PR #1232, opened 2016-10-14
Keep pagination consistent across LTR/RTL (close #1231)

### LINKED ISSUE #1231: Should pagination stay left-to-right in RTL?
See our discussion here for context: https://github.com/mozilla/addons-frontend/pull/1223#discussion_r83313187

Right now the pagination changes Next and Previous around in RTL, but should they be swapped or should we keep Next on the right? @mstriemer points out that Google in Hebrew keep things in the same spots between LTR and RTL.

Paging @designakt for thoughts.

tofumatt:
r? for the pagination being consistent ala Google.

Pinging @atiqueahmedziad, who has done a lot of RTL work for us on the addons-server and might know which is better.

If this is the wrong approach we can close this and leave things as they were, it's here in case we want it. No rush to merge this one because it's a bit of a question 😄 

### Before
screen shot 2016-10-14 at 15 18 44

### After
screen shot 2016-10-14 at 15 17 43

### DIFF
--- a/src/core/components/Paginate/Paginate.scss
+++ b/src/core/components/Paginate/Paginate.scss
@@ -9,6 +9,10 @@
 .Paginator-links {
   display: flex;
   justify-content: space-between;
+
+  [dir=rtl] & {
+    flex-direction: row-reverse;
+  }
 }
 
 .Paginator .Paginator-item {
@@ -25,13 +29,21 @@
 }
 
 .Paginator-previous {
-  &::before {
+  [dir=ltr] &::before {
+    @include arrow-previous();
+  }
+
+  [dir=rtl] &::after {
     @include arrow-previous();
   }
 }
 
 .Paginator-next {
-  &::after {
+  [dir=ltr] &::after {
+    @include arrow-next();
+  }
+
+  [dir=rtl] &::before {
     @include arrow-next();
   }
 }


tofumatt:
lol, the Firefox Responsive mode screenshots are a bit massive, eh?

src/core/components/Paginate/Paginate.scss
```diff
@@ -25,13 +29,29 @@
 }
 
 .Paginator-previous {
-  &::before {
-    @include arrow-previous();
+  [dir=lrt] & {
+    &::before {
```
mstriemer: Could move up a line?

```css
[dir=ltr] &::before { ... }

Comment (tofumatt) [inline src/core/components/Paginate/Paginate.scss]: Hah, right. I was doing something different originally but that makes sense now that I did it this way 👍

File: src/core/components/Paginate/Paginate.scss [hunk 2]
```diff
@@ -25,13 +29,21 @@
 }
 
 .Paginator-previous {
-  &::before {
+  [dir=lrt] &::before {
```
atiqueahmedziad: typo maybe...will it be `ltr` ?

src/core/components/Paginate/Paginate.scss
```diff
@@ -25,13 +29,21 @@
...
+  }
+
+  [dir=rtl] &::after {
     @include arrow-previous();
   }
 }
 
 .Paginator-next {
-  &::after {
+  [dir=lrt] &::after {
```
atiqueahmedziad: Same here

tofumatt:
Hah, yes! Thanks!

designakt:
Looking at other sites besides googles front page I most often see the mirrored. Like in the before version. @brassy- mentioned to have done some work with websites that have also been translated to Hebrew, and they had to mirror pagination as well.

tofumatt:
Sounds like we should keep them as they were then, I'll close this.

PR #1233, opened 2016-10-14
Remove prefer-default-export

### LINKED ISSUE #1230: Remove import/prefer-default-export lint check
I find myself disabling [import/prefer-default-export](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md) quite a bit since we have a lot of utility modules that do not provide a default export. Unless anyone objects, I say we remove this rule from our [config](https://github.com/mozilla/addons-frontend/blob/master/.eslintrc) (the inherited `airbnb` config turned it on) and remove all comments that disable it.

muffinresearch:
Fixes #1230

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -59,6 +59,7 @@
       ],
       "newlines-between": "always"
     }],
+    "import/prefer-default-export": ["off"],
     // This ensures a new line after all import statements.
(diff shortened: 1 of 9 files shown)


tofumatt:
More red than green!
red-celebrate

PR #1234, opened 2016-10-14
WIP: Inline SVG files (fix #1207)

### LINKED ISSUE #1207: Use inline SVG for header so we can style it
As discussed here: https://github.com/mozilla/addons-frontend/pull/1194#discussion_r82698882

tofumatt:
This will inline all SVG files and include them using an `Icon` component instead of `::before` and `::after` CSS attributes. This will allow us to style them.

Right now I'm inlining them into the header of the page, but we should probably just build an SVG file of all the symbols and reference it instead.


mstriemer:
```
 assets/.DS_Store                   | Bin 6148 -> 0 bytes
```

🙌

tofumatt:
I don't even know how that snuck in there!

muffinresearch:
Is is worth closing this one until we get a chance to come back to it if it's not in active review?

tofumatt:
Indeed, closed for now.

PR #1236, opened 2016-10-14
Stop using .development() on server (fix #1235)

### LINKED ISSUE #1235: Don't use `development` method on webpackIsomorphicToolsPlugin
```
[webpack-isomorphic-tools] [error] `.development()` method is now deprecated
(for server-side instance only, not for webpack plugin instance) and has no effect.
Set up a proper `process.env.NODE_ENV` variable instead.
```

tofumatt:
This was bugging me so I stopped it.

Apparently this didn't do anything anymore, which does make me ask: what were we relying on it to do and will it break anything in development?

But everything seems fine so I submit this for review.

### DIFF
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -230,7 +230,8 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
 
 export function runServer({ listen = true, app = appName } = {}) {
   if (!app) {
-    log.fatal(`Please specify a valid appName from ${config.get('validAppNames')}`);
+    log.fatal(
+      `Please specify a valid appName from ${config.get('validAppNames')}`);
     process.exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


muffinresearch:
> Apparently this didn't do anything anymore, which does make me ask: what were we relying on it to do and will it break anything in development?

From  the docs:

> What does .development() method do? It enables development mode. In short, when in development mode, it disables asset caching (and enables asset hot reload), and optionally runs its own "dev server" utility (see port configuration setting). Call it in development webpack build configuration, and, conversely, don't call it in production webpack build configuration.

We run our own dev server, but does this break hot-reloads?

tofumatt:
So there is `.development()` mode in several things and the place I removed it from it was a no-op (in the future it will be removed). I found the docs/messages about that a bit unclear as well.

Again, as it doesn't do anything now it shouldn't break stuff, but I'll check right now to make sure 😊

tofumatt:
Just checked and hot reload still works:

oct-17-2016 15-18-36

muffinresearch:
lgtm r+

PR #1237, opened 2016-10-14
Add a categories page, component, and category search support

tofumatt:
<strike>Still a WIP and missing some tests, though I'm gonna finish that soon.

Will need https://github.com/mozilla/addons-server/issues/3676 to finish.</strike>

------

This adds category support to the frontend. It's a beast of a patch and I'm totally sorry. I owe all reviewers beer at Hawaii.

There are a few style issues that I will happily tweak afterward, but I thought getting this merged before it ballooned further was wise.

I also realised that there isn't a paginator for the categories search results. I can add that here but again would be happy to do it in a follow-up patch to save review bloat.

### Screenshots:

screen shot 2016-11-02 at 16 19 45

screen shot 2016-11-02 at 16 19 49

r?


EnTeQuAk:
The API is now available at `/api/v3/addons/categories/`, e.g `https://addons-dev.allizom.org/api/v3/addons/categories/` for -dev.

tests/client/amo/components/TestCategeories.js
```diff
@@ -0,0 +1,117 @@
...
+  {
+    application: 'android',
+    name: 'Travel',
+    slug: 'travel',
+    type: 'extension',
+  },
+  {
+    application: 'android',
+    name: 'Games',
+    slug: 'games',
```
mstriemer: Should probably have some slugs that aren't just lowercase of the name to be safe.

tests/client/amo/containers/TestHome.js
```diff
@@ -7,11 +7,19 @@ import {
 import { Provider } from 'react-redux';
 
 import createStore from 'amo/store';
-import { getFakeI18nInst } from 'tests/client/helpers';
 import Home from 'amo/containers/Home';
+import { getFakeI18nInst } from 'tests/client/helpers';
 
 
 describe('Home', () => {
+  // function render(props) {
```
mstriemer: 🔥

src/amo/components/Categories.scss
```diff
@@ -0,0 +1,21 @@
...
+.Categories-listItem {
+  background: url('../img/icons/arrow.svg') 100% 50% no-repeat;
+  border-bottom: 2px solid #000;
+  display: block;
+  list-style: none;
+  margin: 0 auto;
+  padding: 0;
+  width: 95%;
+
+  a {
```
tofumatt: No your way is totally better; I've been trying for that too and just missed this one. It means less nesting which is nice.

src/amo/containers/CategoriesPage.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  static defaultProps = {
+    CategoriesComponent: Categories,
+  }
+
+  render() {
+    const { CategoriesComponent, addonType } = this.props;
+    return (
+      <div className="Categories-Page" ref={(ref) => { this.container = ref; }}>
```
tofumatt: I created it solely because it seems like a page and I figured it should have a container in addition to a component.

That said you're right it can be simplified a lot, I've tidied it up where the container does the `asyncConnect` and the component doesn't, and removed the HTML from the container. That makes sense. 👍

src/amo/routes.js
```diff
@@ -14,6 +15,7 @@ export default (
     <IndexRoute component={Home} />
     <Route path="addon/:slug/" component={DetailPage} />
     <Route path="addon/:slug/review/:reviewId/" component={AddonReview} />
+    <Route path="categories/:addonType/" component={CategoriesPage} />
```
tofumatt: If we're aiming for 1:1 URL parity with the old AMO (which I wasn't ever clear on, but I guess we are) then yes.

I guess since the new site will switch over via cookies (and cool URLs don't change) that's best.

src/core/components/Search/SearchResults.js
```diff
@@ -52,23 +71,30 @@ class SearchResults extends React.Component {
...
+    } else if (!loading && searchParamIsPresent && count === 0) {
+      // TODO: Add the extension type, if available, so it says "no extensions"
+      // found that match your search or something.
+      messageText = i18n.gettext('No results were found.');
+    } else if (!searchParamIsPresent && !loading) {
+      messageText = i18n.gettext(
+        "Please enter search terms to search all of Mozilla's Add-ons.");
     }
 
-    const message = messageText ?
```
tofumatt: `messageText` is always set, so right now it is never `null` the code coverage not minding the line being removed proves this.

I could add it back in but there's no code path that hits it now.

mstriemer: Wouldn't the coverage complain if there wasn't a test hitting the else case?

src/core/api/index.js
```diff
@@ -59,12 +60,22 @@ export function callApi({
...
+export function categories({ api }) {
+  return callApi({
+    endpoint: 'addons/categories',
+    schema: { results: arrayOf(categorySchema) },
+    state: api,
+  });
+}
+
+export function search({ api, page, query, auth = false, category, addonType }) {
+  const clientApp = api ? api.clientApp : null;
```
mstriemer: I thought the admin passed `api` for the token?

tofumatt:
Oh, you’re right, in which case now I’m confused. I’ll check it out.

src/amo/routes.js
tofumatt: Oh, actually it's not, just realised.

`categories/:addonType/` lists all the categories for a given addonType (extension, theme, etc.)

I'll add the `s` though 👍

src/core/api/index.js
tofumatt: Turns out it was just a bad test, should be good now.

src/amo/routes.js
mstriemer: Oh, I see there's an extensions home and an extensions category list (likewise for themes).

Does it maybe make sense to reverse this? So we have `/extensions/` and `/extensions/categories/`?

mstriemer: I can see arguments for both ways so just throwing it out there.

tofumatt:
Well the lack of “s” is less code/confusion; the “s” is prettier URLs. But
it’s just a .replace(/s$/, ‘') call right now, so I think it’s worth it.

tofumatt:
Okay, updated things and I think I addressed all the comments. Ready for another r?

muffinresearch:
Re: URLS, we can't deviate from the urls used on AMO until we've done desktop. I've filed https://github.com/mozilla/addons-frontend/issues/1349 to check this along with the rationale behind it.

.eslintrc
```diff
@@ -12,6 +12,7 @@
   },
   "parser": "babel-eslint",
   "rules": {
+    "arrow-body-style": ["off", "always"],
```
mstriemer: I'd generally say do this separately, especially since we had some discussion in an issue about it. Do you mind splitting this out into its own commit before you merge?

src/amo/components/Categories.js
```diff
@@ -0,0 +1,71 @@
...
+import { compose } from 'redux';
+
+import Link from 'amo/components/Link';
+import translate from 'core/i18n/translate';
+
+import './Categories.scss';
+
+
+export function filterAndSortCategories(categories, addonType, clientApp) {
+  return categories.filter((category) => {
```
mstriemer: I really prefer the `()` form here but I seem to be the only one :(

```js
return categories.filter((category) => (
  category.type === addonType && category.application === clientApp;
)).sort((a, b) => a.name > b.name);

File: src/amo/components/Categories.js [hunk 10]
```diff
@@ -0,0 +1,71 @@
...
+    loading: PropTypes.bool.isRequired,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const {
+      addonType, categories, clientApp, error, loading, i18n,
+    } = this.props;
+
+    if (loading && !categories.length) {
```
mstriemer: Should we always do this instead of only when we don't have categories? Why would we have categories and still be loading?

src/amo/components/Categories.js
```diff
@@ -0,0 +1,71 @@
...
+      return <div>{i18n.gettext('Loading...')}</div>;
```
mstriemer: Maybe these `<div>`s should have the `Category` class, and potentially a `<p>` inside of them. That way any general layout stuff is the same.

src/amo/components/Categories.js
```diff
@@ -0,0 +1,71 @@
...
+    return (
+      <div className="Categories">
+        <ul className="Categories-list"
+          ref={(ref) => { this.categories = ref; }}>
+          {categoriesToShow.map((category) => {
+            const queryParams = { category: category.slug, type: addonType };
+            return (
+              <li className="Categories-list-item">
+                <Link className="Categories-link"
+                  to={{ pathname: '/search/', query: queryParams }}>
```
mstriemer: Do we not have a category detail page? There's currently one at https://addons.mozilla.org/en-US/firefox/extensions/search-tools/ and I think this makes way more sense than going to search.

If there's an issue for this that's cool but I feel we should have a detail page that uses this search query in the background.

src/amo/components/CategoryInfo.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  render() {
+    const { addonType, categories, clientApp, slug } = this.props;
+
+    if (!categories) {
+      return null;
+    }
+
+    const categoryMatch = categories.filter((category) => {
```
mstriemer: It feels awkward to keep filtering the categories all over. Can you store them as `{ firefox: { extensions: { search: [] } } }` instead?

src/amo/components/CategoryInfo.js
```diff
@@ -0,0 +1,64 @@
...
+      return (category.application === clientApp && category.slug === slug &&
+        category.type === addonType);
+    });
+    const category = categoryMatch.length ? categoryMatch[0] : false;
```
mstriemer: There's an [`Array.prototype.find()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) that would remove the need for this but restructuring how we store categories is probably better.

src/amo/components/SearchPage.js
```diff
@@ -1,43 +1,51 @@
...
     const paginator = query && count > 0 ?
-      <Paginate LinkComponent={Link} count={count} currentPage={page}
+      <Paginate LinkComponent={LinkComponent} count={count} currentPage={page}
         pathname="/search/" query={{ q: query }} showPages={0} /> : [];
 
     return (
       <div className="search-page">
-        <SearchResults results={results} query={query} loading={loading}
-          count={count} ResultComponent={SearchResult} />
+        <SearchResults CategoryInfoComponent={CategoryInfoComponent}
```
mstriemer: Category info doesn't feel related to search. I'd say it should be its own page that just calls search for its API request.

src/core/actions/search.js
```diff
@@ -4,23 +4,23 @@ import {
   SEARCH_FAILED,
 } from 'core/constants';
 
-export function searchStart(query, page) {
+export function searchStart({ page, query, addonType, app, category }) {
```
mstriemer: This really feels like it's modifying search too much.

.eslintrc
tofumatt: I think an update caused it to start yelling at me so I changed it (or maybe it was just a tonne of warnings).

I will do it if it doesn't cause failures though, sure thing.

src/amo/components/Categories.js
tofumatt: I find the implicit returns harder to follow and also just prefer all functions look the same for consistency. I find implicit return JS to be harder to scan.

src/amo/components/Categories.js
tofumatt: Yeah we should have that–there are URL parity bugs and this would be a case where I should add it.

tofumatt:
Ah, I see. I figured extending the search and allowing it the flexibility to be the one place we'd look for add-ons would be the best option, allowing for really specific searches by both users and admins in the future.

We can totally split it into its own component with likely little duplication, just figured it might be cool to allow a search page with lots of query options.

Parts like including category info in the search page is fair enough, but I felt like any listing of add-ons was essentially a search and would make sense to be treated as one.

My gut reaction is I would rather thing extend search than become its own component. I'll have another look and see if I'm wrong on that and it'd be cleaner separated or if there's at least more than could be split out.

src/core/api/index.js
```diff
@@ -59,12 +60,20 @@ export function callApi({
...
-export function search({ api, page, query, auth = false }) {
+export function categories({ api }) {
+  return callApi({
+    endpoint: 'addons/categories',
+    schema: { results: arrayOf(categorySchema) },
+    state: api,
+  });
+}
+
+export function search({ api, page, query, auth = false, category, addonType }) {
```
mstriemer: I think I'd like this to be more generic for the extra filters. Essentially thinking of it as `search({ query, page, filters })` and `filters` would have the category/addon type in it. Once that's done it might be nicer to wrap this in a `category()` function so it's clear how to get category listings.

```js
export function category({ category, addonType, filters = {}, ...args }) {
  return search({ ...args, filters: { ...filters, category, addonType } });
}

src/amo/routes.js
tofumatt: I think most URLs on the existing site are `:addonType/:somethingSomething` so while I think categories first actually might make more sense, I'll go for consistency given it's a toss-up otherwise anyway.

tofumatt:
Superseded by #1376, which is now merged.

PR #1242, opened 2016-10-17
Add unique staticHost for dev and stage amo APP

jasonthomas:
I copied the current configs for dev and stage disco app and updated the staticHost to reflect the amo app's CDN.

r?


muffinresearch:
r+

PR #1245, opened 2016-10-18
Test with node 6 too

muffinresearch:
Just going to see what travis makes of this to know if we have any compat changes we will need to make for node 6.


muffinresearch:
Travis seems happy with this on first run. I'm going to run this a couple more times and if travis continues to be green I will land it.

* [x] Test run 1
* [ ] Test run 2
* [ ] Test run 3

PR #1257, opened 2016-10-19
Don't rely on DOCKERHUB_REPO env variable during test runs (fix #1254)

### LINKED ISSUE #1254: Circle builds fail for branches that are not on the main repo.
Looks like when we propose branches form our own forks the `$DOCKERHUB_REPO` var isn't set.

Here's an example of the failure.

```
docker build --pull -t $DOCKERHUB_REPO .

invalid value "." for flag -t: Error parsing reference: "." is not a valid repository/tag
See 'docker build --help'.

docker build --pull -t $DOCKERHUB_REPO . returned exit code 125

Action failed: docker build --pull -t $DOCKERHUB_REPO .
```

jasonthomas:
r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,7 +33,7 @@ dependencies:
     # day to keep the cache size down.
     - I="image-$(date +%j).gz"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/docker/$I | docker load; fi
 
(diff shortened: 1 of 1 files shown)


tofumatt:
wfm, I suppose we could set the variable instead, but I'm fine with this if we can't just set the variable.

PR #1263, opened 2016-10-20
Translate accept props

mstriemer:
This updates `translate()` to accept `i18n` from props which removes the need for setting up an `I18nProvider` component in many cases. I refactored the tests for `I18nProvider` since I think they were mostly testing `translate()` and moved those tests into their own file.

It also removes the `I18nProvider` and `Provider` wrappers from tests that did not need it.


kumar303:
Nice!

taco

PR #1268, opened 2016-10-24
Update homepage

### LINKED ISSUE #1264: Implement the homepage category layout
Relates to https://github.com/mozilla/addons-frontend/issues/797

Implement the body of the homepag with static cat or search links. We might not have the links initially so will start with just the layout and go from there.

amo_-_v1_1__sketch

muffinresearch:
Fixes #1264

* Updates the colours
* Adds a rough and ready (unoptimised) first cut of the homepage design.

Will need further iterations but it's a good starting point.

Looks like this:

screen shot 2016-10-24 at 20 20 40


src/amo/css/App.scss
```diff
@@ -66,7 +66,7 @@ a:link {
   @include font-medium();
 
   font-size: 14px;
-  color: #0996f8;
+  color: #0568ba;
```
mstriemer: Can you move this into vars, please?

tests/client/amo/containers/TestHome.js
```diff
@@ -4,13 +4,14 @@ import {
...
 import Home from 'amo/containers/Home';
 
 
 describe('Home', () => {
   it('renders a heading', () => {
-    const root = renderIntoDocument(<Home />);
+    const root = renderIntoDocument(<Home i18n={getFakeI18nInst()} />);
     const rootNode = findDOMNode(root);
-    assert.include(rootNode.querySelector('h1').textContent, 'AMO Home Page');
+    assert.include(rootNode.querySelectorAll('h2')[0].textContent, 'What do you want Firefox to do?');
```
mstriemer: Could leave this as `querySelector` if you're just checking the first one. Doing `.map((el) => el.textContent)` and checking `['What do you want Firefox to do?', 'How do you want Firefox to look?']` might be worth testing if this is worth testing though.

src/amo/containers/Home.js
```diff
@@ -1,11 +1,45 @@
...
+  }
 
-export default class Home extends React.Component {
   render() {
+    const { i18n } = this.props;
     return (
-      <div>
-        <h1>AMO Home Page Hello World</h1>
+      <div className="HomePage">
+        <h2 className="HomePage-h2">{i18n.gettext('What do you want Firefox to do?')}</h2>
```
mstriemer: Can this just be `-heading` or `-subheading` instead of `-h2`?

src/amo/containers/Home.js
```diff
@@ -1,11 +1,45 @@
...
+        <ul className="HomePage-cat-list">
+          <li className="HomePage-block-ads"><a href="#block-ads"><span>{i18n.gettext('Block Ads')}</span></a></li>
```
mstriemer: Is the `<span>` needed? This should probably use `Link` and a stateless component could DRY it up a bit:

```jsx
const CatLink = ({ children, name, ...props }) => (
  <li className={`HomePage-${name}`}><Link {...props}>{children}</Link></li>
);
// ...
<ul className="HomePage-cat-list">
  <CatLink name="block-ads" href="#block-ads">{i18n.gettext('Block Ads')}</CatLink>
</ul>

File: src/amo/css/Home.scss [hunk 5]
```diff
@@ -0,0 +1,146 @@
...
+    bottom: 0;
+    font-size: 10px;
+    left: 0;
+    position: absolute;
+    right: 0;
+    text-decoration: none;
+    top: 0;
+  }
+
+  span {
```
mstriemer: If this is needed can you give it a class?

src/amo/css/Home.scss
```diff
@@ -0,0 +1,146 @@
...
+  background: #fff;
+  border-radius: 6px;
+  font-size: 10px;
+  margin: 0 auto 20px;
+  padding: 10px;
+  text-align: center;
+  text-decoration: none;
+  text-transform: uppercase;
+  width: calc(100% - 40px);
+  display: flex;
```
mstriemer: 🍟

src/amo/containers/Home.js
mstriemer: This should probably be `Block ads` since the other ones are `Share stuff`.

src/amo/containers/Home.js
```diff
@@ -1,11 +1,45 @@
...
+          <li className="HomePage-screenshot"><a href="#screenshot"><span>{i18n.gettext('Screenshot')}</span></a></li>
+          <li className="HomePage-save-stuff"><a href="#save-stuff"><span>{i18n.gettext('Save stuff')}</span></a></li>
+          <li className="HomePage-shop-online"><a href="#shop-online"><span>{i18n.gettext('Shop online')}</span></a></li>
+          <li className="HomePage-be-social"><a href="#share-stuff"><span>{i18n.gettext('Be social')}</span></a></li>
+          <li className="HomePage-share-stuff"><a href="#share-stuff"><span>{i18n.gettext('Share stuff')}</span></a></li>
+        </ul>
+        <a className="HomePage-extensions-link" href="#thems">{i18n.gettext('Browse all extensions')}</a>
```
mstriemer: Should be `#extensions` not `#thems` and use `Link`.

src/amo/containers/Home.js
```diff
@@ -1,11 +1,45 @@
...
+        <h2 className="HomePage-h2">{i18n.gettext('How do you want Firefox to look?')}</h2>
+        <ul className="HomePage-cat-list">
+          <li className="HomePage-wild"><a href="#wild"><span>{i18n.gettext('Wild')}</span></a></li>
+          <li className="HomePage-abstract"><a href="#abstract"><span>{i18n.gettext('Abstract')}</span></a></li>
+          <li className="HomePage-fashionable"><a href="#fashionable"><span>{i18n.gettext('Fashionable')}</span></a></li>
+          <li className="HomePage-scenic"><a href="#scenic"><span>{i18n.gettext('Scenic')}</span></a></li>
+          <li className="HomePage-sporty"><a href="#sporty"><span>{i18n.gettext('Sporty')}</span></a></li>
+          <li className="HomePage-mystical"><a href="#mystical"><span>{i18n.gettext('Mystical')}</span></a></li>
+        </ul>
+        <a className="HomePage-themes-link" href="#thems">{i18n.gettext('Browse all themes')}</a>
```
mstriemer: Should be `#themes` and use `Link`.

tests/client/amo/containers/TestHome.js
muffinresearch: Yeah that sounds like a plan.

src/amo/containers/Home.js
muffinresearch: I'm not convinced moving to a stateless component buys us that much quite yet. Leaving it static may help if we need to move things around and make changes. 

As soon as we decide base this on a data structure we can iterate over then that would make more sense in terms of DRY.

muffinresearch: Moving to Link didn't pose a problem re: child markup so I've left the span in for now since it simplifies the positioning.

src/amo/css/Home.scss
muffinresearch: potato

src/amo/css/App.scss
muffinresearch: I'm going to file a separate issue for this. Seems we don't have any way to set vars for just this app - not sure if that got removed or we never had it - I don't think we should only have core vars. That needs to be fixed and I don't want to block on that for this patch.

Issue #1056, opened 2016-09-12
Document some simple guidelines for our approach to CSS 


muffinresearch:
A few times the question of whether we should adopt some kind of methodology for CSS has come up. If we go this route here's some likely goals of adopting anything along these lines:
- Simplicity - it can't be too complex or we just won't use it.
- Consistency - consistency is king. 
- Sane specificity. E.g. avoiding the kind of crazy specifity hacks we needed to do in the style-refresh.

There are lots of methodologies that exist e.g. ([SMACSS](https://smacss.com/), [BEM](http://getbem.com/) and [suitcss](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md)). Most look to provide some structure to classnames to aid consistent code modularization.

We don't have to opt for any one specific approach necessarily, but whatever we decide would be good to document how we are going to build our component classnames (or if we are wholesale adopting something that exists we can just point to their docs instead).

Bonus points for stylelint rules to help maintain whatever the end result is.

kumar303:
The only concrete one I can think of right now is that we should enforce all component class names to be prefixed with the component name. Sadly, I don't see an existing eslint rule for this but it probably wouldn't be too hard to write one.

muffinresearch:
Are we going to make a distinction between descendants and modifiers?

kumar303:
I personally don't really see the point in making a distinction between descendants and modifiers. [The rationale](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#components) listed in SuitCSS doesn't seem compelling. The only compelling point is specificity but I think we already gain that by prefixing component classes with the component name.

muffinresearch:
> I personally don't really see the point in making a distinction between descendants and modifiers. The rationale listed in SuitCSS doesn't seem compelling. The only compelling point is specificity but I think we already gain that by prefixing component classes with the component name.

Yep, that seems reasonable to me; the simpler the better.

kumar303:
On a closer look at SuitCSS I see it tries to standardize shared css classes with the `u-` (for utility) prefix. I ignored this section before because they called one of their classes `u-floatleft` which sounded gross 🔥  We might need a way to prefix shared classes but we could look at it more on a case by case basis.

mstriemer:
I used the modifier in the updated `SearchInput` component because I think it makes things clearer. @tofumatt mentioned this in https://github.com/mozilla/addons-frontend/pull/1269#discussion_r84780256.

[...]

PR #1269, opened 2016-10-24
Match search input to mocks

mstriemer:
This still needs some cleanup but figured I'd post it for UX comments. @pwalm how does this look?

I'm not crazy about how the cursor appears to the left and then the icon slides over, I'd rather they move together but I couldn't think of a great way to do that.

64px seemed like a lot of padding so I did 16px on each side. Is it supposed to be 64px total? This would be 58px of padding on the input right now (16 \* 2 + 26 [I don't know why the icon is 26px, I guess this should be smaller]).

search-input-short mov

There are 3 elements working together: a full width input that is always left aligned (nothing happens with this so the input works as expected), a label that has the icon and placeholder text, and an icon that does the animation.
- When empty and unfocused the label is shown and centre aligned, JS will position the extra icon underneath the icon you see on load and resize but the extra icon is `opacity: 0` so resizing looks normal.
- When focused the label is hidden using `color: transparent` and `opacity: 0` on its icon, the extra icon immediately gets `opacity: 1` and animates to the left.
- When blurred and there is no text entered the icon animates back, the two icons have matching transitions on opacity with a delay so that once the animation ends the extra icon animates out and the label icon animates in.


tofumatt:
I actually like that sliding search box fwiw, and I think it works well.

src/ui/components/SearchInput/style.scss
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+}
+
+.SearchInput-label {
+  @include font-regular();
+  color: #ccc;
+  display: inline-block;
+  height: 3rem;
+  line-height: 3rem;
```
tofumatt: I thought the usual rule for line-height was to [just use a number and not a unit](http://meyerweb.com/eric/thoughts/2006/02/08/unitless-line-heights/). (Of course, the label won't have children, right? So it matters less but for consistency's sake I'd like unitless.)

src/ui/components/SearchInput/style.scss
```diff
@@ -0,0 +1,90 @@
...
+    color: #fff;
+    opacity: 1;
+    transition: transform 250ms;
+
+    [dir=ltr] & {
+      transform: translateX(0) !important;
+    }
+
+    [dir=rtl] & {
+      // transform: translateX(-32px);
```
tofumatt: 🔥

src/ui/components/SearchInput/index.js
```diff
@@ -0,0 +1,69 @@
...
+        <input
+          {...props} className="SearchInput-input" placeholder={placeholder} id={id} name={name}
+          autoComplete="off" ref={this.inputRefs(inputRef)}
+          onFocus={this.onFocus} onBlur={this.onBlur} />
+      </div>
+    );
+  }
+}
+
+export default SearchInput;
```
tofumatt: You could move the `export default` above and ditch this line.

src/ui/components/SearchInput/index.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+    };
+  }
+
+  render() {
+    const { className, name, inputRef, placeholder, ...props } = this.props;
+    const id = `SearchInput-input-${name}`;
+    return (
+      <div className={classNames(className, 'SearchInput')} ref={(el) => { this.root = el; }}>
+        <i className="Icon-magnifying-glass" ref={(el) => { this.icon = el; }} />
```
tofumatt: I know they aren't in place yet but our thoughts of CSS rules (#1056) seem to say we should keep our component name as a common class prefix, so this should be `SearchInput` and not `Icon`.

src/ui/components/SearchInput/index.js
```diff
@@ -0,0 +1,69 @@
...
+    window.addEventListener('resize', this.setIconPosition);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.setIconPosition);
+  }
+
+  onBlur = () => {
+    if (!this.input.value) {
+      this.root.classList.remove('SearchInput--text');
```
tofumatt: Are we using `--` anywhere else? It seems out of place, though I know there was a thought behind doing it.

src/amo/components/SearchForm.js
```diff
@@ -33,10 +34,10 @@ export class SearchFormBase extends React.Component {
     return (
       <form method="GET" action={pathname} onSubmit={this.handleSearch}
             className="SearchForm-form" ref={(ref) => { this.form = ref; }}>
-        <label className="visually-hidden" htmlFor="q">{_('Search')}</label>
-        <input ref={(ref) => { this.searchQuery = ref; }} type="search" name="q"
-               placeholder={_('Search extensions and themes')}
-               defaultValue={query} className="SearchForm-query" />
+        <SearchInput
+          inputRef={(ref) => { this.searchQuery = ref; }} type="search" name="q"
+          defaultValue={query} className="SearchForm-query" offset={31}
```
tofumatt: What does `offset` do? I couldn't find it elsewhere in the PR.

src/ui/components/SearchInput/style.scss
mstriemer: I was setting this to centre the text vertically since the input bar is `3rem`s high. It might not be needed though, I'll at least extract this to a variable if it's still needed.

src/ui/components/SearchInput/index.js
mstriemer: I'm not sure if we're using it anywhere else but it feels wrong to call it `SearchInput-text` since there's also a `SearchInput-input`.

I know we had discussed using something like [suitcss's naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md) but we weren't definite on the descendant and utility class name conventions. It would seem I agree with their conventions.

src/amo/components/SearchForm.js
mstriemer: Leftover from hacks in `CentredInput`. I will remove it.

pwalm:
- Can the cursor appear after the slide animation instead of together?
- Should be 64px@xxhdi to the left of the magnifying glass. Might seem big because you're working at xhdi or hdi? (hdi would be 22px)
- Mag icon should be 32px@xxhdi

Looks really good, otherwise. Love the smooth transition.

mstriemer:
CSS pixels are at 96DPI then scaled based on screen density and typical viewing distance. Device specific but that's roughly accurate.

I made the icon 16px since I figured it should be half of xxhdi. I wasn't sure of what padding to use so here are some screenshots of different padding amounts. Let me know what works best and I'll go with that.

`padding-start / icon size / padding-end`
screenshot 2016-10-26 09 56 59
screenshot 2016-10-26 09 57 21
screenshot 2016-10-26 09 57 50
screenshot 2016-10-26 09 58 27


> Can the cursor appear after the slide animation instead of together?

I have a couple ideas. Neither of them are pretty, but I'll give it a go.

pwalm:
24 / 16 / 12 looks the best to me. Thanks for mocking up those options, Mark.

mstriemer:
Updated look, the focus is delayed until the transition ends using the lesser of two hacks.

searchinput mov

@tofumatt this should probably get another look at it, I made several changes and I think I addressed all of your comments.

src/ui/components/SearchInput/index.js
```diff
@@ -0,0 +1,88 @@
...
+        this.animateIcon.removeEventListener('transitionend', setFocus);
+      };
+      this.animateIcon.addEventListener('transitionend', setFocus);
+    }
+  }
+
+  setIconPosition = () => {
+    const { left: labelLeft } = this.labelIcon.getBoundingClientRect();
+    const { left: animateLeft } = this.animateIcon.getBoundingClientRect();
+    this.animateIcon.style.transform = `translateX(${labelLeft - animateLeft}px)`;
```
tofumatt: I forget if I already asked but this is correct in RTL, right?

src/ui/components/SearchInput/style.scss
```diff
@@ -0,0 +1,91 @@
...
+    & .Icon-magnifying-glass {
+      opacity: 0;
+      transition: none;
+    }
+  }
+
+  .SearchInput-animation-icon {
+    color: #fff;
+    opacity: 1;
+    transform: translateX(0) !important;
```
tofumatt: If this `!important` is really needed, a comment explaining why would be handy; it will surely prevent a refactoring future developer from trying to hunt its purpose down trying to remove it. 😉

src/ui/components/SearchInput/index.js
mstriemer: I believe so but I'll double check. I remember trying to figure out how to handle the RTL case and then I think avoiding it entirely and doing the difference of x positions worked in both cases because `translateX()` is always from the left.

src/ui/components/SearchInput/style.scss
mstriemer: It's needed because the regular position is set using inline styles. I could remove the value instead but this felt cleaner.

mstriemer:
I added a comment that the `!important` is to overwrite inline styles.

RTL works as expected:
searchinput-rtl mov

tofumatt:
Sweet, looks good to merge!

PR #1270, opened 2016-10-24
chore(package): update eslint-plugin-react to version 6.4.1

mstriemer:
Greenkeeper change with lint fix.

https://greenkeeper.io/
Fixes #1204

### DIFF
--- a/src/amo/components/MastHead.js
+++ b/src/amo/components/MastHead.js
@@ -35,13 +35,13 @@ export class MastHeadBase extends React.Component {
         </div>
         {children}
         <header className="MastHead-header">
-          { isHomePage ?
-            <h1 ref={(ref) => { this.title = ref; }}
-                className="MastHead-title MastHead-homepage">
-              { headerTitle }
-            </h1> :
-            <a ref={(ref) => { this.title = ref; }}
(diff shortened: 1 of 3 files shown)


src/amo/components/MastHead.js
```diff
@@ -35,13 +35,13 @@ export class MastHeadBase extends React.Component {
...
         <header className="MastHead-header">
-          { isHomePage ?
-            <h1 ref={(ref) => { this.title = ref; }}
-                className="MastHead-title MastHead-homepage">
-              { headerTitle }
-            </h1> :
-            <a ref={(ref) => { this.title = ref; }}
-               href="/" className="MastHead-title">{ headerTitle }</a> }
+          {isHomePage
+            ? <h1 ref={(ref) => { this.title = ref; }} className="MastHead-title MastHead-homepage">
```
tofumatt: Huh, interesting. I wouldn't think to do it this way, but alright!

mstriemer: I originally refactored into an `if` and `let` but it was enormous so I figured there must be a simpler way. This makes it more clear that there's a ternary going on I guess.

tofumatt:
Yeah I definitely prefer this to an if statement.

PR #1308, opened 2016-10-27
Remove padding compensation for FF50+

muffinresearch:
Removes the padding if the browser version is >= 50 -  uses the https://github.com/linagora/mozilla-version-comparator lib for comparisons.

/en-US/firefox/discovery/pane/**49**/Darwin/normal
discover_add-ons

/en-US/firefox/discovery/pane/**50**/Darwin/normal
discover_add-ons


muffinresearch:
> Seriously though looks good. "comparator" is an odd choice given it's usually a thing that compares two things, if my electrical engineering isn't too rusty. ⚡️

Well to be fair the lib does compare two browser versions, v1 vs v2 - https://github.com/linagora/mozilla-version-comparator/blob/master/mozilla-version-comparator.js#L119

tofumatt:
Oh, so it does. I clearly spaced.

muffinresearch:
> If you use parseInt(browserVersion, 10) < 50 instead then the tests all pass except for the bogus and undefined ones. It seems correct to me that we wouldn't set the padding though because we know browsers less than 50 give us the version string so it's more likely that the padding isn't required than it is IMO. So those tests should be flipped around.

@mstriemer Oh interesting point, I didn't realise parseInt was so liberal to give something back as long as the first char is a number - TIL!. 

I agree re: flipping the fallback around, I was thinking the exact same thing just after landing it. 

I'm just checking we've got nothing coming up that needs a point version comparison if not I think using parseInt makes total sense and I'll change this and drop the dep if we need more (we probably will in AMO) we can add this dep back as and when. Thanks for the feedback. 

Whilst I'm at it I'm going to double-check analytics re: screen widths to make sure the breakpoint is going to be OK. Expect a follow-up branch.

PR #1309, opened 2016-10-27
Add Link component for auto-URL-prefixes (fix #1136)

### LINKED ISSUE #1136: Work out the best way to deal with link prefixes
I added a link to the homepage '/' with the locale prefixing this will result in a redir to `/en-US/firefox/`.

Can we deal with this with routing links? Or, should we set a base element to point to `/%locale%/%appname%`

The goal is whenever we link something in the app if the locale (e.g. `en-US`) and app prefix (e.g:  `firefox`, `android`) is set links should automatically be prefixed with those paths so we don't redirect on every new link.

tofumatt:
Finally got around to doing this as I was being driven mad by all the `lang` passing.

Moves all of our AMO links to use a local `<Link />` component that still uses `to` instead of the propsed `href` in #1136 (so it looks the same as other React links).
r?

### DIFF
--- a/src/amo/components/Footer.js
+++ b/src/amo/components/Footer.js
@@ -1,8 +1,8 @@
 import React, { PropTypes } from 'react';
-import { Link } from 'react-router';
 import { compose } from 'redux';
 
 import LanguagePicker from 'amo/components/LanguagePicker';
+import Link from 'amo/components/Link';
 import translate from 'core/i18n/translate';
 
 import './Footer.scss';
@@ -11,25 +11,22 @@ import './Footer.scss';
 export class FooterBase extends React.Component {
   static propTypes = {
     i18n: PropTypes.object.isRequired,
-    lang: PropTypes.string.isRequired,
   }
 
   render() {
-    const { i18n, lang } = this.props;
+    const { i18n } = this.props;
 
     return (
       <footer className="Footer">
-        <LanguagePicker lang={lang}
-          ref={(ref) => { this.languagePicker = ref; }} />
+        <LanguagePicker ref={(ref) => { this.languagePicker = ref; }} />
         <ul className="Footer-links">
           <li>
-            <Link to={`/${lang}/firefox/privacy`}
-              className="Footer-link">
+            <Link to={'/privacy/'} className="Footer-link">
               {i18n.gettext('Privacy policy')}
             </Link>
           </li>
           <li>
-            <Link to={`/${lang}/firefox/legal`} className="Footer-link">
+            <Link to={'/legal/'} className="Footer-link">
               {i18n.gettext('Legal notices')}
             </Link>
           </li>
--- a/src/amo/components/Link.js
+++ b/src/amo/components/Link.js
@@ -1,0 +1,46 @@
+import path from 'path';
+
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+import { Link } from 'react-router';
+
+
+export class LinkBase extends React.Component {
+  static propTypes = {
+    base: PropTypes.string,
+    children: PropTypes.node,
+    to: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
+  }
+
+  static defaultProps = {
+    base: '',
+  }
+
+  render() {
+    const { base, children, to } = this.props;
+
+    let linkTo = to;
+    if (typeof to === 'string' && to.startsWith('/')) {
+      linkTo = path.join(base, to);
+    } else if (to && to.pathname) {
+      linkTo = {
+        ...to,
+        pathname: to.pathname.startsWith('/') ?
+          path.join(base, to.pathname) : to.pathname,
+      };
+    }
+
+    return <Link {...this.props} to={linkTo}>{children}</Link>;
+  }
+}
+
+export function mapStateToProps(state) {
+  return {
+    base: `/${state.api.lang}/${state.api.clientApp}`,
+  };
+}
+
+export default compose(
+  connect(mapStateToProps),
+)(LinkBase);
--- a/src/amo/components/SearchPage.js
+++ b/src/amo/components/SearchPage.js
@@ -1,15 +1,17 @@
 import React, { PropTypes } from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
 
+import Link from 'amo/components/Link';
 import Paginate from 'core/components/Paginate';
 import SearchResults from 'core/components/Search/SearchResults';
 
 import SearchResult from './SearchResult';
 
 
-export default class SearchPage extends React.Component {
+export class SearchPageBase extends React.Component {
   static propTypes = {
     count: PropTypes.number,
-    lang: PropTypes.string.isRequired,
     loading: PropTypes.bool.isRequired,
     page: PropTypes.number,
     results: PropTypes.array,
@@ -17,17 +19,25 @@ export default class SearchPage extends React.Component {
   }
 
   render() {
-    const { count, lang, loading, page, query, results } = this.props;
-    const pathname = `/${lang}/firefox/search/`;
+    const { count, loading, page, query, results } = this.props;
     const paginator = query && count > 0 ?
-      <Paginate count={count} pathname={pathname} query={{ q: query }}
-        currentPage={page} showPages={0} /> : [];
+      <Paginate LinkComponent={Link} count={count} currentPage={page}
+        pathname="/search/" query={{ q: query }} showPages={0} /> : [];
+
     return (
       <div className="search-page">
         <SearchResults results={results} query={query} loading={loading}
-          count={count} ResultComponent={SearchResult} lang={lang} />
+          count={count} ResultComponent={SearchResult} />
         {paginator}
       </div>
     );
   }
 }
+
+export function mapStateToProps(state) {
+  return { clientApp: state.api.clientApp, lang: state.api.lang };
+}
+
+export default compose(
+  connect(mapStateToProps),
+)(SearchPageBase);
--- a/src/amo/containers/Home.js
+++ b/src/amo/containers/Home.js
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 import { compose } from 'redux';
-import { Link } from 'react-router';
 
+import Link from 'amo/components/Link';
 import translate from 'core/i18n/translate';
 
 import 'amo/css/Home.scss';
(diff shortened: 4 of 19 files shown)


src/amo/components/Footer.js
```diff
@@ -11,25 +11,22 @@ import './Footer.scss';
...
           <li>
-            <Link to={`/${lang}/firefox/privacy`}
-              className="Footer-link">
+            <Link to={'/firefox/privacy'} className="Footer-link">
               {i18n.gettext('Privacy policy')}
             </Link>
           </li>
           <li>
-            <Link to={`/${lang}/firefox/legal`} className="Footer-link">
+            <Link to={'/firefox/legal'} className="Footer-link">
```
muffinresearch: Presumably this should get the app (firefox) removed? Also looks like you'll get a double slash if `to` has a leading slash.

tofumatt:
Oh shoot, good catch.

src/amo/components/Footer.js
tofumatt: You're right, I wanted to emulate `Link` as much as possible so I figured the leading slash should be in `to`. I'll fix it up so that's the case, but if people disagree it can be easily changed.

src/amo/components/Link.js
```diff
@@ -0,0 +1,27 @@
...
+  static propTypes = {
+    base: PropTypes.string,
+    children: PropTypes.node,
+    to: PropTypes.string.isRequired,
+  }
+
+  render() {
+    const { base, children, to } = this.props;
+
+    return (<Link {...this.props} to={`${base}${to}`}>{children}</Link>);
```
mstriemer: The `()` shouldn't be necessary if this fits on one line.

src/amo/components/Link.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+export function mapStateToProps(state) {
+  return { base: `/${state.api.lang}/${state.api.clientApp}` };
```
mstriemer: I think this looks clearer across 3 lines even if it's only one thing.

src/amo/components/SearchPage.js
```diff
@@ -17,17 +20,28 @@ export default class SearchPage extends React.Component {
   }
 
   render() {
-    const { count, lang, loading, page, query, results } = this.props;
-    const pathname = `/${lang}/firefox/search/`;
+    const {
+      clientApp, count, lang, loading, page, query, results,
+    } = this.props;
+    const pathname = `/${lang}/${clientApp}/search/`;
```
mstriemer: I think if you make `Paginate` accept a `Link` prop or use our internal `Link` component you could just pass it `/search/` and remove the redux stuff from this component.

src/amo/containers/Home.js
```diff
@@ -24,7 +24,8 @@ export class HomePageBase extends React.Component {
           <li className="HomePage-be-social"><Link to="#share-stuff"><span>{i18n.gettext('Be social')}</span></Link></li>
           <li className="HomePage-share-stuff"><Link to="#share-stuff"><span>{i18n.gettext('Share stuff')}</span></Link></li>
         </ul>
-        <Link className="HomePage-extensions-link" href="#extensions">{i18n.gettext('Browse all extensions')}</Link>
+        <Link className="HomePage-extensions-link"
```
mstriemer:
```jsx
<Link ...>
  {text}
</Link>
```

Probably reads a bit nicer.

tofumatt:
On the subject of supported `Link` args:

Absolute paths are what we've used so far and what is encouraged. The docs actually state:

> If it's a string it represents the absolute path to link to, e.g. /users/123 (relative paths are not supported).

So if it's a string we just prepend `base` and then pass it through to `<Link>`. It will take relative paths, and that's why I allowed `base` overrides (in case someone wants to make a relative link for whatever reason). But realistically it's discouraged and we wouldn't need it.

But the pathname bit I forgot about. I'm adding that now (needed for the Paginate upgrade anyway!) and will add tests and push. Cheers!

muffinresearch:
I think it would make sense to call our prefixed `Link` something other than `Link` e.g. `PrefixedLink`. This way it makes it more obvious when we're using our link versus the default in case we needed to use the original for something? It also makes it clearer not to expect the exact same arg handling.

Another use case I've seen on AMO, which is maybe not necessary to worry about just now since it's a bit of a minor detail / edge case, is preserving the lang for external mozilla properties. It's quite nice if the user is looking at our site with a chosen lang to retain the language in the link to some other moz property so that site doesn't need to fall-back to an accept-language based redirect which might not provide the same result if they'd chosen a specific lang.

In addition iirc there are parts of AMO that have the lang prefix but no user app in the url (e.g. no firefox or android). The best bet is probably to just update this when we need it though.

tofumatt:
Yeah making this more flexible we can do in the future.

Right now the args ARE the exact same, I figured that would be nice for:

* easily recognisable for existing React coders
* easier to swap out

We can prefix it, but I don’t think we’d ever want a NON prefixed link, so
to me calling it “InternalLink” is just extra typing.

muffinresearch:
The args are the same but we're always prefixing so the `to` passed to the real `Link` is always absolute.

muffinresearch:
I'm wondering if we should be blowing up if the `to` prop doesn't start with '/' otherwise wouldn't input like `../foo`  replace the prefix?

muffinresearch:
> The args are the same but we're always prefixing so the to passed to the real Link is always absolute.

Ah I got confused by https://github.com/mozilla/addons-frontend/pull/1309#pullrequestreview-6150943

The official docs say:

> If it's a string it represents the absolute path to link to, e.g. /users/123 (relative paths are not supported).

src/amo/components/Link.js
```diff
@@ -0,0 +1,27 @@
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+import { Link } from 'react-router';
+
+
+export class InternalLinkBase extends React.Component {
```
kumar303: nit: could this just be `LinkBase`? That's the convention we've [mostly] been following since the module is `Link.js`.

src/amo/components/Link.js
kumar303: [path.join()](https://nodejs.org/api/path.html#path_path_join_paths) or `path.normalize()` might be handy to cope with double slashes when joining `base` and `to`.

src/amo/components/Link.js
kumar303: oh, whoops, nevermind. I see that it needs to work with `Link` from `react-router`

tofumatt: I don't mind calling it `LinkBase`, no reason not to and as mentioned before, I don't think it needs to be prefixed with `Internal` actually.

src/amo/components/Link.js
tofumatt: Good call!

tofumatt: lol train wifi

tofumatt:
Okay, I've addressed all the comments in here I think, another r?

src/amo/components/Link.js
```diff
@@ -0,0 +1,44 @@
...
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+import { Link } from 'react-router';
+
+
+export class LinkBase extends React.Component {
+  static propTypes = {
+    base: PropTypes.string,
+    children: PropTypes.node,
+    to: PropTypes.object.isRequired,
```
muffinresearch: Can be an object or a string?

src/amo/components/Link.js
```diff
@@ -0,0 +1,44 @@
...
+
+    let linkTo;
+    if (typeof to === 'string') {
+      linkTo = path.join(base, to);
+    } else if (to && to.pathname) {
+      linkTo = { ...to, pathname: path.join(base, to.pathname) };
+    } else {
+      linkTo = to;
+    }
+
```
muffinresearch: I'm not too sure about the use-case for supporting the object approach.

It might be worth making sure the string output for LinkTo still starts with base after the join since it would be odd if the base part could be clobbered, for that you'd just supply a different base prop.

src/amo/components/Link.js
tofumatt: Oh, yeah, I figured `object` was the catch-all. I just looked up exactly how `PropTypes` work and will fix.

`Link` actually allows undefined `to` prop that will not output a `href`... dunno what type that should be.

src/amo/components/Link.js
tofumatt: I believe I do test for that, but the object approach is used in Paginate so it's worth supporting.

src/amo/components/Link.js
muffinresearch: > Link actually allows undefined to prop that will not output a href... dunno what type that should be.

I'd hope we'd not ever use that, but yeah I suppose that should still work.

tests/client/amo/components/TestSearchResult.js
```diff
@@ -4,15 +4,21 @@ import {
...
 import { getFakeI18nInst } from 'tests/client/helpers';
 
 
 describe('<SearchResult />', () => {
   function renderResult(result) {
+    const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
+
     return findRenderedComponentWithType(renderIntoDocument(
-      <SearchResult i18n={getFakeI18nInst()} result={result} lang="en-US" />
+      <Provider store={createStore(initialState)}>
```
mstriemer: Too bad we need to wrap all this stuff with `<Provider>` now. I guess there's always a trade off.

It might be kind of nice to simplify this though (not here just thinking out loud):

```jsx
function renderWithProvider(node, { store = createStore() }) {
  const name = node.type;  // I'm not sure if this works?
  return findRenderedComponentWithType(renderIntoDocument((
    <Provider store={store}>{node}</Provider>
  ), name)).getWrappedInstance();
}

Comment (tofumatt) [inline tests/client/amo/components/TestSearchResult.js]: I thought about doing that and will file an issue for it; it's a fair bit of boilerplate but I tried my hand at it really quickly and it needed some tweaking, and I tend to prefer writing easy app code over easy test code 😅

Comment (tofumatt) [inline tests/client/amo/components/TestSearchResult.js]: Wow there are too many "ands" in that statement. I need dinner.

File: tests/client/amo/components/TestLink.js [hunk 10]
```diff
@@ -0,0 +1,70 @@
...
+
+  it('normalizes the path with a string', () => {
+    const root = render({ base: '/foo/', to: '/test' });
+
+    assert.equal(
+      findRenderedComponentWithType(root, Link).props.to, '/foo/test');
+  });
+
+  it('normalizes the path with an object', () => {
+    const root = render({ base: '/foo/', to: { pathname: '/test' } });
```
tofumatt: Hmm, I guess we could do that. Alright, I'll do it and add a test. We aren't using relative links anywhere but I suppose that's a more expected behaviour. URLs are weird.

tofumatt: I actually think the best API would just be to prepend everything and make it so the leading `/` in to `to` is meaningless but just a style thing, but I think that's ultimately bad.

tofumatt:
Yeah, agreed. I’m finishing the patch up and I’m just gonna pass it through.

PR #1321, opened 2016-10-31
Make footer stick to bottom of the viewport

### LINKED ISSUE #1320: Footer background cut off before end of viewport
I believe when I added the footer this wasn't the case, so this might be a regression.

Screenshot here:

screenshot 2016-10-30 22 40 37

The black colour should go until the end of the viewport but doesn't.

muffinresearch:
Fixes #1320

Before:

add-ons_for_firefox

After:

add-ons_for_firefox

### DIFF
--- a/src/amo/components/Footer.scss
+++ b/src/amo/components/Footer.scss
@@ -4,6 +4,7 @@
   background: $footer-color;
   color: $text-color-message;
   display: block;
+  justify-content: flex-end;
   padding: 10px 10px 20px;
 
   &::after {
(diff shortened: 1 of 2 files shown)


src/amo/components/Footer.scss
```diff
@@ -4,6 +4,7 @@
   background: $footer-color;
   color: $text-color-message;
   display: block;
+  justify-content: flex-end;
```
tofumatt: Oh, haha, ignore my comment about this being hard, I forgot about flex.

Issue #1286, opened 2016-10-25
Implement the extensions home page


muffinresearch:
See the sketch file for more info:

![amo_-_v1_sketch](https://cloud.githubusercontent.com/assets/1514/19685446/9a4605e2-9ab4-11e6-998c-38158b09c356.png)

ValentinaPC:
Verified on AMO-dev FF51(Android 6.0.1)
Video: 
![videotogif_2017 02 07_09 58 04](https://cloud.githubusercontent.com/assets/15685960/22682772/5b3aa3ec-ed1d-11e6-98fe-7270ff055317.gif)

@muffinresearch : the "Browse by categories" section was removed?
Thanks!

ValentinaPC:
Observed that my question has a fix in v2. Marking this verified.

Issue #1285, opened 2016-10-25
Implement the themes home page


muffinresearch:
See the sketch file in the assets folder for more details:

![amo_-_v1_sketch](https://cloud.githubusercontent.com/assets/1514/19685426/75a0aa94-9ab4-11e6-9513-c20d8f33347b.png)

ValentinaPC:
Verified as fixed on AMO-dev FF51(Android 6.0.1)
Postfix video:
![videotogif_2017 02 10_15 58 58](https://cloud.githubusercontent.com/assets/15685960/22829253/5b96fe96-efaa-11e6-909a-2519c3780fab.gif)

Browse by category and welcome mat will be added in v2.
Also themes layout are being fixed.
Marking this verified.

PR #1337, opened 2016-11-03
Add featured API support (fix #1335)

### LINKED ISSUE #1335: Add a featured API
Support calling the featured API to get a list of add-ons.

tofumatt:
Needed for #1286 and #1285, plus I just wanted a tiny PR and something quick to knock out that didn't touch other things they needed while #1237 is in review.

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -113,3 +113,12 @@ export function fetchProfile({ api }) {
     state: api,
   });
 }
+
+export function featured({ addonType, api }) {
+  return callApi({
+    endpoint: 'addons/featured',
+    params: { app: api.clientApp, type: addonType },
+    schema: { results: arrayOf(addon) },
+    state: api,
+  });
+}
--- a/tests/client/core/api/test_api.js
+++ b/tests/client/core/api/test_api.js
@@ -142,6 +142,49 @@ describe('api', () => {
     });
   });
 
+  describe('featured add-ons api', () => {
+    function mockResponse() {
+      return Promise.resolve({
+        ok: true,
+        json() {
+          return Promise.resolve({
+            results: [
+              { slug: 'foo' },
+              { slug: 'food' },
+              { slug: 'football' },
+            ],
+          });
+        },
+      });
+    }
+
+    it('sets the app, lang, and type query', () => {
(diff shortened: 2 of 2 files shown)


src/core/api/index.js
```diff
@@ -113,3 +113,12 @@ export function fetchProfile({ api }) {
     state: api,
   });
 }
+
+export function featured({ addonType, api }) {
+  return callApi({
+    endpoint: 'addons/featured',
+    params: { app: api.clientApp, type: addonType },
+    schema: addon,
```
mstriemer: Shouldn't this be similar to [search](https://github.com/mozilla/addons-frontend/blob/6eed6fd46d1d5cde08fd46a104d82e8696f846e0/src/core/api/index.js#L66) where it's a list of add-ons?

```
 schema: { results: arrayOf(addon) },

File: tests/client/core/api/test_api.js [hunk 2]
```diff
@@ -142,6 +142,36 @@ describe('api', () => {
...
+              { slug: 'food' },
+              { slug: 'football' },
+            ],
+          });
+        },
+      });
+    }
+
+    it('sets the app, lang, and type query', () => {
+      // FIXME: This shouldn't fail if the args are in a different order.
```
mstriemer: I think the order of `Object.keys({ clientApp, lang })` is guaranteed since ES6 so it should always work.

tests/client/core/api/test_api.js
```diff
@@ -142,6 +142,36 @@ describe('api', () => {
...
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/addons/featured/?app=android&type=theme&lang=en-US`)
+        .once()
+        .returns(mockResponse());
+      return api.featured({
+        addonType: 'theme',
+        api: { clientApp: 'android', lang: 'en-US' },
+      })
+        .then(() => mockWindow.verify());
```
mstriemer: Can you check the value returned here too? As I suggested above I don't think it's returning the list as expected.

tofumatt:
Okay, pushed an update. r?

tofumatt:
Ahaha, I didn't even notice! 😄

PR #1351, opened 2016-11-07
Scroll restoration

### LINKED ISSUE #1323: Scroll position is maintained for new pages

* Visit https://amo.addons-dev.allizom.org/en-GB/firefox/ in responsive mode.
* Make a search e.g. "foo" https://amo.addons-dev.allizom.org/en-GB/firefox/search/?q=foo
* Scroll down and click one of the last results

### What happened?

The details page is loaded at the same scroll position as the click:

e.g: 

add-ons_for_firefox

The scroll position should be the top of a new page after navigating. However, clicking back should keep the original scroll position.

muffinresearch:
Fixes #1323

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,6 +180,7 @@
     "react-onclickoutside": "5.7.1",
     "react-redux": "4.4.5",
     "react-router": "2.8.1",
+    "react-router-scroll": "0.3.3",
     "redux": "3.6.0",
     "redux-connect": "4.0.2",
     "redux-logger": "2.7.3",
--- a/src/core/client/base.js
+++ b/src/core/client/base.js
@@ -6,8 +6,9 @@ import Jed from 'jed';
 import React from 'react';
 import { render } from 'react-dom';
 import { Provider } from 'react-redux';
-import { Router, browserHistory } from 'react-router';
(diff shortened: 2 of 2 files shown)


tofumatt:
Oh, one question, where does this scroll to? I'd think it'd make more sense if we loaded new pages just under the search bar, especially on mobile. Loading with the header in view on the page load isn't as useful to me.

tofumatt:
What I mean is scrolling to `.App-content` (we could add an ID if needed) would be a lot better that just to the top of the page.

muffinresearch:
> What I mean is scrolling to .App-content (we could add an ID if needed) would be a lot better that just to the top of the page.

That would be weird though since that's not what a browser would do.

muffinresearch:
> (we could add an ID if needed)

If we want a fragment id and use that in the detail links - then that should be possible but that's out of scope for the purposes of this patch.

tofumatt:
> That would be weird though since that's not what a browser would normally do which is the behaviour that this middleware is providing.

Fair enough. It feels like the proper "mobile" experience (thinking about how mobile app/browser toolbars will get out of the way when navigating through an app, and the user can scroll bar up for the search bar). Anywho, out of scope, you're right 😄

PR #1354, opened 2016-11-08
Add "More information" to Details page (fix #1294)

### LINKED ISSUE #1294: Implement the details page more information widget
See https://docs.google.com/document/d/13g8tmoII8oCNrUXOl2HegsQst7WEsXukBnIqPUNrqb0/edit#heading=h.ofi3bxtmofqf

amo_-_v1_sketch

tofumatt:
Adds "More information" to addon detail pages.

### Before
screenshot 2016-11-08 14 13 38

### After
screenshot 2016-11-08 14 25 34

### DIFF
--- a/src/amo/components/MoreInfo.js
+++ b/src/amo/components/MoreInfo.js
@@ -1,0 +1,80 @@
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+
+import Link from 'amo/components/Link';
+import translate from 'core/i18n/translate';
+
+import './MoreInfo.scss';
+
+
+export class MoreInfoBase extends React.Component {
+  static propTypes = {
+    addon: PropTypes.object.isRequired,
+    i18n: PropTypes.object.isRequired,
+    versionDetails: PropTypes.object,
+  }
+
+  render() {
+    const { addon, i18n, versionDetails } = this.props;
+    const license = versionDetails ? versionDetails.license : null;
+
+    if (versionDetails && versionDetails.loading) {
+      return (
+        <section className="MoreInfo">
+          <h2 className="MoreInfo-header">{i18n.gettext('More information')}</h2>
+          <div className="MoreInfo-contents">Loading...</div>
+        </section>
+      );
+    }
+
+    // TODO: Use the addonType for the privacy text, so it reads
+    // "for this extension", "for this theme", etc.
+    return (
+      <section className="MoreInfo">
+        <h2 className="MoreInfo-header">{i18n.gettext('More information')}</h2>
+
+        <dl className="MoreInfo-contents">
+          {addon.homepage ? <dt>{i18n.gettext('Website')}</dt> : null}
+          {addon.homepage ? (
+            <dd>
+              <a href={addon.homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+            </dd>
+          ) : null}
+          <dt>{i18n.gettext('Version')}</dt>
+          <dd>{addon.current_version.version}</dd>
+          <dt>{i18n.gettext('Last updated')}</dt>
+          <dd>{addon.last_updated}</dd>
+          {license ? (
+            <dt ref={(ref) => { this.licenseHeader = ref; }}>
+              {i18n.gettext('License')}
+            </dt>
+          ) : null}
+          {license ? (
+            <dd>
+              <a href={license.url}
+                ref={(ref) => { this.licenseLink = ref; }}>{license.name}</a>
+            </dd>
+          ) : null}
+          {addon.has_privacy_policy ?
+            <dt>{i18n.gettext('Privacy Policy')}</dt> : null}
+          {addon.has_privacy_policy ?
+            (
+              <dd>
+                <Link to={`/addons/addon/${addon.slug}/privacy-policy/`}
+                  ref={(ref) => { this.privacyPolicyLink = ref; }}>
+                  {i18n.gettext('Read the privacy policy for this add-on')}
+                </Link>
+              </dd>
+            ) : null
+          }
+          {versionDetails && versionDetails.error ? 'VersionLoad failed' : null}
+        </dl>
+      </section>
+    );
+  }
+}
+
+export default compose(
+  translate({ withRef: true }),
+)(MoreInfoBase);
--- a/src/amo/components/MoreInfo.scss
+++ b/src/amo/components/MoreInfo.scss
@@ -1,0 +1,38 @@
+@import "~core/css/inc/mixins";
+@import "~core/css/inc/vars";
+
+.MoreInfo-header {
+  @include addonSection();
+
+  border-top-left-radius: $border-radius-default;
+  border-top-right-radius: $border-radius-default;
+  font-size: $font-size-default;
+  margin-bottom: 1px;
+  text-align: center;
+}
+
+.MoreInfo-contents {
+  @include addonSection();
+
+  border-bottom-left-radius: $border-radius-default;
+  border-bottom-right-radius: $border-radius-default;
+  font-size: $font-size-s;
+
+  dt {
+    font-weight: bold;
+    line-height: 1.2;
+    margin: 0;
+    padding: 0;
+  }
+
+  dd {
+    line-height: 1.2;
+    margin: 0 0 3px;
+    padding: 0;
+
+    a:link {
+      font-size: $font-size-s;
+      font-weight: 400;
+    }
+  }
+}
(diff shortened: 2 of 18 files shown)


tests/client/amo/components/TestMoreInfo.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  it('renders loading when loading more info', () => {
+    const rootNode = findDOMNode(render({ versionDetails: { loading: true } }));
+
+    assert.equal(rootNode.querySelector('.MoreInfo-contents').textContent,
+      'Loading...');
+  });
+
+  it('does not render a homepage if none exists', () => {
```
kumar303: I find that tests like this become brittle over time. For example, some new data could be added to `addon` that is unrelated to this test. This might break the test. You could keep it robust by taking an approach like this:
````javascript
import deepcopy from 'deepcopy';
const partialAdon = deepcopy(addon);
delete partialAddon.homepage;
// Run the test with partialAddon
````
See: https://www.npmjs.com/package/deepcopy

File: tests/client/amo/components/TestMoreInfo.js [hunk 2]
```diff
@@ -0,0 +1,82 @@
...
+
+  it('renders the privacy policy and link', () => {
+    const root = render();
+
+    assert.equal(findDOMNode(root.privacyPolicyLink).textContent,
+      'Read the privacy policy for this add-on');
+    assert.equal(root.privacyPolicyLink.props.to,
+      '/addons/addon/my-addon/privacy-policy/');
+  });
+});
```
kumar303: I don't see a test for rendering the version number or the last updated date. Did I miss them?

src/amo/store.js
```diff
@@ -7,10 +7,13 @@ import api from 'core/reducers/api';
 import auth from 'core/reducers/authentication';
 import reviews from 'amo/reducers/reviews';
 import search from 'core/reducers/search';
+import version from 'core/reducers/version';
 
 export default function createStore(initialState = {}) {
   return _createStore(
-    combineReducers({ addons, api, auth, search, reviews, reduxAsyncConnect }),
+    combineReducers({
+      addons, api, auth, search, reviews, reduxAsyncConnect, version,
```
kumar303: I think it's confusing to call this `version`. It would make more sense to me as `versions` since it's a map of version objects by ID.

src/core/reducers/version.js
```diff
@@ -0,0 +1,26 @@
...
+  loading: false,
+};
+
+export default function version(state = initialState, action) {
+  const { payload } = action;
+  switch (action.type) {
+    case VERSION_GET:
+      return { ...state, ...payload, loading: true };
+    case VERSION_LOADED:
+      return { ...state, ...payload.result, loading: false };
```
kumar303: It would be helpful to see a comment here explaining that `payload.result` will expand to a map of version objects by ID

tests/client/core/api/test_api.js
```diff
@@ -142,6 +142,53 @@ describe('api', () => {
     });
   });
 
+  describe('versionDetail api', () => {
+    function mockResponse() {
```
kumar303: This doesn't look like it needs to be a function. You could just reference the resolved promise object.

tofumatt:
After chatting on IRC this patch could be vastly simplified if https://github.com/mozilla/addons-server/issues/3964 was fixed, so I think this can be put aside until we see about that issue.

tests/client/core/api/test_api.js
kumar303: Example:
````javascript
const response = Promise.resolve(...);
mockWindow.expects('fetch')
  .withArgs(`${apiHost}/api/v3/addons/addon/my-addon/versions/3/?lang=fr`)
  .once()
  .returns(response);
````

File: src/core/reducers/version.js [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+
+const initialState = {
+  loading: false,
+};
+
+export default function version(state = initialState, action) {
+  const { payload } = action;
+  switch (action.type) {
+    case VERSION_GET:
+      return { ...state, ...payload, loading: true };
```
kumar303: I think this needs `error: false` in case the last state was an error state

src/core/reducers/version.js
kumar303: same thing: I think this needs `error: false` in case the last state was an error state

src/core/utils.js
```diff
@@ -105,6 +106,56 @@ export function loadAddonIfNeeded(
...
+  }
+
+  // Check for this add-on's current version details in the existing state.
+  if (state.version && state.version[addon.current_version.id]) {
+    dispatch(versionGet({ slug, versionID: addon.current_version.id }));
+    log.info(`Found version ${addon.current_version.id} for addon ${addon.id} in state`);
+    return state.version[addon.current_version.id];
+  }
+
+  return fetchVersionDetails({
```
kumar303: Yeah, I think https://github.com/mozilla/addons-server/issues/3964 will make all this code way simpler

src/amo/components/MoreInfo.js
```diff
@@ -0,0 +1,80 @@
...
+
+        <dl className="MoreInfo-contents">
+          {addon.homepage ? <dt>{i18n.gettext('Website')}</dt> : null}
+          {addon.homepage ? (
+            <dd>
+              <a href={addon.homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+            </dd>
+          ) : null}
+          <dt>{i18n.gettext('Version')}</dt>
```
tofumatt: Yeah, I thought "Version" would do the trick. Especially in the case of a version "number" that maybe had an `-rc` or something in it.

src/amo/components/MoreInfo.js
```diff
@@ -0,0 +1,80 @@
...
+          <dd>{addon.current_version.version}</dd>
+          <dt>{i18n.gettext('Last updated')}</dt>
+          <dd>{addon.last_updated}</dd>
```
tofumatt: Indeed! I put this in here and forgot to format it. Thanks for pointing it out. If it's easy I'll do it here; if not I'll file an issue.

src/amo/components/MoreInfo.js
```diff
@@ -0,0 +1,80 @@
...
+            (
+              <dd>
+                <Link to={`/addons/addon/${addon.slug}/privacy-policy/`}
+                  ref={(ref) => { this.privacyPolicyLink = ref; }}>
+                  {i18n.gettext('Read the privacy policy for this add-on')}
+                </Link>
+              </dd>
+            ) : null
+          }
+          {versionDetails && versionDetails.error ? 'VersionLoad failed' : null}
```
tofumatt: Hahaha yes. Sorry.

src/amo/components/MoreInfo.scss
```diff
@@ -0,0 +1,38 @@
...
+.MoreInfo-contents {
+  @include addonSection();
+
+  border-bottom-left-radius: $border-radius-default;
+  border-bottom-right-radius: $border-radius-default;
+  font-size: $font-size-s;
+
+  dt {
+    font-weight: bold;
+    line-height: 1.2;
```
tofumatt: Actually, not a SASS trick! https://developer.mozilla.org/en/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values

src/amo/containers/DetailPage.js
```diff
@@ -19,17 +20,20 @@ export class DetailPageBase extends React.Component {
 
 function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
+  const addon = state.addons[slug];
   return {
-    addon: state.addons[slug],
+    addon,
     slug,
+    versionDetails: addon && state.version[addon.current_version.id] ?
+      state.version[addon.current_version.id] : state.version,
```
tofumatt: I think it's because of how my reducers are returning values which isn't great. Agreed, this should be fixed.

src/amo/components/MoreInfo.js
Ryuno-Ki: Well, Angular has a [date filter](https://docs.angularjs.org/api/ng/filter/date). I would be surprised if React wouldn't provide one O:

I assume, it uses moment.js internally and derives the preferred language from somewhere else.

src/core/actions/version.js
```diff
@@ -0,0 +1,29 @@
...
+    type: VERSION_GET,
+    payload: { slug, versionID },
+  };
+}
+
+export function versionLoad({ entities }) {
+  return {
+    type: VERSION_LOADED,
+    payload: {
+      result: entities ? entities.versions : null,
```
Ryuno-Ki: Why should `entities` be null? The ternary operator checks for a *falsy* value. The `payload.result` would be `null`.

Issue #1291, opened 2016-10-25
Hookup the install button


muffinresearch:
This will be a bit complicated because of the limitations on origin for the install apis. 

It also needs to take into account the fallback see https://github.com/mozilla/addons-frontend/issues/946

![add-ons_for_firefox](https://cloud.githubusercontent.com/assets/1514/19685896/a2b382c0-9ab6-11e6-9a1d-e0f7e83b84c3.png)

mstriemer:
QA:

This includes updates to how the install button is handled on the disco pane to make it work on the new mobile pages. Can you please test that the install buttons are still working as expected with these changes.

ValentinaPC:
Verified all buttons from disco pane using prefs for -dev (+`extensions.webapi.testing` on true) and -prod on FF 53.0a1 (2016-11-17) (32-bit) and work as before.
Let me know if there is anything else to be checked.

PR #1358, opened 2016-11-10
Extract Switch into a really-dumb component

mstriemer:
This is going to break the build, I will fix it in the morning.

switch-fixed mov

> screenshot 2016-11-09 22 24 59
> Ignore the "success" one, I fixed it like in the gif. The progress=null ones are really `Infinity` and `-Infinity`.

Supports #1291.


src/ui/components/Switch/Switch.scss
```diff
@@ -246,7 +238,7 @@ $installStripeColor2: #00c42e;
 }
 
 @for $i from 1 through 100 {
```
kumar303: ooh fancy

src/ui/components/Switch/index.js
```diff
@@ -0,0 +1,69 @@
...
+    return (
+      <div className={classes} onClick={onClick}
+        data-progress={hasProgress ? this.getProgress() : 0}>
+        <input
+          id={identifier}
+          className="visually-hidden"
+          checked={checked}
+          disabled={disabled}
+          onChange={onChange}
+          ref={(ref) => { this.themeData = ref; }}
```
kumar303: The ref `this.themeData` looks like it's still accessed by `InstallButton` in `handleClick`. 

More broadly, the approach of using data attributes to know which theme to install seems very un-react to me. Since everything is a react component with its own data already, you shouldn't need data attributes. This makes me think that the install action isn't attached to the right component.

mstriemer: The `data-browsertheme` attribute is required unfortunately. Previewing a theme works by dispatching an event on an element that has the `data-browsertheme` attribute with a JSON blob in it. No, I'm not kidding.

I'm not sure what the requirements are for this, perhaps we can just do `dispatchEvent(document.createElement(...))` or maybe that element needs to have an event happening on it already (this is how it's currently done).

I'll take a look at the ref.

mstriemer: We were switching themes while the video played before so we should be able to dispatch on anything. I think you're right the theme stuff should come out of `Switch`. I'll try moving it.

tofumatt: > No, I'm not kidding.

I lol'd.

mstriemer: Turns out the element needs to be in the document (boo!) so I added a wrapper to store the data 😞.

mstriemer:
Removed the theme stuff from `Switch` and pushed an extra commit to show the check mark on `ENABLED` too, I thought we just wanted it on the initial install but we currently always show it.

kumar303:
ok, I'll take another look

src/core/components/InstallButton/index.js
```diff
@@ -71,6 +70,18 @@ export class InstallButtonBase extends React.Component {
     return i18n.sprintf(label, { name });
   }
 
+  getDownloadProgress() {
```
kumar303: I have a couple thoughts on this:

1. This function / logic is only used by `Switch`. Why not move it there?

2. I think it's a bad pattern to pass multiple properties through a component simply because a child component needs them, i.e. passing `status` and `downloadProgress` from `Addon` to `InstallButton` to `Switch`. Using redux to fetch download progress from the state only where it's needed (i.e. `Switch`) would solve this problem nicely. The way I would do it would be to create a `downloadTransactionId` in the `Addon` parent when the user initiates installation then pass that property all the way down: `InstallButton` > `Switch`. The `Addon` could continue dispatching download progress data but it would need to add a transaction ID. `Switch` could fetch download progress directly from the state based on `downloadTransactionId`. This model would still keep `Switch` relatively dumb because it wouldn't know anything about add-ons or versions, only about the current download transaction. The key difference in this model is that the entire chain of components is only re-rendered *once* when the `downloadTransactionId` changes, not all the time as `downloadProgress` gets updated repeatedly.

My first suggestion is probably worth doing (if you agree). The second suggestion is a much bigger change and probably not worth spending time on today.

src/ui/components/Switch/Switch.scss
```diff
@@ -110,101 +110,101 @@ $installStripeColor2: #00c42e;
...
+  &.Switch--success {
+    input + label::before {
+      background: $switchBackgroundOn url('./tick.svg') no-repeat 35% 50%;
 
-      &::before {
-        @include stripes(
-          $installStripeColor1,
-          $installStripeColor2,
-          $switchBackgroundOn);
+      [dir=rtl] & {
```
kumar303: I didn't know about this kind of parent/child inversion with the ampersand, seems handy

tests/client/ui/components/TestSwitch.js
```diff
@@ -0,0 +1,108 @@
...
+  function renderButton(props = {}) {
+    const renderProps = {
+      dispatch: sinon.spy(),
+      label: 'Hey',
+      name: 'sup',
+      ...props,
+    };
+
+    return renderIntoDocument(
+      <Switch {...renderProps} />);
```
kumar303: could be a one liner

src/core/components/InstallButton/index.js
mstriemer: 1. `InstallButton` is a wrapper around `Switch` that provides the hooks for installing an add-on or theme and tells the `Switch` what state to be in. This logic is how an add-on install should be displayed by a switch. Likewise if we had a switch that saved a user setting it might only do the `Infinity` and `-Infinity` case if it saves on toggle.
2. I think fetching data from the redux store at the lowest level can only limit our options in how something is used. By creating components that don't interact with redux at the lowest level we should be able to wrap them in very basic containers that just provide some logic like this one. This way the code for creating this pretty switch that we want to use in multiple places isn't tied to downloads.

> The key difference in this model is that the entire chain of components is only re-rendered once when the downloadTransactionId changes, not all the time as downloadProgress gets updated repeatedly.

Do you mean this prevents the `Addon` component from re-rendering? The `Switch` will always need to re-render so there's no benefit there. This sounds like an optimisation on something that's already optimised (React virtual DOM diffing).

mstriemer: 1. `InstallButton` is a wrapper around `Switch` that provides the hooks for installing an add-on or theme and tells the `Switch` what state to be in. This logic is how an add-on install should be displayed by a switch. Likewise if we had a switch that saved a user setting it might only do the `Infinity` and `-Infinity` case if it saves on toggle.
2. I think fetching data from the redux store at the lowest level can only limit our options in how something is used. By creating components that don't interact with redux at the lowest level we should be able to wrap them in very basic containers that just provide some logic like this one. That way the code for creating this pretty switch that we might use in multiple places isn't tied to downloads.

> The key difference in this model is that the entire chain of components is only re-rendered once when the downloadTransactionId changes, not all the time as downloadProgress gets updated repeatedly.

Do you mean this prevents the `Addon` component from re-rendering? The `Switch` will always need to re-render so there's no benefit there. This sounds like an optimisation on something that's already optimised (React virtual DOM diffing).

mstriemer:
Going to merge this. If you think I should still make those changes then let's talk about it more here or in a new issue.

src/core/components/InstallButton/index.js
kumar303: I don't think it's necessary right now to make this architectural change but I was just thinking that we could use redux in a way that still makes the component *dumb*. Instead of accessing redux data with `downloadProgressId` it could be more generic, like `switchId`. Thus, it wouldn't need to be download specific. 

By using redux, I think it allows us to separate the *producers* of data from the *consumers* of data better. The `Addon` component is the one that dispatches download progress data but `Switch` is really the only one that consumes it so I don't see a good reason for `Addon` to consume the data only to pass it down through its children until it reaches `Switch`.

Issue #1316, opened 2016-10-28
Create a Provider test helper for components that use AMO's internal `<Link>`


tofumatt:
See: https://github.com/mozilla/addons-frontend/pull/1309/files#r85552951

kumar303:
It would be helpful to add a provider helper (for i18n and redux) to all tests by default

kumar303:
It's possible that moving to a universal link component will fix this: https://github.com/mozilla/addons-frontend/issues/2053

willdurand:
I don't think this is valid anymore. We do not use `Provider` and inject props directly on the components, often using `shallowUntilTarget()` and a base component.

PR #1365, opened 2016-11-10
Add Category Page (close #1183)

### LINKED ISSUE #1183: Implement a categories component/page
Implement the design in #803 and create a page to view the categories.

tofumatt:
Okay, this *only* implements the category listing page and not the addons inside a result. That'll come next as a separate PR so this is easier to review.

This is largely just separated out of the work in #1237 with issues addressed.

screen shot 2016-11-10 at 18 06 13

### DIFF
--- a/src/amo/containers/CategoryList.js
+++ b/src/amo/containers/CategoryList.js
@@ -1,0 +1,30 @@
+import { compose } from 'redux';
+import { asyncConnect } from 'redux-connect';
+import { connect } from 'react-redux';
+
+import Categories from 'amo/components/Categories';
+import { loadCategoriesIfNeeded } from 'core/utils';
+
+
+export function mapStateToProps(state, ownProps) {
+  const addonType = ownProps.params.addonType.replace(/s$/, '');
+  const clientApp = state.api.clientApp;
+  const categories = state.categories.categories[clientApp][addonType] ?
+    state.categories.categories[clientApp][addonType] : {};
+
+  return {
+    addonType,
+    categories,
+    error: state.categories.error,
+    loading: state.categories.loading,
+  };
+}
+
+export default compose(
+  asyncConnect([{
+    deferred: true,
+    key: 'CategoriesPage',
+    promise: loadCategoriesIfNeeded,
+  }]),
+  connect(mapStateToProps),
+)(Categories);
--- a/src/core/reducers/categories.js
+++ b/src/core/reducers/categories.js
@@ -1,0 +1,65 @@
+import config from 'config';
+
+import {
+  CATEGORIES_GET,
+  CATEGORIES_LOAD,
+  CATEGORIES_FAILED,
+} from 'core/constants';
+
+
+export function emptyCategoryList() {
+  return config.get('validClientApplications')
+    .reduce((object, appName) => ({ ...object, [appName]: {} }), {});
+}
+
+const initialState = {
+  categories: emptyCategoryList(),
+  error: false,
+  loading: false,
+};
+
+export default function categories(state = initialState, action) {
+  const { payload } = action;
+  let categoryList;
+
+  switch (action.type) {
+    case CATEGORIES_GET:
+      return { ...state, ...payload, loading: true };
+    case CATEGORIES_LOAD:
+      categoryList = emptyCategoryList();
+      payload.result.forEach((result) => {
+        // If the API returns data for an application we don't support,
+        // we'll ignore it for now.
+        if (!categoryList[result.application]) {
+          return;
+        }
+
+        if (!categoryList[result.application][result.type]) {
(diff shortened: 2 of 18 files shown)


tests/client/amo/components/TestCategeories.js
```diff
@@ -0,0 +1,101 @@
...
+      clientApp: 'android',
+      categories,
+    };
+    const initialState = {
+      api: { clientApp: 'android', lang: 'fr' },
+      categories: { ...categories },
+    };
+
+    return findDOMNode(findRenderedComponentWithType(renderIntoDocument(
+      <Provider store={createStore(initialState)}>
```
mstriemer: Do you need the provider here? I feel it shouldn't be necessary.

tests/client/amo/components/TestCategeories.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  it('renders Categories', () => {
+    const root = render({
+      addonType: 'extension',
+      error: false,
+      loading: false,
+    });
+
+    assert.ok(root.querySelector('.Categories-list'));
```
mstriemer: Might be nice to check `assert.equal(root.textContent, 'Games\nTravel');` or something similar.

tests/client/amo/containers/TestCategoryList.js
```diff
@@ -0,0 +1,24 @@
+import { mapStateToProps } from 'amo/containers/CategoryList';
+
+
+describe('<CategoryList />', () => {
+  it('maps state to props', () => {
+    const props = mapStateToProps({
+      api: { clientApp: 'android', lang: 'pt' },
+      categories: {
+        categories: { android: {}, firefox: {} },
```
mstriemer: Should probably have some data in here so we know it's picking the right one.

tests/client/core/reducers/test_categories.js
```diff
@@ -0,0 +1,189 @@
...
+        type: 'extension',
+      },
+      {
+        application: 'netscape',
+        name: 'I should not appear',
+        slug: 'i-should-not-appear',
+        type: 'extension',
+      },
+    ];
+    const state = categories(initialState, {
```
mstriemer: can you move this to a `before()` function?

src/core/reducers/categories.js
```diff
@@ -0,0 +1,71 @@
...
+
+        categoryList[result.application][result.type].push(result);
+      });
+
+      Object.keys(categoryList).forEach((appName) => {
+        Object.keys(categoryList[appName]).forEach((addonType) => {
+          categoryList[appName][addonType] = categoryList[appName][addonType]
+            .sort((a, b) => a.name > b.name)
+            .reduce((object, value) => {
+              // eslint-disable-next-line no-param-reassign
```
tofumatt: Ah, true, good call.

src/amo/containers/CategoryList.js
```diff
@@ -0,0 +1,30 @@
...
+import { connect } from 'react-redux';
+
+import Categories from 'amo/components/Categories';
+import { loadCategoriesIfNeeded } from 'core/utils';
+
+
+export function mapStateToProps(state, ownProps) {
+  const addonType = ownProps.params.addonType.replace(/s$/, '');
+  const clientApp = state.api.clientApp;
+  const categories = state.categories.categories[clientApp][addonType] ?
```
tofumatt: I definitely like that idea, but think it'd be something we should do later. Thanks for adding the comment though, the proposed idea is much nicer.

tests/client/amo/components/TestCategeories.js
tofumatt: It's because we're using our internal `<Link>` component in that component. 😞

Would be removed once #1316 is fixed.

src/amo/containers/CategoryList.js
```diff
@@ -0,0 +1,30 @@
+import { compose } from 'redux';
+import { asyncConnect } from 'redux-connect';
+import { connect } from 'react-redux';
+
+import Categories from 'amo/components/Categories';
+import { loadCategoriesIfNeeded } from 'core/utils';
+
+
+export function mapStateToProps(state, ownProps) {
+  const addonType = ownProps.params.addonType.replace(/s$/, '');
```
tofumatt: That's actually a fair idea. I'm gonna file an issue because I think it requires an additional error message/test and I'd rather split it out, but 👍

Ryuno-Ki: Is `addonType` translatable? Then the "always ends in an s" trick wouldn't work in other languages than English so the `externalAddonTypeMap` would be needed anyway.

tofumatt: No, it's not translated as it's in the URL, which isn't localised.

Issue #1366, opened 2016-11-10
Adjust the review text prompt based on star selection


kumar303:
### Describe the problem and steps to reproduce it:

Go to an add-on detail page, select a one or two star rating.

### What happened?

Your review is saved (if there isn't an error) and on the next screen the review prompt asks you "Tell the world why you think this extension is fantastic!" The placeholder for the review starts with "Tell us what you love about this extension."

### What did you expect to happen?

It should say something more relevant to the feelings of the reviewer who is not happy with the add-on.

vcarciu:
Verified as fixed : 
![mobilefantastic](https://cloud.githubusercontent.com/assets/1583842/22654014/145dac14-ec94-11e6-9837-e9f5be14e0ab.png)

Notes:
@kumar303  In description , we asked for another text for 1 or 2 stars rating, but I observed that same text is added for a 3 star rating. I think that this is reasonable, but I want to be sure , so please confirm

PR #1368, opened 2016-11-10
Use an appropriate prompt based on the star rating

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1366

### DIFF
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -83,25 +83,38 @@ export class AddonReviewBase extends React.Component {
       );
     }
 
+    let placeholder;
+    let prompt;
+    if (review.rating && (review.rating > 3)) {
+      prompt = i18n.gettext(
+        'Tell the world why you think this extension is fantastic!'
+      );
+      placeholder = i18n.gettext(
+        'Tell us what you love about this extension. Be specific and concise.'
+      );
+    } else {
+      prompt = i18n.gettext(
+        'Tell the world about this extension.'
+      );
+      placeholder = i18n.gettext(
+        'Tell us about your experience with this extension. ' +
+        'Be specific and concise.'
+      );
+    }
+
     // TODO: I guess we should load the existing review text so it
     // can be edited? That flow needs more thought.
     return (
       <div className="AddonReview">
         <h2 className="AddonReview-header">{i18n.gettext('Write a review')}</h2>
-        <p>
-          {i18n.gettext(
-            'Tell the world why you think this extension is fantastic!'
-          )}
-        </p>
+        <p ref={(ref) => { this.reviewPrompt = ref; }}>{prompt}</p>
         {errorMessage}
         <form onSubmit={this.onSubmit} ref={(ref) => { this.reviewForm = ref; }}>
           <textarea
             className="AddonReview-textarea"
             ref={(ref) => { this.reviewTextarea = ref; }}
             name="review"
-            placeholder={i18n.gettext(
-              'Tell us what you love about this extension. ' +
-              'Be specific and concise.')} />
+            placeholder={placeholder} />
           <div className="AddonReview-button-row">
             <button className="AddonReview-button AddonReview-back-button"
               onClick={this.goBackToAddonDetail}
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonReview.js
```diff
@@ -83,25 +83,38 @@ export class AddonReviewBase extends React.Component {
       );
     }
 
+    let placeholder;
+    let prompt;
+    if (review.rating && (review.rating > 3)) {
```
tofumatt: The braces here make me expect something more and could be removed.

src/amo/components/AddonReview.js
```diff
@@ -83,25 +83,38 @@ export class AddonReviewBase extends React.Component {
...
+      );
+      placeholder = i18n.gettext(
+        'Tell us what you love about this extension. Be specific and concise.'
+      );
+    } else {
+      prompt = i18n.gettext(
+        'Tell the world about this extension.'
+      );
+      placeholder = i18n.gettext(
+        'Tell us about your experience with this extension. ' +
```
tofumatt: We should see about using the linter's `singleLineString` here (https://muffinresearch.co.uk/removing-leading-whitespace-in-es6-template-strings/)

src/amo/components/AddonReview.js
kumar303: what do you mean by braces? Did you mean parenthesis?

src/amo/components/AddonReview.js
tofumatt: It's a template string, so I think it does? We wrapped calls in the linter with a dummy gettext but never did extractions, so I'm unsure.

src/amo/components/AddonReview.js
tofumatt: Yeah, sorry. My brain got jumbled I meant:

```js
if (review.rating && review.rating > 3) { 
```
would work and I think is easier to read. I prefer the parens for more complicated statements inside `if`s.

kumar303: Alright. I was nervous about the `&&` being applied first but `5 && 5 = 5` so I guess that's ok!

src/amo/components/AddonReview.js
kumar303: it looks like `dedent` will work: https://github.com/mozilla/addons-frontend/blob/master/src/disco/containers/DiscoPane.js#L99

tofumatt: Aha, it seems they beat us to the "publish to npm bit", yeah that's totally what we were doing in the linter, we just rolled our own (oops).

👍

kumar303: I don't think it's really possible here. Even with `dedent`, you can't create a space at the break which is necessary for the placeholder to read correctly in the textarea.

tofumatt:
Oh, that’s true.


Okay, anything, was just a nitpick, it’s cool as is!


On 10 November 2016 at 22:55:30, Kumar McMillan (notifications@github.com)
wrote:

*@kumar303* commented on this pull request.
------------------------------

In src/amo/components/AddonReview.js
<https://github.com/mozilla/addons-frontend/pull/1368>:


> @@ -83,25 +83,38 @@ export class AddonReviewBase extends React.Component {
       );
     }

+    let placeholder;
+    let prompt;
+    if (review.rating && (review.rating > 3)) {
+      prompt = i18n.gettext(
+        'Tell the world why you think this extension is fantastic!'
+      );
+      placeholder = i18n.gettext(
+        'Tell us what you love about this extension. Be specific and concise.'
+      );
+    } else {
+      prompt = i18n.gettext(
+        'Tell the world about this extension.'
+      );
+      placeholder = i18n.gettext(
+        'Tell us about your experience with this extension. ' +

I don't think it's really possible here. Even with dedent, you can't create
a space at the break which is necessary for the placeholder to read
correctly in the textarea.

—
You are receiving this because you were assigned.
<https://github.com/mozilla/addons-frontend/pull/1368>, or mute the thread

kumar303:
Thanks for the late night review 🍻

PR #1370, opened 2016-11-11
Update Polish (pl) localization of AMO Frontend

piotrdrag:
CC @splewako

### DIFF
--- a/locale/pl/LC_MESSAGES/amo.po
+++ b/locale/pl/LC_MESSAGES/amo.po
@@ -3,242 +3,243 @@ msgstr ""
 "Project-Id-Version: amo\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2016-11-10 09:55+0000\n"
-"PO-Revision-Date: 2016-11-10 09:55+0000\n"
-"Last-Translator: Automatically generated\n"
-"Language-Team: none\n"
+"PO-Revision-Date: 2016-11-11 15:01+0100\n"
+"Last-Translator: Piotr Drąg <piotrdrag@gmail.com>\n"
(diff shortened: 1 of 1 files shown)


locale/pl/LC_MESSAGES/amo.po
```diff
@@ -3,242 +3,243 @@ msgstr ""
...
-msgstr ""
+msgstr "Kategoria"
 
 #: src/amo/components/AddonMeta.js:24
 msgid "Used by"
-msgstr ""
+msgstr "Używane przez"
 
 #: src/amo/components/AddonMeta.js:29
 msgid "Sentiment"
```
piotrdrag: I have no idea what it means in this context.

tofumatt:
I think these are supposed to be added [via pontoon](https://pontoon.mozilla.org/), where we will then pull from. Sorry for the confusion.

cc @muffinresearch to make sure I'm not wrong about that.

piotrdrag:
We, the Polish localization team, do not use web translation services.

locale/pl/LC_MESSAGES/amo.po
tofumatt: It's part of some placeholder HTML so it might changed soon, but good point.

I think it means "User sentiment about the add-on", as it's followed by "89% love it". But that text isn't even wrapped in gettext and this is clearly placeholder, so I think either way it can be left alone for now.

tofumatt:
Ah, okay. I’m not very familiar with the localisation process; my apologies
:-)

caitmuenster:
Thanks so much for the fix, Piotr! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#November_2016). Happy hacking!

PR #1372, opened 2016-11-14
greenkeeper helmet 3.1.0

muffinresearch:
* Supercedes #1336

Once this lands need to do quick manual check to diff the header output on -dev.

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
     "dompurify": "0.8.4",
     "express": "4.14.0",
     "extract-text-webpack-plugin": "1.0.1",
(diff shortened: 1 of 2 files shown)


muffinresearch:
I think we might want to turn on includeSubdomains but we don't currently set it for AMO either. I don't think there's a reason not to send that but we should consider that separately just so we make sure there aren't domains that would be impacted that would cause a problem. 

The HSTS settings here provide the same output as we have currently. I've filed https://github.com/mozilla/addons/issues/284 to deal with considering turning that on for both addons-frontend and addons-server.

Issue #1088, opened 2016-09-19
Indicate that a user has already rated an add-on


kumar303:
In https://github.com/mozilla/addons-frontend/issues/1017 the UI was implemented to present the smileys that a user can select to give their overall rating of an add-on. If the user returns to this page after rating the add-on, they cannot re-rate the add-on. How do we visually indicate that they have already taken this action?

UX flow tracker: #927
This supports mozilla/addons#151

kumar303:
@pwalm let me know your thoughts on this. I see that [an older revision](https://github.com/mozilla/addons-frontend/issues/800#issuecomment-238436613) had some grey smileys. Maybe we can use an approach like that?

kumar303:
Actually, now that we have star ratings (https://github.com/mozilla/addons-frontend/issues/1288) the UX for this will be super easy. It will just look like a selected star range.

pwalm:
Need anything else on this issue from me, @kumar303?

kumar303:
Nope, we're good. I realized that the selected star view is in the Sketch mock.

ValentinaPC:
Verified as fixed on FF51(Android 6.0.1)
After an add-on is rated, second time the user goes to add-on details page, a selected star range can be seen.
Postfix screenshot: 
![screenshot_20170206-160718](https://cloud.githubusercontent.com/assets/15685960/22650237/eda98406-ec86-11e6-830c-7cd7c0d3f865.png)

PR #1374, opened 2016-11-14
Render saved star ratings on detail page

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1088

* Saved stars are rendered as "selected"
* The rating button is disabled since you can't rate an add-on twice

### DIFF
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -29,10 +29,15 @@ export function submitReview({
     });
 }
 
-export function getUserReviews({ userId }) {
-  // Warning: if you rely on this API as is, you'll have to implement
-  // your own paging.
-  return callApi({
-    endpoint: `accounts/account/${userId}/reviews`,
+export function getUserReviews({ userId } = {}) {
+  return new Promise((resolve) => {
+    if (!userId) {
+      throw new Error('userId cannot be falsey');
+    }
+    // Warning: if you rely on this API as is, you'll have to implement
+    // your own paging.
+    resolve(callApi({
+      endpoint: `accounts/account/${userId}/reviews`,
+    }));
   });
 }
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -1,4 +1,5 @@
 import { callApi } from 'core/api';
+import log from 'core/logger';
 
 /*
  * POST/PATCH an add-on review using the API.
@@ -33,15 +34,40 @@ export function submitReview({
     });
 }
 
-export function getUserReviews({ userId } = {}) {
+export function getUserReviews({
+  userId, addonId, onlyTheLatest = false,
+} = {}) {
   return new Promise((resolve) => {
     if (!userId) {
       throw new Error('userId cannot be falsey');
     }
-    // Warning: if you rely on this API as is, you'll have to implement
-    // your own paging.
     resolve(callApi({
       endpoint: `accounts/account/${userId}/reviews`,
     }));
-  });
+  })
+    .then((response) => {
+      // TODO: implement paging through response.next
+      if (response.next) {
+        throw new Error('paging is not yet implemented');
+      }
+      return response.results;
+    })
+    .then((reviews) => {
+      if (addonId) {
+        log.info(`Filtering user ${userId} reviews by addonId ${addonId}`);
+        return reviews.filter((review) => review.addon.id === addonId);
+      }
+      return reviews;
+    })
+    .then((reviews) => {
+      if (onlyTheLatest) {
+        log.info(`Only fetching the latest review by user ${userId}`);
+        const latest = reviews.filter((review) => review.is_latest);
+        if (latest.length === 0) {
+          return null;
+        }
+        return latest[0];
+      }
+      return reviews;
+    });
 }
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -28,3 +28,11 @@ export function submitReview({
       }));
     });
 }
+
+export function getUserReviews({ userId }) {
+  // Warning: if you rely on this API as is, you'll have to implement
+  // your own paging.
+  return callApi({
+    endpoint: `accounts/account/${userId}/reviews`,
+  });
+}
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -7,6 +7,10 @@ export function submitReview({
   rating, apiState, addonSlug, versionId, body, reviewId,
 }) {
   const data = { rating, version: versionId, body };
+  if (reviewId) {
+    // You cannot update the version of an existing review.
+    data.version = undefined;
+  }
 
   return new Promise(
     (resolve) => {
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -34,9 +34,7 @@ export function submitReview({
     });
 }
 
-export function getUserReviews({
-  userId, addonId, onlyTheLatest = false,
-} = {}) {
+export function getUserReviews({ userId, addonId } = {}) {
   return new Promise((resolve) => {
     if (!userId) {
       throw new Error('userId cannot be falsey');
@@ -58,16 +56,16 @@ export function getUserReviews({
         return reviews.filter((review) => review.addon.id === addonId);
       }
       return reviews;
-    })
+    });
+}
+
+export function getLatestUserReview(params) {
+  return getUserReviews(params)
     .then((reviews) => {
-      if (onlyTheLatest) {
-        log.info(`Only fetching the latest review by user ${userId}`);
-        const latest = reviews.filter((review) => review.is_latest);
-        if (latest.length === 0) {
-          return null;
-        }
-        return latest[0];
+      const latest = reviews.filter((review) => review.is_latest);
+      if (latest.length === 0) {
+        return null;
       }
-      return reviews;
+      return latest[0];
     });
 }
--- a/src/amo/components/OverallRating.js
+++ b/src/amo/components/OverallRating.js
@@ -158,6 +158,8 @@ export const mapDispatchToProps = (dispatch) => ({
           throw new Error(
             `Unexpectedly received more than one review for
             user ${userId}, addonId ${addonId}, versionId ${versionId}`);
+        } else {
+          log.info('No reviews found for', { addonId, versionId, userId });
         }
       });
   },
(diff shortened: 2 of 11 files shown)


tests/client/amo/components/TestOverallRating.js
```diff
@@ -182,31 +253,254 @@ describe('OverallRating', () => {
...
+      store.dispatch(setJWT(userAuthToken({
+        user_id: userId,
+      })));
+    }
+
+    it('copies api state to props', () => {
+      signIn();
+
+      const state = store.getState();
+      assert(state.api.token, 'a valid token exists in state');
```
mstriemer: It might make sense to move this to `signIn()`.

src/amo/components/OverallRating.js
```diff
@@ -70,12 +104,70 @@ export class OverallRatingBase extends React.Component {
...
+    userId,
+  };
+};
 
 export const mapDispatchToProps = (dispatch) => ({
+
+  loadSavedRating({ addonId, versionId, userId }) {
+    return getUserReviews({ userId })
+      .then((response) => {
+        const relevantReviews = response.results.filter((review) => {
```
kumar303: Fetching all user reviews for all add-ons ever is a hack and I'd like it to replace it with a query that gets exactly the review we need. Thus, I don't think it makes sense to put them all in state before the new API filter lands (they are needed in state).

mstriemer: Might be nice to move the filter into api and call it `getUserReview({ userId, addonId })` then. Paper over it until we can optimise it.

src/amo/css/OverallRating.scss
```diff
@@ -6,22 +6,28 @@
...
-.OverallRating-choice {
+.OverallRating-choices button {
   background: url('../img/ratings/big-star.svg') center/$star-size no-repeat;
   cursor: pointer;
   height: $star-size + 10px;
   padding-top: 80px;
   width: $star-size + 10px;
 }
 
+.OverallRating-choices button:disabled {
```
kumar303: The dilemma with always setting this class is that there are two types of buttons: 1) ones that you can click on and 2) ones that are informative but that you can't click on. This seems like the easiest way to support both cases. Each case is pretty different visually, mostly in all the hover logic.

tests/client/amo/components/TestOverallRating.js
kumar303: I don't think this assertion needs to happen for every test. I only added it here because the assertion below it has the possibly to silently fail if `state.api` is falsey. If we introduce static Flow typing then that wouldn't be an issue!

src/amo/css/OverallRating.scss
mstriemer: Couldn't you accomplish that with `:enabled` and `:disabled`?

kumar303:
> I don't think this should only be showing a review for the current version. I feel like users would find that very confusing.

The visual feedback in this patch is to tell the user: "you already rated this add-on + version at X stars." If they write a review for version 1, the add-on gets updated to version 2, and they return to the details page, they will not see their saved review. I think that is the expected behavior because this allows them to write a review of the new version. Otherwise, how could they rate the newer version? It would need a lot of UX thought to get that right.

> I haven't taken a look but do you know if we save the review against the current version or the version the user has installed? I would guess the current version.

My patch saves the review against whatever version they are viewing (which is most likely the current one). It doesn't take into account their installed version. It's also possible for a user to review an add-on they have never installed. 

Actually, what happened [in the old site](https://github.com/mozilla/addons-server/blob/master/src/olympia/reviews/views.py#L223) is that a review was *always* submitted for the current version. Instead of that, I made it submit a review for whatever version they are viewing, which is most likely the current version but possibly not.

kumar303:
Oops, I misunderstood this. The user should always be able to update their rating even after they saved it. I'll fix the patch for that.

src/amo/api/index.js
```diff
@@ -28,3 +28,16 @@ export function submitReview({
...
+
+export function getUserReviews({ userId } = {}) {
+  return new Promise((resolve) => {
+    if (!userId) {
+      throw new Error('userId cannot be falsey');
+    }
+    // Warning: if you rely on this API as is, you'll have to implement
+    // your own paging.
+    resolve(callApi({
+      endpoint: `accounts/account/${userId}/reviews`,
```
kumar303: I think it will just return the first page of results. I hope to replace it with https://github.com/mozilla/addons-server/issues/3986 asap though.

src/amo/components/OverallRating.js
```diff
@@ -70,12 +104,70 @@ export class OverallRatingBase extends React.Component {
...
-  userId: state.auth && state.auth.userId,
-});
+export const mapStateToProps = (state, componentProps) => {
+  const userId = state.auth && state.auth.userId;
+  let userReview;
+
+  // Look for an already saved review by this user for this add-on/version.
+  if (userId && state.reviews) {
+    const allUserReviews = state.reviews[userId];
+    log.info(`Checking state for review by user ${userId},
```
kumar303: Sure, added. Dedent only trims leading whitespace, not new lines though.

mstriemer: Ah, I guess we really want [`oneLine`](https://github.com/declandewet/common-tags#oneline). Doesn't look like we have a babel plugin for it though. Could import it I guess since this isn't in a translation.

kumar303:
@mstriemer sorry for the churn but this is now ready for another review. The major changes were:

* The stars are always functional now. When you re-rate an add-on it is updated in the API accordingly.
* When the component looks for a saved user review, it fetches the latest one.
* The reviews are stored in state properly now (keyed by review ID) which should allow the code to scale.

tests/client/amo/testApi.js
```diff
@@ -92,5 +92,134 @@ describe('amo.api', () => {
...
+        .returns(Promise.resolve({
+          results: [],
+          next: '/reviews/next-page/',
+        }));
+
+      return getUserReviews({ userId: 123 })
+        .then(() => {
+          throw new Error('unexpected success');
+        })
+        .catch((error) => {
```
mstriemer: This will catch the error in the `then()` which works now but if someone changes the assertion in the catch it could start silently passing. Might be better to use the 2 argument for of `then()` but this is fine too.

src/amo/components/OverallRating.js
```diff
@@ -70,25 +97,57 @@ export class OverallRatingBase extends React.Component {
...
-        // TODO: when we have a user_id in the API response, we
-        // could probably use that instead.
-        // https://github.com/mozilla/addons-server/issues/3672
-        dispatch(setReview({
-          addonId,
-          rating: review.rating,
-          versionId: review.version.id,
-          userId,
-        }));
+        dispatch(setReview(review));
```
mstriemer: Will this always update the old rating if one exists? It probably makes sense to update the old review if it is for the same version otherwise seed the form with the old review and submit a new one on change.

src/amo/components/OverallRating.js
```diff
@@ -70,25 +97,57 @@ export class OverallRatingBase extends React.Component {
...
+        if (review) {
+          dispatch(setReview(review));
+        } else {
+          log.info(
+            `No saved review found for userId ${userId}, addonId ${addonId}`);
+        }
+      });
+  },
+
+  submitRating({ router, addonSlug, ...params }) {
```
kumar303: this is actually only submitting a rating though. It only submits the `rating` property of the review. It leaves the `body` null because that comes later. The data model is a bit weird. I don't mind changing everything to "review" but that was my rationale.

src/amo/components/OverallRating.js
kumar303: It updates the old review but only by review ID, not by version. The form is seeded with the latest old review, yes.

tests/client/amo/testApi.js
kumar303: ah, yes, I always forget about that. We should probably add a helper for this pattern like we have in [web-ext](https://github.com/mozilla/web-ext/blob/master/tests/unit/helpers.js#L92).

mstriemer: It probably makes sense to add a helper. I'm not sure what web-ext has would work. It looks like using that you could still get tests passing when they should fail.

```js
const mock = sinon.mock(something, 'someProp').expects('foo');
return doSomething()
  .then(makeSureItFails())
  .catch(() => {
    // Oops, error is "This test unexpectedly succeeded without an error"
    mock.verify();
   });
```

It looks like [chai-as-promised](https://github.com/domenic/chai-as-promised#assert-interface) has something for this:

```js
assert.isRejected(doSomething(), /userId cannot be falsey/);
// Looks like this would work, too:
assert.isRejected(doSomething(), /userId cannot be falsey/)
  .then(() => mock.verify());

src/amo/components/OverallRating.js
mstriemer: Would updating by review ID update the version or leave it as is?

src/amo/components/OverallRating.js
mstriemer: Okay, I wasn't sure if that was the case. It probably won't be as bad here as on Marketplace. Rating might be a better term (although the API isn't using it) since reviewers review add-ons. Seems unlikely we'd get content ratings added to the mix like Marketplace had.

kumar303: I just changed it all to "review" since the API calls them all reviews.

src/amo/components/OverallRating.js
kumar303: The API helper prevents changing the version anytime you are editing a review: https://github.com/mozilla/addons-frontend/pull/1374/files#diff-3d585cb73e6d00650d856aaf80203732R13

Without this check, the API returns an error since you can't change the version.

mstriemer: Right, I'm wondering if it should be:

```js
function prepareReview(review, addon) {
  if (review.id) {
    if (review.version !== addon.version) {
      // Clear id so this is added against the latest version.
      return { ...review, id: undefined, version: addon.version };
    }
    // Clear version so the current review is updated.
    return { ...review, version: undefined };
  }
  // New review, submit it normally.
  return review;
}

kumar303:
Thanks for all the reviews! I'm going to merge it but I can still address changes in another patch if needed.

PR #543, opened 2016-06-06
Add events to disco pane

muffinresearch:
Fixes #507
Fixes #550

PR #1382, opened 2016-11-16
Extract mozAddonManager listeners to core/installAddon

mstriemer:
This is a huge diff most it's mostly just moving code. Here's what I did:

1. Moved `setCurrentStatus` entirely into `mapDispatchToProps` in disco/Addon.
2. Moved `mapDispatchToProps` to core/installAddon.
3. Added a helper `withInstallHelpers` to core/installAddon to hookup the listeners to a component.
4. Made `withInstallHelpers` take a `src` parameter to set the install source (previously hardcoded to `discovery-promo`).
5. Move files from disco to core and update imports.

Hopefully reviewing this isn't too bad since it's largely unchanged, just moved around.

Supports #1291.


src/core/constants.js
```diff
@@ -97,3 +101,28 @@ export const acceptedInstallTypes = [
...
+export const INSTALL_CATEGORY = 'AMO Addon / Theme Installs';
+export const UNINSTALL_CATEGORY = 'AMO Addon / Theme Uninstalls';
+export const CLICK_CATEGORY = 'AMO Addon / Theme Clicks';
+
+// Error used to know that the setEnable method on addon is
+// not available.
+export const SET_ENABLE_NOT_AVAILABLE = 'SET_ENABLE_NOT_AVAILABLE';
+
+// Add-on statuses for mozAddonManager events.
+export const globalEventStatusMap = {
```
tofumatt: Shouldn't this be `GLOBAL_EVENT_STATUS_MAP`?

src/core/constants.js
```diff
@@ -97,3 +101,28 @@ export const acceptedInstallTypes = [
...
+  onInstalled: INSTALLED,
+  onUninstalling: UNINSTALLING,
+  onUninstalled: UNINSTALLED,
+  onEnabling: ENABLING,
+  onDisabling: DISABLING,
+};
+
+// The events here are set directly on mozAddonManager
+// they will be fired by addons and themes.
+export const globalEvents = Object.keys(globalEventStatusMap);
```
tofumatt: Same here, should this be `GLOBAL_EVENTS`?

src/core/addonManager.js
```diff
@@ -1,13 +1,10 @@
 /* global window */
-
 import log from 'core/logger';
 import {
+  installEventList,
```
tofumatt: Seems like this should be `INSTALL_EVENT_LIST` if it's a constant, even though it seems this is from a former commit and not introduced here.

src/core/addonManager.js
```diff
@@ -24,9 +21,16 @@ export function getAddon(guid, { _mozAddonManager = window.navigator.mozAddonMan
     });
 }
 
-export function install(_url, eventCallback,
-  { _mozAddonManager = window.navigator.mozAddonManager } = {}) {
-  const url = addQueryParams(_url, { src: 'discovery-promo' });
+export function install(_url, eventCallback, {
+  _mozAddonManager = window.navigator.mozAddonManager,
+    src,
```
tofumatt: What's up with this indentation?

src/core/addonManager.js
```diff
@@ -24,9 +21,16 @@ export function getAddon(guid, { _mozAddonManager = window.navigator.mozAddonMan
...
+} = {}) {
+  if (src === undefined) {
+    return new Promise(() => {
```
tofumatt: Sometimes my Promise-foo is rusty, but wouldn't

```js
return Promise.reject(new Error('No src for add-on install'))
```

work here just as well?

src/core/installAddon.js
```diff
@@ -0,0 +1,192 @@
...
+  SHOW_INFO,
+  START_DOWNLOAD,
+  UNINSTALL_CATEGORY,
+  UNINSTALLED,
+  UNINSTALLING,
+} from 'core/constants';
+import * as addonManager from 'core/addonManager';
+
+export function makeProgressHandler(dispatch, guid) {
+  return (addonInstall, e) => {
```
tofumatt: I dislike one-letter-variables. I know the convention is `e` for `event`, but could we call this `event`?

src/core/installAddon.js
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+
+    return {
+      setCurrentStatus() {
+        const { guid, installURL } = ownProps;
+        const payload = { guid, url: installURL };
+        return _addonManager.getAddon(guid)
+        .then(
+          (addon) => {
```
tofumatt: This seems like it would be better on the previous line.

src/core/installAddon.js
```diff
@@ -0,0 +1,192 @@
...
+          // Dispatch a generic error should the success/error functions throw.
+          dispatch({
+            type: INSTALL_STATE,
+            payload: { guid, status: ERROR, error: FATAL_ERROR },
+          });
+        });
+      },
+
+      enable({ _showInfo = showInfo } = {}) {
+        const { guid, name, iconUrl, i18n } = ownProps;
```
tofumatt: Haha the obsessive in me wants these to be alphabetised. 😉

src/core/installAddon.js
```diff
@@ -0,0 +1,192 @@
...
+        return _addonManager.enable(guid)
+        .then(() => {
```
tofumatt: This should be indented.

tests/client/core/TestAddonManager.js
```diff
@@ -91,12 +91,20 @@ describe('addonManager', () => {
...
         .then(() => {
           fakeInstallObj.onDownloadProgressListener(fakeEvent);
           assert.ok(fakeCallback.calledWith(fakeInstallObj, fakeEvent));
         });
     });
+
+    it('requires a src', () => (
+      addonManager.install(
+        fakeInstallUrl, fakeCallback, { _mozAddonManager: fakeMozAddonManager })
+          .then(
```
tofumatt: Could this be:

```js
.then(() => assert.ok(false, 'unexpected success'))
.catch((e) => assert.equal(e.message, 'No src for add-on install'))
```
instead?

muffinresearch: If the unexpected success assertion throws in the `then` we want the test runner to catch it so we get the "unexpected success" message. 

If there was a catch as you suggest you'd only see the exception was wrong which is makes the reason for the failure less clear.

See http://jsbin.com/yatofoc/edit?js,console for an example of the differences.

tofumatt: Ah, right on. Rarely see the second argument to `then()` these days so it just caught me off-guard. 👍

muffinresearch: Yeah I wish these aspects of Promises were more obvious - I always forget the precise details of the subtle differences.

src/core/constants.js
mstriemer: It would seem that way. Just looked up https://github.com/mozilla/addons-frontend/pull/543 which appears to be where they came from.

Not sure if @muffinresearch had a reason in mind to lowercase them but I didn't comment on it. It looks like the strings are uppercase but the ones that are arrays or objects are camelCase.

src/core/addonManager.js
mstriemer: My indenter wants it to go there, not sure why. I moved all the args to the next line.

src/core/addonManager.js
mstriemer: I was cautious to do it that way because I wasn't 100% sure. It looks like the tests still pass with this so should be good.

src/core/installAddon.js
mstriemer: I always avoided using `event` because it was highlighted differently in my edit back in the day. I finally bothered to look into it and it looks like (at least in Firefox) `event` is set if you have something like `<p onClick="doSomething(event)">` but it appears it is not set anywhere else. I'll change it.

🌟 The more you (or at least I) know.

muffinresearch:
@mstriemer It would be good to highlight any install button related to changes to QA so they know to test the buttons on the disco pane a bit more as these PRs land.

src/core/installAddon.js
mstriemer: Sure, you end up with a `}, () => {` for the catch part but that's how I normally would have written it.

src/core/constants.js
mstriemer: I changed them to `SCREAMING_SNAKE_CASE`.

PR #1388, opened 2016-11-17
Hook up install button

mstriemer:
This needs a good cleanup. We're spreading the addon and installation all over the place and add-ons have `status` and installations have `status` and they are incompatible. I might take @kumar303's advice and just pass the `guid` instead, that should be all we need to retrieve the data.

Could probably extract some of the other setup stuff into `withInstallHelpers` too.

~~TODO: File an issue to get `InfoDialog` setup and responsive.~~ https://github.com/mozilla/addons-frontend/issues/1414

install-addon-amo mov

Supports #1291.


mstriemer:
I filed https://github.com/mozilla/addons-frontend/issues/1416 to clean up mashing the `addon` and `installation` props together and filed https://github.com/mozilla/addons-frontend/issues/1417 to extract some more code into core/installAddon.

This should be ready for review now. I tested it in amo and disco.

src/amo/components/AddonDetail.js
```diff
@@ -110,4 +116,7 @@ class AddonDetail extends React.Component {
   }
 }
 
-export default translate({ withRef: true })(AddonDetail);
+export default compose(
+  translate({ withRef: true }),
+  withInstallHelpers({ src: 'dp-btn-primary' }),
```
tofumatt: I'm guessing this means "DiscoPane Button Primary" but why is it wrtn lke tht?

src/amo/containers/DetailPage.js
```diff
@@ -18,11 +18,14 @@ export class DetailPageBase extends React.Component {
...
-function mapStateToProps(state, ownProps) {
+export function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
+  const addon = state.addons[slug];
+  const installation = state.installations[addon.guid] || { status: UNKNOWN };
   return {
-    addon: state.addons[slug],
-    slug,
+    addon,
+    ...addon,
```
tofumatt: I'm guessing #1416 is filed for this, right?

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -16,23 +15,26 @@ import { fakeAddon } from 'tests/client/amo/helpers';
 import { getFakeI18nInst } from 'tests/client/helpers';
 
 
-function render({ addon = fakeAddon, ...customProps } = {}) {
+function render({ addon = fakeAddon, setCurrentStatus = sinon.spy(), ...customProps } = {}) {
   const i18n = getFakeI18nInst();
   const initialState = { api: { clientApp: 'android', lang: 'pt' } };
   const props = {
+    ...addon,
```
tofumatt: The spread (`...addon`) is added after `addon` above (https://github.com/mozilla/addons-frontend/pull/1388/files#diff-d736a30c9a5a02800a9c3f44a64734bdR27)

Not sure if that matters, just noticed it.

src/amo/components/AddonDetail.js
mstriemer: I'm not sure why, that's what we expect in GA though so I don't think we can change it.

src/amo/containers/DetailPage.js
mstriemer: Yup.

tests/client/amo/components/TestAddonDetail.js
mstriemer: I doubt it will make a difference in the end but it should be consistent. I'll update it.

PR #1392, opened 2016-11-18
Add more info to addon details page (fix #1294)

### LINKED ISSUE #1294: Implement the details page more information widget

tofumatt:
So! I've fixed this up and sent it as a separate PR. Supersedes #1354.

Adds "More information" to addon detail pages. (Same as before.)

### Before
screenshot 2016-11-08 14 13 38

### After
screenshot 2016-11-08 14 25 34

### DIFF
--- a/src/amo/components/AddonMoreInfo.js
+++ b/src/amo/components/AddonMoreInfo.js
@@ -1,0 +1,72 @@
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+
+import Link from 'amo/components/Link';
+import translate from 'core/i18n/translate';
+
+import './AddonMoreInfo.scss';
+
+
+export class AddonMoreInfoBase extends React.Component {
+  static propTypes = {
+    addon: PropTypes.object.isRequired,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { addon, i18n } = this.props;
+
+    // TODO: Use the addonType for the privacy text, so it reads
+    // "for this extension", "for this theme", etc.
+    return (
+      <section className="AddonMoreInfo">
+        <h2 className="AddonMoreInfo-header">
+          {i18n.gettext('More information')}
+        </h2>
+
+        <dl className="AddonMoreInfo-contents">
+          {addon.homepage ? <dt>{i18n.gettext('Website')}</dt> : null}
+          {addon.homepage ? (
+            <dd>
+              <a href={addon.homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+            </dd>
+          ) : null}
+          <dt>{i18n.gettext('Version')}</dt>
+          <dd ref={(ref) => { this.version = ref; }}>
+            {addon.current_version.version}
+          </dd>
+          <dt>{i18n.gettext('Last updated')}</dt>
+          <dd>{addon.last_updated}</dd>
+          {addon.current_version.license ? (
+            <dt ref={(ref) => { this.licenseHeader = ref; }}>
+              {i18n.gettext('License')}
+            </dt>
+          ) : null}
+          {addon.current_version.license ? (
+            <dd>
+              <a href={addon.current_version.license.url}
+                ref={(ref) => { this.licenseLink = ref; }}>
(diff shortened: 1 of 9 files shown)


src/amo/components/AddonMoreInfo.js
```diff
@@ -0,0 +1,75 @@
...
+              <a href={addon.homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+            </dd>
+          ) : null}
+          <dt>{i18n.gettext('Version')}</dt>
+          <dd ref={(ref) => { this.version = ref; }}>
+            {addon.current_version.version}
+          </dd>
+          <dt>{i18n.gettext('Last updated')}</dt>
+          <dd>{addon.last_updated}</dd>
```
tofumatt: I still need to format this date, my bad.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -25,9 +27,11 @@ function render({ addon = fakeAddon, ...customProps } = {}) {
   };
 
   return findRenderedComponentWithType(renderIntoDocument(
-    <I18nProvider i18n={i18n}>
-      <AddonDetail {...props} />
-    </I18nProvider>
+    <Provider store={createStore(initialState)}>
```
tofumatt: Because an add-on now uses the `AddonMoreInfo` component which has an AMO-relative `Link` component, this needed to be added so the tests wouldn't fail.

src/amo/components/AddonMoreInfo.js
tofumatt: I'd like this to look like: screenshot 2016-11-18 20 47 23
with moment installed.

src/amo/components/AddonMoreInfo.js
```diff
@@ -0,0 +1,75 @@
...
+  static propTypes = {
+    addon: PropTypes.object.isRequired,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { addon, i18n } = this.props;
+
+    if (!addon) {
+      return null;
```
kumar303: I'd like to see this check done in the parent instead:
````javascript
{addon ? <AddonMoreInfo addon={addon} /> : null}
````

Otherwise, what purpose would  `AddonMoreInfo` serve for `null` add-ons? Anyway, the `addon` property is declared to be required which I think is the right thing to do.

File: src/amo/components/AddonMoreInfo.js (hunk 1 above)
Comment (kumar303) [inline src/amo/components/AddonMoreInfo.js]: yes, filing small issues to fix up things like this is the best way to keep moving forward IMO

File: tests/client/amo/components/TestAddonDetail.js (hunk 2 above)
Comment (kumar303) [inline tests/client/amo/components/TestAddonDetail.js]: I've addressed this type of thing by passing in a stub `AddonMoreInfo` component to isolate it from the test (it has its own tests anyway). Adding a `Provider` doesn't seem so bad in this case though.

File: tests/client/amo/components/TestAddonMoreInfo.js [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+        <AddonMoreInfoBase i18n={getFakeI18nInst()} addon={fakeAddon}
+          {...props} />
+      </Provider>
+    ), AddonMoreInfoBase);
+  }
+
+  it('does not render if no add-on exists', () => {
+    const root = render({ addon: null });
+
+    assert.equal(root.textContent, null);
```
kumar303: Following my comment above, this test would move up to the parent.

src/amo/components/AddonMoreInfo.js
tofumatt: Yeah, turns out I was being overly cautious; the parent component won't render without an `addon` anyway and it's required here too, so this is overkill. I moved the check up but it didn't even matter then, so I've just removed this.

tofumatt:
Addressed the changes, should be good for another look now.

Issue #1371, opened 2016-11-11
Change all raw state representations to redux dispatches in test code


kumar303:
A lot of our unit tests for functions like `mapStateToProps` pass in a raw state representation (i.e. a plain object). This is dangerous because it's hard to keep raw objects in sync with reducer code as the state evolves.

Here is an example:

````javascript
const state = {
  api: {
    lang: 'en-US',
    clientApp: 'firefox',
    token: 'secret-token',
  },
};
const props = mapStateToProps(state);
// make assertions about the props
````

Instead, let's use a real store and dispatch real actions.

````javascript
import createStore from 'amo/store';
import { setJWT } from 'core/actions';
import { userAuthToken } from 'tests/client/helpers';

const store = createStore();
// This would create the equivalent state and would more 
// closely resemble how a user signs into the real app.
store.dispatch(setJWT(userAuthToken()));
const props = mapStateToProps(store.getState());
// make assertions about the props
````

tofumatt:
Bonus: it will likely make many of our tests a lot shorter and easier to parse!

kumar303:
Let's move everything to ducks first https://github.com/mozilla/addons-frontend/issues/1299

tofumatt:
We've closed https://github.com/mozilla/addons-frontend/issues/1452 but we should implement it when we do this.

kumar303:
This is mostly done. We've written all new code this way for a long time now.

willdurand:
> This is mostly done. We've written all new code this way for a long time now.

Yep, and on this basis, I am closing this issue now.

PR #1398, opened 2016-11-21
Add category page (close #1378)

### LINKED ISSUE #1378: Show search results for categories on /categories/ pages
Allow `/extensions/alerts-updates/` URLs to show all add-ons in a category.

tofumatt:
Add the actual category URL (matches addons-server) including pagination.

Screenshot:

nov-21-2016 12-21-02

### DIFF
--- a/src/amo/containers/CategoryPage.js
+++ b/src/amo/containers/CategoryPage.js
@@ -1,0 +1,41 @@
+import { connect } from 'react-redux';
+import { asyncConnect } from 'redux-connect';
+import { compose } from 'redux';
+
+import SearchPage from 'amo/components/SearchPage';
+import {
+  filtersMatch,
+  loadByCategoryIfNeeded,
+  parsePage,
+} from 'core/searchUtils';
+
+
+export function mapStateToProps(state, ownProps) {
+  const filters = {
+    addonType: ownProps.params.addonType,
+    category: ownProps.params.slug,
+    clientApp: ownProps.params.application,
+  };
+  const pathname = `/${filters.addonType}s/${filters.category}/`;
+  const queryParams = { page: parsePage(ownProps.location.query.page) };
+
+  if (filtersMatch(state.search.filters, filters)) {
+    return {
+      hasSearchParams: true,
+      filters,
+      pathname,
+      queryParams,
+      ...state.search,
+    };
+  }
+
+  return { hasSearchParams: true, pathname, queryParams };
+}
+
+export default compose(
+  asyncConnect([{
+    deferred: true,
+    promise: loadByCategoryIfNeeded,
+  }]),
+  connect(mapStateToProps),
+)(SearchPage);
--- a/src/core/searchUtils.js
+++ b/src/core/searchUtils.js
@@ -92,8 +92,32 @@ export function loadSearchResultsIfNeeded({ store: { dispatch, getState }, locat
     ...location.query,
     clientApp: state.api.clientApp,
   });
+
   if (!isLoaded({ state: state.search, page, filters })) {
     return performSearch({ dispatch, page, api: state.api, auth: state.auth, filters });
+  }
+  return true;
+}
+
+export function loadByCategoryIfNeeded(
+  { store: { dispatch, getState }, location, params }
+) {
+  const state = getState();
+  const filters = {
+    addonType: params.addonType,
+    category: params.slug,
+    clientApp: params.application,
+  };
+  const page = parsePage(location.query.page);
+
+  if (!isLoaded({ state: state.search, page, filters })) {
+    return performSearch({
+      api: state.api,
+      auth: state.auth,
+      dispatch,
+      filters,
+      page,
+    });
   }
   return true;
 }
--- a/tests/client/amo/containers/TestCategoryPage.js
+++ b/tests/client/amo/containers/TestCategoryPage.js
@@ -1,0 +1,56 @@
+import { mapStateToProps } from 'amo/containers/CategoryPage';
+import createStore from 'amo/store';
+import { searchStart } from 'core/actions/search';
+
+
+describe('CategoryPage.mapStateToProps()', () => {
+  let filters;
+  let ownProps;
+
+  before(() => {
+    filters = {
+      addonType: 'theme',
+      category: 'ad-block',
+      clientApp: 'firefox',
+    };
+    ownProps = {
+      location: { query: {} },
+      params: {
+        application: 'firefox',
+        addonType: 'theme',
+        slug: 'ad-block',
+      },
+    };
+  });
+
+  it('passes the search state if the filters and state matches', () => {
+    const store = createStore();
+    store.dispatch(searchStart({ filters }));
+    const props = mapStateToProps(store.getState(), ownProps);
+
+    assert.deepEqual(props, {
+      count: 0,
+      filters,
+      hasSearchParams: true,
+      loading: true,
+      page: undefined,
+      pathname: '/themes/ad-block/',
+      queryParams: { page: 1 },
+      results: [],
+    });
+  });
+
+  it('does not pass search state if the filters and state do not match', () => {
+    const store = createStore();
+    store.dispatch(searchStart({ filters }));
+    const mismatchedState = store.getState();
+    mismatchedState.search.filters.clientApp = 'nothing';
+    const props = mapStateToProps(mismatchedState, ownProps);
+
+    assert.deepEqual(props, {
+      hasSearchParams: true,
+      pathname: '/themes/ad-block/',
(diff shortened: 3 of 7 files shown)


tests/client/amo/containers/TestCategoryPage.js
```diff
@@ -0,0 +1,57 @@
+import deepcopy from 'deepcopy';
+
+import { mapStateToProps } from 'amo/containers/CategoryPage';
+
+
+describe('CategoryPage.mapStateToProps()', () => {
+  const state = {
```
kumar303: I think it would be less fragile to create this state by dispatching real actions. It should be just as easy to do. Example: https://github.com/mozilla/addons-frontend/issues/1371

tests/client/amo/containers/TestCategoryPage.js
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+
+    assert.deepEqual(props, {
+      hasSearchParams: true,
+      pathname: '/themes/ad-block/',
+      queryParams: { page: 1 },
+    });
+  });
+});
```
kumar303: This appears to be missing a test for the case when `state.search` is empty (i.e. `{}`).

tests/client/core/test_searchUtils.js
```diff
@@ -16,3 +21,77 @@ describe('searchUtils mapStateToProps()', () => {
     assert.deepEqual(props, { hasSearchParams: false });
   });
 });
+
+describe('searchUtils loadCategoryResultsIfNeeded()', () => {
+  const state = {
```
kumar303: I'd also like to this state built by dispatching real actions, if possible

tests/client/core/test_searchUtils.js
```diff
@@ -16,3 +21,77 @@ describe('searchUtils mapStateToProps()', () => {
...
+      getState: () => stateWithPage,
+    };
+    const location = { query: { page } };
+    const params = {
+      addonType: 'extension',
+      application: 'firefox',
+      slug: 'ad-block',
+    };
+    assert.strictEqual(
+      loadCategoryResultsIfNeeded({ store, location, params }), true);
```
kumar303: I think this test would be better if it verified that `mockApi` is never called, which is a stronger indicator that data is not fetched from the API.

src/core/searchUtils.js
```diff
@@ -92,6 +92,22 @@ export function loadSearchResultsIfNeeded({ store: { dispatch, getState }, locat
...
     clientApp: state.api.clientApp,
   });
+
+  if (!isLoaded({ state: state.search, page, filters })) {
+    return performSearch({ dispatch, page, api: state.api, auth: state.auth, filters });
+  }
+  return true;
+}
+
+export function loadCategoryResultsIfNeeded({ store: { dispatch, getState }, location, params }) {
```
kumar303: This could be shortened slightly to `loadByCategoryIfNeeded`

src/core/searchUtils.js
```diff
@@ -92,6 +92,22 @@ export function loadSearchResultsIfNeeded({ store: { dispatch, getState }, locat
...
+  const page = parsePage(location.query.page);
+  const state = getState();
+  const filters = {
+    addonType: params.addonType,
+    category: params.slug,
+    clientApp: params.application,
+  };
+
   if (!isLoaded({ state: state.search, page, filters })) {
```
kumar303: Why do the search results need to be cached? I think it should make an API request every time. Otherwise, the single page app will go stale -- the only way a user can obtain newly added results is to refresh the page. The browser will already cache the API call as needed so I don't really see the value in adding a new caching layer on top of that.

src/core/searchUtils.js
```diff
@@ -92,6 +92,22 @@ export function loadSearchResultsIfNeeded({ store: { dispatch, getState }, locat
     ...location.query,
     clientApp: state.api.clientApp,
   });
+
+  if (!isLoaded({ state: state.search, page, filters })) {
+    return performSearch({ dispatch, page, api: state.api, auth: state.auth, filters });
+  }
+  return true;
```
tofumatt: I agree actually, as it makes testing weird too. That said:

it-was-like-that-when-i-got-here

I've filed an issue for it (#1403), as changing this across the app would be out-of-scope for the PR.

src/core/searchUtils.js
tofumatt: Good point. Again, it was like that when I got here; the search page currently caches results. When flipping through paginated results it makes going back and forth between pages quick.

I've filed #1404, but I'll remove it from this bit of code because why not and I agree; we can always add it back.

src/amo/containers/CategoryPage.js
```diff
@@ -0,0 +1,41 @@
...
+import SearchPage from 'amo/components/SearchPage';
+import {
+  filtersMatch,
+  loadCategoryResultsIfNeeded,
+  parsePage,
+} from 'core/searchUtils';
+
+
+export function mapStateToProps(state, ownProps) {
+  const filters = {
```
tofumatt: But these filters are declared here to compare them with state and make sure everything matches. They're passed to `if (filtersMatch([...])) {`, so moving them into there would mean duplication.

They're only passed to the state if they match.

tests/client/amo/containers/TestCategoryPage.js
tofumatt: I don't think that should ever happen, as it's set by `searchStart`. Am I missing a path to an empty search state?

tofumatt:
Okay, I think I fixed all the issues (mostly regarding state testing) in the PR, and created issues for the things that were slightly out-of-scope. Ready for another r?

tofumatt:
Odd: https://travis-ci.org/mozilla/addons-frontend/builds/177831776 is failing but passing locally, and the error is quite strange.

mstriemer:
Try rebasing with master. That's what I was running into yesterday and switching back to npm seemed to fix it.

tofumatt:
Ah right, cool. Figured this was related, cheers.

src/amo/containers/CategoryPage.js
kumar303: Ah, ok, I see on a second look that all of these do need to be defined outside of the if block.

src/core/searchUtils.js
```diff
@@ -92,8 +92,28 @@ export function loadSearchResultsIfNeeded({ store: { dispatch, getState }, locat
...
     clientApp: state.api.clientApp,
   });
+
   if (!isLoaded({ state: state.search, page, filters })) {
     return performSearch({ dispatch, page, api: state.api, auth: state.auth, filters });
   }
   return true;
 }
+
+export function loadByCategoryIfNeeded(
```
kumar303: This name no longer makes sense :) I think `loadByCategory` would be fine.

tests/client/amo/containers/TestCategoryPage.js
```diff
@@ -0,0 +1,56 @@
...
+      queryParams: { page: 1 },
+      results: [],
+    });
+  });
+
+  it('does not pass search state if the filters and state do not match', () => {
+    const store = createStore();
+    store.dispatch(searchStart({ filters }));
+    const mismatchedState = store.getState();
+    mismatchedState.search.filters.clientApp = 'nothing';
```
kumar303: Using dispatch (like above) is the way to do it IMO. By using dispatch you are not coupling this code with the raw state representation. Here would be a way to rewrite it:
````javascript
store.dispatch(searchStart({ filters: { ...filters, clientApp: 'nothing' } }));
const mismatchedState = store.getState();
````

File: tests/client/amo/containers/TestCategoryPage.js (hunk 2 above)
Comment (kumar303) [inline tests/client/amo/containers/TestCategoryPage.js]: The component may render with `state.search = {}` for a split second if the API response is slow but, no worries, we could always add the test later if a bug appears with handling empty state.

File: tests/client/core/test_searchUtils.js [hunk 11]
```diff
@@ -16,3 +20,38 @@ describe('searchUtils mapStateToProps()', () => {
     assert.deepEqual(props, { hasSearchParams: false });
   });
 });
+
+describe('searchUtils loadByCategoryIfNeeded()', () => {
```
kumar303: `loadByCategoryIfNeeded` would need to be renamed too

src/core/searchUtils.js
mstriemer: Is this making another API request on the client after JS loading on initial render? I think that's why we've normally included the caching.

tests/client/amo/containers/TestCategoryPage.js
tofumatt: Oops, good point. Will do.

PR #1400, opened 2016-11-21
Ensure installURL is set before calling withInstallHelpers

mstriemer:
The props weren't getting set for the updated `withInstallHelpers` HOC so it was failing on install. I made it throw now if it is misconfigured.

Just had to reorder the `compose()` in `Addon` to fix this.

Fixes mozilla/addons-server#4035.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,7 @@
     "bunyan": "1.8.5",
     "camelcase": "4.0.0",
     "classnames": "2.2.5",
+    "common-tags": "^1.4.0",
     "config": "1.24.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -160,6 +160,7 @@
     "bunyan": "1.8.5",
     "camelcase": "4.0.0",
     "classnames": "2.2.5",
+    "common-tags": "^1.4.0",
```
tofumatt: 👍

mstriemer:
Superseded by #1407.

Issue #1390, opened 2016-11-18
Always compare existing add-on reviews to the current version


kumar303:
### Describe the problem and steps to reproduce it:

* Submit a 5 star rating for an add-on named X at version 1
* Wait for the author of X to publish version 2 of their add-on
* Go back to the details page for add-on X
* You will see your saved star rating
* Decide that you really hate it; change your star rating to only 1 star

### What happened?

The code will see that your existing rating (review) was for an old version so it won't update it. It will instead submit a new rating for the version your are viewing.

### What did you expect to happen?

It should submit a new rating but not for the version you are viewing. It should submit a rating for the *current version* of the add-on which may not be the one you are viewing.

See https://github.com/mozilla/addons-frontend/pull/1374 for where this feature first landed.

PR #1406, opened 2016-11-21
Updating an old review should create a new review against the current add-on version

kumar303:
This is pretty edge-casey since we don't currently have a way to view history add-on versions but I think it's the right thing to do. It should future proof the site for when we do show multiple versions (or is that YAGNI? Sigh).

Fixes https://github.com/mozilla/addons-frontend/issues/1390

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -99,9 +99,7 @@ class AddonDetail extends React.Component {
         <section className="overall-rating">
           <h2>{i18n.gettext('Rate your experience')}</h2>
           <OverallRating
-            addonName={addon.name}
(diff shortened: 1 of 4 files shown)


mstriemer:
I don't think this is an edge case. If I leave a review and then the add-on gets updated (seems quite likely) then I'd hit this, right?

kumar303:
> If I leave a review and then the add-on gets updated (seems quite likely) then I'd hit this, right?

Yes, but my patch doesn't really fix that because `current_version` will always point to the data retrieved when you first loaded the page. There is no easy way to fix what you mentioned. Perhaps we need to submit a `null` version and let the API figure it out?

PR #1407, opened 2016-11-21
Ensure installURL is set before calling withInstallHelpers 

mstriemer:
The props weren't getting set for the updated `withInstallHelpers` HOC so it was failing on install. I made it throw now if it is misconfigured.

Just had to reorder the `compose()` in `Addon` to fix this.

This also tells Travis to use `npm install` instead of `yarn` to install things since their update seems to have broken the build? Maybe we need to update some config but I'm not sure what.

Fixes mozilla/addons-server#4035.

Moved from https://github.com/mozilla/addons-frontend/pull/1400 since when Travis runs yarn it seems to break that PR for good.


mstriemer:
Since @tofumatt approved #1400 before Travis got mad at me and Travis is now happy I'll merge this.

Issue #1381, opened 2016-11-16
Allow existing review text to be edited


kumar303:
When you rate an add-on and return to its detail page, you will see your saved star rating (as of https://github.com/mozilla/addons-frontend/issues/1088). You can also select a new star rating which will update your chosen rating. This redirects you to the review text entry.

The review text entry needs to be populated with your saved text. Currently it is empty.

madalincm:
Verified as fixed in FF50(Android 6.0) in addons-dev.allizom.org
After taping on the stars the user is redirected and can edit the previous review.
Marking as verified fixed.

PR #1409, opened 2016-11-21
Allow existing review text to be edited

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1381

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -4,6 +4,7 @@ export const setReview = (review, reviewOverrides = {}) => ({
   type: SET_REVIEW,
   data: {
     addonId: review.addon.id,
+    addonSlug: review.addon.slug,
     body: review.body,
     id: review.id,
     isLatest: review.is_latest,
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -2,7 +2,7 @@ import { SET_REVIEW } from 'amo/constants';
 
 export const setReview = (review, reviewOverrides = {}) => ({
   type: SET_REVIEW,
-  data: {
+  payload: {
     addonId: review.addon.id,
     addonSlug: review.addon.slug,
     body: review.body,
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -3,12 +3,13 @@ import { SET_REVIEW } from 'amo/constants';
 export const setReview = (review, reviewOverrides = {}) => ({
   type: SET_REVIEW,
   data: {
-    id: review.id,
     addonId: review.addon.id,
-    rating: review.rating,
-    versionId: review.version.id,
+    body: review.body,
+    id: review.id,
     isLatest: review.is_latest,
+    rating: review.rating,
     userId: review.user.id,
+    versionId: review.version.id,
     ...reviewOverrides,
   },
 });
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -94,8 +94,6 @@ export class AddonReviewBase extends React.Component {
       );
     }
 
-    // TODO: I guess we should load the existing review text so it
-    // can be edited? That flow needs more thought.
     return (
       <div className="AddonReview">
         <h2 className="AddonReview-header">{i18n.gettext('Write a review')}</h2>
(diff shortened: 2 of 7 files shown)


src/amo/components/AddonReview.js
```diff
@@ -112,7 +104,9 @@ export class AddonReviewBase extends React.Component {
             className="AddonReview-textarea"
             ref={(ref) => { this.reviewTextarea = ref; }}
             name="review"
-            placeholder={placeholder} />
+            placeholder={placeholder}>
+            {this.props.review.body}
```
mstriemer: Looks like this could use `review.body` instead.

src/amo/components/AddonReview.js
```diff
@@ -147,13 +148,9 @@ export function loadAddonReview(
...
-      dispatch(setReview(review));
-      const reviewData = {
-        rating: review.rating,
-        addonSlug: slug,
-        id: review.id,
-      };
-      return reviewData;
+      const action = setReview(review);
+      dispatch(action);
+      return action.data;
```
mstriemer: data ~> payload

src/amo/reducers/reviews.js
```diff
@@ -22,12 +22,7 @@ export default function reviews(state = initialState, { data, type }) {
     case SET_REVIEW: {
       const existingReviews =
         state[data.userId] ? state[data.userId][data.addonId] : {};
```
mstriemer: data ~> payload

tests/client/amo/components/TestOverallRating.js
```diff
@@ -357,18 +357,14 @@ describe('OverallRating', () => {
     });
 
     it('sets a user review to the latest matching one in state', () => {
-      const isLatest = true;
       signIn({ userId: fakeReview.user.id });
 
-      store.dispatch(setReview(fakeReview, { isLatest }));
+      const action = setReview(fakeReview, { isLatest: true });
+      store.dispatch(action);
+      const dispatchedReview = action.data;
```
mstriemer: data ~> payload

tests/client/amo/components/TestOverallRating.js
```diff
@@ -402,25 +398,20 @@ describe('OverallRating', () => {
...
       signIn({ userId: fakeReview.user.id });
 
-      const reviewBase = {
-        versionId: fakeReview.version.id,
-        rating: 5,
-      };
+      function createReview(overrides) {
+        const action = setReview(fakeReview, overrides);
+        store.dispatch(action);
+        return action.data;
```
mstriemer: data ~> payload

src/amo/actions/reviews.js
```diff
@@ -3,12 +3,14 @@ import { SET_REVIEW } from 'amo/constants';
 export const setReview = (review, reviewOverrides = {}) => ({
```
mstriemer: Is `reviewOverrides` needed? Couldn't the caller do this just as easily? Looks like it's only for tests.

```jsx
const action = setReview({ ...fakeReview, isLatest: true });
// rather than
const action = setReview(fakeReview, { isLatest: true });

File: src/amo/actions/reviews.js [hunk 7]
```diff
@@ -3,12 +3,14 @@ import { SET_REVIEW } from 'amo/constants';
...
   type: SET_REVIEW,
   data: {
```
kumar303: Thanks, good call

src/amo/actions/reviews.js
mstriemer: Couldn't you do this though? I don't see it doing anything special.

```js
const state = reviews(state, setReview({
      ...fakeReview,
      id: 1,
      userId: 1,
      addonId: 1,
      rating: 1,
    }));

kumar303: The first argument, `review`, is an object from the API. It gets denormalized [like this](https://github.com/kumar303/addons-frontend/blob/817985a0faa8d83fba3100bce7f4f1933f8efc2b/src/amo/actions/reviews.js#L12) so the code you suggested won't work (it would add `review.userId` which would do nothing because `review.user.id` is the only property used).

kumar303:
Thanks for the review!

src/amo/actions/reviews.js
mstriemer: Oh, I see. Yeah that makes sense.

Issue #1032, opened 2016-09-06
Handle API errors in client-side rendered views


muffinresearch:
We now pass up errors from the API in server-side rendered views, however we will need to surface errors correctly in client-side code since that's a different code path.

redux-connect dispatches a `LOAD_FAIL` action, so we could look to handle that.

PR #1418, opened 2016-11-23
Create a shared error handler component

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1032

This is a very early patch to get some feedback on an idea for error handling.

This is the idea:
* Any component will be able to display informative errors returned by its API requests
* The main application component could display unhandled errors (this needs work but it's possible through state introspection)
* By using the redux state for errors, we can capture the error state in redux devtools and do replay/diagnostics

You can test it like this:
* Submit and approve an add-on
* Go to the add-on listing
* Log in as the same user who created the add-on
* Try to submit a rating for the add-on
* You should see the error below

screenshot 2016-11-23 16 44 05


kumar303:
@mstriemer @muffinresearch @tofumatt any thoughts on this? Please ignore test failures, lint errors, nits, and the names of things. I expect this will change drastically but I ran out of time before the holidays so hey :)

tofumatt:
I definitely love the idea of it (components can handle their errors locally, but unhandled errors being handled app-wide is nice, especially for more edge-case errors). I'll have a look through the code now.

src/core/api/index.js
```diff
@@ -46,21 +46,37 @@ export function callApi({
 
   return fetch(apiURL, options)
     .then((response) => {
+      return response.json().then((jsonResponse) => {
```
mstriemer: These `then`s should move down a line when you tidy this up.

kumar303:
I simplified the approach so that wrapped components get an `errorHandler` object that they can pass directly to `callApi()`. I'm going to try and keep this patch small by finishing out this approach with tests and better error parsing.

In a future patch, I think we could show arbitrary API errors at the global app level like this within `callApi()`:

````
if (!errorHandler) {
  errorHandler = getDefaultErrorHandler();
}
````
This would be connected to the main App component which would probably just render the error at the top of the screen. 

This would give us two choices:
- Any component could use `withErrorHandling()` to render API errors themselves. They would do this by passing their `errorHandler` object explicitly to `callApi()`.
- The default behavior would be for components to operate without any special error handling, i.e. like passing `errorHandler=null` to `callApi()`. This would invoke the default error message display.

**Comment by kumar303** [comment]
oops, I broke the server tests in legitimate ways. Let me take a look...

File: src/core/errorHandler.js [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+            <ul className="ErrorHandler-list">
+              {errorMessages}
+            </ul>
+          );
+        }
+
+        return (
+          <div>
+            {errorInfo}
+            <WrappedComponent {...props} />
```
mstriemer: If you move creating the error handler to `mapDispatchToProps()` then you could remove this component and just return the `connect()` call on the wrapped component. To render the errors you could expose the `errorInfo` bit above as `ErrorList` and that could be imported instead. That would let the caller put the error messages wherever makes sense.

```jsx
import { ErrorList, withErrorHandling } from 'core/errorHandler';
class AllReviewsBase extends React.Component {
  render() {
    const { error, reviews } = this.props;
    return (
      <div className="AllReviews">
        <h1>All Reviews</h1>
        <ErrorList error={error} />
        <ul>
          {reviews.map((review) => <li>{review.text}</li>)}
        </ul>
       </div>
    );
  }
}
export default withErrorHandling({ name: "AllReviews" })(AllReviewsBase);
```

This would also avoid putting that extra `<div />` in the DOM too.

src/core/actions/errors.js
```diff
@@ -0,0 +1,15 @@
+import { SET_ERROR } from 'core/constants';
+
+export function setError({ error, id } = {}) {
```
kumar303: Technically it's not needed, yes, but without it the error would be ``'_ref is undefined'`` which is way more confusing than ``'id cannot be empty'``. This is one thing I hate about ES6! Maybe the native error for this will be more sensical.

src/core/errorHandler.js
```diff
@@ -0,0 +1,79 @@
...
+        );
+      }
+    }
+
+    const mapStateToProps = (state) => {
+      log.info(`Looking for errors in state with ID ${id}`);
+      return { error: state.errors[id] };
+    };
+
+    const mapDispatchToProps = (dispatch) => ({ dispatch });
```
kumar303: It needs to be per instance, yeah, so that the ID is unique. You reminded me that I forgot a test for that too!

tests/client/core/testErrorHandler.js
```diff
@@ -0,0 +1,140 @@
...
+  const ComponentWithErrorHandling =
+    withErrorHandling({ id, name: 'SomeComponent' })(SomeComponent);
+
+  const provider = renderIntoDocument(
+    <Provider store={store}>
+      <ComponentWithErrorHandling />
+    </Provider>
+  );
+  const component = findRenderedComponentWithType(provider, SomeComponent)
+    .getWrappedInstance();
```
kumar303: Oh, huh. All the tests still pass without `.getWrappedInstance()`, nice. Sometimes I get missing property errors without it but I guess I don't fully understand how it works.

src/core/errorHandler.js
kumar303: This looks like a more flexible approach but it would also add more boilterplate. I think the behavior of displaying an error at the top would be adequate for most components so what I'd like to do is proceed with the current approach and expand it to your more flexible approach later as an option in a future patch. It could be done maybe more like `withErrorHandling({ manageErrorView: false, ... })` for components that want to manage the view themselves.

src/core/reducers/errors.js
```diff
@@ -0,0 +1,54 @@
...
+ * from it. Otherwise, an array containing a generic error message is returned.
+ */
+function getMessagesFromError(error) {
+  let messages = [gettext('An unexpected error occurred')];
+  log.info('Extracting messages from error object:', error);
+
+  if (error && error.response && error.response.data) {
+    const apiMessages = [];
+
+    Object.keys(error.response.data).forEach((key) => {
```
kumar303: It's not ideal as is but it's only a few extra lines of code. I'm not sure it's worth introducing lodash as a top level dep just for this (I cringe at the thought of more greenkeeper spam).

mstriemer: It would be nice to use lodash in other places where we map/iterate over objects. The `Objects.keys(obj).map((k) => obj[k])` thing is a bit verbose and `_.map(obj, (v, k) => v)` is definitely nicer. We could look at doing that later though.

kumar303:
@mstriemer could you take another look? I was thinking that your suggestion about exposing `ErrorList` makes the wrapper more flexible but I don't think that flexibility will be needed for most components. Maybe we can add it later in a future patch if we need it? The downside is that it would require more boilerplate to add the `ErrorList` each time and if someone forgets to add an `ErrorList` then the error messages would disappear which would be confusing.

src/core/api/index.js
```diff
@@ -45,19 +59,29 @@ export function callApi({
   const apiURL = `${API_BASE}/${endpoint}/${queryString}`;
 
   return fetch(apiURL, options)
-    .then((response) => {
+    .then((response) => response.json().then(
```
mstriemer: The second `then` should get dropped down a line. Moving it outside of the `.json()` call would probably be nicer.

```js
fetch()
  .then((response) => response.json())
  .then((jsonResponse) => ..., ...)
  .then(({ response, jsonResponse }) => ...);

kumar303: that won't work because the second then needs to access the `response` variable

kumar303: Anyway, promises are dumb and hard to read. If we switched to async/await it would be a try/catch block and it would be a lot easier to read.

kumar303:
Thanks for all the reviews!

PR #1419, opened 2016-11-24
Fix page loading bug (close #1401)

### LINKED ISSUE #1401: page 2 of search results are broken (passing count: 0)
The second page of search results are broken: https://addons-dev.allizom.org/en-GB/firefox/search/?page=2&q=test

Not sure why it's happening or when it started, but threw me for a loop on a branch when I thought I messed something up, only to see master was broken...

tofumatt:
This should be a simple fix for the second page not loading.

### DIFF
--- a/tests/client/amo/components/TestAddonDetail.js
+++ b/tests/client/amo/components/TestAddonDetail.js
@@ -6,7 +6,10 @@ import {
 } from 'react-addons-test-utils';
 import { Provider } from 'react-redux';
 
-import { AddonDetailBase, allowedDescriptionTags } from 'amo/components/AddonDetail';
+import {
+  AddonDetailBase,
+  allowedDescriptionTags,
+} from 'amo/components/AddonDetail';
 import { OverallRatingWithI18n } from 'amo/components/OverallRating';
 import createStore from 'amo/store';
 import I18nProvider from 'core/i18n/Provider';
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "classnames": "2.2.5",
     "common-tags": "^1.4.0",
     "config": "1.24.0",
+    "deep-eql": "2.0.1",
     "dompurify": "0.8.4",
     "express": "4.14.0",
     "extract-text-webpack-plugin": "1.0.1",
--- a/src/amo/containers/CategoryPage.js
+++ b/src/amo/containers/CategoryPage.js
@@ -1,13 +1,10 @@
+import deepEqual from 'deep-eql';
 import { connect } from 'react-redux';
 import { asyncConnect } from 'redux-connect';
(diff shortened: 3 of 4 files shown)


mstriemer:
The `filtersMatch()` function should check all keys in both `filters` and `otherFilters`. This is essentially `chai.assert.deepEqual()` but returning a boolean instead of throwing. It looks like they use a library called `deep-eql`.

Might be nice to just use `deep-eql` instead. Could be overkill if we just use it once though.

I'm now too deep into this (is that a pun? unintended). Spreading the two sets of props together would fix the function in this case and that should get tests that show the props need to match. That means we could be iterating over each prop twice though that sucks. Putting it into a `Set` would fix that but `Set`s don't have the `every()` method so we'd need to convert it back to an `Array`. You could create a third object and use its keys though (lol, wtf why is this so hard?). Actually this one looks kind of nice (provided there isn't something else I haven't thought of).

```js
const filtersMatch = (filters, otherFilters) =>
  Object.keys({ ...filters, ...otherFilters })
    .every((k) => filters[k] === otherFilters[k]);
```

Just using `deep-eql` may be the best solution, we know it's a good implementation.

mstriemer:
I forgot the link to [`deep-eql`](https://github.com/chaijs/deep-eql/blob/145dd2629e0e1e90e4fabaf0f8b8635ad3015a7a/index.js#L118). Looks like it's just used as:

```js
import deepEqual from 'deep-eql';

const filtersMatch = (x, y) => deepEqual(x, y);

**Comment by Ryuno-Ki** [comment]
In Python, you would actual convert a list to a set and back to a list to make its elements unique …
How high is the penalty if this would be done in JavaScript? That is, how does a transpiled set looks like?

**Comment by tofumatt** [comment]
@mstriemer I'm fine with just using `deep-equal`, we could re-invent the wheel but it has tests already and we know it works well, I'll just use it.

**Comment by tofumatt** [comment]
Okay, switched to using `deep-eql`. r?

File: tests/client/amo/components/TestAddonDetail.js [hunk 1]
```diff
@@ -6,8 +6,9 @@ import {
 } from 'react-addons-test-utils';
 import { Provider } from 'react-redux';
 
-import AddonDetail, { allowedDescriptionTags }
-  from 'amo/components/AddonDetail';
+import AddonDetail, {
+  allowedDescriptionTags,
+} from 'amo/components/AddonDetail';
```
mstriemer: Looks like this could fit on one line. It's also probably why you have a conflict so maybe just leave it as-is.

tofumatt: Ah, indeed, previously it had two imports and needed the spacing? Or something. I'll fix it 👍

PR #1420, opened 2016-11-24
Blur search input on search (fix #1384)

### LINKED ISSUE #1384: Virtual keyboard should be hidden after searching
When you perform a search the keyboard remains visible and the search field remains focused. This is a little annoying as it forces you to manually close the keyboard and makes it less obvious that you have in fact triggered a search.

We should hide the keyboard when they submit the form. Perhaps clearing focus will do this. I'm not really sure.

tofumatt:
Simple fix to hide virtual keyboard (I did the screenshot on the simulator because I broke my Android simulator and I'm too lazy to fix it but it should work).

Before:
nov-24-2016 00-55-59

After:
nov-24-2016 00-48-52

### DIFF
--- a/src/amo/components/SearchForm.js
+++ b/src/amo/components/SearchForm.js
@@ -28,6 +28,7 @@ export class SearchFormBase extends React.Component {
 
   handleSearch = (e) => {
     e.preventDefault();
+    this.searchQuery.input.blur();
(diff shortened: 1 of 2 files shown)


tofumatt:
Done, thanks!

PR #1425, opened 2016-11-25
Translate SearchForm (fix #1362)

### LINKED ISSUE #1362: Search form is missing translatable strings.
`src/amo/components/SearchForm.js` is using a gettext placeholder and not the jed i18n object. 

The gettext placeholder is a no-op (a legacy of the admin search) and as a result strings there are not extracted or translated.

tofumatt:
Had this issue open in my tabs and figured I'd knock it out.

### DIFF
--- a/src/amo/components/SearchForm.js
+++ b/src/amo/components/SearchForm.js
@@ -1,9 +1,10 @@
 import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
+import { compose } from 'redux';
 
 import { loadEntities } from 'core/actions';
 import { fetchAddon } from 'core/api';
-import { gettext as _ } from 'core/utils';
+import translate from 'core/i18n/translate';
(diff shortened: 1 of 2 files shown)


src/amo/components/SearchForm.js
```diff
@@ -69,4 +74,7 @@ export function mapDispatchToProps(dispatch) {
   };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SearchFormBase);
+export default compose(
+  connect(mapStateToProps, mapDispatchToProps),
+  translate({ withRef: true }),
```
mstriemer: I know we seem to use `withRef: true` all over but I'm not sure we actually need it.

tofumatt:
Yeah, same. I just do it out of convention at this point.

Maybe we should file a bug about it.

PR #1429, opened 2016-11-26
Link fallback for installation without mozAddonManager

### LINKED ISSUE #946: Install button fallback for FF < 48 and FFAndroid < 50
Once we have the new install button up and running on the amo pages one consideration is what we need to do for clients that don't have the API available to them.

We should probably feature detect that and fall-back to a pre-48 install route when that happens.

This will need some UX for the button.

mstriemer:
Buttons on http://localhost:3000 and switches on https://discovery.addons.mozilla.org.

install-addon-fallback mov
install-addon-one-click mov
install-theme-fallback mov
install-theme-one-click mov

TODO: Make sure we aren't calling `setCurrentStatus()` anywhere it won't work, or make it a no-op if we don't have `mozAddonManager`. That's probably nicer.

Fixes #946.

### DIFF
--- a/src/core/components/InstallButton/index.js
+++ b/src/core/components/InstallButton/index.js
@@ -1,126 +1,42 @@
 import React, { PropTypes } from 'react';
+import { compose } from 'redux';
 
+import InstallSwitch from 'core/components/InstallSwitch';
+import { THEME_TYPE } from 'core/constants';
 import translate from 'core/i18n/translate';
-import {
-  DOWNLOADING,
-  DISABLED,
-  ENABLED,
-  ENABLING,
-  INSTALLING,
-  INSTALLED,
-  THEME_TYPE,
-  UNINSTALLED,
-  UNINSTALLING,
-  UNKNOWN,
-  validAddonTypes,
-  validInstallStates as validStates,
-} from 'core/constants';
 import { getThemeData } from 'core/themePreview';
-import Switch from 'ui/components/Switch';
+import Button from 'ui/components/Button';
 
 export class InstallButtonBase extends React.Component {
   static propTypes = {
-    downloadProgress: PropTypes.number,
-    enable: PropTypes.func,
-    guid: PropTypes.string.isRequired,
-    handleChange: PropTypes.func,
+    addon: PropTypes.object.isRequired,
+    hasAddonManager: PropTypes.bool.isRequired,
     i18n: PropTypes.object.isRequired,
-    install: PropTypes.func.isRequired,
     installTheme: PropTypes.func.isRequired,
-    installURL: PropTypes.string,
-    name: PropTypes.string.isRequired,
-    slug: PropTypes.string.isRequired,
-    status: PropTypes.oneOf(validStates),
-    type: PropTypes.oneOf(validAddonTypes),
-    uninstall: PropTypes.func.isRequired,
+    status: PropTypes.string.isRequired,
   }
 
-  static defaultProps = {
-    status: UNKNOWN,
-    downloadProgress: 0,
-  }
-
-  getLabel() {
-    const { i18n, name, status } = this.props;
-    let label;
-    switch (status) {
-      case DOWNLOADING:
-        label = i18n.gettext('Downloading %(name)s.');
-        break;
-      case INSTALLING:
-        label = i18n.gettext('Installing %(name)s.');
-        break;
-      case ENABLED:
-      case INSTALLED:
-        label = i18n.gettext('%(name)s is installed and enabled. Click to uninstall.');
-        break;
-      case DISABLED:
-        label = i18n.gettext('%(name)s is disabled. Click to enable.');
-        break;
-      case UNINSTALLING:
-        label = i18n.gettext('Uninstalling %(name)s.');
-        break;
-      case UNINSTALLED:
-        label = i18n.gettext('%(name)s is uninstalled. Click to install.');
-        break;
-      default:
-        label = i18n.gettext('Install state for %(name)s is unknown.');
-        break;
-    }
-    return i18n.sprintf(label, { name });
-  }
-
-  getDownloadProgress() {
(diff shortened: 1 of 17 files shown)


mstriemer:
The diff is fairly large since I moved `InstallButton` into `InstallSwitch` and made `InstallButton` pick `InstallSwitch` or a fallback. I could move the fallback into a component but it's pretty simple so I left it inline.

Expect an update at some point to fix calling `setCurrentStatus()` regardless of whether it will work or not. It doesn't seem to break anything but dumps an error in the console.

mstriemer:
Last build passed other than a linting issue. Should be good for review now.

mstriemer:
Had another linting issue. Fixed now.

src/core/installAddon.js
```diff
@@ -45,12 +47,16 @@ export function makeProgressHandler(dispatch, guid) {
...
-  return {
-    installTheme(node, guid, name, _themeAction = themeAction) {
-      _themeAction(node, THEME_INSTALL);
-      _tracking.sendEvent({ action: 'theme', category: INSTALL_CATEGORY, label: name });
-    },
+export function makeMapStateToProps({ WrappedComponent }) {
+  return function mapStateToProps(state, ownProps, { _tracking = tracking } = {}) {
+    return {
+      WrappedComponent,
+      hasAddonManager: typeof window !== 'undefined' && addonManager.hasAddonManager(),
```
kumar303: both `hasAddonManager` and `installTheme` don't appear to use `state` in any way. Do they need to be in `mapStateToProps`? Could they just be added directly to props in `withInstallHelpers` ? The `mapStateToProps` function is a bit heavy since it's executed on every state change.

src/ui/components/Button/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+export default class Button extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    href: PropTypes.string,
+  }
+
+  render() {
+    const { children, className, href, ...rest } = this.props;
```
kumar303: looks like `href` can be omitted since `...rest` will put it in props but I know sometimes it's nice to declare important props explicitly.

tests/client/core/TestAddonManager.js
```diff
@@ -36,6 +36,16 @@ describe('addonManager', () => {
     };
   });
 
+  describe('hasAddonManager', () => {
+    it('does if mozAddonManager is in navigator', () => {
```
kumar303: this would be less awkward to read for me if it was: `it('returns true if mozAddonManager...')`.

tests/client/core/TestAddonManager.js
```diff
@@ -36,6 +36,16 @@ describe('addonManager', () => {
...
+      assert.ok(addonManager.hasAddonManager({ navigator: { mozAddonManager: {} } }));
+    });
+
+    it('does not if mozAddonManager is not in navigator', () => {
```
kumar303: same here: `it('returns false if mozAddonManager...')`

tests/client/core/TestAddonManager.js
```diff
@@ -36,6 +36,16 @@ describe('addonManager', () => {
...
+      assert.notOk(addonManager.hasAddonManager({}));
```
kumar303: calling this like `addonManager.hasAddonManager()` would prevent someone from accidentally removing the param initializer.

tests/client/core/TestInstallAddon.js
```diff
@@ -40,11 +44,37 @@ describe('withInstallHelpers', () => {
...
     assert.throws(() => {
       withInstallHelpers({})(() => {});
     }, /src is required/);
   });
+
+  it('sets the current status in componentDidMount with an addonManager', () => {
+    const setCurrentStatus = sinon.spy();
+    const WrappedComponent = () => <div />;
+    renderIntoDocument(
+      <WithInstallHelpers WrappedComponent={WrappedComponent} hasAddonManager
```
kumar303: nit: `WrappedComponent={() => <div />}` would make this a bit more concise

tests/client/core/components/TestInstallButton.js
```diff
@@ -1,195 +1,68 @@
...
+    const root = shallowRender(
+      <InstallButtonBase
+        addon={addon} foo="foo" hasAddonManager={false} i18n={i18n}
+        installThemeAction={installThemeAction} />);
+    assert.equal(root.type, Button);
+    assert.deepEqual(root.props, {
+      children: 'Install Theme',
+      'data-browsertheme': JSON.stringify(themePreview.getThemeData(addon)),
+      onClick: installThemeAction,
+    }, sinon.format(root.props));
```
kumar303: using `sinon.format()` is interesting! You reminded me that we need to solve this better though: https://github.com/mozilla/addons-frontend/issues/1447

tests/client/core/TestAddonManager.js
Ryuno-Ki: Plus, the suggestion of @kumar303 has the advantage of not being doubled negated. Double negation tends to be harder to understand.

src/core/components/InstallButton/index.js
```diff
@@ -1,126 +1,46 @@
...
 }
 
-export default translate()(InstallButtonBase);
+export function installTheme(event) {
+  event.preventDefault();
+  themeAction(event.currentTarget, THEME_INSTALL);
+}
+
+export const setInstallThemeAction = (Component) => (props) => (
+  <Component {...props} installThemeAction={installTheme} />
```
mstriemer: This felt weird when I did it. I'm not sure why I did this. That seems reasonable.

src/ui/components/Button/index.js
mstriemer: Since it gets used in the `if` I thought it would be good to pull it out explicitly.

PR #1438, opened 2016-11-29
Extensions page 1286

tofumatt:
Blah, this became quite big with all the tests, but it was a lot to implement to get the whole page together. I think I did the right thing creating actions/reducers for each one.

This fixes #1285 and fixes #1286.

### Screenshots

screen shot 2016-11-29 at 03 12 03

screen shot 2016-11-29 at 03 14 15

screen shot 2016-11-29 at 03 14 17

-----

r? to whoever feels like it. Probably no one. Sorry this is massive.

PR #1439, opened 2016-11-29
Add actions and reducers for #1285 and #1286

tofumatt:
This is half the size of the other PR (#1438)–it implements just the actions/reducers needed to deal with featured, highly rated, and popular add-ons (which will be on the extensions and themes homepage).


tofumatt:
I'll make another PR (or rebase #1438 onto this) once this is ready.

src/amo/actions/featured.js
```diff
@@ -0,0 +1,27 @@
+import {
+  FEATURED_GET,
+  FEATURED_LOADED,
+  FEATURED_FAILED,
+} from 'core/constants';
+
+
+export function featuredGet({ addonType }) {
```
kumar303: I find it makes the most sense to read action functions as verbs. Perhaps `getFeatured` ?

src/amo/actions/featured.js
```diff
@@ -0,0 +1,27 @@
...
+  return {
+    type: FEATURED_GET,
+    payload: { addonType },
+  };
+}
+
+export function featuredLoad({ entities, result, addonType }) {
```
kumar303: `featuredLoaded` ?

src/amo/actions/featured.js
```diff
@@ -0,0 +1,27 @@
...
+  return {
+    type: FEATURED_LOADED,
+    payload: { entities, result, addonType },
+  };
+}
+
+export function featuredFail({ addonType }) {
```
kumar303: `featuredFailed`?

src/amo/reducers/highlyRated.js
```diff
@@ -0,0 +1,33 @@
+import {
+  HIGHLY_RATED_GET,
+  HIGHLY_RATED_LOADED,
+  HIGHLY_RATED_FAILED,
+} from 'core/constants';
+
+const initialState = {
+  count: 0,
+  filters: {},
```
kumar303: why do you need to track `filters` in state for `highlyRated` and `popular` add-on data? It looks like (in your other patch) that the filters are only used to make the API call. The API call populates the data in state so why do you need to save the filters? You should already have the filtered results.

src/amo/reducers/featured.js
```diff
@@ -0,0 +1,33 @@
+import {
+  FEATURED_GET,
+  FEATURED_LOADED,
+  FEATURED_FAILED,
+} from 'core/constants';
+
+const initialState = {
+  count: 0,
+  loading: false,
+  results: [],
```
kumar303: I see three separate reducers: `featured`, `highlyRated`, and `popular` but they all look nearly identical. I also see a lot of repetition in all the code and tests that cover their logic. What if you combined them all into one reducer and just tracked the "buckets" of add-on results in separate keys? more like:

````javascript
const initialState = {
  featured: {
    count: 0,
    loading: false,
    results: [],
  },
  highlyRated: {
    count: 0,
    loading: false,
    results: [],
  },
  popular: {
    count: 0,
    loading: false,
    results: [],
  },
};
````

Comment (tofumatt) [inline src/amo/reducers/featured.js]: That actually works for me, but I worried about too tightly coupling things that could maybe be used separately. I think you're right though, and one reducer could be used. I think I got a bit tripped up using `Promise.all()` and assigning things to the right state but maybe I need to give it another go.

I'll try combining them all and making the tweaks suggested and see how it goes. Thanks 👍

File: src/amo/actions/featured.js (hunk 1 above)
Comment (tofumatt) [inline src/amo/actions/featured.js]: I agree but we haven't been doing that in the past. I'm okay moving toward it; it makes them harder to scan through in a file (here they all stand out as `featured[ACTION]` but I'll move to it here and file an issue to consolidate them.

**Comment by tofumatt** [comment]
You're right about combing them as this is now about a third of the code it was before. I'll need to do a bit of extra to merge all three responses but this is worth it. 👍

**Comment by tofumatt** [comment]
This is ready for another r? btw

File: src/amo/reducers/landing.js [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+  loading: false,
+  popular: { count: 0, results: [] },
+};
+
+export default function landing(state = initialState, action) {
+  const { payload } = action;
+  let newState;
+  switch (action.type) {
+    case LANDING_GET:
+      return { ...state, ...payload, loading: true };
```
kumar303: I would rather see `addonType: payload.addonType` instead of `...payload` because it's not super clear otherwise. Should `initialState` have a default value for `addonType`? Maybe that would be more clear.

src/amo/actions/landing.js
```diff
@@ -0,0 +1,29 @@
...
+
+export function getLanding({ addonType }) {
+  return {
+    type: LANDING_GET,
+    payload: { addonType },
+  };
+}
+
+export function loadLanding(
+  { addonType, entities, featured, highlyRated, popular }
```
kumar303: The reducer is going to throw reference errors if any one of `featured`, `highlyRated`, or `popular` is missing. If I'm correct about that, you should raise an error here if any one of those payload keys is falsey.

src/amo/reducers/landing.js
```diff
@@ -0,0 +1,36 @@
...
+export const initialState = {
+  featured: { count: 0, results: [] },
+  highlyRated: { count: 0, results: [] },
+  loading: false,
+  popular: { count: 0, results: [] },
+};
+
+export default function landing(state = initialState, action) {
+  const { payload } = action;
+  let newState;
```
kumar303: I find that it's a little cleaner to only define this in the exact switch case that you need it for. You might have tried that and got a lint error because you have to use brackets to declare a new scope:
````javascript
switch (action.type) {
  case LANDING_LOADED: {
    const newState = { ...state, loading: false };
    newState[key] = ...
    return newState;
  } 
}
````

File: src/amo/reducers/landing.js [hunk 9]
```diff
@@ -0,0 +1,36 @@
...
+      ['featured', 'highlyRated', 'popular'].forEach((key) => {
+        newState[key] = {
+          count: payload[key].result.count,
+          results: payload[key].result.results.map((slug) => payload.entities.addons[slug]),
+        };
+      });
+
+      return newState;
+    case LANDING_FAILED:
+      return initialState;
```
kumar303: This would not preserve the previously declared `addonType`. Is that important?

src/amo/reducers/landing.js
```diff
@@ -0,0 +1,36 @@
+import {
+  LANDING_GET,
+  LANDING_LOADED,
+  LANDING_FAILED,
+} from 'core/constants';
+
+
+export const initialState = {
```
kumar303: is `entities` used? I see that some actions are setting it and some tests are checking for it. If so, a default value (even if it's `null`) would help me understand the data structure better. Is `entities` really necessary though if we already have `results` arrays?

kumar303:
Just a thought: actions are the interface into data storage. I feel like all reducer tests should be executed with actions. Thus, a reducer test shouldn't ever need to begin with raw state objects. I used this pattern here and it felt right: https://github.com/mozilla/addons-frontend/pull/1418/files#diff-c696d213dc19550ac8bc5c0fac172451R21

Maybe this is just something we should think about more. Just thinking out loud ;)

src/amo/reducers/landing.js
tofumatt: We use `entities` in the `LOADED` reducer to create our `results` object, yeah. But we don't use it/access it outside the reducer/in the components.

src/amo/reducers/landing.js
tofumatt: Ah, neat. Yeah I was getting scope errors and didn't know you could force a new scope like that 👍

src/amo/reducers/landing.js
tofumatt: I don't think there should be a default value set but I've set one to `null` so the data structure is clearer. And I switched to `addonType: payload.addonType`; you're right the explicitness is better 👍

src/amo/reducers/landing.js
tofumatt: I don't know if it would be used but it should; I've changed it.

src/amo/reducers/landing.js
kumar303: aha, I see it now on this line: https://github.com/mozilla/addons-frontend/pull/1439/files#diff-cba04c1c93442d633523fc29d40f97e0R26

src/amo/reducers/landing.js
```diff
@@ -0,0 +1,36 @@
...
+  let newState;
+  switch (action.type) {
+    case LANDING_GET:
+      return { ...state, ...payload, loading: true };
+    case LANDING_LOADED:
+      newState = { ...state, loading: false };
+      ['featured', 'highlyRated', 'popular'].forEach((key) => {
+        newState[key] = {
+          count: payload[key].result.count,
+          results: payload[key].result.results.map((slug) => payload.entities.addons[slug]),
```
kumar303: shouldn't `payload.entities` be `payload[key].entities` ? Otherwise results for other keys will get erased.

tofumatt: Indeed! This was a bug when I separated out the reducers, it was `payload[key].entities` originally. Thanks!

tofumatt:
Okay, I think I've addressed all the comments. Ready for another r?

src/amo/actions/landing.js
kumar303: Now that I understand this better, I think you can remove `entities`, right?

src/amo/actions/landing.js
```diff
@@ -0,0 +1,29 @@
+import {
+  LANDING_GET,
+  LANDING_LOADED,
+  LANDING_FAILED,
+} from 'core/constants';
+
+
+export function getLanding({ addonType }) {
```
kumar303: I think it would be nice to throw an error here if `addonType` is falsey

src/amo/actions/landing.js
kumar303: It would be nice to do some payload checking here:
* make sure `addonType` is set
* make sure at least one of the `featured`, `highlyRated`, or `popular` keys are set
* check for `result`, and `entities` within a set key

I don't know if that's overkill but it seems like it might help prevent typo'd calls.

I want to land this soon: https://github.com/mozilla/addons-frontend/issues/1461

tests/client/amo/actions/test_landing.js
```diff
@@ -0,0 +1,46 @@
...
+
+describe('LANDING_LOADED', () => {
+  const entities = sinon.stub();
+  const response = {
+    featured: sinon.stub(),
+    highlyRated: sinon.stub(),
+    popular: sinon.stub(),
+  };
+  const action = actions.loadLanding({
+    addonType: 'theme', entities, ...response });
```
kumar303: I don't think `entities` is needed here

tests/client/amo/actions/test_landing.js
```diff
@@ -0,0 +1,46 @@
...
+
+  it('sets the type', () => {
+    assert.equal(action.type, 'LANDING_LOADED');
+  });
+
+  it('sets the payload', () => {
+    assert.deepEqual(action.payload, {
+      addonType: 'theme', entities, ...response });
```
kumar303: again, `entities` can be removed.

This is an example of where I think you'll get better tests by following a pattern like: set data using an action -> make assertions about how the *reducer stored the data*. Instead, what you're doing is making assertions about the action payload which isn't so helpful.

tests/client/amo/reducers/testLanding.js
```diff
@@ -0,0 +1,131 @@
...
+        featured: { foo: 'bar' },
+        highlyRated: { count: 0 },
+        loading: false,
+        popular: { results: [] },
+      };
+      const {
+        addonType, featured, highlyRated, loading, popular,
+      } = landing(initialState, {
+        type: 'LANDING_GET',
+        payload: { addonType: 'theme' },
```
kumar303: I think this would be a better test if you passed `actions.loadLanding()` directly to `landing()`.

tests/client/amo/reducers/testLanding.js
```diff
@@ -0,0 +1,131 @@
...
+      const entities = {
+        addons: {
+          bar: { slug: 'bar' },
+          foo: { slug: 'foo' },
+          food: { slug: 'food' },
+        },
+      };
+      const { featured, highlyRated, popular } = landing(initialData, {
+        type: 'LANDING_LOADED',
+        payload: {
```
kumar303: again, I think passing `actions.loadActions()` directly would make this a more robust test

src/amo/actions/landing.js
tofumatt: I'd like to land #1461 but I think it's overkill for this particular method, especially as we don't tend to do it elsewhere IIRC.

PR #1440, opened 2016-11-29
update yarn lock file after greenkeeper updates

kumar303:


mstriemer:
I originally got a different diff but after updating yarn to the latest version I get the same one. Guess the format just changed a little.

kumar303:
Ah, ok! That makes more sense. Thanks for checking. I pulled from master, updated via yarn, and this time the lock was more sane; it showed just the packages that got bumped.

PR #1442, opened 2016-11-30
Theme preview

### LINKED ISSUE #1422: Theme preview on detail page
When holding down on the theme image on the detail page we should provide a theme preview.

> I like "press" instead of "tap". Tapping is just like flicking the switch. Pressing is a deliberate action: press down and you preview, release and you go back to what it was before. Made the label/icon a bit smaller:
> screen shot 2016-11-18 at 1 54 57 pm

> Oh, nix that "icon only" view.

Extracted from #946.

mstriemer:
I'm going to look into fixing the dimensions of the image in the morning which should make the label look better.

screen shot 2016-11-30 at 14 48 13

theme-preview-finalish mp4

> [With label (missing icon) (old)](https://cloud.githubusercontent.com/assets/211578/20743427/e9853634-b69b-11e6-8da7-6f93942d13a0.png)
>[Previewing (and sometimes scrolling) (old)](https://cloud.githubusercontent.com/assets/211578/20743367/8c9b6ace-b69b-11e6-8e12-864866fc6414.gif) 

Fixes #1422.

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -4,11 +4,12 @@ import { compose } from 'redux';
 
 import AddonMeta from 'amo/components/AddonMeta';
 import AddonMoreInfo from 'amo/components/AddonMoreInfo';
-import InstallButton from 'core/components/InstallButton';
 import DefaultOverallRating from 'amo/components/OverallRating';
 import ScreenShots from 'amo/components/ScreenShots';
 import 'amo/css/AddonDetail.scss';
 import fallbackIcon from 'amo/img/icons/default-64.png';
+import InstallButton from 'core/components/InstallButton';
+import { THEME_TYPE } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import { isAllowedOrigin, nl2br, sanitizeHTML } from 'core/utils';
 import translate from 'core/i18n/translate';
@@ -35,7 +36,10 @@ export class AddonDetailBase extends React.Component {
   static propTypes = {
     OverallRating: PropTypes.element,
     addon: PropTypes.object.isRequired,
+    getBrowserThemeData: PropTypes.func.isRequired,
     i18n: PropTypes.object.isRequired,
+    previewTheme: PropTypes.func.isRequired,
+    resetPreviewTheme: PropTypes.func.isRequired,
     setCurrentStatus: PropTypes.func.isRequired,
   }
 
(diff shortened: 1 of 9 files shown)


mstriemer:
I updated the image so it is now positioned on the top right since I believe that's usually where the most interesting part of a theme is located. Aspect ratio is maintained using `object-fit: cover` so it will crop the left hand side until it is as wide as the source image then will start cropping the bottom and scale up the image as it grows beyond that (the latter shouldn't happen until desktop).

Should be good for review now. I'm going to file an issue for the icon and to figure out how to wait on the preview until it looks like the user is pressing not scrolling (this will likely add a 300ms or so delay but will be less annoying if you're just trying to scroll).

src/amo/components/AddonDetail.js
```diff
@@ -45,6 +49,50 @@ export class AddonDetailBase extends React.Component {
...
+      const label = i18n.gettext('Press to preview');
+      return (
+        <div
+          className="AddonDetail-theme-header"
+          id="AddonDetail-theme-header"
+          data-browsertheme={getBrowserThemeData()}
+          onTouchStart={this.onTouchStart}
+          onTouchEnd={this.onTouchEnd}
+        >
+          <label className="AddonDetail-theme-header-label" htmlFor="AddonDetail-theme-header">
```
tofumatt: Nitpick, but looks like this is over 80 lines.

src/amo/css/AddonDetail.scss
```diff
@@ -1,16 +1,12 @@
...
-    background: $masthead-color;
-    color: $header-font-color;
-  }
-
   hr {
     margin: 0 20px;
   }
 
-  .icon,
+  .AddonDetail-icon,
```
tofumatt: I don't think this needs to be nested (anymore).

src/amo/css/AddonDetail.scss
```diff
@@ -66,3 +39,58 @@
...
+.AddonDetail-theme-header-label {
+  $label-margin: 12px;
+  @include start($theme-header-horizontal-padding + $label-margin);
+
+  background: rgba(255, 255, 255, 0.5);
+  border-radius: 8px;
+  bottom: $label-margin;
+  color: #fff;
+  font-size: 12px;
+  line-height: 12px;
```
tofumatt: We usually use unitless line height values.

src/amo/css/AddonDetail.scss
mstriemer: I think the other rules still need to be nested so I'll leave this here until those are brought up to date with our more specific class names.

src/amo/components/AddonDetail.js
mstriemer: It's surprisingly only 79 characters wide, but I believe we have a lint rule to prevent 100+.

src/amo/css/AddonDetail.scss
tofumatt: Oh, true. 👍

PR #1448, opened 2016-12-02
Truncate and style description (fix #1296)

### LINKED ISSUE #1296: Update the "about this add-on" block and add truncation.
This data already exists and is represented on the page. Fixing this should only require:
- Truncate the display of content and show a read more link. Any truncation will need to take care to not break the HTML.
- Update the styles.

Here's a snippet from the latest sketch file:

amo_-_v1_sketch

tofumatt:
Screenshot:
ezgif com-383fe5fed6

-----

This moves some things from the `AddonDetail` component into an `AddonDescription` component (which is nice for smaller components used to build bigger ones) so it looks like more than it actually is as I moved the tests about too.

It also shows the description in the new style and deals with truncation using size rather than text content (anything too tall is rendered with a "show more" link that just sets a new class). Stuff that is short enough just shows with no truncation.

### DIFF
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -1,0 +1,66 @@
+import classNames from 'classnames';
+import React, { PropTypes } from 'react';
+import ReactDOM from 'react-dom';
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+import Card from 'ui/components/Card';
+
+import './ShowMoreCard.scss';
+
+
+export class ShowMoreCardBase extends React.Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    header: PropTypes.node,
+    footer: PropTypes.node,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
+  }
+
+  componentDidMount() {
+    this.expandIfDescriptionIsShortEnough();
+  }
+
+  expandIfDescriptionIsShortEnough = () => {
+    // If the add-on description is short enough it doesn't need a "show more"
+    // link, we'll expand the description by default.
+    if (ReactDOM.findDOMNode(this.contents).clientHeight < 100) {
+      this.setState({ expanded: true });
+    }
+  }
+
+  expandText = (event) => {
+    event.preventDefault();
+    this.setState({ expanded: true });
+  }
+
+  render() {
+    const { children, className, header, footer, i18n } = this.props;
+    const { expanded } = this.state;
+
+    return (
+      <Card className={classNames('ShowMoreCard', className, {
+        'ShowMoreCard--expanded': expanded,
+      })} header={header} footer={footer}>
+        <div className="ShowMoreCard-contents"
+          ref={(ref) => { this.contents = ref; }}>
+          {children}
+        </div>
+        <a className="ShowMoreCard-revealMoreLink" href="#show-more"
+          onClick={this.expandText}>
+          {i18n.gettext('Show more info…')}
+        </a>
+      </Card>
+    );
+  }
+}
+
+export default compose(
+  translate({ withRef: true }),
+)(ShowMoreCardBase);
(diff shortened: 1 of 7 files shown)


src/amo/components/AddonDescription.js
```diff
@@ -0,0 +1,86 @@
...
+    return (
+      <Card className={classNames('AddonDescription', {
+        'AddonDescription--expanded': expanded,
+      })} header={header}>
+        <div className="AddonDescription-contents"
+          ref={(ref) => { this.contents = ref; }}
+          dangerouslySetInnerHTML={
+            sanitizeHTML(nl2br(addon.description), allowedDescriptionTags)
+          } />
+        <a className="AddonDescription-revealMoreLink" href="#more"
```
mstriemer: We normally use `kebab-case` other than the component name from what I've seen.

src/amo/components/AddonDescription.js
```diff
@@ -0,0 +1,86 @@
...
+          onClick={this.expandText}>
+          {i18n.gettext('Show more info…')}
```
mstriemer: I feel like "Show more" might be better text since there's a section called "More information".

src/amo/components/AddonDescription.js
```diff
@@ -0,0 +1,86 @@
...
+  'code',
+  'em',
+  'i',
+  'li',
+  'ol',
+  'strong',
+  'ul',
+];
+
+export class AddonDescriptionBase extends React.Component {
```
mstriemer: It looks like this component doesn't have anything specific to an add-on's description other than it using `addon.description`. You could use `children` instead and make this reusable.

```jsx
// Leaving `maxheight` out might be good in case we want to
// change it at different screen sizes.
<CollapsableCard maxheight={100} header={`About this ${addon.type}`}>
  <div dangerouslySetInnerHTML={prepareDescription(addon.description} />
</CollapsableCard>
```

It might be good to make the default be expanded and then collapse it in `componentDidMount()` if it is too tall. That way if JS is disabled you can still read the whole description.

tests/client/amo/components/TestAddonDescription.js
```diff
@@ -0,0 +1,144 @@
...
+import { fakeAddon } from 'tests/client/amo/helpers';
+import { getFakeI18nInst } from 'tests/client/helpers';
+
+
+function render(props) {
+  const i18n = getFakeI18nInst();
+  const initialState = { api: { clientApp: 'android', lang: 'pt' } };
+
+  return findRenderedComponentWithType(renderIntoDocument(
+    <Provider store={createStore(initialState)}>
```
mstriemer: Is the provider and store needed? I don't see anything using redux state.

tests/client/amo/components/TestAddonDescription.js
```diff
@@ -0,0 +1,144 @@
...
+        description:
+          '<a href="javascript:alert(document.cookie)" onclick="sneaky()">placeholder</a>',
+      },
+    });
+    const anchor = rootNode.querySelector('a');
+    assert.equal(anchor.attributes.onclick, null);
+    assert.equal(anchor.attributes.href, null);
+  });
+//
+//   it('renders an amo CDN icon image', () => {
```
mstriemer: 👻

tofumatt:
Ah, good call given it's a "show more" card effectively. Cool, I'll separate it out and get rid of some of the extra code; the tests were copied right from the AddonDetail one and I missed a few bits. Cheers!

src/ui/components/ShowMoreCard/index.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
+  }
+
+  componentDidMount() {
+    this.expandIfDescriptionIsShortEnough();
+  }
+
+  expandIfDescriptionIsShortEnough = () => {
```
mstriemer: This name should likely change.

src/ui/components/ShowMoreCard/index.js
```diff
@@ -0,0 +1,66 @@
...
+    children: PropTypes.node,
+    className: PropTypes.string,
+    header: PropTypes.node,
+    footer: PropTypes.node,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
```
mstriemer: Would be nice to default this to `expanded: true` and close it with JS.

src/ui/components/ShowMoreCard/index.js
tofumatt: Oh shoot, good point.

Ryuno-Ki:
It looks like the "show more info" link would be the only one on that page. Otherwise I'd like to remind you on https://github.com/mozilla/addons-frontend/issues/1296#issuecomment-257966681

PR #1453, opened 2016-12-02
Fix comments in #1448

tofumatt:
A few tweaks mentioned after I merged, sorry!

### DIFF
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -1,13 +1,18 @@
+/* eslint-disable react/no-danger */
+
 import classNames from 'classnames';
 import React, { PropTypes } from 'react';
 import ReactDOM from 'react-dom';
 import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
+import { sanitizeHTML } from 'core/utils';
 import Card from 'ui/components/Card';
 
 import './ShowMoreCard.scss';
 
+
+const MAX_HEIGHT = 100;
 
 export class ShowMoreCardBase extends React.Component {
   static propTypes = {
@@ -20,24 +25,24 @@ export class ShowMoreCardBase extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { expanded: false };
+    this.state = { expanded: true };
   }
 
   componentDidMount() {
-    this.expandIfDescriptionIsShortEnough();
-  }
-
-  expandIfDescriptionIsShortEnough = () => {
-    // If the add-on description is short enough it doesn't need a "show more"
-    // link, we'll expand the description by default.
-    if (ReactDOM.findDOMNode(this.contents).clientHeight < 100) {
-      this.setState({ expanded: true });
-    }
+    this.truncateToMaxHeight(ReactDOM.findDOMNode(this.contents));
   }
 
   expandText = (event) => {
     event.preventDefault();
     this.setState({ expanded: true });
+  }
+
+  truncateToMaxHeight = (contents) => {
+    // If the contents are short enough they don't need a "show more" link; the
+    // contents are expanded by default.
+    if (contents.clientHeight > MAX_HEIGHT) {
+      this.setState({ expanded: false });
+    }
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


src/ui/components/ShowMoreCard/index.js
```diff
@@ -55,6 +60,10 @@ export class ShowMoreCardBase extends React.Component {
         <a className="ShowMoreCard-revealMoreLink" href="#show-more"
           onClick={this.expandText}>
           {i18n.gettext('Show more info…')}
```
mstriemer: Can this text be "Read more"? I find it confusing that there is a block below called "More information" and this says "Show more info...".

src/ui/components/ShowMoreCard/index.js
```diff
@@ -55,6 +60,10 @@ export class ShowMoreCardBase extends React.Component {
...
+
+          <span className="visually-hidden">
+            {i18n.gettext("(show more of this add-on's description)")}
```
mstriemer: Ideally this wouldn't reference add-on description at all.

src/ui/components/ShowMoreCard/index.js
mstriemer: It looks like Facebook has a "Continue Reading" link on long posts in their Android app. I'm not sure if there is any extra context provided on Android but on the desktop web it just shows "Read More".

Not saying we should necessarily do what they are but it could be that these sort of links are commonplace and adding more context might not make sense. I could see if you had something like "**Read more** of Alice's article" that would be bad.

I'm not sure what the solution is but this feels too verbose and "Read More" alone does feel like without context you couldn't tell if the site was not screen reader friendly or if it was one of these "expand section" links.

`<span className="visually-hidden">Expand to</span> Read More`?

src/ui/components/ShowMoreCard/index.js
```diff
@@ -16,30 +16,35 @@ export class ShowMoreCardBase extends React.Component {
     header: PropTypes.node,
     footer: PropTypes.node,
     i18n: PropTypes.object.isRequired,
+    maxHeight: PropTypes.number.isRequired,
+  }
+
+  static defaultProps = {
+    maxHeight: 100,
```
mstriemer: I'm fine with passing this in but since we don't know if we'll always want this I'd say it's probably best to leave it out for now.

My reasoning is that at some point we might want to have "sizes" rather than "heights". So on mobile we might want this to be 100px but maybe on desktop we want 200px but we want that to be the "small" card.

src/ui/components/ShowMoreCard/index.js
Ryuno-Ki: Yes, adding content for screenreader-only should be sufficient. It needs to be a children of the same `<a>` tag to get read out.

I've found a [chrome extension about accessibility developer tools](https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb) recently, but haven't give it a try. The idea is to be able to inspect the part of the DOM relevant for screenreaders.

src/ui/components/ShowMoreCard/index.js
tofumatt: Alright, less code for now!

tofumatt:
Okay, fixed up the issues, ready for another r?

src/ui/components/ShowMoreCard/index.js
```diff
@@ -54,7 +56,11 @@ export class ShowMoreCardBase extends React.Component {
         </div>
         <a className="ShowMoreCard-revealMoreLink" href="#show-more"
           onClick={this.expandText}>
-          {i18n.gettext('Show more info…')}
+          <span className="visually-hidden">
```
mstriemer: This might not translate well if we want to flip the two strings in a certain language. It might need to be `i18n.gettext('<span class="visually-hidden">Expand to</span> Read More')`. Putting HTML in our translations is definitely not ideal though.

There should at least be an l10n comment explaining how these strings are used.

It would be nice if we could get the text formatted somehow but that could be confusing. Something like this, but actually reasonable:

```jsx
const { label, text } = magicallyGetI18nValues(
  i18n.gettext('label(Expand to) text(Read More)'));
// ...
  <span className="visually-hidden">{label.trim()}</span>{text.trim()}

File: src/ui/components/ShowMoreCard/index.js [hunk 5]
```diff
@@ -54,7 +56,11 @@ export class ShowMoreCardBase extends React.Component {
...
+            {i18n.gettext('Expand to')}
+          </span>
+
+          {i18n.gettext('Read more…')}
```
mstriemer: I don't see an ellipsis in the mocks and I'm not a huge fan of them. I think "Read more" can stand on its own.

tofumatt:
@mstriemer If you're online and want to just double-check this–I had to use `dangerouslySetInnerHTML` to get the `i18n` `<span>` to show, but I wrapped it in `sanitizeHTML` so even if bad localised data got through it should be fine.

mstriemer:
Good call sanitizing it. That looks perfect.

PR #1463, opened 2016-12-07
Build locales 2016 12 07

muffinresearch:


mstriemer:
A bit hard to verify but I don't see anything standing out that looks wrong. r+

PR #1464, opened 2016-12-07
Refactor homepage cat CSS and reduce nesting

muffinresearch:
Before:

modules_pour_firefox

After:

modules_pour_firefox

### DIFF
--- a/src/amo/css/Home.scss
+++ b/src/amo/css/Home.scss
@@ -8,7 +8,7 @@
   text-align: center;
 }
 
-.HomePage-cat-list {
+.HomePage-category-list {
   display: flex;
   flex-wrap: wrap;
   overflow: auto;
@@ -16,8 +16,8 @@
   width: 100%;
 }
 
-.HomePage-cat-li {
-  background-color: #fff;
+.HomePage-category-li {
+  background-color: $base-color;
   background-size: 50% auto;
   background-position: 50% 35%;
   background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


src/amo/css/Home.scss
```diff
@@ -14,45 +14,34 @@
...
-    text-align: center;
-    width: 30%;
-
-    &::before {
-      content: '';
-      display: block;
-      padding-bottom: 100%;
-    }
-  }
+.HomePage-cat-li {
```
tofumatt: I know the extra characters make the HTML longer but can we make this `.HomePage-category-li`? The abbreviations always trip me up and I'd like to discourage their use 😄

src/amo/css/Home.scss
```diff
@@ -14,45 +14,34 @@
...
+  background-color: #fff;
```
tofumatt: We should use `$base-color` from `~core/css/inc/vars` here instead, just in case we ever want to change these colours.

src/amo/css/Home.scss
```diff
@@ -103,49 +92,50 @@
 }
 
 .HomePage-block-ads {
-  background: url('../img/home/block-ads.svg') no-repeat 50% 35%;
```
tofumatt: Very nice 👍

src/amo/css/Home.scss
muffinresearch: Good catch!

PR #1465, opened 2016-12-08
Include eye icon in theme preview label

### LINKED ISSUE #1444: Include eye icon on theme image label
The mocks have an eye icon in the label on the theme image. Once we have that asset we should add it to `Icon` and drop it in the label.

> screen shot 2016-11-18 at 1 54 57 pm

Extracted from #1422 and #1442.

mstriemer:
screen shot 2016-12-08 at 11 10 09

Fixes #1444.


pwalm:
Can we make the button a wee bit shorter? Other than that, looks great :D

mstriemer:
Sat down with Phil and he said the height was okay (it's too skinny when it's shorter) but he wanted more padding between the text.

mstriemer:
I updated the screenshot with the new padding.

mstriemer:
Oh wow, I didn't notice that it has no tests at all. Looks like I'm depending on coverage more than I should.

I filed https://github.com/mozilla/addons-frontend/issues/1491 to switch the component to export something like `<EyeIcon />` so that we don't use icons that aren't setup.

kumar303:
thanks, looks good. r+

PR #1469, opened 2016-12-09
Add language picker (fix #1289)

### LINKED ISSUE #1289: Hook up locales to the language drop-down
The language chooser doesn't currently have any data in it.

add-ons_for_firefox

tofumatt:
Just grabbed the `language.json` from product details and tossed it in here rather than dealing with importing it and running it and all that jazz. That should be fine (as discussed with @kumar303) and we can always add to it if we add locales outside of what it has in it in the future.

dec-08-2016 17-06-46

### DIFF
--- a/src/amo/components/LanguagePicker.js
+++ b/src/amo/components/LanguagePicker.js
@@ -1,33 +1,69 @@
+/* global window */
 import React, { PropTypes } from 'react';
 import { compose } from 'redux';
+import { connect } from 'react-redux';
 
+import languages from 'core/languages';
 import translate from 'core/i18n/translate';
+import { addQueryParams } from 'core/utils';
 
 import './LanguagePicker.scss';
 
 
+export function changeLocaleURL({ currentLocale, location, newLocale }) {
+  const newPath = location.pathname.replace(new RegExp(`^/${currentLocale}/`),
+    `/${newLocale}/`);
+  return addQueryParams(newPath, location.query);
+}
+
 export class LanguagePickerBase extends React.Component {
   static propTypes = {
+    currentLocale: PropTypes.string.isRequired,
     i18n: PropTypes.object.isRequired,
+    location: PropTypes.object.isRequired,
+    _window: PropTypes.object,
+  }
+
+  onChange = (event) => {
+    event.preventDefault();
+    this.changeLanguage(event.target.value);
+  }
+
+  changeLanguage(newLocale) {
+    const { currentLocale, location, _window } = this.props;
+
+    if (currentLocale !== newLocale) {
+      const newURL = changeLocaleURL({ currentLocale, location, newLocale });
+      // We change location because a locale change requires a full page
+      // reload to get the new translations, etc.
+      (_window || window).location = newURL;
+    }
   }
 
   render() {
-    const { i18n } = this.props;
+    const { currentLocale, i18n } = this.props;
 
     return (
       <div className="LanguagePicker">
         <h3 className="LanguagePicker-header">
           {i18n.gettext('Browse in your language')}
         </h3>
-        <select className="LanguagePicker-selector"
-          ref={(ref) => { this.selector = ref; }}>
-          <option value="en-US">English (US)</option>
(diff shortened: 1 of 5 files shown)


src/amo/components/LanguagePicker.js
```diff
@@ -1,14 +1,59 @@
...
+      (_window || window).location = newURL;
+    }
+  }
+
+  languages() {
+    const { currentLocale } = this.props;
+
+    return config.get('langs').map((locale) => {
+      if (languageJSON[locale]) {
+        const isCurrentLocale = currentLocale === locale ? 'selected' : '';
```
mstriemer: You should be able to set `value={currentLocale}` on the select and react will do this for you. See https://facebook.github.io/react/docs/forms.html#the-select-tag.

src/amo/components/LanguagePicker.js
```diff
@@ -1,14 +1,59 @@
+/* global window */
+import config from 'config';
 import React, { PropTypes } from 'react';
 import { compose } from 'redux';
+import { connect } from 'react-redux';
 
+import { languageJSON } from 'core/languages';
```
mstriemer: I doubt we're going to add to that file so this could probably be `import languages from 'core/languages';`.

src/amo/components/LanguagePicker.js
```diff
@@ -1,14 +1,59 @@
...
+      // We change location because a locale change requires a full page
+      // reload to get the new translations, etc.
+      (_window || window).location = newURL;
+    }
+  }
+
+  languages() {
+    const { currentLocale } = this.props;
+
+    return config.get('langs').map((locale) => {
```
mstriemer: Could this check get moved into `core/languages`? I feel it could filter the full list based on config before it exports it for convenience.

src/amo/components/LanguagePicker.js
```diff
@@ -1,14 +1,59 @@
...
+      if (languageJSON[locale]) {
+        const isCurrentLocale = currentLocale === locale ? 'selected' : '';
+        return (
+          <option value={locale} selected={isCurrentLocale}>
+            {languageJSON[locale].native}
+          </option>
+        );
+      }
+
+      return '';
```
mstriemer: You can return `null` to render nothing for this lang. React might be putting `<!-- react-text: 115 --><!-- /react-text: 115 -->` for these empty strings.

src/amo/components/LanguagePicker.js
```diff
@@ -1,14 +1,59 @@
...
 
+import { languageJSON } from 'core/languages';
 import translate from 'core/i18n/translate';
+import { addQueryParams } from 'core/utils';
 
 import './LanguagePicker.scss';
 
 
+export function changeLocaleURL({ currentLocale, location, newLocale }) {
+  const newPath = location.pathname.replace(`/${currentLocale}/`,
```
tofumatt: Unfortunately `split` doesn't return the remaining contents as one long string but instead just stops adding to the array it makes:

```node
> '/en-US/firefox/nowhere/'.split('/')
[ '', 'en-US', 'firefox', 'nowhere', '' ]
> '/en-US/firefox/nowhere/'.split('/', 2)
[ '', 'en-US' ]
> '/en-US/firefox/nowhere/'.split('/', 3)
[ '', 'en-US', 'firefox' ]
> '/en-US/firefox/nowhere/'.split('/', 1)
[ '' ]
> '/en-US/firefox/nowhere/'.split('/', 2)
[ '', 'en-US' ]
```

So using `split` would be more code and end up doing the same thing, especially as I realised we should only be matching the first `/locale/` in the URL (in the off chance an addon or other URL had `/locale/` in it we don't want to replace it).

So I've improved the Regex to only match the first instance and I think it's fine; the alternate with `split` involves more code that would go through the same process.

src/amo/components/LanguagePicker.js
tofumatt: Nice! I noticed that in the console output too after you pointed this out, oops 🙈

PR #1481, opened 2016-12-14
Mamo cookie off 1480

### LINKED ISSUE #1480: View desktop link should set the mamo cookie to "off"
Currently the "mamo" cookie is deleted, instead the value of the cookie should be updated to "off".

This will help for when we want to explicitly want to enforce viewing of the old site. Once this is done we will also need to check that a mobile UA with the mamo cookie set to "off" sees the old site as this will be required for when we start to re-purpose the mamo cookie as a "beta" opt-in.

muffinresearch:
Fixes #1480

### DIFF
--- a/src/amo/containers/App.js
+++ b/src/amo/containers/App.js
@@ -47,7 +47,7 @@ export class AppBase extends React.Component {
   onViewDesktop = (event, { window_ = window, cookie_ = cookie } = {}) => {
     event.preventDefault();
     if (window_ && window_.location) {
-      cookie_.remove('mamo', { path: '/' });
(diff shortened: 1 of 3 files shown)


muffinresearch:
> Looks good, this will need an ops change as well I'm guessing, unless it is already checking the contents of the cookie and not just its existence. cc @jasonthomas

Once landed, I'll test this and file a separate issue if needed.

tofumatt:
Cool, sounds good.

muffinresearch:
Tested on -dev and everything is working perfectly.

PR #1493, opened 2016-12-16
Add extension/theme landing pages

### LINKED ISSUE #1285: Implement the themes home page

### LINKED ISSUE #1286: Implement the extensions home page

tofumatt:
Fixes #1285
Fixes #1286 

### Screenshots

screen shot 2016-12-16 at 15 57 55

screen shot 2016-12-16 at 15 58 03

screen shot 2016-12-16 at 15 58 05

### Video

dec-16-2016 16-03-42

### DIFF
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import React, { PropTypes } from 'react';
 
 import SearchResult from 'amo/components/SearchResult';
@@ -10,15 +11,17 @@ export default class AddonsCard extends React.Component {
   static propTypes = {
     addons: PropTypes.array.isRequired,
     children: PropTypes.node,
+    className: PropTypes.string,
     footer: PropTypes.node,
     header: PropTypes.node,
   }
 
   render() {
-    const { addons, children, footer, header } = this.props;
+    const { addons, children, className, footer, header } = this.props;
 
     return (
-      <Card header={header} footer={footer} className="AddonsCard"
+      <Card header={header} footer={footer}
+        className={classNames('AddonsCard', className)}
         ref={(ref) => { this.cardContainer = ref; }}>
         {children}
         {addons && addons.length ? (
--- a/src/amo/components/LandingPage.js
+++ b/src/amo/components/LandingPage.js
@@ -1,0 +1,121 @@
+import classNames from 'classnames';
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+import { asyncConnect } from 'redux-connect';
+import { connect } from 'react-redux';
+
+import LandingAddonsCard from 'amo/components/LandingAddonsCard';
+import { loadLandingAddons } from 'amo/utils';
+import translate from 'core/i18n/translate';
+
+import './LandingPage.scss';
+
+
+export class LandingPageBase extends React.Component {
+  static propTypes = {
+    addonType: PropTypes.string.isRequired,
+    featuredAddons: PropTypes.array,
+    highlyRatedAddons: PropTypes.array,
+    popularAddons: PropTypes.array,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  contentForType(addonType) {
+    const { i18n } = this.props;
+
+    const contentForTypes = {
+      extension: {
+        featuredHeader: i18n.gettext('Featured extensions'),
+        // TODO: Add this search/route, see:
+        // https://github.com/mozilla/addons-frontend/issues/1535
+        featuredFooterLink: '#/extensions/featured',
+        featuredFooterText: i18n.gettext('More featured extensions'),
+        popularHeader: i18n.gettext('Most popular extensions'),
+        popularFooterLink: {
+          pathname: '/search/',
+          query: { addonType: 'extension', sort: 'hotness' },
+        },
+        popularFooterText: i18n.gettext('More popular extensions'),
+        highlyRatedHeader: i18n.gettext('Top rated extensions'),
+        highlyRatedFooterLink: {
+          pathname: '/search/',
+          query: { addonType: 'extension', sort: 'rating' },
+        },
+        highlyRatedFooterText: i18n.gettext('More highly rated extensions'),
+      },
+      theme: {
+        featuredHeader: i18n.gettext('Featured themes'),
+        // TODO: Add this search/route, see:
+        // https://github.com/mozilla/addons-frontend/issues/1535
+        featuredFooterLink: '#/themes/featured',
+        featuredFooterText: i18n.gettext('More featured themes'),
+        popularHeader: i18n.gettext('Most popular themes'),
+        popularFooterLink: {
+          pathname: '/search/',
+          query: { addonType: 'theme', sort: 'hotness' },
+        },
+        popularFooterText: i18n.gettext('More popular themes'),
+        highlyRatedHeader: i18n.gettext('Top rated themes'),
+        highlyRatedFooterLink: {
+          pathname: '/search/',
+          query: { addonType: 'theme', sort: 'rating' },
+        },
+        highlyRatedFooterText: i18n.gettext('More highly rated themes'),
+      },
+    };
+
+    if (contentForTypes[addonType]) {
+      return contentForTypes[addonType];
+    }
+
+    throw new Error(`No LandingPage content for addonType: ${addonType}`);
+  }
+
+  render() {
+    const {
+      addonType, featuredAddons, highlyRatedAddons, popularAddons,
+    } = this.props;
+
+    const html = this.contentForType(addonType);
+
+    return (
+      <div className={classNames('LandingPage', `LandingPage-${addonType}`)}>
+        <LandingAddonsCard addons={featuredAddons}
+          className="FeaturedAddons" header={html.featuredHeader}
+          footerLink={html.featuredFooterLink}
+          footerText={html.featuredFooterText} />
+
+        <LandingAddonsCard addons={highlyRatedAddons}
+          className="HighlyRatedAddons" header={html.highlyRatedHeader}
+          footerLink={html.highlyRatedFooterLink}
+          footerText={html.highlyRatedFooterText} />
+
+        <LandingAddonsCard addons={popularAddons}
+          className="PopularAddons" header={html.popularHeader}
+          footerLink={html.popularFooterLink}
+          footerText={html.popularFooterText} />
+      </div>
+    );
+  }
+}
+
+export function singularizeAddonType(state, ownProps) {
+  return { addonType: ownProps.params.pluralAddonType.replace(/s$/, '') };
+}
+
+export function mapStateToProps(state) {
+  return {
+    featuredAddons: state.landing.featured.results,
+    highlyRatedAddons: state.landing.highlyRated.results,
+    popularAddons: state.landing.popular.results,
+  };
+}
+
+export default compose(
+  asyncConnect([
+    { deferred: true, promise: loadLandingAddons },
+  ]),
+  connect(mapStateToProps),
+  connect(singularizeAddonType),
+  translate({ withRef: true }),
+)(LandingPageBase);
--- a/src/amo/components/LandingPage.scss
+++ b/src/amo/components/LandingPage.scss
@@ -1,0 +1,5 @@
+@import "~core/css/inc/vars";
+
+.LandingPage {
+  padding: $padding-page;
+}
--- a/src/amo/routes.js
+++ b/src/amo/routes.js
@@ -7,6 +7,7 @@ import AddonReview from './components/AddonReview';
 import App from './containers/App';
 import CategoryList from './containers/CategoryList';
 import CategoryPage from './containers/CategoryPage';
+import LandingPage from './components/LandingPage';
 import Home from './containers/Home';
 import DetailPage from './containers/DetailPage';
 import SearchPage from './containers/SearchPage';
@@ -21,5 +22,6 @@ export default (
     <Route path=":addonType/:slug/" component={CategoryPage} />
     <Route path="fxa-authenticate" component={HandleLogin} />
     <Route path="search/" component={SearchPage} />
+    <Route path=":pluralAddonType/" component={LandingPage} />
   </Route>
 );
--- a/src/amo/utils.js
+++ b/src/amo/utils.js
@@ -1,0 +1,28 @@
+import { LANDING_PAGE_ADDON_COUNT } from 'amo/constants';
+import { getLanding, loadLanding, failLanding } from 'amo/actions/landing';
+import { featured as featuredAPI, search } from 'core/api';
+
+
+export function fetchLandingAddons({ addonType, api, dispatch }) {
+  dispatch(getLanding({ addonType }));
+
+  const filters = { addonType, page_size: LANDING_PAGE_ADDON_COUNT };
+  const landingRequests = [
+    featuredAPI({ api, filters }),
+    search({ api, filters: { ...filters, sort: 'rating' }, page: 1 }),
+    search({ api, filters: { ...filters, sort: 'hotness' }, page: 1 }),
+  ];
+
+  return Promise.all(landingRequests)
+    .then(([featured, highlyRated, popular]) => dispatch(
+      loadLanding({ addonType, featured, highlyRated, popular })
+    ))
+    .catch(() => dispatch(failLanding({ addonType })));
+}
+
+export function loadLandingAddons({ store: { dispatch, getState }, params }) {
+  const state = getState();
+  const addonType = params.pluralAddonType.replace(/s$/, '');
+
+  return fetchLandingAddons({ addonType, api: state.api, dispatch });
+}
--- a/tests/client/amo/components/TestLandingPage.js
+++ b/tests/client/amo/components/TestLandingPage.js
@@ -1,0 +1,112 @@
+import React from 'react';
+import {
+  renderIntoDocument,
+  findRenderedComponentWithType,
+} from 'react-addons-test-utils';
+import { findDOMNode } from 'react-dom';
+import { Provider } from 'react-redux';
+
+import * as landingActions from 'amo/actions/landing';
+import {
+  LandingPageBase,
+  mapStateToProps,
+  singularizeAddonType,
+} from 'amo/components/LandingPage';
+import createStore from 'amo/store';
+import I18nProvider from 'core/i18n/Provider';
+import { fakeAddon } from 'tests/client/amo/helpers';
+import { getFakeI18nInst } from 'tests/client/helpers';
+
+
+describe('<LandingPage />', () => {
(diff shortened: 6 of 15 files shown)


tofumatt:
r? Sorry I added the reviewer status but I'm guessing @kumar303 doesn't get the emails and I know they don't show up in assigned yet 😞 

I'll fix this conflict quickly in a moment.

kumar303:
Argh, I totally do get the emails but they are no different than other people's review comments (which I typically skim over) so they are easy to mix. Yeah, I think assigning PRs to me is the best way because I check that from the dashboard regular. I'll dig into this...

tofumatt:
No probs, it’s easy to set both, I just didn’t want it to be too noisy.
Evidently I needed to be more noisy 😉 🥁

src/amo/utils.js
```diff
@@ -0,0 +1,53 @@
...
+      popular: response[2],
+    })))
+    .catch(() => dispatch(failLanding({ addonType })));
+}
+
+export function loadLandingAddons({ store: { dispatch, getState }, params }) {
+  const state = getState();
+  const addonType = params.addonType.replace(/s$/, '');
+
+  if (!(addonType === state.addonType && !state.loading)) {
```
kumar303: Shouldn't this be `state.landing.addonType` and `state.lading.loading` ?

src/amo/utils.js
```diff
@@ -0,0 +1,53 @@
...
+    return fetchLandingAddons({ addonType, api: state.api, dispatch });
+  }
+
+  return true;
```
kumar303: I would prefer to see `Promise.resolve(true)` but maybe I'll give up. I hate `asyncConnect` so much.

tests/client/amo/components/TestLandingPage.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+describe('<LandingPage />', () => {
+  function render({ ...props }) {
+    const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
+
+    return findDOMNode(findRenderedComponentWithType(renderIntoDocument(
+      <Provider store={createStore(initialState)}>
+        <I18nProvider i18n={getFakeI18nInst()}>
+          <LandingPage {...props} />
```
kumar303: Because of all the HOC hotness, I think you can just do it like this now, right?

    <LandingPage store={createStore(initialState)} i18n={getFakeI18nInst()} {...props} />

tests/client/amo/components/TestLandingPage.js
```diff
@@ -0,0 +1,88 @@
...
+        {
+          authors: [{}],
+          name: 'Pop',
+        },
+        {
+          authors: [{}],
+          name: 'Pop again',
+        },
+      ],
+    };
```
kumar303: Did you try this by using real actions to set up the data? I think it would make the test better. It would look something like this:
````javascript
const store = createStore(initialState);
store.dispatch(setLanding({
  addonType: 'themes',
  featuredAddons: [
    {
      authors: [{}],
      name: 'Howdy again',
    },
  ],
  highlyRatedAddons: ...
}));
````

File: tests/client/amo/components/TestLandingPage.js [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+    const root = render(props);
+
+    assert.deepEqual(
+      Object.values(root.querySelectorAll('.SearchResult-heading'))
```
kumar303: We generally use refs to get at dom things like this. I guess I don't feel too strongly about it but using refs would be more consistent.

tests/client/amo/containers/TestExtensionsPage.js
```diff
@@ -0,0 +1,26 @@
...
+
+describe('ExtensionsPage mapStateToProps()', () => {
+  it('sets the props', () => {
+    const state = {
+      landing: {
+        featured: { results: 'featured' },
+        highlyRated: { results: 'high' },
+        popular: { results: 'pop' },
+      },
+    };
```
kumar303: Let me know if I'm off base here but I really think setting up raw state is dangerous. It makes it hard to iterate on the structure of our internal state. If you used actions to set up the state then it would mitigate that problem because you'd get test failures in all the right places. Is it too cumbersome to use the actions?

tests/client/amo/test_utils.js
```diff
@@ -0,0 +1,69 @@
+import createStore from 'amo/store';
+import * as landingActions from 'amo/actions/landing';
+import * as api from 'core/api';
+import { loadLandingAddons } from 'amo/utils';
+
+
+describe('AMO utils loadLandingAddons()', () => {
```
kumar303: mocha output isn't as nice as node/py.test but when you're hunting for a test file, using real file paths is more helpful. I'd do it like this:
````javascript
describe('amo/utils', () => {
  describe('loadLandingAddons()', () => {
    ...
  });
});
````

File: tests/client/amo/test_utils.js [hunk 8]
```diff
@@ -0,0 +1,69 @@
...
+      })
+      .returns(Promise.resolve({ entities, result }));
+
+    return loadLandingAddons({ store, params: ownProps.params })
+      .then(() => {
+        mockApi.verify();
+        assert.strictEqual(loadLandingAddons({
+          store: {
+            dispatch: sinon.stub(),
+            getState: () => ({
```
kumar303: this seems wrong. It should be returning `state.landing.addonType...` (I think?). Anyway, why do you need a stub store? Can't you just use the real one?

src/amo/containers/ExtensionsPage.js
```diff
@@ -0,0 +1,27 @@
+import { compose } from 'redux';
+import { asyncConnect } from 'redux-connect';
+import { connect } from 'react-redux';
+
+import LandingPage from 'amo/components/LandingPage';
+import { loadLandingAddons } from 'amo/utils';
+
+
+export function setAddonType(state, ownProps) {
+  return { addonType: ownProps.params.addonType.replace(/s$/, '') };
```
kumar303: We only use International English, not British English. (Just kidding.)

src/amo/utils.js
tofumatt: I actually agree, though @mstriemer pointed out there is an ever-so-slight performance difference in returning true.

But it's like `1ms` right? Do we care that much? I'm happy to use `Promise.resolve()` if we don't care, but I'm also obsessive enough that I'll take the odd API for a relatively easy `1ms` perf gain.

src/amo/components/LandingPage.js
```diff
@@ -0,0 +1,75 @@
...
+        highlyRatedFooter: (
+          <Link to="test">{i18n.gettext('More highly rated themes')}</Link>
+        ),
+      },
+    };
+
+    const html = contentForTypes[addonType];
+
+    return (
+      <div className={`LandingPage LandingPage-${addonType}`}>
```
tofumatt: Quite, thanks for catching it. ES6 string interpolation makes me lazy 😉

src/amo/components/LandingPage.scss
```diff
@@ -0,0 +1,9 @@
+@import "~core/css/inc/vars";
+
+.LandingPage {
+  padding: $padding-page;
+
+  .AddonsCard {
```
tofumatt: Totally right, thanks for noticing. I'll fix that up, should be easy.

src/amo/utils.js
```diff
@@ -0,0 +1,53 @@
...
+        addonType,
+        page_size: LANDING_PAGE_ADDON_COUNT,
+        sort: 'hotness',
+      },
+      page: 1,
+    }),
+  ];
+
+  return Promise.all(landingRequests)
+    .then((response) => dispatch(loadLanding({
```
tofumatt: Omg that works?! Cool.

src/amo/containers/ExtensionsPage.js
Ryuno-Ki: Uhh, US English became international? I prefer BE. Closer to home :-)

tests/client/amo/components/TestLandingPage.js
tofumatt: omg wow that's so much nicer, yeah 👍

tofumatt: Actually no because sub-components need the `i18n` too. I could do shallow render though, I'll try that.

src/amo/utils.js
kumar303: one tick on the event loop? Maybe. Seems like a wasted optimization effort. What's nice about returning a promise is that we can use these functions as utilities in other functions more easily. It doesn't break `asyncConnect` either way so I don't feel strongly.

tests/client/amo/components/TestLandingPage.js
tofumatt: Agreed, not using them definitely led to an error earlier. I will move to that. Sorry I keep not doing it, I promise I'm learning, I think these tests are just a bit old.

tests/client/amo/containers/TestExtensionsPage.js
tofumatt: No you're totally right. I'll fix it, we should never do this. 👍 👍 👍

tests/client/amo/test_utils.js
tofumatt: For sure. I'd love to tidy up all the test descriptions some lazy afternoon.

tests/client/amo/components/TestLandingPage.js
tofumatt: In this case I'm just testing the (dumb) component and involving actions seems to complicate the test and test things that are covered elsewhere in this test. This component doesn't even use the state so involving actions on this one seems too much.

tofumatt: (I did change it to use `fakeAddon` though which tidied it up a bit.)

tests/client/amo/components/TestLandingPage.js
tofumatt: I often find the ref tests more annoying to write (and they clutter up the HTML), especially when I end up just wanted to test rendered output.

In this case I'm testing things rendered by subcomponents and I find this to be much easier without any harm than dependency injection and refs.

tofumatt:
Style updated with blue small-caps:
screen shot 2016-12-22 at 23 14 58

tofumatt:
Okay, I addressed the PR comments I think and this is ready for another r?

I tweaked the style (screenshot above) and moved to actions for things pretty much everywhere which are much nicer and caught some bugs. 👍

tests/client/amo/components/TestLandingPage.js
kumar303: Yeah, that makes sense. I use both.

tests/client/amo/components/TestLandingPage.js
kumar303: Oh yeah, you're totally right. I thought this component was listening to state to get the data.

src/amo/components/AddonsCard/index.js
```diff
@@ -22,7 +22,8 @@ export default class AddonsCard extends React.Component {
         ref={(ref) => { this.cardContainer = ref; }}>
         {children}
         {addons && addons.length ? (
-          <ul className="AddonsCard-list">
+          <ul className="AddonsCard-list"
+            ref={(ref) => { this.cardList = ref; }}>
```
kumar303: Hmm, is this ref used anywhere? I don't see it

src/amo/components/LandingPage.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  contentForType(addonType) {
+    const { i18n } = this.props;
+
+    const contentForTypes = {
+      extension: {
+        featuredHeader: i18n.gettext('Featured extensions'),
+        featuredFooter: (
+          <Link className="LandingPage-more-link" to="test">
```
kumar303: `to="test"`? Is there another bug to fill these in?

src/amo/components/LandingPage.js
```diff
@@ -0,0 +1,118 @@
...
+    featuredAddons: PropTypes.array,
+    highlyRatedAddons: PropTypes.array,
+    popularAddons: PropTypes.array,
+    i18n: PropTypes.object.isRequired,
+  }
+
+  contentForType(addonType) {
+    const { i18n } = this.props;
+
+    const contentForTypes = {
```
kumar303: I see a lot of repetition here, especially with all the `<Link>` tags that mostly look the same. I think the only difference between `themes` and `extensions` will be the text and destinations. It would be nice to DRY this up with that in mind. It could happen in the next path when you fill in the destinations though.

src/amo/components/LandingPage.js
```diff
@@ -0,0 +1,118 @@
...
+  render() {
+    const {
+      addonType, featuredAddons, highlyRatedAddons, popularAddons,
+    } = this.props;
+
+    const html = this.contentForType(addonType);
+
+    return (
+      <div className={classNames('LandingPage', `LandingPage-${addonType}`)}>
+        <AddonsCard addons={featuredAddons} className="FeaturedAddons"
```
kumar303: You might find it easy to DRY up the repetition of all the footer `<Link>`s by moving each of these into a new component:
````javascript
class FeaturedAddonCard extends React.Component {
  render() {
    const footer = (
      <Link className="LandingPage-more-link" to="test">{this.props.footerText}</Link>
    );
    return (
      <AddonsCard addons={this.props.highlyRatedAddons} className="{this.props.className}"
          header={this.props.headerText} footer={footer} />
    );
  }
}
````

File: src/amo/routes.js [hunk 17]
```diff
@@ -21,5 +22,6 @@ export default (
     <Route path=":addonType/:slug/" component={CategoryPage} />
     <Route path="fxa-authenticate" component={HandleLogin} />
     <Route path="search/" component={SearchPage} />
+    <Route path=":pluralAddonType/" component={LandingPage} />
```
kumar303: thanks, the `pluralAddonType` makes it way less ambiguous now

src/amo/utils.js
```diff
@@ -0,0 +1,32 @@
...
+    featuredAPI({ api, filters }),
+    search({ api, filters: { ...filters, sort: 'rating' }, page: 1 }),
+    search({ api, filters: { ...filters, sort: 'hotness' }, page: 1 }),
+  ];
+
+  return Promise.all(landingRequests)
+    .then(([featured, highlyRated, popular]) => dispatch(
+      loadLanding({ addonType, featured, highlyRated, popular })
+    ))
+    .catch(() => dispatch(failLanding({ addonType })));
```
kumar303: are failures displayed anywhere? I need to figure out how to make `asyncConnect` work with the new API error handler.

src/amo/utils.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function loadLandingAddons({ store: { dispatch, getState }, params }) {
+  const state = getState();
+  const addonType = params.pluralAddonType.replace(/s$/, '');
+
+  if (!(addonType === state.landing.addonType && !state.landing.loading)) {
```
kumar303: parsing `!!state.landing.loading` is hard on my brain. This would be more clear to me:
````javascript
if (addonType !== state.landing.addonType || state.landing.loading) {...}
````
Although now I'm confused about the loading part of that if statement.

File: tests/client/amo/components/TestLandingPage.js [hunk 20]
```diff
@@ -0,0 +1,117 @@
...
+  it('throws if add-on type is not supported', () => {
+    assert.throws(() => {
+      render({ addonType: 'XUL' });
+    }, 'No LandingPage content for addonType: XUL');
+  });
+
+  describe('mapStateToProps()', () => {
+    it('sets the props', () => {
+      const store = createStore({ application: 'android' });
+      // store.dispatch(landingActions.getLanding({ addonType }));
```
kumar303: looks like an old comment

tests/client/amo/components/TestLandingPage.js
```diff
@@ -0,0 +1,117 @@
...
+    const root = render({
+      addonType: 'theme',
+    });
+
+    assert.include(root.textContent, 'Featured themes');
+    assert.include(root.textContent, 'More featured themes');
+  });
+
+  it('renders each add-on when set', () => {
+    const props = {
```
kumar303: I feel like this would be a great place to use `store.dispatch()` to populate the data because it will make sure that the rendering is tied directly to realistic state data. You can still still use `...fakeAddon` of course.

tests/client/amo/components/TestLandingPage.js
```diff
@@ -0,0 +1,117 @@
...
+    );
+  });
+
+  it('throws if add-on type is not supported', () => {
+    assert.throws(() => {
+      render({ addonType: 'XUL' });
+    }, 'No LandingPage content for addonType: XUL');
+  });
+
+  describe('mapStateToProps()', () => {
```
kumar303: Oh, I get your comment about not deleting much code with this new patch. To be honest, I don't think we need this test as long as you use `store.dispatch()` in the rendering test above. 

We typically isolate state mapper tests from component tests but only for when a real `store` object would make testing the component too difficult. In this case, it doesn't.

To put it another way: if you remove this test you'll still get 100% code coverage because the other rendering test will have to rely on mapStateToProps() to pass.

tests/client/amo/test_utils.js
```diff
@@ -0,0 +1,68 @@
...
+import * as api from 'core/api';
+import { loadLandingAddons } from 'amo/utils';
+
+
+describe('amo/utils', () => {
+  describe('loadLandingAddons()', () => {
+    const addonType = 'theme';
+    let ownProps;
+
+    before(() => {
```
kumar303: should this be `beforeEach` or do you only want it to run once?

src/amo/components/LandingPage.js
tofumatt: Yeah I thought that too and was gonna do it in a tidy-up bug to avoid too many components and more tests in this PR, but I'll do it here if you're cool with that.

src/amo/utils.js
tofumatt: Not yet, I wasn't sure how to hook them up, I just figured I should have an error action. Is it superfluous or is there a way I should do that?

src/amo/utils.js
tofumatt: Yeah this is actually too much, modelled in my head off the more complex checks for search.

In this case `addonType !== state.landing.addonType` is fine though that would make the page eventually stale... though this data isn't nearly as changed as even search results.

Basically this shouldn't be here I think... I'll remove it.

tests/client/amo/components/TestLandingPage.js
tofumatt: Removing this knocks test coverage down a bit but I'll move things around and see if I can use store for this, cheers. 👍

tofumatt:
Alright, addressed everything and created an issue for the one bit that's outside the scope of this PR (and possibly mobile V1).

Good for another r?

PR #1498, opened 2016-12-18
Remove UITour code 

### LINKED ISSUE #1380: Remove UITour code
The UITour code didn't work properly inside of `about:addons` and we went with a fallback. The dialog is now being implemented in Firefox for all installs so we will no longer need this code.

Any usage of this code should currently be wrapped in a config check for `useUiTour`.

atiqueahmedziad:
Fixes #1380 
@mstriemer please check if I have done it right.

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -27,7 +27,6 @@ module.exports = {
     'trackingEnabled',
     'trackingId',
     'trackingSendInitPageView',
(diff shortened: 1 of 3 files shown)


atiqueahmedziad:
@mstriemer done with updating the calls :)

mstriemer:
Great, thanks for the patch!

PR #1505, opened 2016-12-19
Don't show "no results" in between searches

### LINKED ISSUE #1485: Don't clear out search results when running a new search
Right now the search results page shows "No results found" in between a loaded set of results and a next page because we remove the results as part of the "start search" action.

We shouldn't clear them out.

Semi-related we should have a loading spinner there.

tofumatt:
Fixes #1485.

This fixes some of the wonky search result text we had in the past. It also preserves the search results on screen during the `SEARCH_STARTED` action so they stay until new results are loaded, which is less jarring.

Obvious the "searching..." text isn't great, we might UX on that in the future, but already this is better.

dec-19-2016 11-12-44

### DIFF
--- a/src/amo/components/SearchResults.js
+++ b/src/amo/components/SearchResults.js
@@ -31,25 +31,25 @@ class SearchResults extends React.Component {
 
     let messageText;
 
-    if (hasSearchParams && loading) {
+    if (loading) {
       messageText = i18n.gettext('Searching...');
-    } else if (!loading && count === 0) {
+    } else if (count === 0 && hasSearchParams) {
       if (query) {
         messageText = i18n.sprintf(
           i18n.gettext('No results were found for "%(query)s".'), { query });
-      } else if (hasSearchParams) {
+      } else {
         // TODO: Add the extension type, if available, so it says
         // "no extensions" found that match your search or something.
         messageText = i18n.gettext('No results were found.');
-      } else {
-        messageText = i18n.gettext(
-          'Please enter a search term to search Mozilla Add-ons.');
       }
+    } else if (!hasSearchParams) {
+      messageText = i18n.gettext(
+        'Please enter a search term to search Mozilla Add-ons.');
     }
 
     return (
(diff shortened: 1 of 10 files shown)


src/amo/components/SearchResults.js
```diff
@@ -31,25 +31,25 @@ class SearchResults extends React.Component {
...
       }
+    } else if (!hasSearchParams) {
+      messageText = i18n.gettext(
+        'Please enter a search term to search Mozilla Add-ons.');
     }
 
     return (
       <div ref={(ref) => { this.container = ref; }} className="SearchResults">
-        <AddonsCard addons={results}>
+        <AddonsCard addons={hasSearchParams ? results : null}>
```
tofumatt: I'm not sure I test for this specifically, gonna look into it.

tofumatt: Added one here: https://github.com/mozilla/addons-frontend/pull/1505/files#diff-93393917ee7e2eb6bbb9f980ae677f78R24

mstriemer:
I like that it puts the message at the top and keeps the results on the page, nice touch!

tests/client/amo/containers/TestSearch.js
```diff
@@ -23,13 +24,16 @@ describe('Search.mapStateToProps()', () => {
   };
 
   it('passes the search state if the URL and state query matches', () => {
-    const props = mapStateToProps(state, { location: { query: { q: 'ad-block' } } });
+    const props = mapStateToProps(state,
```
mstriemer: I really prefer the PEP8 style where all of the arguments need to be at the same indentation level.

```js
const props = mapStateToProps(
  state,
  { location: { query: { q: 'ad-block' } } });
// or
const props = mapStateToProps(
  state, { location: { query: { q: 'ad-block' } } });
// or, preferably not, bot possibly
const props = mapStateToProps(state,
                              { location: { query: { q: 'ad-block' } } });

Issue #1484, opened 2016-12-14
Show all unhandled API errors at the top of the page (global error handler)


kumar303:
As of https://github.com/mozilla/addons-frontend/pull/1418 components can choose to handle API errors themselves. Otherwise, we should show the error at the top of the page so that they don't get lost in the console.

kumar303:
After a first stab, this is harder than I thought it would be. You can't really handle all API errors from a central place because subsequent, unrelated API requests (made after the error) will erase the error from the UI. The erasing feature was intended to clear the error when you retry a failed request.

This was my failed attempt: https://github.com/mozilla/addons-frontend/pull/1496

kumar303:
I unassigned this for now since https://github.com/mozilla/addons-frontend/pull/1508 landed which takes care of all cases, I think.

kumar303:
I feel like we know how to work with custom error handlers now so this isn't really a concern.

PR #1508, opened 2016-12-19
Hook up AddonReview for API error handling

kumar303:
At first I wanted to make a global error handler (https://github.com/mozilla/addons-frontend/issues/1484) but that got hairy so I set it aside.

As a compromise, I looked through all AMO components that work with the API and added `withErrorHandling()` to make sure they display errors. I didn't look at any of the API actions that are executed via `asyncConnect` because that's not compatible with our higher order error handler component (specifically, there is no way to get the `errorHandler` property from the component props, unless I'm missing a way?). This leaves only one component: `AddonReview`. Am I forgetting anything?

Adding API error handling for it is pretty straight forward. The new behavior looks like this:

screenshot 2016-12-19 15 07 39

### DIFF
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -7,6 +7,7 @@ import { withRouter } from 'react-router';
 import { submitReview } from 'amo/api';
 import { setReview } from 'amo/actions/reviews';
 import { callApi } from 'core/api';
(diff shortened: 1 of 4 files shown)


mstriemer:
Actually can you remove the padding above the error message?

kumar303:
> Actually can you remove the padding above the error message

Good call, done.

PR #1509, opened 2016-12-20
Screenshot viewer on add-on detail

### LINKED ISSUE #936: Get add-on screenshots working
The add-on detail page was hooked up to API data in https://github.com/mozilla/addons-frontend/issues/883 but the screenshot widget is not working yet. It needs images and navigation, etc.

mstriemer:
Going to look through this diff one more time later tonight but wanted to get the PR up.

screenshots-portrait mov

screenshots-landscape mov

Fixes #936.

### DIFF
--- a/src/amo/components/ScreenShots.js
+++ b/src/amo/components/ScreenShots.js
@@ -9,6 +9,30 @@ import 'amo/css/ScreenShots.scss';
 
 const HEIGHT = 200;
 const WIDTH = 320;
+const PHOTO_SWIPE_OPTIONS = {
+  closeEl: true,
+  captionEl: true,
+  fullscreenEl: false,
+  zoomEl: false,
+  shareEl: false,
+  counterEl: true,
+  arrowEl: true,
+  preloaderEl: true,
+};
+
+const formatPreviews = (previews) => (
+  previews.map((preview) => ({
+    src: preview.image_url,
+    thumbnail_src: preview.thumbnail_url,
+    h: HEIGHT,
+    w: WIDTH,
+    title: preview.caption,
+  }))
+);
+
+export const thumbnailContent = (item) => (
+  <img src={item.src} className="ScreenShots-image" height={HEIGHT} width={WIDTH} alt="" />
+);
 
 export default class ScreenShots extends React.Component {
   static propTypes = {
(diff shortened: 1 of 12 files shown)


kumar303:
This looks shiny!

mstriemer:
@pwalm I have a version with the scroll indicator like in the mocks too but it unfortunately dropped the frame rate down to 10-20 fps on my OnePlus 3 so I'm thinking that might be a deal breaker on lower level hardware.

If you really want that to be a thing then let me know and I can look into optimising it but I'm not seeing what else I can do with it right now.

We'll need the dimensions of the images in the API response to make them format a little nicer. I have a start on a patch that I'll look at landing later but this is probably good enough for now.

src/amo/components/ScreenShots.js
```diff
@@ -1,41 +1,59 @@
...
+  render() {
+    const { previews } = this.props;
+    const items = previews.map((preview) => ({
+      src: preview.image_url,
+      thumbnail_src: preview.thumbnail_url,
+      h: HEIGHT,
+      w: WIDTH,
+      title: preview.caption,
+    }));
+    const photoSwipeOptions = {
```
tofumatt: Setting this in the `render()` when it doesn't ever change based on props seems odd to me; I don't _really_ think the performance difference matters, but it just doesn't seem like it needs to be here.

If it could be moved out that'd be cool.

src/amo/components/ScreenShots.js
```diff
@@ -1,41 +1,59 @@
...
+      closeEl: true,
+      captionEl: true,
+      fullscreenEl: false,
+      zoomEl: false,
+      shareEl: false,
+      counterEl: true,
+      arrowEl: true,
+      preloaderEl: true,
+    };
     return (
```
tofumatt: A newline here to separate out the return and the method setup would be nice.

src/amo/components/ScreenShots.js
```diff
@@ -1,41 +1,59 @@
...
+          <PhotoSwipeGallery
+            className="ScreenShots-list" close={this.onClose} items={items}
+            options={photoSwipeOptions} thumbnailContent={this.thumbnailContent} />
+        </div>
       </div>
     );
   }
 }
-
-export default translate({ withRef: true })(ScreenShotsBase);
```
tofumatt: 🎉

tests/client/amo/components/TestScreenShots.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { renderIntoDocument } from 'react-addons-test-utils';
+import { PhotoSwipeGallery } from 'react-photoswipe';
+
+import ScreenShots from 'amo/components/ScreenShots';
+import { shallowRender } from 'tests/client/helpers';
+
+describe('<ScreenShots />', () => {
+  const onePxImg = 'data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=';
```
tofumatt: Bytes are cheap, we can call this `onePixelImage`

tofumatt: Unless you are just trying to get it under 80 chars... in which case I guess it's okay but I really dislike short variable names like this. 😃

tofumatt: Also this is only used in one test, it could go into there as it really only relates to it (for now, anyway).

pwalm:
This looks SUPER good. Instead of a scrollbar, can we style the "1/5" to look nicer? I'll work on styling the description copy as well.

PR #1514, opened 2016-12-21
I18n dates 1395

tofumatt:
This adds support for moment and loading moment locales as part of our own (previously JSON, now JS) localisation files that we dynamically require depending on locale.

It's a bit of a change as we now have JS inside our locales. I've been at this several days and the way I got the locales in is a bit hairy but it works great in usage and avoids us loading all of moment's locales at once.

This adds moment info into the disco pane as well but seeing as disco pane doesn't use moment we might well be better off only adding them in for AMO. Let me know on that front, it's an easy change to make as it's all in the build script.

-----

While I was at it I added in the REASON I wanted moment; this tweak to the addon details (see the **last updated** bit:

screenshot 2016-12-21 16 29 14
screenshot 2016-12-21 16 29 02


tofumatt:
Oh my, the locale changes make this PR wild... though I'm not sure the tests will work without them? I'll try to create a PR without them, hopefully it'll work and I'll build the locales separate.

bin/build-locales
```diff
@@ -39,5 +40,23 @@ poFiles.forEach((pofile) => {
     format: 'jed1.x',
     fuzzy: config.get('po2jsonFuzzyOutput'),
   });
-  fs.writeFileSync(jsonfile, json);
+
+  // Add the moment locale JS into our locale file, if one is available.
+  var defineLocale = null;
+  try {
+    const momentLocale = locale.replace('_', '-').toLowerCase();
+    defineLocale = new Function(
```
mstriemer: I might be worth commenting on what this does. At first I thought it would eval but it looks like it will actually end up wrapping the file in a function and printing that to file.

bin/build-locales
```diff
@@ -39,5 +40,23 @@ poFiles.forEach((pofile) => {
...
+      fs.readFileSync(
+        `./node_modules/moment/locale/${momentLocale}.js`, 'utf8'
+      )
+    );
+  } catch (e) {
+    if (locale !== 'en_US') {
+      console.info(`No moment i18n available for ${locale}; consider adding one or creating a mapping.`);
```
mstriemer: This line is over 100 characters long. I'm guessing we don't lint the `bin` folder...

src/amo/components/AddonMoreInfo.js
```diff
@@ -32,7 +32,9 @@ export class AddonMoreInfoBase extends React.Component {
             {addon.current_version.version}
           </dd>
           <dt>{i18n.gettext('Last updated')}</dt>
-          <dd>{addon.last_updated}</dd>
+          <dd>
+            {i18n.moment(addon.last_updated).fromNow()} ({i18n.moment(addon.last_updated).format('MMMM Do YYYY')})
```
mstriemer: Should this be localised? I'm not sure if parens translate well and some languages might want something like "1 year ago on December 1, 2015"

src/core/i18n/translate.js
```diff
@@ -14,6 +19,23 @@ export default function translate(options = {}) {
       constructor(props, context) {
         super(props, context);
         this.i18n = context.i18n;
+
+        if (this.i18n) {
```
mstriemer: Can this move to server/base and client/base? Strictly speaking a component doesn't need to use `translate()` and it looks like this code might run more than once and keep configuring moment.

Changing this to `this.i18n && !this.i18n.momet` might fix that but moving it to where `i18n` is created probably makes the most sense.

src/core/server/base.js
```diff
@@ -182,17 +182,18 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
...
-              jedData = require(`json!../../locale/${locale}/${appInstanceName}.json`);
+              i18nData = require(
+                `../../locale/${locale}/${appInstanceName}.js`);
             }
           } catch (e) {
             log.info(`Locale JSON not found or required for locale: "${locale}"`);
             log.info(`Falling back to default lang: "${config.get('defaultLang')}".`);
           }
-          const i18n = new Jed(jedData);
+          const i18n = new Jed(i18nData);
```
mstriemer: Setting up moment here makes the most sense to me, provided there's nothing preventing that.

```js
function makeI18n(i18nData) {
  const i18n = new Jed(i18nData);
  moment.locale(/* yadda yadda */);
  i18n.moment = moment;
  return i18n;
}

const i18n = makeI18n(i18nData);

File: src/core/i18n/translate.js (hunk 4 above)
Comment (Ryuno-Ki) [inline src/core/i18n/translate.js]: Typo: !this.i18n.mome*n*t

File: bin/build-locales (hunk 2 above)
Comment (tofumatt) [inline bin/build-locales]: I don't think so, but also I forgot if we had `dedent` here. But I bet we do, I'll add it.

File: bin/build-locales (hunk 1 above)
Comment (tofumatt) [inline bin/build-locales]: Yeah, without doing this we end up writing `[object Object]`. Indeed though, it warrants an explanation.

File: bin/build-locales (hunk 2 above)
Comment (tofumatt) [inline bin/build-locales]: Actually maybe we don't. I'll just concat it here; the `bin/` folder is less pretty but I'll try not to make it a total garbage heap 👍

Comment (mstriemer) [inline bin/build-locales]: I think you can use `oneLine` from common-tags. Pretty sure that got added back in recently.

**Comment by tofumatt** [comment]
Moved things into `client/base.js` and `server/base.js`, which is much nicer, agreed.

I added some comments and cleaned things up too.

Unrelated but some long lines in `base.js` were irking me so I fixed them but in a separate commit; if it's too noisy let me know and I'll put it elsewhere.

Ready for another r?

**Comment by tofumatt** [comment]
Oh I also stopped adding moment locales to the disco pane. They're only added to AMO for now (it's hardcoded in the `build-locales` file but I'm just thinking we could probably add that as a config value, so maybe I'll do that in a few).

File: bin/build-locales [hunk 6]
```diff
@@ -39,5 +41,35 @@ poFiles.forEach((pofile) => {
...
+
+  // Add the moment locale JS into our locale file, if one is available and
+  // we're building for AMO (which is the only app that uses moment right
+  // now).
+  if (appName === 'amo') {
+    var defineLocale = null;
+    try {
+      const momentLocale = locale.replace('_', '-').toLowerCase();
+      // We're using `new Function()` here to create a function out of the
+      // raw code in the file; if we require it here it will evaluate here
```
mstriemer: I think mentioning why this is being wrapped in a function more explicitly might be nice. Something like:

```js
// We're using `new Function()` here to create a function out of the
// raw code in the file; this function won't be executed but will be
// written out by `toSource()` so that it can be used later.

File: src/amo/components/AddonMoreInfo.js [hunk 7]
```diff
@@ -32,7 +32,16 @@ export class AddonMoreInfoBase extends React.Component {
             {addon.current_version.version}
           </dd>
           <dt>{i18n.gettext('Last updated')}</dt>
-          <dd>{addon.last_updated}</dd>
+          <dd>
+            {i18n.sprintf(
+              // L10n: This will output, in English:
+              // "2 months ago (Dec 12 2016)"
```
mstriemer: That should probably be `Dec 12, 2016`.

src/amo/components/AddonMoreInfo.js
```diff
@@ -32,7 +32,16 @@ export class AddonMoreInfoBase extends React.Component {
...
+              i18n.gettext('%(timeFromNow)s (%(date)s)'), {
+                timeFromNow: i18n.moment(addon.last_updated).fromNow(),
+                date: i18n.moment(addon.last_updated).format('MMMM Do YYYY'),
```
mstriemer: I think we'll want to use the `ll` format for this. That will give `Dec 12, 2016` in English and `12 déc. 2016` in French.

src/core/i18n/utils.js
```diff
@@ -173,3 +175,26 @@ export function getLanguage({ lang, acceptLanguage } = {}) {
...
+// Create an i18n object with a translated moment object available we can
+// use for translated dates across the app.
+export function makeI18n(i18nData, _Jed = Jed) {
+  const i18n = new _Jed(i18nData);
+
+  // This adds the correct moment locale for the active locale so we can get
+  // localised dates, times, etc.
+  if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
+    i18n.options._momentDefineLocale();
+    moment.locale(makeMomentLocale(i18n.options.locale_data.messages[''].lang));
```
mstriemer: It looks like you can actually set a locale for an [instance of moment](http://momentjs.com/docs/#/i18n/instance-locale/). Maybe that would be better?

```js
i18n.moment = moment();
if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
  i18n.options._momentDefineLocale();
  i18n.moment.locale(makeMomentLocale(i18n.options.locale_data.messages[''].lang));
}
return i18n;

File: src/core/server/base.js [hunk 10]
```diff
@@ -259,16 +273,19 @@ export function runServer({ listen = true, app = appName } = {}) {
...
+          routes, createStore, { appInstanceName: app });
         if (listen === true) {
           server.listen(port, host, (err) => {
             if (err) {
               reject(err);
             }
-            log.info(`🔥  Addons-frontend server is running [ENV:${env}] [APP:${app}] ` +
-                     `[isDevelopment:${isDevelopment}] [isDeployed:${isDeployed}] ` +
-                     `[apiHost:${config.get('apiHost')}] [apiPath:${config.get('apiPath')}]`);
-            log.info(`👁  Open your browser at http://${host}:${port} to view it.`);
```
mstriemer: We allow 100 character lines elsewhere so this line and some of the others could probably stay as they were.

tests/client/core/i18n/test_utils.js
```diff
@@ -387,4 +389,54 @@ describe('i18n utils', () => {
...
+        }
+      };
+    });
+
+    beforeEach(() => {
+      // FIXME: Our moment is not immutable so we reset it before each test.
+      // This is annoying to work around because of the locale `require()`s
+      // and it only affects tests so it'd be nice to fix but doesn't break
+      // anything.
+      moment.locale('en');
```
mstriemer: If we were localising the moment instance you could check that `moment.locale()` is still `en` when setting the locale data to `fr` and remove this.

src/amo/components/AddonMoreInfo.js
tofumatt: Oh, nice. Done.

src/core/server/base.js
tofumatt: They are a lot easier to read under 80 and it still outputs the same:

screenshot 2016-12-22 23 37 15
vs
screenshot 2016-12-22 23 37 07

src/core/i18n/utils.js
tofumatt: I tried that, but frustratingly the docs don't seem to point out that the `moment()` returned by `moment()` is a `moment` object with `new Date()` set as the time. There isn't a way to make new moments like `moment(addon.last_updated)`. And I don't see an API to reset the date, so this doesn't actually work :-(

mstriemer: That is really unfortunate. I guess we could attach a function that passes the args to moment(), sets the locale and returns it.

It might not be worth the complexity but if it works first try might be worth a shot. If it doesn't work right away I'd say leave it.

PR #1520, opened 2016-12-22
Downgrade redux packages

mstriemer:
Upgrading redux and redux-connect to 5.x seems to have broken the `withInstallHelpers()` HOC. It looks like it is because the discoResults have an `addon` property which is the guid and that is overriding the `addon` object that gets set later or because the way `ownProps` are passed from a parent component's `mapStateToProps()` has changed.


mstriemer:
Yeah, I was thinking that too. You wouldn't notice unless `mozAddonManager` was available but maybe there's still a way to test it.

tofumatt:
Shoot, sorry! Thanks for this. Merged it as I headed out the door and
didn't get the chance to hit it locally.

- tofumatt

kumar303:
I don't think it's a reasonable expectation to test every package upgrade manually. We would never upgrade any packages that way.

Issue #1536, opened 2016-12-29
Camel case API responses in a consistent fashion


mstriemer:
### Describe the problem and steps to reproduce it:
Currently the [`DiscoPane`](https://github.com/mozilla/addons-frontend/blob/6d4258986c9343b58b0fa59856de72df01fe8a1c/src/disco/containers/DiscoPane.js#L123) component is camel casing the add-on keys when they are passed into the child component. On AMO we have not been camel casing keys at all.

### What happened?
The `showInfo` helper expects an `iconUrl` in `ownProps` but AMO provides `icon_url`.

### What did you expect to happen?
AMO should provide `iconUrl` or disco should provide `icon_url`.

### Anything else we should know?
We should be consistently camel casing in API responses or in reducers. Doing it in [`callApi`](https://github.com/mozilla/addons-frontend/blob/6d4258986c9343b58b0fa59856de72df01fe8a1c/src/core/api/index.js#L73) would be a one line change (other than fixing all the code we broke):

```js
    .then(({ response, jsonResponse }) => {
      if (response.ok) {
        return camelCaseProps(jsonResponse);
      }
```

The `camelCaseProps` helper will only do one level of keys, so we should make that function recursive or use [humps](https://www.npmjs.com/package/humps) instead which provides a `camelizeKeys` helper that will perform the operation on nested objects too.

muffinresearch:
Discussing this we prefer not to break the relationship between props and the real api responses.

PR #1537, opened 2016-12-29
Responsive InfoDialog for AMO

### LINKED ISSUE #1414: Show InfoDialog when an add-on is installed
The mobile frontend will be able to install add-ons once #1388 lands. The data is being set in state for the `InfoDialog` to appear but since it isn't being rendered and doesn't support mobile.

Update `InfoDialog` to support mobile, move it to `core/` and render it in `App` so that it can be displayed on install.

mstriemer:
info-dialog-responsive mov

Fixes #1414.


src/core/reducers/addons.js
```diff
@@ -26,6 +26,7 @@ export default function addon(state = initialState, action) {
       } else if (thisAddon.current_version && thisAddon.current_version.files.length > 0) {
         newState[key] = {
           ...thisAddon,
+          iconUrl: thisAddon.icon_url,
```
mstriemer: This is quite unforunate but I looked into switching everything over to camel cased and it looks like a good amount of work. I filed #1536.

src/core/containers/InfoDialog/index.js
```diff
@@ -39,4 +41,19 @@ export class InfoDialogBase extends React.Component {
   }
 }
 
-export default translate()(onClickOutside(InfoDialogBase));
+export const InfoDialogBase = compose(
```
tofumatt: 👍

src/disco/containers/DiscoPane.js
```diff
@@ -87,7 +85,7 @@ export class DiscoPaneBase extends React.Component {
   }
 
   render() {
-    const { results, i18n, AddonComponent, showInfoDialog, infoDialogData } = this.props;
+    const { results, i18n, AddonComponent } = this.props;
```
tofumatt: While this is being changed, `AddonComponent` should be first.

mstriemer:
> Why the move/rename of src/core/components/InfoDialog/index.js → src/core/containers/InfoDialog/index.js?

I moved it because it now handles its own redux state. I thought that was our distinction between components and containers. I think the distinction was inspired by https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0 originally but after reading it again it probably makes more sense to move the base stuff into `ui/components/InfoDialog` and make this `core/containers/InstallDialog`. I'm not sure that would provide much benefit right now though. I still think it makes sense for it to be a container but maybe we're not on the same page about that distinction.

mstriemer:
I guess I could have put the `ShowInfoDialog` bit into `core/containers/InfoDialog` and left the base stuff alone in the components folder.

PR #1538, opened 2016-12-29
Hook up homepage categories

### LINKED ISSUE #1290: Hook up category links on the homepage
The homepage is pretty much an empty shell. Once we know and have the category pages to link to we should update the links.

add-ons_for_firefox

mstriemer:
I had to swap out some stuff because we didn't have categories that mapped very well. Some of these aren't exact mappings and might be worth renaming as well. For example "Screenshot" links to "Photos & Media" which is mostly video downloaders.

Swapped extensions:
**Save stuff** for **Find news** (Feeds, News & Blogging)
**Share stuff** for **Play games** (Sports & Games) -- these are mostly for game info/streaming stuff.

Swapped themes:
**Mystical** for **Solid** (Solid)

screen shot 2016-12-29 at 16 36 07-fullpage

Fixes #1290.


mstriemer:
@pwalm thoughts on the icons? The news and games ones kind of work so I left them. How about the cateogries, something you'd prefer instead? I don't know Scott's GitHub handle so I can't mention him =\

src/amo/containers/Home.js
```diff
@@ -6,6 +7,22 @@ import translate from 'core/i18n/translate';
...
+    </Link>
+  </li>
+);
+CategoryLink.propTypes = {
+  name: PropTypes.string.isRequired,
+  slug: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+};
+const ExtensionLink = (props) => <CategoryLink type="extensions" {...props} />;
```
tofumatt: Some newlines here to separate out things would be nice.

src/amo/containers/Home.js
```diff
@@ -6,6 +7,22 @@ import translate from 'core/i18n/translate';
 
 import 'amo/css/Home.scss';
 
+const CategoryLink = ({ name, slug, text, type }) => (
```
tofumatt: I'd prefer components complex enough to have proptypes to just be in their own files (I just recently did this in an upcoming PR and find it's easier to grok things when there is a single component per file). Plus I dislike declaring proptypes like that.

tofumatt: To be clear I think the `ThemeLink` and `ExtensionLink` components are fair (and could even be exported from `CategoryLink` but they should be in their own file.

src/amo/containers/Home.js
```diff
@@ -17,49 +34,25 @@ export class HomePageBase extends React.Component {
...
-          <li className="HomePage-category-li HomePage-shop-online">
-            <Link to="#shop-online"><span>{i18n.gettext('Shop online')}</span></Link>
-          </li>
-          <li className="HomePage-category-li HomePage-be-social">
-            <Link to="#share-stuff"><span>{i18n.gettext('Be social')}</span></Link>
-          </li>
-          <li className="HomePage-category-li HomePage-share-stuff">
-            <Link to="#share-stuff"><span>{i18n.gettext('Share stuff')}</span></Link>
-          </li>
+          <ExtensionLink name="block-ads" slug="security-privacy" text={i18n.gettext('Block ads')} />
```
tofumatt: I've been a fan of using `children` over a prop argument like `text` as I think it makes the JSX more readable (especially for newbies) and more like HTML, which I like.

src/amo/css/Home.scss
```diff
@@ -136,6 +146,6 @@
   background-image: url('../img/home/sporty.svg');
 }
 
-.HomePage-mystical {
-  background-image: url('../img/home/mystical.svg');
```
tofumatt: The mystical SVG should get deleted then, right?

tofumatt:
Looks like you've got a linting error in `Home.scss`: https://travis-ci.org/mozilla/addons-frontend/jobs/187575704#L338

src/amo/containers/Home.js
mstriemer: This feels really specific to me to put in its own file and they aren't exported right now.

I just think this is nicer than doing a helper like `this.extensionLink({ name, slug, text })`.

src/amo/containers/Home.js
mstriemer: I was torn on this. I went with `text` since it makes it a little shorter but I'll switch it to `children`.

src/amo/css/Home.scss
mstriemer: Yeah, we can recover it if we add it back.

src/amo/containers/Home.js
mstriemer: Also I don't think this is "complex enough to have proptypes" we just require proptypes for everything even if it isn't really useful.

mstriemer:
Filed #1560 to review the categories/icons.

pwalm:
@mstriemer I will provide new icons for all of these today, we're using the old FxOS emoji now.

pwalm:
Icons are [here](https://github.com/mozilla/addons-frontend/tree/master/assets/collection%20icons)

PR #1541, opened 2016-12-30
fix: Use singular addonType in category search

### LINKED ISSUE #1540: Use singular "extension" and "theme" not "extensions" and "themes" in category search
The response to all category searches is the same because we are using `type: 'extensions'` instead of `type: 'extension'`. The API ignores our params in this case instead of 400ing (see: https://github.com/mozilla/addons-server/issues/4181).

Just need to single-case these arguments and it'll be fine. Not sure if this is a regression or not, but as I think it was working at some point maybe it is. Anyway, easy fix.

### LINKED ISSUE #1545: API calls for themes are using "theme" and not "persona" 😠
So I'm using `theme` in most places for tests and also actual API calls in the app but lightweight themes, the kind we want to fetch, are `type: 'persona'`. So that needs to be changed.

cc @diox: I assume there's no way to change the API at this point and rename `theme` to `oldTheme` or `completeTheme` as this'll be a weird relic going forward and presumably we're the biggest API consumer?

tofumatt:
Fix #1540.
Fix #1545.


mstriemer:
I think this should be apiAddonType or something because it isn't necessarily singular or plural it's one for the API and one for display. 

We've got another bug here too since we're using "theme" for the API version of "themes" but it should actually be "persona". A display to API add-on type lookup object would be best in my opinion.

I'd say do the conversion in the API helper and then we don't need to draw the distinction elsewhere in the code base.

tofumatt:
Yeah, agreed.  I filed #1545 about the theme/persona bit; I'd rather just fix it there across the board if that's okay. Thoughts?

mstriemer:
Fixing this all in one go sounds good to me 👍

tofumatt:
Fair enough, `apiAddonType` works for me, especially as like you say eventually it'll be converting `themes` to `persona`, lulz.

mstriemer:
Yeah =/

If you can do the conversion late enough, in the API call code rather than the component I think we can just refer to it as `addonType` everywhere else though.

tofumatt:
I guess in the new version we’ll never really want to ACTUALLY request a
real theme so passing around “theme” in the code and translating that to
“persona” in/around callApi() would be nicer to deal with. 👍🏻

tofumatt:
Heh, I knew https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/Categories.js#L53 and https://github.com/mozilla/addons-frontend/blob/master/src/amo/containers/CategoryPage.js#L17 would bite me eventually.

Changed to `apiAddonType` but also just addresses bits of #1545 at the same time because it made sense. Now we use constants most places. I'll do a more thorough pass when I address #1545 in a separate PR.

r?

src/core/utils.js
```diff
@@ -153,3 +157,17 @@ export function browserBase64Decode(str) {
     return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
   }).join(''));
 }
+
+export function apiAddonType(addonType) {
+  if (!API_ADDON_TYPES_MAPPING[addonType]) {
```
mstriemer: This should probably use `hasOwnProperty()` or use a `Map` or be an object created with `Object.create(null)`. Currently I believe this test would fail:

```js
assert.throws(() => {
  apiAddonType('hasOwnProperty'); // This would return a function
});
```

http://localhost:3000/en-US/firefox/hasOwnProperty/ ~> 500 error

src/core/constants.js
```diff
@@ -42,6 +42,22 @@ export const validAddonTypes = [
...
 ];
+// Mapping of the add-on types we show in URLs, etc. and what they map
+// to in the API (and how they're represented internally in the app).
+//
+// Examples:
+// * '/extensions/' -> 'extension'
+// * '/themes/' -> 'persona'
+export const API_ADDON_TYPES_MAPPING = {
+  extensions: EXTENSION_TYPE,
+  themes: API_THEME_TYPE,
```
tofumatt: I agree, but:

it-was-like-that-when-i-got-here

I'll see about getting rid of `API_THEME_TYPE` and just making it `THEME_TYPE`.

src/core/constants.js
```diff
@@ -42,6 +42,22 @@ export const validAddonTypes = [
...
+};
+export const VISIBLE_ADDON_TYPES_MAPPING = Object.keys(API_ADDON_TYPES_MAPPING)
+  .reduce((object, key) => {
+    // eslint-disable-next-line no-param-reassign
+    object[API_ADDON_TYPES_MAPPING[key]] = key;
```
tofumatt: Oh, right. I knew I was doing this wrong at the time but was lazy.

tofumatt:
Okay, changed the constant name (this resulted in a lot of changed files but makes it MUCH clearer what type of extension we're dealing with).

Addressed other comments too, r?

mstriemer:
I'm not sure why the old `THEME_TYPE` references changed to `LEGACY_XUL_THEME_TYPE`? Didn't we want it to be `"persona"` all along but we were using the wrong string? So changing `THEME_TYPE` to `"persona"` and leaving those in place should have been fine.

tofumatt:
The Disco pane uses the legacy type so I assumed it should stay that way.
Unless it’s also been wrong all along or the API calls for the disco pane
accept `theme` instead of `persona` as the type. Or the type isn’t ever
sent as a param.

mstriemer:
Oh, disco pain.

I think we need to be consistent throughout. It looks like the reducer [1] is just setting the type to `THEME_TYPE` if it looks like a theme so changing it should be fine. It looks like there's probably a lot of test code that will fail though, unfortunately. Those should be using `THEME_TYPE` instead of `'theme'` though.

[1] https://github.com/mozilla/addons-frontend/blob/e1ac56af59704010d1dd2616a7ca729c700f1421/src/core/reducers/addons.js#L23

tofumatt:
> Those should be using THEME_TYPE instead of 'theme' though.

Yeah caught lots of them already, the remaining ones'll be easy to fix too and it's nice to change 'em anyway.

I'll change everything over then!

tofumatt:
Turns out there were lots of instances where defined constants weren't being used, especially (and confusingly) in the tracking code where `'theme'` _is_ the right value. So yeah, gonna include that fix here and say #1545 gets fixed by this now.

tofumatt:
Ready for another r?

src/core/installAddon.js
```diff
@@ -36,14 +36,20 @@ import {
...
+
 
 export function installTheme(
   node, addon, { _themeAction = themeAction, _tracking = tracking } = {},
 ) {
   const { name, status, type } = addon;
   if (type === THEME_TYPE && [DISABLED, UNINSTALLED, UNKNOWN].includes(status)) {
     _themeAction(node, THEME_INSTALL);
-    _tracking.sendEvent({ action: 'theme', category: INSTALL_CATEGORY, label: name });
+    _tracking.sendEvent({
```
mstriemer: I really wish the `tracking` module was smarter and these calls could be more like:

```js
if (type === THEME_TYPE && /* ... */) {
  _tracking.themeInstalled(name);
} /* ... */
```

But that isn't the case. Anyway I find the name `DISCO_DATA_THEME` a bit weird and this code is run by mobile pages as well so maybe `TRACKING_TYPE_THEME` or similar would be better?

src/core/utils.js
```diff
@@ -153,3 +158,21 @@ export function browserBase64Decode(str) {
     return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
   }).join(''));
 }
+
+export function apiAddonType(addonType) {
+  if (!Object.prototype.hasOwnProperty.call(
```
mstriemer: Oh, crazy eslint rules. Probably makes sense but shucks is that ugly. Would saving `hasOwnProperty` off work? I can't decide if it's better to be explicit or try and make it look a little nicer. Either way works for me.

```js
const hasOwnProperty = Object.prototype.hasOwnProperty.call;

export function apiAddonType(addonType) {
  if (!hasOwnProperty(API_ADDON_TYPES_MAPPING, addonType)) { }
  /* ... */
}

src/core/installAddon.js
tofumatt: it-was-like-that-when-i-got-here

But yes, I'm up for changing it. Also I find it weird we have localised constant files, but that's a separate issue 😄 

I'll change it. 😊

src/core/utils.js
tofumatt: Yeah I was pretty surprised by that. I guess I see the sense but it's weird.

I tried the `const` thing, but it's less clear and also it didn't work. So I'll leave it as-is and be like "oh JavaScript".

wat

tofumatt:
Pedantic PRs are the best; then we don't subject anyone else to this code 😄. Thanks!

tofumatt:
While we're renaming constants actually I dislike that the constants that started this whole thing are inversely namespaced; they're `EXTENSION_TYPE` and `THEME_TYPE` but they should be `TYPE_EXTENSION` and `TYPE_THEME`.

I'ma rename them if that's cool? I'll wait to merge 'til you say okay 😉

mstriemer:
If you're using `TRACKING_TYPE_THEME` then maybe `ADDON_TYPE_THEME` would be good?

tofumatt:
That works for me.

Turns out the tracking types were NEVER constants and I shouldn’t have used
DISCO_ stuff for them, but I’m fixing that now. There was a lot of stuff in
tracking that was just using magic constants.

Mostly done, should have it up soon but it should be good to merge. Let me
know if it sounds like it needs another r? but I think it’s okay.

mstriemer:
Should be good to merge! Thanks

PR #1543, opened 2016-12-30
fix: Use convertFiltersToQueryParams on LandingPage

### LINKED ISSUE #1542: addontype query param is sent instead of type from landing page
Easy fix, coming in a second.

tofumatt:
fix #1542.

I should have been using `convertFilterstoQueryParams` here and wasn't, so the extension type wasn't being passed to search.

### DIFF
--- a/src/amo/components/LandingPage.js
+++ b/src/amo/components/LandingPage.js
@@ -28,7 +28,10 @@ export class LandingPageBase extends React.Component {
         featuredHeader: i18n.gettext('Featured extensions'),
         // TODO: Add this search/route, see:
         // https://github.com/mozilla/addons-frontend/issues/1535
-        featuredFooterLink: '#/extensions/featured',
+        featuredFooterLink: {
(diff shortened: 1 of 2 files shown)


tofumatt:
Yeah, there's another PR I have in the queue for that which will use real URLs... I guess I can just make them real now and fill in the pages later. I'll do that an merge it in.

tofumatt:
Oh and the `Link` component's `to` prop converts query objects to params which I like more than actually manually adding in the query string ourselves.

src/amo/components/LandingPage.js
```diff
@@ -28,7 +28,10 @@ export class LandingPageBase extends React.Component {
         featuredHeader: i18n.gettext('Featured extensions'),
         // TODO: Add this search/route, see:
```
tofumatt: Actually I have the TODO above here to convert it later; I'll do that then.

PR #1544, opened 2016-12-30
Add search sort

### LINKED ISSUE #1297: Implement the search sort widget
Since this is essentially changing the search the easiest way to implement this is probably going to be with a list of links that are revealed when the sort is expanded as opposed to trying to use a form control which will get messy if there's implicit submission.

The whole thing should be keyboard accessible.

See https://docs.google.com/document/d/13g8tmoII8oCNrUXOl2HegsQst7WEsXukBnIqPUNrqb0/edit#heading=h.22jk2sku8h5t

amo_-_v1_sketch

tofumatt:
Fixes #1297.

Right now we're using "relevance" from the mocks as the default but I'm not sure that's exactly what the default search from the API actually is. http://addons-server.readthedocs.io/en/latest/topics/api/addons.html#search says "sorted by weekly downloads" is the default, so perhaps we need to add a "relevance" sort option (which should be the default).

cc @diox: should I file a bug for addons-server to get "relevance" as a sort option? Or is calling the default sort "relevance" appropriate enough here?

## Screenshots:
screenshot 2016-12-30 15 42 15
screenshot 2016-12-30 15 42 11

## Video:
dec-30-2016 15-41-35
(Note: is keyboard accessible)

### DIFF
--- a/src/amo/components/SearchPage.js
+++ b/src/amo/components/SearchPage.js
@@ -3,6 +3,7 @@ import React, { PropTypes } from 'react';
 import Link from 'amo/components/Link';
 import Paginate from 'core/components/Paginate';
 import SearchResults from 'amo/components/SearchResults';
+import SearchSort from 'amo/components/SearchSort';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 
 import SearchResult from './SearchResult';
@@ -29,6 +30,7 @@ export default class SearchPage extends React.Component {
     ResultComponent: SearchResult,
     filters: {},
     pathname: '/search/',
+    results: [],
   }
 
   render() {
@@ -42,12 +44,16 @@ export default class SearchPage extends React.Component {
       <Paginate LinkComponent={LinkComponent} count={count} currentPage={page}
         pathname={pathname} queryParams={queryParams} showPages={0} />
     ) : [];
+    const searchSort = hasSearchParams && results.length ? (
+      <SearchSort filters={filters} pathname={pathname} />
+    ) : null;
 
     return (
       <div className="SearchPage">
+        {searchSort}
         <SearchResults ResultComponent={ResultComponent} count={count}
           hasSearchParams={hasSearchParams} loading={loading} results={results}
-          filters={filters} />
+          filters={filters} pathname={pathname} />
         {paginator}
       </div>
     );
--- a/src/amo/components/SearchSort/SearchSort.scss
+++ b/src/amo/components/SearchSort/SearchSort.scss
@@ -1,0 +1,82 @@
+@import '~core/css/inc/mixins';
+@import '~core/css/inc/vars';
+
+@mixin arrow-link-right-aligned() {
+  background-image: url('~amo/img/icons/arrow-link-color.svg');
+  margin-right: 0;
+  position: absolute;
+  right: 8px;
+  top: 14px;
+}
+
+.SearchSort {
+  background: $base-color;
+  border-radius: $border-radius-default;
+  border: 0;
+  margin: 0 ($padding-page * 2) $padding-page;
+  overflow: hidden;
+}
+
+.SearchSort-toggle {
+  @include button($background: $base-color, $color: $link-color);
+
+  // We do this so the button doesn't have weird edges on hover when the
+  // sort options are expanded.
+  border-bottom-left-radius: 0;
+  border-bottom-right-radius: 0;
+  box-sizing: border-box;
+  display: block;
+  font-size: $font-size-s;
+  font-variant: small-caps;
+  font-weight: normal;
+  line-height: 2;
+  margin: 0;
+  min-width: 40%;
+  position: relative;
+  text-align: center;
+
+  &::before {
+    background: url('~amo/img/icons/sort.svg') 50% 50% no-repeat;
+    background-size: cover;
+    content: '';
+    display: inline-block;
+    height: 16px;
+    left: -8px;
+    position: relative;
+    top: 4px;
+    width: 16px;
+  }
+
+  &::after {
+    @include arrow-down();
+    @include arrow-link-right-aligned();
+
+    .SearchSort--visible & {
+      @include arrow-up();
+      @include arrow-link-right-aligned();
+    }
+  }
+
+  &:focus {
+    background: transparentize($link-color, 0.8);
+    box-shadow: none;
+  }
+}
+
+.SearchSort-list {
+  display: none;
+  margin: 0;
+  padding: 0;
+  align-items: center;
+  flex-flow: row;
+  flex-wrap: wrap;
+
+  .SearchSort--visible & {
+    display: flex;
+  }
+}
+
+.SearchSort-list-item {
+  display: flex;
+  width: 50%;
+}
--- a/src/amo/components/SearchSort/index.js
+++ b/src/amo/components/SearchSort/index.js
@@ -1,0 +1,85 @@
+import classNames from 'classnames';
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+
+import SearchSortLink from 'amo/components/SearchSort/SearchSortLink';
+import translate from 'core/i18n/translate';
+
+import './SearchSort.scss';
+
+
+const DEFAULT_SORT = 'relevance';
+
+export class SearchSortBase extends React.Component {
+  static propTypes = {
+    filters: PropTypes.object.isRequired,
+    i18n: PropTypes.object.isRequired,
+    pathname: PropTypes.string.isRequired,
+  }
+
+  state = { sortVisible: false };
+
+  onClick = (event) => {
+    event.preventDefault();
+    // Blur the search sort toggle if we're closing it, but only if
+    // we've clicked on it; if the link was keyboard activated we don't want
+    // to mess with focus, see:
+    // https://github.com/mozilla/addons-frontend/pull/1544#discussion_r96052790
+    this.toggleSort({ allowBlur: true });
+  }
+
+  onKeyPress = (event) => {
+    event.preventDefault();
+    this.toggleSort();
+  }
+
+  sortOptions() {
+    const { i18n } = this.props;
+
+    return [
+      { sort: 'updated', text: i18n.gettext('Recently Updated') },
+      { sort: 'relevance', text: i18n.gettext('Relevance') },
+      { sort: 'users', text: i18n.gettext('Most Users') },
+      { sort: 'rating', text: i18n.gettext('Top Rated') },
+    ];
+  }
+
+  toggleSort({ allowBlur = false } = {}) {
+    if (allowBlur === true && this.state.sortVisible) {
+      this.searchToggle.blur();
+    }
+    this.setState({ sortVisible: !this.state.sortVisible });
+  }
+
+  render() {
+    const { filters, i18n, pathname } = this.props;
+    const { sortVisible } = this.state;
+    const currentSort = filters.sort || DEFAULT_SORT;
+
+    return (
+      <div className={classNames('SearchSort', {
+        'SearchSort--visible': sortVisible,
+      })}>
+        <a className="SearchSort-toggle" href="#SearchSortOptions"
+          onClick={this.onClick} onKeyPress={this.onKeyPress}
+          ref={(ref) => { this.searchToggle = ref; }}>
+          {i18n.gettext('Sort')}
+        </a>
+        <ul id="SearchSortOptions" className="SearchSort-list">
+          {this.sortOptions().map((option) => (
+            <li className="SearchSort-list-item">
+              <SearchSortLink currentSort={currentSort} filters={filters}
+                pathname={pathname} sort={option.sort}>
+                {option.text}
+              </SearchSortLink>
+            </li>
+          ))}
+        </ul>
+      </div>
+    );
+  }
+}
+
+export default compose(
+  translate({ withRef: true }),
+)(SearchSortBase);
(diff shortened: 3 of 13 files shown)


diox:
The docs are a bit wrong in this case: if you do specify a `q` parameter *and* don't specify a `sort`, then it sorts by relevance. Said differently, it sorts by relevance by default as long as you have a query.

I'll fix the docs.

tofumatt:
@diox In that case, can we add an explicit `sort=reference` param that does the same as the default then? I'll file the issue in server, but it'd be handy to not have to special-case "empty equal relevance" in the front-end code for this sorter.

diox:
s/reference/relevance/ but yes, that works for me. Assign it to me directly and I'll get right on it.

mstriemer:
Nice! Thanks for including the keyboard version in the video 👏 🎉

src/amo/components/SearchSort/SearchSort.scss
```diff
@@ -0,0 +1,82 @@
...
+
+.SearchSort {
+  background: $base-color;
+  border-radius: $border-radius-default;
+  border: 0;
+  margin: 0 5% 10px;
+  overflow: hidden;
+}
+
+.SearchSort-toggle {
```
mstriemer: These styles aren't specific to a `SearchSort`, just that dropdown component. It would be nice to move this into a `Picker` or `Dropdown` element or something (maybe prefix with `Inline` but that could be screen size dependent).

src/amo/components/SearchSortLink/index.js
```diff
@@ -0,0 +1,36 @@
...
+import Link from 'amo/components/Link';
+
+import './SearchSortLink.scss';
+
+
+export function queryString(filters) {
+  return makeQueryString(convertFiltersToQueryParams(filters));
+}
+
+export default class SearchSortLink extends React.Component {
```
mstriemer: I don't feel like these small components really need to be in their own file. It does make the main component a bit smaller though.

Perhaps it would be better to include this within `components/SearchSort` instead of giving it its own component directory? That way it's sort of an internal component.

```
amo/components/SearchSort/index.js
amo/components/SearchSort/SearchSort.scss
amo/components/SearchSort/SearchSortLink.js
amo/components/SearchSort/SearchSortLink.scss

File: tests/client/amo/components/TestSearchSort.js [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+    // Open the sort options.
+    Simulate.click(rootNode.querySelector('.SearchSort-toggle'));
+    assert.ok(searchToggle.blur.notCalled);
+
+    // Close the sort options, which should blur the toggle link.
+    Simulate.click(rootNode.querySelector('.SearchSort-toggle'));
+    assert.ok(searchToggle.blur.called);
+  });
+  //
+  // it('is expanded by default', () => {
```
mstriemer: Commented out tests here, can't tell if they're useful or not.

src/amo/components/SearchSort/index.js
```diff
@@ -0,0 +1,79 @@
...
+    return [
+      { sort: 'updated', text: i18n.gettext('Recently Updated') },
+      { sort: 'relevance', text: i18n.gettext('Relevance') },
+      { sort: 'users', text: i18n.gettext('Most Users') },
+      { sort: 'rating', text: i18n.gettext('Top Rated') },
+    ];
+  }
+
+  toggleSort(searchToggle) {
+    // const searchToggle = _searchToggle || this.searchToggle;
```
mstriemer: 🔥

tests/client/amo/components/TestSearchSort.js
```diff
@@ -0,0 +1,93 @@
...
+
+    Simulate.click(rootNode.querySelector('.SearchSort-toggle'));
+
+    assert.include(rootNode.className, 'SearchSort--visible');
+  });
+
+  it('blurs the search sort toggle when closing the sort options', () => {
+    const searchToggle = { blur: sinon.stub() };
+    const root = render({
+      _searchToggle: searchToggle,
```
mstriemer: If you use `event.currentTarget` in `onClick` then you should be able to accomplish this without `_searchToggle`.

```js
// May need to set preventDefault, too.
Simulate.click(root.searchToggle, { currentTarget: searchToggle });
```

Alternatively, rather than passing in `_searchToggle` you could check that it is set to the right element then stub it.

```js
assert.ok(root.searchToggle.classList.contains('SearchSort-toggle'));
sinon.stub(root, 'searchToggle', searchToggle);

File: tests/client/amo/components/TestSearchSort.js (hunk 3 above)
Comment (Ryuno-Ki) [inline tests/client/amo/components/TestSearchSort.js]: Why don't you [skip them](https://mochajs.org/#inclusive-tests)? This way they will be reported as _pending_ and won't be forgotten.

File: src/amo/components/SearchPage.js [hunk 6]
```diff
@@ -45,9 +46,12 @@ export default class SearchPage extends React.Component {
 
     return (
       <div className="SearchPage">
+        {hasSearchParams ? (
```
tofumatt: I feel like it's weird to show it with no params (and arguably whenever there are no results, actually).

Hitting that search page without a query gives a "please enter a search term" message. Adding bits that only appear once there are results to filter seems like bad UI/UX to me.

src/amo/components/SearchSort/index.js
```diff
@@ -0,0 +1,79 @@
...
+    const { sortVisible } = this.state;
+    const currentSort = filters.sort || 'relevance';
+
+    return (
+      <div className={classNames('SearchSort', {
+        'SearchSort--visible': sortVisible,
+      })}>
+        <a className="SearchSort-toggle" href="#SearchSortOptions"
+          ref={(ref) => { this.searchToggle = ref; }} onClick={this.onClick}>
+          {i18n.gettext('Sort')}
```
tofumatt: I guess. Problem with that is I'm not sure where the icon should be LTR vs RTL and I don't feel like it's content so including it in the link feels weird. I like the idea of the Icon component (I saw it and my first crack at this used that), but in usage it feels odd, especially as there are two icons here (one left, one right).

I guess in this one it's not too bad, I'll take a crack at it.

src/amo/components/SearchSort/SearchSort.scss
```diff
@@ -0,0 +1,82 @@
...
+  position: absolute;
+  right: 8px;
+  top: 14px;
+}
+
+.SearchSort {
+  background: $base-color;
+  border-radius: $border-radius-default;
+  border: 0;
+  margin: 0 5% 10px;
```
tofumatt: You're right, I think I was just being wrong.

src/amo/components/SearchSortLink/index.js
tofumatt: I'm cool with that. I much prefer components in their own file; the consistency helps me and I think definitely helps newbies. Determining "what is big enough for its own file and what should just be another component declared as a `const` in the same file" is way more overhead than "New... File" 😉

tests/client/amo/components/TestSearchSort.js
tofumatt: @Ryuno-Ki We know how to skip tests, these were remnants of a copy/paste job that I didn't entirely clean up. We don't need to mark them as pending.

@mstriemer From copy/paste of a similar component. None are useful and the previous tests cover things; will 🔥! 😄

src/amo/components/SearchSort/SearchSort.scss
tofumatt: Yeah, eventually moving this out would be nice but I think that's a refactor we can do later, if that's cool?

src/amo/components/SearchSort/index.js
tofumatt: Actually, because the icons here already require CSS moving about and moving to Icons everywhere would be a better bug IMO (I'm happy to move to them, but it's more than just here we'd need to), can we just file a bug to move every icon to an `<Icon>` component if that makes sense?

src/amo/components/SearchPage.js
mstriemer: > Adding bits that only appear once there are results to filter seems like bad UI/UX to me.
> 
> I should change this to check for results as well though; no sense in showing it when there isn't anything to sort.

I feel I was suggesting the same thing as the first sentence, but then the second one says it should only appear when there are results 😕

It does seem a bit pointless to include it when there are no results to filter but just always having it there makes sense to me too. Maybe you want to set it to "Top rated" and then enter your search term.

In the end it doesn't matter much but I would lean towards the consistency of it always being there.

src/amo/components/SearchSort/SearchSort.scss
mstriemer: It was really convenient to have that `Card` component ready to use but I don't think we have any other uses for this one lined up other than "it looks like something we'll reuse" so that's cool.

src/amo/components/SearchPage.js
tofumatt: I think I meant "showing stuff" and not "adding stuff". The happy path doesn't get users to a blank search page but if they went there and there are no results I wouldn't want a sort. And also if they search and get "no results" a search sort is weird and also useless. The entire page changes on search and based on whether there are results or not (the paginator does the same thing), so I think conditionally showing/hiding search sort makes sense.

tofumatt:
It should be EASY to separate out, and we could totally do it. I just wanna
get this merged :-)

tofumatt:
I think I addressed most comments, could you have another look?

src/amo/components/SearchSort/index.js
```diff
@@ -0,0 +1,79 @@
...
+      { sort: 'relevance', text: i18n.gettext('Relevance') },
+      { sort: 'users', text: i18n.gettext('Most Users') },
+      { sort: 'rating', text: i18n.gettext('Top Rated') },
+    ];
+  }
+
+  toggleSort() {
+    // Blur the search sort toggle if we're closing it.
+    if (this.state.sortVisible) {
+      this.searchToggle.blur();
```
mstriemer: I'm a little curious what this is for. Does it hide the focus style once the dropdown closes?

tests/client/amo/components/TestSearchSort.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('blurs the search sort toggle when closing the sort options', () => {
+    const fakeSearchToggle = { blur: sinon.stub() };
+    const root = render({ filters: { query: 'test' }, pathname: '/search/' });
+
+    // Open the sort options.
+    root.toggleSort();
+
+    sinon.stub(root, 'searchToggle', fakeSearchToggle);
```
mstriemer: Looks like you could stub this before the first `root.toggleSort()` and check it wasn't called on open then check it was called on close. Not a bit deal but should shorten the test a little.

mstriemer:
I restarted that test that failed at one point, could be legit.

Probably worth marking the issues as needs more qa and getting some extra testing done on the disco pane to ensure nothing broke.

tofumatt:
I think it’s just a timeout thing; one set of Travis tests passed and the
others didn’t and their UI tests. :-/

- tofumatt

On 13 January 2017 at 19:58:09, Mark Striemer (notifications@github.com)
wrote:

I restarted that test that failed at one point, could be legit.

Probably worth marking the issues as needs more qa and getting some extra
testing done on the disco pane to ensure nothing broke.

—
You are receiving this because you authored the thread.
<https://github.com/mozilla/addons-frontend/pull/1544#issuecomment-272532850>,
or mute the thread

src/amo/components/SearchSort/index.js
```diff
@@ -0,0 +1,79 @@
...
+export class SearchSortBase extends React.Component {
+  static propTypes = {
+    filters: PropTypes.object.isRequired,
+    i18n: PropTypes.object.isRequired,
+    pathname: PropTypes.string.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { sortVisible: false };
```
tofumatt: Oh that's much nicer, cool. I'll try that. 👍

src/amo/components/SearchSort/index.js
tofumatt: Yeah. If you're *closing* the search sort it is a lot visually nicer to blur it. I thought it would make sense for keyboard users too, but maybe it should only be `onClick` and not `onKeyDown` so it doesn't reset a keyboard user's tab position (which I assume it will do, but have to admit am not sure about).

tofumatt:
I'll check into the test and make the requested changes, thanks!

src/amo/components/SearchSort/index.js
mstriemer: I was curious about the keyboard situation. After watching the video I think doing it in `onClick` is probably best. Keyboard is probably still want it focused.

tofumatt: The `onClick` (and by extension taps on mobile) now passes an argument to blur the sort when closed. Keyboard focus is not changed when closed:

jan-14-2017 18-29-15

Good call 👍

tests/client/amo/components/TestSearchSort.js
tofumatt: Yeah these tests were a bit off and I was stuck getting them to work. Turns out I was doing something a bit off and the best solution was just a `sinon.spy(root.searchToggle, 'blur')`. It's clearer now 👍

src/amo/components/SearchPage.js
```diff
@@ -42,12 +44,16 @@ export default class SearchPage extends React.Component {
       <Paginate LinkComponent={LinkComponent} count={count} currentPage={page}
         pathname={pathname} queryParams={queryParams} showPages={0} />
     ) : [];
+    const searchSort = hasSearchParams && results.length ? (
+      <SearchSort filters={filters} pathname={pathname} />
+    ) : [];
```
tofumatt: I agree; the tests are more wordy to write when it's `null` but I'll do it since you mentioned it and I was already irked by it. 😄

PR #1547, opened 2016-12-30
fix(package): update normalizr to version 3.0.1

### LINKED ISSUE #1555: Update normalizr code to 3.0.0
normalizr did a [major rewrite for version 3](https://github.com/paularmstrong/normalizr/blob/master/CHANGELOG.md#v300), so we'll need to update all our normalizr code to use their new API.

We'll also need to update our own docs where we [mention normalizr usage](https://github.com/mozilla/addons-frontend/blob/master/docs/adding-a-page.md#hitting-the-api).

tofumatt:
Closes #1462
Fixes #1555

https://greenkeeper.io/

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -3,7 +3,7 @@
 import url from 'url';
 
 import 'isomorphic-fetch';
-import { Schema, arrayOf, normalize } from 'normalizr';
+import { schema as normalizrSchema, normalize } from 'normalizr';
 import config from 'config';
 
 import log from 'core/logger';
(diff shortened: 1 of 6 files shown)


tofumatt:
I updated our code to work with `normalizr`'s new API. r?

src/core/api/index.js
```diff
@@ -12,9 +12,12 @@ import { convertFiltersToQueryParams } from 'core/searchUtils';
 
 const API_BASE = `${config.get('apiHost')}${config.get('apiPath')}`;
 
-export const addon = new Schema('addons', { idAttribute: 'slug' });
-export const category = new Schema('categories', { idAttribute: 'slug' });
-export const user = new Schema('users', { idAttribute: 'username' });
+export const addon = new normalizrSchema.Entity(
```
mstriemer: This is a lot longer than I'd like. Maybe save `Entity` off into a const and use that?

src/disco/api.js
```diff
@@ -1,17 +1,17 @@
...
 import { callApi } from 'core/api';
 import { getGuid } from 'core/reducers/addons';
 
-export const discoResult =
-  new Schema('discoResults', { idAttribute: (result) => getGuid(result.addon) });
-export const addon = new Schema('addons', { idAttribute: getGuid });
-discoResult.addon = addon;
+
+export const addon = new schema.Entity('addons', {}, { idAttribute: getGuid });
+export const discoResult = new schema.Entity('discoResults',
```
mstriemer: Can you break early here and drop `'discoResults',` down a line too?

Issue #1535, opened 2016-12-24
Add "Featured" search page


tofumatt:
https://github.com/mozilla/addons-frontend/pull/1493#discussion_r93798968 points out that we need to link to the popular and top rated searches for the "More popular add-ons" and "More highly rated add-ons" links on the landing pages (#1285 and #1286).

I'm not sure we have a "Featured" page (eg, can it be done using a search) or if we need to create a page with pagination that loads all featured add-ons.

Alternately–do we even want a "more featured add-ons" search page? A big list of "featured" add-ons almost seems odd to me, but I guess we can tackle that later as it isn't hard to build.

tofumatt:
Actually, [Featured add-ons are sorted randomly](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#featured) and like I mentioned it's weird to have a search of featured add-ons... so we can build out a page of featured add-ons but it's going to be a weird UX as the landing page of 5 add-ons could have different results than the full page.

Anyway, I'll build out the featured page, but yeah... maybe we need to tweak the API for this? Or get all featured add-ons with a large `page_size` argument to `/api/v3/addons/featured/`

ValentinaPC:
Verified as fixed on AMO-dev (Android 6.0.1)
Postfix screenshots (featured extensions and featured themes pages):
![fe](https://cloud.githubusercontent.com/assets/15685960/22683069/dee911a0-ed1e-11e6-883c-76a6675f3e09.png)
![ft1](https://cloud.githubusercontent.com/assets/15685960/22683068/dee89a04-ed1e-11e6-922e-06547876df50.png)

Issue #1521, opened 2016-12-22
Handle asyncConnect LOAD_FAIL errors better


kumar303:
### Describe the problem and steps to reproduce it:

* Decorate a component with `asyncConnect`
* Add a callback with a typo that will throw an exception, such as `ReferenceError: store is not defined`

### What happened?

If you have `disableSSR: false` (the default, more like production), the page is a 500 and you see this unrelated error at the bottom of the server log:

````
[1] {"name":"server","app":"amo","hostname":"zenith.local","pid":6592,"level":50,"err":{"message":"Cannot read property 'status' of undefined","name":"TypeError","stack":"TypeError: Cannot read property 'status' of undefined\n    at /Users/kumar/dev/addons-frontend/src/core/server/base.js:213:31\n    at run (/Users/kumar/dev/addons-frontend/node_modules/core-js/modules/es6.promise.js:87:22)\n    at /Users/kumar/dev/addons-frontend/node_modules/core-js/modules/es6.promise.js:100:28\n    at flush (/Users/kumar/dev/addons-frontend/node_modules/core-js/modules/_microtask.js:18:9)\n    at _combinedTickCallback (internal/process/next_tick.js:67:7)\n    at process._tickCallback (internal/process/next_tick.js:98:9)"},"msg":"Cannot read property 'status' of undefined","time":"2016-12-22T16:45:00.746Z","v":0}
```` 

The actual error was dispatched via redux as `LOAD_FAIL` and is buried in the log output:
````
[1] %c action color: #03A9F4; font-weight: bold { type: '@redux-conn/LOAD_FAIL',
[1]   payload:
[1]    { key: 'initialData',
[1]      error:
[1]       ReferenceError: store is not defined
[1]           at /Users/kumar/dev/addons-frontend/src/amo/components/AddonReviewList.js:88:27
[1]           at run (/Users/kumar/dev/addons-frontend/node_modules/core-js/modules/es6.promise.js:87:22)
[1]           at /Users/kumar/dev/addons-frontend/node_modules/core-js/modules/es6.promise.js:100:28
[1]           at flush (/Users/kumar/dev/addons-frontend/node_modules/core-js/modules/_microtask.js:18:9)
[1]           at _combinedTickCallback (internal/process/next_tick.js:67:7)
[1]           at process._tickCallback (internal/process/next_tick.js:98:9) } }
````

[...]

PR #1551, opened 2016-12-31
Add featured search page

tofumatt:
Fixes https://github.com/mozilla/addons-frontend/issues/1535

Adds the list of "more featured add-ons", which the mocks want.

Includes a bit of clean-up in its own commit as I didn't see the harm.

### Screenshots
screen shot 2016-12-30 at 17 49 30
screen shot 2016-12-30 at 17 49 14

### Video
dec-30-2016 17-48-55


src/amo/actions/featured.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function loadFeatured({ addonType, entities, result }) {
+  return {
+    type: FEATURED_LOADED,
+    payload: { addonType, entities, result },
+  };
+}
+
+export function failFeatured({ addonType }) {
```
kumar303: Everything related to `FEATURED_FAILED` doesn't seem to be used in any meaningful way. I see that the action gets dispatched when the exception fails but that's it. I say we remove it all for now until we hook it up async-connect errors properly. In other words, it will probably be a YAGNI feature once we address https://github.com/mozilla/addons-frontend/issues/1521

src/amo/utils.js
```diff
@@ -1,8 +1,31 @@
...
 
 
+export function fetchFeaturedAddons({ addonType, api, dispatch }) {
+  dispatch(getFeatured({ addonType }));
+
+  const filters = { addonType, page_size: FEATURED_ADDONS_TO_LOAD };
+
+  return featuredAPI({ api, filters })
+    .then((response) => dispatch(
+      loadFeatured({ addonType, ...response })
```
kumar303: I'd rather see `entities` and `result` instead of `...response`. Since there are only two keys I don't think it saves much to use a spread.

tests/client/amo/components/TestFeaturedAddons.js
```diff
@@ -0,0 +1,79 @@
...
+
+
+describe('<FeaturedAddons />', () => {
+  const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
+
+  function render({ ...props }) {
+    return findDOMNode(findRenderedComponentWithType(renderIntoDocument(
+      <Provider store={createStore(initialState)}>
+        <I18nProvider i18n={getFakeI18nInst()}>
+          <FeaturedAddonsBase i18n={getFakeI18nInst()} {...props} />
```
kumar303: do you need to pass in `i18n` again even though you use the provider?

tests/client/amo/components/TestFeaturedAddons.js
```diff
@@ -0,0 +1,79 @@
...
+    assert.deepEqual(
+      Object.values(root.querySelectorAll('.SearchResult-heading'))
+        .map((heading) => heading.textContent),
+      ['Howdy', 'Howdy again']
+    );
+  });
+
+  it('throws if add-on type is not supported', () => {
+    assert.throws(() => {
+      render({ addonType: 'XUL' });
```
kumar303: ha

tests/client/amo/reducers/testFeatured.js
```diff
@@ -0,0 +1,66 @@
+import * as actions from 'amo/actions/featured';
+import featured, { initialState } from 'amo/reducers/featured';
+
+describe('featured reducer', () => {
+  it('defaults to loading', () => {
```
kumar303: it defaults to loading? That's not what it looks like from the test.

src/amo/actions/featured.js
tofumatt: Fair enough. It at least shows a failure action during development but I guess that's it. I'll remove them, but in their own commit so we can always easily fetch them if wanted. 😄

src/amo/components/FeaturedAddons/index.js
```diff
@@ -0,0 +1,63 @@
...
+        </h2>
+        <SearchResults count={results.length} hasSearchParams={hasSearchParams}
+          loading={loading} results={results} />
+      </div>
+    );
+  }
+}
+
+export function mapStateToProps(state) {
+  return { hasSearchParams: true, ...state.featured };
```
tofumatt: Totally, good call.

tests/client/amo/components/TestFeaturedAddons.js
tofumatt: No, this started using `FeaturedAddonsBase` but could use `FeaturedAddons` without the `i18n` prop. We use the providers because we're rendering subcomponents and it's nice to check their data.

tests/client/amo/reducers/testFeatured.js
tofumatt: But fortunately I only need to change the text, as it shouldn't default to loading 😄

tests/client/amo/components/TestFeaturedAddons.js
tofumatt: Actually using `shallowRender` here might be doable and would clean this up, as removing the `i18n` prop here breaks stuff but including `FeaturedAddons` instead of `FeaturedAddonsBase` means passing stuff into the `<Provider>` from each test and is gross.

I'll try to shallowrender it if it's easy.

tofumatt: I moved it to `shallowRender`, which arguably makes the properties we're checking a bit less obvious but the tests are less brittle and way fewer lines so 👍

tofumatt:
All changes addressed, should be good for another r? (Sorry for the delay on getting fixes in on this one!)

src/amo/components/FeaturedAddons/index.js
```diff
@@ -0,0 +1,66 @@
...
+          loading={loading} results={results} />
+      </div>
+    );
+  }
+}
+
+export function mapStateToProps(state) {
+  return {
+    addonType: state.featured.addonType,
+    hasSearchParams: true,
```
kumar303: could this just be set in `defaultProps` ?

src/amo/components/SearchPage.js
```diff
@@ -27,20 +22,19 @@ export default class SearchPage extends React.Component {
...
     const {
-      LinkComponent, ResultComponent, count, filters, hasSearchParams,
-      loading, page, pathname, results,
+      LinkComponent, count, filters, hasSearchParams, loading, page, pathname,
+      results,
     } = this.props;
     const queryParams = this.props.queryParams ||
       convertFiltersToQueryParams(filters);
-    const paginator = count && hasSearchParams > 0 ? (
+    const paginator = count > 0 && hasSearchParams ? (
```
kumar303: since `count` is always used like a number, maybe make the property default to 0 for sanity?

tests/client/amo/components/TestFeaturedAddons.js
```diff
@@ -0,0 +1,76 @@
...
+  mapStateToProps,
+} from 'amo/components/FeaturedAddons';
+import createStore from 'amo/store';
+import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import { fakeAddon } from 'tests/client/amo/helpers';
+import { getFakeI18nInst, shallowRender } from 'tests/client/helpers';
+
+
+describe('<FeaturedAddons />', () => {
+  const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
```
kumar303: This should use [signedInApiState](https://github.com/mozilla/addons-frontend/blob/70f359d87cb70e2ecf2edf905ffedbb3e661e778/tests/client/amo/helpers.js#L45) like `{ api: signedInApiState }`

tests/client/amo/components/TestFeaturedAddons.js
```diff
@@ -0,0 +1,76 @@
...
+
+  function render({ ...props }) {
+    return shallowRender(
+      <FeaturedAddonsBase i18n={getFakeI18nInst()} {...props} />
```
kumar303: If you did `<FeaturedAddons store={store} ...>` (i.e. use the default component, not the base) won't that invoke `mapStateToProps` automatically?

tests/client/amo/components/TestFeaturedAddons.js
```diff
@@ -0,0 +1,76 @@
...
+    );
+  }
+
+  it('renders a FeaturedAddons page with no add-ons set', () => {
+    const store = createStore(initialState);
```
kumar303: it looks like `store =` could be moved up to a `beforeEach()`

tests/client/amo/helpers.js
```diff
@@ -12,6 +12,7 @@ export const fakeAddon = {
     id: 123,
     license: { name: 'tofulicense', url: 'http://license.com/' },
     version: '2.0.0',
+    files: [],
```
kumar303: Ha. How did this work before? I just added it too. https://github.com/mozilla/addons-frontend/pull/1586/files#diff-2e76cee7f42cc26c94ef51183f1b5f86R16

tests/client/amo/test_utils.js
```diff
@@ -1,15 +1,50 @@
...
+import { loadFeaturedAddons, loadLandingAddons } from 'amo/utils';
 
 
 describe('amo/utils', () => {
+  describe('loadFeaturedAddons()', () => {
+    const addonType = ADDON_TYPE_EXTENSION;
+    let ownProps;
+
+    beforeEach(() => {
+      ownProps = {
```
kumar303: It doesn't look like this is shared by multiple tests anymore. I suggest moving it closer to the source for easier readability until it needs to be shared.

src/amo/utils.js
```diff
@@ -1,10 +1,36 @@
...
+  LANDING_PAGE_ADDON_COUNT,
+} from 'amo/constants';
+import { getFeatured, loadFeatured } from 'amo/actions/featured';
 import { getLanding, loadLanding, failLanding } from 'amo/actions/landing';
 import { featured as featuredAPI, search } from 'core/api';
 import { SEARCH_SORT_POPULAR, SEARCH_SORT_TOP_RATED } from 'core/constants';
 import { apiAddonType } from 'core/utils';
 
 
+export function fetchFeaturedAddons({ addonType, api, dispatch }) {
```
kumar303: I don't understand why you need this as a separate function. Can't it just be part of `loadFeaturedAddons` ? Maybe it will become more clear in the next patch?

tests/client/amo/test_utils.js
```diff
@@ -1,15 +1,50 @@
...
+      const entities = sinon.stub();
+      const result = { results: [] };
+
+      mockApi
+        .expects('featured')
+        .once()
+        .withArgs({ api: {}, filters: { addonType, page_size: 25 } })
+        .returns(Promise.resolve({ entities, result }));
+
+      return loadFeaturedAddons({ store, params: ownProps.params })
```
kumar303: This looks like it should be a test for `fetchFeaturedAddons` instead since it sets up mock expectations for the API call. I don't see `fetchFeaturedAddons` tested directly though and I commented above that you could just merge it into `loadFeaturedAddons` anyway and be done (unless I'm missing something).

src/amo/utils.js
tofumatt: I guess it can be part of it. This was just convention in previous similar utils functions (categories and search) and the smaller functions make it easier to test and such. I like the small functions personally, but if you'd rather it all together I can put them together.

Though since I'm not discretely testing it as you mentioned, I'll either do that or merge them; no use doing it halfway!

tests/client/amo/test_utils.js
tofumatt: 👍

tests/client/amo/helpers.js
tofumatt: Right?!

src/amo/components/SearchPage.js
tofumatt: Good call.

src/amo/components/FeaturedAddons/index.js
tofumatt: Thanks, indeed!

tests/client/amo/components/TestFeaturedAddons.js
tofumatt: Because I'm using `shallowRender` (to avoid having to add `<Provider>`s and render a bunch of children), that seems to not work. I get errors like:

```
✖ renders a FeaturedAddons page with themes HTML
      Firefox 50.0.0 (Mac OS X 10.12.0)
    root.props.children is undefined
    @webpack:///tests/client/amo/components/TestFeaturedAddons.js:44:4 <- test-runner.js:126801:6
```

tofumatt:
I think I addressed the comments; I tested `fetchFeaturedAddons` discretely though the test for `loadLandingAddons` is similar as when I tried to mock `fetchFeaturedAddons` using DI it got weird. I'll investigate more if you like though it's similar to other tests of similar methods...

If you find the methods should be combined I'll just do that. Let me know.

tests/client/amo/test_utils.js
```diff
@@ -1,29 +1,81 @@
...
+  beforeEach(() => {
+    ownProps = {
+      params: {
+        application: 'android',
+        visibleAddonType: 'extensions',
+      },
+    };
+  });
+
+  describe('fetchFeaturedAddons()', () => {
```
kumar303: This test is exactly the same as `loadFeaturedAddons()`. I think it would make your life way easier if you merge the code for `fetchFeaturedAddons` into `loadFeaturedAddons` and only keep the test for `loadFeaturedAddons`. I know you mentioned that it follows a convention but it doesn't seem helpful in this context.

tofumatt: Yeah, I see the light 😉 👍

tofumatt:
Combined the functions, so should be good now 😄

Thanks for being patient while I slowly worked through all the issues 😆

tests/client/amo/helpers.js
kumar303: You may want to merge master into your branch just to check that this change merges cleanly. However, if github isn' complaining it's probably fine.

tofumatt: I rebased master onto this and pushed it here. Rebased cleanly so I think git was smart enough to realise we did the same thing ^_^

PR #1554, opened 2017-01-01
Add overlay and card overlay components

### LINKED ISSUE #1553: Add overlay and card overlay components
For #1526, we should have generic UI components that overlay contents and can be hidden/shown on command. Specific to #1526 we'll want an overlay that uses the `<Card>` component.

I've got one built already (`<Overlay>`) and a `<CardOverlay>` in the works.

tofumatt:
Adds a very generic overlay and more useful `OverlayCard` for the restyle of the reviews (and probably other things too, this seemed a useful thing to make generic).

The component automatically stretches to fill the screen and has a max height of 75% of the viewport.

Fixes #1553.

### Screenshot

screenshot 2017-01-01 16 50 37

### DIFF
--- a/src/ui/components/Overlay/Overlay.scss
+++ b/src/ui/components/Overlay/Overlay.scss
@@ -21,3 +21,15 @@
   width: 100%;
   z-index: 1000;
 }
+
+.Overlay-contents {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  left: 0;
+  position: fixed;
+  top: 50%;
+  transform: translate(0, -50%);
+  width: 100%;
+  z-index: 1001;
+}
--- a/src/ui/components/Overlay/Overlay.scss
+++ b/src/ui/components/Overlay/Overlay.scss
@@ -1,0 +1,23 @@
+@import "~core/css/inc/vars";
+
+.Overlay {
+  display: none;
+}
+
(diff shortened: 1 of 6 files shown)


kumar303:
Should this description link to https://github.com/mozilla/addons-frontend/issues/1553 ?

tofumatt:
> Should this description link to #1553 ?

I put it in the commit message but yes! Added.

kumar303:
> I put it in the commit message but yes! Added.

Because github is buggy, I think it's better to leave issue numbers out of commit messages and always put them in the PR description instead. Here are some reasons:
* Every time you push a new commit to your branch (or rebase), the issue gets updated with a link to the commit which creates noise
* When we squash and merge we might lose one of the commits that mention the issue which means it may not get auto-closed

tofumatt:
Yeah, I’ve been trying to do that as you mentioned before. I’m still trying
but I forget a lot.

tofumatt:
Yeah, I figured it could always be folded into the review, but I thought this was totally the kind of component we'd want again so I figured why not.

kumar303:
ok, feel free to merge it in and we'll build on top of it. I don't know if we'll need all the methods in here but it will be easier to figure that out once we build something concrete (like the review form) with it.

PR #1566, opened 2017-01-05
Allow scrolling when previewing theme

### LINKED ISSUE #1445: Don't preview a theme if the user wants to scroll
Tap the theme header image and try to scroll. The theme is previewed, scrolling does not happen.

### What happened?
The theme header touch events have prevented default so that the user isn't shown a context menu when long pressing but this cancels scrolling too.

If I scroll quickly I should get scrolling. If I keep my finger within the theme header image for a few hundred ms I should get a preview.

theme-preview-finalish mp4

Extracted from #1422 and #1442.

mstriemer:
This fixes the not being able to scroll while previewing a theme bug by replacing the `` with a `<div>` that has a background image once the component mounts. This is done to avoid the CSP errors that we'd get for rendering on the server with inline styles.

> [div version?](https://cloud.githubusercontent.com/assets/211578/21670248/7beff1c8-d2d8-11e6-85e6-bc527e1abc97.png)
> [img version?](https://cloud.githubusercontent.com/assets/211578/21670249/7c11d0f4-d2d8-11e6-8889-0fc127c4cff0.png)
> Screenshots of the two versions, I forget which is which but they look the same

theme-preview-with-scrolling mp4

Fixes #1445.

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -47,10 +47,19 @@ export class AddonDetailBase extends React.Component {
     OverallRating: DefaultOverallRating,
   }
 
+  constructor(props) {
+    super(props);
+    this.state = { mounted: false };
+  }
+
+  componentDidMount() {
+    // Disabling react/no-did-mount-set-state because it is to prevent additional renders, but
+    // that's exactly what we want in this case (use background-image with JS).
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ mounted: true });
+  }
+
   onTouchStart = (event) => {
-    // We preventDefault so that the image long-press menu doesn't appear.
-    // This unfortunately prevents scrolling :(
-    event.preventDefault();
     this.props.previewTheme(event.currentTarget);
   }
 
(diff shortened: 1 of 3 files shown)


src/amo/components/AddonDetail.js
```diff
@@ -47,10 +47,19 @@ export class AddonDetailBase extends React.Component {
...
+  constructor(props) {
+    super(props);
+    this.state = { mounted: false };
+  }
+
+  componentDidMount() {
+    // Disabling react/no-did-mount-set-state because it is to prevent additional renders, but
+    // that's exactly what we want in this case (use background-image with JS).
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ mounted: true });
```
kumar303: Can you help me understand why this is needed? Why is the `` tag needed at all? For server side rendering?

mstriemer: Yeah, the `` tag is rendered on the server since we can't put `style='background-image: url("/path/to/previewURL.png")'` in the HTML served from the server.

Since long pressing on an image pops up the context menu (and it looks like disabling that isn't necessarily cross-browser and can be prevented by the user) this switches it to a `<div>` with a background image when the page loads. The background image won't open the context menu and I disabled text selection (which appears to be well supported).

kumar303: Interesting, I never thought about how SSR never mounts any component. I suppose that makes sense!

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -223,7 +228,31 @@ describe('AddonDetail', () => {
...
+    const rootNode = findDOMNode(root);
+    root.setState({ mounted: false });
+
+    let image = rootNode.querySelector('.AddonDetail-theme-header-image');
+    assert.equal(image.tagName, 'IMG');
+    assert.ok(image.classList.contains('AddonDetail-theme-header-image'));
+    assert.equal(image.src, 'https://amo/preview.png');
+    assert.equal(image.alt, 'Press to preview');
+
+    root.componentDidMount();
```
kumar303: this seems a little redundant to me since you already have a "renders ...as a background image when mounted" test

mstriemer: Yeah, it's a bit odd. I wanted to test that `componentDidMount()` is what's switching it though. I'll pull this out and add a test that just looks at `mounted`.

PR #1569, opened 2017-01-06
White text in search placeholder

mstriemer:
Make the text white, as it is in the mocks.

screen shot 2017-01-05 at 22 49 25

### DIFF
--- a/src/ui/components/SearchInput/style.scss
+++ b/src/ui/components/SearchInput/style.scss
@@ -43,7 +43,6 @@ $transition-duration: 250ms;
   }
 
   .SearchInput-animation-icon {
-    color: #fff;
     opacity: 1;
     // Use !important to overwrite style properties added in JS.
     transform: translateX(0) !important;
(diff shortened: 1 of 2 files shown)


src/ui/components/SearchInput/style.scss
```diff
@@ -56,7 +56,7 @@ $transition-duration: 250ms;
   // Include the start margin on the end so that icon and text is centred together.
   @include margin-end($icon-margin-start);
 
-  color: #ccc;
+  color: #fff;
```
tofumatt: Should we use `$base-color` here instead? `$base-color` is `#fff` and presumably we'd want to update most #fff's across the board as our "base white" if/when we changed this.

mstriemer:
I removed the styles for `color` from SearchInput/styles.scss since they are specific to the header. The only `color` styles in there now are to set the label/placeholder to be transparent when focused. The SearchForm now sets the styles as that's the place we'd like to change the colour. Theoretically if you used this elsewhere your regular `input` styles would kick in now and it should just work.

Issue #1293, opened 2016-10-25
Show user count data on add-on detail page


muffinresearch:
For V1, we will only implement the user count metadata. Note this likely shares data with the "more information widget"

See https://docs.google.com/document/d/13g8tmoII8oCNrUXOl2HegsQst7WEsXukBnIqPUNrqb0/edit#heading=h.2ltzvz63pc6h

![amo_-_v1_sketch](https://cloud.githubusercontent.com/assets/1514/19686010/4f2b7cb0-9ab7-11e6-95fa-13ae37e86ca5.png)

mstriemer:
I don't think we have any way of detecting "Editor's Choice". We could presumably show currently featured add-ons but I don't think that data is currently surfaced in the API.

For downloads the value we have on mobile currently and in the API is "weekly downloads".

It might be nice to include the star rating up top, not sure if we can reuse the star icon from the rating widget but that might work.

kumar303:
According to the latest comment on the doc linked above, all of this has been pushed to V2 *except* the user count metadata. We need to show the user count for V1. I will update the title of this issue.

ValentinaPC:
Verified as fixed on AMO-dev FF51(Android 6.0.1).
User count data is displayed in add-on details page.
Postfix screenshot: 
![u1](https://cloud.githubusercontent.com/assets/15685960/22686450/e07272b0-ed2c-11e6-821c-8f9f34e59f56.png)

PR #1579, opened 2017-01-06
Show average daily users on add-on page

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1293

screenshot 2017-01-06 16 04 00

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -146,7 +146,7 @@ export class AddonDetailBase extends React.Component {
           <h2 className="visually-hidden">
             {i18n.gettext('Extension Metadata')}
           </h2>
-          <AddonMeta />
+          <AddonMeta averageDailyUsers={addon.average_daily_users} />
           <InstallButton {...this.props} />
         </section>
 
--- a/tests/client/amo/components/TestAddonDetail.js
+++ b/tests/client/amo/components/TestAddonDetail.js
@@ -326,11 +326,9 @@ describe('AddonDetail', () => {
     assert.ok(rootNode.querySelector('.AddonMoreInfo-contents'));
   });
 
-  it('renders meta data', () => {
-    const root = render({ addon: fakeAddon });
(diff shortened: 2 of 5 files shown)


tests/client/amo/components/TestAddonDetail.js
```diff
@@ -324,4 +325,12 @@ describe('AddonDetail', () => {
 
     assert.ok(rootNode.querySelector('.AddonMoreInfo-contents'));
   });
+
+  it('renders meta data', () => {
+    const root = render({ addon: fakeAddon });
```
mstriemer: Mind testing the user count directly? Either check that it's `100` since that's the value in `fakeAddon` or explicitly define it (I prefer the latter).

```js
const addon = { ...fakeAddon, average_daily_users: 25 };
// ...
assert.equal(metaData.props.averageDailyUsers, 25);

File: src/amo/components/AddonDetail.js [hunk 2]
```diff
@@ -146,7 +146,7 @@ export class AddonDetailBase extends React.Component {
           <h2 className="visually-hidden">
             {i18n.gettext('Extension Metadata')}
           </h2>
-          <AddonMeta />
+          <AddonMeta averageDailyUsers={addon.average_daily_users} />
```
kumar303: I think it's ok as a separate component because it connects to the `api.lang` state which pinpoints the exact place we'd need to do a re-render when that value changes. Plus, we will eventually put more metadata in there.

tests/client/amo/components/TestAddonDetail.js
kumar303: Oh, sure, then I wouldn't need the dumb `assert.ok()`

mstriemer: I totally missed that. I still prefer an explicit value but good to see that's covered.

src/amo/components/AddonDetail.js
mstriemer: Ah, yeah that's cool.

PR #1583, opened 2017-01-09
Fix camelcase and bail on build failure

### LINKED ISSUE #1575: Minify JS bundle
Our JS bundle isn't being minified in production. Gzip is likely making this not terrible but we can likely squeeze out some more bytes by minifying.

Current disco pane js: https://addons-discovery.cdn.mozilla.net/disco-6fc7e9484a5f0795d9fa.js

muffinresearch:
Fixes #1575

This fixes the build to blow up on error. See the comments in the source since this required an extra hook to ensure uglify errors cause a build to fail.

Interestingly even if we'd have been using `--bail` it wouldn't have failed with the uglify error.

If you want to try the first issue, pull just the first commit and see the build fail.

Second commit replaces camelcase with humps.


package.json
```diff
@@ -26,7 +26,7 @@
   },
   "betterScripts": {
     "build": {
-      "command": "npm run clean && npm run version-check && webpack --verbose --display-error-details --progress --colors --config webpack.prod.config.babel.js",
+      "command": "npm run clean && npm run version-check && webpack --bail --verbose --display-error-details --progress --colors --config webpack.prod.config.babel.js",
```
kumar303: Weird that bailing wasn't the default behavior.

muffinresearch: Yeah that is odd and somewhat unexpected.

mstriemer:
While there's a build related issue open I see this in my build output but can't tell if it's an issue or not:

```
 69% 1074/1089 build modulesPostCSS Config could not be loaded. Please check your PostCSS Config.
 68% 1074/1106 build modulesPostCSS Config could not be loaded. Please check your PostCSS Config.
 67% 1079/1127 build modulesPostCSS Config could not be loaded. Please check your PostCSS Config.
 69% 1124/1134 build modulesPostCSS Config could not be loaded. Please check your PostCSS Config.
 69% 1147/1152 build modulesPostCSS Config could not be loaded. Please check your PostCSS Config.

muffinresearch:
Yeah I think that's because disco doesn't require the use of postCSS, so I think that's likely a non-issue.

PR #1584, opened 2017-01-09
Show install switch by default

### LINKED ISSUE #1568: On loading disco pane big blue buttons appear
Open disco pane and watch it load. Upon loading you will see big blue buttons first that then disappear and the actual on/off install toggles appear. (see gif)
disco-pane-load-20170105

I did expect no blue buttons to appear at first, but only see the toggles.

It is confusing to see something appear only to be gone before I know what it is/was.

mstriemer:
This makes the install buttons default to the switch when rendering on the server then flips to the blue button if `mozAddonManager` isn't supported once the page loads.

No support for `<noscript>` yet (I have it working in dev but need to get around CSP still for production).

disco-mozaddonmanager mov
> With `mozAddonManager`

disco-no-mozaddonmanager mov
> No `mozAddonManager`

Fixes #1568.

### DIFF
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -10,7 +10,10 @@ import { addQueryParams } from 'core/utils';
 
 
 export function hasAddonManager({ navigator } = {}) {
-  return typeof window !== 'undefined' && 'mozAddonManager' in (navigator || window.navigator);
+  if (!navigator && typeof window === 'undefined') {
+    return undefined;
+  }
+  return 'mozAddonManager' in (navigator || window.navigator);
 }
 
 export function getAddon(guid, { _mozAddonManager = window.navigator.mozAddonManager } = {}) {
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -9,8 +9,10 @@ import {
 import { addQueryParams } from 'core/utils';
 
 
-export function hasAddonManager({ navigator } = {}) {
-  if (!navigator && typeof window === 'undefined') {
+const testHasWindow = () => typeof window !== 'undefined';
+
+export function hasAddonManager({ hasWindow = testHasWindow, navigator } = {}) {
+  if (!navigator && !hasWindow()) {
     return undefined;
   }
   return 'mozAddonManager' in (navigator || window.navigator);
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -12,6 +12,8 @@ import { addQueryParams } from 'core/utils';
 const testHasWindow = () => typeof window !== 'undefined';
 
 export function hasAddonManager({ hasWindow = testHasWindow, navigator } = {}) {
+  // Returns undefined if it cannot be determined if mozAddonManager is supperted (likely server
+  // rendering with no window). Otherwise returns true/false based on mozAddonManager in navigator.
   if (!navigator && !hasWindow()) {
     return undefined;
   }
--- a/src/core/components/InstallButton/styles.scss
+++ b/src/core/components/InstallButton/styles.scss
@@ -1,15 +1,16 @@
 .InstallButton {
   .InstallButton-button {
-    display: none;
     white-space: nowrap;
   }
 
   &.InstallButton--use-button {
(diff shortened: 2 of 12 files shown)


mstriemer:
Build should be passing now.

src/core/addonManager.js
```diff
@@ -9,8 +9,13 @@ import {
 import { addQueryParams } from 'core/utils';
 
 
-export function hasAddonManager({ navigator } = {}) {
-  return typeof window !== 'undefined' && 'mozAddonManager' in (navigator || window.navigator);
+const testHasWindow = () => typeof window !== 'undefined';
+
+export function hasAddonManager({ hasWindow = testHasWindow, navigator } = {}) {
+  if (!navigator && !hasWindow()) {
+    return undefined;
```
muffinresearch: Any reason to return undefined instead of `false`?

src/core/addonManager.js
```diff
@@ -9,8 +9,13 @@ import {
...
+  }
+  return 'mozAddonManager' in (navigator || window.navigator);
```
muffinresearch: It feels like the `if` and this line could be combined as a single return - although that would make for a longer line...

src/core/components/InstallSwitch/index.js
```diff
@@ -101,7 +101,7 @@ export class InstallSwitchBase extends React.Component {
 
   render() {
     const browsertheme = JSON.stringify(getThemeData(this.props));
-    const { handleChange, slug, status } = this.props;
+    const { handleChange, slug, status, ...props } = this.props;
```
muffinresearch: `props` seems a bit vague, would something like `otherProps` be a bit more representative?

src/core/addonManager.js
mstriemer: So you can tell the difference between "not supported" and "we don't know if it's supported".

src/core/addonManager.js
mstriemer: I don't think so since we want the if to return `undefined` not a boolean.

src/core/addonManager.js
muffinresearch: Ah ok, it might be good to comment the nuance here.

src/core/addonManager.js
muffinresearch: Yep, that all makes sense now based on the previous comment.

mstriemer:
@kumar303 we might include this in the push tomorrow if it's ready. You can see how it looked before in the linked issue.

src/core/components/InstallButton/styles.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  &.InstallButton--use-button {
+    .InstallButton-switch {
+      display: none;
+    }
+  }
+
+  &.InstallButton--use-button {
+    .InstallButton-button {
```
kumar303: you can collapse this into the `&.InstallButton--use-button` block up above

tests/client/disco/components/TestAddon.js
```diff
@@ -48,15 +48,9 @@ function renderAddon({ setCurrentStatus = sinon.stub(), ...props }) {
...
-    let root;
-
-    beforeEach(() => {
-      root = renderAddon(result);
-    });
-
     it('renders a default error overlay with no close link', () => {
       const data = { ...result, status: ERROR, setCurrentStatus: sinon.stub() };
-      root = renderAddon(data);
+      const root = renderAddon({ addon: data, ...data });
```
kumar303: All of these seem like evidence to suggest that `addon` doesn't need to be a property on `<Addon>` but that's an exercise for a future patch

mstriemer: There's an issue filed for the opposite. The data from `addon` is used but it's being spread into the rest of the props as well. I think it would be better to have the addon props in `addon` instead.

kumar303: Yeah, seems fine. Sometimes I get nervous about how we rely on returned API results a little too implicitly though.

Issue #1287, opened 2016-10-25
Implement the all reviews page


muffinresearch:
See the sketch file for more details:

![amo_-_v1_sketch](https://cloud.githubusercontent.com/assets/1514/19685654/93a17bc6-9ab5-11e6-8c6d-cf8c70781e41.png)

vcarciu:
Verified as fixed : 
![reviews](https://cloud.githubusercontent.com/assets/1583842/22337699/44509be2-e3ee-11e6-9eee-012ec016e95f.jpg)

Issue #1104, opened 2016-09-22
Test all React components with Enzyme


kumar303:
It may be worth trying out [Enzyme](http://airbnb.io/enzyme/) to replace `Simulate.click()` and other 'react-addons-test-utils' things. A few things jump out from reading the docs:
- Shallow rendering looks like it works better
- It looks like event simulations are implemented better
- finding nested components looks easier

tofumatt:
The "future" section seems especially promising if I'm reading it right; they want to focus on better event propagation in shallowly-rendered components. I'm hoping that means things like "clicking a button or enter in an `<input>` will fire a `<form>`'s `onSubmit` event", but maybe I'm being too optimistic.

muffinresearch:
Fixing this should also mean removing all of our home-grown shallow rendering functionality.

tofumatt:
> Fixing this should also mean removing all of our home-grown shallow rendering functionality.

I ended up adding Enzyme easily then seeing about moving all our shallow rendering tests over. Turns out it was pretty easy, so I've filed #2539 and have a PR incoming.

kumar303:
Here's how to find all the remaining tests that need converting to Enzyme: https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=renderIntoDocument&type=

kumar303:
Only 32 tests left 😭

muffinresearch:
Now there's only 20 left.

willdurand:
Only 5 files (excluding two files scheduled for deletion):

- `tests/unit/core/testErrorHandler.js`
- `tests/unit/amo/components/TestRatingManager.js`
- `tests/unit/core/components/TestInfoDialog.js`
- `tests/unit/ui/components/TestOverlayCard.js`
- `tests/unit/ui/components/TestRating.js`

AlexandraMoga:
@willdurand could you please add a "qa:not needed" label?
Thanks

PR #1586, opened 2017-01-09
Show all add-on reviews

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1287

First, the detail page has a link to all reviews. This will change once we have the 'helpful reviews' section.
screenshot 2017-01-13 20 19 32

Here are all the reviews. The usernames look a little funky but in practice these will be what the user has set their display name to. If they don't set a display name it will fall back to their email. I think the prompt for display name is pretty good so the majority should have them. Also, in an upcoming patch, all reviews will have a title.
screenshot 2017-01-13 19 49 27


kumar303:
@mstriemer sorry this is so massive! Thanks for digging in. I didn't have time to split it apart but it's pretty much all for the same feature anyway.

kumar303:
@mstriemer the conflicts with your pwalm-mock updates should be resolved now

src/amo/api/index.js
```diff
@@ -41,6 +41,25 @@ export function submitReview({
...
+    }
+    resolve(callApi({
+      endpoint: `addons/addon/${addonSlug}/reviews`,
+      method: 'GET',
+    }));
+  })
+    .then((response) => {
+      // TODO: implement paging through response.next
+      if (response.next) {
+        throw new Error('paging is not yet implemented');
```
kumar303: Yes it would because otherwise the additional reviews would be hidden. I feel like paging should probably be a release blocker but I didn't want to wedge it in here since it could get complex with recursion.

mstriemer: Isn't not having a pager enough though? When testing this against dev it was a bit annoying to have to find an add-on with few enough reviews to fit on one page.

src/amo/components/AddonDetail.js
```diff
@@ -113,6 +114,31 @@ export class AddonDetailBase extends React.Component {
...
+        <Link className="AddonDetail-all-reviews-link"
+          to={`/addon/${addon.slug}/reviews/`}>
+          {linkText}
+        </Link>
+      );
+    } else {
+      content = <span>{i18n.gettext('No reviews yet')}</span>;
+    }
+
+    return <div className="AddonDetail-read-reviews-footer">{content}</div>;
```
kumar303: I was able to remove the span but the div is still needed for testing -- I need a css class to access the footer content.

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+  renderReview(review) {
+    return (
+      <li className="AddonReviewList-li">
+        <h3>{review.title}</h3>
+        <p>
+          {review.body}
+        </p>
+        <div className="AddonReviewList-by-line">
+          <Rating rating={review.rating} readOnly />
+          {this.props.i18n.gettext('from')} {review.userName},
```
kumar303: oh yeah, duh

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+  render() {
+    const { i18n, initialData } = this.props;
+    if (!initialData) {
+      // TODO: add a spinner
+      return <div>{i18n.gettext('Loading...')}</div>;
+    }
+
+    const { reviews, addon } = initialData;
+    const subTitle = i18n.sprintf(
+      i18n.gettext('for %(addonName)s'), { addonName: addon.name });
```
kumar303: sure, doesn't bother me either way. I removed the leading 'for'

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+}
+
+export function loadAddonReviews({ addonSlug, dispatch }) {
+  return getAddonReviews({ addonSlug })
+    .then((allReviews) => {
+      // Ignore reviews with null bodies as those are incomplete.
+      // For example, the user selected a star rating but hasn't submitted
+      // review text yet.
+      const reviews = allReviews.filter((review) => Boolean(review.body));
```
kumar303: `(review) => review.body` was just confusing to me. I actually deleted it because I thought it was a mistake left over from refactoring. I think `Boolean()` makes it more clear that it's testing truthiness and honestly I'm not sure if checking null is totally the right thing to do. This just leaves the code unchanged as it was before but makes it more explicit in checking truthiness.

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+      throw new Error('missing URL param slug (add-on slug)');
+    }
+    resolve();
+  })
+    .then(() => Promise.all([
+      _loadAddonReviews({ addonSlug: slug, dispatch: store.dispatch }),
+      loadAddonIfNeeded({ store, params }),
+    ]))
+    .then(([reviews]) => {
+      const addon = findAddon(store.getState(), slug);
```
kumar303: Yes! But `loadAddonIfNeeded` was written to be very inconsistent. It sometimes returns the add-on and sometimes loads it into state. Rather than shave a bunch of yaks trying to convert all code that calls it (which is a lot), I just oped for the simpler path to fetch from state, which will work in every case. This pattern of inconsistent loader return values is bad though and I hate `asyncConnect` for encouraging it.

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+      const initialData = { addon, reviews };
```
kumar303: I hate asyncConnect so much. I find the magical conversion of `key` into properties very confusing so I wanted to assign it here explicitly to make it a little bit more clear.

src/amo/components/AddonReviewList.js
```diff
@@ -0,0 +1,119 @@
...
+      return initialData;
+    });
+}
+
+export default compose(
+  asyncConnect([{
+    key: 'initialData',
```
kumar303: Ah, right, I forgot about this. I should try and fix it to pull everything from state. The problem I ran into was that `slug` only gets passed via URL params which is only available in the `asyncConnect` callback. But let me see if I can figure out another way. I hate the magic `asyncConnect` key thing but we do use it in a few places.

src/amo/css/AddonReviewList.scss
```diff
@@ -0,0 +1,91 @@
...
+  .AddonReviewList-by-line {
+    display: flex;
+    justify-content: flex-start;
+    font-size: 0.9em;
+    color: $sub-text-color;
+  }
+
+  .Rating {
+    margin: 0 5px 0 0;
+    min-width: ($star-size + $star-gutter) * 5;
```
kumar303: I thought so at first but the stars got truncated. On second look, this was some kind of auto-collapsing or borders that `float` was doing. I switched it to flex and your math is working now. I am also moving all of these into Rating.

src/amo/api/index.js
kumar303: hmm, yeah, I suppose so. I'll change it to a warning in the console

src/core/css/inc/vars.scss
```diff
@@ -7,6 +7,7 @@ $header-font-color: #fff;
 $text-color-default: #333;
 $text-color-headings: $text-color-default;
 $text-color-message: #fff;
+$sub-text-color: #6a6a6a;
```
kumar303: I fixed the one in `SearchResult.scss`

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -2,16 +2,20 @@ import React from 'react';
 import { findDOMNode } from 'react-dom';
 import {
   Simulate,
+  scryRenderedComponentsWithType,
```
kumar303: I think it's supposed to be "scary." One day maybe we can switch it all to the new hotness. https://github.com/mozilla/addons-frontend/issues/1104

kumar303:
@mstriemer thanks for patiently slogging through that review. Great stuff! I merged a bunch of code/styles with `SearchResult` using a new shared `CardList` and also moved a bunch of styles to the Rating widget. Could you take another look?

src/amo/components/AddonReviewList.js
mstriemer: It looks like we've got a few instances of `!!thingToConvert` and one other instance of `Boolean(thingToConvert)`. I don't really care which we use but it would be good to be consistent. `!!` is definitely shorter but if you've never seen it it's probably confusing.

src/amo/components/AddonDetail.js
mstriemer: You should be able to grab the footer with like `.AddonDetail-overall-rating .Card-footer`.  We don't have a great way of stating that a selector is "public" for a component but I doubt that one would change.

tests/client/amo/components/TestAddonReviewList.js
```diff
@@ -151,53 +168,109 @@ describe('amo/components/AddonReviewList', () => {
...
 
+  describe('mapStateToProps', () => {
+    let store;
+
+    beforeEach(() => {
+      store = createStore();
+    });
+
+    function getMappedProps({
+      addonSlug = fakeAddon.slug, params = { addonSlug },
```
kumar303: Yes, it kind of hurts the brain to read it but beats moving the statement down into the function body and wrapping it in `if (params !== undefined) { ...`

src/amo/components/AddonReviewList.js
```diff
@@ -80,40 +85,37 @@ export function loadAddonReviews({ addonSlug, dispatch }) {
...
+    return Promise.reject(new Error('missing URL param addonSlug'));
+  }
+  return Promise.all([
+    loadAddonReviews({ addonSlug, dispatch: store.dispatch }),
+    loadAddonIfNeeded({ store, params: { slug: addonSlug } }),
+  ]);
+}
+
+export function mapStateToProps(state, ownProps) {
+  if (!ownProps || !ownProps.params || !ownProps.params.addonSlug) {
```
kumar303: Yes it could in a situation where the redux state changes after the page initially loads (e.g. someone submits a review which loads it into state).

src/ui/components/Rating/styles.scss
```diff
@@ -38,3 +35,25 @@ $star-size: 46px;
...
+  background-image: url('./img/yellow-star.svg');
+}
+
+.Rating--small {
+  $star-size: 13px;
+  $star-gutter: 4px;
+
+  .Rating-star-group {
+    display: flex;
+    min-width: $star-size * 5 + $star-gutter * 4;
```
kumar303: ah, I can totally remove `min-width`, thanks. That was leftover from when I was using floats (for whatever reason).

kumar303:
@mstriemer thanks for slogging through this one

PR #1530, opened 2016-12-22
Support read-only star ratings

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1529

PR #1590, opened 2017-01-10
Clean up Rating widgets

kumar303:
This applies some cleanup to https://github.com/mozilla/addons-frontend/pull/1530


kumar303:
@mstriemer what do you think? I don't want to over-engineer this which is why I moved it back to AMO for now. I doubt we'll ever need a rating manager anywhere else.

PR #1593, opened 2017-01-11
Install button noscript fallback

### LINKED ISSUE #1604: Noscript fallback for install buttons
Add some CSS that will default to the blue install button when the user does not have javascript enabled.

mstriemer:
This will read a `noscript.css` file from the app's root and include that in `<noscript><style>{contents}</style></noscript>` if it exists. It also sets the CSP header so that this will work in production.

I have some updates to make this work with webpack as well but I'm running into some problems with webpack-isomorphic-tools so just putting this up for now to look at.

Fixes #1604.

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,6 @@
     "chalk": "1.1.3",
     "cheerio": "0.22.0",
     "concurrently": "3.1.0",
-    "csp-parse": "0.0.2",
     "css-loader": "0.26.1",
     "deepcopy": "0.6.3",
     "eslint": "3.13.1",
--- a/src/amo/noscript.css
+++ b/src/amo/noscript.css
@@ -1,0 +1,6 @@
+.InstallButton .InstallButton-button {
+  display: inline-block !important;
+}
+.InstallButton .InstallButton-switch {
+  display: none !important;
+}
--- a/src/amo/noscript.css
+++ b/src/amo/noscript.css
@@ -1,6 +1,8 @@
+/* !important is needed to show the button, make sure switch is hidden */
 .InstallButton .InstallButton-button {
   display: inline-block !important;
 }
+
 .InstallButton .InstallButton-switch {
   display: none !important;
 }
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -1,3 +1,4 @@
+import crypto from 'crypto';
 import fs from 'fs';
 import path from 'path';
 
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -213,7 +213,6 @@
     "chalk": "1.1.3",
     "cheerio": "0.22.0",
     "concurrently": "3.1.0",
-    "csp-parse": "0.0.2",
```
mstriemer: `csp-parse` was lower casing the entire CSP header string which was breaking the tests for the SHA since they're case sensitive. There was a PR on the repo with no response from the maintainer. Turns out parsing the header really isn't very hard so I wrote one.

src/amo/noscript.css
```diff
@@ -0,0 +1,6 @@
+.InstallButton .InstallButton-button {
+  display: inline-block !important;
```
tofumatt: Can we have comments next to all `!important` declarations? I find it really helps when doing (inevitable) CSS cleanup and prevents someone trying to be clever and remove the `!important`.

src/amo/noscript.css
```diff
@@ -0,0 +1,6 @@
...
+}
+.InstallButton .InstallButton-switch {
```
tofumatt: I'm assuming this isn't built output; if so can we have a newline here?

src/core/server/base.js
```diff
@@ -41,6 +42,18 @@ const errorPageText = {
   500: 'Internal Server Error',
 };
 
+function getNoScriptStyles({ appName }) {
+  const cssPath = path.join(config.get('basePath'), `src/${appName}/noscript.css`);
+  try {
+    return fs.readFileSync(cssPath);
+  } catch (e) {
+    if (e.code !== 'ENOENT') {
```
tofumatt: Should we at least log that the noscript file doesn't exist as a `debug()` log call, since we're catching an error?

tests/server/helpers.js
```diff
@@ -24,3 +25,14 @@ export function checkSRI(res) {
       'anonymous', 'script should have crossorigin attr');
   });
 }
+
+export function parseCSP(rawCsp) {
+  return camelizeKeys(
```
tofumatt: Nitpick but could we import `camelizeKeys` as `camelCaseKeys`? It feels like a more descriptive function name and camelize isn't a real word.

If you don't want to just ignore this comment, it's fine.

tests/server/helpers.js
```diff
@@ -24,3 +25,14 @@ export function checkSRI(res) {
       'anonymous', 'script should have crossorigin attr');
   });
 }
+
+export function parseCSP(rawCsp) {
```
tofumatt: Cool! Neat to not rely on an unmaintained package then 😄

mstriemer:
Forgot the video before merging. Also added an extra commit with the debug line. Although it doesn't show by default and I'm not sure how that's changed so maybe not too useful...

discopane-noscript mov

PR #1594, opened 2017-01-11
[WIP] Webpack noscript install fallback

mstriemer:
This is still work in progress.

This is the webpack version of #1593. It will build the bundle if you run `npm run build` but not if you run `npm run dev:disco`. I guess there's some magic going on in webpack-isomorphic-tools?

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -31,6 +31,10 @@ module.exports = {
 
   staticHost,
 
+  webpackEntryPoints: {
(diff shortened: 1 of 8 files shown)


mstriemer:
It looks like there is no way to get the CSS bundle out of webpack-isomorphic-tools, it only generates the JS file that will load the CSS, which is a little useless in a `<noscript>` tag.

I filed #1598 to look into moving to a webpack generated bundle once we have moved to webpack 2.

PR #1595, opened 2017-01-11
Update docs on running UI tests

kumar303:
@davehunt does this look right? I was getting `UsageError: --driver must be specified` when running it the old way. This brings it closer to how the `.travis.yml` file runs the tests.

### DIFF
--- a/tests/ui/README.md
+++ b/tests/ui/README.md
@@ -25,7 +25,7 @@ The tests must be run in Firefox 48 or later.
 To run against a different environment, pass in a value for `--base-url`, like so:
 
 ```bash
-PYTEST_ADDOPTS="--base-url https://discovery.addons.mozilla.org" tox -e discopane-ui-tests
+tox -e discopane-ui-tests -- --base-url=https://discovery.addons.mozilla.org
 ```
 
 The pytest plugin that we use for running tests has a number of advanced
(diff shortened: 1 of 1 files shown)


tests/ui/README.md
```diff
@@ -25,7 +25,7 @@ The tests must be run in Firefox 48 or later.
 To run against a different environment, pass in a value for `--base-url`, like so:
 
 ```bash
-$ tox -e discopane-ui-tests -- --base-url https://discovery.addons.mozilla.org
+PYTEST_ADDOPTS="--base-url https://discovery.addons.mozilla.org" tox -e discopane-ui-tests
```
davehunt: The driver argument should be picked up from https://github.com/mozilla/addons-frontend/blob/master/tests/ui/setup.cfg#L2 so this shouldn't be necessary. For what it's worth, f you update to the latest version of pytest-base-url you can just add an environment variable of `PYTEST_BASE_URL` with the value of the URL. Same goes for Travis.

davehunt: Ah, my guess is that the space between `--base-url` and `https://discovery.addons.mozilla.org` is causing path confusion when locating `setup.cfg`. You could also try replacing the space with an `=`.

kumar303:
Thanks @davehunt , adding the equal sign did fix it. That's kinda weird though.

PR #1596, opened 2017-01-11
Try to get UI tests working

kumar303:


kumar303:
ok, so, we needed a PostCSS config file for some reason. Weird.

kumar303:
@muffinresearch I merged this to get the UI tests working ASAP but could you review it? Do you think the empty config will conflict with [settings.postcss](https://github.com/mozilla/addons-frontend/blob/master/webpack.prod.config.babel.js#L142) in the prod config?

Issue #1391, opened 2016-11-18
Deploy dependencies to production with yarn


kumar303:
### Describe the problem and steps to reproduce it:

When we deploy AMO/discopane to production the install step uses `npm`

### What happened?

This installs all dependencies as it should but it does not verify the contents of each package.

### What did you expect to happen?

We should switch to [yarn](https://yarnpkg.com/en/docs/install#linux-tab) which will verify the contents of each package we install. This will make sure none of them have been tampered with

### Anything else we should know?

To bootstrap our system, i.e. install `yarn` itself, we can use their RPM: https://yarnpkg.com/en/docs/install#linux-tab

This issue should also involve updating all the developer to documentation so that devs all use `yarn` locally.

mstriemer:
We should also add a test that verifies there are no changes to `yarn.lock` when installing on CI.

This should be as easy as checking the exit code of a shell command. With git available it should just be `git diff --exit-code yarn.lock`. Without git we can copy the file before install and use `diff yarn.lock.orig yarn.lock`.

mstriemer:
There's actually a [`yarn check`](https://yarnpkg.com/en/docs/cli/check) command that seems to do this. It has a `--integrity` flag that we should pass when deploying.

kumar303:
I moved the `yarn check` bit to https://github.com/mozilla/addons-frontend/issues/3034

PR #1611, opened 2017-01-12
Use yarn on Travis

mstriemer:
Supports #1391.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
   - TOXENV=discopane-ui-tests GUI=1 PYTEST_ADDOPTS=--base-url=http://localhost:4000
   - TOXENV=flake8
 cache:
+  yarn: true
   directories:
   - node_modules
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ cache:
   directories:
   - node_modules
 before_install:
+- npm install -g npm@3 
 - if [ $TOXENV ]; then sudo pip install tox; fi
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then wget -O geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v0.10.0/geckodriver-v0.10.0-linux64.tar.gz; fi
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then gunzip -c geckodriver.tar.gz | tar xopf -; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,14 +18,10 @@ cache:
   directories:
   - node_modules
 before_install:
-- npm install -g npm@3
-- npm --version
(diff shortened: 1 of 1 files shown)


mstriemer:
Looks like Travis passes now. Hopefully this is accurate but who knows:

First run combined time was 36 min 45 sec.
Second run combined time was 27 min 20 sec.
Lint time went from 2 min 30 sec to 1 min 32 sec. 

That ain't bad.

tofumatt:
Nice!

- tofumatt

.travis.yml
```diff
@@ -18,14 +18,11 @@ cache:
   directories:
   - node_modules
 before_install:
-- npm install -g npm@3
-- npm --version
+- npm install -g npm@3
```
kumar303: maybe add a comment saying that yarn needs this (to avoid confusion)

mstriemer: I'm not sure if yarn needs it but I figured I'd leave it in so that the check script can ensure we're all on at least npm@3 in case someone does an accidental `npm install` or uses `npm` primarily.

I don't actually know why this was required but I think @muffinresearch would know more.

mstriemer:
So starting the discopane-ui-tests build takes 6 minutes and then the tests run in < 30 seconds. Maybe we can optimise that at some point.

kumar303:
Instead of [sleep 60](https://github.com/mozilla/addons-frontend/blob/master/.travis.yml#L33) here we could write a script to watch the `docker logs` output and exit when the server has started. That's not going to save us much time though. They are slow because they start a browser for each test case. Maybe we could leave an instance running somehow.

kumar303:
Hmm, I don't think we need to run the UI tests for every version of Node we support either. We should just run them for the version we use in prod.

muffinresearch:
Fwiw There's an open ops ticket for moving to node 6 here https://bugzilla.mozilla.org/show_bug.cgi?id=1323994 I need to follow-up but hopefully that's something that will happen in q1

mstriemer:
It only took 20 seconds to run the two tests, but it took 262 seconds to run:

```
if [[ $TOXENV == 'discopane-ui-tests' ]]; then docker run --cidfile=./docker-container.txt -d -p 4000:4000 -e NODE_APP_INSTANCE=disco -e NODE_ENV=uitests $(docker build -q .) /bin/sh -c "npm run build && npm run start" && sleep 60 && docker logs --tail=all $(cat ./docker-container.txt); fi
```

Trying the docker stuff locally it looks like it takes about a minute to start docker and 30 seconds to do the build so the sleep 60 is probably fine. I'm guessing this would take a little longer on Travis but 3 times longer to start docker plus the minute of sleeping seems like a lot. Maybe we can optimise the docker setup at some point.

I made it use `yarn` locally and it was way slower. Although I bet we're installing dependencies twice, Travis sets it up and then the docker container does it again and blows away the stuff Travis did. Maybe we can just skip the `npm install` in the Dockerfile.

PR #1612, opened 2017-01-12
Update to mocks and pwalm's comments

### LINKED ISSUE #1597: More touch ups to detail, search and header
From comments in #1570.

- [x] copy should be larger and white in the search bar
- [x] "by [developer name]" should be the same size as the extension name
- [x] let's use the "read more" button at the bottom of the "about this extension" block instead of the faded text under the link. We'll still keep the fade, but can we make it shorter? Like, 20px transition.
screen shot 2017-01-06 at 3 52 41 pm
- [x] some body copy looks grey, it should all be #000000
- [x] One big, site wide thing: Can we change the blue header zone to dark grey? Things will clash less and the switch will stand-out better.
screen shot 2017-01-06 at 3 54 51 pm
top: #343A40
search bar: #495057

mstriemer:
You can click these things to see the image (fancy).

<details>
<summary>Button</summary>

Button
</details>

<details><summary>Switch (off)</summary>

Switch (off)
</details>

<details><summary>Switch (on)</summary>

Switch (on)
</details>

Fixes #1597.

### DIFF
--- a/src/amo/components/AddonMoreInfo.scss
+++ b/src/amo/components/AddonMoreInfo.scss
@@ -1,5 +1,5 @@
 @import "~core/css/inc/mixins";
-@import "~core/css/inc/vars";
+@import "~amo/css/inc/vars";
 
 .AddonMoreInfo-contents {
   margin: 0;
--- a/src/ui/components/Card/Card.scss
+++ b/src/ui/components/Card/Card.scss
@@ -36,4 +36,10 @@
   border-bottom-right-radius: $border-radius-default;
   margin-top: 1px;
   text-align: center;
+
+  a {
+    color: $link-color;
+    font-variant: all-small-caps;
+    text-decoration: none;
+  }
 }
--- a/src/ui/components/Card/Card.scss
+++ b/src/ui/components/Card/Card.scss
@@ -35,11 +35,14 @@
   border-bottom-left-radius: $border-radius-default;
   border-bottom-right-radius: $border-radius-default;
   margin-top: 1px;
+  padding: 0;
   text-align: center;
 
-  a {
+  > a {
     color: $link-color;
+    display: block;
     font-variant: all-small-caps;
+    padding: 10px;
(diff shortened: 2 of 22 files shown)


src/amo/components/AddonMoreInfo.scss
```diff
@@ -1,5 +1,5 @@
 @import "~core/css/inc/mixins";
-@import "~core/css/inc/vars";
+@import "~amo/css/inc/vars";
```
mstriemer: I think we should import vars from amo now. #psa

tofumatt: I guess that makes sense, but I'd think a lot of things would have shared styles between AMO and Disco and DEFINITELY in the future should between the admin and AMO, right? That way they won't slowly drift apart (as they definitely have on addons-server).

Though if we don't have good UI tests for them it'll be a PITA, so I guess I see reasons for both.

mstriemer: I guess I could have updated the core styles but the base font colour and masthead colour didn't feel like they belonged in core. The amo vars import the core vars so common stuff can still go in core and the import can be from amo.

src/ui/components/Card/Card.scss
```diff
@@ -35,5 +35,14 @@
   border-bottom-left-radius: $border-radius-default;
   border-bottom-right-radius: $border-radius-default;
   margin-top: 1px;
+  padding: 0;
   text-align: center;
+
+  > a {
```
kumar303: * I think this should apply to all anchors, not just the direct descendant because injecting a footer can be done in so many different ways
* This should also apply to `:link`, `:active`, `:visited`, and `:hover`
* Could you add `font-weight: normal;` while you're in here? The mocks have that

I hit some trouble with the old one too: https://github.com/mozilla/addons-frontend/pull/1586/files#diff-5f781c3e4951964eb41f3c8da5d0cfbdR4

PR #1614, opened 2017-01-13
Downgrade karma-webpack from 2.0.0 to 1.8.1

mstriemer:
I was a little surprised the tests passed but didn't see and upgrade docs so figured they just rewrote a bunch and decided on a major bump.

This entirely broke the unit tests even though they kept passing 😠

Backing out the change, not sure if it's meant for webpack 2. We can deal with it later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -234,7 +234,7 @@
     "karma-mocha-reporter": "2.2.0",
     "karma-sinon": "1.0.5",
     "karma-sourcemap-loader": "0.3.7",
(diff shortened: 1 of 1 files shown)


mstriemer:
Build is still green and the tests actually ran. Yay

kumar303:
Whoa! We need this silent failure not to happen again. https://github.com/mozilla/addons-frontend/issues/1615

mstriemer:
Definitely need to fix that. Thanks for filing!

PR #1618, opened 2017-01-13
Switch to node:6-slim and perform build optimization. fixes #1120

### LINKED ISSUE #1120: Prepare for node 6
Node 6 is moving into LTS at the begining of October.

We should start using it for development and add it to the travis testing matrices. We'll be looking to move to using node 6 in production from mid Q4 onwards as soon as we're happy the linter and the addons-frontend will be happy running on it.  

I'll file an ops bug, and link back to this from there as we'll need to have closed this issue before we do the upgrade on the server.

jasonthomas:
Switching to the -slim image reduces the image size by 500MB, currently it is 1.3GB.

r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,21 @@
-FROM node:4.6
+FROM node:6-slim
 
 # Install node_modules into a different directory to avoid npm/npm#9863.
 RUN mkdir -p /srv/node
 ADD package.json /srv/node/
 WORKDIR /srv/node
-RUN npm install -g npm@3
-RUN npm install
+
+RUN buildDeps=' \
+    git \
+    ' && \
+    # install deps
+    apt-get update -y && \
+    apt-get install -y --no-install-recommends $buildDeps && \
+	npm install -g npm@3 && \
+	npm install && npm cache clean && \
+    # cleanup
+    # apt-get purge -y $buildDeps && \
+    rm -rf /var/lib/apt/lists/*
 
 ADD . /srv/code/
 WORKDIR /srv/code


tofumatt:
#1130 is a pull request here, did you reference the wrong issue?

(btw nice that's an impressive size reduction!)

jasonthomas:
I meant to reference #1120. I've updated the commit message and PR.

muffinresearch:
@jasonthomas is there a way to point to a tag that will get us any version beyond 6.9.x just thinking if this would mean we'd get stuck to 6.9.x if 6.10 came out for example.

muffinresearch:
> @jasonthomas is there a way to point to a tag that will get us any version beyond 6.9.x just thinking if this would mean we'd get stuck to 6.9.x if 6.10 came out for example.

I suppose a downside of handling this automatically is if for some reason a node update broke us it wouldn't be that great.

jasonthomas:
Yes we would be stuck to 6.9 until we decided to change the build to 6.10. If you prefer we can try setting it to nodejs:6-slim and it should reference the latest 6.x on build time.

muffinresearch:
I think there's a trade-off, presumably 6.9-slim would get patch updates in any case, but just not upgrade the minor version e.g. when 6.10 was released?

The main thing I was thinking about was avoiding the scenario that happened with centos a while back, where images pinned to a too specific version meant a lot of docker users using that image missed out on an important security update.

jasonthomas:
> I think there's a trade-off, presumably 6.9-slim would get patch updates in any case, but just not upgrade the minor version e.g. when 6.10 was released? 

Correct. Right now it is 6.9.4 and when a new point release is released (like 6.9.5) the image would be updated at build time.

> The main thing I was thinking about was avoiding the scenario that happened with centos a while back, where images pinned to a too specific version meant a lot of docker users using that tag missed out on an important security update.

If security updates happen as a point release then the image would get updated. Not sure how it would be handled for older versions.

[1] https://hub.docker.com/_/node/

muffinresearch:
@jasonthomas if we're currently going to get to patch changes I think it might make sense to also take minor versions too since node should be following semver and both minor and patch updates should be backwards compatible  - what do you think?

That should then mean we don't automatically get breaking changes (through a major version update).

PR #1619, opened 2017-01-13
Don't bail if watch arg is supplied

### LINKED ISSUE #1103: Recover from syntax error while running karma in a loop
- run the karma dev server in a loop, for example `npm run unittest:dev`
- create a syntax error in one of the source files
### What happened?

The karma runner dies with something like this:

```
Module build failed: SyntaxError: Unexpected token (19:4)

  17 |     i18n = getFakeI18nInst(),
  18 |     createRating = () => {}
> 19 |     ...customProps
     |     ^
  20 |   } = {}
  21 | ) {
  22 |   const props = {

21 09 2016 22:36:12.588:ERROR [karma]: [Error: Syntax error encountered, bailing]
```

The karma loop should continue running, waiting for you to fix the syntax error. Otherwise, you have to fix it and wait for karma to restart which takes a couple seconds and slows you down.

I don't know if this is possible but it's worth looking into!

muffinresearch:
Fixes #1103 

@kumar303 Give this a spin it - should not bail when running with `npm run unittest:dev` when there's an syntax error.

Note: if the syntax error is in a test file then the syntax error info will be above the rest of the test output. If it's a src file it will not run the tests. Coverage is the useful guide in the first case.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -51,8 +51,8 @@ const newWebpackConfig = Object.assign({}, webpackConfigProd, {
             }
           });
 
-          // Bail if syntax errors were encountered.
-          if (flagError) {
+          // Bail if syntax errors were encountered and we're not watching.
+          if (flagError && process.argv.indexOf('--watch') === -1) {
             throw new Error('Syntax error encountered, bailing');
           }
         }
@@ -150,7 +150,6 @@ module.exports = function karmaConf(conf) {
       'karma-sourcemap-loader',
       'karma-webpack',
     ],
-    autoWatch: true,
     browsers: ['Firefox'],
     singleRun: false,
     concurrency: Infinity,
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -39,22 +39,32 @@ const newWebpackConfig = Object.assign({}, webpackConfigProd, {
     function WebpackWarningPlugin() {
       this.plugin('done', (_stats) => {
         const stats = _stats;
-        let flagError;
+        let flagException = false;
 
-        if (stats.compilation.warnings.length) {
-          // Log each of the warnings
-          stats.compilation.warnings.forEach((_warning) => {
-            const warning = _warning.message || _warning || '';
-            if (warning.indexOf('SyntaxError') > -1) {
-              console.log(warning);
-              flagError = true;
-            }
-          });
+        const loggedExceptions = [
+          'SyntaxError',
+        ];
 
-          // Bail if syntax errors were encountered and we're not watching.
-          if (flagError && process.argv.indexOf('--watch') === -1) {
-            throw new Error('Syntax error encountered, bailing');
+        function strContainsException(str) {
+          return loggedExceptions.some((substring) => str.includes(substring));
+        }
+
+        ['errors', 'warnings'].forEach((key) => {
+          const loggedStack = stats.compilation[key];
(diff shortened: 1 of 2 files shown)


kumar303:
Hmm. I don't see a report on the syntax error. Do you? I did this:

1. Start `npm run unittest:dev`
2. Watched the tests pass once
3. Caused a syntax error in `ui/components/Rating/index.js`

It showed a confusing error and no indication of the syntax error.

screenshot 2017-01-13 12 30 34

muffinresearch:
Is the line pointed to below Cannot find module relevant?

muffinresearch:
Oh I see, probably not because RatingManager is importing the one with the error.

muffinresearch:
I think that should probably be looked at separately unless for some strange reason that behaviour is new with this patch?

muffinresearch:
Actually for me I got:

```
Firefox 50.0.0 (Mac OS X 10.11.0) ERROR  ReferenceError: whatever is not defined
  at webpack:///src/amo/api/index.js:4:0 <- test-runner.js:97995
```

Which was relevant to the thing I'd added.

muffinresearch:
Ah ok I think I've got this sorted. Seems there's a need to look at warnings *and* errors and catch SyntaxErrors in both. I also realised my example above is a `ReferenceError` and not a `SyntaxError` too. I'll add support for nicer errors for ReferenceError too at the same time.

kumar303:
Yeah, the warnings. Huh, I feel like we had to solve this for payments too.

kumar303:
Crap, you're right though. The behavior I mentioned is not related to your patch. I was doing this to trigger it:
````diff
diff --git a/src/ui/components/Rating/index.js b/src/ui/components/Rating/index.js
index 4622e74..d4729a5 100644
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -8,7 +8,7 @@ import './styles.scss';
 
 export default class Rating extends React.Component {
   static propTypes = {
-    rating: PropTypes.number,
+    rating = PropTypes.number,
     readOnly: PropTypes.boolean,
     onSelectRating: PropTypes.func,
   }
````

**Comment by muffinresearch** [comment]
Ok I've added another commit that catches `SyntaxError`s in `stats.compilation.errors` as well as `stats.compilation.warnings`.

ReferenceErrrors aren't surfaced in webpack so I can't get at them here. However the line indicated in the karma output is enough of a clue. I left the array of exceptions in place since it's nice if we do need to extend this to catch something else later on.

File: karma.conf.js [hunk 1]
```diff
@@ -39,22 +39,32 @@ const newWebpackConfig = Object.assign({}, webpackConfigProd, {
...
+
+        function strContainsException(str) {
+          return loggedExceptions.some((substring) => str.includes(substring));
+        }
+
+        ['errors', 'warnings'].forEach((key) => {
+          const loggedStack = stats.compilation[key];
+          if (loggedStack && loggedStack.length) {
+            stats.compilation[key].forEach((item) => {
+              const message = item.message || item || '';
```
kumar303: could `item` really be falsey here? If so, `item.message` is going to throw. If not, then you probably don't need some of those or statements.

muffinresearch: Yep that's a good point I think `item.message || ''` is good enough.

PR #1623, opened 2017-01-15
fix: Header link always white (fix #1253)

### LINKED ISSUE #1253: Header colour isn't white at first
I think this is related to us using the `a:link` selector instead of just `a` (why do we do that?)

screenshot 2016-10-19 00 30 35

So we need to set the `.Masthead-title` element to `color: #fff` or whatever colour it should be.

tofumatt:
This ensures the header colour is always white, even if you haven't visited the link.

Side note: I removed the `isHomePage` bit because we weren't using it (it was never set) and I think leaving the header as a link all the time is fine and helps set expectation that it's a link to home if the user hovers over/tabs through it even on the homepage. Personally I think it's too clever 😉


muffinresearch:
> Side note: I removed the isHomePage bit because we weren't using it (it was never set) and I think leaving the header as a link all the time is fine and helps set expectation that it's a link to home if the user hovers over/tabs through it even on the homepage. Personally I think it's too clever 😉

The original designs had a different treatment of the homepage header - is that not happening any more?

Also if we're not setting this up, what's the h1 on the homepage? The other reason for this approach you need an h1 on the homepage and the only really valid h1 is the name of the site. If it needs to be linked up all the time then I can live with that - but we do need an h1.

tofumatt:
> The original designs had a different treatment of the main header on the homepage - is that not happening any more?

Oh, did they? Shoot, I'll double-check that then. I didn't ever notice it, at least. My bad.

tofumatt:
The header seems to always be the same in the V1 mocks:
screenshot 2017-01-16 18 59 40

But I've added the `<h1>` back in ^_^

src/amo/components/MastHead.js
```diff
@@ -35,15 +30,12 @@ export class MastHeadBase extends React.Component {
...
-          {isHomePage
-            ? <h1 ref={(ref) => { this.title = ref; }}
-                className="MastHead-title MastHead-homepage">
-              {headerTitle}
-            </h1>
-            : <Link ref={(ref) => { this.title = ref; }} to="/"
-                className="MastHead-title">
-              {headerTitle}
-            </Link>}
+          <h1>
```
muffinresearch: This will mean there's more than one h1 on pages like details etc.

muffinresearch:
> The header seems to always be the same in the V1 mocks:

Ah ok, that's fine - it's clearly changed since the initial UX - but it seems like a good idea for consistency.

src/amo/components/MastHead.js
tofumatt: Oh shoot, I misunderstood: never have the `<h1>`. 😅

muffinresearch: So there should be an h1 on the homepage for the title, and not elsewhere. The original behaviour would have done that but just need the isHomepage bool to be true for the homepage.

muffinresearch: FWIW: Since the markup needs to be conditional I originally left the link out when isHomepage is true since it's a link to the page you're on so it seemed superfluous.

tofumatt: Gotcha. I guess the `<h1>` on the homepage makes sense. I like the discoverability realising "oh this is a link it will always get me home" by it being a link on the homepage. If a link to home on home is considered _bad_ UX I can certainly remove it, but I find it handy.

muffinresearch: > I like the discoverability realising "oh this is a link it will always get me home" by it being a link on the homepage. If a link to home on home is considered bad UX I can certainly remove it, but I find it handy.

Sounds good - I'm not too worried about leaving the link if you think it's better being consistent. 👍

kumar303:
Does this still need a review or did @muffinresearch cover it? LGTM at a glance.

tofumatt:
I think @muffinresearch covered it, you’re off the hook :-)

tofumatt:
Addressed all the comments, added `isHomePage` passing through from `App.js` and added some tests for it. Set for another r?

tests/client/amo/components/TestMastHead.js
```diff
@@ -30,24 +30,25 @@ describe('MastHead', () => {
...
 
-  it('renders a link when isHomepage is false', () => {
+  it('always renders a link in the header', () => {
     const root = renderMastHead({
-      isHomePage: false,
       children: FakeChild,
       SearchFormComponent: FakeChild,
     });
-    const titleLink = findDOMNode(root).querySelectorAll('.MastHead-title')[0];
+    const titleLink = findDOMNode(root).querySelector('.MastHead-title');
```
mstriemer: Should also check that the `h1` isn't rendered.

src/amo/containers/App.js
```diff
@@ -73,15 +75,20 @@ export class AppBase extends React.Component {
       InfoDialogComponent,
       MastHeadComponent,
       children,
+      clientApp,
       i18n,
+      lang,
       location,
     } = this.props;
+    const isHomePage = location.pathname === `/${lang}/${clientApp}/`;
```
mstriemer: I really wish there was a better way of checking this but react-router isn't too helpful here. It looks like the last `/` is optional, so that should be accounted for.

tofumatt: Oh, I see. I'm so used to us forcing trailing slash so I didn't know that was valid. Will fix, thanks.

PR #1625, opened 2017-01-16
Fix locale extract

tofumatt:
Pretty simple fix; just ignore the locale files in the `l10n` webpack config.


muffinresearch:
There's also a need to make sure the extract command exits with a non-zero exit code on errors like this just so we don't get caught out should we automate extraction. 

That could be a follow-up but since without this patch provides a good example, it might be nice to use this opportunity.

tofumatt:
We now bail on extraction errors in this patch: fdb06bd0085e8b0b460b2470e1232f8f033d21af

PR #1627, opened 2017-01-16
Localise user count in search results

### LINKED ISSUE #1608: Localise user count in search results
The user count in search results is not localised the same as it is on the detail page.

screen shot 2017-01-12 at 10 41 43 am
> Search page

screen shot 2017-01-12 at 10 41 27 am
> Detail page

The search results should have a comma in the number in English, `606,461 users` rather than `606461 users`.

mstriemer:
I updated how the lang is acquired. Rather than using state it uses `i18n.lang` and the lang is passed to the `makeI18n()` function to be set rather than pulling it from the first message (I'm curious what that would do if the string wasn't translated or we maybe didn't have any translations for a locale).

I updated the string to match the one from `AddonMeta` since they have different names for the number so it would have caused there to be an extra translation string (I think).

Screenshots:

<details><summary>English</summary>screen shot 2017-01-16 at 16 31 59
</details>

<details><summary>French</summary>

screen shot 2017-01-16 at 16 45 00
</details>
&nbsp;

Fixes #1608.

### DIFF
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -187,6 +187,8 @@ export function makeI18n(i18nData, lang, _Jed = Jed) {
   const i18n = new _Jed(i18nData);
   i18n.lang = lang;
 
+  i18n.formatNumber = (number) => number.toLocaleString(lang);
+
   // This adds the correct moment locale for the active locale so we can get
   // localised dates, times, etc.
   if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -183,14 +183,15 @@ export function makeMomentLocale(locale) {
 
 // Create an i18n object with a translated moment object available we can
 // use for translated dates across the app.
-export function makeI18n(i18nData, _Jed = Jed) {
+export function makeI18n(i18nData, lang, _Jed = Jed) {
   const i18n = new _Jed(i18nData);
+  i18n.lang = lang;
 
   // This adds the correct moment locale for the active locale so we can get
   // localised dates, times, etc.
   if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
     i18n.options._momentDefineLocale();
-    moment.locale(makeMomentLocale(i18n.options.locale_data.messages[''].lang));
+    moment.locale(makeMomentLocale(i18n.lang));
   }
 
   // We add a translated "moment" property to our `i18n` object
--- a/tests/client/amo/components/TestSearchResult.js
+++ b/tests/client/amo/components/TestSearchResult.js
@@ -12,12 +12,12 @@ import { getFakeI18nInst } from 'tests/client/helpers';
 
 
 describe('<SearchResult />', () => {
-  function renderResult(result) {
-    const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
+  function renderResult(result, { lang = 'en-GB' } = {}) {
+    const initialState = { api: { clientApp: 'android', lang } };
 
     return findRenderedComponentWithType(renderIntoDocument(
       <Provider store={createStore(initialState)}>
-        <SearchResult i18n={getFakeI18nInst()} addon={result} />
+        <SearchResult i18n={getFakeI18nInst({ lang })} addon={result} />
       </Provider>
     ), SearchResult).getWrappedInstance();
   }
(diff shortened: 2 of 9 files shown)


tests/client/helpers.js
```diff
@@ -71,7 +71,7 @@ export function unexpectedSuccess() {
 /*
  * Creates a stand-in for a jed instance,
  */
-export function getFakeI18nInst() {
+export function getFakeI18nInst({ lang = 'en-US' } = {}) {
```
tofumatt: I know this is only a test, but do we declare `en-US` as the default locale anywhere else? Would be nice to use a constant rather than the magic one here.

tests/client/amo/components/TestSearchResult.js
```diff
@@ -53,7 +53,14 @@ describe('<SearchResult />', () => {
...
     const users = findRenderedDOMComponentWithClass(root, 'SearchResult-users');
-    assert.equal(users.textContent, '553 users');
+    assert.equal(users.textContent, '5,253 users');
+  });
+
+  it('localises the user count', () => {
+    const localisedRoot = renderResult(result, { lang: 'fr' });
+    const users = findRenderedDOMComponentWithClass(localisedRoot, 'SearchResult-users');
+    // \xa0 is a non-breaking space.
+    assert.match(users.textContent, /5\xa0253/);
```
tofumatt: L'espace est bizarre ici pour moi.

How many results?
"5 253 !"

English must feel so cramped in comparison. 😉

src/core/i18n/utils.js
```diff
@@ -183,14 +183,15 @@ export function makeMomentLocale(locale) {
...
+export function makeI18n(i18nData, lang, _Jed = Jed) {
   const i18n = new _Jed(i18nData);
+  i18n.lang = lang;
 
   // This adds the correct moment locale for the active locale so we can get
   // localised dates, times, etc.
   if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
     i18n.options._momentDefineLocale();
-    moment.locale(makeMomentLocale(i18n.options.locale_data.messages[''].lang));
+    moment.locale(makeMomentLocale(i18n.lang));
```
tofumatt: That is quite a bit nicer, cheerio 👍

tofumatt:
Actually, only thing is: do we want to keep all our `i18n` stuff as methods off `i18n`? Maybe `i18n.numberFormat(number)` or even `i18n.toLocaleString(number)` would be nicer and more consistent with how we translate everything else?

Just a thought.

mstriemer:
I like the idea of `i18n.toLocaleString()`, I'll add it.

kumar303:
I'd like to preserve the ability to dispatch a Redux action and change `lang` everywhere -- that's what using `mapStateToProps` in `AddonMeta` did. This approach won't preserve it, as far as I can tell. 

I know that our current language switcher doesn't work this way since it does a redirect but that's a hack IMO and we should fix it. Redux is made for state changes like this.

kumar303:
Maybe if `I18nProvider` was the only one to use `mapStateToProps` (which would reset `i18n.lang`) then it would propagate all the way through?

src/core/i18n/utils.js
```diff
@@ -187,6 +187,8 @@ export function makeI18n(i18nData, lang, _Jed = Jed) {
   const i18n = new _Jed(i18nData);
   i18n.lang = lang;
 
+  i18n.formatNumber = (number) => number.toLocaleString(lang);
```
muffinresearch: yay

mstriemer:
We'd need to load the new i18n data and likely create a new `Jed` instance in `I18nProvider`. I don't think any of these changes prevent this.

The react docs [1] suggest that you do not update things that are in context because a parent component can return `false` in `shouldComponentUpdate()` which will prevent the children being checked, but they may depend on context.

They link to an article with potential solutions but I haven't read through it. Sounds like there is going to be some tricky bits getting the app to re-render after a lang change since we have that in context. Either way I don't think these changes prevent, or make more difficult, updating the app when the lang changes.

[1] https://facebook.github.io/react/docs/context.html#updating-context

kumar303:
Yes, the blog post you linked to says not to do exactly what you did :) None of this is a blocker for release but I think this patch would have been just as easy to write using Redux state. I'd like to fix it in the future because Redux will make the app a lot easier to work on. Filed: https://github.com/mozilla/addons-frontend/issues/1640 (low priority)

kumar303:
To be clear: I know there are other challenges preventing us from ditching the locale redirect (like, how do we load the new translations?) but the redirect was already hooked up to Redux state so we might as well use it to prepare for the future.

mstriemer:
I didn't modify the context. I changed a value that it had when it was created, that is totally fine.

If we want to make lang switching without a full reload possible then we will need to modify the `i18n` object that is returned in the context because it has all of the strings that are being translated.

PR #1629, opened 2017-01-16
Disable CSP for development

kumar303:
We used to have support for `CSP: false`. We should probably just bring that back but until then, this seems worth documenting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,31 +130,12 @@ to learn more about how configuration is applied.
 
 When developing locally with a webpack server, the randomly generated asset
 URL will fail our Content Security Policy (CSP) and clutter your console
-with errors. You can turn off all CSP errors by adding this to any of your
-local config files, such as `local-development-amo.js`:
+with errors. You can turn off all CSP errors by settings CSP to `false`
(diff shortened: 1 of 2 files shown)


muffinresearch:
It might be nicer to add some config and use that flag on the server to avoid the setup of the CSP middleware at all?

Better still would be adding the necessary rules to make the development server mode not cause CSP errors, but that will definitely take some doing.

kumar303:
ok, let me try this. I think last time I tried it I was overwhelmed by the lack of unit testability of `server/base.js`

kumar303:
actually, https://github.com/mozilla/addons-frontend/commit/66b62233e0296012b785b69412853ce36e49552f broke my config hack anyway :clown_face: What happens is that introducing a nonce for the style source [disables unsafe-inline](https://blogs.dropbox.com/tech/2015/09/unsafe-inline-and-nonce-deployment/).

kumar303:
r? It seems to work. I can't see an easy way to write a test for this but since it's just for development I'm not worried about it breaking without us knowing about it.

muffinresearch:
Nice, it's much cleaner like this. r+

PR #1630, opened 2017-01-17
Revert "Update react-redux to the latest version 🚀"

muffinresearch:
Fixes mozilla/addons-server#4312

Reverts mozilla/addons-frontend#1591 since it appears to cause a regression https://github.com/mozilla/addons-server/issues/4312


muffinresearch:
> r+ blarg, shame when we don't have a test that fails for something dependency related.

I've filed https://github.com/mozilla/addons-frontend/issues/1631 to cover investigating this. Adding a regression test is a good idea! I've added that to the notes in that issue.

PR #1636, opened 2017-01-18
fix(package): update react-helmet to version 4.0.0

tofumatt:
Closes #1624

https://greenkeeper.io/


tofumatt:
Seems like we're good here https://github.com/nfl/react-helmet/blob/master/CHANGELOG.md#400

Issue #1606, opened 2017-01-12
Handle async connect failures (LOAD_FAIL) on the client


kumar303:
### Describe the problem and steps to reproduce it:

- Load some page
- Navigate to another page that uses `asyncConnect` to load its data but fails in a API 404, 500, or whatever

### What happened?

There is no indication of the error in the UI

### What did you expect to happen?

It should show the user that there was an error

### Anything else we should know?

This works as expected on the server (you see a 404 page, etc). Maybe we can just re-use that server logic on the client. Relates to https://github.com/mozilla/addons-frontend/issues/307

Issue #1033, opened 2016-09-06
Provide error page HTML via components.


muffinresearch:
At the moment we use basic strings for errors. These are not translated and they aren't HTML. 

To fix this we should have a base error component and then specialise that for specific errors where we need to override the messaging.

muffinresearch:
Relates https://github.com/mozilla/addons-frontend/issues/307

kumar303:
Check out https://github.com/mozilla/addons-frontend/pull/1637 which lays a bit of foundation for rendering resource errors such as 404, 500, etc.

ValentinaPC:
Please add some STR for us to check this or the label "qa: not needed". Thanks!

tofumatt:
QA: To verify simply visit a page that results in error (easiest is a 404 Not Found page) and verify it is rendered with HTML and is not a plaintext string.

... although because addons-server provides our 404 errors right now this would be hard to test, so I can mark this as QA not needed for now, though when we move to using the frontend for everything it's worth testing. What would QA prefer?

ValentinaPC:
I'll try to find a 404 page to check this.
I'll also wanna add a legacy frontend label, if it is ok with you, to test this after the big changes coming up in our way.

ValentinaPC:
Verified this on AMO-dev FF53(Android 6.0.1)
Accessing https://addons-dev.allizom.org/en-US/android/mobile/ on my Android device will load a 404 page rendered with HTML.
Postfix screenshot:
![mobile](https://cloud.githubusercontent.com/assets/15685960/26311929/92b990d8-3f0e-11e7-8daa-7fedd29d8813.png)

Also, "`file an issue`" form the 404 page shouldn't redirect to "https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Contact" as "Report an issue" does, and not to the Github login page?
Thanks!

tofumatt:
The file an issue page should not go directly to GitHub but to https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Contact, yes. That's the current behaviour, right?

ValentinaPC:
> The file an issue page should not go directly to GitHub but to https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Contact, yes. That's the current behaviour, right?

The `file an issue` link redirects to Github - this is the current behavior.

PR #1637, opened 2017-01-18
Handle async-connect errors client side

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1606

This is just a quick fix until we do https://github.com/mozilla/addons-frontend/issues/1033 . It prints simple text client side like 'Not Found' for 404s, just like it would on the server side.


tests/client/amo/containers/TestApp.js
```diff
@@ -35,59 +37,50 @@ describe('App', () => {
...
-    class MyComponent extends React.Component {
-      render() {
-        return <p>The component</p>;
-      }
-    }
-    const i18n = getFakeI18nInst();
-    const location = sinon.stub();
-    const root = renderIntoDocument(
-      <AppBase i18n={i18n} isAuthenticated
+  function render({ children = [], ...customProps } = {}) {
```
tofumatt: Hah, thanks for this, I saw it when I was working on #1253 and thought "we should put this in a render function". 👍

kumar303:
Thanks for the review. I switched it to a HOC because that would make it easier to use on other top-level apps. We probably don't need to worry about that until later though.

PR #1641, opened 2017-01-19
[WIP] Store translations in redux, support lang updates without reload

mstriemer:
This adds support for updating the language without reloading the entire page. Formatting the numbers works but moment support was removed (should be easy to add back in with `serialize` but I just wanted this to work).

TODO:

- [ ] Update the URL when language changes.
- [ ] Support moment.
- [ ] Make the tests pass.
- [ ] Shave a few yaks.

The `I18nProvider` is no longer needed as the `translate()` HOC will create the i18n instance for components to use (this may use more memory but that's what has to happen I think).

The change is essentially to put the translations into redux instead of passing the i18n instance in with the translations included and have `translate()` pull the translations and create the `i18n` instance. The loading of strings on the client side has moved from `client/base.js` to `LanguagePicker` (I don't know how this will perform in production mode but I think hypothetically it should work).

translations-without-reload mov

I'm just putting this here so the code is sitting around somewhere, I'm not sure how much work it will take to make it mergeable.

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -46,7 +46,7 @@ poFiles.forEach((pofile) => {
   // Add the moment locale JS into our locale file, if one is available and
   // we're building for AMO (which is the only app that uses moment right
   // now).
-  if (appName === 'amo') {
+  if (false && appName === 'amo') {
     var defineLocale = null;
     try {
       const momentLocale = locale.replace('_', '-').toLowerCase();
--- a/src/amo/components/LanguagePicker.js
+++ b/src/amo/components/LanguagePicker.js
@@ -1,8 +1,10 @@
 /* global window */
+import config from 'config';
 import React, { PropTypes } from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 
+import { setLang } from 'core/actions';
 import languages from 'core/languages';
 import translate from 'core/i18n/translate';
 import { addQueryParams } from 'core/utils';
@@ -20,8 +22,7 @@ export class LanguagePickerBase extends React.Component {
   static propTypes = {
     currentLocale: PropTypes.string.isRequired,
     i18n: PropTypes.object.isRequired,
-    location: PropTypes.object.isRequired,
-    _window: PropTypes.object,
+    dispatch: PropTypes.func.isRequired,
   }
 
   onChange = (event) => {
@@ -30,13 +31,15 @@ export class LanguagePickerBase extends React.Component {
   }
 
   changeLanguage(newLocale) {
-    const { currentLocale, location, _window } = this.props;
+    const { currentLocale } = this.props;
 
     if (currentLocale !== newLocale) {
-      const newURL = changeLocaleURL({ currentLocale, location, newLocale });
-      // We change location because a locale change requires a full page
-      // reload to get the new translations, etc.
-      (_window || window).location = newURL;
+      const appName = config.get('appName');
+      // eslint-disable-next-line max-len, global-require, import/no-dynamic-require
+      require(`bundle?name=[name]-i18n-[folder]!../../locale/${newLocale}/${appName}.js`)((i18nData) => {
+        this.props.dispatch(setLang(newLocale));
+        this.props.dispatch({ type: 'SET_I18N', payload: i18nData });
+      });
     }
   }
 
--- a/src/core/i18n/translate.js
+++ b/src/core/i18n/translate.js
@@ -1,4 +1,7 @@
 import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+
+import { makeI18n } from 'core/i18n/utils';
 
 
 function getDisplayName(component) {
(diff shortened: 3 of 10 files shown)


src/amo/components/LanguagePicker.js
```diff
@@ -30,13 +31,15 @@ export class LanguagePickerBase extends React.Component {
...
+    const { currentLocale } = this.props;
 
     if (currentLocale !== newLocale) {
-      const newURL = changeLocaleURL({ currentLocale, location, newLocale });
-      // We change location because a locale change requires a full page
-      // reload to get the new translations, etc.
-      (_window || window).location = newURL;
+      const appName = config.get('appName');
+      // eslint-disable-next-line max-len, global-require, import/no-dynamic-require
+      require(`bundle?name=[name]-i18n-[folder]!../../locale/${newLocale}/${appName}.js`)((i18nData) => {
```
kumar303: whoa, that's pretty wild!

src/core/i18n/translate.js
```diff
@@ -49,6 +44,12 @@ export default function translate(options = {}) {
 
     Translate.displayName = `Translate[${getDisplayName(WrappedComponent)}]`;
 
-    return Translate;
+    function mapStateToProps({ api, i18n }) {
+      return {
+        i18n: makeI18n(i18n, api.lang),
```
kumar303: I don't think we'd need to store the i18n object in state, we only need the `lang` and `i18nData` in state. By connecting to those directly we would only re-render when their values change [1]. We could expose the `i18n` prop in a HOC by decorating every component that needs it with something like this:
````javascript
function withI18nProp(WrappedComponent) {
  function Localizer(props) {
    const { i18nData, lang, ...otherProps } = props;
    return <WrappedComponent i18n={makeI18n(i18nData, lang)} {...otherProps} />;
  }

  function mapStateToProps(state) {
    // Only re-render when lang or i18nData changes:
    return { lang: state.api.lang, i18nData: state.api.i18nData },
  }

  return compose(
    connect(mapStateToProps),
  )(Localizer);
}
````
[1] I don't think redux will descend into an object to diff all its keys so putting a complete `i18n` object in state might cause problems.

**Comment by kumar303** [comment]
Yeah! This looks doable

File: src/core/i18n/translate.js (hunk 2 above)
Comment (mstriemer) [inline src/core/i18n/translate.js]: I think that's what this is doing.

Comment (mstriemer) [inline src/core/i18n/translate.js]: When you wrap your component in `translate()` it pulls the translation strings and language from state and passes a new `i18n` instance through props.

Comment (mstriemer) [inline src/core/i18n/translate.js]: Oh, are you suggesting I move the `makeI18n()` call into the HOC instead of leaving it in `mapStateToProps`?

File: src/amo/components/LanguagePicker.js (hunk 1 above)
Comment (mstriemer) [inline src/amo/components/LanguagePicker.js]: Yeah, we had it in `client/base.js` before but it's quite the line.

File: src/core/i18n/translate.js (hunk 2 above)
Comment (kumar303) [inline src/core/i18n/translate.js]: I think you're right. The code I offered isn't any different than yours. At first I thought that we should prevent `makeI18n()` from getting called on *every* state change regardless of whether the lang state changes or not. Neither my code nor your code prevents that (I think?). Well, it's probably a premature optimization anyway.

If we measure and decide we do need to optimize this, we can probably use the [reselect approach](http://redux.js.org/docs/recipes/ComputingDerivedData.html) with a key on `lang`.

Comment (kumar303) [inline src/core/i18n/translate.js]: Also, I need to better understand when redux decides to render a `connect()`'d component or not. Their docs are not helpful.

Comment (mstriemer) [inline src/core/i18n/translate.js]: Making `translate()` a pure component might fix this if we move the call into the component itself but I am not sure. It is quite possible that this would break the underlying child components however since I the redux store is in the context and it is probably being updated to propagate changes. There might be some extra magic going on to keep redux from breaking but I'm not sure.

Reselect definitely seems like the right approach to me.

File: bin/build-locales [hunk 3]
```diff
@@ -46,7 +46,7 @@ poFiles.forEach((pofile) => {
   // Add the moment locale JS into our locale file, if one is available and
   // we're building for AMO (which is the only app that uses moment right
   // now).
-  if (appName === 'amo') {
+  if (false && appName === 'amo') {
```
mstriemer: This was added to avoid the moment code being unserialisable as JSON. I believe we've changed how we setup moment but it might be a `require` still so some extra work may still be required.

src/core/i18n/utils.js
```diff
@@ -191,10 +191,10 @@ export function makeI18n(i18nData, lang, _Jed = Jed) {
 
   // This adds the correct moment locale for the active locale so we can get
   // localised dates, times, etc.
-  if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
-    i18n.options._momentDefineLocale();
-    moment.locale(makeMomentLocale(i18n.lang));
-  }
+  // if (i18n.options && typeof i18n.options._momentDefineLocale === 'function') {
```
mstriemer: Moment related again, the moment data was removed so this wouldn't work as it did before.

Issue #1557, opened 2017-01-03
Logging out from addons-server does not log you out from addons-frontend


muffinresearch:
### Describe the problem and steps to reproduce it:

* Go to https://addons-dev.allizom.org as a desktop site.
* Login
* Click view mobile site link in the footer.
* Note the link "logout" is present in the header.
* Clicking logout sent me to the FXA auth page.
* Going to the desktop site I could logout
* Returning to the mobile site it was still showing "logout".

### What did you expect to happen?

* Clearing cookies was required to restore the correct logged out state.
* The logged-in state should stay in-sync between desktop and mobile.
* The mobile page Logout link should logout and not send the user to the login page.

muffinresearch:
Note: Fxa for -dev seems to have broke (JS error) - so maybe that's related?

ValentinaPC:
@muffinresearch : going on https://addons-dev.allizom.org from you mobile device and logging in, you will not me able to **log out** because clicking on the "Log Out" button will redirect you to FxA "_Sign in_" page.
Is this a correct behavior? Not being able to log out only from desktop site? Thanks!

mstriemer:
This is not intended. I'll split that into its own bug though, thanks.

ValentinaPC:
Verified as fixed on AMO-dev FF51(Win 7, Android 6.0.1) for the STR described and the result:
> The logged-in state should stay in-sync between desktop and mobile.
Postfix video:
![logout](https://cloud.githubusercontent.com/assets/15685960/23120590/9dcb7af4-f765-11e6-9c2e-b8b8b6fdc429.gif)

PR #1643, opened 2017-01-20
Make log out work on mobile pages

### LINKED ISSUE #1632: Log out tries to log you in again
The log in/out button has its text update when the user authenticates however the action that is performed when it is clicked is always "start log in process".

If the user is already logged in the "Log out" button should "log out" the user.

Related to #1557.

mstriemer:
Logs you out in the same way as it does on the admin. We have an extra cookie that we'll need to clear on addons-server. I'll submit a follow up patch for that later, tracked in #1557 and an addons-server issue near you.

This also clears the JWT from the `api` reducer on log out. That was definitely a bug.

Maybe we should get rid of the authentication reducer. I filed #1644 for that, I think it was intended to store info about the user but that can go in a "user" reducer and shouldn't store the JWT.

I also renamed `setJWT` to `setJwt` since it was bothering me, in a separate commit.

[demo gif (7.5MB)](https://cloud.githubusercontent.com/assets/211578/22137092/91609c20-de9e-11e6-915b-d0f78fec2f03.gif)

Fixes #1632.

### DIFF
--- a/src/amo/components/MastHead.js
+++ b/src/amo/components/MastHead.js
@@ -12,7 +12,7 @@ import './MastHead.scss';
 export class MastHeadBase extends React.Component {
   static propTypes = {
     i18n: PropTypes.object.isRequired,
-    isHomePage: PropTypes.bool,
+    isHomePage: PropTypes.bool.isRequired,
     location: PropTypes.object.isRequired,
     SearchFormComponent: PropTypes.node.isRequired,
     query: PropTypes.string,
--- a/src/amo/components/MastHead.js
+++ b/src/amo/components/MastHead.js
@@ -2,6 +2,7 @@ import React, { PropTypes } from 'react';
 
 import Link from 'amo/components/Link';
 import SearchForm from 'amo/components/SearchForm';
+import AuthenticateButton from 'core/components/AuthenticateButton';
 import translate from 'core/i18n/translate';
 
 import 'mozilla-tabzilla/css/_tabzilla.scss';
@@ -10,9 +11,9 @@ import './MastHead.scss';
 
 export class MastHeadBase extends React.Component {
   static propTypes = {
-    children: PropTypes.node,
-    isHomePage: PropTypes.bool.isRequired,
     i18n: PropTypes.object.isRequired,
+    isHomePage: PropTypes.bool,
+    location: PropTypes.object.isRequired,
     SearchFormComponent: PropTypes.node.isRequired,
     query: PropTypes.string,
   }
@@ -23,9 +24,7 @@ export class MastHeadBase extends React.Component {
   }
 
   render() {
-    const {
-      SearchFormComponent, children, i18n, isHomePage, query,
-    } = this.props;
+    const { SearchFormComponent, i18n, isHomePage, location, query } = this.props;
     const headerLink = (
       <Link className="MastHead-title" to="/">
         {i18n.gettext('Firefox Add-ons')}
@@ -34,10 +33,12 @@ export class MastHeadBase extends React.Component {
 
     return (
       <div className="MastHead">
-        <div id="tabzilla">
-          <a href="https://www.mozilla.org">Mozilla</a>
+        <div className="MastHead-top-row">
+          <div id="tabzilla">
+            <a href="https://www.mozilla.org">Mozilla</a>
+          </div>
+          <AuthenticateButton className="MastHead-auth-button" size="small" location={location} />
         </div>
-        {children}
         <header className="MastHead-header">
           {isHomePage
             ? <h1 className="MastHead-title-wrapper">{headerLink}</h1>
--- a/src/amo/components/MastHead.scss
+++ b/src/amo/components/MastHead.scss
@@ -60,6 +60,7 @@
 
   .Icon {
     @include margin-end(0.5em);
+
     vertical-align: top;
   }
 }
--- a/src/amo/components/MastHead.scss
+++ b/src/amo/components/MastHead.scss
@@ -1,15 +1,5 @@
 @import "~amo/css/inc/vars";
 
-#tabzilla {
-  left: 14px;
-  position: absolute;
-
-  [dir=rtl] & {
-    left: auto;
-    right: 14px;
-  }
-}
-
 .MastHead {
   background: $masthead-color;
   min-height: 100px;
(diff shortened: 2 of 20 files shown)


tofumatt:
Why `setJWT` over `setJwt`? I find things like `useAPI` over `useApi` more readable because I'm used to seeing things like JWT or API in all-caps.

src/amo/containers/App.js
```diff
@@ -57,10 +59,10 @@ export class AppBase extends React.Component {
   }
 
   accountButton() {
-    const { handleLogIn, i18n, isAuthenticated, location } = this.props;
+    const { handleLogIn, handleLogOut, i18n, isAuthenticated, location } = this.props;
     return (
       <button className="button AccountButton"
-              onClick={() => handleLogIn(location)}
+              onClick={() => (isAuthenticated ? handleLogOut() : handleLogIn(location))}
```
tofumatt: You encouraged me to make all handler methods `onClick() {}` and then pass off to other methods from there. I'd dig that here, especially so we don't have to use a ternary to set which method to call.

src/core/actions/index.js
```diff
@@ -1,18 +1,23 @@
...
+  LOG_OUT_USER,
   SET_JWT,
   SET_CLIENT_APP,
   SET_LANG,
   ENTITIES_LOADED,
   SET_CURRENT_USER,
 } from 'core/constants';
 
-export function setJWT(token) {
+export function setJwt(token) {
```
tofumatt: As mentioned, I prefer `setJWT`. But not enough to r- this 😉

kumar303: camel case is hard

mstriemer: Yeah, I think we've been formatting it like this so things like `setJwtToken` are a bit clearer. I think that looks better than `setJWTToken`.

mstriemer: If we want to change things that's cool but I think most of our acronyms are formatted like this. I know we have a `callApi` too anyway.

tofumatt: Yeah I'm cool with the revised format. I like `setJWT` more than `setJwt` but I like your example where `setJwtToken` is better and it's easier to grok the first rule so let's do that 👍

src/amo/containers/App.js
mstriemer: This is a good call. I did however end up moving the button into its own component and that is now included by `MastHead` instead of being the only passed in child to it from `App`. I also refactored the CSS in `MastHead` a bit for the header to use flexbox  (I added a `div` but it's shorter) and an `Icon` (in `AuthenticateButton`).

src/amo/containers/App.js
```diff
@@ -102,13 +86,8 @@ export class AppBase extends React.Component {
   }
 }
 
-export const setupMapStateToProps = (_window) => (state) => ({
-  clientApp: state.api.application,
```
mstriemer: This had a bug in it (looking for `application` instead of `clientApp`) so there was never an `h1` on the homepage :(

src/amo/components/MastHead.js
```diff
@@ -34,10 +33,12 @@ export class MastHeadBase extends React.Component {
 
     return (
       <div className="MastHead">
-        <div id="tabzilla">
-          <a href="https://www.mozilla.org">Mozilla</a>
+        <div className="MastHead-top-row">
+          <div id="tabzilla">
+            <a href="https://www.mozilla.org">Mozilla</a>
+          </div>
+          <AuthenticateButton className="MastHead-auth-button" size="small" location={location} />
```
mstriemer: Unfortunately need to pass `location` throughout, I didn't see a way of getting it from react-router. Could update `AuthenticateButton` to just use `window.location.pathname` instead. I'm happy with either.

mstriemer:
I made the button white now, since that's what it was before. I'm not sure if it was grey for a reason but it didn't match the other grey so I checked the mocks and it matches the tabzilla button there.

Updated screenshots:

screen shot 2017-01-25 at 23 22 30 screen shot 2017-01-25 at 23 22 40

src/amo/components/MastHead.js
tofumatt: I thought there was a `withRouter` HOC you could pass to `compose`? Would that work?

src/amo/components/MastHead.scss
```diff
@@ -52,38 +43,23 @@
...
-  top: 15px;
-
-  [dir=rtl] & {
-    left: 15px;
-    right: auto;
-  }
+.MastHead-top-row {
+  display: flex;
+  margin: 0 15px;
+  justify-content: space-between;
```
tofumatt: I always hear maxresdefault

src/amo/containers/App.js
tofumatt: Shoot, thanks for catching it.

src/amo/components/MastHead.scss
```diff
@@ -52,38 +43,24 @@
...
-  color: $header-font-color;
-  font-size: 12px;
-  padding: 8px;
-  position: absolute;
-  right: 15px;
-  top: 15px;
+.MastHead-top-row {
+  display: flex;
+  margin: 0 15px;
+  justify-content: space-between;
```
tofumatt: I always hear maxresdefault

src/ui/components/Icon/styles.scss
```diff
@@ -8,18 +8,22 @@
...
 }
 
 .Icon-user {
-  background: url('./user.svg') 0 50% no-repeat;
+  background: url('./user.svg') center no-repeat;
+  background-size: contain;
+}
+
+.Icon-user-dark {
+  background: url('./user-dark.svg') center no-repeat;
```
mstriemer: Hopefully we can get `Icon` using `<svg>` tags so we can do stuff like this with CSS at some point. It's kind of annoying to need to update the XML to change the stroke colour which causes another asset to be downloaded :(

tofumatt: Yeah, I started on it but it was too much work to convert old icons (at the time). It's a great lazy Sunday patch that I'm hoping to do in the future; CSS control over SVG is like one of the best reasons to use them :-)

src/amo/components/MastHead.js
mstriemer: There is but I didn't see `location` being exposed on the router in the docs. I'll try that and see if it just isn't documented.

mstriemer: It looks like the `router` has two methods that look like they might work, `createLocation()` and `createHref()`. Calling these without any arguments results in `TypeError: location is undefined` so I don't think they will work :(

mstriemer:
Getting network errors on the ui tests but I think these failures are irrelevant so I'm going to merge this. Got another branch that is going to be blocked on this right away.

PR #1646, opened 2017-01-20
Adding Nightly System Add-ons to the ignore list.

designakt:
Since updating this list I identified 4 more system add-ons in Nightly.

### DIFF
--- a/src/disco/constants.js
+++ b/src/disco/constants.js
@@ -17,4 +17,8 @@ export const DISCO_DATA_GUID_IGNORE_LIST = [
   'loop@mozilla.org', // Firefox Hello
   'aushelper@mozilla.org', // Application Update Service Helper
   'webcompat@mozilla.org', // Web Compat
+  'flyweb@mozilla.org', // FlyWeb
+  'formautofill@mozilla.org', // Form Autofill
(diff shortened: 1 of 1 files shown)


muffinresearch:
Are these all present in a completely new profile?

designakt:
@muffinresearch I did not check that. Why is that relevant?

muffinresearch:
> @muffinresearch I did not check that. Why is that relevant?

Because that ensure they are there by default as opposed to something that got added via usage.

designakt:
Why does that matter? Any system addon will be installed by us, without the user needing to know about addons. So even if it got added because of usage they still can be new addon users.

muffinresearch:
> Why does that matter? Any system addon will be installed by us, without the user needing to know about addons. So even if it got added because of usage they still can be new addon users. 

But they should be present in a new profile created with that specific version of nightly too. Checking that makes sure there's nothing being added that shouldn't be.

designakt:
I found a safer way to identify system add-ons:
in about:telemetry -> Environment -> Addons i can check for isSystem=true
So we can be sure the ones I added are system add-ons, no matter if they appear in a new profile or not.

PR #1648, opened 2017-01-23
Don't use relative moment require paths

### LINKED ISSUE #1628: 500 when accessing a locale other than en-US
When you access a language other than English a 500 error is returned.

### What happened?
The moment locale code is attempting to load a `../moment` module which does not appear to exist.

Moment locales load fine and the page is served in the requested language.

```
[1] [2017-01-16T22:59:55.026Z] ERROR: server/58760 on mstriemer-30951: Cannot find module '../moment' (app=amo, err.code=MODULE_NOT_FOUND)
[1]     Error: Cannot find module '../moment'
[1]         at Function.Module._resolveFilename (module.js:325:15)
[1]         at Function._module2.default._resolveFilename (/Users/markstriemer/work/addons-frontend/node_modules/require-hacker/babel-transpiled-modules/require hacker.js:423:34)
[1]         at Function.module._load (/Users/markstriemer/work/addons-frontend/node_modules/piping/lib/launcher.js:24:23)
[1]         at Module.require (module.js:353:17)
[1]         at require (internal/module.js:12:17)
[1]         at /Users/markstriemer/work/addons-frontend/src/locale/fr/amo.js:70:51
[1]         at Object.anonymous [as _momentDefineLocale] (/Users/markstriemer/work/addons-frontend/src/locale/fr/amo.js:68:3)
[1]         at makeI18n (/Users/markstriemer/work/addons-frontend/src/core/i18n/utils.js:193:18)
[1]         at /Users/markstriemer/work/addons-frontend/src/core/server/base.js:232:24
[1]         at run (/Users/markstriemer/work/addons-frontend/node_modules/core-js/modules/es6.promise.js:87:22)

### PR DESCRIPTION
**Pull request by tofumatt**
The issue was looking for `../moment` and since we're already playing with the locales the easiest thing was to modify the `require` directly. This fixes #1628.

### CODE REVIEW

**Comment by tofumatt** [comment]
d87b2d0bf800a4a694279ef67dfd0a224133aa1e is the commit you want to r? in this PR; the other commit is just the chore of rebuilding the locale files to have the right require path.

File: webpack.dev.config.babel.js [hunk 1]
```diff
@@ -128,4 +126,15 @@ export default Object.assign({}, webpackConfig, {
     new webpack.IgnorePlugin(/webpack-stats\.json$/),
     webpackIsomorphicToolsPlugin.development(),
   ],
+  resolve: {
```
muffinresearch: Is this not already there via the dev config being derived from the prod one?

tofumatt: Actually yes, sorry this was a holder from testing and should be removed, thanks!

src/locale/id/amo.js
```diff
@@ -20,32 +20,32 @@ module.exports = { domain:"messages",
       "View desktop site":[ "Lihat situs desktop" ],
       "Browse in your language":[ "Jelajahi dalam bahasa Anda" ],
       "Firefox Add-ons":[ "Pengaya Firefox" ],
-      "How are you enjoying your experience with %(addonName)s?":[ "" ],
-      "screenshot %(imageNumber)s of %(totalImages)s":[ "" ],
-      "Average rating: %(rating)s out of 5":[ "" ],
-      "No ratings":[ "" ],
-      "%(users)s user":[ "" ],
-      "Log out":[ "" ],
+      "How are you enjoying your experience with %(addonName)s?":[ "Bagaimana pengalaman Anda menggunakan %(addonName)s?" ],
```
kumar303: Why does this patch include translations? Did you mean to include these?

tofumatt: Yeah, because the locale files were causing errors and needed to be rebuilt to fix them; I figured I'd include them here. Otherwise this won't show as fixed until we rebuilt the locales which won't be until next week.

tofumatt:
Yeah I'm not extremely happy with how we're doing this either but every JS i18n date library has locales that include code (for things like "1st, 2nd, 3rd" type numbers in different kinds of languages)... after a fair bit of messing about this was the best/most performant way to include it with our locales.

Definitely up for a better way if found though.

kumar303:
Adding a note to my self: we chatted on IRC about auto-generating a require statement instead, which seems to work: https://dpaste.de/9v8u

tofumatt:
Huzzah, this is updated to not copy/paste the moment locale and instead require it.

I think the reason this used to not work in the first place is incredibly dumb oversight; I forgot to check for the locale in moment first and just added the require. So locales like `ast` (which moment doesn't yet support) would cause errors. Once I fixed that it worked and this is way better.

Thanks for the suggestion on IRC @kumar303! Ready for another r!

bin/build-locales
```diff
@@ -50,15 +50,15 @@ poFiles.forEach((pofile) => {
     var defineLocale = null;
     try {
       const momentLocale = locale.replace('_', '-').toLowerCase();
+      // Check for the locale first; if it doesn't exist we don't have
+      // a moment locale that matches.
+      fs.accessSync(`./node_modules/moment/locale/${momentLocale}.js`);
```
kumar303: To make the access check more effective, you could share the same variable by doing:
````js
const localeModule = `moment/locale/${momentLocale}.js`;
fs.accessSync(`./node_modules/${localeModule}`);
// ...
defineLocale = new Function(`require('${localeModule}');`);
````

File: bin/build-locales [hunk 4]
```diff
@@ -50,15 +50,15 @@ poFiles.forEach((pofile) => {
...
-      // written out by `toSource()` so that it can be used later (at runtime,
-      // by moment).
+      // raw code; this function won't be executed but will be written out by
+      // `toSource()` so that it can be used later (at runtime, by moment).
       defineLocale = new Function(
-        fs.readFileSync(
-          `./node_modules/moment/locale/${momentLocale}.js`, 'utf8'
-        )
-      );
+        `require('moment/locale/${momentLocale}.js');`);
```
kumar303: do you still need `new Function()` ?

kumar303: Since requiring this module is magical, I think it would make sense to add a comment to the auto-generated code like:

    // By requiring this module, the new locale is defined and registered internally for moment.

tofumatt: Yeah, because if we include `require` in there directly it will evaluate and throw an error, and if we make it a string `toSource` will output it as one. It's now quite harmless; a bit ugly but that's all.

tofumatt: Works for me! 👍

tofumatt:
Changes made (and bit about `new Function` usage explained). I think this should be good now, r?

bin/build-locales
```diff
@@ -50,15 +50,18 @@ poFiles.forEach((pofile) => {
...
-        fs.readFileSync(
-          `./node_modules/moment/locale/${momentLocale}.js`, 'utf8'
-        )
-      );
+      // raw code; this function won't be executed but will be written out by
+      // `toSource()` so that it can be used later (at runtime, by moment).
+      defineLocale = new Function(`
+        // By requiring this module, the new locale is defined and
+        // registered internally for moment.js
+        require('${localeModulePath}');`);
```
kumar303: I guess it's ok that the module path ends in `.js` now? That's probably fine.

tofumatt: Good point; I can remove it though, no need for it.

tofumatt:
(I forgot I didn't merge this.)

Issue #1639, opened 2017-01-19
Star rating on search page is always shown as 4/5


mstriemer:
The search page has a star count but it is always 4/5. There's now a `Rating` component that can be set to `readOnly` and it will render the stars for us instead of using the placeholder image.

<img src="https://cloud.githubusercontent.com/assets/211578/22094999/27978f46-ddd7-11e6-831b-81765a35a955.png" alt="screen shot 2017-01-18 at 23 37 20" style="max-width:100%;" width="300">

While we're at it the CSS could probably be updated to use `flex` instead of `float` for the icon. The icon should also be vertically centred. There are 4 headings in each component as well. I could see the add-on name being a heading (it's an `h2` right now) but the rating, author and number of users probably don't need to be in `h3` tags.

kumar303:
I can take this

cosminbadescuSV:
This issue is verified as fixed on AMO-dev on Firefox 50.1.0 (Android 7.1.1) 
After performing a search the stars from search results are displayed correctly.
Here is a video:
![stars](https://cloud.githubusercontent.com/assets/15885467/22366579/1bef43f4-e486-11e6-8a1b-25f7cb103cc5.gif)

PR #1653, opened 2017-01-25
Show real ratings in search results

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1639

The screenshot isn't too exciting because it pretty much looks the same. The stars are just a bit smaller to be consistent with the review listing page. Looking at the Sketch file, I think the smaller stars are more accurate.

screenshot 2017-01-25 17 45 48


src/amo/components/SearchResult.js
```diff
@@ -38,7 +25,9 @@ export class SearchResultBase extends React.Component {
           <section className="SearchResult-main">
             <img className="SearchResult-icon" src={addon.icon_url} alt="" />
             <h2 className="SearchResult-heading">{addon.name}</h2>
-            {rating}
+            <div className="SearchResult-rating">
+              <Rating rating={addon.ratings.average} readOnly size="small" />
```
tofumatt: "small" here feels like a magic constant.

src/ui/components/Rating/index.js
```diff
@@ -66,18 +71,35 @@ export default class Rating extends React.Component {
   }
 
   render() {
-    const { readOnly, size } = this.props;
+    const { i18n, rating, readOnly, size } = this.props;
     const cls = classNames('Rating', {
```
tofumatt: Could this be `className` instead of `cls`? I find those short vars hard to read best of times but this one actually confused me a bit.

src/ui/components/Rating/index.js
```diff
@@ -66,18 +71,35 @@ export default class Rating extends React.Component {
...
       'Rating--editable': !readOnly,
       'Rating--small': size === 'small',
```
tofumatt: Maybe not super-related to this PR but is there a set of valid sizes? I feel like we should throw if the size isn't valid, especially as the values aren't passed around as constants.

src/ui/components/Rating/index.js
```diff
@@ -66,18 +71,35 @@ export default class Rating extends React.Component {
...
     });
+    let description;
+    if (rating) {
+      description = i18n.sprintf(
+        // L10n: This is for showing the star rating of an add-on.
+        i18n.ngettext('Rated %(starCount)s star',
```
tofumatt: Might be better to right `Rated %(starCount)s out of 5`; visually it's clear the rating is out of five but a screenreader saying "it's 4 stars" won't give much context (eg "4 out of what? 4 out of 5? Out of 10?")

The "stars" bit doesn't even add much either, so I think a plain number here (X of out 5) would work better; bonus it's less complex code.

tests/client/ui/components/TestRating.js
```diff
@@ -92,6 +98,33 @@ describe('ui/components/Rating', () => {
     });
   });
 
+  it('renders selected stars nearest to an average rating', () => {
```
tofumatt: This description confuses me a bit, I feel like it's saying "should round down average ratings to stars" or something, as the crux of the test seems to be that we don't round up.

src/ui/components/Rating/index.js
kumar303: Yeah, that's a fair point. All possible values are defined in the PropType above but that's pretty much useless. I patched it to throw an error. If we were using Flow all of this would be way easier.

src/ui/components/Rating/index.js
kumar303: ahh, no need for `ngettext`, I like it

src/amo/components/SearchResult.js
kumar303: It throws an error now if you specify an invalid value

kumar303:
looks like this test bustage is already on master (unrelated to my patch) so I'm merging

kumar303:
test breakage is from https://github.com/mozilla/addons-frontend/pull/1618

Issue #1638, opened 2017-01-19
Add-on rating is not displayed on detail page


mstriemer:
We show the add-on's rating on the search results but not on the detail page. It would be nice to see the rating there.

@pwalm any thoughts on where this should go?

Here's a quick mock, I'm not sure how we want to fit the rating and the count or maybe just the rating.

![screen shot 2017-01-18 at 23 34 11](https://cloud.githubusercontent.com/assets/211578/22094940/c6d1f3f4-ddd6-11e6-9b9a-9ffdbd720433.png)

pwalm:
The only thing I would change is remove the number of reviews and balance it with the number of users so the icons are the same size and the text sits on the same line. Looks a bit weird with the two lines of text stacked like that.

kumar303:
@pwalm thanks for taking a look. I filed an issue to make that change https://github.com/mozilla/addons-frontend/issues/1698

pwalm:
Thanks @kumar303!

madalincm:
Verified as fixed in FF51(Android 7.1.1) addons-dev.allizom.org
The rating is now displayed in add-on detail page.

![screenshot_20170206-173627](https://cloud.githubusercontent.com/assets/15685441/22653671/d0469bcc-ec92-11e6-803b-b732e8301b72.png)

The problem about the text and the balancing is being tracked in #1698. Marking this bug as verified fixed.

PR #1667, opened 2017-01-27
Show average rating on detail page

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1638

screenshot 2017-01-27 15 18 26

screenshot 2017-01-27 15 17 50


src/amo/components/AddonMeta.js
```diff
@@ -8,24 +8,54 @@ import 'amo/css/AddonMeta.scss';
...
       { total: i18n.formatNumber(averageDailyUsers) },
     );
+    let starCount;
+    if (averageRating) {
+      starCount = i18n.sprintf(
+        i18n.gettext('%(averageRating)s out of 5'),
+        { averageRating: i18n.formatNumber(averageRating) },
+      );
+    } else {
+      starCount = i18n.gettext('Not yet rated');
```
tofumatt: This is pedantic but I wonder if `'No ratings'` (and `'No reviews'`, below) would be a bit easier to grok and read. Not yet rated feels almost like an invitation to action ("YOU could be the first to rate it") but here it really isn't and I think it's weirder to parse as-is.

Could we change it?

kumar303: I think there are pros and cons to each one but I don't feel too strongly. I think the best thing to do is just land it and get feedback from UX. We had requested feedback on this anyway and hadn't received it yet so other things about it might change.

tofumatt: Cool, sounds good 👍

kumar303:
The test failure was already on master.

Issue #1682, opened 2017-01-30
server tests time out sometimes when calling runServer()


kumar303:
Sometimes the server tests time out when calling [server/base:runServer()](https://github.com/mozilla/addons-frontend/blob/master/src/core/server/base.js#L270) from [TestDetail](https://github.com/mozilla/addons-frontend/blob/master/tests/server/amo/TestDetailsPage.js#L16).

kumar303:
The patch in https://github.com/mozilla/addons-frontend/issues/1684 will hopefully show us the real error next time.

kumar303:
I haven't seen this happen since the patch mentioned above was merged. I think the next timeout will be easier to diagnose so I'm closing this one.

Issue #1559, opened 2017-01-03
Add sentry middleware to node service


muffinresearch:
We should add the sentry middleware to the node service so we get exception details into sentry.

Doing this via a middleware will probably make the most sense: https://docs.sentry.io/clients/node/integrations/express/

muffinresearch:
I think this should be in v1 otherwise debugging server errors in the live service will be much harder.

PR #1684, opened 2017-01-30
Report runServer() failures better

kumar303:
This addresses https://github.com/mozilla/addons-frontend/issues/1682 but only in that it should show us what the real error was.

### DIFF
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -267,19 +267,23 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
   return app;
 }
 
-export function runServer({ listen = true, app = appName } = {}) {
-  if (!app) {
-    throw new Error(
-      `Please specify a valid appName from ${config.get('validAppNames')}`);
-  }
-
+export function runServer({
+  listen = true, app = appName, exitProcess = true,
+} = {}) {
   const port = config.get('serverPort');
   const host = config.get('serverHost');
 
   const isoMorphicServer = new WebpackIsomorphicTools(
     WebpackIsomorphicToolsConfig);
-  return isoMorphicServer
-    .server(config.get('basePath'))
+
+  return new Promise((resolve) => {
+    if (!app) {
+      throw new Error(
+        `Please specify a valid appName from ${config.get('validAppNames')}`);
+    }
+    resolve();
+  })
+    .then(() => isoMorphicServer.server(config.get('basePath')))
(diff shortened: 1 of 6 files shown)


kumar303:
@muffinresearch does this look ok?

src/core/server/base.js
```diff
@@ -312,6 +311,7 @@ export function runServer({ listen = true, app = appName } = {}) {
     })
     .catch((err) => {
       log.error({ err });
+      throw err;
```
muffinresearch: What the rationale on re-throwing a caught error?

src/core/server/base.js
```diff
@@ -269,9 +269,8 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
 
 export function runServer({ listen = true, app = appName } = {}) {
   if (!app) {
-    log.fatal(
+    throw new Error(
       `Please specify a valid appName from ${config.get('validAppNames')}`);
-    process.exit(1);
```
muffinresearch: Is there a reason to throw and not bail at this point?

kumar303: I didn't want the test suite to be completely killed on this line: https://github.com/mozilla/addons-frontend/blob/master/tests/server/amo/TestDetailsPage.js#L16 I'd like it to fail with the error instead. Actually, I can't really tell if `process.exit(1)` was the right thing to do. When I add that back, `npm run dev:amo` just keeps spawning the process in a loop as if it's trying to start it again. Hmm.

src/core/server/base.js
kumar303: Will it get logged elsewhere? I wanted to be sure it would.

src/core/server/base.js
muffinresearch: I suppose that might be something that needs to change contextually. Having the process exit when the server is run in production due to a config error is likely to still be desirable.

src/core/server/base.js
muffinresearch: I would have thought the logging should be enough. Another potentially relevant issue is the one for adding the sentry client which should help for tracking down caught and uncaught server exceptions. https://github.com/mozilla/addons-frontend/issues/1559

kumar303:
@muffinresearch can you take another look? This puts back the `process.exit(1)` but for the server scripts only.

muffinresearch:
Looks good to me.

PR #1685, opened 2017-01-30
Error pages

### LINKED ISSUE #1033: Provide error page HTML via components.

### LINKED ISSUE #1605: Check that all API requests have withErrorHandling
We need to make sure all our components/API requests use error handling for mobile v1 until we get #1484 done.

### LINKED ISSUE #1616: 404s are 500s
Go to http://127.0.0.1:3000/en-US/firefox/404/, should be a 404 but instead it's a 500. 

The trace is:

```
Error: No LandingPage content for addonType: 404
[1]         at LandingPageBase.contentForType (/Users/Whatevs/code/addons-frontend/src/amo/components/LandingPage.js:77:11)
[1]         at LandingPageBase.render (/Users/Whatevs/code/addons-frontend/src/amo/components/LandingPage.js:85:23)
[1]         at /Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:796:21
[1]         at measureLifeCyclePerf (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:75:12)
[1]         at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:795:25)
[1]         at ReactCompositeComponentWrapper._renderValidatedComponent (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:822:32)
[1]         at ReactCompositeComponentWrapper.performInitialMount (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:362:30)
[1]         at ReactCompositeComponentWrapper.mountComponent (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:258:21)
[1]         at Object.mountComponent (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactReconciler.js:46:35)
[1]         at ReactCompositeComponentWrapper.performInitialMount (/Users/Whatevs/code/addons-frontend/node_modules/react-dom/lib/ReactCompositeComponent.js:371:34)
```

### LINKED ISSUE #1673: reduxAsyncConnect stops further navigation
In working on the error pages (and have been struggling with them for awhile). I thought originally I just wasn't clearly the set error handlers but that's not the case. It seems that `reduxAsyncConnect`, after handling an error either with our `handleResourceError` HOC or by `catch()`ing the promises in `utils.js` or similar, stops all further rendering of the app. I've tried this with everything from a simple hard-coded string in `App.js` to render error components which will display but then stop all further updates of components.

What's especially strange is the actions are clearly still being sent and they are updating state, including state that I've added to `App.js` container in `amo/containers/App.js` (see screenshot) but the components do not re-render. It seems as soon as there's an error component or string rendered in `App.js` further rendering (but not navigation, action disptaching, or state/store changes) stops.

Either I'm missing something incredibly simple but not obvious, or something is wrong here.

jan-29-2017 14-37-04

---

I should note this happens with both links on the page (when I had an error component that rendered links) and the back button. Sometimes the second link click would not be caught by the router and would force a page reload upon navigation. But I can't exactly tell what the steps to reproduce there are so not sure it's important.

---

### Basic steps to reproduce

1. Edit https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/SearchResult.js#L22 to change the URL to `/addon/${addon.slug}-not/`
2. Navigate to any search result page (eg http://localhost:3000/en-US/firefox/search/?q=test)
3. Click on a link
4. Get an error page
5. Navigate back

Note that nothing happens.

tofumatt:
feat: Add error pages

fix: Landing Page throws 404 not 500

* fixes #1033
* fixes #1605
* fixes #1616
* fixes #1673

This adds 404 and 500 pages. You can access them at `/401/`, `/404/`, and `/500/` if you're on the dev environment (not sure if that's enabled on -dev or not). You can also of course navigate to any page that should 404.

-----

### Before

screen shot 2017-02-03 at 15 29 39

screen shot 2017-02-03 at 15 29 26

screen shot 2017-02-03 at 15 29 47

### After

screen shot 2017-02-03 at 14 13 40

screen shot 2017-02-03 at 12 53 32

screen shot 2017-02-03 at 14 05 43


tofumatt:
Oddly, the page will refresh if you change the component to a string: 0b38b35a9dba6967344b38a61a87538847c9f4e4

jan-30-2017 18-53-27

src/amo/utils.js
```diff
@@ -54,3 +56,12 @@ export function loadLandingAddons({ store: { dispatch, getState }, params }) {
 
   return fetchLandingAddons({ addonType, api: state.api, dispatch });
 }
+
+export function getErrorComponent(status) {
+  switch (status) {
+    case 404:
+      return NotFound;
+    default:
+      return NotFound;
```
tofumatt: Obviously a placeholder.

kumar303:
I think the error you're seeing when rendering a component (instead of a string) is because the URL is not changing during navigation. It looks like something in the react-redux-router change is preventing the URL from changing when navigating to links?

Here is a detailed explanation: The URL is used by the add-on detail page to get the slug. When the 404 is initially shown, this is fine because it's a 404 anyway. However, when you dispatch the CLEAR_ERROR action, it attempts to re-render the add-on detail page because there is no longer an error. The first step in re-rendering is calling `mapStateToProps()` which throws an error because no slug is in the URL: DetailPage.js:24 Uncaught TypeError: Cannot read property 'guid' of undefined.

kumar303:
If you have time (today or tomorrow), we should jump on Vidyo to discuss this. I think it's going in the right direction but there could be some complications. I'd also like to see it merged into [the resource error](https://github.com/mozilla/addons-frontend/blob/master/src/core/resourceErrors/decorator.js#L27) so that fetching functions don't need to call `showError()` and I have ideas for how to do that.

kumar303:
I do want to move to react-redux-router but it feels like it could be a big change that we'll need a lot of QA on. Is there another way to do it? I'm still not sure that clearing the error when the page first loads is going to work well enough. I think it's worth exploring this idea:
* Let the resource decorator continue to handle errors
* When the error gets rendered, "pop it off the stack" (this part could be a little tricky because dispatching actions isn't really meant for this)
* On next navigation, it won't be shown

Let's chat about it in the morning.

tofumatt:
> The first step in re-rendering is calling mapStateToProps() which throws an error because no slug is in the URL: DetailPage.js:24 Uncaught TypeError: Cannot read property 'guid' of undefined.

That's interesting, helpful, and makes sense. I never saw that as an error. 😞

kumar303:
> I never saw that as an error. 

You only see it in Chrome because Firefox lost source map support (temporarily) in the new devtools UI. Sadly, I think we should all be developing in Chrome until we get source maps back.

tofumatt:
Oh shoot, terribly sorry (I did notice I didn’t see them but I never thought it’d hide the error like that. Sorry) :-(

- tofumatt

tofumatt:
> When the error gets rendered, "pop it off the stack" (this part could be a little tricky because dispatching actions isn't really meant for this)
The error state is stored in the state as well though, and passed into `handleResourceError` via its `mapStateToProps`... so marking an error page as read (which yeah is a bit weird and involves passing dispatch as a prop to the component so it can dispatch a `CLEAR` event once mounted) causes it to reload as its state has changed. Unless I'm missing something here, it will remove itself.

kumar303:
ok, yeah, that's not going to work. I think clearing the error (like you are doing) on navigation will work but it won't handle the case when the rendered page itself triggers a new error. Maybe that's fine if we are limiting this functionality to the 404/500/etc handler for now.

src/core/store.js
```diff
@@ -4,6 +4,16 @@ import { applyMiddleware, compose } from 'redux';
 import createLogger from 'redux-logger';
 import config from 'config';
 
+
+// HEY TOFUMATT THIS IS WHAT YOU WANT
+const tofuWare = store => next => action => {
```
mstriemer: Does the [`onEnter`](https://github.com/ReactTraining/react-router/blob/master/docs/Glossary.md#enterhook) hook from react-router always get called? I think you might be able to add one to the `App` route instead of switching to react-redux-router.

mstriemer:
I gave the code a quick look and reading the comments it seems like trying to clear the last error is going to suck because that's not how redux is supposed to work.

It might make sense to maintain an object of `errorId: errorObject` as well as the id of the last error. That error id (or I guess the error itself) could be cleared on navigation. Using the URL might also work but there could be some case where we add a "retry" button or something and the URL doesn't change but we want to clear the error.

Sorry if this isn't helpful, I'm a little behind here but it seemed like it might help (or maybe that's already what's happening?) I can't really tell.

src/amo/containers/App.js
```diff
@@ -77,7 +85,7 @@ export class AppBase extends React.Component {
           SearchFormComponent={SearchForm} isHomePage={isHomePage} location={location}
           query={query} ref={(ref) => { this.mastHead = ref; }} />
         <div className="App-content">
-          {children}
+          {errorPage.error ? 'errorContent' : children}
```
mstriemer: You might be looking to change this later but you could move all the error logic into a generic `ErrorHandler` component and have it selectively render its children instead.

```jsx
// <App />
<div className="App-content">
  <ErrorHandler>
    {children}
  </ErrorHandler>
</div>

// <ErrorHandler />
render() {
  const { children, errorPage } = this.props;
  if (errorPage.error /* errorPage.lastError? */) {
    return <getErrorComponent(errorPage.status) />;
  }
  return children;
}
```

tofumatt: Yeah, that's a good idea; I like that. 👍

tofumatt:
I'll note that I actually have this working locally on the client AND server and am just gonna clean it up/add tests then push. Thanks for all the help, both! 😄

src/amo/containers/App.js
kumar303: That's pretty much what we already have: https://github.com/mozilla/addons-frontend/blob/master/src/core/resourceErrors/decorator.js#L27 I know you are trying to get something simple working first as a proof though.

tofumatt:
Huzzah! This is now complete. It took a fair bit of work and also a large amount of lines solely because we needed distinct error pages for a few statuses and I thought dedicated components were nice for them (especially because we can return them elsewhere like in [`LandingPage.js`](https://github.com/mozilla/addons-frontend/pull/1685/files#diff-acdfe8ba62bf292cef622537c989648aR94))

If you want I can split the error components into their own commit/PR but essentially `amo/NotAuthorized`, `amo/NotFound`, and `amo/ServerError` (and their tests) are largely boilerplate of components/component tests. That's 326 lines of this PR, so the actual stuff that's doing anything is less than the magic 500 number. 😉 

Anyway, if you need this split up let me know, but this gets errors rendering as nice components inside App.js, renders them on the client and on the server, and sends the right HTTP Status down when they're SSR.

src/core/server/base.js
```diff
@@ -54,17 +52,52 @@ function getNoScriptStyles({ appName }) {
...
+    store.dispatch(setClientApp(res.locals.clientApp));
+  } else {
+    log.warn(`No clientApp for this ${status} error page`);
   }
-  return res.status(adjustedStatus).end(error);
-}
 
-const appName = config.get('appName');
+  store.dispatch({
+    type: REDUX_CONNECT_LOAD_FAIL,
```
tofumatt: I'm dispatching this action because we already respond to it but arguably we should have our own action that does the same thing but isn't named after a redux action–as this one is almost certainly not caused by redux-connect.

muffinresearch:
Can you check and add a screenshot to show what error pages will look like for the disco pane? Could be a follow-up  if it needs additional work, since they're generally not something a user would see for the disco pane since it's wrapped by the browser by default.

tofumatt:
Ah, yeah, I'll do a round of screenshots. 👍

src/amo/containers/Home.js
```diff
@@ -71,6 +71,9 @@ export class HomePageBase extends React.Component {
         <Link className="HomePage-themes-link" to="/themes/">
           {i18n.gettext('Browse all themes')}
         </Link>
+        <Link className="HomePage-themes-link" to="/404/">
```
kumar303: Looks like test code.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +115,22 @@ export class LandingPageBase extends React.Component {
...
 
-export function mapStateToProps(state, ownProps) {
+export function mapStateToProps(
+  state, ownProps, _apiAddonType = apiAddonType, _log = log
+) {
+  let addonType;
+  try {
+    addonType = _apiAddonType(ownProps.params.visibleAddonType);
+  } catch (err) {
+    if (err.message.match('not found in API_ADDON_TYPES_MAPPING')) {
```
kumar303: How hard would it be to throw and catch a custom error like `AddonTypeNotFound` ? I'm not sure if we've done custom error classes yet. If not, you can use [es6-error](https://github.com/bjyoungblood/es6-error) to make `instanceof` work.

src/core/server/base.js
kumar303: How about doing it just like redux-connect does it?
````js
import { loadFail } from 'redux-connect/lib/store';
import { createApiError } from 'core/api';

const apiError = createApiError({
  apiURL: res.url,
  response: { status },
});
store.dispatch(loadFail('SomeKey', apiError));
````

File: tests/client/amo/components/TestErrorPage.js [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+    const store = createStore(signedInApiState);
+    const payload = {
+      error: {
+        response: {
+          otherData: 'something',
+          status: 404,
+        },
+      },
+    };
+    store.dispatch({ type: REDUX_CONNECT_LOAD_FAIL, payload });
```
kumar303: None of the tests should dispatch this directly. They need to use the reduc-connect method:
````js
import { loadFail } from 'redux-connect/lib/store';
store.dispatch(loadFail('SomeKey', apiError));
````
This is because we need the test to ensure that our mimicry of redux-connect doesn't break.

File: src/amo/containers/Home.js (hunk 5 above)
Comment (tofumatt) [inline src/amo/containers/Home.js]: Haha, yes, oops! 😄

File: src/amo/components/LandingPage.js (hunk 6 above)
Comment (tofumatt) [inline src/amo/components/LandingPage.js]: I don't think we have a lot of (any?) custom error types yet but I bet that wouldn't be hard. I like it 👍

File: tests/client/amo/components/TestErrorPage.js (hunk 7 above)
Comment (tofumatt) [inline tests/client/amo/components/TestErrorPage.js]: Indeed, coolio I'll do that.

File: src/amo/components/LandingPage.js (hunk 6 above)
Comment (tofumatt) [inline src/amo/components/LandingPage.js]: Looking at es6-error I'm worried about the TODOs ("Better browser compatibility" and "Browser tests"). Given this site will likely be used on Firefox for iOS (right?) and I'd want it to be useful everywhere I don't want to use something that won't work there.

I'll try it as it's likely an easy patch but if it breaks in Safari or Edge that'd be annoying.

Comment (kumar303) [inline src/amo/components/LandingPage.js]: I'm not too worried about that because we use babel to compile all of our code to ES5, which has a lot of tweaks for browser safety. We use es6-error on web-ext and haven't run into any problems. Are Safari and Edge platforms we need to support?

Comment (tofumatt) [inline src/amo/components/LandingPage.js]: Ah I guess since we compile to ES5 it should be fine.

Personally I'm for maintaining broad-as-possible browser support, as in the future Edge in particular should support Web Extensions. I'm not sure if this is an explicit product goal, but the idea of us being able to serve Edge and other browsers Web Extensions seems awesome to me. And I like us being good web advocates and making stuff work everywhere 😄 

Still, should work, I'll give it a go.

**Comment by tofumatt** [comment]
Alright, fixed all the comments so far, thanks!

File: src/core/reducers/errorPage.js [hunk 8]
```diff
@@ -0,0 +1,31 @@
...
+      if (state.clearOnNext) {
+        return { ...state, ...initialState };
+      }
+      return { ...state, clearOnNext: true };
+    case REDUX_CONNECT_LOAD_FAIL:
+      return {
+        ...state,
+        error: payload.error,
+        hasError: true,
+        statusCode: payload.error.response.status,
```
kumar303: Since redux-connect will set `error` to anything that gets thrown (by, for example, `callApi()`) then you can't rely on this having the property `response`. This needs to be defensive otherwise it dies saying "error does not have a response property" or something.

I found it while testing which led me to https://github.com/mozilla/addons-frontend/issues/1701 (but that's completely unrelated!)

tofumatt: Yeah fair enough; I thought `createApiError` would fix it... but evidently not 😆

No worries, easy to be defensive about.

tofumatt: I'm going to default to `500` if I don't have a set status (inside `response`). I think that's reasonable, let me know if I'm off thinking that.

tofumatt:
Re: disco pane errors–I have an extra commit that improves them and actually moves the base ErrorPage component into core with some generic error pages that make it easier to add error handling to any app, but I'd rather submit it as a cleanup PR right after this if that's okay. It's only about 170 lines (of mostly HTML) but I don't want to add anything more to this already large-by-line PR.

So I won't add disco pane shots to this PR as to not confuse things, but I'll add in a follow-up.

### Before

screen shot 2017-02-03 at 15 29 47

screen shot 2017-02-03 at 15 29 39

screen shot 2017-02-03 at 15 29 26

### After

screen shot 2017-02-03 at 14 13 40

screen shot 2017-02-03 at 14 05 43

screen shot 2017-02-03 at 12 53 32

tofumatt:
Also I added routes active in development mode only that let devs access these errors by going to `/401/`, /404/`, and `/500/`.

src/amo/routes.js
```diff
@@ -26,6 +30,12 @@ export default (
     <Route path=":visibleAddonType/:slug/" component={CategoryPage} />
     <Route path="fxa-authenticate" component={HandleLogin} />
     <Route path="search/" component={SearchPage} />
+    <Route path="401/"
+      component={config.get('isDevelopment') ? NotAuthorized : NotFound} />
+    <Route path="404/" component={NotFound} />
```
tofumatt: I guess this is technically a bit redundant, but I like telegraphing the intent here to other developers and making it convention that we put our error pages in an easily accessible place for testing.

src/amo/utils.js
```diff
@@ -50,7 +53,23 @@ export function fetchLandingAddons({ addonType, api, dispatch }) {
...
+}
 
-  return fetchLandingAddons({ addonType, api: state.api, dispatch });
+export function getErrorComponent(status) {
+  switch (status) {
+    case 401:
+      return NotAuthorized;
+    case 404:
+      return NotFound;
+    case 500:
```
tofumatt: Again, not needed but I like to see the explicit mapping so I thought this was okay; if it's annoying or you'd like it removed I'm fine with that, though.

tests/client/amo/containers/TestApp.js
```diff
@@ -137,24 +130,6 @@ describe('App', () => {
...
-      response: { status: 404 },
-    });
-    store.dispatch(reduxConnectLoadFail('someKey', apiError));
-
-    const root = renderIntoDocument(
-      <WrappedApp store={store} />
-    );
-
-    const rootNode = findDOMNode(root);
-    assert.include(rootNode.textContent, 'Not Found');
```
kumar303: Since we're not using Flow to guarantee that `ErrorPage` is what you say it is, it would make sense to include a test that looks like this. Otherwise, it's hard to make sure that `ErrorPage` is configured in the App component correctly. It would be sort of like a "smoke test."

tests/client/amo/test_utils.js
```diff
@@ -87,5 +92,35 @@ describe('amo/utils', () => {
           mockApi.verify();
         });
     });
+
+    it('returns a rejected Promise if the addonsType is wrong', () => {
+      ownProps.params.visibleAddonType = '404';
```
kumar303: I know this value can be anything but 404 is sort of confusing. I would suggest something like "addon-with-a-typo"

tests/client/amo/test_utils.js
```diff
@@ -87,5 +92,35 @@ describe('amo/utils', () => {
...
+      const store = createStore({ application: 'android' });
+
+      return loadLandingAddons({ store, params: ownProps.params })
```
kumar303: I know that the `beforeEach` function protects it, but mutating objects like this is risky and I think an approach with cloning reads just as well:
````js
return loadLandingAddons({ 
  store, 
  params: { ...ownProps.params, visibleAddonType: '404' }, 
})
  .then(...);
````

File: tests/client/core/reducers/test_errorPage.js [hunk 14]
```diff
@@ -0,0 +1,81 @@
...
+      assert.equal(state.clearOnNext, true);
+
+      store.dispatch(endGlobalLoad());
+      state = getErrorPageState(store);
+      assert.deepEqual(state.statusCode, initialState.statusCode);
+    });
+  });
+
+  describe('REDUX_CONNECT_LOAD_FAIL', () => {
+    it('sets clearOnNext then clears it next time', () => {
```
kumar303: this spec text is the same as the previous one. Is it still accurate?

src/core/server/base.js
```diff
@@ -54,17 +53,53 @@ function getNoScriptStyles({ appName }) {
...
-  let error = getErrorMsg(adjustedStatus);
-  if (!error) {
-    adjustedStatus = 500;
-    error = getErrorMsg(adjustedStatus);
+const appName = config.get('appName');
+
+function showErrorPage(res, createStore, status, error = {}) {
+  const store = createStore();
+
+  // Get SRI for deployed services only.
```
tofumatt: Agreed, I was being a bit lazy here, I'll consolidate.

src/core/server/base.js
```diff
@@ -54,17 +53,53 @@ function getNoScriptStyles({ appName }) {
...
+    includeSri: isDeployed,
+    noScriptStyles: '',
+    sriData,
+    store,
+    trackingEnabled: convertBoolean(config.get('trackingEnabled')),
+  };
+
+  const HTML = ReactDOM.renderToString(
+    <ServerHtml {...pageProps} />);
+  const httpStatus = NestedStatus.rewind();
```
tofumatt: Fair enough. I liked the API a lot but I didn't dig too deep on how they worked. I'd be fine converting to Redux in a future patch.

src/core/server/base.js
```diff
@@ -192,6 +218,10 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
         log.warn(`No clientApp for this URL: ${req.url}`);
       }
 
+      if (!renderProps) {
+        return showErrorPage(res, createStore, 404);
```
tofumatt: Agreed actually; I was being lazy. Thanks 👍

src/core/utils.js
```diff
@@ -163,7 +164,8 @@ export function apiAddonType(addonType) {
   if (!Object.prototype.hasOwnProperty.call(
     API_ADDON_TYPES_MAPPING, addonType
   )) {
-    throw new Error(`"${addonType}" not found in API_ADDON_TYPES_MAPPING`);
+    throw new NoAddonTypeError(
```
tofumatt: agreed, wfm

src/amo/components/LandingPage.js
```diff
@@ -77,15 +79,25 @@ export class LandingPageBase extends React.Component {
...
     const {
       addonType, featuredAddons, highlyRatedAddons, popularAddons,
     } = this.props;
 
-    const html = this.contentForType(addonType);
+    let html;
+    try {
+      html = this.contentForType(addonType);
+    } catch (err) {
+      if (err instanceof NoAddonTypeError) {
```
tofumatt: Heh, I actually had it, tested it, then took it out as I thought it wouldn't be needed. Fair enough, will restore 😄

src/amo/components/NotAuthorized/index.js
```diff
@@ -0,0 +1,51 @@
...
+  static propTypes = {
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { i18n } = this.props;
+
+    const fileAnIssueText = i18n.gettext(oneLine`
+      If you are signed in and think this message is an error, please
+      <a href="https://github.com/mozilla/addons-frontend/issues/new/">file
```
tofumatt: Oops, totally true, sorry I forgot. 👍

src/amo/components/ServerError/index.js
```diff
@@ -0,0 +1,71 @@
...
+import { compose } from 'redux';
+import NestedStatus from 'react-nested-status';
+
+import Link from 'amo/components/Link';
+import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import translate from 'core/i18n/translate';
+import { sanitizeHTML, visibleAddonType } from 'core/utils';
+import Card from 'ui/components/Card';
+
+import 'amo/components/ErrorPage/ErrorPage.scss';
```
tofumatt: Yeah, I actually have that ready in a separate patch that builds on this PR (it also adds generic errors for disco pane and that can be used anywhere). But it's an extra few hundred lines and I wanted to keep this PR small. I have done exactly what you suggested in https://github.com/mozilla/addons-frontend/tree/generic-error-page; I'll make a PR with it after this if that's okay.

src/amo/components/ServerError/index.js
```diff
@@ -0,0 +1,71 @@
...
+        <Card className="ErrorPage ServerError"
+          header={i18n.gettext('Server Error')}>
+          <p>
+            {i18n.gettext(oneLine`
+              Sorry, but there was an error with our server and we couldn't
+              complete your request. We have logged this error and will
+              investigate it.`)}
+          </p>
+
+          <h2>{i18n.gettext('Suggested Pages')}</h2>
```
tofumatt: Touché.

src/core/reducers/errorPage.js
```diff
@@ -0,0 +1,42 @@
...
+  clearOnNext: false,
+  error: null,
+  hasError: false,
+  statusCode: null,
+};
+
+export default function errorPage(state = initialState, action) {
+  const { payload } = action;
+  switch (action.type) {
+    case REDUX_CONNECT_END_GLOBAL_LOAD:
```
tofumatt: I believe I tried that and the issue is that then it would clear error state and try to render the non-error-page content and error. I'll double-check, but I think I needed to use `END_GLOBAL_LOAD` to prevent errors from throwing.

tests/client/amo/components/TestLandingPage.js
```diff
@@ -101,9 +101,30 @@ describe('<LandingPage />', () => {
...
+  it('renders not found if add-on type is not supported', () => {
+    const root = render({ addonType: 'XUL' });
+    assert.include(root.textContent, 'Page not found');
+  });
+
+  it('renders not found if add-on type is not supported', () => {
+    assert.throws(() => {
+      // Stubbing in a fake i18n here will cause contentForType to throw.
+      // This is easier/lazier than doing dependency injection for that method.
+      render({ addonType: 'doesntmatter', i18n: null });
```
tofumatt: Yeah, it's just a fair bit more code. I should know any time I write a two-line comment to explain a lousy test I should just Do It Right™.

src/amo/components/ServerError/index.js
kumar303: ah, yeah, doing it in another patch is cool

src/amo/components/ServerError/index.js
kumar303: http://touchechicago.com/

src/amo/containers/App.js
```diff
@@ -9,17 +9,18 @@ import { compose } from 'redux';
 import 'core/fonts/fira.scss';
 import 'amo/css/App.scss';
 import SearchForm from 'amo/components/SearchForm';
+import ErrorPage from 'amo/components/ErrorPage';
```
kumar303: I should have mentioned that the reason I like it is you can read the `<ErrorPage>` code (i.e. where the component is used) more naturally.

src/core/reducers/errorPage.js
kumar303: oh, right, that sounds familiar from when I was testing an earlier patch

tests/client/amo/components/TestLandingPage.js
kumar303: You might be able to delete this test if you end up moving the visibleAddonType logic around as I suggested but I'm not sure

src/amo/components/ServerError/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+export class ServerErrorBase extends React.Component {
+  static propTypes = {
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { i18n } = this.props;
+
+    const fileAnIssueText = i18n.gettext(oneLine`
```
tofumatt: The text in each message is a bit tailored to the error, so it seems like more work to create a component that is essentially just filling in the blanks of `i18n.sprintf`... which admittedly isn't much of a lovely API, but adding a shared component for this seems overkill (it will be a lot more code/tests for something easy to do):

```js
const fileAnIssueText = i18n.sprintf(i18n.gettext(oneLine`
      If you are signed in and think this message is an error, please
      <a href="%(url)s">file an issue</a>. Tell us where you came from
      and what you were trying to access, and we'll fix the issue.`),
      { url: 'https://github.com/mozilla/addons-frontend/issues/new/' });
```

src/amo/components/ServerError/index.js
tofumatt: Actually, this is a good idea but mind if I do it in my consolidating-these-errors-into-a-shared-component follow-up PR?

src/amo/components/ServerError/index.js
kumar303: oh, my bad, I didn't realize the text was different

src/amo/components/ServerError/index.js
kumar303: sure, whatever is easiest

tests/client/amo/components/TestServerError.js
```diff
@@ -0,0 +1,42 @@
...
+    ), ServerError));
+  }
+
+  it('renders a server error', () => {
+    const rootNode = render();
+
+    assert.include(rootNode.textContent,
+      'but there was an error with our server and');
+  });
+});
```
tofumatt: These components are only rendered by `ErrorPage` if there is an error in state, so they would never render children without an error. https://github.com/mozilla/addons-frontend/pull/1685/files/0631c06567d37732491ffab0e69193ffc04960ec#diff-422611d6ba2ee20db53e3927e61b1ca6R29 covers `ErrorPage` doing that.

tests/client/amo/components/TestLandingPage.js
```diff
@@ -102,29 +105,16 @@ describe('<LandingPage />', () => {
...
-      render({ addonType: 'doesntmatter', i18n: null });
+      render({
+        _apiAddonType: null,
+        params: { visibleAddonType: 'doesnotmatter' },
+      });
     });
   });
-
-  it('throws an error if a different error is encountered ', () => {
-    const fakeError = new Error('Ice Cream Error');
```
tofumatt: Sad I have to get rid of this 🍦

src/amo/components/NotAuthorized/index.js
```diff
@@ -0,0 +1,51 @@
...
+      <NestedStatus code={401}>
+        <Card className="ErrorPage NotAuthorized"
+          header={i18n.gettext('Not Authorized')}>
+          <p>
+            {i18n.gettext(oneLine`
+              Sorry, but you aren't authorized to access this page. If you
+              aren't signed in, try signing in using the link at the top
+              of the page.`)}
+          </p>
+
```
tofumatt: Sadly not; tried and it failed the linter 😢

tofumatt:
Okay, sorry I forgot to get to this on the weekend and then was late on this today, but I addressed all the PR issues (I think) so good for another r?

tests/client/amo/components/TestServerError.js
kumar303: oh my god duh, I think my head was about to explode when I left that review comment

tests/client/amo/test_utils.js
```diff
@@ -9,15 +12,17 @@ import {
...
   loadLandingAddons,
 } from 'amo/utils';
+import { unexpectedSuccess } from 'tests/client/helpers';
 
 
 describe('amo/utils', () => {
   let ownProps;
 
-  beforeEach(() => {
+  before(() => {
```
tofumatt: Err, yes. True. I'm not sure why I didn't do that, haha.

tests/client/amo/components/TestLandingPage.js
```diff
@@ -101,9 +102,17 @@ describe('<LandingPage />', () => {
     );
   });
 
-  it('throws if add-on type is not supported', () => {
+  it('renders not found if add-on type is not supported', () => {
+    const root = render({ params: { visibleAddonType: 'XUL' } });
+    assert.include(root.textContent, 'Page not found');
+  });
+
+  it('throws an error if something other than an unknown addonType', () => {
```
tofumatt: Huzzah, that is much better 👍

src/core/server/base.js
```diff
@@ -160,55 +197,25 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
     ) => {
       cookie.plugToRequest(req, res);
 
-      if (err) {
```
tofumatt: Yeah, sorry, there isn't test coverage for the server code so I did a fair bit of manual stuff but haven't written many tests for it yet. I'll add that back in though as after I refactored the code it's easy to do. Thanks for catching it.

src/amo/components/LandingPage.js
```diff
@@ -5,36 +5,50 @@ import { asyncConnect } from 'redux-connect';
...
   ADDON_TYPE_THEME,
   SEARCH_SORT_POPULAR,
   SEARCH_SORT_TOP_RATED,
 } from 'core/constants';
-import { apiAddonType, visibleAddonType } from 'core/utils';
+import { AddonTypeNotFound } from 'core/errors';
+import log from 'core/logger';
+import {
+  apiAddonType,
+  visibleAddonType as _visibleAddonType,
```
tofumatt: Indeed, that works for me. Because we use `visibleAddonType` as a variable this got messy 😒

tests/client/amo/components/TestLandingPage.js
```diff
@@ -101,9 +102,17 @@ describe('<LandingPage />', () => {
...
     assert.throws(() => {
-      render({ addonType: 'XUL' });
-    }, 'No LandingPage content for addonType: XUL');
+      render({
+        _apiAddonType: null,
+        params: { visibleAddonType: 'doesnotmatter' },
+      });
+    });
```
tofumatt: The return of 🍦!

tofumatt:
Fixes made, should be good 😄

Issue #1546, opened 2016-12-30
Long author names float icon too far to the right


tofumatt:
### Describe the problem and steps to reproduce it:
Have a long author name in the search results

### What happened?
The icon is pushed (partially) off-screen.

### What did you expect to happen?
The author name should wrap and not affect icon positioning.

### Anything else we should know?
Seems I made this with floats but really should use flexbox; I'll update the styles and it should fix it.

### Screenshot

<img width="308" alt="screenshot 2016-12-30 17 29 01" src="https://cloud.githubusercontent.com/assets/90871/21569542/aa1dce0c-ceb5-11e6-98aa-841e3e0fd772.png">

aniketkudale:
I'd like to work on this!

tofumatt:
Hi @aniketkudale, sounds good! We can't assign issues to people outside the organisation on GitHub, but feel free to work on this and send a PR when it's ready.

Ryuno-Ki:
@tofumatt Maybe introducing a label as workaround? Something like __assigned__?

aayushsanghavi:
@tofumatt any progress on this issue? If not, would it be fine if I worked on it?

tofumatt:
@aayushsanghavi No progress yet, if you wanted to work on it that would be awesome! 👍

mstriemer:
If someone's feeling ambitious there is a similar bug on the add-on detail page which references the same add-on #1607.

PR #1699, opened 2017-02-02
Long author names float icon too far to the right

aayushsanghavi:
#1546 
Using div wrappers around text content and image, flexbox styling and word wrap.

Before - 
before
After - 
after


src/core/css/SearchResult.scss
```diff
@@ -73,11 +81,4 @@
...
   display: flex;
-}
-
-.SearchResult-icon {
-  display: inline-block;
-  float: right;
-  height: 38px;
-  width: 38px;
-}
+}
```
tofumatt: Super nitpick here but we like to have newlines at the end of files; often you can get your text editor to add these automatically when saving. Helps to avoid wondering if a file prematurely ends 😄

src/core/css/SearchResult.scss
```diff
@@ -23,7 +23,15 @@
...
 .SearchResult-main {
+  display: flex;
+  flex-direction: row;
   flex-grow: 1;
+  justify-content: space-between;
+  width: 100%;
+}
+
+.SearchResult-text {
+  width: 90%;
```
tofumatt: Where did you arrive at these percentages? We tend to use `px` values in the app and not `%`.

aayushsanghavi: I used % because what I've observed is that it makes responsiveness better compared to when px values are used. I've faced issues earlier in my projects with resizing when using pixels but that becomes fine with the use of %. So following that notion I used %.

src/core/css/SearchResult.scss
aayushsanghavi: Haha okay. I'll add that. Will keep it in mind next time.

tofumatt: Heh, not that nitpicky apparently: https://travis-ci.org/mozilla/addons-frontend/jobs/197795481#L340

You can run `npm test` before you push your PR to get test results; it will include style linting like this so you don't have to make as many changes after pushing 😄

aayushsanghavi: Okay. Thanks for suggesting that. Will make sure to run that next time.

src/core/css/SearchResult.scss
tofumatt: It does, though for now we've been using `px margin` over `% width` as it keeps the margins consistent across devices and means content can fill a lot of the screen. Plus we have a fairly different design coming up for desktop that will involve a fair number of style overrides/differences anyway, so for now we are keeping things responsive by basically just scaling up the mobile view on desktop, as this v1 is planned to only be the default for mobile devices anyway.

Your point is valid but for this project, for now, please use `px` similar to what was used before to keep things from being full-width. 😄

aayushsanghavi: I tried using pixels but it fails below a certain browser width. Should I use a media query for that?

src/core/css/SearchResult.scss
aayushsanghavi: I changed the width of SearchResult-text to px and added media queries to make it responsive.

muffinresearch:
This might be an over-simplification but I think you achieve the same outcome with this and no markup changes:

```css
// Ensures this element doesn't go wider than its parent.
.SearchResult-main  {
   max-width: 100%;
}

.SearchResult-author {
   word-wrap: break-word;
   display: block;
}
```

aayushsanghavi:
@muffinresearch yes, you are right. That does work out and it definitely makes the styling simpler. Thank you for your suggestion.

aayushsanghavi:
Before - 
before
After - 
after

src/core/css/SearchResult.scss
```diff
@@ -59,6 +60,8 @@
 
 .SearchResult-author {
   margin-right: 0.5em;
+  word-wrap: break-word;
+  display: block;
```
muffinresearch: Generally we try to alpha-sort descriptors e.g. `display` before `margin...`. It's not always appropriate - which makes it difficult to use a style-lint rule to automatically flag it - but generally it does aid readability.

muffinresearch: Also I appreciate my example wasn't in order too, so my bad!

muffinresearch:
@aayushsanghavi  great, I'm glad that worked out.

Feel free to update the overall description and screenshots to match the patch changes. Also adding `Fixes` before the `#ISSUENUM` will link this up when it lands.

For bonus points you could manually rebase this branch and squash your commits [1]. It just helps keep the PR tidy and limited to the main changes you're planning to land. If you get stuck then leave it and we can always squash when we merge this.

[1] http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

tofumatt:
Tested this locally and it works well, thanks! (This solution is easier, sorry for leading you down the flexbox road in the issue summary I made 😆)

I'll alphabetize the order and squash up the commits myself just to get this merged. Thanks @aayushsanghavi! 👍

aayushsanghavi:
@muffinresearch thanks for those tips. I'll keep them in mind. I'm learning a lot of new things here on mozilla.
@tofumatt I appreciate your help with that. Thanks!

Issue #1701, opened 2017-02-02
callApi() on a 404 URL throws TypeError: Body has already been consumed.


kumar303:
### Describe the problem and steps to reproduce it:

Make a component that does `callApi()` with a URL that is a 404

### What happened?

In Firefox, it throws:
````
TypeError: Body has already been consumed.
````
and something in similar in Chrome.

### What did you expect to happen?

It should throw a well-formed `apiError` instead.

### Anything else we should know?

Well, first, the 404 response is not in JSON (see https://github.com/mozilla/addons-server/issues/4576) but that's OK. The `callApi()` was designed to handle this by falling back to `.text()`, but this is what triggers the error.

This is kind of a fetch gotcha. The solution is to clone the stream instead, like this:
````js
fetch(url).then(function(response) {
  return response.clone().json().catch(function() {
    return response.text();
  });
});
````
(example from https://jakearchibald.com/2015/thats-so-fetch/)

Issue #1473, opened 2016-12-11
Move all promises to async / await syntax


kumar303:
Promises can be very hard to read, especially when we need to do specific exception handling that should not affect the rest of the promise chain. Let's consider *fully* moving to [async / await](https://ponyfoo.com/articles/understanding-javascript-async-await) instead. 

You can grep for `\.then` in `src` to get an idea of where there are still promises. You will notice a few examples where we will still need to use promises, such as a top-level call like [createClient()](https://github.com/mozilla/addons-frontend/blob/473d5c9b267fe85e9fdc68f88884f0eee8e930a3/src/amo/client.js#L10).

Contributors: I recommend beginning with a small patch where you only fix one or two instances.

aayushsanghavi:
@kumar303 I would like to take a shot at this. Could you tell me where all we need to change the code to `async-await`?

tofumatt:
You would need to convert all Promises code over to the async/await syntax. This might also involve learning flow given we annotate some Promises code with flow. But going through the Promises code would be a good start, as that's what we want to convert to use async/await.

aayushsanghavi:
Could you link some files that use promises?

tofumatt:
Unfortunately I won't be able to link to all the files; a fair bit of the work in this patch is the task of identifying all the Promises code–once I'd have found them all it would be easier for me to complete the patch than link to all the files.

You can look through the codebase with an editor like [Atom](https://atom.io/) for Promises code (a good sign in `.then(`.

Are you familiar with Promises and with async/await?

aayushsanghavi:
Haha okay. I'll look up the files first, then.
Neither am I familiar with async/await nor with promises but I hope to learn about them through this issue. I'll read up on it from the link @kumar303 left in the description. Then I'll try working on the code and file a PR if possible. If it doesn't work out, I'll drop it,

Sounds good?

tofumatt:
That sounds fine, yup! Always feel free to try working on something, cheers.

aayushsanghavi:
Hey @tofumatt we've got to look for `.then` in the `src` directory only right? I happened to search it in the root directory of the project and got an astonishingly huge number of results. I guess thats not what we want right?

tofumatt:
`src/` and `tests/` at least, I'm not sure if we use them elsewhere. You'll want to ignore `node_modules`, obviously, as it will be full of third-party libraries using Promises.

Razaik25:
[...]

Issue #1252, opened 2016-10-18
Relax eslint arrow-body-style rule


kumar303:
The [arrow-body-style](http://eslint.org/docs/rules/arrow-body-style) rule makes us prefer to use arrow shortcuts when possible. However, I think it makes long promise chains harder to read and harder to refactor.

Take this example:

``` javascript
it('requires URL params',
  () => loadAddonReview({ params: {} })
    .then(() => assert(false, 'unexpected success'))
    .catch((error) => {
      assert.match(error.message, /missing URL params/);
     })
);
```

I think it would be easier to follow the promise chain if written with a return statement. It's also a lot easier to refactor when you need to declare variables at the top.

``` javascript
it('requires URL params', () {
  return loadAddonReview({ params: {} })
    .then(() => assert(false, 'unexpected success'))
    .catch((error) => {
      assert.match(error.message, /missing URL params/);
     });
});
```

mstriemer:
You could probably use `()` to make the formatting a bit nicer then if you want to add multiple lines you just switch them to `{}` and add the return.

This could create some ambiguity around whether something is being returned or not but it does format more nicely in my opinion.

```js
it('requires URL params', () => (
  loadAddonReview({ params: {} })
    .then(() => assert(false, 'unexpected success'))
    .catch((error) => {
      assert.match(error.message, /missing URL params/);
     })
));

tofumatt:
I really like seeing the explicit returns in JS myself as well; I don't feel like implicit returns are a thing in JS like they are in Ruby, so I like always seeing the explicit one. I'd be for this rule being enabled 👍

Ryuno-Ki:
Yepp, implicit returns in Ruby made me scratch my head in the past. Especially when debugging oO

PR #1702, opened 2017-02-02
Handle non-JSON responses in callApi() correctly

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1701

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -67,7 +67,7 @@ export function callApi({
           (jsonResponse) => ({ response, jsonResponse }),
           (error) => {
             log.warn('Could not parse response as JSON:', error);
-            return response.text().then((textResponse) =>
+            return response.clone().text().then((textResponse) =>
               ({ response, jsonResponse: { text: textResponse } })
             );
           }
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -1,4 +1,5 @@
 /* global fetch */
+/* eslint-disable arrow-body-style */
 
 import url from 'url';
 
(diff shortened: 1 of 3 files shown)


src/core/api/index.js
```diff
@@ -60,15 +61,18 @@ export function callApi({
...
-      (jsonResponse) => ({ response, jsonResponse }),
-      (error) => {
-        log.warn('Could not parse response as JSON:', error);
-        return response.text().then((textResponse) =>
-          ({ response, jsonResponse: { text: textResponse } })
+    .then((response) => {
+      return response.clone().json()
+        .then(
+          (jsonResponse) => ({ response, jsonResponse }),
+          (error) => {
```
tofumatt: I feel like I've asked this before for `fetch` stuff but I forget: any reason to use the much-less-seen error argument over `.catch`?

src/core/api/index.js
```diff
@@ -1,4 +1,5 @@
 /* global fetch */
+/* eslint-disable arrow-body-style */
```
tofumatt: I'm guessing this is for https://github.com/mozilla/addons-frontend/pull/1702/files#diff-9c6af2bc9310bfed1e429d02f827c559R65? If so, maybe this should just be a `disable-next-line`?

It's unrelated but I'd love to switch over to `arrow-body-style` always; I find it so much easier to scan.

kumar303: You're going to fight me for it! It's about to die https://github.com/mozilla/addons-frontend/issues/1252 I generally agree that single value returns should always use arrow body but long promise chains get really hard to read. In fact, when I first submitted the code for review, mark had trouble with it because it was hard to follow the `response` variable reference.

src/core/api/index.js
kumar303: The same code won't work with catch. If I put a catch above this and recover, it will execute the `jsonResponse` then function. I don't think there's another way to do it.

These kind of patterns is why promise chains are terrible! If we move to async/await (https://github.com/mozilla/addons-frontend/issues/1473) everything will be way easier to read.

src/core/api/index.js
tofumatt: Oh maybe I was unclear, I actually would prefer we used explicit `return` *always*; I dislike the implicit return syntax and invariable refactoring of the code once it becomes multi-line/statement.

Anywho, this is better, I was just nitpicking about disabling a rule whole-file over line, really 😄

src/core/api/index.js
tofumatt: Ahh, gotcha. Yeah the fetch API is a bit odd. I need to learn the async/await syntax, though it def looks nicer 😄

tofumatt:
I think these test failures are just `servertest` related but I'm not sure. r+ to the code but be sure it passes tests; I re-ran them a few times last night and they're still failing...

src/core/api/index.js
kumar303: oh, I see what you mean. Well, I like doing things like `someArray.filter((value) => value === true)`. I wouldn't want to give that up.

Issue #1526, opened 2016-12-22
Add title to review entry, update styles


kumar303:
The review entry form has changed a bit in the latest Sketch file. We need to bring the style up to date and also add title entry which wasn't there before.

Here is the mock but check the Sketch file for the latest:

![screenshot 2016-12-22 15 10 00](https://cloud.githubusercontent.com/assets/55398/21440509/ee00c128-c858-11e6-8631-7f8e09b21839.png)

vcarciu:
With keyboard open, the scroll should work inside _Write a review_ form , because else you will not be able to submit the review : 
![scrollreview](https://cloud.githubusercontent.com/assets/1583842/22787913/9ade03e8-eee6-11e6-9072-9085aa3618fa.gif)

kumar303:
@vcarciu D'oh. Good catch, thanks. Can you open a new bug for this? That will make it easier to track because https://github.com/mozilla/addons-frontend/issues/1739 is about to land soon as well (which will remove the title).

vcarciu:
@kumar303 : logged https://github.com/mozilla/addons-frontend/issues/1773

kumar303:
Thanks

Issue #1156, opened 2016-09-28
Show data loading indicator for overall rating submission


kumar303:
While the user is submitting an [overall rating](https://github.com/mozilla/addons-frontend/issues/1017) the UI should show some kind of loading indicator. Maybe a simple spinner would be enough.

UX flow tracker: #927

muffinresearch:
Relates to https://github.com/mozilla/addons-frontend/issues/1219

kumar303:
This should get fixed by https://github.com/mozilla/addons-frontend/issues/2179

Ryuno-Ki:
As I said in https://github.com/mozilla/addons-frontend/issues/2179#issuecomment-289306398 watch out for a11y support.

ValentinaPC:
Verified this on AMO-dev FF56(Win 7)
While hovering over the rating stars, no matter if a review was already submitted or a new one is made, any selection I made, same info is displayed, "no ratings" in my case.
Please see the video:
![no rating](https://user-images.githubusercontent.com/15685960/31070711-0c6f755a-a76a-11e7-9cdc-07a4bc0cebb6.gif)

Am I missing something?

kumar303:
The 'No ratings' label is confusing and will be fixed in https://github.com/mozilla/addons-frontend/issues/3231

As for the original issue here, I'm re-opening it because there is still no visual indication that submitting a rating is *in progress*. For slow connections, this may cause confusion and lead to double submissions. 

All in all, it's not an urgent bug but I think we could do something here like render the stars read-only or something.

kumar303:
Now that we have some production usage, submitting a review is a bit slow (almost one second on average). We should definitely add a loading indicator.

![amo prod frontend apis usage performance datadog](https://user-images.githubusercontent.com/55398/32347998-7a30a224-bfe0-11e7-919c-a9229cd6b060.png)

PR #1706, opened 2017-02-04
Use a popover to enter in the review text

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1526

First, this fixes handling fatal fetch() errors which we somehow didn't know were not handled:

screenshot 2017-02-03 18 12 50

More importantly, the review entry screen is now up to spec (I don't think the focus lines will look the same on mobile but we probably need to keep these for accessibility):

screenshot 2017-02-03 18 13 56

screenshot 2017-02-03 18 14 18


mstriemer:
Unrelated nit: can the background red be changed to one of the reds in the palette from the mocks. `#f00` is a little jarring.

screen shot 2017-02-03 at 6 25 48 pm

src/amo/components/AddonReview.js
```diff
@@ -19,40 +17,52 @@ export class AddonReviewBase extends React.Component {
...
-    router: PropTypes.object.isRequired,
     updateReviewText: PropTypes.func,
   }
 
-  onSubmit = (event) => {
+  constructor(props) {
+    super(props);
+    this.state = { reviewTitle: null };
+    if (props.review) {
+      // TODO: also do this in componentWillReceiveProps.
```
kumar303: oops, I need to do this

src/ui/components/Overlay/index.js
```diff
@@ -39,8 +49,12 @@ export default class Overlay extends React.Component {
       <div className={classNames('Overlay', className, {
         'Overlay--visible': this.state.visible,
       })} ref={(ref) => { this.overlayContainer = ref; }}>
-        <div className="Overlay-background"
-          ref={(ref) => { this.overlayBackground = ref; }} />
+        <div
+          onClick={this.onClickBackground}
+          ref={(ref) => { this.overlayBackground = ref; }}
+          onTouchEnd={this.onClickBackground}
```
kumar303: note to self: this needs a test.

kumar303:
Oh yeah, totally. I didn't know we had that in the palette.

src/ui/components/Overlay/index.js
```diff
@@ -39,8 +49,12 @@ export default class Overlay extends React.Component {
       <div className={classNames('Overlay', className, {
         'Overlay--visible': this.state.visible,
       })} ref={(ref) => { this.overlayContainer = ref; }}>
-        <div className="Overlay-background"
-          ref={(ref) => { this.overlayBackground = ref; }} />
+        <div
+          onClick={this.onClickBackground}
+          onTouchEnd={this.onClickBackground}
```
kumar303: I'm actually not sure if I need this. It looks like React maybe handles this internally for `onClick` and even works around some iOS bugs. I guess I should try it and see.

mstriemer: A tap will fire the touch events but also the mouse events including click (unless you `preventDefault` somewhere higher in the chain, with click being the last event)

src/amo/components/AddonReview.js
tofumatt: Still need to 😉

src/amo/components/AddonReview.js
```diff
@@ -19,40 +17,52 @@ export class AddonReviewBase extends React.Component {
...
+      this.state.reviewTitle = props.review.title;
+      this.state.reviewBody = props.review.body;
+    }
+    this.overlayCard = null;
+    this.reviewTextarea = null;
+    this.reviewTitleInput = null;
+  }
+
+  onSubmit = (event, { overlayCard = this.overlayCard } = {}) => {
```
tofumatt: Generally, @mstriemer advocated for splitting all our `onX` methods apart into two things: the `onSubmit` type thing that would take the event as the argument and then another method (here, I guess it'd be `updateReview`) that actually had all the logic. It makes it easy to know what to name the event handlers themselves and means multiple handlers can call the same function. It should means testing the logic of the reviewupdate method doesn't need a stubbed out event.

Short version is we've been keeping the event handler and logic in separate methods which is a bit more code but easier to follow in my and @mstriemer's opinion. Could we do that here?

src/amo/components/AddonReview.js
```diff
@@ -19,40 +17,52 @@ export class AddonReviewBase extends React.Component {
...
-      body,
+      body: reviewBody,
+      title: reviewTitle,
       addonSlug: this.props.review.addonSlug,
       errorHandler: this.props.errorHandler,
       reviewId: this.props.review.id,
       apiState: this.props.apiState,
-      router: this.props.router,
     };
     // TODO: render a progress indicator in the UI.
```
tofumatt: I think there's already an issue for this, you could add it to the comment if you like 😄

(I think it's https://github.com/mozilla/addons-frontend/issues/1156, right?)

src/amo/components/AddonReview.js
```diff
@@ -74,31 +84,38 @@ export class AddonReviewBase extends React.Component {
       );
     }
 
+    const titlePlaceholder = i18n.gettext('Give your review a title');
```
tofumatt: Probably separate but I find it so weird we require titles in reviews; it'd be nice if failing receiving one we just extracted the first sentence or ignored it. I always hate being asked for titles in my reviews; makes me feel like I'm doing my High School English papers again.

src/amo/components/AddonReview.js
```diff
@@ -74,31 +84,38 @@ export class AddonReviewBase extends React.Component {
...
-              type="submit" value={i18n.gettext('Submit')}
-            />
+              ref={(ref) => { this.reviewTitleInput = ref; }}
+              type="text"
+              className="AddonReview-title"
+              name="reviewTitle"
+              value={reviewTitle}
+              onInput={this.onTitleInput}
+              placeholder={titlePlaceholder} />
+            <textarea
```
tofumatt: We generally put as many props as possible on a single line (it's even [done below](https://github.com/mozilla/addons-frontend/pull/1706/files#diff-8520bc744fd8142c8f1e5d422038bb41R115)), any reason these are all newline'd?

src/amo/css/AddonReview.scss
```diff
@@ -1,52 +1,51 @@
...
-  // Code point: &#12296;
-  // TODO: switch this to an SVG.
-  content: '〈 ';
+.AddonReview-textarea {
+  min-height: 200px;
 }
 
-.AddonReview-button {
+.AddonReview-submit {
+  font-size: 0.8em;
```
tofumatt: While I see a few in this file we have tended to use `px` values instead of `em` most of the time, and we have a lot of sizes in `vars.scss` for font-sizes. If you could use them that would be more consistent/easier to adjust later.

src/amo/css/AddonReview.scss
```diff
@@ -1,52 +1,51 @@
...
+.AddonReview-form-input {
+  display: flex;
+  flex-direction: column;
+  margin: 0;
+  padding: 0 10px;
 }
 
 .AddonReview-header {
-  font-size: $font-size-l;
+  font-size: $font-size-m + 8px;
```
tofumatt: This seems odd, makes me wonder why we have the variable. Was this for testing or is the large too large?

src/core/errorHandler.js
```diff
@@ -31,41 +36,57 @@ export class ErrorHandler {
 
 class ErrorHandlerComponent extends React.Component {
   static propTypes = {
+    autoRenderErrors: PropTypes.boolean,
```
tofumatt: Because we have a defaultProp this should be marked as `isRequired`, right?

src/core/errorHandler.js
```diff
@@ -31,41 +36,57 @@ export class ErrorHandler {
...
+        </ul>
+      );
+    }
+
+    const errorHandler = new ErrorHandler({
+      errorContent,
+      id: errorHandlerId,
+      dispatch,
+    });
+    const allProps = { ...props, errorHandler };
```
tofumatt: I'd be fine with just writing `<WrappedComponent { ...props, errorHandler } />` below, but it's up to you 😄

tofumatt: Oh, never mind, I see it's called twice; this makes sense.

src/ui/components/OverlayCard/OverlayCard.scss
```diff
@@ -1,8 +1,9 @@
 @import "~core/css/inc/vars";
 
 .OverlayCard {
-  margin: 0 $padding-page;
-  width: 100%;
+  margin: 0;
+  width: 90%;
```
tofumatt: Why this change? We tend not to use `%` widths and instead `px` margins to keep the margins in-line with the mocks.

tests/client/ui/components/TestOverlay.js
```diff
@@ -39,6 +39,18 @@ describe('<Overlay />', () => {
     assert.include(root.overlayContainer.className, 'Overlay--visible');
   });
 
+  it('hides when you click the background', () => {
```
tofumatt: Seems like an `Esc` `keyPress` would also be appropriate, especially for accessibility as I'm not sure there is a "close window" button.

That's my bad as I wrote the component, but I'm just realising it now.

tests/client/ui/components/TestOverlay.js
```diff
@@ -39,6 +39,18 @@ describe('<Overlay />', () => {
...
+    const root = render({ visibleOnLoad: true });
+    Simulate.click(root.overlayBackground);
+    assert.notInclude(root.overlayContainer.className, 'Overlay--visible');
+  });
+
+  it('hides when you touch the background', () => {
+    const root = render({ visibleOnLoad: true });
+    Simulate.touchEnd(root.overlayBackground);
```
tofumatt: Sometimes I forget with things like `touchEnd`, but if a touch starts outside the background but finishes on it, will it still hide the background? Because that seems wrong.

I'd have thought `onClick` was enough here, but maybe I'm wrong.

src/amo/components/AddonReview.js
mstriemer: I think I prefer the new line for each prop if it doesn't all fit version. The trailing `/>` is good by me too but I do find it a little odd still.

```jsx
            <textarea 
              ref={(ref) => { this.reviewTextarea = ref; }}
              className="AddonReview-textarea"
              onInput={this.onBodyInput}
              name="review"
              value={reviewBody}
              placeholder={placeholder}
            />

File: src/core/css/inc/vars.scss [hunk 15]
```diff
@@ -8,7 +8,7 @@ $text-color-default: #333;
 $text-color-headings: $text-color-default;
 $text-color-message: #fff;
 $sub-text-color: #6a6a6a;
-$form-border-color: #f1f1f1;
+$form-border-color: $primary-font-color;
```
mstriemer: 👏

src/core/errorHandler.js
mstriemer: The eslint rule we're blocked on upgrading on ([require-default-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md)) is asking for the opposite I think. If it isn't required it should have a default.

src/core/errorHandler.js
```diff
@@ -31,41 +36,57 @@ export class ErrorHandler {
...
+
+    if (error && autoRenderErrors) {
```
mstriemer: Does passing `autoRenderErrors = false` prevent this from rendering errors? This seems odd to me.

src/amo/components/AddonReview.js
kumar303: I dislike abstraction for the sake of abstraction. If two different event handlers needed to call `updateReviewText()` then that would be a good reason to split it into a new function but for now there's only one. Another reason to split it out would be to make testing easier but that's not a problem here.

src/amo/components/AddonReview.js
kumar303: I agree! https://github.com/mozilla/addons-frontend/issues/1739

src/amo/components/AddonReview.js
kumar303: When there are more than 4 properties, I find them easier to scan if each is on a new line. I can balance the closing tags on these, that looks a bit better.

src/amo/css/AddonReview.scss
kumar303: I was just trying to get it looking like the mock. I'll add a new variable for it.

src/amo/css/AddonReview.scss
kumar303: Yeah, true. I switched to a font size variable. I think `em` still makes sense for the margins and paddings because that makes them all relative to font-size.

src/core/errorHandler.js
kumar303: when it has a default value, it doesn't make sense to mark it as required because otherwise the default value would never get used.

tofumatt: Hah, okay, that makes sense but I only thought it through halfway and took `isRequired` in this case to be more/less a developer note that it requires a `defaultProp`. I should fix some of my components that use `isRequired` when they shouldn't, in that case. 😨

src/core/errorHandler.js
kumar303: Yes, you are correct. By passing `autoRenderErrors = false`, the component is responsible for rendering errors itself. However, if the component forgets to do so then that could be confusing. I will log a warning so that at least there's something in the console.

This feature actually came up in the original review but I wanted to wait until we absolutely needed it. Now we need it!

tests/client/ui/components/TestOverlay.js
kumar303: yeah, I think onClick is enough. I'll remove the touch handler.

tests/client/ui/components/TestOverlay.js
kumar303: hmm, maybe. We should get some accessibility feedback on the overlay for sure.

src/core/errorHandler.js
mstriemer: I think it makes more sense to split out the rendering and the display than to pass something to this one entry point to tell it to only do one or the other.

You could pull the `connect` bits out into a `withErrorHandler` function which could be used by `withErrorHandling` and this component instead of telling the HOC to skip rendering errors.

src/core/errorHandler.js
kumar303: meh, propTypes are a joke. I would rather spend time on porting us to Flow :D

src/core/errorHandler.js
```diff
@@ -31,41 +36,63 @@ export class ErrorHandler {
...
+    let errorContent;
     if (error) {
-      return (
-        <div>
-          <ul className="ErrorHandler-list">
-            {error.messages.map((msg) => <li>{msg}</li>)}
-          </ul>
-          {wrappedOutput}
-        </div>
+      errorContent = (
```
mstriemer: I think you could move this inside of `ErrorHandler` to put the logic closer to `getError()`.

mstriemer:
I really don't think adding `autoRenderErrors` is a good idea. It's tying the error handling logic more tightly with there being one component that renders its errors locally. You could move the JSX into `ErrorHandler` which would allow a component to render the error anywhere but I think having a different component and passing it the error makes the most sense.

It really seems to me that these parts should be broken up rather than having options added for only running part of the logic. There is a wrapper that creates an `ErrorHandler` and passes it in as props, a component that renders an error, and a HOC that puts the two together.

If these were all separate parts then I think it would be possible to render the authenticate button's errors in the main content area of the page in #1707.

src/core/errorHandler.js
kumar303: Good idea, fixed

mstriemer:
This is essentially what I [suggested back when error handling was originally added](https://github.com/mozilla/addons-frontend/pull/1418#discussion_r90366987). Then I thought we may as well let the component do both, but if the wrapper is what's easiest we can have the best of both worlds.

> If you move creating the error handler to `mapDispatchToProps()` then you could remove this component and just return the `connect()` call on the wrapped component. To render the errors you could expose the `errorInfo` bit above as `ErrorList` and that could be imported instead. That would let the caller put the error messages wherever makes sense.
>
> ```jsx
> import { ErrorList, withErrorHandling } from 'core/errorHandler';
> class AllReviewsBase extends React.Component {
>   render() {
>     const { error, reviews } = this.props;
>     return (
>       <div className="AllReviews">
>         <h1>All Reviews</h1>
>         <ErrorList error={error} />
>         <ul>
>           {reviews.map((review) => <li>{review.text}</li>)}
>         </ul>
>        </div>
>     );
>   }
> }
> export default withErrorHandling({ name: "AllReviews" })(AllReviewsBase);
> ```
>
> This would also avoid putting that extra `<div />` in the DOM too.

We could leave `withErrorHandling` as is by just having it wrap the other two.

```jsx
// ErrorList is the getError() bit from ErrorHandler
// withErrorHandler is the connect() bit from current withErrorHandling

export function withErrorHandling(...args) {
  return (WrappedComponent) => {
    const Component = ({ error, ...props }) => (
      error
        ? <div><ErrorList error={error} /><WrappedComponent {...props} /></div>
        : <WrappedComponent {...props} />)
    return withErrorHandler(...args)(Component);
  };
}

**Comment by kumar303** [comment]
Yep, it's totally your idea :) I thought it was a good idea at the time but I was worried about the complexity and didn't think that we'd need it. I still don't think we really need it but the Overlay is a unique case because its parent div needs to be absolutely positioned meaning that no outer divs will ever be visible. 

I'll see if I can make these into separate components but right now the HOC is, I think, a nice abstraction around the Redux state. It makes it so the decorated component doesn't have to know anything about Redux state at all.

I also added a bunch of docs to the error handler code so that it's hopefully a bit easier to use.

**Comment by mstriemer** [comment]
I'm not suggesting we get rid of the HOC. I'm just suggesting that we give meaningful names to its different parts and export those as well. This would remove the need for `autoRenderErrors` since you could just grab the part that hooks up to redux and be done.

**Comment by kumar303** [comment]
@mstriemer splitting it into two decorators seems a lot cleaner, thanks. Could you do another review? Do you still think I should render the errors using a standalone component?

File: src/core/errorHandler.js [hunk 18]
```diff
@@ -11,61 +11,75 @@ function generateHandlerId({ name = '' } = {}) {
...
+ *
+ * class SomeComponent extends React.Component {
+ *   static propTypes = {
+ *     errorHandler: PropTypes.object.isRequired,
+ *   }
+ *   render() {
+ *     const { errorHandler } = this.props;
+ *     return (
+ *       <div>
+ *         {errorHandler.hasError() ? errorHandler.renderError() : null}
```
mstriemer: That looks good to me 👍

src/core/errorHandler.js
```diff
@@ -74,15 +88,75 @@ export function withErrorHandling({ name, id } = {}) {
...
+
       return (state) => ({
-        WrappedComponent,
-        errorHandlerId: instanceId,
         error: state.errors[instanceId],
+        instanceId,
       });
     };
 
+    const mapDispatchToProps = (dispatch) => ({ dispatch });
```
mstriemer: This should be the default implementation of `mapDispatchToProps`, so passing `undefined` instead might work.

PR #1707, opened 2017-02-04
Log out from server

### LINKED ISSUE #1557: Logging out from addons-server does not log you out from addons-frontend

mstriemer:
Logging in and out now works across both addons-server and addons-frontend. The local development setup is unchanged, based on whether the app is running on the same domain as the API.

login-logout-same-domain mov

Fixes #1557.

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -184,3 +184,13 @@ export function categories({ api }) {
     state: api,
   });
 }
+
+export function logOutFromServer({ api }) {
+  return callApi({
+    auth: true,
+    credentials: true,
+    endpoint: 'accounts/session',
+    method: 'delete',
+    state: api,
+  });
+}
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -1,7 +1,5 @@
 /* global window */
-import config from 'config';
 import React, { PropTypes } from 'react';
-import cookie from 'react-cookie';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
@@ -56,16 +54,8 @@ export const mapStateToProps = (state) => ({
 
 export const mapDispatchToProps = (dispatch) => ({
   handleLogOut({ api }) {
-    const { location } = window;
-    if (config.get('apiHost') === `${location.protocol}//${location.host}`) {
-      // Deployed config, server manages cookie.
-      return logOutFromServer({ api })
-        .then(() => dispatch(logOutUser()));
-    }
-    // Local development, we manage our own cookie.
-    cookie.remove(config.get('cookieName'), { path: '/' });
-    dispatch(logOutUser());
-    return Promise.resolve();
+    return logOutFromServer({ api })
+      .then(() => dispatch(logOutUser()));
   },
 });
 
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import { logOutUser } from 'core/actions';
-import { startLoginUrl } from 'core/api';
+import { logOutFromServer, startLoginUrl } from 'core/api';
 import translate from 'core/i18n/translate';
 import Button from 'ui/components/Button';
 import Icon from 'ui/components/Icon';
@@ -14,6 +14,7 @@ import Icon from 'ui/components/Icon';
 
 export class AuthenticateButtonBase extends React.Component {
   static propTypes = {
+    api: PropTypes.object.isRequired,
     className: PropTypes.string,
     handleLogIn: PropTypes.func.isRequired,
     handleLogOut: PropTypes.func.isRequired,
@@ -23,9 +24,9 @@ export class AuthenticateButtonBase extends React.Component {
   }
 
   onClick = () => {
-    const { handleLogIn, handleLogOut, isAuthenticated, location } = this.props;
+    const { api, handleLogIn, handleLogOut, isAuthenticated, location } = this.props;
     if (isAuthenticated) {
-      handleLogOut();
+      handleLogOut({ api });
     } else {
       handleLogIn(location);
     }
@@ -45,6 +46,7 @@ export class AuthenticateButtonBase extends React.Component {
 }
 
 export const mapStateToProps = (state) => ({
+  api: state.api,
   isAuthenticated: !!state.api.token,
   handleLogIn(location, { _window = window } = {}) {
     // eslint-disable-next-line no-param-reassign
@@ -53,9 +55,17 @@ export const mapStateToProps = (state) => ({
 });
 
 export const mapDispatchToProps = (dispatch) => ({
-  handleLogOut() {
+  handleLogOut({ api }) {
+    const { location } = window;
+    if (config.get('apiHost') === `${location.protocol}//${location.host}`) {
+      // Deployed config, server manages cookie.
+      return logOutFromServer({ api })
+        .then(() => dispatch(logOutUser()));
+    }
+    // Local development, we manage our own cookie.
     cookie.remove(config.get('cookieName'), { path: '/' });
     dispatch(logOutUser());
+    return Promise.resolve();
   },
 });
 
--- a/tests/client/amo/helpers.js
+++ b/tests/client/amo/helpers.js
@@ -1,3 +1,5 @@
+import { signedInApiState as coreSignedInApiState } from '../helpers';
+
 export const fakeAddon = {
   id: 1234,
   guid: '1234@my-addons.firefox',
@@ -50,8 +52,7 @@ export const fakeReview = {
 /*
  * Redux store state for when a user has signed in.
  */
-export const signedInApiState = {
-  lang: 'en-US',
+export const signedInApiState = Object.freeze({
+  ...coreSignedInApiState,
   clientApp: 'firefox',
-  token: 'secret-token',
-};
+});
--- a/tests/client/core/api/test_api.js
+++ b/tests/client/core/api/test_api.js
@@ -4,7 +4,7 @@ import config from 'config';
 import * as api from 'core/api';
 import { ADDON_TYPE_THEME } from 'core/constants';
 import { ErrorHandler } from 'core/errorHandler';
-import { unexpectedSuccess } from 'tests/client/helpers';
+import { signedInApiState, unexpectedSuccess } from 'tests/client/helpers';
 
 
 export function generateHeaders(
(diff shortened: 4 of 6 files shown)


mstriemer:
To test this locally:

This depends on https://github.com/mozilla/addons-server/pull/4590 so make sure that's included in your addons-server.

1. Start your addons-server
    * You may need to add '127.0.0.1' to `ALLOWED_HOSTS` to get the proxy to work
2. Start a proxy server
    * I wrote a [node script](https://gist.github.com/7e55934a49b085106d3e47e67b4e385d) to do this, start it with `sudo node proxy.js` (sudo to use port 80/override docker)
    * You'll need to install `http-proxy` to use this (should be the only dep)
3. Start the frontend `STATIC_HOST=http://olympia.dev API_HOST=http://olympia.dev npm run dev:amo`
4. Go to http://olympia.dev in responsive mode with an android device set and you should see the mobile pages, desktop will get you desktop
    * If it isn't working verify the proxy server is spamming "got request", if it isn't try hitting the URL in different combinations of curl, private browsing, chrome, chrome incognito, closing browsers, etc until some sort of cache gets busted somewhere (this may just be an issue with my machine, unsure)

src/core/components/AuthenticateButton/index.js
```diff
@@ -53,9 +55,17 @@ export const mapStateToProps = (state) => ({
 });
 
 export const mapDispatchToProps = (dispatch) => ({
-  handleLogOut() {
+  handleLogOut({ api }) {
+    const { location } = window;
+    if (config.get('apiHost') === `${location.protocol}//${location.host}`) {
+      // Deployed config, server manages cookie.
+      return logOutFromServer({ api })
+        .then(() => dispatch(logOutUser()));
```
kumar303: dang, we really need to see global errors. Looks like a failure here will just die in the console. https://github.com/mozilla/addons-frontend/issues/1484

src/core/api/index.js
```diff
@@ -170,3 +170,13 @@ export function categories({ api }) {
...
 }
+
+export function logOutFromServer({ api }) {
+  return callApi({
+    auth: true,
+    credentials: true,
+    endpoint: 'accounts/session',
+    method: 'delete',
+    state: api,
+  });
```
kumar303: does it make any sense to check the result to make sure `ok` is true? Or maybe a better question: will there be any `200` responses to this request that indicate a failure?

tests/client/core/components/TestAuthenticateButton.js
```diff
@@ -60,18 +60,35 @@ describe('<AuthenticateButton />', () => {
...
+    handleLogOut({ api: {} });
     assert.notOk(store.getState().api.token);
     assert.ok(cookie.remove.calledWith('authcookie', { path: '/' }));
   });
 
+  it('asks the server to clear the cookie and JWT in handleLogOut when on the API host', () => {
+    sinon.stub(api, 'logOutFromServer').returns(Promise.resolve());
+    const _config = { cookieName: 'authcookie', apiHost: 'http://localhost:9876' };
+    sinon.stub(config, 'get', (key) => _config[key]);
+    const apiConfig = { token: 'some.jwt.string' };
```
kumar303: this should be [helpers.signedInApiState](https://github.com/mozilla/addons-frontend/blob/master/tests/client/amo/helpers.js#L53)

src/core/components/AuthenticateButton/index.js
mstriemer: Maybe the error can be flagged as "global" and we can put a display only component in `App` that renders global errors. I guess clearing these could be tricky but we could clear all global errors on navigation or put a close button on the error.

```jsx
// AuthenticateButton
return logOutFromServer({ api })
  .then(() => success())
  .catch((error) => dispatch(addError({
    error, name: 'AuthenticateButton', global: true,
   })));

// App
<div className="content">
  <ErrorHandler global />
  {/* ... */}
</div>

File: src/core/api/index.js (hunk 2 above)
Comment (mstriemer) [inline src/core/api/index.js]: If it's a `200` I think it should be okay. I just put `{ "ok": true }` in the response because it seemed weird to return nothing.

File: tests/client/core/components/TestAuthenticateButton.js (hunk 3 above)
Comment (mstriemer) [inline tests/client/core/components/TestAuthenticateButton.js]: Cool. I might make that a function while I'm at it so we don't have a test accidentally modify it.

File: src/core/components/AuthenticateButton/index.js [hunk 4]
```diff
@@ -53,9 +55,17 @@ export const mapStateToProps = (state) => ({
 });
 
 export const mapDispatchToProps = (dispatch) => ({
-  handleLogOut() {
+  handleLogOut({ api }) {
+    const { location } = window;
+    if (config.get('apiHost') === `${location.protocol}//${location.host}`) {
```
mstriemer: I'm probably going to branch off of this and start that. The API is still in review so there's no rush on merging this but it feels like it's probably better off as a separate patch to me.

src/core/components/AuthenticateButton/index.js
kumar303: Yeah, I tried that in https://github.com/mozilla/addons-frontend/pull/1496 but couldn't think of a way to clear the error. I like the idea of adding a 'close' button so that the user can clear it. I'm not sure yet if we have a stable hook for when a new page gets rendered but maybe we do.

mstriemer:
The proxy server stuff is in https://github.com/mozilla/addons-frontend/pull/1766. Does this look okay without the proxy? I wouldn't mind landing them separately.

src/core/components/AuthenticateButton/index.js
kumar303: I think you should remove this before merging. It will require the proxy to work but that's ok. That will come in the next patch.

tests/client/amo/helpers.js
```diff
@@ -50,8 +50,9 @@ export const fakeReview = {
 /*
  * Redux store state for when a user has signed in.
  */
-export const signedInApiState = {
+export const signedInApiState = (overrides = {}) => ({
```
kumar303: What happened here? Did some test accidentally mutate this object? If that was the only problem, you could probably fix it like this:
````js
export const signedInApiState = Object.freeze({
  lang: 'en-US',
  clientApp: 'firefox',
  token: 'secret-token',
});
````

File: tests/client/core/api/test_api.js [hunk 6]
```diff
@@ -467,4 +467,21 @@ describe('api', () => {
...
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/accounts/session/?lang=en-US`, {
+          credentials: 'include',
+          headers: { authorization: 'Bearer my-token' },
+          method: 'DELETE',
+        })
+        .once()
+        .returns(mockResponse);
+      return api.logOutFromServer({
+        api: { lang: 'en-US', token: 'my-token' },
```
kumar303: this can use `signedInApiState` too, no?

tests/client/amo/helpers.js
mstriemer: Nothing mutated it but it didn't very safe to be sharing that object across tests in multiple files. `Object.freeze` should work too. I'll do that.

tests/client/core/api/test_api.js
mstriemer: I added a `signedInApiState` to `tests/client/helpers` and used that. The one in `tests/client/amo/helpers` now imports from `../helpers` and adds `clientApp`. They're both wrapped in `Object.freeze`.

src/core/components/AuthenticateButton/index.js
mstriemer: Removing this will mean that logging out doesn't work when you aren't using the proxy server. If we always use the proxy then that would be fine, but it could be a bit annoying as it is. Of course you can just use dev tools to clear the cookies in development so that may not be a big deal.

mstriemer:
I've rebased both of the branches with master and they can be merged together to test it.

I'm not sure about the with proxy vs without proxy for local development. I like the idea of there being one code path for both but I'm not sure if the proxy is what we want to use all the time in development. It would probably be fine though.

I can delete the local-specific version of the code, but we might want to consider making the proxy the default then. This would mean running your own addons-server rather than using https://addons-dev.allizom.org for the API is the default too.

src/core/components/AuthenticateButton/index.js
kumar303: I'm concerned with including the if branch for this code because it seems fragile. I think letting the logout feature break for non-proxy developers is fine. We will tell everyone to use the proxy and if we run into problems with it then we could add the if branch back in.

kumar303:
> I'm not sure if the proxy is what we want to use all the time in development. 

Yes, I think we should. It will allow us to develop with same-origin functionality that matches prod -- this is super important.

> I can delete the local-specific version of the code, but we might want to consider making the proxy the default then.

Yes, I think we should make the proxy the default. I am fine with introducing it as a separate command (`npm run dev:amo:proxy`) but that should only be temporary until we replace `npm run dev:amo` with it.

mstriemer:
> Yes, I think we should make the proxy the default. I am fine with introducing it as a separate command (npm run dev:amo:proxy) but that should only be temporary until we replace npm run dev:amo with it.

Should I just flip them now, or should we hold off until we're sure things are working properly? ie `npm run dev:amo` and `npm run dev:amo:proxyless`.

kumar303:
> ie npm run dev:amo and npm run dev:amo:proxyless.

yes! I have a pending review in another tab where I am suggesting that :) I had suggested `dev:amo:no-proxy` for the other but it doesn't matter.

kumar303:
or we could use a config param like `disableProxy=true`

tests/client/core/api/test_api.js
```diff
@@ -467,4 +467,21 @@ describe('api', () => {
...
+        .withArgs(`${apiHost}/api/v3/accounts/session/?lang=en-US`, {
+          credentials: 'include',
+          headers: { authorization: 'Bearer secret-token' },
+          method: 'DELETE',
+        })
+        .once()
+        .returns(mockResponse);
+      return api.logOutFromServer({
+        api: signedInApiState,
+      });
```
kumar303: I think this needs to call `mockWindow.verify()`, right?

mstriemer: Well that's embarrassing. Yes, yes it does.

PR #1731, opened 2017-02-07
fix: fetch hangs on server-side (fix #1726)

### LINKED ISSUE #1726: All fetch requests with responses larger than 16kb fail on the server
Turns out #1401 was reopened but the issue encountered is unrelated to the original bug and is, I believe, related to https://github.com/bitinn/node-fetch/blob/master/LIMITS.md.

593b5742a76449879a344a6114ef3ad272253672 introduced usage of `response.clone()` in our fetch code, which works fine on the client, but **fails on node with any response larger than `16kb`** (eg: most search responses) because of the size of the response. Note that a response with no results will work fine: https://addons-dev.allizom.org/en-US/firefox/search/?q=rjkngkwhnkwjnhkwrjnhkjrwnhkjwnrhkwjnhkwh but any large response fails. This is true for search (https://addons-dev.allizom.org/en-US/android/search/?q=Fire) and non-search ( https://addons-dev.allizom.org/en-US/firefox/extensions/) pages

Note that navigating to these pages on the client works without issue.

-----

Looks like we can't use `response.clone` on the server in this way.

tofumatt:
r?

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -5,6 +5,7 @@ import url from 'url';
 
 import 'isomorphic-fetch';
 import { schema as normalizrSchema, normalize } from 'normalizr';
+import { oneLine } from 'common-tags';
 import config from 'config';
 
 import log from 'core/logger';
@@ -62,27 +63,33 @@ export function callApi({
 
   return fetch(apiURL, options)
     .then((response) => {
-      return response.clone().json()
-        .then(
-          (jsonResponse) => ({ response, jsonResponse }),
-          (error) => {
-            log.warn('Could not parse response as JSON:', error);
-            return response.clone().text().then((textResponse) =>
-              ({ response, jsonResponse: { text: textResponse } })
-            );
-          }
-        );
+      const contentType = response.headers.get('Content-Type').toLowerCase();
+
+      // This is a bit paranoid, but we ensure the API returns a JSON response
+      // (see https://github.com/mozilla/addons-frontend/issues/1701).
+      // If not we'll store the text response in JSON and log an error.
+      // If the JSON parsing fails; we log the error and return an "unknown
+      // error".
+      if (contentType === 'application/json') {
+        return response.json()
+          .then((jsonResponse) => ({ response, jsonResponse }));
+      }
+
+      log.warn(oneLine`Response from API was not JSON (was Content-Type:
+        ${contentType})`, response);
+      return response.text().then((textResponse) => (
+        { jsonResponse: { text: textResponse }, response }
+      ));
     })
     .then(({ response, jsonResponse }) => {
       if (response.ok) {
         return jsonResponse;
       }
 
-      // If response is not ok we'll throw.
+      // If response is not ok we'll throw an error.
       // Note that if callApi is executed by an asyncConnect() handler,
       // then redux-connect will catch this exception and
-      // dispatch a LOAD_FAIL action which puts the error in state.
-
+      // dispatch a LOAD_FAIL action which puts the error in the state.
       const apiError = createApiError({ apiURL, response, jsonResponse });
       if (errorHandler) {
         errorHandler.handle(apiError);
--- a/tests/client/core/api/test_api.js
+++ b/tests/client/core/api/test_api.js
@@ -1,4 +1,4 @@
-/* global window */
+/* global Response, window */
 import config from 'config';
 
 import * as api from 'core/api';
@@ -7,18 +7,26 @@ import { ErrorHandler } from 'core/errorHandler';
 import { unexpectedSuccess } from 'tests/client/helpers';
 
 
+export function generateHeaders(
+  headerData = { 'Content-Type': 'application/json' }
+) {
+  const response = new Response();
+  Object.keys(headerData).forEach((key) => (
+    response.headers.append(key, headerData[key])
+  ));
+  return response.headers;
+}
+
 function createApiResponse({
   ok = true, jsonData = {}, ...responseProps
 } = {}) {
   const response = {
     ok,
+    headers: generateHeaders(),
     json: () => Promise.resolve(jsonData),
     ...responseProps,
   };
-  return Promise.resolve({
-    ...response,
-    clone: () => response,
-  });
+  return Promise.resolve(response);
 }
 
 describe('api', () => {
@@ -72,9 +80,7 @@ describe('api', () => {
       sinon.stub(errorHandler, 'handle');
 
       return api.callApi({ endpoint: 'resource', errorHandler })
-        .then(() => {
-          assert(false, 'unexpected success');
-        }, () => {
+        .then(unexpectedSuccess, () => {
           assert.ok(errorHandler.handle.called);
           const args = errorHandler.handle.firstCall.args;
           assert.deepEqual(args[0].response.data.non_field_errors,
@@ -84,26 +90,34 @@ describe('api', () => {
 
     it('handles error responses with JSON syntax errors', () => {
       mockWindow.expects('fetch').returns(createApiResponse({
-        ok: false,
         json() {
           return Promise.reject(
             new SyntaxError('pretend this was a response with invalid JSON'));
-        },
-        text() {
-          return Promise.resolve('actual error response');
         },
       }));
 
       const errorHandler = newErrorHandler();
       sinon.stub(errorHandler, 'handle');
 
-      return api.callApi({ endpoint: 'resource', errorHandler })
+      return api.callApi({ endpoint: 'resource' })
+        .then(unexpectedSuccess)
+        .catch((err) => {
+          assert.equal(err.message,
+            'pretend this was a response with invalid JSON');
+        });
+    });
+
+    it('handles non-JSON responses', () => {
+      mockWindow.expects('fetch').returns(createApiResponse({
+        headers: generateHeaders({ 'Content-Type': 'text/plain' }),
+        text() {
+          return Promise.resolve('some text response');
+        },
+      }));
+
+      return api.callApi({ endpoint: 'resource' })
         .then(() => {
-          assert(false, 'unexpected success');
-        }, () => {
-          assert.ok(errorHandler.handle.called);
-          const args = errorHandler.handle.firstCall.args;
-          assert.equal(args[0].response.data.text, 'actual error response');
+          mockWindow.verify();
         });
     });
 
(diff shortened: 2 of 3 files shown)


src/core/api/index.js
```diff
@@ -62,27 +63,35 @@ export function callApi({
...
     .then((response) => {
-      return response.clone().json()
-        .then(
-          (jsonResponse) => ({ response, jsonResponse }),
-          (error) => {
+      function makeTextResponseJson(textResponse) {
+        return { jsonResponse: { text: textResponse }, response };
+      }
+
+      if (response.headers.get('Content-Type') === 'application/json') {
```
kumar303: for extreme paranoia, how about `response.headers.get('Content-Type').toLowerCase()` ?

src/core/api/index.js
```diff
@@ -62,27 +63,35 @@ export function callApi({
...
+          .then((jsonResponse) => ({ response, jsonResponse }))
+          .catch((error) => {
             log.warn('Could not parse response as JSON:', error);
-            return response.clone().text().then((textResponse) =>
-              ({ response, jsonResponse: { text: textResponse } })
-            );
-          }
-        );
+
+            return response.text()
```
kumar303: This is going to fail with a TypeError because the stream is drained at this point. Until we process the stream in chunks, we don't have a way to fix it. I say, for now, just let `response.json()` throw an error. Maybe it would be nice to catch the error, log the response headers, and then re-throw it.

tests/client/core/api/test_api.js
```diff
@@ -7,18 +7,26 @@ import { ErrorHandler } from 'core/errorHandler';
 import { unexpectedSuccess } from 'tests/client/helpers';
 
 
+const defaultHeaders = {
+  'Content-Type': 'application/json',
+};
+
+function generateHeaders(headerData = defaultHeaders) {
```
kumar303: I would find this slightly more easy to read if it were all in the same place:
````js
function generateHeaders(
  headerData = { 'Content-Type': 'application/json' }
) {
  ...
}
````

File: src/core/api/index.js (hunk 1 above)
Comment (tofumatt) [inline src/core/api/index.js]: Totally 👍

Comment (mstriemer) [inline src/core/api/index.js]: The [docs](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get) suggest this would be an array. Although it does include a footnote we'll probably want to handle (unless universal-fetch does this for us but I would suspect not):

> [1] Prior to Firefox 52, get() only returned the first value in the specified header, with getAll() returning all values. From 52 onwards, get() now returns all values and getAll() has been deleted.

I'm also a bit curious how this would handle the server returning `content-type`, hopefully that string is case-insensitive.

File: src/core/api/index.js (hunk 2 above)
Comment (mstriemer) [inline src/core/api/index.js]: We could just return "Unknown error" here since something went horribly wrong.

File: src/core/api/index.js (hunk 1 above)
Comment (kumar303) [inline src/core/api/index.js]: ugh, maybe it needs to be `response.headers.get('Content-Type').toString().toLowerCase()`

File: tests/client/core/api/test_api.js [hunk 4]
```diff
@@ -7,18 +7,26 @@ import { ErrorHandler } from 'core/errorHandler';
...
+  return {
+    get: (name) => headerData[name],
```
mstriemer: You could use a real `Response` object and return its `headers`. In fact we should probably use a real `Response` object in `createApiResponse` but one step at a time is probably fine.

Note: This won't identify the Firefox issue since we're running against release which is 51 but after the next update I think these tests will start to fail if we use a real `Response` (which is good).

mstriemer: I tried it out and the tests pass, going to see if they fail on 52+.

```js
function generateHeaders(headerData = defaultHeaders) {
  const response = new Response();
  Object.keys(headerData).forEach((key) => response.headers.append(key, headerData[key]));
  return response.headers;
}

File: src/core/api/index.js (hunk 2 above)
Comment (tofumatt) [inline src/core/api/index.js]: Arg, fair point. Yeah the log will output that the JSON from the server was received but malformed, and then we can output "Unknown error, bad JSON" or whatnot 👍

File: tests/client/core/api/test_api.js (hunk 4 above)
Comment (mstriemer) [inline tests/client/core/api/test_api.js]: Actually, it appears as though the `'Content-type'` header is special? In Chrome and Firefox it always returns and with multiple values it returns them comma-separated. I guess this means the MDN docs are incorrect.

Comment (tofumatt) [inline tests/client/core/api/test_api.js]: Weird. I noticed usage of `Content-type` elsewhere in tests but figured it was just an alias, like you said.

Hmmm, so what should I do here, just check for `Content-Type` as before? Does `Content-type` work in browsers other than Firefox and Chrome?

File: src/core/api/index.js (hunk 1 above)
Comment (mstriemer) [inline src/core/api/index.js]: It appears as though this will actually return a string and the "content-type" header is special.

The string does appear to be case-insensitive. Setting the header with `'content-type'` and retrieving it with `'Content-Type'` worked for me.

File: tests/client/core/api/test_api.js (hunk 4 above)
Comment (mstriemer) [inline tests/client/core/api/test_api.js]: I went to try this in Safari since that's the only other browser I have installed right now neither `fetch` nor `Response` is supported. So, maybe for all other browsers it works as expected because it isn't supported? 😜

File: src/core/api/index.js (hunk 1 above)
Comment (tofumatt) [inline src/core/api/index.js]: Cheers, thanks for checking it out 👍

File: tests/client/core/api/test_api.js (hunk 4 above)
Comment (tofumatt) [inline tests/client/core/api/test_api.js]: Our code just polyfills it though... right? I'd hate to write code, even test code, that fails to run in Safari (rendering all our tests useless). 😢

Comment (mstriemer) [inline tests/client/core/api/test_api.js]: Our tests do not all pass in Safari (and it doesn't really work well with karma for some reason) but it looks like this updated test using `Response` passes.

Comment (tofumatt) [inline tests/client/core/api/test_api.js]: Ah, fair enough. Okay then, I'll stick to using `Response`. 👍

File: src/core/api/index.js (hunk 1 above)
Comment (mstriemer) [inline src/core/api/index.js]: So the "conten-type" header isn't special but MDN was out of date. I checked the spec and Firefox and Chrome matched the spec. I updated MDN.

**Comment by tofumatt** [comment]
All set; added some comments at @muffinresearch's suggestion and cleaned up the issues mentioned. Good for another r?

**Comment by tofumatt** [comment]
Sorry, I broke the server tests but fixed them. Didn't want to copy the helper so I hacked in the response headers... but if you want I can likely abstract that helper so server and client tests can use it, I just wanted to get this out before I went offline for a few hours to squash/dinner.

File: src/core/api/index.js [hunk 5]
```diff
@@ -62,27 +63,44 @@ export function callApi({
...
+          .catch((error) => {
             log.warn('Could not parse response as JSON:', error);
-            return response.clone().text().then((textResponse) =>
-              ({ response, jsonResponse: { text: textResponse } })
-            );
-          }
-        );
+
+            return makeTextResponseJson(
+              `Unknown Error: Could not parse response from ${apiURL} as JSON`);
```
kumar303: I don't think this is the right thing to do because it relies on the response status to produce an error. If the response body triggers a JSON parse error but its status is 200 then `callApi()` will happily return it as if everything is rosy. I think you should just let this throw an error. If it becomes a problem in real life at least we'll see an error in the console.

tests/client/core/api/test_api.js
```diff
@@ -1,24 +1,33 @@
...
 import * as api from 'core/api';
 import { ADDON_TYPE_THEME } from 'core/constants';
 import { ErrorHandler } from 'core/errorHandler';
 import { unexpectedSuccess } from 'tests/client/helpers';
 
 
+export function generateHeaders(
+  headerData = { 'Content-Type': 'application/json' }
+) {
+  const response = new Response();
```
kumar303: yeah, this is way more future proof 🎉

tests/client/core/api/test_api.js
```diff
@@ -1,24 +1,33 @@
...
+  Object.keys(headerData).forEach((key) => (
+    response.headers.append(key, headerData[key])
+  ));
+  return response.headers;
+}
+
 function createApiResponse({
   ok = true, jsonData = {}, ...responseProps
 } = {}) {
   const response = {
```
kumar303: would it be crazy to use a real response instance here? Let's not hold up this patch if it's too much work.

tests/client/core/api/test_api.js
```diff
@@ -89,23 +96,51 @@ describe('api', () => {
...
+      });
+
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/resource/?lang=`, {
+          method: 'GET', headers: {},
+        })
+        .once()
+        .returns(response);
+      return api.callApi({ endpoint: 'resource', method: 'GET' })
+        .then(() => (mockWindow.verify()));
```
kumar303: does this need parentheses?

tests/client/core/api/test_api.js
tofumatt: I think we can do it but like @mstriemer baby steps on using real `Response` 😄

tofumatt:
Okay, addressed the comments.

src/core/api/index.js
```diff
@@ -62,27 +63,38 @@ export function callApi({
...
+      // This is a bit paranoid, but we ensure the API returns a JSON response
+      // (see https://github.com/mozilla/addons-frontend/issues/1701).
+      // If not we'll store the text response in JSON and log an error.
+      // If the JSON parsing fails; we log the error and return an "unknown
+      // error".
+      if (contentType === 'application/json') {
+        return response.json()
+          .then((jsonResponse) => {
+            return { response, jsonResponse };
+          });
```
kumar303: this can be a little more concise now:
````js
.then((jsonResponse) => ({ response, jsonResponse })
````

File: src/core/api/index.js [hunk 10]
```diff
@@ -62,27 +63,38 @@ export function callApi({
...
+      }
+
+      log.warn(oneLine`Response from API was not JSON (was Content-Type:
+        ${contentType})`, response);
+      return response.text()
+        .then(makeTextResponseJson);
```
kumar303: this can go back to a one-liner:
````js
.then((textResponse) => ({ jsonResponse: { text: textResponse }, response }));
````

tests/client/core/api/test_api.js
kumar303: I think it would be wise to use a real response. Can you file an issue so we don't forget?

src/core/api/index.js
tofumatt: I did this on purpose because @muffinresearch and I were both chatting and found it hard to read with all the nesting and implicit returns.

tofumatt: I guess it's fine for now though, I'll move it back to a one-liner, especially as this is now cleaner.

kumar303: I find it easier to read on a single line since the object is not complex

Issue #1739, opened 2017-02-07
Review titles are awkward to write


kumar303:
As of https://github.com/mozilla/addons-frontend/pull/1706 , writing a review for an add-on requires you to enter a title. The titles make the list of reviews look nice but it's not so nice as a user to think of a title. 

For me, I often get tripped up by this because my first inclination is to say something simple, e.g. "I like the way this add-on helps me do X." Thinking of a title slows me down.

Maybe the title could be auto-populated with the first sentence of the review if the user wants to leave it blank? Maybe we could make titles optional? Maybe we could put the title input underneath the body input? Just thinking out loud.

pwalm:
Or... no title? I'm very happy to have no title. It removes one more barrier of "[blank slate syndrome](http://www.urbandictionary.com/define.php?term=Blank%20Page%20Syndrome)"

tofumatt:
We'd probably need to update the server but I'd be happy with no title or an optional title at the very least. 👍

muffinresearch:
We like the idea of removing the title. We can always add it back if there's more rationale to do so later.

kumar303:
I'll file an issue to consider the review list layout now that none will have titles. It might need tweaking.

vcarciu:
Verified as fixed : 
![notitlerew](https://cloud.githubusercontent.com/assets/1583842/22822951/442828ea-ef8a-11e6-8f63-df6b40b6ae24.png)

PR #1765, opened 2017-02-08
Remove review title from entry form

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1739

### DIFF
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -35,9 +35,7 @@ export class AddonReviewBase extends React.Component {
   componentWillReceiveProps(nextProps) {
     const { review } = nextProps;
     if (review) {
-      this.setState({
-        reviewBody: review.body,
-      });
+      this.setState({ reviewBody: review.body });
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/amo/components/AddonReview.js
```diff
@@ -22,33 +22,29 @@ export class AddonReviewBase extends React.Component {
...
     }
     this.overlayCard = null;
     this.reviewTextarea = null;
-    this.reviewTitleInput = null;
   }
 
   componentWillReceiveProps(nextProps) {
     const { review } = nextProps;
     if (review) {
       this.setState({
```
tofumatt: I bet this could be one line now, if you wanted.

src/amo/components/AddonReview.js
```diff
@@ -104,15 +94,6 @@ export class AddonReviewBase extends React.Component {
         <form onSubmit={this.onSubmit} ref={(ref) => { this.reviewForm = ref; }}>
           <div className="AddonReview-form-input">
             {errorHandler.hasError() ? errorHandler.renderError() : null}
-            <input
```
tofumatt: Did this never have a label? Looks like the `<textarea>` is missing one too. I know it's unrelated to this patch, but could you add one?

kumar303: The textarea doesn't have a label in the mock. Are you suggesting we should add a hidden one? I'm not too familiar with a11y -- what purpose does that serve?

tofumatt: Yeah, a hidden one will give extra context to screenreaders. Something simple like "Your review text" should be fine.

kumar303: ok, I added a hidden label

kumar303:
Thanks for the review

PR #1766, opened 2017-02-08
Proxy server for AMO development

### LINKED ISSUE #1716: Start a proxy server with the dev server
In our deployed version the frontend and the server are on the same domain but locally they are on different domains.

### What happened?
We had to add code to handle both cases to the logout handler.

One code path for all environments.

We should start a proxy server along with the dev server. This can also detect mobile and proxy to mobile/desktop (possibly optionally for convenience).

There is a working proxy (with some config changes) in a [gist](https://gist.github.com/mstriemer/7e55934a49b085106d3e47e67b4e385d).

Related to #1707.

mstriemer:
Start the proxy server with `npm run dev:amo`. You need to be running an API server at `http://olympia.dev`. Access the site as usual at http://localhost:3000.

Fixes #1716.

Depends on https://github.com/mozilla/addons-server/pull/4681.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,8 @@ To see this report while running tests locally, type:
 
 A proxy server is provided for running the AMO app with the API on the same host as the frontend.
 This provides a setup that is closer to production than running the frontend on its own. The
-default configuration for this is to use a local addons-server for the API. This comes
-pre-configured to use a local addons-server which can be setup according to the
+default configuration for this is to use a local addons-server for the API which can be setup
+according to the
 [addons-server docs](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
 Docker is the preferred method of running addons-server.
 
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ Generic scripts that don't need env vars. Use these for development:
 | Script                  | Description                                           |
 |-------------------------|-------------------------------------------------------|
 | npm run dev:admin       |  Starts the dev server (admin app)                    |
-| npm run dev:amo         |  Starts the dev server (amo)                          |
-| npm run dev:amo:proxy   |  Starts the dev server with proxy (amo)               |
+| npm run dev:amo         |  Starts the dev server and proxy (amo)                |
+| npm run dev:amo:no-proxy|  Starts the dev server without proxy (amo)            |
 | npm run dev:disco       |  Starts the dev server (discovery pane)               |
 | npm run eslint          |  Lints the JS                                         |
 | npm run stylelint       |  Lints the SCSS                                       |
@@ -95,17 +95,19 @@ To see this report while running tests locally, type:
 
     open ./coverage/index.html
 
-### Running development AMO similar to production
+### Running AMO for local development
 
 A proxy server is provided for running the AMO app with the API on the same host as the frontend.
-This provides a setup that is closer to production than running the frontend on its own. The only
-supported method of doing this is to also be running an instance of addons-server at
-http://olympia.dev (this is its default configuration in development mode). You can find
-instructions on settings up addons-server in the
+This provides a setup that is closer to production than running the frontend on its own. The
+default configuration for this is to use a local addons-server for the API. This comes
+pre-configured to use a local addons-server which can be setup according to the
 [addons-server docs](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
 Docker is the preferred method of running addons-server.
 
-To run the development server with the proxy use `npm run dev:amo:proxy`.
+This should work without any issues with a local addons-server and can also be configured to use
+`https://addons-dev.allizom.org` as the API but the addons-server views will not all work as
+expected. If you would like to use `https://addons-dev.allizom.org` for the API you should use the
+`npm run dev:amo:no-proxy` command to start the server without the proxy.j
 
 ### Configuring for local development
 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Generic scripts that don't need env vars. Use these for development:
 |-------------------------|-------------------------------------------------------|
 | npm run dev:admin       |  Starts the dev server (admin app)                    |
 | npm run dev:amo         |  Starts the dev server (amo)                          |
+| npm run dev:amo:proxy   |  Starts the dev server with proxy (amo)               |
 | npm run dev:disco       |  Starts the dev server (discovery pane)               |
 | npm run eslint          |  Lints the JS                                         |
 | npm run stylelint       |  Lints the SCSS                                       |
@@ -93,6 +94,18 @@ The continuous integration process will give you a link to view the report.
 To see this report while running tests locally, type:
 
     open ./coverage/index.html
+
+### Running development AMO similar to production
+
+A proxy server is provided for running the AMO app with the API on the same host as the frontend.
+This provides a setup that is closer to production than running the frontend on its own. The only
+supported method of doing this is to also be running an instance of addons-server at
+http://olympia.dev (this is its default configuration in development mode). You can find
+instructions on settings up addons-server in the
+[addons-server docs](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
+Docker is the preferred method of running addons-server.
+
+To run the development server with the proxy use `npm run dev:amo:proxy`.
 
 ### Configuring for local development
 
--- a/README.md
+++ b/README.md
@@ -104,9 +104,12 @@ according to the
 [addons-server docs](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
 Docker is the preferred method of running addons-server.
 
-This should work without any issues with a local addons-server and can also be configured to use
-`https://addons-dev.allizom.org` as the API but the addons-server views will not all work as
-expected. If you would like to use `https://addons-dev.allizom.org` for the API you should use the
+Authentication will work when initiated from addons-frontend and will persist to addons-server but
+it will not work when logging in from an addons-server page. See
+[mozilla/addons-server#4684](https://github.com/mozilla/addons-server/issues/4684) for more
+information on fixing this.
+
+If you would like to use `https://addons-dev.allizom.org` for the API you should use the
 `npm run dev:amo:no-proxy` command with an `API_HOST` to start the server without the proxy. For
 example: `API_HOST=https://addons-dev.allizom.org npm run dev:amo:no-proxy`.
 
--- a/README.md
+++ b/README.md
@@ -107,7 +107,8 @@ Docker is the preferred method of running addons-server.
 This should work without any issues with a local addons-server and can also be configured to use
 `https://addons-dev.allizom.org` as the API but the addons-server views will not all work as
 expected. If you would like to use `https://addons-dev.allizom.org` for the API you should use the
-`npm run dev:amo:no-proxy` command to start the server without the proxy.j
+`npm run dev:amo:no-proxy` command with an `API_HOST` to start the server without the proxy. For
+example: `API_HOST=https://addons-dev.allizom.org npm run dev:amo:no-proxy`.
 
 ### Configuring for local development
 
--- a/bin/proxy.js
+++ b/bin/proxy.js
@@ -1,6 +1,7 @@
 require('babel-register');
 
 const http = require('http');
+const url = require('url');
 
 const bunyan = require('bunyan');
 const config = require('config');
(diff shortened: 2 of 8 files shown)


bin/proxy.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
+require('babel-register');
+
+const http = require('http');
```
kumar303: can you use imports? That would be nice because it will activate import linting.

bin/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+
+const proxy = httpProxy.createProxyServer({ secure: false });
+const apiHost = config.get('proxyApiHost', null) || config.get('apiHost');
+const frontendHost = `http://${config.get('serverHost')}:${config.get('serverPort')}`;
+
+console.log(`apiHost: ${apiHost}`);
+console.log(`frontendHost: ${frontendHost}`);
+
+function getHost(req) {
+  if (!req.headers['user-agent'].includes('Android') || req.url.startsWith('/api/')) {
```
kumar303: Can we remove the user agent check? If we need to develop on the old desktop site, I think it's understandable to not use the proxy for that. You could just check for more URL prefixes like `developers`, `editors`, etc to support what we need from desktop during local development, right?

bin/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+  console.log(`${req.method} ${req.url}`);
+  return proxy.web(req, res, { target: getHost(req), changeOrigin: true, autoRewrite: true });
+});
+
+proxy.on('proxyRes', (proxyRes, req) => {
+  console.log(` ${proxyRes.statusCode} ~> ${proxyRes.client._host}${req.url}`);
+});
+
+proxy.on('error', (error, req) => {
+  console.log(` ERR ~> ${req.url}\n${error}`);
```
kumar303: console.error ?

package.json
```diff
@@ -10,10 +10,12 @@
     "clean": "rimraf './dist/*!(.gitkeep)' './webpack-assets.json'",
     "dev:admin": "better-npm-run dev:admin",
     "dev:amo": "better-npm-run dev:amo",
+    "dev:amo:proxy": "better-npm-run dev:amo:proxy",
```
kumar303: The last line when this starts up is:
````
INFO: server/35663 on indigo.local: 👁  Open your browser at http://127.0.0.1:3333 to view it. (app=amo)
````
Is there an easy way to output another line mentioning the proxy? Maybe not.

File: bin/proxy.js [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+
+const config = require('config');
+const httpProxy = require('http-proxy');
+
+
+const proxy = httpProxy.createProxyServer({ secure: false });
+const apiHost = config.get('proxyApiHost', null) || config.get('apiHost');
+const frontendHost = `http://${config.get('serverHost')}:${config.get('serverPort')}`;
+
+console.log(`apiHost: ${apiHost}`);
```
kumar303: I think prefixing this would help, like: 
````
console.log(`proxy: apiHost: ${apiHost}`)
````

File: bin/proxy.js [hunk 6]
```diff
@@ -0,0 +1,39 @@
...
+console.log(`frontendHost: ${frontendHost}`);
```
kumar303: same here:
````
console.log(`proxy: frontendHost: ${frontendHost}`)
````

File: bin/proxy.js [hunk 7]
```diff
@@ -0,0 +1,39 @@
...
+proxy.on('proxyRes', (proxyRes, req) => {
+  console.log(` ${proxyRes.statusCode} ~> ${proxyRes.client._host}${req.url}`);
+});
+
+proxy.on('error', (error, req) => {
+  console.log(` ERR ~> ${req.url}\n${error}`);
+});
+
+const port = parseInt(config.get('proxyPort', '3333'), 10);
+console.log(`🚦 Proxy running on port ${port}`);
```
kumar303: I think it would be helpful if this were a complete URL one could copy/paste into their browser

bin/proxy.js
kumar303: Hmm, I guess relying on user agent is ok for now. We'll definitely have to solve it when we start on the desktop site though.

My main concern was developing in Chrome (which is needed for source map support). I just found their responsive mode. It's a little clunky but I guess it will do: https://developers.google.com/web/tools/chrome-devtools/device-mode/

bin/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+
+function getHost(req) {
+  if (!req.headers['user-agent'].includes('Android') || req.url.startsWith('/api/')) {
+    return apiHost;
+  }
+  return frontendHost;
+}
+
+const server = http.createServer((req, res) => {
+  console.log(`${req.method} ${req.url}`);
```
kumar303: Oh, I forgot to mention. This creates a lot of noise in the logs since we already see frontend server logging. Could this log message only go out for 4xx or 5xx responses?

bin/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+  return proxy.web(req, res, { target: getHost(req), changeOrigin: true, autoRewrite: true });
+});
+
+proxy.on('proxyRes', (proxyRes, req) => {
+  console.log(` ${proxyRes.statusCode} ~> ${proxyRes.client._host}${req.url}`);
```
kumar303: Similarly, I don't think this is so helpful either. Maybe the logging could only happen for 4xx and 5xx responses?

bin/proxy.js
mstriemer: Do you mean `import http from 'http'`? I don't think that works until you're in a file that was imported. I guess I could make a file that just does `require('babel-register');require('./proxy.js')` and call that instead though.

kumar303: I tried it and it doesn't work. That's surprising to me though. I thought `require('babel-register')` was magic enough to make it work.

mstriemer: I think that preps everything for the next import. So any file that this file `require`s can use `import`, but this file still can't.

mstriemer:
Each request now only generates one log statement, it will look like this:

```
[2] [2017-02-11T03:12:45.044Z]  INFO: proxy/10335 on MarkBook Pro: 200 ~> http://127.0.0.1:3333/en-US/android/ (app=amo)
```

When the app starts with the proxy server you will see:

```
[1] [2017-02-11T03:30:29.485Z]  INFO: server/10762 on MarkBook Pro: 🔥  Addons-frontend server is running [ENV:development] [APP:amo] [isDevelopment:true] [isDeployed:false] [apiHost:http://localhost:3000] [apiPath:/api/v3] (app=amo)
[1] [2017-02-11T03:30:29.485Z]  INFO: server/10762 on MarkBook Pro: 🚦  Proxy detected, frontend running at http://127.0.0.1:3333. (app=amo)
[1] [2017-02-11T03:30:29.485Z]  INFO: server/10762 on MarkBook Pro: 👁  Open your browser at http://localhost:3000 to view it. (app=amo)
```

I left the file where it is, so some JS features like imports and the spread operator aren't supported in the proxy file. If you want me to move it I can but as it is right now it's just missing imports which isn't a big deal to me.

README.md
```diff
@@ -94,6 +95,18 @@ To see this report while running tests locally, type:
 
     open ./coverage/index.html
 
+### Running development AMO similar to production
```
kumar303: These docs look good, thanks, but when the commands get merged this will need some re-organizing. The no-proxy section should explicitly mention how to connect to `addons-dev.allizom.org`.

bin/proxy.js
```diff
@@ -0,0 +1,45 @@
...
+const httpProxy = require('http-proxy');
+
+const log = bunyan.createLogger({
+  name: 'proxy',
+  app: config.get('appName'),
+  serializers: bunyan.stdSerializers,
+});
+
+const proxy = httpProxy.createProxyServer();
+const apiHost = config.get('proxyApiHost', null) || config.get('apiHost');
```
kumar303: Is `https` the problem with using `addons-dev.allizom.org`? If so, could this script die in an informative error if `apiHost` starts with `https://` ?

bin/proxy.js
```diff
@@ -0,0 +1,45 @@
...
+const frontendHost = `http://${config.get('serverHost')}:${config.get('serverPort')}`;
+
+log.info(`apiHost: ${apiHost}`);
+log.info(`frontendHost: ${frontendHost}`);
+
+function getHost(req) {
+  if (!req.headers['user-agent'].includes('Android') || req.url.startsWith('/api/')) {
```
kumar303: what were your thoughts on replacing the user agent check with a whitelist of `/developers`, `/editors`, and `/admin` (or whatever) ? That will prep us for inevitable desktop site development. Or will it be annoying not having a working index page for the desktop site?

bin/proxy.js
mstriemer: So I wasn't entirely sure what the problem was but I looked into it some more.

There was an issue with cookies (these needed to be rewritten by the proxy). I set that up so that the proxy will strip the `Domain` and `Secure` portions of the cookie and you can log in on the frontend and interact with the site (woo). Leaving your first review works, but it doesn't seem to get loaded and you can't update it (this might not be proxy related).

If you try to use the site on a non-mobile browser then it looks like it works but you can't log in since you'll get redirected to `https://addons-dev.allizom.org` by FxA. We could work around this but we'd need to do something in addons-server.

So for now it looks like defaulting to `http://olympia.dev` seems like a good default and using -dev will sort of work if you want. Well enough that we probably don't need to prevent it, anyway.

bin/proxy.js
mstriemer: It would be nice to be able to see both. I guess this should use the `mamo` cookie, then check that list, then go to the frontend.

mstriemer: I updated it to just use the `mamo` cookie. So it will go to mobile unless the cookie is set to `off` or the path starts with `/api/`. This is probably a good enough start until work starts on the desktop frontend and we have a list of the URLs we want to allow setup somewhere.

mstriemer:
Alright, the default is now to run the proxy in development for AMO. To run it without the proxy you need to specify the `API_HOST` unfortunately. I think this is required without updating the app config to be aware of the proxy and since it isn't the recommended way of running the site in development I think this is fine.

* `npm run dev:amo` ~> Use the proxy with `http://olympia.dev` as the API.
* `API_HOST=http://olympia.dev npm run dev:amo:no-proxy` ~> No proxy with `http://olympia.dev` as the API.

mstriemer:
So actually now that I'm testing this some more login doesn't work with a local addons-server at `localhost:3000` either, since it does the same thing that addons-dev does and redirects to olympia.dev. So I guess -dev and olympia.dev are in the same boat here.

I guess I only tested the login-from-amo case on localhost, oops =\

I made some updates to the FxA config on FxA's end and updated the authenticate view to take the `config` param and I got it working. I'm guessing we could make this work by telling your local addons-server to use the `local` config instead of the `default` config. I'm not sure if there's an easy way that we could do this through the docker setup, but it would mean running the frontend addons-server and the regular addons-server in slightly different configurations.

I'm not sure how much we want to mess with this, maybe logging in from the frontend only is fine. I'll still need to land an addons-server patch with these changes to get the login to persist between both versions (and only if you log in from mobile).

mstriemer:
I just submitted mozilla/addons-server#4681 to support using the authenticate flow. This will now hit the API's authenticate view which will redirect back to the frontend when logging in with the proxy server. Since we now expect the redirect to go to `/api/v3/accounts/authenticate/` we'll still hit the login API if you are running without the proxy.

This means when you login through addons-frontend you are also logged in to addons-server. Unfortunately the reverse still doesn't work since addons-server will try to redirect to olympia.dev if you log in from there.

bin/proxy.js
```diff
@@ -0,0 +1,75 @@
...
+  proxyRes.headers['set-cookie'] = proxyCookies.map((rewrittenCookie) => {
+    if (!req.connection.encrypted) {
+      return rewrittenCookie.replace(/;\s*?(Secure)/i, '');
+    }
+    return rewrittenCookie;
+  });
+}
+
+function getHost(req) {
+  const useDesktop = req.headers.cookie && cookie.parse(req.headers.cookie).mamo === 'off';
```
kumar303: Hmm, I guess this means you can't simply type `http://localhost:3000/en-US/developers/` in the same browser that you're developing the mobile site on (i.e. in two different tabs). That's probably fine since toggling the cookie is easy enough.

mstriemer: Yeah, I figured at some point in the future we'd need to come up with a solution for this and I didn't bother spending any time on it here.

src/amo/routes.js
```diff
@@ -26,7 +26,7 @@ export default (
     <Route path=":visibleAddonType/categories/" component={CategoryList} />
     <Route path=":visibleAddonType/featured/" component={FeaturedAddons} />
     <Route path=":visibleAddonType/:slug/" component={CategoryPage} />
-    <Route path="fxa-authenticate" component={HandleLogin} />
+    <Route path="/api/v3/accounts/authenticate/" component={HandleLogin} />
```
mstriemer: I think the production deployment maps `/api/*` to addons-server and won't try the frontend.

kumar303: Cool

bin/proxy.js
kumar303: Works for me

PR #1772, opened 2017-02-09
Add welcome mat

### LINKED ISSUE #1740: Welcome mat is missing
The area below the search bar on the Explore page of AMO Mobile is missing:
screen shot 2017-02-07 at 2 52 04 pm

If we're waiting on copy, I think Scott might already have something along these lines.

muffinresearch:
Fixes #1740 

Before:

screen shot 2017-02-09 at 15 22 31

After:

screen shot 2017-02-09 at 15 21 55

### DIFF
--- a/src/amo/containers/Home.js
+++ b/src/amo/containers/Home.js
@@ -34,6 +34,21 @@ export class HomePageBase extends React.Component {
     const { i18n } = this.props;
     return (
       <div className="HomePage">
+        <div className="HomePage-welcome">
+          <p className="HomePage-welcome-text">
+            {i18n.gettext(dedent`Extensions are special features you can add to Firefox.
+            Themes let you change your browser's appearance.`)}
+          </p>
+          <div className="HomePage-welcome-links">
+            <Link className="HomePage-extensions-link" to="/extensions/">
+              {i18n.gettext('Extensions')}
+            </Link>
+            <Link className="HomePage-themes-link" to="/themes/">
+              {i18n.gettext('Themes')}
+            </Link>
+          </div>
+        </div>
+
         <h2 className="HomePage-subheading">{i18n.gettext('What do you want Firefox to do?')}</h2>
         <ul className="HomePage-category-list">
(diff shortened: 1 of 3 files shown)


src/amo/containers/Home.js
```diff
@@ -34,6 +34,21 @@ export class HomePageBase extends React.Component {
     const { i18n } = this.props;
     return (
       <div className="HomePage">
+        <div className="HomePage-welcome">
+          <p className="HomePage-welcome-text">
+            {i18n.gettext(dedent`Extensions are special features you can add to Firefox.
+            Themes let you change your browser's appearance.`)}
+          </p>
+          <div className="HomePage-welcome-links">
+            <Link className="HomePage-extensions-link" to="/extensions/">
```
tofumatt: This is pedantic, but I've been using `visibleAddonType(ADDON_TYPE_EXTENSION)` instead of hardcoding `extensions`, just to avoid magic constants. I guess URLs shouldn't change but it seems a bit nicer. 🤷‍♀️

src/amo/containers/Home.js
```diff
@@ -34,6 +34,21 @@ export class HomePageBase extends React.Component {
...
+              {i18n.gettext('extensions')}
```
tofumatt: Should this be lowercase? I guess it doesn't matter as the CSS styles it to drop-caps but it seems like it should be capitalised.

muffinresearch: Yep, that makes sense I'll at least upper-case the first letter the rest is presentation.

src/amo/containers/Home.js
muffinresearch: It would be event better to tie back to the routes like django's named urls. Maybe that's something we can look at - it might even be easier with the newer version of the router?

tofumatt: Yeah, agreed. I think react-router used to have a feature like `routeFor` or something way back but they ditched it. I'm not too bothered by it as URLs shouldn't change, just a thought 😄

Issue #1738, opened 2017-02-07
Your new review will replace latest review 


vcarciu:
Steps to reproduce:
1.On Android, go to any add-ons with reviews
2.Write a new review

Expected results:
Number of reviews will increase and all reviews are visible

Actual results:
Number of reviews is the same and we can observe that your review replaced latest review for that add-on

Please see video for this issue : 
![reviewreplace](https://cloud.githubusercontent.com/assets/1583842/22700940/e20dfac6-ed64-11e6-9e89-d0be78d28332.gif)

kumar303:
Hi @vcarciu ! Thanks, I can reproduce this on dev too. If I refresh the pages a couple times I can see both reviews so it's some kind of caching issue but it might be client-side caching, I'm not sure yet.

vcarciu:
Verified as fixed : 
![lastrew](https://cloud.githubusercontent.com/assets/1583842/22823144/071d42a4-ef8b-11e6-9386-da7b252a6512.gif)

PR #1774, opened 2017-02-09
Refresh add-on after posting a review to update review count

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1738

After reading the bug, here's what you need to know about the patch:
* The behavior where the review is missing from the list seemed to be solved by https://github.com/mozilla/addons-frontend/pull/1706 (I could reproduce it before but not after). This patch did not need to address that.
* The review count was not getting updated on the add-on detail page after submitting a new review. This patch addresses that.

### DIFF
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -45,10 +45,10 @@ export class AddonReviewBase extends React.Component {
   }
 
   onSubmit = (event, { overlayCard = this.overlayCard } = {}) => {
+    const { reviewBody, reviewTitle } = this.state;
     event.preventDefault();
     event.stopPropagation();
 
-    const { reviewBody, reviewTitle } = this.state;
     const addonSlug = this.props.review.addonSlug;
     const apiState = this.props.apiState;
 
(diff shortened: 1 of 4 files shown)


src/amo/components/AddonReview.js
```diff
@@ -46,20 +48,24 @@ export class AddonReviewBase extends React.Component {
...
-      apiState: this.props.apiState,
     };
     // TODO: render a progress indicator in the UI.
     // https://github.com/mozilla/addons-frontend/issues/1156
     return this.props.updateReviewText(params)
-      .then(() => {
-        overlayCard.hide();
-      });
+      // This will update the review count on the add-on detail page.
+      .then(() => this.props.refreshAddon({ addonSlug, apiState }))
```
kumar303: This will hold the window open while it makes another API request. Should I close it immediately and let that run in the background? Deferring it to the background means the review count update will be delayed.

tofumatt: I actually think that's okay UI; usually the problem with showing an empty/skeleton page then loading in content after the API request is done is content jumping around the page. In this case it's just the review count number which shouldn't case content to jump. I say close it and run the API request in the background if it's not a PITA.

kumar303: Cool, I changed made the refresh happen after closing the window. This should make it feel more snappy.

Issue #1779, opened 2017-02-09
Updating review text shows the old text when you press submit


kumar303:
### Describe the problem and steps to reproduce it:

* rate an add-on on the detail page
* write a review
* save the review
* select a new star rating to change it
* enter in new review text
* press save again

### What happened?

For a split second, while the review is saving, the form contains the old text.

### What did you expect to happen?

It should only show the new text you just entered otherwise the user might think the site has a glitch.

vcarciu:
Verified as fixed :  
![replacementtext](https://cloud.githubusercontent.com/assets/1583842/22888483/8dfa7b86-f20e-11e6-8e32-770b6542b8a4.gif)

PR #1780, opened 2017-02-09
Don't reset review state when submitting a new review

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1779

Here's an explanation of what was happening:
* Let's say a user creates a rating and a written review
* The user returns to edit the written review
* When the user presses 'submit' on their text change, the component patches the API with the new review object and waits
* Meanwhile, the error handler resets itself since a new API request has been initiated. This triggers a re-render of the AddonReview component
* The AddonReview component renders using the old review object from state and sets the form text to the old text. To the user, it looks like pressing the submit button just erased what they had written.
* The API response finally completes and puts the new review object in state. By the time the overlay for the form has already closed.

The solution to this problem (in this patch) is to optimistically dispatch the new review object before the API request has completed. When the API responds, the resulting object will also be dispatched to state.

### DIFF
--- a/src/amo/components/AddonReview.js
+++ b/src/amo/components/AddonReview.js
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import { submitReview } from 'amo/api';
-import { setReview } from 'amo/actions/reviews';
+import { setDenormalizedReview, setReview } from 'amo/actions/reviews';
 import { refreshAddon } from 'core/utils';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
@@ -19,6 +19,7 @@ export class AddonReviewBase extends React.Component {
     i18n: PropTypes.object.isRequired,
     refreshAddon: PropTypes.func.isRequired,
     review: PropTypes.object.isRequired,
+    setDenormalizedReview: PropTypes.func.isRequired,
     updateReviewText: PropTypes.func.isRequired,
   }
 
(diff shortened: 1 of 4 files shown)


kumar303:
Hmm, I might try this patch again tomorrow by dispatching an 'update review' action instead.

tofumatt:
> Hmm, I might try this patch again tomorrow by dispatching an 'update review' action instead.

Okay, I'll hold off reviewing for now if that's okay?

kumar303:
@tofumatt ok, this is ready to review.

I tried to use an update approach but it's not possible because of how the review state is currently structured. Since using an update approach would be a big change I'm going to file that as a separate issue.

src/amo/components/AddonReview.js
```diff
@@ -47,18 +48,34 @@ export class AddonReviewBase extends React.Component {
...
       errorHandler: this.props.errorHandler,
       reviewId: this.props.review.id,
+      ...newReviewParams,
     };
     // TODO: render a progress indicator in the UI.
     // https://github.com/mozilla/addons-frontend/issues/1156
+
+    // First, put the new review in state so that the
+    // component doesn't re-render with old data while
+    // our API request is in the air.
```
tofumatt: This is nitpicky, but "is in the air" is a bit of a colloquialism and might confuse an ESL developer or new contributor. "while our API request is in progress" or "before we get a response for our API request" might be a bit clearer.

src/amo/components/AddonReview.js
```diff
@@ -47,18 +48,34 @@ export class AddonReviewBase extends React.Component {
...
+    this.props.setDenormalizedReview(updatedReview);
+
+    // Next, update the review for real with an API call.
```
tofumatt: Again, a bit nitpicky, but "update the review for real" might be better written as "update the review on the server with an API call"

tests/client/amo/actions/testReviews.js
```diff
@@ -1,11 +1,46 @@
...
+import {
+  denormalizeReview,
+  setDenormalizedReview,
+  setReview,
+  setAddonReviews,
+} from 'amo/actions/reviews';
+import { SET_REVIEW } from 'amo/constants';
 import { fakeAddon, fakeReview } from 'tests/client/amo/helpers';
 
+// See reducer tests for more coverage of review actions.
```
tofumatt: I will be so happy when we just combine reducers and actions 😄

kumar303:
Thanks, that code is a bit odd so I think it helps to have understandable comments.

Issue #1749, opened 2017-02-08
After dismissing keyboard, it is not triggered again at subsequent search


vcarciu:
Steps to reproduce:
1.On Android, go to addons-dev.allizom.org
2.Tap on search field
3.Dismiss the keyboard by pressing back button
4.Tap a Category from add-ons
5. Tap search field again

Expected results:
Keyboard is successfully opened

Actual results:
The keyboard is not triggered, focus is not present, tapping multiple times does nothing

Please see video for this issue : 
![keyboardnottriggered](https://cloud.githubusercontent.com/assets/1583842/22738839/c399d5d0-ee11-11e6-9117-671cc6e95773.gif)

mstriemer:
This worked for me the first few times but I was eventually able to reproduce.

I have a feeling this is because of the animation and it is somehow breaking. The input won't focus until the magnifying glass animates over to the left. It looks like sometimes the animation skips and then focusing breaks.

Maybe this should just `setTimeout(() => this.input.focus(), animationLength)` instead of waiting on the `transitionend` event.

This is somewhat related to #1443 but that issue is just annoying, it doesn't necessarily break the functionality. Hopefully both of these can be fixed at once though.

Possible solution: making the input `color: transparent` appears to hide the cursor. Maybe it can be transparent until the time we expect the animation to end, then back to white.

We should fix the animation too, not sure what's up with that.

vcarciu:
Verified as fixed using different search scenarios , but I will add video only for original bug description : 
![keyboardfix](https://cloud.githubusercontent.com/assets/1583842/22926108/f311725e-f2b3-11e6-9ce5-48ae393bf0d3.gif)

PR #1790, opened 2017-02-10
Fix SearchInput issues

### LINKED ISSUE #1443: Search box is unresponsive until animation ends
Types characters are ignored while the magnifying glass is animating.

1. Click on the empty search bar.
2. Immediately start typing.

### What happened?
The first few characters are ignored until the animation ends.

The characters I type after clicking should all appear in the search bar regardless of the animation state.

uuuuck mov

mstriemer:
There were two problems with `SearchInput`, when typing immediately after focus keystrokes would be dropped, and when dismissing the virtual keyboard the animation would stop working and this prevented the input from being focused again.

This patch fixes both problems. The text is now transparent until the animation ends so keystrokes are not lost (instead of cancelling the focus event and manually focusing later).

The virtual keyboard issue was due to how the icon is moved over for animating (there are two icons, one on the label and one that animates to the left-aligned position). The `resize` event fires when the keyboard opens or closes and the icon's position would be recalculated but the calculation only worked if the icon had not been positioned yet. The original location of the icon is now saved when the calculation is first done which allows the `setIconPosition` function to correctly position the icon every time it is called.

search-keyboard-dismissal mp4

Fixes #1443 and #1749.

### DIFF
--- a/src/ui/components/SearchInput/index.js
+++ b/src/ui/components/SearchInput/index.js
@@ -31,7 +31,7 @@ export default class SearchInput extends React.Component {
   onBlur = () => {
     this.setState({ focus: false });
     if (!this.value) {
-      // Animation will start if there is no value.
+      // There is no transitionstart event, but animation will start if there is no value.
       this.setState({ animating: true });
     }
   }
@@ -39,7 +39,7 @@ export default class SearchInput extends React.Component {
   onFocus = () => {
     this.setState({ focus: true });
     if (!this.value) {
-      // Animation will start if there is no value.
+      // There is no transitionstart event, but animation will start if there is no value.
       this.setState({ animating: true });
     }
   }
--- a/src/ui/components/SearchInput/index.js
+++ b/src/ui/components/SearchInput/index.js
@@ -53,9 +53,11 @@ export default class SearchInput extends React.Component {
   };
 
   setIconPosition = () => {
+    if (!this.animateLeft) {
+      this.animateLeft = this.animateIcon.getBoundingClientRect().left;
+    }
     const { left: labelLeft } = this.labelIcon.getBoundingClientRect();
-    const { left: animateLeft } = this.animateIcon.getBoundingClientRect();
-    this.animateIcon.style.transform = `translateX(${labelLeft - animateLeft}px)`;
+    this.animateIcon.style.transform = `translateX(${labelLeft - this.animateLeft}px)`;
   }
 
   get value() {
--- a/src/ui/components/SearchInput/index.js
+++ b/src/ui/components/SearchInput/index.js
@@ -77,11 +77,17 @@ export default class SearchInput extends React.Component {
         ref={(el) => { this.root = el; }}
       >
         <Icon
-          name="magnifying-glass" className="SearchInput-animation-icon"
+          name="magnifying-glass"
+          className="SearchInput-animation-icon"
           getRef={(el) => { this.animateIcon = el; }}
-          onTransitionEnd={this.onTransitionEnd} />
+          onTransitionEnd={this.onTransitionEnd}
+        />
         <label className="SearchInput-label" htmlFor={id}>
-          <Icon name="magnifying-glass" getRef={(el) => { this.labelIcon = el; }} />
+          <Icon
+            name="magnifying-glass"
+            className="SearchInput-label-icon"
+            getRef={(el) => { this.labelIcon = el; }}
+          />
           {placeholder}
         </label>
         <input
--- a/src/ui/components/SearchInput/index.js
+++ b/src/ui/components/SearchInput/index.js
@@ -16,7 +16,7 @@ export default class SearchInput extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { focus: false, value: props.defaultValue };
+    this.state = { animating: false, focus: false, value: props.defaultValue };
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


src/ui/components/SearchInput/index.js
```diff
@@ -30,32 +30,34 @@ export default class SearchInput extends React.Component {
 
   onBlur = () => {
     this.setState({ focus: false });
+    if (!this.value) {
+      // Animation will start if there is no value.
```
mstriemer: There is no `transitionstart` event so we need to manually detect when we will start animating.

src/ui/components/SearchInput/style.scss
```diff
@@ -20,19 +20,16 @@ $transition-duration: 250ms;
...
-    opacity: 0;
-    position: absolute;
-    transition: opacity 1ms $transition-duration, transform $transition-duration;
-  }
+  opacity: 0;
+  position: absolute;
 }
 
 .SearchInput--text {
-  @include text-align-start();
```
mstriemer: This was just moving the label to the left which is hidden in this state so there was no point (and it might have contributed to messing up position calculations in `setIconPosition`.

src/ui/components/SearchInput/index.js
tofumatt: Makes sense, but I'd be fine if we added that comment to the actual code 😄

src/ui/components/SearchInput/style.scss
```diff
@@ -46,6 +43,15 @@ $transition-duration: 250ms;
     opacity: 1;
     // Use !important to overwrite style properties added in JS.
     transform: translateX(0) !important;
+  }
+}
+
+.SearchInput--animating {
+  .SearchInput-animation-icon {
+    transition: opacity 1ms $transition-duration, transform $transition-duration;
```
tofumatt: I see `1ms` and `$transition-duration`; I didn't know about that syntax. What does it mean? I thought the third argument was the transition curve, so `duration` seems like the wrong variable.

tests/client/ui/TestSearchInput.js
```diff
@@ -43,40 +51,45 @@ describe('<SearchInput />', () => {
...
-      assert.notOk(event.preventDefault.called);
-      resolve();
-    })
-  ));
+  it('clears the --animating class on transitionend', () => {
+    const root = renderIntoDocument(<SearchInput name="foo" />);
+    root.setState({ animating: true });
+    assert.ok(root.root.classList.contains('SearchInput--animating'));
+    Simulate.transitionEnd(root.animateIcon);
+    assert.notOk(root.root.classList.contains('SearchInput--animating'));
```
tofumatt: I would find these a bit easier to read if we did

```js
assert.include(root.root.classList, 'SearchInput--animating');
Simulate.transitionEnd(root.animateIcon);
assert.notInclude(root.root.classList, 'SearchInput--animating');
```

It's shorter but also I think a bit more distinct. 🤷‍♀️

src/ui/components/SearchInput/index.js
mstriemer: Haha, end of the day Mark was trying, he really was! I guess if I'm writing comments on GitHub to explain something they just belong in the code 😛.

src/ui/components/SearchInput/style.scss
mstriemer: The first one is the duration, the second is the delay. So the icon will get hidden after `$transition-duration` which is when the label re-appears. I might be able to just set this on `SearchInput--animating` now since we have that class. I'll fix it or add a comment.

tests/client/ui/TestSearchInput.js
mstriemer: `root.root.classList` is one of those things that is like an array but not actually an array. So `assert.include` fails if you try and use it.

```
  <SearchInput />
    ✖ clears the --animating class on transitionend
      Firefox 51.0.0 (Mac OS X 10.12.0)
    object tested must be an array, an object, or a string, but domtokenlist given

      + expected - actual

      
AssertionError@node_modules/chai/chai.js:5553:18
[12]</module.exports@node_modules/chai/chai.js:4285:11
include@node_modules/chai/chai.js:495:5
assert@node_modules/chai/chai.js:4121:24
[6]</module.exports/assert.include@node_modules/chai/chai.js:2994:5
@webpack:///tests/client/ui/TestSearchInput.js:89:4 <- test-runner.js:168913:6

mstriemer: Some helpers might make this clearer.

```js
const assertHasClass = (el, className) =>
  assert.ok(el.classList.contains(className));
const assertNotHasClass = (el, className) => 
  assert.notOk(el.classList.contains(className));

  it('clears the --animating class on transitionend', () => {
    const root = renderIntoDocument(<SearchInput name="foo" />);
    root.setState({ animating: true });
    assertHasClass(root.root, 'SearchInput--animating');
    Simulate.transitionEnd(root.animateIcon);
    assertNotHasClass(root.root, 'SearchInput--animating');
  });

tofumatt: Oh, shoot, didn't realise that. The helpers totally read better, if that's not a PITA I like it 👍

Otherwise though just a comment explaining why is fine so someone like me who thinks they're clever doesn't bother with it again. I'd say whichever is shorter and more obvious wins (the comment versus the helper) 😉

src/ui/components/SearchInput/style.scss
tofumatt: Oh gosh, right! I totally knew this but forgot. I think the comment helps because not everyone is familiar with that delay value (and those like me who know but forget will have their memory jogged). 👍

tofumatt:
All the explanations/comments here make total sense, thanks 👍

Issue #1782, opened 2017-02-10
'No reviews yet' is not styled correctly on add-on detail page


kumar303:
### Describe the problem and steps to reproduce it:

* Go to an add-on detail page for an add-on that has 0 reviews

### What happened?

The 'No reviews yet' text isn't styled correctly

<img width="387" alt="screenshot 2017-02-09 18 10 55" src="https://cloud.githubusercontent.com/assets/55398/22808774/53636186-eef3-11e6-803e-53a19bfe23a5.png">

### What did you expect to happen?

It should match the mock

vcarciu:
Verified as fixed : 
![norewstyle](https://cloud.githubusercontent.com/assets/1583842/22970634/73f1670e-f37b-11e6-953b-074e75053a70.png)

Issue #1771, opened 2017-02-09
White dot present in the warning for adding review for your own addon


vcarciu:
Steps to reproduce:
1.Submit an Android add-on on -dev and approve it
2.Try to rate your own add-on
3.Observe the warning

Expected results:
Warning is correctly displayed , no UI issues

Actual results:
White dot displayed in the left side of the warning without reason. 

NOTES:
We should round the corners of the red frame in order to meet the general amo styling. We can write _can not_ in a single  _cannot_

Please see screenshot for this issues : 

![ownreview](https://cloud.githubusercontent.com/assets/1583842/22789624/32ea92f0-eeec-11e6-9cd8-1b4498cc2d63.png)

muffinresearch:
It might be better to also say "their own add-on" instead of "its own add-on"?

kumar303:
Huh, that's weird about the white dot. Thanks. I'll make the style more consistent too.

RE: the error message, that will need to be fixed in https://github.com/mozilla/addons-server so someone could start by filing a bug there to change it. It's an error returned by this API endpoint: http://addons-server.readthedocs.io/en/latest/topics/api/reviews.html#post

vcarciu:
Verified as fixed : 
![ownaddon](https://cloud.githubusercontent.com/assets/1583842/22969596/b7c95d0a-f377-11e6-8b8b-0341dde81284.png)

PR #1793, opened 2017-02-10
Fix rating error style and fix 'No reviews yet' style

kumar303:
* Fixes https://github.com/mozilla/addons-frontend/issues/1771
  * The rating error no longer has a white dot
* Fixes https://github.com/mozilla/addons-frontend/issues/1782
  * The 'No reviews yet' text is now styled more consistently. We don't technically have a mock of this.
* The error style is more consistent with the palette now

screenshot 2017-02-13 18 04 47

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -114,9 +114,10 @@ export class AddonDetailBase extends React.Component {
     );
   }
 
-  readReviewsFooter() {
-    const { addon, i18n } = this.props;
+  renderRatingsCard() {
+    const { addon, i18n, RatingManager } = this.props;
     let content;
+    let footerPropName;
 
     if (addon.ratings.count) {
       const count = addon.ratings.count;
@@ -125,6 +126,7 @@ export class AddonDetailBase extends React.Component {
         { count: i18n.formatNumber(count) },
       );
 
+      footerPropName = 'footerLink';
       content = (
         <Link className="AddonDetail-all-reviews-link"
           to={`/addon/${addon.slug}/reviews/`}>
@@ -132,14 +134,29 @@ export class AddonDetailBase extends React.Component {
         </Link>
       );
     } else {
+      footerPropName = 'footerText';
       content = i18n.gettext('No reviews yet');
     }
 
-    return <div className="AddonDetail-read-reviews-footer">{content}</div>;
+    const props = {
+      [footerPropName]: (
+        <div className="AddonDetail-read-reviews-footer">{content}</div>),
+    };
+    return (
+      <Card
+        header={i18n.gettext('Rate your experience')}
+        className="AddonDetail-overall-rating"
+        {...props}>
+        <RatingManager
+          addon={addon}
+          version={addon.current_version}
+        />
+      </Card>
+    );
   }
 
   render() {
-    const { RatingManager, addon, i18n } = this.props;
+    const { addon, i18n } = this.props;
 
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
@@ -192,15 +209,7 @@ export class AddonDetailBase extends React.Component {
             } />
         </ShowMoreCard>
 
-        <Card
-          header={i18n.gettext('Rate your experience')}
-          footer={this.readReviewsFooter()}
-          className="AddonDetail-overall-rating">
-          <RatingManager
-            addon={addon}
-            version={addon.current_version}
-          />
-        </Card>
+        {this.renderRatingsCard()}
 
         <AddonMoreInfo addon={addon} />
       </div>
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -115,7 +115,7 @@ export class AddonDetailBase extends React.Component {
   }
 
   renderRatingsCard() {
-    const { addon, i18n, RatingManager } = this.props;
+    const { RatingManager, addon, i18n } = this.props;
     let content;
     let footerPropName;
 
--- a/src/core/css/ErrorHandler.scss
+++ b/src/core/css/ErrorHandler.scss
@@ -2,6 +2,7 @@
 
 .ErrorHandler-list {
   text-align: center;
+  border-radius: $border-radius-default;
   background-color: $error-background-color;
   color: $error-text-color;
   padding: 10px;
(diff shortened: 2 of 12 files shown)


src/core/css/ErrorHandler.scss
```diff
@@ -7,4 +7,8 @@
   padding: 10px;
   margin: 0;
   margin-bottom: 10px;
+
+  > li {
```
tofumatt: Usually we assign class names and just target the elements directly. I think there are some silly performance advantages to that in browsers too, but largely it's just our convention. So this would be `.ErrorHandler-listItem` or whatever.

src/ui/components/Card/Card.scss
```diff
@@ -45,9 +46,7 @@
   a:visited {
     color: $link-color;
     display: block;
-    font-variant: all-small-caps;
     font-weight: normal;
-    padding: 10px;
```
tofumatt: Hmm, this will reduce the tap target on the link itself. Before it would take up the entire footer which was a lot nicer; now if you tap on the edge of the footer it won't register a click/tap event. I think we might need to change the `<Card>` component to accept a `footerUrl` and wrap the contents in a link if we want the link to fill the entire footer.

Unless React can tell what the element inside `.Card-footer` is and apply a style that either pads the link or the `.Card-footer` class itself based on the element (though that is certainly more magically and less obvious).

So I think we need to change `<Card footer={footerHTMLWithALink}>` to `<Card footer={footerHTML} footerLink={'/someplace'}>`.

We can do that outside this PR if you want as it's not related, but we'll have to file an issue first.

kumar303: oh yeah, true, I'll see if I can preserve the big click target

tofumatt: Cool, up to you if you want to do it in this PR or just file an issue for it and do it outside of here, hence my not-explicit-approve-or-request-changes review.

If you want to file the issue and do it later I'm r+ though. Otherwise I'll have another look once it's finished.

src/core/css/ErrorHandler.scss
kumar303: I really don't see the point in trying to optimize our css for theoretical performance that we have no measurement of whatsoever. I know there are blog posts but those blog posts get outdated and even the right to left problem isn't so much of a problem anymore (and this was from 2011): https://calendar.perfplanet.com/2011/css-selector-performance-has-changed-for-the-better/

Adding css classes to every sub-element within a component is annoying and I just don't think it's worth it. I would rather see some performance metrics to prove what we should and should not be doing.

src/ui/components/Card/Card.scss
kumar303: I fixed the click target

kumar303:
@tofumatt could you take another look?

src/core/css/ErrorHandler.scss
tofumatt: If that's the case it's fine, my two _actually compelling_ reasons for using classes are:

1. it's convention we've been following thus far
2. it means we target things very specifically and never rely on nesting for styles

Obviously the refutation of those points is easy: 1. who cares 2. `<li>` elements are always inside `<ul>`

So wfm; in the future I won't write out classes on `<li>`s either and it'll likely be nicer 😄

src/amo/components/AddonDetail.js
```diff
@@ -114,9 +114,10 @@ export class AddonDetailBase extends React.Component {
     );
   }
 
-  readReviewsFooter() {
-    const { addon, i18n } = this.props;
+  renderRatingsCard() {
+    const { addon, i18n, RatingManager } = this.props;
```
tofumatt: Properties `ThatStartWithCaps` should go first if we're alphabetising props (as we tend to).

src/ui/components/Card/index.js
```diff
@@ -8,12 +8,26 @@ export default class Card extends React.Component {
...
 
   render() {
-    const { children, className, footer, header } = this.props;
+    const { children, className, footerText, footerLink, header } = this.props;
+
+    let footer;
+    let footerClass;
+    if (footerText && footerLink) {
+      throw new Error(
+        'You cannot specify footerLink and footerText at the same time');
```
tofumatt: fwiw I would have done `footer` and `footerLink` without this error.

It's fine for this PR but I'd actually imagine supplying a link would wrap the footer in the `<Link>` (and because HTML5 lets us wrap pretty much anything in an `<a>` that'd be fine). I would have supplied just the URL too. Essentially I'd think the API would be:

```jsx
<Card footer={<div>Content</div>} footerLink=`/addon/${addon.slug}` />
// Footer content => <a href="/addon/my-addon/" class="Card-footer-link"><div>Content</div></a>
```

That said, it's for a theoretical use case in the future when we'd wrap any content.

tofumatt:
Of course, you have to fix those `<a href="#">` test links because:

travis-says-no

src/ui/components/Card/index.js
kumar303: Yeah, I saw that in the other component but this is a little different because `footerLink` is a node of components. It's not a single `Link`, it's just a node that contains one or more `Links`. There is no way to inject `footerText` into it.

kumar303:
Hmm, yeah, `footerText` vs `footerLink` is a bit awkward. I guess if we force the caller to specify the link destination ([instead of a component](https://github.com/mozilla/addons-frontend/pull/1793/files#diff-0e4838cb628814da46134b8ff01e933dL129)) then that would be less awkward?

tofumatt:
Yeah, I think just specifying a link destination is best. We could still use external URLs if we wanted as `<Link>` accepts links starting with `/https?/` and it would cut down on boilerplate.

src/ui/components/Card/index.js
tofumatt: As mentioned, if we made `footerLink` a URL instead, we could do it.

kumar303:
I think creating a `Link` component internally out of a `footerUrl` and `footerText` is getting a little too hairy because we're making a lot of custom links like this: https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/ShowMoreCard/index.js#L55-L65

I think it's too much to put in this patch for now. Oh well.

tofumatt:
Oh shoot, that’s a good point. Yeah, too much for this patch, we can check it out later. All good for now!

src/core/css/ErrorHandler.scss
kumar303: Trying to keep AMO consistent is a bit dangerous since everything is in flux but that's a fair point about selector specificity. I switched it to a class.

Issue #842, opened 2016-07-27
Add exclusions for prefix middleware.


muffinresearch:
AMO has exclusions for the prefix middleware, so we're probably going to need something similar in addons-frontend.

muffinresearch:
Going to leave this one until we need it.

muffinresearch:
> Going to leave this one until we need it.

We're going to need this to support the redirection based on 404s it should match the exclusions list in addons-server.

muffinresearch:
List is here: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L262

On looking at this there's also a list of locale exclusions, we'll need to investigate if that's needed too: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L262

vcarciu:
Can you please add some STRs for this bug or add _qa:not needed_ label?

tofumatt:
On the new AMO frontend visit a URL like `/en-us/about` and `/developers/` to make sure you're served a page from the desktop site and don't get a 404 error.

PR #1798, opened 2017-02-12
Add URL exception prefixes in middleware

tofumatt:
This adds support for URLs like `/en-US/developers/`

Fixes https://github.com/mozilla/addons-frontend/issues/842


src/core/middleware.js
```diff
@@ -45,7 +50,19 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
     // We skip prepending an app if we'd previously prepended a lang and the
     // 2nd part of the URL is now a valid app.
     log.info('Application in URL is valid following prepending a lang.');
-  } else if (!hasValidClientApp) {
+  } else if (
+    (hasValidUrlException || isValidUrlException(URLParts[1], { _config }))
```
kumar303: If I delete `hasValidUrlException` from this statement, the tests still pass. Is it needed? If so, please add test coverage.

src/core/middleware.js
```diff
@@ -20,11 +24,12 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
 
   const hasValidLang = isValidLang(langFromURL);
   const hasValidClientApp = isValidClientApp(appFromURL, { _config });
+  const hasValidUrlException = isValidUrlException(appFromURL, { _config });
 
   let prependedLang = false;
 
   if ((hasValidLang && langFromURL !== lang) ||
-      (!hasValidLang && hasValidClientApp)) {
+      (!hasValidLang && (hasValidClientApp || hasValidUrlException))) {
```
kumar303: If I remove `!hasValidLang` from this if statement, all the tests still pass. Is it needed? If so, can you add some test coverage?

kumar303: If I remove `|| hasValidUrlException` from the if statement, all tests still pass. Is it needed? Does it need test coverage?

src/core/middleware.js
```diff
@@ -45,7 +50,19 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
...
+    if (hasValidLang) {
+      log.info('Exception in URL found; we fallback to addons-server.');
+
+      res.status(404).end(
+        'This page exists only on addons-server; returning 404 to fallback.');
+      return next();
+    }
+
+    log.info('Exception in URL found; prepending lang to URL.');
+  } else if (!hasValidClientApp && !hasValidUrlException) {
```
kumar303: When I remove `&& !hasValidUrlException` from the if statement, all tests pass. Is this one needed? Can you add test coverage?

config/default-amo.js
```diff
@@ -23,4 +23,27 @@ module.exports = {
       ],
     },
   },
+  validUrlExceptions: [
```
kumar303: Could you put a comment above this saying that it needs to be kept in sync with https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L262 for now?

Also, in a separate PR, could you add a comment to the addons-server settings? It should say that this list will eventually be handled in addons-frontend, maybe with a link back to this PR for now.

It's a shame that we'll have to keep them in sync.

tofumatt: Totally, I meant to do it (in both places) and it slipped my mind. 👍

tofumatt:
Thanks for helping out with the test coverage; as mentioned this code is a bit odd and some erroneous stuff stayed in while I missed one thing! All fixed up now, ready for another r?

src/core/middleware.js
```diff
@@ -20,11 +24,12 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
 
   const hasValidLang = isValidLang(langFromURL);
   const hasValidClientApp = isValidClientApp(appFromURL, { _config });
+  const hasValidUrlException = isValidUrlException(appFromURL, { _config });
 
   let prependedLang = false;
 
   if ((hasValidLang && langFromURL !== lang) ||
-      (!hasValidLang && hasValidClientApp)) {
+      (hasValidClientApp || hasValidUrlException)) {
```
kumar303: I don't think you need these parentheses since they are all being or'd together

src/core/middleware.js
```diff
@@ -45,6 +50,19 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
     // We skip prepending an app if we'd previously prepended a lang and the
     // 2nd part of the URL is now a valid app.
     log.info('Application in URL is valid following prepending a lang.');
+  } else if (
+    hasValidUrlException || isValidUrlException(URLParts[1], { _config })
```
kumar303: if I remove `hasValidUrlException ||`, the tests all pass. Can you add test coverage for this or remove it? Or is that just a side effect of checking `URLParts[1]`?

src/core/middleware.js
tofumatt: Indeed, sorry, I'm terrible at adding them in.

src/core/middleware.js
tofumatt: I think it's a side effect of checking for `isValidUrlException(URLParts[1], { _config })` but I'll see about adding a test that fails if it's gone. It could be that it's unneeded because at this point `URLParts[1]` should always be the app/exceptionURL so you're right it might be erroneous. If it is I'll change around the code so it's easier to follow; like you mentioned this middleware is a bit hairy to grok.

tofumatt:
Okay, tweaked things so it's more obvious to trace test coverage, etc.

Sorry I mistakenly did a `git commit --amend`. I promise I'll stop in the future and just rebase once it's r+! 😢

PR #1808, opened 2017-02-14
fix: Long author names cause horizontal scrolling

### LINKED ISSUE #1607: Long author names cause horizontal scrolling
A long author name without spaces will cause the page to have horizontal scrollbars. This text should wrap instead. Check the same thing for the add-on name too.

See https://amo.addons-dev.allizom.org/en-US/firefox/addon/update-scanner/.

aayushsanghavi:
Fixes #1607 
Before - (contains a horizontal scroll bar at the bottom)
before
After - (no horizontal scroll bar at the bottom)
after

### DIFF
--- a/src/amo/css/AddonDetail.scss
+++ b/src/amo/css/AddonDetail.scss
@@ -21,6 +21,7 @@ $page-margin: 20px 10px;
 .AddonDetail-author {
   display: block;
   opacity: 0.75;
+  word-wrap: break-word;
 
   a,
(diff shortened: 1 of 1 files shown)


kumar303:
Thanks @aayushsanghavi ! This was really helpful as we are scrambling right now to push out the first public version of these new mobile pages.

aayushsanghavi:
@kumar303 great! I'm glad to have been of help.

Issue #1711, opened 2017-02-06
Change "Mozilla Add-ons" to "Firefox Add-ons" in the empty search string message


madalincm:
### Steps to reproduce
1. Open addons-dev.allizom.org on an Android device.
2. Tap on the search field, leave it empty and perform a search.

### Expected results:
- We show all the available add-ons
- We show a correct message like: "Please enter... to search Firefox Add-ons" not "Mozilla Add-ons"

### Actual results:
No add-ons are shown. The following message is displayed: "Please enter a search term to search Mozilla Add-ons".

### Notes/Issues
Verified in FF51(Android 7.1.1) in addons-dev.allizom.org
![screenshot_20170206-171015](https://cloud.githubusercontent.com/assets/15685441/22652465/3b2e36c4-ec8f-11e6-83e6-09d166d2eefb.png)

tofumatt:
I think this is intended behaviour; to me returning results for a totally empty search is weird and kinda useless. We do handle this case, it's just we choose to display a message instead of show all add-ons (which is effectively what an empty search would show).

I think this works as intended.

madalincm:
What about the message then? It is ok to be: "Please enter a search term to search Mozilla Add-ons" and not: "Please enter a search term to search Firefox Add-ons"

tofumatt:
Actually that's a good point. We should change it, I'll reopen the bug just to edit the message, but I don't think that's a V1 blocker so I've removed the label.

tofumatt:
Oh sorry I misread the original bug and realise it was more about the message than the empty results. My bad!

eviljeff:
mentor @tofumatt 
If this is your first contribution, please refer to https://wiki.mozilla.org/Add-ons/Contribute/Code on how to get started.

kumar303:
fixed in https://github.com/mozilla/addons-frontend/pull/1815

vcarciu:
Verified as fixed : 
![searchfirefox](https://cloud.githubusercontent.com/assets/1583842/22971249/c2802462-f37d-11e6-819b-21dd2c30d434.png)

PR #1815, opened 2017-02-14
fix: Change "Mozilla Add-ons" to "Firefox Add-ons" in the empty search string message

aayushsanghavi:
#1711 
After - 
after


aayushsanghavi:
Sure thing!

Issue #1527, opened 2016-12-22
Handle displaying 401 API errors


kumar303:
### Describe the problem and steps to reproduce it:
Create a 401 API error in the UI (not on the server). This would do it:
* Go to an add-on page
* make sure you're not logged in
* Click on a star to add a rating

### What happened?

The `SET_ERROR` action payload looks like:
````
{
  response: {
    apiURL: "http://olympia.dev/api/v3/addons/addon/food-truck-pie/reviews/?lang=en-US",
    data: {
      detail: "Authentication credentials were not provided."
    },
    status: 401,
  }
}
````

````
error TypeError: val.forEach is not a function
    at http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:78017:14
    at Array.forEach (native)
    at http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:78015:41
    at getMessagesFromError (http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:78033:8)
    at errors (http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:78049:20)
    at combination (http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:33022:30)
    at i (<anonymous>:1:18467)
    at u (<anonymous>:1:18753)
    at <anonymous>:1:21682
    at Object.dispatch (http://127.0.0.1:3001/amo-0236d939da573fb964ad.js:32444:23)
````

### What did you expect to happen?

* no stack trace
* error shown at the top of the page

### Anything else we should know?

Eventually we should support prompting for log in when this happens but until then the error reporter should at least work

tofumatt:
I think I fixed this in #1685; will double-check.

kumar303:
Sadly, no, that patch didn't fix it but I think it's the same problem (something is trying to iterate over the response).

I think I'm getting a different stack trace for it now:
````
webpack:///./src/core/reducers/errors.js:20

Uncaught (in promise) TypeError: val.forEach is not a function
    at http://127.0.0.1:3001/amo-7822798f3661faa94ad3.js:90197:14
    at Array.forEach (native)
    at http://127.0.0.1:3001/amo-7822798f3661faa94ad3.js:90195:41
    at getMessagesFromError (http://127.0.0.1:3001/amo-7822798f3661faa94ad3.js:90213:8)
    at errors (http://127.0.0.1:3001/amo-7822798f3661faa94ad3.js:90229:20)
[...]

PR #1818, opened 2017-02-14
Handle displaying 401 errors and other non-form-validation API responses

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1527

These are unlikely but possible:

screenshot 2017-02-14 16 28 27

### DIFF
--- a/src/core/errorHandler.js
+++ b/src/core/errorHandler.js
@@ -38,7 +38,17 @@ export class ErrorHandler {
     return (
       <ul className="ErrorHandler-list">
         {this.capturedError.messages.map(
-          (msg) => <li className="ErrorHandler-list-item">{msg}</li>
+          (msg) => {
+            let msgString = msg;
+            if (typeof msgString === 'object') {
+              // This is an unlikely scenario where an API response
+              // contains nested objects within objects. If this
+              // happens in real life let's make it prettier.
+              // Until then, let's just prevent a stack trace.
+              msgString = JSON.stringify(msg);
+            }
+            return <li className="ErrorHandler-list-item">{msgString}</li>;
+          }
         )}
       </ul>
     );
--- a/src/core/reducers/errors.js
+++ b/src/core/reducers/errors.js
@@ -17,15 +17,19 @@ function getMessagesFromError(error) {
 
     Object.keys(error.response.data).forEach((key) => {
       const val = error.response.data[key];
-      val.forEach((msg) => {
-        if (key === 'non_field_errors') {
-          // Add generic messages for the API response.
-          apiMessages.push(msg);
-        } else {
-          // Add field specific error messages.
-          apiMessages.push(`${key}: ${msg}`);
-        }
-      });
+      if (Array.isArray(val)) {
+        val.forEach((msg) => {
+          if (key === 'non_field_errors') {
+            // Add generic messages for the API response.
+            apiMessages.push(msg);
+          } else {
+            // Add field specific error messages.
+            apiMessages.push(`${key}: ${msg}`);
+          }
+        });
+      } else {
+        apiMessages.push(val);
+      }
     });
 
     if (apiMessages.length) {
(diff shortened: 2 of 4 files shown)


src/core/errorHandler.js
```diff
@@ -38,7 +38,17 @@ export class ErrorHandler {
     return (
       <ul className="ErrorHandler-list">
         {this.capturedError.messages.map(
-          (msg) => <li className="ErrorHandler-list-item">{msg}</li>
+          (msg) => {
+            let msgString = msg;
+            if (typeof msgString === 'object') {
+              // This is an unlikely scenario where an API response
```
tofumatt: Looks good, but there should be an issue about it, right? (Or if there is and I missed it we should link it in this comment.)

src/core/reducers/errors.js
```diff
@@ -17,15 +17,19 @@ function getMessagesFromError(error) {
...
-          // Add generic messages for the API response.
-          apiMessages.push(msg);
-        } else {
-          // Add field specific error messages.
-          apiMessages.push(`${key}: ${msg}`);
-        }
-      });
+      if (Array.isArray(val)) {
+        val.forEach((msg) => {
+          if (key === 'non_field_errors') {
```
tofumatt: This seems like a magic constant; we should store this in `constants.js`.

src/core/reducers/errors.js
```diff
@@ -17,15 +17,19 @@ function getMessagesFromError(error) {
...
+            // Add generic messages for the API response.
+            apiMessages.push(msg);
+          } else {
+            // Add field specific error messages.
+            apiMessages.push(`${key}: ${msg}`);
```
tofumatt: Is this output to the user? If so it seems like an issue for i18n; not all locales will use the "Something: more info" grammar/symbols.

src/core/reducers/errors.js
kumar303: Why? That means if you want to figure out what it is you'd have to look in `constants.js`.

src/core/reducers/errors.js
kumar303: Oh yeah, true. Maybe `msg` would be enough (which should be localized).

src/core/errorHandler.js
kumar303: There's no issue and it probably won't happen because I don't know of an API response that does this but it is possible. If it does happen, someone from QA will see a weird JSON object and bug us. But without my protective code, React will throw an error saying that the child node can't be an object.

src/core/reducers/errors.js
tofumatt: If it's the only place it's used I guess it's fine, but if this is a special key from the API telling the front-end it's a generic error rather than one associated with a specific key (as it seems to be), storing that value in a constant seems our usual route, similar to using `ADDON_TYPE_THEME`

Maybe it's just because `'non_field_errors'` wasn't immediately obvious to me. A comment would help maybe.

src/core/errorHandler.js
tofumatt: Ah, okay, I wasn't sure if it was something that happened but was super-rare. Just being protective of the possible bad future = 👍

src/core/reducers/errors.js
kumar303: Constants introduce a level of abstraction so I think they are only good if more than one function uses the string. I can add a comment linking to the API docs http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#bad-requests

tofumatt: Good point. That's perfect. 👍

kumar303:
Good call on the comments, thanks.

I looked into the `msg` thing but it's tricky. The localization for something like a missing `body` in a review is 'This field is required'. When I join key and message we get 'body: This field is required' but obviously `body` won't be localized. I think it would be pretty confusing though if it just said 'This field is required' because you wouldn't know which one. 

I think for now I'm going to leave the key prefix in there but we should think of a better solution (in the API perhaps).

Coincidentally, I can no longer get the API to say 'body: This field is required' and I'm not sure why (https://github.com/mozilla/addons-frontend/issues/1837). The API docs say that it is not required, heh. Because of this, I don't think we will have any part of the UI (for the v1 release) that shows field errors so we can worry about this later.

PR #1819, opened 2017-02-15
feat: Add trailing slashes to URLs (fix #843)

### LINKED ISSUE #843: Consider adding an add-slashes middleware.
AMO automatically 301 redirects urls missing a slash to one with a slash. There are a few exceptions but generally this is fairly consistent. 

Presumably we'll need something similar for addons-frontend?

tofumatt:
This adds a config option to enable a middleware that adds trailing
slashes to all URLs that don't have one. It's only enabled on AMO
for now as Disco Pane definitely have URLs that are declared without
trailing slash and this does the dumbest thing possible which is
redirect all requests.

### DIFF
--- a/src/core/middleware.js
+++ b/src/core/middleware.js
@@ -102,3 +102,15 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
 
   return next();
 }
+
+export function trailingSlashesMiddleware(req, res, next) {
+  const URLParts = req.originalUrl.split('?');
+
+  // If the URL doesn't end with a trailing slash, we'll add one.
+  if (URLParts[0].substr(-1) !== '/') {
+    URLParts[0] = `${URLParts[0]}/`;
+    return res.redirect(301, URLParts.join('?'));
+  }
+
+  return next();
+}
--- a/config/default-amo.js
+++ b/config/default-amo.js
@@ -2,7 +2,6 @@ const amoCDN = 'https://addons.cdn.mozilla.net';
 const staticHost = 'https://addons-amo.cdn.mozilla.net';
 
 module.exports = {
-  fxaConfig: 'amo',
   CSP: {
     directives: {
       formAction: [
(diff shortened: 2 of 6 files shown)


tofumatt:
Sorry for more middleware reviews!

src/core/middleware.js
```diff
@@ -81,3 +81,15 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
 
   return next();
 }
+
+export function trailingSlashesMiddleware(req, res, next) {
+  const URLParts = req.originalUrl.split('?');
```
kumar303: Do you know if `req.originalUrl` is URL encoded or is it already decoded? I like how simple your patch is but I wonder if we should go with something more battle tested like [url.parse()](https://nodejs.org/docs/latest/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost).

tofumatt: Good question. I tried it with multiple `?` in the URL (https://github.com/mozilla/addons-frontend/pull/1819/files#diff-6b5535b4fe81a5387879c0140e976318R184), but I'll see if `url.parse` will work fine (I'm sure they've tested their code more than I did!)

muffinresearch:
I notice that AMO does have urls without a slash https://addons-dev.allizom.org/en-US/about will this work ok with that?

tofumatt:
> I notice that AMO does have urls without a slash https://addons-dev.allizom.org/en-US/about will this work ok with that?

Oh gosh, that's not great.

I think the way it works (because `/about/` is redirected to `/about` on the server) it might break. I'll check into this. I wanted to avoid more complex code that only redirected if the frontend found a 404... though even then this will be weird because `/about` would 404 on the frontend... then redirect to `/about/`, which 404s to fallback to the server which 301s on the server to `/about`.

*sigh*

src/core/middleware.js
tofumatt: It's encoded; seems like my code covers it.

tofumatt:
I was worried I'd have to add a hack but actually this works fine with URLs like `/en-US/about`. It won't work with things like `/en-US/firefox/about` if they exist. If that's a problem we'll need a more elaborate fix or we'll need to ditch this for now.

tofumatt:
I added a few server tests to ensure what @muffinresearch asked about was true (URL exceptions aren't redirected because they're served as 404 before the `trailingSlashes` middleware is used).

PR #1824, opened 2017-02-15
fix: Hide search sort in categories (fix #1626)

### LINKED ISSUE #1626: Sort appears in category pages but doesn't work
Visit http://localhost:3000/en-GB/firefox/extensions/feeds-news-blogging/ and activate the search sort.

### What happened?

Nothing.

The sort should work... or it shouldn't appear (but the former is better).

The URLs are largely hooked up properly (though there is some erroneous data in there), but the params are likely just not passed through.

-----

sortmobile

tofumatt:
Removes the search sort from categories as it isn't needed there and was buggy.

-----

### Before
sortmobile

### After
screenshot 2017-02-15 18 19 29

### DIFF
--- a/src/amo/containers/CategoryPage.js
+++ b/src/amo/containers/CategoryPage.js
@@ -1,4 +1,5 @@
 import deepEqual from 'deep-eql';
+import React from 'react';
 import { connect } from 'react-redux';
 import { asyncConnect } from 'redux-connect';
 import { compose } from 'redux';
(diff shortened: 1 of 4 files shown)


src/amo/containers/CategoryPage.js
```diff
@@ -9,6 +9,7 @@ import { apiAddonType } from 'core/utils';
 
 
 export function mapStateToProps(state, ownProps) {
+  const enableSearchSort = false;
```
kumar303: I guess overloading `mapStateToProps` for more than just state is OK but it seems a little weird. I think you can use composition to accomplish this a little cleaner:
````js
function CategoryPageBase(props) {
  return <SearchPage enableSearchSort={false} hasSearchParams {...props} />;
}

export default compose(
  ...,
  connect(mapStateToProps),
)(CategoryPageBase);
````

tofumatt: Oh that's nicer and I agree. I'll give it a go.

tofumatt: I did that and adjusted the tests (which become ever-so-slightly more verbose) but I like the compositon over overloading `mapStateToProps`, thanks 👍

PR #1825, opened 2017-02-15
fix: &nbsp; in FR string (close #1513)

### LINKED ISSUE #1513: "How are you enjoying your experience?" is localised with a &nbsp; in French
There should just be a regular space before the "?" in the sentence but the localisation includes a `&nbsp;` instead, which we escape and thus display as escaped text in French.

This happens on the add-ons details page (example link: https://addons-dev.allizom.org/fr/firefox/addon/test-ie/)

See attached screenshot:

screenshot 2016-12-21 16 29 45

-----

We may want to point out to localisers that we can't accept HTML in our most of our translations, I'm thinking.

tofumatt:
I committed the built file so QA could test it instantly and because why not? It'll just get overwritten later and this makes it easier to QA.

### Before
screenshot 2017-02-15 19 22 41

### After
screenshot 2017-02-15 19 20 12


kumar303:
Since I'm too lazy to copy/paste could you include issue numbers like #1513 in the pull request description :) I know it will auto-close either way so definitely not a biggie.

tofumatt:
> Since I'm too lazy to copy/paste could you include issue numbers like #1513 in the pull request description :) I know it will auto-close either way so definitely not a biggie.

ahaha I totally am too and I was reviewing something awhile ago when that bugged me so I should know better. Noted 😆

PR #1827, opened 2017-02-15
fix: Add protocol to homepage URLs (fix #1647)

### LINKED ISSUE #1647: Prepend protocol to links that are missing one
Looks like the links in "More information" are relative:

screenshot 2017-01-23 13 53 50

Example:

https://addons-dev.allizom.org/en-US/firefox/addon/awesome-screenshot-plus-/

tofumatt:
Adds `http://` to homepage URLs that are missing it; will fix https://addons-dev.allizom.org/en-US/firefox/addon/awesome-screenshot-plus-/

Also omits the `http://` in the actual text of the URL in the page, which looks nicer.

I guess this means `gopher://` and `ftp://` homepages won't work. 😉

Seriously though it means a protocol other than HTTP(S) won't work, but I figure that's fine.

### DIFF
--- a/src/amo/components/AddonMoreInfo.js
+++ b/src/amo/components/AddonMoreInfo.js
@@ -17,14 +17,21 @@ export class AddonMoreInfoBase extends React.Component {
   render() {
     const { addon, i18n } = this.props;
 
+    let homepage = addon.homepage ? addon.homepage.trim() : null;
+    if (homepage && !homepage.match(/^https?:\/\//)) {
+      homepage = `http://${homepage}`;
+    }
+
     return (
       <Card className="AddonMoreInfo" header={i18n.gettext('More information')}>
         <dl className="AddonMoreInfo-contents">
           {addon.homepage ? <dt>{i18n.gettext('Website')}</dt> : null}
-          {addon.homepage ? (
+          {homepage ? (
             <dd>
-              <a href={addon.homepage}
-                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+              <a href={homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>
+                {homepage.replace(/^https?:\/\//, '')}
+              </a>
             </dd>
           ) : null}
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonMoreInfo.js
```diff
@@ -17,14 +17,21 @@ export class AddonMoreInfoBase extends React.Component {
...
         <dl className="AddonMoreInfo-contents">
           {addon.homepage ? <dt>{i18n.gettext('Website')}</dt> : null}
-          {addon.homepage ? (
+          {homepage ? (
             <dd>
-              <a href={addon.homepage}
-                ref={(ref) => { this.homepageLink = ref; }}>{addon.homepage}</a>
+              <a href={homepage}
+                ref={(ref) => { this.homepageLink = ref; }}>
+                {homepage.replace(/^https?:\/\//, '')}
```
kumar303: It might make sense to pull out `/^https?:\/\//` into a shared constant

tofumatt:
> I guess there's another bug to route all these through outgoing?

Hmm, that's actually a good point. I made this patch in case bogus data was returned but we decided it's the API's sole job to create outgoing links (because we need a secret to create them and keeping their creation in one place made sense).

But now I'm thinking the API should be prefixing these anyway and stripping out the `http://` on outgoing links might be weird.

I guess all this solves the issue of some weird data on at least `-dev`...

src/amo/components/AddonMoreInfo.js
```diff
@@ -17,14 +17,21 @@ export class AddonMoreInfoBase extends React.Component {
   render() {
     const { addon, i18n } = this.props;
 
+    let homepage = addon.homepage;
+    if (homepage && !homepage.match(/^https?:\/\//)) {
```
muffinresearch: Should this trim whitespace at all? Hopefully it won't have it but would be good to check. Because ' http://...' would end up being prefixed even though within an href it would still work.

tofumatt: It should never happen, but good point: why not?

PR #1839, opened 2017-02-16
fix: Encode query params (fix #1680)

### LINKED ISSUE #1680: After performing a search for the symbol  “&” then performing a refresh of the web page, the message “Internal Server Error” is displayed
Perform a search after the symbol “&”, then refresh the web page.

### What happened?
The message “Internal Server Error” is displayed.

A refresh of the web page.

Verified on FF51.0 (Android 7.1.1) in AMO-dev server.
Here is a video:
internalsymbol

tofumatt:
Fix #1680.

Can't believe we weren't doing this!

### Before
screenshot 2017-02-16 19 53 45
screenshot 2017-02-16 19 53 35

### After
screen shot 2017-02-16 at 19 51 57

### DIFF
--- a/src/amo/components/SearchForm.js
+++ b/src/amo/components/SearchForm.js
@@ -25,9 +25,10 @@ export class SearchFormBase extends React.Component {
 
   goToSearch(query) {
     const { api, pathname } = this.props;
-    const encodedQuery = query ? encodeURIComponent(query) : '';
-    this.context.router.push(
-      `/${api.lang}/${api.clientApp}${pathname}?q=${encodedQuery}`);
+    this.context.router.push({
+      pathname: `/${api.lang}/${api.clientApp}${pathname}`,
+      query: { q: query },
+    });
   }
 
   handleSearch = (e) => {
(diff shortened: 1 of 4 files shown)


src/amo/components/SearchForm.js
```diff
@@ -25,8 +25,9 @@ export class SearchFormBase extends React.Component {
 
   goToSearch(query) {
     const { api, pathname } = this.props;
+    const encodedQuery = query ? encodeURIComponent(query) : '';
     this.context.router.push(
```
kumar303: I think it would be safer to rely on the [router API](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#pushpathorloc):
````js
this.context.router.push({ pathname: ..., query: ... });
````

tofumatt: Oh, true. I guess the previous way was just too naive (with the string interpolation). 👍

tofumatt:
Okay, using `router.push` now 😄

PR #1845, opened 2017-02-17
more system addons - commit fix

designakt:
Added more system addons and the new default dark & light themes...
Accidentally committed without creating a pull request yesterday, creating a pull request now.


muffinresearch:
The trailing spaces are causing the lint to fail.

Issue #1838, opened 2017-02-16
Switch to new review API endpoints


kumar303:
This is to adjust the UI so that it connects to the new review API endpoint: https://github.com/mozilla/addons-server/pull/4692

For QA, please check that this doesn't create any regressions for the following actions:
* Go to add-on detail page, add a star rating, edit star rating
* After selecting a star, create a review, edit the review
* From the add-on detail page, tap 'read N reviews' to see a list of all reviews

kumar303:
@muffinresearch @vcarciu oof, sorry. We should have waited on the frontend patch before landing the backend patch. The dev server will be broken until this lands. In all honesty, it took more effort/time to fix the frontend than I thought it would.

muffinresearch:
@kumar303 np - at least this way one patch will close multiple issues :)

PR #1849, opened 2017-02-17
Switch review components to new API endpoints

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1838

oof, this took way more effort than I thought it would.

### DIFF
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -69,7 +69,7 @@ export function getLatestUserReview({ user, addon } = {}) {
     if (!user || !addon) {
       throw new Error('Both user and addon must be specified');
     }
-    // The API will only return the latest user review.
+    // The API will only return the latest user review for this add-on.
     resolve(getReviews({ user, addon }));
   })
     .then((reviews) => {
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -50,9 +50,7 @@ export function getReviews({ user, addon, ...params } = {}) {
     }
     resolve(callApi({
       endpoint: 'reviews/review',
-      params: {
-        user, addon, ...params,
-      },
+      params: { user, addon, ...params },
     }));
   })
     .then((response) => {
@@ -78,8 +76,7 @@ export function getLatestUserReview({ user, addon } = {}) {
       } else if (reviews.length === 0) {
         return null;
       }
-      throw new Error(
-        'Unexpectedly received multiple review objects: ' +
-        `${JSON.stringify(reviews)}`);
+      throw new Error(dedent`Unexpectedly received multiple review objects:
+        ${JSON.stringify(reviews)}`);
     });
 }
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -8,36 +8,30 @@ export function submitReview({
   addonId,
   rating,
   apiState,
-  addonSlug,
   title,
   versionId,
   body,
   reviewId,
   ...apiCallParams
 }) {
-  const data = { rating, version: versionId, body, title };
-  if (reviewId) {
-    // You cannot update the version of an existing review.
-    data.version = undefined;
-  } else {
-    if (!addonId) {
-      throw new Error('addonId is required when posting a new review');
-    }
-    data.addon = addonId;
-  }
-
   return new Promise(
     (resolve) => {
-      // TODO: remove addonSlug entirely.
-      if (!addonSlug) {
-        throw new Error('addonSlug is required to build the endpoint');
-      }
+      const data = { rating, version: versionId, body, title };
       let method = 'POST';
       let endpoint = 'reviews/review';
+
       if (reviewId) {
         endpoint = `${endpoint}/${reviewId}`;
         method = 'PATCH';
+        // You cannot update the version of an existing review.
+        data.version = undefined;
+      } else {
+        if (!addonId) {
+          throw new Error('addonId is required when posting a new review');
+        }
+        data.addon = addonId;
       }
+
       resolve(callApi({
         endpoint,
         body: data,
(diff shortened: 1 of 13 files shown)


src/amo/api/index.js
```diff
@@ -42,30 +43,16 @@ export function submitReview({
...
-    if (!userId) {
-      throw new Error('userId cannot be falsey');
+    if (!user && !addon) {
+      throw new Error('Either user or addon must be specified');
     }
     resolve(callApi({
-      endpoint: `accounts/account/${userId}/reviews`,
+      endpoint: 'reviews/review',
+      params: {
+        user, addon, ...params,
```
tofumatt: Looks like this might fit on one line, but that's a nitpick.

src/amo/api/index.js
```diff
@@ -74,23 +61,25 @@ export function getUserReviews({ userId, addonId } = {}) {
...
-      const latest = reviews.filter((review) => review.is_latest);
-      if (latest.length === 0) {
+      if (reviews.length === 1) {
+        return reviews[0];
+      } else if (reviews.length === 0) {
         return null;
       }
-      return latest[0];
+      throw new Error(
+        'Unexpectedly received multiple review objects: ' +
```
tofumatt: Usually we'd use template strings here (and something like `oneLine` to combine them rather than string concat. I guess if we don't have an eslint rule preventing it though it's fine. I understand this is actually fewer lines/imports than using `oneLine` or `dedent` (though I like them more than string concat, especially when you're presumably casting it to a string anyway with the template string on the next line.

src/amo/components/AddonReview.js
```diff
@@ -40,43 +37,38 @@ export class AddonReviewBase extends React.Component {
...
+      reviewId: review.id,
       ...newReviewParams,
     };
     // TODO: render a progress indicator in the UI.
     // https://github.com/mozilla/addons-frontend/issues/1156
 
-    // First, dispatch the new review to state so that the
-    // component doesn't re-render with old data while
+    // Dispatch the new review to state so that the
+    // component doesn't re-render with stale data while
```
tofumatt: That is a much clearer comment, thanks 👍

src/core/api/index.js
```diff
@@ -20,7 +20,15 @@ export const category = new Entity('categories', {}, { idAttribute: 'slug' });
 export const user = new Entity('users', {}, { idAttribute: 'username' });
 
 export function makeQueryString(query) {
-  return url.format({ query });
+  const resolvedQuery = { ...query };
+  Object.keys(resolvedQuery).forEach((key) => {
+    if (resolvedQuery[key] === undefined) {
```
tofumatt: If sending empty keys triggers bugs (and is also just ugly, I'm all for removing them) we could be more defensive about empty values (eg: `null` and `''` should also cause key deletion).

tests/client/amo/components/TestAddonReview.js
```diff
@@ -92,8 +94,8 @@ describe('AddonReview', () => {
           apiState: signedInApiState,
         });
 
-        // Make sure the overlay was hidden after finishing.
-        assert.ok(overlayCard.hide.called);
+        assert.ok(onReviewSubmitted.called,
+                  'onReviewSubmitted() should have been called after updating');
```
tofumatt: Super nitpick but I think we're sticking to just two-space indentation rather than alignment.

tests/client/amo/components/TestRatingManager.js
```diff
@@ -134,16 +138,44 @@ describe('RatingManager', () => {
...
+      });
+  });
+
+  it('renders and configures AddonReview after submitting a rating', () => {
+    const userReview = setReview(fakeReview).payload;
+    const FakeAddonReview = sinon.spy(() => <div />);
+    const root = render({ AddonReview: FakeAddonReview, userReview });
+
+    assert.equal(FakeAddonReview.called, false,
+                 'expected AddonReview to initially not be visible');
```
tofumatt: Nitpick again but we've been doing two-space instead of alignment.

tests/client/amo/testApi.js
```diff
@@ -165,86 +138,63 @@ describe('amo.api', () => {
...
         }));
 
-      return getLatestUserReview({
-        userId: 123,
-        addonId: fakeReview.addon.id,
-      })
-        .then((review) => {
-          assert.deepEqual(review, latestReview);
+      return getLatestUserReview({ user: 123, addon: fakeReview.addon.id })
+        .then(() => assert(false, 'unexpected success'), (error) => {
```
tofumatt: We have an `unexpectedSuccess` helper that prevents writing out the `() => assert(false, 'unexpected success')` as many times and I think makes the code cleaner.

Also, why not use `catch` here instead of the second promise argument? (We usually use catch)

src/amo/api/index.js
kumar303: oh duh, forgot about that

src/core/api/index.js
kumar303: Whoops, good call. I also synced up with mat on what DRF expects for boolean values.

tests/client/amo/components/TestAddonReview.js
kumar303: Sure. That would be easy to add a lint rule for so we should do it. It's not easy to add lint rules for vertical alignment.

tests/client/amo/testApi.js
kumar303: If I use `catch()` it will also catch the 'unexpected success' error which could cause confusion. By using the second callback to `then()`, it will only catch true failures.

kumar303:
Thanks for the review. I got input from mat and synced up the falsey/truthy handling with how DRF wants us to do it.

Issue #1696, opened 2017-02-02
All user-generated links should use outgoing


muffinresearch:
As noted in https://github.com/mozilla/addons-frontend/issues/1647#issuecomment-274632001 there's currently some api responses with links that are not setup to be run through outgoing. This needs to be fixed prior to release.

Relates to https://github.com/mozilla/addons-server/issues/4270 which was the bug filed for fixing this in the API.

Since outgoing links are signed I'm feel like fixing it in the API is the way to go to keep the signing in one place. This is issue is just here to track that we fix it in some way or other ahead of putting this into production.

tofumatt:
How will signed links work with things like https://github.com/mozilla/addons-frontend/pull/1685/files#diff-746716be1ee643b487345fd8abf59e5aR23 ?

Seems like we need to be able to generate outgoing links on the client side for pages where we have outgoing links, unless we aren't interested in tracking them, or want to hardcode in the particular outgoing URLs into our templates (which I guess would work if we put the link content in a variable and only set outgoing in PROD or something, but that seems weird).

muffinresearch:
So I should clarify that as all user-generated links, links we add are ok. I've updated the title  - good catch!

mstriemer:
I think technically that github link should be outgoing as well. Probably not as big of a deal since we clearly trust github to some extent by using it as a service but the same reasons for using outgoing would apply I believe.

If we've got some technical reason we can't do it it's probably not a huge deal, but if we're making outgoing links client side I think we may as well.

muffinresearch:
That specific example is fine since a) we defined the link b) the translation is run through DOMPurify in addition. 

I think we're ok if we're supplying the links even if they're 3rd party domains. However if there's a case where we have UGC containing links then all of those must be run through outgoing. Since the source of all that data should be coming from the API I think we should be good to look to get this resolved by mozilla/addons-server#4270 unless there's some other source of 3rd party content I've missed in which case we should re-think this and start working on adding a way to generate outgoing links in the frontend code.

muffinresearch:
This is covered by the API so marking this one as closed.

PR #1860, opened 2017-02-20
feat: Use summary text if description is empty

### LINKED ISSUE #1858: Populate "About this extension" section with the summary if there is no other info
Steps to reproduce:
1.On Android, open https://addons-dev.allizom.org/en-US/android/addon/gotomeeting
2.Observe _About this extension_ section

Expected results:
This section should load the summary if there are no other info about the addon

Actual results: 
Section is empty

NOTES:
How it looks on desktop : 
about

How it looks on Mobile : 
aboutmobile

### LINKED ISSUE #1861: Links from summary are not working on mobile
Steps to reproduce:
1.On Android , go to https://addons-dev.allizom.org/en-US/android/addon/clean-links
2.Observe the summary section under add-on title and author

Expected results:
If there are links in summary , users should be able to open them by tap

Actual results:
Links are not clickable (they are working on Desktop)

NOTES:
Desktop link to same add-on : https://addons-dev.allizom.org/en-US/firefox/addon/clean-links/?src=search

summarylinks

tofumatt:
fixes #1858
fixes #1861

### Before
aboutmobile

### After
screen shot 2017-02-20 at 19 07 43

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -161,6 +161,10 @@ export class AddonDetailBase extends React.Component {
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
 
+    const descriptionSanitized = addon.description && addon.description.length ?
+      sanitizeHTML(nl2br(addon.description), allowedDescriptionTags) :
+      sanitizeHTML(nl2br(addon.summary), allowedDescriptionTags);
+
     const title = i18n.sprintf(
       // L10n: Example: The Add-On <span>by The Author</span>
       i18n.gettext('%(addonName)s %(startSpan)sby %(authorList)s%(endSpan)s'), {
@@ -204,9 +208,7 @@ export class AddonDetailBase extends React.Component {
         )} className="AddonDescription">
           <div className="AddonDescription-contents"
             ref={(ref) => { this.addonDescription = ref; }}
-            dangerouslySetInnerHTML={
-              sanitizeHTML(nl2br(addon.description), allowedDescriptionTags)
-            } />
+            dangerouslySetInnerHTML={descriptionSanitized} />
         </ShowMoreCard>
 
         {this.renderRatingsCard()}
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -160,11 +160,11 @@ export class AddonDetailBase extends React.Component {
 
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
-
-    const descriptionSanitized = addon.description && addon.description.length ?
-      sanitizeHTML(nl2br(addon.description), allowedDescriptionTags) :
-      sanitizeHTML(nl2br(addon.summary), allowedDescriptionTags);
-
+    const description = addon.description ? addon.description : addon.summary;
+    const descriptionSanitized = sanitizeHTML(
+      nl2br(description), allowedDescriptionTags);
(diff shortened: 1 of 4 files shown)


muffinresearch:
With a minor change this patch could probably fix https://github.com/mozilla/addons-frontend/issues/1861 too.

tofumatt:
> With a minor change this patch could probably fix #1861 too.

It seems the server links have outgoing URLs so I'm not sure, we might need the API to prefix them so it could depend on #1696. I think on our side it's just a matter of linkifying them but then they wouldn't be using outgoing, so it might be better to have the API render out HTML 😢

muffinresearch:
DOMPurify is removing the links so there's nothing to add. Probably just requires allowing the same tags as the descriptions.

src/amo/components/AddonDetail.js
```diff
@@ -161,6 +161,10 @@ export class AddonDetailBase extends React.Component {
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
 
+    const descriptionSanitized = addon.description && addon.description.length ?
+      sanitizeHTML(nl2br(addon.description), allowedDescriptionTags) :
```
kumar303: I think you should pull out `sanitizeHTML(nl2br(...), allowedDescriptionTags)` into a new variable on the next line to avoid repetition.

src/amo/components/AddonDetail.js
```diff
@@ -161,6 +161,10 @@ export class AddonDetailBase extends React.Component {
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
 
+    const descriptionSanitized = addon.description && addon.description.length ?
```
kumar303: Isn't `addon.description.length` redundant? An empty string will always be falsey.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -139,6 +139,23 @@ describe('AddonDetail', () => {
                    'About this extension');
   });
 
+  it('uses the summary as the description if no description exists', () => {
+    const addon = { ...fakeAddon };
```
kumar303: This is the right thing to do but you reminded me that we should add `Object.freeze()` to both `fakeAddon` and `fakeReview` just like [signedInApiState](https://github.com/mozilla/addons-frontend/blob/master/tests/client/amo/helpers.js#L55). Could you add this in your patch?

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -139,6 +139,23 @@ describe('AddonDetail', () => {
...
+    delete addon.description;
+    const rootNode = renderAsDOMNode({ addon });
+    assert.equal(
+      rootNode.querySelector('.AddonDescription-contents').textContent,
+      addon.summary);
+  });
+
+  it('uses the summary as the description if no description exists', () => {
+    const addon = { ...fakeAddon, description: '' };
```
kumar303: For cases like this I think it's helpful to make an explicit dependency on the `summary` property, like this:
````js
const addon = { ...fakeAddon, description: '', summary: 'short text' };
...
assert.equal(
  rootNode.querySelector('.AddonDescription-contents').textContent,
  'short text');
````

src/amo/components/AddonDetail.js
tofumatt: True, sorry, I forget which languages do that and which don't.

tofumatt:
Aha yeah I now see this can fix #1861 as well, I'll add that fix in 👍

PR #1862, opened 2017-02-21
fix: Balance icons in detail

### LINKED ISSUE #1698: Balance star and user icons in add-on detail metadata
If you go to the detail for an add-ons page you see some metadata at the top about how many users it has and what its average rating is (screenshot below). Let's tweak this design based on @pwalm's comment (https://github.com/mozilla/addons-frontend/issues/1638#issuecomment-276976722) -- let's make the star only have one line of text underneath it so it looks balanced with the user icon.

2e9f2b04-e4a4-11e6-878d-a5f85774b067

tofumatt:
Fix #1698.

This balances the icon size of the user icon and the star icon. The best way to keep the max amount of context seemed to be adding a line below with the number of ratings (not reviews) if available. That way the numerical rating is given context (1/5 with 1,000 reviews is different than 1/5 with 3 reviews). See the screenshots below, but I think this is the best option.

The number of reviews is displayed lower in the page, which I think is better.

### Before
2e9f2b04-e4a4-11e6-878d-a5f85774b067

### After
screen shot 2017-02-21 at 11 53 17
screen shot 2017-02-21 at 13 33 37

### DIFF
--- a/src/amo/components/AddonMeta.js
+++ b/src/amo/components/AddonMeta.js
@@ -16,7 +16,7 @@ export class AddonMetaBase extends React.Component {
     const { addon, i18n } = this.props;
     const averageDailyUsers = addon.average_daily_users;
     const averageRating = addon.ratings.average;
-    const addonReviewCount = addon.ratings.count;
+    const addonRatingCount = addon.ratings.count;
 
     const userCount = i18n.sprintf(
       i18n.ngettext('%(total)s user', '%(total)s users', averageDailyUsers),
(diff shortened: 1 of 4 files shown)


src/amo/components/AddonMeta.js
```diff
@@ -16,7 +16,7 @@ export class AddonMetaBase extends React.Component {
     const { addon, i18n } = this.props;
     const averageDailyUsers = addon.average_daily_users;
     const averageRating = addon.ratings.average;
-    const addonReviewCount = addon.ratings.count;
+    const addonRatingCount = addon.ratings.count;
```
tofumatt: This was mislabelled, it seems. The API says that is a count of ratings, not reviews. https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail

kumar303: Yes, that's true. Technically, a review is just a rating that also has text but this count could contain ratings without text.

PR #1865, opened 2017-02-22
feat: Add loading indicator to search (fix #1219)

### LINKED ISSUE #1219: Implement a nice search result loading placeholder component
Instead of showing "Searching..." while we request and load more search results, we should show at least a spinner, but I think it'd be quite nice if we took a page out of the kind of wireframe-with-no-content loading screens on things like Facebook. The search result styles are consistent, so a nice animation with their boxes would make the loading screen possibly feel faster, if I'm making any sense.

tofumatt:
fixes #1219

### Video
feb-22-2017 00-51-11


tofumatt:
This is ready for review but it's not a blocker, just had it ready and figured I'd push it.

Feel free to leave this until later if you're pressed for time (or someone else can r? it if they're bored)

src/ui/components/LoadingIndicator/LoadingIndicator.scss
```diff
@@ -0,0 +1,17 @@
+.LoadingIndicator {
+  height: 48px;
+  width: 48px;
+
+  &.animating {
+    animation: spin 1.1s 1 cubic-bezier(0.42, 0.26, 0.04, 1.21);
```
kumar303: ooh fancy

src/ui/components/LoadingIndicator/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+  static defaultProps = {
+    children: null,
+    timeToReset: 500,
+  }
+
+  restartAnimation = (event) => {
+    const { timeToReset } = this.props;
+    const iconDOM = ReactDOM.findDOMNode(this.loadingIcon);
+    iconDOM.classList.remove('animating');
```
kumar303: Why is this needed? Can't you do it like this?
````js
restartAnimation = (event) => {
  ...
  this.setState({ animating: false });
}

render() {
  const className = classNames('LoadingIndicator', {
    animating: this.state.animating,
  });
  return <Icon className={className} name="fox" />;
}
````

File: src/amo/components/SearchResults.js [hunk 3]
```diff
@@ -29,10 +30,20 @@ class SearchResults extends React.Component {
...
     let messageText;
 
     if (loading) {
-      messageText = i18n.gettext('Searching...');
+      loadingMessage = (
+        <div className="LoadingIndicator-container"
+          ref={(ref) => { this.LoadingIndicatorContainer = ref; }}>
+          <LoadingIndicator>
+            <span className="visually-hidden">
+              {i18n.gettext('Searching…')}
```
kumar303: is the ellipsis okay for localization?

src/amo/components/SearchResults.js
```diff
@@ -29,10 +30,20 @@ class SearchResults extends React.Component {
...
+    let loadingMessage;
     let messageText;
 
     if (loading) {
-      messageText = i18n.gettext('Searching...');
+      loadingMessage = (
+        <div className="LoadingIndicator-container"
+          ref={(ref) => { this.LoadingIndicatorContainer = ref; }}>
+          <LoadingIndicator>
+            <span className="visually-hidden">
```
kumar303: It seems like it might be common to include hidden text like this. I think this would be nicer as:
````js
<LoadingIndicator altText={i18n.gettext('Searching…')} />
````

File: src/ui/components/LoadingIndicator/index.js [hunk 5]
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  static defaultProps = {
+    children: null,
+    timeToReset: 500,
+  }
+
+  restartAnimation = () => {
+    const { timeToReset } = this.props;
+    const iconDOM = ReactDOM.findDOMNode(this.loadingIcon);
```
kumar303: Why is this needed? Can't you do it like this?
````js
restartAnimation = (event) => {
  ...
  this.setState({ animating: false });
}

render() {
  const className = classNames('LoadingIndicator', {
    animating: this.state.animating,
  });
  return <Icon className={className} name="fox" />;
}
````

File: src/ui/components/LoadingIndicator/LoadingIndicator.scss (hunk 1 above)
Comment (tofumatt) [inline src/ui/components/LoadingIndicator/LoadingIndicator.scss]: I have @sarahmonster to thank 😄

File: src/amo/components/SearchResults.js (hunk 4 above)
Comment (tofumatt) [inline src/amo/components/SearchResults.js]: Oh that's nicer, agreed.

File: src/amo/components/SearchResults.js (hunk 3 above)
Comment (tofumatt) [inline src/amo/components/SearchResults.js]: Yeah, it's okay. The triple dots were there before and that's worse (and we use ellipsis elsewhere).

File: src/ui/components/LoadingIndicator/index.js (hunk 2 above)
Comment (tofumatt) [inline src/ui/components/LoadingIndicator/index.js]: Totally. Sometimes I forget about React state stuff but that's better AND easier to test. 👍

File: src/ui/components/LoadingIndicator/index.js [hunk 6]
```diff
@@ -0,0 +1,46 @@
...
+    altText: null,
+    timeToReset: 500,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { animating: true };
+  }
+
+  restartAnimation = () => {
```
mstriemer: You shouldn't need any JS to do this. The timing will be slightly different but roughly accurate to what you had.

```scss
.LoadingIndicator {
  animation: spin 1.6s infinite cubic-bezier(0.42, 0.26, 0.04, 1.21);
  height: 48px;
  width: 48px;
}

@keyframes spin {
  0% {
    transform: rotate(0deg);
  }

  66% {
    transform: rotate(720deg);
  }

  100% {
    transform: rotate(720deg);
  }
}

Comment (tofumatt) [inline src/ui/components/LoadingIndicator/index.js]: Oh shoot that's cool (and makes sense). Okay, I'll give that a go, cheers.

File: tests/client/ui/components/TestLoadingIndicator.js [hunk 7]
```diff
@@ -0,0 +1,37 @@
...
+    assert.isTrue(root.state.animating);
+
+    Simulate.animationEnd(iconNode);
+
+    assert.isFalse(root.state.animating);
+
+    setTimeout(() => {
+      assert.isTrue(root.state.animating);
+      done();
+    }, 10);
```
kumar303: These kind of tests are hard to write but doing it like this may fail occasionally since the timing is dependent on how fast the Travis VM is processing events. Also, slowing down our test suite by 10 milliseconds should be avoided when possible.

I think you have a few options:
1. Figure out how to return a promise from `restartAnimation()` and call it directly. You could add a separate test to make sure the `animationEnd` event is hooked up.
2. Use mock injection to replace `setTimeout` altogether. You could do it by passing it as a prop, for example. With this approach you'd have to trigger the callback manually, like `mockSetTimeout.firstCall.args[0]()` to simulate `setTimeout` usage.
3. Maybe if you move it all to CSS you can delete these tests entirely :)

tofumatt: As @mstriemer mentioned seems it can be done in CSS; I'm just fixing up another PR then gonna do that here. Should tidy it up a million.

mstriemer: These tests shouldn't be needed if this is done in CSS but sinon provides [fake timers](http://sinonjs.org/docs/#clock) that can be used to test code with `setTimeout()` synchronously if we need it elsewhere.

We're currently using them in [TestHoverIntent.js](https://github.com/mozilla/addons-frontend/blob/89ff059ead85a4c813d018388ed093d4a1b4eaf4/tests/client/core/components/TestHoverIntent.js#L51).

tofumatt:
Ready for another r? This one is pure CSS, thanks @mstriemer

src/ui/components/LoadingIndicator/index.js
```diff
@@ -0,0 +1,27 @@
...
+import './LoadingIndicator.scss';
+
+
+export default class LoadingIndicator extends React.Component {
+  static propTypes = {
+    altText: PropTypes.node,
+  }
+
+  static defaultProps = {
+    altText: null,
```
kumar303: I guess this could default to `gettext('Loading…')`, right?

tests/client/ui/components/TestLoadingIndicator.js
```diff
@@ -0,0 +1,19 @@
...
+
+import LoadingIndicator from 'ui/components/LoadingIndicator';
+
+
+describe('<LoadingIndicator />', () => {
+  function render(props) {
+    return renderIntoDocument(<LoadingIndicator {...props} />);
+  }
+
+  it('renders alText', () => {
```
kumar303: typo: altText

src/ui/components/LoadingIndicator/index.js
tofumatt: True, actually, good call. 👍

PR #1872, opened 2017-02-22
Fix theme module layout

### LINKED ISSUE #1746: Layout of theme module is incorrect
Theme module doesn't look anything like mocks:

Currently:
screen shot 2017-02-07 at 3 24 56 pm

Mocks:
screen shot 2017-02-07 at 3 25 25 pm

muffinresearch:
Fixes #1746

Fixes the theme layout - I made an assumption that we'd want to keep ratings. I also didn't look at the hyphenating of the text since that's covered elsewhere and needs discussion anyway.

Before: 

screen shot 2017-02-22 at 21 41 25

After:

screen shot 2017-02-22 at 21 41 11

I also added some belt and braces fallbacks if images aren't set. End users are unlikely to see this but it covers the eventuality and we already did this on the details page for icons.

It will need extending for themes details pages at some point too but that could be done when revisiting the functionality for trying on a theme.

Extensions:

screen shot 2017-02-22 at 19 51 43

Themes: 
screen shot 2017-02-22 at 19 46 54

### DIFF
--- a/src/core/css/SearchResult.scss
+++ b/src/core/css/SearchResult.scss
@@ -34,6 +34,10 @@
   margin: 0;
   padding: 0;
   text-decoration: none;
+
+  .SearchResult--theme & {
+    float: left;
+  }
 }
 
 .SearchResult-rating {
@@ -46,6 +50,11 @@
     margin: 0;
     justify-content: flex-start;
   }
+
+  .SearchResult--theme & {
+    float: right;
+    width: auto;
+  }
 }
 
 .SearchResult-author,
@@ -56,6 +65,10 @@
   font-weight: normal;
   margin: 0;
   padding: 0;
+
+  .SearchResult--theme & {
+    clear: both;
+  }
 }
 
 .SearchResult-author {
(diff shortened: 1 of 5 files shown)


src/core/css/SearchResult.scss
```diff
@@ -83,4 +96,37 @@
...
+  .SearchResult--theme & {
+    display: block;
+    float: right;
+    height: 64px;
+    object-fit: cover;
+    object-position: top right;
+    width: 100%;
+
+    [dir=rtl] & {
+      object-position: top left;
```
muffinresearch: One thing, I'm not sure if a real RTL browser will show a theme back to front? If so we should transform the images and do the same thing in the details page too.

tests/client/amo/components/TestSearchResult.js
```diff
@@ -84,4 +86,42 @@ describe('<SearchResult />', () => {
     assert.equal(node.textContent,
                  `Rated ${fakeAddon.ratings.average} out of 5`);
   });
+
+  it('Displays a placeholder if the icon is malformed', () => {
```
tofumatt: The nit-pickiest of all nitpicks but the "D" should be lowercased here.

tests/client/amo/components/TestSearchResult.js
```diff
@@ -84,4 +86,42 @@ describe('<SearchResult />', () => {
...
+    };
+    const themeRoot = renderResult(themeResult);
+    const iconPlaceholder = findRenderedDOMComponentWithClass(
+      themeRoot, 'SearchResult-icon');
+    const iconSrc = iconPlaceholder.src;
+    assert.ok(iconSrc.startsWith('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEAAA'), iconSrc);
+  });
+
+
+  it('Adds theme specific class', () => {
```
tofumatt: Same as above; our style has been to use lowercase for the test descriptions.

PR #1875, opened 2017-02-23
fix: Android lacks theme categories

### LINKED ISSUE #1723: No content in Themes categories
Steps to reproduce:
1. Load AMO-dev on your Android device
2. Click on any theme category i.e Wild

Expected results:
Themes are displayed.

Actual results:
No themes displayed.

Notes/Issues:
- "Browse all themes" link is working as expected.
  
Verified on AMO-dev FF51(Android 6.0.1)
Screenshot for this issue: 
wt

tofumatt:
Fixes #1723

Admittedly a very hacky fix but it works.

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -9,6 +9,7 @@ import { schema as normalizrSchema, normalize } from 'normalizr';
 import { oneLine } from 'common-tags';
 import config from 'config';
 
+import { ADDON_TYPE_THEME } from 'core/constants';
 import log from 'core/logger';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 
@@ -118,12 +119,32 @@ export function callApi({
 }
 
 export function search({ api, page, auth = false, filters = {} }) {
+  const _filters = { ...filters };
+  if (!_filters.clientApp && api.clientApp) {
+    log.debug(
+      `No clientApp found in filters; using api.clientApp (${api.clientApp})`);
+    _filters.clientApp = api.clientApp;
+  }
+  // TODO: This loads Firefox personas (lightweight themes) for Android
+  // until github.com/mozilla/addons-frontend/issues/1723#issuecomment-278793546
+  // and https://github.com/mozilla/addons-server/issues/4766 are addressed.
+  // Essentially: right now there are no categories for the combo
+  // of "Android" + "Themes" but Firefox lightweight themes will work fine
+  // on mobile so we request "Firefox" + "Themes" for Android instead.
+  // Obviously we need to fix this on the API end so our requests aren't
+  // overridden, but for now this will work.
+  if (
+    _filters.clientApp === 'android' && _filters.addonType === ADDON_TYPE_THEME
+  ) {
+    log.info(dedent`addonType: ${_filters.addonType}/clientApp:
+      ${_filters.clientApp} is not supported. Changing clientApp to "firefox"`);
+    _filters.clientApp = 'firefox';
+  }
   return callApi({
     endpoint: 'addons/search',
     schema: { results: [addon] },
     params: {
-      app: api.clientApp,
-      ...convertFiltersToQueryParams(filters),
+      ...convertFiltersToQueryParams(_filters),
       page,
     },
     state: api,
--- a/tests/client/core/api/test_api.js
+++ b/tests/client/core/api/test_api.js
@@ -296,6 +296,34 @@ describe('api', () => {
         .then(() => mockWindow.verify());
     });
 
+    it('changes theme requests for android to firefox results', () => {
+      // FIXME: This shouldn't fail if the args are in a different order.
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/addons/search/?app=firefox&type=persona&page=3&lang=en-US`)
+        .once()
+        .returns(mockResponse());
+      return api.search({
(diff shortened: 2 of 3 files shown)


src/amo/containers/CategoryPage.js
```diff
@@ -14,10 +14,21 @@ export function CategoryPageBase(props) {
 }
 
 export function mapStateToProps(state, ownProps) {
+  // HACK: This loads Firefox personas (lightweight themes) for Android
+  // until https://github.com/mozilla/addons-frontend/issues/1723#issuecomment-278793546
```
kumar303: In addition to linking here, I think it would be good to file an addons-server bug asking to insert the new mobile categories and link to that issue

src/amo/containers/CategoryPage.js
```diff
@@ -14,10 +14,21 @@ export function CategoryPageBase(props) {
 }
 
 export function mapStateToProps(state, ownProps) {
+  // HACK: This loads Firefox personas (lightweight themes) for Android
```
kumar303: I would suggest changing this to `TODO` since we already have those. It would give us one less thing to grep for.

src/core/api/index.js
```diff
@@ -111,7 +111,11 @@ export function search({ api, page, auth = false, filters = {} }) {
     endpoint: 'addons/search',
     schema: { results: [addon] },
     params: {
-      app: api.clientApp,
+      // We allow overrides to clientApp because of
+      // https://github.com/mozilla/addons-frontend/issues/1723
+      // When it stops requiring overrides to the clientApp this
+      // can likely be restored to `api.clientApp` only.
```
kumar303: I think this feature is fine. I don't think you'd need to remove it later.

src/core/searchUtils.js
```diff
@@ -108,10 +108,22 @@ export function loadByCategoryIfNeeded(
...
+
+  // HACK: This loads Firefox personas (lightweight themes) for Android
+  // until https://github.com/mozilla/addons-frontend/issues/1723#issuecomment-278793546
+  // is more fully addressed.
+  // Essentially: right now there are no categories for the combo
+  // of "Android" + "Themes" but Firefox lightweight themes will work fine
+  // on mobile so we request them.
+  // Obviously we need to fix this on the API end so our requests aren't
+  // overridden, but for now this will work.
+  const clientApp = params.application === 'android' ?
```
kumar303: The `loadByCategoryIfNeeded` function and the `mapStateToProps` function both look like they end up calling `core/api:search()`. Can't you just put the android->firefox hack in `search()` so that it only gets switched in one place?

tests/client/core/api/test_api.js
```diff
@@ -244,6 +244,20 @@ describe('api', () => {
         .then(() => mockWindow.verify());
     });
 
+    it('allows overrides to clientApp', () => {
+      // FIXME: This shouldn't fail if the args are in a different order.
```
kumar303: ugh, yeah, I've noticed this too. It will be a bit annoying to fix.

src/core/api/index.js
```diff
@@ -111,7 +111,11 @@ export function search({ api, page, auth = false, filters = {} }) {
...
+      app: filters.clientApp || api.clientApp,
```
kumar303: is there a test for this? I can't spot it.

src/core/searchUtils.js
tofumatt: Yeah actually, I can do it in there for sure. I think I put it in two places because it compares the filters sent to callApi and those in state... but putting in it `search` is better I think.

src/core/api/index.js
tofumatt: It's here: https://github.com/mozilla/addons-frontend/pull/1875/files#diff-da8c6fe27817853c3f04c9897c485535R247

But it's implicit because it's also about switching android -> firefox. I'll make it clearer.

src/amo/containers/CategoryPage.js
tofumatt: Done, filed https://github.com/mozilla/addons-server/issues/4766

tofumatt:
Ready for another r?

src/core/api/index.js
```diff
@@ -107,12 +108,28 @@ export function callApi({
 }
 
 export function search({ api, page, auth = false, filters = {} }) {
+  const _filters = { ...filters };
+  if (!_filters.clientApp && api.clientApp) {
```
kumar303: I think it would be nice to log a debug message here in case something surprising happens

src/core/api/index.js
```diff
@@ -107,12 +108,28 @@ export function callApi({
...
+  // until github.com/mozilla/addons-frontend/issues/1723#issuecomment-278793546
+  // and https://github.com/mozilla/addons-server/issues/4766 are addressed.
+  // Essentially: right now there are no categories for the combo
+  // of "Android" + "Themes" but Firefox lightweight themes will work fine
+  // on mobile so we request "Firefox" + "Themes" for Android instead.
+  // Obviously we need to fix this on the API end so our requests aren't
+  // overridden, but for now this will work.
+  if (
+    _filters.clientApp === 'android' && _filters.addonType === ADDON_TYPE_THEME
+  ) {
```
kumar303: this could use debug logging too since setting `firefox` could be surprising

tofumatt: Good call. 😄

PR #1885, opened 2017-02-23
Hook up Sentry for error reporting

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1559

We'll need ops to set `SENTRY_ENV` for each server but that can come in https://github.com/mozilla/addons-frontend/issues/1886

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,26 @@ Example: Building and running a production instance of the admin app:
 NODE_APP_INSTANCE=admin NODE_ENV=production npm run build && npm run start
 ```
 
+## Hooking up Sentry for error reporting
+
+To hook up [Sentry](https://sentry.io/) for reporting application errors, you
+need to add some configuration.
+
+First, define this environment variable before starting the server:
+````
+SENTRY_DSN=https://<key>@sentry.io/<project>
+````
+
+Next, make sure your application config allows you to simulate errors:
+````js
+module.exports = {
+  allowErrorSimulation: true,
+};
+````
+
+This will allow you to make a request to `/simulate-error/`. You should
+see a 500 response and Sentry should have a record of the actual error.
+
 ## Overview and rationale
 
 This project will hold distinct front-ends e.g:
--- a/README.md
+++ b/README.md
@@ -178,26 +178,6 @@ Example: Building and running a production instance of the admin app:
 NODE_APP_INSTANCE=admin NODE_ENV=production npm run build && npm run start
 ```
 
-## Hooking up Sentry for error reporting
-
-To hook up [Sentry](https://sentry.io/) for reporting application errors, you
-need to add some configuration.
-
-First, define this environment variable before starting the server:
-````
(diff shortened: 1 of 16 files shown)


config/default.js
```diff
@@ -175,4 +176,10 @@ module.exports = {
   fxaConfig: null,
 
   proxyEnabled: false,
+
+  // If true, enable a route that explicitly triggers a server error
+  // to test our internal error handler.
+  allowErrorSimulation: false,
```
tofumatt: For the test routes for error pages (`/404`, `/500`, etc.) I just checked for `isDevelopment`. If the user is in development mode AND added a `SENTRY_DSN` environment variable I'd say that's enough to say they want error simulation and the extra config is just more work/docs/etc.

tests/server/admin/TestViews.js
```diff
@@ -3,7 +3,7 @@ import request from 'supertest-as-promised';
 
 import { checkSRI, parseCSP, runTestServer } from '../helpers';
 
-describe('Search App GET requests', () => {
+describe('Admin views', () => {
```
tofumatt: 😅

src/core/server/base.js
```diff
@@ -115,6 +116,14 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
   const app = new Express();
   app.disable('x-powered-by');
 
+  const sentryDsn = config.get('sentryDsn');
+  if (sentryDsn) {
+    Raven.config(sentryDsn).install();
+    app.use(Raven.errorHandler());
+  } else {
+    log.warn('Sentry reporting is disabled because sentryDsn was falsey');
```
tofumatt: "was falsey" seems odd to write in a log. Could this be a bit more instructive?

`'Sentry reporting is disabled. Set config.sentryDsn to enable reporting.'`

config/stage.js
```diff
@@ -28,4 +28,6 @@ module.exports = {
       ],
     },
   },
+
+  allowErrorSimulation: true,
```
tofumatt: Ah, okay, my comment about just checking `isDevelopment` isn't quite fair here, because it's off on stage but we want error simulation.

README.md
```diff
@@ -178,6 +178,26 @@ Example: Building and running a production instance of the admin app:
...
+
+To hook up [Sentry](https://sentry.io/) for reporting application errors, you
+need to add some configuration.
+
+First, define this environment variable before starting the server:
+````
+SENTRY_DSN=https://<key>@sentry.io/<project>
+````
+
+Next, make sure your application config allows you to simulate errors:
```
tofumatt: This is enabled by default in everything but production; I would argue this section can just be removed as it's instructing the developer to enable something that's already enabled and is just extra work for nothing.

kumar303: ok, sure. I guess it was just documentation for ops to follow but maybe that's not so important.

Issue #1880, opened 2017-02-23
500 error page is blank


muffinresearch:
### Describe the problem and steps to reproduce it:

On addons-dev go to: /en-GB/firefox/users/edit/
Note the blank grey page.

### What happened?

A blank 500 error page. The traceback on the server is:

```
[1] [2017-02-23T11:24:01.392Z] ERROR: server/71328 on hanuman.local: "users" not found in API_ADDON_TYPES_MAPPING (app=amo)
[1]     AddonTypeNotFound: "users" not found in API_ADDON_TYPES_MAPPING
[1]         at apiAddonType (/Users/Whatevs/code/addons-frontend/src/core/utils.js:176:11)
[1]         at Object.loadByCategoryIfNeeded [as promise] (/Users/Whatevs/code/addons-frontend/src/core/searchUtils.js:112:16)
[1]         at /Users/Whatevs/code/addons-frontend/node_modules/redux-connect/lib/helpers/utils.js:184:38
[1]         at Array.reduce (native)
[1]         at _loop3 (/Users/Whatevs/code/addons-frontend/node_modules/redux-connect/lib/helpers/utils.js:182:46)
[1]         at /Users/Whatevs/code/addons-frontend/node_modules/redux-connect/lib/helpers/utils.js:200:7
[1]         at iterateOverResults (/Users/Whatevs/code/addons-frontend/node_modules/redux-connect/lib/helpers/utils.js:53:12)
[1]         at process._tickCallback (internal/process/next_tick.js:103:7)
```

There's also `Unhandled promise rejection` errors in the console.

### What did you expect to happen?

A missing addon type should probably be a 404. But also the 500 error page should be being displayed correctly.

Looks like this:

<img width="350" alt="screen shot 2017-02-22 at 10 10 50 am" src="https://cloud.githubusercontent.com/assets/211578/23220477/c51ce9c0-f8e7-11e6-8689-d1f636249fb0.png">

kumar303:
redux-connect strikes again. Whenever one of our callbacks accidentally throws an error instead of returning a promise, that error [is not caught by redux-connect](https://github.com/makeomatic/redux-connect/blob/master/modules/helpers/utils.js#L133) so it disappears on the event loop.

I have an idea for a fix that makes sure our callbacks never do that. 

[...]

Issue #1403, opened 2016-11-21
Always return a Promise for functions passed to asyncConnect


tofumatt:
Right now we return `true` for certain functions passed to `asyncConnect`, which is technically supported but is weird as it usually returns a Promise. Even returning `Promise.resolve(true)` would be nicer, probably.

See: https://github.com/mozilla/addons-frontend/pull/1398#discussion_r88937885

(Side note, we do this a lot in places where we're caching results, so maybe we just want to not cache stuff.)

kumar303:
I think returning a consistent value (a Promise) always makes everything easier: tests, calling code. The biggest win is that the calling code won't have to write defensive code like:
````javascript
const val = loadSomeData();
if (isPromise(val)) {
  return val.then(...)
}

kumar303:
I think https://github.com/mozilla/addons-frontend/pull/1890 partially addresses this. It makes sure that no errors are thrown. Making sure all callbacks return promises when successful seems good but I'm not sure if it will be so helpful.

kumar303:
We're going to remove async connect in https://github.com/mozilla/addons-frontend/issues/2356

PR #1890, opened 2017-02-23
Show a 500 page for asyncConnect loaders that throw

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1880

This makes sure that any `promise` callback passed to `asyncConnect` never throws an error. As for the bug in question, the request will now render a 500 page. This fix should also prevent any other page from circumventing our 500 handler in the same way.

### DIFF
--- a/src/admin/containers/AddonPage/index.js
+++ b/src/admin/containers/AddonPage/index.js
@@ -2,7 +2,7 @@ import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
 import { asyncConnect } from 'redux-connect';
 
-import { gettext as _, loadAddonIfNeeded } from 'core/utils';
+import { gettext as _, loadAddonIfNeeded, safePromise } from 'core/utils';
 import NotFound from 'core/components/ErrorPage/NotFound';
 import JsonData from 'admin/components/JsonData';
 
@@ -132,7 +132,7 @@ function mapStateToProps(state, ownProps) {
 const CurrentAddonPage = asyncConnect([{
   key: 'AddonPage',
   deferred: true,
-  promise: loadAddonIfNeeded,
+  promise: safePromise(loadAddonIfNeeded),
 }])(connect(mapStateToProps)(AddonPage));
 
 export default CurrentAddonPage;
--- a/src/admin/containers/AddonPage/index.js
+++ b/src/admin/containers/AddonPage/index.js
@@ -1,8 +1,8 @@
 import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { asyncConnect } from 'redux-connect';
+import { compose } from 'redux';
 
-import { gettext as _, loadAddonIfNeeded, safePromise } from 'core/utils';
+import { gettext as _, loadAddonIfNeeded, safeAsyncConnect } from 'core/utils';
 import NotFound from 'core/components/ErrorPage/NotFound';
 import JsonData from 'admin/components/JsonData';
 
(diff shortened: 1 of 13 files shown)


src/admin/containers/AddonPage/index.js
```diff
@@ -132,7 +132,7 @@ function mapStateToProps(state, ownProps) {
 const CurrentAddonPage = asyncConnect([{
```
tofumatt: I know it's not really related to this patch but it'd be nice to see `compose` here as we use it elsewhere and it makes things a bit easier to read 😄

src/admin/containers/AddonPage/index.js
```diff
@@ -132,7 +132,7 @@ function mapStateToProps(state, ownProps) {
...
   key: 'AddonPage',
   deferred: true,
-  promise: loadAddonIfNeeded,
+  promise: safePromise(loadAddonIfNeeded),
```
tofumatt: I wonder if we should have a helper function. `deferred: true, promise: safePromise(realPromiseMethod)` seems like a pattern we use a lot.

src/core/utils.js
```diff
@@ -201,3 +201,31 @@ export function getErrorComponent(status) {
...
+ * You would use it like this:
+ *
+ * export default compose(
+ *   asyncConnect([{
+ *     deferred: true,
+ *     promise: safePromise(loadSomeData),
+ *   }])
+ * )(YourComponentBase)
+ */
+export const safePromise = (promiseOrNot) => (...args) => {
```
tofumatt: I think `promiseOrNot` is supposed to be a joke-y variable but I don't get it 😉

tests/client/core/test_utils.js
```diff
@@ -537,3 +538,31 @@ describe('visibleAddonType', () => {
...
+    const callback = sinon.spy(() => Promise.resolve());
+    const asPromised = safePromise(callback);
+    return asPromised('one', 'two', 'three').then(() => {
+      assert.ok(callback.called, 'callback was never called');
+      assert.deepEqual(callback.firstCall.args, ['one', 'two', 'three']);
+    });
+  });
+
+  it('catches errors and returns them as rejected promises', () => {
+    const msg = 'well, that was unfortunate';
```
tofumatt: Just a nitpick but I dislike variables like `msg` when `message` or even `errorMessage` fits.

src/core/utils.js
kumar303: It was meant to be literal, not a joke. The value is either a promise or not. I can give it more of a generic name though.

tofumatt: Hah, true. I think knowing we want to move to #1403 it can just be called Promise and the caveat can be explained in a comment if needed.

Unless I'm misreading this.

src/admin/containers/AddonPage/index.js
kumar303: Yeah, good call. This crossed my mind too.

kumar303:
Thanks for the review

src/core/utils.js
```diff
@@ -201,3 +202,47 @@ export function getErrorComponent(status) {
...
+  const safeConfigs = configs.map((conf) => {
+    if (!conf.promise) {
+      // This is the only way we use asyncConnect() for now.
+      throw new Error(
+        dedent`Expected safeAsyncConnect() config to define a promise:
+        ${JSON.stringify(conf)}`);
+    }
+    return {
+      ...conf,
+      deferred: true,
```
tofumatt: I would expect this to be _before_ `...conf` so it can be overridden. We don't do that yet but the option would be nice in case we want to do some special UI loading in a specific component. Filed #1909.

PR #1895, opened 2017-02-24
chore: Add "tr" to langs

tofumatt:
I reformatted them as well because I hate shifting stuff like that around,
even when it's rare. Makes it easier to add on GitHub too where the margins
aren't as obvious.


muffinresearch:
r+

Issue #1898, opened 2017-02-24
"Server Error" page displayed while page refresh for some add-ons


ValentinaPC:
Steps to reproduce:
1. Load AMO-dev homepage
2. Search for "Venezuela" and access "Spanish (Venezuela) spell check dictionary" add-on details page
3. Refresh the page

Expected results:
There are no display or layout issues.

Actual results:
A "Server Error" page is loaded.

Notes/Issue:
- issue can be reproduce on desktop if on add-on details page from step 2 - https://addons-dev.allizom.org/en-US/android/addon/diccionario-espa%C3%B1ol-venezuela/?src=ss we click "View Mobile Site"
- the URL stays the same (https://addons-dev.allizom.org/en-US/android/addon/diccionario-espa%C3%B1ol-venezuela/?src=ss)
- other examples:
   - https://addons-dev.allizom.org/mk/android/addon/google-translate-auto-espa%C3%B1ol/?src=ss

Verified on FF51(Android 6.0.1, Win 7).
Issue is reproducing on AMO-dev.
Video for this issue on device:
![videotogif_2017 02 24_17 06 07](https://cloud.githubusercontent.com/assets/15685960/23308967/91fbb54c-fab5-11e6-8357-c190b323030e.gif)
Video for this issue on desktop:
![desktop](https://cloud.githubusercontent.com/assets/15685960/23309300/d4b9da20-fab6-11e6-87f3-7ad91a884781.gif)

I think that this is localization related. Supposition.

kumar303:
Good catch, thanks. Here is the stacktrace:
<img width="643" alt="screenshot 2017-02-24 09 57 26" src="https://cloud.githubusercontent.com/assets/55398/23310310/bdfddc26-fa77-11e6-8f2e-967b9a3679df.png">

kumar303:
`index.js:92` is `./src/amo/components/RatingManager/index.js:92`

ValentinaPC:
Verified this on AMo-dev FF51(Win 7, Android 6.0.1) and is still reproducing on both mobile and desktop.
Video from desktop:
![desktop](https://cloud.githubusercontent.com/assets/15685960/23353547/af0ce8dc-fcd5-11e6-8672-1a10b6f78224.gif)

kumar303:
Whoops. This was an entirely separate bug that was hidden by the first. A fix is forthcoming in https://github.com/mozilla/addons-frontend/pull/1920

ValentinaPC:
[...]

PR #1900, opened 2017-02-24
Fix log in to rate prompt for all add-ons

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1898

### DIFF
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -102,11 +102,13 @@ export class RatingManagerBase extends React.Component {
       case ADDON_TYPE_THEME:
         return i18n.gettext('Log in to rate this theme');
       case ADDON_TYPE_EXTENSION:
+        return i18n.gettext('Log in to rate this extension');
       default:
         if (!validAddonTypes.includes(addonType)) {
           throw new Error(`Unknown extension type: ${addonType}`);
         }
-        return i18n.gettext('Log in to rate this extension');
+        log.warn(`Using generic prompt for add-on type: ${addonType}`);
+        return i18n.gettext('Log in to rate this add-on');
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/amo/components/RatingManager/index.js
```diff
@@ -81,23 +88,37 @@ export class RatingManagerBase extends React.Component {
...
+      case ADDON_TYPE_SEARCH:
+        return i18n.gettext('Log in to rate this search engine');
+      case ADDON_TYPE_THEME:
+        return i18n.gettext('Log in to rate this theme');
+      case ADDON_TYPE_EXTENSION:
+      default:
+        if (!validAddonTypes.includes(addonType)) {
+          throw new Error(`Unknown extension type: ${addonType}`);
+        }
+        return i18n.gettext('Log in to rate this extension');
```
tofumatt: I understood this after a second but it might be more immediately obvious (and technically a bit more accurate) as:

```js
case ADDON_TYPE_EXTENSION:
  return i18n.gettext('Log in to rate this extension');
default:
  if (!validAddonTypes.includes(addonType)) {
    throw new Error(`Unknown extension type: ${addonType}`);
  }
  return i18n.gettext('Log in to rate this add-on');
```

And maybe even log that it's an unknown extension type not covered here. If we had a new "theme" type that was valid but not handled here it would be called, specifically, an "extension" when our grab-bag term "add-on" is more appropriate.

tests/client/amo/components/TestRatingManager.js
```diff
@@ -242,6 +201,81 @@ describe('RatingManager', () => {
...
+    });
+
+    it('renders a login prompt for themes', () => {
+      const prompt = getAuthPromptForType(ADDON_TYPE_THEME);
+      assert.include(prompt, 'theme');
+    });
+
+    it('cannot render a login prompt for unknown extension types', () => {
+      assert.throws(
+        () => getAuthPromptForType('xul'), /Unknown extension type: xul/);
```
tofumatt: 😆

src/amo/components/RatingManager/index.js
kumar303: Hmm, from all the latest UX work it looks like we call add-ons "extensions" now.

tofumatt: Ooops. My bad.

memo

kumar303: I guess it doesn't matter what we call it. I just don't want the site to be broken.

PR #1910, opened 2017-02-24
Use setTimeout() to simulate an event loop error

kumar303:
As a follow-up to https://github.com/mozilla/addons-frontend/issues/1559 I'd like to also see what Sentry does with an error thrown completely asynchronously in the event loop.


tofumatt:
Seems it fails the server tests https://travis-ci.org/mozilla/addons-frontend/jobs/205144289#L902

kumar303:
heh, whoops

PR #1915, opened 2017-02-25
fix: add locale URL exception (fix #1867)

### LINKED ISSUE #1867: Add-ons cannot be installed from disco pane when mamo cookie is set
Prerequisites:
- `extensions.webservice.discoverURL` is set to point to -dev (https://discovery.addons-dev.allizom.org/%LOCALE%/firefox/discovery/pane/%VERSION%/%OS%/%COMPATIBILITY_MODE%)
- `xpinstall.signatures.dev-root` set to true
- browser used: FF54

Steps to reproduce:
1. On AMO-dev (desktop) homepage,  go to footer and click "View Mobile Site"
2. Go to about:addons and try to install an add-on i.e. Video DownloadHelper

Expected results:
The add-on is installed.

Actual results:
The add-on cannot be installed.

Notes/Issues:
- issue is not reproducing for -stage or -production.
- issue is not reproducing on -dev either if first step is not fulfilled
Verified on FF54(Win 7). Issue is reproducing on AMO-dev.
Screenshot for this issue:
about_addons

### LINKED ISSUE #1991: RelaodEvery does not install and shows connection problem
Steps to reproduce : 

1. Go to https://addons-dev.allizom.org/en-US/firefox/addon/reloadevery/
2. Then use the slider button beside the star button to download it
  

Actual Result :

A pop-up comes which shows connection problem 

Expected Result :

It should have been installed. And if it is not compatible with mobile device then the pop-up should say that "it is not compatible with mobile device"  

Attachment  -

17198039_1282377388519532_929326456_n

tofumatt:
fix #1867
fix #1991

Sorry, more `middleware.js` URL handling. This patch allows root URLs like `/downloads/*` or `/contribute.json` to be served properly without infinite redirects.

### DIFF
--- a/config/default-amo.js
+++ b/config/default-amo.js
@@ -25,9 +25,34 @@ module.exports = {
   },
   enableTrailingSlashesMiddleware: true,
   fxaConfig: 'amo',
+
+  // This needs to be kept in sync with addons-server's SUPPORTED_NONLOCALES
+  // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L274
+  // These are URLs that are ignored by our prefix middleware that will add
+  // a locale (e.g. `en-US`) to any URL that doesn't have a valid locale.
+  // These are all URLs that should not get a locale prepended to the URL,
+  // because they are locale-independant, like `/firefox/downloads/`.
+  validLocaleUrlExceptions: [
+    'contribute.json',
+    'google1f3e37b7351799a5.html',
+    'robots.txt',
+    'services',
+    'downloads',
+    // This isn't in addons-server, but instead will cause a redirect to another
+    // host.
+    'blocklist',
+    'static',
+    'user-media',
+    '__version__',
+  ],
+
   // This needs to be kept in sync with addons-server's SUPPORTED_NONAPPS
   // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L262
-  validUrlExceptions: [
+  // These are URLs that are ignored by our prefix middleware that will add
+  // a clientApp (e.g. `android`) to any URL that doesn't have a valid
+  // clientApp. These are all URLs that don't require a clientApp in them
+  // because they are app-independant, like `/en-US/developers/`.
+  validClientAppUrlExceptions: [
     'about',
     'admin',
     'apps',
--- a/src/core/middleware.js
+++ b/src/core/middleware.js
@@ -3,7 +3,8 @@ import config from 'config';
 import {
   getClientApp,
   isValidClientApp,
-  isValidUrlException,
+  isValidLocaleUrlException,
+  isValidClientAppUrlException,
 } from 'core/utils';
 import { getLanguage, isValidLang } from 'core/i18n/utils';
 import log from 'core/logger';
@@ -21,16 +22,34 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
   // Get language from URL or fall-back to detecting it from accept-language header.
   const acceptLanguage = req.headers['accept-language'];
   const { lang, isLangFromHeader } = getLanguage({ lang: langFromURL, acceptLanguage });
+  // Get the application from the UA if one wasn't specified in the URL (or
+  // if it turns out to be invalid).
+  const application = getClientApp(req.headers['user-agent']);
 
   const hasValidLang = isValidLang(langFromURL);
+  const hasValidLocaleException = isValidLocaleUrlException(appFromURL, { _config });
   const hasValidClientApp = isValidClientApp(appFromURL, { _config });
-  let hasValidUrlException = isValidUrlException(appFromURL, { _config });
+  let hasValidClientAppUrlException = isValidClientAppUrlException(
+    appFromURL, { _config });
 
-  let prependedLang = false;
+  let isApplicationFromHeader = false;
+  let prependedOrMovedApplication = false;
 
-  if (
+  if (hasValidLocaleException) {
+    log.info(dedent`Second part of URL is a locale exception (${URLParts[1]});
+      make sure the clientApp is valid`);
+
+    // Normally we look for a clientApp in the second part of a URL, but URLs
+    // that match a locale exception don't have a locale so we look for the
+    // clientApp in the first part of the URL.
+    if (!isValidClientApp(langFromURL, { _config })) {
+      URLParts[0] = application;
+      isApplicationFromHeader = true;
+      prependedOrMovedApplication = true;
+    }
+  } else if (
     (hasValidLang && langFromURL !== lang) || hasValidClientApp ||
-    hasValidUrlException
+    hasValidClientAppUrlException
   ) {
     // Replace the first part of the URL if:
     // * It's valid and we've mapped it e.g: pt -> pt-PT.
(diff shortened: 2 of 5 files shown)


config/default-amo.js
```diff
@@ -24,6 +24,21 @@ module.exports = {
   },
   enableTrailingSlashesMiddleware: true,
   fxaConfig: 'amo',
+
+  // This needs to be kept in sync with addons-server's SUPPORTED_NONLOCALES
+  // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L274
+  validRootUrlExceptions: [
```
tofumatt: Relevant bit in `addons-server` is ready to merge once this PR gets merged. https://github.com/mozilla/addons-server/pull/4772

kumar303: I'm confused. Is this supposed to be `validRootExceptions` or is the whitelisted key supposed to be `validRootUrlExceptions`? Either way, they are out of sync.

muffinresearch: Would it be clearer to keep the config keys between the server and frontend more similar? Having the locale in the name makes it a bit clear what this does to me somehow.

kumar303: I was confused because `validRootUrlExceptions` looks like a typo. Is it not? I don't see it initialized in https://github.com/mozilla/addons-frontend/blob/44b614eba8c7692d903c7806fcd183e962da4882/config/default.js

kumar303: Also, the `isValidRootException()` function doesn't check it so it won't be used by anything.

tofumatt: Omg yeah and the way we mock out the tests it still passes. Maybe I changed the keys after I tested it in the browser and messed up. /me fixes...

tofumatt: > Would it be clearer to keep the config keys between the server and frontend more similar? Having the locale in the name makes it a bit clear what this does to me somehow.

Yeah, I actually misunderstood this a bit because during testing I saw some URL redirects and didn't realise one was to the CDN. I need to change the patch a bit.

tofumatt:
I got confused by this because the addons-server code includes both `contribute.json` and `downloads` in their list... but `contribute.json` is redirected to unless it's at the root and `/downloads/` is redirected to `/firefox/downloads/`.

So it seems the code isn't actually accurate on the server and there could be other redirects happening. I'll do my best to replicate what the server does by trial+error 😢

tofumatt:
Working on this and tidying up the middleware in the process as it was getting a bit hard to understand and outgrowing its original code paths/setup. Patch should be coming later tonight as I'm meeting up with friends who are leaving town so I'll be back later.

tofumatt:
Key names were wrong and I misunderstood the patch. Turns out it was for `/firefox/downloads` not `/downloads`

So I tweaked it and added a bit more test coverage. Ready for another r?

I'm sorry this too me forever 😒

tests/client/core/test_middleware.js
```diff
@@ -68,17 +70,51 @@ describe('Prefix Middleware', () => {
     assert.deepEqual(fakeRes.set.firstCall.args, ['vary', []]);
   });
 
-  it('should render a 404 when a URL exception is found', () => {
+  it('should render 404 when a localeURL exception exists', () => {
```
muffinresearch: Presumably this is only the in a case where we don't have a route?

config/default-amo.js
```diff
@@ -24,9 +24,24 @@ module.exports = {
...
+    'blocklist',
+    'static',
+    'user-media',
+    '__version__',
+  ],
+
   // This needs to be kept in sync with addons-server's SUPPORTED_NONAPPS
   // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L262
-  validUrlExceptions: [
+  validClientAppUrlExceptions: [
```
muffinresearch: Nice, this is easier to understand.

tests/client/core/test_middleware.js
tofumatt: Yeah. Well, all of our routes are prepended with `/:lang/:clientApp/` so we don't have routes that look like these.

I guess if we were eventually serving them it might be weird but I don't think we'd do so with the React Router anyway, so I think it's fine.

muffinresearch: My point was the test comment makes it sound like a 404 is tied to the localeURL exception - which might be confusing in the future if we ever needed a route without a locale.

tofumatt: Hmm... does `a localeUrlException triggers a 404` work any better? It clearly ties it to a config value. I could add a comment saying "URLs without a locale will still work as long as they aren't in this config"...

muffinresearch: How about `should render 404 for localeURL exception when route does not exist`?

tofumatt: Oh, no, so it renders a 404 no matter what. The middleware is hit before the routes are accessed and they are in the config–it makes sense to intercept them right away.

If we wanted to serve routes like `/firefox/downloads/foo` but 404 on other things this patch would need to be more complicated.

tofumatt:
@muffinresearch and I chatted about this and agreed that we should stop the middleware from sending out the 404s and just prevent the redirects, but I'll do that in a patch later because we want this to land for release and it doesn't break anything for now.

The middleware shouldn't be sending the 404 though and I'll file a bug to fix it before landing this.

config/default-amo.js
```diff
@@ -24,9 +24,24 @@ module.exports = {
   },
   enableTrailingSlashesMiddleware: true,
   fxaConfig: 'amo',
+
+  // This needs to be kept in sync with addons-server's SUPPORTED_NONLOCALES
```
kumar303: I think it would help to add a sentence about what the map is for -- something simple like "These URLs do not require a locale prefix."

config/default-amo.js
```diff
@@ -24,9 +24,24 @@ module.exports = {
...
+  // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L274
+  validLocaleUrlExceptions: [
+    'contribute.json',
+    'google1f3e37b7351799a5.html',
+    'robots.txt',
+    'services',
+    'downloads',
+    'blocklist',
```
kumar303: Since this isn't in addons-server maybe you should add a comment that it exists to support a redirect to the other host?

src/core/middleware.js
```diff
@@ -21,42 +22,70 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
...
+    prependedOrMovedApplication = true;
   } else if (!hasValidLang) {
     // If lang wasn't valid or was missing prepend one.
     log.info(`Prepending lang to URL: ${lang}`);
     URLParts.splice(0, 0, lang);
     prependedLang = true;
     // If we've prepended the lang to the URL we need to re-check our
     // URL exception and make sure it's valid.
-    hasValidUrlException = isValidUrlException(URLParts[1], { _config });
+    hasValidClientAppUrlException = isValidClientAppUrlException(URLParts[1], { _config });
```
kumar303: I know we don't force 80 char lines but I would be grateful if this was under 80 because I like my split screen :)

src/core/middleware.js
```diff
@@ -21,42 +22,70 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
...
   }
 
-  let isApplicationFromHeader = false;
-
-  if (!hasValidClientApp && prependedLang &&
-      isValidClientApp(URLParts[1], { _config })) {
+  if (
+    !hasValidClientApp && prependedLang &&
```
kumar303: If I remove `prependedLang` from the statement, all tests still pass.

kumar303: If I remove `!hasValidClientApp` from the statement, all tests still pass.

src/core/middleware.js
```diff
@@ -21,42 +22,70 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
...
+    isValidClientApp(URLParts[1], { _config })
+  ) {
     // We skip prepending an app if we'd previously prepended a lang and the
     // 2nd part of the URL is now a valid app.
     log.info('Application in URL is valid following prepending a lang.');
-  } else if (hasValidUrlException) {
-    if (hasValidLang) {
+  } else if (
+    !hasValidClientApp && prependedOrMovedApplication &&
```
kumar303: If I remove `!hasValidClientApp`, the tests still pass

src/core/middleware.js
```diff
@@ -21,42 +22,70 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
...
+    isValidLocaleUrlException(URLParts[1], { _config })
```
kumar303: if I remove this, all tests still pass

tofumatt:
Looking at the test coverage it was actually good (it covers all URL forms we could expect) but I just had extra-wordy conditionals (there was no need to check for one thing when it was checked earlier, etc.). I've tidied them up and added the requested comments 😄

config/default-amo.js
```diff
@@ -25,9 +25,34 @@ module.exports = {
   },
   enableTrailingSlashesMiddleware: true,
   fxaConfig: 'amo',
+
+  // This needs to be kept in sync with addons-server's SUPPORTED_NONLOCALES
+  // settings value: https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L274
+  // These are URLs that are ignored by our prefix middleware that will add
+  // a locale (e.g. `en-US`) to any URL that doesn't have a valid locale.
+  // These are all URLs that should not prepend the locale to the URL if it's
```
kumar303: Thanks for adding extra comments. I think this sentence needs to be reworded though. "These are all URLs that should not get a locale prepended ... " ?

tofumatt:
Cool. I've filed an issue (#2003) to improve it (hopefully reducing complexity) in the future, but this works for launch.

PR #1920, opened 2017-02-27
Encode non-ascii URLs as UTF-8

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1898

The "refresh and it breaks" error strikes again! This time it looks like a bug in [node-fetch](https://github.com/bitinn/node-fetch/issues/245) which Firefox/Chrome do not suffer from. According to the `node-fetch` docs, request bodies are already encoded to UTF-8 so I think all we need is the URL.


src/core/api/index.js
```diff
@@ -86,9 +88,11 @@ export function callApi({
 
       log.warn(oneLine`Response from API was not JSON (was Content-Type:
         ${contentType})`, response);
-      return response.text().then((textResponse) => (
-        { jsonResponse: { text: textResponse }, response }
-      ));
+      return response.text().then((textResponse) => {
+        log.warn(`ACTUAL URL: ${apiURL}`);
```
kumar303: hah, whoops

kumar303: good to know I can fool @tofumatt with debugging code though 😮

tofumatt: Oh shoot I thought that was a log info and you wanted it there, hahaha

PR #1922, opened 2017-02-27
added more system addons

designakt:
more system addons that weren't present on my machine but in other versions/released of Firefox.
Identified via SystemAddons release ftp as described here: https://wiki.mozilla.org/Firefox/Go_Faster/Releasing_an_add-on_mechanics


andymckay:
This is going to get tedious, I wonder if mozilla/addons-server#4695 will help.

muffinresearch:
> This is going to get tedious, I wonder if mozilla/addons-server#4695 will help.

FWIW  there's also a bug open about removing the JSON sent to the disco pane and there's some discussion about retaining the ability to show if user has installed add-ons. 

See https://bugzilla.mozilla.org/show_bug.cgi?id=1340312

PR #1923, opened 2017-02-27
Button for trying on themes

### LINKED ISSUE #1657: After previewing a theme, it remains loaded after a refresh or when navigating to other pages
Preview any theme on Android.

### What happened?
The theme is loaded but after a refresh or if you navigate to other web pages it remains loaded in the web page.

The theme should not be visible, after a refresh of the web page or when you navigate to other web pages.

Verified on FF50.1.0 (Android 7.1.1) on -dev server.
Here is a video:
previewtheme

### LINKED ISSUE #1677: Pressing on the label “Press to Preview” flashes the theme over another theme that is already installed
Install a theme, then press on the label “Press to Preview” from another theme.

### What happened?
After pressing on the label “Press to Preview”, the theme you want to preview flashes over the installed theme.

Pressing on the label “Press to Preview” should not flash the theme you want to preview over the installed theme.

Press&hold will display the theme you want to preview, until you release it.
Verified on FF51.0 (Android 7.1.1) in AMO-dev server.

Here is a video: 
flash

### LINKED ISSUE #1776: Preview themes using a toggle
Currently previewing themes is achieved by pressing and holding on the theme image. This isn't super intuitive and currently has a bug if you tap really quickly.

This interaction should be changed to **Tap to preview** and then switch to **Cancel preview** once the theme is being previewed.

This should resolve #1657 and #1677.

### LINKED ISSUE #1777: Remove background for theme image
Theme images are provided as an `` on the server but that is switched out for a CSS background image once the page loads to avoid prevent the long-press interaction for an image which was preventing scrolling.

Once #1776 lands we should be able to remove this code.

### LINKED ISSUE #1879: Check RTL theme position in a true RTL FF edition
This can be filed under "Sweat the details" I'm not sure if LW theme images are inverted when installed. If they are we should transform the images on the site so the themes are a true reflection of the FF UI.

* [x] Are theme images inverted when installed in a RTL language download of FF?
* ~If the answer to the first question is yes, invert the theme images on amo.~
* [ ] If the answer to the first question is no we should stop inverting the background anchor position for RTL - see comment below for CSS file references.

muffinresearch:
Fixes #1657
Fixes #1677
Fixes #1776
Fixes #1777 
Fixes #1879 

- Themes now previewed on a click. 
- Click again to turn off the theme preview. 
- Preview button goes if the theme is seen as installed + enabled
- Preview goes away when you navigate away (on component unmount)
- Previewing now dispatches in order to provide toggling based on state.

screen shot 2017-02-27 at 20 59 18

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -10,7 +10,7 @@ import Link from 'amo/components/Link';
 import 'amo/css/AddonDetail.scss';
 import fallbackIcon from 'amo/img/icons/default-64.png';
 import InstallButton from 'core/components/InstallButton';
-import { ADDON_TYPE_THEME } from 'core/constants';
+import { ADDON_TYPE_THEME, ENABLED } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import { isAllowedOrigin, ngettext, nl2br, sanitizeHTML } from 'core/utils';
 import translate from 'core/i18n/translate';
@@ -83,7 +83,7 @@ export class AddonDetailBase extends React.Component {
           ref={(el) => { this.wrapper = el; }}
           onClick={this.onClick}
         >
-          {status !== 'ENABLED' ?
+          {status !== ENABLED ?
             <button className="Button AddonDetail-theme-header-label" htmlFor="AddonDetail-theme-header">
               <Icon name="eye" className="AddonDetail-theme-preview-icon" />
               {label}
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -41,69 +41,53 @@ export class AddonDetailBase extends React.Component {
     addon: PropTypes.object.isRequired,
     getBrowserThemeData: PropTypes.func.isRequired,
     i18n: PropTypes.object.isRequired,
+    isPreviewingTheme: PropTypes.bool.isRequired,
     location: PropTypes.object.isRequired,
-    previewTheme: PropTypes.func.isRequired,
-    resetPreviewTheme: PropTypes.func.isRequired,
+    resetThemePreview: PropTypes.func.isRequired,
+    themePreviewNode: PropTypes.element,
+    status: PropTypes.string.isRequired,
+    toggleThemePreview: PropTypes.func.isRequired,
   }
 
   static defaultProps = {
     RatingManager: DefaultRatingManager,
   }
 
-  constructor(props) {
-    super(props);
-    this.state = { mounted: false };
+  componentWillUnmount() {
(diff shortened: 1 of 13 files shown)


muffinresearch:
Test fails are UI related

src/core/css/SearchResult.scss
```diff
@@ -104,10 +104,6 @@
     object-fit: cover;
     object-position: top right;
     width: 100%;
-
-    [dir=rtl] & {
```
muffinresearch: This is removed since for RTL FF the header images are still anchored right so there's no need to change that when the site is being viewed in RTL.

src/core/installAddon.js
```diff
@@ -72,15 +72,32 @@ export function makeProgressHandler(dispatch, guid) {
...
-    previewTheme(node, _themeAction = themeAction) {
-      _themeAction(node, THEME_PREVIEW);
-    },
-    resetPreviewTheme(node, _themeAction = themeAction) {
-      _themeAction(node, THEME_RESET_PREVIEW);
+    toggleThemePreview(node, _themeAction = themeAction, _log = log) {
+      const theme = addon && addon.guid ? addon : null;
+      if (theme && theme.status !== ENABLED) {
+        if (!theme.isPreviewingTheme) {
+          _log.info('Previewing theme');
```
muffinresearch: Yep, that's a good call

src/amo/components/AddonDetail.js
```diff
@@ -41,69 +41,53 @@ export class AddonDetailBase extends React.Component {
...
-          onTouchStart={this.onTouchStart}
-          onTouchEnd={this.onTouchEnd}
           ref={(el) => { this.wrapper = el; }}
+          onClick={this.onClick}
         >
-          <label className="AddonDetail-theme-header-label" htmlFor="AddonDetail-theme-header">
-            <Icon name="eye" className="AddonDetail-theme-preview-icon" />
-            {label}
-          </label>
+          {status !== 'ENABLED' ?
```
muffinresearch: Ah good catch this should be a constant.

src/core/installAddon.js
```diff
@@ -72,15 +72,32 @@ export function makeProgressHandler(dispatch, guid) {
 }
 
 export function mapStateToProps(state, ownProps) {
+  const guid = ownProps.guid || (ownProps.addon && ownProps.addon.guid);
```
muffinresearch: As discussed on vidyo this will resolve to a guid on the addon object if a guid directly on ownProps does not exist. Since this is now used in multiple places it should be factored out to it's own function.

muffinresearch:
Just tested disco theme previews etc with this code via Charles and it's all looking good.

Issue #1886, opened 2017-02-23
Configure SENTRY_DSN on dev/stage/prod


kumar303:
In order to hook up our error reporting (https://github.com/mozilla/addons-frontend/issues/1559) we need the `SENTRY_DSN` environment variable declared in dev/stage/prod for all front-end applications: amo, discopane, and admin. This DSN should be the same one [used by addons-server](https://github.com/mozilla/addons-server/blob/master/src/olympia/conf/dev/settings.py#L228).

jasonthomas:
Shouldn't we use a different SENTRY_DSN  (which means a new project) for each app/env combination? I've set it up that way thinking that would be ideal, but just want to verify.

That being said, I didn't receive any tracebacks when hitting the /simulate-error/ endpoint. I need to look into further why tracebacks are not coming through to sentry.

kumar303:
Sure, whatever makes sense for the dsn. The simulate error view should also
log to the console which may help to figure out if it's working.

jasonthomas:
I need assistance debugging the sentry integration, `raven test SENTRY_DSN` works but /simulate-error/ route from the app are not coming through into sentry [1].

I noticed that `app.use(Raven.requestHandler());` is not configured, do we need that here? I patched -dev but didn't seem to do anything useful [2]

Also [2] mentions that the request and error handler needs to be used before another other middleware, I tried to see if DEBUG=express:* would show anything useful and looks like others might be loaded first? 

```
Feb 27 16:55:30 ip-172-31-39-205 docker[18866]: npm info ok
Feb 27 16:55:35 ip-172-31-39-205 docker[18866]: Mon, 27 Feb 2017 16:55:35 GMT express:application set "x-powered-by" to true
Feb 27 16:55:35 ip-172-31-39-205 docker[18866]: Mon, 27 Feb 2017 16:55:35 GMT express:application set "etag" to 'weak'
Feb 27 16:55:35 ip-172-31-39-205 docker[18866]: Mon, 27 Feb 2017 16:55:35 GMT express:application set "etag fn" to [Function: wetag]
Feb 27 16:55:35 ip-172-31-39-205 docker[18866]: Mon, 27 Feb 2017 16:55:35 GMT express:application set "env" to 'dev'
Feb 27 16:55:35 ip-172-31-39-205 docker[18866]: Mon, 27 Feb 2017 16:55:35 GMT express:application set "query parser" to 'extended'
[...]

PR #1925, opened 2017-02-27
Fix Raven (Sentry) middleware

kumar303:
Jason noticed that this was missing when working on https://github.com/mozilla/addons-frontend/issues/1886


kumar303:
The test failures here are not related to the patch.

PR #1934, opened 2017-02-28
Split error simulation into two URLs: sync and async

kumar303:
This relates to https://github.com/mozilla/addons-frontend/issues/1559

In trying to debug why Sentry won't show our synchronous error (maybe it's https://github.com/getsentry/raven-node/issues/130 ?) I'm splitting the error simulation into two URLs: one for sync errors and one for async errors. This seems like it will be helpful anyway, just to test each kind of error in isolation.


src/core/containers/error-simulation/utils.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable react/prop-types */
+import defaultConfig from 'config';
+import React from 'react';
+
+import NotFound from 'core/components/ErrorPage/NotFound';
+
+export const render404WhenNotAllowed = (Component) => ({
+  config = defaultConfig, ...props
+}) => {
+  if (!config.get('allowErrorSimulation')) {
```
tofumatt: This is a cool idea but I think it should take an argument that gets passed to the `config.get` call; then it's an actual utility function and not specific to this one config call. Because we could use this for the other error page routes only active in development mode.

```jsx
export const render404Unless = (configToCheck) => (Component) => [...]

if (config.get(configToCheck) {
  return <Component />
}

return <NotFound />
```
Type of thing.

kumar303: oh yeah, good idea

src/admin/routes.js
```diff
@@ -1,8 +1,10 @@
-import config from 'config';
 import React from 'react';
 import { IndexRoute, Route } from 'react-router';
 
-import SimulateError from 'core/containers/SimulateError';
+import SimulateAsyncError from
+  'core/containers/error-simulation/SimulateAsyncError';
+import SimulateSyncError from
```
Ryuno-Ki: In Node the variants are differentiated as well. But there the sync one has a longer name (to discourage usage?).

PR #1940, opened 2017-03-01
Disco locales 2017 02 01

muffinresearch:


muffinresearch:
Test fail was because of the nn-NO missing locale.

PR #1955, opened 2017-03-03
Make perPage default for pagination match the API

### LINKED ISSUE #1946: "No results were found" for a search term while navigating to Search Results last pages
Steps to reproduce:
1. Load AMO-dev on your device
2. Make a search for "lang" - https://addons-dev.allizom.org/en-US/android/search/?page=1&q=lang
3. Navigate to page 5

Expected results:
Results are displayed even if page 5 is the last page.

Actual results:
"No results were found for "lang"." message is displayed on page 5.

Notes/Issues:
- reproducing for other search terms with results displayed on multiple pages.
Another example: 
- search term "japan"
- search results in 26 pages
- page 23 is not returning any results

Verified on AMO-dev FF51(Android 6.0.1)
Video for this issue:
videotogif_2017 03 02_11 01 03

muffinresearch:
Fixes #1946

This is a quick fix for #1946, the reason for the empty pages was we were assuming pages of 20 when the API by default returns 25.

The best way to fix this properly would be to have the API expose the perPage value in the results. That way the pagination would dynamically react to the API response and it would prevent needing to match a request page size with the pagination output configuration.

~I'll file an issue for that separately.~ See https://github.com/mozilla/addons-server/issues/4837 for an API issue to update the response to include the number of results per page in the response.

### DIFF
--- a/src/core/components/Paginate/index.js
+++ b/src/core/components/Paginate/index.js
@@ -29,7 +29,7 @@ export class PaginateBase extends React.Component {
 
   static defaultProps = {
     LinkComponent: Link,
-    perPage: 20,
+    perPage: 25, // The default number of results per page returned by the API.
     showPages: 0,
   }
 


eviljeff:
Can't the front-end just handle whatever number of results it receives, and check `response.next_link` isn't `None` to see when the end of the results has been received?

src/core/components/Paginate/index.js
```diff
@@ -29,7 +29,7 @@ export class PaginateBase extends React.Component {
 
   static defaultProps = {
     LinkComponent: Link,
-    perPage: 20,
+    perPage: 25,
```
kumar303: Can you add a comment saying that this is hard coded to the API's default page size?

muffinresearch:
> Can't the front-end just handle whatever number of results it receives, and check response.next_link isn't None to see when the end of the results has been received?

That wouldn't help with counting the number of pages though since they're exposed in the UI from page 1.

src/core/components/Paginate/index.js
muffinresearch: Yep, good call.

kumar303:
> Can't the front-end just handle whatever number of results it receives, and check response.next_link isn't None to see when the end of the results has been received?

I was thinking about that too. We can probably figure it out just by comparing `results.length` to `count` in the API response.

muffinresearch:
> I was thinking about that too. We can probably figure it out just by comparing results.length to count in the API response.

But that would break as soon as you get a remainder e.g. the last page which has only a few results instead of 25.

kumar303:
Yeah, I was thinking you could do some math to handle that case but it would probably be super annoying.

eviljeff:
afaik the api accepts a `page_size` parameter and can return whatever page size the front-end wants.

PR #1957, opened 2017-03-03
Disable install button for themes/add-ons on non-Firefox browsers

### LINKED ISSUE #1881: Fallback install buttons are shown for all UAs even if they can't install add-ons

Visit addons-dev with chrome on ios or chrome android or FF ios.

### What happened?

Since there's no mozAddonManager there's a fallback to install buttons which won't do anything when pressed since the UA doesn't work with FF extensions.

A notice to show that the UA is not supported with a link to where to download FF from if relevant.
The buttons should also be disabled if they can't be used.

The notice will need UX input. This is something that can come as a post-release update.

kumar303:
Fixes #1881 

This is a partial fix for the problem. All it does is dim and disable the install button for add-ons/themes (including the theme preview button) in browsers like Chrome where we know for sure add-ons aren't supported. It also disables buttons in Firefox iOS but I can be persuaded not to do that if someone has a different idea.

Screenshots from Chrome desktop:

screenshot 2017-03-06 13 14 29
screenshot 2017-03-06 13 14 53

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -72,12 +72,6 @@
     "jsx-a11y/no-static-element-interactions": "off",
     "no-console": "error",
     "no-plusplus": "off",
-    "no-restricted-syntax": [
-      'error',
-      'ForInStatement',
-      'LabeledStatement',
-      'WithStatement',
-    ],
     "no-underscore-dangle": "off",
     "space-before-function-paren": ["error", "never"],
     // The airbnb default of this rule mainly encourages `shape` over `object`
--- a/.eslintrc
+++ b/.eslintrc
@@ -72,6 +72,12 @@
     "jsx-a11y/no-static-element-interactions": "off",
     "no-console": "error",
     "no-plusplus": "off",
+    "no-restricted-syntax": [
+      'error',
+      'ForInStatement',
+      'LabeledStatement',
+      'WithStatement',
+    ],
     "no-underscore-dangle": "off",
     "space-before-function-paren": ["error", "never"],
     // The airbnb default of this rule mainly encourages `shape` over `object`
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -89,7 +89,6 @@ export class AddonDetailBase extends React.Component {
       const label = isPreviewingTheme ? i18n.gettext('Cancel preview') : i18n.gettext('Tap to preview');
       const imageClassName = 'AddonDetail-theme-header-image';
       const headerImage = <img alt={label} className={imageClassName} src={previewURL} />;
-      const buttonIsDisabled = !clientSupportsAddons();
 
       return (
         <div
@@ -101,7 +100,7 @@ export class AddonDetailBase extends React.Component {
         >
           {status !== ENABLED ?
             <button
-              disabled={buttonIsDisabled}
+              disabled={!clientSupportsAddons()}
               className="Button AddonDetail-theme-header-label"
               htmlFor="AddonDetail-theme-header">
               <Icon name="eye" className="AddonDetail-theme-preview-icon" />
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -12,7 +12,13 @@ import fallbackIcon from 'amo/img/icons/default-64.png';
 import InstallButton from 'core/components/InstallButton';
 import { ADDON_TYPE_THEME, ENABLED } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
-import { isAllowedOrigin, ngettext, nl2br, sanitizeHTML } from 'core/utils';
+import {
+  clientSupportsAddons as _clientSupportsAddons,
+  isAllowedOrigin,
+  ngettext,
+  nl2br,
+  sanitizeHTML,
+} from 'core/utils';
 import translate from 'core/i18n/translate';
 import Card from 'ui/components/Card';
 import Icon from 'ui/components/Icon';
@@ -39,6 +45,7 @@ export class AddonDetailBase extends React.Component {
   static propTypes = {
     RatingManager: PropTypes.element,
     addon: PropTypes.object.isRequired,
+    clientSupportsAddons: PropTypes.func,
     getBrowserThemeData: PropTypes.func.isRequired,
     i18n: PropTypes.object.isRequired,
     isPreviewingTheme: PropTypes.bool.isRequired,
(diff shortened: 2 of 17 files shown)


kumar303:
whoops, hold up. I forgot to disable the tap to preview button for themes.

src/core/css/inc/mixins.scss
```diff
@@ -58,6 +58,18 @@
   outline: none;
 }
 
+@mixin disabled($background) {
+  @if lightness($background) > 80 {
+    background: adjust-hue(darken(desaturate($background, 85%), 13%), 6);
+  }
+
```
kumar303: eslint wanted me to add this empty line but I don't know why

src/amo/components/AddonDetail.js
```diff
@@ -74,6 +89,7 @@ export class AddonDetailBase extends React.Component {
       const label = isPreviewingTheme ? i18n.gettext('Cancel preview') : i18n.gettext('Tap to preview');
       const imageClassName = 'AddonDetail-theme-header-image';
       const headerImage = <img alt={label} className={imageClassName} src={previewURL} />;
+      const buttonIsDisabled = !clientSupportsAddons();
```
tofumatt: Reading `button is disabled if client does not support add-ons` reads like a double negative and was harder for my brain to grok. I guess `buttonEnabled` or `installationEnabled` here, then `disabled={!buttonEnabled}` is just as confusing and double-negative though. `disabled` is an annoying attribute 🤷‍♀️

src/core/components/InstallButton/index.js
```diff
@@ -22,29 +24,49 @@ export class InstallButtonBase extends React.Component {
...
+      className,
+      hasAddonManager,
+      i18n,
+      size,
+    } = this.props;
     const useButton = hasAddonManager !== undefined && !hasAddonManager;
     let button;
+    const buttonIsDisabled = !clientSupportsAddons();
+    const buttonClass = classNames('InstallButton-button', {
+      'InstallButton-button--disabled': buttonIsDisabled,
```
tofumatt: Hmm... I know we use classes everywhere but I feel like `class="button-disabled"` is the butt of many jokes about CSS frameworks. But it's consistent with our other stuff and I like that we keep our targets from a lot of cascading, so I think this is fine.

src/core/components/InstallButton/index.js
```diff
@@ -22,29 +24,49 @@ export class InstallButtonBase extends React.Component {
...
-          className="InstallButton-button">
+          className={buttonClass}>
           {i18n.gettext('Install Theme')}
         </Button>
       );
     } else {
       button = (
-        <Button href={addon.installURL} size={size} className="InstallButton-button">
+        <Button
+          href={buttonIsDisabled ? null : addon.installURL}
```
tofumatt: Maybe we should file a bug to show a link to the add-on XPI directly if the install button is disabled. Could be handy for debugging but also as a failsafe if our detection went wrong or something.

src/core/css/inc/mixins.scss
tofumatt: Yeah that's weird. 😒

src/core/css/inc/mixins.scss
```diff
@@ -58,6 +58,18 @@
   outline: none;
 }
 
+@mixin disabled($background) {
```
tofumatt: I like this, btw. 👍

.eslintrc
```diff
@@ -72,6 +72,12 @@
     "jsx-a11y/no-static-element-interactions": "off",
     "no-console": "error",
     "no-plusplus": "off",
+    "no-restricted-syntax": [
+      'error',
+      'ForInStatement',
```
tofumatt: This was here so we'd explicitly go through and find instances where we were using `for... in` loops and remove them eventually. I know it's annoying to write it all the time but I thought we'd agree it would be an eventual code quality thing so writing them out would make it more obvious we don't encourage it and easy to grok for when replacing.

https://github.com/mozilla/addons-frontend/pull/1379#issuecomment-261549708

If we want to just say "okay we'll use `for... in`" I guess that's fine but it seems at odds with our general non-mutating stuff style. I'd be happy to stop using it eventually 😄

tests/client/helpers.js
```diff
@@ -114,3 +114,59 @@ export const signedInApiState = Object.freeze({
   lang: 'en-US',
   token: 'secret-token',
 });
+
+export const userAgents = {
```
tofumatt: Handy 👍

tests/client/core/test_utils.js
```diff
@@ -174,61 +123,101 @@ describe('getClientApplication', () => {
...
+    }
+  });
+
+  it('returns false for Chrome desktop', () => {
+    for (const ua of userAgents.chromeAndroid) {
+      assert.strictEqual(clientSupportsAddons(ua), false, `UA string: ${ua}`);
+    }
+  });
+
+  it('returns true for Firefox desktop', () => {
```
tofumatt: I don't see a Firefox Android test here. Did I just miss it?

src/core/components/InstallButton/index.js
kumar303: It's necessary because one implementation uses a `<button>` tag and the other uses an `<a>` tag.

tofumatt: Ah, makes sense. Cool with it anyway, but that makes it extra clear 😄 👍

src/core/css/inc/mixins.scss
mstriemer: Maybe it should be nested with the close paren? I see it used to be like this:

```scss
@if ... {
} @else {
}

src/core/components/InstallButton/index.js
kumar303: Oh, good call. I'll put the href back. Actually, you made me realize that we want to expose XPI links to search crawlers which will probably trigger "not supported" rules

tests/client/core/test_utils.js
kumar303: Good catch, thanks

.eslintrc
kumar303: Oh, huh, thanks for linking to the comment. I wish there was a way to make this more obvious to the next person who accidentally removes them all but I got nothing. I put the ignore comments back.

tests/client/core/test_utils.js
kumar303: no, it needs one, thanks

src/core/css/inc/mixins.scss
kumar303: That's exactly how it was before I moved it into a function. It triggers a lint error now that it's in a function. I see other instances inside functions that do not trigger an error. Shrug. It must be a bug in stylelint.

PR #1959, opened 2017-03-03
Hack around bug in PhotoSwipe

### LINKED ISSUE #1712: Screenshots preview is not working for second add-on
Steps to reproduce:
1. Open AMO-dev on your Android device
2. Load details page for an add-on
3. Navigate through screenshots in full mode preview
4. Go to another add-on details and try to navigate to its screenshots in full mode preview

Expected results:
User can navigate through screenshots in full mode preview with not problems.

Actual results:
Screenshots section turns into with for the second add-on. Screenshot cannot longer be previewed.

Notes/Issues:
Verified on AMO-dev FF51(Android 6.0.1).
Video for this issue: 
noscrrn

muffinresearch:
Fixes #1712 

Original function is here: https://github.com/minhtranite/react-photoswipe/blob/05602e52755cc69d1f530a64a65cc5b6b0b225b9/src/PhotoSwipeGallery.js#L36

This is a bit of a hack but it overloads the function used to get thumbnails so that the second time photoswipe is loaded it continues to work.

This will need coverage (or skips to coverage) but if you have a moment it would be spin this up in review and check it solves the problem.

### DIFF
--- a/src/amo/components/ScreenShots.js
+++ b/src/amo/components/ScreenShots.js
@@ -18,7 +18,9 @@ const PHOTO_SWIPE_OPTIONS = {
   counterEl: true,
   arrowEl: true,
   preloaderEl: true,
-  getThumbBoundsFn: function getThumbBoundsFn(index) {
+  // Overload getThumbsBoundsFn as workaround to
+  // https://github.com/minhtranite/react-photoswipe/issues/23
+  getThumbBoundsFn: /* istanbul ignore next */ function getThumbBoundsFn(index) {
     const thumbnail = document.querySelectorAll('.pswp-thumbnails')[index];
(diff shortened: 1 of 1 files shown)


Ryuno-Ki:
Have  you filled a ticket against upstream?

muffinresearch:
> Have you filled a ticket against upstream?

No, because one already exists.

Ryuno-Ki:
Could you link it here, so others could take a look for when it gets resolved, too?

Or better, open a ticket that the workaround gets removed once the fix lands upstream.

tofumatt:
Note as well: that repo is quite old and seemingly unmaintained, given some open pull requests from last year with no author/maintainer comments.

Long term we'll likely want to fork it or something. But this is good for next week 😄

Issue #1933, opened 2017-02-28
Send client side errors to Sentry


kumar303:
When an error happens on the client, let's [send it to Sentry](https://sentry.io/for/javascript/). Sentry for server side errors was hooked up in https://github.com/mozilla/addons-frontend/issues/1559

muffinresearch:
@kumar303 does this still make sense to keep in as a release blocker or would we be ok to patch this post release?

kumar303:
I think it would be nice to block release on this because otherwise we'll be flying blind. However, I'm trying to work out the server errors first since they seem more important. In any case, if we had to release without client error reporting, nothing will *break* in the philosophical sense because we won't know what's broken.

kumar303:
This is broken because CSP is blocking the Sentry API. Oops.

PR #1978, opened 2017-03-06
Fix CSP rules for Sentry client reporting

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1933

### DIFF
--- a/config/default-admin.js
+++ b/config/default-admin.js
@@ -1,4 +1,5 @@
-const amoCDN = 'https://addons.cdn.mozilla.net';
+import { amoProdCDN } from './lib/shared';
+
 const staticHost = 'https://addons-admin.cdn.mozilla.net';
 
 module.exports = {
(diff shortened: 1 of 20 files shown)


config/default-admin.js
```diff
@@ -13,7 +14,7 @@ module.exports = {
       styleSrc: [staticHost],
       imgSrc: [
         "'self'",
-        amoCDN,
+        amoProdCDN,
```
muffinresearch: I like the name change - it's clearer.

config/default-admin.js
```diff
@@ -1,4 +1,5 @@
-const amoCDN = 'https://addons.cdn.mozilla.net';
+import { amoProdCDN } from './lib/shared';
+
 const staticHost = 'https://addons-admin.cdn.mozilla.net';
 
 module.exports = {
```
muffinresearch: It's a bit weird being module.exports and using es6 imports.

kumar303: Hmm, I don't think there's an easy way to convert all of these to es6 without turning each object key into `export const ...`

kumar303:
Thanks. Let's try it on dev...

PR #1980, opened 2017-03-07
Disable incompat versions and browsers

### LINKED ISSUE #1935: Disable buttons for incompatible FF clients
If the version of the add-on is not supported by the FF client we should disable the button as a minimum. We should also add a basic message since the switches don't support copy by themselves and I don't think it's great to rely on a tooltip for that.

Note this change will also apply to the fallback buttons in the case where mozAddonManager is not available *and* the Add-on is incompatible with that version.

tofumatt:
Fixes #1935.

This prevents non-Firefox browsers and incompatible versions of Firefox (based on an addon's `current_version.compatibility[clientApp].max` and `current_version.compatibility[clientApp].min` values) from showing the install buttons and theme preview toggles as enabled.

I need to tidy up a lot of tests as they introduced state into some components and I also need to add tests, but I'm gonna do it in the morning when I get up 😄 

(This works but is quite rough now so no sense doing prelim review 😉)

### Screenshots

screen shot 2017-03-07 at 04 01 31
screen shot 2017-03-07 at 04 01 14
screen shot 2017-03-07 at 04 00 54
screen shot 2017-03-07 at 04 00 43


src/core/utils.js
```diff
@@ -305,3 +307,44 @@ export function render404IfConfigKeyIsFalse(
...
+  if (browser.name === 'Firefox' && os.name !== 'iOS') {
+    // Do version checks, if this add-on has minimum or maximum version
+    // requirements.
+    // The mozilla-version-comparator API is quite strange; a result of
+    // `1` means the first argument is higher in version than the second.
+    if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
+      return false;
+    }
+
+    // A result of `-1` means the second argument is a lower version than the
```
muffinresearch: Thanks for adding these comments!

tofumatt:
Ready for r?

tests/client/amo/components/TestAddonCompatibility.js
```diff
@@ -0,0 +1,109 @@
...
+    const root = render({
+      isCompatibleWithUserAgent: () => false,
+      userAgent: userAgents.chrome[0],
+    });
+    const rootNode = findDOMNode(root);
+
+    // This is localised to the current locale, which is en-US in our
+    // signedInApiState helper.
+    assert.equal(rootNode.querySelector('a').href,
+      'https://www.mozilla.org/en-US/firefox/');
```
kumar303: to avoid mistakes, I prefer to see these kind of tests done with a non-default lang. You could alter `render()` to accept an override to the `signedInApiState` or something like that.

tests/client/amo/components/TestAddonCompatibility.js
```diff
@@ -0,0 +1,109 @@
...
+      minVersion: '11.0',
+      // This is Firefox 10 on Linux.
+      userAgent: userAgents.firefox[0],
+    });
+    const rootNode = findDOMNode(root);
+
+    // This is localised to the current locale, which is en-US in our
+    // signedInApiState helper.
+    assert.equal(rootNode.querySelector('a').href,
+      'https://www.mozilla.org/en-US/firefox/');
```
kumar303: Same lang comment applies

tests/client/amo/components/TestAddonCompatibility.js
```diff
@@ -0,0 +1,109 @@
...
+    assert.include(rootNode.textContent,
+      'You need to download Firefox to install this add-on.');
+  });
+
+  it('renders a notice for old versions of Firefox', () => {
+    const root = render({
+      minVersion: '11.0',
+      // This is Firefox 10 on Linux.
+      userAgent: userAgents.firefox[0],
```
kumar303: When you move this to `userAgentData` (parsed) then this test will be easier to read. Right now it's hard to know what the version comparison is because you have to look up the user agent string.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -21,24 +22,23 @@ import createStore from 'amo/store';
...
-    clientSupportsAddons: () => true,
+    isCompatibleWithUserAgent: () => true,
     getBrowserThemeData: () => '{}',
     i18n,
     location: { pathname: '/addon/detail/' },
     // Configure AddonDetail with a non-redux depdendent RatingManager.
     RatingManager: RatingManagerWithI18n,
     setCurrentStatus,
-    store: createStore(initialState),
+    store: createStore(signedInApiState),
```
kumar303: This should be `store: createStore({ api: signedInApiState })`

tests/client/core/test_utils.js
```diff
@@ -713,3 +727,104 @@ describe('render404IfConfigKeyIsFalse', () => {
...
+      addon: null, clientApp: 'firefox' });
+
+    assert.equal(maxVersion, null);
+    assert.equal(minVersion, null);
+  });
+});
+
+describe('isCompatibleWithUserAgent', () => {
+  it('should mark non-Firefox UAs as incompatible', () => {
+    isCompatibleWithUserAgent({ userAgent: userAgents.chrome[0] });
```
kumar303: I don't see any assertions. What is this test doing? It also looks like a dupe of https://github.com/mozilla/addons-frontend/blob/43b3dfd55121535d1acf249153b02161e214225f/tests/client/core/test_utils.js#L193

tests/client/core/test_utils.js
```diff
@@ -713,3 +727,104 @@ describe('render404IfConfigKeyIsFalse', () => {
...
+  });
+
+  it('should mark Firefox for iOS as incompatible', () => {
+    isCompatibleWithUserAgent({ userAgent: userAgents.firefoxIOS[0] });
```
kumar303: I don't see any assertions. What is this test doing? It also looks like a dupe of https://github.com/mozilla/addons-frontend/blob/43b3dfd55121535d1acf249153b02161e214225f/tests/client/core/test_utils.js#L225

tests/client/core/test_utils.js
```diff
@@ -713,3 +727,104 @@ describe('render404IfConfigKeyIsFalse', () => {
...
+  });
+
+  it('should mark Firefox 10 as incompatible with a minVersion of 10.1', () => {
+    isCompatibleWithUserAgent({
```
kumar303: I don't see any assertions here either but this test and the ones below it should move into the [existing suite](https://github.com/mozilla/addons-frontend/blob/43b3dfd55121535d1acf249153b02161e214225f/tests/client/core/test_utils.js#L176).

src/amo/components/AddonDetail.js
```diff
@@ -198,6 +217,7 @@ export class AddonDetailBase extends React.Component {
           </h2>
           <AddonMeta addon={addon} />
           <InstallButton {...this.props} />
+          <AddonCompatibility maxVersion={maxVersion} minVersion={minVersion} />
```
tofumatt: Yeah that's a solid idea. I rushed through bits of this and found the double usage of `isCompatibleWithUserAgent()` a code smell. That idea makes sense to me!

src/core/utils.js
```diff
@@ -305,3 +293,53 @@ export function render404IfConfigKeyIsFalse(
...
+  }
+
+  return { maxVersion, minVersion };
+}
+
+export function isCompatibleWithUserAgent({
+  userAgent, maxVersion, minVersion,
+} = {}) {
+  // If the userAgent is false-y, we'll assume it isn't valid.
+  if (!userAgent) {
```
tofumatt: Oh, good call.

src/core/utils.js
```diff
@@ -305,3 +293,53 @@ export function render404IfConfigKeyIsFalse(
...
+  const { browser, os } = UAParser(userAgent);
+
+  // We need a Firefox browser compatible with add-ons (Firefox for iOS does
+  // not currently support add-ons).
+  if (browser.name === 'Firefox' && os.name !== 'iOS') {
+    // Do version checks, if this add-on has minimum or maximum version
+    // requirements.
+    // The mozilla-version-comparator API is quite strange; a result of
+    // `1` means the first argument is higher in version than the second.
+    if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
```
tofumatt: Oh, thanks, I'll check that out. The whole API is kinda weird... arguably _we_ should just ignore `*` but if it does (their docs are basically the tests 😒) then I'll just add a test for us and pass it through to the lib.

src/core/reducers/api.js
```diff
@@ -13,6 +14,8 @@ export default function api(state = {}, action) {
       return { ...state, lang: action.payload.lang };
     case SET_CLIENT_APP:
       return { ...state, clientApp: action.payload.clientApp };
+    case SET_USER_AGENT:
+      return { ...state, userAgent: action.payload.userAgent };
```
tofumatt: Ah, fair enough.

src/core/utils.js
tofumatt: Just tested an `maxVersion` works but a `minVersion` of `'*'` is marked as higher than any version.

That should not happen and a `minVersion` of `'*'` kinda makes no sense, but what should we do with that? I'm fine marking a `minVersion` of `'*'` as compatible, but arguably it's not valid at all...

src/core/utils.js
```diff
@@ -305,3 +293,53 @@ export function render404IfConfigKeyIsFalse(
...
+      return false;
+    }
+
+    // If we made it here we're compatible (yay!)
+    return true;
+  }
+
+  // This means the client is Firefox for iOS or not Firefox, so we mark
+  // the user agent as incompatible.
+  return false;
```
tofumatt: To do that I'd need to pass in `i18n` to the function because we're in `utils`.

But I guess we could have error codes that we pass back to the component, I'll try that.

src/core/utils.js
kumar303: I van verify that `maxVersion=*` will happen automatically for all WebExtensions, even if they don't specify anything in their manifest. I'm trying to find out about `minVersion`. To be safe, we should probably support `minVersion=*` in our code.

kumar303: To clarify, what I can confirm is that WebExtensions will automatically get `minVersion=48` when they don't specify anything in their manifest.

src/core/utils.js
kumar303: Passing `i18n` to the function doesn't seem so bad. I think returning error codes seems ok too but it would be more work for you since you'd have more tests to write.

src/core/utils.js
tofumatt: Cool. I just tested it locally and `maxVersion = '*'` totally works.

tofumatt: I'll make minVersion work but will log an error about it.

tofumatt:
Okay! All set for another r? This got a bit bigger but I think it's must more robust now, and includes handling the Install Switch which wasn't happening before I think.

kumar303:
Looks like maybe some lang related test failures

tofumatt:
I think it's because of navigator being undefined in the server tests. Should be an easy fix but I lazily walked away from my computer last night thinking the tests would pass.

- Matthew Riley MacPherson (sent from mobile)

tofumatt:
Okay, tests fixed, sorry about that!

tests/client/amo/components/TestAddonCompatibilityError.js
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  it('renders a notice for non-Firefox browsers', () => {
+    const root = render({
+      lang: 'es',
+      userAgentInfo: { browser: { name: 'Chrome' }, os: {} } });
+    const rootNode = findDOMNode(root);
+
+    // This is localised to the current locale, which is en-US in our
+    // signedInApiState helper.
```
kumar303: You can delete this comment. And, yeah, I like the explicit approach of passing in `lang` better.

tests/client/amo/components/TestAddonCompatibilityError.js
```diff
@@ -0,0 +1,114 @@
...
+  it('renders a notice for old versions of Firefox', () => {
+    const root = render({
+      lang: 'fr',
+      minVersion: '11.0',
+      userAgentInfo: { browser: { name: 'Firefox', version: '8.0' }, os: {} },
+    });
+    const rootNode = findDOMNode(root);
+
+    // This is localised to the current locale, which is en-US in our
+    // signedInApiState helper.
```
kumar303: This comment can also be deleted.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -291,19 +296,31 @@ describe('AddonDetail', () => {
...
-      clientSupportsAddons: () => true,
+      isCompatibleWithUserAgent: () => ({ compatible: true }),
     });
     const button = rootNode.querySelector('.AddonDetail-theme-header-label');
     assert.equal(button.disabled, false);
   });
 
+  it('disables install switch for unsupported clients', () => {
+    const rootNode = renderAsDOMNode({
+      addon: {
```
kumar303: `addon` does not seem relevant to this test so you can delete it and rely on the default

tests/client/amo/containers/TestApp.js
```diff
@@ -141,6 +143,19 @@ describe('App', () => {
...
   });
 
+  it('sets up a callback for setting add-on status', () => {
+    const dispatch = sinon.spy();
+    const { setUserAgent } = mapDispatchToProps(dispatch);
+    const userAgent = 'tofubrowser';
+    const { browser, os } = UAParser(userAgent);
+
+    setUserAgent(userAgent);
+    assert.ok(dispatch.calledWith({
```
kumar303: I think it would be better to stub out `setUserAgentAction` and just assert that `dispatch` gets called with the result of your stub `setUserAgentAction`. This will allow the `setUserAgentAction` function to evolve independently of the `App` component.

tests/client/core/actions/test_index.js
```diff
@@ -28,6 +32,28 @@ describe('core actions setClientApp', () => {
...
+
+  it('allow an empty userAgent value', () => {
+    const userAgent = '';
+    const { browser, os } = UAParser(userAgent);
+
+    assert.deepEqual(actions.setUserAgent(''), {
+      type: 'SET_USER_AGENT',
+      payload: { userAgent, userAgentInfo: { browser, os } },
+    });
+  });
```
kumar303: The `UAParser()` docs are not clear on how it handles garbage UA strings. Can you add a test for `UAParser('^^**garbage__')` or something like that?

tests/client/core/test_utils.js
```diff
@@ -172,56 +180,150 @@ describe('getClientApp', () => {
...
+      os: { name: 'Windows' },
+    };
+    assert.deepEqual(isCompatibleWithUserAgent({
+      maxVersion: '*', userAgentInfo }),
+      { compatible: true });
+  });
+
+  it('should log warning when minVersion is "*"', () => {
+    // Note that this should never happen as addons-server will mark a
+    // WebExtension with no minVersion as having a minVersion of "48".
```
kumar303: It might still happen for non-webextension add-ons or for webextensions that specify `*` explicitly in their manifest but I don't know

tofumatt: Yeah I think that should be technically invalid. The API says that a lack of a minimum version will mark `min` as `null`. Given the comparator library doesn't address minimum versions of `'*'` I think that's a sign to say they aren't valid.

tests/client/core/actions/test_index.js
tofumatt: I do have a test for empty strings and a test that uses `'tofubrowser'` which is not valid. But it's in `App.js` and aims to test something else so I'll add one here as well.

tests/client/amo/components/TestAddonDetail.js
tofumatt: Copypasta strikes again. Thanks.

src/core/components/InstallButton/index.js
```diff
@@ -37,15 +43,25 @@ export class InstallButtonBase extends React.Component {
...
       size,
+      userAgentInfo,
     } = this.props;
     const useButton = hasAddonManager !== undefined && !hasAddonManager;
     let button;
-    const buttonIsDisabled = !clientSupportsAddons();
+
+    // Test add-on capability (is the client Firefox?) and compatibility
+    // (is the client a valid version of Firefox to run this add-on?).
+    const { maxVersion, minVersion } = getCompatibleVersions({
```
tofumatt: Totally true. Good call; I like it.

Like you said I'll keep the functions separate in the module, but I'll add `getClientCompatibility`. 👍

src/core/constants.js
```diff
@@ -68,6 +68,12 @@ export const VISIBLE_ADDON_TYPES_MAPPING = Object.keys(API_ADDON_TYPES_MAPPING)
     [API_ADDON_TYPES_MAPPING[key]]: key,
   }), {});
 
+// Incompatibility codes for
```
tofumatt: Hah, I meant to write something and clearly got distracted. 😖

src/amo/components/AddonDetail.js
```diff
@@ -100,7 +118,7 @@ export class AddonDetailBase extends React.Component {
         >
           {status !== ENABLED ?
             <button
-              disabled={!clientSupportsAddons()}
+              disabled={!this.isCompatible()}
```
tofumatt: Mostly I dislike passing around data like that when there's already a method to do it. I don't think `isCompatible()` is a bad thing to call multiple times. It's a cleaner API to me, having to pass less stuff around.

src/amo/containers/App.js
```diff
@@ -43,10 +47,28 @@ export class AppBase extends React.Component {
...
+      handleGlobalEvent,
+      mozAddonManager,
+      setUserAgent,
+      userAgent,
+    } = this.props;
+
+    // If userAgent isn't set in state it could be we couldn't get one from
+    // the request headers on our first (server) request. If that's the case
+    // we try to load them from navigator.
+    const _navigator = navigator || window.navigator;
```
tofumatt: I did that and then removed it because I think elsewhere we didn't have a test for navigator? But yeah good point, I'll add that back and run a test.

src/core/actions/index.js
```diff
@@ -35,6 +38,15 @@ export function setLang(lang) {
   };
 }
 
+export function setUserAgent(userAgent) {
+  const { browser, os } = UAParser(userAgent);
```
tofumatt: It's documented in code examples in the README:

screenshot 2017-03-10 01 25 40

tofumatt:
```js
    var result = parser.getResult();
    // this will also produce the same result (without instantiation):
    // var result = UAParser(uastring);

    console.log(result.browser);        // {name: "Chromium", version: "15.0.874.106"}
    console.log(result.device);         // {model: undefined, type: undefined, vendor: undefined}
    console.log(result.os);             // {name: "Ubuntu", version: "11.10"}
    console.log(result.os.version);     // "11.10"
    console.log(result.engine.name);    // "WebKit"
    console.log(result.cpu.architecture);   // "amd64"
```

It breaking would be somewhat easy to diagnose as it'd be in a patch that upped the version and we could see where/how we're using `UAParser`. I think it'd be alright, and the syntax I'm using is more concise without giving up anything–the `getBrowser()` code isn't really any better to read.

src/core/reducers/api.js
```diff
@@ -13,6 +14,12 @@ export default function api(state = {}, action) {
       return { ...state, lang: action.payload.lang };
     case SET_CLIENT_APP:
       return { ...state, clientApp: action.payload.clientApp };
+    case SET_USER_AGENT:
+      return {
+        ...state,
+        userAgent: action.payload.userAgent,
+        userAgentInfo: action.payload.userAgentInfo,
```
tofumatt: I saw your tweet about this and totally realised I did it, heh. Makes sense 👍

src/amo/components/AddonDetail.js
tofumatt: Actually I changed my mind and did this.

tofumatt:
Okay, all comments addressed, thanks a lot for the thorough re-re-views. Definitely feel like this patch is a lot nicer than when it started 😄

kumar303:
I'm taking another look right now. I think you can fix the servertest failures by doing `typeof navigator !== 'undefined'`

tofumatt:
Oh weird, the server tests didn't fail locally. I don't think. Maybe they did. You're right, I'll fix that...

tests/client/amo/components/TestAddonCompatibilityError.js
```diff
@@ -0,0 +1,106 @@
...
+import { signedInApiState } from 'tests/client/amo/helpers';
+import { getFakeI18nInst } from 'tests/client/helpers';
+import I18nProvider from 'core/i18n/Provider';
+
+
+describe('AddonCompatibilityError', () => {
+  function render({ ...props }) {
+    const api = {
+      ...signedInApiState,
+      lang: props.lang || signedInApiState.lang,
```
kumar303: This can simply be `lang: props.lang` because the object spread will already fall back to `signedInApiState.lang`

tests/client/amo/components/TestAddonCompatibilityError.js
```diff
@@ -0,0 +1,106 @@
...
+  INCOMPATIBLE_OVER_MAX_VERSION,
+  INCOMPATIBLE_UNDER_MIN_VERSION,
+} from 'core/constants';
+import { signedInApiState } from 'tests/client/amo/helpers';
+import { getFakeI18nInst } from 'tests/client/helpers';
+import I18nProvider from 'core/i18n/Provider';
+
+
+describe('AddonCompatibilityError', () => {
+  function render({ ...props }) {
```
kumar303: Once you add all the undefined checks, you'll need to set up some default props:
````js
function render({ ...customProps } = {}) {
  const props = {
    minVersion: null,
    maxVersion: null,
    ...customProps,
  };
}
````

File: tests/client/amo/components/TestAddonDetail.js [hunk 28]
```diff
@@ -452,3 +465,13 @@ describe('AddonDetail', () => {
     });
   });
 });
```
kumar303: Since you missed passing some props to `AddonCompatibilityError` I think it would be good to add a test for that. You could use a stub component and make assertions about what props it received.

tests/client/amo/containers/TestApp.js
```diff
@@ -141,6 +146,16 @@ describe('App', () => {
     assert.ok(dispatch.calledWith({ type: INSTALL_STATE, payload }));
   });
 
+  it('sets up a callback for setting add-on status', () => {
```
kumar303: I think this test spec is wrong

tests/client/amo/containers/TestApp.js
```diff
@@ -155,6 +170,21 @@ describe('App', () => {
     assert.equal(lang, 'de');
   });
 
+  it('sets the userAgent as props', () => {
+    const store = createStore();
+    store.dispatch(setUserAgentAction('tofubrowser'));
```
kumar303: 👍

tests/client/core/actions/test_index.js
```diff
@@ -28,6 +30,26 @@ describe('core actions setClientApp', () => {
...
+    const userAgent = userAgents.chrome[0];
+
+    assert.deepEqual(actions.setUserAgent(userAgent), {
+      type: 'SET_USER_AGENT',
+      payload: { userAgent },
+    });
+  });
+
+  it('allow an empty userAgent value', () => {
+    const userAgent = '';
```
kumar303: what about another test that uses `userAgent = undefined` ?

tests/client/core/reducers/test_api.js
```diff
@@ -34,6 +38,26 @@ describe('api reducer', () => {
...
+    const { browser, os } = UAParser(userAgent);
+
+    assert.deepEqual(
+      api(existingState, actions.setUserAgent(userAgent)),
+      { ...existingState, userAgent, userAgentInfo: { browser, os } });
+  });
+
+  it('allows empty userAgent', () => {
+    const existingState = { bar: 'baz' };
+    const userAgent = '';
```
kumar303: I think it would be good to add another test like this one that tries to pass an `undefined` `userAgent`

tests/client/core/reducers/test_api.js
```diff
@@ -34,6 +38,26 @@ describe('api reducer', () => {
...
+    const { browser, os } = UAParser(userAgent);
+
+    assert.deepEqual(
+      api(existingState, actions.setUserAgent(userAgent)),
+      { ...existingState, userAgent, userAgentInfo: { browser, os } });
+  });
```
kumar303: I'd like to see a reducer test that passes in a garbage `userAgent` value. I know you have one for the component but that's not going to touch reducer code which does the actual parsing.

tests/client/core/test_utils.js
```diff
@@ -172,56 +180,169 @@ describe('getClientApp', () => {
...
+    assert.deepEqual(
+      isCompatibleWithUserAgent({ minVersion: '9.0', userAgentInfo }),
+      { compatible: false, reason: INCOMPATIBLE_FIREFOX_FOR_IOS }
+    );
+  });
+
+  it('is incompatible with garbage user agent', () => {
+    const userAgent = '^^**garbage__';
+    assert.deepEqual(
+      isCompatibleWithUserAgent({ userAgentInfo: UAParser(userAgent) }),
```
kumar303: I think it would be more useful to put this test in the reducer since this is really just an exercise for the parsing code.

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+  render() {
+    const {
+      i18n,
+      lang,
+      log,
+      maxVersion,
+      minVersion,
+      reason,
+      userAgentInfo,
```
tofumatt: If `navigator.userAgent` is undefined the app init will still set it to `{ browser, os } = UAParser(undefined)`. That will be a lot of null values but it'll be set, as the start of every request sets the state of `api.userAgent` and `api.userAgentInfo`. So this should never be empty.

But I'll add the empty object anyway, doesn't hurt.

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -0,0 +1,93 @@
...
+    } = this.props;
+    const downloadUrl = `https://www.mozilla.org/${lang}/firefox/`;
+    let message = i18n.gettext(
+      'This add-on is not compatible with your web browser.');
+
```
tofumatt: Good point. 👍

src/core/utils.js
```diff
@@ -305,3 +297,77 @@ export function render404IfConfigKeyIsFalse(
...
+    // first.
+    if (minVersion && mozCompare(browser.version, minVersion) === -1) {
+      if (minVersion === '*') {
+        _log.error(oneLine`minVersion of "*" was passed to
+          isCompatibleWithUserAgent(); bad add-on version data`,
+          { browserVersion: browser.version, minVersion }
+        );
+      }
+
+      return { compatible: false, reason: INCOMPATIBLE_UNDER_MIN_VERSION };
```
tofumatt: I've done it in the `getClientCompatibility` method below instead as it was easier to add them there and made more sense to return them there too.

tests/client/amo/components/TestAddonCompatibilityError.js
tofumatt: I didn't know it would fallback! Neat.

src/amo/components/AddonCompatibilityError/index.js
kumar303: On the client side, it's possible for any component to render for a split second without any redux data. This could happen during navigation from one page to another but it won't ever happen server side (the redux data will always be pre-loaded).

tests/client/amo/components/TestAddonCompatibilityError.js
kumar303: Well, I should clarify that it only fall back if `props.lang` is undefined. If `props.lang` is `null` or an empty string, it won't fall back.

src/amo/components/AddonCompatibilityError/index.js
tofumatt: Ah, right.

src/amo/components/AddonDetail.js
```diff
@@ -197,7 +209,10 @@ export class AddonDetailBase extends React.Component {
             {i18n.gettext('Extension Metadata')}
           </h2>
           <AddonMeta addon={addon} />
-          <InstallButton {...this.props} />
+          <InstallButton {...this.props} disabled={!compatible} />
+          {!compatible ? (
+            <AddonCompatibilityError reason={reason} />
```
tofumatt: Thanks! I added errors if they weren't supplied because obviously that's needed 😄

tofumatt:
Okay, I think I've addressed all the PR comments and extended the test coverage. Thanks for catching the lack of min/max versions!

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -0,0 +1,94 @@
...
+    } = this.props;
+    const downloadUrl = `https://www.mozilla.org/${lang}/firefox/`;
+    let message;
+
+    if (typeof maxVersion === 'undefined') {
+      throw new Error('maxVersion is required; it cannot be undefined');
+    }
+    if (typeof minVersion === 'undefined') {
+      throw new Error('minVersion is required; it cannot be undefined');
+    }
```
kumar303: I think this would be a better place to check if `reason` is undefined because that's definitely a programming mistake.

src/amo/containers/App.js
```diff
@@ -42,11 +47,30 @@ export class AppBase extends React.Component {
...
+    // from the request headers on our first (server) request. If that's the
+    // case we try to load them from navigator.
+    if (!userAgent && _navigator && _navigator.userAgent) {
+      log.info(
+        'userAgent not in state on App load; using navigator.userAgent.');
+      setUserAgent(_navigator.userAgent);
+    }
+
     // Use addonManager.addChangeListener to setup and filter events.
     _addChangeListeners(handleGlobalEvent, mozAddonManager);
```
kumar303: Hmm, I would feel better if this were called before `setUserAgent` because dispatching a redux action will cause re-rendering.

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -0,0 +1,94 @@
...
+      message = i18n.sprintf(i18n.gettext(oneLine`This add-on requires a
+        <a href="%(downloadUrl)s">newer version of Firefox</a> (at least
+        version %(minVersion)s). You are using Firefox %(yourVersion)s.`
+      ), {
+        downloadUrl,
+        minVersion,
+        yourVersion: userAgentInfo.browser.version,
+      });
+    } else {
+      throw new Error('AddonCompatibilityError requires a "reason" prop');
```
kumar303: This isn't the correct place to throw an error. If you get to this `else` branch it means `reason` was undefined (which is better handled up above) *or* we forgot to map an error message to this reason. I think you should allow this latter case and just show a generic message. It would probably be good to log a warning about it though.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -291,19 +298,27 @@ describe('AddonDetail', () => {
...
+  });
+
   it('disables a theme preview for unsupported clients', () => {
     const rootNode = renderAsDOMNode({
       addon: {
         ...fakeAddon,
         type: ADDON_TYPE_THEME,
       },
-      clientSupportsAddons: () => false,
+      getClientCompatibility: getClientCompatibilityFalse,
```
kumar303: Does this test throw an error if you delete some required props from `<AddonCompatibilityError />` within the `AddonDetail` render function? If not, I'd like to see a test that makes sure `AddonDetail` is passing the right props to `AddonCompatibilityError`. You could use a stub component (passed via props) and make assertions about what props it received.

tests/client/core/actions/test_index.js
```diff
@@ -28,6 +30,35 @@ describe('core actions setClientApp', () => {
...
+  it('creates the SET_USER_AGENT action', () => {
+    const userAgent = userAgents.chrome[0];
+
+    assert.deepEqual(actions.setUserAgent(userAgent), {
+      type: 'SET_USER_AGENT',
+      payload: { userAgent },
+    });
+  });
+
+  it('allow an empty userAgent value', () => {
```
kumar303: nit: allows...

tests/client/core/actions/test_index.js
```diff
@@ -28,6 +30,35 @@ describe('core actions setClientApp', () => {
...
+    const userAgent = '';
+
+    assert.deepEqual(actions.setUserAgent(userAgent), {
+      type: 'SET_USER_AGENT',
+      payload: { userAgent },
+    });
+  });
+
+  it('allow an undefined userAgent value', () => {
```
kumar303: nit: allows...

tests/client/core/actions/test_index.js
kumar303: I think this test is no longer useful here. It should move to the reducer since that's where the parsing logic happens.

tests/client/core/actions/test_index.js
kumar303: Same thing here: this test should move to the reducer since that's where the parsing logic happens.

tests/client/core/test_utils.js
```diff
@@ -172,56 +181,169 @@ describe('getClientApp', () => {
...
+      browser: { name: 'Firefox', version: '8.0' },
+      os: { name: 'iOS' },
+    };
+    assert.deepEqual(
+      isCompatibleWithUserAgent({ minVersion: '9.0', userAgentInfo }),
+      { compatible: false, reason: INCOMPATIBLE_FIREFOX_FOR_IOS }
+    );
+  });
+
+  it('is incompatible with garbage user agent', () => {
```
kumar303: since this moved to reducer tests you can probably just delete it

tests/client/amo/components/TestAddonDetail.js
tofumatt: Well, `<AddonCompatibilityError />` throws an error if any of the required props are missing, so yes. I had to go back and fix some tests that had them missing. I can add an explicit one where `getClientCompatibility()` doesn't return all required props and throws. It's largely covered elsewhere already though, if I understand correctly.

tofumatt:
Think I got everything! r?

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -61,6 +61,13 @@ function renderAsDOMNode(...args) {
 }
 
 describe('AddonDetail', () => {
+  const getClientCompatibilityFalse = () => ({
+    compatible: false,
+    maxVersion: null,
+    minVersion: null,
+    reason: INCOMPATIBLE_NOT_FIREFOX,
+  });
```
kumar303: It's a bit hard to keep return values like this for tests in sync with real code. Flow is really helpful here. I will nominate a Flow experiment for the sprint: https://github.com/mozilla/addons-frontend/issues/2044

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -291,19 +298,39 @@ describe('AddonDetail', () => {
...
+    });
+    assert.isTrue(
+      rootNode.querySelector('.InstallButton-switch input').disabled);
+  });
+
+  it('throws an error if compatibility props are missing', () => {
+    assert.throws(() => {
+      renderAsDOMNode({
+        getClientCompatibility: () => ({
+          compatibility: false,
```
kumar303: When you need to delete props you can't use a spread operator but the code you have here is still susceptible to future test failures when the object signature changes. An easy way to protect against that is this pattern:

````js
// This would be defined at the top of the test.
const incompatibleClientResult = {
  compatible: false,
  maxVersion: null,
  minVersion: null,
  reason: INCOMPATIBLE_NOT_FIREFOX,
};
// Now, in your test you can write this:
const result = { ...incompatibleClientResult };
delete result.minVersion;
assert.throws(
  () => renderAsDOMNode({ getClientCompatibility: () => result },
   /minVersion is required/,
);
````

File: tests/client/core/test_utils.js [hunk 48]
```diff
@@ -172,56 +181,161 @@ describe('getClientApp', () => {
...
+      assert.deepEqual(
+        isCompatibleWithUserAgent({ userAgentInfo: UAParser(userAgent) }),
+        { compatible: false, reason: INCOMPATIBLE_FIREFOX_FOR_IOS },
+        `UA string: ${userAgent}`);
     });
   });
 
-  it('returns false for non-string user agent values', () => {
-    assert.strictEqual(clientSupportsAddons(false), false);
+  it('should mark Firefox for iOS as incompatible reason, not version', () => {
```
kumar303: I'm not sure what this sentence is saying

src/core/utils.js
```diff
@@ -305,3 +297,80 @@ export function render404IfConfigKeyIsFalse(
...
+    }
+
+    // If we made it here we're compatible (yay!)
+    return { compatible: true };
```
kumar303: Now that I can see the big picture, I think this should set `reason` to `null` because otherwise this result may cause failures in things that check for an `undefined` value. I know that nothing should be using the `reason` value when `compatible` is `false` but `null` makes it more compatible with the other modes just in case.

tests/client/amo/components/TestAddonDetail.js
tofumatt: wfm, cool

tests/client/core/test_utils.js
tofumatt: Yeah I tried to be short. It's saying "if Firefox for iOS is the browser, the error code should be related to Firefox iOS not the version mismatch". I'll make it make more sense.

tofumatt:
🎉

Issue #1975, opened 2017-03-06
AddType not found should not be a 500


muffinresearch:
Looks like certain paths lead to checking addonType and when the addonType is found it throws which results in a 500 error. 

One example is user urls which don't have a mobile component.

https://addons-dev.allizom.org/en-US/firefox/user/ensolis/

See both of these for examples of the same error:

- https://sentry.prod.mozaws.net/operations/addons-frontend-amo-dev/issues/384916/
- https://sentry.prod.mozaws.net/operations/addons-frontend-amo-dev/issues/384914/

I would think that most cases like this should actually be a 404?

muffinresearch:
Would be nice to fix this for release if possible.

mstriemer:
Could we set up nginx to redirect the `/blah/blah/user/.*` URLs to addons-server?

muffinresearch:
> Could we set up nginx to redirect the /blah/blah/user/.* URLs to addons-server?

Could do, but it seems odd to me that there's an assumption that the first part of the url after `/%locale%/%clientApp%/` is a `addonType`. 

We're going to need a user page for desktop soon in any case.

tofumatt:
> Could do, but it seems odd to me that there's an assumption that the first part of the url after `/%locale%/%clientApp%/` is a `addonType`.

It's because of how react-router works; we can't supply a regex for the URL (which is really just `/themes/` and `/extensions/` for our purposes) so the route is `/:addonType/` and that catches everything 😞

kumar303:
The react-router docs are not great but I don't see an easy way to fix this. I guess we could just add a route for `user` so that it is forced into a 404? https://github.com/ReactTraining/react-router/blob/master/docs/guides/RouteMatching.md

kumar303:
hmm, clicking on the user from here is still a 500 for me https://addons-dev.allizom.org/en-US/firefox/addon/tile-tabs/

ValentinaPC:
Verified on FF51(Android 6.0.1) using AMO-dev: 500 page for me too.
Please see video: 
![videotogif_2017 03 07_16 42 24](https://cloud.githubusercontent.com/assets/15685960/23661488/925ae668-0355-11e7-849a-2cfd8c17283c.gif)

muffinresearch:
@ValentinaPC this patch isn't live yet the rev that's on -dev is 0fc05a5a7a8e4a954eff9e5842661e1c8a2120b5 which is before that patch landed.

[...]

PR #1981, opened 2017-03-07
Force a 404 on user pages

kumar303:
Kinda sorta addresses https://github.com/mozilla/addons-frontend/issues/1975

Dumbest fix ever? I couldn't fully test this locally but doing `curl -I http://localhost:3000/en-US/firefox/user/eXXile/` gave me a 500 before and a 404 after. I think our local proxy script is a little different which is maybe why I couldn't fully reproduce it when clicking on a user link from the details page.


muffinresearch:
r+

PR #2017, opened 2017-03-08
Fix invisible search input for long search label text

### LINKED ISSUE #2015: Text entered in search box is not visible in French language
Steps to reproduce:
1.On Android , open AMO-dev
2.Change the language to French
3.Enter some text in search box

Expected results:
Search text can be entered without issues

Actual results:
Text entered is not visible

NOTES:
Not reproducible for another languages, but we were not able to find another language where default text from search box is so long that it needs to be truncated 

notvisible

muffinresearch:
Fixes #2015

### DIFF
--- a/src/ui/components/SearchInput/index.js
+++ b/src/ui/components/SearchInput/index.js
@@ -42,6 +42,10 @@ export default class SearchInput extends React.Component {
       // There is no transitionstart event, but animation will start if there is no value.
       this.setState({ animating: true });
     }
+    // Handle the possibility that onTransitionEnd will never fire if
+    // the CSS animation is a no-op due to the size of text in the label.
+    // CSS Animation time + 25ms.
+    this.onTransitionEndTimeout = window.setTimeout(this.onTransitionEnd, 275);
   }
 
   onInput = (e) => {
@@ -49,6 +53,9 @@ export default class SearchInput extends React.Component {
   }
 
   onTransitionEnd = () => {
+    if (this.onTransitionEndTimeout) {
+      window.clearTimeout(this.onTransitionEndTimeout);
+    }
     this.setState({ animating: false });
   };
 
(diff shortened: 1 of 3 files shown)


src/ui/components/SearchInput/style.scss
```diff
@@ -26,6 +26,8 @@ $icon-margin-end: 12px;
 
   opacity: 0;
   position: absolute;
+  // If changing this delay please update the JS setTimeout
+  // in the SearchInput component.
```
kumar303: Thanks for the comments :)

src/ui/components/SearchInput/index.js
```diff
@@ -42,6 +42,9 @@ export default class SearchInput extends React.Component {
       // There is no transitionstart event, but animation will start if there is no value.
       this.setState({ animating: true });
     }
+    // Handle the possibility that onTransitionEnd will never fire if
+    // the CSS animation is a no-op due to the size of text in the label.
+    window.setTimeout(this.onTransitionEnd, 250);
```
kumar303: What about doing something like this?
````js
this.simulatedTransitionEnd = window.setTimeout(this.onTransitionEnd, 250);
````
That way the real `onTransitionEnd` can call `clearTimeout()` on it. I imagine without clearing it, the animation might look jumpy? But I don't know. If it won't make it jump then we probably don't need it.

kumar303: I'd like to see the `setTimeout` logic covered by a test. It should be pretty easy using [fake timers like this](https://github.com/mozilla/addons-frontend/blob/master/tests/client/core/components/TestHoverIntent.js#L33).

muffinresearch: It doesn't look like it jumps but would need to see it on a real device to know for sure.

muffinresearch: What would be the purpose of the test, just to see that onTransitionEnd got called?

Issue #1744, opened 2017-02-07
Layout of Extensions home is incorrect


pwalm:
- Search bar should say "Search Extensions" only
- Missing "welcome mat" describing what extensions are
- Missing "Browse by Category" button

Currently:
<img width="350" alt="screen shot 2017-02-07 at 3 13 44 pm" src="https://cloud.githubusercontent.com/assets/6797635/22715910/095795f0-ed48-11e6-85e0-504d099cec5a.png">

Mock:
<img width="443" alt="screen shot 2017-02-07 at 3 14 02 pm" src="https://cloud.githubusercontent.com/assets/6797635/22715919/14b26e48-ed48-11e6-8dc2-6e312fbd5656.png">

mstriemer:
I thought we moved the searching for extensions/themes specifically to v2.

Looks like this still needs copy.

tofumatt:
> I thought we moved the searching for extensions/themes specifically to v2.

We did, and we're missing copy for the extensions bit.

If we get the copy we can add that to the V1 but it's not something in the old app and the search is a v2 thing.

aayushsanghavi:
@tofumatt I have taken up #1745, so I'll make the patch general to fix this issue too.

mstriemer:
@aayushsanghavi here's the icon, I thought we had that one already but it looks like we just have a blue one.

[extensions.zip](https://github.com/mozilla/addons-frontend/files/821692/extensions.zip)

mstriemer:
@devaneymoz we need some copy for this welcome mat as well. Looks like the mock has some ideas about what to say but it isn't complete.

aayushsanghavi:
@mstriemer how about this - I submit my patch for now and open an issue for changing the text on the extensions page. That way, the issue won't have to be kept waiting.

mstriemer:
We have this string on the discovery pane (roughly): "Install powerful tools that make browsing faster and safer, add-ons make your browser yours." Let's go with that and we can file to change it later.

pwalm:
Love it! Ship it!

aayushsanghavi:
@mstriemer I've opened a PR but there is some merge conflict issue.

ValentinaPC:
> Search bar should say "Search Extensions" only

This is not changed, @tofumatt, is it ok?

Everything else seems ok.
Verified on FF52(Android 6.0.1) using AMO-dev.
Postfix screenshots:
[...]

Issue #1745, opened 2017-02-07
Layout of Themes home is incorrect


pwalm:
- Search bar should say "Search Themes" only
- Missing "welcome mat" describing what themes are
- Missing "Browse by Category" button

Currently:
<img width="346" alt="screen shot 2017-02-07 at 3 18 17 pm" src="https://cloud.githubusercontent.com/assets/6797635/22716033/ae0addb4-ed48-11e6-9740-fc7f1c5eb3dc.png">

Mocks:
<img width="434" alt="screen shot 2017-02-07 at 3 22 59 pm" src="https://cloud.githubusercontent.com/assets/6797635/22716142/5990270c-ed49-11e6-9205-fb8333352d6c.png">

mstriemer:
I thought we moved the searching for extensions/themes specifically to v2.

Looks like this still needs copy.

mstriemer:
@devaneymoz do you have copy for this?

Here's the icon: [themes.zip](https://github.com/mozilla/addons-frontend/files/804315/themes.zip).

I don't think we can update the search text yet since we only support searching for extensions and themes together right now.

mstriemer:
If you have sketch the [AMO - V1.sketch](https://github.com/mozilla/addons-frontend/blob/master/assets/sketch%20files/AMO%20-%20V1.sketch) file has the mocks in it and you can export the icons from there. If you don't have sketch and need another icon in the future just ping one of the developers and we can extract it for you.

aayushsanghavi:
@mstriemer thanks for the theme icon. Also, could you tell me what icon/glyphicon is to be used for the browse categories button.
And how do I change the placeholder according to the addonType? Any ideas on that?

mstriemer:
Should have looked a little closer. I don't see that icon being used yet, so here it is [bullet-list.zip](https://github.com/mozilla/addons-frontend/files/804674/bullet-list.zip).

These icons should go in the [`Icon`](https://github.com/mozilla/addons-frontend/tree/1244c0dcaff3af740e43a55544a1f1064d730a52/src/ui/components/Icon) component so they're easier to re-use.

aayushsanghavi:
@mstriemer thanks a lot! I'll get back to you in case I have a doubt.

devaneymoz:
For copy under the header "Themes" how about we go with:

Change your browser’s appearance. Choose from thousands of themes to give Firefox the look you want.

aayushsanghavi:
[...]

PR #1848, opened 2017-02-17
fix: Layout of extension module is incorrect

aayushsanghavi:
fixes: #1743 
After - 
after

PR #2024, opened 2017-03-08
fix: Layout of themes and extensions home

aayushsanghavi:
Fixes: #1744 #1745 
There is some merge conflict which would need to be fixed.

After - 
screenshot from 2017-03-09 00-30-05
screenshot from 2017-03-09 00-30-15


aayushsanghavi:
This PR somehow contains commits from my previous PR https://github.com/mozilla/addons-frontend/pull/1848.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
     }
 
     const { addonType, html } = content;
+    const themeText = 'Change your browser’s appearance. Choose from thousands of themes to give Firefox the look you want.';
```
mstriemer: These will need to be wrapped in `i18n.gettext()`. You can get `i18n` from props by adding it to line 96.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
...
+            </span>
+            <p className="Addon-Header-Text-Content">
+              {addonType === 'persona' ? themeText : extensionsText }
+            </p>
+          </div>
+        </div>
+
+        <button className="Browse-Button">
+          <Icon name="browse" />
+          <span className="Browse-by-Category-Text">browse by category</span>
```
mstriemer: This will need to be wrapped in `i18n.gettext()`, too. Can you format it as "Browse by category" as well, please?

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
     }
 
     const { addonType, html } = content;
+    const themeText = 'Change your browser’s appearance. Choose from thousands of themes to give Firefox the look you want.';
+    const extensionsText = 'Install powerful tools that make browsing faster and safer, add-ons make your browser yours.';
 
     return (
       <div className={classNames('LandingPage', `LandingPage-${addonType}`)}>
+
+        <div className="Addon-Header">
```
mstriemer: Our class names are formatted as `ComponentName-element-name[--state]`. So this should be something like `LandingPage-header`. The other classes will need to be updated as well.

You shouldn't need `--state` in any of these but that would be for something like install switches where it would have a normal class like `InstallSwitch` and then when it is in the installing state (flipped on with animated stripes) you add `InstallSwitch--installing` then once it is installed you remove that and add `InstallSwitch--installed`.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
...
+          <Icon name={classNames(`${addonType}`)} className="Addon-Icon" />
+          <div className="Addon-Header-Text">
+            <span className="Addon-Header-Text-Bold">
```
mstriemer: Try not to describe the look of the element but instead what element it is and add the look in the CSS. So this should be something like `LandingPage-heading` and then the fact that it is bold is stored in the CSS.

This could probably be an `h1` too.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
...
+              {addonType === 'persona' ? 'Themes' : 'Extensions' }
```
mstriemer: These strings should be translated.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +109,29 @@ export class LandingPageBase extends React.Component {
...
+            </span>
+            <p className="Addon-Header-Text-Content">
+              {addonType === 'persona' ? themeText : extensionsText }
+            </p>
+          </div>
+        </div>
+
+        <button className="Browse-Button">
```
mstriemer: This should link to the category page for this extension type. We have a `Link` component in `amo/components/Link` and this should use that. There are styles for the white/blue colour scheme in `Home.scss`.

Ideally the common styles would move into `Link` and this colour scheme can be chosen with something like `<Link appearance="light">...</Link>`. Since `Link` uses the `@include button()` mixin you should be able to use the colour from the homepage link and set a class like `Link--light` that will use the mixin as `@include button(#fff, $link-color)`.

src/amo/components/LandingPage.scss
```diff
@@ -3,3 +3,69 @@
...
   padding: $padding-page;
 }
+
+.Addon-Header {
+  margin: 10px;
+  max-width: 100%;
+  text-align: center;
+}
+
+.Icon-persona {
```
mstriemer: These icon styles can go into the `Icon` component's styles. The icon files themselves can be moved next to the component as well.

src/amo/components/LandingPage.scss
```diff
@@ -3,3 +3,69 @@
...
+  margin-top: 5px;
+}
+
+.Browse-Button {
+  background-color: white;
+  border-radius: 9px;
+  color: blue;
+  height: 55px;
+  font-size: 12px;
+  margin-left: calc(50% - 150px);
```
mstriemer: It would be nice to merge this style with the homepage's buttons. The styles there use `calc` as well but I think they could just be `margin: 0 20px` instead.

You should be able to remove the `border-radius`, `font-size` and `width` styles.

src/amo/components/LandingPage.scss
```diff
@@ -3,3 +3,69 @@
...
+  background-size: cover;
+  display: inline-block;
+  height: 47.8px;
+  margin-bottom: 15px;
+  width: 49px;
+}
+
+.Addon-Header-Text {
+  display: inline-block;
+  margin-left: 20px;
```
mstriemer: We need to support RTL as well so anything like `margin-left`, `padding-left` or `left` will need to be flipped for RTL.

There are some helpers you can use in `mixins.scss`. This could instead be `@include margin-start(20px)`.

src/amo/components/LandingPage.scss
```diff
@@ -3,3 +3,69 @@
...
+  background: url('~amo/img/icons/bullet-list.svg') 50% 50% no-repeat;
+  background-size: cover;
+  height: 30px;
+  margin-right: 1em;
+  width: 30px;
+}
+
+.Browse-by-Category-Text {
+  bottom: 10px;
+  position: relative;
```
mstriemer: These styles seem a little odd to me. Using `position: relative` and `position: absolute` should generally be avoided in my opinion. If you need to move it up or down a bit can you use `margin` instead? Hopefully this shouldn't be needed though. Setting the `line-height` to match the container height can work to vertically centre but that can get wonky if the text wraps in a more verbose language than English.

Maybe take a look at how the buttons on the homepage work for this.

mstriemer:
If you want to clear out those old commits I find the best way is to start a new branch off of master and use `git cherry-pick` to pull in the comments you want from oldest to most recent. If it looks good you can delete this branch and recreate it from the new one you made.

aayushsanghavi:
@mstriemer thanks a lot for the review. I'll get back to you with all the changes by tomorrow.
Also, I pulled the upstream master branch and I'm having a merge conflict there too so I'll have to fix that first. I didn't know about git cherry-pick. You taught me something new today! :)

src/amo/components/LandingPage.scss
aayushsanghavi: I tired it with margin, padding and line-height but that didn't work. Then I tried the position which seemed to work somehow.
I'll have a look at the home page button styling and try to replicate that here too.

src/amo/components/LandingPage.js
aayushsanghavi: How do I translate them?

aayushsanghavi:
@mstriemer I've read and understood your comments but I guess while making the changes I'll require your help. I'll ping you on IRC in that case, if that's fine with you.

src/amo/components/LandingPage.js
mstriemer: Using `i18n.gettext()`, they need to be wrapped individually.

```js
// There's also an extra space before the closing } that should be removed
{addonType === 'persona' ? i18n.gettext('Themes') : i18n.gettext('Extensions')}

File: src/amo/components/LandingPage.js (hunk 6 above)
Comment (aayushsanghavi) [inline src/amo/components/LandingPage.js]: @mstriemer can you please share the links of the categories page for each of the addon types?

Comment (mstriemer) [inline src/amo/components/LandingPage.js]: The extension categories should work with `<Link href="/extensions/categories/" />` and themes are at `<Link href="/themes/categories/" />`. For some reason the themes view doesn't return anything on stage.

@tofumatt do you know why there are no theme categories at https://addons.allizom.org/en-US/android/themes/categories/?

File: src/amo/components/LandingPage.js [hunk 11]
```diff
@@ -108,9 +111,32 @@ export class LandingPageBase extends React.Component {
     }
 
     const { addonType, html } = content;
+    const themeText = i18n.gettext('Change your browser’s appearance. Choose from thousands of themes to give Firefox the look you want.');
+    const extensionsText = i18n.gettext('Install powerful tools that make browsing faster and safer, add-ons make your browser yours.');
 
     return (
       <div className={classNames('LandingPage', `LandingPage-${addonType}`)}>
+
+        <div className="LandingPage-Header">
```
mstriemer: Only the component name should be capitalised for the classes, `LandingPage-header`.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +111,32 @@ export class LandingPageBase extends React.Component {
...
+          <Icon name={classNames(`${addonType}`)} />
```
mstriemer: Just `<Icon name={addonType} />` should work here.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +111,32 @@ export class LandingPageBase extends React.Component {
...
+          <div className="LandingPage-Header-Text">
+            <h1 className="LandingPage-Heading">
+              {addonType === 'persona' ? i18n.gettext('Themes') : i18n.gettext('Extensions')}
+            </h1>
+            <p className="LandingPage-Heading-Content">
+              {addonType === 'persona' ? i18n.gettext(themeText) : i18n.gettext(extensionsText)}
```
mstriemer: These are already translated and don't need the extra `i18n.gettext()` call.

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,44 @@
...
+
+.LandingPage-Heading {
+  font-size: 20px;
+}
+
+.LandingPage-Heading-Content {
+  margin-top: 5px;
+}
+
+Link--light {
```
mstriemer: This is missing the `.` before `Link--light`. This should be moved into the styles for `Link` too.

src/ui/components/Icon/styles.scss
```diff
@@ -32,3 +32,29 @@
   background: url('./user-dark.svg') center no-repeat;
   background-size: contain;
 }
+
+.Icon-browse {
+  background: url('./bullet-list.svg') 50% 50% no-repeat;
+  background-size: cover;
+  height: 30px;
+  margin-right: 1em;
```
mstriemer: This should use `@include margin-end(1em)`. The margin styles for these icons should probably move into where they're being used, too. You can add `className="LandingPage-browse-icon"` to the `Icon` call and then move this to a rule in LandingPage.scss, same with the other two.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +111,32 @@ export class LandingPageBase extends React.Component {
...
+            </p>
+          </div>
+        </div>
+
+        <Link className="Browse-Button" appearance="light"
```
mstriemer: This should actually use `Button` from 'ui/components/Button'. You'll need to add support for the light appearance too. Something like:

```diff
diff --git a/src/ui/components/Button/index.js b/src/ui/components/Button/index.js                                                                                                                                                            
index 6b9e0160..b02c89cc 100644                                                                                                                                                                                                               
--- a/src/ui/components/Button/index.js                                                                                                                                                                                                       
+++ b/src/ui/components/Button/index.js                                                                                                                                                                                                       
@@ -6,6 +6,7 @@ import './Button.scss';                                                                                                                                                                                                       
                                                                                                                                                                                                                                              
 export default class Button extends React.Component {                                                                                                                                                                                        
   static propTypes = {                                                                                                                                                                                                                       
+    appearance: PropTypes.string,                                                                                                                                                                                                            
     children: PropTypes.node,                                                                                                                                                                                                                
     className: PropTypes.string,                                                                                                                                                                                                             
     href: PropTypes.string,                                                                                                                                                                                                                  
@@ -13,14 +14,16 @@ export default class Button extends React.Component {                                                                                                                                                                     
   }                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                              
   static defaultProps = {                                                                                                                                                                                                                    
+    appearance: undefined,                                                                                                                                                                                                                   
     size: 'normal',                                                                                                                                                                                                                          
   };                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                              
   render() {                                                                                                                                                                                                                                 
-    const { children, className, href, size, ...rest } = this.props;                                                                                                                                                                         
+    const { appearance, children, className, href, size, ...rest } = this.props;                                                                                                                                                             
     const props = {                                                                                                                                                                                                                          
       className: classNames('Button', className, {                                                                                                                                                                                           
         'Button--small': size === 'small',                                                                                                                                                                                                   
+        'Button--light': appearance === 'light',                                                                                                                                                                                             
       }),                                                                                                                                                                                                                                    
       ...rest,                                                                                                                                                                                                                               
     };

File: src/amo/components/LandingPage.scss (hunk 14 above)
Comment (mstriemer) [inline src/amo/components/LandingPage.scss]: This should move to `Button` actually, it will call `Link` and provide the button styles.

File: src/amo/components/LandingPage.scss [hunk 17]
```diff
@@ -1,5 +1,44 @@
...
+@import "~core/css/inc/mixins";
 
 .LandingPage {
   padding: $padding-page;
 }
+
+.LandingPage-Header {
+  margin: 10px;
+  max-width: 100%;
+  text-align: center;
```
mstriemer: I think you can remove this and the other `text-align`s from this file.

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,44 @@
...
+  margin-top: 5px;
+}
+
+Link--light {
+  @include button(#fff, $link-color);
+}
+
+.Browse-Button {
+  height: 55px;
+  margin-left: calc(50% - 150px);
```
mstriemer: I think you can do: `margin: 0 10px; width: 100%` instead of this.

src/ui/components/Icon/styles.scss
```diff
@@ -32,3 +32,29 @@
...
+  width: 30px;
+}
+
+.Icon-persona {
+  background: url('./themes.svg') 50% 50% no-repeat;
+  background-size: cover;
+  display: inline-block;
+  height: 50px;
+  margin-bottom: 15px;
```
mstriemer: I don't think this bottom margin will work properly. It looks about right in English but the size of the container beside it can change based on the language. You can instead use flexbox to vertically centre it.

You'll want to set it up so you've got two containers that you can align within the parent, and fortunately it's already setup like that.

Then in CSS you can set the parent to `display: flex` and and tell the icon to be centred vertically.

```scss
.LandingPage-header {
  display: flex;
}

.LandingPage-icon {
  align-self: center;
}
```

There's a good [guide to flex box on CSS Tricks](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) that I always end up referencing since there are so many options.

aayushsanghavi: Hahaha yeah when I was working on Mozilla's Network Pulse, my mentor mmavis shared the same link. It is a very nicely written blog.

aayushsanghavi:
@mstriemer thanks for your review and sorry for the late response.
I've made the changes as you suggested but for some reason in the button component I couldn't work with the --light appearance class. I couldn't override the style rules of `@include button();` in the initial lines of Button.scss. Please help with that. For now I've specified my own background-color and color rules in the Browse-button class.

src/ui/components/Button/index.js
```diff
@@ -6,21 +6,24 @@ import './Button.scss';
...
     size: 'normal',
   };
 
   render() {
-    const { children, className, href, size, ...rest } = this.props;
+    const { appearance, children, className, href, size, ...rest } = this.props;
     const props = {
       className: classNames('Button', className, {
         'Button--small': size === 'small',
+        'Button--light' : appearance === 'appearance',
```
aayushsanghavi: @mstriemer  so sorry for this, you had to review such mistakes. I feel like laughing at myself.

aayushsanghavi:
@mstriemer I'll read through the review after Saturday, if that's okay. I have college exams going on.
And **thank you so much** for taking the time out to review my patch as well as for your patience. Also, I apologize for making such stupid mistakes. I'll make sure that next time I thoroughly read through my code before I commit it. Thanks again. You're awesome!!

mstriemer:
Don't worry about it, it's looking good! Good luck with the exams.

aayushsanghavi:
extensions
thmes
Current look of the extensions and themes home page

Issue #2029, opened 2017-03-09
Cannot register on Stage from Android


vcarciu:
Steps to reproduce:
1.On Android, go to AMO-Stage and try to register

Expected results:
After pressing confirmation link from confirmation email received, user is successfully logged to AMO-Stage

Actual results:
Server error page is displayed and user is not logged in

NOTES:
Same like https://github.com/mozilla/addons-frontend/issues/1660 , but on Stage

muffinresearch:
Hmm this seems odd because the patch for that should have landed 

Here's the sentry report for it: https://sentry.prod.mozaws.net/operations/addons-frontend-amo-stage/issues/385485/

muffinresearch:
Ah ok - this sounds like it might be FXA config related. 

The patch I referred to above being landed was actually to force a 404 for `user/[USER]` whereas this path is `users/edit/`. 

Whilst config can probably prevent us hitting this URL, it would be good to add a route to force a 404 for this case so that we do still fall-back to the non-mobile site without blowing up.

muffinresearch:
Fwiw: I've just tried the same thing on our prod instance and a 500 occurs there the same.

kumar303:
The workaround in https://github.com/mozilla/addons-frontend/issues/2034 didn't work. Here's what happens: 
* `addons-frontend` starts by serving a `404` for `/users/edit/`
* This tells the proxy to let `addons-server` serve `/users/edit/`
* `addons-server` sees `/users/edit/` and redirects to `/users/edit`
* `addons-frontend` sees `/users/edit` and adds a slash to the end
* hilarity ensues

kumar303:
The server change landed on stage and I was just about to sign in with a brand new account 🍨

vcarciu:
Verified as fixed

vcarciu:
Please see postfix video : 
![regstage](https://cloud.githubusercontent.com/assets/1583842/23792204/aabca6b6-058e-11e7-852b-00158b6bed31.gif)

vcarciu:
Please see postfix video : 
![regstage](https://cloud.githubusercontent.com/assets/1583842/23799068/dae59390-05af-11e7-8925-203dd2215d0b.gif)

PR #2034, opened 2017-03-09
Force all /users/ URLs to be served by addons-server

kumar303:
Temporary workaround for https://github.com/mozilla/addons-frontend/issues/2029

The real fix will arrive in https://github.com/mozilla/addons-server/pull/4890

### DIFF
--- a/src/amo/routes.js
+++ b/src/amo/routes.js
@@ -29,7 +29,10 @@ export default (
     <IndexRoute component={Home} />
     <Route path="addon/:slug/" component={DetailPage} />
     <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
-    {/* Hack to make the proxy serve this URL from addons-server */}
+    {/* These user routes are to make the proxy serve each URL from */}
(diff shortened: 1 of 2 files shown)


src/amo/routes.js
```diff
@@ -29,7 +29,10 @@ export default (
     <IndexRoute component={Home} />
     <Route path="addon/:slug/" component={DetailPage} />
     <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
-    {/* Hack to make the proxy serve this URL from addons-server */}
+    {/* These user routes are to make the proxy serve each URL from */}
+    {/* addons-server until we can fix the :visibleAddonType route below. */}
+    {/* https://github.com/mozilla/addons-frontend/issues/2029 */}
+    <Route path="users/edit/" component={NotFound} />
```
muffinresearch: I wonder if this should have a placholder since there's quite a lot of options for other urls one could hit: https://github.com/mozilla/addons-server/blob/master/src/olympia/users/urls.py#L20

kumar303: Good call

Issue #2037, opened 2017-03-10
Redirect loop on /users/edit[/]


muffinresearch:
### Describe the problem and steps to reproduce it:
From https://github.com/mozilla/addons-frontend/issues/2029#issuecomment-285452339

Whilst the fxa config has been updated to fix the redirection post-registration the redirect loop still exists.
 
- Login and click my profile.
- And then edit profile.
- Then click "view mobile site" in the footer.
- addons-frontend starts by serving a 404 for /users/edit/
- This tells the proxy to let addons-server serve /users/edit/
- addons-server sees /users/edit/ and redirects to /users/edit
- addons-frontend sees /users/edit and adds a slash to the end
- hilarity ensues

### What happened?

A redirect loop

### What did you expect to happen?

Fallback to /users/edit (we could also not show the "View mobile site" link on this page)
Longer term this will be fixed by added pages for these views in the new mobile site.

muffinresearch:
One way to prevent this behaviour would be to have exclusions for adding slashes.

muffinresearch:
> One way to prevent this behaviour would be to have exclusions for adding slashes.

Or another way would be to add a slash on the amo-server side. Might mean we'd need to change the redirect endpoint for server logins to avoid an unnecessary 301.

kumar303:
We could also put some middleware right at the front of the chain that responds with 404s to known addons-server URLs. However, this could be a little confusing and won't work for when someone is loading pages on the client.

tofumatt:
QA: we'd love to cherry-pick this patch into this week's tag, so if you could test this and the URLs it should cover (`/users/edit`, etc.) on `-dev` and confirm it's working that would be great. ❤️

ValentinaPC:
Verified this on AMO-dev FF52(Win 7) while following the first 3 steps (Login and click my profile->click edit profile->click "view mobile site" in the footer) and obtained the following results:

 - default scenario will reload "https://addons-dev.allizom.org/en-US/firefox/users/edit#" page on desktop
![1scenario](https://cloud.githubusercontent.com/assets/15685960/23991548/6542b11c-0a42-11e7-9a81-1db1b19e74c1.gif)


 - +Clicking on "Other Applications" like Thunderbird and Seamonkey or My Collections-> addons-frontend starts by serving a 404 page and a Server Error page 
[...]

PR #2046, opened 2017-03-13
Revert "Force all /users/ URLs to be served by addons-server (#2034)"

kumar303:
This will make https://github.com/mozilla/addons-frontend/issues/2037 go away but only until we shadow some other URL.

This reverts commit f42934e542ae7383a8302296d85ac7255238177c because a 500 is going to be more helpful until we have implementations of these pages.


muffinresearch:
r+

Issue #2054, opened 2017-03-13
Fix tests in the Paginate component


kumar303:
There are multiple tests for the Paginate comment that are failing silently due to not returning the promises. It also needs a bit of cleanup.

muffinresearch:
I wonder if something like the `catch-or-return` rule here [1] would have caught this or at least some of it?

[1] https://github.com/xjamundx/eslint-plugin-promise#promise-rules

kumar303:
> I wonder if something like the catch-or-return rule here [1] would have caught this or at least some of it?

Good idea! Filed https://github.com/mozilla/addons-frontend/issues/2065

Issue #1720, opened 2017-02-07
No pagination in All Written Reviews page


ValentinaPC:
Steps to reproduce:
1. Load AMO-dev on your Android device
2. Go to details page for an add-on with reviews
3. Click to "Read all `x` reviews"

Expected results:
Review page is loaded. User can navigate through pages and see all reviews.

Actual results:
There isn't any navigation possibility.

Notes/Issues:
Verified on AMO-dev FF51(Android 6.0.1)
Video for this issue:
![videotogif_2017 02 07_08 32 17](https://cloud.githubusercontent.com/assets/15685960/22680409/729780d0-ed10-11e6-9874-fd376361ddd0.gif)

kumar303:
Yes, I can confirm. There are TODO comments in the code to handle paging: https://github.com/mozilla/addons-frontend/blob/master/src/amo/api/index.js#L53

muffinresearch:
I think we can probably make this one not a blocker.

ValentinaPC:
Verified as fixed on AMO-dev FF53(Android 6.0.1) for themes and add-ons.
Postfix screenshot:
![2017-04-28_0956](https://cloud.githubusercontent.com/assets/15685960/25517623/ec45251e-2bf8-11e7-83ee-9e589ccb6556.png)

PR #2055, opened 2017-03-13
Fix tests for Paginate

kumar303:
Welp, I fell down this rabbit hole while trying to patch https://github.com/mozilla/addons-frontend/issues/1720

Fixes https://github.com/mozilla/addons-frontend/issues/2054


src/core/components/Paginate/Paginate.scss
```diff
@@ -1,17 +1,17 @@
 @import '~core/css/inc/mixins';
 
-.Paginator-page-number {
+.Paginate-page-number {
```
tofumatt: 👍

tests/client/core/components/TestPaginatorLink.js
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    it('passes query params to the link', () => {
+      const page = 3;
+      const queryParams = { color: 'red' };
+      const LinkComponent = sinon.spy(() => <div />);
+      renderLink({ queryParams, page, LinkComponent });
+
+      assert.ok(LinkComponent.called, 'custom LinkComponent was not rendered');
+      const props = LinkComponent.firstCall.args[0];
```
tofumatt: This is a nitpick but I kinda like when simpler tests like this declare things at the top, then do all their assertions at the bottom in succession.

It makes it easier to scan, but also doesn't imply anything is changing between the assertions. Because here nothing is different between `assert.ok(LinkComponent.called)` and `assert.deepEqual(props.to.query, [...])`, I think moving the `const props = [...]` line above the assertions makes sense.

We have more complex tests where we test state before/after a click, and that's when I expect to see variable definitions, etc. between assertions.

kumar303: Ah, yeah, I agree about doing setup first and assertions later. However, the `props` assignment can't be moved up because it relies on the stub getting called correctly. If I move it up above `assert.ok(stub.called)` then the test fails in a harder to diagnose error in case it wasn't called:
````
LinkComponent.firstCall is null
````

tofumatt:
Oh right! Oops :-)

kumar303:
Thanks for the review. I tested this locally with my search results but I'll give it a go on dev as well.

Issue #2059, opened 2017-03-14
Some strings will not be translated due to the use of oneLine


muffinresearch:
We have a babel transform `dedent` which needs to be used for template strings wrapped in gettext calls.

The reason for this is that dedent is a babel transform that happens before the extraction process. Since the point that a different indent removal lib was introduced (`oneLine`) anything that's a translated template string that doesn't use dedent is not extracted.

See https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=i18n.gettext%28oneLine

muffinresearch:
Having converted all that list to use dedent there's still a problem with extraction on lines like these:  https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonCompatibilityError/index.js#L67

tofumatt:
Is there a way we can use the same type of setup for oneLine? I like that it doesn’t add weird newlines making it easier to test string contents, and also just doesn’t output erroneously `\n`.

ValentinaPC:
Please add some STR for us to check this or the label "qa: not needed". Thanks!

PR #2063, opened 2017-03-14
Update amo locales 2017 03 14

muffinresearch:
This required some local patches to the babel-gettext-extractor but it can now statically evaluate template literals (thanks to this patch on the upstream of our upstream [1])  and just an equivalent of on the extracted strings pre-emptively. I'll be pushing that patch to the fork in due course and updating the dep to point to that patch.

[1] https://github.com/jruchaud/babel-gettext-plugin/pull/4/files


muffinresearch:
> We chatted about this, makes sense. Maybe worth updating the i18n docs to say gettext() calls need to use plain template strings if multiline and our gettext extraction will handle them.

Good call, I'll do that when I do the lib bump to actually fix: https://github.com/mozilla/addons-frontend/issues/2059

muffinresearch:
Ah actually there's a problem with this approach in that the real gettext call needs to be run through the same oneLine process, since extract and runtime are two different things. Will see if I can override gettext to make that work.

muffinresearch:
Superceded by #2086

PR #2066, opened 2017-03-14
Fix redirect loop

### LINKED ISSUE #2037: Redirect loop on /users/edit[/]

tofumatt:
Adds a config that prevents trailing slashes from being added to certain URLs.

fixes #2037

### DIFF
--- a/config/default-amo.js
+++ b/config/default-amo.js
@@ -78,6 +78,26 @@ module.exports = {
     'user-media',
     '__version__',
   ],
+  // These URLs are exceptions to our trailing slash URL redirects; if we
+  // find a URL that matches this pattern we won't redirect to the same url
+  // with an appended `/`. This is usually because if we redirect, it will
+  // cause a redirect loop with addons-server; see:
+  // https://github.com/mozilla/addons-frontend/issues/2037
+  //
+  // We use $lang and $clientApp as placeholders so we can have URLs in this
+  // list that don't include those URL pieces, if needed.
+  validTrailingSlashUrlExceptions: [
+    // User URLs, found in:
+    // https://github.com/mozilla/addons-server/blob/master/src/olympia/users/urls.py#L20
+    '/$lang/$clientApp/user/abuse',
+    '/$lang/$clientApp/user/rmlocale',
+    '/$lang/$clientApp/users/ajax',
+    '/$lang/$clientApp/users/delete',
+    '/$lang/$clientApp/users/edit',
+    '/$lang/$clientApp/users/login',
+    '/$lang/$clientApp/users/logout',
+    '/$lang/$clientApp/users/register',
+  ],
 
   trackingEnabled: true,
   trackingId: 'UA-36116321-7',
--- a/src/core/middleware.js
+++ b/src/core/middleware.js
@@ -19,15 +19,18 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
   // or missing.
   const [langFromURL, appFromURL] = URLParts;
 
-  // Get language from URL or fall-back to detecting it from accept-language header.
+  // Get language from URL or fall-back to detecting it from accept-language
+  // header.
   const acceptLanguage = req.headers['accept-language'];
-  const { lang, isLangFromHeader } = getLanguage({ lang: langFromURL, acceptLanguage });
+  const { lang, isLangFromHeader } = getLanguage({
+    lang: langFromURL, acceptLanguage });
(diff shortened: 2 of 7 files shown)


tests/client/core/test_middleware.js
```diff
@@ -248,6 +255,50 @@ describe('Trailing Slashes Middleware', () => {
...
+      originalUrl: '/firefox/no/trailing',
+      headers: {},
+    };
+    trailingSlashesMiddleware(
+      fakeReq, fakeRes, fakeNext, { _config: fakeConfig });
+    assert.notOk(fakeRes.redirect.called);
+    assert.ok(fakeNext.called);
+  });
+
+  it('should handle trailing slash exceptions with $lang/$clientApp', () => {
```
muffinresearch: I may have missed it, but it would be good to ensure that clientApp and or locale prefixed urls still get redirs if not excluded.

muffinresearch:
I wonder if it's worth exploring using react-router to handle the slashes instead of a middleware? 

The middleware will apply to everything regardless of whether we're actually handling that route or not. If we did this in the router to routes we care about that are missing them it might potentially remove a lot of unecessary redirections? It would also then mean we're always consolidating on directory style urls consistently but should not impact anything addons-server currently does differently.

This would also negate the need to wrangle prefixes.

See https://github.com/ReactTraining/react-router/issues/820#issuecomment-256814655 for an example it seems way simpler imho.

config/default-amo.js
```diff
@@ -78,6 +78,17 @@ module.exports = {
...
+  // These URLs are exceptions to our trailing slash URL redirects; if we
+  // find a URL that matches this pattern we won't redirect to the same url
+  // with an appended `/`. This is usually because if we redirect, it will
+  // cause a redirect loop with addons-server; see:
+  // https://github.com/mozilla/addons-frontend/issues/2037
+  //
+  // We use $lang and $clientApp as placeholders so we can have URLs in this
+  // list that don't include those URL pieces, if needed.
+  validTrailingSlashUrlExceptions: [
+    '/$lang/$clientApp/user/edit',
```
muffinresearch: This is nice for completeness but definitely makes the implementation more involved due to needing to wrangle the prefixes. I wonder if this could still be useful if it worked with either globs or even if this array defines path suffixes to match against e.g. don't redirect if the path endsWith `/user/edit` or are there cases where that would fall apart?

tofumatt: I suppose that would work fine, we'd still need to split on `?` to make sure we didn't get messed with by query strings but I suppose it could work.

tests/client/core/test_middleware.js
tofumatt: Ah, I didn't actually test that but I can. Sure.

tofumatt:
> If it hasn't already I wonder if it's worth exploring using react-router to handle the slashes instead of a middleware?

I like that idea, I'll give it a go.

tofumatt:
We talked about this on vidyo and for now we'll still need the middleware.

I've extended the test coverage and fixed an issue with query strings. I've also added back the routes backed out in #2046. And then some.

Another r? would be great.

src/core/middleware.js
```diff
@@ -129,13 +133,39 @@ export function prefixMiddleWare(req, res, next, { _config = config } = {}) {
   return next();
 }
 
-export function trailingSlashesMiddleware(req, res, next) {
-  const URLParts = req.originalUrl.split('?');
+export function trailingSlashesMiddleware(
+  req, res, next, { _config = config } = {}
```
muffinresearch: Does this need to be indented?

tofumatt: Yeah, it breaks 80 chars if not.

muffinresearch:
@tofumatt r+wc - I reckon let's land this and test it out on -dev.

PR #2069, opened 2017-03-15
chore: Add eslint rule for Promises (fix #2065)

### LINKED ISSUE #2065: Add ES lint rule to catch broken promises in the test suite

Create a test with a promise but forget to return the promise or call `done()`.

### What happened?

It could fail silently. See https://github.com/mozilla/addons-frontend/issues/2054

An ES lint rule should yell at you!

Something in here should do it: https://github.com/xjamundx/eslint-plugin-promise#promise-rules

If adding is too intrusive, let's at least add it just to `tests/.eslintrc`.

tofumatt:
I spent a few hours doing mindless tile replacement so I felt like doing mindless code linting cleanup too. 😉 

Adds returns to all Promises and ensures all test code that
uses a Promise returns it so all code is executed.

This is a lot of noise but and makes for slightly more verbose
tests with "needless" `return true;` statements, but it will
prevent us from writing unexecuted test code in the future. 👍🏻

-----

This ends up making the code a fair bit more verbose but I think it's worth it. I know it looks like a lot of noise here but it keeps us writing smarter Promises.

### DIFF
--- a/package.json
+++ b/package.json
@@ -253,6 +253,7 @@
     "eslint-config-airbnb": "^13.0.0",
     "eslint-plugin-import": "^2.2.0",
     "eslint-plugin-jsx-a11y": "^2.2.3",
+    "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-react": "^6.10.3",
     "fetch-mock": "^5.10.0",
     "file-loader": "^0.11.1",
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -1,9 +1,18 @@
 {
+  "extends": [
+    "plugin:promise/recommended",
+  ],
+  "plugins": [
+    "promise",
+  ],
   "rules": {
     "import/no-extraneous-dependencies": ["error", {
       // Allow dev-dependencies in this directory.
       "devDependencies": true
     }],
+    "promise/always-return": "off",
+    "promise/avoid-new": "off",
+    "promise/no-nesting": "off",
   },
   "settings": {
     "import/resolver": {


tests/server/disco/TestViews.js
```diff
@@ -51,11 +54,14 @@ describe('Discovery Pane GET requests', () => {
     .then((res) => {
       assert.equal(res.header.location,
         '/en-US/firefox/discovery/pane/48.0/Darwin/normal');
+      return true;
     }));
 
   it('should set an HSTS header', () => request(app)
```
muffinresearch: This is using implicit returns - is the extra return value still needed in the then if it was an explicit return? 

This is a separate problem but the more I think about it we should look to standardise on explicit returns and never use the the implicit arrow return style.

tests/client/core/test_utils.js
```diff
@@ -516,7 +512,7 @@ describe('loadCategoriesIfNeeded', () => {
     return loadCategoriesIfNeeded(props).then(() => {
       assert(dispatch.calledWith(action), 'dispatch not called');
       mockApi.verify();
-      mockActions.verify();
+      return mockActions.verify();
```
muffinresearch: Seems a bit weird to return the last one of multiple calls. Is this needed to pass lint? If so perhaps this one should be excluded instead.

muffinresearch:
I'd be interested to hear other opinions but in general I think it would make more sense to do an exclusion for cases that don't require a return value rather than add an arbitrary return value of either `true` or just returning the last call of several.

At least if this rule flags things it will make everyone think about the case that's right for the context and they can then choose to add the return value if required or add an exclusion to tell eslint for that case we know what we're doing.

tests/client/core/test_utils.js
tofumatt: Yeah we could `return true` instead but a return is always required for any Promise.

tofumatt:
While I'm fine with the idea of just having a lot of exclusions when they don't make sense *in theory*, in practice it's more work to remember the eslint rule and add it in than it is to just always `return true` in a promise.

tests/server/disco/TestViews.js
tofumatt: It is, yeah.

I would love to stop using implicit returns. I *really* dislike their syntax and I also dislike that they make our code feel less homogeneous, especially because the eslint rule *requires* them if they can be used.

tests/client/core/test_utils.js
muffinresearch: Does this eslint rule enforce the return of the original promise *and* returns in the `then()`?

tests/client/amo/components/TestAddonReviewList.js
```diff
@@ -164,6 +164,7 @@ describe('amo/components/AddonReviewList', () => {
           assert.ok(dispatch.called);
           const expectedAction = setAddonReviews({ addonSlug, reviews });
           assert.deepEqual(dispatch.firstCall.args[0], expectedAction);
+          return true;
```
kumar303: returning `true` is going to get really annoying. I say we remove this rule. Instead of extending `plugin:promise/recommended` can you just add what we need?

tests/client/amo/testApi.js
```diff
@@ -216,9 +211,9 @@ describe('amo.api', () => {
         .returns(Promise.resolve({ results: [] }));
 
       return getLatestUserReview()
-        .then(unexpectedSuccess, (error) => {
-          assert.match(error.message, /user and addon must be specified/);
-        });
+        .then(unexpectedSuccess, (error) =>
```
kumar303: I'd like to see braces around a statement like this because it's a common need to add more assertions later. If you remove the return rule you could remove the braces.

tofumatt: I agree, I would really like to ditch implicit returns everywhere actually. The rule is annoying about it 😒

tofumatt:
I agree that these changes are too much and the `return true;` is annoying, but I wanted to see what they'd look like and get others' opinion on it. I'll see about a less prescriptive rule set that catches issues like we had in the test suite without forcing so much extra code.

The rule didn't really highlight any bad patterns in our code (the only nested promises were around variable scoping and make sense. So that's good! 👍

tofumatt:
Hi @kumar303,

Okay, obviously this could've been simpler. It will now fail if a test does not return a promise, tested locally to fail but obviously feel free to double-check if you like 😄

Issue #2074, opened 2017-03-15
Add-ons cannot be installed from Disco Pane by clicking on the switch button


ValentinaPC:
Prerequisites:
- `extensions.webservice.discoverURL` pref is set to point stage/dev

Steps to reproduce:
1. Load about:addons page
2. Try to install an add-on by clicking on the switch button

Expected results:
Add-on is installed.

Actual results:
Add-on cannot be installed.

Notes/Issues:
- add-on can be installed from AMO public pages (i.e https://addons.allizom.org/en-US/firefox/addon/video-downloadhelper/)
- add-ons can be installed from search results pages in about:addons

Verified on FF55(Win 7). Issue is reproducing in AMO-stage and dev.
Video for this issue:
![installdiscopane](https://cloud.githubusercontent.com/assets/15685960/23954086/0b0122a4-099f-11e7-834c-158fe50b41e7.gif)

muffinresearch:
This is caused by the change to disable buttons if the addon is seen to be incompatible. According to the mobile page result for that add-on (on stage [1]) it's incompatible with 54.

[1] https://addons.allizom.org/en-US/firefox/addon/video-downloadhelper/

![add-ons_for_firefox](https://cloud.githubusercontent.com/assets/1514/23960318/8c76645c-099f-11e7-8d90-e6b913b79bce.png)

muffinresearch:
This could be a release blocker if in prod this is going to cause buttons to be disabled in the disco pane when we release.

tofumatt:
Taking a look now.

tofumatt:
For 54 everything seems fine, but Nightly is incompatible with https://addons.mozilla.org/en-GB/firefox/addon/youtube-high-definition/

<img width="110" alt="screenshot 2017-03-15 20 23 13" src="https://cloud.githubusercontent.com/assets/90871/23969177/681b4c94-09bd-11e7-9bb5-4e5aa9794a19.png">

So we should add a message for incompatible add-ons in the disco pane... but we should probably change the add-ons in the disco pane. I think we missed an update but it would be worth cherry-picking into the tag or Nightly will get a greyed-out add-on.

mstriemer:
I thought we only took the minimum version into account in the past? It shows as installable on the desktop site in Nightly.

mstriemer:
(Otherwise every add-on would need a new release for each new Firefox version)

tofumatt:
[...]

PR #2078, opened 2017-03-16
Revert compat changes

muffinresearch:


tofumatt:
Wait, actually. Hold up on merging this because https://github.com/mozilla/addons-frontend/issues/2074#issuecomment-287029732 seems to indicate just ignoring max version will do the trick. That would be an easier patch to land and I think much better.

PR #2086, opened 2017-03-16
Strip indent from extract and gettext

### LINKED ISSUE #2059: Some strings will not be translated due to the use of oneLine

muffinresearch:
Fixes #2059

I've now updated the fork of babel-gettext-extractor [1] after a bit of a clean-up and overhaul adding more tests and fixing up the lint.

This branch updates to the rev of that lib with the new features this now supports removing leading indents from extracted values. 

This patch also wraps our Jed instance functions to support stripping the same indent from keys because otherwise the lookups would fail. This means from now on any multiline translations should use untagged template literals and not use dedent or oneLine. 

I'll also add a commit to this to update the docs so it's clear and we should also look to add a eslint rule to check for tagged template literals as args to any Jed gettext methods.

Lastly to ensure tests see the actual Jed output this changes the FakeJed to be a spy wrapped real Jed. Otherwise several tests look at translation strings broke due to leading whitespace in the text.

[1]  https://github.com/muffinresearch/babel-gettext-extractor/commits/master

## Note

After landing we'll need to extract and merge both amo and the disco pane so that the new strings currently missing can be translated.

### DIFF
--- a/src/core/components/ErrorPage/NotFound/index.js
+++ b/src/core/components/ErrorPage/NotFound/index.js
@@ -1,5 +1,4 @@
 import React, { PropTypes } from 'react';
-import { oneLine } from 'common-tags';
 import { compose } from 'redux';
 import NestedStatus from 'react-nested-status';
 
@@ -24,7 +23,7 @@ export class NotFoundBase extends React.Component {
         <div className="ErrorPage NotFound">
           <h1>{i18n.gettext('Page not found')}</h1>
           <p>
-            {i18n.gettext(oneLine`
+            {i18n.gettext(`
               Sorry, but we can't find anything at the URL you entered.`)}
           </p>
 
--- a/src/core/components/ErrorPage/NotFound/index.js
+++ b/src/core/components/ErrorPage/NotFound/index.js
@@ -23,8 +23,7 @@ export class NotFoundBase extends React.Component {
         <div className="ErrorPage NotFound">
           <h1>{i18n.gettext('Page not found')}</h1>
           <p>
-            {i18n.gettext(`
-              Sorry, but we can't find anything at the URL you entered.`)}
+            {i18n.gettext("Sorry, but we can't find anything at the URL you entered.")}
           </p>
 
           <p>
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -181,6 +181,17 @@ export function makeMomentLocale(locale) {
   return locale.replace('_', '-').toLowerCase();
 }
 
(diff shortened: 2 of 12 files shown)


src/core/i18n/utils.js
```diff
@@ -181,6 +181,16 @@ export function makeMomentLocale(locale) {
   return locale.replace('_', '-').toLowerCase();
 }
 
+// Functionality based on oneLine form declandewet/common-tags https://goo.gl/4PzaJI
+// If this function is altered muffinresearch/babel-gettext-extractor also needs
```
tofumatt: Nitpick but I think you should link to the URL of the repo here, not just the name. Makes it easier to scan and a bit more obvious (I did a double-take when reading it!)

src/core/i18n/utils.js
```diff
@@ -181,6 +181,16 @@ export function makeMomentLocale(locale) {
...
+// to be updated.
+function oneLineTranslationString(str) {
```
tofumatt: I'm guessing this was copy/pasted and we can't use `oneLine` directly.

Can we change the variable name to `string` or `i18nString`? I'm just wary of no vowel variables.

src/core/components/ErrorPage/NotFound/index.js
```diff
@@ -24,7 +23,7 @@ export class NotFoundBase extends React.Component {
         <div className="ErrorPage NotFound">
           <h1>{i18n.gettext('Page not found')}</h1>
           <p>
-            {i18n.gettext(oneLine`
+            {i18n.gettext(`
               Sorry, but we can't find anything at the URL you entered.`)}
```
tofumatt: Nitpick but this looks like it didn't even need `oneLine`. Oops!

muffinresearch: Heh, yeah I'll fix.

muffinresearch:
> I think we can also remove the dedent preprocessing we had in the babel config now, right?

Yeah it's being used in a few other places, and it still needs to be done for disco too.

src/core/i18n/utils.js
muffinresearch: Yep, np

PR #2097, opened 2017-03-18
feat: Show compatibility errors in Disco Pane (fix #2095)

### LINKED ISSUE #2095: Show compatibility errors in Disco Pane
We don't show disabled add-on reasoning in the Disco pane; just a disabled button/switch. It wouldn't hurt to at least add a little message to it so users encountering the page know why an add-on is disabled. This came up in #2074, and while those add-ons shouldn't have been disabled, a message would have made it more obvious 😄 

There's no UX for this so I'm implementing something basic and dumb–largely this should never happen, but the message is better than no message.

tofumatt:
Working on showing errors in the disco pane; tests are failing right now and it's low-priority so just leaving here now but will revisit later.

### DIFF
--- a/src/disco/components/AddonCompatibilityError/index.js
+++ b/src/disco/components/AddonCompatibilityError/index.js
@@ -1,0 +1,68 @@
+/* eslint-disable react/no-danger */
+import React, { PropTypes } from 'react';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+
+import {
+  INCOMPATIBLE_FIREFOX_FOR_IOS,
+  INCOMPATIBLE_UNDER_MIN_VERSION,
+} from 'core/constants';
+import translate from 'core/i18n/translate';
+import { sanitizeHTML } from 'core/utils';
+
+import './style.scss';
+
+
+// Messages in the disco pane are a bit less specific as we don't care about
+// non-Firefox clients and the copy space is limited.
+export class AddonCompatibilityErrorBase extends React.Component {
+  static propTypes = {
+    i18n: PropTypes.object.isRequired,
+    minVersion: PropTypes.string.isRequired,
+    reason: PropTypes.string.isRequired,
+  }
+
+  render() {
+    const {
+      i18n,
+      minVersion,
+      reason,
+    } = this.props;
+    let message;
+
+    if (typeof reason === 'undefined') {
+      throw new Error('AddonCompatibilityError requires a "reason" prop');
+    }
+    if (typeof minVersion === 'undefined') {
+      throw new Error('minVersion is required; it cannot be undefined');
+    }
+
+    if (reason === INCOMPATIBLE_FIREFOX_FOR_IOS) {
+      message = i18n.gettext(
+        'Firefox for iOS does not currently support add-ons.');
+    } else if (reason === INCOMPATIBLE_UNDER_MIN_VERSION) {
+      message = i18n.gettext(
+        'This add-on does not support your version of Firefox.');
+    } else {
+      // Unknown reasons are fine on the Disco Pane because we don't
+      // care about non-FF clients.
+      message = i18n.gettext('This add-on does not support your browser.');
+    }
+
+    return (
+      <div
+        className="AddonCompatibilityError"
+        dangerouslySetInnerHTML={sanitizeHTML(message, ['a'])}
+      />
+    );
+  }
+}
+
+export function mapStateToProps(state) {
+  return { lang: state.api.lang };
+}
+
+export default compose(
+  connect(mapStateToProps),
+  translate({ withRef: true }),
+)(AddonCompatibilityErrorBase);
--- a/src/disco/components/AddonCompatibilityError/style.scss
+++ b/src/disco/components/AddonCompatibilityError/style.scss
@@ -1,0 +1,16 @@
+@import "~ui/css/vars";
+
+.AddonCompatibilityError {
+  background: $report-base-color;
+  font-size: 12px;
+  margin: 0;
+  padding: 5px;
+  text-align: center;
+  width: 100%;
+
+  &,
+  a,
+  a:link {
+    color: $white;
+  }
+}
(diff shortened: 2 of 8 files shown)


muffinresearch:
@tofumatt - this looks to have bit-rotted, can it be updated and completed?

muffinresearch:
sad_keanu

tofumatt:
Alright, updated and fixed everything, here are some screenshots:

### Small screen
screenshot 2017-07-08 02 45 11

### Desktop
screenshot 2017-07-08 02 44 46
screenshot 2017-07-08 02 44 48

src/disco/components/AddonCompatibilityError/style.scss
```diff
@@ -0,0 +1,16 @@
...
+  background: $report-base-color;
+  font-size: 12px;
+  margin: 0;
+  padding: 5px;
+  text-align: center;
+  width: 100%;
+
+  &,
+  a,
+  a:link {
```
muffinresearch: Any other states needed e.g: visited, hover, focus, active  etc or are the defaults ok?

src/amo/components/AddonReviewList.js
```diff
@@ -120,6 +120,7 @@ export class AddonReviewListBase extends React.Component {
           { authorName: review.userName, timestamp });
 
       const reviewBodySanitized = sanitizeHTML(nl2br(review.body), ['br']);
```
muffinresearch: Is this intermediate `const` needed?

src/disco/components/AddonCompatibilityError/index.js
```diff
@@ -0,0 +1,87 @@
...
+    let message;
+
+    if (typeof reason === 'undefined') {
+      throw new Error('AddonCompatibilityError requires a "reason" prop');
+    }
+    if (typeof minVersion === 'undefined') {
+      throw new Error('minVersion is required; it cannot be undefined');
+    }
+
+    if (reason === INCOMPATIBLE_NOT_FIREFOX) {
```
muffinresearch: Should never really happen since non non-FF browsers should ever see the disco pane.

tests/unit/disco/components/TestAddon.js
```diff
@@ -319,8 +359,9 @@ describe('<Addon />', () => {
       themeImage = findDOMNode(root).querySelector('.theme-image');
       const preventDefault = sinon.spy();
       Simulate.click(themeImage, { preventDefault });
+
       expect(preventDefault.called).toBeTruthy();
-      expect(installTheme.calledWith(themeImage, data.addon)).toBeTruthy();
+      sinon.assert.calledWith(installTheme, themeImage);
```
muffinresearch: is `data.addon` no longer passed?

muffinresearch:
This is probably a @pwalm question but I wonder if the red could be dropped back a bit? The sea of red for a non-FF browser might be a bit much.

tests/unit/disco/components/TestAddon.js
```diff
@@ -143,22 +141,31 @@ describe('<Addon />', () => {
...
+
+      expect(findDOMNode(root).querySelector('.notification.restart'))
+        .toBeFalsy();
     });
 
     it('renders the heading', () => {
       const root = renderAddon({ addon: result, ...result });
-      expect(root.heading.textContent).toContain('test-heading');
+
+      expect(findDOMNode(root).querySelector('.heading').textContent)
```
muffinresearch: Iirc `findDOMNode` was planned to be deprecated see https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-find-dom-node.md

See also https://facebook.github.io/react/docs/react-dom.html#finddomnode re: use being discouraged.

tests/unit/disco/components/TestAddonCompatibilityError.js
```diff
@@ -0,0 +1,92 @@
...
+    const link = root.find('.AddonCompatibilityError').render().find('a');
+
+    expect(link.attr('href')).toEqual('https://www.mozilla.org/es/firefox/');
+    expect(root.render().text()).toEqual(
+      'You need to download Firefox to install this add-on.');
+  });
+
+  it('renders a notice for old versions of Firefox', () => {
+    const root = render({
+      lang: 'fr',
```
muffinresearch: Seems a little odd to pass lang and then be looking for english translations.

src/amo/components/AddonReviewList.js
tofumatt: It tidies up the next lines and makes it a bit more readable. I went back and forth on it and this was the nicest version.

src/disco/components/AddonCompatibilityError/index.js
tofumatt: True, but they could hit the URL and it doesn't hurt to have nice messages. Could also happen if the UA was weird and might be useful in a bug report...

That said it increases code size. If you want, I can remove this if we care a lot less about other clients in the Disco Pane, which I guess we do.

tests/unit/disco/components/TestAddon.js
tofumatt: It is, and we'll be moving to enzyme, but I didn't want to refactor these tests at the time. I could do it now if you like but this isn't our only usage of `findDomNode` 😄

tests/unit/disco/components/TestAddonCompatibilityError.js
tofumatt: Oh that's true, I'll tweak that.

src/disco/components/AddonCompatibilityError/style.scss
tofumatt: The defaults are fine, this just keeps the colour white which it will stay for those other states too.

tests/unit/disco/components/TestAddon.js
tofumatt: This was actually a case of "how did this test ever work?" since updating the `withInstallButton` HOC. I fixed it up using enzyme... but we should definitely clean up this test suite in general...

pwalm:
The red is a bit much, isn't it. I would suggest we put that "You need to..." notification between the intro copy and the first add-on, and remove or make inactive the "Add to Firefox"/"Install Theme" buttons.

PR #2100, opened 2017-03-19
fix: Layout of themes and extensions home

mstriemer:
This is #2024 but rebased with master to remove the extra commits and fix a test failure on master.

cc @aayushsanghavi 

screen shot 2017-03-22 at 11 37 21 screen shot 2017-03-22 at 11 37 31


aayushsanghavi:
@mstriemer thank you so much for this. So with this up, I guess we should close #2024 and ask someone to review this PR. Also, the changes you've made in your last commit look good, thanks again.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +112,43 @@ export class LandingPageBase extends React.Component {
     }
 
     const { addonType, html } = content;
+    const themeText = i18n.gettext(
+      oneLine`Change your browser's appearance. Choose from thousands of themes to give Firefox the
```
tofumatt: We no longer use templates string tags in `i18n.gettext`, so we can remove `oneLine` here and anywhere else.

src/amo/components/LandingPage.js
```diff
@@ -20,6 +21,8 @@ import {
   visibleAddonType as getVisibleAddonType,
 } from 'core/utils';
 import translate from 'core/i18n/translate';
+import Icon from 'ui/components/Icon/index';
+import Button from 'ui/components/Button/index';
```
tofumatt: Very pedantic here, but we could alphabetise these 😄

src/amo/components/LandingPage.js
```diff
@@ -108,9 +112,43 @@ export class LandingPageBase extends React.Component {
...
+    const themeText = i18n.gettext(
+      oneLine`Change your browser's appearance. Choose from thousands of themes to give Firefox the
+      look you want.`);
+    const extensionsText = i18n.gettext(
+      oneLine`Install powerful tools that make browsing faster and safer, add-ons make your browser
+      yours.`);
 
     return (
       <div className={classNames('LandingPage', `LandingPage-${addonType}`)}>
+
```
tofumatt: This newline doesn't seem needed.

src/amo/components/LandingPage.js
```diff
@@ -108,9 +112,43 @@ export class LandingPageBase extends React.Component {
...
+        <div className="LandingPage-header">
+          <div className="LandingPage-header-top">
+            <Icon name={visibleAddonType} />
+            <div className="LandingPage-header-text">
+              <h1 className="LandingPage-heading">
+                {addonType === ADDON_TYPE_THEME
```
tofumatt: This means anything not a theme gets output as an extension but that's not 100% true. If this were a dictionary or open search plugin it would say extension. I guess that's kinda how we treat them, but I'd rather see a mapping for this similar to what we do elsewhere so it's "Theme"/"Extension"/"Add-on" if unknown.

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,58 @@
...
+
+.LandingPage-header-top {
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  margin-bottom: $padding-page * 2;
+}
+
+.LandingPage-header-text {
+  @include margin-start(20px);
```
tofumatt: 👍

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,58 @@
...
+
+  display: inline-block;
+  max-width: 350px;
```
tofumatt: How does this look on tablets/desktop/larger screens? Where does this number come from?

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,58 @@
...
+}
+
+.LandingPage-heading {
+  font-size: 14px;
```
tofumatt: These should be using variables (preferably the font size variables we already have).

src/amo/components/LandingPage.scss
```diff
@@ -1,5 +1,58 @@
...
+
+.LandingPage-browse-button,
+.LandingPage-browse-button:link {
+  align-items: center;
+  display: flex;
+  font-size: 14px;
+  font-variant: all-small-caps;
+  justify-content: center;
+  margin: 0 $padding-page;
+  padding: 1em;
```
tofumatt: We don't use `em`  most of the time, so this seems out-of-place. I'd rather stick to the same units everywhere, `px` here would be better I think.

src/ui/components/Icon/styles.scss
```diff
@@ -37,3 +38,26 @@
...
+  background-size: cover;
+  height: 20px;
+  width: 20px;
+}
+
+.Icon-themes {
+  background: url('./themes.svg') center no-repeat;
+  background-size: cover;
+  height: 50px;
+  min-width: 41.5px;
```
tofumatt: These values (and the ones below) seem very strange. Should the SVG have better sizes or defaults?

aayushsanghavi: @tofumatt I had adjusted these values to preserve the aspect ratio of the original icons. Also, I fixed one parameter to 50px (49.5px I guess for the extensions icon) and adjusted the other parameter accordingly because 50px seemed about right after being rendered on the screen.

aayushsanghavi:
Thanks for the review @tofumatt!

src/amo/components/LandingPage.js
mstriemer: It looks like we don't raise lint errors for lines that have strings in them over 100 columns? I don't get how that works but I guess that makes this easier.

src/amo/components/LandingPage.scss
mstriemer: Here it is at 800x600. I just made up the number for something that seemed like a reasonable width to wrap at.

screen shot 2017-03-22 at 18 26 35

src/ui/components/Icon/styles.scss
mstriemer: The extensions icon is 139x134 and the themes icon is 112x135. I think generally the values we get from the mocks are 3x bigger than what we want and those are pretty closed to what these are so I just divided by 3 in each dimension. The sizes are slightly smaller now.

mstriemer:
I think all the comments have been addressed.

tofumatt:
Awesome, yes, sorry I've just been at other stuff during the work week in London. r+, I'll squash and merge if everyone's cool with that.

aayushsanghavi:
@tofumatt thanks a lot for merging it!

PR #2102, opened 2017-03-19
feat: Metadata header tweaks (fix #1888)

### LINKED ISSUE #1888: Improve ratings and user count meta data
From @pwalm's comment in https://github.com/mozilla/addons-frontend/issues/1698#issuecomment-282094051 :

Proposed:
screen shot 2017-02-23 at 11 24 20 am
We kill the icons and just use text for users/reviews. We'll run into less errors with long l10n translations.

tofumatt:
Fixes #1888.

This adds half-stars to the `Ratings` component when `readOnly` is set to `true`. I thought this was useful because it adds a bit more precision to add-on review rating averages.

The `Rating` component still renders only full-step for reviews.

Screenshots:

### Mobile

screen shot 2017-03-19 at 23 25 01
screen shot 2017-03-19 at 23 24 54
screen shot 2017-03-19 at 23 24 47
screen shot 2017-03-19 at 23 24 42

### Desktop

screenshot 2017-03-19 23 26 53
screenshot 2017-03-19 23 26 45


src/ui/components/Rating/index.js
```diff
@@ -46,6 +51,34 @@ export class RatingBase extends React.Component {
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
+    // We render read-only star ratings with half-stars, to give users a
+    // better representation of the average rating.
+    if (readOnly) {
```
tofumatt: I went through this a few times making it work with tests and trying to make it as concise as possible, but I dunno if it's that accessible. It's weird code that write. Happy to hear about ways it would be considered more accessible; it works but it's not lovely.

kumar303: Can you not merge this into the non-read-only code? There is a lot of repetition here which will make it hard to maintain.

Also, I think half-stars should also be rendered in non-read-only mode.

tofumatt:
Added half-star tests and made it so the code paths are closer. As discussed in-person, only `readOnly` reviews show half-stars because averages can be floats but ratings are integers in five steps.

src/ui/components/Rating/index.js
```diff
@@ -50,16 +55,19 @@ export class RatingBase extends React.Component {
       const props = {
         className: classNames('Rating-choice', {
           'Rating-selected-star': rating && thisRating <= rating,
+          // Half stars are the result of averages rounded to the nearest
+          // 0.5 place; only show them when readOnly is true.
+          'Rating-half-star': (readOnly && rating && thisRating > rating &&
```
tofumatt: We don't really need to round it anymore because if rating was `3.6` and we're on `rating === 4` then `thisRating (4) - 0.5 <= rating (3.6)` is true. So we don't have to round anywhere here.

tofumatt: The rating we get from the API is a `float`... if we round the rating to an integer then we can't check for half stars.

tests/client/ui/components/TestRating.js
```diff
@@ -103,11 +103,31 @@ describe('ui/components/Rating', () => {
     });
   });
 
-  it('rounds down average ratings to an integer', () => {
+  it('rounds readOnly average ratings to nearest 0.5 multiple', () => {
+    // This should be treated like a rating of 3.
```
tofumatt: Correct, bad copypasta, I'll fix.

tofumatt:
Tweaked the test comment; I think the other code is correct but if it's not obvious let me know.

src/ui/components/Rating/index.js
kumar303: ok, I think I see

kumar303: I didn't realize it was a float but I think it would still work in non-read only mode. Anyway, yeah, it's not important since I doubt we'll ever let people submit half-star ratings.

tofumatt:
Regarding the `readOnly` bit... you were right, no reason to bother checking for it. Removed that check, because the API should only send integers to non-`readOnly` instances we use this component.

Issue #2062, opened 2017-03-14
Change client / server logger names for filtering in Sentry


kumar303:
Both our server side error logger and client side logger is named `javascript` because that's the default. Let's [change](https://docs.sentry.io/clients/javascript/config/) them so that each one has unique names like `client-js` and `server-js` (or whatever). This will allow us to [filter the events](https://sentry.prod.mozaws.net/operations/addons-frontend-amo-prod/?query=is%3Aunresolved%20logger%3Ajavascript).

Client is configured here: https://github.com/mozilla/addons-frontend/blob/master/src/core/client/base.js#L22

Server is configured here: https://github.com/mozilla/addons-frontend/blob/master/src/core/server/base.js#L140

PR #2109, opened 2017-03-20
Configure Sentry logger for filtering

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2062


kumar303:
I tested it locally, seems to be working:

https://sentry.prod.mozaws.net/operations/addons-frontend-test/?query=logger%3A%22client-js%22
https://sentry.prod.mozaws.net/operations/addons-frontend-test/?query=logger%3A%22server-js%22

PR #2124, opened 2017-03-22
Build locales in build script

muffinresearch:
This removes all the built locale JS from the tree and instead builds it as part of the build step.


tofumatt:
more-red-than-green
red-celebrate

bin/build-locales
```diff
@@ -15,65 +16,70 @@ const chalk = require('chalk');
 const toSource = require('tosource');
 const oneLine = require('common-tags').oneLine;
 
-const appName = config.get('appName');
+const app = config.get('appName');
+const buildAll = process.env.ADDONS_FRONTEND_BUILD_ALL === '1';
```
tofumatt: Should this just be any value? `true` would be nicer to use on the command line... if that would ever get used locally?

tofumatt: I see what this is... I feel like accepting `true` as well might be nice 😄

tofumatt: (But also a comment explaining what this does exactly would be handy. It took me a second to grok.

tofumatt:
Oh, this fails servertests though because of `let` usage.

PR #2125, opened 2017-03-22
Allow non-ints for readOnly (#1888)

tofumatt:
Sorry I said I tweaked this locally and forgot to push. Just doesn't bother making a difference between `readOnly` and not `<Ratings>` for half-stars.


tofumatt:
@kumar303 requested this change and we agreed to it but I forgot to merge it in... so i'm just gonna do it here, hope that's okay 😄

kumar303:
No problem, thanks for the follow-up

Issue #2123, opened 2017-03-22
Support non-JWT authentication tokens for login/logout


kumar303:
Steps to test this change:
* log in
* log out
* navigate the site and make sure you're still logged in

The session authentication token will no longer be a JWT as of https://github.com/mozilla/addons-server/issues/4907 so the frontend code needs to be adjusted accordingly.

ValentinaPC:
> Steps to test this change:
> 
>     log in
>     log out
>     navigate the site and make sure you're still logged in

"make sure that we are still logged in", after log out?
Should we check this on mobile pages?
Thanks!

tofumatt:
Yes, this change was for the mobile pages so it should be checked on them :-)

ValentinaPC:
I like your sens of humor, but, sorry to insist, not sure I follow - what part of "log out and make sure you are still logged in" is normal? :)

tofumatt:
Ahaha okay I see. Sorry!

Each of those steps are separate, not one after the other. So just make sure login works, logout works, and that after login you can navigate the site and stay logged in. Sorry! 😄

kumar303:
Hah! My bad, I can see how that was really unclear now. I just updated the description.

ValentinaPC:
It was fun, nevertheless. Thanks for clarification! Will check :)

ValentinaPC:
Verified as fixed on AMO-dev FF52(Android 6.0.1).
I've navigated through mobile pages, being logged in or out, or tried to make actions that require to be logged in, while logged out and everything behaves as expected.
Note: A video showing this would be too long.

PR #2129, opened 2017-03-22
Supports the new API auth scheme which switched away from JWTs

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2123

This needs to land in sync with https://github.com/mozilla/addons-server/pull/4960

### DIFF
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -22,7 +22,7 @@ import { createApiError } from 'core/api';
 import ServerHtml from 'core/containers/ServerHtml';
 import { prefixMiddleWare, trailingSlashesMiddleware } from 'core/middleware';
 import { convertBoolean } from 'core/utils';
-import { setClientApp, setLang, setJwt, setUserAgent } from 'core/actions';
+import { setClientApp, setLang, setAuthToken, setUserAgent } from 'core/actions';
 import log from 'core/logger';
 import {
   getDirection,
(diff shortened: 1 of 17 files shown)


src/core/server/base.js
```diff
@@ -22,7 +22,7 @@ import { createApiError } from 'core/api';
 import ServerHtml from 'core/containers/ServerHtml';
 import { prefixMiddleWare, trailingSlashesMiddleware } from 'core/middleware';
 import { convertBoolean } from 'core/utils';
-import { setClientApp, setLang, setJwt, setUserAgent } from 'core/actions';
+import { setClientApp, setLang, setAuthToken, setUserAgent } from 'core/actions';
```
tofumatt: Nitpick but we tend to alphabetise imports. But like, only fix if you need to fix something else too.

tests/client/core/api/test_api.js
```diff
@@ -446,7 +446,7 @@ describe('api', () => {
   });
 
   describe('login', () => {
-    const response = { token: 'use.this.jwt' };
+    const response = { token: 'use:this:auth-token' };
```
tofumatt: This token is different from the others, why is that?

tests/client/core/reducers/test_api.js
```diff
@@ -11,15 +11,16 @@ describe('api reducer', () => {
...
+    assert.deepEqual(
+      api({ foo: 'bar' }, actions.setAuthToken(token)),
+      { foo: 'bar', token });
   });
 
-  it('clears the JWT on log out', () => {
+  it('clears the auth token on log out', () => {
     assert.deepEqual(
-      api({ lang: 'fr', clientApp: 'firefox', token: 'secret' },
+      api({ lang: 'fr', clientApp: 'firefox', token: 'user:auth:token' },
```
tofumatt: Dunno if using `signedInApiState` here makes more sense, but I guess it doesn't buy us anything. Just consistency, really.

kumar303: Not sure how I missed that, thanks

PR #2133, opened 2017-03-22
Use fastclick

### LINKED ISSUE #1883: Consider using fastclick or similar.
I notice when using an android device sometimes clicks can be hard to activate  (e.g. the logout button) and having to hit it multiple times feels clunky.

We should at implementing some kind of fast click support e.g: https://www.npmjs.com/package/react-fastclick

tofumatt:
Fix #1883.


tofumatt:
Just managed to test this locally and it seems to improve touch/click speeds. r?

src/core/client/base.js
```diff
@@ -16,6 +17,8 @@ import log from 'core/logger';
 
 
 export default function makeClient(routes, createStore) {
+  FastClick.attach(document.body);
```
kumar303: I would like to see this added after the Sentry logging so that we can get reports of FastClick errors :)

While you're in here, could you put a comment above the Sentry configuration saying something like "This should come before any other code to catch and report errors" ?

tofumatt:
Cool, updated.

PR #2141, opened 2017-03-23
fix: Don't play the disco pane video in tests (fix #765)

### LINKED ISSUE #765: Prevent Disco pane video test from creating a memory leak, crashing
If you run the tests with `npm run unittest:dev` which runs the unit tests with a watcher, eventually node runs out of memory. 

~~When running the tests in debug mode I noticed the video is actually being played in the disco pane tests. We should fix that.~~ (This is fixed but the memory leak is still happening.)

tofumatt:
This fixes #765, so our tests shouldn't have memory leaks and also shouldn't spook me when I'm testing with loud volume settings late at night.

### DIFF
--- a/src/disco/containers/DiscoPane.js
+++ b/src/disco/containers/DiscoPane.js
@@ -59,14 +59,12 @@ export class DiscoPaneBase extends React.Component {
   }
 
   showVideo = (e) => {
-    const { _tracking, _video } = this.props;
-    if (_video) {
-      this.video = _video;
-    }
+    const { _tracking } = this.props;
+    const _video = this.props._video || this.video;
 
     e.preventDefault();
     this.setState({ showVideo: true });
-    this.video.play();
+    _video.play();
     _tracking.sendEvent({
       action: 'play',
       category: VIDEO_CATEGORY,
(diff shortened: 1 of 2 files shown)


src/disco/containers/DiscoPane.js
```diff
@@ -34,13 +34,17 @@ export class DiscoPaneBase extends React.Component {
     results: PropTypes.arrayOf(PropTypes.object).isRequired,
     _addChangeListeners: PropTypes.func,
     _tracking: PropTypes.object,
+    // We allow the video to be overloaded for testing, see:
+    // https://github.com/mozilla/addons-frontend/issues/765
```
kumar303: I don't think it's necessary to add a comment. This is a common pattern that we use elsewhere. The properties above it have defaults already too.

src/disco/containers/DiscoPane.js
```diff
@@ -55,7 +59,11 @@ export class DiscoPaneBase extends React.Component {
   }
 
   showVideo = (e) => {
-    const { _tracking } = this.props;
+    const { _tracking, _video } = this.props;
+    if (_video) {
+      this.video = _video;
```
kumar303: Re-assigning `this.video` makes me nervous because refs are kind of magical. I think we need to get this working:
````js
const { _video } = this.props._video || this.video;
````

tofumatt:
All fixed, ready for another look.

PR #2166, opened 2017-03-23
feat: Add a Loading bar to the site

### LINKED ISSUE #1983: Implement the UX for navigation and data loading
In order to smooth out some of the kinks in the way navigation works in the mobile pages it would be good to define the way we want navigation to work so that it feels intuitive to the user.

@pwalm do you have thoughts on how navigating between pages should work are we just aiming for making it work as a traditional site would or should we be looking to aim for something that provides some kind of transition?

It would be good to cover general page navigation and how loading indicators should be surfaced. If there are specific interactions that aren't yet covered we could add details for those here too.

Related issues:

* Navigation flow is strange https://github.com/mozilla/addons-frontend/issues/1953 
* Scroll position is not preserved pressing back after exploring non-search lists  https://github.com/mozilla/addons-frontend/issues/1954

tofumatt:
Fixes #1983. Now we have a cool loading bar!

This also removes the old spinning fox loading indicator we had because this indicator is global.

### Video

mar-23-2017 19-31-55


package.json
```diff
@@ -195,6 +196,7 @@
     "mozilla-version-comparator": "1.0.2",
     "normalize.css": "5.0.0",
     "normalizr": "3.2.2",
+    "npm": "4.4.1",
```
muffinresearch: Was this intended?

package.json
```diff
@@ -187,6 +187,7 @@
     "fastclick": "1.0.6",
     "helmet": "3.5.0",
     "humps": "2.0.0",
+    "i": "0.3.5",
```
muffinresearch: Where is this used?

tofumatt: Nope!

tofumatt:
Okay, should be ready for r? but Travis is being weird so I'll see about clearing its caches and getting `yarn` to run on it...

kumar303:
Patching `callApi()` could also be done in a future patch since it might get a little bit hairy.

tofumatt:
Yeah I'd rather the `callApi` bit happened separately as this doesn't break any of it and adds a lot to the UX of the site. I've filed #2179.

PR #2171, opened 2017-03-24
fix: Travis yarn install

tofumatt:
Travis `yarn --frozen-lockfile` installs were failing, seeing if this fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ cache:
   directories:
   - node_modules
 before_install:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,7 +18,7 @@ cache:
   directories:
   - node_modules
 before_install:
-- npm install -g npm@3
+- npm install -g yarn
```
kumar303: I think it would be helpful to add `yarn --version` after this

PR #2177, opened 2017-03-24
chore: Remove round package (close #2173)

### LINKED ISSUE #2173: Remove round
Investigate removing the `round` dep and replacing it with something simpler.

tofumatt:
Removes the `round` package.

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -1,7 +1,6 @@
 import classNames from 'classnames';
 import React, { PropTypes } from 'react';
 import { compose } from 'redux';
-import round from 'round';
 
 import log from 'core/logger';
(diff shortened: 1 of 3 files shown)


src/ui/components/Rating/index.js
```diff
@@ -89,8 +88,9 @@ export class RatingBase extends React.Component {
     }
     let description;
     if (rating) {
+      const roundedRating = Math.round(rating * 2) / 2;
```
kumar303: oh, right, `Math.round()` doesn't let you specify precision. The division hack is confusing. 

How about this instead?
````js
{ rating: i18n.formatNumber(rating.toFixed(1)) }
````

tofumatt:
Changed to `toFixed`. 👍

Issue #2061, opened 2017-03-14
Make 'Error calling API' more useful in Sentry


kumar303:
When an API error is thrown, we get a report in Sentry [like this](https://sentry.prod.mozaws.net/operations/addons-frontend-amo-prod/issues/385976/) but it's not so useful.

Here are some suggestions to improve it:

* Add the URL to the message so that at least Sentry makes a new event
* Figure out how to not "filter" the details of the API error. This is what Sentry does to it for some reason:

<img width="898" alt="screenshot 2017-03-14 10 33 54" src="https://cloud.githubusercontent.com/assets/55398/23908445/0cc9da20-08a2-11e7-892a-32a1921f4af4.png">

kumar303:
> Figure out how to not "filter" the details of the API error. 

Ah, this was filtered for [a good reason](https://docs.sentry.io/learn/sensitive-data/) -- I think it contained an authorization related error.

PR #2178, opened 2017-03-24
Make API error message more useful for Sentry

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2061


kumar303:
> This doesn't solve (or seem to solve/mention) the [Filtered] data...

Whoops, I forgot about that part. Thanks. I looked into it and we actually want to keep filtering data like that. More info: https://github.com/mozilla/addons-frontend/issues/2061#issuecomment-289052205

PR #2183, opened 2017-03-25
chore: Remove nb_NO disco locale file

tofumatt:
This snuck into a merge somehow, not sure why. Removing it as we build locales on-the-fly.


kumar303:
Ah! I was wondering about why it keeps getting deleted locally.

PR #2185, opened 2017-03-25
feat: Support OpenSearch plugins (fix #2050)

### LINKED ISSUE #2050: Open search plugins cannot be installed

Relates to this old addons-server issue: https://github.com/mozilla/addons-server/issues/3581

There's no functionality to enable installation of a opensearch add-on in the new mobile pages. 

I'm not sure quite how searches are showing up in the new mobile pages I would have thought the search should just be returning themes and add-ons but they are there if you look for them.

### What happened?

Find an opensearch add-on and try and install it. E.g:  https://addons.mozilla.org/en-US/android/addon/google-dictionary-53151/?src=search

If the fallback button is shown it links to XML as per https://github.com/mozilla/addons-server/issues/3581 -if the addon switch install button is shown it will look like it's installing and then fail.

It should install as normal.

The installation needs to be special-cased to work for opensearch in a similar way to the way it works for lightweight themes as mozAddonManager doesn't handle search installs.

tofumatt:
Fixes #2050.

This allows users to install OpenSearch plugins if their browser is compatible (eg is Firefox). Because OpenSearch installs use their own JS API (`window.external.AddSearchProvider`) I opted not to use the `InstallSwitch` component and always use the button for search because we can't really tell when they're installed and the switch + prompt makes for a weird UX in my opinion.

mar-25-2017 14-22-39


tests/client/amo/components/TestAddonDetail.js
```diff
@@ -313,6 +317,34 @@ describe('AddonDetail', () => {
       rootNode.querySelector('.InstallButton-switch input').disabled);
   });
 
+  it('disables install switch and uses button for OpenSearch plugins', () => {
+    const AddSearchProviderSpy = sinon.stub(window.external,
```
kumar303: `window.external` is a non-standard property so I'd rather see a fake `_window` property passed into the component.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -313,6 +317,34 @@ describe('AddonDetail', () => {
...
+      'AddSearchProvider');
+    const rootNode = renderAsDOMNode({
+      addon: { ...fakeAddon, type: ADDON_TYPE_OPENSEARCH },
+    });
+    const installButton = rootNode.querySelector('.InstallButton-button');
+    assert.equal(installButton.textContent, 'Add to Firefox');
```
kumar303: This logic lives in `<InstallButton>` not `<AddonDetail>` so I think it would make more sense to move these tests to `TestInstallButton`

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -313,6 +317,34 @@ describe('AddonDetail', () => {
...
+
+    Simulate.click(installButton);
+
+    assert.ok(AddSearchProviderSpy.calledWith(fakeAddon.installURL));
+  });
+
+  it('disables the button/does nothing if OpenSearch is not available', () => {
```
kumar303: Same thing here: I think this should move to `TestInstallButton`

tests/client/amo/components/TestAddonDetail.js
tofumatt: Oh, oops, you're totally right. Dunno why I put it here. Will move it!

tests/client/amo/components/TestAddonDetail.js
kumar303: It probably relies on a getter but if you pass in the `window` object like `_window` (or something) then `_window.external` should trigger the getter.

tofumatt:
Sorry, this fell off my radar and I didn't get it fixed. I think I've fixed the issues, ready for another r?

src/core/components/InstallButton/index.js
```diff
@@ -72,6 +84,32 @@ export class InstallButtonBase extends React.Component {
...
+      const onClick = (event) => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        if (
+          !buttonIsDisabled && _window.external &&
+          'AddSearchProvider' in _window.external
+        ) {
+          _log.info('Adding OpenSearch Provider', { addon });
+          _window.external.AddSearchProvider(addon.installURL);
```
kumar303: is there any way to pass a callback or something in case of an error? At the very least we should log errors. But I don't see any way to do it if these docs are accurate: https://developer.mozilla.org/en-US/docs/Web/API/Window/sidebar/Adding_search_engines_from_Web_pages

tests/client/core/components/TestInstallButton.js
```diff
@@ -155,4 +156,72 @@ describe('<InstallButton />', () => {
...
+      _window: fakeWindow,
+    });
+    const installButton = rootNode.querySelector('.InstallButton-button');
+    assert.equal(installButton.textContent, 'Add to Firefox');
+
+    Simulate.click(installButton);
+
+    assert.equal(fakeLog.info.firstCall.args[0], 'Adding OpenSearch Provider');
+    assert.ok(
+      fakeWindow.external.AddSearchProvider.calledWith(fakeAddon.installURL));
```
kumar303: I'd prefer `assert.equal(fakeWindow.external.AddSearchProvider.firstCall.args[0], ...)` because the failure message will be more informative.

tests/client/core/components/TestInstallButton.js
```diff
@@ -155,4 +156,72 @@ describe('<InstallButton />', () => {
...
+  });
+
+  it('disables the button/does nothing if OpenSearch is not available', () => {
+    const fakeLog = { info: sinon.stub() };
+    const fakeWindow = { external: undefined };
```
kumar303: It would be good to add another test for an empty `window.external` like `const fakeWindow = { external: {} };`

tests/client/core/test_utils.js
```diff
@@ -916,6 +917,21 @@ describe('getCompatibleVersions', () => {
...
+      ...fakeAddon,
+      current_version: {
+        compatibility: {},
+      },
+      type: ADDON_TYPE_OPENSEARCH,
+    };
+    getCompatibleVersions({
+      _log: fakeLog, addon: openSearchAddon, clientApp: 'firefox' });
+    assert.include(fakeLog.info.firstCall.args[0],
+      `addon is type ${ADDON_TYPE_OPENSEARCH}`);
```
kumar303: I don't think checking the log message is so important. I'd rather see this test assert the output of `getCompatibleVersions`. But you could do both.

src/core/components/InstallButton/index.js
```diff
@@ -72,6 +84,32 @@ export class InstallButtonBase extends React.Component {
           {i18n.gettext('Install Theme')}
         </Button>
       );
+    } else if (addon.type === ADDON_TYPE_OPENSEARCH) {
```
kumar303: If `_window.external.AddSearchProvider` is not defined, the button will do nothing which isn't very helpful. Could we disable the button instead or render an error message?

src/core/components/InstallButton/index.js
tofumatt: Sadly there's no callback or error event. But if the function doesn't exist it should never be called and it's an API that to my mind only Firefox implements. So it shouldn't ever fail... 🤷‍♀️

src/core/components/InstallButton/index.js
tofumatt: Shoot, I thought I disabled it but maybe that got lost when I moved more of the logic into the button? Yes, it should be disabled, I'll do that.

tofumatt:
Okay I've actually reworked this a bit and moved the check into the compatibility util checker function (where it really should have been anyway).

All set for another r?

src/amo/components/RatingManager/index.js
```diff
@@ -97,7 +97,7 @@ export class RatingManagerBase extends React.Component {
         return i18n.gettext('Log in to rate this dictionary');
       case ADDON_TYPE_LANG:
         return i18n.gettext('Log in to rate this language pack');
-      case ADDON_TYPE_SEARCH:
+      case ADDON_TYPE_OPENSEARCH:
         return i18n.gettext('Log in to rate this search engine');
```
kumar303: Should we call this a 'search plugin?' That's what MDN calls them but sometimes it says 'search engine plugin.'

src/core/components/InstallButton/index.js
```diff
@@ -72,6 +84,29 @@ export class InstallButtonBase extends React.Component {
...
+      const onClick = (event) => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        if (compatible) {
```
kumar303: It looks like this if statement is not necessary since the button will be disabled. If so, I think you should remove the if statement because at least we'd see an error in the console if `onClick` was triggered by a programming mistake -- this would help us track down the bug.

src/core/utils.js
```diff
@@ -385,6 +391,13 @@ export function isCompatibleWithUserAgent({
       return { compatible: false, reason: INCOMPATIBLE_UNDER_MIN_VERSION };
     }
 
+    if (
+      addon.type === ADDON_TYPE_OPENSEARCH &&
+      !(_window.external && 'AddSearchProvider' in _window.external)
+    ) {
+      return { compatible: false, reason: INCOMPATIBLE_NO_OPENSEARCH };
```
kumar303: Yeah, this makes way more sense 🌴 🏖

tests/client/core/components/TestInstallButton.js
```diff
@@ -155,4 +161,53 @@ describe('<InstallButton />', () => {
...
+  it('disables the OpenSearch button if not compatible', () => {
+    const root = render({
+      addon: { ...fakeAddon, type: ADDON_TYPE_OPENSEARCH },
+      getClientCompatibility: getClientCompatibilityFalseOpenSearch,
+    });
+
+    assert.equal(root.type, 'div');
+    const buttonComponent = root.props.children[1];
+    assert.equal(buttonComponent.type, Button);
+    assert.equal(buttonComponent.props.className,
```
kumar303: could this just be `assert.include(buttonComponent.props.className, 'InstallButton-button--disabled')` ?

src/amo/components/RatingManager/index.js
tofumatt: Yes actually! "search plugin" I think works (because it's not always searching via a search engine).

tests/client/core/components/TestInstallButton.js
tofumatt: Indeed, good call!

PR #2187, opened 2017-03-25
fix: Use Firefox categories for Android themes (fix #2170)

### LINKED ISSUE #2170: Android Theme categories don't exist so "Browse Theme Categories" link goes nowhere
Because #1875, this happens: https://github.com/mozilla/addons-frontend/issues/1745#issuecomment-288984336

We'll need to catch that request as well and stub in the `clientApp` as Firefox instead of Android until Android has its own theme categories...

tofumatt:
fix #2170

Similar to the solution used for searching Android Themes, except we patch the data in the reducer/state instead of the API call because the category list API just returns all categories for all clientApps.

### Before
mar-25-2017 23-05-52

### After
mar-25-2017 23-05-00

### DIFF
--- a/src/core/reducers/categories.js
+++ b/src/core/reducers/categories.js
@@ -1,6 +1,7 @@
 import config from 'config';
 
 import {
+  ADDON_TYPE_THEME,
   CATEGORIES_GET,
   CATEGORIES_LOAD,
   CATEGORIES_FAILED,
@@ -50,6 +51,16 @@ export default function categories(state = initialState, action) {
             ), {});
         });
       });
+
+      // Android doesn't have any theme categories but because all lightweight
+      // themes (personas) are installable on Firefox Desktop and Android
+      // we share categories and themes across clientApps.
+      // See: https://github.com/mozilla/addons-frontend/issues/2170
+      //
+      // This can be removed once
+      // https://github.com/mozilla/addons-server/issues/4766 is fixed.
+      categoryList.android[ADDON_TYPE_THEME] = categoryList
+        .firefox[ADDON_TYPE_THEME];
 
       return {
         ...state,
--- a/src/core/reducers/categories.js
+++ b/src/core/reducers/categories.js
@@ -1,3 +1,4 @@
+import { oneLine } from 'common-tags';
 import config from 'config';
 
 import {
@@ -6,6 +7,7 @@ import {
   CATEGORIES_LOAD,
   CATEGORIES_FAILED,
 } from 'core/constants';
+import log from 'core/logger';
 
 
 export function emptyCategoryList() {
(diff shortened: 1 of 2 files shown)


tofumatt:
This would be good to land before today's tag because there are now links to these pages from the improve Extensions/Themes landing pages' headers (#2100). Otherwise users will click links that result in empty categories.

src/core/reducers/categories.js
```diff
@@ -51,6 +52,16 @@ export default function categories(state = initialState, action) {
...
       });
 
+      // Android doesn't have any theme categories but because all lightweight
+      // themes (personas) are installable on Firefox Desktop and Android
+      // we share categories and themes across clientApps.
+      // See: https://github.com/mozilla/addons-frontend/issues/2170
+      //
+      // This can be removed once
+      // https://github.com/mozilla/addons-server/issues/4766 is fixed.
+      categoryList.android[ADDON_TYPE_THEME] = categoryList
```
kumar303: You could use braces to limit the scope of the `categories` variable to just this switch case

kumar303: I know it's not part of this patch but could you log a warning [when we skip an app category](https://github.com/mozilla/addons-frontend/blob/f333d5eb389a017de2641692cf260c5bfc0e6476/src/core/reducers/categories.js#L35)? That could be very confusing if we have incorrect values configured for valid apps.

kumar303: I think it would be nice to log a warning right above this since it would be confusing if we fix the API result and forget to remove this line of code.

kumar303: Maybe add `TODO` in the comment? I've been grepping for those to find temporary code.

tests/client/core/reducers/test_categories.js
```diff
@@ -103,20 +103,27 @@ describe('categories reducer', () => {
     });
 
     it('sets the categories', () => {
+      // Notice all Firefox theme cateories are also set as Android theme
```
kumar303: typo: categories

tests/client/core/reducers/test_categories.js
```diff
@@ -126,6 +133,18 @@ describe('categories reducer', () => {
               slug: 'anime',
               type: ADDON_TYPE_THEME,
             },
+            naturé: {
```
kumar303: since this entire `ADDON_TYPE_THEME` object is a duplication of the Firefox one, how about pulling it out into a shared constant? I think that will also make the intention of the test case more clear.

src/core/reducers/categories.js
tofumatt: 👍  Totally. The more logs the better.

tests/client/core/reducers/test_categories.js
tofumatt: Good call.

PR #2190, opened 2017-03-27
fix: Make search result pagination match the UI mock

### LINKED ISSUE #2064: Make search result pagination match the UI mock

Perform a search that results in many pages. Scroll down to the previous/next buttons. Example: https://addons-dev.allizom.org/en-US/android/search/?page=1&q=food

### What happened?

The buttons don't match the UI mock. They are are not centered or at least they are missing some margins.

They should match the UI mocks closer.

screenshot 2017-03-14 12 55 30

This is not fallout from https://github.com/mozilla/addons-frontend/issues/2054 :) I checked before and after. The layout was the same.

aayushsanghavi:
Fixes #2064 
Before -
before

After -
paginate

### DIFF
--- a/src/core/components/Paginate/Paginate.scss
+++ b/src/core/components/Paginate/Paginate.scss
@@ -12,14 +12,13 @@
 }
 
 .Paginate {
-  padding: 0 10%;
-  margin-bottom: 5%;
+  padding: 0 30px;
+  margin-bottom: 30px;
 }
 
 .Paginate .Paginate-item {
-  @include button($background: #fff, $color: #000);
+  @include button($background: $base-color, $color: $link-color);
 
-  color: $link-color;
   display: inline-block;
   font-size: $font-size-s;
   font-variant: small-caps;
--- a/src/core/components/Paginate/Paginate.scss
+++ b/src/core/components/Paginate/Paginate.scss
@@ -11,16 +11,21 @@
   justify-content: space-between;
 }
 
+.Paginate {
+  padding: 0 10%;
+  margin-bottom: 5%;
+}
+
 .Paginate .Paginate-item {
   @include button($background: #fff, $color: #000);
 
+  color: $link-color;
   display: inline-block;
   font-size: $font-size-s;
   font-variant: small-caps;
   font-weight: normal;
   line-height: 2;
-  margin: 0;
-  min-width: 40%;
+  min-width: 48%;
   text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


src/core/components/Paginate/Paginate.scss
```diff
@@ -11,16 +11,21 @@
   justify-content: space-between;
 }
 
+.Paginate {
+  padding: 0 10%;
+  margin-bottom: 5%;
+}
+
 .Paginate .Paginate-item {
   @include button($background: #fff, $color: #000);
```
tofumatt: I know this isn't part of your patch, but could you change this line to use colour variables too?

Actually this should be using your colour that you used below. It should be: `button($background: $base-color, $color: $link-color);`.

src/core/components/Paginate/Paginate.scss
```diff
@@ -11,16 +11,21 @@
...
 
+  color: $link-color;
```
tofumatt: You can remove this line once you change the `button` mixin.

src/core/components/Paginate/Paginate.scss
```diff
@@ -11,16 +11,21 @@
   justify-content: space-between;
 }
 
+.Paginate {
+  padding: 0 10%;
+  margin-bottom: 5%;
```
tofumatt: Can we use an explicit pixel value here instead of `%`?

src/core/components/Paginate/Paginate.scss
```diff
@@ -11,16 +11,21 @@
   justify-content: space-between;
 }
 
+.Paginate {
+  padding: 0 10%;
```
tofumatt: If possible I think an explicit pixel value would be better; for wider screens, especially on mobile, it will allow the buttons to reach closer to the edge of the screen which is better for usability.

aayushsanghavi:
@tofumatt thanks for the review. It isn't much work so I guess I should get done with it soon.

aayushsanghavi:
@tofumatt I changed the % values to pixels and this is how it looks -

mobile devices -

mobile


tablet devices -

tablet


laptop devices -

laptop

aayushsanghavi:
@tofumatt could you have a look again?

tofumatt:
Hi @aayushsanghavi, looking now. really sorry this slipped through. I've been a bit busy and distracted the past few weeks!

aayushsanghavi:
@tofumatt thanks for the merge!
You've been working hard on the project so it slips out, which is why I reminded you :)

PR #2195, opened 2017-03-27
Use redux-saga

### LINKED ISSUE #2150: Try redux-saga
As discussed in London, I'll try a patch for a single component that uses `reduxAsyncConnect` to see if `redux-saga` is any easier to write/code/test/use.

Related to this we might want to look into allowing components to render without content.

tofumatt:
Adds support for redux-saga and server-side rendering of sagas.

This patch also gets around redux-saga not blocking on request by simply rendering placeholder content for the components it renders, which I think is a nicer solution anyway and something we should do more of in the future. It's more manual but creates a much nicer feel and I think will make the client-side jankiness we loathe much less of an issue.

Closes #2150.

I think I've cleaned this up enough it's ready for r?

-----

Screenshot of the new loading indicator:

may-05-2017 14-40-30


kumar303:
Were you able to find out if `redux-saga` supports an isomorphic render on the server? We should figure that out first before going too far with it.

tofumatt:
Yeah, they have an issue about it and have a special END action with example usage that allows server side rendering. I'll actually need to add that before this patch is considered complete.

So we should be able to do server side rendering with saga.

tofumatt:
`END` issue is here: https://github.com/redux-saga/redux-saga/issues/255

So it's implemented and people use it, apparently the docs aren't perfect but good enough, heh.

tofumatt:
The test failures here are totally confounding me. Any idea what's going on?

kumar303:
> The test failures here are totally confounding me. Any idea what's going on?

The error doesn't make a whole lot of sense -- I guess you don't get it locally? One thing that comes to mind is that we need to support async functions in babel somehow. If it's working for you locally already then maybe it's set up. We use the `async` and `await` keywords on web-ext but perhaps that's a different approach? For that we have to specify [regenerator: true](https://github.com/mozilla/web-ext/blob/97404a9d3b803376dcc732a28e26aae63d0b3aa9/.babelrc#L10) and we use the [regenerator-runtime](https://github.com/mozilla/web-ext/blob/97404a9d3b803376dcc732a28e26aae63d0b3aa9/package.json#L69). Does `addons-frontend` need something like that?

tofumatt:
Yeah it’s not failing locally at all, and this used to pass on Travis… but recently it seems a tonne of branches are failing, including this one once pushed again.

I don’t think I have anything special set up locally that would cause it to fail on travis; all my changes are checked in.

tofumatt:
Basically the only issue I've hit that isn't easily solvable is that `redux-async-connect` sends its `END_GLOBAL_LOAD` event before the saga has finished to the loading bar won't display. There doesn't seem to be a straightforward way around this but I'll look further. This should be ready for review soon and fortunately changing something over to a saga is pretty easy so I should be able to do it for search in another PR and get the loading bar working then.

src/amo/components/Categories.js
```diff
@@ -1,29 +1,53 @@
...
+
   render() {
     const { addonType, categories, error, loading, i18n } = this.props;
 
     if (loading) {
       return (
         <div className="Categories">
-          <p>{i18n.gettext('Loading...')}</p>
+          <span className="visually-hidden">{i18n.gettext('Loading…')}</span>
+          <ul className="Categories-list"
```
tofumatt: I think some of this could be refactored so the categories were just an array of `<LoadingText />` elements and if `loading === true` they are displayed in `<span>` instead of `<a>` tags. I can refactor this to duplicate less markup I think.

src/amo/components/Categories.js
```diff
@@ -1,29 +1,53 @@
...
     categories: PropTypes.object.isRequired,
+    clientApp: PropTypes.string.isRequired,
     error: PropTypes.bool,
     loading: PropTypes.bool.isRequired,
     i18n: PropTypes.object.isRequired,
   }
 
+  componentWillMount() {
+    const { addonType, categories, clientApp, dispatch } = this.props;
+    if (!Object.values(categories).length) {
```
tofumatt: I would say this is the big difference for redux-saga: our components are now expected to load/work without data and then dispatch their own actions to get said data if that's the case. For this simple case of categories it was easy and I ended up reloading them from state because they nearly never change and it made sense. For things like search I suspect there'd be more effort involved though in reality it will mostly be shuffling around existing code. But I think this component now makes a lot more sense to look at: if no categories exist (basically only happens on first load) it fetches them and shows a placeholder until then. The mechanism for loading is in the saga, yes, but I think this is a bit easier to follow.

src/amo/components/Categories.js
```diff
@@ -1,29 +1,53 @@
...
+      dispatch(showLoading());
```
tofumatt: This should have been in the saga, my bad. Will move it.

src/core/server/base.js
```diff
@@ -302,10 +303,34 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
...
+          }
+
+          if (!rootSaga) {
+            return hydrateOnClient({ props, pageProps, res });
+          }
+
+          const sagas = store.runSaga(rootSaga);
+          // We need to render once because it will force components
+          // with sagas to call the sagas and load their data.
+          ReactDOM.renderToString(<ServerHtml {...pageProps} {...props} />);
```
kumar303: Why not just integrate this with `hydrateOnClient()` like:

````js
const HTML = ReactDOM.renderToString(<ServerHtml {...pageProps} {...props} />);
...
return sagas.done.then(() => {
  return hydrateOnClient({ HTML, props, pageProps, res });
});
````

````js
function hydrateOnClient({ HTML, res, props = {}, pageProps }) {
  let renderedHtml = HTML;
  if (!renderedHtml) {
    renderedHtml =
      ReactDOM.renderToString(<ServerHtml {...pageProps} {...props} />);
  }
  const componentDeclaredStatus = NestedStatus.rewind();
  return res.status(componentDeclaredStatus)
    .send(`<!DOCTYPE html>\n${renderedHtml}`)
    .end();
}
````

File: src/core/store.js [hunk 5]
```diff
@@ -30,11 +30,13 @@ const PROMISE_PREFIXES = {
 export function middleware({
   _config = config, _createLogger = createLogger,
   _window = typeof window !== 'undefined' ? window : null,
+  sagaMiddleware = () => { return (next) => (action) => next(action); },
```
kumar303: This seems awkward to me. Why not just have it accept arbitrary additions like this?

````js
export function middleware({
  _config = config, _createLogger = createLogger,
  _window = typeof window !== 'undefined' ? window : null,
  additions = [],
} = {}) {
  if (_config.get('isDevelopment')) {
    return compose(
      applyMiddleware(
        _createLogger(),
        ...additions,
        loadingBarMiddleware(PROMISE_PREFIXES),
      ),
      _window && _window.devToolsExtension ?
        _window.devToolsExtension() : (createStore) => createStore
    );
  }

  return compose(
    applyMiddleware(
      ...additions,
      loadingBarMiddleware(PROMISE_PREFIXES),
    ),
    (createStore) => createStore,
  );
}
````

File: src/ui/components/LoadingText/styles.scss [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+  animation: placeHolderShimmer 3s infinite cubic-bezier(0.65, 0.05, 0.36, 1);
+  background: $text-color-default;
+  display: inline-block;
+  height: 1rem;
+  line-height: 1;
+  margin: 0;
+  vertical-align: middle;
+}
+
+@for $percentage from 1 through 100 {
```
kumar303: This looks like it will add 300 lines to our CSS bundle because of the duplicated rules. Can that be avoided? Could it use a CSS transition or something instead? Or could it do a gradual fade-in like a fake progress indicator?

tests/client/amo/components/TestCategories.js
```diff
@@ -88,3 +106,41 @@ describe('Categories', () => {
     assert.equal(root.textContent, 'Failed to load categories.');
   });
 });
+
+describe('mapStateToProps', () => {
+  it('maps state to props', () => {
+    const props = mapStateToProps({
+      api: { clientApp: 'android', lang: 'pt' },
+      categories: {
```
kumar303: could you dispatch an action to produce this state instead?

tests/client/amo/components/TestCategories.js
```diff
@@ -88,3 +106,41 @@ describe('Categories', () => {
     assert.equal(root.textContent, 'Failed to load categories.');
   });
 });
+
+describe('mapStateToProps', () => {
+  it('maps state to props', () => {
+    const props = mapStateToProps({
+      api: { clientApp: 'android', lang: 'pt' },
```
kumar303: you can dispach `setClient` and `setLang` to produce this state

tests/client/amo/sagas/testCategories.js
```diff
@@ -0,0 +1,70 @@
...
+import { CATEGORIES_FETCH } from 'core/constants';
+
+
+const categories = {};
+
+describe('categoriesSaga', () => {
+  it('should get Api from state then make API request to categories', () => {
+    const fetchCategoriesGenerator = fetchCategories();
+
+    let next = fetchCategoriesGenerator.next();
```
kumar303: Oof, I know you are just following the docs but I think this testing approach is very fragile and not very helpful. I agree with all the same problems raised in this issue: https://github.com/redux-saga/redux-saga/issues/518

I did some digging and [this library](https://github.com/wix/redux-saga-tester) looks the most promising as a solution. Do you want to try rewriting the tests with that to see how it feels?

tests/client/core/actions/test_categories.js
```diff
@@ -1,49 +1,43 @@
...
 
   it('sets the type', () => {
-    assert.equal(action.type, 'CATEGORIES_GET');
+    assert.equal(action.type, 'CATEGORIES_FETCH');
   });
 
-  it('sets the query', () => {
-    assert.deepEqual(action.payload, params);
-  });
+  // it('sets the query', () => {
```
kumar303: Should this be removed or put back?

src/amo/routes.js
```diff
@@ -32,7 +32,15 @@ export default (
     <IndexRoute component={Home} />
     <Route path="addon/:slug/" component={DetailPage} />
     <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
-    <Route path=":visibleAddonType/categories/" component={CategoryList} />
+    {/* These user routes are to make the proxy serve each URL from */}
```
tofumatt: Indeed. I messed up resolving a conflict and should have kept these removed. Will fix!

src/amo/components/Categories.js
```diff
@@ -1,29 +1,51 @@
...
+    clientApp: PropTypes.string.isRequired,
     error: PropTypes.bool,
     loading: PropTypes.bool.isRequired,
     i18n: PropTypes.object.isRequired,
   }
 
+  componentWillMount() {
+    const { addonType, categories, clientApp, dispatch } = this.props;
+    if (!Object.values(categories).length) {
+      dispatch(categoriesFetch({ addonType, clientApp }));
```
tofumatt: Ah thanks, good catch. It works when testing in the browser too (I spent a lot of time looking at it when doing the loading text placeholder) so I knew it worked and didn't give it much thought. 👍

tests/client/amo/sagas/testCategories.js
tofumatt: Totally. I saw that library too and indeed even made a mistake during a refactor forgetting to re-assign `next` to the next result of the generator.

I'll give it a go with that lib.

src/ui/components/LoadingText/styles.scss
muffinresearch: If this is controlled only by client-side code then setting style props directly is the answer for this sort of thing, since react should poke the style properties directly and does not cause CSP issues. See also https://github.com/mozilla/addons-frontend/issues/2321

The only caveat to using style props is for server rendering because the initial server render would use inline-styles and they *would* be blocked by CSP.

tofumatt: In this particular case, at least for now, that loading text is only seen during client-side loading. Though that could change in the future if we wanted certain stuff that loaded only on the client-side.

src/amo/sagas/categories.js
```diff
@@ -0,0 +1,34 @@
...
+
+// worker Saga: will be fired on every CATEGORIES_FETCH action.
+export function* fetchCategories() {
+  try {
+    yield put(showLoading());
+    const api = yield select(getApi);
+    const response = yield call(categoriesApi, { api });
+    yield put(categoriesLoad(response));
+  } catch (err) {
+    yield put(categoriesFail(err));
```
tofumatt: That sounds good to me. I think we could possibly even do that in the root saga, but I'm not sure. I started with rudimentary error handling specific to this component, but we should be able to be quite flexible in the future 👍

src/amo/store.js
```diff
@@ -36,6 +39,10 @@ export default function createStore(initialState = {}) {
       search,
     }),
     initialState,
-    middleware(),
+    middleware({ sagaMiddleware }),
   );
+
+  store.runSaga = sagaMiddleware.run;
```
tofumatt: Ten four. There are a few examples of this usage and it was quick and easy, but you're right. I'll go through and change it to export both items, it'll be a bit noisier in the diff.

src/core/server/base.js
tofumatt: If that happens we'll be rendering HTML that doesn't have any data loaded from sagas. If you see the comments, we actually have to render twice when doing sagas on the server.

The first render will get any empty components to dispatch an action to fetch data with sagas. Once they're finished we have all that data in the store and THEN we can render and actually use the data.

Because we always need to render in `hydrateOnClient` (and it's called in a few places as the "final page load" function) we can just render after loading the sagas once to get them to dispatch all their actions.

src/ui/components/LoadingText/styles.scss
tofumatt: I think I'll go ahead with a client-side only solution because this loading animation is client-side only. I'll document that fact though, just in case it confuses anyone.

src/amo/components/Categories.js
tofumatt: I ended up just passing in props directly, see: https://github.com/mozilla/addons-frontend/pull/2195/files#diff-eaa60fc7f4b8c5473dadb4f134abbe3eR85

tests/client/amo/components/TestCategories.js
tofumatt: Definitely. This is largely old code just moved but I'll refactor, thanks 👍

tests/client/core/actions/test_categories.js
tofumatt: Oops, should be removed! Thanks.

tofumatt:
Okay! I think I fixed up everything, ready for another r?

tests/client/amo/sagas/testIndex.js
```diff
@@ -0,0 +1,15 @@
+import { fork } from 'redux-saga/effects';
+
+import rootSagas from 'amo/sagas';
+import categoriesSaga from 'amo/sagas/categories';
+
+
+describe('amo rootSagas', () => {
+  it('should get Api from state then make API request to categories', () => {
```
kumar303: is this comment accurate?

tests/client/amo/sagas/testIndex.js
```diff
@@ -0,0 +1,15 @@
...
+    const sagaGenerator = rootSagas();
+
+    assert.deepEqual(sagaGenerator.next().value, [
+      fork(categoriesSaga),
```
kumar303: Hmm, this test doesn't seem so useful. It looks like it will be hard to maintain, especially as sagas grow. Maybe you could just call `sagaGenerator.next()` to make sure no errors are thrown?

tests/client/amo/sagas/testUtils.js
```diff
@@ -0,0 +1,11 @@
+import { getApi } from 'amo/sagas/utils';
+import { signedInApiState } from 'tests/client/amo/helpers';
+
+
+describe('Saga utils', () => {
+  it('should return API state', () => {
+    const state = { api: signedInApiState, somethingElse: true };
+
+    assert.deepEqual(getApi(state), signedInApiState);
```
kumar303: This would be a better test if it used the `SagaTester`. I don't know how hard that is but the `redux-saga-tester` docs have some good examples of adding a saga function to the tester.

tests/client/core/reducers/test_categories.js
```diff
@@ -10,7 +17,20 @@ describe('categories reducer', () => {
...
+        [ADDON_TYPE_OPENSEARCH]: [],
+        [ADDON_TYPE_THEME]: [],
+      },
+      firefox: {
+        [ADDON_TYPE_DICT]: [],
+        [ADDON_TYPE_EXTENSION]: [],
+        [ADDON_TYPE_LANG]: [],
+        [ADDON_TYPE_OPENSEARCH]: [],
+        [ADDON_TYPE_THEME]: [],
+      },
```
kumar303: Instead of hard coding this list, I think it would be better to call `emptyCategoryList()` directly. I don't think you gain much by hard coding all the values.

tests/client/core/reducers/test_categories.js
```diff
@@ -24,12 +44,24 @@ describe('categories reducer', () => {
...
+          [ADDON_TYPE_OPENSEARCH]: [],
+          [ADDON_TYPE_THEME]: [],
+        },
+        firefox: {
+          [ADDON_TYPE_DICT]: [],
+          [ADDON_TYPE_EXTENSION]: [],
+          [ADDON_TYPE_LANG]: [],
+          [ADDON_TYPE_OPENSEARCH]: [],
+          [ADDON_TYPE_THEME]: [],
+        },
```
kumar303: same comment here about `emptyCategoryList()`

tests/client/amo/sagas/testIndex.js
tofumatt: We do essentially the same thing for reducers right now... though we might be able to use the redux saga testing utils to make it better. I forgot to use them in this test. I'll see about that and failing that do your thing, yeah.

tests/client/amo/sagas/testIndex.js
tofumatt: Oops, no, holdover from when the test was different.

tests/client/amo/sagas/testCategories.js
```diff
@@ -0,0 +1,119 @@
...
+    // Next action is loading the categories returned by the API.
+    assert.deepEqual(calledActions[2], actions.categoriesFail(error));
+
+    // Last action is to hide the loading bar.
+    assert.deepEqual(calledActions[3], hideLoading());
+
+    mockApi.verify();
+  });
+
+  it('should yield takeEvery() for the main generator', () => {
```
tofumatt: It ensures we response to all category actions instead of just one. Though actually because of how the categories API works it might be better to just have one request... I'll see.

src/core/server/base.js
```diff
@@ -251,12 +252,15 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
...
       let locale;
       let pageProps;
+      let sagaMiddleware;
       let store;
 
       try {
         cookie.plugToRequest(req, res);
 
-        store = createStore();
+        const _store = createStore();
```
tofumatt: I was getting syntax errors when I did that.

src/amo/components/Categories.js
```diff
@@ -46,14 +60,27 @@ export class CategoriesBase extends React.Component {
...
+          <div className="Categories-loadingText visually-hidden">
+            {i18n.gettext('Loading categories.')}
+          </div>
+        ) : null}
+        <ul
+          className="Categories-list"
+          ref={(ref) => { this.categories = ref; }}
+        >
+          {categories.map((category) => (
+            <li className="Categories-list-item" key={category.slug}>
```
tofumatt: Probably not a huge deal but that's a good point. I'll grab the index and put that in there if `category.slug` is undefined.

kumar303: Since it just needs to be a unique key, you could use the index for both cases

src/core/server/base.js
kumar303: Ah, ok. In that case, renaming `_store` to `storeData` would read better to me.

tests/client/amo/sagas/testCategories.js
kumar303: Okay, I think I see what you were going for. I think it would be better to test the actual logic with an approach more like this:
````js
sagaTester.dispatch(actions.categoriesFetch());
sagaTester.dispatch(actions.categoriesFetch()); // dispatch another
assert.equal(sagaTester.numCalled(actions.categoriesLoad({ result })), 2);
````

File: tests/client/amo/sagas/testIndex.js (hunk 16 above)
Comment (kumar303) [inline tests/client/amo/sagas/testIndex.js]: I know, the current test we have for registering reducers feels a bit overkill for me. I get how we want to add coverage to this code but other than that it feels like a duplication of effort.

File: tests/client/amo/sagas/testCategories.js [hunk 23]
```diff
@@ -0,0 +1,119 @@
...
+    const mockApi = sinon.mock(api);
+    const error = sinon.stub();
+
+    mockApi
+      .expects('categories')
+      .once()
+      .withArgs({
+        api: { ...initialState.api },
+      })
+      .throws(error);
```
tofumatt: `mockApi.throws(error);` seems to cause an error because `mockApi.throws is not a function`. I can trim it down to `mockApi.expects('categories').throws(error);`

tests/client/amo/sagas/testCategories.js
tofumatt: Just to note the actual API for `numCalled` is to just pass the action type and not the entire `{ type, payload }` object (as would result from this). I blindly copied this then wondered why it didn't work for about five minutes, hah!

It's actually:

```js
    sagaTester.dispatch(actions.categoriesFetch());
    sagaTester.dispatch(actions.categoriesFetch());

    assert.equal(sagaTester.numCalled(CATEGORIES_LOAD), 2);
```

tests/client/amo/sagas/testIndex.js
tofumatt: Fair enough actually and I think doing `rootSagas().next()` does not throw is fine. I did that and coverage remained the same; I agree that's a better and more maintainable test. 👍

tofumatt:
Okay, I think I addressed all the comments and tightened up the tests. I realised as well I was using `signedInApiState` which while I know technically isn't too bad to use, is still raw state and I want to get away from using it so I used the store and dispatched the actions. Yay!

kumar303:
> I was using signedInApiState ... and I want to get away from using it so I used the store and dispatched the actions

Good call! That was a bad habit I introduced as a baby step to DRY'ing up the state-based tests. We should fix that everywhere one day.

Issue #2044, opened 2017-03-13
Experiment with introducing Flow for static analysis


kumar303:
I keep seeing a lot of patterns that are going to be hard to maintain without something like [Flow](https://flowtype.org/).

Here are the main problems I think we can solve: 
* When tests mock an object's signature, we need to keep it in sync with that of the real object
* When components are instantiated, we need to make sure that all of its required props are used

This issue is to experiment with *introducing* Flow because we can't convert everything at once. Let's start with opting in a few files in AMO (using the `@flow` comment) and see how hard it is.

PR #2196, opened 2017-03-27
Begin adding Flow annotations

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2044

I have entered into the valley of darkness [1] to seek out the Light of Truth. *Who will walk with me*?

I started by annotating *one* component (I swear). I tried to follow its dependencies and annotate each one but I didn't finish because we have so much shared logic. I think this paints an interesting picture of what Flow can buy us.

Here are some observations:

* Declaring API response objects will be very helpful not only in catching errors but also in understanding how to work with the API. In my own experience, one already has to start by understanding the API data (either from the docs or with browser devtools) so using Flow will simply require us to document our findings.
* Declaring state objects will be very helpful for all kinds of things (reducers, components) because we have lots of code that depends on specific state structures. It has been pretty hard to catch all mistakes using our test suite.
* We can finally validate our use of component properties!
* Static typing reveals the truth about what your code is trying to accomplish and I have already spotted a few code problems after adding annotations.
* Flow adds verbosity and the syntax sometimes looks odd but I think the verbosity will make understanding how objects interact with each other much easier.
* Flow-typing tests is going to be hard because it will require some architectural changes around dependency injection. It may take longer to use Flow in our tests.

Let me know what you think!

[1] during my 8-hour flight back from London to Chicago


README.md
```diff
@@ -84,6 +86,66 @@ or have `InfoDialog` in their behavior text.
 Any option after the double dash (`--`) gets sent to `mocha`. Check out
 [mocha's usage](https://mochajs.org/#usage) for ideas.
 
+### Flow
+
+There is limited support for using [Flow](https://flowtype.org/)
+to check for problems in the source code.
+
+Here is how to develop with constant feedback about Flow issues:
```
tofumatt: I would rephrase this to `To check for flow issues during development on any file change, run:`

README.md
```diff
@@ -84,6 +86,66 @@ or have `InfoDialog` in their behavior text.
...
+for hints on how to solve common Flow errors.
+
+To add flow coverage to a source file, put a `/* @flow */` comment at the top.
+The more source files you can opt into Flow, the better.
+
+Here is our Flow manifesto:
+
+* The point of Flow is to help you declare the intention of your code and
+  help your team understand the implications of changes to interfaces and data
+  structures. It adds some verbosity and overhead but JavaScript hates you so
```
tofumatt: 😆

artifacts/README.md
```diff
@@ -0,0 +1 @@
+Ephemeral artifacts are saved to this directory.
```
tofumatt: This could use more comments or explanation.

package.json
```diff
@@ -118,7 +120,7 @@
       }
     },
     "test": {
-      "command": "npm run version-check && npm run unittest && npm run servertest && npm run eslint && npm run stylelint",
+      "command": "npm run version-check && npm run unittest && npm run servertest && npm run flow:check && npm run eslint && npm run stylelint",
```
tofumatt: While I usually like running linters after "real" tests, I wonder if running flow first might help us find type errors in our code (exactly what flow can be useful for). Maybe we want to run flow before other tests?

src/amo/components/RatingManager/index.js
```diff
@@ -18,34 +21,54 @@ import {
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
 import DefaultRating from 'ui/components/Rating';
+/* eslint-disable no-duplicate-imports */
```
tofumatt: Oh, shame, that will get annoying 😦

src/core/api/index.js
```diff
@@ -8,9 +9,14 @@ import { schema as normalizrSchema, normalize } from 'normalizr';
 import { oneLine } from 'common-tags';
 import config from 'config';
 
+import { initialApiState } from 'core/reducers/api';
 import { ADDON_TYPE_THEME } from 'core/constants';
 import log from 'core/logger';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
+/* eslint-disable no-duplicate-imports */
```
tofumatt: No way to import types separately, like `import { ErrorHandler }, type { ErrorHandlerType } from 'core/errorHandler';`?

Ryuno-Ki:
I'm curious about flow now :-)

Reading through their guides I found that they have library definitions (or libdefs), which would make you not „pollute“ your source code. You use it for `define` already from what I can tell.

This project uses Babel already, but libdefs may be an option if you don't want to introduce a transpile step (say, it's a small step).
I am about to investigate that.

What would be helpful is getting some documentation generated out of flow type declarations.

kumar303:
> Reading through their guides I found that they have library definitions (or libdefs), which would make you not „pollute“ your source code. 

I would rather see type definitions next to the code that uses them. We already do transpilation (for ES7 features) so I don't think it's a burden. I find that hiding library definitions in magic include files adds too much obscurity -- I prefer to see explicit imports for each definition.

Ryuno-Ki:
Yeah, I understand. It was meant as a side note :-)

artifacts/README.md
kumar303: Ha, oops. I didn't mean to commit this. It was replaced by the `flow` directory.

package.json
kumar303: Yeah, I like that. I think solving Flow errors first is better because sometimes test failures can be hard to decipher when you've made a programming mistake.

src/amo/components/RatingManager/index.js
kumar303: Yeah, maybe there is a better way to fix it

src/core/api/index.js
kumar303: Nope:
````
src/core/api/index.js:12
 12: import { initialApiState }, type { ApiStateType } from 'core/reducers/api';
                               ^ Unexpected token ,
````
There are hundreds of issues open about "import type" so I'm not sure if someone is thinking about this or not.

**Comment by kumar303** [comment]
> What would be helpful is getting some documentation generated out of flow type declarations.

Yeah, this could be interesting. Sometimes auto-generated docs are not very readable though. It looks like [react-docgen](https://github.com/reactjs/react-docgen) has support for Flow but they don't support imported type aliases (yet).

**Comment by kumar303** [comment]
> I think we need to include a link to a good flow primer for new contributors because it makes the code quite foreign to look at.

Good call. I updated the readme with some links. The Flow docs just got a revamp (like, a few days ago) and I think they are way nicer now than they used to be.

File: src/amo/components/RatingManager/index.js (hunk 5 above)
Comment (muffinresearch) [inline src/amo/components/RatingManager/index.js]: This plugin can distinguish between flow imports and normal ones: see https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md

File: flow/libs/dedent.js.flow [hunk 7]
```diff
@@ -0,0 +1,3 @@
+// I'm not sure exactly what this should be. I was using this issue as a guide.
+// https://github.com/facebook/flow/issues/2616#issuecomment-289257544
+declare function dedent(params: Array<*>): string;
```
muffinresearch: dedent is getting removed so we should kill it an then it doesn't matter.

muffinresearch:
I like the idea that flow will help us catch problems and enforce good practices.

I too don't like the syntax much and find it confusing but I'm aware that's mostly a gut reaction and comes from not having ever used it.

Like @tofumatt said there's a lot of overhead involved and whilst part of that is likely to go away as we get used to using flow (or at least for those of us that haven't been using it regularly) I share the concerns around contributors - this will likely raise the bar even higher for entry and this project already has a big learning curve if you're looking to contribute bigger patches.

I think we need to consider if this will slow us down. I would imagine the only way to do that would be to try it out properly and then make an honest assessment after having spent some time with it if the value it provides outweighs the additional overhead to the development process.

I also wonder if it would be good to start by limiting flow to: 

- State, Actions + reducers e.g: https://flow.org/en/docs/frameworks/redux/
- props e.g. https://flow.org/en/docs/frameworks/react/ 

rather than going to town adding it to everything. But that's just a thought.

kumar303:
> I share the concerns around contributors - this will likely raise the bar even higher for entry and this project already has a big learning curve if you're looking to contribute bigger patches.

Intuitively I thought the same thing but I actually found that the opposite was true for web-ext. Flow helped contributors considerably, especially in how it doesn't let you make common newbie JavaScript mistakes that slow down development. 

For example, I've seen Flow help pinpoint issues like accidentally sending the wrong arguments to functions. Without Flow, these kinds of things are very hard to track down because you usually just see cryptic errors like 'object is undefined'. Experienced JavaScript developers have developed OCD habits to avoid these mistakes but inexperienced devs do not.

The key thing about Flow is that it makes you *plan out* the intention of your code which I think will be really helpful for new contributors since a reviewer can gain more insight into their thought process.

> I also wonder if it would be good to start by limiting flow to... state, actions, reducers, props

Once you opt a module into Flow you will have to annotate all exported functions otherwise it will yell at you. You could use suppression comments but I think that might make the code even more unreadable. I think the easiest way to introduce Flow will be one module at a time but we can go as slow as we need to.

In addition to the objects you mentioned, I'd really like to see API request/responses defined as Flow types because this is currently the hardest part of AMO to understand (without Flow).

src/amo/components/RatingManager/index.js
kumar303: Whoa, hah. We were already using `import/no-duplicates` so we didn't even need the core `no-duplicate-imports`. That was confusing. Fixed, thanks!

muffinresearch:
> Intuitively I thought the same thing but I actually found that the opposite was true for web-ext. Flow helped contributors considerably, especially in how it doesn't let you make common newbie JavaScript mistakes that slow down development.

Ah nice, that's good to know.

kumar303:
@tofumatt I've been chasing master to keep from avoiding conflicts. Have you had a chance to take a closer look?

tofumatt:
I had a quick look today but didn't finish, really sorry and I know this is a big patch we don't want to bitrot.

My concerns about contributors has been alleviated by your experience with web-ext, so I'll give a more thorough review this weekend.

Apologies for the delay!

src/core/components/AuthenticateButton/index.js
```diff
@@ -8,27 +10,37 @@ import { logOutFromServer, startLoginUrl } from 'core/api';
...
+import type { DispatchFn } from 'core/types/reduxTypes';
 
+type HandleLogInFn = (
+  location: UrlFormatParams, options?: {| _window: typeof window |}
+) => void;
+
+type HandleLogOutFn = ({| api: ApiStateType |}) => Promise<void>;
+
+type AuthenticateButtonProps = {|
+  api: ApiStateType,
```
tofumatt: Admittedly these are nicer to write than `PropTypes` and way more useful. That alone is a great selling point.

SLOWLY BEING CONVERTED! 😆

src/core/reducers/api.js
```diff
@@ -7,8 +8,44 @@ import {
...
+
+export const initialApiState = {
+  token: null,
+  lang: null,
+  clientApp: null,
+  userAgent: null,
+  userAgentInfo: null,
+};
+
+export default function api(
```
tofumatt: Some of this syntax is a bit hairy to someone new to flow. Any chance there's something you could point to (a docs page on what this is) for more foreign syntax like this? I don't even know what to search for.

src/core/types/addonTypes.js
```diff
@@ -0,0 +1,82 @@
+/* @flow */
```
tofumatt: This is extremely useful "code-as-docs". 👍

src/core/types/coreTypes.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
```
tofumatt: Do these filenames all need `Types` as suffixes? They're already in a folder named `types`–can we nix the `Types` suffix?

tests/client/amo/reducers/testReviews.js
```diff
@@ -122,14 +146,14 @@ describe('amo.reducers.reviews', () => {
...
       id: 1,
-      isLatest: true,
+      is_latest: true,
     }));
 
-    state = reviews(state, setReview(fakeReview, {
+    state = reviews(state, setFakeReview({
       id: 2,
-      isLatest: false,
+      is_latest: false,
```
tofumatt: I like that we aren't messing with these anymore 👍

src/amo/components/RatingManager/index.js
```diff
@@ -18,34 +21,52 @@ import {
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
 import DefaultRating from 'ui/components/Rating';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { UserReviewType } from 'amo/actions/reviews';
+import type { SubmitReviewParams } from 'amo/api';
+import type { UrlFormatParams } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+import type { DispatchFn } from 'core/types/reduxTypes';
+import type { AddonType, AddonTypeProp, AddonVersionType }
```
kumar303: Our check only enforces line length. I prefer it like this since the the import list fits all on one line but I can change it to your example, no biggie.

src/amo/components/RatingManager/index.js
```diff
@@ -18,34 +21,52 @@ import {
...
+  from 'core/types/addonTypes';
 
 import './styles.scss';
 
+type LoadSavedReviewFn = ({|
+  userId: number,
+  addonId: number,
+|}) => Promise<*>;
```
kumar303: This is a [function type](https://flow.org/en/docs/types/functions/#toc-function-types). The return value is a `Promise` which is defined as a [generic](https://flow.org/en/docs/types/generics/) since a promise resolve to a specific value. In this case, I'm using `*` to indicate that the promise can resolve to any value. I replaced this with the literal string `any` which isn't exactly the same but it's easier to understand.

src/amo/components/RatingManager/index.js
```diff
@@ -18,34 +21,52 @@ import {
...
+
+type SubmitReviewFn = (SubmitReviewParams) => Promise<void>;
+
+type RatingManagerProps = {|
+  AddonReview: typeof DefaultAddonReview,
+  AuthenticateButton: typeof DefaultAuthenticateButton,
+  Rating: typeof DefaultRating,
+  addon: AddonType,
```
kumar303: Yeah, I think with large object definitions we will need to alphabetize. I attempted it here but misplaced one of the attributes. Fixed.

src/amo/components/RatingManager/index.js
```diff
@@ -18,34 +21,52 @@ import {
...
+  errorHandler: ErrorHandlerType,
+  apiState: ApiStateType,
+  i18n: Object,
```
kumar303: Yeah, in a future patch I'd like to define `i18n` because we rely on `i18n.gettext(...)` everywhere.

src/amo/components/RatingManager/index.js
```diff
@@ -193,7 +222,14 @@ export const mapStateToProps = (state, ownProps) => {
   };
 };
 
-export const mapDispatchToProps = (dispatch) => ({
+type DispatchMappedProps = {|
+  loadSavedReview: LoadSavedReviewFn,
```
kumar303: Sure. I like `Func` because otherwise these names are going to get really long.

src/core/actions/index.js
```diff
@@ -32,28 +47,48 @@ export function setClientApp(clientApp) {
...
+export function loadEntities(entities: Array<Object>): LoadEntitiesAction {
   return {
     type: ENTITIES_LOADED,
     payload: { entities },
   };
 }
 
-export function setCurrentUser(username) {
+export type SetCurrentUserAction = {|
+  type: typeof SET_CURRENT_USER,
```
kumar303: Oh, whoops, good catch. My intention here was to say the action `type` could only be equal to `'SET_CURRENT_USER'` so that we can link that in the reducer and keep everything in sync. That `typeof` call only returns the type which is `string`, like you said. Duh. I need to think about this more in https://github.com/mozilla/addons-frontend/issues/2260 . For now I just changed them all to strings.

src/core/components/AuthenticateButton/index.js
kumar303: yeah, being able to look at a prop that expects an object and follow it to the type definition to know what the object looks like is a major win.

src/core/reducers/api.js
kumar303: The new docs are pretty nice. Were they not enough? Should I link to this [cheat sheet](http://www.saltycrane.com/blog/2016/06/flow-type-cheat-sheet/) ? I don't know how it's well it's maintained though.

kumar303:
@tofumatt thanks for the thorough review. I expanded the docs to help clarify everything and addressed each issue. Can you take another look?

src/core/reducers/api.js
tofumatt: Mostly I was confused by the `&` here: https://github.com/mozilla/addons-frontend/pull/2196/files/732cebdb6a9b2987b505687be588a4bfc94667ee#diff-503196c93f1c7e904f62dff0de0bb124R44

README.md
```diff
@@ -84,6 +86,70 @@ or have `InfoDialog` in their behavior text.
...
+* Check out the [getting started](https://flow.org/en/docs/getting-started/) guide.
+* Read through the [web-ext guide](https://github.com/mozilla/web-ext/blob/master/CONTRIBUTING.md#check-for-flow-errors)
+  for hints on how to solve common Flow errors.
+
+To add flow coverage to a source file, put a `/* @flow */` comment at the top.
+The more source files you can opt into Flow, the better.
+
+Here is our Flow manifesto:
+
+* We use Flow to **declare the intention of our code** and help others
```
tofumatt: These docs are really awesome now, thanks so much for doing another pass. Super easy to grok and the links are great 👍

README.md
```diff
@@ -84,6 +86,70 @@ or have `InfoDialog` in their behavior text.
...
+}
+````
+
+* Use [Exact object types](https://flowtype.org/en/docs/types/objects/#toc-exact-object-types)
+  via the pipe syntax (`{| key: ... |}`) when possible. Sometimes the
+  spread operator triggers an error like
+  'Inexact type is incompatible with exact type' but that's a
+  [bug](https://github.com/facebook/flow/issues/2405).
+  You can use the `Exact<T>` workaround from
+  [`src/core/types/coreTypes`](https://github.com/mozilla/addons-frontend/blob/master/src/core/types/coreTypes.js)
```
tofumatt: Why do we call the file `coreTypes` when it's already in the `types/` dir? I think I at least meant to ask but forgot or it got buried. Seems redundant though.

README.md
```diff
@@ -84,6 +86,70 @@ or have `InfoDialog` in their behavior text.
...
+  if you have to. This is meant as a working replacement for
+  [$Exact<T>](https://flow.org/en/docs/types/utilities/#toc-exact).
+* Try to avoid loose types like `Object` or `any` but feel free to use
+  them if you are spending too much time declaring types that depend on other
+  types that depend on other types, and so on.
+* You can add a `$FLOW_FIXME` comment to skip a Flow check if you run
+  into a bug or if you hit something that's making you bang your head on
+  the keyboard. If it's something you think is unfixable then use
+  `$FLOW_IGNORE` instead. Please explain your rationale in the comment and link
+  to a github issue if possible.
```
tofumatt: Nitpick but GitHub is a proper noun.

package.json
```diff
@@ -14,6 +14,8 @@
     "dev:amo:no-proxy": "better-npm-run dev:amo:no-proxy",
     "dev:disco": "better-npm-run dev:disco",
     "eslint": "eslint .",
+    "flow:check": "flow check",
+    "flow:dev": "chokidar .flowconfig flow/ src/ tests/ -i flow/logs/flow.log -c 'flow status' --initial",
```
tofumatt: I forget, is this safe to run if you don't have `chokidar` installed globally? I'm rusty on my `npm run` command knowledge. Maybe it's okay?

src/amo/api/index.js
```diff
@@ -13,10 +50,16 @@ export function submitReview({
   body,
   reviewId,
   ...apiCallParams
-}) {
+}: SubmitReviewParams): Promise<ApiReviewType> {
   return new Promise(
     (resolve) => {
-      const data = { rating, version: versionId, body, title };
+      const data = {
```
tofumatt: Nitpick but `postData`, `ratingData`, or `reviewData` might be a bit less generic. `data` is a vague variable name.

src/core/api/index.js
```diff
@@ -50,10 +63,29 @@ export function createApiError({ apiURL, response, jsonResponse }) {
   return apiError;
 }
 
+type CallApiParams = {|
+  endpoint: string,
```
tofumatt: This would be nice to alphabetise I think.

src/core/types/addonTypes.js
```diff
@@ -0,0 +1,82 @@
...
+  version: string,
+|};
+
+export type AddonAuthorType = {|
+  name: string,
+  url: string,
+|};
+
+export type AddonTypeProp =
+  typeof ADDON_TYPE_DICT |
```
tofumatt: These look like they should be `||` `OR`s not bitwise `OR`s, as I assume we want `string` here not `0`. Odd that the type being `0` didn't mess stuff up somewhere...

src/core/types/addonTypes.js
```diff
@@ -0,0 +1,82 @@
...
+  previews: Array<Object>,
+  public_stats: boolean,
+  ratings: {|
+    count: number,
+    average: number,
+  |},
+  review_url: string,
+  slug: string,
+  status: 'beta'
+    | 'lite'
```
tofumatt: Oh, I see single pipes being used here too. I know the pipes can surround the object syntax and it's something special in Flow. But are they also used as regular `OR` in Flow? What would a double pipe here do? Normally I'm used to catching bitwise `OR` operators as typos so they set off alarm bells in my head. 😖

src/core/types/coreTypes.js
tofumatt: What's the reason for wanting `types` in the folder and filename?

src/core/types/reduxTypes.js
```diff
@@ -0,0 +1,7 @@
+/* @flow */
+
+// This defines some Redux interfaces that we want to depend on.
+// It may be possible to use an official library for this.
+// See: https://github.com/reactjs/react-redux/pull/389
+// and: https://github.com/reactjs/redux/pull/1887/files#diff-46d86d39c8da613247f843ee8ca43ebc
+export type DispatchFn = (action: Object) => void;
```
tofumatt: This should be changed to `DispatchFunc` like other function type names.

src/core/types/addons.js
```diff
@@ -0,0 +1,67 @@
...
+  previews: Array<Object>,
+  public_stats: boolean,
+  ratings: {|
+    count: number,
+    average: number,
+  |},
+  review_url: string,
+  slug: string,
+  status: 'beta'
+    | 'lite'
```
tofumatt: These look like they should be `||` `OR`s not bitwise `OR`s. Odd that this resulting in `0` didn't mess stuff up somewhere...

Are they also used as regular `OR` in Flow? What would a double pipe here do? Normally I'm used to catching bitwise `OR` operators as typos so they set off alarm bells in my head. 😖

README.md
kumar303: Oh, I did change the name but I forgot to update this link! Fixed.

package.json
kumar303: Yep, it's safe. "In addition to the shell's pre-existing PATH, [npm run](https://docs.npmjs.com/cli/run-script) adds node_modules/.bin to the PATH provided to scripts."

src/amo/api/index.js
kumar303: Sure, I changed it. However, I feel like a short and concise variable name makes code easier to understand when it only spans 10 lines or so. Otherwise the verbosity can slow a reader down.

kumar303: Actually, naming it `review` instead achieves the same conciseness that `data` provided.

tofumatt: Sweet, I didn't even consider that, oops 😲

src/core/api/index.js
kumar303: I did another sweep and I think all param lists should be alphabetized now

src/core/types/addons.js
kumar303: It is a [Union type](https://flow.org/en/docs/types/unions/) so I think it's correct but, huh, maybe there is a [bug](https://github.com/facebook/flow/issues/3717) here because it's not protecting me from referencing an invalid `status` value (which is what I intended). Well, at the very least it's good documentation and when the bug is fixed it will produce errors.

kumar303:
@tofumatt thanks, this helped me dig into a few more issues. It should be ready for another look.

PR #2204, opened 2017-03-29
fix: Remove all comments from production build (fix #2113)

### LINKED ISSUE #2113: AMO prod is including docstring comments in JS bundle
See the JS bundle file on https://addons.mozilla.org/en-GB/firefox/; we should remove the docstrings with uglify or whatever.

tofumatt:
This saves about 6kB which is pretty tiny, but it might save more in the future and is worth doing. It's an easy optimisation so why not?


Ryuno-Ki:
I wonder, whether some comments need to be preserved due to legal reasons (license).

PR #2228, opened 2017-04-04
Extract/Merge AMO locales

kumar303:


muffinresearch:
This one looks like it was a no-op since nothing was added in the extract. The rest is just msgmerge noise. I think there's no need to land this.

kumar303:
@muffinresearch huh. How did you decipher that? Just curious so I know for next time. I noticed the reformatted IDs. Maybe I'm using a different version of gettext (perhaps this could be fixed by using a pure Node lib like https://github.com/smhg/gettext-parser).

muffinresearch:
@kumar303 Look at the extract commit, there's nothing but a date update. 

The rest comes from running msgmerge which will make various formatting adjustments.

kumar303:
Aha, ok. I was confused why the extract commit didn't change anything.

Issue #2238, opened 2017-04-06
Prompt the user to refresh the page when auth tokens expire


kumar303:
### Describe the problem and steps to reproduce it:

Open a page on AMO, let your auth token expire (see https://github.com/mozilla/addons-frontend/issues/2130), and try to do something requiring auth like submit a star rating. 

**For QA**: This will be really hard to reproduce. The first `addons.mozilla.org` token expirations will occur on April 28, 2017 but could happen anytime after that.

### What happened?

You see an error like this:
<img width="357" alt="screenshot 2017-04-04 17 27 50" src="https://cloud.githubusercontent.com/assets/55398/24778203/f8dec44e-1aee-11e7-8414-164fde495cf8.png">

### What did you expect to happen?

Ideally, this could be avoided by anticipating the token expiration (https://github.com/mozilla/addons-frontend/issues/2237) but there will inevitably be cases where that fails. For those cases, we should say something like "click here to refresh the page" (but in a more sensible message). 

To be clear, this will be a *last ditch effort* to recover from the expired token. When the user refreshes the page, they will be logged out which will make more sense. Single page apps are so fun!

kumar303:
The new screen looks like this:

<img width="361" alt="screenshot 2017-04-10 16 31 03" src="https://cloud.githubusercontent.com/assets/55398/24920599/752efa66-1ead-11e7-9283-ef0f8cf214d7.png">

When the user reloads, they are logged out. Adding this button gives them a chance to save any work that might get lost.

Issue #2071, opened 2017-03-15
No-op gettext/ngettext is being used and will result in untranslated strings.


muffinresearch:
Before localization there was a no-op gettext and ngettext in core/utils. The problem is that this is being used in some places instead of `i18n.ngettext` this will result in untranslated strings.

E.g: https://github.com/mozilla/addons-frontend/blob/c884314b9827bcbb420a1a9547f1bf73d22c90d3/src/amo/components/AddonDetail.js#L131

kumar303:
Let's do this after https://github.com/mozilla/addons-frontend/issues/2373

Issue #2237, opened 2017-04-06
Anticipate expiring auth tokens and switch to a logged out state


kumar303:
### Describe the problem and steps to reproduce it:

Open a page on AMO, let your auth token expire (see https://github.com/mozilla/addons-frontend/issues/2130), and try to do something requiring auth like submit a star rating. 

### What happened?

You see an error like this:
<img width="357" alt="screenshot 2017-04-04 17 27 50" src="https://cloud.githubusercontent.com/assets/55398/24778130/8a1aa12c-1aee-11e7-889a-b11eded68884.png">

### What did you expect to happen?

The frontend code should anticipate the time at which the token will expire and change the UI to a logged out state. It might be nice to show a message saying "you have been logged out" but at the very least everything will make more sense in a logged out state.

kumar303:
I think it should be pretty easy to preserve review text in local storage: https://github.com/mozilla/addons-frontend/issues/2322

muffinresearch:
@kumar303 I needed to use olympia.dev to get this working but it did look to work fine running locally.

The only thing that could improve this even more would be to actually have some kind of notification to the user perhaps providing directly a button to log in again so that it's clearer what actually happened. Unless you notice the login button change from `Logout` to `Log in / Sign up` you might wonder what happened.

PR #2241, opened 2017-04-07
Prompt the user to refresh the page if their session expires

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2238

This is a baby step toward helping users cope with expired sessions. It's definitely not ideal but it's a quick patch that may be worth landing. 

screenshot 2017-04-10 16 31 03

Here are some things to do later:

* We can probably set a timer and anticipate when a session will expire and change the state. This is a harder patch and will be done in https://github.com/mozilla/addons-frontend/issues/2237 Even after doing that patch, it will still be possible to get a signature expired result though (because the timer won't be exact).
* ~~We can switch to detecting a code instead of the 'Signature expired' string, which is https://github.com/mozilla/addons-server/issues/5122~~
* ~~I thought it would be nice to show a better message like 'Your session has expired. You need to log in again' but this was difficult because we only have `i18n` objects inside components. Maybe I can think of a way.~~

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -174,3 +174,6 @@ export const GLOBAL_EVENT_STATUS_MAP = {
 // The events here are set directly on mozAddonManager
 // they will be fired by addons and themes.
 export const GLOBAL_EVENTS = Object.keys(GLOBAL_EVENT_STATUS_MAP);
+
+// API error codes.
+export const API_ERROR_SIGNATURE_EXPIRED = 'ERROR_SIGNATURE_EXPIRED';
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -179,4 +179,6 @@ export const GLOBAL_EVENTS = Object.keys(GLOBAL_EVENT_STATUS_MAP);
 export const ERROR_UNKNOWN = 'ERROR_UNKNOWN';
 // API error codes. These values match the error codes defined here:
 // http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#unauthorized-and-permission-denied
+export const API_ERROR_DECODING_SIGNATURE = 'ERROR_DECODING_SIGNATURE';
+export const API_ERROR_INVALID_HEADER = 'ERROR_INVALID_HEADER';
 export const API_ERROR_SIGNATURE_EXPIRED = 'ERROR_SIGNATURE_EXPIRED';
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -175,5 +175,8 @@ export const GLOBAL_EVENT_STATUS_MAP = {
 // they will be fired by addons and themes.
 export const GLOBAL_EVENTS = Object.keys(GLOBAL_EVENT_STATUS_MAP);
 
-// API error codes.
+// Generic error codes.
+export const ERROR_UNKNOWN = 'ERROR_UNKNOWN';
+// API error codes. These values match the error codes defined here:
+// http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#unauthorized-and-permission-denied
 export const API_ERROR_SIGNATURE_EXPIRED = 'ERROR_SIGNATURE_EXPIRED';
--- a/src/core/reducers/errors.js
+++ b/src/core/reducers/errors.js
@@ -11,7 +11,7 @@ import log from 'core/logger';
  * http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#responses
  */
 function getMessagesFromError(error) {
-  let data = {
+  let errorData = {
     code: ERROR_UNKNOWN,
     messages: [],
   };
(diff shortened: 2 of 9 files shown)


src/core/reducers/errors.js
```diff
@@ -12,7 +12,10 @@ import { gettext } from 'core/utils';
  * http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#responses
  */
 function getMessagesFromError(error) {
-  let messages = [gettext('An unexpected error occurred')];
+  let data = {
+    messages: [gettext('An unexpected error occurred')],
```
tofumatt: Oh, so this isn't (and wasn't ever) translated. 😢

src/core/reducers/errors.js
```diff
@@ -41,12 +44,23 @@ function getMessagesFromError(error) {
...
+      data = { ...data, messages: apiMessages };
     } else {
       log.warn('API error response did not contain any messages', error);
     }
+
+    if (
+      error.response.data.detail &&
+      // TODO: switch to checking for a code after
+      // https://github.com/mozilla/addons-server/issues/5122
+      error.response.data.detail === 'Signature has expired.'
```
tofumatt: This should at least be in a `gettext` call so we can grep for it, but I would really like for us to translate this message.

Also it should be changed. At least "Your session has expired." instead of "signature" which is confusing and less common terminology.

src/core/reducers/errors.js
```diff
@@ -41,12 +44,23 @@ function getMessagesFromError(error) {
...
+    ) {
+      // In this case, the only way to recover from the error is to
+      // refresh the page. It is impossible to fully avoid this state.
```
tofumatt: Is there any reason not to refresh the page for the user in this case? Forms are often saved by browsers but we should test that. Maybe it could be a post-merge new issue/patch.

src/core/reducers/errors.js
tofumatt: Seems like, looking through the rest of the code, just an error code here that the component maps to a string, would be better and let us translate it.

src/core/reducers/errors.js
kumar303: Argh. We need to fix https://github.com/mozilla/addons-frontend/issues/2071

src/core/reducers/errors.js
kumar303: Reloading the page is pretty disruptive. It's hard to know for sure that the user won't lose any work so I didn't want to do it automatically. I think hitting this use case will be uncommon so I don't think we need to optimize it too much.

kumar303:
@tofumatt thanks for the review -- could you take another look? I addressed everything and luckily https://github.com/mozilla/addons-server/issues/5122 has already landed so I switched to using codes.

kumar303:
I updated the screenshot since the message is better now

src/core/reducers/errors.js
```diff
@@ -12,14 +12,21 @@ import { gettext } from 'core/utils';
  * http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#responses
  */
 function getMessagesFromError(error) {
-  let messages = [gettext('An unexpected error occurred')];
+  let data = {
+    code: undefined,
```
tofumatt: I feel like this should be its own code (`UNKNOWN_ERROR`) and that way the component can use real `i18n.gettext` for it.

I'm guessing this path should never happen but I think we might as well handle it correctly.

src/core/reducers/errors.js
```diff
@@ -12,14 +12,21 @@ import { gettext } from 'core/utils';
  * http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#responses
  */
 function getMessagesFromError(error) {
-  let messages = [gettext('An unexpected error occurred')];
+  let data = {
```
tofumatt: `data` as a variable name is always a bit of a code smell to me. I know it's not a big difference but even `errorMessageData` or `errorData` would help it feel a bit less generic...

src/ui/components/ErrorList/index.js
```diff
@@ -0,0 +1,67 @@
...
+  static defaultProps = {
+    _window: typeof window !== 'undefined' ? window : {},
+  };
+
+  render() {
+    const { _window, code, className, i18n, messages } = this.props;
+    const items = [];
+
+    messages.forEach((msg) => {
+      let msgString = msg;
```
tofumatt: I think we can write out `message` here in both vars instead of `msg`.

src/core/constants.js
```diff
@@ -174,3 +174,11 @@ export const GLOBAL_EVENT_STATUS_MAP = {
 // The events here are set directly on mozAddonManager
 // they will be fired by addons and themes.
 export const GLOBAL_EVENTS = Object.keys(GLOBAL_EVENT_STATUS_MAP);
+
+// Generic error codes.
+export const ERROR_UNKNOWN = 'ERROR_UNKNOWN';
+// API error codes. These values match the error codes defined here:
+// http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#unauthorized-and-permission-denied
```
tofumatt: Hyper nitpick but this can be HTTPS

PR #2275, opened 2017-04-17
Fix css/sass build failures

kumar303:
This should stop failures like this from happening: https://travis-ci.org/mozilla/addons-frontend/jobs/222881961

It appears that Travis had some cached artifacts of `node-sass` lying around and when it upgraded to a new LTS Node version, `node-sass` was left in a broken state.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ before_install:
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then gunzip -c geckodriver.tar.gz | tar xopf -; fi
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then chmod +x geckodriver && sudo mv geckodriver /usr/local/bin; fi
 before_script:
-- echo "TODO: npm rebuild node-sass"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,6 +24,9 @@ before_install:
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then gunzip -c geckodriver.tar.gz | tar xopf -; fi
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then chmod +x geckodriver && sudo mv geckodriver /usr/local/bin; fi
 before_script:
+# Make sure we have the right node-sass. See:
+# https://github.com/webpack-contrib/css-loader/issues/240#issuecomment-252661267
+- npm rebuild node-sass
```
kumar303: This takes 5 seconds which sucks. Any ideas for improvement?

kumar303:
Trying to delete the cache instead to see if that fixes it

tofumatt:
Arrrgh! I didn't think of that. Thanks so much! I'll run through all the old PRs and restart them as needed.

kumar303:
Actually, we'll also need to delete the cache for any PR that is currently failing. There is sadly no 'delete all' button unless I'm, missing it. https://travis-ci.org/mozilla/addons-frontend/caches

PR #2290, opened 2017-04-20
Switch to a logged out state when auth tokens expire

### LINKED ISSUE #2237: Anticipate expiring auth tokens and switch to a logged out state

kumar303:
Fixes #2237 

This handles the case when a user's auth token expires while they are using the app. Previously in https://github.com/mozilla/addons-frontend/issues/2238 the solution was to show an error and prompt the user to reload the page (note that due to obscure race conditions, this is still possible). 

In this patch, the user will be logged out automatically when the token expires.

Here are some scenarios to give you an example of what that will feel like:

## A happy-ish scenario
- Log in, load an add-on detail page
- Allow time to lapse so that your auth token expires
- Instead of seeing stars for rating, you will suddenly see the *Log in to rate this add-on* button

## A sad scenario
- Log in, load an add-on detail page
- Select a star rating
- Begin entering a text review
- As you are typing, be unfortunate enough to have your auth token expire
- You will now lose everything you typed and you will see the *Log in to rate this add-on* button

The sad scenario is sad but keep in mind that without this patch, the user will not be able to submit the review anyway. 

If that sounds okay, we could move forward with this patch and do a follow-up patch that will add a special case for components like this (ones that do text entry) so that the user sees a message to log out instead. I could maybe try to implement it in this patch but it could get complicated.

Another possibility is to follow-up with a patch that shows a warning before expiration time, like, "Your session is about to expire in 2 minutes."

Thoughts?


config/default-amo.js
```diff
@@ -104,4 +104,10 @@ module.exports = {
 
   // https://sentry.prod.mozaws.net/operations/addons-frontend-amo-prod/
   publicSentryDsn: 'https://dbce4e759d8b4dc6a1731d3301fdaab7@sentry.prod.mozaws.net/183',
+
+  // The amount of time (in seconds) that an auth token lives for.
+  // This needs to match the SESSION_COOKIE_AGE in addons-server:
+  // https://github.com/mozilla/addons-server/blob/master/src/olympia/lib/settings_base.py#L990
+  authTokenValidFor: 2592000, // 30 days
+  // TODO: patch addons-server with a comment about keeping in sync with this.
```
tofumatt: Worth filing an issue for this if you haven't already, just so it doesn't get lost.

tests/client/amo/containers/TestApp.js
```diff
@@ -204,4 +206,118 @@ describe('App', () => {
...
+      const props = {
+        authToken: userAuthToken(),
+        logOutUser: sinon.stub(),
+        ...customProps,
+      };
+      return render(props);
+    }
+
+    beforeEach(() => {
+      clock = sinon.useFakeTimers(Date.now());
```
tofumatt: I love this API.

tests/client/amo/containers/TestApp.js
```diff
@@ -204,4 +206,118 @@ describe('App', () => {
...
+      clock.restore();
+    });
+
+    it('logs out when the token expires', () => {
+      const authTokenValidFor = 10; // seconds
+      const logOutUser = sinon.stub();
+
+      renderAppWithAuth({ authTokenValidFor, logOutUser });
+
+      const fuzz = 3; // acount for rounding the offset calculation.
```
tofumatt: Typo: should be "account"

tests/client/amo/containers/TestApp.js
```diff
@@ -204,4 +206,118 @@ describe('App', () => {
...
+    it('only sets one timer when receiving new props', () => {
+      const authTokenValidFor = 10; // seconds
+      const authToken = userAuthToken();
+      const logOutUser = sinon.stub();
+
+      const root = render({ authToken, authTokenValidFor, logOutUser });
+      // Simulate updating the component with new properties.
+      root.componentWillReceiveProps({ authToken });
+
+      const fuzz = 3; // acount for rounding the offset calculation.
```
tofumatt: "account"

tests/client/amo/containers/TestApp.js
```diff
@@ -204,4 +206,118 @@ describe('App', () => {
...
+
+    it('does not log out until the token expires', () => {
+      const authTokenValidFor = 10; // seconds
+      const logOutUser = sinon.stub();
+
+      renderAppWithAuth({ authTokenValidFor, logOutUser });
+
+      clock.tick(5 * 1000); // 5 seconds
+      assert.notOk(logOutUser.called,
+        'expected logOutUser() to NOT be called');
```
tofumatt: Nitpick: "NOT to be called" is technically correct and I think a bit better.

src/amo/containers/App.js
```diff
@@ -1,46 +1,65 @@
...
-    clientApp: PropTypes.string.isRequired,
-    handleGlobalEvent: PropTypes.func.isRequired,
-    i18n: PropTypes.object.isRequired,
-    lang: PropTypes.string.isRequired,
-    location: PropTypes.object.isRequired,
-    mozAddonManager: PropTypes.object,
-    setUserAgent: PropTypes.func.isRequired,
-    userAgent: PropTypes.string,
-  }
+  mastHead: Node;
```
kumar303: It's an annotation of a [class property](https://flow.org/en/docs/types/classes/#toc-class-fields-properties) which is cool because [things like refs](https://github.com/kumar303/addons-frontend/blob/556b2a8a890706aa50708e142f7c7522fbc67a67/src/amo/containers/App.js#L199) tend to be a bit obscure if you're seeing them for the first time in test code. With the Flow annotations, you can always look at the top of the class and see all the ref properties.

src/amo/containers/App.js
```diff
@@ -74,9 +96,76 @@ export class AppBase extends React.Component {
         'userAgent not in state on App load; using navigator.userAgent.');
       setUserAgent(_navigator.userAgent);
     }
+
+    if (authToken) {
+      log.info('Setting a logout timer when the token expires');
+      this.setLogOutTimer(authToken);
```
kumar303: Ah, yeah, that's better. Thanks.

src/amo/containers/App.js
```diff
@@ -74,9 +96,76 @@ export class AppBase extends React.Component {
...
+        `Auth token "${encodedTimestamp}" triggered this base62 error: "${base62Error}"`);
+      return;
+    }
+    // If the encoded timestamp was malformed it will be 0 or negative.
+    if (createdAt <= 0) {
+      log.error(oneLine`Got an invalid timestamp from auth token;
+        encoded value: ${encodedTimestamp}; decoded value: ${createdAt}`);
+      return;
+    }
+    const readableCreatedAt = new Date(createdAt * 1000).toString();
```
kumar303: I didn't think of that. Hmm. Looking at `moment().format()` I don't think it would buy much though.

src/amo/containers/App.js
```diff
@@ -74,9 +96,76 @@ export class AppBase extends React.Component {
...
+    this.scheduledLogout = setTimeout(() => {
+      log.info('Logging out because the auth token has expired');
+      logOutUser();
+    }, expirationTime * 1000);
+  }
+
+  onViewDesktop = (
+    event: Event,
+    {
+      window_ = window, cookie_ = cookie,
```
kumar303: Heh, looks like I inherited some legacy. I fixed them, thanks.

config/default-amo.js
kumar303: Thanks for the reminder, I did actually forget :) https://github.com/mozilla/addons-server/issues/5284

kumar303:
Thanks for the review.

Issue #2019, opened 2017-03-08
Add-on review submission is blocked with a "...can't leave more than one review for the same add-on..." message


ValentinaPC:
Steps to reproduce:
1. Load AMO-dev details page for an add-on that wasn't yet reviewed by you (on your device)
2. Tap some review stars but change you mind by clicking outside the review submission popup
3. Refresh the page
4. Try to add a review for the same add-on

Expected results:
Review submission popup is displayed.

Actual results:
A message saying that you cannot add 2 reviews for the same add-on is displayed.

Notes/Issues:
- after step 2 the review counter is incremented (even if no review was submitted)
- at refresh rating star, selected at step 2, disappears
- second review for the same add-on should load edit review popup 

Verified on AMO-dev FF51(Android 6.0.1). Issue is reproducing on -dev and -stage.
Video for this issue:
![videotogif_2017 03 08_18 07 04](https://cloud.githubusercontent.com/assets/15685960/23712452/a3d663a0-042b-11e7-8b28-822e481f7a23.gif)

kumar303:
I can't reproduce this. When I refresh the page, the saved star rating is there. Can you reproduce it every time? Could you attach a console log when it happens?

ValentinaPC:
I can reproduce this every time on device or "mobile view" for desktop.
Please see the screenshot for the logs:
![2017-03-09_1022](https://cloud.githubusercontent.com/assets/15685960/23741690/71329752-04b2-11e7-8e99-af1d775ae44b.png)

kumar303:
Thanks for the console log, that's helpful. It looks like some kind of API call is failing but I can't find the error in our server logs. 

Does this happen for every add-on? Can you provide an example of an add-on it happens for on dev?

Also, your STR doesn't mention how you first logged in. Did you do it with the 'log in to rate this add-on' or before?

ValentinaPC:
Your welcome!

> Also, your STR doesn't mention how you first logged in. Did you do it with the 'log in to rate this add-on' or before?

I was already logged. Login was made normally, from "Log in /Sign up" button.

> Does this happen for every add-on? Can you provide an example of an add-on it happens for on dev?

You must use and add-on you did not reviewed before. It must be your first time. 
And it is mandatory to refresh the page after dismissing first attempt to review. (as in step 3)
Happens for me for any add-on.

Examples of add-ons:
- https://addons-dev.allizom.org/en-US/android/addon/youtube-10423/
[...]

PR #2294, opened 2017-04-21
Work around cache issues by filtering latest review more precisely

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2019

### DIFF
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -88,7 +88,12 @@ export function submitReview({
 
 type GetReviewsParams = {|
   addon: number,
+  filter?: string,
+  page?: number,
+  page_size?: number,
   user: number,
+  version?: number,
+  show_grouped_ratings?: boolean,
 |};
 
 export function getReviews(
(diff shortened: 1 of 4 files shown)


src/amo/api/index.js
```diff
@@ -88,7 +88,12 @@ export function submitReview({
 
 type GetReviewsParams = {|
   addon: number,
+  filter?: string,
+  page?: number,
+  page_size?: number,
   user: number,
+  version?: number,
+  show_grouped_ratings?: boolean,
```
tofumatt: nitpick: this should be above `user: number,`

src/amo/api/index.js
```diff
@@ -115,17 +120,19 @@ export function getReviews(
...
-    if (!user || !addon) {
-      throw new Error('Both user and addon must be specified');
+    if (!user || !addon || !version) {
+      throw new Error('user, addon, and version must be specified');
     }
-    // The API will only return the latest user review for this add-on.
-    resolve(getReviews({ user, addon }));
+    // The API will only return the latest user review for this add-on
+    // and version.
+    resolve(getReviews({ user, addon, version }));
```
tofumatt: Sort of irking me these aren't alphabetised everywhere but I mean it's not that big of a deal.

tests/client/amo/testApi.js
```diff
@@ -204,20 +216,21 @@ describe('amo.api', () => {
...
 
-      return getLatestUserReview({ user: 123, addon: 321 })
+      return getLatestUserReview({ user: 123, addon: 321, version: 456 })
         .then((review) => {
           assert.strictEqual(review, null);
         });
     });
 
-    it('requires a user and addon', () => {
+    it('requires a user, addon, and version', () => {
```
tofumatt: Technically we only changed the error message and don't actually discretely test for each case–it would be nice to have three tests here, each one making sure that two out of three params still triggers the error.

src/amo/api/index.js
kumar303: heh, sorry, it slows me down mentally. I think when the list spans multiple lines we should make sure it's alphabetized since the goal of alphabetization is helping you find what you're looking for. When there are three items, it's easy to find what you're looking for.

tests/client/amo/testApi.js
kumar303: Yeah, true. Those kind of tests are a bit annoying and hard to get right but I'll add them.

PR #2310, opened 2017-04-25
Handle cancelled install

### LINKED ISSUE #1435: Handle onInstallCancelled
With required permissions landing for WebExtensions a user will now be able to cancel an install by not accepting the permissions dialog. We should catch this and display a message to the user that the install was cancelled.

Firefox work in [bug 1308309](https://bugzilla.mozilla.org/show_bug.cgi?id=1308309).

webextensions-permissions-prompt

tofumatt:
Fixes #1435.

Allows an installation to be cancelled after the switch is toggled on. Resets the switch to uninstalled.

Screenshot:

apr-25-2017 18-45-30

### DIFF
--- a/src/core/reducers/installations.js
+++ b/src/core/reducers/installations.js
@@ -3,6 +3,7 @@ import {
   DOWNLOADING,
   ERROR,
   INSTALLED,
+  INSTALL_CANCELLED,
   INSTALL_COMPLETE,
   INSTALL_ERROR,
   INSTALL_STATE,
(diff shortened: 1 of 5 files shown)


tofumatt:
cc8f22d7a6a67d270eedd4a1d2ee5d084ce2522e is largely the relevant change here; sorry I just had a bunch of whitespace nits I couldn't unsee in the other commit.

src/core/reducers/installations.js
```diff
@@ -13,10 +14,12 @@ import {
   UNINSTALL_COMPLETE,
   acceptedInstallTypes,
 } from 'core/constants';
+import log from 'core/logger';
 
 
 export default function installations(state = {}, { type, payload }) {
   if (!acceptedInstallTypes.includes(type)) {
+    log.info(`${type} not in acceptedInstallTypes so action is being ignored.`);
```
tofumatt: I added this because this was totally unexpected to me and it tripped me up for a good thirty minutes writing this patch 😢 
Hopefully this will prevent anyone else from getting confused by this bit of code... though I'm wondering why it's there at all...

tests/client/core/TestInstallAddon.js
```diff
@@ -278,6 +279,17 @@ describe('withInstallHelpers inner functions', () => {
       }));
     });
 
+    it('resets status to uninstalled on onInstallCancelled', () => {
+      const dispatch = sinon.spy();
+      const guid = '{my-addon}';
+      const handler = makeProgressHandler(dispatch, guid);
+      handler({ state: 'STATE_SOMETHING' }, { type: 'onInstallCancelled' });
+      assert(dispatch.calledWith({
```
kumar303: It's not a big deal but I don't like this style because the failure doesn't give you any information. If you switched to this style you would possibly be able to figure out the fix faster:

````js
assert.deepEqual(dispatch.firstCall.args[0], {
  type: INSTALL_CANCELLED,
  payload: { guid },
});
````

File: tests/client/core/reducers/test_installations.js [hunk 3]
```diff
@@ -216,6 +217,32 @@ describe('installations reducer', () => {
...
+  it('updates the status on INSTALL_CANCELLED', () => {
+    const state = {
+      'my-addon@me.com': {
+        guid: 'my-addon@me.com',
+        url: 'https://cdn.amo/download/my-addon.xpi',
+        downloadProgress: 100,
+        status: DOWNLOAD_PROGRESS,
+      },
+    };
+    assert.deepEqual(
```
kumar303: The only important changes are `status` and `downloadProgress` so this would be more refactor-friendly as the following assertion:
````js
const newState = installations(state, {
  type: INSTALL_CANCELLED,
  payload: {
    guid: 'my-addon@me.com',
  },
});
assert.equal(newState['my-addon@me.com'].downloadProgress, 0);
assert.equal(newState['my-addon@me.com'].status, UNINSTALLED);
````

src/core/reducers/installations.js
mstriemer: I think this is so it doesn't do any work for unrelated actions. Since type could be anything like "CALL_API" or "SET_USER_TOKEN". Unfortunately since this is adding a new action it's a bit annoying.

Will this log on every non-installation specific action now?

tests/client/core/TestInstallAddon.js
tofumatt: Agreed actually. I copy+pasted this test because it was so similar to another but I too prefer `firstCall`. I'll change it over.

src/core/reducers/installations.js
tofumatt: Well what's weird is in most reducers we just check for a bunch of action `type`s (with either an `if/else` or `switch` control), act on them, and `return state;` if none of the types match.

Here we return early if something isn't in a constant array when we have to define the types we act on anyway. It seems like we're checking twice here and it's very different to the rest of our code.

mstriemer: Yeah, it definitely doesn't fit the mold of what we normally do. I think it had more setup that I wanted to guard against executing unnecessarily before but it looks pretty safe now.

The check could be removed along with the `/* instalbul ignore else */` comment and then the else could just return state by the looks of it.

Issue #2325, opened 2017-04-28
Cannot edit a star rating (without review text)


kumar303:
### Describe the problem and steps to reproduce it:

* Log in
* Go to an add-on that you have not rated
* Select a star rating but do not enter any text for the review (tap outside the box to close it)
* Simulate returning to the site by clicking the browser refresh button on the URL

### What happened?

The stars you had saved disappeared. If you try to rate the add-on again you see an error like 'you cannot leave more than one review'

### What did you expect to happen?

The saved stars should appear selected. You should not get an error when you try to edit the rating.

### Anything else we should know?

This is broken in production.

Here is some more info from the API perspective: https://github.com/mozilla/addons-server/issues/5288

The root cause is that a null `review.body` (which is due to no text being entered) triggers some code that excludes the review from the filter -- this makes the site think no saved rating exists. Part of the complication here is that the API request to query for existing reviews is not made with authentication even though the user is signed in.

kumar303:
I'm landing a quick fix to this. Here are the follow-ups that need to happen:
* We need a new API filter: https://github.com/mozilla/addons-server/issues/5297
* We need to fix the frontend code to use that filter: https://github.com/mozilla/addons-frontend/issues/2327
* We should really provide a way to visualize star rating reviews (instead of hide them) https://github.com/mozilla/addons-frontend/issues/2328

SVSanta:
Verified as fixed using AMO-dev FF55(Win 10 64Bit).
Postfix video:
![videotogif_2017 05 02_12 56 37](https://cloud.githubusercontent.com/assets/28138132/25616761/3d58a86c-2f47-11e7-83e8-d7ff4efdab95.gif)

kumar303:
Note that this was a regression of https://github.com/mozilla/addons-frontend/issues/2019

PR #2326, opened 2017-04-28
Fix editing of star ratings by adding auth to API request

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2325

This is the quickest fix I can think of but it needs some follow-up. This works around the new logic introduced by https://github.com/mozilla/addons-server/pull/5227 which now hides empty reviews (i.e. star ratings) by default. I'd like for it not to hide star ratings by default but let's do that in future patches. For now, this passes authentication to the API for list requests which will unhide reviews created by the current user.

### DIFF
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -88,6 +88,7 @@ export function submitReview({
 
 type GetReviewsParams = {|
   addon?: number,
+  apiState?: ApiStateType,
   filter?: string,
   page?: number,
   page_size?: number,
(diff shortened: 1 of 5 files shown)


src/amo/api/index.js
```diff
@@ -99,35 +100,39 @@ type GetReviewsParams = {|
...
+  { apiState, user, addon, ...params }: GetReviewsParams = {}
 ): Promise<GetReviewsApiResponse> {
   return new Promise((resolve) => {
     if (!user && !addon) {
       throw new Error('Either user or addon must be specified');
     }
     resolve(callApi({
+      // Make an authenticated request if an API token exists.
+      auth: Boolean(apiState && apiState.token),
+      state: apiState,
```
tofumatt: Nitpick here but I find it easier to read key/value args like this when they're sorted alphabetically.

kumar303: whoops, I had renamed this param which is why it got out of order

kumar303:
Thanks for the review. I listed all the follow-ups here: https://github.com/mozilla/addons-frontend/issues/2325#issuecomment-298101464

Issue #2322, opened 2017-04-27
Preserve unsubmitted review text


kumar303:
### Describe the problem and steps to reproduce it:

* Log in
* Rate an add-on and begin entering a text review
* Be unfortunate enough to have your [auth token expire](https://github.com/mozilla/addons-frontend/issues/2237) (or have your browser crash or your tab accidentally closed, etc)

### What happened?

The text you entered will be lost (in most cases).

### What did you expect to happen?

The text should not be lost. It should appear in the text field when you return to rate the add-on again.

### Anything else we should know?

This is more important now that https://github.com/mozilla/addons-frontend/issues/2237 landed

ValentinaPC:
The review text is preserved.
Verified this on AMO-dev FF53(Win 7) while in Mobile Site View.
Postfix video:
![close](https://cloud.githubusercontent.com/assets/15685960/26300480/dfebd3d4-3ee6-11e7-9d9e-792b18557e1c.gif)

PR #2329, opened 2017-04-29
Preserve add-on review text in local storage in case of crashes

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2322

This protects the user from losing a review they are writing in a few cases:
* they get logged out automatically (https://github.com/mozilla/addons-frontend/issues/2237)
* their browser crashes
* they accidentally close their tab or something


src/core/localStore.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(
+    id: string,
+    { localForage = defaultLocalForage }: LocalStoreOptions = {},
+  ) {
+    this.id = id;
+    this.localForage = localForage;
+    configureLocalForage({ localForage });
+  }
+
+  getData(): Promise<Object | null> {
```
tofumatt: Ah okay I see how this works now, we limit our storage engine to only use one key and put the data in there.

Maybe I'm being pedantic but I don't think adding `Data` to all of these methods helps with anything. I think we could remove them and it would make as much sense.

I might just be allergic to the word "data" in code though. Maybe I should just take some anti-allergy meds and be quiet 😉

src/amo/components/AddonReview.js
```diff
@@ -7,43 +12,82 @@ import { setDenormalizedReview, setReview } from 'amo/actions/reviews';
...
+  }
+
+  onSubmit = (event: Event) => {
     const { apiState, errorHandler, onReviewSubmitted, review } = this.props;
     const { reviewBody } = this.state;
     event.preventDefault();
     event.stopPropagation();
 
-    const newReviewParams = { body: reviewBody };
+    const newReviewParams = { body: reviewBody || undefined };
```
tofumatt: Seems like the storage library you wrote will return `null` for `reviewBody` here if it's not found. Any reason we need `undefined` here instead of `null`?

kumar303: The review params here are for setting a review object which is typed independently of the local storage thing. If I don't initialize this to `undefined`, Flow says:
````
src/amo/components/AddonReview.js:90
 90:     const newReviewParams = { body: reviewBody };
                                         ^^^^^^^^^^ null. This type is incompatible with
  8:   body?: string,
              ^^^^^^ string. See: src/amo/actions/reviews.js:8

src/amo/components/AddonReview.js:90
 90:     const newReviewParams = { body: reviewBody };
                                         ^^^^^^^^^^ null. This type is incompatible with
 33:   body?: string,
              ^^^^^^ string. See: src/amo/api/index.js:33
````
This is because the [body parameter](https://github.com/kumar303/addons-frontend/blob/4f5ea85cf6cd94d356241d2c5e1ece1c299de77d/src/amo/actions/reviews.js#L8) can be undefined but not `null` or `false` or anything else.

tofumatt: ah gotcha

kumar303:
Thanks for the review. I renamed a bunch of things to make it more clear what the intention is.

PR #2337, opened 2017-05-02
fix(package): update normalize.css to version 7.0.0

kumar303:
Closes #2188 

See https://github.com/mozilla/addons-frontend/pull/2188 for discussion about the changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,7 +196,7 @@
     "moment": "2.18.1",
     "mozilla-tabzilla": "0.5.1",
     "mozilla-version-comparator": "1.0.2",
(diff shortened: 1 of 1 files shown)


muffinresearch:
Looks like the bigger changes that were in version 6 have been reversed: see https://github.com/necolas/normalize.css/compare/5.0.0...7.0.0 for the actual changeset if we merge this.

kumar303:
Oh, hey, thanks for digging up the diff. Yeah, I think we will be fine. The only change I see is that fieldsets now have different padding but we were already resetting that in `App.scss`.

Issue #2088, opened 2017-03-17
Clean-up forced 404 routes once new nginx config is released


muffinresearch:
Once the nginx configs from https://github.com/mozilla/addons-frontend/issues/2082 are rolled out to stage and prod it would be good to remove the forced 404 routes that should no longer be needed.

See https://github.com/mozilla/addons-frontend/blob/c8e567e7b6b7098368123d12fa698855c8e00d44/src/amo/routes.js#L32

kumar303:
This will be a little confusing (for local development only) since we won't see 404s anymore. I think we can get back to 404s after https://github.com/mozilla/addons-frontend/issues/2346

Issue #2082, opened 2017-03-16
Explicit nginx routes for amo mobile pages 


muffinresearch:
Relates to https://github.com/mozilla/addons-frontend/issues/2037

Currently we are routing traffic to the mobile pages based on:

* Presence of mamo cookie with value "on"
* OR, a mobile UA string

In addition to this we have a fallback which will fallback based on the mobile pages providing a 404. The fallback mostly works but there are a few cases which are causing problems:

* Any time the addons-server app issues a redirect it will be subject to being routed to the mobile pages again. 
* If the mobile page urls do not match the addons-server server urls there can be a game of redirect tennis which results in an infinite redir loop.
* If when on a  fallback page the client issues an XHR that might be routed incorrectly.

To fix this we could set up an explicit list of routes that the mobile pages responds to as this will be less prone to problems - the only downside will be one of maintenance. I'll add comments to the routes to make it clear that ops need to be told when new routes are added.

Based on https://github.com/mozilla/addons-frontend/blob/e356c928debbb226b57540b28e8a9b3d49cbaa91/src/amo/routes.js#L27

Here's the list of routes:

**All** routes are optionally prefixed with `/%locale%/(android|firefox)` though we also fill in both the locale and or the app if one does not exist.

```
/
/addon/%slug%/
/addon/%slug%/reviews/
/(dictionary|extensions|language|search|themes)/categories/
/(dictionary|extensions|language|search|themes)/featured/
/(dictionary|extensions|language|search|themes)/%slug%/
/search/
/simulate-async-error/
/simulate-sync-error/
/simulate-client-error/
/(dictionary|extensions|language|search|themes)/
```

muffinresearch:
@bqbn If it's possible to set this up we can verify it on -dev and stage and once verified get this out into the production as soon as possible.

muffinresearch:
@bqbn I just made a change because I just noticed this line: https://github.com/mozilla/addons-frontend/blob/master/src/core/constants.js#L65

```
persona -> themes
extension -> extensions
```

bqbn:
I've pushed some nginx changes to -dev for this. Please test and let us know how things are looking.

muffinresearch:
@bqbn thanks will check it out.

ValentinaPC:
@muffinresearch : I've checked all the routes that I've mentioned in #2037 using AMO-dev FF52(Win7) - clean profile - and did not see any changes.
As far as read above, some changes should be o -dev, yes?

[...]

PR #2347, opened 2017-05-03
Clean up 404 routes that we don't need anymore

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2088


kumar303:
@muffinresearch re: https://github.com/mozilla/addons-frontend/issues/2082 ...

> I'll add comments to the routes to make it clear that ops need to be told when new routes are added.

Did that comment ever get added? Should I add it?

muffinresearch:
> Did that comment ever get added? Should I add it?

Oops, I think that slipped through the net. Good catch - please do.

muffinresearch:
Since it doesn't matter leaving these, would it be worth folding this into the react-router upgrade to avoid having any weird falling-through locally?

kumar303:
> would it be worth folding this into the react-router upgrade to avoid having any weird falling-through locally?

I think it's ok to do the two patches independently. The lack of 404 routes will cause a 500 locally which isn't really any more or less helpful than a 404.

PR #2363, opened 2017-05-09
chore(package): update postcss-loader to version 2.0.3

kumar303:
Closes #2354


muffinresearch:
Big version bump. The conflicts are fixed, but it needs checking to make sure the config is correct.

Issue #2060, opened 2017-03-14
Make it easier to see what version of addons-frontend is deployed


kumar303:
We have a [``__version__``](https://github.com/mozilla/addons-frontend/blob/master/src/core/server/base.js#L188) URL but it gets overshadowed by addons-server:

````
$ curl https://addons-dev.allizom.org/__version__ | json_pp
{
   "build" : "https://circleci.com/gh/mozilla/addons-server/5405",
   "version" : "",
   "source" : "https://github.com/mozilla/addons-server",
   "commit" : "faf35752ba24e1ec08b2b52a00fa588d90d0a688"
}
````

You can trick out nginx to see the frontend version but this is tough if you forget the cookie name:
````
$ curl -H 'Cookie: mamo=on'  https://addons-dev.allizom.org/__version__ | json_pp
{
   "version" : "",
   "source" : "https://github.com/mozilla/addons-frontend",
   "build" : "https://circleci.com/gh/mozilla/addons-frontend/1739",
   "commit" : "b24bb7eaeed61efe37a8e1a42820d1ed646ad49f"
}
````

Let's make it easier. Perhaps it's enough just to change the URL to `__frontend_version__`?

kumar303:
This relies on https://github.com/mozilla/addons-frontend/issues/2370 for the URL to go live

PR #2366, opened 2017-05-09
A better URL for checking the deployed version

kumar303:
This introduces a new `/__frontend_version__` URL as a convenience so that you don't have to set a cookie when requesting `/__version__` on AMO. The old `/__version__` URL will still exist.

Fixes https://github.com/mozilla/addons-frontend/issues/2060


muffinresearch:
This will likely need a nginx config addition.

tests/server/amo/TestViews.js
```diff
@@ -62,4 +67,45 @@ describe('AMO GET Requests', () => {
       assert.equal(res.header.location,
         '/en-US/firefox/search/');
     }));
+
+  it('should return the application version', () => {
+    const projectRoot = path.join(__dirname, '..', '..', '..');
```
muffinresearch: Iirc there might be a config entry that is essentially the root of the project

kumar303: Oh, neat. Thanks.

PR #2379, opened 2017-05-12
Remove noop ngettext usage from amo

### LINKED ISSUE #2378: Remove the no-op ngettext usage from amo
Relates to #2071 but this is just to cover the amo parts so we can get it done ahead of the admin app removal.

This is the problem now:

modules_pour_firefox

Once fixed the `READ ALL N REVIEWS` should be translated, correctly.

muffinresearch:
Fixes #2378 

Before: 
modules_pour_firefox

After:
modules_pour_firefox

### DIFF
--- a/src/amo/components/AddonDetail.js
+++ b/src/amo/components/AddonDetail.js
@@ -17,7 +17,6 @@ import { withInstallHelpers } from 'core/installAddon';
 import {
   isAllowedOrigin,
   getClientCompatibility as _getClientCompatibility,
-  ngettext,
   nl2br,
   sanitizeHTML,
 } from 'core/utils';
@@ -128,7 +127,7 @@ export class AddonDetailBase extends React.Component {
     if (addon.ratings.count) {
       const count = addon.ratings.count;
       const linkText = i18n.sprintf(
-        ngettext('Read %(count)s review', 'Read all %(count)s reviews', count),
+        i18n.ngettext('Read %(count)s review', 'Read all %(count)s reviews', count),
         { count: i18n.formatNumber(count) },
       );
 


tests/client/amo/components/TestAddonDetail.js
```diff
@@ -441,12 +441,15 @@ describe('AddonDetail', () => {
...
     it('prompts you to read many reviews', () => {
+      const i18n = getFakeI18nInst();
       const root = reviewFooterDOM({
         ratingsCount: 5,
+        i18n,
       });
       const footer =
         root.querySelector('.AddonDetail-read-reviews-footer');
       assert.equal(footer.textContent, 'Read all 5 reviews');
+      assert.ok(i18n.ngettext.calledWith('Read %(count)s review', 'Read all %(count)s reviews', 5));
```
kumar303: Instead of this assertion, I think it would be nicer to make a new test that renders with `{ ratingsCount: 1 }` and asserts the value of `footer.textContent`. I know that it was an accident to not call `i18n.ngettext()` in the first place but I don't think adding a regression test for that type of mistake is very helpful.

muffinresearch: So do you think it's better to leave the test change out entirely? The no-op ngettext would have passed a test showing the english singular phrase for a single rating.

kumar303: I think the best thing would be to add a new test for the singular phrase. I don't think it's so important to add a regression test for the fact that we are now using `i18.ngettext()` instead of `ngettext()`. It seems unlikely that we'll make that same mistake again (after that function is removed) and the way you were testing the implementation (rather than the functionality) didn't seem so useful.

muffinresearch: Ah, ok - that's already handled here: https://github.com/mozilla/addons-frontend/blob/2754e5c7faf719f65527fc853afb4243876d7df3/tests/client/amo/components/TestAddonDetail.js#L438

kumar303: Oh I didn't know that. Then I'd suggest just removing this assertion. It's up to you though.

kumar303: ...which you already did! Thanks 🚢

Issue #1761, opened 2017-02-08
Upgrade to webpack 2


tofumatt:
I was gonna just do this in the Greenkeeper PR but it's a fair bit of changing so closing the Greenkeeper PRs for now and leaving this as a code quality issue.

https://webpack.js.org/guides/migrating/

muffinresearch:
Nominating this for consideration for q2 - this would probably be tied quite closely to any potential changes wrt webpack-isomorphic-tools.

muffinresearch:
Also assumes https://github.com/mozilla/addons-frontend/issues/353 which was closed since upgrading to webpack2 assumes a clean-up of the webpack config at the same time.

PR #2380, opened 2017-05-12
Upgrade to webpack 2

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/1761

Webpack 1 to 2 migration guide: https://webpack.js.org/guides/migrating/

Here's what I tested:
* `npm run dev:amo`, `npm run unittest:dev`, `bin/extract-locales`
* reloading JS and CSS
* source maps

I had trouble getting `npm run start` to work to mimic production. Doing `npm run build` is fine and the server starts but the bundle URLs are a 404. How do you simulate the CDN locally? I see the bundle files in `dist` but I'm not sure how to serve them. Is there a special value to use for `amoCDN` and `staticHost` in the config?

I also had trouble figuring out how to test that the `postcss-loader` is working. Any ideas?


package.json
```diff
@@ -285,15 +285,15 @@
     "semver": "^5.3.0",
     "shelljs": "^0.7.7",
     "sinon": "^2.2.0",
-    "sri-stats-webpack-plugin": "^0.7.3",
+    "sri-stats-webpack-plugin": "github:sopherio/sri-stats-webpack-plugin#f23de19e1d84dab50e59076e05bcafa508ba64c4",
```
kumar303: This is pretty unfortunate since the patch to make it webpack 2 compatible is just a dependency bump. More info: https://github.com/mikechau/sri-stats-webpack-plugin/issues/71

karma.conf.js
```diff
@@ -24,16 +23,14 @@ const coverageReporters = [{
...
 const newWebpackConfig = Object.assign({}, webpackConfigProd, {
+  devtool: 'inline-source-map',
+  module: {
+    rules: getRules({ babelQuery, bundleStylesWithJs: true }),
+  },
+  output: undefined,
+  entry: undefined,
   plugins: [
-    new webpack.DefinePlugin({
-      CLIENT_CONFIG: JSON.stringify(clientConfig),
```
kumar303: In the shared `getPlugins()` array, the `process.env.NODE_ENV` value is added to `DefinePlugin` unlike before. This doesn't seem to cause any problems though.

webpack-common.js
```diff
@@ -0,0 +1,173 @@
...
+
+  const plugins = [
+    new webpack.DefinePlugin({
+      CLIENT_CONFIG: JSON.stringify(clientConfig),
+      'process.env.NODE_ENV': JSON.stringify('production'),
+    }),
+    // Replaces server config module with the subset clientConfig object.
+    new webpack.NormalModuleReplacementPlugin(
+      /config$/, 'core/client/config.js'),
+    // Substitutes client only config.
```
muffinresearch: The comment should be updated to refer to logging.

webpack-common.js
```diff
@@ -0,0 +1,173 @@
...
+    },
+    ...styleRules,
+    {
+      test: /\.svg$/,
+      use: [{ loader: 'svg-url-loader', options: { limit: 10000 } }],
+    }, {
+      test: /\.jpg$/,
+      use: [{
+        loader: 'url-loader',
+        options: { limit: 10000, mimetype: 'image/jpeg' },
```
muffinresearch: 10000 could be a `const` at the top of the file.

webpack.prod.config.babel.js
```diff
@@ -27,7 +24,9 @@ for (const app of appsBuildList) {
 const settings = {
   devtool: 'source-map',
   context: path.resolve(__dirname),
-  progress: true,
+  devServer: {
+    progress: true,
```
muffinresearch: What does this setting relate to - is it ok in a prod config?

package.json
muffinresearch: I wonder if we should consider an alternative lib e.g: https://github.com/waysact/webpack-subresource-integrity (there might be others, though this says it supports both webpack1 and 2).

The bit I'm not keen on is that this relies on a fork patch, which in turn loads another fork patch. Whilst you've been able to specify a hash in the first case, you can't in the second. To fix potential integrity issue we'd need to have our own forks or, it might be easier to use something else?

webpack-common.js
```diff
@@ -0,0 +1,173 @@
...
+      exclude: /node_modules/,
+      loader: 'babel-loader',
+      query: babelQuery,
+    },
+    ...styleRules,
+    {
+      test: /\.svg$/,
+      use: [{ loader: 'svg-url-loader', options: { limit: 10000 } }],
+    }, {
+      test: /\.jpg$/,
```
muffinresearch: Nice work on the DRY'ing up.

silicagelpack

webpack-common.js
```diff
@@ -0,0 +1,173 @@
...
+    {
+      test: /\.jsx?$/,
+      exclude: /node_modules/,
+      loader: 'babel-loader',
+      query: babelQuery,
+    },
+    ...styleRules,
+    {
+      test: /\.svg$/,
+      use: [{ loader: 'svg-url-loader', options: { limit: 10000 } }],
```
muffinresearch: Should this have a mimetype added?

webpack.prod.config.babel.js
```diff
@@ -128,19 +75,26 @@ const settings = {
...
 if (config.get('enablePostCssLoader')) {
-  settings.postcss = [
-    autoprefixer({ browsers: ['last 2 versions'] }),
-  ];
+  settings.plugins.push(
+    new webpack.LoaderOptionsPlugin({
+      options: {
+        context: path.resolve(__dirname),
+        postcss: [
+          autoprefixer({ browsers: ['last 2 versions'] }),
```
muffinresearch: One way it should be possible to know this is working is that if when it's disabled via conf the CSS should be smaller. That does assume there's something in the CSS that would need to be prefixed with though.

muffinresearch:
This is great - I made some comments, I think it might be nice to switch out the sri plugin to something better maintained. 

I also added @tofumatt to the reviewers since more eyes on this the merrier.

This is probably one to land after the tag just so there's more time to fix up any nits that come up in use.

webpack.prod.config.babel.js
kumar303: Yeah, it's just for the webpack dev server. I can just move it to the dev config though.

webpack-common.js
kumar303: I just read the docs on these loaders. I can add one here but why are we specifying mime types at all? The docs say that a mime type is inferred from file extension.

muffinresearch: If it's not needed then that's less code to maintain here 👍

webpack.prod.config.babel.js
kumar303: Looks like it's working. I searched `-ms-flex-pack` in `dist/amo-eb2fa8bf265a819a3276f84cff2ff9b2.css` before and after to verify.

kumar303:
@muffinresearch I was able to switch us to `webpack-subresource-integrity` but unfortunately it took some custom code. Can you take another look? The switch occurred in https://github.com/mozilla/addons-frontend/pull/2380/commits/eb720de3d6c88fa296307ec9e1bbb067d0d327e5

kumar303:
I had to do some fix-ups so I guess you'll have to dig for it in the diff: https://github.com/mozilla/addons-frontend/pull/2380/files#diff-c325001c4cdd93809a99cbe1cb1b201cR10

PR #2391, opened 2017-05-17
Generate debug locales with potools

### LINKED ISSUE #2390: Use potools to generate the debug locales
See https://github.com/mozilla/potools/ for details

muffinresearch:
Fixed #2390


locale/dbr/LC_MESSAGES/amo.po
```diff
@@ -5,631 +5,586 @@ msgstr ""
...
 
 #: src/amo/components/AddonDetail.js:152
 msgid "Rate your experience"
-msgstr "‮ᴚɐʇǝ ʎonɹ ǝxdǝɹıǝuɔǝ"
+msgstr "ǝɔuǝıɹǝdxǝ ɹnoʎ ǝʇɐᴚ"
 
 #: src/amo/components/AddonDetail.js:182
 msgid "%(addonName)s %(startSpan)sby %(authorList)s%(endSpan)s"
-msgstr "‮%(ɐppouNɐɯǝ)s %(sʇɐɹʇSdɐu)sqʎ %(ɐnʇɥoɹ⅂ısʇ)s%(ǝupSdɐu)s"
+msgstr "%(endSpan)s%(authorList)s ʎq%(startSpan)s %(addonName)s"
```
muffinresearch: Gah, this is a case that isn't handled properly, since when substituted this will mean the placeholders are reversed :/

muffinresearch: Have filed https://github.com/mozilla/potools/issues/1 to make a fix via potools.

bin/debug-locales
```diff
@@ -39,6 +39,6 @@ let poFile;
 for (locale of debugLocales) {
   potFile = path.join(templateDir, `${appName}.pot`);
   poFile = path.join(localeDir, locale, 'LC_MESSAGES', `${appName}.po`);
-  rewrite = locale === 'dbl' ? 'unicode' : 'flipped';
-  shell.exec(`podebug -i ${potFile} -o ${poFile} --rewrite ${rewrite}`);
+  rewrite = locale === 'dbl' ? 'unicode' : 'mirror';
+  shell.exec(`node_modules/.bin/potools debug ${potFile} --format ${rewrite} --output ${poFile}`);
```
kumar303: Is there an easy way to put `potools` on PATH? I guess if it were converted to `npm run debug-locales` then it would be but I don't know how easy that is.

muffinresearch: Yeah, I'm not sure either, this seems like the easiest way since it's a dev-dep.

muffinresearch:
@kumar303 this is ready for another look.

bin/debug-locales
kumar303: Would it better to do `path.join(__dirname, '..', 'node_modules', ...)` ? Otherwise it won't work unless you're running it from the root dir but I guess it would be weird not to.

docs/i18n.md
```diff
@@ -56,14 +55,8 @@ NODE_APP_INSTANCE=[MY_APP] bin/merge-locales
...
 ### Building the debug locales
 
-Using this command Generates a unicode debug locale and a flipped rtl debug
-locale using [podebug](https://bit.ly/1r3yFy5).
-
-`podebug` is required for this step, you can install it with:
-
- * OSX: `brew install translate-toolkit`
- * Debian Linux variants: `sudo apt-get install translate-toolkit`
- * For other platforms see the [podebug docs](https://bit.ly/1r3yFy5)
```
kumar303: Thanks for killing all these steps 🍰

PR #2398, opened 2017-05-18
Add missing keys

### LINKED ISSUE #2397: Fix missing keys
Would be good to eliminate these warnings since they might also affect performance if items rendered via a loop don't have a stable id.

[1] Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `LanguagePickerBase`. See https://fb.me/react-warning-keys for more information.

[1] Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `RatingBase`. See https://fb.me/react-warning-keys for more information.

muffinresearch:
Fixes #2397


muffinresearch:
@kumar303 Good call - I'll file an issue to investigate if we can get an early warning if this is missed in future.

PR #2400, opened 2017-05-18
Enable eslint rule for missing jsx keys

### LINKED ISSUE #2399: Fail early if keys are missing from an array of JSX elements
Would be nice to have something e.g. a lint rule to prevent keys being missed from elements provided via a loop: https://facebook.github.io/react/docs/lists-and-keys.html#keys

Maybe there's an existing lint available to catch it, or some other way to fail harder earlier?

muffinresearch:
Fixes #2399

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -83,12 +83,12 @@
     "react/prefer-stateless-function": "off",
     "react/jsx-indent-props": "off",
     "react/jsx-closing-bracket-location": "off",
+    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
     "react/jsx-first-prop-new-line": "off",
+    "react/jsx-key": "error",
     "react/no-string-refs": "error",
     // FIXME: Deprecate use of findDOMNode https://github.com/mozilla/addons-frontend/issues/968
     "react/no-find-dom-node": "off",
-    // FIXME: Switch using .jsx for jsx files https://github.com/mozilla/addons-frontend/issues/969.
-    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }]
   },
   "settings": {
     "import/ignore": [
--- a/src/admin/containers/AddonPage/index.js
+++ b/src/admin/containers/AddonPage/index.js
@@ -33,6 +33,7 @@ class AddonPage extends React.Component {
     ];
     if (addon.homepage) {
       items.push([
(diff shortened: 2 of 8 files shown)


.eslintrc
```diff
@@ -83,12 +83,13 @@
     "react/prefer-stateless-function": "off",
     "react/jsx-indent-props": "off",
     "react/jsx-closing-bracket-location": "off",
+    // FIXME: Switch using .jsx for jsx files https://github.com/mozilla/addons-frontend/issues/969.
```
kumar303: It looks like this was copied from the old code and I don't think it's relevant. Can it be removed?

src/admin/containers/AddonPage/index.js
```diff
@@ -33,19 +33,22 @@ class AddonPage extends React.Component {
     ];
     if (addon.homepage) {
       items.push([
+        // eslint-disable-next-line react/jsx-key
```
kumar303: Why can't these ones get keys?

src/ui/components/Rating/index.js
```diff
@@ -71,6 +71,7 @@ export class RatingBase extends React.Component {
       }
 
       return (
+        // eslint-disable-next-line react/jsx-key
```
kumar303: Ugh. I was afraid of this. There is a `key` defined in the `props` object but it looks like the linter isn't able to recognize it. Could you file an upstream bug?

tests/client/amo/containers/TestApp.js
```diff
@@ -92,6 +92,7 @@ describe('App', () => {
         return <p>The component</p>;
       }
     }
+    // eslint-disable-next-line react/jsx-key
     const root = render({ children: [<MyComponent />] });
```
kumar303: You could also just add `key="key"` here.

src/admin/containers/AddonPage/index.js
muffinresearch: It's actually using data in the object to make the key later on. The eslint rule just sees any jsx in an array as needing a key so this is a false positive.

.eslintrc
muffinresearch: 👍

src/ui/components/Rating/index.js
muffinresearch: One was already filed here: https://github.com/yannickcr/eslint-plugin-react/issues/613

Issue #2403, opened 2017-05-18
Prevent accidental modules from getting imported


kumar303:
If you import a module by name and that name happens to be a top-level directory in the project, then you might get confusing results. This is because the top-level directory is added to webpack's resolution path (which is only necessary to import test files).

PR #2404, opened 2017-05-18
Stop exposing the top-level directory as a module path

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2403


webpack.prod.config.babel.js
```diff
@@ -72,9 +72,9 @@ const settings = {
   resolve: {
     alias: {
       'normalize.css': 'normalize.css/normalize.css',
+      'tests': path.resolve('./tests'),
```
tofumatt: This needs to be unquoted for lint.

muffinresearch: It should be possible to pull this line when switching to jest.

tests/server/TestConfig.js
```diff
@@ -1,69 +1,6 @@
...
-  describe(`Config for ${appName}`, () => {
-    let oldNodeEnv;
-
-    before(() => {
-      oldNodeEnv = process.env.NODE_APP_INSTANCE;
-    });
-
-    beforeEach(() => {
-      process.env.NODE_APP_INSTANCE = appName;
-      process.env.NODE_ENV = 'production';
```
kumar303: These tests are tricky because they depend on the production config which is overidable with a [local config](https://github.com/mozilla/addons-frontend#building-and-running-services). They are creating conflicts when `config/local-production-amo.js` is present. On closer inspection, they are just checking that the configs are sane. Do we really need tests for that?

kumar303:
@tofumatt thanks, it should be ready for another look now.

@muffinresearch if you have a min to look that would be helpful.

tofumatt:
Looks like lint is still failing: https://travis-ci.org/mozilla/addons-frontend/jobs/234100092#L529

tests/server/TestConfig.js
tofumatt: They test the configs as they are? I'd say no... if we commit bad prod configs we'll see it on stage or something... and we rarely edit them. And that sounds like a fragile and easily fooled test anyway. I'd say it's fine to ditch 👍

Issue #1961, opened 2017-03-03
Add unit tests for server middleware and request handlers (i.e. baseServer())


kumar303:
We have integration tests for some app specific parts of [our middleware and request handlers](https://github.com/mozilla/addons-frontend/blob/master/src/core/server/base.js#L115) but we also need to add more unit-ish tests to exercise all the different types of error handling and recovery. I say unit-ish because ideally they will still use real web requests, just with fake routes.

kumar303:
Beware this code will change when we switch to the new router https://github.com/mozilla/addons-frontend/issues/2161

muffinresearch:
The main middlewares are now unit tests under Jest. There are other parts of the base server that need coveraget so I think this can stay open until that's done unless that's covered elsewhere.

willdurand:
We do have tests for the `baseServer` for a while now, closing.

Issue #837, opened 2016-07-26
Clean up the server tests


muffinresearch:
The server tests are starting to get clunky - it would be great if we could ditch them and move as much code to unitests as possible.
- CSP tests, we could look at passing the config to the middleware and check the output of the middleware directly (saving a whole request/response) setup via supertest.
- SRI tests, we should see if  we can move them to the HTML component tests.
- The basic request/response tests will be superceded by the UI tests once live.
- It would  be nice not to need to update the package.json with a new servertest run just because we added a new app.

muffinresearch:
I think this was mostly cleaned up with the migration to Jest.

Issue #217, opened 2016-04-14
Add coverage for the server tests


muffinresearch:
Ideally this would reflect in the overall coverage as opposed to being separate.

muffinresearch:
Fixed by #2413

PR #2413, opened 2017-05-22
Setup Jest

### LINKED ISSUE #2372: Try switching to Jest for running tests
The new [Jest test runner](https://github.com/facebook/jest) seems to add enough awesome that we should try it. Some highlights:
* Speedier tests by splitting things across workers
* No more automatic mocking
* Better control of running partial tests

muffinresearch:
Fixes #2372

Caveats: 

* Jest runs all tests under node so anything that relies on browser APIs needs to be mocked
* Config points to the node config rather than the client config in tests (unifying the configs is something that really needs to happen because of this)
* I've ripped out most of the server tests but I think with some middleware refactoring most of what is covered by them can be re-instated.
* Tests are no longer being run in a browser.

The wins:

* Much faster to run tests since it side-steps the webpack build.
* Much easier to control the testing to point to a single file or watch for only the tests that have changed etc.
* There's no separation for coverage since all tests are now run in one go, there's no longer any separate server tests.

TODOS:

* [x] Fix the lint errors 
* [x] Remove duplication of getClientConfig.
* [x] File issue for refactoring config (that needs to be a separate branch)  See https://github.com/mozilla/addons-frontend/issues/2422
* [x] Refactor middlewares and unit test independently - this should negate the need for the csp config tests. 
* [x] File issue to unify the logging - I don't think there's a ongoing need for bunyan vs console logging, having just one logger would make things simpler. See https://github.com/mozilla/addons-frontend/issues/2423
* [x] Convert to expect instead of assert.


jest.config.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+  return clientConfig;
+}
+
+module.exports = {
+  coveragePathIgnorePatterns: [
+    '<rootDir>/node_modules/',
+    '<rootDir>/src/core/server/webpack-isomorphic-tools-config.js',
+    '<rootDir>/tests/',
+    '<rootDir>/config/',
```
muffinresearch: Maybe this is actually useful?

src/core/i18n/utils.js
```diff
@@ -242,7 +242,8 @@ export function makeI18n(i18nData: I18nConfig, lang: string, _Jed: Jed = Jed) {
 
   // TODO: move all of this to an I18n class that extends Jed so that we
   // can type-check all the components that rely on the i18n object.
-  i18n.formatNumber = (number) => number.toLocaleString(lang);
+  // Note: the available locales for tests are controlled in tests/setup.js
+  i18n.formatNumber = (number) => new Intl.NumberFormat(lang).format(number);
```
muffinresearch: Browser support for this should be good enough. The change was necessary since it's a more universal approach and works for testing too.

tests/client/amo/components/TestAddonDetail.js
```diff
@@ -375,7 +375,7 @@ describe('AddonDetail', () => {
       getBrowserThemeData: () => '{"the":"themedata"}',
     });
     const header = rootNode.querySelector('.AddonDetail-theme-header');
-    assert.equal(header.dataset.browsertheme, '{"the":"themedata"}');
```
muffinresearch: `dataset` not supported for some reason running via node.

tests/client/core/reducers/test_authentication.js
```diff
@@ -43,7 +43,7 @@ describe('authentication reducer', () => {
       });
       assert.throws(
         () => setAndReduceToken(token),
-        Error, /Error parsing auth token .* JSON\.parse: unexpected character/);
+        Error, /Error parsing auth token "{"malformed JSON"}/);
```
muffinresearch: These kind of changes were necessary since the exception messages are subtly different under node.

tests/server/TestCSPConfig.js
```diff
@@ -21,19 +21,26 @@ const apiHosts = {
...
     it(`should have a source-list config for ${env}`, () => {
       process.env.NODE_ENV = env;
+      jest.resetModules();
       const cdnHost = cdnHosts[env];
       const apiHost = apiHosts[env];
-      const config = requireUncached('config');
+      const config = require('config');
       const cspConfig = config.get('CSP').directives;
-      assert.include(cspConfig.scriptSrc, cdnHost);
+      expect(cspConfig.scriptSrc).toContain(cdnHost);
```
muffinresearch: This was a test to see the better assertion failure messaging you get using expect - will need to migrate to that in due course.

jest.config.js
```diff
@@ -0,0 +1,52 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+require('babel-register');
+const config = require('config');
+
+function getClientConfig(_config) {
```
muffinresearch: This duplication should be removed, it was to work around an issue importing.

tests/server/TestCSPConfig.js
```diff
@@ -21,19 +21,26 @@ const apiHosts = {
...
+
+  after(() => {
+    process.env.NODE_ENV = oldNodeEnv;
   });
 
   // eslint-disable-next-line no-restricted-syntax
   for (const env of deployedEnvs) {
     it(`should have a source-list config for ${env}`, () => {
       process.env.NODE_ENV = env;
+      jest.resetModules();
```
kumar303: Could you add a comment for why that's needed?

tests/setup.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+} else {
+  // No `Intl`, so use and load the polyfill.
+  // eslint-disable-next-line global-require
+  global.Intl = require('intl');
+}
+
+
+const realSinon = sinon;
+global.before = beforeAll;
```
kumar303: I think it would be better to update all of this code, i.e. just rename `before()` to `beforeAll()`.

tests/client/core/i18n/test_utils.js
```diff
@@ -270,7 +270,7 @@ describe('i18n utils', () => {
       assert.deepEqual(result, [
         { lang: 'en', quality: 0.7 },
         { lang: 'fil', quality: 0.5 },
-      ], sinon.format(result));
```
muffinresearch: Because sinon.format is depreacted.

tests/client/core/i18n/test_utils.js
```diff
@@ -446,6 +446,7 @@ describe('i18n utils', () => {
     });
 
     it('localised formatting a number', () => {
+      // Skipped because node gives a different result.
```
muffinresearch: Nice catch -  I left the comment in by mistake after removing changing back from `it.skip()` -> `it()`

tests/setup.js
muffinresearch: Yeah I agree - I was just trying to limit what was needed in one go :)

muffinresearch: FWIW I find it's easy to confuse `beforeAll` with `beforeEach`, but we should still be consistent with Jest conventions.

package.json
```diff
@@ -18,13 +18,11 @@
...
-    "servertest": "bin/config-check.js && ADDONS_FRONTEND_BUILD_ALL=1 npm run build && better-npm-run servertest && better-npm-run servertest:amo && better-npm-run servertest:disco && better-npm-run servertest:admin",
     "start": "npm run version-check && NODE_PATH='./:./src' node bin/server.js",
-    "test": "bin/config-check.js && better-npm-run test",
-    "unittest": "bin/config-check.js && better-npm-run unittest",
-    "unittest:dev": "bin/config-check.js && better-npm-run unittest:dev",
-    "unittest:server": "bin/config-check.js && better-npm-run unittest:server",
-    "unittest:run": "bin/config-check.js && better-npm-run unittest:run",
+    "test-ci": "bin/config-check.js && better-npm-run test-ci",
+    "test": "bin/config-check.js && jest",
+    "test-watch": "bin/config-check.js && jest --watch",
```
muffinresearch: It depends, I find it's been useful to have shortcuts for all options. Sometime you don't want `--watch`.

package.json
```diff
@@ -18,13 +18,11 @@
...
     "lint": "npm run eslint && npm run stylelint",
-    "servertest": "bin/config-check.js && ADDONS_FRONTEND_BUILD_ALL=1 npm run build && better-npm-run servertest && better-npm-run servertest:amo && better-npm-run servertest:disco && better-npm-run servertest:admin",
     "start": "npm run version-check && NODE_PATH='./:./src' node bin/server.js",
-    "test": "bin/config-check.js && better-npm-run test",
-    "unittest": "bin/config-check.js && better-npm-run unittest",
-    "unittest:dev": "bin/config-check.js && better-npm-run unittest:dev",
-    "unittest:server": "bin/config-check.js && better-npm-run unittest:server",
-    "unittest:run": "bin/config-check.js && better-npm-run unittest:run",
+    "test-ci": "bin/config-check.js && better-npm-run test-ci",
+    "test": "bin/config-check.js && jest",
```
muffinresearch: Linting and flow are slow that's why I think it makes sense to split them out from a typical command.

tests/client/core/TestInstallAddon.js
```diff
@@ -96,6 +97,10 @@ describe('withInstallHelpers inner functions', () => {
     mapDispatchToProps = makeMapDispatchToProps({ WrappedComponent, src });
   });
 
+  beforeEach(() => {
+    configStub = sinon.stub(config, 'get').withArgs('server').returns(false);
```
muffinresearch: Or we can turn on resetModules to always be run, then you can mock modules without cause for concern. See https://facebook.github.io/jest/docs/en/configuration.html#resetmodules-boolean

Also I've not personally done this until now, but I notice it is already used elsewhere in the test suite.

muffinresearch: From the homepage:

>Jest parallelizes test runs across workers to maximize performance. Console messages are buffered and printed together with test results. Sandboxed test files and automatic global state resets for every test so no two tests conflict with each other.

tests/emptyModule.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * @providesModule EmptyModule
```
muffinresearch: Yep, good call.

jest.config.js
```diff
@@ -0,0 +1,52 @@
...
+  coveragePathIgnorePatterns: [
+    '<rootDir>/node_modules/',
+    '<rootDir>/src/core/server/webpack-isomorphic-tools-config.js',
+    '<rootDir>/tests/',
+    '<rootDir>/config/',
+  ],
+  globals: {
+    CLIENT_CONFIG: getClientConfig(config),
+  },
+  moduleDirectories: ['node_modules', 'src', '<rootDir>'],
```
muffinresearch: Nice catch!

tests/client/core/TestInstallAddon.js
kumar303: ok, cool, it sounds like the mocking is concurrency aware. We may need to double check that sinon's sandbox also works the same way. If the sinon sandbox gets reset after every test then I think it will be fine.

package.json
kumar303: Yeah, I suppose a separate command could be useful. I think we should definitely optimize for the watch case, i.e. make it the easiest one to type: `npm test` :)

package.json
```diff
@@ -18,13 +18,11 @@
...
-    "servertest": "bin/config-check.js && ADDONS_FRONTEND_BUILD_ALL=1 npm run build && better-npm-run servertest && better-npm-run servertest:amo && better-npm-run servertest:disco && better-npm-run servertest:admin",
     "start": "npm run version-check && NODE_PATH='./:./src' node bin/server.js",
-    "test": "bin/config-check.js && better-npm-run test",
-    "unittest": "bin/config-check.js && better-npm-run unittest",
-    "unittest:dev": "bin/config-check.js && better-npm-run unittest:dev",
-    "unittest:server": "bin/config-check.js && better-npm-run unittest:server",
-    "unittest:run": "bin/config-check.js && better-npm-run unittest:run",
+    "test-ci": "bin/config-check.js && better-npm-run test-ci && cat ./coverage/lcov.info | coveralls",
+    "test": "bin/config-check.js && jest",
+    "test-watch": "bin/config-check.js && jest --watch",
```
kumar303: I think this one will be used the most. If you called this `test` then it could be `npm test` which is easy to type. You could rename the one without watching to `test-once` or something.

package.json
```diff
@@ -18,13 +18,11 @@
...
     "stylelint": "stylelint --syntax scss **/*.scss",
     "lint": "npm run eslint && npm run stylelint",
-    "servertest": "bin/config-check.js && ADDONS_FRONTEND_BUILD_ALL=1 npm run build && better-npm-run servertest && better-npm-run servertest:amo && better-npm-run servertest:disco && better-npm-run servertest:admin",
     "start": "npm run version-check && NODE_PATH='./:./src' node bin/server.js",
-    "test": "bin/config-check.js && better-npm-run test",
-    "unittest": "bin/config-check.js && better-npm-run unittest",
-    "unittest:dev": "bin/config-check.js && better-npm-run unittest:dev",
-    "unittest:server": "bin/config-check.js && better-npm-run unittest:server",
-    "unittest:run": "bin/config-check.js && better-npm-run unittest:run",
+    "test-ci": "bin/config-check.js && better-npm-run test-ci && cat ./coverage/lcov.info | coveralls",
```
kumar303: This would make more sense to me as `test-all` 🚲  since the ci part is not really relevant to what it does

tests/client/core/TestInstallAddon.js
muffinresearch: The sinon we use is actually a sandbox and that gets implicitly reset here https://github.com/mozilla/addons-frontend/pull/2413/files#diff-81e799b66ac20327dd8a1b8bb1f5f802R69

package.json
muffinresearch: It is more relevant that it seems at first glance, since this pipes coverage data to coveralls too, so it's not really something someone should run locally.

package.json
muffinresearch: Have you tried it in practice? I can definitely change it especially as I won't be using it as much as you will - but, I'd definitely suggest you try it out, watch can get in the way and I don't think it makes sense as a default. 

Also not sure if you've been using `yarn` as a command runner but you can do `yarn test`, `yarn test-watch` negating the need for the `run` sub command which is way saner.

muffinresearch: Hmm actually create-react-app defaults to watching, ok I'll change this around....

muffinresearch: Ok you got it! 🍰

kumar303:
btw, here are a bunch of issues that may be closable after this patch:
* Clean up server tests https://github.com/mozilla/addons-frontend/issues/837
* Add coverage to server tests https://github.com/mozilla/addons-frontend/issues/217
* Add unit tests for middleware https://github.com/mozilla/addons-frontend/issues/1961

kumar303:
Sorry, we had a false alarm on stage so I didn't get a chance to look at this in detail.

Do you have some tips for working with `jest --watch`? It feels a lot slower to me. Previously, once karma started up it only took about 5-6 seconds to run/re-run the entire suite. Now it takes 35 seconds. Do you know what is causing it to be so slow? Is it just the addition of what used to be server tests?

muffinresearch:
> Do you have some tips for working with jest --watch? It feels a lot slower to me. Previously, once karma started up it only took about 5-6 seconds to run/re-run the entire suite. Now it takes 35 seconds. 

Are you running all the tests all the time? The expectation is that you'd limit the watch to a specific file or test case using the filtering options.

I suppose the main differences here are the karma test suite does all the compilation up front and then runs and re-runs in an open browser. Whereas jest will re-build related files on the fly and re-test. 

The real wins with Jest come from watching the things you care about for near instant feedback.

Reducing test overhead could also help, I would imagine more shallow rendering would reduce test time as well as making the tests themselves better.

It looks like the tests that stand out as slow are the ones that are using dom methods more heavily I would imagine all of that could be replaced with shallow rendering via enzyme and that would.

> Do you know what is causing it to be so slow? Is it just the addition of what used to be server tests?

No I don't think so - there's no supertest involved at the moment at all.

tests/client/amo/containers/TestSearch.js
```diff
@@ -170,14 +167,10 @@ describe('CurrentSearchPage.loadSearchResultsIfNeeded()', () => {
...
-      assert(
-        dispatch.firstCall.calledWith(
-          searchActions.searchStart({ filters, page })),
-          'searchStart not called');
-      assert(
-        dispatch.secondCall.calledWith(
-          searchActions.searchLoad({ filters, entities, result })),
-          'searchLoad not called');
+      expect(dispatch.firstCall.calledWith(
+        searchActions.searchStart({ filters, page }))).toBeTruthy();
```
kumar303: mmm, I want to kill all `calledWith()` assertions one day. The failure message is never helpful. Anyway, that's for later.

tests/client/amo/reducers/testLanding.js
```diff
@@ -52,11 +52,11 @@ describe('landing reducer', () => {
         addonType, featured, highlyRated, loading, popular,
       } = landing(initialState, getLanding({ addonType: ADDON_TYPE_THEME }));
 
-      assert.equal(addonType, ADDON_TYPE_THEME);
-      assert.equal(loading, true);
-      assert.deepEqual(featured, { foo: 'bar' });
-      assert.deepEqual(highlyRated, { count: 0 });
-      assert.deepEqual(popular, { results: [] });
+      expect(addonType).toEqual(ADDON_TYPE_THEME);
+      expect(loading).toEqual(true);
```
kumar303: This should be `.toBe(true)` to get strict equality, right? I think there are a couple more in here which might be worth hunting down.

tests/client/core/middleware/test_cspMiddleware.js
```diff
@@ -0,0 +1,162 @@
...
+import MockExpressRequest from 'mock-express-request';
+import MockExpressResponse from 'mock-express-response';
+import parse from 'content-security-policy-parser';
+
+import log from 'core/logger';
+import { csp, getNoScriptStyles } from 'core/middleware';
+
+
+describe('CSP Middleware', () => {
+  const exisitingNodeEnv = process.env.NODE_ENV;
```
kumar303: typo: `exisitingNodeEnv`

package.json
```diff
@@ -229,8 +165,8 @@
     "babel-core": "^6.24.1",
     "babel-eslint": "^7.2.3",
     "babel-gettext-extractor": "git+https://github.com/muffinresearch/babel-gettext-extractor.git#0d39d3882bc846e7dcb6c9ff6463896c96920ce6",
+    "babel-jest": "20.0.3",
```
muffinresearch: I wonder if we can change a config so that yarn knows to do that?

tests/client/amo/containers/TestSearch.js
muffinresearch: The best way to use that is to use sinon.assert directly. That will give a better message when it fails.

tests/client/amo/reducers/testLanding.js
muffinresearch: This won't be a problem but strictly speaking single value could use toBe it might be faster for starters.

In any case, for something like this if loading was 'blah' you'd get:

```
Comparing two different types of values. Expected boolean but received string.
```

package.json
kumar303: I did look for a config setting but there doesn't seem to be one. There's only one setting to cover both prod and dev prefs.

tests/client/TestPackageJson.js
```diff
@@ -0,0 +1,9 @@
+import packageJson from 'package';
```
kumar303: I don't think we should rely on using the root directory as a module path because there are a lot of folders and they could accidentally overshadow something in `node_modules`. In fact, I thought I fixed this in https://github.com/mozilla/addons-frontend/pull/2404 but I guess not, heh.

Instead, I think you should do:
````js
const packageJson = JSON.parse(readFileSync(path.join(config.get('basePath'), 'package.json')));
````

btw, this is a great test. Good idea!

Comment (muffinresearch) [inline tests/client/TestPackageJson.js]: @kumar303 the rootDir is currently added to jest module paths. I think things might break if I remove it, but I'll see what happens.

File: jest.config.js [hunk 23]
```diff
@@ -21,6 +20,10 @@ module.exports = {
     // Use the client-side logger by default for tests.
     '^core/logger$': '<rootDir>/src/core/client/logger',
     '^core/window$': '<rootDir>/src/core/browserWindow',
+    // Alias tests for tests to be able to import helpers.
+    '^tests/(.*)$': '<rootDir>/tests/$1',
+    // Alias package.json so it can be imported.
+    '^package$': '<rootDir>/package.json',
```
kumar303: Why does `package` need to look like a module? I think this is confusing because when you look at `import package from 'package'` it's not obvious where that comes from. Can't we just do `JSON.parse(readFileSync(...))`?

jest.config.js
```diff
@@ -21,6 +20,10 @@ module.exports = {
     // Use the client-side logger by default for tests.
     '^core/logger$': '<rootDir>/src/core/client/logger',
     '^core/window$': '<rootDir>/src/core/browserWindow',
+    // Alias tests for tests to be able to import helpers.
+    '^tests/(.*)$': '<rootDir>/tests/$1',
```
kumar303: 👍

muffinresearch:
> Unless there is a compelling reason, I'd rather not introduce a magic package import.

Ok, will do it the boring way :)

muffinresearch:
@kumar303 Magic removed!

PR #2420, opened 2017-05-23
Prefix url

### LINKED ISSUE #2419: Privacy policy URL isn't prefixed
The "Privacy policy" link on https://addons-dev.allizom.org/en-US/android/addon/url-fixer/?src=ss and any other page is missing client app and locale.

Privacy Policy links to https://addons-dev.allizom.org/addon/url-fixer/privacy/ but should link to https://addons-dev.allizom.org/en-US/android/addon/url-fixer/privacy/

tofumatt:
Fixes #2419.

Simple fix to add clientApp and locales to `<Link href="/foo">` urls.

### DIFF
--- a/src/amo/components/Link.js
+++ b/src/amo/components/Link.js
@@ -19,11 +19,16 @@ export class LinkBase extends React.Component {
   }
 
   render() {
-    const { base, children, href, to } = this.props;
+    const { base, children, href, to, ...customProps } = this.props;
 
-    if (typeof href === 'string' && href.startsWith('/')) {
-      let linkHref = path.join(base, href);
-      return <Link {...this.props} href={linkHref}>{children}</Link>;
+    if (typeof href === 'string' && typeof to !== 'undefined') {
+      throw new Error(
+        'Cannot use "href" prop and "to" prop in the same Link component');
+    }
+
+    if (typeof href === 'string') {
+      const linkHref = href.startsWith('/') ? path.join(base, href) : href;
+      return <a {...customProps} href={linkHref}>{children}</a>;
     }
 
     let linkTo = to;
@@ -37,7 +42,7 @@ export class LinkBase extends React.Component {
       };
     }
 
-    return <Link {...this.props} to={linkTo}>{children}</Link>;
+    return <Link {...customProps} to={linkTo}>{children}</Link>;
   }
 }
 
--- a/src/amo/components/Link.js
+++ b/src/amo/components/Link.js
@@ -10,6 +10,7 @@ export class LinkBase extends React.Component {
   static propTypes = {
     base: PropTypes.string,
     children: PropTypes.node,
+    href: PropTypes.string,
     to: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
   }
 
@@ -18,7 +19,12 @@ export class LinkBase extends React.Component {
   }
 
   render() {
-    const { base, children, to } = this.props;
+    const { base, children, href, to } = this.props;
+
+    if (typeof href === 'string' && href.startsWith('/')) {
+      let linkHref = path.join(base, href);
+      return <Link {...this.props} href={linkHref}>{children}</Link>;
+    }
 
     let linkTo = to;
     if (typeof to === 'string' && to.startsWith('/')) {
--- a/tests/client/amo/components/TestLink.js
+++ b/tests/client/amo/components/TestLink.js
@@ -3,6 +3,7 @@ import {
   findRenderedComponentWithType,
   renderIntoDocument,
 } from 'react-addons-test-utils';
+import { findDOMNode } from 'react-dom';
 import { Link } from 'react-router';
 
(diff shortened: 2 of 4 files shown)


src/amo/components/Link.js
```diff
@@ -18,7 +19,12 @@ export class LinkBase extends React.Component {
   }
 
   render() {
-    const { base, children, to } = this.props;
+    const { base, children, href, to } = this.props;
+
+    if (typeof href === 'string' && href.startsWith('/')) {
+      const linkHref = path.join(base, href);
+      return <Link {...this.props} href={linkHref}>{children}</Link>;
```
kumar303: This adds too much duplicate code which introduces a maintenance burden and possibly a debugging headache one day.

I would like to see the calling code updated to pass in `to` like `<Link to={`/addon/${addon.slug}/privacy/`} ... />`.

If you really think we should support `href` then I'd like to ideally convert all `<Link>` calls to use `href` (could be done in a follow-up) and/or use more of an aliasing approach to cut down on duplicated code. For example:
````js
let linkTo = to;
if (linkTo === undefined and href !== undefined) {
  linkTo = href;
}
if (typeof linkTo === 'string' && linkTo.startsWith('/')) {
  // ...
}
````

Comment (tofumatt) [inline src/amo/components/Link.js]: My concern with `to` is that it will result in a 404 because those URLs don't exist in the front-end yet but it will try to load that route and 404. So I need this rendered as an `<a>` tag that doesn't use `react-router`.

I can add that as a comment and file an issue to remove it.

Alternatively I can create a mapStateToProps for the component relying on this and just use the `api` state, ditching the extra stuff for the Link component. Which do you reckon is best?

Comment (kumar303) [inline src/amo/components/Link.js]: Oh, sorry! I thought the new line was exactly the same as [the old line](https://github.com/mozilla/addons-frontend/blob/69501369755923cc1e7f8afe67b28941d60d0fb6/src/amo/components/Link.js#L40) but now I see that they use different attributes.

But now I'm totally confused because I don't see the `href` property documented for [`Link`](https://github.com/ReactTraining/react-router/blob/v2.8.1/docs/API.md#link). What does it do?

Comment (tofumatt) [inline src/amo/components/Link.js]: Ah, I guess it's not documented but it will output a regular a tag with a href attribute that won't try to do any client-side routing. It basically outputs an <a> tag that isn't handled by React

File: src/amo/components/Link.js [hunk 2]
```diff
@@ -18,7 +19,12 @@ export class LinkBase extends React.Component {
   }
 
   render() {
-    const { base, children, to } = this.props;
+    const { base, children, href, to } = this.props;
```
kumar303: I think it would be helpful to throw an error if *both* `to` and `href` are defined because that would probably be a mistake. I can see making that mistake if I was changing a lot of anchor tags to `Link`s maybe.

src/amo/components/Link.js
```diff
@@ -18,7 +19,12 @@ export class LinkBase extends React.Component {
...
+
+    if (typeof href === 'string' && href.startsWith('/')) {
+      const linkHref = path.join(base, href);
+      return <a {...this.props} href={linkHref}>{children}</a>;
```
kumar303: I don't think spreading `this.props` is correct here. I think you meant to allow custom props to be passed along, right? If so, an approach like this would be safer:
````js
const { base, children, href, to, ...customProps } = this.props;
// ...
return <a {...customProps} href={linkHref}>{children}</a>;
````

File: tests/client/amo/components/TestLink.js [hunk 4]
```diff
@@ -90,4 +90,23 @@ describe('<Link />', () => {
...
+
+    assert.notInclude(findDOMNode(root, 'a').href, '/base/');
+    assert.include(findDOMNode(root, 'a').href, 'test');
+  });
+
+  it('normalizes the href path with a string', () => {
+    const root = render({ base: '/foo/', href: '/test' });
+
+    assert.include(findDOMNode(root, 'a').href, '/foo/test');
+  });
```
kumar303: It would be nice to add a test that ensures the children of `<a ...>{children}</a>` are rendered.

src/amo/components/Link.js
tofumatt: Ah, fair enough. The `Link` component doesn't do that but no reason we can't be stricter 👍

src/amo/components/Link.js
tofumatt: Oops, indeed! Shall do.

src/amo/components/Link.js
```diff
@@ -18,7 +19,20 @@ export class LinkBase extends React.Component {
   }
 
   render() {
-    const { base, children, to } = this.props;
+    const { base, children, href, to, ...customProps } = this.props;
+
+    if (
+      typeof href === 'string' && href.startsWith('/') &&
```
kumar303: Looks like you can remove this check since it's not used in the if statement below anymore: `href.startsWith('/')`

PR #2427, opened 2017-05-24
Extensions categories have layout issues for RTL lang

### LINKED ISSUE #2168: Extensions categories have layout issues for RTL lang
Steps to reproduce:
1. Load Extensions categories page on your device https://addons-dev.allizom.org/en-US/android/extensions/categories/
2. Change the language to a RTL one i.e. https://addons-dev.allizom.org/ar/android/extensions/categories/

Expected results:

Actual results:
Categories arrows are not RTL.

Notes/Issues:
Verified on AMO-dev FF52(Win 7).
Screenshot for this issue:

tsl143:
Fixes #2168

This is my version of solution, hope this helps.

### DIFF
--- a/src/amo/components/Categories.scss
+++ b/src/amo/components/Categories.scss
@@ -10,6 +10,10 @@
   list-style: none;
   margin: 0 auto;
   padding: 0;
+
+  [dir=rtl] & {
+    transform: scale(-1, 1);
+  }
 }
 
 .Categories-link {
@@ -19,6 +23,6 @@
   text-decoration: none;
 
   [dir=rtl] & {
-    padding-right: 1em;
+    transform: scale(-1, 1);
   }
 }
--- a/src/amo/components/Categories.scss
+++ b/src/amo/components/Categories.scss
@@ -12,7 +12,8 @@
   padding: 0;
 
   [dir=rtl] & {
-    transform: scale(-1, 1);
+    background-image: url('../img/icons/arrow-left.svg');
+    background-position-x: 0%;
   }
 }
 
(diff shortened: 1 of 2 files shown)


tsl143:
Before
before

After
after

tsl143:
Before

before

After
after

src/amo/components/Categories.scss
```diff
@@ -10,11 +10,19 @@
   list-style: none;
   margin: 0 auto;
   padding: 0;
+
+  [dir=rtl] & {
+    transform: scale(-1, 1);
```
tofumatt: We actually don't need to scale the entire element; that's more effort on the renderer and could change other content. We should be changing the background element used and changing its position. eg:

```scss
[dir=rtl] & {
  background-image: url('../img/icons/arrow-left.svg');
  background-position-x: 0%;
}
```

You'll need to create an inverted arrow icon if we don't have one. Shame we aren't using an icon component here...

tsl143:
@tofumatt , I created the arrow-left.svg by changing the coordinates of arrow.svg, will that do?

muffinresearch:
I would suggest moving the svg background to be in generated content for both cases and then just transform that for RTL which saves duplicating the images and separates the arrow from the content.

tsl143:
@tofumatt @muffinresearch 
So whats the final decision?
Shall I go for new svg image or transforms?

PS : I prefer transforms to save redundancy.

tofumatt:
@muffinresearch is right, better to transform, but only transform the SVG and not the entire element.

tsl143:
I am not sure how to transform just the background image. Or Is it creating new image by using transform in polygon tag in svg?
@tofumatt

tsl143:
Tried with pseudo elements too, https://github.com/mozilla/addons-frontend/pull/2439

muffinresearch:
Superceded by #2439

PR #2439, opened 2017-05-25
pseudo element approach

### LINKED ISSUE #2168: Extensions categories have layout issues for RTL lang

tsl143:
Fixes #2168

### DIFF
--- a/src/amo/components/Categories.scss
+++ b/src/amo/components/Categories.scss
@@ -10,6 +10,10 @@
   list-style: none;
   margin: 0 auto;
   padding: 0;
+
+  [dir=rtl] & {
+    background: none;
+  }
 }
 
 .Categories-link {
@@ -17,4 +21,13 @@
   display: block;
   padding: 1em 0;
   text-decoration: none;
+
+  [dir=rtl] &::before {
+    background: url('../img/icons/arrow.svg') 100% 50% no-repeat;
+    content: '';
+    float: left;
+    height: 19px;
+    transform: scale(-1, 1);
+    width: 11px;
+  }
 }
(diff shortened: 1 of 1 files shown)


muffinresearch:
If you use generated content here: https://github.com/mozilla/addons-frontend/pull/2439/files#diff-0e3835cf21cbc3cf9ec95501830508b1L7 then you only need a transform for the rtl case.

tsl143:
@muffinresearch  Removed background image from `.Categories-list-item` and added pseudo element on `.Categories-link`

src/amo/components/Categories.scss
```diff
@@ -17,4 +16,17 @@
   display: block;
   padding: 1em 0;
   text-decoration: none;
+
+  &::before {
+    background: url('../img/icons/arrow.svg') 100% 50% no-repeat;
+    content: '';
+    float: right;
```
muffinresearch: I think if you set `display` to `inline-block` it might negate the need for floats.

tsl143:
@muffinresearch But implementing `inline-block` will incur alignment (both vertical and left/right)

muffinresearch:
> @muffinresearch But implementing inline-block will incur alignment (both vertical and left/right)

Yep, now I've spun it up and look at it in practice you're absolutely right. Sorry for the noise! This is ready to 🚢

tsl143:
Awesome :)

caitmuenster:
Woohoo! You're on fire this month, @tsl143! Thanks for your contributions!

PR #2444, opened 2017-05-25
fix: Protect against empty category results (fix #2443)

### LINKED ISSUE #2443: Categories reducer doesn't check for empty result
https://github.com/mozilla/addons-frontend/blob/1bc558d7b83ca9d3dd6e0bd793e42949a95edee5/src/core/reducers/categories.js#L44 doesn't check if `result` isn't `null` which causes our tests to complain more than they should.

Also, `result` should really be `category` 😄

tofumatt:
This fixes #2443 and just stops spamming the logs, really. It shouldn't really happen outside tests anyway.

### DIFF
--- a/src/core/reducers/categories.js
+++ b/src/core/reducers/categories.js
@@ -38,23 +38,30 @@ export default function categories(state = initialState, action) {
     case CATEGORIES_LOAD:
       {
         const categoryList = emptyCategoryList();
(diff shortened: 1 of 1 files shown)


tofumatt:
Feel free to wait until Jest is pushed and I'll rebase against it, though these checks are tested implicitly elsewhere so I didn't explicitly test them...

Issue #2450, opened 2017-05-25
Document new test setup with Jest


kumar303:
Document the new hotness for running tests now that we have https://github.com/mozilla/addons-frontend/issues/2372

Issue #2223, opened 2017-04-03
Re-organize project documentation


kumar303:
We should fix up our project docs. Here are some ideas:
* Move as much as possible out of README and into [docs](https://github.com/mozilla/addons-frontend/tree/master/docs)
* Move `docs/index.md` to `docs/README.md` so that it shows up on github better
* ~~Move contributor guide type of stuff to `CONTRIBUTING.md`~~ or perhaps just put a link in `CONTRIBUTING.md` to `docs/` ?

kumar303:
See https://github.com/mozilla/addons-frontend/pull/2699 for the contributing doc.

PR #2451, opened 2017-05-25
Document the new test setup with Jest

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2450


README.md
```diff
@@ -152,12 +161,16 @@ function getAllAddons({ categoryId }: GetAllAddonsParams = {}) {
...
+````
+yarn test-coverage
+````
 
-    open ./coverage/index.html
+This will print a table of files showing the percentage of code coverage.
+The uncovered lines will be shown in the right column but you can open
+the full report in a browser:
+
+    open coverage/lcov-report/index.html
```
tofumatt: This only works on Mac OS, for what it's worth. Also might be nice to be consistent with block code style as the other ones use the tilde.

README.md
```diff
@@ -202,7 +215,7 @@ module.exports = {
 When you start up your front-end discover pane server, it will now apply
```
tofumatt: This should be "front-end Discovery Pane" instead of "discover pane".

README.md
```diff
@@ -350,3 +363,5 @@ still can.
 * Code written in ES2015+
 * Universal rendering via node
 * Unit tests with high coverage (aiming for 100%)
+
+[jest]: http://facebook.github.io/jest/docs/en/getting-started.html
```
tofumatt: Should be HTTPS.

tinfoil-hat

docs/adding-a-page.md
```diff
@@ -98,7 +98,7 @@ export default (
 
 ### Starting the server
 
-Now that the component is setup we can run `npm start dev:search` and navigate to
+Now that the component is setup we can run `yarn dev:search` and navigate to
```
tofumatt: Technically this page is very out of date since it still references the search app which is now admin (which we're likely deleting). But c'est la vie!

README.md
```diff
@@ -23,68 +23,77 @@ Never submit security-related bugs through a Github Issue or by email.
...
-#### Run all unit tests in a loop
+Here are a few tips:
 
-You can use `npm run unittest:dev` to run all unit tests in a loop while you
-edit the source code.
+* When you start `yarn test`, you can switch to your code editor and begin
+  adding test files or changing existing code. As you save each file, [jest][]
+  will only run tests related to the code you change.
+* If you had typed `a` when you first started then [jest][] will continue to
+  run the full suite even when you change specific files. Type `o` to switch
```
muffinresearch: It does sort of do that if you run watch without pressing anything, since it seems to look at what changed since the last commit.

docs/adding-a-page.md
muffinresearch: Yeah I would remove any references to search (and admin) since it won't work the app is called admin, and the app is very close to being deleted.

README.md
```diff
@@ -23,68 +23,77 @@ Never submit security-related bugs through a Github Issue or by email.
...
-| npm run lint            |  Runs all the JS + SCSS linters                       |
-| npm run version-check   |  Checks you have the minimum node + npm versions      |
-| npm test                |  Runs the unittest, servertests + lint                |
-| npm run unittest        |  Runs just the unittests                              |
-| npm run unittest:dev    |  Runs the unittests and watches for changes           |
-| npm run unittest:server |  Starts a unittest server for use with `unittest:run` |
-| npm run unittest:run    |  Executes unittests (requires `unittest:server`)      |
-| npm run servertest      |  Runs the servertests                                 |
+* type `yarn` to install all dependencies
+* type `yarn dev:amo` to start a dev server
```
muffinresearch: That's an interesting idea. I also wondered about us considering making the addons-frontend repo a single app and binning the separation between the disco pane and amo once admin is gone. It would remove a number of complications though we would need to retain them as separate entry points from a webpack perspective. That's a discussion for another time though.

README.md
```diff
@@ -23,68 +23,77 @@ Never submit security-related bugs through a Github Issue or by email.
 
 ## Requirements
 
-* Node 6.x LTS
+* You need [Node](https://nodejs.org/) 6.x which is the current
+  [LTS](https://github.com/nodejs/LTS) (long term support) release.
+* To make life easier, also install [yarn](https://yarnpkg.com/en/).
```
muffinresearch: Since we're pushing yarn as the command runner and dep installer it would make sense to make it sound a bit less optional (or alternatively clarify how to run commands with `npm run` in its place)

README.md
```diff
@@ -23,68 +23,77 @@ Never submit security-related bugs through a Github Issue or by email.
...
+* If you see something like `Error watching file for changes: EMFILE` on Mac OS
+  then `brew install watchman` might fix it.
+  See https://github.com/facebook/jest/issues/1767
 
-#### Run a subset of the unit tests
+#### Run a subset of the tests
 
-If you don't want to run the entire unit test suite, first you have to start a
-unittest server:
+By default, `yarn test` will only run a subset of tests that relate to the code
```
muffinresearch: As mentioned, that depends. It seems to default to running anything it finds since the last commit. But if there is nothing it just sits there waiting for a prompt.

tofumatt: Neat. I didn't realise that. Might be worth quickly pointing that out in the docs then, but that sounds good.

README.md
muffinresearch: FWIW looks like a "open coverage" key option will be available directly from jest in the not too distant.

README.md
muffinresearch: Seems to be only available under http

tofumatt: Huh, that's weird, GitHub pages I thought had HTTPS 😞

README.md
kumar303: > Is there a way to launch in o mode by default?

It does launch in this mode by default :) However, if you press the `a` key (which intuitively is what I thought to do) then it will switch you out of the `o` mode so just don't do that.

tofumatt:
ahhhh, cool :-)

README.md
muffinresearch: > Huh, that's weird, GitHub pages I thought had HTTPS 😞

Ah no you were right! For some reason I tried https://facebook.github.io/jest/docs/en/getting-started.html and got the http version, maybe that was the awesome bar correcting me if I'd been linked to that in the past? ¯\_(ツ)_/¯

kumar303: An https link works for me. Sure, I can change it.

tofumatt:
Oh, yeah, it’ll do that.

docs/adding-a-page.md
kumar303: This whole guide is based around the search app. I'm not going to spend time on updating it for this patch. TBH I was just replacing all instances of `npm run`.

README.md
kumar303: sure

docs/adding-a-page.md
tofumatt: Yeah I just sort of was pointing it out to keep in mind... no need to update it right now, there are other issues for that.

Sorry for the noise!

muffinresearch: Fwiw `yarn dev:search` won't even run though.

kumar303: https://github.com/mozilla/addons-frontend/issues/2223 would be a good place to fix up this guide

Issue #2459, opened 2017-05-26
Add build step to travis


muffinresearch:
Since Jest side-steps webpack we should add a build step to the list of travis tasks so that we know if the build breaks.

PR #2463, opened 2017-05-26
Simplify the TravisCI build command

kumar303:
Follow up to #2459

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ node_js:
 - '6'
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


muffinresearch:
Yep that's way nicer. Be sure to double-check scripts are actually running for each part of the matrix.

r+ if green

kumar303:
Oh my god, why doesn't TravisCI document this crap! Finally got it to work.

kumar303:
yak-shaving-day_480x360

Issue #2373, opened 2017-05-11
Remove the admin app


kumar303:
As [discussed](https://groups.google.com/a/mozilla.com/d/topic/addons-dev-internal/ZSbSnYJa9GA/discussion) on `addons-dev-internal`, we decided it's a good idea to remove the admin app. Here's why:

* The code is outdated enough to where adding new features to it would pose a burden
  * A lot of modules were copy/pasted rather than shared so there is a lot of bitrot
* No one is using the admin app in its current form
* It's slowing down our test suite
* If we are able to spin it up again to make it useful, it will be easier to take a fresh approach to it using the current `addons-frontend` idioms and patterns
* We can always reference the old code

muffinresearch:
Blocked by https://github.com/mozilla/addons-frontend/issues/2385

PR #2466, opened 2017-05-26
Remove the admin app

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2373


kumar303:
@muffinresearch d'oh. We'd need to bump the coverage threshold down to to 93% for this since so much code was deleted.

muffinresearch:
> @muffinresearch d'oh. We'd need to bump the coverage threshold down to to 93% for this since so much code was deleted.

I think it's more that the test coverage decreased since we are setup to not allow decreases at all. 

Would it take much to cover the missing parts?

kumar303:
> Would it take much to cover the missing parts?

I haven't looked into it but it would be annoying to take time away from more pressing code quality issues. I'd like to do it eventually, for sure.

muffinresearch:
> I haven't looked into it but it would be annoying to take time away from more pressing code quality issues. I'd like to do it eventually, for sure.

If you want to leave it as a follow-up then it's fine, but I would have thought that 0.4% drop should not amount to too many lines needing coverage.

r+wc

more-red-than-green-award-sm

tofumatt:
> More red than green

An understatement. 👍

PR #2474, opened 2017-05-29
Add --watch with coverage reports to package.json

### LINKED ISSUE #2473: Restore ability to watch tests and generate coverage reports
After moving to Jest (#2372) we lost the ability to watch tests and build coverage at the same time. I found this very useful as I'll often run the tests in watch mode to regenerate coverage reports and find where I'm missing lines and if I'm testing this appropriately.

I'll have a patch to restore this in a few minutes. 😄

tofumatt:
Closes #2473.

I wanted this functionality back as I find it handy. Also changed up the name of the tests (and adjusted the CI command to run it). A few tweaks to the README while I was at it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,22 +40,22 @@ The easiest way to manage multiple node versions in development is to use
 
 Here are some commands you can run:
 
-| Command              | Description                                                                     |
-|----------------------|---------------------------------------------------------------------------------|
-| yarn dev:admin       |  Starts the dev server (admin app)                                              |
-| yarn dev:amo         |  Starts the dev server and proxy (amo)                                          |
-| yarn dev:amo:no-proxy|  Starts the dev server without proxy (amo)                                      |
-| yarn dev:disco       |  Starts the dev server (discovery pane)                                         |
-| yarn flow:check      |  Check for Flow errors and exit                                                 |
-| yarn flow:dev        |  Continuously check for Flow errors                                             |
-| yarn eslint          |  Lints the JS                                                                   |
-| yarn stylelint       |  Lints the SCSS                                                                 |
-| yarn lint            |  Runs all the JS + SCSS linters                                                 |
-| yarn version-check   |  Checks you have the required dependencies                                      |
-| yarn test            |  Enters [jest][] in watch mode                                                  |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,22 +40,22 @@ The easiest way to manage multiple node versions in development is to use
 
 Here are some commands you can run:
 
-| Command              | Description                                                                     |
```
tofumatt: This whitespace makes it hard to edit (and harder to read) the doc in an 80 char editor and doesn't add anything, so I removed it.

README.md
```diff
@@ -40,22 +40,22 @@ The easiest way to manage multiple node versions in development is to use
...
-| yarn dev:amo:no-proxy|  Starts the dev server without proxy (amo)                                      |
-| yarn dev:disco       |  Starts the dev server (discovery pane)                                         |
-| yarn flow:check      |  Check for Flow errors and exit                                                 |
-| yarn flow:dev        |  Continuously check for Flow errors                                             |
-| yarn eslint          |  Lints the JS                                                                   |
-| yarn stylelint       |  Lints the SCSS                                                                 |
-| yarn lint            |  Runs all the JS + SCSS linters                                                 |
-| yarn version-check   |  Checks you have the required dependencies                                      |
-| yarn test            |  Enters [jest][] in watch mode                                                  |
-| yarn test-ci         |  Runs all continuous integration checks. This is only meant to run on TravisCI. |
```
kumar303: It looks like `test-ci` got deleted. Can you put it back?

tofumatt:
Ah, I removed it from the readme as it's only intended for CI use and not usual developers... So I felt it was just noise in the docs (and what it does is evident in the package.json where other undocumented commands are).

Is it worth restoring for routine development?

kumar303:
> I felt it was just noise in the docs (and what it does is evident in the package.json where other undocumented commands are).

One day some developer is going to be debugging a problem in CI and think, huh, how does CI work anyway? Thus, I think it's worth documenting.

As for the other undocumented commands, I guess if they're just supporting commands that are used to DRY up the rest then we should prefix them with an underscore or something. But that can be done in another patch.

tofumatt:
Alrighty then, I will restore it :-)

Matthew Riley MacPherson

Issue #2477, opened 2017-05-29
Add "Explore", "Extensions", and "Themes" to header/search context


tofumatt:
We need the "Explore", "Extensions", and "Themes" links in the header and for them to be contextually active based on where you are on the site.

To do this we'll need to implement #2468 first.

ValentinaPC:
Verified as fixed on AMO-dev FF53(Android 7.1.2)
Postfix screenshots:
![header](https://user-images.githubusercontent.com/15685960/27170562-0457c006-51b7-11e7-8663-865038459a07.png)

@tofumatt : I have 2 questions for you
1. "empty" search is not possible while in Explore, but possible while in Extensions or Themes - is this ok?
2. while in Extensions or Themes homepages, the titles of every group of extensions (Featured extensions for example) is not centered - is this intended?

Thanks!

tofumatt:
To answer:

1. Ah yes, this is because technically searching from the extensions and themes pages, `type=extension` is added to the search query and thus the component doesn't see it as an empty search. I think this is okay... but if you wanted to file an issue about it that's alright, we could improve the handling of that
2. Yes, the new mocks shifted the headers in the cards to the left rather than centre.

ValentinaPC:
> Yes, the new mocks shifted the headers in the cards to the left rather than centre.

Shouldn't be aligned with extensions names?

Thanks for your reply!

tofumatt:
When #2560 is complete the extension name/content should look as intended but the header is where it is supposed to be as per the mocks. It looks a bit odd now but should be fixed once #2560 is ready.

Issue #2468, opened 2017-05-27
Store current addon type for addon/context in store


tofumatt:
### Describe the problem and steps to reproduce it:
The new header (#2436) highlights Explore, Extensions, or Themes based on the URL currently, but it should really use the current "context" of add-on type we're viewing. Most of the time our URLs provide the context for the add-on type we're exploring (Extension or Theme) but once we go to the Add-on detail page that info isn't in the URL so we should store it (or check it from the `SectionLinks` component) and use that to inform which link we highlight.

vcarciu:
Can you please add some STRs to this issue or mark it as "qa: not needed" ?

tofumatt:
You should see the explore, extensions, and themes in the header and in the search placeholder.

vcarciu:
@tofumatt : then we can mark this as Verified after responding to questions from https://github.com/mozilla/addons-frontend/issues/2477

tofumatt:
Questions answered, cheers!

vcarciu:
Thanks!

PR #2478, opened 2017-05-29
feat: Implement Desktop header

### LINKED ISSUE #1609: Re-implement the configuration for auto-prefixing (via postcss-loader)
Since the post-css auto-prefixing config was originally setup postcss-loader looks like it has changed. As a result it's likely that the auto-prefixing is not working additionally this was causing an error in the uitests. (See https://github.com/mozilla/addons-frontend/pull/1596)

`postcss-loader` allows for a standalone config file. It also allows for config via webpack. 

We should remove the standalone config file and re-implement/check the configuration via webpack and make sure amo gets cross-browser css prefixes when built, and disco pane doesn't (because the audience for the disco pane is just Firefox).

If postcss-loader gets the config as expected then it should stop complaining about the missing config.

A boolean config value controls the setup of the config here: https://github.com/mozilla/addons-frontend/blob/master/webpack.prod.config.babel.js#L141

The current docs for postcss setup are here: https://github.com/postcss/postcss-loader#webpack-1-1

### LINKED ISSUE #2436: Create a Desktop Header (for detail and landing pages)
Create a desktop header for inner/detail and one for landing pages.

### Landing Pages
screenshot 2017-05-25 16 05 20

### Detail Pages
screenshot 2017-05-25 13 43 22

### LINKED ISSUE #2438: Create a better, more flexible button element using the new specs
Our current usage of "button" elements is a bit all-over-the-map and leading to it being frustrating to add new ones for the desktop header. I have specs for all possible button colours and sizes and we should make ours consistent and a bit easier to work with.

This might mean using some common classes for them rather than our current mixins directly which leads to slightly different buttons everywhere...

### LINKED ISSUE #2467: <Link> component `href` prop doesn't allow unprefixed links
Try to use `<Link href="/en-US/developers/">Developer Hub</Link>` and it will outptut `<a href="/en-US/firefox/en-US/developers/">Developer Hub</a>`. But in this case I want it to output `<a href="/en-US/developers/">Developer Hub</a>`.

Easy fix, just need to add a `prefix` prop that can be set to `false` to disable prefixing. I've actually done this in my #2436 patch, just filing here for posterity and to remind me to write extra tests.

tofumatt:
Add a desktop header (without the section links, they're coming in #2477 because we need #2468 first and I didn't want this patch to be too huge).

* Fixes #1609
* Fixes #2436
* Fixes #2438 
* Fixes #2467

-----

### Desktop
screenshot 2017-05-30 00 39 23

### Mobile
screenshot 2017-05-30 00 40 37


src/amo/components/LanguagePicker.scss
```diff
@@ -7,6 +7,8 @@ $selector-padding: 10px;
 }
 
 .LanguagePicker-selector {
+  -moz-appearance: none;
```
tofumatt: These weren't being generated locally but I'll double-check to see if I can reproduce that. Maybe it was a weird local config error...

tofumatt:
r?

This had a fair amount of lines changed but a lot of it is adding a tonne of button/spec variables and the like.

tofumatt:
The only test failure here is in the admin app which will soon be removed. So I'ma just wait until #2466 is merged and rebase on top of it.

src/admin/components/SearchForm.js
```diff
@@ -56,16 +55,16 @@ export class AdminSearchFormBase extends React.Component {
...
-        <button className="button button-middle" type="submit" title="Enter"
+        <Button className="button-middle" type="submit" title="Enter"
                 ref={(ref) => { this.submitButton = ref; }}
-                onClick={this.handleSearch}>
+                onClick={this.handleSearch} styleName="action">
           {_('Search')}
-        </button>
-        <button className="button button-end button-inverse"
+        </Button>
+        <Button className="button-end" inverse styleName="action"
```
muffinresearch: My immediate reaction to this is I find this less instant than the previous classNames. I wonder if it might be simpler to abstract the `link` vs `button` but otherwise just pass things like className through.

Too many props might make it more complicated to understand e.g.`inverse` prop vs `styleName` etc.

src/amo/components/Header/index.js
```diff
@@ -0,0 +1,93 @@
...
+      i18n,
+      isHomePage,
+      lang,
+      location,
+      query,
+    } = this.props;
+    const headerLink = (
+      <Link className="Header-title" to="/">
+        <Icon className="Header-addons-icon" name="extensions-complex" />
+        {i18n.gettext('Firefox Add-ons')}
```
muffinresearch: Would be good to add a comment for l10ners about not translating "Firefox". This test in the extaction shows what is looked for: https://github.com/mozilla/babel-gettext-extractor/blob/master/test/test.js#L55

src/amo/components/Header/index.js
```diff
@@ -0,0 +1,93 @@
...
+      </Link>
+    );
+
+    return (
+      <div className="Header">
```
muffinresearch: Could this whole thing be a `<header>` rather than just below for the content?

src/amo/css/inc/vars.scss
```diff
@@ -14,7 +13,9 @@ $accent-color: #ebebeb;
 $breakpoints: (
   small: '(max-width: 499px)',
   medium: '(min-width: 500px)',
-  large: '(min-width: 720px)'
+  large: '(min-width: 720px)',
+  desktop: only screen and (min-width: 720px),
+  mobile: only screen and (max-width: 719px),
```
muffinresearch: On reflection I feel that the mix of device agnostic naming with device class naming might be confusing in the future once we have a mix of all of that across the code base. 

Maybe we should pick one, use that and not have a mix?

I'm not tied to the completely agnostic sizes but I think a single system will be better than two. 

My own two cents - I've used more of a mobile/desktop split in the past and what I don't like about that is when you might have a mobile that is actually going to use the desktop breakpoint the concept starts to fall apart. Although, I can buy it a little bit more if the intent is more generic, e.g. mobile and desktop is more a division between 100% width (e.g. mobile layout) and adding additional columns (desktop).

Maybe we only really need a couple of breakpoints though, because I would imagine the rest are going to be specific to the content and would therefore be defined specifically and sit with the related styles - they might actually? I would imagine we're not even using much of these small, medium and large breakpoints yet?

Also since the site is mobile first, there shouldn't be much of a use-case for using the small/mobile breakpoint since that *is* the default and only larger breakpoints should modify it.

Here's some additional food for thought: 

* https://medium.freecodecamp.com/the-100-correct-way-to-do-css-breakpoints-88d6a5ba1862
* https://seesparkbox.com/foundry/there_is_no_breakpoint

src/ui/components/Button/Button.scss
```diff
@@ -1,20 +1,51 @@
...
-  }
+.Button--styleName-action {
+  @include button(
+    $background: $action-base-color,
+    $background-active: $action-active-color,
+    $background-hover: $action-hover-color
+  );
+}
+
+.Button--styleName-confirm {
```
muffinresearch: I think this could be simpler as `.Button--confirm` sticking to the modifier convention.

tests/client/ui/components/TestButton.js
```diff
@@ -7,18 +7,22 @@ import Button from 'ui/components/Button';
...
+    expect(button.props.className).toContain('Button');
+    expect(button.props.className).toContain('Foo');
     expect(button.props.onClick).toBe(onClick);
     expect(button.props.children).toEqual('My button!');
   });
 
   it('renders a link with an href', () => {
     const href = 'https://addons.mozilla.org';
-    const button = shallowRender(<Button className="Bar" to={href}>Link text!</Button>);
+    const button = shallowRender(
```
muffinresearch: Maybe an opportunity to start using enzyme?

src/admin/components/SearchForm.js
tofumatt: Having anything more than a single way to make buttons seems to really lead to inconsistency and also a bit of confusion around making buttons. This way groking for buttons in the code base you'll find only one thing, because often when we make "buttons" we're really just talking style. Looking at the code in complete isolation it can be a bit weird (because it's just a capital letter away from an HTML component) but when you see it's imported from `ui/components` I think it makes more sense than the `className` approach.

src/amo/components/Header/styles.scss
```diff
@@ -0,0 +1,60 @@
...
+  background: $header-base-color;
+  min-height: 100px;
+  position: relative;
+}
+
+.Header-content {
+  padding: 10px 0;
+  text-align: center;
+
+  @include respond-to(mobile) {
```
muffinresearch: Mobile styling should be the default so this should be inverted.

src/amo/components/Link.js
```diff
@@ -12,33 +12,36 @@ export class LinkBase extends React.Component {
...
+      const linkHref = href.startsWith('/') && prefix ?
+        path.join(base, href) : href;
       return <a {...customProps} href={linkHref}>{children}</a>;
     }
 
     let linkTo = to;
-    if (typeof to === 'string' && to.startsWith('/')) {
-      linkTo = path.join(base, to);
+    if (typeof to === 'string') {
+      linkTo = to.startsWith('/') && prefix ? path.join(base, to) : to;
```
muffinresearch: Is there a need to handle protocol free links e.g. `//google.com` here? Just because that starts with `/` but isn't the same thing as a relative path.

src/amo/css/inc/vars.scss
tofumatt: Using a mobile breakpoint is handy for hiding things only on mobile rather than *showing* them on all other breakpoints, especially if we ever added new ones: we wouldn't have to go back and add them to lists of "all-but-mobile" `@include respond-to` blocks.

I am cool sticking with one way of defining breakpoint. I get these arguments but as we discussed I feel the frequent/usual *intent* of breakpoints is divorced from the recommendation in the abstract. We, ourselves, are always talking *mobile vs desktop* not *small vs large*.

I'd like to have another think about it, ultimately I'm okay with using small/medium/large... but I don't think that's usually how devs or designers are thinking. Even if it's more pure in concept, I actually think it requires a translation when speaking about intent.

src/ui/components/Icon/styles.scss
```diff
@@ -76,3 +80,11 @@
   min-width: 46.3px;
```
muffinresearch: Is this precision needed?

src/amo/components/Header/styles.scss
tofumatt: As mentioned above I'm not sure how I feel about that: this is something that only happens on the mobile breakpoint–were we to add new breakpoints we would need to go back and change everything that uses `respond-to(desktop)` even though this is something we only want to happen on very small screens.

src/amo/components/Link.js
tofumatt: That isn't an intent, no. The `to` attribute is really only for local links and I'm not in love with this behaviour but I'm just emulating what the original Link component does.

Maybe what we should really be doing is throwing an error about relative URLs because really we should never be using them.

src/ui/components/Icon/extensions-complex.svg
```diff
@@ -0,0 +1,165 @@
+
+<svg width="44px" height="48px" viewBox="0 0 44 48" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
muffinresearch: Presumably this replaces the fox icon everywhere? I think this one might need checking before we use this icon for the entire site.

src/ui/components/Icon/styles.scss
tofumatt: I'm guessing  it's a holdover from the whole "divide things by 3" issue with sketch files. Also maybe not but I didn't write that line 😉

I'll have a look though; `46.5` might be okay.

src/ui/components/Icon/extensions-complex.svg
tofumatt: This is a new icon as I didn't think it should replace the fox everywhere. This only replaces the header.

src/amo/components/LanguagePicker.scss
kumar303: We only [configure the auto-prefixer](https://github.com/mozilla/addons-frontend/blob/master/webpack.prod.config.babel.js#L91) for prod builds so you can test it there. Maybe our 'last 2 versions' setting needs to be tweaked?

kumar303: Actually, you can probably just set `enablePostCssLoader: true` in your local dev config.

src/ui/components/Button/index.js
```diff
@@ -9,31 +9,52 @@ import './Button.scss';
...
+    } = this.props;
     const props = {
-      className: classNames('Button', className, {
-        'Button--small': size === 'small',
-        'Button--light': appearance === 'light',
+      className: classNames('Button', className, [
+        `Button--size-${size}`,
+        `Button--styleName-${styleName}`,
+      ], {
+        'Button--inverse': inverse,
```
kumar303: what does inverse mean? Can there be a more descriptive name for this?

tests/client/ui/components/TestButton.js
kumar303: It might be a rabbit hole though 🔥

kumar303:
The `admin` rapture has taken all who may pass. Only sinners are left.

tofumatt:
Thanks for all the comments! I'm gonna do another pass removing admin stuff and incorporating feedback. Should be ready soon!

src/amo/components/LanguagePicker.scss
tofumatt: Ah, right on. Any reason we don't have that on in development? It's definitely handy and doesn't really add much bloat...

tests/client/ui/components/TestButton.js
muffinresearch: rabbit-hole

src/amo/components/LanguagePicker.scss
kumar303: I can't think of any reason. We should probably configure it on.

tofumatt: Cool, I'll just enable it and include it in this patch. If you'd rather it separate just let me know and I'll file an issue/PR.

src/amo/components/Header/index.js
```diff
@@ -0,0 +1,93 @@
...
+        <div className="Header-top-row">
+          <div className="Header-buttons">
+            {/*
+              We don't use the Link component here because we only want to
```
tofumatt: Yeah that's way better and means not having to map `state.api` to the `props` of the component using the `<Link>` too. Will do! 👍

tofumatt:
I did a fair bit of improvements to buttons, some tests, etc. and it ended up being a fair bit more work to polish. I think this is close now but I'm heading out and might not get to it until tomorrow morning. I'll update the request for review once I'm happy with it but it's nearly there!

src/ui/components/Icon/extensions-complex.svg
muffinresearch: As mentioned on irc - this new icon for the header isn't yet signed off so it might need a change before the tag. Phil is going to let us know before then.

tofumatt: Yeah, I think I'll actually just use the old one until we get sign-off and make that a separate bug 👍

src/amo/components/Header/index.js
tofumatt: `<header class="Header">`

src/amo/css/inc/vars.scss
tofumatt: fwiw I've switched to declaring only medium and large.

Having a mobile breakpoint is an anti-pattern; if we want to do mobile-first then we shouldn't have a way to target only mobile 😄

tests/client/ui/components/TestButton.js
tofumatt: Yeah gonna do that... in a different PR 😉

src/amo/components/Header/styles.scss
tofumatt: I've changed my tune, agreed.

src/amo/components/Header/index.js
```diff
@@ -0,0 +1,79 @@
...
+      i18n,
+      isHomePage,
+      location,
+      query,
+    } = this.props;
+    const headerLink = (
+      <Link className="Header-title" to="/">
+        <Icon className="Header-addons-icon" name="fox-light" />
+        {
+          // l10n: "Firefox" should not be translated. :-)
```
muffinresearch: This won't match the translator comment format currently. https://github.com/mozilla/babel-gettext-extractor/blob/master/index.js#L27

tests/client/amo/components/TestHeader.js
```diff
@@ -19,40 +19,41 @@ class FakeChild extends React.Component {
...
-    const root = renderMastHead({
+    const root = renderHeader({
       children: FakeChild,
       SearchFormComponent: FakeChild,
     });
-    const titleLink = findDOMNode(root).querySelector('.MastHead-title');
+    const titleLink = findDOMNode(root).querySelector('.Header-title');
     const h1Tag = findDOMNode(root).querySelector('h1');
 
     expect(!h1Tag).toBeTruthy();
```
muffinresearch: Might be more direct to use `.toBeFalsey()` and drop the `!`.

src/core/css/inc/mixins.scss
```diff
@@ -164,41 +163,71 @@ $default-arrow-margin: 3px;
...
+  .Icon {
+    @include margin-end(6px);
+
+    height: 16px;
+    vertical-align: top;
+    width: 16px;
+  }
+
+  &.Button--small {
+    &,
```
muffinresearch: What's this parent ref needed for?

src/ui/components/Button/Button.scss
```diff
@@ -1,20 +1,51 @@
...
+    $background: $neutral-base-color,
+    $background-active: $neutral-active-color,
+    $background-hover: $neutral-hover-color
+  );
 }
 
-.Button--light,
-.Button--light:link {
-  @include button(#fff, $link-color);
+.Button--light {
```
muffinresearch: This mixin usage seems to imply a lot of duplication. I wonder whether there's a better way using the cascade directly.

src/amo/components/Header/index.js
tofumatt: Ah, I think I copied it from elsewhere 😒

Yup: https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/ShowMoreCard/index.js#L60

Though I think that one was my fault too 😆

tofumatt: So I want:

```js
// translators: "Firefox" should not be translated. :-)
```

right?

src/ui/components/Button/Button.scss
tofumatt: I bet there is. I'd rather refactor it in a separate PR, but I think it can be improved.

tests/client/ui/components/TestButton.js
```diff
@@ -7,18 +7,22 @@ import Button from 'ui/components/Button';
 describe('<Button />', () => {
```
muffinresearch: In case we don't already would be nice to have a test to ensure that classNames does the right thing if a duplicate Button class is passed in.

src/amo/components/Header/index.js
muffinresearch: Yeah I think so, you can try extraction to be sure.

src/core/css/inc/mixins.scss
tofumatt: We need to target `&:link` explicitly because of all the styles we have that use `a:link` (which I think we do for a reason? but I forget it and find it really frustrating), but some Buttons are actual `button` tags so the `&` targets them.

src/amo/components/Header/index.js
muffinresearch: I'd be in favour of adding to what the extractor supports  too.

tests/client/amo/components/TestHeader.js
tofumatt: Good catch. Probably lazy copypasta here.

src/amo/components/Header/index.js
tofumatt: Alright I'll do that instead. I'd rather support `translators`, `l10n`, and `i18n`. But I'll update this for now so it's not blocked on that patch.

src/core/css/inc/mixins.scss
muffinresearch: Ah ok, maybe something to look at if cascade can be used in place of the mixin later.

src/amo/components/Link.js
```diff
@@ -9,37 +9,75 @@ import { Link } from 'react-router';
...
+  }
+
+  urlPrefix({ clientApp, lang, prependClientApp, prependLang } = {}) {
+    if (prependClientApp && prependLang) {
+      return `/${lang}/${clientApp}/`;
+    } else if (prependClientApp) {
+      return `/${clientApp}/`;
+    } else if (prependLang) {
+      return `/${lang}/`;
+    }
```
kumar303: This could be DRY'd up to:
```js
const prefix = [];
if (prependLang) {
  prefix.push(lang);
}
if (prependClientApp) {
  prefix.push(clientApp);
}
if (prefix.length) {
  return `/${prefix.join('/')}/`;
}
return null;
```

src/amo/components/Link.js
```diff
@@ -9,37 +9,75 @@ import { Link } from 'react-router';
...
+        (to && to.pathname && !to.pathname.startsWith('/'))
+      )
+    ) {
+      throw new Error(
+        '"to" prop cannot contain a relative path; it must start with a "/".');
+    }
+
     if (typeof href === 'string') {
-      const linkHref = href.startsWith('/') ? path.join(base, href) : href;
+      const linkHref = urlPrefix ? path.join(urlPrefix, href) : href;
```
kumar303: I don't think this should have ever used `path.join()` because that will use back slashes on Windows. It should always join a URL with forward slashes -- maybe with [`url.resolve()`](https://nodejs.org/api/url.html#url_url_resolve_from_to)?

src/amo/components/Link.js
```diff
@@ -9,37 +9,75 @@ import { Link } from 'react-router';
...
       return <a {...customProps} href={linkHref}>{children}</a>;
     }
 
     let linkTo = to;
-    if (typeof to === 'string' && to.startsWith('/')) {
-      linkTo = path.join(base, to);
+    if (typeof to === 'string') {
+      linkTo = urlPrefix ? path.join(urlPrefix, to) : to;
```
kumar303: Same problem here with `path.join()`.

src/amo/components/Link.js
```diff
@@ -9,37 +9,75 @@ import { Link } from 'react-router';
...
     } else if (to && to.pathname) {
       linkTo = {
         ...to,
-        pathname: to.pathname.startsWith('/') ?
-          path.join(base, to.pathname) : to.pathname,
+        pathname: urlPrefix ? path.join(urlPrefix, to.pathname) : to.pathname,
```
kumar303: `path.join()` again

src/core/css/inc/mixins.scss
```diff
@@ -77,8 +77,7 @@
       @content;
     }
   } @else {
-    @warn 'Unfortunately, no value could be retrieved from `#{$breakpoint}`. '
-      + 'Please make sure it is defined in `$breakpoints` map.';
+    @warn 'Breakpoint `#{$breakpoint}` not found in `$breakpoints` map.';
```
kumar303: It's so impolite now. jk

tests/client/ui/components/TestButton.js
tofumatt: Turns out `classNames` doesn't filter out duplicate classNames.

tofumatt: screenshot 2017-06-05 17 28 09

tests/client/amo/components/TestLink.js
```diff
@@ -4,119 +4,193 @@ import {
...
 
-import { LinkBase, mapStateToProps } from 'amo/components/Link';
+import Link, { mapStateToProps } from 'amo/components/Link';
+import createStore from 'amo/store';
+import { setClientApp, setLang } from 'core/actions';
 
 
 describe('<Link />', () => {
-  const defaultState = { api: { clientApp: 'android', lang: 'fr' } };
+  const { store } = createStore();
```
kumar303: I think it's a lot safer to wrap this kind of thing in `beforeEach()`. That way you can guarantee that no dispatched actions from one test will leak into another.

config/development.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
   cookieSecure: false,
 
   enableClientConsole: true,
+  enablePostCssLoader: true,
```
tofumatt: Good point actually. Will-do.

src/amo/components/Header/index.js
```diff
@@ -0,0 +1,79 @@
...
+            <AuthenticateButton
+              className="Header-auth-button Button--action Button--small"
+              location={location}
+              noIcon
+            />
+          </div>
+        </div>
+        <div className="Header-content">
+          {isHomePage
+            ? <h1 className="Header-title-wrapper">{headerLink}</h1>
```
tofumatt: Nope. The `<h1>` should be "the most important thing" on the page; the site's title is important on the homepage (a landing page) telling users "this is what this is" but on an add-on's page we'd want `<h1>My cool add-on name</h1>`.

You can technically have multiple `<h1>` tags in the page (especially in a `<header>` versus `<article>`, etc. but it's confusing and we don't need that amount of headers.

src/amo/components/Link.js
tofumatt: Wasn't sure if that was preferred but I like it better too. 👍

src/amo/components/Link.js
tofumatt: Shoot, good point. I'll tweak that.

src/core/css/inc/mixins.scss
tofumatt: I'm assuming a Canadian wrote the original one? @mstriemer?

config/development.js
tofumatt: Weird, it was actually in the default and off for disco.

src/core/css/inc/mixins.scss
mstriemer: Somewhat [surprisingly](https://github.com/mozilla/addons-frontend/commit/b4fbc2641061382b5d3fea8597bf52ad4e738e99) it wasn't a Canadian. I would have believed you though ;)

src/amo/components/Header/index.js
kumar303: Okay, makes sense. One other minor thing: `isHomePage` feels like an implementation detail. Calling it `headerIsH1` would make more sense to me.

src/ui/components/Button/index.js
```diff
@@ -9,31 +9,32 @@ import './Button.scss';
 
 export default class Button extends React.Component {
   static propTypes = {
-    appearance: PropTypes.string,
```
mstriemer: I'm curious about the move to specifying a `className` instead of an appearance. Since this component defines the class names I think it makes sense to instead ask it to use the right class rather than have each caller need to know the class to add.

If you don't like the `appearance` prop, you could have boolean props for the classes too.

```jsx
<Button alert>Cancel</Button>
<Button action>Continue</Button>

Comment (tofumatt) [inline src/ui/components/Button/index.js]: I actually agree–something like that was originally what I did (though I like your API better!) and now that you brought it up I'm seemingly not the only one. Knowing the classnames is weird but also verbose.

I'll move to your method... I like `<Button action>` a lot and it's easier to have saner defaults then without having to deal with managing `classNames`.

I might do it in another PR so this one doesn't take forever but 👍

Comment (kumar303) [inline src/ui/components/Button/index.js]: I agree. Asking each caller to know the class name puts an undue mental burden on them (they have to look it up in the css or try to find an example of a similar call). 

Also, with Flow annotations, an open ended string property like this is easier to deal with. For example, you can glance at the props type and know what to do:

```js
type ButtonProps = {|
  appearance: 'light' | 'dark',
  ...
|};

export class Button extends React.Component {
  props: ButtonProps;
  ...
}
```

mstriemer: Yeah, the `appearance` version isn't quite as clean as the boolean props but it makes it a lot easier to implement within the component.

There's probably something you could do to avoid writing a bunch of `classNames({ 'Button--light': this.props.light, 'Button--dark': this.props.dark, 'Button--enough-already': ... })` though.

tofumatt: I definitely prefer @mstriemer's API (`<Button alert small>Foo</Button>`)... I don't care if the internal implementation for the prop checking isn't as nice because we use that less than the component itself. The only benefit to `appearance` I can see is easily enforcing only one type of appearance mod... eg you can't have `appearance: 'small large light dark'`.

But then we can just do that with something like:

```js
if ([props.light, props.dark, props.small, props.large].reduce((prop) => return !!prop; )) {
  throw new Error('Conflicting style properties supplied to Button component');
}
```

Which I think is fine. If we can do that with Flow all the better...

src/amo/components/Header/index.js
tofumatt: The thing is that `isHomePage` is something that the `currentView` reducer (#2514) is going to implement. Basically we'll know: "What is the current addonType we're looking at?", "Are we in the Explore mode?", and "Are we on the home page?"

I don't feel like it's an implementation detail; it's a special context that is worth knowing about for certain components, so I think it's worth maintaining.

src/amo/components/Link.js
tofumatt: Sadly `url.resolve` doesn't do that... nothing in node does actually, and people have been asking for it... but there's an npm library that does I've added.

src/amo/components/Addon/index.js
```diff
@@ -104,14 +105,16 @@ export class AddonBase extends React.Component {
           ref={(el) => { this.wrapper = el; }}
           onClick={this.onClick}
         >
-          {installStatus !== ENABLED ?
-            <button
+          {installStatus !== ENABLED ? (
+            <Button
+              className="AddonDetail-theme-header-label Button--neutral"
```
kumar303: This needs to be `Addon` after my rename. Sorry not sorry 🌵 🍰

src/amo/components/Addon/index.js
```diff
@@ -104,14 +105,16 @@ export class AddonBase extends React.Component {
...
               disabled={!compatible}
-              className="Button Addon-theme-header-label"
-              htmlFor="Addon-theme-header">
-              <Icon name="eye" className="Addon-theme-preview-icon" />
+              htmlFor="AddonDetail-theme-header"
```
kumar303: `Addon-theme-header`

src/amo/components/Addon/index.js
```diff
@@ -104,14 +105,16 @@ export class AddonBase extends React.Component {
...
+            >
+              <Icon name="eye" className="AddonDetail-theme-preview-icon" />
```
kumar303: Addon...

src/amo/components/Addon/styles.scss
```diff
@@ -70,7 +71,16 @@ $page-margin: 20px 10px;
   }
 }
 
-.Addon-theme-header {
+.AddonDetail-theme-header-label {
```
kumar303: Addon...

src/amo/components/Addon/styles.scss
```diff
@@ -70,7 +71,16 @@ $page-margin: 20px 10px;
...
+  $label-margin: 12px;
+
+  @include start($label-margin);
+
+  bottom: $label-margin;
+  position: absolute;
+}
+
+.AddonDetail-theme-header {
```
kumar303: Addon...

src/amo/components/Header/index.js
kumar303: ok, I'll take a look at it in the other patch

src/amo/components/Link.js
```diff
@@ -1,5 +1,4 @@
-import path from 'path';
-
+import join from 'join-url';
```
kumar303: what about calling this `joinUrl`? Otherwise `join` seems a bit too generic as a global module var. Not a big deal, just something I noticed when reading the code.

config/default-disco.js
```diff
@@ -57,6 +57,8 @@ module.exports = {
   // add custom dimension data.
   trackingSendInitPageView: false,
 
+  // This is disabled for the disco pane because Firefox is the only target
+  // so we don't need auto-prefixing, etc.
```
kumar303: Thanks, this comment is helpful.

src/ui/components/Button/index.js
```diff
@@ -9,31 +9,32 @@ import './Button.scss';
...
     children: PropTypes.node,
     className: PropTypes.string,
+    href: PropTypes.string,
     to: PropTypes.string,
-    size: PropTypes.string.isRequired,
   }
 
-  static defaultProps = {
-    appearance: undefined,
```
kumar303: Is there an issue on file to put something like `appearance` back? I would rather see it done in this patch because I think it will be hard to track down all the hard coded class names that are used in component instances. It would be ok to do it in a cleanup patch though.

src/ui/components/Button/index.js
kumar303: Using boolean props or a single appearance prop are both fine with me. I would just like to not see hard coded class names.

src/ui/components/Button/index.js
tofumatt: I added it to #2517. I would rather clean it up elsewhere as the classNames are actually pretty easy to track down and I think it will be easier/nicer as a separate patch. There was already some inconsistency in our button usage and I cleaned that up in this patch. The next one can deal with the `<Button>` prop API.

src/amo/components/Link.js
tofumatt: Works for me actually, will do. This is just copied from their example, I didn't give it any thought but you're right.

src/ui/components/Button/index.js
muffinresearch: > I agree. Asking each caller to know the class name puts an undue mental burden on them (they have to look it up in the css or try to find an example of a similar call).

@kumar303 Heh it's interesting to see the different viewpoints. 

FWIW I argued in the opposite direction. Having multiple props (as this did earlier) can create ambiguity about what they affect unless it's very clear through naming what changes style versus other attributes of the button. 

Simply passing through the className prop unambiguously informs the developer what affects the styling and it makes clear the relationship between the JSX markup and the CSS without having to inspect it running on a page or drill into the component code to know what's what.

Using props to control styles used adds a layer of indirection as opposed to following a long established web convention.

If the props route is seen as the way forward I think it would be important to establish some  conventions around naming and have decent docs so it's easy for anyone to know what changes will provide the necessary result easily.

For a real-world example here's how react-bootstrap use props. https://react-bootstrap.github.io/components.html#buttons-props. It's not too bad if you have the docs to hand and they have some naming conventions to attempt make it a bit more intuitive.

It's worth trying it because (and maybe this is somewhat related to familiarity) but react-bootstrap seemed to be more annoying to use than the non-react version that just used structured classnames.

PR #2480, opened 2017-05-30
Fallback to number.toLocaleString when Intl not available

### LINKED ISSUE #2470: Tapping is only working after second attempt on AMO-dev
Steps to reproduce:
1. Load https://addons-dev.allizom.org/en-US/android/ on your Android device
2. Tap on "Extensions"

Expected results:
"Extensions" page is loaded.

Actual results:
A blue loading bar is displayed, while loading the "Extensions" page is not finalized.

Notes/Issues:
- refreshing the page loads the desired page.
- happens with every link from homepage, or extensions/themes categories
- tapping again on the item, loads the page
- no errors displayed in the console

Verified on AMO-dev FF53(Android 7.1.2). Issue is only reproducing on DEV.
Video for this issue:
videotogif_2017 05 29_10 29 58

### LINKED ISSUE #2471: Installation is not working for themes/extensions on AMO-dev
Steps to reproduce:
1. Load any theme/extensions details page on your Android device i.e. https://addons-dev.allizom.org/en-US/android/addon/shamrocks2/
2. Observe the install button

Expected results:
The installation button is enabled.

Actual results:
The installation button is disabled.

Notes/Issues:
- not reproducing on desktop while in Responsive Design Mode - View Mobile Site
- only reproducing on AMO-dev
-  "tap to preview" button is not working also
- all `prefs` are set

Verified on FF53(Android 6.0.1, 7.1.2).
Video for this issue:
videotogif_2017 05 29_12 41 06

### LINKED ISSUE #2472: "Read more" link is not available in add-on details page
Steps to reproduce:
1. Load an add-on details page that has a long description on your device i.e. https://addons-dev.allizom.org/en-US/android/addon/quitnow/
2. Check out the "Read More" link

Expected results:
Link is displayed and working.

Actual results:
Link is not displayed.

Notes/Issues:
Verified on FF53(Android 6.0.1, 7.1.2). Issue is **only** reproducing on dev.
Video for his issue:
videotogif_2017 05 29_16 36 42

muffinresearch:
Fixes #2471 
Fixes #2472
Fixes #2470

Intl object is not available on FF Android until v54 see https://bugzilla.mozilla.org/show_bug.cgi?id=1215247

### DIFF
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -243,8 +243,16 @@ export function makeI18n(i18nData: I18nConfig, lang: string, _Jed: Jed = Jed) {
   // TODO: move all of this to an I18n class that extends Jed so that we
   // can type-check all the components that rely on the i18n object.
   // Note: the available locales for tests are controlled in tests/setup.js
-  // $FLOW_IGNORE
-  i18n.formatNumber = (number) => new Intl.NumberFormat(lang).format(number);
+  i18n.formatNumber = (number) => {
+    // $FLOW_IGNORE
+    if (typeof Intl === 'object' && Object.prototype.hasOwnProperty.call(Intl, 'NumberFormat')) {
+      return new Intl.NumberFormat(lang).format(number);
+    }
+    // Intl is not yet supported on FF Android though it is expected to land in 54
+    // See https://bugzilla.mozilla.org/show_bug.cgi?id=1215247
+    /* istanbul ignore next */
+    return number.toLocaleString(lang);
+  };
 
   // This adds the correct moment locale for the active locale so we can get
   // localised dates, times, etc.


src/core/i18n/utils.js
```diff
@@ -244,7 +244,15 @@ export function makeI18n(i18nData: I18nConfig, lang: string, _Jed: Jed = Jed) {
   // can type-check all the components that rely on the i18n object.
   // Note: the available locales for tests are controlled in tests/setup.js
   // $FLOW_IGNORE
-  i18n.formatNumber = (number) => new Intl.NumberFormat(lang).format(number);
+  i18n.formatNumber = (number) => {
+    if (Intl && Intl.NumberFormat) {
```
tofumatt: Surprised this won't error in node/on the server as `Intl is undefined` because it wouldn't exist at all? Should this be `typeof Intl !== 'undefined'` just to be extra safe?

kumar303: Yeah, I think `typeof Intl !== 'undefined'` would be better but I don't know if that's just my IE scars talking.

kumar303: Hmm, Flow failed because `Intl` cannot be resolved. I'm not sure how that worked before. It definitely needs a `/* global ... */` definition.

kumar303: Oh, because it was being ignored before

muffinresearch: I think worrying about redefinition of undefined is less of a problem these days, but if everyone feels happier we can go belt and braces. 

For fun here's a comprehensive post on all the in and outs of redefining `undefined` http://adripofjavascript.com/blog/redefining-undefined.html

muffinresearch:
> I agree that this needs a test but it looks good enough to land just to unbreak stuff.

Since the fallback doesn't exist in node, do you think a test is necessary to check the if statement via mocks?

src/core/i18n/utils.js
tofumatt: Yeah, I worry less about redefining `undefined`... but more just about `Intl` being an undefined object in the global scope. Sometimes that makes things error weirdly.

We need to use `typeof Foo !== 'undefined'` elsewhere for `window` so I'd stick to it here. If we wanna change it we can do it app-wide in the future. It's just that fails the server tests, IIRC.

kumar303:
> do you think a test is necessary to check the if statement via mocks?

Yes, definitely, because it sets up an expectation of how our code will operate with the real object. It won't protect against the *real object changing* but it will protect against someone else accidentally regressing our intention which could easily happen as the code evolves through refactoring.

src/core/i18n/utils.js
tofumatt: Yup, causes errors: https://travis-ci.org/mozilla/addons-frontend/jobs/237582984#L540

kumar303:
> Looks like we need to change the check for Intl to use typeof or it fails tests.

It's a Flow error so the easiest fix is to restore the `$FLOW_IGNORE` comment that was suppressing the error before.

src/core/i18n/utils.js
muffinresearch: That's flow because I moved the flow ignore comment.

tofumatt:
Ah okay fair enough.

muffinresearch:
One thing we'll need to also check with this is making sure we have the right locales on the server since the main problem with Intl on node is that node won't necessarily have many locales by default. This is already patched for the tests but there might be a more direct way to do it in docker for production.

That also begs the question since number.toLocaleString doesn't exist in node, how did this ever work for a server render without an error or was it luck that `i18n.format()` was never in a server render?

PR #2484, opened 2017-05-30
Added note about how to debug Flow problems

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2490

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Here are some commands you can run:
 | yarn dev:amo:no-proxy     | Start the dev server without proxy (amo) |
 | yarn dev:disco            | Start the dev server (discovery pane) |
 | yarn flow                 | Run Flow. By default this checks for errors and exits |
+| yarn flow:check           | Explicitly check for Flow errors and exit |
 | yarn flow:dev             | Continuously check for Flow errors |
 | yarn eslint               | Lint the JS |
 | yarn stylelint            | Lint the SCSS |
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Here are some commands you can run:
 | yarn dev:amo              | Start the dev server and proxy (amo) |
 | yarn dev:amo:no-proxy     | Start the dev server without proxy (amo) |
 | yarn dev:disco            | Start the dev server (discovery pane) |
-| yarn flow:check           | Check for Flow errors and exit |
+| yarn flow                 | Run Flow. By default this checks for errors and exits |
 | yarn flow:dev             | Continuously check for Flow errors |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -160,6 +160,9 @@ function getAllAddons({ categoryId }: GetAllAddonsParams = {}) {
   the keyboard. If it's something you think is unfixable then use
   `$FLOW_IGNORE` instead. Please explain your rationale in the comment and link
   to a GitHub issue if possible.
+* If you're stumped on why some Flow annotations aren't working, try using
```
tofumatt: This would require `flow` being installed as a global, can we have these as `npm run`/`yarn` command instead?

kumar303: `flow` doesn't have to be global. It's available in `node_modules/.bin/` just like any other locally installed command. I have this in my `~/.bash_profile` and I highly recommend it:
```shell
# Get the bin for any local NodeJS module on cwd.
PATH="./node_modules/.bin:${PATH}"
```

tofumatt: Right, but when you install anything to `node_modules/.bin/` eg `node_modules/.bin/flow` `npm run flow` will just work without you needing to amend your paths/profile. This is a common paradigm and is friendly to new contributors/Windows users/folks who don't want to add a folder to their `$PATH` that could have anything in it.

I'd much rather us not rely on more globals than we need. I don't have that line in my profile, for instance and I'd rather us not need to add instructions on adding things to paths for new developers when `npm run` already works fine 😄

kumar303: We don't rely on `flow type-at-pos` for anything. These docs are just giving a hint that the command exists and can be helpful. I don't think it makes sense to create `yarn` shortcuts for every single thing you can do with `flow`. There are a lot of things that are useful. For example, I use `flow stop` and `flow start` a lot to deal with the server. The number of `yarn` commands we have is already overwhelming which is itself problematic for contributors.

kumar303: Ooh, I just remembered that unlike with `npm`, you can pass args to `yarn` shortcuts more easily. I can add a `yarn flow` shortcut which will allow things like `yarn flow type-at-pos ...`.

tofumatt: That totally works for me; yeah I don't want to add commands for everything but `yarn flow --stuff` is perfect. Thanks! 😄

kumar303: Keep in mind that `yarn flow --stuff` doesn't work and that's where things get annoying. You have to escape out of yarn with dashes: `yarn flow -- --stuff --other-options`. It works fine for sub-commands though.

tofumatt: Ah yes, sorry, I think I knew that but lazily typed it. Thanks 👍

PR #2485, opened 2017-05-30
Add missing coverage for Intl fallback

muffinresearch:
Moved the Number format instantiation outside the function and added coverage rather than skipping it for the fallback.


muffinresearch:
@kumar303 feel free to tell me where all the flow stuff is wrong, I find it difficult.

muffinresearch:
Ah I see there's this now: https://github.com/mozilla/addons-frontend/pull/2486 I can use that instead.

tests/client/core/i18n/test_utils.js
```diff
@@ -445,9 +445,29 @@ describe('i18n utils', () => {
...
+      const i18n = utils.makeI18n({}, 'de', FakeJed);
+      const fakeNumberFormat = {
+        format: sinon.stub(),
+      };
+      i18n.formatNumber(9518231, { _NumberFormat: fakeNumberFormat });
+      sinon.assert.calledWith(fakeNumberFormat.format, 9518231);
+    });
+
+    it('falls-back to number.toLocaleString if Intl is undefined', () => {
+      const i18n = utils.makeI18n({}, 'de', FakeJed, false);
```
muffinresearch: How do you do that with flow? I'm not finding any docs that explain how to handle things like options objects with defaults in function signatures? 

e.g: 

```js
function foo(bar, { baz = 'something' } = {}){
 // blah
}
```

muffinresearch:
@kumar303 this is ready for another look.

src/core/i18n/utils.js
```diff
@@ -234,22 +234,36 @@ type I18nConfig = {|
   _momentDefineLocale?: Function,
 |};
 
+type makeI18nOptions = {|
+  // FIXME this should be tied back to the Intl type.
+  _Intl?: any,
```
muffinresearch: This was the hack I used to pass the flow checking so any ideas about how to get this to work properly would be much appreciated.

.eslintrc
```diff
@@ -13,6 +13,7 @@
     "jest": true,
```
kumar303: Oh, huh. I must have missed this in the original patch. What about moving all of these test-specific globals to [tests/.eslintrc](https://github.com/mozilla/addons-frontend/blob/master/tests/.eslintrc)?

.eslintrc
```diff
@@ -13,6 +13,7 @@
...
     "sinon": true,
     "webpackIsomorphicTools": true,
+    "Intl": true,
```
kumar303: Since we're only using this in one file, how about `/* global Intl */`?

src/core/i18n/utils.js
```diff
@@ -234,22 +234,36 @@ type I18nConfig = {|
...
+  { _Intl = Intl }: makeI18nOptions = {}
+) {
   const i18n = new _Jed(i18nData);
   i18n.lang = lang;
 
   // TODO: move all of this to an I18n class that extends Jed so that we
   // can type-check all the components that rely on the i18n object.
   // Note: the available locales for tests are controlled in tests/setup.js
+  if (typeof _Intl === 'object' && Object.prototype.hasOwnProperty.call(_Intl, 'NumberFormat')) {
+    i18n.NumberFormat = new _Intl.NumberFormat(lang);
```
kumar303: nit: the convention in our JS is to start class instances with a lower case, like `i18n.numberFormat =`. I think we have some lint rules for this too but they probably don't catch object assignments.

src/core/i18n/utils.js
```diff
@@ -234,22 +234,36 @@ type I18nConfig = {|
...
+  }
+
   i18n.formatNumber = (number) => {
-    if (typeof Intl === 'object' && Object.prototype.hasOwnProperty.call(Intl, 'NumberFormat')) {
-      return new Intl.NumberFormat(lang).format(number);
+    if (typeof i18n.NumberFormat !== 'undefined') {
```
kumar303: since it may not be obvious which formatter is in use, I think it would be nice to log some debug statements here and down below to say which formatter is being used.

tests/client/core/i18n/test_utils.js
```diff
@@ -445,9 +445,34 @@ describe('i18n utils', () => {
...
+      const fakeIntl = {
+        NumberFormat: () => {
+          return {
+            format: formatStub,
+          };
+        },
+      };
+      const i18n = utils.makeI18n({}, 'de', FakeJed, { _Intl: fakeIntl });
+      i18n.formatNumber(9518231);
+      sinon.assert.calledWith(formatStub, 9518231);
```
kumar303: I don't really see how this test is any different from the one above where you just spy on the polyfilled global like `sinon.spy(Intl, 'NumberFormat')`. You could delete this test IMO.

tests/client/core/i18n/test_utils.js
```diff
@@ -445,9 +445,34 @@ describe('i18n utils', () => {
...
+
+    it('falls-back to number.toLocaleString if Intl is undefined', () => {
+      const i18n = utils.makeI18n({}, 'de', FakeJed, { _Intl: false });
+      const numberFormatSpy = sinon.spy(Intl, 'NumberFormat');
+      expect(i18n.formatNumber(1234)).toEqual('1,234');
+      // The calling of toLocalString is inferred by not calling
+      // Intl.formatNumber since there's no way to mock a prototype builtin.
+      sinon.assert.notCalled(numberFormatSpy);
     });
   });
```
kumar303: I think it would be helpful to add another test that simulates the case where `Intl.NumberFormat` is undefined, like  `utils.makeI18n(..., { _Intl: {} })`

.eslintrc
muffinresearch: Sure, I didn't actually know that existed 👍

tests/client/core/i18n/test_utils.js
muffinresearch: Yep good idea.

PR #2486, opened 2017-05-30
Add Flow types for the Intl object

kumar303:
This is a follow-up to https://github.com/mozilla/addons-frontend/pull/2480

I think this would be useful because we won't forget to be protective about accessing the `Intl` object in the future. If you forget to check it, you get:

```
src/core/i18n/utils.js:248
248:     return new Intl.NumberFormat(lang).format(number);
                         ^^^^^^^^^^^^ property `NumberFormat`. Property cannot be accessed on possibly undefined value
248:     return new Intl.NumberFormat(lang).format(number);
                    ^^^^ undefined
```

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -11,6 +11,9 @@
 
 [libs]
 # TODO: this can go away after
+# https://github.com/facebook/flow/issues/2801
+./flow/libs/intl.js.flow
+# TODO: this can go away after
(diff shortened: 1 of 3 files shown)


muffinresearch:
@kumar303 I'm going to ship this so that I can rebase my patch against it.

PR #2488, opened 2017-05-30
docs: Make need for local config extra clear

### LINKED ISSUE #2487: Subresource integrity checks fail when building production locally
Followed https://github.com/mozilla/addons-frontend/#building-and-running-services and used Charles to serve the Add-ons CDN to my local machine, running an HTTP server in `dist/`.

The code builds and the server starts and accepts/returns requests, but fails to load CSS and JS.

I've checked and the files exist. I have unmodified configs and no local configs.

I've successfully done this in the past.

### What happened?
The page is loaded without resources.

The page loads with CSS/JS.

This is blocking #2409 because I need to test the prod version to test it.

Should be noted this happens on master.

tofumatt:
Fixes #2487

I missed this reading the docs then remembered it but I think calling extra attention to it is worthwhile. Maybe we should even have a "build prod" shortcut command that creates this file and does the `build && start` dance.

Now that the tests yell at you for having a `config/local-production-amo.js` it might be nice to be able to easily create one. (My editor ignores files in `.gitignore` so making it is extra annoying, heh)

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,15 +253,15 @@ The env vars are:
 | yarn start  |  Starts the express server (requires env vars)      |
 | yarn build  |  Builds the libs (all apps) (requires env vars)     |
 
-Example: Building and running a production instance of the AMO app:
+**Example:** Building and running a production instance of the AMO app:
 
 ```
 NODE_APP_INSTANCE=amo NODE_ENV=production yarn build
 NODE_APP_INSTANCE=amo NODE_ENV=production yarn start
 ```
 
-To run the app locally in production mode you'll need to create a config file
-that looks something like this. Name it exactly as `config/local-production-amo.js`.
+**Note: To run the app locally in production mode you'll need to create a config file for local production builds.**
+It must be saved as `config/local-production-amo.js` and should look like:
 
 ```js
 import { apiStageHost, amoStageCDN } from './lib/shared';


README.md
```diff
@@ -253,15 +253,17 @@ The env vars are:
...
+**Example:** Building and running a production instance of the AMO app:
 
 ```
 NODE_APP_INSTANCE=amo NODE_ENV=production yarn build
 NODE_APP_INSTANCE=amo NODE_ENV=production yarn start
 ```
 
-To run the app locally in production mode you'll need to create a config file
-that looks something like this. Name it exactly as `config/local-production-amo.js`.
+**Note: To serve assets on production builds locally you must also create a local production config:**
```
kumar303: It's more than just serving assets though. This sentence is pretty much a duplicate of the one below it. Maybe the old sentence should just be marked bold instead?

kumar303:
> Maybe we should even have a "build prod" shortcut command that creates this file and does the build && start dance.

I would rather fix all the tests that rely on a production config so that we can leave `local-production-*` configs lying around.

In all honesty, how often do we need to run the production site locally? Were you trying to debug a production issue?

tofumatt:
Fair enough. Yeah, I was trying to debug the loading bar CSP issues and needed to build the production site to test it.

kumar303:
> I was trying to debug the loading bar CSP issues

Interesting. That's the same exact thing I was trying to debug last time. Other than that, I've never had to build a production site. Having to figure out how to simulate production is not something you want to do when you need to solve a production issue though 🗡 🔥 🐉

PR #2491, opened 2017-05-31
Extract/merge AMO locales

kumar303:


muffinresearch:
@kumar303 could you run the debug locale generation as part of this?

PR #2499, opened 2017-06-01
fix(package): update raven to version 2.0.2

muffinresearch:
Closes #2375

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "normalize.css": "7.0.0",
     "normalizr": "3.2.2",
     "prop-types": "15.5.10",
-    "raven": "1.2.1",
+    "raven": "2.0.2",
     "raven-js": "3.16.1",
     "react": "15.5.4",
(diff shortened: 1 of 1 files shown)


muffinresearch:
See the release details here:

https://github.com/getsentry/raven-node/releases

**Note specifically:** 

 * Node processes with raven will now exit in exactly the same situations as if raven were not present
   * Previously, there were failure scenarios where raven would erroneously cause the process to continue to run when it should have shut down
   * **Be aware when upgrading**: it is possible that with raven-node 2.0, your node process will shut down from exceptions where it previously did not

tofumatt:
I ran this locally, hit pages that throw, browsed around, ran the tests, and things seemed fine with no premature exits from exceptions. I'm gonna say this is okay.

tofumatt:
Looks like #2615 supercedes this though; tested that and it works as well so will merge that.

PR #2501, opened 2017-06-01
Fix flow defs for Intl

### LINKED ISSUE #2500: Fix the flow definition for Intl so that the any escape hatch doesn't need to be used.
See https://github.com/mozilla/addons-frontend/blob/508590c8ae8c8adbdeb75cfdff978c3011564476/src/core/i18n/utils.js#L239 for details.

muffinresearch:
Fixes #2500

After using the flow env to try things out with a reduced test case this seems to work. It appears that you can't use `|` for global definitions or at least removing that makes it work.

See: http://bit.ly/2rI8UbW for the try flow env to see the reduced test case. (Works better in Chrome sadly)

### DIFF
--- a/flow/libs/intl.js.flow
+++ b/flow/libs/intl.js.flow
@@ -1,15 +1,17 @@
-// This is only a partial definition of
-// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
+// This is only a partial definition of Intl (https://mzl.la/1dsIHLk MDN)
 
 declare class Intl$NumberFormat {
   constructor(locales: string | Array<string>, options?: Object): void;
-  format(number: Number): string;
+  format(number: number): string;
 }
 
 declare type IntlType = {
   NumberFormat: Class<Intl$NumberFormat>,
 }
 
-// Mark it as possibly undefined since
-// Safari and some recent versions of Firefox for Android do not implement it.
-declare var Intl: typeof undefined | IntlType;
+
+// Intl can be undefined since Safari and some recent versions of Firefox 
+// for Android do not implement it.
+// The expected declaration here is `declare var Intl: void | IntlType;`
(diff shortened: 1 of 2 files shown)


flow/libs/intl.js.flow
```diff
@@ -10,6 +9,4 @@ declare type IntlType = {
   NumberFormat: Class<Intl$NumberFormat>,
 }
 
-// Mark it as possibly undefined since
-// Safari and some recent versions of Firefox for Android do not implement it.
-declare var Intl: typeof undefined | IntlType;
+declare var Intl: IntlType;
```
kumar303: I think it's fine to add this as a workaround but ideally we want to say that `Intl` *can* be undefined in some browsers. This will make sure that we don't forget to write defensive code around it.

Can you put a comment in here linking to this Flow bug? I think this is what we're hitting. https://github.com/facebook/flow/issues/4077

flow/libs/intl.js.flow
```diff
@@ -1,5 +1,4 @@
-// This is only a partial definition of
-// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
+// This is only a partial definition of Intl (https://mzl.la/1dsIHLk MDN)
 
 declare class Intl$NumberFormat {
   constructor(locales: string | Array<string>, options?: Object): void;
```
kumar303: I just noticed that `format(number: Number)` should be `format(number: number)` -- [the lower case one is what we want](https://flow.org/en/docs/types/primitives/#toc-numbers). I always forget about the difference. Can you add that change? It doesn't help with the bug we're hitting but it will prevent other errors.

flow/libs/intl.js.flow
kumar303: In the comments you may want to point out that the correct definition is:
```
declare var Intl: void | IntlType;
```

PR #2510, opened 2017-06-02
Add an ackrc file

kumar303:


kumar303:
Yeah, this is a magic file that will be recognized by Ack: https://beyondgrep.com/documentation/ack-2.18-man.html#ACKRC-LOCATION-SEMANTICS

It looks like `_ackrc` is also supported which would stop it from being hidden. Would that be better?

tofumatt:
Actually I wish the dotfiles WERE hidden by default on GitHub or something. My issue is the clutter. I think `.ackrc` is fine :-)

Issue #2519, opened 2017-06-05
Consolidate add-on detail views before converting it to saga


kumar303:
The add-on detail components needed a lot of cleanup before porting it to a saga so I think it makes sense to land the cleanup patch first. This is in preparation for https://github.com/mozilla/addons-frontend/issues/2442

For QA: It would be helpful if you could check for regressions on the add-on detail page. Some things to check:
- Can you install add-ons?
- Does theme previewing still work?
- Can you install themes?
- Does the detail page still look right after the theme is installed?

kumar303:
It looks like this caused a regression: https://github.com/mozilla/addons-frontend/issues/2531

kumar303:
Actually, I'm tracking the regression in https://github.com/mozilla/addons-frontend/issues/2531 so let's just close this since it landed in https://github.com/mozilla/addons-frontend/pull/2520

vcarciu:
Tried all scenarios from description and some more scenarios around and everything looks fine.
Marking bug as verified.

kumar303:
Thanks for taking a look. Keep in mind that the fix for https://github.com/mozilla/addons-frontend/issues/2531 hasn't on dev landed yet.

Issue #2494, opened 2017-05-31
Create a test helper for common API dispatch store


tofumatt:
We have been moving tests to use `dispatch` instead of raw state, usually on an ad-hoc basis, which has led to this block of code often being repeated:

```js
import createStore from 'amo/store';
import { setClientApp, setLang } from 'core/actions';

const { store } = createStore();
store.dispatch(setClientApp('android'));
store.dispatch(setLang('fr'));
```

It's good that we're doing this because it's safer and more realistic than raw state (eg `signedInApiState`) but we should abstract this store/state into a test helper because we need it nearly everywhere and it'd be handy to be able to load up a freshly-created store like this instead:

```js
import { testStore } from 'tests/client/helpers'; // or whatever

const store = testStore();
console.log(store.getState().api);
// => would output { clientApp, lang, etc. }

// It could even take *specific* args like: 
const store = testStore({ clientApp, lang });
// Which would dispatch `setClientApp` and `setLang` with custom values,
// rather than using raw state.
```

kumar303:
Yeah, a helper to dispatch sign-in actions would be awesome

kumar303:
heh, I was searching my awesome bar for this and found a similar issue (this is more to convert everything over): https://github.com/mozilla/addons-frontend/issues/2000

kumar303:
This got fixed in https://github.com/mozilla/addons-frontend/pull/2520

PR #2514, opened 2017-06-03
Add current section context to store

tofumatt:
Fixes #2468.
Fixes #2477.

Stores the section of the site (Explore/Extensions/Themes) the user is exploring. Also stores whether or not we're on the home page.

Depends on #2436; this is based off the code in that branch.

### Before
screenshot 2017-05-30 00 39 23

### After
screenshot 2017-06-03 12 51 55
screenshot 2017-06-03 12 51 51
screenshot 2017-06-03 12 51 45

Issue #2437, opened 2017-05-25
Replace sinon calls wrapped with expect to use sinon.assert directly


muffinresearch:
If we wrap things like `stub.calledWith('blah')` with assertions then it hides useful error messages that tell us that the args were wrong etc.

A better solution is to cover to using [sinon.assert](http://legacy.sinonjs.org/docs/#assertions) directly OR use something like [sinon-jasmine](https://github.com/froots/jasmine-sinon). 

I have already tried the latter and it appears to be out of date so it might be easier just to use sinon.asserts for now.

tofumatt:
We've often tended to use `stubFunction.firstCall.args[0]` to get around this which is a less cool API, so being able to use `calledWith()` directly sounds 👍

muffinresearch:
Here's an example of this in action:

<img width="771" alt="node" src="https://cloud.githubusercontent.com/assets/1514/26488989/745faedc-41fd-11e7-91ec-1ae0c14a3899.png">

kumar303:
We have converted most calls to `sinon.assert` but you can see the stragglers here: https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=calledWith&type=

willdurand:
In addition to that, we do not need to use `sinon.assert.called()` when we use `sinon.assert.calledWith()` on the same spy, in the same test case.

PR #2520, opened 2017-06-05
Consolidate add-on detail views before converting it to saga

kumar303:
Closes https://github.com/mozilla/addons-frontend/issues/2519


src/amo/containers/DetailPage.js
```diff
@@ -1,37 +0,0 @@
...
-  }
-}
-
-export function mapStateToProps(state, ownProps) {
-  const { slug } = ownProps.params;
-  const addon = state.addons[slug];
-  const installation = state.installations[addon.guid] || { status: UNKNOWN };
-  return {
-    addon,
-    ...addon,
```
kumar303: As far as I could tell, none of these `addon` props were actually getting used

src/amo/containers/DetailPage.js
```diff
@@ -1,37 +0,0 @@
...
-    ...installation,
```
kumar303: As far as I could tell, only `status` was getting used here. The `isPreviewingTheme` and `themePreviewNode` were getting set by [`withInstallHelpers()`](https://github.com/kumar303/addons-frontend/blob/ef5669d0230e1a40746c14f0a2fc0da1ef765aff/src/core/installAddon.js#L101-L102).

kumar303:
@tofumatt this saga conversion patch is getting hairy so let's try and land just the cleanup bits first before actually switching it to sagas.

tofumatt:
Works for me, doing all the cleanup in one patch and then the move to saga in the next will actually make it way easier to grok what's *really* changing versus what's just kinda shuffling around 👍

src/amo/components/AddonDetail/index.js
```diff
@@ -240,14 +243,24 @@ export class AddonDetailBase extends React.Component {
   }
 }
 
-export function mapStateToProps(state) {
+export function mapStateToProps(state, ownProps) {
+  const { slug } = ownProps.params;
+  const addon = state.addons[slug];
```
tofumatt: When this is saga of course these will have to accept `state.addons[slug]` being empty but I know that's for a future patch 😄

src/amo/containers/App.js
```diff
@@ -227,7 +228,7 @@ export function mapDispatchToProps(dispatch: DispatchFunc) {
       dispatch(logOutUserAction());
     },
     handleGlobalEvent(payload: InstalledAddon) {
-      dispatch({ type: INSTALL_STATE, payload });
+      dispatch(setInstallState(payload));
```
tofumatt: Everywhere we switch to an action function over a plain object is 👍

src/amo/containers/DetailPage.js
tofumatt: I think there's actually an issue open for this (saying we should stop/avoid using `...addon` spread) so if that's true we should at least reference #1416... maybe something in disco is still using it though.

src/amo/containers/DetailPage.js
```diff
@@ -1,37 +0,0 @@
-import React from 'react';
```
tofumatt: Happy to see another container go away. 👍

src/amo/routes.js
```diff
@@ -30,7 +30,7 @@ import ServerError from './components/ErrorPage/ServerError';
 export default (
   <Route path="/:lang/:application" component={App}>
     <IndexRoute component={Home} />
-    <Route path="addon/:slug/" component={DetailPage} />
+    <Route path="addon/:slug/" component={AddonDetail} />
```
tofumatt: I wonder if `AddonDetail` should just be `Addon`? I think that gets to the core of what that page is and it's very central to the site so such a bold name doesn't seem off to me. We don't have a generic `Addon` component to display bits of an `Addon` and I wonder if we ever will... or if we do we can go back to calling this one `AddonDetail`.

Just a thought, I don't care *that* much either way, but given there is no `Addon` or `AddonSnippet` or anything else like that, I wonder why we're specific with this component's name. It could just be `Addon`.

src/core/reducers/installations.js
```diff
@@ -23,12 +23,26 @@ export type InstalledAddon = {
   guid: $PropertyType<AddonType, 'guid'>,
   isPreviewingTheme?: boolean,
   needsRestart?: boolean,
-  status?: $PropertyType<AddonType, 'status'>,
+  // TODO: merge with core/constants.validInstallStates
```
tofumatt: Thanks for adding this note 👍

tests/unit/amo/components/TestAddonDetail.js
```diff
@@ -313,6 +322,30 @@ describe('AddonDetail', () => {
...
+  it('enables a theme preview for non-enabled add-ons', () => {
+    const rootNode = renderAsDOMNode({
+      addon: {
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+      },
+      status: UNKNOWN,
+    });
+    const button = rootNode.querySelector('.AddonDetail-theme-header-label');
+    expect(button).toBeTruthy();
```
tofumatt: This is a bit nit-picky but I've been trying to always at least separate tests into two sections with a newline: one is actions to set up the tests, the other is a list of expectations we're asserting. So this would simply become:

```js
it('enables a theme preview for non-enabled add-ons', () => {
  const rootNode = renderAsDOMNode({
    addon: {
      ...fakeAddon,
      type: ADDON_TYPE_THEME,
    },
    status: UNKNOWN,
  });
  const button = rootNode.querySelector('.AddonDetail-theme-header-label');

  expect(button).toBeTruthy();
});
```

I think that really helps highlight what we're testing and makes the tests less likely to be chunks of code a future developer glazes over. I find that whitespace really helps illustrate intent and importance.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -28,7 +27,7 @@ import I18nProvider from 'core/i18n/Provider';
 import Rating from 'ui/components/Rating';
 import { fakeAddon, fakeReview } from 'tests/unit/amo/helpers';
 import {
-  apiResponsePage, getFakeI18nInst, unexpectedSuccess,
+  apiResponsePage, createFetchAddonResult, getFakeI18nInst, unexpectedSuccess,
```
tofumatt: I prefer each of these methods on their own line but I get this was already this way before. I feel like we've preferred:

```js
import {
  bar,
  foo,
  etc,
} from 'whatever';
```

going forward but we don't have an eslint rule so I won't complain if this lands 😉

tests/unit/amo/helpers.js
```diff
@@ -58,3 +65,20 @@ export const signedInApiState = Object.freeze({
   ...coreSignedInApiState,
   clientApp: 'firefox',
 });
+
+export function dispatchSignInActions({
```
tofumatt: ❤️ ❤️ ❤️ !

This probably won't entirely close #2494 because we'd need to get every test using it but that's awesome.

tests/unit/amo/helpers.js
```diff
@@ -58,3 +65,20 @@ export const signedInApiState = Object.freeze({
...
+  lang = 'en-US',
+  userAgent = sampleUserAgent,
+} = {}) {
+  store.dispatch(setAuthToken(authToken));
+  store.dispatch(setClientApp(clientApp));
+  store.dispatch(setLang(lang));
+  store.dispatch(setUserAgent(userAgent));
+  return {
+    store,
+    state: store.getState(),
```
tofumatt: Nice.

tests/unit/core/TestInstallAddon.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import { renderIntoDocument } from 'react-addons-test-utils';
 
 import createStore from 'amo/store';
+import { setInstallState } from 'core/actions/installations';
```
tofumatt: Just realised this is called `setInstallState` but I think going forward we should use `reducerNameActionVerb` as our template for actions. For instance we do `categoriesFetch`, `categoriesLoad`, `categoriesFail`. I think this helps us identify the type of action much better and while it's a bit awkward from an English-speaker's perspective re: word-order I think it's the best way to make things consistent and easily grokable.

E.g. I think this should be `installStateSet`. Or even `installSet`.

Fwiw I used `currentViewSet` for a similar action in #2514 and it felt natural.

tests/unit/core/TestInstallAddon.js
```diff
@@ -110,10 +110,10 @@ describe('withInstallHelpers inner functions', () => {
         dispatch, { _addonManager: getFakeAddonManagerWrapper(), guid, installURL });
       return setCurrentStatus()
         .then(() => {
-          expect(dispatch.calledWith({
-            type: INSTALL_STATE,
-            payload: { guid, status: ENABLED, url: installURL },
-          })).toBeTruthy();
+          sinon.assert.calledWith(
```
tofumatt: I get that this will throw if it doesn't work but it feels a bit weird to not use our usual test API. Will this give us a better sense of the arg mismatch than `expect()`? If so I am all for it, otherwise it feels a bit weird.

kumar303:
Hmm, the coverage decrease doesn't look like anything I touched.

src/amo/components/AddonDetail/index.js
kumar303: Yeah, that will require some additional tweaks in render

src/amo/routes.js
kumar303: > We don't have a generic Addon component to display bits of an Addon

I think we may use one eventually if we ever get around to DRY'ing up the Disco overlap (I swear we had an issue for this but I can't find it). Honestly, we probably won't get around to it so I don't mind renaming it.

tests/unit/core/TestInstallAddon.js
kumar303: I prefer to use verb phrasing that reads like English. This helps the reader understand what's happening. In response to your point about helping us identify the type of action better, is that really important? *Set install state* tells you what it does, no? I found `categoriesFetch` awkward to read because I had to unpack it backwards but it didn't seem important enough to complain about when reviewing the patch.

tests/unit/core/TestInstallAddon.js
kumar303: Yeah, the failure message makes it easier to solve. For example:

screenshot 2017-06-05 22 22 33


We have an issue on file to convert everything eventually: https://github.com/mozilla/addons-frontend/issues/2437

tofumatt: Sweet, I thought as much but just wanted to make sure 👍

tests/unit/core/TestInstallAddon.js
tofumatt: While I think I still prefer the action name first (I usually care more about which action I'm dealing with more), I see your point and maybe I'm just weird for wanting the verb after.

Okay cool, going forward I'll note to prefer `verb-subject(object)` order.

PR #2522, opened 2017-06-06
Fix Intl error android

### LINKED ISSUE #2521: Intl TypeError on android devices
The root cause of this is due to Intl being undeclared resulting in a ReferenceError

muffinresearch:
Fixes #2521

loki-ooph

### DIFF
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -246,7 +246,9 @@ export function makeI18n(
   i18nData: I18nConfig,
   lang: string,
   _Jed: Jed = Jed,
-  { _Intl = Intl }: makeI18nOptions = {}
+  // Checks required to guard against ReferenceError when Intl is not defined.
+  // $FLOW_FIXME
+  { _Intl = typeof Intl !== 'undefined' ? Intl : undefined }: makeI18nOptions = {}
 ) {
   const i18n = new _Jed(i18nData);
   i18n.lang = lang;


src/core/i18n/utils.js
```diff
@@ -246,7 +246,7 @@ export function makeI18n(
   i18nData: I18nConfig,
   lang: string,
   _Jed: Jed = Jed,
-  { _Intl = Intl }: makeI18nOptions = {}
+  { _Intl = window.Intl }: makeI18nOptions = {}
```
kumar303: This is [used in server code](https://github.com/mozilla/addons-frontend/blob/master/src/core/server/base.js#L32) so it needs to be protected by `if (typeof window !== 'undefined')...` or something like that.

muffinresearch: Ah yep you're right. 🤕

src/core/i18n/utils.js
```diff
@@ -1,4 +1,4 @@
-/* global Intl */
+/* global Intl, window */
```
kumar303: You don't need to declare this as a global anymore

kumar303:
bah, looks like https://github.com/facebook/flow/issues/4077

I suggest adding a [`$FLOW_FIXME`](https://github.com/mozilla/addons-frontend/blob/master/.flowconfig#L28) comment which is different than an ignore comment in that it's something we should be able to search for and fix later.

Issue #2161, opened 2017-03-23
Update to react-router-4


tofumatt:
This will be a fair bit of work but filing here as an issue on our radar, so we can close the Greenkeeper PRs and track it here.

kumar303:
It took me a while to wrap my head around this one but now I've hit a road block. `redux-async-connect` relies on [the old router API](https://github.com/Rezonans/redux-async-connect/blob/master/modules/ReduxAsyncConnect.js#L2) and since it declares `react-router` as a peer dep, it doesn't have its own isolated copy. 

We could fork it and hack `package.json` to declare it as a dep but eyeballing what it does with `ReactRouter` tells me that it still may not work. It looks like we need to replace `redux-async-connect` with `redux-saga` so let's wait on the router upgrade until then.

kumar303:
Here's my patch that was very broken. I doubt we'll need to reference it but maybe.
[react-router-4-upgrade.patch.txt](https://github.com/mozilla/addons-frontend/files/984951/react-router-4-upgrade.patch.txt)

kumar303:
Here is the migration guide: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md

kumar303:
I've been trying to connect components to react-redux-router's state but [this is apparently unsupported](https://github.com/reactjs/react-router-redux/issues/205#issuecomment-173087259) because the state is not always in sync with react-router (wtf). After reading some issues, react-router seems like a big mess.

I've been looking at http://router5.github.io/ which may be a good alternative. The redux support looks more solid.

kumar303:
http://router5.github.io/docs/understanding-router5.html

> [router5] also works very well with state containers like Redux: your state container is placed between your view and your router, and your view subscribes to state updates.

This is exactly what I want in a router ^

As far as I can tell, you can't do this with react-router-redux because react-router doesn't offer enough hooks. E.G. You still have to use route handlers for initial component rendering.

muffinresearch:
@kumar303 if that provides all the nice things like regexes etc and is a better fit for our setup then I'm +1 on us moving in that direction.

willdurand:
There is a new router library that looks interesting: https://reach.tech/router.

willdurand:
For QA: the website should continue to work as expected, there should be no issue with links or page reload.

AlexandraMoga:
[...]

Issue #181, opened 2016-04-05
Dispatch react-router changes with Redux (i.e. track URLs in state)


kumar303:
Consider migrating the router to [react-router-redux](https://github.com/reactjs/react-router-redux). It would provide a couple benefits:
- Since all routing is part of the redux state then time travel will work when debugging (https://github.com/mozilla/addons-frontend/issues/29)
- All components that rely on the URL / query string parameters for their state can simply map to the state rather than rely on other components to pass them the data
- It should make testing simpler since state gets turned into properties
- It should make things more future proof since component context (used by the router) will not be needed anymore (I think). Component context isn't really recommended: https://facebook.github.io/react/docs/context.html
- It would keep all app data flow uni-directional since navigating to a link would just dispatch a redux action like any other event. This should help for sanity when trying to understand how the app works.

kumar303:
react-router 4 may have built in redux support

kumar303:
Also, if SSR is a problem, there is also http://www.electrode.io/docs/redux_router_engine.html to consider

kumar303:
In https://github.com/mozilla/addons-frontend/pull/2523, there is a new reducer that needs to use data from the URL. Switching to redux for routing should make that easier.

PR #2523, opened 2017-06-06
Add current section context to store

### LINKED ISSUE #2468: Store current addon type for addon/context in store

### LINKED ISSUE #2477: Add "Explore", "Extensions", and "Themes" to header/search context

tofumatt:
Fixes #2468.
Fixes #2477.

Stores the section of the site (Explore/Extensions/Themes) the user is exploring. Also stores whether or not we're on the home page.

### Before
screenshot 2017-05-30 00 39 23

### After
screenshot 2017-06-03 12 51 55
screenshot 2017-06-03 12 51 51
screenshot 2017-06-03 12 51 45


src/amo/actions/currentView.js
```diff
@@ -0,0 +1,9 @@
+import { CURRENT_VIEW_SET } from 'core/constants';
+
+
+export function currentViewSet({ addonType, isExploring, isHomePage }) {
```
tofumatt: I'm gonna change this to be the preferred action name type.

src/amo/reducers/currentView.js
```diff
@@ -0,0 +1,38 @@
+/* @flow */
+import { CURRENT_VIEW_SET } from 'core/constants';
+
+
+export type currentViewTypes = {|
+  // TODO: Create an addonType Type that is one of type (ADDON_TYPE_EXTENSION,
+  // ADDON_TYPE_THEME, etc.)
+  addonType: string | null,
+  isExploring: boolean | void,
```
kumar303: Instead of `addonType` and `isExploring`, couldn't you just have a single value that indicates what the context is? I think this would mirror the menu UI a bit more. Example:

```js
type ViewContextState = {|
  context: 'exploring' | 'extensions' | 'themes',
|}
```

You could still support dispatches like `setAddonType(addonType)` and just reduce them as needed, like:

```js
switch (action.type) {
  case 'SET_ADDON_TYPE':
    switch(action.payload.addonType) {
      case 'extension':
        return { ...state, context: 'extensions' };      
      case 'persona':
        return { ...state, context: 'themes' };
    }
}

```

src/amo/reducers/currentView.js
```diff
@@ -0,0 +1,38 @@
...
+  isExploring: false,
+  isHomePage: false,
+};
+
+export default function currentView(
+  state : currentViewTypes = initialState, action : currentViewActionType
+) {
+  const { payload } = action;
+  switch (action.type) {
+    case CURRENT_VIEW_SET:
```
kumar303: To me, `CURRENT_VIEW_CONTEXT` would make more sense because this state describes the context in which one is viewing add-ons. I think you could also drop *current* since that's implied by the persistent nature of the redux state, making it `VIEW_CONTEXT`.

src/amo/reducers/currentView.js
```diff
@@ -0,0 +1,38 @@
+/* @flow */
+import { CURRENT_VIEW_SET } from 'core/constants';
+
+
+export type currentViewTypes = {|
+  // TODO: Create an addonType Type that is one of type (ADDON_TYPE_EXTENSION,
+  // ADDON_TYPE_THEME, etc.)
+  addonType: string | null,
+  isExploring: boolean | void,
+  isHomePage: boolean | void,
```
kumar303: When I look at where `isHomePage: true` is dispatched, it seems to be a duplication of `isExploring` since that's the home page. Can you just say `const isHomePage = isExploring` ? Or is there a future scenario in the desktop mocks that I'm missing?

src/amo/components/Header/index.js
```diff
@@ -68,12 +69,18 @@ export class HeaderBase extends React.Component {
...
         </div>
+        <SectionLinks location={location} />
         <SearchFormComponent pathname="/search/" query={query} />
       </header>
     );
   }
 }
 
+export function mapStateToProps(state) {
+  return { isHomePage: state.currentView.isHomePage };
```
kumar303: It looks like [`amo/components/App/index`](https://github.com/mozilla/addons-frontend/blob/8a0c7baa68f8b0824fe93b6e91914159521c674d/src/amo/components/App/index.js#L196) also needs to be updated to get `isHomePage` from state.

src/amo/reducers/currentView.js
```diff
@@ -0,0 +1,38 @@
...
+  isHomePage: false,
+};
+
+export default function currentView(
+  state : currentViewTypes = initialState, action : currentViewActionType
+) {
+  const { payload } = action;
+  switch (action.type) {
+    case CURRENT_VIEW_SET:
+      return {
```
kumar303: It's good practice to always inherit the previous state even though you don't need to:

```js
return {
  ...state,
  addonType: ...,
  isExploring: ...,
  isHomePage: ...,
}
```

src/amo/components/Addon/index.js
```diff
@@ -69,8 +71,19 @@ export class AddonBase extends React.Component {
     getClientCompatibility: _getClientCompatibility,
   }
 
+  componentWillMount() {
+    const { addon, dispatch } = this.props;
+
+    dispatch(setCurrentView({ addonType: addon.type }));
```
kumar303: I don't think this is the right place to dispatch but it's ok for now. I think we need to move to a saga first. After that we can do something like this in the saga:
```js
export function* fetchAddon() {
  // This is a mythical saga to load an add-on for the detail page.
  const response = yield call(getAddon, { api, addonSlug });
  const addon = yield put(setAddon(response));
  // After the add-on is loaded, we can dispatch its type.
  yield put(setAddonType(addon.type));
}
```

tofumatt:
The state is *often* in the URL but not always. An add-on's detail page is one place where the action is dispatched (so that landing on a theme sets the header and search context to "themes") but the URL makes no mention of "extensions" or "themes"--it's just `/addon/my-theme-slug/`. It's in a few places. For instance:

`/en-US/firefox/extensions/` vs `/en-US/firefox/search?q=foo&type=extension` vs `/en-US/firefox/addon/an-extension-not-a-theme-but-how-would-you-tell/`.

I see the point that using redux-router (which, hey, I'd like to use also!) would cut down on the number of dispatches but ultimately we'll need to do some based on component data and not the route.

src/amo/reducers/currentView.js
tofumatt: Exploring happens when you're searching both extensions and themes but also when you're on the home page; it's basically anytime you are not specifically in extensions or themes. So:

`/` = `isHomePage === true`, `isExploring === true`
`/search?q=foo` = `isHomePage === false`, `isExploring === true`
`/search?q=foo&type=extension` = `isHomePage === false`, `isExploring === false`

src/amo/reducers/currentView.js
tofumatt: Yeah, that's fine actually. Works for me.

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -88,21 +109,47 @@ describe('<SearchForm />', () => {
...
     })).toBeTruthy();
   });
 });
 
 describe('SearchForm mapStateToProps', () => {
   it('passes the api through', () => {
-    const api = { clientApp: 'firefox', lang: 'de', token: 'someauthtoken' };
-    expect(mapStateToProps({ foo: 'bar', api })).toEqual({ api });
+    const { store } = createStore();
+    store.dispatch(actions.setAuthToken(userAuthToken()));
```
tofumatt: Ah right, I did this before that patch. Coolio!

src/amo/reducers/currentView.js
kumar303: ok, thanks for the breakdown by URL. I guess `isHomePage` is necessary but only for the header to know when to use an h1 or not.

src/amo/reducers/currentView.js
tofumatt: I like putting the action verb in the action context to I'd like `VIEW_CONTEXT_SET` but I agree.

tests/unit/amo/containers/TestHome.js
```diff
@@ -6,21 +6,24 @@ import {
...
 import { getFakeI18nInst } from 'tests/unit/helpers';
-import Home from 'amo/containers/Home';
 
 
 describe('Home', () => {
   it('renders a heading', () => {
-    const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
-    const { store } = createStore(initialState);
+    const { store } = createStore();
+    store.dispatch(setClientApp('android'));
```
tofumatt: It works with `dispatchSignInActions` so I'll use that for now if it's okay. Technically it's dispatching more than is needed but that's okay.

tofumatt:
Ready for another review! Cleaned up all your comments and the tests, but as discussed on IRC this is still using the less-elegant dispatch-from-components for now, as we want to wait to integrate `react-redux-router` (#181) until we upgrade `react-router` (#2161).

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -88,21 +113,60 @@ describe('<SearchForm />', () => {
...
   });
 
+  it('passes addonType when set', () => {
+    root = renderIntoDocument(
+      <SearchFormWrapper addonType={ADDON_TYPE_EXTENSION} />
+    ).root;
+    form = root.form;
+    input = root.searchQuery.input;
+
+    expect(!router.push.called).toBeTruthy();
```
kumar303: This would be easier to read as `sinon.assert.notCalled(router.push)`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -88,21 +113,60 @@ describe('<SearchForm />', () => {
...
+    input.value = '& 26 %';
+    Simulate.click(root.submitButton);
+    expect(router.push.calledWith({
```
kumar303: `sinon.assert.calledWith()`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -88,21 +113,60 @@ describe('<SearchForm />', () => {
...
     })).toBeTruthy();
   });
 });
 
 describe('SearchForm mapStateToProps', () => {
   it('passes the api through', () => {
-    const api = { clientApp: 'firefox', lang: 'de', token: 'someauthtoken' };
-    expect(mapStateToProps({ foo: 'bar', api })).toEqual({ api });
+    const { store } = dispatchSignInActions();
+    store.dispatch(setViewContext(VIEW_CONTEXT_HOMEPAGE));
```
kumar303: is it necessary to dispatch the view context for this test?

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -88,21 +113,60 @@ describe('<SearchForm />', () => {
...
+
+    const state = store.getState();
+
+    expect(mapStateToProps(state).api).toEqual(state.api);
+  });
+
+  it('passes the context through', () => {
+    const { store } = dispatchSignInActions();
+    store.dispatch(setViewContext(VIEW_CONTEXT_HOMEPAGE));
```
kumar303: why is this dispatched?

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -88,4 +91,26 @@ describe('<SearchPage />', () => {
...
+
+  it('sets the viewContext to the addonType if addonType exists', () => {
+    const fakeDispatch = sinon.stub();
+    const filters = { addonType: ADDON_TYPE_EXTENSION, query: 'test' };
+
+    render({ count: 0, dispatch: fakeDispatch, filters });
+
+    expect(
+      fakeDispatch.calledWith(setViewContext(ADDON_TYPE_EXTENSION))
+    ).toBe(true);
```
kumar303: `sinon.assert.calledWith()`

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -88,4 +91,26 @@ describe('<SearchPage />', () => {
...
+
+  it('sets the viewContext to exploring if no addonType found', () => {
+    const fakeDispatch = sinon.stub();
+    const filters = { query: 'test' };
+
+    render({ count: 0, dispatch: fakeDispatch, filters });
+
+    expect(
+      fakeDispatch.calledWith(setViewContext(VIEW_CONTEXT_EXPLORE))
+    ).toBe(true);
```
kumar303: `sinon.assert.calledWith()`

tests/unit/amo/containers/TestHome.js
```diff
@@ -6,21 +6,21 @@ import {
...
   it('renders a heading', () => {
-    const initialState = { api: { clientApp: 'android', lang: 'en-GB' } };
-    const { store } = createStore(initialState);
+    const { store } = dispatchSignInActions();
+    const fakeDispatch = sinon.stub();
 
     const root = findRenderedComponentWithType(renderIntoDocument(
       <Provider store={store}>
-        <Home i18n={getFakeI18nInst()} />
+        <HomeBase dispatch={fakeDispatch} i18n={getFakeI18nInst()} />
```
kumar303: This can just be `dispatch={store.dispatch}`. I think it's risky to mix real dispatches (via `dispatchSignInActions()`) with a stub dispatcher. 

I think this was ghost code anyway. The fake dispatcher doesn't appear to be used anymore.

src/amo/reducers/viewContext.js
```diff
@@ -0,0 +1,33 @@
...
+|}
+
+export const initialState = { context: VIEW_CONTEXT_EXPLORE };
+
+export default function viewContext(
+  state : ViewContextType = initialState, action : ViewContextActionType
+) {
+  const { payload } = action;
+  switch (action.type) {
+    case VIEW_CONTEXT_SET:
```
tofumatt: So the actual type I feel is handy to name with the reducer first solely for the purpose of organisation in the constants file. You have all the `CATEGORY_$VERB` actions grouped together naturally that way. I think we should change them all to make action verb first (so `SET_VIEW_CONTEXT` over `VIEW_CONTEXT_SET` and `FETCH_CATEGORIES` over `CATEGORIES_FETCH`) but I think we should do it en-masse and organise the constants so each reducer's actions are someplace.

That said the confusion for `VIEW_CONTEXT_SET` being similar to `VIEW_CONTEXT_HOMEPAGE` is something I noticed myself and since you brought it up too I'll change it in this patch 😄

src/amo/actions/viewContext.js
```diff
@@ -0,0 +1,9 @@
+import { VIEW_CONTEXT_SET } from 'core/constants';
+
+
+export function setViewContext(context) {
```
tofumatt: Oh shoot yes it's not much work and I meant to add it here 👍

src/amo/components/LandingPage.js
```diff
@@ -41,9 +44,33 @@ export class LandingPageBase extends React.Component {
...
+
+  setViewContextType() {
+    const { apiAddonType, dispatch, params } = this.props;
+
+    try {
+      const addonType = apiAddonType(params.visibleAddonType);
+      dispatch(setViewContext(addonType));
+    } catch (err) {
+      if (err instanceof AddonTypeNotFound) {
+        log.info('AddonTypeNotFound in setViewContextType()', err);
```
tofumatt: Because of the way the router works the app can hit this page instead of a 404, thinking getting `addonType` as whatever is in the URL after `/en-US/firefox/` so `/en-US/firefox/foo/` passes addonType `foo`.

#2161 should really help with that...

src/amo/components/SearchForm.js
```diff
@@ -25,10 +31,10 @@ export class SearchFormBase extends React.Component {
   }
 
   goToSearch(query) {
-    const { api, pathname } = this.props;
+    const { addonType, api, pathname } = this.props;
     this.context.router.push({
       pathname: `/${api.lang}/${api.clientApp}${pathname}`,
-      query: { q: query },
+      query: { q: query, type: addonType },
```
tofumatt: Won't this push `?q=searchTerm&viewContext=extension`? That's not what we want.

src/amo/components/SearchForm.js
```diff
@@ -61,8 +77,12 @@ export class SearchFormBase extends React.Component {
   }
 }
 
-export function mapStateToProps({ api }) {
-  return { api };
+export function mapStateToProps(state) {
+  return {
+    addonType: validAddonTypes.includes(state.viewContext.context) ?
```
tofumatt: We often want to know if an addonType is passed to the component, and the keys passed around are import to methods that make up our query strings/API requests, etc.

We often need to know not only what the addonType is but also if there *is* an addonType.

src/amo/reducers/viewContext.js
```diff
@@ -0,0 +1,33 @@
...
+  ADDON_TYPE_THEME,
+  VIEW_CONTEXT_EXPLORE,
+  VIEW_CONTEXT_HOMEPAGE,
+  VIEW_CONTEXT_SET,
+} from 'core/constants';
+
+
+export type ViewContextType = {|
+  context: ADDON_TYPE_EXTENSION | ADDON_TYPE_THEME | VIEW_CONTEXT_EXPLORE |
+    VIEW_CONTEXT_HOMEPAGE,
```
tofumatt: Oddly it doesn't seem to be working though. They're the only context we allow and use but if I change the values or even change that union to "number" flow still says "No errors!". So it doesn't seem to be working.

tests/unit/amo/components/TestCategories.js
```diff
@@ -45,26 +46,56 @@ describe('Categories', () => {
...
+    expect(fakeDispatch.calledWith(setCurrentView({
+      addonType: ADDON_TYPE_EXTENSION,
+    }))).toBeTruthy();
+
+    // Make sure that we update the addonType when `componentDidUpdate()`
+    // is called. This will happen when changing from one route that uses
+    // this component to another–the props will be updated so
+    // `componentDidUpdate()` is called without the component being
+    // mounted again.
+    // TODO: This feels naughty; can it be done better?
```
tofumatt: Yeah I think Enzyme will give us better hooks into component life-cycle when using `mount`. But I'll look at that when we start converting things over.

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -137,26 +164,33 @@ describe('<LandingPage />', () => {
...
+    const rootNode = renderNode({ params: { visibleAddonType: 'XUL' } });
+    expect(rootNode.textContent).toContain('Page not found');
   });
 
   it('throws for any error other than an unknown addonType', () => {
     expect(() => {
       render({
         apiAddonType: () => { throw new Error('Ice cream'); },
-        params: { visibleAddonType: 'doesnotmatter' },
+        params: { visibleAddonType: 'does not matter' },
```
tofumatt: Actually this code is here to protect against 404s that hit this page because of the lack of regexes for routes in the router. Anything at the top level gets sent to this component right now so if it isn't a `validAddonType` we 404. I guess really this value would be a valid addonType but something else would cause the error so I'll change this to `ADDON_TYPE_EXTENSION`. Thanks.

tests/unit/amo/components/TestSearchForm.js
tofumatt: No it's not. Must have added it when I moved to the dispatch helper as other tests used it. Will remove.

tofumatt:
TODO:

- [ ] Fix merge conflicts
- [ ] Add dedicated tests for `SectionLinks` component (only tested indirectly right now and not all paths are fully tested)

tofumatt:
Cheers, thanks for all the review on this. All set for another one.

src/amo/reducers/viewContext.js
```diff
@@ -0,0 +1,32 @@
...
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_THEME,
+  VIEW_CONTEXT_EXPLORE,
+  VIEW_CONTEXT_HOME,
+  SET_VIEW_CONTEXT,
+} from 'core/constants';
+
+
+export type ViewContextType = {|
+  context: ADDON_TYPE_EXTENSION | ADDON_TYPE_THEME | VIEW_CONTEXT_EXPLORE |
```
rpl: @tofumatt if I'm not wrong, flow doesn't allow (at least currently) to use a `const` as a type, which is unfortunate because it is still possible to make flow able to check this kind of union type, but "by using a more verbose strategy", 
e.g. something similar to the following strategy should work:

- explicitly defining the types of these constants: 

```js
/* @flow */
...

// Define and export the types related to the constants.
export type ViewContextExploreType = 'VIEW_CONTEXT_EXPLORE';
export type ViewContextHomeType = 'VIEW_CONTEXT_HOME';

// Define and export the values related to the constants.
export const VIEW_CONTEXT_EXPLORE: ViewContextExploreType = 'VIEW_CONTEXT_EXPLORE';
export const VIEW_CONTEXT_HOME: ViewContextHomeType = 'VIEW_CONTEXT_HOME';
```

- Import the types related to the constants and define the union using these types:

```js
import type {
  ViewContextHomeType,
  ViewContextExploreType,
  ...
} from './const';

export type ViewContextType = {|
  context: ViewContextHomeType | ViewContextExploreType | ...,
|}
```

This way, flow should be able to check that the following code is not valid because the assigned string does not pass the flow types checks:

```js
var thisShouldBeInvalid: ViewContextType = {context: "this string is not of the valid for this type"};
```

tofumatt: Ah, thanks for the info. Strangely even if I changed them to strings it still didn't work?

That syntax is obnoxious 😢

tofumatt: I tried this and flow just takes forever to initialise. :-(

kumar303: > I tried this and flow just takes forever to initialise. :-(

I think this is a bug. I've seen it. You can do `yarn flow start` before `yarn flow:dev` to work around it.

kumar303: https://github.com/mozilla/addons-frontend/issues/2556

src/amo/components/LandingPage.js
kumar303: The site still works if this error gets thrown, right? I think this should throw an error, especially because I don't want us to forget about it when we fix the routing issue.

src/amo/components/SearchForm.js
kumar303: > Won't this push ?q=searchTerm&viewContext=extension? That's not what we want.

Yes, why not? I think `viewContext` makes more sense than `addonType` because it's only used to indicate the view context. It would make the code that parses the query string way more straight forward IMO.

src/amo/components/SearchForm.js
kumar303: But the `addonType` in this property was only introduced in your patch. Can you give me an example in this component where `addonType` is used for something other than the view context? In every place I see it, it's used as the view context so why not simplify the component (and the tests) by calling it what it is:

```js
return {
  viewContext: state.viewContext.context,
}
```

src/core/types/addons.js
```diff
@@ -1,4 +1,9 @@
 /* @flow */
+import { validAddonTypes } from 'core/constants';
+
+// You have to just laugh at this line!
+export type AddonTypeType = validAddonTypes;
```
kumar303: This doesn't look like it needs to be exported anymore. Your patch will be less funny but hey, that's life.

src/amo/components/SearchForm.js
tofumatt: `addonType` is already used by the search code to limit the scope of add-on type returned by the API, so yes it's already used and there is already code and tests written for it. That's why this patch adds so little; we're just using that already supported feature for filtering by `addonType` to also inform the current view context.

We don't have values for `viewContext` that *arent'* addonType, and we would need to go back and change lots of search code to change this. So I don't think this should change to `viewContext`

src/amo/components/SearchForm.js
```diff
@@ -39,7 +45,17 @@ export class SearchFormBase extends React.Component {
   }
 
   render() {
-    const { api, i18n, pathname, query } = this.props;
+    const { addonType, api, i18n, pathname, query } = this.props;
+
+    let placeholderText;
+    if (addonType === ADDON_TYPE_EXTENSION) {
```
tofumatt: Again, we already used `addonType` in other search bits so I'd rather let search things refer to variables used for searching; the current view context is a bit more abstracted from search.

src/amo/reducers/viewContext.js
tofumatt: Cool, that worked. But now even if I change everything to what @rpl suggested it doesn't complain if I say "context can only be a number" or I submit a bad value.

src/amo/components/LandingPage.js
tofumatt: I don't think so. I'll double-check but throwing the error would cause something similar to what happened with #2545; if a component throws and we don't catch it the page will 500.

tofumatt:
I stopped exporting the type as requested but addressed the issue of `viewContext` over `addonType` in the comments. As mentioned: https://addons.mozilla.org/en-GB/firefox/search/?q=test&type=persona is already a search term that works and this relies on the `addonType` key being set in Search's `filters` property. Changing `addonType` to `viewContext` will change that URL to https://addons.mozilla.org/en-GB/firefox/search/?q=test&viewContext=persona which doesn't work (and is also less clear).

src/amo/components/LandingPage.js
tofumatt: The issue with that is this page is hit any time and URL that matches `/*/` is accessed. I suppose in production this shouldn't *really* happen because we serve 404s usually from `addons-server` though any `Link` component pointing to a bad URL would cause this to throw.

We already do exactly what I'm doing here elsewhere in the component and have for awhile, in the interest of not throwing erroneous 500 errors when we're really just hitting a 404. And if there is an invalid `addonType` this will at least log the error.

It should be noted that `apiAddonType` will throw so we'll never `dispatch` with an invalid `addonType`.

If you're 100% sure about throwing here I've added it as a commit but that's a change from what this component already does and I'm doing this here because `apiAddonType` throws when it's handed an invalid `addonType`, which only happens here because of the router.

kumar303: Okay, I understand what's happening now. There is a legit 404 case that we need to handle. Thanks for taking the time to explain it in detail.

Here's what I propose we do. Let me know what you think.

1. revert your last commit so that the error gets caught, logged, and ignored in `setViewContextType()` and the `<NotFound/>` component gets rendered as before.
2. add a comment to the catch/ignore code in `setViewContextType()` explaining that the error is handled more properly in `render()`
3. add a `TODO` comment in `render()` above the `<NotFound/>` code saying that it should be removed after https://github.com/mozilla/addons-frontend/issues/2161

tofumatt:
Cool, thanks for the review and for the idea to comment those try/catch blocks–I remembered why they were there but looking bad it was totally undocumented.

Thanks! All set for another review.

PR #2524, opened 2017-06-06
feat: New Footer

### LINKED ISSUE #2518: Implement the desktop footer
We should change the footer colour over to match the header, and implement the new style from the mocks. (Minus the "view mobile site" link because the site will be responsive. Actually that link should be "Leave AMO beta" on desktop and "View Desktop Site" on mobile sizes...)

screenshot 2017-06-05 17 39 54

tofumatt:
Fix #2518.

Simple style changes to the footer that brings it in line with the new style.

### Desktop

screenshot 2017-06-06 18 59 20
screenshot 2017-06-06 18 58 07

### Mobile

screenshot 2017-06-06 18 59 12
screenshot 2017-06-06 18 58 12


src/amo/components/Footer/index.js
```diff
@@ -0,0 +1,79 @@
...
+                ref={(ref) => { this.desktopLink = ref; }}>
+                {i18n.gettext('View classic desktop site')}
+              </a>
+            </li>
+          </ul>
+        </div>
+
+        <a
+          className="Footer-mozilla-link"
+          href="https://mozilla.org/"
```
tofumatt: I guess we could add lang to this as well; I'll do that.

src/amo/components/Footer/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+  padding: 20px;
+
+  @include respond-to(medium) {
+    display: flex;
+    justify-content: space-between;
+  }
+}
+
+.Footer-language-and-links {
+  max-width: 220px;
```
tofumatt: This is pretty protective against screens that are lower than `220px`... realistically we could probably just make this `width: 220px` as that's what's in the desktop mocks and fits on even phones smaller than `280px` wide. Let me know.

src/amo/components/LanguagePicker.scss
```diff
@@ -9,25 +8,26 @@ $selector-padding: 10px;
...
   border: 0;
   border-radius: $border-radius-default;
-  color: $link-color;
+  color: $white;
   display: block;
   font-size: $font-size-s;
   line-height: 1;
-  padding: $selector-padding $selector-padding #{$selector-padding - 1} #{$selector-padding * 3};
+  max-width: 220px;
+  padding: 5px 15px 3.5px 30px;
```
tofumatt: Could probably use a mixin for this; I'll see if we have one.

src/amo/components/Footer/index.js
```diff
@@ -0,0 +1,79 @@
...
+    const homepageText = i18n.gettext("Go to Mozilla's homepage");
+
+    return (
+      <footer className="Footer">
+        <div className="Footer-language-and-links">
+          <LanguagePicker location={location}
+            ref={(ref) => { this.languagePicker = ref; }} />
+          <ul className="Footer-links">
+            <li>
+              <a href={`https://www.mozilla.org/${lang}/privacy/websites/`}
```
muffinresearch: Seeing the lang made me wonder - does mozilla.org handle it gracefully if we get them a lang they don't cater for?

src/amo/components/Footer/index.js
```diff
@@ -0,0 +1,79 @@
...
+          </ul>
+        </div>
+
+        <a
+          className="Footer-mozilla-link"
+          href={`https://mozilla.org/${lang}/`}
+          title={homepageText}
+        >
+          <Icon className="Footer-mozilla-logo" name="mozilla" />
+          <span className="visually-hidden">{homepageText}</span>
```
muffinresearch: Can the Icon have an alt attribute and then this span would not be needed?

src/amo/components/Footer/index.js
tofumatt: There seem to be certain locales it will redirect to if it doesn't have one available, but they need to be known (so `en-CA` redirects to `en-US`). I'll see about all of them, but given it will 404 if it doesn't redirect, we could just omit the locale to be safe. moz.org should figure out the user's locale like we do so that'll only an annoying experience if you manually set your locale on AMO to be different than the one assigned to you.

src/amo/components/Footer/index.js
tofumatt: I think I'd need to add it but that's a good idea.

src/amo/components/Footer/index.js
muffinresearch: Yeah it's a shame that sometimes it falls apart. I know MDN in the past suffered from problems if the locale was one it didn't know.

tofumatt:
Okay I added the alt text support to `Icon` and removed the lang from the footer URLs as they had the potential to route to 404s. Ready for another r?

tofumatt:
Shoot, asked for a review and forgot to force push because I rebased this onto master. It's all set for r? now

src/ui/components/Icon/index.js
```diff
@@ -4,18 +4,29 @@ import classNames from 'classnames';
...
     }
-    return <i className={classNames('Icon', `Icon-${name}`, className)} {...props} />;
+
+    // If alt text was included, we'll render that in a hidden span.
+    if (alt) {
+      props.children = <span className="visually-hidden">{alt}</span>;
+    }
+
+    return (
+      <i className={classNames('Icon', `Icon-${name}`, className)} {...props} />
```
muffinresearch: Ah I didn't realise this compont use used an `i` element under the covers. This makes the alt attribute a little bit weird but I suppose the alt convention works even if it's not actually an alt attribute on an img.

PR #2529, opened 2017-06-06
chore: Use isAllowedOrigin in AddonReviewList

### LINKED ISSUE #2528: addonReviewList not using isAllowedOrigin
EDIT: Issue updated; I should just change my CDN host but we should fix the `AddonReviewList` component to check origin. I noticed this issue in the first place because it didn't check.

------

Original issue:

If you use `API_HOST=https://addons.mozilla.org yarn dev:amo` as I often do for more robust data for testing, you'll notice all the add-on icons are missing. Actually, a lot of data is missing for any icon that uses `isAllowedOrigin`. I'm guessing we do this so we only even bother rendering images for CDN hosts we allow and otherwise use the fallback icon which is nice, but I'd like an override for development so I can view all the remote content when I'm working with external server data.

Interestingly the `AddonReviewList` component doesn't use it so it shows the icon.

tofumatt:
Fixes #2528.

Just adds this check to the AddonReviewList.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -123,7 +123,9 @@ export function loadAddonIfNeeded(
   return _refreshAddon({ addonSlug: slug, apiState: state.api, dispatch });
 }
 
-export function isAllowedOrigin(urlString, { allowedOrigins = [config.get('amoCDN')] } = {}) {
+export function isAllowedOrigin(urlString, {
+  allowedOrigins = [config.get('amoCDN')],
+} = {}) {
   let parsedURL;
   try {
     parsedURL = url.parse(urlString);
@@ -131,6 +133,7 @@ export function isAllowedOrigin(urlString, { allowedOrigins = [config.get('amoCD
     log.error(`invalid urlString provided to isAllowedOrigin: ${urlString}`);
     return false;
   }
+
   return allowedOrigins.includes(`${parsedURL.protocol}//${parsedURL.host}`);
 }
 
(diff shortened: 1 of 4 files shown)


tofumatt:
Cool, this is ready now 😄

src/core/utils.js
```diff
@@ -123,14 +123,18 @@ export function loadAddonIfNeeded(
   return _refreshAddon({ addonSlug: slug, apiState: state.api, dispatch });
 }
 
-export function isAllowedOrigin(urlString, { allowedOrigins = [config.get('amoCDN')] } = {}) {
+export function isAllowedOrigin(urlString, {
+  allowedOrigins = [config.get('amoCDN')],
+  allOriginsAllowed = config.get('allOriginsAllowed'),
```
muffinresearch: Where is this used in the function?

muffinresearch: Ah I see I think this can be removed now.

tofumatt: Oops, yeah sorry should have removed it. Probably why the tests are failing too (lint). Will send a patch in a sec.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -114,12 +115,23 @@ describe('amo/components/AddonReviewList', () => {
       expect(byLine).toContain(fakeReview.user.name);
     });
 
-    it('renders an icon in the header', () => {
+    it('renders the fallback icon if the origin is not allowed', () => {
```
kumar303: The descriptions of these two tests look like they're swapped, no? This test looks like it expects a real icon URL and the other expects a fallback.

tofumatt: Oops, you are correct. My bad 😞

Issue #2531, opened 2017-06-07
Add-ons installation button is disabled


ValentinaPC:
Steps to reproduce:
1. Load any add-on details page on your Android device i.e. https://addons.allizom.org/en-US/android/addon/phony/?src=hp-dl-featured
2. Observe the installation button

Expected results:
The button is enabled.

Actual results:
The button is disabled.

Notes/Issues:
- issue is reproducing on bot stage and dev
- issue is reproduced for both themes and add-ons
- issue is reproduced on both FF55 and FF52
- pref `extensions.webapi.testing`  is set to true as it was before
- "tap to preview" is also affected
- issue is also reproducing on desktop while in Mobile Site View

Screenshots for this issue:
![2017-06-07_1224](https://user-images.githubusercontent.com/15685960/26876169-845cc4c0-4b8d-11e7-8b7d-88fd26830da4.png)

muffinresearch:
I'm looking into it...

muffinresearch:
It's a regression from https://github.com/mozilla/addons-frontend/commit/0a0e650b25deb74841802e387f93920eda02135d - reverting that commit restores expected functionality.

muffinresearch:
This won't revert cleanly on master but a revert on the tag will work if we need to just get things working as quickly as possible and take more time with the patch for master.

muffinresearch:
@kumar303 also let me know if you need any help with reproducing this locally.

muffinresearch:
Actually I'm going to go ahead and revert in the tag to unblock QA sooner.

muffinresearch:
Reverted for the tag see https://github.com/mozilla/addons-frontend/compare/2017.06.01...2017.06.08-2 

@ValentinaPC this will mean it should be fixed on stage (once deployed shortly) but won't be fixed on -dev. Leaving this bug open to track the fix that needs to land in master.

tofumatt:
Shoot, sorry I didn't notice this myself.

ValentinaPC:
Verified this on **stage** while using FF53(Android 7.1.2).
The install button is enabled (verified also for Nightly)
Video: 
![videotogif_2017 06 07_19 12 26](https://user-images.githubusercontent.com/15685960/26889020-6886112a-4bb5-11e7-862b-839a41d88896.gif)

ValentinaPC:
Depends on #2545

Issue #1416, opened 2016-11-23
Don't spread add-on and installation in mapStateToProps


mstriemer:
`InstallButton` and some add-on components depend on properties of `addon` and `installation` to be set at the root rather than being namespaced. This means the status of an add-on and the status of its installation can collide and cause annoying errors where `status` is expected to be an installation status but is actually an add-on status.

We should return `{ addon, installation }` from mapStateToProps instead of `{ ...addon, ...installation }`.

kumar303:
It looks like discopane's Addon is the only one left: https://github.com/mozilla/addons-frontend/blob/fd669beb91ea311a8d1b8de97cc946a40189f00b/src/disco/components/Addon.js#L227

kumar303:
I plan to fix the ones that were in `amo`'s `AddonDetail` as part of https://github.com/mozilla/addons-frontend/pull/2520

kumar303:
> I plan to fix the ones that were in amo's AddonDetail as part of #2520

Sigh. That didn't work out so well. I added them back in https://github.com/mozilla/addons-frontend/pull/2536 Things get complicated because the reducer also spreads `addon.them_data` onto `addon`.

tofumatt:
We also do a pretty gnarly spread here https://github.com/mozilla/addons-frontend/blob/master/src/core/reducers/addons.js#L32 which is causing #2569.

AlexandraMoga:
Could you please add STR or the label qa not needed?
Thanks

willdurand:
@AlexandraMoga sorry. The add-on detail page should be displayed correctly, and installing/uninstalling an add-on should work on AMO. On Disco, add-ons should also be displayed correctly, and installing/uninstalling an add-on should work as expectedtoo.

AlexandraMoga:
Verified fixed on -dev and stage, with FF62 and Nightly 64, Wn10x64 and Android 8.0
Add-on details page and disco pane look fine. Installing/uninstalling working as expected.

PR #2536, opened 2017-06-07
Fix install button in the context of mozAddonManager

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2531

This was a regression introduced by https://github.com/mozilla/addons-frontend/pull/2520

Maybe one day we can fix the spreads (https://github.com/mozilla/addons-frontend/issues/1416) but I was finding it too difficult to track down all their usage.


src/amo/components/Addon/index.js
```diff
@@ -253,11 +253,22 @@ export class AddonBase extends React.Component {
 export function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
   const addon = state.addons[slug];
-  const installation = state.installations[addon.guid];
+  const installedAddon = state.installations[addon.guid] || {};
 
   return {
     addon,
-    installStatus: installation ? installation.status : UNKNOWN,
+    // These spreads obscure a lot of hidden properties but they
```
tofumatt: It would be good to link to the regression issue where we realised here just so a developer has extra context for this.

tests/unit/amo/components/TestAddon.js
```diff
@@ -557,4 +557,26 @@ describe('mapStateToProps', () => {
 
     expect(installStatus).toEqual(UNKNOWN);
   });
+
+  it('sets random addon object properties', () => {
```
tofumatt: Rather than "sets random addon props" (which makes it seem like maybe *literally* random props are set) I would say "passes all addon object properties".

A link to the issue explaining why would help too.

tests/unit/amo/components/TestAddon.js
```diff
@@ -557,4 +557,26 @@ describe('mapStateToProps', () => {
...
+    signIn();
+    const description = 'whatever';
+    fetchAddon({ addon: { ...fakeAddon, description } });
+    const props = _mapStateToProps();
+
+    // Make sure that top-level addon props get passed along.
+    expect(props.description).toEqual(description);
+  });
+
+  it('sets random install object properties', () => {
```
tofumatt: Same here, I'd avoid "random" and give a bit more context.

src/amo/components/Addon/index.js
kumar303: Well, I'm sure that's just one of many regressions that could occur! I don't know how helpful it would be. There is a lot of theme interaction that would break too.

kumar303: I linked to the issue about removing the spreads which explains the problem and links to the regression

Issue #2537, opened 2017-06-07
Document how to develop for mozAddonManager locally


kumar303:
This is for developers

PR #2538, opened 2017-06-07
Document how to develop for mozAddonManager locally

kumar303:
Closes https://github.com/mozilla/addons-frontend/issues/2537


tofumatt:
billy-d_approves

tofumatt:
(I'll try this tomorrow and give it a proper review then, in case I run into something with the process.)

kumar303:
I was pushing changes to check the Markdown preview :) It should be ready for review now.

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,173 @@
+# Develop features for mozAddonManager
+
+The AMO and Discovery Pane apps use the
+[`mozAddonManager`](https://bugzilla.mozilla.org/show_bug.cgi?id=1310752)
+web API to achieve a more seamless add-on installation user experience.
+However, this API is only available to a limited list of domains.
+If you go to https://addons.mozilla.org (production) in Firefox then `mozAddonManager`
+is available on the page. If you go to a development site then it's not.
+
+## Turning on mozAddonManager for development
```
muffinresearch: These docs don't appear to mention installation. I'm guessing that will implicitly work if you're installing from prod (e.g. you're using prod apis) but if the apis are pointing anywhere else will require additional steps.

muffinresearch:
Really glad there's a better way than having people use charles. 👍

docs/moz-addon-manager.md
kumar303: I'm not totally following. Since the key to accessing the APIs is the host in your URL bar, installation doesn't apply, unless I misunderstood your comment? I assume you meant installing with a production [`API_HOST`](https://github.com/mozilla/addons-frontend#running-amo-for-local-development)?

muffinresearch: So I think the overall question I'm asking is if you follow these steps can you install an addon in local development?

muffinresearch: Also related is the [`xpinstall.signatures.dev-root`](https://wiki.mozilla.org/Add-ons/Projects/DiscoveryImprovements#Testing) pref required to allow install signed add-ons from -dev or stage - and there's other instructions re: the general xpinstall settings to allow a site to have permission to install addons.

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+
+To access `mozAddonManager` on a development site like
+https://addons-dev.allizom.org or https://addons.allizom.org
+go to `about:config` in Firefox and set this property to `true`:
```
tofumatt: It's worth adding a note here that IIRC this will mess up all the existing add-on installs in your browser which really sucks, so use this option with caution.

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+Refresh the page, open the JavaScript console, and you should see a global `mozAddonManager` object.
+
+## Developing on Android
+
+The presence of `mozAddonManager` on Firefox for Android will activate the
+add-on installation switch. Again, you will see this on
+https://addons.mozilla.org (production) but not on a development site.
+
+To activate it on a development site, first
+make sure you set the preference up above. If you're testing this on an Android
```
tofumatt: I would name the preference here. E.g. "set the `webapi.testing` preference as mentioned above".

tofumatt: I'd remove the "this" e.g.

"If you're testing on an Android device..."

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+device then the preference is enough to access `mozAddonManager`.
+
+If you're testing this on desktop Firefox, you need
```
tofumatt: Again I'd remove "this" as it's an unclear pronoun and is easier to read without.

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+It will ask you what version of Firefox you want to build.
+Choose:
+
+```
+Firefox for Desktop
+```
+
+Do not choose the artifact build because unfortunately you have to compile c++ code.
```
tofumatt: Nitpick but it's a capital C in C++ isn't it?

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+
+```
+
+This patch will:
+1. allow you to use an HTTP connection
+2. allow any page at `localhost` (on any port) to access `mozAddonManager`.
+
+With this patch applied, you are ready to build Firefox. Once again, make sure
+you aren't configured for an artifact build since that won't work
+(you'll need to build c++ code). Build it!
```
tofumatt: Again, C++

docs/moz-addon-manager.md
```diff
@@ -0,0 +1,204 @@
...
+### Updating your Firefox source code
+
+You will need to periodically update your Firefox source code and rebuild it
+with your patch.
+
+First, set aside your patch by reverting the changes or using something like the
+[shelve](https://www.mercurial-scm.org/wiki/ShelveExtension) command:
+
+```
+hg shelve
```
tofumatt: Even mercurial's command names are silly. 😂

docs/moz-addon-manager.md
muffinresearch: IIRC isn't it the `xpinstall.signatures.dev-root` pref that does that since that's related to the -dev/stage signing certs which then invalidates the install of everything that was signed with the production cert.

tofumatt: Ahhh okay gotcha. That makes sense. Just worth noting it as it's not obvious why it happens and is definitely a pain.

mstriemer:
I tried reading the mitmproxy docs since it looks like it might be a nicer solution than Charles but my knowledge of how it works is quite limited. I don't totally understand how it works but it seems like it should be able to accomplish the same thing and has an option of configuration from a file.

This could allow shipping the config with addons-frontend which might be nice and could avoid building a local Firefox.

This documentation looks good, setting up Charles is a total pain so it's nice to have a different option. Just thought I'd mention mitmproxy in case someone decides they want to poke around with it.

docs/moz-addon-manager.md
kumar303: > can you install an addon in local development

Yes because you would still be connecting to your local docker container for the devhub

kumar303: > Also related is the xpinstall.signatures.dev-root pref required to allow install signed add-ons from -dev or stage

Oh, fun times! That will be an exercise for a future date. I am assuming the patch to make a custom Firefox allow add-on installations from localhost will not be easy.

docs/moz-addon-manager.md
kumar303: Ha. I suppose it is.

kumar303:
> mitmproxy ... could allow shipping the config with addons-frontend which might be nice and could avoid building a local Firefox.

Indeed! Building c++ especially makes it a pain since you can't do an artifact build.

Issue #2543, opened 2017-06-08
Document the weekly locale extraction process better


kumar303:
The existing docs need to make it a bit more clear about the end to end process

PR #2544, opened 2017-06-08
Document the weekly locale extraction process better

kumar303:
Closes https://github.com/mozilla/addons-frontend/issues/2543

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -103,13 +103,17 @@ files. Do this in your branch and commit:
 NODE_APP_INSTANCE=amo bin/merge-locales
 ```
 
-At this point, you may want to check for
+Keep an eye out for
 [fuzzy strings](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html)
 by running `git diff` and searching for a comment that looks like
 `# fuzzy`. This comment means the localization may not exactly match the source text;
 a localizer needs to review it.
-For applications like `disco` (The Discovery Pane) you may not want
-to merge a pull request until the fuzzy strings are resolved.
+As per our configuration, the application will not display fuzzy translations. These
+strings will fall back to English.
+
+In some rare cases you may wish to remove the
+`fuzzy` marker to prevent falling back to English. Discuss it with a team
+member before removing `fuzzy` markers.
 
 Commit and continue to the next step:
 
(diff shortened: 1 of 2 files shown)


kumar303:
@muffinresearch TBH I wasn't sure what to say in the case where fuzzy strings are in the merge diff

docs/i18n.md
```diff
@@ -27,47 +28,122 @@ NODE_APP_INSTANCE=[MY_APP] NODE_PATH='./:./src' bin/create-locales
...
+NODE_APP_INSTANCE=amo bin/merge-locales
+```
+
+At this point, you may want to check for
+[fuzzy strings](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html)
+by running `git diff` and searching for a comment that looks like
+`# fuzzy`. This comment means the localization may not exactly match the source text;
+a localizer needs to review it.
+For applications like `disco` (The Discovery Pane) you may not want
+to merge a pull request until the fuzzy strings are resolved.
```
muffinresearch: You really have to merge them or localizers won't see them / or fix them.

I think the main points are about checking that the fuzzy markers are expected. E.g. you'd expect to see them when you extract a new string that's a partial match to a key that already exists.

Developers should look out for where code changes cause translations to be marked fuzzy e.g. by changing the copy or changes to the extraction code. In exceptional cases you may prefer to remove fuzzy markers manually to avoid falling back to an english string even if the resulting translation isn't exact any more. 

Removing the fuzzy marker manually will make it less obvious to translators that the translation needs updating so use that option with care and or contact the localizers mailing list to let them know.

An important point is that a fuzzy marked translations will *not* be used in the application by default. This is controlled by config. Allowing fuzzy translations would almost always result in bad translations being used in the application so turning that on should be done with caution.

docs/i18n.md
```diff
@@ -27,47 +28,122 @@ NODE_APP_INSTANCE=[MY_APP] NODE_PATH='./:./src' bin/create-locales
...
 ```
-# build the debug locale files.
-NODE_APP_INSTANCE=[MY_APP] bin/debug-locales
+git push origin amo-locales
 ```
 
+Request a review for the pull request and merge it like any other.
+If necessary, repeat the process for the next application.
+
 ### Building the JS locale files
```
muffinresearch: You could note that this part is now automatically run for deployments and tests.

muffinresearch:
Thanks for doing this.

PR #2551, opened 2017-06-10
Create responsive detail page

### LINKED ISSUE #2430: Create a desktop Add-on details page
Create a desktop version of the add-on detail page according to mocks.

### LINKED ISSUE #2550: "About this add-on" should be on right side of desktop view
The desktop pages put the "About this add-on" on the left side of the screen, but because we expand that context down it renders very strangely with an add-on that has a reasonable amount of text in the description once expanded:

jun-10-2017 17-34-16

I think the issue is to move the "About this add-on" text to the right side of the screen, either above or below the screenshots (prob below).

tofumatt:
This is the first crack at a responsive detail page for extension. Fixes #2430.

Note that I reversed the screenshots and info from the mocks because of #2550. I'd argue this closes #2550 if there aren't any complains from UX.

### Desktop
screen shot 2017-06-10 at 22 02 40
-----
jun-10-2017 21-46-28

### Mobile
screen shot 2017-06-10 at 22 02 34

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -1,95 +1,127 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
 @import "~ui/css/vars";
-@import "~ui/components/Icon/vars";
 
-$page-margin: 20px 10px;
+.Addon {
+  padding: 20px 10px;
+}
 
-// Put the margin on the immediate children so that the header can be full width and blue.
-.Addon > * {
-  margin: $page-margin;
+.Addon-header {
+  display: flex;
+  flex-flow: column;
+  justify-content: space-between;
+}
+
+.Addon-icon-image {
+  height: 48px;
+  width: 48px;
+
+  @include respond-to(medium) {
+    height: 64px;
+    width: 64px;
+  }
+}
+
+.Addon-title-section {
+  display: flex;
+  flex: auto;
+  flex-flow: row wrap;
+  order: 2;
+
+  .InstallButton {
+    @include margin-start(auto);
+  }
 }
 
 .Addon-title {
-  margin-top: 10px;
-}
-
-.Addon-title-heading {
+  color: $black;
   font-size: $font-size-l;
-  margin-bottom: 10px;
+  margin: 10px 0;
+  width: 100%;
   word-wrap: break-word;
+
+  @include respond-to(medium) {
+    font-size: $font-size-xl;
+  }
 }
 
 .Addon-author {
   display: block;
-  opacity: 0.75;
   word-wrap: break-word;
 
+  &,
   a,
   a:link {
-    color: $header-font-color;
-    font-size: $font-size-l;
+    color: $black;
+    font-size: $font-size-m;
+    font-weight: normal;
     text-decoration: none;
+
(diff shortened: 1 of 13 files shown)


src/amo/components/Addon/styles.scss
```diff
@@ -1,95 +1,127 @@
...
 
-.Addon-theme-header-label {
-  $label-margin: 12px;
+// Details section with lots of flexbox stuff, on all but mobile displays.
+@include respond-to(medium) {
+  .Addon-details {
+    @include margin-end(10px);
 
-  @include start($label-margin);
+    display: grid;
```
kumar303: Ooh fancy.

src/amo/components/SearchForm.scss
```diff
@@ -10,6 +10,6 @@
 .SearchInput-input,
 .SearchForm-form {
   color: $base-color;
-  font-size: $font-size-m;
+  font-size: 14px;
```
kumar303: any reason why this can't use a variable?

src/ui/components/Card/index.js
```diff
@@ -12,10 +12,25 @@ export default class Card extends React.Component {
...
     footerText: PropTypes.node,
     header: PropTypes.node,
+    photonStyle: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    // New-style cards that we'll roll out gradually so we don't break
+    // older card components. Photon is the name of the new Firefox
+    // design language.
+    photon: false,
```
kumar303: This should be `photonStyle`, right?

src/ui/components/Card/index.js
```diff
@@ -12,10 +12,25 @@ export default class Card extends React.Component {
     footerLink: PropTypes.node,
     footerText: PropTypes.node,
     header: PropTypes.node,
+    photonStyle: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    // New-style cards that we'll roll out gradually so we don't break
```
kumar303: I'm not sure exactly what this is saying. I think it would help if it were rewritten as a complete sentence.

webpack-common.js
```diff
@@ -144,7 +144,10 @@ export function getPlugins({ excludeOtherAppLocales = true } = {}) {
         options: {
           context: path.resolve(__dirname),
           postcss: [
-            autoprefixer({ browsers: ['last 2 versions'] }),
+            autoprefixer({
+              browsers: ['last 2 versions'],
+              grid: false,
```
kumar303: Cool

src/amo/components/SearchForm.scss
tofumatt: Nope, I didn't have one for a size, but I'll make one.

kumar303: Do we have a pixel width guide for desktop yet that outlines all the sizes we should be using? That was really handy on mobile.

tofumatt: We do. Unfortunately from the mocks that search text is like the one place there isn't a shared style and the one place we use `14px` 😕

I'll make this a new size that's just `$font-size-search` because even though it's a one-off I like that the hard-coded pixel value stuck out to you so I'll keep us using vars. 👍

PR #2554, opened 2017-06-12
fix: Use installStatus, not add-on status for InstallButton (fix #2545)

### LINKED ISSUE #2545: Add-on details page is returning server error
Steps to reproduce:
1. Load details page for an add-on/theme on device i.e. https://addons-dev.allizom.org/en-US/android/addon/mobile-password-manager/?src=hp-dl-featured

Expected results:
Page is loaded with no display or layout issues.

Actual results:
A server error page is loaded.

Notes/Issues:
- may be a regression from #2531 ?
- tapping on add-on name works after second action, first tap returns an error in the console (screenshot from desktop while in Mobile Site View):
first_click

Verified on AMO-dev FF53(Android 7.1.2). Issue is only reproduced on dev.
Video: 
videotogif_2017 06 09_12 42 20

tofumatt:
Fixes #2545

We were relying on weird object spread for this in the past and with the cleanup it got removed. Added a few tests so hopefully we don't get bit by this again.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -174,6 +174,7 @@ export class AddonBase extends React.Component {
       clientApp,
       getClientCompatibility,
       i18n,
+      installStatus,
       userAgentInfo,
     } = this.props;
 
@@ -219,6 +220,8 @@ export class AddonBase extends React.Component {
             {...this.props}
             className="Button--action Button--small"
             disabled={!compatible}
+            ref={(ref) => { this.installButton = ref; }}
+            status={installStatus}
           />
           {!compatible ? (
             <AddonCompatibilityError maxVersion={maxVersion}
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -219,6 +220,8 @@ export class AddonBase extends React.Component {
             {...this.props}
             className="Button--action Button--small"
             disabled={!compatible}
+            ref={(ref) => { this.installButton = ref; }}
```
muffinresearch: Is this used?

tofumatt: Yup, in the tests https://github.com/mozilla/addons-frontend/pull/2554/files#diff-25c04fb93f8e7aaba2269e69a7182c11R328

kumar303:
oh my god, invisible spreads are dead to me!

Thanks for fixing this. Nice approach with the tests -- I'm sad that those weren't there before.

PR #2568, opened 2017-06-14
feat: Add desktop version of theme page (fix #2559)

### LINKED ISSUE #2559: Create a desktop Theme details page
Create the themes page for the addon detail page. Adding this as blocks release because after landing #2551, the theme details page looks a bit ugly.

tofumatt:
Adds support for themes on the desktop site and fixes issues from the last update.

This also improves the HTML and CSS used to create the same page for extensions (#2430).

### Mobile
screen shot 2017-06-14 at 00 52 28
-----------------
screen shot 2017-06-14 at 00 52 32
-----------------
screen shot 2017-06-14 at 00 57 59
-----------------
screen shot 2017-06-14 at 00 58 01

### Desktop
screenshot 2017-06-14 00 58 06
-----------------
screenshot 2017-06-14 00 57 55

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -1,4 +1,5 @@
 /* eslint-disable react/no-danger */
+import classNames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { compose } from 'redux';
@@ -107,7 +108,7 @@ export class AddonBase extends React.Component {
         >
           {installStatus !== ENABLED ? (
             <Button
-              className="Addon-theme-header-label Button--neutral"
+              className="Addon-theme-header-label Button--action"
               disabled={!compatible}
               htmlFor="Addon-theme-header"
             >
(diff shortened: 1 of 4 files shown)


src/amo/components/Addon/index.js
```diff
@@ -181,9 +182,13 @@ export class AddonBase extends React.Component {
     const authorList = addon.authors.map(
       (author) => `<a href="${author.url}">${author.name}</a>`);
     const description = addon.description ? addon.description : addon.summary;
+    // Themes lack a summary so we do the inverse :-/
+    // TODO: We should file an API bug about this...
+    const summary = addon.summary ? addon.summary : addon.description;
+
     const descriptionSanitized = sanitizeHTML(
       nl2br(description), allowedDescriptionTags);
-    const summarySanitized = sanitizeHTML(addon.summary, ['a']);
```
muffinresearch: Are these vars only used once? If so it might be nice to keep the use of the sanitization function closer to the `dangerouslySetInnerHTML` prop usage?

tofumatt: Sure.

muffinresearch: The thought I was having is that it might make it easier to have a future lint rule that all content passed as HTML via dangerouslySetInnerHTML. should be wrapped in a sanitization function.

muffinresearch:
This can be a follow-up but I think there's some breakpoint tuning needed - it could probably stay 100% width longer see:

add-ons_for_firefox

tofumatt:
I'll see about the breakpoints or setting min-widths on those left sections really quickly, good catch.

muffinresearch:
There's some small rtl issues at the similar width:

xo i _ o _suo-pp

Not sure why the RTL chars are broken there (see the bottom)  - maybe those glyphs don't exist in the font?

tofumatt:
I wondered about content max-width but I say let the actual page expand as much as possible... max-width can also hide issues with display or make things hard to account for in my experience. But prose like the details should wrap after around 500px so I'll look into doing that in a follow-up.

Read more like being off on desktop probably makes sense, although when we start adding more boxes underneath it might be useful; add-on descriptions can be wildly long and it will dominate that section of the page.

tofumatt:
Not sure why coveralls is complaining but coverage for those files is still at 100% and I didn't really touch much code even so merging...

PR #2581, opened 2017-06-15
fix: Don't spread theme_data description

### LINKED ISSUE #2233: "None" text listed for theme with no description
Steps to reproduce:

1. Load https://addons.allizom.org/en-US/android/ (or any version of AMO)
2. Search for a theme without a description. Example: search for 'Flag of France'
3. Tap the result to view the theme detail page

Expected results:
No text displayed in the "About this persona" section.

Actual results:
"None" text is displayed.

Notes/Issues:
* If you refresh the page while on the detail view, the "None" goes away

Verified on FF52(Win 7). Issue is reproduced all around AMO-servers.

none

### LINKED ISSUE #2569: Theme without description have layout issues in details page
Steps to reproduce:
1. Load details page on desktop for a theme without a description i.e. https://addons-dev.allizom.org/en-US/firefox/addon/cute-pastel-flowers/

Expected results:

Actual results:
"None" text is affecting the review section.

Notes/Issues:
Verified on AMO-dev FF53(Android 7.1.2).
Screenshot for this issue:
none2

tofumatt:
Hide the "About this extension" when description and summary are blank.

Fixes #2233.
Fixes #2569.

### Desktop
screenshot 2017-06-15 01 20 19

### Mobile
screen shot 2017-06-15 at 01 20 35

### DIFF
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -30,6 +30,14 @@ export default function addon(state = initialState, action) {
         newState[key] = {
           ...thisAddon,
           ...thisAddon.theme_data,
+          // TODO: Remove this when
+          // https://github.com/mozilla/addons-frontend/issues/1416 is fixed.
+          // theme_data will contain `description: 'None'` when the description
+          // is actually `null` and we don't want to set that on the addon
+          // itself so we reset it in case it's been overwritten.
+          //
+          // See also https://github.com/mozilla/addons-server/issues/5650.
+          description: thisAddon.description,
           guid: getGuid(thisAddon),
           type: ADDON_TYPE_THEME,
         };
--- a/tests/unit/core/reducers/test_addons.js
+++ b/tests/unit/core/reducers/test_addons.js
@@ -1,85 +1,108 @@
-import addons from 'core/reducers/addons';
+import { loadEntities } from 'core/actions';
 import { ADDON_TYPE_THEME } from 'core/constants';
+import addons, { denormalizeAddon } from 'core/reducers/addons';
+import { createFetchAddonResult } from 'tests/unit/helpers';
+import { fakeAddon } from 'tests/unit/amo/helpers';
+
 
 describe('addon reducer', () => {
-  let originalState;
+  let addonsState;
 
   beforeEach(() => {
-    originalState = { foo: { slug: 'foo' }, bar: { slug: 'bar' } };
+    const anotherFakeAddon = { ...fakeAddon, slug: 'testing123', id: 64 };
+
+    addonsState = addons({},
+      loadEntities(createFetchAddonResult(fakeAddon).entities));
+    addonsState = addons(addonsState,
+      loadEntities(createFetchAddonResult(anotherFakeAddon).entities));
   });
 
   it('returns the old state', () => {
-    expect(originalState).toBe(addons(originalState, { type: 'BLAH' }));
+    expect(addonsState)
+      .toEqual(addons(addonsState, { type: 'UNRELATED_ACTION' }));
   });
 
   it('stores addons from entities', () => {
-    const state = addons(originalState, {
-      payload: {
-        entities: {
-          addons: {
-            baz: { slug: 'baz' },
-          },
-        },
-      },
+    const anotherFakeAddon = { ...fakeAddon, slug: 'testing1234', id: 6401 };
+
+    const newAddonsState = addons(addonsState,
+      loadEntities(createFetchAddonResult(anotherFakeAddon).entities));
+    expect(newAddonsState).toEqual({
+      ...addonsState,
+      [anotherFakeAddon.slug]: denormalizeAddon(anotherFakeAddon),
     });
-    expect(state).toEqual({ foo: { slug: 'foo' }, bar: { slug: 'bar' }, baz: { slug: 'baz' } });
   });
 
   it('pulls down the install URL from the file', () => {
-    const fileOne = { url: 'https://a.m.o/download.xpi' };
-    const fileTwo = { file: 'data' };
     const addon = {
+      ...fakeAddon,
       slug: 'installable',
       current_version: {
-        files: [fileOne, fileTwo],
+        ...fakeAddon.current_version,
+        files: [{ url: 'https://a.m.o/download.xpi' }, { file: 'data' }],
       },
     };
+
     expect(
-      addons(undefined, { payload: { entities: { addons: { installable: addon } } } })
+      addons(undefined, loadEntities(createFetchAddonResult(addon).entities))
     ).toEqual({
-      installable: {
+      installable: denormalizeAddon({
         ...addon,
         installURL: 'https://a.m.o/download.xpi',
-      },
+      }),
     });
   });
 
   it('sets the icon_url as iconUrl', () => {
     const addon = {
+      ...fakeAddon,
       slug: 'installable',
       icon_url: 'http://foo.com/img.png',
     };
+
     expect(
-      addons(undefined, { payload: { entities: { addons: { installable: addon } } } })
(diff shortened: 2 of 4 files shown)


src/core/reducers/addons.js
```diff
@@ -30,6 +30,12 @@ export default function addon(state = initialState, action) {
         newState[key] = {
           ...thisAddon,
           ...thisAddon.theme_data,
+          // TODO: Remove this when
+          // https://github.com/mozilla/addons-frontend/issues/1416 is fixed.
+          // theme_data will contain `description: 'None'` when the description
```
kumar303: do we have an API bug on file for this yet?

src/core/reducers/addons.js
```diff
@@ -30,6 +30,12 @@ export default function addon(state = initialState, action) {
         newState[key] = {
           ...thisAddon,
           ...thisAddon.theme_data,
```
kumar303: table-flip

src/core/reducers/addons.js
```diff
@@ -30,6 +30,12 @@ export default function addon(state = initialState, action) {
...
+          // TODO: Remove this when
+          // https://github.com/mozilla/addons-frontend/issues/1416 is fixed.
+          // theme_data will contain `description: 'None'` when the description
+          // is actually `null` and we don't want to set that on the addon
+          // itself so we reset it in case it's been overwritten.
+          description: thisAddon.description,
```
kumar303: yep, good call 🌤

tests/unit/core/reducers/test_addons.js
```diff
@@ -82,4 +82,38 @@ describe('addon reducer', () => {
       },
     });
   });
+
+  it('does not use description from theme_data', () => {
+    // See: https://github.com/mozilla/addons-frontend/issues/2569
+    // TODO: Remove this when
```
kumar303: I don't think you need a TODO here. It's enough to put it in the offending code.

tests/unit/core/reducers/test_addons.js
```diff
@@ -82,4 +82,38 @@ describe('addon reducer', () => {
...
+              slug: 'baz',
+              id: 42,
+              theme_data: { theme_thing: 'some-data', description: 'None' },
+              type,
+            },
+          },
+        },
+      },
+    });
+    expect(state).toEqual({
```
kumar303: I think checking the whole state is overkill and will introduce a maintenance burden. Can you try to only assert the `description` property? I think this will do it but make sure it fails without your fix:

```js
expect(state).toEqual(expect.objectContaining({
  baz: { description: null },
})).
```

tests/unit/core/reducers/test_addons.js
```diff
@@ -82,4 +82,38 @@ describe('addon reducer', () => {
       },
     });
   });
+
+  it('does not use description from theme_data', () => {
+    // See: https://github.com/mozilla/addons-frontend/issues/2569
+    // TODO: Remove this when
+    // https://github.com/mozilla/addons-frontend/issues/1416 is fixed.
+    const type = ADDON_TYPE_THEME;
+    const state = addons(originalState, {
```
kumar303: I think it will be easy to convert this test file to use real actions instead of simulated actions. I wouldn't suggest that if I thought it would slow you down significantly, I swear.  😁 

There are only 5 tests to fix! Here's [how to use the `loadEntities` action](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/core/test_utils.js#L423).

src/core/reducers/addons.js
kumar303: Not you fault, I know. I just had to let it out.

tofumatt: Oh I'm right there with you.

src/core/reducers/addons.js
tofumatt: I've filed https://github.com/mozilla/addons-server/issues/5650 but I suspect it might be a requirement because we put that object in the weird `data-browsertheme` or whatever. Still, filed! I'll add it to the comment.

tests/unit/core/reducers/test_addons.js
tofumatt: Agreed, just doing what was already done but I guess I'm refactoring the tests anyway 😉

kumar303: > ...just doing what was already done

Nothing is that easy on `addons-frontend` !

tofumatt: I forgot to do this part, one second.

tofumatt: Done. The actual test just looks like:

```js
expect(state).toMatchObject({
    baz: { theme_thing: 'some-data', },
});
```

tests/unit/core/reducers/test_addons.js
```diff
@@ -1,28 +1,36 @@
...
+    addonsState = addons({},
+      loadEntities(createFetchAddonResult(fakeAddon).entities));
+    addonsState = addons(addonsState,
+      loadEntities(createFetchAddonResult(anotherFakeAddon).entities));
   });
 
   it('returns the old state', () => {
-    expect(originalState).toBe(addons(originalState, { type: 'BLAH' }));
+    expect(addonsState)
+      .toEqual(addons(addonsState, { type: 'UNRELATED_ACTION' }));
```
kumar303: 👍 

I always get confused by unrealistic values -- this is better, thanks :)

tests/unit/core/reducers/test_addons.js
```diff
@@ -34,13 +42,14 @@ describe('addon reducer', () => {
         files: [fileOne, fileTwo],
```
kumar303: It would be nice if `addon` here extended `fakeAddon` just to future proof the properties:
```js
const addon = {
  ...fakeAddon,
  slug: 'installable',
  current_version: {
    ...fakeAddon.current_version,
    files: [fileOne, fileTwo],
  },
};
```

tests/unit/core/reducers/test_addons.js
```diff
@@ -49,37 +58,48 @@ describe('addon reducer', () => {
       slug: 'installable',
```
kumar303: Same thing here. The `addon` object should extend from `fakeAddon`.

tests/unit/core/reducers/test_addons.js
```diff
@@ -49,37 +58,48 @@ describe('addon reducer', () => {
...
-    const type = ADDON_TYPE_THEME;
-    const state = addons(originalState, {
-      payload: {
-        entities: {
-          addons: {
-            baz: { slug: 'baz', id: 42, theme_data: { theme_thing: 'some-data' }, type },
-          },
-        },
-      },
+    const theme = {
```
kumar303: this should extend from `fakeAddon` to future proof the properties (in case new required properties are introduced).

tests/unit/core/reducers/test_addons.js
```diff
@@ -49,37 +58,48 @@ describe('addon reducer', () => {
...
-        guid: '42@personas.mozilla.org',
-        type: ADDON_TYPE_THEME,
-      },
+  });
+
+  it('does not use description from theme_data', () => {
+    // See: https://github.com/mozilla/addons-frontend/issues/2569
+    // Can be removed when
+    // https://github.com/mozilla/addons-frontend/issues/1416 is fixed.
+    const theme = {
```
kumar303: this should extend from `fakeAddon`

tests/unit/core/reducers/test_addons.js
tofumatt: Yeah agreed actually, I was lazy, thanks for calling me out on it ^_^

tofumatt:
Good ideas; added `...fakeAddon` to all the objects.

PR #2583, opened 2017-06-15
fix: yarn test no longer runs lint test

aayushsanghavi:
Fixes: #2489 

Modified `test-once` to include lint tests

PR #2590, opened 2017-06-15
Update README commands

tofumatt:
See #2583.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Here are some commands you can run:
 | yarn test                 | Run all tests (Enters [jest][] in `--watch` mode) |
 | yarn test-coverage        | Run all tests and generate code coverage report (Enters [jest][] in `--watch` mode) |
 | yarn test-coverage-once   | Run all tests, generate code coverage report, then exit |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ Here are some commands you can run:
 | yarn test                 | Run all tests (Enters [jest][] in `--watch` mode) |
 | yarn test-coverage        | Run all tests and generate code coverage report (Enters [jest][] in `--watch` mode) |
 | yarn test-coverage-once   | Run all tests, generate code coverage report, then exit |
-| yarn test-once            | Run all tests with [jest][], then exit |
+| yarn test-once            | Run all tests with [jest][], run all JS + SCSS linters, then exit |
```
kumar303: Now that this command no longer lets us send additional args to `jest`, I think you can simplify the description to:

Run all tests and JS + SCSS linters, then exit

What I mean specifically is that previously you could have done `yarn test-once tests/some/file.js` but that's no longer possible (which is fine, you can use `yarn test` for that).

tofumatt: Cool, will do 👍

PR #2591, opened 2017-06-15
fix: Log in button works on first click

### LINKED ISSUE #1904: "Login to rate this extension" button is working only at second click
Steps to trreproduce:
1.On Android , log in to AMO-dev and be sure that you are logged out
2.Open any add-on and go to review section
3.Press the _Login to rate this extension_ button

Expected results:
Login page is opened

Actual results:
You have to press the _Login to rate this extension_ button twice

NOTES:
Reproducible on many devices , every time

Please see video for the issue(observe that the button was pressed twice and FXA page started to load only second time) : 
secondclick

tofumatt:
Fixes #1904

This is frustrating but because the `<button>` element used for the "Log in to review this extension" button doesn't actually relate to anything the mobile browser will use its hover and focus states after first tap. It's a mess.

The easiest fix is to put a `href="#"` on it and turn it into a link. That isn't great but it's the best we can do for now.

### DIFF
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -66,10 +66,6 @@ export class AuthenticateButtonBase extends React.Component {
     // appearance will transition to the hover state onClick when using a
     // mobile browser. This is the cause of
     // https://github.com/mozilla/addons-frontend/issues/1904
-    //
-    // We tried to use https://css-tricks.com/annoying-mobile-double-tap-link-issue/#article-header-id-2
-    // and media queries to fix this but the `href` solution worked so I'm
-    // going with this.
     return (
       <Button onClick={this.onClick} href="#" {...otherProps}>
         {noIcon ? null : <Icon name="user-dark" />}
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -60,8 +60,18 @@ export class AuthenticateButtonBase extends React.Component {
     const buttonText = isAuthenticated ?
       logOutText || i18n.gettext('Log out') :
       logInText || i18n.gettext('Log in/Sign up');
+
+    // The `href` is required because a <button> element with a :hover effect
+    // and/or focus effect (that is not part of a form) that changes its
+    // appearance will transition to the hover state onClick when using a
+    // mobile browser. This is the cause of
+    // https://github.com/mozilla/addons-frontend/issues/1904
+    //
+    // We tried to use https://css-tricks.com/annoying-mobile-double-tap-link-issue/#article-header-id-2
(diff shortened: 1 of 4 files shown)


tests/unit/amo/helpers.js
```diff
@@ -66,17 +66,30 @@ export const signedInApiState = Object.freeze({
   clientApp: 'firefox',
 });
 
-export function dispatchSignInActions({
+export function dispatchAnonymousActions({
```
kumar303: I think `dispatchClientMetaData` would more accurately express what this is doing. It's simulating how the client will set its lang and user agent.

src/core/css/inc/mixins.scss
```diff
@@ -219,17 +219,8 @@ $default-arrow-margin: 3px;
...
   }
 
-  // This is required because a <button> element with a :hover effect
-  // that changes its appearance will transition to the hover state onClick
-  // when using a mobile browser.
-  // This is the cause of
-  // https://github.com/mozilla/addons-frontend/issues/1904
-  //
-  // For more info, see: https://css-tricks.com/annoying-mobile-double-tap-link-issue/#article-header-id-2
-  @media (pointer: fine) {
```
kumar303: Won't we still need something like this for all other buttons? I liked how it was a generic solution that would apply globally. If it doesn't make sense for this patch, maybe just file an issue.

src/core/components/AuthenticateButton/index.js
```diff
@@ -60,8 +60,18 @@ export class AuthenticateButtonBase extends React.Component {
...
       logOutText || i18n.gettext('Log out') :
       logInText || i18n.gettext('Log in/Sign up');
+
+    // The `href` is required because a <button> element with a :hover effect
+    // and/or focus effect (that is not part of a form) that changes its
+    // appearance will transition to the hover state onClick when using a
+    // mobile browser. This is the cause of
+    // https://github.com/mozilla/addons-frontend/issues/1904
+    //
+    // We tried to use https://css-tricks.com/annoying-mobile-double-tap-link-issue/#article-header-id-2
```
kumar303: I don't think you need to include this last sentence because it's more relevant to the development process and not the code itself. Keeping comments concise always helps since other devs don't always read comments.

tests/unit/amo/helpers.js
tofumatt: wfm 👍

src/core/css/inc/mixins.scss
tofumatt: I kept testing it and it didn't really apply globally. It was very spotty and ultimately that element shouldn't have been an actual `<button>` anyway if it did nothing without JS.

I liked it too but it seemed to work intermittently 😞

tofumatt: Filed #2594

PR #2593, opened 2017-06-15
Fix Disco buttons

### LINKED ISSUE #2592: Missing padding from install button on disco pane in prod.
The gutter has gone from the right-hand side of the install button.

add-ons_manager

Need to check this on RTL too and make sure the fix doesn't impact amo.

tofumatt:
Fixes #2592.

### Screenshots
screenshot 2017-06-15 19 05 53
screenshot 2017-06-15 19 02 25

### DIFF
--- a/src/disco/css/Addon.scss
+++ b/src/disco/css/Addon.scss
@@ -53,38 +53,12 @@ $addon-padding: 20px;
     }
   }
 
-  .Addon-install-button {
-    margin: 0 20px 0 30px;
-
-    [dir=rtl] & {
-      margin: 0 30px 0 20px;
-    }
-  }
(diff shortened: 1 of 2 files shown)


src/disco/css/Addon.scss
```diff
@@ -53,20 +53,7 @@ $addon-padding: 20px;
     }
   }
 
-  .Addon-install-button {
```
muffinresearch: Are these redundant due to markup changes, or is this just controlled elsewhere?

tofumatt: The style is now controlled elsewhere for the most part so these were causing issues.

Issue #2573, opened 2017-06-14
Convert the add-on detail page to using sagas


kumar303:
Instead of using async-connect to pre-load the add-on detail page, we should use a saga. 

Steps to test for regressions:
* Load the add-on detail page, make sure it works. 
* Try refreshing the detail page, make sure it still works.

This is a step towards https://github.com/mozilla/addons-frontend/issues/2356

vcarciu:
Verified around scenarios from description and everything looks fine.

vcarciu:
Yesterday , everything looked ok , but I started to see again the black bars before add-on is loaded on -dev : 
Mobile: 
![blackbarmobile](https://user-images.githubusercontent.com/1583842/27592560-74932946-5b5d-11e7-9a09-a04104632e9a.gif)

Desktop: 
![blackbardesktop](https://user-images.githubusercontent.com/1583842/27592575-7ee5354c-5b5d-11e7-99a7-957388f4146e.gif)

kumar303:
Oh, yeah, the black bars are a new feature we added -- they are expected :) They are placeholders for the real text while it's still loading in the background. Before the black bars, the page would not load until all data was available. With this new strategy, the page loads immediately (with black bars) which feels a lot faster to the user.

vcarciu:
@kumar303 : thanks!

PR #2602, opened 2017-06-16
Convert add-on detail page to use sagas

### LINKED ISSUE #2442: Convert search and add-on detail page to sagas
Instead of using async-connect to pre-load the add-on detail page, we should use a saga. As part of this, let's also stop populating the add-on detail page with data from search results because it will fix bugs like https://github.com/mozilla/addons-frontend/issues/2435

Steps to test this for QA:
* Load the add-on detail page, make sure it works. 
* Try refreshing the detail page, make sure it still works.
* Try searching for an add-on and navigating to its detail page, make sure it works.

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2573

Here is what it looks like before the add-on has loaded:

screenshot 2017-06-20 15 04 51
screenshot 2017-06-20 15 05 24


kumar303:
@tofumatt for some reason the `<LoadingText>` component in the title doesn't show up. I may need some help to figure that one out. Is it Flex weirdness since they render as empty divs?

kumar303:
Whoops, I forgot to handle 404s. This is not ready for a review yet.

kumar303:
@tofumatt 404s are implemented now. It's ready for review.

src/amo/components/Addon/index.js
```diff
@@ -73,9 +80,13 @@ export class AddonBase extends React.Component {
   }
 
   componentWillMount() {
```
tofumatt: I notice there's nothing here for `componentDidUpdate`–if an add-on linked to another add-on the components on the page wouldn't change but the props would update and we would want to do things like update the `viewContext`.

Of course, it would be nice if we did that more often with routing–as in this case we could get away with using the URL to know `addonType` but for now we should `dispatch(setViewContext())` on `componentDidUpdate`.

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
-    const title = i18n.sprintf(
-      // L10n: Example: The Add-On <span>by The Author</span>
-      i18n.gettext('%(addonName)s %(startSpan)sby %(authorList)s%(endSpan)s'), {
-        addonName: addon.name,
-        authorList: authorList.join(', '),
-        startSpan: '<span class="Addon-author">',
-        endSpan: '</span>',
-      });
+    let errorBanner = null;
+    if (errorHandler.hasError()) {
```
tofumatt: I haven't seen this HOC used much but that's really nice in usage. 👍

src/amo/components/Addon/index.js
```diff
@@ -289,9 +361,13 @@ export class AddonBase extends React.Component {
 }
 
 export function mapStateToProps(state, ownProps) {
+  // TODO: make this function handle a falsy addon
```
tofumatt: Seems like this checks for a falsy add-on below–is this TODO still needed?

src/core/sagas/addons.js
```diff
@@ -0,0 +1,40 @@
...
+import { fetchAddon as fetchAddonFromApi } from 'core/api';
+import { FETCH_ADDON } from 'core/constants';
+import { ErrorHandler } from 'core/errorHandler';
+import log from 'core/logger';
+import type { FetchAddonAction } from 'core/actions/addons';
+
+import { getApi } from './utils';
+
+
+export function* fetchAddon(
```
tofumatt: I'm only commenting here to say this code is very easy to follow and is making me happy we're moving to sagas. It makes our async code feel a bit more idiomatic too, compared to the other stuff which was a bit less consistently structured. 👍

tests/unit/amo/components/TestAddon.js
```diff
@@ -1,4 +1,5 @@
 /* global window */
+import { shallow } from 'enzyme';
```
tofumatt: 🎉

tests/unit/amo/components/TestAddon.js
```diff
@@ -96,6 +123,72 @@ describe('Addon', () => {
...
+      dispatch: fakeDispatch,
+      params: { slug: slugParam },
+    });
+
+    // Since there's no add-on, it should be fetched on load.
+    sinon.assert.calledWith(
+      fakeDispatch, fetchAddonAction({ errorHandler, slug: slugParam }));
+
+    // These should be empty:
+    expect(root.find(InstallButton)).toHaveLength(0);
```
tofumatt: Oh cool, I didn't know you could use components; I've always been using selectors. That's awesome.

tests/unit/amo/components/TestAddon.js
```diff
@@ -96,6 +123,72 @@ describe('Addon', () => {
...
+    expect(root.find(AddonCompatibilityError)).toHaveLength(0);
```
tofumatt: Not sure if https://github.com/blainekasten/enzyme-matchers#tobeempty is a better fit for these tests, but it's up to you. I think I've written tests with `expect(root.find('.selector')).toHaveLength(0);` before and I'm fine with it.

tofumatt: I do like that `toHaveLength(0)` and `toHaveLength(1)` contrast well with each other in this test though, so that's a reason to keep using it over the `toBeEmpty()` matcher.

tests/unit/amo/components/TestAddon.js
```diff
@@ -96,6 +123,72 @@ describe('Addon', () => {
...
+      dispatch: sinon.stub(),
+    });
+
+    const root = shallowRender({ errorHandler });
+    expect(root.find(ErrorList)).toHaveLength(1);
+  });
+
+  it('renders 404 page for missing add-on', () => {
+    const id = 'error-handler-id';
+    const store = createStore().store;
```
tofumatt: Could use `const { store } = createStore();` here... and I'd argue we should try to just use `dispatchClientMetadata()` over raw `createStore()` wherever possible because it loads data we can use if we ever have to and is just as easy to use.

But up to you on the usage of `createStore()` vs the dispatch helpers. Obviously these tests are mostly not dispatching things so it doesn't really matter.

tofumatt:
Looks like there are style regressions from this patch, likely to do with the change in the HTML output by the `Addon` component. A theme should look like:

screenshot 2017-06-20 01 04 45

But in your patch looks like:

screenshot 2017-06-20 01 05 51

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
-            />
+    const summaryProps = {};
+    let summaryPlaceholder = null;
+    if (addon) {
+      // Themes lack a summary so we do the inverse :-/
+      // TODO: We should file an API bug about this...
+      const summary = addon.summary ? addon.summary : addon.description;
+      summaryProps.dangerouslySetInnerHTML = sanitizeHTML(summary, ['a']);
+    } else {
+      summaryPlaceholder = <LoadingText />;
```
tofumatt: If someone thinks they're clever and tries to remove this else block like I did locally, and just set the placeholder, we'll get an invariant error because React won't allow `.children` and `.dangerousSetInnerHTML`. I think it might be nice to just remove the placeholder variable and set `summaryProps.children = <LoadingText />` here.

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
-              className="Addon-summary"
-              dangerouslySetInnerHTML={sanitizeHTML(summary, ['a'])}
-            />
+    const summaryProps = {};
+    let summaryPlaceholder = null;
+    if (addon) {
+      // Themes lack a summary so we do the inverse :-/
+      // TODO: We should file an API bug about this...
+      const summary = addon.summary ? addon.summary : addon.description;
+      summaryProps.dangerouslySetInnerHTML = sanitizeHTML(summary, ['a']);
```
tofumatt: I think we should remove the `/* eslint-disable react/no-danger */` at the top of the file and around the `render()` method–instead we should just use `// eslint-disable-next-line react/no-danger`, as it's a good thing to call out and recognise every time we use it 😄

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
+        i18n.gettext('%(addonName)s %(startSpan)sby %(authorList)s%(endSpan)s'), {
+          addonName: addon.name,
+          authorList: authorList.join(', '),
+          startSpan: '<span class="Addon-author">',
+          endSpan: '</span>',
+        }
+      );
+      titleProps.dangerouslySetInnerHTML = sanitizeHTML(title, ['a', 'span']);
+    } else {
+      titlePlaceholder = <LoadingText />;
```
tofumatt: I tried locally and I think this would benefit from a higher `minWidth` which is a property on `<LoadingText />` – it sets the minimum % the text block will fill. As titles aren't usually a few letters something like 70% or even 100% might work better for a lot of these short fields.

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
+              }
+            </section>
+
+            <section className="Addon-metadata">
+              {this.headerImage({ compatible: isCompatible })}
+
+              <h2 className="visually-hidden">
+                {i18n.gettext('Extension Metadata')}
+              </h2>
+              {addon ? <AddonMeta addon={addon} /> : <LoadingText />}
```
tofumatt: I think this would benefit from `AddonMeta` being able to handle an empty `addon` and rendering each section as `<LoadingText />` instead of just one bar, eg:

```js
if (!addon) {
      return (
        <div className="AddonMeta">
          <div className="AddonMeta-item AddonMeta-users">
            <h3 className="visually-hidden"><LoadingText /></h3>
            <p className="AddonMeta-text"><LoadingText /></p>
            <p className="AddonMeta-text AddonMeta-review-count">
              <LoadingText />
            </p>
            <Rating className="AddonMeta-Rating" rating={0} readOnly
              styleName="small" />
          </div>
        </div>
      );
    }
```

Though probably better implemented in the component like you did in `Addon` so we aren't duplicating the HTML.

src/amo/components/Addon/index.js
```diff
@@ -173,114 +185,174 @@ export class AddonBase extends React.Component {
...
-                }
-              />
-            </ShowMoreCard>
-          ) : null}
-
+          {this.renderShowMoreCard()}
           {this.renderRatingsCard()}
 
-          <AddonMoreInfo addon={addon} />
+          {addon ? <AddonMoreInfo addon={addon} /> : null}
```
tofumatt: Same here as for `AddonMeta`; handling an empty `addon` would be good and we should render this component with the LoadingText. That way fewer elements appear out of nowhere once the add-on loads and the page looks less empty whilst loading.

src/amo/components/Addon/index.js
kumar303: actually, we don't even need it because the linter isn't smart enough to follow the `summaryProps` object 😆

src/amo/components/Addon/index.js
kumar303: Whoa, using `.children` makes everything way easier!

src/amo/components/Addon/index.js
tofumatt: Oh no! 😆

src/amo/components/Addon/index.js
kumar303: `AddonMoreInfo` was pretty tough to convert. I don't think it's worth the effort right now.

kumar303:
@tofumatt I fixed the style regressions, thanks for catching that. It should be ready for another look now.

src/amo/components/Addon/index.js
```diff
@@ -13,21 +13,24 @@ import DefaultRatingManager from 'amo/components/RatingManager';
 import ScreenShots from 'amo/components/ScreenShots';
 import Link from 'amo/components/Link';
 import fallbackIcon from 'amo/img/icons/default-64.png';
+import { fetchAddon } from 'core/actions/addons';
+import { withErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
-import { ADDON_TYPE_THEME, ENABLED, UNKNOWN } from 'core/constants';
+import {
+  ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME, ENABLED, UNKNOWN,
```
kumar303: > Nit but once we start doing more imports than fit on a line it's one-per-line

it's pretty easy to fix once they can't fit on a line anymore

src/amo/components/Addon/index.js
kumar303: This is fixed now

tofumatt: Thanks; sorry it was an awkward place to leave a comment as I know it won't be cleared. 😅

tests/unit/amo/components/TestAddon.js
```diff
@@ -91,11 +118,123 @@ describe('Addon', () => {
...
+    root.componentWillReceiveProps({ addon: fakeAddon });
+    // The view context should not be dispatched.
+    sinon.assert.notCalled(fakeDispatch);
+  });
+
+  it('does not update view context unless there is an addon', () => {
+    const fakeDispatch = sinon.stub();
+    const root = render({ dispatch: fakeDispatch });
+    fakeDispatch.reset();
+    root.componentWillReceiveProps({});
```
tofumatt: I think this can use `root.setProps({})` as well.

tests/unit/ui/components/TestLoadingText.js
```diff
@@ -0,0 +1,24 @@
+import { shallow } from 'enzyme';
```
tofumatt: Damn, did this not have any tests before? I'm sorry 😳

tests/unit/ui/components/TestLoadingText.js
```diff
@@ -0,0 +1,24 @@
...
+  const render = (props = {}) => shallow(<LoadingText {...props} />);
+
+  it('renders LoadingText element with className', () => {
+    const root = render();
+    expect(root).toHaveClassName('LoadingText');
+  });
+
+  it('lets you set a fixed width', () => {
+    const root = render({ width: 55 });
+    expect(root.prop('style')).toMatchObject({ width: '55%' });
```
tofumatt: This can be `expect(root).toHaveProp('style', { width: '55%' });` because I don't think we expect the component to be setting style props other than width.

tofumatt: But if you want to leave it as is I suppose it's more extensible. Up to you.

src/amo/components/Addon/index.js
```diff
@@ -73,9 +80,21 @@ export class AddonBase extends React.Component {
...
+    if (addon) {
+      dispatch(setViewContext(addon.type));
+    } else {
+      dispatch(fetchAddon({ slug: params.slug, errorHandler }));
+    }
+  }
 
-    dispatch(setViewContext(addon.type));
+  componentWillReceiveProps({ addon: newAddon }) {
+    const { addon: oldAddon, dispatch } = this.props;
```
kumar303: I don't think it's the greatest syntax. It's actually a shortcut for this which is different from the code you posted:

```js
const oldAddon = this.props.addon;
const dispatch = this.props.dispatch;
```

tofumatt: Ohh, wow. That's not what I expect, making it a bit hard to read. Maybe it would be nicer to do:

```js
const { dispatch } = this.props;
const addon = this.props.oldAddon;
```

I just find that syntax opposite to what I expect.

src/amo/components/Addon/index.js
```diff
@@ -173,69 +193,127 @@ export class AddonBase extends React.Component {
         header={i18n.gettext('Rate your experience')}
         className="Addon-overall-rating"
         {...props}>
-        <RatingManager
-          addon={addon}
-          location={location}
-          version={addon.current_version}
-        />
+        {addon ?
+          <RatingManager
```
kumar303: Actually, yeah, I was thinking about this too. The 'No reviews yet' is an answer to a question asked, which is 'how many reviews are there for this add-on?' In this case, it is not the right answer to give. I'll see if I can change it without much churn.

tofumatt: I commented below–it looks like it *should* be straightforward.

But if it becomes a beast I'm happy to wait for another patch.

src/amo/components/Addon/index.js
kumar303: I don't think we should add workarounds for standard JavaScript syntax. It is what it is and [it's well documented](). It will be around for a long time :) We can help teach contributors if they get stuck with it.

kumar303: Plus, your workaround suggestion adds an extra line of code -- less code is always easier to read IMO.

tofumatt: Fair enough!

I don't think it's easier to read, but it is a standard syntax so at least someone can look it up. I *think* there are language features I think we don't use because they aren't pleasant and I'd nominate this one as well... I think it's easier to reason about what this is doing with the two lines of code, and it's not logic just assignment so I think it doesn't add complexity.

src/amo/components/Addon/index.js
kumar303: The 'No reviews yet' is fixed.

As for `AddonMoreInfo`, I don't think it's worth spending time on but, sure, I filed and marked it as `contrib: welcome` https://github.com/mozilla/addons-frontend/issues/2629

src/amo/components/AddonMeta.js
```diff
@@ -40,10 +46,14 @@ export class AddonMetaBase extends React.Component {
...
           <h3 className="visually-hidden">{i18n.gettext('Used by')}</h3>
-          <p className="AddonMeta-text">{userCount}</p>
-          <p className="AddonMeta-text AddonMeta-review-count">{reviewCount}</p>
-          <Rating className="AddonMeta-Rating" rating={averageRating} readOnly
-            styleName="small" />
+          <p className="AddonMeta-text AddonMeta-user-count">
+            {userCount}
+          </p>
+          <p className="AddonMeta-text AddonMeta-review-count">
+            {reviewCount}
```
kumar303: Done

src/core/actions/addons.js
```diff
@@ -0,0 +1,29 @@
...
+export type FetchAddonAction = {|
+  type: string,
+  payload: {|
+    errorHandlerId: string,
+    slug: string,
+  |},
+|};
+
+export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddonAction {
+  if (!errorHandler) {
```
kumar303: Flow does protect against this but, yeah, not all components opt into Flow. Until then we'll need checks like this.

tests/unit/amo/components/TestAddon.js
```diff
@@ -91,11 +118,123 @@ describe('Addon', () => {
     sinon.assert.calledWith(fakeDispatch, setViewContext(fakeAddon.type));
   });
 
+  it('dispatches setViewContext on update', () => {
+    const fakeDispatch = sinon.stub();
+    const root = render({ dispatch: fakeDispatch });
+    fakeDispatch.reset();
+    root.componentWillReceiveProps({
```
kumar303: oops, I forgot to convert the rest of this suite to use Enzyme. Heading down the rabbit hole, brb! 😓

tests/unit/amo/components/TestAddon.js
```diff
@@ -91,11 +118,123 @@ describe('Addon', () => {
...
+    root.componentWillReceiveProps({ addon: fakeAddon });
+
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeAddon.type));
+  });
+
+  it('only dispatches setViewContext for a new addon type', () => {
+    const fakeDispatch = sinon.stub();
+    const root = render({ addon: fakeAddon, dispatch: fakeDispatch });
+    fakeDispatch.reset();
+    // Update with the same addon (this apparently happens IRL).
```
kumar303: Surprisingly, it does update with the *same* add-on for some reason. I discovered this because the page got stuck in an infinite loop and crashed Firefox.

tests/unit/ui/components/TestLoadingText.js
kumar303: Oh, it did. This test file was just never renamed after the component changed names (all I did here was rename it).

tests/unit/ui/components/TestLoadingText.js
kumar303: I think it helps for maintenance to keep assertions isolated to exactly what they are testing. In this case, the test is *only* checking for the width style rule. It should be flexible enough to allow new style rules in the future without the test breaking.

tests/unit/amo/components/TestAddon.js
tofumatt: Ah, weird. I think `componentDidUpdate` is only called if the props changed whereas `componentWillUpdate` is called either way maybe? My bad if I suggested the wrong one. We should double-check we're using the right one according to the lifecycle docs: https://facebook.github.io/react/docs/react-component.html#the-component-lifecycle

tests/unit/ui/components/TestLoadingText.js
tofumatt: Oh, cool! Thanks!

tests/unit/amo/components/TestAddon.js
tofumatt: Cool, let me know if you think it warrants another review; should be easy to do and quick!

tests/unit/ui/components/TestLoadingText.js
tofumatt: Cool. I do really like `toMatchObject` for that 👍

src/amo/components/Addon/index.js
kumar303: Once I got used to it it no longer seemed foreign. I really wish it looked more like this but oh well:

```js
const { addon as oldAddon } = this.props;
```
^ This is how I read it in my head now which helps me parse it.

tofumatt: omg, agreed. Why didn't they just do that?!

That actually helps though, thanks 👍

kumar303:
Thanks for all the reviews.

Issue #2598, opened 2017-06-16
Theme icon from "All written reviews" page is too small


ValentinaPC:
Steps to reproduce:
1. Load "All written reviews" page for a theme on device i.e. https://addons-dev.allizom.org/en-US/android/addon/one-piece-luffy-and-sun-256856/reviews/
2. Observe theme icon

Expected results:
Theme icon is aligned.

Actual results:
Theme icon is smaller than extensions icons from the same page and is not aligned with theme title and author.

Notes/Issues:
- reproducing on moth mobile pages and new desktop pages.
Verified on FF53(Android 7.1.2, Win 7). Issue is reproducing on all AMO servers.
Screenshots for this issue: 
![theme](https://user-images.githubusercontent.com/15685960/27221739-5e343e66-5292-11e7-89bd-173abcce9267.png)
![theme_icon](https://user-images.githubusercontent.com/15685960/27221747-6215bd02-5292-11e7-949d-9bda12866cb8.png)

aayushsanghavi:
@ValentinaPC should the review list header look like this -

![theme_icon](https://user-images.githubusercontent.com/19590302/27231408-0c3d9520-52a2-11e7-8b63-f20ae0753939.png)

The image has to have a height and width equal to the height of the text divs?

ValentinaPC:
Let's ask @tofumatt .
Matt, what do you think about this?

tofumatt:
Hi there, sorry my late reply.

Yes it should look like that, at least aligned like that... the icon is `32px x 32px` so it shouldn't be stretched anymore than that. `32px` icons already look blurry on modern devices with retina and better screens.

aayushsanghavi:
@tofumatt should we resize all theme header icons to 32x32, for uniformity? What do you suggest?

tofumatt:
Yes. We shouldn't be stretching them beyond the file size; preferably we should have higher resolution files but that's a whole thing...

Maybe we do and I'm just seeing bad add-ons. Can you find add-one whose icons are above 32px?

- Matthew Riley MacPherson (sent from mobile)

muffinresearch:
See also https://github.com/mozilla/addons-server/issues/5805

vcarciu:
Verified as fixed : 
![screenshot_20170707-112106 1](https://user-images.githubusercontent.com/1583842/27950003-60e9e24c-6308-11e7-9bdb-5076e51125b2.png)

PR #2609, opened 2017-06-20
fix: Theme icon from "All written reviews" page is too small

aayushsanghavi:
Fixes: #2598 

Resized theme icon to make it bigger

After - 
screenshot from 2017-06-20 07-56-15

### DIFF
--- a/src/amo/css/AddonReviewList.scss
+++ b/src/amo/css/AddonReviewList.scss
@@ -9,17 +9,18 @@
 }
 
 .AddonReviewList-header {
-  margin: 15px 0;
+  margin: 15px 0 25px;
   display: flex;
 
   .AddonReviewList-header-text {
+    padding: 5px 0;
+
     > h2,
     > h3 {
       font-size: 1.1em;
     }
 
     > h2 {
-      padding-top: 10px;
       margin: 0 0 5px;
     }
 
@@ -40,6 +41,13 @@
 
   .AddonReviewList-header-icon {
     margin: 0 20px;
+    width: 54px;
+    height: 54px;
+  }
+
+  .AddonReviewList-header-icon img {
+    width: 54px;
+    height: auto;
   }
 }
 
--- a/src/amo/css/AddonReviewList.scss
+++ b/src/amo/css/AddonReviewList.scss
@@ -10,12 +10,7 @@
 
 .AddonReviewList-header {
   margin: 15px 0;
-  display: -ms-flexbox;
-  display: -webkit-flex;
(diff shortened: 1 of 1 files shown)


src/amo/css/AddonReviewList.scss
```diff
@@ -40,6 +41,13 @@
 
   .AddonReviewList-header-icon {
     margin: 0 20px;
+    width: 54px;
+    height: 54px;
+  }
+
+  .AddonReviewList-header-icon img {
+    width: 54px;
```
tsl143: Heads up!
This results in stretching of image to 54x54 which is actually 32x32.

aayushsanghavi: Yes it does. It even shrinks images that are 64x64 to 54x54 but there is no alternative besides resizing the images to the same dimensions because we deal with images of different sizes for almost every theme/extension.
I thought this approach would be helpful because it provides a standard icon size throughout the application, fixed sized margins and paddings which otherwise, we would have to dynamically change in the JavaScript code for vertically centered alignment. Also, the height of the text div in the header section is 54px so it was a natural choice.
The only real issue would be that small images might look a little blurry when resized to a bigger dimension.

aayushsanghavi:
@tofumatt I'm not resizing the theme header icons now. I've centered them with the header text to give the alignment we need. Take a look -

screenshot from 2017-06-22 13-04-38

Also, if you can please take a look at the PR too. Thanks!

src/amo/css/AddonReviewList.scss
```diff
@@ -10,7 +10,13 @@
 
 .AddonReviewList-header {
   margin: 15px 0;
+  display: -ms-flexbox;
+  display: -webkit-flex;
   display: flex;
+  -ms-flex-align: center;
```
muffinresearch: You can leave vendor prefixing out, postCSS should be doing that automatically.

aayushsanghavi: So just having 
```
display: flex;
align-items: center;
```
should suffice right?

tofumatt: Yup, that's all you need.

aayushsanghavi: Thanks! And what do you think about the [image above](https://github.com/mozilla/addons-frontend/pull/2609#issuecomment-310301144)? That's the alignment we need for the icons right?

aayushsanghavi:
@tofumatt please take a look at this one when you're free 😄

Issue #2355, opened 2017-05-08
Expose X_IS_MOBILE_AGENTS header in the app


muffinresearch:
A header `X_IS_MOBILE_AGENTS` is sent to the mobile app via nginx [1]. This can be used to do different things based on the user's preferences or the UA causing them to get the mobile site.

This should be exposed to the app in a generic way so that decisions can be made in components to tune the UX to suit mobile or desktop if not present.

This should be set both in the case of the mamo cookie being set *and* if the mamo cookie is not present and it's a mobile UA.

The first step would be to set something up without relying on it so that it can be tested. We'll also need a good process for emulating this in development too.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1361323

kumar303:
> We'll also need a good process for emulating this in development too.

This part worries me a bit. What do we need to use `X_IS_MOBILE_AGENTS` for that we can't already do?

muffinresearch:
@kumar303 it combines the funtionality of the cookie and or UA string parsing in one place.

kumar303:
oh, ok, so we can remove some custom cookie parsing that the frontend app already does? That sounds good.

muffinresearch:
@kumar303 right now we're not really doing anything with the cookie since this is just a mobile site at the moment. However, if we want to do different things for desktop or mobile (e.g. change amount of data fetched, or withhold functionality for mobile) this provides a good solution.

muffinresearch:
We've not needed this yet so I'm thinking we should just close this out until such point that we do.

PR #2618, opened 2017-06-21
Create a desktop search page

### LINKED ISSUE #2560: Implement a desktop search page
Implement a desktop version of the search page according to the mocks.

tofumatt:
Fixes #2560.

This one was a bit intense! I cleaned up the folder structure and ended up refactoring/cleaning up the `SearchResult` CSS. Pretty much everything here is a style change except the new `SearchContextCard` at the top of the page which shows result counts, etc.

We had, in my opinion, more than enough space for add-on descriptions in the medium/"tablet" view so I added them. It felt okay on a tablet, cue "developers hate whitespace" jokes.

Here is a bunch of screenshots:

### Small
screen shot 2017-06-21 at 13 34 14
screen shot 2017-06-21 at 13 34 12

### Medium
screen shot 2017-06-21 at 13 34 30
screen shot 2017-06-21 at 13 34 27

### Large/Desktop
screenshot 2017-06-21 13 34 03
screenshot 2017-06-21 13 34 00


src/amo/components/AddonsCard/styles.scss
```diff
@@ -0,0 +1,16 @@
+@import "~amo/css/inc/vars";
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+@include respond-to(large) {
+  ul.AddonsCard-list {
```
muffinresearch: Is it possible to factor out the `ul`?

src/amo/components/AddonsCard/styles.scss
```diff
@@ -0,0 +1,18 @@
+@import "~amo/css/inc/vars";
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+ul.AddonsCard-list {
```
muffinresearch: Is it possible to factor out the `ul`?

src/amo/components/Search/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  .Card {
+    margin: $padding-page;
+
+    @include respond-to(large) {
+      margin: 0;
+    }
+  }
+
+  .SearchSort {
```
muffinresearch: Do these need to be nested?

src/amo/components/Search/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+.Search {
+  @include respond-to(large) {
+    display: grid;
+    grid-auto-flow: column dense;
+    grid-gap: 10px;
+    grid-template-columns: minmax(300px, 35%) auto;
+    padding: $padding-page;
+  }
+
+  .Card {
```
muffinresearch: Is this an override of something generic?

src/amo/components/SearchSort/SearchSort.scss
```diff
@@ -65,6 +73,10 @@
     background: transparentize($link-color, 0.8);
     box-shadow: none;
   }
+
+  @include respond-to(large) {
```
muffinresearch: Any time we're using `display: none` we should also consider the alternative of with-holding the content. 

Of changing content output based on UA or the mamo cookie is not equal to a viewport size based decision. See also https://github.com/mozilla/addons-frontend/issues/2355 (this is probably something for later rather than now).

src/ui/components/CardList/styles.scss
```diff
@@ -11,7 +14,7 @@
     padding: 0;
 
     > li {
```
muffinresearch: Side note: Would be nice to swap things like this out for distinct classes when the opportunity presents itself.

src/amo/components/AddonsCard/styles.scss
tofumatt: I can have a look!

src/amo/components/Search/styles.scss
tofumatt: It adds padding to the cards on the search page, but it might be possible to do without the nesting, I'll see.

src/amo/components/Search/styles.scss
tofumatt: They do in order to be scoped to this particular container; I didn't want the `SearchSort` to have these styles by default–I figured with things like margins we can set them in the container's styles scoped like this so different containers could position it slightly differently.

That said maybe I could just pass in a `className` so in general I'll see about improving the nesting here.

src/amo/components/SearchSort/SearchSort.scss
tofumatt: The issue with that is especially as we're using viewports that aren't tied to device sizes but just sane breakpoints... you'd end up without content when you turned your phone around.

In general I'd say "we should only load important content and if we're hiding a LOT of stuff entirely except desktop... yes maybe it isn't needed". But unless we can say a specific component should only load with a certain UA... I kinda dislike loading different sets of content. Take, for example, the search page on a Nexus 6 in portrait versus landscape:

### Portrait
screen shot 2017-06-21 at 14 12 21

### Landscape
screen shot 2017-06-21 at 14 12 24

### Landscape (w/o content loaded)
screen shot 2017-06-21 at 14 13 06

----

I think there's enough room to show the content and furthermore we should load it to prevent weirdness with different UAs getting/not getting content. I think that will make things harder to QA/test and debug, in addition to being a worse UX.

muffinresearch: Yep, I completely agree about the trade-offs that's why I mentioned the header approach is not equal to a viewport size based decision. Especially since the header available has no granularity say compared to using a device database to customize responses to the device more closely.

One place where with-holding content might be a benefit in the future is if significant page weight could be saved on mobile as opposed to just hiding the content via styling. 

This is something that can be considered on a case-by-case and as with any optimizations we can look at it later when we're specifically looking at ways to reduce page size / increase perf.

tofumatt: Yeah, I thought that too... with sagas we can have the notion of server-versus-client-side sagas. One way to do this might be to have the UA and viewport be stored in state and only fire off actions/saga requests for certain bits of data if the viewport is big enough.

I think that's only useful if the perf gain outweighs all the complexities that go with such a solution, but I'd be okay with that.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -0,0 +1,158 @@
...
+    let { store } = props;
+
+    if (!store) {
+      store = dispatchClientMetadata().store;
+      store.dispatch(searchStart({ filters: { query: 'test' } }));
+    }
+
+    return shallow(
+      <SearchContextCardBase
+        {...mapStateToProps(store.getState())}
```
kumar303: This is interesting. I think Enzyme gives us an opportunity to stop using the base component (which was always a hack).

However, [HOC rendering](https://github.com/airbnb/enzyme/issues/539) isn't supported very well yet.

Until it is, we could do it like this (I tested this locally). It's a bit hacky, what do you think? By rendering the real component chain I think we will get better coverage.

```diff
diff --git a/tests/unit/amo/components/TestSearchContextCard.js b/tests/unit/amo/components/TestSearchContextCard.js
index 6504093e..5e8861f4 100644
--- a/tests/unit/amo/components/TestSearchContextCard.js
+++ b/tests/unit/amo/components/TestSearchContextCard.js
@@ -1,10 +1,7 @@
 import { shallow } from 'enzyme';
 import React from 'react';
 
-import {
-  SearchContextCardBase,
-  mapStateToProps,
-} from 'amo/components/SearchContextCard';
+import SearchContextCard from 'amo/components/SearchContextCard';
 import { searchLoad, searchStart } from 'core/actions/search';
 import { dispatchClientMetadata, fakeAddon } from 'tests/unit/amo/helpers';
 import { getFakeI18nInst } from 'tests/unit/helpers';
@@ -19,13 +16,16 @@ describe('SearchContextCard', () => {
       store.dispatch(searchStart({ filters: { query: 'test' } }));
     }
 
-    return shallow(
-      <SearchContextCardBase
-        {...mapStateToProps(store.getState())}
+    const options = { context: { store } };
+
+    const root = shallow(
+      <SearchContextCard
         i18n={getFakeI18nInst()}
         {...props}
-      />
-    );
+      />, options
+    )
+
+    return root.first().shallow(options).first().shallow(options);
   }
 
   it('should render a card', () => {
```

If we add a new HOC to the chain, the test code will need an update but only in one place.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -0,0 +1,158 @@
...
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header'))
+      .toIncludeText('2 add-ons found');
+  });
+
+  it('should use singular form when only one result is found', () => {
+    const { store } = dispatchClientMetadata();
+    store.dispatch(searchLoad({
```
kumar303: It looks like you could replace this call with a helper and save quite a bit of code. In other words, the `result` part can be calculated automatically with a call like:

```js
dispatchSearchResults({ 
  store, 
  addons: { [fakeAddon.slug]: fakeAddon  },
  filters: { query: 'test' },
});
```

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -39,93 +18,104 @@ describe('<SearchResult />', () => {
     name: 'A search result',
     slug: 'a-search-result',
   };
-  const root = renderResult(result);
+
+  function render({ addon = baseAddon, lang = 'en-GB', ...props } = {}) {
+    return shallow(
+      <SearchResultBase
```
kumar303: I don't think you need to use the base component. All I see is a `translate()` wrapper so you could use the unwrapping approach above and test the final component directly.

src/amo/components/Search/index.js
```diff
@@ -71,3 +89,8 @@ export default class SearchPage extends React.Component {
     );
   }
 }
+
+export default compose(
+  safeAsyncConnect([{ promise: loadSearchResultsIfNeeded }]),
```
tofumatt: Yeah porting Search to saga will be awesome but later is a good idea. It'll require a lot of UX tweaks 😄

src/amo/components/SearchResult/index.js
```diff
@@ -0,0 +1,101 @@
...
+                {i18n.gettext('No theme preview available')}
+              </p>
+            )}
+          </div>
+
+          <div className="SearchResult-contents">
+            <h2 className="SearchResult-name">{addon.name}</h2>
+            <p
+              className="SearchResult-summary"
+              dangerouslySetInnerHTML={sanitizeHTML(addon.summary)}
```
tofumatt: It probably should... really I wonder if we should be doing that in the reducer.

src/amo/containers/CategoryPage.js
```diff
@@ -3,13 +3,13 @@ import React from 'react';
...
 
-import SearchPage from 'amo/components/SearchPage';
+import { SearchBase } from 'amo/components/Search';
 import { loadByCategoryIfNeeded, parsePage } from 'core/searchUtils';
 import { apiAddonType, safeAsyncConnect } from 'core/utils';
 
 
 export function CategoryPageBase(props) {
-  return <SearchPage enableSearchSort={false} {...props} />;
+  return <SearchBase enableSearchSort={false} {...props} />;
```
tofumatt: I don't quite follow: that's already how we switched from Category and it was pretty straightforward; once we switch to saga it would be a matter of checking in `SearchBase` for the search results given the filter set. If that component always is set up to dispatch when it doesn't have the right search data things will be good as pages like `amo/components/Category` also use the `SearchBase` component to render out their data.

To be the separate component used to load the data is the one we export using compose. I don't think it should go in a separate file though.

tests/unit/amo/components/TestSearchContextCard.js
tofumatt: I was with you until `return root.first().shallow(options).first().shallow(options);` 😆

I suppose we could write a helper for that, although I liked importing things directly from `enzyme`, it felt so clean!

What happens when we have multiple things in the `compose` chain like a `connect()` call and a `translate()` call?

kumar303: You will need to call `.first().shallow()` for each function passed to `compose`. I was thinking we could maybe do it in a less hacky way with a pattern like this:

```js
// in the component:
export wrappers = [
  translate(),
  connect(mapStateToProps),
];

export default compose(...wrappers)(MyComponent);

// In the test:

import MyComponent, { wrappers } from 'src/MyComponent.js'

// In render():
const root = unwrap(<MyComponent {...props} />, { wrappers });

// in unwrap():
let root = root.shallow(); // top-level
wrappers.forEach(() => {
  // Unwrap each HOC:
  root = root.first().shallow();
});
```

tofumatt: I think that's a great approach. 👍

I'd be happy to go through all the components that need it and do it in one PR rather than in this one (it's already pretty big). But let me know if you'd rather me do it here.

src/amo/components/SearchResult/index.js
tofumatt: Filed #2626.

tofumatt:
As we discussed on IRC I think the component setup here is okay for now and shouldn't make Saga refactoring complicated, so this is ready for another r? if that's okay. Let me know if there's still stuff you think should be refactored but I think this should be good and it will help make #2625 easier to land too.

src/amo/components/SearchResult/index.js
tofumatt: Though I opted not to use description here because it has the potential to be very long and I don't want all that text on the search results page. We will deal with it better in #2626 where if we find a description but no summary we can make a shortened summary from description I think. But for now they are fine blank.

tests/unit/amo/components/TestSearchContextCard.js
kumar303: Let's do it in https://github.com/mozilla/addons-frontend/issues/2643

Issue #2608, opened 2017-06-19
Convert add-on review list to sagas


kumar303:
Instead of using async-connect to pre-load data for the add-on review listing, we should use a saga. 

Steps to test for regressions:
* Load an add-on detail page for an add-on that has reviews
* Click 'Read N reviews'
* Make sure the review listing still works

This is a step towards https://github.com/mozilla/addons-frontend/issues/2356

vcarciu:
@kumar303 : Please let me know if this is ok and will be solved in another issue, or this should be reopened: 
![reviewsbeforeloading](https://user-images.githubusercontent.com/1583842/27592843-2f0582ba-5b5e-11e7-8d3c-45fefa34c547.gif)

kumar303:
Hi @vcarciu , your screencast looks right. What were you expecting to happen that should be solved in another issue? Thanks for testing it out.

PR #2619, opened 2017-06-21
chore: Remove prefix from AddonReviewList and AddonReview

tofumatt:
This patch literally just renames `AddonReview` to `Review` and `AddonReviewList` to `ReviewList`. (It also moves them and their CSS into a folder (`index.js` and `styles.scss`).

I am about to do the desktop review page but found it harder to remember that the review components always had the `Addon` prefix which I think is unneeded: we only have reviews for addons so the namespacing just creates more typing and a harder-to-find component in alphabetical lists.

I figured I'd do this outside the style patch so it was less likely to cause conflicts, as I know @kumar303 wants to work on review sagas soon too.

### DIFF
--- a/src/amo/components/Review/index.js
+++ b/src/amo/components/Review/index.js
@@ -22,9 +22,9 @@ import type { ErrorHandler as ErrorHandlerType } from 'core/errorHandler';
 import type { ElementEvent } from 'core/types/dom';
 import type { DispatchFunc } from 'core/types/redux';
 
-import 'amo/css/AddonReview.scss';
+import './styles.scss';
 
-type AddonReviewProps = {|
+type ReviewProps = {|
   apiState?: ApiStateType,
   createLocalState: typeof defaultLocalStateCreator,
   debounce: typeof defaultDebounce,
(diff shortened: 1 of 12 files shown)


src/amo/components/Review/index.js
```diff
@@ -37,33 +37,38 @@ type AddonReviewProps = {|
...
 
-type AddonReviewState = {|
+type ReviewState = {|
   reviewBody: ?string,
 |};
 
-export class AddonReviewBase extends React.Component {
+// Changing this will invalidate any saved add-ons for users because it
+// changes the prefix we use in our offline storage key.
+const LOCALSTATE_PREFIX = 'AddonReview';
```
tofumatt: Added this because I realised any future over-zealous search+replaces might not catch this and it would render saved reviews un-get-able.

tofumatt:
We can do it later if this is going to block things, but I think yes there is a reason to do them: prefixing things with `Addon`, especially when it isn't something *directly* related to an addon is weird and makes it harder for me to find the component. I look for `Review` when looking for a review, not `AddonReview`. Even when we also have `Reviewer`, I'd rather:

* `Review`
* `ReviewList`
* `Reviewer`

than:

* `Addon`
* `AddonCompatibilityError`
* `AddonMeta`
* `AddonReview` <-- sandwiched in between here even though it's less related
* `AddonsCard`

In general I'd like to move away from prefixing all the components with `Addon` in general, if there are a much of components related directly to the `addon` object maybe they should go in a folder? But Reviews feel like one of the "higher-level" objects in the app and marking them as "part of the Addon" always confuses me.

I just find it easier to read code where things are `Review` over `AddonReview`. It also makes it easier to find and reason about in my file tree viewer in Atom.

kumar303:
> I think yes there is a reason to do them: prefixing things with Addon, especially when it isn't something directly related to an addon is weird and makes it harder for me to find the component. 

An add-on review is directly related to an add-on. It is a review that a user makes about an add-on and nothing else. But I suppose you could make that argument about any component inside the `src/amo` directory.

> I look for Review when looking for a review, not AddonReview. Even when we also have Reviewer, I'd rather...
> * Review
> * ReviewList
> * Reviewer


Looking at this list, I don't know what `Reviewer` would be used for. We will more likely have a "review" page where an editor will enter a review for an add-on. In that case, we can probably call it an `EditorialReview`. 

It doesn't really make sense to try and predict the future of DevHub anyway (which is a ways off), so, simplifying `AddonReview` to `Review` wouldn't bother me. However, if you really want to rename it, can it wait until after https://github.com/mozilla/addons-frontend/issues/2608 lands?

tofumatt:
To clarify "directly related to an add-on" I mean something like `AddonMeta` or `AddonCompatibilityError` which takes an add-on as its "main" prop and renders related info. So I think it's good to have those components so we don't have a monolithic `Addon` component, but something like `User` or `Review` is separate enough (even though yes every `Addon` has an author and every review an add-on).

Happy to wait until other things land, this only took about twenty minutes to do so I can do it again later. It's a few renames and a search/replace.

tofumatt:
Gonna close this for now; I reserve the right to redo it later but this is just noise right now.

PR #2625, opened 2017-06-22
feat: Add Desktop Category headers (fix #2563)

### LINKED ISSUE #2563: Implement a desktop categories page
Implement a desktop categories page of all categories for each add-on type.

tofumatt:
Fixes #2563.

### Small
screen shot 2017-06-22 at 20 54 49

### Medium / Large
screenshot 2017-06-22 20 54 31


tests/unit/amo/components/TestCategory.js
```diff
@@ -1,24 +1,59 @@
...
     const root = render();
-    expect(root.prop('enableSearchSort')).toEqual(false);
+
+    expect(root).toHaveClassName('Category');
+  });
+
+  it('dispatches categoriesFetch if category is falsy', () => {
+    render({ category: null });
+
+    sinon.assert.called(fakeDispatch);
```
kumar303: This should check that the right action is dispatched

tests/unit/amo/components/TestCategoryHeader.js
```diff
@@ -0,0 +1,101 @@
...
+  it('omits addonType if there is no category', () => {
+    const root = render({ category: null });
+
+    expect(root).not.toHaveClassName('CategoryHeader--type-undefined');
+  });
+
+  it('renders LoadingText if category is null (e.g. is loading)', () => {
+    const root = render({ category: null });
+
+    expect(root.find('.CategoryHeader-name')).toIncludeText('<LoadingText />');
```
kumar303: Huh, I didn't know you could use `toIncludeText` to check for a component

kumar303: I guess it's not a big deal but I like how `expect(root.find('.CategoryHeader-name').find(LoadingText)).toHaveLength(1)` would allow you to change the `width` property without having to update the test.

tests/unit/core/test_utils.js
```diff
@@ -957,3 +972,107 @@ describe('getClientCompatibility', () => {
...
+  it('handles a false-y category', () => {
+    const category = null;
+    const categoryColor = getCategoryColor(category);
+
+    expect(categoryColor).toEqual(1);
+  });
+
+  it('all valid addonTypes are in CATEGORY_COLORS', () => {
+    expect(() => {
+      validAddonTypes.forEach((addonType) => {
```
kumar303: with Flow you don't need tests like this. Just sayin' :) If this loop were more expensive I'd be more worried but it's no big deal.

tests/unit/core/test_utils.js
```diff
@@ -957,3 +972,107 @@ describe('getClientCompatibility', () => {
...
+
+  it('throws on unrecognised addonType', () => {
+    expect(() => {
+      getCategoryColor({ id: 5, type: 'NOT_A_REAL_TYPE' });
+    }).toThrowError(
+      'addonType "NOT_A_REAL_TYPE" not found in CATEGORY_COLORS');
+  });
+
+  it('deals with hundreds of categories', () => {
+    for (let i = 1; i < 800; i++) {
```
kumar303: Is it really necessary to run this 800 times? It seems a bit excessive and might slow down the test suite. Can you do a smoke test of the upper range instead somehow?

tofumatt: Totally, I was being lazy and meant to revisit this. 👍

src/amo/components/CategoryHeader/index.js
```diff
@@ -0,0 +1,83 @@
...
+            <h1 className="CategoryHeader-name">
+              {category ? category.name : <LoadingText />}
+            </h1>
+            <div className="CategoryHeader-description">
+              {category ? (
+                <p className="CategoryHeader-paragraph">
+                  {category.description}
+                </p>
+              ) : (
+                <p className="CategoryHeader-paragraph">
```
tofumatt: We need an element wrapping things in React so that won't work because we have three elements below. It's better to do it this way than put a <div> inside the <p> around the loading texts.

tests/unit/amo/components/TestCategoryHeader.js
tofumatt: Oh, no, that's a good point. I'll change it! 👍

tofumatt:
Thanks for the review, set for another r?

src/amo/components/CategoryHeader/index.js
kumar303: ah, right, I see

tests/unit/amo/components/TestCategoryHeader.js
```diff
@@ -0,0 +1,110 @@
...
+    const root = render({ category });
+
+    expect(root.find('.CategoryHeader-description'))
+      .toIncludeText('Browse all themes in this category');
+  });
+
+  it('omits addonType if there is no category', () => {
+    const root = render({ category: null });
+
+    expect(root).not.toHaveClassName('CategoryHeader--type-undefined');
```
kumar303: I guess technically this would be `'CategoryHeader--type-null'` if the code didn't account for it? It's a bit hard to write tests like this because you're anticipating a bug that doesn't exist. Could you use `expect(root).toHaveClassName(...)` for a more positive assertion?

tests/unit/amo/components/TestCategoryHeader.js
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  it('omits color if there is no category', () => {
+    const root = render({ category: null });
+
+    expect(root).not.toHaveClassName(
+      'CategoryHeader--category-color-undefined');
```
kumar303: Same thing here: could this somehow use `expect(root).toHaveClassName(...)` instead?

tests/unit/core/test_utils.js
```diff
@@ -957,3 +972,106 @@ describe('getClientCompatibility', () => {
...
+  });
+
+  it('deals with high category IDs', () => {
+    for (let i = 750; i < 800; i++) {
+      const category = { id: i, type: ADDON_TYPE_THEME };
+      const categoryColor = getCategoryColor(category);
+
+      expect(categoryColor)
+        .toBeLessThanOrEqual(CATEGORY_COLORS[ADDON_TYPE_THEME]);
+      expect(categoryColor).toBeGreaterThanOrEqual(1);
```
kumar303: Would it also be helpful to check that this number is less than 12?

tests/unit/core/test_utils.js
```diff
@@ -957,3 +972,106 @@ describe('getClientCompatibility', () => {
...
+    }
+  });
+
+  it('has a different number of colors for different addonTypes', () => {
+    const category = { id: 11, type: ADDON_TYPE_EXTENSION };
+    const categoryColor = getCategoryColor(category);
+
+    expect(categoryColor).toEqual(1);
+    expect(categoryColor).toBeLessThan(11);
```
kumar303: This assertion seems unnecessary as long as you are using `toEqual()` right before it

src/core/utils.js
```diff
@@ -393,3 +394,42 @@ export function getClientCompatibility({
 
   return { compatible, maxVersion, minVersion, reason };
 }
+
+export function getCategoryColor(category) {
+  if (!category) {
+    throw new Error('category is required.');
+  }
+
+  if (!CATEGORY_COLORS[category.type]) {
```
kumar303: Just a nit but you could move `const maxColors = CATEGORY_COLORS[category.type];` up above this and reference `maxColors` instead.

src/core/utils.js
```diff
@@ -393,3 +394,42 @@ export function getClientCompatibility({
...
+    throw new Error(
+      `addonType "${category.type}" not found in CATEGORY_COLORS.`);
+  }
+
+  const maxColors = CATEGORY_COLORS[category.type];
+  if (category.id > maxColors) {
+    const color = parseInt(category.id / maxColors, 10);
```
kumar303: Would this be any easier with a [modulo](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Remainder_())?

tests/unit/amo/components/TestCategoryHeader.js
tofumatt: Yeah I agree. It means I'll have to check for exact class name because I want to make sure certain ones don't show, so it's a BIT more of a pain to maintain but that's better. Will change.

tests/unit/core/test_utils.js
tofumatt: Oops, good call.

src/core/utils.js
tofumatt: It probably would actually, I'll do that if it makes sense (which it probably does).

PR #2638, opened 2017-06-27
adds getIconUrl function

### LINKED ISSUE #2623: Get add-on icon more safely
We use this line of code a lot:

```js
const iconURL = isAllowedOrigin(addon.icon_url) ? addon.icon_url : fallbackIcon;
```

We should wrap this up in a function because it would be a shame if someone forgot to check the origin of the icon URL.

tsl143:
Fixes #2623


src/core/utils.js
```diff
@@ -137,6 +136,10 @@ export function isAllowedOrigin(urlString, {
   return allowedOrigins.includes(`${parsedURL.protocol}//${parsedURL.host}`);
 }
 
+export function getIconUrl(addon, fallback) {
+  return (addon && isAllowedOrigin(addon.icon_url)) ? addon.icon_url : fallback;
```
kumar303: I think it would be nice to make the `fallback` parameter be optional. By default it can use `'amo/img/icons/default-64.png'` like the original code did.

tsl143:
@kumar303 I tried importing png icon in util to make it as fallback icon but I get following error while building package
`SyntaxError: /<...>/addons-frontend/src/amo/img/icons/default-64.png: Unexpected character '�' (1:0)`

I tried to add `import fallbackIcon from 'amo/img/icons/default-64.png';` in `core/util`

tsl143:
@kumar303 Updated the code, but for some reason, Travis didnt run the test on PR

src/core/imageUtils.js
```diff
@@ -0,0 +1,10 @@
+/* global navigator, window */
```
kumar303: I don't see navigator or window being used so you can delete this

src/core/imageUtils.js
```diff
@@ -0,0 +1,10 @@
...
+/* eslint-disable react/prop-types */
```
kumar303: I don't see any react components so you can delete this

src/core/imageUtils.js
```diff
@@ -0,0 +1,10 @@
...
+
+import { isAllowedOrigin } from 'core/utils';
+import fallbackIcon from 'amo/img/icons/default-64.png';
+
+export function getIconUrl(addon) {
```
kumar303: This should be named `getAddonIconUrl` because I'm pretty sure we'll need to add one for themes later.

tests/unit/core/test_utils.js
```diff
@@ -540,6 +542,17 @@ describe('isAllowedOrigin', () => {
   });
 });
 
+describe('getIconUrl', () => {
```
kumar303: This should be moved to a new file named `tests/unit/core/testImageUtils.js`

src/core/imageUtils.js
```diff
@@ -0,0 +1,10 @@
...
+  return (addon && isAllowedOrigin(addon.icon_url)) ? addon.icon_url : fallbackIcon;
```
kumar303: This should have a line above it checking that `addon` is not empty, like:

```js
if (!addon) {
  throw new Error('addon cannot be empty');
}
```
This will also need a unit test

tests/unit/core/test_utils.js
```diff
@@ -540,6 +542,17 @@ describe('isAllowedOrigin', () => {
...
+  const addon = { ...fakeAddon };
+
+  it('return icon url as in fake addon', () => {
+    expect(getIconUrl(addon)).toEqual('https://addons.cdn.mozilla.net/webdev-64.png');
+  });
+  it('return fallback icon in case of non allowed origin', () => {
+    expect(getIconUrl({ icon_url: 'https://xyz.com/a.png' })).toEqual(fallbackIcon);
```
kumar303: This should inherit from `fakeAddon` like: 

```js
getIconUrl({ ...fakeAddon, icon_url: 'https://xyz.com/a.png' })
```

tests/unit/core/test_utils.js
```diff
@@ -540,6 +542,17 @@ describe('isAllowedOrigin', () => {
   });
 });
 
+describe('getIconUrl', () => {
+  const addon = { ...fakeAddon };
+
+  it('return icon url as in fake addon', () => {
+    expect(getIconUrl(addon)).toEqual('https://addons.cdn.mozilla.net/webdev-64.png');
+  });
+  it('return fallback icon in case of non allowed origin', () => {
```
kumar303: These two functions should be separated by a new line for readability

tests/unit/core/test_utils.js
```diff
@@ -540,6 +542,17 @@ describe('isAllowedOrigin', () => {
   });
 });
 
+describe('getIconUrl', () => {
+  const addon = { ...fakeAddon };
+
+  it('return icon url as in fake addon', () => {
+    expect(getIconUrl(addon)).toEqual('https://addons.cdn.mozilla.net/webdev-64.png');
```
kumar303: For a test like this I think it helps to input a specific value that you're depending on even though the default value is sufficient. This makes the test very explicit and easier to understand. Example:

```js
const allowedIcon = 'https://addons.cdn.mozilla.net/any-icon.png';
expect(getIconUrl({ ...fakeAddon, icon_url: allowedIcon })).toEqual(allowedIcon);
```

tsl143:
@kumar303 I did required changes but there is one issue,
By adding 
`  if (!addon) {
 	throw new Error('addon cannot be empty');
  }`

few test fails 
` ● amo/components/AddonReviewList › <AddonReviewListBase/> › requires an addon prop to produce a URL

    expect(function).toThrowError(regexp)
    
    Expected the function to throw an error matching:
      /cannot access addonURL/
    Instead, it threw:
      addon cannot be empty `

src/core/imageUtils.js
```diff
@@ -0,0 +1,9 @@
+import { isAllowedOrigin } from 'core/utils';
+import fallbackIcon from 'amo/img/icons/default-64.png';
+
+export function getAddonIconUrl(addon) {
+  if (!addon) {
+    throw new Error('addon cannot be empty');
```
kumar303: Ah, you were right. This should not throw an error. In the case of a falsy addon, the `getAddonIconUrl()` function should return the `fallbackIcon`

tsl143:
@kumar303 All set from my side :)

tests/unit/core/test_imageUtils.js
```diff
@@ -0,0 +1,16 @@
...
+  const allowedIcon = 'https://addons.cdn.mozilla.net/webdev-64.png';
+
+  it('return icon url as in fake addon', () => {
+    expect(getAddonIconUrl({ ...fakeAddon, icon_url: allowedIcon })).toEqual(allowedIcon);
+  });
+
+  it('return fallback icon in case of non allowed origin', () => {
+    expect(getAddonIconUrl({ ...fakeAddon, icon_url: 'https://xyz.com/a.png' })).toEqual(fallbackIcon);
+  });
+});
```
kumar303: This needs one more test now that makes sure a falsy add-on (like `getAddonIconUrl(null)`) will return `fallbackIcon`

tsl143: @kumar303 stay tuned working on it rn

tsl143:
@kumar303 All set, is it fine now to get merged?

PR #2648, opened 2017-06-28
fix: Complete themes are valid add-ons

### LINKED ISSUE #2646: Complete themes cause an error because they are unsupported
See the error here: https://sentry.prod.mozaws.net/operations/addons-frontend-amo-dev/issues/618227/

Visit https://addons-dev.allizom.org/en-GB/android/addon/firefox-2-theme-for-firefox-3x/ when *signed out*, and there will be an error. This is because we don't support complete themes in `validAddonTypes`, but evidently we should 😄 

### What happened?
There was an error.

No error!

Easy fix, incoming.

tofumatt:
Fixes #2646.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -136,7 +136,6 @@ export class AddonBase extends React.Component {
           className="Addon-theme-header"
           id="Addon-theme-header"
           data-browsertheme={getBrowserThemeData()}
-          ref={(el) => { this.wrapper = el; }}
           onClick={this.onClick}
         >
           {installStatus !== ENABLED ? (
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -50,7 +50,11 @@ export const ADDON_TYPE_EXTENSION = 'extension';
 export const ADDON_TYPE_LANG = 'language';
 export const ADDON_TYPE_OPENSEARCH = 'search';
 export const ADDON_TYPE_THEME = 'persona';
+// TODO: Remove ADDON_TYPE_COMPLETE_THEME once we don't support complete
+// themes.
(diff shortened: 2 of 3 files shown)


src/amo/components/Addon/index.js
```diff
@@ -136,7 +136,6 @@ export class AddonBase extends React.Component {
           className="Addon-theme-header"
           id="Addon-theme-header"
           data-browsertheme={getBrowserThemeData()}
-          ref={(el) => { this.wrapper = el; }}
```
tofumatt: This was unused and caused a non-100% coverage error so I deleted it. I know it's unrelated but figured it was worth it.

src/core/constants.js
```diff
@@ -50,7 +50,9 @@ export const ADDON_TYPE_EXTENSION = 'extension';
 export const ADDON_TYPE_LANG = 'language';
 export const ADDON_TYPE_OPENSEARCH = 'search';
 export const ADDON_TYPE_THEME = 'persona';
+export const ADDON_TYPE_CLASSIC_THEME = 'theme';
```
kumar303: can we just call this `ADDON_TYPE_COMPLETE_THEME`?

src/core/constants.js
```diff
@@ -50,7 +50,9 @@ export const ADDON_TYPE_EXTENSION = 'extension';
...
 export const validAddonTypes = [
+  ADDON_TYPE_CLASSIC_THEME, // Remove once we don't support complete themes.
```
kumar303: One day we might start grepping `TODO` so it would probably make sense to add that to the comment so we don't forget

src/core/constants.js
tofumatt: Sure, no problem.

PR #2670, opened 2017-07-03
Fix Multiline Review issue #2441

### LINKED ISSUE #2441: Multi line review is written in a single line
Steps to reproduce:
1.On Android, go to any add-on and add a multi line review
2.Open the review in order to see it

Expected results:
Review should be displayed multi line 

Actual results:
Review is displayed in a single line

multiline

tsl143:
Fix #2441 
Before 

screen shot 2017-07-03 at 5 07 26 pm

After
screen shot 2017-07-03 at 5 06 52 pm

### DIFF
--- a/src/amo/components/AddonReviewList.js
+++ b/src/amo/components/AddonReviewList.js
@@ -107,33 +107,29 @@ export class AddonReviewListBase extends React.Component {
     return `${this.addonURL()}reviews/`;
   }
 
-  setReviewBody(review) {
-    let reviewBody = <p><LoadingText /></p>;
-    if (review) {
-      const reviewBodySanitized = sanitizeHTML(nl2br(review.body), ['br']);
-      reviewBody = <p dangerouslySetInnerHTML={reviewBodySanitized} />;
-    }
-    return reviewBody;
-  }
-
   renderReview({ review, key }: RenderReviewParams) {
     const { i18n } = this.props;
 
     let byLine;
+    let reviewBody;
     if (review) {
       const timestamp = i18n.moment(review.created).fromNow();
       // L10n: Example: "from Jose, last week"
       byLine = i18n.sprintf(
         i18n.gettext('from %(authorName)s, %(timestamp)s'),
           { authorName: review.userName, timestamp });
+
+      const reviewBodySanitized = sanitizeHTML(nl2br(review.body), ['br']);
+      reviewBody = <p dangerouslySetInnerHTML={reviewBodySanitized} />;
     } else {
       byLine = <LoadingText />;
+      reviewBody = <p><LoadingText /></p>;
     }
 
     return (
       <li className="AddonReviewList-li" key={key}>
         <h3>{review ? review.title : <LoadingText />}</h3>
-        {this.setReviewBody(review)}
+        {reviewBody}
         <div className="AddonReviewList-by-line">
           {review ?
             <Rating styleName="small" rating={review.rating} readOnly />
--- a/src/amo/components/AddonReviewList.js
+++ b/src/amo/components/AddonReviewList.js
@@ -110,8 +110,8 @@ export class AddonReviewListBase extends React.Component {
   setReviewBody(review) {
     let reviewBody = <p><LoadingText /></p>;
     if (review) {
-      const content = sanitizeHTML(nl2br(review.body), ['br']);
-      reviewBody = <p dangerouslySetInnerHTML={content} />;
+      const reviewBodySanitized = sanitizeHTML(nl2br(review.body), ['br']);
+      reviewBody = <p dangerouslySetInnerHTML={reviewBodySanitized} />;
     }
     return reviewBody;
   }
--- a/src/amo/components/AddonReviewList.js
+++ b/src/amo/components/AddonReviewList.js
@@ -11,7 +11,7 @@ import { fetchAddon } from 'core/actions/addons';
 import Paginate from 'core/components/Paginate';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
-import { findAddon } from 'core/utils';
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonReviewList.js
```diff
@@ -107,6 +107,13 @@ export class AddonReviewListBase extends React.Component {
     return `${this.addonURL()}reviews/`;
   }
 
+  formatReviewBody(text) {
+    const formattedReview = text.split('\n').map((i) => {
+      return <div key={i}>{i}</div>;
```
tofumatt: Because these reviews are already inside a `<p>` tag, I'd prefer not putting divs inside them. This is a bit of a hack anyway because the divs are block-level elements.

A better approach would be to convert the newline characters to `<br>` tags. We already have a utility function `nl2br` in `core/utils` you can use for that.

tsl143:
@tofumatt  I already tried using `<br>` but it renders as follows
screen shot 2017-07-03 at 6 00 29 pm

And for sure we must not use `DIV` inside `p`, can we use `span` instead and use `display:block`?

tofumatt:
It's rendering the `<br />` tags because React will escape tags inside content as a security precaution. To render HTML from a variable inside a React component you'll need to use `<p dangerousSetInnerHTML={santizeHTML(content, ['br'])} />` where `content` is the result of the `nl2br(review.body)`.

src/amo/components/AddonReviewList.js
```diff
@@ -107,6 +107,15 @@ export class AddonReviewListBase extends React.Component {
     return `${this.addonURL()}reviews/`;
   }
 
+  setReviewBody(review) {
+    let reviewBody = <p><LoadingText /></p>;
+    if (review) {
+      const content = sanitizeHTML(nl2br(review.body), ['br']);
+      reviewBody = <p dangerouslySetInnerHTML={content} />;
```
tofumatt: I'd rename `content` to something more meaningful; `content` is vague but `reviewBodySanitized` would be good.

src/amo/components/AddonReviewList.js
```diff
@@ -107,6 +107,15 @@ export class AddonReviewListBase extends React.Component {
     return `${this.addonURL()}reviews/`;
   }
 
+  setReviewBody(review) {
```
tofumatt: `reviewBody` or `reviewBodyHTML` would be better method names; this doesn't actually set anything.

src/amo/components/AddonReviewList.js
tsl143: Okay, also I am writing html render test case for this as the shallow render fails dangerouslysetInnerHtml.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -60,6 +67,43 @@ describe('amo/components/AddonReviewList', () => {
...
+        location: { query: {} },
+        params,
+        reviewCount: loadedReviews && loadedReviews.length,
+        reviews: loadedReviews,
+        ...customProps,
+      };
+
+      const store = createStore().store;
+
+      return findDOMNode(findRenderedComponentWithType(renderIntoDocument(
```
tofumatt: We will be migrating away from these tests and only want to test with enzyme. What is the error you're encountering with `shallow`? I believe we have other tests where it works just fine with `dangerouslySetInnerHTML`.

tsl143: It renders Just blank HTML.
I got this read - https://github.com/airbnb/enzyme/issues/630

tsl143: screen shot 2017-07-03 at 11 15 25 pm

tofumatt: Ah yes, that makes sense actually reading that comment.

You'll want to use https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md or possibly https://github.com/airbnb/enzyme/blob/master/docs/api/render.md. You should definitely use enzyme though, let me know if mount works or if you need help with the tests, I can probably help you write it if you get stuck.

tsl143: Sure, I will give it a try right now :)

tsl143:
@tofumatt , Finally with all check passed :)
Screenshots are same as before.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -278,7 +278,7 @@ describe('amo/components/AddonReviewList', () => {
         .toHaveText(fakeReview.title);
 
       expect(root.find('.AddonReviewList-li p'))
-        .toHaveText(fakeReview.body);
+        .toHaveHTML(`<p>${fakeReview.body}</p>`);
```
tofumatt: This isn't actually checking for a `<br />` tag, which is the point of this patch. This test should actually revert to `.toHaveText(fakeReview.body);` I would think (it should still pass).

tsl143: `toHaveText` does not renders the `dangerouslySetInnerHTML`, it gives blank `p` instead.

tofumatt: Ah okay, no problem then 😀

tsl143:
@tofumatt Done as requested.
Honestly, I am more than happy rn, legit everyday I am learning new things. I was really scared of writing testcases and always choose to skip it, but I cant believe I am writing them and now I know why writing test cases is important.
Thanks a lot for being patient while I commit mistakes :)

tofumatt:
@tsl143 Awesome! I remember when I feared writing tests too, and you're right: it feels really confidence-inspiring once you get them sorted knowing your code will be safer to refactor/change! 👍

PR #2684, opened 2017-07-04
Invalid category slugs should result in a 404

### LINKED ISSUE #2084: Invalid category slugs should result in a 404 rather than no results.
If a CategoryPage slug passed is invalid it should be a 404 rather than just show no results. 

See https://addons-dev.allizom.org/en-US/firefox/extensions/whatever/

Presumably to do this we'd need to fetch a list of cats for the clientApp dynamically. 

We should do this differently per-client application in any case since currently we don't change the list of categories based on app.

tsl143:
Fixes #2084  - Invalid category slugs should result in a 404 rather than no results. 

Also this will remove console errors due to invalid categories.

**Before**

screen shot 2017-07-04 at 12 35 36 pm

**After**
screen shot 2017-07-04 at 12 32 52 pm


tsl143:
Some click on homepage tiles also gives 404 due to https://github.com/mozilla/addons-frontend/issues/2685

muffinresearch:
@tsl143 thanks for the patch! 

Could you add a test to cover this behaviour and then this can be landed.

tests/unit/amo/components/TestCategory.js
```diff
@@ -36,6 +43,23 @@ describe('Category', () => {
     );
   }
 
+  function renderHTML({ ...props }) {
+    const { store } = createStore();
+
+    return findDOMNode(findRenderedComponentWithType(renderIntoDocument(
```
tofumatt: This test should be using enzyme.

tests/unit/amo/components/TestCategory.js
```diff
@@ -48,6 +72,11 @@ describe('Category', () => {
     sinon.assert.calledWith(fakeDispatch, categoriesFetch());
   });
 
+  it('renders 404 page for missing/invalid category', () => {
+    const root = renderHTML({ category: null });
```
tofumatt: As mentioned you should be able to use enzyme for this test, though you might need to use `mount` or `render` because we're loading other components.

src/amo/components/Category/index.js
```diff
@@ -32,6 +33,10 @@ export class CategoryBase extends React.Component {
   render() {
     const { category, ...searchProps } = this.props;
 
+    if (!category) {
```
tofumatt: `category` will also be `null` if this page is loading, which will happen with a saga. We should be proactive and make this handle both the loading case and the empty case.

Looks like the [categories reducer](https://github.com/mozilla/addons-frontend/blob/master/src/core/reducers/categories.js#L37) adds a loading state we can use to check if the page is loading or has already loaded. If `loading` is `false` and `category` is false-y then we can return `<NotFound />`; otherwise this will 404 during category loads.

tsl143:
I have put `loading === false` already, but finding it hard to put enzyme test for same, trying rn.

tsl143:
@tofumatt , rewritten the test case with enzyme :)

Also all test cases passed at my local. One failed with Travis though giving following error. Is this a server issue?

screen shot 2017-07-10 at 3 42 53 pm

tsl143:
Ping @tofumatt

muffinresearch:
>Also all test cases passed at my local. One failed with Travis though giving following error. Is this a server issue?

That's likely to be a spurious issue. You could try rebasing against master.

tsl143:
@muffinresearch Also can you review this ?

tests/unit/amo/components/TestCategory.js
```diff
@@ -48,6 +66,12 @@ describe('Category', () => {
     sinon.assert.calledWith(fakeDispatch, categoriesFetch());
   });
 
+  it('return 404 if category is falsy', () => {
```
muffinresearch: Would be good to have a test to cover when loading is true as well.

tsl143:
@caitmuenster Done.

tsl143:
ping @tofumatt @muffinresearch

tofumatt:
Hi @tsl143, we've seen the issue and are flagged to have another look; we've just been very busy with some other issues in the project. In the future there's no need to ping us as when we're following the PR we get notified when you push new commits.

tsl143:
okay.

muffinresearch:
I think this is looking pretty good. @tofumatt are you happy for this patch to land?

tests/unit/amo/components/TestCategory.js
```diff
@@ -1,14 +1,16 @@
...
 import { CategoryBase, mapStateToProps } from 'amo/components/Category';
 import { SearchBase } from 'amo/components/Search';
 import createStore from 'amo/store';
 import { categoriesFetch } from 'core/actions/categories';
 import { searchStart } from 'core/actions/search';
 import { ADDON_TYPE_THEME } from 'core/constants';
+import I18nProvider from 'core/i18n/Provider';
 import { getFakeI18nInst } from 'tests/unit/helpers';
-
+import NotFound from 'amo/components/ErrorPage/NotFound';
```
tofumatt: Nitpick but this should be above Line 5 `import { SearchBase } from 'amo/components/Search'; `. Not sure why ESLint didn't catch it 😢

tests/unit/amo/components/TestCategory.js
```diff
@@ -36,6 +38,22 @@ describe('Category', () => {
     );
   }
 
+  function mountRender(props = { loading: false }) {
+    const { store } = createStore();
```
tofumatt: Instead of using `createStore` you should use the test helper `dispatchClientMetadata()` (which we should probably rename, sorry!) from `tests/unit/amo/helpers`. It dispatches things like lang and clientApp which should be available when using providers, links, etc.

tests/unit/amo/components/TestCategory.js
```diff
@@ -48,6 +66,18 @@ describe('Category', () => {
     sinon.assert.calledWith(fakeDispatch, categoriesFetch());
   });
 
+  it('return 404 if category is falsy and loading is false', () => {
```
tofumatt: Nitpick but this should be `it('should return [...]')`.

tests/unit/amo/components/TestCategory.js
```diff
@@ -48,6 +66,18 @@ describe('Category', () => {
...
+    const root = mountRender();
+
+    expect(root.find(NotFound)).toHaveLength(1);
+  });
+
+  it('not return 404 if category is falsy and loading is true', () => {
```
tofumatt: Nitpick but this should be `it('should not return [...]')`.

tofumatt:
If @tsl143 can't get to this before the tag I'll pick it up so we can get it merged. 😄

tofumatt:
Nice, that was fast! 👍

Once this passes tests I'll get it merged. Thanks @tsl143!

tsl143:
You are welcome.
Just the right time in my timezone :)

tsl143:
@tofumatt it completed all checks here
but on my local machine when I run `yarn lint` it gives following error.

`/Users/trishulgoel/Documents/addons-frontend/tests/unit/core/server/testServer.js
  6:23  error  Unable to resolve path to module 'supertest'  import/no-unresolved`

tofumatt:
That is because we updated the deps in master and you need to run either npm install or yarn install.

tsl143:
Oh got it.
This is good to merge then.

tests/unit/amo/components/TestCategory.js
muffinresearch: I don't think we have import sorting turned on. I think there's some linting based on groupings of internal and external imports etc but otherwise no other sorting. See also: https://github.com/airbnb/javascript/blob/c43fc749c4213e4574a4b4a3d0fe8784d5cd5b5e/packages/eslint-config-airbnb-base/rules/es6.js#L151

PR #2695, opened 2017-07-06
Extensions page title is not correctly aligned in RTL #2689

### LINKED ISSUE #2689: [RTL] Extensions page title is not correctly aligned in RTL
Steps to reproduce:
1.Go to AMO-dev , mobile view
2.Select extensions tab and then switch to any RTL language (https://addons-dev.allizom.org/ar/firefox/extensions/)

Expected results:
Everything looks on in RTL

Actual results:
Extensions page title is not correctly aligned in RTL

Please see a video for this issue:
extrtl

tsl143:
Fixes #2689 

**Before:**
screen shot 2017-07-06 at 4 01 35 pm

**After:**
screen shot 2017-07-06 at 4 13 19 pm

### DIFF
--- a/src/amo/components/LandingPage/styles.scss
+++ b/src/amo/components/LandingPage/styles.scss
@@ -9,6 +9,7 @@
   @include margin-end(20px);
 }
 
+// In LTR, the spacing actually looks correct because of the icon itself has built-in margins in its design.
 .LandingPage-icon--extension {
   [dir=rtl] & {
     margin-left: 20px;
--- a/src/amo/components/LandingPage/styles.scss
+++ b/src/amo/components/LandingPage/styles.scss
@@ -5,10 +5,6 @@
   padding: $padding-page;
 }
 
-.LandingPage-icon--persona {
-  @include margin-end(20px);
-}
-
 .LandingPage-header {
   align-items: center;
   display: flex;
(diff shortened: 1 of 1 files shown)


src/amo/components/LandingPage/styles.scss
```diff
@@ -31,6 +27,8 @@
   height: 89px;
   width: 225px;
 
+  @include margin-end(20px);
```
tofumatt: As mentioned in the issue, in LTR the spacing actually looks correct because of the (mostly) whitespace in the icon itself. If you zoom in you can see there is a faint squiggle:

screenshot 2017-07-06 19 18 28

This is rare case where we want separate styles for RTL, I think. So the themes spacing should return and `margin-left: 20px;` should be applied to `.LandingPage-icon`.

src/amo/components/LandingPage/styles.scss
```diff
@@ -9,6 +9,12 @@
   @include margin-end(20px);
 }
 
+.LandingPage-icon--extension {
+  [dir=rtl] & {
```
tofumatt: This could use a comment explaining why we are doing this only on rtl. Something like: `// In LTR, the spacing actually looks correct because of the icon itself has built-in margins in its design.`

tsl143: @tofumatt Done :)

Issue #2698, opened 2017-07-06
Erroneous trailing quote


muffinresearch:
See https://github.com/mozilla/addons-frontend/blob/cc13364b2cca75e9b6ab131f0cead976b13a1e19/src/amo/components/SearchContextCard/index.js#L46

Extraction translation is here:

https://github.com/mozilla/addons-frontend/blob/cc13364b2cca75e9b6ab131f0cead976b13a1e19/locale/templates/LC_MESSAGES/amo.pot#L516

```
#: src/amo/components/SearchContextCard/index.js:45
msgid "%(count)s add-on found\""
msgid_plural "%(count)s add-ons found"
msgstr[0] ""
msgstr[1] "

```

Originally filed here: https://github.com/mozilla/addons/issues/436

tofumatt:
Fixed in #2701

SVSanta:
Please add some STR or the label "qa: not needed". Thanks!

PR #2701, opened 2017-07-07
Fix Erroneous trailing quote #2698

tsl143:
Erroneous trailing quote #2698

### DIFF
--- a/locale/templates/LC_MESSAGES/amo.pot
+++ b/locale/templates/LC_MESSAGES/amo.pot
@@ -514,7 +514,7 @@ msgid "No add-ons found"
 msgstr ""
 
 #: src/amo/components/SearchContextCard/index.js:45
-msgid "%(count)s add-on found\""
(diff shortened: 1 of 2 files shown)


locale/templates/LC_MESSAGES/amo.pot
```diff
@@ -514,7 +514,7 @@ msgid "No add-ons found"
 msgstr ""
 
 #: src/amo/components/SearchContextCard/index.js:45
-msgid "%(count)s add-on found\""
+msgid "%(count)s add-on found"
```
tofumatt: You shouldn't need to modify/commit this file. We'll do that as part of our locale extract/merge. Just changing the source should be fine.

muffinresearch:
Following this landing we might want to consider manually updating the key in the translations so we don't result in fuzzy translations OR alternatively we can just extract and it will get fixed up eventually.

PR #2703, opened 2017-07-07
 Blank search will show no results for Explore tab #2702

### LINKED ISSUE #2702: Blank search will show no results for Explore tab
Steps to reproduce:
1.On your Android device, go to AMO-dev
2.While Explore tab is selected,  perform a blank search

Expected results:
If blank search is allowed, a list of results is displayed

Actual results:
Number of results is displayed, but the list of results is missing

NOTES:
For Extensions or Themes tabs , blank search will show the results

blanksearchandroid

tsl143:
Fix  #2702


tsl143:
Video of fix

ezgif-1-6f8bf63081

muffinresearch:
As mentioned in the linked issue #2702 I think this needs UX before we move forward with a solution.

muffinresearch:
@tsl143 we now have the UX for this: https://github.com/mozilla/addons-frontend/issues/2702#issuecomment-321400696 - let me know if you want to continue working on it. 

The additions will need some tests added to cover the changes.

tsl143:
Yes I am working on it right now.
I am planning to update following test to cover the new change
https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestSearchContextCard.js#L75

tsl143:
**Before:**
**1)**
screen shot 2017-08-18 at 3 38 25 pm
**2)**
screen shot 2017-08-18 at 3 37 34 pm

**After:**
screen shot 2017-08-18 at 3 37 34 pm

tsl143:
Ping @muffinresearch , this is ready for review :)

src/amo/components/Search/index.js
```diff
@@ -120,10 +121,19 @@ export class SearchBase extends React.Component {
       <SearchSort filters={filters} pathname={pathname} />
     ) : null;
 
+    if (!queryParams.q) {
+      return (
+        <div className="Search">
+          <Card className="SearchContextCard">
+            <h1 className="SearchContextCard-header">Enter a search term and try again.</h1>
```
willdurand: Shouldn't we use `i18n` here?

Here is an example: https://github.com/mozilla/addons-frontend/blob/93548677fa52daabf950c6f947c89a5f6867429e/src/disco/components/Footer/index.js#L26

tsl143: Actually :)

tsl143:
Ping @muffinresearch

tofumatt:
@tsl143 Thanks for your patch but please refrain from pinging a user especially multiple times within a day. @muffinresearch is away on holiday for a week and the team is pretty busy, so we may take some time to get to reviewing a patch. 😄

If you just let us know something is ready for review we'll all see it and will get to it as soon as we can. Thanks!

src/amo/components/Search/index.js
```diff
@@ -120,10 +124,19 @@ export class SearchBase extends React.Component {
       <SearchSort filters={filters} pathname={pathname} />
     ) : null;
 
+    if (!queryParams.q) {
```
tofumatt: This should be using `filters`, not `queryParams`. It's a bit of a style/convention difference but we use `filters` everywhere internally as they are better-labelled. For instance, you can use `filters.query` here instead of `queryParams.q`, which is more readable and informative.

Additionally, this should account for an empty search (eg `''`), so the check should actually be `if (!filters.query || filters.query.length === 0) {`

src/amo/components/Search/index.js
```diff
@@ -120,10 +124,19 @@ export class SearchBase extends React.Component {
...
+      return (
+        <div className="Search">
+          <Card className="SearchContextCard">
+            <h1 className="SearchContextCard-header">{i18n.gettext('Enter a search term and try again.')}</h1>
```
tofumatt: This line looks quite long, can you move the content into its own line eg:

```js
<h1 className="SearchContextCard-header">
  {i18n.gettext('Enter a search term and try again.')}
</h1>
```

src/amo/components/Search/index.js
```diff
@@ -120,10 +124,19 @@ export class SearchBase extends React.Component {
...
+          </Card>
+        </div>
+      );
+    }
+
     return (
       <div className="Search">
         {errorHandler.renderErrorIfPresent()}
-
```
tofumatt: Why was this removed? The whitespace here was used to divide up different sections for readability.

src/amo/components/Search/index.js
```diff
@@ -152,6 +165,7 @@ export function mapStateToProps(state) {
 }
 
 export default compose(
+  translate({ withRef: true }),
```
tofumatt: I don't think you'll actually need the `withRef: true` here; we use it elsewhere but I think that was for an old testing thing or something? Try leaving it out and if it still works it's fine.

tests/unit/amo/components/TestSearch.js
```diff
@@ -153,6 +159,12 @@ describe('Search', () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
+  it('should render "Enter a search term and try again." msg if no search term', () => {
+    const root = mountRender({ filters: { query: null } });
```
tofumatt: Why does this need to use `mount` over `shallow`?

tsl143:
Ah, my bad, I need to be more patient :P

tests/unit/amo/components/TestSearch.js
```diff
@@ -153,6 +159,12 @@ describe('Search', () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
+  it('should render "Enter a search term and try again." msg if no search term', () => {
```
tofumatt: The spec doesn't need the exact error text; `it('should render an error message on empty search', () => {` would be better.

tests/unit/amo/components/TestSearchResults.js
```diff
@@ -48,7 +42,6 @@ describe('<SearchResults />', () => {
     const root = renderResults({ filters: {} });
     const addonsCard = findRenderedComponentWithType(root, AddonsCard);
 
-    expect(root.message.textContent).toContain('enter a search term');
```
tofumatt: I think you can change this test's spec name... this component won't render an error anymore (that's good!) but just renders an empty result set. Maybe changing it to "renders no results when there are no search filters" would work?

tests/unit/amo/components/TestSearch.js
```diff
@@ -159,10 +154,10 @@ describe('Search', () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
-  it('should render "Enter a search term and try again." msg if no search term', () => {
-    const root = mountRender({ filters: { query: null } });
+  it('should render an error message on empty search', () => {
+    const root = render({ filters: { query: null } });
 
-    expect(root.find(Card)).toHaveText('Enter a search term and try again.');
+    expect(root.find('h1.SearchContextCard-header')).toHaveText('Enter a search term and try again.');
```
tofumatt: You can just target the class without the `h1` tag; that's usually what we do and it means we can always change the tag name later 😄

src/amo/components/Search/index.js
```diff
@@ -120,6 +124,18 @@ export class SearchBase extends React.Component {
       <SearchSort filters={filters} pathname={pathname} />
     ) : null;
 
+    if (!filters.query || filters.query.length === 0) {
```
tofumatt: Shoot, I just realised this means a valid search without a query will error.

You should add a test for a search with `filters: { category: 'some-category', query: null }` and make sure it still renders results. That way we won't get caught out by this bug again. I guess really this should really just use `!hasSearchFilters(filters)` from `core/searchUtils`;

tsl143:
Done :)

src/amo/components/Search/index.js
tsl143: Are we planning category wise search?
Because for now we get addons from categories by https://addons.mozilla.org/en-US/firefox/extensions/photos-music-videos/

tofumatt:
Category is a valid search param and the Categories component uses the search component to get add-ons with a particular category value. This Search component is a more generic component used to render all kinds of searches (as that’s the API we’re always using in the background).

tsl143:
In that case I need to rework on approach. As the search parameters have
addon_type and hence will fail purpose of this patch.
I mean if we are on extensions tab and go for blank search,  it will still
give results.

On Aug 22, 2017 6:46 PM, "tofumatt ☕️" <notifications@github.com> wrote:

Category is a valid search param and the Categories component uses the
search component to get add-ons with a particular category value. This
Search component is a more generic component used to render all kinds of
searches (as that’s the API we’re always using in the background).

—
You are receiving this because you were mentioned.
<https://github.com/mozilla/addons-frontend/pull/2703#issuecomment-324022718>,
or mute the thread

tsl143:
I was commuting so wasnt able to completely describe the problem.
So if I put `(!hasSearchFilters(filters))` instead of `(!hasSearchFilters(filters).query || filters.query.length === 0)`

Following are the results
screen shot 2017-08-22 at 8 49 53 pm
screen shot 2017-08-22 at 8 50 08 pm

I need to do some changes.

tsl143:
Hi,
So I have put the search term text check only when `pathname = '/search/'` so as to restrict that check only if searchbar is used.

src/amo/components/Search/index.js
```diff
@@ -120,6 +124,18 @@ export class SearchBase extends React.Component {
       <SearchSort filters={filters} pathname={pathname} />
     ) : null;
 
+    if (pathname === '/search/' && (!filters.query || filters.query.length === 0)) {
```
tofumatt: While this approach will technically work it's fragile (what if we add another search URL or change it?) and it also seems to be untested.

The best approach would be to have this check in the `SearchPage` component because it's the only place we currently care about having a search query.

tsl143:
Done, and yes via SearchPage its cleaner and effective solution.

tsl143:
@tofumatt sorry for pinging again but its almost a week.

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -23,4 +32,16 @@ describe(__filename, () => {
...
     });
   });
+
+  it('should render Search results on search with query', () => {
+    const root = render({ filters: { query: 'burger' } });
+
+    expect(root.find(Search)).toHaveLength(1);
+  });
+
+  it('should render an error message on empty search', () => {
```
tofumatt: There should be an extra test here for a search that's not `null` but is an empty string, eg: `const root = render({ filters: { query: '' } });`

tests/unit/amo/components/TestSearch.js
```diff
@@ -153,6 +153,12 @@ describe('Search', () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
+  it('should render category results', () => {
```
tofumatt: I would add "even with empty query".

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -1,9 +1,18 @@
...
+import { SearchPageBase, mapStateToProps } from 'amo/components/SearchPage';
+import Search from 'amo/components/Search';
 import { CLIENT_APP_ANDROID } from 'core/constants';
 import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
+import { getFakeI18nInst } from 'tests/unit/helpers';
 
 
 describe(__filename, () => {
+  function render(props = {}) {
+    return shallow(<SearchPageBase {...{ ...props, i18n: getFakeI18nInst() }} />);
```
tofumatt: This should use `shallowUntilTarget(<SearchPage />, SearchPageBase)` instead so we can test the composed component. `shallowUntilTarget` is in `tests/unit/helpers`.

src/amo/components/SearchPage/index.js
```diff
@@ -1,18 +1,45 @@
...
+          <Card className="SearchContextCard">
+            <h1 className="SearchContextCard-header">
+              {i18n.gettext('Enter a search term and try again.')}
+            </h1>
+          </Card>
+        </div>
+      );
+    }
+
+    return <Search filters={filters} />;
```
tofumatt: I was surprised to see all we needed to pass were filters! Nice and tidy 😄

tsl143:
@tofumatt I removed following test cases as now the mentioned `props` are not included and also we have tests for `seacrhFilters` and `pagination` in https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestSearch.js#L55

`doesn't duplicate the clientApp in the URL in the queryParams`
`sets the paginationQueryParams from filters`
`enables search filters`

But I think, this affected the code coverage, I feel stuck.

tsl143:
I realized, I shouldn't remove `doesn't duplicate the clientApp in the URL in the queryParams` test, so I modified it and coverage test passed :)

tsl143:
@tofumatt , anything remaining from my side on this?

tofumatt:
Sorry, but turns out the code as-is causes #3096 so this will need a bit more work. I'm re-opening #2702 because of that.

PR #2708, opened 2017-07-08
fix(package): update redux-logger to version 3.0.6

tofumatt:
Closes #2209.

Just changed `import createLogger from 'react-redux';` to `import { createLogger } from 'react-redux';` as https://github.com/evgenyrodionov/redux-logger/releases/tag/3.0.0 recommends. Tested and it passes tests and logs in browser 👍


tofumatt:
Dammit, I forgot to push my changes and this is causing an error. Sorry 😞

PR #2710, opened 2017-07-08
fix(package): update jsdom to version 11.1.0

tofumatt:
Closes #2303


tofumatt:
r? This is just the new JSDom API to get a `window` as of v10, we could use the old one but it's an easy upgrade.

PR #2720, opened 2017-07-09
 Add-on with long name is causing layout issues #2415

### LINKED ISSUE #2415: Add-on with long name is causing layout issues
Steps to reproduce:
1. Submit an add-on with long name i.e. https://addons-dev.allizom.org/en-US/android/addon/lastpass-password-manager/?src=search (without spaces)
2. Access its details page

Expected results:

Actual results
Add-on name is not truncated or rendered as expected.

Notes/Issues:
- add-on details page: long name causes horizontal scrolling
- listing pages: icon is moved from its place

Verified on FF53(Android 6.0.1). Issue is reproducing on AMO-dev and stage.
Screenshots for this issue:
untitled

tsl143:
Fix #2415 (the review list part)

**Before**
screen shot 2017-07-09 at 08 28 36

**After**
screen shot 2017-07-09 at 08 32 55

### DIFF
--- a/src/amo/css/AddonReviewList.scss
+++ b/src/amo/css/AddonReviewList.scss
@@ -14,6 +14,7 @@
   align-items: center;
 
   .AddonReviewList-header-text {
+    //Flex implies min-width: auto, so the text overflows, to over come that change min-width.
     min-width: 1%;
 
     > h2,
--- a/src/amo/css/AddonReviewList.scss
+++ b/src/amo/css/AddonReviewList.scss
@@ -34,10 +34,10 @@
     a:visited,
     a:hover,
     a:active {
-      text-decoration: none;
       color: $primary-font-color;
       font-weight: normal;
-      word-wrap: break-word;
+      overflow-wrap: break-word;
+      text-decoration: none;
     }
   }
 
@@ -48,7 +48,7 @@
 }
 
 .AddonReviewList-li {
-  word-wrap: break-word;
+  overflow-wrap: break-word;
 
   > h3 {
     padding: 0;
--- a/src/amo/css/AddonReviewList.scss
+++ b/src/amo/css/AddonReviewList.scss
@@ -14,6 +14,8 @@
   align-items: center;
 
   .AddonReviewList-header-text {
+    min-width: 1%;
+
     > h2,
     > h3 {
       font-size: 1.1em;
@@ -35,6 +37,7 @@
       text-decoration: none;
       color: $primary-font-color;
       font-weight: normal;
+      word-wrap: break-word;
     }
   }
 


src/amo/css/AddonReviewList.scss
```diff
@@ -14,6 +14,8 @@
   align-items: center;
 
   .AddonReviewList-header-text {
+    min-width: 1%;
```
tofumatt: Odd, why is this needed?

src/amo/css/AddonReviewList.scss
```diff
@@ -35,6 +37,7 @@
       text-decoration: none;
       color: $primary-font-color;
       font-weight: normal;
+      word-wrap: break-word;
```
tofumatt: Could you change this to `overflow-wrap`? I feel like we might be using `word-wrap` elsewhere but we shouldn't 😅 .

This is a bit of a nitpick as [`word-wrap` aliases to `overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) but it would be nice to use the "proper" attribute. 😄

tsl143:
For min-width:
Flex implies min-width: auto, so the width overflows, to over come that we need to change min-width to 1%.
https://drafts.csswg.org/css-flexbox-1/#min-size-auto

https://stackoverflow.com/questions/12022288/how-to-keep-a-flex-item-from-overflowing-due-to-its-text

tofumatt:
Ah right, makes sense. Could you add a comment there, just explaining that? Afterward I'll get this merged, thanks :-)

Matthew Riley MacPherson

muffinresearch:
@tofumatt if we use `overflow-wrap` does `word-wrap` get added by postcss? Just looks like `word-wrap` still has wider browser support? http://caniuse.com/#feat=wordwrap (Partial support means only word-wrap is supported).

muffinresearch:
Fwiw: we're using word-wrap quite a bit already too: https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=word-wrap&type=

tsl143:
@tofumatt done.

tofumatt:
I'd like to change them all over... I'd think postcss would convert for us but that's a good question. @tsl143 does the generated css output include word-wrap when you look in the devtools?

tsl143:
@tofumatt Compiled CSS has `overflow-wrap: break-word;`

tofumatt:
Shoot, okay, I guess we should actually use `word-wrap` then. Sorry @tsl143, can you change it back? I am dumb! 😭

tsl143:
@tofumatt Done.

Issue #2704, opened 2017-07-07
Discopane UI tests are returning 1 and causing CI to report failures


tofumatt:
Travis is seeing `TypeError: Cannot set property 'getDocsInfo' of undefined` and can't run `yarn`.

I'm having a look, others seem to be reporting a similar error.

tofumatt:
As [others reported](https://github.com/yarnpkg/yarn/issues/3851#issuecomment-313703599), running it again meant `yarn` worked.

The discopane tests are still failing, returning `1` when they pass.

kumar303:
I'm on PTO today but yesterday I contacted TravisCI support because it doesn't make sense that our `docker logs` command is failing (at least I think that's what's failing). Also, I noticed that the `travis-build` revision got bumped when our job started failing.

If anyone wants to try this before I'm back, they suggested the following:

"A possible close approximation of the issue might be to run it locally with our Docker images, as described here: https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally

It is also possible to turn on the debug feature for this repository, so that you can have access to a running VM https://docs.travis-ci.com/user/running-build-in-debug-mode/ "

BTW, it seems unlikely that their build deployment caused the error although it's still possible. They sent me this diff of the release which they said mostly contained housekeeping changes: https://github.com/travis-ci/travis-build/compare/75a27e5e4...6094b6ae8

If no one gets to it before me I'll try in debug mode and try to see if `docker logs` is really the culprit or not.

tofumatt:
I tried doing this but I'll admit I'm a bit unfamiliar with exactly how to run the UI tests from a checkout. I tried building the docker image inside the travis docker image, but running the test-ci command had the travis VM complain that it couldn't find a running docker :-/

I'll try again later when my brain is less hazy.

tofumatt:
Oh I think I'm dumb and probably needed to enable the right config... having another look now. Sorry, I've never run travis VMs/dockers locally before!

tofumatt:
I couldn't quite figure out how to get things set up locally. The Travis docker images aren't, as I understand it, what we're running on because we ourselves use docker. Docker isn't running on those images and I don't get how to start it... but even so it feels like it's not the same process as what's happening on Travis so I didn't keep exploring. 😢

kumar303:
I'll take a look on Monday

tofumatt:
[...]

PR #2722, opened 2017-07-09
fix: Disable docker log for now to fix tests

tofumatt:
Trying this to see what happens, because these constantly "failing" tests are annoying and a bad thing–they're gonna make me think something is fine when it isn't soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
 - if [ $GUI ]; then export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start && sleep 3; fi
 script:
 - if [[ $TOXENV == 'discopane-ui-tests' ]]; then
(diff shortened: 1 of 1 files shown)


tofumatt:
This prevents the tests from running `docker logs` on Travis which seems to cause an error.

This is a stop-gap but as you can see from the test output it still runs the tests (successfully) so I think this is a good temporary fix to stop all of ours builds from constantly stating they've (falsely) failed to build.

tofumatt:
If someone from QA could let us know why the `docker logs` command is even needed that'd be great... the test output seems to still be here and the tests are run, so I'm not sure what they're for...

kumar303:
`docker logs` is pretty essential since it helps us figure out what's going on (server side) when the py.test suite fails. I'll work on fixing it in https://github.com/mozilla/addons-frontend/issues/2704

PR #2727, opened 2017-07-10
Show Docker logs again in functional tests

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2704


kumar303:
Even though the tests mysteriously fixed themselves, I researched this and could see some legitimate causes for failure. It sucks to replace one line of piped bash commands with 200 lines of JS but I think it helps buy us many things:

* The UI tests should be faster now since they no longer sleep for a minute at startup
* The UI tests should be more reliable -- sleeping for a minute might not have been enough when Travis was really busy
* I think some of the missing command failure output will be solved now
* You can now run the UI tests locally (this wasn't so easy before on Mac OS X)
* Docker gets a compatible log driver by default which is my theory for what was causing the failure (without this fix we were at the mercy of system defaults which could vary between Travis machines)

bin/start-func-test-server.js
```diff
@@ -0,0 +1,191 @@
...
+const path = require('path');
+const childProcess = require('child_process');
+
+const root = path.join(__dirname, '..');
+console.log(`Working directory: ${root}`);
+const containerIdFile = path.join(root, 'docker-container-id.txt');
+
+// TODO: make these configurable when we need to start
+// servers for multiple apps.
+const appInstance = 'disco';
```
tofumatt: Using `process.env.NODE_APP_INSTANCE` would be a pretty easy fix here I think and buy us most of that. I think it's already passed to the tests on travis... if not you could easily add it to the env vars.

bin/start-func-test-server.js
```diff
@@ -0,0 +1,191 @@
...
+
+function exec(cmd, argParts, { quiet = false } = {}) {
+  const cmdString = `${cmd} ${argParts.join(' ')}`;
+  if (!quiet) {
+    logDivider('BEGIN exec');
+    console.log(`Exec: ${cmdString}`);
+  }
+  return new Promise((resolve, reject) => {
+    childProcess.exec(cmdString, { cwd: root }, (error, stdout, stderr) => {
+      let gotOutput = false;
```
tofumatt: Personally I found `gotOutput` a bit of an odd variable name to grok at first. Maybe `receivedOutput` would be better?

bin/start-func-test-server.js
kumar303: Heh, I thought so too but passing env vars to commands in [.travis.yml](https://github.com/kumar303/addons-frontend/blob/4eea1340c0ec9d3e398c47aded50a2e32367d741/.travis.yml#L30) is not straight forward. I think it's best to deal with it later. I think we may have to create an `amo` and a `disco` `package.json` script.

tofumatt: Ah okay, no problem.

PR #2728, opened 2017-07-10
fix(package): update react-onclickoutside to version 6.1.1

tofumatt:
Closes #2525


tofumatt:
The [upgrade to v6](https://github.com/Pomax/react-onclickoutside/compare/v5.11.1...v6.0.0) was around using `React.createClass`, so after tests pass and testing this locally we're good to merge.

Issue #2742, opened 2017-07-11
Allow search results to be loaded without add-ons


kumar303:
This is needed to show the [featured add-ons page](https://github.com/mozilla/addons-frontend/issues/2620) before it has fully loaded with data.

PR #2743, opened 2017-07-11
Allow search results to load without add-on data

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2742

This patch doesn't apply it anywhere but here's an example of what's possible:

search-results-loading mov

### DIFF
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -33,7 +33,7 @@ export default class AddonsCard extends React.Component {
         {addons && addons.length ? (
           <ul className="AddonsCard-list">
             {addons.map((addon) => (
-              <SearchResult addon={null} key={addon.slug} />
+              <SearchResult addon={addon} key={addon.slug} />
             ))}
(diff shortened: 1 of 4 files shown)


kumar303:
@tofumatt I figured you might also need this for search data so I pulled it out into a separate patch.

kumar303:
To test it locally you'd have to change some code like `<SearchResult addon={addon} />` to `<SearchResult addon={null} />`.

tofumatt:
Omg, awesome!

kumar303:
Should the margin only be applied for the loading case? Or for all cases?

kumar303:
Oh wait, nevermind, you have `.LoadingText` in there. I see.

kumar303:
Yeah, that looks better. Let me know what you think:

screenshot 2017-07-11 17 19 14

tofumatt:
Yeah, when the actual content is there the right margins are applied, because there's a header tag.

tofumatt:
That's great 👍

Issue #2785, opened 2017-07-12
Google Analytics is used to track users.


alsternerd:
### Describe the problem and steps to reproduce it:
Google Analytics is used on the about:addons site while browsing not installed AddOns.

### What happened?
It's tracking users and giving data to google directly without any consent.

<hr>

**EDIT** from @kumar303: Mozilla has a [legal contract with Google](https://bugzilla.mozilla.org/show_bug.cgi?id=697436#c14) that prevents them from using our Google Analytics data for mining or from sharing it with third parties, among other privacy-protecting provisions.

<hr>

### What did you expect to happen?
No tracking anywhere in a browser that advertises with privacy.

### Anything else we should know?
Yes, remove all tracking everywhere. (General Bug)

tofumatt:
As I mentioned in #1107: we *will not* be removing analytics support entirely. It is extremely useful to us and we have already weighed the cost/benefit of using tracking.

*That said*, I think we should only be enabling tracking for users who have opted-in to sharing their data usage in Firefox. I'd guess respecting the Telemetry checkbox (https://www.dropbox.com/s/blcumfu2vzrvt9f/Screenshot%202017-07-12%2014.13.00.png?dl=0) would work. Or maybe having a new checkbox if piggy-backing on that one is problematic from a legal perspective.

tofumatt:
Actually, @muffinresearch pointed out we could probably just observe Do Not Track here, because this pane is actually a web page loaded in an iFrame inside the browser page. That might be faster to ship.

I'm definitely for giving users the option to disable this.

tofumatt:
Wanted to address your position though:

We don't give the "data directly to Google". See the discussion here: https://bugzilla.mozilla.org/show_bug.cgi?id=858839. The short version is:

> **tl;dr:** We now have an option to opt-out of Google doing anything with the data that Google Analytics collections on Mozilla websites. GA tracking is anonymous and at the aggregate level and we use it to improve the experience of our websites.

We are collecting aggregate and non-identifiable data in numbers to ensure our development/UX changes are met well. We can respect privacy and still have analytics; in fact Mozilla's aim is for an experience that values user privacy **and** usability (I'd say Apple also wants UX that fits that mold, as an example). We need some data, anonymised and aggregated, to do this.

dertuxmalwieder:
> We now have an option to opt-out

Privacy: Option to opt-in.

Anti-privacy: Option to opt-out.

sleeksorrow:
[...]

PR #2787, opened 2017-07-12
Respect Do Not Track

MaxLeiter:
Ref: https://github.com/mozilla/addons-frontend/issues/2785

### DIFF
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -79,7 +79,7 @@ export function getPageProps({ noScriptStyles = '', store, req, res }) {
     noScriptStyles,
     sriData,
     store,
-    trackingEnabled: convertBoolean(config.get('trackingEnabled')),
+    trackingEnabled: convertBoolean(config.get('trackingEnabled')) && convertBoolean(req.app.get('dnt')),
   };
 }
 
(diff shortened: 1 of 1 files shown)


ianb:
Typically Mozilla properties use this DNT function: https://github.com/schalkneethling/dnt-helper/blob/master/js/dnt-helper.js

(Mostly it's that IE defaulted it on, so we don't believe that represents user intent on that browser)

MaxLeiter:
This page is only accessible from Firefox. No need for additional bloat.

MaxLeiter:
Good idea, @tofumatt. Will do

MaxLeiter:
Not really in state, but pretty simple solution. I don't have time to dig around the codebase and figure out a more complex solution.

kumar303:
Google Analytics (in the way [Mozilla has legally contracted Google to operate it](https://bugzilla.mozilla.org/show_bug.cgi?id=697436#c14)) does not count as the T in DNT [as defined by the EFF](https://www.eff.org/deeplinks/2011/02/what-does-track-do-not-track-mean) so I'm not sure this is the right thing to do.

tofumatt:
Yeah... I threw out the idea to use DNT as a way to not load GA but it seems the Tracking Protection pref is the more appropriate thing to use.

That will require more work but I'd say Mozilla's position is GA is not to be blocked by DNT. I was just throwing out ideas, seems my DNT one was wrong.

tofumatt:
So in fact I was wrong and we _should not_ be using GA if DNT is on. I'll close this as I think the AMO team will investigate this and handle it, but thanks a lot for your quick patch @MaxLeiter!

PR #2790, opened 2017-07-12
Make sure things run on a secure version of Node.js

Pomax:
adds an "engine" section so that Node can be pegged to a secure (LTS) version - see https://nodejs.org/en/blog/vulnerability/july-2017-security-releases/

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
   "description": "Universal front-end projects to complement addons-server.",
   "main": "index.js",
   "private": true,
+  "engines": {
+    "node": "^6.11.1"
+  },
   "scripts": {
     "build": "bin/build-checks.js && better-npm-run build",
     "build-locales": "bin/build-locales",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "private": true,
   "engines": {
-    "node": "^6.11.1"
+    "node": "6"
   },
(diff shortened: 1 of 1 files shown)


tofumatt:
We should update our NVMRC while we're at it; I'll add it to this patch.

package.json
```diff
@@ -4,6 +4,9 @@
   "description": "Universal front-end projects to complement addons-server.",
   "main": "index.js",
   "private": true,
+  "engines": {
+    "node": "^6.11.1"
```
kumar303: Can this just be `"node": "6"`?  Otherwise, I supposed we'd have to update it anytime we move to a new minor release of Node 6.

kumar303:
Thanks @Pomax . I couldn't find anything in https://nodejs.org/en/blog/vulnerability/july-2017-security-releases/ in relation to how the `engine` field is used for security purposes. Can you elaborate on what I'm missing?

package.json
Pomax: Note the `^` which marks it as "any version of 6, but at least 6.11.1"

Pomax:
@kumar303 it's a signal to devs (npm will complain if it sees the wrong version of node) and is also used by various deploy systems such as Heroku to determine the minimum acceptable version of Node.js to use.

Pomax:
@kumar303 updated the PR

PR #2794, opened 2017-07-13
feat: Disable tracking/Google Analytics when DNT is enabled

### LINKED ISSUE #2792: Do not serve GA assets when DNT is enabled
It is Mozilla's website policy to not serve Google Analytics when DNT is enabled in a browser. We should check for a DNT header and not serve Google Analytics if one is present (or at least not run them, I'm not sure how we include/bundle them right now).

This issue will disable tracking for both AMO and Disco Pane (and any other site that uses our `Tracking` class).

tofumatt:
Fix #2792.

This patch disables the GA script from being included and sets Tracking as being off when DNT is set. It respects this on both the server and client.

Works for both AMO and the Disco Pane, so that's cool 👍 

Tested locally with tracking set to enabled and DNT on/off and things worked as expected.

One review for this should be fine, whoever can get to it first 😄

### DIFF
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -1,5 +1,6 @@
-/* global window */
+/* global navigator, window */
 /* eslint-disable no-underscore-dangle */
+import { oneLine } from 'common-tags';
 import config from 'config';
 
 import { convertBoolean } from 'core/utils';
@@ -13,6 +14,30 @@ import {
 } from 'core/constants';
 
 
+export function isDoNotTrackEnabled({
+  _log = log,
+  _navigator = typeof navigator !== 'undefined' ? navigator : null,
+  _window = typeof window !== 'undefined' ? window : null,
+} = {}) {
+  if (!_navigator || !_window) {
+    return false;
+  }
+
+  // We ignore things like `msDoNotTrack` because they are for older,
+  // unsupported browsers and don't really respect the DNT spec. This
+  // covers new versions of IE/Edge, Firefox from 32+, Chrome, Safari, and
+  // any browsers built on these stacks (Chromium, Tor Browser, etc.).
+  const dnt = _navigator.doNotTrack || _window.doNotTrack;
+  if (dnt === '1') {
+    _log.log(oneLine`[TRACKING]: Do Not Track Enabled; Google Analytics not
+      loaded and tracking disabled.`);
+    return true;
+  }
+
+  // Known DNT values not set, so we will assume it's off.
+  return false;
+}
+
 export class Tracking {
 
   constructor({ trackingId, trackingEnabled, trackingSendInitPageView, _log = log } = {}) {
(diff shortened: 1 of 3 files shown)


src/core/tracking.js
```diff
@@ -105,8 +105,23 @@ export function getAction(type) {
   }[type] || TRACKING_TYPE_INVALID;
 }
 
+export function doNotTrackEnabled({
```
muffinresearch: I think generally we're prefixing boolean vars and functions with with `is` so this could be `isDoNotTrackEnabled`  &lt;/nitpick&gt;

tests/unit/core/test_tracking.js
```diff
@@ -127,3 +127,43 @@ describe('getAction', () => {
     expect(getAction('whatever')).toEqual('invalid');
   });
 });
+
+describe('Do Not Track', () => {
+  it('should respect DNT when enabled', () => {
+    expect(doNotTrackEnabled({
+      _navigator: { doNotTrack: '1' },
+      _window: {},
+    })).toBeTruthy();
```
muffinresearch: All the use of `toBeTruthy` / `toBeFalsy` could be `toBe(true)` and `toBe(false)`.

src/core/tracking.js
```diff
@@ -105,8 +105,23 @@ export function getAction(type) {
...
+  _navigator = typeof navigator !== 'undefined' ? navigator : null,
+  _window = typeof window !== 'undefined' ? window : null,
+} = {}) {
+  if (!_navigator || !_window) {
+    return false;
+  }
+
+  const dnt = _navigator.doNotTrack || _window.doNotTrack;
```
muffinresearch: IE 9/10 uses `navigator.msDoNotTrack` - it might be nice to add that for completeness.

src/core/tracking.js
```diff
@@ -105,8 +105,23 @@ export function getAction(type) {
...
+  return dnt === '1';
```
muffinresearch: Is this guaranteed to be a string on all browsers?

src/core/tracking.js
muffinresearch: It would be handy if this function logged the result.

tofumatt: Good call!

src/core/tracking.js
tofumatt: Yeah, but those versions don't really respect the spec either and enable it by default. I've stuck with newer browsers who use DNT as opt-in. They're also older versions than we generally target on AMO even (for CSS we target the past two versions).

tests/unit/core/test_tracking.js
tofumatt: Given the actual code checks for truthiness over `=== true` I thought this was fine, but that's true. I'll change it 😄

src/core/tracking.js
tofumatt: Good call. The tracking code logs whether or not GA is on but this is separate and worth logging ("Why is `trackingEnabled` set to `true` but I have no logging?!")

src/core/tracking.js
tofumatt: All the docs indicate as much, and my testing in Safari, Chrome, and Firefox say yes https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack#Browser_compatibility

Gonna test Edge later as my Windows PC is doing an update dance. 😓

tofumatt:
All set for another review, issues addressed and added a comment about why we're ignoring old IE. 😄

PR #2799, opened 2017-07-13
fix: Check for DNT correctly

### LINKED ISSUE #2797: Add missing tests for DNT functionality in Tracking object
As a follow up to https://github.com/mozilla/addons-frontend/issues/2792 we need tests for the `Tracking` object to make sure it gets [disabled when DNT is enabled](https://github.com/mozilla/addons-frontend/commit/1312536f6e1b075c508bccb2de7f050876eba172#diff-d174de3c0985bc0d979157795f96ae2bR50).

For QA, here's how to check for regressions:
* Load AMO in a tab and in the Discovery Pane
* Make sure Google Analytics gets loaded when DNT (Do Not Track) is disabled
* Make sure Google Analytics does not get loaded when DNT is enabled

### LINKED ISSUE #2798: Google Analytics is disabled for everyone

Load the site after https://github.com/mozilla/addons-frontend/issues/2792

### What happened?

Google Analytics is never loaded

We should load GA for anyone who doesn't have DNT enabled.

This logic is backwards: https://github.com/mozilla/addons-frontend/commit/1312536f6e1b075c508bccb2de7f050876eba172#diff-d174de3c0985bc0d979157795f96ae2bR134 It should be `!isDoNotTrackEnabled()` otherwise it will never be enabled for either case.

However, I recommend removing isDoNotTrackEnabled() from that line altogether since it's redundant (and makes the code more complicated). DNT is already checked in the constructor.

tofumatt:
Adds a test ensuring GA is disabled when DNT is true, as well.

Fixes #2798

### DIFF
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -39,15 +39,20 @@ export function isDoNotTrackEnabled({
 }
 
 export class Tracking {
-
-  constructor({ trackingId, trackingEnabled, trackingSendInitPageView, _log = log } = {}) {
+  constructor({
+    _isDoNotTrackEnabled = isDoNotTrackEnabled,
+    _log = log,
+    trackingEnabled,
+    trackingId,
+    trackingSendInitPageView,
+  } = {}) {
     if (typeof window === 'undefined') {
       /* istanbul ignore next */
       return;
     }
     this._log = _log;
     this.id = trackingId;
-    this.enabled = trackingEnabled && trackingId && !isDoNotTrackEnabled();
+    this.enabled = trackingEnabled && trackingId && !_isDoNotTrackEnabled();
     this.logPrefix = `[GA: ${this.enabled ? 'ON' : 'OFF'}]`;
     this.sendInitPageView = trackingSendInitPageView;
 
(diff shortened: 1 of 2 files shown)


src/core/tracking.js
```diff
@@ -131,10 +136,7 @@ export function getAction(type) {
 }
 
 export default new Tracking({
-  trackingEnabled: (
-    convertBoolean(config.get('trackingEnabled')) &&
-    isDoNotTrackEnabled()
-  ),
+  trackingEnabled: convertBoolean(config.get('trackingEnabled')),
```
kumar303: I don't think this variable is useful. It's redundant and makes the code more complicated and harder to test (e.g. there is no test coverage of the config reading functionality at all). Let's *not* fix it in this patch though :) because it will require updating the tests. I'll file a lower priority follow-up.

tests/unit/core/test_tracking.js
```diff
@@ -48,6 +48,18 @@ describe('Tracking', () => {
     ).toBe(true);
   });
 
+  it('should log OFF when not enabled due to Do Not Track', () => {
+    tracking = new Tracking({
+      _isDoNotTrackEnabled: () => true,
+      _log: {
+        info: sinon.stub(),
+      },
+      trackingEnabled: true,
```
kumar303: For example, it would be pretty hard to add coverage for the config logic here. Again, not important enough to block on.

tests/unit/core/test_tracking.js
```diff
@@ -48,6 +48,18 @@ describe('Tracking', () => {
...
+      trackingId: 'whatever',
+    });
+    sinon.assert.calledWith(tracking._log.info, sinon.match(/OFF/));
```
kumar303: This doesn't actually test if GA is enabled or not but I don't want to block the patch. I'll file for a fixup.

tofumatt:
Cool, thanks 👍

PR #2804, opened 2017-07-13
Fix up tests and add more logging for Tracking

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2797

* Adds better tests for things like config reading logic
* DRYs up the tests
* Adds better logging so we can see exactly how GA gets enabled/disabled

### DIFF
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -43,6 +43,7 @@ export class Tracking {
     _isDoNotTrackEnabled = isDoNotTrackEnabled,
   } = {}) {
     if (typeof window === 'undefined') {
+      /* istanbul ignore next */
       return;
     }
     this._log = log;
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -40,6 +40,7 @@ export function isDoNotTrackEnabled({
 
 export class Tracking {
   constructor({
+    _config = config,
     _isDoNotTrackEnabled = isDoNotTrackEnabled,
     _log = log,
     trackingEnabled,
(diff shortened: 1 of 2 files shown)


src/core/tracking.js
```diff
@@ -40,37 +39,43 @@ export function isDoNotTrackEnabled({
...
+      this.enabled = false;
+    } else if (!this.id) {
+      this.log('Disabled because trackingId was empty')
+      this.enabled = false;
+    } else if (_isDoNotTrackEnabled()) {
+      this.log(oneLine`Do Not Track Enabled; Google Analytics not
+        loaded and tracking disabled`);
+      this.enabled = false;
+    } else {
+      this.log('Google Analytics is enabled');
```
tofumatt: I wonder: should we log `"Enable Do Not Track to disable analytics on about:addons and addons.mozilla.org"`? Because there aren't any docs about this and it might be helpful.

kumar303: I dunno, I really don't want to imply that we are actually "tracking" people with GA because we're really not.

tofumatt: The only people who will ever see this are probably technical and those who see "Google Analytics" as something more nefarious than it is in our usage. I don't think we're implying tracking or encouraging them to opt-out, I'm just trying to give curious users who would actually hit this log output choice and save someone from filing a bug asking "how can I disable this?"

kumar303: Also, logs are disabled in production so I don't see how it would be so helpful

tofumatt: Hah, oh shoot. Good point! Alright, just ignore me then 😄

PR #2807, opened 2017-07-15
chore: Combine Categories and CategoriesHeader

### LINKED ISSUE #2653: Merge Categories and CategoriesHeader
#2649 merged a component that is quite similar to the `Categories` component because there were a few differences and it was nice to land. But really we should get them to be the same because it should be easy to reuse them.

tofumatt:
Close #2653.

This consolidates CategoriesHeader and Categories into one component (it also moves some tests to enzyme).

Finally it pretties up the `/extensions/categories/` page a bit from what was there before (we still don't have mocks for it as it's a new/rarely-visited page, but I think this is definitely better)

### Before
screen shot 2017-07-19 at 22 41 39

### After
screen shot 2017-07-19 at 22 40 35


tests/unit/amo/components/TestCategories.js
```diff
@@ -1,160 +1,100 @@
...
 
-    expect(
-      root.querySelector('.Categories-list').textContent
-    ).toEqual('GamesTravel');
+    expect(root).toHaveClassName('Categories');
   });
 
-  it('renders loading when loading', () => {
-    const root = renderDomNode({
+  it('it renders loading text when loading', () => {
```
muffinresearch: Superfluous `it`

tests/unit/amo/components/TestCategories.js
```diff
@@ -1,160 +1,100 @@
...
     });
 
-    expect(root.textContent).toContain('Loading');
+    expect(root.find('.Categories-loading-info'))
+      .toIncludeText('Loading categories.');
   });
 
-  it('renders a message when there are no categories', () => {
-    const root = renderDomNode({
+  it('it renders LoadingText components when loading', () => {
```
muffinresearch: superfluous `it`

tests/unit/amo/components/TestCategories.js
```diff
@@ -1,160 +1,100 @@
...
     });
 
-    expect(root.textContent).toEqual('No categories found.');
+    expect(root.find('.Categories-loading-text').find(LoadingText))
+      .toHaveLength(8);
   });
 
-  it('renders an error', () => {
-    const root = renderDomNode({
+  it('it renders an error message if there was an error', () => {
```
muffinresearch: superfluous `it`

tests/unit/amo/components/TestCategories.js
```diff
@@ -1,160 +1,100 @@
...
-
-    const props = mapStateToProps(store.getState(), {
-      params: { visibleAddonType: 'extensions' },
-    });
 
-    expect(props).toEqual({
-      addonType: ADDON_TYPE_EXTENSION,
-      categories: {
-        Games: {
+  it('it renders categories if they exist', () => {
```
muffinresearch: superfluous `it`

tests/unit/amo/components/TestCategoriesPage.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+
+    expect(props).toEqual({
+      categories: {
+        [ADDON_TYPE_COMPLETE_THEME]: {},
+        [ADDON_TYPE_DICT]: {},
+        [ADDON_TYPE_EXTENSION]: {
+          Games: {
+            application: 'android',
+            name: 'Games',
```
muffinresearch: If the lang is dispatched is 'fr' should the localized data be reflected?

tests/unit/amo/components/TestCategories.js
```diff
@@ -1,160 +1,100 @@
...
     const fakeDispatch = sinon.stub();
-    const root = render({
-      addonType: ADDON_TYPE_EXTENSION,
-      dispatch: fakeDispatch,
-    });
 
-    sinon.assert.calledWith(
-      fakeDispatch, setViewContext(ADDON_TYPE_EXTENSION));
-
-    // Make sure that we update the addonType when `componentDidUpdate()`
```
muffinresearch: Was this cleaned up in this patch so that the comment + TODO doesn't need to be kept?

tofumatt: Yup, now we use enzyme's `setProps` which calls `componentDidUpdate()` as part of the React Component Lifecycle rather than us calling the method directly, which was what felt off about that code: https://github.com/mozilla/addons-frontend/pull/2807/files/118b437e8e627d07b27d72b0cdf222f6a14c95c1#diff-08275f1fe3a947e23a90f5ae53d3e3c3R63

tests/unit/amo/components/TestCategoriesPage.js
tofumatt: Actually that call didn't need to be there so I removed the setLang.

PR #2815, opened 2017-07-17
new system add-on

designakt:
via https://ftp.mozilla.org/pub/system-addons/tls13-middlebox/


designakt:
@muffinresearch can I consider your fix a positive review?

muffinresearch:
@designakt For future reference it's better to squash the commits rather than do a merge commit when there's multiple commits in a PR - see the drop-down in the merge button.

Issue #2806, opened 2017-07-14
Discovery Pane is still cached when DNT pref is changed


kumar303:
### Describe the problem and steps to reproduce it:

- In Firefox (Release or Nightly), start with a new profile. Alternatively, you can empty your cache and make sure DNT (Do Not Track) is disabled in about:preferences
- Load https://discovery.addons.mozilla.org/en-US/firefox/discovery/pane/54.0/Darwin/normal in a tab
- Using Devtools, check network tab and notice that Google Analytics is loaded. You'll see a request for https://www.google-analytics.com/analytics.js which comes from the script tag in the server HTML
- Go to about:preferences and enable DNT
- Go back to the tab with https://discovery.addons.mozilla.org/ , press enter in the URL bar (do not press refresh)

### What happened?

You will still see a request for https://www.google-analytics.com/analytics.js

### What did you expect to happen?

You should not see a request for https://www.google-analytics.com/analytics.js since DNT is enabled

### Anything else we should know?

We need the server to send Vary: DNT so that the browser doesn't use a cached response

tofumatt:
We want to do this for AMO as well, because we also don’t Vary on DNT there.

kumar303:
As far as I can tell, this is only needed for Discopane since that's the only place we tell the browser to cache the response so hard. However, I'm going to add it to all apps just in case.

kumar303:
Actually, no we do need it for AMO too, my bad

kumar303:
I verified that this patch fixed it, using the STR on https://discovery.addons-dev.allizom.org/en-US/firefox/discovery/pane/56.0a1/Darwin/normal

vcarciu:
Verified on Release and Nightly , -dev and Stage and everything looks fine. Marking bug as verified.

PR #2817, opened 2017-07-17
Tell the browser to vary its cache by DNT (Do Not Track)

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2806

To test this locally you'll need to simulate how the Discopane cache headers are set. You can do that by adding this to a local config:
```js
module.exports = {
  // Simulate Discopane's caching strategy.
  cacheAllResponsesFor: 60 * 60, // 1 hour
  trackingEnabled: true,
};
```


kumar303:
I started looking at adding test coverage to `server/base` for this but that patch would take a lot longer. Since adding Vary support is high priority, we should probably land before then.

src/core/server/base.js
```diff
@@ -199,6 +199,16 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
...
 
+    const cacheAllResponsesFor = config.get('cacheAllResponsesFor');
+    if (cacheAllResponsesFor) {
+      log.warn(oneLine`Sending a Cache-Control header so that the client caches
+        all requests for ${cacheAllResponsesFor} seconds`);
+      res.set('Cache-Control', `public, max-age=${cacheAllResponsesFor}`);
+    }
+
+    // Vary the cache on Do Not Track headers.
+    res.vary('DNT');
```
tofumatt: Usually when I see the DNT header written in lowercase, but I presume this doesn't really matter.

src/core/middleware/prefixMiddleware.js
```diff
@@ -109,14 +109,12 @@ export function prefixMiddleware(req, res, next, { _config = config } = {}) {
     // Collect vary headers to apply to the redirect
     // so we can make it cacheable.
     // TODO: Make the redirects cacheable by adding expires headers.
-    const varyHeaders = [];
     if (isLangFromHeader) {
-      varyHeaders.push('accept-language');
+      res.vary('accept-language');
```
tofumatt: 👍

src/core/server/base.js
kumar303: The browser requests the header in upper-case as `DNT: 1` but it shouldn't matter since I believe `res.vary()` accepts a case-insensitive header value. This is not documented though.

src/core/server/base.js
```diff
@@ -199,6 +199,16 @@ function baseServer(routes, createStore, { appInstanceName = appName } = {}) {
       webpackIsomorphicTools.refresh();
     }
 
+    const cacheAllResponsesFor = config.get('cacheAllResponsesFor');
+    if (cacheAllResponsesFor) {
```
muffinresearch: Should this have additional controls to prevent being used outside development?

src/core/middleware/prefixMiddleware.js
muffinresearch: This is so much better!

kumar303:
> Is -dev caching in the same way as prod/stage etc to enable this to be validated there?

Yep, I can use the STR to see the bug by visiting https://discovery.addons-dev.allizom.org/en-US/firefox/discovery/pane/56.0a1/Darwin/normal in a normal tab.

src/core/server/base.js
kumar303: good idea :)

tests/unit/core/server/testServer.js
```diff
@@ -0,0 +1,54 @@
...
+      return {
+        store: createStore(combineReducers({ reduxAsyncConnect })),
+        sagaMiddleware: null,
+      };
+    }
+
+    const app = baseServer(stubRoutes, createStoreAndSagas, {
+      appInstanceName: 'testapp',
+    });
+    return supertest(app);
```
muffinresearch: I wish there was a way to do this without causing a server to be run and listen on a port. But if the usage is minimal this should be a good way to get increased coverage of the server.

tests/unit/core/server/testServer.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import Helmet from 'react-helmet';
+import { Router, Route } from 'react-router';
+import { createStore, combineReducers } from 'redux';
+import { reducer as reduxAsyncConnect } from 'redux-connect';
+import supertest from 'supertest';
```
muffinresearch: Ah supertest, we meet again!

tests/unit/core/server/testServer.js
kumar303: yeah, me too. I looked high and low but couldn't find a non-server alternative. I guess no one does it? I was surprised though. I'm not totally thrilled with using the server but it seems pretty fast, I guess.

kumar303:
> Somehow this results in less coverage

Huh. It even says I increased it:
screenshot 2017-07-17 13 47 55

kumar303:
Hopefully 93% is still above the threshold. We definitely got more coverage here so I'm not sure what calculation is throwing things off.

PR #2819, opened 2017-07-17
feat: Add command to run dev:amo with staging data

### LINKED ISSUE #2535: Document CDN host to change when modifying API host during development
If you change your `API_HOST`, to get images to appear you also need to modify your `amoCDN` (see: https://github.com/mozilla/addons-frontend/issues/2528#issuecomment-306752163).

We should document how to do this and when you'd want to.

I actually feel like a cool option would just be `API_DATA=prod yarn dev:amo` and it would set all options needed to load another dataset locally (because sometimes docker is borked but oftentimes the docker data just isn't as full-featured). Thoughts on that idea?

tofumatt:
Close #2535.

Add some docs and a few convience commands to run the dev server with:

* dev/staging API data
* dev/staging CDN setting
* `CSP: false` so the CDN data is loaded

Given this is often how we develop locally anyway I figured these settings were alright.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,9 @@ Here are some commands you can run:
 | Command                     | Description |
 |-----------------------------|-------------|
 | yarn dev:amo                | Start the dev server and proxy (amo) |
+| yarn dev:amo:dev            | Start the dev server and proxy (amo) using data from the dev server (https://addons-dev.allizom.org/) |
 | yarn dev:amo:no-proxy       | Start the dev server without proxy (amo) |
+| yarn dev:amo:stage          | Start the dev server and proxy (amo) using data from the staging server (https://addons.allizom.org/) |
 | yarn dev:disco              | Start the dev server (discovery pane) |
 | yarn flow                   | Run Flow. By default this checks for errors and exits |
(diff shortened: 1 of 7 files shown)


tofumatt:
Has a test failure but my battery is dying. I'll look in a few hours once I'm home. 😢

tofumatt:
Okay... pretty much duped a test but I think it seems okay... let me know what you think.

README.md
```diff
@@ -43,7 +43,9 @@ Here are some commands you can run:
 | Command                     | Description |
 |-----------------------------|-------------|
 | yarn dev:amo                | Start the dev server and proxy (amo) |
+| yarn dev:amo:dev            | Start the dev server and proxy (amo) using data from the dev server (https://addons-dev.allizom.org/) |
```
tofumatt: I thought about this, but adding more ENV vars would then require the config to check for them and change things based on it. I'm okay with that approach but it's a bit less obvious/easy-to-document-in-the-README. It's also more code as here we're just piggy-backing on existing configs–I think even here with the extra CSP code/tests it's less work.

Issue #2833, opened 2017-07-19
Rename local dev commands for simplicity


kumar303:
I think `yarn dev:amo` and `yarn dev:disco` can just become `yarn amo` and `yarn disco` respectively because why not. Also, it will make `yarn amo:dev` (a local dev server connected to a hosted dev server) less confusing.

tofumatt:
Yes! I'm in total agreement. 👍

PR #2834, opened 2017-07-19
Simplify development server commads

kumar303:
Closes https://github.com/mozilla/addons-frontend/issues/2833

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Here are some commands you can run:
 | yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/) |
 | yarn amo:no-proxy           | Start the dev server without a proxy (for amo) using data from Docker |
 | yarn amo:stage              | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/) |
-| yarn dev:disco              | Start the dev server (for Discovery Pane) using data from Docker |
+| yarn disco                  | Start the dev server (for Discovery Pane) using data from the dev server (https://addons-dev.allizom.org/) |
 | yarn flow                   | Run Flow. By default this checks for errors and exits |
 | yarn flow:check             | Explicitly check for Flow errors and exit |
 | yarn flow:dev               | Continuously check for Flow errors |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -34,20 +34,20 @@ The easiest way to manage multiple node versions in development is to use
 ## Get started
 
 * type `yarn` to install all dependencies
-* type `yarn dev:amo:dev` to start a local dev server that connects to a
-  hosted development environment
+* type `yarn amo:stage` to start a local server that connects to a
```
kumar303: I think stage data is probably the best for newbies since it's cleaner and has icons (etc.)

README.md
```diff
@@ -34,20 +34,20 @@ The easiest way to manage multiple node versions in development is to use
...
-| yarn dev:amo                | Start the dev server/proxy (for amo) using data from Docker |
-| yarn dev:amo:dev            | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/) |
-| yarn dev:amo:no-proxy       | Start the dev server without a proxy (for amo) using data from Docker |
-| yarn dev:amo:stage          | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/) |
-| yarn dev:disco              | Start the dev server (for Discovery Pane) using data from Docker |
+| yarn amo                    | Start the dev server/proxy (for amo) using data from Docker |
+| yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/) |
+| yarn amo:no-proxy           | Start the dev server without a proxy (for amo) using data from Docker |
+| yarn amo:stage              | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/) |
+| yarn disco                  | Start the dev server (for Discovery Pane) using data from the dev server (https://addons-dev.allizom.org/) |
```
kumar303: In testing, I realized that disco still defaults to using dev data

README.md
tofumatt: Agreed, actually. Data on -dev is actually kind of a mess.

README.md
tofumatt: I feel like we should change that, but it can be a separate patch. I guess Disco data is pretty standard though.

kumar303: I think it's because we don't auto-generate featured Discopane add-ons in Docker so it would be a bit hard to set up

kumar303:
@muffinresearch @tsl143 @aayushsanghavi heads up that when you next pull from master you'll need to run `yarn amo` instead of `yarn dev:amo` to start a local server. Also note that it's easier to develop with stage data now (which is the cleanest). You can simply run `yarn amo:stage`.

Issue #2620, opened 2017-06-21
Convert featured add-ons to load data with a saga


kumar303:
Convert the featured add-ons component to load its data with a saga instead of async-connect.

Steps to test for regressions:
* Go to [more featured add-ons](https://addons-dev.allizom.org/en-US/android/extensions/featured/)
* Make sure the page still works

This is a step towards https://github.com/mozilla/addons-frontend/issues/2356

ValentinaPC:
Verified this on AMO-dev FF54(Win 7).
The link you provided returns a 502 Bad Gateway page.... as everything on dev (https://github.com/mozilla/addons-server/issues/6026).
Please see screenshot:
![2017-07-27_0954](https://user-images.githubusercontent.com/15685960/28657943-ca37dbb0-72b1-11e7-86ec-6425b8954743.png)

kumar303:
The dev server is back! Sorry about that. See https://github.com/mozilla/addons-server/issues/6026

ValentinaPC:
@kumar303 : I you sure the link - https://addons-dev.allizom.org/en-US/android/extensions/featured/ - is correct?
Shouldn't be: https://addons-dev.allizom.org/en-US/android/extensions/?sort=featured ?
Because accessing the your link returns a 404 page.
![404](https://user-images.githubusercontent.com/15685960/28705793-0b771dac-737a-11e7-885b-1c4e03d3dba4.png)

Thanks!

kumar303:
Yes, the link is correct: https://addons-dev.allizom.org/en-US/android/extensions/featured/

You need to first make sure you are "pinned" to the desktop site by way of a cookie. Since the desktop site is not live yet you need to view the mobile site first by clicking 'View mobile site' in the footer of https://addons-dev.allizom.org/en-US/firefox/ As long as you are using a desktop browser, it will show you the new desktop site. It's a bit confusing, sorry about that!

kumar303:
Actually, it would be better to verify this particular feature on an Android device since that is where we need to protect against regressions the most.

ValentinaPC:
Verified this on AMO-dev FF54(Win 7, Android 7.2.1)
Both https://addons-dev.allizom.org/en-US/android/extensions/featured/ and https://addons-dev.allizom.org/en-US/android/themes/featured/ are working as expected.
Tried also for different languages - works as expected.
[...]

PR #2838, opened 2017-07-19
Convert featured add-ons page to load with a saga

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2620

This mostly looks the same as the current featured add-ons pages but now it appears to load faster and also shows API errors. Examples:

screenshot 2017-07-19 17 04 15
screenshot 2017-07-19 17 03 49
screenshot 2017-07-19 16 26 41


src/amo/components/FeaturedAddons/index.js
```diff
@@ -38,19 +59,39 @@ export class FeaturedAddonsBase extends React.Component {
     }
   }
 
+  loadDataIfNeeded(nextProps = {}) {
+    const { addonType: lastAddonType } = this.props;
+    const {
+      errorHandler, dispatch, loading, results,
+    } = { ...this.props, ...nextProps };
+
+    const nextParams = nextProps && nextProps.params ?
```
tofumatt: I think this method could use with a few comments explaining, in plain English, what's happening. I can read through it but it took me awhile and a few re-reads to truly get what was happening. I think a comment above each section stating what it does and what it's checking for would help a lot.

src/amo/components/FeaturedAddons/index.js
```diff
@@ -38,19 +59,39 @@ export class FeaturedAddonsBase extends React.Component {
...
   render() {
-    const { addonType, hasSearchParams, loading, results } = this.props;
+    const { errorHandler, hasSearchParams, loading, results } = this.props;
 
     return (
       <div className="FeaturedAddons">
+        {errorHandler.hasError() ? errorHandler.renderError() : null}
         <h2 className="FeaturedAddons-header">
-          {this.headerForAddonType(addonType)}
+          {this.headerForAddonType()}
```
tofumatt: I was gonna ask about this eventually, but I know we do a fair amount of prop passing from `render()` methods to other component class methods. It makes testing easier... or I guess is how we'd setup methods prop injection for testing... but seeing you not use it here makes things cleaner.

I guess in the future we should just read props inside the method itself and not pass them around inside components as there's no need.

src/amo/sagas/featured.js
```diff
@@ -0,0 +1,41 @@
...
+import { createErrorHandler, getApi } from 'core/sagas/utils';
+
+
+export function* fetchFeaturedAddons(
+  { payload: { errorHandlerId, addonType } }
+) {
+  const errorHandler = createErrorHandler(errorHandlerId);
+  try {
+    yield put(showLoading());
+    const api = yield select(getApi);
```
tofumatt: I think it would be nice to give any `yield` statement some breathing room. I feel like this would be a bit clearer as:

```js
yield put(showLoading());

const filters = { addonType, page_size: FEATURED_ADDONS_TO_LOAD };
const response = yield call(featuredApi, { api, filters });

yield put(loadFeatured({
  addonType,
  entities: response.entities,
  result: response.result,
}));
```

I think of each `yield` as a sort of "section" of the saga and it's helpful to break them up.

tests/unit/amo/components/TestFeaturedAddons.js
```diff
@@ -6,59 +6,284 @@ import {
...
+
+    root.setProps({
+      // Update an unrelated parameter.
+      hasSearchParams: false,
+    });
+
+    sinon.assert.notCalled(dispatch);
+  });
+
+  it('does not fetch featured add-ons when results are loaded + empty', () => {
```
tofumatt: Just a note that these test descriptors are all really helpful and awesome–they describe each state very clearly and make reviewing these tests a breeze. Thanks! 👍

tests/unit/amo/components/TestFeaturedAddons.js
```diff
@@ -6,59 +6,284 @@ import {
...
+    const addonType = ADDON_TYPE_EXTENSION;
+    const dispatch = sinon.stub();
+
+    store.dispatch(_loadFeatured({ resultMap: {} }));
+    // This will make props.results === []
```
tofumatt: Thanks for the comment 👍

tests/unit/amo/components/TestFeaturedAddons.js
```diff
@@ -6,59 +6,284 @@ import {
...
-    expect(root.childAt(0)).toIncludeText('More Featured Themes');
+    expect(root.find('.FeaturedAddons-header'))
+      .toIncludeText('More Featured Themes');
+  });
+
+  it('renders a header for a configured addonType', () => {
+    const root = render({
+      addonType: ADDON_TYPE_THEME,
+      params: {
+        visibleAddonType: VISIBLE_ADDON_TYPES_MAPPING[ADDON_TYPE_EXTENSION],
```
tofumatt: This is fine but we do have a helper (`visibleAddonType`) that is probably worth using in case this ever becomes more complex than a mapping.

tests/unit/amo/reducers/testFeatured.js
```diff
@@ -20,7 +20,9 @@ describe('featured reducer', () => {
   describe('FEATURED_GET', () => {
     it('sets the initialState', () => {
       const { addonType, loading, results } = featured(
-        initialState, actions.getFeatured({ addonType: 'theme' }));
+        initialState, actions.getFeatured({
+          addonType: 'theme', errorHandlerId: 'some-error-handler',
```
tofumatt: This should use a constant rather than a hard-coded string, as I can't tell whether this should be an `ADDON_TYPE_COMPLETE_THEME` or `ADDON_TYPE_PERSONA` from the context.

tofumatt: (Though I see now this was probably my bad originally, very sorry.)

shame-cube

src/amo/components/FeaturedAddons/index.js
```diff
@@ -1,32 +1,54 @@
...
+    params: PropTypes.object.isRequired,
     results: PropTypes.array,
   }
 
   static defaultProps = {
     hasSearchParams: true,
   }
 
+  componentWillMount() {
+    this.props.dispatch(setViewContext(this.requestedAddonType()));
```
tofumatt: This is not a big deal but I'm so used to use doing `const { dispatch } = this.props;` that seeing `this.props.dispatch` really threw me off. I know it's an extra line or two do make the `const` declaration but for the particular task of getting component props it's so common we do `const { prop } = this.props;` that I'd prefer we always did it so I can tell what's a prop.

That said I did a quick search in the project and `this.props.foo` happens about a dozen times, so it's fine to merge, I guess I just never noticed it or saw components using it. 🤷‍♀️

tofumatt:
Also those screenshots look nice, cool to see the loading text in the featured section. 👍

kumar303:
Sorry, it wasn't ready for review yet (I forgot to put WIP). I need to push some more stuff but I'll read your comments first in case the new diff causes them to collapse.

tofumatt:
Shoot, sorry, I was trying to be helpful but was overzealous. I knew you were going on PTO soon so I figured I'd chime in early, but yeah I should have waited until you requested review.

Sorry about that! 😓

src/amo/components/FeaturedAddons/index.js
kumar303: Well, the old code had a bug because the `addonType` function parameter wasn't used at all. I wanted to convert this to Flow (which would have exposed it) but I didn't have time.

kumar303:
Thanks for the early review. It should be officially ready now but I haven't had a chance to address all comments. I need to catch a yoga class so I'll have to fix it up when I'm back from PTO.

tests/unit/amo/components/TestFeaturedAddons.js
```diff
@@ -2,68 +2,307 @@ import { shallow } from 'enzyme';
...
+    const dispatch = sinon.stub();
+    const customErrorHandler = new ErrorHandler({
+      id: 'custom-error-handler',
+      dispatch: sinon.stub(),
+    });
+    render({
+      dispatch,
+      errorHandler: customErrorHandler,
+      params: {
+        visibleAddonType: VISIBLE_ADDON_TYPES_MAPPING[addonType],
```
kumar303: Sweet. I didn't know we had that.

src/amo/components/AddonsCard/index.js
```diff
@@ -13,15 +13,45 @@ export default class AddonsCard extends React.Component {
...
+    const searchResults = [];
+
+    if (addons && addons.length) {
+      addons.forEach((addon) => {
+        searchResults.push(
+          <SearchResult addon={addon} key={addon.slug} />
+        );
+      });
+    } else if (loading) {
+      for (let count = 0; count < placeholderTotal; count++) {
```
kumar303: It's literally only one character shorter and it's pretty obscure so I don't think it's worth it:
```js
      Array(placeholderCount).fill(0).map((value, count) => {
        searchResults.push(
          <SearchResult key={count} />
        );
      });
```

vs.

```js      
      for (let count = 0; count < placeholderCount; count++) {
        searchResults.push(
          <SearchResult key={count} />
        );
      }
```

src/amo/components/FeaturedAddons/index.js
kumar303: Sure, I added some comments. This component is suffering from tracking URL state in two places which is why it's a bit confusing. It could be simplified once we track URL state purely in redux: https://github.com/mozilla/addons-frontend/issues/181

PR #2851, opened 2017-07-24
feat: Improve font/line-height on horizontal addons cards

### LINKED ISSUE #2661: Fonts for add-ons in horizontal card view should be modified
@sarahmonster pointed out that a larger font size with a lighter font weight would look good for the desktop horizontal card add-ons. She suggested:

```css
font-size: 18px;
font-weight: 500;
line-height: 1.2;
```

and it looked good, so I'll do that. I think that's closer to the mocks anyway and I sized them down like a doofus.

tofumatt:
Close #2661.

### Before
screenshot 2017-07-24 18 54 32
screenshot 2017-07-24 18 54 29

### After
screenshot 2017-07-25 15 42 45
screenshot 2017-07-25 15 41 16

### DIFF
--- a/src/amo/components/AddonsCard/styles.scss
+++ b/src/amo/components/AddonsCard/styles.scss
@@ -54,10 +54,6 @@
         width: auto;
       }
 
-      .SearchResult-name {
-        font-size: $font-size-s;
-      }
-
       .SearchResult-metadata,
       .SearchResult-summary {
         display: none;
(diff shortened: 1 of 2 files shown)


muffinresearch:
If the text is larger could the weight be reduced since it's larger and equally as heavy compared to the headings? 

This is probably worth getting UX input on too if you haven't already.

muffinresearch:
A lighter-weight as well as increased font-size was the original suggestion in the linked issue.

tofumatt:
The mocks actually have them with a lot less weight actually. I'll give that another go. I think the text is a bit too small in the mocks but these are actually maybe too big. I'll do another quick pass.

PR #2853, opened 2017-07-25
fix: Categories temporary 404 while loading

### LINKED ISSUE #2846: Tapping on any category loads a 404 page for a few seconds
Steps to reproduce:
1. Load AMO homepage on you device https://addons-dev.allizom.org/en-US/android/
2. Tap on a category i.e. "Blocks ads"

Expected results:
Category is loaded without any display or layout issues.

Actual results:
A 404 page is loaded for a few seconds.

Notes/Issues:
- happening first time the page is loaded - reproduce is you must load the page again in a new tab;
- reproducing on new desktop pages also;
- may be a regression from #2084 

Verified on FF54(Win 7). Issue is reproducing around all AMO servers.
Video for this issue:
videotogif_2017 07 24_11 04 39

tofumatt:
Fix #2846.

We weren't checking to see if we had *any* category data yet (or even did a proper loading check) so we'd hit a temporary 404 screen during the load. Because we fetch all categories at once this just marks whether or not we've fetched them and only renders a 404 for a non-existent category once we've compared it to the fetched categories.

### Before
videotogif_2017 07 24_11 04 39

### DIFF
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -67,23 +67,28 @@ export function mapStateToProps(state, ownProps) {
     { ...state.search.filters, page: parsePage(state.search.page) },
     { ...filters, page: queryParams.page },
   );
-  if (filtersMatchState) {
-    return {
-      addonType: filters.addonType,
-      category,
-      filters,
-      pathname,
-      queryParams,
-      ...state.search,
-    };
-  }
 
-  return {
+  const loading = state.categories.loading || state.search.loading;
+
+  const props = {
     addonType: filters.addonType,
     category,
+    loading,
     pathname,
     queryParams,
   };
+
+  if (filtersMatchState) {
+    return {
+      ...props,
+      filters,
+      count: state.search.count,
+      page: state.search.page,
+      results: state.search.results,
+    };
+  }
+
+  return props;
 }
 
 export default compose(
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -8,10 +8,11 @@ import { SearchBase } from 'amo/components/Search';
 import createStore from 'amo/store';
 import { categoriesFetch } from 'core/actions/categories';
 import { searchStart } from 'core/actions/search';
-import { ADDON_TYPE_THEME } from 'core/constants';
+import { ADDON_TYPE_THEME, CLIENT_APP_FIREFOX } from 'core/constants';
 import I18nProvider from 'core/i18n/Provider';
 import { getFakeI18nInst } from 'tests/unit/helpers';
 import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
+
 
 describe('Category', () => {
   let category;
@@ -67,8 +68,11 @@ describe('Category', () => {
     sinon.assert.calledWith(fakeDispatch, categoriesFetch());
   });
 
-  it('should return 404 if category is falsy and loading is false', () => {
-    const root = mountRender();
+  it('should return 404 if no category was found and loading is false', () => {
+    const root = mountRender({
+      category: null,
+      loading: false,
+    });
 
     expect(root.find(NotFound)).toHaveLength(1);
   });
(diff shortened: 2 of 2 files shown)


src/amo/components/Category/index.js
```diff
@@ -67,11 +68,17 @@ export function mapStateToProps(state, ownProps) {
     { ...state.search.filters, page: parsePage(state.search.page) },
     { ...filters, page: queryParams.page },
   );
+
+  const hasFetched = state.categories.hasFetched;
```
muffinresearch: Could this be derived from the category data rather than needing to be passed as a prop?

tofumatt:
It could but I figured doing it once in the reducer made a bit more sense-categories are fetched essentially all at once so knowing whether we've made the request seemed worth storing once.

- Matthew Riley MacPherson (sent from mobile)

src/amo/components/Category/index.js
```diff
@@ -67,11 +68,17 @@ export function mapStateToProps(state, ownProps) {
...
+  const loading = state.categories.loading || state.search.loading;
```
kumar303: This looks like the right thing to do. Once you make this `loading` prop effective I believe that you won't need `hasFetched`.

src/amo/components/Category/index.js
```diff
@@ -67,11 +68,17 @@ export function mapStateToProps(state, ownProps) {
...
+
   if (filtersMatchState) {
     return {
       addonType: filters.addonType,
       category,
       filters,
+      hasFetched,
+      loading,
```
kumar303: This prop is not affecting anything because `...state.search` (below) looks to be overriding it.

src/amo/components/Category/index.js
```diff
@@ -67,11 +68,17 @@ export function mapStateToProps(state, ownProps) {
...
       pathname,
       queryParams,
       ...state.search,
```
kumar303: I know this is old code but could you remove the spread and instead reference each one of the `state` attributes?

src/amo/components/Category/index.js
kumar303: I don't think you need to introduce this new state attribute since you already have it as `state.categories.loading`. My hunch is that you added it because the new `loading` prop was not taking effect (I explained that below).

tofumatt:
Yeah that makes sense upon looking at this again. I'll do the check in the component, should be ready after lunch.

src/amo/components/Category/index.js
tofumatt: It needs to be a prop as it's based on data in *all* categories but the prop that the component gets is just one category.

tofumatt:
Changed the check to be inside the component's `mapStateToProps` (we don't have all the required data in the component itself and passing in every category just for the check seemed silly). Ready for another r?

tofumatt:
> I don't understand why you need to loop through all loaded categories in the state mapper when there is only one API fetch that populates the categories (not multiple).

I'm just looping through until I find a populated one because it could be there are legitimately no dictionary categories for android, but I can't know if we fetched anything until I see at least one populated category. Previously I set `hasFetched` in state to mark "we made the request and got all the categories" but that was r-'d

That said, it seems object spread strikes again. I didn't clue in to the fact that the spread overwrites the loading prop I see, I think. Rawr. I feel dumb.

I'll try your patch and tweak the test. really sorry I didn't notice that 😞

tofumatt:
Gosh, you're right, I got really tripped up by the spread.

Ready for another r? Basically it's your patch with better loading checks and a few extra tests.

Sorry 'bout that.

tests/unit/amo/components/TestCategory.js
```diff
@@ -139,11 +143,44 @@ describe('Category.mapStateToProps()', () => {
...
       addonType: ADDON_TYPE_THEME,
       category: null,
+      loading: true,
       pathname: '/themes/ad-block/',
       queryParams: { page: 1 },
     });
   });
+
+  it('sets loading to true if categories and search are loading', () => {
+    const { store } = createStore();
```
kumar303: You could move this to `beforeEach` and share it for every test

tofumatt: Good call, will do.

PR #2860, opened 2017-07-26
Eslint 4.3

tofumatt:
eslint complained about a LOT of indentation errors here. I ran `eslint --fix`... but I wonder what the diff looks like and if it's what we wanted. Posting here to see the diff.

Closes #2721

### DIFF
--- a/package.json
+++ b/package.json
@@ -209,7 +209,7 @@
     "css-loader": "^0.28.3",
     "deepcopy": "^0.6.3",
     "enzyme": "^2.8.2",
-    "eslint": "^3.19.0",
+    "eslint": "^4.3.0",
     "eslint-config-amo": "^0.1.1",
     "eslint-plugin-promise": "^3.5.0",
     "fetch-mock": "^5.10.0",
--- a/src/amo/api/index.js
+++ b/src/amo/api/index.js
@@ -52,40 +52,39 @@ export function submitReview({
   body,
   reviewId,
   ...apiCallParams
-}: SubmitReviewParams): Promise<ApiReviewType> {
-  return new Promise(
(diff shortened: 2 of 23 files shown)


muffinresearch:
I think most of these changes look ok - not sure why the test is showing still in progress it completed already. 

Maybe there's a few cases here where a manual update to the formatting would do things slightly differently.

muffinresearch:
@tofumatt what do you think about removing the space after `async` that one seems a little odd to me.

muffinresearch:
I'll file an issue to fix that in eslint-config-amo.

muffinresearch:
@tofumatt I patched the eslint config so this should allow the space behind the async keyword now.

muffinresearch:
@tofumatt if you can rebase this it should be pretty much good to go...

tofumatt:
Ten four, I’ll take a look after the standup.

tofumatt:
Sorry all for the merge conflicts this'll likely create! 😅

willdurand:
@tofumatt I get these warnings now:

```
can't resolve reference #/definitions/basicConfig from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
```

There is a similar issue on the airbnb preset: https://github.com/airbnb/javascript/issues/1488, which is related to https://github.com/yannickcr/eslint-plugin-react/pull/1292.

Do you see these messages too?

tofumatt:
Yeah, I’ve seen those too but they are apparently okay to ignore. They don’t cause anything to fail and are because of a dependency issue :-/

PR #2862, opened 2017-07-26
chore(package): update file-loader to version 1.0.0

tofumatt:
Closes #2861


tofumatt:
Ran locally, all seemed to work and the tests pass. Looking at the changelog it seems not to affect us so 👍

PR #2864, opened 2017-07-27
Fix screenshots overlap in safari

### LINKED ISSUE #2847: Images in Screenshots section overlaps the container if there are more than 1 images in Safari
Images in Screenshots section overlaps the container if there are more than 1 images in Safari

Steps to reproduce:
1. Open [https://addons.mozilla.org/en-US/firefox/search/?q=videos&type](https://addons.mozilla.org/en-US/firefox/search/?q=videos&type) in Safari
2. Go to any extension in the result and check for screenshots which has more than 1 image.
3. The images overlaps the About this extension container.

### What happened?
The images overlaps the About this extension container.

The images should be inside the screenshots container with a horizontal scroll

screen shot 2017-07-24 at 6 22 10 pm

muffinresearch:
Fixes #2847

Before: 

before

After:

after

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -162,6 +162,9 @@
 
   .Addon-screenshots {
     grid-column: 1 / 2;
+    // overflow required to fix content overlap in Safari.
+    // See https://github.com/mozilla/addons-frontend/issues/2847
+    overflow-x: hidden;
   }
 
   .Addon-overall-rating {


src/amo/components/Addon/styles.scss
```diff
@@ -162,6 +162,8 @@
 
   .Addon-screenshots {
     grid-column: 1 / 2;
+    // overflow required to fix #2847
```
tofumatt: Please link to the full URL for better UX 😁

PR #2866, opened 2017-07-27
Workaround transparency issue in Safari

### LINKED ISSUE #2865: Transparent gradient renders Grey in Safari
With the mamo cookie set visit: https://addons.allizom.org/en-US/firefox/addon/download-youtube/
in Safari - not the grey line above the read more link

Safari:

add-ons_for_firefox

Firefox: 

add-ons_for_firefox

muffinresearch:
Fixes #2865

Before:

add-ons_for_firefox 2

After:
add-ons_for_firefox

### DIFF
--- a/src/ui/components/ShowMoreCard/ShowMoreCard.scss
+++ b/src/ui/components/ShowMoreCard/ShowMoreCard.scss
@@ -2,7 +2,10 @@
 
 .ShowMoreCard-contents {
   &::after {
-    background: linear-gradient(transparent, $base-color);
+    // rgba(255, 255, 255, 0) instead of transparent prevents
+    // grey line in Safari.
+    // See: https://github.com/mozilla/addons-frontend/issues/2865
+    background: linear-gradient(rgba(255, 255, 255, 0), $base-color);
     bottom: 0;
     content: "";
     height: 20px;


src/ui/components/ShowMoreCard/ShowMoreCard.scss
```diff
@@ -2,7 +2,9 @@
 
 .ShowMoreCard-contents {
   &::after {
-    background: linear-gradient(transparent, $base-color);
+    // rgba(255,255,255,0) instead of transparent prevents
+    // grey line in Safari see #2865
```
tofumatt: Can you link to the actual issue URL rather than the number? It makes it easier to get to and more obvious what that is... we know to search a GitHub repo for issues but a newbie might not. 😀

tofumatt: Also I'd put a full-stop after "Safari" and make "See: https://github.com[...]" its own sentence.

src/ui/components/ShowMoreCard/ShowMoreCard.scss
```diff
@@ -2,7 +2,9 @@
...
+    background: linear-gradient(rgba(255,255,255,0), $base-color);
```
tofumatt: I think the linter doesn't like the lack of spaces here. Also to prevent using non-variable colours you could do `transparent($base-color, 1)` or whatever the SCSS was, but that's maybe superfluous.

Issue #2516, opened 2017-06-05
Categories page does not surface API errors.


muffinresearch:
Locally if the local API isn't running directly reloading a category page (/en-US/android/extensions/)  results in a 200 with no results. There is also no error message shown to indicate something has gone wrong.

I would expect this to be a 500 for a reload, and both server + client-side requests should at least show an error page.

muffinresearch:
Not sure if this has always done this or if it's a side-effect of the saga implementation?

tofumatt:
It’s a saga one I believe, though it SHOULD say “no categories found”. So something isn’t 100% right, I’ll have a look.

ValentinaPC:
Please add some STR for us to check this or the label "qa: not needed". Thanks!

tofumatt:
You would need to run the site locally to get this to work I think, running it, loading a categories page, shutting down the API server, and then refreshing it. I think this is likely a non-QA-needed issue, correct me if I'm wrong anyone!

kumar303:
That's correct. @ValentinaPC if you have time to test for any regressions, that would still be helpful. The category page should still be working as it did before.

ValentinaPC:
> That's correct. @ValentinaPC if you have time to test for any regressions, that would still be helpful. The category page should still be working as it did before.

Sure thing!

ValentinaPC:
Verified for regression while loading all categories pages for extensions and themes, and were not found.

PR #2867, opened 2017-07-27
Handle API errors on Categories pages

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2516

* Adds a generic error handler to the main category listing and category detail pages
* Removes the custom error handling logic for categories
* Fixes broken Flow typing

Here are some screenshots of what an unexpected error might look. Keep in mind that most API errors will be more informative but those are harder to simulate.

On mobile:

screenshot 2017-07-27 11 57 26
screenshot 2017-07-27 14 01 35

On desktop:

screenshot 2017-07-27 12 08 19
screenshot 2017-07-27 14 01 19


muffinresearch:
Would it be possible to show a full error page for errors like these instead of the inline error?

kumar303:
> Would it be possible to show a full error page for errors like these instead of the inline error?

Do you mean a 500 error page? I think that would be possible but it would be better as a follow-up patch because it will require some customization of the ErrorHandler.

muffinresearch:
> Do you mean a 500 error page? I think that would be possible but it would be better as a follow-up patch because it will require some customization of the ErrorHandler.

Yes essentially the equivalent of a 500 error page. Mainly because it side-steps the possibility of other broken UI being displayed alongside the error message. As a follow-up would be totally fine.

src/amo/components/Categories/index.js
```diff
@@ -93,37 +109,49 @@ export class CategoriesBase extends React.Component {
...
-                >
-                  {category.name}
-                </Button>
-              </li>
-            ))}
+            {categories.map((category) => {
+              // Flow cannot figure out CategoryType in this case.
+              // See https://github.com/facebook/flow/issues/2174
+              // and https://github.com/facebook/flow/issues/2221
+              // $FLOW_IGNORE
```
tofumatt: Confusing that this is not a disable/enable like eslint 😢

src/amo/components/Categories/index.js
```diff
@@ -93,37 +109,49 @@ export class CategoriesBase extends React.Component {
...
+              const name = category.name;
+              // $FLOW_IGNORE
```
tofumatt: Oh wait I see this is a `disable-next-line`-type comment.

src/amo/components/Category/styles.scss
```diff
@@ -0,0 +1,3 @@
+.Category .ErrorList {
```
tofumatt: Because `.ErrorList` is shared and overridden in so many places we should be aware that it's going to eventually have some possibly weird cascades. This is fine for now and nothing that needs refactoring, just something to remain vigilant about 😄

src/amo/sagas/categories.js
```diff
@@ -5,25 +5,23 @@ import { hideLoading, showLoading } from 'react-redux-loading-bar';
...
+  const errorHandler = createErrorHandler(errorHandlerId);
   try {
     yield put(showLoading());
     const api = yield select(getApi);
     const response = yield call(categoriesApi, { api });
     yield put(categoriesLoad(response));
-  } catch (err) {
-    log.warn('Categories failed to load:', err);
-    yield put(categoriesFail(err));
+  } catch (error) {
```
tofumatt: Thanks, I think short variables names are silly. 👍  (If I did this, which I think I did, I was wrong.)

src/core/actions/categories.js
```diff
@@ -1,13 +1,15 @@
 import {
   CATEGORIES_FETCH,
   CATEGORIES_LOAD,
-  CATEGORIES_FAIL,
 } from 'core/constants';
 
-export function categoriesFetch() {
+export function categoriesFetch({ errorHandlerId } = {}) {
+  if (!errorHandlerId) {
```
tofumatt: I wish we could do some kind of meta-programming for JS variable names and write a convenience function for things like this.

src/core/constants.js
```diff
@@ -100,7 +100,6 @@ export const SEARCH_SORT_TOP_RATED = 'rating';
 export const FETCH_ADDON = 'FETCH_ADDON';
 export const CATEGORIES_FETCH = 'CATEGORIES_FETCH';
 export const CATEGORIES_LOAD = 'CATEGORIES_LOAD';
-export const CATEGORIES_FAIL = 'CATEGORIES_FAIL';
```
tofumatt: Thanks for removing these as you add errorHandlers. I admit I was slow to understand how they worked but I definitely like their flexibility more, even if they're a bit less obvious than just dispatching error actions. 👍

src/core/reducers/categories.js
```diff
@@ -25,7 +24,6 @@ export function emptyCategoryList() {
 
 const initialState = {
   categories: emptyCategoryList(),
-  error: false,
```
tofumatt: 👍

src/core/reducers/categories.js
```diff
@@ -34,7 +32,7 @@ export default function categories(state = initialState, action) {
 
   switch (action.type) {
     case CATEGORIES_FETCH:
-      return { ...state, ...payload, loading: true };
```
tofumatt: Yes! Thanks so much. 👍

tests/unit/amo/components/TestCategoriesPage.js
```diff
@@ -1,159 +1,21 @@
...
-  it('renders loading when loading', () => {
-    const root = render({
-      addonType: ADDON_TYPE_EXTENSION,
-      categories: [],
-      error: false,
-      loading: true,
-    });
-
-    expect(root.find('.Categories-loading-info'))
-      .toIncludeText('Loading');
```
tofumatt: Do we not load this text anywhere anymore? It is slightly useful for a11y purposes.

tests/unit/amo/components/TestCategoriesPage.js
```diff
@@ -1,159 +1,21 @@
...
-  it('renders a message when there are no categories', () => {
-    const root = render({
-      addonType: ADDON_TYPE_EXTENSION,
-      categories: [],
-      error: false,
-      loading: false,
-    });
-
-    expect(root.find('.Categories-none-loaded-message'))
-      .toIncludeText('No categories found.');
```
tofumatt: This seems like a useful message to maintain if users try to load categories for an `addonType` that doesn't have categories. Did this move and I just didn't notice, or was it removed?

src/amo/components/Categories/index.js
```diff
@@ -93,37 +109,49 @@ export class CategoriesBase extends React.Component {
...
   }
 }
 
-export function mapStateToProps(state) {
+export function mapStateToProps(
+  state: {| api: ApiStateType, categories: CategoriesStateType |}
+) {
   const clientApp = state.api.clientApp;
-  const categories = state.categories.categories[clientApp];
+  const categories = state.categories.categories[clientApp || ''];
```
tofumatt: This seems quite odd and sort of fragile, why the `''`?

tests/unit/core/actions/test_categories.js
```diff
@@ -2,10 +2,22 @@ import * as actions from 'core/actions/categories';
...
 
 describe('CATEGORIES_FETCH', () => {
-  const action = actions.categoriesFetch();
+  function _categoriesFetch({ errorHandlerId = 'some-handler-id' } = {}) {
+    return actions.categoriesFetch({ errorHandlerId });
+  }
 
   it('sets the type', () => {
-    expect(action.type).toEqual('CATEGORIES_FETCH');
+    expect(_categoriesFetch().type).toEqual('CATEGORIES_FETCH');
```
tofumatt: Do we have a constant for this? If so I think we should use it.

tests/unit/core/reducers/test_categories.js
```diff
@@ -25,16 +25,11 @@ describe('categories reducer', () => {
     expect(loading).toEqual(false);
   });
 
-  it('defaults to not error', () => {
```
tofumatt: I've mentioned this before, but I really like removing error state from the reducers. 👍

src/core/actions/categories.js
kumar303: When we use Flow 100% we won't need it. We could also try [invariant](https://github.com/zertosh/invariant) in the meantime but it will still require the boilerplate code.

tofumatt: Is that still a goal though? I thought the outcome of our discussions in San Francisco was that Flow was useful for propTypes and maybe state/API annotations, but not every single function.

tests/unit/amo/components/TestCategoriesPage.js
kumar303: We still do. This assertion just moved to https://github.com/kumar303/addons-frontend/blob/4979224ecc6caa2bba873f76a6634847758e043c/tests/unit/amo/components/TestCategories.js#L64-L65

tests/unit/amo/components/TestCategoriesPage.js
kumar303: The message was still there in the component but it looks like it no longer has test coverage. Good catch. Fixed.

src/core/actions/categories.js
kumar303: > When we use Flow 100% we won't need it. 

^ That was a hypothetical statement to illustrate how we can only be sure that our app is calling action creators in the right way with something like Flow or with if statements that check every required variable for truthiness. Take your pick. 

Personally, I think this is an excellent example of how Flow can help us. If our app calls an action creator incorrectly it could be really hard to track down since that information is completely lost in the trail of dispatched actions (we have no trail of function calls to action creators).

kumar303:
Thanks for the last one. It's ready for another review.

Issue #2737, opened 2017-07-11
Convert search to load data with a saga


tofumatt:
We should migrate the search to use reduxs-saga. Related to #2356.

Steps to test for regressions:

* Search for add-ons
* Search should load instantly with loading text placeholders while searching
* Content should then appear

Blank searches should not be allowed and should display an error.

tofumatt:
Because of the DNT stuff I'm gonna push this to the next milestone, sorry all 😞

ValentinaPC:
> Search should load instantly with loading text placeholders while searching

- like this:
![place](https://user-images.githubusercontent.com/15685960/29372900-21ca4eea-82b5-11e7-9ac1-cd9115a6cf27.png)
?

tofumatt:
Yes :-)

ValentinaPC:
> Blank searches should not be allowed and should display an error.
- blank search are blocked only in Explore - intended?

Verified this on AMO-dev FF54(Win 7)
Search is working as expected.

Issue #2857, opened 2017-07-26
Convert Category detail page to load data with a saga


kumar303:
We already have a categories saga but we need to update the Category detail page to use it.

Here's how to test for regressions:
* From the home page, click 'Browse all extensions' (or themes)
* Click on 'Explore all categories'
* Click on one of the categories
* Make sure the category detail page loads and works fine

This is a step towards https://github.com/mozilla/addons-frontend/issues/2356

tofumatt:
This uses the search utils though, right? Effectively this is part of the search change because behind-the-scenes they're just using the same search component, I thought.

kumar303:
Do you mean https://github.com/mozilla/addons-frontend/issues/2737 will cover converting the Category detail page to a saga? If so, I think it should be done in a separate patch since `Category` is a top-level component.

tofumatt:
Well, SearchBase will start using a saga so I'd imagine it will implicitly convert. Okay but I'll try to keep it separate.

kumar303:
I took a quick look at this. It definitely needs a saga implementation for search first but I still think it can be kept for a separate patch, after https://github.com/mozilla/addons-frontend/issues/2737

kumar303:
This was fixed in #2737 ❤️

ValentinaPC:
Verified this on AMO-dev FF54(Win 7).
For each category issue #2953 is reproducing....don't know if is the same root cause or not?
Please see video:
![cat](https://user-images.githubusercontent.com/15685960/29371597-6df40c74-82b1-11e7-8577-b2e7ecff77e6.gif)

PR #2870, opened 2017-07-28
Search sagas

### LINKED ISSUE #2871: Create Search Saga code
To make #2737 easier to review I've separated out the saga code itself so the patches are smaller. This issue is just to add the saga support to the app.

tofumatt:
Adds the sagas and tests for them in preparation of #2737. Doesn't use them yet. 😄 

I separated this bit out so things are easier to review.

Fixes #2871.

### DIFF
--- a/src/amo/sagas/index.js
+++ b/src/amo/sagas/index.js
@@ -1,7 +1,7 @@
 // Disabled because of
 // https://github.com/benmosher/eslint-plugin-import/issues/793
 /* eslint-disable import/order */
-import { fork } from 'redux-saga/effects';
+import { all, fork } from 'redux-saga/effects';
 /* eslint-enable import/order */
 
 import categories from 'amo/sagas/categories';
@@ -13,11 +13,11 @@ import addons from 'core/sagas/addons';
 
 // Export all sagas for this app so runSaga can consume them.
 export default function* rootSaga() {
-  yield [
+  yield all([
     fork(addons),
     fork(categories),
     fork(featured),
     fork(landing),
     fork(reviews),
-  ];
+  ]);
 }
--- a/src/amo/store.js
+++ b/src/amo/store.js
@@ -19,27 +19,29 @@ import search from 'core/reducers/search';
 import { middleware } from 'core/store';
 
 
+export const AMO_REDUCERS = {
+  addons,
+  api,
+  auth,
+  categories,
+  errors,
+  errorPage,
+  featured,
+  infoDialog,
+  installations,
+  landing,
+  loadingBar: loadingBarReducer,
+  reduxAsyncConnect,
+  reviews,
+  search,
+  viewContext,
+};
+
 export default function createStore(initialState = {}) {
   const sagaMiddleware = createSagaMiddleware();
 
   const store = _createStore(
-    combineReducers({
-      addons,
-      api,
-      auth,
-      categories,
-      errors,
-      errorPage,
-      featured,
-      infoDialog,
-      installations,
-      landing,
-      loadingBar: loadingBarReducer,
-      reduxAsyncConnect,
-      reviews,
-      search,
-      viewContext,
-    }),
+    combineReducers(AMO_REDUCERS),
     initialState,
     middleware({ sagaMiddleware }),
   );
--- a/src/amo/store.js
+++ b/src/amo/store.js
@@ -19,29 +19,27 @@ import search from 'core/reducers/search';
 import { middleware } from 'core/store';
 
 
-export const AMO_REDUCERS = {
-  addons,
-  api,
-  auth,
-  categories,
-  errors,
-  errorPage,
-  featured,
-  infoDialog,
-  installations,
-  landing,
-  loadingBar: loadingBarReducer,
-  reduxAsyncConnect,
-  reviews,
-  search,
-  viewContext,
-};
-
 export default function createStore(initialState = {}) {
   const sagaMiddleware = createSagaMiddleware();
 
   const store = _createStore(
-    combineReducers(AMO_REDUCERS),
+    combineReducers({
+      addons,
+      api,
+      auth,
+      categories,
+      errors,
+      errorPage,
+      featured,
+      infoDialog,
+      installations,
+      landing,
+      loadingBar: loadingBarReducer,
+      reduxAsyncConnect,
+      reviews,
+      search,
+      viewContext,
+    }),
     initialState,
     middleware({ sagaMiddleware }),
   );
--- a/src/core/actions/search.js
+++ b/src/core/actions/search.js
@@ -5,10 +5,10 @@ import {
 } from 'core/constants';
 
 
-export function searchStart({ filters, page, results }) {
+export function searchStart({ errorHandlerId, filters, page, results }) {
   return {
     type: SEARCH_STARTED,
-    payload: { filters, page, results },
+    payload: { errorHandlerId, filters, page, results },
   };
 }
 
--- a/src/core/actions/search.js
+++ b/src/core/actions/search.js
@@ -6,6 +6,19 @@ import {
 
 
 export function searchStart({ errorHandlerId, filters, page, results }) {
+  if (!errorHandlerId) {
+    throw new Error('errorHandlerId is required');
+  }
+  if (!filters) {
+    throw new Error('filters are required');
+  }
+  if (!page) {
+    throw new Error('page is required');
+  }
+  if (!results) {
+    throw new Error('results are required');
+  }
+
   return {
     type: SEARCH_STARTED,
     payload: { errorHandlerId, filters, page, results },
--- a/src/core/sagas/search.js
+++ b/src/core/sagas/search.js
@@ -1,7 +1,7 @@
 // Disabled because of
 // https://github.com/benmosher/eslint-plugin-import/issues/793
 /* eslint-disable import/order */
-import { call, put, select, takeEvery } from 'redux-saga/effects';
+import { call, put, select, takeLatest } from 'redux-saga/effects';
 /* eslint-enable import/order */
 
 import { searchLoad } from 'core/actions/search';
@@ -37,5 +37,5 @@ export function* fetchSearchResults({ payload }) {
 }
 
 export default function* searchSaga() {
-  yield takeEvery(SEARCH_STARTED, fetchSearchResults);
+  yield takeLatest(SEARCH_STARTED, fetchSearchResults);
 }
--- a/src/core/sagas/search.js
+++ b/src/core/sagas/search.js
@@ -8,7 +8,7 @@ import { searchLoad } from 'core/actions/search';
 import { search as searchApi } from 'core/api';
 import { SEARCH_STARTED } from 'core/constants';
 import log from 'core/logger';
-import { createErrorHandler, getApi, getAuth } from 'core/sagas/utils';
+import { createErrorHandler, getState } from 'core/sagas/utils';
 
 
 export function* fetchSearchResults({ payload }) {
@@ -20,10 +20,14 @@ export function* fetchSearchResults({ payload }) {
   try {
     const { filters, page } = payload;
 
-    const api = yield select(getApi);
-    const auth = yield select(getAuth);
+    const state = yield select(getState);
 
-    const response = yield call(searchApi, { api, auth, filters, page });
+    const response = yield call(searchApi, {
+      api: state.api,
+      auth: state.auth,
+      filters,
+      page,
+    });
 
     yield put(searchLoad({ page, filters, ...response }));
   } catch (error) {
--- a/src/core/sagas/search.js
+++ b/src/core/sagas/search.js
@@ -1,0 +1,37 @@
+// Disabled because of
+// https://github.com/benmosher/eslint-plugin-import/issues/793
+/* eslint-disable import/order */
+import { call, put, select, takeEvery } from 'redux-saga/effects';
+/* eslint-enable import/order */
+
+import { searchLoad } from 'core/actions/search';
+import { search as searchApi } from 'core/api';
+import { SEARCH_STARTED } from 'core/constants';
+import log from 'core/logger';
+import { createErrorHandler, getApi, getAuth } from 'core/sagas/utils';
+
+
+export function* fetchSearchResults({ payload }) {
+  const { errorHandlerId } = payload;
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const { filters, page } = payload;
+
+    const api = yield select(getApi);
+    const auth = yield select(getAuth);
+
+    const response = yield call(searchApi, { api, auth, filters, page });
+
+    yield put(searchLoad({ page, filters, ...response }));
+  } catch (error) {
+    log.warn(`Search results failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* searchSaga() {
+  yield takeEvery(SEARCH_STARTED, fetchSearchResults);
+}
--- a/src/core/sagas/search.js
+++ b/src/core/sagas/search.js
@@ -28,8 +28,9 @@ export function* fetchSearchResults({ payload }) {
       filters,
       page,
     });
+    const { entities, result } = response;
 
-    yield put(searchLoad({ page, filters, ...response }));
+    yield put(searchLoad({ filters, entities, result }));
   } catch (error) {
     log.warn(`Search results failed to load: ${error}`);
     yield put(errorHandler.createErrorAction(error));
--- a/src/core/sagas/utils.js
+++ b/src/core/sagas/utils.js
@@ -1,9 +1,7 @@
-/* @flow */
 import { ErrorHandler } from 'core/errorHandler';
-import log from 'core/logger';
-import type { ApiStateType } from 'core/reducers/api';
+import defaultLog from 'core/logger';
 
-export function createErrorHandler(id: string): typeof ErrorHandler {
+export function createErrorHandler(id, { log = defaultLog } = {}) {
   return new ErrorHandler({
     id,
     // Make sure the dispatch() method can't be used. A saga will yield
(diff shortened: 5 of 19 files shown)


tests/unit/core/sagas/testUtils.js
```diff
@@ -6,13 +6,23 @@ import SagaTester from 'redux-saga-tester';
...
 
 
 describe('Saga utils', () => {
+  it('does not allow usage of dispatch from a saga', () => {
+    const fakeLog = { error: sinon.stub() };
+    const errorHandler = createErrorHandler('some-error-handler', {
+      log: fakeLog });
+    errorHandler.dispatch('ANYTHING');
+
+    sinon.assert.calledWith(fakeLog.error,
```
tofumatt: This was lacking test coverage so I added it.

src/amo/sagas/index.js
```diff
@@ -12,10 +12,10 @@ import addons from 'core/sagas/addons';
 
 // Export all sagas for this app so runSaga can consume them.
 export default function* rootSaga() {
-  yield [
+  yield all([
```
kumar303: Was this necessary? I think `fork()` already instructs the middleware to make a non-blocking call.

src/core/actions/search.js
```diff
@@ -5,10 +5,10 @@ import {
 } from 'core/constants';
 
 
-export function searchStart({ filters, page, results }) {
+export function searchStart({ errorHandlerId, filters, page, results }) {
```
kumar303: Since all of these parameters are required, the function should throw an error if any are empty. This is important because action creators are one part of the redux flow that we can't capture and replay. In other words, if there is a bug in some calling code where it's passing an empty parameter to an action creator, it might be hard to track down.

src/core/sagas/utils.js
```diff
@@ -1,9 +1,7 @@
-/* @flow */
```
kumar303: Sorry if this was giving you grief. I think it's important to keep Flow where we have it, especially since the problem I mentioned above (possible missing params in your action creator) are protected nicely by Flow. Here is a patch to put it back for you:

```diff
diff --git a/src/core/sagas/utils.js b/src/core/sagas/utils.js
index 65e66069..f35ab181 100644
--- a/src/core/sagas/utils.js
+++ b/src/core/sagas/utils.js
@@ -1,7 +1,12 @@
+/* @flow */
 import { ErrorHandler } from 'core/errorHandler';
 import defaultLog from 'core/logger';
+import type { ApiStateType } from 'core/reducers/api';
 
-export function createErrorHandler(id, { log = defaultLog } = {}) {
+export function createErrorHandler(
+  id: string,
+  { log = defaultLog }: {| log: typeof defaultLog |} = {}
+): typeof ErrorHandler {
   return new ErrorHandler({
     id,
     // Make sure the dispatch() method can't be used. A saga will yield
@@ -12,11 +17,11 @@ export function createErrorHandler(id, { log = defaultLog } = {}) {
 }
 
 // Convenience function to extract API info.
-export function getApi(state) {
+export function getApi(state: {| api: ApiStateType |}): ApiStateType {
   return state.api;
 }
 
 // Convenience function to extract authentication info.
-export function getAuth(state) {
+export function getAuth(state: Object) {
   return state.auth;
 }
```

src/core/sagas/utils.js
```diff
@@ -14,6 +12,11 @@ export function createErrorHandler(id: string): typeof ErrorHandler {
 }
 
 // Convenience function to extract API info.
-export function getApi(state: {| api: ApiStateType |}): ApiStateType {
+export function getApi(state) {
   return state.api;
 }
+
+// Convenience function to extract authentication info.
+export function getAuth(state) {
```
kumar303: can't we just use one function now?

```js
type StateType = {|
  api: ApiStateType,
  auth: Object,
|};

export function getState(state: StateType) {
  return state;
}
```

tests/unit/core/sagas/testUtils.js
kumar303: Thanks for adding test coverage! We really need to get back up to 100% coverage so we don't miss stuff like this.

src/amo/sagas/index.js
tofumatt: Not using `all()` causes a deprecation warning from `redux-saga`. See the release notes in: https://github.com/redux-saga/redux-saga/releases/tag/v0.15.0

I can add that link to the comment/commit if you'd like! 😄

src/core/actions/search.js
tofumatt: Fair enough. I think results might not actually be required in a refactor later but I will make them required for now.

src/core/sagas/utils.js
tofumatt: I removed it because I find that syntax really head-scratching and it really polluted an otherwise quite straightforward file. I honestly have little idea looking at that function signature what that code does anymore, and I can't imagine a relative ES6 newcomer groking it well.

src/core/sagas/utils.js
tofumatt: I guess so. I was thinking it was cleaner not to just extract the entire state in case we want to manipulate certain things in the future, but I guess we would have state extraction helpers elsewhere that weren't saga-specific, so yeah.

src/amo/sagas/index.js
kumar303: No need, I believe you! I see in the release notes that it's no different than before; it's just more explicit. Works for me.

src/core/sagas/utils.js
kumar303: One thing I find really nice about annotating the `state` object is that you know exactly what you're getting. You can follow `ApiStateType` and find out what all the `state.api.*` properties are that you get access to. This would be hugely informative to an ES6 newcomer who is trying to understand how our app works. In fact, they could be pretty lost without it! Obviously it's not helpful in the case of `state.auth` because we don't have types for that one yet.

tofumatt: Heh, I like the idea but an ES6 newcomer is likely to be intimidated by flow's syntax.

Now we're returning all of the state object in the new `getState` function. I guess I can add `Object` to the rest of the state objects (or omit them as you did below) but that feels even worse: we are half-documenting the keys returned by the helper.

src/core/actions/search.js
tofumatt: Making these required meant a lot of changes, so the PR diff definitely grew.

tofumatt:
Shoot, I think I forgot to add coverage for the required params. I'll do that now.

src/core/sagas/utils.js
```diff
@@ -1,19 +1,17 @@
...
-      'ErrorHandler cannot dispatch from a saga'),
+    dispatch: () => log.error('ErrorHandler cannot dispatch from a saga'),
   });
 }
 
-// Convenience function to extract API info.
-export function getApi(state: {| api: ApiStateType |}): ApiStateType {
-  return state.api;
+// Convenience function to extract state info.
+export function getState(state) {
```
kumar303: I think an annotation of `api: ApiStateType` would be really helpful here, especially since `state` is now completely non-intuitive. You mentioned that it's only half-documented since we don't have an `auth` definition but half is better than nothing. How could someone even look at this function and know what to do with `state`?

Also, I'm fine with not adding `/* @flow */` to new files but removing it from existing ones? You need to provide a solid argument for that and I don't see one here. It's hard enough for us to get decent Flow coverage; what's the point of reducing our coverage? If it's slowing you down on this patch, just let me know and I can fix it up on master once you land it.

tests/unit/amo/components/TestCategory.js
```diff
@@ -124,7 +124,12 @@ describe('Category.mapStateToProps()', () => {
   });
 
   it('passes the search state if the filters and state matches', () => {
-    store.dispatch(searchStart({ filters, results: [] }));
+    store.dispatch(searchStart({
```
kumar303: This should use a helper like `_searchStart()` that sets default argument values so that you aren't repeating yourself with arguments that are not relevant to what you are testing.

tests/unit/core/actions/test_search.js
```diff
@@ -1,15 +1,60 @@
...
   });
 
   it('sets the query and existing results', () => {
-    expect(action.payload).toEqual({ filters: { query: 'foo' }, page: 5, results: [] });
+    expect(action.payload).toEqual(payload);
+  });
+
+  it('throws an error if errorHandlerId is empty', () => {
+    expect(() => {
+      actions.searchStart({
```
kumar303: These also should all use a helper function that sets default values since those values aren't relevant to each test

tests/unit/core/sagas/testSearch.js
```diff
@@ -0,0 +1,106 @@
...
+  it('clears the error handler', async() => {
+    const entities = sinon.stub();
+    const result = sinon.stub();
+
+    const filters = { query: 'test' };
+
+    mockApi
+      .expects('search')
+      .once()
+      .withArgs({ api: apiState, auth: authState, filters, page: 1 })
```
kumar303: I don't think you need to make an assertion about the calling arguments here because there is already a test for that. This specific test only needs an API response to set up a condition for clearing the error.

tests/unit/core/sagas/testSearch.js
```diff
@@ -0,0 +1,106 @@
...
+      .returns(Promise.resolve({ entities, result }));
+
+    _searchStart({
+      filters,
+      page: 1,
```
kumar303: `page` and `results` look like arguments that are not relevant to this test or the ones below it. They can be moved into `_searchStart()` as defaults so that you don't have to repeat them each time.

src/core/sagas/search.js
```diff
@@ -0,0 +1,41 @@
...
+
+    yield put(searchLoad({ page, filters, ...response }));
+  } catch (error) {
+    log.warn(`Search results failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* searchSaga() {
+  yield takeEvery(SEARCH_STARTED, fetchSearchResults);
```
kumar303: Does this need to process parallel search requests? I don't think we have any page that does that currently. Maybe `takeLatest` is enough so that the last request gets cancelled?

src/amo/store.js
```diff
@@ -19,27 +19,29 @@ import search from 'core/reducers/search';
 import { middleware } from 'core/store';
 
 
+export const AMO_REDUCERS = {
```
kumar303: any reason why this is in all caps? I feel like that's a bad habit we've gotten into.

tests/unit/core/sagas/testUtils.js
```diff
@@ -6,30 +6,36 @@ import SagaTester from 'redux-saga-tester';
...
-    store.dispatch(setLang('en-US'));
-
+    const { store } = dispatchSignInActions();
     const state = store.getState();
 
     const sagaTester = new SagaTester({
-      initialState: { api: state.api },
-      reducers: { api: apiReducer },
+      initialState: state,
+      reducers: AMO_REDUCERS,
```
kumar303: what do we gain by passing all of AMO's reducers to this tester? I am struggling to think of a benefit. Every reducer gets execute on every action dispatch so it will introduce slowness. The point of this test is just to make sure the entire state is exposed but you can still prove that with a limited reducer map such as `{ api: apiReducer, auth: authReducer }`.

kumar303:
Please also merge in master and fix the call to `yield select(getApi)` in `amo/sagas/landing`.

src/core/sagas/utils.js
tofumatt: I find the syntax for objects with default values (eg `{ log = defaultLog }: {| log: typeof defaultLog |} = {}`) to be totally hostile and it's really frustrating to me. But I think I've made the best case I can against flow already which is that I find the syntax to be actively distracting.

I think `state` is largely understood to be our state object, the result of `store.getState()` from our redux `store` object. That's not intuitive if you haven't used a react/redux app (or maybe even haven't used our app), but I think flow's annotations are even less intuitive than the `state` object in this function.

When we're often declaring the types in a separate file I still need to check somewhere else to see exactly what `ApiStateType` is anyway. I've yet to hit a bug that flow would have caught. That's perhaps a bit chicken and egg because we lack any substantial flow coverage, but to me the syntax fights with me more than the bugs it would catch were we to have coverage.

src/core/sagas/search.js
tofumatt: Good call. I think I confused calling the `search` API multiple times (which the landing page does) with the saga accepting multiple requests.

src/amo/store.js
tofumatt: Heh, I feel like it's just how one does constants, so nothing other than convention.

tests/unit/core/sagas/testUtils.js
tofumatt: The way the test was written the entire state was being returned so using all the reducers made it work... I'll take a look again and see about checking a few things. I thought this was less fragile than picking specific reducers which might eventually change.

tests/unit/core/sagas/testSearch.js
tofumatt: You're right, I hastily added them in after adding the checks in the action creator, but they should be in a helper function. Sorry about that.

tofumatt: They actually may both go away in a future patch too, so best to keep 'em in one place so they're easier to refactor out. 👍

tests/unit/core/sagas/testUtils.js
tofumatt: Oh I was silly, yeah there's a fine way to do it with specific reducers. Changed!

tofumatt:
Restored flow and added the helpers to the test files. Also removed adding all the reducers.

tests/unit/core/actions/test_search.js
```diff
@@ -1,15 +1,53 @@
...
+      errorHandlerId: 'Search',
+      filters: { query: 'foo' },
+      page: 1,
+      results: [],
+    });
+  });
+
+  it('throws an error if errorHandlerId is empty', () => {
+    expect(() => {
+      _searchStart({ errorHandlerId: undefined });
```
kumar303: Huh. I didn't even know `undefined` would override the default attributes in the spread but it totally does.

tests/unit/core/sagas/testSearch.js
```diff
@@ -0,0 +1,96 @@
...
+  it('clears the error handler', async() => {
+    const entities = sinon.stub();
+    const result = sinon.stub();
+
+    const filters = { query: 'test' };
+
+    mockApi
+      .expects('search')
+      .once()
+      .withArgs({ api: apiState, auth: authState, filters, page: 1 })
```
kumar303: I don't think you need to make an assertion about the calling arguments here because there is already a test for that. This specific test only needs an API response to set up a condition for clearing the error.

src/core/sagas/utils.js
```diff
@@ -1,19 +1,24 @@
...
 import { ErrorHandler } from 'core/errorHandler';
-import log from 'core/logger';
+import defaultLog from 'core/logger';
 import type { ApiStateType } from 'core/reducers/api';
 
-export function createErrorHandler(id: string): typeof ErrorHandler {
+
+export function createErrorHandler(
+  id: string,
+  { log = defaultLog }: {| log: typeof defaultLog |} = {}
```
tofumatt: There's nothing readable about that syntax to me. 😢 It looks like information overload and it reads like the annotation isn't an addendum to the code/argument but sandwiched in the middle of the default argument declaration. So I'm reading about what the argument is in the code and "bam!" in the middle is annotation before it reverts to code. I think this is the most frustrating part of flow to me: it isn't an annotation as much as it is an interruption.

src/core/sagas/utils.js
```diff
@@ -1,19 +1,24 @@
...
   });
 }
 
-// Convenience function to extract API info.
-export function getApi(state: {| api: ApiStateType |}): ApiStateType {
-  return state.api;
+// Convenience function to extract state info.
+export function getState(
+  state: {| api: ApiStateType, auth: Object |}
+): Object {
```
tofumatt: I find flow's error outputs to be hard to follow as well, but I see what that's doing.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -36,7 +36,12 @@ describe('SearchContextCard', () => {
 
   it('should render "searching" while loading without query', () => {
     const { store } = dispatchClientMetadata();
-    store.dispatch(searchStart({ filters: {} }));
+    store.dispatch(searchStart({
```
tofumatt: True! Good catch.

tests/unit/core/actions/test_search.js
tofumatt: Yeah it's kind of weird but I think it explains well enough what's happening.

tests/unit/core/sagas/testSearch.js
tofumatt: Oh, true, sorry!

src/core/sagas/search.js
```diff
@@ -0,0 +1,41 @@
...
+    const state = yield select(getState);
+
+    const response = yield call(searchApi, {
+      api: state.api,
+      auth: state.auth,
+      filters,
+      page,
+    });
+
+    yield put(searchLoad({ page, filters, ...response }));
```
kumar303: Oops, I just noticed the `...response` spread. Can you change this to explicit parameters? I know this is just copied from the old code but we've had too much trouble with spreads. 

Also, `searchLoad()` doesn't take a `page` argument at all. Flow helps to catch these types of problems but until you're using it, please double check all function signatures manually.

tests/unit/core/sagas/testSearch.js
```diff
@@ -0,0 +1,85 @@
...
+  it('searches the API for add-ons', async() => {
+    const entities = sinon.stub();
+    const result = sinon.stub();
+
+    const filters = { query: 'test' };
+
+    mockApi
+      .expects('search')
+      .once()
+      .withArgs({ api: apiState, auth: authState, filters, page: 1 })
```
kumar303: You don't need to make an assertion about the calling arguments here because there is already a test for that. This specific test only needs an API response to set up a condition for clearing the error. You can delete `withArgs()` on this line.

src/core/sagas/search.js
tofumatt: Totally, makes the code so much nicer anyhow!

tofumatt:
Changes made, set for another r?

kumar303:
...if Travis says so

kumar303:
Thanks for doing this in a separate patch. I'm going to use the saga for converting the Category page (https://github.com/mozilla/addons-frontend/issues/2857). I'll land it so I can get started!

Issue #2859, opened 2017-07-26
Convert LandingPage to load data with a saga


kumar303:
Convert the AMO LandingPage component to load its data using a saga.

To test for regressions:
* Go to the home page
* Click on 'Browse all extensions' (or themes)
* This is served by the LandingPage component. Make sure the page still works.

ValentinaPC:
Verified this on AMO-dev FF54(Win 7).
Both pages are loaded, but for a second the other page is loaded, for example while clicking "Browse all extensions" loads extensions page with themes, and vice-versa.
Logged separate issue for this.
Leaving this unmarked as verified until we figure it out if has something to do with the new issue or not.

PR #2876, opened 2017-07-31
Convert LandingPage to load data with a saga

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2859

The page looks the same as before except now there are loading placeholders and you can also see errors:

screenshot 2017-07-31 12 12 11
screenshot 2017-07-31 12 12 34


src/amo/actions/landing.js
```diff
@@ -1,31 +1,38 @@
...
+    payload: { addonType, errorHandlerId },
   };
 }
 
 export function loadLanding({ addonType, featured, highlyRated, popular }) {
+  if (!addonType) {
+    throw new Error('addonType cannot be empty');
+  }
+  if (!featured) {
+    throw new Error('featured cannot be empty');
```
tofumatt: Do you think adding `Addons` to the end of each variable would help with readability? `featured cannot be empty` is a bit ambiguous to me. 🤷‍♀️

src/amo/components/LandingAddonsCard/index.js
```diff
@@ -13,10 +13,17 @@ export default class LandingAddonsCard extends React.Component {
     footerLink: PropTypes.object.isRequired,
     footerText: PropTypes.string.isRequired,
     header: PropTypes.node.isRequired,
+    loading: PropTypes.boolean,
+  }
+
+  static defaultProps = {
+    loading: false,
```
tofumatt: Why is this needed? Don't we usually default to loading being false in the reducer?

src/amo/components/LandingPage/index.js
```diff
@@ -5,22 +5,22 @@ import PropTypes from 'prop-types';
...
   SEARCH_SORT_TOP_RATED,
 } from 'core/constants';
-import { AddonTypeNotFound } from 'core/errors';
+import { withErrorHandler } from 'core/errorHandler';
 import log from 'core/logger';
 import {
-  apiAddonType as getApiAddonType,
-  safeAsyncConnect,
+  apiAddonType,
+  apiAddonTypeIsValid,
```
tofumatt: Cannot wait for react-router 4 😆

src/amo/components/LandingPage/index.js
```diff
@@ -37,61 +37,88 @@ const ICON_MAP = {
 
 export class LandingPageBase extends React.Component {
   static propTypes = {
-    apiAddonType: PropTypes.func.isRequired,
-    contentForType: PropTypes.func,
+    addonTypeOfResults: PropTypes.string,
```
tofumatt: I would think `addonType` is sufficient because we use that everywhere else.

src/amo/components/LandingPage/index.js
```diff
@@ -37,61 +37,88 @@ const ICON_MAP = {
...
+    errorHandler: PropTypes.object.isRequired,
     featuredAddons: PropTypes.array,
     highlyRatedAddons: PropTypes.array,
+    loading: PropTypes.boolean,
     popularAddons: PropTypes.array,
     i18n: PropTypes.object.isRequired,
     params: PropTypes.objectOf({
       visibleAddonType: PropTypes.string.isRequired,
     }).isRequired,
+    resultsLoaded: PropTypes.bool.isRequired,
```
tofumatt: This could probably be computed from `featuredAddons.length || highlyRatedAddons.length || popularAddons.length` (with a helper function I guess?) rather than needing its own prop.

src/amo/components/LandingPage/index.js
```diff
@@ -214,20 +233,23 @@ export class LandingPageBase extends React.Component {
           footerText={html.featuredFooterText}
           footerLink={html.featuredFooterLink}
           header={html.featuredHeader}
+          loading={loading}
```
tofumatt: I think it's best to do in a future patch but technically these loading states shouldn't be shared; there are three API requests so we could load the data as it came in. I guess we'd want to A/B test that experience but my first thought is we should display data as we load it; if the featured data for some reason takes an extra `400ms` to load we should still display the other stuff, so arguably each request should have its own loading property.

src/amo/components/LandingPage/index.js
```diff
@@ -37,61 +37,88 @@ const ICON_MAP = {
...
-    try {
-      const addonType = apiAddonType(params.visibleAddonType);
-      dispatch(setViewContext(addonType));
-    } catch (err) {
-      if (err instanceof AddonTypeNotFound) {
-        log.info('AddonTypeNotFound in setViewContextType()', err);
-      } else {
-        throw err;
-      }
+  getLandingDataIfNeeded() {
```
tofumatt: Actually, if this function just took a `force = true` argument that made it load the data (which only `componentWillMount` would need here), you wouldn't have to deal with `resultsLoaded` at all. If the component mounts we always need to fetch data, and we only need to fetch new data if its updated and the `addonType` has changed.

Not sure if that's better. I think it's less code/state to manage but let me know what you think.

src/amo/sagas/landing.js
```diff
@@ -0,0 +1,58 @@
...
+      addonType ${addonType}: ${error}`);
+
+    yield put(errorHandler.createErrorAction(error));
+  } finally {
+    yield put(hideLoading());
+  }
+}
+
+export default function* landingSaga() {
+  yield takeEvery(LANDING_GET, fetchLandingAddons);
```
tofumatt: Arguably this could be `take` instead of `takeEvery` which would halt the old request if a user changed from one page to another. But this is probably fine–I think we do `takeEvery` everywhere and I don't think it'll be an issue...

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -17,31 +13,43 @@ import {
...
+      params: { visibleAddonType: visibleAddonType(ADDON_TYPE_EXTENSION) },
+      resultsLoaded: false,
+      ...props,
+    };
+  }
 
-    return findRenderedComponentWithType(renderIntoDocument(
+  function render(props = {}) {
+    return shallow(<LandingPageBase {...renderProps(props)} />, {
+      lifecycleExperimental: true,
```
tofumatt: Hah, this explains why you could use `componentDidUpdate` and have them work with the tests. Neato. 👍

tests/unit/amo/helpers.js
```diff
@@ -120,7 +120,8 @@ export function dispatchSearchResults({
   return { store };
 }
 
-export function createFeaturedApiResult(results) {
-  // Return a normalized result just like core.api.featured() does.
+export function createAddonsApiResult(results) {
```
tofumatt: 👍

tests/unit/amo/reducers/testLanding.js
```diff
@@ -96,7 +93,7 @@ describe('landing reducer', () => {
         },
       };
       const { highlyRated } = landing({
-        ...initialData,
+        ...initialState,
```
tofumatt: 👍

tests/unit/amo/sagas/testLanding.js
```diff
@@ -0,0 +1,123 @@
...
+      apiState = state.api;
+      sagaTester = new SagaTester({
+        initialState: state,
+        reducers: { landing: landingReducer, api: apiReducer },
+      });
+
+      sagaTester.start(landingSaga);
+    });
+
+    function _getLanding(overrides = {}) {
```
tofumatt: This seems to be a common pattern in our tests, wonder if we could write a helper for it... or if it's not needed.

src/amo/components/LandingPage/index.js
kumar303: It cannot because the initial state of the reducer sets all of those to a zero-length array (which does not indicate that the request to load them has been fulfilled). Plus, if we add a new add-on result set then the if statement would need updating which seems clumsy to me.

src/amo/components/LandingPage/index.js
kumar303: I don't think that would handle the case where data was loaded on the server. In that case, `componentWillMount` should do nothing because all of the data has already been loaded.

src/amo/components/LandingPage/index.js
kumar303: Sure, that could be done as an optimization in the future but both the reducer and action creator would need refactoring because right now they 1) retrieve data from the 3 API calls at once and 2) do not store a distinct loading flag for each API result. It will also make the component more complex since it will have to juggle multiple loading flags. 

The important part is that the API requests are made in parallel so at least we have that so far.

tests/unit/amo/components/TestLandingPage.js
kumar303: Yeah, unfortunately it's not documented and the patch to do so is bitrotting :( https://github.com/airbnb/enzyme/pull/1033

tests/unit/amo/sagas/testLanding.js
kumar303: I think we can do a pass soon on all the existing saga tests and come up with some shared helpers. I see a few patterns emerging.

src/amo/components/LandingPage/index.js
kumar303: I wanted to link this specifically to results because there is already `params.visibleAddonType` which is the `addonType` of the URL (i.e. what was requested). It's a bit confusing so I think it helps to tell them apart.

src/amo/components/LandingPage/index.js
kumar303: It might make things easier if we had one component per type of add-on result set (like `FeaturedAddons`, `HighlyRatedAddons`, and `PopularAddons`) then we could give them all their own state/reducer/saga as well and a lot of things would fall into place more naturally. I didn't want to change too much in this patch though.

src/amo/sagas/landing.js
kumar303: Hmm. I'm not sure that `take` is right because it could make the page appear unresponsive for slow requests. It looks like we don't need to handle parallel requests though (and that could lead to confusing results) so I changed it to `takeLatest`. I think that will make the page still feel responsive if there is one hiccup request yet an otherwise snappy API. We'll probably have to tweak these later on after real world usage.

src/amo/components/LandingAddonsCard/index.js
kumar303: Ah, right, I suppose neither of those defaults are needed.

src/amo/components/LandingPage/index.js
tofumatt: Ah, okay.

src/amo/components/LandingPage/index.js
tofumatt: Ahhh, right. I didn't think of that.

kumar303:
Thanks for the review!

PR #2884, opened 2017-08-02
Fixes  Long search term is not wrapped #2880

### LINKED ISSUE #2880: Long search term is not wrapped
Steps to reproduce:
1. Make a search while typing a long search term without spaces (will be quicker on your device) i.e. https://addons.mozilla.org/en-US/android/search/?q=gheriuiergherighsdifghdjgsdklfghdfkjghsdfjksfghsdfgfd

Expected results:

Actual results:
The text is not wrapped and causes a layout issue in the "0 results for search_term" header.

Notes/Issues:
- reproducing on new desktop pages also;

Verified on AMO-prod FF54(Win 7, Android 7.2.1).
Screenshots for this issue:
device
desktop

tsl143:
Fixes #2880 

**Before:**
screen shot 2017-08-02 at 11 58 40 am

**After**
screen shot 2017-08-02 at 11 55 23 am

### DIFF
--- a/src/amo/components/Search/styles.scss
+++ b/src/amo/components/Search/styles.scss
@@ -53,6 +53,7 @@
       grid-row: 4;
     }
   }
+
   .SearchResults-message {
     word-wrap: break-word;
   }
--- a/src/amo/components/Search/styles.scss
+++ b/src/amo/components/Search/styles.scss
@@ -53,4 +53,7 @@
       grid-row: 4;
     }
   }
+  .SearchResults-message {
+    word-wrap: break-word;
+  }
 }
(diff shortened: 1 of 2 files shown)


muffinresearch:
I think this needs a similar rule for the "No results were found for..." part

tsl143:
**Before : message **
screen shot 2017-08-02 at 9 18 52 pm


**After : message**
screen shot 2017-08-02 at 9 19 29 pm

src/amo/components/Search/styles.scss
```diff
@@ -53,4 +53,7 @@
       grid-row: 4;
     }
   }
+  .SearchResults-message {
```
muffinresearch: Needs a newline above.

tsl143:
@muffinresearch done :)

muffinresearch:
> @muffinresearch done :)
But this newline issue was not addressed in lint test, shall we write a lint rule for that?

Yep good call -  I agree it would be better handled by stylelint. If you'd like to file a bug then we can look and see if stylelint has an existing rule for that.

Issue #2886, opened 2017-08-02
Category pages are 404s with content


muffinresearch:
### Describe the problem and steps to reproduce it:

Go to https://addons.mozilla.org/en-US/firefox/extensions/photos-music-videos/

### What happened?

Status code is a 404

<img width="599" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/28871038-9702e320-777a-11e7-9f4b-7562a0baca0f.png">

### What did you expect to happen?

It should be a 200

Issue reproduces on all servers.

muffinresearch:
Downgrading to a p2 since this will affect things like indexation but site usage is not impacted.

kumar303:
Hmm. My patch didn't seem to fix it on dev. It worked locally though. I'll take a closer look.

ValentinaPC:
Verified as fixed on AMO-dev FF54(Win 7).
Checked console for all extensions and themes categories and everything was `200`.
Postfix screenshot:
![200](https://user-images.githubusercontent.com/15685960/28908413-32daa650-782c-11e7-9b50-dc2457072d0b.png)

PR #2891, opened 2017-08-02
Make sure all apps set a default status code

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2886

**The problem:**

The [fix for Category 404s](https://github.com/mozilla/addons-frontend/pull/2684/files#diff-4623346018e51a6970a0e7b3ffb9e325R38) introduced a scenario where the [first render](https://github.com/mozilla/addons-frontend/blob/a1ccd6fde1b87c536046f69b9ec32f3888262e31/src/core/server/base.js#L312-L314) of the app declares a 404 status code but [later](https://github.com/mozilla/addons-frontend/blob/a1ccd6fde1b87c536046f69b9ec32f3888262e31/src/core/server/base.js#L320-L323) it does not. In the final successful render, nothing was declaring a 200 status code.

This bug was probably happening in a few other parts of the app too.

**The solution:**

Both AMO and Discopane will now always declare a 200 status code by default. This will get overidden by error responses because in those cases, the default `App` is not rendered.

We should also stop using `NestedStatus`: https://github.com/mozilla/addons-frontend/issues/2892

### DIFF
--- a/tests/unit/disco/containers/TestApp.js
+++ b/tests/unit/disco/containers/TestApp.js
@@ -1,9 +1,11 @@
 import React from 'react';
+import { shallow } from 'enzyme';
 import { findDOMNode } from 'react-dom';
 import {
   findRenderedComponentWithType,
   renderIntoDocument,
 } from 'react-addons-test-utils';
+import NestedStatus from 'react-nested-status';
 import { Provider } from 'react-redux';
 import { loadFail } from 'redux-connect/lib/store';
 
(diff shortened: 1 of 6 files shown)


tofumatt:
Huh, I didn't know we'd need to do this! Thanks for figuring that out. 👍

kumar303:
It's only needed because we do double rendering, yay 🍰

tofumatt:
Ahhhhhhh! I was wondering if that was related. Clearly time to go "serverless".

kumar303:
@muffinresearch this could probably use a good amount of testing. Do you think we really need to rush out a hotfix for it?

tests/unit/disco/containers/TestApp.js
```diff
@@ -20,14 +22,19 @@ class MyComponent extends React.Component {
   }
 }
 
-function renderApp(extraProps = {}, store = createStore().store) {
-  const props = {
+function renderProps(customProps = {}) {
```
tofumatt: Thanks that's nicer.

tofumatt:
Oh, but Travis doesn't like it 😢

tofumatt:
Looks like there's a flow error re: `react-nested-status`. After that's fixed I'm r+ though, code looks good.

tofumatt:
My $0.02: Given this code feels like it might output a `200` status code when it should actually be another one I'd say the fix is riskier than the bug... I think it's probably okay to wait on the fix.

kumar303:
> this code feels like it might output a 200 status code when it should actually be another one

Yeah, I was worried about that too. I manually tested the real error pages and they are responding with the correct statuses but it could use more testing.

PR #2895, opened 2017-08-03
 Fixes No Read More link for add-ons with long description #2890 

### LINKED ISSUE #2890: No "Read More" link for add-ons with long description
Steps to reproduce:
1. Load https://addons.allizom.org/en-US/firefox/extensions/ page
2. Click on any add-on with long description i.e. Omnibar
3. Observe About this extension section

Expected results:
There is a "Read More" link and a part of description is hidden.

Actual results:
Entire description is displayed. No "Read More" link.

Notes/Issues:
- "Read More" link is displayed after page refresh.
- reproducing also on device.

Verified on AMO-stage(Win 7, Android 7.2.1). Reproducing on all servers.
Video fro this issue:
read more

tsl143:
Fixes #2890


tsl143:
Before
before

After
after

muffinresearch:
This is an aside but I notice the content in the first screenshot shows escaped HTML.

muffinresearch:
This could use a regression test.

tsl143:
I am not sure how to go for regression test, it will help me if you can point me to a link :)

muffinresearch:
@tsl143 the idea would be to have a test that sets up the scenario that fails _without your patch_. That way if something changes in the future that regresses the behaviour the test will fail an we'll know.

tsl143:
Ohkay,
I will try :)

muffinresearch:
@tsl143 let me know if you're stuck - also if you can name your prs related to what they fix it helps us know what's in the review queue at a glance.

tsl143:
@muffinresearch , actually for https://github.com/mozilla/addons-frontend/pull/2895#issuecomment-319917141, the escaped HTML is desired as the API response is following 
screen shot 2017-08-04 at 9 49 15 am


Also I am not sure what to write in test case, as this only occurs if user navigates to page on very first time, once component is mounted it will not occur again.

tsl143:
Also in https://github.com/mozilla/addons-frontend/blob/master/tests/unit/ui/components/TestShowMoreCard.js#L20
it says `we don't have a clientHeight in the tests` so if we already set `{ expanded: false }` this can not be tested.

muffinresearch:
@tsl143 ah ok I see it now it looks like bad data: https://addons-dev.allizom.org/en-US/firefox/addon/clickclean/

muffinresearch:
@tsl143 re: the tests is there not a way to exercise the lifecycle methods to bring about the original issue so you can get a failing test (without your patch).

tsl143:
When `componentDidMount` is executed `contents.clientHeight` = `20` (may be it didn't receive the API response hence no props received), hence it doesn't show read more . So I changed `componentDidMount` to `componentWillReceiveProps`, whenever the response is received, `contents.clientHeight` given actual content height.
Now I am not sure how to write a fail test case for this.

muffinresearch:
> When componentDidMount is executed contents.clientHeight = 20 (may be it didn't receive the API response hence no props received), hence it doesn't show read more . So I changed componentDidMount to componentWillReceiveProps, whenever the response is received, contents.clientHeight given actual content height.
Now I am not sure how to write a fail test case for this.

Ah I think I understand this better now - thanks for clarifying the issue. So it's not a case of the truncation not being called it was just being called too early.

tsl143:
@muffinresearch You are welcome :)
So is it good to merge now?

muffinresearch:
So I think that minimizes what can be done with testing in the way I was originally thinking.

I have a concern that changing moving where this happens in the life-cycle might not fix this in the future but this is still probably better than nothing unless we try to change to something based on content (e.g. char counts) - although that gets difficult on the basis of viewport widths.

I think the only thing left to do is have a test covering that the function is called (unless we already have that) when props are received and it may also be worth checking that we're not calling the truncation more times than needed: 

> Note that React may call this method even if the props have not changed, so make sure to compare the current and next values if you only want to handle changes. This may occur when the parent component causes your component to re-render.

From https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops

tsl143:
@muffinresearch , really happy to do this specific bug, got a much better insights about test cases :)
Also I have used `/* eslint-env browser */` first time to stop linting errors due to native functions.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -57,4 +58,19 @@ describe('<ShowMoreCard />', () => {
     const rootNode = findDOMNode(root);
     expect(rootNode.textContent).toContain('Hello I am description');
   });
+
+  it('executes truncateToMaxHeight when it recieves props', () => {
+    const node = document.createElement('div');
+    let ref;
+    ReactDOM.render(<ShowMoreCardBase i18n={getFakeI18nInst()} ref={(e) => { ref = e; }} />, node,
+      () => {
+        jest.spyOn(ref, 'truncateToMaxHeight');
```
muffinresearch: We use sinon for spies since the API is a little bit more complete and intuitive.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -57,4 +58,19 @@ describe('<ShowMoreCard />', () => {
...
+        ReactDOM.render(<ShowMoreCardBase
```
muffinresearch: What's the rationale for rendering twice?  Couldn't the spy be setup and check based on the first render?

tsl143: The component has been rerendered to call the function again. The reference of first call is bound to spy and then it can actually spy the method on second call.
I looked various ways to achieve this but found nothing better for redux component.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -57,4 +58,19 @@ describe('<ShowMoreCard />', () => {
     const rootNode = findDOMNode(root);
     expect(rootNode.textContent).toContain('Hello I am description');
   });
+
+  it('executes truncateToMaxHeight when it recieves props', () => {
+    const node = document.createElement('div');
+    let ref;
+    ReactDOM.render(<ShowMoreCardBase i18n={getFakeI18nInst()} ref={(e) => { ref = e; }} />, node,
```
kumar303: This test is a bit too fragile. Testing the `componentWillReceiveProps()` method is easier with [enzyme](https://github.com/airbnb/enzyme):

```js
import { shallow } from 'enzyme';

const root = shallow(<ShowMoreCardBase i18n={getFakeI18nInst()} />);
const component = root.instance();

const contentNode = ReactDOM.findDOMNode(instance.contents);
const truncateToMaxHeight = sinon.spy(component, 'truncateToMaxHeight');
root.setProps(); // simulate any kind of update to properties

sinon.assert.calledWith(truncateToMaxHeight, contentNode);
```

We already have enzyme but this suite hasn't been converted yet. Try the code above and let me know if it works.

tsl143: Got an error `AssertError: expected truncateToMaxHeight to be called with arguments `
Also replaced `ReactDOM.findDOMNode(instance.contents);` with `ReactDOM.findDOMNode(component.contents);` as `instance` not defined

tsl143:
@kumar303 , so `mount` did the trick. with `shallow` the `contentNode` was null so got that error.
Thanks for the guidance :)

PR #2896, opened 2017-08-03
Static pages - about us and review guide #2783

### LINKED ISSUE #2783: Add static pages
Related to #2748, we need new static pages on the frontend. Should be easy!

- [ ] About
- [ ] Rating Policy

tsl143:
Fixes #2783 
About us page


src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,113 @@
...
+                <a href="https://wiki.mozilla.org/Add-ons/Reviewers/Applying">{i18n.gettext('become an Add-on Reviewer')}</a>!&nbsp;
+                {i18n.gettext('Our reviewers are add-on fans with a technical background who review add-ons for code quality and stability.')}
+              </li>
+              <li>
+                {i18n.gettext('Help improve this website. It\'s open source, and you can file bugs and submit patches.')}&nbsp; <a href="https://github.com/mozilla/olympia/issues"> {i18n.gettext('GitHub')}</a>&nbsp;
+                {i18n.gettext('contains all of our current bugs, legacy bugs can still be found in Bugzilla.')}
+              </li>
+            </ul>
+          </p>
+          <p>{i18n.gettext('If you\'re interested in add-ons but not quite as technical, there are still ways to help:')}</p>
```
muffinresearch: You can use a template string e.g: `` `Blah` `` for these strings to avoid needing to escape the quotes. Be sure not to use any template tags though since our extraction tool will automatically fix leading indents with just the straight template tag.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,113 @@
...
+                <Link to={'/developers'}>{i18n.gettext('Make your own add-on')}</Link>.&nbsp;
+                {i18n.gettext('We provide free hosting and update services and can help you reach a large audience of users.')}
+              </li>
+              <li>
+                {i18n.gettext('If you have add-on development experience,')}&nbsp;
+                <a href="https://wiki.mozilla.org/Add-ons/Reviewers/Applying">{i18n.gettext('become an Add-on Reviewer')}</a>!&nbsp;
+                {i18n.gettext('Our reviewers are add-on fans with a technical background who review add-ons for code quality and stability.')}
+              </li>
+              <li>
+                {i18n.gettext('Help improve this website. It\'s open source, and you can file bugs and submit patches.')}&nbsp; <a href="https://github.com/mozilla/olympia/issues"> {i18n.gettext('GitHub')}</a>&nbsp;
```
muffinresearch: These two lines could be one, and a placeholder should be used to substitute in the link.

muffinresearch:
Please can you add a screenshot of what this looks like?

tsl143:
Fixed the escape string with template strings, but lint is throwing errors
`  29:27  error  Strings must use singlequote  quotes`
  `39:29  error  Strings must use singlequote  quotes`
  `62:28  error  Strings must use singlequote  quotes`
  `74:31  error  Strings must use singlequote  quotes`
  `79:28  error  Strings must use singlequote  quotes`
  `94:27  error  Strings must use singlequote  quotes`

Following is screenshot of page
screen shot 2017-08-03 at 14 24 53-fullpage

muffinresearch:
@tsl143 ok - that's an edge case our lint config doesn't allow for - I've submitted this PR and we'll update it to allow for that. https://github.com/mozilla/eslint-config-amo/pull/17

tsl143:
Also created review_guide page and updated testcases for both pages.

screen shot 2017-08-05 at 11 50 44-fullpage

tsl143:
@muffinresearch also this.

muffinresearch:
@tsl143 you'll need to rebase on master to clear the lint errors. I'll take a look at this today.

tsl143:
I did `git pull upstream master`, still linting errors

muffinresearch:
@tsl143 sorry about that, the versioning of devDeps means the updated eslint config isn't used yet. An incoming patch will fix that shortly.

tsl143:
Oh okay, np :)

muffinresearch:
@tsl143 that dep should be updated now.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+export class AboutBase extends React.Component {
+  static propTypes = {
+    i18n: PropTypes.object.isRequired,
+  }
+
+  render() {
+    const { i18n } = this.props;
+
+    const helpText = i18n.sprintf(i18n.gettext(`Help improve this website. It's open source, and you can file bugs and submit patches. <a href="%(url)s">GitHub</a> contains all of our current bugs, legacy bugs can still be found in Bugzilla.`),
```
muffinresearch: You can use newlines to keep this more readable. 

We'd ideally like to keep HTML out of the translation strings as much as possible. If the text in the link isn't translatable then you can use a placeholder for the entire link.

If the strings are translatable then we'd prefer to use a start and end tag placeholder. See https://addons-frontend.readthedocs.io/en/latest/i18n/#guidance-on-html-in-translations for examples.

muffinresearch:
Anywhere there's something that looks to be one chunk of text when rendered should be a single translation and not split into multiple gettext calls.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,111 @@
...
+          </ul>
+        </section>
+        <section>
+          <h2>{i18n.gettext('I have a question')}</h2>
+          <p>{i18n.gettext('A good place to start is our')}&nbsp;
+            <Link to="/faq"><abbr title="{i18n.gettext('Frequently Asked Questions')}">{i18n.gettext('FAQ')}</abbr></Link>.&nbsp;
+            {i18n.gettext(`If you don't find an answer there, you can`)}&nbsp;<a href="https://discourse.mozilla-community.org/c/add-ons">{i18n.gettext('ask on our forums')}</a>.
+          </p>
+          <p>{i18n.gettext('If you really need to contact someone from the Mozilla team, please see our')}&nbsp;
+            <Link to="/developers/docs/policies/contact">{i18n.gettext('contact information')}</Link> {i18n.gettext('page.')}
```
muffinresearch: You could save a redirect and just point to `https://developer.mozilla.org/en-US/Add-ons#Contact_us` for this - also avoids the complication with `Link`.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,111 @@
...
+            </li>
+            <li dangerouslySetInnerHTML={sanitizeHTML(forumText, ['a'])} />
+          </ul>
+        </section>
+        <section>
+          <h2>{i18n.gettext('This sounds great! How can I get involved?')}</h2>
+          <p>{i18n.gettext(`There are plenty of ways to get involved. If you're on the technical side:`)}
+            <ul>
+              <li>
+                <Link to={'/developers'}>{i18n.gettext('Make your own add-on')}</Link>.&nbsp;
```
muffinresearch: This is probably one of the few places we can probably live with splitting up the translations since I don't think we can really mix jsx inside a gettext call.  If there are other ideas on how we could deal with that better they'd be welcomed.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,111 @@
...
+              <a href="https://discourse.mozilla-community.org/c/add-ons">{i18n.gettext('forums')}</a>.
+            </li>
+            <li>
+              {i18n.gettext('Review add-ons on the site. Add-on authors are more likely to improve their add-ons and write new ones when they know people appreciate their work.')}
+            </li>
+          </ul>
+        </section>
+        <section>
+          <h2>{i18n.gettext('I have a question')}</h2>
+          <p>{i18n.gettext('A good place to start is our')}&nbsp;
```
muffinresearch: Is the non-breaking space necessary?

src/amo/components/StaticPages/ReviewGuide/index.js
```diff
@@ -0,0 +1,79 @@
...
+            <li>{i18n.gettext('Include HTML, links, source code or code snippets. Reviews are meant to be text only.')}</li>
+            <li>{i18n.gettext('Make false statements, disparage add-on authors or personally insult them.')}</li>
+            <li>{i18n.gettext('Include your own or anyone else’s email, phone number, or other personal details.')}</li>
+            <li>{i18n.gettext('Post reviews for an add-on you or your organization wrote or represent.')}</li>
+            <li>{i18n.gettext('Criticize an add-on for something it’s intended to do. For example, leaving a negative review of an add-on for displaying ads or requiring data gathering, when that is the intended purpose of the add-on, or the add-on requires gathering data to function.')}</li>
+          </ul>
+        </section>
+        <section>
+          <h2>{i18n.gettext('Frequently Asked Questions about Reviews')}</h2>
+          <p className="staticSubHeading">{i18n.gettext('How can I report a problematic review?')}</p>
```
muffinresearch: A `dl` would probably be a better fit for this content.

src/amo/routes.js
```diff
@@ -30,6 +32,8 @@ import ServerError from './components/ErrorPage/ServerError';
 export default (
   <Route path="/:lang/:application" component={App}>
     <IndexRoute component={Home} />
+    <Route path="about" component={About} />
+    <Route path="review_guide" component={ReviewGuide} />
```
muffinresearch: It's preferable to use a hyphen instead of underscore in a url - sometimes underscores get lost in underlines and are less obvious.

tests/unit/amo/components/TestStaticPages.js
```diff
@@ -0,0 +1,39 @@
...
+import { AboutBase } from 'amo/components/StaticPages/About';
+import { ReviewGuideBase } from 'amo/components/StaticPages/ReviewGuide';
+import { getFakeI18nInst } from 'tests/unit/helpers';
+
+
+describe('About', () => {
+  function render() {
+    return shallow(
+      <AboutBase
+        i18n={getFakeI18nInst()}
```
muffinresearch: Unnecessary indentation

tests/unit/amo/components/TestStaticPages.js
```diff
@@ -0,0 +1,39 @@
...
+
+    expect(root).toHaveClassName('AboutPage');
+  });
+});
+
+describe('ReviewGuide', () => {
+  function render() {
+    return shallow(
+      <ReviewGuideBase
+        i18n={getFakeI18nInst()}
```
muffinresearch: Unnecessary indentation

tsl143:
All changes have been implemented except
- New line for const variable works but not for JSX (unknown character error)
- `Link` translation still no idea so left as it is
- `non-breaking space ` still at few places as `getText` trims space.
- Used `hyphen` for `review-guide` but the URL at legacy website has `review_guide`

muffinresearch:
> Used hyphen for review-guide but the URL at legacy website has review_guide

Good catch - it'll be necessary to add a route to 301 that.

src/amo/components/StaticPages/About/index.js
```diff
@@ -52,8 +62,8 @@ export class AboutBase extends React.Component {
           <ul>
             <li dangerouslySetInnerHTML={sanitizeHTML(blogText, ['a'])} />
             <li>
-              {i18n.gettext('We often post news, tips, and tricks to our Twitter account,')}&nbsp;
-              <a href={'http://twitter.com/mozamo'}>{i18n.gettext('mozamo')}</a>
+              {i18n.gettext('We often post news, tips, and tricks to our Twitter account, ')}&nbsp;
+              <a href={'http://twitter.com/mozamo'}>mozamo</a>
```
muffinresearch: This should be a substitution, rather than implicit concatenation, as a translation's word order could be different.

muffinresearch: Lets' remove the `&nbsp;` too.

src/amo/components/StaticPages/About/index.js
```diff
@@ -63,23 +73,21 @@ export class AboutBase extends React.Component {
...
+                <Link to={'/developers'}>{i18n.gettext('Make your own add-on')}</Link>
+                {i18n.gettext('. We provide free hosting and update services and can help you reach a large audience of users.')}
               </li>
               <li>
-                {i18n.gettext('If you have add-on development experience,')}&nbsp;
-                <a href="https://wiki.mozilla.org/Add-ons/Reviewers/Applying">{i18n.gettext('become an Add-on Reviewer')}</a>!&nbsp;
-                {i18n.gettext('Our reviewers are add-on fans with a technical background who review add-ons for code quality and stability.')}
+                {i18n.gettext('If you have add-on development experience, ')}&nbsp;
+                <a href="https://wiki.mozilla.org/Add-ons/Reviewers/Applying">{i18n.gettext('become an Add-on Reviewer')}</a>
+                {i18n.gettext('! Our reviewers are add-on fans with a technical background who review add-ons for code quality and stability.')}
```
muffinresearch: The `!` looks like a typo?

src/amo/components/StaticPages/About/index.js
```diff
@@ -63,23 +73,21 @@ export class AboutBase extends React.Component {
...
-                <Link to={'/developers'}>{i18n.gettext('Make your own add-on')}</Link>.&nbsp;
-                {i18n.gettext('We provide free hosting and update services and can help you reach a large audience of users.')}
+                <Link to={'/developers'}>{i18n.gettext('Make your own add-on')}</Link>
+                {i18n.gettext('. We provide free hosting and update services and can help you reach a large audience of users.')}
               </li>
               <li>
-                {i18n.gettext('If you have add-on development experience,')}&nbsp;
-                <a href="https://wiki.mozilla.org/Add-ons/Reviewers/Applying">{i18n.gettext('become an Add-on Reviewer')}</a>!&nbsp;
-                {i18n.gettext('Our reviewers are add-on fans with a technical background who review add-ons for code quality and stability.')}
+                {i18n.gettext('If you have add-on development experience, ')}&nbsp;
```
muffinresearch: This section should be all one block of `gettext` with the link a substitution via `sprintf` using a start and end placeholder as done elswhere.

src/amo/components/StaticPages/ReviewGuide/index.js
```diff
@@ -62,11 +62,11 @@ export class ReviewGuideBase extends React.Component {
         </section>
         <section>
           <h2>{i18n.gettext('Frequently Asked Questions about Reviews')}</h2>
-          <p className="staticSubHeading">{i18n.gettext('How can I report a problematic review?')}</p>
+          <dl className="staticSubHeading">{i18n.gettext('How can I report a problematic review?')}</dl>
```
muffinresearch: A `dl` needs a `dt` and `dd` for each item within it.

src/amo/components/StaticPages/About/index.js
```diff
@@ -18,12 +18,22 @@ export class AboutBase extends React.Component {
   render() {
     const { i18n } = this.props;
 
-    const helpText = i18n.sprintf(i18n.gettext(`Help improve this website. It's open source, and you can file bugs and submit patches. <a href="%(url)s">GitHub</a> contains all of our current bugs, legacy bugs can still be found in Bugzilla.`),
-      { url: 'https://github.com/mozilla/olympia/issues' });
-    const blogText = i18n.sprintf(i18n.gettext(`Our <a href="%(url)s">Add-ons Blog</a> is regularly updated with information for both add-on enthusiasts and developers.`),
-      { url: 'http://blog.mozilla.com/addons/' });
-    const forumText = i18n.sprintf(i18n.gettext(`Our <a href="%(url)s">forums</a> are a good place to interact with the add-ons community and discuss upcoming changes to AMO.`),
-      { url: 'https://discourse.mozilla-community.org/c/add-ons' });
+    const helpText = i18n.sprintf(i18n.gettext(`Help improve this website.
```
muffinresearch: For context, and since there's a lot of these, I think it might be helpful to put them inline rather than have these intermediate `const`s.

tsl143:
@muffinresearch , sorry for not replying early I was on vacation (extended weekend :) )
I updated the changes. For https://github.com/mozilla/addons-frontend/pull/2896#discussion_r132923640
There are lots of `const`s but putting them inline makes code a bit unreadable and shabby, can we leave it as it is?

tsl143:
Ping @muffinresearch

muffinresearch:
> @muffinresearch , sorry for not replying early I was on vacation (extended weekend :) )
I updated the changes. For #2896 (comment)
There are lots of consts but putting them inline makes code a bit unreadable and shabby, can we leave it as it is?

Whilst I see where you're coming from, I think that keeping the context is really helpful for ensuring the end result makes sense.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,140 @@
...
+            />
+            <li>
+              {i18n.gettext('Review add-ons on the site. Add-on authors are more likely to improve their add-ons and write new ones when they know people appreciate their work.')}
+            </li>
+          </ul>
+        </section>
+        <section>
+          <h2>{i18n.gettext('I have a question')}</h2>
+          <p>{i18n.gettext('A good place to start is our')}&nbsp;
+            <Link to="/faq"><abbr title="{i18n.gettext('Frequently Asked Questions')}">FAQ</abbr></Link>
```
muffinresearch: I don't think this page exists yet either.

src/amo/routes.js
```diff
@@ -30,6 +32,8 @@ import ServerError from './components/ErrorPage/ServerError';
 export default (
   <Route path="/:lang/:application" component={App}>
     <IndexRoute component={Home} />
+    <Route path="about" component={About} />
+    <Route path="review-guide" component={ReviewGuide} />
```
muffinresearch: This router config will result in the about page being under `/en-US/firefox/about` which will also mean there's a `/en-US/android/about` too which would result in duplicate content.

It needs to use purely a locale prefix so that you end up with: `/en-US/about` without the `clientApp` (firefox/android) part. 

This will probably require changes to the routes to have more than one `Route` component.

We also still a redirect for `review_guide` to `review-guide`.

src/amo/components/StaticPages/ReviewGuide/index.js
```diff
@@ -0,0 +1,102 @@
...
+                    <li>{i18n.gettext('What features did you like or dislike?')}</li>
+                    <li>{i18n.gettext('Was it useful?')}</li>
+                    <li>{i18n.gettext('Was it easy to use?')}</li>
+                    <li>{i18n.gettext('Will you continue to use this add-on?')}</li>
+                  </ul>
+                </li>
+                <li>{i18n.gettext('Take a moment to read your review before submitting it to minimize typos.')}</li>
+              </ul>
+            </dd>
+          </dl>
```
muffinresearch: You can remove the closing dl and opening dl and have a single definition list rather than one per question.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,140 @@
...
+            {i18n.gettext(`. If you don't find an answer there, you can`)}&nbsp;<a href="https://discourse.mozilla-community.org/c/add-ons">{i18n.gettext('ask on our forums')}</a>.
```
muffinresearch: This should make the link part of the translation as done elsewhere otherwise the context is split up and that might make it hard to localise.

muffinresearch: The non-breaking space should be removed too.

src/amo/components/StaticPages/About/index.js
```diff
@@ -0,0 +1,140 @@
...
+            {
+              start_link: '<a href="https://developer.mozilla.org/en-US/Add-ons#Contact_us">',
+              end_link: '</a>',
+            }), ['a'])}
+          />
+        </section>
+        <section>
+          <h2>{i18n.gettext('Who works on this website?')}</h2>
+          <p>{i18n.gettext('Over the years, many people have contributed to this website, including both volunteers from the community and a dedicated AMO team. A list of significant contributors can be found on our ')}&nbsp;
+            <Link to="/pages/credits">{i18n.gettext('Site Credits')}</Link>&nbsp;{i18n.gettext(' page')}.
```
muffinresearch: @tofumatt, is there some equivalent JS API we could use to derive links so that we don't need to split up strings that use `Link`?

src/amo/components/StaticPages/About/index.js
tsl143: `i18n.gettext` trims the start and end spaces, so I need to put `&nbsp;`

src/amo/routes.js
willdurand: Maybe this could be the way to configure the new routes:

```
diff --git a/src/amo/routes.js b/src/amo/routes.js
index af60673..6b4e912 100644
--- a/src/amo/routes.js
+++ b/src/amo/routes.js
@@ -32,8 +32,8 @@ import ServerError from './components/ErrorPage/ServerError';
 export default (
   <Route path="/:lang/:application" component={App}>
-    <Route path="about" component={About} />
-    <Route path="review-guide" component={ReviewGuide} />
+    <Route path="/:lang/about/" component={About} />
+    <Route path="/:lang/review-guide/" component={ReviewGuide} />
     <Route path=":visibleAddonType/categories/" component={CategoriesPage} />
```

muffinresearch:
@tsl143 I'm going to take this patch and make a few updates to get it landed. In doing so it's blocked by https://github.com/mozilla/addons-frontend/issues/3162 for the route changes so that'll need resolving first. Apologies for this taking a long time.

src/amo/routes.js
tsl143: hi @willdurand, 
`<Route path="/:lang/about/" component={About} />`
`<Route path="/:lang/review-guide/" component={ReviewGuide} />`
doesn't helps
On `http://localhost:3000/en-GB/about/`, I get this
`This page does not exist in addons-frontend. Returning 404; this error should trigger upstream (usually addons-server) to return a valid response.`

tsl143:
@muffinresearch , that will help :)
It has been actually long time.

src/amo/components/StaticPages/About/index.js
muffinresearch: This isn't needed anymore since the FAQ page and credits pages are being dropped.

src/amo/routes.js
muffinresearch: See https://github.com/mozilla/addons-frontend/issues/3162 which is a blocker.

muffinresearch:
Closing in favour of #3207

Issue #1894, opened 2017-02-24
Navigating from one page to another in extensions categories, shows “No results were found” before loading the content on the page


raresmsv:
# Environment
Firefox: Nightly 54.0a1
Mobile device: Samsung S7 (Android 6.0.1)

# Steps to reproduce
1. Go to AMO mobile –dev homepage.
2. Tap on an extensions category (make sure the content is spread over multiple pages).
3. Navigate to the next page (tap Next at the bottom the list) and observe the behavior until the page is loaded.

# Expected results
A loading animation (loading spinner) indicates the content is loading.

# Actual results
“No results were found” message is displayed for a few moments (up to 3 seconds), then the page is loaded.

# Notes
Reproducibility: 5/5.

gif
![no results were found message](https://cloud.githubusercontent.com/assets/26005101/23366817/cef582cc-fd10-11e6-8b84-bb96cc0799bc.gif)

ValentinaPC:
This is happening on new desktop pages also.
Please see video:
![no results](https://user-images.githubusercontent.com/15685960/29114574-72e4469a-7cfd-11e7-9c90-c2802e66cee8.gif)

kumar303:
whoops, this is a regression from https://github.com/mozilla/addons-frontend/issues/2859 Thanks. I'll fix this.

tofumatt:
I think it might be covered by the sagas, may not be needed. Might want to wait until we merge those unless you feel it’s worth a hotfix for.

- tofumatt

kumar303:
ok, I'll try it in your branch first

kumar303:
It's still happening in your branch and actually it's broken, heh (page 2 loads for a minute then unloads). I'll keep an eye on it after reviewing your patch.

tofumatt:
Oh shoot, right. That is a bug and I forgot about it. I'll mark the saga branch as WIP again then. My bad.

kumar303:
This was fixed in https://github.com/mozilla/addons-frontend/pull/2909 I didn't close it earlier because we had other unrelated bugs with this page that would have made QA verification difficult (those are also fixed now).

ValentinaPC:
Verified this as fixed on AMO-dev FF55(Win 7).
Navigation through categories pages results display loading indicators with no other issues.
Postfix video:
![nav](https://user-images.githubusercontent.com/15685960/29453551-b0f29484-8412-11e7-86fd-777afd3cecd0.gif)

PR #2909, opened 2017-08-06
feat: Use sagas for search

### LINKED ISSUE #2665: Empty type query param in search from homepage
When searching from the home page, an empty `type` query param is added to the URL. See: https://github.com/mozilla/addons-frontend/issues/2513#issuecomment-312585463

### What happened?
An empty `type` param is in the URL.

No type param is set when it's empty.

### LINKED ISSUE #2737: Convert search to load data with a saga

tofumatt:
* Fixes #2665
* Fixes #2737

Will likely fix other things too; this moves to use sagas for search and tidies up a fair bit of search code in the process.


tofumatt:
This gets rid of the weird `hasSearchParams` props we used to use and puts search information only in state, instead of the weird state + location medley we had in the past.

I think I maintained test coverage but this was a bit of a "touching code in a lot of places" patch and maybe I missed something. The coverage is maintained but let me know if I missed something.

tofumatt:
Sorry, forgot about https://github.com/mozilla/addons-frontend/issues/1894#issuecomment-321319395, will ask for review again once it's fixed.

tofumatt:
Tests failing here are just lint, I'll fix these after dinner.

tests/unit/core/reducers/test_search.js
```diff
@@ -18,16 +18,16 @@ describe('search reducer', () => {
...
-          filters: { query: 'bar' },
-          loading: false,
-          results: [{ slug: 'bar' }],
-        },
-        { type: 'SEARCH_STARTED', payload: { filters: { query: 'foo' } } });
+      const state = search({
+        filters: { query: 'bar' },
+        loading: false,
+      }, {
+        type: 'SEARCH_STARTED',
```
kumar303: this should be calling `searchStart()`.

tests/unit/core/test_searchUtils.js
```diff
@@ -1,100 +1,64 @@
...
-import { ADDON_TYPE_THEME } from 'core/constants';
-import { loadByCategoryIfNeeded, mapStateToProps } from 'core/searchUtils';
+import {
+  isLoaded,
+  parsePage,
+} from 'core/searchUtils';
 
 
-describe('searchUtils mapStateToProps()', () => {
+describe('Search.isLoaded()', () => {
```
kumar303: I know we're lax about module level descriptions but I'm not sure what `Search` is in this context. Can it be something that will be more helpful when viewed in the console? I think `core/searchUtils/isLoaded()` would be helpful but `isLoaded` would be fine too. The problem is that jest doesn't always provide useful stack traces (e.g. wonky event loop stuff).

tests/unit/core/test_searchUtils.js
```diff
@@ -1,100 +1,64 @@
...
   const state = {
-    api: { lang: 'fr-CA' },
-    search: {
-      filters: { clientApp: 'firefox', query: 'foo' },
-      hasSearchParams: true,
-    },
+    page: 2,
+    filters: { query: 'ad-block' },
+    loading: false,
+    results: [{ slug: 'ab', name: 'ad-block' }],
```
kumar303: Since you already have:

https://github.com/mozilla/addons-frontend/blob/e43b9acdca9d903aab38310aa1764a3b8411c68e/tests/unit/amo/helpers.js#L103

It might not be hard to convert this test suite to dispatched/reduced state rather than hard-coded state. If it starts taking a lot of time then don't worry about it.

tofumatt:
Thanks a tonne for the review. Sorry I didn't fix the lint errors yet but I'll get them when I fix the address your comments. This wasn't a big Patch in terms of lines of code but it touched a lot of stuff--thanks for catching a lot of little things I missed after looking at it for so long 👍

src/amo/components/Category/index.js
```diff
@@ -46,7 +44,7 @@ export class CategoryBase extends React.Component {
       <div className="Category">
         {errorHandler.hasError() ? errorHandler.renderError() : null}
         <CategoryHeader category={category} />
-        <SearchBase enableSearchSort={false} hasSearchParams {...searchProps} />
+        <Search enableSearchSort={false} {...searchProps} />
```
tofumatt: Good call, I was lazy.

src/amo/components/Search/index.js
```diff
@@ -40,12 +40,29 @@ export class SearchBase extends React.Component {
...
+
+    if (!deepEqual(oldFilters, newFilters)) {
+      this.dispatchSearch({ filters: newFilters });
+    }
+  }
+
+  dispatchSearch({ filters = {} } = {}) {
+    const { dispatch } = this.props;
+
+    if (hasSearchFilters(filters)) {
```
tofumatt: The intent was not to dispatch the search if there aren't enough filters for a (meaningful) search. I'll see about a test and also moving the logic here, maybe there's a move obvious place for it.

kumar303: ok, then maybe you can combine this if statement with the `!deepEqual(...)` one above? It would be nice to see the conditional logic all in one place (that was the source of my confusion). As mentioned in previous comments, I think you will need to do the `!deepEqual(...)` every time anyway, e.g. to handle the server side rendering case, so you might as well combine them.

src/amo/components/Search/index.js
```diff
@@ -40,12 +40,29 @@ export class SearchBase extends React.Component {
...
+      dispatch(searchStart({ errorHandlerId: 'Search', filters }));
```
tofumatt: Shoot, right!

src/amo/components/Search/index.js
```diff
@@ -60,49 +77,66 @@ export class SearchBase extends React.Component {
       LinkComponent,
       count,
       enableSearchSort,
+      errorHandler,
```
tofumatt: Indeed, sorry I missed it but my lint caught it after and it's been added!

src/amo/components/Search/index.js
```diff
@@ -60,49 +77,66 @@ export class SearchBase extends React.Component {
...
       filters,
-      hasSearchParams,
       loading,
-      page,
+      paginationQueryParams,
```
tofumatt: The issue with that is it needlessly pollutes the query params for things like category pages, because filters like `addonType` and `category` already exist elsewhere in the URL. Without explicitly specifying `paginationQueryParams` category navigating from the first to second page does this:

```
# First page's URL
/en-GB/firefox/feeds-news-blogging/
# Second page's URL
/en-US/firefox/extensions/feeds-news-blogging/?app=firefox&category=feeds-news-blogging&page=2&type=extension
```

The paginator will now (by default/with no `paginationQueryParams`) using the `filters` it's supplied to do exactly what you suggested, but this way we can make slightly nicer/less-repeated URLs.

tofumatt: This is a bit odd though when first looked at though, agreed. I'll add some comments to it.

src/amo/components/Search/index.js
```diff
@@ -60,49 +77,66 @@ export class SearchBase extends React.Component {
...
+
+    const page = parsePage(filters.page);
+
+    // If paginator params aren't specified, we convert filters instead.
+    const queryParams = paginationQueryParams ||
       convertFiltersToQueryParams(filters);
-    const paginator = count > 0 && hasSearchParams ? (
+
+    // TODO: Make the count need to be higher than one page's worth of
+    // add-ons.
```
tofumatt: It should be a separate bug, I'll file. It would be nice not to show the paginator if there are 23 add-ons (eg both prev/next pages will be disabled) but the `page_size` is 25.

src/amo/components/Search/index.js
```diff
@@ -60,49 +77,66 @@ export class SearchBase extends React.Component {
...
+  return {
+    count: state.search.count,
+    loading: state.search.loading,
+    results: state.search.results,
+  };
+}
+
 export default compose(
-  safeAsyncConnect([{ promise: loadSearchResultsIfNeeded }]),
+  withErrorHandling({ name: 'Search' }),
```
tofumatt: Ah. Those functions are *super* similarly named, it might be better to make them a bit easier to tell apart.

I've filed: #2931

src/amo/components/Search/index.js
tofumatt: I've filed #2932 and I think can just remove this comment actually.

src/amo/components/SearchPage/index.js
```diff
@@ -0,0 +1,18 @@
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import Search from 'amo/components/Search';
+import { convertQueryParamsToFilters } from 'core/searchUtils';
+
+
+export function mapLocationToProps(state, ownProps) {
```
tofumatt: Ah, fair enough especially for grep-ing. This function's intent is more to grab location info but that's a good point. 👍

src/amo/components/Search/index.js
tofumatt: My bad, the paginator itself doesn't do that as it's too generic, but the search component will. So for certain searches we just use the filters, but here we specify which ones to include.

src/core/reducers/search.js
```diff
@@ -1,34 +1,35 @@
...
+        filters: payload.filters,
+        results: [],
+        loading: true,
+      };
     case SEARCH_LOADED:
       return {
         ...state,
         count: payload.result.count,
         loading: false,
-        filters: payload.filters,
```
tofumatt: I thought that too, but I didn't like dispatching the same data again when it wouldn't have changed (extra tests, code, for nothing really). If we change the search to allow multiple searches we'd not only need to change the saga but also our search code to hold multiple sets of results for different filters. So I figured it was better to remove it as they shouldn't change.

Writing a test to ensure filters haven't been overwritten is essentially writing our own test for saga's `takeLatest` which would be redundant and a bit messy/verbose, but for what it's worth in local testing the `takeLatest` saga definitely prevents old results overwriting new ones.

src/core/sagas/search.js
```diff
@@ -18,7 +19,8 @@ export function* fetchSearchResults({ payload }) {
   yield put(errorHandler.createClearingAction());
 
   try {
-    const { filters, page } = payload;
+    const { filters } = payload;
+    const page = parsePage(filters.page);
```
tofumatt: I think this would be better set in the reducer though, actually. I'll make sure there's test coverage for it but I think it should be there rather than in the saga.

tests/unit/amo/components/TestSearch.js
```diff
@@ -73,28 +77,39 @@ describe('<Search />', () => {
     expect(sort.prop('pathname')).toEqual(props.pathname);
   });
 
-  it('does not render a SearchSort when there are no filters', () => {
-    const root = render({ hasSearchParams: false, results: [] });
+  it('does not render a SearchSort when there are no results', () => {
+    const root = render({ count: 0 });
```
tofumatt: Sounds good, search is important and it'd be great to get it using real dispatches 👍

src/core/sagas/search.js
tofumatt: Wait, my bad, this should be here but yeah I'll add coverage.

tests/unit/core/test_searchUtils.js
tofumatt: Agreed, though in this case the test could actually be deleted. Woohoo!

tests/unit/core/test_searchUtils.js
tofumatt: Actually this suite was outdated, but yeah I'm always for converting stuff.

tests/unit/amo/components/TestSearch.js
tofumatt: Tweaking this test turned out to be a bit more complicated and is a big change in terms of lines. I think it's worth doing (and did it for this test) but a few are more complicated... I'd rather leave it for another patch if that's okay.

tofumatt:
Alright, comments addressed and ready for another review. Thanks for all the comments. I'll go through and see what other issues this fixes, as there are a fair number of search annoyances this clears up.

src/amo/components/Search/index.js
```diff
@@ -26,12 +28,12 @@ export class SearchBase extends React.Component {
     count: PropTypes.number,
     dispatch: PropTypes.func.isRequired,
     enableSearchSort: PropTypes.bool,
+    errorHandler: PropTypes.object.isRequired,
     filters: PropTypes.object,
-    hasSearchParams: PropTypes.bool.isRequired,
+    filtersFromState: PropTypes.object,
```
kumar303: could this be `filtersForResults`? I think that would make it more obvious that it links directly to the `results` property. Saying it comes from `state` is a bit vague since pretty much everything comes from state (even `filters` will come from state once we switch to redux-router).

tests/unit/amo/components/TestCategories.js
```diff
@@ -49,6 +47,36 @@ describe('<Categories />', () => {
...
+      dispatch,
+    });
+
+    sinon.assert.calledWith(dispatch, setViewContext(ADDON_TYPE_EXTENSION));
+    sinon.assert.calledTwice(dispatch);
+
+    root.setProps();
+
+    // Dispatch should not be called again because no new props were set.
+    sinon.assert.calledTwice(dispatch);
```
kumar303: You could also do it this way to capture the intention a bit better:

```js
dispatch.reset();
root.setProps();
sinon.assert.notCalled(dispatch);
```

tests/unit/amo/components/TestFeaturedAddons.js
```diff
@@ -180,14 +180,13 @@ describe('<FeaturedAddons />', () => {
...
       ...mappedProps,
-      hasSearchParams: true,
     });
 
     dispatch.reset();
 
     root.setProps({
       // Update an unrelated parameter.
-      hasSearchParams: false,
+      something: false,
```
kumar303: I like your approach in the other test better (which I didn't think of in this one, heh):

```js
root.setProps();
```

I think updating a non-existent property isn't helpful because it could produce an error one day if we get strict about validating props.

tests/unit/core/test_utils.js
```diff
@@ -1086,3 +1087,33 @@ describe('getCategoryFromState', () => {
...
+  it('treats "0" as 1', () => {
+    expect(parsePage('0')).toBe(1);
+  });
+
+  it('treats 0 as 1', () => {
+    expect(parsePage(0)).toBe(1);
+  });
+
+  it('treats empty strings as 1', () => {
+    expect(parsePage('')).toBe(1);
```
kumar303: maybe one more test for `expect(parsePage(undefined)).toBe(1);` ?

src/amo/components/Category/index.js
```diff
@@ -24,7 +18,10 @@ export class CategoryBase extends React.Component {
     category: PropTypes.object,
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
+    filters: PropTypes.object,
     loading: PropTypes.boolean,
+    paginationQueryParams: PropTypes.object,
```
tofumatt: Oh, thanks, good catch.

src/amo/components/Category/index.js
```diff
@@ -56,48 +65,31 @@ export function mapStateToProps(state, ownProps) {
...
   };
   const pathname = `/${ownProps.params.visibleAddonType}/${filters.category}/`;
-  const queryParams = { page: parsePage(ownProps.location.query.page) };
+  const paginationQueryParams = { page: filters.page };
 
   const category = getCategoryFromState({
     addonType: filters.addonType,
     categorySlug: filters.category,
-    clientApp: filters.clientApp,
+    clientApp: state.api.clientApp,
```
tofumatt: Sure thing. I'll see if a quick follow-up patch for converting would work, I think `shallowUntilTarget()` wasn't available when I started the patch 😄

src/amo/components/Search/index.js
tofumatt: Yeah actually, I think that's a better name anyway. I think you suggested it first and I thought this would be okay but I agree, it's not great. Maybe `filtersUsedForResults` if that's not too verbose.

src/amo/components/Search/index.js
```diff
@@ -60,49 +85,73 @@ export class SearchBase extends React.Component {
...
+
+    // We allow specific paginationQueryParams instead of always using
+    // convertFiltersToQueryParams(filters) so certain search filters
+    // aren't repeated if they are elsewhere in the URL. This is useful
+    // for pages like the category page which contain `addonType` and
+    // `category` in their URLs
+    // (eg: `/extensions/categories/feed-news-blogging/`) so they don't
+    // need them in the queryParams.
+    //
+    // If paginator params aren't specified, we fallback to filters.
```
tofumatt: Oh, yeah, that is a bit cleaner. The only one to worry about is the implicit `page=1` but paginator will always sort that out. As you say we'll leave this for now but I've filed: #2937

tests/unit/amo/components/TestCategories.js
tofumatt: Argh, right. That's what I wanted, sweet.

tests/unit/amo/components/TestFeaturedAddons.js
tofumatt: Oh, good call, I didn't think of that. Will do.

src/amo/components/Category/index.js
tofumatt: Actually, wait, I don't see `location` being used in `CategoryBase`.

src/amo/components/Category/index.js
tofumatt: For now I switched to `dispatchClientMetadata()` instead of `createStore()` and I'll add an extra test to make sure it's pulling out a `category`.

src/amo/components/Category/index.js
kumar303: it's used here, no? https://github.com/mozilla/addons-frontend/blob/395c53e29af341a7508ca4e04b387cabe80a37bc/src/amo/components/Category/index.js#L68

src/amo/components/Search/index.js
kumar303: I like `filtersUsedForResults`

src/amo/components/Category/index.js
tofumatt: Oh, would that count? It's using `ownProps` because the router is passing that to `mapStateToProps`, but the `CategoryBase` component doesn't actually use it at all, just `page`.

Do we include properties used by `mapStateToProps` in the `*Base` component's `PropTypes`? I thought not, because they never actually use `ownProps`/`location`.

kumar303: Yes, it does count. The wrapped component (`Category`) receives the `location` property. To put it another way, when this PropType declaration moves to Flow, the `mapStateToProps` function would trigger an error saying "`ownProps` does not have a `location` property" or something.

tofumatt: Ah shoot, I didn't know that. Also I missed adding it, I'll send a quick follow-up now. Sorry 😓

kumar303: Yeah, no worries, we can fix it up when porting the prop types to Flow. That proptype eslint module is way buggy. I usually just put a comment at the to to disable it!

Issue #2902, opened 2017-08-03
Convert the Discovery Pane to load data with sagas


kumar303:
The Discovery Pane needs to load data with a saga instead of async-connect.

To test for regressions:
* Load the discovery pane via `about:addons`
* Make sure the add-on data still loads
* Make sure themes can be previewed and installed
* Make sure extensions can be installed/uninstalled

SVSanta:
Verified as fixed on AMO-dev FF57(Win 10).
Postfix gif:
![sagas](https://user-images.githubusercontent.com/28138132/29271622-0734a866-8105-11e7-920d-8172c99dfbbf.gif)

Issue #2905, opened 2017-08-04
Show a 500 page for unexpected errors in ErrorHandler


kumar303:
### Describe the problem and steps to reproduce it:

Create a 500 error on a component that uses an ErrorHandler. Locally with `yarn amo`, you can do it like:

- shut down your local API
- go to the extensions landing page

### What happened?

You will see a banner that says 'Unexpected error' and in the console it will explain how the page can't connect to the API. The page will also appear to be trapped in a loading state.

<img width="818" alt="screenshot 2017-09-12 09 45 29" src="https://user-images.githubusercontent.com/55398/30332107-2736f5d2-979f-11e7-97f5-7765c6971cc1.png">

### What did you expect to happen?

It should show a 500 error page and not render any of the real component.

PR #2915, opened 2017-08-08
Convert Discovery Pane to load data with a saga

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2902

This converts `DiscoPane` to use a saga and also fixes a lot of issues in the tests with how they were hard-coding state data or otherwise using unrealistic data. I also converted some tests to Enzyme but only where I needed to use it most.

The behavior should be pretty much the same. We won't even see loading bars since the data is loaded server side. The only new feature is that any kind of catastrophic API error will look like this (until is https://github.com/mozilla/addons-frontend/issues/2905 fixed).

screenshot 2017-08-08 16 53 36


tests/unit/disco/sagas/testDiscovery.js
```diff
@@ -0,0 +1,104 @@
+import SagaTester from 'redux-saga-tester';
```
tofumatt: A comment on this file name: we tend to use `Disco` everywhere so calling this `testDiscovery` rather than `testDisco` stood out to me. Of course, "disco" is a weird code name/shorthand itself, so up to you. Just figured I'd point it out.

tests/unit/disco/sagas/testDiscovery.js
```diff
@@ -0,0 +1,104 @@
...
+import { getDiscoResults, loadDiscoResults } from 'disco/actions';
+import discoResultsReducer from 'disco/reducers/discoResults';
+import discoverySaga from 'disco/sagas/discovery';
+import { dispatchSignInActions } from 'tests/unit/amo/helpers';
+import {
+  createFetchDiscoveryResult, fakeDiscoAddon,
+} from 'tests/unit/disco/helpers';
+
+
+describe('disco/sagas/discovery', () => {
```
tofumatt: I've noticed this is a pattern you've been using naming tests recently. While I think it's okay (it's readable and gives some insight into what the test is describing and what its context is), it's quite fragile as if a test moves (but maybe otherwise stayed the same) the description would need updating, compared to just calling it `Discovery Saga`. Jest does a great job of telling us which files are being run and where any errors are found as well, so I'd say using paths as test descriptors isn't a great habit to get into.

Let's say you had an error in the first test–writing "Discovery Saga" and getting an error here would result in Jest telling us:

```
disco/sagas/discovery.js:52
Discovery Sagas > fetches landing page addons from the API
```

instead of:

```
disco/sagas/discovery.js:52
disco/sagas/discovery > fetches landing page addons from the API
```

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -1,158 +1,172 @@
...
-      const props = helpers.mapStateToProps({
-        addons: { one: { slug: 'one' }, two: { slug: 'two' } },
-        discoResults: [{ addon: 'two' }],
-        infoDialog: {},
-      });
-      expect(props.results).toEqual([{ slug: 'two', addon: 'two' }]);
+      // This is removed by the reducer.
+      delete addon.theme_data;
+
+      // This is a magical theme guid created by the reducer for some reason.
```
mstriemer: Firefox appends the @personas.mozilla.org bit internally. That's how you look it up with mozAddonManager.

src/disco/containers/DiscoPane.js
```diff
@@ -152,21 +162,10 @@ export class DiscoPaneBase extends React.Component {
...
-    return Promise.resolve();
-  }
-  return getDiscoveryAddons({ api: state.api })
-    .then(({ entities, result }) => {
-      dispatch(loadEntities(entities));
-      dispatch(discoResults(result.results.map((r) => entities.discoResults[r])));
-    });
+export function loadedAddons(state) {
+  return state.discoResults.map(
+    (result) => ({ ...result, ...state.addons[result.addon] })
```
kumar303: but...

white-walkers-what-will-happen-when-i-remove-the-spreads


These are old spreads 😢 

They were just moved from: https://github.com/mozilla/addons-frontend/pull/2915/files#diff-0a5f326b61d0fe1d07f721f0d66cacaaL156

src/disco/reducers/discoResults.js
```diff
@@ -1,8 +1,13 @@
-import { DISCO_RESULTS } from 'disco/constants';
+import { LOAD_DISCO_RESULTS } from 'disco/constants';
 
 export default function discoResults(state = [], { type, payload }) {
-  if (type === DISCO_RESULTS) {
-    return payload.results;
+  switch (type) {
+    case LOAD_DISCO_RESULTS: {
+      const { entities, result } = payload;
+      // Unwrap the magical discoResults API schema.
```
kumar303: It is magical because you can't figure out how it transforms the results of the API response just by looking at it.

kumar303: I think the comment is helpful because it's not obvious why the reducer is altering `result.results` in this way. The comment should help someone understand that it's because of the `discoResults` API schema which they can look up in `disco/api.js`.

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -1,158 +1,172 @@
...
     });
 
     it('gets bigger and smaller when clicked', () => {
       const root = render();
-      Simulate.click(root.querySelector('.play-video'));
-      expect(root.querySelector('.show-video')).toBeTruthy();
-      Simulate.click(root.querySelector('.close-video a'));
-      expect(root.querySelector('.show-video')).toBeFalsy();
+      root.find('.play-video').simulate('click', fakeEvent);
+      expect(root.find('.show-video')).toHaveLength(1);
```
kumar303: ah yeah, I forgot about `hasClassName`, thanks

tests/unit/disco/containers/TestDiscoPane.js
kumar303: thanks @mstriemer ! I added a comment about that to the reducer

src/disco/containers/DiscoPane.js
tofumatt: Oh shoot, okay, I thought they were new. Yeah. Let's not mess with them.

src/disco/reducers/discoResults.js
tofumatt: Your comment here makes more sense than the comment 😄

Could you add a note saying the API schema that transforms `result.results` can be found in the `disco/api.js` file? Until #2917 is fixed it's probably best to over-comment that.

tests/unit/disco/sagas/testDiscovery.js
kumar303: oh, I wish it were that simple but it's not. There are many situations (because JavaScript) where Jest will give us no indication *whatsoever* which test file is producing an error. Here is an example:

screenshot 2017-08-11 10 15 36

If that spec wasn't named `disco/sagas/discovery` how would you figure out where the error is coming from? You'd have to start grepping for phrases in the test description which is cumbersome because we have a lot of overlapping test descriptions. 

I see your point about renaming files and forgetting to update specs but until we have a better solution I'd like to put the test file path in all specs. I've definitely struggled to locate failing tests that have vague spec names before and I want to save my future self from that pain.

tofumatt: Hmmm, interesting. Fair enough, that's a good point. You can search the project for the text in the spec, but I see the value of this for sure.

Works for me!

kumar303: oh man! I just realized that since Jest runs in Node (unlike our old karma suite) we can totally use `__filename`. Check it:

```diff
diff --git a/tests/unit/disco/sagas/testDiscovery.js b/tests/unit/disco/sagas/testDiscovery.js
index 3b98cf7a..9ee16c8b 100644
--- a/tests/unit/disco/sagas/testDiscovery.js
+++ b/tests/unit/disco/sagas/testDiscovery.js
@@ -14,7 +14,7 @@ import {
 } from 'tests/unit/disco/helpers';


-describe('disco/sagas/discovery', () => {
+describe(__filename, () => {
   describe('fetchDiscoveryAddons', () => {
     let apiState;
     let errorHandler;
```

screenshot 2017-08-11 10 23 57

tofumatt: Hah! Maybe we need our own `describe` override/function that adds that to the spec name?

Nice though, very clever 😄

kumar303: I will fix this in all the things! https://github.com/mozilla/addons-frontend/issues/2928

tests/unit/disco/sagas/testDiscovery.js
kumar303: sure, I can call it `testDisco` if you want. I like disco. https://www.youtube.com/watch?v=yuLz4T6YXWs

src/disco/containers/DiscoPane.js
```diff
@@ -96,11 +101,12 @@ export class DiscoPaneBase extends React.Component {
     // TODO: Add captions see https://github.com/mozilla/addons/issues/367
     /* eslint-disable jsx-a11y/media-has-caption */
 
-    const { AddonComponent, results, i18n } = this.props;
+    const { AddonComponent, errorHandler, results, i18n } = this.props;
     const { showVideo } = this.state;
 
     return (
       <div id="app-view" ref={(ref) => { this.container = ref; }}>
+        {errorHandler.hasError() ? errorHandler.renderError() : null}
```
tofumatt: I notice we do this a lot, probably worth having `errorHandler.renderError()` just output `null` if no errors exist as it's often what we do... separate thing though.

tests/unit/core/server/testServer.js
```diff
@@ -31,13 +32,22 @@ describe('core/server/base', () => {
 
   function testClient({ stubRoutes = defaultStubRoutes } = {}) {
     function createStoreAndSagas() {
+      const sagaMiddleware = createSagaMiddleware();
       return {
-        store: createStore(combineReducers({ reduxAsyncConnect })),
-        sagaMiddleware: null,
+        store: createStore(
+          combineReducers({ reduxAsyncConnect }),
+          undefined,
```
tofumatt: Whoa, I must have missed this before. Why the `undefined` here (at least over `null` or something else empty?)

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -1,158 +1,173 @@
...
     });
 
     it('gets bigger and smaller when clicked', () => {
       const root = render();
-      Simulate.click(root.querySelector('.play-video'));
-      expect(root.querySelector('.show-video')).toBeTruthy();
-      Simulate.click(root.querySelector('.close-video a'));
-      expect(root.querySelector('.show-video')).toBeFalsy();
+      root.find('.play-video').simulate('click', fakeEvent);
+      expect(root.find('header')).toHaveClassName('.show-video');
```
tofumatt: Thanks these read better 👍

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -163,14 +178,60 @@ describe('AddonPage', () => {
...
+      const dispatch = sinon.stub();
+      const configSource = { server: true };
+      const configStub = { get: (key) => configSource[key] };
+      expect(helpers.mapDispatchToProps(dispatch, { _config: configStub }))
+        .toEqual({ dispatch });
+    });
+  });
+
+  describe('constructor', () => {
+    it('gets discovery results when results are 0 length', () => {
```
tofumatt: This phrasing is a bit awkward. Maybe `gets discovery results if results are empty`?

tests/unit/disco/sagas/testDisco.js
```diff
@@ -0,0 +1,104 @@
...
+import { getDiscoResults, loadDiscoResults } from 'disco/actions';
+import discoResultsReducer from 'disco/reducers/discoResults';
+import discoSaga from 'disco/sagas/disco';
+import { dispatchSignInActions } from 'tests/unit/amo/helpers';
+import {
+  createFetchDiscoveryResult, fakeDiscoAddon,
+} from 'tests/unit/disco/helpers';
+
+
+describe(__filename, () => {
```
tofumatt: Is this still worth doing after #2928 was found to be okay?

I actually think this is a cute idea but I'd rather a helper did it universally–`describe(__filename, () => {` is sort of awkward to read.

If you want to leave it for now I guess that's okay though, no biggie... we aren't THAT strict about spec names and I think that's fine.

tests/unit/core/server/testServer.js
kumar303: Not my fault :) `createStore` is just an [awkward API](http://redux.js.org/docs/api/createStore.html). Passing `undefined` seems like the least intrusive way to skip the argument (I do not want to define the second argument).

tofumatt: Yeah, I know it's not your fault 😄 

Makes sense, but just because it's a weird API call it could use a comment–even what you wrote here works for me.

src/disco/containers/DiscoPane.js
kumar303: I like that. It could even be a good one for William. https://github.com/mozilla/addons-frontend/issues/2929

kumar303:
Thanks. I was hoping to land this sooner for more testing but we need to get it out. I'll check it on dev.

PR #2918, opened 2017-08-09
Add dimension3 to addons-frontend

### LINKED ISSUE #2827: Be able to tell apart the new frontend from the legacy frontend in analytics
To be able to check things like page perf metrics against the old site we need a good way to tell them apart. 

I'm thinking that a custom dimension would be the best way to do that, but there might be an alternative.

muffinresearch:
Fixes #2827

### DIFF
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -71,11 +71,14 @@ export class Tracking {
       /* eslint-disable */
       // Snippet from Google UA docs: http://bit.ly/1O6Dsdh
       window.ga = window.ga || function() {(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+      /* eslint-enable */
       ga('create', this.id, 'auto');
       if (convertBoolean(_config.get('trackingSendInitPageView'))) {
         ga('send', 'pageview');
       }
-      /* eslint-enable */
+      // Set a custom dimension; this allows us to tell which front-end
+      // (addons-frontend vs addons-server) is being used in analytics.
+      ga('set', 'dimension3', 'addons-frontend');
     }
   }
 
--- a/tests/unit/core/test_tracking.js
+++ b/tests/unit/core/test_tracking.js
@@ -75,11 +75,17 @@ describe('Tracking', () => {
     sinon.assert.calledWith(window.ga, 'send', 'pageview');
   });
 
+  it('should set dimension3', () => {
+    createTracking();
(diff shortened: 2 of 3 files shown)


src/core/tracking.js
```diff
@@ -75,6 +75,8 @@ export class Tracking {
       if (convertBoolean(_config.get('trackingSendInitPageView'))) {
         ga('send', 'pageview');
       }
+      // Set dimension to be able to tell apart the new frontend from the old.
```
tofumatt: This is a nitpick but I'd change the wording here to be clearer:

```js
// Set a custom dimension; this allows us to tell which front-end
// (addons-frontend vs addons-server) is being used in analytics.
```

tests/unit/core/test_tracking.js
```diff
@@ -75,11 +75,17 @@ describe('Tracking', () => {
...
 
+  it('should set dimension3', () => {
+    createTracking();
+    sinon.assert.calledWith(window.ga, 'set', 'dimension3', 'addons-frontend');
+  });
+
   it('should not send initial page view when disabled', () => {
     createTracking({ trackingSendInitPageView: false });
-    // Make sure only 'create' was called, not 'send'
+    // Make sure only 'create' and  'set' was called, not 'send'
```
tofumatt: "were called", not "was called"

tests/unit/core/test_tracking.js
```diff
@@ -75,11 +75,17 @@ describe('Tracking', () => {
...
     sinon.assert.calledWith(window.ga, 'create');
-    sinon.assert.callCount(window.ga, 1);
+    sinon.assert.calledWith(window.ga, 'set', 'dimension3', 'addons-frontend');
```
tofumatt: Seeing this string repeated three times across files makes me think "use a constant!" but up to you.

muffinresearch: @tofumatt for which bit?

tofumatt: Oops, sorry, `'addons-frontend'`

muffinresearch: Assuming you're talking about 'addons-frontend' or 'dimension3' I think it's probably not necessary since it's unlikely to get used further than this.

tofumatt:
Yeah that’s fine, leave as-is then :-)

Issue #2691, opened 2017-07-05
Try out bundlesize to track the size of our assets


kumar303:
Try out [bundlesize](https://github.com/siddharthkp/bundlesize) to track the size of our assets in CI.

kumar303:
Hmm, maybe it won't work. https://github.com/siddharthkp/bundlesize/issues/39

kumar303:
Huh. It fails on master only possibly due to not using the Github integration. See https://github.com/siddharthkp/bundlesize/issues/93

kumar303:
See https://github.com/mozilla/addons-frontend/pull/4277 for the configuration code that we started on.

PR #2920, opened 2017-08-09
Check bundle size in CI against a threshold

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2691


kumar303:
@muffinresearch I was mistaken. This patch doesn't require GitHub access because all it does is fail if we dip below our own hard-coded threshold. 

It doesn't give us graphs over time or anything fancy. Obviously when we hit the max we'll have to bump the threshold but at least that will force us to consider the impact of such a change.

kumar303:
The output is kind of terrible since our build log is too big for Travis. Here's where it shows the bundle sizes in the raw log:

screenshot 2017-08-09 12 17 54

And now my memory has been jogged. They do offer [TravisCI integration](https://github.com/siddharthkp/bundlesize#2-build-status) which is where I got rejected when requesting access for it. I'm not sure what "cool features" it provides so it would probably be worth landing the patch without it.

muffinresearch:
I wonder if we should make the test output quiet for travis (with --silent IIRC) on the basis that individual failures should still be visible - I found the amount of noise got in the way recently too when looking for what had failed.

kumar303:
> Might be nice to ... track locale bundles

I was looking into this at first but for some reason locale files aren't getting built with this command. I couldn't figure out why and I don't want to spend too much time on it. Maybe you have some ideas? I filed it here: https://github.com/mozilla/addons-frontend/issues/2923

kumar303:
> I wonder if we should make the test output quiet for travis?

Hmm, I don't see a way to do this easily. Maybe we can fix it up later.

Thanks for the review!

Issue #2643, opened 2017-06-27
Unwrap HOC components to make shallow render tests more effective


kumar303:
Since we use shallow Enzyme rendering now we no longer have to test the base (i.e. unwrapped) components. We should be able to introduce a helper that unwraps them automatically.

Example: https://github.com/mozilla/addons-frontend/pull/2618#discussion_r123585638

kumar303:
Enzyme doesn't have great support for doing this kind of unwrapping: https://github.com/airbnb/enzyme/issues/539

PR #2926, opened 2017-08-10
Make it possible to test default component instances (with helper function)

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2643

This introduces a helper to perform a shallow render on a component that may be wrapped in other HOCs (higher order components). This allows us to test default components in a manner similar to how the production app operates.

I only converted a few tests to prove if the concept will work. I think it will work. 

The nicest result is that we can simply dispatch actions to load data rather than mimic what `redux/connect/mapStateToProps` does under the hood. Ultimately, it should provide much better test coverage for each component.


tests/unit/amo/components/TestSectionLinks.js
```diff
@@ -24,14 +34,9 @@ describe('SectionLinks Component', () => {
     expect(root.find('.SectionLinks-link').length).toEqual(3);
   });
 
-  it('renders no active links without context', () => {
-    const root = render({ viewContext: null });
```
kumar303: Interestingly, this scenario is no longer possible to test because it's not possible in the real app!

tests/unit/core/TestInstallAddon.js
```diff
@@ -231,7 +231,7 @@ describe('withInstallHelpers inner functions', () => {
         guid,
         installURL,
       });
-      const setCurrentStatus = root.prop('setCurrentStatus');
+      const { setCurrentStatus } = root.instance().props;
```
tofumatt: Hmmm, it feels like it would be better/easier to use if `root.instance()` was what was returned by `shallowUntilTarget`. That way it feels exactly like regular enzyme `shallow`. Migrating from one to another is easier then and it's less boilerplate. I'm seeing `root.instance()` a lot; that means we should code it out 😄

tests/unit/core/components/TestInstallButton.js
```diff
@@ -34,11 +34,13 @@ describe('<InstallButton />', () => {
...
     hasAddonManager: true,
     i18n: getFakeI18nInst(),
+    store: createStore().store,
     ...customProps,
   });
 
-  const render = (props) =>
-    shallow(<InstallButtonBase {...renderProps(props)} />);
+  const render = (props) => shallowUntilTarget(
+    <InstallButton {...renderProps(props)} />, InstallButtonBase
```
tofumatt: Would it be possible to know how many HOCs we're `compose`-ing and just get that element or do we need to have the base component as well?

tests/unit/testHelpers.js
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable react/no-multi-comp */
+import { shallow } from 'enzyme';
+import React, { Component } from 'react';
+import { compose } from 'redux';
+
+import { shallowUntilTarget } from 'tests/unit/helpers';
+
+describe('helpers', () => {
+  describe('shallowUntilTarget', () => {
```
tofumatt: Meta testing! 👍

tests/unit/core/TestInstallAddon.js
kumar303: `shallow()` does not return `root.instance()`, it returns a `ShallowWrapper`. I'd rather make `shallowUntilTarget()` work exactly like `shallow()`. If I make it return `root.instance()` then `root.find(SubComponent)` won't work and you'd be stuck. 

The reason this patch had to switch to `root.instance()` is because I was trying to unwrap the component hierarchy manually in the previous code and I was off by one level :) It should not have been possible to call `root.prop()` on the output before -- `shallow()` would not have let you do that.

tests/unit/core/components/TestInstallButton.js
kumar303: There are a couple other options:

* We could hard code the number of HOCs we wrap in the tests and call the function like `shallowUntilTarget(<Component /> { hocToUnwrap: 4 })`. But then we'd have to manually update that number any time we add/remove an HOC which I don't like.
* We could export a list of HOCs in the component and then use that in the tests like: `shallowUntilTarget(<MyComponent /> { hocToUnwrap:  MyComponentHOCs.length })`. This really isn't too different.

Considering the options, I still think looking for a base component is the best one.

kumar303: Another option that I never took seriously:

* We could put a special property on each HOC so that the unwrapper can pop off each one automatically based on presence of the property. This feels dirty and will be error prone since you'd have to remember to decorate each HOC with the special property.

kumar303:
@tofumatt thanks for the review. I responded to your change requests -- let me know if you have any other ideas to try.

tests/unit/core/components/TestInstallButton.js
tofumatt: The only other option that's sort of like option 2 would be to setup a convention of declaring each HOC in an array we export as a property of each component which would clean up the test API, eg:

```js
// src/amo/components/Search.js
export class SearchBase extends React.Component {
  [...]
}

export const HigherOrderComponents = [
  connect(mapStateToProps),
  translate(),
];

const Search = compose(...HigherOrderComponents)(SearchBase);

Search._higherOrderComponents = HigherOrderCompoents;

export default Search;
```

But writing that out it's way more boilerplate and annoyance than just specifying `shallowUntilTarget(<Search>, SearchBase)`. So let's stick with your existing API!

tests/unit/core/TestInstallAddon.js
tofumatt: Ahhhhhhhhh. Oops. Okay, so `root.props()` for most components will continue to work as it does now, this was just a bit off in the past.

Cool, thanks!

tofumatt:
Thanks for the clarifications, I think this is good to merge now that I understand things and we've thought about alternate APIs for unwrapping things. I'd love it if it was more automatic but I think this is still a pretty good API and it's better than what we've been doing so 👍

kumar303:
> I'd love it if it was more automatic

I don't know that it ever can be simply because HOCs are just like any other components - they aren't special. Maybe the airbnb folks will think of something better in https://github.com/airbnb/enzyme/issues/539

Issue #2087, opened 2017-03-17
mobile AMO allows installing incompatible extensions


ValentinaPC:
Steps to reproduce:
1. On Desktop, select an Android add-on that is not compatible with your platform i.e. https://addons.mozilla.org/en-US/android/addon/cloud-viewer
2. Click "View Mobile Site"
3. Install the add-on
4. Go in about:addons Extensions tab and see your add-on

Expected results:
Add-on is installed and disabled. Reason of disabling the add-on is that it is incompatible with your platform.

Actual results:
Reason of disabling the add-on is that it is incompatible with your version of Firefox.

Notes/Issues:
 - Installing the add-on on device, using the same version of Firefox, as before, will work without any messages.

Verified on FF52(Win 7, Android 6.0.1) on AMO-prod.
Video for this issue:
![wrongmsg](https://cloud.githubusercontent.com/assets/15685960/24036706/728ff016-0b03-11e7-9410-b7987b6f4e1e.gif)

muffinresearch:
@ValentinaPC note that the discovery pane is *not* the list of addons in about:addons. It is merely the page with the curated list of addons which looks like this:

![add-ons_manager](https://cloud.githubusercontent.com/assets/1514/24037191/d9ec2f86-0af4-11e7-8827-782dc83c61d0.png)

ValentinaPC:
> Actually fixing the incorrect message in about:addons#Extensions would require a bugzilla bug to be raised.

Logged a Bugzilla issue under: https://bugzilla.mozilla.org/show_bug.cgi?id=1349135

designakt:
To prevent this problem from arising, we should implement @muffinresearch suggestion.
> I think the bug we can action here is that the mobile page should have a message about platform compatibility and it should follow desktop in not allowing installation for an incorrect platform.

designakt:
@pwalm can you please take this - you new desktop might already have a UI for this.

pwalm:
- No install button
- Replace install button with "Not available for your platform" copy from desktop site (will have to break it into two lines)
- Align copy to the right

muffinresearch:
@tofumatt would you have a chance to look at this? It would be good to add this messaging to avoid confusion especially if the platforms allow installation of incorrect add-ons.

@jvillalobos mentioned developers are also getting confused by this when viewing /firefox/ from android.

tofumatt:
> developers are also getting confused by this when viewing /firefox/ from android.

[...]

Issue #2911, opened 2017-08-07
Follow compatibility rules for the add-ons install buttons


diox:
Now the the API exposes `is_strict_compatibility_enabled` boolean, together with the `compatibility` object on versions, addons-frontend should be able to disable the install buttons when an add-on is not supposed to be compatible with the client's version.

The basic rules are:
- Client's version needs to be higher or equal than `compatibility` `min` for the appropriate app regardless of `is_strict_compatibility_enabled` 
- Client's version needs to be lower or equal than `compatibility` `max` for the appropriate app if `is_strict_compatibility_enabled` is enabled, ignored otherwise

When we're done implementing all compatibility-related bugs in addons-server (see [the full plan](https://github.com/mozilla/addons-server/issues/5277#issuecomment-318074010)), the behavior for incompatible add-ons should be to hide the install button entirely when an add-on is not compatible, and instead show a message, `This add-on is not compatible with your version of {0}` together with a link to `View other versions`. you can see the last part in action in https://github.com/mozilla/addons-server/issues/6028#issuecomment-319713741 (hiding the install button itself has not been implemented yet at the time I'm filing this)

muffinresearch:
@diox I think we've done most of this - can you see anything missing?

diox:
#2888 implemented the compatibility rules with strict compatibility etc a while back.

What the frontend is still missing though is the link to download the add-on anyway or see older versions. This is better tracked in #3021 though.

PR #2934, opened 2017-08-14
feat: Disable legacy extensions for Firefox 57+ (fix #2888)

### LINKED ISSUE #2888: Install buttons should be disabled for legacy extensions if 57+
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1374607

If using a 57+ browser the install button should be disabled in amo and the disco pane preventing an installation of a legacy extension.

Currently as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1374607 if the pref `extensions.legacy.enabled` is false it still looks like you can install a legacy extension except it is disabled immediately.

The disco pane has already moved to only showing web-ext so this is lower priority there but likely if we fix this for the general install button case it will also make sure this happens for disco too in the unlikely event non web-ext content was shown.

tofumatt:
Fixes #2888.

When an addon has a type of `extension` but is not flagged as a webextension, we disable installation and show an error.

### Screenshot on FF 57
screenshot 2017-08-14 15 41 33


src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -53,8 +54,12 @@ export class AddonCompatibilityErrorBase extends React.Component {
 
     if (reason === INCOMPATIBLE_NOT_FIREFOX) {
       message = i18n.sprintf(i18n.gettext(`You need to
-        <a href="%(downloadUrl)s">download Firefox</a> to install this add-on.`
+        <a href="%(downloadUrl)s">download Firefox</a> to install this
+        add-on.`
```
kumar303: was this line too long? Github makes it hard to tell. If it's not too long, it would probably better to put this back to where it was. Otherwise, our localizers will have a lot of busy work fixing fuzzy matches.

src/core/utils.js
```diff
@@ -360,6 +362,27 @@ export function isCompatibleWithUserAgent({
         https://github.com/mozilla/addons-frontend/issues/2074`);
     }
 
+    // Check to see if:
```
kumar303: Why do we need to make a special case for this? We already have code that checks version compatibility. Can't you just re-use that? Here is an idea:

```diff
diff --git a/src/core/utils.js b/src/core/utils.js
index c91213fa..12cf9e29 100644
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -310,6 +310,18 @@ export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
     if (addon.current_version.compatibility[clientApp]) {
       maxVersion = addon.current_version.compatibility[clientApp].max;
       minVersion = addon.current_version.compatibility[clientApp].min;
+    } else if (
+      addon.type === ADDON_TYPE_EXTENSION &&
+      clientApp === 'firefox' &&
+      (
+        !addon.current_version || !addon.current_version.files ||
+        !addon.current_version.files.length ||
+        addon.current_version.files[0].is_webextension !== true
+      )
+    ) {
+      // This is a legacy extension that is only supported by
+      // Firefox 57+
+      minVersion = '57.0';
     } else if (addon.type === ADDON_TYPE_OPENSEARCH) {
       _log.info(oneLine`addon is type ${ADDON_TYPE_OPENSEARCH}; no
         compatibility info found but this is expected.`, { addon, clientApp });

```

After that, I suspect that you'll get all the comparison code for free. You'll just need to display the error message which might simultaneously fix https://github.com/mozilla/addons-frontend/issues/2087 🍰

src/amo/components/AddonCompatibilityError/index.js
tofumatt: It was. Actually, the tags should really be removed from the translation anyway and made into `%(startLink)s` and `%(endLink)s`, so maybe I'll do those both at the same time. That makes it worth the work of the fuzzy match I think.

src/core/utils.js
tofumatt: Dammit. You're totally right! Nice catch, I'll do that.

tofumatt: Wait, you're marking the min version here as 57 but not the max version, which is what this issue is for. This issue fixes extensions that aren't 57-compatible being marked as such because we *don't* and *shouldn't* actually honour the `maxVersion` attribute.

tofumatt:
From your patch suggestion maybe it wasn't clear what this patch should do, but it's to not allow non-WebExtensions from appearing as installable on 57+.

We can't shouldn't really rely on `maxVersion` because it's set from the API but also sort of weird: https://github.com/mozilla/addons-frontend/blob/master/src/core/utils.js#L352

We could special-case that code even further, which I'd rather not, or just create a new "anything not a webext isn't allowed in 57" rule.

muffinresearch:
@tofumatt this issue is related and might provide a better guide for the rules: See https://github.com/mozilla/addons-frontend/issues/2911

tofumatt:
Huh, oh, shoot, I didn't realise that. Okay then, I'll take another crack at this with that info in mind. Cheers!

tofumatt:
Alright, redid things to be a bit more straightforward, thanks!

Most of the lines changed here are because I moved to enzyme for this component as well 😄

tofumatt:
Shoot, sorry, forgot to request review again for this... 😢

tests/unit/amo/helpers.js
```diff
@@ -26,7 +26,10 @@ export const fakeAddon = Object.freeze({
     id: 123,
     license: { name: 'tofulicense', url: 'http://license.com/' },
     version: '2.0.0',
-    files: [],
+    files: [{
+      is_webextension: true,
+    }],
+    is_strict_compatibility_enabled: true,
```
kumar303: This should not default to to `true` in the shared `fakeAddon` object; it's an unlikely value

src/core/utils.js
```diff
@@ -354,9 +355,14 @@ export function isCompatibleWithUserAgent({
     // confused by this as tofumatt was.
     // See: https://github.com/mozilla/addons-frontend/issues/2074#issuecomment-286983423
     if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
-      _log.info(oneLine`maxVersion ${maxVersion} for add-on lower than browser
-        version ${browser.version}, but add-on still marked as compatible
-        because we largely ignore maxVersion. See:
+      // See: https://github.com/mozilla/addons-server/issues/5277
+      if (addon.current_version.is_strict_compatibility_enabled) {
+        return { compatible: false, reason: INCOMPATIBLE_LEGACY_EXTENSION };
```
kumar303: This looks like the right thing to do but this message should be more specific to strict compatibility. We already have `INCOMPATIBLE_UNDER_MIN_VERSION` so I would suggest `INCOMPATIBLE_OVER_MAX_VERSION` which is to say the user's Firefox version is higher than the max version allowed by the extension.

Also, keep in mind that any add-on can opt into strict compatibility checks, not just "legacy" add-ons. A web extension add-on can say that it's using an experimental API which will be removed in Firefox 59, for example. Furthermore, beta add-on releases typically say they are only compatible with the current version of Firefox to dissuade ongoing usage.

src/core/utils.js
```diff
@@ -354,9 +355,14 @@ export function isCompatibleWithUserAgent({
     // confused by this as tofumatt was.
     // See: https://github.com/mozilla/addons-frontend/issues/2074#issuecomment-286983423
     if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
-      _log.info(oneLine`maxVersion ${maxVersion} for add-on lower than browser
-        version ${browser.version}, but add-on still marked as compatible
-        because we largely ignore maxVersion. See:
+      // See: https://github.com/mozilla/addons-server/issues/5277
+      if (addon.current_version.is_strict_compatibility_enabled) {
```
kumar303: Related: it's worth adding a comment over here that `INCOMPATIBLE_UNDER_MIN_VERSION` is enforced [regardless of strict compatibility](https://github.com/mozilla/addons-frontend/issues/2911): https://github.com/mozilla/addons-frontend/blob/e32a09b41e699055ccd7b35f4da711c6f207c42b/src/core/utils.js#L379

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -53,8 +54,12 @@ export class AddonCompatibilityErrorBase extends React.Component {
...
       ), { downloadUrl });
+    } else if (reason === INCOMPATIBLE_LEGACY_EXTENSION) {
+      message = i18n.gettext(`This is a legacy extension that isn't
+        compatible with your version of Firefox.`);
```
kumar303: I don't think this should use the word *legacy* because version incompatibility applies to future add-ons too. Jorge had this suggestion in https://github.com/mozilla/addons-server/issues/5277#issuecomment-315185528 : "This add-on is not compatible with your version of Firefox."

src/core/utils.js
```diff
@@ -354,9 +355,14 @@ export function isCompatibleWithUserAgent({
     // confused by this as tofumatt was.
     // See: https://github.com/mozilla/addons-frontend/issues/2074#issuecomment-286983423
     if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
-      _log.info(oneLine`maxVersion ${maxVersion} for add-on lower than browser
-        version ${browser.version}, but add-on still marked as compatible
-        because we largely ignore maxVersion. See:
+      // See: https://github.com/mozilla/addons-server/issues/5277
```
kumar303: I don't think this bug is relevant because `current_version.is_strict_compatibility_enabled` is not a new concept (it's just newly exposed in the API). The bug you linked to was just about automatically setting a bunch of add-ons to use strict compatibility. Developers can still manually turn on strict compatibility.

src/core/utils.js
tofumatt: Ah okay, I wasn't aware. The compatibility rules are all a bit dizzying. I'll modify the message, cheers 👍

tofumatt:
Sorry for the delay on this. Made all requested changes, ready for another review.

tests/unit/amo/components/TestAddonCompatibilityError.js
```diff
@@ -33,72 +29,106 @@ describe('AddonCompatibilityError', () => {
...
-      lang: 'es',
+      lang: 'en-GB',
       reason: INCOMPATIBLE_NOT_FIREFOX,
-      userAgentInfo: { browser: { name: 'Chrome' }, os: {} } });
-    const rootNode = findDOMNode(root);
+      userAgentInfo: { browser: { name: 'Chrome' }, os: {} },
+    });
+
+    expect(
+      root.find('.AddonCompatibilityError').render().find('a').attr('href')
```
kumar303: That's interesting. I've been doing this more like `expect(root.find('.localized').html()).toContain('<a ...')` but I like your approach better.

tofumatt: I think I started that way too, but yeah I like not having to actually type out the tag. I think I only figured out I could do that in this patch 😄

PR #2939, opened 2017-08-15
Add renderErrorIfPresent() method in ErrorHandler

### LINKED ISSUE #2929: Simplify code to render errors from ErrorHandler
We use this pattern a lot in our components:

```js
<SomeRoot>
</SomeRoot>
```

Let's introduce a new method that will execute the same condition under the hood:

```js
<SomeRoot>
</SomeRoot>
```

willdurand:
Fixes #2929

---

This PR introduces a `renderErrorIfPresent()` method to the `ErrorHandler` as requested in #2929.

### DIFF
--- a/tests/unit/core/testErrorHandler.js
+++ b/tests/unit/core/testErrorHandler.js
@@ -4,6 +4,7 @@ import { findDOMNode } from 'react-dom';
 import { findRenderedComponentWithType, renderIntoDocument }
   from 'react-addons-test-utils';
 import { createStore, combineReducers } from 'redux';
+import { shallow } from 'enzyme';
 
 import I18nProvider from 'core/i18n/Provider';
 import { createApiError } from 'core/api/index';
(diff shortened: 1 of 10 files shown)


tests/unit/core/testErrorHandler.js
```diff
@@ -263,5 +263,15 @@ describe('errorHandler', () => {
...
     });
+
+    it('returns no component if it does not have an error', () => {
+      const handler = new ErrorHandler();
+      expect(handler.renderErrorIfPresent()).toBe(null);
+    });
+
+    it('returns a component if it has an error', () => {
+      const handler = new ErrorHandler({ capturedError: new Error() });
+      expect(handler.renderErrorIfPresent()).not.toBe(null);
```
tofumatt: It's worth checking that we rendered the intended component here more specifically. I think:

```js
expect(handler.renderErrorIfPresent().find(ErrorList))).toHaveLength(1);
```

should do the trick.

willdurand: That is way better indeed.

I need to use some sort of Enzyme wrapper though or did I miss a testing lib that allows to write this that way?

tofumatt: Ah we [naughtily inject the enzyme testing wrapper into all our test files](https://github.com/mozilla/addons-frontend/blob/master/tests/setup.js#L6), so you should be able to just use that out of the box 😄

willdurand: Mmmh, but the `find()` method call must be done on a Enzyme wrapped component if I understand correctly, and `renderErrorIfPresent()` returns a plain old React component.

tofumatt: Oh, shoot, right. If you wrapped it in `shallow()` that might work?

willdurand: Yes, surrounded by `<div>` :)

PR #2944, opened 2017-08-15
Rename withErrorHandling() to withRenderedErrorHandler()

### LINKED ISSUE #2931: Rename withErrorHandling to withErrorHandlerRenderer
I was a bit confused to learn we have two HOCs for our error rendering: `withErrorHandling` and `withErrorHandler` (https://github.com/mozilla/addons-frontend/pull/2909#discussion_r132560152). They're very close and I think could be easily mistaken for each other as-is–plus they don't clearly differentiate themselves based on name, so we should rename the one that automatically outputs to something more meaningful.

willdurand:
Fixes #2931 

---

Rename a HOC to avoid confusion, cf. #2931.

### DIFF
--- a/src/core/errorHandler.js
+++ b/src/core/errorHandler.js
@@ -59,13 +59,8 @@ export class ErrorHandler {
 export type ErrorHandlerType = typeof ErrorHandler;
 
 /*
- * This is a decorator that gives a component the ability to handle errors.
- *
- * The decorator will assign an ErrorHandler instance to the errorHandler
- * property.
- *
- * For convenience, you can use withErrorHandling() instead which will
- * additionally renders the error automatically.
+ * For convenience, you can use `withRenderedErrorHandler()` which renders the
+ * error automatically at the beginning of the component's output.
  *
  * Example:
  *
@@ -77,7 +72,7 @@ export type ErrorHandlerType = typeof ErrorHandler;
  *     const { errorHandler } = this.props;
  *     return (
  *       <div>
- *         {errorHandler.hasError() ? errorHandler.renderError() : null}
+ *         {errorHandler.hasErrorIfPresent()}
  *         <div>some content</div>
  *       </div>
  *     );
@@ -141,10 +136,10 @@ export function withErrorHandler({ name, id }) {
  * }
  *
  * export default compose(
- *   withErrorHandling({ name: 'SomeComponent' }),
+ *   withRenderedErrorHandler({ name: 'SomeComponent' }),
  * )(SomeComponent);
  */
-export function withErrorHandling({ name, id } = {}) {
+export function withRenderedErrorHandler({ name, id } = {}) {
   return (WrappedComponent) => {
     function ErrorBanner(props) {
       // eslint-disable-next-line react/prop-types
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import { oneLine } from 'common-tags';
 
-import { withErrorHandling } from 'core/errorHandler';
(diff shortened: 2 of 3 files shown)


src/core/errorHandler.js
```diff
@@ -64,7 +64,7 @@ export type ErrorHandlerType = typeof ErrorHandler;
  * The decorator will assign an ErrorHandler instance to the errorHandler
  * property.
  *
- * For convenience, you can use withErrorHandling() instead which will
+ * For convenience, you can use withRenderedErrorHandler() instead which will
```
tofumatt: This is unrelated but that comment is probably a bit too verbose. I think it would be fine if it just said:

```js
* For convenience, you can use `withRenderedErrorHandler()` which renders
* the error automatically at the beginning of the component's output.
```

thoughts?

willdurand: Would you also like to get rid of the example? Or is it just that sentence that I should rephrase?

willdurand: I went ahead and kept the example below your suggestion.

src/core/errorHandler.js
```diff
@@ -59,13 +59,8 @@ export class ErrorHandler {
 export type ErrorHandlerType = typeof ErrorHandler;
 
 /*
- * This is a decorator that gives a component the ability to handle errors.
- *
- * The decorator will assign an ErrorHandler instance to the errorHandler
- * property.
```
kumar303: Thanks for this patch @willdurand !

Your addition about `withRenderedErrorHandler()` looks right but some of the previous documentation got chopped off. Can you open a new PR to put it back? The doc block should look like this to make it clear that the component will get an `errorHandler` prop:

```js
/*
 * This is a decorator that gives a component the ability to handle errors.
 *
 * The decorator will assign an ErrorHandler instance to the errorHandler
 * property.
 *
 * For convenience, you can use `withRenderedErrorHandler()` which renders the
 * error automatically at the beginning of the component's output.
 *
 * Example ...
 */
```

kumar303:
@willdurand Also, be sure to use the [squash and merge](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) button so that we always get one commit in the master log. This helps keep it readable and will also be convenient for you since you won't have to worry about rebasing and squashing as you do fix-ups in your pull request branch.

willdurand:
> Also, be sure to use the squash and merge button so that we always get one commit in the master log.

@kumar303 so we always have 1 PR = 1 commit?

src/core/errorHandler.js
willdurand: Doh! @tofumatt [suggested to shorten the comment](https://github.com/mozilla/addons-frontend/pull/2944#discussion_r133213126), so I thought I could get rid of these lines.

tofumatt:
Yeah, we aim to keep one commit per PR whenever possible. Generally we want one commit per issue. If you had a pr that perhaps solved two issues with two atomic commits that could be okay, but otherwise we find the single commit per issue makes things easy to read in the logs, cherry pick for hot-fixes, and revert.

src/core/errorHandler.js
willdurand: see: https://github.com/mozilla/addons-frontend/pull/2954

willdurand:
Good to know, thanks @kumar303 and @tofumatt

kumar303:
Yes, and after I said that I noticed @tofumatt uses merge commits (which introduces two commits). If that's your preference (to keep your branches in sync) then I guess that's okay but I would rather see a PR squashed and rebased on master into a single commit (no merge commit) because that makes cherry-picks and reverts a lot easier. This is what the github Squash and Merge button does. You have to force-delete your local branches with that approach though.

willdurand:
Oh you mean fast forward in addition? I am definitely not a fan but it makes sense if commits are squashed anyway (the 1 PR = ! commit rule).

PR #2946, opened 2017-08-15
Fix test suite (i18n) // Make sure that nodes uses full icu

### LINKED ISSUE #2945: Test suite is failing depending on the locale

Run: `yarn run test-once`

### What happened?

On my laptop, I have two failing test cases related to the `i18n` module:

```
Summary of all failing tests
 FAIL  tests/unit/core/i18n/test_utils.js
  ● i18n utils › makeI18n › falls-back to number.toLocaleString if Intl is not an object

    expect(received).toEqual(expected)

    Expected value to equal:
      "1,234"
    Received:
      "1.234"

      at Object.<anonymous> (tests/unit/core/i18n/test_utils.js:464:41)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)

  ● i18n utils › makeI18n › falls-back to number.toLocaleString when Intl is missing a NumberFormat constructor

    expect(received).toEqual(expected)

    Expected value to equal:
      "12,345"
    Received:
      "12 345"

      at Object.<anonymous> (tests/unit/core/i18n/test_utils.js:474:41)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)


Test Suites: 1 failed, 136 passed, 137 total
Tests:       2 failed, 1584 passed, 1586 total
```

All green 💚

willdurand:
Fixes #2945 

---

This should fix the test suite, hopefully.

The first test case is fixed by expecting the same output as the previous test case (which has the same input now, only a different implementation under the hood, but I guess we expect the same output).

The second test case was not accurate. In French, commas are used to create decimal numbers, so 12000 (twelve thousands) must be written either `12000` or `12 000`, but `12,000` means twelve to me.


willdurand:
Well, of course it is failing. My understanding of [this page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) is that locales can be passed to this `toLocaleString()` method, but [Twitter](https://twitter.com/wesbos/status/750423643323297792?lang=en) says something different, and Node 6 does not help 😢

willdurand:
Updated the code to "cheat"... If I understand correctly, we want to check whether `Number.toLocaleString()` has been called, so maybe the result does not matter as far as it is the same?

muffinresearch:
Checking the display I think we have a bug here in that the server and the browser disagree on the output of `foo = 1000; foo.toLocaleString('fr')`;

Under node 6.10.3 I get:
```
> foo = 1000; foo.toLocaleString('fr')
'1,000'
```

Under Firefox I get:
```
foo = 1000; foo.toLocaleString('fr')
"1 000"
```

Jest will use node so it will be providing whatever node outputs for that. 

The output might also vary depending on what locales the node has available which might be why it's failing for you. 

We also need to figure out what needs to change so that the server render isn't different to what the js outputs since currently I see a flash of a comma which is then replaced when the js loads looking at https://addons-dev.allizom.org/en-US/firefox/extensions/feeds-news-blogging/?page=3 for example.

willdurand:
Yeah node 6 seems not good, but I also tried node 8 and it did not help much.

Firefox is correct: `1000` must be `1 000` (or `1000`) in French, but certainly not `1,000`, so the test suite is failing and I got the right results by chance :)

muffinresearch:
Where did your node come from? I think we are mostly using node installed via nvm on OSX. If yours is packaged differently it might explain the differences,

willdurand:
Homebrew

tofumatt:
Same, actually, that’s when I started to hit the bug.

nvm actually adds a LOT of load time to my shell initialisation (on every new shell) so I stopped using it. I’m guessing `nvm` just doesn’t care about locales (as it’s largely a development tool) but homebrew’s version does.

willdurand:
I reverted my second attempt to reveal the bug.

I use a Node version that is compiled with `system-icu`(`icu4c`), which might explain why locales are supported and the code behaves as intended (once fixed).

willdurand:
Required [full-icu-npm](https://github.com/unicode-org/full-icu-npm) (found by @muffinresearch) and it seems better now.

muffinresearch:
Using my nvm node to run tests locally the tests fail because the node used by the jest runner isn't picking up the icu package.

sh

willdurand:
> Using my nvm node to run tests locally the tests fail because the node used by the jest runner isn't picking up the icu package.

😞  Weird that Travis does not catch it though. It also uses `nvm`..

muffinresearch:
> 😞 Weird that Travis does not catch it though. It also uses nvm..

That's because travis is running this: https://github.com/mozilla/addons-frontend/pull/2946/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L106

willdurand:
Doh! totally makes sense indeed.

willdurand:
So, I should add `npm run node` before all `jest` calls.

package.json
```diff
@@ -25,20 +25,22 @@
...
+    "node": "NODE_PATH='./:./src' NODE_ICU_DATA='./node_modules/full-icu' node",
+    "start": "npm run version-check && npm run node bin/server.js",
+    "start-func-test-server": "npm run node bin/start-func-test-server.js",
     "test-ci": "bin/config-check.js && better-npm-run test-ci && cat ./coverage/lcov.info | coveralls",
     "test": "bin/config-check.js && jest --watch",
     "test-coverage": "bin/config-check.js && jest --coverage --watch",
     "test-coverage-once": "bin/config-check.js && jest --coverage",
     "test-once": "bin/config-check.js && jest && npm run lint",
-    "version-check": "node bin/version-check.js",
+    "version-check": "npm run node bin/version-check.js",
```
muffinresearch: Since this just checks the version it probably doesn't really need to run with icu support.

willdurand: Yup.

muffinresearch:
> So, I should add npm run node before all jest calls.

I think if you use better-npm-run (as per the test-ci command) you can export the env var and when node is used in jest it should do the right thing.

willdurand:
> I think if you use better-npm-run (as per the test-ci command) you can export the env var and when node is used in jest it should do the right thing.

Should I move the `node` script in a `better-npm-run` script too?

willdurand:
@muffinresearch does that work better for you now?

muffinresearch:
> Should I move the node script in a better-npm-run script too?

Might be cleaner whilst we continue to use better-npm-run.

> @muffinresearch does that work better for you now?

Yep `yarn test`, `yarn test-once`  and  `yarn test-coverage` looks good. 👍

willdurand:
Ok I guess I got it right now (hopefully).

willdurand:
ping @jasonthomas and @bqbn. If the app is not started with `yarn start`, you may want to install a Node version with full icu on the servers.

kumar303:
> If the app is not started with yarn start, you may want to install a Node version with full icu on the servers.

The production site gets started using [`npm start`](https://github.com/mozilla/addons-frontend/blob/master/Dockerfile#L34) but this should accomplish the same thing as `yarn start`; that is, it will [set `NODE_ICU_DATA` as needed](https://github.com/mozilla/addons-frontend/blob/0b10967f94db00ef1b2df8358f91c7ef31edb50e/package.json#L28).

Thanks for tracking this down!

PR #2952, opened 2017-08-16
Show loading indicators when switching between landing pages

### LINKED ISSUE #2882: Loading indicators aren't show when switching between landing pages
Sorry I didn't catch this when doing local testing, I must not have noticed as I was on a fast enough connection that blinking would hide the fact that no loading bars appeared.

If you are on a landing page (`/extensions/` or `/themes/`) then navigate to the other landing page, loading indicators are not shown and the old content remains until the new content is loaded from the server.

### What happened?
The old content remains until the new content loads.

There to be an intermediate loading state while the route changes but the load is taking place.

### Screencasts

Loading animation happens on first landing page load, but not on subsequent loads:

willdurand:
Fixes #2882 

---

This PR enables the loader when we change from one landing page to another by resetting the add-ons sets when a new request is triggered. In addition, I fixed a visual glitch by reducing the size of the loader related to the number of users (a width of `100%` put the loader on a new line).

### DIFF
--- a/tests/unit/amo/reducers/testLanding.js
+++ b/tests/unit/amo/reducers/testLanding.js
@@ -1,6 +1,7 @@
 import { getLanding } from 'amo/actions/landing';
 import landing, { initialState } from 'amo/reducers/landing';
 import { ADDON_TYPE_THEME } from 'core/constants';
+import { fakeAddon } from 'tests/unit/amo/helpers';
 
 
 describe('landing reducer', () => {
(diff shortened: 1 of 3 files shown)


tests/unit/amo/reducers/testLanding.js
```diff
@@ -53,6 +53,30 @@ describe('landing reducer', () => {
 
       expect(state.resultsLoaded).toEqual(false);
     });
+
+    it('resets each set of add-ons', () => {
+      const entities = {
+        addons: {
+          bar: { slug: 'bar' },
```
tofumatt: I think it would be best to use actual add-ons here rather than the stubs. It should be as easy as using `import { fakeAddon } from 'tests/unit/amo/helpers';` and doing: `bar: { ...fakeAddon, slug: 'bar' },`.

Just in case the reducers start to rely more on the structure of the add-on.

willdurand: Oh nice, I did not know about it sorry 😕

tofumatt:
No worries :-)

Issue #2912, opened 2017-08-07
Pass `appversion` when searching, to only see add-ons compatible with the client's version


diox:
When doing a search on addons-frontend, to follow addons-server behavior, we should pass the `appversion` parameter. That will enable filtering on the search API and only return compatible add-ons as per [compatibility rules](https://github.com/mozilla/addons-frontend/issues/2911#issue-248445514). Per mozilla/addons-server#5482 however, this should only be applied if the client is Firefox/Firefox for Android 57 or newer.

Note that in addition, addons-server search has a sidebar with versions on the left-hand side of the page. I am not sure whether or not this is supposed to stay in the new site - I personally don't feel it's very useful: the only versions present are the ones listed in the results themselves - it only allows you to refine a search further, but that does not bring any value to the user (who does not care about whether the add-ons presented are compatible with say, Firefox 4 or not...)

Issue #1250, opened 2016-10-18
Convert setState() to redux actions


kumar303:
Some components use `this.setState()` for various things. Here's an example of showing an error:

```
this.setState({ error: 'you must enter a blah before blah' });
```

This is OK but we completely lose any benfits of Redux, such as:
- time travel, which is nice for debugging state changes 
- error diagnostics, which is something I think we should add in the future
  - i.e. we can post the stack of redux state changes to an API endpoint from the client and replay the state changes on our local machines to reproduce the bug.

Let's switch `this.setState({ error: '...' })` to `this.props.setUIState({ error: '...' })` via `withUIState()` (see https://github.com/mozilla/addons-frontend/issues/5343).

This may end up sharing functionality with https://github.com/mozilla/addons-frontend/issues/1032 but it will be a little different.

muffinresearch:
We should also review other use of `setState`

kumar303:
One exception is using `setState` for typing into a text box. Using redux for this would probably be overkill.

kumar303:
Maybe do this one at the same time too https://github.com/mozilla/addons-frontend/issues/2345

tofumatt:
https://github.com/mozilla/addons-frontend/pull/2995#discussion_r135146819 is another place we should convert in the future.

willdurand:
Given that we have issues for each occurrence where it makes sense to use the `withUIState()` HOC, I am closing this issue now.

PR #2958, opened 2017-08-16
Search hints/autocomplete reducer/sagas

### LINKED ISSUE #2747: Add search hints/autocomplete
Add a search hint/autocomplete as per mocks. Relies on https://github.com/mozilla/addons-server/issues/3545.

willdurand:
Refs #2747

---

This PR adds reducer/sagas/api for the autocomplete/suggestion/hint search feature.

## TODO

- [x] Add a saga
- [x] Add a reducer
- [x] Implement API call in `core/api`
- [x] Tests

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -308,3 +308,22 @@ export function logOutFromServer({ api }: {| api: ApiStateType |}) {
     state: api,
   });
 }
+
+type AutocompleteParams = {|
+  api: ApiStateType,
+  filters: {|
+    query: string,
+    addonType?: string,
+  |},
+|};
+
+export function autocomplete({ api, filters }: AutocompleteParams) {
+  return callApi({
+    endpoint: 'addons/autocomplete',
+    params: {
+      app: api.clientApp,
+      ...convertFiltersToQueryParams(filters),
+    },
+    state: api,
+  });
+}
--- a/src/core/reducers/autocomplete.js
+++ b/src/core/reducers/autocomplete.js
@@ -1,6 +1,6 @@
-export const AUTOCOMPLETE_LOADED = 'core/autocomplete/AUTOCOMPLETE_LOADED';
-export const AUTOCOMPLETE_STARTED = 'core/autocomplete/AUTOCOMPLETE_STARTED';
-export const AUTOCOMPLETE_CANCELLED = 'core/autocomplete/AUTOCOMPLETE_CANCELLED';
+export const AUTOCOMPLETE_LOADED = 'AUTOCOMPLETE_LOADED';
+export const AUTOCOMPLETE_STARTED = 'AUTOCOMPLETE_STARTED';
+export const AUTOCOMPLETE_CANCELLED = 'AUTOCOMPLETE_CANCELLED';
 
 const initialState = {
   loading: false,
@@ -42,16 +42,11 @@ export default function reducer(state = initialState, action = {}) {
 
   switch (action.type) {
     case AUTOCOMPLETE_CANCELLED:
-      return {
-        ...state,
-        loading: false,
-        suggestions: [],
-      };
+      return initialState;
     case AUTOCOMPLETE_STARTED:
       return {
-        ...state,
+        ...initialState,
         loading: true,
-        suggestions: [],
       };
     case AUTOCOMPLETE_LOADED:
       return {
--- a/src/core/reducers/autocomplete.js
+++ b/src/core/reducers/autocomplete.js
@@ -1,0 +1,69 @@
+export const AUTOCOMPLETE_LOADED = 'core/autocomplete/AUTOCOMPLETE_LOADED';
+export const AUTOCOMPLETE_STARTED = 'core/autocomplete/AUTOCOMPLETE_STARTED';
+export const AUTOCOMPLETE_CANCELLED = 'core/autocomplete/AUTOCOMPLETE_CANCELLED';
+
+const initialState = {
+  loading: false,
+  suggestions: [],
+};
+
+export function autocompleteStart({ errorHandlerId, filters }) {
+  if (!errorHandlerId) {
+    throw new Error('errorHandlerId is required');
+  }
+
+  if (!filters) {
+    throw new Error('filters are required');
+  }
+
+  return {
+    type: AUTOCOMPLETE_STARTED,
+    payload: { errorHandlerId, filters },
+  };
+}
+
+export function autocompleteCancel() {
+  return { type: AUTOCOMPLETE_CANCELLED };
+}
+
+export function autocompleteLoad({ results }) {
+  if (!results) {
+    throw new Error('results are required');
+  }
+
+  return {
+    type: AUTOCOMPLETE_LOADED,
+    payload: { results },
+  };
+}
+
+export default function reducer(state = initialState, action = {}) {
+  const { payload } = action;
+
+  switch (action.type) {
+    case AUTOCOMPLETE_CANCELLED:
+      return {
+        ...state,
+        loading: false,
+        suggestions: [],
+      };
+    case AUTOCOMPLETE_STARTED:
+      return {
+        ...state,
+        loading: true,
+        suggestions: [],
+      };
+    case AUTOCOMPLETE_LOADED:
+      return {
+        ...state,
+        loading: false,
+        suggestions: payload.results
+        // TODO: Remove this when `null` names are not returned. See:
+        // https://github.com/mozilla/addons-server/issues/6189
+          .filter((result) => result.name !== null)
+          .map((result) => result.name),
+      };
+    default:
+      return state;
+  }
+}
--- a/src/core/sagas/autocomplete.js
+++ b/src/core/sagas/autocomplete.js
@@ -1,0 +1,40 @@
+// Disabled because of
+// https://github.com/benmosher/eslint-plugin-import/issues/793
+/* eslint-disable import/order */
+import { call, put, select, takeLatest } from 'redux-saga/effects';
+/* eslint-enable import/order */
+
+import { AUTOCOMPLETE_STARTED, autocompleteLoad } from 'core/reducers/autocomplete';
+import { autocomplete as autocompleteApi } from 'core/api';
+import log from 'core/logger';
+import { createErrorHandler, getState } from 'core/sagas/utils';
+
+
+export function* fetchAutocompleteResults({ payload }) {
+  const { errorHandlerId } = payload;
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const { filters } = payload;
+
+    const state = yield select(getState);
+
+    const response = yield call(autocompleteApi, {
+      api: state.api,
+      auth: state.auth,
+      filters,
+    });
+    const { results } = response;
+
+    yield put(autocompleteLoad({ results }));
+  } catch (error) {
+    log.warn(`Autcomplete results failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* autocompleteSaga() {
+  yield takeLatest(AUTOCOMPLETE_STARTED, fetchAutocompleteResults);
+}
--- a/src/core/sagas/autocomplete.js
+++ b/src/core/sagas/autocomplete.js
@@ -23,7 +23,6 @@ export function* fetchAutocompleteResults({ payload }) {
 
     const response = yield call(autocompleteApi, {
       api: state.api,
-      auth: state.auth,
       filters,
     });
     const { results } = response;
--- a/tests/unit/amo/helpers.js
+++ b/tests/unit/amo/helpers.js
@@ -138,11 +138,11 @@ export function createAddonsApiResult(results) {
   return normalize({ results }, { results: [addonSchema] });
 }
 
-export function createFakeAutocompleteResult({ name, lang = null }) {
+export function createFakeAutocompleteResult({ name = 'suggestion-result' } = {}) {
   return {
     id: Date.now(),
     icon_url: `https://example.org/${name}.png`,
-    name: lang ? { [lang]: name } : name,
-    url: `https://example.org/${lang}/firefox/addons/${name}/`,
+    name,
+    url: `https://example.org/en-US/firefox/addons/${name}/`,
   };
 }
--- a/tests/unit/amo/helpers.js
+++ b/tests/unit/amo/helpers.js
@@ -137,3 +137,12 @@ export function createAddonsApiResult(results) {
   // For example: core.api.featured(), core.api.search()...
   return normalize({ results }, { results: [addonSchema] });
 }
+
+export function createFakeAutocompleteResult({ name, lang = null }) {
+  return {
+    id: Date.now(),
+    icon_url: `https://example.org/${name}.png`,
+    name: lang ? { [lang]: name } : name,
+    url: `https://example.org/${lang}/firefox/addons/${name}/`,
+  };
+}
--- a/tests/unit/core/api/test_api.js
+++ b/tests/unit/core/api/test_api.js
@@ -231,7 +231,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.search({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         auth: true,
         filters: { page: parsePage(3), query: 'foo' },
       })
@@ -303,7 +303,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.search({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         filters: {
           addonType: ADDON_TYPE_THEME,
           clientApp: CLIENT_APP_ANDROID,
@@ -320,7 +320,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.search({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         filters: { clientApp: 'firefox', page: parsePage(3), query: 'foo' },
       })
         .then(() => mockWindow.verify());
@@ -362,7 +362,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.featured({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         filters: { addonType: ADDON_TYPE_THEME },
       })
         .then((response) => {
@@ -583,7 +583,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.categories({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
       })
         .then(() => mockWindow.verify());
     });
@@ -623,7 +623,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.autocomplete({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         filters: {
           query: 'foo',
         },
@@ -637,7 +637,7 @@ describe('api', () => {
         .once()
         .returns(mockResponse());
       return api.autocomplete({
-        api: { clientApp: 'android', lang: 'en-US' },
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
         filters: {
           query: 'foo',
           addonType: ADDON_TYPE_THEME,
--- a/tests/unit/core/api/test_api.js
+++ b/tests/unit/core/api/test_api.js
@@ -13,6 +13,7 @@ import {
   unexpectedSuccess,
   userAuthToken,
 } from 'tests/unit/helpers';
+import { createFakeAutocompleteResult } from 'tests/unit/amo/helpers';
 
 
 export function generateHeaders(
@@ -601,6 +602,47 @@ describe('api', () => {
         .once()
         .returns(mockResponse);
       return api.logOutFromServer({ api: signedInApiState })
+        .then(() => mockWindow.verify());
+    });
+  });
+
+  describe('autocomplete api', () => {
+    const mockResponse = () => createApiResponse({
+      jsonData: {
+        results: [
+          createFakeAutocompleteResult({ name: 'foo' }),
+          createFakeAutocompleteResult({ name: 'food' }),
+          createFakeAutocompleteResult({ name: 'football' }),
+        ],
+      },
+    });
+
+    it('sets the app, lang, and query', () => {
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/addons/autocomplete/?app=android&q=foo&lang=en-US`)
+        .once()
+        .returns(mockResponse());
+      return api.autocomplete({
+        api: { clientApp: 'android', lang: 'en-US' },
+        filters: {
+          query: 'foo',
+        },
+      })
+        .then(() => mockWindow.verify());
+    });
+
+    it('optionally takes addon type as filter', () => {
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/addons/autocomplete/?app=android&q=foo&type=persona&lang=en-US`)
+        .once()
+        .returns(mockResponse());
+      return api.autocomplete({
+        api: { clientApp: 'android', lang: 'en-US' },
+        filters: {
+          query: 'foo',
+          addonType: ADDON_TYPE_THEME,
+        },
+      })
         .then(() => mockWindow.verify());
     });
   });
--- a/tests/unit/core/reducers/test_autocomplete.js
+++ b/tests/unit/core/reducers/test_autocomplete.js
@@ -1,0 +1,94 @@
+import reducer, {
+  autocompleteCancel,
+  autocompleteLoad,
+  autocompleteStart,
+  AUTOCOMPLETE_STARTED,
+  AUTOCOMPLETE_LOADED,
+} from 'core/reducers/autocomplete';
+import { createFakeAutocompleteResult } from 'tests/unit/amo/helpers';
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('initializes properly', () => {
+      const { loading, suggestions } = reducer(undefined);
+      expect(loading).toBe(false);
+      expect(suggestions).toEqual([]);
+    });
+
+    it('handles AUTOCOMPLETE_CANCELLED', () => {
+      const { loading, suggestions } = reducer(undefined, autocompleteCancel());
+      expect(loading).toBe(false);
+      expect(suggestions).toEqual([]);
+    });
+
+    it('handles AUTOCOMPLETE_STARTED', () => {
+      const { loading, suggestions } = reducer(undefined, { type: AUTOCOMPLETE_STARTED });
+      expect(loading).toBe(true);
+      expect(suggestions).toEqual([]);
+    });
+
+    it('handles AUTOCOMPLETE_LOADED', () => {
+      const payload = {
+        results: [
+          createFakeAutocompleteResult({ name: 'foo' }),
+          createFakeAutocompleteResult({ name: 'bar' }),
+          createFakeAutocompleteResult({ name: 'baz' }),
+        ],
+      };
+
+      const { loading, suggestions } = reducer(undefined, {
+        type: AUTOCOMPLETE_LOADED,
+        payload,
+      });
+      expect(loading).toBe(false);
+      expect(suggestions).toEqual(['foo', 'bar', 'baz']);
+    });
+
+    it('excludes AUTOCOMPLETE_LOADED results with null names', () => {
+      const payload = {
+        results: [
+          createFakeAutocompleteResult({ name: 'foo' }),
+          createFakeAutocompleteResult({ name: null }),
+          createFakeAutocompleteResult({ name: 'baz' }),
+        ],
+      };
+
+      const { loading, suggestions } = reducer(undefined, {
+        type: AUTOCOMPLETE_LOADED,
+        payload,
+      });
+      expect(loading).toBe(false);
+      expect(suggestions).toEqual(['foo', 'baz']);
+    });
+  });
+
+  describe('autocompleteStart()', () => {
+    it('throws an error if arguments are missing', () => {
+      expect(() => {
+        autocompleteStart({});
+      }).toThrow(/errorHandlerId is required/);
+      expect(() => {
+        autocompleteStart({ errorHandlerId: 'id' });
+      }).toThrow(/filters are required/);
+    });
+
+    it('constructs the AUTOCOMPLETE_STARTED action', () => {
+      const payload = {
+        errorHandlerId: 'id',
+        filters: {},
+      };
+      expect(autocompleteStart(payload)).toEqual({
+        type: AUTOCOMPLETE_STARTED,
+        payload,
+      });
+    });
+  });
+
+  describe('autocompleteLoad()', () => {
+    it('throws an error if arguments are missing', () => {
+      expect(() => {
+        autocompleteLoad({});
+      }).toThrow(/results are required/);
+    });
+  });
+});
--- a/tests/unit/core/reducers/test_autocomplete.js
+++ b/tests/unit/core/reducers/test_autocomplete.js
@@ -2,8 +2,6 @@ import reducer, {
   autocompleteCancel,
   autocompleteLoad,
   autocompleteStart,
-  AUTOCOMPLETE_STARTED,
-  AUTOCOMPLETE_LOADED,
 } from 'core/reducers/autocomplete';
 import { createFakeAutocompleteResult } from 'tests/unit/amo/helpers';
 
@@ -15,6 +13,12 @@ describe(__filename, () => {
       expect(suggestions).toEqual([]);
     });
 
+    it('ignore unrelated actions', () => {
+      const state = { loading: false, suggestions: ['foo'] };
+      const newState = reducer(state, { type: 'UNRELATED' });
+      expect(newState).toEqual(state);
+    });
+
     it('handles AUTOCOMPLETE_CANCELLED', () => {
       const { loading, suggestions } = reducer(undefined, autocompleteCancel());
       expect(loading).toBe(false);
@@ -22,41 +26,34 @@ describe(__filename, () => {
     });
 
     it('handles AUTOCOMPLETE_STARTED', () => {
-      const { loading, suggestions } = reducer(undefined, { type: AUTOCOMPLETE_STARTED });
+      const { loading, suggestions } = reducer(undefined, autocompleteStart({
+        errorHandlerId: 'any-error-handler-id',
+        filters: { q: 'search string' },
+      }));
       expect(loading).toBe(true);
       expect(suggestions).toEqual([]);
     });
 
     it('handles AUTOCOMPLETE_LOADED', () => {
-      const payload = {
-        results: [
-          createFakeAutocompleteResult({ name: 'foo' }),
-          createFakeAutocompleteResult({ name: 'bar' }),
-          createFakeAutocompleteResult({ name: 'baz' }),
-        ],
-      };
+      const results = [
+        createFakeAutocompleteResult({ name: 'foo' }),
+        createFakeAutocompleteResult({ name: 'bar' }),
+        createFakeAutocompleteResult({ name: 'baz' }),
+      ];
 
-      const { loading, suggestions } = reducer(undefined, {
-        type: AUTOCOMPLETE_LOADED,
-        payload,
-      });
+      const { loading, suggestions } = reducer(undefined, autocompleteLoad({ results }));
       expect(loading).toBe(false);
       expect(suggestions).toEqual(['foo', 'bar', 'baz']);
     });
 
     it('excludes AUTOCOMPLETE_LOADED results with null names', () => {
-      const payload = {
-        results: [
-          createFakeAutocompleteResult({ name: 'foo' }),
-          createFakeAutocompleteResult({ name: null }),
-          createFakeAutocompleteResult({ name: 'baz' }),
-        ],
-      };
+      const results = [
+        createFakeAutocompleteResult({ name: 'foo' }),
+        createFakeAutocompleteResult({ name: null }),
+        createFakeAutocompleteResult({ name: 'baz' }),
+      ];
 
-      const { loading, suggestions } = reducer(undefined, {
-        type: AUTOCOMPLETE_LOADED,
-        payload,
-      });
+      const { loading, suggestions } = reducer(undefined, autocompleteLoad({ results }));
       expect(loading).toBe(false);
       expect(suggestions).toEqual(['foo', 'baz']);
     });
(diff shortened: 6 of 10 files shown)


willdurand:
@tofumatt @kumar303 as you might have seen, I picked #2747 to start working on new layers of the app. I'd like your input if possible to know if I go in the right direction.

willdurand:
Ok so it is almost implemented (no tests..). I have updated the screenshot.

The UX is not that great since the Autocomplete API returns heterogeneous results. I would expect smaller and smaller result sets as I type new letters.

CI complains because of the lack of tests but I'd like to get input on the architecture. Bonus: I introduced the first [ducks](https://github.com/erikras/ducks-modular-redux). There are only one reducer and one saga.

src/amo/components/SearchForm.js
```diff
@@ -102,10 +158,12 @@ export function mapStateToProps(state) {
     addonType: validAddonTypes.includes(state.viewContext.context) ?
       state.viewContext.context : null,
     api: state.api,
+    suggestions: state.autocomplete.suggestions,
   };
 }
 
 export default compose(
+  withErrorHandler({ name: 'Autocomplete' }),
```
willdurand: Might not be the right name here. Should it be `Search` or `SearchFormBase` instead?

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,50 @@
...
+    type: AUTOCOMPLETE_LOADED,
+    payload: { results },
+  };
+}
+
+export default function reducer(state = initialState, action) {
+  const { payload } = action;
+  switch (action.type) {
+    case AUTOCOMPLETE_STARTED:
+      return state;
```
willdurand: At first, I wanted to reset the `suggestions` but then the list disappears and then appears again, which was not nice..

src/ui/components/SuggestedResults/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+import './style.scss';
+
+type Props = {
+  query: string,
+  suggestions: Array<string>,
+  onSelect: Function,
+};
+
+const renderSuggestion = (query: string, suggestion: string, onSelect: Function) => {
```
willdurand: This could maybe move to a separate named component inside the `SuggestedResults` one.

src/amo/components/SearchForm.js
```diff
@@ -23,12 +26,31 @@ export class SearchFormBase extends React.Component {
...
   }
 
   static contextTypes = {
     router: PropTypes.object,
   }
 
+  constructor(props: Object) {
+    super(props);
+
+    this.state = {
```
tofumatt: We tend not to use internal component state for anything. I know we had an issue open about removing `setState` usage from existing components (#1250) which I'm not sure is finished, but I don't think we should add any new `setState` usage. I'd rather everything be in the redux store instead.

src/amo/components/SearchForm.js
tofumatt: `SearchForm` seems like the right name; I think using the component's name (the same name we use for the file) is our usual pattern.

src/core/reducers/autocomplete.js
tofumatt: I think resetting the suggestions would be good as I am often frustrated by autocompletes that leave old results on screen as I search. Perhaps we should just have a more robust loading screen. Instead of [returning `null` here](https://github.com/mozilla/addons-frontend/pull/2958/files#diff-c2d5de2068dbee8f3313dfa396c573b0R94) you could make the suggestions an array of `LoadingText` components as we do elsewhere in the app if there is a search happening.

I think if the query is deleted and the search hides/appears that's fine.

Let me know if that makes sense.

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,50 @@
...
+    case AUTOCOMPLETE_LOADED:
+      return {
+        ...state,
+        suggestions: payload.results
+          // TODO: cf. https://github.com/mozilla/addons-server/issues/6189
```
tofumatt: I am such an uncultured Canadian, I had to look up "cf." 😅

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,50 @@
+export const AUTOCOMPLETE_LOADED = 'core/autocomplete/AUTOCOMPLETE_LOADED';
```
tofumatt: We've tended to put these in the constants files in the past.

Though honestly this is probably way nicer.

Not sure about the path though; I don't think it is really needed/helpful because it's easy to grep for when debugging and it just adds size to the app/console output/etc.

src/ui/components/SearchInput/index.js
```diff
@@ -13,33 +13,47 @@ export default class SearchInput extends React.Component {
...
   componentDidMount() {
     this.setIconPosition();
     window.addEventListener('resize', this.setIconPosition);
   }
 
+  componentWillReceiveProps(nextProps) {
+    const { defaultValue } = this.props;
+
+    if (defaultValue !== nextProps.defaultValue) {
+      this.setState({ value: nextProps.defaultValue || '' });
```
tofumatt: Again, I think we should avoid `setState` and just use redux instead.

src/ui/components/SuggestedResults/index.js
tofumatt: I'm gonna guess this is an unpopular opinion but I don't like stateless components aesthetically nor do I think we should be using them. I get it's a bit more boilerplate to write a class-based component but it's what we do most of the time and unless the entire app is some functional-paradise (for me, a nightmare!) I don't think it's worth selectively using this way of making components. I expect to see a function not a component when I see this syntax.

src/ui/components/SuggestedResults/style.scss
```diff
@@ -0,0 +1,34 @@
+@import "~ui/css/vars";
+$item-padding: 5px 52px;
+
+.SuggestedResults {
+  background-color: $white;
+  color: $link-color;
+  padding-bottom: 10px;
+  padding-top: 10px;
+
+  &-list {
```
tofumatt: Huh. I didn't know about this but it makes sense.

That said I think explicitly writing `.SuggestedResults-item { [...styles] }` is better: it's easier to grep for and easier to immediately understand.

src/ui/components/SuggestedResults/style.scss
```diff
@@ -0,0 +1,34 @@
...
+    text-align: left;
+    width: 100%;
+
+    &:hover,
+    &:active,
+    &:focus {
+      background-color: $background-gray;
+    }
+
+    .is-highlighted {
```
tofumatt: I think just `highlighted` is fine, though we've tended to use very-scoped (and admittedly verbose) selectors for things like this: `.SearchResults-item--highlighted` is the usual approach, the double dashes signifying it's a utility/modifier instead of a different element.

src/ui/components/SuggestedResults/index.js
```diff
@@ -0,0 +1,54 @@
...
+  let parts;
+  if (query.replace(/\s+/, '') === '') {
+    parts = [suggestion];
+  } else {
+    parts = suggestion.split(new RegExp(`(${escapeStringRegexp(query)})`, 'gi'));
```
tofumatt: Sorry, I'm not immediately understanding what this does. Could you explain/add some documentation to it?

src/core/api/index.js
```diff
@@ -308,3 +308,19 @@ export function logOutFromServer({ api }: {| api: ApiStateType |}) {
...
+
+type AutocompleteParams = {|
+  api: ApiStateType,
+  query: string,
+|};
+
+export function autocomplete({ api, query }: AutocompleteParams) {
+  return callApi({
+    endpoint: 'addons/autocomplete',
+    params: {
```
tofumatt: The [autocomplete API](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#get--api-v3-addons-autocomplete-) also supports `addonType` which we should set if the `viewContext` is `ADDON_TYPE_EXTENSION`, `ADDON_TYPE_THEME`, etc. The intent of the "sections" of the site "Explore, Extensions, Themes, etc." is that searches are scoped to that add-on type when you're in that section (at least by default). So only when on the homepage/"Explore" should you search all add-on types.

Also we have a bug open for it for regular search but we might as well use `appversion` here so only compatible add-ons are returned.

src/core/reducers/autocomplete.js
kumar303: With the ducks pattern, I think it's more helpful if they remain in this file. That way you don't have to open a new file to reference the value (since this file will be the only place they are ever referenced).

willdurand:
@tofumatt thank you so much, that is very helpful! 

Do you have an idea on how the suggestions should behave on desktop? I would expect to use the arrows. The current component was my first attempt, but mobile first.. I played with react-autosuggest and it seems like a good fit. Is it ok to use third-party components?

For Scss, should I leverage its features like `&`? Or I stick to full classnames? 

I do not really understand the choices regarding functional components and no UI state. I tend to only write functional components 🙁 Class-based components are useful for lifecycle methods and UI state, but you say we should not use UI state. For me there are two different states: app state (data) and UI state.

willdurand:
(ooops wrong button)

src/core/sagas/autocomplete.js
```diff
@@ -0,0 +1,40 @@
...
+
+    yield put(autocompleteLoad({ results }));
+  } catch (error) {
+    log.warn(`Autcomplete results failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* autocompleteSaga() {
+  yield takeLatest(AUTOCOMPLETE_STARTED, fetchAutocompleteResults);
```
kumar303: You may need to debounce this (there's a [recipe](https://redux-saga.js.org/docs/recipes/) for it) but that can be tweaked later too.

src/core/reducers/autocomplete.js
willdurand: 👍

src/core/reducers/autocomplete.js
willdurand: Yes, this was to show how to use the ducks pattern. It must be exported here for the saga.

src/core/reducers/autocomplete.js
willdurand: Oops sorry for that. Would `refs` be better?

src/ui/components/SearchInput/index.js
kumar303: One thing that might work better as `setState` is typing into a text box (like for this use case). The way React does it is every character changes the state of the component. This might be a little bit overkill for Redux since each character typed would trigger every reducer function in our app.

src/ui/components/SearchInput/index.js
```diff
@@ -73,7 +93,7 @@ export default class SearchInput extends React.Component {
   }
 
   render() {
-    const { className, name, placeholder, ...props } = this.props;
```
kumar303: thanks for expanding the rest/spread!

src/ui/components/SearchInput/index.js
willdurand: I have never seen a controlled input being connected directly to redux I must say. That is why I think there is a UI state and a app state (redux).

src/ui/components/SearchInput/index.js
kumar303: You should double check every component that was using `SearchInput` previously to make sure you expanded everything that was in use.

src/core/reducers/autocomplete.js
tofumatt: Yeah, to be clear I definitely prefer them this way–in the file.

src/ui/components/SuggestedResults/index.js
kumar303: functional components are fine by me (it's less code thus easier to read) but capitalizing it as `RenderSuggestion` would be better to follow our component naming conventions.

src/core/reducers/autocomplete.js
tofumatt: I think something more complete would be better: ``// TODO: Remove this when `null` names are not returned. See: https://github.com/mozilla/addons-server/issues/6189``

That's assuming I understand this correctly and that's what this is saying 😄

src/ui/components/SearchInput/index.js
tofumatt: Oh, hmm, good point actually. Okay 👍

src/ui/components/SuggestedResults/index.js
```diff
@@ -0,0 +1,54 @@
+/* @flow */
+/* eslint react/no-array-index-key: 0 */
+import React from 'react';
+import escapeStringRegexp from 'escape-string-regexp';
+
+import './style.scss';
+
+type Props = {
```
kumar303: object types should always be exact if possible, i.e. using pipes:

```js
type Props = {|
  query: string,
|};
```

src/ui/components/SuggestedResults/index.js
kumar303: Oh, I see that this is not used like a traditional React component. I think it should be a traditional component where `query`, `suggestion`, and `onSelect` are its properties.

src/ui/components/SuggestedResults/index.js
```diff
@@ -0,0 +1,54 @@
...
+        })}
+      </button>
+    </li>
+  );
+};
+
+const SuggestedResults = ({ query, suggestions, onSelect }: Props) => (
+  <div className="SuggestedResults">
+    <ul className="SuggestedResults-list">
+      {suggestions.map((s) => renderSuggestion(query, s, onSelect))}
```
kumar303: As mentioned up above, I think this should be a React component, used like:

```js
<ul className="SuggestedResults-list">
  {suggestions.map((s) => <Result query={query} suggestion={s} onSelect={onSelect} />)}
</ul>
```

kumar303: Just as a separate point (for future reference), I prefer to see any function with more than one argument to take an object of parameters, like `renderSuggestion({ query, suggestion: s, onSelect }))`. It makes it easier to look at the function call and see what's going on.

tests/unit/amo/helpers.js
```diff
@@ -137,3 +137,14 @@ export function createAddonsApiResult(results) {
   // For example: core.api.featured(), core.api.search()...
   return normalize({ results }, { results: [addonSchema] });
 }
+
+export function createFakeAutocompleteResult(name, lang = 'en-US') {
```
kumar303: again, here, I'd like to functions using a single object for multiple parameters, like `createFakeAutocompleteResult({ name, lang = 'en-US' })`

kumar303:
As you mentioned, [react-autosuggest](https://github.com/moroshko/react-autosuggest) looks good, especially since it's ARIA compliant (which we strive for).

As for functional components, I agree that we should use them whenever we don't need lifecycle functions. We've seen too many bugs around improper `componentWillReceiveProps` logic and, as mentioned, they are the future of React that we can embrace today.

src/ui/components/SuggestedResults/index.js
willdurand: This is actually a simple function that returns a component indeed. I did that to highlight the query in the suggestions like it is depicted in the mockups.

src/core/reducers/autocomplete.js
willdurand: You got it right! 😃 

More seriously, sorry for that, it was not clear at all.

src/ui/components/SuggestedResults/index.js
willdurand: Oops, forgot about it. I have just updated my snippet too.

src/ui/components/SuggestedResults/index.js
willdurand: Roger that!

src/ui/components/SearchInput/index.js
willdurand: Yes, it seems not used elsewhere.

src/core/sagas/autocomplete.js
willdurand: I will do that!

willdurand:
@tofumatt I have refactored the whole feature and provided new Gifs. It is way better now, but I am not sure about the random loading widths. It feels more natural on small screens, but maybe too "blinky" on desktop.

willdurand:
Going to write tests now

tofumatt:
@willdurand Ah, agreed on the random widths! Fortunately the component accepts a `width` prop you can use to set the exact width of the component (in percentages, IIRC). I think the autocomplete results would look nicer with a uniform width so you can just set them all to be `width={70}` (or whatever width looks the best, feel free to experiment)

tofumatt:
Also I would say the grey background for the add-on that is selected looks odd when it's the first selection because the colour is very similar to the background and it makes the page look broken. I think the new mocks (they were *very* recently updated) have a different selector but at any rate I think it should be changed.

It also seems like the first result is selected/focused automatically during search which I don't think should be the case.

Largely though looks good!

willdurand:
I have updated the Gifs with a slightly better UI.

src/amo/components/SearchForm/Suggestion.js
```diff
@@ -0,0 +1,39 @@
+/* @flow */
+/* eslint react/no-array-index-key: 0 */
+import React from 'react';
+import escapeStringRegexp from 'escape-string-regexp';
+
+type Props = {|
+  search: string,
```
tofumatt: I think `searchQuery` or `query` might work better here because it's the term we use across the rest of the app for a string used to search the API. This is part of why I was confused by the code in `Suggestion`

src/amo/components/SearchForm/Suggestion.js
```diff
@@ -0,0 +1,39 @@
...
+  suggestion: string,
+|};
+
+const Suggestion = ({ search, suggestion }: Props) => {
+  let parts;
+  if (search.replace(/\s+/, '') === '') {
+    parts = [suggestion];
+  } else {
+    // split the suggestion by the search typed by the user so that it can be
```
tofumatt: Ahhhh, I see what this is doing now. I think using a word like "highlight" over identify will help illustrate what this is doing. A tweak to the docs could be:

```js
// Separate the existing search query text from the suggested search term
// returned from the autocomplete API. This allows us to highlight the search
// query already typed by the user in the text using the `.match-search` className
// below.
```

I'll admit I still don't understand what "since template strings are not by default" means.

src/amo/components/SearchForm/Suggestion.js
```diff
@@ -0,0 +1,39 @@
...
+    // escaped since template strings are not by default.
+    parts = suggestion.split(new RegExp(`(${escapeStringRegexp(search)})`, 'gi'));
+  }
+
+  return (
+    <p className="Suggestion">
+      {parts.map((text, idx) => {
+        if (text.toLowerCase() === search.toLowerCase()) {
+          return (
+            <span className="match-search" key={`${suggestion}-${text}=${idx}`}>
```
tofumatt: We don't have a tonne of typed-up guidelines but [we do have some for CSS](https://github.com/mozilla/addons-frontend/blob/master/docs/css.md). It would be great if you used component names to make `className`s easily searchable/identifiable that'd be great ^_^

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,211 @@
...
+import LoadingText from 'ui/components/LoadingText';
+import { withErrorHandler } from 'core/errorHandler';
+import {
+  autocompleteAbort,
+  autocompleteStart,
+} from 'core/reducers/autocomplete';
+import Suggestion from 'amo/components/SearchForm/Suggestion';
+
+import 'core/css/inc/lib.scss';
+import './style.scss';
```
tofumatt: We've been using `styles.scss` not `style.scss`.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,211 @@
...
+  handleSuggestionSelected = (e, { suggestionValue }) => {
+    this.setState({ search: suggestionValue }, () => {
+      this.goToSearch(suggestionValue);
+    });
+  }
+
+  renderSuggestion = (suggestion) => {
+    const { search } = this.state;
+
+    if (this.props.loadingSuggestions && suggestion === LOADING_SUGGESTION_KEY) {
```
tofumatt: Is `suggestion === LOADING_SUGGESTION_KEY` needed when the component is already marked as loading (and thus wanting to render `<LoadingText>`?

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,211 @@
...
+        className="SearchForm-form"
+        ref={(ref) => { this.form = ref; }}
+      >
+        <label
+          className="visually-hidden"
+          htmlFor="q"
+        >
+          {i18n.gettext('Search')}
+        </label>
+        <Autosuggest
```
tofumatt: So many props! 😄

src/amo/components/SearchForm/style.scss
```diff
@@ -0,0 +1,54 @@
...
+    color: $black;
+    margin: 0;
+    padding: 0;
+  }
+
+  &-item {
+    cursor: pointer;
+    outline: none;
+    padding: $item-padding;
+    text-align: left;
```
tofumatt: This should use the `@include text-align-start()` mixin so we get free RTL support.

src/amo/components/SearchForm/style.scss
```diff
@@ -0,0 +1,54 @@
...
+
+.SearchInput-input,
+.SearchForm-form {
+  color: $base-color;
+  font-size: $font-size-m-smaller;
+  white-space: nowrap;
+}
+
+.SearchForm-suggestions {
+  &-list {
```
tofumatt: As mentioned, I'm not a fan of this style of writing classNames because it makes it harder to search the exact className when searching through the codebase and it's less obvious to read. I like using `&` for states of a component like `:hover` but here it's a different element/className and I don't think the nesting here buys us anything; it's a tiny bit of shorthand but it makes it look like this relies on being a child element of `.SearchForm-suggestions` which it doesn't. We've been avoiding nesting to prevent future errors and keep selectors really basic.

Sorry if I said this already and missed a question but I think some of your questions got buried by the new commits 😞

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+    type: AUTOCOMPLETE_LOADED,
+    payload: { results },
+  };
+}
+
+export default function reducer(state = initialState, action = {}) {
+  const { payload } = action;
+
+  switch (action.type) {
+    case AUTOCOMPLETE_ABORTED:
```
tofumatt: This is a nitpick but `AUTOCOMPLETE_CANCELLED` would be a bit nicer; it's a less harsh, programmer-y type word. Technically aborted is the wrong word anyway: it means stopped prematurely *because of a fault*; this action happens because a user stops/changes their search.

I'm not trying to be a dictionary nerd, but we use `STARTED` instead of `INITIALISED` most of the time, I think `CANCELLED` beats `ABORTED`.

tests/unit/amo/components/TestSearchFormSuggestion.js
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  it('renders itself', () => {
+    const props = {};
+    const root = shallowComponent(props);
+
+    expect(root.find('.Suggestion')).toHaveLength(1);
+  });
+
+  it('highlights parts of the suggestion that match the search', () => {
```
tofumatt: "match the search term" or "match the search query" would make this spec a bit clearer.

tests/unit/core/reducers/test_autocomplete.js
```diff
@@ -0,0 +1,90 @@
...
+        payload,
+      });
+      expect(loading).toBe(false);
+      expect(suggestions).toEqual(['foo', 'baz']);
+    });
+  });
+
+  describe('autocompleteStart()', () => {
+    it('throws an error if arguments are missing', () => {
+      expect(() => autocompleteStart({})).toThrow(/errorHandlerId is required/);
```
tofumatt: I feel like we've been out of the habit of using implicit returns. I get they're shorter but I find them a fair bit harder to read, especially here in the test where we aren't focused on a function's *return value* but instead its *side effect* (throwing the exception). I feel like this test would be easier to immediately grok as:

```js
expect(() => {
  autocompleteStart({});
}).toThrow(/errorHanderId is required/);
```

I know it's more to write but I think it gives the code room to breathe (an aesthetic preference I'll admit) but also more accurately represents what the test is about.

I'd prefer no implicit returns because I don't like concise syntax like that, but I think they're really only useful when they fit on one line and are actually intended to return something.

tests/unit/core/reducers/test_autocomplete.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    it('constructs the AUTOCOMPLETE_STARTED action', () => {
+      const args = {
+        errorHandlerId: 'id',
+        filters: {},
+      };
+      expect(autocompleteStart(args)).toEqual({
+        type: AUTOCOMPLETE_STARTED,
+        payload: {
```
tofumatt: This could be a bit shorter, as it doesn't need the spreads. You could even do:

```js
const payload = {
  errorHandlerId: 'id',
  filters: {},
};
expect(autocompleteStart(payload)).toEqual({
  type: AUTOCOMPLETE_STARTED,
  payload,
});
```

tests/unit/core/reducers/test_autocomplete.js
```diff
@@ -0,0 +1,90 @@
...
+          ...args,
+        },
+      });
+    });
+  });
+
+  describe('autocompleteLoad()', () => {
+    it('throws an error if arguments are missing', () => {
+      expect(() => autocompleteLoad({})).toThrow(/results are required/);
```
tofumatt: Again, see my previous comments about not using the return values of implicit return functions.

tests/unit/ui/components/TestSearchInput.js
```diff
@@ -1,6 +1,7 @@
 /* global window */
 import React from 'react';
 import { renderIntoDocument, Simulate } from 'react-addons-test-utils';
+import sinon from 'sinon';
```
tofumatt: You shouldn't need to explicitly import `sinon`; it's loaded into our tests by default.

tofumatt:
Oh also: I'm working on #2912 and I think we should stop adding a bunch of files to `core/api`. You don't have to change this patch but in the future I think we shouldn't keep adding them to `index.js`.

tofumatt:
Sorry, I think the conversations in this thread got a bit confusing. For posterity I just want to add my thoughts/address your original questions.

> For Scss, should I leverage its features like &? Or I stick to full classnames?

I think sticking to full classnames is better/clearer. It makes the code easier to grep and means we don't have to nest things as much. When I see indents in (S)CSS I assume nested selectors; in the case of `&-item` selectors it's shorthand but not actual nesting/cascades, so I'd like to avoid that for clarity.

> I do not really understand the choices regarding functional components and no UI state. I tend to only write functional components 🙁 Class-based components are useful for lifecycle methods and UI state, but you say we should not use UI state. For me there are two different states: app state (data) and UI state.

I think @kumar303 made the point for functional components and I'm *okay* with using them but my argument for not using them is I find it a lot harder to read something as a component when it isn't extending `React.Component` -- it's less obvious at a glance if something is intended to be a component or is a function. I guess that's the whole point of functional components: "who cares as long as this is functional?" so maybe I'm just being daft, but I like the consistency around "all components look one way". The other reason, as @kumar303 mentioned, to use non-functional components is lifecycle methods (and we'll use them a fair bit with sagas), even when we tend to (mostly) stay away from component state. If a component will need lifecycle methods eventually it means refactoring it which is annoying to me.

Saying that "functional components are the future of React" implies to me that old-style components are going away or will be deprecated. I think it's better to state the concrete benefits of using them, which I have since found to be performance... which is a good enough reason for me.

Aesthetically and even just conceptually I find functional components less friendly but I think I'm a minority on that front not only on the team but in the React community in general. 😅 

Aside from performance I'm not convinced of any of the [wins of functional components](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc), but for UI elements I guess they do make sense and we'd be silly not to take advantage of that React optimisation. So I will refrain from complaints about their usage in the future 😄 

-----

Apologies that read sort of like a dialogue with myself where I realised functional components do have a benefit. I guess I'll leave it here for posterity. Seemingly I really only care about the SCSS classnames now 😉

tofumatt:
Also one last point: PropTypes suck for functional components but then I guess that's a good reason to use `flow` for PropTypes 😄

src/amo/components/SearchForm/style.scss
willdurand: Yes I was not 100% sure so I left things as is (_i.e._ stuck to my habits..). Thank you and sorry for reiterating this though.

src/amo/components/SearchForm/style.scss
willdurand: Awesome!

src/core/reducers/autocomplete.js
willdurand: Roger that! I had `STOPPED` in mind first, but was not so nice. I did not think about `CANCELLED` but what you say here makes a lot of sense 💪

tests/unit/core/reducers/test_autocomplete.js
willdurand: I do agree, and will now stick to that.

tests/unit/ui/components/TestSearchInput.js
willdurand: oh ok, cool!

willdurand:
> Oh also: I'm working on #2912 and I think we should stop adding a bunch of files to core/api. You don't have to change this patch but in the future I think we shouldn't keep adding them to index.js.

Agreed.

willdurand:
> Apologies that read sort of like a dialogue with myself where I realised functional components do have a benefit. I guess I'll leave it here for posterity. Seemingly I really only care about the SCSS classnames now 😉

Well, I wanted to start such a feature because I knew it would raise many questions/comments. I think I have collected enough information now to be in sync with both of you. I do apologize to take so much of your time for that..

willdurand:
> Overall I think this is looking really good. It's a big patch for sure: in the future it might be nice to land actions/reducers/sagas first and then the components. @kumar303 and I have done that in the past and it's been nice.

Thanks! I was not sure about how to ship this feature. I have been a bit confused by how you approach PRs, commits and the Git history. I'll send smaller diffs next time.

According to the mockups, the UI has changed. Maybe I should get rid of the UI changes in this PR (only keeping reducer/saga), and open a new one for UI. WDYT?

tofumatt:
No need to apologise! Dialogue is always good 😄

If it's not a huge pain to do the actions/reducers/sagas/api in one patch then the UI in the next that would be great! I think we have a good sense of the data the reducers should handle so it's fine to review them in isolation. Then we can review the component/UI changes separately 👍

tofumatt:
Oh, and if anything isn't clear and you think we should write it up clearly in docs let us know. I think it's a worthwhile exercise and as it's just documenting what we already know it shouldn't take ages. If you had questions or still do that weren't clear feel free to file an issue and I'll try my best to answer them and commit it as a contributing/process doc.

willdurand:
> If it's not a huge pain to do the actions/reducers/sagas/api in one patch then the UI in the next that would be great!

@tofumatt done :)

tests/unit/core/sagas/test_autocomplete.js
```diff
@@ -0,0 +1,79 @@
...
+    mockApi
+      .expects('autocomplete')
+      .once()
+      .returns(Promise.resolve({ results }));
+
+    _autocompleteStart({ filters });
+
+    // The saga should respond by dispatching the autocomplete loaded action.
+    await sagaTester.waitFor(AUTOCOMPLETE_LOADED);
+    mockApi.verify();
```
kumar303: This should follow up by asserting that `results` is passed to `autocompleteLoad()` correctly.

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+      };
+    case AUTOCOMPLETE_LOADED:
+      return {
+        ...state,
+        loading: false,
+        suggestions: payload.results
+        // TODO: Remove this when `null` names are not returned. See:
+        // https://github.com/mozilla/addons-server/issues/6189
+          .filter((result) => result.name !== null)
+          .map((result) => result.name),
```
kumar303: I haven't looked at the UI patch yet but why is this reduced to an array of names? Why not keep it as an array of `result` objects? I'm thinking that you'll at least need IDs.

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
+export const AUTOCOMPLETE_LOADED = 'core/autocomplete/AUTOCOMPLETE_LOADED';
+export const AUTOCOMPLETE_STARTED = 'core/autocomplete/AUTOCOMPLETE_STARTED';
+export const AUTOCOMPLETE_CANCELLED = 'core/autocomplete/AUTOCOMPLETE_CANCELLED';
+
+const initialState = {
+  loading: false,
+  suggestions: [],
```
kumar303: I haven't looked at the UI patch yet but defaulting `suggestions` to an empty array makes it look no different than an API response of zero results. Maybe that's okay but previously this has caused problems (especially due to server rendering) where we have no way to tell if an initial API request had been made or not. We would have solved this by defaulting `suggestions` to `null` to indicate that no API request has been made. I would suggest trying to work with that model unless you find yourself writing too much defensive code around it.

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+    payload: { results },
+  };
+}
+
+export default function reducer(state = initialState, action = {}) {
+  const { payload } = action;
+
+  switch (action.type) {
+    case AUTOCOMPLETE_CANCELLED:
+      return {
```
kumar303: since there's only one dimension to the data, you can just reset it:

```js
return initialState;
```

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
...
+        ...state,
+        loading: false,
+        suggestions: [],
+      };
+    case AUTOCOMPLETE_STARTED:
+      return {
```
kumar303: To avoid some repetition and guarantee that the state gets reset, this could be done as:

```js
return { ...initialState, loading: true };
```

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,69 @@
+export const AUTOCOMPLETE_LOADED = 'core/autocomplete/AUTOCOMPLETE_LOADED';
```
willdurand: That's Ducks style, but I agree, no need for prefix here.

src/core/reducers/autocomplete.js
willdurand: I am ok with that, it should not require too much extra code.

I did not do that because autocompletion looks like a "secondary" feature to me: if no result is shown to the user for some reasons, that's fine because she is typing in the search input anyway. Nothing is really broken because user is looking for something _via_ the search input and relying on it for quick/direct access to add-ons is still not natural/common (IMO).

src/core/reducers/autocomplete.js
willdurand: I am a bit reluctant to expose too many things in the state.

At first, I needed names only because that's what I had to display. In the UI PR, I have added the `url` and `iconUrl` because that's what I needed to match the new spec.

tests/unit/amo/helpers.js
```diff
@@ -137,3 +137,12 @@ export function createAddonsApiResult(results) {
   // For example: core.api.featured(), core.api.search()...
   return normalize({ results }, { results: [addonSchema] });
 }
+
+export function createFakeAutocompleteResult({ name, lang = null }) {
+  return {
+    id: Date.now(),
+    icon_url: `https://example.org/${name}.png`,
+    name: lang ? { [lang]: name } : name,
```
willdurand: > However, maybe you meant it to be a simulation of an add-on with multiple locales?

Exactly, removing now.

tests/unit/core/api/test_api.js
```diff
@@ -604,4 +605,45 @@ describe('api', () => {
...
+      },
+    });
+
+    it('sets the app, lang, and query', () => {
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/addons/autocomplete/?app=android&q=foo&lang=en-US`)
+        .once()
+        .returns(mockResponse());
+      return api.autocomplete({
+        api: { clientApp: 'android', lang: 'en-US' },
```
willdurand: Ok, replacing it in the whole file then.

tests/unit/core/sagas/test_autocomplete.js
willdurand: Indeed. I must admit I have mimicked the whole search feature and this is missing there too 🙈

tests/unit/core/sagas/test_autocomplete.js
```diff
@@ -0,0 +1,79 @@
...
+
+  beforeEach(() => {
+    errorHandler = createStubErrorHandler();
+    mockApi = sinon.mock(api);
+    const initialState = dispatchSignInActions().state;
+    sagaTester = new SagaTester({
+      initialState,
+      reducers: {
+        api: apiReducer,
+        auth: authReducer,
```
willdurand: nothing, removing it.

src/core/reducers/autocomplete.js
kumar303: ok, fixing it in the next patch is fine

src/core/reducers/autocomplete.js
```diff
@@ -0,0 +1,64 @@
+export const AUTOCOMPLETE_LOADED = 'AUTOCOMPLETE_LOADED';
+export const AUTOCOMPLETE_STARTED = 'AUTOCOMPLETE_STARTED';
+export const AUTOCOMPLETE_CANCELLED = 'AUTOCOMPLETE_CANCELLED';
+
+const initialState = {
+  loading: false,
+  suggestions: [],
```
kumar303: As you mentioned, since autocomplete is a secondary feature, it's probably fine to default it to an empty array. Just keep this pattern in mind for future reducers: this approach will never allow the app to tell if an initial API request has been made or not (which might pose a problem).

tests/unit/amo/helpers.js
```diff
@@ -137,3 +137,12 @@ export function createAddonsApiResult(results) {
   // For example: core.api.featured(), core.api.search()...
   return normalize({ results }, { results: [addonSchema] });
 }
+
+export function createFakeAutocompleteResult({ name, lang = 'en-US' }) {
```
kumar303: `lang` can be removed until you need it

kumar303: maybe `name` should default to something like ...

```js
createFakeAutocompleteResult({ name = 'suggestion-result' } = {}) {
}
```

?

tests/unit/core/sagas/test_autocomplete.js
```diff
@@ -0,0 +1,83 @@
...
+
+  function _autocompleteStart(params) {
+    sagaTester.dispatch(autocompleteStart({
+      errorHandlerId: 'create-stub-error-handler-id',
+      ...params,
+    }));
+  }
+
+  it('calls the API for suggestions', async () => {
+    const results = [createFakeAutocompleteResult({ name: 'first-result' })];
```
kumar303: As mentioned up above, it would be simpler to call this as `createFakeAutocompleteResult()` since the name is not relevant to the test.

tests/unit/core/sagas/test_autocomplete.js
```diff
@@ -0,0 +1,83 @@
...
+
+    mockApi
+      .expects('autocomplete')
+      .once()
+      .returns(Promise.resolve({ results }));
+
+    _autocompleteStart({ filters });
+
+    // The saga should respond by dispatching the autocomplete loaded action.
+    await sagaTester.waitFor(AUTOCOMPLETE_LOADED);
```
kumar303: I know you copied this from the last test (d'oh) but I think using action creators would make it a bit more robust:

```js
const expectedLoadAction = autocompleteLoad(results);

await sagaTester.waitFor(expectedLoadAction.type);
mockApi.verify();

const loadAction = sagaTester.getCalledActions()[2];
```

src/core/reducers/autocomplete.js
willdurand: Roger that 😃

Issue #2949, opened 2017-08-15
Categories never load (via server rendering)


kumar303:
### Describe the problem and steps to reproduce it:

- Go to the all-categories page, such as https://addons-dev.allizom.org/en-US/firefox/extensions/categories/
- Click on a category, like Bookmarks
- Bookmark-related add-ons load as expected
- Click the browser Refresh button to reload it

### What happened?

Bookmark-related add-ons never load. You just see the placeholders.

### What did you expect to happen?

Bookmark-related add-ons should load.

### Anything else we should know?

This is probably fallout from https://github.com/mozilla/addons-frontend/issues/2737 but I think the component already had loading issues that were just previously hidden.

kumar303:
This one was tricky to track down but I have a patch ready (it just needs test fixups and cleanup). The issue is that when `Category` first renders on the server, [this logic](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/Category/index.js#L45-L47) returns `NotFound` instead of rendering `Search`. Because of that, `Search` does not get a chance to dispatch its data loading actions. Confusingly, the second render (after sagas have ended) will cause `Search` to trigger its loading actions. By that time it's too late because no sagas are running to process them.

kumar303:
Everything seems to be working on dev. I'll cherry-pick to stage.

kumar303:
btw, I now know a ton about how sagas work; their `SagaMonitor` feature is pretty handy: https://redux-saga.js.org/docs/api/index.html#sagamonitor

kumar303:
I'm a bit worried this bug will happen again so I filed https://github.com/redux-saga/redux-saga/issues/1132

kumar303:
I just tested on stage. Everything looks fine.

SVSanta:
Verified as fixed on AMO-stg FF55(Win 10, Android 7.2.1).
Tested on all categories for Extensions, Themes for Firefox, Android, Thunderbird and SeaMonkey.
Postfix gif:
![categories](https://user-images.githubusercontent.com/28138132/29405885-f83b6166-8347-11e7-8f51-a793f622b001.gif)

PR #2960, opened 2017-08-16
Fixes category loading problem (server side)

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2949

Notes:
* `Category` loads data only when needed now
* `Search` is rendered while loading which is necessary so it can dispatch its own loading actions
* Category state is initialized properly but that required some re-design of a few things
* I spent extra time making sure the test coverage was right since this fixed a serious bug
* I was relying server logging to track this down so I cleaned up a bunch of PropType typos
* Since we need to cherry-pick this ASAP, I can address minor change requests in a follow-up

### DIFF
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -49,7 +49,6 @@ describe('Category', () => {
     }
     const state = store.getState();
 
-    // TODO: move to shallowUntilTarget
     return {
       ...mapStateToProps(state),
       dispatch: store.dispatch,
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -18,18 +18,9 @@ import { ErrorHandler } from 'core/errorHandler';
 import I18nProvider from 'core/i18n/Provider';
 import { visibleAddonType } from 'core/utils';
 import { createStubErrorHandler, getFakeI18nInst } from 'tests/unit/helpers';
-import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
+import { dispatchClientMetadata, fakeCategory } from 'tests/unit/amo/helpers';
 import ErrorList from 'ui/components/ErrorList';
 
-
-const fakeCategory = {
-  id: 5,
-  application: CLIENT_APP_FIREFOX,
-  description: 'I am a cool category for doing things',
-  name: 'Testing category',
-  slug: 'test',
-  type: ADDON_TYPE_THEME,
-};
 
 describe('Category', () => {
   let category;
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -50,9 +50,9 @@ describe('Category', () => {
     }
     const state = store.getState();
 
+    // TODO: move to shallowUntilTarget
     return {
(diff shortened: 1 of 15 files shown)


tests/unit/amo/components/TestCategory.js
```diff
@@ -75,163 +81,229 @@ describe('Category', () => {
...
+        params: {
+          slug: targetCategory.slug,
+          visibleAddonType: visibleAddonType(decoyCategory.type),
+        },
+      });
+
+      expect(root.find(NotFound)).toHaveLength(1);
+    });
+
+    it('renders 404 for bogus addonType', () => {
```
tofumatt: Bogus seemed like a weird word to use here but the literal definition is "not genuine or true", so I guess it's super-accurate! 😅

tests/unit/amo/helpers.js
```diff
@@ -62,6 +63,17 @@ export const fakeReview = Object.freeze({
   title: 'Review Title',
 });
 
+export const fakeCategory = Object.freeze({
+  application: CLIENT_APP_FIREFOX,
+  description: 'I am a cool category for doing things',
```
tofumatt: 😎

tests/unit/amo/components/TestCategory.js
```diff
@@ -75,163 +81,229 @@ describe('Category', () => {
...
+      errorHandlerId: errorHandler.id,
+    }));
   });
 
-  it('should render an error without a category too', () => {
-    const errorHandler = createStubErrorHandler(
-      new Error('example of an error')
-    );
-    const root = render({ errorHandler, category: null, loading: false });
+  it('does not fetch categories when already loaded', () => {
```
kumar303: It would be fine because `!categories` will not be true for `[]` (an empty API response of categories). That would be a good test case to add though.

tests/unit/amo/components/TestCategory.js
```diff
@@ -75,163 +81,229 @@ describe('Category', () => {
...
+    _categoriesFetch();
+    const root = render({}, { autoDispatchCategories: false });
 
-  function _searchStart(props = {}) {
-    store.dispatch(searchStart({
-      errorHandlerId: 'Search',
-      ...props,
-    }));
-  }
+    expect(root.instance().props.loading).toEqual(true);
```
kumar303: It doesn't support it since `instance()` is not a `ShallowWrapper`

tests/unit/amo/components/TestCategory.js
```diff
@@ -75,163 +81,229 @@ describe('Category', () => {
...
-    _searchStart({ filters });
-    const props = mapStateToProps(store.getState(), ownProps);
+    function _render(
+      props = {}, { autoDispatchCategories = true, ...options } = {}
+    ) {
+      if (autoDispatchCategories) {
+        _categoriesLoad({ result: [decoyCategory, targetCategory] });
+      }
+      return render(props, {
+        // Since we loaded our own, tell the parent helper not to.
```
kumar303: It is a complete sentence. It's in an imperative mood ("tell..." is a command).

tofumatt: "not to" seems an odd way to end a sentence. "Not to what?" was my first response reading the sentence.

PR #2969, opened 2017-08-19
feat: Pass appversion param to searches

### LINKED ISSUE #2912: Pass `appversion` when searching, to only see add-ons compatible with the client's version

### LINKED ISSUE #2967: Search tests are duplicated
I was working on #2912 and noticed that during a search refactor we seem to have included two duplicated search API spec names and tests. As part of the PR for #2912 I'll be removing them, but filing here for posterity.

tofumatt:
Fixes #2912.
Fixes #2967.

This passes the appversion param to any search if the user agent is Firefox for Desktop or Firefox for Android. In #2912 @diox said to send it only for Firefox 57+ but from what I can see there's no harm in sending it for any version as searches are still returned for versions as low as `~10` (I forget the exact number but it was something like that). I'm guessing stage just had no add-ons for versions below that.

@diox–is there any danger to always sending `appversion` even if the version is below 57?


src/core/api/index.js
```diff
@@ -162,56 +162,6 @@ export function callApi({
     .then((response) => (schema ? normalize(response, schema) : response));
 }
 
-type SearchParams = {|
```
tofumatt: I moved this to its own file because I was finding this file got really crowded and I wanted to search files for "search" to edit this API but had to find `core/api/index.js` instead. It's isolated in 1f4600bd744dd8e3c0f5a4563a516aae70e8434e so if you just want to review the feature change it's in the second commit: fb0f11167f3561763e330a2da53bbb0afe9480c7

tests/unit/core/api/test_index.js
```diff
@@ -209,141 +190,6 @@ describe('api', () => {
...
-        filters: { page: parsePage(3), query: 'foo' },
-      })
-        .then(unexpectedSuccess, (err) => {
-          expect(err.response.status).toEqual(401);
-          expect(err.response.apiURL).toEqual(url);
-        });
-    });
-  });
-
-  describe('search api', () => {
```
tofumatt: These search tests were duped, probably from an older refactor. I consolidated things and got rid of a few duplicated tests 👍

diox:
> @diox–is there any danger to always sending appversion even if the version is below 57?

It's not a danger per se - it's supported, but you should not do it, at least not by default. That was a request from product to help users from older Firefoxes to still find extensions once the majority has switched to webextensions and recent APIs. See the discussion in https://github.com/mozilla/addons-server/issues/5482

tofumatt:
Ah, so passing `appversion=50.0` would still only return WebExtensions?

There is (going to be) a "Show Legacy Extensions" checkbox in the search filters (it's off by default), but I suppose it could always be enabled for 57 < users.

diox:
Passing `appversion=50.0` would return any extension marked as compatible with Firefox 50.0 or greater. Whether or not they are webextensions is irrelevant for this parameter - we just expect most extensions to have switched to webextensions at some point, using recent APIs, making them incompatible with older versions.

I haven't followed the plans for the new frontend, but I'm not sure I see the value in a specific checkbox to show legacy extensions. If you pass `appversion` for FF 57 or higher, and nothing for the rest, then FF 57 users will automatically see webextensions, and the rest will see everything.

tofumatt:
Well, for instance uBlock Origin right now won’t appear as its main version is not 57+ compatible but the development version is, so showing legacy and then going to dev channel could be useful. Or just browsing old extensions? The first sounds like a bad flow though and the second seems not actually useful, so fair point. Maybe we don’t want to show old extensions at all, at least not without a hidden param not exposed in UI (useful for debugging/expert users but not something we should surface).

But it sounds like passing appversion=52 isn’t harmful if the browser used is ESR, right? Current ESR is 52 (IIRC) so adding that param on the frontend so add-ons that aren’t compatible with 52 is good, right?

I guess I’m not understanding why we would only want to pass appversion for browsers above 57. I plan to pass the current userAgent’s version so it’s not like I’m hard-coding appversion=57, if that wasn’t clear.

diox:
> Well, for instance uBlock Origin right now won’t appear as its main version is not 57+ compatible but the development version is, so showing legacy and then going to dev channel could be useful. Or just browsing old extensions? The first sounds like a bad flow though and the second seems not actually useful, so fair point. Maybe we don’t want to show old extensions at all, at least not without a hidden param not exposed in UI (useful for debugging/expert users but not something we should surface).

We don't really want most end-users to install development versions of add-ons (they might not be maintained as well as the regular version, or just generally be unstable - they're for testing purposes only). Plus, by the time Firefox 57 is released, most popular add-ons should have a webext as their main version.

> But it sounds like passing appversion=52 isn’t harmful if the browser used is ESR, right? Current ESR is 52 (IIRC) so adding that param on the frontend so add-ons that aren’t compatible with 52 is good, right? I guess I’m not understanding why we would only want to pass appversion for browsers above 57. I plan to pass the current userAgent’s version so it’s not like I’m hard-coding appversion=57, if that wasn’t clear.

By November, we expect many add-ons that were previously compatible with ESR to have their last public version be a WebExtension using APIs that didn't exist in ESR. If you were to send `appversion`, ESR users would not see those add-ons.

The compromise that was decided in the issue I linked above is to not filter those add-ons out - instead, we don't do any version filtering for Firefox < 57, and the ESR users can still see the incompatible add-ons, even if they might not be able to install them (the compatibility rules still apply to the install button as normal). The incompatibility message was tweaked to point to the page with all versions of the add-ons, where those users can find and install the latest legacy version that is still compatible with their Firefox.

tofumatt:
Ahhhhh okay, sorry I read through the issue but found it hard to make sense of–that clears it up a lot.

Makes sense, I'll send the param only when the UA is 57+. Thanks for the explanation, @diox!

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,301 @@
...
+import { createApiResponse } from 'tests/unit/core/api/helpers';
+import { dispatchSignInActions } from 'tests/unit/amo/helpers';
+import { unexpectedSuccess } from 'tests/unit/helpers';
+
+
+describe(__filename, () => {
+  let api;
+  let mockWindow;
+  const apiHost = config.get('apiHost');
+  const esrFirefoxUA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.2.1) Gecko/20100101 Firefox/52.2.1';
```
tofumatt: I can add these to the existing userAgent helpers actually, let me know and I can do that in a little fix-up commit.

tests/unit/core/api/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+  headerData = { 'Content-Type': 'application/json' }
+) {
+  const response = new Response();
+  Object.keys(headerData).forEach((key) => (
+    response.headers.append(key, headerData[key])
+  ));
+  return response.headers;
+}
+
+export function createApiResponse({
```
kumar303: I think this could move to `tests/unit/core/helpers` because there are other places that could benefit from a more realistic response. I'm thinking of saga tests and anywhere that we resolve API mocks with a response.

src/core/api/search.js
```diff
@@ -0,0 +1,86 @@
+import { oneLine } from 'common-tags';
```
tofumatt: Oops, thanks!

src/core/api/search.js
```diff
@@ -0,0 +1,86 @@
...
+
+  // If the browser is Firefox or Firefox for Android, send the appversion
+  // param to get only add-ons compatible with this version.
+  // Otherwise this flag is ignored.
+  if (
+    api.userAgentInfo.browser.name === 'Firefox' &&
+    api.userAgentInfo.os.name !== 'iOS' &&
+    (!_filters.addonType || _filters.addonType === ADDON_TYPE_EXTENSION)
+  ) {
+    if (_filters.showLegacyExtensions === true) {
```
tofumatt: Fair point. This patch and some other stuff surround versions/legacy definitely had a simplistic view of 57/not-57. As you've mentioned that isn't helpful; I just have to clear it out this patch as well 👍

src/core/api/search.js
```diff
@@ -0,0 +1,86 @@
...
+      log.info(oneLine`filters.showLegacyExtensions is true; removing
+        filters.compatibleWithVersion because we are searching for allows
+        extensions`);
+      delete _filters.compatibleWithVersion;
+    } else if (parseInt(api.userAgentInfo.browser.version, 10) >= 57) {
+      log.debug(oneLine`Setting appVersion to current application version so
+        only relevant add-ons are surfaced.`);
```
tofumatt: Right on, will do.

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,301 @@
...
+  const newAndroidUA = 'Mozilla/5.0 (Android 4.4; Tablet; rv:57.0) Gecko/57.0 Firefox/57.0';
+  const newFirefoxForIOSUA = 'Mozilla/5.0 (iPad; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4';
+  const newFirefoxUA = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:57.0) Gecko/20100101 Firefox/57.1';
```
tofumatt: Yeah good point. I'll actually just move them to the `userAgent` helpers. I'd like to improve that helper so it has useful keys (eg `userAgent.firefox57`) or something but that's for the future 😉 

I'll just move them to the existing helper so everything's at least in one place.

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,301 @@
...
+    mockWindow = sinon.mock(window);
+  });
+
+  function mockResponse(responseProps = {}) {
+    return createApiResponse({
+      jsonData: {
+        results: [
+          { slug: 'foo' },
+          { slug: 'food' },
+          { slug: 'football' },
```
tofumatt: Good call, totally.

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,301 @@
...
+
+    return search({
+      api,
+      auth: true,
+      filters: { page: parsePage(3), query: 'foo' },
+    })
+      .then(() => mockWindow.verify());
+  });
+
+  it('sets appversion if addonType is extension', () => {
```
tofumatt: Yeah, the spec is out of date from when my assumptions were more simplistic

tofumatt:
Okay this is ready for another review!

src/core/reducers/api.js
```diff
@@ -20,16 +20,21 @@ import type {
 import type { Exact } from 'core/types/util';
 
 type UserAgentInfoType = {|
-  browser: string,
-  os: string,
+  browser: {
+    name?: string,
+    version?: string,
```
kumar303: Thanks, this is really helpful. Huh, did `UAParser` change? Or maybe this was always typed wrong. Could you add in the other property? I see it like this in the console:

```js
browser: {
  name?: string,
  version?: string,
  major?: string,
}
```

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,252 @@
...
+      .withArgs(`${apiHost}/api/v3/addons/search/?app=firefox&page=3&q=foo&lang=en-US`)
+      .once()
+      .returns(mockResponse());
+    return _search().then(() => mockWindow.verify());
+  });
+
+  it('sets appversion if Firefox version is 57 or above', () => {
+    api = dispatchSignInActions({
+      clientApp: CLIENT_APP_FIREFOX,
+      userAgent: userAgents.firefox[5],
```
kumar303: It's pretty important to know that this user agent is for 57+. What about making an alias to it at the top of the file like:

```js
const firefox57 = userAgents.firefox[5];
```

tests/unit/core/api/test_search.js
```diff
@@ -0,0 +1,252 @@
...
+      .returns(mockResponse());
+
+    return _search().then(() => mockWindow.verify());
+  });
+
+  // See: https://github.com/mozilla/addons-frontend/pull/2969#issuecomment-323551742
+  it('does not set appversion if version is below 57', () => {
+    api = dispatchSignInActions({
+      clientApp: CLIENT_APP_FIREFOX,
+      userAgent: userAgents.firefox[4],
```
kumar303: Same thing here. It would be easier to follow this test if there was an alias like `userAgent: firefoxESR52`

src/core/reducers/api.js
tofumatt: I think it's always been wrong? Well, it's possible this patch (which checks on properties inside `userAgentInfo` exposed it being wrong, yeah.

I'll add in `major` as well, flow didn't catch it but you're right it should be there 👍

tests/unit/core/api/test_search.js
tofumatt: Fair enough; like I said I'd like to label the versions with keys instead of numbers, but that can be in a later patch.

PR #2971, opened 2017-08-21
Search hints/autocomplete UI

### LINKED ISSUE #2747: Add search hints/autocomplete

willdurand:
Fix #2747 
~~⚠️  depends on #2958~~

---

This PR adds the UI components for the autocomplete search feature. I tried to use the most recent style as much as possible, but did not change the search UI. I have added a new mixin `arrow-big-blue()`, similar to the other arrow mixins. The name comes from the icon in Sketch.


tofumatt:
If it's okay with you I'm just gonna wait until #2958 is merged to review this so it's a bit easier to see what's in this patch? 😅

willdurand:
> If it's okay with you I'm just gonna wait until #2958 is merged to review this so it's a bit easier to see what's in this patch? 😅

Yes! In any cases, the last commit is the one to review https://github.com/mozilla/addons-frontend/pull/2971/commits/ddd3f791522e78adde69ec60f1cafcaaa828d802. I kept both branches in sync.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  handleSuggestionsClearRequested = () => {
+    this.props.dispatch(autocompleteCancel());
+  }
+
+  handleSuggestionSelected = (e, { suggestion }) => {
+    e.preventDefault();
+    this.setState({ searchValue: '' }, () => {
+      this.context.router.push(suggestion.url);
```
kumar303: Huh. I'm not sure how this works. In my devtools panel, `suggestion.url` seems to start with `https://addons-dev.allizom.org`, not my local server, yet it loads the page correctly on my local server.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,215 @@
...
+      <form
+        method="GET"
+        action={`/${api.lang}/${api.clientApp}${pathname}`}
+        onSubmit={this.handleSearch}
+        className="SearchForm-form"
+        ref={(ref) => { this.form = ref; }}
+      >
+        <label
+          className="visually-hidden"
+          htmlFor="q"
```
kumar303: could this use the `input` ID instead maybe?

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,215 @@
...
+            suggestionsList: 'SearchForm-suggestions-list',
+            suggestion: 'SearchForm-suggestions-item',
+            suggestionHighlighted: 'SearchForm-suggestions-item--highlighted',
+          }}
+        />
+        <button
+          className="visually-hidden"
+          onClick={this.handleSearch}
+          ref={(ref) => { this.submitButton = ref; }}
+          title="Enter"
```
kumar303: Even though it's hidden, this should be localized like `title={this.props.i18n.gettext('Enter')}` for accessibility. When you move old code around you get to own it 🌵

src/core/css/inc/mixins.scss
```diff
@@ -162,6 +162,27 @@ $default-arrow-margin: 3px;
   }
 }
 
+@mixin arrow-big-blue($margin: $default-arrow-margin) {
```
kumar303: This doesn't seem like something that would be used very often. Can it live in `SearchForm/styles.scss` for now?

src/amo/components/SearchForm/styles.scss
```diff
@@ -0,0 +1,76 @@
...
+
+.SearchForm-suggestions-item:hover,
+.SearchForm-suggestions-item:active,
+.SearchForm-suggestions-item:focus,
+.SearchForm-suggestions-item--highlighted {
+  color: $link-color;
+  white-space: normal;
+
+  .Suggestion::after {
+    @include arrow-big-blue();
```
kumar303: What if the component loaded this with `ui/components/Icon` instead?

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,215 @@
...
+  handleSuggestionSelected = (e, { suggestion }) => {
+    e.preventDefault();
+    this.setState({ searchValue: '' }, () => {
+      this.context.router.push(suggestion.url);
+    });
+  }
+
+  renderSuggestion = (suggestion) => {
+    if (this.props.loadingSuggestions) {
+      return <LoadingText width={60} />;
```
willdurand: Yes, we agree on doing this with @tofumatt. Other solutions were:

- to leave the suggestions list open all the time
- to see this list blinking on every API call

Having the loading bars is nicer than the two previous options.

src/amo/components/SearchForm/index.js
willdurand: > When you move old code around you get to own it 🌵

🙏

willdurand:
I have updated the code and added a few more test cases.

src/amo/components/SearchForm/index.js
tofumatt: That seems like something we should at least have a test for (is there one?)... if that behaviour changed it would cause some weird bugs.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,236 @@
...
+
+    if (this.props.query !== query) {
+      this.setState({ searchValue: query || '' });
+    }
+  }
+
+  getSuggestions() {
+    if (this.props.loadingSuggestions) {
+      // 10 is the maximum number of results returned by the API
+      return Array(10).fill({ component: <LoadingText width={60} /> });
```
kumar303: I guess this is a little weird (my suggestion, I know) but you should probably attempt to protect against colliding with real `suggestion` object properties. Perhaps calling it `{ loadingComponent: <LoadingText width={60} /> }` is enough.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,236 @@
...
+
+  handleSuggestionSelected = (e, { suggestion }) => {
+    e.preventDefault();
+    this.setState({ searchValue: '' }, () => {
+      this.props.router.push(suggestion.url);
+    });
+  }
+
+  renderSuggestion = (suggestion) => {
+    if (this.props.loadingSuggestions) {
```
kumar303: this is a duplication of logic which is unnecessary and also introduces a maintenance burden. I suggest reacting to the data instead, like this:

```js
if (suggestion.component) {
  // The suggestion is already a rendered component.
  return suggestion.component;
}
```

src/amo/components/SearchForm/styles.scss
```diff
@@ -0,0 +1,80 @@
...
+  outline: none;
+  padding: $item-padding;
+  width: 100%;
+
+  .Icon-arrow-big-blue {
+    display: none;
+    float: right;
+
+    [dir=rtl] & {
+      float: left;
```
kumar303: I believe we're trying to use flexbox instead of floats everywhere because it's more predictable across browsers but no biggie if it's harder to implement with flex.

src/ui/components/Icon/styles.scss
```diff
@@ -115,3 +115,14 @@
...
 }
+
+.Icon-arrow-big-blue {
+  @include icon($name: "arrow-big-blue");
+
+  width: 18px;
+
+  [dir=rtl] & {
+    right: auto;
+    transform: rotate(180deg);
```
kumar303: Nice!

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -40,140 +39,226 @@ describe('<SearchForm />', () => {
...
+      const state = store.getState();
+
+      expect(mapStateToProps(state).suggestions).toEqual([
+        {
+          iconUrl: result.icon_url,
+          name: result.name,
+          url: result.url,
+        },
+      ]);
+      expect(mapStateToProps(state).loadingSuggestions).toBe(false);
```
kumar303: It would be nice to have another test that asserts `loadingSuggestions` gets set to true. You could set it up with a dispatch like:

```js
store.dispatch(autocompleteStart(...));
```

tests/unit/amo/components/TestSearchFormSuggestion.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import Suggestion from 'amo/components/SearchForm/Suggestion';
+import Icon from 'ui/components/Icon';
+import { fakeAddon } from 'tests/unit/amo/helpers';
+
+
+describe(__filename, () => {
+  const shallowComponent = (props) => {
```
kumar303: typically we just call this function `render()` but I suppose it doesn't matter

tests/unit/amo/components/TestSearchFormSuggestion.js
```diff
@@ -0,0 +1,35 @@
...
+      iconUrl: fakeAddon.icon_url,
+      ...props,
+    };
+
+    return shallow(<Suggestion {...allProps} />);
+  };
+
+  it('renders itself', () => {
+    const props = {};
+    const root = shallowComponent(props);
```
kumar303: You could call this as `shallowComponent()` if you change the signature to:

```js
const shallowComponent = (props = {}) => {
  // ...
}
```

src/amo/components/SearchForm/styles.scss
```diff
@@ -0,0 +1,80 @@
...
+  color: $link-color;
+  white-space: normal;
+
+  .Icon-arrow-big-blue {
+    display: inline-block;
+  }
+}
+
+.Suggestion {
+  height: 20px;
```
kumar303: When suggestions arrive and they replace the `LoadingText` placeholders, I see everything expand by about 3 or 4 pixels. Did you look into why that's happening? It might just need some `.Suggestion .LoadingText {...}` styles.

src/amo/components/SearchForm/index.js
kumar303: I was playing with this and I think these settings make the placeholders look more like the real results. This makes the transition less jarring to me as well.

```diff
diff --git a/src/amo/components/SearchForm/index.js b/src/amo/components/SearchForm/index.js
index 9253d326..20cd578f 100644
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -74,7 +74,9 @@ export class SearchFormBase extends React.Component {
   getSuggestions() {
     if (this.props.loadingSuggestions) {
       // 10 is the maximum number of results returned by the API
-      return Array(10).fill({ component: <LoadingText width={60} /> });
+      return Array(10).fill({
+        component: <LoadingText minWidth={20} range={12} />
+      });
     }
 
     return this.props.suggestions;
```

kumar303: Also, if you include a generic add-on icon when loading placeholders then I think it would make it more obvious that autocomplete results are about to load.

kumar303: I was thinking more about this and I know what the problem is. The responsibility of loading logic should be in `Suggestion` like this:

```js
Array(10).fill(<Suggestion loading={true} />);
```

kumar303: Also, adding a generic icon ([`getAddonIconUrl()`](https://github.com/mozilla/addons-frontend/blob/master/src/core/imageUtils.js#L4)) will make the loading results look more realistic and it should also make it more obvious that autocompletion is about to happen.

src/core/reducers/autocomplete.js
```diff
@@ -53,10 +53,14 @@ export default function reducer(state = initialState, action = {}) {
...
-        // TODO: Remove this when `null` names are not returned. See:
-        // https://github.com/mozilla/addons-server/issues/6189
+          // TODO: Remove this when `null` names are not returned. See:
+          // https://github.com/mozilla/addons-server/issues/6189
           .filter((result) => result.name !== null)
-          .map((result) => result.name),
+          .map((result) => ({
+            name: result.name,
+            url: result.url,
+            iconUrl: result.icon_url,
```
kumar303: This needs to use [`getAddonIconUrl(result)`](https://github.com/mozilla/addons-frontend/blob/master/src/core/imageUtils.js#L4).

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,236 @@
...
+    e.preventDefault();
+    this.searchInput.blur();
+    this.goToSearch(this.searchInput.value);
+  }
+
+  handleSearchChange = (e) => {
+    this.setState({ searchValue: e.target.value });
+  }
+
+  handleSuggestionsFetchRequested({ value }) {
```
willdurand: @kumar303 I have never used [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) before, especially not for bound methods. To me, it is expected to find bound methods in the constructor as described [in the React doc](https://facebook.github.io/react/docs/handling-events.html). Honestly, I get more confused by your approach than mine, but I understand your concern and I can update my code.

src/amo/components/SearchForm/index.js
willdurand: That makes sense 👍

src/amo/components/SearchForm/index.js
kumar303: I find that snippet in the react docs (`this.handleClick = this.handleClick.bind(this)`) to be a horrendous hack :/

The main problem I have with that approach is just that there's no way to look at the method and *know* that it has been replaced. You have to also think to look in the constructor which should not be a requirement for figuring out how things work. 

If you were to do it in the constructor using a *new* property then I'd be fine with that since one would know to search for the property by name. In other words, I just don't want to see a method by the same name get replaced.

willdurand: I am fine with that, let's make things (more) explicit.

src/amo/components/SearchForm/index.js
willdurand: I thought this `Suggestion` would be a very small/dumb component, but it makes a lot of sense to add it the loading logic, and I like this `getAddonIconUrl()`, which I did not know about 😖 . This should solve a lot of issues you mentioned.

tests/unit/amo/components/TestSearchFormSuggestion.js
willdurand: There is also a `render()` function in Enzyme but it does different things than `shallow()`.

src/amo/components/SearchForm/index.js
willdurand: I have a slightly different idea based on your comments. I am not a huge fan of having totally different arguments passed to methods.

src/amo/components/SearchForm/styles.scss
willdurand: It is even more obvious with default icons added. It seems to be a missing `vertical-align`.

src/amo/components/SearchForm/index.js
willdurand: I found a different way to debounce bound methods, I did not know about it before:

``` js
class C extends React.Component {
  method = debounce(({ args }) => {
    // method body
  }, 200)
}
```

willdurand:
Ok, I have added more tests and reworked this loading thing.

The `SearchForm` is fairly well tested now (🤞 ). I have slightly updated its code to ease the addition of my tests. I don't think we need a component wrapper, or at least the existing tests were not showing the need for it.

willdurand:
Gifs updated!

tests/unit/helpers.js
```diff
@@ -241,6 +241,12 @@ export function createFakeEvent(extraProps = {}) {
   };
 }
 
+export function createFakeChangeEvent(value = '') {
```
kumar303: Typing `createFakeEvent({ target: { value: '' } })` seems fine to me. Why create a new helper?

kumar303: Another option is to move `createFakeChangeEvent` directly into `TestSearchForm.js` since that's the only place it's used.

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,241 @@
...
+
+    if (this.props.query !== query) {
+      this.setState({ searchValue: query || '' });
+    }
+  }
+
+  getSuggestions() {
+    if (this.props.loadingSuggestions) {
+      // 10 is the maximum number of results returned by the API
+      return Array(10).fill({
```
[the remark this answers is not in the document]
willdurand: It is not possible since Autosuggest needs data, not components. This is the job of `renderSuggestion()` to decide how to display the suggestions.

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,333 @@
...
-    input = root.searchQuery.input;
+    it('updates the location on form submit', () => {
+      sinon.assert.notCalled(router.push);
+      input.value = 'adblock';
+      Simulate.click(root.submitButton);
+      sinon.assert.called(router.push);
+    });
+
+    it('passes addonType when set', () => {
+      root = renderIntoDocument(getComponentUnderTest({
```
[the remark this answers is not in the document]
willdurand: I did not touch this test case. I was not sure about rewriting the whole file, but yeah I can definitely do that

src/amo/components/SearchForm/index.js
```diff
@@ -0,0 +1,241 @@
...
+        name: 'loading-text',
+        iconUrl: getAddonIconUrl(),
+        loading: true,
+      });
+    }
+
+    return this.props.suggestions.map((suggestion) => ({
+      ...suggestion,
+      loading: false,
```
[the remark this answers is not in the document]
willdurand: I wanted to keep a consistent 
 shape for the Autosuggest `suggestions`.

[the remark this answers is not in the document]
willdurand: Also, it does not make really sense to have `loading` in the reducer, because it is a pure UI prop. That is why I think it makes sense to augment the `suggestions` in the methods used by the `Autosuggest` component.

willdurand:
Added more tests, and verified with local coverage. I also converted the whole `SearchForm` test file to use Enzyme.

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
 
-  it('updates the location on form submit', () => {
-    sinon.assert.notCalled(router.push);
-    input.value = 'adblock';
-    Simulate.click(root.submitButton);
-    sinon.assert.called(router.push);
-  });
+    it('blurs the form on submit', () => {
+      const blurSpy = sinon.stub(wrapper.instance().searchInput, 'blur');
+      expect(blurSpy.called).not.toBeTruthy();
```
kumar303: use `sinon.assert.notCalled()` here

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      expect(wrapper.state('searchValue')).toEqual('');
+    });
+
+    it('fetches suggestions on focus', () => {
+      const dispatch = sinon.spy();
+      wrapper = mountComponent({
+        query: 'foo',
+        dispatch,
+      });
+      // no call to handleSuggestionsFetchRequested() until the input has focus,
```
kumar303: Try to use complete sentences in comments to make them easier to read. I think this would be "Expect no call to ..." ?

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      // even if there is already a `searchValue`
+      expect(dispatch.calledOnce).toBe(false);
```
kumar303: use `sinon.assert.callCount()` or `sinon.assert.noCalled()`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      // this is needed to trigger handleSuggestionsFetchRequested()
+      wrapper.find('input').simulate('focus');
+      expect(dispatch.calledOnce).toBe(true);
```
kumar303: `sinon.assert.callCount()`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      expect(dispatch.calledWith(autocompleteStart({
```
kumar303: use `sinon.assert.calledWith()`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+    it('clears suggestions when input is cleared', () => {
+      const dispatch = sinon.spy();
+      wrapper = mountComponent({
+        query: 'foo',
+        dispatch,
+      });
+      // clearing the input calls handleSuggestionsClearRequested()
+      wrapper.find('input').simulate('change', createFakeChangeEvent());
+      expect(dispatch.calledOnce).toBe(true);
+      expect(dispatch.calledWith(autocompleteCancel())).toBe(true);
```
kumar303: both of these need updating

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      wrapper = mountComponent({ suggestions: [] });
+      wrapper.find('input').simulate('focus');
+      expect(wrapper.find(Suggestion)).toHaveLength(0);
+      expect(wrapper.find(LoadingText)).toHaveLength(0);
+    });
+
+    it('displays 10 loading bars when suggestions are loading', () => {
+      wrapper = mountComponent({
+        suggestions: [],
+        loadingSuggestions: true,
```
kumar303: Instead of setting props explicitly, this should dispatch the `autocompleteStart` action and do something like:

```js
wrapper = mountComponent(mapStateToProps(store.getState()));
```

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+        query: 'foo',
+        suggestions: autocompleteState.suggestions,
+      });
+      expect(wrapper.state('searchValue')).toBe('foo');
+
+      wrapper.find('input').simulate('focus');
+      wrapper.find(Suggestion).simulate('click');
+      expect(wrapper.state('searchValue')).toBe('');
+      expect(router.push.calledOnce).toBe(true);
+      expect(router.push.calledWith(result.url)).toBe(true);
```
kumar303: do a search in this file for `calledOnce` and `calledWith` so you can catch all of these

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
+      expect(wrapper.state('searchValue')).toEqual('baz');
+
+      wrapper.instance().handleSuggestionSelected(createFakeEvent(), {
+        suggestion: {
+          ...autocompleteState.suggestions[0],
+          loading: true,
+        },
+      });
+      expect(wrapper.state('searchValue')).toEqual('baz');
+      expect(router.push.called).toBe(false);
```
kumar303: `sinon.assert.called()`

tests/unit/ui/components/TestSearchInput.js
```diff
@@ -112,7 +112,42 @@ describe('<SearchInput />', () => {
...
-    const root = renderIntoDocument(<SearchInput name="foo" defaultValue="yo" />);
+    const root = renderIntoDocument(<SearchInput name="foo" value="yo" />);
     expect(root.value).toEqual('yo');
   });
+
+  it('forwards the onFocus event', () => {
+    const spy = sinon.spy();
+    const root = renderIntoDocument(<SearchInput name="foo" onFocus={spy} />);
+    Simulate.focus(root.input);
+    expect(spy.calledOnce).toBe(true);
```
kumar303: use `sinon.assert.called()`

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,386 @@
...
-
-  beforeEach(() => {
-    router = { push: sinon.spy() };
-    root = renderIntoDocument(<SearchFormWrapper />).root;
-    form = root.form;
-    input = root.searchQuery.input;
-  });
+  const mountComponent = (props = {}) => {
+    return mount(
+      <SearchFormBase
```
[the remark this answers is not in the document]
willdurand: Yeah, I am a bit confused by all the different ways of doing things to be honest. That is why I tried to keep things "as is". I really agree with what you're saying/showing here.

tests/unit/amo/components/TestSearchForm.js
willdurand: mmh actually I should have noticed that most of the `spy` rely on `sinon.assert`, not jest `expect`..

willdurand:
I have updated the test cases :)

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,394 @@
...
+
+    it('ignores loading suggestions that are selected', () => {
+      const result = createFakeAutocompleteResult();
+      const { store } = dispatchAutocompleteResults({ results: [result] });
+      const { autocomplete: autocompleteState } = store.getState();
+
+      wrapper = mountComponent({ query: 'baz' });
+      expect(wrapper.state('searchValue')).toEqual('baz');
+
+      wrapper.instance().handleSuggestionSelected(createFakeEvent(), {
```
kumar303: did `wrapper.find(Suggestion).at(0).simulate('click')` not work for this?

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,394 @@
...
+        suggestion: {
+          ...autocompleteState.suggestions[0],
+          loading: true,
+        },
+      });
+      expect(wrapper.state('searchValue')).toEqual('baz');
+      sinon.assert.notCalled(router.push);
+    });
+
+    it('does not fetches suggestions when there is not value', () => {
```
kumar303: it does not fetch suggestions when there is no value

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,394 @@
...
+        query: 'ad',
+        addonType: ADDON_TYPE_EXTENSION,
+        dispatch,
+      });
+
+      wrapper.find('input').simulate('focus');
+
+      sinon.assert.callCount(dispatch, 1);
+      sinon.assert.calledWith(dispatch, autocompleteStart({
+        errorHandlerId: createStubErrorHandler().id,
```
kumar303: heh, this only works by coincidence. I think you need to move the variable to the upper scope so you can reference it. Example:

```diff
diff --git a/tests/unit/amo/components/TestSearchForm.js b/tests/unit/amo/components/TestSearchForm.js
index a4d678ef..2915f09a 100644
--- a/tests/unit/amo/components/TestSearchForm.js
+++ b/tests/unit/amo/components/TestSearchForm.js
@@ -32,6 +32,7 @@ import {
 describe(__filename, () => {
   const pathname = '/search/';
   const api = { clientApp: 'firefox', lang: 'de' };
+  let errorHandler;
   let router;
   let wrapper;
   let form;
@@ -46,7 +47,7 @@ describe(__filename, () => {
         i18n={getFakeI18nInst()}
         loadingSuggestions={false}
         suggestions={[]}
-        errorHandler={createStubErrorHandler()}
+        errorHandler={errorHandler}
         dispatch={() => {}}
         router={router}
         debounce={(callback) => (...args) => callback(...args)}
@@ -67,6 +68,7 @@ describe(__filename, () => {
       wrapper = mountComponent();
       form = wrapper.find('form');
       input = wrapper.find('input');
+      errorHandler = createStubErrorHandler();
     });
 
     it('renders a form', () => {
```

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -1,179 +1,394 @@
...
+} from 'core/reducers/autocomplete';
 
 
-describe('<SearchForm />', () => {
+describe(__filename, () => {
   const pathname = '/search/';
   const api = { clientApp: 'firefox', lang: 'de' };
   let router;
-  let root;
+  let wrapper;
```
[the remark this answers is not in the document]
willdurand: sure, I like it better too.

tests/unit/amo/components/TestSearchForm.js
willdurand: oops 😞

tests/unit/amo/components/TestSearchForm.js
willdurand: No because it is tricky to render `Suggestion` without actual suggestions. I mean, I cannot simulate a click on "nothing", I have to force the call to `handleSuggestionSelected()`.

tests/unit/amo/components/TestSearchForm.js
willdurand: indeed, thanks!

tests/unit/amo/components/TestSearchForm.js
willdurand: I cannot make it work 😢

kumar303: ok, no worries. It seemed worth a try!

willdurand:
Fixed the tests, should be better now. Thanks for the reviews!

willdurand:
Finally, thank you and sorry for the delay!

PR #2972, opened 2017-08-21
Add badge for featured addons

### LINKED ISSUE #2752: Add badge for featured (featured add-on indicator)
If an add-on is featured, that should be exposed as a badge (as per the UX mocks).

API landed in https://github.com/mozilla/addons-server/issues/5769.

willdurand:
Fix #2752

---

This PR adds a badge to all featured addons, on the addon page. A new `Badge` component has been created as the mockups show more badges.
## Screenshots

screen shot 2017-08-21 at 16 15 09
screen shot 2017-08-21 at 16 15 12
screen shot 2017-08-21 at 16 15 16


src/ui/components/Badge/styles.scss
```diff
@@ -0,0 +1,35 @@
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+@mixin icon($name) {
+  background: url("~amo/img/icons/badge-#{$name}.svg") 50% 50% no-repeat;
```
tofumatt: I would rather keep the badges:

1. in the UI folder
2. Inside the badge component folder

We (mostly) do this with `Icon` and I think it's a bit easier to manage.

src/ui/components/Badge/styles.scss
```diff
@@ -0,0 +1,35 @@
...
+
+.Badge {
+  @include font-bold();
+  @include margin-end(10px);
+
+  background-color: $neutral-base-color;
+  border-radius: 15px;
+  display: inline-block;
+  font-size: $font-size-s;
+  line-height: 16px;
```
tofumatt: We should just be using numbers for line height, not pixels. (https://developer.mozilla.org/en/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values)

It might not matter a tonne in this particular instance but I think it's a good practice. I guess we should have a style it rule...

willdurand: Oh I must admit I did not know this. Thank you for the link!

src/amo/components/Addon/index.js
```diff
@@ -308,6 +311,15 @@ export class AddonBase extends React.Component {
             <h1 className="Addon-title" {...titleProps} />
             <p className="Addon-summary" {...summaryProps} />
 
+            <div className="Addon-badges">
+              {isFeatured ? (
+                <Badge
+                  type="featured"
+                  label={i18n.sprintf(i18n.gettext('Featured %(addonType)s'), { addonType })}
```
tofumatt: So in fact, after discussing #2979, we should just have a switch case for this translation for all the add-on types.

We tend to only feature extensions and themes though, so I think the switch/case can just be for `ADDON_TYPE_EXTENSION`, `ADDON_TYPE_THEME`, and then a default with "add-on".

kumar303: Also, just in general, note that you can't really use placeholders in localizable strings. The only times we use it are for things like HTML (like anchor URLs).

willdurand: Oh ok, I took inspiration here actually: https://github.com/mozilla/addons-frontend/blob/d759b35145a429a3e90c6d5fd691d5207bcdd6ef/src/amo/components/Addon/index.js#L228-L230.

src/ui/components/Badge/index.js
```diff
@@ -0,0 +1,17 @@
+/* @flow */
+import React from 'react';
+
+import './styles.scss';
+
+type Props = {|
+  label: string,
+  type?: string,
```
kumar303: This should be a [literal type](https://flow.org/en/docs/types/literals/) that only accepts a single string, e.g. `type?: 'featured'`. As we add more types we can adjust it. Also, sadly, I think you should also throw an error if an unexpected type is passed in. This is because we don't have great Flow coverage yet.

src/ui/components/Badge/styles.scss
```diff
@@ -0,0 +1,35 @@
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+@mixin icon($name) {
```
kumar303: This is pretty cool. We should use it in [`Icon/styles.scss`](https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/Icon/styles.scss) :) but if that's too big of a change for this patch then maybe just a file an issue to fix it up later.

willdurand:
@kumar303 I have updated the code to include an `Icon`, and also refactored the `Icon` style.

willdurand:
(added a few more test cases to increase coverage.)

src/ui/components/Icon/styles.scss
```diff
@@ -24,90 +28,90 @@
...
 .Icon-fox-light {
-  background: url("./fox-light.svg") center no-repeat;
-  background-size: contain;
+  @include icon($name: "fox-light");
 }
 
 .Icon-mozilla {
-  background: url("./mozilla.svg") center no-repeat;
-  background-size: contain;
+  @include icon($name: "mozilla");
```
tofumatt: Nice! Thanks for cleaning these up! 👍

src/amo/components/Addon/index.js
```diff
@@ -292,14 +306,17 @@ export class AddonBase extends React.Component {
...
 
     let isCompatible = false;
+    let isFeatured = false;
     let compatibility;
     if (addon) {
       compatibility = getClientCompatibility({
         addon, clientApp, userAgentInfo,
       });
       isCompatible = compatibility.compatible;
+      isFeatured = addon.is_featured;
```
kumar303: I know this file isn't using Flow but could you add `is_featured` to the [Addon type](https://github.com/willdurand/addons-frontend/blob/cb3c68935ee7b2fe966660fdb7f55ef88d2912c1/src/core/types/addons.js#L39) while you're in here?

src/ui/components/Icon/styles.scss
```diff
@@ -1,20 +1,24 @@
...
+  background: url("./#{$name}.svg") center no-repeat;
+  background-size: contain;
   content: "";
   display: inline-block;
   height: $default-icon-size;
   width: $default-icon-size;
 }
 
 .Icon-magnifying-glass {
-  background: url("~amo/img/icons/magnifying-glass.svg") center no-repeat;
```
kumar303: Sigh. I see you found some cobwebs. I don't see the file path to `magnifying-glass.svg` used directly so I think moving it was safe. Thanks.

src/ui/components/Icon/styles.scss
```diff
@@ -24,90 +28,90 @@
...
-  background: url("./multitasking-octopus.svg") center no-repeat;
-  background-size: contain;
+  @include icon($name: "multitasking-octopus");
+
   height: 199.3px;
   width: 366px;
 }
+
+.Icon-featured {
+  @include icon($name: "featured");
```
kumar303: this doesn't need a height and width?

tests/unit/amo/components/TestAddon.js
```diff
@@ -745,4 +746,32 @@ describe('mapStateToProps', () => {
 
     expect(addon).toEqual(undefined);
   });
+
+  it('displays a badge when the addon is featured', () => {
+    const addon = { ...fakeAddon, is_featured: true };
+    const rootNode = renderAsDOMNode({ addon });
+
+    expect(rootNode.querySelector('.Badge-featured').textContent).toContain('Featured Extension');
```
kumar303: You may have seen examples of testing like this elsewhere but they are all relics of the past from before we had shallow rendering via enzyme. I would like to see more of a "unit testing" approach here where you are only making assertions about `<Badge/>`'s public properties. This is a safe approach because `Badge` already has its own unit tests. This is a robust approach because it allows the implementation of `Badge` to evolve independently. Here is a way to achieve the same result:

```js
const root = shallowRender({ addon });

expect(root.find(Badge)).toHaveProp('type', 'featured');
expect(root.find(Badge)).toHaveProp('label', 'Featured Extension');
```

tests/unit/amo/components/TestAddon.js
```diff
@@ -745,4 +746,32 @@ describe('mapStateToProps', () => {
...
+  });
+
+  it('adds a different badge label when a "theme" addon is featured', () => {
+    const addon = { ...fakeAddon, is_featured: true, type: ADDON_TYPE_THEME };
+    const rootNode = renderAsDOMNode({ addon });
+
+    expect(rootNode.querySelector('.Badge-featured').textContent).toContain('Featured Theme');
```
kumar303: Try to see if you can apply the same approach to the rest of these tests

tests/unit/ui/components/TestBadge.js
```diff
@@ -0,0 +1,31 @@
...
+import Icon from 'ui/components/Icon';
+
+
+describe(__filename, () => {
+  it('renders a badge', () => {
+    const badge = shallow(<Badge label="super badge" />);
+
+    expect(badge).toHaveClassName('Badge');
+    expect(badge.find(Icon)).toHaveLength(0);
+    expect(badge.children()).toIncludeText('super badge');
```
kumar303: It would be nice to see an assertion making sure that `Icon` gets an `alt` property since that's essential for accessibility.

src/ui/components/Icon/styles.scss
tofumatt: Icons have a default height/width IIRC.

src/ui/components/Icon/styles.scss
willdurand: Yes, I checked that too, thanks

src/ui/components/Icon/styles.scss
willdurand: Yes, previously there was a default height/width configured with a variable. I kept this behavior.

willdurand:
@kumar303 thanks, I updated the code

PR #2973, opened 2017-08-21
	Add badge for needs restart

### LINKED ISSUE #2753: Add badge for needs restart
If an add-on requires a restart, we should add a badge (in UX mocks).

Uses https://github.com/mozilla/addons-server/issues/5772, which is complete.

### LINKED ISSUE #2990: Installation button is on the wrong side on device
Steps to reproduce:
1. Load details page for an add-on on your Android device using FF55 i.e. https://addons-dev.allizom.org/en-US/android/addon/cloud-viewer/?src=ss
2. Observe installation button

Expected results:
Button is in the right side.

Actual results:
Button is on the left side.

Notes/Issues:
- not reproducing on desktop pages

Verified on AMO-dev FF55(Android 7.2.1). Issue is only reproducing on -dev.
May be related with #2888 or with #2752 .
Screenshot for this issue:

willdurand:
Fix #2753 
Fix #2990 

~~⚠️  depends on #2972~~

---

Adds a new badge for addons that need restart. Note: I used the same tip as for the `installURL` field to expose `is_restart_required`.

## Screenshots

screen shot 2017-08-23 at 17 23 54

screen shot 2017-08-23 at 17 28 44

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -315,7 +315,7 @@ export class AddonBase extends React.Component {
       });
       isCompatible = compatibility.compatible;
       isFeatured = addon.is_featured;
-      isRestartRequired = addon.is_restart_required;
+      isRestartRequired = addon.isRestartRequired;
     }
 
     return (
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -307,6 +307,7 @@ export class AddonBase extends React.Component {
 
     let isCompatible = false;
     let isFeatured = false;
+    let isRestartRequired = false;
     let compatibility;
     if (addon) {
       compatibility = getClientCompatibility({
@@ -314,6 +315,7 @@ export class AddonBase extends React.Component {
       });
       isCompatible = compatibility.compatible;
       isFeatured = addon.is_featured;
+      isRestartRequired = addon.is_restart_required;
     }
 
     return (
@@ -329,6 +331,12 @@ export class AddonBase extends React.Component {
                 <Badge
                   type="featured"
                   label={this.getFeaturedText(addonType)}
+                />
+              ) : null}
+              {isRestartRequired ? (
+                <Badge
+                  type="restart-required"
+                  label={i18n.gettext('Restart Required')}
                 />
               ) : null}
             </div>
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -17,6 +17,7 @@ export function denormalizeAddon(apiAddon) {
       ...apiAddon,
       // Set iconUrl to be consistent between disco and amo.
       iconUrl: apiAddon.icon_url,
+      is_restart_required: apiAddon.is_restart_required || false,
     };
   }
   return apiAddon;
@@ -48,6 +49,7 @@ export default function addon(state = initialState, action) {
         newState[key] = {
           ...thisAddon,
           installURL: thisAddon.current_version.files[0].url,
+          is_restart_required: thisAddon.current_version.files[0].is_restart_required,
         };
       } else {
         newState[key] = thisAddon;
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -17,7 +17,6 @@ export function denormalizeAddon(apiAddon) {
       ...apiAddon,
       // Set iconUrl to be consistent between disco and amo.
       iconUrl: apiAddon.icon_url,
-      is_restart_required: apiAddon.is_restart_required || false,
     };
   }
   return apiAddon;
@@ -28,7 +27,10 @@ export default function addon(state = initialState, action) {
   if (payload && payload.entities && payload.entities.addons) {
     const newState = { ...state };
     Object.keys(payload.entities.addons).forEach((key) => {
-      const thisAddon = payload.entities.addons[key];
+      const thisAddon = {
+        ...payload.entities.addons[key],
+        isRestartRequired: false,
+      };
       if (thisAddon.theme_data) {
         newState[key] = {
           ...thisAddon,
@@ -48,8 +50,10 @@ export default function addon(state = initialState, action) {
       } else if (thisAddon.current_version && thisAddon.current_version.files.length > 0) {
         newState[key] = {
           ...thisAddon,
-          installURL: thisAddon.current_version.files[0].url,
-          is_restart_required: thisAddon.current_version.files[0].is_restart_required,
+          installURL: thisAddon.current_version.files[0].url || '',
+          isRestartRequired: thisAddon.current_version.files.some(
+            (file) => !!file.is_restart_required
+          ),
         };
       } else {
         newState[key] = thisAddon;
--- a/src/ui/components/Badge/styles.scss
+++ b/src/ui/components/Badge/styles.scss
@@ -28,7 +28,7 @@
 .Badge-restart-required {
   background-color: $yellow-50;
 
-  .Icon-restart-required {
+  .Icon-restart {
     margin-top: -2px;
   }
 }
--- a/src/ui/components/Badge/styles.scss
+++ b/src/ui/components/Badge/styles.scss
@@ -10,16 +10,25 @@
   display: inline-block;
   font-size: $font-size-s;
   line-height: 1.3;
+  margin-top: 10px;
   padding: 8px 12px 6px;
   text-align: center;
+
+  .Icon {
+    @include margin-end(10px);
+
+    vertical-align: middle;
+  }
 }
 
 .Badge-featured {
   background-color: $teal-50;
+}
 
-  .Icon-featured {
-    @include margin-end(10px);
+.Badge-restart-required {
+  background-color: $yellow-50;
 
-    vertical-align: middle;
+  .Icon-restart-required {
+    margin-top: -2px;
   }
 }
--- a/src/ui/components/Icon/styles.scss
+++ b/src/ui/components/Icon/styles.scss
@@ -116,6 +116,6 @@
   @include icon($name: "featured");
 }
 
-.Icon-restart-required {
+.Icon-restart {
   @include icon($name: "restart");
 }
--- a/src/ui/components/Icon/styles.scss
+++ b/src/ui/components/Icon/styles.scss
@@ -115,3 +115,8 @@
 .Icon-featured {
   @include icon($name: "featured");
 }
+
+.Icon-restart,
+.Icon-restart-required {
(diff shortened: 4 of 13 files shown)


src/amo/components/Addon/index.js
```diff
@@ -292,12 +293,17 @@ export class AddonBase extends React.Component {
...
+    let isRestartRequired = false;
     let compatibility;
     if (addon) {
       compatibility = getClientCompatibility({
         addon, clientApp, userAgentInfo,
       });
       isCompatible = compatibility.compatible;
+      isFeatured = addon.is_featured;
+      // themes do not have this attribute
+      isRestartRequired = addon.is_restart_required || false;
```
tofumatt: It would be nicer to just set this in the reducer rather than adding this check here. Would that be a huge pain?

src/amo/components/Addon/index.js
```diff
@@ -308,6 +314,21 @@ export class AddonBase extends React.Component {
             <h1 className="Addon-title" {...titleProps} />
             <p className="Addon-summary" {...summaryProps} />
 
+            <div className="Addon-badges">
+              {isFeatured ? (
+                <Badge
+                  type="featured"
+                  label={i18n.sprintf(i18n.gettext('Featured %(addonType)s'), { addonType })}
```
tofumatt: Sadly, `addonType` is not translated. We have to handle this in a few places by doing if/else or switch statements for all the different addon types... so it seems what we really need is an `i18n.addonType` helper.

tests/unit/core/reducers/test_addons.js
```diff
@@ -105,4 +106,24 @@ describe('addon reducer', () => {
       baz: { description: null },
     });
   });
+
+  it('pulls down the "is_restart_required" attribute from the file', () => {
```
tofumatt: This is a nitpick but `reads `is_restart_required` attribute from the file` might be a bit clearer? I dunno, "pulls down" seems overly verbose 😄

willdurand: I reused the sentence that was used for `installURL` 😄

tofumatt: We are bad role models! 😆

willdurand:
Updated code to fix things similar to those spotted in #2972. Only the second commit is relevant for this PR.

willdurand:
Fixed #2990 and updated the screenshots.

kumar303:
Can you add Fix #2990 to the description so it gets closed on merge?

willdurand:
@kumar303 did you delete a recent comment of yours here?

kumar303:
> @kumar303 did you delete a recent comment of yours here?

Yeah, sorry. I forgot what it was but it no longer applied.

src/ui/components/Icon/styles.scss
```diff
@@ -115,3 +115,8 @@
 .Icon-featured {
   @include icon($name: "featured");
 }
+
+.Icon-restart,
```
willdurand: It is not used, but I added this alias for consistency purpose.

tests/unit/core/reducers/test_addons.js
```diff
@@ -105,4 +106,24 @@ describe('addon reducer', () => {
...
     });
   });
+
+  it('reads "is_restart_required" attribute from the file', () => {
+    const addon = {
+      ...fakeAddon,
+      slug: 'installable',
+      current_version: {
+        ...fakeAddon.current_version,
+        files: [{ is_restart_required: true }, { file: 'data' }],
```
willdurand: ah, bad copy/paste again it seems 😞

src/amo/components/Addon/index.js
```diff
@@ -331,6 +333,12 @@ export class AddonBase extends React.Component {
                   label={this.getFeaturedText(addonType)}
                 />
               ) : null}
+              {isRestartRequired ? (
```
willdurand: I like the idea of a helper, but I do not really like having a state that is 1-1 with the API response. Can we change that?

I could add this helper to ship this feature/fix, but I would like to see a clear separation between app state and API responses. Can I open an issue for that?

kumar303: Yeah, I see what you're saying. I really want to throw that `addons` reducer out the window but let's do this. Let's continue with your approach. I'll add another review with some change requests to it.

src/core/reducers/addons.js
```diff
@@ -17,6 +17,7 @@ export function denormalizeAddon(apiAddon) {
       ...apiAddon,
       // Set iconUrl to be consistent between disco and amo.
       iconUrl: apiAddon.icon_url,
+      is_restart_required: apiAddon.is_restart_required || false,
```
kumar303: This doesn't look like it's necessary since you already have one below in the `else if (thisAddon.current_version...` block.

src/core/reducers/addons.js
```diff
@@ -48,6 +49,7 @@ export default function addon(state = initialState, action) {
         newState[key] = {
           ...thisAddon,
           installURL: thisAddon.current_version.files[0].url,
+          is_restart_required: thisAddon.current_version.files[0].is_restart_required,
```
kumar303: What if we called this `isRestartRequired` to make it a bit more obvious that it's a mapped property?

src/core/reducers/addons.js
kumar303: Well, even though it's not necessary (since it will only be false for themes or extensions without files) I see that you were using this code to initialize it to `false`. My main confusion here is that it looks like `is_restart_required` is being accessed on the API object which looks like a typo since the API doesn't return it.

First, I think renaming it to `isRestartRequired` will help.

Second, how about initializing it more directly, like this?

```diff
diff --git a/src/core/reducers/addons.js b/src/core/reducers/addons.js
index 1b051c34..1a935e2b 100644
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -28,7 +28,10 @@ export default function addon(state = initialState, action) {
   if (payload && payload.entities && payload.entities.addons) {
     const newState = { ...state };
     Object.keys(payload.entities.addons).forEach((key) => {
-      const thisAddon = payload.entities.addons[key];
+      const thisAddon = {
+        ...payload.entities.addons[key],
+        isRestartRequired: false,
+      };
       if (thisAddon.theme_data) {
         newState[key] = {
           ...thisAddon,
```

src/ui/components/Badge/styles.scss
```diff
@@ -10,16 +10,25 @@
...
+}
 
-  .Icon-featured {
-    @include margin-end(10px);
+.Badge-restart-required {
+  background-color: $yellow-50;
 
-    vertical-align: middle;
+  .Icon-restart-required {
+    margin-top: -2px;
```
willdurand: With margin:

screen shot 2017-08-24 at 17 54 26

Without:

screen shot 2017-08-24 at 17 54 22

What I can do is use a different `vertical-align` though. What's the best?

willdurand: Well actually, `vertical-align` changes the other badge, so I'd like to keep this negative margin..

willdurand:
Code updated and ready for a new review.

src/ui/components/Icon/styles.scss
kumar303: > It is not used, but I added this alias for consistency purpose.

How do you mean? I don't see any other styles like this in `Icon/styles.scss`. If it's unused I don't see why we need this style.

kumar303: Oh, is it because `$name` is `restart`? Why not make the name `restart-required`?

tests/unit/core/reducers/test_addons.js
```diff
@@ -105,4 +112,18 @@ describe('addon reducer', () => {
       baz: { description: null },
     });
   });
+
+  it('exposes "isRestartRequired" attribute from current version files', () => {
```
kumar303: This needs a test to make sure it sets `isRestartRequired` when at least one of the files declares it. You could modify the helper a bit and set it up like:

```js
const addon = createFakeAddon({
  files: [
    { ...fakeAddon.currentVersion.files[0], is_restart_required: false },
    { ...fakeAddon.currentVersion.files[0], is_restart_required: true },
  ],
});
```

tests/unit/core/reducers/test_addons.js
```diff
@@ -105,4 +112,18 @@ describe('addon reducer', () => {
...
+    const addon = createFakeAddon({ extraFileProps: { is_restart_required: true } });
```
kumar303: Since this hits a special branch of the reducer code, I think it would be good to write a test for the opposite of this too, set up like:

```js
const addon = createFakeAddon({ extraFileProps: { is_restart_required: false } });
```

src/ui/components/Icon/styles.scss
willdurand: Yes, I thought this could be used somewhere else actually..

willdurand:
Just added a few more tests, thanks!

src/ui/components/Badge/styles.scss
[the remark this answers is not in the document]
tofumatt: I'm fine with the negative margin here, fwiw. As long as this matches the spec I think it's 👍

src/ui/components/Icon/styles.scss
```diff
@@ -115,3 +115,7 @@
 .Icon-featured {
   @include icon($name: "featured");
 }
+
+.Icon-restart-required {
```
tofumatt: I think the name should match the icon file/etc. So this should just be called `.Icon-restart`.

tests/unit/core/reducers/test_addons.js
```diff
@@ -105,4 +112,72 @@ describe('addon reducer', () => {
...
+    ).toEqual({
+      [addon.slug]: denormalizeAddon({
+        ...addon,
+        installURL: '',
+        isRestartRequired: true,
+      }),
+    });
+  });
+
+  it('sets "isRestartRequired" to "false" when addon does not need restart', () => {
```
tofumatt: This is a near-OCD nitpick but I like the markdown convention of using backticks for things that are "code". So I'd rather see things like object properties and `false` written with <code>\`</code> instead of `"` around them. Wouldn't r- thing, but I'd probably change it next time I edited the file :wink:

src/ui/components/Icon/styles.scss
kumar303: Ah, yes, renaming this to `.Icon-restart` would have cleared up a lot of the confusion I had initially

Issue #2431, opened 2017-05-24
Integrate greenkeeperio/greenkeeper-lockfile


muffinresearch:
Looks like greenkeeper now have a package for keeping the yarn.lock up to date. 

Would be nice to look at integrating that on travis. See https://github.com/greenkeeperio/greenkeeper-lockfile for details.

willdurand:
No need to do this anymore, it is built-in! https://greenkeeper.io/docs.html#lockfiles

PR #2975, opened 2017-08-21
Update yarn.lock file

willdurand:


kumar303:
Thanks. If you pull from master, run yarn, and get a new lock file just push it up to master (no need for a PR). Maybe we can automate this one day which we started tracking here: https://github.com/mozilla/addons-frontend/issues/2431

willdurand:
@kumar303 ok thanks, will do next time. I was not sure about pushing directly to `master`.

Issue #2781, opened 2017-07-12
[tracker] User Profiles


tofumatt:
Create a page that allows the user to edit their profile:

 - [x] ~~Add users reducers/sagas #4030~~
 - [x] Add user profile view #2749 
 - [x] Add profile editing (AKA Account Settings) #3144

See [this document](https://docs.google.com/document/d/1cJ1k8HsAb6taauvJMDLOiYMa1McfUXNgEtax3hoToPc/edit#heading=h.zbgeax3cwqnz) for a break down of all user profile features.

kumar303:
Everything in this tracker has been nominated for MVP removal so I'm nominating the tracker.

muffinresearch:
Reviewed and removed from the MVP.

muffinresearch:
Refs: #3761

willdurand:
Closing this issue. I will open a new "tracker" issue for the edit page.

PR #2977, opened 2017-08-21
Fix sign-in while running a server from amo:dev

### LINKED ISSUE #2941: Cannot login with the dev server
Hi!

It looks like the dev server has an issue to authenticate users. 

1. run `yarn amo:dev`
2. go to the web interface and click "Log in/Sign up"
3. authenticate (maybe create an account first)
4. observe a 404 Page not found, without being authenticated

### What happened?

404 Page not found. Below is the console output:

```
index.js:57 [amo] Skipping componentWillMount() because visibleAddonType is invalid: fxa-authenticate
index.js:180 [amo] Rendering 404 because visibleAddonType is invalid: fxa-authenticate
```

Being authenticated.

kumar303:
Fixes #2941 

This does a few things:
* Both `amo:dev` and `amo:stage` go through the proxy which means both the frontend site and the backend API are served on the same domain like production. This should help catch same origin issues locally.
* Sign-ins work on `amo:dev`
* Sign-ins do not work on `amo:stage` (like before) because we need to configure a new FxA client. I'll do that in a follow-up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,10 +67,11 @@
       }
     },
     "amo:stage": {
-      "command": "better-npm-run start-dev",
+      "command": "better-npm-run start-dev-proxy",
       "env": {
         "AMO_CDN": "https://addons-stage-cdn.allizom.org",
-        "API_HOST": "https://addons.allizom.org",
+        "PROXY_API_HOST": "https://addons.allizom.org",
+        "FXA_CONFIG": "local",
         "CSP": false,
         "NODE_APP_INSTANCE": "amo"
       }
--- a/package.json
+++ b/package.json
@@ -50,10 +50,12 @@
       }
     },
     "amo:dev": {
-      "command": "better-npm-run start-dev",
+      "command": "better-npm-run start-dev-proxy",
       "env": {
         "AMO_CDN": "https://addons-amo-dev-cdn.allizom.org",
         "API_HOST": "https://addons-dev.allizom.org",
+        "PROXY_API_HOST": "https://addons-dev.allizom.org",
+        "FXA_CONFIG": "local",
         "CSP": false,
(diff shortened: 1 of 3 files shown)


willdurand:
🎉  💯 

Just tested this PR and it works well. Thanks @kumar303. It also works for newly created accounts (which did not work either before), except that the user gets redirected to `http://localhost:3000/en-US/firefox/users/edit`, which does not exit (404). Logout works too.

kumar303:
@willdurand great, thanks for testing it out. I am reminded now that the redirection 404 (for new accounts) is a known issue that will be fixed when https://github.com/mozilla/addons-frontend/issues/2781 lands. However, with my patch you will be logged in when you type in the URL `localhost:3000` after the 404 so it's not too annoying.

tofumatt:
Just testing this now, I was beat up by a migraine most of today, sorry! 😢

package.json
```diff
@@ -66,10 +67,11 @@
       }
     },
     "amo:stage": {
-      "command": "better-npm-run start-dev",
+      "command": "better-npm-run start-dev-proxy",
```
tofumatt: Shoot! My bad, I think this change was my fault 😢

kumar303:
btw, I also tested add-on/theme installation through a [custom Nightly build](https://github.com/mozilla/addons-frontend/blob/master/docs/moz-addon-manager.md#developing-with-a-local-server)

kumar303:
@willdurand be sure to clear your cookies when switching between `yarn amo:dev`, `yarn amo`, and `yarn amo:stage` since they all share the same domain (`localhost:3000`).

willdurand:
ah good to know, thanks!

PR #2979, opened 2017-08-21
Add i18n.addonType helper

### LINKED ISSUE #2978: addonType is not translated; create a helper for this
Right now we translate entire strings because the `addonType` we use internally is not translated. This leads to a lot of duplicated effort on the part of translators and more code for us. See stuff like https://github.com/mozilla/addons-frontend/blob/cfcd838c812b1d0f2aa84c47c498f971e4a417d0/src/amo/components/RatingManager/index.js#L123

Really we should just have an `i18n.addonType(addonType)` that returns a translated name.

willdurand:
Fix #2978

### DIFF
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -5,6 +5,15 @@ import Jed from 'jed';
 import moment from 'moment';
 
 import log from 'core/logger';
+import {
+  ADDON_TYPE_COMPLETE_THEME,
+  ADDON_TYPE_DICT,
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_LANG,
+  ADDON_TYPE_OPENSEARCH,
+  ADDON_TYPE_THEME,
+  validAddonTypes,
+} from 'core/constants';
 
 const defaultLang = config.get('defaultLang');
 const langs = config.get('langs');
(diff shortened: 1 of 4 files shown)


src/core/i18n/utils.js
```diff
@@ -287,6 +295,29 @@ export function makeI18n(
       oneLineTranslationString(pluralKey), val);
   };
 
+  // Because the `addonType` we use internally is not translated, we needs this
```
tofumatt: Nitpick editorialising, but this could be:

```js
// The `addonType` we use internally is not translated, so we use this
// helper to return translated add-on type names.
```

tests/unit/core/i18n/test_utils.js
```diff
@@ -475,5 +486,25 @@ describe('i18n utils', () => {
...
+
+    describe('addonType()', () => {
+      it('translates internal addon types', () => {
+        const i18n = utils.makeI18n({}, 'en', FakeJed);
+        expect(i18n.addonType(ADDON_TYPE_DICT)).toEqual('dictionary');
+        expect(i18n.addonType(ADDON_TYPE_EXTENSION)).toEqual('extension');
+        expect(i18n.addonType(ADDON_TYPE_LANG)).toEqual('language pack');
+        expect(i18n.addonType(ADDON_TYPE_OPENSEARCH)).toEqual('search plugin');
+        expect(i18n.addonType(ADDON_TYPE_THEME)).toEqual('theme');
+        // this one seems deprecated
```
tofumatt: It is but it will still appear for older (ESR) clients I think? I would include it as "theme" as well.

src/core/i18n/utils.js
```diff
@@ -287,6 +295,29 @@ export function makeI18n(
...
+  // helper to return the translated names without effort.
+  i18n.addonType = function addonType(type) {
+    switch (type) {
+      case ADDON_TYPE_DICT:
+        return i18n.gettext('dictionary');
+      case ADDON_TYPE_LANG:
+        return i18n.gettext('language pack');
+      case ADDON_TYPE_OPENSEARCH:
+        return i18n.gettext('search plugin');
+      case ADDON_TYPE_THEME:
```
tofumatt: I think you can just add `case ADDON_TYPE_COMPLETE_THEME:` here; as long as we "support" it we should return a value.

kumar303:
This is my first time seeing #2978 . I don't think it's safe to replace `%(addonsType)s` in localizable strings because it's hard to anticipate all the grammatical fluctuations. For example, some languages may use different conjugations of the word *language pack* depending on the context of the sentence. The rules for pluralization can also vary between language.

tofumatt:
As mentioned on IRC we actually shouldn't do this, as advised by localisers. There are too many edge-cases we can't anticipate. Sorry about that @willdurand  😢

Issue #1820, opened 2017-02-15
Store auth token/payload in an `account` or `user` reducer


mstriemer:
We need to display some information about the logged in user in the UI. Specifically in #1751, we'd like to display the username instead of the text **Log out** in `AuthenticateButton` when the user is logged in.

The auth token contains the user's email, user ID and username. That data should be parsed from the auth token and put into this reducer.

Blocks #1751, supports #1644.

mstriemer:
You can find some information on how to parse JWTs at https://jwt.io/.

mstriemer:
Actually we already have a function to [decode the user ID](https://github.com/mozilla/addons-frontend/blob/711bf11f5af28c8a0e0904bedfeecc9f299a5cc8/src/core/reducers/authentication.js#L10). That could be updated to return the username and email address too and then that could be hooked up to this reducer.

kumar303:
Yeah, it stores the [userId in state](https://github.com/mozilla/addons-frontend/blob/711bf11f5af28c8a0e0904bedfeecc9f299a5cc8/src/core/reducers/authentication.js#L38) so that could easily be extended to store user name / email.

muffinresearch:
This will be impacted by moving to an HMAC token.

kumar303:
This is a dupe of https://github.com/mozilla/addons-frontend/issues/1644 or at least they can be done at the same time

Issue #1751, opened 2017-02-08
Log Out button should have user's name on it 


vcarciu:
Steps to reproduce:
1.On Android, go to addons-dev.allizom.org and log in with valid credentials

Expected results:
According to mocks, the log-out button should contain user's display name : 
![usernamemocks](https://cloud.githubusercontent.com/assets/1583842/22748839/15f32710-ee34-11e6-8f42-38953c6dd38d.png)

Actual results:
Button's text is "Log out"

Please see screenshot for this issue : 
![usernameactual](https://cloud.githubusercontent.com/assets/1583842/22748875/31e1b6e4-ee34-11e6-83a5-0005cd27e2dc.png)

tofumatt:
I think we could move this out of V1 to V2. Log out seems to do enough for now, and there is minimal space for long usernames here.

mstriemer:
It would be nice to know who you're logged in as so I wouldn't be opposed to this. It does seem strange to click your username and get logged out so we'd need to add a dropdown menu to this, potentially. Perhaps we could cheat and do tap your name, text changes to log out, tap again you get logged out.

vcarciu:
Maybe we can split the button in 2 sections , one with log out and one with the name which will redirect to user's info(I can see that this is requested in mocks too)

tofumatt:
> It would be nice to know who you're logged in as so I wouldn't be opposed to this. It does seem strange to click your username and get logged out so we'd need to add a dropdown menu to this, potentially. Perhaps we could cheat and do tap your name, text changes to log out, tap again you get logged out.

Yeah I like that. I guess I was thinking more for v1. But yeah, eventually showing your username would be nice.

mstriemer:
@pwalm what would you like us to do here? Is just **Log out** okay for now or would you rather see the name as well?

pwalm:
We had is mocked up so that it shows the user name, the user would tap it to go to their profile/settings page, and log out would be at the bottom of that. Can we do that?

aayushsanghavi:
I would like to take this up. @tofumatt what variable/object has the username that we would have to display - https://github.com/mozilla/addons-frontend/blob/master/src/core/components/AuthenticateButton/index.js#L36 ?

tofumatt:
@aayushsanghavi That variable would be in the redux state somewhere, I'm not sure where. If you don't have a good understanding of Redux yet this might be a pain bug to take on.

[...]

Issue #1644, opened 2017-01-20
Remove authentication reducer


mstriemer:
The `authentication` reducer has a lot of duplication with the `api` reducer. It seems to store the JWT and some info about the user but that user info doesn't seem to be exposed anywhere (on amo anyway). It does look like some code is checking if `auth.token` is set to detect if the user is logged in.

It is probably a good idea to remove the `authentication` reducer and add any user-specific data that we need to a `user` or `account` reducer.

There was a bug with logging out due to this that was fixed in #1643.

kumar303:
oof, yeah, I'd love to clean up some other reducers too

PR #2989, opened 2017-08-23
Retrieve user profile on SET_AUTH_TOKEN 🔥

willdurand:
Refs #1820 #1644 #1751

---

This PR does a few things, but the main goal was to fetch the current logged user profile. I hooked into the `SET_AUTH_TOKEN` action to fetch the user profile with a saga. This action is cool because it is used in both SSR and not SSR.

### What has changed?

- The `userId` has moved to the `user` reducer now (it was in `api` since #2983).
- I removed the JWT token parsing to exclusively rely on the user profile data. This is now gone forever, but if we want to validate this token somehow, maybe we could re-introduce the method with a different name, _i.e._ `validateToken()`.

### What has been added?

- a `user` ducks that owns the current logged user profile
- a `user` saga that fetches the user profile on `SET_AUTH_TOKEN`
- a selector `isAuthenticated` in `user` reducer to know whether user is logged in

### What is missing? // TODO

- [x] write tests to cover the change in `core/server/base`. The `setAuthToken()` call to dispatch `SET_AUTH_TOKEN` was triggered too early and the saga could not catch it. I had to move the saga instantiation earlier. See this part of the code for further details: https://github.com/willdurand/addons-frontend/blob/master/src/core/server/base.js#L244-L260
- [x] how to handle error in the `user` saga? Since it is not created by UI, I don't know how to pass a error handler (id). Not sure if I have to actually..

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -22,7 +22,6 @@ const Entity = normalizrSchema.Entity;
 
 export const addon = new Entity('addons', {}, { idAttribute: 'slug' });
 export const category = new Entity('categories', {}, { idAttribute: 'slug' });
-export const user = new Entity('users', {}, { idAttribute: 'username' });
 
 export function makeQueryString(query: { [key: string]: * }) {
   const resolvedQuery = { ...query };
@@ -212,15 +211,6 @@ export function startLoginUrl(
   }
   const query = makeQueryString(params);
   return `${API_BASE}/accounts/login/start/${query}`;
-}
-
-export function fetchProfile({ api }: {| api: ApiStateType |}) {
-  return callApi({
-    endpoint: 'accounts/profile',
-    schema: user,
-    auth: true,
-    state: api,
-  });
 }
 
 type FeaturedParams = {|
--- a/src/core/api/user.js
+++ b/src/core/api/user.js
@@ -4,6 +4,10 @@ import type { ApiStateType } from 'core/reducers/api';
 
 
 export function userProfile({ api }: {| api: ApiStateType |}) {
+  if (!api) {
+    throw new Error('The api state is required.');
+  }
+
   return callApi({
     auth: true,
     endpoint: 'accounts/profile',
--- a/src/core/api/user.js
+++ b/src/core/api/user.js
@@ -1,0 +1,12 @@
+/* @flow */
+import { callApi } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+
+
+export function userProfile({ api }: {| api: ApiStateType |}) {
+  return callApi({
+    auth: true,
+    endpoint: 'accounts/profile',
+    state: api,
+  });
+}
--- a/src/core/containers/LoginRequired/index.js
+++ b/src/core/containers/LoginRequired/index.js
@@ -4,12 +4,8 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import LoginPage from 'core/components/LoginPage';
+import { isAuthenticated } from 'core/reducers/user';
 
-export function mapStateToProps(state) {
-  return {
-    authenticated: !!state.user.id,
-  };
-}
 
 // This class is exported for testing outside of redux.
 export class LoginRequiredBase extends React.Component {
@@ -26,6 +22,12 @@ export class LoginRequiredBase extends React.Component {
     }
     return <LoginPage location={location} />;
   }
+}
+
+export function mapStateToProps(state) {
+  return {
+    authenticated: isAuthenticated(state),
+  };
 }
 
 export default compose(
--- a/src/core/containers/LoginRequired/index.js
+++ b/src/core/containers/LoginRequired/index.js
@@ -7,7 +7,7 @@ import LoginPage from 'core/components/LoginPage';
 
 export function mapStateToProps(state) {
   return {
-    authenticated: !!state.api.token,
+    authenticated: !!state.user.id,
   };
 }
 
--- a/src/core/reducers/api.js
+++ b/src/core/reducers/api.js
@@ -1,5 +1,4 @@
 /* @flow */
-import base64url from 'base64url';
 import UAParser from 'ua-parser-js';
 
 import {
@@ -9,7 +8,6 @@ import {
   SET_CLIENT_APP,
   SET_USER_AGENT,
 } from 'core/constants';
-import log from 'core/logger';
 import type {
   SetAuthTokenAction,
   LogOutUserAction,
@@ -36,7 +34,6 @@ export type ApiStateType = {
   token: ?string,
   userAgent: ?string,
   userAgentInfo: UserAgentInfoType,
-  userId: ?number,
 };
 
 export const initialApiState: ApiStateType = {
@@ -45,26 +42,7 @@ export const initialApiState: ApiStateType = {
   token: null,
   userAgent: null,
   userAgentInfo: { browser: {}, os: {} },
-  userId: null,
 };
-
-function getUserIdFromAuthToken(token) {
-  let data;
-  try {
-    const parts = token.split(':');
-    if (parts.length < 3) {
-      throw new Error('not enough auth token segments');
-    }
-    data = JSON.parse(base64url.decode(parts[0]));
-    log.info('decoded auth token data:', data);
-    if (!data.user_id) {
-      throw new Error('user_id is missing from decoded data');
-    }
-    return data.user_id;
-  } catch (error) {
-    throw new Error(`Error parsing auth token "${token}": ${error}`);
-  }
-}
 
 export default function api(
   state: Exact<ApiStateType> = initialApiState,
@@ -79,10 +57,6 @@ export default function api(
       return {
         ...state,
         token: action.payload.token,
-        // Extract user data from the auth token (which is loaded from a cookie
-        // on each request). This doesn't check the token's signature
-        // because the server is responsible for that.
-        userId: getUserIdFromAuthToken(action.payload.token),
       };
     case SET_LANG:
       return { ...state, lang: action.payload.lang };
@@ -99,7 +73,7 @@ export default function api(
       };
     }
     case LOG_OUT_USER:
-      return { ...state, token: null, userId: null };
+      return { ...state, token: null };
     default:
       return state;
   }
--- a/src/core/reducers/user.js
+++ b/src/core/reducers/user.js
@@ -1,0 +1,37 @@
+import { LOG_OUT_USER } from 'core/constants';
+
+
+const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
+
+export const initialState = {
+  id: null,
+  username: null,
+};
+
+export const userProfileLoaded = ({ profile }) => {
+  if (!profile) {
+    throw new Error('The profile parameter is required.');
+  }
+
+  return {
+    type: USER_PROFILE_LOADED,
+    payload: { profile },
+  };
+};
+
+export default function reducer(state = initialState, action = {}) {
+  const { payload } = action;
+
+  switch (action.type) {
+    case USER_PROFILE_LOADED:
+      return {
+        ...state,
+        id: payload.profile.id,
+        username: payload.profile.username,
+      };
+    case LOG_OUT_USER:
+      return initialState;
+    default:
+      return state;
+  }
+}
--- a/src/core/reducers/user.js
+++ b/src/core/reducers/user.js
@@ -26,6 +26,10 @@ export const userProfileLoaded = ({ profile }: Object) => {
   };
 };
 
+export const isAuthenticated = (state: { user: UserStateType }) => {
+  return !!state.user.id;
+};
+
 export default function reducer(
   state: UserStateType = initialState,
   action: Action = {}
--- a/src/core/reducers/user.js
+++ b/src/core/reducers/user.js
@@ -1,14 +1,21 @@
+/* @flow */
 import { LOG_OUT_USER } from 'core/constants';
 
 
 const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
 
-export const initialState = {
+type Action = Object;
+export type UserStateType = {
+  id: ?number,
+  username: ?string,
+};
+
+export const initialState: UserStateType = {
   id: null,
   username: null,
 };
 
-export const userProfileLoaded = ({ profile }) => {
+export const userProfileLoaded = ({ profile }: Object) => {
   if (!profile) {
     throw new Error('The profile parameter is required.');
   }
@@ -19,7 +26,10 @@ export const userProfileLoaded = ({ profile }) => {
   };
 };
 
-export default function reducer(state = initialState, action = {}) {
+export default function reducer(
+  state: UserStateType = initialState,
+  action: Action = {}
+): UserStateType {
   const { payload } = action;
 
   switch (action.type) {
--- a/src/core/reducers/user.js
+++ b/src/core/reducers/user.js
@@ -2,7 +2,7 @@
 import { LOG_OUT_USER } from 'core/constants';
 
 
-const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
+const LOAD_USER_PROFILE = 'LOAD_USER_PROFILE';
 
 type Action = Object;
 export type UserStateType = {
@@ -15,13 +15,13 @@ export const initialState: UserStateType = {
   username: null,
 };
 
-export const userProfileLoaded = ({ profile }: Object) => {
+export const loadUserProfile = ({ profile }: Object) => {
   if (!profile) {
     throw new Error('The profile parameter is required.');
   }
 
   return {
-    type: USER_PROFILE_LOADED,
+    type: LOAD_USER_PROFILE,
     payload: { profile },
   };
 };
@@ -37,7 +37,7 @@ export default function reducer(
   const { payload } = action;
 
   switch (action.type) {
-    case USER_PROFILE_LOADED:
+    case LOAD_USER_PROFILE:
       return {
         ...state,
         id: payload.profile.id,
--- a/src/core/sagas/user.js
+++ b/src/core/sagas/user.js
@@ -1,0 +1,32 @@
+// Disabled because of
+// https://github.com/benmosher/eslint-plugin-import/issues/793
+/* eslint-disable import/order */
+import { call, put, select, takeLatest } from 'redux-saga/effects';
+/* eslint-enable import/order */
+
+import { SET_AUTH_TOKEN } from 'core/constants';
+import { getState } from 'core/sagas/utils';
+import { userProfileLoaded } from 'core/reducers/user';
+import { userProfile as userProfileApi } from 'core/api/user';
+
+// This saga is not triggered by the UI but on the server side, hence do not
+// have a `errorHandler`. We do not want to miss any error because it would
+// mean no ways for the users to log in, so we let the errors bubble up.
+export function* fetchUserProfile({ payload }) {
+  const { token } = payload;
+
+  const state = yield select(getState);
+
+  const response = yield call(userProfileApi, {
+    api: {
+      ...state.api,
+      token,
+    },
+  });
+
+  yield put(userProfileLoaded({ profile: response }));
+}
+
+export default function* userSaga() {
+  yield takeLatest(SET_AUTH_TOKEN, fetchUserProfile);
+}
--- a/src/core/sagas/user.js
+++ b/src/core/sagas/user.js
@@ -6,7 +6,7 @@ import { call, put, select, takeLatest } from 'redux-saga/effects';
 
 import { SET_AUTH_TOKEN } from 'core/constants';
 import { getState } from 'core/sagas/utils';
-import { userProfileLoaded } from 'core/reducers/user';
+import { loadUserProfile } from 'core/reducers/user';
 import { userProfile as userProfileApi } from 'core/api/user';
 
 // This saga is not triggered by the UI but on the server side, hence do not
@@ -24,7 +24,7 @@ export function* fetchUserProfile({ payload }) {
     },
   });
 
-  yield put(userProfileLoaded({ profile: response }));
+  yield put(loadUserProfile({ profile: response }));
 }
 
 export default function* userSaga() {
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -3,7 +3,7 @@ import path from 'path';
 
 import 'babel-polyfill';
 import { oneLine } from 'common-tags';
-import config from 'config';
+import defaultConfig from 'config';
 import Express from 'express';
 import helmet from 'helmet';
 import Raven from 'raven';
@@ -36,16 +36,10 @@ import I18nProvider from 'core/i18n/Provider';
 import WebpackIsomorphicToolsConfig from './webpack-isomorphic-tools-config';
 
 
-const env = config.util.getEnv('NODE_ENV');
-// This is a magic file that gets written by deployment scripts.
-const version = path.join(config.get('basePath'), 'version.json');
-const isDeployed = config.get('isDeployed');
-const isDevelopment = config.get('isDevelopment');
+export function getPageProps({ noScriptStyles = '', store, req, res, config }) {
+  const appName = config.get('appName');
+  const isDeployed = config.get('isDeployed');
 
-
-const appName = config.get('appName');
-
-export function getPageProps({ noScriptStyles = '', store, req, res }) {
   // Get SRI for deployed services only.
   const sriData = (isDeployed) ? JSON.parse(
     fs.readFileSync(path.join(config.get('basePath'), 'dist/sri.json'))
@@ -86,9 +80,9 @@ export function getPageProps({ noScriptStyles = '', store, req, res }) {
   };
 }
 
-function showErrorPage({ createStore, error = {}, req, res, status }) {
+function showErrorPage({ createStore, error = {}, req, res, status, config }) {
   const { store } = createStore();
-  const pageProps = getPageProps({ store, req, res });
+  const pageProps = getPageProps({ store, req, res, config });
 
   const componentDeclaredStatus = NestedStatus.rewind();
   let adjustedStatus = status || componentDeclaredStatus || 500;
@@ -116,7 +110,13 @@ function hydrateOnClient({ res, props = {}, pageProps }) {
     .end();
 }
 
-function baseServer(routes, createStore, { appSagas, appInstanceName = appName } = {}) {
+function baseServer(routes, createStore, {
+  appSagas,
+  appInstanceName = null,
+  config = defaultConfig,
+} = {}) {
+  const appName = appInstanceName !== null ? appInstanceName : config.get('appName');
+
   const app = new Express();
   app.disable('x-powered-by');
 
@@ -127,8 +127,7 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
     log.info(`Sentry reporting configured with DSN ${sentryDsn}`);
     // The error handler is defined below.
   } else {
-    log.warn(
-      'Sentry reporting is disabled; Set config.sentryDsn to enable it.');
+    log.warn('Sentry reporting is disabled; Set config.sentryDsn to enable it.');
   }
 
   app.use(middleware.logRequests);
@@ -156,6 +155,9 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
 
   // Show version/commit information as JSON.
   function viewVersion(req, res) {
+    // This is a magic file that gets written by deployment scripts.
+    const version = path.join(config.get('basePath'), 'version.json');
+
     fs.stat(version, (error) => {
       if (error) {
         log.error(`Could not stat version file ${version}: ${error}`);
@@ -175,7 +177,8 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
   // Return 200 for csp reports - this will need to be overridden when deployed.
   app.post('/__cspreport__', (req, res) => res.status(200).end('ok'));
 
-  if (appInstanceName === 'disco' && isDevelopment) {
+  const isDevelopment = config.get('isDevelopment');
+  if (appName === 'disco' && isDevelopment) {
     app.get('/', (req, res) =>
       res.redirect(302, '/en-US/firefox/discovery/pane/48.0/Darwin/normal'));
   }
@@ -224,12 +227,13 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
       }
       if (!renderProps) {
         log.info(`match() did not return renderProps for ${req.url}`);
-        return showErrorPage({ createStore, status: 404, req, res });
+        return showErrorPage({ createStore, status: 404, req, res, config });
       }
 
       let htmlLang;
       let locale;
       let pageProps;
+      let runningSagas;
       let sagaMiddleware;
       let store;
 
@@ -239,23 +243,34 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
         const storeAndSagas = createStore();
         sagaMiddleware = storeAndSagas.sagaMiddleware;
         store = storeAndSagas.store;
+
+        let sagas = appSagas;
+        if (!sagas) {
+          // eslint-disable-next-line global-require, import/no-dynamic-require
+          sagas = require(`${appName}/sagas`).default;
+        }
+        runningSagas = sagaMiddleware.run(sagas);
+
         const token = cookie.load(config.get('cookieName'));
         if (token) {
           store.dispatch(setAuthToken(token));
         }
 
-        pageProps = getPageProps({ noScriptStyles, store, req, res });
+        pageProps = getPageProps({ noScriptStyles, store, req, res, config });
         if (config.get('disableSSR') === true) {
-          log.warn(
-            'Server side rendering disabled; responding without loading');
-          return hydrateOnClient({ res, pageProps });
+          // This stops all running sagas.
+          store.dispatch(END);
+
+          return runningSagas.done.then(() => {
+            log.warn('Server side rendering is disabled.');
+            return hydrateOnClient({ res, pageProps });
+          });
         }
 
         htmlLang = pageProps.htmlLang;
         locale = langToLocale(htmlLang);
       } catch (preLoadError) {
-        log.info(
-          `Caught an error in match() before loadOnServer(): ${preLoadError}`);
+        log.info(`Caught an error in match() before loadOnServer(): ${preLoadError}`);
         return next(preLoadError);
       }
 
@@ -266,14 +281,11 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
           try {
             if (locale !== langToLocale(config.get('defaultLang'))) {
               // eslint-disable-next-line global-require, import/no-dynamic-require
-              i18nData = require(
-                `../../locale/${locale}/${appInstanceName}.js`);
+              i18nData = require(`../../locale/${locale}/${appName}.js`);
             }
           } catch (e) {
-            log.info(
-              `Locale JSON not found or required for locale: "${locale}"`);
-            log.info(
-              `Falling back to default lang: "${config.get('defaultLang')}".`);
+            log.info(`Locale JSON not found or required for locale: "${locale}"`);
+            log.info(`Falling back to default lang: "${config.get('defaultLang')}".`);
           }
           const i18n = makeI18n(i18nData, htmlLang);
 
@@ -292,12 +304,6 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
           }
 
           const props = { component: InitialComponent };
-          let sagas = appSagas;
-          if (!sagas) {
-            // eslint-disable-next-line global-require, import/no-dynamic-require
-            sagas = require(`${appName}/sagas`).default;
-          }
-          const runningSagas = sagaMiddleware.run(sagas);
 
           // We need to render once because it will force components to
           // dispatch data loading actions which get processed by sagas.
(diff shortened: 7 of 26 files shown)


tests/unit/core/server/testServer.js
```diff
@@ -6,57 +6,74 @@ import { reducer as reduxAsyncConnect } from 'redux-connect';
...
-      return {
-        store: createStore(
-          combineReducers({ reduxAsyncConnect }),
-          // Do not define an initial state.
-          undefined,
-          applyMiddleware(sagaMiddleware),
-        ),
-        sagaMiddleware,
-      };
+      store = createStore(
```
kumar303: I know it's convenient but it's too confusing to let `testClient()` have the side effect of creating a store in the outer scope. Let me know if you need ideas for how to fix it.

tests/unit/core/server/testServer.js
```diff
@@ -6,57 +6,74 @@ import { reducer as reduxAsyncConnect } from 'redux-connect';
...
-      appSagas: fakeSaga,
+      appSagas: appSagas || fakeSaga,
       appInstanceName: 'testapp',
     });
     return supertest(app);
   }
 
   describe('app', () => {
     it('varies on DNT', async () => {
-      const response = await testClient().get('/en-US/firefox/').end();
```
kumar303: Unfortunately, `.end()` must always be the final call to the `testClient()` chain. This is because `end()` will shut down the server. So, without it, every one of these tests will be spawning a server that will never get shut down. It's the most terrible interface ever written. See https://github.com/visionmedia/supertest/issues/100

I originally couldn't think of a way to do this automatically but maybe we can just use a wrapper around `testClient()` ?

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +100,66 @@ describe('core/server/base', () => {
       );
 
       const response = await testClient({ stubRoutes })
-        .get('/en-US/firefox/simulation-of-a-non-existent-page').end();
+        .get('/en-US/firefox/simulation-of-a-non-existent-page')
+      ;
```
kumar303: for consistency with the rest of the code base, the semi-colon should always go at the end of the line.

We'll add a rule 😇 https://github.com/mozilla/eslint-config-amo/issues/24

src/core/sagas/user.js
```diff
@@ -0,0 +1,34 @@
...
+    const response = yield call(userProfileApi, {
+      api: {
+        ...state.api,
+        token,
+      },
+    });
+
+    yield put(userProfileLoaded({ profile: response }));
+  } catch (error) {
+    log.warn(`Current user profile failed to load: ${error}`);
```
willdurand: perfect!

src/core/sagas/user.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+export default function* userSaga() {
+  yield takeLatest(SET_AUTH_TOKEN, fetchUserProfile);
```
willdurand: We need the `username` in the header for instance, so I think we'll need to fetch the user profile anyway. I chose this action because it seems to be done only once on first hit (+ page reloads).

src/core/reducers/user.js
```diff
@@ -0,0 +1,33 @@
...
+
+
+const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
+
+const initialState = {
+  id: null,
+  username: null,
+};
+
+export const userProfileLoaded = ({ profile }) => {
```
willdurand: oh right!

src/core/api/index.js
```diff
@@ -327,3 +328,5 @@ export function autocomplete({ api, filters }: AutocompleteParams) {
     state: api,
   });
 }
+
+export const userProfile = userApi.userProfile;
```
willdurand: heh I dont know, I saw @tofumatt doing something similar too.

tests/unit/core/server/testServer.js
willdurand: Then, I will pass the store to the `testClient()`

tests/unit/core/server/testServer.js
willdurand: Oh ok 😢  

I did that to get rid of warnings saying that there was a "double callback" problem, see: https://github.com/visionmedia/supertest/issues/141.

tests/unit/core/server/testServer.js
willdurand: oups

willdurand:
I have updated the code.

src/core/server/base.js
```diff
@@ -230,6 +230,7 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
       let htmlLang;
       let locale;
       let pageProps;
+      let runningSagas;
```
tofumatt: Any reason this needs to be scoped up here? I know the scoping in this file is a bit broad and that's fine, just double-checking. The GitHub diff on this one is a bit harder to get context from.

tests/unit/core/api/test_user.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    it('fetches the current user profile', () => {
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/accounts/profile/?lang=en-US`)
+        .once()
+        .returns(mockResponse());
+
+      return api.userProfile({
+        api: { clientApp: CLIENT_APP_ANDROID, lang: 'en-US' },
```
tofumatt: This should use `dispatchClientMetadata().store.getState().api`. I know it's way more verbose/work but we should definitely get away from stubbing out raw state. We used to do it and it always leads to refactors or oddness.

tests/unit/core/containers/TestServerHtml.js
```diff
@@ -12,14 +12,14 @@ import FakeApp, {
 } from 'tests/unit/core/server/fakeApp';
 
 describe('<ServerHtml />', () => {
-  const _helmentCanUseDOM = Helmet.canUseDOM;
```
tofumatt: lol Nice catch! 😆

tests/unit/core/reducers/test_user.js
```diff
@@ -0,0 +1,42 @@
...
+import reducer, {
+  userProfileLoaded,
+} from 'core/reducers/user';
+import { createUserProfileResponse } from 'tests/unit/helpers';
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('initializes properly', () => {
+      const { id, username } = reducer(undefined);
+      expect(id).toBe(null);
```
tofumatt: I'd be fine if we just used `toEqual` everywhere because it's easier to read/rewrite/etc. I don't think we'll get bit by using it over `toBe`. @muffinresearch and I chatted about this and it's just nicer to have fewer matchers IMO.

That said if you left it here I'd be fine with it. Just a note.

tests/unit/core/reducers/test_user.js
```diff
@@ -0,0 +1,42 @@
...
+      expect(username).toEqual(null);
+    });
+
+    it('ignores unrelated actions', () => {
+      const state = { id: 12345, username: 'john' };
```
tofumatt: Could this use the `initialiState` from the actual reducer instead? I just like to stick with realistic data.

src/core/sagas/user.js
```diff
@@ -0,0 +1,29 @@
...
+import { getState } from 'core/sagas/utils';
+import { userProfileLoaded } from 'core/reducers/user';
+import { userProfile as userProfileApi } from 'core/api/user';
+
+export function* fetchUserProfile({ payload }) {
+  const { token } = payload;
+
+  const state = yield select(getState);
+
+  const response = yield call(userProfileApi, {
```
tofumatt: I think this should have a `try/catch` around it... at least that's what we do for all of our other sagas and then we use `yield put(errorHandler.createErrorAction(error));` to dispatch an error.

See something like the search saga: https://github.com/mozilla/addons-frontend/blob/master/src/core/sagas/search.js#L21

tests/unit/core/sagas/test_user.js
```diff
@@ -0,0 +1,63 @@
...
+    const expectedCalledAction = userProfileLoaded({ profile });
+
+    await sagaTester.waitFor(expectedCalledAction.type);
+    mockApi.verify();
+
+    const calledAction = sagaTester.getCalledActions()[1];
+    expect(calledAction).toEqual(expectedCalledAction);
+  });
+
+  it('lets exceptions to be thrown', () => {
```
tofumatt: I think this test spec could be better written as `it('handles exceptions and dispatches an error', () => { [...] })`

But then as mentioned before I think the way this saga handles errors isn't how we tend to. I guess we should write some documentation at least on error handling, because we have a HOC and reducer for dispatching errors that components can consume.

tests/unit/core/reducers/test_user.js
willdurand: can I `sed -i s/toBe/toEqual` in another PR maybe?

src/core/server/base.js
willdurand: That's because `store` and `sagaMiddleware` are there too..

tests/unit/core/reducers/test_user.js
tofumatt: Yeah, sure, it's really no biggie. I just figured I'd mention for future patches. I think `toEqual` is a nicer matcher too; `toBe` implies more of an exact type-check not a value-check.

tests/unit/core/reducers/test_user.js
willdurand: That would mean exposing the `initialState`?

src/core/server/base.js
tofumatt: Cool, that's what I thought, makes sense! 👍

src/core/sagas/user.js
willdurand: Yeah I know how this works, but the UI gives the error handler ID, which is not possible here. That's why @kumar303 suggested to bubble up the errors: https://github.com/mozilla/addons-frontend/pull/2989#discussion_r134889238.

tests/unit/core/reducers/test_user.js
tofumatt: Yeah, [we do that in a fair number of reducers](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/featured.js#L4), I think it's fine.

src/core/sagas/user.js
tofumatt: Ahhhh, right. Okay, can you add a comment in here explaining why this saga is different than usual then and why we aren't using `errorHandler`? Sorry I missed some of the comments here.

willdurand: Yes, just updated the code.

tests/unit/core/sagas/test_user.js
```diff
@@ -0,0 +1,63 @@
...
+    const error = new Error('some API error maybe');
+    mockApi
+      .expects('userProfile')
+      .returns(Promise.reject(error));
+
+    sagaTester.dispatch(setAuthToken(userAuthToken()));
+
+    expect.assertions(1);
+    return rootTask.done.catch((e) => {
```
kumar303: I find `expect.assertions(N)` to be hard to maintain as apps grow. They also result in poor failure messages.

Here's a more robust way to do it:

```js
return rootTask.done
  .then(() => {
    throw new Error('unexpected success');
  })
  .catch((sagaError) => {
    mockApi.verify();
    expect(sagaError).toBe(error);
  });
```

btw, we try to avoid single character variable names in the code base.

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +111,68 @@ describe('core/server/base', () => {
       );
 
       const response = await testClient({ stubRoutes })
-        .get('/en-US/firefox/simulation-of-a-non-existent-page').end();
+        .get('/en-US/firefox/simulation-of-a-non-existent-page')
+        .end();
```
kumar303: I had an idea for making this better. What do you think of this?

```js
const response = await testClient({ stubRoutes })
  .get('/en-US/firefox/simulation-of-a-non-existent-page');
shudownServer(response);
```

`shudownServer()` would just call `response.end()` but it would make it really clear what's happening. Also, we can link to https://github.com/visionmedia/supertest/issues/100 in a comment.

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +111,68 @@ describe('core/server/base', () => {
...
+    });
+
+    it('dispatches setAuthToken() if cookie is present', async () => {
+      const token = userAuthToken();
+      const { store, sagaMiddleware } = createStoreAndSagas({
+        reducers: { reduxAsyncConnect, api: apiReducer },
+      });
+      const response = await testClient({ store, sagaMiddleware })
+        .get('/en-US/firefox/')
+        .set('cookie', `${config.get('cookieName')}="${token}"`)
```
kumar303: there's no `.cookie()` method?

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +111,68 @@ describe('core/server/base', () => {
...
+      expect(response.statusCode).toEqual(200);
+      expect(api.token).toEqual(token);
+    });
+
+    it('runs sagas as early as possible', async () => {
+      const mockResponse = () => createApiResponse({
+        jsonData: createUserProfileResponse({ id: 42, username: 'babar' }),
+      });
+
+      mockWindow.expects('fetch')
```
kumar303: I think it would be better just to check `mockUserApi.expects('userProfile')`

tests/unit/core/server/testServer.js
```diff
@@ -23,32 +36,47 @@ describe('core/server/base', () => {
...
+      return { store, sagaMiddleware };
     }
 
     // eslint-disable-next-line no-empty-function
     function* fakeSaga() {}
 
-    const app = baseServer(stubRoutes, createStoreAndSagas, {
-      appSagas: fakeSaga,
+    const app = baseServer(stubRoutes, _createStoreAndSagas, {
+      appSagas: appSagas || fakeSaga,
```
kumar303: Can you also try to add test coverage for the `disableSSR=true` case? I think you could do it with this patch:

```js
diff --git a/src/core/server/base.js b/src/core/server/base.js
index e2f727a9..5927a744 100644
--- a/src/core/server/base.js
+++ b/src/core/server/base.js
@@ -3,7 +3,7 @@ import path from 'path';
 
 import 'babel-polyfill';
-import config from 'config';
+import defaultConfig from 'config';
 import Express from 'express';
 import helmet from 'helmet';
 import Raven from 'raven';
@@ -116,7 +116,10 @@ function hydrateOnClient({ res, props = {}, pageProps }) {
     .end();
 }
 
-function baseServer(routes, createStore, { appSagas, appInstanceName = appName } = {}) {
+function baseServer(routes, createStore, {
+  appSagas, appInstanceName = appName,
+  config = defaultConfig,
+} = {}) {
   const app = new Express();
   app.disable('x-powered-by');
 
```

tests/unit/core/reducers/test_user.js
```diff
@@ -0,0 +1,51 @@
...
+    it('initializes properly', () => {
+      const { id, username } = reducer(undefined);
+      expect(id).toEqual(null);
+      expect(username).toEqual(null);
+    });
+
+    it('ignores unrelated actions', () => {
+      const state = {
+        ...initialState,
+        id: 12345,
```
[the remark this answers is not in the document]
willdurand: But how can I be sure it does not change the values? The `initialState` has `null` values, which I think are not the best values for this test case.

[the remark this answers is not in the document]
willdurand: I used an action creator to have a non-null state.

tests/unit/core/sagas/test_user.js
```diff
@@ -0,0 +1,63 @@
...
+
+
+describe(__filename, () => {
+  let mockApi;
+  let sagaTester;
+  let rootTask;
+
+  beforeEach(() => {
+    mockApi = sinon.mock(api);
+    const initialState = dispatchSignInActions().state;
```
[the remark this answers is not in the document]
willdurand: of course!

tests/unit/core/server/testServer.js
willdurand: I did not find anything. This PR seems to confirm this: https://github.com/visionmedia/supertest/pull/410.

willdurand: I found a `cookie` method on the response only...

tests/unit/core/server/testServer.js
willdurand: This works and I see the coverage in the report, but I haven't found a good assertion yet.

tests/unit/core/server/testServer.js
tofumatt: I dig it.

willdurand: mmh this is not as easy as what I thought since `await` resolves the promise. I do not get a `supertest` instance but a response (with `req` and `res`) that is the same with or without `end()`. In other words, I cannot call `.end()` on `response`.

willdurand: A solution could be to transform the `testClient()` function into something like `createApp()`, then have a call to `supertest(app)` to get the test client, and later shutdown the server:

```
const app = createApp(...);
const client = supertest(app);

// ...

app.close();
```

WDYT? Slighty more verbose, but explicit.

willdurand:
I did not change the `.end()` since I dont know how to do it properly. We could maybe `listen()` to the `app` and therefore have full control over the server.. WDYT?

kumar303:
> I did not change the .end() since I dont know how to do it properly.

Aw, whoops, I see what you mean about `await`. Thanks for giving it a try. We'll just have to remember to call `.end()` for now.

src/core/reducers/api.js
```diff
@@ -48,24 +46,6 @@ export const initialApiState: ApiStateType = {
   userId: null,
```
kumar303: `userId` should be removed from `initialApiState`. I think this will break some things. Those things will need to use `state.user.id` instead.

willdurand: doh, bad rebase it seems because I am pretty sure I removed it :x

willdurand:
> There are a few problems:
> 
> First, the token is getting set but not the user object. Maybe the user fetching saga isn't running?

wow that's pretty bad. I think I have been tricked by my login state... I will double check and fix that.

> Secondly, some components are checking api.token and some are checking user.id to see if the user is logged in. All components should check user.id (I think?). We need to pick one.

heh yeah. The only thing I have done here is getting rid of `userId` in the `api` reducer (well, except that I introduced it again during a bad rebase...). I think #1644 mentions the use of `state.token` to determine whether a user is authenticated. I do agree with you we should pick one and I'd like to choose the `user.id`. Yet, I did not do it here because I did not want to increase the diff size.

Should I do the change here or open a new issue and send a new patch for the `user.id` vs `state.token`?

willdurand:
> I noticed a problem with disableSSR: true. Set that in your local config and restart your server. Go to an extension detail page and log in. Scroll down a bit. It should show you star ratings to select from but it does not:

mmh, I cannot reproduce this actually.

willdurand:
> wow that's pretty bad. I think I have been tricked by my login state... I will double check and fix that.

pff, I am silly I guess. The test case for that passes by chance. I was not 100% confident with the test case for that, because I did not find a good assertion 😪

willdurand:
Ok, I was able to reproduce the issue in the test suite. I fixed it and made sure we had assertions for it. I use [`cheerio`](https://cheerio.js.org/) because it is a Enzyme dependency.

src/core/reducers/user.js
```diff
@@ -1,14 +1,21 @@
+/* @flow */
 import { LOG_OUT_USER } from 'core/constants';
 
 
 const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
 
-export const initialState = {
+type Action = Object;
```
willdurand: TODO: have a better `Action` type. I find it difficult and a bit superfluous to maintain this type in general.

willdurand:
Added a last commit with `state.token` changes.

tests/unit/core/server/testServer.js
```diff
@@ -23,33 +33,50 @@ describe('core/server/base', () => {
...
-        store: createStore(
-          combineReducers({ reduxAsyncConnect }),
-          // Do not define an initial state.
-          undefined,
-          applyMiddleware(sagaMiddleware),
-        ),
-        sagaMiddleware,
-      };
+  function createStoreAndSagas({
+    reducers = { reduxAsyncConnect },
```
kumar303: I think this should default to:

```js
{ reduxAsyncConnect, api: apiReducer, user: userReducer }
```

because really the server requires all of those to operate. I don't see a strong case for customizing the reducer list in any of the tests below.

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,112 @@ describe('core/server/base', () => {
...
+        .set('cookie', `${defaultConfig.get('cookieName')}="${token}"`)
+        .end();
+
+      const { api } = store.getState();
+
+      expect(response.statusCode).toEqual(200);
+      expect(api.token).toEqual(token);
+    });
+
+    it('runs sagas as early as possible', async () => {
```
kumar303: The "early" part is an implementation detail. I think this spec could be "it fetches the user profile when given a token"

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,112 @@ describe('core/server/base', () => {
...
+
+      const { api, user } = store.getState();
+
+      expect(response.statusCode).toEqual(200);
+      expect(api.token).toEqual(token);
+      expect(user.id).toEqual(42);
+      expect(user.username).toEqual('babar');
+      mockUserApi.verify();
+
+      // Parse the HTML response to retrieve the serialized redux state.
```
kumar303: This is a great test! However, I think it should just be a standalone test. The spec could be "it serializes the redux state in html"

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,112 @@ describe('core/server/base', () => {
...
+        .returns(Promise.resolve(profile));
+
+      const token = userAuthToken();
+      const { store, sagaMiddleware } = createStoreAndSagas({
+        reducers: { reduxAsyncConnect, api: apiReducer, user: userReducer },
+      });
+      // We use `cloneDeep()` to allow modifications on the `config` object,
+      // since a call to `get()` makes it immutable. This is the case in the
+      // previous test cases (on `defaultConfig`).
+      const config = configUtil.cloneDeep(defaultConfig);
```
kumar303: Interesting approach. I like it because it makes the config object realistic. It may be a little heavyweight since config objects are big but I'm not worried for this patch.

We actually have a few instances of config mocking that might benefit from this approach instead.

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,112 @@ describe('core/server/base', () => {
...
+      const { api, user } = store.getState();
+
+      expect(response.statusCode).toEqual(200);
+      expect(api.token).toEqual(token);
+      expect(user.id).toEqual(42);
+      expect(user.username).toEqual('babar');
+      mockUserApi.verify();
+    });
+
+    it('sends the redux-saga END action when `disableSSR` is `true`', async () => {
```
kumar303: This spec should be more related to the actual feature. I suggest this: "it fetches the user profile even when SSR is disabled"

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,112 @@ describe('core/server/base', () => {
...
+      expect(api.token).toEqual(token);
+    });
+
+    it('runs sagas as early as possible', async () => {
+      const profile = createUserProfileResponse({ id: 42, username: 'babar' });
+
+      mockUserApi
+        .expects('userProfile')
+        .once()
+        .returns(Promise.resolve(profile));
```
kumar303: I would like to see another test that does `.returns(Promise.reject(new Error('example of an API error')))` and asserts a 500, like `expect(response.statusCode).toEqual(500);`

tests/unit/helpers.js
```diff
@@ -276,3 +276,23 @@ export function createApiResponse({
...
+    homepage: '',
+    id,
+    is_addon_developer: false,
+    is_artist: false,
+    location: '',
+    name: '',
+    num_addons_listed: 0,
+    occupation: '',
+    picture_type: '',
+    picture_url: 'https://addons-dev-cdn.allizom.org/static/img/zamboni/anon_user.png',
```
kumar303: maybe this can use the `amoCDN` config property too?

src/core/api/user.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+import { callApi } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+
+
+export function userProfile({ api }: {| api: ApiStateType |}) {
```
kumar303: It would be pretty confusing if the caller forgot to pass a value for `api` so I think this should throw an error if it's empty.

tests/unit/amo/helpers.js
```diff
@@ -108,11 +110,15 @@ export function dispatchClientMetadata({
 
 export function dispatchSignInActions({
   authToken = userAuthToken(),
+  userId = 12345,
   ...otherArgs
 } = {}) {
   const { store } = dispatchClientMetadata(otherArgs);
 
   store.dispatch(setAuthToken(authToken));
+  store.dispatch(userProfileLoaded({
```
kumar303: oh yeah, good idea

src/core/containers/LoginRequired/index.js
```diff
@@ -7,7 +7,7 @@ import LoginPage from 'core/components/LoginPage';
 
 export function mapStateToProps(state) {
   return {
-    authenticated: !!state.api.token,
+    authenticated: !!state.user.id,
```
kumar303: It looks like you didn't have to update any tests for this which is a red flag. Can you try to add test coverage for it?

src/core/server/base.js
```diff
@@ -36,25 +36,25 @@ import I18nProvider from 'core/i18n/Provider';
...
-const isDevelopment = config.get('isDevelopment');
+const version = path.join(defaultConfig.get('basePath'), 'version.json');
+const isDeployed = defaultConfig.get('isDeployed');
+const isDevelopment = defaultConfig.get('isDevelopment');
 
 
-const appName = config.get('appName');
+const appName = defaultConfig.get('appName');
 
 export function getPageProps({ noScriptStyles = '', store, req, res }) {
```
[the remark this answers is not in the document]
willdurand: Yeah, that definitely makes sense.

src/core/reducers/user.js
```diff
@@ -0,0 +1,47 @@
...
+const USER_PROFILE_LOADED = 'USER_PROFILE_LOADED';
+
+type Action = Object;
+export type UserStateType = {
+  id: ?number,
+  username: ?string,
+};
+
+export const initialState: UserStateType = {
+  id: null,
```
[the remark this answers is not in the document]
willdurand: All have been replaced. `api.token` is used for the API calls, but not for the authentication status checks anymore. I think having a new prop makes sense. Another option would be a selector:

```
// in core/reducers/user
const isAuthenticated = (state) => {
    return !!state.user.id;
};
```

The benefit is to not have to update the code if we want to change internal state things, and the state does not get bigger.

tests/unit/core/server/testServer.js
willdurand: great, it simplifies the test cases.

tests/unit/core/server/testServer.js
willdurand: I was thinking about adding a helper function but could not find a similar snippet.

src/core/api/user.js
willdurand: can we open an issue for that? this should be done across all `core/api/*` functions.

willdurand: can we open an issue for that? this should be done across all `core/api/*` functions.

src/core/containers/LoginRequired/index.js
willdurand: It is actually covered by the change in `dispatchSignInActions()` 😉

src/core/api/user.js
kumar303: I don't think it should be done for all API functions. In this case, it's only necessary to require `api` because you have to be authenticated to make a request to the profile endpoint.

willdurand: ok, adding the check here then

src/core/containers/LoginRequired/index.js
kumar303: oh I see, that's cool. The helper is doing its job well.

You could edit the specs in `TestLoginRequired` but it's not a big deal. I would propose the following changes:

* old: it sets authenticated to true when there is a token
  * new: it sets authenticated to true
* old: it sets authenticated to false when there is not a token
  * new: it sets authenticated to false

src/core/reducers/user.js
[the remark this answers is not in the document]
kumar303: Sure. A selector sounds good.

willdurand:
Ok I think this is ready. I had to pull your patch from #3023 because test suite was failing on Travis and I wanted to make sure it was not because of my changes. I'll get rid of this commit before merging this PR (or rebase when #3023 gets merged).

willdurand:
Ok I think this is ready.

@kumar303 I had to pull your patch from #3023 because test suite was failing on Travis and I wanted to make sure it was not because of my changes. I'll get rid of this commit before merging this PR (or rebase when #3023 gets merged).

willdurand:
I have updated the code, and also changed the action type/creator I introduced to match the naming pattern as per https://github.com/mozilla/addons-frontend/pull/3025#discussion_r135909932.

src/core/reducers/user.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+};
+
+export const isAuthenticated = (state: { user: UserStateType }) => {
+  return !!state.user.id;
+};
+
+export default function reducer(
+  state: UserStateType = initialState,
+  action: Action = {}
```
kumar303: for future reference, this would look more like:

```js
action: LoadUserProfileAction | LogOutUserAction
```

...but we'd have to switch away from constants to strings so that Flow [can figure out which action is which](https://flow.org/en/docs/react/redux/#toc-typing-redux-reducers) in the switch.

src/core/server/base.js
```diff
@@ -333,27 +339,29 @@ function baseServer(routes, createStore, { appSagas, appInstanceName = appName }
...
     log.error({ err: error }); // log the stack trace too.
-    return showErrorPage({ createStore, error, status: 500, req, res });
+    return showErrorPage({ createStore, error, status: 500, req, res, config });
   });
 
   return app;
 }
 
 export function runServer({
-  listen = true, app = appName, exitProcess = true,
```
kumar303: Hmm, I can't see where any callers were using this `app` parameter. I guess it's safe to remove.

tests/unit/core/components/TestAuthenticateButton.js
```diff
@@ -65,31 +70,32 @@ describe('<AuthenticateButton />', () => {
...
+    const { store } = dispatchSignInActions();
     const _window = { location: '/foo' };
     const location = { pathname: '/bar', query: { q: 'wat' } };
-    const startLoginUrlStub =
-      sinon.stub(api, 'startLoginUrl').returns('https://a.m.org/login');
-    const { handleLogIn } = mapStateToProps({
-      auth: {},
-      api: { lang: 'en-GB' },
-    });
+    const startLoginUrlStub = sinon.stub(api, 'startLoginUrl').returns('https://a.m.org/login');
```
kumar303: just FYI, the reason our lines tend to be under 80 chars is because I like to use three splits in Vim. I haven't been able to convince people that our 100 line limit is too long for that :)

tests/unit/core/containers/TestLoginRequired.js
```diff
@@ -41,15 +42,13 @@ describe('<LoginRequired />', () => {
...
 
-    it('sets authenticated to true when there is a token', () => {
-      expect(mapStateToProps(store.getState()))
-        .toEqual({ authenticated: true });
+    it('sets authenticated to true', () => {
+      expect(mapStateToProps(store.getState())).toEqual({ authenticated: true });
     });
 
-    it('sets authenticated to false when there is not a token', () => {
+    it('it sets authenticated to false', () => {
```
kumar303: technically the `it` isn't needed but Jest doesn't prefix our specs with "it" like mocha did so they aren't readable anyway, heh.

tests/unit/core/sagas/test_user.js
```diff
@@ -0,0 +1,66 @@
...
+      .expects('userProfile')
+      .returns(Promise.reject(error));
+
+    sagaTester.dispatch(setAuthToken(userAuthToken()));
+
+    return rootTask.done
+      .then(() => {
+        throw new Error('unexpected success');
+      })
+      .catch((expectedError) => {
```
kumar303: We tend to only use *expected* for variables set by the tests, not variables returned by the program under test. No need to bother with changing it, just though I'd mention it.

tests/unit/core/server/testServer.js
```diff
@@ -83,9 +110,151 @@ describe('core/server/base', () => {
...
+
+      const token = userAuthToken();
+      const { store, sagaMiddleware } = createStoreAndSagas();
+      // We use `cloneDeep()` to allow modifications on the `config` object,
+      // since a call to `get()` makes it immutable. This is the case in the
+      // previous test cases (on `defaultConfig`).
+      const config = configUtil.cloneDeep(defaultConfig);
+      config.disableSSR = true;
```
kumar303: I don't think it's so important to test this with `disableSSR=true`. It would be better not to just because it will simulate production better.

willdurand: I see what you mean, and I will add another similar test for that.

I'd like to keep this one though as it actually covers the fact that with `disableSSR=true`, sagas are executed correctly and both user profile and token are passed to the client side. The bug you mentioned earlier this week was because of that.

willdurand:
@tofumatt I am going to fix the last @kumar303's comments before merging it. Would you have something to raise or could you unlock your review status?

tests/unit/core/components/TestAuthenticateButton.js
willdurand: ah I only have two splits in Vim 😅 

I think that 80 chars is not enough, but I am sold! Let's go for 80 chars.

tests/unit/core/containers/TestLoginRequired.js
willdurand: oh good catch, that was actually a mistake 😞

tests/unit/core/server/testServer.js
willdurand: The state computed on the server is still passed to the client _via_ the HTML. `SET_AUTH_TOKEN` is not often called on the client because it is already done by the server. That is also why I put these assertions in the `disableSSR=true` test case. That is the HTML I get with `disableSSR=true`:

``` html
<!DOCTYPE html>
<html lang="en-US" dir="ltr" data-reactroot="" data-reactid="1" data-react-checksum="1242348688">
 <head data-reactid="2">
  <meta charset="utf-8" data-reactid="3" />
  <meta name="viewport" content="width=device-width, initial-scale=1" data-reactid="4" />
  <link rel="shortcut icon" href="/favicon.ico" data-reactid="5" />
  <title data-react-helmet="true" data-reactid="6"></title>
  <noscript data-reactid="7">
   <style data-reactid="8">/* !important is needed to show the button, make sure switch is hidden */
.InstallButton .InstallButton-button {
  display: inline-block !important;
}

.InstallButton .InstallButton-switch {
  display: none !important;
}
</style>
  </noscript>
 </head>
 <body data-reactid="9">
  <div id="react-view" data-reactid="10"></div>
  <script type="application/json" id="redux-store-state" data-reactid="11">{"addons":{},"api":{"clientApp":"firefox","lang":"en-US","token":"eyJhdXRoX2hhc2giOiJiNTRhNTBiYjYzMDg0OTVkMzQ3OTU5MGIyMDg4MmEyZDkwZWM3ZmJjIiwidXNlcl9pZCI6MTA2MzkxNTZ9:1dmdwc:DHZayyST3Y2Ms6SE7memWNz5l7Y","userAgent":"Mozilla\u002F5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit\u002F537.36 (KHTML, like Gecko) Chrome\u002F59.0.3071.115 Safari\u002F537.36","userAgentInfo":{"browser":{"name":"Chrome","version":"59.0.3071.115","major":"59"},"os":{"name":"Mac OS","version":"10.12.6"}}},"autocomplete":{"loading":false,"suggestions":[]},"categories":{"categories":null,"loading":false},"errors":{},"errorPage":{"clearOnNext":false,"error":null,"hasError":false,"statusCode":null},"featured":{"addonType":null,"loading":false,"results":[]},"infoDialog":{},"installations":{},"landing":{"addonType":null,"featured":{"count":0,"results":[]},"highlyRated":{"count":0,"results":[]},"loading":false,"popular":{"count":0,"results":[]},"resultsLoaded":false},"loadingBar":0,"reduxAsyncConnect":{"loaded":false,"loadState":{}},"reviews":{"byAddon":{}},"search":{"count":0,"filters":{},"loading":false,"results":[]},"user":{"id":10639156,"username":"willdurand"},"viewContext":{"context":"VIEW_CONTEXT_EXPLORE"}}</script>
  <script src="//127.0.0.1:3001/amo-c623c6e9e4951a016e4b.js" data-reactid="12"></script>
 </body>
</html>
```

I have updated both test cases now. The `disableSSR=true` one checks that the state is correct (so that login works with such a configuration). The second test case about making sure the state is serialized into HTML now simulates the production configuration.

kumar303: > I'd like to keep this one though as it actually covers the fact that with disableSSR=true, sagas are executed correctly and both user profile and token are passed to the client side.

Adding both test cases is fine but I didn't explain myself very well in my suggestion to only include a test for the `disableSSR: false` case (to emulate production). Here is some more context.

The `core/server/testServer.js` suite performs high level integration tests--that is, at the server->client level. These types of tests are good but they are also dangerous because they tempt you to test for a matrix of possibilities and that matrix can be vast. The more test cases you add, the slower the suite becomes.

I think the best approach for an app like ours is to rely on unit tests as much as possible and then try to use static analysis (like Flow) to make sure our units are glued together right. We may not ever get to that last part because I haven't been able to convince the team of how Flow can accomplish this for us.

So, we should still do integration level tests but we should use them sparingly. For this case, we actually [already have a unit test](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/core/containers/TestServerHtml.js#L101) for rendering the redux state to the client. We were lacking an integration test for it so that's why I requested adding one (it could definitely catch a bug or two) but I don't think we need to try and cover every edge case.

willdurand: Oh, but you are absolutely right! Thanks for the explanations.

Issue #2992, opened 2017-08-23
Document how to install missing fonts for Sketch


kumar303:
When developers set up Sketch for viewing the UX mocks they need to install some fonts

willdurand:
Can we archive the mobile Sketch file? It seems outdated now, and I have relied on the wrong one at the beginning 😞

tofumatt:
I think that’s fine; feel free to move them in a PR. I don’t think they’ve been updated in ages...

PR #2993, opened 2017-08-23
Document how to install fonts for Sketch

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2992

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,13 +347,7 @@ be sure to clear your cookies.
 
 ### Working with UX Mocks
 
-When implementing user interfaces you will need to refer to the
-[Sketch](https://www.sketchapp.com/) mocks that are located in the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -345,6 +345,16 @@ be sure to clear your cookies.
 
 **NOTE**: At this time, it's not possible to sign in using this approach.
 
+### Working with UX Mocks
+
+When implementing user interfaces you will need to refer to the
```
tofumatt: Minor note: I don't think we need to wrap README prose because GiHub will give us nicer text diffs if we don't and also it's kind of a pain. But do whatever you prefer.

willdurand: I think it is also good for editors or our own `git diff`.

kumar303: Sure, I can do that. I didn't have soft wrapping enabled in vim but I just fixed that.

Issue #2745, opened 2017-07-11
Show tools menus/links for signed in, authenticated users


tofumatt:
If a user is signed in to AMO and has any relevant permissions for tools, we should display a tools menu.

Note: this doesn't seem to be in the UX mocks so this needs UX. It might simply fit well in the user menu from #2744.

Tools menus are:

* Reviewer Tools
* Admin Tools
* Localizer Tools

Existing menu:

<img width="333" alt="screenshot 2017-07-12 00 46 21" src="https://user-images.githubusercontent.com/90871/28095714-de7b06e6-669b-11e7-82d3-b079064d22d5.png">

willdurand:
Looks like it is now part of the mockups, see: https://github.com/mozilla/addons-frontend/issues/2744#issuecomment-323378139.

vcarciu:
The specs have changed a lot here, but we can keep this open until implemented. New specs : 
https://user-images.githubusercontent.com/6797635/29692108-37b5b102-88e3-11e7-9ae5-a11f9a6358fa.png

willdurand:
#2994 has added the menu. I keep this issue open as we need to add more links now.

willdurand:
@tofumatt it seems we does not deal with permissions in the app yet. We should start by doing this before being able to add the following links to the menu:

- Reviewer Tools
- Admin Tools
- Localizer Tools

tofumatt:
@willdurand They are sent during the login API step (http://addons-server.readthedocs.io/en/latest/topics/api/auth_internal.html#get--api-v3-accounts-login-)... which I'm just realising could cause issues if permissions are updated during site usage, but maybe we can still get them without logging in again–I forget how often we re-check the auth endpoint for new info about the existing user.

kumar303:
I was late to the IRC conversation but I don't think we need to block this on getting any API changes. We can only get permissions on login currently but that will be fine 90% of the time.

willdurand:
@kumar303 yes but we have not way to get it on the server right now. We could add more information to the cookie though. However, this is not nice for UX if you have to logout/login to get access to something you did not have access to before.

kumar303:
Yes, adding permissions to the user profile endpoint would be ideal. I was just saying doing a logout/login after someone grants you admin privileges is something we could live with if needed.

willdurand:
Ah ok. I have requested the API change at: https://github.com/mozilla/addons-server/issues/6415. I am going to work on the cookie-based version.

[...]

PR #2994, opened 2017-08-23
Add dropdown menu and display username (when logged in)

### LINKED ISSUE #1751: Log Out button should have user's name on it

willdurand:
Fix #1751 
Refs #2745
~~⚠️  depends on #2989~~

---

This PR introduces a new `DropdownMenu` component according to the recent mock-ups. It displays the username if she is logged in, and a simple menu linking to account view/edit and providing the "log out" action. This PR also fixes warnings related to the `Header` and `Link` components.

## Screenshots

screen shot 2017-09-08 at 13 34 30
screen shot 2017-09-08 at 13 34 28
screen shot 2017-09-08 at 13 34 26

capture d ecran - 2017-09-08 a 13 34 48

capture d ecran - 2017-09-08 a 13 34 53

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -89,15 +89,15 @@ export class HeaderBase extends React.Component {
               text={username}
               className="Header-authenticate-button Header-button"
             >
-              <DropdownMenuItem>{i18n.gettext('Account')}</DropdownMenuItem>
+              <DropdownMenuItem>{i18n.gettext('My Account')}</DropdownMenuItem>
               <DropdownMenuItem>
                 <Link href={`/user/${username}/`}>
-                  {i18n.gettext('My profile')}
+                  {i18n.gettext('View Profile')}
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
                 <Link href="/users/edit">
-                  {i18n.gettext('Account settings')}
+                  {i18n.gettext('Edit Profile')}
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem onClick={this.handleLogOut} detached>
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -6,28 +6,51 @@ import DownloadFirefoxButton from 'amo/components/DownloadFirefoxButton';
 import Link from 'amo/components/Link';
 import SearchForm from 'amo/components/SearchForm';
 import SectionLinks from 'amo/components/SectionLinks';
-import AuthenticateButton from 'core/components/AuthenticateButton';
+import AuthenticateButton, {
+  createHandleLogOutFunction,
+} from 'core/components/AuthenticateButton';
+import { isAuthenticated } from 'core/reducers/user';
 import { VIEW_CONTEXT_HOME } from 'core/constants';
 import translate from 'core/i18n/translate';
 import Icon from 'ui/components/Icon';
+import DropdownMenu from 'ui/components/DropdownMenu';
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
 
 import './styles.scss';
 
 
 export class HeaderBase extends React.Component {
   static propTypes = {
+    api: PropTypes.object.isRequired,
+    handleLogOut: PropTypes.func.isRequired,
     i18n: PropTypes.object.isRequired,
+    isAuthenticated: PropTypes.bool.isRequired,
     isHomePage: PropTypes.bool.isRequired,
     location: PropTypes.object.isRequired,
-    query: PropTypes.string,
+    query: PropTypes.string.isRequired,
+    username: PropTypes.string,
   }
 
-  static defaultPropTypes = {
+  static defaultProps = {
     isHomePage: false,
+    query: '',
+  }
+
+  handleLogOut = (event) => {
+    event.preventDefault();
+
+    this.props.handleLogOut({ api: this.props.api });
   }
 
   render() {
-    const { i18n, isHomePage, location, query } = this.props;
+    const {
+      i18n,
+      isHomePage,
+      location,
+      query,
+      username,
+    } = this.props;
+
     const headerLink = (
       <Link className="Header-title" to="/">
         <Icon className="Header-addons-icon" name="fox" />
@@ -60,12 +83,35 @@ export class HeaderBase extends React.Component {
           <DownloadFirefoxButton
             className="Header-download-button Header-button"
           />
-          <AuthenticateButton
-            className="Header-authenticate-button Header-button Button--action
+
+          {this.props.isAuthenticated ? (
+            <DropdownMenu
+              text={username}
+              className="Header-authenticate-button Header-button"
+            >
+              <DropdownMenuItem>{i18n.gettext('Account')}</DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link href={`/user/${username}/`}>
+                  {i18n.gettext('My profile')}
+                </Link>
+              </DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link href="/users/edit">
+                  {i18n.gettext('Account settings')}
+                </Link>
+              </DropdownMenuItem>
+              <DropdownMenuItem onClick={this.handleLogOut} detached>
+                {i18n.gettext('Log out')}
+              </DropdownMenuItem>
+            </DropdownMenu>
+          ) : (
+            <AuthenticateButton
+              className="Header-authenticate-button Header-button Button--action
               Button--outline-only Button--small"
-            location={location}
-            noIcon
-          />
+              location={location}
+              noIcon
+            />
+          )}
         </div>
 
         <SearchForm
@@ -78,11 +124,20 @@ export class HeaderBase extends React.Component {
   }
 }
 
-export function mapStateToProps(state) {
-  return { isHomePage: state.viewContext.context === VIEW_CONTEXT_HOME };
-}
+export const mapStateToProps = (state) => {
+  return {
+    api: state.api,
+    isHomePage: state.viewContext.context === VIEW_CONTEXT_HOME,
+    username: state.user.username,
+    isAuthenticated: isAuthenticated(state),
+  };
+};
+
+export const mapDispatchToProps = (dispatch) => ({
+  handleLogOut: createHandleLogOutFunction(dispatch),
+});
 
 export default compose(
-  connect(mapStateToProps),
+  connect(mapStateToProps, mapDispatchToProps),
   translate(),
 )(HeaderBase);
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -71,7 +71,7 @@ export class AuthenticateButtonBase extends React.Component {
     // https://github.com/mozilla/addons-frontend/issues/1904
     return (
       <Button
-        href={`#log-${isAuthenticated ? 'out' : 'in'}`}
+        href={`#${isAuthenticated ? 'logout' : 'login'}`}
         className={className}
         onClick={this.onClick}
       >
--- a/src/core/components/AuthenticateButton/index.js
+++ b/src/core/components/AuthenticateButton/index.js
@@ -21,13 +21,13 @@ type HandleLogInFunc = (
   location: ReactRouterLocation, options?: {| _window: typeof window |}
 ) => void;
 
-type HandleLogOutFunc = ({| api: ApiStateType |}) => Promise<void>;
+type HandleLogOutFunction = ({| api: ApiStateType |}) => Promise<void>;
 
 type AuthenticateButtonProps = {|
   api: ApiStateType,
   className?: string,
   handleLogIn: HandleLogInFunc,
-  handleLogOut: HandleLogOutFunc,
+  handleLogOut: HandleLogOutFunction,
   i18n: Object,
   isAuthenticated: boolean,
   location: ReactRouterLocation,
@@ -58,7 +58,7 @@ export class AuthenticateButtonBase extends React.Component {
 
   render() {
     const {
-      i18n, isAuthenticated, logInText, logOutText, noIcon, ...otherProps
+      i18n, isAuthenticated, logInText, logOutText, noIcon, className,
     } = this.props;
     const buttonText = isAuthenticated ?
       logOutText || i18n.gettext('Log out') :
@@ -70,7 +70,11 @@ export class AuthenticateButtonBase extends React.Component {
     // mobile browser. This is the cause of
     // https://github.com/mozilla/addons-frontend/issues/1904
     return (
-      <Button onClick={this.onClick} href="#" {...otherProps}>
+      <Button
+        href={`#log-${isAuthenticated ? 'out' : 'in'}`}
+        className={className}
+        onClick={this.onClick}
+      >
         {noIcon ? null : <Icon name="user-dark" />}
         {buttonText}
       </Button>
@@ -99,16 +103,21 @@ export const mapStateToProps = (
 });
 
 type DispatchMappedProps = {|
-  handleLogOut: HandleLogOutFunc,
+  handleLogOut: HandleLogOutFunction,
 |};
+
+export const createHandleLogOutFunction = (
+  dispatch: DispatchFunc
+): HandleLogOutFunction => {
+  return ({ api }) => {
+    return logOutFromServer({ api }).then(() => dispatch(logOutUser()));
+  };
+};
 
 export const mapDispatchToProps = (
   dispatch: DispatchFunc
 ): DispatchMappedProps => ({
-  handleLogOut({ api }) {
-    return logOutFromServer({ api })
-      .then(() => dispatch(logOutUser()));
-  },
+  handleLogOut: createHandleLogOutFunction(dispatch),
 });
 
 export default compose(
--- a/src/ui/components/DropdownMenu/index.js
+++ b/src/ui/components/DropdownMenu/index.js
@@ -1,0 +1,67 @@
+/* @flow */
+import React from 'react';
+import classNames from 'classnames';
+import onClickOutside from 'react-onclickoutside';
+
+import Icon from 'ui/components/Icon';
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
+
+import './styles.scss';
+
+
+type Props = {|
+  text: string,
+  children: Array<DropdownMenuItem>,
+  className?: string,
+|};
+
+export class DropdownMenuBase extends React.Component {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = { buttonIsActive: false };
+  }
+
+  state: {| buttonIsActive: bool |};
+  props: Props;
+
+  handleOnClick = (event: SyntheticEvent) => {
+    event.preventDefault();
+
+    this.setState((previousState) => ({
+      buttonIsActive: !previousState.buttonIsActive,
+    }));
+  }
+
+  handleClickOutside = () => {
+    this.setState({ buttonIsActive: false });
+  }
+
+  render() {
+    const { children, className, text } = this.props;
+
+    return (
+      <div
+        className={classNames('DropdownMenu', className, {
+          'DropdownMenu--active': this.state.buttonIsActive,
+        })}
+      >
+        <button
+          className="DropdownMenu-text"
+          onClick={this.handleOnClick}
+        >
+          {text}
+          <Icon name="inverted-caret" />
+        </button>
+
+        {children && (
+          <ul className="DropdownMenu-items">
+            {children}
+          </ul>
+        )}
+      </div>
+    );
+  }
+}
+
+export default onClickOutside(DropdownMenuBase);
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -9,6 +9,22 @@
   padding-bottom: 6px;
   padding-top: 6px;
   position: relative;
+
+  // This is needed because `:hover` on mobile prevents the main button to
+  // close the dropdown menu. On mobile, `.DropdownMenu--active` controls the
+  // visibility of the menu. On "large" devices (and above), `:hover` also
+  // opens the menu.
+  @include respond-to(large) {
+    &:hover {
+      .DropdownMenu-items {
+        display: block;
+      }
+
+      .Icon-inverted-caret {
+        opacity: 1;
+      }
+    }
+  }
 }
 
 .DropdownMenu-text {
@@ -26,30 +42,8 @@
     opacity: 0.6;
     transition: opacity $transition-short ease-in-out;
     width: 9px;
-  }
-}
 
-.DropdownMenu--active {
-  .DropdownMenu-items {
-    display: block;
-  }
-
-  .Icon-inverted-caret {
-    opacity: 1;
-  }
-}
-
-// This is needed because `:hover` on mobile prevents the main button to close
-// the dropdown menu. On mobile, `.DropdownMenu--active` controls the
-// visibility of the menu. On "large" devices (and above), `:hover` also opens
-// the menu.
-@include respond-to(large) {
-  .DropdownMenu:hover {
-    .DropdownMenu-items {
-      display: block;
-    }
-
-    .Icon-inverted-caret {
+    .DropdownMenu--active & {
       opacity: 1;
     }
   }
@@ -84,5 +78,9 @@
     top: -4px;
     transform: rotate(45deg);
     width: 8px;
+  }
+
+  .DropdownMenu--active & {
+    display: block;
   }
 }
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -7,7 +7,7 @@
 
   display: inline-block;
   padding-bottom: 6px;
-  padding-top: 6px;
+  padding-top: 5px;
   position: relative;
 
   // This is needed because `:hover` on mobile prevents the main button to
@@ -38,7 +38,7 @@
   .Icon-inverted-caret {
     @include margin-start(5px);
 
-    height: 10px;
+    height: 8px;
     opacity: 0.6;
     transition: opacity $transition-short ease-in-out;
     width: 9px;
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -1,0 +1,88 @@
+@import "~core/css/inc/mixins";
+@import "~amo/css/inc/vars";
+@import "~ui/css/vars";
+
+.DropdownMenu {
+  @include text-align-start();
+
+  display: inline-block;
+  padding-bottom: 6px;
+  padding-top: 6px;
+  position: relative;
+}
+
+.DropdownMenu-text {
+  @include font-regular();
+
+  color: $white;
+  cursor: pointer;
+  font-size: $font-size-s;
+  font-weight: normal;
+  text-decoration: none;
+
+  .Icon-inverted-caret {
+    @include margin-start(5px);
+
+    height: 10px;
+    opacity: 0.6;
+    width: 9px;
+  }
+}
+
+.DropdownMenu--active {
+  .DropdownMenu-items {
+    display: block;
+  }
+
+  .Icon-inverted-caret {
+    opacity: 1;
+  }
+}
+
+// This is needed because `:hover` on mobile prevents the main button to close
+// the dropdown menu. On mobile, `.DropdownMenu--active` controls the
+// visibility of the menu. On "large" devices (and above), `:hover` also opens
+// the menu.
+@include respond-to(large) {
+  .DropdownMenu:hover {
+    .DropdownMenu-items {
+      display: block;
+    }
+
+    .Icon-inverted-caret {
+      opacity: 1;
+    }
+  }
+}
+
+.DropdownMenu-items {
+  @include font-regular();
+  @include end(0);
+
+  background: $white;
+  border: 0;
+  border-radius: $border-radius-default;
+  box-shadow: 0 0 2px transparentize($black, 0.5);
+  display: none;
+  list-style-type: none;
+  margin: 0;
+  padding: 4px 24px 20px;
+  position: absolute;
+  top: 30px;
+  white-space: nowrap;
+  width: auto;
+  z-index: 10;
+
+  &::after {
+    @include end(36px);
+
+    background-color: $white;
+    content: "";
+    display: block;
+    height: 8px;
+    position: absolute;
+    top: -4px;
+    transform: rotate(45deg);
+    width: 8px;
+  }
+}
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -15,7 +15,6 @@
   @include font-regular();
 
   color: $white;
-  cursor: pointer;
   font-size: $font-size-s;
   font-weight: normal;
   text-decoration: none;
@@ -25,6 +24,7 @@
 
     height: 10px;
     opacity: 0.6;
+    transition: opacity $transition-short ease-in-out;
     width: 9px;
   }
 }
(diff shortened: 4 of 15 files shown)


src/amo/components/Header/index.js
```diff
@@ -60,12 +84,31 @@ export class HeaderBase extends React.Component {
...
-            className="Header-authenticate-button Header-button Button--action
+
+          {isAuthenticated ? (
+            <DropdownMenu
+              title={username}
+              className="Header-authenticate-button Header-button"
+            >
+              <DropdownMenuItem title={i18n.gettext('Actions')} />
+              <DropdownMenuItem
+                title={i18n.gettext('Add-ons Home Page')}
```
willdurand: This is maybe not the best link, but the "log out" alone does not look nice 😞

src/amo/components/Header/index.js
```diff
@@ -78,11 +121,22 @@ export class HeaderBase extends React.Component {
...
+    username: state.user.username,
+    isAuthenticated: isUserAuthenticated(state),
+  };
+};
+
+export const mapDispatchToProps = (dispatch) => ({
+  handleLogOut: ({ api }) => {
+    return logOutFromServer({ api }).then(() => dispatch(logOutUser()));
+  },
+});
```
willdurand: I am a bit sad to duplicate the logic of the `AuthenticateButton`. Is that OK? Should I handle things differently?

src/ui/components/Icon/styles.scss
```diff
@@ -144,3 +144,7 @@
 .Icon-restart {
   @include icon($name: "restart");
 }
+
+.Icon-caron {
+  @include icon($name: "caron");
```
willdurand: @tofumatt I know it is not a "caron" _per se_ (not even sure what a caron actually is) but I was searching for "inverted caret" and found this word 😝

willdurand:
This PR should be ready for a first round of comments 😃 

I dont really know what to do on small screens and not sure about the dimensions in the UI spec (especially for the `font-size`). It seems bigger than the existing UI (I tried to stick to it though).

willdurand:
I dont understand why it is not building on Travis 😢

willdurand:
@tofumatt r? 😃

src/amo/components/Header/index.js
tofumatt: Ah, I see what you mean now. So in the future I think stuff like "My Profile" or "My collections" would work. Linking to "View Profile" and/or "Edit Profile/Settings" for now I think is better than "Add-ons Home Page". Sure, it will fallback to the old addons-server code but I think that's a better link.

src/ui/components/DropdownMenu/index.js
```diff
@@ -0,0 +1,33 @@
+/* @flow */
+import React from 'react';
+import classNames from 'classnames';
+
+import Icon from 'ui/components/Icon';
+
+import DropdownMenuItem from './DropdownMenuItem';
```
tofumatt: I don't like this style of putting a separate component file inside a component's directory. I understand this is a child component only this component is expected to use, but if that's the case I'd rather see it in the same file.

But I'd rather this be in a separate folder as its own component. It makes it easy to trace a single component to a file/folder and keeps the organisation rules clear/simple.

src/ui/components/DropdownMenu/index.js
```diff
@@ -0,0 +1,33 @@
...
+  title: string,
+  children: Array<DropdownMenuItem>,
+  className?: string,
+|};
+
+const DropdownMenu = (props: Props) => (
+  <div className={classNames('DropdownMenu', props.className)}>
+    <button href="#" className="DropdownMenu-title">
+      {props.title}
+      <Icon name="caron" />
```
tofumatt: What is a "caron"?

src/ui/components/Icon/styles.scss
tofumatt: Oh, hah, question answered.

I would rather just call it `inverted-caret`. If you had to google what it was called I'd say it's not obvious 😄

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
     expect(root.find('.Header-title').prop('children'))
       .toContain('Firefox Add-ons');
   });
+
+  it('displays `login` text when user is not connected', () => {
+    const { store } = dispatchClientMetadata();
+    const fakeI18n = getFakeI18nInst();
+    const allProps = { i18n: fakeI18n, location: {}, query: '' };
+
+    const wrapper = mount(
```
tofumatt: Is it not possible to just pass `store` and `i18n` directly to `Header`? Or will that cause issues with child components that get rendered?

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+      .toEqual('Register or Log in');
+    expect(wrapper.find(DropdownMenu)).toHaveLength(0);
+  });
+
+  it('displays a menu and the username when user is logged in', () => {
+    const { store } = dispatchSignInActions({ username: 'babar' });
+    const fakeI18n = getFakeI18nInst();
+    const allProps = { i18n: fakeI18n, location: {}, query: '' };
+
+    const wrapper = mount(
```
tofumatt: This looks duplicated enough it could be a helper function inside the test.

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+        </I18nProvider>
+      </Provider>
+    );
+
+    wrapper.find(DropdownMenuItem).find('button').simulate('click');
+
+    mockApi.verify();
+  });
+
+  describe('mapStateToProps', () => {
```
tofumatt: Do we need these tests here? @kumar303 has pointed out we can test `mapStateToProps` are working by testing with dispatch and looking at the output of the rendered component itself based on what `mapStateToProps` sends. I suppose testing `mapStateToProps` is more direct unit testing but it feels more removed.

tests/unit/ui/components/TestDropdownMenuItem.js
```diff
@@ -0,0 +1,71 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import DropdownMenuItem from 'ui/components/DropdownMenu/DropdownMenuItem';
+import Link from 'amo/components/Link';
+
+
+describe(__filename, () => {
+  it('renders a section when only `title` is specified', () => {
+    const item = shallow(<DropdownMenuItem title="A section" />);
```
tofumatt: The `title` attribute here seems like it would be better suited to just being the `children` attribute. It's not rendering a `title` attribute which is a common HTML attribute, so re-using that here for what is effectively `children` (these get rendered as text inside the button/link components) is confusing.

tests/unit/ui/components/TestDropdownMenuItem.js
```diff
@@ -0,0 +1,71 @@
...
+    expect(item).toIncludeText('A section');
+  });
+
+  it('cannot receive `to` and `onClick` props at the same time', () => {
+    expect(() => {
+      shallow(
+        <DropdownMenuItem
+          title="A section"
+          to="/"
+          onClick={jest.fn()}
```
tofumatt: Interesting. I guess there's nothing wrong with this (it's just a dummy function) but it might be more consistent to use `sinon.stub()`, just because already use that everywhere and it's our usual dummy function, even when we aren't testing it being called.

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -0,0 +1,113 @@
+@import "~core/css/inc/mixins";
+@import "~amo/css/inc/vars";
+@import "~ui/css/vars";
+
+$link-font-size: 16px;
+
+.DropdownMenu {
+  @include text-align-start();
+
+  cursor: pointer;
```
tofumatt: This actually creates a UX problem: the entire dropdown has a pointer cursor so there's no cursor style change on hover state of the actual link/button elements that are clickable versus the entire `DropdownMenu` which is not clickable. See this GIF:



Because there's only a colour change on hover state this will be entirely unclear to colourblind users. Instead of adding a text decoration or something else on hover though just making only the actual clickable elements show the link cursor should do the trick.

Basically: this needs to work like the "Developer Hub" link does.

tofumatt:
I'm clearing @kumar303 from review on this one because I don't think we both need to review all your PRs. 😄

src/amo/components/Header/index.js
```diff
@@ -6,28 +6,52 @@ import DownloadFirefoxButton from 'amo/components/DownloadFirefoxButton';
 import Link from 'amo/components/Link';
 import SearchForm from 'amo/components/SearchForm';
 import SectionLinks from 'amo/components/SectionLinks';
+import { logOutUser } from 'core/actions';
+import { logOutFromServer } from 'core/api';
 import AuthenticateButton from 'core/components/AuthenticateButton';
+import { isAuthenticated as isUserAuthenticated } from 'core/reducers/user';
```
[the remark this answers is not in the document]
willdurand: Eslint was complaining but I found another way.

src/ui/components/DropdownMenu/DropdownMenuItem.js
```diff
@@ -0,0 +1,49 @@
+/* @flow */
+import React from 'react';
+import classNames from 'classnames';
+
+import Link from 'amo/components/Link';
+
+
+type Props = {|
+  title: string,
```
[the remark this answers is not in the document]
willdurand: 🤣

src/ui/components/DropdownMenu/DropdownMenuItem.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  const classnames = classNames('DropdownMenuItem-link', {
+    'DropdownMenuItem-link--detached': detached,
+  });
+
+  if (to) {
+    return (
+      <li className="DropdownMenuItem">
+        <Link className={classnames} to={to}>{title}</Link>
```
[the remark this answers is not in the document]
willdurand: I would like to avoid spreading `props`. I could have a whitelist of `allowedProps` though.

src/ui/components/DropdownMenu/DropdownMenuItem.js
```diff
@@ -0,0 +1,49 @@
...
+      </li>
+    );
+  }
+
+  return (
+    <li className="DropdownMenuItem">
+      <button className={classnames} onClick={onClick}>
```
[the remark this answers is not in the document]
willdurand: I wanted a simple clickable item, but I guess I could use the existing `Button` and unstyle it. Also, I read this issue: https://github.com/mozilla/addons-frontend/issues/1904. Should I do add `href="#"` too?

tests/unit/ui/components/TestDropdownMenuItem.js
willdurand: Oh that makes a lot of sense!

tests/unit/ui/components/TestDropdownMenuItem.js
willdurand: Typo 😢

src/ui/components/DropdownMenu/DropdownMenuItem.js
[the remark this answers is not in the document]
tofumatt: Fair enough, but it would be nice to have some kind of test ensuring new props supposed by `Link` were passed. Dunno if that's possible; failing that a comment might be nice. But yeah a whitelist of allowed props works for me.

src/ui/components/DropdownMenu/DropdownMenuItem.js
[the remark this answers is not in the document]
tofumatt: I think this is fine then, actually. Our button component is as much about UI/style as it is a button. I think you can leave this as-is... provided you tested it on Firefox for Android and the first click works. Because yeah otherwise you'll want something that eventually outputs an `<a href="#">` because of the strangeness of the first/second click 😒

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+    it('gets props from store', () => {
+      const { store } = dispatchClientMetadata();
+      store.dispatch(setViewContext(VIEW_CONTEXT_EXPLORE));
+      expect(mapStateToProps(store.getState())).toEqual({
+        api: store.getState().api,
+        isAuthenticated: false,
```
kumar303: I agree with what @tofumatt said. You should not need to test `mapStateToProps` directly (in most cases) when you're using `shallowUntilTarget`. Test the mapped properties instead which will also help us cover our Redux integration. Here is an example of how to test the properties themselves:

```js
it('sets isAuthenticated from state', () => {
  store.dispatch(setViewContext(VIEW_CONTEXT_EXPLORE));
  
  expect(root.instance().props.isAuthenticated).toEqual(true);
});
```

*However*, this kind of test isn't so useful. What we really want to know is how the `isAuthenticated` prop used. It is used to show either a dropdown or an authenticate button and, at a glance, I think you already have tests to cover that logic :)

src/amo/components/Header/index.js
```diff
@@ -78,11 +121,22 @@ export class HeaderBase extends React.Component {
...
+    api: state.api,
+    isHomePage: state.viewContext.context === VIEW_CONTEXT_HOME,
+    username: state.user.username,
+    isAuthenticated: isUserAuthenticated(state),
+  };
+};
+
+export const mapDispatchToProps = (dispatch) => ({
+  handleLogOut: ({ api }) => {
+    return logOutFromServer({ api }).then(() => dispatch(logOutUser()));
```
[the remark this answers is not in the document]
willdurand: I came with  `createHandleLogOut()` function that returns the function given a `dispatch`.

tests/unit/amo/components/TestHeader.js
willdurand: Ok, I am fine with that. Thanks!

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+    const wrapper = mount(
+      <Provider store={store}>
+        <I18nProvider i18n={fakeI18n}>
+          <Header {...allProps} />
+        </I18nProvider>
+      </Provider>
+    );
+
+    expect(wrapper.find(AuthenticateButton).text())
+      .toEqual('Register or Log in');
```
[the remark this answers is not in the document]
willdurand: Ok, will give this helper a try then, thanks!

src/ui/components/DropdownMenu/DropdownMenuItem.js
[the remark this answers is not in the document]
willdurand: mmh, maybe I can actually pass `Link` directly.

src/amo/components/Header/index.js
```diff
@@ -6,28 +6,52 @@ import DownloadFirefoxButton from 'amo/components/DownloadFirefoxButton';
...
 
 export class HeaderBase extends React.Component {
   static propTypes = {
     i18n: PropTypes.object.isRequired,
+    api: PropTypes.object.isRequired,
     isHomePage: PropTypes.bool.isRequired,
     location: PropTypes.object.isRequired,
-    query: PropTypes.string,
+    query: PropTypes.string.isRequired,
+    isAuthenticated: PropTypes.bool.isRequired,
```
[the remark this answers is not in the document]
willdurand: Note: I use to put `isRequired` props first, then optional once. It does not play nice with alphabetisation though, so I'll try not to do that anymore and sort things alphabetically instead 😛

src/ui/components/DropdownMenu/styles.scss
willdurand: I see, this should be fixed.

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+    const wrapper = mount(
+      <Provider store={store}>
+        <I18nProvider i18n={fakeI18n}>
+          <Header {...allProps} />
+        </I18nProvider>
+      </Provider>
+    );
+
+    expect(wrapper.find(DropdownMenu)).toHaveLength(1);
+    expect(wrapper.find(DropdownMenu)).toIncludeText('babar');
```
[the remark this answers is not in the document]
willdurand: oh shoot, thanks for that!

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +63,95 @@ describe(__filename, () => {
...
+
+    const wrapper = mount(
+      <Provider store={store}>
+        <I18nProvider i18n={fakeI18n}>
+          <Header {...allProps} />
+        </I18nProvider>
+      </Provider>
+    );
+
+    wrapper.find(DropdownMenuItem).find('button').simulate('click');
```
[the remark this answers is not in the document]
willdurand: I always forget that our code coverage is so high that we can afford having very isolated test cases.. sorry for that :/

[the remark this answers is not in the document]
kumar303: Well, we are not at 100% so in a case like this it's worth checking the unit tests for `DropdownMenuItem` to make sure it covers what you are relying on.

willdurand:
Thanks for the reviews, I came up with a new menu, and fixed most of the issues hopefully. I did not know how to use `shallowUntilTarget()` but now I get it: very cool!

willdurand:
@tofumatt r?

src/amo/components/Header/index.js
```diff
@@ -60,12 +83,35 @@ export class HeaderBase extends React.Component {
           <DownloadFirefoxButton
             className="Header-download-button Header-button"
           />
-          <AuthenticateButton
-            className="Header-authenticate-button Header-button Button--action
+
+          {this.props.isAuthenticated ? (
+            <DropdownMenu
```
tofumatt: I didn't know where to put this but this shouldn't have a cursor that changes:



It's confusing because clicking on my username doesn't do anything. So I don't think that button, when signed in, should have a cursor style. That might mean an extra `className` when the user is signed in so we can style it differently.

src/amo/components/Header/index.js
```diff
@@ -60,12 +83,35 @@ export class HeaderBase extends React.Component {
...
+              className="Header-authenticate-button Header-button"
+            >
+              <DropdownMenuItem>{i18n.gettext('Account')}</DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link href={`/user/${username}/`}>
+                  {i18n.gettext('My profile')}
+                </Link>
+              </DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link href="/users/edit">
```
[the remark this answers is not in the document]
willdurand: According to https://addons.mozilla.org/en-US/firefox/, this is it, without a trailing slash..

src/ui/components/DropdownMenuItem/index.js
```diff
@@ -0,0 +1,57 @@
...
+|};
+
+const DropdownMenuItem = ({ children, onClick, detached = false }: Props) => {
+  const childIsComponent = typeof children === 'object';
+
+  if (
+    childIsComponent &&
+    children.type && children.type.displayName !== 'Connect(LinkBase)'
+  ) {
+    throw new Error(oneLine`Only the "Link" component is supported as a child
```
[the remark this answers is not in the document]
willdurand: My initial idea was to restrict the menu to `Link` or `text` entries. There are Flow types but it will not be caught because Flow is not used everywhere yet. That is why I have added such a check (semi-related: https://github.com/mozilla/addons-frontend/pull/2972#discussion_r134330441).

Maybe it is not needed though.

src/ui/components/DropdownMenuItem/index.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  const classnames = classNames('DropdownMenuItem-link', {
+    'DropdownMenuItem-link--detached': detached,
+  });
+
+  if (childIsComponent) {
+    return (
+      <li className="DropdownMenuItem">
+        {React.cloneElement((children: Link), { className: classnames })}
```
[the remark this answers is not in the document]
willdurand: That's because I wanted to pass the `className` prop to the `children`. Not sure if there are performance implications, see: https://facebook.github.io/react/docs/react-api.html#cloneelement.

`(children: Link)` is indeed Flow syntax.

src/ui/components/DropdownMenuItem/index.js
```diff
@@ -0,0 +1,57 @@
...
+    );
+  }
+
+  const classnames = classNames('DropdownMenuItem-link', {
+    'DropdownMenuItem-link--detached': detached,
+  });
+
+  if (childIsComponent) {
+    return (
+      <li className="DropdownMenuItem">
```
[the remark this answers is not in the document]
willdurand: I have been reading this PR again and yes, this could simplify this component a lot.

src/ui/components/DropdownMenuItem/styles.scss
```diff
@@ -0,0 +1,38 @@
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+$section-font-size: 11px;
```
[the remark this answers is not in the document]
willdurand: Do you know why the mock-ups have different sizes btw?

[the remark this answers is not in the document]
tofumatt: I'm not quite sure but it seems to be a frustrating Sketch issue? It seems a lot of stuff in Sketch is more accurate if you take something like 2/3 of the stated sizes? I just sort of look at it at 100% on my screen then make the HTML resemble that. It's admittedly not great, I'm sorry 😒

willdurand:
@tofumatt I got rid of the `cloneElement` but kept the extra check on the `children` prop. I can definitely remove it but I don't really know what to do to be honest..

I have removed the custom `font-size` variables since removing `uppercase` on the sections made them slightly too small (it was 11px before). I used the existing `small` size (12px) and it looks better now.

src/ui/components/DropdownMenuItem/styles.scss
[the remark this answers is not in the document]
willdurand: ok I will apply the 2/3 rule then. thanks

src/ui/components/DropdownMenu/index.js
```diff
@@ -0,0 +1,67 @@
...
+|};
+
+export class DropdownMenuBase extends React.Component {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = { buttonIsActive: false };
+  }
+
+  state: {| buttonIsActive: bool |};
```
tofumatt: Huh, nitpick but I thought we'd put these before `constructor`.

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+  font-size: $font-size-s;
+  font-weight: normal;
+  text-decoration: none;
+
+  .Icon-inverted-caret {
+    @include margin-start(5px);
+
+    height: 10px;
+    opacity: 0.6;
+    transition: opacity $transition-short ease-in-out;
```
tofumatt: merci !

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+    width: 9px;
+  }
+}
+
+.DropdownMenu--active {
```
tofumatt: Here's a weird thought: we should target, as the root selector, the element we want to style and if there are contextual styles applied to it from a state like hover/active, they should be inside the selector. So instead of a section of restyles based on `.DropdownMenu--active {}` being set you'd do:

```scss
.DropdownMenu-items {
  [all the normal styles here]

  .DropdownMenu--active & {
    display: block;
  }
}
```

I've tried to do that (though maybe not at first, but tried going forward about three months ago) and I found it works well and makes the file easier to understand. I thought I said so in the CSS guides but forget now.

Anyway, just a thought!

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -0,0 +1,88 @@
...
+  .Icon-inverted-caret {
+    opacity: 1;
+  }
+}
+
+// This is needed because `:hover` on mobile prevents the main button to close
+// the dropdown menu. On mobile, `.DropdownMenu--active` controls the
+// visibility of the menu. On "large" devices (and above), `:hover` also opens
+// the menu.
+@include respond-to(large) {
```
tofumatt: Similarly: I prefer when the selectors are at the top and the `respond-to(large) {}` things are inside them. It makes it easier to reason about an element's style rather than a screen size's style, because some things are the same across sizes but others aren't. I think it's a better way to organise and would prefer we do:

```scss
.SomeClass {
  @include respond-to(large) {
    display: block;
  }
}
```

going forward. Sorry if I missed that in previous reviews. If it's missing from the CSS guides I should add it.

src/ui/components/DropdownMenuItem/index.js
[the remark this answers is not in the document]
tofumatt: I think there's nothing technically preventing another element from being a child so this check seems unneeded. I get that it might be awkward to add a non-linked element in there but I say we remove it for now.

src/ui/components/DropdownMenuItem/styles.scss
```diff
@@ -0,0 +1,35 @@
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+.DropdownMenuItem-section {
+  color: $grey-50;
```
tofumatt: This can be imported from `~photon-colors/colors' now.

tests/unit/ui/components/TestDropdownMenuItem.js
```diff
@@ -0,0 +1,82 @@
...
+        A button that is detached from the rest of the menu
+      </DropdownMenuItem>
+    );
+
+    expect(item).toHaveClassName('DropdownMenuItem');
+    expect(item).toHaveClassName('DropdownMenuItem-link');
+    expect(item).toHaveClassName('DropdownMenuItem--detached');
+  });
+
+  it('throws an error if child component is not a `Link`', () => {
```
tofumatt: Again, I think this isn't needed now that we aren't checking the component type. 😄

src/ui/components/DropdownMenu/index.js
willdurand: nope^^

```
/Users/william/projects/mozilla/addons-frontend/src/ui/components/DropdownMenu/index.js
  20:3  error  state should be placed after constructor  react/sort-comp
```

src/ui/components/DropdownMenu/styles.scss
willdurand: I like it!

src/ui/components/DropdownMenuItem/index.js
```diff
@@ -0,0 +1,35 @@
...
+import React from 'react';
+import classNames from 'classnames';
+
+import Link from 'amo/components/Link';
+
+import './styles.scss';
+
+
+type Props = {|
+  children: string | Link,
```
kumar303: This looks like it should be `children?:` since it's optional for the case when `onClick` is defined, I think?

src/ui/components/DropdownMenuItem/index.js
```diff
@@ -0,0 +1,35 @@
...
+  detached?: boolean,
+  onClick?: Function,
+|};
+
+const DropdownMenuItem = ({ children, onClick, detached = false }: Props) => {
+  const childIsComponent = typeof children === 'object';
+  const classnames = classNames('DropdownMenuItem', {
```
kumar303: be careful with overshadowing names like this. I know it will pass lint checks since it has a lower case `n` but this makes it a little confusing to read. I would suggest using a different name like `classes` or `computedClassNames` or `_classNames`.

src/core/components/AuthenticateButton/index.js
```diff
@@ -99,16 +103,21 @@ export const mapStateToProps = (
 });
 
 type DispatchMappedProps = {|
-  handleLogOut: HandleLogOutFunc,
+  handleLogOut: HandleLogOutFunction,
 |};
 
+export const createHandleLogOutFunction = (
```
kumar303: Since this is now a shared unit, it needs tests of its own. This is necessary because it allows us to update its implementation without having to rewrite the tests of each component that uses it.

tests/unit/amo/components/TestHeader.js
```diff
@@ -40,4 +59,35 @@ describe(__filename, () => {
...
+    expect(wrapper.find(DropdownMenu)).toHaveProp('text', 'babar');
+  });
+
+  it('allows a signed-in user to log out', () => {
+    const { store } = dispatchSignInActions({ username: 'babar' });
+    const wrapper = renderHeader({ store });
+    const mockApi = sinon.mock(api);
+
+    mockApi
+      .expects('logOutFromServer')
```
kumar303: This component is relying on `props.handleLogOut()` so you should be mocking the expectation for how `handleLogOut()` is called. As I mentioned in a comment, the `createHandleLogOutFunction` needs tests of its own which will look like the one here -- i.e. `mockApi.expects('logOutFromServer')`. 

This kind of approach allows us to separate the responsibilities of each unit that makes up the application.

In order to inject a `sinon.stub()` for `props.handleLogOut` you may need to change [this line](https://github.com/willdurand/addons-frontend/blob/dfa71a8f67fd9322146114e65e1ed778c086d4ff/src/amo/components/Header/index.js#L137) but I'm not sure:

```js
export const mapDispatchToProps = (dispatch, ownProps) => ({
  handleLogOut: ownProps.handleLogOut || createHandleLogOutFunction(dispatch),
});
```

tests/unit/ui/components/TestDropdownMenu.js
```diff
@@ -0,0 +1,76 @@
...
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
+import Icon from 'ui/components/Icon';
+import { createFakeEvent, shallowUntilTarget } from 'tests/unit/helpers';
+
+
+describe(__filename, () => {
+  it('renders a menu', () => {
+    const menu = shallowUntilTarget(
+      <DropdownMenu text="Menu" />,
+      DropdownMenuBase
```
kumar303: since this is repeated for every call, you should definitely add a helper at the top of the test suite.

tests/unit/ui/components/TestDropdownMenu.js
```diff
@@ -0,0 +1,76 @@
...
+
+    // User clicks the menu main button.
+    menu.find('.DropdownMenu-text').simulate('click', createFakeEvent());
+    expect(menu).toHaveClassName('DropdownMenu--active');
+
+    // User clicks somewhere else.
+    // The first `instance()` call is Enzyme API, the second `getInstance()`
+    // call is react-onclickoutside API. See:
+    // https://github.com/Pomax/react-onclickoutside#but-how-can-i-access-my-component-it-has-an-api-that-i-rely-on
+    root.instance().getInstance().handleClickOutside();
```
kumar303: This seems fragile. Could you do it like this?

```js
const root = mount(
  <div className="wrapper">
    <DropdownMenu text="Menu" />
  </div>
);
root.find('.wrapper').simulate('click');
```

src/ui/components/DropdownMenu/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+import Icon from 'ui/components/Icon';
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
+
+import './styles.scss';
+
+
+type Props = {|
+  text: string,
+  children: Array<DropdownMenuItem>,
```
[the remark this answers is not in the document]
willdurand: we need flow 53 for that :(

tests/unit/ui/components/TestDropdownMenu.js
willdurand: no, this did not work.. that was my first attempt. I tried to add other children to the `wrapper` too, but no success.

PR #2995, opened 2017-08-23
feat: Create SearchFilters component to replace SearchSort

### LINKED ISSUE #2503: Import colours/styles from Photon package
We should import colours/styles/etc. from https://github.com/FirefoxUX/photon, which handily has, for instance, a list of colours we'd want for our buttons, etc.

### LINKED ISSUE #2611: Sort header section borders are cut on the corners
Steps to reproduce:
1.On Android , select any field from header (ie. Extensions)
2.Press "More popular extensions " link
3.Expand the "Sort" section

Expected results:
No UI issues

Actual results:
Sort header section borders are cut on the corners

Notes : 
Reproducible also on Desktop with mobile view

borders

### LINKED ISSUE #2810: On hover event over the selected Search option, the background color turns white

1.Search for a query in search bar on browser ( Firefox, Safari )
2. Go to search results and use search filters. -relevance (default), top rates, most users, and recently updated on left panel. 
3. On hovering/focussing the mouse over the selected Search option, the background color turns white (which is by selection blue) and the text color is also white.

 The hover event  shouldn't make background white

screen shot 2017-07-16 at 12 56 09 am

### LINKED ISSUE #2956: Implement new Search Sort widget
The new search sort widget will be more standard (no custom weird buttons) and look like:

screenshot 2017-08-16 16 35 47

This will fix:

* #2611
* #2810

tofumatt:
* Fixes #2503
* Fixes #2611
* Fixes #2810
* Fixes #2956

This includes a new `ToggleOnMobileCard` to replace `SearchSort`'s toggle on mobile that allows us to render a card that is expanded on desktop but can be folded up/expanded on mobile/tablets.

Just realising now that (`ToggleOnMobileCard`) might not be the best component name, so let me know if anyone has better ideas for that 😄 

### Screenshots!

screenshot 2017-08-23 18 27 28

-------

This doesn't implement "Show legacy extensions" (I don't think it's worth including–needs a product decision) or "Show Featured add-ons" (needs an API). They will be easy to add in a follow-up patch.


src/ui/components/Icon/styles.scss
```diff
@@ -40,6 +40,16 @@
   width: 112px;
 }
 
+.Icon-triangle-down {
+  background: url("./triangle-down.svg") center no-repeat;
```
willdurand: I thought we had the `icon` mixin in `master` already? Did you rebase lately?

src/ui/components/ToggleOnMobileCard/index.js
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable react/no-danger */
+
+import classNames from 'classnames';
+import React from 'react';
+import PropTypes from 'prop-types';
+// import ReactDOM from 'react-dom';
```
willdurand: 😛

tofumatt: Oops! Should delete! This was created from an old (naughty) component! 😅

src/ui/components/Icon/styles.scss
tofumatt: Oh, maybe. Right now we don't handle SVGs well so this could be a different colour... I'll double-check it.

src/ui/components/ToggleOnMobileCard/index.js
```diff
@@ -0,0 +1,72 @@
...
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+// import { sanitizeHTML } from 'core/utils';
```
willdurand: 😛

src/ui/components/Icon/styles.scss
tofumatt: Wait, sorry, I just realised exactly what you were asking. Yeah this PR started awhile back so I was missing it I think. I'll fix that!

tofumatt:
I pushed a few commits because this had conflicts. Still good to review though.

src/amo/components/SearchFilters/styles.scss
```diff
@@ -0,0 +1,22 @@
+@import "~photon-colors/colors";
+// @import "~amo/css/inc/vars";
+// @import "~core/css/inc/mixins";
```
kumar303: These can be deleted

src/amo/components/SearchPage/index.js
```diff
@@ -1,10 +1,39 @@
+import React from 'react';
```
kumar303: If you're going to use flow this needs `/* @flow */`

src/amo/components/SearchFilters/index.js
```diff
@@ -0,0 +1,185 @@
...
+      { children: i18n.gettext('Theme'), value: ADDON_TYPE_THEME },
+    ];
+  }
+
+  operatingSystemOptions() {
+    const { i18n } = this.props;
+
+    return [
+      // HACK: Should we be using at least `null` here or will that
+      // be oddly cast?
```
kumar303: I think an empty string value makes sense since it's mapping to query string params.

src/amo/components/SearchFilters/index.js
```diff
@@ -0,0 +1,185 @@
...
+    router: PropTypes.object,
+  }
+
+  onSelectElementChange = (event) => {
+    const { clientApp, filters, lang, pathname } = this.props;
+    event.preventDefault();
+
+    const newFilters = { ...filters };
+
+    // TODO: Make this better; paginator resets it anyway
```
kumar303: Can you comment on why it needs to be deleted?

src/ui/components/ToggleOnMobileCard/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = { expanded: false };
+  }
+
+  onClick = (event) => {
+    event.preventDefault();
+
+    this.setState({ expanded: !this.state.expanded });
```
kumar303: It would be nicer to use Redux here but I guess you'd have to generate an ID [like the ErrorHandler](https://github.com/mozilla/addons-frontend/blob/master/src/core/errorHandler.js#L97) and this patch is already massive so no biggie.

src/ui/components/ToggleOnMobileCard/index.js
```diff
@@ -0,0 +1,68 @@
...
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+import Card from 'ui/components/Card';
+import Icon from 'ui/components/Icon';
+
+import './styles.scss';
+
+
+export class ToggleOnMobileCardBase extends React.Component {
```
kumar303: This name seems odd to me for what it does. How about just calling it `ExpandableCard` ? Either way, it would be nice to see a comment at the top explaining how it's a card that is collapsed by default on mobile.

src/ui/components/Icon/styles.scss
```diff
@@ -44,6 +44,16 @@
...
+  background-size: contain;
+}
+
+.Icon-triangle-down-black {
+  background: url("./triangle-down-black.svg") center no-repeat;
```
kumar303: I checked out this branch locally and I don't see an arrow icon in the header (in mobile view). Maybe it's because you need to use the new `icon` mixin?

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -0,0 +1,142 @@
...
+import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
+import {
+  createStubErrorHandler,
+  getFakeI18nInst,
+  shallowUntilTarget,
+} from 'tests/unit/helpers';
+
+
+describe(__filename, () => {
+  const fakeEvent = { preventDefault: sinon.stub() };
```
kumar303: you can use `createFakeEvent()` from `tests/unit/helpers.js`.

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -0,0 +1,142 @@
...
+        store={store}
+        pathname={pathname}
+        {...props}
+      />, SearchFiltersBase,
+      { shallowOptions: { context } }
+    );
+  }
+
+  beforeEach(() => {
+    context = { router: { push: sinon.stub() } };
```
kumar303: I don't think you need this. You should be able to access the router simply as `props.router`, no?

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  it('renders a SearchFilters component', () => {
+    const root = render();
+
+    expect(root.find('.SearchFilters')).toHaveLength(1);
+  });
+
+  it('changes the URL when a new addonType filter is selected', () => {
+    const root = render({ filters: { query: 'Music player' } });
```
kumar303: Thanks for adding a realistic value for `query`. I think that kind of thing helps understand the tests better.

src/ui/components/ToggleOnMobileCard/index.js
tofumatt: Oh but that's a good point. I'll just link to this comment from #1250.

src/ui/components/Icon/styles.scss
tofumatt: Yeah I bet after rebasing this was messed up, and I should use the mixin anyway but forgot even after it was pointed out, sorry.

tests/unit/amo/components/TestSearchFilters.js
tofumatt: Oh, perfect then!

tests/unit/amo/components/TestSearchFilters.js
tofumatt: Oh, I didn't know that. I'll try.

tofumatt:
Thanks for the r? I know this turned out pretty massive 😢

I'm working on the header styles right now so I'll probably get to tweaking this tomorrow.

src/ui/components/ToggleOnMobileCard/index.js
tofumatt: Yeah, it's weird to name. It actually can't be expanded at all on desktop (there's no point), it's just expandable on smaller screens. But I think `ExpandableCard` is still a better name, agreed 👍

tests/unit/amo/components/TestSearchFilters.js
tofumatt: Nope, actually, that won't work when tested. We do it this way elsewhere and using `withRouter` won't work well unless we have at least react-router 3. Also `withRouter` is really just a convenience anyway; [using `context` is still okay](https://github.com/ReactTraining/react-router/issues/3350#issuecomment-211577962).

tofumatt:
Alrighty, this is ready for another review.

src/amo/components/SearchPage/index.js
```diff
@@ -1,18 +1,60 @@
...
-import { convertQueryParamsToFilters } from 'core/searchUtils';
+import {
+  convertFiltersToQueryParams,
+  convertQueryParamsToFilters,
+} from 'core/searchUtils';
 
 
-export function mapStateToProps(state, ownProps) {
+type PropTypes = {|
+  filters: Object,
```
kumar303: Since you're opting for Flow and this already has a type definition, you might as well hook it up. It will give you protection around using unknown filters.

```diff
diff --git a/src/amo/components/SearchPage/index.js b/src/amo/components/SearchPage/index.js
index 82df0709..21b6b50d 100644
--- a/src/amo/components/SearchPage/index.js
+++ b/src/amo/components/SearchPage/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+/* global $PropertyType */
 import React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -8,10 +9,11 @@ import {
   convertFiltersToQueryParams,
   convertQueryParamsToFilters,
 } from 'core/searchUtils';
+import type { SearchParams } from 'core/api/search';
 
 
 type PropTypes = {|
-  filters: Object,
+  filters: $PropertyType<SearchParams, 'filters'>,
   pathname: string,
 |};
 
diff --git a/src/core/api/search.js b/src/core/api/search.js
index 737e154b..b5cfd0da 100644
--- a/src/core/api/search.js
+++ b/src/core/api/search.js
@@ -12,7 +12,7 @@ import type { ApiStateType } from 'core/reducers/api';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 
 
-type SearchParams = {|
+export type SearchParams = {|
   api: ApiStateType,
   auth: boolean,
   // TODO: Make a "searchFilters" type because these are the same args
```

src/core/constants.js
```diff
@@ -97,6 +97,13 @@ export const VIEW_CONTEXT_HOME = 'VIEW_CONTEXT_HOME';
 export const SEARCH_SORT_POPULAR = 'hotness';
 export const SEARCH_SORT_TOP_RATED = 'rating';
 
+// Operating system for add-ons
+export const OS_WINDOWS = 'windows';
+export const OS_MAC = 'mac';
+export const OS_LINUX = 'linux';
+// TODO: Do we need this? We have clientApp.
```
kumar303: Yeah, I expected to see it as an option in the operating system drop down. Users will need to filter for Android only extensions somehow. Is that happening in a separate patch?

src/ui/components/Icon/styles.scss
```diff
@@ -44,6 +44,16 @@
   width: 112px;
 }
 
+.Icon-triangle-down {
+  background: url("./triangle-down.svg") center no-repeat;
```
kumar303: This should be `@include icon($name: "triange-down")`

src/ui/components/Icon/styles.scss
kumar303: Same thing here

src/ui/components/ToggleOnMobileCard/styles.scss
```diff
@@ -0,0 +1,61 @@
...
+  &:link {
+    color: $grey-60;
+  }
+
+  @include respond-to(large) {
+    cursor: default;
+  }
+}
+
+.ToggleOnMobileCard-ToggleArrow {
```
kumar303: Is this supposed to render an arrow in the header to indicate that tapping on the header will expand it? I don't see an arrow:

screenshot 2017-08-28 15 53 43

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -0,0 +1,143 @@
...
+        store={store}
+        pathname={pathname}
+        {...props}
+      />, SearchFiltersBase,
+      { shallowOptions: { context } }
+    );
+  }
+
+  beforeEach(() => {
+    context = { router: { push: sinon.stub() } };
```
kumar303: It looks like the only variable you need to declare in the upper scope is `router`.

tests/unit/ui/components/TestToggleOnMobileCard.js
```diff
@@ -0,0 +1,64 @@
...
+function render(props) {
+  // We have to mount because of the usage of `setState` in this component.
+  return mount(<ToggleOnMobileCardBase i18n={getFakeI18nInst()} {...props} />);
+}
+
+describe(__filename, () => {
+  let fakeEvent;
+
+  beforeEach(() => {
+    fakeEvent = { preventDefault: sinon.stub() };
```
kumar303: this should use `createFakeEvent()`

tests/unit/amo/components/TestSearchFilters.js
tofumatt: I just realised, after looking at another patch, I was using `withRouter` wrong. This can use the `withRouter` HOC; I'll move it over.

src/core/constants.js
tofumatt: It might be worth doing as a separate patch because changing this OS flag would change the `clientApp` in the URL, not the OS in the search params.

I'd rather do it in a follow-up patch as it's different behaviour and sort of unrelated to this change. If anything it should be a select/switch/etc. elsewhere that is site-wide.

tofumatt:
I've fixed up the comments here and filed #3035 as a separate issue. Ready for another r?

Sorry I got distracted by the new header.

tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -0,0 +1,59 @@
+import { mount } from 'enzyme';
+import React from 'react';
+
+import ExpandableCard from 'ui/components/ExpandableCard';
+import { createFakeEvent, getFakeI18nInst } from 'tests/unit/helpers';
+
+
+function render(props) {
+  // We have to mount because of the usage of `setState` in this component.
```
kumar303: It looks like you only need `mount()` for the interaction with `Card` and only for one test: *it toggles when clicked*. I think it would be good to use `shallowUntilTarget()` for the other tests to keep it lean.

src/amo/components/SearchPage/index.js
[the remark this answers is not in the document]
tofumatt: Shoot, yes, I'm sorry. It just slipped my mind making fix-ups. Will do!

src/ui/components/ExpandableCard/index.js
```diff
@@ -0,0 +1,68 @@
...
+    const { expanded } = this.state;
+
+    const headerWithExpandLink = (
+      <a
+        aria-checked={expanded}
+        className="ExpandableCard-ToggleLink"
+        href="#toggle-filters"
+        onClick={this.onClick}
+        role="switch"
+        title={i18n.gettext('Tap to toggle contents')}
```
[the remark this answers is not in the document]
tofumatt: Sure, works for me.

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -16,7 +77,34 @@ describe(__filename, () => {
...
       expect(mapStateToProps(state, { location })).toEqual({
         filters: {
-          clientApp: CLIENT_APP_ANDROID,
+          page: 2,
+          query: 'burger',
+        },
+      });
+    });
+
+    it('returns ignores clientApp in location', () => {
```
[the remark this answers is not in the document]
tofumatt: Oh it shouldn't have `returns` in there, my bad.

tests/unit/ui/components/TestExpandableCard.js
tofumatt: Ah, fair enough. I guess there's probably a decent speed difference when all the tests are considered in using `shallow` whenever possible.

kumar303: In addition to speed, `mount()` also makes tests brittle since a problem with a component farther down in the render tree could cause the test to fail.

tofumatt: Ah, true.

PR #3013, opened 2017-08-28
Update documentation

### LINKED ISSUE #3016: Add a note about testing in documentation
The main goal is to describe how to write (unit) tests for this project.

willdurand:
Fix #3016 

---

This PR updates the project documentation a bit:

- removed the `How to add a page` file since it is outdated and misleading
- fixed a typo in the `CSS` documentation
- added a new rule to the `CSS` documentation, related to the scss parent selector
- added a new page gathering everything I learnt writing tests over the last two weeks


docs/adding-a-page.md
```diff
@@ -1,337 +0,0 @@
-# addons-frontend
-
-This will outline what is required to add a page to the project. A basic knowledge of
-[react](https://facebook.github.io/react/docs/getting-started.html) and
-[redux](http://redux.js.org/) is assumed.
-
-**Note:** This page is very out-of-date and does not reflect our practices anymore. We now use [redux-saga](https://github.com/redux-saga/redux-saga) for API requests. See `amo/components/Categories.js` for a more modern example of a component that makes API/async requests for data.
```
tofumatt: Rather than removing this page entirely, we could leave this section and mention these docs do need to be expanded. Pointing someone to a modern component with sagas + `shallowUntilTarget` tests is still better than saying nothing. I think code makes for lousy documentation but if we direct people to right bits of code it still beats just saying nothing.

docs/css.md
```diff
@@ -1,11 +1,12 @@
 # Our Approach to Writing CSS
 
-We use [SCSS](http://sass-lang.com/documentation/file.SCSS_FOR_SASS_USERS.html) to write our CSS. All components should keep their CSS local to their own component (eg. use `Button/style.scss` next to `Button/index.js`). Here are a few guidelines for writing scalable CSS:
+We use [SCSS](http://sass-lang.com/documentation/file.SCSS_FOR_SASS_USERS.html) to write our CSS. All components should keep their CSS local to their own component (e.g. use `Button/styles.scss` next to `Button/index.js`). Here are a few guidelines for writing scalable CSS:
```
tofumatt: Oops, my bad, I wrote this too; should have caught the `style.scss` mixup! 😅

docs/css.md
```diff
@@ -1,11 +1,12 @@
...
 * **Don't nest selectors; use more specific class names instead**: prefer `<ul  className="List"><li className="List-item"></li></ul>` and the selector `.List-item` over `.List li`. This seems verbose but it prevents nesting bugs and confusion as the codebase grows. It also makes it easier to scan CSS rules at a glance.
-* **When overrides make sense, keep them in the parent component and use CSS nesting:** this is the only time we use nesting instead of specifically-named class selectors. There are a few cases where tiny adjustments may be made to `margin`, etc. for a child component in only one context. If modifying the `margin` of `Result` inside `SearchResultList`, it is appropriate to declare `.SearchResultList .Result` in `SearchResultList/style.scss`.
+* **Don't use the [parent selector](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#Referencing_Parent_Selectors_____parent-selector), except for pseudo-classes**: prefer full class names, it is easier to find. Using `&` for pseudo-class such as `:hover`, `:active`, etc. is allowed though.
```
tofumatt: Thanks 👍

docs/testing.md
```diff
@@ -0,0 +1,68 @@
+# Testing
+
+We want to maintain a project with a high coverage (aiming for 100%). Our main [coverage criterion](https://en.wikipedia.org/wiki/Code_coverage#Coverage_criteria) is **branch coverage**, that is making sure all possible code branches are covered by test. [Coveralls](https://coveralls.io/github/mozilla/addons-frontend) helps us monitor the status of our test suite coverage and it is a good idea to double-check its output when you are opening a Pull Request. You can also [generate a coverage report locally](https://github.com/mozilla/addons-frontend/#code-coverage).
```
tofumatt: The second-to-last sentence here could be a bit more concise and probably cut down. It might flow better simply as:

```markdown
[Coveralls](https://coveralls.io/github/mozilla/addons-frontend) monitors our test coverage for every pull request. You can make sure you haven't decreased coverage by checking the Pull Request on GitHub.
```

The tests will actually fail if test coverage drops (enough).

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+
+## General guidelines
+
+- Imports should be alphabetized, even in test file.
```
tofumatt: test files

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+- Comments should be full sentences to ease readability.
+- There are a lot of helpers in the `tests/unit/helpers` and `tests/unit/amo/helpers` modules.
```
tofumatt: "There are a lot of helpers in the `tests/unit/helpers` and `tests/unit/amo/helpers` modules–please use them."

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+- Use the action creators instead of hard coding props. This applies to both UI components and reducers/sagas.
```
tofumatt: "Use action creators in `amo/actions/` and `core/actions/` instead of hard-coding `dispatch()` arguments or state data for tests. This applies to both UI components and sagas."

(We don't dispatch from reducers so I don't think it matters there, unless you meant for tests... in which case I suppose you can leave it in.)

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+- Try to use constants (likely from `core/constants`) as much as you can. Corollary: avoid hard coded values.
```
tofumatt: I worry about the word "corollary" for ESL folks 😅

"As much as you can" is a bit of a weird guideline too. Maybe this could be:

```markdown
Use constants (see `core/constants`) when using the same value across files. This avoids hard-coding values and [magic constants](https://en.wikipedia.org/wiki/Magic_constant).
```

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+``` js
+describe(__filename, () => {
+});
+```
+
+## Spies/Stubs/Mocks and Sinon.JS
+
+We use [sinon](http://sinonjs.org/) for spies, stubs and mocks. In addition, we use [sinon assertions](http://sinonjs.org/releases/v3.2.1/assertions/) over Jest expectations because failure messages are more descriptive.
+
+No need to import `sinon`, it is already done globally in the test suite configuration.
```
tofumatt: "There is no need to import `sinon`, it is already in the global scope for all test files."

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+When creating a new test file, start with a `describe()` block that takes the current file name as first argument. This makes easy to find/edit a failing test case as Jest will display the test file in its output:
+
+``` js
+describe(__filename, () => {
+});
+```
+
+## Spies/Stubs/Mocks and Sinon.JS
+
+We use [sinon](http://sinonjs.org/) for spies, stubs and mocks. In addition, we use [sinon assertions](http://sinonjs.org/releases/v3.2.1/assertions/) over Jest expectations because failure messages are more descriptive.
```
tofumatt: An example of what the `expect` vs `sinon.assert.calledWith` code would look like here would be handy.

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+
+No need to import `sinon`, it is already done globally in the test suite configuration.
+
+## Testing reducers and sagas
+
+For sagas/reducers, there are two useful helpers: `dispatchClientMetadata()` and `dispatchSignInActions()` (`tests/unit/amo/helpers`) that should be used to initialize state in a realistic manner. The former is used to obtain a non-authenticated state while the latter returns an authenticated state.
+
+When you need a `errorHandler` or a `errorHandlerId`, use the `createStubErrorHandler()` helper from `tests/unit/helpers`.
+
+When asserting for exceptions/errors, do not omit parentheses with an arrow function expression. It explicits what is expected to throw an error:
```
tofumatt: I find this section a bit confusing. Could you rephrase? I think what you're getting across is "don't use ES6 shorthand functions/implicit return functions".

docs/testing.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Testing UI components
+
+We use [Enzyme](http://airbnb.io/enzyme/docs/api/index.html) for testing UI components (React components). You should test the final component ideally. Below are a few rules regarding Enzyme:
+
+- Prefer `shallow()` over `mount()` when it makes sense.
+- Assert components on public properties (props), _e.g._:
+
+    ``` js
```
tofumatt: This don't need to be indented, just the three backticks will work fine. Also you don't need the space before JS (we tend not to use it) but I don't really care 😄

docs/adding-a-page.md
willdurand: Indeed!

docs/testing.md
willdurand: Sounds good to me

docs/testing.md
willdurand: good catch, thank

docs/testing.md
willdurand: Sounds good to me

docs/testing.md
willdurand: Ah right!

docs/testing.md
willdurand: I indented this block because I want it to be part of the last bullet point. If I dont indent, it starts a new paragraph.

tofumatt: Ah, oops, that makes sense then! Carry on 😄

docs/testing.md
willdurand: I actually had to use an action creator to init a `state`.

willdurand:
Pushed fixes. Sorry for the diff on the `adding-a-page.md` file...

docs/testing.md
```diff
@@ -0,0 +1,81 @@
...
+There is no need to import sinon, it is already in the global scope for all test files.
+
+
+## Testing reducers and sagas
+
+For sagas/reducers, there are two useful helpers: `dispatchClientMetadata()` and `dispatchSignInActions()` (`tests/unit/amo/helpers`) that should be used to initialize state in a realistic manner. The former is used to obtain a non-authenticated state while the latter returns an authenticated state.
+
+When you need a `errorHandler` or a `errorHandlerId`, use the `createStubErrorHandler()` helper from `tests/unit/helpers`.
+
+When asserting for exceptions/errors, do not use ES6 shorthand functions/implicit return functions. It explicits what is expected to throw an error:
```
kumar303: Why not? I think this code would have been fine:

```js
expect(() => methodThatThrowsAnError())
  .toThrow(/expected error message/);
```

docs/testing.md
```diff
@@ -0,0 +1,81 @@
...
+const expectedLoadAction = autocompleteLoad(results);
+
+await sagaTester.waitFor(expectedLoadAction.type);
+mockApi.verify();
+
+const loadAction = sagaTester.getCalledActions()[2];
+expect(loadAction).toEqual(expectedLoadAction);
+```
+
+## Testing UI components
```
kumar303: It might be good to show an example of using `shallowUntilTarget()` in this section.

docs/adding-a-page.md
kumar303: Yeah, this section is problematic but I don't know if you should spent time to rewrite it yet. As matt mentioned, maybe just link to a new-ish component as an example.

docs/testing.md
tofumatt: Here's my issue with that (I mentioned this in a previous review somewhere):

Implicit return functions are useful when it's a shorthand for a quick, lambda-like function who return value you want. I don't think implicit-return functions are a good shorthand to use when you aren't testing the return value of a function–in many cases you'd be using them and they wouldn't even have a return value or at least you wouldn't be using it. You're testing a side-effect of calling the function (e.g. throwing an exception) not the value the function returns. But seeing implicit returns in the code makes me scan for a test involving the returned value. So while:

```js
expect(sortAnArray([1, 3, 2])).toEqual([1, 2, 3]);
```
is fine, I don't think:
```js
expect(() => sortAnArray('notAnArray')).toThrow(/Not an array/);
```
is proper as you're using implicit return just to save on typing. You wouldn't write that test as:
```js
expect(() => {
  return sortAnArray('notAnArray');
}).toThrow(/Not an array/);
```
... you'd omit the return. So I don't like using them.

kumar303: This is so much more concise though 😁 

```js
expect(() => sortAnArray('notAnArray')).toThrow(/Not an array/);
```
It makes the test more readable.

kumar303: To be more clear about what I mean. In response to:

> Implicit return functions are useful when it's a shorthand for a quick, lambda-like function [whose] return value you want.

I think arrow functions serve a secondary purpose of conciseness which can really aid in readability.

tofumatt: Again though, I don't think it *is* more readable. Seeing an implicit return I read the return value as important but it's just a by-product of a shorter syntax. With the offending/error-throwing code on it's own line, without a return statement, I can see:

1. the return value is not important/relevant to that error being thrown
2. this is code that has a side-effect
3. we're testing the side-effect, as no data is being passed in any direction

I don't think concise === more readable.

kumar303: > Seeing an implicit return I read the return value as important

But you can just look at the context and immediately know that it's not. 

*shrug*

Anyway, this sentence needs fixing:

> It explicits what is expected to throw an error:

I'm not really sure what it meant to say so I don't have a good suggestion.

willdurand: > But you can just look at the context and immediately know that it's not.

While I also think brackets are overkill and it is not too complicated to read, I see a similarity with single-letter variable names here.

Example: `promise.catch((e) => handleError(e))`. Here, the context is self-explanatory I think. `e` is an error, yet writing `error` does not take too much time, and makes things crystal clear.

> I'm not really sure what it meant to say so I don't have a good suggestion.

What I want to say here is that using `() => {}` highlights the function that will throw an error (which is what we expect).

willdurand: I just rephrased the second sentence on this line.

PR #3018, opened 2017-08-29
Fetch addon on slug update

### LINKED ISSUE #3017: Accessing second search suggestion is not working as expected
Steps to reproduce:
1. Load Extensions homepage https://addons-dev.allizom.org/en-US/firefox/extensions/
2. Search for something i.e. "love"
3. Click on a search suggestion 
4. Search for something else i.e. "hate"
5. Click on a search suggestion

Expected results:
Corresponding add-on details page is loaded.

Actual results:
The page is continuously loading.

Notes/Issues:
- issue is reproducing in Themes page also

Verified on FF55(Win 7, Android 7.2.1). Issue is only reproducing on dev.
Video for this issue:
loading search res

#2747

willdurand:
Fix #3017

---

This PR updates an addon page whenever its slug changes.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -90,11 +90,16 @@ export class AddonBase extends React.Component {
     }
   }
 
-  componentWillReceiveProps({ addon: newAddon }) {
-    const { addon: oldAddon, dispatch } = this.props;
+  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
+    const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
+
     const oldAddonType = oldAddon ? oldAddon.type : null;
     if (newAddon && newAddon.type !== oldAddonType) {
       dispatch(setViewContext(newAddon.type));
+    }
+
+    if (params.slug !== newParams.slug) {
+      dispatch(fetchAddon({ slug: newParams.slug, errorHandler }));
     }
   }
 
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -200,6 +200,32 @@ describe('Addon', () => {
     expect(root.find(AddonMeta).prop('addon')).toEqual(null);
   });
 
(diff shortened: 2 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -90,12 +90,17 @@ export class AddonBase extends React.Component {
...
-    const { addon: oldAddon, dispatch } = this.props;
+  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
+    const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
+
     const oldAddonType = oldAddon ? oldAddon.type : null;
     if (newAddon && newAddon.type !== oldAddonType) {
       dispatch(setViewContext(newAddon.type));
     }
+
+    if (params.slug !== newParams.slug) {
```
tofumatt: I'm not sure if an extension can have the same slug as a theme, but just in case (and to at least render an error rather than not update state) we should check if the `addonType` OR the `slug` doesn't match, because right now `/extensions/foo/` -> `/themes/foo/` wouldn't cause a new dispatch.

tests/unit/amo/components/TestAddon.js
```diff
@@ -200,6 +200,32 @@ describe('Addon', () => {
     expect(root.find(AddonMeta).prop('addon')).toEqual(null);
   });
 
+  it('does not dispatch fetchAddon action when slug is the same', () => {
```
tofumatt: As mentioned above, I think we should have a test for "same slug; different addonType".

willdurand:
> I think this is good but I think we should also test for addonType, just so we don't ignore an extension->theme route change with the same URL.

From what I am seeing in the routing, we have `/addon/${slug}` so I dont think we can have the same slug for different addon types. I do not say we should not cover this case, but I am actually unable to use anything else than `params.slug` *before* having fetched an addon.

tofumatt:
Oh, wait, gosh, you're totally right.

Okay nevermind then this is actually fine.

Issue #3024, opened 2017-08-29
Make add-on reducer state more explicit


kumar303:
It's really tough to see what properties we actually store in the add-on reducer state. This has led to numerous bugs and has also made it tricky to add new add-on properties.

I'd like to start by restructuring the reducer a bit and getting rid of all spreads, carefully. This will be focused on preparing to solve the following bugs:

* Lack of platform support: https://github.com/mozilla/addons-frontend/issues/2998
* Incomplete theme objects: https://github.com/mozilla/addons-frontend/issues/3014

For QA: please test for regressions in AMO and the Discovery Pane for anything related to installing add-ons and themes.

willdurand:
I was feeling the same when thinking about new features to add. Semi-related, I have converted the addon code into a ducks module: https://github.com/willdurand/addons-frontend/commit/933f81cfd86ece933e6149186fe4a2f168c6b609.

PR #3025, opened 2017-08-29
Transform addons reducer/actions/action creators into a ducks module

willdurand:
Refs #3024 

---

Edit: fixed by rebase

~~Current build fails because of a locale test fixed by #3023:~~

```
Summary of all failing tests
 FAIL  tests/unit/TestLocalesConfig.js
  ● Locale Config › should have a corresponding ka entry the locale dir in the config

    expect(array).toContain(value)
    
    Expected array:
      ["af", "ar", "ast", "bg", "bn-BD", "ca", "cs", "da", "de", "dbl", "dbr", "dsb", "el", "en-GB", "en-US", "es", "eu", "fa", "fi", "fr", "fy-NL", "ga-IE", "he", "hsb", "hu", "id", "it", "ja", "kab", "ko", "mk", "mn", "ms", "nl", "nb-NO", "nn-NO", "pl", "pt-BR", "pt-PT", "ro", "ru", "sk", "sl", "sq", "sv-SE", "te", "th", "tr", "uk", "ur", "vi", "zh-CN", "zh-TW"]
    To contain value:
      "ka"
      
      at Object.<anonymous> (tests/unit/TestLocalesConfig.js:32:25)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
```


src/core/reducers/addons.js
```diff
@@ -1,8 +1,40 @@
+/* @flow */
 import { ADDON_TYPE_THEME } from 'core/constants';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { AddonType } from 'core/types/addons';
+
+
+export const FETCH_ADDON = 'FETCH_ADDON';
```
willdurand: Do we have a naming pattern for action types btw? poke @tofumatt

tofumatt: We have been inconsistent but we want `VERB_NOUN` so `FETCH_ADDON`, `LOAD_ADDON`, etc. makes sense.

willdurand: `FETCH_ADDON` to trigger the API call, and `LOAD_ADDON` to update the state with the API results, is that correct?

tofumatt: Yup, that's what we want. And when starting things like autocomplete or a search it would probably be `START_SEARCH`, `START_AUTOCOMPLETE`

willdurand: wonderful!

tests/unit/core/actions/test_addons.js
```diff
@@ -1,7 +1,8 @@
-import { fetchAddon } from 'core/actions/addons';
+import { fetchAddon } from 'core/reducers/addons';
 import { createStubErrorHandler } from 'tests/unit/helpers';
 
-describe('core.actions.addons', () => {
+
+describe(__filename, () => {
```
kumar303: I think this file should just be merged into [`tests/unit/core/reducers/test_addons.js`](https://github.com/willdurand/addons-frontend/blob/933f81cfd86ece933e6149186fe4a2f168c6b609/tests/unit/core/reducers/test_addons.js).

src/core/reducers/addons.js
```diff
@@ -22,8 +54,12 @@ export function denormalizeAddon(apiAddon) {
   return apiAddon;
 }
 
-export default function addon(state = initialState, action) {
+export default function addon(
+  state: Object = initialState,
+  action: Object = {}
```
kumar303: heh, yeah. I will adjust this when fixing the reducer. I'll also move `loadEntities` over here but only after bathing it in bleach and renaming it to `loadAddons` 💥

kumar303:
> Current build fails because of a locale test fixed by #3023

d'oh. I thought that failure was only in my build. It landed so you can merge in master.

willdurand:
Thanks for the review, this should be ok now.

Issue #2767, opened 2017-07-12
Add link to " More add-ons by these developers"


tofumatt:
Relies on the API here https://github.com/mozilla/addons-server/issues/5786, already complete.

Note the feature is labeled differently when there’s a [single dev](Note the feature is labeled differently when there’s a single dev or multiple devs.) or [multiple devs](https://addons.mozilla.org/en-GB/firefox/addon/wikipedia-panel/).

willdurand:
It seems it is not a link but a section/panel.

willdurand:
I have asked for an API update for this feature: https://github.com/mozilla/addons-server/issues/6285.

willdurand:
I am a bit confused by the Sketch file, I reused our `SearchResult` component to display the other addons but it does not look like the mock-ups. Yet, other parts of the app don't look like the mock-ups (cc @tofumatt @kumar303)

<img width="510" alt="screen shot 2017-09-06 at 17 25 41" src="https://user-images.githubusercontent.com/217628/30120225-87b8adcc-9328-11e7-9091-f18ebd142006.png">

tofumatt:
Hmm, that style will appear in an `AddonsCard` if you use `type="horizontal"`. That is admittedly not a great API, sorry about that 😢

https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/LandingAddonsCard/index.js#L35

willdurand:
Ah thanks, this is a nice one.. I was about to re-create the same thing.. So now I have it, but I still get this style. In the mock-ups, the icon should be on top of the text, not on the left.

<img width="490" alt="screen shot 2017-09-07 at 14 32 48" src="https://user-images.githubusercontent.com/217628/30163477-703e9754-93d9-11e7-941f-8d44cff4f92e.png">

Issue #2917, opened 2017-08-08
Create explicit API data mappings (i.e. replace normalizr) for AMO


kumar303:
We use magical schemas (via [normalizr](https://github.com/paularmstrong/normalizr)) to store API data in a mappable format. Sometimes this helps look up objects better but other times we have code that *unmaps* it entirely.

Instead of using `normalizr`, let's just map what we need explicitly in our reducers. It might require more code but the mapping logic will be a lot easier to follow. This is what our reducers are for!

---

Let's do this for AMO!

For QA: please make sure add-ons are loaded correctly and no information are missing (icons, names, titles, authors)

tofumatt:
Sounds good, normalizr is a kind of weird API for me anyway.

kumar303:
This is annoying to work around: https://github.com/paularmstrong/normalizr/issues/290

kumar303:
Grep the code for this issue number. I'm adding some TODOs.

ioanarusiczki:
Looks good to me. Verified on AMO dev with FF62-Win10

PR #3027, opened 2017-08-30
Add new addonsByAuthors reducer/saga to retrieve other add-ons by these authors

willdurand:
Refs #2767

---

This PR adds a new reducer (`addonsByAuthors`) to deal with other add-ons by developers (see #2767). ~~It would make sense to put it into the `addons` reducer but I think we should not touch it anymore until #3024 is resolved.~~ That is why I created a new reducer.

There is also a saga to retrieve add-ons based on a set of authors. I set the number of results to retrieve for each author to `6` because it is the number of add-ons displayed [on the existing website](https://addons.mozilla.org/en-US/firefox/addon/english-gb-language-pack/). It is configurable in the reducer though.

I reused the same principle as in the [`reviews` reducer](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/reviews.js) and exposed a `byAddon` map.

It can be used as follows:

<details>
<summary>patch</summary>

``` diff
diff --git a/src/amo/components/Addon/index.js b/src/amo/components/Addon/index.js
index 45d468d..8b60362 100644
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -13,6 +13,7 @@ import NotFound from 'amo/components/ErrorPage/NotFound';
 import DefaultRatingManager from 'amo/components/RatingManager';
 import ScreenShots from 'amo/components/ScreenShots';
+import { fetchOtherAddonsByAuthors } from 'amo/reducers/addonsByAuthors';
 import { fetchAddon } from 'core/reducers/addons';
 import { withErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
@@ -57,6 +58,7 @@ export class AddonBase extends React.Component {
     installStatus: PropTypes.string.isRequired,
     toggleThemePreview: PropTypes.func.isRequired,
     userAgentInfo: PropTypes.object.isRequired,
+    addonsByAuthors: PropTypes.object.isRequired,
   }
 
   static defaultProps = {
@@ -69,12 +71,19 @@ export class AddonBase extends React.Component {
 
     if (addon) {
       dispatch(setViewContext(addon.type));
+      dispatch(fetchOtherAddonsByAuthors({
+        addonType: addon.type,
+        authors: addon.authors.map((author) => author.name),
+        slug: addon.slug,
+      }));
     } else {
       dispatch(fetchAddon({ slug: params.slug, errorHandler }));
     }
   }
 
-  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
+    const { addon: newAddon, params: newParams } = nextProps;
     const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
 
     const oldAddonType = oldAddon ? oldAddon.type : null;
@@ -85,6 +94,16 @@ export class AddonBase extends React.Component {
     if (params.slug !== newParams.slug) {
       dispatch(fetchAddon({ slug: newParams.slug, errorHandler }));
     }
+
+    // TODO: also check when addonsByAuthors is not empty to avoid extra API calls
+    if (newAddon && oldAddon !== newAddon) {
+      dispatch(fetchOtherAddonsByAuthors({
+        addonType: newAddon.type,
+        authors: newAddon.authors.map((author) => author.name),
+        slug: newAddon.slug,
+      }));
+    }
   }
 
   componentWillUnmount() {
@@ -408,9 +427,13 @@ export class AddonBase extends React.Component {
+
   let installedAddon = {};
+  let addonsByAuthors = {};
+
   if (addon) {
     installedAddon = state.installations[addon.guid] || {};
+    addonsByAuthors = state.addonsByAuthors.byAddon[addon.slug] || {};
   }
 
   return {
@@ -431,6 +454,7 @@ export function mapStateToProps(state, ownProps) {
     installStatus: installedAddon.status || UNKNOWN,
     clientApp: state.api.clientApp,
     userAgentInfo: state.api.userAgentInfo,
+    addonsByAuthors,
   };
 }
```

</details>

### DIFF
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -1,0 +1,135 @@
+/* @flow */
+import type { AddonType } from 'core/types/addons';
+
+
+type State = {
+  byAddonSlug: { [string]: AddonType },
+};
+
+export const initialState: State = {
+  byAddonSlug: {},
+};
+
+export const OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE = 6;
+
+// For further information about this notation, see:
+// https://github.com/mozilla/addons-frontend/pull/3027#discussion_r137661289
+export const FETCH_OTHER_ADDONS_BY_AUTHORS: 'FETCH_OTHER_ADDONS_BY_AUTHORS'
+  = 'FETCH_OTHER_ADDONS_BY_AUTHORS';
+export const LOAD_OTHER_ADDONS_BY_AUTHORS: 'LOAD_OTHER_ADDONS_BY_AUTHORS'
+  = 'LOAD_OTHER_ADDONS_BY_AUTHORS';
+
+type FetchOtherAddonsByAuthorsParams = {|
+  addonType: string,
+  authors: Array<string>,
+  errorHandlerId: string,
+  slug: string,
+|};
+
+type FetchOtherAddonsByAuthorsAction = {|
+  type: typeof FETCH_OTHER_ADDONS_BY_AUTHORS,
+  payload: FetchOtherAddonsByAuthorsParams,
+|};
+
+export const fetchOtherAddonsByAuthors = (
+  { addonType, authors, errorHandlerId, slug }: FetchOtherAddonsByAuthorsParams
+): FetchOtherAddonsByAuthorsAction => {
+  if (!errorHandlerId) {
+    throw new Error('An errorHandlerId is required');
+  }
+
+  if (!slug) {
+    throw new Error('An add-on slug is required.');
+  }
+
+  if (!addonType) {
+    throw new Error('An add-on type is required.');
+  }
+
+  if (!authors) {
+    throw new Error('Authors are required.');
+  }
+
+  if (!Array.isArray(authors)) {
+    throw new Error('The authors parameter must be an array.');
+  }
+
+  return {
+    type: FETCH_OTHER_ADDONS_BY_AUTHORS,
+    payload: {
+      addonType,
+      authors,
+      errorHandlerId,
+      slug,
+    },
+  };
+};
+
+type LoadOtherAddonsByAuthorsParams = {|
+  slug: string,
+  addons: Array<AddonType>,
+|};
+
+type LoadOtherAddonsByAuthorsAction = {|
+  type: typeof LOAD_OTHER_ADDONS_BY_AUTHORS,
+  payload: LoadOtherAddonsByAuthorsParams,
+|};
+
+export const loadOtherAddonsByAuthors = (
+  { addons, slug }: LoadOtherAddonsByAuthorsParams
+): LoadOtherAddonsByAuthorsAction => {
+  if (!slug) {
+    throw new Error('An add-on slug is required.');
+  }
+
+  if (!addons) {
+    throw new Error('A set of add-ons is required.');
+  }
+
+  return {
+    type: LOAD_OTHER_ADDONS_BY_AUTHORS,
+    payload: { slug, addons },
+  };
+};
+
+type Action =
+  | FetchOtherAddonsByAuthorsAction
+  | LoadOtherAddonsByAuthorsAction;
+
+const reducer = (
+  state: State = initialState,
+  action: Action
+): State => {
+  switch (action.type) {
+    case FETCH_OTHER_ADDONS_BY_AUTHORS: {
+      const payload = action.payload;
+
+      return {
+        ...state,
+        byAddonSlug: {
+          ...state.byAddonSlug,
+          [payload.slug]: undefined,
+        },
+      };
+    }
+    case LOAD_OTHER_ADDONS_BY_AUTHORS: {
+      const payload = action.payload;
+
+      return {
+        ...state,
+        byAddonSlug: {
+          ...state.byAddonSlug,
+          [payload.slug]: payload.addons
+            // This ensures we do not display the main add-on in the list of
+            // "add-ons by these authors".
+            .filter((addon) => addon.slug !== payload.slug)
+            .slice(0, OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE),
+        },
+      };
+    }
+    default:
+      return state;
+  }
+};
+
+export default reducer;
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -101,32 +101,26 @@ const reducer = (
   action: Action
 ): State => {
   switch (action.type) {
-    case FETCH_OTHER_ADDONS_BY_AUTHORS: {
-      const payload = action.payload;
-
+    case FETCH_OTHER_ADDONS_BY_AUTHORS:
       return {
         ...state,
         byAddonSlug: {
           ...state.byAddonSlug,
-          [payload.slug]: undefined,
+          [action.payload.slug]: undefined,
         },
       };
-    }
-    case LOAD_OTHER_ADDONS_BY_AUTHORS: {
-      const payload = action.payload;
-
+    case LOAD_OTHER_ADDONS_BY_AUTHORS:
       return {
         ...state,
         byAddonSlug: {
           ...state.byAddonSlug,
-          [payload.slug]: payload.addons
+          [action.payload.slug]: action.payload.addons
             // This ensures we do not display the main add-on in the list of
             // "add-ons by these authors".
-            .filter((addon) => addon.slug !== payload.slug)
+            .filter((addon) => addon.slug !== action.payload.slug)
             .slice(0, OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE),
         },
       };
-    }
     default:
       return state;
   }
--- a/src/amo/sagas/addonsByAuthors.js
+++ b/src/amo/sagas/addonsByAuthors.js
@@ -1,0 +1,47 @@
+import { call, put, select, takeLatest } from 'redux-saga/effects';
+import { SEARCH_SORT_POPULAR } from 'core/constants';
+import {
+  FETCH_OTHER_ADDONS_BY_AUTHORS,
+  OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE,
+  loadOtherAddonsByAuthors,
+} from 'amo/reducers/addonsByAuthors';
+import { search as searchApi } from 'core/api/search';
+import log from 'core/logger';
+import { createErrorHandler, getState } from 'core/sagas/utils';
+
+
+export function* fetchOtherAddonsByAuthors({ payload }) {
+  const { errorHandlerId, authors, slug, addonType } = payload;
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
+
+    const response = yield call(searchApi, {
+      api: state.api,
+      filters: {
+        addonType,
+        author: authors.join(','),
+        // We need one more add-on than the number to display because the API
+        // may return the main add-on and we cannot tell the API to exclude it.
+        page_size: OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE + 1,
+        sort: SEARCH_SORT_POPULAR,
+      },
+    });
+
+    // TODO: remove the line below and pass `response.addons` directly once
+    // https://github.com/mozilla/addons-frontend/issues/2917 is done.
+    const addons = Object.values(response.entities.addons || {});
+
+    yield put(loadOtherAddonsByAuthors({ addons, slug }));
+  } catch (error) {
+    log.warn(`Search for addons by authors results failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* addonsByAuthorsSaga() {
+  yield takeLatest(FETCH_OTHER_ADDONS_BY_AUTHORS, fetchOtherAddonsByAuthors);
+}
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -10,7 +10,6 @@ import config from 'config';
 
 import { initialApiState } from 'core/reducers/api';
 import log from 'core/logger';
-import searchApi from 'core/api/search';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import type { ApiStateType } from 'core/reducers/api';
@@ -268,5 +267,3 @@ export function autocomplete({ api, filters }: AutocompleteParams) {
     state: api,
   });
 }
-
-export const search = searchApi;
--- a/tests/unit/amo/reducers/test_addonsByAuthors.js
+++ b/tests/unit/amo/reducers/test_addonsByAuthors.js
@@ -15,7 +15,7 @@ describe(__filename, () => {
       expect(state).toEqual(initialState);
     });
 
-    it('ignore unrelated actions', () => {
+    it('ignores unrelated actions', () => {
       // Load some initial state to be sure that an unrelated action does not
       // change it.
       const state = reducer(undefined, loadOtherAddonsByAuthors({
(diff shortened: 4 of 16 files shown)


src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,60 @@
...
+export function* fetchOtherAddonsByAuthors({ payload }) {
+  const { errorHandlerId, authors, slug } = payload;
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
+
+    const responses = yield all(authors.map((author) => call(searchApi, {
```
kumar303: we could probably ask for an API patch to allow filtering by a list of authors

src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,60 @@
...
+      api: state.api,
+      filters: { author, page_size: NB_ADDONS_MAX_PER_AUTHOR },
```
kumar303: This should probably [sort by hotness](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#addon-search-sort) or something

src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,60 @@
...
+    })));
+
+    // This computes a map of add-ons based on the N previous API calls (N
+    // being the number of authors). It eliminates duplicates (_.e.g._ 3 people
+    // authored the same add-on results in 3 times the same add-on). It also
+    // omits the add-on (identified by `slug`) that is related to these other
+    // add-ons, so that we will not list the same add-on already displayed in
+    // the detail view.
+    const addons = responses.reduce((previousAddons, response) => {
```
kumar303: I don't think this is worth the effort and complexity (and added test coverage). The reducer will actually do this naturally because it is creating a map of add-ons by ID (slug). Also, if we get an API patch for filtering by multiple authors then we won't need it at all.

src/amo/sagas/addonsByAuthors.js
willdurand: Asked here: https://github.com/mozilla/addons-server/issues/6285.

src/amo/sagas/addonsByAuthors.js
willdurand: Of course yes!

src/amo/sagas/addonsByAuthors.js
willdurand: If we dont get the filter for a list of authors, then I am afraid we are forced to do that. We need a distinct set of add-ons and I think the saga should handle this.

willdurand:
> Hmm, I kinda like this as a separate reducer/saga.

Thanks. Okay so I continue this way then. If we can get a new filter param for the authors, then I can get rid of the multiple API calls in the saga, but the rest will remain the same (that is why I think it makes sense to compute the result set into the saga).

> I think once we fix up the add-ons reducer we could adjust this one to simply keep a map of add-on ID to array of other add-on IDs (or slugs).

Yes, and it could be used elsewhere maybe.

src/amo/sagas/addonsByAuthors.js
kumar303: Oh, wait, I didn't read the reducer code close enough. I thought it was storing a *map* of add-ons but I see it's storing an array of them. In any case, I find that this sort of code should live in the reducer because that's its job anyway.

willdurand:
I have updated this PR.  The second commit adds support for the `addonType` because the mockups show that (it is "More extensions by devs" and "More themes by artists", see: https://github.com/mozilla/addons-frontend/issues/2767#issuecomment-325648856).

⚠️  be sure to use the `dev` API to test this PR, since it uses the updated `author` query parameter: https://github.com/mozilla/addons-server/pull/6289.

willdurand:
The API needs one more change: https://github.com/mozilla/addons-server/issues/6346. I can update the code in #2994 (which depends on this PR).

src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,45 @@
...
+
+  try {
+    const state = yield select(getState);
+
+    const response = yield call(searchApi, {
+      api: state.api,
+      filters: {
+        addonType,
+        author: authors.join(','),
+        // We need one more add-on than the number to display because the API
```
tofumatt: Is there a maximum number of add-ons the API will return, even if `page_size` is set quite large? It seems like this could require pagination...

src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,45 @@
...
+        // may return the main add-on and we cannot tell the API to exclude it.
+        page_size: NB_OTHERS_ADDONS_BY_AUTHORS + 1,
+        sort: SEARCH_SORT_POPULAR,
+      },
+    });
+
+    const addons = Object.values(response.entities.addons || {});
```
tofumatt: This seems like something the reducer should be doing, not the saga.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -0,0 +1,128 @@
...
+      const state = reducer(undefined);
+      expect(state).toEqual(initialState);
+    });
+
+    it('ignore unrelated actions', () => {
+      const state = {
+        ...initialState,
+        // We slightly modify the state to be sure that an unrelated action
+        // does not change it.
+        byAddon: {
```
tofumatt: I'm sort of confused by this name `byAddon`. What is it supposed to mean?

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -0,0 +1,112 @@
...
+    }));
+  }
+
+  it('calls the API to retrieve other add-ons', async () => {
+    const addons = [fakeAddon];
+    const slug = fakeAddon.slug;
+    const authors = ['mozilla', 'johnedoe'];
+
+    mockSearchApi
+      // We use `default` because `search` is the default exported function in
```
tofumatt: Oh, that sucks. You *can* import it from `core/api` but otherwise we should fix this... should we export it twice or something? This is pretty inelegant-looking.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+export const initialState: State = {
+  byAddon: {},
+};
+
+export const NB_OTHERS_ADDONS_BY_AUTHORS = 6;
+
+export const FETCH_OTHER_ADDONS_BY_AUTHORS = 'FETCH_OTHER_ADDONS_BY_AUTHORS';
+const LOAD_OTHER_ADDONS_BY_AUTHORS = 'LOAD_OTHER_ADDONS_BY_AUTHORS';
```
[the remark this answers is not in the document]
willdurand: Ideally we should not export these types, but yes let's be consistent.

src/amo/sagas/addonsByAuthors.js
willdurand: Not sure to understand. This would be a different feature to me. Here we just want to show a few other add-ons on the detail page. Displaying all the add-ons of a given author would rather use the search feature/UI.

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
tofumatt: If we wanted to stop doing that (I think we tend only to use them in sagas aside from tests) we could start using `fetchCategoriesAction().type` over `FETCH_CATEGORIES`.

src/amo/sagas/addonsByAuthors.js
tofumatt: Ohhhhhhhh. Okay then yes, totally. I understand now. Ignore my pagination comments then, I'm daft.

src/amo/sagas/addonsByAuthors.js
willdurand: Then it is not clear to me what goes in saga and what not.

For instance here, without Normalizr I would pass `response.addons` and I would not have to change any line of code in the reducer. Yet we have this weird Normalizr layer that changes the response payload, transforms a simple array into a hash map, and does not return anything when array is empty (reported by @kumar303 yesterday).

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
willdurand: I'd like to do that

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: I borrowed this in another reducer, review maybe. The idea is to normalize the state as described here: http://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html.

Here I can ask the state: "give me other addons for the authors of this addon (slug)".

src/amo/sagas/addonsByAuthors.js
tofumatt: Oh, right. Okay, as long as we have normalizr (which, I agree, is annoying) this makes sense. I guess we could add a TODO here that this can be removed once we ditch normalizr (#2917).

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,120 @@
...
+
+type FetchOtherAddonsByAuthorsParams = {|
+  errorHandlerId: string,
+  addonType: string,
+  authors: Array<string>,
+  slug: string,
+|};
+
+type FetchOtherAddonsByAuthorsAction = {|
+  type: 'FETCH_OTHER_ADDONS_BY_AUTHORS',
```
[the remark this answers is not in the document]
kumar303: It's tricky for Flow to figure out the `const` value because any dynamic JS expression can be used to set the value of the `const`.

However, there is a way to at least keep the repetition in one place:

```js
export const 
  // This declares the `const` as a literal type, not a string type.
  FETCH_OTHER_ADDONS_BY_AUTHORS: 'FETCH_OTHER_ADDONS_BY_AUTHORS'
  // This sets the actual JS value.
  = 'FETCH_OTHER_ADDONS_BY_AUTHORS';

type FetchOtherAddonsByAuthorsAction = {|
  // Now you can reference its type.
  type: typeof FETCH_OTHER_ADDONS_BY_AUTHORS,
|}
```

However, if we had full Flow coverage then I don't see why we would need to define a constant at all. We could just use the string and Flow would warn us when values did not match up. The main reason we use `const` values now is because the linter will warn us about invalid imports. Since we won't reach full Flow coverage we should probably keep doing that.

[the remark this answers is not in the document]
kumar303: Here's another approach of creating a typed reducer without repetition: https://gist.github.com/rendro/4c9f4ddea8133bbddc3df427a8f62b9e

tests/unit/amo/sagas/test_addonsByAuthors.js
willdurand: For some reasons, I could not mock the `core/api` ... Maybe because I also require `core/api/search` in the code.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -0,0 +1,129 @@
...
+        errorHandlerId: 'any-error-handler-id',
+        slug: 'addon-slug',
+        authors,
+        addonType: ADDON_TYPE_THEME,
+      });
+      expect(payload.authors).toEqual(authors);
+    });
+
+    it('transforms a single author into an array of authors', () => {
+      const authors = 'john';
```
kumar303: Ah, this answers my question for why you needed `concat()`. I think we can live without this shortcut since it's not overly verbose to type `authors = ['john']`. This kind of optimization is typical in JavaScript but it's only good for introducing bugs IMHO :/

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -0,0 +1,129 @@
...
+        fetchOtherAddonsByAuthors({
+          errorHandlerId: 'id',
+          slug: 'addon-slug',
+          type: ADDON_TYPE_THEME,
+          addonType: ADDON_TYPE_THEME,
+        });
+      }).toThrow(/Authors are required/);
+    });
+
+    it('returns an array of authors in payload', () => {
```
kumar303: I don't think this kind of test is very useful. You will be covering this logic in the saga test.

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -0,0 +1,108 @@
...
+    }));
+  }
+
+  it('calls the API to retrieve other add-ons', async () => {
+    const addons = [fakeAddon];
+    const slug = fakeAddon.slug;
+    const authors = ['mozilla', 'johnedoe'];
+
+    mockApi
+      .expects('search')
```
kumar303: This should add some expectations about the arguments given to `search()`. This will add better coverage for the action creator payload and how the saga handles it.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,122 @@
...
+
+type State = {
+  byAddon: { [string]: AddonType },
+};
+
+export const initialState: State = {
+  byAddon: {},
+};
+
+export const OTHERS_ADDONS_BY_AUTHORS_TO_LOAD = 6;
```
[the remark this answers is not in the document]
willdurand: I grep'ed the code for `page_size` and found `FEATURED_ADDONS_TO_LOAD`. I like `FEATURED_ADDONS_TO_LOAD` though.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,122 @@
...
+): State {
+  const payload: Object = action.payload || {};
+
+  switch (action.type) {
+    case FETCH_OTHER_ADDONS_BY_AUTHORS:
+      return state;
+    case LOAD_OTHER_ADDONS_BY_AUTHORS:
+      return {
+        ...state,
+        byAddon: {
```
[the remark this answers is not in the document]
willdurand: I am fine with `byAddonSlug`, thanks!

src/core/api/index.js
```diff
@@ -10,7 +10,7 @@ import config from 'config';
 
 import { initialApiState } from 'core/reducers/api';
 import log from 'core/logger';
-import searchApi from 'core/api/search';
+import { search as searchApi } from 'core/api/search';
```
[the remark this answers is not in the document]
willdurand: Yup, that works for me

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: Ok, sounds good to me

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
kumar303: Sure, no biggie. However, it should be `OTHER_ADDONS_BY_AUTHORS_TO_LOAD`, right?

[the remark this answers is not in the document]
willdurand: indeed!

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,122 @@
...
+  authors: Array<string>,
+  slug: string,
+|};
+
+type FetchOtherAddonsByAuthorsAction = {|
+  type: typeof FETCH_OTHER_ADDONS_BY_AUTHORS,
+  payload: FetchOtherAddonsByAuthorsParams,
+|};
+
+export const fetchOtherAddonsByAuthors = (
```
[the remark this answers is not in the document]
willdurand: As discussed on IRC, let's use `=>` for all functions.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -0,0 +1,126 @@
...
+
+const reducer = (
+  state: State = initialState,
+  action: Action
+): State => {
+  const payload: Object = action.payload || {};
+
+  switch (action.type) {
+    case FETCH_OTHER_ADDONS_BY_AUTHORS:
+      return state;
```
kumar303: I think this should reset the loaded add-ons in case there are any:

```js
return {
  ...state,
  byAddonSlug: {
    ...state.byAddonSlug,
    [payload.slug]: undefined,
  },
};
```

willdurand:
Eeeerf, fixed the last batch of issues.. I'll review this PR myself before requesting a review again.

src/amo/sagas/addonsByAuthors.js
```diff
@@ -0,0 +1,47 @@
...
+    const state = yield select(getState);
+
+    const response = yield call(searchApi, {
+      api: state.api,
+      filters: {
+        addonType,
+        author: authors.join(','),
+        // We need one more add-on than the number to display because the API
+        // may return the main add-on and we cannot tell the API to exclude it.
+        page_size: OTHERS_ADDONS_BY_AUTHORS_TO_LOAD + 1,
```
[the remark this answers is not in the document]
willdurand: Follow-up: https://github.com/mozilla/addons-server/issues/6376

willdurand:
Should be ready now.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -0,0 +1,170 @@
...
+          // This is the case where there are more add-ons loaded than needed.
+          addons: Array(OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE + 2).fill(fakeAddon),
+        }));
+        expect(state.byAddonSlug[slug])
+          .toHaveLength(OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE);
+      });
+    });
+
+    describe('fetchOtherAddonsByAuthors()', () => {
+      it('resets the loaded add-ons', () => {
```
kumar303: This should be added to the existing `fetchOtherAddonsByAuthors` describe block

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -0,0 +1,170 @@
...
+      // change it.
+      const state = reducer(undefined, loadOtherAddonsByAuthors({
+        slug: fakeAddon.slug,
+        addons: [fakeAddon],
+      }));
+      const newState = reducer(state, { type: 'UNRELATED' });
+      expect(newState).toEqual(state);
+    });
+
+    describe('loadOtherAddonsByAuthors()', () => {
```
kumar303: I think all of these can be merged into the other `loadOtherAddonsByAuthors` describe block

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -0,0 +1,133 @@
...
+    mockApi.verify();
+
+    const loadAction = sagaTester.getCalledActions()[2];
+    expect(loadAction).toEqual(expectedLoadAction);
+  });
+
+  it('clears the error handler', async () => {
+    _fetchOtherAddonsByAuthors({ authors: [], slug: fakeAddon.slug });
+
+    await sagaTester.waitFor(CLEAR_ERROR);
```
kumar303: This might be a bit more future proof as `sagaTester.waitFor(errorHandler.createClearingAction().type)` (although you probably want to create a variable so you can use it in the assertion).

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: Correct. I wanted to separate reducer and action creators, but now they have the same name so.. it makes sense to merge them. Thanks!

PR #3029, opened 2017-08-30
Remove confusing/redundant title attribute on search button

### LINKED ISSUE #3028: Search button title is not clear
Translators [do not really understand the purpose/meaning of the title of the (hidden) search `button`](https://github.com/mozilla/addons-frontend/commit/9fb119bf608a649d0ecad0c5de6d0e8401327918). It was not translated before, that is why it has only been revealed recently.

~~I propose to update it from `Enter` to `Search`.~~

I propose to remove this `title` attribute, the button text should be enough.

willdurand:
Fix #3028 

---

Because there is already a textual content on this button, I do not think the `title` attribute adds anything meaningful.

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -213,7 +213,6 @@ export class SearchFormBase extends React.Component {
           className="visually-hidden"
           onClick={this.handleSearch}
           ref={(ref) => { this.submitButton = ref; }}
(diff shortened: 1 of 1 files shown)


willdurand:
The job for the selenium tests is failing, and I am not sure to understand why: https://travis-ci.org/mozilla/addons-frontend/jobs/270016781. I checked and it seems the button `title` isnt used in these tests..

PR #3033, opened 2017-08-30
Use yarn in Dockerfile and functional test command

### LINKED ISSUE #1391: Deploy dependencies to production with yarn

willdurand:
Fix #1391

---

This PR fixes an issue with the functional test suite on Travis: the Docker image could not be built.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,7 @@ RUN buildDeps=' \
     # install deps
     apt-get update -y && \
     apt-get install -y --no-install-recommends $buildDeps && \
-	npm update -g npm@3 && \
-	npm install && npm cache clean && \
+    yarn install && \
     # cleanup
     # apt-get purge -y $buildDeps && \
     rm -rf /var/lib/apt/lists/*
@@ -31,4 +30,4 @@ RUN ln -s /srv/node/node_modules
 ENV SERVER_HOST 0.0.0.0
 ENV SERVER_PORT 4000
 
-CMD npm start
+CMD yarn start
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,16 @@ WORKDIR /srv/node
 
 RUN buildDeps=' \
     git \
+    yarn \
     ' && \
+    # `apt-transport-https` is required to use https deb repositories
+    apt-get update -y && \
+    apt-get install -y --no-install-recommends apt-transport-https && \
+    # configure Yarn repository, see: https://yarnpkg.com/en/docs/install#linux-tab
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg|apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
+    # the base image installs yarn, let's be sure we use ours
+    rm -f /usr/local/bin/yarn /usr/local/bin/yarnpkg && \
     # install deps
     apt-get update -y && \
     apt-get install -y --no-install-recommends $buildDeps && \
--- a/bin/start-func-test-server.js
+++ b/bin/start-func-test-server.js
@@ -125,7 +125,7 @@ new Promise((resolve) => {
       // This will make sure we can read the logs.
       '--log-driver=json-file',
       imageId,
-      '/bin/sh -c "npm run build && npm run start"',
+      '/bin/sh -c "yarn build && yarn start"',
     ];
     return exec('docker', runArgs);
   })


kumar303:
> jason: +1 to switching to yarn

ops says yes

Their custom start command already uses it, heh. https://github.com/mozilla-services/cloudops-deployment/pull/1067/files

willdurand:
> Their custom start command already uses it, heh. https://github.com/mozilla-services/cloudops-deployment/pull/1067/files

I dont have acces to this file :(

kumar303:
It's puppet config to start the same Docker container with a custom command. I'm not really sure why it's needed but it's already in their pipeline so *shrug*.

screenshot 2017-08-30 12 21 13

willdurand:
I have updated the `Dockerfile` to use `yarn` from the `apt` repo.

Dockerfile
```diff
@@ -7,12 +7,20 @@ WORKDIR /srv/node
 
 RUN buildDeps=' \
     git \
+    yarn \
     ' && \
+    # `apt-transport-https` is required to use https deb repositories
+    apt-get update -y && \
+    apt-get install -y --no-install-recommends apt-transport-https && \
```
kumar303: `apt-transport-https` should be moved to the `buildDeps` list up above. After that, you won't need this duplicate `apt-get install` command. Unless ... does this need to be installed before yarn?

Dockerfile
```diff
@@ -7,12 +7,20 @@ WORKDIR /srv/node
 
 RUN buildDeps=' \
     git \
+    yarn \
     ' && \
+    # `apt-transport-https` is required to use https deb repositories
+    apt-get update -y && \
```
kumar303: This already happens down below. Did you mean to delete the other one?

Dockerfile
```diff
@@ -7,12 +7,20 @@ WORKDIR /srv/node
...
+    apt-get install -y --no-install-recommends apt-transport-https && \
+    # configure Yarn repository, see: https://yarnpkg.com/en/docs/install#linux-tab
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg|apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
+    # the base image installs yarn, let's be sure we use ours
+    rm -f /usr/local/bin/yarn /usr/local/bin/yarnpkg && \
```
kumar303: Is this really necessary? Doesn't `apt-get install yarn` just do what we want?

Dockerfile
willdurand: that was not a mistake. I need to fetch `apt-transport-https` and install it first. otherwise I cannot do anything because the `https://` dep repo will not be supported. So, I need to install it separately, hence these redundant commands.

Dockerfile
willdurand: I am not sure how `PATH` is configured, and I dont trust `update-alternatives` to take care of it. That would be safer in my opinion.

willdurand:
IRC logs:

```
[19:37:59]  <@kumar>	willdurand: do you need to call apt-get update -y twice? Is it because the new source was added? https://github.com/mozilla/addons-frontend/pull/3033#discussion_r136136984
[19:38:46]  <willdurand>	kumar: yes, one for transport-https, one for yarn
```

```
<@kumar>	willdurand: and for this one was it just a precaution or was it using the wrong version? https://github.com/mozilla/addons-frontend/pull/3033/files#r136137907
[19:41:22]  <willdurand>	precaution
[19:41:33]  <willdurand>	but it would be unfortunate not to use the right one
[19:42:25]  <@kumar>	willdurand: it feels a little hacky because I'm not sure if the paths will be right. I'm pretty sure apt-get install will do the right thing. I would rather see it removed for now until we run into issues.
[19:42:42]  <willdurand>	I took the paths from the main dockerfile
[19:42:52]  <@kumar>	right, but they may change, I dunno
[19:43:20]  <willdurand>	yeah. that would still be harmless though (-f)
[19:44:24]  <@kumar>	willdurand: would apt-get ever put the new one there? would we be rm'ing the new one accidentally ever?
[19:45:01]  <willdurand>	it should be rm'ed before apt-get install no?
[19:45:23]  <@kumar>	oh, duh, right
[19:46:02]  <willdurand>	but anyway it's up to you really. I think it is safer that way, but I am fine not removing the existing ones
[19:46:38]  <@kumar>	ok, I don't think we'll need it but I suppose it's fine.
```

PR #3041, opened 2017-08-31
Use local Yarnpkg GPG key in Docker

### LINKED ISSUE #3037: Add yarn gpg key to the repository instead of fetching it each time
Let's fix up the Dockerfile after https://github.com/mozilla/addons-frontend/pull/3033 so that we use a yarn gpg key that has been committed to the `addons-frontend` repo instead of fetching it from the Internet each time. 

We can follow the pattern here: https://github.com/mozilla/addons-server/blob/master/Dockerfile#L6-L13

willdurand:
Fix #3037

---

This PR changes the way we install the yarnpkg GPG key to fetch `yarn` from their deb repository.


Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN mkdir -p /srv/node
 ADD package.json /srv/node/
 WORKDIR /srv/node
 
+COPY docker/yarnpkg.gpg.key /tmp/yarnpkg.gpg.key
```
kumar303: Could you add a comment saying that this lives at https://dl.yarnpkg.com/debian/pubkey.gpg ? It would be bizarre if we ever had to update it so I don't think it's a problem that we are making a local copy.

kumar303: I think it would make sense to put the key somewhere long-lived rather than `tmp`. Also, I find it better for sanity if the local `docker` directory mirrors the server's filesystem. I suggest this:

```
COPY docker/etc/pki/yarnpkg.gpg.key /etc/pki/yarnpkg.gpg.key
```

willdurand: It is only useful once, at docker build-time, then it is in `apt`. That's why I put it in `/tmp`, but I can change that.

willdurand:
updated!

PR #3048, opened 2017-08-31
Configure codecov

### LINKED ISSUE #3012: Coveralls failing causes the tests to fail
https://travis-ci.org/mozilla/addons-frontend/jobs/268732728 failed because coveralls ran into a server error. We should possibly run coveralls in a separate task, if possible, so we can tell it's failing more obviously.

muffinresearch:
Fixes #3012 

Ok [looks good](https://codecov.io/gh/mozilla/addons-frontend) - this is ready for an r?


muffinresearch:
> Are these instructions for checking coverage locally still the same?

Yep, that's part of Jest and independent of this.

tofumatt:
https://github.com/mozilla/addons-frontend/blob/master/docs/testing.md Will need updating to not mention coveralls though.

muffinresearch:
> https://github.com/mozilla/addons-frontend/blob/master/docs/testing.md Will need updating to not mention coveralls though.

Good catch, that's fixed now.

Issue #2760, opened 2017-07-12
Add release notes on add-on details


tofumatt:
If release notes are available we should show them. May need a new section, probably at the top.

kumar303:
We have release notes: https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail-object

kumar303:
From `AMO Desktop - Master.sketch` :

<img width="486" alt="screenshot 2017-08-17 16 28 04" src="https://user-images.githubusercontent.com/55398/29434585-30823a92-8369-11e7-9728-be4296cf7be7.png">

kumar303:
Hmm, we may want to wait on https://github.com/mozilla/addons-server/issues/6268 otherwise I'd have to create a version saga (which do we need anyway, eventually).

ValentinaPC:
Verified this on AMO-dev FF55(Win 7).
Version notes are listed for each version.
Postfix screenshot:
![rel_notes](https://user-images.githubusercontent.com/15685960/30015280-affb7b6a-9158-11e7-8e48-698253a1fb73.png)

@kumar303 : Is there anywhere else where this notes should be displayed?

ValentinaPC:
Found it! :P
Verified as fixed on AMO-dev FF55(Win 7, Android 7.2.1).
Release notes are listed in new desktop pages for current add-on version.
Postfix screenshots:
![release_notes-last_vers](https://user-images.githubusercontent.com/15685960/30107954-12282192-9309-11e7-9059-90892b33d555.png)
![device](https://user-images.githubusercontent.com/15685960/30107958-152e26f2-9309-11e7-9a77-b887d2a49ba3.png)

kumar303:
oh yeah, sorry, I should have included a URL

PR #3050, opened 2017-08-31
Show version release notes

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2760

Here are some examples using https://addons-dev.allizom.org/en-US/firefox/addon/fastest-search/

screenshot 2017-08-31 16 30 35
screenshot 2017-08-31 16 32 28

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -22,8 +22,8 @@ import {
 import { withInstallHelpers } from 'core/installAddon';
 import {
   getClientCompatibility as _getClientCompatibility,
-  nl2br,
   sanitizeHTML,
+  sanitizeUserHTML,
 } from 'core/utils';
 import { getAddonIconUrl } from 'core/imageUtils';
 import translate from 'core/i18n/translate';
@@ -37,22 +37,6 @@ import Badge from 'ui/components/Badge';
 
 import './styles.scss';
 
-
-export const allowedDescriptionTags = [
-  'a',
-  'abbr',
-  'acronym',
-  'b',
-  'blockquote',
-  'br',
-  'code',
-  'em',
-  'i',
-  'li',
-  'ol',
-  'strong',
-  'ul',
-];
 
 export class AddonBase extends React.Component {
   static propTypes = {
@@ -237,8 +221,7 @@ export class AddonBase extends React.Component {
       if (!description || !description.length) {
         return null;
       }
-      descriptionProps.dangerouslySetInnerHTML = sanitizeHTML(
-        nl2br(description), allowedDescriptionTags);
+      descriptionProps.dangerouslySetInnerHTML = sanitizeUserHTML(description);
     } else {
       descriptionProps.children = <LoadingText width={100} />;
     }
@@ -255,6 +238,32 @@ export class AddonBase extends React.Component {
         />
       </ShowMoreCard>
     );
+  }
+
+  renderVersionReleaseNotes() {
+    const { addon, i18n } = this.props;
+    if (!addon) {
+      return null;
+    }
+
+    const currentVersion = addon.current_version;
+    if (!currentVersion || !currentVersion.release_notes) {
+      return null;
+    }
+
+    const header = i18n.sprintf(
+      i18n.gettext('Release notes for %(addonVersion)s'),
+      { addonVersion: currentVersion.version }
+    );
+    const releaseNotes = sanitizeUserHTML(currentVersion.release_notes);
+
+    /* eslint-disable react/no-danger */
+    return (
+      <Card className="AddonDescription-version-notes" header={header}>
+        <p dangerouslySetInnerHTML={releaseNotes} />
+      </Card>
+    );
+    /* eslint-enable react/no-danger */
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


kumar303:
@tofumatt well, huh, any idea how I broke the grid layout? This is me staring at the grid inspector with http://localhost:3000/en-US/firefox/addon/fastest-search/ (from `yarn amo:dev`) wondering how grids work:

<details>
  <summary>Screenshot</summary>

screenshot-2017-8-31 add-ons for firefox

</details>

src/core/utils.js
```diff
@@ -94,6 +94,35 @@ export function nl2br(text) {
...
+ *
+ * This also converts new lines to breaks (<br />) as a convenience when
+ * users did not write entirely in HTML.
+ *
+ * This is meant to display things like an add-on's description or version
+ * release notes. The allowed tags are meant to match what you see in the
+ * Developer Hub when you hover over the *Some HTML Supported* link under
+ * the textarea field.
+ */
+export function sanitizeUserHTML(text) {
```
tofumatt: Ohh, smart 👍

tests/unit/amo/components/TestAddon.js
```diff
@@ -393,22 +392,17 @@ describe('Addon', () => {
...
-    }
+  it('allows some HTML tags in the description', () => {
+    const root = shallowRender({
+      addon: {
+        ...fakeAddon,
+        description: '<b>super</b> <i>cool</i> <blink>add-on</blink>',
+      },
+    });
+    const contents = root.find('.AddonDescription-contents');
+    expect(contents.html()).toMatch(
```
tofumatt: I think this could use `.toHaveHTML()` rather than `.html()).toMatch()`. I think you wouldn't need the RegExp then.

tests/unit/core/test_utils.js
```diff
@@ -1130,3 +1131,27 @@ describe('parsePage', () => {
     expect(parsePage(undefined)).toBe(1);
   });
 });
+
+describe('sanitizeUserHTML', () => {
+  const sanitize = (...args) => sanitizeUserHTML(...args).__html;
+
+  it('converts new lines to breaks', () => {
```
tofumatt: 👍

tests/unit/core/test_utils.js
```diff
@@ -1130,3 +1131,27 @@ describe('parsePage', () => {
...
+    expect(sanitize(`
+      first
+      second
+    `).trim()).toEqual('<br>      first<br>      second<br>');
+  });
+
+  it('allows some tags', () => {
+    const customHtml =
+      '<b>check</b> <i>out</i> <a href="http://mysite">my site</a>';
+    expect(sanitize(customHtml)).toMatch(new RegExp(
```
tofumatt: Again, this can probably use [`.toHaveHTML`](https://github.com/blainekasten/enzyme-matchers#tohavehtmlhtmlstring)

src/amo/components/Addon/index.js
```diff
@@ -295,7 +304,7 @@ export class AddonBase extends React.Component {
       const authorList = addon.authors.map(
         (author) => `<a href="${author.url}">${author.name}</a>`);
       const title = i18n.sprintf(
-        // L10n: Example: The Add-On <span>by The Author</span>
+        // translators: Example: The Add-On <span>by The Author</span>
```
tofumatt: Thanks 👍

tests/unit/core/test_utils.js
```diff
@@ -1130,3 +1131,27 @@ describe('parsePage', () => {
     expect(parsePage(undefined)).toBe(1);
   });
 });
+
+describe('sanitizeUserHTML', () => {
```
tofumatt: I think this could use another test for supplying a `null` value that isn't a string. We probably don't want it... but we should either treat it like an empty string or throw an error. I just don't know what will happen looking at the code, so I think we should have a test to define the expected behaviour.

tofumatt:
Ah, I've run into this, I forget the fix but I'll get it posted in a few.

tofumatt:
Alright, I see what was up. Easy fix and it future-proofs more items 😄

You should add, to `src/amo/components/Addon/styles.scss`

```scss
.AddonDescription-version-notes {
  grid-column: 1;
}
```

and in that file change:

```scss
.AddonDescription {
  grid-column: 2;
  grid-row: 1 / 5;
}
```

to

```scss
.AddonDescription {
  grid-column: 2;
  grid-row: 1 / span 5;
}
```

kumar303:
> Easy fix and it future-proofs more items

Awesome, thanks. That fixed it. I need to learn grids, the tooling is pretty nice.

tests/unit/amo/components/TestAddon.js
kumar303: It looks like [.toHaveHTML()](https://github.com/blainekasten/enzyme-matchers#tohavehtmlhtmlstring) matches the entire HTML so I'd have to include the enclosing div and all its classes.

tofumatt:
Ah okay then, not worth it 😊

kumar303:
> I also wonder if they should be higher up in the HTML/mobile order because release notes are topical/relevant, especially compared to the "More Info" section.

Yeah, I'm not sure either. The mock actually has a support card coming before it too. We should probably review this whole page with Phil in a UX open house once we have all the elements in place.

tests/unit/core/test_utils.js
kumar303: Good call, I added it.

src/amo/components/Addon/index.js
```diff
@@ -257,6 +240,32 @@ export class AddonBase extends React.Component {
...
+
+    const header = i18n.sprintf(
+      i18n.gettext('Release notes for %(addonVersion)s'),
+      { addonVersion: currentVersion.version }
+    );
+    const releaseNotes = sanitizeUserHTML(currentVersion.release_notes);
+
+    /* eslint-disable react/no-danger */
+    return (
+      <ShowMoreCard className="AddonDescription-version-notes" header={header}>
```
tofumatt: Thanks, this is a lot nicer on mobile.

PR #3051, opened 2017-09-01
fix: Remove filters from categories

### LINKED ISSUE #3039: Filter options don't work in any category pages
Steps to reproduce:
1. Load any add-on/theme category i.e. https://addons-dev.allizom.org/en-US/firefox/extensions/bookmarks/
2. Change a filter criteria

Expected results:
Filter criteria is changed. Results are listed accordingly. 

Actual results:
Nothing happens.

Notes/Issues:
Verified on AMO-dev FF55(Win 7, Android 7.2.1).
Video for this issue:
filter

Related to #2956

tofumatt:
Fixes #3039.

This is sort of a stop-gap as we *should* eventually support filters here. But we didn't before so this at least prevents the bug.

### Before
screenshot 2017-09-01 02 43 25

### After
screenshot 2017-09-01 02 43 10

### DIFF
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -9,6 +9,7 @@ import Search from 'amo/components/Search';
 import { categoriesFetch } from 'core/actions/categories';
 import { withErrorHandler } from 'core/errorHandler';
 import log from 'core/logger';
+import { convertFiltersToQueryParams } from 'core/searchUtils';
 import { apiAddonType, parsePage } from 'core/utils';
 
 import './styles.scss';
@@ -58,7 +59,9 @@ export class CategoryBase extends React.Component {
       page: parsePage(location.query.page),
     };
     const pathname = `/${params.visibleAddonType}/${categorySlug}/`;
-    const paginationQueryParams = { page: filters.page };
+    const paginationQueryParams = convertFiltersToQueryParams({
+      page: filters.page,
+    });
 
     let category;
     if (categories) {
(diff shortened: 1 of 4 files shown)


src/amo/components/Category/index.js
```diff
@@ -58,7 +59,9 @@ export class CategoryBase extends React.Component {
       page: parsePage(location.query.page),
     };
     const pathname = `/${params.visibleAddonType}/${categorySlug}/`;
-    const paginationQueryParams = { page: filters.page };
+    const paginationQueryParams = convertFiltersToQueryParams({
```
tofumatt: This doesn't change anything right now because `page` is one of the few `filters` that has the same key as in `queryParams` but this is what we should be doing, so I did it here.

tofumatt:
Cool, thanks.

Yeah the issue is the categories page seems not to always update/set the right params. It shouldn't be terrible to fix but the main annoyance is changing the URL on filter change. Anyway, I'll comment further on the issue 😄

willdurand:
> Yeah the issue is the categories page seems not to always update/set the right params. It shouldn't be terrible to fix but the main annoyance is changing the URL on filter change. Anyway, I'll comment further on the issue 😄

Also I guess we should disable some filters, e.g., the add-on type one since categories are per type already.

willdurand:
Follow-up: https://github.com/mozilla/addons-frontend/issues/3053, feel free to update it.

PR #3057, opened 2017-09-01
ShowMoreCard should expand if updated content changes

### LINKED ISSUE #3026: Short description has "Read More" link
Steps to reproduce:
1. Load details page for an add-on with short description i.e. https://addons-dev.allizom.org/en-US/firefox/addon/adblock-edge/

Expected results:
No "Reed More" link.

Actual results:
"Read More" link is present even if the description is 1 row long.

Notes/Issues:
Verified on FF55(Win 7, Android 7.2.1). Issue is reproducing on both dev and stage.
Screenshot for this issue:

Could be related to #3017

willdurand:
Fix #3026 

---

This PR fixes an issue when the description of an add-on has a smaller size than the previously browsed one.

### DIFF
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -3,7 +3,6 @@
 import classNames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
-import ReactDOM from 'react-dom';
 import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
@@ -29,15 +28,13 @@ export class ShowMoreCardBase extends React.Component {
   }
 
   componentWillReceiveProps() {
-    this.truncateToMaxHeight(ReactDOM.findDOMNode(this.contents));
+    this.setState({ expanded: true }, () => {
+      this.truncateToMaxHeight(this.contents);
+    });
   }
 
   onClick = (event) => {
     event.preventDefault();
-    this.expandText();
-  }
-
-  expandText() {
     this.setState({ expanded: true });
   }
 
--- a/tests/unit/ui/components/TestShowMoreCard.js
+++ b/tests/unit/ui/components/TestShowMoreCard.js
@@ -13,7 +13,7 @@ function render(props) {
   );
 }
 
-describe('<ShowMoreCard />', () => {
+describe(__filename, () => {
   it('reveals more text when clicking "show more" link', () => {
     const root = render({ children: 'Hello I am description' });
     const rootNode = findDOMNode(root);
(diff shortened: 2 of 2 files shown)


tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -69,4 +69,23 @@ describe('<ShowMoreCard />', () => {
 
     sinon.assert.calledWith(truncateToMaxHeight, contentNode);
   });
+
+  it('expands if new child content is smaller', () => {
+    const root = mount(
+      <ShowMoreCardBase i18n={getFakeI18nInst()}>
+        This should be a very long content, but we cannot get `clientHeight` in
```
tofumatt: Nitpick but this would be better written as "This would be very long content, but [...]"

src/ui/components/ShowMoreCard/index.js
```diff
@@ -29,15 +29,13 @@ export class ShowMoreCardBase extends React.Component {
   }
 
   componentWillReceiveProps() {
-    this.truncateToMaxHeight(ReactDOM.findDOMNode(this.contents));
+    this.setState({ expanded: true }, () => {
+      this.truncateToMaxHeight(ReactDOM.findDOMNode(this.contents));
```
tofumatt: While we're here, should we still be using `ReactDOM.findDOMNode`? I forget if we can do this with refs or not but maybe we could?

willdurand:
@tofumatt thanks for the review, just addressed your comments.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -69,4 +69,23 @@ describe('<ShowMoreCard />', () => {
 
     sinon.assert.calledWith(truncateToMaxHeight, contentNode);
   });
+
+  it('expands if new child content is smaller', () => {
+    const root = mount(
+      <ShowMoreCardBase i18n={getFakeI18nInst()}>
+        This would be a very long content, but we cannot get `clientHeight` in
```
tofumatt: Nitpick: should be "This would be very long content" not "a very long content"

willdurand: doh 😅

src/ui/components/ShowMoreCard/index.js
```diff
@@ -29,15 +28,13 @@ export class ShowMoreCardBase extends React.Component {
   }
 
   componentWillReceiveProps() {
-    this.truncateToMaxHeight(ReactDOM.findDOMNode(this.contents));
+    this.setState({ expanded: true }, () => {
+      this.truncateToMaxHeight(this.contents);
```
kumar303: This will work in production but I don't think it will work for our test suite. The problem is that React [does not guarantee](https://facebook.github.io/react/docs/react-component.html#setstate) that the callback to `setState()` will be invoked immediately. Thus, the test you wrote will probably pass *most* of the time but not all of the time. If the `setState()` callback is delayed then your test will fail.

There are a few ways to solve this. We could try to wedge promises in here so that the test can use a promise chain. I don't see an easy way to do that.

Alternatively, we could keep the process synchronous which would make the test you already wrote pass 100% of the time. I think restoring `componentWillReceiveProps()` to how it was and adding this `else` block would accomplish that:

```js
truncateToMaxHeight = (contents) => {
  // If the contents are short enough they don't need a "show more" link; the
  // contents are expanded by default.
  if (contents.clientHeight > MAX_HEIGHT) {
    this.setState({ expanded: false });
  } else {
    this.setState({ expanded: true });
  }
}
```

kumar303: Well, it's possible that Enzyme has some trickery to ensure all batched state changes are applied but they don't document this. I'm skeptical since this kind of state change is triggered from *within* the component (not from an Enzyme method).

willdurand: What we could do is reuse the `callback` from Enzyme `setState()` (see: https://github.com/airbnb/enzyme/pull/617) and ensure the test case asserts things once the `setState()` has been triggered (and use `done()` to ensure the test does not silently fail)

kumar303: Ugh, yeah, we really can't rely on `setState()` being synchronous anywhere. https://github.com/airbnb/enzyme/issues/509#issuecomment-237362893

It would be nice to use promises instead of `done()` but your approach will work. Creating a promise wrapper for everywhere we rely on `wrapper.setState()` in the tests is probably a better long term solution.

PR #3058, opened 2017-09-01
Avoid fetching suggestions all the time

### LINKED ISSUE #3030: Autocomplete files when clicking on an already-run search
If you type a search term into the search box, let the results load, click away, and then focus the search again, it will run the autocomplete search a second time despite having local results. This creates an unneeded network request and also feels a bit too frantic.

### What happened?
The autocomplete action dispatched again despite the filters all being the same.

The existing results are simply loaded/shown again because the filters haven't changed.

willdurand:
Fix #3030 

---

This PR slightly changes the autocomplete feature to avoid too many API calls.

This was more complicated than I thought 😞 Hopefully the new test cases are meaningful enough but they should cover all the edge cases I encountered.

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -7,6 +7,7 @@ import { compose } from 'redux';
 import Autosuggest from 'react-autosuggest';
 import { withRouter } from 'react-router';
 import defaultDebounce from 'simple-debounce';
+import deepEqual from 'deep-eql';
 
 import Suggestion from 'amo/components/SearchSuggestion';
 import {
@@ -60,6 +61,7 @@ export class SearchFormBase extends React.Component {
 
     this.state = {
       searchValue: props.query || '',
+      oldAutocompleteFilters: {},
     };
   }
 
@@ -112,7 +114,7 @@ export class SearchFormBase extends React.Component {
     this.setState({ searchValue: event.target.value });
   }
 
-  handleSuggestionsFetchRequested = this.props.debounce(({ value }) => {
+  handleSuggestionsFetchRequested = this.props.debounce(({ value, reason }) => {
     if (!value) {
       log.debug(oneLine`Ignoring suggestions fetch requested because value
         is not supplied: ${value}`);
@@ -126,10 +128,22 @@ export class SearchFormBase extends React.Component {
       filters.addonType = addonType;
     }
 
-    dispatch(autocompleteStart({
-      errorHandlerId: errorHandler.id,
-      filters,
-    }));
+    if (
+      // See: https://github.com/moroshko/react-autosuggest#onsuggestionsfetchrequested-required.
+      reason === 'input-focused' &&
+      deepEqual(filters, this.state.oldAutocompleteFilters)
+    ) {
+      log.debug(oneLine`No autocomplete start dispatched: search input focused
+        but filters have not changed.`);
+      return;
+    }
+
+    this.setState({ oldAutocompleteFilters: filters }, () => {
+      dispatch(autocompleteStart({
+        errorHandlerId: errorHandler.id,
+        filters,
+      }));
+    });
   }, 200)
 
   handleSuggestionsClearRequested = () => {
--- a/tests/unit/amo/components/TestSearchForm.js
+++ b/tests/unit/amo/components/TestSearchForm.js
@@ -247,6 +247,179 @@ describe(__filename, () => {
       }));
     });
 
+    it('does not fetch suggestions twice on focus when search has not changed', () => {
+      const { store } = dispatchAutocompleteResults({ results: [
+        createFakeAutocompleteResult(),
+        createFakeAutocompleteResult(),
+      ] });
+      const { autocomplete: autocompleteState } = store.getState();
+
+      const dispatch = sinon.spy();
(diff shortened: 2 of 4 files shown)


src/amo/components/SearchForm/index.js
```diff
@@ -55,6 +57,7 @@ export class SearchFormBase extends React.Component {
 
     this.state = {
       searchValue: props.query || '',
+      lastAutocompleteFilters: {},
```
tofumatt: Could this be either `previousAutoCompleteFilters` or `oldFilters`? Last sounds like the latest and is confusing, so I think this could be better named.

src/amo/components/SearchForm/index.js
```diff
@@ -119,10 +123,22 @@ export class SearchFormBase extends React.Component {
       filters.addonType = addonType;
     }
 
-    dispatch(autocompleteStart({
-      errorHandlerId: errorHandler.id,
-      filters,
-    }));
+    if (
+      // See: https://github.com/moroshko/react-autosuggest#onsuggestionsfetchrequested-required.
```
tofumatt: Thanks for linking to that, very handy 👍

src/amo/components/SearchForm/index.js
```diff
@@ -119,10 +123,22 @@ export class SearchFormBase extends React.Component {
...
+      reason === 'input-focused' &&
+      deepEqual(filters, this.state.lastAutocompleteFilters)
+    ) {
+      log.debug(oneLine`Ignoring suggestions fetch requested because it has been
```
tofumatt: I think this could be a bit clearer/concise, maybe just:

"No autocomplete start dispatched: search input focused but filters have not changed."

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -219,6 +219,173 @@ describe(__filename, () => {
...
+        dispatch,
+        suggestions: autocompleteState.suggestions,
+      });
+
+      sinon.assert.notCalled(dispatch);
+
+      // User types 'a' in the search input.
+      wrapper.find('input').simulate('change', createFakeChangeEvent('a'));
+
+      // User clicks somewhere else on the UI, triggering the
```
tofumatt: These comments in the test describing the actions the user would take are fantastic, thank you much!

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -219,6 +219,173 @@ describe(__filename, () => {
...
+
+      // User focuses the search input.
+      wrapper.find('input').simulate('focus');
+
+      expect(wrapper.find(Suggestion)).toHaveLength(
+        autocompleteState.suggestions.length
+      );
+
+      sinon.assert.callCount(dispatch, 2);
+      sinon.assert.calledWith(dispatch, autocompleteStart({
```
tofumatt: I think this test would be ever more robust if you reset the dispatch before `wrapper.setProps({ addonType: ADDON_TYPE_THEME });`. Technically this isn't checking the call order, which I know is minor but if you instead made sure it's called once after each event we expect the dispatch on we can check the order too.

It's super minor but I think it's better overall.

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -219,6 +219,173 @@ describe(__filename, () => {
...
+
+      // User types 'adb' in the search input, which requires new suggestions
+      // to be loaded from the API.
+      wrapper.find('input').simulate('change', createFakeChangeEvent('adb'));
+
+      expect(wrapper.find(Suggestion)).toHaveLength(
+        autocompleteState.suggestions.length
+      );
+
+      sinon.assert.callCount(dispatch, 2);
```
tofumatt: Same here; I think doing the check before `wrapper.find('input').simulate('change', createFakeChangeEvent('adb'));` and then doing `dispatch.reset()` then checking again would be a bit more robust and also more clearly show which assertion is related to specific actions we're simulating in the test.

src/amo/components/SearchForm/index.js
willdurand: Would `oldAutocompleteFilters` work for you? I started with `lastDispatched(Autocomplete)Filters`, but that was waaaay too long, so I shortened the name, but agreed that now it feels weird to have `last`.

tofumatt: `oldAutocompleteFilters` works for me; we use `oldFilters` elsewhere. I think `previous` is better still but you're right it's verbose and we're already using the `old` prefix elsewhere so let's do that.

src/amo/components/SearchForm/index.js
```diff
@@ -126,10 +128,22 @@ export class SearchFormBase extends React.Component {
...
+      // See: https://github.com/moroshko/react-autosuggest#onsuggestionsfetchrequested-required.
+      reason === 'input-focused' &&
+      deepEqual(filters, this.state.oldAutocompleteFilters)
+    ) {
+      log.debug(oneLine`No autocomplete start dispatched: search input focused
+        but filters have not changed.`);
+      return;
+    }
+
+    this.setState({ oldAutocompleteFilters: filters }, () => {
```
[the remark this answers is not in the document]
willdurand: Ok I am going to remove the async callback.

PR #3059, opened 2017-09-01
[WIP] Try prettier

muffinresearch:
💣 **DO NOT MERGE.** 💣

This branch does a setup of Prettier and runs it on the code base so everyone can see what it will do.

Feel free to use the review to comment on any bits this is changing that are problematic.


.eslintrc
```diff
@@ -1,15 +1,20 @@
...
     "ga": true,
   },
   "parser": "babel-eslint",
   "plugins": [
-    "jest"
+    "jest",
+    "prettier",
   ],
   "rules": {
+    "prettier/prettier": ["error", {"singleQuote": true, "trailingComma": 'es5'}],
```
muffinresearch: This duplicate of the .prettierrc conf will eventually be eliminated, Prettier's config usage is new and the eslint plugin needs to catch up.

bin/start-func-test-server.js
```diff
@@ -32,22 +32,26 @@ function shell(cmd, args) {
       cwd: root,
     });
 
-    child.stdout.on('data', (data) => {
```
muffinresearch: I wish this part was configurable :(

bin/proxy.js
```diff
@@ -15,12 +15,14 @@ const log = bunyan.createLogger({
 
 const proxy = httpProxy.createProxyServer();
 const apiHost = config.get('proxyApiHost', null) || config.get('apiHost');
-const frontendHost = `http://${config.get('serverHost')}:${config.get('serverPort')}`;
+const frontendHost = `http://${config.get('serverHost')}:${config.get(
+  'serverPort'
+)}`;
```
willdurand: this is a bit weird

bin/proxy.js
```diff
@@ -32,7 +34,7 @@ const array = (value) => {
 function unsecureCookie(req, res, proxyRes) {
   const proxyCookies = array(proxyRes.headers['set-cookie']);
   // eslint-disable-next-line no-param-reassign
-  proxyRes.headers['set-cookie'] = proxyCookies.map((rewrittenCookie) => {
+  proxyRes.headers['set-cookie'] = proxyCookies.map(rewrittenCookie => {
```
willdurand: ESLint used to yell at me when I was doing this (which is actually valid..)

tofumatt: I *thoroughly* hate this syntax. I think we should ban it.

.eslintrc
willdurand: Don't you have to specify `--parser flow`? I cannot remember why I have that in my Prettier settings..

bin/proxy.js
tofumatt: Agreed. I'd rather use `oneLine` than this...

muffinresearch:
I'm going to close this for now - I'm thinking we're probably better off using eslint rules for now. Generally that approach hasn't got in the way and we're already unified across the amo related js projects save for web-ext.

bin/start-func-test-server.js
kumar303: personally, I really hate having to type parentheses all the time :) It's very common to make an arrow function with only one param. We have lint rules to catch mistakes.

tofumatt: I value consistency and scan-ability--using parens all the time makes it easier to pick out a function as they always look the same (this is why I dislike implicit return functions as well). We spend more time reading code than writing it, so I think the effort to type something matters less than the effort for a human to scan/parse/read it. 😄

kumar303: It just takes a little getting used to. When your eye sees the arrow (`=>`) it immediately knows what's going on.

tofumatt: While the `=>` is a bit of a unique set of characters I definitely look for more than just `=>` when looking for a function. I see parens `()` as a reference to something callable.

My main issue though is having to train my brain and other developers to learn two ways to do the same thing. This is why I also dislike implicit returns: rather than always writing a function as `(param) => { return value; }` I need to write it in a different way depending on context or my style choices or whatever. And developers need to learn that rule. Or prettier would transform their effort of writing the most expressive and clear version to something condensed because it didn't like the verbosity used in this case. And if you're using object de-structuring (which is sort of just using one argument in theory, right?) you still need parens.

I think it's a code smell to have multiple ways of writing the same thing which is why I advocate for always using full function syntax. I think having prettier or whatever always convert to full functions would be way better. Anyone new to the code wouldn't have to learn/wonder why things are written different ways in different places. The whole reason for code style is consistency is key 😄

willdurand:
We finally did it! See #5313

PR #3063, opened 2017-09-03
Send platform in search and autocomplete

### LINKED ISSUE #2706: Add support for appversion and platform in search query params
Related to https://bugzilla.mozilla.org/show_bug.cgi?id=1263313 we're going to need to add support for `appversion` and `platform` query strings in search urls.

tofumatt:
Fixes #2706 by sending the operating system (basically only for desktop searches right now, as we filter for android compatibility with `clientApp` instead).

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -26,7 +26,7 @@ import {
   autocompleteStart,
 } from 'core/reducers/autocomplete';
 import {
-  convertOperatingSystemToFilterName,
+  convertOSToFilterValue,
   convertFiltersToQueryParams,
 } from 'core/searchUtils';
 import Icon from 'ui/components/Icon';
@@ -101,7 +101,7 @@ export class SearchFormBase extends React.Component {
       filters.addonType = addonType;
     }
 
-    filters.operatingSystem = convertOperatingSystemToFilterName(
+    filters.operatingSystem = convertOSToFilterValue(
       userAgentInfo.os.name);
 
     return filters;
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -93,7 +93,7 @@ export class SearchFormBase extends React.Component {
     }));
   }
 
-  setFilters(newFilters) {
+  createFiltersFromQuery(newFilters) {
     const { addonType, userAgentInfo } = this.props;
     const filters = { ...newFilters };
 
@@ -109,7 +109,7 @@ export class SearchFormBase extends React.Component {
 
   goToSearch(query) {
     const { api, pathname, router } = this.props;
-    const filters = this.setFilters({ query });
+    const filters = this.createFiltersFromQuery({ query });
 
     router.push({
       pathname: `/${api.lang}/${api.clientApp}${pathname}`,
@@ -144,7 +144,7 @@ export class SearchFormBase extends React.Component {
       return;
     }
 
-    const filters = this.setFilters({ query: value });
+    const filters = this.createFiltersFromQuery({ query: value });
 
     this.setState({
       autocompleteIsOpen: true,
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -21,11 +21,14 @@ import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import { getAddonIconUrl } from 'core/imageUtils';
 import log from 'core/logger';
-import { convertFiltersToQueryParams } from 'core/searchUtils';
 import {
   autocompleteCancel,
   autocompleteStart,
 } from 'core/reducers/autocomplete';
+import {
+  convertOperatingSystemToFilterName,
+  convertFiltersToQueryParams,
+} from 'core/searchUtils';
 import Icon from 'ui/components/Icon';
 
 import './styles.scss';
@@ -49,6 +52,7 @@ export class SearchFormBase extends React.Component {
       url: PropTypes.string.isRequired,
       iconUrl: PropTypes.string.isRequired,
     })).isRequired,
+    userAgentInfo: PropTypes.object.isRequired,
   }
 
   static defaultProps = {
@@ -90,12 +94,14 @@ export class SearchFormBase extends React.Component {
   }
 
   goToSearch(query) {
-    const { addonType, api, pathname, router } = this.props;
+    const { addonType, api, pathname, router, userAgentInfo } = this.props;
     const filters = { query };
 
     if (addonType) {
       filters.addonType = addonType;
     }
+    filters.operatingSystem = convertOperatingSystemToFilterName(
+      userAgentInfo.os.name);
 
     router.push({
       pathname: `/${api.lang}/${api.clientApp}${pathname}`,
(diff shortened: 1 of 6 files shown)


src/core/api/search.js
```diff
@@ -57,25 +56,12 @@ export default function search(
     _filters.clientApp = 'firefox';
   }
 
-  // Themes are cross-platform and searching for them with an operatingSystem
-  // filter will result in no results, so we delete it for now.
-  // Can be deleted once https://github.com/mozilla/addons-server/issues/6206
```
tofumatt: Note: this is only fixed on dev right now, so make sure to test this against dev or it will error.

src/amo/components/SearchForm/index.js
```diff
@@ -88,12 +92,14 @@ export class SearchFormBase extends React.Component {
...
   goToSearch(query) {
-    const { addonType, api, pathname, router } = this.props;
+    const { addonType, api, pathname, router, userAgentInfo } = this.props;
     const filters = { query };
 
     if (addonType) {
       filters.addonType = addonType;
     }
+    filters.operatingSystem = convertOperatingSystemToFilterName(
+      userAgentInfo.os.name);
```
kumar303: It looks to me like we need to DRY this up a bit since we set `filters` in the same way in two places. We need something like `const filters = this.setFilters(newFilters)` that uses values from `newFilters` and falls back to values from `this.props`.

src/core/searchUtils.js
```diff
@@ -35,6 +41,14 @@ export function convertQueryParamsToFilters(params) {
   }, {});
 }
 
+export function convertOperatingSystemToFilterName(name) {
```
kumar303: I think this is actually `convertOperatingSystemToFilterValue`, right? Also, I think abbreviating it to `OS` would make the code a bit easier to read but no biggie.

src/core/searchUtils.js
```diff
@@ -35,6 +41,14 @@ export function convertQueryParamsToFilters(params) {
...
+  if (name in operatingSystems) {
+    return operatingSystems[name];
+  }
+
```
kumar303: It would be nice to see a warning in the log about unknown operating systems. I suspect we'll have to fix some edge cases like `Linux (x86)` due to UA parsing errors -- just a hunch! Hopefully not.

src/core/searchUtils.js
```diff
@@ -35,6 +41,14 @@ export function convertQueryParamsToFilters(params) {
...
+  return '';
```
kumar303: should this be `undefined` so that the query string doesn't get a value?

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -12,52 +12,73 @@ import {
...
-  const mountComponent = ({
-    store = dispatchClientMetadata().store,
+  function mountComponent({
+    pathname = '/search/',
+    query = 'foo',
+    store = dispatchSignInActions().store,
     ...props
-  } = {}) => {
+  } = {}) {
     return mount(
```
kumar303: why do so many tests need to use `mount()`? Is it just that they haven't been converted to use `shallow()` yet?

willdurand: Isn't it because we use `simulate()` on components that are not `root`? Would `shallowUntilTarget()` allow that?

src/core/searchUtils.js
tofumatt: An empty value will also be removed... but yes, that would make more sense semantically.

tests/unit/amo/components/TestSearchForm.js
tofumatt: It's because they are checking state, actually. I'll add some comments. A few are using it because they were supplying props directly and not dispatching and it was a fair bit of wrangling to convert them. I can see about migrating the last few.

src/core/searchUtils.js
tofumatt: That's totally true, thanks.

src/core/searchUtils.js
tofumatt: Yes and yes. 👍

src/amo/components/SearchForm/index.js
tofumatt: Oh, good call. I'll have a look at that.

tofumatt:
I'll make some of the requested changes and fix the merge conflict. Thanks.

tofumatt:
Sorry for the delay, this is finally ready for another review.

tofumatt:
Okay conflicts resolved, good for another r?

src/amo/components/SearchForm/index.js
```diff
@@ -217,8 +226,8 @@ export class SearchFormBase extends React.Component {
     };
 
     const autocompleteIsOpen = this.state.autocompleteIsOpen &&
-      // This prevents the input to look like Autosuggest is open when there is
-      // no result coming from the API.
+      // This prevents the input to look like Autosuggest is open when
+      // there is no result coming from the API.
```
willdurand: You wrap at 72 chars right, not 80?

src/core/api/search.js
```diff
@@ -57,25 +56,12 @@ export default function search(
...
   // compatible with this version.
   if (
     api.userAgentInfo.browser.name === 'Firefox' &&
-    api.userAgentInfo.os.name !== 'iOS' &&
-    // TODO: Remove this check once
-    // https://github.com/mozilla/addons-server/issues/6206 is fixed.
-    // Right now sending the `appversion` param to search will result
-    // in no themes being returned.
-    (!_filters.addonType || _filters.addonType === ADDON_TYPE_EXTENSION)
+    api.userAgentInfo.os.name !== 'iOS'
```
willdurand: Can we have a constant for that?

src/amo/components/SearchForm/index.js
tofumatt: I wrap at 80. 😅

src/core/api/search.js
tofumatt: Hmmm, I'll see if the `UAParser` lib we use exposes constants. That's a good call though. These are all from an external library but we can write constants if they don't.

tests/unit/amo/components/TestSearchForm.js
```diff
@@ -280,19 +326,19 @@ describe(__filename, () => {
       sinon.assert.callCount(dispatch, 1);
       sinon.assert.calledWith(dispatch, autocompleteStart({
         errorHandlerId: errorHandler.id,
-        filters: { query: 'a' },
+        filters: { operatingSystem: OS_WINDOWS, query: 'a' },
       }));
     });
 
-    it('fetches suggestions twice on focus when search has not changed but add-type changed', () => {
+    it('fetches suggestions twice on focus when search has not changed but add-on type changed', () => {
```
willdurand: :bowtie:

tests/unit/core/test_searchUtils.js
```diff
@@ -41,4 +44,46 @@ describe(__filename, () => {
...
+
+  describe('convertOSToFilterValue', () => {
+    function getOSNameFromUserAgent(userAgent) {
+      const { store } = dispatchClientMetadata({ userAgent });
+      return store.getState().api.userAgentInfo.os.name;
+    }
+
+    it('converts Windows to filter', () => {
+      const osName = getOSNameFromUserAgent(userAgents.firefox[1]);
+      const filterName = convertOSToFilterValue(osName);
```
kumar303: No biggie but `filterName` in all of these look like they're left over from before the name change

tofumatt: Hah, good catch.

PR #3065, opened 2017-09-04
Fix wrong class name on SearchForm

### LINKED ISSUE #3064: Search box bottom corners are not rounded after deleting searchterm
Steps to reproduce:
1.Go to https://addons-dev.allizom.org/en-US/firefox/ , mobile view
2.Enter a search text in search box , then delete it

Expected results:
Search box is not affected

Actual results:
Search box bottom corners are not rounded after deleting searchterm

rounded

willdurand:
Fix #3064

---

This PR fixes a visual bug on the search form, after deleting search terms.

~~⚠️  There is a remaining glitch, likely due to our `debounce` default implementation:~~ => replaced `simple-debounce` lib by `lodash.debounce` one (which is configurable).


willdurand:
Looks good now but I had to change the debounce library that was in use.

src/amo/components/SearchForm/index.js
```diff
@@ -144,7 +144,9 @@ export class SearchFormBase extends React.Component {
         filters,
       }));
     });
-  }, 200)
+  // Configure debounce (200ms) with `leading: true` so that the method is
+  // triggered immediately and subsequent calls are debounced.
+  }, 200, { leading: true, trailing: false })
```
tofumatt: Is this a value we would want to use elsewhere in the app? It seems like 200ms is a good default for debouncing. Maybe we want to store this as a constant somewhere?

willdurand: The `AddonReview` component (the only other one that uses `debounce` so far) is configured with 800ms.

tofumatt: Ah, fair enough. Okay then, not worth it 😄

PR #3073, opened 2017-09-06
Make add-on reducer state more explicit

### LINKED ISSUE #3024: Make add-on reducer state more explicit

kumar303:
Fixes #3024 

This will hopefully make the central add-on reducer easier to deal with for features and bug fixes but it has potential to break a lot of stuff. It would help if you could test out add-on/theme installs. I've manually tested AMO and the Discopane.


src/core/reducers/addons.js
```diff
@@ -1,15 +1,36 @@
 /* @flow */
 import { ADDON_TYPE_THEME } from 'core/constants';
 import type { ErrorHandlerType } from 'core/errorHandler';
-import type { AddonType } from 'core/types/addons';
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
 
 
+export const ADDONS_LOADED = 'ADDONS_LOADED';
```
willdurand: Can we have `LOAD_ADDON` instead? See: https://github.com/mozilla/addons-frontend/pull/3025#discussion_r135909932

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
+    is_disabled: apiAddon.is_disabled,
+    is_experimental: apiAddon.is_experimental,
+    is_featured: apiAddon.is_featured,
+    is_source_public: apiAddon.is_source_public,
+    last_updated: apiAddon.last_updated,
+    latest_unlisted_version: apiAddon.latest_unlisted_version,
+    name: apiAddon.name,
+    previews: apiAddon.previews,
+    public_stats: apiAddon.public_stats,
+    ratings: apiAddon.ratings || undefined,
```
willdurand: Why does this one need this?

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
+  // (which has a header and description) with a minimal add-on object.
+  // For example, the minimal add-on object does not have a description
+  // property so the spread should not override `description`.
+  addon = removeUndefinedProps(addon);
+
+  return addon;
 }
 
-export default function addon(
+const initialState = {};
```
willdurand: Flow type?

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
     // This mimics how Firefox appends @personas.mozilla.org internally.
     // It's needed to look up themes in mozAddonManager.
-    return `${result.id}@personas.mozilla.org`;
+    return `${addon.id}@personas.mozilla.org`;
   }
-  return result.guid;
+  return addon.guid;
+}
+
+export function removeUndefinedProps(object: Object): Object {
```
willdurand: Is this method needed because there are spreads in different places? Otherwise I would argue that the state should have a consistent shape.

src/core/actions/index.js
```diff
@@ -69,15 +68,3 @@ export function setUserAgent(userAgent: string): SetUserAgentAction {
     payload: { userAgent },
   };
 }
-
-export type LoadEntitiesAction = {|
-  payload: {| entities: Array<Object> |},
-  type: string,
-|};
-
-export function loadEntities(entities: Array<Object>): LoadEntitiesAction {
```
tofumatt: Yay!

src/core/reducers/addons.js
```diff
@@ -1,15 +1,36 @@
...
+
+export type LoadAddonsAction = {|
+  payload: {| addons: ExternalAddonMap |},
+  type: string,
+|};
 
-type Action = Object;
-type AddonState = Object;
+export function loadAddons(
+  entities: {| addons?: ExternalAddonMap |}
```
tofumatt: As mentioned previously I think it would be great to keep flow definitions out of the function signatures if they're being included. They really crowd the function signature and make it very difficult to parse exactly what's happening inside it.

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
+    support_email: apiAddon.support_email,
+    support_url: apiAddon.support_url,
+    tags: apiAddon.tags,
+    type: apiAddon.type,
+    url: apiAddon.url,
+    weekly_downloads: apiAddon.weekly_downloads,
+
+    // These are custom properties not in the API response.
+
+    // TODO: remove this if possible. I think it was added by mistake
```
tofumatt: 👍

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
+        textcolor: apiAddon.theme_data.textcolor,
+        updateURL: apiAddon.theme_data.updateURL,
+        version: apiAddon.theme_data.version,
+      }),
     };
   }
-  return apiAddon;
+
+  if (apiAddon.current_version && apiAddon.current_version.files.length > 0) {
+    // TODO: support specific platform files.
```
tofumatt: All these comments are very helpful 🥂!

src/core/reducers/addons.js
```diff
@@ -37,69 +58,151 @@ export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddon
...
-      }
-      newState[key] = denormalizeAddon(newState[key]);
-    });
-    return newState;
+  switch (action.type) {
+    case ADDONS_LOADED: {
+      const { addons } = action.payload;
+      const newState = { ...state };
+      Object.keys(addons).forEach((key) => {
+        newState[key] = createInternalAddon(addons[key]);
```
tofumatt: 🎉💃🎉 THANK YOU THIS IS SO MUCH EASIER TO READ 🎉💃🎉

src/core/sagas/addons.js
```diff
@@ -7,9 +7,8 @@ import { hideLoading, showLoading } from 'react-redux-loading-bar';
 import { call, put, select, takeEvery } from 'redux-saga/effects';
 /* eslint-enable import/order */
 
-import { loadEntities } from 'core/actions';
 import { fetchAddon as fetchAddonFromApi } from 'core/api';
-import { FETCH_ADDON } from 'core/reducers/addons';
+import { loadAddons, FETCH_ADDON } from 'core/reducers/addons';
```
tofumatt: Nitpick but I think we usually load constants first. It technically is alphabetical that way but I guess until we have a rule for sorting imports... 😞

src/core/sagas/addons.js
```diff
@@ -25,7 +24,7 @@ export function* fetchAddon(
     yield put(showLoading());
     const state = yield select(getState);
     const response = yield call(fetchAddonFromApi, { api: state.api, slug });
-    yield put(loadEntities(response.entities));
+    yield put(loadAddons(response.entities));
```
tofumatt: This is just way easier to read too. Thanks a million.

src/core/types/addons.js
```diff
@@ -20,15 +20,39 @@ export type AddonAuthorType = {|
...
+  previewURL?: string,
+  textcolor?: string,
+  updateURL?: string,
+  version?: string,
+|};
+
+/*
+ * This is the external API representation of an add-on.
+ *
+ * This is a detailed API response. Not all API responses
```
tofumatt: Linking to the API docs from here would be cool just so developers could understand better the varied details in responses. Maybe just http://addons-server.readthedocs.io/en/latest/topics/api/addons.html ?

tests/unit/core/reducers/test_addons.js
```diff
@@ -123,65 +176,42 @@ describe(__filename, () => {
...
-        isRestartRequired: false,
-      }),
-    });
+    const state = addons(undefined,
+      loadAddons(createFetchAddonResult(addon).entities));
+    expect(state[addon.slug].isRestartRequired).toBe(false);
   });
 
-  it('sets `isRestartRequired` to `true` when at least one of the files declares it', () => {
+  it('sets `isRestartRequired` to `true` when some files declare it', () => {
```
tofumatt: if the aim here was to shorten this spec I'd say "when any file declares it".

src/core/reducers/addons.js
kumar303: ah, right, thanks

src/core/reducers/addons.js
kumar303: Yes it is needed for spreads. I added a comment about it below. It's going to be hard to remove the spreads and I think it's a low priority fix.

src/core/reducers/addons.js
kumar303: ah, whoops, good catch. That was leftover from me battling with Flow :/

src/core/reducers/addons.js
willdurand: > I think it's a low priority fix.

👍

src/core/reducers/addons.js
kumar303: There is a type for it -- you can see it assigned in the function signature: https://github.com/mozilla/addons-frontend/pull/3073/files/39fb66595b5b8033cd43288e61a19f35d9f32dbf#diff-21aa4f8ea4fba6406233649d0ce65beaR193

src/core/reducers/addons.js
kumar303: The issue I have with making them separate is that it slows down the reader. If you want to know what the type is, you have to cross-reference it from somewhere else. I understand how separation can be helpful for big definitions but this one only spans about 30 characters so I don't think that warrants separation.

src/core/types/addons.js
kumar303: good idea

kumar303:
Thanks for the reviews! Let's start testing this on dev...

Issue #3077, opened 2017-09-07
Make tests more predictable by removing setState callbacks


kumar303:
I don't think our tests should rely on `setState()` being synchronous as it's [unclear](https://github.com/airbnb/enzyme/issues/509#issuecomment-237362893) that this will always work in enzyme as its libraries evolve and get upgraded.

We have a few problems in the test suite related to this but first let's try not to rely on `setState` callbacks.

Here are two places that (hopefully) can be fixed with different approaches:

* https://github.com/mozilla/addons-frontend/pull/3058#pullrequestreview-60648776
* https://github.com/mozilla/addons-frontend/pull/3057#pullrequestreview-60631026

willdurand:
should be fixed now.

PR #3081, opened 2017-09-08
Fix search input corners and autocomplete

### LINKED ISSUE #3078: Quickly typing a search term does not autocomplete it

Quickly type a word in the search box

### What happened?

The last few characters you type will not be included in the search term. You can verify this in the network panel -- inspect the API request made to the `autocomplete` endpoint. 

For example, if you type "ad block" quickly, the search term will just be "a". If you type a bit slower, you may get more of the word but probably not the whole thing.

Autocompletion should always search for the full word you type.

This is a regression from fixing https://github.com/mozilla/addons-frontend/issues/3064 . We chatted about it in IRC and I think `handleSuggestionsFetchRequested` can immediately call `setState` without a debounce to change its class name. The call to `dispatch(autocompleteStart())` should still be debounced.

willdurand:
Fix #3078
Refs #3077

---

This PR reverts the trailing edge debouncing implementation introduced in #3065. It also get rid of the async `setState` callback. Last but not the least, it provides a different fix for the search input corners.

Having a `isOpen` in the reducer was misleading: with the recent `autocompleteCancel()` change (cancelling the API call and everything else), the input could be desynchronized. Now, the `SearchForm` is synchronized with `Autosuggest` at the component level and it works better.

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -35,7 +35,6 @@ export class SearchFormBase extends React.Component {
   static propTypes = {
     addonType: PropTypes.string,
     api: PropTypes.object.isRequired,
-    autocompleteIsOpen: PropTypes.bool.isRequired,
     className: PropTypes.string,
     debounce: PropTypes.func.isRequired,
     dispatch: PropTypes.func.isRequired,
@@ -60,8 +59,9 @@ export class SearchFormBase extends React.Component {
     super(props);
 
     this.state = {
-      searchValue: props.query || '',
+      autocompleteIsOpen: false,
       oldAutocompleteFilters: {},
+      searchValue: props.query || '',
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/amo/components/SearchForm/index.js
```diff
@@ -160,7 +166,7 @@ export class SearchFormBase extends React.Component {
       return;
     }
 
-    this.setState({ searchValue: '' }, () => {
+    this.setState({ autocompleteIsOpen: false, searchValue: '' }, () => {
       this.props.router.push(suggestion.url);
```
kumar303: I think you could safely remove `router.push()` from the `setState` callback too. I don't think it needs to wait for `setState` to resolve -- they are independent actions. It can be:

```js
this.setState({ autocompleteIsOpen: false, searchValue: '' });
this.props.router.push(suggestion.url);
```

src/amo/components/SearchForm/index.js
```diff
@@ -114,20 +114,32 @@ export class SearchFormBase extends React.Component {
...
 
-    const { addonType, dispatch, errorHandler } = this.props;
+    const { addonType } = this.props;
     const filters = { query: value };
 
     if (addonType) {
       filters.addonType = addonType;
     }
 
+    this.setState({ autocompleteIsOpen: true });
```
kumar303: To make things a little less complex, you could just add the filter change to this and remove it from the dispatch helper:

```js
this.setState({ 
  autocompleteIsOpen: true, oldAutocompleteFilters: filters, 
});
```

src/amo/components/SearchForm/index.js
```diff
@@ -114,20 +114,32 @@ export class SearchFormBase extends React.Component {
...
-  handleSuggestionsFetchRequested = this.props.debounce(({ value, reason }) => {
+  dispatchAutocompleteStart = this.props.debounce(({ filters }) => {
+    const { dispatch, errorHandler } = this.props;
+
+    dispatch(autocompleteStart({
+      errorHandlerId: errorHandler.id,
+      filters,
+    }));
+    this.setState({ oldAutocompleteFilters: filters });
+  }, 200, { leading: true })
```
kumar303: I know you're relying on defaults here but since you are already specifying an option, it looks like the `debounce` is configured for leading edge only. I would suggest adding them both: `{ leading: true, trailing: true }`.

kumar303: How does firing this off on the leading edge help? I don't think the first character that someone types would be useful.

willdurand: I inverted `leading` and `trailing` it seems.

willdurand:
I have made the requested changes and played with it. Seems to work the same (maybe better without leading edge debouncing 😋 )

PR #3083, opened 2017-09-08
More add-ons by authors UI

### LINKED ISSUE #2767: Add link to " More add-ons by these developers"

willdurand:
Fix #2767
~~⚠️ depends on #3027~~

---

This PR adds a new section on the add-on detail page to show related add-ons (made by any of the authors).

~~⚠️  I am waiting for https://github.com/mozilla/addons-server/issues/6346 to get more add-ons by authors. Right now, it does not work well (user `name` must be her `username`).~~ => available in `dev`.

The card is not displayed if there is no "other add-ons", which is the same behavior as the old website.

### Themes

screen shot 2017-09-08 at 18 38 12
screen shot 2017-09-08 at 18 38 27
screen shot 2017-09-08 at 18 38 21

### Extensions

screen shot 2017-09-08 at 18 38 53
screen shot 2017-09-08 at 18 39 05
screen shot 2017-09-08 at 18 39 01

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -125,10 +125,6 @@ export class AddonBase extends React.Component {
     this.props.toggleThemePreview(event.currentTarget);
   }
 
-  getAuthorUsernames(addon) {
-    return addon.authors.map((author) => author.username);
-  }
-
   getFeaturedText(addonType) {
     const { i18n } = this.props;
 
@@ -147,7 +143,7 @@ export class AddonBase extends React.Component {
 
     dispatch(fetchOtherAddonsByAuthors({
       addonType: addon.type,
-      authors: this.getAuthorUsernames(addon),
+      authors: addon.authors.map((author) => author.username),
       errorHandlerId: errorHandler.id,
       slug: addon.slug,
     }));
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -18,7 +18,12 @@ import { fetchAddon } from 'core/reducers/addons';
 import { withErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
 import {
-  ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME, ENABLED, UNKNOWN,
+  ADDON_TYPE_DICT,
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_LANG,
+  ADDON_TYPE_THEME,
+  ENABLED,
+  UNKNOWN,
 } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import {
@@ -75,26 +80,20 @@ export class AddonBase extends React.Component {
       params,
     } = this.props;
 
-    if (addon) {
-      dispatch(setViewContext(addon.type));
-      dispatch(fetchOtherAddonsByAuthors({
-        addonType: addon.type,
-        authors: this.getAuthorUsernames(addon),
-        errorHandlerId: errorHandler.id,
-        slug: addon.slug,
-      }));
-    } else {
-      dispatch(fetchAddon({ slug: params.slug, errorHandler }));
+    // This makes sure we do not try to dispatch any new actions in the case
+    // of an error.
+    if (!errorHandler.hasError()) {
+      if (addon) {
+        dispatch(setViewContext(addon.type));
+        this.dispatchFetchOtherAddonsByAuthors({ addon });
+      } else {
+        dispatch(fetchAddon({ slug: params.slug, errorHandler }));
+      }
     }
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
     const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
-    const {
-      addon: newAddon,
-      addonsByAuthors: newAddonsByAuthors,
-      params: newParams,
-    } = nextProps;
 
     const oldAddonType = oldAddon ? oldAddon.type : null;
     if (newAddon && newAddon.type !== oldAddonType) {
@@ -105,13 +104,8 @@ export class AddonBase extends React.Component {
       dispatch(fetchAddon({ slug: newParams.slug, errorHandler }));
     }
 
-    if (newAddon && oldAddon !== newAddon && !newAddonsByAuthors) {
-      dispatch(fetchOtherAddonsByAuthors({
-        addonType: newAddon.type,
-        authors: this.getAuthorUsernames(newAddon),
-        errorHandlerId: errorHandler.id,
-        slug: newAddon.slug,
-      }));
+    if (newAddon && oldAddon !== newAddon) {
+      this.dispatchFetchOtherAddonsByAuthors({ addon: newAddon });
     }
   }
 
@@ -146,6 +140,17 @@ export class AddonBase extends React.Component {
       default:
         return i18n.gettext('Featured Add-on');
     }
+  }
+
+  dispatchFetchOtherAddonsByAuthors({ addon }) {
+    const { dispatch, errorHandler } = this.props;
+
+    dispatch(fetchOtherAddonsByAuthors({
+      addonType: addon.type,
+      authors: this.getAuthorUsernames(addon),
+      errorHandlerId: errorHandler.id,
+      slug: addon.slug,
+    }));
   }
 
   headerImage({ compatible } = {}) {
@@ -310,13 +315,35 @@ export class AddonBase extends React.Component {
 
     let header;
     switch (addonType) {
+      case ADDON_TYPE_DICT:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More dictionaries by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More dictionaries by these translators'),
+          authorNames.length
+        );
+        break;
+
       case ADDON_TYPE_EXTENSION:
         header = i18n.ngettext(
           i18n.sprintf(
             i18n.gettext('More extensions by %(author)s'),
             { author: authorNames[0] }
           ),
-          i18n.gettext('More extensions by these authors'),
+          i18n.gettext('More extensions by these developers'),
+          authorNames.length
+        );
+        break;
+
+      case ADDON_TYPE_LANG:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More language packs by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More language packs by these translators'),
           authorNames.length
         );
         break;
@@ -327,7 +354,7 @@ export class AddonBase extends React.Component {
             i18n.gettext('More themes by %(author)s'),
             { author: authorNames[0] }
           ),
-          i18n.gettext('More themes by these authors'),
+          i18n.gettext('More themes by these artists'),
           authorNames.length
         );
         break;
@@ -338,7 +365,7 @@ export class AddonBase extends React.Component {
             i18n.gettext('More add-ons by %(author)s'),
             { author: authorNames[0] }
           ),
-          i18n.gettext('More add-ons by these authors'),
+          i18n.gettext('More add-ons by these developers'),
           authorNames.length
         );
     }
@@ -507,12 +534,12 @@ export function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
   const addon = state.addons[slug];
 
+  let addonsByAuthors;
   let installedAddon = {};
-  let addonsByAuthors = [];
 
   if (addon) {
-    installedAddon = state.installations[addon.guid] || {};
     addonsByAuthors = state.addonsByAuthors.byAddonSlug[addon.slug];
+    installedAddon = state.installations[addon.guid] || {};
   }
 
   return {
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -13,6 +13,7 @@ import NotFound from 'amo/components/ErrorPage/NotFound';
 import DefaultRatingManager from 'amo/components/RatingManager';
 import ScreenShots from 'amo/components/ScreenShots';
 import Link from 'amo/components/Link';
+import { fetchOtherAddonsByAuthors } from 'amo/reducers/addonsByAuthors';
 import { fetchAddon } from 'core/reducers/addons';
 import { withErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
@@ -34,6 +35,7 @@ import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
 import ShowMoreCard from 'ui/components/ShowMoreCard';
 import Badge from 'ui/components/Badge';
+import AddonsCard from 'amo/components/AddonsCard';
 
 import './styles.scss';
 
@@ -57,6 +59,7 @@ export class AddonBase extends React.Component {
     installStatus: PropTypes.string.isRequired,
     toggleThemePreview: PropTypes.func.isRequired,
     userAgentInfo: PropTypes.object.isRequired,
+    addonsByAuthors: PropTypes.array.isRequired,
   }
 
   static defaultProps = {
@@ -65,17 +68,33 @@ export class AddonBase extends React.Component {
   }
 
   componentWillMount() {
-    const { addon, dispatch, errorHandler, params } = this.props;
+    const {
+      addon,
+      dispatch,
+      errorHandler,
+      params,
+    } = this.props;
 
     if (addon) {
       dispatch(setViewContext(addon.type));
+      dispatch(fetchOtherAddonsByAuthors({
+        addonType: addon.type,
+        authors: this.getAuthorUsernames(addon),
+        errorHandlerId: errorHandler.id,
+        slug: addon.slug,
+      }));
     } else {
       dispatch(fetchAddon({ slug: params.slug, errorHandler }));
     }
   }
 
-  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
+  componentWillReceiveProps(nextProps) {
     const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
+    const {
+      addon: newAddon,
+      addonsByAuthors: newAddonsByAuthors,
+      params: newParams,
+    } = nextProps;
 
     const oldAddonType = oldAddon ? oldAddon.type : null;
     if (newAddon && newAddon.type !== oldAddonType) {
@@ -84,6 +103,15 @@ export class AddonBase extends React.Component {
 
     if (params.slug !== newParams.slug) {
       dispatch(fetchAddon({ slug: newParams.slug, errorHandler }));
+    }
+
+    if (newAddon && oldAddon !== newAddon && !newAddonsByAuthors) {
+      dispatch(fetchOtherAddonsByAuthors({
+        addonType: newAddon.type,
+        authors: this.getAuthorUsernames(newAddon),
+        errorHandlerId: errorHandler.id,
+        slug: newAddon.slug,
+      }));
     }
   }
 
@@ -101,6 +129,10 @@ export class AddonBase extends React.Component {
 
   onClick = (event) => {
     this.props.toggleThemePreview(event.currentTarget);
+  }
+
+  getAuthorUsernames(addon) {
+    return addon.authors.map((author) => author.username);
   }
 
   getFeaturedText(addonType) {
@@ -266,6 +298,67 @@ export class AddonBase extends React.Component {
     /* eslint-enable react/no-danger */
   }
 
+  renderMoreAddonsByAuthors = () => {
+    const { addon, addonsByAuthors, i18n } = this.props;
+
+    if (!addon || !addonsByAuthors || addonsByAuthors.length === 0) {
+      return null;
+    }
+
+    const addonType = addon.type;
+    const authorNames = addon.authors.map((author) => author.name);
+
+    let header;
+    switch (addonType) {
+      case ADDON_TYPE_EXTENSION:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More extensions by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More extensions by these authors'),
+          authorNames.length
+        );
+        break;
+
+      case ADDON_TYPE_THEME:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More themes by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More themes by these authors'),
+          authorNames.length
+        );
+        break;
+
+      default:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More add-ons by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More add-ons by these authors'),
+          authorNames.length
+        );
+    }
+
+    const classnames = classNames('AddonDescription-more-addons', {
+      'AddonDescription-more-addons--theme': addonType === ADDON_TYPE_THEME,
+    });
+
+    return (
+      <AddonsCard
+        addons={addonsByAuthors}
+        className={classnames}
+        header={header}
+        showMetadata={false}
+        showSummary={false}
+        type="horizontal"
+      />
+    );
+  }
+
   render() {
     const {
       addon,
@@ -402,6 +495,7 @@ export class AddonBase extends React.Component {
 
           {addon ? <AddonMoreInfo addon={addon} /> : null}
           {this.renderVersionReleaseNotes()}
+          {this.renderMoreAddonsByAuthors()}
         </div>
       </div>
     );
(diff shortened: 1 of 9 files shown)


src/amo/components/Addon/index.js
```diff
@@ -266,6 +298,67 @@ export class AddonBase extends React.Component {
...
+            i18n.gettext('More add-ons by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More add-ons by these authors'),
+          authorNames.length
+        );
+    }
+
+    const classnames = classNames('AddonDescription-more-addons', {
+      'AddonDescription-more-addons--theme': addonType === ADDON_TYPE_THEME,
```
willdurand: Do we want to support the `ADDON_TYPE_COMPLETE_THEME` here?

tofumatt: No, we shouldn't show those I'd say.

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+        addonsByAuthors,
+      });
+
+      expect(root.find('.AddonDescription-more-addons--theme'))
+        .toHaveLength(1);
+      expect(root.find(SearchResult)).toHaveLength(addonsByAuthors.length);
+    });
+  });
+
+  it('displays a badge when the addon is featured', () => {
```
willdurand: moved the tests below here because they were not in the appropriate `describe()`

src/amo/components/Addon/index.js
willdurand: Oki, this is not the case anyway because the API call passes the `addonType` as filters.

willdurand:
I think this is good for a quick review, especially visual feedback based on the screenshots.

src/amo/components/Addon/index.js
```diff
@@ -65,17 +68,33 @@ export class AddonBase extends React.Component {
...
+      }));
     } else {
       dispatch(fetchAddon({ slug: params.slug, errorHandler }));
     }
   }
 
-  componentWillReceiveProps({ addon: newAddon, params: newParams }) {
+  componentWillReceiveProps(nextProps) {
     const { addon: oldAddon, dispatch, errorHandler, params } = this.props;
+    const {
```
tofumatt: I mean we use this syntax often enough but it's actually fewer lines and arguably easier to grok to read:

```js
const newAddon = nextProps.addon;
const newAddonsByAuthors = nextProps.addonsByAuthors;
const newParams = nextProps.params;
```

but just being silly really, this is fine... 😆

src/amo/components/Addon/index.js
```diff
@@ -266,6 +298,67 @@ export class AddonBase extends React.Component {
...
+
+    let header;
+    switch (addonType) {
+      case ADDON_TYPE_EXTENSION:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More extensions by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More extensions by these authors'),
```
tofumatt: I think this should be "More extensions by these developers"

src/amo/components/Addon/index.js
```diff
@@ -266,6 +298,67 @@ export class AddonBase extends React.Component {
...
+        );
+        break;
+
+      case ADDON_TYPE_THEME:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More themes by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More themes by these authors'),
```
tofumatt: And this could be something cute like "More themes by these artists".

src/amo/components/Addon/index.js
```diff
@@ -266,6 +298,67 @@ export class AddonBase extends React.Component {
...
+        );
+        break;
+
+      default:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More add-ons by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More add-ons by these authors'),
```
tofumatt: I'd use "developers" here too.

Might be worth adding dedicated strings for "by these translators" for lang packs and dictionaries, if it's not a pain.

src/amo/components/Addon/index.js
tofumatt: Yeah I was gonna say in general we don't tend to get complete themes unless a specific add-on is loaded, usually.

src/amo/components/Addon/styles.scss
```diff
@@ -148,6 +148,46 @@
...
+      border-radius: $border-radius-default;
+    }
+
+    @include respond-to(large) {
+      .SearchResult.SearchResult--theme {
+        margin: 0 5px;
+      }
+    }
+
+    @include respond-to(desktop) {
```
tofumatt: We don't (or shouldn't) have a size named `desktop` referring to a device. @muffinresearch and I chatted about this ages ago so I can't find the thread but we decided we'd stick to using generic "size" descriptors. I think right now we have `medium`, `large`, and `extraLarge`.

src/amo/css/inc/vars.scss
```diff
@@ -18,6 +18,7 @@ $breakpoints: (
   medium: only screen and (min-width: 500px),
   large: only screen and (min-width: 720px),
   extraLarge: only screen and (min-width: 900px),
+  desktop: only screen and (min-width: 1025px),
```
tofumatt: `1025px` is a pretty large desktop! My own laptop browser isn't always that wide. It seems this should fix onto an `extraLarge` display, but if not we should at least stick to the size-type naming. Something like `fullWidth` or `largest`? 🤔

src/amo/components/SearchResult/index.js
```diff
@@ -19,15 +19,22 @@ export class SearchResultBase extends React.Component {
   static propTypes = {
     addon: PropTypes.object,
     i18n: PropTypes.object.isRequired,
+    showMetadata: PropTypes.bool,
+    showSummary: PropTypes.bool,
```
tofumatt: Neato... though couldn't we just do this with CSS? I don't think it hurts to render the content.

tofumatt: I guess if we NEVER want to render it this makes more sense, but it feels like something that could be done easier with CSS...

tests/unit/amo/components/TestAddon.js
```diff
@@ -101,6 +107,20 @@ function shallowRender(...args) {
   return shallow(<AddonBase {...props} />, { context: { i18n: props.i18n } });
 }
 
+// We use `mount` to test the i18n logic on the Addon component.
```
tofumatt: Can we not use `shallowUntilTarget`?

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+    });
+
+    it('does not fetch the other add-ons when they are already loaded', () => {
+      const fakeDispatch = sinon.stub();
+      const root = shallowRender({ addon: fakeAddon, dispatch: fakeDispatch });
+      fakeDispatch.reset();
+
+      const newAddon = { ...fakeAddon, slug: 'new-addon-slug' };
+      // These `addonsByAuthors` are related to the `newAddon` above, and it
+      // simulates the case when user has previously browsed this `newAddon`
```
tofumatt: I know I'm nitpicking but <code>had previously visited the `newAddon` page, navigated away, and then visited the page again</code> would be clearer I think 😄

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+        .toHaveLength(0);
+    });
+
+    it('is hidden when other addons are not loaded yet', () => {
+      const root = shallowRender({ addonsByAuthors: undefined });
+      expect(root.find('.AddonDescription-more-addons'))
+        .toHaveLength(0);
+    });
+
+    it('is hidden when there is no other addon', () => {
```
tofumatt: there are no other add-ons.

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+
+    it('displays the developer name when add-on is an extension', () => {
+      const root = mountMoreAddonsByAuthors({
+        addonsByAuthors: [fakeAddon],
+      });
+      expect(root.find('.Card-header'))
+        .toIncludeText('More extensions by Krupa');
+    });
+
+    it('displays the artist name when add-on is a theme', () => {
```
tofumatt: Hey look you even called them artists here 😄

tests/unit/amo/helpers.js
```diff
@@ -82,6 +83,7 @@ export const fakeTheme = Object.freeze({
   authors: [{
     name: 'Madonna',
```
tofumatt: This is unrelated but MaDonna is so great we should respect her real name. The "D" here should be uppercase; [that's how she styles it](https://addons.mozilla.org/en-GB/firefox/user/MaDonna/themes).

Technically her username is MaDonna too. 😄

src/amo/components/Addon/styles.scss
tofumatt: Also looking at the screenshots seems like this would work on `extraLarge`. But maybe I'm wrong–if we need the new size let's just relabel it.

src/amo/css/inc/vars.scss
willdurand: I was not used to such breakpoints to be honest 😄  (see Bootstrap ones for instance: https://getbootstrap.com/docs/3.3/css/#responsive-utilities-classes).

I was not happy with `extraExtraLarge`, hence `desktop` but I knew this would raise a red flag 😛  I am fine with not using device names.

What would be the "desktop" reference width by the way? I think 1000px is pretty common now (GitHub has 980px for example).

src/amo/components/Addon/index.js
```diff
@@ -85,6 +104,15 @@ export class AddonBase extends React.Component {
     if (params.slug !== newParams.slug) {
       dispatch(fetchAddon({ slug: newParams.slug, errorHandler }));
     }
+
+    if (newAddon && oldAddon !== newAddon && !newAddonsByAuthors) {
+      dispatch(fetchOtherAddonsByAuthors({
```
kumar303: you could DRY this up a bit by making a shared instance method:

```js
dispatch(this.fetchOtherAddonsByAuthors({ addon: newAddon }));
```

tests/unit/amo/components/TestAddon.js
willdurand: Doh! I changed this to singular form right before pushing it 😅

src/amo/components/Addon/index.js
```diff
@@ -408,9 +502,13 @@ export class AddonBase extends React.Component {
 export function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
   const addon = state.addons[slug];
+
   let installedAddon = {};
+  let addonsByAuthors = [];
```
kumar303: I think this should default to `undefined` which will be consistent with when they haven't loaded yet. I.E. `state.addonsByAuthors.byAddonSlug[addon.slug]` will return `undefined` before they are loaded.

src/amo/components/Addon/index.js
```diff
@@ -65,17 +68,33 @@ export class AddonBase extends React.Component {
...
+      addon,
+      dispatch,
+      errorHandler,
+      params,
+    } = this.props;
 
     if (addon) {
       dispatch(setViewContext(addon.type));
+      dispatch(fetchOtherAddonsByAuthors({
+        addonType: addon.type,
```
kumar303: It's kind of a bummer that you have to wait for `addon.type` here because it means the two add-on fetching API requests can't happen in parallel. This also means you could have gone with `byAddonId` in the state instead, which is a bit more stable. However, that's probably not worth changing.

src/amo/css/inc/vars.scss
kumar303: These breakpoints have been named after sizing, not device. Could we call it `extraWide` maybe?

src/core/types/addons.js
```diff
@@ -18,6 +18,7 @@ export type AddonVersionType = {|
 export type AddonAuthorType = {|
   name: string,
   url: string,
+  username: string,
```
kumar303: Hmm, this field isn't documented. Can you file a bug to update the docs? https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+
+    it('does not fetch the other add-ons when they are already loaded', () => {
+      const fakeDispatch = sinon.stub();
+      const root = shallowRender({ addon: fakeAddon, dispatch: fakeDispatch });
+      fakeDispatch.reset();
+
+      const newAddon = { ...fakeAddon, slug: 'new-addon-slug' };
+      // These `addonsByAuthors` are related to the `newAddon` above, and it
+      // simulates the case when user has previously browsed this `newAddon`
+      // and navigates to it again.
```
kumar303: I think for this case we should just load them again. This kind of caching has the potential to introduce a lot of subtle bugs so let's wait until we absolutely need it. The browser would have cached the API response so we probably don't even need to invent a new caching layer.

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+      expect(root.find('.AddonDescription-more-addons'))
+        .toHaveLength(0);
+    });
+
+    it('displays the developer name when add-on is an extension', () => {
+      const root = mountMoreAddonsByAuthors({
+        addonsByAuthors: [fakeAddon],
+      });
+      expect(root.find('.Card-header'))
+        .toIncludeText('More extensions by Krupa');
```
kumar303: The i18n components are a little tricky to test but you don't need `mount()`. You can still do it with `shallow()` by checking the HTML:

```js
expect(root.find('.Card-header').html())
  .toContain('More extensions by Krupa');
```

If you need to check for links or something you could do it like this:

```js
expect(root.find('.Card-header').render().find('a'))
  .toHaveLength(1);
```

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +771,242 @@ describe('Addon', () => {
...
+          type: ADDON_TYPE_OPENSEARCH,
+        },
+        addonsByAuthors: [fakeAddon],
+      });
+      expect(root.find('.Card-header'))
+        .toIncludeText('More add-ons by these authors');
+    });
+
+    it('displays more add-ons by authors', () => {
+      const addonsByAuthors = [
```
kumar303: Unless you can convince me otherwise, I think all mapped props should be set with a dispatch, not raw state. This helps keep everything in sync.

```js
store.dispatch(loadOtherAddonsByAuthors({ 
  slug: fakeAddon.slug, 
  addons: [
    { ...fakeAddon, slug: 'addon-1' },
    { ...fakeAddon, slug: 'addon-2' },
    { ...fakeAddon, slug: 'addon-3' },
  ], 
}));
const root = mountMoreAddonsByAuthors({ store });
```

We usually create a helper specific to the test suite like `_loadOtherAddonsByAuthors()`.

src/amo/css/inc/vars.scss
willdurand: Ok actually `extraLarge` is quite ok.

src/core/types/addons.js
willdurand: It seems to be documented here https://github.com/mozilla/addons-server/pull/6385/files#diff-d054974b06ee8bc0688e68b6331256ec but not deployed yet.

tests/unit/amo/components/TestAddon.js
willdurand: Oki

tests/unit/amo/components/TestAddon.js
willdurand: I don't see how to do that yet since switching to `mount` gives these errors:

```
  ● Addon › more add-ons by authors › indicates when other add-ons are themes

    TypeError: Cannot read property 'sprintf' of undefined

      at SearchResultBase.renderResult (src/amo/components/SearchResult/index.js:114:27)
      at SearchResultBase.render (src/amo/components/SearchResult/index.js:127:25)
```

tofumatt: You'll need to use:

```js

shallowUntilTarget(<MyComponent i18n={getFakeI18nInst()} />, MyComponentBase);
```

willdurand: I'll give it a try but I thought that was already there: https://github.com/willdurand/addons-frontend/blob/0ddc2d7368430cd8a7fcd48e4cd82cdc53b3c0cd/tests/unit/amo/components/TestAddon.js#L105-L108. Does `shallowUntilTarget` do something different?

tofumatt: `shallowUntilTarget` will use the wrapped component which will automatically add in the `i18n` to the props via the `compose(translate(), [...])` wrapping the component.

tofumatt: So come to think of it you might not even need to add in the fakeI18n thing, but I forget 😆

src/core/types/addons.js
kumar303: I think RTD is having trouble updating. I'm checking on it...

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +787,284 @@ describe('Addon', () => {
...
+        { ...fakeAddon, slug: 'addon-1' },
+        { ...fakeAddon, slug: 'addon-2' },
+        { ...fakeAddon, slug: 'addon-3' },
+      ];
+
+      store.dispatch(_loadAddon({ addon }));
+      store.dispatch(_loadOtherAddonsByAuthors({ addon, addonsByAuthors }));
+
+      const mappedProps = mapStateToProps(
```
kumar303: This approach looks good but I think you should take this opportunity to render it properly with `shallowUntilTarget`:

```diff
diff --git a/tests/unit/amo/components/TestAddon.js b/tests/unit/amo/components/TestAddon.js
index 3c04fe9c..bcc2ad55 100644
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -10,7 +10,7 @@ import { Provider } from 'react-redux';
 import { match } from 'react-router';
 
 import { setViewContext } from 'amo/actions/viewContext';
-import {
+import Addon, {
   AddonBase,
   mapStateToProps,
 } from 'amo/components/Addon';
@@ -58,6 +58,7 @@ import {
   createStubErrorHandler,
   getFakeI18nInst,
   sampleUserAgentParsed,
+  shallowUntilTarget,
 } from 'tests/unit/helpers';
 import ErrorList from 'ui/components/ErrorList';
 import LoadingText from 'ui/components/LoadingText';
@@ -107,6 +108,11 @@ function renderAsDOMNode(...args) {
   return findDOMNode(root);
 }
 
+function renderComponent(...args) {
+  const props = renderProps(...args);
+  return shallowUntilTarget(<Addon {...props} />, AddonBase);
+}
+
 function shallowRender(...args) {
   const props = renderProps(...args);
   return shallow(<AddonBase {...props} />, { context: { i18n: props.i18n } });
@@ -976,11 +982,9 @@ describe(__filename, () => {
       store.dispatch(_loadAddon({ addon }));
       store.dispatch(_loadOtherAddonsByAuthors({ addon, addonsByAuthors }));
 
-      const mappedProps = mapStateToProps(
-        store.getState(), { params: { slug: addon.slug } }
-      );
-
-      const root = shallowRender({ ...mappedProps });
+      const root = renderComponent({
+        params: { slug: addon.slug }, store,
+      });
 
       expect(root.find('.AddonDescription-more-addons--theme'))
         .toHaveLength(0);
@@ -1001,11 +1005,9 @@ describe(__filename, () => {
       store.dispatch(_loadAddon({ addon }));
       store.dispatch(_loadOtherAddonsByAuthors({ addon, addonsByAuthors }));
 
-      const mappedProps = mapStateToProps(
-        store.getState(), { params: { slug: addon.slug } }
-      );
-
-      const root = shallowRender({ ...mappedProps });
+      const root = renderComponent({
+        params: { slug: addon.slug }, store,
+      });
 
       expect(root.find('.AddonDescription-more-addons--theme'))
         .toHaveLength(1);
```

tests/unit/amo/components/TestAddon.js
```diff
@@ -748,6 +787,284 @@ describe('Addon', () => {
...
+        },
+        addonsByAuthors: [fakeAddon],
+      });
+      expect(root.find('.AddonDescription-more-addons'))
+        .toHaveProp('header', 'More language packs by these translators');
+    });
+
+    it('displays a plural form when dictionary has multiple authors', () => {
+      const root = shallowRender({
+        addon: {
```
kumar303: I know this test suite has ancient lore from every period in history so it's not worth converting the whole thing to `shallowUntilTarget` but ideally we would. I am uneasy that you are adding a bunch of this though:

```js
const root = shallowRender({
  addon: ...,
  addonsByAuthors: ...,
});
```

I would like to see these using the `dispatch` approach along with `shallowUntilTarget`, if possible. If creating `store` each time becomes too repetitive, you can move it to `beforeEach()` and then change `_loadAddon()` to reference the global store. This is what we do in other modern tests.

tests/unit/amo/components/TestAddon.js
willdurand: thanks for the patch!

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +810,268 @@ describe('Addon', () => {
       );
     });
   });
+
+  describe('more add-ons by authors', () => {
+    const dispatchLoadAddonsAndOtherAddonsByAuthors = ({
```
kumar303: I'd say you can just call this `dispatchAddonData` since it's only used within the context of the 'more add-ons by authors' test.

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +810,268 @@ describe('Addon', () => {
...
+      addon, addonsByAuthors,
+    }) => {
+      const { store } = dispatchClientMetadata();
+
+      store.dispatch(_loadAddons({ addon }));
+      store.dispatch(_loadOtherAddonsByAuthors({ addon, addonsByAuthors }));
+
+      return {
+        store,
+        state: store.getState(),
```
kumar303: this property doesn't appear to be necessary

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +810,268 @@ describe('Addon', () => {
...
+      });
+      const root = renderComponent({ params: { slug: addon.slug }, store });
+
+      return root.find('.AddonDescription-more-addons');
+    };
+
+    it('fetches the other add-ons by authors', () => {
+      const addon = fakeAddon;
+      const fakeDispatch = sinon.stub();
+      shallowRender({ addon, dispatch: fakeDispatch });
```
kumar303: You should be able to do it this way:

```js
const { store } = dispatchLoadAddonsAndOtherAddonsByAuthors({ addon });
const fakeDispatch = sinon.spy(store, 'dispatch');

sinon.assert.calledWith(fakeDispatch, ...);
```

willdurand:
`shallowRender` are gone now, thanks @kumar303!

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +823,277 @@ describe('Addon', () => {
...
+
+    it('does not fetch the other add-ons when add-on is the same', () => {
+      const addon = fakeAddon;
+      const { store } = dispatchAddonData({ addon });
+      const fakeDispatch = sinon.spy(store, 'dispatch');
+
+      const root = renderComponent({ params: { slug: addon.slug }, store });
+      // `fakeAddon` is the API representation of an add-on, which is slightly
+      // different than the representation of an add-on in the state. That is
+      // we need to retreive the add-on from the state.
```
kumar303: typo: "That is why we..." and "...retrieve".

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +823,277 @@ describe('Addon', () => {
...
+      const addonFromState = store.getState().addons[addon.slug];
```
kumar303: Maybe this would be better as ...?

```js
const addonFromState = mapStateToProps(store.getState()).addon;
```

tests/unit/amo/components/TestAddon.js
```diff
@@ -768,6 +823,277 @@ describe('Addon', () => {
...
+
+    it('fetches the other add-ons when a new add-on is loaded', () => {
+      const addon = fakeAddon;
+      const { store } = dispatchAddonData({ addon });
+      const fakeDispatch = sinon.spy(store, 'dispatch');
+
+      const root = renderComponent({ params: { slug: addon.slug }, store });
+      fakeDispatch.reset();
+
+      const newAddon = { ...fakeAddon, slug: 'new-addon-slug' };
```
kumar303: I suppose you could fix this one too. It's not a big deal though.

PR #3089, opened 2017-09-09
chore: Reorganise API code and tests

tofumatt:
This just separates out a lot of the code in core/api. I was finding it a real pain to read that file and while we've moved to using separate files for new code I wanted to separate it out because the tests and API code is hard to make sense of right now.

This didn't take too long and is solely reorganising. If you think it's not worth the time though just say so and you can close this, it's not the hugest deal.

### DIFF
--- a/src/core/api/authentication.js
+++ b/src/core/api/authentication.js
@@ -1,0 +1,55 @@
+import url from 'url';
+
+import config from 'config';
+
+import { API_BASE, callApi, makeQueryString } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+import type { ReactRouterLocation } from 'core/types/router';
+
+
+type LoginParams = {|
+  api: ApiStateType,
+  code: string,
+  state: string,
+|};
+
+export function login({ api, code, state }: LoginParams) {
+  const params = { config: undefined };
+  const configName = config.get('fxaConfig');
+  if (configName) {
+    params.config = configName;
+  }
+  return callApi({
+    endpoint: 'accounts/login',
+    method: 'POST',
+    body: { code, state },
+    params,
+    state: api,
+    credentials: true,
+  });
+}
+
+export function logOutFromServer({ api }: {| api: ApiStateType |}) {
+  return callApi({
+    auth: true,
+    credentials: true,
+    endpoint: 'accounts/session',
+    method: 'DELETE',
+    state: api,
+  });
+}
+
+export function startLoginUrl(
+  { location }: {| location: ReactRouterLocation |},
+) {
+  const configName = config.get('fxaConfig');
+  const params = {
+    config: undefined,
+    to: url.format({ ...location }),
+  };
+  if (configName) {
+    params.config = configName;
+  }
+  const query = makeQueryString(params);
+  return `${API_BASE}/accounts/login/start/${query}`;
+}
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -4,25 +4,24 @@ import url from 'url';
 
 import utf8 from 'utf8';
 import 'isomorphic-fetch';
-import { schema as normalizrSchema, normalize } from 'normalizr';
+import { normalize } from 'normalizr';
 import { oneLine } from 'common-tags';
 import config from 'config';
 
 import { initialApiState } from 'core/reducers/api';
 import log from 'core/logger';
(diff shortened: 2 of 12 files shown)


src/core/api/index.js
```diff
@@ -162,111 +161,22 @@ export function callApi({
...
+export const login = authenticationApi.login;
+export const logOutFromServer = authenticationApi.logOutFromServer;
+export const startLoginUrl = authenticationApi.startLoginUrl;
+// Category APIs
+export const categories = categoriesApi.categories;
+// Featured Add-on APIs
+export const featured = featuredApi.featured;
+// Search and Autocomplete APIs
+export const autocomplete = searchApi.autocomplete;
+export const search = searchApi.search;
```
willdurand: Can we stop doing this maybe? See: https://github.com/mozilla/addons-frontend/pull/3027#discussion_r137825532.

tofumatt: Yes I would like that... but it was more work especially with normalizr. I'll see about removing it in another commit though!

willdurand: Can't we just add per API `import` where we rely on `core/api`?

tofumatt: We can but there are some bugs with normalizr and stuff in the tests. But I'll give it a go again... 😄

tofumatt: Okay, I just spent a few hours trying to remove these imports and going through the code/tests changing calls from `import * as api from 'core/api'` to more specific imports but it's a total mess. Even after some basic cleanup there are like 50 failing tests and the causes aren't obvious. I feel like this branch is a definite improvement and moving to importing the functions from each file will be a lot easier once we ditch normalizr (#2917).

So I'm just gonna propose the PR as it is. I think it's a better jumping off point to fix #2917 but I don't want to go down the rabbit hole any further for now.

willdurand: Oh ok, sorry for the extra work 😔

tofumatt: Oh, no worries. It was worth trying but I realised eventually it just wasn't worth pursuing right now 😄

src/core/api/index.js
```diff
@@ -162,111 +161,22 @@ export function callApi({
...
-
-export const search = searchApi;
+// Schemas
+// TODO: Remove this when
+// https://github.com/mozilla/addons-frontend/issues/2917 is fixed.
+export const addon = addonApi.addonSchema;
+export const category = categoriesApi.categorySchema;
+
+// Add-on APIs
+export const fetchAddon = addonApi.fetchAddon;
```
kumar303: I don't see the benefit in aliasing all of these submodules. It just creates confusion and requires a bunch of boilerplate code. Why can't we just update all of the import statements? If the test failures are confusing, you could could push a commit with the aliases removed and I'll take a look.

src/core/api/authentication.js
```diff
@@ -0,0 +1,55 @@
+import url from 'url';
```
kumar303: All of these new files need `/* @flow */` at the top

tofumatt:
My thought was it's easier to read the files and deal with the code/tests when it's not a tonne of barely-related functions inside the same file. It's easier at a glance to see "oh this is the file for search, this one for categories". Aside from using the `callApi` method there is little in common with these functions.

Aliasing things means the code works as before but is at least more organised... especially the tests. The aliases are there because of the way the tests do:

```js
import * as api from 'core/api';

const mockApi = sinon.mock(api);
```

kumar303:
```js
import * as api from 'core/api';

const mockApi = sinon.mock(api);
```

This seems easy enough to change, no?

```js
// In test_search.js, for example...

import * as api from 'core/api/search';

const mockApi = sinon.mock(api);
```

If you need two mocks in the same test file then some tests probably need moving to their respective file. If it doesn't make sense to move them you can just create two mocks:

```js
import * as searchApi from 'core/api/search';
import * as categoriesApi from 'core/api/categories';

const mockSearchApi = sinon.mock(searchApi);
const mockCategoriesApi = sinon.mock(categoriesApi);
```

tofumatt:
I didn't try that approach but normalizr would sometimes throw errors if I tried something similar. I'll try again but more than an hour's work and I'd say best to save this for later...

tofumatt:
Bit rot. I'll leave the branch here as a template, maybe.

PR #3092, opened 2017-09-11
Restrict max content width

### LINKED ISSUE #2680: Restrict max-width of content
There's a few pages that don't look too good if run on a maximized browser with lots of pixels.

E.g:

add-ons_for_firefox

Would this work better if constrained? E.g: here's an example using 1400px:

add-ons_for_firefox

muffinresearch:
Fixes #2680

Before:

screen shot 2017-09-11 at 18 33 20

After: 

screen shot 2017-09-11 at 18 29 59

### DIFF
--- a/src/amo/components/App/styles.scss
+++ b/src/amo/components/App/styles.scss
@@ -44,6 +44,11 @@ body {
   padding: 0 0 $padding-page;
 }
 
+.App-content-wrapper {
+  margin: 0 auto;
+  max-width: $max-content-width;
+}
+
 h1,
 h2 {
   @include font-bold();
--- a/src/amo/components/App/index.js
+++ b/src/amo/components/App/index.js
@@ -209,9 +209,11 @@ export class AppBase extends React.Component {
             ref={(ref) => { this.header = ref; }}
           />
           <div className="App-content">
-            <ErrorPage getErrorComponent={getErrorComponent}>
-              {children}
-            </ErrorPage>
+            <div className="App-content-wrapper">
+              <ErrorPage getErrorComponent={getErrorComponent}>
+                {children}
+              </ErrorPage>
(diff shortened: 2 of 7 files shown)


src/amo/components/App/styles.scss
```diff
@@ -44,6 +44,11 @@ body {
   padding: 0 0 $padding-page;
 }
 
+.App-content > div {
```
muffinresearch: Thinking a dedicated wrapper would be a plan, to save the possibility of this being brittle in the future.

muffinresearch:
@pwalm here's a quick example of limiting the content width does this look to match what you expected?

tofumatt:
What happens here with the review window modal? Does the blue fill the page or just the content's max-width?

src/amo/components/App/styles.scss
tofumatt: Yeah this should be changed I'd say before we merge, it should be easy, right?

muffinresearch:
I'm also trying some other options I'm not sure the full-width header/footer are that nice:

Here's an example with just a white background:

full-screen-white-background

muffinresearch:
And with the photon background colour: 

add-ons_for_firefox

muffinresearch:
Hmm going back to the original layout as per this PR, it's better than these last two actually.

willdurand:
I also prefer the full-width header/footer.

pwalm:
full-width header/footer ftw! Looks great. :D

muffinresearch:
Ok, I'll fix the wrapper and we can try it out for real. Thanks!

muffinresearch:
@tofumatt - good call I've pushed a fix for the modals too:

add-ons_for_firefox

PR #3093, opened 2017-09-12
Tighten up header spec

muffinresearch:
A few tweaks to tighten up the spec on the header. There's some ambiguity in the mocks/specs re:  the gutter size. I've gone for the smaller one represented in the mocks. We can always adjust with the var later. 

Overall this gives the logo more room to breath. I quite like how the intended design has the header and footer on a different grid to the main content.

Before large:
screen shot 2017-09-12 at 10 43 44

Before medium:
screen shot 2017-09-12 at 10 44 11

After large:
screen shot 2017-09-12 at 10 43 29

After medium (updated)
screen shot 2017-09-12 at 11 28 59

### DIFF
--- a/src/amo/components/Footer/styles.scss
+++ b/src/amo/components/Footer/styles.scss
@@ -10,6 +10,7 @@
   @include respond-to(medium) {
     display: flex;
     justify-content: space-between;
+    padding: 20px $header-footer-gutter;
   }
 
   @include respond-to(large) {
@@ -48,8 +49,6 @@
   margin-top: 10px;
 
   @include respond-to(medium) {
-    @include margin-end(10px);
-
     margin-top: auto;
   }
 }
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -14,6 +14,7 @@
   @include respond-to(medium) {
     grid-template-columns: auto auto;
     min-height: 98px;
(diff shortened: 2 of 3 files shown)


tofumatt:
I don't see mobile screenshots but my chief concern is the explore/extensions/themes like on a tablet. They're really jammed into the home page link. I'm not a huge fan of the aesthetics of it but worse I think those are way too close of touch targets.

How did this feel on a tablet or large mobile?

muffinresearch:
@tofumatt I left mobile out of it since they're unaffected by these changes since all modifications are to the medium breakpoint.

I can add a bit more space between the logo wordmark and the nav for medium easily enough to match the space that was there before. Good catch 👍

> I'm not a huge fan of the aesthetics of it...

This way is closer to the spec (albeit unless I've missed something I don't think there is a medium spec to follow), and I think a bit more space under and around the header feels right. The bit I'm less sure about is the grid breaking aspect it's a bit quirky but that seems to be as designed - FWIW the spec has a an even bigger left/right gutter for the header `32px` instead of `24px`.

muffinresearch:
Ok, I've reduced the padding at the bottom of the medium breakpoint and updated the medium screenshot.

tofumatt:
I actually meant the aesthetics of the header link being close to the section links. The other padding is cool. I'll have another look, thanks!

PR #3094, opened 2017-09-12
I3061: Handles 401 error for unathuorized addons

### LINKED ISSUE #3061: [layout issue] Unapproved addon should give 'Not Found' error
Navigate to https://addons.mozilla.org/en-US/firefox/addon/share-backported/

### What happened?
Gives "You do not have permission to perform this action." error
(Its an unapproved addon [under review])

Addon detail page should be visible as in legacy website.
[update] As per new frontend specifications this should throw 404 error.

**Mobile website view**
screen shot 2017-09-03 at 12 42 14 pm

**Legacy view**
screen shot 2017-09-03 at 12 42 28 pm

Also layout shows loading icons, which shouldn't be visible in any case other than loading time.

tsl143:
Fixes #3061


tsl143:
Merged wrong commits

PR #3095, opened 2017-09-12
fix:Handles 401 error for unathuorized addons

tsl143:
Fixes: #3061 

Unable to have screenshot as no addon gives `401` at local (or may be I am just unaware)

But I got 401 error from `addons.mozilla.org`
screen shot 2017-09-12 at 6 11 24 pm

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -280,8 +280,8 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.error('Captured API Error:', errorHandler.capturedError);
-      if (errorHandler.capturedError.responseStatusCode === 404 ||
-          errorHandler.capturedError.responseStatusCode === 401
+      if (errorHandler.capturedError.responseStatusCode === 401 ||
+          errorHandler.capturedError.responseStatusCode === 404
       ) {
         return <NotFound />;
       }
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -279,7 +279,9 @@ export class AddonBase extends React.Component {
 
     let errorBanner = null;
     if (errorHandler.hasError()) {
-      log.error('Captured API Error:', errorHandler.capturedError);
+      log.warn('Captured API Error:', errorHandler.capturedError);
+      // A 401 is made to look like a 404 on purpose.
+      // See https://github.com/mozilla/addons-frontend/issues/3061
       if (errorHandler.capturedError.responseStatusCode === 401 ||
           errorHandler.capturedError.responseStatusCode === 404
       ) {
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -280,7 +280,9 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.error('Captured API Error:', errorHandler.capturedError);
-      if (errorHandler.capturedError.responseStatusCode === 404) {
+      if (errorHandler.capturedError.responseStatusCode === 404 ||
+          errorHandler.capturedError.responseStatusCode === 401
+      ) {
         return <NotFound />;
       }
       // Show a list of errors at the top of the add-on section.
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -41,7 +41,7 @@ import {
 import InstallButton from 'core/components/InstallButton';
 import { ErrorHandler } from 'core/errorHandler';
 import I18nProvider from 'core/i18n/Provider';
-import { dispatchSignInActions, fakeAddon } from 'tests/unit/amo/helpers';
+import { dispatchSignInActions, dispatchClientMetadata, fakeAddon } from 'tests/unit/amo/helpers';
 import {
   createFetchAddonResult,
   createStubErrorHandler,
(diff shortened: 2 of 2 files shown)


tests/unit/amo/components/TestAddon.js
```diff
@@ -255,6 +255,28 @@ describe('Addon', () => {
...
+      response: { status: 401 },
+      apiURL: 'https://some/api/endpoint',
+      jsonResponse: { message: 'Authentication credentials were not provided.' },
+    });
+    store.dispatch(setError({ id, error }));
+    const capturedError = store.getState().errors[id];
+    expect(capturedError).toBeTruthy();
+
+    // Set up an error handler from state like withErrorHandler().
+    const errorHandler = new ErrorHandler({
```
willdurand: I think you can actually use `createStubErrorHandler()` from `tests/unit/helpers`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -255,6 +255,28 @@ describe('Addon', () => {
...
+    const { store } = createStore();
+
+    const error = createApiError({
+      response: { status: 401 },
+      apiURL: 'https://some/api/endpoint',
+      jsonResponse: { message: 'Authentication credentials were not provided.' },
+    });
+    store.dispatch(setError({ id, error }));
+    const capturedError = store.getState().errors[id];
+    expect(capturedError).toBeTruthy();
```
willdurand: Do we actually need this? It does not seem to cover any code that you have changed or am I missing something?

tests/unit/amo/components/TestAddon.js
tofumatt: We have a test helper for this: `import { createStubErrorHandler } from 'tests/unit/helpers';` and then you can just do: `createStubErrorHandler(capturedError);`

tests/unit/amo/components/TestAddon.js
```diff
@@ -255,6 +255,28 @@ describe('Addon', () => {
     expect(root.find(NotFound)).toHaveLength(1);
   });
 
+  it('renders NotFound page for unauthorized add-on - 401 error', () => {
+    const id = 'error-handler-id';
+    const { store } = createStore();
```
tofumatt: You should use `dispatchClientMetadata()` here to get a store as usually initialised by the app instead. It's in `tests/unit/amo/helpers`. But you can use it in the exact same way as `createStore()`, so: `const { store } = dispatchClientMetadata();` will work 😄

tests/unit/amo/components/TestAddon.js
tofumatt: This check could be a bit more specific; maybe test the text of the error message?

src/amo/components/Addon/index.js
```diff
@@ -280,7 +280,9 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.error('Captured API Error:', errorHandler.capturedError);
-      if (errorHandler.capturedError.responseStatusCode === 404) {
+      if (errorHandler.capturedError.responseStatusCode === 404 ||
+          errorHandler.capturedError.responseStatusCode === 401
```
tofumatt: Two points:

1. nitpick, but could you put 401 first? Just find it easier to scan with incremental counts
2. I think you should add a comment explaining why this is here--even just something like "We send 404 for unauthorised add-ons to prevent unauthorised users from even knowing if the add-on exists." and then a link to the issue

tsl143:
@tofumatt , updated as required,
I need to dive more into helpers they are actually helping :P

tests/unit/amo/components/TestAddon.js
```diff
@@ -255,6 +255,24 @@ describe('Addon', () => {
...
+    const id = 'error-handler-id';
+    const store = dispatchClientMetadata().store;
+
+    const error = createApiError({
+      response: { status: 401 },
+      apiURL: 'https://some/api/endpoint',
+      jsonResponse: { message: 'Authentication credentials were not provided.' },
+    });
+    store.dispatch(setError({ id, error }));
+    const capturedError = store.getState().errors[id];
```
kumar303: I know you moved it per request but I think you should put this line back, including the comment:

```js
// This makes sure the error was dispatched to state correctly.
expect(capturedError).toBeTruthy();
```

This pattern is a bit clunky (sorry) but I think it's the right way to do it for now.

src/amo/components/Addon/index.js
```diff
@@ -280,7 +280,9 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.error('Captured API Error:', errorHandler.capturedError);
-      if (errorHandler.capturedError.responseStatusCode === 404) {
+      if (errorHandler.capturedError.responseStatusCode === 401 ||
```
kumar303: I know Jorge said we should do this but I think it's a bit odd so we should add this comment to prevent a code reader from thinking it was done by mistake:

```js
// A 401 is made to look like a 404 on purpose.
// See https://github.com/mozilla/addons-frontend/issues/3061
```

src/amo/components/Addon/index.js
```diff
@@ -280,7 +280,9 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.error('Captured API Error:', errorHandler.capturedError);
```
kumar303: While you're in here, could you make this `log.warn` instead of `log.error`? That was a mistake that I'm only noticing right now. When we use `log.error` it shows up in our error log as something we need to fix. This is not something that needs fixing; it is an expected outcome of the user's action.

tsl143:
@kumar303 I did the changes,
comments make it a bit more understandable :)
But I really don't understand why should we use
`// This makes sure the error was dispatched to state correctly.`
`expect(capturedError).toBeTruthy();`

PR #3113, opened 2017-09-13
Use photon colors for log in button

### LINKED ISSUE #3112: Update sign-in button link to use photon colours
Currently looks like:

add-ons_for_firefox

Mocks have:

amo_desktop_-_master_sketch

muffinresearch:
Fixes #3112

Before:

add-ons_for_firefox

After:

add-ons_for_firefox


muffinresearch:
> Looks good, but one thing though: don't you have to import @import "~photon-colors/colors"?

It's currently working implicitl, but that's a good catch, it'll make sense to make it explicit.

PR #3116, opened 2017-09-13
Update category link colours

### LINKED ISSUE #3114: Update category colours to use photon colours
Currently we have:

add-ons_for_firefox

The mocks have:
amo_desktop_-_master_sketch

muffinresearch:
Fixes #3114 

This updates the category colours to match the current spec. Note the colours in the spec are not actually photon colours but they're photon-like hence not using the photon vars.

I removed the use of button default classes to reduce the specificity issues and make these links their own thing. This enabled dedicated hover/focus styling much more easily and avoiding the pale blue default.

The photon colours removal was to make sure we're using the packaged vars not hard-coded ones - despite me not using the vars in this patch the import was needed to expose the photon vars elsewhere.

I removed another un-used CSS variable here too.

Before: 

add-ons_for_firefox

After:

add-ons_for_firefox

### DIFF
--- a/src/amo/components/Categories/index.js
+++ b/src/amo/components/Categories/index.js
@@ -137,7 +137,8 @@ export class CategoriesBase extends React.Component {
               return (
                 <li className="Categories-item" key={name}>
                   <Button
-                    className={`Categories-link Categories--category-color-${getCategoryColor(category)}`}
+                    className={`Categories-link
+                      Categories--category-color-${getCategoryColor(category)}`}
                     to={`/${visibleAddonType(addonType)}/${slug}/`}
                   >
                     {name}
--- a/src/amo/components/Categories/index.js
+++ b/src/amo/components/Categories/index.js
@@ -137,9 +137,7 @@ export class CategoriesBase extends React.Component {
               return (
                 <li className="Categories-item" key={name}>
                   <Button
-                    className={`Categories-link Button--action
-                      Button--small
-                      Categories--category-color-${getCategoryColor(category)}`}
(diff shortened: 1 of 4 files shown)


src/amo/components/Categories/index.js
```diff
@@ -137,9 +137,7 @@ export class CategoriesBase extends React.Component {
               return (
                 <li className="Categories-item" key={name}>
                   <Button
-                    className={`Categories-link Button--action
-                      Button--small
-                      Categories--category-color-${getCategoryColor(category)}`}
+                    className={`Categories-link Categories--category-color-${getCategoryColor(category)}`}
```
willdurand: Could you please wrap at 80 chars? I think that was the purpose of the previous indentation.

src/amo/components/Categories/styles.scss
```diff
@@ -14,15 +14,34 @@
 }
 
 .Categories-item {
-  @include margin-end(5px);
+  @include margin-end(7px);
 
   display: inline-block;
-  margin: 0 0 5px;
+  margin: 0 0 7px;
```
willdurand: Does these two values (`7px`) are somehow related? If so, maybe we could have a variable.

willdurand: Does these two values (`7px`) are somehow related? If so, maybe we could have a variable.

src/amo/components/Categories/index.js
willdurand: Could you please wrap at 80 chars? I think that was the purpose of the previous indentation.

PR #3117, opened 2017-09-13
Add more links to the menu

willdurand:
Refs #2745

---

This PR adds more links to the menu to be able to submit add-ons. I also took the opportunity to clean a few things related to the menu.

screen shot 2017-09-13 at 15 57 13

capture d ecran - 2017-09-13 a 15 57 22

### DIFF
--- a/src/amo/components/Link.js
+++ b/src/amo/components/Link.js
@@ -19,7 +19,6 @@ export class LinkBase extends React.Component {
     prependClientApp: PropTypes.bool,
     prependLang: PropTypes.bool,
     to: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
-    dispatch: PropTypes.func,
   }
 
   static defaultProps = {
@@ -50,7 +49,6 @@ export class LinkBase extends React.Component {
       clientApp,
       children,
       external,
-      dispatch,
       href,
       lang,
       prependClientApp,
--- a/src/amo/components/Link.js
+++ b/src/amo/components/Link.js
@@ -19,7 +19,7 @@ export class LinkBase extends React.Component {
     prependClientApp: PropTypes.bool,
     prependLang: PropTypes.bool,
     to: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
-    dispatch: PropTypes.func.isRequired,
+    dispatch: PropTypes.func,
   }
 
   static defaultProps = {
--- a/src/ui/components/DropdownMenu/index.js
+++ b/src/ui/components/DropdownMenu/index.js
@@ -11,7 +11,8 @@ import './styles.scss';
 
 type Props = {|
   text: string,
-  children: Array<DropdownMenuItem>,
+  // TODO: Replace Array by React.ChildrenArray when we upgrade Flow to 0.53+.
+  children?: Array<React.Element<typeof DropdownMenuItem>>,
   className?: string,
 |};
 
--- a/tests/unit/amo/components/TestHeader.js
+++ b/tests/unit/amo/components/TestHeader.js
@@ -6,7 +6,6 @@ import Link from 'amo/components/Link';
 import AuthenticateButton from 'core/components/AuthenticateButton';
 import DropdownMenu from 'ui/components/DropdownMenu';
 import DropdownMenuItem from 'ui/components/DropdownMenuItem';
-import * as api from 'core/api';
 import { VIEW_CONTEXT_HOME } from 'core/constants';
 import {
   dispatchClientMetadata,
@@ -77,17 +76,13 @@ describe(__filename, () => {
 
   it('allows a signed-in user to log out', () => {
     const { store } = dispatchSignInActions({ username: 'babar' });
-    const wrapper = renderHeader({ store });
-    const mockApi = sinon.mock(api);
+    const handleLogOut = sinon.stub();
 
-    mockApi
-      .expects('logOutFromServer')
-      .once()
-      .returns(Promise.resolve());
+    const wrapper = renderHeader({ store, handleLogOut });
 
     const onClick = wrapper.find(DropdownMenuItem).last().prop('onClick');
     onClick(createFakeEvent());
 
-    mockApi.verify();
+    sinon.assert.called(handleLogOut);
   });
 });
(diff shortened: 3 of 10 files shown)


src/amo/components/Link.js
```diff
@@ -19,7 +19,7 @@ export class LinkBase extends React.Component {
     prependClientApp: PropTypes.bool,
     prependLang: PropTypes.bool,
     to: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
-    dispatch: PropTypes.func.isRequired,
+    dispatch: PropTypes.func,
```
kumar303: What is dispatch used for in this component? I guess it's in the destructor only to make sure it gets excluded from `customProps`?

src/ui/components/DropdownMenu/index.js
```diff
@@ -11,7 +11,8 @@ import './styles.scss';
 
 type Props = {|
   text: string,
-  children: Array<DropdownMenuItem>,
+  // TODO: Replace Array by React.ChildrenArray when we upgrade Flow to 0.53+.
```
kumar303: Ah, yeah, I saw your comment about that. I'll try to look at that upgrade soon. We have some things to fix.

tests/unit/amo/components/TestHeader.js
```diff
@@ -77,17 +76,13 @@ describe(__filename, () => {
...
-      .expects('logOutFromServer')
-      .once()
-      .returns(Promise.resolve());
+    const wrapper = renderHeader({ store, handleLogOut });
 
     const onClick = wrapper.find(DropdownMenuItem).last().prop('onClick');
     onClick(createFakeEvent());
 
-    mockApi.verify();
+    sinon.assert.called(handleLogOut);
```
kumar303: This needs to also make sure `handleLogOut` was called with `{ api }` since that's a crucial part of its signature. This is the type of thing Flow can be really good for. It would tell you if you're calling `handleLogOut()` incorrectly even though it's a stub.

tests/unit/core/components/TestAuthenticateButton.js
```diff
@@ -132,4 +109,33 @@ describe('<AuthenticateButton />', () => {
...
+      const handleLogOut = createHandleLogOutFunction(store.dispatch);
+      const apiState = state.api;
+      // This makes sure the state contains a token because user is logged in.
+      expect(apiState.token).toBeTruthy();
+
+      return handleLogOut({ api: apiState })
+        .then(() => {
+          sinon.assert.calledWith(api.logOutFromServer, { api: apiState });
+          // This makes sure the logOutUser() action has been dispatched.
+          expect(store.getState().api.token).toBeFalsy();
```
kumar303: Why is this needed? Checking that the API was called correctly is enough, no?

src/amo/components/Link.js
willdurand: Yes, this is needed here because I want to exclude it from the list passed to `a`.

tests/unit/core/components/TestAuthenticateButton.js
willdurand: That was part of the previous test I merged and the spec reflected it so I thought that was cool to keep.

willdurand: Ah, edit: this is also because once `logOutFromServer()` has been called, we dispatch the `logOutUser()` action, and the result is that the token should be null'ed.

tests/unit/amo/components/TestHeader.js
willdurand: ah sure!

src/amo/components/Link.js
kumar303: How does `Link` get the dispatch property to begin with? I don't see a mapper for it.

willdurand: I think it is because of the spread everywhere 😞

tests/unit/core/components/TestAuthenticateButton.js
kumar303: Okay, I understand. It's not necessary to test that the state gets reduced because that's already covered in https://github.com/mozilla/addons-frontend/blob/071db107cb4fdac16a28071dfa60400164ed501a/tests/unit/core/reducers/test_api.js#L23

In other words, this code should rely on the interface (which is the `logOutFromServer` action), not its implementation.

You can do this by asserting that the action is dispatched:

```js
const dispatchSpy = sinon.spy(store, 'dispatch');
...
sinon.assert.calledWith(dispatchSpy, logOutFromServer());
```

src/amo/components/Link.js
willdurand: Going to remove actually. It does not fix the real issue.

willdurand:
I have updated the tests!

tests/unit/amo/components/TestHeader.js
```diff
@@ -77,17 +76,15 @@ describe(__filename, () => {
...
-    const mockApi = sinon.mock(api);
+    const handleLogOut = sinon.stub();
 
-    mockApi
-      .expects('logOutFromServer')
-      .once()
-      .returns(Promise.resolve());
+    const wrapper = renderHeader({ store, handleLogOut });
 
     const onClick = wrapper.find(DropdownMenuItem).last().prop('onClick');
```
kumar303: Using `.last()` seems a bit fragile. You could give it a class like `Header-logOut` and find that instead.

tests/unit/core/components/TestAuthenticateButton.js
```diff
@@ -132,4 +111,44 @@ describe('<AuthenticateButton />', () => {
...
+  });
+
+  describe('createHandleLogOutFunction()', () => {
+    it('logs out a signed-in user on the server and client sides', () => {
+      sinon.stub(api, 'logOutFromServer').returns(Promise.resolve());
+
+      const { store, state } = dispatchSignInActions();
+      const dispatchSpy = sinon.spy(store, 'dispatch');
+
+      const handleLogOut = createHandleLogOutFunction(dispatchSpy);
```
kumar303: Heh, I had to looks this up in the docs. I was going to suggest passing `store.dispatch` since the intention is a bit more clear but they are identical so it's fine either way.

muffinresearch:
FWIW: The earlier test fail was an intermittent failure on just the uitests - cleared by re-starting the job on CI.

PR #3126, opened 2017-09-13
Add add-on DB ID to MoreInfoCard

### LINKED ISSUE #2750: Reveal add-on ID by clicking/hovering on icon
According to @jvillalobos, this exists on the current site (though I couldn't get it to work–maybe it's a staff-only feature?) and is useful for debugging.

I'm guessing it's the database ID that should be shown here? @jvillalobos?

### LINKED ISSUE #3127: Show LoadingText in AddonMoreInfo Card on details page
Right now during loading states the "More information" card is hidden, but it *should* render with LoadingText content. I was adding something (the DB ID, #2750) to this card and realised it was an easy addition so I folded it into #3126, filing here for posterity/QA.

tofumatt:
This PR looks huge because I moved the tests to enzyme (and thus had to change from using `ref` because `shallow` won't find refs. I like putting classNames on stuff anyway.

* Fixes #2750
* Fixes #3127 

Adds the add-on's database ID to the more info card. Useful for admins/debugging/etc.

Check out 62b85c0a607a35f0459425e0c51b47acdb1edb1f for the meat of the change.

### Before
screenshot 2017-09-13 22 28 59

### After
screenshot 2017-09-13 22 25 57

------

While I was here it was super-easy to add LoadingText support to this component so I did:

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -87,9 +87,7 @@ export class AddonMoreInfoBase extends React.Component {
             </dd>
           ) : null}
           {addon.has_privacy_policy ? (
-            <dt
-              className="AddonMoreInfo-privacy-policy-title"
-            >
+            <dt className="AddonMoreInfo-privacy-policy-title">
               {i18n.gettext('Privacy Policy')}
             </dt>
           ) : null}
@@ -104,9 +102,7 @@ export class AddonMoreInfoBase extends React.Component {
             </dd>
           ) : null}
           {addon.has_eula ? (
-            <dt
-              className="AddonMoreInfo-eula-title"
-            >
+            <dt className="AddonMoreInfo-eula-title">
               {i18n.gettext('End-User License Agreement')}
             </dt>
           ) : null}
@@ -118,6 +114,20 @@ export class AddonMoreInfoBase extends React.Component {
               >
                 {i18n.gettext('Read the license agreement for this add-on')}
               </Link>
+            </dd>
+          ) : null}
+          {addon.id ? (
+            <dt
+              className="AddonMoreInfo-database-id-title"
+              title={i18n.gettext(`This ID is useful for debugging and
+                identifying your add-on to site administrators.`)}
+            >
+              {i18n.gettext('Internal Database ID')}
+            </dt>
+          ) : null}
+          {addon.id ? (
+            <dd className="AddonMoreInfo-database-id-content">
+              {addon.id}
             </dd>
           ) : null}
         </dl>
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -7,7 +7,7 @@ import translate from 'core/i18n/translate';
 import { trimAndAddProtocolToUrl } from 'core/utils';
 import Card from 'ui/components/Card';
 
-import './AddonMoreInfo.scss';
+import './styles.scss';
 
 
 export class AddonMoreInfoBase extends React.Component {
@@ -22,38 +22,42 @@ export class AddonMoreInfoBase extends React.Component {
     let homepage = trimAndAddProtocolToUrl(addon.homepage);
     if (homepage) {
       homepage = (
-        <li><a href={homepage} ref={(ref) => { this.homepageLink = ref; }}>
+        <li><a className="AddonMoreInfo-homepage-link" href={homepage}>
           {i18n.gettext('Homepage')}
         </a></li>
       );
     }
+
     let supportUrl = trimAndAddProtocolToUrl(addon.support_url);
     if (supportUrl) {
       supportUrl = (
-        <li><a href={supportUrl} ref={(ref) => { this.supportLink = ref; }}>
+        <li><a className="AddonMoreInfo-support-link" href={supportUrl}>
           {i18n.gettext('Support Site')}
         </a></li>
       );
     }
 
     return (
-      <Card className="AddonMoreInfo" header={i18n.gettext('More information')}>
+      <Card
+        className="AddonMoreInfo"
+        header={i18n.gettext('More information')}
+      >
         <dl className="AddonMoreInfo-contents">
           {homepage || supportUrl ? (
-            <dt ref={(ref) => { this.linkTitle = ref; }}>
+            <dt className="AddonMoreInfo-links-title">
               {i18n.gettext('Add-on Links')}
             </dt>
           ) : null}
           {homepage || supportUrl ? (
-            <dd className="AddonMoreInfo-contents-links">
-              <ul className="AddonMoreInfo-contents-links-list">
+            <dd className="AddonMoreInfo-links-contents">
+              <ul className="AddonMoreInfo-links-contents-list">
                 {homepage}
                 {supportUrl}
               </ul>
             </dd>
           ) : null}
           <dt>{i18n.gettext('Version')}</dt>
-          <dd ref={(ref) => { this.version = ref; }}>
+          <dd className="AddonMoreInfo-version">
             {addon.current_version.version}
           </dd>
           <dt>{i18n.gettext('Last updated')}</dt>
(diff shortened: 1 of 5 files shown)


src/amo/components/AddonMoreInfo/index.js
```diff
@@ -19,41 +20,61 @@ export class AddonMoreInfoBase extends React.Component {
   render() {
     const { addon, i18n } = this.props;
 
+    if (!addon) {
+      return (
```
kumar303: The header gets repeated here and the loading styles also don't seem to match up too well. What about using a helper like this?

```js
return this.renderCard({
  homepage: <LoadingText width={55} />,
  supportUrl: <LoadingText width={55} />
  version: <LoadingText width={55} />,
  lastUpdated: <LoadingText width={25} />,
  // etc...
});
```

When you have an add-on, you could just fill it in with real values or nulls as necessary:

```js
return this.renderCard({
  homepage,
  supportUrl,
  version: addon.current_version.version,
  lastUpdated: i18n.sprintf(...),
  // etc...
});
```

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -1,34 +1,36 @@
...
-    return findRenderedComponentWithType(renderIntoDocument(
-      <Provider store={store}>
-        <AddonMoreInfoBase
-          i18n={getFakeI18nInst()}
-          addon={fakeAddon}
-          {...props}
-        />
-      </Provider>
-    ), AddonMoreInfoBase);
+    return shallowUntilTarget(
```
kumar303: nice

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -46,87 +49,117 @@ describe('<AddonMoreInfo />', () => {
...
     const root = render({ addon });
 
-    expect(findDOMNode(root.privacyPolicyLink).textContent).toEqual('Read the privacy policy for this add-on');
-    // TODO: Change this to an internal `<Link>` tag and use `expect().toBe`
-    // once https://github.com/mozilla/addons-frontend/issues/1828 is fixed.
-    expect(root.privacyPolicyLink.props.href).toContain('/addon/chill-out/privacy/');
+    expect(root.find('.AddonMoreInfo-privacy-policy-title'))
+      .toHaveText('Privacy Policy');
+    expect(root.find('.AddonMoreInfo-privacy-policy-link'))
+      .toHaveProp('children', 'Read the privacy policy for this add-on');
```
kumar303: Could you use [`.children()`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/children.html) here?

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -46,87 +49,117 @@ describe('<AddonMoreInfo />', () => {
...
+    expect(root.find('.AddonMoreInfo-privacy-policy-link'))
+      .toHaveProp('href', '/addon/chill-out/privacy/');
+  });
+
+  it('does not render a EULA if none exists', () => {
+    const partialAddon = { ...fakeAddon, has_eula: false };
```
kumar303: This was only called `partialAddon` in the other test because one of its properties got deleted.

src/amo/components/AddonMoreInfo/index.js
tofumatt: Yeah, I could make it more DRY.

Also yeah I'll tweak the styles a bit, I just sort of noticed that. Thanks!

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: Oh, I guess so. I forgot about it! 😅

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: True! There's also the fact that a lot of the tests rely on the default props in `fakeAddon` but I like the pattern of explicitly setting the props we're testing so it's visible the relationship. I'll quickly tidy up the tests...

src/amo/components/AddonMoreInfo/index.js
tofumatt: Oh to be clear though I want the `<dt>` *and* `<dd>`s to appear as `<LoadingText />` because which sections appear are pretty varied, so I don't want to render:

```
<dt>Privacy Policy</dt>
<dd><LoadingText /></dd>
```
because it might turn out the add-on doesn't even have a Privacy Policy. So I'm just rendering a few dt/dd sets.

kumar303: ok, in that case you can at least share the header like:

```js
return this.renderCard(
    <dt><LoadingText width={55} /></dt>
    <dd><LoadingText width={25} /></dd>
    <dt><LoadingText width={55} /></dt>
    <dd><LoadingText width={25} /></dd>
  </dl>
);
```

However, I'd think for the purpose of making the loading screen match the final content you'd want to preserve the headers like `<dt>Privacy Policy</dt>`.

tofumatt: Indeed, that's what I ended up with.

Unfortunately we can't do that because the headers that appear are quite variable. Nearly everything is conditionally rendered. I should be able to match the size at least with the right margins.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -0,0 +1,169 @@
...
+            </Link>
+          </dd>
+        ) : null}
+        {addon.id ? (
+          <dt
+            className="AddonMoreInfo-database-id-title"
+            title={i18n.gettext(`This ID is useful for debugging and
+              identifying your add-on to site administrators.`)}
+          >
+            {i18n.gettext('Internal Database ID')}
```
muffinresearch: Not sure if there was other guidance on naming but maybe we could avoid the use of Internal and Database they make it look a little like something that shouldn't be public. Would `Add-on ID` or just simply `ID` or `Identifier`.

tofumatt: "Site Identifier" maybe?

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -0,0 +1,169 @@
...
+              href={`/addon/${addon.slug}/eula/`}
+            >
+              {i18n.gettext('Read the license agreement for this add-on')}
+            </Link>
+          </dd>
+        ) : null}
+        {addon.id ? (
+          <dt
+            className="AddonMoreInfo-database-id-title"
+            title={i18n.gettext(`This ID is useful for debugging and
```
willdurand: No need for `oneLine`?

tofumatt: Very confusingly (I think we mention in the `i18n` docs) our `gettext` processing needs no `oneLine`/other template string modifiers. So in this one case `oneLine` will get applied automatically. 😓

muffinresearch: That's right. If you use a template tag the babel extraction will not be able to extract the string correctly. I would love an eslint rule to catch this case, but it's way down on the list of priorities unfortunately.

src/amo/components/AddonMoreInfo/index.js
muffinresearch: > "Site Identifier" maybe?

Works for me

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -0,0 +1,169 @@
...
+          </dd>
+        ) : null}
+        <dt>{i18n.gettext('Version')}</dt>
+        <dd className="AddonMoreInfo-version">
+          {addon.current_version.version}
+        </dd>
+        <dt>{i18n.gettext('Last updated')}</dt>
+        <dd>
+          {i18n.sprintf(
+            // L10n: This will output, in English:
```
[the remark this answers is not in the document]
tofumatt: Oh, shoot, you're right!

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -0,0 +1,169 @@
...
+            // "2 months ago (Dec 12 2016)"
+            i18n.gettext('%(timeFromNow)s (%(date)s)'), {
```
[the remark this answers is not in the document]
tofumatt: It being parenthetical needs localisation. Some locales might not want to use parens, format it differently, etc.

Moment localises the times but the `i18n` we're doing here is about the structure. It's pretty minor but it matters 😄

PR #3128, opened 2017-09-13
Update GPG key for yarn

### LINKED ISSUE #3129: Disco Pane tests (and yarn install) is failing because of yarn GPG Key change
Yarn updated their public key and so we get this error:

```
GPG error: https://dl.yarnpkg.com stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY E074D16EB6FF4DE3
```

We'll likely need to cherry-pick this into release or the server installs/running will fail in the same way as Disco Pane tests were here: https://travis-ci.org/mozilla/addons-frontend/jobs/275220669

tofumatt:
Should fix the failing Disco Pane errors.

Fixes #3129


docker/etc/pki/yarnpkg.gpg.key
```diff
@@ -1,4 +1,5 @@
 -----BEGIN PGP PUBLIC KEY BLOCK-----
+Version: GnuPG v2
```
kumar303: This makes me think they maybe just upgraded GPG or something

kumar303:
@willdurand may also want to take a look at this

docker/etc/pki/yarnpkg.gpg.key
tofumatt: That's what I thought too. The key is a lot longer but I forget if public keys have consistent lengths because I totally forget how GPG's encryption works 😅

tofumatt:
Isn't he out the rest of the week? Or do you just mean it's good to be aware of?

kumar303:
Yeah, just something to be aware of

tofumatt:
And yeah I'm wary of fetching GPG keys as a build step. Even if this failed during the prod build the build just wouldn't be promoted and we'd investigate; I think that's a lot better than fetching the keys in an automated fashion, even if it takes longer to fix.

kumar303:
Well, if they rotate it every couple weeks then it will get annoying. We'll see. Their [install docs](https://yarnpkg.com/en/docs/install#linux-tab) show fetching it from the URL but maybe it's just documented that way for convenience.

willdurand:
Looks good to me. I hope they do not change the key on every new version..

tofumatt:
This is the first time I've noticed this happen so hopefully it's not common. Hoping it's just caused by a major GPG version change.

kumar303:
> This is the first time I've noticed this happen 

We only started using yarn (correctly) for deploys as of https://github.com/mozilla/addons-frontend/pull/3033 (landed 15 days ago)

tofumatt:
Ahhh, shoot, that’s why then. :-(

PR #3136, opened 2017-09-14
Make the loading indicators more subtle

### LINKED ISSUE #2899: Lighten the colour of the loading indicator bars
The grey used in the loading indicator is quite dark which make the effect a bit jarring. It would be less so if the grey used was knocked back a notch.

We might want to get UX input on an appropriate shade for this.

muffinresearch:
Fixes #2899

This makes use of a photon grey which is lighter than the current grey used as the base colour for the loading indicator animations. I've reduced the lightest opacity so it doesn't get too light. 

Before (ignore the arrow it's not relevant to this)

untitled_2

After (updated to grey-30 after feedback)

untitled_3

### DIFF
--- a/src/ui/components/LoadingText/styles.scss
+++ b/src/ui/components/LoadingText/styles.scss
@@ -1,4 +1,5 @@
 @import "~amo/css/inc/vars";
+@import "~photon-colors/colors";
 
 @keyframes placeHolderShimmer {
   0% {
@@ -6,7 +7,7 @@
   }
 
   25% {
-    opacity: 0.15;
+    opacity: 0.25;
   }
 
   50% {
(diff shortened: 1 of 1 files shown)


muffinresearch:
@pwalm thoughts?

pwalm:
Can we go lighter? How about #F9F9FA?

muffinresearch:
> Can we go lighter? How about #F9F9FA?

That'll be way too light. The thing is the animation works by changing the opacity of the grey used, so it ends up various shades lighter than the base grey as the animation goes through the frames.

If you think lighter is the way to go here's a screenshot of using grey-30 as the base:

untitled_3

Does that look better?

pwalm:
Ship it!

PR #3139, opened 2017-09-14
Add display name to user state and use it in menu

### LINKED ISSUE #3118: Don't set the user menu label to username if there is a display name available
Steps to reproduce:
1.Login to amo-dev , mobile view
2.Be sure that your account have both username and display name filled

Expected results:
Display name should be used as user menu label 

Actual results:
Username is used

NOTES:
If username is deleted from account settings, then the label will be set to "anonymous ... " even if we have a display name 
Reproducible on Android devices too

username

willdurand:
Fix #3118

---

This PR adds the `displayName` to the `user` state and uses it in the menu when possible.

Before (or without a display name):

screen shot 2017-09-14 at 15 50 51

After (I usually don't use my second name but I needed it for testing 😛 ):

screen shot 2017-09-14 at 15 50 38

### DIFF
--- a/src/core/reducers/user.js
+++ b/src/core/reducers/user.js
@@ -8,11 +8,13 @@ type Action = Object;
 export type UserStateType = {
   id: ?number,
   username: ?string,
+  displayName: ?string,
 };
 
 export const initialState: UserStateType = {
   id: null,
   username: null,
+  displayName: null,
 };
 
 export const loadUserProfile = ({ profile }: Object) => {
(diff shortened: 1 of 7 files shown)


src/core/reducers/user.js
```diff
@@ -30,6 +32,17 @@ export const isAuthenticated = (state: { user: UserStateType }) => {
   return !!state.user.id;
 };
 
+export const selectDisplayName = (state: { user: UserStateType }) => {
+  const displayName = state.user.displayName;
+  if (displayName && displayName !== '') {
+    return displayName;
+  }
+
+  // We fallback to the username if no display name has been defined by the
```
tofumatt: Weird, I thought `display_name` would be `null` if the user hadn't picked one. I'd also sort of expect it to fall back to `username`.

I guess there's nothing wrong with this approach but I'm surprised we'd ever get an empty string like that back from the API.

src/core/reducers/user.js
```diff
@@ -30,6 +32,17 @@ export const isAuthenticated = (state: { user: UserStateType }) => {
   return !!state.user.id;
 };
 
+export const selectDisplayName = (state: { user: UserStateType }) => {
+  const displayName = state.user.displayName;
+  if (displayName && displayName !== '') {
```
tofumatt: I always like doing a `displayName.length` check for empty strings instead, but that's just me. I tend to use that idiom in the code but if it's varied do whichever you like 😄

PR #3145, opened 2017-09-14
Remove old HandleLogic component

### LINKED ISSUE #3146: HandleLogin component is not used
It seems the `HandleLogin` component is never called. We should remove it because it is misleading. I thought I could hook into it to update the cookie, but that's not the case. The `login()` API method is not used either so it should also be removed.

We double-checked with ops and they did not find any calls to [the `login` endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/auth_internal.html#get--api-v3-accounts-login-) in the logs (asked on IRC).

willdurand:
Fix #3146

---

This PR removes useless files to avoid confusion.


src/core/containers/HandleLogin/index.js
```diff
@@ -1,94 +0,0 @@
...
-import { connect } from 'react-redux';
-import { compose } from 'redux';
-import config from 'config';
-import { withRouter } from 'react-router';
-
-import { setAuthToken } from 'core/actions';
-import { login } from 'core/api';
-import LoginPage from 'core/components/LoginPage';
-import log from 'core/logger';
-import { browserBase64Decode, gettext as _ } from 'core/utils';
```
kumar303: It looks like you can delete `browserBase64Decode` too.

willdurand: thanks!

kumar303: It didn't have any tests either, heh

PR #3161, opened 2017-09-15
Fix Search header doesn't mention type #3154

### LINKED ISSUE #3154: Search header doesn't mention type
Currently our search header says: `9 results for "foo"` but the mocks and scope doc suggest it should be `9 extension results for "foo"`.

Current site:

add-ons_for_firefox

Mocks:

amo_desktop_-_master_sketch

PunitGr:
Fixes #3154 
------------

This PR fixes the `Search Header mention type issue` which appears while searching a query.

I've updated the `SearchText` string by using `addonType` from `filters` which fixes this bug. Also, changed the `addonType` text to plural if search results are more than 1.

### Screenshots

* Test Case: **All Passed**

screen shot 2017-09-15 at 10 18 47 pm

* Search Preview after the fix:

screen shot 2017-09-15 at 10 19 35 pm

screen shot 2017-09-15 at 10 41 12 pm

* After updating test cases

screen shot 2017-09-16 at 3 02 46 am

Thanks!

### DIFF
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -60,6 +60,12 @@ export class SearchContextCardBase extends React.Component {
           '%(count)s themes found for "%(query)s"',
           count), { count: i18n.formatNumber(count), query }
         );
+      } else {
+        searchText = i18n.sprintf(i18n.ngettext(
+          '%(count)s result for "%(query)s"',
+          '%(count)s results for "%(query)s"',
+          count), { count: i18n.formatNumber(count), query }
+        );
       }
     } else if (!loading && query) {
       searchText = i18n.sprintf(i18n.ngettext(
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -5,6 +5,12 @@ import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
 import Card from 'ui/components/Card';
+import {
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_DICT,
+  ADDON_TYPE_LANG,
+  ADDON_TYPE_THEME,
+} from 'core/constants';
 
 import './styles.scss';
 
@@ -28,15 +34,24 @@ export class SearchContextCardBase extends React.Component {
     const { addonType } = filters;
 
     let searchText;
+    let addOn;
+
+    if (addonType === ADDON_TYPE_EXTENSION) {
+      addOn = i18n.ngettext('extension', 'extensions', count);
+    } else if (addonType === ADDON_TYPE_DICT) {
+      addOn = i18n.ngettext('dictionary', 'dictionaries', count);
+    } else if (addonType === ADDON_TYPE_LANG) {
+      addOn = i18n.ngettext('language pack', 'language packs', count);
+    } else if (addonType === ADDON_TYPE_THEME) {
+      addOn = i18n.ngettext('theme', 'themes', count);
+    }
 
     if (!loading && query && addonType) {
-      const addon = addonType === 'persona' ? 'theme' : addonType;
-      searchText = i18n.sprintf(i18n.ngettext(
-        '%(count)s %(addon)s found for "%(query)s"',
-        '%(count)s %(addon)ss found for "%(query)s"',
-        count), { count: i18n.formatNumber(count), addon, query }
+      searchText = i18n.sprintf(i18n.gettext(
+        '%(count)s %(addOn)s found for "%(query)s"'
+      ), { count: i18n.formatNumber(count), addOn, query }
       );
-    } else if (!loading && query) {
+    } else if (!loading && query && !addonType) {
       searchText = i18n.sprintf(i18n.ngettext(
         '%(count)s result for "%(query)s"',
         '%(count)s results for "%(query)s"',
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -34,24 +34,34 @@ export class SearchContextCardBase extends React.Component {
     const { addonType } = filters;
 
     let searchText;
-    let addOn;
-
-    if (addonType === ADDON_TYPE_EXTENSION) {
-      addOn = i18n.ngettext('extension', 'extensions', count);
-    } else if (addonType === ADDON_TYPE_DICT) {
-      addOn = i18n.ngettext('dictionary', 'dictionaries', count);
-    } else if (addonType === ADDON_TYPE_LANG) {
-      addOn = i18n.ngettext('language pack', 'language packs', count);
-    } else if (addonType === ADDON_TYPE_THEME) {
-      addOn = i18n.ngettext('theme', 'themes', count);
-    }
 
     if (!loading && query && addonType) {
-      searchText = i18n.sprintf(i18n.gettext(
-        '%(count)s %(addOn)s found for "%(query)s"'
-      ), { count: i18n.formatNumber(count), addOn, query }
-      );
-    } else if (!loading && query && !addonType) {
+      if (addonType === ADDON_TYPE_EXTENSION) {
+        searchText = i18n.sprintf(i18n.ngettext(
+          '%(count)s extension found for "%(query)s"',
+          '%(count)s extensions found for "%(query)s"',
+          count), { count: i18n.formatNumber(count), query }
+        );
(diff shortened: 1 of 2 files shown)


src/amo/components/SearchContextCard/index.js
```diff
@@ -25,10 +25,18 @@ export class SearchContextCardBase extends React.Component {
   render() {
     const { count, filters, i18n, loading } = this.props;
     const { query } = filters;
+    const { addonType } = filters;
 
     let searchText;
 
-    if (!loading && query) {
+    if (!loading && query && addonType) {
+      const addon = addonType === 'persona' ? 'theme' : addonType;
```
tofumatt: Unfortunately `addonType` is not localised, so this won't work in any language that isn't English (and even then, `addonType` isn't very readible if it's something like `langpack`).

So this will need to be more of an `if`/`else` case like we have here: https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/SearchForm/index.js#L207

Sorry about that!

PunitGr: Alright! I don't know why I didn't think of it. Sorry, my bad!

tofumatt: No worries, we've made the same mistake too! 😉

src/amo/components/SearchContextCard/index.js
```diff
@@ -25,10 +31,27 @@ export class SearchContextCardBase extends React.Component {
...
     const { count, filters, i18n, loading } = this.props;
     const { query } = filters;
+    const { addonType } = filters;
 
     let searchText;
+    let addOn;
 
-    if (!loading && query) {
+    if (addonType === ADDON_TYPE_EXTENSION) {
+      addOn = i18n.ngettext('extension', 'extensions', count);
```
tofumatt: Oh, this is clever but unfortunately it's best to put the entire string in here instead. Some locales have complex pluralisation rules and it's best to have the entire string in context. I know it's a bit less elegant that way, but if you look at the example I referenced earlier that's how we do it. Could you put the entire string (in both singular and plural forms) in one gettext call?

src/amo/components/SearchContextCard/index.js
```diff
@@ -25,10 +31,27 @@ export class SearchContextCardBase extends React.Component {
...
+    } else if (addonType === ADDON_TYPE_THEME) {
+      addOn = i18n.ngettext('theme', 'themes', count);
+    }
+
+    if (!loading && query && addonType) {
+      searchText = i18n.sprintf(i18n.gettext(
+        '%(count)s %(addOn)s found for "%(query)s"'
+      ), { count: i18n.formatNumber(count), addOn, query }
+      );
+    } else if (!loading && query && !addonType) {
```
tofumatt: I think the `!addonType` is redundant; because you test for `addonType` above and this is otherwise the same check you can just remove that check here 😄

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -104,4 +109,125 @@ describe('SearchContextCard', () => {
...
+      },
+    });
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header'))
+      .toIncludeText('2 extensions found for "test"');
+  });
+
+  // addonType = ADDON_TYPE_EXTENSION
```
tofumatt: You don't need these comments here, you wrote really good test specs actually (👍🏻) and the `dispatchSearchResults` calls show which type you're using so you can remove these comments.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -104,4 +109,125 @@ describe('SearchContextCard', () => {
...
+      },
+    });
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header'))
+      .toIncludeText('1 dictionary found for "test"');
+  });
+
+  it('should render plural form when multipe results are found with addonType as ADDON_TYPE_DICT', () => {
```
tofumatt: Typo here: multipe should be multiple. 😄

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -104,4 +109,125 @@ describe('SearchContextCard', () => {
...
+    });
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header'))
+      .toIncludeText('2 themes found for "test"');
+  });
+
+  // addonType = ADDON_TYPE_DICT
+  it('should render singular form when only one result is found with addonType as ADDON_TYPE_DICT', () => {
```
tofumatt: I think you can just end with `ADDON_TYPE_DICT` instead of `addonType as ADDON_TYPE_DICT`. Similar for the rest of the test specs.

They're otherwise really good though, thanks for writing really clear test specs! 👍

src/amo/components/SearchContextCard/index.js
PunitGr: Yes sure, thanks for letting me know regarding this. Will keep this in mind in future.

PunitGr:
@tofumatt any more changes would you recommend?

PunitGr:
@tofumatt , I have updated the code. Please let me know if you recommend any more changes. Thanks.

src/amo/components/SearchContextCard/index.js
```diff
@@ -25,10 +31,37 @@ export class SearchContextCardBase extends React.Component {
...
+          '%(count)s language packs found for "%(query)s"',
+          count), { count: i18n.formatNumber(count), query }
+        );
+      } else if (addonType === ADDON_TYPE_THEME) {
+        searchText = i18n.sprintf(i18n.ngettext(
+          '%(count)s theme found for "%(query)s"',
+          '%(count)s themes found for "%(query)s"',
+          count), { count: i18n.formatNumber(count), query }
+        );
+      }
```
kumar303: This could use an `else` block that shows some kind of generic message like '12 results found for...'

PunitGr:
@kumar303, I was gonna do that before but couldn't find a test case regarding the same when there is an `addonType`. I've updated it.
Thanks

kumar303:
Thanks. The `else` branch now needs test coverage (unless I'm missing it).

PunitGr:
I don't think the `else` statement works. Since it doesn't get any response from the `server api`. And if we don't get any `response` no result is rendered on the `frontend`. Don't you think there should be a response if the `addonType` someone mentioned doesn't exist in the database?

kumar303:
It's an edge case but it's still possible. You can get trigger a "valid" search with this archaic type (if you edit the URL):

http://localhost:3000/en-US/firefox/search/?platform=mac&q=firefox&type=theme

This `theme` type has been replaced by `persona` so we shouldn't use it but it should still render a heading in this case. I'm more worried about when we add a new type and maybe forget to update the code. We could either throw an error or display a generic message -- I like displaying a generic message better.

PunitGr:
It works for the cases if `addonType` exists in the DB, even for `themes`. But the problem occurs when there is no searched `addonType` like for [http://localhost:3000/en-US/firefox/search/?platform=mac&q=firefox&type=theme](http://localhost:3000/en-US/firefox/search/?platform=mac&q=firefox&type=xyz). 
Since there is no `record` for `xyz` in the DB.  The search gets stuck and no response is rendered.

PunitGr:
@kumar303, Take this for an example,
screen shot 2017-09-19 at 11 58 08 pm

kumar303:
> http://localhost:3000/en-US/firefox/search/?platform=mac&q=firefox&type=xyz.
> Since there is no record for xyz in the DB. The search gets stuck and no response is rendered.

That looks like a separate bug. It's not so important since no user would see it while using the site.

PunitGr:
@kumar303 , Okay.
I have added the test cases for `else` block.
Thanks!

caitmuenster:
Thanks so much for the patch, @PunitGr! Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#September_2017). 

If you'd like to set up a profile on [mozillians.org](https://mozillians.org/en-US/), I would be happy to vouch for you.

Welcome onboard! I look forward to seeing you around. :)

PunitGr:
Hi @caitmuenster Thank you so much for the invitation. I've set up my profile. I would be grateful if you can vouch for my profile here [https://mozillians.org/en-US/u/PunitGr/](https://mozillians.org/en-US/u/PunitGr/)

caitmuenster:
\o/ You're all vouched!

PR #3163, opened 2017-09-15
fix: Make screenshots big

### LINKED ISSUE #2571: Screenshots from new desktop are too small
Steps to reproduce:
1. Go to any add-on details pages and press "View Mobile Site" from the bottom of the page (https://addons-dev.allizom.org/en-US/firefox/addon/findbar-tweak/)
2. Open screenshots to full size in order to view them

Expected results:
Screenshots should be at least same size like on the classic desktop site

Actual results:
Screenshots are too small, and the details can be hardly distinguished

screenshotsize

tofumatt:
Fix #2571.

Well, this wasn't hard.

The `Screenshots` component we're using frustratingly *requires* height and width params so I couldn't just leave this blank or set them to `100%`, but this works well enough as it's as big as the max size of the size and are basically as big as the screenshots come down the wire from AMO anyway, for now.

### DIFF
--- a/tests/unit/amo/components/TestScreenShots.js
+++ b/tests/unit/amo/components/TestScreenShots.js
@@ -25,15 +25,15 @@ describe('<ScreenShots />', () => {
         title: 'A screenshot',
         src: 'http://img.com/one',
         thumbnail_src: 'http://img.com/1',
-        h: 200,
-        w: 320,
+        h: 1024,
+        w: 1366,
       },
       {
         title: 'Another screenshot',
         src: 'http://img.com/two',
         thumbnail_src: 'http://img.com/2',
-        h: 200,
-        w: 320,
+        h: 1024,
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestScreenShots.js
```diff
@@ -25,15 +25,15 @@ describe('<ScreenShots />', () => {
         title: 'A screenshot',
         src: 'http://img.com/one',
         thumbnail_src: 'http://img.com/1',
-        h: 200,
-        w: 320,
+        h: 1024,
+        w: 1366,
```
willdurand: Would it make sense to export the variables in the component and reuse them here?

tofumatt: Not a bad idea! Thanks! 😄

PR #3164, opened 2017-09-16
add in line under themes

### LINKED ISSUE #2197: Disco Pane - light colored or translucent themes problematic in display
#### Proposed fix:

Ok, let's add a `1px stroke #F5F5F5` in between the theme image and the info block below:
screen shot 2017-04-03 at 11 21 21 am

----

#### Original issue

Because of the frame-less way themes are displayed in Disco Pane, light-colored, translucent, or border-less themes are avoided because they appear to just float in space. 

For example, consider the awkwardness of a theme like this in Disco:

https://addons.mozilla.org/firefox/addon/cat-glasses/

... should we consider addressing in some way? Apply some type of border? 

@pwalm, @designakt

andymckay:
Fixes #2197

screenshot 2017-09-15 16 56 03
screenshot 2017-09-15 16 55 43

It's subtle.

### DIFF
--- a/src/disco/css/Addon.scss
+++ b/src/disco/css/Addon.scss
@@ -37,6 +37,7 @@ $addon-padding: 20px;
   }
 
   .theme-image {
+    border-bottom: 2px solid #f5f5f5;
     border-top-left-radius: 3px;
     border-top-right-radius: 3px;
     display: block;


tofumatt:
The mocks make it look more like a 2px line; yours is indeed more subtle. Next to those thick border/shadows around the content it seems *really* faint. How does it look with a `2px` border?

tofumatt:
screenshot 2017-09-16 01 14 56
Don't mean to armchair design but I think using a 2px border is a bit nicer given the colour is so faint.

andymckay:
I'm easy, I would actually make it 1px and slightly darker. But hey, made it 2px.

tofumatt:
Side note, the branch name `disco-line` made me smile.

3a166d6c4da51e9ab74c22ba71ceff8b

Issue #3148, opened 2017-09-14
Read more link doesn't show more content for release notes


muffinresearch:
## STR:

* Go to https://addons-dev.allizom.org/en-US/firefox/addon/requestpolicy/
* Click "Read more" on the release notes:

### What happens:

The read more link disappears but there's no additional content

### What should happen:

A "Read more" link should not be required if the content is already visible.

Before clicking "read more":

<img width="511" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30445836-44f1c750-997f-11e7-84d9-d44ebde1a542.png">

After clicking "read more":

<img width="532" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30445882-653850ec-997f-11e7-9e06-84950b234a89.png">

tsl143:
This is probably because `.ShowMoreCard-contents` have `max-height: 150px` and we test `truncateToMaxHeight` with `MAX_HEIGHT = 100`

ValentinaPC:
Verified as fixed on AMO-dev FF55(Win 7).
Postfix screenshot:
![rn](https://user-images.githubusercontent.com/15685960/30592215-931b5a18-9d4e-11e7-9349-7166e88f5e43.png)

Issue #3045, opened 2017-08-31
Restrict line length to 80 characters (adjust eslint rule)


kumar303:
Code lines restricted to 80 characters lets you use many horizontal split screens in your editor which is nice.

```diff
diff --git a/.eslintrc b/.eslintrc
index c9d84498..6d8085f9 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,13 @@
     "jest"
   ],
   "rules": {
+    "max-len": ["error", 80, 2, {
+      ignoreUrls: true,
+      ignoreComments: false,
+      ignoreRegExpLiterals: true,
+      ignoreStrings: false,
+      ignoreTemplateLiterals: false,
+    }],
     // FIXME: Deprecate use of findDOMNode https://github.com/mozilla/addons-frontend/issues/968
     "react/no-find-dom-node": "off",
   },
```

We discussed it here: https://github.com/mozilla/addons-frontend/pull/3036#discussion_r136206566

Unfortunately, `yarn eslint -- --fix` cannot help with this. 

**Please ask before fixing this issue as it will be hard to review and could introduce merge conflicts**.

kumar303:
As a stepping stone we could maybe do `ignoreStrings: true` and `ignoreTemplateLiterals: true`.

willdurand:
I think I mentioned it already, but maybe [Prettier](https://github.com/prettier/prettier) could help here. At least, it would avoid introducing bugs.

muffinresearch:
@kumar303 this could be something worth considering prettier for as suggested by @willdurand.

muffinresearch:
@willdurand snap! - I've played with it I think it's pretty good it really does format the code like we mostly have it. The only real nit is the requirement of a post-save / pre-commit hook.

muffinresearch:
> The only real nit is the requirement of a post-save / pre-commit hook.

As long as we can either fail when files don't follow it / or have a bot add a commit having run it, it would probably not get in the way.

muffinresearch:
FWIW: I started a patch to do 80chars and it was epic. On that basis alone, switching to use prettier would probably be quite a lot easier to implement.

willdurand:
Well, we can maybe run prettier time to time? Given the intensity on code reviews, I dont think things will diverge too much from our standards.

kumar303:
[...]

PR #3166, opened 2017-09-17
fix: Read more link doesn't show more content for release notes

tsl143:
Fixes: #3148 

**Before:** 

screen shot 2017-09-17 at 4 45 36 pm

**After:**

screen shot 2017-09-17 at 4 44 37 pm

### DIFF
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -13,7 +13,7 @@ import './ShowMoreCard.scss';
 
 // This refers to height of `Card-contents` CSS class,
 // beyond which it will add read more link.
-export const MAX_HEIGHT = 145;
+export const MAX_HEIGHT = 150;
 
 export class ShowMoreCardBase extends React.Component {
   static propTypes = {
@@ -40,7 +40,7 @@ export class ShowMoreCardBase extends React.Component {
   truncateToMaxHeight = (contents) => {
     // If the contents are short enough they don't need a "show more" link; the
     // contents are expanded by default.
-    if (contents.clientHeight > MAX_HEIGHT) {
+    if (contents.clientHeight >= MAX_HEIGHT) {
       this.setState({ expanded: false });
     } else {
       this.setState({ expanded: true });
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -12,7 +12,7 @@ import Card from 'ui/components/Card';
 import './ShowMoreCard.scss';
 
 
-const MAX_HEIGHT = 100;
+const MAX_HEIGHT = 145;
 
 export class ShowMoreCardBase extends React.Component {
   static propTypes = {
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -11,7 +11,8 @@ import Card from 'ui/components/Card';
 
 import './ShowMoreCard.scss';
 
-// This refers to height of Card-contents class, beyond which it will add read more link
+// This refers to height of `Card-contents` CSS class,
+// beyond which it will add read more link.
 export const MAX_HEIGHT = 145;
 
 export class ShowMoreCardBase extends React.Component {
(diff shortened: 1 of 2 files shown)


tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -40,7 +40,7 @@ describe(__filename, () => {
 
   it('truncates the contents if they are too long', () => {
     const root = render({ children: 'Hello I am description' });
-    root.truncateToMaxHeight({ clientHeight: 101 });
+    root.truncateToMaxHeight({ clientHeight: 146 });
```
willdurand: Could you export `MAX_HEIGHT` in the `ShowMoreCard` component and use it here? Like:

```
root.truncateToMaxHeight({ clientHeight: MAX_HEIGHT + 1 });
```

src/ui/components/ShowMoreCard/index.js
```diff
@@ -12,7 +12,7 @@ import Card from 'ui/components/Card';
 import './ShowMoreCard.scss';
 
 
-const MAX_HEIGHT = 100;
+const MAX_HEIGHT = 145;
```
willdurand: We should maybe add a comment saying that it is related to the CSS rule (if I remember correctly).

tests/unit/ui/components/TestShowMoreCard.js
tsl143: This make things easy :)

src/ui/components/ShowMoreCard/index.js
```diff
@@ -11,8 +11,8 @@ import Card from 'ui/components/Card';
 
 import './ShowMoreCard.scss';
 
-
-const MAX_HEIGHT = 100;
+// This refers to height of Card-contents class, beyond which it will add read more link
```
willdurand: I would write:

    `.Card-contents` CSS class

A period is missing at the end of the sentence, which should be wrapped to 80 chars too.

tsl143:
@wagnerand , updated the comment.
Also if we can put 80 char/line for comment limit in lint rules, that will help.

src/ui/components/ShowMoreCard/index.js
```diff
@@ -11,8 +11,9 @@ import Card from 'ui/components/Card';
 
 import './ShowMoreCard.scss';
 
-
-const MAX_HEIGHT = 100;
+// This refers to height of `Card-contents` CSS class,
+// beyond which it will add read more link.
+export const MAX_HEIGHT = 145;
```
willdurand: Why not having `150` actually? I cannot get the issue mentioned in #3148 fixed with this value.

src/ui/components/ShowMoreCard/index.js
```diff
@@ -40,7 +40,7 @@ export class ShowMoreCardBase extends React.Component {
   truncateToMaxHeight = (contents) => {
     // If the contents are short enough they don't need a "show more" link; the
     // contents are expanded by default.
-    if (contents.clientHeight > MAX_HEIGHT) {
+    if (contents.clientHeight >= MAX_HEIGHT) {
```
tsl143: Need this as, when content loads the due to max height the height of content equals 150px, but again on `componentWillReceiveProps`, `truncateToMaxHeight` is invoked so even at 150px we need to set `{ expanded: false }`

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -40,7 +40,7 @@ describe(__filename, () => {
 
   it('truncates the contents if they are too long', () => {
     const root = render({ children: 'Hello I am description' });
-    root.truncateToMaxHeight({ clientHeight: 101 });
+    root.truncateToMaxHeight({ clientHeight: MAX_HEIGHT + 1 });
```
muffinresearch: What is the extra 1px needed for?

tsl143: @muffinresearch , to test `(contents.clientHeight >= MAX_HEIGHT)` this condition.
Initially we were doing it manually `101 px` but then @willdurand suggested to export the variable.

kumar303:
> Also if we can put 80 char/line for comment limit in lint rules, that will help.

It's going to be a big change. I don't think we have bandwidth to do it right now. https://github.com/mozilla/addons-frontend/issues/3045

Issue #3147, opened 2017-09-14
Broken 404, for all reviews.


muffinresearch:
Go to https://addons-dev.allizom.org/en-US/firefox/addon/languagetool/reviews/

Shows content with unresolved loading indicators:

<img width="1401" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30443737-7bb05ba0-9978-11e7-8a33-f7734e16a80f.png">

muffinresearch:
This is possibly a dupe of one of the other reviews pages bugs.

PR #3167, opened 2017-09-17
fix: Broken 404, for all reviews

tsl143:
Fixes: #3147 
I have also considered `401` error , referring from #3061.
**Before:**
screen shot 2017-09-17 at 5 48 00 pm

**After:**
screen shot 2017-09-17 at 5 47 01 pm

### DIFF
--- a/src/amo/components/AddonReviewList.js
+++ b/src/amo/components/AddonReviewList.js
@@ -16,6 +16,7 @@ import { getAddonIconUrl } from 'core/imageUtils';
 import log from 'core/logger';
 import Link from 'amo/components/Link';
 import CardList from 'ui/components/CardList';
+import NotFound from 'amo/components/ErrorPage/NotFound';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import type { UserReviewType } from 'amo/actions/reviews';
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonReviewList.js
```diff
@@ -141,6 +142,18 @@ export class AddonReviewListBase extends React.Component {
     const {
       addon, errorHandler, location, params, i18n, reviewCount, reviews,
     } = this.props;
+
+    if (errorHandler.hasError()) {
+      log.warn('Captured API Error:', errorHandler.capturedError);
+      // A 401 is made to look like a 404 on purpose.
+      // See https://github.com/mozilla/addons-frontend/issues/3061
+      if (errorHandler.capturedError.responseStatusCode === 401 ||
```
tofumatt: That's for an add-on, but will requesting reviews for an add-on that would 401 also send a 401 unauthorised? I guess it's likely, maybe @diox knows off the top of his head.

I guess we should also 404 here then.

tsl143: Even I had that confusion, that's why I put both check :)

PR #3171, opened 2017-09-18
feat: Add Report abuse API/sagas

### LINKED ISSUE #3170: Create Report abuse sagas/API
We need an API/reducer/saga for reporting an add-on for abuse. This is part of #2759, but I want to split up the api/reducer/saga work and the UI work to make code easier to review.

tofumatt:
Fixes #3170.

This is just the API/reducer/saga code needed to allow us to report add-ons for abuse. After this is merged I'll send the PR with the actual UI.

### DIFF
--- a/src/core/api/abuse.js
+++ b/src/core/api/abuse.js
@@ -4,20 +4,20 @@ import type { ApiStateType } from 'core/reducers/api';
 
 
 export type ReportAddonParams = {|
-  addon: string, // id/guid/slug
+  addonSlug: string,
   api: ApiStateType,
-  auth: boolean,
+  auth?: boolean,
   message: string,
 |};
 
 export function reportAddon(
-  { addon, api, auth = false, message }: ReportAddonParams
+  { addonSlug, api, auth = false, message }: ReportAddonParams
 ) {
   return callApi({
     auth,
     endpoint: 'abuse/report/addon',
     method: 'POST',
-    params: { addon, message },
+    params: { addon: addonSlug, message },
     state: api,
   });
 }
--- a/src/core/api/abuse.js
+++ b/src/core/api/abuse.js
@@ -1,0 +1,23 @@
+/* @flow */
+import { callApi } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+
+
+export type ReportAddonParams = {|
+  addon: string, // id/guid/slug
+  api: ApiStateType,
+  auth: boolean,
+  message: string,
+|};
+
+export function reportAddon(
+  { addon, api, auth = false, message }: ReportAddonParams
+) {
+  return callApi({
+    auth,
+    endpoint: 'abuse/report/addon',
+    method: 'POST',
+    params: { addon, message },
+    state: api,
+  });
+}
--- a/src/core/api/abuse.js
+++ b/src/core/api/abuse.js
@@ -6,15 +6,14 @@ import type { ApiStateType } from 'core/reducers/api';
 export type ReportAddonParams = {|
   addonSlug: string,
   api: ApiStateType,
-  auth?: boolean,
   message: string,
 |};
 
 export function reportAddon(
-  { addonSlug, api, auth = false, message }: ReportAddonParams
+  { addonSlug, api, message }: ReportAddonParams
 ) {
   return callApi({
-    auth,
+    auth: true,
     endpoint: 'abuse/report/addon',
     method: 'POST',
     params: { addon: addonSlug, message },
--- a/src/core/reducers/abuse.js
+++ b/src/core/reducers/abuse.js
@@ -62,7 +62,7 @@ export const initialState = {
 
 type ReducerState = {|
   bySlug: {
-    [addonSlug: string]: Array<{| message: string, reporter: string |}>,
+    [addonSlug: string]: {| message: string, reporter: string |},
   },
 |};
 
@@ -71,14 +71,6 @@ export default function abuseReducer(
   action: Object
 ) {
   switch (action.type) {
-    case SEND_ADDON_ABUSE_REPORT: {
-      const { addonSlug } = action.payload;
-      if (state.bySlug[addonSlug]) {
-        throw new Error('You already reported this add-on.');
-      }
-
-      return state;
-    }
     case LOAD_ADDON_ABUSE_REPORT: {
       const { addon, message, reporter } = action.payload;
       return {
--- a/src/core/reducers/abuse.js
+++ b/src/core/reducers/abuse.js
@@ -1,0 +1,81 @@
+/* @flow */
+import type { ErrorHandlerType } from 'core/errorHandler';
+
+
+export const LOAD_ADDON_ABUSE_REPORT = 'LOAD_ADDON_ABUSE_REPORT';
+export const SEND_ADDON_ABUSE_REPORT = 'SEND_ADDON_ABUSE_REPORT';
+
+type LoadAddonAbuseReportType = {
+  addon: {|
+    guid: string,
+    id: number,
+    slug: string,
+  |},
+  message: string,
+  reporter: Object | null,
+};
+
+export function loadAddonAbuseReport(
+  { addon, message, reporter }: LoadAddonAbuseReportType
+) {
+  if (!addon) {
+    throw new Error('addon is required');
+  }
+  if (!message) {
+    throw new Error('message is required');
+  }
+  if (reporter === undefined) {
+    throw new Error('reporter cannot be undefined');
+  }
+
+  return {
+    type: LOAD_ADDON_ABUSE_REPORT,
+    payload: { addon, message, reporter },
+  };
+}
+
+type SendAddonAbuseReportAction = {|
+  addonSlug: string,
+  errorHandler: ErrorHandlerType,
+  message: string,
+|};
+
+export function sendAddonAbuseReport(
+  { addonSlug, errorHandler, message }: SendAddonAbuseReportAction
+) {
+  if (!addonSlug) {
+    throw new Error('addonSlug is required');
+  }
+  if (!errorHandler) {
+    throw new Error('errorHandler is required');
+  }
+  if (!message) {
+    throw new Error('message is required');
+  }
+
+  return {
+    type: SEND_ADDON_ABUSE_REPORT,
+    payload: { addonSlug, errorHandlerId: errorHandler.id, message },
+  };
+}
+
+export const initialState = {};
+
+type ReducerState = {};
+
+export default function abuseReducer(
+  state: ReducerState = initialState,
+  action: Object
+) {
+  switch (action.type) {
+    case LOAD_ADDON_ABUSE_REPORT: {
+      const { addon, message, reporter } = action.payload;
+      return {
+        ...state,
+        [addon.slug]: { message, reporter },
+      };
+    }
+    default:
+      return state;
+  }
+}
--- a/src/core/reducers/abuse.js
+++ b/src/core/reducers/abuse.js
@@ -36,18 +36,18 @@ export function loadAddonAbuseReport(
 
 type SendAddonAbuseReportAction = {|
   addonSlug: string,
-  errorHandler: ErrorHandlerType,
+  errorHandlerId: string,
   message: string,
 |};
 
 export function sendAddonAbuseReport(
-  { addonSlug, errorHandler, message }: SendAddonAbuseReportAction
+  { addonSlug, errorHandlerId, message }: SendAddonAbuseReportAction
 ) {
   if (!addonSlug) {
     throw new Error('addonSlug is required');
   }
-  if (!errorHandler) {
-    throw new Error('errorHandler is required');
+  if (!errorHandlerId) {
+    throw new Error('errorHandlerId is required');
   }
   if (!message) {
     throw new Error('message is required');
@@ -55,7 +55,7 @@ export function sendAddonAbuseReport(
 
   return {
     type: SEND_ADDON_ABUSE_REPORT,
-    payload: { addonSlug, errorHandlerId: errorHandler.id, message },
+    payload: { addonSlug, errorHandlerId, message },
   };
 }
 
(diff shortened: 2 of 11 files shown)


tofumatt:
The two tests that are failing are oddly totally unrelated and not code I touched. Trying to see what's up with that...

tofumatt:
Hmmm, oddly b62d2d07557c4bda335f42f23a154d27589b6d1f fixed it. Strange.

kumar303:
Just a small lint error in there now

tests/unit/core/reducers/test_abuse.js
```diff
@@ -0,0 +1,125 @@
...
+      let message;
+      let response;
+
+      beforeEach(() => {
+        addon = { ...fakeAddon, slug: 'Ego' };
+        message = 'I am Groot!';
+        response = createFakeAddonAbuseReport({ addon, message });
+      });
+
+      it('saves the abuse report response to the reducer', () => {
```
kumar303: I'd like to see some additional tests that make sure multiple reports for different add-ons (or the same add-on) preserve existing state since that code can break in a refactor.

tests/unit/core/sagas/test_abuse.js
```diff
@@ -0,0 +1,94 @@
...
+    sagaTester.dispatch(sendAddonAbuseReport({
+      addonSlug: fakeAddon.slug,
+      errorHandler: createStubErrorHandler('my-abuse-error'),
+      message: 'Testing',
+      ...params,
+    }));
+  }
+
+  it('calls the API for abuse', async () => {
+    const addon = { ...fakeAddon, slug: 'fancy' };
```
kumar303: I don't think this should inherit from `...fakeAddon` since that will include a lot more fields than what's returned from the API.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type SendAddonAbuseReportAction = {|
+  addonSlug: string,
+  errorHandler: ErrorHandlerType,
+  message: string,
+|};
+
+export function sendAddonAbuseReport(
+  { addonSlug, errorHandler, message }: SendAddonAbuseReportAction
```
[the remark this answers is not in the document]
tofumatt: Ah, okay! I think different actions handle it different ways and I just looked at prior art that used the entire object as I forgot which we preferred. That makes sense, I'll change it over.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,81 @@
...
+
+  return {
+    type: SEND_ADDON_ABUSE_REPORT,
+    payload: { addonSlug, errorHandlerId: errorHandler.id, message },
+  };
+}
+
+export const initialState = {};
+
+type ReducerState = {};
```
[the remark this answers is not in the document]
tofumatt: Yeah, I think I forgot to do this one as I wasn't consuming this state anywhere yet so didn't think much of it. Your suggestion makes sense.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,81 @@
...
+export default function abuseReducer(
+  state: ReducerState = initialState,
+  action: Object
+) {
+  switch (action.type) {
+    case LOAD_ADDON_ABUSE_REPORT: {
+      const { addon, message, reporter } = action.payload;
+      return {
+        ...state,
+        [addon.slug]: { message, reporter },
```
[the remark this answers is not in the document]
tofumatt: Ahhh. Right on, I like it. Will do!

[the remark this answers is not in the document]
tofumatt: Hah, so I will try to give the cliffnotes version here but I can explain more on video or IRC if you like...

The current abuse reports aren't stored anywhere. Or rather, they aren't stored in the API or even in a DB if I recall correctly. They're emailed to admins and people take a look. So we're storing them essentially to prevent the same user submitting multiple abuse reports in a single session. But that's it for now. They aren't stored anywhere so I'm just storing them here because it seems the right thing to do *and* I can refuse multiple reports per-session. But yeah, that's about it. The abuse reporting features are pretty under-developed 😢

src/core/sagas/abuse.js
```diff
@@ -0,0 +1,50 @@
...
+      reporter: response.reporter,
+    }));
+  } catch (error) {
+    log.warn(`Reporting add-on for abuse failed: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* abuseSaga() {
+  yield takeEvery(SEND_ADDON_ABUSE_REPORT, reportAddon);
```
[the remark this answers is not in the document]
tofumatt: I suppose so. Will `takeLatest` automatically cancel the old one though?

But I guess `takeLatest` should be fine, yeah.

tests/unit/core/api/test_abuse.js
```diff
@@ -0,0 +1,69 @@
...
+      return createApiResponse({
+        jsonData: createFakeAddonAbuseReport({ addon, message }),
+      });
+    }
+
+    it('should allow anonymous users to report an add-on', () => {
+      const message = 'I do not like this!';
+
+      mockWindow.expects('fetch')
+        .withArgs(`${apiHost}/api/v3/abuse/report/addon/?addon=cool-addon&message=I%20do%20not%20like%20this!&lang=en-US`)
```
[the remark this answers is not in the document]
tofumatt: Ah, you're totally right, I think you've pointed that out before and I forgot. A lot of times I just copy these tests from elsewhere as a template.

Maybe in Q4 we can maintain some nice templates for starting new components, component tests, api tests, etc. 😆

tests/unit/core/api/test_abuse.js
```diff
@@ -0,0 +1,69 @@
...
+        .once()
+        .returns(mockResponse({
+          addon: { ...fakeAddon, slug: 'cool-addon' },
```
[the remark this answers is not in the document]
tofumatt: Yeah, I actually have a helper that transforms an entire `addon` to just what the abuse report returns. I'm using it elsewhere but made it part-way through the patch and forgot to use it here. I'll add it in!

src/core/reducers/abuse.js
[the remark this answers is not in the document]
kumar303: > I'm just storing them here because ... I can refuse multiple reports per-session

In that case, maybe a lookup structure keyed by add-on ID and reporter?

```js
return {
  ...state,
  bySlug: {
    ...state.bySlug,
    [addon.slug]: {
      ...state.bySlug[addon.slug],
      [reporter]: message,
    },
  },
};
```

We should definitely only build state for what we need :)

tests/unit/core/sagas/test_abuse.js
tofumatt: I just checked and this uses `createFakeAddonAbuseReport` which only returns the relevant parts of the add-on in the response. So this is okay.

tests/unit/core/reducers/test_abuse.js
tofumatt: We shouldn't allow multiple abuse reports for the same add-on (I was thinking I'd add that test once I made the UI for it too) but I'll do it here. And yeah, multiple reports across add-ons should be tested, good call.

tofumatt:
Alright, keyed things by slug and made the other requested changes. Should be good for another review!

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+}
+
+export const initialState = {
+  bySlug: {},
+};
+
+type ReducerState = {|
+  bySlug: {
+    [addonSlug: string]: Array<{| message: string, reporter: string |}>,
```
kumar303: The way you have it now is not an Array, it's

```js
[addonSlug: string]: {| message: string, reporter: string |},
```

tests/unit/core/api/test_abuse.js
```diff
@@ -0,0 +1,91 @@
...
+
+describe(__filename, () => {
+  let mockApi;
+
+  beforeEach(() => {
+    mockApi = sinon.mock(api);
+  });
+
+  describe('reportAddon', () => {
+    function _reportAddon(extraArguments = {}) {
```
kumar303: This helper doesn't seem to be helping. I'd suggest using `reportAddon()` directly instead.

tests/unit/core/reducers/test_abuse.js
```diff
@@ -0,0 +1,164 @@
...
+} from 'core/reducers/abuse';
+import { dispatchClientMetadata, fakeAddon } from 'tests/unit/amo/helpers';
+import { createFakeAddonAbuseReport } from 'tests/unit/helpers';
+
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('initializes properly', () => {
+      const state = abuseReducer(initialState, { type: 'UNRELATED_ACTION' });
+      expect(state).toEqual({ bySlug: {} });
```
kumar303: I think it would be better to make this assertion:

```js
expect(state).toEqual(initialState);
```
Otherwise you'd have to update the test every time it changes which doesn't seem necessary.

tests/unit/core/reducers/test_abuse.js
```diff
@@ -0,0 +1,164 @@
...
+      it('saves the abuse report response to the reducer', () => {
+        const state = abuseReducer(
+          initialState, loadAddonAbuseReport(response));
+
+        expect(state.bySlug[addon.slug].message).toEqual(message);
+      });
+
+      it('requires an addon', () => {
+        expect(() => {
+          loadAddonAbuseReport({
```
kumar303: These tests will be easier to maintain if they use the same approach as above:

```js
const partialParams = { ...defaultParams };
delete partialParams.addon;
loadAddonAbuseReport(partialParams);
```

src/core/api/abuse.js
```diff
@@ -0,0 +1,23 @@
...
+  api: ApiStateType,
+  auth?: boolean,
+  message: string,
+|};
+
+export function reportAddon(
+  { addonSlug, api, auth = false, message }: ReportAddonParams
+) {
+  return callApi({
+    auth,
```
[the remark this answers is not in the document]
tofumatt: Oh! Good to know! I always found it weird. Okay, cool! Thanks.

src/core/reducers/abuse.js
tofumatt: Aha, right. I copied your flow thing and wasn't paying attention. My bad.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,95 @@
...
+
+export default function abuseReducer(
+  state: ReducerState = initialState,
+  action: Object
+) {
+  switch (action.type) {
+    case SEND_ADDON_ABUSE_REPORT: {
+      const { addonSlug } = action.payload;
+      if (state.bySlug[addonSlug]) {
+        throw new Error('You already reported this add-on.');
```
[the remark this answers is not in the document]
tofumatt: Well, reports can be anonymous. But yeah, I threw this error here but thought: the saga will catch this and dispatch the error that the component would actually watch for and render a real message about.

But without seeing that in action it's weird. This isn't the user-surfaced message, but I figured it at least tests the logic.

Still think it's worth deleting? I could leave it as-is and tweak it in the UI patch, but also fine to just ditch it for now.

tests/unit/core/api/test_abuse.js
tofumatt: Hah, yeah, true. It used to help! 😉

tests/unit/core/reducers/test_abuse.js
tofumatt: True, I think I just got lazy. Will do!

src/core/reducers/abuse.js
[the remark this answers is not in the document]
tofumatt: Okay actually I see what you mean. I'll remove it here and handle it in the UI patch.

tofumatt:
Alrighty, last round of changes addressed.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,87 @@
...
+
+type ReducerState = {|
+  bySlug: {
+    [addonSlug: string]: {| message: string, reporter: string |},
+  },
+|};
+
+export default function abuseReducer(
+  state: ReducerState = initialState,
+  action: Object
```
kumar303: btw, the way to type this would be:

```js
action: SendAddonAbuseReportAction
```

but then you also have to type your constants as string literals [like this](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/addonsByAuthors.js#L17) so Flow can figure out which action object is being operated on in each switch case:

```js
export const LOAD_ADDON_ABUSE_REPORT: 'LOAD_ADDON_ABUSE_REPORT' = 'LOAD_ADDON_ABUSE_REPORT';
```

`Object` is ok if you want to leave it as is but you will get some nice safety checks in your reducer code.

tests/unit/core/api/test_abuse.js
```diff
@@ -0,0 +1,85 @@
...
+        })
+        .once()
+        .returns(mockResponse({
+          addon: { ...fakeAddon, slug: 'auth-addon' },
+          message,
+        }));
+      return reportAddon({
+        addonSlug: 'auth-addon',
+        api: apiState,
+        auth: true,
```
kumar303: This can be deleted now

tests/unit/core/api/test_abuse.js
```diff
@@ -0,0 +1,85 @@
...
+        addonSlug: 'cool-addon',
+        api: apiState,
+        message,
+      })
+        .then(() => {
+          mockApi.verify();
+        });
+    });
+
+    it('should allow signed-in users to report an add-on', () => {
```
kumar303: It would be nice to update this spec to avoid confusion. The API call applies to both signed in or anonymous users.

src/core/reducers/abuse.js
```diff
@@ -0,0 +1,87 @@
...
+  };
+}
+
+export const initialState = {
+  bySlug: {},
+};
+
+type ReducerState = {|
+  bySlug: {
+    [addonSlug: string]: {| message: string, reporter: string |},
```
kumar303: I think it should be `reporter: Object | null`

tests/unit/core/api/test_abuse.js
tofumatt: Yeah I actually think now that `auth: true` is always set this can just be removed.

Issue #3176, opened 2017-09-19
Keep "More extensions by author_name" section in the same position


ValentinaPC:
Steps to reproduce:
1. Load details page for an add-on that has release notes i.e. https://addons-dev.allizom.org/en-US/firefox/addon/xnotifier/
2. Load details page for an add-on that does not have release notes i.e. https://addons-dev.allizom.org/en-US/firefox/addon/ratopuse-weasp-vulturkal/
3. Compare "More extensions by author_name" section position from both pages.

Expected results:
The section is positioned in the same place in both pages.

Actual results:
The "More extensions by author_name" section is after More information section for the add-on with release notes and before this section for the add-on without release notes.

Notes/Issues:
Verified on AMO-dev FF55(Win 7).
Screenshot for this issue: 
![more](https://user-images.githubusercontent.com/15685960/30591419-6b05f40a-9d4b-11e7-9ac5-ae566bf24a0a.png)

tofumatt:
This looks like it could be a grid issue but I don't even see "More Extensions" for xnotifier.

Ah, I see. Please make sure to specify the viewport size you're using when filing issues like this as they are out-of-order on mobile but one is entirely obscured on Desktop.

See:

<img width="1280" alt="screenshot 2017-09-19 13 22 10" src="https://user-images.githubusercontent.com/90871/30592189-b90cc092-9d3d-11e7-9cc8-fa3de89101a3.png">
<img width="1280" alt="screenshot 2017-09-19 13 22 07" src="https://user-images.githubusercontent.com/90871/30592188-b904c82e-9d3d-11e7-8219-751c1911d719.png">

I'm considering it the same issue though as it's a grid bug.

I'll try to get this fixed up before the tag right now as it should be easy and is sort of annoying 😄

tofumatt:
Oddly this is worse than reported. I can't see more add-ons on the frontend by the author at all... yet they appear on the old site. Something is amiss and this will require more investigation...

ValentinaPC:
@tofumatt : my initial screenshots look like I am using small browser size but I am not, just cropped the important part of the screen.
I am using same as always: 1600 x 900 px. 
Thanks!

tsl143:
@tofumatt, the `more addons by author` section, will be visible on chrome for this sepcific addon
`http://localhost:3000/en-US/firefox/addon/xnotifier/`.
[...]

Issue #3130, opened 2017-09-14
"More extensions by author" card is not loaded on the server


tofumatt:
I just noticed that the "More extensions by [author]" content is only loaded client-side, after the content has been sent from the server.

We should be loading this server-side and not seeing it load in after the page has been sent over the wire.

### Screenshot
![2017-09-14 01 30 22](https://user-images.githubusercontent.com/90871/30407211-d246a27e-98ee-11e7-8853-7a72142cc912.gif)

willdurand:
Indeed, but how is this possible actually? We are not waiting for the saga to finish in SSR? Oh. No. We need an add-on because the addon type is required. That's the problem I think..

willdurand:
The way data fetching and sagas work on the server makes it nearly impossible to fix this issue, so this is a wontfix.

PR #3179, opened 2017-09-19
fix: More extensions by author_name - section

tsl143:
Fixes: #3176 

**Before/After** 
screen shot 2017-09-19 at 10 33 13 pm
screen shot 2017-09-19 at 10 32 53 pm

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -254,7 +254,7 @@
 
   .AddonMoreInfo {
     grid-column: 1 / 2;
-    grid-row: 4 / span 1;
+    grid-row: 5 / span 1;
 
     .Addon--has-more-than-0-addons.Addon-persona & {
       grid-row: 3 / span 1;


muffinresearch:
@tsl143 thanks for the pathc. Please can you add some details with rationale for the approach taken? This will help anyone not very familiar with CSS grids (myself included) to understand the fix (and learn something along the way) 👍

tsl143:
@muffinresearch I too am trying to understand the CSS-grids (and I am loving it:) )
I followed https://css-tricks.com/snippets/css/complete-guide-grid/#article-header-id-26
So for this issue refer `grid-row: <start-line> / <end-line> | <start-line> / span <value>;`
Here we are following second pattern `grid-row: <start-line> / span <value>;`,
Initially `.AddonMoreInfo` was in 4th line, now we have 5 lines(boxes) So I gave it last line number that is  `grid-row: 5 / span 1;`

Hope that helps.

muffinresearch:
If we specify something to be on the 5th line what happens if the thing that should be 4th line doesn't render e.g. if for that add-on data is missing?

tsl143:
It will auto adjust, taking the last line for row consideration ie. 4th row.

tofumatt:
Oddly I'm still not seeing these being loaded at all. I wonder if this is related to #3130?

tofumatt:
Oh wait, maybe it's related to a weird API bug where versions 57+ are adding in the `appversion=57` param to any search. I bet that's it.

PR #3182, opened 2017-09-20
Workaround erroneous span focus and outline user button on focus.

### LINKED ISSUE #3110: User menu is not keyboard accessible
The dropdown user menu needs to be keyboard accessible otherwise those links are completely obscured to anyone that can't use a mouse.

The username menu should be focusable, and when focused that should open the menu and make it possible to tab down through the links.

add-ons_for_firefox

muffinresearch:
Fixes #3110

~~This is a workaround for #3110 - I can't find a root cause as to why the span is focusable. Since it doesn't need to be focused adding an explicit attribute fixes the problem.~~

~~At some point it would be nice to understand what the root cause of the span being focusable even when it's tabIndex property says it's -1 already.~~

~~Trying to reproduce it in isolation doesn't appear to be easy.~~

~~This also adds the outline so that the button is outlined on focus.~~

The fix to prevent the span focus was to remove the `overflow:hidden` which was done in a different PR.

This  PR adds the outline for FF and the default webkit focus back for chrome etc:

## Firefox:

add-ons_for_firefox

add-ons_for_firefox

## Chrome

add-ons_for_firefox

add-ons_for_firefox
#

### DIFF
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -34,6 +34,13 @@
 }
 
 .DropdownMenu-button {
+  &:focus {
+    // Fix outline for Firefox (see #3110)
+    outline: 1px dotted $white;
+    // Restore default outline for webkit.
+    outline: auto 5px -webkit-focus-ring-color;
+  }
+
   .DropdownMenu-button-text {
     @include font-regular();
 


muffinresearch:
This seems to mess with the default focus on chrome.

muffinresearch:
Further looking into this and trying to isolate it found that the problem is coming in via the CSS. I've not had time yet to bisect the CSS to find the cause, once I do I'm expecting this should be fairly easy to remedy.

muffinresearch:
The fix for the span focus was making overflow visible on the span. This was previously fixed so now the only additional fix is making the focus style more obvious.

Issue #3184, opened 2017-09-20
Make the search look more obviously focused


muffinresearch:
Unfocused: 

<img width="308" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30636871-91632458-9dee-11e7-9041-6094e78a98ef.png">

Focused:

<img width="311" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30636888-a319fece-9dee-11e7-809f-47246238d8d2.png">

For styling we should look at something similar for what we do for the logged-out sign in button.

<img width="336" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30636950-d6496b40-9dee-11e7-8e82-5f2ca0ca8211.png">

AlexandraMoga:
Verified in AMO-Dev Mobile view and the search field appears focused as expected:

![focused search](https://user-images.githubusercontent.com/31961530/30648209-2d54c952-9e26-11e7-8e7f-5ee4308711cb.png)

PR #3187, opened 2017-09-20
fix: Make the search look more obviously focused

tsl143:
Fixes: #3184 

**Before:**
screen shot 2017-09-20 at 3 32 44 pm

**After:**
screen shot 2017-09-20 at 3 35 32 pm

### DIFF
--- a/src/amo/components/SearchForm/styles.scss
+++ b/src/amo/components/SearchForm/styles.scss
@@ -28,6 +28,11 @@
   white-space: nowrap;
   width: 100%;
 
+  &:focus {
+    border: 1px solid;
+    @include focus();
+  }
+
   @include respond-to(medium) {
     @include padding-end(28px);
   }
--- a/src/amo/components/SearchForm/styles.scss
+++ b/src/amo/components/SearchForm/styles.scss
@@ -17,7 +17,7 @@
   @include padding-end(6px);
   @include padding-start(28px);
 
-  border: 1px solid $teal-50;
+  border: 1px solid $white;
   border-radius: $border-radius-s;
   color: $black;
   height: 28px;
@@ -28,8 +28,11 @@
   white-space: nowrap;
   width: 100%;
 
-  &:focus {
+  &:focus,
+  &:hover {
     @include focus();
+
+    border-color: $teal-50;
   }
 
   @include respond-to(medium) {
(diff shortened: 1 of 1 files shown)


muffinresearch:
@tsl143 that was fast!

I think the border colour for the input should be `$teal-50` since the outline on a white background gets lost and this will make it stand out.

To prevent the placeholder text moving we'll probably want a default border applied.

Whilst we're here it might be nice to use `$teal-70` as a hover style just as a border only.

muffinresearch:
The aforementioned teal vars are in https://github.com/mozilla/addons-frontend/blob/159de138dfc377937db810e1e3b3f9a1c954fb65/src/amo/components/SearchFilters/styles.scss#L1 in case you need to import it.

tsl143:
@muffinresearch , updated the code with border,
Also we dont need to import `~photon-colors/colors` as its already imported in mixins.

tsl143:
**Without focus:**
screen shot 2017-09-20 at 4 24 55 pm

**With focus:**
screen shot 2017-09-20 at 4 25 02 pm

src/amo/components/SearchForm/styles.scss
```diff
@@ -17,7 +17,7 @@
   @include padding-end(6px);
   @include padding-start(28px);
 
-  border: 0;
+  border: 1px solid $teal-50;
```
muffinresearch: `$teal-50` should only be on focus - but a default border will be needed too - probably transparent or the same as the current input background.

muffinresearch: Sorry I realise I screwed up my description above and didn't qualify that teal-50 was for the focus style only. I've edited that now.

src/amo/components/SearchForm/styles.scss
```diff
@@ -28,6 +28,13 @@
   white-space: nowrap;
   width: 100%;
 
+  &:focus,
+  &:hover {
+    @include focus();
```
muffinresearch: Can we split `:focus` and `:hover` so that the  `focus()` mixin is *only* used for focus?

The hover style should be a 1px border using `$teal-70` rather than `$teal-50`.

tsl143:
I have put `focus` after `hover` so as to prioritize `$teal-50` when user types in the box.

tsl143:
Yes that makes it look better :)

Issue #3186, opened 2017-09-20
Fix the minor difference between button height and search box height


muffinresearch:
On mobile there's a small difference between the search box height and the login button. 

![screen-shot-2017-09-20-at-10 56 30](https://user-images.githubusercontent.com/1514/30638328-f6fdfbf4-9df2-11e7-8a53-a873132a0123.png)

It's a really minor difference, but would be nice to fix it for consistency.

muffinresearch:
See also https://github.com/mozilla/addons-frontend/issues/3185 - which affects a related area and could be fixed at the same time.

tofumatt:
I think there’s actually supposed to be a height difference… but at least vertically centering the button inside would be nicer. I think it was supposed to be… if the login button were moved 1px up that would work I think.

muffinresearch:
For me, the height difference stands out a mile and looks like a bug, but if aligning-top helps alleviate my inner sense of discord that could work too. Either way it's super minor issue.

vcarciu:
This looks fixed now : 
![align](https://user-images.githubusercontent.com/1583842/30738507-fb6e4c10-9f92-11e7-8f1f-ed37692a8620.png)

PR #3191, opened 2017-09-20
fix : height fix for search box and register btn

tsl143:
Fix: #3186 

**Before:**
screen shot 2017-09-20 at 6 13 11 pm

**After:**
screen shot 2017-09-20 at 6 12 41 pm

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -98,7 +98,7 @@
 
   grid-column: 3 / span 2;
   grid-row: 1 / 2;
-  margin-top: 11px;
+  margin-top: 10px;
 
   @include respond-to(medium) {
     grid-column: 4 / -1;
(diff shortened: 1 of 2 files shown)


tofumatt:
Frustratingly now it's misaligned by being one pixel too high. It would be nice if there were a `1px` margin on both top and bottom.

What's the padding like on the search box? Maybe it's best to knock it down `1px`.

muffinresearch:
Looks to me like the register button could easy handle some more of bottom padding and less top padding. The text looks off-centre though that might be partly an optical illusion.  

Might be possible to add the missing 1px that way too. It would be nice to get the text baselines of the input placeholder and the register button text more closely aligned.

tofumatt:
That button style is from the spec, though. Changing just the one button is frustrating because then it doesn’t follow the spec. Adding the padding will change things elsewhere.

muffinresearch:
> That button style is from the spec, though. Changing just the one button is frustrating because then it doesn’t follow the spec. Adding the padding will change things elsewhere.

Yep, that's a good point.

tsl143:
@muffinresearch, Again nsp-check fails after master merge, 
But the fix @tofumatt suggested works
screen shot 2017-09-21 at 2 50 15 pm

muffinresearch:
Thanks @tsl143

Issue #3185, opened 2017-09-20
Menu jumps vertically on logout


muffinresearch:
When logging out the user menu jumps vertically see the following animated gif:

![jumping-menu-logout](https://user-images.githubusercontent.com/1514/30638537-a6685fa8-9df3-11e7-8379-8e0f3ca8a259.gif)

Fixing this requires matching the logged in heights/padding/margin etc with the styles for logged-out so there's no vertical displacement.

AlexandraMoga:
verified on FF56 (Win 10 and Android 7.0)

Postfix video:

![vertical jump menu](https://user-images.githubusercontent.com/31961530/31075857-3fb4d2d2-a780-11e7-98b9-ed631c596f8f.gif)

PR #3194, opened 2017-09-21
fix: Menu jumps vertically on logout

tsl143:
Fixes: #3185 

Fixed GIF
menu

### DIFF
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -6,7 +6,7 @@
   @include text-align-start();
 
   display: inline-block;
-  padding-bottom: 6px;
+  padding-bottom: 2px;
   padding-top: 4px;
   position: relative;
 
@@ -24,6 +24,10 @@
         opacity: 1;
       }
     }
+  }
+
+  @include respond-to(medium) {
+    padding-bottom: 6px;
   }
 }
 
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -6,6 +6,7 @@
   @include text-align-start();
 
   display: inline-block;
+  height: 27px;
   padding-bottom: 2px;
   padding-top: 4px;
   position: relative;
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -7,7 +7,7 @@
 
   display: inline-block;
   padding-bottom: 6px;
-  padding-top: 8px;
+  padding-top: 4px;
   position: relative;
 
   // This is needed because `:hover` on mobile prevents the main button to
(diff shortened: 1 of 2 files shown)


tsl143:
@muffinresearch, `yarn nsp-check` even fails at master branch. Did we merge something break-ish ?

tofumatt:
This seems fine, can you just confirm it doesn't change anything on other screen sizes? 😄

tofumatt:
(that's responsive mode in Firefox at 320 x 480, btw)

tsl143:
@tofumatt , got confused between screen size variables then I got `mobile-first` concept :)
That's cool, and I am going to use that in my projects too.

muffinresearch:
> @muffinresearch, yarn nsp-check even fails at master branch. Did we merge something break-ish ?
See https://github.com/mozilla/addons-frontend/issues/3195

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -6,8 +6,8 @@
   @include text-align-start();
 
   display: inline-block;
-  padding-bottom: 6px;
-  padding-top: 8px;
+  padding-bottom: 2px;
+  padding-top: 4px;
```
willdurand: I should have left a comment obviously. The `padding-bottom` was important to avoid the menu to auto-close itself (because of focus lost) when user moves the cursor from username to the menu itself.

willdurand: I tested locally and it is all good.

willdurand: I still have a different between the menu and login button. It seems there is a `0.25px` difference. Could you set the height of the `DropdownMenu` to `27px`?

tsl143: @willdurand , I was looking at desktop version where `height: 27px` had no effect :(
Then I saw the mobile version :P , I forgot #mobileFirst
Here you go -
Mobile:
mobile


Desktop:
ezgif-4-2d44bd1686

willdurand: Yeah, sorry for that. There is actually an issue on desktop now :-(

willdurand:
Fixing the height on mobile devices make it work like a charm. However, this broke the desktop mode (focus lost when trying to reach the menu from the username):

willdurand:
@tsl143 sorry, can you try adding `height: 30px` for desktop? (in `@include respond-to(large)`)

tsl143:
@willdurand Instead what about adding height auto there?

willdurand:
> @willdurand Instead, how about adding height: auto there?

if that works, why not. I tested locally with `30px` to be honest, so I now it should be fine. I'll test with `auto` then :)

kumar303:
@willdurand has confirmed that the issue in https://github.com/mozilla/addons-frontend/pull/3194#pullrequestreview-64244062 is resolved.

Thanks @tsl143 !

Issue #3195, opened 2017-09-21
yarn nsp-check is failing re: tough-cookie ReDOS


muffinresearch:
This is due to https://nodesecurity.io/advisories/525

Running locally to get the extended tree shows tough-cookie is coming in via:

`mozilla-addons-frontend@0.0.1 > jsdom@11.2.0 > request-promise-native@1.0.4 > tough-cookie@2.3.2`

I think this being a dep of jsdom would mean this should only impacts tests if at all. This needs to be confirmed.

The upstream issue is: https://github.com/salesforce/tough-cookie/issues/92

tphan18:
I have the same vulnerability report this morning from a bunch of dependencies like 
 > react-scripts@1.0.13 > jest@20.0.4 > jest-cli@20.0...

kumar303:
Looks like we don't have any patches yet. We should probably get master back to green so we don't miss any unrelated failures. This should do it in a `.nsprc`:

```
{
  "exceptions": ["https://nodesecurity.io/advisories/525"]
}
```

kumar303:
When we get an upstream fix, let's remove the exception added in https://github.com/mozilla/addons-frontend/pull/3211

kumar303:
> I think this being a dep of jsdom would mean this should only impacts tests if at all. 

I also see us using `tough-cookie` for `chokidar` which is just a file watcher which we use purely for development.

```
$ npm ls tough-cookie
mozilla-addons-frontend@0.0.1 /Users/kumar/dev/addons-frontend
├─┬ chokidar-cli@1.2.0
│ └─┬ chokidar@1.7.0
│   └─┬ fsevents@1.1.2
│     └─┬ node-pre-gyp@0.6.36
│       └─┬ request@2.81.0
│         └── tough-cookie@2.3.2
├─┬ jsdom@11.2.0
│ └── tough-cookie@2.3.2
├── UNMET PEER DEPENDENCY react@15.6.2
├── UNMET PEER DEPENDENCY react-redux@5.0.6
└── UNMET PEER DEPENDENCY webpack@3.1.0
```

PR #3203, opened 2017-09-22
feat: Add a link to the stats dashboard for add-on owners

### LINKED ISSUE #2751: User count should link to stats dashboard
If signed-in user has access to an add-on's statistics dashboard, the user count should link to it.

Alternatively we could add a link next to the count with (View stats), as the contextual link on a small element like that feels like sub-par UX.

tofumatt:
Fixes #2751.

Links to the stats dashboard from the user count on the `AddonMeta` section and also adds a link in the dashboard.

In order to see this locally you'll be signed in and visit an add-on you own.

### Screenshots
screenshot 2017-09-22 02 50 47
screenshot 2017-09-22 02 41 53

### DIFF
--- a/src/amo/components/AddonMeta.js
+++ b/src/amo/components/AddonMeta.js
@@ -1,22 +1,29 @@
+/* @flow */
 import React from 'react';
-import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { compose } from 'redux';
 
+import Link from 'amo/components/Link';
 import translate from 'core/i18n/translate';
+import { isAddonAuthor } from 'core/utils';
+import type { AddonType } from 'core/types/addons';
 import LoadingText from 'ui/components/LoadingText';
 import Rating from 'ui/components/Rating';
 
 import 'amo/css/AddonMeta.scss';
 
 
+type PropTypes = {
+  addon: AddonType | null,
+  i18n: Object,
+  userId: number | null,
+}
+
 export class AddonMetaBase extends React.Component {
-  static propTypes = {
-    addon: PropTypes.object.isRequired,
-    i18n: PropTypes.object.isRequired,
-  }
+  props: PropTypes;
 
   render() {
-    const { addon, i18n } = this.props;
+    const { addon, i18n, userId } = this.props;
     const averageRating = addon ? addon.ratings.average : null;
     const addonRatingCount = addon ? addon.ratings.count : null;
 
@@ -49,7 +56,14 @@ export class AddonMetaBase extends React.Component {
         <div className="AddonMeta-item AddonMeta-users">
           <h3 className="visually-hidden">{i18n.gettext('Used by')}</h3>
           <p className="AddonMeta-text AddonMeta-user-count">
-            {userCount}
+            {addon && isAddonAuthor({ addon, userId }) ? (
+              <Link
+                href={`/addon/${addon.slug}/statistics/`}
+                title={i18n.gettext('Click to view statistics')}
+              >
+                {userCount}
+              </Link>
+            ) : userCount}
           </p>
           <p className="AddonMeta-text AddonMeta-review-count">
             {reviewCount}
@@ -66,6 +80,13 @@ export class AddonMetaBase extends React.Component {
   }
 }
 
+export const mapStateToProps = (state: Object) => {
+  return {
+    userId: state.user.id,
+  };
+};
+
 export default compose(
-  translate({ withRef: true }),
+  connect(mapStateToProps),
+  translate(),
 )(AddonMetaBase);
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -1,25 +1,30 @@
+/* @flow */
 import React from 'react';
-import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import Link from 'amo/components/Link';
 import ReportAbuseButton from 'amo/components/ReportAbuseButton';
 import translate from 'core/i18n/translate';
-import { trimAndAddProtocolToUrl } from 'core/utils';
+import type { AddonType } from 'core/types/addons';
+import { isAddonAuthor, trimAndAddProtocolToUrl } from 'core/utils';
 import Card from 'ui/components/Card';
 import LoadingText from 'ui/components/LoadingText';
 
 import './styles.scss';
 
 
+type PropTypes = {
+  addon: AddonType | null,
+  i18n: Object,
+  userId: number | null,
+}
+
 export class AddonMoreInfoBase extends React.Component {
-  static propTypes = {
-    addon: PropTypes.object.isRequired,
-    i18n: PropTypes.object.isRequired,
-  }
+  props: PropTypes;
 
   listContent() {
-    const { addon, i18n } = this.props;
+    const { addon, i18n, userId } = this.props;
 
     if (!addon) {
       return this.renderDefinitions({
@@ -52,6 +57,14 @@ export class AddonMoreInfoBase extends React.Component {
     return this.renderDefinitions({
       homepage,
       supportUrl,
+      statsLink: addon && isAddonAuthor({ addon, userId }) ? (
+        <a
+          className="AddonMoreInfo-stats-link"
+          href={`/addon/${addon.slug}/statistics/`}
+        >
+          {i18n.gettext('Visit stats dashboard')}
+        </a>
+      ) : null,
       version: addon.current_version.version,
       versionLastUpdated: i18n.sprintf(
         // translators: This will output, in English:
@@ -110,6 +123,7 @@ export class AddonMoreInfoBase extends React.Component {
   renderDefinitions({
     homepage = null,
     supportUrl = null,
+    statsLink = null,
     privacyPolicyLink = null,
     eulaLink = null,
     version,
@@ -118,7 +132,7 @@ export class AddonMoreInfoBase extends React.Component {
     versionHistoryLink,
     betaVersionsLink = null,
     addonId,
-  }) {
+  }: Object) {
     const { i18n } = this.props;
     return (
       <dl className="AddonMoreInfo-contents">
@@ -171,6 +185,12 @@ export class AddonMoreInfoBase extends React.Component {
           </dt>
         ) : null}
         {betaVersionsLink ? <dd>{betaVersionsLink}</dd> : null}
+        {statsLink ? (
+          <dt className="AddonMoreInfo-stats-title">
+            {i18n.gettext('Usage Statistics')}
+          </dt>
+        ) : null}
+        {statsLink ? <dd>{statsLink}</dd> : null}
         <dt
           className="AddonMoreInfo-database-id-title"
           title={i18n.gettext(`This ID is useful for debugging and
@@ -201,6 +221,13 @@ export class AddonMoreInfoBase extends React.Component {
   }
 }
 
+export const mapStateToProps = (state: Object) => {
+  return {
+    userId: state.user.id,
+  };
+};
+
 export default compose(
+  connect(mapStateToProps),
   translate(),
 )(AddonMoreInfoBase);
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -58,12 +58,12 @@ export class AddonMoreInfoBase extends React.Component {
       homepage,
       supportUrl,
       statsLink: addon && isAddonAuthor({ addon, userId }) ? (
-        <a
+        <Link
           className="AddonMoreInfo-stats-link"
           href={`/addon/${addon.slug}/statistics/`}
         >
           {i18n.gettext('Visit stats dashboard')}
-        </a>
+        </Link>
       ) : null,
       version: addon.current_version.version,
       versionLastUpdated: i18n.sprintf(
@@ -75,12 +75,12 @@ export class AddonMoreInfoBase extends React.Component {
         }
       ),
       versionLicenseLink: addon.current_version.license ? (
-        <a
+        <Link
           className="AddonMoreInfo-license-link"
           href={addon.current_version.license.url}
         >
           {addon.current_version.license.name}
-        </a>
+        </Link>
       ) : null,
       privacyPolicyLink: addon.has_privacy_policy ? (
         <Link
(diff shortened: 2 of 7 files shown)


tofumatt:
Failing test is #3195 again.

src/amo/components/AddonMeta.js
```diff
@@ -1,22 +1,28 @@
 import React from 'react';
```
kumar303: This needs `/* @flow */` if you're going to use it for prop types.

src/amo/components/AddonMeta.js
```diff
@@ -66,6 +79,13 @@ export class AddonMetaBase extends React.Component {
   }
 }
 
+export const mapStateToProps = (state) => {
+  return {
+    userId: state.user ? state.user.id : null,
```
kumar303: Due to the beauty of Redux, `state.user` will never be falsy. You can just do `userId: state.user.id` -- `id` will be null when the user is not signed in.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -164,6 +180,13 @@ export class AddonMoreInfoBase extends React.Component {
   }
 }
 
+export const mapStateToProps = (state) => {
+  return {
+    userId: state.user ? state.user.id : null,
```
kumar303: same thing here

src/amo/components/AddonMeta.js
```diff
@@ -1,22 +1,28 @@
...
 import LoadingText from 'ui/components/LoadingText';
 import Rating from 'ui/components/Rating';
 
 import 'amo/css/AddonMeta.scss';
 
 
+type PropTypes = {
+  addon: AddonType | null,
+  i18n: Object,
+  userId: number | null,
```
willdurand: Not sure we have discussed this yet (at least not with me), but Flow users tend to use `?number` to mark a nullable/maybe value: https://flow.org/en/docs/types/maybe/.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -1,24 +1,28 @@
...
+  i18n: Object,
+  userId: number | null,
+}
+
 export class AddonMoreInfoBase extends React.Component {
-  static propTypes = {
-    addon: PropTypes.object.isRequired,
-    i18n: PropTypes.object.isRequired,
-  }
+  props: PropTypes;
```
willdurand: Thanks!

src/amo/components/AddonMeta.js
willdurand: Also I would go for strict typing if possible.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -1,22 +1,34 @@
...
-    addon: fakeAddon,
-    i18n: getFakeI18nInst(),
-    ...customProps,
-  };
-  return shallow(<AddonMetaBase {...props} />);
-}
 
 describe('<AddonMeta>', () => {
+  function render({
+    addon = fakeAddon,
```
kumar303: We should get into the habit of doing `addon = createInternalAddon(fakeAddon)` because the internal addon object is what this component will receive. You can import it from https://github.com/mozilla/addons-frontend/blob/master/src/core/reducers/addons.js#L83

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -53,6 +65,31 @@ describe('<AddonMeta>', () => {
       });
       expect(getUserCount(root)).toMatch(/^1\.000/);
     });
+
+    it('links to stats if add-on author is viewing the page', () => {
+      const addon = {
```
kumar303: this should also be `const addon = createInternalAddon({ ...fakeAddon, ... })`

src/amo/components/AddonMeta.js
```diff
@@ -49,7 +55,14 @@ export class AddonMetaBase extends React.Component {
         <div className="AddonMeta-item AddonMeta-users">
           <h3 className="visually-hidden">{i18n.gettext('Used by')}</h3>
           <p className="AddonMeta-text AddonMeta-user-count">
-            {userCount}
+            {addon && isAddonAuthor({ addon, userId }) ? (
```
willdurand: There is probably no need for `addon &&` because it is checked in the `isAddonAuthor()` function.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -1,22 +1,34 @@
...
-function render({ ...customProps } = {}) {
-  const props = {
-    addon: fakeAddon,
-    i18n: getFakeI18nInst(),
-    ...customProps,
-  };
-  return shallow(<AddonMetaBase {...props} />);
-}
 
 describe('<AddonMeta>', () => {
```
willdurand: You could update the describe name to use `__filename` if you wish.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -53,6 +65,31 @@ describe('<AddonMeta>', () => {
...
       expect(getUserCount(root)).toMatch(/^1\.000/);
     });
+
+    it('links to stats if add-on author is viewing the page', () => {
+      const addon = {
+        ...fakeAddon,
+        slug: 'coolio',
+        authors: [
+          {
+            id: 11,
```
kumar303: Even though `fakeAddon` is not complete, it's good practice for future maintenance to always extend from the base object, like:

```js
authors: [
  {
    ...fakeAddon.authors[0],
    id: 11,
    name: 'tofumatt',
  },
],
```

The `fakeAddon` author could use some more fields in it too.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -1,22 +1,34 @@
...
 import React from 'react';
 
-import { AddonMetaBase } from 'amo/components/AddonMeta';
-import { fakeAddon } from 'tests/unit/amo/helpers';
-import { getFakeI18nInst } from 'tests/unit/helpers';
+import AddonMeta, { AddonMetaBase } from 'amo/components/AddonMeta';
+import Link from 'amo/components/Link';
+import {
+  dispatchSignInActions,
+  dispatchClientMetadata,
```
willdurand: This should be imported first.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -53,6 +65,31 @@ describe('<AddonMeta>', () => {
...
+            name: 'tofumatt',
+            picture_url: 'http://cdn.a.m.o/myphoto.jpg',
+            url: 'http://a.m.o/en-GB/firefox/user/tofumatt/',
+            username: 'tofumatt',
+          },
+        ],
+      };
+      const root = render({
+        addon,
+        store: dispatchSignInActions({ userId: 11 }).store,
```
kumar303: instead of using a magic number, this should reference a variable for the `userId` value.

src/amo/components/AddonMoreInfo/index.js
willdurand: We could also pass the `isAddonAuthor` (bool) prop instead of computing this value in the `render()` method. The `isAddonAuthor()` function could be a selector living into `reducers/user` (like `isAuthenticated()`). This will make refactoring of the user info a bit easier if needed.

That being said, no need to change it if you don't want to :)

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -182,4 +187,29 @@ describe(__filename, () => {
     expect(root.find('.AddonMoreInfo-database-id-content'))
       .toHaveText('9001');
   });
+
+  it('links to stats if add-on author is viewing the page', () => {
+    const addon = {
```
kumar303: This should be `addon = createInternalAddon({ ...fakeAddon, slug: ... })`.

Also, while you're in there could you fix the [`render()` ](https://github.com/mozilla/addons-frontend/blob/e3f014c003ca6153da56688abe2658304f3ac30f/tests/unit/amo/components/TestAddonMoreInfo.js#L24) function?

```js
function render(props) {
    return shallowUntilTarget(
      <AddonMoreInfo
        addon={createInternalAddon(fakeAddon)}
        i18n={getFakeI18nInst()}
        store={store}
        {...props}
      />,
      AddonMoreInfoBase
    );
}
```

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -182,4 +187,29 @@ describe(__filename, () => {
...
+      ...fakeAddon,
+      slug: 'coolio',
+      authors: [
+        {
+          id: 11,
```
kumar303: same comment here about extending from `fakeAddon.authors[0]`

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -53,6 +65,31 @@ describe('<AddonMeta>', () => {
       });
       expect(getUserCount(root)).toMatch(/^1\.000/);
     });
+
+    it('links to stats if add-on author is viewing the page', () => {
```
willdurand: Do we have the opposite test case somewhere?

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -182,4 +187,29 @@ describe(__filename, () => {
...
+          name: 'tofumatt',
+          picture_url: 'http://cdn.a.m.o/myphoto.jpg',
+          url: 'http://a.m.o/en-GB/firefox/user/tofumatt/',
+          username: 'tofumatt',
+        },
+      ],
+    };
+    const root = render({
+      addon,
+      store: dispatchSignInActions({ userId: 11 }).store,
```
kumar303: same comment here about referencing a variable for `userId`.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -5,7 +5,12 @@ import React from 'react';
 import AddonMoreInfo, {
   AddonMoreInfoBase,
 } from 'amo/components/AddonMoreInfo';
-import { dispatchClientMetadata, fakeAddon } from 'tests/unit/amo/helpers';
+import Link from 'amo/components/Link';
+import {
+  dispatchSignInActions,
+  dispatchClientMetadata,
```
willdurand: This should be imported first.

tests/unit/core/test_utils.js
```diff
@@ -209,6 +209,48 @@ describe('getClientApp', () => {
...
+  it('returns true when userId is in add-on author object', () => {
+    expect(isAddonAuthor({ addon, userId: 5838591 })).toEqual(true);
+  });
+
+  it('returns true when userId is not in add-on author object', () => {
+    expect(isAddonAuthor({ addon, userId: 5838592 })).toEqual(false);
+  });
+
+  it('returns false when addon is not set', () => {
+    expect(isAddonAuthor({ addon: null, userId: 5838591 })).toEqual(false);
```
kumar303: Since the code relies on falsyness (not just null) then I think you need another test for the `addon: undefined` case. 

Ideally the function should throw if `addon` or `userId` is undefined (since it could have been a calling mistake) but I think React might actually set `addon` to undefined in some cases because JavaScript is awful.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -53,6 +65,31 @@ describe('<AddonMeta>', () => {
...
+      const addon = {
+        ...fakeAddon,
+        slug: 'coolio',
```
willdurand: 🥇

tests/unit/core/test_utils.js
```diff
@@ -209,6 +209,48 @@ describe('getClientApp', () => {
...
+        username: 'tofumatt',
+      },
+    ],
+  };
+
+  it('returns true when userId is in add-on author object', () => {
+    expect(isAddonAuthor({ addon, userId: 5838591 })).toEqual(true);
+  });
+
+  it('returns true when userId is not in add-on author object', () => {
```
willdurand: it returns **false**

tests/unit/core/test_utils.js
```diff
@@ -209,6 +209,48 @@ describe('getClientApp', () => {
...
+    expect(isAddonAuthor({ addon, userId: 5838592 })).toEqual(false);
+  });
+
+  it('returns false when addon is not set', () => {
```
willdurand: when addon is **null**

"not set" could be tested with `addon: undefined` no?

tests/unit/core/test_utils.js
```diff
@@ -209,6 +209,48 @@ describe('getClientApp', () => {
...
+  it('returns false when addon.authors is not set', () => {
+    expect(isAddonAuthor({ addon: {}, userId: 5838591 })).toEqual(false);
+  });
+
+  it('returns false when userId is not set', () => {
+    expect(isAddonAuthor({ addon, userId: null })).toEqual(false);
+  });
+
+  it('returns false if add-on has no authors', () => {
+    const partialAddon = { ...addon, authors: [] };
```
kumar303: I don't see `authors: null` covered which is in the if branch. However, I don't know if that's very realistic.

src/amo/components/AddonMeta.js
kumar303: I guess it's up to the design of the component. If it were `userId?` then `undefined` would be allowed. I think allowing `undefined` is risky and we should avoid it whenever possible. I like using `null` values to indicate that the property value is still loading.

tests/unit/amo/components/TestAddonMeta.js
tofumatt: Should `fakeAddon` just get wrapped in `createInternalAddon()`? I noticed you doing that in another test file a lot and it seems like boilerplate. Maybe we want a helper like `fakeAddon({ ...extraProps })` instead of the object we manipulate now.

tests/unit/core/test_utils.js
tofumatt: It will work, so I added a test case. I'm not sure either but good to test it to make sure.

src/amo/components/AddonMeta.js
tofumatt: Ah, true!

tests/unit/amo/components/TestAddonMeta.js
tofumatt: NIIIIIICE catch 👍  🤣

tests/unit/amo/components/TestAddonMeta.js
tofumatt: Good catch, we didn't. I added one.

tofumatt:
Thanks for the reviews, both. Ready for another r?

src/amo/components/AddonMeta.js
tofumatt: Annoyingly, flow can't understand I won't access that property unless `isAddonAuthor` is `true` (which it never would be if `addon` were `null`), so I left this in.

tofumatt:
This is all rebased and ready for a final r? before the tag if possible 😄

kumar303:
Oh, also, in my patch I changed `Link` to `a` because it will link to an external site (the legacy one), correct? That's how the other definitions were being rendered.

PR #3211, opened 2017-09-22
chore: Ignore nsp issue for JSDom

### LINKED ISSUE #3195: yarn nsp-check is failing re: tough-cookie ReDOS

tofumatt:
See #3195.

As discussed because we don't use `FetchExternalResources` for JSDom it seems we're all good here.

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,0 +1,3 @@
+{
+  "exceptions": ["https://nodesecurity.io/advisories/525"]
+}


kumar303:
I updated the description so that it won't close #3195 . We should close that when we land the upstream patches. Ignoring it is just a workaround for now.

tofumatt:
> I updated the description so that it won't close #3195 . We should close that when we land the upstream patches. Ignoring it is just a workaround for now.

Ah, fair enough!

PR #3217, opened 2017-09-22
Allow links in the disco add-on descriptions

### LINKED ISSUE #3213: Allows links in the summary for recommended add-ons.
Currently recommended add-ons have links removed by DOMPurify:

add-ons_manager

We need to look to allow links in the sanitization step so they don't get removed from the content entirely.

muffinresearch:
Fixes #3213

### DIFF
--- a/tests/unit/disco/components/TestAddon.js
+++ b/tests/unit/disco/components/TestAddon.js
@@ -1,5 +1,6 @@
-import { shallow } from 'enzyme';
 import React from 'react';
+import { oneLine } from 'common-tags';
+import { shallow } from 'enzyme';
 
 import { AddonBase, mapStateToProps } from 'disco/components/Addon';
 import { setInstallState } from 'core/actions/installations';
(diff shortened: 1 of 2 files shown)


tests/unit/disco/components/TestAddon.js
```diff
@@ -245,6 +246,31 @@ describe(__filename, () => {
       );
     });
 
+    it('purifies an editorial description with a bad link', () => {
```
tofumatt: Thanks for testing this! 👍

tests/unit/disco/components/TestAddon.js
```diff
@@ -245,6 +246,31 @@ describe(__filename, () => {
...
+      const data = {
+        ...result,
+        description: 'This is a <a href="javascript:alert(1)">description</a>',
+      };
+      const root = renderAddon({ addon: data, ...data });
+      expect(root.find('.editorial-description').html()).toContain(
```
kumar303: For future reference, @tofumatt discovered that you can do this which is pretty handy:

```js
const anchor = root.find('.editorial-description').render().find('a');
```

tofumatt: Hah, shoot, that's true, I didn't even think of that. I always think to do it myself but forget when reviewing 😅

Would be nice to do that before merging.

muffinresearch: What's the intermediate render for?  Couldn't you do `root.find('.editorial-description a')`

muffinresearch: I'm going to land this for now but I'm not against doing a separate update to fix the entire file for consistency if we're going to make any changes to how these assertions are done once I understand the rationale a bit better. Presumably the idea would be to move to more object representation based assertions?

kumar303: > Couldn't you do `root.find('.editorial-description a')`

It doesn't work, possibly in relation to this [bug](https://github.com/airbnb/enzyme/issues/419).

> I'm not against doing a separate update to fix the entire file for consistency

I personally don't think it's worth a follow-up. Your assertion works fine and will be easy enough to maintain. There are other situations where an HTML based assertion could be more of a maintenance burden.

Issue #2564, opened 2017-06-13
Implement a desktop "all reviews" page


tofumatt:
Implement the review page for desktop, relates to #2552.

tofumatt:
It turns out there isn't a mock for this in the mocks... maybe it was overlooked?

@pwalm Can you confirm there's no desktop version of "all reviews" (after the list that's on the details page)? No worries if not, the mobile will work okay, just want to confirm I'm not missing it in the mocks.

pwalm:
I am missing it, but it won't be hard to throw together. Kind of a hectic week, can I have it to you by Friday?

tofumatt:
Oh gosh yeah that's fine. No rush at all, this Friday or even the next one is fine.

Matthew Riley MacPherson

tofumatt:
Arg, I thought I updated this issue with the image but I must've not clicked submit. Here's the mock from @pwalm:

![all reviews page - 1366px](https://user-images.githubusercontent.com/90871/28961468-5033ec10-78fa-11e7-9bc9-c56ce106ae8a.png)

kumar303:
Is there an API bug on file to get a tally of ratings grouped by star count? We do not currently have this data: http://addons-server.readthedocs.io/en/latest/topics/api/addons.html#get--api-v3-addons-addon-(int-id|string-slug|string-guid)-

kumar303:
whoops, we already have the grouped rating data in the API: http://addons-server.readthedocs.io/en/latest/topics/api/reviews.html#list-reviews

ValentinaPC:
Verified All Review page in AMO-dev FF56(Win 7) and this is how it looks like now:
![2017-10-27_1227](https://user-images.githubusercontent.com/15685960/32097439-4c4c1136-bb12-11e7-9cf5-dfdfeaddd5cd.png)
- no write a review button
- no average rating stars section
- rating stars from each add-on are gray (logged https://github.com/mozilla/addons-frontend/issues/3661)

@tofumatt : Looks like mocks and actual page are very different. Is there an intend to implement those options? Or this is how it should look like? Thanks a lot!

pwalm:
Should look like this:
<img width="979" alt="screen shot 2017-10-27 at 12 26 43 pm" src="https://user-images.githubusercontent.com/6797635/32099804-207f796c-bb12-11e7-901f-ed863e02cef9.png">

ValentinaPC:
But, it looks like this:
[...]

Issue #2775, opened 2017-07-12
Add an interface/popup for developers to reply to reviews


tofumatt:
Add a page that developers can use to reply to a review.

muffinresearch:
I'm not sure but I don't see any UX for this in the sketch file, but I might have missed it? @pwalm do you have something for this?

pwalm:
There is no ux for this on paper, but I have it in my brain. Lemme mock it up and post it here.

kumar303:
I'm nominating for MVP removal because I think we could do this as a follow-up feature.

jvillalobos:
Being able to reply to reviews can be pretty important for developers. I wouldn't object to dropping the All Reviews page from the MVP, but if we keep it this should also be included.

kumar303:
> I wouldn't object to dropping the All Reviews page from the MVP

Just to be clear, we currently do have a mobile listing of all reviews: https://addons.mozilla.org/en-US/firefox/addon/youtube-high-definition/reviews/ (which is reachable on desktop)

The All Reviews issue we have marked as MVP is to give it a desktop style.

kumar303:
@pwalm I'm going to start with the mobile styles since we can use the same overlay as writing/editing a review

kumar303:
Also @tofumatt added a nice inline textarea UX that might be nice here so that the context of the review you are replying to is fully in view.

kumar303:
The inline textarea (added for abuse reports) appears like this when you click a button. This might work well for review replies but I'm going to look at both that and the overlay and start with whichever is easiest to implement. We can always fix it up later.

<img width="321" alt="screenshot 2017-09-26 09 56 03" src="https://user-images.githubusercontent.com/55398/30867290-29b7d26a-a2a1-11e7-8612-27e1ee06e571.png">

pwalm:
First pass at a flow for this:
1) Developer sees a "Reply to this user" below each review.
<img width="1281" alt="screen shot 2017-09-28 at 2 42 21 pm" src="https://user-images.githubusercontent.com/6797635/30991858-c2e517c0-a45b-11e7-9099-f1ea22c37c81.png">

2) Launches a modal window with a reply field.
3) Dev can preview reply.
4) Dev can then publish.
<img width="1296" alt="screen shot 2017-09-28 at 2 42 57 pm" src="https://user-images.githubusercontent.com/6797635/30991900-f06fbf56-a45b-11e7-89fd-2c7768f83bdc.png">

[...]

PR #3220, opened 2017-09-22
Allow users to edit their reviews

### LINKED ISSUE #3180: There is no obvious way to edit your review

* go to AMO and visit any add-ons details page.
* Write a review

### What happened?

You have no obvious way to edit the review

There should be an edit button for your own review

The review is actually editable but you have to click a star as if you were rating the add-on again (which doesn't make sense). See https://github.com/mozilla/addons-frontend/issues/3075 for original report.

kumar303:
Fixes #3180 

I focused mainly on mobile styles only since we will do desktop styles in https://github.com/mozilla/addons-frontend/issues/2564

Here is the new edit flow. A key difference is that the stars are now editable in the overlay.

edit-review-flow mov

(The blue line at the end is just a glitch in how I converted the movie.)


package.json
```diff
@@ -244,7 +244,7 @@
     "intl": "^1.2.5",
     "intl-locales-supported": "^1.0.0",
     "jest": "^21.0.2",
-    "jest-enzyme": "^3.2.0",
+    "jest-enzyme": "^3.8.3",
```
kumar303: This fixed a bug I was seeing in `expect()`

src/amo/components/AddonReview/index.js
```diff
@@ -179,7 +200,7 @@ export class AddonReviewBase extends React.Component {
           <input
             className="AddonReview-submit"
             type="submit"
-            value={i18n.gettext('Submit review')}
+            value={i18n.gettext('Update review')}
```
kumar303: It's pretty much always an update whenever you're on this screen. You can only see this the first time by selecting a star rating beforehand.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+  onClickToEditReview = (event: SyntheticEvent) => {
+    event.preventDefault();
+    this.setState({ editingReview: true });
+  }
+
+  onEscapeReviewOverlay = () => {
+    // Even though an escaped overlay will be hidden, we still have to
+    // synchronize our show/hide state otherwise we won't be able to
+    // show the overlay after it has been escaped.
```
kumar303: This is a bit lame since we have a chain of callbacks now. I explained the issue in https://github.com/mozilla/addons-frontend/issues/3222 but I think using callbacks is a quicker fix for now.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -1,122 +1,143 @@
...
+    ...customProps
+  } = {}) => {
+    const props = {
+      i18n: getFakeI18nInst(),
+      location: { query: {} },
+      params,
+      store,
+      ...customProps,
+    };
+    return shallowUntilTarget(
```
kumar303: Believe it or not I didn't want to spend time converting this whole suite to shallow rendering. However, since I had to refactor the `reviews` state *and* be sure I didn't break things along the way, it was really helpful to switch to shallow rendering first. It turned out to be a nice validation of this `shallowUntilTarget` pattern because you can see how this component's reliance on mapped state is fully tested now.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -0,0 +1,121 @@
...
+                onReviewSubmitted={this.onReviewSubmitted}
+                review={review}
+              />
+              : null
+            }
+            <Button
+              onClick={this.onClickToEditReview}
+              className="AddonReviewListItem-edit-button Button--action Button--small"
+            >
+              {i18n.gettext('Edit')}
```
tofumatt: This button could definitely use some more context. I also don't really think it should be a button. I think stylistically looking like a link with a large tap-target would be better. Where there are multiple buttons it might look weird and things like "Report this review" will be link-styled, not button-styled. So I think this should be "Edit my review" and should be an `<a>` tag instead with link styling.

Changing this to a `<a>` tag should be fine, I think the only adjustment you'd need would be a `preventDefault()` for the `onClick`...

src/amo/components/AddonReviewList/index.js
```diff
@@ -55,6 +52,8 @@ export class AddonReviewListBase extends React.Component {
   }
 
   loadDataIfNeeded(nextProps?: AddonReviewListProps) {
+    const lastAddon = this.props.addon;
```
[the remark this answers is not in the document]
kumar303: It only seems necessary to me if you need to destruct multiple variables at once. Otherwise, what's the point?

tests/unit/amo/components/TestAddonReview.js
```diff
@@ -29,37 +35,55 @@ function fakeLocalState(overrides = {}) {
...
-    refreshAddon: () => Promise.resolve(),
-    review: defaultReview,
-    setDenormalizedReview: () => {},
-    updateReviewText: () => Promise.resolve(),
-    ...customProps,
+describe(__filename, () => {
+  let store;
+
+  beforeEach(() => {
+    store = createStore().store;
```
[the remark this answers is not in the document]
kumar303: This component doesn't use `clientApp` or `api` or anything which is why I didn't use it. I suppose I could just so someone doesn't copy/paste it and get confused.

src/amo/components/AddonReview/index.js
kumar303: There's not an easy way. I would have to make the internal state of `RatingManager` more complex and that test suite doesn't test the click-star-then-edit flow in its entirety (due to react-test-utils limitations).

I'll just revert it back to `Submit review` for now since that makes sense in either case.

kumar303:
@tofumatt thanks, it should be ready for another review. I gave the Edit anchor tag `href="#"` because I'm not sure how else to do that. I don't know if it's worth the effort of making a button style that looks like a link.

src/amo/components/AddonReviewList/index.js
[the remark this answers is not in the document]
tofumatt: Fair enough.

src/amo/components/AddonReviewList/index.js
```diff
@@ -204,7 +173,11 @@ export class AddonReviewListBase extends React.Component {
         <CardList>
           <ul>
             {allReviews.map((review, index) => {
-              return this.renderReview({ review, key: String(index) });
+              return (
+                <li className="AddonReviewList-li" key={String(index)}>
```
tofumatt: I'm not sure we use this class anywhere anymore (based on its removal from the stylesheet below) and it's a bit awkwardly named anyway. I'd be fine just removing it if it isn't used anywhere.

src/amo/components/AddonReviewListItem/styles.scss
```diff
@@ -0,0 +1,41 @@
+@import "~amo/css/inc/vars";
+
+.AddonReviewListItem {
+  word-wrap: break-word;
+
+  > h3 {
```
tofumatt: I know these are from a previous patch but if it'd be possible to label this with a `className` and select them directly instead that'd be a bit less fragile. I've liked us not nesting selectors, especially direct selectors like this. I know it's way more verbose but it should help with HTML/CSS refactors.

If not though, it's okay.

kumar303:
I fixed the focus and looked into the other fixes but they weren't straight forward. I think on desktop we can do away with these modals. They were originally designed for mobile so clicking on the sides isn't really an issue (the gutter is pretty small). The modals might still be good UX for mobile screens.

I'll land this and then look at replacing the modal for desktop edits as part of https://github.com/mozilla/addons-frontend/issues/2775

The failing tests will be resolved on master soon by https://github.com/mozilla/addons-frontend/pull/3244

tofumatt:
> I think on desktop we can do away with these modals. They were originally designed for mobile so clicking on the sides isn't really an issue (the gutter is pretty small). The modals might still be good UX for mobile screens.

All sounds good 👍 Thanks!

Issue #3229, opened 2017-09-25
Link to All versions / All development channel versions from add-on detail page


kumar303:
Before implementing the [All versions / All development channel versions](https://github.com/mozilla/addons-frontend/issues/2770) pages we can link to those existing pages on the old site. This will at least let users (e.g. ESR or pre-57 users) attempt to find compatible versions.

vcarciu:
Verified as fixed : 
![see versions](https://user-images.githubusercontent.com/1583842/30873628-8968c3d6-a2f6-11e7-95f2-60d6e98704b0.gif)

PR #3230, opened 2017-09-25
Link to Version History / Development Channel

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3229

This adds two external links to the More Info card for viewing and installing alternative versions.

screenshot 2017-09-25 16 49 33

I also made the loading screen match the final rendered card more closely which should make the loading transition less jarring.

screenshot 2017-09-25 12 55 39

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -27,7 +27,6 @@ export class AddonMoreInfoBase extends React.Component {
         versionLicense: <LoadingText minWidth={20} />,
         addonId: <LoadingText minWidth={20} />,
         versionHistoryLink: <LoadingText minWidth={20} />,
-        betaVersionsLink: <LoadingText minWidth={20} />,
       });
     }
 
@@ -159,10 +158,12 @@ export class AddonMoreInfoBase extends React.Component {
           {i18n.gettext('Version History')}
         </dt>
         <dd>{versionHistoryLink}</dd>
-        <dt className="AddonMoreInfo-beta-versions-title">
-          {i18n.gettext('Beta Versions')}
-        </dt>
-        <dd>{betaVersionsLink}</dd>
+        {betaVersionsLink ? (
+          <dt className="AddonMoreInfo-beta-versions-title">
+            {i18n.gettext('Beta Versions')}
+          </dt>
+        ) : null}
+        {betaVersionsLink ? <dd>{betaVersionsLink}</dd> : null}
         <dt
           className="AddonMoreInfo-database-id-title"
           title={i18n.gettext(`This ID is useful for debugging and
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -132,9 +132,13 @@ export class AddonMoreInfoBase extends React.Component {
             </ul>
           </dd>
         ) : null}
-        <dt>{i18n.gettext('Version')}</dt>
+        <dt className="AddonMoreInfo-version-title">
+          {i18n.gettext('Version')}
+        </dt>
         <dd className="AddonMoreInfo-version">{version}</dd>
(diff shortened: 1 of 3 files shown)


tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -161,17 +168,8 @@ describe(__filename, () => {
       .toHaveProp('href', '/addon/chill-out/eula/');
   });
 
-  it('does not render an add-on ID if none exists', () => {
```
kumar303: I can't think of any situation where this would happen

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -57,6 +49,75 @@ export class AddonMoreInfoBase extends React.Component {
...
+        >
+          {i18n.gettext('See all versions')}
+        </a>
+      ),
+      versionDevChannelLink: (
+        <a
+          className="AddonMoreInfo-dev-channel-link"
+          href={`/addon/${addon.slug}/versions/beta`}
+        >
+          {i18n.gettext('See experimental versions')}
```
tofumatt: I've mentioned this elsewhere but I think on the new front-end we should be referring to these as "Beta versions" not "Development" or "Experimental". It's a confusing name for users and the intent of these versions is bleeding edge features/fixes for adventurous users–exactly what a beta is 😄

So I think this could be:

```html
<dt>Version History</dt>
<dd>See all versions</dd>
<dt>Beta Version</dt>
<dd>See all beta versions</dd>
```

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -73,79 +134,45 @@ export class AddonMoreInfoBase extends React.Component {
...
+        </dt>
+        <dd className="AddonMoreInfo-database-id-content">
+          {addonId}
+        </dd>
+        <dt className="AddonMoreInfo-version-history-title">
+          {i18n.gettext('Version History')}
+        </dt>
+        <dd>{versionHistoryLink}</dd>
+        <dt className="AddonMoreInfo-dev-channel-title">
+          {i18n.gettext('Development Channel')}
```
tofumatt: I like the less jarring transition, but for the Beta Channel you might need to do a check to make sure `current_beta_version` exists. As I understand it not all add-ons have/require a beta version, so it'd be weird if this header appeared with the loading content then went away.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -73,79 +134,45 @@ export class AddonMoreInfoBase extends React.Component {
...
+        </dt>
+        <dd>{versionDevChannelLink}</dd>
```
tofumatt: I'd rename the referenced to "beta" instead of "development" or "experimental" as well.

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: Oh, that's true. Dunno what I was thinking. 🙈

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -182,4 +180,24 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({ ...fakeAddon, slug: 'some-slug' });
+    const root = render({ addon });
+
+    expect(root.find('.AddonMoreInfo-version-history-title'))
+      .toHaveLength(1);
+    const link = root.find('.AddonMoreInfo-version-history-link');
+    expect(link).toHaveProp('href', `/addon/${addon.slug}/versions/`);
+  });
+
+  it('links to development channel versions', () => {
```
tofumatt: As mentioned, I think this should check to make sure one exists and render nothing if not.

Also we should call it beta. 😄

kumar303:
Good ideas, thanks. I always found *development channel* to be a bit non-intuitive.

@tofumatt it should be ready for another review.

PR #3242, opened 2017-09-26
fix: Put cards in proper order

### LINKED ISSUE #3241: Addon Detail Page Card order doesn't match mocks
The card order should look like:

screenshot 2017-09-26 13 28 52
screenshot 2017-09-26 13 28 46

I've got a quick patch to change this.

QA: Note this won't update all the styles, but will at least keep the order presented here.

tofumatt:
fix #3241 

Now looks like this, which matches the order in the mocks:

screenshot 2017-09-26 13 52 07

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -202,6 +202,7 @@
       // This span makes sure the left widget does not move vertically when
       // there is a long description.
       grid-row: 1 / span 100000;
+      overflow-x: hidden;
     }
 
     .AddonDescription-version-notes {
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/styles.scss
```diff
@@ -202,6 +202,7 @@
       // This span makes sure the left widget does not move vertically when
       // there is a long description.
       grid-row: 1 / span 100000;
+      overflow-x: hidden;
```
willdurand: Is this needed for the same as line 222?

tofumatt: No, this isn’t Safari-related. It happens on Firefox.

Issue #3243, opened 2017-09-26
Prevent denial of service in Express server


kumar303:
There is a patch release fix in Express for https://nodesecurity.io/advisories/526 so let's land it.

PR #3244, opened 2017-09-26
Upgrade Express to prevent denial of service

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3243


kumar303:
The `yarn.lock` diff shows that we get the fix (which is in `fresh` 0.5.2). We still have `fresh` 0.3.0 getting installed but I verified that this is only for `mock-express-request`.

Issue #3140, opened 2017-09-14
Collection view


muffinresearch:
Refs: https://github.com/mozilla/addons-frontend/issues/2782

Mocks are in the Sketch file:

<img width="710" alt="amo_desktop_-_master_sketch" src="https://user-images.githubusercontent.com/1514/30435430-a78733b4-9961-11e7-9d55-6077548b6293.png">

muffinresearch:
Iconography is likely to need to dropped for the first release.

muffinresearch:
Otherwise the page to show a collection is fairly vital for representing curated content so should stay in the MVP.

muffinresearch:
This should reflect themes as well as extensions since a collection can contain both.

pwalm:
All three versions:
![extensions collection - 1366px](https://user-images.githubusercontent.com/6797635/30937512-e0a152ea-a38b-11e7-8cf9-0bc713c30b15.png)

![themes collection - 1366px](https://user-images.githubusercontent.com/6797635/30937519-e55fb196-a38b-11e7-9c1c-539f4e125624.png)

![collection mixed - 1366px](https://user-images.githubusercontent.com/6797635/30937525-e95f1c50-a38b-11e7-8816-367c3b36a8d2.png)

tofumatt:
@pwalm: I have a general question about the ways extensions/themes are split up here (and on other pages)–what happens if there's only one theme in the list? Or if the search results surface themes that are more relevant for a search? Should we always de-emphasize themes?

For collections this is a bit less of an issue but for searches the API returns results largely based on relevancy–would it be possible to mock something up that mixed the two results together a bit more? It means fewer API requests (site is faster) but would also mean sorting by relevancy is a lot more straightforward/accurate.

devaneymoz:
From an Editorial perspective, it's not an MVP to have both themes and extensions within the same collection. 

Also, I didn't think Search returns mixed these add-on types?

tofumatt:
Search results can return mixed result types if no “add-on type” filter is set. Right now we just render them each on one row.

willdurand:
Quick question about these screens: where do we fetch the images displayed on the left side? I don't think it is a blocker for MVP, but I am curious.

ValentinaPC:
Is this on -dev? Because, to me, everything looks the same. I am missing something? Thanks!

kumar303:
[...]

PR #3256, opened 2017-09-27
Add reducer/saga/api for collection view

willdurand:
Refs #3140

---

This PR adds the reducer, saga, and API needed to build the collection view. It has a simplified state that could be extended later but for now, I focus on getting this collection view up and running as part of our MVP.

I do not reset the whole `current` collection when fetching a collection because it is nicer in the UI: only the list of add-ons has a "loading indicator" but the card with the detail on the left side do not change. I can't think of any issue with this so far, since we do not list collections in a list view. See:

I could maybe add a separate action for that, but I took the easy path.

### DIFF
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -20,13 +20,16 @@ export const getCollectionDetail = (
   }
 
   return callApi({
-    auth: Boolean(api && api.token),
+    auth: true,
     endpoint: `accounts/account/${user}/collections/${slug}`,
     state: api,
   });
 };
 
-type GetCollectionAddonsParams = GetCollectionParams & {| page?: number |};
+type GetCollectionAddonsParams = {|
+  ...GetCollectionParams,
+  page?: number,
+|};
 
 export const getCollectionAddons = (
   { api, page, slug, user }: GetCollectionAddonsParams
@@ -39,7 +42,7 @@ export const getCollectionAddons = (
   }
 
   return callApi({
-    auth: Boolean(api && api.token),
+    auth: true,
     endpoint: `accounts/account/${user}/collections/${slug}/addons`,
     params: { page },
     state: api,
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -1,0 +1,47 @@
+/* @flow */
+import { callApi } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+
+
+type GetCollectionParams = {|
+  api: ApiStateType,
+  slug: string,
+  user: string | number,
+|};
+
+export const getCollectionDetail = (
+  { api, slug, user }: GetCollectionParams
+) => {
+  if (!slug) {
+    throw new Error('slug is required');
+  }
+  if (!user) {
+    throw new Error('user is required');
+  }
+
+  return callApi({
+    auth: Boolean(api && api.token),
+    endpoint: `accounts/account/${user}/collections/${slug}`,
+    state: api,
+  });
+};
+
+type GetCollectionAddonsParams = GetCollectionParams & {| page?: number |};
+
+export const getCollectionAddons = (
+  { api, page, slug, user }: GetCollectionAddonsParams
+) => {
+  if (!slug) {
+    throw new Error('slug is required');
+  }
+  if (!user) {
+    throw new Error('user is required');
+  }
+
+  return callApi({
+    auth: Boolean(api && api.token),
+    endpoint: `accounts/account/${user}/collections/${slug}/addons`,
+    params: { page },
+    state: api,
+  });
+};
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -37,7 +37,7 @@ type FetchCollectionParams = {|
 |};
 
 type FetchCollectionAction = {|
-  type: 'FETCH_COLLECTION',
+  type: typeof FETCH_COLLECTION,
   payload: FetchCollectionParams,
 |};
 
@@ -69,7 +69,7 @@ type FetchCollectionPageParams = {|
 |};
 
 type FetchCollectionPageAction = {|
-  type: 'FETCH_COLLECTION_PAGE',
+  type: typeof FETCH_COLLECTION_PAGE,
   payload: FetchCollectionPageParams,
 |};
 
@@ -136,7 +136,7 @@ type LoadCollectionParams = {|
 |};
 
 type LoadCollectionAction = {|
-  type: 'LOAD_COLLECTION',
+  type: typeof LOAD_COLLECTION,
   payload: LoadCollectionParams,
 |};
 
@@ -162,7 +162,7 @@ type LoadCollectionPageParams = {|
 |};
 
 type LoadCollectionPageAction = {|
-  type: 'LOAD_COLLECTION_PAGE',
+  type: typeof LOAD_COLLECTION_PAGE,
   payload: LoadCollectionPageParams,
 |};
 
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -2,8 +2,12 @@
 import { createInternalAddon } from 'core/reducers/addons';
 import type { AddonType, ExternalAddonType } from 'core/types/addons';
 
-export const FETCH_COLLECTION = 'FETCH_COLLECTION';
-export const LOAD_COLLECTION = 'LOAD_COLLECTION';
+export const FETCH_COLLECTION: 'FETCH_COLLECTION' = 'FETCH_COLLECTION';
+export const LOAD_COLLECTION: 'LOAD_COLLECTION' = 'LOAD_COLLECTION';
+export const FETCH_COLLECTION_PAGE: 'FETCH_COLLECTION_PAGE'
+  = 'FETCH_COLLECTION_PAGE';
+export const LOAD_COLLECTION_PAGE: 'LOAD_COLLECTION_PAGE'
+  = 'LOAD_COLLECTION_PAGE';
 
 export type CollectionType = {
   addons: Array<AddonType>,
@@ -32,12 +36,17 @@ type FetchCollectionParams = {|
   user: number | string,
 |};
 
+type FetchCollectionAction = {|
+  type: 'FETCH_COLLECTION',
+  payload: FetchCollectionParams,
+|};
+
 export const fetchCollection = ({
   errorHandlerId,
   page,
   slug,
   user,
-}: FetchCollectionParams = {}) => {
+}: FetchCollectionParams = {}): FetchCollectionAction => {
   if (!errorHandlerId) {
     throw new Error('errorHandlerId is required');
   }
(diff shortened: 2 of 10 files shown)


src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
+/* @flow */
+import { createInternalAddon } from 'core/reducers/addons';
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
+
+export const FETCH_COLLECTION = 'FETCH_COLLECTION';
+export const LOAD_COLLECTION = 'LOAD_COLLECTION';
+
+export type CollectionType = {
```
willdurand: I did not put strict here because Flow complains in the reducer (when I spread the state..)

willdurand:
Ready for a first review. I assigned both but feel free to unassign the other :)

src/amo/reducers/collections.js
kumar303: > I did not put strict here because Flow complains in the reducer (when I spread the state..)

Ugh, yeah, I still have that issue in my awesomebar history https://github.com/facebook/flow/issues/2405

kumar303: You could possibly keep this exact and then import `$Shape` from `src/core/types/util.js` and use `$Shape<CollectionType>` only where you need it for spreads. If that's a hassle then it's no biggie.

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+  };
+};
+
+type CollectionAddonsResults = Array<{|
+  addon: ExternalAddonType,
+  downloads: number,
+  notes: string | null,
+|}>;
+
+// The API returns more attributes than what is listed below.
```
kumar303: I think you should list them all in the type

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+type CollectionDetail = {
+  addon_count: number,
+  author: {
```
kumar303: this type should be exact if possible

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+    name: string,
+  },
+  description: string | null,
+  id: number,
+  modified: string,
+  name: string,
+};
+
+type LoadCollectionParams = {|
+  addons: {
```
kumar303: this type should be exact if possible

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+type CollectionAddonsResults = Array<{|
+  addon: ExternalAddonType,
+  downloads: number,
+  notes: string | null,
+|}>;
+
+// The API returns more attributes than what is listed below.
+type CollectionDetail = {
```
kumar303: If I'm understanding correctly, this is the API response that eventually becomes `CollectionType`, correct? If so, you should name this one `ExternalCollectionType` to make it clear.

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+  detail: CollectionDetail,
+  items: CollectionAddonsResults,
+|};
+
+export const createInternalCollection = ({
+  detail,
+  items,
+}: CreateInternalCollectionParams): CollectionType => ({
+  addons: items.map((item) => {
+    // This allows to have a consistent way to manipulate addons in the app.
```
kumar303: Good idea

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+  slug: string,
+  user: number | string,
+|};
+
+export const fetchCollection = ({
+  errorHandlerId,
+  page,
+  slug,
+  user,
+}: FetchCollectionParams = {}) => {
```
kumar303: This should include a type for the returned action object.

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+  description: detail.description,
+  id: detail.id,
+  lastUpdatedDate: detail.modified,
+  name: detail.name,
+  numberOfAddons: detail.addon_count,
+});
+
+const reducer = (
+  state: CollectionsState = initialState,
+  action: Object
```
kumar303: I think it would be helpful to add a type for the action objects that you will be operating on. Due to how Flow comprehends switch statements, you will need the constant-to-literal hack to make that work:

```js
export const FETCH_COLLECTION: 'FETCH_COLLECTION' = 'FETCH_COLLECTION';
```

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,160 @@
...
+): CollectionsState => {
+  switch (action.type) {
+    case FETCH_COLLECTION:
+      return {
+        // Current collection can be null if state is the initial state,
+        // otherwise we have already loaded a collection. We cannot load
+        // another collection from the collection view, so we can keep the
+        // detail of the current collection and only use "loading indicator"
+        // for the add-ons.
+        current: state.current === null ? null : {
```
kumar303: Why not just always reset it like `current: null` while fetching?

src/amo/sagas/collections.js
```diff
@@ -0,0 +1,48 @@
...
+}) {
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
+
+    const { detail, addons } = yield all({
+      detail: call(api.getCollectionDetail, {
```
kumar303: It seems unnecessary to make two API calls. Could you file an issue requesting to add a `collection` parameter to the response of https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-list ?

src/amo/sagas/collections.js
```diff
@@ -0,0 +1,48 @@
...
+
+    yield put(loadCollection({ addons, detail }));
+  } catch (error) {
+    log.warn(`Collection failed to load: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
+  }
+}
+
+export default function* collectionsSaga() {
+  yield takeEvery(FETCH_COLLECTION, fetchCollection);
```
kumar303: I think we should default to `takeLatest` unless we know that we need to process parallel requests.

tests/unit/amo/api/test_collections.js
```diff
@@ -0,0 +1,159 @@
...
+  dispatchClientMetadata,
+  dispatchSignInActions,
+} from 'tests/unit/amo/helpers';
+
+
+describe(__filename, () => {
+  let mockApi;
+
+  const getParams = ({
+    apiState = dispatchClientMetadata().store.getState().api,
```
kumar303: Since you reference in a few places where it's not relevant to the test, you might as well initialize it in `beforeEach()`

tests/unit/amo/api/test_collections.js
```diff
@@ -0,0 +1,159 @@
...
+        .withArgs({
+          auth: false,
+          endpoint: 'accounts/account/user-id-or-name/collections/some-slug',
+          state: apiState,
+        })
+        .once()
+        .returns(createApiResponse());
+
+      return getCollectionDetail(params)
+        .then(() => {
```
kumar303: For all new tests, let's use `async/await`.

tests/unit/amo/api/test_collections.js
```diff
@@ -0,0 +1,159 @@
...
+    it('makes an authenticated request when API state allows it', () => {
+      const apiState = dispatchSignInActions().store.getState().api;
+      const params = getParams({ apiState });
+
+      mockApi
+        .expects('callApi')
+        .withArgs({
+          auth: true,
+          endpoint:
+          'accounts/account/user-id-or-name/collections/some-slug',
```
kumar303: I think it helps for readability to indent continuations like this but we don't have a lint rule so meh

tests/unit/amo/api/test_collections.js
```diff
@@ -0,0 +1,159 @@
...
+        .once()
+        .returns(createApiResponse());
+
+      return getCollectionDetail(params)
+        .then(() => {
+          mockApi.verify();
+        });
+    });
+
+    it('makes an authenticated request when API state allows it', () => {
```
kumar303: You can delete this test when `auth` is always `true`

tests/unit/amo/api/test_collections.js
```diff
@@ -0,0 +1,159 @@
...
+        .once()
+        .returns(createApiResponse());
+
+      return getCollectionAddons(params)
+        .then(() => {
+          mockApi.verify();
+        });
+    });
+
+    it('makes an authenticated request when API state allows it', () => {
```
kumar303: You can delete this test too after `auth` is always `true`.

src/amo/sagas/collections.js
```diff
@@ -0,0 +1,48 @@
...
+    slug,
+    user,
+  },
+}) {
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
```
kumar303: Does this need `yield put(showLoading())` ?

src/amo/api/collections.js
```diff
@@ -0,0 +1,47 @@
...
+) => {
+  if (!slug) {
+    throw new Error('slug is required');
+  }
+  if (!user) {
+    throw new Error('user is required');
+  }
+
+  return callApi({
+    auth: Boolean(api && api.token),
```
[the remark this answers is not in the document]
willdurand: Oh that's what this field means.. I did not know actually. Thanks!

src/amo/api/collections.js
```diff
@@ -0,0 +1,47 @@
...
+    endpoint: `accounts/account/${user}/collections/${slug}`,
+    state: api,
+  });
+};
+
+type GetCollectionAddonsParams = GetCollectionParams & {| page?: number |};
```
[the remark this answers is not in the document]
willdurand: Ah well, if this works, sure it is 200% better.

src/amo/reducers/collections.js
willdurand: Ah right, I guess I was lazy and omitted all the Action types because there were a lot of types already.

src/amo/reducers/collections.js
willdurand: Ok, I can do that. I wanted to keep things small.

src/amo/reducers/collections.js
willdurand: Not exactly, the `CollectionType` is the representation I need in the app, but it merges different result sets because there are 2 endpoints to call.

src/amo/reducers/collections.js
willdurand: Yep, as you spotted before, I did not handle the Action types at all..

src/amo/reducers/collections.js
willdurand: Because I want to keep the collection information on the left side (see the Gif in the PR description). Resetting it would force me to add loading indicators for this card (on the left), but I don't really want to do that because it feels weird (as a user).

src/amo/sagas/collections.js
willdurand: Ah yes, maybe indeed.

src/amo/sagas/collections.js
willdurand: Ah yes, I already talked to @eviljeff but did not open the issue yet. Will do!

tests/unit/amo/api/test_collections.js
willdurand: Yes, and I think I edited the file and reformatted it entirely. Unfortunately, my vim does not want to keep such indentations 😢 

I'll fix this!

src/amo/reducers/collections.js
kumar303: Aha, I think I understand what's going on. I suggest that you do it in two separate actions. The first would be `FETCH_COLLECTION` which fetches an entirely new collection (which would happen during navigation) -- this would always reset `current` to `null`. The second would be `FETCH_COLLECTION_PAGE` which would only happen during pagination -- this would simply reset `addons` to `[]`. 

This would solve the issue where loading an entirely new collection will leave information up about the old collection. However, if this introduces too much complexity then I guess it would be fine as is.

willdurand: I decided to do this, it takes some time though..

willdurand:
Ok updated Flow types, added Action types and added new actions to fetch and load collection add-ons pages.

src/amo/sagas/collections.js
willdurand: I did not add this as discussed on IRC.

src/amo/sagas/collections.js
willdurand: See: https://github.com/mozilla/addons-server/issues/6548.

src/amo/reducers/collections.js
```diff
@@ -0,0 +1,265 @@
...
+
+type FetchCollectionParams = {|
+  errorHandlerId: string,
+  page?: number,
+  slug: string,
+  user: number | string,
+|};
+
+type FetchCollectionAction = {|
+  type: 'FETCH_COLLECTION',
```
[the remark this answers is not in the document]
willdurand: ah right, sorry for that.

PR #3263, opened 2017-09-28
chore: Add error handler output to ReportAbuseButton

### LINKED ISSUE #3262: Add-on abuse report UI does not show errors

* Using a local development server, go to an add-on detail page
* Click Report this add-on for abuse
* Enter some text for the abuse report
* Stop your development server. This will cause the API to respond with a 500 error. I'm not sure of any other way to trigger an API error.
* Click the Send abuse report button

### What happened?

Nothing. The UI is stuck in a state where you can't dismiss the prompt:

screenshot 2017-09-27 15 06 47

The UI should render the error caught by `errorHandler`

tofumatt:
Fix #3262

Add error handler to the report abuse button.

Looks like this:

screenshot 2017-09-28 00 57 25

### DIFF
--- a/tests/unit/amo/components/TestReportAbuseButton.js
+++ b/tests/unit/amo/components/TestReportAbuseButton.js
@@ -1,48 +1,59 @@
 import { mount } from 'enzyme';
 import React from 'react';
 
-import ReportAbuseButton from 'amo/components/ReportAbuseButton';
+import ReportAbuseButton, {
+  ReportAbuseButtonBase,
+} from 'amo/components/ReportAbuseButton';
+import { setError } from 'core/actions/errors';
+import I18nProvider from 'core/i18n/Provider';
 import {
   disableAbuseButtonUI,
   enableAbuseButtonUI,
   loadAddonAbuseReport,
   sendAddonAbuseReport,
 } from 'core/reducers/abuse';
+import ErrorList from 'ui/components/ErrorList';
 import { dispatchClientMetadata, fakeAddon } from 'tests/unit/amo/helpers';
 import {
   createFakeAddonAbuseReport,
   createFakeEvent,
   createStubErrorHandler,
   getFakeI18nInst,
+  shallowUntilTarget,
 } from 'tests/unit/helpers';
 
 
 describe(__filename, () => {
-  function renderMount({
-    addon = { ...fakeAddon, slug: 'my-addon' },
-    errorHandler = createStubErrorHandler(),
-    store = dispatchClientMetadata().store,
-    ...props
-  } = {}) {
+  const defaultRenderProps = {
+    addon: { ...fakeAddon, slug: 'my-addon' },
+    errorHandler: createStubErrorHandler(),
+    i18n: getFakeI18nInst(),
+    store: dispatchClientMetadata().store,
+  };
+
+  function renderMount({ ...props } = {}) {
     return mount(
-      <ReportAbuseButton
-        addon={addon}
-        errorHandler={errorHandler}
-        i18n={getFakeI18nInst()}
-        store={store}
-        {...props}
-      />
+      <I18nProvider i18n={props.i18n || defaultRenderProps.i18n}>
+        <ReportAbuseButton {...defaultRenderProps} {...props} />
+      </I18nProvider>
+    );
+  }
+
+  function renderShallow({ ...props } = {}) {
+    return shallowUntilTarget(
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestReportAbuseButton.js
```diff
@@ -304,4 +308,16 @@ describe(__filename, () => {
...
   });
+
+  it('renders an error if one exists', () => {
+    const errorHandler = createStubErrorHandler();
+    const { store } = dispatchClientMetadata();
+    store.dispatch(setError({
+      error: new Error('something bad'),
+      id: errorHandler.id,
+    }));
+    const root = renderMount({ errorHandler, store });
```
willdurand: I am not sure if it is relevant here, but in many other places we did this instead:

https://github.com/mozilla/addons-frontend/blob/5a33c7458ef20c33a27cbed1fb73d0bf2577715a/tests/unit/amo/components/TestAddonReviewList.js#L289-L299

tests/unit/amo/components/TestReportAbuseButton.js
```diff
@@ -25,13 +27,15 @@ describe(__filename, () => {
...
-        errorHandler={errorHandler}
-        i18n={getFakeI18nInst()}
-        store={store}
-        {...props}
-      />
+      <I18nProvider i18n={getFakeI18nInst()}>
+        <ReportAbuseButton
+          addon={addon}
+          errorHandler={errorHandler}
+          i18n={getFakeI18nInst()}
```
willdurand: We could maybe extract a variable here since `getFakeI18nInst()` is called twice.

tests/unit/amo/components/TestReportAbuseButton.js
tofumatt: I saw that but it seems a bit more verbose way to do it... the error is still being dispatched and clearly showing up in the component in my test path, so I'm not sure what the extra tests gain.

tests/unit/amo/components/TestReportAbuseButton.js
```diff
@@ -304,4 +309,16 @@ describe(__filename, () => {
...
   });
+
+  it('renders an error if one exists', () => {
+    const errorHandler = createStubErrorHandler();
+    const { store } = dispatchClientMetadata();
+    store.dispatch(setError({
+      error: new Error('something bad'),
+      id: errorHandler.id,
+    }));
+    const root = renderMount({ errorHandler, store });
```
kumar303: I don't see any advantage to doing a mounted render and it has several disadvantages for future maintenance. I would like to see a shallow render where you just check that the component exists:

```js
```

tests/unit/amo/components/TestReportAbuseButton.js
kumar303: I think it's ok as is because it dispatches `setError()` with a real store. 

If I recall correctly, the `AddonReview` test suite couldn't use this approach because its render method wasn't triggering Redux state mapping like it should have. It was rendering a base component directly.

tests/unit/amo/components/TestReportAbuseButton.js
kumar303: Hmm, this was my only change request but github doesn't show it in the context of my review for some reason.

tofumatt:
Good point, I didn't need to mount it. A lot of the rest of the test suite uses mount because of the UI interaction and stuff... all fixed up.

kumar303:
> the rest of the test suite uses mount because of the UI interaction

I didn't get to review the original patch (lucky for you!) but I don't see anything in the suite that won't work with shallow rendering. You can replace this:

```js
const textarea = root.find('.ReportAbuseButton-textarea textarea');
textarea.node.value = 'Opera did it first!';
textarea.simulate('change');
```

with:

```js
const textarea = root.find('.ReportAbuseButton-textarea textarea');
textarea.simulate('change', createFakeEvent({
  target: { value: 'Opera did it first!' }
}));
```

And instead of testing the implementation of `<Textarea />` you should just check that the component gets instantiated with the right parameters. This tests the *interface* of the component which is something that the developer will likely make a good effort to preserve.

There's no good reason to fix this up now but just take note for future patches. We should always avoid `mount()` whenever possible.

tofumatt:
Actually, I think it was that the component originally used `setState` a lot. I changed it over to use redux but might’ve not updated the tests as they would have still worked :-(

Issue #3100, opened 2017-09-12
Pagination doesn't match the mocks


muffinresearch:
The mocks have a new pagination: 

<img width="621" alt="amo_desktop_-_master_sketch" src="https://user-images.githubusercontent.com/1514/30333663-5b19bda4-97d5-11e7-907e-f1bc37fa041a.png">

Site currently has: 

<img width="1517" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30333594-2dac1196-97d5-11e7-8267-28bbbbf8ab4d.png">

This could be a post-release fix-up or something to focus on at the end if we have time.

tsl143:
Can I pick this one?

tofumatt:
Yes, feel free to work on this. Let us know if you are and we'll mark it `contrib:assigned`, thanks.

tofumatt:
The touch targets here I'm suspicious of actually so this will have different versions for desktop/mobile. Eg. mobile should still have large(r) buttons, maybe fewer pages with more padding.

tsl143:
Great, I am picking this up.

tsl143:
Great, I am picking this up.

tofumatt:
Hi @tsl143, are you still working on this one or are you focusing on other issues? No worries either way, but I'll unassign this if you're busy with other stuff–I know you've been closing a tonne of issues lately so no worries if you don't have time for this one! 😄

tsl143:
Hi @tofumatt , thanks :)
I started on this, but got occupied with something, I may not be working on this for now. Please feel free to unassign this.

willdurand:
Unassigned you @tsl143, thanks for all your contributions btw!

willdurand:
I need this for the user profile page.

willdurand:
For QA, the new paginator still looks a bit different (please see #5336) but here are the things to test:

- make sure there are more pages displayed by default
- make sure that links work
- on small screens, the paginator should look different (i.e. only prev/next buttons as before)

vcarciu:
Verified as fixed : 
![pagination](https://user-images.githubusercontent.com/1583842/41716471-35a7e10a-755f-11e8-80ba-96d86dbaa627.gif)

Also verified for small screens and everything works as expected.

PR #3268, opened 2017-09-28
Collection view

### LINKED ISSUE #3140: Collection view

willdurand:
Fix #3140 
~~⚠️ depends on #3256~~

---

This PR adds a new page: collection view. I reuse the `AddonsCard` component that displays a list of add-ons. It does not match the mock-ups yet because this component needs some love but I believe this should be done in a separate PR. Same for the pagination (see #3100).

## Screenshots

Loading:

screen shot 2017-09-29 at 15 59 55

Pagination:

screen shot 2017-09-28 at 13 58 32

Extensions:

screen shot 2017-09-28 at 13 58 29

Themes:

screen shot 2017-09-28 at 13 58 26

<details>
<summary>Smaller screens (more screenshots!)</summary>

screen shot 2017-09-28 at 13 59 17

screen shot 2017-09-28 at 13 59 13

screen shot 2017-09-28 at 13 59 08

screen shot 2017-09-28 at 13 59 01

</details>

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -1,0 +1,202 @@
+/* @flow */
+import React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import { withErrorHandler } from 'core/errorHandler';
+import log from 'core/logger';
+import translate from 'core/i18n/translate';
+import { parsePage } from 'core/utils';
+import { fetchCollection, fetchCollectionPage } from 'amo/reducers/collections';
+import type {
+  CollectionsState,
+  CollectionType,
+} from 'amo/reducers/collections';
+import AddonsCard from 'amo/components/AddonsCard';
+import Link from 'amo/components/Link';
+import NotFound from 'amo/components/ErrorPage/NotFound';
+import Paginate from 'core/components/Paginate';
+import Card from 'ui/components/Card';
+import LoadingText from 'ui/components/LoadingText';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { ReactRouterLocation } from 'core/types/router';
+
+import './styles.scss';
+
+
+type Props = {|
+  collection: CollectionType | null,
+  dispatch: Function,
+  errorHandler: ErrorHandlerType,
+  i18n: Object,
+  loading: boolean,
+  location: ReactRouterLocation,
+  params: {|
+    slug: string,
+    user: string,
+  |},
+|};
+
+export class CollectionBase extends React.Component {
+  componentWillMount() {
+    this.loadDataIfNeeded();
+  }
+
+  componentWillReceiveProps(nextProps: Props) {
+    this.loadDataIfNeeded(nextProps);
+  }
+
+  loadDataIfNeeded(nextProps?: Props) {
+    const { collection, errorHandler, loading, params } = {
+      ...this.props,
+      ...nextProps,
+    };
+
+    if (errorHandler.hasError()) {
+      log.warn('Not loading data because of an error.');
+      return;
+    }
+
+    if (loading) {
+      return;
+    }
+
+    let collectionChanged = false;
+    let addonsPageChanged = false;
+    let location = this.props.location;
+
+    if (nextProps && nextProps.location) {
+      const nextLocation = nextProps.location;
+
+      if (location.pathname !== nextLocation.pathname) {
+        collectionChanged = true;
+        location = nextLocation;
+      }
+
+      if (location.query.page !== nextLocation.query.page) {
+        addonsPageChanged = true;
+        location = nextLocation;
+      }
+    }
+
+    if (!collection || collectionChanged) {
+      this.props.dispatch(fetchCollection({
+        errorHandlerId: errorHandler.id,
+        page: parsePage(location.query.page),
+        slug: params.slug,
+        user: params.user,
+      }));
+
+      return;
+    }
+
+    if (collection && addonsPageChanged) {
+      this.props.dispatch(fetchCollectionPage({
+        errorHandlerId: errorHandler.id,
+        page: parsePage(location.query.page),
+        slug: params.slug,
+        user: params.user,
+      }));
+    }
+  }
+
+  props: Props;
+
+  url() {
+    const { params } = this.props;
+
+    return `/collections/${params.user}/${params.slug}/`;
+  }
+
+  renderCollection() {
+    const { collection, i18n, loading, location } = this.props;
+
+    const addons = collection ? collection.addons : [];
+
+    return (
+      <div className="Collection-wrapper">
+        <Card className="Collection-detail">
+          <h1 className="Collection-title">
+            {collection ? collection.name : <LoadingText />}
+          </h1>
+          <p className="Collection-description">
+            {collection ? collection.description : <LoadingText />}
+          </p>
+          <div className="Collection-metadata">
+            <dl className="Collection-number-of-addons">
+              <dd>
+                {collection ? collection.numberOfAddons : <LoadingText />}
+              </dd>
+              <dt>{i18n.gettext('Addons')}</dt>
+            </dl>
+            <dl className="Collection-author-name">
+              <dd className="Collection-author-name-value">
+                {collection ? collection.authorName : <LoadingText />}
+              </dd>
+              <dt>{i18n.gettext('Creator')}</dt>
+            </dl>
+            <dl className="Collection-last-updated">
+              <dd>
+                {collection ?
+                  i18n.moment(collection.lastUpdatedDate).format('ll')
+                  : <LoadingText />}
+              </dd>
+              <dt>{i18n.gettext('Last updated')}</dt>
+            </dl>
+          </div>
+        </Card>
+        <div className="Collection-items">
+          <AddonsCard
+            addons={addons}
+            loading={!collection || loading}
+          />
+          {collection && (
+            <Paginate
+              LinkComponent={Link}
+              count={collection.numberOfAddons}
+              currentPage={parsePage(location.query.page)}
+              pathname={this.url()}
+            />
+          )}
+        </div>
+      </div>
+    );
+  }
+
+  render() {
+    const { errorHandler } = this.props;
+
+    if (errorHandler.hasError()) {
+      log.warn('Captured API Error:', errorHandler.capturedError);
+      // 401 and 403 are for an add-on lookup is made to look like a 404 on
+      // purpose. See: https://github.com/mozilla/addons-frontend/issues/3061.
+      if (errorHandler.capturedError.responseStatusCode === 401 ||
+        errorHandler.capturedError.responseStatusCode === 403 ||
+        errorHandler.capturedError.responseStatusCode === 404
+      ) {
+        return <NotFound />;
+      }
+    }
+
+    return (
+      <div className="Collection">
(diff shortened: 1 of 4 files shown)


tofumatt:
Looks niiiice.

src/amo/components/Collection/index.js
```diff
@@ -0,0 +1,202 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import { withErrorHandler } from 'core/errorHandler';
+import log from 'core/logger';
+import translate from 'core/i18n/translate';
+import { parsePage } from 'core/utils';
+import { fetchCollection, fetchCollectionPage } from 'amo/reducers/collections';
+import type {
```
tofumatt: I thought types went at the bottom? 😉

src/amo/components/Collection/index.js
```diff
@@ -0,0 +1,202 @@
...
+  CollectionsState,
+  CollectionType,
+} from 'amo/reducers/collections';
+import AddonsCard from 'amo/components/AddonsCard';
+import Link from 'amo/components/Link';
```
tofumatt: I find without sorting these imports just feel incredibly scattered and hard to sort through.

I think we should force imports to be ordered sooner than later...

src/amo/components/Collection/index.js
```diff
@@ -0,0 +1,202 @@
...
+import Paginate from 'core/components/Paginate';
+import Card from 'ui/components/Card';
+import LoadingText from 'ui/components/LoadingText';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { ReactRouterLocation } from 'core/types/router';
+
+import './styles.scss';
+
+
+type Props = {|
```
tofumatt: I thought we usually called these "PropTypes". I don't care which ones but we should be consistent.

tofumatt: Actually I think I slightly prefer "PropTypes" 😉 because I think it helps not confuse things with actual props and signifies intent better to users used to using `prop-types`

src/amo/components/Collection/index.js
```diff
@@ -0,0 +1,202 @@
...
+        <Card className="Collection-detail">
+          <h1 className="Collection-title">
+            {collection ? collection.name : <LoadingText />}
+          </h1>
+          <p className="Collection-description">
+            {collection ? collection.description : <LoadingText />}
+          </p>
+          <div className="Collection-metadata">
+            <dl className="Collection-number-of-addons">
+              <dd>
```
tofumatt: `<dd>` first?! 😱

Seriously though I see why, I think that makes sense. It's just odd to see 😄

tests/unit/amo/components/TestCollection.js
```diff
@@ -0,0 +1,445 @@
...
+    const wrapper = renderComponent({
+      errorHandler,
+      location,
+      params: { slug, user },
+      store,
+    });
+    fakeDispatch.reset();
+
+    // This will trigger the componentWillReceiveProps() method.
+    wrapper.setProps({
```
tofumatt: Can we dispatch on the `store` we pass to `renderComponent` here to do this instead? It might be a bit nicer.

tests/unit/amo/components/TestCollection.js
```diff
@@ -0,0 +1,445 @@
...
+    const wrapper = renderComponent({
+      errorHandler,
+      location,
+      params: { slug, user },
+      store,
+    });
+    fakeDispatch.reset();
+
+    // This will trigger the componentWillReceiveProps() method.
+    wrapper.setProps({ location: newLocation });
```
tofumatt: Could we dispatch here instead as well?

tests/unit/amo/components/TestCollection.js
```diff
@@ -0,0 +1,445 @@
...
+
+    // User clicks on 'next' pagination link.
+    store.dispatch(fetchCollectionPage({
+      errorHandlerId: errorHandler.id,
+      page: 2,
+      slug,
+      user,
+    }));
+
+    wrapper.setProps(mapStateToProps(store.getState()));
```
tofumatt: Why do we have to `setProps` in so many places? Is `shallow` missing certain lifecycle stuff? I thought there was a special option to enable that? Or we could use `mount`? I'd like to approximate dispatches as closely as possible if we can and it's not a huge pain.

tofumatt: I think it's `lifecycleExperimental: true` we want to use, like here: https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestLandingPage.js#L37

src/amo/components/Collection/index.js
willdurand: 😅

src/amo/components/Collection/index.js
willdurand: Flow uses `Props` https://flow.org/en/docs/react/components/, that's why I use `Props`, but I am fine with `PropTypes`.

willdurand:
> So it seems like 404s don't work.

oh now, I broke something obviously. Sorry 😢 

> Also theme previews don't seem to show in the search results, but show once I'm in the detail view

Huh, weird. Can it be the component that is faulty? I will investigate.

willdurand:
Oh shoot, I see what happens for the 404s.. It only works with SSR disabled (which I used to test 404s), but when an error is thrown, `loading` is still `true` ...

tests/unit/amo/components/TestCollection.js
willdurand: `shallow` does not trigger lifecycle methods on purpose. We use `setProps()` quite often actually, and I thought that was the way to go since it does exactly what we want here, that is triggering `componentWillReceiveProps()`.

tests/unit/amo/components/TestCollection.js
willdurand: `location` and `params` are not handled by redux.

tests/unit/amo/components/TestCollection.js
willdurand: `location` and `params` are not handled by redux.

tests/unit/amo/components/TestCollection.js
willdurand: Note that here I use a `store` and dispatch the proper actions so isn't it "as closely as possible" already? I am open to suggestions, please let me know.

tofumatt:
Okay, works for me with `setProps` then, I think my only issue is with the 404.

tests/unit/amo/components/TestCollection.js
tofumatt: Yeah I think you're right, never mind me 😄

tofumatt:
In this case I think it's just r+wc, things like type order. Other issues should be split out into their own bugs.

tests/unit/amo/components/TestCollection.js
```diff
@@ -0,0 +1,445 @@
...
+    const wrapper = renderComponent({ store });
+
+    const id = 'error-handler-id';
+    const error = createApiError({
+      response: { status: 500 },
+      apiURL: 'https://some/api/endpoint',
+      jsonResponse: { message: 'Nope.' },
+    });
+    store.dispatch(setError({ id, error }));
+    const capturedError = store.getState().errors[id];
```
kumar303: This pattern pre-dates https://github.com/mozilla/addons-frontend/pull/3010 as well as `shallowUntilTarget()`.

I think you can do it just by dispatching an error which makes this a lot simpler.

```js
  id: 'some-id',
  dispatch: store.dispatch,
});
errorHandler.handle(new Error('an unexpected error'));

// You can only stub out dispatch after handling the error.
const fakeDispatch = sinon.stub(store, 'dispatch');

wrapper.setProps({ errorHandler });
```

tests/unit/amo/components/TestCollection.js
kumar303: > location and params are not handled by redux.

I'm looking forward to https://github.com/mozilla/addons-frontend/issues/181 for that

tests/unit/amo/components/TestCollection.js
kumar303: Similar to https://github.com/mozilla/addons-frontend/pull/3268/files#r142172324 I'm confused about why you need to call `setProps()` again too. When you call `store.dispatch()` it should automatically trigger the update, no? Do you need to call `wrapper.update()` maybe? Or does it work without `setProps()` if you introduce `lifecycleExperimental: true` ?

tests/unit/amo/components/TestCollection.js
```diff
@@ -0,0 +1,445 @@
...
+    const id = 'error-handler-id';
+    const store = dispatchClientMetadata().store;
+
+    const error = createApiError({
+      response: { status: 401 },
+      apiURL: 'https://some/api/endpoint',
+      jsonResponse: { message: 'unauthorized' },
+    });
+    store.dispatch(setError({ id, error }));
+    const capturedError = store.getState().errors[id];
```
kumar303: This test and a couple others could probably be simplified too

tests/unit/amo/components/TestCollection.js
kumar303: Actually, if you are only doing this for updates then you can probably do it like this:

```js
const wrapper = render(...);

const errorHandler = wrapper.instance().props.errorHandler;
errorHandler.handle(new Error('an unexpected error'));

const fakeDispatch = sinon.stub(store, 'dispatch');

wrapper.setProps({ errorHandler });
```

tests/unit/amo/components/TestCollection.js
willdurand: I did not know about `lifecycleExperimental` to be honest, and having `experimental` in an option does not sound very robust.

My main (and only) argument in favor of `setProps()` is that it is very explicit.

tests/unit/amo/components/TestCollection.js
willdurand: ah thanks. did not know this and I was pretty sure we merged code with the code I had here. The new code is definitely better though.

tests/unit/amo/components/TestCollection.js
willdurand: So without `setProps()` it does not do anything but it makes sense according to the `shallow` documentation. Calling `update()` does not do anything. Let's see if I can use `lifecycleExperimental` now.

willdurand: mmh, no success with `lifecycleExperimental` 😢 I tried to pass it as `shallowOptions` to `shallowUntilTarget` but it did not work, and could not make it work with a regular `shallow`.

willdurand: I left a comment to explain why `setProps` is needed here.

kumar303: Huh, that's weird how `lifecycleExperimental` doesn't fix it. It's fine as is, I just wish I knew it why it's not working :/

willdurand: Yeah I am not entirely done with this, I want to spend 30 more minutes on it before I give up.

willdurand: Nope, no luck. Follow-up at: #3317.

PR #3271, opened 2017-09-28
fix: Do not show version or history for themes

### LINKED ISSUE #3265: Unusable "See all versions" for themes
Steps to reproduce:
1. Load details page for a theme i.e. https://addons.allizom.org/en-US/firefox/addon/summer-ladybug/
2. Observe "More information" section

Expected results:
There is no "See all versions" link for themes.

Actual results:
There is a "See all versions" link fro themes that redirects to an extension empty "version history" page.

Notes/Issues:
Verified on AMO-stage FF55/FF58(Win 7, Android 7.2.1).
Video for this issue:
all vrs

tofumatt:
Fix #3265

Prevents showing version info is the version is "0" and only show versions for add-ons that support version history.

Themes more info cards now look like:

screenshot 2017-09-28 17 34 17

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -7,7 +7,11 @@ import Link from 'amo/components/Link';
 import ReportAbuseButton from 'amo/components/ReportAbuseButton';
 import translate from 'core/i18n/translate';
 import type { AddonType } from 'core/types/addons';
-import { isAddonAuthor, trimAndAddProtocolToUrl } from 'core/utils';
+import {
+  addonHasVersionHistory,
+  isAddonAuthor,
+  trimAndAddProtocolToUrl,
+} from 'core/utils';
 import Card from 'ui/components/Card';
 import LoadingText from 'ui/components/LoadingText';
 
@@ -28,11 +32,9 @@ export class AddonMoreInfoBase extends React.Component {
 
     if (!addon) {
       return this.renderDefinitions({
-        version: <LoadingText minWidth={20} />,
         versionLastUpdated: <LoadingText minWidth={20} />,
         versionLicense: <LoadingText minWidth={20} />,
         addonId: <LoadingText minWidth={20} />,
-        versionHistoryLink: <LoadingText minWidth={20} />,
       });
     }
 
@@ -65,7 +67,8 @@ export class AddonMoreInfoBase extends React.Component {
           {i18n.gettext('Visit stats dashboard')}
         </Link>
       ) : null,
-      version: addon.current_version.version,
+      version: addonHasVersionHistory(addon) ?
+        addon.current_version.version : null,
       versionLastUpdated: i18n.sprintf(
         // translators: This will output, in English:
         // "2 months ago (Dec 12 2016)"
@@ -99,14 +102,14 @@ export class AddonMoreInfoBase extends React.Component {
         </Link>
       ) : null,
       addonId: addon.id,
-      versionHistoryLink: (
+      versionHistoryLink: addonHasVersionHistory(addon) ? (
         <Link
           className="AddonMoreInfo-version-history-link"
           href={`/addon/${addon.slug}/versions/`}
         >
           {i18n.gettext('See all versions')}
         </Link>
-      ),
+      ) : null,
       // Since current_beta_version is just an alias to the latest beta,
       // we can assume that no betas exist at all if it is null.
       betaVersionsLink: addon.current_beta_version ? (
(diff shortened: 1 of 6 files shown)


src/amo/components/AddonMoreInfo/index.js
```diff
@@ -65,7 +67,9 @@ export class AddonMoreInfoBase extends React.Component {
           {i18n.gettext('Visit stats dashboard')}
         </Link>
       ) : null,
-      version: addon.current_version.version,
+      // Themes are sent with a version of "0", which we ignore.
+      version: addon.current_version.version !== '0' ?
```
kumar303: This needs to check `addon.type` similar to how `addonHasVersionHistory()` works.

I was curious if a real extension could have a version of "0" and indeed it can:

screenshot 2017-10-02 09 50 34

tofumatt: Hah! Good catch, I didn't think that'd be allowed.

tofumatt:
Nice catch! Fixed up and ready for another review.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -65,7 +67,9 @@ export class AddonMoreInfoBase extends React.Component {
           {i18n.gettext('Visit stats dashboard')}
         </Link>
       ) : null,
-      version: addon.current_version.version,
+      // Themes are sent with a version of "0", which we ignore.
```
kumar303: This can be deleted

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -254,8 +261,61 @@ describe(__filename, () => {
...
 
-  it('links to version history', () => {
-    const addon = createInternalAddon({ ...fakeAddon, slug: 'some-slug' });
+  // Add-ons can have a zero version! So we originally thought to check
+  // for a "0" version value (which themes will have set). But they are
+  // valid–this test is here to remind us of this :-)
+  //
+  // See:
+  // https://github.com/mozilla/addons-frontend/pull/3271#discussion_r142159199
+  it('shows verson even if it is zero', () => {
```
kumar303: I am not a fan of adding tests like this unless they are a regression test for a specific bug that actually happened. The problem is that every test adds more time to the suite so adding tests that don't prevent any known bugs do incur a cost.

If you really think we need a reminder, I would suggest adding a comment to the Flow type like this:

```diff
diff --git a/src/core/types/addons.js b/src/core/types/addons.js
index 6be02718..3a45668c 100644
--- a/src/core/types/addons.js
+++ b/src/core/types/addons.js
@@ -38,6 +38,7 @@ export type AddonVersionType = {|
   // The `text` property is omitted from addon.current_version.license.
   license: { name: string, url: string },
   reviewed: Date,
+  // This is the developer defined version number. It could, for example, be set to "0".
   version: string,
 |};
 
```

Also, it could be helpful to update [`fakeTheme`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/helpers.js#L96) so that it has `version: "0"`.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -264,6 +324,47 @@ describe(__filename, () => {
...
+
+    expect(root.find('.AddonMoreInfo-version-history-title'))
+      .toHaveLength(0);
+    expect(root.find('.AddonMoreInfo-version-history-link'))
+      .toHaveLength(0);
+  });
+
+  it('omits version history for themes', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
```
kumar303: This should use `...fakeTheme`

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: It's a bug that would have happened and I think tests make it more obvious than the type. Even if we had complete flow coverage "0" would be a valid type but one might still think to hide it as an invalid version. The tests are here to tell us how the app should work and I think a version of "0", because it isn't immediately obvious as valid, should be called out as such.

tests/unit/amo/helpers.js
```diff
@@ -124,6 +124,7 @@ export const fakeTheme = Object.freeze({
     version: '1.0',
   },
   type: ADDON_TYPE_THEME,
+  version: '0',
```
kumar303: This should have been:

```js
current_version: {
  ...fakeAddon.current_version,
  version: 0,
},
```

tofumatt: Blarg, right, I'm a tool. I'll fix it up and merge it if it's green.

tofumatt: Though it should be a string, not a number.

Issue #3277, opened 2017-09-29
Add an inline text form component for review replies


kumar303:
I wanted to use the inline text form from abuse reporting (https://github.com/mozilla/addons-frontend/issues/2759) in the upcoming reply-to-review feature (https://github.com/mozilla/addons-frontend/issues/2775). This issue is to extract a component that will work for both of them.

PR #3278, opened 2017-09-29
Add an inline text form component for review replies

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3277

I'm still working on the UI in another branch. This is what the UX looks like:

dismissable-textarea mov

### DIFF
--- a/src/ui/components/DismissibleTextForm/index.js
+++ b/src/ui/components/DismissibleTextForm/index.js
@@ -28,6 +28,7 @@ type PropTypes = {|
   i18n: Object,
   isSubmitting?: boolean,
   placeholder?: string,
+  submitButtonClassName?: string,
   submitButtonText?: string,
   submitButtonInProgressText?: string,
   text?: string,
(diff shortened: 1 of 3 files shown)


kumar303:
The review reply is a little different than the report abuse form but they both have a form that you can show/hide. This adds a component that I think should work for both and so far it is hooked up for the review reply (which works).

src/ui/components/DismissibleTextForm/index.js
```diff
@@ -0,0 +1,140 @@
...
+              'DismissibleTextForm-dismiss--disabled': isSubmitting,
+            })}
+            href="#dismiss"
+            onClick={this.onDismiss}
+          >
+            {i18n.gettext('Dismiss')}
+          </a>
+          <Button
+            type="submit"
+            className="DismissibleTextForm-submit Button--action Button--small"
```
tofumatt: We'll want a prop we can pass here to control the button's style; on the report abuse form we style this as `Button--alert` or something to make it red.

I think just adding a `buttonClassName` and adding it in with `classNames()` should do the trick.

kumar303:
Thanks for the quick review

Issue #3280, opened 2017-09-29
Hook up the reply to review API


kumar303:
This hooks up the reply to review API https://addons-server.readthedocs.io/en/latest/topics/api/reviews.html#reply

The UI for this will be done in https://github.com/mozilla/addons-frontend/issues/2775

PR #3283, opened 2017-09-29
Hook up the reply-to-review API

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3280

This adds functions for the reply-to-review API but the UI is forthcoming. See https://github.com/mozilla/addons-frontend/issues/2775

I had to touch some ancient code for this so I cleaned up a few things (but not everything).


tests/unit/helpers.js
```diff
@@ -63,6 +63,7 @@ export function getFakeAddonManagerWrapper({
 }
 
 export function unexpectedSuccess() {
+  // TODO: fix this so that the failure message is better.
```
kumar303: This is coming in the next patch. The problem is that `expect(false).toBe(true)` produces a terrible error message that doesn't help you figure out what went wrong.

kumar303:
@willdurand @tofumatt I tagged you both just to see who's available but it only needs one review. Nothing out of the ordinary is happening in here.

src/amo/api/reviews.js
```diff
@@ -87,6 +87,38 @@ export function submitReview({
...
+  originalReviewId,
+  title,
+}: ReplyToReviewParams = {}): Promise<ApiReviewType> => {
+  return new Promise((resolve) => {
+    const endpoint = `reviews/review/${originalReviewId}/reply/`;
+
+    resolve(callApi({
+      auth: true,
+      body: {
+        body,
```
tofumatt: Haha, I always laugh when stuff like this happens 😄

tests/unit/amo/api/test_reviews.js
```diff
@@ -123,37 +134,35 @@ describe('amo.api', () => {
...
-        mockApi.verify();
-      });
+      await submitReview(params);
+      mockApi.verify();
     });
   });
 
   describe('getReviews', () => {
-    it('allows you to fetch reviews by any param', () => {
+    it('allows you to fetch reviews by any param', async () => {
```
tofumatt: Sorry I didn't catch this before but I'm a bit confused by this spec. Is it just saying it can fetch the reviews by addon ID instead of slug?

tests/unit/amo/sagas/testReviews.js
```diff
@@ -1,7 +1,7 @@
 import { hideLoading, showLoading } from 'react-redux-loading-bar';
 import SagaTester from 'redux-saga-tester';
 
-import * as amoApi from 'amo/api';
+import * as reviewsApi from 'amo/api/reviews';
```
tofumatt: Thanks for moving those API methods to their own folder btw, it makes them a lot easier to sort through 😄

tests/unit/amo/api/test_reviews.js
kumar303: Hmm, it's old code so I'm not sure. I think "it allows you to fetch reviews with filters" will be better.

PR #3285, opened 2017-09-30
feat: Add more section to header

### LINKED ISSUE #3281: Add More Section to header
Turns out the UX of having the "Extensions" link in the header also be a dropdown was quite bad and I've opted to return to the old "More..." link–it's in a previous mock-up.

We're going to need this for the Language Tools and Search add-ons so I'm gonna add it.

tofumatt:
fix #3281.

This adds links to Language Tools, Search Tools, and other apps to the header. This is in advance of language tools page.

The Explore link is the same as the home link, so we hide it on smaller screens because otherwise the header gets too crammed.

### Screenshots

screenshot 2017-09-30 01 29 56
screenshot 2017-09-30 01 29 50
screenshot 2017-09-30 01 29 25
screenshot 2017-09-30 01 29 19

----

I know we argued again the "More" link but using "Extensions" as both a link and a dropdown was awkward UX and inconsistent. Putting the link inside that dropdown seemed like bad UX because then it was extra work (and two taps on a touch device) to get to Extensions landing page which is a very important page. So I settled on the more link, especially because I think the Dropdown @willdurand added is quite nice 😄

### DIFF
--- a/src/amo/components/SectionLinks/index.js
+++ b/src/amo/components/SectionLinks/index.js
@@ -1,41 +1,73 @@
+/* @flow */
 import classNames from 'classnames';
 import React from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
 
 import Link from 'amo/components/Link';
-import type { ViewContextType } from 'amo/reducers/viewContext';
+import { setClientApp } from 'core/actions';
 import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
+  CLIENT_APP_ANDROID,
+  CLIENT_APP_FIREFOX,
   VIEW_CONTEXT_EXPLORE,
   VIEW_CONTEXT_HOME,
+  VIEW_CONTEXT_LANGUAGE_TOOLS,
 } from 'core/constants';
 import translate from 'core/i18n/translate';
 import { visibleAddonType } from 'core/utils';
+import DropdownMenu from 'ui/components/DropdownMenu';
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
+import type { ViewContextType } from 'amo/reducers/viewContext';
+import type { ApiStateType } from 'core/reducers/api';
 
 import './styles.scss';
 
 
-type SectionLinksProps = {
+type SectionLinksProps = {|
   className?: string,
+  clientApp: string,
+  dispatch: Function,
   i18n: Object,
+  router: Object,
   viewContext: ViewContextType,
-}
+|};
 
 export class SectionLinksBase extends React.Component {
+  setClientApp = (event: Object) => {
+    event.preventDefault();
+
+    const { dispatch, router } = this.props;
+
+    const clientApp = event.currentTarget.getAttribute('data-clientapp');
+
+    dispatch(setClientApp(clientApp));
+    router.push(event.currentTarget.getAttribute('href'));
+  }
+
   props: SectionLinksProps;
 
   render() {
-    const { className, i18n, viewContext } = this.props;
+    const { className, clientApp, i18n, viewContext } = this.props;
     const isExploring = [VIEW_CONTEXT_EXPLORE, VIEW_CONTEXT_HOME]
       .includes(viewContext);
+
+    let forBrowserNameText;
+    if (clientApp === CLIENT_APP_FIREFOX) {
+      forBrowserNameText = i18n.gettext('for Firefox');
+    } else if (clientApp === CLIENT_APP_ANDROID) {
+      forBrowserNameText = i18n.gettext('for Android');
+    }
 
     return (
       <ul className={classNames('SectionLinks', className)}>
         <li>
           <Link
-            className={classNames('SectionLinks-link', { 'SectionLinks-link--active': isExploring })}
+            className={classNames('SectionLinks-link', 'SectionLinks-explore', {
+              'SectionLinks-link--active': isExploring,
+            })}
             to="/"
           >
             {i18n.gettext('Explore')}
@@ -44,7 +76,8 @@ export class SectionLinksBase extends React.Component {
         <li>
           <Link
             className={classNames('SectionLinks-link', {
-              'SectionLinks-link--active': viewContext === ADDON_TYPE_EXTENSION,
+              'SectionLinks-link--active': viewContext ===
+                ADDON_TYPE_EXTENSION,
             })}
             to={`/${visibleAddonType(ADDON_TYPE_EXTENSION)}/`}
           >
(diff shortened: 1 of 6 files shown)


tofumatt:
Moving this review to @willdurand to even out review load 😄

src/amo/components/SectionLinks/index.js
```diff
@@ -2,40 +2,71 @@ import classNames from 'classnames';
...
 import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
+  CLIENT_APP_ANDROID,
+  CLIENT_APP_FIREFOX,
   VIEW_CONTEXT_EXPLORE,
   VIEW_CONTEXT_HOME,
+  VIEW_CONTEXT_LANGUAGE_TOOLS,
 } from 'core/constants';
+import type { ApiStateType } from 'core/reducers/api';
```
willdurand: I tend to put the `import type` at the end of the `import` block, but I don't know if it is the rule here. Also, if you use Flow without `// @flow` at the top of the file, it will not have any effect if I remember correctly.

src/amo/components/SectionLinks/index.js
```diff
@@ -2,40 +2,71 @@ import classNames from 'classnames';
...
 import translate from 'core/i18n/translate';
 import { visibleAddonType } from 'core/utils';
+import DropdownMenu from 'ui/components/DropdownMenu';
+import DropdownMenuItem from 'ui/components/DropdownMenuItem';
 
 import './styles.scss';
 
 
 type SectionLinksProps = {
```
willdurand: Could be stricter.

src/amo/components/SectionLinks/index.js
```diff
@@ -2,40 +2,71 @@ import classNames from 'classnames';
...
   className?: string,
+  clientApp: string,
+  dispatch: Function,
   i18n: Object,
+  router: Object,
   viewContext: ViewContextType,
 }
 
 export class SectionLinksBase extends React.Component {
+  setClientApp = (event) => {
```
willdurand: Could be nice to have a type for this event so that we know where it comes from (or at least have a hint).

src/amo/components/SectionLinks/index.js
```diff
@@ -2,40 +2,71 @@ import classNames from 'classnames';
...
-    const { className, i18n, viewContext } = this.props;
+    const { className, clientApp, i18n, viewContext } = this.props;
     const isExploring = [VIEW_CONTEXT_EXPLORE, VIEW_CONTEXT_HOME]
       .includes(viewContext);
 
+    let forBrowserNameText;
+    if (clientApp === CLIENT_APP_FIREFOX) {
+      forBrowserNameText = i18n.gettext('for Firefox');
+    } else if (clientApp === CLIENT_APP_ANDROID) {
+      forBrowserNameText = i18n.gettext('for Firefox for Android');
```
willdurand: Just to be sure: it is really "for Firefox for Android"? In both cases, it is for Firefox no? Why indicating that it is for Android when you probably run Firefox for Android on your Android device?

src/amo/components/SectionLinks/index.js
```diff
@@ -2,40 +2,71 @@ import classNames from 'classnames';
...
+    }
+
     return (
       <ul className={classNames('SectionLinks', className)}>
         <li>
           <Link
-            className={classNames('SectionLinks-link', { 'SectionLinks-link--active': isExploring })}
+            className={classNames('SectionLinks-link', 'SectionLinks-explore',
+              { 'SectionLinks-link--active': isExploring }
+            )}
```
willdurand: Can we format this block like this?

```
className={classNames('SectionLinks-link', 'SectionLinks-explore', {
  'SectionLinks-link--active': isExploring,
})
```

I find it more readable and the diff will be slightly better if we need to add a new class in the future.

src/amo/components/SectionLinks/index.js
willdurand: I tested the feature locally and shouldn't this be "for Android"?

When you click on "Other Browser Sites > Add-ons for Firefox", you get:

screen shot 2017-10-01 at 00 55 49

The first menu section header is named exactly like the link you've navigated to. However, if you click "Other Browser Sites > Add-ons for Android", then you get a slightly different section header:

screen shot 2017-10-01 at 00 55 54

tests/unit/amo/components/TestSectionLinks.js
```diff
@@ -64,4 +72,67 @@ describe('SectionLinks Component', () => {
...
+
+    root.find(`.SectionLinks-clientApp-android`)
+      .simulate('click', fakeEvent);
+
+    sinon.assert.called(fakeEvent.preventDefault);
+    sinon.assert.calledWith(getAttributeSpy, 'data-clientapp');
+    sinon.assert.calledWith(getAttributeSpy, 'href');
+    sinon.assert.calledWith(dispatchSpy, setClientApp(CLIENT_APP_ANDROID));
+    sinon.assert.calledWith(fakeRouter.push, `/en-US/${CLIENT_APP_ANDROID}/`);
+  });
```
willdurand: I would have expected a test to make sure we render a `DropdownMenu`.

tofumatt:
Yeah, all the links fallback to the old site when not run locally. And the language tools page is in another PR so that should come soon. 😄

src/amo/components/SectionLinks/index.js
tofumatt: Oh, I guess that's a fine way to organise them. I tend to sort things by the filename we're importing from but separating out types wfm. 👍

tofumatt: Oh, and yeah I constantly forget to add the flow annotation, my bad 😞

src/amo/components/SectionLinks/index.js
tofumatt: I tried using `Event` and `SyntheticEvent` but they complain about our (reasonable) use of `currentTarget` (and care about `target`) too.

src/amo/components/SectionLinks/index.js
tofumatt: I wanted to be clear that we were targetting "Firefox for Android"... but that's a good point, it's confusing and inconsistent. Good catch; I'll switch it to just "for Android". Reads better, too. 👍

tests/unit/amo/components/TestSectionLinks.js
tofumatt: Oh, good call. Will add one.

src/amo/components/SectionLinks/index.js
willdurand: Yeah, this is a mess, and this reply does not really help: https://github.com/facebook/flow/issues/4381.

PR #3288, opened 2017-09-30
chore: Add LanguageTools API/sagas

### LINKED ISSUE #3287: Create Language Tools API/Sagas
Supports #2777.

This will allow us to call the language tools API to get dictionaries and language packs on the new front-end.

tofumatt:
close #3287

Adds support to getting language tools and adding them into the add-ons state.


src/core/api/languageTools.js
```diff
@@ -0,0 +1,17 @@
...
+import { callApi } from 'core/api';
+import type { ApiStateType } from 'core/reducers/api';
+
+
+export type LanguageToolsParams = {|
+  api: ApiStateType,
+|};
+
+export function languageTools({ api }: LanguageToolsParams) {
+  return callApi({
```
kumar303: This should set `auth: true` so that the user's token will be sent in the request in case they're signed in . I'm really not sure why `auth` defaults to `false` in `callApi()`. If you meant to never send a login token then this should explicitly set `auth: false`.

src/core/api/languageTools.js
```diff
@@ -0,0 +1,17 @@
...
+    endpoint: 'addons/language-tools',
```
kumar303: The API docs suggest that this URL should have a trailing slash

src/core/sagas/languageTools.js
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    const state = yield select(getState);
+
+    const response = yield call(languageToolsApi, {
+      api: state.api,
+      auth: true,
+    });
+
+    yield put(loadLanguageTools({ addons: response.results }));
```
kumar303: Since these are just add-ons (with a reduced fieldset) you should call `loadAddons()` here. I don't think a separate `loadLanguageTools()` action is necessary. This will cause the add-on objects to always be stored with `createInternalAddon()` which I think is a good idea. We will then have a consistent representation of all add-on objects even if a bunch of the fields are null or undefined. This also means you would need to update `createInternalAddon()` so that it sets the language tool specific fields to null or undefined when they aren't set.

src/core/types/addons.js
```diff
@@ -47,6 +47,12 @@ export type AddonAuthorType = {|
   username: string,
 |};
 
+type LanguageToolType = {|
```
kumar303: I find this name misleading since it implies it's the full type of a language tool add-on. Instead, it's used here to indicate the extra fields possessed by language tools. I think that's fine given some other limitations we have for expressing types in this file. I would just recommend changing the name to `LanguageToolData` which follows the pattern we have for things like `ThemeData` within this file.

tests/unit/core/api/test_languageTools.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+describe(__filename, () => {
+  let mockApi;
+
+  beforeEach(() => {
+    mockApi = sinon.mock(api);
+  });
+
+  describe('languageTools API', () => {
```
kumar303: new test files should use async/await

tests/unit/core/sagas/testLanguageTools.js
```diff
@@ -0,0 +1,83 @@
...
+
+describe(__filename, () => {
+  let errorHandler;
+  let mockApi;
+  let sagaTester;
+
+  beforeEach(() => {
+    errorHandler = createStubErrorHandler();
+    mockApi = sinon.mock(api);
+    const initialState = dispatchClientMetadata().state;
```
kumar303: This could be moved inline:

```js
sagaTester = new SagaTester({
  initialState: dispatchClientMetadata().state,
});
```

src/core/api/languageTools.js
willdurand: Can we fill an issue about this if no issue exists yet ? It seems we all get caught by that. I would suggest to rename this to `sendCredentials` or something that is less misleading than `auth`.

src/core/api/languageTools.js
tofumatt: `callApi` adds the trailing slash.

src/core/sagas/languageTools.js
tofumatt: Using loadAddons means I have to use normalizr or set up the entities weirdness to get the actions to work. I really didn't want to do that as we want to get closer to using the responses from the API.

I would rather add a TODO: that says this can be converted to loadAddons once #2917 is closed.

src/core/types/addons.js
tofumatt: Makes sense, thanks.

tofumatt:
As mentioned at https://github.com/mozilla/addons-frontend/pull/3288#discussion_r142170670 I think changing this to use `loadAddons()` isn't possible until #2917 is closed, but I've made the other requested changes.

If you think it's worth adding a normalizr mapping to make them return similar results for now, Iet me know.

src/core/api/languageTools.js
kumar303: Yes, that would be great. I am not aware of an issue so please file it.

src/core/sagas/languageTools.js
kumar303: ok, then I would suggest making a new action creator called `loadAddonsFromResults()` with a TODO comment in `loadAddons()` saying that it is deprecated. The key change I'd like to see is that the reducer uses `createInternalAddon(addons[key])` not `createInternalLanguageTool(addons[key])` (I don't think we need that function).

kumar303:
I think we need to always store a consistent `addon` object in state. I added a different idea for how to do that in https://github.com/mozilla/addons-frontend/pull/3288#discussion_r142195468

src/core/sagas/languageTools.js
kumar303: The next step would be this:

> ...you would need to update createInternalAddon() so that it sets the language tool specific fields to null or undefined when they aren't set.

This enables every function working with the state data to always gets the same object, no matter what type of add-on it is. This will cut down on surprises.

tofumatt:
Okay, this is now using `LOAD_ADDONS` with a different action (`loadAddonResults`) that doesn't rely on normalizr stuff. It also means I've updated the flow types, etc.

tofumatt:
The test failure on CircleCI is something about npm3 install checksums... I don't think it's related...

src/core/types/addons.js
```diff
@@ -99,6 +104,7 @@ export type ExternalAddonType = {|
   is_source_public: boolean,
   last_updated: Date,
   latest_unlisted_version: ?AddonVersionType,
+  locale_disambiguation?: string,
```
kumar303: When an `ExternalAddonType` is a language tool it will only have [a limited field set](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#language-tools). Because of this, it would be good to go through [`ExternalAddonType`](https://github.com/mozilla/addons-frontend/blob/74779218ba2a2a5522f0c1e5459ac5d93c041e62/src/core/types/addons.js#L86) and add question marks to all the fields that would be undefined in the case of a language tool. This will help us catch parts of our code that might throw an error for undefined field values.

However, if the Flow errors force you to write overly defensive code then we may need to take a different approach. Try it and see. The other approach is to use literal strings for `addon.type` but it will require opt'ing in more files to Flow and a bit of effort.

tests/unit/core/api/test_languageTools.js
```diff
@@ -0,0 +1,46 @@
...
+    it('calls the language tools API', async () => {
+      const apiState = dispatchClientMetadata().store.getState().api;
+
+      mockApi
+        .expects('callApi')
+        .withArgs({
+          auth: true,
+          endpoint: 'addons/language-tools',
+          method: 'GET',
+          params: { app: apiState.clientApp },
```
kumar303: Since we don't have Flow coverage in this file, I'm going to suggest something a bit more annoying but safer:

```js
const clientApp = CLIENT_APP_ANDROID;
const apiState = dispatchClientMetadata({ clientApp }).store.getState().api;

mockApi
  ...
  .withArgs({
    ...
    params: { app: clientApp },
  });
```

tests/unit/core/sagas/testLanguageTools.js
```diff
@@ -0,0 +1,82 @@
...
+    const loadAction = sagaTester.getCalledActions()[2];
+    expect(loadAction).toEqual(expectedLoadAction);
+  });
+
+  it('clears the error handler', async () => {
+    sagaTester.dispatch(fetchLanguageTools({
+      errorHandlerId: errorHandler.id,
+    }));
+
+    await sagaTester.waitFor(CLEAR_ERROR);
```
kumar303: This would be more maintainable as:

```js
const clearingAction = errorHandler.createClearingAction();
await sagaTester.waitFor(clearingAction.type);

expect(sagaTester.getCalledActions()[1])
  .toEqual(clearingAction);
```

src/core/sagas/languageTools.js
```diff
@@ -0,0 +1,40 @@
...
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
+
+    const response = yield call(languageToolsApi, {
+      api: state.api,
+      auth: true,
```
kumar303: This is not a valid parameter

tests/unit/core/sagas/testLanguageTools.js
```diff
@@ -0,0 +1,82 @@
...
+    sagaTester.start(languageToolsSaga);
+  });
+
+  it('calls the API for language tools', async () => {
+    const addon = { ...fakeAddon, slug: 'fancy' };
+    const response = { results: createFakeLanguageAddon({ addon }) };
+
+    mockApi
+      .expects('languageTools')
+      .once()
```
kumar303: it's worth doing `withArgs({ api: apiState })` here

src/core/types/addons.js
tofumatt: Generally we will fetch that info on the details page, but I getcha. It's not trivial to make the change but I think it makes sense... I'll try for like twenty minutes or so.

tofumatt: Most of the issues are around GUIDs though, which maybe do exist and we could get the API to send down.

tests/unit/core/sagas/testLanguageTools.js
tofumatt: True! Confusingly the `callApi` takes the `state` param but it really is `apiState`. I'll add the `withArgs` here but I've also filed #3346.

tests/unit/core/sagas/testLanguageTools.js
tofumatt: Ah, true. We do it the way I committed elsewhere in the app. I've filed: #3347

tofumatt:
Changes made. I've filed https://github.com/mozilla/addons-server/issues/6586 because language tools have a GUID but just aren't being sent. It will actually be useful to have it for installation right from this page in the future so I've requested it and it prevents us from marking the GUID optional in flow (and thus nearly all the flow errors I hit when marking things optional).

tests/unit/core/api/test_languageTools.js
```diff
@@ -0,0 +1,49 @@
...
+          auth: true,
+          endpoint: 'addons/language-tools',
+          method: 'GET',
+          params: { app: CLIENT_APP_ANDROID },
+          state: apiState,
+        })
+        .once()
+        .returns(mockResponse([createFakeLanguageAddon()]));
+
+      await languageTools({ api: apiState });
```
kumar303: I think it's helpful to also verify the result of the promise in an assertion.

tests/unit/core/sagas/testLanguageTools.js
```diff
@@ -0,0 +1,84 @@
...
+      .withArgs({ api: sagaTester.getState().api })
```
kumar303: Whoa, I didn't know about `sagaTester.getState()` 🍒

tests/unit/core/api/test_languageTools.js
tofumatt: Will do!

Issue #3234, opened 2017-09-26
Make fakeAddon a helper using createInternalAddon, not an object


tofumatt:
Using `fakeAddon` isn't perfect and @kumar303 pointed out we should wrap it in `createInternalAddon()` from `core/reducers/addons`, but this leads to a LOT of boilterplate in tests. See: https://github.com/mozilla/addons-frontend/pull/3203#discussion_r140777651

What we should really have is a `fakeAddon()` helper that takes extra props, overriding the defaults in the `fakeAddon` object and returns the result of `createInternalAddon()`.

willdurand:
We started to use various `createXYZ` helpers, so maybe we could introduce a `createFakeAddon` helper like:

```js
const createFakeAddon = (props = {}) => createInternalAddon({ ...fakeAddon, ...props });
```

That would ease the migration.

bobsilverberg:
We just need to be careful that everyone understands the naming conventions. `fakeAddon` is an *external* addon, not an internal addon (as are all of the other `fakeXxx` objects afaik). The name `createFakeAddon` therefore seems a bit ambiguous to me: does it create an external addon or an internal addon? It sounds from the above comment that we should assume that any `createXxx` function creates an internal object, but that doesn't seem necessarily intuitive to me.

Personally, I'm happy with just having `fakeXxx` objects that are external objects and knowing that I can call `createInternalXxx` in order to transform them into internal objects if I need an internal object. I feel that results in tests that are easy to follow without having to make any assumptions about what a `createFakeXxx` function might be returning.

As an aside, I've also come across a few tests recently where we were using an internal object when we really should have been using an external object. This can happen especially with `addon` because the internal object (prior to my recent refactor) contained all of the properties of the external object. This is another reason why I think it's valuable for tests to be explicit about whether they are using internal objects or external objects.

willdurand:
> Personally, I'm happy with just having `fakeXxx` objects that are external objects and knowing that I can call `createInternalXxx` in order to transform them into internal objects if I need an internal object. I feel that results in tests that are easy to follow without having to make any assumptions about what a `createFakeXxx` function might be returning.
> 
> As an aside, I've also come across a few tests recently where we were using an internal object when we really should have been using an external object. This can happen especially with `addon` because the internal object (prior to my recent refactor) contained all of the properties of the external object. This is another reason why I think it's valuable for tests to be explicit about whether they are using internal objects or external objects.

[...]

PR #3320, opened 2017-10-02
Use internal addon representation and themeData

### LINKED ISSUE #3318: No preview of theme in collection view
The collection view should display a preview of the themes, currently it shows "no preview available" but this is not correct. It should display a preview of these themes for instance:

screen shot 2017-10-02 at 23 52 04

### LINKED ISSUE #3319: Use internal addon representation in all reducers
It turns out that we do not use the internal representation of add-ons in all reducers. Let's leverage `createInternalAddon()` from `core/reducers/addons` to fix this so that all components deal with the exact same representation of an add-on (that is `AddonType`).

willdurand:
Fix #3318
Fix #3319 

---

This PR ensures we use an internal addon representation and exposes a `themeData` property to this internal representation. It helped fixing an issue with collections not displaying a preview (because the `collections` reducer uses internal addon representation).

(Note: I am aware of #3234 but did not want to open a big PR)

Before:

screen shot 2017-10-02 at 23 52 04

After:

screen shot 2017-10-03 at 00 51 36

### DIFF
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -1,5 +1,6 @@
 /* @flow */
-import type { AddonType } from 'core/types/addons';
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
+import { createInternalAddon } from 'core/reducers/addons';
 
 
 type State = {
@@ -67,7 +68,7 @@ export const fetchOtherAddonsByAuthors = (
 
 type LoadOtherAddonsByAuthorsParams = {|
   slug: string,
-  addons: Array<AddonType>,
+  addons: Array<ExternalAddonType>,
 |};
 
 type LoadOtherAddonsByAuthorsAction = {|
@@ -118,7 +119,8 @@ const reducer = (
             // This ensures we do not display the main add-on in the list of
             // "add-ons by these authors".
             .filter((addon) => addon.slug !== action.payload.slug)
-            .slice(0, OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE),
+            .slice(0, OTHER_ADDONS_BY_AUTHORS_PAGE_SIZE)
+            .map((addon) => createInternalAddon(addon)),
         },
       };
     default:
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -172,6 +172,8 @@ export function createInternalAddon(
         updateURL: apiAddon.theme_data.updateURL,
         version: apiAddon.theme_data.version,
       }),
+      // We should be using this property.
+      themeData: removeUndefinedProps({ ...apiAddon.theme_data }),
     };
   }
 
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -4,7 +4,11 @@ import { oneLine } from 'common-tags';
 import { ADDON_TYPE_THEME } from 'core/constants';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import log from 'core/logger';
-import type { AddonType, ExternalAddonType } from 'core/types/addons';
+import type {
+  AddonType,
+  ExternalAddonType,
+  ThemeData,
+} from 'core/types/addons';
 
 
 export const LOAD_ADDONS = 'LOAD_ADDONS';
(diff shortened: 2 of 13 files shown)


src/amo/reducers/addonsByAuthors.js
```diff
@@ -67,7 +68,7 @@ export const fetchOtherAddonsByAuthors = (
 
 type LoadOtherAddonsByAuthorsParams = {|
   slug: string,
-  addons: Array<AddonType>,
+  addons: Array<ExternalAddonType>,
```
kumar303: whoops, good catch

src/core/reducers/addons.js
```diff
@@ -172,6 +172,8 @@ export function createInternalAddon(
         updateURL: apiAddon.theme_data.updateURL,
         version: apiAddon.theme_data.version,
       }),
+      // We should be using this property.
+      themeData: removeUndefinedProps({ ...apiAddon.theme_data }),
```
kumar303: This can just be `themeData: apiAddon.theme_data` because `removeUndefinedProps()` was only necessary to make the spread not override existing property values with `undefined`.

kumar303: I think we need to introduce a mapper and call this like `themeData: createInternalThemeData(apiAddon)` because of several reasons:

* Even though it adds verbosity, I think it's helpful to see all the fields assigned one by one
* When a new field is added to the API, we won't introduce the new field by surprise
* We already have to fix the mapping for one of the fields (`description`) because of a bug
* We may need to re-map others later on too.

You can move [all of this stuff](https://github.com/willdurand/addons-frontend/blob/11d457c73768a0e6b4861ef8803e3bc17ba485b1/src/core/reducers/addons.js#L146-L174) into  a new `createInternalThemeData()` function.

tests/unit/amo/reducers/testLanding.js
```diff
@@ -103,7 +104,10 @@ describe('landing reducer', () => {
       });
       expect(state.featured.count).toEqual(2);
       expect(state.featured.results)
-        .toEqual([{ slug: 'foo' }, { slug: 'food' }]);
+        .toEqual([
+          createInternalAddon({ slug: 'foo' }),
```
kumar303: these should both be `{ ...fakeAddon, slug: '...' }`

src/core/reducers/addons.js
```diff
@@ -172,6 +172,8 @@ export function createInternalAddon(
         updateURL: apiAddon.theme_data.updateURL,
         version: apiAddon.theme_data.version,
       }),
+      // We should be using this property.
```
kumar303: I think it would be better to put a comment up above where `theme_data` gets merged into `addon` saying "use addon.themeData[themeProp] instead of addon[themeProp]"

tests/unit/core/reducers/test_addons.js
```diff
@@ -85,10 +85,11 @@ describe(__filename, () => {
     const state = addons(undefined,
       loadAddons(createFetchAddonResult(theme).entities));
 
-    const expectedTheme = {
+    const expectedTheme = createInternalAddon({
```
kumar303: This should not use `createInternalAddon()`. This test is a bit weird but the intention of it was to manually recreate the theme addon to test that the mapper is doing what we expect it to. If you use `createInternalAddon()` then this test is pointless since it relies on that for the mapping. Instead of manually creating the entire theme addon, the test was using some spreads for the mapped properties that exactly match the original. Maybe it needs some comments to clarify this?

tests/unit/core/reducers/test_search.js
```diff
@@ -66,13 +67,19 @@ describe('search reducer', () => {
 
     it('sets the results', () => {
       const { results } = getNextState();
-      expect(results).toEqual([{ slug: 'foo' }, { slug: 'food' }]);
+      expect(results).toEqual([
+        createInternalAddon({ slug: 'foo' }),
```
kumar303: I won't mark each one but a couple more of these need to change to `{ ...fakeAddon, slug: '...' }`

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -185,6 +186,7 @@ describe(__filename, () => {
           windows: undefined,
         },
         isRestartRequired: false,
+        themeData,
```
kumar303: ugh, this is an old test and it's not very useful. It may be worth just using `createInternalAddon({ ...fakeDiscoAddon, ... })` for the assertions here.

willdurand: Looks like we are not using the internal addon representation here, or I don't understand how it works right now :/

willdurand: ah, should be ok now.

willdurand:
@kumar303 updated the code, thanks for your comments! ready for a second look.

Issue #3079, opened 2017-09-07
Layout issues on iOS Firefox & Safari


wagnerand:
### Describe the problem and steps to reproduce it:

Filing in one single issue since they seem closely related.
STR:
1. https://addons.mozilla.org/en-US/firefox/addon/ubo-scope/

### What happened?

The blue fox logo, 'Explore' and 'Themes' are not where they should be.
On Firefox, they even move a bit as soon as you scroll down.
On Safari, they don't.

Also, the description is cut, at least on Safari.

### What did you expect to happen?

Correct placement and no moving...

### Safari:
![img_4346](https://user-images.githubusercontent.com/873258/30184278-dd92ff72-941d-11e7-9ee8-13067499f50b.PNG)

### Firefox, top of the page:
![img_4347](https://user-images.githubusercontent.com/873258/30184280-dd94da4a-941d-11e7-9125-6e6c999ef480.PNG)

### Firefox, scrolled down a bit:
![img_4348](https://user-images.githubusercontent.com/873258/30184279-dd9485ea-941d-11e7-9e27-3f36e4f2a851.PNG)

ExE-Boss:
On Firefox for iOS version 9.0 on iOS 11, the layout looks like this:

## Vertical: (Top bar redacted for privacy)
![Firefox for iOS 9.0 on iOS 11 – AMO (Vertical)](https://user-images.githubusercontent.com/3889017/30675015-49f841da-9e7e-11e7-88aa-427683f5a77b.png)

### Zoomed out:
![Firefox for iOS 9.0 on iOS 11 – AMO (Vertical, zoomed out)](https://user-images.githubusercontent.com/3889017/30675011-44146014-9e7e-11e7-8f88-f72dfd2550f1.png)

## Horizontal:
![Firefox for iOS 9.0 on iOS 11 – AMO (Horizontal)](https://user-images.githubusercontent.com/3889017/30675033-63026750-9e7e-11e7-900a-abd0bfdbd4ca.png)

### Zoomed out:
![Firefox for iOS 9.0 on iOS 11 – AMO (Horizontal, zoomed out)](https://user-images.githubusercontent.com/3889017/30675034-6305aa64-9e7e-11e7-8518-e189ff8c2894.png)

wagnerand:
I would recommend opening a new issue.

muffinresearch:
This looks a lot better today. I don't think there's anything else to do.

ioanarusiczki:
[...]

PR #3331, opened 2017-10-03
Fix header on iOS 11

ExE-Boss:
This fixes the header on iOS 11 by setting `max-width` to the display width.

## Issue checklist

* [x] This PR relates to an existing open issue and there are no existing
      PRs open for the same issue.
* [ ] Add `Fixes #ISSUENUM` at the top of your PR. (#3079, only fixes iOS 11, not iOS 10)
* [x] Add a description of the the changes introduced in this PR.
* [ ] The change has been successfully run locally. (I can’t run `yarn amo` for some reason, I use Windows 10)
* [ ] Add tests to cover the changes added in this PR.
* [ ] Add before and after screenshots (Only for changes that impact the UI). (Before available at https://github.com/mozilla/addons-frontend/issues/3079#issuecomment-331026857, after unavailable because of `yarn` issues)

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -8,6 +8,7 @@
   background: $ink-90;
   display: grid;
   grid-template-columns: min-content auto;
+  max-width: 100vw;
   min-height: 112px;
   padding: 0 12px;
 


ExE-Boss:
This has already been mostly fixed by #3498 and #3502. Mostly because there is still the occasional issue when going back a page, but it gets immediately fixed once you scroll, which is unfortunately not fixed by this or the aforementioned PRs.

Issue #3086, opened 2017-09-08
Update Category Page


muffinresearch:
Category pages currently look like:

![screen shot 2017-09-08 at 18 32 42](https://user-images.githubusercontent.com/1514/30223783-32c836a0-94c4-11e7-80aa-f81d81e76ee8.png)

But the mocks have:

<img width="794" alt="amo_desktop_-_master_sketch" src="https://user-images.githubusercontent.com/1514/30223768-29c94328-94c4-11e7-8726-0c833e5cab3b.png">

muffinresearch:
Based on the mock it's unclear what a lot of this data should be. @pwalm do we know what the shelves should be? I would have thought the current solution of using a search is probably more accurate for representing the content of a category in which case it would make sense to close this and action https://github.com/mozilla/addons-frontend/issues/3053 instead

tofumatt:
Ah, shoot, those specs are new.

pwalm:
I'm thinking the shelves will have similar data to what category pages have now:
- Featured
- Most Popular
- Top Rated
- Recently Added

Not sure what a "search widget" is, though? What would that add to the page?

muffinresearch:
Following a review with @pwalm this should not be a search results page but instead should follow the data we have on the existing category pages on the legacy frontend:

<img width="1082" alt="alerts___updates____add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30423904-379931c6-993c-11e7-845d-22e02ade457d.png">

Recently added was one that was OK to drop: From the scope doc "Top add-on lists: Featured, Top Rated, Most Popular".

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7).
All categories pages are changed and using the new form.
Postfix screenshots:
![addons](https://user-images.githubusercontent.com/15685960/31280653-ba9fd64e-aab5-11e7-86b1-5f0bdea01ffa.png)
![themes](https://user-images.githubusercontent.com/15685960/31280658-bd957d22-aab5-11e7-8a4b-36215a3bd32d.png)

Issues: https://github.com/mozilla/addons-frontend/issues/1952 and https://github.com/mozilla/addons-frontend/issues/3169 is happening in each category.

PR #3343, opened 2017-10-04
Add reducer/saga logic for new category page

willdurand:
Refs #3086

---

This PR adds the logic required to build the new category pages. Please see #3323 for the UI.

### DIFF
--- a/src/amo/sagas/landing.js
+++ b/src/amo/sagas/landing.js
@@ -10,20 +10,29 @@ import { LANDING_PAGE_ADDON_COUNT } from 'amo/constants';
 import { featured as featuredApi } from 'core/api';
 import { search as searchApi } from 'core/api/search';
 import {
-  LANDING_GET, SEARCH_SORT_POPULAR, SEARCH_SORT_TOP_RATED,
+  LANDING_GET,
+  SEARCH_SORT_POPULAR,
+  SEARCH_SORT_TOP_RATED,
 } from 'core/constants';
 import log from 'core/logger';
 import { createErrorHandler, getState } from 'core/sagas/utils';
 
 
 export function* fetchLandingAddons(
-  { payload: { errorHandlerId, addonType } }
+  { payload: { addonType, category, errorHandlerId } }
 ) {
   const errorHandler = createErrorHandler(errorHandlerId);
   try {
     const state = yield select(getState);
     const { api } = state;
-    const filters = { addonType, page_size: LANDING_PAGE_ADDON_COUNT };
+    const filters = {
+      addonType,
+      page_size: LANDING_PAGE_ADDON_COUNT,
+    };
+
+    if (category !== null) {
+      filters.category = category;
+    }
 
     const [featured, highlyRated, popular] = yield all([
       call(featuredApi, { api, filters }),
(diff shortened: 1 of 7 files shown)


tests/unit/amo/actions/test_landing.js
```diff
@@ -1,84 +1,125 @@
...
-  });
-  const action = getLanding(getActionParams());
-
-  it('sets the type', () => {
-    expect(action.type).toEqual('LANDING_GET');
-  });
-
-  it('sets the filters', () => {
-    expect(action.payload).toEqual({
+describe(__filename, () => {
```
willdurand: I have added a new `describe` block with `__filename`, that is why there is a larger diff than expected in this file.

willdurand:
So I have extracted this PR to reduce the diff of the UI PR (#3323).

tests/unit/amo/reducers/testLanding.js
```diff
@@ -138,4 +145,25 @@ describe(__filename, () => {
       expect(highlyRated).toEqual('hello');
     });
   });
+
+  describe('GET_LANDING_FOR_CATEGORY', () => {
+    it('indicates when fetching landing data with category', () => {
```
tofumatt: Indicates what? I'm not exactly sure what this spec means.

tests/unit/amo/sagas/testLanding.js
```diff
@@ -116,4 +122,116 @@ describe('amo/sagas/landing', () => {
...
+    let mockSearchApi;
+    let sagaTester;
+
+    beforeEach(() => {
+      errorHandler = createStubErrorHandler();
+      mockApi = sinon.mock(api);
+      mockSearchApi = sinon.mock(searchApi);
+      sagaTester = new SagaTester({
+        initialState: dispatchClientMetadata().state,
+        reducers: { landing: landingReducer, api: apiReducer },
```
tofumatt: You did https://github.com/mozilla/addons-frontend/pull/3343/files#diff-c339389f32102f7b6fc32a1dd8062aabR29 which I liked but then here it's on one line and not even sorted. 😅

willdurand:
> This works, but I wonder about just sending an extra filter to the LANDING_GET action. category is a vaild filter, so I don't see why we have a new action/etc. for this if you could instead just add category as an argument to fetchLandingAddons.

works for me. I don't remember why I changed, but initial work only added a new filter param indeed.

willdurand:
@tofumatt hi, sorry for that, it should be ready for another look

src/amo/sagas/landing.js
```diff
@@ -10,20 +10,29 @@ import { LANDING_PAGE_ADDON_COUNT } from 'amo/constants';
...
   try {
     const state = yield select(getState);
     const { api } = state;
-    const filters = { addonType, page_size: LANDING_PAGE_ADDON_COUNT };
+    const filters = {
+      addonType,
+      page_size: LANDING_PAGE_ADDON_COUNT,
+    };
+
+    if (category !== null) {
```
tofumatt: Maybe I'm missing it (the diff is a bit tough to read because of the re-org, but that's nice!), but I don't see a test for this. If the category is false-y at all, really, it shouldn't be set, and we should make sure it's not passed as a filter in that case.

willdurand: I have updated the condition. There are two tests, one with the category specified, the other without. I have added a new test to cover the case where `category` is given with `undefined` as value.

PR #3351, opened 2017-10-05
fix: Render carousel on client to fix CSP errors (fix #3349)

### LINKED ISSUE #3349: Homepage carousel fails CSP
The homepage fails to load with CSP enabled on first page load if you are loading the home page (eg. https://addons-dev.allizom.org/en-GB/firefox/). This results in the homepage looking messed up:

screenshot 2017-10-04 22 01 48

Navigate to the homepage from another page and it's fine:

screenshot 2017-10-04 22 02 10

This is because CSP is blocking it from loading a `<style>` tag.

Possibly related to https://github.com/FormidableLabs/nuka-carousel/issues/182. Hopefully an easy fix!

tofumatt:
fix #3349 

No need for two reviews, just whoever gets to it first (feel free to clear the other person once you start reviewing it).

This fixes the CSP error on the carousel by rendering it on the client after `componentDidMount` runs. It's not lovely but it looks nice enough without huge jumps and it works. Pretty much any carousel, by their nature modifying style props, would require this 😢

### DIFF
--- a/src/ui/components/Carousel/index.js
+++ b/src/ui/components/Carousel/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import classNames from 'classnames';
 import React from 'react';
 import { compose } from 'redux';
 import NukaCarousel from 'nuka-carousel';
@@ -54,7 +55,11 @@ export class CarouselBase extends React.Component {
     }
 
     return (
-      <Card className="Carousel">
+      <Card
+        className={classNames('Carousel', {
+          'Carousel--first-render-wrapper': !this.state.hasRendered,
+        })}
+      >
         {this.state.hasRendered ? (
           <NukaCarousel
             autoplay
@@ -72,7 +77,7 @@ export class CarouselBase extends React.Component {
             }))}
           </NukaCarousel>
         ) : (
-          <div className="Carousel--server-render">
+          <div className="Carousel--first-render">
             {/*
               We render the sections here to keep the server-side content
               similar to the client-side content so the re-render isn't as
--- a/src/ui/components/Carousel/index.js
+++ b/src/ui/components/Carousel/index.js
@@ -1,7 +1,8 @@
 /* @flow */
-import NukaCarousel from 'nuka-carousel';
+import defaultConfig from 'config';
 import React from 'react';
 import { compose } from 'redux';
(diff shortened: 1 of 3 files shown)


tofumatt:
Note: I tested this using a production build with CSP enabled. You can too by following: https://github.com/mozilla/addons-frontend#building-and-running-services

willdurand:
I am reviewing this PR and I cannot get the expected result it seems 😞 I have followed the instructions in the README but the carousel does not render correctly on page hit, see the Gif where I have to navigate to another page before getting it right:

willdurand:
Mmmh the HTML sent by the server does not contain any `Carousel` HTML:

screen shot 2017-10-05 at 10 38 25

willdurand:
The HTML sent by the server does not have any `Carousel` HTML in it:

screen shot 2017-10-05 at 10 38 25

tofumatt:
Okay, rebased onto master and all set for review. Sorry I mucked this up.

src/ui/components/Carousel/index.js
```diff
@@ -16,31 +18,80 @@ type PropTypes = {|
...
-        }))}
-      </NukaCarousel>
-    </Card>
-  );
-};
+        {this.state.hasRendered ? (
+          <NukaCarousel
+            autoplay
+            autoplayInterval={4000}
+            cellAlign={getDirection(i18n.lang) === 'ltr' ? 'left' : 'right'}
```
willdurand: Didn't see this before, but shouldn't we have constants for that? It would avoid case issues. If so, then I could open a PR because it is unrelated.

src/ui/components/Carousel/styles.scss
```diff
@@ -48,6 +48,38 @@ $extraHeightForCard: 28px;
   }
 }
 
+.Carousel--first-render-wrapper .Card-contents {
+  // HACK: This fixes RTL glitches in the NukaCarousel we will eventually
+  // fix properly. This means te content in RTL doesn't change order after
+  // the client re-render.
+  direction: ltr;
```
willdurand: 👍

src/ui/components/Carousel/index.js
willdurand: See 3362

PR #3356, opened 2017-10-05
Remove View Classic Desktop Site from footer

### LINKED ISSUE #3352: Remove "View Classic Desktop Site" from footer
If we don't provide an escape hatch to the old Django site we should remove the "View Classic Desktop Site" link/action from the footer.

To be clear I think this is a good idea 👍

tsl143:
Fixes #3352 

Also updated the test cases with enzyme approach.

Before
screen shot 2017-10-05 at 4 06 56 pm

After
screen shot 2017-10-05 at 4 06 56 pm

### DIFF
--- a/src/amo/components/Footer/index.js
+++ b/src/amo/components/Footer/index.js
@@ -13,13 +13,12 @@ import './styles.scss';
 
 export class FooterBase extends React.Component {
   static propTypes = {
-    handleViewDesktop: PropTypes.func.isRequired,
     i18n: PropTypes.object.isRequired,
     location: PropTypes.object.isRequired,
   }
 
   render() {
-    const { handleViewDesktop, i18n, location } = this.props;
+    const { i18n, location } = this.props;
     const homepageText = i18n.gettext("Go to Mozilla's homepage");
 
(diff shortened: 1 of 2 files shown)


willdurand:
@tsl143 you should rebase I think.

tsl143:
Done, got conflict due to my commit itself :D

willdurand:
> Done, got conflict due to my commit itself :D

Yeah. Note that I see a merge rather than a rebase here :) You should only have one commit, but that does not matter much.

muffinresearch:
As a heads-up - having discussed this with Jorge last week, we're going to only remove this post-launch when all the remaining pages are completed.

tofumatt:
Hey @tsl143,

Thanks for your work on this patch, which was solid but apparently came too soon. It's a good basis for a future patch (I especially like the update to enzyme 👍), but I'm gonna close this as we clearly aren't ready to merge this and the patch has since developed a few conflicts.

I'm gonna mark the issue as one not to touch as well, because if after six months we don't want to close it I think it's better to just file again once we're *actually* ready for it.

Issue #3309, opened 2017-10-02
Add collection based shelf to Homepage [2/5]


muffinresearch:
Breaking this out from #3152

Add a shelf of content on the homepage to feature a defined collection.

https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit#heading=h.mky7j18wsuga

For release the collection is going to be called: "Change your tabs" and points to  

https://addons.mozilla.org/firefox/collections/mozilla/change-up-your-tabs/

Not sure if there's specific UX for this  - I'd suggest starting with the same as the default shelf and  then we can specialize the look and feel separately.

E.g see this look and feel for  Most Popular content: 

<img width="1433" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/31087670-d24c2040-a795-11e7-8183-c19f2764d663.png">

Order - based  on the doc this is shelf no.2

muffinresearch:
Marked as UX because we  could use explicit UX for all the homepage shelves.

willdurand:
@muffinresearch this collection is not available in dev or stage, so don't know how QA will be able to test it. How can we add it to those environments?

ValentinaPC:
Verified as fixed on AMO-dev and AMO-stage using FF56(Win 7, Android 7.2.1).
Postfix screenshot:
![2017-10-09_1356](https://user-images.githubusercontent.com/15685960/31335262-74ccacb8-acfa-11e7-8c4b-d5a67c433b24.png)
![2017-10-09_1402](https://user-images.githubusercontent.com/15685960/31335280-84240c10-acfa-11e7-89a4-3f951c83f0af.png)

willdurand:
@ValentinaPC sorry, the issue description is a bit misleading. You should look for a second shelf. The one you highlight here is the first one (see #3308).

ValentinaPC:
The collection is listed on AMO-dev FF56 (Win 7), but not on Android.
But, if we click on "Browse this collection" link and refresh the loaded page, user is redirected to the old AMO coll page.
Please see video:
![refresh](https://user-images.githubusercontent.com/15685960/31336395-5abbe0f0-acff-11e7-8379-51c884c04505.gif)

Reopening.

willdurand:
ah right. I think this is an Nginx (configuration) issue.

willdurand:
This is now available in dev.

ValentinaPC:
[...]

PR #3365, opened 2017-10-05
Add most popular extensions shelf to the homepage

### LINKED ISSUE #3308: Add most popular extensions shelf to the homepage [1/5]
Breaking this out from #3152

Add a shelf  of content  on the homepage to  feature  "Most Popular" content.

https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit#heading=h.mnb8tk6s8el0

Not sure if there's specific UX for this see below for what's in the mock - I'd suggest starting with the same as the default shelf and then we can specialize the look and feel separately.

**NOTE: We already have this content on the extensions landing page:**

add-ons_for_firefox

Order: based on the doc  this is the first  shelf under the carousel.

willdurand:
Fix #3308

---

This PR adds the "most popular extensions" shelf to the homepage. It also adds all we need to add the other shelves. I have created a new reducer/saga to be able to fetch all the data in one round.

For example: here is the code to add the second shelf (#3309): https://gist.github.com/willdurand/88a24904de4e6b2567fb2e5ab7f63676 (no test and I used `happyholidays` because dev/stage does not have the required collection).

## Screenshot

screen shot 2017-10-05 at 18 00 51

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -6,12 +6,17 @@ import { compose } from 'redux';
 
 import { setViewContext } from 'amo/actions/viewContext';
 import HomeCarousel from 'amo/components/HomeCarousel';
+import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import Link from 'amo/components/Link';
+import { fetchHomeAddons } from 'amo/reducers/home';
 import {
+  ADDON_TYPE_EXTENSION,
   CLIENT_APP_ANDROID,
   CLIENT_APP_FIREFOX,
+  SEARCH_SORT_POPULAR,
   VIEW_CONTEXT_HOME,
 } from 'core/constants';
+import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import Card from 'ui/components/Card';
 
@@ -27,6 +32,7 @@ export const CategoryLink = ({ children, name, slug, type }) => {
     </li>
   );
 };
+
 CategoryLink.propTypes = {
   children: PropTypes.node.isRequired,
   name: PropTypes.string.isRequired,
@@ -46,13 +52,19 @@ export class HomeBase extends React.Component {
   static propTypes = {
     clientApp: PropTypes.string.isRequired,
     dispatch: PropTypes.func.isRequired,
+    errorHandler: PropTypes.object.isRequired,
     i18n: PropTypes.object.isRequired,
+    popularExtensions: PropTypes.array.isRequired,
   }
 
   componentWillMount() {
-    const { dispatch } = this.props;
+    const { dispatch, errorHandler, popularExtensions } = this.props;
 
     dispatch(setViewContext(VIEW_CONTEXT_HOME));
+
+    if (popularExtensions.length === 0) {
+      dispatch(fetchHomeAddons({ errorHandlerId: errorHandler.id }));
+    }
   }
 
   extensionsCategoriesForClientApp() {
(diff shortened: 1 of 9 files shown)


willdurand:
Should be ready for a first look, first in, first served, thanks!

tofumatt:
I've got it. I'm happy to do some reviews while I watch the push 😄

tests/unit/amo/containers/TestHome.js
```diff
@@ -104,8 +134,57 @@ describe('Home', () => {
...
+      pathname: '/search/',
+      query: {
+        addonType: ADDON_TYPE_EXTENSION,
+        sort: SEARCH_SORT_POPULAR,
+      },
+    });
+    expect(shelf).toHaveProp('loading', true);
+  });
+
+  it('dispatches an action to fetch the add-ons to display', () => {
```
tofumatt: Should we have another test to ensure it renders the content and doesn't set loading once the add-ons have been dispatched?

tests/unit/amo/reducers/test_home.js
```diff
@@ -0,0 +1,70 @@
+import reducer, {
```
tofumatt: I think labelling this `homeReducer` will improve the test's readability a lot. We do that elsewhere, I think it's worth doing here too.

tests/unit/amo/containers/TestHome.js
willdurand: indeed, added new assertions for that.

src/amo/components/Home/index.js
```diff
@@ -46,13 +52,19 @@ export class HomeBase extends React.Component {
...
+    popularExtensions: PropTypes.array.isRequired,
   }
 
   componentWillMount() {
-    const { dispatch } = this.props;
+    const { dispatch, errorHandler, popularExtensions } = this.props;
 
     dispatch(setViewContext(VIEW_CONTEXT_HOME));
+
+    if (popularExtensions.length === 0) {
```
kumar303: We should get out of the habit of using empty arrays to indicate that data has not yet loaded. It is possible to receive an empty array as a legitimate API result and in that case your code will get stuck in an infinite dispatch loop. It's unlikely for the popular extensions API to return an empty array but when using a local development server that isn't configured right, it could happen. It would be safer to do this:

```js
if (popularExtensions === null) {
  dispatch(fetchHomeAddons(...));
}
```

And then you'd need to initialize the reducer like that too:

```js
export const initialState: HomeState = {
  popularExtensions: null,
};
```

src/amo/components/Home/index.js
```diff
@@ -129,12 +141,27 @@ export class HomeBase extends React.Component {
   }
 
   render() {
-    const { i18n } = this.props;
+    const { i18n, popularExtensions } = this.props;
 
     return (
       <div className="Home">
```
kumar303: This needs `{errorHandler.renderErrorIfPresent()}`

tofumatt: Shoot, I noticed that too and must've forgot to hit submit on the comment 😭

src/amo/components/Home/index.js
tofumatt: That's a good point, and we probably do this elsewhere. I've filed #3372.

willdurand: Indeed. For my defense, I did it on purpose because there are always popular extensions (in theory) and I removed it in the upcoming PR about the 2nd shelf in favor of a `resultsLoaded` flag.

willdurand: That being said, I should not have done that.

src/amo/components/Home/index.js
willdurand: ooops, will add it in the next PR, sorry for that I forgot it.

PR #3367, opened 2017-10-05
fix: Reduce asset size (close #3366)

### LINKED ISSUE #3366: Carouesl assets are gigantic
I naively pulled and committed very large images to the carousel component. We should compress them–@muffinresearch pointed out right now they're like 18MB.

shame-cube

tofumatt:
close #3366


tofumatt:
Maybe as a build step? I ran them through Pixelmator's compress for web. `jpegoptim` didn't seem to make a huge diff, but I did it, why not? 😄 
screenshot 2017-10-05 17 50 50

muffinresearch:
Nice - much better 👍

Props for the quick turnaround.

PR #3369, opened 2017-10-05
Add collection based shelf to Homepage

### LINKED ISSUE #3309: Add collection based shelf to Homepage [2/5]

willdurand:
Fix #3309

---

This PR adds the featured collection shelf. It adds a test for the missing error rendering mentioned in https://github.com/mozilla/addons-frontend/pull/3365#discussion_r143070227.

⚠️ this collection does not exist in stage/dev 😢  I manually changed to `happy-2015` to test it, but I dont know what else to do. It could maybe part of the `config` (since it has different configs depending on the envs) but I did not see any similar config data...

## (Fake) Screenshot

screen shot 2017-10-06 at 15 06 34

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -166,7 +166,7 @@ export class HomeBase extends React.Component {
 
         <LandingAddonsCard
           addons={popularExtensions}
-          className="PopularExtensions"
+          className="Home-PopularExtensions"
           header={i18n.gettext('Most popular extensions')}
           footerText={i18n.gettext('More popular extensions')}
           footerLink={{
@@ -181,7 +181,7 @@ export class HomeBase extends React.Component {
 
         <LandingAddonsCard
           addons={featuredCollection}
-          className="FeaturedCollection"
+          className="Home-FeaturedCollection"
           header={i18n.gettext('Change your tabs')}
           footerText={i18n.gettext('Browse this collection')}
           footerLink={{ pathname:
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -48,22 +48,31 @@ export const ThemeLink = (props) => {
   return <CategoryLink type="themes" {...props} />;
 };
 
+export const FEATURED_COLLECTION_SLUG = 'change-up-your-tabs';
+export const FEATURED_COLLECTION_USER = 'mozilla';
+
 export class HomeBase extends React.Component {
   static propTypes = {
     clientApp: PropTypes.string.isRequired,
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
+    featuredCollection: PropTypes.array.isRequired,
     i18n: PropTypes.object.isRequired,
     popularExtensions: PropTypes.array.isRequired,
+    resultsLoaded: PropTypes.bool.isRequired,
   }
 
   componentWillMount() {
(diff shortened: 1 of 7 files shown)


src/amo/reducers/home.js
```diff
@@ -80,14 +106,29 @@ const reducer = (
...
-      const { popularExtensions } = action.payload;
+      const {
+        featuredCollection,
+        popularExtensions,
+      } = action.payload;
 
       return {
         ...state,
+        featuredCollection: featuredCollection.results
+          .slice(0, LANDING_PAGE_ADDON_COUNT)
```
willdurand: I do this here because the [Collections API](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#list) does not allow to specify the number of results we want.

tofumatt: Works but sounds like a bug worth filing: we could probably cut down on the response data size if we could send a `page_size` query.

src/amo/reducers/home.js
```diff
@@ -80,14 +106,29 @@ const reducer = (
...
+          .map((item) => (
+            createInternalAddon(item.addon)
```
tofumatt: If this is already on a newline and wrapped in something I'd say typing out return makes this a lot more scannable.

I think this should either be:

```js
.map((item) => createInternalAddon(item.addon))
```
or
```js
.map((item) => {
  return createInternalAddon(item.addon);
})
```

If our lint rules are to blame we should update them; this seems like the worst of both worlds. Though to be clear I prefer the second, more explicit and scannable one 😄

src/amo/components/Home/index.js
```diff
@@ -159,7 +176,18 @@ export class HomeBase extends React.Component {
               sort: SEARCH_SORT_POPULAR,
             },
           }}
-          loading={popularExtensions === null}
+          loading={resultsLoaded === false}
+        />
+
+        <LandingAddonsCard
+          addons={featuredCollection}
+          className="FeaturedCollection"
```
tofumatt: We'll often prefix a classname of a child component with the parent component's name, so this should be `Home-FeaturedCollection`.

tofumatt:
Oh and regarding this collection not being on stage and -dev: I would write up instructions for QA In the original issue to create this collection. It’s the best we can do for now until we improve data on dev and stage.

willdurand:
> Oh and regarding this collection not being on stage and -dev: I would write up instructions for QA In the original issue to create this collection. It’s the best we can do for now until we improve data on dev and stage.

I have talked to @muffinresearch today and he managed to create the collection in dev (and stage soon).

tofumatt:
Ah cool. Sorry I didn’t see that. 

Matthew Riley MacPherson

PR #3373, opened 2017-10-06
fix: Don't rely on empty arrays as being unloaded

### LINKED ISSUE #3372: Don't rely on empty arrays as being unloaded
As noted in https://github.com/mozilla/addons-frontend/pull/3365#discussion_r143069482, we sometimes treat an empty array as one that hasn't loaded yet, but strictly speaking this isn't true. We could get a legitimate empty response from the server that would keep us constantly dispatching.

We should follow @kumar303's idea in the linked comment and differentiate between unloaded and loaded but empty.

tsl143:
Fixes #3372 
There were few other array length checks such as in `amo/api/review.js` but those were being returned from API as an array so `=== null` check wouldn't work.

### DIFF
--- a/src/amo/reducers/home.js
+++ b/src/amo/reducers/home.js
@@ -6,7 +6,7 @@ export const FETCH_HOME_ADDONS: 'FETCH_HOME_ADDONS' = 'FETCH_HOME_ADDONS';
 export const LOAD_HOME_ADDONS: 'LOAD_HOME_ADDONS' = 'LOAD_HOME_ADDONS';
 
 export type HomeState = {
-  popularExtensions: ?Array<AddonType>,
+  popularExtensions: Array<AddonType>| null,
 };
 
 export const initialState: HomeState = {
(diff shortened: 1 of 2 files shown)


src/amo/reducers/home.js
```diff
@@ -6,11 +6,11 @@ export const FETCH_HOME_ADDONS: 'FETCH_HOME_ADDONS' = 'FETCH_HOME_ADDONS';
 export const LOAD_HOME_ADDONS: 'LOAD_HOME_ADDONS' = 'LOAD_HOME_ADDONS';
 
 export type HomeState = {
-  popularExtensions: Array<AddonType>,
+  popularExtensions: ?Array<AddonType>,
```
tofumatt: I think you could also write ` | null` which is a bit clearer.

tsl143:
> Not sure what you meant by the api returning an array meaning we can’t check, but this is largely about our default reducer state. So if there’s anywhere else we set an array of items to zero in the reducer we should change them to null...

Got it.

willdurand:
@tsl143 cool stuff, I did that in my other PR too, but if you're done with this one, let's merge it and I'll base my work on yours. (note that this will disappear 😂 )

tsl143:
@willdurand I updated this already :)

willdurand:
@tsl143 no other place were we do this?

tsl143:
> no other place were we do this?

We use `.length` compare in `amo/api/review.js` but that uses api directly, not by reducer.

willdurand:
I see, thanks!

PR #3384, opened 2017-10-06
Use semantically transparent element for icons

### LINKED ISSUE #3376: Accessibility issue on using 'i' tags for icons
Just found that the project has been using `i` tags for using icons. But that doesn't comply with accessibility standards. `i` tags should be used only to make a text italic. May be we could replace `i`  with `span` tags all over the project. Wont be needing any difference in styling but also improves the accessibility of AMO. what do you think?

muffinresearch:
Fixes #3376

### DIFF
--- a/src/ui/components/Icon/index.js
+++ b/src/ui/components/Icon/index.js
@@ -23,7 +23,7 @@ export default class Icon extends React.Component {
     }
 
     return (
-      <i className={classNames('Icon', `Icon-${name}`, className)} {...props} />
+      <span className={classNames('Icon', `Icon-${name}`, className)} {...props} />
     );
   }
 }


muffinresearch:
> r+, but do we use <Icon> inside any <span> elements already? I think we're technically not supposed to do that. But I suppose we're technically not supposed to use `<i>` either... though arguably `<i>` isn't semantic, just <em> is. But whatever!

spans can be nested that's not a problem. 

Whilst `i` was always historically used for presentational reasons it does have some meaning:

> The i element represents a span of text in an alternate voice or mood, or otherwise offset from the normal prose in a manner indicating a different quality of text, such as a taxonomic designation, a technical term, an idiomatic phrase from another language, a thought, or a ship name in Western texts.

From https://www.w3.org/TR/2011/WD-html5-author-20110809/the-i-element.html

tofumatt:
Huh, neato!

And cool, didn’t know nesting spans was okay!

PR #3386, opened 2017-10-06
Add an rg .ignore file for better searching

kumar303:


tofumatt:
So many dots! 😱

kumar303:
@willdurand form your rg experience is there anything else we need to add?

kumar303:
> So many dots!

heh, yeah. I considered removing `.ackrc` but maybe someone still needs ack for certain things.

willdurand:
I mostly use the default config since it relies on `.gitignore` already. I guess this is fine.

Issue #3370, opened 2017-10-05
Change style of reply-to-review form


kumar303:
To see this UI element:
* Sign in as a developer of an add-on
* Go to the detail page of your add-on
* Click to read all reviews
* Click Reply to this review
* Enter some text

This issue is to make UI changes based on the latest mocks, shown in Sketch and below:

<img width="1463" alt="screenshot 2017-10-05 16 33 44" src="https://user-images.githubusercontent.com/55398/31253556-0165a46c-a9eb-11e7-8871-95b0f317d152.png">

ValentinaPC:
@kumar303: according to mocks, it should be a preview reply option, but I cannot see it. It was removed?

tofumatt:
Yes, the Pull request mentioned it I think but maybe not the issue, whoops!

The preview was dropped for this issue but should be implemented later, if there isn’t an issue for it it should be filed.

- Matthew Riley MacPherson (sent from mobile)

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7, Android 7.2.1)
Reply to this review option is working as expected. Checked also using a RTL lang.
Postfix screenshots:
![2017-10-10_1305](https://user-images.githubusercontent.com/15685960/31381114-41eaebd0-adbc-11e7-874d-98b326ac115b.png)
![2017-10-10_1306](https://user-images.githubusercontent.com/15685960/31381117-46369946-adbc-11e7-8220-10c199302d91.png)

PR #3394, opened 2017-10-07
Update styles for the reply-to-review form

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3370

- I left the text-shadow in the buttons even though the mock doesn't have it
- I didn't implement the preview-reply part because that's a lot of effort--it can come later
- The mock uses `$grey-50` (I think?) for the Cancel button which is why I didn't go with the `$neutral-base-color`
- I didn't add the placeholder guidelines (tips for writing a review) since that needs some product guidance. I'm not sure what tips we should be giving.

screenshot 2017-10-09 09 50 19
screenshot 2017-10-09 09 50 01

screenshot 2017-10-06 18 54 58

This is what the centered buttons looked like (this change got reverted):

screenshot 2017-10-06 18 54 06

screenshot 2017-10-06 18 54 37


tofumatt:
The text shadow in the buttons is site-wide and from our use of pre-photon colours; without the shadow the buttons had accessibility problems because of their text size.

We should be able to fix that in #3398, which I've filed as a sort of tracker issue post-MVP.

src/ui/components/Button/Button.scss
```diff
@@ -30,6 +30,15 @@
   );
 }
 
+// This is intended for counter-actions like cancelling or undoing, etc.
+.Button--counter-action {
```
tofumatt: I get the notion that this is "counter" to the original action but I think that's a bit of unclear phrasing, especially as people might read this as a number "Counter" (which could also have a button).

"cancel" or "undo" would be a better name.

src/ui/components/DismissibleTextForm/styles.scss
```diff
@@ -1,30 +1,64 @@
...
   display: flex;
   justify-content: space-between;
+  margin-top: 6px;
+
+  @include respond-to(medium) {
+    justify-content: center;
+    margin-top: 24px;
+
+    .Button {
+      margin-left: 48px;
```
tofumatt: This should use `@include margin-start(48px);` so it works in RTL as well.

src/ui/components/DismissibleTextForm/styles.scss
```diff
@@ -1,30 +1,64 @@
...
+      padding: 0 64px;
```
tofumatt: Is this supposed to centre the buttons? In the mocks it looks like they should be centered... though I'd argue that's worse for usability. Hacing them be farther away from each other prevents accidental taps (mainly on touch devices) and having them be on the edges of the screen makes it easier to tap when on a phone/tablet.

I see in the mocks it's:
screenshot 2017-10-08 13 58 34

I also just think it looks a bit off. @pwalm: Could we align theses on the left/right instead of the centre?

src/ui/components/DismissibleTextForm/styles.scss
```diff
@@ -1,30 +1,64 @@
...
+
+      &:first-child {
+        margin-left: 0;
```
tofumatt: Should be `margin-start` as well.

src/ui/components/DismissibleTextForm/styles.scss
kumar303: This adds left and right padding to make the buttons wide like the mocks have them.

src/ui/components/Button/Button.scss
kumar303: I think counter-action is more versatile since it's generic enough to convey the concept of cover cancel, dismiss or undo but I can change it, no problem.

kumar303:
@tofumatt I think the edge buttons look weird on a wide screen but I'm not picky so I changed them for you and updated the screenshots. It is ready for another review.

src/ui/components/DismissibleTextForm/styles.scss
```diff
@@ -1,30 +1,58 @@
...
 .DismissibleTextForm-buttons {
   display: flex;
   justify-content: space-between;
+  margin-top: 6px;
+
+  @include respond-to(medium) {
+    margin-top: 24px;
+
+    .Button {
+      padding: 0 64px;
```
tofumatt: I see now this is for making the button bigger.

I think this should be a button option (something like `Button--large` similar to our existing `Button--small`). I know it's a pain but I have seen a lot of code bases with tiny variances on buttons because they are modified in a non-central place.

It's fine to do it as a follow-up bug if it's a pain here, but we should at least have it filed (and maybe linked to #3398) to have a large button style any component can use rather than individual components changing button styles. It's fine to align/position buttons here, but I think anything we do *inside* the button should be in the button style as an option.

kumar303: Good call, I addressed it. I was hesitating before only because I didn't know if we wanted to apply the style globally or not. Using a new button style is a safe approach because it won't be applied to any existing buttons.

Issue #3391, opened 2017-10-06
Ensure package dependency versions are exact


kumar303:
### Describe the problem and steps to reproduce it:

Add a new item to `dependencies` in `package.json` with a caret version like `^5.1.0`.

### What happened?

No CI tests failed.

### What did you expect to happen?

Our CI tests should fail because dependency versions need to be exact.

### Anything else we should know?

Without exact dependency versions, deploying to production might unintentionally upgrade a package.

PR #3397, opened 2017-10-08
fix: more menu layout issue

### LINKED ISSUE #3300: [RTL] 'More' menu layout appears sectioned on browser width <720
Steps to reproduce:

1. Go to AMO homepage
2. Select a RTL language (i.e. https://addons-dev.allizom.org/fa/firefox/)
3. Change page width to <720
4.  Open the 'More' menu

Expected result:
'More' menu is displayed without any layout issues

Actual result
'More' menu  appears sectioned on the right hand side

Notes:
Tested on FF56, Win10
Reproduces also on Android device in landscape mode

sectioned more mnu

tsl143:
Fixes #3300 
Before:
screen shot 2017-10-08 at 8 16 34 am

After
screen shot 2017-10-08 at 8 16 03 am

### DIFF
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -105,4 +105,8 @@
   .DropdownMenu--active & {
     display: block;
   }
+
+  @include respond-to(extraLarge) {
+    max-width: 300px;
+  }
 }
--- a/src/ui/components/DropdownMenu/styles.scss
+++ b/src/ui/components/DropdownMenu/styles.scss
@@ -81,6 +81,7 @@
   display: none;
   list-style-type: none;
   margin: 0;
+  max-width: 245px;
   padding: 4px 24px 20px;
   position: absolute;
   top: 30px;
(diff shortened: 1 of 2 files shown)


tsl143:
`discopane-ui-tests` fail for this, any specific reason?

tsl143:
I think this is due to #3391

tofumatt:
Sometimes the disco pane hi tests just fail/tie,out, unfortunately 😐

I’ve restarted them, maybe that will help! 😅

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -81,6 +81,7 @@
   display: none;
   list-style-type: none;
   margin: 0;
+  max-width: 245px;
```
tofumatt: My only thought is maybe this could be a bit wider on really large displays? Maybe for the extraLarge viewport size it could be 300?

And maybe even by default something like 260? I guess that’s a nitpick but just trying to limit when this would happen.

Or was 245 the biggest you could go to without errors?

tsl143: I got max 243 max without errors, so chose 245.
But adding 300 for `extraLarge` will be good too.

PR #3400, opened 2017-10-08
WIP: feat: Add an option to randomize carousels

### LINKED ISSUE #3332: Randomize header content order
Doing this as a follow-up bug, but after discussing with @devaneymoz and @pwalm, we're gonna randomize the header content by default.

tofumatt:
Adds an option to randomly sort a carousel's sections and enables it on the Home page carousel.

Fix #3332.

Randomly sorting things led to even more issues with the difference between server/client load, and I thus far have settled on thinking we really shouldn't load the carousel on the client because of the weirdness of it until the client render happens.

I think this might be the best solution but I'll think about it a bit more and tidy this up tomorrow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
     "jed": "1.1.1",
     "join-url": "2.0.0",
     "jsdom": "11.3.0",
+    "knuth-shuffle": "1.0.1",
     "localforage": "1.5.1",
     "lodash.debounce": "4.0.8",
     "moment": "2.18.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -171,6 +171,7 @@
     "jed": "1.1.1",
     "join-url": "2.0.0",
     "jsdom": "11.3.0",
+    "knuth-shuffle": "1.0.1",
```
tofumatt: Ensures a truly random sort, plus unit testing sort code is obnoxious.

muffinresearch:
@tofumatt can this be closed out in light of https://github.com/mozilla/addons-frontend/issues/3438?

tofumatt:
Yeah I meant to clean these issues up once the carousel was changed out, but this can be deleted now 😄

PR #3412, opened 2017-10-09
Fix landing page

### LINKED ISSUE #3383: Improve "More Featured Extensions/Themes" section
Steps to reproduce:
1. Load Themes homepage https://addons-dev.allizom.org/en-US/firefox/themes/
2. Go to a theme category that has no featured themes i.e. https://addons-dev.allizom.org/en-US/firefox/themes/websites/
3. Go back to Themes homepage
4. Observe "More Featured Themes" section

Expected results:
There are no displayed or layout issues.

Actual results:
The featured section is hidden.

Notes/Issues:
- section reappears after page refresh

**Enhancement suggestion for featured section** - do not show the more link if less then 4 featured add-ons available.

Screenshot for this issue:
featured

willdurand:
Fix #3383

---

This PR makes sure a landing page fetches data again after having loaded a category page. It happens because we use the same (redux) action with an optional parameter (`category`) and the `LandingPage` component needed to check it to determine if it had to get data or not. I also made sure that we did not dispatch `setViewContext` too often.

This diff is rather large because I rewrote most of the test file with `shallowUntilTarget()`.

### DIFF
--- a/src/amo/components/LandingPage/index.js
+++ b/src/amo/components/LandingPage/index.js
@@ -41,6 +41,8 @@ export class LandingPageBase extends React.Component {
     // `componentWillReceiveProps()`.
     // eslint-disable-next-line react/no-unused-prop-types
     addonTypeOfResults: PropTypes.string.isRequired,
+    // This is a bug; context is used in `setViewContextType()`.
+    // eslint-disable-next-line react/no-unused-prop-types
     context: PropTypes.string.isRequired,
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
@@ -85,6 +87,7 @@ export class LandingPageBase extends React.Component {
   getLandingDataIfNeeded(nextProps = {}) {
     const {
       addonTypeOfResults,
+      dispatch,
       errorHandler,
       loading,
       params,
@@ -97,8 +100,8 @@ export class LandingPageBase extends React.Component {
     const requestedAddonType = apiAddonType(params.visibleAddonType);
 
     if (!loading && !errorHandler.hasError() &&
-      (!resultsLoaded || addonTypeOfResults !== requestedAddonType)) {
-      this.props.dispatch(getLanding({
+        (!resultsLoaded || addonTypeOfResults !== requestedAddonType)) {
+      dispatch(getLanding({
         addonType: requestedAddonType,
         errorHandlerId: errorHandler.id,
       }));
@@ -106,10 +109,10 @@ export class LandingPageBase extends React.Component {
   }
 
   setViewContextType(nextProps = {}) {
-    const { params } = { ...this.props, ...nextProps };
+    const { context, params } = { ...this.props, ...nextProps };
     const addonType = apiAddonType(params.visibleAddonType);
 
-    if (this.props.context !== addonType) {
+    if (context !== addonType) {
       this.props.dispatch(setViewContext(addonType));
     }
   }
@@ -276,7 +279,7 @@ export function mapStateToProps(state) {
 }
 
 export default compose(
-  withErrorHandler({ name: 'LandingPage' }),
   connect(mapStateToProps),
-  translate({ withRef: true }),
+  translate(),
+  withErrorHandler({ name: 'LandingPage' }),
 )(LandingPageBase);
--- a/src/amo/components/LandingPage/index.js
+++ b/src/amo/components/LandingPage/index.js
@@ -37,17 +37,22 @@ const ICON_MAP = {
 
 export class LandingPageBase extends React.Component {
   static propTypes = {
-    addonTypeOfResults: PropTypes.string,
+    // This is a bug; addonTypeOfResults is used in
+    // `componentWillReceiveProps()`.
+    // eslint-disable-next-line react/no-unused-prop-types
+    addonTypeOfResults: PropTypes.string.isRequired,
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
-    featuredAddons: PropTypes.array,
-    highlyRatedAddons: PropTypes.array,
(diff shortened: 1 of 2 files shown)


src/amo/components/LandingPage/index.js
```diff
@@ -37,17 +37,23 @@ const ICON_MAP = {
 
 export class LandingPageBase extends React.Component {
   static propTypes = {
-    addonTypeOfResults: PropTypes.string,
+    // This is a bug; addonTypeOfResults is used in
+    // `componentWillReceiveProps()`.
+    // eslint-disable-next-line react/no-unused-prop-types
```
kumar303: I say disable it at the top -- it's annoying and full of bugs! If you search for components that disable it, you'll find a lot.

src/amo/components/LandingPage/index.js
```diff
@@ -63,43 +69,49 @@ export class LandingPageBase extends React.Component {
...
   }
 
-  setViewContextType() {
-    const { dispatch, params } = this.props;
+  setViewContextType(nextProps = {}) {
+    const { params } = { ...this.props, ...nextProps };
     const addonType = apiAddonType(params.visibleAddonType);
-    dispatch(setViewContext(addonType));
+
+    if (this.props.context !== addonType) {
```
kumar303: Should `context` be fetched more like this?

```js
const props = { ...this.props, ...nextProps };
if (props.context ...) {
}
```

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -15,27 +18,40 @@ import {
...
-describe('<LandingPage />', () => {
-  function renderProps(props = {}) {
+describe(__filename, () => {
+  function renderProps({
+    store = dispatchClientMetadata().store,
+    ...otherProps
+  } = {}) {
     return {
-      dispatch: sinon.stub(),
+      dispatch: store.dispatch,
```
kumar303: You can delete this. It gets set automatically by `mapDispatchToProps()`.

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -68,97 +82,85 @@ describe('<LandingPage />', () => {
...
-    const addonType = ADDON_TYPE_EXTENSION;
-    const dispatch = sinon.stub();
-    const errorHandler = new ErrorHandler({ id: 'some-id', dispatch });
-    render({
-      dispatch,
-      errorHandler,
-      params: { visibleAddonType: visibleAddonType(addonType) },
-      resultsLoaded: false,
-    });
+    const { store } = dispatchClientMetadata();
```
kumar303: This tends to be so common that I think it's better to make a var and initialize it in `beforeEach()`.

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -68,97 +82,85 @@ describe('<LandingPage />', () => {
...
+    sinon.assert.calledWith(fakeDispatch, setViewContext(ADDON_TYPE_EXTENSION));
+    sinon.assert.callCount(fakeDispatch, 1);
   });
 
   it('does not dispatch getLanding when there is an error', () => {
-    const dispatch = sinon.stub();
+    const { store } = dispatchClientMetadata();
+    const fakeDispatch = sinon.stub(store, 'dispatch');
+
     const errorHandler = new ErrorHandler({
```
kumar303: it should be possible to rewrite this as:

```js
  id: 'some-id', dispatch: store.dispatch,
});
errorHandler.handle(new Error('some error'));

render({ errorHandler, store });
```

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -303,4 +278,56 @@ describe('<LandingPage />', () => {
...
+      addonType,
+      featured: createAddonsApiResult([]),
+      highlyRated: createAddonsApiResult([]),
+      popular: createAddonsApiResult([]),
+    }));
+
+    const fakeDispatch = sinon.stub(store, 'dispatch');
+    render({ errorHandler, store });
+
+    sinon.assert.callCount(fakeDispatch, 2);
```
kumar303: This can go at the end of the test since the assertion message won't be so good. The other assertion messages will be more informative.

src/amo/components/LandingPage/index.js
```diff
@@ -63,43 +69,49 @@ export class LandingPageBase extends React.Component {
...
       addonTypeOfResults,
-      dispatch,
       errorHandler,
       loading,
       params,
       resultsLoaded,
-    } = this.props;
+    } = {
+      ...this.props,
+      ...nextProps,
```
kumar303: If I comment this out all tests pass so the updating logic needs better coverage.

src/amo/components/LandingPage/index.js
tofumatt: But then we lose out on it for the entire file and it's still useful more than it isn't. Removing it means we'll miss out on opportunities to remove unused prop-types in the future, which is handy.

tests/unit/amo/components/TestLandingPage.js
willdurand: ah right

tests/unit/amo/components/TestLandingPage.js
willdurand: actually, that was needed for the `renderAndMount()` method that I did not touch.

src/amo/components/LandingPage/index.js
willdurand: I have added a test case to cover this case.

src/amo/components/LandingPage/index.js
kumar303: porting the prop types to Flow will get us much better coverage of prop usage.

PR #3413, opened 2017-10-09
Fetch featured themes on Android

### LINKED ISSUE #3408: Themes categories have issues on Android
Steps to reproduce:
1. Load Themes homepage using "Add-ons for Android" option https://addons-dev.allizom.org/en-US/android/themes/
2. Click on any category i.e. Abstract

Expected results:
Corresponding category page is loaded without any display or layout issues.

Actual results:
Add-ons are continuously loading. For some categories an error message are displayed on top of the page.

Notes/Issues:
- reproducing only using "Add-ons for Android" option.

Screenshot for this issue:
invalid

willdurand:
Fix #3408

---

This PR fixes an issue with the themes for Android by porting the same workaround used in the search API to the featured API, since both are different.

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -13,6 +13,7 @@ import log from 'core/logger';
 import {
   addVersionCompatibilityToFilters,
   convertFiltersToQueryParams,
+  fixFiltersForAndroidThemes,
 } from 'core/searchUtils';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import type { ApiStateType } from 'core/reducers/api';
@@ -207,11 +208,12 @@ type FeaturedParams = {|
 |};
 
 export function featured({ api, filters, page }: FeaturedParams) {
+  const newFilters = fixFiltersForAndroidThemes({ api, filters });
+
   return callApi({
     endpoint: 'addons/featured',
     params: {
-      app: api.clientApp,
-      ...convertFiltersToQueryParams(filters),
+      ...convertFiltersToQueryParams(newFilters),
       page,
     },
     schema: { results: [addon] },
--- a/src/core/searchUtils.js
+++ b/src/core/searchUtils.js
@@ -3,6 +3,7 @@ import { oneLine } from 'common-tags';
 import {
   ADDON_TYPE_THEME,
   CLIENT_APP_ANDROID,
+  CLIENT_APP_FIREFOX,
 } from 'core/constants';
 import log from 'core/logger';
 
@@ -132,8 +133,8 @@ export const fixFiltersForAndroidThemes = ({ api, filters }) => {
   ) {
     log.info(oneLine`addonType: ${newFilters.addonType}/clientApp:
       ${newFilters.clientApp} is not supported. Changing clientApp to
-      "firefox"`);
-    newFilters.clientApp = 'firefox';
+      "${CLIENT_APP_FIREFOX}"`);
+    newFilters.clientApp = CLIENT_APP_FIREFOX;
   }
 
   return newFilters;
--- a/src/core/searchUtils.js
+++ b/src/core/searchUtils.js
@@ -1,5 +1,9 @@
 import { oneLine } from 'common-tags';
 
+import {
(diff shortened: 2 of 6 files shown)


src/core/api/index.js
```diff
@@ -207,11 +211,37 @@ type FeaturedParams = {|
 |};
 
 export function featured({ api, filters, page }: FeaturedParams) {
+  const newFilters = { ...filters };
```
kumar303: This workaround code looks identical to that in `search()` so you should be calling it from a shared function. I would suggest giving it a name that stands out so that we are motivated to fix it soon :) Maybe like:

```js
const newFilters = fixFiltersForAndroidThemes({ filters, api });
```

tests/unit/core/api/test_index.js
```diff
@@ -264,6 +302,38 @@ describe(__filename, () => {
           return mockWindow.verify();
         });
     });
+
+    it('does not change clientApp for Android extensions', () => {
```
kumar303: Once you move the workaround code into a shared function, I don't think you need to test it here because that would be a duplication of the tests in search().

src/core/api/index.js
willdurand: Fair enough :)

willdurand:
So I've extracted the fix in a new helper function but kept the tests at the API level. Let me know if we should inject the helper to be able to mock it in the tests (this seems a bit odd to me though).

tests/unit/core/api/test_index.js
```diff
@@ -235,7 +239,40 @@ describe(__filename, () => {
...
+            },
+            result: {
+              results: ['foo', 'food', 'football'],
+            },
+          });
+          return mockWindow.verify();
+        });
+    });
+
+    it('sets app to `firefox` when addon type is theme for Android', () => {
```
kumar303: Github makes it hard to comment on this new test you added but I think it does too much. All it should do is add the `.withArgs('.../?app=firefox...')` assertion. It doesn't need to check the entities in the response because that's already covered.

To answer the question about mock injection: I think your approach here is fine. This will serve as a smoke test to make sure that the function is calling the helper but it won't check all edge cases.

willdurand: Oh, that's what you meant by smoke test! I thought it had something to do with an injected mock or so... Sorry, I completely misunderstood you. Everything you say here makes sense, going to change that!

willdurand:
Ready for a third look, sorry 😕

src/core/searchUtils.js
```diff
@@ -103,3 +107,34 @@ export function hasSearchFilters(filters) {
...
+  // Obviously we need to fix this on the API end so our requests aren't
+  // overridden, but for now this will work.
+  if (
+    newFilters.clientApp === CLIENT_APP_ANDROID &&
+    newFilters.addonType === ADDON_TYPE_THEME
+  ) {
+    log.info(oneLine`addonType: ${newFilters.addonType}/clientApp:
+      ${newFilters.clientApp} is not supported. Changing clientApp to
+      "firefox"`);
+    newFilters.clientApp = 'firefox';
```
kumar303: This should use `CLIENT_APP_FIREFOX`

tests/unit/core/test_searchUtils.js
```diff
@@ -171,4 +177,70 @@ describe(__filename, () => {
...
+      };
+
+      const newFilters = fixFiltersForAndroidThemes({ filters });
+      expect(newFilters).toEqual(filters);
+    });
+
+    it('does not do anything when clientApp is `firefox`', () => {
+      const filters = {
+        addonType: ADDON_TYPE_THEME,
+        clientApp: CLIENT_APP_FIREFOX,
```
kumar303: This isn't a great test because it will never fail. You could make it fail by changing the filter to `clientApp: 'some-bogus-client'` and writing the spec more like "it does not change filters unless clientApp is `android`"

willdurand: ah indeed, thanks!

Issue #3266, opened 2017-09-28
Add-ons incompatible with your platform look installable


ValentinaPC:
Steps to reproduce:
1. Load details page for an add-on that is not compatible with your platform i.e. https://addons.allizom.org/en-US/firefox/addon/webext-no-platf/
2. See the installation button

Expected results:
The install button is disabled. A red banner saying that the add-on is not compatible with your platform is displayed.

Actual results:
No banner regarding the add-on platform incompatibility. Install button is enabled. At click the switch is moved but the color is not turned to green. Add-on is not installed.

Notes/Issues:
Verified on AMO-stage FF55(Win 7). Issue is also reproducing on -dev.
- refreshing the page will put the switch button in its correct position.
Screenshot for this issue:
![noplatform](https://user-images.githubusercontent.com/15685960/30956918-5a7c15f4-a441-11e7-8cc6-8dff980f60ad.png)

tofumatt:
@kumar303, I wonder if this is related to the new platform install feature?

- Matthew Riley MacPherson (sent from mobile)

ValentinaPC:
@muffinresearch 👍 
- example: https://addons.allizom.org/en-US/developers/addon/webext-no-platf/versions
- platforms - anything beside the platform you are using, in my case: Linux and Mac
![2017-10-02_0944](https://user-images.githubusercontent.com/15685960/31066558-737fa79c-a756-11e7-8fa3-2f50e308105a.png)

kumar303:
> I wonder if this is related to the new platform install feature?

Yes, sort of. What's happening here is that no install URL exists for the add-on since it doesn't support the platform. However, why did the code even get that far.

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7).
Postfix screenshot:
![no_plat](https://user-images.githubusercontent.com/15685960/31435255-a10693f0-ae87-11e7-8b6e-cf286d7a385e.png)

PR #3419, opened 2017-10-09
Show incompatible error for missing platform URLs

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3266

Previously when an add-on declared itself compatible with a platform yet was missing a platform file, we just let the user try to install it. This resulted in a console error because there was no install URL. This patch shows an incompatible error banner and disables the install button instead.


src/core/utils.js
```diff
@@ -417,6 +420,22 @@ export function isCompatibleWithUserAgent({
       return { compatible: false, reason: INCOMPATIBLE_NO_OPENSEARCH };
     }
 
+    // Even if an extension's version is marked compatible,
```
tofumatt: These are the edge cases that make AMO so special 😆

tests/unit/amo/components/TestAddonCompatibilityError.js
```diff
@@ -10,25 +10,30 @@ import {
...
 
 
 describe(__filename, () => {
-  const defaultUserAgentInfo = {
-    browser: { name: 'Firefox' },
-    os: { name: 'Plan 9' },
+  let store;
+
+  beforeEach(() => {
+    store = createStore().store;
```
tofumatt: This could use `dispatchClientMetadata` for a more accurate/complete "default" store.

tests/unit/amo/components/TestAddonCompatibilityError.js
```diff
@@ -42,11 +47,10 @@ describe(__filename, () => {
   }
 
   it('renders a notice for non-Firefox browsers', () => {
-    const root = render({
-      lang: 'en-GB',
-      reason: INCOMPATIBLE_NOT_FIREFOX,
-      userAgentInfo: { browser: { name: 'Chrome' }, os: {} },
+    _dispatchClientMetadata({
```
tofumatt: Not sure why this was made when we could use the existing `dispatchClientMetadata` helper.

tests/unit/core/test_utils.js
```diff
@@ -495,23 +526,72 @@ describe('isCompatibleWithUserAgent', () => {
...
+      UAParser(userAgentsByPlatform.windows.firefox40);
+
+    expect(isCompatibleWithUserAgent({ addon, userAgentInfo }))
+      .toEqual({
+        compatible: false,
+        reason: INCOMPATIBLE_UNSUPPORTED_PLATFORM,
+      });
+  });
+
+  it('allows non-extensions to have mismatching platform files', () => {
```
tofumatt: Does this ever happen? I didn't think themes could/did set platform.

tests/unit/amo/components/TestAddonCompatibilityError.js
kumar303: With this helper I don't have to pass `store` into the function every time

tests/unit/amo/components/TestAddonCompatibilityError.js
kumar303: This component relies on `lang` and `userAgent` so I wanted each test to call `dispatchClientMetadata()` instead since the values are important. I can add a comment to this if it's helpful though.

tests/unit/core/test_utils.js
kumar303: I'm not sure but since this code is only working around a bug in extensions I wanted to limit it there until we know we need it elsewhere.

PR #3426, opened 2017-10-10
Integrate dennis linting for .po files.

### LINKED ISSUE #3336: Lint l10n translations with dennis-cmd
Part of https://github.com/mozilla/addons-frontend/issues/529

To avoid things like broken placeholders or other obvious problems.

* [ ] automate and document dennis-cmd installation
* [ ] Integrate dennis-cmd into linting package.json, potentially run it after `build-locales` script
* [ ] add to travis to run on every commit to catch these problems early on in the development process

EnTeQuAk:
Fixes #3336

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,9 @@ commands = flake8 {posargs:.}
 [testenv:dennis-lint]
 deps = dennis==0.9
 # Run via `bash -c` to make sure the file globbing are properly resolved on the bash
-commands = bash -c 'dennis-cmd lint --errorsonly locale/*/LC_MESSAGES/*.po'
+commands =
+  bash -c 'dennis-cmd lint --errorsonly locale/*/LC_MESSAGES/*.po'
+  bash -c 'dennis-cmd lint --errorsonly locale/templates/LC_MESSAGES/*.pot'
 whitelist_externals =
   bash
 
--- a/tox.ini
+++ b/tox.ini
@@ -31,7 +31,10 @@ commands = flake8 {posargs:.}
 
 [testenv:dennis-lint]
 deps = dennis==0.9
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -29,5 +29,12 @@ commands = pytest tests/ui/test_discopane.py {posargs}
 deps = flake8==3.3.0
 commands = flake8 {posargs:.}
 
+[testenv:dennis-lint]
+deps = dennis==0.9
+# Run via `bash -c` to make sure the file globbing are properly resolved on the bash
+commands = bash -c 'dennis-cmd lint --errorsonly locale/*/LC_MESSAGES/*.po'
```
muffinresearch: This won't check the pot files? Should they be validated too? E.g. https://github.com/mozilla/addons-frontend/tree/master/locale/templates/LC_MESSAGES

tox.ini
```diff
@@ -29,5 +29,12 @@ commands = pytest tests/ui/test_discopane.py {posargs}
 deps = flake8==3.3.0
 commands = flake8 {posargs:.}
 
+[testenv:dennis-lint]
+deps = dennis==0.9
+# Run via `bash -c` to make sure the file globbing are properly resolved on the bash
```
muffinresearch: s/file globbing are properly resolved on the bash/file globbing is properly resolved in the shell/

tox.ini
EnTeQuAk: Fair point, we never validated the .pot files in olympia but might be a good idea. I'll add it here, won't hurt I guess.

muffinresearch: Yeah the .pot files should be good, but this should hopefully help us in a case where the extraction process does something odd.

kumar303:
@EnTeQuAk thanks for adding this!

PR #3440, opened 2017-10-11
fix: limit search term to 100 characters

### LINKED ISSUE #3429: Prevent searches with a query of more than 100 characters in length
https://github.com/mozilla/addons-server/issues/6604 will introduce a limit to the `q` parameter in the search API. When called with a query longer than 100 characters, it will answer with a 400 Bad Request. 

The frontend should prevent the user from making such a request by limiting the user input (and/or be prepared to handle the potential 400 Bad Request response)

tsl143:
Fixes #3429

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -33,7 +33,7 @@ import Icon from 'ui/components/Icon';
 
 import './styles.scss';
 
-
+const SEARCH_TERM_MAX_LENGTH = 100;
 export class SearchFormBase extends React.Component {
   static propTypes = {
     addonType: PropTypes.string,
@@ -125,7 +125,10 @@ export class SearchFormBase extends React.Component {
   }
 
   handleSearchChange = (event) => {
-    this.setState({ searchValue: event.target.value });
+    const searchValue = event.target.value || '';
+    if (searchValue.length <= SEARCH_TERM_MAX_LENGTH) {
+      this.setState({ searchValue });
+    }
   }
 
   dispatchAutocompleteStart = this.props.debounce(({ filters }) => {
@@ -160,8 +163,9 @@ export class SearchFormBase extends React.Component {
         but filters have not changed.`);
       return;
     }
-
-    this.dispatchAutocompleteStart({ filters });
+    if (value.length <= SEARCH_TERM_MAX_LENGTH) {
+      this.dispatchAutocompleteStart({ filters });
+    }
   }
 
   handleSuggestionsClearRequested = () => {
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestSearchForm.js
```diff
@@ -271,6 +271,13 @@ describe(__filename, () => {
 
       wrapper.find('input').simulate('change', createFakeChangeEvent(undefined));
       expect(wrapper.state('searchValue')).toEqual('');
+
+      // Tests for allowing only 100 characters as input
+      wrapper.find('input').simulate('change', createFakeChangeEvent('asdfghjklpasdfghjklpasdfghjklpasdfghjklpasdfghjklpasdfghjklpasdfghjklpasdfghjklpasdfghjklp98765432111'));
```
willdurand: haha maybe we could use [`String.repeat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat):

```
'c'.repeat(100)
```

src/amo/components/SearchForm/index.js
```diff
@@ -125,7 +125,10 @@ export class SearchFormBase extends React.Component {
   }
 
   handleSearchChange = (event) => {
-    this.setState({ searchValue: event.target.value });
+    const searchValue = event.target.value || '';
+    if (searchValue.length <= SEARCH_TERM_MAX_LENGTH) {
```
willdurand: We could maybe `trim()` before `.length`?

src/amo/components/SearchForm/index.js
```diff
@@ -160,8 +163,9 @@ export class SearchFormBase extends React.Component {
         but filters have not changed.`);
       return;
     }
-
-    this.dispatchAutocompleteStart({ filters });
+    if (value.length <= SEARCH_TERM_MAX_LENGTH) {
```
willdurand: Maybe this check could happen earlier?

tsl143:
@willdurand `t` instead of `c` coz its Trishul :P

tsl143:
@willdurand anything more to do on this?

src/amo/components/SearchForm/index.js
```diff
@@ -125,7 +125,10 @@ export class SearchFormBase extends React.Component {
   }
 
   handleSearchChange = (event) => {
-    this.setState({ searchValue: event.target.value });
+    const searchValue = event.target.value || '';
```
willdurand: Can you add a test case to cover this condition? I think you want to test the case where `event.target.value` is not set or something, and ensure it fallbacks to empty string, avoiding any further issue.

willdurand:
thanks @tsl143!

Issue #3310, opened 2017-10-02
Add curated link "subject shelf" to homepage [3/5]


muffinresearch:
Breaking out from #3152 

https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit#heading=h.bcmv6y37ccp

This  one is a hard-coded curated collection. 

This one needs specialised UX since it's a hard-coded list of links to a collection  - see the doc for the content. 

This is what is known as a subject shelf - this requires iconography.

Order: this is shelf no. 3

willdurand:
Hi @pwalm, we need you here :)

pwalm:
Icons! More icons! I think some of the Subject Icons will cover these, but i'll design options for the theme icons this week.

pwalm:
Subject shelf icons in situ:
<img width="1175" alt="screen shot 2017-10-10 at 3 04 51 pm" src="https://user-images.githubusercontent.com/6797635/31413058-71b13ba2-adcc-11e7-8ac7-36b0abbed187.png">

Only had to create two new icons: "screenshots" and "wild". SVGs of those files are [here](https://github.com/mozilla/addons-frontend/tree/master/assets/icons/subjects).

willdurand:
Cool, thanks!

devaneymoz:
@pwalm your [Subject Shelf icon designs](https://github.com/mozilla/addons-frontend/tree/master/assets/icons/subjects) are looking great! 

Just so we're in sync on Subject Shelf icons for launch, I had these four specifically noted because they will lead to strong curated collections by November: 
- [Ad blockers](https://addons.mozilla.org/firefox/collections/mozilla/ad-blockers/)
- [Password Managers](https://addons.mozilla.org/firefox/collections/mozilla/password-managers/)
- [Bookmarks](https://addons.mozilla.org/firefox/collections/mozilla/bookmark-managers/)
- [Watching Videos](https://addons.mozilla.org/firefox/collections/mozilla/watching-videos/)

... however if UX calls for more than four Subject Shelf icons to be displayed on the Homepage, please let me know asap so I can make programming adjustments, additions.

willdurand:
Hi @pwalm and @devaneymoz, here is how it looks so far:

[...]

PR #3442, opened 2017-10-11
Add up & coming shelf

### LINKED ISSUE #3312: Add  Up & Coming shelf to the homepage [5/5]
Split out from #3152

https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit#heading=h.rf8x4i1of5cm

Should follow existing shelf  styling so we  can update all later: 

Here's an example of that for most popular data:

add-ons_for_firefox

willdurand:
Fix #3312

---

This PR adds the fifth shelf on the homepage.

ℹ️ It is the 5th shelf but the 3rd has not been added yet (#3310), therefore it is the last (and 4th) shelf.

screen shot 2017-10-11 at 17 16 14

### DIFF
--- a/tests/unit/amo/containers/TestHome.js
+++ b/tests/unit/amo/containers/TestHome.js
@@ -21,6 +21,7 @@ import {
   CLIENT_APP_ANDROID,
   CLIENT_APP_FIREFOX,
   SEARCH_SORT_POPULAR,
+  SEARCH_SORT_TRENDING,
   VIEW_CONTEXT_HOME,
 } from 'core/constants';
 import { ErrorHandler } from 'core/errorHandler';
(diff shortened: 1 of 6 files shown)


tests/unit/amo/containers/TestHome.js
```diff
@@ -164,6 +165,19 @@ describe(__filename, () => {
     expect(shelf).toHaveProp('loading', true);
   });
 
+  it('renders a featured collection shelf', () => {
```
willdurand: this one was missing :/

tests/unit/amo/sagas/test_home.js
```diff
@@ -142,7 +161,7 @@ describe(__filename, () => {
 
       mockSearchApi
         .expects('search')
-        .once()
+        .twice()
```
tofumatt: Is this just because we call it twice for different results? Should we test if only one response is bad, as well?

willdurand: Well, either we put `twice` or nothing actually, otherwise sinon will complain.

willdurand: No actually it is `twice` or ... it fails because of `sinon` 😓 

I changed this test to use the collection API, which is used only once so that it keeps the same behavior as before.

PR #3449, opened 2017-10-11
feat: New static carousel

### LINKED ISSUE #3332: Randomize header content order

### LINKED ISSUE #3430: Replace homepage carousel with three static panes
Chatted with @pwalm about this, and we're looking to replace the carousel with a set of static panes, something like:

5_alternatives_to_using_a_carousel_on_your_website_homepage

This will allow all "carousel" items to be viewed at once, prevents CSP issues, should increase load times (by being able to load smaller assets), and should be a bit nicer overall. Plus we don't have to implement an infinite carousel so it's more time for us to knock out other bugs.

tofumatt:
* fix #3430
* fix #3332

### Small
screen shot 2017-10-11 at 23 57 23

### Medium
screen shot 2017-10-11 at 23 57 29

### Large
screen shot 2017-10-11 at 23 57 32

### RTL
(It works now, so easy!)
screen shot 2017-10-12 at 00 47 30

Responsive: https://www.dropbox.com/s/mjjbmb7e4rpe5et/2017-10-11%2023.57.04.gif?dl=0


tofumatt:
This implements the new carousel with random items. In the future we can do stuff like webpack compress the images, etc., but for now I just wanted to get this out as the diff is big enough 😄

devaneymoz:
@tofumatt are we settled on going with three carousel images? (one hero and two smaller ones?). 

@pwalm thoughts on doing something visually to distinguish editorial headlines ("Block ads") from headlines that are extension titles ("YouTube High Definition")? I could see this getting confusing, especially if an extension title sort of sounds like a headline (content like "AdBlock for Firefox" or "Best Video Downloader" come to mind)

tofumatt:
@devaneymoz, yeah, that's what we discussed with UX. The sections are random though; there are currently four pieces of content to pull from–one will be the "hero" image on the left and the other two are the "side" images.

tofumatt:
I think adding the Add-on's icon for content that's a specific add-on would be good.

But then I still don't think the collection content is useful; I think it's better to feature specific add-ons over collections. But at the very least, it's best to avoid "ad-blocking" collections when we might also feature something like "uBlock" (and I think uBlock should *always* be in the rotation 😄 ).

muffinresearch:
@tofumatt was the wide left and right gutter a conscious decision? It definitely makes it stand out more - but I wonder if how it would look if the left + right gutter was reduced to match the gap between items?

tofumatt:
No, largely it’s just from outdated gutters in the card styling in general. Indeed I can tighten it up here.... I did during development but was just gonna do it card-wide as part of the overall style tweaks.

But I’ll do it here because this is the homepage 😀

Matthew Riley MacPherson

src/core/reducers/carousels.js
```diff
@@ -0,0 +1,40 @@
...
+  };
+};
+
+export const initialState = {};
+
+export default function randomOrderReducer(state = initialState, action) {
+  switch (action.type) {
+    case SET_CAROUSEL_ORDER: {
+      const { name, random, sections } = action.payload;
+      const orderArray = [...Array(sections.length).keys()].slice(0, MAX_ITEMS);
```
muffinresearch: Could this just be `[...sections.keys()].slice(0, MAX_ITEMS);` ?

tests/unit/amo/reducers/testCarousels.js
```diff
@@ -0,0 +1,86 @@
+import knuthShuffle from 'knuth-shuffle';
+
+import carouselsReducer, {
+  initialState,
+  setCarouselOrder,
+} from 'core/reducers/carousels';
+
+
+describe(__filename, () => {
```
muffinresearch: Nice, I like this idea, but would it break if we were running tests in a browser in the future? Just I think it's likely that it may be possible to run tests in a browser at some point in the future.

tests/unit/amo/reducers/testCarousels.js
```diff
@@ -0,0 +1,86 @@
...
+      sections: [1, 2, 3],
+    }));
+
+    sinon.assert.notCalled(knuthShuffleSpy);
+    expect(state).toMatchObject({
+      NotAKnownKey1234: { order: [0, 1, 2] },
+    });
+  });
+
+  it('randomizes the order if random prop is passed', () => {
```
muffinresearch: It might be nice to add an assertion to make sure the original data isn't modified in terms of order by the shuffle.

src/ui/components/Carousel/index.js
```diff
@@ -1,98 +1,60 @@
...
+    const {
+      carousels,
+      dispatch,
+      name,
+      random,
+      sections,
+    } = this.props;
+
+    if (!carousels[name]) {
+      dispatch(setCarouselOrder({ name, random, sections }));
```
muffinresearch: One question - how does the randomization work in terms server-side renders vs client ones?

willdurand: Since the randomization happens in the reducer, the client side should receive the same ordered list of sections (server state being serialized to be passed to the client).

src/amo/components/HomeCarousel/index.js
```diff
@@ -78,7 +80,11 @@ export class HomeCarouselBase extends React.Component {
   render() {
     return (
       <div className="HomeCarousel">
-        <Carousel sections={this.carouselSections()} />
+        <Carousel
```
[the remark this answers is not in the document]
tofumatt: I thought about that but find "Hero" such an inside-baseball term. Even I don't quite know what it means. Maybe HeroBanner or something, or even HeroCarousel? Sure thing though.

src/core/reducers/carousels.js
tofumatt: No, because I want array indexes, not the actual content itself. The content doesn't get stored/modified in the reducer, just the order it's stored in.

tests/unit/amo/reducers/testCarousels.js
tofumatt: Well, we do this every where already, so maybe, but at that point we'll need to deal with it 😉

tests/unit/amo/reducers/testCarousels.js
tofumatt: Oh, sure thing.

src/ui/components/Carousel/index.js
tofumatt: Yup, that's the only reason it's stored in the reducer, otherwise this would be way over-engineered 😉 .

src/core/reducers/carousels.js
tofumatt: Oh snap, true. Alright, I'll fix this up. 👍

PR #3457, opened 2017-10-12
fix: consistent pagination layout

### LINKED ISSUE #3436: Keep consistency for nav arrows between search results and collections details pages
Steps to reproduce:
1. Load a collection details page that contains more than 25 add-ons i.e. https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/privacy/
2. Observe navigation arrows
3. Search for something i.e. "test" - https://addons-dev.allizom.org/en-US/firefox/search/?platform=windows&q=test
4. Observe navigation arrows

Expected results:
Pagination arrows are positioned in the same place.

Actual results:
Pagination arrows position differ.

Notes/Issues:
Verified on AMO-dev FF56(Win 7)
Screenshots:

tsl143:
Fixes #3436 

Before:
screen shot 2017-10-12 at 5 51 37 pm

After:
screen shot 2017-10-12 at 5 59 15 pm

### DIFF
--- a/src/amo/components/Search/styles.scss
+++ b/src/amo/components/Search/styles.scss
@@ -53,8 +53,7 @@
 
   .Paginate {
     @include respond-to(large) {
-      grid-column: 1 / -1;
-      grid-row: 4;
+      grid-column: 2;
     }
   }
 


willdurand:
Saw your comment in the issue, I personally find this is too big on desktops, and I'd go for the other way around actually. It makes sense to have pagination below the list only in my opinion, and it does not affect the mobile view anyway.

tsl143:
> Saw your comment in the issue, I personally find this is too big on desktops, and I'd go for the other way around actually. It makes sense to have pagination below the list only in my opinion, and it does not affect the mobile view anyway.

@willdurand, yes that's the way it should be, shall we wait for @pwalm, or already implement this?

willdurand:
This is going to be updated post-release anyway (#3100), so I'd just go ahead but maybe @tofumatt has some more thoughts on it before you change anything again :)

muffinresearch:
@tsl143 I agree with Will  - If it's not too much trouble to do the opposite then having the pagination  under the content (like the first screenshot) on balance looks better.

willdurand:
Hi @tsl143, I went ahead and adjusted your PR 😉

willdurand:
Updated screenshot for the search results:

screen shot 2017-10-16 at 19 07 42

willdurand:
@tofumatt since I modified @tsl143's PR, I'd like a quick r? Thanks!

PR #3483, opened 2017-10-12
make sure collections get loaded when params change

### LINKED ISSUE #3482: Collections not updated on the client

1. Go to the "Block ads" collection from homepage
2. Return to the homepage
3. Navigate to the "Change your tabs" collection

### What happened?

You still get the old content, i.e. the data of the "Block ads" collection

The new collection "Change your tabs" is loaded

willdurand:
Fix #3482

---

This PR makes sure collections get loaded correctly on the client.

## Gifs

Before:

After:

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -79,6 +79,13 @@ export class CollectionBase extends React.Component {
       }
     }
 
+    if (collection && (
+      collection.slug !== params.slug ||
+      collection.authorUsername !== params.user
+    )) {
+      collectionChanged = true;
+    }
+
     if (!collection || collectionChanged) {
       this.props.dispatch(fetchCollection({
         errorHandlerId: errorHandler.id,
--- a/tests/unit/amo/components/TestCollection.js
+++ b/tests/unit/amo/components/TestCollection.js
@@ -29,14 +29,18 @@ import {
 
 
 describe(__filename, () => {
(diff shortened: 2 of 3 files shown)


tests/unit/amo/components/TestCollection.js
```diff
@@ -29,14 +29,18 @@ import {
 
 
 describe(__filename, () => {
+  // These values match the `createFakeCollectionDetail()` helper data.
+  const defaultUser = 'johndoe';
+  const defaultSlug = 'my-addons';
```
kumar303: what about this?

```js
const defaultCollection = createFakeCollectionDetail();
const defaultUser = defaultCollection.author.username;
const defaultSlug = defaultCollection.slug;
```

src/amo/components/Collection/index.js
```diff
@@ -79,6 +79,13 @@ export class CollectionBase extends React.Component {
       }
     }
 
+    if (collection && (
+      collection.slug !== params.slug ||
+      collection.authorUsername !== params.user
+    )) {
+      collectionChanged = true;
+    }
```
kumar303: Ugh, the loading logic in this component is really complicated. It's my first time seeing it. I know it's necessary because of SSR though. Your new logic looks fine, I'm just complaining out loud :/

tests/unit/amo/components/TestCollection.js
willdurand: much better!

Issue #3458, opened 2017-10-12
'Publish reply' button appears inactive after trying to publish an empty review


AlexandraMoga:
Steps to reproduce:

1. Sign in as an add-on developer
2. Go to an add-on you own
3. Click to read all reviews
4. Reply to one of the reviews
5. Enter only blank spaces and tap on Publish reply
6. Enter text and observe the 'Publish reply' button

Expected result:
The 'Publish reply' button is active once text is introduced

Actual result:
The 'Publish reply' button looks like it is inactive

Notes:  
Tapping the "apparently inactive" button once text is introduced will publish the reply
This can be reproduced only on device (AMO-dev - FF58, Android 7.0)
Relates to [#3415](https://github.com/mozilla/addons-frontend/issues/3415)

Video for the issue:
![publish button](https://user-images.githubusercontent.com/31961530/31497937-3f63687a-af69-11e7-8b10-78f03ad92f20.gif)

kumar303:
This is working as expected. A user should not be able to publish a blank reply because that would be confusing. How would an empty reply be helpful? I [made a typo in the issue you mentioned](https://github.com/mozilla/addons-frontend/issues/3415#issuecomment-336166466) which I think is what caused the confusion. Sorry about that.

kumar303:
If I misunderstood the issue just let me know.

kumar303:
I think I may know what you are referring to. When you have your mouse hovering over the Publish button, it has a hover style that makes the button look disabled.

kumar303:
Was that it?

AlexandraMoga:
@kumar303 : the issue described by me does not imply that the user should be allowed to submit an empty review. 
I was pointing specifically to the way the "Publish reply" button is behaving when making an attempt to leave an empty review and then typing actual text.
When the text field contains only blank spaces, the Publish button is disabled, which is correct. But if the user tries to type text after attempting to submit an empty review, the button still looks as it was inactive.
Hope this makes more sense now. If not, I will try to make the scenario more clear.

kumar303:
> the button still looks as it was inactive.

This is just because your mouse was hovering over the button. Try moving your mouse cursor off of the button. I don't think you'll be able to reproduce it anymore.

AlexandraMoga:
As mentioned in the bug, this can be reproduced only on device, so the mouse hovering does not apply.
What you see in the gif is the actual reproduction from my device screen, but I use the mouse pointer to make it more clear in the video what options are selected.

[...]

PR #3485, opened 2017-10-12
Fix double-tap and hover bug in submit button for mobile

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3458

* The form looks and behaves the same as before
* I haven't been able to test this on an Android device. Do you think it will fix the hover issue?
* I started to document how to do local development on Android but it started taking so much time I just gave up and documented how far I got (which is better than nothing)

### DIFF
--- a/src/ui/components/Button/index.js
+++ b/src/ui/components/Button/index.js
@@ -1,8 +1,10 @@
 import classNames from 'classnames';
+import { oneLine } from 'common-tags';
 import React from 'react';
 import PropTypes from 'prop-types';
 
 import Link from 'amo/components/Link';
+import log from 'core/logger';
 
 
 import './Button.scss';
@@ -11,6 +13,7 @@ export default class Button extends React.Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
+    disabled: PropTypes.boolean,
     href: PropTypes.string,
     to: PropTypes.string,
   }
@@ -23,7 +26,11 @@ export default class Button extends React.Component {
       to,
       ...rest
     } = this.props;
-    const props = { className: classNames('Button', className), ...rest };
+    const props = { ...rest };
+
+    const setClassName = (...classConfig) => {
+      return classNames('Button', className, ...classConfig);
+    };
 
     if (href || to) {
       if (href) {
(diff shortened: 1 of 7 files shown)


kumar303:
@tofumatt since I couldn't find a way to test this, could you take a quick look? I know you've dealt with this before.

src/ui/components/Button/index.js
```diff
@@ -23,7 +26,11 @@ export default class Button extends React.Component {
       to,
       ...rest
     } = this.props;
-    const props = { className: classNames('Button', className), ...rest };
+    const props = { ...rest };
+
+    const setClassName = (...classConfig) => {
+      return classNames('Button', className, ...classConfig);
```
tofumatt: Nit but this will output just the class `disabled`, right? I'd think we'd want `Button--disabled`... I think this would work with `classConfig.map((name) => `Button--${name}`)`

src/ui/components/Button/index.js
```diff
@@ -35,9 +42,18 @@ export default class Button extends React.Component {
...
+          event.preventDefault();
+          log.warn(oneLine`Not calling onClick() for Button link to
+            ${props.href || props.to} because it is disabled`);
+        };
+      }
       return <Link {...props}>{children}</Link>;
     }
 
-    return <button {...props}>{children}</button>;
+    return <button className={setClassName()} {...props}>{children}</button>;
```
tofumatt: I'd think `setClassName({ disabled: props.disabled })` down here would work just fine; if `props.disabled` is false-y `classNames` won't output that class. That's easier to follow for me; this approach was a bit confusing to trace.

kumar303: When a `<button>` has the property `disabled={true}` then it already gets styled correctly because of [`:disabled`](https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/mixins.scss#L260).

src/ui/components/Button/index.js
kumar303: Yes, I intended to use the general purpose [`.disabled`](https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/mixins.scss#L261) style. Were you thinking of `InstallButton--disabled`?

kumar303:
Thanks for the reviews! I'm going to get it on dev so I can test it but let me know if it needs any follow-ups.

PR #3494, opened 2017-10-13
Fix chrome height bug

### LINKED ISSUE #3015: Search results page has a height of more than 17000px on Chrome/Desktop view
When displaying search results, _e.g._, https://addons-dev.allizom.org/en-US/firefox/search/?q=ad+block, then Chrome users will show an extra height:

screen shot 2017-08-28 at 17 30 06

It does not happen on small screens (< 768px).

---
Chrome Version 59.0.3071.115 (Official Build) (64-bit) on MacOS.

muffinresearch:
Fixes #3015

Before: 

add-ons_for_firefox

After the fix there's no extra space rendered under the search results:

add-ons_for_firefox

### DIFF
--- a/src/amo/components/Collection/styles.scss
+++ b/src/amo/components/Collection/styles.scss
@@ -55,7 +55,7 @@
     display: grid;
     grid-auto-flow: column dense;
     grid-gap: 0 10px;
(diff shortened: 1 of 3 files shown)


willdurand:
:tada:

Issue #3495, opened 2017-10-13
Add API/actions/sagas for flagging reviews


kumar303:
* We need to hook up the API for flagging reviews (as spam, etc)
* We need action creators and sagas

This supports https://github.com/mozilla/addons-frontend/issues/2774

Issue #2774, opened 2017-07-12
Add UI to flag a review for spam, abuse, etc


tofumatt:
This is a UI that lets a user flag an add-on review for the following reasons:
* It's spam
* It contains inappropriate language
* It is a bug report or support request (not a review)

Related to #2772.

You will be able to flag a review for a different reason than what's listed here in ﻿﻿https://github.com/mozilla/addons-frontend/issues/3573

muffinresearch:
This is probably going to follow the UX for reviews.

kumar303:
The UX for this form can follow https://github.com/mozilla/addons-frontend/issues/2775#issuecomment-332973335 (which is also in Sketch now)

ValentinaPC:
Verified this on AMO-dev FF56(Win 7).
The "Flag" blue link and options were added.
Postfix screenshots:
![flag](https://user-images.githubusercontent.com/15685960/31879708-e510acda-b7e6-11e7-81cf-ccb715714c91.png)
![rtl](https://user-images.githubusercontent.com/15685960/31879838-555c3af4-b7e7-11e7-9a2b-014c37d76d8e.png)

logged a separate issue for anything that seems to not work as expected around flag option (https://github.com/mozilla/addons-frontend/issues/3598).
We can mark this verified fixed.

Issue #1461, opened 2016-12-07
Protect function args with invariant


kumar303:
We have a lot of functions that protect against incorrect arguments like this:
````javascript
export function doTheThing({ error, id } = {}) {
  if (!id) {
    throw new Error('id cannot be empty');
  }
  if (!error) {
    throw new Error('error cannot be empty');
  }
  ...
}
````
We should use [invariant](https://github.com/zertosh/invariant) instead because that's what it's meant for. It will also give us a bit more granularity about what type of argument we are expecting.

kumar303:
It might be more helpful to focus on moving more parts of the app to Flow instead

kumar303:
In separate discussions, @tofumatt had requested not to use Flow so I'll re-open this.

tofumatt:
Heh, it's definitely made it in more places... I'll happily admit while I wouldn't elect to use it it's caught a few bugs and I've found it handy a fair few times. 😄

But yeah, given we protect a lot of action creators with and without flow, this might still be handy for the "without-flow" bit.

kumar303:
Also, adding Flow coverage to all the test files would be a near impossible feat so I don't think we'd ever get there.

muffinresearch:
@kumar303 @tofumatt where are we at with this? Do you think this is still something we'd like to do?

tofumatt:
Invariant would decrease the size of our bundle by changing our error messages in production, but that’s it. I think it’s worth keeping open but we already throw when required args aren’t supplied in most functions these days.

- Matt (Sent from mobile)

kumar303:
I'm getting a little annoyed with writing `if (!param) { throw new Error(...) }` and adding tests to cover each case so ... 😎

kumar303:
We're using invariant for all new code now.

PR #3496, opened 2017-10-13
Add API/actions/sagas for flagging reviews

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3495

* Adds low-level API stuff needed for the UI (https://github.com/mozilla/addons-frontend/issues/2774)
* Adds type safety to the reviews reducer
* Fixes issues with the review saga test that I hadn't noticed before

For context, this is how I envision the UI will work:
* The user goes to the list of all add-on reviews
* They see one that needs flagging
* The click on 'Flag this review' and get a dropdown similar to the More menu item. 
* The can select things like: Flag for spam, inappropriate language, misplaced bug report, or other (this will allow them to enter a text explanation)

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -1,15 +1,18 @@
 /* @flow */
 import {
-  SHOW_EDIT_REVIEW_FORM,
-  SHOW_REPLY_TO_REVIEW_FORM,
   FETCH_REVIEWS,
   HIDE_EDIT_REVIEW_FORM,
   HIDE_REPLY_TO_REVIEW_FORM,
   SEND_REPLY_TO_REVIEW,
+  SEND_REVIEW_FLAG,
   SET_ADDON_REVIEWS,
   SET_REVIEW,
   SET_REVIEW_REPLY,
+  SET_REVIEW_WAS_FLAGGED,
+  SHOW_EDIT_REVIEW_FORM,
+  SHOW_REPLY_TO_REVIEW_FORM,
 } from 'amo/constants';
+import type { FlagReviewReasonType } from 'amo/constants';
 import type {
   ExternalReviewReplyType, ExternalReviewType,
 } from 'amo/api/reviews';
@@ -51,7 +54,7 @@ export function denormalizeReview(
 }
 
 export type SetReviewAction = {|
-  type: string,
+  type: typeof SET_REVIEW,
   payload: UserReviewType,
 |};
 
@@ -70,7 +73,7 @@ type SetReviewReplyParams = {|
 |};
 
 export type SetReviewReplyAction = {|
-  type: string,
+  type: typeof SET_REVIEW_REPLY,
   payload: SetReviewReplyParams,
 |};
 
@@ -96,7 +99,7 @@ type FetchReviewsParams = {|
 |};
 
 export type FetchReviewsAction = {|
-  type: string,
+  type: typeof FETCH_REVIEWS,
   payload: {|
     addonSlug: string,
     errorHandlerId: string,
@@ -129,7 +132,7 @@ export const setDenormalizedReview = (
 };
 
 export type SetAddonReviewsAction = {|
-  type: string,
+  type: typeof SET_ADDON_REVIEWS,
   payload: {|
     addonSlug: string,
     reviewCount: number,
@@ -173,7 +176,7 @@ type SendReplyToReviewParams = {|
 |};
 
 export type SendReplyToReviewAction = {|
-  type: string,
+  type: typeof SEND_REPLY_TO_REVIEW,
   payload: SendReplyToReviewParams,
 |};
 
@@ -202,7 +205,7 @@ type ReviewIdActionParams = {|
 
 export const reviewIdAction = (
   { reviewId, type }: ReviewIdActionParams = {}
-) => {
+): any => {
   if (!reviewId) {
     throw new Error('The reviewId parameter is required');
   }
@@ -214,7 +217,7 @@ type ShowEditReviewFormParams = {|
 |};
 
 export type ShowEditReviewFormAction = {|
-  type: string,
+  type: typeof SHOW_EDIT_REVIEW_FORM,
   payload: ShowEditReviewFormParams,
 |};
 
@@ -229,7 +232,7 @@ type ShowReplyToReviewParams = {|
 |};
 
 export type ShowReplyToReviewFormAction = {|
-  type: string,
+  type: typeof SHOW_REPLY_TO_REVIEW_FORM,
   payload: ShowReplyToReviewParams,
 |};
 
@@ -244,7 +247,7 @@ type HideEditReviewFormParams = {|
 |};
 
 export type HideEditReviewFormAction = {|
-  type: string,
+  type: typeof HIDE_EDIT_REVIEW_FORM,
   payload: HideEditReviewFormParams,
 |};
 
@@ -259,7 +262,7 @@ type HideReplyToReviewFormParams = {|
 |};
 
 export type HideReplyToReviewFormAction = {|
-  type: string,
+  type: typeof HIDE_REPLY_TO_REVIEW_FORM,
   payload: HideReplyToReviewFormParams,
 |};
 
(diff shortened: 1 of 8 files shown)


kumar303:
r? @tofumatt or @willdurand 

It might be good to add a comment in here when you begin the review. This will allow me to make sure the PR is fully up to date. Since I'll be working on the UI patch in the background I may need to push tweaks to this as they come up.

src/amo/actions/reviews.js
```diff
@@ -1,15 +1,18 @@
 /* @flow */
 import {
-  SHOW_EDIT_REVIEW_FORM,
-  SHOW_REPLY_TO_REVIEW_FORM,
   FETCH_REVIEWS,
   HIDE_EDIT_REVIEW_FORM,
   HIDE_REPLY_TO_REVIEW_FORM,
   SEND_REPLY_TO_REVIEW,
+  SEND_REVIEW_FLAG,
```
tofumatt: I don't like the word "flag" as it's vague and a bit unclear. For the other thing we used `SEND_ABUSE_REPORT` so I'd say `REPORT_REVIEW_FOR_ABUSE` or `SEND_REVIEW_ABUSE_REPORT`?

src/amo/actions/reviews.js
```diff
@@ -1,15 +1,18 @@
...
   SET_ADDON_REVIEWS,
   SET_REVIEW,
   SET_REVIEW_REPLY,
+  SET_REVIEW_WAS_FLAGGED,
```
tofumatt: Nitpick but if this were clearer like `SET_REVIEW_MARKED_ABUSIVE` that might be nice.

src/amo/actions/reviews.js
```diff
@@ -268,3 +271,58 @@ export const hideReplyToReviewForm = (
...
+
+export type FlagReviewAction = {|
+  type: typeof SEND_REVIEW_FLAG,
+  payload: FlagReviewParams,
+|};
+
+export const flagReview = (
+  { errorHandlerId, note, reason, reviewId }: FlagReviewParams = {}
+): FlagReviewAction => {
+  if (!errorHandlerId) {
```
tofumatt: We should really have a helper for this, heh.

src/amo/constants.js
```diff
@@ -1,13 +1,37 @@
...
-export const SHOW_EDIT_REVIEW_FORM = 'SHOW_EDIT_REVIEW_FORM';
-export const SHOW_REPLY_TO_REVIEW_FORM = 'SHOW_REPLY_TO_REVIEW_FORM';
-export const FETCH_REVIEWS = 'FETCH_REVIEWS';
-export const HIDE_EDIT_REVIEW_FORM = 'HIDE_EDIT_REVIEW_FORM';
-export const HIDE_REPLY_TO_REVIEW_FORM = 'HIDE_REPLY_TO_REVIEW_FORM';
-export const SET_ADDON_REVIEWS = 'SET_ADDON_REVIEWS';
-export const SET_REVIEW = 'SET_REVIEW';
-export const SET_REVIEW_REPLY = 'SET_REVIEW_REPLY';
-export const SEND_REPLY_TO_REVIEW = 'SEND_REPLY_TO_REVIEW';
+export const SHOW_EDIT_REVIEW_FORM: 'SHOW_EDIT_REVIEW_FORM' =
```
tofumatt: I hate to complain and I've been quiet about a lot of flow creeping into parts of the codebase that aren't proptypes, but this is baffling to look at. I get full flow coverage would mean never needing constants, but as we use them, forcing the flow coverage to the constant values is really unfriendly.

src/amo/constants.js
```diff
@@ -1,13 +1,37 @@
...
+export const SET_ADDON_REVIEWS: 'SET_ADDON_REVIEWS' = 'SET_ADDON_REVIEWS';
+export const SET_REVIEW: 'SET_REVIEW' = 'SET_REVIEW';
+export const SET_REVIEW_REPLY: 'SET_REVIEW_REPLY' = 'SET_REVIEW_REPLY';
+export const SET_REVIEW_WAS_FLAGGED: 'SET_REVIEW_WAS_FLAGGED' =
+  'SET_REVIEW_WAS_FLAGGED';
+export const SEND_REPLY_TO_REVIEW: 'SEND_REPLY_TO_REVIEW' =
+  'SEND_REPLY_TO_REVIEW';
+export const SEND_REVIEW_FLAG: 'SEND_REVIEW_FLAG' = 'SEND_REVIEW_FLAG';
+
+export const REVIEW_FLAG_REASON_SPAM: 'review_flag_reason_spam' =
```
tofumatt: Any reason these are lowercase?

src/amo/sagas/reviews.js
```diff
@@ -70,7 +72,33 @@ function* handleReplyToReview(
   }
 }
 
+function* handleFlagReview(
```
tofumatt: As mentioned, I don't really like flag as user-facing OR developer-facing terminology. I think using "abuse report" whenever possible is clearer and sets up expectations better. So I'd rather this be `handleReviewAbuseReport` or something similar.

src/amo/actions/reviews.js
```diff
@@ -268,3 +271,58 @@ export const hideReplyToReviewForm = (
...
+    throw new Error('The errorHandlerId parameter is required');
+  }
+  if (!reason) {
+    throw new Error('The reason parameter is required');
+  }
+  if (!reviewId) {
+    throw new Error('The reviewId parameter is required');
```
tofumatt: All of the "the"s in here seem redundant and are unlike the usual missing param error messages we have, so they can probably be removed.

tests/unit/amo/sagas/testReviews.js
```diff
@@ -3,13 +3,19 @@ import SagaTester from 'redux-saga-tester';
...
   setAddonReviews,
   sendReplyToReview,
   setReview,
   setReviewReply,
+  setReviewWasFlagged,
 } from 'amo/actions/reviews';
-import { SET_ADDON_REVIEWS } from 'amo/constants';
+import {
+  REVIEW_FLAG_REASON_OTHER,
+  REVIEW_FLAG_REASON_SPAM,
```
tofumatt: Oh wait, I see, the reasons vary. So maybe ignore my comments about flagging as abusive. But I still think "report" is a better verb than "flag".

src/amo/actions/reviews.js
kumar303: I filed a bug for it a while ago https://github.com/mozilla/addons-frontend/issues/1461

src/amo/actions/reviews.js
kumar303: I think making it a complete sentence will help someone reading the exception message to parse it quicker and act on it quicker. I don't request that change in other people's code because it's not so important--it's just a personal preference that I think makes it easier for other developers.

src/amo/constants.js
kumar303: They map to specific values that the API accepts https://addons-server.readthedocs.io/en/latest/topics/api/reviews.html#post--api-v3-reviews-review-(int-id)-flag-

src/amo/sagas/reviews.js
kumar303: It's not just an abuse report. You "flag" a review so that an admin can inspect it. You might flag it for abuse but you might also flag it for other reasons.

src/amo/actions/reviews.js
tofumatt: I'm fine with either, mostly I think we should just be consistent 😄

src/amo/constants.js
kumar303: It's not great but utilizing real action types adds a lot of value. It gives us safety checks in how each reducer case handles its action object. Reducer code is one of the most critical parts of the application. Our unit tests cannot always cover using invalid payload properties--those bugs are very easy to introduce when refactoring action code.

> a lot of flow creeping into parts of the codebase

This code was already Flow-typed, it just wasn't working.

tofumatt: Fair enough. I think I'm just cranky about the syntax. I've had more coffee and now I'm fine with it, ignore me 😉

src/amo/constants.js
tofumatt: Ah, right on. Might be worth a comment explaining what these are then, but makes sense.

src/amo/sagas/reviews.js
tofumatt: Yeah, I realised this after but I think maybe just "report" might be a better verb than "flag".

src/amo/constants.js
kumar303: Also, since our test code doesn't have Flow coverage, it would be risky to switch away from constants. If one of these constants got renamed, we might miss a test. Hopefully it would fail but sometimes tests do a lot of stubbing.

kumar303:
@tofumatt Thanks for the quick review! Since the functionality is not limited to abuse reports, your change requests don't really fit with the logic. Could you take another look and revise your change requests? Is there a different verb for *flag* that you'd like me to use? I like to match the API whenever it makes sense and in this case *flag* makes sense since we're marking the review for admin inspection. It's not a *report* necessarily.

tofumatt:
Okay, upon further review of the code/comments/API docs I get it. Yeah, +1 to being consistent everywhere, flag it is.

PR #3498, opened 2017-10-13
Fix safari header moving around

### LINKED ISSUE #3497: Safari header grid moves around
Visit https://addons.mozilla.org in safari and you'll see something like this (this is RTL) but you get the idea:

localhost_3000_ar_firefox_themes_

Navigate and the login button part off-screen in the screenshot above moves backwards and forwards.

Here's a gif showing the craziness that ensures:

safari-weirdness

muffinresearch:
Fixes #3497 

Before:

safari-weirdness

After:

safari-weirdness-fixed

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -12,13 +12,13 @@
   padding: 0 12px;
 
   @include respond-to(medium) {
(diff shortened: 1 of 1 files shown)


ExE-Boss:
This has partially resolved #3079 (only on iOS 11).

Issue #3501, opened 2017-10-13
Search box length was regressed


muffinresearch:
In fixing the header the search box is smaller in large viewports:

On prod:

<img width="1360" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/31560352-d94d1c8c-b04b-11e7-897d-c5210f4deeb1.png">

Locally:

<img width="1400" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/31560377-fb34ea3c-b04b-11e7-944a-8ff51eeef6c2.png">

AlexandraMoga:
Verified this on FF56 with Windows 10 and macOS 10.12.6. However, I wasn't able to spot a difference between the Search box layout on either of the systems or AMO servers.

Production:
![image](https://user-images.githubusercontent.com/31961530/31618583-037d4850-b29b-11e7-863c-cf82fee9984c.png)


-Dev:
![image](https://user-images.githubusercontent.com/31961530/31618554-f070c728-b29a-11e7-9887-1cb9f231f435.png)

@muffinresearch : Am I missing some verification steps here?

willdurand:
@AlexandraMoga hi, you should not see any difference between the width of the search bar in -dev and its width in production. According to your screenshots, it looks ok to me.

AlexandraMoga:
@willdurand Thanks a lot for clarifying this.

PR #3502, opened 2017-10-13
Fix search box visual length

muffinresearch:
Fixes: #3501
Before:

add-ons_for_firefox

After:

add-ons_for_firefox

Still works in safari too.

localhost_3000_ar_firefox_

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -18,7 +18,7 @@
   }
 
   @include respond-to(extraLarge) {
-    grid-template-columns: max-content 1fr auto;
+    grid-template-columns: max-content 1fr 1fr;
     grid-template-rows: 46px auto;
     margin: 0 auto;
     max-width: $max-content-width;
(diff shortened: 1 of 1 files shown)


src/amo/components/Header/styles.scss
```diff
@@ -18,7 +18,7 @@
   }
 
   @include respond-to(extraLarge) {
-    grid-template-columns: max-content 1fr auto;
+    grid-template-columns: max-content 1fr 1fr;
```
tofumatt: lol I was going to ask about this other `auto` in the other patch and thought "naww I'm sure he tested it and it's fine".

muffinresearch: Heh, yeah - saw some lovely visual regression tools demo'd at fronteers - stuff like this makes that look really useful.

willdurand: so we should forget about `auto` for any css grid stuff, right?

ExE-Boss:
This has partially resolved #3079 (only on iOS 11).

Issue #3137, opened 2017-09-14
"More featured" is inconsistent compared to "More highly rated" and "More popular"


muffinresearch:
See https://addons-dev.allizom.org/en-US/firefox/extensions/featured/

This is inconsistent with "more highly rated" and "more popular" which are just search results pages.

<img width="1408" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30427498-f727fbf2-9947-11e7-93f9-ed3f11a6364e.png">

@pwalm not sure there's specific mocks for these - can we consolidate on a search page based approach?

AlexandraMoga:
Verified as fixed on AMO-dev with FF56 (Win10 and Android 7.0)

Postfix:
![image](https://user-images.githubusercontent.com/31961530/31667197-5fe2f326-b357-11e7-91b8-2eec5ec1d7b8.png)

![image](https://user-images.githubusercontent.com/31961530/31667340-d9907f5e-b357-11e7-976f-e0fd94be5b4d.png)

PR #3504, opened 2017-10-13
Get rid of the old featured API in favor of the search one

### LINKED ISSUE #3307: Featured add-ons should be compatible with user browser
The current implementation of the featured add-ons retrieval relies on the [Featured API endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#featured). This is not good since we try not to show incompatible add-ons to the users. This endpoint has no notion of `appversion` parameter for instance.

The featured add-ons retrieval does not rely on the Search API but ideally we should do that. I have filled an issue on addons-server to ask for a new `sort` type (random) so that it mimics the existing Featured API endpoint: https://github.com/mozilla/addons-server/issues/6560.

This would help us to always retrieve compatible featured add-ons.

Once we have the API for that, we could remove the code to retrieve the featured add-ons (in `core/api/index`) and use the regular Search API (in `core/api/search`).

willdurand:
Fix #3307

---

This PR removes calls to the Featured API in favor of the Search API so that we can leverage all its parameters (including `appversion`) and return compatible add-ons only. The UI update will follow in #3137.

### DIFF
--- a/src/amo/sagas/featured.js
+++ b/src/amo/sagas/featured.js
@@ -6,8 +6,8 @@ import { call, put, select, takeEvery } from 'redux-saga/effects';
 
 import { loadFeatured } from 'amo/actions/featured';
 import { FEATURED_ADDONS_TO_LOAD } from 'amo/constants';
-import { featured as featuredApi } from 'core/api';
-import { FEATURED_GET } from 'core/constants';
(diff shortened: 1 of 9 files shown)


tofumatt:
I can review this really quickly before I go meet a friend for dinner 👍

PR #3524, opened 2017-10-16
Fix search filters

### LINKED ISSUE #3121: Filter menu options change at page refresh
Steps to reproduce:
1. Load Themes homepage https://addons-dev.allizom.org/en-US/firefox/themes/
2. Click on "More Highly Rates Themes"
3. Observe Filter menu
4. Refresh the page

Expected results:
Filter menu stays the same.

Actual results:
Filter menu options change.

Notes/Issues:
- reproducing on other pages like: search results from homepage, "More Highly Rates Extensions", "More Popular Themes", "More Popular Extensions".
- wherever you are filters turn in to "Relevance", "Extensions", "Windows" not mater what.

Verified on AMO-dev FF55(Win 7). Not reproducing on device.
For the Type option of Filter menu - the issue is also reproducing on prod.

Video for this issue:
change

willdurand:
Fix #3121

---

So, this PR fixes a weird bug related to the search filters being updated with the wrong values on page refresh. First, I thought that was an issue related to the `key` prop not correctly set on each `option`, so I changed that but it did not work actually... Yet, this still fixes something. I started to investigate more, but it was not easy to trigger the problem. Here are a few steps to make it work (or fail depending on your point of view):

- go to https://addons-dev.allizom.org/en-US/firefox/themes/
- click on "more trending themes"
- refresh (might not be necessary)
- go to "extensions" (header link)
- go to "more trending extensions" (well, looks good to me)
- refresh and 💥 

I thought that was a bug in the `render()` of the `SearchFilters` component, not being updated with the right `props`, but no. It gets the correct `props`, all the time. Ha! I checked on https://github.com/facebook/react but found related issues that have all been fixed... So I tried the previous scenario in Chrome and I could not reproduce it. Doh! I tried many other things to reproduce in Chrome but I was unsuccessful.

I then googled for known issues with Firefox and I found this: [Why doesn't Firefox show the correct default select option?](https://stackoverflow.com/questions/1479233/why-doesnt-firefox-show-the-correct-default-select-option), related to this [bugzilla ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=46845). 💡 Firefox sometimes thinks it is a good idea to keep the old values of a `form`, but here it is not really a good idea, so we have to tell it not to do that.

That's how the `<form autoComplete="off">` landed in this patch 😅

### DIFF
--- a/src/amo/components/SearchFilters/index.js
+++ b/src/amo/components/SearchFilters/index.js
@@ -126,74 +126,76 @@ export class SearchFiltersBase extends React.Component {
         className="SearchFilters"
         header={i18n.gettext('Filter results')}
       >
-        <label
-          className="SearchFilters-label"
-          htmlFor="SearchFilters-Sort"
-        >
-          {i18n.gettext('Sort by')}
-        </label>
-        <select
-          className="SearchFilters-select"
-          id="SearchFilters-Sort"
-          name="sort"
-          onChange={this.onSelectElementChange}
-          value={filters.sort || 'relevance'}
-        >
-          {this.sortOptions().map((option) => {
-            return <option key={option.name} {...option} />;
-          })}
-        </select>
+        <form autoComplete="off">
+          <label
+            className="SearchFilters-label"
+            htmlFor="SearchFilters-Sort"
+          >
+            {i18n.gettext('Sort by')}
+          </label>
+          <select
+            className="SearchFilters-select"
+            id="SearchFilters-Sort"
+            name="sort"
+            onChange={this.onSelectElementChange}
+            value={filters.sort || 'relevance'}
+          >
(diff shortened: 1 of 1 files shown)


muffinresearch:
Does this submit anything  to the current page if you hit enter?

willdurand:
> Does this submit anything to the current page if you hit enter?

@muffinresearch I was not able to trigger any submit on FF Nightly and Chrome (tried to focus the form, enter, cmd + enter, enter outside of the form, and enter on a select box opened with no change).

willdurand:
@muffinresearch are there any new comments apart from the question about hitting `enter`?

muffinresearch:
> @muffinresearch are there any new comments apart from the question about hitting enter?

No - that was the only one.

Issue #1096, opened 2016-09-20
Make redux actions hot reloadable again


kumar303:
### Describe the problem and steps to reproduce it:

Run a dev server with the react hot reloader enabled. Example:

```
npm run dev:amo
```

Change the code in any action creator.
### What happened?

The old code is still running.
### What did you expect to happen?

The hot reloader should automatically reload the action code so that you don't have to refresh the entire page and potentially lose your state.
### Anything else we should know?

It looks like this was a regression in some upstream module but there are a couple different issues. See:
- https://github.com/reactjs/redux/issues/782
- https://github.com/gaearon/babel-plugin-react-transform/issues/26

React Hot Loader 3.0 should fix this. Preview: https://github.com/gaearon/react-hot-boilerplate/pull/61

willdurand:
Here is a patch for that: https://github.com/mozilla/addons-frontend/commit/b5f49aae9eeb05622ab5923e12e84eb42791f462

See also this (minor) issue: https://github.com/mozilla/addons-frontend/pull/5652#discussion_r205609278

AlexandraMoga:
Please add STR or the label 'qa:not needed'
Thanks

PR #3529, opened 2017-10-17
chore(package): update react-hot-loader to version 3.1.1

willdurand:
Closes #3434

### DIFF
--- a/package.json
+++ b/package.json
@@ -260,7 +260,7 @@
     "postcss-loader": "^2.0.6",
     "potools": "^0.1.0",
     "react-addons-test-utils": "^15.5.1",
-    "react-hot-loader": "^1.3.0",
+    "react-hot-loader": "^3.1.1",
     "react-test-renderer": "^15.5.4",
     "react-transform-hmr": "^1.0.4",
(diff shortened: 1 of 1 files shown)


willdurand:
So HMR still works for me with this PR. We should definitely switch to 3.x, I don't think 1.x will ever receive fixes.

kumar303:
I'm kinda nervous since I don't want anything to break our dev flow before we ship. If you tested it I guess it's fine. We could always revert it.

willdurand:
@tofumatt have you tried this one? 🙏

tofumatt:
I’ve not but I can give it a quick try now. Is there a big benefit to the upgrade? :-)

willdurand:
> Is there a big benefit to the upgrade?

One more PR merged 😋  (and also improvements like better SSR support and reloads, apparently)

We can wait though.

tofumatt:
Fair enough :-) I’ll give it a quick go.

kumar303:
Go for it. It can't possibly break action reloading since that's already broken 😞  https://github.com/mozilla/addons-frontend/issues/1096

PR #3530, opened 2017-10-17
Upgrade moment to 2.9.1

willdurand:
Closes #3425


willdurand:
Note: I could not create a new PR based on #3425 for unknown reasons (reported to GitHub support, but no reply yet).

PR #3532, opened 2017-10-17
fix(package): update raven-js to version 3.19.1

willdurand:


willdurand:
According to the [CHANGELOG](https://github.com/getsentry/raven-js/blob/master/CHANGELOG.md), this looks good to merge.

PR #3535, opened 2017-10-17
fix(package): update raven to version 2.2.1

willdurand:


willdurand:
reviewed the [changes](https://github.com/getsentry/raven-node/releases) and it looked good. I had to create the PR myself because GitHub did weird things with the Greenkeeper ones.

PR #3561, opened 2017-10-18
fix: Use InstallButton all the time on AMO (fix #3557)

### LINKED ISSUE #3557: Change install switch to install button on AMO
According to our user research the install switch performs very poorly so we should disable it and use the fallback `InstallButton` everywhere on AMO for now. In the future we'll want to improve the button, but switching to it for MVP.

tofumatt:
Force the InstallButton to always be a button on AMO.

Note that the Disco Pane should still stay a switch because it's nice to have a toggle there and it's less an issue.

### After

screenshot 2017-10-18 21 35 53

### DIFF
--- a/tests/unit/core/components/TestInstallButton.js
+++ b/tests/unit/core/components/TestInstallButton.js
@@ -205,6 +205,25 @@ describe(__filename, () => {
     expect(button).toHaveProp('disabled', true);
   });
 
+  it('renders a switch button if useButton is false', () => {
+    const root = render({ useButton: false });
+
+    expect(root).toHaveClassName('InstallButton--use-switch');
+  });
+
+  it('renders a button if useButton is true', () => {
+    const root = render({ useButton: true });
+
+    expect(root).toHaveClassName('InstallButton--use-button');
+
+    const button = root.childAt(1);
+
+    expect(button.type()).toEqual(Button);
+    expect(button).toHaveClassName('Button--action');
+    expect(button).toHaveClassName('InstallButton-button');
+    expect(button.children()).toIncludeText('Add to Firefox');
+  });
+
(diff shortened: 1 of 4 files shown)


tests/unit/core/components/TestInstallButton.js
```diff
@@ -205,6 +205,38 @@ describe(__filename, () => {
     expect(button).toHaveProp('disabled', true);
   });
 
+  it('renders a switch button if useButton is false', () => {
+    const root = render({
+      addon: createInternalAddon({
```
willdurand: Why is this needed?

tests/unit/core/components/TestInstallButton.js
```diff
@@ -205,6 +205,38 @@ describe(__filename, () => {
...
+        ...fakeAddon,
+        hasAddonManager: true,
+      }),
```
kumar303: I think you meant:

```js
const root = render({
  addon: createInternalAddon({
    ...fakeAddon,
  }),
  hasAddonManager: true,
});
```

And in the other test.

tests/unit/core/components/TestInstallButton.js
tofumatt: Turns out: it's not! Copied from another test and assumed it was needed, but it wasn't.

tests/unit/core/components/TestInstallButton.js
tofumatt: Yeah, bad assumptions in copypasta, fixing up now.

muffinresearch:
Does dead code elimination mean we're not including the mozAddonManager install button code and CSS?

tofumatt:
No, because it’s still used on Disco Pane and a fair bit of the code will be useful in the future. I think the safest/quickest route is just to supply the `useButton` prop for now.

PR #3574, opened 2017-10-19
UI to flag an add-on review for spam, etc

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2774

Here is the main functionality:

addons-frontend-flag-review mov

screenshot 2017-10-20 08 54 06

You have to be logged in first or else you'll see this. This is a functional button that will log you in:
screenshot 2017-10-19 16 18 28

You can't flag your own review (duh):
screenshot 2017-10-19 16 39 12

If there's an error (like a 500 API error) you'll see the message and you can try again. Encountering an error is not something we expect to happen. I had to break the login code to show this message.
screenshot 2017-10-19 16 32 35


kumar303:
r? @tofumatt or @willdurand 

Sorry for the epic patch.

tests/unit/amo/components/TestFlagReview.js
```diff
@@ -0,0 +1,120 @@
...
+      reviewId: review.id,
+    }));
+
+    // The user triggers an error:
+    errorHandler.handle(new Error('Unexpected API error'));
+
+    const root = render({ errorHandler, reason });
+    expect(root.find(ErrorList)).toHaveLength(1);
+
+    // It should still display a button so they can try again.
```
tofumatt: 👍

tests/unit/amo/components/TestFlagReviewMenu.js
```diff
@@ -0,0 +1,161 @@
...
+
+      const flag = menu.find('.FlagReviewMenu-flag-bug-support-item')
+        .find(FlagReview);
+      expect(flag).toHaveProp('review', review);
+      expect(flag).toHaveProp('reason', REVIEW_FLAG_REASON_BUG_SUPPORT);
+      expect(flag).toHaveProp('wasFlaggedText');
+      expect(flag).toHaveProp('buttonText');
+    });
+
+    it('does not prompt you to flag a response as a bug/support', () => {
```
tofumatt: I get why this is but it just took me a few seconds to parse, understand the intention here, and then realise why. Feel free to disagree but I think a comment explaining why "responses can't be flagged as bug reports" might be handy. 😄

kumar303: ok, I added a comment to `FlagReviewMenu` about it.

src/ui/components/TooltipMenu/styles.scss
```diff
@@ -48,31 +48,42 @@
...
-      @include font-medium();
-
+    a:link,
+    button {
       color: $grey-90;
-      cursor: pointer;
       font-size: $font-size-m-smaller;
       line-height: 1.6;
+      overflow: hidden;
+      text-align: left;
```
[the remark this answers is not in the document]
kumar303: whoops, that was a mistake

kumar303:
Thanks for the review!

> are we allowed to flag a review as multiple things?

The API does allow it and I guess you might want to flag a review for both spam and inappropriate language -- maybe flagging it twice would raise its priority for admins? Or maybe flagging it once would be enough.

Anyway, I was flying blind without UX specs here so we will definitely need to iterate on this. I'm also not sure if the menu should auto-close when you're done flagging + lots of little things.

Issue #2998, opened 2017-08-24
Cannot install add-on if it supports multiple platforms


kumar303:
### Describe the problem and steps to reproduce it:

* Go to an add-on that supports multiple platforms
* Example: https://addons-dev.allizom.org/en-US/firefox/addon/ratopuse-weasp-vulturkal/
* Click install

### What happened?

You may or may not get the correct add-on file for your platform when you click the install button.

### What did you expect to happen?

Depending on what platform you're on (Linux, Windows, Mac OS X) you should always get the right add-on file during installation.

### Anything else we should know?

Here are some real world cases that might trigger this issue:
* There are a lot of legacy add-ons with multiple platform files
* A webextension add-on that requires native messaging to talk to a desktop application might want to say it only supports Mac and Windows but not Linux. This add-on would have multiple files and would trigger this bug. I think 1Password is an example of this but they seem to self-host their add-on.
* When a developer is submitting their add-on they may choose to deselect the 'all platforms' box for whatever reason and select the other platform boxes explicitly. This would also trigger the bug.
* An extension may contain an Android version and a desktop version but I can't think of any real reason why you'd do that.

You can take a look at the API representation of the example add-on with:

```
curl https://addons-dev.allizom.org/api/v3/addons/addon/ratopuse-weasp-vulturkal/
```

This is what the `files` array looks like:

```js
  "files" : [
         {
            "is_restart_required" : false,
            "id" : 253517,
            "size" : 4162,
            "is_webextension" : true,
            "permissions" : [],
            "platform" : "linux",
            "created" : "2017-08-24T14:47:04Z",
            "status" : "public",
            "hash" : "sha256:638e8aa4fc91e58cb003f163b1e129b362aa991a0460bcf9101daff7f8a4b653",
            "url" : "https://addons-dev.allizom.org/firefox/downloads/file/253517/ratopuse_weasp_vulturkal-1.0-an+fx-linux.xpi?src="
         },
         {
            "is_restart_required" : false,
            "id" : 253518,
            "is_webextension" : true,
            "size" : 4161,
            "platform" : "mac",
            "permissions" : [],
            "status" : "public",
            "created" : "2017-08-24T14:48:26Z",
            "url" : "https://addons-dev.allizom.org/firefox/downloads/file/253518/ratopuse_weasp_vulturkal-1.0-an+fx-mac.xpi?src=",
[...]

PR #3575, opened 2017-10-19
Give XPIs to all Linux/Unix users

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2998

For user agent reference, see https://github.com/faisalman/ua-parser-js


tofumatt:
Got this one too.

src/core/installAddon.js
```diff
@@ -46,8 +46,23 @@ import {
...
+  USER_AGENT_OS_BSD_NETBSD,
+  USER_AGENT_OS_BSD_OPENBSD,
+  USER_AGENT_OS_BSD_PC,
   USER_AGENT_OS_LINUX,
+  USER_AGENT_OS_LINUX_ARCH,
+  USER_AGENT_OS_LINUX_CENTOS,
+  USER_AGENT_OS_LINUX_DEBIAN,
+  USER_AGENT_OS_LINUX_FEDORA,
+  USER_AGENT_OS_LINUX_GENTOO,
+  USER_AGENT_OS_LINUX_GNU,
```
tofumatt: [But aren't *all* Linuxes GNU? Stallman would be angry.](https://www.gnu.org/gnu/linux-and-gnu.en.html) 😉

src/core/installAddon.js
```diff
@@ -165,9 +180,26 @@ export function makeMapDispatchToProps({ WrappedComponent, src }) {
 
 const userAgentOSToPlatform = {
   [USER_AGENT_OS_ANDROID]: OS_ANDROID,
-  [USER_AGENT_OS_LINUX]: OS_LINUX,
   [USER_AGENT_OS_MAC]: OS_MAC,
   [USER_AGENT_OS_WINDOWS]: OS_WINDOWS,
+  // Not all of these are strictly Linux but giving them a Linux XPI
```
tofumatt: Should we maybe file a bug to rename "Linux" to "*NIX" or something in the UI? I dunno, both options are kinda meh but I wonder which is better...

tests/unit/core/TestInstallAddon.js
```diff
@@ -297,6 +297,20 @@ describe(__filename, () => {
         .toEqual('https://a.m.o/files/linux.xpi');
     });
 
+    it('finds a Linux Ubuntu install URL', () => {
```
tofumatt: Can we have an extra one that is 100% not Linux but UNIX instead, just for completeness/clarity? I know a user was asking if we could load Linux XPIs for BSD users, so it'd be nice to have a test that spells out that intention and makes sure it stays true.

kumar303: Good call. I had actually made a typo (UNIX not Unix) so it wasn't even working. I'll double check the mapping but it's probably overkill to add a test for each one.

tofumatt: Yeah every one is definitely excessive, I was just thinking at least one "Unix". 😄

src/core/installAddon.js
willdurand: 😈

PR #3577, opened 2017-10-20
fix: Make abuse button less intense

### LINKED ISSUE #3553: Report abuse button is too prominent
https://www.dropbox.com/s/a7bfp2ip84mz1gp/Screenshot%202017-10-18%2010.28.56.png?dl=0

In the new design, the Report Abuse button is too prominent; it comes before ratings and is more noticeable than ratings. We should move it below ratings and make it less prominent.

tofumatt:
fix #3553

Makes the "Report this add-on for abuse" button less intense and make it not the most obvious CTA on the page by making the "Add to Firefox" button larger and it smaller.

Upon further reflection, yeah, I really made this too out there. The new design uses the action button style until the "Send report" button where a red style appears, making it clear the action is serious 😄 

### Before
screenshot 2017-10-20 01 41 07

screen shot 2017-10-20 at 01 41 36

### After
screenshot 2017-10-20 01 41 09

screen shot 2017-10-20 at 01 41 30

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -495,7 +495,7 @@ export class AddonBase extends React.Component {
             {addon ?
               <InstallButton
                 {...this.props}
-                className="Button--action Button--small"
+                className="Button--wide"
                 disabled={!isCompatible}
                 ref={(ref) => { this.installButton = ref; }}
                 status={installStatus}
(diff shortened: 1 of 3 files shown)


tofumatt:
I guess we could. It was still a bit bold but that’s a good point. I’ll try again and use it if it looks good. 😊

- Matt (Sent from mobile)

tofumatt:
Tried with a small red button and it looks good, so final design is:
screenshot 2017-10-20 16 00 22

willdurand:
Vendu! 👍

pwalm:
Why did the button get all small? the scale shouldn't have changed.

Issue #3583, opened 2017-10-20
Make Travis CI test log easier to read


kumar303:
Our CI test log is too big (https://github.com/mozilla/addons-frontend/issues/2980) so you always have read the raw log. Let's make it easier to read.

PR #3584, opened 2017-10-20
Disable jest colors for CI

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3583


kumar303:
The jest output is a bit easier for me to read now. There are still escape sequences for webpack colors and colors in Travis's own output but the test output itself doesn't have them.

PR #3585, opened 2017-10-20
Give the design some fresh air on desktop (implement explore page)

### LINKED ISSUE #3558: Launch Content for Explore Page
- Content displayed in the image below is accurate and matches to @devaneymoz's [launch content doc](https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit#)
- Make sure copy matches what is in the doc/mock up
- Icons are already designed and in the repo
- Theme icons should link to their respective categories, while the extension icons will link to the collections @devaneymoz has made

explore - launch content - 1366px

willdurand:
Fix #3558
~~⚠️  Depends on #3581~~

---

This PR implements the explore page as mentioned in #3558:

- the shelves have been sorted as requested
- the UI should be closer to what is expected

Because some components are used in other pages, I actually did a first pass on all the pages. It does not affect any screen below the `large` breakpoint.

## Screenshots

screen shot 2017-10-20 at 19 21 03

screen shot 2017-10-20 at 19 21 34
screen shot 2017-10-20 at 19 21 47
screen shot 2017-10-20 at 19 21 58
screen shot 2017-10-20 at 19 22 11
screen shot 2017-10-20 at 19 22 15
screen shot 2017-10-20 at 19 22 26
screen shot 2017-10-20 at 19 22 34
screen shot 2017-10-20 at 19 22 56


src/core/css/inc/mixins.scss
```diff
@@ -385,4 +386,8 @@ $default-arrow-margin: 3px;
   background: $base-color;
   overflow: hidden;
   padding: 10px 20px;
+
+  @include respond-to(large) {
+    padding: 15px 36px;
```
kumar303: where is 15px used? Should it be 12px since that's a value used more frequently in the specs?

kumar303:
@pwalm I have two questions:

* Should we follow the left-right order of icons for Shelf 1 and Shelf 5 as described in the [content doc](https://docs.google.com/document/d/1PeBCNbx0V-5NReKdyv1c14Za6sVjTJ60-ondGaJDBBg/edit) or in the Sketch file? They are both different.
* Do we still have shelf headers? The content docs does not mention headers but the mock up does. For example: "Themes change how Firefox looks." This header seems a bit redundant which is why I'm wondering.

willdurand:
Thanks!

I totally overlooked the hero banner as there was no highlighted changes in the google doc, my bad 😞

src/core/css/inc/mixins.scss
willdurand: If I go for `12px` then I'll have to change the `height` of the header card. `15px` is the value I read in Sketch (padding-bottom of the title in a header card) and is also used to vertically align the title.

willdurand:
I have updated the HeroBanner:

screen shot 2017-10-23 at 20 14 46

PR #3586, opened 2017-10-20
Reset page filter on change

### LINKED ISSUE #3579: Filtering by "Featured add-ons only" results in a 404 error for an inaccessible page number
Steps to reproduce:
1.Go to AMO homepage and perform a string search (i.e. 'block' - https://addons-dev.allizom.org/en-GB/firefox/search/?platform=windows&q=block)
2.Navigate to the next pages
3. Filter by "Featured add-ons only"

Expected result:
Featured add-ons are displayed, according to the selected filter

Actual result:
An 'invalid result' is displayed and the page is stuck in loading mode

Notes:
Reproduced on all AMO servers with FF56 (Win10 and Android 7.1.2)

image

willdurand:
Fix #3579

---

This PR resets the `page` search filter when one of these filters is updated.

### DIFF
--- a/src/amo/components/SearchFilters/index.js
+++ b/src/amo/components/SearchFilters/index.js
@@ -39,7 +39,7 @@ export class SearchFiltersBase extends React.Component {
   onSelectElementChange = (event) => {
     event.preventDefault();
 
-    const { clientApp, filters, lang, pathname, router } = this.props;
+    const { filters } = this.props;
     const newFilters = { ...filters };
 
     // Get the filter we're supposed to change and set it.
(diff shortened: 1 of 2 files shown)


src/amo/components/SearchFilters/index.js
```diff
@@ -79,6 +84,11 @@ export class SearchFiltersBase extends React.Component {
       newFilters.featured = true;
     }
 
+    if (newFilters.page) {
+      // We reset the page as filters have changed.
+      newFilters.page = 1;
+    }
+
```
willdurand: At some point, this and the `router.push` could be a helper (or extracted into a new method). Worth doing there or?

kumar303: > At some point, this and the router.push could be a helper (or extracted into a new method). Worth doing there or?

Yeah, I'd say you can share it in component method:

```js
doSearch({ newFilters, filtersWereChanged = false }) {
  const { ... } = this.props;

  if (filtersWereChanged && newFilters.page) {
    // Since it's now a new search, reset the page.
    newFilters.page = 1;
  }

  router.push({
      pathname: `/${lang}/${clientApp}${pathname}`,
      query: convertFiltersToQueryParams(newFilters),
  });
}
```

PR #3587, opened 2017-10-20
Add HTML titles

### LINKED ISSUE #2184: <title> is rarely changed
We use `react-helmet` which lets us change the `<title>` tag across pages but we almost never change it from the default `Firefox Add-ons` title. We should change the title per-page–it makes it easy to tell pages apart and also probably helps with search.

willdurand:
Fix #2184

---

This PR adds HTML titles on each page and also adapts the title according to the `clientApp`.

For the search results, I tried to have titles for each page that we link in the app. It does not aim at covering every cases for now.

## Some screenshots

screen shot 2017-10-20 at 22 22 34
screen shot 2017-10-20 at 22 22 40
screen shot 2017-10-20 at 22 22 49
screen shot 2017-10-20 at 22 22 59
screen shot 2017-10-20 at 22 23 06
screen shot 2017-10-20 at 22 23 12
screen shot 2017-10-20 at 22 23 25
screen shot 2017-10-20 at 22 23 29
screen shot 2017-10-20 at 22 23 38
screen shot 2017-10-20 at 22 23 47
screen shot 2017-10-20 at 22 23 51
screen shot 2017-10-20 at 22 24 04

screen shot 2017-10-20 at 22 22 04
screen shot 2017-10-20 at 22 22 12
screen shot 2017-10-20 at 22 22 17
screen shot 2017-10-20 at 22 22 26

### DIFF
--- a/tests/unit/amo/containers/TestApp.js
+++ b/tests/unit/amo/containers/TestApp.js
@@ -7,6 +7,7 @@ import {
 import NestedStatus from 'react-nested-status';
 import { Provider } from 'react-redux';
 import { loadFail } from 'redux-connect/lib/store';
+import Helmet from 'react-helmet';
 
 import App, {
   AppBase,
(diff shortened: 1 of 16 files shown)


tests/unit/amo/containers/TestApp.js
```diff
@@ -189,6 +195,24 @@ describe('App', () => {
     expect(root.find(NestedStatus)).toHaveProp('code', 200);
   });
 
+  it('configures a default HTML title for Firefox', () => {
+    const { store } = dispatchClientMetadata({ clientApp: CLIENT_APP_FIREFOX });
+    const root = shallowUntilTarget(
```
kumar303: even though it's only used in two places I think it would be nice to make a helper function for this like `shallowRender()` or something.

willdurand: ah that makes sense, thanks! 💯

PR #3589, opened 2017-10-20
Remove "Site Identifier" from listing pages

### LINKED ISSUE #3567: Remove "Site Identifier" from listing pages
After some discussion we've decided to not show the add-on ID on the listing page. We might bring it back later on, but just visible to certain people. As long as the numerical ID can be found somewhere in the HTML code of the listing page, admins can figure it out.

willdurand:
Fix #3567

---

This PR removes the site identifier in the meta info of an add-on. It adds this identifier as a data attribute (`data-site-identifier`) set on the `.Addon` element.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -464,6 +464,7 @@ export class AddonBase extends React.Component {
           'Addon--has-more-than-0-addons': numberOfAddonsByAuthors > 0,
           'Addon--has-more-than-3-addons': numberOfAddonsByAuthors > 3,
         })}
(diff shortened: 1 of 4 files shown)


src/amo/components/Addon/index.js
```diff
@@ -464,6 +464,7 @@ export class AddonBase extends React.Component {
           'Addon--has-more-than-0-addons': numberOfAddonsByAuthors > 0,
           'Addon--has-more-than-3-addons': numberOfAddonsByAuthors > 3,
         })}
+        data-site-identifier={addon ? addon.id : null}
```
kumar303: Good idea.

PR #3602, opened 2017-10-23
Add target=_blank attribute to disco links only

### LINKED ISSUE #3554: Clicking on author name on the detail page opens new tab
On an add-on detail page, when you click on the author's name underneath the name of the add-on, a new tab opens that takes you to the user profile page.

It should not open a new tab.

willdurand:
Fix #3554

---

This PR restricts the DOM Purify hook that adds `target="_blank"` to links (actually, "targetable" elements) to the `disco` project. Before it was used everywhere.

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -77,10 +77,10 @@ export function isValidClientApp(value, { _config = config } = {}) {
   return _config.get('validClientApplications').includes(value);
 }
 
-export function sanitizeHTML(text, allowTags = []) {
+export function sanitizeHTML(text, allowTags = [], _purify = purify) {
   // TODO: Accept tags to allow and run through dom-purify.
   return {
-    __html: purify.sanitize(text, { ALLOWED_TAGS: allowTags }),
+    __html: _purify.sanitize(text, { ALLOWED_TAGS: allowTags }),
   };
 }
 
--- a/src/disco/purify.js
+++ b/src/disco/purify.js
@@ -1,0 +1,17 @@
+import createDOMPurify from 'dompurify';
+
+import universalWindow from 'core/window';
+
+
+const discoPurify = createDOMPurify(universalWindow);
+
+discoPurify.addHook('afterSanitizeAttributes', (node) => {
+  // Set all elements owning target to target=_blank
+  // and add rel="noopener noreferrer".
+  if ('target' in node) {
+    node.setAttribute('target', '_blank');
+    node.setAttribute('rel', 'noopener noreferrer');
+  }
+});
+
+export default discoPurify;
--- a/src/disco/utils.js
+++ b/src/disco/utils.js
@@ -1,0 +1,7 @@
+import { sanitizeHTML } from 'core/utils';
+import purify from 'disco/purify';
+
+
+export const sanitizeHTMLWithNewTabLinks = (text, allowTags = []) => {
+  return sanitizeHTML(text, allowTags, purify);
+};
--- a/src/disco/utils.js
+++ b/src/disco/utils.js
@@ -2,6 +2,7 @@ import { sanitizeHTML } from 'core/utils';
 import purify from 'disco/purify';
 
 
-export const sanitizeHTMLWithNewTabLinks = (text, allowTags = []) => {
+export const sanitizeHTMLWithExternalLinks = (text, allowTags = []) => {
+  // This purify instance is configured with a hook to fix link targets.
   return sanitizeHTML(text, allowTags, purify);
 };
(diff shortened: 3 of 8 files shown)


src/disco/utils.js
```diff
@@ -0,0 +1,7 @@
+import { sanitizeHTML as baseSanitizeHTML } from 'core/utils';
+import purify from 'disco/purify';
+
+
+export const sanitizeHTML = (text, allowTags = [], _purify = purify) => {
```
muffinresearch: This could be renamed to reflect that the default behaviour is to modify all links to open in an new tab.

src/disco/purify.js
```diff
@@ -0,0 +1,13 @@
+import purify from 'core/purify';
```
muffinresearch: I wonder if there's a use-case for the external link treatment in amo in the future? That might mean we'd want two purify instances centrally rather than splitting the usage by application.

src/disco/utils.js
muffinresearch: Although I suppose that then risks that future usage would try and import this from here and modify the existing purify instance.

src/disco/purify.js
willdurand: I find very intrusive to force `target="_blank"` on links in general, and I don't think we have [valid use cases](https://css-tricks.com/use-target_blank/) for them on AMO. That's why I ended up with this separation.

src/disco/utils.js
willdurand: I can totally do this actually, but then I'll remove the `_purify` parameter. I only need to come up with a good name 😬

tests/unit/core/utils/test_index.js
```diff
@@ -839,3 +838,20 @@ describe('sanitizeUserHTML', () => {
...
+describe('sanitizeHTML', () => {
+  it('does not change links', () => {
+    const html = '<a href="http://example.org">link</a>';
+    expect(sanitizeHTML(html, ['a'])).toEqual({
+      __html: html,
+    });
+  });
+
+  // This is a built-in feature of dompurify.
+  it('removes `target` attribute on links', () => {
```
muffinresearch: Could this test fail if the disco test ran before it and modified the purify instance adding the hook?

src/disco/purify.js
muffinresearch: @willdurand yeah I agree with that - I can't see a use-case at the moment.

src/disco/utils.js
muffinresearch: `sanitizeHTMLNewTabLinks`?

willdurand: `sanitizeHTMLWithNewTabLinks`?

muffinresearch: > sanitizeHTMLWithNewTabLinks?

It's long-winded but very clear. 👍

tests/unit/core/utils/test_index.js
willdurand: ah oops, I guess yes.

src/disco/utils.js
willdurand: ⛵️

willdurand:
Thanks for the comments, I've updated the code with better naming and fixed the way we construct the `purify` objects.

src/core/utils/index.js
```diff
@@ -77,10 +77,10 @@ export function isValidClientApp(value, { _config = config } = {}) {
   return _config.get('validClientApplications').includes(value);
 }
 
-export function sanitizeHTML(text, allowTags = []) {
+export function sanitizeHTML(text, allowTags = [], _purify = purify()) {
```
muffinresearch: This will create a new instance for every call. 

I'm not sure how heavy this is but we might want to be sure this doesn't cause a perf issue.

willdurand: indeed 😞 

So I guess I have no choice but duplicating the creation of the purify instance.

src/disco/utils.js
```diff
@@ -0,0 +1,7 @@
+import { sanitizeHTML } from 'core/utils';
+import purify from 'disco/purify';
+
+
+export const sanitizeHTMLWithNewTabLinks = (text, allowTags = []) => {
```
kumar303: I think `sanitizeHTMLWithExternalLinks` reads a bit better since opening target-blank URLs in a new tab can be changed in one's browser settings.

src/disco/utils.js
```diff
@@ -0,0 +1,7 @@
...
+  return sanitizeHTML(text, allowTags, purify);
```
kumar303: It would help to add a comment saying that `purify` is configured to fix link targets since it's not totally obvious unless you think to dive into `disco/purify`.

PR #3606, opened 2017-10-23
Fix min height of language tool entries

### LINKED ISSUE #3454: Small issue in "Dictionaries & Language Packs" page for small browser dimensions
Steps to reproduce:
1. Load "Dictionaries & Language Packs" page on your device (or on your desktop resized FF 56 browser) https://addons-dev.allizom.org/en-US/firefox/language-tools/ but using Add-ons for Firefox option
2. Navigate through page

Expected results:

Actual results:
Small issue in gray background sections.

Notes/Issues:
- reproducing only for small browser dimensions, while using "Add-ons for Firefox" option, because it has results for both dictionaries and lang packs.

Screenshot for this issue:

willdurand:
Fix #3454

---

This PR fixes a visual glitch when some entries are not present in a given language on the "Dictionaries and Language Packs" page and on small screens. I used `31px` because that was the height for other regular rows.

Before:

screen shot 2017-10-23 at 19 55 23

After:

screen shot 2017-10-23 at 19 36 10

### DIFF
--- a/src/amo/components/LanguageTools/styles.scss
+++ b/src/amo/components/LanguageTools/styles.scss
@@ -2,6 +2,8 @@
 @import "~core/css/inc/mixins";
 @import "~ui/css/vars";
 
+$cell-padding: 6px;
+
 .LanguageTools {
   margin: 12px;
 }
@@ -41,6 +43,10 @@
 
   th,
   td {
-    padding: 6px;
+    padding: $cell-padding;
+  }
+
+  td {
+    min-height: $font-size-m + (2 * $cell-padding);
   }
 }


src/amo/components/LanguageTools/styles.scss
```diff
@@ -43,4 +43,8 @@
   td {
     padding: 6px;
   }
+
+  td {
+    min-height: 31px;
```
kumar303: Could you use line-height or reference a variable relating to font size maybe? It seems a bit fragile.

willdurand: `line-height` did not work, I guess that's because there is no content. I'll try to come up with a variable.

src/amo/components/LanguageTools/styles.scss
```diff
@@ -41,6 +43,10 @@
 
   th,
   td {
-    padding: 6px;
+    padding: $cell-padding;
+  }
+
+  td {
+    min-height: $font-size-m + (2 * $cell-padding);
```
kumar303: Nice

PR #3616, opened 2017-10-24
Allow to find addons by ID

### LINKED ISSUE #3610: Loading add-on page using numerical id instead of slug leads to blank page
STR:

willdurand:
Fix #3610

---

This PR allows to load add-ons by ID so that we can load a detail page by ID or slug.

The problem is that we rely on `normalizr` after having fetched the add-ons, which uses the add-on `slug` as key. When user wants to load a page by ID, all the API calls are correctly executed but then `normalizr` changes the response payload for the add-ons. The `addons` reducer then loads the add-ons by `slug`, not by `ID`. Because of this, the UI does not find the requested add-on and is in error state.

Pros: it is BC, safe and it works
Cons: it adds more data to the `addon` reducer

What do you think?

### DIFF
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -57,7 +57,9 @@ export type FetchAddonAction = {|
   |},
 |};
 
-export function fetchAddon({ errorHandler, slug }: FetchAddonParams): FetchAddonAction {
+export function fetchAddon(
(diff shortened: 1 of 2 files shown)


willdurand:
Yes, works for me! I close this PR and will open a new one with a server redirect implementation.

Issue #3618, opened 2017-10-24
Link based install button doesn't set a src


kumar303:
### Describe the problem and steps to reproduce it:

- Go to an add-on detail page
- Hover over the install button

### What happened?

You see a link to the XPI but without a `src` parameter in the URL

### What did you expect to happen?

The URL should include a `src` for statistics. See https://addons-server.readthedocs.io/en/latest/topics/api/download_sources.html#id1

### Anything else we should know?

Originally filed as https://github.com/mozilla/addons-frontend/issues/3600

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7).
The `src` param is listed in the URL while hovering over the Add to Firefox install button.
Postfix screenshot:
![addon](https://user-images.githubusercontent.com/15685960/31990662-6af1e90c-b97e-11e7-927a-0c1b95e3634b.png)

PR #3619, opened 2017-10-24
Add a src parameter to the install button URL

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3618

### DIFF
--- a/src/core/installAddon.js
+++ b/src/core/installAddon.js
@@ -167,6 +167,8 @@ export function makeMapDispatchToProps({ WrappedComponent, src }) {
 
     if (config.get('server')) {
       // Return early without validating properties.
+      // I think this returns early because a user agent prop isn't
+      // guaranteed on the server.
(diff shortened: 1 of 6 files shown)


src/core/installAddon.js
```diff
@@ -223,11 +226,12 @@ const userAgentOSToPlatform = {
  *
  * type FindInstallUrlParams = {|
  *   installURLs: $PropertyType<AddonType, 'installURLs'>,
+ *   src?: string,
  *   userAgentInfo: UserAgentInfoType,
  * |};
```
willdurand: Why is this in a comment btw?

kumar303: Because converting this file to Flow would be an arduous task

kumar303:
>  the src query parameter should not be added if there is no value

Done.

kumar303:
Thanks for the review!

Issue #3615, opened 2017-10-24
Gray glitch while tapping in the search field


ValentinaPC:
Steps to reproduce:
1. Load AMO-dev Extensions homepage https://addons-dev.allizom.org/en-US/firefox/extensions/ using your device
2. Tap on the search field

Expected results:
There are no display or layout issues.

Actual results:
Entire search field turns gray for a second.

Notes/Issues:
- no reproducing on desktop 
Reproducing on all AMO-servers using both FF56 and FF58 under Android 7.2.1.
Video for this issue:
![videotogif_2017 10 24_14 42 36](https://user-images.githubusercontent.com/15685960/31941325-162b16e6-b8ca-11e7-8911-049c39423095.gif)

vcarciu:
Verified as fixed : 
![telefonsearch](https://user-images.githubusercontent.com/1583842/32060434-6a997d86-ba77-11e7-910b-930638115cd6.gif)

PR #3623, opened 2017-10-24
Prevent a flicker when focusing search box on mobile

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3615

This only happens on mobile so I didn't make a screencast. I was able to reproduce the issue [with these steps](https://github.com/mozilla/addons-frontend/#configuring-an-android-device-for-local-development) before fixing it.

### DIFF
--- a/src/amo/components/SearchForm/styles.scss
+++ b/src/amo/components/SearchForm/styles.scss
@@ -17,6 +17,8 @@
   @include padding-end(6px);
   @include padding-start(28px);
 
+  // This prevents a flicker of black when focusing on mobile.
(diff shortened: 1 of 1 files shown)


src/amo/components/SearchForm/styles.scss
```diff
@@ -17,6 +17,8 @@
   @include padding-end(6px);
   @include padding-start(28px);
 
+  // This prevents a flicker of black when focusing on mobile.
```
muffinresearch: Good call adding a comment.

Issue #3600, opened 2017-10-23
Link based install button doesn't appear to send explicit tracking


muffinresearch:
The switch-based button had explicit tracking based on the various states of interaction. Having removed that to use a direct link button, we lost that tracking.

STR
- Go to an add-on detail page
- Open the network panel in Devtools
- Click the install button

Expected: a request is made to `https://www.google-analytics.com/r/collect`
Actual: no request is made

QA Notes:
- Please also make sure the Discopane install button does not have regressions. It should already be sending tracking events.

muffinresearch:
Since the type of installation has changed I think an appropriate fix for this would be to know we're on the details page and add `src=dp-btn-primary` which should mean we're still sending a stat for the same install as before. 

We can consider adding back more complete instrumentation at the same time as bringing back some of the features of mozAddonManager which is work that's still to be defined and specced out.

kumar303:
Sorry, I was a bit confused and thought this was comprised of two issues. I moved the `src` part to https://github.com/mozilla/addons-frontend/issues/3618

Maybe I'll just leave this one open so I can verify that we're still sending tracking events.

vcarciu:
@kumar303 : I cand see any request https://www.google-analytics.com/r/collect . Am I doing something wrong? : 
![analytics](https://user-images.githubusercontent.com/1583842/32106483-778f1a4c-bb34-11e7-9bfe-2dea71983a57.gif)

kumar303:
That's weird. What Firefox version are you using? I can see the request when installing that same add-on using 58.0a1 (2017-10-25) on Mac:
<img width="1631" alt="screenshot 2017-10-27 20 40 14" src="https://user-images.githubusercontent.com/55398/32130132-e9a8df28-bb57-11e7-9d97-b726c93de25d.png">

kumar303:
@vcarciu make sure you have *do not track* disabled.

vcarciu:
@kumar303 : Not sure what happened, do not track was not enabled, but anyway Today it is working fine : 
![ga](https://user-images.githubusercontent.com/1583842/32166668-410c5f1a-bd6f-11e7-80d9-ab39706c0e1b.png)

I observed that if you are trying to install a webext from about:addons search results, no google analytics request is made. Please let me know if this is ok (I used latest Nightly on AMO-dev): 
![noga](https://user-images.githubusercontent.com/1583842/32168039-b09832c4-bd73-11e7-9621-876c29813062.gif)

[...]

PR #3626, opened 2017-10-24
Track analytics of the standalone install button

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3600

- is this the right way to do it?
- the `mozAddonManager` does it like this but only *after* a successful installation. I'm not sure if we need to set this event apart.
- the old site seems to be using a different analytics framework so I'm not sure if this matches what's going on there.

### DIFF
--- a/tests/unit/core/TestInstallAddon.js
+++ b/tests/unit/core/TestInstallAddon.js
@@ -21,6 +21,7 @@ import {
   INSTALL_CATEGORY,
   INSTALL_CANCELLED,
   INSTALL_FAILED,
+  INSTALL_STARTED_CATEGORY,
   INSTALLED,
   INSTALLING,
   OS_ALL,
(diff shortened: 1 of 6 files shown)


muffinresearch:
Would it be difficult to listen to the success event with the basic button if mozAddonManager is available? One issue with this is we're now using the non-mozAddonManager fallback button as the main button so this does need to work in both cases.

As you say the instrumentation is most useful based on the success events since the cancellation at the permissions prompt will result in no complete install.

If adding the success listening is better to be done when the changes to the button happen (as mentioned in #3600 we're still in need of a spec for that and that's not going to happen pre-release) maybe it makes sense to move forward with this but to refer to this separately as a new "install start" event and then we can add that into the instrumentation of both the switch and the new button to better understand when users bail on the install.

kumar303:
@muffinresearch ok, I switched to an install started event so this is ready for another look. I also adjusted the InstallSwitch code to track an install started event just so we don't forget when we merge them.

> Would it be difficult to listen to the success event with the basic button if mozAddonManager is available

It's possible but I don't think it would be worth the effort. We need to merge the InstallSwitch functionality into the fallback button so we can do it later.

muffinresearch:
> It's possible but I don't think it would be worth the effort. We need to merge the InstallSwitch functionality into the fallback button so we can do it later.

Sounds good.

tests/unit/core/TestInstallAddon.js
```diff
@@ -935,6 +936,36 @@ describe(`${__filename}: withInstallHelpers`, () => {
...
+      return install(addon)
+        .then(() => {
+          // Even though the install() promise fails, it gets caught
+          // and resolved successfully.
+          sinon.assert.calledWith(fakeTracking.sendEvent, {
+            action: TRACKING_TYPE_EXTENSION,
+            category: INSTALL_STARTED_CATEGORY,
+            label: addon.name,
+          });
+          sinon.assert.calledOnce(fakeTracking.sendEvent);
```
muffinresearch: Nice

kumar303:
Good call. I added STR and QA notes to the issue

PR #3632, opened 2017-10-25
fix: Close dropdown on click

### LINKED ISSUE #3452: "More..." dropdown list is not closed if "Dictionaries & Language Packs" is clicked
Steps to reproduce:
1. Load AMO-dev page on your device https://addons-dev.allizom.org/en-US/android/
2. Navigate to "Dictionaries & Language Packs"

Expected results:
Corresponding page is loaded. The drop down list is closed.

Actual results:
Corresponding page is loaded. The drop down list is not closed.

Notes/Issues:
- reproducing on desktop too, while the hover is still on the option;
Video for this issue:
videotogif_2017 10 12_10 13 01

tofumatt:
Fix #3452.

Changes to use a JS hover to allow us to dismiss the menu on a click.

### DIFF
--- a/src/amo/components/SectionLinks/styles.scss
+++ b/src/amo/components/SectionLinks/styles.scss
@@ -67,7 +67,7 @@
   .DropdownMenu-button {
     display: block;
     margin-top: 2px;
-    padding: 0;
+    padding: 0 0 2px;
   }
 
   .DropdownMenu-button-text {
--- a/src/ui/components/DropdownMenu/index.js
+++ b/src/ui/components/DropdownMenu/index.js
@@ -1,8 +1,10 @@
 /* @flow */
+import { oneLine } from 'common-tags';
 import * as React from 'react';
 import classNames from 'classnames';
 import onClickOutside from 'react-onclickoutside';
 
+import log from 'core/logger';
 import Icon from 'ui/components/Icon';
 import DropdownMenuItem from 'ui/components/DropdownMenuItem';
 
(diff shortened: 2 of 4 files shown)


src/amo/components/SectionLinks/styles.scss
```diff
@@ -67,7 +67,7 @@
   .DropdownMenu-button {
     display: block;
     margin-top: 2px;
-    padding: 0;
+    padding: 0 0 2px;
```
tofumatt: This prevents any issues when moving the mouse from the button to the menu–if this isn't here it might cause a `mouseLeave` event to be fired and the menu will disappear.

muffinresearch:
Was there a reason to not use onClickOutside?

willdurand:
> Was there a reason to not use onClickOutside?

isn't it because we are actually inside the component? This HOC is used if I remember correctly.

tofumatt:
Yeah, we're clicking inside the component. The issue is clicks *inside* the component on a link that should dismiss the dropdown (and *would* in tradition navigation because it would reload the page, resetting the dropdown state) don't dismiss it so it hangs out there until the user hovers away. It's even worse on mobile where the dropdown takes up a lot of screen and there is no hover, so a user needs to click "Language Tools" and then click elsewhere to get it to dismiss.

src/ui/components/DropdownMenu/index.js
```diff
@@ -34,18 +36,44 @@ export class DropdownMenuBase extends React.Component<Props, State> {
     }));
   }
 
+  handleOnClickForLinks = (event: SyntheticEvent<any>) => {
+    // If a link inside the menu is clicked, we should close the dropdown.
+    // See: https://github.com/mozilla/addons-frontend/issues/3452
+    if (event.target && event.target.tagName === 'A') {
```
willdurand: Are we 100% sure it is a uppercase tag name? I think so but we never know.

src/ui/components/DropdownMenu/styles.scss
```diff
@@ -11,22 +11,6 @@
...
-      .DropdownMenu-items {
-        display: block;
-      }
-
-      .Icon-inverted-caret {
-        opacity: 1;
-      }
-    }
-  }
-
```
willdurand: Why is this removed? Not needed anymore?

willdurand:
STRs:

- go with the mouse on the menu button/title
- click on it

This patch in action:



It actually closes the dropdown menu. It is opened on hover, but when you click on the menu button/title, it is closed.

- Current behavior on -dev: (I click many times when the cursor is on the button/title)



Same steps keep the dropdown menu open.

src/ui/components/DropdownMenu/styles.scss
tofumatt: Yeah, because the hover state is now managed in JS.

tofumatt:
Actually, that was intentional (there is even a test case for it: https://github.com/mozilla/addons-frontend/pull/3632/files#diff-d411d62dcb1f4ae57c93cac125dba2c8R99).

This means a user can hide/show the menu with a click, which they may want to do. I thought there was no harm in it as the main menu items are never clickable links but just show/hide actions.

willdurand:
hmm does that still work on mobile devices?

I actually triggered this bug the first time I interacted with the menu, because I am never sure of the `hover` behavior. I tend to click to things to keep them open.

tofumatt:
It still works on mobile because the first click opens it. I guess I could make it so that hover sets a different state and then the first click doesn't do anything. But another click would still close it? Would that work for you?

willdurand:
If that's quick and easy, yes. Otherwise let's forget about it.

muffinresearch:
> Yeah, we're clicking inside the component. The issue is clicks inside the component on a link that should dismiss the dropdown (and would in tradition navigation because it would reload the page, resetting the dropdown state) don't dismiss it so it hangs out there until the user hovers away. It's even worse on mobile where the dropdown takes up a lot of screen and there is no hover, so a user needs to click "Language Tools" and then click elsewhere to get it to dismiss.

Aha right, that makes perfect sense.

willdurand:
@tofumatt anything I can do here?

src/ui/components/DropdownMenu/index.js
tofumatt: That's how the DOM API works and React's `SyntheticEvent` normalizes the content across browsers, so it should always be uppercase.

tofumatt:
I don't there's anything needing to be done other than a review? Does it look good to you or is there still stuff that needs addressing?

willdurand:
I was waiting for a reply to:

> [tofumatt] I guess I could make it so that hover sets a different state and then the first click doesn't do anything. But another click would still close it? Would that work for you?

> [willdurand] If that's quick and easy, yes. Otherwise let's forget about it.

tofumatt:
Oh, shoot, right. Okay I can do that.

tofumatt:
@willdurand Okay, I've patched it so the first click after a hover won't close the menu... but clicks afterward do, which I think is sensible behaviour.

Issue #3627, opened 2017-10-25
Link to review guidelines from the review dialog


dessant:
The legacy frontend has some useful tips for users which probably prevent at least a couple of invalid reviews. These tips are missing from the new frontend, there should be a prominent [link](https://addons.mozilla.org/en-US/review_guide) with a short paragraph if the new design does not allow a large block of text.

Legacy site:
![screenshot from 2017-10-25 06-41-04](https://user-images.githubusercontent.com/7513068/31979940-07bcce28-b952-11e7-8324-d8dbadd4babf.png)

New site:
![screenshot from 2017-10-25 06-42-12](https://user-images.githubusercontent.com/7513068/31979945-0fe963e0-b952-11e7-8b5a-cdd11f0b36d8.png)

kumar303:
That's a good idea, thanks. We also had a UX mock-up with tips in it but I forgot that we had copy already. We probably just need to make the existing tips a little more concise so they are easy to scan at a glance.

kumar303:
Thanks again for filing this. Showing the review guidelines somehow slipped through the cracks.

A fix has landed (on our dev server only) that makes the review entry form look like this:

<img width="320" alt="screenshot 2017-10-25 17 37 18" src="https://user-images.githubusercontent.com/55398/32064650-0cc2201e-ba40-11e7-82a9-2527230c554a.png">
<img width="321" alt="screenshot 2017-10-25 17 37 30" src="https://user-images.githubusercontent.com/55398/32064658-13376882-ba40-11e7-89ac-9a51910dcbb4.png">

I did not add any explicit tips to that form since that will need a bit more UX thought. The guidelines page does have them though. You can see it here if you click on 'view mobile site' : https://addons.mozilla.org/en-US/review_guide

dessant:
I like the outcome, thanks for the quick turnaround! ❤️

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7, Android 7.2.1)
Postfix video:
![rev g](https://user-images.githubusercontent.com/15685960/32092529-a9f574d2-bb01-11e7-969b-a00e9cd6df6e.gif)

PR #3635, opened 2017-10-25
Link to review writing guidelines

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3627

Before:

screenshot 2017-10-25 14 37 46

After:

screenshot 2017-10-25 17 37 18
screenshot 2017-10-25 17 37 30
screenshot 2017-10-25 17 37 55

### DIFF
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -8,7 +8,7 @@ import { compose } from 'redux';
 
 import { submitReview } from 'amo/api/reviews';
 import { setDenormalizedReview, setReview } from 'amo/actions/reviews';
-import { refreshAddon } from 'core/utils';
+import { refreshAddon, sanitizeHTML } from 'core/utils';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import defaultLocalStateCreator, { LocalState } from 'core/localState';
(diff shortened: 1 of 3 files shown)


kumar303:
Ultimately I think this overlay should be replaced. See https://github.com/mozilla/addons-frontend/issues/3636

src/amo/components/AddonReview/index.js
```diff
@@ -164,30 +164,41 @@ export class AddonReviewBase extends React.Component<Props, State> {
     }
 
     let placeholder;
-    let prompt;
+    let promptText;
     if (review.rating && review.rating > 3) {
-      prompt = i18n.gettext(
-        'Tell the world why you think this extension is fantastic!'
+      promptText = i18n.gettext(
+        'Tell the world why you think this extension is fantastic! Please read our %(linkStart)sreview guidelines%(linkEnd)s first.'
```
willdurand: This is more than 80chars :)

src/amo/components/AddonReview/index.js
```diff
@@ -164,30 +164,41 @@ export class AddonReviewBase extends React.Component<Props, State> {
...
       );
       placeholder = i18n.gettext(
         'Tell us what you love about this extension. Be specific and concise.'
       );
     } else {
-      prompt = i18n.gettext('Tell the world about this extension.');
+      promptText = i18n.gettext(
+        'Tell the world about this extension. Please read our %(linkStart)sreview guidelines%(linkEnd)s first.'
```
willdurand: same here

src/amo/components/AddonReview/index.js
kumar303: I forgot that we use babel to extract gettext messages so it's probably safe to use back ticks. I'll update it.

kumar303:
> This blue overlay is ugly.

It will probably go away in https://github.com/mozilla/addons-frontend/issues/3636

kumar303:
I made the prompt wording a bit better (screenshots updated).

src/amo/components/AddonReview/index.js
```diff
@@ -164,30 +164,43 @@ export class AddonReviewBase extends React.Component<Props, State> {
...
+        Please follow our %(linkStart)sreview guidelines%(linkEnd)s.`
+      );
       placeholder = i18n.gettext(
         'Tell us about your experience with this extension. ' +
         'Be specific and concise.'
       );
     }
 
+    const prompt = i18n.sprintf(promptText, {
+      linkStart: '<a href="/review_guide">', linkEnd: '</a>',
```
muffinresearch: We could really use a JS API for what  our `Link` component does - since this will presumably be a server redir + render with no locale?

This also looks like a case where we should consider opening the guidelines in a new tab since it will replace the current action the user is doing. That is unless clicking back recreates the modal and if the back button works then this can be ignored.

Bear in mind DOMPurify will remove target attrs by default and if you add target="_blank" you'll need rel="noopener noreferrer" to not leak the window object (albeit this is way less of an issue linking to ourselves but probably good to be consistent in that pattern in case it got copied for an external link later).

muffinresearch: We can always file any relevant issues as follow-ups to keep moving forward.

kumar303: I considered target=blank but I could not figure out how to get `sanitizeHTML()` to allow that. Is there a trick?

willdurand: I dont think it is feasible because of `dompurify`

Issue #3598, opened 2017-10-23
A review can be flagged for multiple reasons


ValentinaPC:
Steps to reproduce:
1. Load All reviews page for an add-on with reviews i.e. https://addons-dev.allizom.org/en-US/firefox/addon/imgur-uploader/reviews/
2. Click on all 3 flagging options

Expected results:
Review can be flagged only once.

Actual results:
All 3 flagging options can be picked for the same review.

notes/Issues:
- _flag link is also not changing in "marked as flagged" or something like this giving the user the chance to flag that review as many times as he wants_;
- in Moderated reviews page only the first selected flagging option is stored 

Verified on AMo-dev FF56(Win 7).
Video for this issue:
![flag0v](https://user-images.githubusercontent.com/15685960/31880413-1ec193ca-b7e9-11e7-88bd-7c374d9c4c10.gif)

kumar303:
This is working as intended. However, I too am not sure if the intention was correct. I think we need some UX and product related guidance on what to do here.

kumar303:
> in Moderated reviews page only the first selected flagging option is stored

What do you mean by *moderated* reviews page? What do you mean be *stored*?

ValentinaPC:
@kumar303 : 
- the initial title was more general because not only was possible to flag multiple times a review, but the fact that if I select 2 options, only the one action is recorded in moderated review queue (https://reviewers.addons-dev.allizom.org/en-US/editors/queue/reviews)
Example of selecting all 3 flagging options:
![2017-10-24_0913](https://user-images.githubusercontent.com/15685960/31927235-d74b1170-b89b-11e7-8ffb-8af29a159dcf.png)

kumar303:
Ah! I didn't know about that page. Thanks. If only one flag reason is shown in the reviewer tools then the UI should match that.

ValentinaPC:
We should check the intended behavior from Reviewer Tools with @wagnerand :), just to be sure.
Thanks guys!

kumar303:
> in Moderated reviews page only the first selected flagging option is stored

Correct me if I'm wrong but I observed that the last selected flagging reason is shown in the reviewer tools. Because of that I allowed the user to switch their reason for flagging the review. They can still only choose one. This would allow someone to change their mind or make a correction if they accidentally flagged for the wrong reason.

ValentinaPC:
Verified this on AMO-dev FF56(Win 7).
[...]

PR #3639, opened 2017-10-25
Only flag a review for a single reason

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3598

* After you flag a review, the prompt changes from *Flag* to *Flagged*
* The UI indicates that you can only flag a review for a single reason
* You can still change your mind and switch the reason

Screencast:

addons-frontend-flag-review mov


kumar303:
The coverage reduction doesn't seem related to my patch. I think it happened because I deleted some code.

PR #3644, opened 2017-10-26
Update hero colors/patterns

### LINKED ISSUE #3613: Hero banners have no background images
Steps to reproduce:
1. Load AMO-dev homepage https://addons-dev.allizom.org/en-US/firefox/
2. Observe hero banners

Expected results:

Actual results:
The banners have no background images.

Notes/issues:
Screenshot for this issue:

willdurand:
Fix #3613

---

This PR adds new colors and patterns to the hero banner and fixes the font style. It is not 100% random (we could have more [combinations](https://en.wikipedia.org/wiki/Combination)) but I made sure to have distinct colors/patterns and a pretty understandable code. 

I used the order of the sections computed in the reducer as a source of randomness because it is pretty reliable and does not differ between client and server.

## Screenshots

screen shot 2017-10-26 at 12 19 23
screen shot 2017-10-26 at 12 18 55
screen shot 2017-10-26 at 12 19 05
screen shot 2017-10-26 at 12 19 18
screen shot 2017-10-26 at 12 20 59
screen shot 2017-10-26 at 12 20 55


src/amo/components/HomeHeroBanner/styles.scss
```diff
@@ -1,19 +1,160 @@
-@import "~photon-colors/colors";
+/* stylelint-disable indentation */
+$colors: ("blue", #228ae6, #1c7cd6)
```
willdurand: it is more readable that way, that is why I disabled the style rule.

src/amo/components/HomeHeroBanner/styles.scss
```diff
@@ -1,19 +1,160 @@
...
 
-.HomeHeroBanner {
-  .HeroSection-styleName--Home-featured-extensions {
-    background: $teal-70;
-  }
+/* stylelint-enable indentation */
 
-  .HeroSection-styleName--Home-youtube-high-definition {
-    background: $green-70;
+@function encodeColor($string) { /* stylelint-disable-line */
```
willdurand: it seems `@function` and `@return` (below) are not enabled by default. It is possible to update the stylelint config though: https://github.com/stylelint/stylelint/issues/1612.

src/amo/components/HomeHeroBanner/styles.scss
```diff
@@ -1,19 +1,160 @@
...
+}
+
+@mixin polka-dots($foreground, $background) {
+  background-color: $background;
+  background-image: url("data:image/svg+xml,%3Csvg width='20' height='20' viewBox='0 0 20 20' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='#{encodeColor($foreground)}' fill-opacity='0.4' fill-rule='evenodd'%3E%3Ccircle cx='3' cy='3' r='3'/%3E%3Ccircle cx='13' cy='13' r='3'/%3E%3C/g%3E%3C/svg%3E");
+}
+
+@mixin floor-tile($foreground, $background) {
+  background-color: $background;
+  background-image: url("data:image/svg+xml,%3Csvg width='30' height='30' viewBox='0 0 30 30' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M0 10h10v10H0V10zM10 0h10v10H10V0z' fill='#{encodeColor($foreground)}' fill-opacity='0.4' fill-rule='evenodd'/%3E%3C/svg%3E");
```
muffinresearch: If you point at the file, this should be inlined as base64 encoded svg by default.

willdurand: I actually need to modify the `fill` property, that's why I chose to inline them here.

muffinresearch:
Looks good.

- There's no hover styles at all at the moment, might be nice to have add them as part of this I think the spec in the original issue included some guidance on the colours there.
- For me the contrast is too between the main colour and the pattern is a little too low - on some of the colours the background pattern is really hard to see.
- The randomization seems to be server-side only - but it's like that currently so we could always look to do something with that later.

src/amo/components/HomeHeroBanner/styles.scss
muffinresearch: Oh I see you're using vars. 

The only other way would have been to use the SVG with transparency instead.

src/amo/components/HomeHeroBanner/styles.scss
muffinresearch: Yep it would be nice to add that generally - can be a follow-up if that's easier.

muffinresearch:
Here's an example of the contrast I see - Purple and blue on the primary content seem to be the ones that stand out as hard to see:

add-ons_for_firefox

add-ons_for_firefox

willdurand:
> For me the contrast is too between the main colour and the pattern is a little too low - on some of the colours the background pattern is really hard to see.

Ah good catch, I just fixed that.

src/amo/components/HomeHeroBanner/styles.scss
willdurand: We had a config file already, so I have just updated it.

muffinresearch:
Contrast is better still a little low on my internal screen. Here's the blue + purple from before:

add-ons_for_firefox

add-ons_for_firefox

muffinresearch:
@tofumatt how's this looking for you?

I'm thinking assuming there are no other review nits - we could ship this as a cherry-pick and look at additional contrast/colour tweaks for the release if required.

willdurand:
Normal:

screen shot 2017-10-26 at 13 48 03

Main hero section is hovered:

screen shot 2017-10-26 at 13 48 06

Issue #3421, opened 2017-10-09
Don't store addons in top-level of addons reducer


tofumatt:
Right now we store addons in the top of the addons reducer rather than organzing them in a dictionary under a key of the main reducer state object. Right now we have:

```js
{
  addonsReducer: {
    addonSlug: { [...addon] },
    anotherAddonSlug: { [...addon] },
  },
  [...]
}
```

But this prevents us from storing any other info in this reducer, like languagePack slugs as we wanted to in https://github.com/mozilla/addons-frontend/pull/3289#discussion_r143288040

We want:

```js
{
  addonsReducer: {
    bySlug: {
      addonSlug: { [...addon] },
      anotherAddonSlug: { [...addon] },
    },
    otherReducerData,
  },
  [...]
}
```

In addition and since #3657, we store an `addon` twice in the state. We should not do this. Instead, we should store the `addon` only once and reference it by ID in all other parts of the state (a `slug` could change).

willdurand:
> In addition and since #3657, we store an addon twice in the state. We should not do this. Instead, we should store the addon only once and reference it by ID in all other parts of the state (a slug could change).

Yeah, this is causing:

1. twice the amount of data supplied to the `LanguageTools` component and many React warnings because keys are duplicate
2. the languages packs to be indexed twice with numeric values: one being the `id` and the other being the index (natural integer) because it uses `loadAddonResults()`, which does not rely on normalized (normalizr) data

kumar303:
Since this issue doesn't cover it fully, here is my wish list for the implementation :)

- we should only store the add-on object once, by ID
- all other mappings should link to the add-on. For example, `bySlug` should be a mapping like `{ slugOne: 1234, slugTwo: 5432 }` where the key is the add-on slug and the value is the ID of the add-on.

willdurand:
I have removed the `qa: not needed` badge because the patch is sensitive. For QA, please make sure:

- Disco Pane continues to work as expected
- AMO continues to work as expected (add-ons load fine, add-ons can be installed)

vcarciu:
@willdurand : not sure if this is related to this bug , but searches in disco pane are not working as expected for me in -dev.  With all disco pane prefs set to -dev, if you search for "firefox" for example , there are no results, even if we have many compatible webextensions found with same search term on AMO. Please let me know if you think that it is related to this issue(you can see in video that some searchterms are working but another one not) : 
[...]

PR #3657, opened 2017-10-26
Redirect add-on detail page when loaded by ID

### LINKED ISSUE #3610: Loading add-on page using numerical id instead of slug leads to blank page

willdurand:
Fix #3610

---

This PR allows to send a server redirect (in the node server) from the react app when an add-on slug is numeric (i.e. it is its ID and not its usual slug). There is not client side support.


tests/unit/amo/components/TestAddon.js
```diff
@@ -404,6 +406,24 @@ describe(__filename, () => {
...
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+    renderComponent({
+      // We set the numeric `id` as slug.
+      params: { slug: addon.id }, store,
+    });
+
+    sinon.assert.calledWith(fakeDispatch, sendServerRedirect({
+      status: 301,
+      url: `/en-US/${clientApp}/addon/${addon.slug}/`,
+    }));
```
kumar303: This could benefit from `sinon.assert.calledOnce()` to make sure it was the only thing dispatched.

src/amo/reducers/redirectTo.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const sendServerRedirect = (
+  { status, url }: SendServerRedirectParams
+): SendServerRedirectAction => {
+  if (!status) {
+    throw new Error('status is required');
+  }
+  if (!url) {
+    throw new Error('url is required');
+  }
```
kumar303: I like how you put "server" in the name. It would also be nice if this had some extra logging when called client side -- just another thing to dissuade someone from using it or to catch bugs. 

```js
if (!config.get('server')) {
  log.warn('sendServerRedirect() currently does nothing when run from client code');
}
```

No one should ever see this message (hopefully!).

tests/unit/core/reducers/test_addons.js
```diff
@@ -42,18 +42,22 @@ describe(__filename, () => {
         ...createInternalAddon(anotherFakeAddon),
         isRestartRequired: false,
       },
+      [anotherFakeAddon.id]: {
+        ...createInternalAddon(anotherFakeAddon),
+        isRestartRequired: false,
```
kumar303: This code predates `createInternalAddon()`. You can change them both like this:

```diff
diff --git a/tests/unit/core/reducers/test_addons.js b/tests/unit/core/reducers/test_addons.js
index 6eee9e50..23c783cd 100644
--- a/tests/unit/core/reducers/test_addons.js
+++ b/tests/unit/core/reducers/test_addons.js
@@ -36,16 +36,11 @@ describe(__filename, () => {
     const newState = addons(firstState,
       loadAddons(createFetchAddonResult(anotherFakeAddon).entities));
 
+    const internalAddon = createInternalAddon(anotherFakeAddon);
     expect(newState).toEqual({
       ...firstState,
-      [anotherFakeAddon.slug]: {
-        ...createInternalAddon(anotherFakeAddon),
-        isRestartRequired: false,
-      },
-      [anotherFakeAddon.id]: {
-        ...createInternalAddon(anotherFakeAddon),
-        isRestartRequired: false,
-      },
+      [anotherFakeAddon.slug]: internalAddon,
+      [anotherFakeAddon.id]: internalAddon,
     });
   });
 

```

tests/unit/core/server/testServer.js
```diff
@@ -262,5 +272,31 @@ describe(__filename, () => {
...
+
+      const client = testClient({
+        routes: amoRoutes,
+        appSagas: addonsSaga,
+        sagaMiddleware,
+        store,
+      });
+
+      const response = await client
+        .get(`/en-US/firefox/addon/${addon.id}/`)
```
kumar303: I'm not a fan of creating functional tests like this. I know it's nice for regressions but they are always a headache to maintain. This test suite is meant to be a unit test of server functionality (but naturally it integrates a few lower level units).

What I'd like to see instead is a stub route added for a brand new component (just for testing) that dispatches the server redirect action in its `componentWillMount()` code. The test should look more like this: https://github.com/willdurand/addons-frontend/blob/aa085219b8379a9fa24bcd0bc02d6290f3decc4a/tests/unit/core/server/testServer.js#L106

This test will independently verify that any component which dispatches a redirect action will trigger the server redirect. Since we already have a unit test to guarantee that `<Addon>` dispatches this action then we can trust that it will work.

willdurand: I agree. Ultimately, I'd like to see more integration tests like this, or even better, some end-to-end tests.

src/core/server/base.js
```diff
@@ -261,69 +270,76 @@ function baseServer(routes, createStore, {
...
-              <Provider store={store} key="provider">
-                <ReduxAsyncConnect {...renderProps} />
-              </Provider>
-            </I18nProvider>
-          );
-
-          const errorPage = store.getState().errorPage;
-          if (errorPage && errorPage.hasError) {
-            log.info(`Error page was dispatched to state: ${errorPage.error}`);
-            throw errorPage.error;
```
kumar303: I see that you deleted this. You are brave 😁  Did you find and delete the tests too? Maybe it wasn't tested, I'm not sure.

src/amo/store.js
```diff
@@ -43,6 +44,7 @@ export default function createStore(initialState = {}) {
       installations,
       landing,
       reduxAsyncConnect,
+      redirectTo,
```
kumar303: It would be good for sanity to also add this to the Discopane. We won't need it there but since Discopane also runs from `server/base.js` it will make sure that `store.getState().redirectTo` is never undefined.

src/core/reducers/addons.js
```diff
@@ -287,7 +289,14 @@ export default function addonsReducer(
...
       const newState = { ...state };
       Object.keys(addons).forEach((key) => {
-        newState[key] = createInternalAddon(addons[key]);
+        const addon = createInternalAddon(addons[key]);
+
+        // We index add-ons by id and slug to be able to retrieve them by any
+        // of these parameters. This is needed to redirect an add-on detail
+        // page loaded by ID.
+        // See: https://github.com/mozilla/addons-frontend/issues/342
+        newState[addon.id] = addon;
```
kumar303: Could you file an issue to stop storing `addon` in state twice and link it to a TODO comment here? It will be an annoying patch to write but we should eventually store the `addon` only once and reference it by ID in all other parts of the state.

src/amo/store.js
willdurand: ah that makes sense, even though I think there is a check for that.

src/core/reducers/addons.js
willdurand: hmm isnt it this one? https://github.com/mozilla/addons-frontend/issues/3421

src/core/server/base.js
willdurand: I did not find any tests.. I wanted to get rid of more code to earn a "more red than green" badge but it seems we still use some of this `errorPage`, and I prefer to do that in a follow-up patch.

src/core/server/base.js
```diff
@@ -261,69 +270,76 @@ function baseServer(routes, createStore, {
...
+      try {
+        await loadOnServer({ ...renderProps, store });
+
+        // eslint-disable-next-line global-require
```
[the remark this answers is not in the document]
willdurand: thanks!

src/core/reducers/addons.js
kumar303: That issue is different but related. I think it would be fine to edit the description of that issue to make it work for both :)

willdurand: indeed!

src/core/server/base.js
willdurand: This change is now reverted in this PR after our discussion on IRC.

PR #3667, opened 2017-10-27
Set min-height on the Hero link, not the container

willdurand:
Before:

screen shot 2017-10-27 at 20 42 49

After:

screen shot 2017-10-27 at 20 38 15

### DIFF
--- a/src/ui/components/HeroSection/styles.scss
+++ b/src/ui/components/HeroSection/styles.scss
@@ -5,7 +5,6 @@
 
 .HeroSection {
   border-radius: $border-radius-default;
-  min-height: 128px;
   position: relative;
 }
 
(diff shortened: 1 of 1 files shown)


kumar303:
Your *after* screenshot shows a stretched download button but I don't see it locally with the patch.

kumar303:
Is there an issue logged about this? I can't quite tell what it was fixing.

willdurand:
Yes I noticed that on the screenshot, I think it is a HMR glitch... Could not reproduce and the change cannot impact that part of the UI anyway.

willdurand:
No, no issue. It fixes an issue on browsers other than Firefox. The text on the small heros should be aligned on the bottom.

Issue #3669, opened 2017-10-27
All themes are disabled and cannot be installed


kumar303:
### Describe the problem and steps to reproduce it:

* Go to a theme detail page using a desktop browser or Android

### What happened?

The theme is disabled

![screenshot-2017-10-27 eyeballs2 - add-ons for firefox](https://user-images.githubusercontent.com/55398/32126354-d0720ace-bb35-11e7-9e43-0d532269d178.png)

### What did you expect to happen?

It should be enabled

### Anything else we should know?

Regression from https://github.com/mozilla/addons-frontend/issues/3649

vcarciu:
Verified as fixed : 
![themeunavailable](https://user-images.githubusercontent.com/1583842/32232313-743b9742-be60-11e7-8d12-db329251fce8.png)

PR #3670, opened 2017-10-27
Only require extensions to declare clientApp compatibility

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3669

This fixes not only themes but also dictionaries and search providers.

### DIFF
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -19,7 +19,17 @@ import log from 'core/logger';
 export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
   let maxVersion = null;
   let minVersion = null;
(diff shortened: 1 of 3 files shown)


src/core/utils/compatibility.js
```diff
@@ -31,7 +48,6 @@ export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
       _log.info(oneLine`addon is type ${ADDON_TYPE_OPENSEARCH}; no
```
willdurand: Does this message still apply? It seems this `if` block does not do anything anymore.

kumar303: It never really did anything actually. I should remove it. Good call.

PR #3672, opened 2017-10-27
ltr-rtl string to constant change

### LINKED ISSUE #3362: i18n `getDirection()` helper should rely on constants
In `core/i18n/utils`, the `getDirection()` function should not use hardcoded values for `ltr` and `rtl` because then we need to test with the hardcoded values when using this helper:

We should rather have `LTR` and `RTL` as constants, and use them. When working on this issue, please "grep" for `'rtl'`, `'ltr'` and `getDirection()` and update the code accordingly.

PhillypHenning:
Fixes #3362 

## Description
Changes all 'rtl' and 'ltr' strings to instead use constants that have been added to `core/constants.js`

## Local Test
`yarn test-once` && `yarn lint`

pr1-postlint

## Comments

Is testing required for this change and if so where would I be placing the testing for this?

### DIFF
--- a/locale/he/LC_MESSAGES/amo.po
+++ b/locale/he/LC_MESSAGES/amo.po
@@ -849,9 +849,8 @@ msgid "See more trending themes"
 msgstr "צפייה בערכות נושא נוספות שמובילות לאחרונה"
 
 #: src/amo/components/LandingPage/index.js:167
-#, fuzzy
 msgid "See more top rated themes"
-msgstr "ערכות הנושא המדורגות ביותר"
+msgstr ""
 
 #: src/amo/components/LandingPage/index.js:197
 msgid "Change your browser's appearance. Choose from thousands of themes to give Firefox the look you want."


src/core/constants.js
```diff
@@ -212,3 +212,7 @@ export const CATEGORY_COLORS = {
   [ADDON_TYPE_OPENSEARCH]: 12,
   [ADDON_TYPE_THEME]: 12,
 };
+
+// Left-to-right and right-to-left convience variables
```
tofumatt: This should be convenience 😀

tofumatt: But I don’t think this comment adds anything so it can be removed.

src/ui/components/Carousel/index.js
```diff
@@ -64,7 +65,7 @@ export class CarouselBase extends React.Component {
           <NukaCarousel
             autoplay
             autoplayInterval={4000}
-            cellAlign={getDirection(i18n.lang) === 'ltr' ? 'left' : 'right'}
+            cellAlign={getDirection(i18n.lang) === LTR ? 'left' : 'right'}
```
tofumatt: This file doesn’t exist anymore; you’ll want to update your code by pulling the newest master and rebasing on top of master. It’s why you have conflicts.

PhillypHenning:
@tofumatt no problem, I fix it up tonight.

PhillypHenning:
Hey @tofumatt 
So I'm running into a few issues during my `make test` regardless of if it is a fresh fetch and pull. I would love some help looking at these if you have time.

Issue #3603, opened 2017-10-23
statistics link should be there for all user for Addon.public_stats == true


eviljeff:
STR
* Create an add-on and mark its stats as public
* View the add-on detail page
* Make sure you are logged out
* Refresh the page

You should still see a link to the stats but you do not.

(On the backend the API data exposes this as `public_stats` = `true`)

kumar303:
> The stats link should also be shown for all add-ons where the user has `Stats:View` also, though this is arguably a lower priority fix.

Moved to https://github.com/mozilla/addons-frontend/issues/3604

ValentinaPC:
@kumar303 : I have a question - for admins, in old pages, we displayed the stats no matter if they where set public or not in Dev Hub, shouldn't be the case of the new frontend?
Thanks!

kumar303:
> for admins, in old pages, we displayed the stats no matter if they where set public or not

Yes, the new site should do that too but it was a little harder to implement so we are doing it in a separate issue: https://github.com/mozilla/addons-frontend/issues/3604

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7)
Postfix screenshot:
![2017-11-02_1353](https://user-images.githubusercontent.com/15685960/32324793-3badb088-bfd5-11e7-8bae-6a96f3b054b6.png)

PR #3679, opened 2017-10-30
I3603 - fixes public stats

tsl143:
Fixes: #3603

Before
screen shot 2017-10-30 at 6 10 22 pm

After
screen shot 2017-10-30 at 6 09 27 pm

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -57,7 +57,7 @@ export class AddonMoreInfoBase extends React.Component<Props> {
     return this.renderDefinitions({
       homepage,
       supportUrl,
-      statsLink: addon && isAddonAuthor({ addon, userId }) ? (
+      statsLink: addon && (isAddonAuthor({ addon, userId }) || addon.public_stats) ? (
         <Link
           className="AddonMoreInfo-stats-link"
           href={`/addon/${addon.slug}/statistics/`}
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonMeta/index.js
```diff
@@ -30,7 +31,14 @@ export class AddonMetaBase extends React.Component<Props> {
       userCount = null;
       userTitle = i18n.gettext('Users');
     } else if (averageDailyUsers) {
-      userCount = i18n.formatNumber(averageDailyUsers);
+      userCount = addon && addon.public_stats ? (
```
tofumatt: I removed the stats from the user count in this part of the page; there is a link to the statistics dashboard in the `More Info` card, that's where this link should be added: https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonMoreInfo/index.js#L60

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -27,6 +31,23 @@ describe(__filename, () => {
     );
   }
 
+  function mountRender({
```
tofumatt: I don't think you should need `mount` for this test; obviously you'll want to change it to be in `AddonMoreInfo`, but this should work with `shallowUntilRender`.

kumar303:
Hi @tsl143 - thanks for getting this patch started, we really appreciate it. Since this issue was high-ish priority (since all the [mvp](https://github.com/mozilla/addons-frontend/issues?q=is%3Aissue+is%3Aopen+label%3A%22priority%3A+mvp%22) issues are finished) I moved it to https://github.com/mozilla/addons-frontend/pull/3701 where I wanted a few minor changes just to make the intention of the test more explicit. I know you've been very fast to resolve change requests but we have a release cut off early on Tuesday so I want to make sure this gets in.

PR #3680, opened 2017-10-30
fix(package): update react-super-responsive-table to version 4.0.2

willdurand:
Closes #3656

Huge bump but it seems nothing has really changed.. only build/release stuff.


willdurand:
Actually, it needs some style update since the bundled CSS has changed and now it looks like this:

screen shot 2017-10-30 at 14 36 04

muffinresearch:
@willdurand thanks for checking by spinning this up!

PR #3681, opened 2017-10-30
fix(package): update knuth-shuffle to version 1.0.8

willdurand:
Closes #3675

Nothing fancy obviously, the author has moved his code to another Git platform.


willdurand:
Tried locally and we still get random stuff 🎉

Issue #3664, opened 2017-10-27
New review submitted on all-reviews page is not visible


kumar303:
### Describe the problem and steps to reproduce it:

* Go to an add-on detail page
* Click on *Read N reviews* to see the list of all reviews
* Click a star (from the left sidebar) to begin reviewing the add-on
* Write a review, click Submit

### What happened?

The review does not show up on the page

### What did you expect to happen?

The review should show up on the page

### Anything else we should know?

If you reload the page you can see the review

AlexandraMoga:
Verified as fixed on AMO-dev with FF56 (Win10 and Android 7.0)

![pagerefresh](https://user-images.githubusercontent.com/31961530/32235034-0a8ac0fa-be67-11e7-81dc-4d6d1c640bd8.gif)

PR #3698, opened 2017-10-30
Reload review list after submitting review

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3664

Screencast after the fix:

addons-frontend-fix-reviews mov


src/amo/components/AddonReviewList/index.js
```diff
@@ -265,10 +290,11 @@ export class AddonReviewListBase extends React.Component<Props> {
   }
 }
 
-export function mapStateToProps(
-  state: {| user: UserStateType, reviews: ReviewState |},
-  ownProps: Props,
-) {
+type AppState = {|
```
willdurand: I would not have expected strict type here because the `AppState` looks incomplete to me.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -174,19 +174,42 @@ describe('RatingManager', () => {
...
+    return root.onSelectRating(5)
+      .then(() => {
+        sinon.assert.called(FakeAddonReview);
+        const props = FakeAddonReview.firstCall.args[0];
+
+        // Simulate pressing submit in the review UI.
+        props.onReviewSubmitted();
+        sinon.assert.called(parentOnReviewSubmitted);
+      });
+  });
```
willdurand: It took me a while to understand the aim of this test case 😕

src/amo/components/AddonReviewList/index.js
kumar303: It helps to make it strict because it will catch any mistakes within the mapper function. However, we should really define a centralized `AppState` type and reference that in every mapper function.

tests/unit/amo/components/TestRatingManager.js
kumar303: I added more comments. This test suite uses some old patterns but I didn't want to spend too much time modernizing it.

willdurand: thanks!

PR #3699, opened 2017-10-30
Expose support email

### LINKED ISSUE #3696: Show support e-mail address in More Information box
Developers can provide a Support URL, a Support E-mail address, both, or neither. The new front-end currently doesn't support the e-mail address.

willdurand:
Fix #3696 

---

screen shot 2017-10-30 at 18 21 26

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -54,9 +54,26 @@ export class AddonMoreInfoBase extends React.Component<Props> {
       );
     }
 
+    let supportEmail = addon.support_email;
+    if (supportEmail && /.+@.+/.test(supportEmail)) {
+      supportEmail = (
+        <li>
+          <a
+            className="AddonMoreInfo-support-email"
+            href={`mailto:${supportEmail}`}
+          >
+            {i18n.gettext('Support Email')}
+          </a>
+        </li>
+      );
+    } else {
+      supportEmail = null;
+    }
+
     return this.renderDefinitions({
       homepage,
       supportUrl,
+      supportEmail,
       statsLink: addon && (isAddonAuthor({ addon, userId }) || addon.public_stats) ? (
         <Link
           className="AddonMoreInfo-stats-link"
@@ -125,6 +142,7 @@ export class AddonMoreInfoBase extends React.Component<Props> {
   renderDefinitions({
     homepage = null,
     supportUrl = null,
+    supportEmail = null,
     statsLink = null,
     privacyPolicyLink = null,
     eulaLink = null,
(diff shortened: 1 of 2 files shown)


muffinresearch:
Before we do this there appears to be some crude antispam measures in place on the current site. We probably should consider if implementing something similar is needed here.

Presumably the url needs to be url encoded?

willdurand:
@muffinresearch I checked the old website and did not find anything special, so I kept things simple. I knew the discussion would happen so I am open to suggestions.

muffinresearch:
> @muffinresearch I checked the old website and did not find anything special, so I kept things simple. I knew the discussion would happen so I am open to suggestions.

Looks like the old site displays in reverse and then in JS inverts it back again - not sure how reliable that is since having a js aware email harvester would be pretty trivial these days.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -137,16 +155,17 @@ export class AddonMoreInfoBase extends React.Component<Props> {
...
     return (
       <dl className="AddonMoreInfo-contents">
-        {homepage || supportUrl ? (
+        {homepage || supportUrl || supportEmail ? (
           <dt className="AddonMoreInfo-links-title">
             {i18n.gettext('Add-on Links')}
           </dt>
         ) : null}
-        {homepage || supportUrl ? (
+        {homepage || supportUrl || supportEmail ? (
```
kumar303: We should add a helper to deal with duplicated if statements. This should work:

```js
const renderNodesIf = (includeContent: boolean, nodes: Array<any>) => {
  return nodes.map((content) => includeContent ? content : null);
}
```

```js
        {renderNodesIf(homepage || supportUrl, [
          <dt className="AddonMoreInfo-links-title">
            {i18n.gettext('Add-on Links')}
          </dt>,
          <dd className="AddonMoreInfo-links-contents">
            <ul className="AddonMoreInfo-links-contents-list">
              {homepage}
              {supportUrl}
            </ul>
          </dd>,
        ])}
```

willdurand:
Ah indeed, I did not pay attention to the HTML sent by the server. That's the problem nowadays, most obfuscations are inefficient. This article might be of interest: http://www.grall.name/posts/1/antiSpam-emailAddressObfuscation.html.

muffinresearch:
> I don't think porting over the JS obfuscation from the old site is worthit.

I agree that we shouldn't just re-implement what we have on the old site.

@jvillalobos do you know if there's anything we should be aware re: privacy around the support emails being provided by developers? As noted above the old site provides something that probably isn't preventing harvesting. Are there any considerations today or is just exposing these as pure mailto links acceptable?

muffinresearch:
Looking through the privacy policy and distribution agreements I don't see a mention of how support email information is handled.

willdurand:
Ok, adding the "do not merge" label until we know if exposing the support email like this is ok.

muffinresearch:
> Ok, adding the "do not merge" label until we know if exposing the support email like this is ok.

Thanks, I should be able to get an answer on this today.

muffinresearch:
@willdurand I checked this at the leadership call and as a js enabled browser shows a default mailto: link on the legacy AMO we're good to land this.

jvillalobos:
> @jvillalobos do you know if there's anything we should be aware re: privacy around the support emails being provided by developers?

Developers should be aware that exposing the email address in a public site will lead to spamming and other issues, so I think it's okay not to port the old code.

PR #3701, opened 2017-10-30
Always link to add-on stats when they are declared public

kumar303:
Fixes: #3603 

This patch was started by @tsl143 - thanks!

**Before:**

While logged out, viewing an add-on that has public stats looked like:

screenshot 2017-10-30 16 41 18

**After:**

screenshot 2017-10-30 16 41 36


kumar303:
For code reviews see https://github.com/mozilla/addons-frontend/pull/3679

Issue #2261, opened 2017-04-10
Developer or admin disabled listings needs updating


jvillalobos:
Visit the Fire.fm listing page on [mobile](https://addons.mozilla.org/android/addon/firefm/) and [desktop](https://addons.mozilla.org/firefox/addon/firefm/) to see the difference. I used the URL to access the page, I don't think it's possible to navigate to it.

Expected: a message indicating the add-on has been disabled by its developers. It'd be good to not show any add-on metadata (icon, developers), since that has been a problem in the past.

jvillalobos:
Pages disabled by an admin, [like this one](https://addons.mozilla.org/android/addon/microsoft-net-framework-assist/), show a "Not Authorized" error. I think, similarly, we should be showing a notice that indicates the add-on was disabled by Mozilla.

kumar303:
What I see on this URL is a Not Authorized error. I don't see any metadata. In this issue are you requesting that we just add a more accurate message about why it was disabled and make sure to hide the metadata?

<img width="715" alt="screenshot 2017-04-11 12 04 44" src="https://cloud.githubusercontent.com/assets/55398/24921048/2858d9a8-1eaf-11e7-98c7-d14e76c52d47.png">

jvillalobos:
Ah, I suppose both cases are being treated the same: Not Authorized if you're not the dev, normal add-on listing if you're the dev (I'm a developer of the add-on in the first example).

Given that, I'm requesting that:
* The page appears the same regardless you being the developer or not.
* Instead of "Not Authorized" explain the add-on was disabled by the developer or Mozilla, depending on the situation. The Suggested Pages section and below all makes sense.

muffinresearch:
Currently disabled addons look very odd:

<img width="1410" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30448738-178b84b0-9987-11e7-83bb-8a56539745d0.png">

So nominating this for MVP as a bug to be fixed.

muffinresearch:
@kumar303 do we actually have any data returned to special case these pages? If we don't we should at least get back to a generic error page that isn't trying to display any addon data, ahead of a more involved fix.

kumar303:
> do we actually have any data returned to special case these pages?

I don't see anything that we can currently use.

Since this is mostly a cosmetic issue for pages that are unlikely to be seen, I think we should defer it until after the MVP release.

kumar303:
[...]

PR #3703, opened 2017-10-31
Show disabled add-on error pages

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2261

**Before:**

screenshot 2017-10-30 20 10 56

**After** 

Developer disabled add-ons:

screenshot 2017-10-30 20 09 01

Admin disabled add-ons:

screenshot 2017-10-30 20 10 06

All other 404s:

screenshot 2017-10-31 09 57 57

### DIFF
--- a/src/amo/components/ErrorPage/NotFound/index.js
+++ b/src/amo/components/ErrorPage/NotFound/index.js
@@ -1,23 +1,27 @@
+/* @flow */
 import React from 'react';
-import PropTypes from 'prop-types';
 import { compose } from 'redux';
 import NestedStatus from 'react-nested-status';
 
 import SuggestedPages from 'amo/components/SuggestedPages';
+import {
+  ERROR_ADDON_DISABLED_BY_ADMIN, ERROR_ADDON_DISABLED_BY_DEV,
+} from 'core/constants';
 import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
 import Card from 'ui/components/Card';
+import type { I18nType } from 'core/types/i18n';
 
 import 'amo/components/ErrorPage/ErrorPage.scss';
 
+type Props = {|
+  errorCode?: string,
+  i18n: I18nType,
+|};
 
-export class NotFoundBase extends React.Component {
-  static propTypes = {
-    i18n: PropTypes.object.isRequired,
-  }
-
+export class NotFoundBase extends React.Component<Props> {
   render() {
-    const { i18n } = this.props;
+    const { errorCode, i18n } = this.props;
 
     const fileAnIssueText = i18n.sprintf(i18n.gettext(`
       If you followed a link from somewhere, please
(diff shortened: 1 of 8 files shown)


src/amo/components/ErrorPage/NotFound/index.js
```diff
@@ -1,30 +1,47 @@
...
       what you were looking for, and we'll do our best to fix it.`),
     { url: 'https://github.com/mozilla/addons-frontend/issues/new/' });
 
+    let explanation;
+    if (errorCode === ERROR_ADDON_DISABLED_BY_DEV) {
+      explanation =
+        i18n.gettext('This add-on has been removed by its author.');
+    } else if (errorCode === ERROR_ADDON_DISABLED_BY_ADMIN) {
+      explanation =
+        i18n.gettext('This add-on has been disabled by an administrator.');
```
tofumatt: Should we just say "disabled by Mozilla"? I think that makes more sense and is a bit clearer.

src/amo/components/ErrorPage/NotFound/index.js
```diff
@@ -1,30 +1,47 @@
...
+    } else {
+      explanation = i18n.gettext(`
+        Sorry, but we can't find anything at the address you entered.
+        If you followed a link to an add-on, it's possible that add-on
+        has been removed by its author.`);
```
tofumatt: This is no longer true, right? Is there even an instance where we don't get the "removed by author" code if this is the case?

src/core/reducers/errors.js
```diff
@@ -11,19 +17,20 @@ import log from 'core/logger';
...
   log.info('Extracting messages from error object:', error);
 
+  const logCodeChange = ({ oldCode, newCode }) => {
+    log.warn(`Replacing error code ${oldCode} with ${newCode}`);
+  };
+
   if (error && error.response && error.response.data) {
+    // Extract a code and messages from the JSON response.
     Object.keys(error.response.data).forEach((key) => {
       const val = error.response.data[key];
```
tofumatt: I know this isn't your fault, but can we change `val` to something more meaningful? At the very least, `value` 😄

src/amo/components/ErrorPage/NotFound/index.js
kumar303: I got the copy from the old site. I was thinking about it and I like how administrator is a bit more AMO-focused and inclusive, meaning we have volunteers who may be helping to disable things (well, I know we have volunteer reviewers but I'm not sure about admins).

src/amo/components/ErrorPage/NotFound/index.js
kumar303: Oh yeah, that's true. I can shorten it.

src/core/reducers/errors.js
kumar303: Sure

PR #3725, opened 2017-10-31
feat: Show a better proxy error during development (close #3721)

### LINKED ISSUE #3721: Have Proxy error in development give more info and refresh page
Sometimes during development I get too excited and load `localhost:3000` before webpack has built everything and the server is ready. The "Proxy error" we get isn't super helpful and might confuse a new developer.

Also, it'd be nice if that error page tried to refresh the page (a meta-refresh would work fine, as this stuff is only seen in development) so I didn't have to manually refresh it lots. That way once the server was ready the page would just reload and I wouldn't have to sit there babysitting it 😄

tofumatt:
Doing this and making it look pretty, which I think was totally worth it, involved just cramming a bunch of stuff inline in the file. I know it'll get outdated, it's dumb, etc. but I think it's nice and it's loads better than what we had before. Plus: the main content can easily be extracted and pasted into a new, similar file. Maybe we can make this better in the future. For now I still like it and find it way more convenient.

Fixes #3721; this makes the proxy error clearer, automatically refreshes the page every 10 seconds so eventually the development site will just load automatically in the background, and gives the page some style points.

If you went to `localhost:3000` before webpack was built in the past you saw:

### Before

screenshot 2017-10-31 19 58 54

And now you see:

### After
screenshot 2017-10-31 19 58 14


bin/loading-page.html
```diff
@@ -0,0 +1,108 @@
+<!DOCTYPE html>
```
tofumatt: Seriously though: this file is hideous and you can just ignore it 😉

Issue #3728, opened 2017-10-31
Redirect to seamonkey and thunderbird apps via user agent


kumar303:
### Describe the problem and steps to reproduce it:

* Set your user agent to that of seamonkey or thunderbird
* Load a page without an app, such as `https://addons-dev.allizom.org/en-US/`

### What happened?

You are redirected to `https://addons-dev.allizom.org/en-US/firefox/`

### What did you expect to happen?

* For a seamonkey user agent you should be redirected to `https://addons-dev.allizom.org/en-US/seamonkey/`
* For a thunderbird user agent you should be redirected to `https://addons-dev.allizom.org/en-US/thunderbird/`

### Anything else we should know?

* seamonkey and thunderbird URLs are still served from the legacy site: https://github.com/mozilla/addons-frontend/issues/3719
* As of https://github.com/mozilla/addons-frontend/issues/3678 the `addons-frontend` server will need to be in charge of redirects

muffinresearch:
Duplicate of https://github.com/mozilla/addons-frontend/issues/3719

muffinresearch:
@kumar303 not sure if this is at cross purposes with #3719, as that issue was about making sure SeaMonkey and Thunderbird UAs only ever see the legacy site not just based on paths - that was already in place.

Let me know if I've misunderstood anything.

kumar303:
@muffinresearch I'm confused. The STR up above on dev lands me https://addons-dev.allizom.org/en-US/firefox/ but it should land me at https://addons-dev.allizom.org/en-US/thunderbird/ (or seamonkey). With my patch it will do so.

muffinresearch:
We talked about this on irc, the confusion was caused by kumar's UA extension not setting the test UA  correctly.

kumar303:
Yeah, I was experiencing some kind of bug where the UA was not getting applied to the request. 

The clarification I was missing is that nginx *will* inspect the user agent. It always serves the legacy site to seamonkey/thunderbird UAs and the legacy site does redirects as expected.

PR #3729, opened 2017-10-31
Redirect to seamonkey/thunderbird based on user agent

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3728

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -222,6 +222,8 @@ module.exports = {
   validClientApplications: [
     'android',
     'firefox',
+    'seamonkey',
+    'thunderbird',
   ],
 
   validLocaleUrlExceptions: [],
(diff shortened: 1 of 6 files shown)


kumar303:
@muffinresearch should we cherry pick this?

config/default.js
```diff
@@ -222,6 +222,8 @@ module.exports = {
   validClientApplications: [
     'android',
     'firefox',
+    'seamonkey',
```
tofumatt: Will this allow the frontend to navigate to seamonkey and thunderbird URLs? Do we want that?

kumar303:
It sounds like you answered the questions but here are the main points:

* After this patch you can browse the site with either of these client apps in the URL and nothing blows up. You can only browse the site like this on your local dev server though.
* On dev, stage, and prod the nginx rules will serve these client app URLs from `addons-server`. However, without this patch it won't redirect correctly for the user agents.

muffinresearch:
We shouldn't need this patch see my comments in #3728

PR #3741, opened 2017-11-01
Display stats links to admins or users with 'Stats:View' permission

### LINKED ISSUE #3604: Admins should be able to always see a stats link

* Log in as an admin or any user with the `Stats:View` permission
* Go to an add-on detail page for an add-on that you are not an owner of

### What happened?

There is no link to add-on statistics

You should see a link to add-on statistics

This is similar to https://github.com/mozilla/addons-frontend/issues/3603

willdurand:
Fix #3604

---

This PR relies on the newly added `permissions` state (in `user` reducer) to display the stats link to any users with the `Stats:View` permission or admins.

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -7,7 +7,7 @@ import Link from 'amo/components/Link';
 import ReportAbuseButton from 'amo/components/ReportAbuseButton';
 import { STATS_VIEW } from 'core/constants';
 import translate from 'core/i18n/translate';
-import { hasPermission, isAdmin } from 'core/reducers/user';
+import { hasPermission } from 'core/reducers/user';
 import type { AddonType } from 'core/types/addons';
 import {
   addonHasVersionHistory,
@@ -26,7 +26,7 @@ type Props = {|
   addon: AddonType | null,
   i18n: I18nType,
   userId: number | null,
-  canViewStats: boolean,
+  hasStatsPermission: boolean,
 |};
 
 const renderNodesIf = (includeContent: boolean, nodes: Array<any>) => {
@@ -35,7 +35,7 @@ const renderNodesIf = (includeContent: boolean, nodes: Array<any>) => {
 
 export class AddonMoreInfoBase extends React.Component<Props> {
   listContent() {
-    const { addon, i18n, userId, canViewStats } = this.props;
+    const { addon, i18n, userId, hasStatsPermission } = this.props;
 
     if (!addon) {
       return this.renderDefinitions({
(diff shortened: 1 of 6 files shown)


src/amo/components/AddonMoreInfo/index.js
```diff
@@ -247,9 +256,10 @@ export class AddonMoreInfoBase extends React.Component<Props> {
   }
 }
 
-export const mapStateToProps = (state: Object) => {
+export const mapStateToProps = (state: {| user: UserStateType |}) => {
   return {
     userId: state.user.id,
+    canViewStats: isAdmin(state) || hasPermission(state, STATS_VIEW),
```
kumar303: Since the concept of `isAddonAuthor()` is checked separately, it might be a useful clarification to name this property more explicitly like `hasStatsPermission`.

src/core/reducers/user.js
```diff
@@ -45,6 +45,21 @@ export const selectDisplayName = (state: { user: UserStateType }) => {
...
 
+export const hasPermission = (
+  state: { user: UserStateType }, permission: string,
+): boolean => {
+  const permissions = state.user.permissions;
+  if (!permissions) {
+    return false;
+  }
+
+  return permissions.includes(permission);
```
kumar303: It's going to be easier to scale this up if you include the admin check as an implementation detail:

```js
return permissions.includes(permission) || isAdmin(state);
```

Admins can do anything.

willdurand: ah that makes sense indeed!

src/amo/components/AddonMoreInfo/index.js
willdurand: That's what I had before and then changed because of `isAdmin()`.

src/core/reducers/user.js
willdurand: I removed the `isAdmin` for now as we're not using it.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -315,6 +317,34 @@ describe(__filename, () => {
...
+    const root = render({
+      addon,
+      store: dispatchSignInActions({ permissions: [STATS_VIEW] }).store,
+    });
+
+    const statsLink = root.find('.AddonMoreInfo-stats-link');
+    expect(statsLink).toHaveLength(1);
+  });
+
+  it('links to stats if user is ADMIN', () => {
```
kumar303: This test should be deleted. It would be testing an implementation detail of `hasPermission()` which is not necessary for this unit test. More importantly, the assertion is a duplication of https://github.com/mozilla/addons-frontend/pull/3741/files#diff-2e12f8c050405872bcaf7b56035e9387R157

willdurand:
Thanks for the review, should be ok now.

Issue #3747, opened 2017-11-02
"Invalid 'type' parameter" red banner for "All" add-on type in each category


ValentinaPC:
Steps to reproduce:
1. Load any add-on category page i.e. https://addons.allizom.org/en-US/firefox/extensions/shopping/
2. Choose "see more" for any sorting option i.e. https://addons.allizom.org/en-US/firefox/search/?category=shopping&sort=hotness&type=extension
3. Change the add-on type to "All"

Expected results:
The search results are refined.

Actual results:
"Invalid 'type' parameter" red banner is displayed.

Notes/Issues:
Verified on AMO-stage FF56(Win 7, Android 7.2.1). 
Reproducing on all servers.

Not sure if is not related with https://github.com/mozilla/addons-frontend/issues/3563 .
Screenshot for this issue:
![2017-11-02_1203](https://user-images.githubusercontent.com/15685960/32320594-0384ba34-bfc7-11e7-93a6-daf012bd6ac0.png)

tofumatt:
😱 Oh my, that's not good.

muffinresearch:
A small step towards improving this would be to fix the error message to not be the size of a house.

The second step is probably to disable the type selector for category searches. 

Long term it would be really great to fix the categories to be the same across types which is a whole separate issue.

ValentinaPC:
Verified this on AMO-dev FF56(Win 7, Android 7.2.1).
Add-on type is no longer listed in any add-on category.
Postfix video:
![cat](https://user-images.githubusercontent.com/15685960/32367170-e56b9d7e-c089-11e7-9622-af1dbaf8a6f2.gif)

PR #3749, opened 2017-11-02
Fix search filters error style

willdurand:
Refs #3747

---

## Screenshots

screen shot 2017-11-02 at 12 11 08

screen shot 2017-11-02 at 12 10 46

screen shot 2017-11-02 at 12 10 58

screen shot 2017-11-02 at 12 11 03

### DIFF
--- a/src/amo/components/Search/styles.scss
+++ b/src/amo/components/Search/styles.scss
@@ -45,7 +45,7 @@
   .SearchResults {
     @include respond-to(large) {
       grid-column: 2;
-      grid-row: 2 / 4;
+      grid-row: 2 / 5;
       // Allows the grid item to contain the content preventing overflow.
       min-width: 0;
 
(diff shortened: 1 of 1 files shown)


src/amo/components/Search/styles.scss
```diff
@@ -45,7 +45,7 @@
   .SearchResults {
     @include respond-to(large) {
       grid-column: 2;
-      grid-row: 2 / 4;
+      grid-row: 2 / 5;
```
muffinresearch: Do we need to specify specific rows for the layout at all? I wonder if we could stop doing that where we can rely on implicit ordering.

willdurand: Not sure we can because of the difference between mobile and desktop pages.

muffinresearch: We can look at that separately, I might be wrong (I'm still getting used to grids) but I think aside from intricate layouts specific rows might be something that could be factored out.

PR #3750, opened 2017-11-02
Disably addonType search filter when category type is set

### LINKED ISSUE #3747: "Invalid 'type' parameter" red banner for "All" add-on type in each category

willdurand:
Fix #3747

---

This PR does not display the add-on type select filter when a category filter is set.

## Screenshot

screen shot 2017-11-02 at 14 32 45

### DIFF
--- a/tests/unit/amo/components/TestSearchFilters.js
+++ b/tests/unit/amo/components/TestSearchFilters.js
@@ -246,6 +246,7 @@ describe(__filename, () => {
   });
 
   it('does not display the addonType filter when a category is defined', () => {
+    // See: https://github.com/mozilla/addons-frontend/issues/3747
     const root = render({ filters: { category: 'abstract' } });
 
     expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
--- a/tests/unit/amo/components/TestSearchFilters.js
+++ b/tests/unit/amo/components/TestSearchFilters.js
@@ -244,4 +244,10 @@ describe(__filename, () => {
       }),
     });
   });
+
+  it('does not display the addonType filter when a category is defined', () => {
+    const root = render({ filters: { category: 'abstract' } });
+
+    expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
+  });
 });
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +244,10 @@ describe(__filename, () => {
       }),
     });
   });
+
+  it('does not display the addonType filter when a category is defined', () => {
```
tofumatt: Referencing the issue here would be good, as this is unexpected behaviour (to me) and I think warrants explanation as to why we do it. Just a link should be good 😄

muffinresearch:
Would it be any better to show it but have it be disabled if we could know in advance that category doesn't exist for that type? It could show a message to explain why underneath as help text.

This would then allow it to be enabled if a category is common between types.

muffinresearch:
We might want to get some UX and product input on this in terms of what the best long term solution should be.

tofumatt:
But category *names* that are common between types would still have a different ID. They’re tied to clientApp and addonType. Nothing should ever really be shared. And I don’t think any are.

- Matt (Sent from mobile)

muffinresearch:
> But category *names* that are common between types would still have a different ID. They’re tied to clientApp and addonType. Nothing should ever really be shared. And I don’t think any are.

Ah ok, not sure why I thought there was overlap. In which case I think this solution makes sense.

Issue #3764, opened 2017-11-02
Not all search filters are included in paginated links


mstriemer:
### Describe the problem and steps to reproduce it:
1. Go to a search page with extra filters, like [`category=web-development&type=extension`](https://addons.mozilla.org/en-US/firefox/search/?category=web-development&type=extension).
2. Click **Next**.

### What happened?
You are now on page 2 of a search for `type=extension`.

### What did you expect to happen?
You are now on page 2 of a search for `category=web-development&type=extension`.

### Anything else we should know?
![search-link](https://user-images.githubusercontent.com/211578/32348621-58cf8954-bfe2-11e7-859f-9edeb8ba581a.jpg)

kumar303:
Nice Firefox theme :)

I could reproduce it on Mac so I tried on Windows in Nightly 58.0a1 2017-11-01. I still couldn't reproduce it. What version were you using?

kumar303:
Oh, sorry, ignore me. I *can* reproduce it on Mac and Windows. I misread the issue.

mstriemer:
Screenshot is from https://twitter.com/prashantak/status/926181742620364800. You'll have to ask him about the theme if you're interested ;)

PR #3765, opened 2017-11-02
Fix pagination in category search

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3764


src/amo/components/SearchPage/index.js
```diff
@@ -18,15 +18,20 @@ type Props = {|
...
+    // Explicitly map query string values from filters.
+    ...convertFiltersToQueryParams({
+      addonType: filters.addonType,
+      featured: filters.featured,
+      operatingSystem: filters.operatingSystem,
+      page: filters.page,
+      query: filters.query,
+      sort: filters.sort,
+      tag: filters.tag,
+    }),
```
willdurand: Why not adding a mapping for `filters.category` here?

kumar303:
Thanks for the review. I didn't fully understand what was happening here but on a second look I see it. 

I actually tried to flow type `searchUtils` which would have helped me understand what's happening. I didn't complete it though because Flow has a lot of bugs around reducing arrays and merging objects and I didn't want to rewrite that code. 

The test I added doesn't cover all filter parameters but doing so seems like it would be excessive. It would be better just to Flow type it.

Issue #3763, opened 2017-11-02
More prominent Firefox branding in the Nav


pwalm:
**Very high priority**: We need to change the puzzle piece to the Firefox icon and adjust our typography to match the Firefox branding guidelines. This needs to be done and launched before the 57 launch.

A good example from Test Pilot:
<img width="1044" alt="screen shot 2017-11-02 at 12 56 38 pm" src="https://user-images.githubusercontent.com/6797635/32347259-4a950a4a-bfcd-11e7-983c-7957dd3daf65.png">

Here is a mock up:
<img width="1253" alt="screen shot 2017-11-02 at 12 56 14 pm" src="https://user-images.githubusercontent.com/6797635/32347271-536b73b6-bfcd-11e7-8f9a-3d4d930007fb.png">

Assets and guidelines [can be found on the Design System](http://design.firefox.com/photon/visuals/product-identity-assets.html).

Spec:
<img width="961" alt="screen shot 2017-11-02 at 1 00 56 pm" src="https://user-images.githubusercontent.com/6797635/32347446-e898e644-bfcd-11e7-9da9-23b4c107eaab.png">

kumar303:
@pwalm I have a few questions:

- The mock you posted has a bit of a pink tint to it that I don't see in the assets. Should I use the official svg asset instead?
- The 36px left margin you spec'd differs from our current 24px margin. Do you really want to change that to 36px? If so, it will look odd if we don't change all margins to 36px, I think.
- Could you give some context as to why we can't use the puzzle icon? I thought it added a unique personality to the site.

pwalm:
- Use the official svg icon [from here](http://design.firefox.com/photon/visuals/product-identity-assets.html).
- Stick with 24px! Sorry, thought we were using 36px. 
- I thought so too, but they want to keep Firefox sites uniform and branded with the Firefox icon. We can still keep the favicon though. ;)

muffinresearch:
Cross-posting this from the PR to keep track of the changes. The following is combined wordmark svg based on @pwalm's spec - this solves one of the issues in that we can use this across all locales and just localize the alternate text to keep visual consistency. (See PR for the additional comments).

Example screenshot in InkScape:
<img width="1124" alt="firefox-addons-logo-horizontal-lockup-1_svg_-_inkscape" src="https://user-images.githubusercontent.com/1514/32387241-bc8c966a-c0bb-11e7-8652-d2c6f7b2468a.png">

Svg in zip file:
[...]

PR #3767, opened 2017-11-02
Replace puzzle icon with Firefox in the header

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3763

Screenshots:

**XL**:

screenshot 2017-11-03 14 37 42

**L**:

screenshot 2017-11-03 14 38 35

**M**:

screenshot 2017-11-03 14 39 01

**S**:

screenshot 2017-11-03 14 39 29

Right to left example:

screenshot 2017-11-03 14 40 26


willdurand:
I noticed this in the mockup from the issue but it is even more visible on your screenshots 1, 3 and 4. Everything looks misaligned 😕

The menu is floating without really being aligned vertically. Maybe we should align it with the search bar (cf. Screenshots 1 and 3), what do you think?

muffinresearch:
@kumar303 I think this will require some tweaking and maybe some refactoring of the header to make it work a bit better since the bigger icon does seem to change everything as Will pointed out. 

Re: the wordmark idea, currently the heading string is being translated. I wonder if "Firefox Add-ons" is seen more as branding then would we still need the translations for it? Test Pilot doesn't appear to have that issue since it doesn't look like they have localisations. 

If it was appropriate to use a wordmark (including "Add-ons"), that might simplify some things and we wouldn't have to make so many considerations in the layout for long translations which is why things work the way they do currently. Of course RTL would also need consideration in any case.

muffinresearch:
I spoke to @jvillalobos about this today and we concluded that we could use an SVG wordmark for the entire logo e.g. `[FFLOGO] Firefox Add-ons` this can then have localized alternate text but keeps a visual consistency for the whole site. 

This should mean the grid can be simplified and it should make the whole thing a bit easier from a responsive perspective.

I'll see about getting an asset to us use for this ASAP.

muffinresearch:
Here's a combined wordmark svg based on @pwalm's spec:

Example screenshot in InkScape:
firefox-addons-logo-horizontal-lockup-1_svg_-_inkscape

Svg in zip file:
[firefox-addons-min.svg.zip](https://github.com/mozilla/addons-frontend/files/1441909/firefox-addons-min.svg.zip)

kumar303:
The wordmark looks way better. I updated the screenshots -- this is ready for another look. 

Some notes:
* I sized up the logo to 48px in the large/medium/small screens instead of 32px like the puzzle icon. I think the text looks better like this because the new logo has a bigger size in proportion to the text.
* I noticed that the distance between menu items is 24px not 36px like the Sketch spec for the header said but I kept it as is
* @willdurand I'm not sure if I fully addressed your comments about aligning things to the search bar. I tried to follow the current site's layout for now instead of improve on it -- maybe we need a follow-up issue?

kumar303:
The build will be fixed after https://github.com/mozilla/addons-frontend/pull/3777

willdurand:
> @willdurand I'm not sure if I fully addressed your comments about aligning things to the search bar. I tried to follow the current site's layout for now instead of improve on it -- maybe we need a follow-up issue?

Yes, we should. It was not really aligned before but the new bigger icon makes it obvious that it needs some love.

I noticed one issue on "focus", I think the color should be `$white`:

screen shot 2017-11-03 at 21 51 13

kumar303:
> Yes, we should. It was not really aligned before but the new bigger icon makes it obvious that it needs some love.

Are you talking about how the menu doesn't line up horizontally with the search bar in the large and medium screenshots up above?

> I noticed one issue on "focus", I think the color should be $white

Oops, I removed that because I didn't think we needed it. I put it back :)

willdurand:
> Are you talking about how the menu doesn't line up horizontally with the search bar in the large and medium screenshots up above?

yes!

PR #3772, opened 2017-11-03
Add data-no-csrf tag to search form

### LINKED ISSUE #3770: Search form missing data-no-csrf attribute
The Search form included on most (all?) pages doesnt have an anti-csrf token (which is fine) and also doesnt have a 'data-no-csrf' attribute (which is not).
This means that the addons.mozilla.org and addons.allizom.org are now failing the baseline tests.

For more background see https://bugzilla.mozilla.org/show_bug.cgi?id=1328560

Is here a good place to add it? https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/SearchForm/index.js#L242-L251
If so I'm happy to submit a PR...

psiinon:
Fixes #3770

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -248,6 +248,7 @@ export class SearchFormBase extends React.Component {
         method="GET"
         onSubmit={this.handleSearch}
         ref={(ref) => { this.form = ref; }}
+        data-no-csrf
       >
         <label
           className="visually-hidden"


src/amo/components/SearchForm/index.js
```diff
@@ -247,7 +247,8 @@ export class SearchFormBase extends React.Component {
         })}
         method="GET"
         onSubmit={this.handleSearch}
-        ref={(ref) => { this.form = ref; }}
+        ref={(ref) => { this.form = ref; }
+        data-no-csrf}
```
willdurand: It looks like there is an issue here. It should be:

```
ref={(ref) => { this.form = ref; }}
data-no-csrf
```

Having a unit test for it would be nice too (an assertion in an existing test case would do the trick)

psiinon: Ha, was wondering about that. Will fix that.
So I've not written / updated any tests on addons before - which one would be a good one to base mine on?

psiinon: Syntax fixed

psiinon:
Think thats fixed now?

tofumatt:
@psiinon Should all form tags we use have a data tag like this? We could probably add a test of our own to check for the attribute, just to make sure.

willdurand:
POST forms should have csrf tokens no?

tofumatt:
Errr yeah I meant `GET` ones. Though actually I'm not sure how our app handles CSRF given every request is sent to the API. I was not very involved in the auth implementation, maybe @kumar303 or @mstriemer know better 😓

diox:
POST forms in addons-frontend should not require a CSRF token since the auth does not depend on cookies automatically sent by the browser, so there is no way for an attacker to force somebody to make a POST request while authenticated (without an XSS)

Only exception would be forms that don't require auth. Are there any?

Issue #3773, opened 2017-11-03
Page level error handling


willdurand:
This issue is derived from #3313. Below is an extract of the discussion:

---

For example if a details page call doesn't find an addon that page should render the NotFound component and if directly rendered by the server it should be a 404 (or maybe 410 gone if explicitly removed).

Similarly a search page request that asks for a page or results that don't exist, should similarly be 404 when directly linked and should also render the NotFound component.

---

The idea is to create a new HOC to deal with page level errors that can be sync'ed between the client and server sides.

willdurand:
This is actually implemented in #3775.

Issue #3387, opened 2017-10-06
Create a separate component for fetching add-ons by author for better errors


kumar303:
The Addon component shows add-ons by other authors as of https://github.com/mozilla/addons-frontend/pull/3083

We should move this to a separate component so that it can have its own error handler. Otherwise, an error might looks like this (example triggered by https://github.com/mozilla/addons-server/issues/6610). This screenshot would have made a confusing bug report.

<img width="799" alt="screenshot 2017-10-06 11 14 44" src="https://user-images.githubusercontent.com/55398/31289016-f100defa-aa8c-11e7-9d8e-280af17cd546.png">

willdurand:
This has been done.

Issue #3313, opened 2017-10-02
Pages stuck in "loading" mode when an error is thrown


willdurand:
When we use a `loading` flag in a reducer and the API layer throws an error (e.g. 400 or 404), then the redux state keeps `loading` set to `true`, which results in a page stuck in "loading" mode sent by the server.

For instance, here is a search result that throws a 400: https://addons.allizom.org/en-US/firefox/search/?platform=pouet&q=ad&type=extension. We do not get any error displayed and the page does not look nice. In #3268, I implemented a similar mechanism that should render a `NotFound` component when a 404 error is caught. Yet this does not work either.

Note: This does not seem to be the case when SSR is disabled.

kumar303:
For handling unexpected errors (likely a 500), see https://github.com/mozilla/addons-frontend/issues/2905

muffinresearch:
Here's an example of what it looks like for https://addons.mozilla.org/en-US/firefox/search/?platform=bogus


<img width="1452" alt="add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/31126272-1df6b1d8-a843-11e7-9eec-ed11ee32b0fd.png">**

muffinresearch:
If this is linked to directly then it's something we should fix pre-release. If it's an edge case we might be able to de-prioritise it for post-release.

kumar303:
Why is this marked as MVP? Would a user ever see this broken loading state by using the site in a normal fashion?

tofumatt:
It's rendering the wrong page (eg not an error page) on the server, which seems bad. I just went to a 404 add-on and it seemed to send it down in a loading state then change to 404 once the client render happened.

kumar303:
Yes but you only encounter a 404 through atypical usage. I agree we should fix it but I don't see a case for marking this as an MVP bug.

willdurand:
I have downgraded this issue but marked it as "priority: post release" because I think we should care about errors (a lot) and I think it also impacts our productivity (I spent some time fighting with the error handler).

willdurand:
One more thing (well many more actually):

- the error handler does not generate the same IDs between the client and the server, resulting is a different state. This is not nice because it means we are not consistent between the server and the client. Most of our error handling code (like rendering an error) is framed upon the state. Yet, because the client and server states are different, the app behaves in a weird manner.

[...]

PR #3775, opened 2017-11-03
Introduce page level error handler

### LINKED ISSUE #3542: Out of range search results pages should be a 404
https://addons.mozilla.org/en-US/firefox/search/?page=30000&platform=mac&q=search

Currently if you attempt to load a page with an out of range page you get a loading page:

add-ons_for_firefox

### LINKED ISSUE #3548: 404 page is loaded for an existing collection
Steps to reproduce:
1. Load a collection page that returns a 404 i.e. https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/ad-blocker/
2. Go back to homepage
3. Go to an existing collection page i.e https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/change-up-your-tabs/

Expected results:
Collection page si loaded.

Actual results:
A 404 page is loaded.

Notes/Issues:
- reproducing only in case of collections vs. collections.
- issue fixed at page refresh

Verified on AMO-dev FF56(Win 7). Reproducing on all AMO servers.
Video for this issue:

willdurand:
Refs #3773 #3313 

Fix #3548
Fix #3542

---

**EDIT:** The text below is the result of my first attempt to solve the issues mentioned in #3313. I kept the different sections below as it gives the context of the problem(s) and also some hints about how I came up with this patch.

## The Search Page

The search page allows to render a "Not Found" page when a 401, 403 or 404 is returned by the API. For instance, out of range pagination now returns a "Not Found" page, whereas an invalid `type` value (e.g. `bom` instead of `persona`) results in a empty search results + error, all rendered by the server:

screen shot 2017-11-03 at 16 25 32

This works by dispatching an action targeting the `search` reducer when an error is caught. It avoids to get stuck in a "loading state". So the search page works as expected, without extra action dispatched, etc. And I don't see how it could end up in a deadlock situation.

**EDIT:** the status code is not reflected yet, but the page never gets stuck in loading mode anymore.

## The Collection Page

There was a known issue implying a page refresh when user navigates to a non-existent collection page, and then tries to navigate to a different (existing) collection page. The problem was caused by the error handler ID being fixed.

This PR fixes it but not exactly how it should. The server successfully renders the error (as expected) in the HTML (which is an improvement of the `master` code). However, I introduced the `componentDidMount()` method in the `Collection` component. This method is only called on the client, **not on the server** and clears an error in the error handler has one. Why?

The error handler is the same for all collections. That's why there was an issue when navigating to a non-existent collection. The error handler captures an error, and then it is over. No way to fetch new data because the error handler **has an error** already.

When the server rendering encounters an error, it also uses the error handler and captures an error, allowing the component to render the error.

screen shot 2017-11-03 at 17 17 17

That's cool to get the correct HTML from the server, but then the client cannot fetch any new collection. By using the `componentDidMount()`, I am sure to be in a "client" context and I can clear the error for future usage, enabling new collections to be fetched and keeping the error handling for potential non-existent collections (this should not happen in theory, but it is covered here).

Unfortunately, `componentDidMount()` triggers `componentWillReceiveProps()` and all the actions are played again on the server, but it ends in the same state! I wish I could get rid of this because it should not be done but I don't know how to do that (yet).

**EDIT:** the code is simpler now, with one error handler per collection. No need to reset any previous error or to share a common error handler.

## The Add-on Detail Page

It is very similar to the previous section, except that there is no need for `componentDidMount()` because the error handler is not used to prevent data fetching. Similar issue: the client side re-runs the actions so: 1. we get the error page from the server, 2. the client tries to load data, 3. the client gets an error and display the error page. Not cool.

Note that the "add-ons by other authors" should get its own component + component-level error handling. It is already mentioned in #3387.

**EDIT:** here too, there is one error handler per add-on, avoiding many problems including the one that executes all the actions on both the server and the client.

## And...

So I wish I could come up with something that says "on the client, when you receive a state containing page level errors, then reset them but do not re-render anything". I tried to dispatch an action on the server code to reset the `errors` state, but it did not work very well. Maybe someone will have a better idea.

I tried to give a unique ID to each page level error handler, based on the props received by the component (e.g. `params.slug`). This did not work well either, and it might be tricky to configure for all components (I did not try it but maybe passing a function that computes this unique ID given `ownProps` to the HOC would work). This *might* be a solution to avoid the problem mentioned in the previous paragraph, but I am not sure and there might be a better way to do that.

**EDIT:** the solution mentioned in the previous paragraph is actually the one I implemented in the last version of this patch.

All in all, I think having page level error handling is worthy (either based on this patch or something else, this is a POC really). I also think that we should update the state when an error occurs (see the `ABORT_*` actions in this patch) in order to avoid infinite "loading".

If we find a solution for these three different pages, then it should work for everything else (hopefully). Feel free to checkout this patch and play with it. Thanks!

---

## Updates 1

I just pushed a second commit with this idea I had while writing this PR description (the function in the HOC). It turns out that it works and resolves most of the pain points I explained previously. It is not all rainbows and unicorns yet, but it seems way better. I don't think updating the state on error helps with anything anymore but it still makes the state better, so I kept these changes too (as a POC, again).

## Updates 2

I updated the code with tests and fixed a few things to make things cleaner than before. It is not a POC or WIP anymore, so I'd like your input now :)

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -623,14 +623,13 @@ export function mapStateToProps(state, ownProps) {
   };
 }
 
+export const extractId = (ownProps) => {
+  return ownProps.params.slug;
+};
+
 export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withFixedErrorHandler({
-    name: __filename,
-    extractId: (ownProps) => {
-      return ownProps.params.slug;
-    },
-  }),
+  withFixedErrorHandler({ name: __filename, extractId }),
 )(AddonBase);
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -20,7 +20,7 @@ import Link from 'amo/components/Link';
 import { fetchOtherAddonsByAuthors } from 'amo/reducers/addonsByAuthors';
 import { fetchAddon } from 'core/reducers/addons';
 import { sendServerRedirect } from 'core/reducers/redirectTo';
-import { withPageErrorHandler } from 'core/errorHandler';
+import { withFixedErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
 import {
   ADDON_TYPE_DICT,
@@ -426,7 +426,12 @@ export class AddonBase extends React.Component {
     if (errorHandler.hasError()) {
       log.warn('Captured API Error:', errorHandler.capturedError);
 
-      if (errorHandler.shouldRenderNotFound()) {
+      // 401 and 403 are made to look like a 404 on purpose.
+      // See: https://github.com/mozilla/addons-frontend/issues/3061.
+      if (errorHandler.capturedError.responseStatusCode === 401 ||
+          errorHandler.capturedError.responseStatusCode === 403 ||
+          errorHandler.capturedError.responseStatusCode === 404
+      ) {
         return <NotFound errorCode={errorHandler.capturedError.code} />;
       }
 
@@ -622,8 +627,10 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withPageErrorHandler({
-    name: 'Addon',
-    extractId: (ownProps) => ownProps.params.slug,
+  withFixedErrorHandler({
+    name: __filename,
+    extractId: (ownProps) => {
+      return ownProps.params.slug;
+    },
   }),
 )(AddonBase);
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -33,7 +33,6 @@ import {
 } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import {
-  getRelativePath,
   nl2br,
   sanitizeHTML,
   sanitizeUserHTML,
@@ -632,5 +631,5 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withFixedErrorHandler({ name: getRelativePath(__filename), extractId }),
+  withFixedErrorHandler({ fileName: __filename, extractId }),
 )(AddonBase);
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -13,6 +13,7 @@ import AddonCompatibilityError from 'amo/components/AddonCompatibilityError';
 import AddonMeta from 'amo/components/AddonMeta';
 import AddonMoreInfo from 'amo/components/AddonMoreInfo';
 import ContributeCard from 'amo/components/ContributeCard';
+import NotFound from 'amo/components/ErrorPage/NotFound';
 import DefaultRatingManager from 'amo/components/RatingManager';
 import ScreenShots from 'amo/components/ScreenShots';
 import Link from 'amo/components/Link';
@@ -426,7 +427,7 @@ export class AddonBase extends React.Component {
       log.warn('Captured API Error:', errorHandler.capturedError);
 
       if (errorHandler.shouldRenderNotFound()) {
-        return errorHandler.renderNotFound();
+        return <NotFound errorCode={errorHandler.capturedError.code} />;
       }
 
       // Show a list of errors at the top of the add-on section.
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -93,30 +93,25 @@ export class AddonBase extends React.Component {
       params,
     } = this.props;
 
-    if (addon) {
-      if (!isNaN(params.slug)) {
-        // We only load add-ons by slug, but ID must be supported too because
-        // it is a legacy behavior.
-        dispatch(sendServerRedirect({
-          status: 301,
-          url: `/${lang}/${clientApp}/addon/${addon.slug}/`,
-        }));
-        return;
+    // This makes sure we do not try to dispatch any new actions in the case
+    // of an error.
+    if (!errorHandler.hasError()) {
+      if (addon) {
+        if (!isNaN(params.slug)) {
+          // We only load add-ons by slug, but ID must be supported too because
+          // it is a legacy behavior.
+          dispatch(sendServerRedirect({
+            status: 301,
+            url: `/${lang}/${clientApp}/addon/${addon.slug}/`,
+          }));
+          return;
+        }
+
+        dispatch(setViewContext(addon.type));
+        this.dispatchFetchOtherAddonsByAuthors({ addon });
+      } else {
+        dispatch(fetchAddon({ slug: params.slug, errorHandler }));
       }
-
-      dispatch(setViewContext(addon.type));
-      this.dispatchFetchOtherAddonsByAuthors({ addon });
-    } else {
-      dispatch(fetchAddon({ slug: params.slug, errorHandler }));
-    }
-  }
-
-  componentDidMount() {
-    const { errorHandler } = this.props;
-
-    if (errorHandler.hasError()) {
-      log.debug('Clearing errors on the client');
-      this.props.dispatch(errorHandler.createClearingAction());
     }
   }
 
@@ -626,5 +621,8 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withPageErrorHandler({ name: 'Addon' }),
+  withPageErrorHandler({
+    name: 'Addon',
+    extractId: (ownProps) => ownProps.params.slug,
+  }),
 )(AddonBase);
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -33,6 +33,7 @@ import {
 } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import {
+  getRelativePath,
   nl2br,
   sanitizeHTML,
   sanitizeUserHTML,
@@ -631,5 +632,5 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withFixedErrorHandler({ name: __filename, extractId }),
+  withFixedErrorHandler({ name: getRelativePath(__filename), extractId }),
 )(AddonBase);
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -13,14 +13,13 @@ import AddonCompatibilityError from 'amo/components/AddonCompatibilityError';
 import AddonMeta from 'amo/components/AddonMeta';
 import AddonMoreInfo from 'amo/components/AddonMoreInfo';
 import ContributeCard from 'amo/components/ContributeCard';
-import NotFound from 'amo/components/ErrorPage/NotFound';
 import DefaultRatingManager from 'amo/components/RatingManager';
 import ScreenShots from 'amo/components/ScreenShots';
 import Link from 'amo/components/Link';
 import { fetchOtherAddonsByAuthors } from 'amo/reducers/addonsByAuthors';
 import { fetchAddon } from 'core/reducers/addons';
 import { sendServerRedirect } from 'core/reducers/redirectTo';
-import { withErrorHandler } from 'core/errorHandler';
+import { withPageErrorHandler } from 'core/errorHandler';
 import InstallButton from 'core/components/InstallButton';
 import {
   ADDON_TYPE_DICT,
@@ -94,25 +93,30 @@ export class AddonBase extends React.Component {
       params,
     } = this.props;
 
-    // This makes sure we do not try to dispatch any new actions in the case
-    // of an error.
-    if (!errorHandler.hasError()) {
-      if (addon) {
-        if (!isNaN(params.slug)) {
-          // We only load add-ons by slug, but ID must be supported too because
-          // it is a legacy behavior.
-          dispatch(sendServerRedirect({
-            status: 301,
-            url: `/${lang}/${clientApp}/addon/${addon.slug}/`,
-          }));
-          return;
-        }
-
-        dispatch(setViewContext(addon.type));
-        this.dispatchFetchOtherAddonsByAuthors({ addon });
-      } else {
-        dispatch(fetchAddon({ slug: params.slug, errorHandler }));
+    if (addon) {
+      if (!isNaN(params.slug)) {
+        // We only load add-ons by slug, but ID must be supported too because
+        // it is a legacy behavior.
+        dispatch(sendServerRedirect({
+          status: 301,
+          url: `/${lang}/${clientApp}/addon/${addon.slug}/`,
+        }));
+        return;
       }
+
+      dispatch(setViewContext(addon.type));
+      this.dispatchFetchOtherAddonsByAuthors({ addon });
+    } else {
+      dispatch(fetchAddon({ slug: params.slug, errorHandler }));
+    }
+  }
+
+  componentDidMount() {
+    const { errorHandler } = this.props;
+
+    if (errorHandler.hasError()) {
+      log.debug('Clearing errors on the client');
+      this.props.dispatch(errorHandler.createClearingAction());
     }
   }
 
@@ -425,14 +429,11 @@ export class AddonBase extends React.Component {
     let errorBanner = null;
     if (errorHandler.hasError()) {
       log.warn('Captured API Error:', errorHandler.capturedError);
-      // 401 and 403 are made to look like a 404 on purpose.
-      // See https://github.com/mozilla/addons-frontend/issues/3061
-      if (errorHandler.capturedError.responseStatusCode === 401 ||
-          errorHandler.capturedError.responseStatusCode === 403 ||
-          errorHandler.capturedError.responseStatusCode === 404
-      ) {
-        return <NotFound errorCode={errorHandler.capturedError.code} />;
+
+      if (errorHandler.shouldRenderNotFound()) {
+        return errorHandler.renderNotFound();
       }
+
       // Show a list of errors at the top of the add-on section.
       errorBanner = errorHandler.renderError();
     }
@@ -625,5 +626,5 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withErrorHandler({ name: 'Addon' }),
+  withPageErrorHandler({ name: 'Addon' }),
 )(AddonBase);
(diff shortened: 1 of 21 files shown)


willdurand:
I have added a bunch of tests and fixed a few things so that it is no more a "POC". I'd love feedback now since it seems OK and it fixes actual issues.

muffinresearch:
@kumar303 would you have a chance to review this? It would be great to get this in if it looks good.

muffinresearch:
@willdurand if you have a moment could you update the description of the issue to reflect that it's ready for review rather than WiP / PoC?

muffinresearch:
I've just tried this and it works pretty well - really nice to see the right status codes for out of range pages etc.

A couple of observations: 

* I saw some weird trouble when going from a 404 to click links, however the issue cleared and I've not seen it since. Might want to double check that's working ok.
* A search with no results is still a 200, I wonder if this should be a 404 (That could easily be a follow-up). See screenshot below:
search_results_-_add-ons_for_firefox

Overall it looks good for all the cases I was looking for - the stuck loading is gone!

src/core/errorHandler.js
```diff
@@ -140,6 +167,29 @@ export function withErrorHandler({ name, id }) {
...
+ * Pass the optional `extractId` function to create "per unique page" level
+ * error handlers. This function takes the component's props and must returns a
+ * unique value based on these props (e.g., based on the `slug`, `uniqueId`,
+ * etc.).
+ *
+ * When this optional function is not supplied, we "fix" the error handler `id`
+ * (e.g., `SomeComponentPage`) otherwise we pass the `name` suffixed by `Page`,
+ * and the HOC will create the final error handler ID thanks to `extractId`.
+ */
+export const withPageErrorHandler = ({ name, extractId = null }) => {
```
kumar303: I think the `extractId` approach will solve a lot of client/server problems and not just for pages. For example, an error handler for each review in the list of reviews could have an `extractId()` that returns the ID of the review object.

Thus, we should call this something different.

* `withFixedErrorHandler()` ?
* `withMemoizedErrorHandler()` ?
* `withKeyedErrorHandler()` ?

kumar303: Since we are introducing this as a new HOC, I suggest making `extractId` a mandatory parameter. If it's falsy, throw an error. If the component can't do `extractId: ...` then it should just use the old `withErrorHandler` HOC.

src/core/reducers/search.js
```diff
@@ -4,13 +4,20 @@ import {
 } from 'core/constants';
 import { createInternalAddon } from 'core/reducers/addons';
 
+
+const ABORT_SEARCH = 'ABORT_SEARCH';
```
kumar303: for consistency with these other action names, it should probably be `SEARCH_ABORTED`.

src/core/reducers/search.js
```diff
@@ -30,6 +37,12 @@ export default function search(state = initialState, action) {
           createInternalAddon(payload.entities.addons[slug])
         )),
       };
+    case ABORT_SEARCH:
+      return {
+        ...state,
```
kumar303: I think this also needs `count: 0`. I wonder why `SEARCH_STARTED` doesn't do that though. Hmm.

src/amo/components/Addon/index.js
```diff
@@ -598,5 +595,8 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withErrorHandler({ name: 'Addon' }),
+  withPageErrorHandler({
+    name: 'Addon',
+    extractId: (ownProps) => ownProps.params.slug,
```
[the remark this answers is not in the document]
willdurand: No it does not anymore but we have an issue for that #3387.

src/amo/components/Collection/index.js
```diff
@@ -207,12 +200,11 @@ export const mapStateToProps = (state: { collections: CollectionsState }) => {
 
 export default compose(
   translate(),
-  withErrorHandler({
-    // This allows to sync the client and the server error handler ids, thus
-    // allowing the client side to be aware of errors thrown on the server.
-    // See: https://github.com/mozilla/addons-frontend/issues/3313
-    id: 'Collection-001',
```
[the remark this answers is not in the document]
willdurand: 🙈

src/amo/components/Collection/index.js
```diff
@@ -207,12 +200,11 @@ export const mapStateToProps = (state: { collections: CollectionsState }) => {
...
+  withPageErrorHandler({
     name: 'Collection',
+    extractId: (ownProps: Props) => {
+      return `${ownProps.params.user}/${ownProps.params.slug}`;
```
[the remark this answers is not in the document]
willdurand: Wow, good catch!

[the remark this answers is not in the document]
willdurand: > Part of my worry with this approach is that it will be hard to construct the right ID. But maybe we should still try it and see what happens?

I tend to agree, but you also reply to yourself correctly. All in all, the whole error handling feature saves us a lot of time and hides complexity. If the trade-off is to only be sure we have correct IDs, then I think it is a fair price we can afford.

src/amo/sagas/collections.js
```diff
@@ -42,6 +43,7 @@ export function* fetchCollection({
   } catch (error) {
     log.warn(`Collection failed to load: ${error}`);
     yield put(errorHandler.createErrorAction(error));
+    yield put(abortFetchCollection());
```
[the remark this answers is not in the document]
willdurand: >  Now that I've seen real world usage, it really isn't feasible to revert to empty states at the component level. It has to be done at the reducer level.

I tried hard, I swear it! 😛

src/amo/components/Addon/index.js
```diff
@@ -401,14 +401,11 @@ export class AddonBase extends React.Component {
...
     if (errorHandler.hasError()) {
       log.warn('Captured API Error:', errorHandler.capturedError);
-      // 401 and 403 are made to look like a 404 on purpose.
-      // See https://github.com/mozilla/addons-frontend/issues/3061
-      if (errorHandler.capturedError.responseStatusCode === 401 ||
-          errorHandler.capturedError.responseStatusCode === 403 ||
-          errorHandler.capturedError.responseStatusCode === 404
-      ) {
+
+      if (errorHandler.shouldRenderNotFound()) {
```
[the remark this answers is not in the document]
willdurand: I reverted this change for the add-ons, and only handled 404 in the other components.

src/amo/components/Addon/index.js
```diff
@@ -598,5 +595,8 @@ export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withErrorHandler({ name: 'Addon' }),
+  withPageErrorHandler({
+    name: 'Addon',
```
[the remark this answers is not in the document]
willdurand: it works!

src/core/errorHandler.js
willdurand: `withFixedErrorHandler()`

willdurand:
I have created a `getRelativePath()` function because `__filename` behaves differently in the browser and in node.

The webpack polyfill for `__filename` in the browser does not return an absolute path but a relative path, relative to the `context` option in the webpack configuration. On the server, it uses the node version, which returns an absolute path. The problem is that we get different results for a same `__filename` call depending on where the code is executed. This function fixes this by mimicking webpack's behavior.

willdurand:
It should be good for a new review.

src/core/utils/index.js
```diff
@@ -366,3 +367,11 @@ export function addonHasVersionHistory(addon) {
     ADDON_TYPE_THEME,
   ].includes(addon.type);
 }
+
+export const getRelativePath = (filename, dirname = __dirname) => {
+  if (dirname) {
+    return path.relative(path.resolve(dirname, '../../../'), filename);
```
kumar303: This looks like it won't work on Windows and we might forget to update it in case `utils/index.js` gets moved.

Since it only needs to guarantee a consistent path, what about a simpler approach more like this?

```js
let fileId = filename;
if (!fileId.startsWith('src')) {
  fileId = fileId.replace(/^.*src/, 'src');
}
```

tests/unit/amo/components/TestCollection.js
```diff
@@ -507,4 +474,30 @@ describe(__filename, () => {
     const wrapper = renderComponent();
     expect(wrapper.find('title')).toHaveLength(0);
   });
+
+  describe('errorHandler - extractId', () => {
+    it('returns a unique ID based on params', () => {
+      const props = {
```
kumar303: for maintainability, this should use the helper like `getProps({ params: { user: 'foo', ... } })`

tests/unit/core/reducers/test_search.js
```diff
@@ -20,6 +23,23 @@ describe(__filename, () => {
...
+        filters: { query: 'foo' },
+      }));
+      expect(state.filters).toEqual({ query: 'foo' });
+      expect(state.loading).toBe(true);
+      expect(state.results).toEqual([]);
+
+      const newState = search(state, abortSearch());
+      expect(newState.filters).toEqual({ query: 'foo' });
+      expect(newState.loading).toBe(false);
+      expect(newState.results).toEqual([]);
```
kumar303: I guess this should assert `count == 0` for completeness.

webpack.dev.config.babel.js
```diff
@@ -69,6 +69,11 @@ export default Object.assign({}, webpackConfig, {
   module: {
     rules: getRules({ babelQuery: BABEL_QUERY, bundleStylesWithJs: true }),
   },
+  node: {
```
kumar303: You don't need this because all configs [extend the prod config](https://github.com/willdurand/addons-frontend/blob/fd31ac940d0fc509509d648f96117293ffc8f29a/webpack.dev.config.babel.js#L56).

src/amo/components/Addon/index.js
```diff
@@ -599,9 +602,13 @@ export function mapStateToProps(state, ownProps) {
...
+export const extractId = (ownProps) => {
+  return ownProps.params.slug;
+};
+
 export default compose(
   translate({ withRef: true }),
   connect(mapStateToProps),
   withInstallHelpers({ src: 'dp-btn-primary' }),
-  withErrorHandler({ name: 'Addon' }),
+  withFixedErrorHandler({ name: getRelativePath(__filename), extractId }),
```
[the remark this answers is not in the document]
willdurand: right but then it introduces a new param to the HOC.

src/amo/components/Collection/index.js
```diff
@@ -172,13 +175,9 @@ export class CollectionBase extends React.Component<Props> {
 
     if (errorHandler.hasError()) {
       log.warn('Captured API Error:', errorHandler.capturedError);
-      // 401 and 403 are for an add-on lookup is made to look like a 404 on
-      // purpose. See: https://github.com/mozilla/addons-frontend/issues/3061.
-      if (errorHandler.capturedError.responseStatusCode === 401 ||
```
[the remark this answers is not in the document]
willdurand: yes, there was no real use case except handling the errors.

src/core/utils/index.js
willdurand: sounds good, will try this.

webpack.dev.config.babel.js
willdurand: right. I forgot to remove it after having tried a few things...

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
kumar303: I think that's fine. Eventually we could get rid of `name` once all of the files have been converted to `fileName`.

src/amo/components/Collection/index.js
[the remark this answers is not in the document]
kumar303: Yeah but did someone explicitly request for 401/403 responses from the collections API to be rendered as 404s? I hope not. I don't see why we would need to do that because the behavior is different from the add-on scenario.

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
willdurand: 👍

src/amo/components/Collection/index.js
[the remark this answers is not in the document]
willdurand: > Yeah but did someone explicitly request for 401/403 responses from the collections API to be rendered as 404s? I hope not.

No I don't think so.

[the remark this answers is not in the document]
muffinresearch: FWIW: There's more info in the linked issue: https://github.com/mozilla/addons-frontend/issues/3061

muffinresearch:
@kumar303 @willdurand if we can get this onto -dev today and it tests out well, I'd like to consider getting this out this week. It fixes some pretty gnarly issues.

src/amo/components/Search/index.js
```diff
@@ -235,5 +245,5 @@ export function mapStateToProps(state) {
 export default compose(
   connect(mapStateToProps),
   translate(),
-  withErrorHandler({ name: 'Search' }),
+  withErrorHandler({ id: getRelativePath(__filename) }),
```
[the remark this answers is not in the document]
willdurand: We do not have a `Fixed` error handler here because there is no implementation of a `extractId` function for a search page.

src/amo/sagas/collections.js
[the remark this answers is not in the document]
willdurand: done in #3838

src/core/utils/index.js
willdurand: Note: I assumed Node would fix the Windows vs UNIX difference as done in many languages.

src/amo/components/Search/index.js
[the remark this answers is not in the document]
willdurand: Hmm, wait. maybe there is a use case for `withFixedErrorHandler` here.

src/amo/components/SearchFilters/index.js
```diff
@@ -227,5 +227,5 @@ export default compose(
   withRouter,
   connect(mapStateToProps),
   translate(),
-  withErrorHandler({ name: 'Search' }),
+  withErrorHandler({ name: 'SearchFilters' }),
```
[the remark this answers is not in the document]
willdurand: I decided to leave it as is (well, fixing the `name` though) and to update all error handlers in a follow-up patch.

willdurand:
I have addressed all the changes, so it should be ready for a new look.

I would like to update the `withErrorHandler` and other components (with or without `withFixedErrorHandler`) in a follow-up patch.

kumar303:
> I'd like to consider getting this out this week.

Since there is no realistic path a user can take to trigger the bugs that this patch fixes, I don't think we should rush it. There are plenty of unrealistic paths :)

tests/unit/core/utils/test_index.js
```diff
@@ -63,795 +64,810 @@ import {
...
-  function createAddonWithType(type) {
-    return createInternalAddon({ ...fakeAddon, type });
-  }
-
-  it('requires an addon', () => {
-    expect(() => {
-      addonHasVersionHistory();
-    }).toThrow('addon is required');
-  });
+describe(__filename, () => {
```
kumar303: You are braver than me for fixing the indent 👀

tests/unit/core/utils/test_index.js
```diff
@@ -63,795 +64,810 @@ import {
...
+  describe('getRelativePath', () => {
+    it('returns a path relative to the project root directory', () => {
+      expect(getRelativePath('/path/to/src/foo/index.js'))
+        .toEqual('src/foo/index.js');
+    });
+
+    it('returns the given filename if `src` is not in it', () => {
+      const filename = 'tests/unit/core/utils/test_index.js';
+
+      expect(getRelativePath(filename)).toEqual(filename);
```
kumar303: I'd like to see an additional test that asserts how `getRelativePath('src/file.js')` results in `src/file.js` -- that is, no stripping.

src/core/utils/index.js
```diff
@@ -366,3 +366,12 @@ export function addonHasVersionHistory(addon) {
     ADDON_TYPE_THEME,
   ].includes(addon.type);
 }
+
+export const getRelativePath = (filename) => {
```
kumar303: This should live in `errorHandler` and be named something more like `normalizeFileNameId()` because I don't want someone to think they can use it to get relative paths :)

tests/unit/core/utils/test_index.js
willdurand: ah correct 👍

src/core/utils/index.js
willdurand: That's right, I was not sure about moving it or not, to be honest. I told myself "there is 99% chance someone will want to get a relative path of a source file and 1% will be @kumar303 telling me to put it in the `errorHandler`".

muffinresearch:
> Since there is no realistic path a user can take to trigger the bugs that this patch fixes, I don't think we should rush it. There are plenty of unrealistic paths :)

Yep makes sense to get it right, I'm happy for it to go out next week on that basis.

tests/unit/core/utils/test_index.js
willdurand: yes, sorry for that actually. it makes the diff less readable..

willdurand: The PR does not add anything in this file anymore since I moved the logic + test cases to the `errorHandler`.

src/amo/components/Search/index.js
```diff
@@ -236,8 +246,12 @@ export function mapStateToProps(state) {
   };
 }
 
+export const extractId = (ownProps) => {
+  return parsePage(ownProps.filters.page);
```
kumar303: I struggled to understand this when I first saw it. I think that this ID does not need to differentiate between component instances because the error handler gets cleared every time the search filters change, right? If so, maybe adding a comment here would be helpful to the next reader.

src/core/errorHandler.js
```diff
@@ -10,6 +11,15 @@ function generateHandlerId({ name = '' } = {}) {
   return `${name}-${Math.random().toString(36).substr(2, 9)}`;
 }
 
+export const normalizeFileNameId = (filename) => {
+  let fileId = filename;
+  if (!fileId.startsWith('src')) {
```
kumar303: I just realized that nothing is obvious about what this code is doing. I would suggest adding a comment saying that on the server we get an absolute path and on the client we get a relative path. This function will make server and client code return the same value.

tests/unit/core/testErrorHandler.js
```diff
@@ -357,4 +393,68 @@ describe('errorHandler', () => {
...
+    });
+
+    it('creates an error handler with a fixed ID', () => {
+      const { component } = createFixedErrorComponent({
+        extractId: () => 'unique-id-based-on-props',
+      });
+      const errorHandler = component.props.errorHandler;
+      expect(errorHandler).toBeInstanceOf(ErrorHandler);
+      expect(errorHandler.id)
+        .toEqual('src/SomeComponent/index.js-unique-id-based-on-props');
```
kumar303: Unlike the other tests, this one relies on `fileName` in its assertion. Thus, you should declare it explicitly when calling `createFixedErrorComponent()`. Relying on implicit defaults can make debugging failed assertions difficult.

src/core/errorHandler.js
willdurand: ah sure, that totally makes sense.

PR #3779, opened 2017-11-03
fix: Combine all and beta versions

### LINKED ISSUE #3768: 'Version History' is superflous to the 'See all versions' link
## First impresion …

> Version History
> See all versions

Where there's any history – any more than one version of an add-on: 

- you need not present the non-linked subheading (Version History) above the link to the history (of all versions). 

## Enhancement request

Be less verbose; save a little space. 

## Thoughts

The phrase _See all versions_ is relatively plain English. (To me, the phrase _Version history_ is fractionally more geeky.)

If the add-on has a beta/development/experimental channel then I'll expect _all versions_ to **truly** list all versions – including those that are experimental (notionally (without ploughing through repos of issues) two columns, with releases to the left). 

Notes to self: 

- whilst _beta_ is a good word, it's debatably not good when alphas are generalised in that way
- _experimental_, yeah, and the purple flask, that works for me. 

### Suggested priority

Low.

tofumatt:
(fix #3768)

Simple tweak–just combines the version links into one section, similar to the links section at the top of the card.

### Before
screenshot 2017-11-03 23 37 14

### After
screenshot 2017-11-03 23 36 14

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -132,22 +132,26 @@ export class AddonMoreInfoBase extends React.Component<Props> {
         </Link>
       ) : null,
       versionHistoryLink: addonHasVersionHistory(addon) ? (
-        <Link
-          className="AddonMoreInfo-version-history-link"
-          href={`/addon/${addon.slug}/versions/`}
-        >
-          {i18n.gettext('See all versions')}
-        </Link>
+        <li>
+          <Link
+            className="AddonMoreInfo-version-history-link"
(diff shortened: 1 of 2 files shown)


willdurand:
You probably want to rebase since the error you get has been fixed.

tofumatt:
Yeah I forgot to push after I did that locally, will do! 😅

grahamperrin:
:+1: thanks, this is good. 

At a later date, maybe rationalise things further … better use of space through condensing, having all version-related subsections in one space (not separated by three other subsections), and so on. 

Dust-settling first. Something for 2018 maybe :-)

PR #3796, opened 2017-11-06
Redirect atype parameters to their corresponding addonType

### LINKED ISSUE #3791: Redirect atype parameters to the new type parameter
via https://www.reddit.com/r/firefox/comments/7aza66/no_search_engines_section_on_mozillas_restyled/

From Firefox in about:preferences#search there's a search tools link https://addons.mozilla.org/en-US/firefox/search/?atype=4

preferences

This shows no results via the new frontend. That link should now be https://addons.mozilla.org/en-US/firefox/search/?type=search

We need to redirect from the former to the latter.

willdurand:
Fix #3791

---

This PR triggers a server-side redirect when `atype` search parameter is detected and we have a corresponding `addonType` for it.

### DIFF
--- a/src/amo/components/SearchPage/index.js
+++ b/src/amo/components/SearchPage/index.js
@@ -5,6 +5,14 @@ import { compose } from 'redux';
 
 import Search from 'amo/components/Search';
 import {
+  ADDON_TYPE_DICT,
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_LANG,
+  ADDON_TYPE_OPENSEARCH,
+} from 'core/constants';
+import { makeQueryString } from 'core/api';
+import { sendServerRedirect } from 'core/reducers/redirectTo';
+import {
   convertFiltersToQueryParams,
   convertQueryParamsToFilters,
 } from 'core/searchUtils';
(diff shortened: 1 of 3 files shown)


src/amo/components/SearchPage/index.js
```diff
@@ -5,39 +5,89 @@ import { compose } from 'redux';
...
-  });
+export class SearchPageBase extends React.Component<Props> {
+  componentWillMount() {
+    const { clientApp, filters, lang, location } = this.props;
+
+    // Map the old `atype` parameter to its corresponding `adddonType`.
+    // See: https://github.com/mozilla/addons-frontend/issues/3791.
+    const newFilters = { ...filters };
+    if (location.query.atype) {
+      switch (String(location.query.atype)) {
```
tofumatt: This logic seems like it would be better placed in a utility function (maybe inside `searchUtils`).

willdurand: I like the idea but I am not sure if we will reuse the logic somewhere else. Since it is a "patch" for a legacy behavior and it triggers a server-side redirect, maybe having everything in the same file makes everything more obvious. That being said, I have no preference.

muffinresearch:
This appears to remove other query parameters.

muffinresearch:
Ah ok I see now it looks like it only removes unexpected query parameters e.g. q=foo is ok but foo=1 is removed.

willdurand:
> Ah ok I see now it looks like it only removes unexpected query parameters e.g. q=foo is ok but foo=1 is removed.

Yes, that's the current behavior actually. Do we want to change that too?

muffinresearch:
I could see that might not be desirable in future in-case there's something on the url that the app doesn't need but something else does. We can probably cross that bridge later though. 

I'm also wondering if this should be a 302 rather than 301 until we remove the old site.

willdurand:
> I could see that might not be desirable in future in-case there's something on the url that the app doesn't need but something else does. We can probably cross that bridge later though.

This patch #3765 may introduce it then.

willdurand:
> I'm also wondering if this should be a 302 rather than 301 until we remove the old site.

What would be the advantage? If we really want to avoid this `atype` parameter, shouldn't we communicate this intent?

muffinresearch:
> I'm also wondering if this should be a 302 rather than 301 until we remove the old site.

Because the new site still exists, and if you switch between the two a 301 is likely to be implicitly cached.

willdurand:
> Because the new site still exists, and if you switch between the two a 301 is likely to be implicitly cached.

oh right, caching 😅

willdurand:
Ok rebased and updated the code with `statusCode = 302`.

willdurand:
@muffinresearch looks good to you now?

muffinresearch:
@willdurand looks good 🛳

Issue #3797, opened 2017-11-06
Header menu is visually misaligned with search bar


kumar303:
### Describe the problem and steps to reproduce it:

* Go to any page using a desktop browser with a small or medium sized window

### What happened?

The header menu is misaligned with the search bar

### What did you expect to happen?

The bottom of the menu should line up with the bottom of the search bar. The mocks don't have an exact version of these screen sizes but the spec for the extra large header shows a bottom alignment.

### Anything else we should know?

Large screen (1432px wide)

<img width="716" alt="screenshot 2017-11-06 08 40 20" src="https://user-images.githubusercontent.com/55398/32446581-fc2d7eaa-c2ce-11e7-9b52-985ce0ee1d0e.png">

![add ons for firefox - large](https://user-images.githubusercontent.com/55398/32446750-7b08f308-c2cf-11e7-926a-5596d566c866.png)

Small screen (1236px wide)

<img width="618" alt="screenshot 2017-11-06 08 42 03" src="https://user-images.githubusercontent.com/55398/32446604-11b6e8f6-c2cf-11e7-8f38-a16d83f01bfc.png">

![add ons for firefox 1](https://user-images.githubusercontent.com/55398/32446631-2744f762-c2cf-11e7-80fe-b84e2559fb2a.png)

AlexandraMoga:
Verified as fixed on AMO-dev with FF56 (Win10 and Android 7.0)

![image](https://user-images.githubusercontent.com/31961530/32487400-4424e5ca-c3b2-11e7-889b-09c95a405729.png)

![image](https://user-images.githubusercontent.com/31961530/32487527-95929d3a-c3b2-11e7-96c2-f0b518d88ac5.png)

PR #3799, opened 2017-11-06
Fix new logo margin

muffinresearch:
Fixes https://github.com/mozilla/addons-frontend/issues/3797

screen shot 2017-11-06 at 15 10 11
screen shot 2017-11-06 at 15 10 02
screen shot 2017-11-06 at 15 09 57
screen shot 2017-11-06 at 15 09 50
screen shot 2017-11-06 at 15 09 41

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -17,7 +17,7 @@
     padding: 0 $header-footer-gutter 10px;
   }
 
-  @include respond-to(extraExtraLarge) {
+  @include respond-to(extraLarge) {
     grid-template-columns: max-content 1fr 1fr;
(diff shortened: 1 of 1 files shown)


muffinresearch:
> The large size looks like it's running really tight against the search bar so I wonder if another even slightly more verbose locale will mess with things, but looks good to me.

Currently there's a max width on there and the search-bar shrinks up to the point that the next break point kicks in.

muffinresearch:
> I see a different between large and medium screen. Vertical alignment is not the same.

Ah yep, I see that now too - I'll fix it for consistency.

muffinresearch:
@willdurand that's updated now if you want to retry?

PR #3800, opened 2017-11-06
Search box should work the same from all sections of the site

### LINKED ISSUE #3689: Search box should work the same from all sections of the site
Instead of having "Find add-ons", "Find extensions", and "Find themes", there should only be "Find add-ons". If users need to filter down results, there are controls in the search results page. I think it's better to provide a consistent experience so users can find any add-on from anywhere on the site.

willdurand:
Fix #3689

---

This PR removes the contextual search depending on the landing page a user is viewing.

### DIFF
--- a/src/amo/components/SearchForm/index.js
+++ b/src/amo/components/SearchForm/index.js
@@ -10,13 +10,6 @@ import defaultDebounce from 'lodash.debounce';
 import deepEqual from 'deep-eql';
 
 import Suggestion from 'amo/components/SearchSuggestion';
-import {
-  ADDON_TYPE_EXTENSION,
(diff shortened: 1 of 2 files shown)


tofumatt:
Hmm, maybe we want to wait on #3689 being discussed before merging this, actually.

willdurand:
> Hmm, maybe we want to wait on #3689 being discussed before merging this, actually.

Yes, adding the right label now.

tofumatt:
We chatted about this and think that the search being global makes the most sense–especially as it's more intuitive to narrow results than expand them.

willdurand:
ah, good news 😺

PR #3805, opened 2017-11-06
Fix download links (add utm query params)

### LINKED ISSUE #3803: Fix links to Firefox marketing/download on new AMO pages

1. Go here (or really to any addon page) *in Chrome*: https://addons.mozilla.org/en-US/firefox/addon/video-downloadhelper/
2. Hover over the green "Download Firefox" button in the header -- the link is https://mozilla.org/firefox/
3. Hover over the red "download Firefox" message in the page -- the link is https://www.mozilla.org/en-US/firefox/
4. Hover over the "Download Firefox" link in the footer -- the link is https://www.mozilla.org/firefox/new/?utm_source=addons.mozilla.org&utm_campaign=footer&utm_medium=referral

### What happened?

The footer link is correct. Other links are incorrect because...
1. Links shouldn't have l10n on them (we should let the destination site handle l10n routing)
2. Links should have link-specific UTM parameters on them
3. "Download Firefox" links should point at the download page at /firefox/new. 

### What needs to happen?

We should replace the **green header link** with this:

https://www.mozilla.org/firefox/new/?utm_source=addons.mozilla.org&utm_medium=referral&utm_campaign=non-fx-button&utm_content=header-download-button

We should replace the **red in-page link** with this:

https://www.mozilla.org/firefox/new/?utm_source=addons.mozilla.org&utm_medium=referral&utm_campaign=non-fx-button&utm_content=install-addon-button

willdurand:
Fix #3803

---

This PR adds `utm` query parameters to the "download firefox" links.

### DIFF
--- a/src/amo/components/AddonCompatibilityError/index.js
+++ b/src/amo/components/AddonCompatibilityError/index.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 
+import { makeQueryStringWithUTM } from 'amo/utils';
 import {
   INCOMPATIBLE_FIREFOX_FOR_IOS,
   INCOMPATIBLE_NO_OPENSEARCH,
@@ -22,7 +23,6 @@ import './style.scss';
 export class AddonCompatibilityErrorBase extends React.Component {
   static propTypes = {
     i18n: PropTypes.object.isRequired,
-    lang: PropTypes.string.isRequired,
     log: PropTypes.object,
     minVersion: PropTypes.string.isRequired,
     reason: PropTypes.string.isRequired,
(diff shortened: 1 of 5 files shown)


src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -37,14 +36,19 @@ export class AddonCompatibilityErrorBase extends React.Component {
...
-      lang,
       log,
       minVersion,
       reason,
       userAgentInfo,
     } = this.props;
-    const downloadUrl = `https://www.mozilla.org/${lang}/firefox/`;
-    let message;
+
+    const queryString = [
```
tofumatt: I'm pretty sure we have an actual `makeQueryParam` helper or something which *might* be nicer just because it already has tests and takes an object as an argument, but this is fine 😄

willdurand: ah, indeed `makeQueryString` it is called 😓  I'll use it!

Issue #3563, opened 2017-10-18
Prevent against using a component with missing properties


kumar303:
### Describe the problem and steps to reproduce it:

Consider a component like this:

```js
import * as React from 'react';
import { compose } from 'redux';

import translate from 'core/i18n/translate';

class ThingBase extends React.Component<{| color: string |}> {
  render() {}
}

export default compose(
  translate(),
)(ThingBase);
```

### What happened?

```js

import Thing from '...';
// This should cause a Flow error because it's missing the color prop.
const thing = (<Thing />);
```

It does not raise any Flow error.

### What did you expect to happen?

It should raise a Flow error [like this](https://flow.org/try/#0JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQduANmugCoAWwAdgOZxYAHjCzcAJumx4YAOgDCJSN1EwAPAG8APnFwQm0AFxxUMKD36aAvgD446unDiVxWKAAoAlHct0fjCNwmcDBcfHAAvHBuqpzmcAD01h40QA):

```
7: const thing = (<Thing />);
                  ^ props of React element `Thing`. This type is incompatible with
3: class Thing extends React.Component<{| color: string |}> {
                                       ^ object type
Property `color` is incompatible:
3: class Thing extends React.Component<{| color: string |}> {
                                       ^ property `color`. Property not found in
7: const thing = (<Thing />);
                  ^ props of React element `Thing`
```

### Anything else we should know?

I think we need to define types for `compose()` and `translate()` as well as any other HOCs we use.

Note that we still get reported errors when prop values are invalid.

kumar303:
Some patches landed to fix this but we use a type hint rather than actually letting Flow follow all permutations introduced by HOCs. There is an example here of how to write code using the type hint: https://github.com/mozilla/addons-frontend#flow

PR #3806, opened 2017-11-06
Add new Contribute card to add-on detail page

### LINKED ISSUE #3149: Implement UI for new contributions feature
We're replacing contributions with a simple link. This needs implementing on the new frontend but we will need UX for it. It should be relatively simple in the sense this won't be much more than a link and some context.

willdurand:
Fix #3149

---

This PR adds a new "Contribute" card to the add-on detail page, exposing a button with a link to the "contributions URL".

## Screenshot

screen shot 2017-11-06 at 20 02 42


src/amo/components/Addon/index.js
```diff
@@ -385,6 +386,22 @@ export class AddonBase extends React.Component {
...
 
+  renderContributeCard() {
+    const { addon } = this.props;
+
+    if (!addon || (addon && !addon.contributions_url)) {
+      return null;
+    }
+
+    return (
+      <Contribute
```
tofumatt: Dunno if we should call this component `ContributeCard`. Maybe?

src/amo/components/Addon/index.js
```diff
@@ -385,6 +386,22 @@ export class AddonBase extends React.Component {
...
+        addonType={addon.type}
```
tofumatt: Given that this card seems only to use props from `addon` I wonder if that should just be the only prop it takes. Then it can have a simpler proptype and can internally handle checking for the right attributes. It could also just return `null` if `addon` is undefined.

src/amo/components/Addon/index.js
```diff
@@ -545,6 +562,8 @@ export class AddonBase extends React.Component {
 
           {this.renderRatingsCard()}
 
+          {this.renderContributeCard()}
```
tofumatt: If the `Contribute` component accepted just `addon` as previously suggested this wouldn't need to be a method with logic but a simple tag: `<Contribute addon={addon} />`.

src/amo/components/Contribute/index.js
```diff
@@ -0,0 +1,91 @@
...
+        >
+          <Icon name="heart" />
+          {i18n.gettext('Contribute now')}
+        </Button>
+      </p>
+    </Card>
+  );
+};
+
+export default translate()(Contribute);
```
tofumatt: I know it's only using one HOC but I think this should still use `compose` for consistency and ease-of-HOC-adding in the future.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1238,6 +1239,46 @@ describe(__filename, () => {
...
   });
+
+  describe('Contribute', () => {
+    it('does not render a Contribute card when there is no add-on', () => {
+      const root = shallowRender({ addon: null, params: { slug: 'some-slug' } });
+
+      expect(root.find(Contribute)).toHaveLength(0);
+    });
+
+    it('does not render a Contribute card when there is no contributions URL', () => {
```
tofumatt: Again, these seem like tests that should be moved to the `Contribute` component. I think if we had it deal with the `addon` object as a whole and just render `null` when things are missing that'd be good and this component would need fewer tests.

tests/unit/amo/components/TestContribute.js
```diff
@@ -0,0 +1,89 @@
...
+  const render = (props = {}) => {
+    const allProps = {
+      contributionsURL: 'https://paypal.me/babar',
+      addonType: ADDON_TYPE_EXTENSION,
+      numberOfAuthors: 1,
+      i18n: fakeI18n(),
+      ...props,
+    };
+
+    return shallow(<Contribute {...allProps} />, {
```
tofumatt: This seems a bit weird, why not just use `shallowUntilTarget`?

src/amo/components/Addon/index.js
```diff
@@ -385,6 +386,22 @@ export class AddonBase extends React.Component {
...
+    const { addon } = this.props;
+
+    if (!addon || (addon && !addon.contributions_url)) {
+      return null;
+    }
+
+    return (
+      <Contribute
+        addonType={addon.type}
+        authorNames={addon.authors.length}
```
kumar303: This prop should be `numberOfAuthors`. Even though `Addon/index.js` doesn't use Flow, we still need to address https://github.com/mozilla/addons-frontend/issues/3563 before this type of thing can be caught.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1238,6 +1239,46 @@ describe(__filename, () => {
...
+        ...fakeAddon,
+        contributions_url: contributionsURL,
+      });
+      store.dispatch(_loadAddons({ addon }));
+
+      const root = renderComponent({ params: { slug: addon.slug }, store });
+
+      expect(root.find(Contribute)).toHaveLength(1);
+      expect(root.find(Contribute))
+        .toHaveProp('contributionsURL', contributionsURL);
```
kumar303: This test should validate all other properties too: `addonType`, `numberOfAuthors`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1238,6 +1239,46 @@ describe(__filename, () => {
...
+      expect(root.find(Contribute)).toHaveLength(0);
+    });
+
+    it('does not render a Contribute card when there is no contributions URL', () => {
+      const contributionsURL = null;
+
+      const { store } = dispatchClientMetadata();
+      const addon = createInternalAddon({
+        ...fakeAddon,
+        contributions_url: contributionsURL,
```
kumar303: This can be `contributions_url: null`

tests/unit/amo/components/TestContribute.js
```diff
@@ -0,0 +1,89 @@
...
+      context: { i18n: allProps.i18n },
+    }).shallow();
```
kumar303: Instead of rendering twice, use `shallowUntilTarget()`. That way it will be a self-maintaining test as we add more HOC decorators.

tests/unit/amo/components/TestContribute.js
```diff
@@ -0,0 +1,89 @@
...
+    const allProps = {
+      contributionsURL: 'https://paypal.me/babar',
+      addonType: ADDON_TYPE_EXTENSION,
+      numberOfAuthors: 1,
+      i18n: fakeI18n(),
+      ...props,
+    };
+
+    return shallow(<Contribute {...allProps} />, {
+      context: { i18n: allProps.i18n },
```
kumar303: You should not need this. If you still get errors about a missing `i18n` prop then the assertions for translation strings may need a new approach.

tests/unit/amo/components/TestContribute.js
```diff
@@ -0,0 +1,89 @@
...
+    expect(root.find('.Contribute-content'))
+      .toHaveText(oneLine`The developer of this extension asks that you help
+        support its continued development by making a small contribution.`);
+  });
+
+  it('displays content for multiple extension developers', () => {
+    const root = render({ numberOfAuthors: 3 });
+    expect(root.find(Card)).toHaveProp('header', 'Support these developers');
+    expect(root.find('.Contribute-content'))
+      .toHaveText(oneLine`The developers of this extension ask that you help
```
kumar303: Since these tests only check that the alternate translation is *used*, you don't have to include the full string in the assertion. You could just check a subset: `.toMatch(/The developers of this extension/)`. That should cut down on having to fix tests when making minor changes to the copy.

src/amo/components/Contribute/index.js
```diff
@@ -0,0 +1,91 @@
...
+      break;
+  }
+
+  return (
+    <Card className="Contribute" header={header}>
+      <p className="Contribute-content">{content}</p>
+      <p>
+        <Button
+          className="Contribute-button Button--action Button--fullwidth"
+          href={contributionsURL}
```
kumar303: According to the [PRD](https://docs.google.com/document/d/1DjXQbl_hZlRe4bqE9B-cLPfq64wYFa19TU9DfsOpckQ/edit) this should be a `target=_blank` link

src/amo/components/Addon/index.js
willdurand: oh no, I badly rebased it seems 😕

tests/unit/amo/components/TestContribute.js
willdurand: It works but there is a warning complaining about missing `i18n` in the context..

willdurand:
I rushed a bit this PR and I am very sorry. This should be looking better now.

tofumatt:
No worries! 😄

I'll have another look now 👍

PR #3807, opened 2017-11-06
Not compatible with Firefox Quantum

### LINKED ISSUE #3780: yellow or red alert for 'Legacy'
For users of 56.x – and [until mid 2018](https://wiki.mozilla.org/RapidRelease/Calendar#Future_branch_dates), users of ESR 52.x: 

- there should be an upfront indication wherever an extension is legacy. 

## Precedents

Established use of a yellow rectangle for LEGACY at `about:addons` within the browser. 

The subtle new look yellow alert alongside _Restart required_ at, for example, https://addons.mozilla.org/addon/quick-locale-switcher-2/

## Thoughts

The requirement for _Legacy_ extensions to be **clearly identifiable** as such probably arises primarily from the new look disappearance of clear identification (green badge) wherever an extension will be compatible with _Firefox 57+_

As there's space for the subtle yellow ⟲ for the relatively rare extensions that will require a restart of Firefox, there must be space for an unmistakable yellow alert for the relatively commonplace extensions that will be unusable with Firefox 57 and greater (all of which will, I guess, disappear from AMO sometime in Q4 2018).

It's highly desirable to have the distinction at a time of **increased likelihood of people 'shopping'** for extensions ahead of the mid-November release of 57. 

### Urgency

If you can (please): 

- Monday 2017-11-06 for a Tuesday freeze and push on Thursday 2017-11-09

– TIA.  

## Related

https://www.reddit.com/r/firefox/comments/7adng2/did_the_addon_page_change/dp9833c/ (9 points, recently up from 7):

> How can I see if an addon is FF 57+ ready now?

https://www.reddit.com/r/firefox/comments/7adng2/did_the_addon_page_change/dp95kcl/ (17 points, rising): 

> … can't tell anymore … 57 or not.

willdurand:
Fix #3780

---

This PR adds a new badge to mark the non-webext add-ons as "not compatible with FF Quantum".

## Screenshots

screen shot 2017-11-06 at 21 06 44
screen shot 2017-11-06 at 20 47 57
screen shot 2017-11-06 at 20 48 05
screen shot 2017-11-06 at 20 48 08

### DIFF
--- a/src/amo/components/AddonBadges/index.js
+++ b/src/amo/components/AddonBadges/index.js
@@ -30,6 +30,9 @@ export const AddonBadgesBase = (props: Props) => {
     }
   };
 
+  const isIncompatible = addon && addon.type === ADDON_TYPE_EXTENSION &&
+    !addon.isWebExtension;
+
   return (
     <div className="AddonBadges">
       {addon && addon.is_featured ? (
@@ -48,6 +51,12 @@ export const AddonBadgesBase = (props: Props) => {
         <Badge
           type="experimental"
           label={i18n.gettext('Experimental')}
+        />
+      ) : null}
+      {isIncompatible ? (
+        <Badge
+          type="not-compatible"
+          label={i18n.gettext('Not compatible with Firefox Quantum')}
         />
       ) : null}
     </div>
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -3,15 +3,18 @@
 @import "~ui/css/vars";
 
 .AddonBadges {
-  @include respond-to(medium) {
+  width: 100%;
+
+  @include respond-to(large) {
     display: flex;
+    flex-flow: row wrap;
   }
 }
 
 .AddonBadges .Badge {
   @include text-align-end();
 
-  @include respond-to(medium) {
+  @include respond-to(large) {
     @include margin-end(6px);
     @include text-align-start();
   }
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -231,6 +231,7 @@ export function createInternalAddon(
       windows: undefined,
     },
     isRestartRequired: false,
+    isWebExtension: false,
   };
 
   if (addon.type === ADDON_TYPE_THEME && apiAddon.theme_data) {
@@ -262,6 +263,9 @@ export function createInternalAddon(
     });
     addon.isRestartRequired = apiAddon.current_version.files.some(
       (file) => !!file.is_restart_required
+    );
+    addon.isWebExtension = apiAddon.current_version.files.some(
+      (file) => !!file.is_webextension
     );
   }
 
--- a/tests/unit/amo/components/TestAddonBadges.js
+++ b/tests/unit/amo/components/TestAddonBadges.js
@@ -126,4 +126,14 @@ describe(__filename, () => {
 
     expect(root.find(Badge)).toHaveLength(0);
   });
+
+  it('does not display a badge when the addon is not an extension', () => {
+    const addon = createInternalAddon(createFakeAddon({
+      type: ADDON_TYPE_THEME,
+      files: [{ is_webextension: false }],
+    }));
+    const root = shallowRender({ addon });
+
+    expect(root.find(Badge)).toHaveLength(0);
+  });
 });
--- a/tests/unit/amo/components/TestAddonBadges.js
+++ b/tests/unit/amo/components/TestAddonBadges.js
@@ -106,4 +106,24 @@ describe(__filename, () => {
 
     expect(root.find(Badge)).toHaveLength(0);
   });
+
+  it('displays a badge when the addon is not a web extension', () => {
+    const addon = createInternalAddon(createFakeAddon({
+      files: [{ is_webextension: false }],
+    }));
+    const root = shallowRender({ addon });
+
+    expect(root.find(Badge)).toHaveProp('type', 'not-compatible');
+    expect(root.find(Badge))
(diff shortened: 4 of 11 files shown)


grahamperrin:
👍 looks good to me in these contexts, thanks. 

It'll be nice to also have the icon in listings e.g. as shown in the shots at https://github.com/mozilla/addons-frontend/issues/3780#issuecomment-342248970

src/amo/components/AddonBadges/index.js
```diff
@@ -50,6 +50,12 @@ export const AddonBadgesBase = (props: Props) => {
           label={i18n.gettext('Experimental')}
         />
       ) : null}
+      {addon && !addon.isWebExtension ? (
```
tofumatt: What if the extension is:

* a theme
* a language tool
* a search plugin

Will it be marked as incompatible then?

src/amo/components/AddonBadges/styles.scss
```diff
@@ -3,15 +3,20 @@
 @import "~ui/css/vars";
 
 .AddonBadges {
-  @include respond-to(medium) {
+  width: 100%;
+
+  @include respond-to(large) {
     display: flex;
+    flex-flow: row wrap;
```
tofumatt: Thanks, I meant to add this and maybe I forgot or was gonna do it in a follow-up. Much appreciated! 👍

muffinresearch:
Is the distance between the 3 labels in the 1st screenshot the same? It looks to me like the distance from the "Restart required" icon to the "Not..." is bigger?

src/amo/components/AddonBadges/index.js
willdurand: oh shoot, that's right. I totally forgot this check :|

src/core/reducers/addons.js
```diff
@@ -262,6 +263,9 @@ export function createInternalAddon(
     addon.isRestartRequired = apiAddon.current_version.files.some(
       (file) => !!file.is_restart_required
     );
+    addon.isWebExtension = apiAddon.current_version.files.some(
```
kumar303: This isn't quite right since you could upload a webextension for desktop and a non-webextension for Android. However, that's a really unrealistic use case and simply highlights how our model of multiple install files is broken. Thus, I think you should keep it as is. I just wanted to mention it. I think switching to `files.all(...)` would make my example case worse from a usability standpoint so don't do that.

kumar303: The old site did `if current_version.files[0].is_webextension == True` so definitely don't to that 😆

willdurand:
> Is the distance between the 3 labels in the 1st screenshot the same? It looks to me like the distance from the "Restart required" icon to the "Not..." is bigger?

Indeed, I did not see it before but I just fixed it:

screen shot 2017-11-06 at 22 07 48

src/core/reducers/addons.js
willdurand: Yes, I thought about this too but told myself "we assumed that for other computed props", let's be consistent..

willdurand:
> Just checked locally and it seems the check for isWebExtension is a bit too narrow as it's only valid for addonType === ADDON_TYPE_WEBEXTENSION. 

@tofumatt heh sorry for that, and thanks for the review. Are there any other rules to determine if an add-on is incompatible? I asked the question in the issue but did not get any answer.

tofumatt:
We do have a function in `utils.js` named `isCompatibleWithUserAgent` but it requires a UA. I think the logic in there is correct though... I wonder about just supplying it a known FF57 UA? But that's a bit naive and strange.

I believe that any add-on that is not compatible with 57+ had `is_strict_compatibility_enabled` set to `true` in the DB so the logic here for a hard-coded browser version of `57` should work? https://github.com/mozilla/addons-frontend/blob/master/src/core/utils/compatibility.js#L93

tests/unit/amo/components/TestAddonBadges.js
```diff
@@ -106,4 +106,24 @@ describe(__filename, () => {
 
     expect(root.find(Badge)).toHaveLength(0);
   });
+
+  it('displays a badge when the addon is not a web extension', () => {
```
tofumatt: There should be an extra test here for it not being a web extension but also not being an `addonType` of `ADDON_TYPE_EXTENSION`.

src/core/reducers/addons.js
kumar303: I don't think there are plans to improve the data model any time soon. Your approach is fine -- someone will file a bug for their bizarro edge case if not :)

grahamperrin:
> … It'll be nice to also have the icon **in listings** e.g. …

I'm in two minds about that. One mind: 

- discourage unnecessary clicks.

Second mind: 

- serendipitous discovery of '**More extensions by …**' the same developer

– and the more can be a **very** good thing, and 👍 it is already disclosed (without a _Show more_ click) in the primary page for an add-on. 

----

Unless I'm missing something (I'm speeding through screenshots, not code), we do not yet have: 

- the **no entry** icon (_Not compatible with Firefox Quantum_) in the **More extensions by …** area: 

2017-11-07 02 24 31 56 0 2 user-agent switcher sdk - add-ons for firefox

Image source: https://addons.mozilla.org/addon/user-agent-switcher-firefox/ viewed with 56.0.2 on FreeBSD-CURRENT.

tests/unit/amo/components/TestAddonBadges.js
willdurand: ah good catch, just added it!

willdurand:
@grahamperrin I think the list view needs more UX input and therefore might be updated later. This PR will close your issue, but I think you should open another issue to ask for a visual hint on the add-on lists.

willdurand:
Follow-up issue: #3814.

PR #3808, opened 2017-11-06
Updated discopane tests and setup server to be able to run in HTTPS m…

### LINKED ISSUE #3634: Updated and extend discopane UI tests
The current discopane UI tests are good but we could use some more.

I am using this issue to track my work.

### LINKED ISSUE #3910: Discopane UI tests fail on Travis against Firefox 57
When we were using `firefox: latest` in `.travis.yml`, the Discopane UI tests started failing with what looks like maybe Selenium driver incompatibilities. Example: https://travis-ci.org/mozilla/addons-frontend/jobs/301948712

As a workaround we are pinning the tests to Firefox 56: https://github.com/mozilla/addons-frontend/pull/3909

```
_______________ ERROR at teardown of test_see_more_addons_button _______________

request = <SubRequest 'driver' for <Function 'test_see_more_addons_button'>>

driver_class = <class 'selenium.webdriver.firefox.webdriver.WebDriver'>

driver_kwargs = {'capabilities': {'marionette': True, 'moz:firefoxOptions': {'profile': 'UEsDBBQAAAAIADtvbks36sPN5wMAAGcNAAAHAAAAdXNlc...AAAABAAEANQAAAAwEAAAAAA=='}}, 'firefox_options': <selenium.webdriver.firefox.options.Options object at 0x7f3ad1a71050>}

    @pytest.yield_fixture

    def driver(request, driver_class, driver_kwargs):

        """Returns a WebDriver instance based on options and capabilities"""

        driver = driver_class(**driver_kwargs)

    

        event_listener = request.config.getoption('event_listener')

        if event_listener is not None:

            # Import the specified event listener and wrap the driver instance

            mod_name, class_name = event_listener.rsplit('.', 1)

            mod = __import__(mod_name, fromlist=[class_name])

            event_listener = getattr(mod, class_name)

            if not isinstance(driver, EventFiringWebDriver):

                driver = EventFiringWebDriver(driver, event_listener())

    

        request.node._driver = driver

        yield driver

>       driver.quit()

.tox/discopane-ui-tests/local/lib/python2.7/site-packages/pytest_selenium/pytest_selenium.py:105: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <selenium.webdriver.firefox.webdriver.WebDriver (session="e5621908-6c24-4672-bf92-9534a05a2d33")>

    def quit(self):

        """Quits the driver and close every associated window."""

        try:

            RemoteWebDriver.quit(self)

        except (http_client.BadStatusLine, socket.error):

            # Happens if Firefox shutsdown before we've read the response from

            # the socket.

            pass

        if "specificationLevel" in self.capabilities:

            self.service.stop()

        else:

>           self.binary.kill()

E           AttributeError: 'NoneType' object has no attribute 'kill'

.tox/discopane-ui-tests/local/lib/python2.7/site-packages/selenium/webdriver/firefox/webdriver.py:187: AttributeError

------------------------------- pytest-selenium --------------------------------
```

jrbenny35:
Fixes #3634 
Fixes #3910

This updates the current discopane tests to newer standards and adds addon and theme install tests.

Thanks for opening a Pull Request (PR), here's a few guidelines as to what we need in your PR before we review it.

I've added and changed some JS code, and my JS is not that good so please take a close look as I'm sure it will need correcting.


jrbenny35:
@kumar303 @tofumatt Can any of you guys take a look at this?

.travis.yml
```diff
@@ -1,40 +1,98 @@
 sudo: required
+_aliases:
+  - &ui-tests
+    addons:
+      firefox: latest-nightly
+      hosts: example.com
```
kumar303: Can you also update [the docs](https://github.com/mozilla/addons-frontend/blob/master/tests/ui/README.md) for how to do this manually? I assume there is an extra step now where you have to add `example.com 127.0.0.1` to `/etc/hosts`, right?

.travis.yml
```diff
@@ -1,40 +1,98 @@
...
+      - wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER/geckodriver-v$GECKODRIVER-linux64.tar.gz
+      - mkdir $HOME/geckodriver && tar xvf /tmp/geckodriver.tar.gz -C $HOME/geckodriver
+      - export PATH=$HOME/geckodriver:$PATH
+      - firefox --version
+      - geckodriver --version
+      - pip install tox
+    before_script:
+      - yarn install
+      - sh -e /etc/init.d/xvfb start
+      - sleep 5
```
kumar303: the old code did `sleep 3` -- could we try going back to that or were you running into issues? The tests need to be as fast as possible.

.travis.yml
```diff
@@ -1,40 +1,98 @@
...
+      - jwm &
+      - sleep 5
+      - yarn start-func-test-server
+    script: tox
+  - &node6
+    addons:
+      firefox: latest
```
kumar303: why does this job need firefox?

.travis.yml
```diff
@@ -1,40 +1,98 @@
...
+    language: node_js
+    node_js: '6'
+    before_install: npm install -g yarn
+  - &node8
+    addons:
+      firefox: latest
```
kumar303: I don't think this one needs firefox either

.travis.yml
```diff
@@ -1,40 +1,98 @@
...
+    - stage:
+      language: python
+      python: 3.6
+      <<: *ui-tests
+      install: nvm install 6
+    - stage:
+      language: python
+      python: 3.6
+      <<: *ui-tests
+      install: nvm install 8
```
kumar303: I don't like how there are always two duplicate stages, one for each version of Node. The old approach to this yaml file using `matrix` did not need that. Is there a way to switch back to `matrix` but still use aliases to share configuration?

tests/ui/conftest.py
```diff
@@ -1,8 +1,52 @@
...
+    capabilities['acceptInsecureCerts'] = True
     return capabilities
+
+
+@pytest.fixture
+def my_base_url(base_url, firefox):
+    """Create custom base url depending on os and firefox version."""
+    ff_version = firefox.browser.firefox_version
+    os = sys.platform
+    return base_url + '/discovery/pane/{0}.0/{1}/normal'.format(ff_version, os)
```
kumar303: This can just use string formatting:

```python
return '{}/discovery/pane/{}.0/{}/normal'.format(base_url, ff_version, os)
```

tests/ui/conftest.py
```diff
@@ -1,8 +1,52 @@
...
+@pytest.fixture
+def discovery_pane(selenium, my_base_url):
+    """Open the discovery pane via the URL."""
+    return DiscoveryPane(selenium, my_base_url).open()
+
+
+@pytest.fixture
+def firefox_options(firefox_options):
+    """Configure Firefox preferences and additonal arguments."""
+    firefox_options.set_preference('extensions.webservice.discoverURL', 'https://discovery.addons.mozilla.org/%LOCALE%/firefox/discovery/pane/%VERSION%/%OS%/%COMPATIBILITY_MODE%')  # noqa
```
kumar303: Why isn't this URL set to `https://example.com:4000`? I'm confused

.travis.yml
jrbenny35: Well its the same, it just look different. The jobs config needs to be more explicitly written whereas the matrix would create all of it.

src/core/server/base.js
```diff
@@ -396,9 +399,18 @@ export function runServer({
         const routes = require(`${appName}/routes`).default;
         const createStore = require(`${appName}/store`).default;
         /* eslint-enable global-require, import/no-dynamic-require */
-        const server = baseServer(
+        let server = baseServer(
           routes, createStore, { appInstanceName: appName });
         if (listen === true) {
+          if (useHttps) {
```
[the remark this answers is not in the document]
jrbenny35: From what I know, yes.

tests/ui/conftest.py
jrbenny35: This is to test that if the URL is wrong, discopane will not load. Kind of an edge case but I thought it was useful.

tox.ini
```diff
@@ -4,30 +4,15 @@ envlist = discopane-ui-tests, flake8
...
-  pytest-metadata==1.2.0
-  pytest-selenium==1.9.0
-  pytest_variables==1.4
-  requests==2.13.0
-  selenium==3.0.2
-  six==1.10.0
-  zope.component==4.3.0
-  zope.event==4.2.0
-  zope.interface==4.3.0
-install_command = pip install --no-deps {opts} {packages}
```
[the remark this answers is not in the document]
jrbenny35: No reason

src/core/server/base.js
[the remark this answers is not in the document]
jrbenny35: Could I do a ```&&``` using ```useHttps``` and that check you posted?

tests/ui/conftest.py
jrbenny35: Actually, my mistake. It appears that creating a custom profile with selenium does not populate this field. Maybe this isn't needed however.

jrbenny35:
So I think I assumed that the yarn tests would need firefox, that is why I changed over to the jobs style config. If you really don't like it I can go back to the matrix style as before.

bin/start-func-test-server.js
```diff
@@ -121,6 +121,8 @@ new Promise((resolve) => {
       `NODE_APP_INSTANCE=${appInstance}`,
       '-e',
       `NODE_ENV=${nodeEnv}`,
+      '-e',
+      `USE_HTTPS=true`,
```
kumar303: I think you should use `-e` here to hard code everything to `https://example.com:4000`. You can follow [this example](https://github.com/mozilla/addons-frontend/#configuring-an-android-device-for-local-development) but you won't need to set `WEBPACK_SERVER_HOST`.

jrbenny35:
@kumar303 r?

jrbenny35:
@kumar303 can you look at this?

kumar303:
I have two other review requests in my queue ahead of this one but I'm making time to clear my queue.

.travis.yml
```diff
@@ -1,42 +1,94 @@
 sudo: required
+_aliases:
+  - &ui-tests
```
kumar303: I get that this is a manual declaration of what we had in the matrix but it adds a lot of noise which will make it really hard to maintain. I suggest some simplifications:

- Use `node-prod` and `node-next` so we can update the version numbers easily
- Use more aliases to clean up duplicate code
- Preserve the old order of stage execution
- Remove the old `$TASK` approach which was only needed for the matrix
- Only run crucial jobs in `node-prod` and `node-next`, not all jobs. Crucial jobs are the build step and CI tests.

Here's my attempt at it. I haven't tried a build with this file though:

<details>
<summary>Show contents

`.travis.yml`

</summary>

```yaml
sudo: required
_aliases:
  - &tox
    language: python
    python: 3.6
    install: pip install tox
    script: tox
  - &node-prod
    # This is the NodeJS version we run in production.
    language: node_js
    node_js: '6'
    before_install: npm install -g yarn
  - &node-next
    # This is the next NodeJS version we will support.
    language: node_js
    node_js: '8'
    before_install: npm install -g yarn
services:
  - docker
cache:
  yarn: true
  directories:
  - node_modules
jobs:
  include:

    # Test the build process.
    - stage:
      <<: *node-prod
      env: ADDONS_FRONTEND_BUILD_ALL=1
      script: yarn build
    - stage:
      <<: *node-next
      env: ADDONS_FRONTEND_BUILD_ALL=1
      script: yarn build

    # Run the unit/integration tests.
    - stage:
      <<: *node-prod
      script: yarn test-ci
    - stage:
      <<: *node-next
      script: yarn test-ci

    - stage:
      <<: *node-prod
      script: yarn lint

    - stage:
      <<: *node-prod
      script: yarn nsp-check

    # Run the functional tests.
    - stage:
      <<: *tox
      <<: *node-prod
      addons:
        firefox: latest-nightly
        hosts: example.com
      env: TOXENV=discopane-ui-tests DISPLAY=:99.0 GECKODRIVER=0.19.1 MOZ_HEADLESS=1 PATH=./node_modules/.bin:$PATH
      install:
        - yarn
        - yarn start-func-test-server &
      before_script:
        - sudo apt-get -qq update
        - sudo apt-get install jwm
        - wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER/geckodriver-v$GECKODRIVER-linux64.tar.gz
        - mkdir $HOME/geckodriver && tar xvf /tmp/geckodriver.tar.gz -C $HOME/geckodriver
        - export PATH=$HOME/geckodriver:$PATH
        - firefox --version
        - geckodriver --version
        - pip install tox
        - jwm &
        - sleep 3
        # Wait for server to be available
        - until $(curl --output /dev/null --silent --head --fail -k https://example.com:4000); do printf '.'; sleep 5; done

    - stage:
      <<: *tox
      env: TOXENV=flake8

    - stage:
      <<: *tox
      env: TOXENV=dennis-lint
```

</details>

src/core/server/base.js
```diff
@@ -378,6 +379,8 @@ export function runServer({
   const host = config.get('serverHost');
   const appName = config.get('appName');
 
+  const useHttps = process.env.USE_HTTPS;
```
kumar303: We should make it more obvious that this is for development only:

```js
const useHttpsForDev = process.env.USE_HTTPS_FOR_DEV;
```

tests/ui/README.md
```diff
@@ -14,22 +14,22 @@ If you think you would like to contribute to the tests by writing or maintaining
 it would be a good idea to create a fork of this repository first, and then clone that.
 GitHub also has great instructions for [forking a repository][git-fork].
 
+### Edit your hosts file
+These tests rely on the url `https://example.com`. You must add a hostname alias to your /etc/hosts
```
kumar303: Thanks! These steps worked for me.

src/core/server/base.js
jrbenny35: Good point.

jrbenny35:
@kumar303 do you guys care if the tests run just against node v6? Or should I have 2 env's for both v6 and v8

bin/local-dev-server-certs/README.md
```diff
@@ -0,0 +1,4 @@
+# Local server SSL certs
+
+These certs are used to enable SSL for local testing of discopane. They could probably be used elsewhere
```
kumar303: This should mention how the certs are specifically for `example.com` because it's not obvious from the encoded cert data.

src/core/server/base.js
```diff
@@ -394,9 +397,21 @@ export function runServer({
...
+          if (useHttpsForDev && (host === 'example.com')) {
+            const options = {
+              key: fs.readFileSync('bin/local-dev-server-certs/my-server.key.pem'),
+              cert: fs.readFileSync('bin/local-dev-server-certs/my-server.crt.pem'),
+              ca: fs.readFileSync('bin/local-dev-server-certs/my-private-root-ca.crt.pem'),
+              passphrase: '',
+            };
+            server = https.createServer(options, server);
+          } else {
+            log.info(
```
kumar303: This is in the wrong place. The statement up above should look more like:

```js
if (useHttpsForDev) {
  if (host === 'example.com') {
    // server = ...
  } else {
    log.info(`To use the HTTPS server you must...`);
  }
}
```

jrbenny35: Any reason for this? I feel having 2 logical statements that don't execute anything differently isn't needed.

kumar303:
If you merge in master, you'll get the fix for https://github.com/mozilla/addons-frontend/pull/4078 (this was causing tests in your branch to fail)

src/core/server/base.js
kumar303: The log message is currently in the wrong place. For example, if `process.env.USE_HTTPS_FOR_DEV` was not set at all (or it was false) then you'd see a message in the console telling you that you are configuring HTTPS wrong. Logging a message in this case would be unnecessary and would be confusing.

jrbenny35: Wait, if we follow your suggestion, wouldn't that need an error message too?

Could we just do an ```else if``` and throw 2 different logs? If ```process.env.USE_HTTPS_FOR_DEV``` is ```False``` then we can just say ```https is not enabled```, then if the host is also not ```example.com``` then we can use the current log message.

jrbenny35: @kumar303 thoughts?

src/core/server/base.js
```diff
@@ -403,17 +403,19 @@ export function runServer({
...
-              passphrase: '',
-            };
-            server = https.createServer(options, server);
-          } else {
-            log.info(
-              `To use the HTTPS server you must serve the site at example.com (host was "${host}")`);
+          if (useHttpsForDev) {
+            if (host === 'example.com') {
+              const options = {
+                key: fs.readFileSync('bin/local-dev-server-certs/my-server.key.pem'),
```
willdurand: Instead of `my-server`, I'd rename the keys with the host name like `example.com.key.pem` and same for the `crt`, because it is a bit confusing (or name the directory differently maybe).

kumar303: Good catch -- this change would be helpful

kumar303:
If you merge in master, you'll get the locale fix. See https://github.com/mozilla/addons-frontend/pull/4096

jrbenny35:
Wooot 😄

PR #3821, opened 2017-11-07
Translate add-on description header

### LINKED ISSUE #2705: "About this <add-on type>" needs l10n for add-on types, and for some locales, a localizable "this"
When viewing an add-on’s or theme’s details, "About this `<type>"` is displayed, probably taken from "About this %(addonType)s". This can be several items, such as add-on, theme, persona or (formerly?) extension. However these items display in English and are not available in the l10n files, so e.g. "Over deze extension" is displayed.

1) Please make sure all available items are available for l10n.
2) The "this" part may depend on the type for some locales, i.e. be gender related. Please make this possible. If needed, it may be best and easier to get rid of the type variable type and create separate strings, as was the case for "About this add-on"and "About this collection".

willdurand:
Fix #2705

---

This PR translates the add-on description header.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -255,23 +255,41 @@ export class AddonBase extends React.Component {
 
   renderShowMoreCard() {
     const { addon, i18n } = this.props;
-    const addonType = addon ? addon.type : ADDON_TYPE_EXTENSION;
 
+    let title;
     const descriptionProps = {};
+
     if (addon) {
-      const description =
-        addon.description ? addon.description : addon.summary;
+      switch (addon.type) {
+        case ADDON_TYPE_DICT:
+          title = i18n.gettext('About this dictionary');
+          break;
+        case ADDON_TYPE_EXTENSION:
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -255,23 +255,41 @@ export class AddonBase extends React.Component {
...
+        case ADDON_TYPE_DICT:
+          title = i18n.gettext('About this dictionary');
+          break;
+        case ADDON_TYPE_EXTENSION:
+          title = i18n.gettext('About this extension');
+          break;
+        case ADDON_TYPE_LANG:
+          title = i18n.gettext('About this language pack');
+          break;
+        case ADDON_TYPE_THEME:
```
tofumatt: We should have another one for `OPENSEARCH`

willdurand: What the name of an `opensearch`?

willdurand: See: https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/RatingManager/index.js#L137.

PR #3841, opened 2017-11-08
Make sure to set the right context in search

### LINKED ISSUE #3689: Search box should work the same from all sections of the site

willdurand:
Fix #3689 (comment)

---

This PR makes sure we set the right `viewContext` during a search.

### DIFF
--- a/tests/unit/amo/components/TestSearch.js
+++ b/tests/unit/amo/components/TestSearch.js
@@ -32,6 +32,7 @@ describe(__filename, () => {
 
   beforeEach(() => {
     props = {
+      context: VIEW_CONTEXT_EXPLORE,
       count: 80,
       dispatch: sinon.stub(),
       errorHandler: createStubErrorHandler(),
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestSearch.js
```diff
@@ -32,6 +32,7 @@ describe(__filename, () => {
 
   beforeEach(() => {
     props = {
+      context: VIEW_CONTEXT_EXPLORE,
```
kumar303: You probably know this but we should always dispatch actions for property values that are set by `mapStateToProps`. The older pattern of setting mapped prop values manually moves us away from reality since it only simulates how the application changes state.

It's not worth a follow-up patch; I know this was an older test that did not use `shallowUntilTarget()` and fixing it would be more effort. For future situations like this, I'd suggest trying to fix old test suites as part of the patch as long as it's not too time consuming (some are harder to convert than others).

Other than this, the patch looks good. Sorry I didn't get a chance to review it in time.

willdurand: Thanks 👍

PR #3844, opened 2017-11-08
Make sure i18n.moment uses the correct locale

### LINKED ISSUE #3538: The active locale leaks between server responses

I haven't been able to prove this but by reading the code, I think this is possible:

* Go to a page that displays a date or time
* Begin a request from one browser in the `en-US` locale
* Begin a request simultaneously from another browser in the `de` locale

### What happened?

Date/time data on the page may be rendered with the wrong locale

Date/time data should always be rendered in the requested locale

The problem is that we use a `moment.locale(locale)` to set the local globally and then we pass the global instance to all components. Due to the asynchronous nature of the program (especially with saga processing), we cannot always rely on this global instance to be the right one.

The fix is to pass an instance of `moment` which will always be configured for the right locale:

```diff
diff --git a/src/core/i18n/utils.js b/src/core/i18n/utils.js
index 7260b3e1..f525c2ba 100644
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -289,6 +289,6 @@ export function makeI18n(
 
   // We add a translated "moment" property to our `i18n` object
   // to make translated date/time/etc. easy.
-  i18n.moment = moment;
+  i18n.moment = moment();
   return i18n;
 }
```

However, this change breaks everything because our code is using the global localization functions. We have to replace code like this...

```js
const timestamp = i18n.moment(dateCreated).fromNow();
//     TypeError: i18n.moment is not a function
//         at AddonReviewListItemBase.render (/Users/kumar/dev/addons-frontend/src/amo/components/AddonReviewListItem/index.js:194:30)
//         ...
//         at Object.renderToString (/Users/kumar/dev/addons-frontend/node_modules/react-dom/lib/ReactServerRendering.js:71:10)
//         at hydrateOnClient (/Users/kumar/dev/addons-frontend/src/core/server/base.js:106:14)
//         at /Users/kumar/dev/addons-frontend/src/core/server/base.js:320:20
//         at process._t
```

...with some alternative. It's not immediately clear to me from the docs how we can replace it. We may need to introduce a proxy function that allows you to get a new moment instance configured with the same locale.

### LINKED ISSUE #3819: "Last updated" section of a detail page changes to another language at page refresh
Similar to https://github.com/mozilla/addons-frontend/issues/3443 but with an add-on detail page.

STR:
- `curl -s https://addons.mozilla.org/en-US/firefox/addon/adblock-plus/` 10 times

Expected Result:
- The page is always the same, fully in english

Actual Result:
- The "Last Updated" section is sometimes in a different language. For instance I just saw `<dt class="AddonMoreInfo-last-updated-title" data-reactid="193">Last updated</dt><dd data-reactid="194">21 小時前 (2017年11月6日)</dd>`.

Happening on prod and stage at least.

willdurand:
Fix #3819
Fix #3538

---

This PR changes the `i18n.moment` property and now only exposes a function to return localized moment instances. It is needed to avoid random locales to be used by moment (because it is used globally). The added test case simulates what is describe in both issues fixed above. The `locale` function has been added for the test suite and works as we expect.

Note that we cannot use global/static `moment` methods anymore like:

- https://momentjs.com/docs/#/durations/
- https://momentjs.com/docs/#/utilities/

Absolutely all other methods are called on an instance of `moment` and therefore will continue to work as intended. I don't think we should rely on any of the other global/static `moment` methods anyway because we would issue similar issues again.

### DIFF
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -300,7 +300,5 @@ export function makeI18n(
     return scopedMoment;
   };
 
-  i18n.moment.locale = () => momentLocale;
-
   return i18n;
 }
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -278,19 +278,29 @@ export function makeI18n(
     i18n.options._momentDefineLocale();
   }
 
-  // This makes sure moment always uses the current locale.
-  moment.locale(makeMomentLocale(i18n.lang));
-
-  // Wrap the core Jed functionality so that we can always strip leading whitespace
-  // from translation keys to match the same process used in extraction.
+  // Wrap the core Jed functionality so that we can always strip leading
+  // whitespace from translation keys to match the same process used in
+  // extraction.
   i18n._dcnpgettext = i18n.dcnpgettext;
   i18n.dcnpgettext = function dcnpgettext(domain, context, singularKey, pluralKey, val) {
     return i18n._dcnpgettext(domain, context, oneLineTranslationString(singularKey),
       oneLineTranslationString(pluralKey), val);
   };
 
-  // We add a translated "moment" property to our `i18n` object
-  // to make translated date/time/etc. easy.
-  i18n.moment = moment;
+  const momentLocale = makeMomentLocale(i18n.lang);
+
+  // We add a translated "moment" property to our `i18n` object to make
+  // translated date/time/etc. easy.
+  i18n.moment = (...params) => {
+    const scopedMoment = moment(...params);
+
+    // This also makes sure moment always uses the current locale.
+    scopedMoment.locale(momentLocale);
+
+    return scopedMoment;
+  };
+
+  i18n.moment.locale = () => momentLocale;
+
   return i18n;
 }
--- a/tests/unit/core/i18n/test_utils.js
+++ b/tests/unit/core/i18n/test_utils.js
@@ -418,7 +418,7 @@ describe('i18n utils', () => {
         },
       };
       const i18n = utils.makeI18n(i18nData, 'fr', FakeJed);
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
       sinon.assert.called(i18nData.options._momentDefineLocale);
     });
 
@@ -431,7 +431,7 @@ describe('i18n utils', () => {
       };
 
       const i18n = utils.makeI18n(i18nData, 'en', FakeJed);
-      expect(i18n.moment.locale()).toEqual('en');
+      expect(i18n.moment().locale()).toEqual('en');
     });
 
     it('always passes the locale to moment', () => {
@@ -442,7 +442,7 @@ describe('i18n utils', () => {
         },
       };
       const i18n = utils.makeI18n(i18nData, 'fr', FakeJed);
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
     });
 
     it('formats a number', () => {
@@ -484,7 +484,7 @@ describe('i18n utils', () => {
       const i18n = utils.makeI18n({}, 'fr', FakeJed);
       // We modify the moment locale, globally.
       moment.locale('de');
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
     });
   });
 });
--- a/tests/unit/core/i18n/test_utils.js
+++ b/tests/unit/core/i18n/test_utils.js
@@ -482,7 +482,8 @@ describe('i18n utils', () => {
 
     it('always returns a scoped moment instance', () => {
       const i18n = utils.makeI18n({}, 'fr', FakeJed);
-      // We modify the moment locale, globally.
+      // Modifying the locale globally below does not affect the instance
+      // created previously.
       moment.locale('de');
       expect(i18n.moment().locale()).toEqual('fr');
     });
--- a/tests/unit/core/i18n/test_utils.js
+++ b/tests/unit/core/i18n/test_utils.js
@@ -487,5 +487,10 @@ describe('i18n utils', () => {
       moment.locale('de');
       expect(i18n.moment().locale()).toEqual('fr');
     });
+
+    it('formats a date', () => {
+      const i18n = utils.makeI18n({}, 'fr', FakeJed);
+      expect(i18n.moment('1988-09-22').format('ll')).toEqual('22 sept. 1988');
+    });
   });
 });
--- a/tests/unit/core/i18n/test_utils.js
+++ b/tests/unit/core/i18n/test_utils.js
@@ -398,15 +398,7 @@ describe('i18n utils', () => {
       }
     }
 
-    beforeEach(() => {
-      // FIXME: Our moment is not immutable so we reset it before each test.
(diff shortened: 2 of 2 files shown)


src/core/i18n/utils.js
```diff
@@ -278,19 +278,29 @@ export function makeI18n(
...
+  i18n.moment = (...params) => {
+    const scopedMoment = moment(...params);
+
+    // This also makes sure moment always uses the current locale.
+    scopedMoment.locale(momentLocale);
+
+    return scopedMoment;
+  };
+
+  i18n.moment.locale = () => momentLocale;
```
kumar303: Can't you just update the tests or am I missing something?

```diff
diff --git a/src/core/i18n/utils.js b/src/core/i18n/utils.js
index 5dabcb426..8c714a640 100644
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -300,7 +300,5 @@ export function makeI18n(
     return scopedMoment;
   };
 
-  i18n.moment.locale = () => momentLocale;
-
   return i18n;
 }
diff --git a/tests/unit/core/i18n/test_utils.js b/tests/unit/core/i18n/test_utils.js
index 156e83bb4..a92a3bf3c 100644
--- a/tests/unit/core/i18n/test_utils.js
+++ b/tests/unit/core/i18n/test_utils.js
@@ -418,7 +418,7 @@ describe('i18n utils', () => {
         },
       };
       const i18n = utils.makeI18n(i18nData, 'fr', FakeJed);
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
       sinon.assert.called(i18nData.options._momentDefineLocale);
     });
 
@@ -431,7 +431,7 @@ describe('i18n utils', () => {
       };
 
       const i18n = utils.makeI18n(i18nData, 'en', FakeJed);
-      expect(i18n.moment.locale()).toEqual('en');
+      expect(i18n.moment().locale()).toEqual('en');
     });
 
     it('always passes the locale to moment', () => {
@@ -442,7 +442,7 @@ describe('i18n utils', () => {
         },
       };
       const i18n = utils.makeI18n(i18nData, 'fr', FakeJed);
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
     });
 
     it('formats a number', () => {
@@ -484,7 +484,7 @@ describe('i18n utils', () => {
       const i18n = utils.makeI18n({}, 'fr', FakeJed);
       // We modify the moment locale, globally.
       moment.locale('de');
-      expect(i18n.moment.locale()).toEqual('fr');
+      expect(i18n.moment().locale()).toEqual('fr');
     });
   });
 });
```

tests/unit/core/i18n/test_utils.js
```diff
@@ -487,5 +479,12 @@ describe('i18n utils', () => {
       sinon.assert.calledWith(toLocaleStringSpy, 'fr');
       sinon.assert.notCalled(numberFormatSpy);
     });
+
+    it('always returns a scoped moment instance', () => {
+      const i18n = utils.makeI18n({}, 'fr', FakeJed);
+      // We modify the moment locale, globally.
```
kumar303: This comment threw me off when I read the test because I wasn't sure why you were modifying the global instance. This comment would have helped me understand why:

```
// Modifying the locale globally does not affect the instance.
```

src/core/i18n/utils.js
```diff
@@ -278,19 +278,29 @@ export function makeI18n(
...
-  i18n.moment = moment;
+  const momentLocale = makeMomentLocale(i18n.lang);
+
+  // We add a translated "moment" property to our `i18n` object to make
+  // translated date/time/etc. easy.
+  i18n.moment = (...params) => {
+    const scopedMoment = moment(...params);
+
+    // This also makes sure moment always uses the current locale.
+    scopedMoment.locale(momentLocale);
```
kumar303: I'm slightly nervous about calling `.locale()` every single time we need to format a string. I took a very quick look and it seems to [use caching](https://github.com/moment/moment/blob/6beb383a0ab0d4c8b3bb56d1393d810db79d486e/src/lib/locale/locales.js#L60) but I didn't inspect everything.

What about this? As far as I can tell [set()](https://momentjs.com/docs/#/get-set/set/) accomplishes the same thing.

```diff
diff --git a/src/core/i18n/utils.js b/src/core/i18n/utils.js
index 5dabcb426..521151508 100644
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -288,16 +288,12 @@ export function makeI18n(
   };
 
   const momentLocale = makeMomentLocale(i18n.lang);
+  const scopedMoment = moment().locale(momentLocale);
 
   // We add a translated "moment" property to our `i18n` object to make
   // translated date/time/etc. easy.
   i18n.moment = (...params) => {
-    const scopedMoment = moment(...params);
-
-    // This also makes sure moment always uses the current locale.
-    scopedMoment.locale(momentLocale);
-
-    return scopedMoment;
+    return scopedMoment.set(...params);
   };
 
   i18n.moment.locale = () => momentLocale;

```

willdurand: Are you sure about this `set()` method? From the doc, it is a...:

> Generic setter, accepting unit as first argument, and value as second

willdurand: I tried to find a constructor that is not `moment()` in the doc but could not find it 😢 I have seen various issues mentioning this approach of calling `locale()` right before `format()`, hence my patch. Of course, I understand your concern.

src/core/i18n/utils.js
willdurand: Ah well, if that works, I'd be very happy to use this. I must admit, I did not find anything in the doc and went for the easy solution, forgetting to try this before 😞

willdurand: Bwaaa I am stuuuupid. This is perfectly fine, thanks for this, this is much better!

tests/unit/core/i18n/test_utils.js
willdurand: ah sorry for that.

src/core/i18n/utils.js
willdurand: I just tried and I can confirm it does not work. Tests pass though because.. well, we dont have a test for that I'd say. Going to add one.

willdurand: Test case added here: https://github.com/mozilla/addons-frontend/pull/3844/commits/58245a8b0a9498976f575afde703854b76785897.

kumar303: Oh wow, good catch! Thanks for the test. I find `moment` so utterly confusing.

tests/unit/core/i18n/test_utils.js
```diff
@@ -487,5 +479,18 @@ describe('i18n utils', () => {
...
+      const i18n = utils.makeI18n({}, 'fr', FakeJed);
+      // Modifying the locale globally below does not affect the instance
+      // created previously.
+      moment.locale('de');
+      expect(i18n.moment().locale()).toEqual('fr');
+    });
+
+    it('formats a date', () => {
+      const i18n = utils.makeI18n({}, 'fr', FakeJed);
+      expect(i18n.moment('1988-09-22').format('ll')).toEqual('22 sept. 1988');
```
kumar303: blargh, can't believe we didn't have a test like that 👅

PR #3857, opened 2017-11-09
 adds comment fot IT at home

### LINKED ISSUE #3757: Please add a hidden string to the landing page for monitoring purpose
The old desktop site has a hidden string on the landing page for Ops to do site availability monitoring, for example https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/templates/addons/home.html#L33-L34 

### What happened?
The new mobile site doesn't have such a string. We're using "Firefox" for now.
Can you add a hidden string on the landing page so we can use to monitor the site availability? Ideally, please put a note that says "not remove it" so it won't get removed by accident.

We have no preference what the hidden string is, as long as it's unique and not too long. :)

tsl143:
Fixes #3757

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -171,6 +171,7 @@ export class HomeBase extends React.Component {
     return (
       <div className="Home">
         <span
+          className="visually-hidden do-not-remove"
           // eslint-disable-next-line react/no-danger
           dangerouslySetInnerHTML={{ __html: '<!-- Godzilla of browsers -->' }}
         />
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -170,7 +170,11 @@ export class HomeBase extends React.Component {
 
     return (
       <div className="Home">
-        <span data-comment="For-IT" dangerouslySetInnerHTML={{ __html: '<!-- Godzilla of browsers -->' }} />
+        <span
+          data-comment="For-IT"
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{ __html: '<!-- Godzilla of browsers -->' }}
+        />
 
         {errorHandler.renderErrorIfPresent()}
 
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -170,6 +170,9 @@ export class HomeBase extends React.Component {
 
     return (
       <div className="Home">
+        {/* IT looks for this comment with nagios, don't remove it. */}
+        <span dangerouslySetInnerHTML={{ __html: '<!-- aww yeah -->' }} />
+
         {errorHandler.renderErrorIfPresent()}
 
         <HomeHeroBanner />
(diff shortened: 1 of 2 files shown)


src/amo/components/Home/index.js
```diff
@@ -170,6 +170,9 @@ export class HomeBase extends React.Component {
 
     return (
       <div className="Home">
+        {/* IT looks for this comment with nagios, don't remove it. */}
+        <span dangerouslySetInnerHTML={{ __html: '<!-- aww yeah -->' }} />
```
tofumatt: So this should probably just be an HTML comment–won't that work? Or will react strip it out?

Also I wish we used a more fun string. This is what IT look for on addons-server so maybe ours should be different.

Maybe we could use "Godzilla of browsers" or something cute 😉

willdurand: > Or will react strip it out?

I am pretty sure React removes all comments.

tofumatt: Ah okay I figured that might be the case. Cool then!

tsl143: Yes react removes all the comments.
And yes, writing test case is much better idea.
`Godzilla of browsers` \o/

tests/unit/amo/containers/TestHome.js
```diff
@@ -168,6 +168,11 @@ describe(__filename, () => {
     });
   });
 
+  it('renders a span with comment for IT', () => {
+    const root = render();
+    expect(root.find('[data-comment="For-IT"]')).toHaveLength(1);
```
willdurand: I'd get rid of the `data-` attribute and I'd try to assert `root.html()` to be sure to find the expected HTML comment.

tsl143: @willdurand , I got the `root.html()` test but I am not sure what else to use if not `data-comment`

willdurand: nothing. why do you need this?

tsl143: How to search for element with specific text, either to put class or a data-attribute?

willdurand: You don't. Can't you use `toContain()`?

tsl143: I tried to use this
`  it('renders a span with comment for IT', () => {`
`   const root = render();`
`    expect(root.find('.Home').html()).toContain('Godzilla of browsers');`
`  });`
which should work but it gives me following error.
screen shot 2017-11-10 at 8 20 55 am

tests/unit/amo/containers/TestHome.js
```diff
@@ -168,6 +168,11 @@ describe(__filename, () => {
     });
   });
 
+  it('renders a span with comment for IT', () => {
+    const root = render();
+    expect(root.find({ dangerouslySetInnerHTML: { __html: '<!-- Godzilla of browsers -->' } })).toHaveLength(1);
```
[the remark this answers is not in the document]
tsl143: done

willdurand:
@tsl143 I have pushed a minor tweak to your branch as I was testing it locally :)

willdurand:
Thanks @tsl143!

Issue #3736, opened 2017-11-01
"Flag" turns into "Flagged" after Edit a review/Developer reply action


ValentinaPC:
Steps to reproduce:
1. Leave a comment to an add-on i.e. https://addons.allizom.org/en-US/firefox/addon/red-cushion/reviews/
2. Edit your review
3. Observe "Flag" link

Expected results:
"Flag" link does not change.

Actual results:
"Flag" link turns to "Flagged".

Notes/Issues:
- reproducing also for developer reply action
Verified on AMO-stage FF56(Win 7). Issue is also reproducing on dev.
Video for this issue:
![flag](https://user-images.githubusercontent.com/15685960/32275544-4ed884c4-bf14-11e7-9073-52b7eee37902.gif)

ValentinaPC:
Verified this on AMO-dev FF56(Win7)
Seems like the issue is still happening.
Please see video:
![flagged](https://user-images.githubusercontent.com/15685960/32652434-eee392c0-c60c-11e7-8006-71b47ea1c819.gif)

kumar303:
We had an issue that broke auto-updates to dev -- sorry about that. The patch should be on dev now. I added more details here: https://github.com/mozilla/addons-frontend/issues/2327#issuecomment-343600156

ValentinaPC:
Verified this on AMO-dev FF56(Win 7).
Flag is not changed to Flagged while edit a review or submitting a developer reply.
Postfix video:
![flag](https://user-images.githubusercontent.com/15685960/32727210-668752f2-c884-11e7-9f82-365e2c6836b1.gif)

PR #3864, opened 2017-11-09
Replying to review no longer marks review as flagged

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3736


willdurand:
Is it time for adding `shouldComponentUpdate()` methods?

kumar303:
> Is it time for implementing the `shouldComponentUpdate()` method?

I don't think so. Redux already does a little bit of that with a shallow comparison of what `mapStateToProps` outputs. In this case, I just had the state mapping wrong.

Issue #3867, opened 2017-11-09
Use a JS API to install addons to side-step connection errors caused by MITM SSL certs


muffinresearch:
Refs: https://github.com/mozilla/addons/issues/525
### Describe the problem and steps to reproduce it:

* Install a custom root CA (instructions below)
* Go to any add-on detail page and click the install button

### What happened?

The add-on fails to install with a window showing this message:

"The add-on failed to download due to a connection failure"

### What did you expect to happen?

If you have set up Firefox to use a custom root CA then it should allow add-on installations too.

### Anything else we should know?

Further analysis of the differences between the new AMO install buttons and the old shows that the old site uses a JS install API [InstallTrigger](https://github.com/mozilla/addons-server/blob/master/static/js/zamboni/buttons.js#L310) and the use of this API doesn't prevent installs from working when a MITM cert is installed.

## Reproducing the install connection failure issue

Using [Charles proxy](https://www.charlesproxy.com/) with the MITM cert installed and SSL proxying  set up to work for addons.mozilla.org  _(Update you only need the CDN host - see screenshot below)_. I can repro the issue on https://addons.mozilla.org using the new frontend.

_Charles SSL Proxying Settings ([requires the Charles Cert to be installed in Firefox first](https://www.charlesproxy.com/documentation/using-charles/ssl-certificates/) I recommend using a different profile for this):_

<img width="584" alt="ssl_proxying_settings" src="https://user-images.githubusercontent.com/1514/32652267-9de8f8a8-c5fb-11e7-84ab-dc5fe27e637a.png">

_Screenshot of the install connection error:_

<img width="962" alt="fullscreen_09_11_2017__19_07" src="https://user-images.githubusercontent.com/1514/32624551-7cc14f38-c581-11e7-91c7-a21d44539447.png">

* With charles setup to proxy SSL for the CDN as per the instructions above.
* Visit any details page
* Click the install button
* You will see the download start and then it will be followed by the connection error (shown in the screenshot above).

## Fixing the issue

There's a couple of possible fixes.

###  Using InstallTrigger

From the console running 

```InstallTrigger.install({URL: "https://addons.mozilla.org/firefox/downloads/file/764482/ublock_origin-1.14.18-an+fx.xpi?src=dp-btn-primary"});``` with Charles enabled successfully starts the install.

If we switch to using this old API for the button then this should prevent the users of AntiVirus software experiencing the issue.

### Using mozAddonManager

[...]

Issue #3871, opened 2017-11-10
Expose file objects mapped by platform in AddonType


kumar303:
The hotfix for https://github.com/mozilla/addons-frontend/issues/3867 requires access to file hashes by platform but `AddonType` doesn't expose that very easily. This issue is to clean up the hacks that were needed to get things working.

kumar303:
Duplicate of https://github.com/mozilla/addons-frontend/issues/4426

PR #3872, opened 2017-11-10
Use InstallTrigger.install() for extensions

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3867

* This switches us to `InstallTrigger`, like the old AMO, which doesn't care if the TLS channel has a non built-in root CA. This is because it relies on the file hash. If a file hash were `undefined` *and* the redirect to the final file did not contain a special hash header then it would fall back to requiring a built-in root CA. 
* Getting a file hash is not too easy because we need to find it based on platform. We do have a file URL and URLs are unique enough so I'm using that to find the hash as a quick fix. The follow-up: https://github.com/mozilla/addons-frontend/issues/3871

### DIFF
--- a/src/core/components/InstallButton/index.js
+++ b/src/core/components/InstallButton/index.js
@@ -1,14 +1,17 @@
-/* global window */
+/* global InstallTrigger, window */
 import classNames from 'classnames';
+import { oneLine } from 'common-tags';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 
+import { getAddonIconUrl } from 'core/imageUtils';
 import InstallSwitch from 'core/components/InstallSwitch';
 import {
   ADDON_TYPE_OPENSEARCH,
   ADDON_TYPE_THEME,
+  INSTALL_CATEGORY,
   INSTALL_STARTED_CATEGORY,
   TRACKING_TYPE_EXTENSION,
   validAddonTypes,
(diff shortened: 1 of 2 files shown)


muffinresearch:
See the latest comments in the original issue - looks like the hash is available via headers so you may not need to pass it explicitly.

muffinresearch:
I tried this running against prod  

First as a control running with charles proxying the prod CDN I get the connection error:
fullscreen_10_11_2017__11_15

Running the patch I get a successful install.

Note: The way I tested was using map local to point to js build from this branch for production. I had to turn off SRI in about:config and use a rewrite to add CORs because using map local in charles doesn't provide CORS headers.

I retested that turning off map local showed the same connection error each time.

src/core/components/InstallButton/index.js
```diff
@@ -74,6 +111,49 @@ export class InstallButtonBase extends React.Component {
...
+        // The old AMO did this so, hey, why not?
+        toString: () => installURL,
+      },
+    }, (xpiURL, status) => {
+      log.debug(oneLine`InstallTrigger completed for "${xpiURL}";
+        status=${status}`);
+
+      if (status === 0) {
+        // The extension was installed successfully.
+        this.trackInstallSucceeded({ addonName: addon.name });
```
muffinresearch: This is cool 👍

kumar303:
from @willdurand 

> kumar: I have tested your patch for InstallTrigger. It says something about it in the console logs locally. next step is to setup Charles with prod data

Thanks, all. Let's get this on dev...

PR #3874, opened 2017-11-10
Fix how we determine Quantum compatibility

### LINKED ISSUE #3868: Check that Mozilla signed extensions show up correctly
https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/ is a Mozilla signed extension. On the new site its showing up as "Not compatible with Firefox Quantum". On the old site it is showing up as a compatible.

Would consider this reasonably high priority (I just added that but feel free to disagree). Here's what mat said in IRC:

mat> the API says it is
mat> the frontend shouldn't use is_webextension or things like that, just rely on the compatibility info

willdurand:
Fix #3868

---

We use a forged User Agent string to be able to use the existing compatibility logic. It is not ideal but it is good-enough short-term. Th "not compatible with Firefox Quantum" badge should only be displayed if an add-on is not compatible with the desktop version of Quantum (`clientApp = firefox`).

## Screenshots

Before:

screen shot 2017-11-09 at 22 59 43

After:

screen shot 2017-11-09 at 23 00 32

**Important:** On Android, we do not show the badge on purpose because this extension is compatible with Quantum Desktop:

screen shot 2017-11-10 at 17 40 42

### DIFF
--- a/src/amo/components/AddonBadges/index.js
+++ b/src/amo/components/AddonBadges/index.js
@@ -35,7 +35,8 @@ export const AddonBadgesBase = (props: Props) => {
     }
   };
 
-  const isIncompatible = isQuantumCompatible({ addon }) === false;
+  const isIncompatible = addon.type === ADDON_TYPE_EXTENSION &&
+    isQuantumCompatible({ addon }) === false;
 
   return (
     <div className="AddonBadges">
--- a/src/amo/components/AddonBadges/index.js
+++ b/src/amo/components/AddonBadges/index.js
@@ -4,6 +4,7 @@ import { compose } from 'redux';
 
 import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
 import translate from 'core/i18n/translate';
+import { isQuantumCompatible } from 'core/utils/compatibility';
 import Badge from 'ui/components/Badge';
 import type { AddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
@@ -19,6 +20,10 @@ type Props = {|
 export const AddonBadgesBase = (props: Props) => {
   const { addon, i18n } = props;
 
+  if (!addon) {
+    return null;
+  }
+
   const getFeaturedText = (addonType: string) => {
     switch (addonType) {
       case ADDON_TYPE_EXTENSION:
@@ -30,24 +35,23 @@ export const AddonBadgesBase = (props: Props) => {
     }
   };
 
-  const isIncompatible = addon && addon.type === ADDON_TYPE_EXTENSION &&
-    !addon.isWebExtension;
+  const isIncompatible = isQuantumCompatible({ addon }) === false;
 
   return (
     <div className="AddonBadges">
-      {addon && addon.is_featured ? (
+      {addon.is_featured ? (
         <Badge
           type="featured"
           label={getFeaturedText(addon.type)}
         />
       ) : null}
-      {addon && addon.isRestartRequired ? (
+      {addon.isRestartRequired ? (
         <Badge
           type="restart-required"
           label={i18n.gettext('Restart Required')}
         />
       ) : null}
-      {addon && addon.is_experimental ? (
+      {addon.is_experimental ? (
         <Badge
           type="experimental"
           label={i18n.gettext('Experimental')}
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -1,10 +1,12 @@
 /* global window */
 import { oneLine } from 'common-tags';
 import mozCompare from 'mozilla-version-comparator';
+import UAParser from 'ua-parser-js';
 
 import {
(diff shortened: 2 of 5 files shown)


src/amo/components/AddonBadges/index.js
```diff
@@ -30,24 +35,23 @@ export const AddonBadgesBase = (props: Props) => {
     }
   };
 
-  const isIncompatible = addon && addon.type === ADDON_TYPE_EXTENSION &&
-    !addon.isWebExtension;
+  const isIncompatible = isQuantumCompatible({ addon }) === false;
```
kumar303: I think it would be safer to do this as:

```js
const isIncompatible = addon.type === ADDON_TYPE_EXTENSION &&
  isQuantumCompatible({ addon }) === false;
```

It's hurting my brain a little bit but I'm worried because [the compatibility object patch](https://github.com/mozilla/addons-frontend/commit/3cecbed3269f715099a6c8c69430bd9da6d6b2da) has not landed in prod yet which means when we cherry pick it, this function may be getting compatibility objects that cannot be trusted.

src/core/utils/compatibility.js
```diff
@@ -160,3 +162,17 @@ export function getClientCompatibility({
...
+export const isQuantumCompatible = ({ addon }) => {
+  // HACK: we use a hardcoded User Agent string corresponding to MacOS/Firefox
+  // 57.0 to determine whether the given `addon` is compatible with Quantum.
+  // The OS is not important as far as it is not `iOS`.
+  //
+  // See: https://github.com/mozilla/addons-frontend/issues/3868.
+  const userAgentInfo = UAParser(oneLine`Mozilla/5.0 (Macintosh; Intel Mac OS X
+    10.12; rv:57.0) Gecko/20100101 Firefox/57.0`);
+  // We set `clientApp` to `firefox` because we do not want to rely on it.
+  const clientApp = CLIENT_APP_FIREFOX;
```
kumar303: Actually, I think we do need to rely on `clientApp` because that's crucial to how the question "is this add-on compatible with 57" is answered. If an add-on is compatible with Firefox for Android 57 but is *not* compatible with Firefox for desktop 57 then we *do not* want to show the "incompatible with Quantum" banner.

tests/unit/core/utils/test_compatibility.js
```diff
@@ -610,4 +614,36 @@ describe(__filename, () => {
...
     });
   });
+
+  describe('isQuantumCompatible', () => {
+    it('returns `true` when add-on is compatible', () => {
+      const addon = createInternalAddon(createFakeAddon({
+        name: 'Firefox Multi-Account Containers',
+        compatibility: {
+          [CLIENT_APP_FIREFOX]: {
+            max: '*',
```
kumar303: You should set this to `57.0` to make it realistic

tests/unit/amo/components/TestAddonBadges.js
```diff
@@ -107,33 +118,34 @@ describe(__filename, () => {
...
-    }));
-    const root = shallowRender({ addon });
+      const root = shallowRender({ addon });
 
-    expect(root.find(Badge)).toHaveLength(0);
+      expect(root.find(Badge)).toHaveLength(0);
+    });
+
+    it('displays a badge when the addon is not compatible with Quantum', () => {
+      const addon = createInternalAddon(createFakeAddon());
```
kumar303: I don't think mocking out `isQuantumCompatible` is such a good idea for this case. Since we have no Flow coverage of the tests it's just too hard to keep the mock in sync with the real object. For starters, your tests do not make any assertions to guarantee that `isQuantumCompatible()` is called with the correct arguments. 

I think you should just set up the `addon` with realistic application data instead. This will naturally guarantee that `isQuantumCompatible()` gets called with the right arguments.

willdurand: That's what I did at the beginning but then I duplicated the same `addons` in the compatibility test file..

tests/unit/core/utils/test_compatibility.js
willdurand: It is realistic since it is the prod data..

willdurand: See: https://addons.mozilla.org/api/v3/addons/addon/multi-account-containers/

src/amo/components/AddonBadges/index.js
willdurand: but then we'll loose the possibility to test other add-on types?

willdurand: nevermind, let's play safe and do something better, later.

tests/unit/core/utils/test_compatibility.js
willdurand: There is a new test with `57.0` as `max`.

kumar303: Oh, cool. I didn't notice that prod uses `*`. I was looking at dev.

PR #3875, opened 2017-11-10
Add link to edit a collection

### LINKED ISSUE #3766: Add link to edit collection from collection view, when user has rights
There's currently no way to go from viewing a collection to editing it. The only way to edit collections right now is to go to View My Collections from the dropdown and then choose to edit one from the list. This doesn't work for collections you don't own, which affects our theme reviewers, who edit the Featured Themes collection.

So, if a user has permission to edit a collection, there should be a link or button in the collection view leading to the edit page for that collection.

willdurand:
Fix #3766

---

This PR adds a link to edit a collection when user is the owner or has the `Collections:Edit` permission.

It also fixes the special case when a user has a collection without any add-ons in it (it was not rendering the number of addons on the left and it the pagination was displayed on the right). Please see the first screenshots.

## Screenshots

Before:

screen shot 2017-11-10 at 16 59 17

After:

screen shot 2017-11-10 at 16 49 57
screen shot 2017-11-10 at 16 54 45

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -11,9 +11,11 @@ import {
   fetchCollectionPage,
 } from 'amo/reducers/collections';
 import NotFound from 'amo/components/ErrorPage/NotFound';
+import { COLLECTIONS_EDIT } from 'core/constants';
 import Paginate from 'core/components/Paginate';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import log from 'core/logger';
+import { hasPermission } from 'core/reducers/user';
 import translate from 'core/i18n/translate';
 import { parsePage } from 'core/utils';
 import Card from 'ui/components/Card';
@@ -24,6 +26,7 @@ import type {
   CollectionType,
 } from 'amo/reducers/collections';
 import type { ErrorHandlerType } from 'core/errorHandler';
+import type { UserStateType } from 'core/reducers/user';
 import type { ReactRouterLocation } from 'core/types/router';
 import type { I18nType } from 'core/types/i18n';
 
@@ -34,6 +37,7 @@ type Props = {|
   collection: CollectionType | null,
   dispatch: Function,
   errorHandler: ErrorHandlerType,
+  hasEditPermission: boolean,
   i18n: I18nType,
   loading: boolean,
   location: ReactRouterLocation,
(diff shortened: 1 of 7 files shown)


src/amo/components/Collection/index.js
```diff
@@ -197,10 +214,22 @@ export class CollectionBase extends React.Component<Props> {
...
+export const mapStateToProps = (
+  state: {| collections: CollectionsState, user: UserStateType |}
+) => {
+  const { current: collection, loading } = state.collections;
+
+  let hasEditPermission = false;
+
+  if (collection) {
+    hasEditPermission = collection.authorId === state.user.id ||
+      hasPermission(state, COLLECTIONS_EDIT);
```
kumar303: Should this be outside of the `if (collection)` block so that admins always get permission rights? But maybe if `collection` is falsy it doesn't matter anyway?

src/ui/components/MetadataCard/index.js
```diff
@@ -25,10 +25,13 @@ const MetadataCard = (props: Props) => {
           throw new Error('title is required');
         }
 
+        const hasContent = content || content === '' ||
```
kumar303: This was hard for me to understand. It would be nice if `metadata` was [typed properly](https://github.com/willdurand/addons-frontend/blob/a3ecc0d751b6bd5dc7e55c4096d9cb5c5ccb6430/src/ui/components/MetadataCard/index.js#L10).

kumar303: What about this?

```js
// <some comment here explaining '0' strings>
const hasContent = content || typeof content === 'string';
```

Using `parseInt()` to check for string numbers seems awkward.

kumar303: This is not a change request, just a comment :)

src/amo/components/Collection/index.js
willdurand: mmh if the collection is not loaded, there is no link. Maybe admins could get a link *before* but I am not sure that is matters a lot.

src/ui/components/MetadataCard/index.js
willdurand: It deserves a comment indeed because I do not check string numbers with `parseInt()`. Numbers are allowed but `0` is interpreted as `false` here..

willdurand: I simplified with a test `content === 0`, it is more obvious I think.

PR #3883, opened 2017-11-13
chore(package): update nsp to version 3.1.0

muffinresearch:
Closes #3809


willdurand:
I think this is not solved: https://github.com/mozilla/addons-frontend/pull/3809#issuecomment-342513473.

muffinresearch:
> I think this is not solved: #3809 (comment).

Yep, it was wishful thinking on my part :)

PR #3889, opened 2017-11-13
Fix pagination when over 25k add-ons

### LINKED ISSUE #3887: New AMO pages don't return correct number of pages after empty search
Steps to reproduce:
1. Make an empty search for themes -  https://addons-dev.allizom.org/en-US/firefox/search/?platform=windows&type=persona
2. Observe number of available pages - 14.006
3. Navigate to page 1000 - https://addons-dev.allizom.org/en-US/firefox/search/?page=1000&platform=windows&type=persona (which is also last working page from 14 006 pages)

Expected results:
There is a 25k add-ons returned by the API and the frontend should take this into account.

Actual results:
"No results were found" page is displayed. 

Notes/Issues:
- after page refresh a 404 page is displayed.
Video for this issue:
nav-themes

willdurand:
Fix #3887

---

This PR fixes the `Paginate` component to not display more pages than allowed by the API. Currently, the API can return up to 25k add-ons. If we go over this limit, we get a 404 (for example, with 25 add-ons per page, we can navigate up to page `1000` and page `1001` will 404).

The diff is larger than expected because I updated the test file.

## Screenshot

screen shot 2017-11-13 at 15 52 42

### DIFF
--- a/src/core/components/Paginate/index.js
+++ b/src/core/components/Paginate/index.js
@@ -7,6 +7,8 @@ import translate from 'core/i18n/translate';
 
 import './Paginate.scss';
 
+// This is the maximal number of add-ons the API can return for a query.
+const MAX_ADDONS = 25000;
 
 function makePageNumbers({ start, end }) {
   const pages = [];
@@ -25,7 +27,10 @@ export class PaginateBase extends React.Component {
     pathname: PropTypes.string.isRequired,
     perPage: PropTypes.number,
     queryParams: PropTypes.object,
-    showPages: PropTypes.number,
+    showPages: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.boolean,
+    ]),
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


willdurand:
See also: https://github.com/mozilla/addons-server/issues/6902

src/core/components/Paginate/index.js
```diff
@@ -7,6 +7,8 @@ import translate from 'core/i18n/translate';
 
 import './Paginate.scss';
 
+// This is the maximal number of add-ons the API can return for a query.
+const MAX_ADDONS = 25000;
```
kumar303: This limit looks like it's only applied to views that utilize Elastic Search (i.e. not views that utilize the database directly): https://github.com/mozilla/addons-server/search?utf8=%E2%9C%93&q=ES_MAX_RESULT_WINDOW&type=

Thus, we shouldn't simulate the limit for all paginated components. In fact, the paginate component may not be working with API data at all so I don't think this is the right place.

willdurand:
Asked for a new API attribute (`page_count`) here: https://github.com/mozilla/addons-server/issues/6919. Closing this PR for now.

PR #3891, opened 2017-11-13
Handle platform=all from old frontend

### LINKED ISSUE #3870: Search passes down "platform=all" parameter to the API if it was in the URL
When searching for add-ons in the new frontend, it filters by platform by default, and you can remove that filtering by selecting "Operating Systems: All" in the dropdown. That's good.

However, when the `platform=all` parameter is present in the URL, the dropdown looks the same *but* the `platform=all` parameter is passed to the API call. That results in fewer add-ons found, because instead of not filtering on platform at all, it filters add-ons that specifically have a file that is available for the special "all" platforms.

This matters because the old frontend does use that "platform=all" parameter. So, as a STR:
- Go to the old site
- Make a search
- In the filters, select Firefox 57.0 (if it doesn't let you, select 56.0 then change it in the URL) and extensions category
- Look at the results
- Click on view the new site
- Look ad the results

Expected:
- You should get the same results

Actual:
- You get a lot fewer add-ons.

willdurand:
Fix #3870

---

This PR triggers a server redirect when `platform=all` is passed in a search query so that we get the same results on both the old and new frontends. It is required because the new frontend expresses "all platforms"  with an empty string, resulting in the filter omitted. The old frontend uses the string `all` though, but if we do that then the API will look for add-ons with a platform/OS named `all` 😛

### DIFF
--- a/src/amo/components/SearchPage/index.js
+++ b/src/amo/components/SearchPage/index.js
@@ -12,6 +12,7 @@ import {
   ADDON_TYPE_THEME,
 } from 'core/constants';
 import { makeQueryString } from 'core/api';
+import log from 'core/logger';
 import { sendServerRedirect } from 'core/reducers/redirectTo';
 import {
   convertFiltersToQueryParams,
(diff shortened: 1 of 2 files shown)


src/amo/components/SearchPage/index.js
```diff
@@ -33,9 +34,25 @@ export class SearchPageBase extends React.Component<Props> {
   componentWillMount() {
     const { clientApp, filters, lang, location } = this.props;
 
+    let shouldRedirect = false;
+    const newFilters = { ...filters };
+
+    // The legacy frontend uses `all` to express "all platforms" but we now use
+    // no parameter (empty) to target all platforms. In addition, the query
+    // parameter `platform` is mapped to the `operatingSystem` filter.
+    // See: https://github.com/mozilla/addons-frontend/issues/3870.
```
kumar303: This comment is helpful, thanks

tests/unit/amo/components/TestSearchPage.js
```diff
@@ -165,6 +166,39 @@ describe(__filename, () => {
...
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+
+    render({ location: { query: { platform: OS_MAC } }, store });
+
+    sinon.assert.notCalled(fakeDispatch);
+  });
+
+  it('redirects without affecting the other parameters', () => {
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+    const query = { atype: 1, page: 123, platform: 'all' };
```
kumar303: Since `atype` has its own mapping logic, this would be a more neutral test without it. If the query were set to `{ page: 123, platform: 'all' }` then I think the test coverage would be just as good.

willdurand: Yeah I wanted to test both at once, `atype` (which has a dedicated logic) and `page` (which is yet another parameter). I've removed `atype`.

willdurand:
Merging as failing CI is not related to this patch and has been fixed in #3901.

PR #3894, opened 2017-11-13
Use a fixed error handler for reviews list

### LINKED ISSUE #3273: 404 initial render for out of range pagination has no content
The first render of a 404 page for **out of range pagination** shows a blank page:

ssr-404

## STR:
URL: https://addons.mozilla.org/en-US/firefox/addon/transparent-standalone-imag/reviews/?page=3

At time of  filing this is Page 3 of 2 pages.

## What happens: 

Looks like the right UI is shown when the JS kicks in which makes this a low priority fix.

##  What should happen

I'd expect a 404 to be the initial render. 
Note: Navigating to this URL shouldn't be possible by default so this is minor in terms of impact.

willdurand:
Fix #3273

---

This PR uses a "fixed" error handler to synchronize the server and client error handlers, avoiding to render a 200 and then a 404 when it is expected to get a 404. Now, the server returns a 404.

### DIFF
--- a/tests/unit/amo/components/TestAddonReviewList.js
+++ b/tests/unit/amo/components/TestAddonReviewList.js
@@ -7,6 +7,7 @@ import {
 import { setViewContext } from 'amo/actions/viewContext';
 import AddonReviewList, {
   AddonReviewListBase,
+  extractId,
 } from 'amo/components/AddonReviewList';
 import AddonReviewListItem from 'amo/components/AddonReviewListItem';
 import NotFound from 'amo/components/ErrorPage/NotFound';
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -45,19 +46,21 @@ describe(__filename, () => {
...
-      params,
+      params: {
+        addonSlug: fakeAddon.slug,
+      },
       store,
       ...customProps,
     };
+  };
+
+  const render = ({ ...customProps } = {}) => {
```
kumar303: It can just be `customProps = {}`

Issue #3895, opened 2017-11-13
Windows-only and Linux-only add-ons are flagged as incompatible with Quantum


kumar303:
### Describe the problem and steps to reproduce it:

* Create a windows-only or linux-only extension that supports Firefox 57
* Go to its listing page

### What happened?

It says *incompatible with Firefox Quantum*

### What did you expect to happen?

It should not include this badge.

### Anything else we should know?

See https://github.com/mozilla/addons-server/issues/6905 for details. Also see this thread: https://discourse.mozilla.org/t/extension-not-compatible-with-firefox-quantum/21460/7

This is a follow-up to https://github.com/mozilla/addons-frontend/issues/3868

ValentinaPC:
Verified as fixed on AMO-dev FF56(Win 7).
Only-Win and only-Linux compatible add-ons are not flagged as incomaptible with Quantum.
Examples:
- https://addons-dev.allizom.org/en-US/developers/addon/webext-linux/versions/1533230
![linux](https://user-images.githubusercontent.com/15685960/32773350-86572b9a-c931-11e7-9cca-ac43733f7370.png)

- https://addons-dev.allizom.org/en-US/firefox/addon/webext-win/
![win](https://user-images.githubusercontent.com/15685960/32773971-8349e0da-c933-11e7-99d2-1d2e8dae4fcf.png)

PR #3896, opened 2017-11-13
Quantum compat is true for all webext / moz addons

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3895

I tested with some dev data. The following should all show up as compatible with Quantum:

- https://addons-dev.allizom.org/en-US/firefox/addon/multi-account-containers/
- https://addons-dev.allizom.org/en-US/firefox/addon/web-ext-windows-only/
- https://addons-dev.allizom.org/en-US/firefox/addon/web-ext-linux-only/
- (any other web extension)

This should show up as incompatible with Quantum:

- https://addons-dev.allizom.org/en-US/firefox/addon/adblock-edge/


tests/unit/core/utils/test_compatibility.js
```diff
@@ -646,8 +720,20 @@ describe(__filename, () => {
...
+      }));
+
+      expect(isQuantumCompatible({ addon })).toEqual(false);
+    });
+
+    it('returns `true` when Android webextension is compatible', () => {
       const addon = createInternalAddon(createFakeAddon({
+        files: [{
+          is_webextension: true,
+          platform: OS_ANDROID,
```
kumar303: When I added this, the old code was failing because of the desktop UA string. Oops.

src/core/utils/compatibility.js
```diff
@@ -165,25 +162,16 @@ export function getClientCompatibility({
 }
 
 export const isQuantumCompatible = ({ addon }) => {
-  // HACK: we use a hardcoded User Agent string corresponding to MacOS/Firefox
-  // 57.0 to determine whether the given `addon` is compatible with Quantum.
-  // The OS is not important as far as it is not `iOS`.
```
willdurand: haha, who wrote this comment? 😓

tests/unit/core/utils/test_compatibility.js
willdurand: 😞

src/core/utils/compatibility.js
```diff
@@ -165,25 +162,16 @@ export function getClientCompatibility({
...
+  // https://github.com/mozilla/addons-frontend/issues/3814
+
+  if (!addon.current_version) {
+    return false;
+  }
+  return addon.current_version.files.some((file) => {
+    // These checks are fragile because future mozilla-signed extensions
+    // may not be Quantum compatible.
+    return file.is_webextension || file.is_mozilla_signed_extension;
+  });
```
willdurand: The `isWebExtension` computed prop was already available. How do you feel about computing `isMozillaSignedExtension` in the `reducer` then?a

PR #3897, opened 2017-11-13
Allow admins to reply to reviews

### LINKED ISSUE #3829: Admins cannot reply to reviews

* Sign in as an admin
* Go to the detail page for an add-on you are not an author of
* Click *Read N reviews*

### What happened?

You do not see a *reply* link to any review

You should be able to reply to reviews

The old site allowed admins to do this: https://github.com/mozilla/addons-server/blob/c0d8e97cb7251d0702c0b8b4c18a5e9d9127680a/src/olympia/reviews/templates/reviews/review.html#L74

willdurand:
Fix #3829

---

This PR allows admins to reply to any reviews, except their own reviews.

### DIFF
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -9,7 +9,10 @@ import FlagReviewMenu from 'amo/components/FlagReviewMenu';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
-import { isAuthenticated } from 'core/reducers/user';
+import {
+  isAdmin,
+  isAuthenticated,
+} from 'core/reducers/user';
 import { isAddonAuthor, nl2br, sanitizeHTML } from 'core/utils';
 import {
   hideEditReviewForm,
@@ -39,6 +42,7 @@ type Props = {|
   editingReview: boolean,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
+  isAdmin: boolean,
   isAuthenticated: boolean,
   isReplyToReviewId?: number,
   i18n: I18nType,
@@ -180,6 +184,7 @@ export class AddonReviewListItemBase extends React.Component<Props> {
       editingReview,
       errorHandler,
       i18n,
+      isAdmin: userIsAdmin,
       isAuthenticated: userIsAuthenticated,
       isReplyToReviewId,
       location,
@@ -264,7 +269,7 @@ export class AddonReviewListItemBase extends React.Component<Props> {
           {
             review && addon && siteUser &&
             !replyingToReview && !review.reply &&
-            isAddonAuthor({ addon, userId: siteUser.id }) &&
+            (isAddonAuthor({ addon, userId: siteUser.id }) || userIsAdmin) &&
             review.userId !== siteUser.id ?
               (
                 <a
@@ -311,6 +316,7 @@ export function mapStateToProps(
   return {
     editingReview,
     isAuthenticated: isAuthenticated(state),
+    isAdmin: isAdmin(state),
     replyingToReview,
     siteUser: state.user,
     submittingReply,
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -6,11 +6,12 @@ import { compose } from 'redux';
 
 import AddonReview from 'amo/components/AddonReview';
 import FlagReviewMenu from 'amo/components/FlagReviewMenu';
+import { ADDONS_EDIT } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
 import {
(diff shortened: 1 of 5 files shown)


src/amo/components/AddonReviewListItem/index.js
```diff
@@ -311,6 +316,7 @@ export function mapStateToProps(
   return {
     editingReview,
     isAuthenticated: isAuthenticated(state),
+    isAdmin: isAdmin(state),
```
kumar303: I would prefer something like `siteUserIsAdmin` to match the `siteUser` prop. Also, this clarifies which user it refers to (this component already deals with add-on owner users and review author users).

src/core/reducers/user.js
```diff
@@ -61,6 +61,10 @@ export const hasPermission = (
   return permissions.includes(permission);
 };
 
+export const isAdmin = (state: { user: UserStateType }): boolean => {
+  return hasPermission(state, ADMIN_SUPER_POWERS);
```
kumar303: What if this did...

```js
if (permissions.includes(ADMIN_SUPER_POWERS)) {
  return true;
}
```

...and then call `isAdmin(state)` from within `hasPermission()`? I guess you'd need a duplicate check on `if (!state.user.permissions)` so I'm not sure it's worth it.

willdurand:
@kumar303 so... this is not that permission? https://github.com/mozilla/addons-server/blob/b9590d7033287d799f91a1db44f34055f1de5441/src/olympia/constants/permissions.py#L9-L10

src/amo/components/AddonReviewListItem/index.js
willdurand: I also changed the prop for `isAuthenticated`.

src/core/reducers/user.js
willdurand: Yeah, I did not want to duplicate the check so I'd prefer keep this as is.

willdurand:
Turned out that you're right. This `*:*` is a very specific permission that is for admins (not a lot of people have it). `Admin:%` was something else and it seems we should never have to use it.

The old frontend was using `Admin:Edit` so I used it here too.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -310,7 +316,8 @@ export function mapStateToProps(
   }
   return {
     editingReview,
-    isAuthenticated: isAuthenticated(state),
+    siteUserIsAuthenticated: isAuthenticated(state),
+    siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
```
willdurand: Maybe `siteUserHasReplyPermission` would be better, but it soooo long..

kumar303: > Maybe `siteUserHasReplyPermission` would be better

I like *has...perm* the best https://i.pinimg.com/736x/e4/91/fa/e491fabe4365fb2b62e193601c44c4d7--hair-perms-historical-hairstyles.jpg

src/core/constants.js
```diff
@@ -233,7 +233,7 @@ export const ADDONS_REVIEW = 'Addons:Review';
 export const THEMES_REVIEW = 'Personas:Review';
 // Can view statistics for all addons, regardless of privacy settings.
 export const STATS_VIEW = 'Stats:View';
-// Admin super powers. Very few users will have this permission.
-export const ADMIN_SUPER_POWERS = 'Admin:%';
 // Can edit all collections.
 export const COLLECTIONS_EDIT = 'Collections:Edit';
+// Super powers. It means absolutely all permissions.
+export const ALL_SUPER_POWERS = '*:*';
```
kumar303: It's so heroic. I love it.

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -270,6 +271,52 @@ describe(__filename, () => {
...
+    });
+
+    const fakeDispatch = sinon.stub(store, 'dispatch');
+    const root = render({ addon, review });
+
+    const editButton = root.find('.AddonReviewListItem-begin-reply');
+    expect(editButton).toHaveLength(1);
+
+    const clickEvent = createFakeEvent();
+    editButton.simulate('click', clickEvent);
```
kumar303: Since the click handler doesn't have any logic to check for an admin, I don't think it's necessary to fire a click event fort this test. I know it makes the feature coverage a bit more thorough but I don't think it would really be that helpful.

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -270,6 +271,52 @@ describe(__filename, () => {
...
+        id: addon.id,
+        slug: addon.slug,
+      },
+    });
+
+    const fakeDispatch = sinon.stub(store, 'dispatch');
+    const root = render({ addon, review });
+
+    expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
+    sinon.assert.notCalled(fakeDispatch);
```
kumar303: I don't think it's necessary to make assertions about the dispatch

PR #3908, opened 2017-11-14
Fix langpack display

muffinresearch:
Fixes https://github.com/mozilla/addons/issues/528

The main change here was to use a unfiltered set of language data so all relevant locales are represented rather than using the default subset that's relevant to our site.

I also added sorting by English locale since this makes the ordering make more sense rather than the sorting on locale code when the display is the native locale string. We could look to improve on this in due course but it matches what's on the old site.

Before: 

dictionaries_and_language_packs_ _add-ons_for_firefox

After:

dictionaries_and_language_packs_ _add-ons_for_firefox

### DIFF
--- a/src/amo/components/LanguageTools/index.js
+++ b/src/amo/components/LanguageTools/index.js
@@ -34,7 +34,7 @@ type LanguageToolListProps = {|
 
 
 // Get languages into a list of objects sorted by the english name.
-const languageList = Object.keys(languages).map((langKey) => {
+const sortedLanguages = Object.keys(languages).map((langKey) => {
   return {
     english: languages[langKey].English,
     locale: langKey,
@@ -178,10 +178,10 @@ export class LanguageToolsBase extends React.Component<Props> {
             </Tr>
           </Thead>
           <Tbody>
-            {addons && addons.length ? languageList.map((lang) => {
+            {addons && addons.length ? sortedLanguages.map((language) => {
               const toolsInLocale = addons ?
                 addons.filter((addon) => {
-                  return addon.target_locale === lang.locale;
+                  return addon.target_locale === language.locale;
                 }) : null;
 
               // This means there are no language tools available in this
@@ -202,18 +202,22 @@ export class LanguageToolsBase extends React.Component<Props> {
                 <Tr
                   className={classNames(
                     'LanguageTools-table-row',
-                    `LanguageTools-lang-${lang.locale}`,
+                    `LanguageTools-lang-${language.locale}`,
                   )}
-                  key={lang.locale}
+                  key={language.locale}
                 >
                   <Td>
-                    <strong>{lang.english}</strong> <span lang={lang.locale}>{lang.native}</span>
+                    <strong>
+                      {language.english}
+                    </strong> <span lang={language.locale}>
+                      {language.native}
+                    </span>
                   </Td>
-                  <Td className={`LanguageTools-lang-${lang.locale}-languagePacks`}>
+                  <Td className={`LanguageTools-lang-${language.locale}-languagePacks`}>
                     {languagePacks.length ?
                       <LanguageToolList addons={languagePacks} /> : null}
                   </Td>
-                  <Td className={`LanguageTools-lang-${lang.locale}-dictionaries`}>
+                  <Td className={`LanguageTools-lang-${language.locale}-dictionaries`}>
                     {dictionaries.length ?
                       <LanguageToolList addons={dictionaries} /> : null}
                   </Td>
--- a/src/amo/components/LanguageTools/index.js
+++ b/src/amo/components/LanguageTools/index.js
@@ -15,7 +15,7 @@ import {
   ADDON_TYPE_LANG,
   VIEW_CONTEXT_LANGUAGE_TOOLS,
 } from 'core/constants';
-import languages from 'core/languages';
+import { languagesUnfiltered as languages } from 'core/languages';
 import translate from 'core/i18n/translate';
 import { fetchLanguageTools } from 'core/reducers/addons';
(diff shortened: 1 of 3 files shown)


src/amo/components/LanguageTools/index.js
```diff
@@ -32,6 +32,19 @@ type LanguageToolListProps = {|
...
 |};
 
+
+// Get languages into a list of objects sorted by the english name.
+const languageList = Object.keys(languages).map((langKey) => {
+  return {
+    english: languages[langKey].English,
+    locale: langKey,
+    native: languages[langKey].native,
+  };
```
willdurand: I think this shape deserves a type:

``` js
type Language = {|
  english: string,
  locale: string,
  native: string,
|};
```

src/amo/components/LanguageTools/index.js
```diff
@@ -32,6 +32,19 @@ type LanguageToolListProps = {|
...
+}).sort((a, b) => {
```
willdurand: Here it should be:

``` js
.sort((languageA: Language, languageB: Language): bool => {
   return languageA.localeCompare(languageB);
});
```

muffinresearch: I wasn't too sure about using this, but then I see we already are: https://github.com/mozilla/addons-frontend/blob/cfcd838c812b1d0f2aa84c47c498f971e4a417d0/src/core/reducers/categories.js#L68

muffinresearch:
I'm going to look at adding a sort test as a follow-up. Just looking at the best way to do it.

src/core/languages.js
```diff
@@ -621,3 +621,5 @@ export default config.get('langs').reduce((object, locale) => {
 
   return object;
 }, {});
+
+export const languagesUnfiltered = languages;
```
willdurand: I dont think we should be doing this.

src/amo/components/LanguageTools/index.js
```diff
@@ -15,7 +15,7 @@ import {
   ADDON_TYPE_LANG,
   VIEW_CONTEXT_LANGUAGE_TOOLS,
 } from 'core/constants';
-import languages from 'core/languages';
+import { languagesUnfiltered as languages } from 'core/languages';
```
willdurand: I would use `languages` here, as before.

src/amo/components/LanguageTools/index.js
```diff
@@ -32,6 +32,17 @@ type LanguageToolListProps = {|
   addons?: Array<AddonType>,
 |};
 
+
+// Get languages into a list of objects sorted by the english name.
+const languageList = Object.keys(languages).map((langKey) => {
```
willdurand: this could be `sortedLanguages`

src/amo/components/LanguageTools/index.js
```diff
@@ -167,10 +178,10 @@ export class LanguageToolsBase extends React.Component<Props> {
             </Tr>
           </Thead>
           <Tbody>
-            {addons && addons.length ? Object.keys(languages).map((langKey) => {
-              const toolsInLocale = addons ? addons
-                .filter((addon) => {
-                  return addon.target_locale === langKey;
+            {addons && addons.length ? languageList.map((lang) => {
```
willdurand: s/lang/language?

src/core/languages.js
muffinresearch: Can you expand on why?

src/amo/components/LanguageTools/index.js
muffinresearch: Can you expand on what you're thinking?

src/core/languages.js
willdurand: I think we should keep the default export. You can make the distinction between filtered (or is sorted?) and not filtered/sorted only where you need this distinction.

It should result in less changes too.

src/amo/components/LanguageTools/index.js
willdurand: I'd require `languages` as before and only create an array of `sortedLanguages`. I dont see why you need this a named export in one file only to alias it to its original name here.

src/core/languages.js
muffinresearch: The default export is completely different data.

muffinresearch: @willdurand see https://github.com/mozilla/addons-frontend/blob/92bd46069a2c9fa4bd0d804f496d17c894657e73/src/core/languages.js#L617

willdurand: oh sorry for that, I did not pay attention to that. What does this code do actually?

any chance to come up with a meaningful name and assign it as default? That would make things easier to understand, and we could `export` the `languages` array (which could be renamed too)

```
export const allLanguages = {};

const reducedLanguages = ...;

export default reducedLanguages;
```

muffinresearch: This is why the original code doesn't work, because it's filtering this list of languages to match the supported locales for the site. This is mentioned in the description. The reason for a distinct export rather than exporting the raw data is to prevent the case of someone making assumptions about it.

muffinresearch: I think this code is mainly used by the locale switcher.

src/amo/components/LanguageTools/index.js
muffinresearch: See my other answer which relates to the same thing.

src/core/languages.js
willdurand: I see, but maybe you could directly rename `languages` to `unfilteredLanguages` and export it in `src/core/languages.js`?

muffinresearch: Yep, that would work - I'll see about doing that in a follow-up patch.

PR #3909, opened 2017-11-14
Pin UI tests to an older Firefox

kumar303:
See https://github.com/mozilla/addons-frontend/issues/3910


jrbenny35:
You probably just need to update geckodriver but I have new tests here:  #3808

PR #3927, opened 2017-11-16
Refactor the addons reducer state

### LINKED ISSUE #3421: Don't store addons in top-level of addons reducer

willdurand:
Fix #3421

---

This PR changes the shape of the `addons` reducer state. It introduces two hash maps to quickly lookup an add-on by GUID (disco) or slug (amo) and several selectors.

Add-ons are stored only once and by ID (slugs can change). The `bySlug` and `byGUID` state attributes are hash maps that store the `id` given a `slug` or `guid`.

If you want to retrieve an add-on by slug, you need to get its `id` first, and then get it in the set of `addons`. To ease that, there is a new `getAddonBySlug()` selector. It is the exact same thing for `guid` and `getAddonByGUID()`.

It is also possible to retrieve an add-on by ID using `getAddonByID()` and to get an array of all add-ons with `getAllAddons()`. These selectors should make things more readable.

---

Follow-up: #3957

### DIFF
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -261,10 +261,53 @@ export function createInternalAddon(
   return addon;
 }
 
-const initialState = {};
+type AddonID = number;
 
-export type AddonState = {
-  [addonSlug: string]: AddonType,
+export type AddonState = {|
+  // Flow wants hash maps with string keys.
+  // See: https://zhenyong.github.io/flowtype/docs/objects.html#objects-as-maps
+  byID: { [addonId: string]: AddonType },
+  byGUID: { [addonGUID: string]: AddonID },
+  bySlug: { [addonSlug: string]: AddonID },
+|};
+
+export const initialState: AddonState = {
+  byID: {},
+  byGUID: {},
+  bySlug: {},
+};
+
+export const getAddonByID = (
+  state: { addons: AddonState }, id: AddonID
+): AddonType | null => {
+  return state.addons.byID[`${id}`] || null;
+};
+
+export const getAddonBySlug = (
+  state: { addons: AddonState }, slug: string
+): AddonType | null => {
+  const addonId = state.addons.bySlug[slug];
+
+  return getAddonByID(state, addonId);
+};
+
+export const getAddonByGUID = (
+  state: { addons: AddonState }, guid: string
+): AddonType | null => {
+  const addonId = state.addons.byGUID[guid];
+
+  return getAddonByID(state, addonId);
+};
+
+export const getAllAddons = (
+  state: { addons: AddonState }
+): Array<AddonType> => {
+  const addons = state.addons.byID;
+
+  // TODO: one day, Flow will get `Object.values()` right but for now... we
+  // have to deal with it.
+  // See: https://github.com/facebook/flow/issues/2221.
+  return Object.keys(addons).map((key: string) => addons[key]);
 };
 
 export default function addonsReducer(
@@ -273,20 +316,35 @@ export default function addonsReducer(
 ) {
   switch (action.type) {
     case LOAD_ADDONS: {
-      const { addons } = action.payload;
-      const newState = { ...state };
-      Object.keys(addons).forEach((key) => {
-        const addon = createInternalAddon(addons[key]);
+      const { addons: loadedAddons } = action.payload;
 
-        // We index add-ons by id and slug to be able to retrieve them by any
-        // of these parameters. This is needed to redirect an add-on detail
-        // page loaded by ID.
-        // See: https://github.com/mozilla/addons-frontend/issues/3610
-        // TODO: https://github.com/mozilla/addons-frontend/issues/3421
-        newState[addon.id] = addon;
-        newState[key] = addon;
+      const byID = { ...state.byID };
+      const byGUID = { ...state.byGUID };
+      const bySlug = { ...state.bySlug };
+
+      Object.keys(loadedAddons).forEach((key) => {
+        const addon = createInternalAddon(loadedAddons[key]);
+        // Flow wants hash maps with string keys.
+        // See: https://zhenyong.github.io/flowtype/docs/objects.html#objects-as-maps
+        byID[`${addon.id}`] = addon;
+
+        if (addon.slug) {
+          bySlug[addon.slug] = addon.id;
+        }
+
+        if (addon.guid) {
+          // `guid` is already "normalized" with the `getGuid()` function in
+          // `createInternalAddon()`.
+          byGUID[addon.guid] = addon.id;
+        }
       });
-      return newState;
+
+      return {
+        ...state,
+        byID,
+        byGUID,
+        bySlug,
+      };
     }
     default:
       return state;
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -304,10 +304,8 @@ export const getAllAddons = (
 ): Array<AddonType> => {
   const addons = state.addons.byID;
 
-  // TODO: one day, Flow will get `Object.values()` right but for now... we
-  // have to deal with it.
-  // See: https://github.com/facebook/flow/issues/2221.
-  return Object.keys(addons).map((key: string) => addons[key]);
+  // $FLOW_FIXME: see https://github.com/facebook/flow/issues/2221.
+  return Object.values(addons);
 };
 
 export default function addonsReducer(
--- a/src/disco/containers/DiscoPane.js
+++ b/src/disco/containers/DiscoPane.js
@@ -13,6 +13,7 @@ import tracking from 'core/tracking';
 import { INSTALL_STATE } from 'core/constants';
 import InfoDialog from 'core/containers/InfoDialog';
 import { addChangeListeners } from 'core/addonManager';
+import { getAddonByGUID } from 'core/reducers/addons';
 import {
   NAVIGATION_CATEGORY,
   VIDEO_CATEGORY,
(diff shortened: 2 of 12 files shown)


willdurand:
@kumar303 I would not say no to a first look :) I am testing this patch locally in the meantime.

src/core/reducers/addons.js
```diff
@@ -286,10 +286,55 @@ export function createInternalAddon(
...
+): AddonType | null => {
+  const addonId = state.addons.bySlug[slug];
+
+  return getAddonByID(state, addonId);
+};
+
+export const getAddonByGUID = (
+  state: { addons: AddonState }, guid: string
+): AddonType | null => {
+  const addonId = state.addons.byGUID[guid];
```
kumar303: This *might* need to use [`getGuid()`](https://github.com/willdurand/addons-frontend/blob/8b810ba3f9d18f729217144c38f6b6ae909538ce/src/core/reducers/addons.js#L121):

```js
const addonId = state.addons.byGUID[getGuid(addon)];
```

but I'm not sure. Does the Discopane work with your current patch? You would need to be using a build with `mozAddonManager` then install an add-on and refresh the page to test it because I think that's only when the bonkers `installations` reducer kicks in.

kumar303: This is how the Discopane API results come in: https://github.com/willdurand/addons-frontend/blob/8b810ba3f9d18f729217144c38f6b6ae909538ce/src/disco/api.js#L11

src/disco/containers/DiscoPane.js
```diff
@@ -172,7 +173,12 @@ export class DiscoPaneBase extends React.Component {
 
 export function loadedAddons(state) {
   return state.discoResults.map(
-    (result) => ({ ...result, ...state.addons[result.addon] })
+    (result) => {
+      return {
+        ...result,
+        ...getAddonByGUID(state, result.addon),
```
kumar303: `normalizer` makes this kind of code so hard to understand. You can see how `result.addon` is mapped here: https://github.com/willdurand/addons-frontend/blob/8b810ba3f9d18f729217144c38f6b6ae909538ce/src/disco/api.js#L7

tests/unit/amo/test_store.js
```diff
@@ -29,6 +29,6 @@ describe('amo createStore', () => {
 
   it('creates an empty store', () => {
     const { store } = createStore();
-    expect(store.getState().addons).toEqual({});
+    expect(store.getState().addons.addons).toEqual({});
```
kumar303: I'm not sure what the point of this test was to be honest. I suppose you could also make it `expect(store.getState()).toEqual(initialState);`. Or maybe just delete it.

Also, I see a `describe(__filename, ...` opportunity in here.

tests/unit/core/reducers/test_addons.js
```diff
@@ -32,42 +39,65 @@ describe(__filename, () => {
...
+      'second-slug': 456,
+    });
   });
 
   it('ignores empty results', () => {
     const addonResults = [];
     const state = addons(undefined,
       loadAddons(createFetchAllAddonsResult(addonResults).entities));
-    expect(Object.keys(state)).toEqual([]);
+    expect(Object.keys(state.addons)).toEqual([]);
```
kumar303: How is this equal to an array and not an [empty object](https://github.com/mozilla/addons-frontend/pull/3927/files#diff-21aa4f8ea4fba6406233649d0ce65beaR298)?

src/core/reducers/addons.js
```diff
@@ -286,10 +286,55 @@ export function createInternalAddon(
   return addon;
 }
 
-const initialState = {};
+type AddonID = string;
```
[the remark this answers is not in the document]
willdurand: maybe I could switch to `number` actually

src/core/reducers/addons.js
```diff
@@ -286,10 +286,55 @@ export function createInternalAddon(
...
-export type AddonState = {
-  [addonSlug: string]: AddonType,
+export type AddonState = {|
+  addons: { [addonId: AddonID]: AddonType },
+  byGUID: { [addonGUID: string]: AddonID },
+  bySlug: { [addonSlug: string]: AddonID },
+|};
+
+const initialState: AddonState = {
+  addons: {},
```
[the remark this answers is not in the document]
willdurand: ah that's interesting, no reason except that I had to start this patch somewhere 😛 Now, it makes sense to use `byID`.

src/core/reducers/addons.js
```diff
@@ -286,10 +286,55 @@ export function createInternalAddon(
...
+  byGUID: {},
+  bySlug: {},
+};
+
+export const getAddonByID = (
+  state: { addons: AddonState }, id: string
+): AddonType | null => {
+  if (!id) {
```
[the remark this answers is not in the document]
willdurand: I wanted to avoid `undefined` as much as possible but yeah that should work too.

src/core/reducers/addons.js
willdurand: Yes, it took me a while to figure out how the disco results were fetched and "normalized".

I think this selector should be as clean as possible and the `getuid` should be used to retrieve the `guid` that will later be passed to it.

src/disco/containers/DiscoPane.js
willdurand: Yes, that's what I figured out. I did not want to rewrite everything but I guess adding some comments would be useful.

tests/unit/amo/test_store.js
willdurand: Yes, I figured out that we could use `initialState`...

Good catch for the `__filename` 😉

tests/unit/core/reducers/test_addons.js
willdurand: because it is using `Object.keys()`?

src/core/reducers/addons.js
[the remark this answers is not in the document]
willdurand: mmh no, it expects a `string` in:

```
Object.keys(addons).map((key: AddonID) => addons[key]);
```

We should do this because `Object.values()` cannot be used with Flow...

willdurand:
While testing this patch with @kumar303, we found an issue related to the language packs. This patch fixes a bug but introduces a regression... This is related to #3939 so let's fix this issue first and then we'll be able to merge this patch.

src/core/reducers/addons.js
tofumatt: FYI I tested this patch with Disco Pane locally on a `mozAddonManager`-patched Nightly and it loaded add-ons and both themes and extensions could be installed and uninstalled, so it appears to work just fine.

src/core/reducers/addons.js
```diff
@@ -261,10 +261,53 @@ export function createInternalAddon(
...
+
+export const getAllAddons = (
+  state: { addons: AddonState }
+): Array<AddonType> => {
+  const addons = state.addons.byID;
+
+  // TODO: one day, Flow will get `Object.values()` right but for now... we
+  // have to deal with it.
+  // See: https://github.com/facebook/flow/issues/2221.
+  return Object.keys(addons).map((key: string) => addons[key]);
```
tofumatt: Can't we just tell flow to ignore this line instead? This seems silly (and while it might not be a huge deal, possibly less efficient in case `Object.values()` is optimised by the browser better than this inline function).

src/disco/containers/DiscoPane.js
```diff
@@ -172,7 +173,15 @@ export class DiscoPaneBase extends React.Component {
 
 export function loadedAddons(state) {
   return state.discoResults.map(
-    (result) => ({ ...result, ...state.addons[result.addon] })
+    (result) => {
+      return {
+        ...result,
+        // `result` comes from the API call in `src/disco/api.js` and
+        // normalizer makes everything complicated...
+        // `addon` is actually the add-on's GUID.
```
tofumatt: I'd clarify with `result.addon` instead of `addon`, just a nitpick though.

willdurand: ah right, thanks

src/core/reducers/addons.js
willdurand: I am fine with ignoring this line but I think it will confuse Flow for the rest of the code. Since I did that for the `languageTools` reducer, I'd like to follow-up in another patch. WDYT?

tofumatt: What happens if you do `// $FlowFixMe` on that one line? If it doesn't mess stuff up I'd prefer our code not be made more onerous because of Flow.

willdurand: Alright doing this now. I'll patch the `languageTools` reducer too then (in a separate PR). I think we do that many times though...

tofumatt: Oh you're right, but I realised that and it started to really bug me. A tool for writing good code shouldn't make us write worse code 😉

willdurand: I could not agree more.

willdurand:
Let's see how it goes now ✨

PR #3948, opened 2017-11-19
Fixes Edge case for Language packs data is not loaded

### LINKED ISSUE #3939: Language packs data is not loaded when opened from a deep-linked language details page
## STR:

* https://addons.mozilla.org/en-US/firefox/addon/british-english-dictionary-2/
* click on "More..." -> "Dictionaries and Language Packs"

## What I expect to happen:

Language packs page should load

## What happens: 

dictionaries_and_language_packs_ _add-ons_for_firefox

There are no requests made for the data. Reloading the page shows the data.

tsl143:
Fixes #3939 
As per comment https://github.com/mozilla/addons-frontend/issues/3939#issuecomment-345421811

### DIFF
--- a/src/amo/components/LanguageTools/index.js
+++ b/src/amo/components/LanguageTools/index.js
@@ -246,7 +246,7 @@ export const mapStateToProps = (
     Object.values(addons).filter((addon) => {
       // I don't know why we need to check for type but flow complains if we
       // don't. 🤷🏼‍
-      return addon && addon.type && (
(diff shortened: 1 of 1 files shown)


willdurand:
Hi @tsl143, thanks for this PR! It looks like a nice fix but it is not the one we want here. Actually, we should not mix the data from the Search API and the Language Tools API because they do not have the same "shape".

I'm working on a patch to not rely on the `addons` reducer anymore and use a `languageTools` reducer instead. Sorry!

PR #3967, opened 2017-11-23
chore: Send params to TAAR (fix #3804)

### LINKED ISSUE #3804: [TAAR] Pass locale/platform to the TAAR-api when opening the Discopane
This is a feature request for the [TAAR project](https://github.com/mozilla/taar/). When Firefox opens the Discopane, it sends along some bit of information (locale, platform, ...). 

We would love this informations to be propagated to the TAAR api to make use of it if some user requesting recommendations is not in our database.

I will file a similar issue in mozilla/addons-server.

tofumatt:
Fix #3804.

These tests are quite old so this patch is pretty meh... I didn't want to spend a lot of time fixing up the tests because it'd be a PITA to review and I need to start on the user pages... there aren't actually many more weeks to this quarter 😓

The lang is already sent to the API so this just adds the platform sent to the Disco Pane to the API request, which addons-server will send to TAAR.

### DIFF
--- a/src/disco/containers/DiscoPane.js
+++ b/src/disco/containers/DiscoPane.js
@@ -33,6 +33,9 @@ export class DiscoPaneBase extends React.Component {
     i18n: PropTypes.object.isRequired,
     location: PropTypes.object.isRequired,
     mozAddonManager: PropTypes.object,
+    params: {
+      platform: PropTypes.string.isRequired,
+    },
     results: PropTypes.arrayOf(PropTypes.object).isRequired,
     _addChangeListeners: PropTypes.func,
     _tracking: PropTypes.object,
@@ -51,7 +54,13 @@ export class DiscoPaneBase extends React.Component {
     super(props);
     this.state = { showVideo: false };
 
-    const { dispatch, errorHandler, location, results } = props;
+    const {
+      dispatch,
+      errorHandler,
+      location,
+      params,
+      results,
+    } = props;
     // TODO: fix this; it's not the right way to detect whether a
     // dispatch is needed. This should look for an undefined value
     // instead of an empty list because an empty list could be a valid
@@ -59,6 +68,7 @@ export class DiscoPaneBase extends React.Component {
     if (!errorHandler.hasError() && !results.length) {
       dispatch(getDiscoResults({
         errorHandlerId: errorHandler.id,
+        platform: params.platform,
         telemetryClientId: location.query.clientId,
       }));
     }
--- a/tests/unit/disco/containers/TestDiscoPane.js
+++ b/tests/unit/disco/containers/TestDiscoPane.js
@@ -67,6 +67,7 @@ describe(__filename, () => {
       dispatch: sinon.stub(),
       i18n,
       location: { query: {} },
+      params: { platform: 'Darwin' },
       results,
       _tracking: fakeTracking,
       _video: fakeVideo,
@@ -75,7 +76,9 @@ describe(__filename, () => {
   }
 
   function render(props = {}) {
-    return shallow(<DiscoPaneBase {...renderProps(props)} />);
+    return shallow(<DiscoPaneBase {...renderProps(props)} />, {
+      params: { platform: 'Darwin' },
+    });
   }
 
   function renderAndMount(customProps = {}) {
@@ -208,7 +211,9 @@ describe(__filename, () => {
       render({ errorHandler, dispatch, ...props });
 
       sinon.assert.calledWith(dispatch, getDiscoResults({
-        errorHandlerId: errorHandler.id, telemetryClientId: undefined,
+        errorHandlerId: errorHandler.id,
+        platform: 'Darwin',
+        telemetryClientId: undefined,
       }));
     });
 
@@ -227,6 +232,7 @@ describe(__filename, () => {
 
       sinon.assert.calledWith(dispatch, getDiscoResults({
         errorHandlerId: errorHandler.id,
+        platform: 'Darwin',
         telemetryClientId: location.query.clientId,
       }));
     });
--- a/tests/unit/disco/test_actions.js
+++ b/tests/unit/disco/test_actions.js
@@ -31,14 +31,27 @@ describe('disco/actions/loadDiscoResults', () => {
 
 describe('disco/actions/getDiscoResults', () => {
   function defaultParams() {
-    return { errorHandlerId: 'some-id', telemetryClientId: 'client-id' };
+    return {
+      errorHandlerId: 'some-id',
+      platform: 'Darwin',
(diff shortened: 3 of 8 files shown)


tests/unit/disco/test_actions.js
```diff
@@ -41,6 +45,13 @@ describe('disco/actions/getDiscoResults', () => {
       .toThrow(/errorHandlerId is required/);
   });
 
+  it('requires platform', () => {
+    const params = defaultParams();
+    delete params.platform;
+    expect(() => getDiscoResults(params))
```
willdurand: I thought we had to make things a little bit more explicit? Something like:

```
expect(() => {
  // func
}).toThrow();
```

tests/unit/disco/test_api.js
```diff
@@ -17,23 +17,30 @@ describe(__filename, () => {
 
   describe('getDiscoveryAddons', () => {
     it('calls the API', () => {
-      getDiscoveryAddons({ api: apiState });
+      getDiscoveryAddons({ api: apiState, platform: 'Windows' });
```
willdurand: Do we have constants for that? Windows, Darwin, etc.

src/disco/containers/DiscoPane.js
```diff
@@ -176,8 +184,9 @@ export function loadedAddons(state) {
   );
 }
 
-export function mapStateToProps(state) {
+export function mapStateToProps(state, ownProps) {
   return {
+    platform: ownProps.params.platform,
```
willdurand: I dont see the `params` prop in the `PropTypes`.

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -203,12 +213,15 @@ describe(__filename, () => {
     it('gets discovery results when results are empty', () => {
       const dispatch = sinon.stub();
       const errorHandler = new ErrorHandler({ id: 'some-id', dispatch });
-      const props = helpers.mapStateToProps(loadDiscoResultsIntoState([]));
+      const props = helpers.mapStateToProps(
+        loadDiscoResultsIntoState([]), ownProps);
 
-      render({ errorHandler, dispatch, ...props });
+      render({ errorHandler, dispatch, platform: 'Darwin', ...props });
```
willdurand: Why is this needed if `props` is constructed using `ownProps` before?

tofumatt: Oh, possibly a holdover. I think it can be removed.

tests/unit/disco/test_actions.js
tofumatt: Indeed! This was copypasta. I'll fix this one and the one I copied it from.

src/disco/containers/DiscoPane.js
tofumatt: Ah, I put `platform` in there because it's what `mapStateToProps` returns. Should I add `params` as well? That's confusing ^_^

willdurand: Yeah, `params` are provided by react-router (like `location`). We need to declare all props, not only the mapped ones (that are redux only).

tests/unit/disco/test_api.js
tofumatt: We don't, they're different from what we usually use–I think Firefox sends them.

src/disco/containers/DiscoPane.js
tofumatt: Ah okay, cool.

tests/unit/disco/test_api.js
willdurand: oh ok, fine then

tofumatt:
Okay, I think I made all the tweaks needed for this.

tests/unit/disco/test_api.js
tofumatt: I don't even know all the values though, it would be great to document them going forward. I'll at least add a note about that...

src/disco/containers/DiscoPane.js
```diff
@@ -33,6 +33,8 @@ export class DiscoPaneBase extends React.Component {
     i18n: PropTypes.object.isRequired,
     location: PropTypes.object.isRequired,
     mozAddonManager: PropTypes.object,
+    params: PropTypes.object.isRequired,
```
willdurand: Well, ideally you want to be more explicit:

```
params: {
  platform: PropTypes.string.isRequired,
},
```

src/disco/containers/DiscoPane.js
```diff
@@ -33,6 +33,8 @@ export class DiscoPaneBase extends React.Component {
...
+    platform: PropTypes.string.isRequired,
```
willdurand: Aaaah I see. I think you dont need to do that. Instead, rely on `params.platform` in the code.

src/disco/containers/DiscoPane.js
```diff
@@ -176,8 +185,9 @@ export function loadedAddons(state) {
   );
 }
 
-export function mapStateToProps(state) {
+export function mapStateToProps(state, ownProps) {
   return {
+    platform: ownProps.params.platform,
```
willdurand: Yeah sorry, actually you should not do that. Remove the `ownProps` arg here, and use `params.platform` in the component.

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -126,14 +129,21 @@ describe(__filename, () => {
   });
 
   describe('mapStateToProps', () => {
+    it('sets platform', () => {
+      const props = helpers.mapStateToProps(
+        loadDiscoResultsIntoState([]), ownProps);
+
+      expect(props.platform).toEqual('Darwin');
+    });
```
willdurand: So yeah, this should not be done here I guess. We should not rely on `ownProps` here.

willdurand: The `params` prop should be added to the `renderProps()` function, example: https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestCollection.js#L45-L48.

tests/unit/disco/containers/TestDiscoPane.js
```diff
@@ -241,7 +247,7 @@ describe(__filename, () => {
           guid: '@guid1',
           slug: 'discovery-addon-1',
         },
-      }]));
+      }]), { params: { platform: 'Darwin' } });
```
willdurand: mmh this looks weird here. What is tested?

tofumatt: Oops it's wrong and should have been removed but was before I was just adding `ownProps` everywhere. I'll remove it 😄

PR #3974, opened 2017-11-24
Remove asnyc-connect (redux-connect) and add react-router-redux

### LINKED ISSUE #181: Dispatch react-router changes with Redux (i.e. track URLs in state)

### LINKED ISSUE #2356: Do all pre-loading with redux-saga, not redux-async-connect

### LINKED ISSUE #3648: Remove unused errorPage reducer for old error pages
This code is no longer used: https://github.com/mozilla/addons-frontend/blob/2e870ae660214ab6873793497f29cca17d9a7049/src/core/server/base.js#L300

The reducer can also be removed: https://github.com/mozilla/addons-frontend/blob/9fd02d91f28d2c96d5541f1883640f6b4bfd6be1/src/core/reducers/errorPage.js

This was an older implementation of rendering error pages when redux async connection failures occurred. We currently handle this directly at the end of the promise chain: https://github.com/mozilla/addons-frontend/blob/9fd02d91f28d2c96d5541f1883640f6b4bfd6be1/src/core/server/base.js#L345

willdurand:
Fix #181
Fix #2356

---

This PR completely removes `redux-async-connect`/`redux-connect` and adds `react-router-redux` instead to keep backward compatibility. (PS: it is the road to React 16)


tests/unit/core/reducers/test_errorPage.js
```diff
@@ -21,23 +24,29 @@ describe('errorPage reducer', () => {
     expect(state).toEqual(initialState);
   });
 
-  describe('REDUX_CONNECT_END_GLOBAL_LOAD', () => {
+  describe('LOCATION_CHANGE', () => {
```
tofumatt: This is so much nicer anyway 👍

PR #3975, opened 2017-11-24
Upgrade to React 16

### LINKED ISSUE #3252: sort by name, and other sort options
https://addons.mozilla.org/firefox/search/?q=tab+groups for example, with the classic desktop view there's a visible option to sort by name. A total of eight sort options. 

No such option with the mobile view. 

## Partial workaround

Manually edit the URL, for example: 

https://addons.mozilla.org/firefox/search/?q=tab+groups&sort=name

– but then the Filter results column wrongly shows that results are sorted by relevance.

willdurand:
~⚠️ depends on #3974~

Closes #3238 
Closes #3247 
Closes #3250 
Closes #3251 
Closes #3259 
Closes #3629 

---

~There is one test case I am not able to fix and I have no clue right now...~


willdurand:
Just rebased this patch!

kumar303:
whoops, `tests/unit/core/server/testServer.js` changed in https://github.com/mozilla/addons-frontend/pull/3944/files

I see that #3252 is in the close list (in the PR description). How does this upgrade close that?

willdurand:
> I see that #3252 is in the close list (in the PR description). How does this upgrade close that?

erf, wrong autocompletion I guess.

willdurand:
Rebased one more time. I think you can have a quick look until I figure out how to fix the skipped test.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -583,7 +583,7 @@ describe(__filename, () => {
       );
     });
 
-    it('resets the page after submitting a review', () => {
+    it.skip('resets the page after submitting a review', () => {
```
willdurand: This is the skipped test I have mentioned in the PR desc and a comment. I am not sure to understand why it is failing, but passing the `router` parameter makes it fail with a very weird message. I'll get back to that with a fresh mind, but a second brain could be useful too :)

tofumatt: From IRC:

```
[20:07]  <tofumatt>	willdurand: hmmm, this failing react 16 test is weird indeed
[20:07]  <tofumatt>	I'm left scratching my head on it itoo
[20:17]  <tofumatt>	willdurand: weirdly, if I don't use shallowUntilTarget but just shallow
[20:17]  <tofumatt>	and then add the extra missing props
[20:17]  <tofumatt>	I can get it to work
```

tofumatt: Changing that test to:

```jsx
it('resets the page after submitting a review', () => {
      const router = { push: sinon.stub() };
      dispatchAddon({ ...fakeAddon });
      const location = { query: { page: 2 } };
      const props = getProps({
        addon: { ...fakeAddon },
        clientApp: CLIENT_APP_FIREFOX,
        dispatch: sinon.stub(),
        errorHandler: createStubErrorHandler(),
        lang: 'en-US',
        reviews: null,
        router,
        location,
      });

      const root = shallow(<AddonReviewListBase {...props} />);

      const manager = root.find(RatingManager);
      expect(manager).toHaveProp('onReviewSubmitted');
      const onReviewSubmitted = manager.prop('onReviewSubmitted');

      // Simulate a review submission.
      onReviewSubmitted();

      sinon.assert.calledWith(router.push, {
        pathname: `/${lang}/${clientApp}${root.instance().url()}`,
        query: { page: 1 },
      });
    });
```

And adding an `import { shallow } from 'enzyme';` still runs the tests and fixes the issue. Not ideal, but it does get us to React 16... 😄

willdurand: This test case + a comment would indeed be a good thing.

I have the feeling that Enzyme 3 is not ultra stable but react 16 would be very nice.

kumar303:
This patch looks good. I'm surprised that so few changes were needed. 🎈 

The [`withRouter`](https://reacttraining.com/react-router/web/api/withRouter) docs mention how we need to fix the ordering of HOCs when mixing with Redux:

```js
// This gets around shouldComponentUpdate
withRouter(connect(...)(MyComponent))

// This does not
connect(...)(withRouter(MyComponent))
```

Our use of `withRouter` looks correct in all places except for `AddonReviewList`, which may explain the test failure. However, I don't know if there is an easy fix here. When I corrected the order I just get a new error:

```
The component had a falsey params.addonSlug parameter

      at Function.mapStateToProps [as mapToProps] (src/amo/components/AddonReviewList/index.js:316:2317)
      ...
```

This makes me think we shouldn't be mocking `location` or `params` in any of the tests. As for `params`, I think this may be an undocumented prop that `withRouter` is injecting, maybe for legacy purposes (unproven theory!).

What if this test dispatches a new router action instead of using mocks? That would simulate the real app nicely. I don't know how hard that is. If it works, we should replace all mock locations and params with that approach but this could be done in a follow-up patch.

tests/unit/amo/components/TestAddonReviewList.js
kumar303: It would be nice to keep using `shallowUntilTarget` because I'm pretty sure this is catching a real problem with our use of `withRouter` (my comment below has details).

tofumatt: Indeed, I would prefer to use `shallowUntilTarget` for the same reasons. I was hoping my patch might help expose the underlying issue, not be what we moved to ulessabdolutely necessary 😊. Your explanation below makes a lot of sense, thanks. Hopefully we can fix this and still use `shallowUntilTarget`.

willdurand:
Given @kumar303's details, it looks like it is also the only component with the `withRouter` HOC and `ownProps` passed to `mapStateToProps()`.

willdurand:
> As for params, I think this may be an undocumented prop that withRouter is injecting

it is not hidden 😉 it is documented here: https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#injected-props.

tests/unit/amo/components/TestAddonReviewList.js
willdurand: So, I ended up using `setProps()` to pass the `router` stub to the base component. It works with `shallowUntilTarget()` even though it is not ideal.

I could not use an action to change the current route because we do not store anything in the state (nothing useful I mean). It is recommended to use the `params` provided by the HOC in the doc: https://github.com/reactjs/react-router-redux#how-do-i-access-router-state-in-a-container-component, and it is also mentioned:

> You should not read the location state directly from the Redux store.

So I believe the `params` approach is fine...

`shallowUntilTarget()` (or Enzyme 3) prevents the `withRouter` HOC to pass the props correctly to the `mapStateToProps()`. Fixing the loading order of the HOC did not help 😞 

Dispatching an action to "push" a route would work only if we use the `Router` component because I think it does more than just passing the `router` to its children (which is one of the reasons we can omit Redux `Provider` component for instance).

willdurand:
rebased to get a green status :)

kumar303:
> it is documented here: https://github.com/ReactTraining/react-router/blob/v3/docs/API.md#injected-props.

Bah, I was looking at an [old version](https://reacttraining.com/react-router/web/api/withRouter) of the docs but the page doesn't make that clear.

So, yeah, we should probably stop mocking all those props 😲 

In most cases, we can probably do `router` stubbing like:

```js
const root = render(...);

const push = sinon.stub(root.instance().props.router, 'push');
// Trigger event that calls router.push() ...
// sinon.assert.calledWith(push, ...);
```

willdurand:
> Bah, I was looking at an old version of the docs but the page doesn't make that clear.

It is actually the latest version. `params` has moved to `match.params` in RR v4.

> So, yeah, we should probably stop mocking all those props 😲

I guess yes, but I am looking at "how" now. It would certainly ease the upgrade from RR v3 to v4.

I tried stubbing like you did but in the failing test case of the `AddonReviewList`, I could not make it work 😞

willdurand:
Ok I have it! The `withRouter()` HOC passes the `params` and `location` props to its children. Inverting the loading order caused `mapStateToProps()` to not receive the correct props, but by switching to a `router` object that looks like what it is expected by the HOC, everything works again.

tests/unit/amo/components/TestAddonReviewList.js
willdurand: More info below 👇

willdurand:
Note: RR v4 provides a `MemoryRouter` that is often recommended for testing the components that use the `withRouter` HOC. We're not there yet though.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -46,12 +46,21 @@ describe(__filename, () => {
...
-  const getProps = ({ ...customProps }) => {
+  const getProps = ({ location, params, router, ...customProps } = {}) => {
     return {
       i18n: fakeI18n(),
-      location: { query: {} },
-      params: {
-        addonSlug: fakeAddon.slug,
+      // The `withRouter()` HOC uses the `router` to pass `params` and
+      // `location` to its children.
+      router: {
```
kumar303: It's a shame that we have to mock the whole router object but I don't have a better solution. I thought it would be possible to start the test with:

```js
import { push } from 'react-router-redux';
...
store.dispatch(push(`/en-US/firefox/addon/${fakeAddon.slug}/reviews/?page=2`))
```

...but I fell down a rabbit hole trying to get that to work. Oh well.

kumar303:
@tofumatt not sure if you've looked yet -- I'd feel better if you also reviewed this one before we merge.

tofumatt:
I had a look a bit before and things looked good, but that was before the router fix/mocks. I'll have another look tonight in about an hour. 👍

tests/unit/amo/components/TestAddonReviewList.js
tofumatt: Yeah not having to mock it would be great. Maybe at least a comment or an issue encouraging us to try to be better in the future 😉

willdurand:
@tofumatt feel free to merge then, I cannot anymore :)

tofumatt:
Ah right, of course! Thank you so much for this one!

Issue #3755, opened 2017-11-02
Add statsd / datadog measurements of NodeJS server side rendering


kumar303:
For each HTTP request, the frontend server makes an API request (at least one) and then renders the React components to a string twice -- once to trigger sagas and again after the initial data has loaded. 

We should add `statsd` timing for this so we can graph it in DataDog.

kumar303:
We have a latency graph from the AWS ELB (load balancer) (see https://github.com/mozilla/addons-frontend/issues/3916#issuecomment-344740803) but we need this data to help keep an eye on server load. We had a load issue in prod: https://github.com/mozilla/addons-frontend/issues/3916

kumar303:
This library seems like the most maintained and up to date: https://github.com/brightcove/hot-shots

Here is the only express middleware implementation I could find. It's using an old datadog lib and it doesn't have any tests so we should probably copy it into our app: https://github.com/AppPress/node-connect-datadog/blob/master/lib/index.js

kumar303:
Depends on https://github.com/mozilla/addons-frontend/issues/3933 to connect to datadog on dev

kumar303:
Looks like it's working (dev only for now) https://app.datadoghq.com/dash/407599/addons-frontend-server

PR #3983, opened 2017-11-27
Use different datadog keys per count/timing (for stats)

kumar303:
Follow up to https://github.com/mozilla/addons-frontend/issues/3755

The timing and count stats need to have different keys.


willdurand:
ah oops I totally missed that.

kumar303:
No worries. I was just fixing my dashboard after looking at the data so I didn't really think it needed a review.

PR #4018, opened 2017-12-01
Fix the way we detect IDs (integer) in slugs

### LINKED ISSUE #3610: Loading add-on page using numerical id instead of slug leads to blank page

willdurand:
Fix #3610

---

In a previous PR, I thought `Number.isInteger` was more readable and explicit but it turns out it does not parse stringified numbers. I reverted the changes to use `!isNaN` and provided a test case.


willdurand:
@tofumatt feel free to merge :)

tofumatt:
Indeed, was just waiting for Travis to be green 😅

kumar303:
JavaScript 
machine filled with bees

PR #4028, opened 2017-12-04
chore: Remove NSP ignore now that fix is deployed

### LINKED ISSUE #3195: yarn nsp-check is failing re: tough-cookie ReDOS

tofumatt:
fix #3195

Pretty sure we fixed this, will let the tests run to make sure.

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,6 +1,4 @@
 {
   "exceptions": [
-    // https://github.com/mozilla/addons-frontend/issues/3989
-    "https://nodesecurity.io/advisories/532"
   ]
 }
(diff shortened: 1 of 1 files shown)


willdurand:
@tofumatt there is a fix for moment now (to get rid of the other exception).

tofumatt:
Hah, cool, I was about to check that too. I guess I could do them in the same PR.

PR #4031, opened 2017-12-04
feat: Add user reducers/sagas/apis

### LINKED ISSUE #4030: Create reducers/sagas for users/user profiles
To start on #2781 we need reducers to manage multiple users' profiles. We'll need API files, sagas, etc. to deal with it.

tofumatt:
Fix #4030.

Adds the reducers/etc. needed for users profiles.

<strike>This doesn't integrate with the user reducer but I guess it could. I didn't look that much into it because it deals with a different set of actions and I found it a bit less clear how the user reducer was used.

If I should integrate them let me know.</strike>


tests/unit/helpers.js
```diff
@@ -404,6 +404,8 @@ export function createUserProfileResponse({
   };
 }
 
+export const createUserAccountResponse = createUserProfileResponse;
```
tofumatt: In the future this might be different but for now I just exported the reference so that if we changed it in the future we can just do it at the source.

kumar303:
> This doesn't integrate with the user reducer but I guess it could.

I think they should be integrated because both will need to store the same detail object: https://addons-server.readthedocs.io/en/latest/topics/api/accounts.html#account-object-self

The existing reducer/actions/sagas seems related: 
* https://github.com/mozilla/addons-frontend/blob/master/src/core/reducers/user.js#L71-L78
* https://github.com/mozilla/addons-frontend/blob/master/src/core/sagas/user.js#L27

Even though the existing saga uses a different API endpoint, it returns the same account detail object.

The current `reducers/user` state has a similar problem to the `collections` one. All the current values need to move to `users.byId` (which would store the account object). The existing entries would need a link like `users.signedInUser`, an ID number pointing to an object in `users.byId`.

tofumatt:
Alright. That makes sense... I'm hoping that's not too much work to integrate things, I'll see about it now.

tofumatt:
Okay, this is ready for review I think. I've combined the users and user reducer, and tidied up a fair amount of stuff. Sorry this ballooned to a wildly-large PR.

src/amo/reducers/users.js
```diff
@@ -0,0 +1,256 @@
...
+  last_login?: string,
+  last_login_ip?: string,
+  is_verified?: boolean,
+  permissions: ?Array<string>,
+  read_dev_agreement?: boolean,
+  // Properties we add to each object.
+  displayName: ?string,
+|};
+
+export type UsersStateType = {
```
kumar303: I don't see anything to indicate that user data is loading. If you think it's fine to add later then it can be postponed. I guess you could add it later like:

```js
export type UsersStateType = {
  currentUserID: number | null,
  byID: { [userId: number]: UserType },
  byUsername: { [username: string]: UserId },
  developerProfilePage: {
    loading: boolean,
    user: UserId | null,
  },
};
```

Theoretically, `currentUser` should have a way to indicate loading state but since that will always happen server side, I'm not worried about it. It would be a lot of work to change that!

src/amo/reducers/users.js
```diff
@@ -0,0 +1,256 @@
...
+  | LogOutUserAction;
+
+const reducer = (
+  state: UsersStateType = initialState,
+  action: Action
+): UsersStateType => {
+  switch (action.type) {
+    case LOAD_CURRENT_USER_ACCOUNT: {
+      const { user } = action.payload;
+      user.displayName = getDisplayName(user);
```
kumar303: This code is duplicated. You should introduce a helper that both reducer actions can use. Something like:

```js
const { user, newState } = addUserToState({ user: action.payload.user, state });
return {
  ...newState,
  currentUserID: user.id,
};
```

tests/unit/amo/api/test_users.js
```diff
@@ -0,0 +1,77 @@
...
+        })
+        .returns(mockResponse());
+
+      return userAccount({ api: state.api, username })
+        .then(() => mockApi.verify());
+    });
+
+    it('throws an error if api state is missing', () => {
+      expect(() => {
+        userAccount({ username: 'sweet-little-duck' });
```
kumar303: I find these kind of tests are a lot easier to maintain if you use a common helper for the params:

```js
const params = getParams();
delete params.api;
expect(() => userAccount(params)).toThrowError(...);
```

tests/unit/amo/reducers/test_users.js
```diff
@@ -0,0 +1,206 @@
...
+      expect(newState).toEqual(state);
+    });
+
+    it('requires errorHandlerId for fetchCurrentUserAccount', () => {
+      expect(() => {
+        fetchCurrentUserAccount({});
+      }).toThrowError('errorHandlerId is required');
+    });
+
+    it('fetchCurrentUserAccount returns action', () => {
```
kumar303: This test doesn't seem useful to me. You will already have coverage of this when testing interaction with the reducer.

tests/unit/amo/reducers/test_users.js
```diff
@@ -0,0 +1,206 @@
...
+      const errorHandlerId = createStubErrorHandler().id;
+      const action = fetchCurrentUserAccount({ errorHandlerId });
+
+      expect(action.type).toEqual(FETCH_CURRENT_USER_ACCOUNT);
+      expect(action.payload).toMatchObject({ errorHandlerId });
+    });
+
+    it('fetchUserAccount returns payload and type', () => {
```
kumar303: Same thing here: if you just test how this action affects the reducer / saga then you don't need this test. It just creates a maintenance burden.

tests/unit/core/server/testServer.js
```diff
@@ -173,53 +173,53 @@ describe(__filename, () => {
...
-      const { api, user } = store.getState();
+      const { api, users } = store.getState();
 
       expect(response.statusCode).toEqual(200);
       expect(api.token).toEqual(token);
-      expect(user.id).toEqual(42);
-      expect(user.username).toEqual('babar');
-      mockUserApi.verify();
+      expect(users.currentUserID).toEqual(42);
+      expect(users.byID[users.currentUserID].username).toEqual('babar');
```
kumar303: you could use `getCurrentUser(state)` here

src/amo/reducers/users.js
```diff
@@ -0,0 +1,256 @@
...
+import { ALL_SUPER_POWERS } from 'core/constants';
+
+
+export const LOG_OUT_USER: 'LOG_OUT_USER' = 'LOG_OUT_USER';
+export const LOAD_CURRENT_USER_ACCOUNT: 'LOAD_CURRENT_USER_ACCOUNT' = 'LOAD_CURRENT_USER_ACCOUNT';
+export const FETCH_CURRENT_USER_ACCOUNT: 'FETCH_CURRENT_USER_ACCOUNT' = 'FETCH_CURRENT_USER_ACCOUNT';
+export const FETCH_USER_ACCOUNT: 'FETCH_USER_ACCOUNT' = 'FETCH_USER_ACCOUNT';
+export const LOAD_USER_ACCOUNT: 'LOAD_USER_ACCOUNT' = 'LOAD_USER_ACCOUNT';
+
+export type UserId = string;
```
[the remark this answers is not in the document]
willdurand: is that type actually used?

src/amo/reducers/users.js
```diff
@@ -0,0 +1,256 @@
...
+
+export type UserType = {|
```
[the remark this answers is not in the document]
tofumatt: Oh yeah, good call 👍🏻

src/amo/reducers/users.js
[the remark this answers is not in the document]
tofumatt: It might not even be... either way I'll tidy it up.

src/amo/reducers/users.js
tofumatt: I was going to add it later in with the actual pages but no harm in doing so now. 😄

src/amo/reducers/users.js
```diff
@@ -0,0 +1,256 @@
...
+
+export function logOutUser(): LogOutUserAction {
+  return {
+    type: LOG_OUT_USER,
+    payload: {},
+  };
+}
+
+export const getCurrentUser = (users: UsersStateType) => {
+  if (!users.currentUserID || !users.byID[users.currentUserID]) {
```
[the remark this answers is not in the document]
tofumatt: Oh, true.

tofumatt:
Cool, this should be ready for another review; sorry I got busy with other junk and travel and didn't get to it earlier.

tests/unit/amo/helpers.js
```diff
@@ -214,18 +219,20 @@ export function dispatchSignInActions({
   authToken = userAuthToken(),
   userId = 12345,
   username = 'user-1234',
-  displayName = null,
+  // eslint-disable-next-line camelcase
+  display_name = null,
```
kumar303: Were you trying to avoid confusion here? Using camel case would help keep things consistent. It seems ok to me just to set like `display_name: displayName`

tests/unit/core/server/testServer.js
```diff
@@ -241,13 +241,13 @@ describe(__filename, () => {
...
```
kumar303: you could use `getCurrentUser(state.users)` here

tests/unit/core/server/testServer.js
```diff
@@ -257,40 +257,41 @@ describe(__filename, () => {
       const reduxStoreState = JSON.parse($('#redux-store-state').html());
 
       expect(reduxStoreState.api).toEqual(api);
-      expect(reduxStoreState.user).toEqual(user);
+      expect(reduxStoreState.users.byID[reduxStoreState.users.currentUserID])
```
kumar303: You could also use `getCurrentUser()` here too

tests/unit/amo/helpers.js
tofumatt: Yeah, because this function/helper specifically creates an object that mimics an API response, I didn't like setting the `display_name` property with our (assigned/computed) `displayName` prop. It's really confusing to set `displayName` which will actually set `display_name` which is checked to set `displayName`. I thought the change disambiguated things; before then I found it hard to follow.

tofumatt:
Thanks for the reviews on this one; it caught a fair few nits I didn't because it's a pretty big diff. 👍

kumar303:
screenshot 2017-12-14 09 38 14

clint-eastwood-nod

PR #4034, opened 2017-12-05
Fix column height calculation in ESR

muffinresearch:
Fixes https://github.com/mozilla/addons/issues/566

Before the fix the search results height is > 50000px high before you get to the pagination.

search_results_for__toggle_javascript__ _add-ons_for_firefox

After the fix - there's no massive gap:

search_results_for__toggle_javascript__ _add-ons_for_firefox

### DIFF
--- a/src/amo/components/Search/styles.scss
+++ b/src/amo/components/Search/styles.scss
@@ -48,6 +48,8 @@
       grid-row: 2 / 5;
       // Allows the grid item to contain the content preventing overflow.
       min-width: 0;
+      // Hack to force column height calculations in ESR 52.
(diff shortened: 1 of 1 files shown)


src/amo/components/Search/styles.scss
```diff
@@ -48,6 +48,8 @@
       grid-row: 2 / 5;
       // Allows the grid item to contain the content preventing overflow.
       min-width: 0;
+      // Hack to force column height calculations in ESR 52.
```
tofumatt: A link to the actual issue here might be nice too 😊

PR #4042, opened 2017-12-06
I4040 - Highlight author names on details page

### LINKED ISSUE #4040: Developer name in header of listing page should be highlighted as a hyperlink
Previously discussed in #3771 and #3936, but the scope of those bugs was different. We should highlight the username in the header of add-on listing pages, to make it clearer that it's a link:
screen shot 2017-11-16 at 3 55 35 pm
The image above also highlights the word "by", but I don't think that should be part of the link, since there can be multiple authors.

tsl143:
Fixes #4040 

**Before:**
screen shot 2017-12-06 at 4 06 45 pm

**After**
screen shot 2017-12-06 at 4 04 40 pm

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -84,7 +84,6 @@
   &,
   a,
   a:link {
-    color: $black;
     font-size: $font-size-m;
     font-weight: normal;
     text-decoration: none;


src/amo/components/Addon/styles.scss
```diff
@@ -84,7 +84,6 @@
   &,
   a,
   a:link {
-    color: $black;
```
kumar303: Hah! Thanks @tsl143 

magic

Issue #4045, opened 2017-12-06
Loading an incomplete add-on causes purple/blank screen while logged in as an admin


ValentinaPC:
Steps to reproduce:
1. Load homepage while logged in https://addons.allizom.org/en-US/firefox/
2. Click on an hero banner that has no stage correspondent i.e. https://addons.allizom.org/en-US/firefox/addon/decentraleyes/
3. Refresh the page
4. Go back to homepage
5. Refresh the page

Expected results:
After step 2 a 404 page should be displayed.

Actual results:
After step 2 a purple page is loaded.
After step 3 the purple page is replaced with a Server Error page and user seems logged out.
After step 5 user seems logged in again.

Notes/Issues:
- issue is not reproducing while logged out.
- issue is also loaded for unexisting add-ons i.e. https://addons.allizom.org/en-US/firefox/addon/photo-screen-saver/

Verified on AMO-stage FF57(Win 7).
Video for this issue:
![login-logout](https://user-images.githubusercontent.com/15685960/33662176-8dc341e8-da93-11e7-8cc1-0e9e01041127.gif)

kumar303:
@ValentinaPC from my observations, this only happens when you sign in using an admin account. For any other user account, it's not happening. Can you verify?

kumar303:
The root cause is https://sentry.prod.mozaws.net/operations/addons-frontend-amo-stage/issues/776253/

<img width="969" alt="screenshot 2017-12-06 10 26 11" src="https://user-images.githubusercontent.com/55398/33672482-ea2315d6-da6f-11e7-93bf-4df7ad513531.png">

ValentinaPC:
@kumar303 : Yes I can confirm that being logged out or being logged in with a non-admin account will not cause the issue.

kumar303:
Another problem is that `decentraleyes` is an incomplete add-on meaning that it was not fully submitted through DevHub. This won't be happening in prod (for this specific add-on). Normal users can't see these add-ons but admins can.

```
$ ssh ...@olympia-web-amostage1...
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> a = Addon.objects.filter(slug='decentraleyes')[0]
>>> a
<Addon: 521554: Decentraleyes>
>>> a.status
0
```

ValentinaPC:
Verified as fixed on AMO-dev FF57(Win 7).
[...]

PR #4049, opened 2017-12-06
Cope with null versions when dealing with incomplete add-ons

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4045

In addition to addressing that issue, this changes the Flow type of `addon.current_version` so that all typed code will be forced to guard against it.

Visiting a broken add-on now looks like this:

screenshot 2017-12-06 17 31 07

### DIFF
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -680,6 +680,16 @@ describe(__filename, () => {
     expect(root.prop('location')).toEqual(location);
   });
 
+  it('does not show a ratings manager without a version', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      current_version: null,
+    });
+    const root = shallowRender({ addon });
+
+    expect(root.find(RatingManagerWithI18n)).toHaveLength(0);
+  });
+
   it('renders a summary', () => {
     const root = shallowRender();
     expect(root.find('.Addon-summary').html()).toContain(fakeAddon.summary);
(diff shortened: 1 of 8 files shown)


tests/unit/amo/components/TestAddon.js
```diff
@@ -680,6 +680,16 @@ describe(__filename, () => {
     expect(root.prop('location')).toEqual(location);
   });
 
+  it('does not show a ratings manager without a version', () => {
```
tofumatt: Should we show a message or reason why we aren't displaying the ratings manager? I know this is an admin-only issue but as you mentioned on the call QA uses admin users a lot when doing manual testing... so they might run into a new error of "the ratings manager disappeared" ^_^

Maybe we should just show a message like "Ratings disabled because this add-on is incomplete" or something?

kumar303: Oh yeah, good idea. I updated the description with a screenshot of what it looks like now. I'm pretty sure no one would find a page like this in real life. It would be hard even for an admin to find one unless they knew to look for an incomplete add-on.

kumar303: And of course dev has an incomplete add-on featured on the home page for some reason 😛

tofumatt: I dig it, thanks 😄

tofumatt: Yeah I was gonna say -dev is always crazypants data.

PR #4060, opened 2017-12-08
I4056 - fixes word break in review list page

### LINKED ISSUE #4056: Add-on name has layout issues in 'Read all reviews' page
Steps to reproduce:
1. Load 'All reviews' page for an add-on with a longer title (i.e. https://addons.allizom.org/en-US/firefox/addon/grammarly-spell-checker/reviews/)
2. Observe the add-on name displayed in the details section

Expected result:
Add-on name is wrapped correctly

Actual result;
Characters are pushed to the second row if the entire word does not fit on the first line

Notes: reproduced on all AMO servers with FF57 (Win10 and Android 7.0)

image
image

tsl143:
Fixes #4056 

Before:
screen shot 2017-12-08 at 2 26 03 pm

After:
screen shot 2017-12-08 at 2 26 28 pm

### DIFF
--- a/src/amo/components/AddonReviewList/styles.scss
+++ b/src/amo/components/AddonReviewList/styles.scss
@@ -53,6 +53,10 @@
   height: 36px;
 }
 
+.AddonReviewList-header-text {
+  min-width: 50%;
+}
+
 .AddonReviewList-header-icon-image {
   display: block;
   height: 36px;
@@ -62,7 +66,7 @@
   font-size: $font-size-m;
   margin: 0;
   padding: 0;
-  word-break: break-all;
+  word-wrap: break-word;
 
   a,
   a:link {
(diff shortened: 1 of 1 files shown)


src/amo/components/AddonReviewList/styles.scss
```diff
@@ -53,6 +53,10 @@
   height: 36px;
 }
 
+.AddonReviewList-header-text {
+  min-width: 50%;
```
tsl143: Rationale for this is [here](https://stackoverflow.com/questions/36150458/flex-item-overflows-container-due-to-long-word-even-after-using-word-wrap#answer-36150764)

Display flex assigns min-width as total width of element if not defined so word-wrap still overflows the container.
hence need to use this with `word-wrap: break-word`

tofumatt: Huh, cool. If you could add a link to that rationale to the CSS in a comment that'd be cool.

tofumatt:
Closed in 4af23d304f4d11d2d6b51a96341aa33ed6b4a5e7.

PR #4072, opened 2017-12-12
I4063-Fixes release note issue

### LINKED ISSUE #4063: Release Notes won't be displayed when they contain some HTML tags
Steps to reproduce:

1. Go to the details page of an addon that contains Release Notes written with some HTML tags (i.e.https://addons-dev.allizom.org/en-US/firefox/addon/test-of-today/)
2. Scroll down to the Release Notes section and observe the text included by the developer
3. Refresh the page and take another look into the Release Notes section

Expected result:
Release Notes are displayed

Actual result:
The notes have disappeared from their section

Notes: I was able to reproduce this issue by using the following HTML tags: `<ul> <ol> <li> <blockquote>`
Reproduces on all AMO servers with FF57 (Win 10 and Android 7.0)

release notes

tsl143:
Fixes #4063 

screen shot 2017-12-13 at 3 52 32 am


tsl143:
> I don't think this should be a <span> but rather a <div>. I also think we should add a test for that to make sure we don't regress this behaviour.

Will modifying test cases do, or we need to write new test case?

tofumatt:
If this doesn’t fail any tests (which it seems it doesn’t) then we should write a new test and reference this issue in it :-)

tsl143:
In order to replace with `div`, I have updated the test case.

kumar303:
What @tofumatt meant is that we should add a regression test for the actual bug which is that a tag like `<ul>` was not showing up within a `<p>` tag. Did you try writing a test for that? I'm not sure how easy it is to test for since it might be a browser rendering issue rather than an html markup issue. Let us know if you get stuck.

kumar303:
Hi @tsl143, let me know if you have any questions about how to write a test for this.

tsl143:
@kumar303 My bad, I was unable to understand @tofumatt point, it makes complete sense now. Updated the test case with previously failed scenario. Please let me know if I need to add anything more.

tofumatt:
Thanks for the extra test case @tsl143, all set and merged! 👍

Issue #4114, opened 2018-01-03
Create a shared Select component


kumar303:
Since the UI for adding add-ons to a collection (https://github.com/mozilla/addons-frontend/issues/2768) needs a select box, we need to move the current select styles to a shared component.

For QA: please test the functionality and look and feel of the [current search filters](https://addons.mozilla.org/en-US/firefox/search/?platform=mac&q=ad+blocker) (the select boxes) for regressions.

vcarciu:
No regressions were found so far in this area. I will mark the issue as verified.

PR #4115, opened 2018-01-03
Extract a shared Select component

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4114

### DIFF
--- a/src/ui/components/Select/styles.scss
+++ b/src/ui/components/Select/styles.scss
@@ -1,0 +1,27 @@
+@import "~photon-colors/colors";
+
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+.Select {
+  @include padding-end(22px);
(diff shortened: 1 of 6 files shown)


src/ui/components/Select/styles.scss
```diff
@@ -0,0 +1,27 @@
+@import "~photon-colors/colors";
+
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+.Select {
+  @include padding-end(22px);
+  @include padding-start(5px);
```
tofumatt: I wonder if all these styles need updating? It can be done later if you like but seeing a lot of multiples of 5 in here when often the new styles are multiples of 6. If it's easy and can be folded into this patch then 👍, otherwise no worries.

kumar303: Good catch, thanks

Issue #4116, opened 2018-01-03
Support next page API URLs


kumar303:
For adding add-ons to a collection (https://github.com/mozilla/addons-frontend/issues/2768) we need to load multiple pages of [user collections](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#list). This requires the following adjustments to our core `callApi()` function:

* It should accept absolute URLs since that's what the API returns in the `next` parameter
* It should preserve query string parameters in API endpoint URLs

For QA: It would be helpful to test some pages with query string parameters for regressions, such as the [search page](https://addons.mozilla.org/en-US/firefox/search/?platform=mac&q=ad+block).

AlexandraMoga:
I've looked for possible regressions around query string parameters but no issues were noticed.
I will mark this as verified fixed (AMO-dev, FF57, Win10)

PR #4117, opened 2018-01-03
Support absolute URLs, query string params in callApi()

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4116

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -25,7 +25,7 @@ const Entity = normalizrSchema.Entity;
 export const addon = new Entity('addons', {}, { idAttribute: 'slug' });
 export const category = new Entity('categories', {}, { idAttribute: 'slug' });
 
-export function makeQueryString(query: { [key: string]: * }) {
+export function makeQueryString(query: { [key: string]: any }) {
   const resolvedQuery = { ...query };
   Object.keys(resolvedQuery).forEach((key) => {
     const value = resolvedQuery[key];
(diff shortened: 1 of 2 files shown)


src/core/api/index.js
```diff
@@ -91,10 +91,35 @@ export function callApi({
...
+    return Promise.reject(
+      new Error(`endpoint URL cannot be falsy: "${endpoint}"`)
+    );
+  }
   if (errorHandler) {
     errorHandler.clear();
   }
-  const queryString = makeQueryString({ ...params, lang: state.lang });
+
+  const urlObj = url.parse(endpoint, true);
```
tofumatt: Just a nitpick but I like descriptive variable names; we're past the grt Unix vwl shrtg yrs 😉

Eg. could this be `urlObject` instead?

src/core/api/index.js
```diff
@@ -116,7 +141,10 @@ export function callApi({
     }
   }
 
-  let apiURL = `${API_BASE}/${endpoint}/${queryString}`;
+  adjustedEndpoint = adjustedEndpoint.endsWith('/') ?
```
tofumatt: Does this mess up any requests? I *think* the API always wants a trailing slash but I feel like we should double-check for any existing requests that don't have them and may be relying on the trailing slash being omitted.

I feel like that might be happening someplace? I know the API sometimes returns different errors for bad requests with/without the trailing slash so maybe that's why I'm thinking it matters, and it might not.

src/core/api/index.js
kumar303: Sure, that's fine. I tend to prefer shorter variables because it helps me read the code faster (e.g. less letters to parse). I can change it though.

src/core/api/index.js
kumar303: The old code [always appended a slash](https://github.com/mozilla/addons-frontend/blob/a60c82a55cb3e53a1a57005e62992e6230489572/src/core/api/index.js#L119) to `endpoint` so that's not a new change.

This new code is just a bit more forgiving. If you make a helper that does `callApi({ endpoint: 'something/' })` it just works whereas before you'd end up with double slashes and your API request would always fail.

Issue #4118, opened 2018-01-03
API functions for adding an add-on to a collection


kumar303:
We need some new API functions for adding an add-on to a collection (https://github.com/mozilla/addons-frontend/issues/2768). Additionally, we need a helper to fetch all pages of an API response.

Issue #3993, opened 2017-11-28
Add an add-on to a new collection


kumar303:
Tracker: https://github.com/mozilla/addons-frontend/issues/2782

The user should be able to:
- Sign in
- Go to the detail listing of an add-on
- Add the add-on to a new collection
- During the process, name the new collection

Note that there is a feature related to this where you can add an extension to an existing collection: https://github.com/mozilla/addons-frontend/issues/2768

pwalm:
Drop down on the add-on detail page. Normal state:
![screen shot 2017-12-04 at 10 47 53 am](https://user-images.githubusercontent.com/6797635/33570125-acb0664e-d8e0-11e7-9ca4-2fdef0d92d3f.png)

Active state:
![screen shot 2017-12-04 at 10 48 08 am](https://user-images.githubusercontent.com/6797635/33570129-b165a622-d8e0-11e7-9d7c-0b55ade790ca.png)

Launches into "Create a collection" modal flow.

tofumatt:
Just a thought (sure how much of the UI state @kumar303 has implemented so far): can we just use the native OS `<select>` styles instead of that heavily customised one? It's a *lot* more effort to implement that custom one (unless there's an easily restyle-able component out there we can grab) and I'm not sure it helps much with usability. Granted, it looks cool, but doing something like the existing LanguagePicker at the bottom of the page would save us a lot of time and I think still look alright (and probably work better, as `select` tags already have all required browser accessibility stuff built-in).

pwalm:
Totally cool with going native.

kumar303:
> Totally cool with going native.

I like it 🍰 

The implementation uses native styles for when you open the menu (just like the search page). I agree that it would have been a huge rabbit hole (with accessibility issues) to re-style that part.

bobsilverberg:
@jvillalobos This feature has not worked since we implemented the new front end, because all we've done is link to the legacy "Create a collection" form, but without passing any information about the add-on.

Would we still like to implement this, enhancing the current functionality to redirect the user to the new "Create a collection" screen, and also remembering the add-on and automatically adding it to the collection after creation? The latter would require some sort of UX change because right now there's no way to indicate that an add-on will be added to the collection after creation, and we do not currently allow users to add add-ons to a collection before creation.

kumar303:
> Would we still like to implement this

I think it's worth implementing. The *Add to collection* button is a nice shortcut and an entry point that will help users discover the feature of creating collections.

[...]

PR #4119, opened 2018-01-03
API functions for adding an add-on to a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4118

### DIFF
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -1,9 +1,5 @@
 /* @flow */
-/* eslint-disable no-await-in-loop */
-import { oneLine } from 'common-tags';
-
-import { callApi } from 'core/api';
-import log from 'core/logger';
+import { callApi, allPages } from 'core/api';
 import type {
   ExternalCollectionAddon, ExternalCollectionDetail,
 } from 'amo/reducers/collections';
@@ -64,30 +60,25 @@ export const getCollectionAddons = (
   return callApi(request);
 };
 
+type GetAllCollectionAddonsParams = {|
+  ...GetCollectionParams,
+  _allPages?: typeof allPages,
+  _getCollectionAddons?: typeof getCollectionAddons,
+|};
+
 export const getAllCollectionAddons = async (
-  { api, slug, user }: GetCollectionParams
+  {
+    api,
+    slug,
+    user,
+    _allPages = allPages,
+    _getCollectionAddons = getCollectionAddons,
+  }: GetAllCollectionAddonsParams
 ): Promise<Array<ExternalCollectionAddon>> => {
-  let allResults = [];
-  let done = false;
-  let nextURL;
-
-  // TODO: make a helper function for this pattern.
-  while (!done) {
-    const response = await getCollectionAddons({
-      api, nextURL, slug, user,
-    });
-    allResults = allResults.concat(response.results);
-
-    if (response.next) {
-      nextURL = response.next;
-      log.debug(oneLine`Fetching next page "${nextURL}" of
-        getCollectionAddons for "${user}/${slug}"`);
-    } else {
-      done = true;
-    }
-  }
-
-  return allResults;
+  const { results } = await _allPages(
+    (nextURL) => _getCollectionAddons({ api, nextURL, slug, user })
+  );
+  return results;
 };
 
 type ListCollectionsParams = {|
@@ -107,27 +98,24 @@ export const listCollections = (
   return callApi({ auth: true, endpoint, state: api });
 };
 
+type GetAllUserCollectionsParams = {|
+  ...ListCollectionsParams,
+  _allPages?: typeof allPages,
+  _listCollections?: typeof listCollections,
+|};
+
 export const getAllUserCollections = async (
-  { api, user }: ListCollectionsParams
+  {
+    api,
+    user,
+    _allPages = allPages,
+    _listCollections = listCollections,
+  }: GetAllUserCollectionsParams
 ): Promise<Array<ExternalCollectionDetail>> => {
-  let allResults = [];
-  let done = false;
-  let nextURL;
-
-  while (!done) {
-    const response = await listCollections({ api, user, nextURL });
-    allResults = allResults.concat(response.results);
-
-    if (response.next) {
-      nextURL = response.next;
-      log.debug(oneLine`Fetching next page "${nextURL}" of
-        listCollections for user "${user}"`);
-    } else {
-      done = true;
-    }
-  }
-
-  return allResults;
+  const { results } = await _allPages(
+    (nextURL) => _listCollections({ api, nextURL, user })
+  );
+  return results;
 };
 
 type GetAllUserAddonCollections = {|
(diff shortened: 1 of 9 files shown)


kumar303:
Let's hold off on this patch since there is a PR open to solve https://github.com/mozilla/addons-server/issues/7167 and that will make the API calls here more sane.

kumar303:
This is ready for review now. Since the UI isn't here to try out, let me know if you have any questions. The end goal for these helpers is to implement this UI (which is working in a separate patch): https://github.com/mozilla/addons-frontend/issues/3993#issuecomment-349064460

The implementation of `getAllUserAddonCollections()` may not perform super well but I'm not sure. I think we should try it. I was hoping we wouldn't have to page through all the user's collections to find which ones match the add-on but we do (because of cache-machine). However, we have just lifted cache-machine on collection endpoints so perhaps soon we can make a better API endpoint for it. Again, it might not even be a problem.

src/amo/api/collections.js
```diff
@@ -28,42 +32,153 @@ export const getCollectionDetail = (
...
   }
 
-  return callApi({
+  const request = {
     auth: true,
-    endpoint: `accounts/account/${user}/collections/${slug}/addons`,
-    params: { page },
+    endpoint:
+      nextURL || `accounts/account/${user}/collections/${slug}/addons`,
+    params: undefined,
```
tofumatt: Huh, I guess this is fine but it seems weird to pass `undefined`. Could we just leave this blank or use `null`? Just seems odd, is all.

src/amo/api/collections.js
```diff
@@ -28,42 +32,153 @@ export const getCollectionDetail = (
...
     state: api,
-  });
+  };
+  if (!nextURL || page) {
```
tofumatt: This seems an odd conditional. I'd think you'd want to check for either `!nextURL && page` (to make sure the param we're setting has a value below) or just check for a false-y `nextURL`... I guess I just find it strange but maybe there's a reason I'm missing here. 😄

src/core/api/index.js
```diff
@@ -265,3 +266,42 @@ export function autocomplete({ api, filters }: AutocompleteParams) {
     state: api,
   });
 }
+
+type GetNextResponseType =
+  (nextURL?: string) => Promise<PaginatedApiResponse<any>>;
+
+type AllPagesOptions = {| pageLimit: number |};
+
+export const allPages = async (
```
tofumatt: This is neat and probably fine for now but as mentioned it might be nice to request an API for this. I'd think a single request/endpoint would be better than what we're doing here 😄

tests/unit/amo/api/test_collections.js
```diff
@@ -141,4 +179,200 @@ describe(__filename, () => {
...
+          state: apiState,
+        })
+        .once()
+        .returns(createApiResponse());
+
+      await addAddonToCollection(params);
+      mockApi.verify();
+    });
+
+    it('posts notes about an addon', async () => {
```
tofumatt: This spec confused me at first because I thought it was about posting notes but really it's about posting an add-on with accompanying notes, right? I think it would be clearer to label it as "posts an add-on and notes" or something like that.

kumar303:
Actually, we found a way to simplify the UX for this which means some of these API helpers aren't needed. I'll prepare a new patch.

kumar303:
Ah whoops, I didn't realize you reviewed it already. Yes! We can definitely fix the performance now because I can remove the API request for getting all collection add-ons entirely. Sorry for the noise, I thought you were on PTO so I was waiting to see if I could get it fixed in time before commenting 😄 I may still be able to get it fixed today.

src/amo/api/collections.js
kumar303: It's necessary so that Flow knows that `params` is a valid key. It actually makes a bit of sense too because adding `request.param` when it's uninitialized could be a mistake.

src/core/api/index.js
kumar303: I removed the code that makes a lot of API requests but we still need to fetch all of the user's collections to fill up the select box.

kumar303:
I removed `getAllUserAddonCollections()` because we no longer need it for the UI. This should be ready for another look.

src/amo/api/collections.js
```diff
@@ -1,6 +1,10 @@
 /* @flow */
-import { callApi } from 'core/api';
+import { callApi, allPages } from 'core/api';
+import type {
+  ExternalCollectionAddon, ExternalCollectionDetail,
```
tofumatt: This is just an aside but I think once we can't fit the entire import statement on one line we might as well do each one on a line; I think it's easier to scan the names if they're on newlines and it makes it easier to add/remove imports. 🤷🏻‍♂️

src/amo/api/collections.js
tofumatt: Would it be okay to use `null` instead? It just seems weird to state a key as `undefined`.

src/amo/api/collections.js
```diff
@@ -28,42 +32,118 @@ export const getCollectionDetail = (
...
-    endpoint: `accounts/account/${user}/collections/${slug}/addons`,
-    params: { page },
+    endpoint:
+      nextURL || `accounts/account/${user}/collections/${slug}/addons`,
+    params: undefined,
     state: api,
-  });
+  };
+  if (page) {
+    request.params = { page };
```
tofumatt: Would there ever be a case where other params exist? Maybe we should be defining `params` as `{}` above and doing `request.params = { ...request.params, page };` here? It might make the above `params: undefined` code a bit easier to logically follow as well.

src/amo/api/collections.js
kumar303: It has to be `undefined` so that the default value for `params` is used when `callApi` is invoked.

In fact, Flow will help you not make this mistake:

```
Error: src/amo/api/collections.js:60
 60:   return callApi(request);
                      ^^^^^^^ object literal. This type is incompatible with the expected param type of
 94: }: CallApiParams): Promise<any> {
        ^^^^^^^^^^^^^ object type. See: src/core/api/index.js:94
  Property `params` is incompatible:
     53:     params: null,
                     ^^^^ null. This type is incompatible with
     77:   params?: Object,
                    ^^^^^^ object type. See: src/core/api/index.js:77
```

src/amo/api/collections.js
kumar303: > Would there ever be a case where other params exist?

No. Here's an explanation that will hopefully clarify things.

This function does not accept `request.params` as input. The line of code you're looking at is a hard coded execution of `callApi()` to handle only two cases:

1. Getting collection add-ons for a given user/collection at page 1 (the default page)
2. Getting collection add-ons at page *N* (via the `next` URL returned by the API)

Also, due to the way spreads work, `request.params` should only be defined if `page` is set (this is why it's wrapped in an if statement). Otherwise, the `page` parameter from the `next` URL will be reset. Imagine this:

```js
const query = {
  // This is how callApi() will extract ?page= from the `next` URL
  ...{ page: 2 },
  // This would be the custom `request.params` value passed to `callApi()`
  ...{ page: undefined }
};
```

In that code the value of `query.page` would be `undefined` which is not what we want. We want it to be `2`. The fix is to execute the function more like:

```js
callApi({ endpoint: nextUrlWithPageParam, params: undefined, ... })
```

Let me know if you have more questions about it.

kumar303: I'll add some comments to the code

kumar303:
> It would probably still be cool to have an API added that's tailored to the result set we want here that would save us from combing through paginated requests to get more info

That's not even possible. To fill a select box with the user's collections we need to get all collections. The API will only give us the first page (25 collections). 

I put a safety guard in there to prevent us from fetching an insane amount of pages but in reality I doubt many users would have more than 25 collections anyway.

I think this is a problem we are going to have to solve eventually for other scenarios. We've been able to avoid thus far because we can provide a UI to navigate between pages. That didn't seem practical for a select box.

Thanks for the review!

PR #4126, opened 2018-01-04
[Docs] - mozilla/addons-frontend - adding separators to README.md file

tiagomoraismorgado88:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
 Front-end infrastructure and code to complement
 [mozilla/addons-server](https://github.com/mozilla/addons-server).
 
+---
+
 ## Security Bug Reports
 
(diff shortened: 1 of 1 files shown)


muffinresearch:
Hi @tiagomoraismorgado88 I'm not sure there's a need for this. The Headings clearly separate the different areas of content already. 

Did you discuss this in an issue already? If not, it might be helpful to do that in future before creating a PR since in a case like this we may have decided this wasn't necessary ahead of time.

It's also worth noting that referencing pull requests to an open issue is detailed as part of the pull request template, see https://github.com/mozilla/addons-frontend/blob/master/PULL_REQUEST_TEMPLATE.md

PR #4130, opened 2018-01-04
Add pyup configuration.

### LINKED ISSUE #4128: Add python dependency updater.
This would help keep the UI tests dependency's updated.

jrbenny35:
Fixes #4128

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,3 @@
+branch: master
+schedule: "everyweek on monday"
+assignees: jrbenny35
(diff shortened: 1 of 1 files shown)


.pyup.yml
```diff
@@ -0,0 +1,3 @@
+branch: master
+schedule: "everyweek on monday"
```
kumar303: Does this mean it will only send one pull request on Monday? Or will it still send *multiple* pull requests for each dependency on Monday? I'd like to only have one pull request.

jrbenny35: According to the docs it will create ONE PR, with all the updates. https://pyup.io/docs/bot/config/#schedules

Issue #2768, opened 2017-07-12
Add an add-on to an existing collection


tofumatt:
Tracker: https://github.com/mozilla/addons-frontend/issues/2782

The user should be able to:
- Sign in
- Go to the detail listing of an add-on
- Add the add-on to an existing collection

UX mock: https://github.com/mozilla/addons-frontend/issues/3993#issuecomment-349064460

jvillalobos:
The [plan for collections](https://docs.google.com/document/d/1DutmS8yco-hNiKJ1-0MrkAkze7Ue88xpA4b-MWx6MLI/edit) eliminates the special-casing of the favorites collection. The Add to Collection feature should let you pick from your existing collections, similarly to how it currently works on AMO theme listings.

kumar303:
I'm nominating this for MVP removal because it would only cripple a small percentage of users if we release without it. If should be enough to ship with read-only views (https://github.com/mozilla/addons-frontend/issues/3140 https://github.com/mozilla/addons-frontend/issues/3141). We can do a follow-up release with this edit feature.

jvillalobos:
Removing from MVP since the old front-end hasn't had this for a while. We want it back, but it's not that urgent.

jvillalobos:
There are many requests for this feature, so moving to P2. I was wrong about the old front-end, because we did support it for themes before moving to the new one.

Not sure if we have UX for this one.

AlexandraMoga:
Verified on AMO-dev with FF57 (Win10 and Android 7.0)

Observations:
- Create a collection function works as expected
- Adding an add-on to an existing collection is possible and the add-on is listed among the other items inside the collection 
- notification messages (success/errors) are triggered correctly

Notes:
There are some issues that concern mostly the layout of some specific elements for which I will create separate cases

![image](https://user-images.githubusercontent.com/31961530/35103678-3d192fac-fc6f-11e7-9be6-195ffc8c9f89.png)

![image](https://user-images.githubusercontent.com/31961530/35103697-507c3efe-fc6f-11e7-833e-2997bc274cd2.png)

PR #4148, opened 2018-01-12
Add an add-on to a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/2768

This implements a simple UI for adding an extension to an existing collection from the detail page. If you add an extension to a collection and then reload the page (or visit it again later) you will not see an indication that it is a part of a collection. This was a deliberate choice to reduce complexity. Instead, you will see a clear and concise error if you try to add it.

UX mock: https://github.com/mozilla/addons-frontend/issues/3993#issuecomment-349064460

addons-frontend-collection mov

### DIFF
--- a/src/amo/components/AddAddonToCollection/index.js
+++ b/src/amo/components/AddAddonToCollection/index.js
@@ -242,6 +242,9 @@ export class AddAddonToCollectionBase extends React.Component<Props> {
       });
     }
 
+    // translators: This is a header for a list meaning Add to [some collection name]
+    const collectionOptLabel = i18n.gettext('Add to…');
+
     return (
       <div className={makeClassName('AddAddonToCollection', className)}>
         {errorHandler.renderErrorIfPresent()}
@@ -253,7 +256,7 @@ export class AddAddonToCollectionBase extends React.Component<Props> {
         >
           {actionOptions}
           {collectionOptions.length ? (
-            <optgroup label={i18n.gettext('Add to…')}>
+            <optgroup label={collectionOptLabel}>
               {collectionOptions}
             </optgroup>
           ) : null}
--- a/src/amo/components/AddAddonToCollection/index.js
+++ b/src/amo/components/AddAddonToCollection/index.js
@@ -61,7 +61,8 @@ export class AddAddonToCollectionBase extends React.Component<Props> {
 
   componentDidMount() {
     // This runs as componentDidMount() to only load data client side,
-    // not server side.
+    // not server side. By skipping server side rendering, no extra API
+    // requests will be made on the server.
     this.loadDataIfNeeded();
   }
 
@@ -70,17 +71,17 @@ export class AddAddonToCollectionBase extends React.Component<Props> {
   }
 
   loadDataIfNeeded(nextProps?: Props) {
-    const allProps = { ...this.props, ...nextProps };
+    const combinedProps = { ...this.props, ...nextProps };
     const {
       dispatch,
       errorHandler,
       loadingUserCollections,
       userCollections,
       siteUserId,
-    } = allProps;
+    } = combinedProps;
 
     if (errorHandler.hasError()) {
-      // Abort loadng data so that the error can be rendered.
+      // Abort loading data so that the error can be rendered.
       return;
     }
 
@@ -155,9 +156,9 @@ export class AddAddonToCollectionBase extends React.Component<Props> {
 
     let progressMessage;
     if (loadingUserCollections) {
-      progressMessage = i18n.gettext('Loading...');
+      progressMessage = i18n.gettext('Loading…');
     } else if (loadingAddonsInCollections) {
-      progressMessage = i18n.gettext('Adding...');
+      progressMessage = i18n.gettext('Adding…');
     }
     if (progressMessage) {
       // Create a disabled select box with a single option.
--- a/src/amo/components/AddAddonToCollection/index.js
+++ b/src/amo/components/AddAddonToCollection/index.js
@@ -45,7 +45,11 @@ type Props = {|
 
 type OnSelectOptionType = () => void;
 
-type SelectData = {| options: Array<Object>, disabled: boolean |};
+type SelectData = {|
+  actionOptions: Array<Object>,
+  collectionOptions: Array<Object>,
+  disabled: boolean,
+|};
 
 export class AddAddonToCollectionBase extends React.Component<Props> {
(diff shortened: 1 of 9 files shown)


tofumatt:
I like the UX. It might be nice to do a follow-up issue that removes the added collection from the `<select>` tag once it's added though (I understand it was avoided to reduce complexity in this PR, but it would be worth doing in the future).

src/amo/components/AddAddonToCollection/index.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    const options = [
+      this.createOption({
+        text: i18n.gettext('Add to collection'), key: 'default',
+      }),
+    ];
+
+    options.push(this.createOption({
+      text: i18n.gettext('Create new collection'),
```
tofumatt: I wonder if a little emoji or glyph or something would help this stand out? Or maybe a blank line underneath it (eg `-----`)? I just feel like using this, it's not clear this is a separate action compared to adding to an existing collection, which I think is important to highlight.

tests/unit/amo/components/TestAddAddonToCollection.js
```diff
@@ -0,0 +1,388 @@
...
+
+  it('lets you specify the css class', () => {
+    const root = render({ className: 'MyClass' });
+
+    expect(root).toHaveClassName('MyClass');
+    expect(root).toHaveClassName('AddAddonToCollection');
+  });
+
+  describe('fetching user collections', () => {
+    it('fetches user collections on first render', () => {
```
tofumatt: Should (can?) we have a test that makes sure this is only called on client-side render? (https://github.com/mozilla/addons-frontend/pull/4148/files#diff-79d1d6288c9a4baa1c738148e96b9b83R63)

tests/unit/amo/components/TestAddAddonToCollection.js
```diff
@@ -0,0 +1,388 @@
...
+        userId,
+        collectionId: 321,
+        collectionSlug: 'some-collection',
+        errorHandlerId: 'error-handler',
+      }));
+
+      const root = render({ addon });
+
+      const select = root.find('.AddAddonToCollection-select');
+      expect(select.props().disabled).toEqual(true);
```
tofumatt: I think you can do `expect(select).toHaveProp('disabled', true);` here and in a few other places, which just outputs better error messages IIRC.

src/amo/components/AddAddonToCollection/index.js
```diff
@@ -0,0 +1,301 @@
...
+    _window: typeof window !== 'undefined' ? window : {},
+  };
+
+  constructor(props: Props) {
+    super(props);
+    this.optionSelectHandlers = {};
+  }
+
+  componentDidMount() {
+    // This runs as componentDidMount() to only load data client side,
```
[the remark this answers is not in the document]
kumar303: Oh. The reason it was client only was because it made some API requests which would have increased the server load. It no longer makes those API requests so I'll put it back on the server.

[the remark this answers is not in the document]
kumar303: Actually, nevermind. It still loads data from the collections API when it first renders. I think we should only do it client-side because otherwise it introduces more load on the server. We don't need the select box for SEO so that should be fine. It won't work with JS disabled so why bother.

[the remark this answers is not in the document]
kumar303: I'll add a comment to explain why

tests/unit/amo/components/TestAddAddonToCollection.js
kumar303: I don't think we should because there's very little chance of regressing the feature of loading only on the client. The feature relies on `componentDidMount()` which is well understood to be client-side only and React is committed to supporting that method for the foreseeable future.

src/amo/components/AddAddonToCollection/index.js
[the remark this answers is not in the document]
tofumatt: Ah, cool, that totally makes sense. Comment is perfect, thanks 👍

tests/unit/amo/components/TestAddAddonToCollection.js
tofumatt: Right, true. Sorry 'bout that.

src/amo/components/AddAddonToCollection/index.js
```diff
@@ -0,0 +1,301 @@
...
+    } else {
+      log.warn(`No handler for option: "${key}"`);
+    }
+  }
+
+  addToCollection(collection: CollectionType) {
+    const { addon, errorHandler, dispatch, siteUserId } = this.props;
+    if (!addon) {
+      throw new Error(
+        'Cannot add to collection because no add-on has been loaded');
```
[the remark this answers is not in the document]
kumar303: No, I don't think a user would ever see this error. This select box gets rendered on the add-on detail page so there will always be an add-on. If not, a 404 will be rendered. An error is thrown here just to help in debugging in case a developer breaks something.

src/amo/components/AddAddonToCollection/index.js
```diff
@@ -0,0 +1,301 @@
...
+      _window,
+      i18n,
+      loadingAddonsInCollections,
+      loadingUserCollections,
+      userCollections,
+    } = this.props;
+
+    let progressMessage;
+    if (loadingUserCollections) {
+      progressMessage = i18n.gettext('Loading...');
```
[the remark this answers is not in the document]
kumar303: Good catch. The existing localizations were indeed for `Loading…`.

src/amo/components/AddAddonToCollection/index.js
kumar303: I put it in an optgroup but we can do some follow-ups if needed.


screenshot 2018-01-17 15 15 11

kumar303: Screenshot: https://user-images.githubusercontent.com/55398/35067665-45066d5e-fb9a-11e7-8d00-06f261e35880.png

tofumatt: Oh, that's totally nice, thanks 👍

kumar303:
Thanks for the review. I added an optgroup but we can change it to something else if needed.

https://user-images.githubusercontent.com/55398/35068262-2a1c9dcc-fb9c-11e7-9888-e5bbe5a43ebf.png

Issue #3706, opened 2017-10-31
Introduce Photon-style notification bar


pwalm:
We need to introduce the photon [notification ("message") bar](http://design.firefox.com/photon/components/message-bars.html) into AMO. I'll provide specs on how it should be integrated into our site properly soon.

For QA: please check the style of notifications against the spec (linked above). Here are some screenshots of what has been changed: https://github.com/mozilla/addons-frontend/pull/4149

Be sure to check the list at the top of https://github.com/mozilla/addons-frontend/pull/4149 for what has *not* been implemented yet.

willdurand:
I have updated the labels since this looks more like a post release feature than a mvp-polish.

kumar303:
In addition to incompatibility notices, the error notifications also need updating. Here are some examples:

<img width="663" alt="screenshot 2017-11-15 12 04 00" src="https://user-images.githubusercontent.com/55398/32852411-6d706e44-c9fd-11e7-9d1c-da23cad6e2e0.png">
<img width="370" alt="screenshot 2017-11-15 12 06 24" src="https://user-images.githubusercontent.com/55398/32852414-6f7bf58c-c9fd-11e7-82d1-6b4b7b0190d9.png">

pwalm:
Implementation of this one is going to take a bit of problem-solving, but the design system team and I are working on it.

pwalm:
Here are the implementation mocks. Basically, the notification should appear in the block to which it pertains (how we've been doing it, but always at the top).
![screen shot 2017-11-20 at 3 33 34 pm](https://user-images.githubusercontent.com/6797635/33046980-3e938ffc-ce08-11e7-9fe5-c4ddeafcd918.png)

Full-width notifications look a bit odd, might have to finesse them a bit.
![screen shot 2017-11-20 at 3 41 21 pm](https://user-images.githubusercontent.com/6797635/33047178-4b085834-ce09-11e7-8b17-7f8e098fd186.png)

AlexandraMoga:
I've verified this on AMO-dev (Win10, Android 7.0) and here are my observations:
1. The suggestion was to have the notifications displayed at the top of the current block, but I see that  the position hasn't changed
2. The [specs](https://design.firefox.com/photon/components/message-bars.html) suggest that the information icon for error messages should be a **`!`**, but we have here an **`i`**
3. The error text was to be white, however links included in this text are blue and not very legible
4. The notification text doesn't seem to be perfectly aligned inside the notification banner
[...]

PR #4149, opened 2018-01-12
Add Photon notifications to AMO

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3706

I needed success notices for the add-to-collection feature so I implemented most of [the Photon notifications](https://design.firefox.com/photon/components/message-bars.html). 

These features have *not* been implemented:
* Generic notices or warnings
* Discopane. I started trying to update this code but it was copy/pasted and has complications due to the flex rules. I can't fit it into this patch.
* Messages are not yet dismissable

screenshot 2018-01-16 11 33 27
screenshot 2018-01-12 12 34 00
screenshot 2018-01-16 11 34 29

Here is a success notice:

screenshot 2018-01-12 08 52 50

Here's an error on the review page:

screenshot 2018-01-16 11 29 51

Here's how it looks with an action button:

screenshot 2018-01-16 11 25 51
screenshot 2018-01-12 09 59 16
screenshot 2018-01-12 09 59 35
screenshot 2018-01-16 11 27 14
screenshot 2018-01-12 10 01 11


willdurand:
😍

tofumatt:
It looks great! 👍 I'll probably get to the review later tonight or tomorrow, but thanks for this!

src/ui/components/Notice/index.js
```diff
@@ -0,0 +1,67 @@
+/* @flow */
+import makeClassName from 'classnames';
+import * as React from 'react';
```
tofumatt: Why not just `import React from 'react';`?

src/ui/components/Notice/styles.scss
```diff
@@ -0,0 +1,97 @@
...
+@import "~ui/css/vars";
+@import "~photon-colors/colors";
+
+.Notice {
+  border-radius: $border-radius-s;
+  display: grid;
+  font-size: $font-size-s;
+  grid-gap: 4px;
+  grid-template-columns: 20px auto;
+  line-height: 15px;
```
tofumatt: `line-height` [should be in unitless measurements](https://css-tricks.com/almanac/properties/l/line-height/#article-header-id-0); I'm guessing this is probably around `1.2` but I'm not sure, you'll have to trial+error it 🙃

src/ui/components/Notice/styles.scss
```diff
@@ -0,0 +1,97 @@
+@import "~core/css/inc/mixins";
+@import "~core/css/inc/vars";
+@import "~ui/css/vars";
+@import "~photon-colors/colors";
+
+.Notice {
+  border-radius: $border-radius-s;
+  display: grid;
+  font-size: $font-size-s;
+  grid-gap: 4px;
```
tofumatt: Most of our base units are `6px` multiples. Should this really be `4px`? If it totally looks right compared to mocks that's okay, but I'd assume 6 is preferred.

src/ui/components/Notice/styles.scss
```diff
@@ -0,0 +1,97 @@
...
+  grid-template-columns: 20px auto;
+  line-height: 15px;
+  min-height: 20px;
```
tofumatt: Same throughout re: `6px` base multiple; these all seem to be multiples of `4` which isn't common in Add-ons or Photon and is in fact a new multiple.

src/ui/components/Notice/styles.scss
```diff
@@ -0,0 +1,97 @@
...
+  // end of the text block when there's no button.
+  @include padding-end(4px);
+}
+
+@mixin notice_button() {
+  &,
+  &.Button--small {
+    border: none;
+    font-size: $font-size-s;
+    line-height: 15px;
```
tofumatt: This should be unitless too.

src/ui/components/Notice/styles.scss
```diff
@@ -0,0 +1,97 @@
...
+    padding: 6px;
+    text-shadow: none;
+    width: max-content;
```
tofumatt: This seems to be unsupported in Edge: https://caniuse.com/#search=max-content

That's not the end of the world IMO but I'd rather us be compatible. Is there any way to replicate what you're doing here and have it work with Edge?

tests/unit/ui/components/TestNotice.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('requires a known type', () => {
+    expect(() => render({ type: 'nope' })).toThrow(/Unknown type: nope/);
+  });
+
+  it('hides action buttons by default', () => {
+    const root = render();
+
+    expect(root).not.toHaveClassName('.Notice-with-action');
+    // There should only be one column besides the icon.
```
tofumatt: This is an uber-nitpick but "besides the icon" is a bit confusing to read–I couldn't tell at first if it meant "other than the icon" or "next to the icon". I think "one column other than the icon column" would clarify things.

kumar303:
Thanks for the review! I'm back from the holiday so I'll start working on this after breakfast. Did you look at [the spec](https://design.firefox.com/photon/components/message-bars.html)? It's pretty nice because it has pixel dimensions and it specifies button placement, etc. Some changes you requested go against that spec. If you would like to do that, we should probably get approval from someone on the Photon project.

src/ui/components/Notice/index.js
kumar303: The [package](https://reactjs.org/blog/2017/09/26/react-v16.0.html#packaging) changed in React 16 so importing `*` makes more things work as expected.

src/ui/components/Notice/styles.scss
kumar303: The [spec](https://design.firefox.com/photon/components/message-bars.html) says it should be 4px.

tofumatt:
Oh, I see. I hadn’t seen that part of the spec.

I definitely don’t like that 😅 But you’re right: it’s part of the spec. I’ll bring it up with people but you’re right: you should follow the spec for now.

src/ui/components/Notice/styles.scss
tofumatt: Ah, okay, that's fine then. 👍

kumar303:
I updated the multi-line message based on a "spec" that @pwalm just gave me 😃 I think this looks better (I adjusted the screenshots). 
img_1255

kumar303:
Tests + lint all pass for me locally

PR #4153, opened 2018-01-13
chore: Add config option to disable version checks in search

### LINKED ISSUE #4152: Non-production sites return no results for 57+ browsers
We (largely, helpfully) send an `appversion` param when searching for add-ons to restrict search results to add-ons that are compatible with the user's UA. This works well in production but causes a lot of strife when using docker or -dev/-stage environments, where the data is out-of-date and nearly all add-ons are not marked as 57+ compatible.

We need a config option that allows us to omit the version param during development so we can get useful results when developing.

tofumatt:
Fixes #4152

Adds a config option (set to `true` for normal environments and `false` for development environments) that ignores version compatibility checks. This means we get add-ons returned in searches when using -stage data.

### Before

screenshot 2018-01-13 00 33 30

### After

screenshot 2018-01-13 00 08 44

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -43,6 +43,13 @@ module.exports = {
   isDeployed: true,
   isDevelopment: false,
 
+  // For all Firefox 57+ (Quantum) UAs, send the `appversion` in all API
+  // search requests. This will return only compatible add-ons, which is
+  // good for UX (it prevents a lot of incompatible old add-ons).
+  // Disable this in development when working with stage data, which is
+  // very out-of-date and mostly not 57+ compatible.
+  restrictSearchResultsToAppVersion: true,
+
   // The canonical list of enabled apps.
   validAppNames,
 
@@ -77,6 +84,7 @@ module.exports = {
     'apiHost',
     'apiPath',
     'appName',
+    'restrictSearchResultsToAppVersion',
     'cookieMaxAge',
     'cookieName',
     'cookieSecure',
--- a/config/development-amo.js
+++ b/config/development-amo.js
@@ -3,6 +3,11 @@ module.exports = {
   proxyApiHost: 'http://olympia.dev',
   proxyPort: 3000,
   proxyEnabled: true,
+  // Setting this to false returns add-ons that are not compatible but means
+  // developers can pull from a much larger dataset on the local/-dev/-stage
+  // servers. Set this to true to only get compatible add-ons (this is what
+  // happens in production) but get a lot fewer add-ons in search results.
+  restrictSearchResultsToAppVersion: false,
   fxaConfig: 'local',
   trackingEnabled: false,
 };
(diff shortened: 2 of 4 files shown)


config/default.js
```diff
@@ -30,6 +30,13 @@ module.exports = {
   // Disables the server side render, handy for debugging.
   disableSSR: false,
 
+  // For all Firefox 57+ (Quantum) UAs, send the `appversion` in all API
+  // search requests. This will return only compatible add-ons, which is
+  // good for UX (it prevents a lot of incompatible old add-ons).
+  // Disable this in development when working with stage data, which is
+  // very out-of-date and mostly not 57+ compatible.
+  checkForVersionCompatiblityInAPIRequests: true,
```
kumar303: This seems to only apply to search so I think it would help to indicate that by name. I would suggest:

```js
restrictSearchResultsToAppVersion: true
```

src/core/searchUtils.js
```diff
@@ -57,10 +59,16 @@ export function addVersionCompatibilityToFilters({
...
     if (browserVersion >= 57) {
-      log.debug(oneLine`Setting "compatibleWithVersion" to current application
+      if (config.get('checkForVersionCompatiblityInAPIRequests')) {
+        log.debug(oneLine`Setting "compatibleWithVersion" to current application
         version (Firefox ${browserVersion}) so only relevant extensions are
         displayed.`);
-      newFilters.compatibleWithVersion = userAgentInfo.browser.version;
+        newFilters.compatibleWithVersion = userAgentInfo.browser.version;
+      } else {
+        log.debug(oneLine`checkForVersionCompatiblityInAPIRequests config set;
```
kumar303: I think `log.warn()` would be more helpful here since disabling a production feature (while in development) is risky.

config/development-amo.js
```diff
@@ -1,5 +1,10 @@
 module.exports = {
   apiHost: 'http://localhost:3000',
+  // Setting this to false returns add-ons that are not compatible but means
+  // developers can pull from a much larger dataset on the local/-dev/-stage
+  // servers. Set this to true to only get compatible add-ons (this is what
+  // happens in production) but get a lot fewer add-ons in search results.
+  checkForVersionCompatiblityInAPIRequests: false,
```
kumar303: Is this absolutely necessary? I would rather see this as a config option that we explicitly set in our local servers only. I am just worried that dev will stray too far from production by disabling this feature.

tofumatt: I thought the `dev` config was for the `dev` server and this config would only affect local development. If I'm wrong on that this wasn't intended and yeah I only want this on local dev.

tofumatt: To clarify: I wanted it enabled by default because otherwise the data set is near useless for development.

I agree: I'd rather us clean up the data on the servers, but I know that's a lot to ask of the server devs so until then I'd rather fix it this way for anyone doing front-end development and not require more docs/manual work to set the config value.

kumar303: Ah, whoops, I was mistaken. Yes, this only applies to local development which I was asking for so carry on 🌴 🌞

PR #4159, opened 2018-01-16
chore: Create generic taarParams object for #4155

### LINKED ISSUE #4155: Pass additional parameters through to the API
TAAR is starting to use some new query string parameters in the disco pane so we should pass through `study` and `branch` as well as the `clientId`.

To make this easier in the future I'm wondering if we should just add a list for params that are OK to pass through, that way we just need to add to the list to send additional parameters.

tofumatt:
fix #4155.

See the params being sent:

screenshot 2018-01-16 13 05 08

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -26,6 +26,7 @@ module.exports = {
     'langMap',
     'publicSentryDsn',
     'rtlLangs',
+    'taarParamsToUse',
     'trackingEnabled',
     'trackingId',
     'trackingSendInitPageView',
@@ -51,6 +52,10 @@ module.exports = {
       mediaSrc: [staticHost],
     },
   },
+  taarParamsToUse: [
+    'platform',
+    'telemetry-client-id',
+  ],
   trackingEnabled: true,
   trackingId: 'UA-36116321-7',
   // We override the initial page view call in order to
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -53,8 +53,10 @@ module.exports = {
     },
   },
   taarParamsToUse: [
+    'branch',
+    'clientId',
     'platform',
-    'telemetry-client-id',
+    'study',
   ],
   trackingEnabled: true,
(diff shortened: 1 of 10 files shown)


config/default-disco.js
```diff
@@ -51,6 +52,12 @@ module.exports = {
       mediaSrc: [staticHost],
     },
   },
+  taarParamsToUse: [
+    'branch',
+    'platform',
+    'study',
+    'telemetry-client-id',
```
muffinresearch: Might be worth a comment here too to say that this is translated to this from clientId.

src/disco/containers/DiscoPane.js
```diff
@@ -67,10 +67,19 @@ export class DiscoPaneBase extends React.Component {
     // instead of an empty list because an empty list could be a valid
     // (yet unlikley) API response.
     if (!errorHandler.hasError() && !results.length) {
+      // We accept all query params here and filter them out based on the
+      // `taarParamsToUse` config value. See:
+      // https://github.com/mozilla/addons-frontend/issues/4155
+      const taarParams = { platform: params.platform, ...location.query };
+      // We translate `clientId` to `'telemetry-client-id'`.
+      if (taarParams.clientId) {
```
muffinresearch: I wonder if this would be better moved to where the filtering happens?  Also it's worth noting this implicitly allows telemetry-client-id as a query param too, not that this probably matters.

Also it looks like `platform` from the URL could be clobbered by `platform` as a query param. You'd need to reverse the order to prevent that. Would be good to add a test for that case.

config/default-disco.js
muffinresearch: If you move the mapping to where the filtering happens this config could specify `clientId` instead, which might be easier to understand.

I'm not sure about the merging of the query params with the bits extracted from the url string e.g. `branch` is a query param wheras `platform` comes from the URL, it might be easier to understand if those are kept separate.

PR #4160, opened 2018-01-16
Add a `Reviewer Tools` link to the user dropdown if they have permission

diox:
Fix mozilla/addons-server/issues/7280

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -10,8 +10,8 @@ import SectionLinks from 'amo/components/SectionLinks';
 import AuthenticateButton, {
   createHandleLogOutFunction,
 } from 'core/components/AuthenticateButton';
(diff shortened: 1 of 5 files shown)


src/amo/components/Header/index.js
```diff
@@ -160,9 +182,18 @@ export const mapStateToProps = (state) => {
     api: state.api,
     isHomePage: state.viewContext.context === VIEW_CONTEXT_HOME,
     siteUser: getCurrentUser(state.users),
+    hasAnyReviewerRelatedPermission:
```
tofumatt: This seems like it's be a good helper (or possibly even an attribute to set on the user inside the reducer).

At the very least could we abstract this into a helper?

diox: Yeah, I wasn't sure initially, but you're right, I'll do that.

Issue #4169, opened 2018-01-17
Make adjustments to Photon notifications


kumar303:
Follow up to https://github.com/mozilla/addons-frontend/issues/3706

Verified on AMO-dev (Win10, Android 7.0) and here are my observations:
1. The suggestion was to have the notifications displayed at the top of the current block, but I see that  the position hasn't changed
2. ~~The [specs](https://design.firefox.com/photon/components/message-bars.html) suggest that the information icon for error messages should be a **`!`**, but we have here an **`i`**~~ See https://github.com/mozilla/addons-frontend/issues/4175
3. The error text was to be white, however links included in this text are blue and not very legible
4. The notification text doesn't seem to be perfectly aligned inside the notification banner
5. At page refresh, the notification banners changes sizes
6. ~~The 'Download Firefox' banner was not added (not so sure if it was intended or not for this patch)~~ See https://github.com/mozilla/addons-frontend/issues/4176

Screenshots:
![image](https://user-images.githubusercontent.com/31961530/35044605-c0fcee34-fb99-11e7-8ec8-9bbc22a652cc.png)
![refresh](https://user-images.githubusercontent.com/31961530/35044619-ca080dec-fb99-11e7-8047-5103aa1c511a.gif)
</br>
![image](https://user-images.githubusercontent.com/31961530/35044790-5567ae4c-fb9a-11e7-998b-cb807d5d0866.png)

AlexandraMoga:
Verified as fixed on AMO-dev with FF57 (Win10, Android 7.0)

![image](https://user-images.githubusercontent.com/31961530/35105980-b4eada20-fc75-11e7-83c1-8d80141507a7.png)
</br>
![image](https://user-images.githubusercontent.com/31961530/35106015-ca852642-fc75-11e7-9666-ec130505d34f.png)
</br>
![pagerefr](https://user-images.githubusercontent.com/31961530/35106028-d46eaf5c-fc75-11e7-9b2c-f7dc16652230.gif)

kumar303:
Thanks again for your help on this @AlexandraMoga . I picked the patch for today's release so that we'll be releasing the corrected notifications. https://public.etherpad-mozilla.org/p/amo-2018.01.18

AlexandraMoga:
@kumar303 I'm glad you considered this helpful!

PR #4177, opened 2018-01-17
Fix Photon notification issues

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4169 (this issue contains a list of what was fixed).

Here is the updated link style as well as a fix for the vertical centering of the message:

screenshot 2018-01-17 17 27 28

I could only reproduce the reload-and-it-jumps bug on dev, not locally. However, while inspecting the markup I think the culprit was putting a `<div>` inside of a `<p>` tag which seems to have created wonky HTML on the server only:

screenshot 2018-01-17 17 16 51

### DIFF
--- a/src/ui/components/Notice/index.js
+++ b/src/ui/components/Notice/index.js
@@ -49,10 +49,12 @@ const Notice = ({
     <div className={finalClass}>
       <div className="Notice-icon" />
       <div className="Notice-column">
-        <p className="Notice-text">
-          {children}
-        </p>
-        {actionButton}
+        <div>
+          <p className="Notice-text">
+            {children}
+          </p>
+          {actionButton}
+        </div>
       </div>
     </div>
   );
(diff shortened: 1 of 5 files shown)


src/ui/components/Notice/index.js
```diff
@@ -49,10 +49,12 @@ const Notice = ({
     <div className={finalClass}>
       <div className="Notice-icon" />
       <div className="Notice-column">
-        <p className="Notice-text">
-          {children}
-        </p>
-        {actionButton}
+        <div>
```
kumar303: This is the best idea I could come up with but I'm open to alternatives. My intention was to use flex in the container for vertical centering but continue relying on inline block display rules to render the message and action button correctly (i.e. not apply flex to the message and the button).

tofumatt: I think it's fine 😄

PR #4178, opened 2018-01-18
Upgrade eslint configs

muffinresearch:
Closes #4174 
Closes #4166 

Assuming this looks good I'll file some follow-up issues to take a look at switching away from`isNaN` to `Number.isNaN` but it'll make sense to split that out, there's some subtle differences we'll need to check out.

There's also some JSX a11y rules that we should take a closer look at e.g. there might be some cases where it was valid that it complained about using a link when it should be a button.

Other notes:

`prefer-destructing` generally works well, however it does need to be skipped when there's no const/let otherwise that causes a syntax error.


muffinresearch:
Last commit was to fix this error with flow:check:

```
$ yarn flow:check
yarn run v1.3.2
$ flow check
Error: src/ui/components/HeroSection/index.js:30
 30:           `HeroSection-styleName--${styleName}`,
                                         ^^^^^^^^^ undefined. This type cannot be coerced to
 30:           `HeroSection-styleName--${styleName}`,
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string
```

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -304,9 +304,11 @@ export function mapStateToProps(
   if (ownProps.review) {
     const view = state.reviews.view[ownProps.review.id];
     if (view) {
+      /* eslint-disable prefer-destructuring */
```
kumar303: I would rather not have to use this comment for assignments everywhere. I think you can turn this off with [`{AssignmentExpression: {array: false, object: false}}`](https://eslint.org/docs/rules/prefer-destructuring). Did you try that?

muffinresearch: Cool, I'll give that a spin.

Issue #4017, opened 2017-11-30
Polyfill request animation in the test suite


kumar303:
We load a polyfill in the web application for React 16 (search for `import 'raf/polyfill'`) but not in the test suite. This litters the console with warnings. We should add that polyfill to the test suite, maybe in `tests/setup.js`.

<img width="982" alt="screenshot 2017-11-30 16 53 16" src="https://user-images.githubusercontent.com/55398/33459363-514951a6-d5ef-11e7-8c53-f2c3458f1ed9.png">

tofumatt:
Good call, thanks for noticing. The tests since moving to Jest have been so noisy that I run them with the --silent argument most of the time, so I missed this entirely. 😓

willdurand:
Why would the reducers yield such warnings though? 🤔

kumar303:
> The tests since moving to Jest have been so noisy...

Yeah, when I get time I'm going to create a console reporter plugin to accomplish this: https://github.com/facebook/jest/issues/4156

kumar303:
This depends on a new polyfill module introduced by https://github.com/mozilla/addons-frontend/pull/4119

I'll fix this when that patch lands.

kumar303:
A few of the tests in https://github.com/mozilla/addons-frontend/pull/4182 kept failing with out of memory errors. I guess the Travis VM doesn't perform well with the polyfill. Instead of spending a lot of time investigating I'm just going to close this for now. Maybe there's another way to make the React warnings go away.

PR #4182, opened 2018-01-18
Add polyfills to test suite just the same as the prod app

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4017

### DIFF
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -1,3 +1,6 @@
+// This adds the same polyfills that the production application relies on.
+import 'core/polyfill';
+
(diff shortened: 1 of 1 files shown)


kumar303:
A couple test runs failed with 'JavaScript heap out of memory'. Hmm. Maybe we can't add a polyfill.

kumar303:
Bah, another out of memory error. Looks like Travis doesn't like the polyfill so I'm closing this for now. I'm not sure it's worth the time to investigate.

PR #4199, opened 2018-01-22
chore: Update homepage content

### LINKED ISSUE #4163: 25 January homepage content programming
For the **Thursday, 25 January** push, please update the AMO homepage programming in accord with what's outlined [here](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#heading=h.e9p8yhnuv39m).

@EnTeQuAk @muffinresearch @diox

tofumatt:
Close #4163.

### Before

screenshot 2018-01-22 18 34 04

### After

screenshot 2018-01-22 18 41 13

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -24,23 +24,20 @@ import Icon from 'ui/components/Icon';
 import './styles.scss';
 
 
-export const FIRST_COLLECTION_SLUG = 'be-more-productive';
+export const FIRST_COLLECTION_SLUG = 'dynamic-media-downloaders';
 export const FIRST_COLLECTION_USER = 'mozilla';
-
-export const SECOND_COLLECTION_SLUG = 'privacy-matters';
-export const SECOND_COLLECTION_USER = 'mozilla';
 
 export class HomeBase extends React.Component {
   static propTypes = {
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
     firstCollection: PropTypes.array.isRequired,
-    secondCollection: PropTypes.array.isRequired,
     featuredExtensions: PropTypes.array.isRequired,
+    featuredThemes: PropTypes.array.isRequired,
     i18n: PropTypes.object.isRequired,
     popularExtensions: PropTypes.array.isRequired,
     resultsLoaded: PropTypes.bool.isRequired,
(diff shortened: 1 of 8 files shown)


muffinresearch:
Not sure it came in via this patch but would be nice to fix these duplicated headings:

add-ons_for_firefox

tofumatt:
Yeah... I don't think that's related to this patch and I think in general this homepage code is a bit... clunky. I think it could use a refactor in general. 🤔

muffinresearch:
> Yeah... I don't think that's related to this patch and I think in general this homepage code is a bit... clunky. I think it could use a refactor in general. 🤔

Probably something to leave until we connect up to the admin tools.

muffinresearch:
It seems possible now to have the same colour in multiple places:

add-ons_for_firefox

tofumatt:
> It seems possible now to have the same colour in multiple places:

Looks like the wrong screenshot was attached 😅

muffinresearch:
> Looks like the wrong screenshot was attached 😅

Yeah skitch was doing weird things - should be good now.

tofumatt:
Ah, good catch. That might've even been possible before but I've tweaked around the colours now and it should be fine.

muffinresearch:
Looks like there's an escaping issue on the collections page, the api appears to provide it to us that way:

See: https://addons.allizom.org/api/v3/accounts/account/mozilla/collections/dynamic-media-downloaders/

dynamic_media_downloaders_ _add-ons_for_firefox

muffinresearch:
r+wc

tofumatt:
Ah, that’s a bit weird... I don’t think we should treat that field as safe/escaped as it’s user input.

- Matt (Sent from mobile)

muffinresearch:
> Ah, that’s a bit weird... I don’t think we should treat that field as safe/escaped as it’s user input.

Yeah I think it's an API bug

muffinresearch:
@tofumatt as a workaround we could use the word "and" for that collection.

muffinresearch:
I've filed https://github.com/mozilla/addons-server/issues/7368 for the escaping

tofumatt:
I'm cool with changing the collection description but I'm not sure I have access to it. I just tried to change it and couldn't see where to 😢

tofumatt:
I'm gonna merge this then as that bug is unrelated...

Issue #4194, opened 2018-01-22
API helpers for editing a collection


kumar303:
We need API helpers to facilitate editing a collection https://github.com/mozilla/addons-frontend/issues/4002

For QA: please check for regressions in any of the existing collection pages

AlexandraMoga:
We looked for regressions around the existing collection functions and everything seems to work as expected.

PR #4201, opened 2018-01-22
API helpers for editing a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4194

### DIFF
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -8,6 +8,7 @@ import { schema as normalizrSchema, normalize } from 'normalizr';
 import { oneLine } from 'common-tags';
 import config from 'config';
 
+import languages from 'core/languages';
 import { initialApiState } from 'core/reducers/api';
 import log from 'core/logger';
 import {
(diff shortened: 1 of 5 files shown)


src/core/api/index.js
```diff
@@ -305,3 +306,15 @@ export const allPages = async (
   // If we get this far the callback may not be advancing pages correctly.
   throw new Error(`Fetched too many pages (the limit is ${pageLimit})`);
 };
+
+export const validateLocalizedString = (localizedString: LocalizedString) => {
```
tofumatt: According to the API docs, if we're sending the `lang` parameter along with the request (which we should be), we shouldn't be getting an object back, just a string in the correct locale. (http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#translated-fields)

Is this because we are sending multiple translated fields?

tofumatt: Ah, just read below, looks like yes. Okay, cool!

tests/unit/amo/api/test_collections.js
```diff
@@ -287,4 +288,95 @@ describe(__filename, () => {
       });
     });
   });
+
+  describe('updateCollection', () => {
+    const _params = (params = {}) => {
```
tofumatt: We usually call these `defaultParams()` and I think that's a bit more instructive.

src/core/api/index.js
tofumatt: Technically though, we're allowed to send a single string to the API if we supply the locale... and I think we shouldn't be supplying the locale (which I'm pretty sure `callApi` does automatically) if we're sending an object.

Issue #4196, opened 2018-01-22
Saga helpers for editing a collection


kumar303:
We need saga helpers to facilitate editing a collection https://github.com/mozilla/addons-frontend/issues/4002

For QA: please check for regressions in any of the existing collection pages

AlexandraMoga:
We've looked for regressions around collection pages but everything seems to be working fine. Marking this as fixed.

Issue #4002, opened 2017-11-28
Edit collection name and description


kumar303:
Tracker: https://github.com/mozilla/addons-frontend/issues/2782

Depends on https://github.com/mozilla/addons-frontend/issues/4194 https://github.com/mozilla/addons-frontend/issues/4196 https://github.com/mozilla/addons-frontend/issues/4198

A user should be able to:
- Sign in
- View an existing collection that they created (the classic site enables creation)
- Edit the name or description. Editing other the other parts of the collection (as seen in the mock) will come in future patches.

Notes:
- We will release this behind a flag and only enable on the dev server. This is because the feature will be incomplete and we want to let users continue on the old site until it's fully ready.

<img width="685" alt="screenshot 2018-01-22 11 21 04" src="https://user-images.githubusercontent.com/55398/35234388-8f9d3152-ff66-11e7-9bb1-27626eb7c463.png">

pwalm:
Edit collection modals (with "edit custom link" toggled on/off):
![screen shot 2017-12-04 at 10 41 42 am](https://user-images.githubusercontent.com/6797635/33569832-c9d2f4f4-d8df-11e7-81f7-1d72dead99b4.png)
![screen shot 2017-12-04 at 10 42 29 am](https://user-images.githubusercontent.com/6797635/33569864-dadcff1a-d8df-11e7-8e6d-d377686bbfe3.png)

[Collections screens here](https://mozilla.invisionapp.com/share/YCCYEOFRH).

ioanarusiczki:
I verified this issue on AMO dev with FF58 (Win10). The buttons work as expected, click outside the box will close the popup, or using "X" button, the user can edit name and description multiple times.
I updated for the observed layout issues #4185.

PR #4209, opened 2018-01-24
Saga helpers for editing a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4196

I envision this new saga, `updateCollection()`, to run when the user presses the submit button ([see mock](https://github.com/mozilla/addons-frontend/issues/4002)). It will do the following:
- Change state to *updating* so that the submit button can be disabled
- Send a patch to the collection API
- Report any API errors and reset state accordingly


src/amo/reducers/collections.js
```diff
@@ -73,6 +76,9 @@ export type CollectionsState = {
       |};
     },
   },
+  collectionUpdates: {
+    [collectionSlug: string]: {| updating: boolean, success?: boolean |},
```
tofumatt: `successful` might be a slightly better flag name and might make the code read a bit nicer, as I'm assuming this flags that an update was successful.

Issue #4179, opened 2018-01-18
Switch from isNaN to Number.isNaN


muffinresearch:
We should consider switching our use of the global `isNaN` to `Number.isNaN` - if we make this change we should make sure that we understand the subtle differences related to coercion in the global function in case we are currently reliant on that behaviour.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

See https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=isNaN&type= for all instances.

KhushilMistry:
Hey, I want to work on this issue.
Will make a pull request as soon as possible.

muffinresearch:
@KhushilMistry ok thank! I'll mark this contrib:assigned, please reference this issue in the PR.

willdurand:
@muffinresearch how does that `Number.isNaN` behave on the server?

KhushilMistry:
Hey, I have made a PR #4214. Can you review this ?

tofumatt:
> Hey, I have made a PR #4214. Can you review this ?

Sure thing, we get notifications of new PRs so someone from the team will take a look next chance we get. There's no need to notify us in the issue 😄

kumar303:
In all cases, we use `isNaN()` to check whether a string is numeric or not. On further inspection, `Number.isNaN()` is not intended for this case. It seems intended for checking if a value is literally `NaN`.

Actually, this led me to discover a bug: https://github.com/mozilla/addons-frontend/issues/4218

I'm closing this for now because I can't think of a case where we need `Number.isNaN()`.

kumar303:
Whoops, the bug I said I discovered was not a bug.

However, I think it would be risky to try and port some things to `Number.isNaN()` and not others. When glancing at our usage of `isNaN()`, it will only be safe to use `Number.isNaN()` in one case. Feel free to re-open if someone thinks we should still pursue it.

willdurand:
FWIW, Airbnb suggests to use `Number.isNaN` only: https://github.com/airbnb/javascript#standard-library--isnan.

PR #4214, opened 2018-01-24
Changed isNaN to Number.isNaN

KhushilMistry:
Fixes Issue #4179

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -103,7 +103,7 @@ export class AddonBase extends React.Component {
     if (!errorHandler.hasError()) {
       if (addon) {
         // eslint-disable-next-line no-restricted-globals
-        if (!isNaN(params.slug)) {
+        if (!Number.isNaN(params.slug)) {
           // We only load add-ons by slug, but ID must be supported too because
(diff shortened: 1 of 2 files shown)


kumar303:
Hi, thanks for getting this started. You have unfortunately picked up a rock only to reveal some bugs crawling underneath 🐞 

I believe the tests are failing because something is `undefined` that shouldn't be. You can see that `isNaN()` and `Number.isNaN()` both handle `undefined` differently:

screenshot 2018-01-24 16 00 06

You'd have to track down what exactly is `undefined`.

As an aside, I really have no idea why `Number.isNaN()` returns `false` for `undefined`. It makes me wonder if we shouldn't be using it.

kumar303:
@KhushilMistry thanks for bringing this to my attention 😄 I have done some research and we should not be using `Number.isNaN()` for any code that is trying to check whether a string is numeric or not.

I am closing this because I've opened a new issue to do the opposite: https://github.com/mozilla/addons-frontend/issues/4218 Feel free to work on that one if you can.

kumar303:
Actually, you can ignore https://github.com/mozilla/addons-frontend/issues/4218 -- it's not a valid bug.

PR #4216, opened 2018-01-24
fix: Reduce homepage CSS (fix #4211)

### LINKED ISSUE #4211: Uncompressed CSS on stage is 5.84mb vs 1.75mb on prod
Just noticed this checking out the cache immutable changes.

I'm not sure what's changed, but we should check this out before release in case there's a been any changes resulting in a lot more CSS being generated etc.

tofumatt:
Wow. So looking into this exposed how brutal our (admittedly more flexible) homepage CSS setup was for file size.

Last week's tag (and similar releases since the introduction of our homepage banners) had a CSS file size of `1.8MB`. This week's tag, with extra numbers of possible banners, because of how we created them, had a CSS size of `6.1MB`.

After this fix, our CSS file is down to ~`500kB`. So yeah. This reduces the dynamic backgrounds/colours found in the old homepage, but it cuts down on maintenance effort in updating the homepage and file size. I think it's worth it.

### DIFF
--- a/src/amo/components/HomeHeroBanner/index.js
+++ b/src/amo/components/HomeHeroBanner/index.js
@@ -23,7 +23,6 @@ export class HomeHeroBannerBase extends React.Component<Props> {
         <HeroSection
           key="wikipedia-context-menu-search"
           linkTo="/addon/wikipedia-context-menu-search/"
-          styleName="Home-wikipedia-context-menu-search"
         >
           <h3>{i18n.gettext('Wikipedia Context Menu Search')}</h3>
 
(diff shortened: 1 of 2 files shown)


willdurand:
Great 👍  So now patterns are fixed and colors are randomized per item?

kumar303:
It looks like everything is fixed (colors and patterns) but I'm not sure if that was intended. I'm still reviewing the code locally.

tofumatt:
The patterns are now fixed. The colours are grouped (so that blues, greens, and violets are separate) such that at build time they’re random. It’s certainly much less random but it’s a lot less filesize. I suppose we could restore some amount of randomness, at least in the pattern, in a future patch. But I think the filesize reduction is worth the loss… especially when UX is talking about replacing this component in the future anyway.

kumar303:
> I think the filesize reduction is worth the loss

I agree. I always found the old code confusing and the looping made me nervous. I would like it if we could avoid loops in Sass code because it's always hard to comprehend the O(N...) matrix we are creating.

kumar303:
Looks like the UI tests failed due to a network error so I restarted them.

Issue #4226, opened 2018-01-25
The first item in the Add to collection menu should be disabled


kumar303:
### Describe the problem and steps to reproduce it:

Go to any add-on detail page. Scroll to *Add to collection* and click on the select box to open the menu.

### What happened?

The first item is *Select a collection…* which looks like any other actionable item.

### What did you expect to happen?

This first item is not an actionable option. It should be disabled but it's not:

![image](https://user-images.githubusercontent.com/31961530/35381852-3fca1cbe-01c6-11e8-96fb-b9177b59f427.png)

AlexandraMoga:
I'm afraid this didn't help much because now the "Create new collection" option is not working anymore:

![coll](https://user-images.githubusercontent.com/31961530/35436938-083b15cc-0299-11e8-92e0-b747b01e939e.gif)

muffinresearch:
FWIW I don't think I have ever seen the "Please select..." option of a select be disabled, I don't think it needs to be. It's a pretty long-standing convention that the first item just has an empty value.

muffinresearch:
Maybe the answer to this could be to change the copy?

muffinresearch:
@kumar303 see my comments above, I'm not sure this is a bug - what do you think? Have you seen this pattern elsewhere?

PR #4227, opened 2018-01-25
Disable the add-to-collection prompt option in the menu

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4226

Before:

screenshot 2018-01-25 11 51 21

After:

screenshot 2018-01-25 11 50 48

### DIFF
--- a/tests/unit/amo/components/TestAddAddonToCollection.js
+++ b/tests/unit/amo/components/TestAddAddonToCollection.js
@@ -296,8 +296,9 @@ describe(__filename, () => {
       });
       expect(promptOption).toHaveProp('disabled', true);
 
-      // Just to cover the error handling code, try selecting the prompt
-      // (first option) which shouldn't do anything.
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAddAddonToCollection.js
```diff
@@ -294,8 +294,10 @@ describe(__filename, () => {
       const promptOption = findOption({
         root, text: 'Select a collection…',
       });
+      expect(promptOption).toHaveProp('disabled', true);
 
-      // Select the prompt (first option) which doesn't do anything.
+      // Just to cover the error handling code, try selecting the prompt
```
tofumatt: I think this comment is a bit oddly-worded and I don't exactly follow what it's documenting. It's being selected to make sure nothing happens, right? It's a bit vague, maybe this could be reworded... I think the original wording is actually a bit clearer.

PR #4228, opened 2018-01-25
Remove randomly generated style colors

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4229

The hash in the filename of the final stylesheet asset is generated from its contents. When we deployed a build to multiple server instances, they were each generating different file names.


tofumatt:
Sorry about this, didn't realise that would break things 😓

kumar303:
No worries, neither did I 😢

PR #4237, opened 2018-01-26
I4135 - submit rating popup font consistency

### LINKED ISSUE #4135: Add-on Review Pop-up: Font inconsistency

1. Go to any add-on
2. Navigate to 'Rate your experience'
3. Click on the stars for your rating

### What happened?
The font for Placeholder text, form text and submit review text is different from the rest of the site (which uses Fira Sans)

Have the same font as the rest of the pages

image

tsl143:
Fixes #4135 
Inconsistent fonts in submit review popup.

Before:
before

After
after

### DIFF
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -35,8 +35,6 @@
 
 .AddonReview-textarea,
 .AddonReview-submit {
-  @include font-regular();
-
   border: 2px solid $form-border-color;
   border-radius: 8px;
   box-shadow: none;
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -35,6 +35,8 @@
 
 .AddonReview-textarea,
 .AddonReview-submit {
+  @include font-regular();
+
   border: 2px solid $form-border-color;
   border-radius: 8px;
   box-shadow: none;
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonReview/styles.scss
```diff
@@ -35,6 +35,8 @@
 
 .AddonReview-textarea,
 .AddonReview-submit {
+  @include font-regular();
```
tofumatt: Is this font messed up for all textareas? If so, we should be changing this at the tag-level, not just here.

tsl143: The inputs throughout the site just have 'sans serif', example in search bar.

tsl143: I realized, we have input only at two places on complete website, one is submit review and other is search box, so shall we use `Fira sans`everywhere?

tsl143:
Oh got one more, Textarea is also used at report abuse, so making it global will help. PR is updated with global thing.

screen shot 2018-01-26 at 5 18 05 pm

tsl143:
@tofumatt it already has `Fire Sans` font as we are using input type `submit` not `button` there.

screen shot 2018-01-26 at 11 43 32 pm

tofumatt:
Oh sweet okay then, in that case this is good to go.

tofumatt:
Oh sweet okay then, in that case this is good to go.

PR #4238, opened 2018-01-26
Disallow random() in SASS

### LINKED ISSUE #4230: Prevent the use of random() in SASS files
Deployments rely on idempotent builds of the static content for the same set of source files. Idempotent builds should result in the same hash for the output file given the same set of input.

Using something that randomizes content at build time such as `random()` in SASS will result in a different hash for subsequent builds. (Of course you might luck out and get two builds that have the same result).

To fix this we should file a way to prevent `random()` being used in our SASS files. 

A stylelint rule to disallow it might be the way forward.

muffinresearch:
Fixes #4230

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -28,6 +28,7 @@
       "ignoreAtRules": ["if", "else"],
     }],
     "function-url-quotes": ["always"],
+    "function-blacklist": ["random"],
     "max-nesting-depth": [3, {
       "ignore": ["blockless-at-rules"],
     }],


muffinresearch:
> Maybe add a link to the issue where we discovered this so it's clear why we're doing this, but this is good. 

Yeah I thought about doing that, I assumed that .stylelintrc didn't allow comments but maybe it does have an allowance for that. I'll check.

kumar303:
Nice!

muffinresearch:
Unfortunately comments in .stylelintrc aren't allowed.

PR #4239, opened 2018-01-26
fix: Allow HTML entities in collection description

### LINKED ISSUE #4213: Assume HTML for collection descriptions
The API currently has a special case for translated content for collection descriptions. This process will handle any HTML in the descriptions. For descriptions with no HTML this may result in `&` being sent by the API as `&amp;`

To fix this we should use DangerouslySetInnerHTML + DOMPurify on collection descriptions, and we can set the allowed tags to an empty list since they shouldn't actually contain unescaped HTML.

Ideally in the future the code behind API wouldn't need to do this treatment of the content, but for now this is a pragmatic solution that gets us the desirable result with a small change on the frontend.

Example of the problem [1] and the API response [2]:

dynamic_media_downloaders_ _add-ons_for_firefox
 
[1] https://addons.allizom.org/en-US/firefox/collections/mozilla/dynamic-media-downloaders/
[2] https://addons.allizom.org/api/v3/accounts/account/mozilla/collections/dynamic-media-downloaders/

refs: https://github.com/mozilla/addons-server/issues/7368

tofumatt:
Fix #4213 

### Before
screenshot 2018-01-26 17 31 09

### After
screenshot 2018-01-26 17 26 07

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -18,7 +18,7 @@ import { withFixedErrorHandler } from 'core/errorHandler';
 import log from 'core/logger';
 import { hasPermission } from 'amo/reducers/users';
 import translate from 'core/i18n/translate';
-import { parsePage } from 'core/utils';
+import { parsePage, sanitizeHTML } from 'core/utils';
 import Card from 'ui/components/Card';
 import LoadingText from 'ui/components/LoadingText';
 import MetadataCard from 'ui/components/MetadataCard';
@@ -142,7 +142,11 @@ export class CollectionBase extends React.Component<Props> {
             {collection ? collection.name : <LoadingText />}
           </h1>
           <p className="Collection-description">
-            {collection ? collection.description : <LoadingText />}
+            {collection ? (
+              <span
+                dangerouslySetInnerHTML={sanitizeHTML(collection.description)}
+              />
+            ) : <LoadingText />}
           </p>
           <MetadataCard
             metadata={[
--- a/tests/unit/amo/components/TestCollection.js
+++ b/tests/unit/amo/components/TestCollection.js
@@ -66,6 +66,22 @@ describe(__filename, () => {
     expect(wrapper.find('.Collection-wrapper')).toHaveLength(1);
   });
 
+  it('allows HTML entities in the Collection description', () => {
+    const { store } = dispatchClientMetadata();
+
+    store.dispatch(loadCurrentCollection({
+      addons: createFakeCollectionAddons(),
+      detail: {
+        ...defaultCollectionDetail,
+        description: 'Apples &amp; carrots',
+      },
+    }));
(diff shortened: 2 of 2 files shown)


src/amo/components/Collection/index.js
```diff
@@ -141,9 +141,12 @@ export class CollectionBase extends React.Component<Props> {
           <h1 className="Collection-title">
             {collection ? collection.name : <LoadingText />}
           </h1>
-          <p className="Collection-description">
-            {collection ? collection.description : <LoadingText />}
-          </p>
+          {collection ? (
+            <p
+              className="Collection-description"
+              dangerouslySetInnerHTML={sanitizeHTML(collection.description)}
```
muffinresearch: My understanding of this field's output is that shouldn't be any unescaped tags in it, so you should be able to tell the sanitizer that no tags are allowed. If you include HTML in the collection description entry, it should appear escaped in the API data already.

tofumatt: `sanitizeHTML()` with only one argument means no tags are allowed, so that's what this is doing 😄

tests/unit/amo/components/TestCollection.js
```diff
@@ -66,6 +66,23 @@ describe(__filename, () => {
     expect(wrapper.find('.Collection-wrapper')).toHaveLength(1);
   });
 
+  it('allows HTML entities in the Collection description', () => {
+    const { store } = dispatchClientMetadata();
+
+    // We need a collection for this test case.
```
kumar303: I know you got this from another test but I don't think it's really helpful. It should be pretty obvious (from the code) that the test needs a collection.

src/amo/components/Collection/index.js
```diff
@@ -141,9 +141,12 @@ export class CollectionBase extends React.Component<Props> {
           <h1 className="Collection-title">
             {collection ? collection.name : <LoadingText />}
           </h1>
-          <p className="Collection-description">
-            {collection ? collection.description : <LoadingText />}
-          </p>
+          {collection ? (
+            <p
```
kumar303: I think it would be nicer if you changed this to `<span dangerouslySetInnerHTML={sanitizeHTML(collection.description)} />`.

That would allow you to keep wrapping it in a common container:

```js
<p className="Collection-description">
  {collection ? <span dangerouslySetInnerHTML={...} /> : <LoadingText />}
</p>
```

tests/unit/amo/components/TestCollection.js
tofumatt: Ah, indeed. Just a bad copypasta. Thanks 👍

src/amo/components/Collection/index.js
tofumatt: Ah, that's a good call. It increases the output of the HTML a bit but it's certainly nicer to write.

Issue #4086, opened 2017-12-18
Unbroken strings are not wrapped for Release Notes


AlexandraMoga:
Steps to reproduce:  

 1. Log into AMO and go to Developer Hub (i.e. https://addons-dev.allizom.org/en-US/developers/)
 2. Click one of your submitted add-ons
 3. In the  Manage Status&Versions section click the Version of your ad-on (such as https://addons-dev.allizom.org/en-US/developers/addon/swarrand/versions/1533356)
 4. Insert a long unbroken text in the Version Notes and Save Changes
 5. Click on View Listing and observe the Release Notes field

Expected result:
Text is wrapped 

Actual result:
Text is not entirely displayed and appears cut

Notes:
Reproduced on both AMO-stage and dev with FF57(Win 10)
Copied from addons-server [#7128](https://github.com/mozilla/addons-server/issues/7128#issuecomment-352232259)
![release notes](https://user-images.githubusercontent.com/33448286/33887271-ebab6552-df51-11e7-97a6-57c78c83886a.png)

ioanarusiczki:
This issue is verified on AMO dev and stage with FF 58 (Win10). 

![release notes](https://user-images.githubusercontent.com/33448286/35671986-aa4bdd1e-0745-11e8-9e67-c3038f02e06d.png)

PR #4248, opened 2018-01-28
I4086: word wrap for show more card contents

tsl143:
Fixes: #4086 

PS: word wrapping on `showmore-card-contents` so that this may not replicate at other places too

Before:
screen shot 2018-01-28 at 8 23 48 pm

After:
screen shot 2018-01-28 at 8 22 23 pm

### DIFF
--- a/src/ui/components/ShowMoreCard/ShowMoreCard.scss
+++ b/src/ui/components/ShowMoreCard/ShowMoreCard.scss
@@ -17,6 +17,7 @@
   max-height: 150px;
   overflow: hidden;
   position: relative;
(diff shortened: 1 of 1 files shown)


muffinresearch:
> That's good but we should add a CSS attribute that adds a hyphen when breaking words apart. I forget what it is but I think that exists... so it's clear a word is being broken apart.

@tofumatt Could we file a new issue for that assuming that is possible to save blocking on landing this? We don't currently do that anywhere else that we break long strings.

tofumatt:
Fair enough: yes 😄

Issue #4198, opened 2018-01-22
Add a new overlay for editing in context (like for collections)


kumar303:
We need an overlay component for editing or creating something in the context of something else. For example, you may need to create a collection from the collection listing screen. We have an existing overlay component but it has a lot of problems. See https://github.com/mozilla/addons-frontend/issues/3636

**For QA**: The actual overlay won't be testable until https://github.com/mozilla/addons-frontend/issues/4198 . For this issue, please check for style regressions on any page with forms (review entry, review replies, search box, etc).

Here is what we need in a new overlay component:

- Click the X to close it
- Click outside of it it close it
- Specify a header
- Specify the bottom row of buttons and their functionality
- Specify the content

Example:
<img width="685" alt="screenshot 2018-01-22 11 21 04" src="https://user-images.githubusercontent.com/55398/35235937-40795a88-ff6b-11e7-8a86-db80c4d7281f.png">

muffinresearch:
@kumar303 do we have details of how these would work on mobile - does the modal just essentially fill the screen?

kumar303:
@pwalm how should this work on mobile?

pwalm:
Great question. I'll have to design breakpoints for this UI, but that's something that needs to happen on quite a few screens (2018!!), so I'll do some work on that.

Short answer: that same, but thinner.

kumar303:
Mobile styles will be added in https://github.com/mozilla/addons-frontend/issues/4258

ioanarusiczki:
I verified this issue on AMO dev using FF58 (Win 10 and Android 8.0.0) Developer Response form changed the display for the Cancel button and Report this add-on for abuse form changed the display for Send abuse report button. They respond to click and look ok. 

The only inconsistency is in Write a review form ->  font size for Placeholder text changed on AMO dev it looks larger than on AMO stage and AMO prod.

![font size on dev](https://user-images.githubusercontent.com/33448286/35808942-34a97c44-0a90-11e8-9143-2bf183e5f1e9.png)
![font size on stage](https://user-images.githubusercontent.com/33448286/35808973-3ab38788-0a90-11e8-9ec6-8a57576462cc.png)

kumar303:
@ioanarusiczki Yes, this text change is fine. Thank you for verifying.

pwalm:
INLINE REVIEWS:
[...]

PR #4257, opened 2018-01-30
Form overlay for editing in context (like for Collections)

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4198

This adds a new overlay that will be used for editing collections (https://github.com/mozilla/addons-frontend/issues/4002).

I am testing it with this patch. After applying it, visit any collection detail page to see an overlay.

```diff
diff --git a/src/amo/components/Collection/index.js b/src/amo/components/Collection/index.js
index 9e09e030c..d3c7900a8 100644
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -20,6 +20,7 @@ import { hasPermission } from 'amo/reducers/users';
 import translate from 'core/i18n/translate';
 import { parsePage } from 'core/utils';
 import Card from 'ui/components/Card';
+import FormOverlay from 'ui/components/FormOverlay';
 import LoadingText from 'ui/components/LoadingText';
 import MetadataCard from 'ui/components/MetadataCard';
 import type {
@@ -169,6 +170,19 @@ export class CollectionBase extends React.Component<Props> {
               </Link>
             </p>
           )}
+          <FormOverlay
+            title={i18n.gettext('Edit collection')}
+            id={`collection-${collection ? collection.slug : ''}`}
+          >
+            <label htmlFor="collection_description">
+              {i18n.gettext('Collection name')}
+            </label>
+            <input
+              id="collection_description"
+              type="text"
+              placeholder={i18n.gettext('Description text?')}
+            />
+          </FormOverlay>
         </Card>
         <div className="Collection-items">
           <AddonsCard

```

Notes:
* I have not implemented mobile styles yet because we don't have mocks. In any case I'll do that in https://github.com/mozilla/addons-frontend/issues/4258
* The form styles will probably need more work but I'll do that in https://github.com/mozilla/addons-frontend/issues/4002

screenshot 2018-01-30 09 35 03
screenshot 2018-01-30 09 37 04


src/ui/components/FormOverlay/index.js
```diff
@@ -0,0 +1,157 @@
+/* @flow */
+/* eslint-disable react/sort-comp */
+import makeClassName from 'classnames';
+import * as React from 'react';
```
tofumatt: This reminded me to file: #4263

src/ui/components/FormOverlay/index.js
```diff
@@ -0,0 +1,157 @@
...
+|};
+
+export class FormOverlayBase extends React.Component<Props> {
+  static defaultProps = {
+    isOpen: true,
+  };
+
+  closeOverlay(event: SyntheticEvent<any>) {
+    const { id, dispatch } = this.props;
+    event.preventDefault();
```
tofumatt: Usually we just need `preventDefault()`. Is there a reason both are needed here? I believe it, but just would be cool to know why 😄

src/ui/components/FormOverlay/styles.scss
```diff
@@ -0,0 +1,62 @@
...
+.FormOverlay-overlay {
+  background-color: $white;
+  border-radius: $border-radius-s;
+  display: block;
+  margin: 12px;
+  min-width: 55%;
+  padding: 24px;
+}
+
+.FormOverlay-closer {
```
tofumatt: Given I'm not sure "closer" is a real noun this read awkwardly to me and I needed to do a double-take. `closeAction`, `closeButton`, etc. might be a bit more descriptive.

src/ui/components/FormOverlay/styles.scss
```diff
@@ -0,0 +1,62 @@
...
+
+.FormOverlay-form-buttons {
+  display: grid;
+  grid-gap: 16px;
+  grid-template-columns: min-content min-content;
+  justify-content: end;
+  margin: 36px 0 0;
+}
+
+.FormOverlay-x-closer {
```
tofumatt: Something like `closeButton-content` would be more descriptive and less directly-related to the content here; this class name as well made me stop and think a bit to figure out what it was.

tests/unit/ui/components/TestFormOverlay.js
```diff
@@ -0,0 +1,172 @@
...
+  it('closes on cancel', () => {
+    const id = 'some-id';
+    sinon.spy(store, 'dispatch');
+    const root = render({ id });
+    root.find('.FormOverlay-cancel').simulate('click', createFakeEvent());
+
+    sinon.assert.calledWith(store.dispatch, closeFormOverlay(id));
+  });
+
+  it('calls back on cancel', () => {
```
tofumatt: This might be better written as `'invokes cancel callback on cancel'`.

tests/unit/ui/components/TestFormOverlay.js
```diff
@@ -0,0 +1,172 @@
...
+    sinon.assert.called(onCancel);
+  });
+
+  it('handles an empty cancel callback', () => {
+    const root = render({ onCancel: undefined });
+    // This should not throw an error.
+    root.find('.FormOverlay-cancel').simulate('click', createFakeEvent());
+  });
+
+  it('calls back on submit', () => {
```
tofumatt: Similarly here: `'invokes submit callback on submit'` would read better.

src/ui/components/FormOverlay/index.js
kumar303: Since we have a lot of handlers all over the app for the same events (like clicks), it can be tricky to predict which handlers will fire as a result of event bubbling. As a precaution, I think we should always add `event.stopPropagation()` to stop bubbling until we know we need bubbling. More info: https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

kumar303:
Whoops, good catch. I fixed the bug where clicking the overlay was closing it. This is ready for another review.

tofumatt:
Looks good now. When I clicked on "Submit" nothing happened but I assume that's because there was no event handler in the example diff. Clicking the X, background, and cancel all closed it so seems to work well.

PR #4277, opened 2018-01-31
chore: Remove bundlesize devDep

### LINKED ISSUE #4275: Remove unused devDep bundlesize
We include bundlesize but aren't currently using it, so we should remove it as a dependency for now.

tofumatt:
Fix #4275.
Close #4273.


tofumatt:
Agreed, I would definitely like to restore it, but for now I figured we'd just keep it tidy and without greenkeeper PR noise.

Issue #2407, opened 2017-05-19
Configure webpack 2 to use tree-shaking for smallest bundle size


kumar303:
Now that we're on webpack 2 we can optimize the bundle by removing unused code via tree-shaking.

Webpack2 supports es modules directly so compiling to commonjs is no longer necessary.

See:
* https://webpack.js.org/guides/tree-shaking/
* https://blog.craftlab.hu/how-to-do-proper-tree-shaking-in-webpack-2-e27852af8b21

muffinresearch:
Worth noting from https://facebook.github.io/jest/docs/en/webpack.html#using-with-webpack-2

> webpack 2 offers native support for ES modules. However, Jest runs in Node, and thus requires ES modules to be transpiled to CommonJS modules. As such, if you are using webpack 2, you most likely will want to configure Babel to transpile ES modules to CommonJS modules only in the test environment.

willdurand:
We can probably do this now that we have babel 7, webpack 4 and a cleaner config.

willdurand:
Looks like there are issues with our current codebase, which prevents tree-shaking to be applied. For example, having `__filename` in our code is a problem (but we need it to generate unique IDs between the client and the server).

willdurand:
So, it looks like we have dead code elimination enabled and it works well. Tree-shaking is not enabled yet, though but it seems to be tricky to enable for some reasons.

The main reason is: CSS order is not respected in the final bundle, which causes issue. Another issue is that some polyfill functions get removed (like `regeneratorRuntime`) but I was able to bring them back. Everything is related to the `sideEffects` config parameter.

My last experiments show that with tree-shaking enabled (assuming CSS order gets fixed), the gain would be < 10KB.

willdurand:
It is also worth mentioning that in most docs, it is mentioned that we should set `modules: false` in `babel-preset-env`'s options. Yet, with Babel 7 (which we use), this is not needed anymore because the default value is `auto`, and in the context of Webpack, it does not generate CommonJS modules (i.e. it keeps the `import`/`export` statements, so that tree-shaking can work).

I have noticed NO difference with or without `modules: false`.

willdurand:
So. I thought I could setup tree shaking on our apps, but that's not working as intended for many reasons. I am re-stating what I wrote in #7042 below.

TL;DR: fully enabling tree shaking on our apps will not provide any immediate benefits.

[...]

PR #4286, opened 2018-01-31
chore: Update react imports (close #4263)

### LINKED ISSUE #4263: Move to new React imports (import * as React from 'react';)
In the future we should import React as `import * as React from 'react';` instead of our previous style (`import React from 'react';`). This should be doable as a simple find + replace 😄

tofumatt:
This'll be a massive diff but should merge pretty cleanly. Might as well do this now...

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -191,7 +191,7 @@ Here's an example of a basic component setup for translation:
 
 
 ```javascript
-import React from 'react';
+import * as React from 'react';
 import PropTypes from 'prop-types';
 
 import translate from 'core/i18n/translate';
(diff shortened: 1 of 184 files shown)


tofumatt:
Oh. That kind of sucks… especially as a big benefit of flow to me is checking stuff at build time instead of runtime saving us bundle size. Agreed it’s fine for now but that’s probably worth logging in the performance issue and addressing later...

kumar303:
If we implement tree shaking then the unused exports will go away. We totally should implement that. https://github.com/mozilla/addons-frontend/issues/2407

tofumatt:
Oh right, good call!

PR #4288, opened 2018-02-01
fix: Don't disable Search Plugins on the server (fix #4047)

### LINKED ISSUE #4047: 'Add to Firefox' button looks inactive on Search Tool details page
Steps to reproduce:
1. Go to AMO homepage and from the More... menu select Search Tools
2. Pick an option from the listed results to see its details page (i.e. https://addons-dev.allizom.org/en-US/firefox/addon/3-maps/)
3. Reload the page
4. Observe the 'Add to Firefox' button displayed on the Search tool details page

Expected result:
The Add to Firefox button is active and  blue colored 

Actual result:
The Add to Firefox button is grayed out and appears inactive on hover

Notes: when clicked, the button will trigger the installation process and the search plugin is added to the browser Search bar
Reproduced on AMO -dev and stage with FF57 (Win10)

search tools

tofumatt:
### Before

screenshot 2018-02-01 00 29 31

(Note the install button is disabled, and originally the page will load with an incompatibility notice)

### After

screenshot 2018-02-01 00 27 56

### DIFF
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -46,6 +46,7 @@ export function isCompatibleWithUserAgent({
   minVersion,
   userAgentInfo,
   _log = log,
+  _hasWindow = typeof window !== 'undefined',
   _window = typeof window !== 'undefined' ? window : {},
 } = {}) {
   // If the userAgent is false there was likely a programming error.
@@ -101,6 +102,11 @@ export function isCompatibleWithUserAgent({
 
     if (
       addon.type === ADDON_TYPE_OPENSEARCH &&
+      // If window isn't defined but the browser is Firefox, this is likely a
+      // server-side render, so we shouldn't condemn this search plugin to
+      // incompatibility yet, or it will stay disabled.
+      // See: https://github.com/mozilla/addons-frontend/issues/4047
+      _hasWindow &&
       !(_window.external && 'AddSearchProvider' in _window.external)
     ) {
       return { compatible: false, reason: INCOMPATIBLE_NO_OPENSEARCH };
--- a/tests/unit/core/utils/test_compatibility.js
+++ b/tests/unit/core/utils/test_compatibility.js
@@ -1,4 +1,3 @@
-import { oneLine } from 'common-tags';
 import UAParser from 'ua-parser-js';
 
 import {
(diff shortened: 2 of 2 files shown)


src/core/utils/compatibility.js
```diff
@@ -101,6 +102,11 @@ export function isCompatibleWithUserAgent({
 
     if (
       addon.type === ADDON_TYPE_OPENSEARCH &&
+      // If window isn't defined but the browser is Firefox, this is likely a
+      // server-side render, so we shouldn't condemn this search plugin to
+      // incompatibility yet, or it will stay disabled.
+      // See: https://github.com/mozilla/addons-frontend/issues/4047
+      _hasWindow &&
```
kumar303: This will probably work but it's a bit hacky (checking for `window` is not the real intent here). I think it would be better to use `config.get('server')` to check if we're rendering on the server or not.

src/core/utils/compatibility.js
```diff
@@ -101,6 +102,11 @@ export function isCompatibleWithUserAgent({
...
       !(_window.external && 'AddSearchProvider' in _window.external)
```
kumar303: I think you've uncovered a deeper bug here. It looks like React is rendering server HTML for the button component but not re-rendering it on the client (as it should). I forget exactly how this works but I think it has to do with the way React diffs prop values.

After your patch, the bug isn't fixed. Or, rather, there's a new bug where the button appears enabled for any Firefox client that does *not* define `window.external.AddSearchProvider()`. It should appear disabled.

I was testing it with a patch to simulate a missing `AddSearchProvider` API:

```diff
diff --git a/src/core/utils/compatibility.js b/src/core/utils/compatibility.js
index edfc29d5e..b23722140 100644
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -107,7 +107,7 @@ export function isCompatibleWithUserAgent({
       // incompatibility yet, or it will stay disabled.
       // See: https://github.com/mozilla/addons-frontend/issues/4047
       _hasWindow &&
-      !(_window.external && 'AddSearchProvider' in _window.external)
+      !(_window.external && 'AddSearchProviderNOPE' in _window.external)
     ) {
       return { compatible: false, reason: INCOMPATIBLE_NO_OPENSEARCH };
     }

```

screenshot 2018-02-01 14 22 25

I have a feeling this may be hard to fix 😄 If that's the case, I would suggest filing a bug to fix it later because it's not so important to break Firefox clients without `AddSearchProvider` -- do they even exist?

tofumatt: I think the only client that would identify itself as Firefox but not have the `AddSearchProvider` API is a REALLY old version (unsure) or Firefox for iOS. The latter we care more about and would be covered by a broader incompatibility check so it should be fine.

src/core/utils/compatibility.js
tofumatt: Oh right, good call.

muffinresearch:
I'm going to close this since it's bit-rotted but feel free to re-open it if you have time to update and/or finish it off.

kumar303:
Me, in the r- :

> I think you've uncovered a deeper bug here.

ZOMG I finally figured this out in https://github.com/mozilla/addons-frontend/pull/5450 It's a really annoying React "optimization" where css class names aren't updated in the DOM after a client render introduces new ones.

Issue #4300, opened 2018-02-02
FormOverlay should be closed by default


kumar303:
Since `FormOverlay` is controlled by Redux actions, it should default to a closed state.

This is a follow up to https://github.com/mozilla/addons-frontend/issues/4198

PR #4301, opened 2018-02-02
FormOverlay should be closed by default

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4300

### DIFF
--- a/tests/unit/ui/components/TestFormOverlay.js
+++ b/tests/unit/ui/components/TestFormOverlay.js
@@ -33,37 +33,51 @@ describe(__filename, () => {
     );
   };
 
+  const renderOpen = (customProps = {}) => {
+    const id = customProps.id || 'example-form-id';
+    store.dispatch(openFormOverlay(id));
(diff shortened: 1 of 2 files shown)


tests/unit/ui/components/TestFormOverlay.js
```diff
@@ -33,23 +33,29 @@ describe(__filename, () => {
...
 
     expect(root.find('.FormOverlay-h3')).toHaveText('Some Form');
   });
 
-  it('is open by default', () => {
+  it('is closed by default', () => {
     const root = render({ children: [<input key={1} />] });
 
-    expect(root.find('input')).toHaveLength(1);
+    expect(root.find('input')).toHaveLength(0);
```
tofumatt: Checking for `.FormOverlay` seems more accurate; it's likely there will be an input but we should check for the entire element. Sorry, I know this is mostly from a previous patch but this tweak made me notice it more 😄

PR #4302, opened 2018-02-02
Move Icon files to a sub directory

### LINKED ISSUE #4259: Move Icon files to a sub directory
The [Icon](https://github.com/mozilla/addons-frontend/tree/2d36a5e20268d51f921f10e5a764323bc6188855/src/ui/components/Icon) component has many icon files which are referenced by `styles.scss`. To reduce clutter, let's move those to a sub directory within the same component.

For QA: please check for regressions everywhere that we use icons, such as next/previous page arrows, error notices, search box magnifier, etc.

bobsilverberg:
Fixes #4259 

This moves the image files and updates the required styles.scss files.


kumar303:
When it's ready for another look, re-request a review from me. That will put it in my queue. The Github UI does not make this obvious which is why I'm pointing it out, heh.

src/ui/components/Icon/styles.scss
```diff
@@ -2,7 +2,7 @@
 @import "~core/css/inc/mixins";
 
 @mixin icon($name) {
-  background: url("./#{$name}.svg") center no-repeat;
+  background: url("./img/icons/#{$name}.svg") center no-repeat;
```
kumar303: I think you should put everything in `./img`, not `./img/icons/`. I really can't foresee any case where we'd need an extra subdirectory.

PR #4304, opened 2018-02-02
Update the favicon urls to directly point to the CDN

### LINKED ISSUE #3275: Update the favicon urls to directly point to the CDN
Refs: https://github.com/mozilla/addons-frontend/issues/3274

Hitting /favicon.ico  currently redirects to the CDN for each deployed environment.

This should be fixed by setting the url to point to the CDN relevant to each deployment.

PROD: https://addons.cdn.mozilla.net/favicon.ico
STAGE: https://addons-stage-cdn.allizom.org/favicon.ico
DEV:  https://addons-dev-cdn.allizom.org/favicon.ico

Since we'll be changing the host, we can drop the manual cache-busting query string param.

bobsilverberg:
Fixes #3275 

This uses the `amoCDN` key from the config object to create the favicon link in
ServerHtml.js.

### DIFF
--- a/tests/unit/core/containers/TestServerHtml.js
+++ b/tests/unit/core/containers/TestServerHtml.js
@@ -5,6 +5,7 @@ import {
   findRenderedDOMComponentWithTag,
   renderIntoDocument,
 } from 'react-dom/test-utils';
+import config from 'config';
 
 import ServerHtml from 'core/containers/ServerHtml';
 import FakeApp, {
@@ -113,7 +114,7 @@ describe('<ServerHtml />', () => {
   it('renders favicon', () => {
     const html = findRenderedDOMComponentWithTag(render(), 'html');
     const favicon = html.querySelector('link[rel="shortcut icon"]');
-    expect(favicon.getAttribute('href')).toEqual('/favicon.ico?v=1');
+    expect(favicon.getAttribute('href')).toEqual(`${config.get('amoCDN')}/favicon.ico`);
   });
 
   it('renders title', () => {
--- a/tests/unit/core/containers/TestServerHtml.js
+++ b/tests/unit/core/containers/TestServerHtml.js
@@ -5,12 +5,12 @@ import {
   findRenderedDOMComponentWithTag,
   renderIntoDocument,
 } from 'react-dom/test-utils';
-import config from 'config';
 
 import ServerHtml from 'core/containers/ServerHtml';
 import FakeApp, {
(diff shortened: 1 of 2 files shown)


tests/unit/core/containers/TestServerHtml.js
```diff
@@ -113,7 +114,7 @@ describe('<ServerHtml />', () => {
   it('renders favicon', () => {
     const html = findRenderedDOMComponentWithTag(render(), 'html');
     const favicon = html.querySelector('link[rel="shortcut icon"]');
-    expect(favicon.getAttribute('href')).toEqual('/favicon.ico?v=1');
+    expect(favicon.getAttribute('href')).toEqual(`${config.get('amoCDN')}/favicon.ico`);
```
kumar303: As is, this test relies on the [test configuration](https://github.com/mozilla/addons-frontend/blob/master/config/test.js) value of `amoCDN`. In practice that's fine but I think it would make the test more robust if you relied on something explicit. This will help with maintenance as well as with understanding the intention of the test. Here's an example of how it would look:

```js
import { getFakeConfig } from 'tests/unit/helpers';
...
const amoCDN = 'https://someserver.net';
const _config = getFakeConfig({ amoCDN });

const html = findRenderedDOMComponentWithTag(render({ _config }), 'html');
expect(favicon.getAttribute('href')).toEqual(`${amoCDN}/favicon.ico`);
```

This change requires you to set up `_config` as a property on `<ServerHtml>` for dependency injection. Here's an example:

```js
export class SomeComponent extends React.Component {
  static PropTypes = {
    _config: PropTypes.object,
  };

  static defaultProps = {
    _config: config,
  }

  ...
}
```

You then need to change your access of `config.get('amoCDN')` to something like:

```js
const { _config } = this.props;
...
_config.get('amoCDN');
```

tests/unit/core/containers/TestServerHtml.js
```diff
@@ -26,6 +27,9 @@ describe('<ServerHtml />', () => {
     getState: () => ({ foo: 'bar' }),
   };
 
+  const amoCDN = 'https://test.cdn.net';
```
kumar303: This is a small request but it would be nicer to move this into the function for the test that uses it to prevent any scope surprises. In other words, no other test needs to access it so it shouldn't go at the top of the file. In the future, if more than one tests needs access than we can move it higher up in scope. The same thing applies to `const _config = ...`.

PR #4309, opened 2018-02-05
fix: Do not overwrite button styles in tooltip lists

### LINKED ISSUE #4303: Text for the 'Log in to flag review' button placed in the Flag review section needs to be white
Follow up from [#2517](https://github.com/mozilla/addons-frontend/issues/2517)
STR:

1. Access AMO while logged out
2. Choose an add-on that has some reviews to view its details page
3. Go the add-on's reviews page and click on 'Flag' review for any of the listed reviews
4. Observe the text color of the 'Log in to flag review' button

Actual result:
Text color is black

Expected result:
Text color should be white

Reproduced on AMO dev FF58, (Win10, Android 7.0)

image

tofumatt:
Fixes #4303, by not overriding button styles inside these elements. The nesting on a lot of this CSS is pretty rough/complex by necessity and this was the most future-proof way to solve it for now. It could use future cleanup.

### Before
screenshot 2018-02-05 12 42 43
screenshot 2018-02-05 12 42 39

### After
screenshot 2018-02-05 12 39 02
screenshot 2018-02-05 12 38 58

### DIFF
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -36,9 +36,11 @@
   a:visited,
   a:hover,
   a:active {
-    color: $blue-50;
-    font-weight: normal;
-    text-decoration: none;
+    &:not(.Button) {
+      color: $blue-50;
+      font-weight: normal;
+      text-decoration: none;
+    }
   }
 }
 
--- a/src/ui/components/TooltipMenu/styles.scss
+++ b/src/ui/components/TooltipMenu/styles.scss
@@ -70,32 +70,39 @@
     a,
     a:link,
     button {
-      @include text-align-start();
-
-      color: $grey-90;
-      font-size: $font-size-m-smaller;
-      line-height: 1.6;
       overflow: hidden;
       text-overflow: ellipsis;
       white-space: nowrap;
       width: 100%;
+
+      &:not(.Button) {
+        @include text-align-start();
+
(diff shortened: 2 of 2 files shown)


src/amo/components/AddonReviewListItem/styles.scss
```diff
@@ -36,9 +36,11 @@
   a:visited,
   a:hover,
   a:active {
-    color: $blue-50;
-    font-weight: normal;
-    text-decoration: none;
+    &:not(.Button) {
```
kumar303: I don't see `button` in the selector list. Why was this needed?

src/ui/components/TooltipMenu/styles.scss
```diff
@@ -70,32 +70,39 @@
...
+        padding: 0;
+        text-decoration: none;
+      }
     }
 
     a:hover,
     a:focus,
     a:active {
-      color: $blue-50;
+      &:not(.Button) {
```
kumar303: Same question: I didn't see `button` in the selector list

src/amo/components/AddonReviewListItem/styles.scss
tofumatt: The buttons are `a` tags here (and in the other place) because they have a `href` attribute.

kumar303: Ah, the login button is an anchor. I didn't think of that.

Issue #4313, opened 2018-02-05
Prevent faking invalid config keys in tests


kumar303:
We use dependency injection to rely on config values in tests. We should prevent a test from accidentally relying on an invalid config value. Such a failure would indicate a typo or an old config key that had been since renamed.

This test made me think that the feature would be useful: https://github.com/mozilla/addons-frontend/pull/4304/files#diff-9c08b23edee72c13339e20d85241c391R116

AlexandraMoga:
@kumar303 is there anything to be done by QA here? This doesn't have an impact on AMO code as far as I can tell but a confirmation would be helpful.

PR #4314, opened 2018-02-05
Prevent faking invalid config keys in tests

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4313

### DIFF
--- a/tests/unit/helpers.js
+++ b/tests/unit/helpers.js
@@ -420,13 +420,14 @@ export function createUserAccountResponse({
 //   ...
 // }
 export const getFakeConfig = (params = {}) => {
-  Object.keys(params).forEach((key) => {
+  // eslint-disable-next-line no-restricted-syntax
+  for (const key of Object.keys(params)) {
     if (!config.has(key)) {
       // This will help alert us when a test accidentally relies
       // on an invalid config key.
       throw new Error(
         `Cannot set a fake value for "${key}"; this key is invalid`);
     }
-  });
+  }
   return Object.assign(configUtil.cloneDeep(config), params);
 };
--- a/tests/unit/helpers.js
+++ b/tests/unit/helpers.js
@@ -420,5 +420,13 @@ export function createUserAccountResponse({
 //   ...
 // }
 export const getFakeConfig = (params = {}) => {
+  Object.keys(params).forEach((key) => {
+    if (!config.has(key)) {
+      // This will help alert us when a test accidentally relies
+      // on an invalid config key.
+      throw new Error(
+        `Cannot set a fake value for "${key}"; this key is invalid`);
+    }
+  });
   return Object.assign(configUtil.cloneDeep(config), params);
 };
--- a/config/default.js
+++ b/config/default.js
@@ -269,4 +269,7 @@ module.exports = {
   // This setting is intended to simulate the Discopane's nginx cache
   // header for development purposes.
   cacheAllResponsesFor: null,
+
+  // See the Discopane config for how this is currently used.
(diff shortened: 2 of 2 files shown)


tests/unit/helpers.js
```diff
@@ -420,5 +420,13 @@ export function createUserAccountResponse({
 //   ...
 // }
 export const getFakeConfig = (params = {}) => {
+  Object.keys(params).forEach((key) => {
```
bobsilverberg: Would `for...of` be better than `forEach`?

tests/unit/helpers.js
```diff
@@ -420,13 +420,14 @@ export function createUserAccountResponse({
 //   ...
 // }
 export const getFakeConfig = (params = {}) => {
-  Object.keys(params).forEach((key) => {
+  // eslint-disable-next-line no-restricted-syntax
+  for (const key of Object.keys(params)) {
```
tofumatt: Oh sorry, I missed this comment before but we generally don't use `for...` loops, right? What's wrong with using `Object.keys().forEach()` as before?

kumar303:
> Do you think we should add a test for this too?

I don't think the helper would benefit much from having its own unit tests. As it is, we have good coverage of its functionality. A unit test might help us better pinpoint a bug but I don't know if it's worth it.

tests/unit/helpers.js
kumar303: It might be good for us to get into the habit of `for...of`. It has some benefits:
- You can use it on any iterable, not just arrays
- You can use a `break`

None of this applies to my patch but it seems reasonable.

I'm not really sure what's up with the eslint rule about it. We use `for...of` in a couple other places already and they each disable the lint rule.

tofumatt: Fair points!

I believe the rule stemmed more from using `for in` rather than `for of` but I forget exactly why. We'd have to go back and look at the issues to see, @muffinresearch might remember as I believe he made that PR.

willdurand: I did not know the `for of` es6 feature, but while looking for more info, I stumbled upon this issue, which might interest both of you: https://github.com/airbnb/javascript/issues/1271.

tofumatt: Yeah, that is a fair point in the issue of `break` being as naughty as a `GOTO`... it's not a lovely flow and if we need to use one maybe we could be writing our code better anyway.

Anyhow, I mostly just like:

* things being functions (`forEach`, `map`, etc.)
* sticking to one way of doing things (we tend to use array methods lots and we can convert things to arrays easily when they aren't already an array, meaning our objects are more predictable)
* I don't like adding code that goes against our lint rules when the reason is 🤷‍♂️ and the other code looks just as readable.

bobsilverberg: I had originally suggested it because it's the standard for the JS I've been writing for Firefox, and one of the justifications for that is that it is apparently faster. I'm not sure how much we care about that, especially in test code, and I agree that having a standard and sticking to it makes sense. If we'd rather stick with forEach, and make that a standard, I'm totally fine with that.

tofumatt: Yeah, that kind of micro-optimisation isn't really something I think we care about in the app... I'd say it's better to write readable code over fast code in this instance.

That said `for of` isn't terribly worse for readability, so it's not as big a deal (things like decrementing loops for performance always bothered me, though I forget if that's still a thing).

Agreed on consistency, mostly. Let's just pick one way to do it; I like `forEach` because it seems to encourage side effects less and other reasons I stated, but largely I just want consistency 😄

PR #4319, opened 2018-02-06
Make sure the proper viewContext is set when loading a category page

### LINKED ISSUE #4043: Themes menu option is not highlighted while accessing any theme category from homepage
Steps to reproduce:
1. Load a theme category from homepage i.e. https://addons.allizom.org/en-US/firefox/themes/scenery/
2. Observe underlined option from the menu

Expected results:
"Themes" option is underlined.

Actual results:
"Explore" option is underlined.

Notes/Issues:
Verified on AMO-stage FF57(Win 7). Issue is reproducing on all servers.
Screenshot for this issue:

bobsilverberg:
Fixes #4043 

When a category page is visited via direct link, the viewContext is not set to the proper value
which results in the wrong link in SectionLinks being highlighted. This patch adds a dispatch for
SET_VIEW_CONTEXT to the Category component to fix this issue.

I'm not entirely sure about the behaviour in the test for `does not fetch anything, but sets a viewContext when already loaded`. I think we still need to do this (i.e., set the viewContext), and that the test is therefore accurate, but I could be wrong about that.

### DIFF
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { getLanding, loadLanding } from 'amo/actions/landing';
+import { setViewContext } from 'amo/actions/viewContext';
 import Category, { CategoryBase } from 'amo/components/Category';
 import CategoryHeader from 'amo/components/CategoryHeader';
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
@@ -137,14 +138,15 @@ describe(__filename, () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
-  it('fetches categories and landing data when not yet loaded', () => {
+  it('fetches categories and landing data and sets a viewContext when not yet loaded', () => {
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.callCount(fakeDispatch, 2);
+    sinon.assert.callCount(fakeDispatch, 3);
     sinon.assert.calledWithMatch(fakeDispatch, categoriesFetch({
       errorHandlerId: errorHandler.id,
     }));
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
     sinon.assert.calledWith(fakeDispatch, getLanding({
       addonType: fakeCategory.type,
       category: fakeCategory.slug,
@@ -152,7 +154,7 @@ describe(__filename, () => {
     }));
   });
 
-  it('does not fetch anything when already loaded', () => {
+  it('does not fetch anything, but sets a viewContext when already loaded', () => {
     _categoriesFetch();
     _categoriesLoad();
     _getLanding();
@@ -161,7 +163,8 @@ describe(__filename, () => {
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.notCalled(fakeDispatch);
+    sinon.assert.callCount(fakeDispatch, 1);
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
   });
 
   it('does not fetch categories when an empty set was loaded', () => {
@@ -235,14 +238,15 @@ describe(__filename, () => {
     sinon.assert.notCalled(fakeDispatch);
   });
 
-  it('dispatches getLanding when results are not loaded', () => {
+  it('dispatches getLanding and sets a viewContext when results are not loaded', () => {
     _categoriesFetch();
     _categoriesLoad();
 
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.callCount(fakeDispatch, 1);
+    sinon.assert.callCount(fakeDispatch, 2);
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
     sinon.assert.calledWith(fakeDispatch, getLanding({
       addonType: fakeCategory.type,
       category: fakeCategory.slug,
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestCategory.js
```diff
@@ -235,22 +238,23 @@ describe(__filename, () => {
...
+  it('dispatches getLanding and sets a viewContext when results are not loaded', () => {
     _categoriesFetch();
     _categoriesLoad();
 
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.callCount(fakeDispatch, 1);
+    sinon.assert.callCount(fakeDispatch, 2);
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
```
kumar303: This test exercises [a specific if branch](https://github.com/bobsilverberg/addons-frontend/blob/d41b5e99ba24219eab50f1a1127f937497195fa6/src/amo/components/Category/index.js#L110).

It's not necessary to check the view context logic because that's not affected by the if branch. In fact, by checking the view context behavior, you are duplicating the [other test](https://github.com/bobsilverberg/addons-frontend/blob/d41b5e99ba24219eab50f1a1127f937497195fa6/tests/unit/amo/components/TestCategory.js#L141). Such a duplication won't really hurt anything but it could potentially introduce a maintenance burden.

I would suggest these changes:
* Remove the `assert.callCount()` from this test since it's not necessary
* Remove the assertion of the `setViewContext()` call

tests/unit/amo/components/TestCategory.js
```diff
@@ -272,14 +276,15 @@ describe(__filename, () => {
       },
     });
 
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
```
kumar303: Similarly, this test exercises [only a specific if branch](https://github.com/bobsilverberg/addons-frontend/blob/d41b5e99ba24219eab50f1a1127f937497195fa6/src/amo/components/Category/index.js#L113). It's not necessary to make an assertion about the view context logic since that's not affected by the if branch.

tests/unit/amo/components/TestCategory.js
```diff
@@ -307,6 +312,7 @@ describe(__filename, () => {
       category,
       errorHandlerId: errorHandler.id,
     }));
+    sinon.assert.calledWith(fakeDispatch, setViewContext(addonType));
```
kumar303: And again, this test was meant to cover [a specific if branch](https://github.com/bobsilverberg/addons-frontend/blob/d41b5e99ba24219eab50f1a1127f937497195fa6/src/amo/components/Category/index.js#L112) so it's not necessary to add an assertion for the view context logic.

tests/unit/amo/components/TestCategory.js
```diff
@@ -161,7 +163,8 @@ describe(__filename, () => {
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.notCalled(fakeDispatch);
+    sinon.assert.callCount(fakeDispatch, 1);
+    sinon.assert.calledWith(fakeDispatch, setViewContext(fakeCategory.type));
```
kumar303: I saw your question about this test. The test intends to cover [only this if branch](https://github.com/bobsilverberg/addons-frontend/blob/d41b5e99ba24219eab50f1a1127f937497195fa6/src/amo/components/Category/index.js#L95) which is there to prevent fetching data when the data is already loaded. Because of that, I don't think it's necessary to cover the view context logic.

The old approach was to assert that nothing is dispatched but that won't work anymore. I suggest changing the assertion to be specific about category fetching:

```js
import { CATEGORIES_FETCH } from 'core/constants';

sinon.assert.neverCalledWith(fakeDispatch, sinon.match({ type: CATEGORIES_FETCH }));
```

bobsilverberg: Thanks Kumar, that all makes sense. 👍

tests/unit/amo/components/TestCategory.js
```diff
@@ -137,22 +139,23 @@ describe(__filename, () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
-  it('fetches categories and landing data when not yet loaded', () => {
+  it('fetches categories and landing data and sets a viewContext when not yet loaded', () => {
```
kumar303: Just FYI: we try to keep line length to 80 characters. It's not mandatory, just a nice-to-have for people like me who prefer to use horizontal splits. We haven't put a [lint rule](https://github.com/mozilla/addons-frontend/issues/3045) in place because no one has the courage to review the patch 🙂

tests/unit/amo/components/TestCategory.js
```diff
@@ -161,7 +164,8 @@ describe(__filename, () => {
     const fakeDispatch = sinon.stub(store, 'dispatch');
     render({}, { autoDispatchCategories: false });
 
-    sinon.assert.notCalled(fakeDispatch);
+    sinon.assert.callCount(fakeDispatch, 1);
```
kumar303: I'd like to avoid call count assertions unless absolutely necessary because they introduce a maintenance burden. For example, we may later add another unrelated dispatch to the component but that should not break this test.

tests/unit/amo/components/TestCategory.js
```diff
@@ -137,22 +139,24 @@ describe(__filename, () => {
     expect(root.find(ErrorList)).toHaveLength(1);
   });
 
-  it('fetches categories and landing data when not yet loaded', () => {
+  it('fetches categories and landing data ' +
```
muffinresearch: You could have use a template literal here - also there's helper tag called `oneLine` that will remove leading whitespace from the second line. See https://github.com/mozilla/addons-frontend/blob/350227964acd5818bacefb5057108678a463a36e/src/core/server/sriDataPlugin.js#L5 for an example.

PR #4321, opened 2018-02-06
Add edition to allowed params to be passed through to the disco pane API

### LINKED ISSUE #4320: Allow "edition" query param to be passed through to the disco pane API.
A follow-up from https://github.com/mozilla/addons-server/issues/7489 

Once this lands, passing `?edition=china` on the end of a disco pane URL should show different content.

muffinresearch:
Fixes #4320 

Running this locally with `yarn run disco` and hitting http://127.0.0.1:3000/en-US/firefox/discovery/pane/48.0/Darwin/normal?edition=china shows up the different content.

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -55,6 +55,7 @@ module.exports = {
   taarParamsToUse: [
     'branch',
     'clientId',
+    'edition',
     'platform',
     'study',
   ],


config/default-disco.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
   taarParamsToUse: [
```
eviljeff: change to `discoParamsToUse`? 😉

muffinresearch: Good idea 👍 I'll do that as a follow-up patch.

PR #4325, opened 2018-02-06
Display a tooltip for screenshots

### LINKED ISSUE #3917: Screenshots should display tips while hovering them
Describe the problem and steps to reproduce it:

Open:
https://addons.mozilla.org/en-US/firefox/addon/british-english-dictionary-2/
and hover the mouse over the screenshots.

What happened?

No tips are shown (they used to be on the old site) and the screenshots don't fit in the gadget.
What did you expect to happen?

See the tips and the images should fit in the gadget.

Anything else we should know?

bobsilverberg:
Fixes #3917

This adds a title attribute to the img tag that is generated for screenshot thumbnails
which will result in a tooltip when hovering.

It uses the value of item.title for the title, which comes from the caption property of
the preview object returned via the API.

### DIFF
--- a/tests/unit/amo/components/TestScreenShots.js
+++ b/tests/unit/amo/components/TestScreenShots.js
@@ -49,7 +49,7 @@ describe('<ScreenShots />', () => {
   });
 
   it('renders custom thumbnail', () => {
-    const item = { src: 'https://foo.com/img.png' };
+    const item = { src: 'https://foo.com/img.png', title: 'test title' };
     const thumbnail = shallow(thumbnailContent(item));
 
     expect(thumbnail.type()).toEqual('img');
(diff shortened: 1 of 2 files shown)


bobsilverberg:
Before: No tooltip for screenshots on hover:

feb-08-2018 09-31-27

After: A tooltip appears for screenshots on hover:

feb-08-2018 09-29-14

tests/unit/amo/components/TestScreenShots.js
```diff
@@ -49,14 +49,15 @@ describe('<ScreenShots />', () => {
...
-    const item = { src: 'https://foo.com/img.png' };
+    const item = { src: 'https://foo.com/img.png', title: 'test title' };
     const thumbnail = shallow(thumbnailContent(item));
 
     expect(thumbnail.type()).toEqual('img');
     expect(thumbnail.prop('src')).toEqual('https://foo.com/img.png');
     expect(thumbnail.prop('height')).toEqual(HEIGHT);
     expect(thumbnail.prop('width')).toEqual(WIDTH);
     expect(thumbnail.prop('alt')).toEqual('');
+    expect(thumbnail.prop('title')).toEqual('test title');
```
kumar303: I know you were just following the pattern but for future reference we prefer this check now:

```js
expect(thumbnail).toHaveProp('title', 'test title');
```

This produces a better failure message.

PR #4337, opened 2018-02-07
Implement yellow stars for user's add-on rating

### LINKED ISSUE #4029: Implement yellow stars for user's add-on rating
Per UX, the add-on rating stars should be yellow whenever they indicate the site user's personal rating. They should be gray in all other cases (for example: cumulative ratings in add-on listings).

Consult the UX specs before implementing. The spec currently looks like:

screenshot 2017-12-04 15 11 34

bobsilverberg:
Fixes #4029 

Display yellow stars in dialogs where a user is able to specify their rating.


src/ui/components/Rating/img/full-star-gold.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="17px" height="16px" viewBox="0 0 17 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
kumar303: I'm not sure what happened here but this seems to be off by one pixel. When I go to the rating screen and hover over a star, it jumps. I tried fixing it in the SVG markup but it didn't work. I think you have to export it again.

addons-frontend-stars mov

kumar303:
My understanding of https://github.com/mozilla/addons-frontend/issues/4029 is that all visual representations of the user's personal ratings should be yellow. @pwalm is that correct?

If so, I think we should make these stars yellow too. However, I'd suggest filing a separate bug for it.

reviews for video downloadhelper add ons for firefox

src/ui/components/Rating/img/full-star-gold.svg
bobsilverberg: This is probably a result of the hollow star being an export from Sketch and the filled-in star being a copy of an svg already in the repo, but with a different fill colour. I didn't see a filled-in star in sketch for me to export. @pwalm Do you know if there is one in there, and if not might you be able to provide me with a set of hollow and filled-in stars that don't have this issue?

kumar303: Here's a filled-in one https://pastebin.mozilla.org/9077566

I just deleted the inner shape using Sketch. Let me know if that doesn't fix it.

bobsilverberg: Thanks @kumar303, that seems to have fixed it.

kumar303:
Can you file a new bug for https://github.com/mozilla/addons-frontend/pull/4337#issuecomment-363904246 ? We'll get some feedback from @pwalm before implementing it.

pwalm:
@kumar303 yep! A user should always see their star ratings as yellow.

PR #4343, opened 2018-02-08
chore: Add a DefinitionList component

### LINKED ISSUE #4342: Create a styled DL with a renderIfNodes API
The User profile page requires something very similar to:

screenshot 2018-02-08 12 50 49

which also has a handy `renderNodesIf()` helper.

https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonMoreInfo/index.js#L180

I think we could abstract this out into a component for displaying DL-type info with a common style and nicer API.

tofumatt:
fix #4342

While this means rendering a DL looks less like HTML, I think this API is actually a better fit for how we're going to render components like:

screenshot 2018-02-08 14 49 13
and
screenshot 2018-02-08 12 50 49

This is a pattern we're going to have to do more of, so I figured this was worth it.


tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -33,7 +34,7 @@ describe(__filename, () => {
         {...props}
       />,
       AddonMoreInfoBase
-    );
+    ).find(DefinitionList).shallow();
```
kumar303: I think this is a pattern we should avoid because it involves reaching into the implementation of another component. The reason you had to do this was because you used `<DefinitionList items={...} />` instead of rendering children elements. If you switch to children elements then you won't need to dive into `<DefinitionList />`.

I took a quick stab at it and I think it will be possible with children elements. How about something more like this?

```diff
diff --git a/src/amo/components/AddonMoreInfo/index.js b/src/amo/components/AddonMoreInfo/index.js
index 0ae98231b..804a3bb0c 100644
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -171,69 +171,30 @@ export class AddonMoreInfoBase extends React.Component<Props> {
   }: Object) {
     const { i18n } = this.props;
     return (
-      <DefinitionList
-        className="AddonMoreInfo-dl"
-        items={[
-          {
-            dt: i18n.gettext('Add-on Links'),
-            dd: (
-              <ul className="AddonMoreInfo-links-contents-list">
-                {homepage}
-                {supportUrl}
-                {supportEmail}
-              </ul>
-            ),
-            key: 'AddonMoreInfo-links',
-            renderIf: homepage || supportUrl || supportEmail,
-          },
-          {
-            dt: i18n.gettext('Version'),
-            dd: version,
-            key: 'AddonMoreInfo-version',
-            renderIf: version,
-          },
-          {
-            dt: i18n.gettext('versionLastUpdated'),
-            dd: versionLastUpdated,
-            key: 'AddonMoreInfo-last-updated',
-          },
-          {
-            dt: i18n.gettext('License'),
-            dd: versionLicenseLink,
-            key: 'AddonMoreInfo-license',
-            renderIf: versionLicenseLink,
-          },
-          {
-            dt: i18n.gettext('Privacy Policy'),
-            dd: privacyPolicyLink,
-            key: 'AddonMoreInfo-privacy-policy',
-            renderIf: privacyPolicyLink,
-          },
-          {
-            dt: i18n.gettext('End-User License Agreement'),
-            dd: eulaLink,
-            key: 'AddonMoreInfo-eula',
-            renderIf: eulaLink,
-          },
-          {
-            dt: i18n.gettext('Version History'),
-            dd: (
-              <ul className="AddonMoreInfo-links-contents-list">
-                {versionHistoryLink}
-                {betaVersionsLink}
-              </ul>
-            ),
-            key: 'AddonMoreInfo-version-history',
-            renderIf: versionHistoryLink || betaVersionsLink,
-          },
-          {
-            dt: i18n.gettext('Usage Statistics'),
-            dd: statsLink,
-            key: 'AddonMoreInfo-stats',
-            renderIf: statsLink,
-          },
-        ]}
-      />
+      <DefinitionList className="AddonMoreInfo-dl">
+        {(homepage || supportUrl || supportEmail) && (
+          <Definition term={i18n.gettext('Add-on Links')}>
+            <ul className="AddonMoreInfo-links-contents-list">
+              {homepage}
+              {supportUrl}
+              {supportEmail}
+            </ul>
+          </Definition>
+        )}
+        {version && (
+          <Definition term={i18n.gettext('Version')}>
+            {version}
+          </Definition>
+        )}
+        <Definition term={i18n.gettext('versionLastUpdated')}>
+          {versionLastUpdated}
+        </Definition>
+        {versionLicenseLink && (
+          <Definition term={i18n.gettext('License')}>
+            {versionLicenseLink}
+          </Definition>
+        )}
+      </DefinitionList>
     );
   }
 
```

Also, I don't think we need `renderNodesIf()` anymore. I can't remember exactly why we added it but I think it was maybe just a workaround for not having `React.Fragment`.

tofumatt: Oh damn, that's a nice API. It means having to import `DefinitionList` and `Definition[Item]` but I like it. I thought about using children before but my approach was still too programmatic. I'll do that. 👍

tofumatt:
I went with the children approach and I think it's a lot nicer, thanks. This is ready for another review.

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+  children: React.Node,
+  className?: string,
+  title: React.Node,
+|};
+
+export const DefinitionItem = (
+  { children, className = '', title }: DefinitionItemProps = {}
+) => {
+  return (
+    <React.Fragment>
```
tofumatt: Happy to use `<>` here, but I wanted to push this for review last night and did't check to make sure our Babel supported it. I'll do that and update if it's easy, or we can just address it in #4345.

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+import './styles.scss';
+
+
+type DefinitionItemProps = {|
+  children: React.Node,
+  className?: string,
+  title: React.Node,
+|};
+
+export const DefinitionItem = (
```
kumar303: I think it would be nice to just call this `Definition` since that's pretty much what a `<dd>` stands for. The fact that `<dd>` requires a separate `<dt>` tag is really weird to me.

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+  { children, className = '', title }: DefinitionItemProps = {}
+) => {
+  return (
+    <React.Fragment>
+      <dt
+        className={makeClassName(className, {
+          [`${className}-title`]: className && className.length,
```
kumar303: I think we should avoid dynamic class names derived from the configured class name because there is no way to look at the calling code and guess that it's happening.

From what I can tell, there are no `AddonMoreInfo...title` classes in any style sheets so I say you either remove it or replace it with a hard-coded class like `DefinitionItem-title`.

If you really needed a custom class like this then it would be better as a custom property, e.g.:

```js
          <DefinitionItem
            className="AddonMoreInfo-version"
            title={i18n.gettext('Version')}
            titleClassName="AddonMoreInfo-version-title"
          >
            {version}
          </DefinitionItem>
```

When looking at the component instance, the custom styling possibilities would be more obvious.

However, I say you just remove the title class for now since we don't need it.

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+        })}
+      >
+        {title}
+      </dt>
+      <dd
+        className={makeClassName(className, {
+          [`${className}-contents`]: className && className.length,
```
kumar303: For the same reasons as above, I think you should remove this for now

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -41,19 +42,23 @@ describe(__filename, () => {
...
     // These fields will not be visible during loading
     // since they may not exist.
-    expect(root.find('.AddonMoreInfo-links-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-license-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-privacy-policy-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-version-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-version-history-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-eula-title')).toHaveLength(0);
-    expect(root.find('.AddonMoreInfo-beta-versions-title')).toHaveLength(0);
+    expect(root.find({ className: 'AddonMoreInfo-links' })).toHaveLength(0);
```
kumar303: All of these calls seem overly verbose. What's your objection to `root.find('.AddonMoreInfo-links')`? It seems to be working for me locally.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -419,8 +420,10 @@ describe(__filename, () => {
     });
     const root = render({ addon });
 
-    const link = root.find('.AddonMoreInfo-beta-versions-link');
-    expect(link).toHaveProp('href', `/addon/${addon.slug}/versions/beta`);
+    expect(root
+      // .find({ className: 'AddonMoreInfo-version-history-link' })
```
kumar303: This can be deleted

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+import * as React from 'react';
+import makeClassName from 'classnames';
+
+import './styles.scss';
+
+
+type DefinitionItemProps = {|
+  children: React.Node,
+  className?: string,
+  title: React.Node,
```
kumar303: No biggie but `term` might be a better name to illustrate how it's just a wrapper for `<dt>{term}</dt>`

src/ui/components/DefinitionList/index.js
```diff
@@ -0,0 +1,50 @@
...
+      >
+        {children}
+      </dd>
+    </React.Fragment>
+  );
+};
+
+type DefinitionListProps = {|
+  className?: string,
+  children: React.Node,
```
kumar303: You may get some helpful Flow validation if you restrict this to the only type of child the component accepts:

```js
children: React.ChildrenArray<React.Element<typeof DefinitionList>>,
```

src/ui/components/DefinitionList/index.js
tofumatt: That's a good point. In the past they were useful for testing but I don't think that's the case anymore, so works for me.

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: I think the `toHaveLength(0)` will work regardless because it won't find either `. AddonMoreInfo-links` or `{ className: 'AddonMoreInfo-links' }`... but does it work when you are verifying something *does* exist? I'll check and if so I'll switch back.

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: Oops! 🔥

src/ui/components/DefinitionList/index.js
tofumatt: Good call 👍

src/ui/components/DefinitionList/index.js
tofumatt: Oh, I'll do that.

tests/unit/amo/components/TestAddonMoreInfo.js
tofumatt: Indeed, using just `root.find('.Addon[...]')` worked, so I've switched back. 🤷‍♂️

tofumatt:
Okay, comments addressed and should be ready for another review 😄

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -179,7 +183,7 @@ describe(__filename, () => {
     const root = render({ addon });
     const link = root.find('.AddonMoreInfo-license-link');
 
-    expect(root.find('.AddonMoreInfo-license-title')).toHaveText('License');
+    expect(root.find('.AddonMoreInfo-license')).toHaveProp('term', 'License');
     expect(link).toHaveProp('children', 'tofulicense');
```
kumar303: I think it's better to do something like:

```js
expect(link.childAt(0)).toIncludeText('tofulicense');
```

but maybe it doesn't matter.

tofumatt: I don't think it really matters, but if you prefer that I'll change it, totally cool. 👍

Issue #4347, opened 2018-02-08
Fix `yarn amo` for developing with a local addons-server API


kumar303:
The local domain `olympia.dev` [is changing](https://github.com/mozilla/addons-server/pull/7402) to `olympia.test`. We need to update some configs and docs for this new setup.

Note that the upstream changes will *not* affect `yarn amo:dev`, `yarn amo:stage`, or `yarn disco`.

PR #4349, opened 2018-02-08
Fix `yarn amo` for developing with a local addons-server API

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4347 with one config change. The rest of the changes here are for the documentation.

I am testing with Docker against https://github.com/mozilla/addons-server/pull/7402 (we should wait to merge until that has landed).

**For developers**: We need to update our `/etc/hosts` file according to https://github.com/mozilla/addons-server/pull/7402/files#diff-01e2128b6420f1c2f6f27eb5340138dfL70

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,36 +202,41 @@ If you would like to use `https://addons-dev.allizom.org` for data you should us
 `yarn amo:dev` command. See the table of commands up above for similar
 hosted options.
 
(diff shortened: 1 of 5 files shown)


tofumatt:
(Possibly worth sending an email round to developers on addons-dev about the update needed.)

Issue #4351, opened 2018-02-08
Remove "updating collection" state in favor of FormOverlay state


kumar303:
The patch from https://github.com/mozilla/addons-frontend/pull/4209 added a concept of "updating" or "not updating" state. I realized that this is no longer necessary because we can use the same state from the [`FormOverlay`](https://github.com/mozilla/addons-frontend/issues/4300). This is a more generic pattern that can be re-used in future forms.

Issue #4350, opened 2018-02-08
Update-collection saga should reset collection state


kumar303:
The update-collection saga added in https://github.com/mozilla/addons-frontend/issues/4196 supports this user flow:

- open a form to edit a collection
- change some collection data like name, description
- patch the collection using the API
- close the form and **see the newly added collection data**

In order to implement the last part of this (in bold), the saga needs to invalidate the collection data. This will force all components to reload it from the API.

Issue #4323, opened 2018-02-06
Fix how localized params are passed through collection actions/sagas


kumar303:
The API function requires [a localized string](https://github.com/mozilla/addons-frontend/pull/4201/files#diff-f2a5dd9c232205755bd944f602a2e32fR161) but the action/saga was only [passing a regular string](https://github.com/mozilla/addons-frontend/pull/4209/files#diff-5c509a4b5942e40bfb891f933815dc7bR454).

Interestingly, the point of contact with the API function was through the saga and since the saga doesn't have Flow coverage, the invalid types went unnoticed. Flow would have thrown an error.

AlexandraMoga:
Verified on AMO-stage with FF58 (Win10 and Android 7.0). There were no issues spotted around the available collection pages/features.

PR #4352, opened 2018-02-08
Fix update-collection saga, clean up collection reducer

kumar303:
* Fixes https://github.com/mozilla/addons-frontend/issues/4350 by invalidating the collection object after an update
* Fixes https://github.com/mozilla/addons-frontend/issues/4351 by removing the `finishUpdateCollection()` action which is no longer needed

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,9 +13,11 @@
     // See: https://github.com/facebook/flow/issues/1609
     "SyntheticEvent": true,
     // Standard DOM globals:
+    "Generator": true,
     "HTMLElement": true,
     "HTMLInputElement": true,
     "HTMLSelectElement": true,
+    "HTMLTextAreaElement": true,
     "Node": true,
   },
   "parser": "babel-eslint",
--- a/src/amo/sagas/collections.js
+++ b/src/amo/sagas/collections.js
@@ -1,3 +1,4 @@
+/* @flow */
 import { all, call, put, select, takeLatest } from 'redux-saga/effects';
 import {
   ADD_ADDON_TO_COLLECTION,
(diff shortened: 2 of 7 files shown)


.eslintrc
```diff
@@ -13,9 +13,11 @@
     // See: https://github.com/facebook/flow/issues/1609
     "SyntheticEvent": true,
     // Standard DOM globals:
+    "Generator": true,
     "HTMLElement": true,
     "HTMLInputElement": true,
     "HTMLSelectElement": true,
+    "HTMLTextAreaElement": true,
```
kumar303: This will be needed for a future patch but `git add -p` wanted to add it.

src/amo/sagas/collections.js
```diff
@@ -1,3 +1,4 @@
+/* @flow */
```
kumar303: I ported this to Flow because Collections require a lot of complex actions and I can no longer keep it all in my head 😞 

This was a specific bug we had that Flow would have caught: https://github.com/mozilla/addons-frontend/issues/4323

When I converted it to Flow I noticed another bug! The `slug` parameter was incorrect.

src/amo/sagas/collections.js
```diff
@@ -144,36 +169,40 @@ export function* updateCollection({
...
-    });
+    };
+    yield call(api.updateCollection, params);
 
+    // Invalidate the stored collection object. This will force each
+    // component to re-fetch the collection.
+    yield put(deleteCollectionBySlug(collectionSlug));
     yield put(closeFormOverlay(formOverlayId));
-    yield put(finishUpdateCollection({ collectionSlug, successful: true }));
+    yield put(finishFormOverlaySubmit(formOverlayId));
```
tofumatt: Seeing this in the collections saga makes me feel a bit uneasy though. I get it's more generic/reusable but it feels like collections saga/API work shouldn't be so tightly coupled to the UI implementation of a form overlay. What happens if this is done elsewhere without a form overlay?

I know it was like that before and it's the only place it's used now... and also I guess it wouldn't be hard to abstract out so it's fine. I guess it just seem jarring to me to have the update action in the collection saga be fired by a UI state/action. It feels a bit too tightly coupled to me.

kumar303: Yeah, I definitely hear what you're saying and this was on my mind as well.

> What happens if this is done elsewhere without a form overlay?

My current idea is that we would make another saga that shares some of the same update code -- i.e. we'd have two sagas like `updateCollectionFromOverlay` and `updateCollectionFromPage`. This will be easy to do because a saga can dispatch an action to trigger another saga.

But, yeah, I'm not sure if this is the best pattern. To be fair, a saga really is meant to encapsulate a series of actions so maybe it makes sense.

I think we should keep iterating and see what happens. One thing worth mentioning is that we have no intention of hosting an edit collection screen *not* in an overlay 😄 so this may never be an issue.

If you have any other ideas for how to implement it, let me know. My original approach was to make the begin/finish state more specific to the act of *editing a collection* but then I had to infer the form overlay state by inspecting the collection state so it just felt like saving state in two places whereas this patch does it in one.

kumar303:
Thanks for the review. I think we should merge this to let it simmer. Perhaps a better pattern will come into view once we work with a couple other form overlays.

PR #4353, opened 2018-02-08
Remove beta version listing

### LINKED ISSUE #4317: Remove beta version support
This is part of https://github.com/mozilla/addons-server/issues/7163

The frontend currently exposes beta versions. Since we're going to remove this feature, we need to remove all references to beta versions, gated by a waffle flag.

wagnerand:
Fixes #4317.


src/core/types/addons.js
```diff
@@ -167,3 +167,7 @@ export type AddonType = {
   isWebExtension: boolean,
   themeData?: ThemeData,
 };
+
+export type CreateInternalAddonOptions = {|
```
kumar303: This should be moved to `core/reducers/addons.js` right above the definition of `createInternalAddon`.

src/core/types/addons.js
```diff
@@ -167,3 +167,7 @@ export type AddonType = {
...
+  _config?: Object,
```
kumar303: This should be `_config: typeof config` which is the same as `Object` but it helps to make the association.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -423,6 +423,22 @@ describe(__filename, () => {
     expect(link).toHaveProp('href', `/addon/${addon.slug}/versions/beta`);
   });
 
+  it('does not link to beta versions if they are disabled', () => {
```
kumar303: This test should move to https://github.com/mozilla/addons-frontend/blob/master/tests/unit/core/reducers/test_addons.js

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -423,6 +423,22 @@ describe(__filename, () => {
...
+    const fakeConfig = new Map();
```
kumar303: use [`getFakeConfig({ betaVersions: flase })`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/helpers.js#L422) instead

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -423,6 +423,22 @@ describe(__filename, () => {
...
+
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      slug: 'some-slug',
+      current_beta_version: {
+        ...fakeAddon.current_version,
+        version: '3.0.0-beta',
+      },
+    },
+    { _config: fakeConfig });
```
kumar303: I know we don't have great lint rules for this but this should be indented to indicate that it's a continuation of the call up above. I'd suggest something more like:

```js
const addon = createInternalAddon(
  {
    ...fakeAddon,
    slug: 'some-slug',
  },
  { _config: fakeConfig },
);
```

tests/unit/amo/components/TestAddonMoreInfo.js
kumar303: This spec should describe what's happening at a lower level. I suggest something like "it does not expose beta version properties if disabled"

config/default.js
```diff
@@ -272,4 +273,6 @@ module.exports = {
 
   // See the Discopane config for how this is currently used.
   taarParamsToUse: [],
+
+  betaVersions: true,
```
kumar303: Since this config key is a bit vague, I'd like to see some comments explaining how this controls add-on beta version features.

src/core/types/addons.js
wagnerand: Hehe, I stole this from `core/reducers/redirectTo.js#21`.

tests/unit/amo/components/TestAddonMoreInfo.js
wagnerand: Another thing I stole from another test :)

kumar303:
When you're ready for another look, just re-request a review from me so that it shows up in my review queue.

config/default.js
```diff
@@ -274,6 +275,10 @@ module.exports = {
   // See the Discopane config for how this is currently used.
   taarParamsToUse: [],
 
+  // If true, show a link to the beta versions page on the listing page
+  betaVersions: true,
```
kumar303: I made a new 'Feature flags' section in the config (see line 282) so this should move there now.

config/dev-amo.js
```diff
@@ -25,5 +25,7 @@ module.exports = {
     },
   },
 
+  betaVersions: false,
```
kumar303: You should also add this to the local development config: https://github.com/wagnerand/addons-frontend/blob/e38bf9f021a196fb30b957cf64c8db5345dfaa92/config/development-amo.js

PR #4354, opened 2018-02-08
Update homepage content for Feb 15

### LINKED ISSUE #4330: 15 February homepage content programming
For the **Thursday, 15 February** push, can we please update the AMO homepage programming in accord with what's outlined [here](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#heading=h.n5783ro8ws5h)? 

@EnTeQuAk @diox @muffinresearch

bobsilverberg:
Fixes #4330 

This changes all the necessary files for the content update, and also includes a change to the way that collections are requested by and fetched for the home page. A variable number of collections can now be requested without requiring changes to the reducer or the saga.

I feel like a further refactor could be beneficial. It would be nice if the home page could specify which datasets it needs, and the reducer and the saga could  maintain the logic for all of the possible datasets (as opposed to adding and  removing code each time a change is required). Then the saga could retrieve only the requested data, and pass only the requested data back. As this procedure is only in place temporarily, I wasn't sure if it was worth making that change as well, but I'm happy to do so if you think it would be worthwhile.

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -13,7 +13,6 @@ import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
   SEARCH_SORT_POPULAR,
-  SEARCH_SORT_TOP_RATED,
   VIEW_CONTEXT_HOME,
 } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
@@ -24,20 +23,21 @@ import Icon from 'ui/components/Icon';
 import './styles.scss';
 
 
-export const FIRST_COLLECTION_SLUG = 'dynamic-media-downloaders';
-export const FIRST_COLLECTION_USER = 'mozilla';
+export const COLLECTIONS_TO_FETCH = [
+  { slug: 'privacy-matters', user: 'mozilla' },
+  { slug: 're-imagine-search', user: 'mozilla' },
+  { slug: 'addonsofthemonth', user: 'mozilla' },
+];
 
 export class HomeBase extends React.Component {
   static propTypes = {
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
-    firstCollection: PropTypes.array.isRequired,
+    collections: PropTypes.array.isRequired,
     featuredExtensions: PropTypes.array.isRequired,
-    featuredThemes: PropTypes.array.isRequired,
+    popularThemes: PropTypes.array.isRequired,
     i18n: PropTypes.object.isRequired,
-    popularExtensions: PropTypes.array.isRequired,
     resultsLoaded: PropTypes.bool.isRequired,
-    topRatedExtensions: PropTypes.array.isRequired,
   }
 
   componentWillMount() {
@@ -48,8 +48,7 @@ export class HomeBase extends React.Component {
     if (!resultsLoaded) {
       dispatch(fetchHomeAddons({
         errorHandlerId: errorHandler.id,
-        firstCollectionSlug: FIRST_COLLECTION_SLUG,
-        firstCollectionUser: FIRST_COLLECTION_USER,
+        collectionsToFetch: COLLECTIONS_TO_FETCH,
       }));
     }
   }
@@ -155,13 +154,11 @@ export class HomeBase extends React.Component {
   render() {
     const {
       errorHandler,
-      firstCollection,
+      collections,
       featuredExtensions,
-      featuredThemes,
+      popularThemes,
       i18n,
-      popularExtensions,
       resultsLoaded,
-      topRatedExtensions,
     } = this.props;
 
     // translators: The ending ellipsis alludes to a row of icons for each type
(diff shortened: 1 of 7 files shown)


src/amo/components/Home/index.js
```diff
@@ -24,20 +23,21 @@ import Icon from 'ui/components/Icon';
 import './styles.scss';
 
 
-export const FIRST_COLLECTION_SLUG = 'dynamic-media-downloaders';
-export const FIRST_COLLECTION_USER = 'mozilla';
+export const COLLECTIONS_TO_FETCH = [
```
tofumatt: Oh, this is much nicer 👍

src/amo/components/Home/index.js
```diff
@@ -197,72 +194,64 @@ export class HomeBase extends React.Component {
...
-            query: {
-              addonType: ADDON_TYPE_EXTENSION,
-              sort: SEARCH_SORT_TOP_RATED,
-            },
+          addons={collections[1]}
+          className="Home-FeaturedCollection"
+          header={i18n.gettext('Re-imagine search')}
+          footerText={i18n.gettext('See more search extensions')}
+          footerLink={{ pathname:
+            `/collections/${COLLECTIONS_TO_FETCH[1].user}/${COLLECTIONS_TO_FETCH[1].slug}/`,
```
tofumatt: Technically this (and anything else that's just a `pathname`) can be a string instead of an object. So:

```jsx
footerLink={`/collections/${COLLECTIONS_TO_FETCH[1].user}/${COLLECTIONS_TO_FETCH[1].slug}/`}
// instead of:
footerLink={{ pathname:
  `/collections/${COLLECTIONS_TO_FETCH[1].user}/${COLLECTIONS_TO_FETCH[1].slug}/`,
}}
```

src/amo/components/Home/index.js
```diff
@@ -197,72 +194,64 @@ export class HomeBase extends React.Component {
...
-          footerText={i18n.gettext('See more featured extensions')}
-          footerLink={{
-            pathname: '/search/',
-            query: {
-              addonType: ADDON_TYPE_EXTENSION,
-              featured: true,
-            },
+          addons={collections[2]}
+          className="Home-FeaturedCollection"
+          header={i18n.gettext(`February's new featured extensions`)}
```
tofumatt: Hyper copy nitpick but this should really be "February’s new" not "February's new"... but I assume it was copy+pasted so we can blame editorial not engineering 😉

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -46,64 +46,62 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
+            {i18n.gettext('Secure encryption for webmail.')}
           </p>
         </HeroSection>
       ),
       (
         <HeroSection
-          key="facebook-filter"
-          linkTo="/addon/facebook-filter/"
+          key="hero-5"
+          linkTo="/addon/%E7%BF%BB%E8%AF%91%E4%BE%A0-translate-man/"
```
tofumatt: Whoa, that's a heck of a slug...

tofumatt: Oh, I see, it should be `linkTo="/addon/翻译侠-translate-man/"`. Any reason we can't use the unicode characters in the URL?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -116,12 +114,41 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
+          key="hero-9"
+          linkTo="/addon/searchpreview/"
         >
-          <h3>{i18n.gettext('Tree Style Tab')}</h3>
+          <h3>{i18n.gettext('SearchPreview')}</h3>
 
-          <p>{i18n.gettext('Display tabs in a space-saving “tree” layout.')}</p>
+          <p>
+            {i18n.gettext(`Enhance search results with thumbnail previews,
+              popularity ranks & more.`)}
```
tofumatt: I assume this `&` gets escaped by React, right? I think so.

src/amo/sagas/home.js
```diff
@@ -32,20 +29,22 @@ export function* fetchHomeAddons({
   try {
     const state = yield select(getState);
 
+    const collections = [];
+    // eslint-disable-next-line no-restricted-syntax
+    for (const coll of collectionsToFetch) {
```
tofumatt: Can we use a more descriptive variable name for this over `coll`? I think `collection` is fine.

Also, if we're going to keep using `for... of` we shouldn't have it be an eslint violation. I'm fine with doing it, but it should be an okay thing to do 😄

src/amo/components/Home/index.js
bobsilverberg: Thanks, that's nicer.

src/amo/components/HomeHeroBanner/index.js
bobsilverberg: Thanks, that works.

src/amo/components/HomeHeroBanner/index.js
bobsilverberg: Seems to be the case. It displays fine when that item is the large item with the description displayed.

src/amo/sagas/home.js
bobsilverberg: I tried to use `forEach` and `map` instead of `for... of`, based on the conversation in an earlier issue, but couldn't seem to figure out the syntax that would work because of the `yield` and generators. If we're okay with using `for... of` in some situations (which seems to be what you're saying here), I agree we should fix the eslint rule. @tofumatt Is it okay to just add that to this PR and commit, or would you prefer that a separate issue and commit be done for that?

bobsilverberg: Ah, I see that the eslint config doesn't even reside in this repo, so I guess it would have to be a separate commit.

Issue #4258, opened 2018-01-30
Mobile style for Collection editing form overlay


kumar303:
We have a new form overlay that will be used for editing/creating Collections among other things (https://github.com/mozilla/addons-frontend/issues/4198).

We need to implement mobile styles. A spec will follow.

pwalm:
The screen on the right is the mobile user collection and the two on the left are modal options. I'm partial to the one in the middle, it follows some more modern best practices and will grow/contract nicely depending on viewport. Some kinks to work out, but this is a good starting point!

<img width="1260" alt="screen shot 2018-02-01 at 3 17 06 pm" src="https://user-images.githubusercontent.com/6797635/35709042-b9c6a38a-0765-11e8-89c9-dfd97fb26dd5.png">

tofumatt:
The one in the middle is certainly prettier; possibly a bit more work to do generically as the “origin point” would change but we could work toward it. I like them though!

pwalm:
Yeah, i'm just so sick of the look on right. I know we do it on desktop, it's something I'd like to move away from in 2018. More inline input.

bobsilverberg:
I think this issue can be closed. Here is what the edit screen currently looks like at the smallest size:

<img width="333" alt="screenshot 2018-05-17 13 32 39" src="https://user-images.githubusercontent.com/142755/40193950-d7e469b0-59d6-11e8-8b6d-790a573a4d99.png">

What do you reckon @pwalm?

pwalm:
Yep 👍

bobsilverberg:
Great, thanks @pwalm 😄

PR #4356, opened 2018-02-08
UI to edit collection name and description

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4002

Notes:
* This is going to land behind a flag. It will need a lot of follow-ups, like for editing other parts of the collection.
* The mobile styles are happening in https://github.com/mozilla/addons-frontend/issues/4258

screenshot 2018-02-08 17 36 54

screenshot 2018-02-08 17 31 21

### DIFF
--- a/config/dev-amo.js
+++ b/config/dev-amo.js
@@ -24,4 +24,6 @@ module.exports = {
       ],
     },
   },
+
+  enableCollectionEdit: true,
 };
--- a/config/development-amo.js
+++ b/config/development-amo.js
@@ -10,4 +10,5 @@ module.exports = {
   restrictSearchResultsToAppVersion: false,
   fxaConfig: 'local',
   trackingEnabled: false,
+  enableCollectionEdit: true,
 };
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -142,12 +142,7 @@ export class CollectionBase extends React.Component<Props> {
   }
 
   editCollectionLink() {
-    const { _config, hasEditPermission, i18n } = this.props;
-
-    if (!hasEditPermission) {
-      return null;
-    }
-
+    const { _config, i18n } = this.props;
     const props = {};
 
     if (_config.get('enableCollectionEdit')) {
@@ -169,6 +164,7 @@ export class CollectionBase extends React.Component<Props> {
   renderCollection() {
     const {
       collection,
+      hasEditPermission,
       i18n,
       loading,
       location,
@@ -208,7 +204,7 @@ export class CollectionBase extends React.Component<Props> {
               },
             ]}
           />
-          {this.editCollectionLink()}
+          {hasEditPermission ? this.editCollectionLink() : null}
         </Card>
         <div className="Collection-items">
           <AddonsCard
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import config from 'config';
 import * as React from 'react';
 import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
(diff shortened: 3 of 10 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -0,0 +1,176 @@
...
+    super(props);
+    this.state = this.propsToState(props);
+  }
+
+  componentWillReceiveProps(props: Props) {
+    const existingId = this.props.collection && this.props.collection.id;
+    if (props.collection && props.collection.id !== existingId) {
+      // Only reset the form when receiving a collection that the
+      // user is not already editing. This prevents clearing the form
+      // when pressing the submit button.
```
tofumatt: This is fine but would the code be a bit easier to follow if we just had and used an `isSubmitting` flag for this? We could render a "submitting" state this way as well and the check could be a bit clearer.

src/amo/components/CollectionManager/index.js
```diff
@@ -0,0 +1,176 @@
...
+      collection,
+      errorHandler,
+      dispatch,
+      i18n,
+      siteLang,
+      siteUserId,
+    } = this.props;
+
+    if (!collection) {
+      // You'd have to click really fast to access a form without a
```
tofumatt: I like these comments for the edge cases, they're super-helpful 👍

src/amo/components/CollectionManager/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+  return `collection-${collection ? collection.slug : ''}`;
+};
+
+export const mapStateToProps = (
+  state: {| api: ApiStateType, users: UsersStateType |}
+) => {
+  return {
+    siteLang: state.api.lang,
+    siteUserId: state.users.currentUserID,
```
tofumatt: This is just a nitpick but I find `currentUserID` a better variable name to reason about something related to the current user rather than the site; it belongs to the currently signed-in user, not the site... I would rather us use `currentUser` and `currentUserId` everywhere but that's just me.

config/dev-amo.js
```diff
@@ -24,4 +24,6 @@ module.exports = {
       ],
     },
   },
+
+  enableCollectionEdit: true,
```
tofumatt: It might be nice to set this flag as default to on in `development.js` as well... it feels like it should be enabled in development mode no matter what.

src/amo/components/CollectionManager/index.js
kumar303: Since there is already the concept of a *collection user* (she who owns the collection), I think it's helpful to make a distinction of *site user* (she who is using the site). Calling it *current* still seems ambiguous because it could be the current collection owner.

src/amo/components/CollectionManager/index.js
kumar303: This doesn't only apply to submitting, I just used that as an example.

src/amo/components/Collection/index.js
```diff
@@ -124,17 +134,49 @@ export class CollectionBase extends React.Component<Props> {
...
+    const { dispatch } = this.props;
+    event.preventDefault();
+    event.stopPropagation();
+    dispatch(openFormOverlay(COLLECTION_OVERLAY));
+  }
+
+  editCollectionLink() {
+    const { _config, hasEditPermission, i18n } = this.props;
+
+    if (!hasEditPermission) {
```
tofumatt: I feel like this check might be better in the `render()` method as it wasn't clear to me and it would prevent this method from even being called...

tofumatt:
My issue was caused by having my localhost and addons.allizom.org accounts having similar usernames but being signed in using different emails 😱

And my user on the frontend (localhost) was an admin account. This worked when I was the same, non-admin user everywhere editing my own collection. But it looks like admins editing another user's collection gives a 404.

src/amo/components/Collection/index.js
kumar303: Good call.

kumar303:
Good catch. It's ready for another look.

config/development-amo.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
   restrictSearchResultsToAppVersion: false,
   fxaConfig: 'local',
   trackingEnabled: false,
+  enableCollectionEdit: true,
```
tofumatt: 👍 Thanks 😄

src/amo/components/CollectionManager/index.js
```diff
@@ -42,7 +40,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     if (props.collection && props.collection.id !== existingId) {
       // Only reset the form when receiving a collection that the
       // user is not already editing. This prevents clearing the form
-      // when pressing the submit button.
+      // in a few scenarios such as pressing the submit button.
```
tofumatt: Heh, thanks, that helps 😆

Issue #2928, opened 2017-08-11
Show test suite file names in failure output


kumar303:
### Describe the problem and steps to reproduce it:

Trigger a test failure that happens on the event loop. In other words, one that won't produce a stack trace pointing back to the test file.

### What happened?

You have no easy way to figure out which test file produced the error. Example:

<img width="666" alt="screenshot 2017-08-11 10 26 14" src="https://user-images.githubusercontent.com/55398/29220091-aaa16a30-7e7f-11e7-9469-1568ae559127.png">

### What did you expect to happen?

You should be able to easily figure out which test file produced the error.

### Anything else we should know?

We can totally do this now that Jest runs from NodeJS! We can start every test with `describe(__filename, ...)` which will produce errors like this:

<img width="666" alt="screenshot 2017-08-11 10 23 57" src="https://user-images.githubusercontent.com/55398/29220149-cef9cf12-7e7f-11e7-9ac9-fff5fe1756d7.png">

kumar303:
Hah, ignore me. I have *so much console output* (thinking of https://github.com/facebook/jest/issues/4156) that I forgot how Jest prints the failing test name at the top:

<img width="373" alt="screenshot 2017-08-11 11 00 48" src="https://user-images.githubusercontent.com/55398/29221467-7b564dd6-7e84-11e7-9b14-a016ea50ee2b.png">

tofumatt:
That's what I thought happened but figured maybe with a catastrophic failure it wouldn't print.

Nothing to see here then! 😅

kumar303:
I thought about this more and I still think the change would be helpful. What do you think? Here are the benefits I see:

* The bottom line of a failing test will always show the filename. Otherwise, you'd have to scroll up beyond all the console logging to find it.
* We won't have to think of what to type in the initial `describe()` function ever again :) It can just always be `__filename`

tofumatt:
I'm okay with all root-level `describe()`s amounting to `describe(__filename)` but I'd rather there just be a helper–that sort of boilerplate is annoying...

Dunno if that's possible though. I'm cool with doing that though, I can never think of what to name these things!

kumar303:
I don't think a helper would be helpful because you'd have to pass it `__filename` since that is different depending on the file. In other words, a helper would have to look like:

```js
describe(showFile(__filename), ...);
```
which is pointless.

tofumatt:
Oh shoot, duh, right. I didn’t think that one through 😓

PR #4371, opened 2018-02-12
Show LastUpdated date when search results are sorted by "Recently Updated"

### LINKED ISSUE #3783: Show Creation/Update Date when sorting by date-related property
Please add date added or updated to the search results page as in the 'old' version.

I search for new and updated addons daily, so without these dates being initially visible I'd have to keep opening the results to see them.

image

tx,

Ed

bobsilverberg:
Fixes #3783 

This displays a LastUpdated date for each search result if the search included a sort order of "Recently Updated".

Note that the layout and appearance likely needs to change, and I've requested UX feedback in issue #3783. I think the logic can be reviewed at this point as it may need significant changes as well.

### DIFF
--- a/src/amo/components/SearchResult/index.js
+++ b/src/amo/components/SearchResult/index.js
@@ -21,6 +21,7 @@ export class SearchResultBase extends React.Component {
     i18n: PropTypes.object.isRequired,
     showMetadata: PropTypes.bool,
     showSummary: PropTypes.bool,
+    sortedByDate: PropTypes.bool,
   }
 
   static defaultProps = {
@@ -33,8 +34,21 @@ export class SearchResultBase extends React.Component {
     return addon && addon.type === ADDON_TYPE_THEME;
   }
 
+  lastUpdatedText() {
+    const { addon, i18n, sortedByDate } = this.props;
+    const lastUpdated = addon && sortedByDate && i18n.sprintf(
+      // translators: This will output, in English:
+      // "2 months ago (Dec 12 2016)"
+      i18n.gettext('%(timeFromNow)s (%(date)s)'), {
+        timeFromNow: i18n.moment(addon.last_updated).fromNow(),
+        date: i18n.moment(addon.last_updated).format('ll'),
+      }
+    );
+    return lastUpdated && `${i18n.gettext('Last updated')}: ${lastUpdated}`;
+  }
+
   renderResult() {
-    const { addon, i18n, showMetadata, showSummary } = this.props;
+    const { addon, i18n, showMetadata, showSummary, sortedByDate } = this.props;
 
     const isTheme = this.addonIsTheme();
     const averageDailyUsers = addon && addon.average_daily_users;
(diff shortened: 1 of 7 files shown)


src/amo/components/SearchResult/index.js
```diff
@@ -33,8 +34,21 @@ export class SearchResultBase extends React.Component {
...
+    const { addon, i18n, sortedByDate } = this.props;
+    const lastUpdated = addon && sortedByDate && i18n.sprintf(
+      // translators: This will output, in English:
+      // "2 months ago (Dec 12 2016)"
+      i18n.gettext('%(timeFromNow)s (%(date)s)'), {
+        timeFromNow: i18n.moment(addon.last_updated).fromNow(),
+        date: i18n.moment(addon.last_updated).format('ll'),
+      }
+    );
+    return lastUpdated && `${i18n.gettext('Last updated')}: ${lastUpdated}`;
```
kumar303: Due to the way localization works, you can't concatenate two translations. You have to use formatting to make it all in one sting. For example, right-to-left languages would reverse the order of the sentence. I suggest this:

```js
i18n.gettext('Last updated: %(timeFromNow)s (%(date)s)', { date: ... })
```

src/amo/components/SearchResult/styles.scss
```diff
@@ -145,6 +145,8 @@
 }
 
 .SearchResult-author {
+  @include margin-end(10px);
```
kumar303: Were you trying to match something else that is already 10px? I ask because our UI guidelines are to only use multiples of 2. The closest would be 12px. However, there are some old styles that don't follow this.

tests/unit/amo/components/TestSearchResults.js
```diff
@@ -11,6 +11,7 @@ import SearchResults from 'amo/components/SearchResults';
 import I18nProvider from 'core/i18n/Provider';
 import { fakeAddon } from 'tests/unit/amo/helpers';
 import { fakeI18n } from 'tests/unit/helpers';
+import { SEARCH_SORT_UPDATED } from 'core/constants';
 
 
 describe('<SearchResults />', () => {
```
kumar303: This is a pretty old test suite -- while you're in here could you make it `describe(__filename, ...)`? We have [this issue](https://github.com/mozilla/addons-frontend/issues/2928) to fix them all but it will be a lot of work.

src/amo/components/SearchResult/index.js
kumar303: You may want to wait for @pwalm to weight in but I think this is cramming a lot of data into a small space. I would suggest doing something like *Updated 2 days ago* and then including the full date string in a tooltip.

kumar303:
After more discussion in https://github.com/mozilla/addons-frontend/issues/3783 , I don't think we need to add this feature. Even the original person who opened the issue said they don't need it.

bobsilverberg:
#3783 was closed, so this is now obsolete.

PR #4377, opened 2018-02-13
Allow LandingAddonsCard to accept a string for the footerLink property

### LINKED ISSUE #4374: "See more ..." links are not working anymore
Steps to reproduce:
1.Go to amo-dev and try the "See more... " links for "Privacy tools" , "Re-imagine search" and "February’s new featured extensions" collections

Expected results:
Collections are opened without any problems

Actual results:
Nothing happens. Collection cannot be opened.

NOTES:
Tested using Windows 10x64, FF Release, AMO-dev
Not reproducible for "See more popular themes"
Possible related to #4330 

seemore

bobsilverberg:
Fixes #4374 

It should really have been able to do this before, and this fixes the issue so we can now pass simple strings into LandingAddonsCard.footerLink.

Note that I'm fairly sure the way I am checking for the value in the test is wrong, but I wasn't sure how else to verify that the string passed made it into the AddonsCard properly. Feedback is appreciated.

### DIFF
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -35,13 +35,15 @@ export default class LandingAddonsCard extends React.Component {
 
     let footerLinkHtml = null;
     if (addons && addons.length >= placeholderCount) {
-      const linkSearchURL = typeof footerLink === 'string' ?
-        footerLink :
-        {
-          ...footerLink,
-          query: convertFiltersToQueryParams(footerLink.query),
+      let linkTo = footerLink;
+      if (typeof linkTo === 'object') {
+        // As a convenience, fix the query parameter.
+        linkTo = {
+          ...linkTo,
+          query: convertFiltersToQueryParams(linkTo.query),
         };
-      footerLinkHtml = <Link to={linkSearchURL}>{footerText}</Link>;
+      }
+      footerLinkHtml = <Link to={linkTo}>{footerText}</Link>;
     }
 
     return (
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -11,7 +11,7 @@ export default class LandingAddonsCard extends React.Component {
   static propTypes = {
     addons: PropTypes.array.isRequired,
     className: PropTypes.string,
-    footerLink: PropTypes.object.isRequired,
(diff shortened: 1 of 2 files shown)


src/amo/components/LandingAddonsCard/index.js
```diff
@@ -35,10 +35,12 @@ export default class LandingAddonsCard extends React.Component {
 
     let footerLinkHtml = null;
     if (addons && addons.length >= placeholderCount) {
-      const linkSearchURL = {
-        ...footerLink,
-        query: convertFiltersToQueryParams(footerLink.query),
-      };
+      const linkSearchURL = typeof footerLink === 'string' ?
```
kumar303: The old component was simply fixing up the `to` property as a convenience for all the component callers. I think this would better encapsulate the intention of fixing `to` properties:

```js
let linkTo = footerLink;
if (typeof linkTo === 'object') {
  // As a convenience, fix the query parameter.
  linkTo = {
    ...linkTo,
    query: convertFiltersToQueryParams(linkTo.query),
  };
}
footerLinkHtml = <Link to={linkTo}>{footerText}</Link>;
```

A better way to fix this whole thing is to make each calling component pass in their own `<Link/>`. Instead of this:

```js
        <LandingAddonsCard
          addons={featuredExtensions}
          className="Home-FeaturedExtensions"
          header={i18n.gettext('Featured extensions')}
          footerText={i18n.gettext('See more featured extensions')}
          footerLink={{
            pathname: '/search/',
            query: {
              addonType: ADDON_TYPE_EXTENSION,
              featured: true,
            },
          }}
          loading={resultsLoaded === false}
        />
```

Pass in a `<Link/>`:

```js
        <LandingAddonsCard
          addons={featuredExtensions}
          className="Home-FeaturedExtensions"
          header={i18n.gettext('Featured extensions')}
          footerText={i18n.gettext('See more featured extensions')}
          footerLink={
            <Link 
              to={{
                pathname: '/search/',
                query: convertFiltersToQueryParams({
                  addonType: ADDON_TYPE_EXTENSION,
                  featured: true,
                }),
              }} 
            />
          }
          loading={resultsLoaded === false}
        />
```

I'm not sure it's worth the effort though -- just something to think about for future designs.

bobsilverberg: Thanks @kumar303. I made the small fix suggested above so we can get this into the tag. I'll merge it once CI passes. I'll open a separate issue about refactoring LandingAddonsCard.

PR #4384, opened 2018-02-13
Cause incorrect page numbers to return a 404 in "All reviews" pages

### LINKED ISSUE #4081: Incorrect page numbers don't return an error in "All reviews" pages
Steps to reproduce:
1. Load https://addons-dev.allizom.org/en-US/firefox/addon/adblock-edge/reviews/?page=2
2. Change the page number to "0" or a random letter i.e. "x"

Expected results:
A 404 page is displayed.

Actual results:
Page 1 is displayed.

Notes/Issues:
- not reproducing in search results pages;
- not reproducing if the number you enter is bigger than the max page

Verified on AMO-dev FF57(Win 7). Reproducing on all AMO servers.
Video for this issue:
pagen

bobsilverberg:
Fixes #4081 

The problem here is that we are sanitizing the page number via `parsePage` before sending it to the API, which causes any non-numbers and the number 0 to be converted to a 1. When that is passed into `fetchReviews` it of course retrieves the first page of reviews.

Simply removing that sanitization fixes this issue and allows the API to return a 404 as expected, which is similar to how `SearchBase` behaves.

I'm not sure if there are any other repercussions to this change, but everything seems to behave as expected. I do wonder, however, if it might not be a better approach to check the page value before even dispatching `fetchReviews` to avoid making that call if the number is invalid? That's not what `SearchBase` does, but it might be worth considering.

### DIFF
--- a/tests/unit/amo/components/TestAddonReviewList.js
+++ b/tests/unit/amo/components/TestAddonReviewList.js
@@ -215,6 +215,7 @@ describe(__filename, () => {
     });
 
     it('dispatches fetchReviews with an invalid page variable', () => {
+      // We intentionally pass invalid pages to the API to get a 404 response.
       const dispatch = sinon.stub(store, 'dispatch');
       const errorHandler = createStubErrorHandler();
       const addonSlug = fakeAddon.slug;
--- a/tests/unit/amo/components/TestAddonReviewList.js
+++ b/tests/unit/amo/components/TestAddonReviewList.js
@@ -214,6 +214,25 @@ describe(__filename, () => {
       }));
     });
 
+    it('dispatches fetchReviews with an invalid page variable', () => {
+      const dispatch = sinon.stub(store, 'dispatch');
+      const errorHandler = createStubErrorHandler();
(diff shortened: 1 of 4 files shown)


tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -214,6 +214,25 @@ describe(__filename, () => {
       }));
     });
 
+    it('dispatches fetchReviews with an invalid page variable', () => {
+      const dispatch = sinon.stub(store, 'dispatch');
+      const errorHandler = createStubErrorHandler();
+      const addonSlug = fakeAddon.slug;
+      const page = 'x';
```
kumar303: Since it's not totally obvious what this test is about I think it's worth adding a comment to say that we pass invalid pages to the API intentionally to get a 404 response.

bobsilverberg: Good idea. Done!

bobsilverberg:
Thanks for the review, @kumar303. I've opened #4403 as a follow-up.

PR #4391, opened 2018-02-14
Add a query string to the favicon link to force clients to retrieve the newest one

### LINKED ISSUE #4365: The favicon is not correctly displayed on AMO prod
STR:

1. Load AMO-prod - https://addons.mozilla.org/en-US/firefox/
2. Observe the favicon displayed in the page title

Actual result:
The favicon is represented by a green puzzle piece

Expected result:
The favicon is represented by the blue puzzle piece

Notes:
- I see that [#3275](https://github.com/mozilla/addons-frontend/issues/3275) is scheduled for the next push, so it shouldn't be reflected on production, right?
- Reproduced only on AMO-prod, FF58 (WIn10)

image

bobsilverberg:
Fixes #4365 

I was unable to reproduce issue #4365, but @diox was, so we believe it is a caching issue. The favicon is configured to be cached with `Cache-Control: max-age=315360000` so we need to force a refresh on clients who still have the old favicon.

### DIFF
--- a/src/core/containers/ServerHtml.js
+++ b/src/core/containers/ServerHtml.js
@@ -84,7 +84,7 @@ export default class ServerHtml extends Component {
 
   getFaviconLink() {
     const { _config } = this.props;
-    return `${_config.get('amoCDN')}/favicon.ico`;
+    return `${_config.get('amoCDN')}/favicon.ico?v=2`;
   }
 
   render() {
--- a/src/core/containers/ServerHtml.js
+++ b/src/core/containers/ServerHtml.js
@@ -84,7 +84,7 @@ export default class ServerHtml extends Component {
 
   getFaviconLink() {
     const { _config } = this.props;
-    return `${_config.get('amoCDN')}/favicon.ico?v=2`;
+    return `${_config.get('amoCDN')}/favicon.ico?v=${_config.get('faviconVersion')}`;
   }
 
   render() {
--- a/config/default.js
+++ b/config/default.js
@@ -70,6 +70,11 @@ module.exports = {
   apiHost: apiProdHost,
   apiPath: '/api/v3',
 
+  // The version for the favicon.
+  // This should be changed when a new favicon is pushed to the CDN to prevent
+  // client caching.
+  faviconVersion: 2,
+
   // The keys listed here will be exposed on the client.
   // Since by definition client-side code is public these config keys
   // must not contain sensitive data.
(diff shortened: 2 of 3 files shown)


src/core/containers/ServerHtml.js
```diff
@@ -84,7 +84,7 @@ export default class ServerHtml extends Component {
 
   getFaviconLink() {
     const { _config } = this.props;
-    return `${_config.get('amoCDN')}/favicon.ico`;
+    return `${_config.get('amoCDN')}/favicon.ico?v=2`;
```
tofumatt: Could we do this in a more automated fashion? This seems a bit brittle.

bobsilverberg: Do we want a new string to be used for each build, or do we just need to bust the cache this once (for now) because we've got a new icon?

tofumatt: Well, updating it only when we change the icon means benefiting from more caching, but we already invalidate the rest of our (much larger) assets on a new build so I don't think caching only the favicon makes much sense. The favicon is tiny and I think browsers already aggressively cache favicons for sites. And this way when it's changed no other configs/tests need changing; it will Just Work™.

bobsilverberg: Good point. It sounds like the right thing to do. I'll give it a shot.

bobsilverberg: I've looked into this and I'm not sure how to get webpack to help us out here. You suggested that our webpack build will append cache-busting strings to URLs, but I don't see any evidence of that, unless you mean the webpack generated assets like javascript, css, images, etc. I don't think we'll be asking webpack to generate the favicon, as it's being served from the cdn, so what we're currently doing with webpack assets doesn't seem like a good fit.

We could do something as simple as using `Date.now()` as a query string, or, if we can get the webpack build id somehow in `ServerHtml.js` we could use that. I'm not sure how to do the latter. What do you suggest, @tofumatt?

tofumatt: It's worth noting that our build happens on multiple servers right now, which isn't ideal but means using something like `Date.now()` could result in the same page serving different content. The webpack build is made from a hash of contents IIRC though so it should be consistent; storing it someplace might be nice.

I thought all of our links to things like CSS, JS, etc. were processed by webpack somehow. The asset is served from the CDN like all of our resources, but I think the URLs to those resources are still built by webpack.

muffinresearch:
I think it will be fine to move forward with a simple fix for now. The better long term fix can happen once the favicon moves into the project in https://github.com/mozilla/addons-frontend/issues/4407

PR #4406, opened 2018-02-16
Increase the possible height of the add-on description on desktop

### LINKED ISSUE #3787: expand add-on description by default on desktop
Please expand, at least on desktop, the add-on description by default. I don't see any good reason to hide the most important part of the add-on page by default and it's annoying to have to click on every add-on page the "read more" link to read the description.

bobsilverberg:
Fixes #3787 

I know that this patch is highly flawed, but I figured providing some sample code would help illustrate the problems I'm struggling with and make the questions I'm going to ask below more clear...

We want to increase the amount of space that the `ShowMoreCard` allows on desktop, for the add-on description. I increased the allowable space by adding a rule for `respond-to(large)`, which seems to work, but I'm not sure if that's the correct approach. Are we equating desktop with `large`? What about `medium`?

Assuming the above is okay, the other part of the problem is deciding when to show the "Read more" footer. The current code has 150px hardcoded as a limit in the `ShowMoreCard` component, but now the limit is different between desktop and mobile. I tried to solve this with a css-only solution, but couldn't figure out how to do so. I think a css-only solution would be preferable, if it's possible. Not being able to figure that out, I decided to make `maxHeight` a property of the `ShowMoreCard` component, so that we could have a different `maxHeight` value for different instances (which we need). I then tried to figure out if we were on desktop or not in the `Addon` component so I could pass the correct `maxHeight` into the `ShowMoreCard` component. I did this by looking at `userAgentInfo`, but I think that's flawed. I'm also not happy about having to keep the correct maxHeight value in sync between the css file and the js file, but I guess we're already doing that with the current 150pm value.

So, do you have any suggestions for a better approach to this, or if my approach is not terribly off base, how can I improve it as I still think it's quite flawed?

### DIFF
--- a/src/ui/components/ShowMoreCard/ShowMoreCard.scss
+++ b/src/ui/components/ShowMoreCard/ShowMoreCard.scss
@@ -40,7 +40,7 @@
     }
   }
 
-  footer.Card-footer-link {
+  .Card-footer-link {
     @include respond-to(large) {
       display: none;
     }
--- a/src/ui/components/ShowMoreCard/ShowMoreCard.scss
+++ b/src/ui/components/ShowMoreCard/ShowMoreCard.scss
@@ -1,4 +1,5 @@
 @import "~core/css/inc/vars";
+@import "~core/css/inc/mixins";
 
 .ShowMoreCard-contents {
   &::after {
@@ -25,5 +26,23 @@
     }
 
     max-height: none;
+  }
+}
+
+.AddonDescription {
+  .ShowMoreCard-contents {
+    @include respond-to(large) {
+      &::after {
+        height: 0;
+      }
+
+      max-height: none;
+    }
+  }
+
+  footer.Card-footer-link {
+    @include respond-to(large) {
+      display: none;
+    }
   }
 }


bobsilverberg:
Note that the first commit was my initial attempt to remove the Read more entirely and have no limit, but we decided that we wanted to keep some limit on desktop. You can ignore that commit and just look at the combined commits.

bobsilverberg:
Thanks for the tips, @kumar303. I was able to change from using `MAX_HEIGHT` to using  `scrollHeight > clientHeight` and that part is working, but I am struggling to get the css to work. When the Read more link is clicked, the `max-height` is not updated to `none` as we wish. It looks like the rule I added is taking precedence over that rule. I've tried a whole bunch of different ways of specifying/structuring the rules, but none of them seem to work.  I tried making the new rule active only when the section is not expanded using `.AddonDescription:not(.ShowMoreCard--expanded) {` in place of what's there now, but that only succeeded in having the section always be expanded from the get go.

Do you have any advice about what I might try?

bobsilverberg:
@kumar303 I figured out the problem and have come up with a solution that works, but I'm not sure if it's a good approach. The problem was that the logic in [`truncateToMaxHeight` in `ShowMoreCard/index.js`](https://github.com/bobsilverberg/addons-frontend/blob/read-more-3787/src/ui/components/ShowMoreCard/index.js#L42)  was getting triggered when the component was first rendered, without any content (because of lazy loading), and that was causing the state to be `expanded` at that time. We need to wait for the content to load, and yet still have the state as `not expanded` so the test can actually work. I did this by storing an initial size and then checking that in the logic of `truncateToMaxHeight` but I'm worried that's a bit fragile, and depends on the lifecycle.

Please let me know what you think of the approach and if it's not a good one if you have any suggestions for improving it.

src/ui/components/ShowMoreCard/ShowMoreCard.scss
```diff
@@ -1,4 +1,5 @@
 @import "~core/css/inc/vars";
+@import "~core/css/inc/mixins";
```
bobsilverberg: Oops, I'll fix this.

bobsilverberg:
@kumar303 I did do a bit of experimenting with `componentDidUpdate` and found at first that I created an infinite loop by calling `setState` inside `componentDidUpdate`, because `setState` triggers another `componentDidUpdate`. So I need some sort of guard to only call `setState` in `componentDidUpdate` if I need to (e.g., if the size of the component has changed). I still haven't gotten to the point where this actually solves my problem, but it's a good next path to take. I will do as you suggest and leave this one until I'm done with the permissions issue and then spend some more time on it.

tofumatt:
@bobsilverberg Is there any update on this one? From the look of the code it seems to be being clever and expanding the content to the size of the screen/viewport, but in reality I think just *not* condensing it on desktop (basically any size where there's nothing underneath it) for now is fine. If the content isn't all one card on top of another (eg mobile and small tablets) then it should just be expanded.

bobsilverberg:
@tofumatt I was never able to figure out how to get it to work "properly" and was spending so much time on it that I decided to leave it and move on to other things. Simply not condensing it on desktop sounds like a simpler fix, but how do we know if we are on desktop? Is there something I can check in the code for the component, as opposed to relying on css rules for this? Or can you suggest a way to implement that logic via the css?

tofumatt:
My recollection is that the JS is just setting `className`s, not directly modifying the CSS. Just make the desktop version of the class (`@inculde respond-to(large) {}` never have hidden overlflow-y: hidden and no height set. I'd think that'd work.

bobsilverberg:
Thanks @tofumatt. Looking at this again, I realize that I can just go back to the very first commit that I added to this PR, which simply expands the description unconditionally on desktop. I had an issue with the gradient at the bottom of the box, but I've fixed that in the latest commit.

The discussion in issue #3787 did suggest we should use some limit (https://github.com/mozilla/addons-frontend/issues/3787#issuecomment-365093414), but perhaps this is a good enough solution for now, considering a number of people want to see this fixed, and increasing the limit on desktop seems to be a much harder thing to accomplish.

src/ui/components/ShowMoreCard/ShowMoreCard.scss
```diff
@@ -27,3 +28,21 @@
...
+    @include respond-to(large) {
+      &::after {
+        height: 0;
+      }
+
+      max-height: none;
+    }
+  }
+
+  footer.Card-footer-link {
```
muffinresearch: Is this only ever used for "show more" purposes? Does it actually need `footer` for specificity?

bobsilverberg: You're correct, it doesn't need `footer`. I will remove it.

Issue #4393, opened 2018-02-15
Send a parameter in API requests indicating that the API should wrap URLs with outgoing.*.mozaws.net


diox:
## User story

* As a site user, go to an add-on detail page (or a similar page) that lists an external link in its metadata. For example, *homepage* is an external link on the [AdBlock Plus](https://addons.mozilla.org/en-US/firefox/addon/adblock-plus/) page.
* Click on the external link
* The request should be served by https://outgoing.prod.mozaws.net/ (or the equivalent dev/stage site) before redirecting to the final destination

## Context

We want to start returning unwrapped URLs by default in the API, instead of having all of them go through `outgoing.*.mozaws.net`. We still want that wrapping to be active for addons-frontend though, so it needs to send addons-server a parameter that the API will recognize that will enable the wrapping.

See https://github.com/mozilla/addons-server/issues/7513 for more details

vcarciu:
Tested for regressions, everything looks fine.

PR #4408, opened 2018-02-16
Explicitly ask API to wrap outgoing URLs

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4393 by configuring all API requests with the new parameter.

Because all API requests have a new query param, this broke a lot of tests that were too tightly coupled to the entire URL. I fixed all of those by making the tests ignore parts of the URL that were not relevant. This is something we've been meaning to do for a while now. I thought there was an issue on file but I couldn't find it.

Sorry for the noise but this cleanup should make the tests more robust.

### DIFF
--- a/tests/unit/core/api/test_index.js
+++ b/tests/unit/core/api/test_index.js
@@ -1,6 +1,5 @@
 /* global window */
 import querystring from 'querystring';
-import url from 'url';
 
 import config from 'config';
 import utf8 from 'utf8';
@@ -13,6 +12,7 @@ import {
 import {
   createFakeAutocompleteResult,
   dispatchClientMetadata,
+  dispatchSignInActions,
 } from 'tests/unit/amo/helpers';
 import {
   apiResponsePage,
@@ -22,9 +22,8 @@ import {
   getFakeConfig,
   signedInApiState,
   unexpectedSuccess,
-  userAuthToken,
+  urlWithTheseParams,
 } from 'tests/unit/helpers';
-
 
 describe(__filename, () => {
   let mockWindow;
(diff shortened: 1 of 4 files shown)


tests/unit/core/api/test_index.js
```diff
@@ -39,64 +38,78 @@ describe(__filename, () => {
...
+      await api.callApi({ endpoint: 'resource', method: 'get' });
+      mockWindow.verify();
+    });
+
+    it('adds some query string params to all requests', async () => {
+      const lang = 'en-US';
+      const { state } = dispatchClientMetadata({ lang });
+
+      mockWindow.expects('fetch')
+        .withArgs(urlWithTheseParams({ lang, wrap_outgoing_links: true }))
```
tofumatt: This is cool! 👍

tests/unit/helpers.js
```diff
@@ -431,3 +433,25 @@ export const getFakeConfig = (params = {}) => {
...
+ * Example:
+ *
+ * mockWindow.expects('fetch').withArgs(urlWithTheseParams({ page: 1 }))
+ */
+export const urlWithTheseParams = (params) => {
+  return sinon.match((urlString) => {
+    const { query } = url.parse(urlString, true);
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const param of Object.keys(params)) {
```
tofumatt: If we're doing a loop anyway wouldn't it be a better habit to just do `for (const param in params) {` which doesn't create a new array? We can already iterate through params. I get you'll want to check for `hasOwnProperty` then, I guess it's up to you 🤷‍♂️ 

Your way looks nicer but is I guess less efficient. That probably doesn't matter for the tests much but I don't think this is a case where the faster option is harder to read/reason about.

PR #4413, opened 2018-02-19
fix: Use href instead of to for Links in header

### LINKED ISSUE #4397: View/Edit Profile options will trigger a Page not found
STR:

1. Log into AMO-dev
2. Click on View/Edit profile in the user menu

Actual result:
The user is redirected to a Page not found

Expected:
The user is redirected to view/edit his profile

Notes: 
- not sure if this has any connection to [#4342](https://github.com/mozilla/addons-frontend/issues/4342)
- reproduced on AMO-dev, FF58 (Win10x64, Android 7.0)
- the profile page will be displayed after page refresh

Video:
profile error

tofumatt:
(fix #4397)

Looks like this slipped in unintentionally, sorry about that.

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -101,12 +101,12 @@ export class HeaderBase extends React.Component {
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
-                <Link to={`/user/${siteUser.username}/`}>
+                <Link href={`/user/${siteUser.username}/`}>
                   {i18n.gettext('View Profile')}
                 </Link>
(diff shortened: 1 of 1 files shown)


tofumatt:
That’s how the React Router link component works, we just extend its functionality.

I can add a test for this, sure. We’re gonna change it soon but yeah good point :-)

- Matt (Sent from mobile)

kumar303:
> I can add a test for this

I don't think it's worth it since this code will change when the new profile pages land. I think it would be more helpful to add a comment or switch to a new `ExternalLink` component as I mentioned. This could be a follow-up commit.

Issue #2757, opened 2017-07-12
Display permissions on add-on detail page


tofumatt:
On the add-on details page, show all permissions for this add-on with helpful information about what each one means.

muffinresearch:
This will be web-ext only.

kumar303:
I'm nominating this for MVP removal because Firefox already shows you a list of permissions before you install the add-on.

tofumatt:
Yeah, I actually asked about removing it altogether but @pwalm pointed out user research that showed knowing permissions beforehand helped users understand the permissions more and think twice about installing an add-on. So we will want to keep it–if anything we should remove it from installs made on AMO. But that’s another issue 😅

muffinresearch:
This one we need to keep and it's fairly small.

willdurand:
The mock-ups show a link for this information. Where do we have to link? The current (old) website displays a popup.

Second question: the API returns the name of the permissions, but is there any doc about how to translate these terms into sentences people can understand? This is what the popup on the old website does. Should we use the webext doc?

tofumatt:
Hmm, I don't know 😢

These seem like UX questions–do we not show permissions anywhere in the mocks?

If we don't have mocks for this feature and Firefox already shows them I would take this out of MVP. Yes, I understand there are benefits to showing them pre-install... but we'll want to know how to display them 😄

willdurand:
ping @pwalm for UX info about this.

muffinresearch:
Ok I've reviewed dropping this from the MVP with @jvillalobos and @deadsquid and we're good to do so as long as it's prioritized towards the top of the post release features. If all other MVP stuff is done and we still have time and the UX is ready this can be done pre-release.

jvillalobos:
Ping @pwalm for UX on this. It's a high priority in the post-release issues.

jvillalobos:
Also pinging @designakt and @brassy-, since Philip is on PTO.

designakt:
It looks like currently we do not even show the link to the list of permissions. The current mock shows a link to all permissions that should be opened in an overlay. Much like in the current AMO.
![amo-permission-link](https://user-images.githubusercontent.com/8145891/32455567-ee2bee8c-c322-11e7-8ff8-f410d228363c.PNG)

[...]

PR #4427, opened 2018-02-21
Add a platformFiles map to internal addon object and use it to replace installURLs

### LINKED ISSUE #4426: Add a platformFiles map to internal addon object and use it to replace installURLs
This is a spin-off from issue #2757 because that issue is quite large. It will require a `platformFiles` map in order to get the correct permissions for an addon, and that `platformFiles` map can also be used to fulfill the purpose of the current `installURLs` map.

This issue is about creating the `platformFiles` map and removing the `installURLs` map.

bobsilverberg:
Fixes #4426 

As noted in the issue, this is a blocker for issue #2757.

### DIFF
--- a/tests/unit/core/reducers/test_addons.js
+++ b/tests/unit/core/reducers/test_addons.js
@@ -101,21 +101,7 @@ describe(__filename, () => {
       ...extension,
       iconUrl: extension.icon_url,
       platformFiles: {
-        [OS_ALL]: {
-          created: '2014-11-22T10:09:01Z',
-          hash: 'a1b2c3d4',
-          id: 57721,
-          is_mozilla_signed_extension: false,
-          is_restart_required: false,
-          is_webextension: true,
-          permissions: [
-            'activeTab',
-            'webRequest',
-          ],
-          platform: 'all',
-          status: 'public',
-          url: 'https://a.m.o/files/321/addon.xpi',
-        },
+        [OS_ALL]: fakeAddon.current_version.files[0],
         [OS_ANDROID]: undefined,
         [OS_LINUX]: undefined,
         [OS_MAC]: undefined,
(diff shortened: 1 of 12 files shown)


tests/unit/core/reducers/test_addons.js
```diff
@@ -100,8 +100,22 @@ describe(__filename, () => {
     expect(state.byID[extension.id]).toEqual({
       ...extension,
       iconUrl: extension.icon_url,
-      installURLs: {
-        [OS_ALL]: 'https://a.m.o/files/321/addon.xpi',
+      platformFiles: {
+        [OS_ALL]: {
```
kumar303: This test data needs to inherit from a fake file object so that you don't have to repeat all fields of data. We have currently been using [`createFakeAddon().current_version.files[0]`](https://github.com/bobsilverberg/addons-frontend/blob/89f906bbbf0d51dd6f514e5df92746027820a1c3/tests/unit/amo/helpers.js#L299) as a base but maybe we need `createFakeFile()` for clarity.

tests/unit/core/reducers/test_addons.js
```diff
@@ -201,21 +229,67 @@ describe(__filename, () => {
     });
     const state = addons(undefined,
       loadAddons(createFetchAddonResult(addon).entities));
-    expect(state.byID[addon.id].installURLs).toMatchObject({
-      [OS_MAC]: 'https://a.m.o/mac.xpi',
-      [OS_WINDOWS]: 'https://a.m.o/windows.xpi',
-      [OS_ALL]: 'https://a.m.o/all.xpi',
+    expect(state.byID[addon.id].platformFiles).toMatchObject({
```
kumar303: These tests were only checking the platform logic so I'm not sure you need to match the entire object. I would suggest picking a field that will cover just the platform mapping logic.

tests/unit/core/reducers/test_addons.js
bobsilverberg: This was a pretty easy fix using `fakeAddon` in the test, and I don't really think we need a `createFakeFile()` currently, so I just went with the quick fix.

Issue #4270, opened 2018-01-31
Allow external tracking sources through to install buttons


muffinresearch:
Refiled copy of https://github.com/mozilla/addons/issues/615 - see that issue for the original issue details + additional comments.

### Describe the problem and steps to reproduce it:
1. Open the add-on's details page with a tracking parameter, like: `https://addons.mozilla.org/en-US/firefox/addon/example?src=external-example`.
2. Open the "Downloads > by Source" statistics page (`https://addons.mozilla.org/en-US/firefox/addon/example/statistics/downloads/sources/`) and see the table.

### What happened?

The pageview in the first step is counted in "Detail Page" category.

### What did you expect to happen?

It should be counted in a column called `external-example`.

### Anything else we should know?

**For QA**: in addition to verifying the feature, please check for regressions in the install process for all extensions on AMO and the Discopane.

![image](https://user-images.githubusercontent.com/6646473/34776395-b3398cb4-f651-11e7-85b8-d6895bd49995.png)

Here's a [copy of the comment](https://github.com/mozilla/addons/issues/615#issuecomment-358707507) I posted that details what seems to be happening and the root cause of the bug.

If you visit a details page with a src query param e.g: https://addons.mozilla.org/en-US/firefox/addon/privacy-badger17/?src=external-test what is sent when installing the add-on is src=dp-btn-primary whereas I think the expectation is that it would be src=external-test. (You can see this when inspecting network in the browser console)

Looking into the code, as you noted it does look like the src param is extracted from the URL here: https://github.com/mozilla/addons-frontend/blob/6447fa8b0fbea6c14ffdeb879e059b8ab3e0ce35/src/core/installAddon.js#L274

However, it looks like the install function overrides that src here: https://github.com/mozilla/addons-frontend/blob/6c7cc717af0905f310c96f0d7b19eb0952a9dfbd/src/core/addonManager.js#L52

See this line for where this is called: https://github.com/mozilla/addons-frontend/blob/6c7cc717af0905f310c96f0d7b19eb0952a9dfbd/src/core/installAddon.js#L404 - as you can see we get the url via findInstallURL which will add a src, however the _addonManager.install() then overrides it.

jvillalobos:
https://github.com/mozilla/addons-frontend/issues/4372 has details on what we expect of the `src` parameter.

vcarciu:
[...]

PR #4436, opened 2018-02-22
Preserve external tracking source on install

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4270

I tested locally (AMO and Discopane) and it seems to be working.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -64,7 +64,7 @@ export class AddonBase extends React.Component {
     addon: PropTypes.object.isRequired,
     clientApp: PropTypes.string.isRequired,
     // This prop is passed in by withInstallHelpers()
-    defaultSource: PropTypes.string.isRequired,
+    defaultInstallSource: PropTypes.string.isRequired,
     dispatch: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
     getClientCompatibility: PropTypes.func,
@@ -435,7 +435,7 @@ export class AddonBase extends React.Component {
       addon,
       addonsByAuthors,
       clientApp,
-      defaultSource,
+      defaultInstallSource,
       errorHandler,
       getClientCompatibility,
       i18n,
@@ -556,7 +556,7 @@ export class AddonBase extends React.Component {
                     {...this.props}
                     disabled={!isCompatible}
                     ref={(ref) => { this.installButton = ref; }}
-                    defaultSource={defaultSource}
+                    defaultInstallSource={defaultInstallSource}
                     status={installStatus}
                     useButton
                   /> : null
@@ -661,6 +661,6 @@ export default compose(
   withRouter,
   translate({ withRef: true }),
   connect(mapStateToProps),
-  withInstallHelpers({ defaultSource: 'dp-btn-primary' }),
+  withInstallHelpers({ defaultInstallSource: 'dp-btn-primary' }),
   withFixedErrorHandler({ fileName: __filename, extractId }),
 )(AddonBase);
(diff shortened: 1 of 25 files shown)


kumar303:
See https://github.com/mozilla/addons-frontend/issues/4437 about test failures, which appear unrelated to my patch.

src/amo/components/Addon/index.js
```diff
@@ -62,6 +63,8 @@ export class AddonBase extends React.Component {
     RatingManager: PropTypes.element,
     addon: PropTypes.object.isRequired,
     clientApp: PropTypes.string.isRequired,
+    // This prop is passed in by withInstallHelpers()
```
tofumatt: 👍 documenting stuff like this, especially when it's related to the install helper, is really handy. Thanks!

src/core/installAddon.js
```diff
@@ -228,24 +233,42 @@ const userAgentOSToPlatform = {
...
     throw new Error('The userAgentInfo parameter is required');
   }
 
+  let source;
+  if (appendSource) {
+    if (!location) {
+      throw new Error(
+        'The location parameter is required when appendSource is true');
+    }
+    source = location.query.src || defaultSource;
```
tofumatt: I get what this is doing but the logic might feel a bit clearer to me if it were restructured. As you're already using a `let` and redefining `source` (which I think is fine here), maybe:

```js
let source = defaultSource;
if (appendSource) {
  if (!location) {
    throw new Error(
      'The location parameter is required when appendSource is true');
  }
  source = location.query.src;
}
```

I think that's just a bit easier to follow.

tofumatt: Oh wait, I see... if `appendSource` is `false` we shouldn't be setting it at all. Should that ever really happen though? If so... I guess the original code is cool but I just had to read it a few times to totally follow it.

src/core/utils/compatibility.js
```diff
@@ -113,7 +113,9 @@ export function isCompatibleWithUserAgent({
     if (
       addon.type === ADDON_TYPE_EXTENSION &&
       !findInstallURL({
-        platformFiles: addon.platformFiles, userAgentInfo,
+        appendSource: false,
```
tofumatt: Ah, okay, I see... seems we do set this as false a fair bit. So I guess ignore my earlier comment about restructuring the `let source` bit 😅

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -288,7 +289,7 @@ describe('RatingManager', () => {
 
     it('renders an AuthenticateButton', () => {
       const AuthenticateButton = sinon.spy(() => <div />);
-      const location = { pathname: '/some/path/' };
+      const location = fakeRouterLocation();
```
tofumatt: All this is great, thanks 👍

tofumatt:
Rebasing with master should fix the test failures, FYI.

src/amo/components/Addon/index.js
kumar303: I find the pattern of introducing magical props with an HOC very confusing. I would like to avoid it in the future. I want `withInstallHelpers()` to go away.

tofumatt: Ultimately I agree, but this comment is a good stop-gap and will let us refactor it away in the future 😄

kumar303:
> This is all good but reading through I'm realising what src was always meant to be and I don't think `source` and `defaultSource` help clarify it much.

Yep, I agree. I think I can rename it. Let me try...

kumar303:
I'm gonna merge ...

california

Issue #4326, opened 2018-02-06
Rename taarParamsToUse to discoParamsToUse or similar


muffinresearch:
Since the params passed to the discovery API are not all specifically just for TAAR we should now rename it to reflect reality.

kumar303:
Thanks for filing -- I was looking at this code and thinking to myself, hmm, what is TAAR again? It could probably use some more comments too.

tofumatt:
@KhushilMistry is working on this and has a PR ready–it just needs one more tweak and should be good to merge.

PR #4441, opened 2018-02-23
Rename taarParamsToUse to discoParamsToUse

KhushilMistry:
Issue #4326 
Any thing else needs to done ?

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -26,7 +26,7 @@ module.exports = {
     'langMap',
     'publicSentryDsn',
     'rtlLangs',
-    'taarParamsToUse',
+    'discoParamsToUse',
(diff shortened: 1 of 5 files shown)


config/default-disco.js
```diff
@@ -52,7 +52,7 @@ module.exports = {
       mediaSrc: [staticHost],
     },
   },
-  taarParamsToUse: [
+  discoParamsToUse: [
```
tofumatt: The issue said this could use more comments; I agree. I think here adding:

```
// Firefox sets these query params when loading the Discovery Pane.
// These are the allowed list of query params we will forward to AMO for statistics.
discoParamsToUse: [
...
```

would be great.

Issue #3525, opened 2017-10-16
Rating star shows IDs for readonly={true} ratings


tofumatt:
The ratings star component will output an `id` with `Rating-rating-${number}`–this is an issue on the detail page where a readonly and not readonly rating are displayed (one in the top right and your personal rating in the "rate this add-on" card.

It means we have duplicate IDs on the page which is naughty! We probably never need to render out an ID so we should just remove them.

KhushilMistry:
@tofumatt I want to work on this. Should I remove id for both the readonly div and button or any particular one ?

tofumatt:
The IDs should only be removed when `readOnly` is `true`.

willdurand:
@kumar303 is it still an issue given that the `Rating` component has changed recently?

PR #4446, opened 2018-02-23
Rating star ID removed for readonly={true} ratings

KhushilMistry:
Issue #3525

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -62,16 +62,14 @@ export class RatingBase extends React.Component {
           'Rating-half-star': (rating && thisRating > rating &&
             thisRating - 0.5 <= rating),
         }),
-        id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
       };
 
       if (readOnly) {
-        delete props.id;
         return <div {...props} />;
       }
-
+      props.id = `Rating-rating-${thisRating}`;
       return (
         // eslint-disable-next-line react/jsx-key
         <button
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -68,6 +68,7 @@ export class RatingBase extends React.Component {
       };
 
       if (readOnly) {
+        delete props.id;
         return <div {...props} />;
       }
 
--- a/tests/unit/ui/components/TestRating.js
+++ b/tests/unit/ui/components/TestRating.js
@@ -175,6 +175,19 @@ describe('ui/components/Rating', () => {
     expect(findDOMNode(root).textContent).toEqual('Rated 2 out of 5');
   });
 
+  it('renders star button with id in ratings', () => {
+    const root = render();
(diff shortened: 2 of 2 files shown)


src/ui/components/Rating/index.js
```diff
@@ -68,6 +68,7 @@ export class RatingBase extends React.Component {
       };
 
       if (readOnly) {
+        delete props.id;
```
tofumatt: I'm less a fan of deleting the prop and rather only adding the ID to the object if `readOnly` is false.

Also, you'll need to add a test for this so you can make sure it's working as intended 😄

KhushilMistry:
@tofumatt I have done required changes. Can you review it ?

src/ui/components/Rating/index.js
```diff
@@ -62,15 +62,14 @@ export class RatingBase extends React.Component {
           'Rating-half-star': (rating && thisRating > rating &&
             thisRating - 0.5 <= rating),
         }),
-        id: `Rating-rating-${thisRating}`,
```
tofumatt: I guess I said conditionally add, but it'd be nice to keep all the props in one place. Could this be:

```js
id: !readOnly ? `Rating-rating-${thisRating}` : null,
```
If so I think that'd be the easiest to follow.

tests/unit/ui/components/TestRating.js
```diff
@@ -175,6 +175,19 @@ describe('ui/components/Rating', () => {
     expect(findDOMNode(root).textContent).toEqual('Rated 2 out of 5');
   });
 
+  it('renders star button with id in ratings', () => {
+    const root = render();
+    const elementKeys = Object.keys(root.ratingElements);
```
tofumatt: This is a bit of a roundabout way to test for IDs, and there are helpers we have to look for IDs that would make this clearer.

You should be able to simply verify the IDs exist in one test (one should already exist?) and then use `expect(findDOMNode(root).querySelector('#Rating-ID')).toHaveLength(0)` or something similar.

KhushilMistry: Can I use this ? 
``
expect(findDOMNode(root).querySelector('.Rating-choice').id).toHaveLength(0); 
``
But I have doubt that it is checking id for only the first Star.
I also thought of an alternate way like this

``
[1, 2, 3, 4, 5].forEach((rating) => {
        expect(root.ratingElements[rating].id).toHaveLength(0);
});
``

What should I use ?

tofumatt: Sorry for my late reply, but unfortunately those tests aren't getting at what we're looking to check for. I don't think I'll have much time to mentor this issue further, but thanks for your efforts here 👍

KhushilMistry: So should I close this PR ?

tofumatt: It might be best to close it if this hasn't moved forward, but if you'd like to work on another issue there are more issues marked as contributor-friendly. Sorry about this one getting stalled.

KhushilMistry: Okay No Worries. Will close this PR and try to work on some other. Thanks :)

Issue #4375, opened 2018-02-13
"&" is interpreted as "&amp" in Edit collection description


AlexandraMoga:
Follow up from [#4002](https://github.com/mozilla/addons-frontend/issues/4002)

STR:

1. Log into AMO-dev as an admin
2. Select the "Ad Blockers' collection - https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/ad-blockers/
3. Click on Edit this collection
4. Observe the description available in the Edit description section

Actual result:
"&"s are interpreted as "&amp"s

Expected result:
There are no text formatting issues around Edit description for a collection

Notes:
- there was a similar issue addressed here [#4213](https://github.com/mozilla/addons-frontend/issues/4213)
- Reproduced on AMO -dev with FF58, Win10 and Android 7.0

![image](https://user-images.githubusercontent.com/31961530/36154985-cbe12606-10db-11e8-837b-39d6f6ba11b7.png)

vcarciu:
Also "<script>alert('problem')</script>" is interpreted as "&lt;script&gt;alert('problem')&lt;/script&gt;"

AlexandraMoga:
Verified as fixed on AMO-dev, FF58 (Win10x64 and Android 7.0)

![image](https://user-images.githubusercontent.com/31961530/36718873-f8c9dc6e-1bab-11e8-8edc-80503a718cac.png)

PR #4447, opened 2018-02-23
Deal with HTML entities when editing a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4375

Before:

screenshot 2018-02-23 16 53 26

After: 

screenshot 2018-02-23 16 52 42

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -301,7 +301,10 @@ export function addonHasVersionHistory(addon) {
   ].includes(addon.type);
 }
 
-export const stripHtmlEntities = (string) => {
+/*
+ * Decodes HTML entities into their respective symbols.
+ */
+export const decodeHtmlEntities = (string) => {
   const entities = new AllHtmlEntities();
   return entities.decode(string);
 };
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -1,6 +1,7 @@
 /* eslint-disable react/prop-types */
 import url from 'url';
 
+import { AllHtmlEntities } from 'html-entities';
 import config from 'config';
 import * as React from 'react';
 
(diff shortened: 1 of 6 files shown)


src/core/utils/index.js
```diff
@@ -299,3 +300,8 @@ export function addonHasVersionHistory(addon) {
     ADDON_TYPE_THEME,
   ].includes(addon.type);
 }
+
+export const stripHtmlEntities = (string) => {
+  const entities = new AllHtmlEntities();
```
tofumatt: Interesting... previously [we just used `sanitizeHTML()` with no allowed tags](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/Collection/index.js) to filter HTML entities. That's a bit more of an obscured use of that helper but it means we didn't need an extra library.

I'd be tempted to just make this helper that use `text => sanitizeHTML(text, [])` under the covers to allow HTML entities. That way we don't have to include and bundle an extra library. If the tests passed when this helper looked like:

```js
export const stripHtmlEntities = (string) => {
  return sanitizeHTML(string);
}
```

then I think that would be the better route to go.

kumar303:
> previously we just used sanitizeHTML() with no allowed tags to filter HTML entities.

What you linked to in `Collection` is something different. It takes API description data which may contain HTML and renders that HTML safely (by stripping out bad tags). That's all it does.

> I'd be tempted to just make this helper use text => sanitizeHTML(text, []) under the covers to allow HTML entities.

What I need to here is different than what `Collection` does. When I get HTML from the API, I need to convert the HTML entity markup into symbols so the user can see the real symbols (not the HTML) while editing the text. 

Your proposed solution doesn't work because it sees HTML (such as `&amp;`), it knows that this is harmless markup, and leaves it untouched. What I want to do instead is transform `&amp;` into `&` which the [dom purify](https://github.com/cure53/DOMPurify#what-does-it-do) library does not do.

I think `stripHtmlEntities` was a bad name and maybe that caused confusion. I changed it. Can you take another look?

tofumatt:
Oh shoot, you're totally right, I got my wires crossed thinking about the scenarios here. My bad.

Issue #4434, opened 2018-02-22
Remove "slug" word from error messages in edit collections


vcarciu:
Steps to reproduce:
1.Go to amo-dev , login as admin and open any collection (https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/ad-blockers/)
2.Click on "Edit this collection" link
3.Modify the slug with something not allowed (special chars)
4.Save the changes

Expected results:
The error messages is not containing the word "slug" at beginning

Actual results:
The error messages is containing the word "slug" at beginning

Please see a screenshot with the issue:
![slug](https://user-images.githubusercontent.com/1583842/36547281-27967e9e-17f6-11e8-92f7-d89bb9a624ec.png)

vcarciu:
Verified as fixed : 
![sluggy](https://user-images.githubusercontent.com/1583842/36721830-b4d50b38-1bb4-11e8-806b-c246659325d6.png)

PR #4457, opened 2018-02-26
Do not prefix form errors with field keys

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4434


src/core/reducers/errors.js
```diff
@@ -38,16 +38,10 @@ function getMessagesFromError(error) {
         // More info:
         // http://addons-server.readthedocs.io/en/latest/topics/api/overview.html#bad-requests
         value.forEach((msg) => {
```
tofumatt: I know this isn't related to your patch but can we expand this variable name? `message` would be nicer and we have the disk space for those extra bytes 😉

Issue #3219, opened 2017-09-22
Rename denormalizeReview() to createInternalReview()


kumar303:
A pattern has emerged where we take an external API object and store it with a function like [`createInternalAddon()`](https://github.com/mozilla/addons-frontend/blob/cc5baba5d47e508323b3b8f45ea16b94a7f19a58/src/core/reducers/addons.js#L83). This is a good name because it illustrates the intent. 

There is an older function called [`denormalizeReview()`](https://github.com/mozilla/addons-frontend/blob/cc5baba5d47e508323b3b8f45ea16b94a7f19a58/src/amo/actions/reviews.js#L22) that did pretty much the same thing. Let's rename it to `createInternalReview()` since that makes more sense.

Ask before working on this to avoid conflicts with open PRs.

plant99:
Can I work on it? @kumar303

tofumatt:
There is a large review-code-related patch in review right now (#3220) so now is not a good time to do fix this issue as it will create confllicts.

vcarciu:
Reviews looks good. Marking issue as verified.

PR #4459, opened 2018-02-26
Make all visual representations of the user's personal ratings yellow

### LINKED ISSUE #4355: Make all visual representations of the user's personal ratings yellow
This is a follow-up to PR #4337. There is a question about whether all visual representations of the user's personal ratings should be yellow, or if we should only show yellow stars where a user can interact with them to set their rating.

For example, should the stars be yellow in the attached screenshot, @pwalm? 

image

bobsilverberg:
Fixes #4355 

This adds a new styleName to the Rating component to support a different style of stars for a rating submitted by the current user.

Note that as per #4355, the only place this change needed to be made was on the list of reviews where a user's own review should appear with yellow starts, and all other reviews should appear with grey stars.

Before:

screenshot 2018-02-26 13 04 57

After:

screenshot 2018-02-26 12 58 52

### DIFF
--- a/src/amo/components/AddonMeta/index.js
+++ b/src/amo/components/AddonMeta/index.js
@@ -67,7 +67,7 @@ export class AddonMetaBase extends React.Component<Props> {
                   className="AddonMeta-item-header"
                   rating={averageRating}
                   readOnly
-                  styleName="small"
+                  styleSize="small"
                 />
               ),
               title: i18n.gettext('Overall Rating'),
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -233,7 +233,7 @@ export class AddonReviewListItemBase extends React.Component<Props> {
           {review && !isReply ?
             <Rating
               styleSize="small"
-              isOwner={siteUser && review && review.userId === siteUser.id}
+              review={review}
               rating={review.rating}
               readOnly
             />
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -21,7 +21,7 @@ import {
 } from 'amo/actions/reviews';
 import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
-import Rating from 'ui/components/Rating';
+import UserRating from 'ui/components/UserRating';
 import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import type { UserReviewType } from 'amo/actions/reviews';
 import type { ReviewState } from 'amo/reducers/reviews';
@@ -231,10 +231,9 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating
+            <UserRating
               styleSize="small"
               review={review}
-              rating={review.rating}
               readOnly
             />
             : null
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -231,7 +231,14 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating styleName="small" rating={review.rating} readOnly />
+            <Rating
+              styleName={
+                siteUser && review && review.userId === siteUser.id ?
+                  'small-by-user' : 'small'
+              }
+              rating={review.rating}
+              readOnly
+            />
             : null
           }
           {byline}
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -232,10 +232,8 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
             <Rating
-              styleName={
-                siteUser && review && review.userId === siteUser.id ?
-                  'small-by-user' : 'small'
-              }
+              styleSize="small"
+              isOwner={siteUser && review && review.userId === siteUser.id}
               rating={review.rating}
               readOnly
             />
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -1,34 +1,33 @@
 import makeClassName from 'classnames';
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
 import { compose } from 'redux';
 
-import { getCurrentUser } from 'amo/reducers/users';
 import log from 'core/logger';
 import translate from 'core/i18n/translate';
 
 import './styles.scss';
 
 
-const RATING_STYLE_SIZES = ['small', 'large'];
+export const RATING_STYLE_SIZE_TYPES = { small: '', large: '' };
+const RATING_STYLE_SIZES = Object.keys(RATING_STYLE_SIZE_TYPES);
 
 export class RatingBase extends React.Component {
   static propTypes = {
     className: PropTypes.string,
     i18n: PropTypes.object.isRequired,
+    isOwner: PropTypes.bool,
     onSelectRating: PropTypes.func,
     rating: PropTypes.number,
     readOnly: PropTypes.bool,
     styleSize: PropTypes.oneOf(RATING_STYLE_SIZES),
-    review: PropTypes.object,
-    isOwner: PropTypes.bool,
   }
 
   static defaultProps = {
     className: '',
     readOnly: false,
     styleSize: 'large',
+    isOwner: false,
   }
 
   constructor(props) {
@@ -122,16 +121,6 @@ export class RatingBase extends React.Component {
   }
 }
 
-export function mapStateToProps(state, ownProps) {
-  const { review } = ownProps;
-  const siteUser = getCurrentUser(state.users);
-  const isOwner = siteUser && review && review.userId === siteUser.id;
-  return { isOwner };
-}
-
-const Rating = compose(
-  connect(mapStateToProps),
+export default compose(
   translate({ withRef: true }),
 )(RatingBase);
-
-export default Rating;
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -9,7 +9,7 @@ import translate from 'core/i18n/translate';
 import './styles.scss';
 
 
-const RATING_STYLES = ['small', 'large', 'small-monochrome'];
+const RATING_STYLES = ['small', 'large', 'small-monochrome', 'small-by-user'];
 
 export class RatingBase extends React.Component {
   static propTypes = {
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -9,7 +9,7 @@ import translate from 'core/i18n/translate';
 import './styles.scss';
 
 
-const RATING_STYLES = ['small', 'large', 'small-monochrome', 'small-by-user'];
+const RATING_STYLE_SIZES = ['small', 'large'];
 
 export class RatingBase extends React.Component {
   static propTypes = {
@@ -18,13 +18,15 @@ export class RatingBase extends React.Component {
     onSelectRating: PropTypes.func,
     rating: PropTypes.number,
     readOnly: PropTypes.bool,
-    styleName: PropTypes.oneOf(RATING_STYLES),
+    styleSize: PropTypes.oneOf(RATING_STYLE_SIZES),
+    isOwner: PropTypes.bool,
   }
 
   static defaultProps = {
     className: '',
     readOnly: false,
-    styleName: 'large',
+    styleSize: 'large',
+    isOwner: false,
   }
 
   constructor(props) {
@@ -83,11 +85,11 @@ export class RatingBase extends React.Component {
   }
 
   render() {
-    const { className, i18n, rating, readOnly, styleName } = this.props;
-    if (!RATING_STYLES.includes(styleName)) {
+    const { className, i18n, rating, readOnly, styleSize, isOwner } = this.props;
+    if (!RATING_STYLE_SIZES.includes(styleSize)) {
       throw new Error(
-        `styleName=${styleName} is not a valid value; ` +
-        `possible values: ${RATING_STYLES.join(', ')}`);
+        `styleSize=${styleSize} is not a valid value; ` +
+        `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
     }
     let description;
     if (rating) {
@@ -97,8 +99,10 @@ export class RatingBase extends React.Component {
       description = i18n.gettext('No ratings');
     }
 
-    const allClassNames = makeClassName('Rating', `Rating--${styleName}`,
-      className, { 'Rating--editable': !readOnly });
+    const allClassNames = makeClassName(
+      'Rating', `Rating--${styleSize}`, className,
+      { 'Rating--editable': !readOnly }, { 'Rating--by-owner': isOwner }
+    );
 
     return (
       <div
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -1,8 +1,10 @@
 import makeClassName from 'classnames';
 import * as React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { compose } from 'redux';
 
+import { getCurrentUser } from 'amo/reducers/users';
 import log from 'core/logger';
 import translate from 'core/i18n/translate';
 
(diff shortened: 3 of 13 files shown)


src/amo/components/AddonReviewListItem/index.js
```diff
@@ -231,7 +231,14 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating styleName="small" rating={review.rating} readOnly />
+            <Rating
+              styleName={
+                siteUser && review && review.userId === siteUser.id ?
+                  'small-by-user' : 'small'
```
tofumatt: Nitpick but we shouldn't have any whitespace before the colon here. Not sure if our linter will catch it (but it should 😄)

tofumatt: Also, I'd think `own-review` or `my-review` would be a better class name than `small-by-user`. We should decouple ownership `className` from size `className` regardless.

src/ui/components/Rating/index.js
```diff
@@ -9,7 +9,7 @@ import translate from 'core/i18n/translate';
 import './styles.scss';
 
 
-const RATING_STYLES = ['small', 'large', 'small-monochrome'];
+const RATING_STYLES = ['small', 'large', 'small-monochrome', 'small-by-user'];
```
tofumatt: As mentioned, I think we should decouple ownership from size. I guess we sort of messed up by having the "small monochrome" one already (are we even using it anywhere?)

But if we can have them be separate I think that'd be better.

src/ui/components/Rating/styles.scss
```diff
@@ -75,3 +76,13 @@
     background-image: url("~ui/components/Icon/img/star-half.svg");
   }
 }
+
+.Rating--small-by-user {
```
tofumatt: Looks like the size is controlled elsewhere? So this could omit the size as I mentioned and just control the colour.

bobsilverberg:
Thanks for the suggestions @tofumatt. I did decouple size from ownership, and I removed the monochrome styles entirely as they don't seem to be used anywhere. This resulted in quite a few changes, so I thought you might want to take another look before I merge it.

src/amo/components/AddonMeta/index.js
```diff
@@ -67,7 +67,7 @@ export class AddonMetaBase extends React.Component<Props> {
                   className="AddonMeta-item-header"
                   rating={averageRating}
                   readOnly
-                  styleName="small"
+                  styleSize="small"
```
tofumatt: Nice 👍

src/ui/components/Rating/index.js
```diff
@@ -97,8 +99,10 @@ export class RatingBase extends React.Component {
       description = i18n.gettext('No ratings');
     }
 
-    const allClassNames = makeClassName('Rating', `Rating--${styleName}`,
-      className, { 'Rating--editable': !readOnly });
+    const allClassNames = makeClassName(
+      'Rating', `Rating--${styleSize}`, className,
+      { 'Rating--editable': !readOnly }, { 'Rating--by-owner': isOwner }
```
tofumatt: You can actually just use one object here:

```js
const allClassNames = makeClassName(
   'Rating', `Rating--${styleSize}`, className, {
    'Rating--editable': !readOnly,
    'Rating--by-owner': isOwner,
  },
);
```

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -197,6 +198,14 @@ describe(__filename, () => {
     expect(root.find('.AddonReviewListItem-edit')).toHaveLength(0);
   });
 
+  it('assigns isOwner property if you wrote the review', () => {
+    const review = signInAndDispatchSavedReview();
```
tofumatt: I think it'd be better to be explicit about the IDs used here, as is done elsewhere in the files. This works but is not as obvious to read. You can change it to:

```js
const review = signInAndDispatchSavedReview({
  siteUserId: 123, reviewUserId: 123,
});
```

and it's a lot easier to read.

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -197,6 +198,14 @@ describe(__filename, () => {
...
+    const root = render({ review });
+    const rating = root.find(Rating);
+
+    expect(rating).toHaveProp('isOwner', true);
```
tofumatt: I would check for the actual output `className` as well, because we want that rendered when this is true.

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -197,6 +198,14 @@ describe(__filename, () => {
     expect(root.find('.AddonReviewListItem-edit')).toHaveLength(0);
   });
 
+  it('assigns isOwner property if you wrote the review', () => {
```
tofumatt: Might be useful to add an extra test ensuring the className is NOT output when the user isn't the owner of a review as well.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -231,7 +231,12 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating styleName="small" rating={review.rating} readOnly />
+            <Rating
+              styleSize="small"
+              isOwner={siteUser && review && review.userId === siteUser.id}
```
kumar303: The logic of figuring out who owns the review should reside within `Rating` so that the component can be re-used more easily. Consider this: if you want to render `Rating` within another component using this `isOwner` approach, you'd have to copy/paste all the logic and tests for checking who owns the review.

Also, there is an easier way to do this -- a pattern that you might not have seen yet. The `Rating` component can map directly to the logged in user data. In fact, [`AddonReviewListItem`](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonReviewListItem/index.js#L240) already does exactly what you need to do: it checks the `siteUser` against `review.user.Id`. Check out [its mapper](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonReviewListItem/index.js#L316) to see how it works. With that mapper pattern, the caller won't have to pass in anything special to `Rating`.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -231,7 +231,12 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating styleName="small" rating={review.rating} readOnly />
+            <Rating
+              styleSize="small"
+              review={review}
+              rating={review.rating}
```
kumar303: Now that the component accepts a `review` object, it no longer needs to accept `rating`.

src/ui/components/Rating/index.js
```diff
@@ -115,6 +122,22 @@ export class RatingBase extends React.Component {
...
+}
+
+const Rating = compose(
+  connect(mapStateToProps),
   translate({ withRef: true }),
 )(RatingBase);
+
+export default Rating;
+
+// export default compose(
```
kumar303: This can be deleted

src/ui/components/Rating/styles.scss
```diff
@@ -55,23 +55,19 @@
 }
 
 .Rating--small .Rating-choice,
-.Rating--small-monochrome .Rating-choice {
+.Rating--small-by-user .Rating-choice {
```
kumar303: I don't see `Rating--small-by-user` used anywhere -- is that still necessary? If so, it should be `...by-owner` for consistency.

tests/unit/ui/components/TestRating.js
```diff
@@ -27,6 +40,19 @@ function makeFakeEvent() {
   };
 }
 
+function signInAndDispatchSavedReview({ siteUserId, reviewUserId }) {
```
bobsilverberg: I borrowed this code from [`TestAddonReviewListItem`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestAddonReviewListItem.js#L101) , but I shortened it (and inlined some other functions it calls) for use in this file. I considered moving the functions that support this into a helper for use by both of these tests, but this seemed like a reasonable approach as well.

bobsilverberg:
The changes I made are not insignificant, and I had to make some changes to some tests that I didn't fully understand, so I wouldn't mind discussing those at some point. Flagging @kumar303 for another look.

tests/unit/ui/components/TestRating.js
```diff
@@ -27,6 +40,19 @@ function makeFakeEvent() {
...
+  dispatchSignInActions({ store, userId: siteUserId });
+  const review = {
+    ...fakeReview,
+    user: {
+      ...fakeReview.user,
+      id: reviewUserId,
+    },
+  };
+  store.dispatch(setReview(review));
+  return store.getState().reviews.byId[review.id];
```
kumar303: This works but it's unnecessary since the component takes a review object directly. It would be better like this:

```js
import { denormalizeReview } from 'amo/actions/reviews';

const review = {
  ...fakeReview,
  user: { ... },
};
return denormalizeReview(review);
```

A more common pattern you'll see for this type of test setup would look like `createInternalAddon(fakeAddon)`. The pattern is to use a common helper to simulate what the reducer would be doing without executing reducer code.

You'll have to forgive the name, I'd like to [change it](https://github.com/mozilla/addons-frontend/issues/3219) eventually.

tests/unit/ui/components/TestRating.js
```diff
@@ -40,13 +66,29 @@ describe('ui/components/Rating', () => {
...
-      .toThrowError(/styleName=x-large is not a valid value; possible values: small,/);
+  it('is classified as owned if you wrote the review', () => {
+    const review = signInAndDispatchSavedReview({
+      siteUserId: 123, reviewUserId: 123,
+    });
+    const root = render({ review });
+    expect(root.element.className).toContain('Rating--by-owner');
+  });
+
+  it('is not classified as owned if you did not write the review', () => {
```
kumar303: This needs one more test where no user is signed into the site. You can either do it by calling [`dispatchClientMetadata()`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/helpers.js#L203) (which sets up an initial logged out state) or dispatching [`logOutUser()`](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/users.js#L156) (sometimes that's easier).

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -231,7 +231,12 @@ export class AddonReviewListItemBase extends React.Component<Props> {
         {reviewBody}
         <div className="AddonReviewListItem-byline">
           {review && !isReply ?
-            <Rating styleName="small" rating={review.rating} readOnly />
+            <Rating
+              styleSize="small"
+              review={review}
```
kumar303: It would be nice to add some test coverage to make sure that `Rating` gets passed the `review` property since that's an important part of the component interface and we don't have Flow. The test suite for this is old but you can maybe edit [a test like this](https://github.com/bobsilverberg/addons-frontend/blob/eae1d1ca2d834ebad45ff3d7f9dcacc157bf45d2/tests/unit/amo/components/TestAddonReview.js#L110).

src/ui/components/Rating/index.js
```diff
@@ -18,13 +20,15 @@ export class RatingBase extends React.Component {
     onSelectRating: PropTypes.func,
     rating: PropTypes.number,
     readOnly: PropTypes.bool,
-    styleName: PropTypes.oneOf(RATING_STYLES),
+    styleSize: PropTypes.oneOf(RATING_STYLE_SIZES),
+    review: PropTypes.object,
+    isOwner: PropTypes.bool,
```
kumar303: We generally try to alphabetize component props but there's no rule for it.

src/ui/components/UserRating/index.js
```diff
@@ -0,0 +1,55 @@
...
+const mapStateToProps = (
+  state: Object, ownProps: Props
+): $Shape<Props> => {
+  const { review } = ownProps;
+  const siteUser = getCurrentUser(state.users);
+  return { isOwner: !!(siteUser && review && review.userId === siteUser.id) };
+};
+
+export default compose(
+  translate({ withRef: true }),
```
kumar303: `withRef` is left over from historic lore. We haven't fixed all the old code but this can just be `translate()`.

tests/unit/ui/components/TestRating.js
```diff
@@ -40,13 +40,18 @@ describe('ui/components/Rating', () => {
...
+    const root = render({ styleSize: 'small' });
     expect(root.element.className).toContain('Rating--small');
   });
 
-  it('throws an error for invalid styleNames', () => {
-    expect(() => render({ styleName: 'x-large' }))
-      .toThrowError(/styleName=x-large is not a valid value; possible values: small,/);
+  it('classifies as unowned by default', () => {
+    const root = render();
+    expect(root.element.className).not.toContain('Rating--by-owner');
```
kumar303: Can you also add a test for the presence of `Rating--by-owner`? Or did I miss it?

tests/unit/ui/components/TestUserRating.js
```diff
@@ -0,0 +1,82 @@
...
+
+describe('ui/components/UserRating', () => {
+  it('renders a Rating', () => {
+    const props = {
+      review: fakeReview,
+      className: 'my-class',
+      readOnly: true,
+      styleSize: 'small',
+    };
+    const root = render(props);
```
kumar303: I'm not really sure how helpful this test is. Maybe it would be more helpful like this?

```diff
diff --git a/tests/unit/ui/components/TestUserRating.js b/tests/unit/ui/components/TestUserRating.js
index 9bea9a1e8..ddfd2aa1a 100644
--- a/tests/unit/ui/components/TestUserRating.js
+++ b/tests/unit/ui/components/TestUserRating.js
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { denormalizeReview } from 'amo/actions/reviews';
+import Rating from 'ui/components/Rating';
 import UserRating, { UserRatingBase } from 'ui/components/UserRating';
 import { logOutUser } from 'amo/reducers/users';
 import {
@@ -47,7 +48,7 @@ describe('ui/components/UserRating', () => {
       readOnly: true,
       styleSize: 'small',
     };
-    const root = render(props);
+    const root = render(props).find(Rating);
     expect(root).toHaveProp('className', props.className);
     expect(root).toHaveProp('readOnly', props.readOnly);
     expect(root).toHaveProp('styleSize', props.styleSize);
```

tests/unit/ui/components/TestUserRating.js
```diff
@@ -0,0 +1,82 @@
...
+    expect(root).toHaveProp('className', props.className);
+    expect(root).toHaveProp('readOnly', props.readOnly);
+    expect(root).toHaveProp('styleSize', props.styleSize);
+  });
+
+  it('passes the rating from the review to Rating', () => {
+    const root = render({ review: fakeReview });
```
kumar303: This code just happens to work since `review.rating` is not converted but to avoid confusion in the future, this should mimic how the real app works:

```js
const root = render({ review: dernormalizeReview(fakeReview) });
```

tests/unit/ui/components/TestUserRating.js
```diff
@@ -0,0 +1,82 @@
...
+  it('passes isOwned: false to Rating if you did not write the review', () => {
+    const review = signInAndReturnReview({
+      siteUserId: 123, reviewUserId: 456,
+    });
+    const root = render({ review });
+    expect(root).toHaveProp('isOwner', false);
+  });
+
+  it('passes isOwned: false to Rating if no user is logged in', () => {
+    store.dispatch(logOutUser());
```
kumar303: It would be helpful to write this test in a way that fails if you comment out `dispatch(logOutUser())`. I think this would do it:

```js
const ownerId = 123;
const review = signInAndReturnReview({
  siteUserId: ownerId, reviewUserId: ownerId,
});
store.dispatch(logOutUser());
const root = render({ review });
expect(root).toHaveProp('isOwner', false);
```

I would recommend trying to do test-first development if you can. As an alternative, try to make the tests fail by temporarily commenting out code.

tests/unit/ui/components/TestUserRating.js
```diff
@@ -0,0 +1,82 @@
...
+  return denormalizeReview({
+    ...fakeReview,
+    user: {
+      ...fakeReview.user,
+      id: reviewUserId,
+    },
+  });
+}
+
+describe('ui/components/UserRating', () => {
```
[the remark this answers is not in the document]
bobsilverberg: Ah, I thought `describe(__filename, ...)` was the old way of doing it. I knew there was an old and a new way, but I got it backwards. Thanks!

tests/unit/ui/components/TestUserRating.js
bobsilverberg: Thanks, good catch and good advice. I could (and should) have caught that myself.

bobsilverberg:
Thanks for the detailed review, @kumar303. I have addressed all of your comments and this is ready for another look.

bobsilverberg:
Thanks @kumar303!  I hope that's not a prediction for how these changes will affect the site! ;-)

kumar303:
Well, you never know! Cautious optimism

PR #4462, opened 2018-02-26
Display permissions on add-on detail page

### LINKED ISSUE #2757: Display permissions on add-on detail page

bobsilverberg:
Fixes #2757

Now ready for review!


src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+              description={permissionStrings[permission]}
+              key={permission}
+            />
+          );
+        }
+      }
+      return permissionsToDisplay;
+    }
+
+    // Get the permission strings from the appropriate file object.
```
bobsilverberg: I originally had this as a separate function, hoping that I could unit test it (as discussed above), but then I just inlined it. It's definitely missing test coverage currently and the best way to address that might be to extract it out into a helper as I discussed above with `classifyPermission`.

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+      tooManySites: (count) => {
+        return i18n.sprintf(
+          i18n.gettext('Access your data on %(count)s other sites'),
+          { count: i18n.formatNumber(count) },
+        );
+      },
+    };
+
+    // Classify a permission as a host permission or a regular permission.
+    function classifyPermission(permission) {
```
bobsilverberg: I'd like to be able to write unit tests for this, but wasn't sure how to access it directly from a test as it's inside a component. I didn't want to export it as it's not used anywhere else. Perhaps it would be better to move this to a helper module where it could be more easily tested?

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+      const permissions = { hosts: [], permissions: [] };
+
+      // First, categorize them into host permissions and regular permissions.
+      // eslint-disable-next-line no-restricted-syntax
+      for (const permission of addonPermissions) {
+        const { type, value } = classifyPermission(permission);
+        permissions[type].push(value);
+      }
+
+      // Classify the host permissions.
```
bobsilverberg: This code could also be extracted out into a function in a helper to make it more testable. As it is I think I've covered most, if not all of it, in the tests for `PermissionsCard`, so maybe it's not that important.

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+      >
+        <div className="PermissionsCard-subhead">{i18n.gettext('This add-on can:')}</div>
+        <ul className="PermissionsCard-list">
+          {content}
+        </ul>
+        <Button
+          buttonType="neutral"
+          className="PermissionCard-learn-more"
+          href="https://support.mozilla.org/en-US/kb/permission-request-messages-firefox-extensions"
+          external
```
bobsilverberg: I assume we want this to open in a new tab/window, but there doesn't seem to be anything currently in `Button` or `Link` to support that. I suppose I could use `customProps` in `Link`, but I expected there might be a property we could use to indicate we wanted to create a link that opens in a new tab/window.

src/ui/components/Icon/img/external-black.svg
```diff
@@ -0,0 +1,14 @@
+
```
bobsilverberg: I had to create this because the icon we currently have for this, called `external`, is blue, and we want this to be black.

src/ui/components/Icon/styles.scss
```diff
@@ -174,3 +178,90 @@
...
+
+.Icon-permission-unlimitedStorage {
+  @include permission-icon($name: "storage");
+}
+
+.Icon-permission-webNavigation {
+  @include permission-icon($name: "navigation");
+}
+
+.PermissionCard-learn-more.Button--neutral:link.Button--puffy {
```
bobsilverberg: I realize that this is the wrong place for this rule, and it should really be in the stylesheet for `PermissionsCard`, but I had trouble making that work. I wanted to re-use the `icon @mixin` from this file, but couldn't seem to get that to work. Any advice on how to make this work in the right place would be appreciated.

tests/unit/ui/components/TestPermission.js
```diff
@@ -0,0 +1,45 @@
...
+
+import Icon from 'ui/components/Icon';
+import Permission from 'ui/components/Permission';
+
+
+describe('Permission component', () => {
+  const defaultProps = {
+    className: 'my-class',
+    description: 'A decription',
+    key: 'someKey',
```
bobsilverberg: Oops. I need to remove this. I've already added a bunch of comments to this PR, and if I push a new commit I think they'll get hidden, so just assume I'll fix this for now. :P

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+
+export class PermissionsCardBase extends React.Component<Props> {
+  render() {
+    const { addon, i18n, userAgentInfo } = this.props;
+
+    if (!addon) {
+      return null;
+    }
+
+    // These should be kept in sync with Firefox's strings for webextention permissions
```
bobsilverberg: I discussed this approach at length with @tofumatt and @diox via IRC and we talked about putting something automated in place to pull the strings, in all languages, from Firefox. It was decided that that would be too much work to be worthwhile, and just having our translators translate them, as we do with other strings that also exist in Firefox, would be the most practical approach.

I do have a couple of follow-up issues/ideas though:

1. Even if we won't be automating the process to continually keep our strings in sync with Firefox, perhaps we could run a one-off job to populate all the locales with strings that already exist in Firefox, to save our translators the job of having to either translate or copy all of them.

2. With this approach there is nothing in place to inform us when a new permission and string are added to Firefox, and the way it's coded, a new permission that doesn't have a matching string in this map will simply be ignored, so it seems like it will be very easy to miss one when it is added. What do we think we can do about that? What would be the best way to try to ensure we update this map when new strings are added to Firefox?

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+      allUrls: i18n.gettext('Access your data for all websites'),
+      wildcard: (domain) => {
+        return i18n.sprintf(
+          i18n.gettext('Access your data for sites in the %(domain)s domain'),
+          { domain }
+        );
+      },
+      tooManyWildcards: (count) => {
+        return i18n.sprintf(
+          i18n.gettext('Access your data in %(count)s other domains'),
```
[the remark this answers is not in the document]
bobsilverberg: The logic dictates that `count` will never be less than 2, which is why I didn't bother with pluralization. I suppose we could choose to change that logic down the road, in which case we'd be happy that we dealt with pluralization now, so that's an argument in favour of it, but I don't think we strictly need it right now.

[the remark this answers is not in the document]
kumar303: Ah, ok, then this should have a comment up above it so that some do-gooder developer doesn't change it one day thinking they caught a bug.

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+  padding: 0;
+
+  li {
+    display: block;
+    margin-left: 10px;
+    margin-top: 10px;
+    vertical-align: middle;
+
+    .Icon {
+      margin-right: 10px;
```
kumar303: The pixel distance I see in Sketch is `16px`. Also, generally, we go with multiples of 4 but this isn't a hard and fast rule. There are a lot of `10px` measurements on the site left over from early days. This kind of distance is more like `8px` or `12px` in the latest mocks.

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
+.PermissionCard-learn-more {
+  margin: 12px auto 0;
+  width: 100%;
+
+  .Icon {
+    margin-left: 10px;
```
kumar303: I see `11px` in the mocks but since we try to adhere to multiples of 4 (the mocks aren't always exact), I'd suggest `12px`.

kumar303: To handle right-to-left locales, which will reverse the icon/text order, you'll need to use [our mixin](https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/mixins.scss#L144) like: 

```css

.Icon {
  @include margin-start(12px);
}
```

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+.PermissionsCard-list {
+  list-style: none;
+  margin: 10px 0 0;
+  padding: 0;
+
+  li {
+    display: block;
+    margin-left: 10px;
+    margin-top: 10px;
+    vertical-align: middle;
```
kumar303: This only works for table cells. It's better to use flexbox as I suggested up above.

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+    if (!content.length) {
+      return null;
+    }
+
+    return (
+      <Card
+        header={i18n.gettext('Permissions')}
+        className="PermissionsCard"
+      >
+        <div className="PermissionsCard-subhead">{i18n.gettext('This add-on can:')}</div>
```
kumar303: Since this is prose introducing the upcoming unordered list, you could use a `<p>` tag.

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+  margin: 12px auto 0;
+  width: 100%;
+
+  .Icon {
+    margin-left: 10px;
+  }
+}
+
+.PermissionsCard-subhead {
+  color: #737373;
```
kumar303: We try to always use standard Photon colors and if the mocks stray from that then you should ask Phil if it was intentional or not. The easiest way to find the color var is to [ripgrep](https://github.com/BurntSushi/ripgrep) the hex (in lower case). This is how I found it:

```
$ rg 737373 node_modules/
node_modules/photon-colors/colors.scss
56:$grey-50: #737373;
...
```

You can then add the var like:

```css
@import "~photon-colors/colors";

.PermissionsCard-subhead {
  color: $grey-50;
}
```

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.PermissionsCard-list {
+  list-style: none;
+  margin: 10px 0 0;
```
kumar303: The top margin I see in Sketch is `18px` but this can probably be set to `0` since the `li` will already have a top margin of `18px` (which I mentioned up above).

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
+.PermissionCard-learn-more {
+  margin: 12px auto 0;
```
kumar303: I see a top margin of `28px` in the mocks

src/amo/components/PermissionsCard/index.js
kumar303: > I assume we want this to open in a new tab/window

I think it's fine to rely on our current implementation of buttons with the `external` property. I think we intentionally don't open it in a new window because that's something that users can easily decide to do on their own with command+click.

src/amo/components/PermissionsCard/index.js
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    // Get the permission strings from the appropriate file object.
+    const agentOsName =
+      userAgentInfo.os.name && userAgentInfo.os.name.toLowerCase();
+    const platform = userAgentOSToPlatform[agentOsName];
+    const file = addon.platformFiles[platform] || addon.platformFiles[OS_ALL];
+
+    if (!file) {
+      log.debug(oneLine`No file exists for platform "${agentOsName}"
```
kumar303: `oneLineTrim` is a bit nicer since it gets rid of the whitespace too

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+
+.PermissionsCard-list {
+  list-style: none;
+  margin: 10px 0 0;
+  padding: 0;
+
+  li {
+    display: block;
+    margin-left: 10px;
+    margin-top: 10px;
```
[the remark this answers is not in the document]
bobsilverberg: I'm not sure exactly where you're seeing this. When I tried to do as you described it looked like `14px` to me, but maybe I'm looking in the wrong place. Oh, I was looking on the Symbols page at the example of a permissions panel, but I see there's also an example on the Screens page for Extension Detail. I'm not sure which one is more accurate.

src/amo/components/PermissionsCard/styles.scss
bobsilverberg: Hrm, again it looks different between the Screens and the Symbols pages. I guess I'll just go with 12 for now, as 16 looks like too much to me.

src/amo/components/PermissionsCard/styles.scss
bobsilverberg: Thanks for the ripgrep tip - useful!

src/amo/components/PermissionsCard/styles.scss
bobsilverberg: The 28 I see if between the top of the button and the bottom of the text, but that doesn't take the bottom margin of the `<ul>` into account, so it looks too big to me. If I make it 16 it works with the bottom margin to give us an effective 28.

src/amo/components/PermissionsCard/styles.scss
```diff
@@ -0,0 +1,29 @@
...
+    vertical-align: middle;
+
+    .Icon {
```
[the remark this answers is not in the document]
bobsilverberg: I changed this as requested, but personally I think the 24px version is too big.

tests/unit/amo/components/TestPermissionsCard.js
```diff
@@ -0,0 +1,72 @@
...
+  dispatchClientMetadata,
+  fakeAddon,
+  fakePlatformFile,
+} from 'tests/unit/amo/helpers';
+import { fakeI18n, shallowUntilTarget } from 'tests/unit/helpers';
+import Button from 'ui/components/Button';
+import Permission from 'ui/components/Permission';
+
+
+describe('PermissionsCard component', () => {
```
kumar303: top-level describe should always use `__filename`

tests/unit/amo/components/TestPermissions.js
```diff
@@ -0,0 +1,175 @@
...
+} from 'core/reducers/api';
+import {
+  fakeI18n,
+} from 'tests/unit/helpers';
+import {
+  createFakeAddon,
+  fakePlatformFile,
+} from 'tests/unit/amo/helpers';
+
+describe('Permissions module', () => {
```
kumar303: Same here: `__filename`

tests/unit/ui/components/TestPermission.js
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import Icon from 'ui/components/Icon';
+import Permission from 'ui/components/Permission';
+
+
+describe('Permission component', () => {
```
kumar303: This should also use `__filename`

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,237 @@
...
+
+import log from 'core/logger';
+import { OS_ALL } from 'core/constants';
+import { userAgentOSToPlatform } from 'core/installAddon';
+import type { AddonType } from 'core/types/addons';
+import type { UserAgentInfoType } from 'core/reducers/api';
+import type { I18nType } from 'core/types/i18n';
+import Permission from 'ui/components/Permission';
+
+import './styles.scss';
```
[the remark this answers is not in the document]
kumar303: Importing `styles.scss` is kind of magical. All it does is add the style contents to the bundle so you only need to do it once.

src/ui/components/Icon/styles.scss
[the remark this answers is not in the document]
bobsilverberg: I tried this, although adding a new toggle for `externalBlack` seemed less than ideal because one ends up with two toggles that do the same thing (create an `Icon`), which requires extra code to untangle. Instead I added a property for `blackIcon` which simply controls the `name` used for the `Icon`. Note that this still didn't entirely deal with the specificity problem I was having with the css. I still needed to define a style that includes `.Button--neutral:link.Button--puffy` to make it work. :(

bobsilverberg:
Please note that the current eslint errors are expected, as this is assuming it will land on top of PR #4498. Please ignore them for now.

src/amo/components/Link.js
```diff
@@ -10,6 +10,7 @@ import Icon from 'ui/components/Icon';
 
 export class LinkBase extends React.Component {
   static propTypes = {
+    blackIcon: PropTypes.bool,
```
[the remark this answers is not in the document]
bobsilverberg: I was trying to avoid a situation where a caller could set both `external` and `externalDark` and end up with two icons, which is why I split it up into two props that do different things, as opposed to two props which each mean "it's an external link". I've changed it to use `externalDark` instead, and added some logic to avoid two icons.

src/ui/components/Icon/styles.scss
```diff
@@ -65,6 +69,15 @@
   width: 10px;
 }
 
+.Button--neutral:link.Button--puffy {
```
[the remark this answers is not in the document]
bobsilverberg: It does not. :( That was the first thing I tried. I also tried a style for `.Icon.Icon-external-dark` and that didn't work either. In both cases the sizes are overridden by a rule for `.Button--neutral.Button--puffy .Icon, .Button--neutral:link.Button--puffy .Icon`, which is why I ended up with the rule I have. 

I updated my rule to include non-puffy buttons, but without the specific rule for `.Button--neutral:link.Button--puffy` I cannot seem to get it to work.

tests/unit/amo/components/TestPermissions.js
```diff
@@ -0,0 +1,179 @@
...
+
+  describe('getCurrentPermissions', () => {
+    const createAddon = (platforms = [OS_MAC, OS_ALL]) => {
+      const files = [];
+      for (const platform of platforms) {
+        files.push(
+          {
+            ...fakePlatformFile,
+            platform,
+            permissions: [platform],
```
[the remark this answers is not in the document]
bobsilverberg: I did it to make the code more concise, but I see what you mean. The function being tested doesn't really care that the array of strings contains permissions - as far as it's concerned it's just an array of strings. So the fact that we're using an invalid permission string doesn't really matter to the test, but I agree it is a bit confusing. To make the test work I need each platform to return something different in the array of strings so that I can verify I'm getting the correct array back. The simplest thing to do, without adding extra code, was to just use the platform name as the permission string. That way I can create it in the loop, and I know what to expect in the test. 

I have added some extra code so it can use valid permission strings, and I can see how it makes the test less confusing. :)

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  // Get a list of permissions from the correct platform file.
+  getCurrentPermissions(
+    addon: AddonType,
+    userAgentInfo: UserAgentInfoType
+  ): Array<string> {
+    const agentOsName =
+      userAgentInfo.os.name && userAgentInfo.os.name.toLowerCase();
+    const platform = userAgentOSToPlatform[agentOsName];
```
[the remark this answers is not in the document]
bobsilverberg: I added a warning, but for some reason it's made Flow upset. I'm seeing this:

```
change:src/amo/components/PermissionsCard/permissions.js
Error: src/amo/components/PermissionsCard/permissions.js:90
 90:       log.debug(`No platform exists for user agent OS "${agentOsName}"`);
                                                              ^^^^^^^^^^^ undefined. This type cannot be coerced to
 90:       log.debug(`No platform exists for user agent OS "${agentOsName}"`);
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ string
```

which I find confusing for two reasons:

1. It's actually entirely valid to pass `undefined` into a template string, although it will result in the string `undefined` instead of an empty string, which I think is fine in this case.
2. I was surprised that Flow is complaining about this now, when I already had another log statement which uses `agentOsName` in a template string and it didn't complain about that. I see now that the older statement uses `oneLineTrim`, and if I wrap the new log statement in that then Flow ceases complaining, but I don't need `oneLineTrim` for this one, and I don't really understand why that makes a difference.

bobsilverberg:
Regarding `formatPermissions`, I agree it's a bit confusing, but it's doing a lot of work that requires multiple passes. It's code I borrowed from Firefox, so I believe it's a pretty efficient approach, and I broke that code down into a few separate functions in `permissions.js` already. If you have suggestions to make it simpler I'd love to hear them.

src/amo/components/PermissionsCard/permissions.js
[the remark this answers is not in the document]
kumar303: Relying on the string `undefined` is relying on JS quirks which is why Flow forbids it. You could do something like `${agentOsName || '<undefined agentOsName>'}` which would also provide a debugging hint.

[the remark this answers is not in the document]
bobsilverberg: Well, I wasn't going to rely on the string being `undefined`, it's just that the message would be fine with either `underfined` or an empty string, but I guess Flow doesn't know that.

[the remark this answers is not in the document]
bobsilverberg: Ironically, adding that code makes the message so long that I now _do_ need to use `oneLineTrim`. 😉

tests/unit/amo/components/TestPermissions.js
```diff
@@ -0,0 +1,179 @@
...
+
+describe(__filename, () => {
+  let permissionUtils;
+
+  beforeEach(() => {
+    permissionUtils = new PermissionUtils(fakeI18n());
+  });
+
+  describe('getCurrentPermissions', () => {
+    const permissionsForOSs = {
```
kumar303: Thanks. This clarified the intent of the test for me.

tests/unit/amo/components/TestPermissions.js
```diff
@@ -0,0 +1,179 @@
+import UAParser from 'ua-parser-js';
+
+import { PermissionUtils } from 'amo/components/PermissionsCard/permissions';
```
kumar303: I think `TestPermissionUtils.js` would be a nicer name for this test file.

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,237 @@
...
+              className="hostPermission"
+              description={this.permissionStrings[moreKey](remaining)}
+              key={moreKey}
+            />
+          );
+        }
+      };
+
+      formatHostPermissions(wildcards, 'wildcard', 'tooManyWildcards');
+      formatHostPermissions(sites, 'oneSite', 'tooManySites');
```
kumar303: I see "Access your data in 3 other domains" and "Access your data on 2 other sites" for the [same extension](http://localhost:3000/en-US/firefox/addon/all-of-the-permissions/). This seems redundant. What about combining this message into one? It could be filed a follow-up issue if it's too much work.


screenshot 2018-03-08 12 15 26

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,237 @@
...
+  // Get a list of permissions from the correct platform file.
+  getCurrentPermissions(
+    addon: AddonType,
+    userAgentInfo: UserAgentInfoType
+  ): Array<string> {
+    const agentOsName =
+      userAgentInfo.os.name && userAgentInfo.os.name.toLowerCase();
+    const platform = userAgentOSToPlatform[agentOsName];
+
+    if (!platform) {
```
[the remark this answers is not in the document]
bobsilverberg: True, but I think I need to bring back `fakeUserAgent` from https://github.com/mozilla/addons-frontend/pull/4462/commits/07f444f81843b6d6bf5f63205490230180047887#diff-76f7a688842283b52b432e415dcf2e3eR42, or for me to just create a one time object to pass into `getCurrentPermissions` because I don't think we can get an invalid platform from `UAParser`, correct? I guess if I only need it for the one test I can just create an object for that purpose.

src/amo/components/PermissionsCard/permissions.js
bobsilverberg: This is what Firefox does (I'm pretty sure) and I tried to match what Firefox does as much as possible. It does make sense that we don't really need both of those messages, although there is a difference between `sites` and `domains`. It's not possible for the user to differentiate when they both just say `on x other ...` though, granted.

@devaneymoz Do you have an opinion on this?

bobsilverberg: FTR, I confirmed that the above is what Firefox shows. It does not combine "3 other domains" and "2 other sites" into "5 other x".

kumar303: ok, I see the same thing in Firefox. I don't really see how it's helpful but I suppose it's fine to match the Firefox UI, at least for a first pass.

screenshot 2018-03-08 14 09 18

src/amo/components/PermissionsCard/permissions.js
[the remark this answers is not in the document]
kumar303: I agree that testing the log message is unhelpful. It's good enough to execute the statements within the if branch -- that will make sure no errors are thrown.

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,238 @@
...
+          className="hostPermission"
+          description={this.permissionStrings.allUrls}
+          key="allUrls"
+        />
+      );
+    } else {
+      // Formats a list of host permissions.  If we have 4 or fewer, display
+      // them all, otherwise display the first 3 followed by an item that
+      // says "...plus N others".
+      const formatHostPermissions = (list, itemKey, moreKey) => {
```
kumar303: I don't want to hold up this patch so here's a comment for future reference -- it's not a change request.

React is really good at encapsulating formatting code and encouraging you to use composition to put all formatters together. The short lived inline functions here suggest to me that the code should be split into more components. 

In this case, the logic for rendering host/site permissions is quite complex and is also used in more than one place. It might work better as a separate `<HostPermissions />` component that would be responsible for handling `<all_urls>` and wildcards, rendering the right number of `<Permission>` components, and adding the "...and 3 more..." permission. Ultimately, it would need to return a list of `<li>` tags without a parent element which would require a [fragment](https://reactjs.org/docs/fragments.html).

Again, I don't want to hold up this patch but it could be something to work on in a low priority follow-up issue.

src/amo/components/PermissionsCard/permissions.js
```diff
@@ -0,0 +1,238 @@
...
+  getCurrentPermissions(
+    addon: AddonType,
+    userAgentInfo: UserAgentInfoType
+  ): Array<string> {
+    const agentOsName =
+      userAgentInfo.os.name && userAgentInfo.os.name.toLowerCase();
+    const platform = userAgentOSToPlatform[agentOsName];
+
+    if (!platform) {
+      log.error(oneLine`No platform exists for user agent OS
```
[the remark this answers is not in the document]
bobsilverberg: Hmm, I thought we did too, because my editor (which until now wasn't set up to strip trailing spaces) always bugs me about them. I think eslint ignored this one (both in the linter and in my editor) because it's in a template string. I have now updated my editor to automatically strip extra spaces.

src/amo/components/PermissionsCard/permissions.js
bobsilverberg: This is an interesting idea, and one I'd like to pursue, but yes, doing so in a follow-up seems to make sense so we can land this feature sooner rather than later. I opened #4527 to track it.

bobsilverberg:
I'm not sure that you were the one struggling. 😉 Your advice was sound and very useful. Thanks for being persistent. 😃

Issue #4372, opened 2018-02-13
Pass specific src params to details page links


devaneymoz:
It appears the new AMO front-end isn’t passing tracking parameters as it did in the old site. 

In the old site, for example, when you’d click to an individual add-on’s detail page from within the “Featured Extensions” promo module, its passing URL included “src=hp-dl-featured”. With the new front-end, there is no URL change in the path. 

Perhaps the tracking params are hidden? Or are we not tracking discovery flow as we did with the old AMO? 

Perhaps there are more (ideas welcome!), but here are the critical source values we should track: 

Collections (src=collection)

Homepage hero area (src=hp-dl-promo)

Featured (src=featured)

Search (src=search)

Filtered groups
- Popular (src=mostpopular)
- Top-rated (src=rating)
- Recently added (src=created)
- Trending (src=hotness)

jvillalobos:
Flagging this as P2, since it has a negative effect on statistics tracking for developers and us.

diox:
Looks like a dupe of https://github.com/mozilla/addons-frontend/issues/4270 which is already a p2.

jvillalobos:
Closing as a dupe.

muffinresearch:
I don't think this is a dupe. #4270 is about allowing the src parameter in the url to be passed down to the install link, however this is about providing the src parameters to the details page links. I don't see them present if you go to the homepage for example and click through to the details page for an add-on in a collection.

muffinresearch:
As mentioned in the OP fixing this requires passing the relevant src params in the details page links.

kumar303:
What source should we use when linking to add-ons from the new *Re-imagine search* card on the home page?

<img width="850" alt="screenshot 2018-02-26 16 41 06" src="https://user-images.githubusercontent.com/55398/36699933-df29a804-1b13-11e8-995e-3cdc9fcbb5cf.png">

kumar303:
Here are all the home page curated shelves for which I don't see sources. Should I leave the source blank for now?

* Re-imagine search
* Privacy tools
* Dynamic media downloaders

kumar303:
One other question that came up while working on this: should we use `src=collection` for add-ons in the *home page* collections? The fix I added only adds it for when you're on any collection detail page.

AlexandraMoga:
Verified  this on stage and here are my observations:
- **Working**
 `src=collection` -  param sent when accessing an add-on details page from "Customize the way Firefox works..." collections shelf or from a user's collection
[...]

PR #4463, opened 2018-02-26
Use add-on install source links like the old site

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4372 by adding the old add-on install source links. The issue specifies some old home page shelves that we aren't using anymore so I've asked about what to do there. I will file a follow-up issue for that if necessary.

### DIFF
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -1,36 +1,46 @@
+/* @flow */
+/* eslint-disable react/sort-comp */
 import makeClassName from 'classnames';
-import PropTypes from 'prop-types';
 import * as React from 'react';
 
 import SearchResult from 'amo/components/SearchResult';
 import CardList from 'ui/components/CardList';
+import type { AddonType } from 'core/types/addons';
 
 import './styles.scss';
 
 
-export default class AddonsCard extends React.Component {
-  static propTypes = {
-    addons: PropTypes.array.isRequired,
-    children: PropTypes.node,
-    className: PropTypes.string,
-    loading: PropTypes.bool,
-    // When loading, this is the number of placeholders
-    // that will be rendered.
-    placeholderCount: PropTypes.number,
-    type: PropTypes.string,
-    showMetadata: PropTypes.bool,
-    showSummary: PropTypes.bool,
-  }
+type Props = {|
+  addonInstallSource?: string,
+  addons?: Array<AddonType> | null,
+  children?: React.Node,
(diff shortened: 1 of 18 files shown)


src/amo/components/AddonsCard/index.js
```diff
@@ -63,10 +74,13 @@ export default class AddonsCard extends React.Component {
       }
     }
 
+    const allClassNames = makeClassName('AddonsCard', className, {
+      'AddonsCard--horizontal': type === 'horizontal',
+    });
     return (
       <CardList
         {...otherProps}
-        className={makeClassName('AddonsCard', `AddonsCard--${type}`, className)}
```
kumar303: The old component was adding `AddonsCard--link` by default but I can't find any references to that class anywhere. I think it had bitrotted. The only one in use appears to be `AddonsCard--horizontal`.

src/amo/components/SearchResult/index.js
```diff
@@ -1,27 +1,32 @@
+/* @flow */
+/* eslint-disable react/sort-comp */
```
muffinresearch: Does this rule need to be added to our general lint rules, I notice it being used a couple of times?

src/amo/components/Home/index.js
```diff
@@ -220,6 +221,7 @@ export class HomeBase extends React.Component {
         />
 
         <LandingAddonsCard
+          addonInstallSource="mostpopular"
```
muffinresearch: I wonder if these strings should be constants?

src/amo/components/SearchResult/index.js
kumar303: Heh, maybe. It has bugs triggered by Flow otherwise I suppose it works. I would be fine to remove it completely. I've never found it helpful.

src/amo/components/Home/index.js
kumar303: What would that buy us? I could do it but I'm not sure how helpful it would be.

muffinresearch: Aren't they used in multiple places?

kumar303: No, each one is only configured in a single place. I suppose in the future we might put some in multiple places.

muffinresearch: If it's not used in multiple places it's less useful, but if it ever was it might be nice to validate these are one of a list of valid sources.

muffinresearch:
:shipit:

Issue #4284, opened 2018-01-31
Add new routes for user profile and edit user profile pages


tofumatt:
We'll need new routes in nginx to route user profile and edit profile requests to the front-end, but that's dependant on other code landing, so filing this for now so we don't forget 😄 

- [x] User Profile: `/user/:username/` (Depends on #4283, ready)
- [x] User Profile Edit: `/user/:username/edit`
- [x] User Profile Edit: `/users/edit`

tofumatt:
`/user/:username/` is now ready and should be enabled **on `-dev` only** for now 😄

bqbn:
@tofumatt what regex would you recommend we use to match `:username`?

For example, we use https://github.com/mozilla/addons-server/blob/master/src/olympia/addons/urls.py#L9 to match `addon_id`. 

We try as much as possible to keep consistent with the application.

tofumatt:
Whatever the old app used to match usernames would work I guess. Seems like it's here: https://github.com/mozilla/addons-server/blob/master/src/olympia/users/urls.py#L7

I think usernames are largely alphanumeric with dashes, but they can probably have other characters too. I'm not sure if the Django regex is overly generous in terms of characters but you can use it if that works.

bqbn:
`/user/:username/` is now on mobile. 😄

bqbn:
Re-routing for both `/user/:username/edit/` and `/users/edit` is on -dev env now.

willdurand:
Thank you!

AlexandraMoga:
@willdurand  I have checked that both [https://addons-dev.allizom.org/en-US/firefox/**users/edit/**](https://addons-dev.allizom.org/en-US/firefox/users/edit/) and [https://addons-dev.allizom.org/en-US/firefox/**users/edit**](https://addons-dev.allizom.org/en-US/firefox/users/edit) are redirecting to the Update account page on the new frontend. Let me know if there is anything more I could test here.

I also have one question:
I you access one of the above links while being logged out, the new frontend will show a Page not Found (404). My opinion is that the user should be prompted to the FxA log in page instead. This flow was available on the legacy pages for example. What do you think?

AlexandraMoga:
Verified as fixed on AMO-dev, with FF60, WIn10x64

Issue #2632, opened 2017-06-23
Add a helper to render captured 404 errors more easily


kumar303:
Here is a pattern that components will need to do in their `render()`:

```js
if (errorHandler.hasError() &&
    errorHandler.capturedError.responseStatusCode === 404) {
  return <NotFound />;
}
```

We should wrap this into a HOC like `with404Handling()` or maybe just bake it into `withErrorHandler()` with an option to opt out (if needed).

See [`Addon`](https://github.com/mozilla/addons-frontend/blob/a044452cae2dcd084965cd3ef99c71b9eea2fafe/src/amo/components/Addon/index.js#L251) for an example.

kumar303:
This is also related since I think it will require a similar approach where we move rendering logic into the HOC https://github.com/mozilla/addons-frontend/issues/2905

kumar303:
This would still be nice to fix. It probably shouldn't be an HOC, though.

PR #4469, opened 2018-02-28
feat: Add a User Profile page with an info card

### LINKED ISSUE #4283: Show user info in card on profile
We should show the user's name, photo, etc. in a card.

screenshot 2018-01-31 14 56 29

tofumatt:
Fix #4283

Looks like this:

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -97,6 +97,7 @@ module.exports = {
     'defaultLang',
     'enableClientConsole',
     'enableCollectionEdit',
+    'enableUserProfile',
     'fxaConfig',
     'isDeployed',
     'isDevelopment',
@@ -288,4 +289,9 @@ module.exports = {
 
   // If true, show a link to the beta versions page on the listing page.
   betaVersions: true,
+
+  // Show and link to User Profiles using the front-end. These pages
+  // are still in development so this should only be enabled locally and
+  // on `-dev`.
+  enableUserProfile: false,
 };
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -1,0 +1,172 @@
+/* @flow */
+import * as React from 'react';
+import Helmet from 'react-helmet';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import NotFound from 'amo/components/ErrorPage/NotFound';
+import ReportUserAbuse from 'amo/components/ReportUserAbuse';
+import { fetchUserAccount, getUserByUsername } from 'amo/reducers/users';
+import { withErrorHandler } from 'core/errorHandler';
+import translate from 'core/i18n/translate';
+import log from 'core/logger';
+import { removeProtocolFromURL, sanitizeUserHTML } from 'core/utils';
+import Card from 'ui/components/Card';
+import DefinitionList, { Definition } from 'ui/components/DefinitionList';
+import LoadingText from 'ui/components/LoadingText';
+import Rating from 'ui/components/Rating';
+import UserAvatar from 'ui/components/UserAvatar';
+import type { UsersStateType, UserType } from 'amo/reducers/users';
+import type { DispatchFunc } from 'core/types/redux';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+
+type Props = {|
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  params: {| username: string |},
+  user?: UserType,
+|};
+
+export class UserProfileBase extends React.Component<Props> {
+  componentWillMount() {
+    const { dispatch, errorHandler, params, user } = this.props;
+
+    if (!user) {
+      dispatch(fetchUserAccount({
+        errorHandlerId: errorHandler.id,
+        username: params.username,
+      }));
+    }
+  }
+
+  componentWillReceiveProps({ params: newParams }: Props) {
+    const { dispatch, errorHandler, params: oldParams } = this.props;
+
+    if (oldParams.username !== newParams.username) {
+      dispatch(fetchUserAccount({
+        errorHandlerId: errorHandler.id,
+        username: newParams.username,
+      }));
+    }
+  }
+
+  render() {
+    const {
+      errorHandler,
+      i18n,
+      params,
+      user,
+    } = this.props;
+
+    let errorMessage;
+    if (errorHandler.hasError()) {
+      log.warn('Captured API Error:', errorHandler.capturedError);
+
+      if (errorHandler.capturedError.responseStatusCode === 404) {
+        return <NotFound errorCode={errorHandler.capturedError.code} />;
+      }
+
+      errorMessage = errorHandler.renderError();
+    }
+
+    const userProfileHeader = (
+      <React.Fragment>
+        <UserAvatar className="UserProfile-avatar" user={user} />
+
+        <h1 className="UserProfile-name">
+          {user ? user.displayName : <LoadingText />}
+        </h1>
+      </React.Fragment>
+    );
+    const userProfileTitle = i18n.sprintf(
+      i18n.gettext('User Profile for %(user)s'), {
+        user: user ? user.displayName : params.username,
+      }
+    );
+
+    return (
+      <div className="UserProfile">
+        <Helmet>
+          <title>{userProfileTitle}</title>
+        </Helmet>
+
+        {errorMessage}
+
+        <div className="UserProfile-wrapper">
+          <Card
+            className="UserProfile-user-info"
+            header={userProfileHeader}
+          >
+            <DefinitionList className="UserProfile-dl">
+              {user && user.homepage ? (
+                <Definition
+                  className="UserProfile-homepage"
+                  term={i18n.gettext('Homepage')}
+                >
+                  <a href={user.homepage}>
+                    {removeProtocolFromURL(user.homepage)}
+                  </a>
+                </Definition>
+              ) : null}
+              <Definition
+                className="UserProfile-user-since"
+                term={i18n.gettext('User since')}
+              >
+                {user ? i18n.moment(user.created).format('ll')
+                      : <LoadingText />}
+              </Definition>
+              <Definition
+                className="UserProfile-number-of-addons"
+                term={i18n.gettext('Number of add-ons')}
+              >
+                {user ? user.num_addons_listed : <LoadingText />}
+              </Definition>
+              <Definition
+                className="UserProfile-rating-average"
+                term={i18n.gettext('Average rating of developer’s add-ons')}
+              >
+                {user ? (
+                  <Rating
+                    rating={user.average_addon_rating}
+                    readOnly
+                    styleName="small"
+                  />
+                ) : <LoadingText />}
+              </Definition>
+            </DefinitionList>
+
+            {user && user.biography && user.biography.length ? (
+              <div
+                className="UserProfile-biography"
+                // eslint-disable-next-line react/no-danger
+                dangerouslySetInnerHTML={sanitizeUserHTML(user.biography)}
+              />
+            ) : null}
+
+            <ReportUserAbuse className="UserProfile-abuse-button" user={user} />
+          </Card>
+        </div>
+      </div>
+    );
+  }
+}
+
+export function mapStateToProps(
+  state: {| users: UsersStateType |},
+  ownProps: Props,
+) {
+  return {
+    user: getUserByUsername(state.users, ownProps.params.username),
+  };
+}
+
+export default compose(
+  connect(mapStateToProps),
+  translate(),
+  withErrorHandler({ name: 'UserProfile' }),
+)(UserProfileBase);
--- a/src/amo/components/UserProfile/styles.scss
+++ b/src/amo/components/UserProfile/styles.scss
@@ -1,0 +1,48 @@
+@import "~amo/css/inc/vars";
+@import "~core/css/inc/mixins";
+@import "~ui/css/vars";
+
+$avatar-size: 64px;
+
+.UserProfile {
+  margin: $padding-page;
+}
+
+.UserProfile-wrapper {
+  display: flex;
+}
+
+.UserProfile-user-info {
+  @include respond-to(large) {
+    min-width: 300px;
+    width: 35%;
+  }
+}
+
+.UserProfile-avatar {
+  display: block;
+  height: $avatar-size;
+  margin: 0 0 12px;
+  width: $avatar-size;
+}
+
+.UserProfile-name {
+  @include font-regular();
+
+  font-size: $font-size-l;
+  margin: 0;
+
+  @include respond-to(medium) {
+    font-size: $font-size-xl;
+  }
+}
+
+.UserProfile-rating-average .Rating {
+  @include margin(0, 0, 0, -3px);
+
+  justify-content: start;
+}
+
+.UserProfile-abuse-button .ReportUserAbuse-show-more {
+  width: 100%;
+}
--- a/src/amo/routes.js
+++ b/src/amo/routes.js
@@ -2,6 +2,22 @@ import config from 'config';
 import * as React from 'react';
 import { IndexRoute, Route } from 'react-router';
 
+import About from 'amo/components/StaticPages/About';
+import Addon from 'amo/components/Addon';
+import AddonReviewList from 'amo/components/AddonReviewList';
+import App from 'amo/components/App';
+import CategoriesPage from 'amo/components/CategoriesPage';
+import Category from 'amo/components/Category';
+import Home from 'amo/components/Home';
+import LandingPage from 'amo/components/LandingPage';
+import LanguageTools from 'amo/components/LanguageTools';
+import NotAuthorized from 'amo/components/ErrorPage/NotAuthorized';
+import NotFound from 'amo/components/ErrorPage/NotFound';
+import ReviewGuide from 'amo/components/StaticPages/ReviewGuide';
+import SearchPage from 'amo/components/SearchPage';
+import ServerError from 'amo/components/ErrorPage/ServerError';
+import Collection from 'amo/components/Collection';
+import UserProfile from 'amo/components/UserProfile';
 import SimulateAsyncError from
   'core/containers/error-simulation/SimulateAsyncError';
 import SimulateClientError from
@@ -9,27 +25,11 @@ import SimulateClientError from
 import SimulateSyncError from
   'core/containers/error-simulation/SimulateSyncError';
 
-import About from './components/StaticPages/About';
-import Addon from './components/Addon';
-import AddonReviewList from './components/AddonReviewList';
-import App from './components/App';
-import CategoriesPage from './components/CategoriesPage';
-import Category from './components/Category';
-import Home from './components/Home';
-import LandingPage from './components/LandingPage';
-import LanguageTools from './components/LanguageTools';
-import NotAuthorized from './components/ErrorPage/NotAuthorized';
-import NotFound from './components/ErrorPage/NotFound';
-import ReviewGuide from './components/StaticPages/ReviewGuide';
-import SearchPage from './components/SearchPage';
-import ServerError from './components/ErrorPage/ServerError';
-import Collection from './components/Collection';
-
 // If you add a new route here, check that the nginx rules maintained by ops
 // are in sync. See:
 // https://github.com/mozilla-services/puppet-config/tree/master/amo
 export default (
-  <div>
+  <React.Fragment>
     <Route path="/:lang" component={App}>
       <Route path="about" component={About} />
       { /* TODO: Post launch update this URL and redirect see #3374/ */ }
@@ -39,6 +39,7 @@ export default (
       <IndexRoute component={Home} />
       <Route path="addon/:slug/" component={Addon} />
       <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
+      <Route path="user/:username/" component={UserProfile} />
       <Route path="collections/:user/:slug/" component={Collection} />
       <Route path=":visibleAddonType/categories/" component={CategoriesPage} />
       <Route path=":visibleAddonType/:slug/" component={Category} />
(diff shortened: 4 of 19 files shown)


tofumatt:
This is ready for review now 🎉

src/amo/routes.js
```diff
@@ -2,34 +2,34 @@ import config from 'config';
...
-import SearchPage from './components/SearchPage';
-import ServerError from './components/ErrorPage/ServerError';
-import Collection from './components/Collection';
-
 // If you add a new route here, check that the nginx rules maintained by ops
 // are in sync. See:
 // https://github.com/mozilla-services/puppet-config/tree/master/amo
 export default (
-  <div>
+  <React.Fragment>
```
tofumatt: Not wrapping our routes in a `<div>` seemed nice 😆

tofumatt:
Will require #4284 once merged.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+    if (oldParams.username !== newParams.username) {
+      dispatch(fetchUserAccount({
+        errorHandlerId: errorHandler.id,
+        username: newParams.username,
+      }));
+    }
+  }
+
+  getUser(username: string) {
```
tofumatt: Sorry, this needs to be moved to `mapStateToProps`, I was just doing it here lazily while hacking and forgot to fix it. I'll be getting it in a few.

kumar303: ok, let me know when it's ready for review

tofumatt:
Alrighty, change made and should be good.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+      i18n,
+      params,
+      user,
+    } = this.props;
+
+    if (errorHandler.hasError()) {
+      log.warn('Captured API Error:', errorHandler.capturedError);
+
+      if (errorHandler.capturedError.responseStatusCode === 404) {
+        return <NotFound errorCode={errorHandler.capturedError.code} />;
```
kumar303: Ugh. This is annoying because it's custom logic that needs its own tests and maintenance. We will fix it in https://github.com/mozilla/addons-frontend/issues/2632

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+      }
+    );
+
+    return (
+      <div className="UserProfile">
+        <Helmet>
+          <title>{userProfileTitle}</title>
+        </Helmet>
+
+        {errorHandler.renderErrorIfPresent()}
```
kumar303: Since you're already checking for the error, the rendering logic should all happen in the same place to avoid confusion. It can look like this up above:

```js
let errorMessage;
  if (errorHandler.capturedError.responseStatusCode === 404) {
    return <NotFound .../>;
  }

  errorMessage = errorHandler.renderError();
}
```
...
```js
return (
  <div className="UserProfile">
    {errorMessage}
  </div>
);
```

Otherwise it looks like you are forgetting to handle the error in the first if statement.

config/default.js
```diff
@@ -288,4 +289,9 @@ module.exports = {
 
   // If true, show a link to the beta versions page on the listing page.
   betaVersions: true,
+
+  // Show and link to User Profiles using the front-end. These pages
+  // are still in development so this should only be enabled locally and
+  // on `-dev`.
+  enableUserProfile: false,
```
kumar303: Using config flags like this is a great way to ship half-working things. It helps a lot because we can iterate in smaller patches which are easier to review. Your patch here is already small, I just wanted to encourage this type of approach for future patches.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,315 @@
...
+    const { store } = dispatchSignInActions({
+      userProps: {
+        userId: 500,
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(
+      root.find('.UserProfile-user-info').shallow().find('.UserProfile-name')
```
kumar303: Doing a shallow render of `Card` isn't a good approach because it reaches into the implementation of `Card`. This might break if `Card` requires a context property in the future or if it its implementation changes in other ways. 

The `Card` component already has a well established interface that you are relying on: the `header` property. You can just reach in and render it like:

```js
const headerContent = root.find('.UserProfile-user-info').prop('header');
const header = shallow(<div>{headerContent}</div>);
expect(header.find('.UserProfile-name')).toHaveText('tofumatt');
```

This comment applies to other tests that use `root.find(...).shallow()`.

tofumatt:
FYI this is ready.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+        <UserAvatar className="UserProfile-avatar" user={user} />
+
+        <h1 className="UserProfile-name">
+          {user ? user.displayName : <LoadingText />}
+        </h1>
+      </React.Fragment>
+    );
+    const userProfileTitle = i18n.sprintf(
+      i18n.gettext('User Profile for %(user)s'), {
+        user: user ? user.displayName : params.username,
```
kumar303: Good idea on the fallback while loading

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+            header={userProfileHeader}
+          >
+            <DefinitionList className="UserProfile-dl">
+              {user && user.homepage ? (
+                <Definition
+                  className="UserProfile-homepage"
+                  term={i18n.gettext('Homepage')}
+                >
+                  {/* TODO: Make a helper that strips https?:\/\/ from text */}
+                  <a href={user.homepage}>{user.homepage.replace(/^https?:\/\//, '')}</a>
```
kumar303: Well, huh, I'm not sure how Flow lets you call `.replace()`. `user.homepage` can be `null` so I think you need protect against that.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+              >
+                {user ? user.num_addons_listed : <LoadingText />}
+              </Definition>
+              <Definition
+                className="UserProfile-rating-average"
+                term={i18n.gettext('Average rating of developer’s add-ons')}
+              >
+                {user ? (
+                  <Rating
+                    rating={parseFloat(user.average_addon_rating)}
```
kumar303: Why did you need `parseFloat()`? `Rating` already [calls it](https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/Rating/index.js#L98).

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+                  />
+                ) : <LoadingText />}
+              </Definition>
+            </DefinitionList>
+
+            {user && user.biography && user.biography.length ? (
+              <div
+                className="UserProfile-biography"
+                // eslint-disable-next-line react/no-danger
+                dangerouslySetInnerHTML={sanitizeUserHTML(user.biography)}
```
kumar303: The old site supported a bunch of tags. We should probably start by doing the same:

screenshot 2018-03-14 11 59 49

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+          </Card>
+        </div>
+      </div>
+    );
+  }
+}
+
+export function mapStateToProps(
+  state: {| users: UsersStateType |},
+  ownProps: {| params: UserProfileParams |},
```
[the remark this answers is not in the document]
tofumatt: Blarg, oops! Thanks.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+) {
+  return {
+    user: state.users.byID[state.users.byUsername[ownProps.params.username]],
```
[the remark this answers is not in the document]
tofumatt: Good call.

src/amo/components/UserProfile/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+
+type UserProfileParams = {| username: string |};
+
+type Props = {|
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  params: UserProfileParams,
+  user: UserType,
```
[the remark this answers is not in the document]
tofumatt: Oops, yes, you're totally right. Thanks!

src/amo/components/UserProfile/index.js
tofumatt: I saw that in other components and I thought it seemed more verbose... but fair point about it not looking like an error is checked until then otherwise. I'll change it 👍

src/amo/components/UserProfile/index.js
tofumatt: This code is only executed if `user && user.homepage` is truthy, so that's unlikely, but I will tidy this up regardless.

src/amo/components/UserProfile/index.js
tofumatt: Ohhhhhhh, I think it's because of how things are called. `Rating` calls `toFixed(1)` first, but I think the data returned by the API caused an issue with `.toFixed(1)`. Maybe `Rating` needs to be fixed.

src/amo/components/UserProfile/index.js
tofumatt: That's exactly what `sanitizeUserHTML` does 😄:

https://github.com/mozilla/addons-frontend/blob/master/src/core/utils/index.js#L90

src/amo/components/UserProfile/index.js
tofumatt: Ah, and it's because of https://github.com/mozilla/addons-server/issues/7822

I can either add the `parseFloat` to the user reducer or here until https://github.com/mozilla/addons-server/issues/7822 is fixed, or keep the fix to `Rating` I added.

tofumatt:
Thanks for the review, this is good for another look.

Let me know what you think about the best way to work around https://github.com/mozilla/addons-server/issues/7822

src/amo/components/UserProfile/index.js
kumar303: oh duh, I read it as `sanitizeHTML()` for some reason

tofumatt: Hah, well that’s easy to mix up 😅

src/amo/components/UserProfile/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+  @include respond-to(medium) {
+    font-size: $font-size-xl;
+  }
+}
+
+.UserProfile-rating-average .Rating {
+  @include margin-start(-3px);
+
+  justify-content: start;
+  margin: 0;
```
kumar303: Hmm, doesn't this overwrite `margin-start(-3px)`?

src/amo/routes.js
```diff
@@ -39,6 +39,8 @@ export default (
       <IndexRoute component={Home} />
       <Route path="addon/:slug/" component={Addon} />
       <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
+      {config.get('enableUserProfile') &&
```
kumar303: Is this necessary? I don't think the URL is accessible on the frontend until we allow it from nginx, right?

src/core/utils/index.js
```diff
@@ -192,6 +193,13 @@ export function getErrorComponent(status) {
   }
 }
 
+export function removeProtocolFromURL(urlWithProtocol) {
+  invariant(urlWithProtocol, 'url is required');
```
kumar303: `url` should be `urlWithProtocol`

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,318 @@
...
+  });
+
+  it('renders the user avatar', () => {
+    const root = renderUserProfile();
+
+    const headerContent = root.find('.UserProfile-user-info').prop('header');
+    const header = shallow(<div>{headerContent}</div>);
+
+    expect(header.find(UserAvatar)).toHaveProp(
+      'user', getCurrentUser(root.instance().props.store.getState().users));
```
kumar303: It feels a little fragile to rely on `.props.store.getState()`. How about doing it with more setup?

```diff
diff --git a/tests/unit/amo/components/TestUserProfile.js b/tests/unit/amo/components/TestUserProfile.js
index 2291383e5..ec603f585 100644
--- a/tests/unit/amo/components/TestUserProfile.js
+++ b/tests/unit/amo/components/TestUserProfile.js
@@ -94,13 +94,17 @@ describe(__filename, () => {
   });
 
   it('renders the user avatar', () => {
-    const root = renderUserProfile();
+    const username = 'tofumatt';
+    const { state, store } = dispatchSignInActions({
+      userProps: { username },
+    });
+    const root = renderUserProfile({ store, params: { username } });
 
     const headerContent = root.find('.UserProfile-user-info').prop('header');
     const header = shallow(<div>{headerContent}</div>);
 
     expect(header.find(UserAvatar)).toHaveProp(
-      'user', getCurrentUser(root.instance().props.store.getState().users));
+      'user', getCurrentUser(state.users));
   });
 
   it('still passes user prop to avatar while loading', () => {
```

If that setup becomes too repetitive you could add a helper to sync up `params` and `user`.

I also find it helpful to see explicit setup of data for tests that are testing that same data.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,318 @@
...
+    const header = shallow(<div>{headerContent}</div>);
+
+    expect(header.find(UserAvatar)).toHaveProp('user', undefined);
+  });
+
+  it("renders the user's name", () => {
+    const root = renderUserProfile();
+
+    const headerContent = root.find('.UserProfile-user-info').prop('header');
+    const header = shallow(<div>{headerContent}</div>);
```
kumar303: These two lines are repeated a lot. You could move them to a helper like:

```js
const header = getHeaderComponent(root);
```

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,318 @@
...
+
+    expect(header.find('.UserProfile-name').find(LoadingText))
+      .toHaveLength(1);
+  });
+
+  it("renders the user's username if no display name exists", () => {
+    const { store } = dispatchSignInActions({
+      userProps: {
+        display_name: null,
+        userId: 500,
```
kumar303: This doesn't seem relevant to the test

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,318 @@
...
+    const root = renderUserProfile({ params: { username: 'not-loaded' } });
+
+    expect(root.find(ReportUserAbuse)).toHaveLength(1);
+  });
+
+  it('renders a not found page if the API request is a 404', () => {
+    const { store } = dispatchSignInActions();
+    const errorHandler = new ErrorHandler({
+      id: 'some-id',
+      capturedError: { responseStatusCode: 404 },
```
kumar303: This looks pretty fragile. [Here](https://github.com/mozilla/addons-frontend/blob/ec3a4dd837ac92516283f3db53e3c21aaec6efc8/tests/unit/amo/components/TestCollection.js#L485-L493) is a better way to do it.

src/ui/components/Rating/index.js
```diff
@@ -95,7 +95,7 @@ export class RatingBase extends React.Component {
     let description;
     if (rating) {
       description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
-        { rating: i18n.formatNumber(parseFloat(rating.toFixed(1))) });
+        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
```
kumar303: This seems like the right thing to do but can you add a test for it? I didn't see one.

src/amo/components/UserProfile/styles.scss
tofumatt: It doesn't, *but* @tsl143 used `margin()` elsewhere and I should really just use that.

src/amo/routes.js
tofumatt: I guess not, but it seemed more complete/proper. I can remove it, but I just figured leaving it here was a bit of code-as-documentation that this is pref'd. I can just leave a comment though, fair enough.

src/core/utils/index.js
tofumatt: I guess... really it's just saying the argument should be truthy; as it's not a named argument I thought "url" was more helpful. But I guess `urlWithProtocol` will be easier to grep.

tests/unit/amo/components/TestUserProfile.js
tofumatt: Fair enough. It's a bit fragile but I found it less work.

On the second point I totally agree though, so yeah I'll refactor this. It's handy to see all the props set in the test. 👍

tests/unit/amo/components/TestUserProfile.js
tofumatt: Hah, I was thinking that when I refactored it and I was lazy. Thanks for calling me out on that, will fix 👍

tests/unit/amo/components/TestUserProfile.js
tofumatt: It's not, shoot, I remember why I did it and that's no longer relevant. Thanks! 🔥

tests/unit/amo/components/TestUserProfile.js
tofumatt: I feel like I tried that and something broke, but I was probably doing it wrong™.

src/ui/components/Rating/index.js
tofumatt: Will do! @diox helpfully patched the API response but we still feed integers to it in some of the tests and also it just makes sense to do. I'll add it. 👍

tests/unit/amo/components/TestUserProfile.js
tofumatt: It worked, I think I must've just forgot to pass `store.dispatch` last time 😓

tofumatt:
Thanks for the review; changes made, good for another look.

src/amo/components/UserProfile/styles.scss
```diff
@@ -0,0 +1,48 @@
...
+  font-size: $font-size-l;
+  margin: 0;
+
+  @include respond-to(medium) {
+    font-size: $font-size-xl;
+  }
+}
+
+.UserProfile-rating-average .Rating {
+  @include margin(0, 0, 0, -3px);
```
kumar303: Seeing a `-3px` negative margin in a review is hard because I can't tell if the final margin follows our standard spacing. Maybe `Rating` needs a new class that can be toggled on via configuration? I see that search results need to adjust it too. No need to address it here but maybe file as a code quality issue.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,321 @@
...
+  it("renders the user's biography", () => {
+    const { store } = dispatchSignInActions({
+      userProps: {
+        biography: 'Not even vegan!',
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-biography').render().html())
```
kumar303: You can remove the `.render()` part. You only need `.html()`.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,321 @@
...
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+
+    const root = renderUserProfile({
+      params: { username: 'black-panther' },
+      store,
+    });
+
+    dispatchSpy.reset();
+
+    sinon.assert.notCalled(dispatchSpy);
```
kumar303: This is not necessary after calling `.reset()`

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,321 @@
...
```
kumar303: same here

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -0,0 +1,321 @@
...
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-biography')).toHaveLength(0);
+  });
+
+  it('omits an empty biography', () => {
+    const { store } = dispatchSignInActions({
+      userProps: {
+        biography: '',
+        username: 'tofumatt',
```
kumar303: The `username` is repeated everywhere to sync up with `params`. You could make a simple change to DRY it up:

```js
  userProps: defaultUserProps({
    biography: '',
  }),
});
```
Or you could wrap `dispatchSignInActions()` in a helper.

src/amo/components/UserProfile/styles.scss
tofumatt: Good call. I think it might be that in most places it indents itself too far over; in reality I think it needs a class for the original place it was used as I think the CSS was too coupled to the parent component originally. I'll file the follow-up issues, cheers 👍

tests/unit/amo/components/TestUserProfile.js
tofumatt: Whoops, right on.

tests/unit/amo/components/TestUserProfile.js
tofumatt: Indeed, I don't need to test sinon in our tests 😅

tests/unit/amo/components/TestUserProfile.js
tofumatt: Nice, will do!

tests/unit/amo/components/TestUserProfile.js
tofumatt: Actually, no, the `.html()` will return an object. I'll try a different matcher.

tofumatt: I think it's because it's doing dangerouslySetInnerHTML...

src/amo/components/UserProfile/styles.scss
tofumatt: Filed #4573.

PR #4472, opened 2018-02-28
feat: Add User Avatar component

### LINKED ISSUE #4471: Add a user Avatar component
We need a user avatar component that accepts a user object and renders either their avatar with an image tag or the placeholder SVG image we have.

tofumatt:
Adds a user avatar component that renders an anonymous user icon when the user is `null` (loading) or has no picture set.

fix #4471

### DIFF
--- a/tests/unit/amo/helpers.js
+++ b/tests/unit/amo/helpers.js
@@ -218,38 +218,15 @@ export function dispatchClientMetadata({
 
 export function dispatchSignInActions({
   authToken = userAuthToken(),
-  biography = 'I love making add-ons!',
-  created = '2017-08-15T12:01:13Z',
   userId = 12345,
-  username = 'user-1234',
-  /* eslint-disable camelcase */
-  average_addon_rating = 4.3,
-  display_name = null,
-  num_addons_listed = 1,
-  picture_url = `${config.get('amoCDN')}/static/img/zamboni/anon_user.png`,
-  picture_type = '',
-  /* eslint-enable camelcase */
-  permissions = [],
-  homepage = null,
+  userProps = {},
   ...otherArgs
 } = {}) {
   const { store } = dispatchClientMetadata(otherArgs);
 
   store.dispatch(setAuthToken(authToken));
   store.dispatch(loadCurrentUserAccount({
-    user: createUserAccountResponse({
-      id: userId,
-      biography,
-      created,
-      homepage,
-      username,
-      average_addon_rating,
-      display_name,
-      num_addons_listed,
-      permissions,
-      picture_url,
-      picture_type,
-    }),
+    user: createUserAccountResponse({ id: userId, ...userProps }),
   }));
 
   return {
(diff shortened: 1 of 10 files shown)


tests/unit/amo/helpers.js
```diff
@@ -218,11 +218,19 @@ export function dispatchClientMetadata({
 
 export function dispatchSignInActions({
   authToken = userAuthToken(),
+  biography = 'I love making add-ons!',
```
kumar303: Instead of duplicating all these defaults what about accepting a `user` parameter to `dispatchSignInActions()` that overrides `loadCurrentUserAccount({ user })`? You could keep the existing shortcuts used by existing tests. With that change you can make use of the `createUserAccountResponse()` helper.

tests/unit/ui/components/TestUserAvatar.js
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(root.find(Icon).find('.test')).toHaveLength(1);
+  });
+
+  it('renders an anonymous icon if the user has no photo', () => {
+    const { state } = dispatchSignInActions({
+      id: 599,
+      picture_url: 'anonymous.jpg',
+      // An empty picture type means no avatar.
+      picture_type: '',
```
kumar303: Not a big deal but the API docs suggest that this would be `null`.

tofumatt: Yeah, I noticed that and filed https://github.com/mozilla/addons-server/issues/7679. I've got responses from the API that show it as an empty string.

tests/unit/amo/helpers.js
tofumatt: Oh, sweet, I brute-forced this without really looking, but that's much better. Will do!

tofumatt: For what it's worth it wasn't much effort to remove the existing shortcuts and I think it makes it more consistent and removes the temptation/obviousness of adding more shortcuts in the future.

tofumatt:
I found it was easy enough to remove all the dispatchSignInActions shortcuts after talking about it. I know it's unrelated to the original patch but I think it's a nice bit of cleanup/consistency, could you have another look just to make sure it's alright?

tests/unit/amo/helpers.js
```diff
@@ -219,23 +219,14 @@ export function dispatchClientMetadata({
 export function dispatchSignInActions({
   authToken = userAuthToken(),
   userId = 12345,
-  username = 'user-1234',
-  // eslint-disable-next-line camelcase
-  display_name = null,
-  permissions = [],
+  user = {},
```
kumar303: I find this confusing because it looks like the caller is forgetting to use a base object. Take this as an example:

```js
       user: { permissions: [COLLECTIONS_EDIT] },
     });
```

It looks like the caller accidentally forgot to do:

```js
       user: createUserAccountResponse({ permissions: [COLLECTIONS_EDIT] }),
     });
```

But I know it's a useful shortcut. Maybe it would help just to rename the argument?

```js
       userProps: { permissions: [COLLECTIONS_EDIT] },
     });
```

Either way, I like the new approach to `dispatchSignInActions()`.

tofumatt: Oh, that's a good point. Yeah, I'll change it to `userProps`, good call. 👍

PR #4474, opened 2018-02-28
feat: Add User Abuse Reports

### LINKED ISSUE #4473: Create a ReportUserAbuse component
We'll need a component similar to the `ReportAbuseButton` but for the user abuse report feature on the user profile page.

It should use the `DismissibleTextForm` component though, unlike `ReportAbuseButton` which doesn't as it predates that component.

screenshot 2018-02-28 23 01 13

tofumatt:
fix #4473

It will be used in the user profile page implemented in #4469.

### DIFF
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -1,0 +1,176 @@
+/* @flow */
+import makeClassName from 'classnames';
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import {
+  hideUserAbuseReportUI,
+  sendUserAbuseReport,
+  showUserAbuseReportUI,
+} from 'amo/reducers/userAbuseReports';
+import { withErrorHandler } from 'core/errorHandler';
+import translate from 'core/i18n/translate';
+import { sanitizeHTML } from 'core/utils';
+import Button from 'ui/components/Button';
+import DismissibleTextForm from 'ui/components/DismissibleTextForm';
+import type { UserAbuseReportsState } from 'amo/reducers/userAbuseReports';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { I18nType } from 'core/types/i18n';
+import type { UserType } from 'amo/reducers/users';
+import type { OnSubmitParams } from 'ui/components/DismissibleTextForm';
+
+import './styles.scss';
+
+
+type Props = {|
+  abuseReports: UserAbuseReportsState,
+  className?: string,
+  dispatch: Function,
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  user?: UserType,
+|};
+
+export class ReportUserAbuseBase extends React.Component<Props> {
+  props: Props;
+
+  hideReportUI = () => {
+    const { dispatch, user } = this.props;
+
+    if (user) {
+      dispatch(hideUserAbuseReportUI({ user }));
+    }
+  }
+
+  sendReport = (reportData: OnSubmitParams) => {
+    const { dispatch, errorHandler, user } = this.props;
+
+    if (user) {
+      dispatch(sendUserAbuseReport({
+        errorHandlerId: errorHandler.id,
+        message: reportData.text,
+        user,
+      }));
+    }
+  }
+
+  showReportUI = () => {
+    const { dispatch, user } = this.props;
+
+    if (user) {
+      dispatch(showUserAbuseReportUI({ user }));
+    }
+  }
+
+  render() {
+    const { abuseReports, className, errorHandler, i18n, user } = this.props;
+    const abuseReport = user ? abuseReports.byUserId[user.id] : null;
+
+    if (!user) {
+      return null;
+    }
+
+    return (
+      <div
+        className={makeClassName('ReportUserAbuse', className, {
+          'ReportUserAbuse--is-expanded': abuseReport && abuseReport.uiVisible,
+        })}
+      >
+        {errorHandler.renderErrorIfPresent()}
+
+        {
+          (!abuseReport || !abuseReport.uiVisible) &&
+          (!abuseReport || !abuseReport.hasSubmitted) && (
+          <Button
+            buttonType="neutral"
+            className="ReportUserAbuse-show-more"
+            disabled={!user}
+            onClick={this.showReportUI}
+            puffy
+          >
+            {i18n.gettext('Report this user for abuse')}
+          </Button>
+        )}
+
+        {(abuseReport && !abuseReport.hasSubmitted) && (
+          <div className="ReportUserAbuse-form">
+            <h2 className="ReportUserAbuse-header">
+              {i18n.gettext('Report this user for abuse')}
+            </h2>
+
+            <p
+              /* eslint-disable react/no-danger */
+              dangerouslySetInnerHTML={sanitizeHTML(
+                i18n.sprintf(i18n.gettext(
+                  `If you think this user is violating
+                  %(linkTagStart)sMozilla's Add-on Policies%(linkTagEnd)s,
+                  please report this user to Mozilla.`
+                ), {
+                  linkTagStart: '<a href="https://developer.mozilla.org/en-US/Add-ons/AMO/Policy">',
+                  linkTagEnd: '</a>',
+                })
+              )}
+              /* eslint-enable react/no-danger */
+            />
+            <p>
+              {i18n.gettext(
+                `Please don't use this form to report bugs or contact this
+                user; your report will only be sent to Mozilla and not
+                to the add-on developer.`)}
+            </p>
+
+            <DismissibleTextForm
+              isSubmitting={
+                abuseReport.isSubmitting && !errorHandler.hasError()
+              }
+              onDismiss={this.hideReportUI}
+              onSubmit={this.sendReport}
+              placeholder={i18n.gettext(
+                'Explain how this user is violating our policies.'
+              )}
+              submitButtonText={i18n.gettext('Send abuse report')}
+              submitButtonInProgressText={i18n.gettext('Sending abuse report')}
+            />
+          </div>
+        )}
+
+        {abuseReport && abuseReport.hasSubmitted && (
+          <div className="ReportUserAbuse--report-sent">
+            <h3 className="ReportUserAbuse-header">
+              {i18n.gettext('You reported this user for abuse')}
+            </h3>
+
+            <p className="ReportUserAbuse-first-paragraph">
+              {i18n.gettext(
+                `We have received your report. Thanks for letting us know about
+                your concerns with this user.`
+              )}
+            </p>
+
+            <p>
+              {i18n.gettext(
+                `We can't respond to every abuse report but we'll look into
+                this issue.`
+              )}
+            </p>
+          </div>
+        )}
+      </div>
+    );
+  }
+}
+
+export const mapStateToProps = (
+  state: {| userAbuseReports: UserAbuseReportsState |}
+) => {
+  return {
+    abuseReports: state.userAbuseReports,
+  };
+};
+
+export default compose(
+  connect(mapStateToProps),
+  translate(),
+  withErrorHandler({ name: 'ReportUserAbuse' }),
+)(ReportUserAbuseBase);
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -143,7 +143,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
               {i18n.gettext('You reported this user for abuse')}
             </h3>
 
-            <p className="ReportUserAbuse-first-paragraph">
+            <p>
               {i18n.gettext(
                 `We have received your report. Thanks for letting us know about
                 your concerns with this user.`
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -25,11 +25,13 @@ import './styles.scss';
 
 
 type Props = {|
-  abuseReports: UserAbuseReportsState,
   className?: string,
   dispatch: DispatchFunc,
+  hasSubmitted: bool,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  isSubmitting: bool,
+  uiVisible: bool,
   user?: UserType,
 |};
 
@@ -65,24 +67,25 @@ export class ReportUserAbuseBase extends React.Component<Props> {
   }
 
   render() {
-    const { abuseReports, className, errorHandler, i18n, user } = this.props;
-    const abuseReport = user ? abuseReports.byUserId[user.id] : null;
-
-    if (!user) {
-      return null;
-    }
+    const {
+      className,
+      errorHandler,
+      hasSubmitted,
+      i18n,
+      isSubmitting,
+      uiVisible,
+      user,
+    } = this.props;
 
     return (
       <div
         className={makeClassName('ReportUserAbuse', className, {
-          'ReportUserAbuse--is-expanded': abuseReport && abuseReport.uiVisible,
+          'ReportUserAbuse--is-expanded': uiVisible,
         })}
       >
         {errorHandler.renderErrorIfPresent()}
 
-        {
-          (!abuseReport || !abuseReport.uiVisible) &&
-          (!abuseReport || !abuseReport.hasSubmitted) && (
+        {(!uiVisible && !hasSubmitted) && (
           <Button
             buttonType="neutral"
             className="ReportUserAbuse-show-more"
@@ -94,7 +97,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
           </Button>
         )}
 
-        {(abuseReport && !abuseReport.hasSubmitted) && (
+        {(!hasSubmitted) && (
           <div className="ReportUserAbuse-form">
             <h2 className="ReportUserAbuse-header">
               {i18n.gettext('Report this user for abuse')}
@@ -122,9 +125,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
             </p>
 
             <DismissibleTextForm
-              isSubmitting={
-                abuseReport.isSubmitting && !errorHandler.hasError()
-              }
+              isSubmitting={isSubmitting}
               onDismiss={this.hideReportUI}
               onSubmit={this.sendReport}
               placeholder={i18n.gettext(
@@ -136,7 +137,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
           </div>
         )}
 
-        {abuseReport && abuseReport.hasSubmitted && (
+        {(hasSubmitted) && (
           <div className="ReportUserAbuse--report-sent">
             <h3 className="ReportUserAbuse-header">
               {i18n.gettext('You reported this user for abuse')}
@@ -163,10 +164,17 @@ export class ReportUserAbuseBase extends React.Component<Props> {
 }
 
 export const mapStateToProps = (
-  state: {| userAbuseReports: UserAbuseReportsState |}
+  state: {| userAbuseReports: UserAbuseReportsState |},
+  ownProps: {| user?: UserType |},
 ) => {
+  const abuseReport = (
+    ownProps.user && state.userAbuseReports.byUserId[ownProps.user.id]
+  ) ? state.userAbuseReports.byUserId[ownProps.user.id] : {};
+
   return {
-    abuseReports: state.userAbuseReports,
+    hasSubmitted: abuseReport.hasSubmitted || false,
+    isSubmitting: abuseReport.isSubmitting || false,
+    uiVisible: abuseReport.uiVisible || false,
   };
 };
 
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -121,7 +121,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
               {i18n.gettext(
                 `Please don't use this form to report bugs or contact this
                 user; your report will only be sent to Mozilla and not
-                to the add-on developer.`)}
+                to this user.`)}
             </p>
 
             <DismissibleTextForm
@@ -165,7 +165,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
 
 export const mapStateToProps = (
   state: {| userAbuseReports: UserAbuseReportsState |},
-  ownProps: {| user?: UserType |},
+  ownProps: Props,
 ) => {
   const abuseReport = (
     ownProps.user && state.userAbuseReports.byUserId[ownProps.user.id]
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -16,6 +16,7 @@ import Button from 'ui/components/Button';
 import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import type { UserAbuseReportsState } from 'amo/reducers/userAbuseReports';
 import type { ErrorHandlerType } from 'core/errorHandler';
+import type { DispatchFunc } from 'core/types/redux';
 import type { I18nType } from 'core/types/i18n';
 import type { UserType } from 'amo/reducers/users';
 import type { OnSubmitParams } from 'ui/components/DismissibleTextForm';
@@ -26,7 +27,7 @@ import './styles.scss';
 type Props = {|
   abuseReports: UserAbuseReportsState,
   className?: string,
-  dispatch: Function,
+  dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
   user?: UserType,
--- a/src/amo/components/ReportUserAbuse/index.js
+++ b/src/amo/components/ReportUserAbuse/index.js
@@ -42,7 +42,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
     const { dispatch, user } = this.props;
 
     if (user) {
-      dispatch(hideUserAbuseReportUI({ user }));
+      dispatch(hideUserAbuseReportUI({ userId: user.id }));
     }
   }
 
@@ -53,7 +53,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
       dispatch(sendUserAbuseReport({
         errorHandlerId: errorHandler.id,
         message: reportData.text,
-        user,
+        userId: user.id,
       }));
     }
   }
@@ -62,7 +62,7 @@ export class ReportUserAbuseBase extends React.Component<Props> {
     const { dispatch, user } = this.props;
 
     if (user) {
-      dispatch(showUserAbuseReportUI({ user }));
+      dispatch(showUserAbuseReportUI({ userId: user.id }));
     }
   }
 
--- a/src/amo/components/ReportUserAbuse/styles.scss
+++ b/src/amo/components/ReportUserAbuse/styles.scss
@@ -1,0 +1,15 @@
+.ReportUserAbuse-header {
+  margin-bottom: 0;
+}
+
+.ReportUserAbuse-first-paragraph {
+  margin-top: 6px;
+}
+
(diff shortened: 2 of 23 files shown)


tests/unit/ui/components/TestDismissibleTextForm.js
```diff
@@ -107,20 +107,6 @@ describe(__filename, () => {
     sinon.assert.called(onDismiss);
   });
 
-  it('clears the text when dismissing a textarea', () => {
```
tofumatt: This behaviour didn't seem needed for anything and interfered with cancelling an abuse report but then opening it again–it would clear the contents of the textarea. This didn't seem there for a reason and was bad UX (as "cancel" might've been clicked by mistake, undoing a lot of effort in typing the abuse report).

tofumatt:
Sorry this one ended up being pretty massive. 😢

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+
+export type UserAbuseReportsState = {|
+  byUserId: {
+    [userId: number]: UserAbuseReportState,
+  },
+|};
+
+export default function userAbuseReportReducer(
+  state: UserAbuseReportsState = initialState,
+  action: Object
```
kumar303: I think it would be helpful to track real action objects: `action: SendUserAbuseReportAction | LoadUserAbuseReportAction`.

You would then need to add the action type hack (sorry):

```js
export const HIDE_USER_ABUSE_REPORT_UI: 'HIDE_USER_ABUSE_REPORT_UI' = 'HIDE_USER_ABUSE_REPORT_UI';
```

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+
+type SendUserAbuseReportAction = {|
+  errorHandlerId: string,
+  message: string,
+  user: UserType | null,
+|};
+
+export function sendUserAbuseReport(
+  { errorHandlerId, message, user }: SendUserAbuseReportAction = {}
+) {
```
kumar303: It would be helpful to define the returned action type

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+  };
+}
+
+type ShowUserAbuseReportUIType = {|
+  user?: UserType,
+|};
+
+export function showUserAbuseReportUI(
+  { user }: ShowUserAbuseReportUIType = {}
+) {
```
kumar303: This is also missing a returned action type

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+    case LOAD_USER_ABUSE_REPORT: {
+      const { message, reporter, user } = action.payload;
+      return {
+        ...state,
+        byUserId: {
+          ...state.byUserId,
+          [user.id]: {
+            message,
+            reporter,
+            user,
```
kumar303: Do we need the whole user object in state? Could it just be a user ID? Also, `user` is missing from `UserAbuseReportState`. I think fixing the reducer's return type definition will catch that.

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+export const SEND_USER_ABUSE_REPORT = 'SEND_USER_ABUSE_REPORT';
+export const SHOW_USER_ABUSE_REPORT_UI = 'SHOW_USER_ABUSE_REPORT_UI';
+
+
+type HideUserAbuseReportUIType = {|
+  user: UserType,
+|};
+
+export function hideUserAbuseReportUI(
+  { user }: HideUserAbuseReportUIType = {}
```
kumar303: It looks to me like only `userId` is required for this action

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+  return {
+    type: HIDE_USER_ABUSE_REPORT_UI,
+    payload: { user },
+  };
+}
+
+type LoadUserAbuseReportType = {|
+  message: string,
+  reporter: Object | null,
+  user: {|
```
kumar303: Should this be `UserType`?

src/core/api/abuse.js
```diff
@@ -20,3 +19,20 @@ export function reportAddon(
...
   });
 }
+
+export type ReportUserParams = {|
+  api: ApiStateType,
+  message: string,
+  user: UserType,
+|};
+
+export function reportUser({ api, message, user }: ReportUserParams) {
```
kumar303: I'd like to see the return type defined, like `: Promise<ReportUserReponse>` because it would be helpful to know how the saga should deal with it.

src/amo/sagas/userAbuseReports.js
```diff
@@ -0,0 +1,49 @@
+// Disabled because of
```
kumar303: After https://github.com/mozilla/addons-frontend/issues/4323 , I feel like we should add flow types for sagas since they are the "glue" between actions and API calls. It's hard to get these integrations right and our tests can't cover them (we can't make real API requests). This is where type safety really shines.

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,175 @@
...
+export type UserAbuseReportsState = {|
+  byUserId: {
+    [userId: number]: UserAbuseReportState,
+  },
+|};
+
+export default function userAbuseReportReducer(
+  state: UserAbuseReportsState = initialState,
+  action: Object
+) {
```
kumar303: This should specify a return value of `UserAbuseReportsState` so that Flow can help validate each reducer case.

src/amo/sagas/userAbuseReports.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    yield put(loadUserAbuseReport({
+      message: response.message,
+      reporter: response.reporter,
+      user: response.user,
+    }));
+  } catch (error) {
+    log.warn(`Reporting user for abuse failed: ${error}`);
+    yield put(errorHandler.createErrorAction(error));
```
kumar303: In addition to dispatching this action, I think it helps to dispatch an `abortReportUser` action that can reset all the state. The alternative is for the component to check `errorHandler.hasError()` every step of the way but that's hard to get right. The render states for `ReportUserAbuse` are already complex and I'm not sure they fully handle the `errorHandler.hasError()` case.

src/amo/reducers/userAbuseReports.js
tofumatt: Indeed, but I thought passing around the user object consistently made for a nicer API and made the types declared in flow easier to manage (everything is just a `UserType`). I didn't see the harm in passing a user to the action and having the reducer handle getting what it needs.

src/amo/reducers/userAbuseReports.js
tofumatt: No, this is just a subset of data returned by the API for the reported user: https://addons-server.readthedocs.io/en/latest/topics/api/abuse.html#submitting-a-user-abuse-report

src/amo/reducers/userAbuseReports.js
tofumatt: Sure thing!

src/amo/reducers/userAbuseReports.js
tofumatt: As mentioned this isn't the entire user object but just what's returned from the API. That said nothing from them is really needed so fair enough, this can likely just be removed.

src/amo/sagas/userAbuseReports.js
tofumatt: I recall flow annotations on sagas being pretty unintuitive/unhelpful to look at directly so I think that's why I opted out of writing them traditionally. I think it was mostly around the generator functions though.

I'd be fine with a code quality issue for adding flow types to sagas though.

src/amo/components/ReportUserAbuse/index.js
```diff
@@ -0,0 +1,176 @@
...
+    const { dispatch, user } = this.props;
+
+    if (user) {
+      dispatch(showUserAbuseReportUI({ user }));
+    }
+  }
+
+  render() {
+    const { abuseReports, className, errorHandler, i18n, user } = this.props;
+    const abuseReport = user ? abuseReports.byUserId[user.id] : null;
```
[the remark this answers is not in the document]
tofumatt: Shoot, totally true. I think I meant to do that after I was prototyping this with the entire object and just forgot, thanks!

src/amo/components/ReportUserAbuse/index.js
```diff
@@ -0,0 +1,176 @@
...
+import type { UserType } from 'amo/reducers/users';
+import type { OnSubmitParams } from 'ui/components/DismissibleTextForm';
+
+import './styles.scss';
+
+
+type Props = {|
+  abuseReports: UserAbuseReportsState,
+  className?: string,
+  dispatch: Function,
```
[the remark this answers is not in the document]
tofumatt: Thanks! I forget a lot about the types we have for stuff like this.

src/amo/components/ReportUserAbuse/index.js
```diff
@@ -0,0 +1,176 @@
...
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  user?: UserType,
```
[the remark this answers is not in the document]
tofumatt: The current implementation allows the report button to appear (but be disabled) during render when user is false-y. This improves the loading rendering state of the `UserProfile` this component appears in. Sorry it isn't clear from the code now but allowing an empty user object means the final thing looks better.

src/amo/components/ReportUserAbuse/styles.scss
```diff
@@ -0,0 +1,15 @@
+.ReportUserAbuse-header {
+  margin-bottom: 0;
+}
+
+.ReportUserAbuse-first-paragraph {
```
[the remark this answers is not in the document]
tofumatt: Yeah that's fair 😄

tofumatt:
I've made all the requested changes to this and added the abort action. It made the PR quite big but if those were the main issues I figured I'd leave it as the one PR here for history and because all of these components are intertwined anyway. But sorry about the big PR, I'll do these separately in the future–I've tended to try in the past but just got carried away as I needed this component working to put it in the UserProfile page where this stemmed from. 😞 

Anyway, set for another review!

tests/unit/amo/sagas/testUserAbuseReports.js
```diff
@@ -0,0 +1,156 @@
...
+    const abortAction = abortUserAbuseReport({ user });
+    await sagaTester.waitFor(abortAction.type);
+    expect(sagaTester.getCalledActions()[3]).toEqual(abortAction);
+  });
+
+  it('throws an error if multiple reports are submitted for the same user', async () => {
+    const user = createUserAccountResponse({ id: 50 });
+
+    _sendUserAbuseReport({
+      message: 'This add-on is malwaré!',
```
kumar303: This user...

tests/unit/core/api/test_abuse.js
```diff
@@ -52,4 +54,38 @@ describe(__filename, () => {
...
   });
+
+  describe('reportUser', () => {
+    function mockResponse({ message, user }) {
+      return createApiResponse({
+        jsonData: createFakeUserAbuseReport({ message, user }),
+      });
+    }
+
+    it('calls the report add-on abuse API', () => {
```
kumar303: This should use `async`

src/core/api/abuse.js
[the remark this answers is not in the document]
tofumatt: I didn't add return types to these because the responses are documented in the API so it feels like a duplication of effort. I get it increases flow coverage though; I'll add them.

src/amo/components/ReportUserAbuse/index.js
```diff
@@ -0,0 +1,185 @@
...
+                  linkTagEnd: '</a>',
+                })
+              )}
+              /* eslint-enable react/no-danger */
+            />
+            <p>
+              {i18n.gettext(
+                `Please don't use this form to report bugs or contact this
+                user; your report will only be sent to Mozilla and not
+                to the add-on developer.`)}
```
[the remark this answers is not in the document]
tofumatt: They shouldn't be; only add-on developers have a profile visible to regular users. I suppose if you were an admin you'd be able to report a regular user... I'll change it, but it's worth noting non-developer profiles return a 404 for regular users now.

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,251 @@
...
+
+type AbortUserAbuseReportAction = {|
+  type: typeof ABORT_USER_ABUSE_REPORT,
+  payload: AbortUserAbuseReportParams,
+|};
+
+export function abortUserAbuseReport(
+  { user }: AbortUserAbuseReportParams = {}
+): AbortUserAbuseReportAction {
+  if (!user) {
```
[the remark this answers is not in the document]
tofumatt: Neato, we finally did it! 😄

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,251 @@
...
+export const LOAD_USER_ABUSE_REPORT: 'LOAD_USER_ABUSE_REPORT' =
+  'LOAD_USER_ABUSE_REPORT';
+export const SEND_USER_ABUSE_REPORT: 'SEND_USER_ABUSE_REPORT' =
+  'SEND_USER_ABUSE_REPORT';
+export const SHOW_USER_ABUSE_REPORT_UI: 'SHOW_USER_ABUSE_REPORT_UI' =
+  'SHOW_USER_ABUSE_REPORT_UI';
+
+
+type AbortUserAbuseReportParams = {|
+  user: UserType,
```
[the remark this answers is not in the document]
tofumatt: I explained before but passing around the User object when we already have a `UserType` seemed nicer than making a new type–which is just a number so arguably is less of a strict type. There doesn't seem to be a downside to passing around the entire object (and we aren't storing the whole user object again in state), and this way we can use more info later if we like. Also I just liked passing around `user` over `userId: user.id`. I don't see it simplifying anything in terms of the complexity of the code. Also, it arguably makes debugging a bit easier as you can see more about the user when using the Redux devtools and the action's payload.

If there's some performance hit to passing the entire `User` object around (which, if anything, there shouldn't be as it should be passed as a reference in JS rather than as a new integer when we pass just the ID) then I think it makes sense to change, but otherwise it seems to be just an aesthetic choice.

tests/unit/amo/sagas/testUserAbuseReports.js
```diff
@@ -0,0 +1,156 @@
...
+
+    _sendUserAbuseReport({ message, user });
+
+    const expectedLoadAction = loadUserAbuseReport({
+      message: response.message,
+      reporter: response.reporter,
+      user,
+    });
+
+    await sagaTester.waitFor(expectedLoadAction.type);
```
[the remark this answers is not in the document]
tofumatt: Hah, neato! Undocumented makes me scared it'll go away but that's cool! 😉

tests/unit/amo/sagas/testUserAbuseReports.js
tofumatt: Sorry, copypasta from the add-ons abuse reporting tests.

tofumatt:
I made the changes save for changing `user` to `userId`, because I don't think it simplifies things. If I'm missing something there let me know. This is ready for another review; thanks for the invariant tip, it certainly cuts down on test noise too!

src/core/api/abuse.js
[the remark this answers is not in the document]
kumar303: Documentation is great but we can't use it to verify how our program works. Using strong types for the return values of API functions helps us validate saga code and anything else that interacts with API functions. Sagas have broken before because we don't have test coverage -- we don't touch a real API in our tests. This is a crucial part of the application to get right because it's impractical to run tests against a real API.

tests/unit/amo/sagas/testUserAbuseReports.js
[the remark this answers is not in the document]
kumar303: I submitted a patch to their docs but no one seems to care. D'oh. https://github.com/wix/redux-saga-tester/pull/49

src/core/api/abuse.js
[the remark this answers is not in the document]
tofumatt: Ten four. I did add types for it 😄

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+  const defaultRenderProps = {
+    errorHandler: createStubErrorHandler(),
+    i18n: fakeI18n(),
+    store: dispatchClientMetadata().store,
+    user: createUserAccountResponse({ id: 9 }),
+  };
+
+  function renderShallow({ ...props } = {}) {
+    return shallowUntilTarget(
+      <ReportUserAbuse {...defaultRenderProps} {...props} />,
```
kumar303: I don't think this is a good idea to use a static object here (`defaultRenderProps`) because it means all tests share one `store` object. You should either create `store` in `beforeEach()` or create it each time `renderShallow()` is called (by calling a function or something). The `errorHandler` prop would have the same problem.

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+    expect(root.find('.ReportUserAbuse')).toHaveLength(1);
+    expect(root.find('.ReportUserAbuse-show-more'))
+      .toHaveProp('disabled', true);
+  });
+
+  it('renders a button to report an user', () => {
+    const root = renderShallow();
+
+    expect(root.find('.ReportUserAbuse')).toHaveLength(1);
+    expect(root.find('.ReportUserAbuse-show-more').prop('children'))
```
kumar303: Can you not use [`children()`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/children.html)?

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+    expect(root.find('.ReportUserAbuse--is-expanded')).toHaveLength(0);
+  });
+
+  it('shows more content when showReportUI is called', () => {
+    const { store } = dispatchClientMetadata();
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const user = createUserAccountResponse();
+    const root = renderShallow({ store, user });
+
+    root.instance().showReportUI();
```
kumar303: Why not do `root.find('ReportUserAbuse-show-more').simulate('click', createFakeEvent())`? That would be more realistic and would cover the onClick logic.

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+  it('hides more content when hideReportUI is called', () => {
+    const { store } = dispatchClientMetadata();
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const user = createUserAccountResponse();
+    const root = renderShallow({ store, user });
+
+    root.instance().showReportUI();
+    dispatchSpy.reset();
+
+    root.instance().hideReportUI();
```
kumar303: For these kind of callbacks I think it's better to simulate what `DismissibleTextForm` will do. This will enforce the contract that `ReportUserAbuse` has with the interface of `DismissibleTextForm`. Example:

```js
const form = root.find(DismissibleTextForm);
const dismiss = form.prop('onDismiss');
// Simulate a dismissal:
dismiss();
```

If you end up doing this a lot you can make a helper for it like `simulateDimissal(root)`

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+    sinon.assert.calledWith(dispatchSpy, hideUserAbuseReportUI({ user }));
+  });
+
+  it('dispatches the send abuse report action', () => {
+    const { store } = dispatchClientMetadata();
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const user = createUserAccountResponse();
+    const root = renderShallow({ store, user });
+
+    root.instance().sendReport({ text: 'This user is funny' });
```
kumar303: Again, this would be more effective if it were to simulate `onSubmit()`.

tests/unit/amo/components/TestReportUserAbuse.js
```diff
@@ -0,0 +1,142 @@
...
+    expect(root.find('.ReportUserAbuse--report-sent')).toHaveLength(1);
+    expect(root.find('.ReportUserAbuse-form')).toHaveLength(0);
+    expect(root.find('.ReportUserAbuse-show-more')).toHaveLength(0);
+    expect(root.find('button.ReportUserAbuse-send-report')).toHaveLength(0);
+  });
+
+  it('renders an error if one exists', () => {
+    const errorHandler = createStubErrorHandler();
+    const { store } = dispatchClientMetadata();
+    store.dispatch(setError({
```
kumar303: It may be a little safer to use the `ErrorHandler` interface instead of calling the action directly.

```js
      id: 'some-error-handler-id',
      dispatch: store.dispatch,
    });
    errorHandler.handle(new Error('something bad'));
```

We should probably cleanup all the varying examples of how to do this.

src/amo/reducers/userAbuseReports.js
```diff
@@ -0,0 +1,235 @@
...
+
+type LoadUserAbuseReportParams = {|
+  message: string,
+  reporter: {|
+    id: number,
+    name: string,
+    url: string,
+    username: string,
+  |} | null,
+  user: UserType,
```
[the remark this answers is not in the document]
tofumatt: Actually I think I passed the actual user object into the reducer but I get how that's confusing. I forget, but yeah I'll tidy it up.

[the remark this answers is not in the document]
tofumatt: I'm just going to move to all ids to keep things simple. I guess if we had flow coverage everywhere we could ensure the full `UserType` was being passed around, which was my intent here, but failing that the ID prevents us having incomplete or confusing objects in the future.

tests/unit/amo/components/TestReportUserAbuse.js
tofumatt: Agreed. These tests are copypasta from the add-on user abuse reports, but I'll tidy at least these ones up.

tests/unit/amo/components/TestReportUserAbuse.js
tofumatt: Nice catch! I had to change the test to be `.toHaveText()` instead of `.toEqual()`, but that's nicer anyway. 👍

tests/unit/amo/components/TestReportUserAbuse.js
tofumatt: I think there was an issue with it not working, maybe because of shallow render? I'll try again and see...

tofumatt: Oh actually, it was just that I needed to reach into the `DismissibleTextForm` component to do the hideUI action in the next test... but I can probably tidy this up. I'll give it a go...

tests/unit/amo/components/TestReportUserAbuse.js
tofumatt: Ah, that's aces.

tofumatt:
I've made all the requested changes, should be set for another review.

tests/unit/amo/reducers/test_userAbuseReports.js
```diff
@@ -0,0 +1,170 @@
...
+          reporter: null,
+          user: createUserAccountResponse({ id: 50 }),
+        });
+        const secondReport = createFakeUserAbuseReport({
+          message: 'This user is boring',
+          reporter: null,
+          user: createUserAccountResponse({ id: 51 }),
+        });
+
+        store.dispatch(loadUserAbuseReport({
```
kumar303: Why does this use a store instead of calling the reducer directly? Nothing else in this suite does that. It will execute all other reducers on `dispatch()` which seems unnecessary.

I think I asked this in a previous review but I can't find the the thread anymore.

tests/unit/amo/sagas/testUserAbuseReports.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    // Report the same add-on again; this will cause the reducer to throw
+    // an error and the saga should dispatch an error.
+    _sendUserAbuseReport({
+      message: 'Duplicate!',
+      userId,
+    });
+
+    await sagaTester.waitFor(SET_ERROR);
```
kumar303: What throws an error? I don't see any duplicate checking code in https://github.com/mozilla/addons-frontend/pull/4474/files#diff-725633e03c598c1ecedc6e82646a2eedR189 Did you check that this test fails when you remove the duplicate checking code?

src/core/api/abuse.js
```diff
@@ -20,3 +19,38 @@ export function reportAddon(
...
+  user: {|
+    id: number,
+    name: string,
+    url: string,
+    username: string,
+  |},
+|};
+
+export function reportUser(
+  { api, message, user }: ReportUserParams
```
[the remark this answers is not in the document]
tofumatt: Ah, shoot. Right. Testing this means rebasing the user profile patch onto it which I've been trying to avoid doing, and the tests didn't catch this. Thanks.

tests/unit/amo/reducers/test_userAbuseReports.js
tofumatt: I think I just found it nicer than constantly returning the reducer's new state. But I can change it, no worries.

tofumatt:
Thanks for catching that `userId` mixup and the test for dupes which I shouldn't have done. I've added the flow types so this is ready for another review.

src/amo/sagas/userAbuseReports.js
```diff
@@ -0,0 +1,53 @@
...
+} from 'amo/reducers/userAbuseReports';
+import { reportUser as reportUserApi } from 'core/api/abuse';
+import log from 'core/logger';
+import { createErrorHandler, getState } from 'core/sagas/utils';
+import type { SendUserAbuseReportAction } from 'amo/reducers/userAbuseReports';
+import type { ReportUserParams } from 'core/api/abuse';
+
+
+export function* reportUser({
+  // TODO: Use a type for this.
```
kumar303: Looks like it's done!

src/amo/sagas/userAbuseReports.js
```diff
@@ -0,0 +1,53 @@
...
+  },
+}: SendUserAbuseReportAction): Generator<any, any, any> {
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  yield put(errorHandler.createClearingAction());
+
+  try {
+    const state = yield select(getState);
+
+    const params: ReportUserParams = { api: state.api, message, userId };
```
kumar303: Thanks. I think this is really helpful because you shouldn't have to do a big rebase and manually test something (as you pointed out) each time you make small structural changes. Sagas are a key part of our integration.

tests/unit/amo/reducers/test_userAbuseReports.js
```diff
@@ -0,0 +1,167 @@
...
+            message: abuseReport.message,
+            reporter: abuseReport.reporter,
+            userId: abuseReport.user.id,
+          }));
+
+        expect(state.byUserId[abuseReport.user.id].message)
+          .toEqual('I am Groot!');
+      });
+
+      it('requires a defined reporter', () => {
```
kumar303: Since you ported to `invariant`, you don't need this test. We can trust that `invariant()` works as advertised and I don't think it is helpful to add tests for static configuration.

tests/unit/amo/reducers/test_userAbuseReports.js
```diff
@@ -0,0 +1,167 @@
...
+        expect(() => {
+          const partialParams = abuseReportResponse();
+          delete partialParams.reporter;
+
+          loadUserAbuseReport(partialParams);
+        }).toThrow('reporter cannot be undefined');
+      });
+
+      it('allows reporter to be `null`', () => {
```
kumar303: I don't think this test is so useful either

PR #4483, opened 2018-03-02
chore(package): update stylelint-config-standard to version 18.2.0

tofumatt:
Closes #3976.
Closes #4684.


tofumatt:
The main issue here is `no-descending-specificity`. It's an interesting notion but right now I'd be happy to ignore it as it gets in the way a fair bit and the fixes/error messages don't seem very intuitive. Plus we tend to avoid most specificity/nesting bugs with our mostly-unnested, class-based selectors.

.stylelintrc
```diff
@@ -32,6 +32,7 @@
     "max-nesting-depth": [3, {
       "ignore": ["blockless-at-rules"],
     }],
+    "no-descending-specificity": null,
```
tofumatt: I ignored this because it's a tough one to solve and seems a bit unaware of our rules anyway. Getting certain files to pass didn't buy us much and our rules about not nesting/targeting elements directly means we don't actually get bit by the ramifications of *not* following this rule much anyway.

I think it's safe to ignore.

tofumatt:
I think this is fine but just want a review to make sure we're cool ignoring that new stylelint rule.

.stylelintrc
muffinresearch: As an aside, the output of this rule seems to show some pretty heavily specified selectors. I wonder if this indicates some problems we're getting via nesting?

I think the rule seems ok in the sense that it's trying to ensure a stylesheet is easier to grok. Whilst we are quite reliant on specificity it would be nice to clean up these cases at some time so if we side-step this now we could add a bug for a follow-up patch to fix it.

They sum up the rule's rationale quite well [here](https://stylelint.io/user-guide/rules/no-descending-specificity/):

>The clashes of these two mechanisms for prioritization, source order and specificity, can cause some confusion when reading stylesheets. If a selector with higher specificity comes before the selector it overrides, we have to think harder to understand it, because it violates the source order expectation. Stylesheets are most legible when overriding selectors always come after the selectors they override. That way both mechanisms, source order and specificity, work together nicely.

tofumatt: It makes a lot of sense, but it flagged what I considered a lot of false positives in our stylehsheets. I'd fix one issue and another would pop up. Maybe it couldn't make sense of our selectors... or maybe fixing that particular issue is just hairy enough that it needs to be its own code quality issue because it'd be a huge patch.

Even if we want to implement it (fine with me) I'd say making it its own issue would be a better move than blocking the upgrade.

bobsilverberg: It does sound like that's the way we need to go at this point. Just so I can understand, is `no-descending-specificity` a new rule that has been added with a recent release, or has the default value of that rule changed, or has the behaviour of that rule changed?

tofumatt: I think it's a new rule. I'm not sure though, the changelog would have it I think.

PR #4493, opened 2018-03-03
Fixes 4253: added hyphen when word breaks

### LINKED ISSUE #4253: Words are not wrapped correctly in add-on details page - statistics section
Prerequisites: Locale set to German. Italian, Spanish

STR:

1. Go to AMO homepage in responsive design mode (~750 - 1000 width) 
2. Select an Add-on to view its listing page (i.e. https://addons-dev.allizom.org/de/firefox/addon/awesome-screenshot-plus-/)
3. Observe the statistics section in the right corner

Expected result:
There are no layout issues

 Actual result:
 Text is not wrapped correctly

Notes:
- reproduces on AMO dev with FF58, Win10 and Android 7.0

image

tsl143:
Fixes #4253 

This puts hyphen at every word break, to make more sense with i18N.

Before:
screen shot 2018-03-03 at 9 51 42 pm

After:
screen shot 2018-03-03 at 9 49 50 pm

### DIFF
--- a/src/ui/components/MetadataCard/styles.scss
+++ b/src/ui/components/MetadataCard/styles.scss
@@ -19,6 +19,7 @@
   @include text-align-start();
 
   flex: 1;
+  hyphens: auto;
   margin: 12px;
   max-width: 25%;
 


tsl143:
This is iterated in #4495 , if we merge #4495, this can be closed without merge.

PR #4503, opened 2018-03-05
Remove the parsePage helper function

### LINKED ISSUE #4403: Remove the parsePage helper function
[`paresePage`](https://github.com/mozilla/addons-frontend/blob/23f68b9d4effd2f71c0f5d84ae8ea0957eeb4f84/src/core/utils/index.js#L286-L289) in `core/utils` accepts a `page` parameter and converts it to a `1` if it is invalid. We were using this incorrectly in `AddonReviewList` which resulted in issue #4081. It is unlikely that there are any situations in which we actually want to do this, so we should take a look at all the places where `parsePage` is still used and see if we can eliminate its usage and remove the function entirely.

bobsilverberg:
Fixes #4403 

We do not need the functionality of parsePage on the frontend because it is fine to pass `undefined` for `page` into API calls, and in fact we want to do that so the API will produce a 404 for us.

### DIFF
--- a/src/amo/components/Search/index.js
+++ b/src/amo/components/Search/index.js
@@ -29,7 +29,6 @@ import {
   convertFiltersToQueryParams,
   hasSearchFilters,
 } from 'core/searchUtils';
-import { parsePage } from 'core/utils';
 
 import './styles.scss';
 
@@ -191,8 +190,6 @@ export class SearchBase extends React.Component {
       }
     }
 
-    const page = parsePage(filters.page);
-
     // We allow specific paginationQueryParams instead of always using
     // convertFiltersToQueryParams(filters) so certain search filters
     // aren't repeated if they are elsewhere in the URL. This is useful
@@ -209,7 +206,7 @@ export class SearchBase extends React.Component {
       <Paginate
         LinkComponent={LinkComponent}
         count={count}
-        currentPage={page}
+        currentPage={filters.page}
         pathname={pathname}
         queryParams={queryParams}
       />
@@ -254,7 +251,7 @@ export function mapStateToProps(state) {
 // This ID does not need to differentiate between component instances because
 // the error handler gets cleared every time the search filters change.
 export const extractId = (ownProps) => {
-  return parsePage(ownProps.filters.page);
+  return ownProps.filters.page;
 };
 
 export default compose(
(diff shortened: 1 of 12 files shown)


tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -668,7 +668,7 @@ describe(__filename, () => {
         location: fakeRouterLocation(),
       });
 
-      expect(extractId(ownProps)).toEqual(`foobar-1`);
+      expect(extractId(ownProps)).toEqual(`foobar-`);
```
tofumatt: That's kind of weird. Should a null value default to 1 instead of blank? Or maybe the string should be something other than blank? `noPageSet` or something? I dunno, maybe it doesn't matter 🤷‍♂️

src/amo/components/Search/index.js
```diff
@@ -254,7 +251,7 @@ export function mapStateToProps(state) {
 // This ID does not need to differentiate between component instances because
 // the error handler gets cleared every time the search filters change.
 export const extractId = (ownProps) => {
-  return parsePage(ownProps.filters.page);
+  return ownProps.filters.page;
```
tofumatt: Should this maybe return a more useful default than `undefined`?

tests/unit/amo/sagas/test_collections.js
```diff
@@ -34,6 +33,7 @@ import {
 describe(__filename, () => {
   const user = 'user-id-or-name';
   const slug = 'collection-slug';
+  const pageToFetch = 1;
```
tofumatt: Could this just `page`? Then you could use the key/variable name shortcut below:

```js
.withArgs({
   api: state.api,
   page,
   slug,
   user,
})
```

src/amo/components/Search/index.js
bobsilverberg: I discussed this issue with @kumar303 and I believe we decided that just returning undefined in all cases makes sense, because that's what's going to be passed to the API with this change.

tests/unit/amo/components/TestAddonReviewList.js
bobsilverberg: I agree it looks weird, but there doesn't seem to be an actual reason to change it, as all we're doing is creating an `extractId` which nobody will ever see.

tests/unit/amo/sagas/test_collections.js
bobsilverberg: Good call. :-)

tests/unit/amo/components/TestAddonReviewList.js
tofumatt: Fair enough 😄

PR #4509, opened 2018-03-06
Fix Paginate so it does not require currentPage

### LINKED ISSUE #4507: AMO is returning a server error after any search related action
STR:

1. Go to AMO-dev homepage and perform a search

Actual result:
A server error is received after any search action

Expected results:
Search results are listed

Notes:
- reproduces also  for 'See more" links and  Collections listed on homepage

server error

bobsilverberg:
Fixes #4507

### DIFF
--- a/src/core/components/Paginate/index.js
+++ b/src/core/components/Paginate/index.js
@@ -20,7 +20,7 @@ export class PaginateBase extends React.Component {
   static propTypes = {
     LinkComponent: PropTypes.func,
     count: PropTypes.number.isRequired,
-    currentPage: PropTypes.number.isRequired,
+    currentPage: PropTypes.number,
     i18n: PropTypes.object.isRequired,
     pathname: PropTypes.string.isRequired,
     perPage: PropTypes.number,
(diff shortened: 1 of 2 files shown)


src/core/components/Paginate/index.js
```diff
@@ -74,9 +75,6 @@ export class PaginateBase extends React.Component {
     if (count === undefined) {
       throw new Error('The count property cannot be undefined');
     }
-    if (currentPage === undefined) {
```
tofumatt: Shouldn't the `defaultProps` setting mean this doesn't happen? I forget how `defaultProps` work exactly...

bobsilverberg: Correct, which is why I removed the code. `currentPage` will never be undefined, so this `if` statement is superfluous.

tofumatt: I'm wondering what would happen if currentPage was null or 0 due to an error in a parent component. Would be nice to have an error thrown if that will break things.

bobsilverberg: Ah, good point. I'll check that and if it's an issue I'll issue a fix.

bobsilverberg: Guess what? I totally broke pagination with this fix! I'll open a new issue and hopefully it's something I can fix fairly quickly.

tofumatt: That's weird, I got pagination working locally when I tried this...

PR #4518, opened 2018-03-07
15 March homepage content programing update

### LINKED ISSUE #4515: 15 March homepage content programing update
For the Thursday, 15 March push, please update the AMO homepage programming in accord with what's [here](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#heading=h.2dl5oxy3ewcc).

@muffinresearch @diox @EnTeQuAk

bobsilverberg:
Fixes #4515 

I created the collections required for the home page on dev to make testing this possible. I guess we'll have to create them on stage as well once this gets pushed there. Is that something QA could do to support testing, or is it something we're required to do?

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -26,9 +26,9 @@ import './styles.scss';
 
 
 export const COLLECTIONS_TO_FETCH = [
-  { slug: 're-imagine-search', user: 'mozilla' },
-  { slug: 'privacy-matters', user: 'mozilla' },
-  { slug: 'dynamic-media-downloaders', user: 'mozilla' },
+  { slug: 'good-time-tabs', user: 'mozilla' },
+  { slug: 'wikipedia-boosters', user: 'mozilla' },
+  { slug: 'trending-add-ons', user: 'mozilla' },
 ];
 
 export class HomeBase extends React.Component {
@@ -214,8 +214,8 @@ export class HomeBase extends React.Component {
         <LandingAddonsCard
           addons={collections[0]}
           className="Home-FeaturedCollection"
-          header={i18n.gettext('Re-imagine search')}
-          footerText={i18n.gettext('See more search extensions')}
+          header={i18n.gettext('Good time tabs')}
+          footerText={i18n.gettext('See more good time tabs extensions')}
           footerLink={
             `/collections/${COLLECTIONS_TO_FETCH[0].user}/${COLLECTIONS_TO_FETCH[0].slug}/`
           }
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Oops, I need to update the test for the home page for this, which I forgot. Adding that now...

src/amo/components/Home/index.js
```diff
@@ -252,8 +252,8 @@ export class HomeBase extends React.Component {
         <LandingAddonsCard
           addons={collections[2]}
           className="Home-FeaturedCollection"
-          header={i18n.gettext('Dynamic media downloaders')}
-          footerText={i18n.gettext('See more dynamic media downloaders')}
+          header={i18n.gettext('Trending extensions')}
+          footerText={i18n.gettext('See more trending extensions')}
           footerLink={
             `/collections/${COLLECTIONS_TO_FETCH[2].user}/${COLLECTIONS_TO_FETCH[2].slug}/`
```
tofumatt: Should trending extensions be a collection? I guess the programming update would say but I expected this to link to `/search/?sort=hotness&type=extension`

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -22,57 +22,58 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
-          linkTo="/addon/momentumdash/"
+          linkTo="/addon/autoformer/"
         >
-          <h3>{i18n.gettext('Momentum')}</h3>
+          <h3>{i18n.gettext('AutoFormer+')}</h3>
 
           <p>
-            {i18n.gettext(`Replace your new tab with a personal
-              dashboard—to-do lists, weather forecasts and more`)}
+            {i18n.gettext(`Fill out a lot of forms? This might help`)}
```
tofumatt: This doesn't need to be a backtick-quoted string anymore, surprised eslint doesn't complain! I guess it's fine though.

src/amo/components/Home/index.js
bobsilverberg: Good question. @devaneymoz, do you intend for `Trending extensions` to be populated via a collection, or can we just use the query as suggested above, an example of which would be https://addons-dev.allizom.org/en-US/firefox/search/?sort=hotness&type=extension ?

src/amo/components/HomeHeroBanner/index.js
bobsilverberg: I don't think eslint cares about this, but good point. It must have been a longer string before and I neglected to replace the backticks with single quotes. I'll change it.

devaneymoz:
@bobsilverberg in this case the "Trending" collection is a hand-picked collection of emerging content.

tofumatt:
That's a bit confusing as we also have a "trending" search filter. 😄

bobsilverberg:
Thanks @devaneymoz. I agree with @tofumatt but whatever. You're the boss!

tofumatt:
All good then!

devaneymoz:
You guys make a good point about the 'trending' distinction between algorithm and curated! Let's please keep these changes for now but I'll work on finding a better editorial label to distinguish the two cases. 

Man I forgot what great hair Tony Danza has.

PR #4524, opened 2018-03-08
Expose payment flag on public listing page

### LINKED ISSUE #2080: Expose payment flag on public listing page
Following https://github.com/mozilla/addons-server/issues/4954, we need to expose the payment requirement flag on the public listing page.

Our official policy would be like:
A payment is required to use this add-on or the service this add-on relies on, either upfront, after a trial period or in conjunction with a paid software.

This sounds a bit too wordy to me. I think we could simply and go by:
A payment is required to use this add-on or the service it relies on.

//cc @jvillalobos
//cc @pwalm for UX (on the new desktop pages)
Should we expose it on mobile as well?

bobsilverberg:
Fixes #2080

### DIFF
--- a/src/amo/components/AddonBadges/index.js
+++ b/src/amo/components/AddonBadges/index.js
@@ -64,6 +64,12 @@ export const AddonBadgesBase = (props: Props) => {
           label={i18n.gettext('Not compatible with Firefox Quantum')}
         />
       ) : null}
+      {addon.requires_payment ? (
+        <Badge
+          type="requires-payment"
+          label={i18n.gettext('Some features may require payment')}
+        />
+      ) : null}
     </div>
   );
 };
--- a/tests/unit/amo/components/TestAddonBadges.js
+++ b/tests/unit/amo/components/TestAddonBadges.js
@@ -31,6 +31,16 @@ describe(__filename, () => {
     expect(root.html()).toEqual(null);
   });
 
+  it('displays no badges when none are called for', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      type: ADDON_TYPE_EXTENSION,
+    });
+    const root = shallowRender({ addon });
+
+    expect(root.find(Badge)).toHaveLength(0);
+  });
+
   it('displays a badge when the addon is featured', () => {
     const addon = createInternalAddon({
       ...fakeAddon,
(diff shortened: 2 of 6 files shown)


src/amo/components/AddonBadges/index.js
```diff
@@ -64,6 +64,12 @@ export const AddonBadgesBase = (props: Props) => {
           label={i18n.gettext('Not compatible with Firefox Quantum')}
         />
       ) : null}
+      {addon.requires_payment ? (
+        <Badge
+          type="requires-payment"
+          label={i18n.gettext('Some features may require payment')}
```
tofumatt: If the add-on has "requires payment" flagged, we shouldn't be so wishy-washy about "may require" 😄

I think this should just say "Some features require payment". That's more accurate and less confusing. Any issues with that, @jvillalobos?

tests/unit/amo/components/TestAddonBadges.js
```diff
@@ -181,4 +181,24 @@ describe(__filename, () => {
...
+      .toHaveProp('label', 'Some features may require payment');
+  });
+
+  it('does not display a badge when the addon does not require payment', () => {
+    const addon = createInternalAddon(createFakeAddon({
+      requires_payment: false,
+    }));
+    const root = shallowRender({ addon });
+
+    expect(root.find(Badge)).toHaveLength(0);
```
tofumatt: This is fine (and probably mirrors other tests), but technically we just want no badges with the "requires payment" prop.

This might be a bit more accurate as `expect(root.find(Badge).find({ type: 'requires-payment' })).toHaveLength(0);`, but obviously it works as-is. Just means if we added a `Badge` elsewhere we'd need to change this test, making it less atomic 😄

src/amo/components/AddonBadges/index.js
bobsilverberg: This was discussed quite a bit in the issue, and I thought the final word was at https://github.com/mozilla/addons-frontend/issues/2080#issuecomment-366061040, but I guess it doesn't hurt to double check before landing. 😃

jvillalobos: The flag can mean a number of different things (see the expanded label on dev tools), so I think it's okay to keep the slightly vague wording.

tofumatt: Ten four 👍

tests/unit/amo/components/TestAddonBadges.js
bobsilverberg: This is a good point. I changed this test, and the others that are asserting the absence of a particular badge. I also added a test that checks that no badges at all are present when none are called for. This would break if we did add a different `Badge` element somewhere inside the `Addon`, but if we do that we'd probably need to revisit this test anyway.

Perhaps you want to take a final look at this before I merge it, @tofumatt?

bobsilverberg:
Before:

screenshot 2018-03-09 08 15 40

After:

screenshot 2018-03-09 08 16 26

Issue #4510, opened 2018-03-06
Extract AutoSearchInput component


kumar303:
We have an input field that autosuggests search results when you type in it. I need to use this for putting add-ons in a collection (https://github.com/mozilla/addons-frontend/issues/4004) so I will extract it into a shared component.

For QA: Please test for regressions in all features related to the search bar in the header.

ioanarusiczki:
I tested the search feature using the test cases from[ here](https://testrail.stage.mozaws.net/index.php?/suites/view/97&group_by=cases:section_id&group_order=asc) with FF58 (Win10 and Android 8.0) 
It seems to be alright so I mark it as verified fixed.

kumar303:
@ioanarusiczki the issue had not been closed yet. I just landed the patch -- can you verify it again when the patch lands on dev? It could take up to 15 minutes to land but you can check the commit at https://addons-dev.allizom.org/__frontend_version__

ioanarusiczki:
Ups! The issue was verified today by @vcarciu on AMO stage with FF59 (Win10).  I re-checked for android 8.0.
The test results are[ here](https://testrail.stage.mozaws.net/index.php?/runs/view/8825&group_by=cases:section_id&group_order=asc)

PR #4532, opened 2018-03-11
Fixes #3070 - Only initiate autocomplete when 𝑛 characters are present

### LINKED ISSUE #3070: Start autocomplete search after n chars entered
The autocomplete should only start sending requests once there's more than `n` characters because the results for a single char aren't that useful.

In the past typical values for `n` would be 2-3 chars - it'll make sense to make this configurable so it's easily tweaked. Let's start with 2 chars and go from there.

SeanPrashad:
Fixes #3070 - Only initiate autocomplete for the search bar after `𝑛` characters have been entered. In this case, `𝑛` should be a configurable integer value.

#### Game Plan:
- [x] Hard-code functionality to filter only with 2 or more characters present
- [x] Remove hard-coded value and allow configurability on what `𝑛` may be (~_possibly pass in another value to the respective function?_~ create a constant variable to be used globally => **check with devs**)
- [x] Clear the results if under minimum threshold
- [x] Review respective unit tests and update/create new ones as necessary

### DIFF
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -165,7 +165,7 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
 
     if (value.length < SEARCH_TERM_MIN_LENGTH) {
       log.debug(oneLine`Ignoring suggestions fetch because query
-      subceeds min length (${SEARCH_TERM_MIN_LENGTH})`);
+      does not meet the required length (${SEARCH_TERM_MIN_LENGTH})`);
 
       this.props.dispatch(autocompleteCancel());
       return;
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -155,7 +155,7 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
   }
 
   handleSuggestionsFetchRequested = (
-    { value, reason }: OnSuggestionsFetchRequestedParams
+    { value }: OnSuggestionsFetchRequestedParams
   ) => {
     if (!value) {
       log.debug(oneLine`Ignoring suggestions fetch requested because
@@ -183,11 +183,6 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
     const filters = this.createFiltersFromQuery(value);
 
     this.setState({ autocompleteIsOpen: true });
-
-    if (reason === 'input-focused') {
-      log.debug('Ignoring suggestions fetch on search input focus');
-      return;
-    }
 
     this.dispatchAutocompleteStart({ filters });
   }
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -31,6 +31,7 @@ import type { ErrorHandlerType } from 'core/errorHandler';
 
 import './styles.scss';
 
+export const SEARCH_TERM_MIN_LENGTH = 3;
 export const SEARCH_TERM_MAX_LENGTH = 100;
 
 // TODO: port reducers/autocomplete.js to Flow and move this there.
@@ -162,10 +163,20 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
       return;
     }
 
+    if (value.length < SEARCH_TERM_MIN_LENGTH) {
+      log.debug(oneLine`Ignoring suggestions fetch because query
+      subceeds min length (${SEARCH_TERM_MIN_LENGTH})`);
+
+      this.props.dispatch(autocompleteCancel());
+      return;
+    }
+
     if (value.length > SEARCH_TERM_MAX_LENGTH) {
       log.debug(oneLine`Ignoring suggestions fetch because query
         exceeds max length (${SEARCH_TERM_MAX_LENGTH})`
       );
+
+      this.props.dispatch(autocompleteCancel());
       return;
     }
 
@@ -271,6 +282,7 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
 
     const inputProps = {
       className: 'AutoSearchInput-query',
+      minLength: SEARCH_TERM_MIN_LENGTH,
       maxLength: SEARCH_TERM_MAX_LENGTH,
       name: inputName,
       onChange: this.handleSearchChange,
--- a/tests/unit/amo/components/TestAutoSearchInput.js
+++ b/tests/unit/amo/components/TestAutoSearchInput.js
@@ -316,7 +316,7 @@ describe(__filename, () => {
         root, query: 't'.repeat(SEARCH_TERM_MAX_LENGTH + 1),
       });
 
-      sinon.assert.notCalled(dispatchSpy);
+      sinon.assert.called(dispatchSpy);
     });
 
     it('does not fetch suggestions on focus', () => {
--- a/tests/unit/amo/components/TestAutoSearchInput.js
+++ b/tests/unit/amo/components/TestAutoSearchInput.js
@@ -3,7 +3,7 @@ import * as React from 'react';
 import Autosuggest from 'react-autosuggest';
 
 import AutoSearchInput, {
-  extractId, AutoSearchInputBase, SEARCH_TERM_MAX_LENGTH,
+  extractId, AutoSearchInputBase, SEARCH_TERM_MIN_LENGTH, SEARCH_TERM_MAX_LENGTH,
 } from 'amo/components/AutoSearchInput';
 import SearchSuggestion from 'amo/components/SearchSuggestion';
 import {
@@ -305,6 +305,18 @@ describe(__filename, () => {
       fetchSuggestions({ root, query: '' });
 
       sinon.assert.notCalled(dispatchSpy);
+    });
+
+    it('does not fetch suggestions for a really short value', () => {
+      const { store } = dispatchClientMetadata();
+      const dispatchSpy = sinon.stub(store, 'dispatch');
+      const root = render({ store });
+
+      fetchSuggestions({
+        root, query: 't'.repeat(SEARCH_TERM_MIN_LENGTH - 1),
+      });
+
+      sinon.assert.called(dispatchSpy);
     });
 
     it('does not fetch suggestions for a really long value', () => {
--- a/tests/unit/amo/components/TestAutoSearchInput.js
+++ b/tests/unit/amo/components/TestAutoSearchInput.js
@@ -316,7 +316,8 @@ describe(__filename, () => {
         root, query: 't'.repeat(SEARCH_TERM_MIN_LENGTH - 1),
       });
 
-      sinon.assert.called(dispatchSpy);
+      sinon.assert.calledWith(dispatchSpy, autocompleteCancel());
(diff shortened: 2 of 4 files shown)


SeanPrashad:
I've applied a fix that checks if the user's entered value is greater than `SEARCH_TERM_MIN_LENGTH` (_currently configured to 2_) however previously valid search contents still appears:

wip_search_fix

Any thoughts on this?

willdurand:
> Any thoughts on this?

Maybe you should reset the result set when `value.length < SEARCH_TERM_MIN_LENGTH`

tofumatt:
As @willdurand suggested, I think removing the results after you're under the minimum is best.

src/amo/components/SearchForm/index.js
```diff
@@ -27,6 +27,7 @@ import Icon from 'ui/components/Icon';
 
 import './styles.scss';
 
+export const SEARCH_TERM_MIN_LENGTH = 2;
```
tofumatt: If we already have a constant for max length (which I didn't even know about/remember!) I think it's fine to have one for minimum length too. I guess it's more extensible if they're props with defaults set using `static defaultProps = { autocompleteMinLength: 3 };`

But it's not a huge deal either way.

tofumatt: This should be set to `3` because as-is only a `1` character search is too short, but I think we wanted a three character minimum.

src/amo/components/SearchForm/index.js
```diff
@@ -143,7 +144,10 @@ export class SearchFormBase extends React.Component {
       return;
     }
 
-    if (value.length > SEARCH_TERM_MAX_LENGTH) {
+    if (value.length < SEARCH_TERM_MIN_LENGTH) {
+      log.debug(oneLine`SEARCH_TERM_MIN_LENGTH: ${SEARCH_TERM_MIN_LENGTH} shortage`);
```
tofumatt: I don't think we need to log anything here, but if you wanted to this could be:

`log.debug(oneLine`Search term length is too short to dispatch autocomplete action. Minimum length is: ${SEARCH_TERM_MIN_LENGTH}`);`

SeanPrashad:
Thanks to the both of you for your speedy guidance @willdurand & @tofumatt!

I've gone ahead and slapped in the requested changes - only one thing left (_hopefully_): fixing up the tests!

SeanPrashad:
Tests failing:

1. https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestSearchForm.js#L236

2. https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestSearchForm.js#L273

3. https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestSearchForm.js#L321

kumar303:
Hi @SeanPrashad, thanks for your interest in this feature. I would advise that you put this work on pause temporarily until after https://github.com/mozilla/addons-frontend/issues/4510 has landed because your work will conflict with that patch. Sorry, I guess the timing was just coincidental -- the other patch should land in the next couple of days.

SeanPrashad:
Will do @kumar303 - I've subscribed to the other issue and will be keeping an eye out for when that patch lands.

src/amo/components/SearchForm/index.js
```diff
@@ -143,7 +144,12 @@ export class SearchFormBase extends React.Component {
       return;
     }
 
-    if (value.length > SEARCH_TERM_MAX_LENGTH) {
+    if (value.length < SEARCH_TERM_MIN_LENGTH) {
+      log.debug(oneLine`Search term length is too short to dispatch autocomplete action. Minimum length is: ${SEARCH_TERM_MIN_LENGTH}`);
+      // Reset the result set to prevent rendering a previously valid result set
+      this.props.suggestions.length = 0;
```
tofumatt: Mutating the `props` object should be avoided (React props are immutable), so it would be best to handle this another way.

SeanPrashad:
@kumar303 @tofumatt Hey guys - I saw that #4510 was closed earlier today - do I have the green lights to continue? It looks like I'll need to rebase my changes if so.

tofumatt:
@SeanPrashad I think you can rebase and continue, yes.

SeanPrashad:
Substituting `//this.props.suggestions.length = 0;` with `this.props.dispatch(autocompleteCancel());` results in the following behaviour:

kumar_suggestion

Currently I'm looking into a way to close the results pane!

kumar303:
> I've applied a patch that checks if the user's entered value is greater than SEARCH_TERM_MIN_LENGTH (currently configured to 2) however previously valid search content still appears

We do have an action that you can dispatch for clearing search results. You can see it used in [`handleSuggestionsClearRequested()`](https://github.com/mozilla/addons-frontend/blob/dd0afb33cac75326dd0cceb8a19ec5bf9b85bd67/src/amo/components/AutoSearchInput/index.js#L153).

This action [reduces state](https://github.com/mozilla/addons-frontend/blob/dd0afb33cac75326dd0cceb8a19ec5bf9b85bd67/src/core/reducers/autocomplete.js#L55) so that the loading flag is removed which prevents the UI from getting stuck in a loading state.

However, I think the reducer has a bug in it. It should probably also be resetting the suggestions array to clear the results, like:

```diff
diff --git a/src/core/reducers/autocomplete.js b/src/core/reducers/autocomplete.js
index 3f4460a81..157817f02 100644
--- a/src/core/reducers/autocomplete.js
+++ b/src/core/reducers/autocomplete.js
@@ -56,6 +56,7 @@ export default function reducer(state = initialState, action = {}) {
       return {
         ...state,
         loading: false,
+        suggestions: [],
       };
     case AUTOCOMPLETE_STARTED:
       return {

```

After that, you can probably dispatch `autocompleteCancel()` when the search term is too long or too short and it should clear any previously fetched results.

However, I'm not totally sure if this will do the right thing in all cases. I wanted to point it out because this bug might have been hard to spot if you're not yet familiar with how actions and reducers work.

SeanPrashad:
After implementing @kumar303's suggestions, I noticed that if the user quickly deletes a certain amount of characters, and that amount is below our threshold of 3, the suggestions list still renders.

I was thinking it might be related to using the [default value for Lodash.debounce](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AutoSearchInput/index.js#L6) and might be something worth opening another issue for - see GIF below:

post_kumar_suggestion

kumar303:
That sounds like acceptable behavior. In the real world, someone will be typing an actual word and waiting for results. Limiting the number of characters before searching is just an internal optimization to put less stress on our API endpoint.

tofumatt:
It’s also to prevent unhelpful results. Searching for just “b” tends not to return relevant/useful results.

SeanPrashad:
Understood - I'll move onto fixing the two failing tests found [here](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestAutoSearchInput.js#L310) and [here](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/core/reducers/test_autocomplete.js#L22).

SeanPrashad:
Ok - Ready for another review @tofumatt @kumar303!

See below for a summary of changes/updates:

1. If the user enters a search term below our threshold of **3** characters, the API will not query for the respective suggestion list
    * Also, if the user previously entered a valid search term and the number of characters subceeded the minimum character threshold, the previous suggestion list will be cleared (_required to update reducer as per comment found [here](https://github.com/mozilla/addons-frontend/pull/4532#issuecomment-375693343)_)

2. Update existing unit tests to take into account calling the reducer and the new suggestion list behaviour

3. Added a [new unit test](https://github.com/mozilla/addons-frontend/pull/4532/files#diff-24e7edfea4e8306e6fa15c7f3ede0477R310) to verify the behaviour of searching below our threshold of **3** characters

src/amo/components/AutoSearchInput/index.js
```diff
@@ -162,10 +163,20 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
       return;
     }
 
+    if (value.length < SEARCH_TERM_MIN_LENGTH) {
+      log.debug(oneLine`Ignoring suggestions fetch because query
+      subceeds min length (${SEARCH_TERM_MIN_LENGTH})`);
```
kumar303: I know that subceed is sometimes used colloquially but it threw me off at first. I think it would be better to spell it out as "Ignoring ... because query does not meet the required length (${SEARCH_TERM_MIN_LENGTH})"

tests/unit/amo/components/TestAutoSearchInput.js
```diff
@@ -316,7 +328,7 @@ describe(__filename, () => {
         root, query: 't'.repeat(SEARCH_TERM_MAX_LENGTH + 1),
       });
 
-      sinon.assert.notCalled(dispatchSpy);
+      sinon.assert.called(dispatchSpy);
```
kumar303: Since multiple actions could be dispatched here, you need to make sure it's the right one:

```js
sinon.assert.calledWith(dispatchSpy, autocompleteCancel());
```

You should also add this to make sure your `return` statement is in the right place:

```js
sinon.assert.callCount(dispatchSpy, 1);
```

src/amo/components/AutoSearchInput/index.js
```diff
@@ -162,10 +163,20 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
...
+      this.props.dispatch(autocompleteCancel());
+      return;
+    }
+
     if (value.length > SEARCH_TERM_MAX_LENGTH) {
       log.debug(oneLine`Ignoring suggestions fetch because query
         exceeds max length (${SEARCH_TERM_MAX_LENGTH})`
       );
+
+      this.props.dispatch(autocompleteCancel());
```
kumar303: The [test for long queries](https://github.com/SeanPrashad/addons-frontend/blob/885452970d79f31ab8793bd6b62f51e21dbcbe4c/tests/unit/amo/components/TestAutoSearchInput.js#L322) also needs to make sure that the cancel action was dispatched.

SeanPrashad:
I've updated the tests as per your comments @kumar303! Let me know if there's anything else I can help with ✌🏽

SeanPrashad:
Nice catch - I've removed [that block of code](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AutoSearchInput/index.js#L176) and it's respective test found [here](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestAutoSearchInput.js#L322).

New behaviour after local testing:

new_behaviour

src/amo/components/AutoSearchInput/index.js
```diff
@@ -154,30 +155,35 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
...
       return;
     }
 
     const filters = this.createFiltersFromQuery(value);
 
     this.setState({ autocompleteIsOpen: true });
 
-    if (reason === 'input-focused') {
```
kumar303: Yep, I think this is the right thing to do given that we fixed the behavior of `handleSuggestionsClearRequested()`. @willdurand maybe you remember why we originally needed this? We can always put it back with a follow-up patch.

PR #4534, opened 2018-03-12
Saga helpers for creating a collection

### LINKED ISSUE #4505: Saga helpers for creating a collection
We need saga helpers to facilitate creating a collection #4003 

For QA: please check for regressions in any of the existing collection pages and in particular with the "Edit a collection" feature.

bobsilverberg:
Fixes #4505

This adds code to reducers and sagas for creating a collection, as well as refactoring the code for updating a collection so that create/update can share some code.

It also removes `collectionUpdates` from state as it is never used.

### DIFF
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -80,6 +80,9 @@ export type CollectionsState = {
       |};
     },
   },
+  collectionUpdates: {
+    [collectionSlug: string]: {| updating: boolean, successful?: boolean |},
+  },
 };
 
 export const initialState: CollectionsState = {
@@ -88,6 +91,7 @@ export const initialState: CollectionsState = {
   current: { id: null, loading: false },
   userCollections: {},
   addonInCollections: {},
+  collectionUpdates: {},
 };
 
 type FetchCurrentCollectionParams = {|
@@ -449,7 +453,6 @@ export const addAddonToCollection = ({
 
 export type RequiredModifyCollectionParams = {|
   errorHandlerId: string,
-  formOverlayId: string,
   user: string,
 |};
 
@@ -485,11 +488,9 @@ export type UpdateCollectionAction = {|
 
 export const validateRequiredCollectionParams = ({
   errorHandlerId,
-  formOverlayId,
   user,
 }: RequiredModifyCollectionParams) => {
   invariant(errorHandlerId, 'errorHandlerId is required');
-  invariant(formOverlayId, 'formOverlayId is required');
   invariant(user, 'user is required');
 };
 
@@ -497,14 +498,12 @@ export const createCollection = ({
   errorHandlerId,
   defaultLocale,
   description,
-  formOverlayId,
   name,
   slug,
   user,
 }: CreateCollectionParams = {}): CreateCollectionAction => {
   validateRequiredCollectionParams({
     errorHandlerId,
-    formOverlayId,
     user,
   });
 
@@ -517,7 +516,6 @@ export const createCollection = ({
       errorHandlerId,
       defaultLocale,
       description,
-      formOverlayId,
       name,
       slug,
       user,
@@ -925,6 +923,20 @@ const reducer = (
         state,
         loading: false,
       });
+    }
+
+    case UPDATE_COLLECTION: {
+      const { collectionSlug } = action.payload;
+
+      return {
+        ...state,
+        collectionUpdates: {
+          [collectionSlug]: {
+            ...state.collectionUpdates[collectionSlug],
+            updating: true,
+          },
+        },
+      };
     }
 
     case DELETE_COLLECTION_BY_SLUG: {
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -1,5 +1,6 @@
 /* @flow */
 import { oneLine } from 'common-tags';
+import invariant from 'invariant';
 
 import { createInternalAddon } from 'core/reducers/addons';
 import type { AddonType, ExternalAddonType } from 'core/types/addons';
@@ -32,6 +33,7 @@ export const LOAD_COLLECTION_ADDONS: 'LOAD_COLLECTION_ADDONS'
 export const UPDATE_COLLECTION: 'UPDATE_COLLECTION' = 'UPDATE_COLLECTION';
 export const DELETE_COLLECTION_BY_SLUG: 'DELETE_COLLECTION_BY_SLUG'
   = 'DELETE_COLLECTION_BY_SLUG';
+export const CREATE_COLLECTION: 'CREATE_COLLECTION' = 'CREATE_COLLECTION';
 
 export type CollectionType = {
   addons: Array<AddonType> | null,
@@ -78,9 +80,6 @@ export type CollectionsState = {
       |};
     },
   },
-  collectionUpdates: {
-    [collectionSlug: string]: {| updating: boolean, successful?: boolean |},
-  },
 };
 
 export const initialState: CollectionsState = {
@@ -89,7 +88,6 @@ export const initialState: CollectionsState = {
   current: { id: null, loading: false },
   userCollections: {},
   addonInCollections: {},
-  collectionUpdates: {},
 };
 
 type FetchCurrentCollectionParams = {|
@@ -449,20 +447,83 @@ export const addAddonToCollection = ({
   };
 };
 
-type UpdateCollectionParams = {|
+type RequiredModifyCollectionParams = {|
   errorHandlerId: string,
-  collectionSlug: string,
+  formOverlayId: string,
+  user: string,
+|};
+
+type OptionalModifyCollectionParams = {|
   defaultLocale: ?string,
   description: ?LocalizedString,
+|};
+
+type CreateCollectionParams = {|
+  ...RequiredModifyCollectionParams,
+  ...OptionalModifyCollectionParams,
+  name: LocalizedString,
+  slug: string,
+|};
+
+type UpdateCollectionParams = {|
+  ...RequiredModifyCollectionParams,
+  ...OptionalModifyCollectionParams,
+  collectionSlug: string,
   name: ?LocalizedString,
   slug: ?string,
-  user: string,
+|};
+
+export type CreateCollectionAction = {|
+  type: typeof CREATE_COLLECTION,
+  payload: CreateCollectionParams,
 |};
 
 export type UpdateCollectionAction = {|
   type: typeof UPDATE_COLLECTION,
   payload: UpdateCollectionParams,
 |};
+
+export const validateRequiredCollectionParams = ({
+  errorHandlerId,
+  formOverlayId,
+  user,
+}: RequiredModifyCollectionParams) => {
+  invariant(errorHandlerId, 'errorHandlerId is required');
+  invariant(formOverlayId, 'formOverlayId is required');
+  invariant(user, 'user is required');
+};
+
+export const createCollection = ({
+  errorHandlerId,
+  defaultLocale,
+  description,
+  formOverlayId,
+  name,
+  slug,
+  user,
+}: CreateCollectionParams = {}): CreateCollectionAction => {
+  validateRequiredCollectionParams({
+    errorHandlerId,
+    formOverlayId,
+    user,
+  });
+
+  invariant(name, 'name is required when creating');
+  invariant(slug, 'slug is required when creating');
+
+  return {
+    type: CREATE_COLLECTION,
+    payload: {
+      errorHandlerId,
+      defaultLocale,
+      description,
+      formOverlayId,
+      name,
+      slug,
+      user,
+    },
+  };
+};
 
 export const updateCollection = ({
   errorHandlerId,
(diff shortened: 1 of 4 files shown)


src/amo/sagas/collections.js
```diff
@@ -166,43 +169,56 @@ export function* addAddonToCollection({
   }
 }
 
-export function* updateCollection({
-  payload: {
-    errorHandlerId,
+export function* modifyCollection(
+  action: 'create' | 'update',
+  params: Object,
```
bobsilverberg: I know this should be something more specific than `Object`, but I couldn't figure out what to make it so that Flow was happy, so some advice would be welcome.

src/amo/sagas/collections.js
```diff
@@ -213,25 +229,77 @@ export function* updateCollection({
       `/${lang}/${clientApp}/collections/${user}/${effectiveSlug}/`
     ));
 
+    const slugWasEdited = slug && slug !== collectionSlug;
```
bobsilverberg: I know we don't really need this line, but trying to inline it into the `if` statement below made the code a lot less understandable, so I kept this.

src/amo/sagas/collections.js
```diff
@@ -213,25 +229,77 @@ export function* updateCollection({
...
+    errorHandlerId,
+    defaultLocale,
+    description,
+    formOverlayId,
+    name,
+    slug,
+    user,
+  },
+}: CreateCollectionAction,
+_modifyCollection: typeof modifyCollection = modifyCollection,
```
bobsilverberg: I designed `createCollection` and `updateCollection` to use dependency injection for `modifyCollection`, so we could test `modifyCollection` independently and just check that `createCollection` and `updateCollection` call it appropriately, but I couldn't figure out how to test `modifyCollection` independently using `sagaTester`.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -829,61 +828,6 @@ describe(__filename, () => {
     });
   });
 
-  describe('updateCollection', () => {
```
bobsilverberg: I removed this because there isn't anything left to test here, but I noticed that we don't have coverage for testing just action creators which don't result in any state changes. I suppose there's not much to test, other than that the return result is what we expect, so maybe this is fine. But it seemed weird to add code and not also add a test for it.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -487,4 +488,122 @@ describe(__filename, () => {
       ).not.toContain(closeFormOverlay(formOverlayId).type);
     });
   });
+
+  describe('createCollection', () => {
```
bobsilverberg: Please see my comments above about why this is still two big test cases for `create` and `update` rather than one for `modify`.

src/amo/sagas/collections.js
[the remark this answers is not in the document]
kumar303: If this gets too bonkers then you could just use one saga for both instead of defining separate create and update sagas. What I mean is instead of this:

```js
export default function* collectionsSaga() {
  yield takeLatest(CREATE_COLLECTION, createCollection);
  yield takeLatest(UPDATE_COLLECTION, updateCollection);
}
```

...just use the same handler for both actions:

```js
export default function* collectionsSaga() {
  yield takeLatest(
    [CREATE_COLLECTION, UPDATE_COLLECTION], 
    modifyCollection
  );
}
```
The `modifyCollection` handler could look at `action.type` to know if it got a create/update action.

I don't know if this would help with the testing problem or not, just something that came to mind.

bobsilverberg:
Your idea of using `yield takeLatest([CREATE_COLLECTION, UPDATE_COLLECTION], modifyCollection)` along with a single method really cleaned things up. I think the test cases I wrote make sense - I believe they reduce duplication as much as makes sense and still give us full coverage.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -327,164 +328,248 @@ describe(__filename, () => {
...
-        description: { 'en-US': 'New collection description' },
-        formOverlayId: 'some-form-overlay',
-        name: { 'en-US': 'New collection name' },
-        slug: 'new-slug',
-        user: 543,
-      };
-      const state = sagaTester.getState();
+    describe('common logic', () => {
+      // The common logic can be tested either by dispatching updateCollection
+      // or createCollection, so we just use _updateCollection.
```
kumar303: I like this approach to the tests given the circumstances. The comment helps explain things too.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -327,164 +328,248 @@ describe(__filename, () => {
...
-      const expectedAction = finishFormOverlaySubmit(params.formOverlayId);
+      it('begins and finishes a form submit', async () => {
+        mockApi.expects('updateCollection').returns(Promise.resolve());
+        _updateCollection({ formOverlayId });
+
+        for (const expectedAction of [
+          beginFormOverlaySubmit(formOverlayId),
+          finishFormOverlaySubmit(formOverlayId),
+        ]) {
+          // eslint-disable-next-line no-await-in-loop
```
kumar303: We should delete this rule. We address the problem through code review. There are many legit uses of `await` within a for loop.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -327,164 +328,248 @@ describe(__filename, () => {
...
 
-      _updateCollection(params);
+        const action = await sagaTester.waitFor(expectedAction.type);
+        expect(action).toEqual(expectedAction);
+      });
 
-      const expectedAction = finishFormOverlaySubmit(params.formOverlayId);
+      it('begins and finishes a form submit', async () => {
+        mockApi.expects('updateCollection').returns(Promise.resolve());
+        _updateCollection({ formOverlayId });
```
kumar303: Hmm, yeah, I guess `formOverlayId` is used enough to warrant relying on the shared value. You might as well reap all the benefits and just leave it off everywhere:
```js
_updateCollection()
```

src/amo/sagas/collections.js
```diff
@@ -166,43 +168,70 @@ export function* addAddonToCollection({
...
-      name,
-      slug,
-      user,
-    };
-    yield call(api.updateCollection, params);
 
-    const slugWasEdited = slug && slug !== collectionSlug;
-    const effectiveSlug = slug || collectionSlug;
+    if (creating) {
+      const apiParams = {
```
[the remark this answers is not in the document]
bobsilverberg: That's how I originally had it, but I couldn't seem to get Flow happy with that in place and the other type definitions I had. When I removed it Flow didn't complain, so I assumed that meant it wasn't necessary. I'll try adding it back in and see if I can work through the Flow errors. I may post them here for some help.

tests/unit/amo/sagas/test_collections.js
bobsilverberg: I opened https://github.com/mozilla/eslint-config-amo/issues/53 as a follow-up for this.

src/amo/sagas/collections.js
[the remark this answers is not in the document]
bobsilverberg: Ok, the problem was because in the type for the function parameters, I need to make `name`, `slug` and `collectionSlug` all optional, because some are optional for each of the two actions. That makes the types `CreateCollectionParams` and `UpdateCollectionParams` incompatible. I was able to solve this by adding some calls to `invariant` to ensure that they were not empty when they shouldn't be, and that makes Flow happy. I suppose there's nothing wrong with this, and is in fact appropriate, although I know the the only code we have that calls this also makes sure those properties are populated when needed, so the extra `invariant`s seem unnecessary. I suppose there's no guarantee that there won't be a future caller who doesn't ensure that, which is kind of the point of Flow in the first place.

So I guess this all makes sense and I'm happy with it, and I learned something too. 😄

bobsilverberg:
Please ignore the eslint error for now. It will go away when I rebase this on top of https://github.com/mozilla/addons-frontend/commit/db95de86460733dc83da01ee327e68f8f5d39bbd, but I figured I'd wait for either another review request or an approval before doing that.

kumar303:
I didn't get a review request for this but I saw it mentioned in https://github.com/mozilla/addons-frontend/pull/4564 -- I think it's ready? I'll take another look

src/amo/sagas/collections.js
```diff
@@ -166,43 +171,71 @@ export function* addAddonToCollection({
...
+        description,
+        name,
+        slug,
+        user,
+      };
+      yield call(api.createCollection, apiParams);
+    } else {
+      invariant(collectionSlug,
+        'collectionSlug cannot be empty when updating');
+      const apiParams: UpdateCollectionParams = {
```
kumar303: Hmm. I see a lot of duplication between these two param objects. You can introduce a spread with a shared base object but Flow isn't able to compute all spread properties so it will complain about exact object types. There is a global `$Shape` utility we use to turn an exact type into an inexact type. It's not great because you lose the type safety around unexpected properties but being able to DRY up this code seems better.

Something like this should work as long you can get the disjoint unions to kick in:

```diff
diff --git a/src/amo/sagas/collections.js b/src/amo/sagas/collections.js
index 5cf55a0b6..f48a5a31f 100644
--- a/src/amo/sagas/collections.js
+++ b/src/amo/sagas/collections.js
@@ -202,29 +202,26 @@ export function* modifyCollection({
 
     const state = yield select(getState);
 
+    const baseApiParams = {
+      api: state.api,
+      defaultLocale,
+      description,
+      name,
+      slug,
+      user,
+    };
+
     if (creating) {
       invariant(name, 'name cannot be empty when creating');
       invariant(slug, 'slug cannot be empty when creating');
-      const apiParams: CreateCollectionParams = {
-        api: state.api,
-        defaultLocale,
-        description,
-        name,
-        slug,
-        user,
-      };
+      const apiParams: $Shape<CreateCollectionParams> = { ...baseApiParams };
       yield call(api.createCollection, apiParams);
     } else {
       invariant(collectionSlug,
         'collectionSlug cannot be empty when updating');
-      const apiParams: UpdateCollectionParams = {
-        api: state.api,
+      const apiParams: $Shape<UpdateCollectionParams> = {
+        ...baseApiParams,
         collectionSlug,
-        defaultLocale,
-        description,
-        name,
-        slug,
-        user,
       };
       yield call(api.updateCollection, apiParams);
     }
```

If this turns into a headache then it's no big deal. The most important part is gaining type safety for the API calls, which you already had.

bobsilverberg:
Oh, my mistake. I thought I requested a review yesterday. :-P

tests/unit/amo/reducers/test_collections.js
kumar303: > I removed this ... but I noticed that we don't have coverage for testing just action creators which don't result in any state changes.

We do. Any action creator that doesn't result in state changes exists solely to trigger a saga, which would be covered in [a test like this](https://github.com/mozilla/addons-frontend/blob/3d510e53b1011d83b3665a3de2aaecffaedc5da7/tests/unit/amo/sagas/test_collections.js#L332).

bobsilverberg:
@tofumatt kumar gave me an r+wc at https://github.com/mozilla/addons-frontend/pull/4534#pullrequestreview-104353818, but that was awhile ago and I've made a few changes, so if you don't mind taking a look as well, before I merge it, I would appreciate it. Thanks!

bobsilverberg:
Oops, I think I left some TODOs in there, although they should probably addressed by a follow-up, so it's probably okay.

src/amo/reducers/collections.js
```diff
@@ -449,21 +451,78 @@ export const addAddonToCollection = ({
...
+  name,
+  slug,
+  user,
+}: CreateCollectionParams = {}): CreateCollectionAction => {
+  validateRequiredCollectionParams({
+    errorHandlerId,
+    user,
+  });
+
+  invariant(name, 'name is required when creating');
```
tofumatt: Nitpick: I'd add "a collection" to the end here for clarity.

src/amo/reducers/collections.js
```diff
@@ -449,21 +451,78 @@ export const addAddonToCollection = ({
...
+
+export const createCollection = ({
+  errorHandlerId,
+  defaultLocale,
+  description,
+  name,
+  slug,
+  user,
+}: CreateCollectionParams = {}): CreateCollectionAction => {
+  validateRequiredCollectionParams({
```
tofumatt: This is actually more lines than just using the two invariant lines:

```js
invariant(user, 'user is required');
```

I get it doesn't duplicate the messages but we already dupe them elsewhere and invariant should let us nix all the messages in production anyway. I'd rather us just use invariant directly.

tests/unit/amo/sagas/test_collections.js
tofumatt: I know fast tests are good and we don't want to go overboard, but testing each action feels a bit safer for future refactors for me. I'm okay with this though, it's cool. 🤷‍♂️

tests/unit/amo/sagas/test_collections.js
```diff
@@ -324,140 +325,220 @@ describe(__filename, () => {
...
+            name: params.name,
+            slug: params.slug,
+            user,
+          })
+          .once()
+          .returns(Promise.resolve());
+
+        _updateCollection(params);
+
+        // TODO: Instead of this we should probably update the collectionUpdates state,
```
tofumatt: Can we get an issue filed for this if it's still the case and a code quality thing we should fix in the future?

bobsilverberg: I added issue #4717 to track this, and added a link to that issue into the TODO comment.

PR #4538, opened 2018-03-12
Allow for a collection that does not have any add-ons to be viewed

### LINKED ISSUE #4537: It should be possible to view a collection that does not have any add-ons
This currently returns a 404 because the API call to retrieve the add-ons for the collection returns 0 results. I believe this can be fixed by checking the value of `addon_count` in the result of the API call to retrieve the collection details, and then only trying to retrieve the add-ons for the collection if there are any.

bobsilverberg:
Fixes #4537 

This allows for the above, which previously resulted in a 404, and also adds a message in the area where the add-ons would normally be informing the user that the collection does not have any add-ons yet.

### DIFF
--- a/src/amo/sagas/collections.js
+++ b/src/amo/sagas/collections.js
@@ -2,7 +2,7 @@
 // Disabled because of
 // https://github.com/benmosher/eslint-plugin-import/issues/793
 /* eslint-disable import/order */
-import { all, call, put, select, takeLatest } from 'redux-saga/effects';
+import { call, put, select, takeLatest } from 'redux-saga/effects';
 import { push as pushLocation } from 'react-router-redux';
 /* eslint-enable import/order */
 
@@ -36,11 +36,16 @@ import type {
 } from 'amo/api/collections';
 import type {
   AddAddonToCollectionAction,
+  CollectionAddonsListResponse,
   FetchCurrentCollectionAction,
   FetchCurrentCollectionPageAction,
   FetchUserCollectionsAction,
   UpdateCollectionAction,
 } from 'amo/reducers/collections';
+
+function emptyAddons(): CollectionAddonsListResponse {
+  return { count: 0, next: '', previous: '', results: [] };
+}
 
 export function* fetchCurrentCollection({
   payload: {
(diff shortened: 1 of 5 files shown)


src/amo/sagas/collections.js
```diff
@@ -36,12 +36,17 @@ import type {
...
   AddAddonToCollectionAction,
+  CollectionAddonsListResponse,
   FetchCurrentCollectionAction,
   FetchCurrentCollectionPageAction,
   FetchUserCollectionsAction,
   UpdateCollectionAction,
 } from 'amo/reducers/collections';
 
+function emptyAddons(): CollectionAddonsListResponse {
+  return { count: 0, next: '', previous: '', results: [] };
```
bobsilverberg: I'm not particularly happy about this, but Flow types plus the automated test should help us keep it in sync if the return values from the API method change. I suppose ideally there'd be something in our API code which could deliver a standard `ListResponse`, but I don't think such a thing exists, and even if it did we'd still need to manually keep it in sync.

src/amo/sagas/collections.js
```diff
@@ -70,10 +75,10 @@ export function* fetchCurrentCollection({
       ...baseParams,
       page,
     };
-    const { detail, addons } = yield all({
-      detail: call(api.getCollectionDetail, detailParams),
-      addons: call(api.getCollectionAddons, addonsParams),
-    });
+    const detail = yield call(api.getCollectionDetail, detailParams);
+    const addons = detail.addon_count ?
```
kumar303: Why don't you just make the `addons` parameter of `loadCurrentCollection()` optional? It probably should have always been optional; it has always been possible to load a collection without add-ons. I looked at the action and reducer and I think making it optional would be fine.

tests/unit/amo/helpers.js
```diff
@@ -345,6 +345,8 @@ export const createFakeCollectionDetail = ({
 export const createFakeCollectionAddons = ({ addons = [fakeAddon] } = {}) => {
   return {
     count: addons.length,
+    previous: '',
```
kumar303: You might as well fix this whole function to use the [`apiResponsePage()`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/helpers.js#L236) helper that we already have.

src/amo/sagas/collections.js
kumar303: Actually, let me back up. What is triggering the 404? Is it the API call to get add-ons? The description of your original bug made it sound like the API returns zero results but not a 404.

bobsilverberg: I was sure that the API call to get the add-ons was triggering the 404, but now I cannot reproduce it at all, either on -dev or locally, either on master or with my branch, so I have no idea what's going on. Let's just leave this for now and if I experience it again I'll do some more digging. Sorry for the noise.

bobsilverberg:
As I mentioned at https://github.com/mozilla/addons-frontend/issues/4537#issuecomment-373139954, this was not actually a bug, but the result of me redirecting to a new collection immediately after creating it.

PR #4546, opened 2018-03-13
Update circleci version

### LINKED ISSUE #4519: CircleCI v1.0 no longer available 8/31/2018
CircleCI v1.0 will be unavailable after 8/31/2018. 

### What happened?
CircleCI is sunsetting their v1.0.

We need to translate the circle.yml file into their v2.0 format.

Some documents for translating from v1.0 to v2.0
* https://circleci.com/docs/2.0/migrating-from-1-2/
* https://circleci.com/docs/2.0/config-translation

### LINKED ISSUE #4544: Update circleci to version 2
We should update circleci to version 2 as the new config is a bit cleaner and allows for faster builds. It will also help with the new frontend tests I will be adding.

jrbenny35:
Fixes #4544 fixes #4519


circle.yml
```diff
@@ -1,72 +0,0 @@
-# These environment variables must be set in CircleCI UI
```
kumar303: This comment still applies, right? I still see the same env vars in use.

.circleci/config.yml
```diff
@@ -0,0 +1,97 @@
...
+          printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n'
+          "$CIRCLE_SHA1"
+          "$CIRCLE_TAG"
+          "$CIRCLE_PROJECT_USERNAME"
+          "$CIRCLE_PROJECT_REPONAME"
+          "$CIRCLE_BUILD_URL"
+          > version.json
+      - run: docker build --pull -t addons-frontend .
+      - run: docker images
+        # write the sha256 sum to an artifact to make image verification easier
```
kumar303: I don't think you need this comment now that the command has a name

.circleci/config.yml
```diff
@@ -0,0 +1,97 @@
...
+      - run:
+          name: Write the sha256 sum to an artifact
+          command: |
+            docker images --no-trunc | awk '/^app/ {print $3}' | tee $CIRCLE_ARTIFACTS/docker-image-shasum256.txt
+        # Clean up any old images; save the new one.
```
kumar303: Same goes for this one

.circleci/config.yml
```diff
@@ -0,0 +1,97 @@
...
+          "$CIRCLE_TAG"
+          "$CIRCLE_PROJECT_USERNAME"
+          "$CIRCLE_PROJECT_REPONAME"
+          "$CIRCLE_BUILD_URL"
+          > version.json
+      - run: docker build --pull -t addons-frontend .
+      - run: docker images
+        # write the sha256 sum to an artifact to make image verification easier
+      - run:
+          name: Write the sha256 sum to an artifact
```
kumar303: I think it would be good to add "... for verification" on to the end of this.

circle.yml
jrbenny35: Yes, I didn't change any variables.

kumar303: Sorry, I wasn't very explicit. I meant please put this comment back at the top of the file since it still applies. It seems important.

jrbenny35: Done!

PR #4548, opened 2018-03-13
fix: Make "Download Firefox" notice less red

### LINKED ISSUE #4547: Change "Download Firefox" message bar from red to grey
This red is scary and could be (probably is) negatively affecting Firefox download numbers:
screen shot 2018-03-13 at 8 45 23 am

We need to change it to the ["generic" style](https://design.firefox.com/photon/components/message-bars.html#color).

tofumatt:
Fix #4547.

### Before
screen shot 2018-03-13 at 8 45 23 am

### After
screenshot 2018-03-13 16 19 41

### DIFF
--- a/src/amo/components/AddonCompatibilityError/index.js
+++ b/src/amo/components/AddonCompatibilityError/index.js
@@ -95,8 +95,13 @@ export class AddonCompatibilityErrorBase extends React.Component {
       ), { downloadUrl });
     }
 
+    // Make the "you should download firefox" error message less scary than
+    // the rest of them: https://github.com/mozilla/addons-frontend/issues/4547
+    const noticeType = reason === INCOMPATIBLE_NOT_FIREFOX ?
+      'generic' : 'error';
+
     return (
-      <Notice type="error" className="AddonCompatibilityError">
+      <Notice type={noticeType} className="AddonCompatibilityError">
         <span dangerouslySetInnerHTML={sanitizeHTML(message, ['a'])} />
       </Notice>
     );
--- a/src/ui/components/Notice/index.js
+++ b/src/ui/components/Notice/index.js
@@ -7,8 +7,9 @@ import Button from 'ui/components/Button';
 import './styles.scss';
 
 const errorType: 'error' = 'error';
+const genericType: 'generic' = 'generic';
 const successType: 'success' = 'success';
-const validTypes = [errorType, successType];
+const validTypes = [errorType, genericType, successType];
 
 type Props = {
   action?: Function,
@@ -22,11 +23,9 @@ type Props = {
  * A Photon style notification bar.
  *
  * See https://design.firefox.com/photon/components/message-bars.html
- *
- * TODO: implement generic notices and warnings.
  */
 const Notice = ({
-  action, actionText, children, className, type,
+  action, actionText, children, className, type = 'generic',
 }: Props) => {
   if (!validTypes.includes(type)) {
     throw new Error(`Unknown type: ${type}`);
--- a/src/ui/components/Notice/index.js
+++ b/src/ui/components/Notice/index.js
@@ -25,7 +25,7 @@ type Props = {
  * See https://design.firefox.com/photon/components/message-bars.html
  */
 const Notice = ({
-  action, actionText, children, className, type = 'generic',
+  action, actionText, children, className, type,
 }: Props) => {
   if (!validTypes.includes(type)) {
     throw new Error(`Unknown type: ${type}`);
--- a/src/ui/components/Notice/styles.scss
+++ b/src/ui/components/Notice/styles.scss
@@ -76,6 +76,31 @@
   }
 }
 
+.Notice-generic {
+  .Notice-button {
+    @include button(
+      $background: $blue-60,
+      $background-active: $blue-80,
+      $background-hover: $blue-70,
+      $color: $grey-90
+    );
+    @include notice_button();
+  }
+
+  .Notice-icon {
+    background-image: url("./media/generic-info-mark.svg");
+  }
+
+  background-color: $grey-20;
+
+  &,
+  :link,
+  :hover,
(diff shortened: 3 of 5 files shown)


tests/unit/amo/components/TestAddonCompatibilityError.js
```diff
@@ -68,6 +69,26 @@ describe(__filename, () => {
...
+    _dispatchClientMetadata({
+      userAgent: userAgentsByPlatform.mac.chrome41,
+    });
+    const root = render({ reason: INCOMPATIBLE_NOT_FIREFOX });
+
+    expect(root.find('.AddonCompatibilityError').find(Notice)).toHaveProp(
+      'type', 'generic');
+  });
+
+  it('renders an error notice for other reasons browsers', () => {
```
bobsilverberg: Should this name just be `renders an error notice for other reasons` (i.e., lose the `browsers` at the end)?

kumar303:
Oh, hah. I see that @pwalm is implying that the mock is out of date. I'll take a closer look.

tests/unit/amo/components/TestAddonCompatibilityError.js
tofumatt: Oops, thanks.

tofumatt:
This uses the current Notice component and moves away from the scary red. There are probably other style issues here that need to be changed as the spec has changed, but this is still an improvement.

bobsilverberg:
Actually, following up on what @kumar303 said, the example that @pwalm linked to uses a button for the action, as opposed to an inline link. Is that what we should be doing?

src/amo/components/AddonCompatibilityError/index.js
```diff
@@ -95,8 +95,13 @@ export class AddonCompatibilityErrorBase extends React.Component {
       ), { downloadUrl });
     }
 
+    // Make the "you should download firefox" error message less scary than
+    // the rest of them: https://github.com/mozilla/addons-frontend/issues/4547
+    const noticeType = reason === INCOMPATIBLE_NOT_FIREFOX ?
+      'generic' : 'error';
```
kumar303: I guess it would be hard to switch this to a [download button](https://github.com/mozilla/addons-frontend/blob/62015810feb4797a5fc5019581e745bfd21e5420/src/amo/components/AddonCompatibilityError/index.js#L62)? The `Notice` component already supports it. If it's difficult then we should do it in a follow-up issue.

Call to actions look like this: https://design.firefox.com/photon/components/message-bars.html#call-to-action

src/ui/components/Notice/index.js
```diff
@@ -22,11 +23,9 @@ type Props = {
  * A Photon style notification bar.
  *
  * See https://design.firefox.com/photon/components/message-bars.html
- *
- * TODO: implement generic notices and warnings.
  */
 const Notice = ({
-  action, actionText, children, className, type,
+  action, actionText, children, className, type = 'generic',
```
kumar303: I don't think it's a good idea to default `type` to `generic`. I'd rather the caller see an error when they forget to declare a `type`. If we find that it's cumbersome to declare an explicit `type` all the time then we could revisit it. However, I think `error` would be a more common type.

tofumatt:
So one of the issues here is the download link is in big scary red text and looks like an error, when it's more of a notice. The entire `Notice` component needs updating to be in line with the Photon spec from what I can see.

I can see about changing the link to a button but I just wanted to start by getting it less red and scary...

src/ui/components/Notice/styles.scss
```diff
@@ -76,6 +76,31 @@
   }
 }
 
+.Notice-generic {
+  .Notice-button {
+    @include button(
+      $background: $blue-60,
```
kumar303: Hmm. I don't see blue anywhere in the [action button spec](https://design.firefox.com/photon/components/message-bars.html#call-to-action). Also, does it have the correct hover styles? That mixin is a bit confusing. You may need some overrides similar to the `success` one.

tofumatt:
I looked and I could add the button by messing with the `Notice` component to accept a `href` prop for the `button` it creates from `action` props but it seems messy and I think I'd rather have a follow-up issue to make the Notice/Notice buttons a bit more flexible for this use. If folks would be cool with that let me know and I'll file it and get this into the tag...

src/amo/components/AddonCompatibilityError/index.js
tofumatt: It supports a button but not one with a `href` prop, just an `onClick` action. It's worth doing separately I think if we want this in for this week. I thought it'd be quick but I think it's worth leaving for now.

src/ui/components/Notice/styles.scss
tofumatt: It's in the spec @pwalm linked to in #4547: https://design.firefox.com/photon/components/message-bars.html#color

kumar303:
> I am not sure about the action button colors.

Just to clarify, I requested a change on this because I think it's harder to fix these things later vs the first time. But if you want to do a follow-up issue to fix the colors, that's fine.

> The entire Notice component needs updating to be in line with the Photon spec from what I can see.

Here's an example of adding a button: https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/ErrorList/index.js#L83-L87

Yeah, it probably needs some tweaks to support an `href`. That's fine for a follow-up.

src/ui/components/Notice/styles.scss
kumar303: Oh man, I'm so confused. I see the blue now in what you linked to but it's gray everywhere else.

The call to action part specifies gray values: https://design.firefox.com/photon/components/message-bars.html#call-to-action

The default button looks gray: https://design.firefox.com/photon/components/buttons.html#default-4

tofumatt:
Thanks! I've filed #4549.

Issue #4431, opened 2018-02-22
Cards have tiny or cramped up text


FatOrangutan:
### Describe the problem and steps to reproduce it:

Go to https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/.

Look at the first card, `.Addon-header-info-card`:
![screenshot-2018-2-22 ublock origin add-ons for firefox](https://user-images.githubusercontent.com/8336638/36520249-f9590f64-17b5-11e8-8083-6c2f8409ec03.png)
There is a whole lot of white space and yet the line-height is too small - the `g` in `uBlock Origin` is almost touching the `l` in `Raymond Hill`.

Also, look at the add-on description, `.AddonDescription-contents`:

![screenshot-2018-2-22 ublock origin add-ons for firefox 2](https://user-images.githubusercontent.com/8336638/36520633-0c45c098-17b8-11e8-9e93-9c0ff435cc0e.png)

This is just painful to read. Between the previous design and this one, at least on Windows, the font-size of what is usually the most text heavy section has gone down from 13px to 12px, while the x-height of the font in use has also slightly decreased. All of this while padding and line-height for the same section have both increased. If I change the font-size and the line-height to their older values, the entire section is significantly easier to read while being shorter in total height.

I'm on a 15.5", 1920x1080 screen at 125% scaling. I don't typically have trouble reading text on other sites.

muffinresearch:
@pwalm Assuming I'm reading the mocks in Sketch correctly the default text copy looks to be 15px not 12px - is that right?

bobsilverberg:
Duplicate of #4520

PR #4551, opened 2018-03-14
I4520 : layout change for addon badges

### LINKED ISSUE #4520: Update Addon-header-info-card to match Sketch specs
The Sketch specs for the Addon-header-info-card differ from what we currently have in a number of ways. The layout and formatting of the text is different (which was also reported in issue #4431), and the badges are in a different location.

Sketch has the card looking like:

screenshot 2018-03-07 12 46 58

Whereas the current site looks like:

screenshot 2018-03-07 12 47 45

Duplicates #4431

tsl143:
Fixes #4520 

Before:
screen shot 2018-03-14 at 8 26 49 am

After:
screen shot 2018-03-14 at 8 27 21 am

### DIFF
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -7,19 +7,13 @@
 
   @include respond-to(large) {
     display: flex;
-    flex-flow: row wrap;
+    flex-flow: column;
+    position: absolute;
+    right: 0;
+    width: auto;
   }
 }
 
 .AddonBadges .Badge {
   @include text-align-end();
-
-  @include respond-to(large) {
-    @include margin-end(6px);
-    @include text-align-start();
-  }
-
-  @include respond-to(extraLarge) {
-    @include margin-end($padding-page-l);
-  }
 }
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -12,5 +12,9 @@
 }
 
 .AddonBadges .Badge {
-  @include text-align-end();
+  @include text-align-start();
+
+  @include respond-to(large) {
+    @include text-align-end();
+  }
 }
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -6,11 +6,8 @@
   width: 100%;
 
   @include respond-to(large) {
-    display: flex;
-    flex-flow: column;
-    position: absolute;
-    right: 0;
-    width: auto;
+    grid-column: 2;
+    grid-row: 1;
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/amo/components/AddonBadges/styles.scss
```diff
@@ -7,19 +7,13 @@
 
   @include respond-to(large) {
     display: flex;
-    flex-flow: row wrap;
+    flex-flow: column;
+    position: absolute;
```
tofumatt: It seems weird to absolutely position something and have it be a flex container.

tsl143: yes, I think we can get rid of `display: flex`, if we use `position: absolute`

tsl143:
@tofumatt , in this fix the badges will be stacked vertically, which can overlap with title and author names. So what should be the ideal behavior here? Also max how many badges we are expecting.

bobsilverberg:
Thanks for taking this on, @tsl143. The issue I filed was not just about the badge positioning, but also about some changes to the text and the way it is laid out, some of which is described at https://github.com/mozilla/addons-frontend/issues/4431#issuecomment-369987147. I think it may be fine to only fix the badge positioning in this patch, but we shouldn't close issue #4520 as a result, or we should open a new follow-up issue to address the other changes needed for this card.

tsl143:
@tofumatt got it, I will update this PR.

tsl143:
Used grid to play with position of badges on mobile and desktop.

screen shot 2018-03-15 at 8 07 03 am
screen shot 2018-03-15 at 8 06 51 am

tofumatt:
That’s cool, but seeing them right-aligned on mobile definitely looks weird. Let’s have them aligned to the left on mobile.

- Matt (Sent from mobile)

tsl143:
@tofumatt that will surely look good, in addition, on mobile shall we put icon first, like below?
image

tofumatt:
Yes! Oops, I forgot to even say that but yes that’s awesome. Thanks!

tsl143:
screen shot 2018-03-15 at 11 16 12 pm
screen shot 2018-03-15 at 11 16 01 pm

src/ui/components/Badge/styles.scss
```diff
@@ -9,11 +9,17 @@
...
 
   @include respond-to(medium) {
+    display: flex;
+    flex-direction: row-reverse;
     margin: 0 0 12px;
   }
 
   .Icon {
-    @include margin-start(6px);
+    @include margin(0, 6px, 0, 0);
```
tofumatt: I didn't know we had this function. I'm guessing it just inverts the left/right for RTL?

tofumatt: Hah, I see it is. Cool!

Issue #4280, opened 2018-01-31
Show "Extensions created" on the user profile page


tofumatt:
Create a card that shows extensions a user has created on the profile.

<img width="482" alt="screenshot 2018-01-31 14 50 59" src="https://user-images.githubusercontent.com/90871/35629308-31bf268a-0696-11e8-9bd9-9ea071ba5173.png">

ioanarusiczki:
@tofumatt 
Two questions here:
1. The number of ad-ons displayed is three (when number of submitted ad-ons in dev hub is more than3).  I have an account with many add-ons (30 plus), also many lightweight themes. If I refresh the page or log out then log in later the [User Profile](https://addons-dev.allizom.org/en-US/firefox/user/DevTesting/) page displays same ad-ons and themes. Shouldn't this be updated when refreshing page the way promo banners are on AMO homepage? Or maybe a See more by *Username* link added?

2. Static themes are not displayed yet on the info card. I submitted some static themes and approved for [another account ](https://addons-dev.allizom.org/en-US/firefox/user/Contrating2/)to see if they show up on the info card.

tofumatt:
1. Could you please send a link? That might be better filed as new issue though.

2. Are static themes a new theme type (cc @eviljeff)? Sounds like a new issue, maybe we need to expand the addonType we search for.

- Matt (Sent from mobile)

ioanarusiczki:
@tofumatt 
I filed two separate issues 
the More/Refresh issue https://github.com/mozilla/addons-frontend/issues/4809
the Static Themes https://github.com/mozilla/addons-frontend/issues/4810

Marking this verified fixed on AMO dev with FF59 (Win10).

tofumatt:
Yeah, the cards are explicitly just supposed to show extensions and themes so the other types are ignored. But I didn’t know about the new static theme type. We’ll have to add that.

- Matt (Sent from mobile)

PR #4554, opened 2018-03-14
chore: Allow fetchAddonsByAuthors to search for add-ons without specifying an `exclude_addon` param

### LINKED ISSUE #4556: Allow fetchAddonsByAuthors to omit the exclude_addons param
To implement #4280 and similar issues, we need to be able to fetch add-ons by a user.

Turns out we essentially have this functionality in `fetchAddonsByAuthors` but it requires a `slug` to send to the API as an `exclude_addons` param. But if we make that `slug` optional we can re-use this reducer for "add-ons by this user" on the profile page.

tofumatt:
Part of #4280.

Fixes #4556.

Depends on #4469.

Allows `addonsByAuthor` to fetch an author's add-ons without an `exclude_addon` prop sent to the API. This allows a user's add-ons to be show on the user profile without needing to write a whole new reducer; it just makes the slug here optional and names it a bit more clearly.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -97,6 +97,7 @@ module.exports = {
     'defaultLang',
     'enableClientConsole',
     'enableCollectionEdit',
+    'enableUserProfile',
     'fxaConfig',
     'isDeployed',
     'isDevelopment',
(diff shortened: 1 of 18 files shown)


tofumatt:
Nixing this in favour of #4571.

PR #4555, opened 2018-03-14
Updated file paths and removed caching.

jrbenny35:
Unfortunately circleci doesn't let you test a build that has to run on master until you merge it. My previous PR had some incorrect paths that I didn't know of until it was merged. This should help fix it but I won't really be able to test it until it is on master.

I did walk through the steps within the circleci container via SSH so they should work, but it may be the little nuances of file paths and permissions that we may run into.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
           name: Build docker image and push to repo
           command: |
             [ ! -z $DOCKERHUB_REPO ]
-            docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
+            docker login -u $DOCKER_USER -p $DOCKER_PASS
             docker tag addons-frontend $DOCKERHUB_REPO:latest
             docker push $DOCKERHUB_REPO:latest
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,11 +20,7 @@ jobs:
         # day to keep the cache size down.
       - run:
           name: Build the container ( Next 3 steps )
-<<<<<<< HEAD
-          command: I="image-$(date +%j).gz"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/docker/$I | docker load; fi
-=======
           command: I="image-$(date +%j).gz"; if [[ -e ~/addons-frontend/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/addons-frontend/docker/$I | docker load; fi
->>>>>>> Updated file paths and removed caching.
       - run: >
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         # day to keep the cache size down.
       - run:
           name: Build the container ( Next 3 steps )
-          command: I="image-$(date +%j).gz"; if [[ -e ~/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/docker/$I | docker load; fi
+          command: I="image-$(date +%j).gz"; if [[ -e ~/addons-frontend/docker/$I ]]; then echo "Loading $I"; pigz -d -c ~/addons-frontend/docker/$I | docker load; fi
```
kumar303: Since `$I` is referenced across multiple commands now, can you call it `$IMG`? That will make it easier to search for when trying to figure it out.

.circleci/config.yml
```diff
@@ -34,47 +34,40 @@ jobs:
...
           name: Clean up any old images and save the new one.
           command: |
-            I="image-$(date +%j).gz"; mkdir -p ~/docker; rm ~/docker/*
-            docker save addons-frontend | pigz --fast -c > ~/docker/$I
-            ls -l ~/docker
-      - save_cache:
-          key: docker-cache-{{ checksum "~/docker" }}
-          paths:
-            - ~/docker
+            docker save addons-frontend | pigz --fast -c > ~/addons-frontend/docker/$I
```
kumar303: Are you sure one command can reference a variable (`$I`) from a previous command like this?

.circleci/config.yml
```diff
@@ -34,47 +34,40 @@ jobs:
...
     machine: true
-    working_directory: ~/addons-frontend
     steps:
       - attach_workspace:
           at: .
       - run:
           name: Build docker image and push to repo
           command: |
             [ ! -z $DOCKERHUB_REPO ]
-            docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
```
kumar303: Please also remove `DOCKER_EMAIL` from the comment at the top of the file. It looks like it's no longer needed.

.circleci/config.yml
```diff
@@ -1,7 +1,6 @@
 # These environment variables must be set in CircleCI UI
 #
 # DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
-# DOCKER_EMAIL   - login info for docker hub
 # DOCKER_USER
```
kumar303: Since you removed the docker hub comment, it's no longer obvious what the other variables are. I suggest adding it back like this:

```
# DOCKER_USER - login user for docker hub
# DOCKER_PASS - login password for docker hub
```

jrbenny35: Makes sense, kinda figured this was implied though.

PR #4557, opened 2018-03-14
Upgrade eslint-config-amo to v1.7.0

bobsilverberg:
This will pick up the rule change to turn off the no-await-in-loop rule


kumar303:
Can you remove the disable flags too? https://github.com/mozilla/addons-frontend/search?utf8=%E2%9C%93&q=no-await-in-loop&type=

bobsilverberg:
@tofumatt I noticed last time that Greenkeeper never picked this change up. Do both repos need to be configured for that to work? Maybe eslint-amo-config isn't configured properly?

PR #4559, opened 2018-03-14
Fixed rm command within circleci steps.

jrbenny35:
A small fix


.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - run:
           name: Clean up any old images and save the new one.
```
kumar303: I guess the name can be updated now

jrbenny35:
> If you don't think it needs to clean up then that's fine with me.

With no cache, there is nothing to clean since it will be a fresh clone every time it builds.

PR #4560, opened 2018-03-14
Fix for release jobs on circleci.

jrbenny35:
Loading the image within the release jobs needed to be done.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,9 @@ jobs:
       - attach_workspace:
           at: .
       - run:
+          name: Load saved image
+          command: docker load -i image-*
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -48,6 +48,9 @@ jobs:
     steps:
       - attach_workspace:
           at: .
+      - run:
+          name: Load saved image
+          command: docker load -i image-*
```
kumar303: I don't know how `load` works but shouldn't that be `docker load -i ~/addons-frontend/docker/image-*` ?

jrbenny35: Ah yes, it should be ```docker/image-*```

Issue #4562, opened 2018-03-14
Expose AutoSearchInput label and make it more embeddable


kumar303:
The AutoSearchInput component added in https://github.com/mozilla/addons-frontend/issues/4510 needs some tweaks for https://github.com/mozilla/addons-frontend/issues/4004

* It has a hidden form label. This should be visible by default and configurable.
* It requires a `position: relative` container. This should be fully extracted from `SearchForm`. This will allow it to be embeddable in other components.

**For QA**: Please check for style regressions in the search box at the top of the page

ioanarusiczki:
Verified on AMO dev with FF59 (Win10 and Android 8.0). No style regressions observed.

PR #4563, opened 2018-03-14
Expose AutoSearchInput label, add a position:relative container

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4562

### DIFF
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -49,7 +49,6 @@ type Props = {|
   debounce: typeof defaultDebounce,
   dispatch: DispatchFunc,
   i18n: I18nType,
-  inputLabelIsHidden?: boolean,
   inputLabelText?: string,
   // This is the name property of the <input> tag.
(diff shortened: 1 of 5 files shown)


src/amo/components/AutoSearchInput/index.js
```diff
@@ -96,6 +98,7 @@ export class AutoSearchInputBase extends React.Component<Props, State> {
 
   static defaultProps = {
     debounce: defaultDebounce,
+    inputLabelIsHidden: false,
```
tofumatt: Specifying `inputLabelIsHidden={true}` (or, really, just `<AutoSearchInput inputLabelIsHidden {...otherProps} />` is a weird API to me. It might be better to have this be `showInputLabel` so it would be `<AutoSearchInput showInputLabel={false} {...otherProps} />` when used. I think setting something to `false` for it to be hidden makes a bit more sense cognitively.

Or at least have it be more instructive if it's a boolean you set to true... maybe `hideInputLabel` or something.

PR #4564, opened 2018-03-15
Create a new collection (name, description, URL)

### LINKED ISSUE #4003: Create a new collection (name, description, URL)

Depends on: https://github.com/mozilla/addons-frontend/issues/4504, https://github.com/mozilla/addons-frontend/issues/4505

A user should be able to:

- Sign in
- Create a new collection with a name, description, and URL
- Assigning add-ons to the collection will be done in https://github.com/mozilla/addons-frontend/issues/4502
- Save the collection

This will be similar to editing a collection: https://github.com/mozilla/addons-frontend/issues/4002

bobsilverberg:
Fixes #4003 

This is an initial attempt at this, and I know it has a number of issues, but I thought it worth submitting for a look anyway. That way I can respond to feedback rather than trying to guess what's best to do.

Note that I had to build it on top of what's currently in PR #4534, so there are three extra commits in here. For the purposes of reviewing my work for #4003, you can ignore the changes to:

- src/amo/reducers/collections.js
-  src/amo/sagas/collections.js
- tests/unit/amo/reducers/test_collections.js
- tests/unit/amo/sagas/test_collections.js

I will add some additional comments calling out the areas where I still see issues.

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -146,7 +146,7 @@ export class CollectionBase extends React.Component<Props> {
     const { _config, i18n } = this.props;
     const props = {};
 
-    if (_config.get('enableCollectionEdit')) {
+    if (_config.get('enableNewCollectionsUI')) {
       // TODO: make this a real link when the form is ready for release.
       // https://github.com/mozilla/addons-frontend/issues/4293
       props.to = this.editUrl();
@@ -172,7 +172,7 @@ export class CollectionBase extends React.Component<Props> {
           <AuthenticateButton
             noIcon
             location={location}
-            logInText={i18n.gettext('Sign in to edit this collection')}
+            logInText={i18n.gettext('Log in to edit this collection')}
           />
         );
       }
--- a/src/amo/components/CollectionList/index.js
+++ b/src/amo/components/CollectionList/index.js
@@ -35,11 +35,11 @@ export class CollectionListBase extends React.Component<Props> {
           <AuthenticateButton
             noIcon
             location={location}
-            logInText={i18n.gettext('Sign in to create a collection')}
+            logInText={i18n.gettext('Log in to create a collection')}
           />
         );
       }
-      return (<CollectionManager {...this.props} creating />);
+      return <CollectionManager {...this.props} creating />;
     }
     return null;
   }
--- a/src/amo/components/CollectionList/index.js
+++ b/src/amo/components/CollectionList/index.js
@@ -1,0 +1,71 @@
+/* @flow */
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import CollectionManager from 'amo/components/CollectionManager';
+import { getCurrentUser } from 'amo/reducers/users';
+import AuthenticateButton from 'core/components/AuthenticateButton';
+import translate from 'core/i18n/translate';
+import Card from 'ui/components/Card';
+import type { UsersStateType } from 'amo/reducers/users';
+import type { I18nType } from 'core/types/i18n';
+import type { ReactRouterLocation } from 'core/types/router';
+
+import './styles.scss';
+
+
+export type Props = {|
+  creating: boolean,
+  i18n: I18nType,
+  isLoggedIn: boolean,
+  location: ReactRouterLocation,
+|};
+
+export class CollectionListBase extends React.Component<Props> {
+  static defaultProps = {
+    creating: false,
+  };
+
+  renderManager() {
+    const { creating, i18n, isLoggedIn, location } = this.props;
+    if (creating) {
+      if (!isLoggedIn) {
+        return (
+          <AuthenticateButton
+            noIcon
+            location={location}
+            logInText={i18n.gettext('Sign in to create a collection')}
+          />
+        );
+      }
+      return (<CollectionManager {...this.props} creating />);
+    }
+    return null;
+  }
+
+  render() {
+    return (
+      <div className="CollectionList">
+        <div className="CollectionList-wrapper">
+          <Card className="CollectionList-create">
+            {this.renderManager()}
+          </Card>
+        </div>
+      </div>
+    );
+  }
+}
+
+export const mapStateToProps = (
+  state: {| users: UsersStateType |}
+) => {
+  return {
+    isLoggedIn: !!getCurrentUser(state.users),
+  };
+};
+
+export default compose(
+  translate(),
+  connect(mapStateToProps),
+)(CollectionListBase);
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import { oneLineTrim } from 'common-tags';
 import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
@@ -36,16 +37,16 @@ import type { ReactRouterType } from 'core/types/router';
 import './styles.scss';
 
 type Props = {|
-  creating: boolean,
   collection: CollectionType | null,
   clientApp: ?string,
+  creating: boolean,
+  currentUsername: string,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  isCollectionBeingModified: boolean,
   router: ReactRouterType,
   siteLang: ?string,
-  isCollectionBeingModified: boolean,
-  currentUsername: string,
 |};
 
 type State = {|
@@ -72,10 +73,15 @@ export class CollectionManagerBase extends React.Component<Props, State> {
 
   onCancel = (event: SyntheticEvent<any>) => {
     const {
-      collection, clientApp, errorHandler, router, siteLang,
+      clientApp, collection, creating, errorHandler, router, siteLang,
     } = this.props;
     event.preventDefault();
     event.stopPropagation();
+
+    if (creating) {
+      router.goBack();
+    }
+
     invariant(collection,
       'A collection must be loaded before you can cancel');
     invariant(clientApp,
@@ -97,11 +103,10 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     const {
       creating,
       collection,
+      currentUsername,
       dispatch,
       errorHandler,
-      i18n,
       siteLang,
-      currentUsername,
     } = this.props;
     event.preventDefault();
     event.stopPropagation();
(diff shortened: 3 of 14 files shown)


src/amo/components/Collection/index.js
```diff
@@ -71,11 +75,18 @@ export class CollectionBase extends React.Component<Props> {
...
   loadDataIfNeeded(nextProps?: Props) {
-    const { collection, errorHandler, loading, params } = {
+    const { creating, collection, dispatch, errorHandler, isLoggedIn, loading, params } = {
       ...this.props,
       ...nextProps,
     };
 
+    if (creating) {
+      if (isLoggedIn) {
+        dispatch(openFormOverlay(CREATE_COLLECTION_OVERLAY));
```
bobsilverberg: The approach I took for locating this page was to have an unloaded collection appear in the background as this logic (for creating a collection) seems to make sense to have in the `Collection` component. I'm not totally happy with it, and there's an issue that if you cancel the overlay you are left at an unloaded collection screen, so advice here would be appreciated.

src/amo/components/Collection/index.js
```diff
@@ -166,15 +177,21 @@ export class CollectionBase extends React.Component<Props> {
...
       i18n,
+      isLoggedIn,
       loading,
       location,
     } = this.props;
 
     const addons = collection ? collection.addons : [];
 
+    if (creating && !isLoggedIn) {
+      return <NotAuthorized />;
```
bobsilverberg: This seemed like a logical thing to do.

src/amo/components/Collection/index.js
```diff
@@ -228,7 +245,9 @@ export class CollectionBase extends React.Component<Props> {
           The manager overlay will render full-screen.
           It will be opened/closed based on Redux state.
         */}
-        <CollectionManager collection={collection} />
+        {isLoggedIn ?
```
bobsilverberg: This code wasn't here before, but it seems like we only need to include the `CollectionManager` component if the user is logged in, and this allows that component to be dependent on a logged-in user, which also seems okay.

src/amo/components/Collection/index.js
```diff
@@ -276,8 +295,9 @@ export const mapStateToProps = (
 
   return {
     collection,
-    loading,
     hasEditPermission,
+    isLoggedIn: !!state.users.currentUserID,
```
bobsilverberg: I'm not sure if there is a better/more standard way of doing this.

src/amo/components/CollectionManager/index.js
```diff
@@ -88,16 +99,30 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       return;
     }
 
-    dispatch(updateCollection({
-      collectionSlug: collection.slug,
-      defaultLocale: collection.defaultLocale,
+    if (!this.state.slug) {
```
bobsilverberg: I was going to gate this on `creating` too, but I don't think submitting a form without a `slug` is valid for either edit or create. If a user doesn't change the slug on the edit form, the current value will still be in `this.state.slug`, will it not?

src/amo/components/CollectionManager/index.js
```diff
@@ -88,16 +99,30 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+      errorHandler.addMessage(
+        i18n.gettext('Collection slug cannot be blank')
+      );
+      return;
+    }
+
+    const payload: Object = {
```
bobsilverberg: I tried to make this a type more specific than `Object` but Flow had me running in circles. Again it's a problem with all the optional/mandatory types needed for the two different actions. I tried to solve it by adding calls to `invariant` into the code here for the non-optional parameters, but this time that didn't work. Flow still got mad about some parameters possibly being null, even though I guaranteed they wouldn't be with `invariant`.  Perhaps that was because I have an object here, `payload`, so my calls to `invariant` looked like `invariant(payload.slug)` as opposed to just `invariant(slug)`, and maybe Flow couldn't figure out what I was doing.

src/amo/components/CollectionManager/index.js
```diff
@@ -88,16 +99,30 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+      collectionSlug: (collection && collection.slug) || undefined,
+      defaultLocale: (collection && collection.defaultLocale) || siteLang,
```
bobsilverberg: I'm assuming here that `siteLang` is what we want to set the initial `defaultLocale` to.

src/amo/components/CollectionManager/index.js
```diff
@@ -88,16 +99,30 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+        creating ? CREATE_COLLECTION_OVERLAY : EDIT_COLLECTION_OVERLAY,
       name: { [siteLang]: this.state.name },
-      user: collection.authorUsername,
+      user: (collection && collection.authorUsername) || currentUsername,
       slug: this.state.slug,
-    }));
+    };
+
+    if (creating) {
+      dispatch(createCollection(payload));
```
bobsilverberg: FTR, it was these calls that Flow was upset about when I tried to use a more specific type above. And adding `invariant` into the different cases for different mandatory parameters in here didn't help.

src/amo/components/CollectionManager/index.js
```diff
@@ -118,25 +143,23 @@ export class CollectionManagerBase extends React.Component<Props, State> {
   }
 
   render() {
-    const { collection, errorHandler, i18n, siteLang } = this.props;
-
-    let collectionUrlPrefix = '';
-    if (collection && siteLang) {
```
bobsilverberg: We have an `if` statement above that will prevent the code from getting this far if there is no `collection` (and we're editing) and if there is no `siteLang`, so this statement seemed superfluous.

src/amo/components/CollectionManager/index.js
```diff
@@ -189,9 +212,11 @@ export const extractId = (ownProps: Props) => {
   return `collection-${collection ? collection.slug : ''}`;
 };
 
-export const mapStateToProps = (state: {| api: ApiStateType |}) => {
+export const mapStateToProps = (state: {| api: ApiStateType, users: UsersStateType, |}) => {
+  const currentUser = getCurrentUser(state.users);
   return {
     siteLang: state.api.lang,
+    currentUsername: currentUser && currentUser.username,
```
bobsilverberg: All we need is the current user's username, in the case of `create`, so this seemed like a reasonable way to get it. I suppose another option would have been to pass the username in as a property from the caller. I'm not sure which is best in this situation.

tests/unit/amo/components/TestCollection.js
```diff
@@ -590,7 +593,22 @@ describe(__filename, () => {
...
 
     sinon.assert.calledWith(
-      dispatchSpy, openFormOverlay(COLLECTION_OVERLAY)
+      dispatchSpy, openFormOverlay(EDIT_COLLECTION_OVERLAY)
+    );
+  });
+
+  it('can open the Collection manager for creating', () => {
+    // Turn on the edit-overlay feature.
+    const fakeConfig = getFakeConfig({ enableCollectionEdit: true });
```
bobsilverberg: I realized I don't need much of this, and I also added a test case for the `NotAuthorized` path, but I'm not going to push those changes yet, as I don't want them to hide all the comments I already added.

kumar303:
Hi Bob. The form overlay was causing too many problems so we decided to remove it in https://github.com/mozilla/addons-frontend/issues/4623 There's a PR up for it and it will create some conflicts with your patch. The `CollectionManager` is more or less the same.

Also, without the overlay, I think we might need to bring back the `collectionUpdates` state (but it needs work). https://github.com/mozilla/addons-frontend/issues/4635

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+  renderManager() {
+    const { creating, i18n, isLoggedIn, location } = this.props;
+    if (creating) {
+      if (!isLoggedIn) {
+        return (
+          <AuthenticateButton
+            noIcon
+            location={location}
+            logInText={i18n.gettext('Sign in to create a collection')}
```
tofumatt: Do we use "Sign in" anywhere else on the site? I think this should be "Log in", for consistency.

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,71 @@
...
+          />
+        );
+      }
+      return (<CollectionManager {...this.props} creating />);
```
tofumatt: Probably don't need the parens around this, but no worries if you like 'em.

src/amo/components/CollectionManager/index.js
```diff
@@ -156,23 +179,22 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
 
-    let collectionUrlPrefix = '';
-    if (collection && siteLang) {
-      const apiHost = config.get('apiHost');
-      const { authorUsername } = collection;
-      collectionUrlPrefix =
-        `${apiHost}/${siteLang}/firefox/collections/${authorUsername}/`;
-    }
+    const collectionUrlPrefix =
+      oneLineTrim`${config.get('apiHost')}/${siteLang || ''}/firefox/collections/
```
tofumatt: When would `siteLang` ever not be set?

src/amo/routes.js
```diff
@@ -47,6 +48,12 @@ export default (
         component={CollectionEdit}
       />
       <Route path="collections/:user/:slug/" component={Collection} />
+      {config.get('enableCollectionEdit') ?
```
tofumatt: Why is adding under a feature flag (named editing) but not editing itself? Seems confusing.

src/amo/components/CollectionManager/index.js
tofumatt: I don't see a test for this either, seems like `siteLang` should just always be set.

src/amo/components/CollectionList/index.js
bobsilverberg: It's used in [`Collection`](https://github.com/mozilla/addons-frontend/blob/ceb360b6ab5358eee52baebca0f7d49202409f01/src/amo/components/Collection/index.js#L175), which is where I got the idea, but you're correct that elsewhere on the site we use "Log in".  I'll change it here and for `Collection` as well.

src/amo/components/CollectionManager/index.js
bobsilverberg: I had to do this for Flow at one point, but with the current code yeah, it's not needed - removing!

src/amo/routes.js
bobsilverberg: Kumar added the flag for enableCollectionEdit to hide the new editing screen, but we also want to hide the new creation screen for the same reason. I guess the flag is now misnamed, but I didn't think it made sense to create a second flag for adding. I'll rename it to something different that encompasses both.

bobsilverberg:
Thanks for the review @tofumatt. I've made the requested changes, and here's a screenshot of what the create screen looks like:

screenshot 2018-04-11 10 09 36

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -428,10 +486,18 @@ describe(__filename, () => {
 
     sinon.assert.calledWith(
       fakeRouter.push,
-      `/${lang}/${clientApp}/collections/${username}/${slug}/`
+      `/${newLang}/${clientApp}/collections/${username}/${slug}/`
     );
   });
 
+  it('calls router.goBack() on cancel when creating', () => {
```
tofumatt: Not sure if it's worth it to test that we *don't* call `goBack()` when not creating 🤷‍♂️

bobsilverberg: Well , we do have a test above for `redirects to the detail view on cancel when editing` which checks that `rotuer.push()` is called when not creating, which is kind of the same thing. I suppose it doesn't prove that both `router.goBack()` and `router.push()` are not called, but that doesn't seem like something we'd normally test against.

Issue #4567, opened 2018-03-15
API helper to create/update a collection add-on


kumar303:
For https://github.com/mozilla/addons-frontend/issues/4004 we need an API helper that will let us [add](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-create) an add-on to a collection or [update](https://addons-server.readthedocs.io/en/latest/topics/api/collections.html#collection-add-ons-edit) the notes for an add-on already in a collection.

PR #4568, opened 2018-03-15
API helper to create/update a collection add-on

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4567

### DIFF
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -126,11 +126,13 @@ export const getAllUserCollections = async (
   return results;
 };
 
+type CollectionAddonNotes = string | null;
+
 export type AddAddonToCollectionParams = {|
   addon: string | number,
   api: ApiStateType,
   collection: string | number,
-  notes?: string,
+  notes?: CollectionAddonNotes,
   user: string | number,
 |};
 
@@ -296,12 +298,12 @@ type ModifyCollectionAddonBaseParams = {|
 
 export type CreateCollectionAddonParams = {|
   ...ModifyCollectionAddonBaseParams,
-  notes?: string,
+  notes?: CollectionAddonNotes,
 |};
 
 export type UpdateCollectionAddonParams = {|
   ...ModifyCollectionAddonBaseParams,
-  notes: string | null,
+  notes: CollectionAddonNotes,
 |};
 
 type ModifyCollectionAddonParams =
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -313,7 +313,7 @@ type ModifyCollectionAddonParams =
 export const modifyCollectionAddon = (
   params: ModifyCollectionAddonParams,
 ): Promise<void> => {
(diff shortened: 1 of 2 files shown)


src/amo/api/collections.js
```diff
@@ -285,3 +287,79 @@ export const createCollection = ({
...
+|};
+
+type ModifyCollectionAddonParams =
+  | {| action: 'create', ...CreateCollectionAddonParams |}
+  | {| action: 'update', ...UpdateCollectionAddonParams |};
+
+export const modifyCollectionAddon = (
+  params: ModifyCollectionAddonParams,
+): Promise<void> => {
+  const { action, api, user, collectionSlug, addonId } = params;
```
bobsilverberg: Should these be in alphabetical order, or do we not care about that in this case?

src/amo/api/collections.js
```diff
@@ -285,3 +287,79 @@ export const createCollection = ({
...
+    endpoint = `${endpoint}/${addonId}`;
+  }
+
+  return callApi({ auth: true, body, endpoint, method, state: api });
+};
+
+export const createCollectionAddon = ({
+  api,
+  collectionSlug,
+  addonId,
```
bobsilverberg: These are out of order too, as are the ones below.

src/amo/api/collections.js
```diff
@@ -285,3 +287,79 @@ export const createCollection = ({
...
+  invariant(collectionSlug, 'The collectionSlug parameter is required');
+  invariant(user, 'The user parameter is required');
+
+  let method = 'POST';
+  const body = { addon: addonId, notes: params.notes || null };
+  let endpoint =
+    `accounts/account/${user}/collections/${collectionSlug}/addons`;
+
+  if (action === 'update') {
+    if (typeof params.notes === 'undefined') {
```
bobsilverberg: Why not just `params.notes === 'undefined'`?

bobsilverberg: Oops, I of course meant `param.notes === undefined`.

kumar303: Oh, heh, it is an old habit from IE6 days where referencing an undefined name would throw. I can change it.

src/amo/api/collections.js
kumar303: Personally I don't really care (I use an editor with good name finding capabilities) but I can change it. I don't feel strongly about it. I'll adjust the others too.

kumar303:
Thanks for the review. I was a bit surprised that I got [disjoint unions](https://flow.org/en/docs/types/unions/#toc-disjoint-unions) to work. I'm not sure what is different here from your approach on `modifyCollection`. It's a simpler function. Maybe the other one was triggering a subtle bug in Flow.

kumar303:
I know that including `{ action: 'created' | 'updated' }` as part of each union'd object is key but I did try making that change to your `modifyCollection` patch. It didn't seem to help.

PR #4569, opened 2018-03-15
Create a HostPermissions component to encapsulate the logic for formatting host permission strings

### LINKED ISSUE #4527: Create a HostPermissions component to encapsulate the logic for formatting host permission strings
This is a follow-up to issue #4462, specifically to https://github.com/mozilla/addons-frontend/pull/4462#discussion_r173313244, to simplify the logic in `permissions.js` for formatting permission strings by extracting the logic for host permissions into a separate component.

bobsilverberg:
Fixes #4527


bobsilverberg:
This is now blocked on #4663. I'll rebase it after that lands.

bobsilverberg:
I've rebased this so it is no longer blocked. It's ready for review @tofumatt.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+      if (match[1] === '*') {
+        allUrls = true;
+      } else if (match[1].startsWith('*.')) {
+        wildcards.push(match[1].slice(2));
+      } else {
+        sites.push(match[1]);
+      }
+    }
+
+    // Format the host permissions.  If we have a wildcard for all urls,
```
tofumatt: Nitpick, but we don't use double-spaces after fullstops in our prose. 😄

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+    this.hostPermissions = [];
+  }
+
+  permissionStrings: Object;
+  hostPermissions: Array<React.Element<typeof Permission>>;
+
+  // Add individual Permission components for individual site/domain permissions.
+  formatItems(items: Array<string>, messageType: string) {
+    for (const item of items) {
+      this.hostPermissions.push(
```
tofumatt: I'm finding it hard to run through how things are added to `this.hostPermissions` and why there are added or not because it's a side-effect of a function in some places but directly pushed into in other places.

Is there any reason we have to use `this.hostPermissions.push` rather than just creating an array and returning host permissions in these functions and adding them to an array in `render()` (or in the constructor if that's more efficient).

I don't get why we store them in an array in the object.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+|};
+
+export class HostPermissionsBase extends React.Component<Props> {
+  constructor(props: Props) {
+    const { i18n } = props;
+    super(props);
+    // These should be kept in sync with Firefox's strings for webextention
+    // host permissions which can be found in
+    // https://hg.mozilla.org/mozilla-central/raw-file/tip/browser/locales/en-US/chrome/browser/browser.properties
+    this.permissionStrings = {
```
tofumatt: I would rather this be a function that returns strings based on arguments than an array that we access. This feels less idiomatic to me and harder to read elsewhere in the component because it looks like you're accessing data rather than calling a function.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+  // that says "...plus N others".
+  addHostPermissions(
+    list: Array<string>,
+    singleMessageType:string,
+    moreMessageType:string
+  ) {
+    if (list.length < 5) {
+      this.formatItems(list, singleMessageType);
+    } else {
+      this.formatItems(list.slice(0, 3), singleMessageType);
```
tofumatt: Rather than calculate remaining and passing that as an argument to an element in another array as a description (which I'm just finding hard to follow), if we returned items rather than modified the component's shared array we could do:

```es6
const hostPermissions = this.formatItems(list.slice(0, 4), singleMessageType);

// If there are more than four items in the list, show the first three and replace the fourth with
// a "... plus N others" message.
if (list.length > 4) {
  hostPermissions[3] = this.permissionStrings({ message: moreMessageType, remaining: list.length - 3 });
}
```

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+          key="allUrls"
+        />
+      );
+    } else {
+      this.addHostPermissions(wildcards, 'wildcard', 'tooManyWildcards');
+      this.addHostPermissions(sites, 'oneSite', 'tooManySites');
+    }
+    return (
+      <React.Fragment>
+        {this.hostPermissions}
```
tofumatt: I think it would be better to have this be a variable created in the `render()` method that was modified in one function so it's easier to walk through. I think the side-effect style here/adding to the `hostPermissions` shared array is confusing.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,151 @@
...
+    if (allUrls) {
+      this.hostPermissions.push(
+        <Permission
+          type="hostPermission"
+          description={this.permissionStrings.allUrls}
+          key="allUrls"
+        />
+      );
+    } else {
+      this.addHostPermissions(wildcards, 'wildcard', 'tooManyWildcards');
```
tofumatt: While flow (in theory when used everywhere in our app) prevents us from sending the wrong magic constant to arguments by strictly defining accepted argument values, I still find this a hard API to read and it's still a magic constant. Using a constant variable instead of a string directly makes ambiguities, however rare they may be, less likely to crop up. I also think they're just more idiomatic to me as a programmer, but that might be preference rather than a real best practice.

At any rate, seeing a few string arguments like this makes it hard for me to reason about what this API does. An object as an argument might illuminate things more, eg:

```es6
addHostPermission({
  permissions: wildcards,
  messageType: 'wildcard',
});
```

I'll point out here too that I don't think we need separate "single" and "tooMany" message types. The only time you use the "tooMany" one is a code path of having too many items so you can handle that logic in that function without exposing it elsewhere, if that makes sense.

src/ui/components/HostPermissions/index.js
bobsilverberg: Yes, this is *way* better. Thanks!

bobsilverberg:
Thanks for the review @tofumatt. I've made a bunch of changes to address your comments. It may still not be perfect, but I'm interested to hear what you think of the latest version.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+type GenerateHostPermissionsParams = {|
+  permissions: Array<string>,
+  // eslint-disable-next-line react/no-unused-prop-types
+  messageType: HostPermissionMessageType,
+|};
+
+export class HostPermissionsBase extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+    this.i18n = props.i18n;
```
tofumatt: I worry about this because I don't think the `this.i18n` would get updated on props change according to this code. I can see you did it to avoid passing `i18n` around, but that defeats the purpose of using props, and I think based on the code means if we updated the `i18n` object from the `translate()` HOC won't get updated here (eventually we'd like to be able to switch languages with redux/without a full page load, we had a working POC ages ago so it's doable).

I don't think it's a good idea in general to set a prop like this to an internal variable (especially only on initialisation). I know it's more passing around variables but that's a bit of the point/model of react 🤷‍♂️. I think you should be accessing the props directly, not storing them like this... admittedly this is just a reference as it's an object so *maybe* it would get updated, but I still think it's bad form and this is just here to save passing objects from what I can see.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+        <Permission
+          type="hostPermission"
+          description={this.getPermissionString({ messageType, param: item })}
+          key={item}
+        />
+      );
+    }
+    if (permissions.length > 4) {
+      // Replace the final individual permission with a "too many" permission.
+      hostPermissions[3] = (<Permission
```
tofumatt: Surprised eslint didn't catch this, but on multi-line components we'll start one after the `(`, so this should be:

```js
hostPermissions[3] = (
  <Permission
    type="hostPermission"
    [...etc.]
  />
);
```

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+    for (const permission of permissions) {
+      if (permission === '<all_urls>') {
+        allUrls = true;
+        break;
+      }
+      if (permission.startsWith('moz-extension:')) {
+        continue;
+      }
+      const match = /^[a-z*]+:\/\/([^/]+)\//.exec(permission);
+      if (!match) {
```
tofumatt: When would this happen? These permissions are all validated by AMO, right? This seems like it should be more than an error that we continue with.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    // Format the host permissions. If we have a wildcard for all urls,
+    // a single string will suffice.  Otherwise, show domain wildcards
+    // first, then individual host permissions.
+    if (allUrls) {
+      hostPermissions.push(
+        <Permission
+          type="hostPermission"
+          description={this.getPermissionString({ messageType: 'allUrls' })}
```
tofumatt: I guess it's fine because you have flow but I always prefer things like this as constants. It makes it easier to change their values later... but it's up to you.

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+  messageType: HostPermissionMessageType,
+|};
+
+export class HostPermissionsBase extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+    this.i18n = props.i18n;
+  }
+
+  getPermissionString({ messageType, param, tooMany = false }: GetPermissionStringParams): string {
```
tofumatt: All of the mention of "too many" in this component reads like an error to me, eg: "Too many permissions" feels like an error, but it's really "too many to display so we'll truncate".

Maybe this would be better named `truncatePermissions` or `multipleDomains`?

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,166 @@
...
+            'Access your data on %(param)s other sites',
+            param), { param: this.i18n.formatNumber(param) }
+          );
+        }
+        return this.i18n.sprintf(
+          this.i18n.gettext('Access your data for %(param)s'),
+          { param }
+        );
+      default:
+        return '';
```
tofumatt: This seems like a programmer error and should throw, right? Why would this ever be blank?

src/ui/components/HostPermissions/index.js
bobsilverberg: I'm not sure what I was thinking with that. I don't even need to pass anything around. I can just access `i18n` from `this.props` in the `getPermissionString` function.

src/ui/components/HostPermissions/index.js
bobsilverberg: This can happen, but it's highly unlikely. One example of a pattern that makes it this far is the pattern `*`. 

If the pattern is invalid and wasn't caught by AMO it also won't be applied by the extension, so I think ignoring is fine in this case. This is similar code to what Firefox does when displaying permissions to a user for the dialog.

src/ui/components/HostPermissions/index.js
bobsilverberg: I agree that it's better to have constants. Flow seems to both protect us without them and also require extra code with them, but the latter is better, so I've made the change. :)

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,170 @@
...
+|};
+
+type GenerateHostPermissionsParams = {|
+  permissions: Array<string>,
+  // eslint-disable-next-line react/no-unused-prop-types
+  messageType: HostPermissionMessageType,
+|};
+
+export class HostPermissionsBase extends React.Component<Props> {
+  getPermissionString({ messageType, param, inOther = false }: GetPermissionStringParams): string {
```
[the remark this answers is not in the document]
tofumatt: I think maybe `multiple` would work if you want to be generic, `multipleSites` if you don't mind saying "sites" for domains+sites (my preferences, technically they're all sites), and `multipleDomainsOrSites` if you want to be painfully clear 😉

src/ui/components/HostPermissions/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+  i18n: I18nType;
+
+  // Generates Permission components for a list of host permissions. If we have 4 or
+  // fewer, display them all, otherwise display the first 3 followed by an item
+  // that says "...plus N others".
+  generateHostPermissions({
+    permissions, messageType,
+  }: GenerateHostPermissionsParams): Array<React.Element<typeof Permission>> {
+    const hostPermissions = [];
```
tofumatt: This is so much cleaner and easier to follow, I love it. Thanks! 👍

src/ui/components/HostPermissions/index.js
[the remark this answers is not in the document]
bobsilverberg: I was going to make it `inXOther` because the messages are `Access your data in %(param)s other domains` and `Access your data on %(param)s other sites`, but I didn't like the way `inXOther` looked, so it became just `inOther`, which I admit is pretty meaningless.

I'll just go with `multiple`.

Issue #4281, opened 2018-01-31
Show "Themes created" on the user profile page


tofumatt:
Create a card that shows themes a user has created on the profile.

<img width="482" alt="screenshot 2018-01-31 14 53 44" src="https://user-images.githubusercontent.com/90871/35629455-8db3d58a-0696-11e8-91f3-8fe0bb5aa808.png">

ioanarusiczki:
@tofumatt I've been doing some testing on AMO dev with FF59 (Win10).

I have three questions:

1. The  "More Themes by DevTesting" is displayed in both cases: user view and public view. I don't know if the text "Themes I've created" is expected to be displayed as in the card displayed above and also as it is mentioned in the tracker https://github.com/mozilla/addons-frontend/issues/2749 
2. A number of six themes appears which is fine but only after hovering over the theme the rating stars are shown.  I don't know if that is as expected or if the ratings and number of users should be displayed simultaneously as above.
3. Add to button is not available yet but I guess this is an upcoming implementation.
4. I filed https://github.com/mozilla/addons-frontend/issues/4796 and https://github.com/mozilla/addons-frontend/issues/4797 for what I've noticed while testing.

tofumatt:
1. Eventually yes, but that should be a follow-up issue. It's related to #4774 
2. This is intentional.
3. Correct, this will be a lot of work to implement properly actually :)
4. 👍

ioanarusiczki:
It is verified fixed on AMO dev with FF59 (Win10).

Issue #3957, opened 2017-11-22
Refactor the addonsByAuthors reducer


willdurand:
The current implementation of the `addonsByAuthors` reducer stores all the related add-ons of a given add-on (technically, these "other add-ons" are those also created by any of the author(s) of that add-on).

Once #3927 is merged, the `addons` reducer will be more usable and flexible. Therefore, we could refactor the `addonsByAuthors` reducer as follows.

## Proposal

Instead of storing the entire add-ons in the `addonsByAuthors` reducer, we can add them to the `addons` reducer and only keep a list of `id` in the `addonsByAuthors` reducer:

```
"addonsByAuthors": {
  "byAddonSlug": {
    "add-on-slug": [123, 456],
    "addon-without-others": []
  }
}
```

Once we get the list of IDs, we can retrieve the add-ons by using the `getAddonByID()` selector. We could also introduce a selector to return an array of add-ons given a list of IDs.

willdurand:
I believe this is not up-to-date anymore.

PR #4571, opened 2018-03-15
chore: Allow fetchAddonsByAuthors to search for add-ons without specifying an `exclude_addon` param

### LINKED ISSUE #4556: Allow fetchAddonsByAuthors to omit the exclude_addons param

tofumatt:
Part of #4280.

Fixes #4556.

Allows `addonsByAuthor` to fetch an author's add-ons without an `exclude_addon` prop sent to the API. This allows a user's add-ons to be show on the user profile without needing to write a whole new reducer; it just makes the slug here optional and names it a bit more clearly.

It also stores add-ons fetched by this action in a userId-keyed dictionary, which will be used for showing a user's extensions (#4280) and themes (#4281) on their profile page.


src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,6 +1,8 @@
 /* @flow */
-import type { AddonType, ExternalAddonType } from 'core/types/addons';
+import invariant from 'invariant';
```
willdurand: oh cool 😍 when did you introduce it?

tofumatt: Right?! Very recently, and it’s just piecemeal now but it’s very nice. A lot fewer silly error checking tests especially!

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,123 +1,136 @@
...
+            .slice(0, ADDONS_BY_AUTHORS_PAGE_SIZE)
+            .map((addon) => createInternalAddon(addon))
+        };
+      }
+
+      const addons = action.payload.addons
+        .map((addon) => createInternalAddon(addon));
+
+      for (const addon of addons) {
+        if (addon.authors) {
```
tofumatt: Flow required this check, even though I didn't think an addon could have undefined authors in this case. Maybe I'm missing something.

tofumatt:
I know we're storing a fair bit of possibly duplicate data in this reducer... it is partial search result data though so we don't want to put it in `state.addons`.

A possible refactor here would be to set the `byAddonSlug` and `byUserId` values to just the add-on ID and put all the actual add-ons in their own dictionary.

src/amo/reducers/addonsByAuthors.js
bobsilverberg: I believe that's based on the types, so wherever the type is defined for `addon` it must say that `authors` is optional/can be null.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,139 @@
...
 
 const reducer = (
   state: State = initialState,
   action: Action
 ): State => {
   switch (action.type) {
-    case FETCH_OTHER_ADDONS_BY_AUTHORS:
+    case FETCH_ADDONS_BY_AUTHORS:
+      if (!action.payload.excludeAddonBySlug) {
+        return state;
```
kumar303: Don't you need to reset the `byUserId` state by making empty arrays for each author that is being fetched? That's what we'd typically do. I think you may need a new `byUserName` mapping for this to work, which I've explained in detail below.

src/amo/reducers/addonsByAuthors.js
kumar303: > Flow required this check, even though I didn't think an addon could have undefined authors in this case. 

It's possible that our `ExternalAddonType` is wrong (or maybe it's trying to account for themes?). The [docs](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail) say that `authors` will always be an array but you may want to double check that on the real site before adjusting the type.

It might be a bit more concise to protect against empty `authors` like:

```js
for (const author of addon.authors || []) {
}
```

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,139 @@
...
+        };
+      }
+
+      const addons = action.payload.addons
+        .map((addon) => createInternalAddon(addon));
+
+      for (const addon of addons) {
+        if (addon.authors) {
+          for (const author of addon.authors) {
+            newAddons.byUserId[author.id] = {
```
kumar303: It looks like you also need a `byUserName` mapping since the input to `fetchAddonsByAuthors()` is a string of usernames (if its param type is correct).

If so, this would be a good opportunity to put the add-ons themselves in a `byId` map and store only add-on IDs in all other mappings. However, fixing the existing `byAddonSlug` mapping would take extra effort. I'd suggest leaving that one as is (with a todo comment) and filing an issue to fix it later. You might as well design the new ones correctly though.

Ultimately, the sagas should all be dispatching an additional [`loadAddonResults()`](https://github.com/mozilla/addons-frontend/blob/5af49b851c3a1808b38fd7cf014a5298480a8d6f/src/core/reducers/addons.js#L84) action so that we can let the add-ons reducer store the actual objects but I think that would require more effort so I wouldn't bother for now.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,139 @@
...
+export const LOAD_ADDONS_BY_AUTHORS: 'LOAD_ADDONS_BY_AUTHORS'
+  = 'LOAD_ADDONS_BY_AUTHORS';
 
-type FetchOtherAddonsByAuthorsParams = {|
+type FetchAddonsByAuthorsParams = {|
   addonType: string,
   authors: Array<string>,
   errorHandlerId: string,
-  slug: string,
+  excludeAddonBySlug?: string,
```
kumar303: The original function was used was like this: "show me add-ons by other authors to display on the add-on I'm currently viewing." 

Thus, it's confusing to call this `excludeAddonBySlug` because the logic of excluding this slug from the results is just an implementation detail. I think calling it `slug` is still fine but maybe `addonSlug` would be more explicit. You could also call it `forAddonSlug` to make it more clear that you want to fetch a bunch of other add-ons for the sole purpose of this one add-on slug that the user is viewing.

It kinda makes sense to think of this as an exclusion parameter when *fetching* the add-ons but it no longer makes sense when loading the add-ons or doing any other reducer action. It still makes sense to group them in state like `addonsByAuthors.byAddonSlug` because that's saying here are some other add-ons from the perspective of add-on slug `X`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1128,19 +1133,19 @@ describe(__filename, () => {
...
+        excludeAddonBySlug: newAddon.slug,
       }));
     });
 
     it('puts "add-ons by author" in main content if type is theme', () => {
       const addon = fakeTheme;
       const { store } = dispatchAddonData({
         addon,
-        addonsByAuthors: [{ ...fakeTheme, slug: 'another-slug' }],
+        addonsByAuthors: defaultAddonsByAuthors,
```
kumar303: The old test was using `...fakeTheme` but the helper uses `...fakeAddon`.

I don't think `defaultAddonsByAuthors` is helpful because it obscures the intent of `[{ ...fakeTheme, slug: 'another-slug' }]` which was to make it clear that a *different* slug would be returned by the API.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1185,39 +1190,39 @@ describe(__filename, () => {
     it('displays the developer name when add-on is an extension', () => {
       const root = renderMoreAddons({
         addon: fakeAddon,
-        addonsByAuthors: [{ ...fakeAddon, slug: 'another-slug' }],
+        addonsByAuthors: defaultAddonsByAuthors,
```
kumar303: I don't think it's helpful to use `defaultAddonsByAuthors` anywhere in here because it could introduce subtle bugs if the slugs end up conflicting.

src/amo/reducers/addonsByAuthors.js
tofumatt: Oh shoot, yes, this is totally from an earlier, dumber refactor. But since I'm not using it anywhere yet I didn't catch it.

src/amo/reducers/addonsByAuthors.js
tofumatt: It's more concise but seems a bit less obvious to read. I feel like my `if` statement, while more verbose, is a bit more clear in setting expectation.

src/amo/reducers/addonsByAuthors.js
tofumatt: Yeah, it requires more effort and *also* extra HTTP requests. The amount of data we need for where these reducers are used is fine from the search results. It's not hard to add an extra `loadAddonResults` to a saga once we need it, but for now I thought we'd keep the request count down.

src/amo/reducers/addonsByAuthors.js
tofumatt: Yeah, I was thinking in terms of `fetch`ing but agree that it's awkward elsewhere. `forAddonSlug` is totally the right name and fits everything, merci !

tests/unit/amo/components/TestAddon.js
tofumatt: Shoot, yes, my bad.

tests/unit/amo/components/TestAddon.js
tofumatt: Yeah, I actually think this was a lazy stopgap to something better. I'll improve this.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -38,120 +38,144 @@ describe(__filename, () => {
...
     };
 
-    it('requires an error id', () => {
-      const params = getParams();
-      delete params.errorHandlerId;
-      expect(() => {
-        fetchOtherAddonsByAuthors(params);
-      }).toThrow(/An errorHandlerId is required/);
-    });
+    it('adds each add-on to each author array', () => {
```
tofumatt: This test and the next test fail with seemingly extra add-ons/authors added to data. Their ids match the defaults in `fakeAddon` and `fakeAuthor` but I'm just totally stumped as to what's happening. I think I need a second pair of eyes 👀

kumar303: This fixes it:

```diff
diff --git a/src/amo/reducers/addonsByAuthors.js b/src/amo/reducers/addonsByAuthors.js
index d7cfa3661..a9c26c691 100644
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -1,4 +1,5 @@
 /* @flow */
+import deepcopy from 'deepcopy';
 import invariant from 'invariant';
 
 import { createInternalAddon } from 'core/reducers/addons';
@@ -102,7 +103,7 @@ const reducer = (
         },
       };
     case LOAD_ADDONS_BY_AUTHORS: {
-      const newState = { ...initialState };
+      const newState = deepcopy(initialState);
 
       if (action.payload.forAddonSlug) {
         newState.byAddonSlug = {
```

The way I found out is by typing `t` (equivalent of [`-t`](https://facebook.github.io/jest/docs/en/cli.html)) and entering a regex to only run each test in isolation. They pass in isolation which made me wonder about the deep object references -- that's where the problem is.

kumar303: Ironically, I think `deepcopy` had bitrotted as a dep. We still have it in `package.json` but I can't find anything that uses it.

tofumatt: Software is hard.

Thanks a bunch for looking into this, I figured it was something like that as I noticed commenting out certain tests would change the outcome of others as well, but I was just getting nowhere with it. I owe you a 🍺 for this 🕵️ work.

tofumatt:
I think this is set for another review, then.

src/amo/components/Addon/index.js
```diff
@@ -629,7 +629,12 @@ export function mapStateToProps(state, ownProps) {
   let installedAddon = {};
 
   if (addon) {
-    addonsByAuthors = state.addonsByAuthors.byAddonSlug[addon.slug];
+    const ids = state.addonsByAuthors.byAddonSlug[addon.slug];
+    if (ids) {
+      addonsByAuthors = ids.map((id) => {
```
kumar303: It would be wise to move this to a selector function defined in the `addonsByAuthors` reducer because that makes future refactors of the state a lot easier.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,146 @@
...
+import deepcopy from 'deepcopy';
+import invariant from 'invariant';
+
 import { createInternalAddon } from 'core/reducers/addons';
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
 
 
 type State = {
+  byAddonId: { [number]: Array<AddonType> },
   byAddonSlug: { [string]: AddonType },
```
kumar303: It would be helpful to a todo comment about converting this part of the state later

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,146 @@
...
-        ...state,
-        byAddonSlug: {
-          ...state.byAddonSlug,
-          [action.payload.slug]: undefined,
-        },
-      };
-    case LOAD_OTHER_ADDONS_BY_AUTHORS:
+    case FETCH_ADDONS_BY_AUTHORS:
+      if (!action.payload.forAddonSlug) {
+        return state;
```
kumar303: You still need to reset the `byUserId` / `byUserName` state by making empty arrays for each author that is being fetched, right? Otherwise it might be hard to render a loading state.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -38,120 +38,173 @@ describe(__filename, () => {
...
+      };
+      const params = getParams({
+        addons: [firstAddon, secondAddon, thirdAddon],
+        forAddonSlug: undefined,
+      });
+
+      const newState = reducer(undefined, loadAddonsByAuthors(params));
+
+      expect(newState.byUserId).toEqual({
+        51: [firstAddon.id],
```
kumar303: It would be helpful to see `51`, `61`, and `71` replaced by short variable names.

src/amo/reducers/addonsByAuthors.js
tofumatt: Actually I converted it but flow didn't catch it 🤷‍♂️. I've updated the type.

src/amo/reducers/addonsByAuthors.js
tofumatt: Yup, this was a lazy first patch. Sorry, I've been rushing patches out lately; feeling a bit behind. I'll get better at that 😓

kumar303: I think it's fine to start with incomplete patches. A todo comment to fix it later would have been fine. It's always easier to review smaller patches!

src/amo/reducers/addonsByAuthors.js
kumar303: Huh, weird. If I change the type back to `{ [string]: AddonType }` in your latest patch I definitely see a Flow error:

```
Error: src/amo/reducers/addonsByAuthors.js:92
 92:   return ids ? ids.map((id) => {
                        ^^^ property `map`. Property not found in
 92:   return ids ? ids.map((id) => {
                    ^^^ object type


Found 1 error
```

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,165 @@
...
+
 import { createInternalAddon } from 'core/reducers/addons';
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
 
 
 type State = {
-  byAddonSlug: { [string]: AddonType },
+  // TODO: It might be nice to eventually stop storing add-ons in this
+  // reducer at all and rely on the add-ons in the `addons` reducer.
+  // That said, these are partial add-ons returned from the search
```
kumar303: Ah, right. I thought I'd requested in a previous review to reflect this in the type. No worries, I'll file a follow-up issue.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,165 @@
...
+        newState.byAddonSlug = {
+          ...newState.byAddonSlug,
+          [action.payload.forAddonSlug]: undefined,
+        };
+      }
+
+      // Reset the data for each author requested.
+      for (const authorUsername of action.payload.authors) {
+        // TODO: Reset the userId here too.
+        newState.byUsername[authorUsername] = [];
```
kumar303: `newState.byUsername[authorUsername]` should be set to `undefined` so you can know the results have *not* been fetched yet. A value of `[]` is an empty result set that was returned from the API; you'd have no way to know if it had been loaded or not.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -1,125 +1,165 @@
...
+      if (action.payload.forAddonSlug) {
+        newState.byAddonSlug = {
+          ...newState.byAddonSlug,
+          [action.payload.forAddonSlug]: undefined,
+        };
+      }
+
+      // Reset the data for each author requested.
+      for (const authorUsername of action.payload.authors) {
+        // TODO: Reset the userId here too.
```
kumar303: This seems important -- some loading screens could have subtle bugs without it. Can you file an issue so we don't forget?

src/amo/reducers/addonsByAuthors.js
tofumatt: Right on, fair point.

src/amo/reducers/addonsByAuthors.js
tofumatt: Sure thing. The issue is we only pass usernames to the reducer right now when we should be passing entire user objects. We don't use `userId` anywhere yet I don't think... I could rip it out but I thought it'd be better to keep it in as the tests and such are already written.

src/amo/reducers/addonsByAuthors.js
tofumatt: This reducer has always been storing incomplete add-ons, but moving to using `loadAddons` would be quite a large patch *and* would waste HTTP requests. In reality we need a generalised `partialAddon` reducer or something. Or maybe even a `isPartialAddon` flag in the `addons` reducer.

tofumatt: Hah, looks like @willdurand was ahead of us: #3957

PR #4572, opened 2018-03-16
Initial commit adding functional tests.

### LINKED ISSUE #4530: Add functional UI tests
@kumar303 was suggesting a suite of UI tests, similar to what is found on addons-server, but run within this repo. 

I think this is a good idea and should be pretty straightforward as the config within addons-server is all set.

jrbenny35:
Fixes #4530 

This will run as a CRON job every 2 hours on master. You can kinda see an example [here](https://circleci.com/gh/jrbenny35/addons-frontend/tree/add_functional_ui_test_config).


.circleci/config.yml
```diff
@@ -43,6 +43,25 @@ jobs:
...
+            pip install docker-compose>=1.18
+            docker-compose --version
+      - run:
+          name: Install tox
+          command: pip install tox
+      - run:
+          name: Run tests
+          command: tox -e ui-test-ci
+      - store_artifacts:
+          path: addons-server/ui-test.html
```
kumar303: Is this path correct? It doesn't seem to be given the `working_directory`

setup-functional-tests.sh
```diff
@@ -0,0 +1,11 @@
+git clone --depth 1 https://github.com/mozilla/addons-server.git
+docker-compose -f addons-server/docker-compose.yml -f docker-compose.override.yml up -d --build
+sleep 20
```
kumar303: You should replace this sleep command with something that polls for doneness (it could be fixed in a future patch). You could write a script that runs a loop until some file exists or waits for `docker images` to show the right image or `docker-compose ps` to show the right process.

setup-functional-tests.sh
```diff
@@ -0,0 +1,11 @@
+git clone --depth 1 https://github.com/mozilla/addons-server.git
+docker-compose -f addons-server/docker-compose.yml -f docker-compose.override.yml up -d --build
```
kumar303: How about renaming `docker-compose.override.yml` to `docker-compose.functional-tests.yml` or something? Also, I'd like to see this moved to a subdirectory within `tests` like `tests/ui` (if that one makes sense).

docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
+version: "2.3"
```
kumar303: It would be nice to see a comment at the top explaining that this is used for functional tests

docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
...
+
+services:
+  addons-frontend:
+    build: ../.
```
kumar303: I like how simple this override is 👍 This will share most configuration with `addons-server` which is good.

tox.ini
```diff
@@ -1,13 +1,17 @@
...
 
 [testenv:discopane-ui-tests]
 passenv = DISPLAY MOZ_HEADLESS PYTEST_ADDOPTS PYTEST_BASE_URL
 deps = -r{toxinidir}/tests/ui/requirements/requirements.txt
 install_command = pip install --no-deps {packages}
 commands = pytest --driver Firefox tests/ui/tests/test_discopane.py {posargs}
 
+[testenv:ui-test-ci]
+commands =
+  sh ./setup-functional-tests.sh
```
kumar303: It looks like this runs the tests so I'd call this `run-functional-tests.sh`

kumar303: Can you move this script to `tests/ui` ?

.circleci/config.yml
jrbenny35: Yes because the tests are run within the ```addons-server``` dir and the HTML report gets created there.

docker-compose.override.yml
jrbenny35: I will have to patch addons-server though, but that is no biggie.

jrbenny35:
So I kept the name ```setup-functional-tests``` and added the run step to tox.

.circleci/config.yml
```diff
@@ -43,6 +43,25 @@ jobs:
...
+          command: |
+            set -x
+            pip install docker-compose>=1.18
+            docker-compose --version
+      - run:
+          name: Install tox
+          command: pip install tox
+      - run:
+          name: Run tests
+          command: tox -e ui-test-ci
```
kumar303: Why do we need tox for this? Can't you just move the commands from `tox.ini` in here? What does value does `tox` add?

jrbenny35: Tox runs other commands, like the discopane tests.

kumar303: As discussed on IRC, I think this makes more sense as:

```
command: ./tests/ui/run-functional-tests.sh
```

It needs a [shebang](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang).

I am suggesting the name `run-functional-tests.sh` because then anyone can run it locally, outside of circle-ci.

tests/ui/run-functional-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
+git clone --depth 1 https://github.com/mozilla/addons-server.git
+docker-compose -f addons-server/docker-compose.yml -f tests/ui/docker-compose.functional-tests.yml up -d --build
+until docker-compose -f addons-server/docker-compose.yml -f tests/ui/docker-compose.functional-tests.yml images | grep "addonsserver_addons-frontend_1" ;
+    do printf "."; sleep 1
```
kumar303: Oh, cool. This is nice and simple. It may need a timeout in case of unexpected errors but circle-ci will already have one so let's worry about that only if we need it.

tests/ui/run-functional-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+done
```
kumar303: You probably want to add `echo` after this so that it prints a new line when it finishes.

Issue #4579, opened 2018-03-16
Add styles to view add-ons within the collection editor


kumar303:
As an initial step towards adding add-ons (https://github.com/mozilla/addons-frontend/issues/4590), deleting add-ons (https://github.com/mozilla/addons-frontend/issues/4577), and editing comments (https://github.com/mozilla/addons-frontend/issues/4578), we need to see the add-ons in a collection.

This will just implement styles for collection add-ons, without any functionality.

Here are some notes about what this issue does and does not cover:
* All collection add-ons should be visible and look like the mock
* Some add-on related buttons from the mock **will be missing**
* Clicking on collection add-ons should **do nothing**
* It will be possible to type a name of an add-on and see results
* Selecting a name from the result list **will do nothing**

AlexandraMoga:
Verified on AMO-dev, with FF59 Win10x64 and Android 7.0

Working:
- the edit collection form will display the add-ons that are part of the collection
- add-ons will include icons and names
- a search field that allows searching for add-ons is present
- search results are will show add-ons according to the searched term used (similar to AMO search)

For the rest of the issues found during testing, separate bugs were filed: #4612, #4614, #4616, #4617

PR #4580, opened 2018-03-16
Styles for viewing add-ons within collection manager

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4579 but check out that issue for a list of what this patch adds. It's pretty much just styles.

addons-frontend-collection-scroll mov
addons-frontend-auto-search mov

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -197,7 +197,9 @@ export class CollectionManagerBase extends React.Component<Props, State> {
         <div className="CollectionManager-addons">
           {collection && collection.addons && collection.addons.map(
             (addon) => {
-              return <EditableCollectionAddon addon={addon} />;
+              return (
+                <EditableCollectionAddon key={addon.id} addon={addon} />
+              );
             }
           )}
         </div>
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -4,6 +4,9 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import config from 'config';
 
+import AutoSearchInput from 'amo/components/AutoSearchInput';
+import EditableCollectionAddon
+  from 'amo/components/EditableCollectionAddon';
 import { updateCollection } from 'amo/reducers/collections';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
@@ -139,7 +142,10 @@ export class CollectionManagerBase extends React.Component<Props, State> {
         title={i18n.gettext('Edit collection')}
       >
         {errorHandler.renderErrorIfPresent()}
-        <label htmlFor="collectionName">
+        <label
+          className="CollectionManager-collectionName"
+          htmlFor="collectionName"
+        >
           {i18n.gettext('Collection name')}
         </label>
         <input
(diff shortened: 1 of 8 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -179,6 +200,24 @@ export class CollectionManagerBase extends React.Component<Props, State> {
             value={this.state.slug}
           />
         </div>
+        <AutoSearchInput
+          inputName="collection-addon-query"
+          inputPlaceholder={
+            i18n.gettext('Enter the name of an add-on to include')
```
tofumatt: I'm not sure if this text was in the mocks but I find it a bit awkward; it's a sentence fragment and while it's likely got enough context around it to make sense of it might be tough for translators ("include in what?").

I think something like "Find an add-on to include in this collection" might work better.

src/ui/components/FormOverlay/index.js
```diff
@@ -104,52 +104,52 @@ export class FormOverlayBase extends React.Component<Props> {
...
+              </Button>
+            </div>
+            <h3 className="FormOverlay-h3">{title}</h3>
+          </header>
+          <section className="FormOverlay-content">
+            {children}
+          </section>
+          <footer className="FormOverlay-footer">
+            {/*
+              type=button is necessary to override the default
```
tofumatt: Thanks, that's handy to have as a reference 👍 

Maybe we eventually want a `CancelButton` that does this for us, it seems common.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -349,6 +363,51 @@ describe(__filename, () => {
     expect(state.description).toEqual(secondCollection.description);
   });
 
+  it('renders editable collection add-ons', () => {
+    const extAddon1 = { ...fakeAddon, name: 'uBlock' };
```
tofumatt: I'm guessing `ext` stands for `extension`–I think we can just write `extension1` and `extension2`. That's clearer and isn't confusing as to whether this is an external add-on (type) or an extension add-on 😄

tofumatt: (Or if it should be `externalAddon` then I think it should be fully labelled as such; the short variable names are usually for line-length but I think they make things less obvious/readable/etc.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -349,6 +363,51 @@ describe(__filename, () => {
...
+
+    const search = simulateAutoSearchCallback({
+      root, propName: 'onSearch',
+    });
+    search({ query: 'ad blocker' });
+    // TODO: test onSearch
+    // https://github.com/mozilla/addons-frontend/issues/4590
+  });
+
+  it('handles selecting an add-on', () => {
```
tofumatt: I wonder if these tests should be set to be skipped for now? 🤷‍♂️

kumar303: We need them for test coverage to make sure each handler doesn't throw an error at the very least

tofumatt: Ah, fair point.

src/amo/components/CollectionManager/index.js
```diff
@@ -107,6 +113,18 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     this.setState({ [event.target.name]: event.target.value });
   }
 
+  // eslint-disable-next-line no-unused-vars
+  onSearchAddon = (filters: SearchFilters) => {
```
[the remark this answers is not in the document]
kumar303: Oh yeah, logging it is a good idea. That way you also know that it's working but not implemented.

Issue #4598, opened 2018-03-20
The profile picture thumbnail looks squashed on the user info's card 


ioanarusiczki:
STR:
1. Load AMO dev and go to Edit profile.
2. Select some photo samples for the profile photo. I used some .jpg and .png samples with normal sizes and dimensions.
3. Click update account and then go to My Profile.

Expected result:
The profile picture thumbnail looks alright.

Actual result:
The profile picture thumbnail is squashed. 

Notes:
This issue is reproducible on AMO dev with FF59 (Win10).

This is a 528 KB .png file, dimensions 550x336

![iguana](https://user-images.githubusercontent.com/33448286/37656569-12105062-2c51-11e8-90ba-b2fa0d78b005.gif)

And the 2.28 MB .jpg format with dimensions 2816x2112 

![sunflower](https://user-images.githubusercontent.com/33448286/37656415-aa992eb8-2c50-11e8-9892-1faa918b61c7.gif)

tofumatt:
Ah, I see. It might be better to file a new issue then as the STR are a bit different, thanks for catching that!

- Matthew Riley MacPherson
tofumatt.com

ioanarusiczki:
@tofumatt Done! it's here https://github.com/mozilla/addons-frontend/issues/4629
Verified fixed on AMO dev with FF59 (Win10).

PR #4600, opened 2018-03-20
I4598: fixes improper compression of profile avtar

tsl143:
Fixes: #4598 

Before:
screen shot 2018-03-20 at 7 26 38 pm

After:
screen shot 2018-03-20 at 7 26 15 pm


src/amo/components/UserProfile/styles.scss
```diff
@@ -20,10 +20,15 @@ $avatar-size: 64px;
 }
 
 .UserProfile-avatar {
-  display: block;
```
tsl143: wrapped `img` in a `div`, so no need of `display:block`.
The wrapper will have the height, width and nothing goes out of wrapper div, image has `width:100%` and `height:auto` this way the image will always preserve to its original dimension ratio and also respect teh wrapper size.

src/amo/components/UserProfile/styles.scss
```diff
@@ -20,10 +20,15 @@ $avatar-size: 64px;
...
   height: $avatar-size;
   margin: 0 0 12px;
+  overflow: hidden;
   width: $avatar-size;
+
+  img {
+    width: 100%;
```
tofumatt: Nitpick here, but our attributes are usually sorted alphabetically (I thought the linter would catch it 🤷‍♂️). It makes them easier to scan.

src/amo/components/UserProfile/styles.scss
```diff
@@ -20,10 +20,15 @@ $avatar-size: 64px;
 }
 
 .UserProfile-avatar {
-  display: block;
   height: $avatar-size;
   margin: 0 0 12px;
+  overflow: hidden;
   width: $avatar-size;
+
+  img {
```
tofumatt: This should have a `className` we can target rather than the HTML tag.

src/ui/components/UserAvatar/index.js
```diff
@@ -14,9 +14,14 @@ type Props = {|
...
   const _className = makeClassName('UserAvatar', className);
 
-  return (user && user.picture_type && user.picture_type.length ?
-    <img alt="" className={_className} src={user.picture_url} /> :
-    <Icon className={_className} name="anonymous-user" />
+  return (
+    <div className={_className}>
+      {
+        (user && user.picture_type && user.picture_type.length) ?
+          <img alt="" src={user.picture_url} /> :
```
tofumatt: I would add a className to target directly above. `UserAvatar-image` would be fine I think.

tests/unit/ui/components/TestUserAvatar.js
```diff
@@ -37,13 +37,13 @@ describe(__filename, () => {
...
 
-    expect(root.find('img.test')).toHaveLength(1);
+    expect(root.find('.UserAvatar.test')).toHaveLength(1);
   });
 
   it('renders extra classNames when rendering an Icon component', () => {
     const root = renderUserAvatar({ className: 'test', user: undefined });
 
-    expect(root.find(Icon).find('.test')).toHaveLength(1);
+    expect(root.find(Icon).parent().find('.test')).toHaveLength(1);
```
tofumatt: You can target `root` here instead or `root.find(Icon).parent()`

tests/unit/ui/components/TestUserAvatar.js
```diff
@@ -37,13 +37,13 @@ describe(__filename, () => {
     const user = getCurrentUser(state.users);
     const root = renderUserAvatar({ className: 'test', user });
 
-    expect(root.find('img.test')).toHaveLength(1);
+    expect(root.find('.UserAvatar.test')).toHaveLength(1);
   });
 
   it('renders extra classNames when rendering an Icon component', () => {
```
tofumatt: Actually this test can be removed entirely as we are rendering the className in the root `<div>` now and it isn't contextual based on which component we're outputting.

tsl143:
@tofumatt all set for second look :)

src/amo/components/UserProfile/styles.scss
```diff
@@ -20,10 +20,15 @@ $avatar-size: 64px;
 }
 
 .UserProfile-avatar {
-  display: block;
   height: $avatar-size;
   margin: 0 0 12px;
+  overflow: hidden;
   width: $avatar-size;
+
+  .UserAvatar-image {
```
tofumatt: This doesn't need to be nested anymore, and we tend to avoid nested selectors to keep things simple 😄

tests/unit/ui/components/TestUserAvatar.js
```diff
@@ -22,30 +22,11 @@ describe(__filename, () => {
...
     const root = renderUserAvatar({ user });
 
-    expect(root.find('img.UserAvatar')).toHaveLength(1);
-    expect(root.find('.UserAvatar')).toHaveProp(
+    expect(root.find('.UserAvatar-image')).toHaveLength(1);
+    expect(root.find('.UserAvatar-image')).toHaveProp(
       'src', 'http://tofumatt.com/photo.jpg');
   });
 
-  it('renders extra classNames when rendering an <img> tag', () => {
```
tofumatt: Oops, only one of these should have been removed. We still want to make sure our custom `className` is rendered, so you'll want a test that ensures `<UserAvatar className="foo" />` will pass `expect(root.find('.foo')).toHaveLength(1)`

tsl143:
@tofumatt done!

PR #4605, opened 2018-03-20
Add circleci version 2 config back and update jobs to run on tags.

### LINKED ISSUE #4604: Tags are no longer auto-deploying to stage
When we tag `addons-frontend` on Tuesdays, the tag is automatically deployed to stage via CircleCI. It loks like https://github.com/mozilla/addons-frontend/pull/4546 somehow broke this. We are going to revert it to restore stage. We should put the CircleCI update patch back when we can figure out the issue.

jrbenny35:
Should fix  #4604

Tagging chris for review as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@
 version: 2.0
 
 jobs:
-  build: # build for the master branch
+  build: # build for the master branch and tags
     machine: true
     working_directory: ~/addons-frontend
     steps:
@@ -100,13 +100,13 @@ workflows:
             branches:
               only: master
             tags:
-              only: /\d{4}\.\d{2}\.\d{2}/
+              only: /.*/
       - release-tag:
           requires:
               - build
           filters:
             tags:
-              only: /\d{4}\.\d{2}\.\d{2}/
+              only: /.*/
             branches:
               ignore: /.*/
       - release-master:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,127 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_USER - Login user for docker hub
+# DOCKER_PASS - Login password for docker hub user
+version: 2.0
+
+jobs:
+  build: # build for the master branch
+    machine: true
+    working_directory: ~/addons-frontend
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_USER - Login user for docker hub
+# DOCKER_PASS - Login password for docker hub user
+version: 2.0
+
+jobs:
+  build: # build for the master branch
```
kumar303: This comment is no longer accurate

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+workflows:
+  version: 2
+  build_deploy_release:
+    jobs:
+      - build:
+          filters:
+            branches:
+              only: master
+            tags:
+              only: /\d{4}\.\d{2}\.\d{2}/
```
kumar303: This regex is wrong (tags can also look like `2018.03.20-2`). I don't think it's worth trying to get it right. Let's build on every tag for now until we run into problems with it. Currently, we only use tags for releases so there shouldn't be a problem.

jrbenny35: ah, I just trusted @tofumatt. I suck at regex

tofumatt: Ah, right, I forgot about the dash suffix. But yeah, fair enough, I was being overly cautious 😉

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+
+jobs:
+  build: # build for the master branch and tags
+    machine: true
+    working_directory: ~/addons-frontend
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: sudo apt-get update; sudo apt-get install pigz
```
EnTeQuAk: Does parallel gzip really improve performance here so that installing that dependency, querying all package sources would still be faster? Querying and installing external dependencies always adds another layer of potential intermittent failures so if we don't need them we shouldn't use them

.circleci/config.yml
```diff
@@ -0,0 +1,127 @@
...
+            branches:
+              only: master
+  scheduled-workflow:
+    triggers:
+      - schedule:
+          cron: "0 0,2 * * *"
+          filters:
+            branches:
+              only:
+                - master
```
EnTeQuAk: I do think we want that for add-ons server too :heart_eyes: but both works for me really, it's just that the tests tend to be failing more than needed with intermittent failures and people tend to not look at them anymore.

PR #4611, opened 2018-03-21
chore/more addons card

tofumatt:
Adds a MoreAddonsByAuthorsCard component to replace the AddonsCard components in the Addon component and to be used for the UserProfile.


tofumatt:
This is ready for review now @bobsilverberg. This allows us to easily create more "add-ons by this author(s)" cards across the site and have multiple ones on one page.

This should basically change nothing on the existing site but the card made in the `Addon` component is now its own thing.

tofumatt:
Wait, shoot, I think I messed up a rebase.

tofumatt:
Okay, should now be good to go.

Really sorry it's a pretty massive patch, but the underlying logic in how the old component was working needed changing.

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+        authorNames: newAuthorNames,
+        forAddonSlug: newForAddonSlug,
+      });
+    }
+  }
+
+  dispatchFetchAddonsByAuthors({ addonType, authorNames, forAddonSlug }: Object) {
+    this.props.dispatch(fetchAddonsByAuthors({
+      addonType,
+      authors: authorNames,
```
bobsilverberg: This seems to be the only place where `authors` is the name of the parameter as opposed to `authorNames`. For consistency it might make sense to rename the parameter in `fetchAddonsByAuthors` in the reducer and then you wouldn't need to map `authorNames` to `authors`. I'll add a comment in the reducer as well.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -37,7 +41,7 @@ export const LOAD_ADDONS_BY_AUTHORS: 'LOAD_ADDONS_BY_AUTHORS'
   = 'LOAD_ADDONS_BY_AUTHORS';
 
 type FetchAddonsByAuthorsParams = {|
-  addonType: string,
+  addonType?: string,
   authors: Array<string>,
```
bobsilverberg: I noticed in the `MoreAddonsByAuthorsCard` component that the array of author names is consistently referred to as `authorNames`, but here it's called `authors` (and this method is dispatched from `MoreAddonsByAuthorsCard`). For consistency it might make sense to change this to `authorNames`, plus it _is_ author names, and not author objects, so that's another reason it might be good to change it.

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+          authorNames.length
+        );
+        break;
+      case ADDON_TYPE_THEME:
+        header = i18n.ngettext(
+          i18n.sprintf(
+            i18n.gettext('More themes by %(author)s'),
+            { author: authorNames[0] }
+          ),
+          i18n.gettext('More themes by these artists'),
```
bobsilverberg: Ooh, `artists`. Fancy!

src/amo/components/MoreAddonsByAuthorsCard/styles.scss
```diff
@@ -0,0 +1,59 @@
+@import "~core/css/inc/mixins";
+@import "~amo/css/inc/vars";
+@import "~ui/css/vars";
+
+.MoreAddonsByAuthorsCard {
```
bobsilverberg: Is this a mistake?

src/amo/reducers/addonsByAuthors.js
```diff
@@ -4,27 +4,31 @@ import invariant from 'invariant';
 
 import { createInternalAddon } from 'core/reducers/addons';
 import type {
-  SearchResultAddonType, ExternalAddonType,
+  AddonType, ExternalAddonType,
 } from 'core/types/addons';
 
 
-type State = {
+export type AddonsByAuthors = {|
```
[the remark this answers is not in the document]
tofumatt: Agreed, I meant to and forgot, this variable name is bad and I should feel bad 😆

src/amo/reducers/addonsByAuthors.js
```diff
@@ -4,27 +4,31 @@ import invariant from 'invariant';
...
   // TODO: It might be nice to eventually stop storing add-ons in this
   // reducer at all and rely on the add-ons in the `addons` reducer.
   // That said, these are partial add-ons returned from the search
   // results and fetching all add-on data for each add-on might be too
   // expensive.
-  byAddonId: { [number]: Array<SearchResultAddonType> },
+  byAddonId: { [number]: AddonType },
```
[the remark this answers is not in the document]
tofumatt: Our convention in other reducers is to have `byId` store the object and `byX` store an ID. It's only that though: convention.

I don't wanna bikeshed it too much but because converting them would require going through several reducers and making changes I'd say we should leave it as-is but open a code quality issue about it. You make a fair point about clarifying that it's different.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -4,27 +4,31 @@ import invariant from 'invariant';
...
   byAddonSlug: { [string]: Array<number> },
```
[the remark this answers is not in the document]
tofumatt: I think we can annotate it as `[addonId: number]: AddonType` as well... but your solution documents the code a bit more. I dig it 👍

src/amo/reducers/addonsByAuthors.js
```diff
@@ -4,27 +4,31 @@ import invariant from 'invariant';
...
 
 
-type State = {
+export type AddonsByAuthors = {|
   // TODO: It might be nice to eventually stop storing add-ons in this
   // reducer at all and rely on the add-ons in the `addons` reducer.
   // That said, these are partial add-ons returned from the search
   // results and fetching all add-on data for each add-on might be too
   // expensive.
-  byAddonId: { [number]: Array<SearchResultAddonType> },
```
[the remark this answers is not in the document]
tofumatt: I switched because otherwise I got weird flow errors and I couldn't tell it to be either `AddonType | SearchResultType` even though it's valid without errors. I'll try once more but we run `createInternalAddon` on the addon so it should adhere to the type properly. Maybe there's a `createInternalSearchResultAddon` function I need to use?

src/amo/reducers/addonsByAuthors.js
```diff
@@ -78,32 +83,86 @@ type LoadAddonsByAuthorsAction = {|
...
-    payload: { addons, forAddonSlug },
+    payload: { addons, addonType, authors, forAddonSlug },
   };
 };
 
-export const getAddonsForSlug = (state: State, slug: string) => {
+export const joinAuthorNames = (
+  authorNames: Array<string>, addonType?: string
+) => {
+  return authorNames.sort().join('-') + (addonType ? `-${addonType}` : '');
```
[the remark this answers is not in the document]
tofumatt: We do in general, but here I think it would look worse. We end up with:

```js
return `${authorNames.sort().join('-')}${(addonType ? `-${addonType}` : ''}`;
```

is that even valid? Because we want a nested template string. Maybe ES6 allows that but it looks less comprehensible to me.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -78,32 +83,86 @@ type LoadAddonsByAuthorsAction = {|
...
+  invariant(usernames && usernames.length, 'At least one username is required');
+
+  const ids = usernames.map((username) => {
+    return state.byUsername[username];
+  }).reduce((array, addonIds) => {
+    if (!addonIds || !array) {
+      return null;
+    }
+
+    for (const addonId of addonIds) {
```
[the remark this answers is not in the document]
tofumatt: Probably yeah, but I'm not sure what it buys us for this? I don't think we use `Set` elsewhere in the code (though I think we did on the linter) fwiw

src/amo/reducers/addonsByAuthors.js
```diff
@@ -115,12 +174,10 @@ const reducer = (
         };
       }
 
-      // Reset the data for each author requested.
```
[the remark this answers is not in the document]
tofumatt: Because resetting the data means you can't have multiple requests/reducer datasets for the same author on a page. In the current (pre-this-patch) implementation of this reducer, add-ons for an author were reset whenever the fetch action was dispatched. But if we want to show themes and extensions for a given user on the page we need to have the ability to add add-ons to the author's list without resetting it first. We'll still make new fetch requests when the components load so we won't "cache" stale data, but this way we can have multiple responses on a page.

The old reducer was a bit too naïve about its use-case, this needs to be removed to allow a UserProfile page with multiple cards.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -138,8 +195,16 @@ const reducer = (
       const addons = action.payload.addons
         .map((addon) => createInternalAddon(addon));
 
+      const authorNamesWithAddonType = joinAuthorNames(
+        action.payload.authors, action.payload.addonType);
+
+      newState.forAuthorNamesAndAddonType[authorNamesWithAddonType] = [];
+      newState.loadingFor[authorNamesWithAddonType] = false;
```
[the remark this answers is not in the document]
tofumatt: It doesn't matter as long as it's done in this block. All of the data is returned at once and it's a synchronous action, so it doesn't matter.

src/amo/sagas/addonsByAuthors.js
```diff
@@ -23,7 +23,7 @@ export function* fetchAddonsByAuthors({ payload }) {
       api: state.api,
       filters: {
         addonType,
-        author: authors.join(','),
+        author: authors.sort().join(','),
```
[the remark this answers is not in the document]
tofumatt: It makes the requests a bit more consistent to look at but also helps debugging; if you don't sort the authors it was more of a pain to test things because you have to replicate order everywhere which I didn't at one point and it was a PITA to track down. 😄

src/amo/sagas/addonsByAuthors.js
```diff
@@ -34,13 +34,18 @@ export function* fetchAddonsByAuthors({ payload }) {
...
+    }));
   } catch (error) {
     log.warn(`Search for addons by authors results failed to load: ${error}`);
     yield put(errorHandler.createErrorAction(error));
   }
 }
 
 export default function* addonsByAuthorsSaga() {
-  yield takeLatest(FETCH_ADDONS_BY_AUTHORS, fetchAddonsByAuthors);
+  yield takeEvery(FETCH_ADDONS_BY_AUTHORS, fetchAddonsByAuthors);
```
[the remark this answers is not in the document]
tofumatt: Yes.

src/amo/components/MoreAddonsByAuthorsCard/index.js
tofumatt: Agreed; I didn't change this because it was old code and there were other places I didn't want to change at first... but yeah this should be tidied up 👍

src/amo/components/MoreAddonsByAuthorsCard/index.js
tofumatt: bob_at_easel

src/amo/components/MoreAddonsByAuthorsCard/styles.scss
tofumatt: Hah, yes. 🔥

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
tofumatt: I was wrong; `SearchResultAddonType` here is fine. Restored!

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
tofumatt: Oh wait, right, because I wouldn't have to check using `includes`.

I'll give it a go; if flow doesn't mind because it ends up throwing around a different type I'm for it. 👍

src/amo/reducers/addonsByAuthors.js
```diff
@@ -78,32 +83,86 @@ type LoadAddonsByAuthorsAction = {|
...
+      if (!array.includes(addonId)) {
+        array.push(addonId);
+      }
+    }
+
+    return array;
+  }, []);
+
+  return ids ? (ids
+    .map((id) => {
```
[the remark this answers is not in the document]
tofumatt: I just played around with doing that and with using `Set`, but the issue is right now we can have `null` values (on purpose) and using a `Set` with flow ends up being a bit more annoying especially as `Set` doesn't have all the `Array` methods... so we end up not having a nice chain of array methods, and it's no longer easy to check for `array.includes(addonId)`. We'd largely just move around the safety checks and swap checking `includes()` for converting the `Set` back to an `Array` by moving this around.

If you think it's really gross and a sign that I'm doing it wrong though let me know and I'll look again.

tofumatt:
I've made the requested changes (except ones where I argued against 😄), so I've flagged this for another review. Let me know if anything is still unclear.

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
bobsilverberg: Great, thanks for explaining that. I think the convention is fine, if it is consistent across the repo.

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
bobsilverberg: Yeah, I figured the reason you used `+` was because the template string would get messy. I'm fine with it, just wanted to check.

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
bobsilverberg: Yes, it was to avoid the `includes` check. In fact, we could avoid a few more `includes` checks if we converted the stored objects e.g., `byAddonSlug` and `byUserId` to store a `Set` instead of an `Array`, but your comment below suggests that it's not as simple as it sounds.

It was just a suggestion and I'm fine with leaving it as is for now.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1293,25 +1128,9 @@ describe(__filename, () => {
...
-        { ...fakeTheme },
-      ];
-
-      const root = renderMoreAddons({ addon, addonsByAuthors });
-
-      expect(root).toHaveClassName('.AddonDescription-more-addons--theme');
-      expect(root).toHaveProp('addons', addonsByAuthors.map((addonByAuthor) => (
-        createInternalAddon(addonByAuthor)
-      )));
+      expect(root).toHaveProp('authorNames',
```
bobsilverberg: In this test, which is for checking the `MoreAddonsByAuthorsCard` component for an extension, I think we should also be checking that the component has the expected properties for `addonType` and `forAddonSlug`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1293,25 +1128,9 @@ describe(__filename, () => {
       const root = renderMoreAddons({ addon, addonsByAuthors });
 
       expect(root).not.toHaveClassName('.AddonDescription-more-addons--theme');
-      expect(root).toHaveProp('addons', addonsByAuthors.map((addonByAuthor) => (
-        createInternalAddon(addonByAuthor)
-      )));
-    });
-
-    it('indicates when other add-ons are themes', () => {
```
bobsilverberg: I think we still need a test case that verifies that the `MoreAddonsByAuthorsCard` receives the correct properties when the add-on is a theme.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,456 @@
...
+
+    return shallowUntilTarget(
+      <MoreAddonsByAuthorsCard {...props} />,
+      MoreAddonsByAuthorsCardBase
+    );
+  }
+
+  it('should render a card', () => {
+    const { store } = dispatchClientMetadata();
+    const authorNames = ['test', 'test2', 'test3'];
```
bobsilverberg: I feel like it's dangerous to hard code these values here, and also have them hardcoded in `fakeAddons()` above, as they could possibly get out of sync. I think it would be better to have this as a high level `const` and then use the values of that const in `fakeAddons()` which would assure that the values are always the same in both places.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,456 @@
...
+    });
+
+    sinon.assert.calledWith(dispatchSpy, fetchAddonsByAuthors({
+      addonType: ADDON_TYPE_EXTENSION,
+      authorNames: ['test2'],
+      errorHandlerId: errorHandler.id,
+    }));
+  });
+
+  it('should dispatch a fetch action if props are updated', () => {
```
bobsilverberg: There's no test for just changing the `addonType` prop and not also the `authorNames` prop (although there is one for just changing the `authorNames` prop and not the `addonType` prop), so I think that should be added. I'm not sure if it's better to make this test case longer, or have separate cases for `addonType` and `authorNames` (as you have for `forAddonSlug` below).

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,456 @@
...
+
+    root.setProps({
+      addonType: ADDON_TYPE_EXTENSION,
+      authorNames: ['test2'],
+    });
+
+    sinon.assert.notCalled(dispatchSpy);
+  });
+
+  it('shows dictionaries in header for ADDON_TYPE_DICT', () => {
```
bobsilverberg: There is a lot of repetition in the following 10 tests. It seems like a test helper method could be used to eliminate a lot of this repetition and reduce the cases to just calling a helper with particular arguments and then including a single assertion.

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
tofumatt: Pushing the entire `addon` object right in means it's harder to check with `includes()`. We'd need to do a `deepEqual` or check each value in the array with a `.some()` call on every loop.

Here we loop through a lot of stuff, narrow it down, avoid the extra loop at all if there's nothing, and only then push the entire addon into state.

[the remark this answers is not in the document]
tofumatt: Moving to a `Set` in the future would be cool, assuming it will handle complex objects, but it requires a lot of refactoring. 🤷‍♂️

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+import type { AddonType } from 'core/types/addons';
+import type { I18nType } from 'core/types/i18n';
+import type { DispatchFunc } from 'core/types/redux';
+
+import './styles.scss';
+
+
+type Props = {|
+  addons?: Array<AddonType> | null,
+  addonType?: string,
```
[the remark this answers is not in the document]
tofumatt: It's certainly possible and seemed like a fine idea, so I allowed for the possibility. We aren't doing it anywhere in the app right now but I think it's a fine idea. I guess we can make it required for now if you wanted though.

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  forAddonSlug?: string,
+  i18n: I18nType,
+  loading?: boolean,
+  numberOfAddons: number,
+
+  // AddonCards prop this component also accepts
+  showSummary?: boolean,
+  type?: 'horizontal',
```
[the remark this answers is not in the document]
tofumatt: This is actually copied from elsewhere and is just the values this component (and the component it passes things to) accepts. The idea is instead of constants we can just say the values it accepts in the type.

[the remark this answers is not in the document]
tofumatt: It does nothing with any other value than `undefined` or `'horizontal'`, so leaving this as-is will make flow complain if we pass it anything else, which is intended.

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { AddonType } from 'core/types/addons';
+import type { I18nType } from 'core/types/i18n';
+import type { DispatchFunc } from 'core/types/redux';
+
+import './styles.scss';
+
+
+type Props = {|
+  addons?: Array<AddonType> | null,
```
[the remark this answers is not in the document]
tofumatt: I thought `?` just meant "undefined is cool", not "false-y values are fine". I'll see though and will remove if it works. This feels more explicit though and I think the point of typing things is to document them as explicitly as possible, so I like this more.

src/amo/components/MoreAddonsByAuthorsCard/index.js
```diff
@@ -0,0 +1,203 @@
...
+  i18n: I18nType,
+  loading?: boolean,
+  numberOfAddons: number,
+
+  // AddonCards prop this component also accepts
+  showSummary?: boolean,
+  type?: 'horizontal',
+|};
+
+export class MoreAddonsByAuthorsCardBase extends React.Component<Props> {
```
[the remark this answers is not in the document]
tofumatt: It's also being used in the `Addon` component (in this patch), where it's showing `MoreAddonsByAuthors` but I get you. We could rename it to make things a bit clearer–I'll do that.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -4,27 +4,34 @@ import invariant from 'invariant';
...
-  byAddonId: { [number]: Array<SearchResultAddonType> },
-  byAddonSlug: { [string]: Array<number> },
-  byUserId: { [number]: Array<number> },
-  byUsername: { [string]: Array<number> },
-};
+  byAddonId: { [AddonId]: SearchResultAddonType },
+  byAddonSlug: { [string]: Array<AddonId> },
+  byUserId: { [number]: Array<AddonId> },
+  byUsername: { [string]: Array<AddonId> },
+  forAuthorNamesAndAddonType: { [string]: Array<AddonId> | null },
```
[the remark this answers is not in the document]
tofumatt: Because this is storing the add-ons found *for* a group of authorNames + addonType. But fair enough, I will change it because once I typed that I didn't buy it. 😉

src/amo/reducers/addonsByAuthors.js
```diff
@@ -78,32 +87,86 @@ type LoadAddonsByAuthorsAction = {|
...
 
   return {
     type: LOAD_ADDONS_BY_AUTHORS,
-    payload: { addons, forAddonSlug },
+    payload: { addons, addonType, authorNames, forAddonSlug },
   };
 };
 
-export const getAddonsForSlug = (state: State, slug: string) => {
+export const joinAuthorNames = (
```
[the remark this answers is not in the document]
tofumatt: Yup, originally it was just names but I added the addonType. Will update. 👍

src/amo/reducers/addonsByAuthors.js
```diff
@@ -78,32 +87,86 @@ type LoadAddonsByAuthorsAction = {|
...
+  authorNames: Array<string>, addonType?: string
+) => {
+  return authorNames.sort().join('-') + (addonType ? `-${addonType}` : '');
+};
+
+export const getLoadingForAuthorNames = (
+  state: AddonsByAuthorsState, authorNames: Array<string>, addonType?: string
+) => {
+  return authorNames && authorNames.length ?
```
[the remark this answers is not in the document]
tofumatt: It will potentially return `undefined` over `null` then and I was just trying to limit the possible return values, but fair enough.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
tofumatt: Sounds good 👍

tests/unit/amo/components/TestAddon.js
tofumatt: Good catch!

tests/unit/amo/components/TestAddon.js
tofumatt: Ah right, fair enough. I'll add that since they're displayed in different places... actually maybe I'll just put it in a function so we don't have to test it twice.

tofumatt: Actually enough things are different I'd rather just have another test, but will do.

src/amo/components/MoreAddonsByAuthorsCard/index.js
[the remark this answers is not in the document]
tofumatt: I checked and it still works with the `null` removed, but I like the explicitness. I'll leave it unless it bothers you. To me I see "optional" as different than "accepts null".

src/amo/reducers/addonsByAuthors.js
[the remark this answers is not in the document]
tofumatt: Oh, and I can do it without the check. I'll fix it!

src/amo/components/MoreAddonsByAuthorsCard/index.js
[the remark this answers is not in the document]
bobsilverberg: In that case I think it's fine the way it is.

src/amo/components/MoreAddonsByAuthorsCard/index.js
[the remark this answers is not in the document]
bobsilverberg: Ah, I assumed it could accept other types. What's the point of it accepting a prop if only one possible value is valid?

src/amo/components/MoreAddonsByAuthorsCard/index.js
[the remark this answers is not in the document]
bobsilverberg: Yes, good point. I wasn't differentiating between `null` and `undefined`, but I should.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,456 @@
...
+      authorNames,
+    }));
+
+    const root = render({
+      addonType: ADDON_TYPE_EXTENSION,
+      authorNames,
+      store,
+    });
+
+    expect(root).toHaveClassName('MoreAddonsByAuthorsCard');
```
bobsilverberg: Should we also be checking that other expected props are passed to the `AddonsCard`, like `placehholderCount`, `showSummary` and `type`?

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,456 @@
...
+      addonType: ADDON_TYPE_OPENSEARCH,
+      authorNames: [fakeAuthor.username, 'secondAuthor'],
+      errorHandler,
+      store,
+    });
+
+    expect(root.find(AddonsCard))
+      .toHaveProp('header', 'More add-ons by these developers');
+  });
+});
```
bobsilverberg: I'm not sure that we have coverage for the logic in `mapStateToProps` for `addons`, or is that covered in the test for the `Addon` component?

src/amo/components/MoreAddonsByAuthorsCard/index.js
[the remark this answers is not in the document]
tofumatt: The underlying component can have a horizontal type set or none set, and they'll be different. In the future we might end up having more types.

Arguably there should be a default set that isn't `undefined` but I think that's a code quality bug outside this patch.

tofumatt:
Thanks for the review. I've made all the changes and the tests do look a lot more compact, cheers!

It's ready for another review. If we could get this merged before the weekend it'd be aces as I could prep a few more patches, but no worries if not.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -83,59 +101,80 @@ describe(__filename, () => {
...
       }));
 
       expect(state.byAddonSlug).toMatchObject({ 'addon-slug': undefined });
-      expect(state.byUsername).toMatchObject({ author1: undefined });
+      // It should keep the add-ons by username so they can be added to
+      // as is done on the UserProfile page.
+      expect(state.byUsername).toMatchObject(firstState.byUsername);
+    });
+
+    it('sets the loading state for authorNames + addonType on fetch', () => {
```
bobsilverberg: We're missing a test that checks that `forAuthorNamesAndAddonType` is set to `null` on fetch.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -201,13 +242,74 @@ describe(__filename, () => {
...
+      });
+      let state = reducer(undefined, loadAddonsByAuthors(firstParams));
+
+      const secondParams = getParams({
+        addons: [addons.thirdAddon],
+        forAddonSlug: undefined,
+      });
+      state = reducer(state, loadAddonsByAuthors(secondParams));
+
+      expect(state.byUserId).toEqual({
```
bobsilverberg: The test description says it's for `byUsername`, but this is checking `byUserId`.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -222,10 +324,144 @@ describe(__filename, () => {
...
+      const state = reducer(undefined, loadAddonsByAuthors({
+        addons: Object.values(addons),
+        authorNames: ['fakeUsername'],
+      }));
+
+      expect(getAddonsForUsernames(state, ['nobody'])).toBeNull();
+    });
+  });
+
+  describe('getLoadingForAuthorNames', () => {
```
bobsilverberg: Should we have a test for the case where `getLoadingForAuthorNames` returns `true` after a fetch, and then returns `false` after a load?

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -222,10 +324,144 @@ describe(__filename, () => {
...
+      expect(joinAuthorNames(['author1', 'author2'], ADDON_TYPE_THEME))
+        .toEqual(`author1-author2-${ADDON_TYPE_THEME}`);
+    });
+
+    it('handles a single authorName', () => {
+      expect(joinAuthorNames(['author1'], ADDON_TYPE_THEME))
+        .toEqual(`author1-${ADDON_TYPE_THEME}`);
+    });
+  });
 });
```
bobsilverberg: There don't seem to be any tests for the `forAuthorNamesAndAddonType` state variable.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -201,13 +242,74 @@ describe(__filename, () => {
...
+        forAddonSlug: undefined,
+      });
+      state = reducer(state, loadAddonsByAuthors(secondParams));
+
+      expect(state.byUserId).toEqual({
+        [firstAuthorId]: [addons.firstAddon.id],
+        [secondAuthorId]: [addons.firstAddon.id, addons.secondAddon.id],
+        [thirdAuthorId]: [addons.thirdAddon.id],
+      });
+    });
```
bobsilverberg: There also don't seem to be tests for the `byUsername` state that verify that it is correctly populated.

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -41,14 +43,85 @@ describe(__filename, () => {
...
+      authorNames,
+      forAddonSlug: undefined,
+    });
+
+    await sagaTester.waitFor(loadActionForAddons.type);
+    mockApi.verify();
+
+    let matchingActions = 0;
+    sagaTester.getCalledActions().filter((action) => {
+      return action.type === loadActionForAddons.type;
```
bobsilverberg: I am confused by this. Won't this `filter` only allow actions with a `loadActionForAddons` type, but we are also looking for actions with a `loadActionForThemes` type? Or do both of those have the same `type`? I wonder if there's a less confusing way to write this, as it hurts my head. :(

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -41,14 +43,85 @@ describe(__filename, () => {
...
+        deepEqual(action, loadActionForThemes)
+      ) {
+        matchingActions++;
+      }
+    });
+
+    expect(matchingActions).toEqual(2);
+  });
+
+  it('calls the API multiple times for different addonType', async () => {
```
bobsilverberg: This description doesn't seem to match this test. Maybe it was meant to go with the test above?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -0,0 +1,416 @@
...
+      ...customProps,
+    };
+
+    return shallowUntilTarget(
+      <AddonsByAuthorsCard {...props} />,
+      AddonsByAuthorsCardBase
+    );
+  }
+
+  function renderAddonsWithType({ addonType, multipleAuthors = false } = {}) {
```
bobsilverberg: 👍

tests/unit/amo/reducers/test_addonsByAuthors.js
bobsilverberg: This also doesn't seem to have been addressed.

bobsilverberg:
@tofumatt They seem to be still hidden because I added them as replies to the older, already hidden comments. Stupid GitHub. 😞 If you cannot find my comments I can find them and add them back into the latest diff which should allow them to appear again.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
tofumatt: The logic in `mapStateToProps` is covered because we dispatch actions and are not using the base component directly, but instead are using the one wrapped in `connect`.

Any test using `renderAddonsWithType` dispatches loading actions and extracts the add-ons from the store using `mapStateToProps`.

That said, we could add one more that's explicitly about that as the functionality there is totally covered but all implicitly. I'll add an extra test just to be thorough and atomic.

tofumatt: Wait, scratch that, `should render a card` explicitly tests it. This is fine 😄

tests/unit/amo/reducers/test_addonsByAuthors.js
tofumatt: I don't understand, when would `forAuthorNamesAndAddonType` be `null`? Do you mean have an empty array? This comment isn't clear, but I can add a test for an empty authorName array.

tofumatt: Actually that would never happen so I don't understand this comment.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
bobsilverberg: My concern is that `mapStateToProps` calculates the value of the `addons` prop, but, as far as I can tell, none of the tests, including `should render a card` check that the `addons` prop receives the expected array. It's true that we exercise the function, but I don't see where we validate that it's actually working as expected.

tests/unit/amo/reducers/test_addonsByAuthors.js
bobsilverberg: Maybe I didn't communicate it well, it's not the entire `byAuthorNamesAndAddonType` object that gets set to null, but rather the particular key in question. It happens right here: https://github.com/mozilla/addons-frontend/pull/4611/files#diff-557fb278613a33409aa465086be9abb7R186, and we don't seem to have a test that checks that this is set to `null` upon fetch.

tests/unit/amo/components/TestMoreAddonsByAuthorsCard.js
tofumatt: Oh yeah, I get what you meant. I added a test for addons, good call.

tests/unit/amo/reducers/test_addonsByAuthors.js
tofumatt: Ah, gotcha. I've added it.

tofumatt:
More tests added, good for another review 😄

tofumatt:
Woot!

Comments are the best, no worries 😄

PR #4620, opened 2018-03-21
fix: Don't show next UserProfile when config disabled

### LINKED ISSUE #4619: View profile should not display the new user profile info card on AMO stage
STR:
1. Load AMO stage https://addons.allizom.org/en-US/firefox/ and click on the View Profile from the drop down menu.
2. Observe the page.

Expected result:
The old profile page is loaded as in [here](https://addons.allizom.org/en-US/firefox/user/Ioana-Testing_Stage/)

Actual result:
The new user profile info card is displayed. 

Notes:
This issue is reproducible on AMO stage with FF59 (Win10).
This is not reproducible when the user's name is clicked from an ad-on's detail page.

view profile on stage

tofumatt:
Fixes #4619 

I cleaned these up after an r+wc and just mixed up the props/ternary. Easy r?–I'm dumb.

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -66,7 +66,7 @@ export class HeaderBase extends React.Component {
 
     const viewProfileURL = siteUser ? `/user/${siteUser.username}/` : null;
     const viewProfileLinkProps = config.get('enableUserProfile') ?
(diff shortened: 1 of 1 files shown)


tofumatt:
It was easy not to notice on -dev as well because the nginx routing still sent the requests to the front-end, it just caused a full-page reload.

Issue #4624, opened 2018-03-21
Use a SearchResultAddonType for search results


kumar303:
As a follow-up to https://github.com/mozilla/addons-frontend/issues/4556, we should define the correct type of add-on that we get from the API to avoid confusion. This has caused bugs in the past. Using a more accurate type will help us remember not to rely on non-existent fields. The type is not documented but you can see https://github.com/mozilla/addons-server/issues/7856 for details.

PR #4625, opened 2018-03-21
Use a SearchResultAddonType for search results

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4624

### DIFF
--- a/src/core/types/addons.js
+++ b/src/core/types/addons.js
@@ -31,13 +31,11 @@ type AddonFileType = {|
   url: string,
 |};
 
-export type AddonVersionType = {|
+type PartialAddonVersionType = {|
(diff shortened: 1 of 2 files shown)


src/core/types/addons.js
```diff
@@ -46,14 +44,25 @@ export type AddonVersionType = {|
   version: string,
 |};
 
-export type AddonAuthorType = {|
+export type AddonVersionType = {|
+  ...PartialAddonVersionType,
+  // The `text` property is omitted from addon.current_version.license.
+  license: { name: string, url: string },
+  release_notes?: string,
```
kumar303: whoops, this was missing from the original type

PR #4654, opened 2018-03-28
Fixes #4648: changed anonymous profile picture size

### LINKED ISSUE #4648: The profile picture placeholder from the user info card is too small
Follow up for #4283 

Prerequisites: user has _not_ set a profile picture to their account

STR:

1. Log in to AMO-dev and select 'View profile' from the user menu
2. Observe the profile image in the user info card

Actual result:
The profile picture placeholder is much smaller than what is depicted in the [mocs](https://user-images.githubusercontent.com/90871/35629616-fcb243f4-0696-11e8-8b0b-024e96ec04cd.png)

Expected result:
The profile picture placeholder has the same dimensions as displayed in the mocs

Notes:
- reproduced on AMO-dev, with FF59, WIn10x64 and Android 7.0

image

svitlana-galianova:
Hello,
I have changed the default profile picture's size to 50px #4648. 
Here is a screenshot of what it looks like right now:
screen shot 2018-03-28 at 12 21 03 pm

Please let me know if that is the desired picture size and I will change it if required.

### DIFF
--- a/src/ui/components/Icon/styles.scss
+++ b/src/ui/components/Icon/styles.scss
@@ -80,9 +80,6 @@
 
 .Icon-anonymous-user {
   @include icon($name: "anonymous-user");
-
-  height: 50px;
-  width: 50px;
 }
 
 .Icon-user {
--- a/src/ui/components/Icon/styles.scss
+++ b/src/ui/components/Icon/styles.scss
@@ -80,6 +80,8 @@
 
 .Icon-anonymous-user {
   @include icon($name: "anonymous-user");
+  height:50px;
+  width:50px;
 }
 
 .Icon-user {
--- a/src/ui/components/Icon/styles.scss
+++ b/src/ui/components/Icon/styles.scss
@@ -80,8 +80,9 @@
 
 .Icon-anonymous-user {
   @include icon($name: "anonymous-user");
-  height:50px;
-  width:50px;
+
+  height: 50px;
+  width: 50px;
 }
 
 .Icon-user {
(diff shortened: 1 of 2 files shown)


src/ui/components/Icon/styles.scss
```diff
@@ -80,6 +80,9 @@
 
 .Icon-anonymous-user {
   @include icon($name: "anonymous-user");
+
+  height: 50px;
```
tofumatt: This is the right fix but I wonder if it's the wrong place to put it. This will override this icon size everywhere.

A default size that's larger for this icon seems fine, but I think there's already a style in the UserProfile styles.

The mocks say it should be `64px` though. If you change it to `64px` that'd be great, and in `UserProfile/styles.scss` instead of here.

svitlana-galianova: UserProfile picture container is already 64x64:
screen shot 2018-03-28 at 12 55 45 pm

It seems like icon-anonymous user is the cause of this issue
screen shot 2018-03-28 at 12 56 15 pm

tofumatt: Yes, but you should target the icon in the UserProfile SCSS. I would recommend adding a `className` to the `<Icon />` in `UserProfile` and selecting it that way. We only want this change for the UserProfile, not site-wide.

svitlana-galianova: thanks for your clarifications, misunderstood it a bit and sorry for taking your issue 😅 Next time I will check if anyone is assigned

tofumatt:
This is much better, thanks!

PR #4679, opened 2018-03-30
chore(package): update photon-colors to version 3.0.1

tofumatt:
Closes #4156


tofumatt:
Ran this locally and after modifying the import paths for the new filenames everything worked fine. Just have to make sure @bwinton includes a CHANGELOG and GitHub tags in the future so it's easier to diff different versions of this 😉

PR #4680, opened 2018-03-30
chore(package): update webpack-dev-middleware to version 3.1.0

tofumatt:
Closes #4080


tofumatt:
This breaks the dev server, because it requires webpack 4 (see #4697).

I'm going to add this dependency to that PR and remove this and the older one.

PR #4685, opened 2018-03-30
fix(package): update react-super-responsive-table to version 4.2.2

tofumatt:
Closes #3981
Closes #3680


tofumatt:
Issues resolved so this is now good. 👍 

We have to override a few of the default styles of the responsive table so our rules are a bit weird, but it looks fine.

PR #4690, opened 2018-03-31
Fixes #3231 - Hovering over a star shows the tooltip "No ratings"

### LINKED ISSUE #3231: Hovering over a star shows the tooltip "No ratings"

* On desktop, go to the detail page for an add-on that you want to rate
* Sign in
* Hover over a star that you want to click to select a rating

### What happened?

The tool tip says "No ratings" which doesn't make any sense

screenshot 2017-09-25 17 20 51

It should say "Click the star to rate the add-on" or something.

SeanPrashad:
Fixes #3231 - Let's display a more appropriate tool-tip when the user has not rated the current add-on.

**Before:**

tooltip_before

**After:**

tooltip_after

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -97,7 +97,7 @@ export class RatingBase extends React.Component {
       description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
         { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
     } else {
-      description = i18n.gettext('Click the stars to rate this add-on');
+      description = i18n.gettext('Click to rate this add-on');
     }
 
     const allClassNames = makeClassName(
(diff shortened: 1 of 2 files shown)


src/ui/components/Rating/index.js
```diff
@@ -97,7 +97,7 @@ export class RatingBase extends React.Component {
       description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
         { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
     } else {
-      description = i18n.gettext('No ratings');
+      description = i18n.gettext('Click the stars to rate this add-on');
```
tofumatt: I know this is what's in the issue but I think it was a half throwaway text suggestion. I think even just **"Click to rate this add-on"** would be better, because I think calling them stars isn't that important to screenreaders and sighted users will already see they are stars.

SeanPrashad:
Ok no problem! I think I would be able to help out with the follow-up issue so I'll keep an eye out for it.

Here's the new behaviour:

tooltip_before_postfix

PR #4693, opened 2018-03-31
Add a more intuitive tool-tip for each add-ons rating star

### LINKED ISSUE #4692: Each rating star should have its own title attribute with the rating number in it
1. Go to an add-on detail page (eg https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/)
2. Make sure you're signed in so you can leave a review
3. Hover over the rating stars to rate the add-on
4. Note the title is the same across the entire set of stars

### What happened?
The entire enclosing div (https://github.com/mozilla/addons-frontend/blob/master/src/ui/components/Rating/index.js) has the same title attribute ("Click to rate this add-on").

Each star has its own `title` attribute like `"Click to rate this add-on 1 out of 5"`.

SeanPrashad:
#Fixes #4692 - Each rating star should have its own title attribute with the rating number respective to it.

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -50,7 +50,7 @@ export class RatingBase extends React.Component {
   }
 
   renderRatings() {
-    const { readOnly, i18n } = this.props;
+    const { i18n, readOnly } = this.props;
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
@@ -64,14 +64,12 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
-        title: rating ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
+        title: (rating && !readOnly) ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
           i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
       };
 
       if (readOnly) {
-        props.title = rating ? i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
-          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) :
-          i18n.gettext(`This add-on has not been rated yet.`);
+        props.title = null;
         return <div {...props} />;
       }
 
@@ -87,11 +85,18 @@ export class RatingBase extends React.Component {
   }
 
   render() {
-    const { className, readOnly, styleSize, isOwner } = this.props;
+    const { className, i18n, isOwner, rating, readOnly, styleSize } = this.props;
     if (!RATING_STYLE_SIZES.includes(styleSize)) {
       throw new Error(
         `styleSize=${styleSize} is not a valid value; ` +
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
+    }
+
+    let description = null;
+    if (rating && readOnly) {
+      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'), { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
+    } else {
+      description = i18n.gettext('This add-on has not been rated yet.');
     }
 
     const allClassNames = makeClassName(
@@ -104,9 +109,11 @@ export class RatingBase extends React.Component {
       <div
         className={allClassNames}
         ref={(ref) => { this.element = ref; }}
+        title={description}
       >
         <span className="Rating-star-group">
           {this.renderRatings()}
+          <span className="visually-hidden">{description}</span>
         </span>
       </div>
     );
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -50,7 +50,7 @@ export class RatingBase extends React.Component {
   }
 
   renderRatings() {
-    const { readOnly } = this.props;
+    const { readOnly, i18n } = this.props;
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
@@ -64,9 +64,14 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: rating ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
+          i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
       };
 
       if (readOnly) {
+        props.title = rating ? i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) :
+          i18n.gettext(`This add-on has not been rated yet.`);
         return <div {...props} />;
       }
 
@@ -82,18 +87,11 @@ export class RatingBase extends React.Component {
   }
 
   render() {
-    const { className, i18n, rating, readOnly, styleSize, isOwner } = this.props;
+    const { className, readOnly, styleSize, isOwner } = this.props;
     if (!RATING_STYLE_SIZES.includes(styleSize)) {
       throw new Error(
         `styleSize=${styleSize} is not a valid value; ` +
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
-    }
-    let description = i18n.gettext('This add-on has no ratings.');
-    if (rating) {
-      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
-        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else if (!readOnly) {
-      description = i18n.gettext('Click to rate this add-on');
     }
 
     const allClassNames = makeClassName(
@@ -106,11 +104,9 @@ export class RatingBase extends React.Component {
       <div
         className={allClassNames}
         ref={(ref) => { this.element = ref; }}
-        title={description}
       >
         <span className="Rating-star-group">
           {this.renderRatings()}
-          <span className="visually-hidden">{description}</span>
         </span>
       </div>
     );
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -53,18 +53,19 @@ export class RatingBase extends React.Component {
   // for each individual star, as well as the wrapper
   // that surrounds the read-only set of stars
   renderTitle = (i18n, rating, readOnly, starRating) => {
-    if (!readOnly && !rating) {
-      return i18n.sprintf(i18n.gettext(
-        `Rate this add-on %(starRating)s out of 5.`), { starRating });
-    } else if (!readOnly && rating) {
+    if (readOnly) {
+      if (rating) {
+        return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
+      }
+      return i18n.gettext('This add-on has not been rated yet.');
+    }
+    if (rating) {
       return i18n.sprintf(i18n.gettext(
         `Update your rating to %(starRating)s out of 5.`), { starRating });
-    } else if (readOnly && rating) {
-      return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
-        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
     }
-    // If it's read only with no rating present
-    return i18n.gettext('This add-on has not been rated yet.');
+    return i18n.sprintf(i18n.gettext(
+      `Rate this add-on %(starRating)s out of 5.`), { starRating });
   }
 
   renderRatings() {
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -51,7 +51,7 @@ export class RatingBase extends React.Component {
 
   // Helper function used to render title attributes
   // for each individual star, as well as the wrapper
-  // that surrounds the read-only set of stars
+  // that surrounds the read-only set of stars.
   renderTitle = (rating, readOnly, starRating) => {
     const { i18n } = this.props;
 
@@ -62,10 +62,12 @@ export class RatingBase extends React.Component {
       }
       return i18n.gettext('This add-on has not been rated yet.');
     }
+
     if (rating) {
       return i18n.sprintf(i18n.gettext(
         `Update your rating to %(starRating)s out of 5.`), { starRating });
     }
+
     return i18n.sprintf(i18n.gettext(
       `Rate this add-on %(starRating)s out of 5.`), { starRating });
   }
@@ -112,7 +114,7 @@ export class RatingBase extends React.Component {
     }
 
     // Wrap read only ratings with a description to maintain functionality
-    // for the "Average rating of developer’s add-ons" tooltip
+    // for the "Average rating of developer’s add-ons" tooltip.
     const description = readOnly ? this.renderTitle(rating, readOnly, null) : null;
 
     const allClassNames = makeClassName(
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -93,10 +93,8 @@ export class RatingBase extends React.Component {
     }
 
     let description = null;
-    if (rating && readOnly) {
-      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'), { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else {
-      description = i18n.gettext('This add-on has not been rated yet.');
+    if (readOnly) {
+      description = rating ? description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'), { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) : i18n.gettext('This add-on has not been rated yet.');
     }
 
     const allClassNames = makeClassName(
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -64,12 +64,14 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
-        title: (rating && !readOnly) ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
-          i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
+        title: !readOnly ? i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }) : null,
       };
 
+      if (rating && !readOnly) {
+        props.title = i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating });
+      }
+
       if (readOnly) {
-        props.title = null;
         return <div {...props} />;
       }
 
@@ -93,8 +95,12 @@ export class RatingBase extends React.Component {
     }
 
     let description = null;
+    // Wrap readOnly ratings with a description to maintain functionality
+    // for the "Average rating of developer’s add-ons" tooltip
     if (readOnly) {
-      description = rating ? description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'), { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) : i18n.gettext('This add-on has not been rated yet.');
+      description = rating ? i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) :
+        i18n.gettext('This add-on has not been rated yet.');
     }
 
     const allClassNames = makeClassName(
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -52,7 +52,9 @@ export class RatingBase extends React.Component {
   // Helper function used to render title attributes
   // for each individual star, as well as the wrapper
   // that surrounds the read-only set of stars
-  renderTitle = (i18n, rating, readOnly, starRating) => {
+  renderTitle = (rating, readOnly, starRating) => {
+    const { i18n } = this.props;
+
     if (readOnly) {
       if (rating) {
         return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
@@ -69,7 +71,7 @@ export class RatingBase extends React.Component {
   }
 
   renderRatings() {
-    const { i18n, readOnly } = this.props;
+    const { readOnly } = this.props;
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
@@ -83,7 +85,7 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
-        title: this.renderTitle(i18n, rating, readOnly, thisRating),
+        title: this.renderTitle(rating, readOnly, thisRating),
       };
 
       if (readOnly) {
@@ -102,7 +104,7 @@ export class RatingBase extends React.Component {
   }
 
   render() {
-    const { className, i18n, isOwner, rating, readOnly, styleSize } = this.props;
+    const { className, isOwner, rating, readOnly, styleSize } = this.props;
     if (!RATING_STYLE_SIZES.includes(styleSize)) {
       throw new Error(
         `styleSize=${styleSize} is not a valid value; ` +
@@ -111,7 +113,7 @@ export class RatingBase extends React.Component {
 
     // Wrap read only ratings with a description to maintain functionality
     // for the "Average rating of developer’s add-ons" tooltip
-    const description = readOnly ? this.renderTitle(i18n, rating, readOnly, null) : null;
+    const description = readOnly ? this.renderTitle(rating, readOnly, null) : null;
 
     const allClassNames = makeClassName(
       'Rating', `Rating--${styleSize}`, className, {
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -49,6 +49,24 @@ export class RatingBase extends React.Component {
     this.props.onSelectRating(rating);
   }
 
+  // Helper function used to render title attributes
+  // for each individual star, as well as the wrapper
+  // that surrounds the read-only set of stars
+  renderTitle = (i18n, rating, readOnly, starRating) => {
+    if (!readOnly && !rating) {
+      return i18n.sprintf(i18n.gettext(
+        `Rate this add-on %(starRating)s out of 5.`), { starRating });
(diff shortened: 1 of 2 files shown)


SeanPrashad:
I've added a `title` attribute to each button that's rendered inside `renderRatings()`. Right now, it's just a matter of figuring out the correct syntax so the template renders it correctly 😅

src/ui/components/Rating/index.js
```diff
@@ -79,6 +79,7 @@ export class RatingBase extends React.Component {
         <button
           onClick={this.onSelectRating}
           value={thisRating}
+          title="Rate this add on %(thisRating) out of 5."
```
tofumatt: That's not what you'll want to do; that syntax is only for `gettext` strings. In fact, what you'll want to do here is add a `title` property to each rating above.

tofumatt: (At line 58 here, you'll want to add a title attribute with `Rate this add on ${thisRating} out of 5.`... as long as the addon is not `readOnly`.)

SeanPrashad:
I'm trying to figure out how to access the `readOnly` variable found [here](https://github.com/mozilla/addons-frontend/blob/025b2bc03dc1049ccd9c6db2fb38f48daabf018a/src/ui/components/Rating/index.js#L53).

Ideally we need to do something like:


```js
    // ...
    key: `rating-${thisRating}`,
    ref: (ref) => { this.ratingElements[thisRating] = ref; },
    title: if (readOnly) {`Rate this add-on ${thisRating} out of 5.`}, 
  };
```

src/ui/components/Rating/index.js
```diff
@@ -68,6 +68,7 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: `Rate this add-on ${thisRating} out of 5.`,
```
tofumatt: This should be <code>title: readOnly ? \`Rate this add-on ${thisRating} out of 5.\` : null,</code>

src/ui/components/Rating/index.js
```diff
@@ -93,11 +94,9 @@ export class RatingBase extends React.Component {
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
     }
     let description;
-    if (rating) {
+    if (rating && readOnly) {
```
tofumatt: Thanks for this! A comment here explaining why only `readOnly` rating wrappers should get this description might be good here.

SeanPrashad:
It looks like we're almost there but I'm trying to figure out why the `'Rated %(rating)s out of 5'` isn't rendering after I've rated an add-on.

I've installed [React Developer Tools](https://addons.mozilla.org/en-US/firefox/addon/react-devtools/) and took at look at the `title` attribute when the add-on rating has a value (and is `readOnly` at this point) and it's `undefined`. Now, I think this is due to the fact that `description` never gets initialized [here](https://github.com/mozilla/addons-frontend/pull/4693/files#diff-88b641d04bd9c7d77b5475071f0870b6R96).

This is what RDT is showing me:

react_dev_tools_title_value

I'm going to take another look at why adding `readOnly` to our `if` statement broke things!

SeanPrashad:
**Update**: RDT is very useful! 👌🏽

Upon closer inspection, RDT allows us to set the values for props to mimic specific behaviour such as when an add-on receives a rating. By setting `readOnly` to true after rating an add-on, we're able to see the expected results in action:

react_dev_tools_show_readonly

And when `readOnly` is not set to true, the tooltip behaves as expected:

react_dev_tools_show_readonly_false

Moving onto tests now!

SeanPrashad:
I think everyone is good to go @tofumatt! I've updated the OP with a few GIF's to showcase this patch but here they are once more:

_When the user is rating the add-on for the first time_:

tooltip_after_update

_After the user has submitted a rating for the add-on_:

tooltip_after_update_readonly

src/ui/components/Rating/index.js
```diff
@@ -92,12 +93,13 @@ export class RatingBase extends React.Component {
         `styleSize=${styleSize} is not a valid value; ` +
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
     }
-    let description;
-    if (rating) {
+    let description = 'Click to rate this add-on';
```
tofumatt: Any strings that we show to users need to be wrapped in a gettext call; this string will not get localised for users which would be bad.

That said: why is this being set by default? Since we could be setting a `title` attribute each rating star shouldn't we skip the title on the entire rating wrapper? I don't think this is a good message anyway, it can probably just go.

src/ui/components/Rating/index.js
```diff
@@ -68,6 +68,7 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: !readOnly ? `Rate this add-on ${thisRating} out of 5.` : null,
```
tofumatt: This needs to be translated using `i18n.gettext` and `i18n.sprintf`. You can find examples on how to use them elsewhere in the app.

SeanPrashad:
I've gone ahead and added the respective functions to internationalize our strings! I also removed the title for the wrapper `div`. In addition, I thought it would be an appropriate time to add your suggestion for when a user is rating the add-on after their first time:

react_dev_tools_show_postrating_postfeedback

Another problem though - the tooltip is shared with the smaller version of the stars, found here:

resulting_star_bug

What are your thoughts on this?

src/ui/components/Rating/index.js
```diff
@@ -68,9 +68,12 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: rating ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
```
tofumatt: You aren’t checking for readOnly any,ore, so you’ll need to add that back. That’s why the tool tips are appearing in the small, read only versions.

SeanPrashad: I moved out the check for `readOnly` to the if block that check for `readOnly` found [here](https://github.com/mozilla/addons-frontend/pull/4693/files#diff-88b641d04bd9c7d77b5475071f0870b6R76). The reason is that I could not nest ternary if statements and if I simply had `if (!readOnly && !rating) { ... }`, then tooltip would show `Update your rating to ... ` over the smaller read only stars.

This approach makes sure that the `readOnly` stars always have their `title` property removed and the clickable versions have a tooltip that dynamically reflects if the user is rating the add-on for the first or fifth time!

src/ui/components/Rating/index.js
```diff
@@ -68,9 +68,14 @@ export class RatingBase extends React.Component {
...
+          i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
       };
 
       if (readOnly) {
+        props.title = rating ? i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) :
+          i18n.gettext(`Unable to retrieve rating.`);
```
bobsilverberg: Sorry for the drive-by review, but do you think a better message might be `This add-on has not been rated yet`? The `Unable` suggests (to me at least) that some sort of error occurred.

SeanPrashad: Swapped out the error messages and updated the GIFs in the OP - thanks Bob!

SeanPrashad:
@tofumatt I'd like a review when you get a chance!

Just a quick summary of what we've done so far:

- Removed the `<div>` that wrapped all of the stars (as well as it's title)
- Updated the existing tests for the removed `<div>` to instead check the `title` of the stars
- Added 7 GIFs showcasing the various use cases that were updated

tofumatt:
@SeanPrashad Looks like you've got some merge conflicts with this now so you'll need to rebase your work on top of master first. After that I can take a look!

SeanPrashad:
☑️ Rebase complete - all ready for a review now @tofumatt!

src/ui/components/Rating/index.js
```diff
@@ -68,9 +68,14 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: rating ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
+          i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
       };
 
       if (readOnly) {
+        props.title = rating ? i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
```
tofumatt: The issue with this is it's going to set the title for each star, but if it's `readOnly` that's not needed; a `title` attribute for the entire `<div>` makes more sense.

src/ui/components/Rating/index.js
tofumatt: For this check I'd set `rating && !readOnly`, because I don't think `readOnly` stars should each have the same duplicate title; if its `readOnly`, as mentioned below, we should just put a `title` attribute on the whole enclosing `<div>`.

src/ui/components/Rating/index.js
```diff
@@ -86,19 +91,12 @@ export class RatingBase extends React.Component {
...
 
   render() {
-    const { className, i18n, rating, readOnly, styleSize, isOwner } = this.props;
+    const { className, readOnly, styleSize, isOwner } = this.props;
     if (!RATING_STYLE_SIZES.includes(styleSize)) {
       throw new Error(
         `styleSize=${styleSize} is not a valid value; ` +
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
     }
-    let description = i18n.gettext('This add-on has no ratings.');
```
tofumatt: I don't think this stuff should be removed, except for `description = i18n.gettext('Click to rate this add-on');`

That could probably be changed to set `description = null` so that the "Rate this add X out of 5" is shown for `!readOnly` add-ons.

src/ui/components/Rating/index.js
SeanPrashad: Thanks for the clarification - I actually removed the wrapping div as per your comment found [here](https://github.com/mozilla/addons-frontend/pull/4693#discussion_r178470860). Should we add it back and update it as necessary?

tofumatt: Oh snap, I'm sorry. Yeah, I think it's worth having the wrapper `title` *only* when it's `readOnly`.

SeanPrashad:
Updated as requested @tofumatt!

src/ui/components/Rating/index.js
```diff
@@ -68,9 +68,12 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: (rating && !readOnly) ? i18n.sprintf(i18n.gettext(`Update your rating to %(thisRating)s out of 5.`), { thisRating }) :
```
tofumatt: Looking at this code, if  `!rating && readOnly === true` then we will show "Rate this add-on %(thisRating)s out of 5.", which is not intended.

Did you test this out and get the desired result? This `readOnly` check seems especially superfluous given you set `props.title` to `null` later on.

src/ui/components/Rating/index.js
```diff
@@ -68,9 +68,12 @@ export class RatingBase extends React.Component {
...
+          i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }),
       };
 
       if (readOnly) {
+        props.title = null;
```
tofumatt: I'm not a huge fan of setting, then deleting props. I think it would be better to have an if statement that *adds* `title` to `props` if the right conditions are set.

src/ui/components/Rating/index.js
```diff
@@ -86,18 +89,16 @@ export class RatingBase extends React.Component {
...
-    let description = i18n.gettext('This add-on has no ratings.');
-    if (rating) {
-      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
-        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else if (!readOnly) {
-      description = i18n.gettext('Click to rate this add-on');
+
+    let description = null;
+    if (readOnly) {
+      description = rating ? description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'), { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) }) : i18n.gettext('This add-on has not been rated yet.');
```
tofumatt: This is a really long line; I'm surprised eslint didn't complain, but can we spread this out over a few lines anyway?

tests/unit/ui/components/TestRating.js
```diff
@@ -180,22 +193,35 @@ describe(__filename, () => {
...
+
+    expect(findDOMNode(root).title).toContain('This add-on has not been rated yet.');
   });
 
   it('renders an accessible description for ratings', () => {
     const root = render({ rating: 2 });
-    expect(findDOMNode(root).textContent).toEqual('Rated 2 out of 5');
+
+    [1, 2, 3, 4, 5].forEach((rating) => {
+      expect(root.ratingElements[rating].title).toEqual(`Update your rating to ${rating} out of 5.`);
```
tofumatt: I think you should be testing to make sure these are not set when `readOnly` is set and I'm not sure I see that test anywhere.

tests/unit/ui/components/TestRating.js
```diff
@@ -180,22 +193,35 @@ describe(__filename, () => {
...
+    });
+
+    const rootReadOnly = render({ rating: 2, readOnly: true });
```
tofumatt: This should be part of a different (eg its own) test.

tests/unit/ui/components/TestRating.js
SeanPrashad: I went ahead and added another test for that case, found [here](https://github.com/mozilla/addons-frontend/pull/4693/files#diff-7ac6d4cc535e144232c13e382e52eeecR224)!

src/ui/components/Rating/index.js
SeanPrashad: I've added another if statement to add, or update, the title of `props` if the user has previously rated the add-on. This still feels like it can be improved though so any suggestions are welcomed!

SeanPrashad:
All ready for another review. Here's what the use cases look like:

1. Undefined/Null Rating with read-only as true:

undefinedrating_truereadonly

2. Undefined/Null Rating with read-only as false:

undefinedrating_falsereadonly

3. Rating of 3 with read-only as true:

3rating_truereadonly

4. Rating of 3 with read-only as false:

3rating_falsereadonly

5. Checking that the smaller, read-only tooltip functions as normal:

readonlytooltip

SeanPrashad:
@caitmuenster I noticed that the original issue was closed and was wondering if I should also close this pull request?

caitmuenster:
@SeanPrashad That was totally an accident on my part and definitely my fault! Please keep the pull request open. :)

src/ui/components/Rating/index.js
```diff
@@ -68,8 +68,13 @@ export class RatingBase extends React.Component {
         id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         ref: (ref) => { this.ratingElements[thisRating] = ref; },
+        title: !readOnly ? i18n.sprintf(i18n.gettext(`Rate this add-on %(thisRating)s out of 5.`), { thisRating }) : null,
```
bobsilverberg: I think this `title` setting code would be a bit cleaner if you put both cases into a function, which would return the correct title, instead of setting a default title here, and then conditionally setting it again below.

Also, please try to keep the line length at a maximum of 80 characters.

SeanPrashad:
I've moved the logic for rendering a title into `renderTitle` and included a short comment about what the function is used for @bobsilverberg! I also rebased on master as there were some new tests added which conflicted with some of this patch.

I particularly don't like how messy the function's contents are so please let me know how I can best clean it up (if possible) 👍🏽

src/ui/components/Rating/index.js
```diff
@@ -82,18 +101,18 @@ export class RatingBase extends React.Component {
...
-    if (rating) {
-      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
-        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else if (!readOnly) {
-      description = i18n.gettext('Click to rate this add-on');
+
+    let description = null;
+    // Wrap readOnly ratings with a description to maintain functionality
+    // for the "Average rating of developer’s add-ons" tooltip
+    if (readOnly) {
```
bobsilverberg: This could be shortened to `const description = readOnly ? this.renderTitle(i18n, rating, readOnly, null) : null;`

src/ui/components/Rating/index.js
```diff
@@ -49,8 +49,26 @@ export class RatingBase extends React.Component {
     this.props.onSelectRating(rating);
   }
 
+  // Helper function used to render title attributes
+  // for each individual star, as well as the wrapper
+  // that surrounds the read-only set of stars
+  renderTitle = (i18n, rating, readOnly, starRating) => {
```
bobsilverberg: This could be cleaned up a bit. For one, you could remove all the `else if` statements, because each `if` returns something, so there's no need for any `else`s.

You could also try grouping it a bit, for example:

```js
  renderTitle = (i18n, rating, readOnly, starRating) => {
    if (readOnly) {
      if (rating) {
        return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
      }
      return i18n.gettext('This add-on has not been rated yet.');
    }
    if (rating) {
      return i18n.sprintf(i18n.gettext(
        `Update your rating to %(starRating)s out of 5.`), { starRating });
    }
    return i18n.sprintf(i18n.gettext(
      `Rate this add-on %(starRating)s out of 5.`), { starRating });
  }
```

tests/unit/ui/components/TestRating.js
```diff
@@ -142,13 +142,50 @@ describe(__filename, () => {
...
+
   it('rounds ratings to nearest 0.5 multiple', () => {
     // This should be treated like a rating of 3.5 in text.
-    const root = render({ rating: 3.60001 });
+    const root = render({ rating: 3.60001, readOnly: true });
 
     expect(findDOMNode(root).title).toContain('3.6 out of 5');
   });
 
+  it('converts rating numbers to a float', () => {
```
bobsilverberg: We don't have any code that explicitly converts to float, so we don't really need this test. I believe a similar one was removed in an earlier commit.

tests/unit/ui/components/TestRating.js
```diff
@@ -142,13 +142,50 @@ describe(__filename, () => {
     verifyRating(root);
   });
 
+  it('renders an appropriate title when updating a rating', () => {
+    const root = render({ rating: 3 });
+
+    [1, 2, 3].forEach((rating) => {
+      expect(root.ratingElements[rating].className).toEqual('Rating-choice Rating-selected-star');
```
bobsilverberg: Because this test is just for the `title`, we don't really need to assert anything about the `className` of the rating element.

tests/unit/ui/components/TestRating.js
```diff
@@ -142,13 +142,50 @@ describe(__filename, () => {
...
+      expect(root.ratingElements[rating].title).toEqual(`Update your rating to ${rating} out of 5.`);
+    });
+    [4, 5].forEach((rating) => {
+      expect(root.ratingElements[rating].className).toEqual('Rating-choice');
```
bobsilverberg: As above, we don't really need this assertion.

tests/unit/ui/components/TestRating.js
```diff
@@ -142,13 +142,50 @@ describe(__filename, () => {
...
 
+  it('converts rating numbers to a float', () => {
+    const rootWithInteger = render({ rating: 3, readOnly: true });
+    const rootWithString = render({ rating: '3.60001', readOnly: true });
+
+    expect(findDOMNode(rootWithInteger).title).toContain('3 out of 5');
+    expect(findDOMNode(rootWithString).title).toContain('3.6 out of 5');
+  });
+
+  it('rounds readOnly average ratings to nearest 0.5 multiple', () => {
```
bobsilverberg: This test was replaced with two others above it, specifically the ones for `it('renders correct full stars for a rating')` and `it('renders correct half stars for a rating')`, so this test can be removed.

src/ui/components/Rating/index.js
```diff
@@ -49,8 +49,27 @@ export class RatingBase extends React.Component {
...
+      return i18n.sprintf(i18n.gettext(
+        `Update your rating to %(starRating)s out of 5.`), { starRating });
+    }
+    return i18n.sprintf(i18n.gettext(
+      `Rate this add-on %(starRating)s out of 5.`), { starRating });
+  }
+
   renderRatings() {
-    const { readOnly } = this.props;
+    const { i18n, readOnly } = this.props;
```
bobsilverberg: Oops, sorry I missed this last time, but I don't think you need to accept extract `i18n` from `this.props` here, only to pass it into `this.renderTitle`. You could just add a `const { i18n } = this.props;` at the top of `renderTitle`, I believe.

SeanPrashad: I went ahead and implemented it but it broke about 8 tests (which is probably too large to paste here). Should I go ahead and refactor them?

willdurand:
Hi @SeanPrashad, is it ready for another look? thanks!

SeanPrashad:
Hey @willdurand! Yes it should be ready to go - @bobsilverberg and I were able to polish up one final thing recently.

src/ui/components/Rating/index.js
```diff
@@ -49,6 +49,27 @@ export class RatingBase extends React.Component {
     this.props.onSelectRating(rating);
   }
 
+  // Helper function used to render title attributes
+  // for each individual star, as well as the wrapper
+  // that surrounds the read-only set of stars
```
willdurand: Nitpick: missing period.

src/ui/components/Rating/index.js
```diff
@@ -49,6 +49,27 @@ export class RatingBase extends React.Component {
...
+        return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
+      }
+      return i18n.gettext('This add-on has not been rated yet.');
+    }
+    if (rating) {
+      return i18n.sprintf(i18n.gettext(
+        `Update your rating to %(starRating)s out of 5.`), { starRating });
+    }
+    return i18n.sprintf(i18n.gettext(
```
willdurand: could you add a blank line before this `return`?

src/ui/components/Rating/index.js
```diff
@@ -49,6 +49,27 @@ export class RatingBase extends React.Component {
...
+    const { i18n } = this.props;
+
+    if (readOnly) {
+      if (rating) {
+        return i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5.'),
+          { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
+      }
+      return i18n.gettext('This add-on has not been rated yet.');
+    }
+    if (rating) {
```
willdurand: could you add a blank line before this `if`?

src/ui/components/Rating/index.js
```diff
@@ -82,19 +104,16 @@ export class RatingBase extends React.Component {
...
-    let description = i18n.gettext('This add-on has no ratings.');
-    if (rating) {
-      description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
-        { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else if (!readOnly) {
-      description = i18n.gettext('Click to rate this add-on');
-    }
+
+    // Wrap read only ratings with a description to maintain functionality
+    // for the "Average rating of developer’s add-ons" tooltip
```
willdurand: Nitpick again: missing period.

tests/unit/ui/components/TestRating.js
```diff
@@ -142,9 +142,20 @@ describe(__filename, () => {
...
   });
 
+  it('renders an appropriate title when updating a rating', () => {
+    const root = render({ rating: 3 });
+
+    [1, 2, 3].forEach((rating) => {
+      expect(root.ratingElements[rating].title).toEqual(`Update your rating to ${rating} out of 5.`);
+    });
+    [4, 5].forEach((rating) => {
+      expect(root.ratingElements[rating].title).toEqual(`Update your rating to ${rating} out of 5.`);
```
willdurand: I don't understand why there are two arrays here.

tests/unit/ui/components/TestRating.js
```diff
@@ -171,22 +182,40 @@ describe(__filename, () => {
...
     const root = render({ rating: 0 });
-    expect(findDOMNode(root).textContent).toEqual('Click to rate this add-on');
+
+    [1, 2, 3, 4, 5].forEach((rating) => {
+      expect(root.ratingElements[rating].title).toEqual(`Rate this add-on ${rating} out of 5.`);
+    });
   });
 
-  it('renders "no ratings" if no rating and readOnly', () => {
+  it('renders an accessible description for null ratings and readOnly', () => {
```
willdurand: s/readOnly/read only (or read-only)

tests/unit/ui/components/TestRating.js
tofumatt: That is a holderover from the old test I think, but if the rating is 3 this should be updated to have "Update your rating to X" for everything except 3, and then a "Your rating is currently 3/5" or even just no title for 3...

SeanPrashad: @tofumatt is right - here's the old test for reference: https://github.com/mozilla/addons-frontend/blob/master/tests/unit/ui/components/TestRating.js#L195

SeanPrashad:
I've addressed the nits and moved the read-only tests into the read-only section. I noticed that there was one tests that was repeated so I removed that one.

tests/unit/ui/components/TestRating.js
```diff
@@ -253,22 +244,66 @@ describe(__filename, () => {
...
 
     it('localizes average rating', () => {
       const i18n = fakeI18n({ lang: 'de' });
-      expect(getRating({ rating: 3.5, i18n })).toContain('3,5');
+      expect(getRating({ rating: 3.5, i18n, readOnly: true })).toContain('3,5');
     });
 
     it('renders empty ratings', () => {
-      expect(getRating({ rating: null })).toEqual('Click to rate this add-on');
+      expect(getRating({ rating: null, readOnly: true })).toEqual('This add-on has not been rated yet.');
```
willdurand: You should try to wrap lines at max 80 chars.

SeanPrashad: Thanks for that - I am wondering why linting isn't giving me an error.. locally or via Travis.

willdurand: I dont think we have a rule for that.

SeanPrashad: It might be worthwhile to use [Prettier pre-commit hooks](https://prettier.io/docs/en/precommit.html) to automatically format code  when commits are made based on the existing `eslint` config.

bobsilverberg:
Thanks for all of your hard work on this @SeanPrashad. It looks like @willdurand took a final look and gave it his seal of approval, so I'm going to merge it.

TonnesM:
Please remember to not use trailing periods for action tooltips, as they make a string instructive/imperative rather than infinitive, which will reflect in a number of localizations, but may also affect the meaning for en-US. (Localizers sometimes fully depend on periods being present or not in order to see what’s intended rather than testing them.)

I.e. they would read

"Rate this add-on / Update your rating to %(thisRating)s out of 5!"

instead of 

(User can or wants to) "Rate this add-on / Update their rating to %(thisRating)s out of 5" (by clicking the star)

A change/fix in any next push would be welcome.

bobsilverberg:
@SeanPrashad Do you want to open a follow-up issue for the change suggested above?

SeanPrashad:
Yes! I'll open another PR when I get home from work later today.

PR #4697, opened 2018-03-31
chore(package): update webpack to version 4.x

tofumatt:
Closes #4450


tofumatt:
Worked on this a bit and hacked together things working with webpack 4.1 (we can't go all the way o 4.4.x because of dependency problems IIRC), but locales aren't being loaded.

Pretty much everything else works and we lazy-load locales so I think it's just a matter of changing how we do that.

tofumatt:
FYI: This PR had a bit of prior art in it that might be useful for the 4.x upgrade.

Issue #4652, opened 2018-03-28
Collection name can be left blank by entering only blank spaces in the Name field in the edit collection form


AlexandraMoga:
Follow up for #4623 

STR:

1, Log in to AMO dev and select 'View My Collections' from the user menu
2. Choose a collection to view its details page
3. Click on Edit collection underneath the collection info card
4. Edit the collection name by entering only blank spaces
5. Click on Save Collection

Actual result:
Collection is successfully saved with a blank name

Expected result:
A collection should not be saved when the Name field contains only blank spaces

Notes:
 - reproduced on AMO-dev, with FF59, Win10x64

![blank col name](https://user-images.githubusercontent.com/31961530/38031709-8f51b1ca-32a4-11e8-83b2-4ee236ab974d.gif)

muffinresearch:
This should be validated on the frontend but this also probably needs a API counterpart to prevent it there too.

AlexandraMoga:
Verified as fixed on AMO-dev, FF59, Win10x64 and Android 7.0

![coll blank name](https://user-images.githubusercontent.com/31961530/38188537-0faf5ae8-3665-11e8-96a5-c973d9f897a5.gif)

bobsilverberg:
#4698 is a nice fix, but I am also concerned that the validation in `onSubmit` did not catch this. @tofumatt do you think it's enough to disable the `Save Collection` button, or should we also make sure that there's working validation for this case in `onSubmit`?

tofumatt:
Hmmm... I think us disabling the form is good enough but if the server allows empty collection names we could check... but in the future I expect someone who actually managed to still get `onSubmit` to fire would get an API error.

But at that point it feels like an API bug our error handler would just catch. They can still submit empty data to the API if they really want, we are already stopping all but the most ardent Devtools Inspector punks. 😉

If you wanted to open a new code quality issue for it though, feel free.

PR #4698, opened 2018-04-01
I4652 - disbales submit button if no name

tsl143:
Fixes: #4652 

Before:
screen shot 2018-04-01 at 6 29 26 pm

After: 
screen shot 2018-04-02 at 8 09 21 am

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -170,7 +170,8 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     // https://github.com/mozilla/addons-frontend/issues/4635
     // The collectionUpdates state will handle this but it needs
     // to be hooked up the saga.
-    const formIsDisabled = (!collection || !name.trim());
+    const formIsDisabled = !collection;
+    const isNameBlank = !(name && name.trim().length);
 
     return (
       <form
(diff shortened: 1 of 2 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -169,7 +170,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     // https://github.com/mozilla/addons-frontend/issues/4635
     // The collectionUpdates state will handle this but it needs
     // to be hooked up the saga.
-    const formIsDisabled = !collection;
+    const formIsDisabled = (!collection || !name.trim());
```
tofumatt: I guess the tests pass with this method but I would have thought we needed to test for `.length` here, as `''.trim()` is still truthy.

tsl143:
I assumed, we need to replicate same behavior for cancel as submit. I will change this :)

src/amo/components/CollectionManager/index.js
tsl143: `''.trim()` is falsey, isnt it?

tsl143: image

tsl143:
@tofumatt , updated :)

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -256,6 +256,16 @@ describe(__filename, () => {
     }));
   });
 
+  it('disables submit button when the name is blank', () => {
+    const root = render();
+
+    // Enter in a blank collection name.
+    typeInput({ root, name: 'name', text: '' });
+
+    expect(root.find('.CollectionManager-submit'))
```
tofumatt: Might be nice to add an extra `expect` statement here that makes sure the cancel button *isn't* disabled.

tsl143:
@tofumatt updated test case.

Issue #4645, opened 2018-03-28
Search suggestions list in edit collection form is showing only 4 add-ons and ends abruptly 


AlexandraMoga:
Follow up for #4623 

STR:

1. Log in to AMO-dev and click on a collection to view its details page (i.e. https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/password-managers/)
2. Click on Edit collection option
3. In the search field, enter a search term 
4. Observe the list with add-on suggestions

Actual result:
There are only 4 add-ons displayed and the list looks cut off

Expected result:
The search suggestion list should display more add-ons and should fit entirely inside the edit collection form

Notes:
- on mobile, there are only 3 suggestions displayed
- reproduced on AMO-dev, FF59, WIn10x64 and Android 7.0

**[1] desktop**
![image](https://user-images.githubusercontent.com/31961530/38011340-ba68f560-3265-11e8-9a84-d81aec2ff300.png)

**[2] mobile**
![image](https://user-images.githubusercontent.com/31961530/38011449-2ab788d6-3266-11e8-80db-35c5174e9c83.png)

tofumatt:
Closed by #4699

AlexandraMoga:
Verified as fixed on AMO-dev, FF59, Win10x64 and Android 7.0

![search coll list](https://user-images.githubusercontent.com/31961530/38257971-059a282a-376a-11e8-86a1-a2e431e634a1.gif)

PR #4699, opened 2018-04-01
I4645 - fix autocomplete height in collection

tsl143:
#4645 

The `overflow` of `card-contents` is hidden that causes the issue, limiting `AutoSearchInput` to `max-height: 90px` that is 3 addon.

Before:
screen shot 2018-04-01 at 7 27 00 pm

After:
screen shot 2018-04-02 at 8 39 03 pm

### DIFF
--- a/src/amo/components/Collection/styles.scss
+++ b/src/amo/components/Collection/styles.scss
@@ -31,6 +31,12 @@
   word-wrap: break-word;
 }
 
+.Collection-detail {
+  .Card-contents {
+    overflow: visible;
+  }
+}
+
 .Collection-author-name-value {
   overflow-x: hidden;
   text-overflow: ellipsis;
--- a/src/amo/components/CollectionManager/styles.scss
+++ b/src/amo/components/CollectionManager/styles.scss
@@ -13,11 +13,6 @@
     }
   }
 
-  .AutoSearchInput-suggestions-list {
-    max-height: 90px;
-    overflow-y: auto;
-  }
-
   label {
     display: block;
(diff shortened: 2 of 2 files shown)


tsl143:
> This only restricts the number of add-ons displayed, but that's very inefficient because we're still making a request that asks for more than three add-ons (and are still getting more than three results back).

This implementation has scroll and that displays all 10 suggestions.


> If we can't fit more than three add-ons, the request should only ask for three 😄

I am not sure that's possible rn, as there is no length/count parameter for Autocomplete API - https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#autocomplete

image

tofumatt:
Oh shoot, I didn’t reloads there was scroll. Still that’s a bit of a hidden UX...

Seeing this now I think the overflow should just be removed for (at least) this card. It’s there for some other reason that I forget and maybe isn’t even needed at all now, but I don’t see why it’s needed here.

Can we at least start with removing the overlfow for this card and just display everything without scroll?

tsl143:
@tofumatt following will be the behavior if I remove `overflow: hidden` from collections card (though this will happen if my collection has only 1 or 2 addons). Is it fine?

image

image

image

tofumatt:
I think that’s fine. It’s a bit weird that it goes into the footer, but that’s only temporary and can be dismissed so I think it’s okay.

Let’s go with that!

tsl143:
@tofumatt updated.

tofumatt:
Fixes #4645

PR #4702, opened 2018-04-01
Issue 4287 - Make reviews and ratings clickable on detail page

### LINKED ISSUE #4287: Make reviews and ratings clickable on detail page
Go to any add-on's detail page

### What happened?
The reviews and ratings area on the upper right corner are not clickable.

I expect to be taken to the /reviews page relevant to the add-on when I click on the "Reviews" and "Overall Rating" labels, as well as the number above "Reviews" and the stars above "Overall Rating".

b68aa4ae-f069-49be-8e09-43860be77901

a-one-d-one:
Fixes #4287 : Added linking to review section on extensions page. 

As shown below : 

screen shot 2018-03-25 at 11 01 02 pm

Fixed the linking : 

mar-25-2018 23-04-25


a-one-d-one:
I would like to expand a little bit more on the testing. Since the Ratings are now being wrapped around a link object. It is not able to actually test the string value of the link anymore. there are 5 failed tests of the same kind with different string values being tested. An example would be : 

expect(received).toEqual(expected)
    
    Expected value to equal:
      "No Ratings"
    Received:
      <Connect(LinkBase) className="Addon-all-reviews-link" to="/addon/chill-out/reviews/">No Ratings</Connect(LinkBase)>
    
    Difference:
    
      Comparing two different types of values. Expected string but received object.

The exact test is linked here  https://github.com/mozilla/addons-frontend/blob/ee80bf95b48595697ec66f9e98c816a69c714c3f/tests/unit/amo/components/TestAddonMeta.js#L99

SeanPrashad:
@a-one-d-one - I've pulled in your patch locally and noticed that you might need to go a "few levels deeper" since you've turned the previous `string` value into an `object` that contains a hyperlink.

As such, for failing tests like [this one](https://github.com/mozilla/addons-frontend/pull/4702/files#diff-8a370ba4cce57cf323ac93e5a94e9a73R159), it should look something like:

`expect(getReviewCount(root).title.props.content).toEqual('No Ratings');`

You'll need to update this in multiple places within [`TestAddonMeta.js`](https://github.com/SeanPrashad/addons-frontend/blob/master/tests/unit/amo/components/TestAddonMeta.js) as your code now touches those areas!

tofumatt:
It seems like this patch is still being worked on, so please let us know when it's ready for review. I know I was originally the mentor on this issue I don't think I'll have the chance to help out with it, really sorry 🙁

a-one-d-one:
After checking Travis CI I noticed there is still a test that is failing in reference to a possible null value. Specifically a "slug" found [here](https://travis-ci.org/mozilla/addons-frontend/jobs/360989480#L521). 

From my limited understanding so far about what a slug is that it handles URL's. I believe that the property is having issues with dynamically figuring out the actual URL to the linked review when the page is loaded.

bobsilverberg:
I'm going to take a look at this and see if I can help sort things out. I know @tofumatt is very busy with other things at the moment.

src/amo/components/AddonMeta/index.js
```diff
@@ -58,8 +60,22 @@ export class AddonMetaBase extends React.Component<Props> {
               title: userTitle,
             },
             {
-              content: reviewCount,
-              title: reviewTitle,
+              content: (
+                <Link
+                  className="Addon-all-reviews-link"
+                  to={`/addon/${addon.slug}/reviews/`}
```
bobsilverberg: This is generating an error when `addon` is null, because then one cannot access `addon.slug`. This is the error you were also seeing on Travis.

I think a nice way to deal with this would be to define a function which would create all of the links for you, because all of the link code is the same except for the content, so something like this:

```js
    const makeRatingLink = (content) => {
      if (addon) {
        return (
          <Link
            className="Addon-all-reviews-link"
            to={`/addon/${addon.slug}/reviews/`}
          >
            {content}
          </Link>
        );
      }
      return content;
    };
```

And then you could replace the above (for example) with `content: makeRatingLink(reviewCount)`.

src/amo/components/AddonMeta/index.js
```diff
@@ -70,7 +86,14 @@ export class AddonMetaBase extends React.Component<Props> {
                   styleSize="small"
                 />
               ),
-              title: i18n.gettext('Overall Rating'),
+              title: (
```
bobsilverberg: I believe the rating stars (at line 82, above) are also meant to be turned into a link.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -99,30 +99,30 @@ describe(__filename, () => {
     it('renders a count of multiple ratings', () => {
       const root = renderRatings({ count: 5 });
 
-      expect(getReviewCount(root).content).toEqual('5');
-      expect(getReviewCount(root).title).toEqual('Ratings');
+      expect(getReviewCount(root).content.props.children).toEqual('5');
```
bobsilverberg: Because the component has been changed to send `Link`s to the `MetadataCard` that is what we should be checking. We should check that the property of the `MetadataCard` is the `Link` that we expect it to be. This is true for all of the tests that check anything that has been set to a `Link`.

src/amo/components/AddonMeta/index.js
a-one-d-one: Thanks for response !! Definitely appreciate the feedback. I have used your suggestions to implement this function and now the tests are passing.

src/amo/components/AddonMeta/index.js
```diff
@@ -58,19 +73,12 @@ export class AddonMetaBase extends React.Component<Props> {
...
             },
             {
-              content: reviewCount,
-              title: reviewTitle,
+              content: makeLink(reviewCount),
+              title: makeLink(reviewTitle),
             },
             {
-              content: (
-                <Rating
```
bobsilverberg: When you removed this `Rating` element that caused the stars to disappear from  the screen altogether and be replaced with a number, which is not what we want. The screen should look exactly as it did before, except that the items should be linked to the reviews page.

a-one-d-one: The CSS issue is noted. Will take care of it. I was trying to figure out how to properly do the testing before I mentioned anything about it, since we are on that topic would you mind guiding me in the right direction on how I would go about implementing that. Once again I appreciate your feedback as well as time and effort.

tests/unit/amo/components/TestAddonMeta.js
bobsilverberg: Here's an example of how you can get at the list of links that were passed into the `MetadataCard` and do some assertions about them:

```js
    it('renders links for each rating item', () => {
      const root = renderRatings({ count: 5 });
      const links = root.find(MetadataCard).dive().find(Link);

      expect(links.at(0)).toHaveProp('to', `/addon/${fakeAddon.slug}/reviews/`);
      expect(links.at(0).children()).toHaveText('5');
     ...etc.
    });
```

Where I've checked the first link, via `links.at(0)`, you could do the same for all 4 expected `Link`s.

bobsilverberg:
@a-one-d-one Just checking in to see if you're still working on this. I made a suggestion about what to do in the tests at https://github.com/mozilla/addons-frontend/pull/4702#discussion_r182118597. Please let me know if there's anything else I can do to help this along.

a-one-d-one:
@bobsilverberg sorry for the delay. Added testing as well made sure the page looks exactly as it was before. I hope what I have done is how it should be done.. looking forward to your response.

package.json
```diff
@@ -182,6 +182,7 @@
     "mozilla-version-comparator": "1.0.2",
     "normalize.css": "8.0.0",
     "normalizr": "3.2.2",
+    "photon-colors": "3.0.1",
```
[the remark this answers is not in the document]
bobsilverberg: Agreed - I don't think this file should change. Can you please revert these changes?

src/amo/components/AddonMeta/index.js
```diff
@@ -47,6 +48,40 @@ export class AddonMetaBase extends React.Component<Props> {
...
+            to={`/addon/${addon.slug}/reviews/`}
+          >
+            {content}
+          </Link>
+        );
+      }
+      return content;
+    };
+
+    const ratingLink = (content) => {
```
bobsilverberg: This function is the same as the `makeLink` function, except that instead of outputting the `content` variable inside the `<Link>` it outputs a `<Rating>` component. 

It seems like you could use the `makeLink` function for this too, and just pass the `<Rating>` component into it as `content`. Do you want to give that a try?

tests/unit/amo/components/TestAddonMeta.js
bobsilverberg: I see you added a test that checks the `Link` components below, but I think that logic should replace what you have above.

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -99,30 +100,43 @@ describe(__filename, () => {
...
+      expect(getReviewCount(root).title.props.children).toEqual('Ratings');
     });
 
     it('renders a count of one rating', () => {
       const root = renderRatings({ count: 1 });
 
-      expect(getReviewCount(root).content).toEqual('1');
-      expect(getReviewCount(root).title).toEqual('Rating');
+      expect(getReviewCount(root).content.props.children).toEqual('1');
```
bobsilverberg: As above, rather than looking at `props.children` I think we should be looking at the props of the actual `Link` that was rendered, as I described in the comment at https://github.com/mozilla/addons-frontend/pull/4702/files#r182118597. This logic should replace the logic in all the tests that are using `props.children`.

bobsilverberg:
@a-one-d-one Are you still working on this? I made some comments which have yet to be addressed. Thanks!

bobsilverberg:
@a-one-d-one Thanks for your efforts thus far on this. As we haven't heard back from you in quite some time I'm going to close this PR for now, but please feel free to open a new one if and when you are able to get back to it.

PR #4707, opened 2018-04-02
Fixes #4626: added spacing between 'featured' badge and theme preview in Theme details page

### LINKED ISSUE #4626: No spacing between 'featured' badge and theme preview in Theme details page
Follow up for #4520 

STR:

1. Log in to AMO and select a featured theme to view it's details page (i.e. https://addons.allizom.org/en-GB/firefox/addon/dew-on-spider-web/?src=featured)
2. Observe the position of the featured label

Actual result:
There is no spacing between the featured badge and theme preview

Expected result:
Some space should be present between the featured badge and the theme preview

Notes:
- reproduced on AMO-stage, FF59 Win10x64

image

svitlana-galianova:
I have added spacing between 'featured' badge and theme preview in Theme details page. 
I have modified left margins only by 6px and 12px based on the screen size. #4626 

Before:
screen shot 2018-04-02 at 9 54 22 am

After:
screen shot 2018-04-02 at 9 53 24 am

### DIFF
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -12,9 +12,13 @@
 }
 
 .AddonBadges .Badge {
+  margin: 0 0 0 6px;
+
   @include text-align-start();
 
   @include respond-to(large) {
+    margin: 0 0 0 12px;
+
     @include text-align-end();
   }
 }
--- a/src/amo/components/AddonBadges/styles.scss
+++ b/src/amo/components/AddonBadges/styles.scss
@@ -12,13 +12,11 @@
 }
 
 .AddonBadges .Badge {
-  margin: 0 0 0 6px;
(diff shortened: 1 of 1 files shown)


src/amo/components/AddonBadges/styles.scss
```diff
@@ -12,9 +12,13 @@
 }
 
 .AddonBadges .Badge {
+  margin: 0 0 0 6px;
```
tofumatt: This won't work when the language is RTL, so you'll want to use our RTL-aware mixin instead: `@include margin(0 0 0 6px);`.

Though in this case if you're only setting left margin you can just use `@include margin-start(6px);`

svitlana-galianova: Thanks for the suggestion, I have just updated it

Issue #4359, opened 2018-02-09
Truncated options from 'Flag' review menu end too abruptly after selection


AlexandraMoga:
STR:

1. Log into AMO and set your locale to `/ru/`
2. Pick a rated add-on to view its details page
3. Go to the rating section and click on Read all reviews to view all available reviews for this add-on (i.e. https://addons-dev.allizom.org/ru/firefox/addon/awesome-screenshot-plus-/reviews/)
4. Click on 'Flag' for any of the listed reviews and select the second or the last option from the menu

Actual result:
After selection, the text will end too abruptly (it doesn't seem to have a text-overflow rule applied)

Expected result:
The text for the selected option should end with an 'ellipsis' 

Notes:
- it would also be helpful to show a tooltip for truncated options to give the user the possibility to read the full text
- reproduced on all AMO servers with FF58, Win10 and Android 7.0

![image](https://user-images.githubusercontent.com/31961530/36026803-540a405a-0da1-11e8-81a0-c4f141fa8af3.png)

tofumatt:
@svitlana-galianova Nice job, two issues with one pull request! 🏅 😄

vcarciu:
Verified as fixed  : 
![russian](https://user-images.githubusercontent.com/1583842/38553457-009f042c-3cc8-11e8-8b52-9cf56ed8ec45.png)

PR #4710, opened 2018-04-02
Fixes #4318: Made 'Log in to flag review' button shorter and less weird

### LINKED ISSUE #4318: 'Log in to flag review' button is not correctly centered
Another follow up for [#2517](https://github.com/mozilla/addons-frontend/issues/2517) which sadly I forgot to mention :disappointed:

STR:
1. Access an add-on review's page while logged out (i.e. https://addons-dev.allizom.org/en-US/android/addon/awesome-screenshot-plus-/reviews/)
2. Click on 'Flag' for any of the listed reviews
3. Observe the position of the 'Log in to flag review' button

Actual result:
The button is not correctly centered

Expected result:
The button should be perfectly centered

Notes:
Reproduced on AMO dev with FF58 (Win10 and Android 7.0)

svitlana-galianova:
#4318 : The button was centered already, so I have fixed width of the pop-up window and made it less 'weird' as was suggested in comments. Please let me know if that is something you were looking for or if you would like to make it shorter.

Before:
screen shot 2018-04-02 at 1 08 24 pm

After:
screen shot 2018-04-02 at 1 07 47 pm

### DIFF
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -14,6 +14,10 @@
   .ErrorList {
     margin-top: 6px;
   }
+
+  .TooltipMenu {
+    width: 220px;
+  }
 }
 
 .AddonReviewListItem-review-header {
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -16,7 +16,7 @@
   }
 
   .TooltipMenu {
-    width: 220px;
+    width: fit-content;
   }
 }
 


src/amo/components/AddonReviewListItem/styles.scss
```diff
@@ -14,6 +14,10 @@
   .ErrorList {
     margin-top: 6px;
   }
+
+  .TooltipMenu {
+    width: 220px;
```
tofumatt: I wonder why the width is so much in the first place... this seems good but setting a width on it might mess up other locales. How does this look in German, for instance?

I wonder if `width: auto` as the default here is messing things up and a `width: 100%;` would fix it?

svitlana-galianova: Width 100% on TooltipMenu component is making things even worth:

screen shot 2018-04-05 at 12 56 56 pm

I have changed width to `fit-content` and it looks good in both German and English now:

screen shot 2018-04-05 at 12 52 54 pm

screen shot 2018-04-05 at 12 51 50 pm

I am not sure if it's ok to use `fit-content`, it's supported in (according to [this website](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content)):

Chrome | Edge | Firefox | Internet Explorer | Opera | Safari
-- | -- | -- | -- | -- | --
 29 | 16 | 51 | No | 44 | 10.1
 
This message in English is this weird because it's width is set to be 320px [here](https://github.com/mozilla/addons-frontend/blob/8e471ffcea03dc86ec152b4613a856d1b451e838/src/ui/components/TooltipMenu/styles.scss#L13). In German it looks perfectly fine with width 320px.

svitlana-galianova:
@tofumatt I was looking for new issues and I came across this one #4359 I have checked and this PR fixed #4359 as well. I think that issue should also be closed

tofumatt:
Thanks, I've closed that issue as well. 👍

PR #4713, opened 2018-04-03
fix: Do not show rating option when readOnly set

### LINKED ISSUE #4706: Incorrect tooltip displayed for average add-on rating in user info card
STR:

1. Log in to AMO-dev with a user that has no add-ons submitted
2. Select "View Profile" form the user menu
3. Hover over the rating stars in the user info card

Actual result:
A "Click to rate this add-on" tooltip is displayed, suggesting a possible interaction with the stars

Expected result:
A more suggestive tooltip should be implemented for the "Average rating of developer’s add-ons" when there is no rating available

Notes:
- the same tooltip can also be observed in statistics card in add-on details page
- reproduced on AMO-dev, with FF59, Win10x64
- introduced by #3231 

[1] User info card
image

[2] add-on statistics card
image

tofumatt:
fix #4706

### Before
image

### After
screenshot 2018-04-03 10 23 48

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -92,7 +92,7 @@ export class RatingBase extends React.Component {
         `styleSize=${styleSize} is not a valid value; ` +
         `possible values: ${RATING_STYLE_SIZES.join(', ')}`);
     }
-    let description;
+    let description = i18n.gettext('This add-on has no ratings.');
     if (rating) {
       description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
         { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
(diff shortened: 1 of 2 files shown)


src/ui/components/Rating/index.js
```diff
@@ -96,7 +96,7 @@ export class RatingBase extends React.Component {
     if (rating) {
       description = i18n.sprintf(i18n.gettext('Rated %(rating)s out of 5'),
         { rating: i18n.formatNumber(parseFloat(rating).toFixed(1)) });
-    } else {
+    } else if (!readOnly) {
```
bobsilverberg: Should we be displaying something other than nothing if it's `readOnly` and there are no ratings? Something like `This add-on has not been rated yet`? That seems to be suggested in issue #4706.

tofumatt: Sure, we can do that. I'll make the change.

tofumatt:
@bobsilverberg I've added the message over the empty text.

tests/unit/ui/components/TestRating.js
```diff
@@ -188,6 +188,11 @@ describe(__filename, () => {
     expect(findDOMNode(root).textContent).toEqual('Click to rate this add-on');
   });
 
+  it('renders nothing if no rating and readOnly', () => {
```
bobsilverberg: Now you need to change the description of this test. Sorry.

tofumatt: Whoops, right, heh 😄

Issue #4712, opened 2018-04-03
Add-on badges are not correctly aligned on Android landscape mode


AlexandraMoga:
Follow up for #4542 

STR:
1. Access AMO-dev from an Android device
2. Open an extension details page for an add-on that has some badges (i.e. https://addons-dev.allizom.org/en-US/android/addon/versiontest/)
3. Rotate your device in landscape mode and observe the position of the badges

Actual result:
Badges are aligned on the bottom right of the add-on card

Expected result:
Badges are aligned on the top right, similar to what we have for desktop.

Notes:
- reproduced on all AMO servers with FF59, Android 7.0

![image](https://user-images.githubusercontent.com/31961530/38237580-4822cd3a-3730-11e8-9f88-a6fe0bbe77f1.png)

vcarciu:
Verified as fixed : 
![screenshot_20180411-154121](https://user-images.githubusercontent.com/1583842/38617157-ee2d161c-3d9e-11e8-8ce5-5fe45c798ae1.png)

PR #4718, opened 2018-04-03
fixes I4712 - regression issue for badges

tsl143:
Fixes: #4712

Before: 
screen shot 2018-04-03 at 10 37 35 pm

After:
screen shot 2018-04-03 at 10 37 18 pm


src/amo/components/AddonBadges/styles.scss
```diff
@@ -12,7 +12,6 @@
 }
 
 .AddonBadges .Badge {
-  @include margin-start(6px);
```
tsl143: @tofumatt we need to remove this from #4707 as  `theme image` and `badge` are on different rows on mobile.

tsl143: @tofumatt on desktop its handled in line https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonBadges/styles.scss#L19
On screen less than `large` we don't need this, where badges are not on `right-top` but below title and description.

Themes are not impacted by this patch 
screen shot 2018-04-04 at 8 33 21 am
screen shot 2018-04-04 at 8 33 06 am
screen shot 2018-04-04 at 8 32 54 am

tofumatt: Ah, okay, but is that featured theme badge on the large size off a bit? It seems like there's a lot of padding on it, pushing it to the left a bit more than expected.

Maybe that's just the regular padding.

tsl143:
@tofumatt , on large screens, the badge is right aligned, so margin on left actually doesn't make any difference for badge position(vice versa for RTL), but that helps with theme putting white space between theme image and badge.

Issue #4668, opened 2018-03-30
Create edit user reducer and API requests


tofumatt:
We'll need a reducer/API/saga stuff to send user profile edits to the API. I'll submit them as their own PR before the components.

- [x] APIs (#4721)
- [x] reducers (#4734)
- [ ] sagas (#4736)

PR #4721, opened 2018-04-03
chore: add API for edit user profile (#4668)

tofumatt:
Gonna do the API/reducer/saga stuff for this all separately and maybe it'll make the PRs nice and easy.

Here's the API and tests for editing a user profile. I've exclude the image upload for now as I'll do that last.

Part of #4668.

### DIFF
--- a/tests/unit/amo/api/test_users.js
+++ b/tests/unit/amo/api/test_users.js
@@ -1,10 +1,18 @@
 import * as api from 'core/api';
-import { currentUserAccount, userAccount } from 'amo/api/users';
+import {
+  currentUserAccount,
+  editUserAccount,
+  userAccount,
+} from 'amo/api/users';
+import { getCurrentUser } from 'amo/reducers/users';
 import {
   createApiResponse,
   createUserAccountResponse,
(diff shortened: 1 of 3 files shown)


tests/unit/amo/api/test_users.js
```diff
@@ -30,16 +38,46 @@ describe(__filename, () => {
...
+  describe('editUserAccount', () => {
+    const mockResponse = (newParams) => createApiResponse({
+      jsonData: createUserAccountResponse(newParams),
     });
 
-    it('throws an error if api state is missing', () => {
-      expect(() => {
-        currentUserAccount({});
-      }).toThrowError(/api state is required/);
+    const getParams = (params = {}) => {
```
bobsilverberg: This is just a nit; take it or leave it. Considering that there's only one test that makes use of it, this `getParams` function is kind of unnecessary. It does seem like a good model to follow, and I know we use it in other test files.  I notice the same model is followed for the other tests in this file, which also don't really need it. If  more tests are added later it would be useful, but at this point none of these cases seems to need more tests.

Issue #4635, opened 2018-03-24
Form buttons are not disabled when saving a collection


kumar303:
### Describe the problem and steps to reproduce it:

After https://github.com/mozilla/addons-frontend/issues/4623 lands...
- sign in
- create a collection
- view the collection
- click the edit button to see a form
- click the *Save Collection* button

### What happened?

While the form is saving the submit button is not disabled

### What did you expect to happen?

The submit button should be disabled while saving the form.

### Anything else we should know?

We had a half baked and weirdly named [`collectionUpdates`](https://github.com/mozilla/addons-frontend/blob/1c7be9eacf1389575d1bce57b9480ff990d6b9f1/src/amo/reducers/collections.js#L81-L83) state that was designed to support this kind of thing. It needs a `finishCollectionUpdate()` action and both need to be hooked up to the `updateCollection` saga. Beware that it might get removed in https://github.com/mozilla/addons-frontend/pull/4564.

bobsilverberg:
Note that I have changed #4564 so that I am no longer removing the `collectionUpdates` state, but we still need to hook it up properly so that the state is updated when a collection has finished updating.

vcarciu:
Verified as fixed : 
![disabledbuttons](https://user-images.githubusercontent.com/1583842/38554663-177470ac-3ccc-11e8-8c6e-2ded827eb3fc.gif)

PR #4723, opened 2018-04-03
Properly implement the collectionUpdates state

### LINKED ISSUE #4717: Properly implement the collectionUpdates state
The `collectionUpdates` state was introduced in https://github.com/mozilla/addons-frontend/commit/a1891d5480db8e92e30b0dda5c8f8665f99482d7, but it isn't actually used for anything yet, and isn't updated after a collection has finished updating. We need to complete the implementation of it by doing that, which will also help fix issue #4635.

Once it's done we can also improve the tests in `tests/unit/amo/sagas/test_collections.js` which currently have a TODO comment discussing this.

bobsilverberg:
Fixes #4717 

This finishes implementing the collectionUpdates state which is managed by the collections reducer. The state will be set to `updating: true` when an update is started, and to `updating: false` when an update finishes. In addition, the state will have a `successful: true` if there were no errors, and a `successful: false` if there were errors. It also introduces a `getCollectionUpdateStatus` function which will return the current update status of a collection. This will be used on the CollectionManager screen to address issue #4635, which will be completed in a separate PR.

### DIFF
--- a/src/amo/sagas/collections.js
+++ b/src/amo/sagas/collections.js
@@ -22,6 +22,8 @@ import {
   loadCurrentCollection,
   loadCurrentCollectionPage,
   loadUserCollections,
+  beginCollectionModification,
+  finishCollectionModification,
 } from 'amo/reducers/collections';
 import * as api from 'amo/api/collections';
 import log from 'core/logger';
@@ -182,6 +184,8 @@ export function* modifyCollection(
     user,
   } = payload;
 
+  yield put(beginCollectionModification());
+
   const errorHandler = createErrorHandler(errorHandlerId);
   yield put(errorHandler.createClearingAction());
 
@@ -220,6 +224,7 @@ export function* modifyCollection(
       };
       yield call(api.updateCollection, apiParams);
     }
+    yield put(finishCollectionModification());
 
     const { lang, clientApp } = state.api;
     const effectiveSlug = slug || collectionSlug;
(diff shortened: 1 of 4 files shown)


src/amo/sagas/collections.js
```diff
@@ -244,6 +249,7 @@ export function* modifyCollection(
     }
   } catch (error) {
     log.warn(`Failed to ${type}: ${error}`);
+    yield put(finishCollectionModification());
```
tofumatt: Given this is run in either scenario (`try`/success and `catch` failure), wouldn't it make more sense to put `yield put(finishCollectionModification());` in a `finally {}` block? It changes the order, but I don't think that should matter.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -370,6 +373,50 @@ describe(__filename, () => {
         const action = await sagaTester.waitFor(expectedAction.type);
         expect(action).toEqual(expectedAction);
       });
+
+      it('records the beginning of a modification in state', async () => {
+        const collectionSlug = 'my-slug';
+
+        mockApi
+          .expects('updateCollection')
+          .returns(Promise.resolve());
```
tofumatt: Does updating the collection not return any params? Or do we just go on status code to know if the update was successful or not? It seems odd not to care about the API response and not to mock it.

src/amo/sagas/collections.js
bobsilverberg: Great idea, thanks!

tests/unit/amo/sagas/test_collections.js
bobsilverberg: In the saga we don't do anything with the response. We assume no error means it worked. This is how the code was when Kumar originally wrote it, so I think it's fine.

PR #4727, opened 2018-04-04
Fixes I4719 - disable submit if slug blank(edit collection)

### LINKED ISSUE #4719: Disable Save Collection button if the slug is blank
This is a replacement for issue #4650 and will be similar to the resolution in issue #4698. If the collection slug is blank, or a string which is only spaces, the "Save Collection" button should be disabled.

@tsl143 Are you interested in taking this one, as it's very similar to the work you did for #4698?

tsl143:
Fixes #4719 

Before: 
screen shot 2018-04-04 at 8 24 34 am

After: 
screen shot 2018-04-04 at 8 23 59 am

### DIFF
--- a/tests/unit/amo/components/TestCollectionManager.js
+++ b/tests/unit/amo/components/TestCollectionManager.js
@@ -268,6 +268,18 @@ describe(__filename, () => {
       .toHaveProp('disabled', true);
   });
 
+  it('disables submit button when the slug is blank', () => {
+    const root = render();
+
+    // Enter in a blank collection name.
+    typeInput({ root, name: 'slug', text: '' });
+
+    expect(root.find('.CollectionManager-cancel'))
+      .toHaveProp('disabled', false);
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -268,6 +268,18 @@ describe(__filename, () => {
       .toHaveProp('disabled', true);
   });
 
+  it('disables submit button when the slug is blank', () => {
+    const root = render();
+
+    // Enter in a blank collection name.
+    typeInput({ root, name: 'slug', text: '' });
```
bobsilverberg: I'd like to see a test for a slug which isn't empty, but contains only spaces. We should add one for the name field too, if we don't have one already. I think it's fine to just add it to this case.

tsl143:
Can we get this merged?

bobsilverberg:
Yes!

PR #4733, opened 2018-04-04
 Disable form buttons while a collection is being updated 

### LINKED ISSUE #4635: Form buttons are not disabled when saving a collection

bobsilverberg:
Fixes #4635 

Note that this builds upon the changes in #4723, so it contains the commit from that as well, but only the code in the more recent commit needs to be reviewed. I will land this on top of the changes in #4723 after it lands.


bobsilverberg:
Hmm, I think if I extract a `beginCollectionUpdate` action from the `updateCollection` action, then I could dispatch just that in my test and it all would work fine. It feels like extra work that will just make the overall flow more complex, but it will make the code testable, and that's usually an indication of better code. I guess I'll make that change in #4723 and then revisit this.

bobsilverberg:
This has been rebased on top of #4723 and is ready to review. You can ignore my rambling comments above.

src/amo/components/CollectionManager/index.js
```diff
@@ -128,26 +132,26 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
     log.debug('TODO: handle seaching for add-on', filters);
-  }
+  };
 
   onAddonSelected = (suggestion: SuggestionType) => {
     // TODO: implement onAddonSelected
     // https://github.com/mozilla/addons-frontend/issues/4590
     log.debug('TODO: handle selecting an add-on', suggestion);
-  }
+  };
```
tofumatt: Huh, these are needed now? Weird, I thought they were a linting error. (Unneeded semicolons.)

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -304,6 +307,24 @@ describe(__filename, () => {
       .toHaveProp('disabled', true);
   });
 
+  it('disables and enables form buttons when modification status changes', () => {
+    const renderAndCheckButtons = (shouldBeDisabled) => {
+      const root = render();
+
+      expect(root.find('.CollectionManager-cancel'))
```
tofumatt: Should we be disabling "cancel" at any point? I guess during submit there's the chance the request is posted, we cancel, and get in a weird state forcing a page refresh, but allowing a user to cancel during submission might be nice.

If it's a pain to implement though we could skip it for now.

src/amo/components/CollectionManager/index.js
tofumatt: Seems like not though. 🤷‍♂️

bobsilverberg: They are neither needed nor a linting error. It would have been a linting error if this were Firefox code, but I guess our configuration allows both to include and exclude them. I find having them excluded annoying, as it seems inconsistent to me, which is why I added them.

tests/unit/amo/components/TestCollectionManager.js
bobsilverberg: Allowing a user to cancel during submission would be quite complicated, because we'd then have to check whether the API request actually completed successfully, and if it did we'd have to "undo" it. As this is all async, we wouldn't even know if the API request to update the object had been successful when the cancel button is clicked, which makes it even more complicated. The simplest solution, which still protects against multiple actions, is to just disable the whole form (submit and cancel) until we have a result from the API, which is what this patch does.

PR #4734, opened 2018-04-04
chore: Add edit user reducer actions (#4668)

tofumatt:
Part of #4668.

This adds the reducer code needed to allow us to edit users. The `editUserAccount` action will be used to dispatch an action to a saga that will call the edit API.

That API returns the updated edit user object, so we'll just use the existing `loadUserAccount` action to update the user account in state after editing.

### DIFF
--- a/src/amo/reducers/users.js
+++ b/src/amo/reducers/users.js
@@ -13,6 +13,8 @@ import {
 } from 'core/constants';
 
 
+export const CANCEL_EDIT_USER_ACCOUNT: 'CANCEL_EDIT_USER_ACCOUNT' = 'CANCEL_EDIT_USER_ACCOUNT';
+export const EDIT_USER_ACCOUNT: 'EDIT_USER_ACCOUNT' = 'EDIT_USER_ACCOUNT';
 export const LOG_OUT_USER: 'LOG_OUT_USER' = 'LOG_OUT_USER';
 export const LOAD_CURRENT_USER_ACCOUNT: 'LOAD_CURRENT_USER_ACCOUNT' = 'LOAD_CURRENT_USER_ACCOUNT';
 export const FETCH_USER_ACCOUNT: 'FETCH_USER_ACCOUNT' = 'FETCH_USER_ACCOUNT';
@@ -60,6 +62,7 @@ export type UsersStateType = {
   currentUserID: UserId | null,
   byID: { [userId: UserId]: UserType },
   byUsername: { [username: string]: UserId },
+  isEditingById: { [userId: UserId]: boolean },
   userPageBeingViewed: {
     loading: boolean,
     userId: UserId | null,
@@ -79,6 +82,7 @@ export const initialState: UsersStateType = {
   currentUserID: null,
   byID: {},
   byUsername: {},
+  isEditingById: {},
   userPageBeingViewed: {
     loading: false,
     userId: null,
@@ -99,12 +103,8 @@ export const fetchUserAccount = ({
   errorHandlerId,
   username,
 }: FetchUserAccountParams): FetchUserAccountAction => {
-  if (!errorHandlerId) {
-    throw new Error('errorHandlerId is required');
-  }
-  if (!username) {
-    throw new Error('username is required');
-  }
+  invariant(errorHandlerId, 'errorHandlerId is required');
+  invariant(username, 'username is required');
 
   return {
     type: FETCH_USER_ACCOUNT,
@@ -112,6 +112,50 @@ export const fetchUserAccount = ({
       errorHandlerId,
       username,
     },
+  };
+};
+
+type CancelEditUserAccountParams = {|
+  userId: UserId,
+|};
+
+type CancelEditUserAccountAction = {|
+  type: typeof CANCEL_EDIT_USER_ACCOUNT,
+  payload: CancelEditUserAccountParams,
+|};
+
+export const cancelEditUserAccount = (
+  { userId }: CancelEditUserAccountParams
+): CancelEditUserAccountAction => {
+  invariant(userId, 'userId is required');
+
+  return {
+    type: CANCEL_EDIT_USER_ACCOUNT,
+    payload: { userId },
+  };
+};
+
+type EditUserAccountParams = {|
+  errorHandlerId: string,
+  userFields: UserEditableFieldsType,
+  userId: UserId,
+|};
+
+type EditUserAccountAction = {|
+  type: typeof EDIT_USER_ACCOUNT,
+  payload: EditUserAccountParams,
+|};
+
+export const editUserAccount = ({
+  errorHandlerId, userFields, userId,
+}: EditUserAccountParams): EditUserAccountAction => {
+  invariant(errorHandlerId, 'errorHandlerId is required');
+  invariant(userFields, 'userFields are required');
+  invariant(userId, 'userId is required');
+
+  return {
+    type: EDIT_USER_ACCOUNT,
+    payload: { errorHandlerId, userFields, userId },
   };
 };
 
@@ -127,9 +171,7 @@ type LoadCurrentUserAccountAction = {|
 export const loadCurrentUserAccount = ({
   user,
 }: LoadCurrentUserAccountParams): LoadCurrentUserAccountAction => {
-  if (!user) {
-    throw new Error('user is required');
-  }
+  invariant(user, 'user is required');
 
   return {
     type: LOAD_CURRENT_USER_ACCOUNT,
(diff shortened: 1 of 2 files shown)


src/amo/reducers/users.js
```diff
@@ -60,6 +62,7 @@ export type UsersStateType = {
   currentUserID: UserId | null,
   byID: { [userId: UserId]: UserType },
   byUsername: { [username: string]: UserId },
+  isEditingById: { [userId: UserId]: boolean },
```
bobsilverberg: This is similar to what we just discussed about a flag for a collection being edited. Why do we need to store multiple objects that represent multiple users being edited, as opposed to just a single flag that says "a user is being edited"?

src/amo/reducers/users.js
```diff
@@ -115,6 +115,50 @@ export const fetchUserAccount = ({
   };
 };
 
+type CancelEditUserAccountParams = {|
```
bobsilverberg: I don't understand what `CANCEL_EDIT_USER_ACCOUNT` is for. When would it be called? Can a "user account edit" be cancelled by a user? Can it be cancelled by a saga?

src/amo/reducers/users.js
```diff
@@ -176,14 +216,17 @@ export const getCurrentUser = (users: UsersStateType) => {
...
-    throw new Error(
-      'currentUserID is defined but no matching user found in users state.');
-  }
+  invariant(users.byID[users.currentUserID],
+    'currentUserID is defined but no matching user found in users state.');
 
   return users.byID[users.currentUserID];
 };
 
+export const getUserById = (users: UsersStateType, userId: number) => {
```
bobsilverberg: This isn't used anywhere. Is it something the will be needed for the saga?

src/amo/reducers/users.js
```diff
@@ -261,8 +304,29 @@ export const addUserToState = ({ user, state } : {
   return { byID, byUsername };
 };
 
+export const setUserAsEditing = ({ state, userId } : {
```
bobsilverberg: I don't understand the value of these helpers. They include validations via `invariant`, but so do the actions that cause them to be called (e.g., `setUserAsEditing` is called as a result of a `EDIT_USER_ACCOUNT`, and the `editUserAccount` action function which returns the `EDIT_USER_ACCOUNT` action also includes the validation). So the extra validation for `userId` here isn't really necessary.

It seems like it would be simpler to do away with the helper entirely and just inline the code which sets the `isEditingById` state in the `EDIT_USER_ACCOUNT` case of the reducer function.

src/amo/reducers/users.js
```diff
@@ -261,8 +304,29 @@ export const addUserToState = ({ user, state } : {
...
+  state: UsersStateType,
+  userId: UserId,
+}) : Object => {
+  invariant(state, 'state is required');
+  invariant(userId, 'userId is required');
+
+  return { isEditingById: { ...state.isEditingById, [userId]: true } };
+};
+
+export const setUserAsNotEditing = ({ state, userId } : {
```
bobsilverberg: I feel the same way about this helper.

src/amo/reducers/users.js
```diff
@@ -272,6 +336,24 @@ const reducer = (
   action: Action
 ): UsersStateType => {
   switch (action.type) {
+    case CANCEL_EDIT_USER_ACCOUNT: {
+      const { userId } = action.payload;
+
+      return {
```
bobsilverberg: As an example of what I meant above, instead of calling a `setUserAsNotEditing` helper here, you could just make this:

```js
      return {
        ...state,
        isEditingById: {...state.isEditingById, [userId]: false },
      };
```

src/amo/reducers/users.js
```diff
@@ -288,6 +370,7 @@ const reducer = (
       return {
         ...state,
         ...addUserToState({ state, user }),
+        ...setUserAsNotEditing({ state, userId: user.id }),
```
bobsilverberg: I'm not sure this makes sense. We're assuming that the fact that we're loading a user account means that we're no longer editing. Would it be better to have a `FINISH_EDIT_USER_ACCOUNT` that is explicitly called when we know the edit is finished? That's what I did in #4723, but maybe that's not the best approach.

Oh! Is that what `CANCEL_EDIT_USER_ACCOUNT` is all about?

src/amo/reducers/users.js
tofumatt: For right now I guess we don't, although editing multiple users is a thing I could see happening.

I guess actually, to your point... editing multiple collections could happen too in the future.

Maybe what we want to do is have a special key for a new/unexisting collection/user. Sorry, that sucks because I said you could change it but now I'm stepping back.

That said we ALWAYS have a user id because by the time you're signed in/editing on the front-end the server made you a user. So this way at least in the future we could allow multiple user edits at once.

I guess if I did it the other way I wouldn't say we should change it, but now that this one works and there's no drawback I don't know if it's worth changing? What does it buy us other than a *very* marginally simpler code path? Keying by ID might back debugging easier in some bizarre scenario.

Sorry, I'm clearing stepping back from my earlier suggestion of not caring re: collection ID keys.

src/amo/reducers/users.js
tofumatt: It would get cancelled by failure. If you see the saga patch (upcoming) then yeah: because we set an `isEditing` state to `true` during our `editUserAccountApi` call, we need a way to set `isEditing` to `false` if the request failed.

Happy path is:
1. Click "Submit edits"
2. `EDIT_USER_ACCOUNT` dispatched
3. `editUserAccount` API call made
4. API call finishes and saga dispatches `LOAD_USER_ACCOUNT` with new user info
5. `LOAD_USER_ACCOUNT` sets `isEditing` to `false`

Sad path is:
1. Click "Submit edits"
2. `EDIT_USER_ACCOUNT` dispatched
3. `editUserAccount` API call made
4. API call fails/returns error and saga dispatches `SET_ERROR` with error and `CANCEL_EDIT_USER_ACCOUNT`
5. `LOAD_USER_ACCOUNT` is never called, so `CANCEL_EDIT_USER_ACCOUNT` sets `isEditing` to `false`

src/amo/reducers/users.js
tofumatt: Looks like I was *going* to use it in here and didn't. I'll either update the code to use it or remove it.

src/amo/reducers/users.js
tofumatt: Fair enough. I think I'm just growing to like using selectors and helpers to consolidate this kind of behaviour, but this one is pretty simple. I'll see about removing them.

src/amo/reducers/users.js
tofumatt: We *could* have a separate action, I'm fine with that. Cancel was for errors, but I guess I could just change it to `FINISH` and use it in either place.

src/amo/reducers/users.js
bobsilverberg: How would editing multiple users or collections work? Are you thinking a grid of forms where multiple forms for multiple users/collections would be displayed on a single screen?

src/amo/reducers/users.js
bobsilverberg: Yeah, now I understand about Cancel. I called it Finish in my patch. I find the term `Cancel` confusing because it sounds like you're actually cancelling (i.e., aborting) something, when really it's just about updating the `editing` state. I would prefer that we use Begin and Finish in both patches, if that's okay with you.

src/amo/reducers/users.js
tofumatt: Yeah, for an admin thing it could happen. That said it’s nowhere in the mocks.

I guess the questions are:

1. Is the current code here significantly harder-to-follow than a single `isEditing` flag?
2. Is refactoring to add userIds later very difficult/annoying?
3. Are we likely to ever want multiple user edits?

I think the answer to 1. is no, 2. is yes, 3. is no.

So I'll change it if you feel strongly about it. I guess it'd be a smaller patch/reducer.

src/amo/reducers/users.js
tofumatt: Finish is totally better, wfm 👍

src/amo/reducers/users.js
bobsilverberg: I agree with your answers to the questions, so either way seems fine. I guess for me I tend to never add code that isn't going to be used - you know the old YAGNI chestnut. I might even suggest that the answer to 2. is yes, but not actually *very* difficult - just a pain.

In this user case it doesn't really matter, but with the collections one, not having it keyed avoids the issue of what to use for the key if there is no existing collection id/slug.

Your call. 😄

tofumatt:
Made all the changes and it brought down the lines changed a fair bit too 👍

PR #4736, opened 2018-04-05
Feat/edit user profile sagas

### LINKED ISSUE #4668: Create edit user reducer and API requests

tofumatt:
Fixes #4668

Depends on #4734.

### DIFF
--- a/tests/unit/amo/sagas/test_users.js
+++ b/tests/unit/amo/sagas/test_users.js
@@ -2,6 +2,8 @@ import SagaTester from 'redux-saga-tester';
 
 import usersSaga from 'amo/sagas/users';
 import usersReducer, {
+  finishEditUserAccount,
+  editUserAccount,
   fetchUserAccount,
   loadCurrentUserAccount,
   loadUserAccount,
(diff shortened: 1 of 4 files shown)


tests/unit/amo/sagas/test_users.js
```diff
@@ -37,78 +39,148 @@ describe(__filename, () => {
...
+        .once()
+        .returns(Promise.resolve(user));
+
+      sagaTester.dispatch(editUserAccount({
+        errorHandlerId: errorHandler.id,
+        userFields,
+        userId: user.id,
+      }));
+
+      const expectedCalledAction = editUserAccount({
```
bobsilverberg: I don't think we have to set this up as the expected action, and then verify it was called as expected, because we are literally calling it above in `sagaTester.dispatch`. What we do need to verify is that the `loadUserAccount` action is dispatched, as that's something the saga does, and if we replace this expected action with one for `loadUserAccount` then that will also allow the test to wait long enough to be able to verify the mockApi.

tests/unit/amo/sagas/test_users.js
```diff
@@ -37,78 +39,148 @@ describe(__filename, () => {
...
+        userFields,
+        userId: user.id,
+      });
+
+      const calledAction = await sagaTester.waitFor(expectedCalledAction.type);
+
+      mockApi.verify();
+      expect(calledAction).toEqual(expectedCalledAction);
+
+      // Make sure the finish action is also called.
```
bobsilverberg: We should also verify that `errorHandler.createClearingAction` was called, as that's something else that the saga code does.

tofumatt:
Shoot, I meant to comment on those style changes in the patch, but I left it as a review comment and not a single one, so it messed up.

But I meant to say that I was getting linting errors, and the string formatting changes are unrelated changes but needed to pass the lint on my Mac. I'll see about removing them if I can...

tests/unit/amo/sagas/test_users.js
tofumatt: OMG, shoot, that's totally what I wanted to do but I just messed up. Thanks.

tofumatt:
Sorry for the delay on this one, all set for another review :smile:

PR #4739, opened 2018-04-05
feat: Add custom download URLs to `getClientCompatibility()`

### LINKED ISSUE #4737: Special case the Firefox download link for Facebook Container
Refiled from https://bugzilla.mozilla.org/show_bug.cgi?id=1451542

>The Facebook Container team would like to change the link that takes you to the Firefox install page to instead take you to the moz.org landing page, https://www.mozilla.org/en-US/firefox/facebookcontainer/, for the funnel cake set up for Facebook Container.

>For context, here's a summary of the discussion that led to this request:

>If you get to the Facebook Container AMO page, it will prompt you to download Firefox to install the add-on.

>It would be great for people who go straight to the AMO page from browsers other than Firefox, if we could link to the landing page first (from green download firefox link) and then have users go through the funnel cake from there. Need this change by Thursday, April 4th.

So in a nutshell the Download Firefox link seen by a non-firefox user on https://addons.mozilla.org/en-US/firefox/addon/facebook-container/ should be special-cased to link to  https://www.mozilla.org/firefox/facebookcontainer/ (Note I removed the locale from the latter so that Mozilla.org can handle the locale itself.

facebook_container_ _add-ons_for_firefox

We'll need to see how easy it is to special-case this change. The request was to have it ready for today's push (5th April). If it's a minimal change then we can look to cherry-pick it, but the first step is to investigate and see how viable it is.

tofumatt:
Fixes #4737

### DIFF
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -16,6 +16,12 @@ import { findInstallURL } from 'core/installAddon';
 import log from 'core/logger';
 
 
+// HACK: This is the GUID for the Facebook Container
+// add-on, which has a special-cased download URL supplied.
+// See: https://github.com/mozilla/addons-server/issues/7982
+export const FACEBOOK_CONTAINER_ADDON_GUID = '@contain-facebook';
+export const FACEBOOK_CONTAINER_DOWNLOAD_URL = 'https://www.mozilla.org/firefox/facebookcontainer/';
+
 export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
   let maxVersion = null;
   let minVersion = null;
@@ -155,8 +161,14 @@ export function getClientCompatibility({
     reason = agent.reason;
   }
 
+  let downloadUrl;
+  if (addon && addon.guid === FACEBOOK_CONTAINER_ADDON_GUID) {
+    downloadUrl = FACEBOOK_CONTAINER_DOWNLOAD_URL;
+  }
+
   return {
     compatible: agent.compatible && supportsClientApp,
+    downloadUrl,
     maxVersion,
     minVersion,
     reason,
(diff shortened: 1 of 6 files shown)


src/core/utils/compatibility.js
```diff
@@ -16,6 +16,14 @@ import { findInstallURL } from 'core/installAddon';
 import log from 'core/logger';
 
 
+// HACK: This is the ID on dev/stage/prod for the Facebook Container
+// add-on, which has a special-cased download URL supplied.
+// This should really be a special field set in the database that the front-end
+// can display, but this patch was last-minute and needed a timely turnaround.
+// See: https://github.com/mozilla/addons-server/issues/7982
+export const FACEBOOK_CONTAINER_ADDON_ID = 619864;
+export const FACEBOOK_CONTAINER_DOWNLOAD_URL = 'https://www.mozilla.org/firefox/facebookcontainer/';
```
tofumatt: The issue just asked for this URL, so I wasn't sure if we wanted the UTM links in there as well; I'm not even sure that page responds to them.

But if they do and we need to tweak this patch so the UTM links are always added I can do that easily, just let me know.

tofumatt: Actually, that would make more, sense, I'll update the PR with that tweak so that the UTM query params are always added.

src/core/utils/compatibility.js
```diff
@@ -16,6 +16,14 @@ import { findInstallURL } from 'core/installAddon';
 import log from 'core/logger';
 
 
+// HACK: This is the ID on dev/stage/prod for the Facebook Container
+// add-on, which has a special-cased download URL supplied.
+// This should really be a special field set in the database that the front-end
+// can display, but this patch was last-minute and needed a timely turnaround.
+// See: https://github.com/mozilla/addons-server/issues/7982
+export const FACEBOOK_CONTAINER_ADDON_ID = 619864;
```
muffinresearch: Could this be keyed by guid instead? Won't using the id mean it won't work on -dev or stage?

tofumatt: Oh, snap, sure. I thought the ID was the same everywhere on IRC but I misunderstood. Easily fixed.

PR #4743, opened 2018-04-05
fix: #4621 regressions

### LINKED ISSUE #4621: Abstract out "add-ons by authors" card to use in UserProfile
We currently don't have a reusable card that can show an `AddonsCard` based on a username or list or usernames. I have one nearly ready that we can use to easily add cards to UserProfile for #4280 and #4281.

tofumatt:
Closes #4621.

Restores the correct number of add-ons to the "more add-ons by authors" cards on the add-on details.

### Before
morebyauthornumberextensions
morebyauthornumber

### After
screenshot 2018-04-05 15 17 34
screenshot 2018-04-05 15 17 26

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -490,7 +490,7 @@ export class AddonBase extends React.Component {
                 authorNames={addon.authors.map((author) => author.username)}
                 className="Addon-MoreAddonsCard"
                 forAddonSlug={addon.slug}
-                numberOfAddons={3}
+                numberOfAddons={6}
               />
             )}
 
@@ -524,7 +524,7 @@ export class AddonBase extends React.Component {
               authorNames={addon.authors.map((author) => author.username)}
               className="Addon-MoreAddonsCard"
               forAddonSlug={addon.slug}
-              numberOfAddons={4}
+              numberOfAddons={6}
             />
           )}
         </div>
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -490,7 +490,7 @@ export class AddonBase extends React.Component {
                 authorNames={addon.authors.map((author) => author.username)}
                 className="Addon-MoreAddonsCard"
                 forAddonSlug={addon.slug}
-                numberOfAddons={3}
```
bobsilverberg: Just out of interest, where did the `3` and `4` come from?

tests/unit/amo/components/TestAddon.js
```diff
@@ -1123,6 +1123,9 @@ describe(__filename, () => {
         { ...fakeAddon, slug: 'addon-1', id: 1 },
         { ...fakeAddon, slug: 'addon-2', id: 2 },
         { ...fakeAddon, slug: 'addon-3', id: 3 },
+        { ...fakeAddon, slug: 'addon-4', id: 4 },
+        { ...fakeAddon, slug: 'addon-5', id: 5 },
+        { ...fakeAddon, slug: 'addon-6', id: 6 },
```
bobsilverberg: Would it be worthwhile to make this contain more than 6 add-ons, to test that the limit of 6 is actually being enforced?

src/amo/components/Addon/index.js
tofumatt: The 4 was arbitrary but the three was for a single row. I didn't check prod and I think I tested against some add-ons that just happened to have three as a match locally so thought it was right. Oops.

tests/unit/amo/components/TestAddon.js
tofumatt: Oh, good call. I'll do that. 👍

tofumatt: Actually, the add-ons by authors card does that bit. Arguably these aren't needed but I thought made the intentional clearer so I left them. 🤷‍♂️

Issue #4747, opened 2018-04-05
Discopane ui tests are failing


bobsilverberg:
Starting with https://github.com/mozilla/addons-frontend/commit/3fd1e1acfae57b55171e9cc85bee3fccbd93cc57, the discopane ui tests are failing. See 
https://travis-ci.org/mozilla/addons-frontend/builds/362738197 for the failure related to that commit.

There doesn't seem to be anything in that commit that would trigger a failure.

tofumatt:
I'm going to disable the tests for now as the noise is making it tough to work on PRs, but I'll leave the issue open because it needs to get fixed.

PR #4749, opened 2018-04-05
Fixes #4633: changed the icon on firefox compatibility error

### LINKED ISSUE #4633: Change "Download Firefox" message bar from grey to green
I'd rather we A/B test grey vs. green on the message bar, but since we can't let's just go with [the green style](https://design.firefox.com/photon/components/message-bars.html#color-2) for the "Download Firefox" prompt for non-Fx users.

screen shot 2018-03-23 at 1 46 28 pm

svitlana-galianova:
#4633: I have change the icon of the error message from check mark to Firefox one.

screen shot 2018-04-05 at 3 00 22 pm


tofumatt:
To be clear: the failing build here is because of #4747.

PR #4756, opened 2018-04-06
feat: Add extensions and themes to profile

### LINKED ISSUE #4280: Show "Extensions created" on the user profile page

### LINKED ISSUE #4281: Show "Themes created" on the user profile page

tofumatt:
Fix #4280
Fix #4281

Screenshots:

screenshot 2018-04-06 15 27 54
screenshot 2018-04-06 15 27 52

(they would both show for a user with both extensions and themes, but I couldn't find one 😆)

### DIFF
--- a/tests/unit/amo/components/TestUserProfile.js
+++ b/tests/unit/amo/components/TestUserProfile.js
@@ -1,11 +1,16 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
 
+import AddonsByAuthorsCard from 'amo/components/AddonsByAuthorsCard';
 import UserProfile, { UserProfileBase } from 'amo/components/UserProfile';
 import NotFound from 'amo/components/ErrorPage/NotFound';
 import ReportUserAbuse from 'amo/components/ReportUserAbuse';
 import { fetchUserAccount, getCurrentUser } from 'amo/reducers/users';
 import { createApiError } from 'core/api';
+import {
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_THEME,
+} from 'core/constants';
 import { ErrorHandler } from 'core/errorHandler';
 import ErrorList from 'ui/components/ErrorList';
 import LoadingText from 'ui/components/LoadingText';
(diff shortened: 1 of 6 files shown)


tests/unit/amo/components/TestUserProfile.js
```diff
@@ -290,6 +295,32 @@ describe(__filename, () => {
...
+
+  it('renders AddonsByAuthorsCard for themes', () => {
+    const root = renderUserProfile();
+
+    expect(root.find(AddonsByAuthorsCard).at(1))
+      .toHaveProp('addonType', ADDON_TYPE_THEME);
+  });
+
+  it('renders no AddonsByAuthorsCard if no user found', () => {
+    const root = renderUserProfile({ params: { username: 'not-loaded' } });
```
bobsilverberg: I know this is using existing code, but I find it confusing that passing in the string `not-loaded` into `params.username` results in no user being found. It's not really obvious that that is the case. Looking back at the code in this test, and also in the component, the whole `params` thing seems confusing to me. Why does `UserProfile` accept a `params` object as a property rather than just accepting a `username` as a property? This seems to result in having to jump through hoops in the test like `syncPropsAndParamsUsername`.

This doesn't block this PR at all, but I am confused by all of this so wouldn't mind understanding a bit better via an explanation.

bobsilverberg:
Actually, I did notice one thing when looking at the screen shots that I forgot to mention in my review. I'm not sure it makes sense for the heading to read "More extensions by xxx". It seems that simply "Extensions by xxx" would be better, as the "More" doesn't really make sense in this context.

tofumatt:
Ah, that's true. That's an issue with the `AddonsByAuthorsCard` in general... it has a hardcoded title right now and really we should allow for different ones or overrides. I don't want to introduce more into this PR but I've filed #4774 for that.

PR #4783, opened 2018-04-12
Implement an EditableCollectionAddon component to support the new Edit collections screen

### LINKED ISSUE #4778: Implement an EditableCollectionAddon component to support the new Edit collections screen
As discussed with @pwalm, when a user edits a collection, the list of add-ons displayed on the right-hand side of the screen will change from a standard AddonsCard to a list of editable add-ons with less information (only name and icon), but with a button to remove the add-on from the collection and also a means of adding notes to the add-on.

This will be similar to what appears on the bottom of the form in the current mockups, for example:

screenshot 2018-04-11 10 41 02

This first step is to simply replace the existing list of add-ons with an editable list. The ability to remove an add-on from a collection and to add notes to an add-on will be implemented in follow up issues.

@pwalm will be providing a mockup of this new screen.

bobsilverberg:
Fixes #4778 

Before:

screenshot 2018-04-12 14 57 05

After:

screenshot 2018-04-12 14 56 12


src/amo/components/Collection/index.js
```diff
@@ -234,6 +235,7 @@ export class CollectionBase extends React.Component<Props> {
           <AddonsCard
             addonInstallSource={INSTALL_SOURCE_COLLECTION}
             addons={addons}
+            forCollectionMaintenance={editing}
```
tofumatt: Yeah, seeing this here... that prop should just be called `editing` or `isEditing`.

bobsilverberg: The thing is, I'll be implementing this for "creating" as well very soon, so I was thinking ahead. It doesn't matter if we're editing or creating, in both cases we're "maintaining" and this will be a `true`. So if I call it `editing` now I'll have to change it later. Does that make sense?

bobsilverberg:
The sparseness of this is not only to match the mocks, but also to prepare for the "Create and add add-ons" screen. With that screen when we add an add-on to a not-yet-created collection, the only info we'll have about the add-on is its name, id and iconUrl. Conveniently that's enough to populate this screen until we save the collection. If we want a screen similar to Search Results when creating a collection we'd need to fetch each add-on via an API call in order to display them in the listing, and I'm not sure we want to have to do that.

src/amo/components/Collection/index.js
tofumatt: I still feel like `editing` is a better fit for creating than `maintaining`–I think it's clearer and easier to understand. 🤷‍♂️

bobsilverberg: I don't agree, but it's not that important to me. I'll change it, and if anyone complains when `editing` is also used for `creating` I'll point them to this comment. ;-)

PR #4800, opened 2018-04-13
Add an add-on to a collection from the edit screen

### LINKED ISSUE #4590: Add an add-on to a collection from the edit screen

A user should be able to:

- Sign in
- View an existing collection that they created
- Click *Edit collection* to open the editor overlay
- Type the name of a new add-on in the search box and select a result from the list to add it to the collection

bobsilverberg:
Fixes #4590 

Sorry the PR is so large. The changes to support the feature are fairly large, but I also had to touch a bunch of other files because I had to make a couple of other changes to implement my feature.

These are:
- I needed to add a new const, `DEFAULT_API_PAGE_SIZE` and I thought it best to use that const in all places it should be used. 
- I renamed a helper function from  `createFakeCollectionAddons` to  `createFakeCollectionAddonsListResponse`, because the former wasn't really accurate and I needed to create a new one called `createFakeCollectionAddons` which was accurate.

Those changes resulted in a bunch of extra files being touched, but it should be pretty easy to review those changes.

### DIFF
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -5,6 +5,7 @@ import * as React from 'react';
 
 import EditableCollectionAddon from 'amo/components/EditableCollectionAddon';
 import SearchResult from 'amo/components/SearchResult';
+import { DEFAULT_API_PAGE_SIZE } from 'core/api';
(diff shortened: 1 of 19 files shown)


bobsilverberg:
Thanks for the review @tofumatt :-)

PR #4828, opened 2018-04-17
Fix getAddonsForUsernames so it will work consistently with multiple usernames

### LINKED ISSUE #4818: The "More add-ons by..." card does not appear on the add-on details page when the current add-on has more than one author
STR:

1. Visit the details page for an extension with more than one author, e.g., https://addons-dev.allizom.org/en-US/firefox/addon/easy-passwords/.

Expected:

A card showing "More extensions by these developers" appears towards the bottom of the page (below the Permissions card).

Actual:

The "More extensions by these developers" does not appear on the page.

bobsilverberg:
Fixes #4818 

Prior to this fix, if one of the usernames passed didn't have any add-ons loaded, then the
whole function returned null, even if there were some add-ons loaded for some of the
usernames. This situation can arise because when add-ons are fetched for a set of
usernames not all add-ons are fetched and loaded, so we can end up with a username
without any loaded add-ons.

### DIFF
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -135,20 +135,18 @@ export const getAddonsForUsernames = (
   const ids = usernames.map((username) => {
     return state.byUsername[username];
   }).reduce((array, addonIds) => {
-    if (!addonIds || !array) {
-      return null;
-    }
-
-    for (const addonId of addonIds) {
-      if (!array.includes(addonId)) {
-        array.push(addonId);
+    if (addonIds) {
+      for (const addonId of addonIds) {
+        if (!array.includes(addonId)) {
+          array.push(addonId);
+        }
       }
     }
 
     return array;
   }, []);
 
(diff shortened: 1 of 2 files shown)


src/amo/reducers/addonsByAuthors.js
```diff
@@ -135,20 +135,18 @@ export const getAddonsForUsernames = (
...
     return state.byUsername[username];
   }).reduce((array, addonIds) => {
-    if (!addonIds || !array) {
-      return null;
-    }
-
-    for (const addonId of addonIds) {
-      if (!array.includes(addonId)) {
-        array.push(addonId);
+    if (addonIds) {
```
muffinresearch: Do we have a test covering `addonIds` being undefined?

bobsilverberg: Yes there are a couple, including the one I added for this patch.

bobsilverberg:
Thanks for the review @muffinresearch.

PR #4849, opened 2018-04-19
I4832: number of addons in userprofile page

### LINKED ISSUE #4832: Display more extensions and themes in user profile page
Branched off from #4809.

We should show at least 10 extensions and 10 themes in the user profile page, so it doesn't require too much clicking to cycle through all add-ons for most developers (some theme designers have thousands of those, but we're not optimizing for that case).

It looks like 4 lines of themes in the original mock would show 12 themes, which is sufficient IMO.

tsl143:
Fixes #4832 

Before:
image

After:
screen shot 2018-04-19 at 6 52 57 am


tsl143:
Ohh, my bad, I will update PR soon.

src/amo/components/UserProfile/index.js
```diff
@@ -162,7 +162,7 @@ export class UserProfileBase extends React.Component<Props> {
                 addonType={ADDON_TYPE_EXTENSION}
                 authorDisplayName={[user.displayName]}
                 authorUsernames={[user.username]}
-                numberOfAddons={3}
+                numberOfAddons={10}
```
bobsilverberg: Sorry for not noticing this before, but you've introduced a `EXTENSIONS_BY_AUTHORS_PAGE_SIZE` in the reducer, so we might as well use that value here as well.

src/amo/components/UserProfile/index.js
```diff
@@ -171,7 +171,7 @@ export class UserProfileBase extends React.Component<Props> {
                 addonType={ADDON_TYPE_THEME}
                 authorDisplayName={[user.displayName]}
                 authorUsernames={[user.username]}
-                numberOfAddons={6}
+                numberOfAddons={12}
```
bobsilverberg: As above, we could use `THEMES_BY_AUTHORS_PAGE_SIZE` here.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -346,6 +344,26 @@ describe(__filename, () => {
     sinon.assert.notCalled(dispatchSpy);
   });
 
+  it('should have extensions length equal to numberOfAddons props', () => {
```
bobsilverberg: We don't need these tests, my apologies if my earlier comments suggested this is what was needed. The `numberOfAddons` passed into `AddonsByAuthorsCard` is only used by the `AddonsCard` to determine the number of placeholders, which isn't what's being tested here.

The test that would be useful would be a test of the `UserProfile` component to check that the properties passed into `AddonsByAuthorsCard` is correct for both the card rendered for extensions and the card rendered for themes.

tsl143: @bobsilverberg , following are my points.
1. `numberOfAddons` is not just for placeholders but also for number of addons in the card, have a look here (https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonsByAuthorsCard/index.js#L194), we use `numberOfAddons` to slice out addons.
2. `EXTENSIONS_BY_AUTHORS_PAGE_SIZE` and `THEMES_BY_AUTHORS_PAGE_SIZE` are used for maximum number of addons that will be fetched from API, and then in component based on `numberOfAddons` we show addons.
3. My initial PR was failed because I used `12` in `numberOfAddons` key but didn't change in reducer which was `6`, the testcases written are covering that case for both theme and extension, for example if I put 14 as `numberOfAddons` test value the test should fail as its getting only `12` from reducer.

tsl143:
@bobsilverberg any thoughts about https://github.com/mozilla/addons-frontend/pull/4849#discussion_r185379026

tests/unit/amo/components/TestAddonsByAuthorsCard.js
bobsilverberg: > 1. `numberOfAddons` is not just for placeholders but also for number of addons in the card, have a look here (https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonsByAuthorsCard/index.js#L194), we use `numberOfAddons` to slice out addons.

Ah, yes, I missed the code in `mapStateToProps`. You are correct, so this is good.

> 2. `EXTENSIONS_BY_AUTHORS_PAGE_SIZE` and `THEMES_BY_AUTHORS_PAGE_SIZE` are used for maximum number of addons that will be fetched from API, and then in component based on `numberOfAddons` we show addons.

Ok, given that it seems like the tests should actually request a `numberOfAddons` that is less than the corresponding constant (e.g., `EXTENSIONS_BY_AUTHORS_PAGE_SIZE`) to that we can prove that the list of add-ons rendered is in fact controlled by `numberOfAddons`. Does that make sense?

tsl143: > Ok, given that, it seems like the tests should actually request a numberOfAddons that is less than the corresponding constant (e.g., EXTENSIONS_BY_AUTHORS_PAGE_SIZE) so that we can prove that the list of add-ons rendered is in fact controlled by numberOfAddons. Does that make sense?

Sure that is good, I will update, also there are some conflicts now I will fix up

tsl143:
@bobsilverberg updated :)

src/amo/reducers/addonsByAuthors.js
```diff
@@ -34,7 +35,8 @@ export const initialState: AddonsByAuthorsState = {
   loadingFor: {},
 };
 
-export const ADDONS_BY_AUTHORS_PAGE_SIZE = 6;
+export const EXTENSIONS_BY_AUTHORS_PAGE_SIZE = 10;
+export const THEMES_BY_AUTHORS_PAGE_SIZE = 12;
```
willdurand: You should use these constants in the `UserProfile` component.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -66,24 +68,19 @@ describe(__filename, () => {
...
   function addonsWithAuthorsOfType({ addonType, multipleAuthors = false }) {
     return loadAddonsByAuthors({
-      addons: [
-        {
+      addons: Array(
+        addonType === ADDON_TYPE_THEME ?
+          THEMES_BY_AUTHORS_PAGE_SIZE :
+          EXTENSIONS_BY_AUTHORS_PAGE_SIZE
+      )
+        .fill('')
```
willdurand: Heh, what does this code do exactly? I think a `for` would be more readable.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -105,7 +102,12 @@ describe(__filename, () => {
     );
   }
 
-  function renderAddonsWithType({ addonType, showMore, multipleAuthors = false } = {}) {
+  function renderAddonsWithType({
+    addonType,
+    showMore,
+    multipleAuthors = false,
+    numberOfAddons = 6,
```
willdurand: Where does this `6` come from?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -347,6 +350,26 @@ describe(__filename, () => {
...
+    });
+
+    expect(root.find(AddonsCard).props().addons).toHaveLength(4);
+  });
+
+  it('should have themes length equal to numberOfAddons props', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_THEME,
+      multipleAuthors: false,
+      numberOfAddons: 6,
```
willdurand: This is the default no?

willdurand: ok then could you please use something different than the default value?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
willdurand: there is `fill` with an empty string in between 😉

tsl143: Yes that's deliberate, `Array(<number>)`, creates array with X empty slots, `fill('')` creates dummy value for each slots, and `map` actually inserts the value in it.
I just don't want to use `for` loop.

willdurand: I get that, but it is tricky to grasp because it is unusual.

src/amo/reducers/addonsByAuthors.js
willdurand: mmh but you're using the same values in the `UserProfile` component no?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
tsl143: But isn't that okay?

willdurand: I would like to see a simpler way of doing this, `Array(N).fill('').map()` is not explicit at all, think about newcomers 😉

tests/unit/amo/components/TestAddonsByAuthorsCard.js
tsl143: done

tsl143:
@willdurand I updated the number of addons in test case, is it good to go now?

src/amo/reducers/addonsByAuthors.js
willdurand: I still don't really understand why we do not use these constants in the `UserProfile` component. We are using the exact same values so far because we load/fetch exactly these numbers. I would rather use the constants here.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
willdurand: We don't want to see such hacks in our codebase, I am sorry. I like ES6/Next syntax a lot, but this is not obvious at all. A simple `for` loop before the `return` would be perfect.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -347,6 +350,26 @@ describe(__filename, () => {
...
   });
 
+  it('should have extensions length equal to numberOfAddons props', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_EXTENSION,
+      multipleAuthors: false,
+      numberOfAddons: 4,
+    });
+
+    expect(root.find(AddonsCard).props().addons).toHaveLength(4);
```
willdurand: If `4` is related to `numberOfAddons` above, I would create a variable for that.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -347,6 +350,26 @@ describe(__filename, () => {
...
+  });
+
+  it('should have themes length equal to numberOfAddons props', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_THEME,
+      multipleAuthors: false,
+      numberOfAddons: 3,
+    });
+
+    expect(root.find(AddonsCard).props().addons).toHaveLength(3);
```
willdurand: Same here for the variable `numberOfAddons`

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -347,6 +350,26 @@ describe(__filename, () => {
...
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_EXTENSION,
+      multipleAuthors: false,
+      numberOfAddons: 4,
+    });
+
+    expect(root.find(AddonsCard).props().addons).toHaveLength(4);
+  });
+
+  it('should have themes length equal to numberOfAddons props', () => {
```
willdurand: `it('should display at most numberOfAddons themes', () => {`

tsl143:
@willdurand updated as requested. The build fails though.

src/amo/components/UserProfile/index.js
```diff
@@ -8,6 +8,10 @@ import AddonsByAuthorsCard from 'amo/components/AddonsByAuthorsCard';
 import NotFound from 'amo/components/ErrorPage/NotFound';
 import ReportUserAbuse from 'amo/components/ReportUserAbuse';
 import { fetchUserAccount, getUserByUsername } from 'amo/reducers/users';
+import {
+  EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
+  THEMES_BY_AUTHORS_PAGE_SIZE,
+} from 'amo/reducers/addonsByAuthors';
```
willdurand: Can you put this `import` before the previous one (`amo/reducers/users`)

src/amo/reducers/addonsByAuthors.js
```diff
@@ -3,6 +3,7 @@ import deepcopy from 'deepcopy';
 import invariant from 'invariant';
 
 import { createInternalAddon } from 'core/reducers/addons';
+import { ADDON_TYPE_THEME } from 'core/constants';
```
willdurand: can you put this import before the previous one? (`core/reducers/..`)

src/amo/reducers/addonsByAuthors.js
```diff
@@ -187,11 +189,14 @@ const reducer = (
     }
     case LOAD_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
+      const pageSize = action.payload.addonType === ADDON_TYPE_THEME ?
+        THEMES_BY_AUTHORS_PAGE_SIZE :
+        EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```
willdurand: does that fit on a single line?

```js
const pageSize = action.payload.addonType === ADDON_TYPE_THEME ?
  THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```

src/amo/sagas/addonsByAuthors.js
```diff
@@ -1,9 +1,10 @@
 import { call, put, select, takeEvery } from 'redux-saga/effects';
-import { SEARCH_SORT_TRENDING } from 'core/constants';
+import { ADDON_TYPE_THEME, SEARCH_SORT_TRENDING } from 'core/constants';
 import {
+  EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
   FETCH_ADDONS_BY_AUTHORS,
-  ADDONS_BY_AUTHORS_PAGE_SIZE,
   loadAddonsByAuthors,
+  THEMES_BY_AUTHORS_PAGE_SIZE,
```
willdurand: We usually put the constants first, then the rest and both are alphabetically sorted. Sorry for this..

src/amo/sagas/addonsByAuthors.js
```diff
@@ -18,14 +19,16 @@ export function* fetchAddonsByAuthors({ payload }) {
 
   try {
     const state = yield select(getState);
-
+    const pageSize = addonType === ADDON_TYPE_THEME ?
+      THEMES_BY_AUTHORS_PAGE_SIZE :
+      EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```
willdurand: does that fit on a single line?

```js
const pageSize = addonType === ADDON_TYPE_THEME ?
  THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -4,8 +4,10 @@ import AddonsByAuthorsCard, {
   AddonsByAuthorsCardBase,
 } from 'amo/components/AddonsByAuthorsCard';
 import {
+  EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
   fetchAddonsByAuthors,
   loadAddonsByAuthors,
+  THEMES_BY_AUTHORS_PAGE_SIZE,
```
willdurand: We should have this constant before `fetchAddonsByAuthors`

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -65,25 +67,23 @@ describe(__filename, () => {
   }
 
   function addonsWithAuthorsOfType({ addonType, multipleAuthors = false }) {
+    const addonsLength = addonType === ADDON_TYPE_THEME ?
+      THEMES_BY_AUTHORS_PAGE_SIZE :
+      EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```
willdurand: does that fit on a single line?

```js
const pageSize = addonType === ADDON_TYPE_THEME ?
  THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
```

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -65,25 +67,23 @@ describe(__filename, () => {
...
+    const addons = [];
+
```
willdurand: can you put this blank line before `const addons = [];`

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -65,25 +67,23 @@ describe(__filename, () => {
...
+    for (let i = 0; i < addonsLength; i++) {
+      addons.push({
+        ...fakeAddon,
+        id: i + 1,
+        slug: `foo${i}`,
+        type: addonType,
+        authors: [fakeAuthorOne],
+      });
+    }
```
willdurand: thank you for doing this!

tests/unit/amo/sagas/test_addonsByAuthors.js
```diff
@@ -1,9 +1,10 @@
 import SagaTester from 'redux-saga-tester';
 
 import addonsByAuthorsReducer, {
-  ADDONS_BY_AUTHORS_PAGE_SIZE,
+  EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
   fetchAddonsByAuthors,
   loadAddonsByAuthors,
+  THEMES_BY_AUTHORS_PAGE_SIZE,
```
willdurand: can you put this constant before `fetchAddonsByAuthors`?

tsl143:
@willdurand updated :)

willdurand:
thank you @tsl143!

PR #4854, opened 2018-04-19
Fix rounding of rating stars

### LINKED ISSUE #4784: Average star ratings should be rounded to the nearest half star
Split from #4186 

The problem is that we're only rounding down, rather than rounding to the nearest half-star. So, an average rating of 3.1 should be 3 stars, and an average rating of 3.3 should be 3.5 stars (not 3 stars, as we currently show).

This has been brought up by a couple of developers before, and it is a big disadvantage for those with average ratings close to the top of the half-rating.

bobsilverberg:
Fixes #4784

### DIFF
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -57,13 +57,9 @@ export class RatingBase extends React.Component {
     return [1, 2, 3, 4, 5].map((thisRating) => {
       const props = {
         className: makeClassName('Rating-choice', {
-          'Rating-selected-star': rating && thisRating <= rating,
-          // Half stars are the result of averages rounded to the nearest
-          // 0.5 place. The API should not return floats for your own review
-          // so effectively this only appears in readOnly for now, but there's
-          // nothing stopping the API from supporting half-stars later.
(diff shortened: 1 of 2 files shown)


src/ui/components/Rating/index.js
```diff
@@ -57,13 +57,9 @@ export class RatingBase extends React.Component {
...
       const props = {
         className: makeClassName('Rating-choice', {
-          'Rating-selected-star': rating && thisRating <= rating,
-          // Half stars are the result of averages rounded to the nearest
-          // 0.5 place. The API should not return floats for your own review
-          // so effectively this only appears in readOnly for now, but there's
-          // nothing stopping the API from supporting half-stars later.
-          'Rating-half-star': (rating && thisRating > rating &&
-            thisRating - 0.5 <= rating),
+          'Rating-selected-star': thisRating - rating <= 0.25,
```
muffinresearch: This feels a bit unwieldy but then I guess doing something like `(rating * 2).toFixed() / 2` you've still got to do something that will break that down into displaying each star correctly.

PR #4855, opened 2018-04-21
Fixes #4584: truncated homepage url

### LINKED ISSUE #4584: Long description is not truncated in user info card
Steps to reproduce:
1.Log in to AMO-dev
2.Be sure that you have a long description in user profile
3.From username drop-down menu, chose view profile

Expected results:
The description is truncated and we should have a "View more" button

Actual results:
The description is not truncated and users cannot see complete description

NOTES:
Reproducible also if the homepage is a long address (edge case though)

Please see a screenshot with the issues:
longname

svitlana-galianova:
Hello, 

I have noticed that description is already fixed(broken into several lines), but on the screenshot in the issue #4584, I saw that homepage url also looks a bit out of place. I have truncated it.
screen shot 2018-04-20 at 11 03 42 pm

### DIFF
--- a/src/amo/components/UserProfile/styles.scss
+++ b/src/amo/components/UserProfile/styles.scss
@@ -48,6 +48,12 @@ $avatar-size: 64px;
   }
 }
 
+.UserProfile-homepage {
+  text-overflow: ellipsis;
+  display: block;
+  overflow: hidden;
+}
+
 .UserProfile-rating-average .Rating {
   @include margin(0, 0, 0, -3px);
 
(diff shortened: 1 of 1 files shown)


muffinresearch:
@svitlana-galianova Hi - thanks for the PR. 

Ideally this should have had an issue filed first since whilst it's related to #4584 it's a different case. 

I think it would make sense to discuss with UX the possibility of switching to linking up the text "homepage" or similar rather than to resort to truncation or a long URL.

Would you be interested in filing an issue for the original problem re: the long links being wrapped and then we can discuss it and see if this PR can be adapted to the eventual solution?

muffinresearch:
I'm going to close this until we know the outcome in #4916

PR #4868, opened 2018-04-25
Greenkeeper/base62 2.0.0

muffinresearch:
Fixes #4787

### DIFF
--- a/src/amo/components/App/index.js
+++ b/src/amo/components/App/index.js
@@ -137,8 +137,8 @@ export class AppBase extends React.Component<Props> {
         `Auth token "${encodedTimestamp}" triggered this base62 error: "${base62Error}"`);
       return;
     }
-    // If the encoded timestamp was malformed it will be 0 or negative.
-    if (createdAt <= 0) {
+    // If the encoded timestamp was malformed it will be 0 or negative or NaN
+    if (createdAt <= 0 || Number.isNaN(createdAt)) {
       log.error(oneLine`Got an invalid timestamp from auth token;
         encoded value: ${encodedTimestamp}; decoded value: ${createdAt}`);
       return;
--- a/tests/unit/amo/test_utils.js
+++ b/tests/unit/amo/test_utils.js
@@ -23,7 +23,7 @@ describe('amo/utils', () => {
     });
   });
 
-  describe('amo/utils/getDjangoBase62', () => {
+  describe('getDjangoBase62', () => {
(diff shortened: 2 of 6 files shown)


src/amo/components/App/index.js
```diff
@@ -137,8 +137,8 @@ export class AppBase extends React.Component<Props> {
         `Auth token "${encodedTimestamp}" triggered this base62 error: "${base62Error}"`);
       return;
     }
-    // If the encoded timestamp was malformed it will be 0 or negative.
-    if (createdAt <= 0) {
+    // If the encoded timestamp was malformed it will be 0 or negative or NaN
+    if (createdAt <= 0 || Number.isNaN(createdAt)) {
```
bobsilverberg: Should these conditions be in the reverse order to that `Number.isNaN(createdAt)` shortcuts before checking `createdAt <= 0`, or does that not really matter?

tests/unit/amo/test_utils.js
```diff
@@ -22,4 +22,24 @@ describe('amo/utils', () => {
       expect(getErrorComponent(501)).toEqual(ServerError);
     });
   });
+
+  describe('getDjangoBase62', () => {
+    const base62 = getDjangoBase62();
+
+    it('encodes a example number', () => {
+      expect(base62.encode(535493287492)).toEqual('9QVvGFw');
```
bobsilverberg: As this test uses the same values as the decode test, it might be worth declaring these two values as consts, and then we can be sure that we are using the same values going in both directions and avoid typo related issues.

yarn.lock
```diff
@@ -1234,9 +1234,9 @@ balanced-match@^1.0.0:
   version "1.0.0"
```
bobsilverberg: Did you mean to include this file in the patch? We often don't include these and then push them separately, but I don't think that needs to be a requirement.

tests/unit/amo/test_utils.js
muffinresearch: Good idea 👍

yarn.lock
muffinresearch: I did but I can see it could be easier to leave it out. 

There's a project we can setup which will update lock files for dep changes. Once we do that having to manually push lockfile changes should be a thing of the past.

src/amo/components/App/index.js
muffinresearch: It can still return 0 - I added a test to cover that case.

PR #4869, opened 2018-04-25
Implement a component which uses react-ab to send different Google Analytics pings

### LINKED ISSUE #4866: Implement an AddonRecommendations component which uses TAAR Lite as a data source
Tracker: #4865 

As an initial POC for react-ab, create a component which can be included in a page which sends a different GA ping depending on the cohort the user is in. This will also need code to read and/or set a cookie to maintain the cohort of a user.

bobsilverberg:
Fixes #4866


tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('tracks impressions', () => {
+    const cohort = TAAR_COHORT_INCLUDED;
+    const _cookie = fakeCookie(cohort);
+    const root = render({ _cookie });
+    expect(root).toHaveClassName('AddonRecommendations');
+    // The component must be rendered in order to trigger the onChoice call.
+    root.render();
+    sinon.assert.calledOnce(fakeTracking.sendEvent);
```
muffinresearch: Do you need to call this since the second check implies it?

bobsilverberg: Well, this asserts it was only called once. Maybe not entirely necessary, but seems like a safe thing to do.

src/amo/components/Addon/index.js
```diff
@@ -62,6 +64,7 @@ import './styles.scss';
 
 export class AddonBase extends React.Component {
   static propTypes = {
+    _config: PropTypes.object,
```
willdurand: We usually don't name props with an underscore. I guess it is because it conflicts with the `config` module. I'd personally prefer to import this config module as `defaultConfig` and use `config` as a prop.

src/amo/components/Addon/index.js
```diff
@@ -530,6 +535,10 @@ export class AddonBase extends React.Component {
               numberOfAddons={6}
             />
           )}
+
+          { _config.get('enableAddonRecommendations') && (
```
willdurand: There is an extra space before `_config` 😉

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,106 @@
...
+    _randomizer: Math.random,
+    _tracking: tracking,
+  }
+
+  constructor(props: Props) {
+    super(props);
+    // This will prevent the component from rendering on the server, by only
+    // setting this to true in componentDidMount, which is only run on the
+    // client.
+    this.shouldRender = false;
```
willdurand: Why exactly do we want to use client side rendering only? Because `react-ab` does not really support SSR or is there something else I am missing?

tests/unit/amo/components/TestAddonRecommendations.js
bobsilverberg: As discussed, I will change it to use `calledOnceWith` to replace both assertions.

src/amo/components/Addon/index.js
bobsilverberg: This is a pattern we have used quite a bit, using an underscore in a prop name to indicate we're using it to inject a dependency. We could do what you suggest, renaming the module on import, but then for consistency we should think about changing it elsewhere, and there are a lot of elsewheres.

src/amo/components/AddonRecommendations/index.js
bobsilverberg: The problem is that the call to `tracking.sendEvent` in `onChoice` fails if it's called on the server, and we don't end up recording the impression. We thought this was a good approach, but have since discussed it and decided it would be better to adjust the tracking module to be smart enough to not try to do some thing it cannot do. I was initially worried that we might miss recording the impression if `render` is only called on the server, but I now gather that it will always be called on the client after the file is downloaded.

bobsilverberg:
Thanks for the review, @willdurand!

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,96 @@
...
+  static defaultProps = {
+    _cookie: cookie,
+    _randomizer: Math.random,
+    _tracking: tracking,
+  }
+
+  componentWillMount() {
+    const { _cookie, _randomizer } = this.props;
+
+    // Set a cohort for the experiment.
```
bobsilverberg: This appears to be working when I test it locally, but we're not 100% sure this will actually work when rendered on the server. What do you think @willdurand?

src/amo/components/Addon/index.js
willdurand: I do remember we have that for some non-component files but I thought react components did not have such props with underscore, sorry for the noise then.

src/core/tracking.js
```diff
@@ -83,7 +83,9 @@ export class Tracking {
   }
 
   log(...args) {
-    this._log.info(this.logPrefix, ...args);
+    if (this._log) {
```
willdurand: This change, I guess, needs a test case.

src/amo/components/AddonRecommendations/index.js
willdurand: If that works for you locally, why would it be a problem given that you should also run it on the server locally?

willdurand: ah no wait, this `componentWillMount()` method only works on the server right? So this won't work when rendered on the client if I am not mistaken.

bobsilverberg: That doesn't appear to be the case. If I turn off SSR componentWillMount is still called. There is a note in the [docs](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) that says "This is the only lifecycle hook called on server rendering.", but I interpret that as "out of all the lifecycle hooks, only this is available on the server", but not necessarily as "this is only available on the server".

src/core/tracking.js
bobsilverberg: I agree.  I tried to figure out how to write a test for this, and it's pretty confusing. I think we'd have to use DI for `this._log`, but then that would generally default `this._log` to `log` and therefore we'd no longer be able to check for `this._log`'s existence in the `log` function. I suppose we could reset `this._log` to undefined if `window` is undefined in the constructor, but this all starts to sound like a lot of unnecessary work.

Maybe this approach isn't the best fix. Perhaps it would be better to not `return` at line 47 in the constructor when `typeof window === 'undefined'`, and allow the code to continue to run, and then set `this.enabled = false` if `typeof window === 'undefined'`. I think that would make testing simpler. What do you think @willdurand?

willdurand: if you `null` the parameter, isn't it supposed to work? the default value should not be used because it is not undefined, but the condition should be fired, no?

bobsilverberg: Maybe I didn't describe the problem very well. The code we're trying to test is:

```js
  log(...args) {
    if (this._log) {
      this._log.info(this.logPrefix, ...args);
    }
  }
```

and the code path we want to test is where `this._log` is falsey. *But* in order to verify that `this._log.info()` was not called, we need `this._log` to be a mock, which is not falsey. So I don't see how it's possible to test this exact code, which suggests to me that it's not a good design.

This is a small enough issue that I'd be happy to just let it go, knowing it works because tests that failed without it pass with it, but looking at the code for `tracking.js` it also seems like not the best approach. I thought maybe the approach I suggested above (which involves not bailing out of the constructor immediately if `window` is undefined) might be better. It's not really a requirement that we not log when `window` doesn't exist. I just added that "fix" because it addressed the issue we had at the time, but that issue was based on the current implementation of `Tracking` which I now feel is sub-optimal.

What do you think?

We could also land this as is and file a follow-up bug to address the missing coverage by improving upon `Tracking`.

willdurand: Ah no, sorry it is me. I skipped the paragraph about the constructor being weird..

I am not a fan of landing things and fix later because we might never come back.

Looking at the code and the tests, if we inject `_log` in the constructor (`_log = log`), can't we smoke test `log()` and ensures it does not output an error? I dont think we can assert on `this._log.info()` but we could maybe invert the expectation and assert for no error when there is no `this._log`. That would cover your change.

bobsilverberg: Ok I added a smoke test for this. It's a bit weird but I guess it does the trick.

src/amo/components/AddonRecommendations/index.js
bobsilverberg: @willdurand Do we need to discuss or investigate this further, or do you agree with my assumptions above, based on what I observed?

willdurand: no, you're 100% right.

tests/unit/core/test_tracking.js
```diff
@@ -30,6 +30,14 @@ describe('Tracking', () => {
     window.ga = sinon.stub();
   });
 
+  it('should be able to call log when uninitialized without an error', () => {
+    const tracking = createTracking({
+      _log: null,
+    });
+    // This should not throw an error.
+    tracking.log('A message that will never be sent');
```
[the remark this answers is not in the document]
bobsilverberg: Nice! Much better, thanks. 😄

bobsilverberg:
@willdurand You didn't give this an explicit approval. I think you're good with it based on your comments, but if you wouldn't mind approving the review then I'll know for sure and we can get it merged.

tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,98 @@
...
+  it('uses an existing cookie to determine the variant', () => {
+    const cohort = TAAR_COHORT_INCLUDED;
+    const _cookie = fakeCookie(cohort);
+    const root = render({ _cookie });
+    expect(root).toHaveClassName('AddonRecommendations');
+    expect(root.render().find('h1').hasClass(cohort)).toEqual(true);
+  });
+
+  it('uses the randomizer to set the cohort and cookie when a cookie does not exist', () => {
+    const checkWithRandomizer = (randomNumber, expectedCohort) => {
```
willdurand: Maybe `checkWithFixedRandomizer`?

willdurand: (could not come up with something better, yet I find this function name a bit obscure)

bobsilverberg:
Oops, I didn't see that @willdurand had added a bunch of comments to the last iteration of the code. I will address them in the next PR I issue.

PR #4875, opened 2018-04-26
issue 4830 profile style fix

### LINKED ISSUE #4830: User profile page is not correctly displayed on small devices in portrait orientation
STR:
1. Load AMO dev with FF59 (Win10) with an account that has submitted extensions/themes.
2. Click on View Profile from the drop down menu below user's name.
3. Observe page.

Expected result:
More extensions and More themes sections are displayed correctly.

Actual result:
Sections are truncated. 

Notes:
This issue is not reproducible when the android device is in landscape view.
This issue is reproducible on AMO dev with FF59( Android 8.0).

android user info card

rebmullin:
fixes #4830 - user profile page style fixes

BEFORE EX:
before fixes

AFTER EX:
After fixes

### DIFF
--- a/src/amo/components/AddonsByAuthorsCard/styles.scss
+++ b/src/amo/components/AddonsByAuthorsCard/styles.scss
@@ -20,7 +20,7 @@
       display: grid;
       grid-auto-flow: initial;
       grid-gap: 6px;
-      grid-template-columns: repeat(3, 1fr);
+      grid-template-columns: repeat(3, minmax(32%, 1fr));
     }
   }
 }
--- a/src/amo/components/UserProfile/styles.scss
+++ b/src/amo/components/UserProfile/styles.scss
@@ -14,8 +14,11 @@ $avatar-size: 64px;
 }
 
 .UserProfile-wrapper {
-  display: flex;
-  justify-content: flex-start;
+
+  @include respond-to(large) {
+    display: flex;
+    justify-content: flex-start;
+  }
 }
 
 .UserProfile-user-info {
@@ -63,8 +66,17 @@ $avatar-size: 64px;
 }
 
 .UserProfile-addons-by-author {
-  @include margin(0, 0, 0, 12px);
+  @include margin(12px, 0, 0, 0);
 
-  max-width: 60%;
   width: 100%;
+
+  @include respond-to(large) {
+    @include margin(0, 0, 0, 12px);
+
+    max-width: calc(100% - 300px - 12px);
+  }
+
+  @include respond-to(extraLarge) {
+    max-width: calc(65% - 12px);
+  }
 }


src/amo/components/AddonsByAuthorsCard/styles.scss
```diff
@@ -20,7 +20,7 @@
       display: grid;
       grid-auto-flow: initial;
       grid-gap: 6px;
-      grid-template-columns: repeat(3, 1fr);
+      grid-template-columns: repeat(3, minmax(32%, 1fr));
```
bobsilverberg: I'm not sure what adding the `minmax` for this `horizontal` class does for us? What is the specific case we are protecting against?

src/amo/components/UserProfile/styles.scss
```diff
@@ -63,8 +66,17 @@ $avatar-size: 64px;
...
   width: 100%;
+
+  @include respond-to(large) {
+    @include margin(0, 0, 0, 12px);
+
+    max-width: calc(100% - 300px - 12px);
+  }
+
+  @include respond-to(extraLarge) {
+    max-width: calc(65% - 12px);
```
bobsilverberg: Why do we need a special `max-width` for `extraLarge`?

src/amo/components/AddonsByAuthorsCard/styles.scss
rebmullin: hey @bobsilverberg - i was seeing some weirdness at smaller breakpoints:

Alt text

this seems to do similar thing as original style except it gives it a min so cols are distributed evenly even in ~ mobile where this didn't seem to be happening

src/amo/components/UserProfile/styles.scss
rebmullin: basically when screen size is over 900px (`extraLarge`) the left side is over 300px so to coordinate the right side with that i adjusted max-width to work with the percentage (vs the 100px -300px that is set to work with the `large` bp) .. does that make sense?

src/amo/components/AddonsByAuthorsCard/styles.scss
bobsilverberg: Thanks for the example, @rebmullin. That does illustrate what the problem is, but I still don't understand why adding a min column size fixes that. Do you know why?

src/amo/components/UserProfile/styles.scss
bobsilverberg: Kinda sorta. Where did the number `65%` come from?

bobsilverberg:
Thanks for the answers to my questions @rebmullin. As I approved this PR, you can merge it. As long as a PR is approved by a reviewer, and is passing on CI, it's fine for you to merge it yourself. If there is more than one commit attached please be sure to use the "Squash and merge" option, rather than the "Merge pull request" option.

src/amo/components/UserProfile/styles.scss
muffinresearch: @rebmullin Would be really handy to add a comment with an explanation for what the magic numbers used for things like this, might save some future head-scratching.

src/amo/components/AddonsByAuthorsCard/styles.scss
rebmullin: honestly i am not too sure why the initial code doesn't work but by using this minmax the min val enforces cols size (specifically col 2 and 3) to be redistributed as expected. There are probably different ways to approach this - with max value too or via fit-content.

PR #4879, opened 2018-04-27
issue 4774 addsonauthorscard show/hide More text option

### LINKED ISSUE #4774: AddonsByAuthorsCard always says "More add-ons by user"
See https://github.com/mozilla/addons-frontend/pull/4756#issuecomment-379751761

Basically we should allow an override to the default header here. I think the defaults make sense but an override would be handy.

rebmullin:
fixes #4774  - sets up AddsonByAuthorsCard's so header text will be "More xxx by..." by default or can be changed via bool prop to "xxx by.." 

### DIFF
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -31,6 +31,7 @@ import './styles.scss';
 type Props = {|
   addons?: Array<AddonType>,
   addonType?: string,
+  header?: string,
   authorDisplayName: string,
   authorUsernames: Array<string>,
   className?: string,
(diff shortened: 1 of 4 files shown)


rebmullin:
oh! I see.. thanks for clarifying. Iet me take another look

rebmullin:
hey @bobsilverberg  - made some changes here - please take another look when you can?

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -105,61 +107,81 @@ export class AddonsByAuthorsCardBase extends React.Component<Props> {
...
 
     let header;
+    const headerModifier = showMore ? 'More' : '';
     switch (addonType) {
       case ADDON_TYPE_DICT:
         header = i18n.ngettext(
           i18n.sprintf(
-            i18n.gettext('More dictionaries by %(author)s'),
-            { author: authorDisplayName }
+            i18n.gettext('%(headerModifier)s dictionaries by %(author)s'),
```
bobsilverberg: Unfortunately we cannot do it this way, because we cannot assume anything about where the `headerModifier` will be placed in other languages, plus we need the headerModifier to be able to be translated into other languages.

So we need to wrap the full string, in both cases, in `i18n.gettext`. This makes for a bit more code, but I don't think there's anything we can do to avoid it. You basically need to wrap both the `Dictionaries by %(author)s` and the `More dictionaries by %(author)s` separately in `i18n.gettext`.

rebmullin: hmm...ok ...i'll try again : )

rebmullin:
hey @bobsilverberg  - i made some changes here. how's it look now?

rebmullin:
cool @bobsilverberg  but can you please clarify your update request - do you mean it should NOT have the More text here:

Alt text

bobsilverberg:
> do you mean it should NOT have the More text here...

@rebmullin correct! also for extensions on the profile page.

rebmullin:
@bobsilverberg  - i think it should be good now 🤞  : ) can you please take another look when you can?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -445,4 +446,26 @@ describe(__filename, () => {
...
+  it('shows header text without More text', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_THEME,
+      showMore: false,
+      multipleAuthors: false,
+    });
+
+    expect(root.find(AddonsCard))
+      .toHaveProp('header', `Themes by ${fakeAuthor.name}`);
+  });
```
willdurand: I would like to see more test cases, ideally one for each addon type and number of authors.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -445,4 +446,26 @@ describe(__filename, () => {
...
+
+  it('shows header text starting with More text as default', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_THEME,
+      multipleAuthors: false,
+    });
+
+    expect(root.find(AddonsCard))
+      .toHaveProp('header', `More themes by ${fakeAuthor.name}`);
+  });
```
willdurand: Is this needed since we have existing test cases?

rebmullin: yep -you're right...it's a dupe test and not needed here, thanks

rebmullin:
hey @willdurand  / @bobsilverberg  - I made some updates here... please take another look when you can?

willdurand:
I take it.

PR #4892, opened 2018-04-28
I4880: remove review headers

### LINKED ISSUE #4880: Remove the h3 for the review titles since they aren't added as part of the form.
In https://github.com/mozilla/addons-frontend/issues/1739 review titles were removed from the review form. 

However the display on the all reviews page shows empty h3s for the reviews page.

reviews_for_ublock_origin_ _add-ons_for_firefox

@jvillalobos Do we need to keep displaying titles for historical reviews or can this markup be completely removed?

Even if we do want to keep titles for historical reviews that have that data, we could at least avoid outputting empty markup.

tsl143:
Fixes #4880 [code quality]

### DIFF
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -225,9 +225,6 @@ export class AddonReviewListItemBase extends React.Component<Props> {
 
     return (
       <div className="AddonReviewListItem">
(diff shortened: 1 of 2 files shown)


muffinresearch:
We should hold on landing this until the request for info in #4880 is concluded.

muffinresearch:
Looks good - thanks for the patch @tsl143 👍

PR #4896, opened 2018-04-30
i-4054 remove dev reply url to their own reply

### LINKED ISSUE #4054: Unusable "Reply to this review" link option for a Developer response
Prerequisites:
User is logged with an admin account;

Steps to reproduce:
1. Load "All reviews" page for an add-on i.e. https://addons.allizom.org/en-US/firefox/addon/clip-to-onenote/reviews/
2. Observe link options for a Developer response

Expected results:
No "Reply to this review" option available.

Actual results:
Unusable "Reply to this review" link option.

Notes/Issues:
- a reply to a developer response is not possible - but still display the link; 
- also 2 developers replies for same review are not possible - that is why we **do not list** the "reply to this review" link in this case.

Verified on AMO-stage FF57(Win 7, Android 7)
Screenshot for this issue:

rebmullin:
fixes #4054  - remove reply to review link on dev's reply
BEFORE:

AFTER:
Alt text

### DIFF
--- a/tests/unit/amo/components/TestAddonReviewListItem.js
+++ b/tests/unit/amo/components/TestAddonReviewListItem.js
@@ -526,14 +526,6 @@ describe(__filename, () => {
     expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
   });
 
-  it('hides the reply-to-review link on the developer reply', () => {
-    const developerUserId = 3321;
-    const { addon } = signInAsAddonDeveloper({ developerUserId });
-    const { reply } = _setReviewReply();
-    const root = renderReply({ reply, addon });
-    expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
-  });
-
   it('ignores other review related view actions', () => {
     const thisReview = _setReview({ ...fakeReview, id: 1 });
     const anotherReview = _setReview({ ...fakeReview, id: 2 });
@@ -630,6 +622,13 @@ describe(__filename, () => {
       expect(replyComponent).toHaveProp('addon', addon);
       expect(replyComponent).toHaveProp('review', reply);
       expect(replyComponent).toHaveProp('isReplyToReviewId', review.id);
+    });
+
+    it('hides the reply-to-review link on the developer reply', () => {
+      const developerUserId = 3321;
+      const { addon } = signInAsAddonDeveloper({ developerUserId });
+      const root = renderReply({ addon });
+      expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
     });
 
     it('hides a nested reply when editing it', () => {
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -526,6 +526,14 @@ describe(__filename, () => {
     expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
   });
 
+  it('hides the reply-to-review link on the developer reply', () => {
```
bobsilverberg: Thanks for adding the test @rebmullin. I think this test would make more sense nested inside the `describe('Developer reply to a review')`.

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -526,6 +526,14 @@ describe(__filename, () => {
...
+    const developerUserId = 3321;
+    const { addon } = signInAsAddonDeveloper({ developerUserId });
+    const { reply } = _setReviewReply();
+    const root = renderReply({ reply, addon });
```
bobsilverberg: Because all we are checking is that the reply link is not displayed for any reply, we don't really need to create an `addon` or a `reply` for this test. I think we can just call `renderReply()` as is done in the test for `it('hides ratings for replies')`.

rebmullin: I initially was wondering if we needed a test for this b/c i do see a similar test to what you suggest above but the problem was it wasn't failing there (and we would expect that it would fail - right? - since it was showing this link in the reply). This test always passed. 

With the code changes the only way i could get it to both pass and fail as expected (for testing purposes) was to log in as a developer

I think i can clean it up tho if that helps:

```javascript
  it('hides the reply-to-review link on the developer reply', () => {
      const developerUserId = 3321;
      const { addon } = signInAsAddonDeveloper({ developerUserId });
      const root = renderReply({ addon });
      expect(root.find('.AddonReviewListItem-begin-reply')).toHaveLength(0);
    });
```
i *think* this is b/c renderReply sets up the reply as a regular user so given that we wouldn't ever see this so will always be 0 like you noted above

let me know if that makes any sense or still if there is a better way to address the test : )

tests/unit/amo/components/TestAddonReviewListItem.js
rebmullin: thanks @bobsilverberg  - good pt

tests/unit/amo/components/TestAddonReviewListItem.js
rebmullin: @bobsilverberg  - hopefully i am not over-pinging you now : ) but please review partial update and comment above regarding your suggestion

bobsilverberg: Ah, I hadn't thought that through. I just saw the code and thought we didn't need it, but yeah, if the reply link won't render without a signed-in developer then we definitely need it. Good job making sure the test fails without your patch, as well as passes with your patch!

bobsilverberg:
Note that if we want PRs to auto-close the associated issue, a comment needs to include exactly "fixes #4054" or "closes #4054", other variations, and ones without the "#" won't do it. That's why we generally include "fixes #4054" as the first line in the PR comment.

rebmullin:
ok- good to know. i was wondering about that (how issues get closed).

PR #4901, opened 2018-04-30
Fix ui test timeouts

jrbenny35:
This should fix the "Timed Out" failures on circleci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,7 @@ workflows:
   scheduled-workflow:
     triggers:
       - schedule:
-          cron: "0 0,2 * * *"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -123,7 +123,7 @@ workflows:
   scheduled-workflow:
     triggers:
       - schedule:
-          cron: "0 0,2 * * *"
+          cron: "0 0,12,15,18,21 * * *"
```
bobsilverberg: Just wondering, why this change?

jrbenny35: The integration tests job runs at certain times. This should allow it to run more throughout the development day instead of just at 0:00 UTC time.

PR #4904, opened 2018-05-01
Implement API, saga and reducer code to interact with TAAR Lite service via addons-server

### LINKED ISSUE #4867: Implement API, saga and reducer code to interact with TAAR Lite service via addons-server
Tracker: #4865

- [ ] Update API code to call the TAAR Lite endpoint, which will be [provided by addons-server](https://github.com/mozilla/addons-server/issues/8118).
- [ ] Update saga code to dispatch the API request and load the results into state.
- [ ] Update reducer code to introduce actions for retrieving and loading TAAR Lite results.

bobsilverberg:
Fixes #4867


src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,175 @@
...
+
+type AbortFetchRecommendationsAction = {|
+  type: typeof ABORT_FETCH_RECOMMENDATIONS,
+  payload: AbortFetchRecommendationsParams,
+|};
+
+export const abortFetchRecommendations = ({
+  guid,
+}: AbortFetchRecommendationsParams = {}): AbortFetchRecommendationsAction => {
+  return {
```
willdurand: no `invariant` here?

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,175 @@
...
+  = 'LOAD_RECOMMENDATIONS';
+
+export type Recommendations = {|
+  addons: Array<ExternalAddonType>,
+  fallbackReason: string,
+  loading: boolean,
+  outcome: string,
+|};
+
+export type RecommendationsState = {
```
willdurand: I think we could have a stricter type here.

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,175 @@
...
+
+type GetRecommendationsByGuidParams = {|
+  guid: string,
+  state: RecommendationsState,
+|};
+
+export const getRecommendationsByGuid = (
+  { guid, state }: GetRecommendationsByGuidParams
+): Recommendations | null => {
+  invariant(guid, 'guid is required');
```
willdurand: `state` should be required too, right?

tests/unit/amo/api/test_recommendations.js
```diff
@@ -0,0 +1,37 @@
+import * as api from 'core/api';
+import { getRecommendations } from 'amo/api/recommendations';
+import { createApiResponse } from 'tests/unit/helpers';
+import {
+  dispatchClientMetadata,
+} from 'tests/unit/amo/helpers';
```
willdurand: nitpick: this could be one a single line.

tests/unit/amo/reducers/test_recommendations.js
```diff
@@ -0,0 +1,84 @@
...
+  it('sets the loading flag when fetching recommendations', () => {
+    const guid = 'some-guid';
+    const state = reducer(undefined, fetchRecommendations({
+      errorHandlerId: createStubErrorHandler().id,
+      guid,
+      recommended: true,
+    }));
+
+    expect(state.byGuid[guid].loading).toEqual(true);
+    expect(state.byGuid[guid].addons).toEqual(undefined);
```
willdurand: According to the flow def, there is always an `addon` array that can be empty.

tests/unit/amo/reducers/test_recommendations.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+
```
willdurand: Extra blank line

tests/unit/amo/reducers/test_recommendations.js
```diff
@@ -0,0 +1,84 @@
...
+      fallbackReason,
+      guid,
+      outcome,
+    }));
+
+    const expectedAddons = addons.map((addon) => createInternalAddon(addon));
+
+    const loadedRecommendations = getRecommendationsByGuid({ guid, state });
+
+    expect(loadedRecommendations).not.toEqual(null);
```
willdurand: I understand this line but it is (implicitly) covered by the line below no?

tests/unit/amo/sagas/test_recommendations.js
```diff
@@ -0,0 +1,105 @@
...
+import apiReducer from 'core/reducers/api';
+import { createStubErrorHandler } from 'tests/unit/helpers';
+import {
+  fakeAddon,
+  dispatchClientMetadata,
+} from 'tests/unit/amo/helpers';
+
+
+describe(__filename, () => {
+  let clientData;
```
willdurand: Is this really necessary?

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,175 @@
...
+
+  return state.byGuid[guid] || null;
+};
+
+type Action =
+  | AbortFetchRecommendationsAction
+  | FetchRecommendationsAction
+  | LoadRecommendationsAction
+  ;
```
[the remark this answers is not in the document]
bobsilverberg: I think I got that from https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/collections.js#L724, but you're right, pretty much everywhere else we put it on the end of the lint.

src/amo/reducers/recommendations.js
bobsilverberg: We were unable to make this an exact type because of https://github.com/facebook/flow/issues/2405, but I see that should be fixed in Flow 0.71, which is landing in https://github.com/mozilla/addons-frontend/pull/4917, so let's get that landed and then I think I can update this.

tests/unit/amo/reducers/test_recommendations.js
bobsilverberg: I updated that so it can be null, and updated this test.

tests/unit/amo/reducers/test_recommendations.js
bobsilverberg: True. I'll remove it.

tests/unit/amo/sagas/test_recommendations.js
bobsilverberg: Nope. :)

bobsilverberg:
Thanks for the review @willdurand. I'm submitting my changes for you to have another look. This is currently failing the Flow check as it will need Flow 0.71, so ignore the CI failures for now. I won't land it until those are fixed.

willdurand:
@bobsilverberg thanks, you can rebase now if you want.

bobsilverberg:
This is ready for another review, thanks @willdurand.

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,182 @@
...
+|};
+
+type AbortFetchRecommendationsAction = {|
+  type: typeof ABORT_FETCH_RECOMMENDATIONS,
+  payload: AbortFetchRecommendationsParams,
+|};
+
+export const abortFetchRecommendations = ({
+  guid,
+}: AbortFetchRecommendationsParams = {}): AbortFetchRecommendationsAction => {
```
willdurand: I would not specify a default empty object here because `guid` is mandatory.

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,182 @@
...
+export type FetchRecommendationsAction = {|
+  type: typeof FETCH_RECOMMENDATIONS,
+  payload: FetchRecommendationsParams,
+|};
+
+export const fetchRecommendations = ({
+  errorHandlerId,
+  guid,
+  recommended,
+}: FetchRecommendationsParams = {}): FetchRecommendationsAction => {
```
willdurand: Here too, I would not add the default empty object. It's not really clear to me why we have some actions with default empty objects and others without... I would be in favor of not doing this because it would raise a warning/error in development.

src/amo/store.js
```diff
@@ -51,6 +52,7 @@ export default function createStore({
       landing,
       languageTools,
       redirectTo,
+      recommendations,
```
willdurand: Shouldn't this be before `redirectTo`?

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,182 @@
...
+
+export const ABORT_FETCH_RECOMMENDATIONS: 'ABORT_FETCH_RECOMMENDATIONS'
+  = 'ABORT_FETCH_RECOMMENDATIONS';
+export const FETCH_RECOMMENDATIONS: 'FETCH_RECOMMENDATIONS'
+  = 'FETCH_RECOMMENDATIONS';
+export const LOAD_RECOMMENDATIONS: 'LOAD_RECOMMENDATIONS'
+  = 'LOAD_RECOMMENDATIONS';
+
+export type Recommendations = {|
+  addons: Array<ExternalAddonType> | null,
```
willdurand: Because you use `createInternalAddon()` in the reducer, I suppose the type should be `AddonType`.

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,182 @@
...
+|};
+
+export const loadRecommendations = ({
+  addons,
+  fallbackReason,
+  guid,
+  outcome,
+}: LoadRecommendationsParams = {}): LoadRecommendationsAction => {
+  invariant(addons, 'addons is required');
+  invariant(fallbackReason, 'fallbackReason is required');
```
willdurand: The API documentation says the `fallback_reason` is "nullable": http://addons-server.readthedocs.io/en/latest/topics/api/addons.html#recommendations

src/amo/reducers/recommendations.js
```diff
@@ -0,0 +1,182 @@
...
+export const FETCH_RECOMMENDATIONS: 'FETCH_RECOMMENDATIONS'
+  = 'FETCH_RECOMMENDATIONS';
+export const LOAD_RECOMMENDATIONS: 'LOAD_RECOMMENDATIONS'
+  = 'LOAD_RECOMMENDATIONS';
+
+export type Recommendations = {|
+  addons: Array<ExternalAddonType> | null,
+  fallbackReason: string | null,
+  loading: boolean,
+  outcome: string | null,
```
willdurand: Do we want to specify the allowed values? [The API documentation](http://addons-server.readthedocs.io/en/latest/topics/api/addons.html#recommendations) describes the following possible values:

> outcome (string) – Outcome of the response returned. Will be either: recommended - responses from recommendation service; recommended_fallback - service timed out or returned empty results so we returned fallback; curated - recommended=False was requested so fallback returned.

src/amo/sagas/recommendations.js
```diff
@@ -0,0 +1,48 @@
...
+      = recommendations;
+
+    yield put(loadRecommendations({
+      addons,
+      fallbackReason,
+      guid,
+      outcome,
+    }));
+  } catch (error) {
+    log.warn(`Failed to fetch user collections: ${error}`);
```
willdurand: s/user collections/recommendations

src/amo/store.js
willdurand: In addition, I noticed you have added the reducer here. Do you want to also add the saga in the [`rootSaga`](https://github.com/mozilla/addons-frontend/blob/2111ff06c6a9499d39a7d1db29d828285b6986ee/src/amo/sagas/index.js#L23)?

src/amo/reducers/recommendations.js
bobsilverberg: Agreed. That was code borrowed from another reducer and modified, but with required arguments there's no reason to default to `{}`.

src/amo/reducers/recommendations.js
bobsilverberg: Ah, correct, good call!

src/amo/reducers/recommendations.js
bobsilverberg: Because the app doesn't care what the value is (i.e., all it does is record the value, but it never makes any decisions based on the value), I thought it might make sense to leave this open. Theoretically the API could change the strings, or add a string, and it shouldn't break anything.

However this does seem to go against our use of Flow, so I will update this to a type which allows specific strings, and I'll do the same for `fallbackReason`.

bobsilverberg:
Oops, I notice that I didn't use the new types for `fallbackReason` and `outcome` in [`LoadRecommendationsParams`](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/recommendations.js#L81-L83).  I'll address that in PR #4942.

PR #4912, opened 2018-05-02
Fix the regression introduced by #4909

bobsilverberg:


src/amo/components/Addon/index.js
```diff
@@ -338,6 +338,29 @@ export class AddonBase extends React.Component {
     /* eslint-enable react/no-danger */
   }
 
+  renderAddonsByAuthorsCard(isForTheme) {
```
willdurand: Could we use object destructuring here?

```js
renderAddonsByAuthorsCard({ isForTheme }) { // ...
```

so that we can call it as follows:

```js
{this.renderAddonsByAuthorsCard({ isForTheme: true })}
```

willdurand: My idea is to avoid method calls using boolean arguments without context.

bobsilverberg: Yes, good idea. 👍  Fixed.

Issue #3056, opened 2017-09-01
Remove theme previews


muffinresearch:
Since future web-ext based lightweight themes don't allow preview and install and uninstall is quick and easy I'd like to propose we remove the theme preview feature everywhere.

@jvillalobos are you ok for us to do this sooner rather than later?

jvillalobos:
Yes, I'm okay with doing this. It'll require an announcement in the blog or the forum, so I just need to be pinged when we're ready to ship this.

muffinresearch:
I think this one makes sense to fold into https://github.com/mozilla/addons-frontend/issues/3388

rebmullin:
Hey @muffinresearch

Just digging into this a bit per our discussion and I want to make sure I’m looking in the right place:

![Alt text](https://monosnap.com/image/MC8bNf0e3kE88Sanzu2GP7PpQP5Yka.png)

is this right? I couldn’t find this under addons-frontend however i do see it under the addons-server repo. Assuming this is correct do we need to move this issue over there for tracking? Or let me know if I am missing something : )

muffinresearch:
@rebmullin it's the preview feature on the front end e.g:

<img width="987" alt="spring_garden_serenity_by_madonna_ _add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/39517727-9fee9218-4df8-11e8-89c1-fdd4790611c4.png">

rebmullin:
@muffinresearch  - ahh i see..cool, thank you

muffinresearch:
@rebmullin The preview feature also exists on the disco pane too: if you got to about:addons. I think it makes sense to remove it from there too. @jvillalobos are you ok with that?

<img width="1143" alt="add-ons_manager" src="https://user-images.githubusercontent.com/1514/39539728-5be0f094-4e38-11e8-9cbc-8c66589c03f9.png">

muffinresearch:
If it's easier to split out the disco pane change into a separate issue patch that's fine, we just need to make sure we don't ship a broken disco pane in between.

jvillalobos:
Yes, I agree we should remove it from disco pane too.

rebmullin:
@muffinresearch @jvillalobos  so theme description displays here instead - is this okay? it looks something like this on disco v:
![Alt text](https://monosnap.com/image/jawrwTxCyeUDmTNwmc3ULg5JxQPyQ9.png)

AlexandraMoga:
[...]

Issue #4899, opened 2018-04-30
Unify the display of Lightweight and static themes


muffinresearch:
To faciliate the handling of static themes on addons-frontend we need to unify the display of lightweight (LWT) and static themes.

## Key differences

* Static themes have a single static "baked" preview image. LWTs display the main theme image which is truncated by the container. This will impact the unification of the main image.
* Static themes are installed the same way a web extension is installed. LWT themes use a special API that reads a blob of JSON that defines the theme's assets.

## UX needed

First we need UX that Unifies the following areas to minimise the differences

* Search results
* Shelves with theme results.
* Details page

pwalm:
Thank you for filing this! On it.

jvillalobos:
There was some discussion around this in https://github.com/mozilla/addons-frontend/issues/4661. The main thing I want to make sure we do is not crop the preview image vertically. There are cases where there's horizontal cropping, inevitably, but that's a similar problem for LWT. 

Other than the preview image, I hope we can keep both theme types nearly identical, to avoid confusing users too much about them.

Issue #3954, opened 2017-11-21
Re-implement mozAddonManager features in the install button


muffinresearch:
We need to think about what features of mozAddonManager would be useful to add back to the install button on AMO.

Before we can implement something we need to know from product and UX what features we want to add back and what they should look like.

Here's a few of the features to consider.

* Download progress
* Reflect installation state
* Uninstallation

jvillalobos:
Out of those 3, I don't particularly care about download progress. User research points to uninstallation being very important, and I don't think we can do it without installation state, so I think we should work on those 2.

Unrelated to `mozAddonManager`, but related to the look of the install UI, we also got feedback from user research that we should indicate that an add-on is free to install.

andymckay:
The installation state is really helpful as well for moving stuff out of about:addons, eg: search, so users can tell if that add-on is installed. The inbetween states less so for me.

pwalm:
Proposed states:
<img width="792" alt="screen shot 2017-12-04 at 8 43 25 pm" src="https://user-images.githubusercontent.com/6797635/33590396-db0efc14-d933-11e7-95e3-3e5d56fb0bb3.png">

The left column would be the detail page view button. The right is the list view button. The smaller one would use the tab loading indicator instead of copy to maintain its scale.

For the installed state ("Manage", or the gear icon), can we open a new tab with add-ons management? Instead of only being able to provide one option ("disable"), we could drop them in the add-ons manager so the user could do more that way, and we could make a clearer connection into the browser.

andymckay:
The "Manage extension" is not currently possible, we'd have to file a bug for that. By the way, the disco pane doesn't do disable, it does uninstall.

kumar303:
> The "Manage extension" is not currently possible

Can we not link to `about:addons` with `target=_blank`?

andymckay:
No browser console error: "Security Error: Content at ... may not load or link to about:addons." Should be pretty simple to try and prove me wrong though :)

muffinresearch:
> No, the browser console error is: "Security Error: Content at ... may not load or link to about:addons." Should be pretty simple to try and prove me wrong though :)

Yep I see the same thing trying that.

muffinresearch:
@pwalm - can we get an update on an alternative handling for "Manage" e.g. could it just indicate "Installed"? We need to get something in place for this as soon as we can since Firefox will soon be linking more directly to AMO from about:addons.

pwalm:
[...]

Issue #613, opened 2016-06-17
install theme when users clicks image


designakt:
We currently show a preview of a theme as soon as a user hovers the image. Allowing users to click the image could be a second way to install an image without the need to aim for the on/off toggle.

muffinresearch:
This was fixed by https://github.com/mozilla/addons-frontend/pull/574

PR #4914, opened 2018-05-02
i 3056 removing tap to preview btn

rebmullin:
address issue #3056 

with changes - section looks like this now:
Alt text

before changes:
spring_garden_serenity_by_madonna_ _add-ons_for_firefox

@bobsilverberg / @willdurand - can you please take a look? // cc @muffinserearch


src/amo/components/Addon/index.js
```diff
@@ -154,22 +151,18 @@ export class AddonBase extends React.Component {
...
     } = this.props;
     const previewURL = addon ? addon.previewURL : null;
     const type = addon ? addon.type : ADDON_TYPE_EXTENSION;
     const iconUrl = getAddonIconUrl(addon);
 
     if (type === ADDON_TYPE_THEME) {
-      const label = isPreviewingTheme ? i18n.gettext('Cancel preview') : i18n.gettext('Tap to preview');
       const imageClassName = 'Addon-theme-header-image';
-      const headerImage = <img alt={label} className={imageClassName} src={previewURL} />;
+      const headerImage = <img alt="" className={imageClassName} src={previewURL} />;
```
willdurand: We should probably not leave this attribute empty. We could maybe add `"Preview of <addon.title>"`, what do you think?

src/amo/components/Addon/index.js
```diff
@@ -154,22 +151,18 @@ export class AddonBase extends React.Component {
     this.props.toggleThemePreview(event.currentTarget);
```
willdurand: I think we should remove this as well, no?

bobsilverberg: It looks like we can also remove the entire `componentWillUnmount` method as it looks like that's entirely about dealing with theme previews.

src/amo/components/Addon/index.js
```diff
@@ -154,22 +151,18 @@ export class AddonBase extends React.Component {
...
   }
 
-  headerImage({ compatible }) {
```
bobsilverberg: `headerImage` is still being called at line 435 with `{ compatible: isCompatible }`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -804,58 +791,6 @@ describe(__filename, () => {
       .toHaveProp('defaultInstallSource', INSTALL_SOURCE_DETAIL_PAGE);
   });
 
-  it('enables a theme preview for non-enabled add-ons', () => {
```
bobsilverberg: There still seems to be a test in this file for `it('toggles a theme on click')`, which I think can be removed. I guess it will start to fail when the other properties are removed from the component.

rebmullin: yep - there is definitely more pieces to this than i originally thought. will take another look to remove ALL related preview parts here

willdurand:
@rebmullin let us know when this PR is ready for another look (no pressure though, I am just looking at the open PRs).

rebmullin:
@willdurand  - will do - thanks : )

tests/unit/core/TestThemePreview.js
```diff
@@ -1,8 +1,6 @@
 import { createInternalAddon } from 'core/reducers/addons';
 import themeAction, { getThemeData } from 'core/themePreview';
```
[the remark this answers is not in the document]
rebmullin: hey @bobsilverberg  - from what I can tell this `getThemeData` is connected to installing and getting the theme data if it's installed. I think the only place we can remove this is from the preview pieces (which i think i've done already)

The thing I am confused about is `data-browsertheme` prop itself - I wanted to see this to be sure but I don't see where that traces back to...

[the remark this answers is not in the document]
bobsilverberg: Ok, I think we need to figure out exactly what that does. I would suggest experimenting a bit. For example, try removing that tag and see how it affects the site when run locally. 

If it turns out we do need to keep the `getThemeData` function for something other than previews, I would suggest moving it somewhere other than the core/themePreview.js file, or renaming that file as you suggest.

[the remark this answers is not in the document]
muffinresearch: I can shed some light on this. This function needs a node with a `data-browsertheme` attribute. See https://github.com/mozilla/addons-frontend/blob/85a7e0b30e206eef4a1e712afc4133186385f81b/src/core/themePreview.js#L3 

This is related to the legacy api that handles installation and previewing of lightweight themes.

Here's a snippet of code that shows how it works a bit better https://developer.mozilla.org/ru/docs/Themes-840092-dup/Lightweight_themes. It's in the Russian translation since this content has been removed from the english page a while back it would seem.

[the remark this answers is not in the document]
rebmullin: ok thanks for clarification @muffinresearch. so we'll leave this in place for now where used for install purposes

i'll rename the themePreview file..

muffinresearch:
more-red-than-green-award-sm

rebmullin:
@willdurand / @bobsilverberg  - can you please take another look at this PR when you can  - thx : )

src/core/constants.js
```diff
@@ -153,14 +153,10 @@ export const SET_VIEW_CONTEXT = 'SET_VIEW_CONTEXT';
 export const CLOSE_INFO = 'CLOSE_INFO';
 export const SHOW_INFO = 'SHOW_INFO';
 
-// Theme preview actions.
+// Theme actions.
 export const THEME_INSTALL = 'InstallBrowserTheme';
-export const THEME_PREVIEW = 'PreviewBrowserTheme';
-export const THEME_RESET_PREVIEW = 'ResetBrowserThemePreview';
 export const validThemeActions = [
   THEME_INSTALL,
```
bobsilverberg: It seems a bit odd to have an array with only a single item in it for `validThemeActions`. If there's only one possible valid action, why do we need this at all? That's a rhetorical question, as I can see from the code in `themeInstall.js` why we need this, but maybe that code can change to remove the need for this entirely. `themeInstall.js` includes a function called `themeAction` which accepts an `action` as an argument, but if there is only one valid `action` now, perhaps that function could be renamed to `installTheme` because that's the only thing it's ever going to do, and it would no longer require an `action` argument. What do you think about that, @rebmullin?

src/disco/components/Addon.js
```diff
@@ -116,43 +112,27 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { getBrowserThemeData, i18n, name, previewURL } = this.props;
+    const { i18n, name, previewURL } = this.props;
     if (this.props.type === ADDON_TYPE_THEME) {
       /* eslint-disable jsx-a11y/href-no-hash, jsx-a11y/anchor-is-valid */
```
bobsilverberg: You may not need this comment anymore either, having removed much of the following code. I'd try removing it and see if eslint complains.

src/disco/components/Addon.js
```diff
@@ -116,43 +112,27 @@ export class AddonBase extends React.Component {
...
+        <a
+          href="#"
+          className="theme-image"
         >
-          <a
-            href="#"
-            className="theme-image"
-            data-browsertheme={getBrowserThemeData()}
-            onBlur={this.resetThemePreview}
-            onClick={this.installTheme}
```
bobsilverberg: The old version of the `<a>` tag had an `onClick` property which was assigned to the `this.installTheme` function. Don't we still need that?

src/disco/components/Addon.js
```diff
@@ -281,7 +253,9 @@ export class AddonBase extends React.Component {
                 sanitizeHTMLWithExternalLinks(heading, ['a', 'span'])
               }
             />
-            {this.getDescription()}
+            {type !== ADDON_TYPE_THEME &&
```
bobsilverberg: It might be better to move this condition into the `getDescription` function, so that it could just return nothing if the add-on is a theme.

tests/unit/core/TestThemeInstall.js
```diff
@@ -1,8 +1,6 @@
 import { createInternalAddon } from 'core/reducers/addons';
-import themeAction, { getThemeData } from 'core/themePreview';
-import { THEME_PREVIEW } from 'core/constants';
+import themeAction, { getThemeData } from 'core/themeInstall';
 import { fakeTheme } from 'tests/unit/amo/helpers';
-import { createFakeEvent } from 'tests/unit/helpers';
 
 describe(__filename, () => {
   it('throws for invalid action', () => {
```
bobsilverberg: Based on my comment above, we might not need this anymore.

tests/unit/core/TestThemeInstall.js
```diff
@@ -11,25 +9,6 @@ describe(__filename, () => {
     }).toThrowError('Invalid theme action requested');
   });
 
-  it('sets-up the event for previews', () => {
```
bobsilverberg: I wonder if we need a test like this for the install action. That doesn't seem to be tested anywhere.

tests/unit/disco/components/TestAddon.js
```diff
@@ -378,71 +376,6 @@ describe(__filename, () => {
...
-      themeImage.simulate('focus', fakeEvent);
-      sinon.assert.calledWith(previewTheme, fakeEvent.currentTarget);
-    });
-
-    it('resets theme preview onBlur on theme image', () => {
-      themeImage.simulate('blur', fakeEvent);
-      sinon.assert.calledWith(resetThemePreview, fakeEvent.currentTarget);
-    });
-
-    it('calls installTheme on click', () => {
```
bobsilverberg: As mentioned above, do we not still want this to work, and should we therefore keep this test?

src/disco/components/Addon.js
rebmullin: i see what you mean but i am looking at about:addons and i don't see a click installing a theme...do you? let me know if i am missing something

the only install i see happening is if they toggle here: 
Alt text

which it looks like is implemented with the InstallButton component (below)

tests/unit/disco/components/TestAddon.js
rebmullin: as far as i can see, clicking on anything but the switch does not install the theme so i don't * think * we need this (more info on comments above about this)

bobsilverberg: What's odd about that is that this test was evidently passing before you removed any code, but it sounds like it should have been failing. @muffinresearch Can you comment on this? Are you supposed to be able to install a theme by clicking on the image from disco pane?

src/amo/components/Addon/index.js
```diff
@@ -153,58 +146,29 @@ export class AddonBase extends React.Component {
...
-      const label = isPreviewingTheme ? i18n.gettext('Cancel preview') : i18n.gettext('Tap to preview');
       const imageClassName = 'Addon-theme-header-image';
-      const headerImage = <img alt={label} className={imageClassName} src={previewURL} />;
+      const headerImage = <img alt={title} className={imageClassName} src={previewURL} />;
 
       return (
         <div
           className="Addon-theme-header"
           id="Addon-theme-header"
-          data-browsertheme={getBrowserThemeData()}
```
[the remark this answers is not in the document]
rebmullin: @bobsilverberg  - that is my understanding - that we only need it for install per @muffinresearch's comments above. 

Also I am not sure - initially I thought this might be replaced by the unification work that will be happening here #4899.

tests/unit/core/TestThemeInstall.js
rebmullin: hmm..i do see tests in InstallButton (eg: `calls installTheme when clicked`). I am not sure if this is enough. I'll take another look tho..

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
muffinresearch: You'll want to check, but IIRC clicking the image results in a install. So you will need the data there if that's the case.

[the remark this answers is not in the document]
muffinresearch: #4899 is just about unifying the preview images so that when static and lightweight themes are shown side by side they don't look ood. Once we're fully migrated to static themes it will be possible to plan removal of all lightweight theme code.

src/disco/components/Addon.js
muffinresearch: As mentioned above I think clicking the image resulted in an install too.

tests/unit/disco/components/TestAddon.js
muffinresearch: > What's odd about that is that this test was evidently passing before you removed any code, but it sounds like it should have been failing. @muffinresearch Can you comment on this? Are you supposed to be able to install a theme by clicking on the image from disco pane?

Yep as mentioned in previous comments above, clicking the image also carries out an install.

muffinresearch: Hmm having just checked it, click installs don't appear to currently work.

src/disco/components/Addon.js
muffinresearch: > As mentioned above I think clicking the image resulted in an install too.

Hmm it doesn't appear to currently work that way.

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
muffinresearch: > You'll want to check, but IIRC clicking the image results in a install. So you will need the data there if that's the case.

Click to install definitely isn't working today. But it did in the past so it looks like it regressed at some point. See also https://github.com/mozilla/addons-frontend/issues/613

muffinresearch:
I've filed a new issue to cover the "click theme image to install" regression here: https://github.com/mozilla/addons-frontend/issues/4999

src/disco/components/Addon.js
rebmullin: ok i'll add this code back in : )

rebmullin:
hey @bobsilverberg  - thanks for all the feedback. I made some updates per your comments. Can you please take another  look? hopefully it's close now :)

rebmullin:
also - fixes #4999 - click on disco image installs theme as well

rebmullin:
hey @jvillalobos - this work here is almost ready to land. Just wanted to make sure you’re ok with it shipping on May 24

tests/unit/core/TestThemeInstall.js
```diff
@@ -1,35 +1,8 @@
...
-    };
-    const fakeEvent = createFakeEvent({
-      initEvent: sinon.stub(),
-    });
-    const fakeDoc = {
-      createEvent: sinon.stub(),
-    };
-    fakeDoc.createEvent.returns(fakeEvent);
-
-    themeAction(fakeNode, THEME_PREVIEW, fakeDoc);
```
bobsilverberg: You replaced `themeAction` with `themeInstall`, so shouldn't we now have a test similar to this for `themeInstall`?

rebmullin:
@bobsilverberg  -  I appreciate your time to give me feedback and catching things that I overlooked : )

I added back that test that you suggested. Do you want to give another final review?

jvillalobos:
Is it possible to put this behind a config flag? We don't know when we'll launch static themes yet and I'd rather not drop this feature until we have some good news to balance this removal.

willdurand:
I think we should keep the "state: do not merge", I am not sure we can feature flag such a patch.

muffinresearch:
> Is it possible to put this behind a config flag? We don't know when we'll launch static themes yet and I'd rather not drop this feature until we have some good news to balance this removal.

@jvillalobos The idea with this removal was do this independently of the static themes deliberately. From the comments in the original issue I thought you were happy to do this [as long as we gave you a heads-up when it was ready.](https://github.com/mozilla/addons-frontend/issues/3056#issuecomment-326634671) 

As Will suggest above adding config for this would be difficult at this stage. If we just leave this here unmerged there's a strong chance it will bitrot.  So I'd personally prefer not to block for long on this. 

We're aiming to fix the issues for the display of static themes in the frontend inside Q2, so I think if you said we expect to be in a position to unlock uploading of static themes in Q3 I don't think that's unreasonable. 

/cc @eviljeff @rebmullin to correct me if I'm wrong with that expectation ^

jvillalobos:
> I thought you were happy to do this as long as we gave you a heads-up when it was ready.

That response was given in a context where "uninstall is quick and easy", which is not the case now. There used to be a notification bar with an Undo button that would allow you to quickly Undo installation. Also, at some point we had the switches instead of the buttons, which also made that easy. Now it requires going to the Add-ons Manager to undo the action, so I think it's a more significant loss in functionality.

I'd like this to happen around the same time we launch static themes, so the announcements can be done together and the removal of the functionality is more justified. Depending on the ongoing discussions on dynamic themes and the front-end work on static themes, this could be sooner or later.

eviljeff:
My recollection from previous discussions - most recently the Berlin work week - was that it was a deliberate decision to decouple dropping the tag-to-preview feature with launching static themes so they weren't "blamed" for breaking the feature.

I understand it's a loss in functionality... but it's a loss we were apparently okay with - I don't recall any proposal/request to re-add the switches and/or have Firefox implement the notification bar with undo for static themes, though they would be great things to have.  So is delaying the "loss" for a few more weeks worth the expense of more work/complexity on frontend? (which, I would imagine, will delay the static theme feature being finished and being able to enable submissions)

jvillalobos:
> I don't recall any proposal/request to re-add the switches and/or have Firefox implement the notification bar with undo for static themes

What I'm saying is that the previous decision to drop the preview was based on the notion that there was a notification bar that would allow an easy Undo for an installation. The bar is not there anymore, so it makes this a bigger deal.

Bringing back the toggle behavior of install buttons is https://github.com/mozilla/addons-frontend/issues/3954, which I want to get to in Q3, but that doesn't help with the timing of this issue.

Breaking theme previews for weeks (potentially months) for no other reason than "something else is coming that needs it" is something I'd rather avoid. If it helps reduce the engineering impact of this last minute (sorry) change, I think it's okay to release the changes for disco pane, given it uses the install toggles.

jvillalobos:
Given the work for dynamic themes isn't tied to this anymore and it looks like static themes are on track to being finished this quarter, I think it's okay to unblock this. No need to hide it behind a flag.

rebmullin:
I've rebased the code and this should be ready to go again.

@jvillalobos, do you need to announce anything about this anymore or is this ok to merge?


// cc @muffinresearch  @bobsilverberg

PR #4917, opened 2018-05-02
chore(package): update flow-bin to version 0.71.0

muffinresearch:
Closes #4820

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,11 +7,15 @@
     "ga": true,
     // https://flow.org/en/docs/types/utilities/#toc-propertytype
     "$PropertyType": true,
-    // This is an undocumented utility that ships with Flow
+    // These ar undocumented utilities that ship with Flow
     // https://github.com/facebook/flow/issues/875
     "$Shape": true,
+    "$Keys": true,
     // See: https://github.com/facebook/flow/issues/1609
     "SyntheticEvent": true,
+    "SyntheticInputEvent": true,
+    // See: https://github.com/facebook/flow/issues/5627
+    "TimeoutID": true,
     // Standard DOM globals:
     "Generator": true,
     "HTMLElement": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,7 @@
     "ga": true,
     // https://flow.org/en/docs/types/utilities/#toc-propertytype
     "$PropertyType": true,
-    // These ar undocumented utilities that ship with Flow
+    // These are undocumented utilities that ship with Flow
     // https://github.com/facebook/flow/issues/875
     "$Shape": true,
     "$Keys": true,
--- a/src/amo/reducers/viewContext.js
+++ b/src/amo/reducers/viewContext.js
@@ -10,12 +10,16 @@ import {
 
 
 export type ViewContextType = {|
-  context: ADDON_TYPE_EXTENSION | ADDON_TYPE_THEME | VIEW_CONTEXT_EXPLORE |
-    VIEW_CONTEXT_HOME | VIEW_CONTEXT_LANGUAGE_TOOLS,
+  context:
+    | typeof ADDON_TYPE_EXTENSION
+    | typeof ADDON_TYPE_THEME
+    | typeof VIEW_CONTEXT_EXPLORE
+    | typeof VIEW_CONTEXT_HOME
+    | typeof VIEW_CONTEXT_LANGUAGE_TOOLS,
 |}
 
 export type ViewContextActionType = {|
-  type: string,
+  type: typeof SET_VIEW_CONTEXT,
   payload: ViewContextType,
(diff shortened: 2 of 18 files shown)


willdurand:
Ok, upgrade ready. I tried not to add too much changes here.

.eslintrc
```diff
@@ -7,11 +7,15 @@
     "ga": true,
     // https://flow.org/en/docs/types/utilities/#toc-propertytype
     "$PropertyType": true,
-    // This is an undocumented utility that ships with Flow
+    // These ar undocumented utilities that ship with Flow
```
bobsilverberg: Should be "There are".

src/amo/reducers/viewContext.js
```diff
@@ -10,12 +10,16 @@ import {
 
 
 export type ViewContextType = {|
-  context: ADDON_TYPE_EXTENSION | ADDON_TYPE_THEME | VIEW_CONTEXT_EXPLORE |
-    VIEW_CONTEXT_HOME | VIEW_CONTEXT_LANGUAGE_TOOLS,
+  context:
+    | typeof ADDON_TYPE_EXTENSION
```
bobsilverberg: Why did we have to add `typeof` to each of these? I assume Flow was fine with it before.

src/core/reducers/api.js
```diff
@@ -79,13 +79,16 @@ export const initialApiState: ApiStateType = {
...
+  | SetAuthTokenAction
+  | SetLangAction
+  | SetClientAppAction
+  | SetUserAgentAction
+  | LogOutUserAction;
+
 export default function api(
   state: Exact<ApiStateType> = initialApiState,
-  action: SetAuthTokenAction
-    & SetLangAction
```
bobsilverberg: Hmm, I've not seen that before. Why were we using `&` before? Was that just a mistake that Flow didn't previously catch?

src/core/constants.js
```diff
@@ -48,6 +50,15 @@ export const ADDON_TYPE_THEME = 'persona';
 // TODO: Remove ADDON_TYPE_COMPLETE_THEME once we don't support complete
 // themes.
 export const ADDON_TYPE_COMPLETE_THEME = 'theme';
+
+export type AddonTypeType =
```
bobsilverberg: It seems a shame we have to repeat all of these (both in `AddonTypeType` and in `validAddonTypes`). I think there might be a way to create the type programmatically from the `validAddonTypes` array, rather than having them all repeated like this.

src/core/types/addons.js
```diff
@@ -1,8 +1,5 @@
 /* @flow */
-import { validAddonTypes } from 'core/constants';
-
-// You have to just laugh at this line!
-type AddonTypeType = validAddonTypes;
```
bobsilverberg: Did this work? If so, could we just move this silly line into `core/constants` to remove the duplication? Or is this one of the things the new Flow doesn't like?

.eslintrc
willdurand: Oh snap, yes you are right!

src/core/reducers/api.js
willdurand: We use `&` for intersections but the behavior has changed (got stricter). In short, yes it was a mistake not caught by Flow. It did work because there was no strict type on the redux `type` properties I think.

src/core/constants.js
willdurand: Not really yet, because `$Values<>` does not work for JS array. It is mostly due to how Facebook deals with that, i.e. they use objects to describe the actions:

```js
const Actions = {
  SOME_ACTION: 'SOME_ACTION',
  ANOTHER_ONE: 'ANOTHER_ONE',
};
```

With hash maps like the one above, one can use `$Keys<>` or `$Values<>`. Maybe we should be doing this as well..

src/amo/reducers/viewContext.js
willdurand: That does not work anymore. Flow complains that we are using a string constant and that it should be prefixed with `typeof`. Ideally, we should add types to each constant but then new errors pop up 😞

src/core/constants.js
bobsilverberg: Yeah, I think we're doing that here: https://github.com/mozilla/addons-frontend/blob/72fb82c57cb50815caf16719d30a0d585538cb74/src/ui/components/UserRating/index.js#L19. Is it worth making that change?

willdurand: That would mean changing the code, I'd be in favor of not doing this since `validAddonTypes` is used elsewhere.

bobsilverberg: Fair enough. 👍

bobsilverberg:
Thanks for taking care of this @willdurand. 👏 It's been sitting in our PR queue for some time.

PR #4870, opened 2018-04-25
WIP: Add an Edit User Profile page

tofumatt:
Fixes #3144

Adds an edit user profile page. 🎉

Needs tests.

### Screenshots
screen shot 2018-04-13 at 16 29 30

screen shot 2018-04-13 at 16 29 32

screenshot 2018-04-13 16 29 22

Issue #3144, opened 2017-09-14
Implement Account Settings page


muffinresearch:
Refs: https://github.com/mozilla/addons-frontend/issues/2781

See also https://docs.google.com/document/d/1cJ1k8HsAb6taauvJMDLOiYMa1McfUXNgEtax3hoToPc/edit#heading=h.5cheild9uc3u

This page allows the user to customize their site settings:

 - [x] Edit Username
 - [x] E-mail address (read-only), message about Firefox Accounts, link to manage Firefox Accounts, support link
 - [x] Edit metadata: display name, homepage, location, occupation
 - [ ] Upload photo, delete photo
 - [x] Edit more details: including HTML help for more details
 - [ ] Notification settings, this includes some options which are developer only.
 - [ ] ~~Reason for change, Notes, Anonymize, Delete Account (Admin-only)~~
 - [x] Update Account button

I think any admin specific functionality needs moving to the admin. This will need bugs filing on addons-server.

 Current site this looks like this:

<img width="1008" alt="account_settings____add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30439358-6c529a9a-996b-11e7-9ab5-6dbd961fba73.png">

muffinresearch:
@jvillalobos Nominating this for post-release since the account settings are a fairly large item. With the proviso that we'll put this at the top of the post release features. Is that ok with you?

jvillalobos:
As long as we keep the old pages, it's okay, since they are needed for the account creation process.

muffinresearch:
Ok great, removed.

eviljeff:
Couldn't see an implementation issue yet, so adding this to here:

The message about Firefox Accounts and changing email addresses should say instead that they can change their email address via Firefox Accounts.  https://support.mozilla.org/en-US/kb/change-primary-email-address-firefox-accounts seems like the relevant article about it.

mozilla/addons-server#4925 was logged about this on addons-server - and I'll keep that open until mozilla/addons-server#6451 is fully complete - but the messaging needs to update in the UI also.

willdurand:
@wagnerand do we want admins (or any other user with a specific right like `Users:Edit`) to be able to edit someone else's profile in the frontend? Or can we move this to the admin tool? It seems the API would allow to implement that in the frontend but I don't know whether it is a good idea or even if it is still wanted. (also cc @jvillalobos)

eviljeff:
[...]

PR #4922, opened 2018-05-03
Add an Edit User Profile page

willdurand:
Replaces #4870 - Refs #3144

---

This is the first PR for the user profile edition, allowing to change:

- username
- metadata (display name, location, bio, homepage, occupation)

## Screenshots

screen shot 2018-05-09 at 16 13 43
screen shot 2018-05-09 at 16 13 48
screen shot 2018-05-09 at 16 13 51
screen shot 2018-05-09 at 16 13 54
screen shot 2018-05-09 at 16 14 09
screen shot 2018-05-09 at 16 14 12

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -113,7 +113,7 @@ export class HeaderBase extends React.Component {
                   className="Header-user-menu-view-profile-link"
                   {...viewProfileLinkProps}
                 >
-                  {i18n.gettext('View Profile')}
+                  {i18n.gettext('View My Profile')}
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
@@ -121,7 +121,7 @@ export class HeaderBase extends React.Component {
                   className="Header-user-menu-edit-profile-link"
                   {...editProfileLinkProps}
                 >
-                  {i18n.gettext('Edit Profile')}
+                  {i18n.gettext('Edit My Profile')}
                 </Link>
               </DropdownMenuItem>
 
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -28,7 +28,7 @@ export class HeaderBase extends React.Component {
     isHomePage: PropTypes.bool,
     location: PropTypes.object.isRequired,
     query: PropTypes.string,
-    siteUser: PropTypes.object.isRequired,
+    siteUser: PropTypes.object,
     isReviewer: PropTypes.bool.isRequired,
   }
 
@@ -109,12 +109,18 @@ export class HeaderBase extends React.Component {
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
-                <Link {...viewProfileLinkProps}>
+                <Link
+                  className="Header-user-menu-view-profile-link"
+                  {...viewProfileLinkProps}
+                >
                   {i18n.gettext('View Profile')}
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
-                <Link {...editProfileLinkProps}>
+                <Link
+                  className="Header-user-menu-edit-profile-link"
+                  {...editProfileLinkProps}
+                >
                   {i18n.gettext('Edit Profile')}
                 </Link>
               </DropdownMenuItem>
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -67,6 +67,9 @@ export class HeaderBase extends React.Component {
     const viewProfileURL = siteUser ? `/user/${siteUser.username}/` : null;
     const viewProfileLinkProps = config.get('enableUserProfile') ?
       { to: viewProfileURL } : { href: viewProfileURL };
+    const editProfileURL = siteUser ? '/users/edit' : null;
+    const editProfileLinkProps = config.get('enableUserProfile') ?
+      { to: editProfileURL } : { href: editProfileURL };
 
     return (
       <header className="Header">
@@ -111,7 +114,7 @@ export class HeaderBase extends React.Component {
                 </Link>
               </DropdownMenuItem>
               <DropdownMenuItem>
-                <Link href="/users/edit">
+                <Link {...editProfileLinkProps}>
                   {i18n.gettext('Edit Profile')}
                 </Link>
               </DropdownMenuItem>
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -273,25 +273,6 @@ export class UserProfileEditBase extends React.Component<Props> {
               />
             </Card>
 
-            {/*
-            <Card
-              className="UserProfileEdit--Card"
-              header={i18n.gettext('Notification Settings')}
-            >
-              <p>
-                {i18n.gettext(`From time to time, Mozilla may send you email about
-                  upcoming releases and add-on events. Please select the topics
-                  you are interested in.`)}
-              </p>
-
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="notify">
-                  {i18n.gettext('… when an add-on author responds to my review')}
-                </label>
-                <input selece={!!user.notifyReview} value="1" type="checkbox" />
-              </div>
-            </Card>
-            */}
             <div className="UserProfileEdit-buttons-wrapper">
               <Button
                 buttonType="action"
@@ -308,15 +289,6 @@ export class UserProfileEditBase extends React.Component<Props> {
                   isEditing ? i18n.gettext("Updating user's profile…") :
                     i18n.gettext("Update user's profile")
                 )}
-              </Button>
-
-              <Button
-                buttonType="neutral"
-                className="UserProfileEdit-delete-button UserProfileEdit-button"
-                disabled={!user}
-                type="button"
-              >
-                {i18n.gettext('Delete my account')}
               </Button>
             </div>
           </div>
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -1,3 +1,4 @@
+/* @flow */
 import * as React from 'react';
 import Textarea from 'react-textarea-autosize';
 import Helmet from 'react-helmet';
@@ -13,15 +14,39 @@ import {
 } from 'amo/reducers/users';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
+import log from 'core/logger';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
-import Icon from 'ui/components/Icon';
-import LoadingText from 'ui/components/LoadingText';
+import type { UsersStateType, UserType } from 'amo/reducers/users';
+import type { DispatchFunc } from 'core/types/redux';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { I18nType } from 'core/types/i18n';
 
 import './styles.scss';
 
 
-export class UserProfileEditBase extends React.Component {
+type Props = {|
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  isEditing: boolean,
+  isEditingCurrentUser: boolean,
+  params: {| username: string |},
+  user?: UserType,
+  username: string,
+|};
+
+const getValueOrEmpty = (input) => {
+  if (input && input.value) {
+    return input.value;
+  }
+
+  return '';
+};
+
+export class UserProfileEditBase extends React.Component<Props> {
+  displayName: HTMLInputElement | null;
+
   componentWillMount() {
     const { dispatch, errorHandler, params, user } = this.props;
 
@@ -44,24 +69,35 @@ export class UserProfileEditBase extends React.Component {
     }
   }
 
-  onSubmit = (event) => {
+  onSubmit = (event: SyntheticEvent<any>) => {
     event.preventDefault();
 
     const { dispatch, errorHandler, user } = this.props;
 
+    if (!user) {
+      log.debug('Form has been submitted but no user passed as prop.');
+      return;
+    }
+
     dispatch(editUserAccount({
       errorHandlerId: errorHandler.id,
       userFields: {
-        biography: this.biography.value,
-        display_name: this.displayName.value,
-        homepage: this.homepage.value,
-        location: this.location.value,
-        occupation: this.occupation.value,
-        username: this.username.value,
+        biography: getValueOrEmpty(this.biography),
+        display_name: getValueOrEmpty(this.displayName),
+        homepage: getValueOrEmpty(this.homepage),
+        location: getValueOrEmpty(this.location),
+        occupation: getValueOrEmpty(this.occupation),
+        username: getValueOrEmpty(this.username),
       },
       userId: user.id,
     }));
   }
+
+  biography: HTMLInputElement | null;
+  homepage: HTMLInputElement | null;
+  location: HTMLInputElement | null;
+  occupation: HTMLInputElement | null;
+  username: HTMLInputElement | null;
 
   render() {
     const {
@@ -117,7 +153,7 @@ export class UserProfileEditBase extends React.Component {
               className="UserProfileEdit--Card"
               header={isEditingCurrentUser ? i18n.gettext('Your Account') : (
                 i18n.sprintf(i18n.gettext('User Account for %(username)s'), {
-                  username: username,
+                  username,
                 })
               )}
             >
@@ -290,7 +326,10 @@ export class UserProfileEditBase extends React.Component {
   }
 }
 
-export function mapStateToProps(state, ownProps) {
+export function mapStateToProps(
+  state: { users: UsersStateType },
+  ownProps: Props,
+) {
   const currentUser = getCurrentUser(state.users);
   const user = ownProps.params.username ?
     getUserByUsername(state.users, ownProps.params.username) : currentUser;
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -74,7 +74,14 @@ export class UserProfileEditBase extends React.Component {
   }
 
   render() {
-    const { errorHandler, i18n, isEditingCurrentUser, user, username } = this.props;
+    const {
+      errorHandler,
+      i18n,
+      isEditing,
+      isEditingCurrentUser,
+      user,
+      username,
+    } = this.props;
 
     return (
       <div className="UserProfileEdit">
@@ -136,27 +143,28 @@ export class UserProfileEditBase extends React.Component {
                   username and Firefox Accounts settings.`)}
               </p>
 
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="username">
-                  {i18n.gettext('Username')}
-                </label>
-                {user ? (
-                  <input
-                    id="username"
-                    name="username"
-                    ref={(ref) => { this.username = ref; }}
-                    defaultValue={user.username}
-                  />
-                ) : <LoadingText maxWidth={60} />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="username">
+                {i18n.gettext('Username')}
+              </label>
+              <input
+                className="UserProfileEdit-username"
+                disabled={!user}
+                id="username"
+                name="username"
+                ref={(ref) => { this.username = ref; }}
+                defaultValue={user && user.username}
+              />
 
               <div title={i18n.gettext('Email address cannot be changed')}>
                 <label className="UserProfileEdit--label" htmlFor="email">
                   {i18n.gettext('Email address')}
                 </label>
-                {user ? (
-                  <input disabled defaultValue={user.email} type="email" />
-                ) : <LoadingText maxWidth={60} />}
+                <input
+                  className="UserProfileEdit-email"
+                  disabled
+                  defaultValue={user && user.email}
+                  type="email"
+                />
               </div>
             </Card>
 
@@ -170,18 +178,17 @@ export class UserProfileEditBase extends React.Component {
                   know you better.`)}
               </p>
 
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="displayName">
-                  {i18n.gettext('Display Name')}
-                </label>
-                {user ? (
-                  <input
-                    name="displayName"
-                    ref={(ref) => { this.displayName = ref; }}
-                    defaultValue={user.displayName}
-                  />
-                ) : <LoadingText maxWidth={60} />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="displayName">
+                {i18n.gettext('Display Name')}
+              </label>
+              <input
+                className="UserProfileEdit-displayName"
+                disabled={!user}
+                id="displayName"
+                name="displayName"
+                ref={(ref) => { this.displayName = ref; }}
+                defaultValue={user && user.displayName}
+              />
 
               {/*
                 TODO: Don't show these to users who don't have a public-facing
@@ -190,67 +197,60 @@ export class UserProfileEditBase extends React.Component {
                 no one will see. It also gets in the way of settings,
                 like notifications, below.
               */}
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="homepage">
-                  {i18n.gettext('Homepage')}
-                </label>
-                {user ? (
-                  <input
-                    id="homepage"
-                    name="homepage"
-                    ref={(ref) => { this.homepage = ref; }}
-                    defaultValue={user.homepage}
-                  />
-                ) : <LoadingText maxWidth={60} />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="homepage">
+                {i18n.gettext('Homepage')}
+              </label>
+              <input
+                className="UserProfileEdit-homepage"
+                disabled={!user}
+                id="homepage"
+                name="homepage"
+                ref={(ref) => { this.homepage = ref; }}
+                defaultValue={user && user.homepage}
+              />
 
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="location">
-                  {i18n.gettext('Location')}
-                </label>
-                {user ? (
-                  <input
-                    name="location"
-                    ref={(ref) => { this.location = ref; }}
-                    defaultValue={user.location}
-                  />
-                ) : <LoadingText maxWidth={60} />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="location">
+                {i18n.gettext('Location')}
+              </label>
+              <input
+                className="UserProfileEdit-location"
+                disabled={!user}
+                id="location"
+                name="location"
+                ref={(ref) => { this.location = ref; }}
+                defaultValue={user && user.location}
+              />
 
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="occupation">
-                  {i18n.gettext('Occupation')}
-                </label>
-                {user ? (
-                  <input
-                    className="UserProfileEdit-occupation"
-                    name="occupation"
-                    ref={(ref) => { this.occupation = ref; }}
-                    defaultValue={user.occupation}
-                  />
-                ) : <LoadingText maxWidth={60} />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="occupation">
+                {i18n.gettext('Occupation')}
+              </label>
+              <input
+                className="UserProfileEdit-occupation"
+                disabled={!user}
+                id="occupation"
+                name="occupation"
+                ref={(ref) => { this.occupation = ref; }}
+                defaultValue={user && user.occupation}
+              />
             </Card>
 
             <Card
               className="UserProfileEdit--Card"
               header={i18n.gettext('Your Biography')}
             >
-              <div>
-                <label className="UserProfileEdit--label" htmlFor="biography">
-                  {i18n.gettext(`Introduce yourself to the community. This text
-                    will appear on your user profile page.
-                    (Some HTML supported)`)}
-                </label>
-                {user ? (
-                  <Textarea
-                    className="UserProfileEdit-biography"
-                    name="biography"
-                    ref={(ref) => { this.biography = ref; }}
-                    defaultValue={user.biography}
-                  />
-                ) : <Textarea className="UserProfileEdit-biography" disabled />}
-              </div>
+              <label className="UserProfileEdit--label" htmlFor="biography">
+                {i18n.gettext(`Introduce yourself to the community. This text
+                  will appear on your user profile page.
+                  (Some HTML supported)`)}
+              </label>
+              <Textarea
+                className="UserProfileEdit-biography"
+                disabled={!user}
+                id="biography"
+                name="biography"
+                ref={(ref) => { this.biography = ref; }}
+                defaultValue={user && user.biography}
+              />
             </Card>
 
             {/*
@@ -276,11 +276,18 @@ export class UserProfileEditBase extends React.Component {
               <Button
                 buttonType="action"
                 className="UserProfileEdit-submit-button UserProfileEdit-button"
-                disabled={!user}
+                disabled={!user || isEditing}
                 puffy
                 type="submit"
               >
-                {i18n.gettext('Update my profile')}
+                {/* eslint-disable-next-line no-nested-ternary */}
+                {isEditingCurrentUser ? (
+                  isEditing ? i18n.gettext('Updating your profile…') :
+                    i18n.gettext('Update my profile')
+                ) : (
+                  isEditing ? i18n.gettext("Updating user's profile…") :
+                    i18n.gettext("Update user's profile")
+                )}
               </Button>
 
               <Button
@@ -301,12 +308,12 @@ export class UserProfileEditBase extends React.Component {
 
 export function mapStateToProps(state, ownProps) {
   const currentUser = getCurrentUser(state.users);
-  // const user = null;
   const user = ownProps.params.username ?
     getUserByUsername(state.users, ownProps.params.username) : currentUser;
 
   return {
     currentUser,
+    isEditing: state.users.isEditing,
     isEditingCurrentUser: currentUser && user && currentUser.id === user.id,
     user,
     username: user ? user.username : ownProps.params.username,
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -21,6 +21,7 @@ import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
+import Notice from 'ui/components/Notice';
 import type { UsersStateType, UserType } from 'amo/reducers/users';
 import type { DispatchFunc } from 'core/types/redux';
 import type { ErrorHandlerType } from 'core/errorHandler';
@@ -41,6 +42,10 @@ type Props = {|
   username: string,
 |};
 
+type State = {|
+  displaySuccessMessage: boolean,
+|};
+
 const getValueOrEmpty = (input) => {
   if (input && input.value) {
     return input.value;
@@ -49,8 +54,16 @@ const getValueOrEmpty = (input) => {
   return '';
 };
 
-export class UserProfileEditBase extends React.Component<Props> {
+export class UserProfileEditBase extends React.Component<Props, State> {
   displayName: HTMLInputElement | null;
+
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      displaySuccessMessage: false,
+    };
+  }
 
   componentWillMount() {
     const { dispatch, errorHandler, params, user } = this.props;
@@ -63,14 +76,27 @@ export class UserProfileEditBase extends React.Component<Props> {
     }
   }
 
-  componentWillReceiveProps({ params: newParams }: Props) {
-    const { dispatch, errorHandler, params: oldParams } = this.props;
+  componentWillReceiveProps(props: Props) {
+    const {
+      isEditing: wasEditing,
+      params: newParams,
+    } = props;
+    const {
+      dispatch,
+      errorHandler,
+      isEditing,
+      params: oldParams,
+    } = this.props;
 
     if (oldParams.username !== newParams.username) {
       dispatch(fetchUserAccount({
         errorHandlerId: errorHandler.id,
         username: newParams.username,
       }));
+    }
+
+    if (wasEditing && !isEditing && !errorHandler.hasError()) {
+      this.setState({ displaySuccessMessage: true });
     }
   }
 
@@ -151,14 +177,17 @@ export class UserProfileEditBase extends React.Component<Props> {
           </ul>
         </Card>
 
-        <form
-          action=""
-          className="UserProfileEdit-form"
-          onSubmit={this.onSubmit}
-        >
-          <div>
+        <form className="UserProfileEdit-form" onSubmit={this.onSubmit}>
+          <div className="UserProfileEdit-form-messages">
             {errorHandler.renderErrorIfPresent()}
 
+            {this.state.displaySuccessMessage && (
+              <Notice type="success">
+                {i18n.gettext('Profile successfully updated.')}
+              </Notice>
+            )}
+          </div>
+          <div>
             <Card
               className="UserProfileEdit--Card"
               header={isEditingCurrentUser ? i18n.gettext('Account') : (
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import invariant from 'invariant';
 import * as React from 'react';
 import Textarea from 'react-textarea-autosize';
 import Helmet from 'react-helmet';
@@ -17,7 +18,6 @@ import {
 import { USERS_EDIT } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
-import log from 'core/logger';
 import { sanitizeHTML } from 'core/utils';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
@@ -36,7 +36,6 @@ type Props = {|
   hasEditPermission: boolean,
   i18n: I18nType,
   isEditing: boolean,
-  isEditingCurrentUser: boolean,
   params: {| username: string |},
   user: UserType | null,
   username: string,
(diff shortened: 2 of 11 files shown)


willdurand:
Ok so this should be ready for a first review, thanks!

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+                className="UserProfileEdit-homepage"
+                disabled={!user}
+                id="homepage"
+                name="homepage"
+                ref={(ref) => { this.homepage = ref; }}
+                defaultValue={user && user.homepage}
+                type="url"
+              />
+              <p className="UserProfileEdit--help">
+                {i18n.gettext(`This URL will only be visible if you are a
```
bobsilverberg: "if you are a developer" || "if the user is a developer"

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+                defaultValue={user && user.occupation}
+              />
+            </Card>
+
+            <Card
+              className="UserProfileEdit--Card"
+              header={i18n.gettext('Your Biography')}
+            >
+              <label className="UserProfileEdit--label" htmlFor="biography">
+                {i18n.gettext(`Introduce yourself to the community. This text
```
bobsilverberg: "yourself" & "your"

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+                ref={(ref) => { this.biography = ref; }}
+                defaultValue={user && user.biography}
+              />
+            </Card>
+
+            <div className="UserProfileEdit-buttons-wrapper">
+              <Button
+                buttonType="action"
+                className="UserProfileEdit-submit-button UserProfileEdit-button"
+                disabled={!user || isEditing}
```
bobsilverberg: As we've done with the edit collection form, perhaps we should set the button to disabled if the form is invalid. Specifically, if username is empty.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+
+type Props = {|
+  currentUser: UserType | null,
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  hasEditPermission: boolean,
+  i18n: I18nType,
+  isEditing: boolean,
+  isEditingCurrentUser: boolean,
+  params: {| username: string |},
```
bobsilverberg: I find this `params` prop a bit odd. Why does this need to take an object that only has one property? I understand that we have another prop for `username`, but maybe this could be `requestedUsername`, or something like that? I don't see the utility in having an object.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+
+              <label className="UserProfileEdit--label" htmlFor="username">
+                {i18n.gettext('Username')}
+              </label>
+              <input
+                className="UserProfileEdit-username"
+                disabled={!user}
+                id="username"
+                name="username"
+                ref={(ref) => { this.username = ref; }}
```
bobsilverberg: This is more of a question than a comment. I see the way we're managing this form's state is different from what we've done with the collection edit form. We're using `ref` and `defaultValue`, whereas in the [collection form](https://github.com/mozilla/addons-frontend/blob/72fb82c57cb50815caf16719d30a0d585538cb74/src/amo/components/CollectionManager/index.js#L145-L151) we use an internal state and an `onChange` function to manage it.

What are the pros and cons between the two approaches?

src/amo/routes.js
```diff
@@ -42,7 +43,13 @@ export default (
       <IndexRoute component={Home} />
       <Route path="addon/:slug/" component={Addon} />
       <Route path="addon/:addonSlug/reviews/" component={AddonReviewList} />
-      <Route path="user/:username/" component={UserProfile} />
+      {config.get('enableUserProfile') && (
+        <React.Fragment>
+          <Route path="users/edit" component={UserProfileEdit} />
+          <Route path="user/:username/edit" component={UserProfileEdit} />
+          <Route path="user/:username/" component={UserProfile} />
```
bobsilverberg: This was not previously wrapped in an `if` based on `enableUserProfile`. Should we keep it outside of the `if`?

tests/unit/amo/components/TestHeader.js
```diff
@@ -100,6 +100,28 @@ describe(__filename, () => {
     expect(link.prop('children')).toEqual('View My Collections');
   });
 
+  it('displays a view profile link when user is signed in', () => {
```
bobsilverberg: Should we include tests to show that it does not display the links when there is no signed in user?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+    params = { username: 'tofumatt' },
+    store = dispatchSignInActions({
+      userProps: defaultUserProps(),
+    }).store,
+    ...props
+  } = {}) {
+    return shallowUntilTarget(
+      <UserProfileEdit i18n={i18n} params={params} store={store} {...props} />,
+      UserProfileEditBase,
+      { shallowOptions: { context: { i18n } } }
```
bobsilverberg: What does this do? I haven't seen it in most of the tests.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+    root.setProps({ params });
+
+    sinon.assert.notCalled(dispatchSpy);
+  });
+
+  it('renders a username input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
+
+    expect(root.find('.UserProfileEdit-username')).toHaveLength(1);
```
bobsilverberg: Would it be good to check the value of this field as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a disabled username input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
```
bobsilverberg: This is a confusing test. Is the form "not ready" because the user we are requesting is not the current logged in user? How do we know the current logged in user has access to edit another user's profile?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  function defaultUserProps(props = {}) {
+    return {
+      display_name: 'Matt MacTofu',
+      userId: 500,
+      username: 'tofumatt',
+      ...props,
+    };
+  }
+
+  function syncPropsAndParamsUsername(username) {
```
bobsilverberg: I find the name of this function confusing, and I think the whole thing is extra confusing because of the `params` object which I suggested changing above. This function is really doing two things (iiuc):

- set the `username` prop to what I passed in
- log in with the username that I passed in

If that's the case then perhaps it should be given a name that suggests this.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+    const root = renderUserProfileEdit({ params: { username }, store });
+
+    sinon.assert.calledWith(dispatchSpy, fetchUserAccount({
+      errorHandlerId: root.instance().props.errorHandler.id,
+      username,
+    }));
+  });
+
+  it('does not dispatch fetchUserAccount action if there is no username', () => {
+    const { store } = syncPropsAndParamsUsername('tofumatt');
```
bobsilverberg: Because we don't care about the actual username for this test, just that no username is passed into render, why not make this state the default state and then we can just call `renderUserProfileEdit` below without doing this first and without passing in a `store`?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a disabled username input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
+
+    expect(root.find('.UserProfileEdit-username')).toHaveProp('disabled', true);
+  });
+
+  it('renders a disabled "email" input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: This is the same as the default for `renderUserProfileEdit`, and as we're not checking anything about the actual `username`, this could just be `const root = renderUserProfileEdit();`.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+    expect(root.find('.UserProfileEdit-email')).toHaveLength(1);
+    expect(root.find('.UserProfileEdit-email')).toHaveProp('disabled', true);
+  });
+
+  it('renders a help text for the email and homepage fields', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: As above, this can just be `const root = renderUserProfileEdit();`.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+    expect(root.find('.UserProfileEdit--help')).toHaveLength(2);
```
bobsilverberg: Might it be worth checking the contents of the fields too?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a displayName input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: Should we use `userProps` to set a specific `displayName` for the user and then check for that value in the test as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+  it('renders a disabled displayName input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
+
+    expect(root.find('.UserProfileEdit-displayName'))
+      .toHaveProp('disabled', true);
+  });
+
+  it('renders a homepage input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: Should we use `userProps` to set a specific `homepage` for the user and then check for that value in the test as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a disabled homepage input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
+
+    expect(root.find('.UserProfileEdit-homepage')).toHaveProp('disabled', true);
+  });
+
+  it('renders a location input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: Should we use `userProps` to set a specific `location` for the user and then check for that value in the test as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a disabled location input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
+
+    expect(root.find('.UserProfileEdit-location')).toHaveProp('disabled', true);
+  });
+
+  it('renders a occupation input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: Should we use `userProps` to set a specific `occupation` for the user and then check for that value in the test as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+  });
+
+  it('renders a disabled occupation input field when not ready', () => {
+    const root = renderUserProfileEdit({ params: { username: 'not-ready' } });
+
+    expect(root.find('.UserProfileEdit-occupation')).toHaveProp('disabled', true);
+  });
+
+  it('renders a biography input field', () => {
+    const root = renderUserProfileEdit({ params: { username: 'tofumatt' } });
```
bobsilverberg: Should we use `userProps` to set a specific `biography` for the user and then check for that value in the test as well?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+    expect(root.find('.UserProfileEdit-biography')).toHaveLength(1);
+  });
+
+  it('renders a disabled biography input field when not ready', () => {
```
bobsilverberg: Is it really valuable to have all of these "it renders a disabled..." as separate tests? I think it might be acceptable to just combine them into a single test which checks all of the fields for disabled.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+
+    expect(root.find('.UserProfileEdit-biography')).toHaveProp('disabled', true);
+  });
+
+  it('dispatches editUserAccount action with all fields', () => {
+    const { store } = syncPropsAndParamsUsername('tofumatt');
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const errorHandler = createStubErrorHandler();
+
+    const root = mountUserProfileEdit({ errorHandler, store });
```
bobsilverberg: Why do we need to use `mount` for this test? Is it because of the approach that was taken that binds the form fields to the component's props? I don't think we had to use `mount` in the tests for the edit collection form.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+    const user = getCurrentUser(store.getState().users);
+
+    root.find('.UserProfileEdit-form').simulate('submit', createFakeEvent());
+
+    sinon.assert.calledWith(dispatchSpy, editUserAccount({
```
bobsilverberg: Is this one test with the default values good enough, or should we have (at least) another which sets actual values that are not the defaults and verifies that dispatch is called with the correct values?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,329 @@
...
+    const { store } = dispatchSignInActions();
+    const errorHandler = new ErrorHandler({
+      id: 'some-id',
+      dispatch: store.dispatch,
+    });
+    errorHandler.handle(new Error('unexpected error'));
+
+    const root = renderUserProfileEdit({ errorHandler, store });
+
+    expect(root.find(ErrorList)).toHaveLength(1);
```
bobsilverberg: Should we verify the content of the `ErrorList` as well?

bobsilverberg:
@willdurand There were a few things I noticed as well when running the code locally:

1. After saving there is no confirmation message, which is a bit confusing.
 Do you think we should display a confirmation message?

2. There doesn't appear to be a link to the Edit Profile page from the View Profile page. Should there be one?

3. When the edit page is made compact, the link to get to View My Profile disappears. Should there still be one somewhere?

src/amo/components/Header/index.js
```diff
@@ -67,6 +67,9 @@ export class HeaderBase extends React.Component {
     const viewProfileURL = siteUser ? `/user/${siteUser.username}/` : null;
     const viewProfileLinkProps = config.get('enableUserProfile') ?
       { to: viewProfileURL } : { href: viewProfileURL };
+    const editProfileURL = siteUser ? '/users/edit' : null;
+    const editProfileLinkProps = config.get('enableUserProfile') ?
+      { to: editProfileURL } : { href: editProfileURL };
```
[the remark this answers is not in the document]
willdurand: If I remember correctly, `to` is intercepted by the react-router and always to navigate without page reload. `href` is used for plain old HTML links, which we have to use since the edit/view pages are currently served by the old frontend.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  onSubmit = (event: SyntheticEvent<any>) => {
+    event.preventDefault();
+
+    const { dispatch, errorHandler, user } = this.props;
+
+    // This should never happen in real life, but Flow complains without this
+    // check...
+    if (!user) {
```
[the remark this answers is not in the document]
willdurand: ah right, I forgot about `invariant` already.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+const getValueOrEmpty = (input) => {
+  if (input && input.value) {
+    return input.value;
+  }
+
+  return '';
+};
+
+export class UserProfileEditBase extends React.Component<Props> {
+  displayName: HTMLInputElement | null;
```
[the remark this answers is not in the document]
willdurand: I think that's because it is used before. I'll try to move it but if I remember well, eslint was complaining...

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+}
+
+export function mapStateToProps(
+  state: { users: UsersStateType },
+  ownProps: Props,
+) {
+  const currentUser = getCurrentUser(state.users);
+  const user = ownProps.params.username ?
+    getUserByUsername(state.users, ownProps.params.username) : currentUser;
+  const isEditingCurrentUser = currentUser && user ? currentUser.id ===
```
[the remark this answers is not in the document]
willdurand: I think so yes.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+          </Helmet>
+        )}
+
+        <Card className="UserProfileEdit-user-links">
+          <ul>
+            <li>
+              <Link to={`/user/${username}/`}>
+                {isEditingCurrentUser ?
+                  i18n.gettext('View My Profile') :
+                  i18n.gettext("View user's profile")
```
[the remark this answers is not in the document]
willdurand: ah well, why not :)

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+            <Card
+              className="UserProfileEdit--Card"
+              header={isEditingCurrentUser ? i18n.gettext('Your Account') : (
+                i18n.sprintf(i18n.gettext('User Account for %(username)s'), {
+                  username,
+                })
+              )}
+            >
+              <p className="UserProfileEdit-aside">
+                {i18n.gettext(`Manage basic account information, such as your
```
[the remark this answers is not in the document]
willdurand: That's the content in the mock-ups. I do agree with the alternative when editing another user.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,364 @@
...
+              <input
+                className="UserProfileEdit-username"
+                disabled={!user}
+                id="username"
+                name="username"
+                ref={(ref) => { this.username = ref; }}
+                defaultValue={user && user.username}
+              />
+
+              <div title={i18n.gettext('Email address cannot be changed here')}>
```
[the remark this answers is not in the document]
willdurand: I don't know, only admins or users with a powerful right can do that.

src/amo/components/UserProfileEdit/index.js
willdurand: This prop `params` comes from the react-router.

src/amo/components/UserProfileEdit/index.js
willdurand: Right. It seems like a lot of content/switches for admins though. Maybe there is a better way to do that, or maybe not..

src/amo/routes.js
willdurand: Yes, although we do not have a nginx rule for that in -prod or -stage so it is almost the same result.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Indeed!

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: We tend to write a test case for each intent.

willdurand: We could maybe merge some of them though.

src/amo/components/UserProfileEdit/index.js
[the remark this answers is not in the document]
willdurand: eslint complains if I move this line somewhere else and also when I try to put the other field definitions here..

src/amo/components/UserProfileEdit/index.js
willdurand: I'll try to avoid pronouns as per https://design.firefox.com/photon/copy/writing-for-users.html#point-of-view. That should result in less switches.

src/amo/routes.js
willdurand: I actually believe this was a mistake, isnt it? I believe the flag should control the whole user profile feature.

src/amo/components/UserProfileEdit/index.js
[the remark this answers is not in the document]
bobsilverberg: Too bad. I guess you could add an eslint ignore comment to move it together with the others, or just leave it as is.

src/amo/components/UserProfileEdit/index.js
bobsilverberg: Maybe we should ask @devaneymoz and/or @MeridelW for a review of this copy?

src/amo/components/UserProfileEdit/index.js
bobsilverberg: Ok, at least I understand why it's here now. :-)

src/amo/components/UserProfileEdit/index.js
bobsilverberg: How consistent is the "uncontrolled" approach with the other forms we already have? I'd prefer  one way of doing things over two for consistency across the codebase.

src/amo/routes.js
bobsilverberg: I thought we already had the new User Profile view available on prod, but no, we don't, so I think you're correct that the original was a mistake and it all can be wrapped in the `if`.

src/amo/components/UserProfileEdit/index.js
willdurand: Indeed, and this is a good argument in favor of controlled inputs!

tests/unit/amo/components/TestHeader.js
willdurand: When user is not signed in, I think it should return a 404 and that's what I have done in the new iteration (not pushed yet).

bobsilverberg:
> I don't think I have ever done a "creation" page on addons-frontend, so I don't really what we expect. I would also think that a confirmation message would be nice but do we have that somewhere else already?

I know we have one after an add-on is added to a collection, so that's one place to look. We don't do a confirmation after creating or editing a collection because we return them to the details screen, and that seems like confirmation enough. I suppose you could just return users to the View Profile screen when they click "save" from the Edit Profile screen, and that would be an indication that their update took place.

tests/unit/amo/components/TestHeader.js
bobsilverberg: Maybe an [`AuthenticateButton`](https://github.com/mozilla/addons-frontend/blob/master/src/core/components/AuthenticateButton/index.js) would be even better than a 404.

src/amo/components/UserProfileEdit/index.js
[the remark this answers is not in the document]
willdurand: Don't worry, I'll get rid of these lines anyway because I now use controlled inputs.

tests/unit/amo/components/TestHeader.js
bobsilverberg: Also, this is about the menu, so the 404 or AuthenticateButton aren't really relevant to this - it's just about showing a link or not, right?

willdurand: The legacy frontend redirects the user to FxA so a login button might make sense, but the page will be very much empty.

willdurand: Well the page should not be displayed at all if there is no current logged in user IMO, because we do not really want to show a form in this case.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Waaay better indeed!

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: `mount` is gone now :)

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Do we do that in other test files? I think it is enough because the `ErrorList` component has its own test suite too.

willdurand:
Sooo I think it is ready for another look. I have fixed lots of stuff, and I have open new issues for the rest because this PR is huuuuge! I am sincerely sorry for that @bobsilverberg 😢 I wish you good luck, and thanks for reviewing!

tests/unit/amo/components/TestUserProfileEdit.js
bobsilverberg: Good point, this would be covered by the tests for `ErrorList`.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+import './styles.scss';
+
+
+type Props = {|
+  currentUser: UserType | null,
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  hasEditPermission: boolean,
+  i18n: I18nType,
+  isEditing: boolean,
```
bobsilverberg: I find the name of this prop confusing. To me `isEditing` sounds like a user is looking at the edit screen and making changes, but really this means the user has submitted changes via the edit screen, correct? Maybe `isSaving` would be better?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,554 @@
...
+    const root = renderUserProfileEdit({ store });
+
+    expect(root.find(Notice)).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-submit-button'))
+      .toHaveProp('disabled', true);
+
+    // The user profile has been updated.
+    store.dispatch(finishEditUserAccount());
+
+    const { isEditing } = store.getState().users;
```
bobsilverberg: Why do you need to do this? Can you not just manually set `isEditing` to `false`? That's what you always expect the value to be at this point in the test, correct?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,554 @@
...
+    root.setProps({ isEditing });
+
+    expect(root.find(Notice)).toHaveLength(1);
```
bobsilverberg: Is it worth checking the prop of the `Notice` to see that it has the expected message and also perhaps that it has the expected `type`?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,554 @@
...
+    const root = renderUserProfileEdit({ params, store });
+
+    expect(root.find(NotFound)).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-username'))
+      .toHaveProp('value', user.username);
+
+    const linkItems = root.find('.UserProfileEdit-user-links li');
+    expect(linkItems.at(0).find(Link))
+      .toHaveProp('children', "View user's profile");
+    // We are on the edit page, so we do not use a link for this item.
```
bobsilverberg: I don't understand this comment.

willdurand:
> One thing I think I mentioned in an overall comment last time, and which wasn't addressed, is that I think we should have a link to "Edit my profile" right on the View Profile screen. It seems inconvenient for a user to have to open the menu and choose an item (which they might not even know exists) in order to edit their profile. We might even want to include an edit link for all profile pages if the current user is allowed to edit another user, but I think that's secondary. What do you think?

I think it is correct, but I don't think the mock-ups show an edit link. Also, this page is not finished so I guess we can add a link to the view page later. What do you think?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+    } = props;
+
+    if (oldUsername !== newUsername) {
+      dispatch(fetchUserAccount({
+        errorHandlerId: errorHandler.id,
+        username: newUsername,
+      }));
+
+      this.setState({
+        ...this.getFormValues(newUser),
```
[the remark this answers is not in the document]
willdurand: I will change with `null` and add a comment. We need to reset the component state here.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+      biography,
+      display_name: displayName,
+      homepage,
+      location,
+      occupation,
+      username,
+    } = user;
+
+    return {
+      ...defaultFormValues,
```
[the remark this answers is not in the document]
willdurand: ahm, I don't think it is useful actually. I put this here to always have default values, but all these values should be overridden by the user data.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+          </ul>
+        </Card>
+
+        <form className="UserProfileEdit-form" onSubmit={this.onSubmit}>
+          <div className="UserProfileEdit-form-messages">
+            {errorHandler.renderErrorIfPresent()}
+
+            {this.state.displaySuccessMessage && (
+              <Notice type="success">
+                {i18n.gettext('Profile successfully updated.')}
```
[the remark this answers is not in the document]
willdurand: I see, can we keep this as is for now and iterate? It is the current behavior of the legacy frontend in -prod.

I'd be happy to open an issue and ping UX to know what we should be doing.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+                onChange={this.onFieldChange}
+                value={this.state.displayName}
+              />
+
+              {/*
+                TODO: Don't show these to users who don't have a public-facing
+                user profile page (eg are developers). It's just noise and may
+                encourage them to enter a lot of text (especially the bio) which
+                no one will see. It also gets in the way of settings,
+                like notifications, below.
```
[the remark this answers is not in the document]
willdurand: oh right!

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,554 @@
...
+    root.setProps({ params });
+
+    sinon.assert.notCalled(dispatchSpy);
+  });
+
+  it('renders a username input field', () => {
+    const root = renderUserProfileEdit();
+    expect(root.find('.UserProfileEdit-username')).toHaveLength(1);
+    expect(root.find('.UserProfileEdit-username'))
+      .toHaveProp('value', defaultUserProps.username);
```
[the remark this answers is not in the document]
willdurand: No you're 200% right, we should be doing this and I am daft. I knew this but wanted to write more test cases and tried to save time here...

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,554 @@
...
+      'displayName',
+      'homepage',
+      'location',
+      'occupation',
+      'username',
+    ];
+
+    const root = renderUserProfileEdit();
+
+    expect.assertions(fields.length);
```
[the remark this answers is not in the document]
willdurand: Yes! Because there is a loop below, I let Jest know that I expect `N` assertions, otherwise if something breaks (without an error, e.g., `break`) we will not notice anything.

src/amo/components/UserProfileEdit/index.js
willdurand: It would be indeed better but I did not change this variable name because it is bound to a property in the reducer, and I did not want to change that property. That would be too much changes. I decided not to just change this name here to keep the 1:1 relation.

Can we do this in another PR? I can open a new issue for that.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Yes, but we tend to read the values from the state to avoid hardcoded values.

willdurand:
Ok, hopefully good for a last review :)

I would like to defer to new issues as much non-blocking improvements as possible. The diff is soo huge 😞

src/amo/components/UserProfileEdit/index.js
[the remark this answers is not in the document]
bobsilverberg: Yes, I think an issue should be opened for it. It doesn't have to block landing this PR, but I do think the current flow is sub-optimal.

tests/unit/amo/components/TestUserProfileEdit.js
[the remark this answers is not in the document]
bobsilverberg: Well, like I said, it would be my inclination to write it this way as well, so you're calling both of us daft. ;-)  But, yeah, we sometimes need to write more code than we'd like to. :-)

tests/unit/amo/components/TestUserProfileEdit.js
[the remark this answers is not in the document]
bobsilverberg: Nice!

src/amo/components/UserProfileEdit/index.js
bobsilverberg: Sure, that's fine. Please open an issue for that so the next person to review the code isn't also confused by it. ;-)

bobsilverberg:
> I think it is correct, but I don't think the mock-ups show an edit link. Also, this page is not finished so I guess we can add a link to the view page later. What do you think?

Sure, let's open an issue for that as well and ask for UX input.

src/amo/components/UserProfileEdit/index.js
bobsilverberg: Maybe it would be worth adding a TODO comment here as well, with a link to the issue.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+  username: string,
+|};
+
+type FormValues = {|
+  biography: string | null,
+  displayName: string | null,
+  homepage: string | null,
+  location: string | null,
+  occupation: string | null,
+  username: string | null,
```
bobsilverberg: Can `username` be null? Don't we always have at least a `username`?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+  componentWillReceiveProps(props: Props) {
+    const { isEditing: wasEditing, username: oldUsername } = this.props;
+    const {
+      dispatch,
+      errorHandler,
+      isEditing,
+      username: newUsername,
+    } = props;
+
+    if (oldUsername !== newUsername) {
```
bobsilverberg: I know you changed this because I felt the previous code was confusing, but I think there's a problem here now. It won't properly update the state if the username does change. You can see this if you run the code and update your username, and click "Update", you get a screen with the old username populated and all the other fields empty.

I think it's missing the ability to handle receiving the new user data after `fetchUserAccount` returns the data.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -0,0 +1,475 @@
...
+          </ul>
+        </Card>
+
+        <form className="UserProfileEdit-form" onSubmit={this.onSubmit}>
+          <div className="UserProfileEdit-form-messages">
+            {errorHandler.renderErrorIfPresent()}
+
+            {this.state.displaySuccessMessage && (
+              <Notice type="success">
+                {i18n.gettext('Profile successfully updated.')}
```
bobsilverberg: I think we tend to leave out the period at the end of confirmation messages. For example, in [AddAddonToCollection](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddAddonToCollection/index.js#L233).

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+    return dispatchSignInActions({
+      userProps: { ...defaultUserProps, username },
+    });
+  }
+
+  function renderUserProfileEdit({
+    i18n = fakeI18n(),
+    params = { username: 'tofumatt' },
+    userProps = { ...defaultUserProps },
+    store = null,
```
bobsilverberg: Why do you need this? Without it won't the test below for `!store` still work, as `store` will be `undefined`? If you could remove this you could also remove the eslint comment below.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+    errorHandlerId = createStubErrorHandler().id,
+  }) {
+    store.dispatch(editUserAccount({
+      errorHandlerId,
+      userFields,
+      userId,
+    }));
+  }
+
+  it('renders user profile page for current logged user', () => {
```
bobsilverberg: Nit: "current logged-in user" (or "logged in user"), not sure which is correct.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+      },
+    });
+
+    expect(root.find('.UserProfileEdit-username')).toHaveLength(1);
+    expect(root.find('.UserProfileEdit-username'))
+      .toHaveProp('value', username);
+  });
+
+  it('renders disabled input fields when user to edit is not loaded', () => {
+    const { store } = signInUserWithUsername('current-logged-user');
```
bobsilverberg: Nit: `current-logged-in-user`.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+    const username = 'some username';
+    const root = renderUserProfileEdit({
+      params: { username },
+      userProps: {
+        ...defaultUserProps,
+        username,
+      },
+    });
+
+    expect(root.find('.UserProfileEdit-username')).toHaveLength(1);
```
bobsilverberg: I think we've discussed this before, perhaps with you saying the opposite on one of my PRs, but is it really necessary to have the `toHaveLength(1)` test here when we're also doing a `toHaveProp('value', username)` check immediately below? It just seems redundant.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+  it('renders a submit button', () => {
+    const root = renderUserProfileEdit();
+    const button = root.find('.UserProfileEdit-submit-button');
+
+    expect(button).toHaveLength(1);
+    expect(button.dive()).toHaveText('Update my profile');
+    expect(button).toHaveProp('disabled', false);
+  });
+
+  it('renders a submit button with a different text when user is not the logged user', () => {
```
bobsilverberg: Nit: "not the logged-in user".

tests/unit/amo/components/TestUserProfileEdit.js
bobsilverberg: As a follow-up to this, are we happy to just be testing with an updated `location`, or should we test with all of the editable fields, either by setting all of them and submitting, or each one and then submitting individually in a loop?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+    const { isEditing } = store.getState().users;
+    root.setProps({ isEditing });
+
+    expect(root.find(Notice)).toHaveLength(0);
+
+    expect(root.find('.UserProfileEdit-submit-button'))
+      .toHaveProp('disabled', false);
+  });
+
+  it('renders a Not Found page when logged user cannot edit another user', () => {
```
bobsilverberg: Nit: "logged-in user".

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+    const username = 'current-logged-user';
```
bobsilverberg: Nit: `current-logged-in-user`.

bobsilverberg: I'm going to stop commenting on each of these, but I think any instance of "logged user" should say "logged-in user" (or "logged in user").

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -0,0 +1,601 @@
...
+  });
+
+  it('returns a 404 if current user is not logged in', () => {
+    const { store } = dispatchClientMetadata();
+
+    const root = renderUserProfileEdit({ store });
+
+    expect(root.find(NotFound)).toHaveLength(1);
+  });
+});
```
bobsilverberg: I think this is missing a test case for editing a user and changing the username, because there is currently a problem with that and it didn't cause any tests to fail.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Do you mean getting rid of the `= null` or the whole line?

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: To me, there are two intents: first making sure that we show a field, and then asserting one of its props.

tests/unit/amo/components/TestUserProfileEdit.js
bobsilverberg: I meant getting rid of the whole line, but maybe that won't work?

tests/unit/amo/components/TestUserProfileEdit.js
bobsilverberg: Right, but the second will definitely fail if the field isn't present, so we don't really need the first. But I'm fine leaving it in if you think it adds value.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: If the line gets removed, then we will have to check/assign `props.store`.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: One more thing: if the second line fails, you have to figure out why and there are two options. With two different assertions, you know why it fails directly.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: yes, I am so sorry for that :/

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: right, I totally missed that case 😞

willdurand: I have also added test cases for when we are editing another user profile and the username has changed (the URL must be updated)

willdurand:
Follow-up issues:

- https://github.com/mozilla/addons-frontend/issues/4997
- https://github.com/mozilla/addons-frontend/issues/4998

Issue #1262, opened 2016-10-20
`translate()` should accept i18n as props


mstriemer:
When testing it is rather inconvenient to set up a provider to put `i18n` into the context. If we allowed a parent component to set `i18n` from props then we wouldn't need to set up the provider.

This should just be a matter of reordering how we [set props on the wrapped component](https://github.com/mozilla/addons-frontend/blob/bd025076be7641da6115e6bcde704ad0eec72e85/src/core/i18n/translate.js#L37).

```jsx
render() {
  const extraProps = { i18n: this.i18n };
  if (withRef) {
    extraProps.ref = (ref) => { this.wrappedInstance = ref; };
  }
  return <WrappedComponent {...this.props} {...extraProps} />;
  // Should be:
  return <WrappedComponent {...extraProps} {...this.props} />;
}

PR #4923, opened 2018-05-03
Remove childContextTypes in Translate HOC

willdurand:
We have many warnings and I don't think we use context to pass the
`i18n` prop anymore. This should remove some warnings in the devtools.

### DIFF
--- a/src/core/i18n/translate.js
+++ b/src/core/i18n/translate.js
@@ -41,10 +41,6 @@ export default function translate(options = {}) {
       i18n: PropTypes.object.isRequired,
     };
 
-    Translate.childContextTypes = {
-      i18n: PropTypes.object.isRequired,
-    };
-
     Translate.displayName = `Translate[${getDisplayName(WrappedComponent)}]`;
 
     return Translate;


src/core/i18n/translate.js
```diff
@@ -41,10 +41,6 @@ export default function translate(options = {}) {
       i18n: PropTypes.object.isRequired,
```
muffinresearch: Can you remove this too ^ ? This looks like a left over from when the I18nProvider parts were extracted/refactored. 

I wonder if i18nProvider could possibly be ultimately removed if it's not used that much following what was started in https://github.com/mozilla/addons-frontend/issues/1262 - if not it might need updating to the newer context API if it still requires context.

willdurand: mmh don't we need this to tell the parent component we want the `i18n` object?

I will look at the I18Provider.

muffinresearch:
@willdurand So am I right in thinking that translate doesn't need the `childContextTypes` because it doesn't have a `getChildContext` method?

willdurand:
> @willdurand So am I right in thinking that translate doesn't need the childContextTypes here because it doesn't have a getChildContext method?

yes, that's what this patch fixes.

willdurand:
@muffinresearch here is an example of such a warning:

```
      Warning: Translate[SearchResultBase].childContextTypes is specified but there is no getChildContext() method on the instance. You can either define getChildContext() on Translate[SearchResultBase] or remove childContextTypes from it.
```

PR #4937, opened 2018-05-04
i-4541 fixes slug leading dash issue

### LINKED ISSUE #4541: Impossible to load an add-on detail page when its slug begins with a dash followed by only numbers
1. Submit an add-on with a slug being a dash (`-`) followed by numbers only.
2. Wait for it to be approved, visit its listing page

### What happened?
On the new frontend, the detail page only shows placeholder content.
37206690-6cd2b0e2-23a2-11e8-85ba-61dd1bc2bb60

It should be loaded without any issues. The legacy site works fine in this case.

Changing the slug to a dash followed by some letters works fine.

There are a few hundred add-ons (probably exclusively themes that were imported with a blank name a long time ago, see mozilla/addons-server#7599) on prod that are affected.

rebmullin:
fixes #4541 - adds a few more checks for slug before trying to redirect

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -107,8 +107,15 @@ export class AddonBase extends React.Component {
     // of an error.
     if (!errorHandler.hasError()) {
       if (addon) {
+        const { slug } = params;
+
+        // make sure slug is number or string without leading -
+        // before redirecting
+
         // eslint-disable-next-line no-restricted-globals
-        if (!isNaN(params.slug)) {
+        if (slug && !isNaN(slug)
+          && ((typeof slug === 'string' && slug.charAt(0) !== '-')
+          || (typeof slug === 'number'))) {
           // We only load add-ons by slug, but ID must be supported too because
           // it is a legacy behavior.
           dispatch(sendServerRedirect({
@@ -546,10 +553,13 @@ export function mapStateToProps(state, ownProps) {
   const { slug } = ownProps.params;
   let addon = getAddonBySlug(state, slug);
 
+  // make sure slug is number or string without leading -
   // It is possible to load an add-on by its ID but in the routing parameters,
   // the parameter is always named `slug`.
   // eslint-disable-next-line no-restricted-globals
-  if (slug && !isNaN(slug)) {
+  if (slug && !isNaN(slug)
+    && ((typeof slug === 'string' && slug.charAt(0) !== '-')
+    || (typeof slug === 'number'))) {
     addon = getAddonByID(state, slug);
   }
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -60,12 +60,10 @@ import ShowMoreCard from 'ui/components/ShowMoreCard';
 import './styles.scss';
 
 
-// Find out if slug is invalid.
-const isInvalidSlugID = (slug) => {
+// Find out if slug converts to a positive number/ID.
+const slugIsPositiveID = (slug) => {
   // eslint-disable-next-line no-restricted-globals
-  return slug && !isNaN(slug)
-    && ((typeof slug === 'string' && slug.charAt(0) !== '-')
-    || (typeof slug === 'number'));
+  return !isNaN(slug) && parseInt(slug, 10) > 0;
 };
 
 export class AddonBase extends React.Component {
@@ -117,9 +115,9 @@ export class AddonBase extends React.Component {
       if (addon) {
         const { slug } = params;
 
-        // We want to make sure the slug is invalid before
-        // before we try redirecting.
-        if (isInvalidSlugID(slug)) {
+        // We want to make sure the slug converts to a positive
+        // number/ID before we try redirecting.
+        if (slugIsPositiveID(slug)) {
           // We only load add-ons by slug, but ID must be supported too because
           // it is a legacy behavior.
           dispatch(sendServerRedirect({
@@ -559,7 +557,7 @@ export function mapStateToProps(state, ownProps) {
 
   // It is possible to load an add-on by its ID but in the routing parameters,
   // the parameter is always named `slug`.
-  if (isInvalidSlugID(slug)) {
+  if (slugIsPositiveID(slug)) {
     addon = getAddonByID(state, slug);
   }
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -60,6 +60,14 @@ import ShowMoreCard from 'ui/components/ShowMoreCard';
 import './styles.scss';
 
 
+// Find out if slug is invalid.
+const isInvalidSlugID = (slug) => {
+  // eslint-disable-next-line no-restricted-globals
+  return slug && !isNaN(slug)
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -546,10 +553,13 @@ export function mapStateToProps(state, ownProps) {
...
   let addon = getAddonBySlug(state, slug);
 
+  // make sure slug is number or string without leading -
   // It is possible to load an add-on by its ID but in the routing parameters,
   // the parameter is always named `slug`.
   // eslint-disable-next-line no-restricted-globals
-  if (slug && !isNaN(slug)) {
+  if (slug && !isNaN(slug)
+    && ((typeof slug === 'string' && slug.charAt(0) !== '-')
+    || (typeof slug === 'number'))) {
```
willdurand: This makes me think that we should probably extract all these checks into a single utility function above the component.

src/amo/components/Addon/index.js
```diff
@@ -107,8 +107,15 @@ export class AddonBase extends React.Component {
     // of an error.
     if (!errorHandler.hasError()) {
       if (addon) {
+        const { slug } = params;
+
+        // make sure slug is number or string without leading -
+        // before redirecting
```
willdurand: We tend to write comments as full English sentences.

tests/unit/amo/components/TestAddon.js
```diff
@@ -461,6 +461,25 @@ describe(__filename, () => {
     sinon.assert.callCount(fakeDispatch, 1);
   });
 
+  it('does not dispatch a server redirect when slug is a stringified integer that starts with -', () => {
+    const clientApp = CLIENT_APP_FIREFOX;
```
willdurand: Could you maybe add the link to the GitHub issue in a comment?

tests/unit/amo/components/TestAddon.js
```diff
@@ -461,6 +461,25 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      slug: '-1234',
+    });
+
+    store.dispatch(_loadAddons({ addon }));
+
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+    renderComponent({
+      params: { slug: `-${addon.id}` }, store,
```
willdurand: why not using `addon.slug`?

rebmullin: yep - thanks : )

src/amo/components/Addon/index.js
rebmullin: ahh! ok..

rebmullin:
hey @willdurand  - thanks for the feedback. i've updated... when you can please take another look : )

src/amo/components/Addon/index.js
```diff
@@ -60,6 +60,14 @@ import ShowMoreCard from 'ui/components/ShowMoreCard';
 import './styles.scss';
 
 
+// Find out if slug is invalid.
+const isInvalidSlugID = (slug) => {
```
willdurand: The slug is not really invalid actually, it is "just" not a string no? If so, we should probably update the function name.

src/amo/components/Addon/index.js
```diff
@@ -60,6 +60,14 @@ import ShowMoreCard from 'ui/components/ShowMoreCard';
...
+  // eslint-disable-next-line no-restricted-globals
+  return slug && !isNaN(slug)
+    && ((typeof slug === 'string' && slug.charAt(0) !== '-')
+    || (typeof slug === 'number'));
```
willdurand: We should have a test case for this too and I don't see one.

src/amo/components/Addon/index.js
```diff
@@ -107,8 +115,11 @@ export class AddonBase extends React.Component {
     // of an error.
     if (!errorHandler.hasError()) {
       if (addon) {
-        // eslint-disable-next-line no-restricted-globals
-        if (!isNaN(params.slug)) {
+        const { slug } = params;
+
+        // We want to make sure the slug is invalid before
+        // before we try redirecting.
```
willdurand: `before` appears twice in this sentence.

tests/unit/amo/components/TestAddon.js
```diff
@@ -461,6 +461,26 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      slug: '-1234',
+    });
+
+    store.dispatch(_loadAddons({ addon }));
+
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+    renderComponent({
+      params: { slug: `${addon.slug}` }, store,
```
willdurand: so I guess, we could write: `{ slug: addon.slug },` no?

rebmullin: 🤦‍♂️ yes...

src/amo/components/Addon/index.js
rebmullin: opps..yepp

src/amo/components/Addon/index.js
rebmullin: thanks i'll take another look..

src/amo/components/Addon/index.js
rebmullin: hmm yeh i'll take another look at the name, thanks

src/amo/components/Addon/index.js
rebmullin: hey @willdurand  - do you mean if it's actually a number? I guess i thought this test was covered under `dispatches a server redirect when slug is a numeric ID` - or let me know which piece.. maybe i misunderstood..?

rebmullin: it also checks for all these pieces with the following tests:
`dispatches a server redirect when slug is a numeric ID`
`dispatches a server redirect when slug is a stringified integer`
`does not dispatch a server redirect when slug is a stringified integer that starts with -`

 i believe this covers all these pieces? are you saying these should be combined into one test?

willdurand: It is probably fine. I was a bit confused because we did not check the type before but you have to do it because you check the first char.

willdurand: Maybe we could get rid of the parentheses btw

src/amo/components/Addon/index.js
rebmullin: actually from my understanding it could be a number type (legacy) or it could be a string which looks like a number (1234) - either of these cases should redirect. The problem was there are some strings like '-1234' which look like a number but we need to exclude these cases (not redirect)... so i am not sure of a better name - maybe `isSlugValidNumber` ?

src/amo/components/Addon/index.js
rebmullin: right..i added type check b/c of charAt piece...

i actually had some less parentheses but linting complained (`Unexpected mix of '&&' and '||'  no-mixed-operators`) unless i did i this way..

src/amo/components/Addon/index.js
willdurand: How about `slugIsID`?

rebmullin:
yes..you're right and i think that might be a better/cleaner approach. I will try to update the code this..

PR #4941, opened 2018-05-05
Fixes 4919: autoflls slug

### LINKED ISSUE #4919: When creating a new collection, the slug should be prefilled
As discussed with @pwalm a couple of weeks ago, in order to speed the process of creating a collection, on the "Create a collection" form, we should pre-fill the stub field so the user only has to enter a Collection name in order to save the collection. They will be free to change the slug if they wish, but they won't have to.

We can use whatever the user entered for the name as the basis for the slug, replacing spaces with `-` and removing invalid characters.

tsl143:
Fixes #4919 

autofillslug

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -55,7 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
-  customSlug?: boolean | false,
+  customSlug?: boolean,
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
@@ -155,21 +155,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    const { name = '', value = '' } = event.target;
+    const { name, value = '' } = event.target;
     const { creating } = this.props;
 
     if (creating && name === 'name' && !this.state.customSlug) {
       this.setState({
         slug: value.trim().replace(/[^A-Za-z0-9]/g, '-'),
-        [event.target.name]: event.target.value,
+        [name]: value,
       });
     } else if (creating && name === 'slug' && value.trim() !== '') {
       this.setState({
         customSlug: true,
-        [event.target.name]: event.target.value,
+        [name]: value,
       });
     } else {
-      this.setState({ [event.target.name]: event.target.value });
+      this.setState({ [name]: value });
     }
   };
 
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -55,7 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
-  userEnterSlug?: boolean | false,
+  customSlug?: boolean | false,
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
@@ -155,8 +155,22 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    this.handleSlug(event.target);
-    this.setState({ [event.target.name]: event.target.value });
+    const { name = '', value = '' } = event.target;
+    const { creating } = this.props;
+
+    if (creating && name === 'name' && !this.state.customSlug) {
+      this.setState({
+        slug: value.trim().replace(/[^A-Za-z0-9]/g, '-'),
+        [event.target.name]: event.target.value,
+      });
+    } else if (creating && name === 'slug' && value.trim() !== '') {
+      this.setState({
+        customSlug: true,
+        [event.target.name]: event.target.value,
+      });
+    } else {
+      this.setState({ [event.target.name]: event.target.value });
+    }
   };
 
   onSearchAddon = (filters: SearchFilters) => {
@@ -184,21 +198,6 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       userId: siteUserId,
     }));
   };
-
-
-  handleSlug = (targetElement: Object) => {
-    const { name = '', value = '' } = targetElement;
-    const { creating } = this.props;
-    // only for add collection
-    if (!creating) return true;
-
-    if (name === 'name' && !this.state.userEnterSlug) {
-      this.setState({ slug: value.replace(/[^A-Za-z0-9]/g, '-') });
-    } else if (name === 'slug' && value.trim() !== '') {
-      this.setState({ userEnterSlug: true });
-    }
-    return true;
-  }
 
   propsToState(props: Props) {
     // Decode HTML entities so the user sees real symbols in the form.
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -155,15 +155,18 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    const { name, value = '' } = event.target;
+    const { name, value } = event.target;
     const { creating } = this.props;
+
+    if (value === null || typeof value === 'undefined') return;
+    const trimmedValue = value.trim();
 
     if (creating && name === 'name' && !this.state.customSlug) {
       this.setState({
-        slug: value.trim().replace(/[^A-Za-z0-9]/g, '-'),
+        slug: trimmedValue.replace(/[^A-Za-z0-9]/g, '-'),
         [name]: value,
       });
-    } else if (creating && name === 'slug' && value.trim() !== '') {
+    } else if (creating && name === 'slug' && trimmedValue !== '') {
       this.setState({
         customSlug: true,
         [name]: value,
@@ -202,6 +205,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
   propsToState(props: Props) {
     // Decode HTML entities so the user sees real symbols in the form.
     return {
+      customSlug: false,
       description: props.collection &&
         decodeHtmlEntities(props.collection.description),
       name: props.collection && decodeHtmlEntities(props.collection.name),
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -55,6 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
+  userEnterSlug?: boolean | false,
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
@@ -154,6 +155,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
+    this.handleSlug(event.target);
     this.setState({ [event.target.name]: event.target.value });
   };
 
@@ -182,6 +184,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       userId: siteUserId,
     }));
   };
+
+
+  handleSlug = (targetElement: Object) => {
+    const { name = '', value = '' } = targetElement;
+    const { creating } = this.props;
+    // only for add collection
+    if (!creating) return true;
+
+    if (name === 'name' && !this.state.userEnterSlug) {
+      this.setState({ slug: value.replace(/[^A-Za-z0-9]/g, '-') });
+    } else if (name === 'slug' && value.trim() !== '') {
+      this.setState({ userEnterSlug: true });
+    }
+    return true;
+  }
 
   propsToState(props: Props) {
     // Decode HTML entities so the user sees real symbols in the form.
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -158,7 +158,10 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     const { name, value } = event.target;
     const { creating } = this.props;
 
-    if (value === null || typeof value === 'undefined') return;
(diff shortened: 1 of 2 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -55,6 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
+  userEnterSlug?: boolean | false,
```
willdurand: I would name this property: `customSlug`

src/amo/components/CollectionManager/index.js
```diff
@@ -183,6 +185,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+    // only for add collection
+    if (!creating) return true;
+
+    if (name === 'name' && !this.state.userEnterSlug) {
+      this.setState({ slug: value.replace(/[^A-Za-z0-9]/g, '-') });
+    } else if (name === 'slug' && value.trim() !== '') {
+      this.setState({ userEnterSlug: true });
+    }
+    return true;
+  }
```
willdurand: Maybe this would be better in the `onTextInput` method so that we could only call `setState` one time.

src/amo/components/CollectionManager/index.js
```diff
@@ -183,6 +185,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
 
+
+  handleSlug = (targetElement: Object) => {
+    const { name = '', value = '' } = targetElement;
+    const { creating } = this.props;
+    // only for add collection
+    if (!creating) return true;
+
+    if (name === 'name' && !this.state.userEnterSlug) {
+      this.setState({ slug: value.replace(/[^A-Za-z0-9]/g, '-') });
```
willdurand: It is a good start but we should maybe be a bit more defensive, see when I enter empty spaces:

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,20 @@ describe(__filename, () => {
     }));
   });
 
+  it('autofills slug when name is entered while creating collection', () => {
+    const name = "trishul's collection";
+    const localStore = dispatchClientMetadata().store;
+    dispatchSignInActions({
+      store: localStore,
+    });
```
willdurand: Why do you need this actually? If I am not mistaken, it is already initialized like this in a `beforeEach` hook.

src/amo/components/CollectionManager/index.js
tsl143: I really suck at naming :( 
will update :)

tsl143:
@willdurand I am not sure why checks are failing, but from my side everything is all set.

willdurand:
> @willdurand I am not sure why checks are failing, but from my side everything is all set.

There was an issue with travis-ci earlier today.

src/amo/components/CollectionManager/index.js
```diff
@@ -55,6 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
+  customSlug?: boolean | false,
```
willdurand: I think you don't need `| false` here because we only describe a type, and `boolean` is the type.

src/amo/components/CollectionManager/index.js
```diff
@@ -154,7 +155,22 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+        slug: value.trim().replace(/[^A-Za-z0-9]/g, '-'),
+        [event.target.name]: event.target.value,
+      });
+    } else if (creating && name === 'slug' && value.trim() !== '') {
+      this.setState({
+        customSlug: true,
+        [event.target.name]: event.target.value,
+      });
+    } else {
+      this.setState({ [event.target.name]: event.target.value });
```
willdurand: maybe you want to use `name` and `value` here since you de-structured `event.target` on line 158.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,15 @@ describe(__filename, () => {
...
   });
 
+  it('autofills slug when name is entered while creating collection', () => {
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+
+    typeInput({ root, name: 'name', text: name });
+
+    expect(root).toHaveState('slug', 'trishul-s-collection');
+  });
```
willdurand: You may want to add a test case to cover the `customSlug` case.

src/amo/components/CollectionManager/index.js
tsl143: I tried this before and got flow error
screen shot 2018-05-08 at 7 14 05 am

But figured out the error, so yes now it looks cleaner.

tsl143:
@willdurand updated.

src/amo/components/CollectionManager/index.js
```diff
@@ -154,7 +155,22 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    this.setState({ [event.target.name]: event.target.value });
+    const { name, value = '' } = event.target;
```
willdurand: I think if `value` is either not defined or `null`, we can early return. What do you think?

src/amo/components/CollectionManager/index.js
```diff
@@ -55,6 +55,7 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
+  customSlug?: boolean,
```
willdurand: We should probably also add a default value into the initial state (in the constructor).

src/amo/components/CollectionManager/index.js
```diff
@@ -154,7 +155,22 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
-    this.setState({ [event.target.name]: event.target.value });
+    const { name, value = '' } = event.target;
+    const { creating } = this.props;
+
+    if (creating && name === 'name' && !this.state.customSlug) {
+      this.setState({
+        slug: value.trim().replace(/[^A-Za-z0-9]/g, '-'),
+        [name]: value,
+      });
+    } else if (creating && name === 'slug' && value.trim() !== '') {
```
willdurand: I think we often use `value.trim().length`

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,25 @@ describe(__filename, () => {
...
+
+  it('stops autofill slug when custom slug is entered while creating collection', () => {
```
willdurand: s/stops/does not

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,25 @@ describe(__filename, () => {
...
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+
+    typeInput({ root, name: 'slug', text: 'trishul' });
+    typeInput({ root, name: 'name', text: name });
+
+    expect(root).toHaveState('slug', 'trishul');
```
willdurand: you should create a variable `const slug = 'trishul';`

src/amo/components/CollectionManager/index.js
tsl143: It will not make difference as we have virtue of Boolean, in every comparison undefined will be `falsey`.

src/amo/components/CollectionManager/index.js
tsl143: okay, and also can handle `value.trim()` in this condition.

src/amo/components/CollectionManager/index.js
willdurand: That's not very clean though 😉

tsl143:
@willdurand hoping this settles everything :P

src/amo/components/CollectionManager/index.js
```diff
@@ -154,7 +155,25 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    this.setState({ [event.target.name]: event.target.value });
+    const { name, value } = event.target;
+    const { creating } = this.props;
+
+    if (value === null || typeof value === 'undefined') return;
```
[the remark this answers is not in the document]
tsl143: > We always use (curly) brackets.

lets put a linting rule for this?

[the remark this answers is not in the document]
willdurand: can we do this?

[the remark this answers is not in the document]
willdurand: > !value will not work, as it will consider blank spaces too.

ok I was not sure to be honest, 👍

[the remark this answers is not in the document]
tsl143: > can we do this?

I will give a quick look, meanwhile I will update PR with `{}` so that this can be merged.



> ok I was not sure to be honest, 👍

Me too, I encountered once I wasnt able to enter space in `name`

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -435,6 +435,24 @@ describe(__filename, () => {
...
 
+  it('autofills slug with trimmed collection name', () => {
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+    typeInput({ root, name: 'name', text: `  ${name}  ` });
+
+    expect(root).toHaveState('slug', 'trishul-s-collection');
+  });
+
+
```
willdurand: We should only have one blank line

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -435,6 +435,24 @@ describe(__filename, () => {
...
+  it('doesnt update slug if event value is null', () => {
```
willdurand: s/doesnt/does not

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -435,6 +435,24 @@ describe(__filename, () => {
...
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+    typeInput({ root, name: 'name', text: name });
+    typeInput({ root, name: 'slug' });
```
willdurand: Can you make sure it is `null` by specifying it? If you dont, it is `undefined` no?

tsl143:
phew, sorted !
@willdurand

src/amo/components/CollectionManager/index.js
```diff
@@ -154,7 +155,28 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     event: ElementEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
     event.preventDefault();
-    this.setState({ [event.target.name]: event.target.value });
+    const { name, value } = event.target;
+    const { creating } = this.props;
+
+    if (value === null || typeof value === 'undefined') {
```
willdurand: I think the second part of this condition deserves a test too.

tsl143: I am not sure when are we expecting `event.target.value` to be `undefined` or `null`.
DO we have any case for that?

willdurand: Ah well, I don't know. If that never happens, we should not have this condition no?

tsl143:
@bobsilverberg @willdurand is it good to get merged?

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,53 @@ describe(__filename, () => {
...
+    typeInput({ root, name: 'slug', text: slug });
+    typeInput({ root, name: 'name', text: name });
+
+    expect(root).toHaveState('slug', slug);
+  });
+
+  it('autofills slug with trimmed collection name', () => {
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+    typeInput({ root, name: 'name', text: `  ${name}  ` });
```
willdurand: can you add a blank line before this line? (as you did for the previous test cases)

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,53 @@ describe(__filename, () => {
...
+
+    expect(root).toHaveState('slug', 'trishul-s-collection');
+  });
+
+  it('does not update slug if event value is null', () => {
+    const name = "trishul's collection";
+    const root = render({ collection: null, creating: true });
+    typeInput({ root, name: 'name', text: name });
```
willdurand: same here (for the blank line)

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,53 @@ describe(__filename, () => {
...
+    typeInput({ root, name: 'slug', text: null });
+
+    expect(root).toHaveState('slug', 'trishul-s-collection');
+  });
+
+  it('does not update slug if event value is undefined', () => {
+    const name = "trishul's collection";
+    let slug;
```
willdurand: I guess you don't really need this.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -415,6 +415,53 @@ describe(__filename, () => {
...
+    const root = render({ collection: null, creating: true });
+    typeInput({ root, name: 'name', text: name });
+    typeInput({ root, name: 'slug', text: slug });
```
willdurand: can you write `text: undefined` instead?

tsl143:
Sounds good, but I would prefer merging this PR as I made changes literally
from Airport. I will give another PR for multi consecutive spaces, anyways
that's also kinda not in scope of this issue

willdurand:
I think your PR and therefore feature will be much better with this small change so let's do it here instead. I have been thinking about other tweaks, but could not come up with anything else so I guess it is the last change 😺

tsl143:
Hey @willdurand , it actually looks neater now, updated the PR

tsl143:
@willdurand , thanks for the merge, now we can work on more tweaks in this, please mention me when you create issues for the same :)

PR #4942, opened 2018-05-07
Use taar for recommendations

### LINKED ISSUE #4866: Implement an AddonRecommendations component which uses TAAR Lite as a data source

bobsilverberg:
Fixes #4866 

Note that this builds upon the work in PR #4904, so the first two commits can be ignored for the purpose of this review.

Note as well that I have addressed all of the issues raised against PR #4869 and included those changes in this code.

Also note that all of the analytics reporting hasn't been fully figured out yet, but I think the general design of this component is ready for a review. I think the tests could use a bit of polish, but I wanted to submit the code for review now, with the hope of possibly landing it in this week's tag.

### DIFF
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -56,7 +56,7 @@ export class AddonRecommendationsBase extends React.Component<Props> {
     tracking: defaultTracking,
   };
 
-  componentWillMount() {
+  componentDidMount() {
     const { cookie, randomizer, addon, recommendations } = this.props;
 
     // Set a cohort for the experiment.
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -1,6 +1,5 @@
 /* @flow */
 import makeClassName from 'classnames';
-import { oneLine } from 'common-tags';
 import invariant from 'invariant';
 import * as React from 'react';
 import defaultCookie from 'react-cookie';
@@ -11,10 +10,13 @@ import AddonsCard from 'amo/components/AddonsCard';
 import {
   fetchRecommendations,
   getRecommendationsByGuid,
+  OUTCOME_RECOMMENDED,
 } from 'amo/reducers/recommendations';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
+import log from 'core/logger';
 import defaultTracking from 'core/tracking';
+import LoadingText from 'ui/components/LoadingText';
 import type {
   Recommendations,
   RecommendationsState,
@@ -25,7 +27,7 @@ import type { AddonType } from 'core/types/addons';
 import type { DispatchFunc } from 'core/types/redux';
 
 export const TAAR_IMPRESSION_CATEGORY = 'AMO Addon / Recommendations Shown';
-export const TAAR_COHORT_COOKIE_NAME = 'TAARCohort';
+export const TAAR_COHORT_COOKIE_NAME = 'taar_cohort';
 export const TAAR_COHORT_INCLUDED: 'TAAR_COHORT_INCLUDED'
   = 'TAAR_COHORT_INCLUDED';
 export const TAAR_COHORT_EXCLUDED: 'TAAR_COHORT_EXCLUDED'
@@ -34,15 +36,15 @@ export const TAAR_COHORT_EXCLUDED: 'TAAR_COHORT_EXCLUDED'
 export type CohortName = typeof TAAR_COHORT_INCLUDED | typeof TAAR_COHORT_EXCLUDED;
 
 type Props = {|
-  cookie: typeof defaultCookie,
-  randomizer: () => number,
-  tracking: typeof defaultTracking,
   addon: AddonType | null,
   className?: string,
+  cookie: typeof defaultCookie,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  randomizer: () => number,
   recommendations: Recommendations | null,
+  tracking: typeof defaultTracking,
 |};
 
 
@@ -50,8 +52,8 @@ export class AddonRecommendationsBase extends React.Component<Props> {
   static defaultProps = {
     cookie: defaultCookie,
     randomizer: Math.random,
-    tracking: defaultTracking,
     recommendations: null,
+    tracking: defaultTracking,
   };
 
   componentWillMount() {
@@ -65,7 +67,7 @@ export class AddonRecommendationsBase extends React.Component<Props> {
       cookie.save(TAAR_COHORT_COOKIE_NAME, this.cohort, { path: '/' });
     }
 
-    if (!!addon && !recommendations) {
+    if (addon && !recommendations) {
       this.dispatchFetchRecommendations({
         guid: addon.guid,
         recommended: this.cohort === TAAR_COHORT_INCLUDED,
@@ -84,7 +86,7 @@ export class AddonRecommendationsBase extends React.Component<Props> {
     } = this.props;
 
     // Fetch recommendations when the add-on changes.
-    if (oldAddon !== newAddon && !!newAddon) {
+    if (newAddon && oldAddon !== newAddon) {
       this.dispatchFetchRecommendations({
         guid: newAddon.guid,
         recommended: this.cohort === TAAR_COHORT_INCLUDED,
@@ -92,7 +94,7 @@ export class AddonRecommendationsBase extends React.Component<Props> {
     }
 
     // Send the GA ping when recommendations are loaded.
-    if (oldRecommendations !== newRecommendations && !!newRecommendations) {
+    if (newRecommendations && oldRecommendations !== newRecommendations) {
       const { fallbackReason, loading, outcome } = newRecommendations;
 
       if (loading) {
@@ -103,8 +105,10 @@ export class AddonRecommendationsBase extends React.Component<Props> {
       invariant(outcome, 'outcome is required');
 
       // TODO: Determine the exact format for this output.
-      const action = oneLine`outcome: ${outcome} |
-        fallbackReason: ${fallbackReason || ''}`;
+      let action = outcome;
+      if (fallbackReason) {
+        action = `${action}-${fallbackReason}`;
+      }
       tracking.sendEvent({
         action,
         category: TAAR_IMPRESSION_CATEGORY,
@@ -127,17 +131,28 @@ export class AddonRecommendationsBase extends React.Component<Props> {
     const { className, i18n, recommendations } = this.props;
 
     if (!recommendations) {
+      log.debug(
+        'No recommandations, hiding the AddonRecommendations component.'
+      );
       return null;
     }
 
-    const { addons, loading } = recommendations;
+    const { addons, loading, outcome } = recommendations;
     const classnames = makeClassName('AddonRecommendations', className);
+
+    let header = loading ? <LoadingText width={100} /> : null;
+    if (!header) {
+      header = outcome === OUTCOME_RECOMMENDED ?
+        i18n.gettext('Other users with this extension also installed') :
+        i18n.gettext('Other popular extensions');
+    }
 
     return (
       <AddonsCard
+        addonInstallSource={outcome || ''}
         addons={addons}
         className={classnames}
-        header={i18n.gettext('You might also like...')}
+        header={header}
         loading={loading}
         placeholderCount={4}
         showMetadata
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -1,0 +1,169 @@
+/* @flow */
+import makeClassName from 'classnames';
+import { oneLine } from 'common-tags';
+import invariant from 'invariant';
+import * as React from 'react';
+import defaultCookie from 'react-cookie';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+
+import AddonsCard from 'amo/components/AddonsCard';
+import {
+  fetchRecommendations,
+  getRecommendationsByGuid,
+} from 'amo/reducers/recommendations';
+import { withErrorHandler } from 'core/errorHandler';
+import translate from 'core/i18n/translate';
+import defaultTracking from 'core/tracking';
+import type {
+  Recommendations,
+  RecommendationsState,
+} from 'amo/reducers/recommendations';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { I18nType } from 'core/types/i18n';
+import type { AddonType } from 'core/types/addons';
+import type { DispatchFunc } from 'core/types/redux';
+
+export const TAAR_IMPRESSION_CATEGORY = 'AMO Addon / Recommendations Shown';
+export const TAAR_COHORT_COOKIE_NAME = 'TAARCohort';
+export const TAAR_COHORT_INCLUDED: 'TAAR_COHORT_INCLUDED'
+  = 'TAAR_COHORT_INCLUDED';
+export const TAAR_COHORT_EXCLUDED: 'TAAR_COHORT_EXCLUDED'
+  = 'TAAR_COHORT_EXCLUDED';
+
+export type CohortName = typeof TAAR_COHORT_INCLUDED | typeof TAAR_COHORT_EXCLUDED;
+
+type Props = {|
+  cookie: typeof defaultCookie,
+  randomizer: () => number,
+  tracking: typeof defaultTracking,
+  addon: AddonType | null,
+  className?: string,
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  recommendations: Recommendations | null,
+|};
+
+
+export class AddonRecommendationsBase extends React.Component<Props> {
+  static defaultProps = {
+    cookie: defaultCookie,
+    randomizer: Math.random,
+    tracking: defaultTracking,
+    recommendations: null,
+  };
+
+  componentWillMount() {
+    const { cookie, randomizer, addon, recommendations } = this.props;
+
+    // Set a cohort for the experiment.
+    this.cohort = cookie.load(TAAR_COHORT_COOKIE_NAME);
+    if (this.cohort === undefined) {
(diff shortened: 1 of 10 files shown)


src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+      if (loading) {
+        return;
+      }
+
+      invariant(newAddon, 'newAddon is required');
+      invariant(outcome, 'outcome is required');
+
+      // TODO: Determine the exact format for this output.
+      const action = oneLine`outcome: ${outcome} |
+        fallbackReason: ${fallbackReason || ''}`;
```
willdurand: Yup, I was really confused when I saw this in my dev tools:

```
"eventAction":"outcome: recommended | fallbackReason:"
```

src/amo/reducers/recommendations.js
```diff
@@ -12,7 +12,7 @@ export const LOAD_RECOMMENDATIONS: 'LOAD_RECOMMENDATIONS'
   = 'LOAD_RECOMMENDATIONS';
 
 export type Recommendations = {|
-  addons: Array<ExternalAddonType> | null,
+  addons: Array<AddonType> | null,
```
willdurand: 👍

src/amo/reducers/recommendations.js
```diff
@@ -92,7 +92,6 @@ export const loadRecommendations = ({
   outcome,
 }: LoadRecommendationsParams = {}): LoadRecommendationsAction => {
   invariant(addons, 'addons is required');
-  invariant(fallbackReason, 'fallbackReason is required');
```
willdurand: 👍

src/amo/sagas/index.js
```diff
@@ -31,6 +32,7 @@ export default function* rootSaga() {
     fork(home),
     fork(landing),
     fork(languageTools),
+    fork(recommendations),
```
willdurand: 👍

src/amo/sagas/recommendations.js
```diff
@@ -37,7 +37,7 @@ export function* fetchRecommendations({
       outcome,
     }));
   } catch (error) {
-    log.warn(`Failed to fetch user collections: ${error}`);
+    log.warn(`Failed to fetch recommendations: ${error}`);
```
willdurand: 👍

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+
+export class AddonRecommendationsBase extends React.Component<Props> {
+  static defaultProps = {
+    cookie: defaultCookie,
+    randomizer: Math.random,
+    tracking: defaultTracking,
+    recommendations: null,
+  };
+
+  componentWillMount() {
```
willdurand: I don't think this will be working as intended, for instance, I cannot get any recommendations from the server. It works well in client side mode but not in server side mode:

screen shot 2018-05-07 at 10 49 55

I think you can reproduce this issue by navigating to an addon detail page and hit "reload" to fully reload the page. It does not work because:

1. the server sends the request to fetch the recommandations
2. the state sets a new `recommendation` entry to the recommendations state (with `loading: true`)
3. the server `END`s the sagas but the recommendations saga was not finished
4. the client loads the preloaded state (from the server) with the `recommendation` entry to the recommendations state (with `loading: true`), 💥 deadlock

I see that we explicitly by-pass the server side rendering in `AddAddonToCollection`: https://github.com/mozilla/addons-frontend/blob/2111ff06c6a9499d39a7d1db29d828285b6986ee/src/amo/components/AddAddonToCollection/index.js#L67-L72

willdurand: Using the `componentDidMount()` method instead of `componentWillMount()` would work here.

willdurand: In addition: the screenshot above was taken to show a UI issue (the card height is not big enough)

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+    recommendations: newRecommendations,
+  }: Props) {
+    const {
+      addon: oldAddon,
+      recommendations: oldRecommendations,
+      tracking,
+    } = this.props;
+
+    // Fetch recommendations when the add-on changes.
+    if (oldAddon !== newAddon && !!newAddon) {
```
willdurand: I would write `&& newAddon` here.

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+      guid,
+      recommended,
+    }));
+  }
+
+  render() {
+    const { className, i18n, recommendations } = this.props;
+
+    if (!recommendations) {
+      return null;
```
willdurand: Maybe we should add a `log.debug()` message here, like "No recommandations, hiding the `AddonRecommendations` component". What do you think?

tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,224 @@
...
+  let store;
+
+  beforeEach(() => {
+    fakeTracking = {
+      sendEvent: sinon.spy(),
+    };
+    store = dispatchClientMetadata().store;
+  });
+
+  function doFetchRecommendations({
```
willdurand: Ah I think that's where the `_` naming is useful, so that it does not conflict with imported functions. But I am fine with `do` as a prefix 😉

tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,224 @@
...
+    );
+  }
+
+  it('renders nothing without an addon', () => {
+    const root = render({ addon: null });
+    expect(root).not.toHaveClassName('AddonRecommendations');
+  });
+
+  it('renders an AddonCard when recommendations are loaded', () => {
+    const addons = [createInternalAddon(fakeAddon)];
```
willdurand: I thought we received `Array<ExternalAddonType>` in `loadRecommendations()`?

tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,224 @@
...
+
+    root.setProps({ recommendations: fakeRecommendations });
+
+    sinon.assert.calledOnce(fakeTracking.sendEvent);
+    sinon.assert.calledWith(fakeTracking.sendEvent, {
+      action: `outcome: ${outcome} | fallbackReason: ${fallbackReason}`,
+      category: TAAR_IMPRESSION_CATEGORY,
+      label: fakeAddon.name,
+    });
+  });
```
willdurand: I think we need a test for the early `loading` return in case we have a recommendation but it is still loading.

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+  Recommendations,
+  RecommendationsState,
+} from 'amo/reducers/recommendations';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { I18nType } from 'core/types/i18n';
+import type { AddonType } from 'core/types/addons';
+import type { DispatchFunc } from 'core/types/redux';
+
+export const TAAR_IMPRESSION_CATEGORY = 'AMO Addon / Recommendations Shown';
+export const TAAR_COHORT_COOKIE_NAME = 'TAARCohort';
```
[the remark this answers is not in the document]
bobsilverberg: I don't believe anything has been prescribed regarding the cookie name. Is snake/lower-cased a standard format for cookies?

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+
+    // Set a cohort for the experiment.
+    this.cohort = cookie.load(TAAR_COHORT_COOKIE_NAME);
+    if (this.cohort === undefined) {
+      this.cohort = randomizer() > 0.5 ?
+        TAAR_COHORT_INCLUDED : TAAR_COHORT_EXCLUDED;
+      cookie.save(TAAR_COHORT_COOKIE_NAME, this.cohort, { path: '/' });
+    }
+
+    if (!!addon && !recommendations) {
```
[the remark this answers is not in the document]
bobsilverberg: I thought I ran into a problem with that, when `addon` was `null`, but now I don't see why that would have been. Perhaps it was with a different iteration of the code. Yes, I'll switch it to `addon`.

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,169 @@
...
+    // Fetch recommendations when the add-on changes.
+    if (oldAddon !== newAddon && !!newAddon) {
+      this.dispatchFetchRecommendations({
+        guid: newAddon.guid,
+        recommended: this.cohort === TAAR_COHORT_INCLUDED,
+      });
+    }
+
+    // Send the GA ping when recommendations are loaded.
+    if (oldRecommendations !== newRecommendations && !!newRecommendations) {
```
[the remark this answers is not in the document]
bobsilverberg: Again, I'm not sure why I did this. `newRecommendations` is not a boolean, it's a Recommendations type, so I don't think I need the `!!` at all.

src/amo/components/AddonRecommendations/index.js
bobsilverberg: We definitely need more info from them about how they want this reported, but, looking at the above, I wonder if we should either be omitting `fallbackReason` from the `action` if it's empty, or reporting `fallbackReason: n/a`.

What do you think?

src/amo/components/AddonRecommendations/index.js
bobsilverberg: Can we move everything in `componentWillMount` to `componentDidMount`, including the cookie setting code, or do we just want to move the call to `this.dispatchFetchRecommendations` into `componentDidMount`?

Also, why is this an issue with this particular component? It seems like a common pattern we have to call a fetch action in `componentWillMount` and it isn't causing a problem in other components where we do it.

src/amo/components/AddonRecommendations/index.js
bobsilverberg: Sure.

src/amo/components/AddonRecommendations/index.js
willdurand: > Can we move everything in componentWillMount to componentDidMount, including the cookie setting code, or do we just want to move the call to this.dispatchFetchRecommendations into componentDidMount?

I guess, yes.

> Also, why is this an issue with this particular component? It seems like a common pattern we have to call a fetch action in componentWillMount and it isn't causing a problem in other components where we do it.

It depends on the components I think. Because of the way we fetch the data and stops the sagas on the server, we have to be very careful. For some components, we by-pass data fetching on the server, for some others, the parent component gets the data and we also have the case where there is no `loading` indicator, which basically allows the client to fetch the data again.

tests/unit/amo/components/TestAddonRecommendations.js
bobsilverberg: Good catch, thanks!

src/amo/components/AddonRecommendations/index.js
bobsilverberg: Thanks. I tried to fix the issue with the size of the card when it's loading, but I couldn't get it to work. Increasing the padding at the bottom increased the distance from the bottom of the box to the footer, but didn't increase the size of the box (because of the white background, I think). This may require some more time and/or better css chops, so I think I want to defer that to a follow-up.

willdurand: Sure, I was thinking maybe that's because we have 4 addons here. Can we have 3 "loading" addons instead? Not sure if that would change anything.

bobsilverberg: 3 doesn't make it any better. :(

willdurand: D'oh! Could you please open an issue so that we don't forget?

bobsilverberg:
@willdurand This is ready for another review. I added some tests, dealt with the header, and added an installSource.

src/amo/reducers/recommendations.js
```diff
@@ -10,9 +10,14 @@ export const FETCH_RECOMMENDATIONS: 'FETCH_RECOMMENDATIONS'
...
   = 'LOAD_RECOMMENDATIONS';
+export const OUTCOME_CURATED: 'curated' = 'curated';
+export const OUTCOME_RECOMMENDED: 'recommended' = 'recommended';
+export const OUTCOME_RECOMMENDED_FALLBACK: 'recommended_fallback' = 'recommended_fallback';
 
 export type FallbackReasonType = 'no_results' | 'timeout';
-export type OutcomeType = 'curated' | 'recommended' | 'recommended_fallback';
+export type OutcomeType = typeof OUTCOME_CURATED |
+  typeof OUTCOME_RECOMMENDED |
+  typeof OUTCOME_RECOMMENDED_FALLBACK;
```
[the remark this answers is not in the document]
willdurand: I insist 😛

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,184 @@
...
+    let header = loading ? <LoadingText width={100} /> : null;
+    if (!header) {
+      header = outcome === OUTCOME_RECOMMENDED ?
+        i18n.gettext('Other users with this extension also installed') :
+        i18n.gettext('Other popular extensions');
+    }
+
+    return (
+      <AddonsCard
+        addonInstallSource={outcome || ''}
```
willdurand: Is it ok to have empty string as install source?

tests/unit/amo/helpers.js
```diff
@@ -195,6 +195,13 @@ export const fakeCategory = Object.freeze({
   weight: 1,
 });
 
+export const fakeRecommendations = Object.freeze({
+  addons: [fakeAddon, fakeAddon, fakeAddon, fakeAddon],
```
willdurand: Note: we can also write `Array(4).fill(fakeAddon)`

src/amo/components/AddonRecommendations/index.js
bobsilverberg: I believe so. Looking at the code that actually uses it, in [`SearchResult`](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/SearchResult/index.js#L144-L146), it simply checks `if (addonInstallSource)` and `''` is falsey, so it should yield a correct result.

Having said that, the only time the `outcome` will be missing is if we're rendering the loading version of the card, in which case there won't be anything to link anyway.

tests/unit/amo/helpers.js
bobsilverberg: Nice. :-)

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,184 @@
...
+      const { fallbackReason, loading, outcome } = newRecommendations;
+
+      if (loading) {
+        return;
+      }
+
+      invariant(newAddon, 'newAddon is required');
+      invariant(outcome, 'outcome is required');
+
+      // TODO: Determine the exact format for this output.
```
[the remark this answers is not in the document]
bobsilverberg: I opened #4957.

src/amo/components/AddonRecommendations/index.js
```diff
@@ -0,0 +1,184 @@
...
+    tracking: defaultTracking,
+  };
+
+  componentDidMount() {
+    const { cookie, randomizer, addon, recommendations } = this.props;
+
+    // Set a cohort for the experiment.
+    this.cohort = cookie.load(TAAR_COHORT_COOKIE_NAME);
+    if (this.cohort === undefined) {
+      this.cohort = randomizer() > 0.5 ?
```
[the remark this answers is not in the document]
bobsilverberg: Sure, sounds good.

tests/unit/amo/components/TestAddonRecommendations.js
```diff
@@ -0,0 +1,299 @@
...
+        const cookie = fakeCookie(undefined);
+        const randomizer = fakeRandomizer(expectedRecommendedValue);
+        render({ cookie, randomizer });
+        sinon.assert.calledWith(cookie.save,
+          TAAR_COHORT_COOKIE_NAME, expectedCohort, { path: '/' });
+      };
+
+    checkWithRandomizerValue(true, TAAR_COHORT_INCLUDED);
+    checkWithRandomizerValue(false, TAAR_COHORT_EXCLUDED);
+  });
```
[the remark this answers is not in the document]
bobsilverberg: I actually prefer the elimination of duplication that this provides, but if you don't like it I'll change it. It just seems unfortunate to duplicate all that code in a separate test which will really only differ by one line. But I do understand there's a value in simplicity and understandability.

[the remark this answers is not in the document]
bobsilverberg: Ok, having changed it it's not really so much code, it does look nicer. :-)

bobsilverberg:
Yeah, I saw that comment in the doc, but I wasn't sure what the resolution was. I'll leave a comment in the doc asking for clarification. I think this is still fine to land on dev only for now.

willdurand:
> Yeah, I saw that comment in the doc, but I wasn't sure what the resolution was. I'll leave a comment in the doc asking for clarification. I think this is still fine to land on dev only for now.

Yup, that is why I did not block the PR, :shipit:

src/amo/reducers/recommendations.js
```diff
@@ -15,7 +15,8 @@ export const OUTCOME_RECOMMENDED: 'recommended' = 'recommended';
 export const OUTCOME_RECOMMENDED_FALLBACK: 'recommended_fallback' = 'recommended_fallback';
 
 export type FallbackReasonType = 'no_results' | 'timeout';
-export type OutcomeType = typeof OUTCOME_CURATED |
+export type OutcomeType =
+  typeof OUTCOME_CURATED |
   typeof OUTCOME_RECOMMENDED |
```
willdurand: Can we have the `|` first?

bobsilverberg:
Thanks for all the thorough reviews and excellent suggestions @willdurand. :-)

PR #4943, opened 2018-05-07
Fix server error when navigating back to a collection page

### LINKED ISSUE #4933: Server error received when navigating with browser 'back' button from page 2 to page 1 in collection details
STR:

1. Open AMO and visit a collection with >25 add-ons (i.e. https://addons.mozilla.org/en-US/firefox/collections/davy-jones/optimizefirefox/)
2. Use the navigation buttons to go to the second page
3. Use browser's back button to navigate back to the first page

Actual result:
A 'server error' page is displayed when trying to navigate back to the first collection page 

Expected result:
First page of the visited collection should be displayed when using browser's 'back' button to navigate back

Notes:
- reproduces only when navigating back from page 2 to page 1
- reproduces only in collection details pages
- no specific error noticed in the console or in the network tab
- reproduced on all AMO servers with FF59, Win10x64

page 1 error

willdurand:
Fix #4933

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -136,7 +136,7 @@ export class CollectionBase extends React.Component<Props> {
     if (collection && addonsPageChanged) {
       this.props.dispatch(fetchCurrentCollectionPage({
         errorHandlerId: errorHandler.id,
-        page: location.query.page,
+        page: location.query.page || 1,
         slug: params.slug,
         user: params.user,
       }));
--- a/tests/unit/amo/components/TestCollection.js
+++ b/tests/unit/amo/components/TestCollection.js
@@ -348,6 +348,43 @@ describe(__filename, () => {
     }));
   });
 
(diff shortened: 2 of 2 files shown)


src/amo/components/Collection/index.js
```diff
@@ -136,7 +136,7 @@ export class CollectionBase extends React.Component<Props> {
     if (collection && addonsPageChanged) {
       this.props.dispatch(fetchCurrentCollectionPage({
         errorHandlerId: errorHandler.id,
-        page: location.query.page,
+        page: location.query.page || 1,
```
bobsilverberg: Would it make sense to default `page` to 1 in [`fetchCurrentCollectionPage`](https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/collections.js#L215) instead of dealing with this in the component?

willdurand: I thought about it too, but since `fetchCurrentCollectionPage()` implies to fetch a specific page, I thought it would be better to always require a `page`.

PR #4949, opened 2018-05-07
Remove sudo:required in travis-ci config

willdurand:
Let's see what happens...


willdurand:
This is actually invalid, we need this config for Docker: https://docs.travis-ci.com/user/docker/.

PR #4953, opened 2018-05-07
i 4367 empty review not updating fix

### LINKED ISSUE #4367: Edit review by deleting the previous submitted text does not work
STR:
1. Load AMO dev and go to an add-on's detail page.
2. Submit a review with text, preferably one that has No reviews yet.
3. Edit your review by deleting the previous text.
4. Go back to the add-on's detail page.
5. Observe the number of reviews mentioned in 'Read all <number> reviews' link.

Expected result:
The link shows No reviews yet.

Actual result:
Read 1 review is displayed and when clicked the review contains text.

Notes:
This is reproducible on AMO dev and AMO stage with FF58 (Win10).

edit review by deleting text

rebmullin:
fixes #4367 - updates the review whether textarea is empty or not.  It still saves as review regardless if text is added (this latter part still might need to be discussed, imo - unless i am missing something - it should still be a review. is there someone who i should reach out for this question?)

hey @bobsilverberg  - please take a look at this approach and let me know what you think : )

note - i wasn't 100% sure about the test update and if this was enough

### DIFF
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -100,7 +100,10 @@ export class AddonReviewBase extends React.Component<Props, State> {
     event.preventDefault();
     event.stopPropagation();
 
-    const newReviewParams = { body: reviewBody || undefined };
+    // If reviewBody text is null, we'll assign it as &nbsp;
+    // which will show as hidden in html and api will follow thru
+    // with the update since there is a value
+    const newReviewParams = { body: reviewBody || '&nbsp;' };
     const updatedReview = { ...review, ...newReviewParams };
 
     const params = {
@@ -218,7 +221,7 @@ export class AddonReviewBase extends React.Component<Props, State> {
               className="AddonReview-textarea"
               onInput={this.onBodyInput}
               name="review"
-              value={reviewBody}
+              value={reviewBody === '&nbsp;' ? '' : reviewBody}
               placeholder={placeholder}
             />
           </div>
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAddonReview.js
```diff
@@ -126,6 +126,29 @@ describe(__filename, () => {
       });
   });
 
+  it('a review without text feedback will show as no text in textarea and &nbsp; in review list display', () => {
```
bobsilverberg: We generally try to name tests so they start with a verb, so the "it ..." makes sense, like `it('can...')` and `it('should...')`.

src/amo/components/AddonReview/index.js
```diff
@@ -100,7 +100,10 @@ export class AddonReviewBase extends React.Component<Props, State> {
     event.preventDefault();
     event.stopPropagation();
 
-    const newReviewParams = { body: reviewBody || undefined };
+    // If reviewBody text is null, we'll assign it as &nbsp;
+    // which will show as hidden in html and api will follow thru
+    // with the update since there is a value
+    const newReviewParams = { body: reviewBody || '&nbsp;' };
```
bobsilverberg: I don't think we want to do this. We actually want to store an empty string, rather than `&nbsp;`. I think this is going to have to be dealt with in the saga or api file, rather than in this component. Can you take a look at doing that and see if you can make it work with an empty string?

rebmullin: @bobsilverberg - thanks for the feedback. I will look into more 🔍

bobsilverberg:
@rebmullin I added a comment to issue #4367 explaining that this is now blocked by https://github.com/mozilla/addons-server/issues/8226, so I think it makes sense to close this PR and issue a new one after https://github.com/mozilla/addons-server/issues/8226 is fixed.

Issue #4918, opened 2018-05-02
need final copy for placeholder text on collection forms screens


rebmullin:
Just need final copy on placeholder text on collection form screens (prior to adding addons):

![Alt text](https://monosnap.com/image/C2f305wUUzHSL0hjSZ95kOeh5e4EnI.png)

REF: original issue #4843

@meridelw  // cc @pwalm @bobsilverberg

bobsilverberg:
To document what was discussed in the meeting, there are two states:

1. Shown above, is when a user first opens a page to add a new collection. The collection does not yet exist, and they need to save it first before adding add-ons to it. They will need to provide at least the "Collection name" in order to save it.

2. Shown below, is what a user sees when editing a collection that does not have any add-ons in it. They should be prompted to add some add-ons to the collection, which they can do using the "Search" box in the form.

<img width="1193" alt="screenshot 2018-05-02 16 23 21" src="https://user-images.githubusercontent.com/142755/39547314-314cf7fa-4e25-11e8-8596-4948adbdc18a.png">

bobsilverberg:
From https://github.com/mozilla/addons-frontend/issues/4843#issuecomment-386111623, @MeridelW  said:

I'm not able to see how this actually behaves using the link. Should I be able to?

It's my understanding, based on Bob's comment above, that first the user must create and save the collection and THEN the user adds extensions/themes to it. Is that right? If so, I think the current lay-out may need adjustment:
-In particular, it seems like the "Save Collection" button should appear above the search-and-add function.
-Is is possible to make this more dynamic? See my suggestions below:

    When the user is creating the collection, before they have added anything to it, we would have the following copy appear in the grey space:

"First, create and save your collection. After that, you can add extensions and themes."

    Once the user hits Save, the following copy would appear (I am not sure of its placement visually):

"Add extensions and themes to your collection."

Philip - we may need to huddle up on this so I understand what's feasible!

bobsilverberg:
Thanks @MeridelW

> I'm not able to see how this actually behaves using the link. Should I be able to?

I believe you need to be logged in to see the screen at that [link](https://addons-dev.allizom.org/en-US/firefox/collections/add/). Were you logged in?

> It's my understanding, based on Bob's comment above, that first the user must create and save the collection and THEN the user adds extensions/themes to it. Is that right? If so, I think the current lay-out may need adjustment:
-In particular, it seems like the "Save Collection" button should appear above the search-and-add function. 

[...]

PR #4961, opened 2018-05-08
I4918 collection form placeholder text

rebmullin:
address/ fixes -#4918 - adds placeholder text in the following places:

viewing collection before add-ons are added:
Alt text

viewing collection in edit form and before any add-ons are added:
Alt text

this was original placeholder that was added (i am actually not sure how you navigate here):
Alt text

note - button will be addressed in another issue

Please let me know if this is ok or if anything is still TBD - thanks : )

cc @MeridelW @bobsilverberg @pwalm

### DIFF
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -19,6 +19,7 @@ type Props = {|
   className?: string,
   editing?: boolean,
   loading?: boolean,
+  placeholderText: string | null,
   // When loading, this is the number of placeholders
   // that will be rendered.
   placeholderCount: number,
@@ -39,6 +40,7 @@ export default class AddonsCard extends React.Component<Props> {
     editing: false,
     loading: false,
     placeholderCount: DEFAULT_API_PAGE_SIZE,
+    placeholderText: null,
   }
 
   render() {
@@ -53,6 +55,7 @@ export default class AddonsCard extends React.Component<Props> {
       showMetadata,
       showSummary,
       type,
+      placeholderText,
       ...otherProps
     } = this.props;
 
@@ -87,8 +90,13 @@ export default class AddonsCard extends React.Component<Props> {
       }
     }
 
+    const placeholder = placeholderText ?
+      <p className="AddonsCard-placeholder-text">{placeholderText}</p> : null;
+
     const allClassNames = makeClassName('AddonsCard', className,
-      type && `AddonsCard--${type}`);
+      type && `AddonsCard--${type}`,
+      placeholderText && !addonElements.length && 'AddonCard--placeholder');
+
     return (
       <CardList
         {...otherProps}
@@ -100,7 +108,8 @@ export default class AddonsCard extends React.Component<Props> {
           <ul className="AddonsCard-list">
             {addonElements}
           </ul>
-        ) : null}
+        ) : <React.Fragment>{placeholder}</React.Fragment>
+      }
       </CardList>
     );
   }
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -19,7 +19,6 @@ type Props = {|
   className?: string,
   editing?: boolean,
   loading?: boolean,
-  placeholderText: string | null,
   // When loading, this is the number of placeholders
   // that will be rendered.
   placeholderCount: number,
@@ -40,7 +39,6 @@ export default class AddonsCard extends React.Component<Props> {
     editing: false,
     loading: false,
     placeholderCount: DEFAULT_API_PAGE_SIZE,
-    placeholderText: null,
   }
 
   render() {
@@ -55,7 +53,6 @@ export default class AddonsCard extends React.Component<Props> {
       showMetadata,
       showSummary,
       type,
-      placeholderText,
       ...otherProps
     } = this.props;
 
@@ -90,13 +87,8 @@ export default class AddonsCard extends React.Component<Props> {
       }
     }
 
-    const placeholder = placeholderText ?
-      <p className="AddonsCard-placeholder-text">{placeholderText}</p> : null;
-
     const allClassNames = makeClassName('AddonsCard', className,
-      type && `AddonsCard--${type}`,
-      placeholderText && !addonElements.length && 'AddonCard--placeholder');
-
+      type && `AddonsCard--${type}`);
     return (
       <CardList
         {...otherProps}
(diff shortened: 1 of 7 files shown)


MeridelW:
@rebmullin: Thanks! 

_When the user is creating the collection, the text in the gray box should be:_
**First, create and save your collection. Then you can add extensions and themes.**

_Once the user creates the collection and the "Edit this collection" link screen appears, the text in the gray box should be:_
**Select "Edit this collection" to add extensions and themes.** 
[Question: is it the case that the user will always be adding  extensions/themes when they see this screen? Or will they also see this screen if they are removing extensions/ themes? If the latter, the text should instead be:] **Select "Edit this collection" to add or remove extensions and themes.** 

_When the user hits Save, this copy appears in the grey box:_
**Search for extensions and themes to add to your collection.**

This flow feels a bit clunky, though. Ideally when the user is first creating a collection, they would not get the screen for "Edit this collection" - instead they would remain on the first screen and be directed via the Save button changing and the text in the grey box to add extensions/ themes. Would changing this be within scope? @pwalm

bobsilverberg:
> _Once the user creates the collection and the "Edit this collection" link screen appears, the text in the gray box should be:_
**Select "Edit this collection" to add extensions and themes.** 
[Question: is it the case that the user will always be adding  extensions/themes when they see this screen? Or will they also see this screen if they are removing extensions/ themes? If the latter, the text should instead be:] **Select "Edit this collection" to add or remove extensions and themes.** 

I believe that the only time the user will see this screen is when they have a collection which doesn't contain any add-ons, which happens after they initially save a new collection, but could also happen if they visit the page for an existing collection which doesn't have any add-ons. I think the first text above makes sense in that case, and there will be no need for the second **Select "Edit this collection" to add or remove extensions and themes.** text.

> _When the user hits Save, this copy appears in the grey box:_
**Search for extensions and themes to add to your collection.**

> This flow feels a bit clunky, though. Ideally when the user is first creating a collection, they would not get the screen for "Edit this collection" - instead they would remain on the first screen and be directed via the Save button changing and the text in the grey box to add extensions/ themes. Would changing this be within scope? @pwalm 

I agree that this seems a bit clunky, and agree it would make sense to leave the user on the Edit screen so that they can immediately add add-ons to the collection after the initial save, rather than returning them to the Collection Details page from which they have to click the "Edit this collection" link to get back into editing mode to add add-ons. One issue here, though, is the question of confirming the fact that the collection was saved. Do you think simply changing the placeholder text on the right from **First, create and save your collection. Then you can add extensions and themes.** to **Search for extensions and themes to add to your collection.** is enough of an indication to the user that the collection was saved?

bobsilverberg:
@rebmullin I'm going to hold off on reviewing the code until the questions above are addressed.

bobsilverberg:
@rebmullin I just noticed a problem when I was checking something else. The placeholder text appears when a user is not logged in, but it should not. See the attached screenshot:

screenshot 2018-05-09 15 15 29

rebmullin:
@bobsilverberg  - i'll open a new issue for that so i can tackle that piece right away.. thanks

rebmullin:
@MeridelW  - disregard last question above. I see your feedback in #4978. I will use that as reference here..thanks : )

rebmullin:
@bobsilverberg  - I made the changes per feedback in #4978 (is it better to post those notes here too or reference link is enough?). 
I kind of feel like maybe this should wait till the flow piece is updated as the logic might have to be adjusted too.. or do you think this is ok first? and we can adjust this later if/when needed?

src/amo/components/CollectionList/index.js
```diff
@@ -55,6 +55,9 @@ export class CollectionListBase extends React.Component<Props> {
           </Card>
           {isLoggedIn &&
             <p className="CollectionList-placeholder">
+              {/* TODO: The collections flow is changing a bit (soon), so this
+              needs to be rechecked once that issue is complete.
+              This placeholder text might be unnessary once that update happens. */}
               {i18n.gettext(
```
bobsilverberg: This should be the "First, create your collection. Then you can add extensions and themes.", text which you originally defined in `Collection/index.js`, but really belongs here.

src/amo/components/AddonsCard/index.js
```diff
@@ -19,6 +19,7 @@ type Props = {|
   className?: string,
   editing?: boolean,
   loading?: boolean,
+  placeholderText: string | null,
```
bobsilverberg: I'm not sure it makes sense to add this to the `AddonsCard` at all. I think it makes more sense to just add it to the `Collection` and `CollectionList`.

src/amo/components/Collection/index.js
```diff
@@ -248,6 +255,7 @@ export class CollectionBase extends React.Component<Props> {
             addons={addons}
             editing={editing}
             loading={!collection || loading}
+            placeholderText={placeholderText}
```
bobsilverberg: Rather than pass a `placeholderText` to the `AddonsCard`, I think it would make more sense to just display the placeholder text here *instead* of an `AddonsCard`. If you want to encapsulate the logic/formatting of the placeholder text in a single place, I'd create a new component for it in `src/ui/components/`, but that might be overkill.

src/amo/components/Collection/index.js
```diff
@@ -249,6 +250,11 @@ export class CollectionBase extends React.Component<Props> {
             editing={editing}
             loading={!collection || loading}
           />
+          {!loading && addons.length === 0 &&
+            <p className="Collection-placeholder">{ i18n.gettext(
```
bobsilverberg: This formatting looks a bit odd. It might be better as:

```js
<p className="Collection-placeholder">
  {i18n.gettext('Search for extensions and themes to add to your collection.')}
</p>
```

tests/unit/amo/components/TestCollection.js
```diff
@@ -107,6 +107,12 @@ describe(__filename, () => {
     expect(wrapper.find(AddonsCard)).toHaveProp('loading', true);
   });
 
+  it('renders placeholder text if there are no add-ons', () => {
```
bobsilverberg: We might as well check the text of the placeholder as well.

tests/unit/amo/components/TestCollection.js
```diff
@@ -107,6 +107,12 @@ describe(__filename, () => {
...
+    const wrapper = renderComponent({ collection: null });
+
+    expect(wrapper.find('.Collection-placeholder')).toHaveLength(1);
+  });
+
```
bobsilverberg: We should also have a test that the placeholder does not appear when there are add-ons in the collection.

rebmullin:
thanks @bobsilverberg : )

PR #4962, opened 2018-05-08
Remove stylefmt from dependencies to eliminate nsp-check failure

### LINKED ISSUE #4960: Remove stylefmt from dependencies to eliminate nsp-check failure
Travis' nsp-check is [failing](https://travis-ci.org/mozilla/addons-frontend/jobs/376447158) because of a component which is a dependency of `stylefmt`, which we appear not to use. Removing `stylefmt` as a dependency *should* resolve this issue.

bobsilverberg:
Fixes #4960

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,6 @@
     "redux-saga": "0.16.0",
     "response-time": "2.3.2",
     "serialize-javascript": "1.5.0",
-    "stylefmt": "6.0.0",
     "ua-parser-js": "0.7.18",
     "url": "0.11.0",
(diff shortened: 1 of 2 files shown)


bobsilverberg:
This seems to have resolved the nsp-check failure on Travis.

willdurand:
mmh it seems it is not failing anymore..

bobsilverberg:
> mmh it seems it is not failing anymore..

That's odd. What the heck was going on? I suppose it's still worth merging this, as we don't need the package. Do you agree?

Issue #4951, opened 2018-05-07
The select box to add an add-on to a collection should be hidden until the collection is saved


bobsilverberg:
On the Add a Collection screen (https://addons-dev.allizom.org/en-US/firefox/collections/add/) we should hide the select box that is used to add an add-on to the collection before the collection is saved. It serves no purpose when creating a collection.

xu3u4:
Hi may I take this issue?

willdurand:
Hi @xu3u4, yes of course! The "select box" is the autocomplete search for add-ons.

PR #4966, opened 2018-05-09
i-4951 hide search addon field before collection saved

xu3u4:
Fixes https://github.com/mozilla/addons-frontend/issues/4951

Because only if a collection is saved, it is able to search add-ons. Thus, this PR hide the add-on search field when a collection is creating.

On adding a collection:
image

On editing a collection:
image

### DIFF
--- a/tests/unit/amo/components/TestCollectionManager.js
+++ b/tests/unit/amo/components/TestCollectionManager.js
@@ -251,6 +251,12 @@ describe(__filename, () => {
     expect(root.find('#collectionSlug')).toHaveProp('value', 'new-slug');
   });
 
+  it('hides search add-on select when creating a collection', () => {
+    const root = render({ collection: null, creating: true });
+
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -251,6 +251,12 @@ describe(__filename, () => {
     expect(root.find('#collectionSlug')).toHaveProp('value', 'new-slug');
   });
 
+  it('has no search add-on selection when creating selection', () => {
```
bobsilverberg: This should say `hides search add-on select when creating a collection`.

xu3u4: Ok, thanks!
It looks better!

xu3u4:
@bobsilverberg I have updated. Please check.
Thanks!

PR #4974, opened 2018-05-10
Only display add-on recommendations for extensions

### LINKED ISSUE #4968: Only display add-on recommendations for extensions
The `AddonRecommendations` component that is displayed for an `Addon` should only be displayed if the add-on is an extension. Currently it displays for all add-on types, but that should be fixed.

bobsilverberg:
Fixes #4968

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -467,6 +467,9 @@ export class AddonBase extends React.Component {
     })}`;
 
     const isFireFox = compatibility && compatibility.reason !== INCOMPATIBLE_NOT_FIREFOX;
+    const enableAddonRecommendations =
+      config.get('enableAddonRecommendations') &&
+      addonType === ADDON_TYPE_EXTENSION;
 
     return (
       <div
@@ -552,9 +555,8 @@ export class AddonBase extends React.Component {
 
             {this.renderShowMoreCard()}
 
-            {config.get('enableAddonRecommendations') &&
-             addonType === ADDON_TYPE_EXTENSION && (
-               <AddonRecommendations addon={addon} />
+            {enableAddonRecommendations && (
+              <AddonRecommendations addon={addon} />
             )}
           </div>
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -552,8 +552,9 @@ export class AddonBase extends React.Component {
 
             {this.renderShowMoreCard()}
 
-            {config.get('enableAddonRecommendations') && (
-              <AddonRecommendations addon={addon} />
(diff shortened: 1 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -552,8 +552,9 @@ export class AddonBase extends React.Component {
 
             {this.renderShowMoreCard()}
 
-            {config.get('enableAddonRecommendations') && (
-              <AddonRecommendations addon={addon} />
+            {config.get('enableAddonRecommendations') &&
+             addonType === ADDON_TYPE_EXTENSION && (
```
willdurand: Can we extract these conditions into a new variable above? `enableAddonsRecommendations`?

tests/unit/amo/components/TestAddon.js
```diff
@@ -961,6 +961,24 @@ describe(__filename, () => {
...
 
+  it('does not render recommendations if the add-on is not an extension', () => {
+    const fakeConfig = getFakeConfig({ enableAddonRecommendations: true });
+    for (const addonType of [
+      ADDON_TYPE_COMPLETE_THEME,
+      ADDON_TYPE_DICT,
+      ADDON_TYPE_LANG,
+      ADDON_TYPE_OPENSEARCH,
+      ADDON_TYPE_THEME,
+    ]) {
```
willdurand: I tend to write such loops in a slightly different manner to make the "for each" part more explicit (see below), but I guess I am fine with this way too!

```
[
  ADDON_TYPE_COMPLETE_THEME,
  ...
].forEach((addonType) => {
});
```

bobsilverberg: Tbh, I prefer the `for... of` syntax over `forEach`, and, if it matters at all, it is apparently more performant. I've read that a number of places including [here](https://coderwall.com/p/kvzbpa/don-t-use-array-foreach-use-for-instead). So I would like to keep this code as is, which you say you're fine with too.

src/amo/components/Addon/index.js
bobsilverberg: I can, but I didn't because it's only used once, so I didn't see the point in creating a variable for it. Is it just to improve readability? I'm fine doing it, and I will, but I just want to understand why you think it's important.

willdurand: I think it would be slightly more readable, and if we open the door to other addon types, that will be much better.

Also, I think we had an issue once with multiple conditions chained with `&&` (rendering `0` instead of not rendering anything).

bobsilverberg: Oh yeah, we did. I think that was because one of the conditions returned a `0` instead of a boolean, but yes, this is a good thing to change. 👍

PR #4975, opened 2018-05-10
Set a custom dimension based on the TAAR Lite cohort of the user

### LINKED ISSUE #4957: Set a custom dimension based on the TAAR Lite cohort of the user
We are currently sending a single ping to Google Analytics (GA) for each impression of the AddonRecommendations card. We are sending the `outcome`, which will indicate what was requested from the API and what was received, and appending the `fallbackReason` to that with a dash if there was a fallback reason. We are also sending the name of the add-on in question.

We should confirm whether this is the data that is desired, and if this is the correct format.

@muffinresearch Do you want to run with this, or would you like me to take it?

bobsilverberg:
Fixes #4957

### DIFF
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -28,6 +28,7 @@ import type { DispatchFunc } from 'core/types/redux';
 
 export const TAAR_IMPRESSION_CATEGORY = 'AMO Addon / Recommendations Shown';
 export const TAAR_COHORT_COOKIE_NAME = 'taar_cohort';
+export const TAAR_COHORT_DIMENSION = 'dimension4';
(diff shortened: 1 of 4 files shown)


src/amo/components/AddonRecommendations/index.js
```diff
@@ -67,6 +68,12 @@ export class AddonRecommendationsBase extends React.Component<Props> {
       cookie.save(TAAR_COHORT_COOKIE_NAME, this.cohort, { path: '/' });
     }
 
+    // Set a custom dimension for the cohort.
```
willdurand: Not sure this comment is super useful since the code is pretty straightforward. Maybe we would like to know why we do this instead, what do you think? (or no comment at all)

bobsilverberg: Fair enough. I think I'll just remove it.

PR #4976, opened 2018-05-10
Css fix for add ons (extensions) detail page

### LINKED ISSUE #4878: The add-on and author's name are too close
STR:
1. Submit an add-on that contains "y" "q" "p" "g" in it's name
2. Go to the add-ons detail page
3. Observe the Add-on details section

Expected result:
It's a reasonable space between the Add-on's Name and Author.

Actual result:
The two names are too close, they look jammed in.

Notes:
This issue is reproducible on AMO dev and AMO stage with FF59 (Win10).

I'll give below two examples to make it more clear
too close

and another one 
another example

biskit1:
Fixes #4878 

This PR introduces css changes to the details page of individual add ons. 

The original issue was opened to fix the line containing the author, which at times would overlap with the title (depending on the letters contained in the title, such as `g, y, p, q` and the size of screen) like so: 
<p align="center">
  
</p><br/>

To fix this problem, I just added a `line-height` of `1.5` to the `Addon-author` class:
<p align="center">
  
</p><br/>

While implementing this change and testing the results, I also noticed that at times the title itself ran into the same problem: 
<p align="center">
  
</p>
and another example: 
<p align="center">
  
</p><br/>

So I added a `line-height` property to the `Addon-title` class as well, with a value of `normal` (1.5 was too much for the title): 
<p align="center">
  
</p><br/>

There were no tests found for the files I modified, so I ran `yarn lint` and corrected any formatting errors.

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -72,7 +72,7 @@
   margin: 0 0 $padding-page-l;
   width: 100%;
   word-wrap: break-word;
-  
+
   .Addon-persona & {
     margin: $padding-page-l 0;
   }
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -70,8 +70,9 @@
   grid-column: 1 / span 2;
   margin: 0 0 $padding-page-l;
   width: 100%;
-  word-wrap: break-word;
   line-height: normal;
+  word-wrap: break-word;
+  
 
   .Addon-persona & {
     margin: $padding-page-l 0;
@@ -86,8 +87,9 @@
   @include font-light();
 
   display: block;
-  word-wrap: break-word;
   line-height: 1.5;
+  word-wrap: break-word;
+  
 
   &,
   a,
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -86,6 +86,7 @@
 
   display: block;
   word-wrap: break-word;
+  line-height: 1.5;
 
   &,
   a,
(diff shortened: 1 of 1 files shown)


rebmullin:
Thanks for this @biskit1 : ) Regarding your changes, I still show your line height for `Addon-title` as 1.5 tho.. per your comment above it seems like you meant to change that..?  

I agree that the title's line height could be improved a bit (in case there are 2 lines) but I think `normal` might be a bit too much - what do you think about something a little smaller (but bigger than what it currently is at :)?

biskit1:
Thanks for the feedback @rebmullin! How is `1.3` for `Addon-author`? 
I don't know why I automatically jumped to .5, but I agree it was a bit much.

rebmullin:
that’s a really good point @biskit1 - that we need to consider longer usernames. I am not sure how common those are but it still might be a (edge) case to consider.  Regarding the padding…I think that might be a good solution to try out..we did increase the parent line height to 1.1 and since `Addon-author` will inherit that maybe this will look okay with just the extra padding. Do you want to try that and see it how looks at the different breakpoints?

biskit1:
@willdurand yes. I drafted a response a while ago and I’ve been meaning to post and update the PR. will post this afternoon. Apologies for the delay. 
Thanks for following up on this!

biskit1:
@rebmullin I think it's good with the `padding-top: 3px` for `Addon-author`. 
A couple of breakpoints as an example: 

image

image

Another option is to **just** have `line-height: normal` on the `Addon-title`. I found it nicely spaced the title and author line at the various breakpoints, but I think you might find it to be a little bit too for when the title line wraps: 

image

src/amo/components/Addon/styles.scss
```diff
@@ -85,6 +86,7 @@
   @include font-light();
 
   display: block;
+  padding-top: 3px;
```
willdurand: what is this padding for?

PR #4989, opened 2018-05-14
I4947-addonrecommendations-style-update

### LINKED ISSUE #4947: AddonRecommendations card does not have enough space at the bottom when loading
This may be difficult to reproduce depending on how fast the recommendations are loaded, but here's an STR:

1. Go to the details page for an add-on.
2. Observe the `AddonRecommendations` card which appears below the "About this extension" card on the right-hand side.
3. Try to observe it when it is in its loading state. You can re-trigger this by doing a hard refresh on the page.

Expected:

The card looks good, with proper spacing.

Actual:

There is not enough space at the bottom of the card.

Example:

screen shot 2018-05-07 at 10 49 55

rebmullin:
fixes #4947 - addresses a few style issues for addon recommendations noted in this issue

many user style fix:
Alt text

Loading style fix:
Alt text

Also - on a separate but related issue:

I see this for iPad / ~768px - 1024 sizes:
Alt text
Alt text

basically b/c the device is smaller but column is still on the right there is not a enough room really. I am wondering if could make this look more like the mobile version or if we could maybe go to 3 columns (each col approx ~ 33%) and adjust the padding slightly so it’s not so squished. I just wasn't sure - should I open a new issue for this (assuming there isn’t one open already for this (I’ll check)), or just ask about it here?

@bobsilverberg

### DIFF
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -26,6 +26,8 @@ import type { I18nType } from 'core/types/i18n';
 import type { AddonType } from 'core/types/addons';
 import type { DispatchFunc } from 'core/types/redux';
 
+import './styles.scss';
+
 export const TAAR_IMPRESSION_CATEGORY = 'AMO Addon / Recommendations Shown';
 export const TAAR_COHORT_COOKIE_NAME = 'taar_cohort';
 export const TAAR_COHORT_INCLUDED: 'TAAR_COHORT_INCLUDED'
--- a/src/amo/components/AddonRecommendations/styles.scss
+++ b/src/amo/components/AddonRecommendations/styles.scss
@@ -1,0 +1,52 @@
+@import "~core/css/inc/mixins";
+@import "~amo/css/inc/vars";
+@import "~ui/css/vars";
+
+.AddonRecommendations {
+
+  // stylelint-disable max-nesting-depth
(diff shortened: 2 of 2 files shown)


rebmullin:
cool - thanks @bobsilverberg  - will do.

hey @pwalm - can you take a look ~ ipad sizes? This is what we are seeing on the frontend:
Alt text
Alt text

It's hard to tell here  b/c there are only 3 items but it's spaced for 4 items (so each item is approx 25%). It looks really good at larger breakpoints but they begin to look a little squished at the medium breakpoints. B/c space is less I propose we make it look either like the stacked mobile version:
Alt text

OR maybe we do 2 or 3 items instead of 4 for medium breakpoints..
here is 3 items example:
Alt text

and 2 items example:
Alt text

we could also reduce padding a bit on the medium breakpoints but I think we still might run into similar issue at least for 768px breakpoint.

let me know if you need to see  more or if you have any questions. thanks!

bobsilverberg:
Thanks for adding all that detail @rebmullin. I think it might have been better to open another issue and put that information there, because this really relates to a different issue, and not what this PR is trying to address. This PR is good to merge and close, but there is still a different issue, which is what you described above.

rebmullin:
@bobsilverberg  - sorry i misread your comment above. will reopen a new issue for latter info described. thanks

PR #4992, opened 2018-05-14
I4852-fetch logic adjustment on addons by authors card

rebmullin:
fixes mozilla/addons#11684  - fixes item count on addonsByAuthorCard component when slug is updated

b/c the component was remounting the logic in componentWillReceiveProps was never different so the refetching logic wasn't hitting here. I adjusted logic to (re)fetch always on mount

### DIFF
--- a/tests/unit/amo/components/TestAddonsByAuthorsCard.js
+++ b/tests/unit/amo/components/TestAddonsByAuthorsCard.js
@@ -219,12 +219,9 @@ describe(__filename, () => {
     expect(root).toHaveClassName('AddonsByAuthorsCard');
   });
 
-  // Since it doesn't go through the logic in componentWillReceiveProps
-  // (of comparing changing props values) when it's remounted  - which is
-  // what happens when it doesn't go back to server in some scenarios - we
-  // we want to always make sure to do a fetch on mount to make sure
+  // We want to always make sure to do a fetch to make sure
   // we have the latest addons list.
-  // REF: https://github.com/mozilla/addons-frontend/issues/4852
+  // See: https://github.com/mozilla/addons-frontend/issues/4852
   it('should always fetch addons by authors', () => {
     const { store } = dispatchClientMetadata();
     const dispatchSpy = sinon.spy(store, 'dispatch');


src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -55,43 +54,13 @@ export class AddonsByAuthorsCardBase extends React.Component<Props> {
...
   componentWillMount() {
-    const { addons, addonType, authorUsernames, forAddonSlug } = this.props;
+    const { addonType, authorUsernames, dispatch, errorHandler, forAddonSlug } = this.props;
 
-    if (!addons) {
-      this.dispatchFetchAddonsByAuthors({ addonType, authorUsernames, forAddonSlug });
-    }
-  }
-
-  componentWillReceiveProps({
```
bobsilverberg: I haven't run the code yet, but at first glance this doesn't look right. You've removed a bunch of code, and the associated tests, which suggests to me that those use cases are no longer supported. Why is that ok? Have you tested the cases for which you've removed code to make sure they still work?

Please let me know your response to this and then I'll proceed with the rest of the review.

rebmullin: Hey @bobsilverberg - sorry I wasn’t too clear above. Basically tho this component (`AddonsByAuthorsCard`) gets remounted so when it runs through componentWillReceiveProps, the `nextProps` are never different from the `this.props`. This means the fetch logic located within this conditional - which compares old and new props - never gets reached. This is why it wasn't refetching the data..

I see the code was extracted out recently from Addons ([code update](https://github.com/mozilla/addons-frontend/commit/21a62088218a91f8f3c5527f451903fccf887202#diff-4953c6065645080e8ab07f343314c395))  and checking old vs newProps worked there but again b/c of remounting it doesn't when moved into its own component. 

It’s a big component but I can look into why Addons is remounting this component. I know I can prevent with shouldComponentMount but not sure if will give us what we need either (or its ramifications here). Anyway maybe we can discuss more tomorrow on a call to sort this out and make sure I move forward with the best approach? :)

Regarding tests...I thought I was removing the tests on this check and moving into the mount test. I always keep `yarn test` running to make sure nothing I didn’t break anything altho maybe it sounds like maybe I need to do more..

willdurand: So I don't think we can do this because if you use the browser "back" button between different addon detail pages, `componentWillReceiveProps()` will definitely be called and we should be using it.

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -55,43 +54,13 @@ export class AddonsByAuthorsCardBase extends React.Component<Props> {
   }
 
   componentWillMount() {
```
willdurand: I think it would maybe make sense to always call `dispatchFetchAddonsByAuthors()` in `componentWillMount()`.

src/amo/components/AddonsByAuthorsCard/index.js
rebmullin: Ahh..that's something i didn't see/test. thanks for catching that. I will add that back in here and return those tests

rebmullin:
hey @willdurand  - thanks for the feedback. I've made some updates.  Can you please take another look?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -232,6 +232,28 @@ describe(__filename, () => {
     }));
   });
 
+  it('dispatches fetchAddonsByAuthors on mount', () => {
```
willdurand: So I think we already have such a test case. What we would like to have now is a test case that fails when you revert your change in `src/amo/components/AddonsByAuthorsCard/index.js`, i.e. trying to render the component with a list of add-ons.

rebmullin: thanks for feedback @willdurand! I just need a little bit of clarification : ) Initially it was only fetching on mount if there were NO add-ons. So I am thinking maybe the following test should be removed:
```javascript
  it('should dispatch a fetch action if no add-ons found', () => {
    const dispatchSpy = sinon.spy(store, 'dispatch');

    render({
      addonType: ADDON_TYPE_EXTENSION,
      authorUsernames: ['test2'],
      errorHandler,
      store,
    });

    sinon.assert.calledWith(dispatchSpy, fetchAddonsByAuthors({
      addonType: ADDON_TYPE_EXTENSION,
      authorUsernames: ['test2'],
    }));
  });
```
altho it wasn't failing it's just not need now b/c regardless of amount of add-ons it's going to do this fetch on mount - which it does on initial load and when you move around on client side (without back and forward buttons)...so it does this not only when there are no add-ons. 

When i revert my changes (for testing purposes), i don't see anything failing via tests. But i guess i should add a test to make sure one of the items doesn't have a matching slug to the current slug and make sure that all is correct. Is that what you mean?

willdurand: Right, this test case should not be kept or maybe it should be updated to "should always fetch addons by authors". That would also cover your change, because if you revert your change in the component, this test will fail.

rebmullin: hmm it;s not failing as expected here when i revert the component so i need to look into this a bit more :/

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -213,7 +213,7 @@ describe(__filename, () => {
     expect(root).toHaveClassName('AddonsByAuthorsCard');
   });
 
-  it('should dispatch a fetch action if no add-ons found', () => {
+  it('should always fetch addons by authors', () => {
```
willdurand: I think we should have a comment to not forget about why we want this (with a link to the issue).

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -225,6 +225,7 @@ describe(__filename, () => {
       store,
     });
 
+    sinon.assert.callCount(dispatchSpy, 1);
```
willdurand: I think you dont need this since you have `calledWith` below.

rebmullin:
i * think * it should be ok now : )

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -219,7 +219,13 @@ describe(__filename, () => {
     expect(root).toHaveClassName('AddonsByAuthorsCard');
   });
 
-  it('should dispatch a fetch action if no add-ons found', () => {
+  // Since it doesn't go through the logic in componentWillReceiveProps
+  // (of comparing changing props values) when it's remounted  - which is
+  // what happens when it doesn't go back to server in some scenarios - we
+  // we want to always make sure to do a fetch on mount to make sure
+  // we have the latest addons list.
+  // REF: https://github.com/mozilla/addons-frontend/issues/4852
```
willdurand: I am not sure I really understand that comment to be honest. Could you make it shorter and focused on the issue we want to solve?

In addition, we often use `See:` to link to an issue.

rebmullin:
hey @willdurand  - i modified the comment a bit. Do you mind taking another look to make sure it's better/okay now?

PR #5006, opened 2018-05-15
Add new UserProfileEditPicture component

### LINKED ISSUE #4276: Create a user profile photo upload component
We need a user profile upload component that accepts a user's photo and sends it to the API as part of the edit profile form.

It should look like:

screenshot 2018-01-31 14 25 04

willdurand:
Fix #4276 – ~⚠️ Depends on #5013 and #5014~

---

## Features

- [x] upload a profile picture
- [x] delete a profile picture

### DIFF
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -1,0 +1,72 @@
+/* @flow */
+import * as React from 'react';
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+import Button from 'ui/components/Button';
+import UserAvatar from 'ui/components/UserAvatar';
+import type { UserType } from 'amo/reducers/users';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+
+type Props = {|
+  i18n: I18nType,
+  name: string,
+  onDelete: Function,
+  onSelect: Function,
+  preview: string | null,
+  user: UserType | null,
+|};
+
+export const UserProfileEditPictureBase = ({
+  i18n,
+  name,
+  onDelete,
+  onSelect,
+  preview,
+  user,
+}: Props) => {
+  const altText = user ? i18n.sprintf(
+    i18n.gettext('Profile picture for %(name)s'), { name: user.name }
+  ) : null;
+
+  return (
+    <section className="UserProfileEdit-picture">
+      <label className="UserProfileEdit--label" htmlFor={name}>
+        {i18n.gettext('Profile photo')}
+      </label>
+
+      <UserAvatar
+        altText={altText}
+        preview={preview}
+        user={user}
+      />
+
+      {/* eslint-disable-next-line jsx-a11y/label-has-for */}
+      <label className="FileInput">
+        <input
+          className="FileInput-input"
+          disabled={!user}
+          name={name}
+          onChange={onSelect}
+          type="file"
+        />
+        <span className="Button Button--action Button--puffy">
+          {i18n.gettext('Choose photo...')}
+        </span>
+      </label>
+
+      {(user && user.picture_url) && (
+        <Button onClick={onDelete}>
+          {i18n.gettext('Delete this picture')}
+        </Button>
+      )}
+    </section>
+  );
+};
+
+export default compose(
+  translate(),
+)(UserProfileEditPictureBase);
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import makeClassName from 'classnames';
 import * as React from 'react';
 import { compose } from 'redux';
 
@@ -32,8 +33,16 @@ export const UserProfileEditPictureBase = ({
     i18n.gettext('Profile picture for %(name)s'), { name: user.name }
   ) : null;
 
+  const buttonClass = makeClassName(
+    'UserProfileEditPicture-select-button',
+    'Button Button--action Button--puffy',
+    {
+      'Button--disabled': !user,
+    }
+  );
+
   return (
-    <section className="UserProfileEdit-picture">
+    <section className="UserProfileEditPicture">
       <label className="UserProfileEdit--label" htmlFor={name}>
         {i18n.gettext('Profile photo')}
       </label>
@@ -45,21 +54,24 @@ export const UserProfileEditPictureBase = ({
       />
 
       {/* eslint-disable-next-line jsx-a11y/label-has-for */}
-      <label className="FileInput">
+      <label className="UserProfileEditPicture-file">
         <input
-          className="FileInput-input"
+          className="UserProfileEditPicture-file-input"
           disabled={!user}
           name={name}
           onChange={onSelect}
           type="file"
         />
-        <span className="Button Button--action Button--puffy">
+        <span className={buttonClass}>
           {i18n.gettext('Choose photo...')}
         </span>
       </label>
 
       {(user && user.picture_url) && (
-        <Button onClick={onDelete}>
+        <Button
+          className="UserProfileEditPicture-delete-button"
+          onClick={onDelete}
+        >
           {i18n.gettext('Delete this picture')}
         </Button>
       )}
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -68,12 +68,14 @@ export const UserProfileEditPictureBase = ({
       </label>
 
       {(user && user.picture_url) && (
-        <Button
-          className="UserProfileEditPicture-delete-button"
-          onClick={onDelete}
-        >
-          {i18n.gettext('Delete this picture')}
-        </Button>
+        <p>
+          <Button
+            className="UserProfileEditPicture-delete-button"
+            onClick={onDelete}
+          >
+            {i18n.gettext('Delete this picture')}
+          </Button>
+        </p>
       )}
     </section>
   );
--- a/src/amo/sagas/users.js
+++ b/src/amo/sagas/users.js
@@ -1,5 +1,6 @@
 import { call, put, select, takeLatest } from 'redux-saga/effects';
 import {
+  DELETE_USER_PICTURE,
   EDIT_USER_ACCOUNT,
   FETCH_USER_ACCOUNT,
   finishEditUserAccount,
@@ -8,6 +9,7 @@ import {
 } from 'amo/reducers/users';
 import {
   currentUserAccount as currentUserAccountApi,
+  deleteUserPicture as deleteUserPictureApi,
   editUserAccount as editUserAccountApi,
   userAccount as userAccountApi,
 } from 'amo/api/users';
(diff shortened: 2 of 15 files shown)


willdurand:
So I guess it is almost ready for a first look @bobsilverberg but there are two PRs before that could be reviewed/merged to reduce the diff here.

In addition, there is also a server issue that makes the feature not entirely finished (see: https://github.com/mozilla/addons-server/issues/8265). I think the only impact would be CSS-related (minor).

I could maybe cut this PR in half to reduce the diff even more!

willdurand:
To display the user profile picture (since it does not work because of https://github.com/mozilla/addons-server/issues/8265), I use this patch:

```diff
diff --git a/src/ui/components/UserAvatar/index.js b/src/ui/components/UserAvatar/index.js
index cc28a0b27..f90b15fe8 100644
--- a/src/ui/components/UserAvatar/index.js
+++ b/src/ui/components/UserAvatar/index.js
@@ -20,7 +20,7 @@ const UserAvatar = ({ altText, className, preview, user }: Props) => {

   if (user && !image) {
     image = user.picture_type && user.picture_type.length && user.picture_url
-      && user.picture_url.length ? user.picture_url : null;
+      && user.picture_url.length ? user.picture_url.replace('.png', '_original.png') : null;
   }

   return (

```

willdurand:
I have updated the Gif, and fixed the CSS for the delete button. This is good for a review!

tests/unit/amo/api/test_users.js
```diff
@@ -134,4 +135,33 @@ describe(__filename, () => {
       mockApi.verify();
     });
   });
+
+  describe('deleteUserPicture', () => {
+    const mockResponse = () => createApiResponse({
```
bobsilverberg: Why is this a function rather than just a variable? Also, I know you've only added the one test, but this `mockResponse` function is used a number of times in this file, so maybe it would be worthwhile to either move it to the top of the file, or have a `createFakeUserAccountResponse` in one of the helper files?

tests/unit/amo/api/test_users.js
```diff
@@ -134,4 +135,33 @@ describe(__filename, () => {
...
+      jsonData: createUserAccountResponse(),
+    });
+
+    const getParams = (params = {}) => {
```
bobsilverberg: I think I've brought this up before, but it seems like we're doing a bunch of extra stuff here having a `getParams` function, which can even take an extra set of `params` to override the defaults, but then all we ever do with it is just call `getParams()`. I know it's how we usually do it, but I'm starting to question the value of all that extra code when we never really use it, and could just create the params inline. This is the only test that even has access to `getParams`, and it seems unlikely we'll add any others.

src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -0,0 +1,86 @@
...
+          type="file"
+        />
+        <span className={buttonClass}>
+          {i18n.gettext('Choose photo...')}
+        </span>
+      </label>
+
+      {(user && user.picture_url) && (
+        <p>
+          <Button
```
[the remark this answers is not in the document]
willdurand: Yes, it is intended.

src/amo/sagas/users.js
```diff
@@ -90,8 +92,34 @@ export function* fetchUserAccount({
   }
 }
 
+export function* deleteUserPicture({
+  payload: {
+    errorHandlerId,
+    userId,
+  },
+}) {
```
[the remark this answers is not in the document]
willdurand: Yes, definitely: https://github.com/mozilla/addons-frontend/issues/5080

src/ui/components/UserAvatar/index.js
```diff
@@ -7,20 +7,33 @@ import type { UserType } from 'amo/reducers/users';
...
 |};
 
-const UserAvatar = ({ className, user }: Props) => {
+const UserAvatar = ({ altText, className, preview, user }: Props) => {
   const _className = makeClassName('UserAvatar', className);
 
+  let image = preview && preview.length ? preview : null;
+
+  if (user && !image) {
+    image = user.picture_type && user.picture_type.length && user.picture_url
```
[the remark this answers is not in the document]
willdurand: I don't know, that was there 🤷‍♂️ I think that's just the way of determining whether there is a profile picture for a given user or not.

tests/unit/amo/api/test_users.js
willdurand: We could indeed simplify a bit here.

tests/unit/amo/api/test_users.js
willdurand: mmh right, we could probably move this function to the top of the file.

src/ui/components/UserAvatar/index.js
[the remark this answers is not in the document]
bobsilverberg: Maybe it's worth investigating further, or maybe @eviljeff knows? I don't love having mysterious code that we don't fully understand in our code base.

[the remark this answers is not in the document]
eviljeff: `picture_type` is the mime type of the image [[docs](http://addons-server.readthedocs.io/en/latest/topics/api/accounts.html#account)]
Previously `picture_url` would be a url to a default avatar if there wasn't a user provided picture, so you had to test `picture_type` - which would be `null`/empty if there wasn't a user provided picture - to check if it was set or not.
Since mozilla/addons-server#7679 `picture_url` has been `null` when not set so I don't believe you need that check.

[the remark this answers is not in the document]
willdurand: Yup, let's forget about `picture_type` then.

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
+import * as React from 'react';
```
bobsilverberg: This file is called `TestUserProfilePictureEdit.js` but the component is called `UserProfileEditPicture`.

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
...
+import {
+  fakeI18n,
+  shallowUntilTarget,
+} from 'tests/unit/helpers';
+
+
+describe(__filename, () => {
+  const render = ({ i18n = fakeI18n(), ...props } = {}) => {
+    return shallowUntilTarget(
+      <UserProfileEditPicture i18n={i18n} name="input-name" {...props} />,
```
bobsilverberg: This component has some mandatory props that are not being passed. I guess the tests don't care, and I honestly don't recall if we usually worry about that or not, but I wonder if we should have default props for everything that is mandatory?

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
...
+    const name = 'some-input-name';
+    const root = render({ name });
+
+    expect(root.find('.UserProfileEditPicture')).toHaveLength(1);
+    expect(root.find('.UserProfileEdit--label'))
+      .toHaveText('Profile photo');
+    expect(root.find('.UserProfileEditPicture-file')).toHaveLength(1);
+    expect(root.find('.UserProfileEditPicture-file-input'))
+      .toHaveProp('name', name);
+    expect(root.find('.UserProfileEditPicture-delete-button')).toHaveLength(0);
```
bobsilverberg: Further to my comment above, it seems like we're relying on a non-explicit default state for the component in these tests. For example here we expect there to be no delete button, but that's only because of a specific state, which we didn't set up in the test. It seems like it would be better to set up the component state in the test (by passing explicit props).

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  it('renders a UserAvatar component', () => {
+    const root = render();
+
+    expect(root.find(UserAvatar)).toHaveLength(1);
+    expect(root.find(UserAvatar)).toHaveProp('altText', null);
```
bobsilverberg: Same here, the test doesn't explicitly set itself up for `altText` to be null.

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  it('specifies the alt text of the UserAvatar component when a user is passed', () => {
+    const { state } = dispatchSignInActions();
+    const user = getCurrentUser(state.users);
+
+    const root = render({ user });
+
+    expect(root.find(UserAvatar))
+      .toHaveProp('altText', `Profile picture for ${user.name}`);
```
bobsilverberg: We might as well also test that the `UserAvatar` component received the expected `preview` and `user` props as well.

willdurand:
@bobsilverberg thanks for the first batch of comments, I have updated the code, fixed a few things and rebased. You can have a fresh look if you wish.

tests/unit/amo/components/TestUserProfilePictureEdit.js
```diff
@@ -0,0 +1,133 @@
...
+    const { state } = dispatchSignInActions({
+      userProps: {
+        picture_url: 'https://example.org/pp.png',
+      },
+    });
+    const user = getCurrentUser(state.users);
+
+    const root = render({ user });
+
+    expect(root.find('.UserProfileEditPicture-delete-button')).toHaveLength(1);
```
bobsilverberg: Is it worth checking the text of the button as well?

tests/unit/ui/components/TestUserAvatar.js
```diff
@@ -62,4 +61,12 @@ describe(__filename, () => {
...
   });
+
+  it('renders a preview image when supplied', () => {
+    const preview = 'https://example.org/image.jpg';
+    const root = renderUserAvatar({ preview });
+
+    expect(root.find('.UserAvatar-image')).toHaveLength(1);
+    expect(root.find('.UserAvatar-image')).toHaveProp('src', preview);
+  });
 });
```
bobsilverberg: You added an `altText` prop to `UserAvatar` but it doesn't seem to be tested in here.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -697,4 +704,90 @@ describe(__filename, () => {
...
+
+    // The user profile picture has been successfully deleted.
+    root.setProps({
+      user: {
+        ...user,
+        picture_url: null,
+      },
+    });
+
+    expect(root.find(Notice)).toHaveLength(1);
```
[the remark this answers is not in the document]
willdurand: yes, thanks!

tests/unit/amo/components/TestUserProfilePictureEdit.js
willdurand: 🤦‍♂️

tests/unit/amo/components/TestUserProfilePictureEdit.js
willdurand: That would make sense, thanks!

willdurand:
Thanks for the review @bobsilverberg, looks like I overlooked a bunch of assertions. This is ready for another look.

PR #5009, opened 2018-05-15
Add a confirmation when adding add-ons to a collection

### LINKED ISSUE #4979: Add a confirmation when adding add-ons to a collection
We need to add a confirmation when a user adds something to their collection, so they know it happened even if they don't see it. Similar to how I've mocked it up below (for ratings):
screen shot 2018-05-10 at 4 10 32 pm

Rules:
- The message bar should appear above the "Find add-ons" field.
- It should read: "Extension added" or "Theme added" depending on type.
- It should have the green checkmark glyph.
- It should fade away after 5 seconds.

### LINKED ISSUE #4982: We need tests for CollectionManager.onAddonSelected
I'm not sure how I managed to land this without any tests, but I did, so we need some tests for the "Add to Collection" feature of the CollectionManager.

bobsilverberg:
Fixes #4979 

Also fixes #4982 

Note that the message does not currently fade out. I'm going to open a follow-up issue about that. I also think the placement could be improved as well, which can also be dealt with in a follow-up.

Pending @pwalm's response to https://github.com/mozilla/addons-frontend/issues/4979#issuecomment-389245823, it just uses a generic message rather than one customized by add-on type.
 
Screenshot of message:

screenshot 2018-05-15 16 18 44

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -323,7 +323,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
         </div>
         {this.state.addAddonStatus === ADDON_ADDED_STATE_SUCCESS && (
           <Notice type="success">
-            {i18n.gettext('Add-on added')}
+            {i18n.gettext('Added to collection')}
           </Notice>
         )}
         {!creating &&
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -20,6 +20,7 @@ import translate from 'core/i18n/translate';
 import { decodeHtmlEntities } from 'core/utils';
 import Button from 'ui/components/Button';
 import LoadingText from 'ui/components/LoadingText';
+import Notice from 'ui/components/Notice';
 import type {
   SearchFilters, SuggestionType,
 } from 'amo/components/AutoSearchInput';
@@ -37,7 +38,18 @@ import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
+export const ADDON_ADDED_STATE_PENDING: 'ADDON_ADDED_STATE_PENDING'
+  = 'ADDON_ADDED_STATE_PENDING';
+export const ADDON_ADDED_STATE_SUCCESS: 'ADDON_ADDED_STATE_SUCCESS'
+  = 'ADDON_ADDED_STATE_SUCCESS';
+
+export type AddonAddedStatusType = |
+  typeof ADDON_ADDED_STATE_PENDING |
+  typeof ADDON_ADDED_STATE_SUCCESS |
+  null;
+
 type Props = {|
+  hasAddonBeenAdded: boolean,
   collection: CollectionType | null,
   clientApp: ?string,
   creating: boolean,
@@ -53,16 +65,21 @@ type Props = {|
 |};
 
 type State = {|
+  addAddonStatus: AddonAddedStatusType,
+  customSlug?: boolean,
   description?: string | null,
   name?: string | null,
   slug?: string | null,
-  customSlug?: boolean,
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
-    this.state = this.propsToState(props);
+    const state = this.propsToState(props);
+    this.state = {
+      ...state,
+      addAddonStatus: null,
+    };
   }
 
   componentWillReceiveProps(props: Props) {
@@ -72,6 +89,12 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       // user is not already editing. This prevents clearing the form
       // in a few scenarios such as pressing the submit button.
       this.setState(this.propsToState(props));
+    }
+    if (this.props.hasAddonBeenAdded !== props.hasAddonBeenAdded) {
+      this.setState({
+        addAddonStatus: props.hasAddonBeenAdded ?
+          ADDON_ADDED_STATE_SUCCESS : null,
+      });
     }
   }
 
@@ -205,6 +228,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       page: page || 1,
       userId: siteUserId,
     }));
+    this.setState({ addAddonStatus: ADDON_ADDED_STATE_PENDING });
   };
 
   propsToState(props: Props) {
@@ -297,6 +321,11 @@ export class CollectionManagerBase extends React.Component<Props, State> {
             value={this.state.slug}
           />
         </div>
+        {this.state.addAddonStatus === ADDON_ADDED_STATE_SUCCESS && (
+          <Notice type="success">
+            {i18n.gettext('Add-on added')}
+          </Notice>
+        )}
         {!creating &&
           <AutoSearchInput
             inputName="collection-addon-query"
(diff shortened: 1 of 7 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -56,12 +68,17 @@ type State = {|
   description?: string | null,
   name?: string | null,
   slug?: string | null,
+  addAddonStatus: AddonAddedStatusType,
```
willdurand: Shouldn't this be positioned first?

src/amo/components/CollectionManager/index.js
```diff
@@ -37,7 +38,18 @@ import type { ReactRouterType } from 'core/types/router';
...
 
+export const ADDON_ADDED_STATE_PENDING: 'ADDON_ADDED_STATE_PENDING'
+  = 'ADDON_ADDED_STATE_PENDING';
+export const ADDON_ADDED_STATE_SUCCESS: 'ADDON_ADDED_STATE_SUCCESS'
+  = 'ADDON_ADDED_STATE_SUCCESS';
+
+export type AddonAddedStatusType = |
+  typeof ADDON_ADDED_STATE_PENDING |
+  typeof ADDON_ADDED_STATE_SUCCESS |
+  null;
```
willdurand: We put the pipe first usually :)

```js
export type AdonAddedStatusType =
  | ...
  | ...;
```

src/amo/components/CollectionManager/index.js
```diff
@@ -37,7 +38,18 @@ import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
+export const ADDON_ADDED_STATE_PENDING: 'ADDON_ADDED_STATE_PENDING'
+  = 'ADDON_ADDED_STATE_PENDING';
+export const ADDON_ADDED_STATE_SUCCESS: 'ADDON_ADDED_STATE_SUCCESS'
+  = 'ADDON_ADDED_STATE_SUCCESS';
+
+export type AddonAddedStatusType = |
```
willdurand: So the constants have "state" in their names but here we have "status". Would it make sense to have `AddonAddedStateType`?

src/amo/components/CollectionManager/index.js
```diff
@@ -56,12 +68,17 @@ type State = {|
...
   name?: string | null,
   slug?: string | null,
+  addAddonStatus: AddonAddedStatusType,
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
-    this.state = this.propsToState(props);
+    const state = this.propsToState(props);
```
willdurand: Can't `propsToState()` return a default value for `addAddonStatus`?

src/amo/components/CollectionManager/index.js
willdurand: Also, is it `status` or `state`?

src/amo/components/CollectionManager/index.js
willdurand: I would not put `null` here by the way because the state (or status) should really be either "pending" or "success". The state attribute could be of that type *or* `null`, and we could reuse this type somewhere else, maybe.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -567,8 +580,73 @@ describe(__filename, () => {
...
+    const newState = root.state();
+    expect(newState.addAddonStatus).toEqual(ADDON_ADDED_STATE_PENDING);
+  });
+
+  it('displays a notification after an add-on has been added', () => {
+    const root = render({});
+
+    expect(root.find(Notice)).toHaveLength(0);
+
+    root.setProps({ hasAddonBeenAdded: true });
```
willdurand: So what happens if `hasAddonBeenAdded` is `null` here?

src/amo/components/CollectionManager/index.js
bobsilverberg: Let's make it `status`. I'll update it elsewhere.

tests/unit/amo/components/TestCollectionManager.js
bobsilverberg: `hasAddonBeenAdded` is a boolean which does not accept nulls. I did add an extra test case to verify that the notification is removed when `hasAddonBeenAdded` switches from `true` to `false`.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -625,8 +641,87 @@ describe(__filename, () => {
...
+
+    expect(root.find(Notice)).toHaveLength(1);
+
+    root.setProps({ hasAddonBeenAdded: false });
+
+    expect(root.find(Notice)).toHaveLength(0);
```
willdurand: This test case does not really do what its description says: I don't see the "after a new add-on has been selected" part in the test.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -300,6 +300,18 @@ describe(__filename, () => {
       expect(savedState.collections).toEqual(null);
     });
 
+    it('sets a hasAddonBeenAdded flag when begining to add add-on to collection', () => {
```
willdurand: beginning?

tests/unit/amo/sagas/test_collections.js
```diff
@@ -288,14 +288,71 @@ describe(__filename, () => {
...
+        errorHandlerId: errorHandler.id,
+        slug: collectionSlug,
+        user: params.userId,
+      });
+
+      expect(sagaTester.wasCalled(unexpectedFetchAction.type)).toEqual(false);
+      mockApi.verify();
+    });
+
+    it('posts an add-on to a collection in edit mode', async () => {
```
willdurand: It took me a while to understand what this "edit mode" was.

tests/unit/amo/components/TestCollectionManager.js
bobsilverberg: Thanks for pointing that out. It wasn't a very good test. I think the next version is better.

tests/unit/amo/sagas/test_collections.js
bobsilverberg: I updated the title of the test to hopefully be a bit better.

tests/unit/amo/components/TestCollectionManager.js
willdurand: I did not lock the PR but feel free to request a quick review for the new test case before the merge :)

bobsilverberg: I will :-)

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -625,8 +641,93 @@ describe(__filename, () => {
...
+    expect(root.find(Notice)).toHaveLength(1);
+    expect(root.find(Notice).children()).toHaveText('Added to collection');
+  });
+
+  it('removes the notification after a new add-on has been selected', () => {
+    const root = render({});
+
+    expect(root.find(Notice)).toHaveLength(0);
+
+    root.setProps({ hasAddonBeenAdded: true });
```
willdurand: Can you add a comment to say that we set this prop to `true` in order to simulate an add-on already added previously?

bobsilverberg:
Only discopane ui tests are failing. Merging!

PR #5013, opened 2018-05-16
Allow to call the API with form data

### LINKED ISSUE #5012: Allow `callApi` to send form data
The `callApi` helper transforms the `body` parameter into a JSON string and forces the `content-type` to be `application/json`. In order to [upload a profile picture](https://addons-server.readthedocs.io/en/latest/topics/api/accounts.html#uploading-a-picture), we need a way to send form data (`multipart/form-data`) (see #4276).

https://github.com/mozilla/addons-frontend/blob/5d5103ab4ba76d4b1e0be476af1c98e91d353c0f/src/core/api/index.js#L86-L98

My proposal is to allow `FormData` instances and JS objects as `body`, because the API allows to send both files and other data (fields) at the same time.

willdurand:
Fix #5012

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,7 @@
     "HTMLSelectElement": true,
     "HTMLTextAreaElement": true,
     "Node": true,
-    // Used for uploading profile pictures.
+    // This is used to send form data to the API.
     "FormData": true
   },
   "parser": "babel-eslint",
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,6 +23,8 @@
     "HTMLSelectElement": true,
     "HTMLTextAreaElement": true,
     "Node": true,
+    // Used for uploading profile pictures.
+    "FormData": true
   },
   "parser": "babel-eslint",
   "plugins": [
--- a/src/core/api/index.js
+++ b/src/core/api/index.js
@@ -72,7 +72,7 @@ export function createApiError(
 
 type CallApiParams = {|
   auth?: boolean,
-  body?: Object,
+  body?: Object | FormData,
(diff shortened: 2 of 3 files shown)


src/core/api/index.js
```diff
@@ -141,8 +141,14 @@ export function callApi({
     options.credentials = 'include';
   }
   if (body) {
-    options.body = JSON.stringify(body);
-    options.headers['Content-type'] = 'application/json';
+    if (body instanceof FormData) {
+      options.body = body;
+      // Let the browser sets this header, including the boundary value.
+      delete options.headers['Content-type'];
```
bobsilverberg: This is not set at this point (i.e., `headers` is still just `{}`), so do we need to `delete` it?

.eslintrc
```diff
@@ -23,6 +23,8 @@
     "HTMLSelectElement": true,
     "HTMLTextAreaElement": true,
     "Node": true,
+    // Used for uploading profile pictures.
```
bobsilverberg: It's used for this now, but will this be the only thing it's ever used for? Maybe the comment should be more about the fact that it's needed for API requests?

src/core/api/index.js
willdurand: I wanted to make sure we never set this header by accident, even in the future. In addition, it describes exactly what we need here. We should never set a `content-type` here, otherwise the API request will be invalid. Does that make sense or should I just leave a comment?

bobsilverberg: Yes, I suppose that makes sense. Although I maintain that it's not needed. 😉

PR #5025, opened 2018-05-16
Change the workflow for Add A Collection to leave the user at the edit screen

### LINKED ISSUE #4977: Change the workflow for Add A Collection to leave the user at the edit screen
Currently, after a user saves a new collection, they are taken to the View Collection screen, from whence they must click the "Edit this collection" link in order to get back into edit mode to be able to add add-ons to the collection. We determined that this is not the best user experience.

Instead of this the user should "remain" on the edit collection screen, the Search box, which can be used to add an add-on to the collection, will appear, and the cold start text on the right will change to text that suggests the user add some add-ons to the collection.

### LINKED ISSUE #4978: Change button copy in collections flow
Some copy changes need to happen during the create collection flow:

View collection page
- "Edit this collection" needs to be an actual button (not a link)

Edit collection view
- When giving the collection a name in order to create it, the button should read "Create collection"
- When adding add-ons to the collection, the button should read "Save collection"

I think this was it? Meridel, can you add to clarify anything I've missed?

bobsilverberg:
Fixes #4977 
Also fixes #4978

This addresses the main issue, but I think the UX can be improved. Implementing #4978 will help, and I may just roll that into this PR.

I am also having a problem with coordinating the finishing of the create action with the redirection to the edit page. The initial problem was that, when the user is redirected to the edit page, it has to fetch and load the collection first, so the loading gifs on the right show up causing a flash of content that then gets replaced by the placeholder text.

I've tried a few things and cannot get it quite right. I could use some advice. My current attempt has me loading the collection in the saga after creating it (because we have access to the details after creating it), but that doesn't seem to make the collection available to the Collection component (or it doesn't know it's available). I thought about going a different route and adding yet another flag to the collections redux state to indicate that we just created a new collection, but I don't really like that idea. Maybe that's the proper way to do it?

@willdurand If you can take a look and pass on any advice you have I would appreciate it.

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -24,6 +24,7 @@ import log from 'core/logger';
 import { getCurrentUser, hasPermission } from 'amo/reducers/users';
 import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
+import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
 import LoadingText from 'ui/components/LoadingText';
 import MetadataCard from 'ui/components/MetadataCard';
@@ -160,9 +161,14 @@ export class CollectionBase extends React.Component<Props> {
     }
 
     return (
-      <p className="Collection-edit-link">
-        <Link {...props}>{i18n.gettext('Edit this collection')}</Link>
-      </p>
+      <Button
+        className="Collection-edit-link"
+        buttonType="neutral"
+        puffy
+        {...props}
+      >
+        {i18n.gettext('Edit this collection')}
+      </Button>
     );
   }
 
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -127,7 +127,7 @@ export class CollectionBase extends React.Component<Props> {
       return;
     }
 
-    if (collection && addonsPageChanged) {
+    if (collection && addonsPageChanged && collection.numberOfAddons) {
       this.props.dispatch(fetchCurrentCollectionPage({
         errorHandlerId: errorHandler.id,
         page: location.query.page || 1,
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -260,6 +260,12 @@ export type ExternalCollectionDetail = {|
   uuid: string,
 |};
 
+export type ExternalCollectionDetailWithLocalizedStrings = {|
+  description: {[lang: string]: string} | null,
+  name: {[lang: string]: string},
+  ...ExternalCollectionDetail
+|};
+
 export type CollectionAddonsListResponse = {|
   count: number,
   next: string,
@@ -694,6 +700,24 @@ export const changeAddonCollectionsLoadingFlag = ({
         },
       },
     },
+  };
+};
+
+type LocalizeCollectionDetailParams = {|
+  detail: ExternalCollectionDetailWithLocalizedStrings,
+  lang: string,
+|};
+
+export const localizeCollectionDetail = (
+  { detail, lang }: LocalizeCollectionDetailParams
+): ExternalCollectionDetail => {
+  invariant(detail, 'detail is required for localizeCollectionDetail');
+  invariant(lang, 'lang is required for localizeCollectionDetail');
+
+  return {
+    ...detail,
+    description: detail.description && detail.description[lang],
+    name: detail.name[lang],
   };
 };
 
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -261,9 +261,9 @@ export type ExternalCollectionDetail = {|
 |};
 
 export type ExternalCollectionDetailWithLocalizedStrings = {|
+  ...ExternalCollectionDetail,
   description: {[lang: string]: string} | null,
   name: {[lang: string]: string},
-  ...ExternalCollectionDetail
 |};
 
 export type CollectionAddonsListResponse = {|
@@ -715,9 +715,17 @@ export const localizeCollectionDetail = (
   invariant(lang, 'lang is required for localizeCollectionDetail');
 
   return {
-    ...detail,
+    addon_count: detail.addon_count,
+    author: detail.author,
+    default_locale: detail.default_locale,
     description: detail.description && detail.description[lang],
+    id: detail.id,
+    modified: detail.modified,
     name: detail.name[lang],
+    public: detail.public,
+    slug: detail.slug,
+    url: detail.url,
+    uuid: detail.uuid,
   };
 };
 
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -262,8 +262,8 @@ export type ExternalCollectionDetail = {|
 
 export type ExternalCollectionDetailWithLocalizedStrings = {|
   ...ExternalCollectionDetail,
-  description: {[lang: string]: string} | null,
-  name: {[lang: string]: string},
+  description: LocalizedString | null,
+  name: LocalizedString,
 |};
 
 export type CollectionAddonsListResponse = {|
@@ -714,11 +714,13 @@ export const localizeCollectionDetail = (
   invariant(detail, 'detail is required for localizeCollectionDetail');
   invariant(lang, 'lang is required for localizeCollectionDetail');
 
+  // Flow will not allow us to use the spread operator here, so we have
+  // to repeat all the fields.
   return {
     addon_count: detail.addon_count,
     author: detail.author,
     default_locale: detail.default_locale,
-    description: detail.description && detail.description[lang],
+    description: detail.description ? detail.description[lang] : null,
     id: detail.id,
     modified: detail.modified,
     name: detail.name[lang],
--- a/tests/unit/amo/sagas/test_collections.js
+++ b/tests/unit/amo/sagas/test_collections.js
@@ -523,13 +523,19 @@ describe(__filename, () => {
     });
 
     describe('create logic', () => {
+      const params = {
+        description: { 'en-US': 'Collection description' },
+        name: { 'en-US': 'Collection name' },
+        slug,
+        user,
+      };
+      const collectionDetail = createFakeCollectionDetail();
+      const collectionDetailResponse = {
+        ...collectionDetail,
+        ...params,
+      };
+
       it('sends a request to the collections API', async () => {
-        const params = {
-          description: { 'en-US': 'Collection description' },
-          name: { 'en-US': 'Collection name' },
-          slug,
-          user,
-        };
         const state = sagaTester.getState();
 
         mockApi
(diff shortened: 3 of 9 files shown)


willdurand:
I see that you have updated the tests to reflect the API response of `POST` calls. I guess you know about the bug I mentioned above already.

bobsilverberg:
> I dont know if this is related but when trying this code, I lose some information after the add-ons search bar loads:

Yes, this is precisely the problem I mentioned in the PR comment above, where I asked for advice. Did you see that comment?

willdurand:
> Yes, this is precisely the problem I mentioned in the PR comment above, where I asked for advice. Did you see that comment?

Yes but that was really clear to me, but I dug into the code since then.

willdurand:
> I am also having a problem with coordinating the finishing of the create action with the redirection to the edit page. The initial problem was that, when the user is redirected to the edit page, it has to fetch and load the collection first, so the loading gifs on the right show up causing a flash of content that then gets replaced by the placeholder text.

Like... this Gif?

bobsilverberg:
> Like... this Gif?

Yes, exactly. That's what I'm trying to fix. Because the call to the create API returns the object to us, there's no need to fetch the collection when going to the edit screen, and there's never a need to fetch the collection add-ons because we know there aren't any. So I'm trying to figure out how to tell the `Collection` component, which renders the edit screen, that we already have a loaded collection.

willdurand:
If solving the `201` response payload answers your questions, we might want to fix that at the saga level because only that level knows about the whole app state. Something like this would work:

```diff
diff --git a/src/amo/sagas/collections.js b/src/amo/sagas/collections.js
index e70b8c71a..b0a0fa038 100644
--- a/src/amo/sagas/collections.js
+++ b/src/amo/sagas/collections.js
@@ -248,9 +248,18 @@ export function* modifyCollection(
 
     if (creating) {
       invariant(response, 'response is required when creating');
+
+      // The API returns objects for localized fields, therefore we need to
+      // override these fields with the current user's language.
+      const detail = {
+        ...response,
+        name: response.name[lang],
+        description: response.description[lang],
+      };
+
       // If a new collection was just created, load it so that it will
       // be available when the user arrives at the collection edit screen.
-      yield put(loadCurrentCollection({ addons: [], detail: response }));
+      yield put(loadCurrentCollection({ addons: [], detail }));
       // TODO: There's a problem however that the collection isn't being recognized
       // when we get to the edit page. If we let the edit page call fetchCurrentCollection
       // it works, but then we get the flash of content. Ideally we can find a way to tell
```

bobsilverberg:
> I tried and it seems to do what you want to achieve.

Awesome, thanks @willdurand. Just making that change above makes Flow very angry, so I think I will move it into the reducer. It's good to know that my thinking was sound about loading the collection from the saga, and the problem was just that the format of the response was not as expected. 😄

bobsilverberg:
Ok, I got it working, but I've been spending the past 90 minutes fighting with Flow. I'm off for a bit, but if you could take a look @willdurand and see if you can figure out why Flow is complaining I'd really appreciate it. I've gotten it down to a couple of errors:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/amo/reducers/collections.js:719:59

Cannot get detail.description[lang] because string [1] is incompatible with number [2].

 [2] 253│   description: string | null,
        :
 [1] 708│   lang: string,
        :
     716│
     717│   return {
     718│     ...detail,
     719│     description: detail.description && detail.description[lang],
     720│     name: detail.name[lang],
     721│   };
     722│ };


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/amo/reducers/collections.js:720:23

Cannot get detail.name[lang] because string [1] is incompatible with number [2].

 [2] 256│   name: string,
        :
 [1] 708│   lang: string,
        :
     717│   return {
     718│     ...detail,
     719│     description: detail.description && detail.description[lang],
     720│     name: detail.name[lang],
     721│   };
     722│ };
     723│



Found 2 errors
```

which are very strange because there are no `numbers` involved at all. Maybe it's one of those obscure Flow bugs because of the nested object type I'm using.

Anyway, if you can see anything I might do to fix this I'd really appreciate it.

bobsilverberg:
Ok @willdurand, this is ready for a full review now. Note that I have also addressed issue #4978 because I think that the button changes on the edit screen are required for a decent UX.

New version of the Collection Detail screen (with button instead of link):

screenshot 2018-05-17 12 55 43

New Create collection screen (with specific button text):

screenshot 2018-05-17 12 56 13

tests/unit/amo/sagas/test_collections.js
```diff
@@ -523,13 +524,19 @@ describe(__filename, () => {
     });
 
     describe('create logic', () => {
+      const params = {
+        description: { 'en-US': 'Collection description' },
```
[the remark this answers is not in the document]
bobsilverberg: The problem is that we need to get `state` via `sagaTester.getState()`, and `sagaTester` isn't available here because it is created in `beforeEach`. So I'd need to move a bunch of stuff around and/or introduce some duplication in order to be able to access `state` during this setup process, which I think will just make the whole thing more complicated. So I'd prefer to leave it as it is.

src/amo/reducers/collections.js
```diff
@@ -697,6 +703,32 @@ export const changeAddonCollectionsLoadingFlag = ({
   };
 };
 
+type LocalizeCollectionDetailParams = {|
+  detail: ExternalCollectionDetailWithLocalizedStrings,
+  lang: string,
+|};
+
+export const localizeCollectionDetail = (
```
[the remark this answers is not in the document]
bobsilverberg: Added!

tests/unit/amo/sagas/test_collections.js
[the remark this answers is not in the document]
willdurand: ah, usually we transform the variable into a function like `getParams()`

src/amo/reducers/collections.js
```diff
@@ -260,6 +260,12 @@ export type ExternalCollectionDetail = {|
   uuid: string,
 |};
 
+export type ExternalCollectionDetailWithLocalizedStrings = {|
+  ...ExternalCollectionDetail,
+  description: {[lang: string]: string} | null,
```
willdurand: maybe you want to create a type for this map if we use it in many places.

tests/unit/amo/sagas/test_collections.js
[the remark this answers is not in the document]
bobsilverberg: I'm not following how that would address the issue I described.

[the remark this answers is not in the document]
willdurand:
```js
const getParams = ({ lang }) => {

getParams({ lang: state.api.lang })
```

src/amo/reducers/collections.js
```diff
@@ -697,6 +703,32 @@ export const changeAddonCollectionsLoadingFlag = ({
...
+  { detail, lang }: LocalizeCollectionDetailParams
+): ExternalCollectionDetail => {
+  invariant(detail, 'detail is required for localizeCollectionDetail');
+  invariant(lang, 'lang is required for localizeCollectionDetail');
+
+  return {
+    addon_count: detail.addon_count,
+    author: detail.author,
+    default_locale: detail.default_locale,
+    description: detail.description && detail.description[lang],
```
willdurand: What is `description` if there is no `detail.description`?

src/amo/reducers/collections.js
```diff
@@ -697,6 +703,32 @@ export const changeAddonCollectionsLoadingFlag = ({
...
+  lang: string,
+|};
+
+export const localizeCollectionDetail = (
+  { detail, lang }: LocalizeCollectionDetailParams
+): ExternalCollectionDetail => {
+  invariant(detail, 'detail is required for localizeCollectionDetail');
+  invariant(lang, 'lang is required for localizeCollectionDetail');
+
+  return {
```
willdurand: We should maybe add a comment to indicate that Flow forces us to repeat all fields, and that ideally we should not..

tests/unit/amo/sagas/test_collections.js
```diff
@@ -523,13 +524,19 @@ describe(__filename, () => {
...
+        name: { 'en-US': 'Collection name' },
+        slug,
+        user,
+      };
+      const collectionDetail = createFakeCollectionDetail();
```
willdurand: You can pass `params` to `createFakeCollectionDetail()` directly if I remember correctly.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -543,31 +550,36 @@ describe(__filename, () => {
...
 
-      it('redirects to the new collection after create', async () => {
-        const params = {
-          slug,
-          user,
-        };
-
-        mockApi.expects('createCollection').returns(Promise.resolve());
+      it('redirects to the collection edit screen after create', async () => {
+        mockApi.expects('createCollection').returns(Promise.resolve(collectionDetailResponse));
```
willdurand: Shouldn't we have `once()` in between?

tests/unit/amo/sagas/test_collections.js
```diff
@@ -543,31 +550,36 @@ describe(__filename, () => {
...
           .once()
-          .returns(Promise.resolve());
+          .returns(Promise.resolve(collectionDetailResponse));
 
         _createCollection(params);
 
+        const expectedLoadAction = loadCurrentCollection({
+          addons: [],
+          detail: localizeCollectionDetail({
+            detail: collectionDetailResponse, lang: state.api.lang,
```
willdurand: Could you put `lang` on a new line to make it more visible?

src/amo/components/Collection/index.js
```diff
@@ -127,7 +128,7 @@ export class CollectionBase extends React.Component<Props> {
       return;
     }
 
-    if (collection && addonsPageChanged) {
+    if (collection && addonsPageChanged && collection.numberOfAddons) {
```
willdurand: Sorry but why do we need this change?

tests/unit/amo/sagas/test_collections.js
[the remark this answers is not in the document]
bobsilverberg: Because `collectionDetailResponse` also needs access to `params`, doing this still requires me to duplicate a bit of code, but I did it anyway because it's only one statement duplicated. Thanks for the advice. 👍

src/amo/reducers/collections.js
bobsilverberg: Yes, in fact we already have one, and it's even used in the same file! `LocalizedString` - duh.

src/amo/reducers/collections.js
bobsilverberg: If `detail.description` is `null` then `description` will be `null`. If `detail.description` is `undefined` then `description` will be `undefined`.  What other case are we concerned about?

willdurand: Ok, then I would rather see a ternary operator here, because we only want `null` or a map.

src/amo/components/Collection/index.js
bobsilverberg: Because we don't want to fetch the collection page (which fetches the add-ons) for a collection without any add-ons. When we redirect to the collection edit page after creating a new collection, fetching the add-ons causes a flash of content because it thinks we're waiting to load add-ons.

src/amo/reducers/collections.js
bobsilverberg: Ok, fixed.

PR #5031, opened 2018-05-17
Add user location and occupation on profile page

### LINKED ISSUE #5019: View Profile page does not display location and occupation - only the old page does
STR:
1. Load AMO dev and log in with an account then go to Edit my profile page.
2. Fill in all the fields from Profile section and click Update my profile button.
3. Click View my profile and observe the section below the user/display name.

Expected result:
Homepage, Location and Occupation are displayed on View profile page.

Actual result:
Only the homepage is displayed as information.

Notes:
It is not reproducible with the old view.

location occupation

willdurand:
Fix #5019

---

screen shot 2018-05-17 at 11 23 21

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -118,6 +118,22 @@ export class UserProfileBase extends React.Component<Props> {
                   </a>
                 </Definition>
               ) : null}
+              {user && user.location ? (
+                <Definition
+                  className="UserProfile-location"
+                  term={i18n.gettext('Location')}
+                >
+                  {user.location}
+                </Definition>
+              ) : null}
+              {user && user.occupation ? (
+                <Definition
+                  className="UserProfile-occupation"
+                  term={i18n.gettext('Occupation')}
+                >
+                  {user.occupation}
+                </Definition>
+              ) : null}
               <Definition
                 className="UserProfile-user-since"
                 term={i18n.gettext('User since')}
--- a/tests/unit/amo/components/TestUserProfile.js
+++ b/tests/unit/amo/components/TestUserProfile.js
@@ -178,6 +178,61 @@ describe(__filename, () => {
     expect(root.find('.UserProfile-homepage')).toHaveLength(0);
   });
 
+  it("renders the user's occupation", () => {
+    const occupation = 'some occupation';
+
+    const { store } = dispatchSignInActions({
+      userProps: {
+        occupation,
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-occupation')).toHaveLength(1);
+    expect(root.find('.UserProfile-occupation').children())
+      .toHaveText(occupation);
+  });
+
+  it("omits occupation if the user doesn't have one set", () => {
+    const { store } = dispatchSignInActions({
+      userProps: {
+        occupation: null,
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-occupation')).toHaveLength(0);
+  });
+
+  it("renders the user's location", () => {
+    const location = 'some location';
+
+    const { store } = dispatchSignInActions({
+      userProps: {
+        location,
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-location')).toHaveLength(1);
+    expect(root.find('.UserProfile-location').children()).toHaveText(location);
+  });
+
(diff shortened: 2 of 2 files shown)


tests/unit/amo/components/TestUserProfile.js
```diff
@@ -178,6 +178,61 @@ describe(__filename, () => {
...
+    expect(root.find('.UserProfile-occupation')).toHaveLength(0);
+  });
+
+  it("renders the user's location", () => {
+    const location = 'some location';
+
+    const { store } = dispatchSignInActions({
+      userProps: {
+        location,
+        username: 'tofumatt',
```
rebmullin: same here - do we need username here if it's set up in renderUserProfile?

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -178,6 +178,61 @@ describe(__filename, () => {
...
+    expect(root.find('.UserProfile-occupation')).toHaveLength(1);
+    expect(root.find('.UserProfile-occupation').children())
+      .toHaveText(occupation);
+  });
+
+  it("omits occupation if the user doesn't have one set", () => {
+    const { store } = dispatchSignInActions({
+      userProps: {
+        occupation: null,
+        username: 'tofumatt',
```
rebmullin: do we need username here if it's set up in renderUserProfile?

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -178,6 +178,61 @@ describe(__filename, () => {
...
+    const { store } = dispatchSignInActions({
+      userProps: {
+        location,
+        username: 'tofumatt',
+      },
+    });
+    const root = renderUserProfile({ store });
+
+    expect(root.find('.UserProfile-location')).toHaveLength(1);
+    expect(root.find('.UserProfile-location').children()).toHaveText(location);
```
rebmullin: not sure if this matters in a test really but is there any benefit here to assigning `root.find('.UserProfile-location')` to a var since you use this 2x?

src/amo/components/UserProfile/index.js
```diff
@@ -118,6 +118,22 @@ export class UserProfileBase extends React.Component<Props> {
...
+                </Definition>
+              ) : null}
+              {user && user.occupation ? (
+                <Definition
+                  className="UserProfile-occupation"
+                  term={i18n.gettext('Occupation')}
+                >
+                  {user.occupation}
+                </Definition>
+              ) : null}
```
rebmullin: hey @willdurand this looks really good but i just had a couple of questions on some related topics:

do we want show the LoadingText in this scenario? I am not sure if it would be useful here since they are optional fields and by the time you figure if you are supposed to show i guess it's ready but I saw it on some of the other fields just double checking..

regarding styles - i am new to sketch so i am not 100% i am reading this right and I am not even sure that i am looking at the right mock  : ) (AMO - Desktop Master) but it *looked * like maybe the fonts were a little bigger:
Alt text

what do you think?

the following is def out of scope but noticed the following (should a new / separate issue be opened for this?):
Alt text

tests/unit/amo/components/TestUserProfile.js
willdurand: it seems we do need the username here because `renderUserProfile()` we want a user named `tofumatt`, because that's the default user loaded by `renderUserProfile()`.

tests/unit/amo/components/TestUserProfile.js
willdurand: That's a good question, usually we don't add a variable for that, except when we start to use the same object more than 2 times I tink 😛

src/amo/components/UserProfile/index.js
willdurand: > do we want show the LoadingText in this scenario? I am not sure if it would be useful here since they are optional fields and by the time you figure if you are supposed to show i guess it's ready but I saw it on some of the other fields just double checking..

I did not show the `LoadingText` because those fields are optional and `homepage` (which is also an optional field) was already there, without a loading indicator...

> regarding styles - i am new to sketch so i am not 100% i am reading this right and I am not even sure that i am looking at the right mock : ) (AMO - Desktop Master) but it *looked * like maybe the fonts were supposed to be a little big bigger:

Yes, this is the right file and fonts are indeed different. The font sizes in the app and in the Sketch files are unfortunately different. I think that's because we still have to adjust a few things in reality and cannot take the Sketch values as is.

> the following is def out of scope but noticed the following (should a new / separate issue be opened for this?):

mmh yes we definitely have to open a new issue here, good catch!

rebmullin:
@willdurand  thanks for clarifying things!  Good job : )

Issue #2749, opened 2017-07-12
Add a profile page for a user


tofumatt:
We need a profile page for any user on the site. This might turn into a bit of a tracker as different elements will be needed based on the user's profile (reviews written, add-ons authored, etc.)

tofumatt:
This will close #2012.

tofumatt:
** Import note:** To mitigate spamming, the future plan for user profile pages is to _only have them accessible to the public if the user has published add-ons_.

tofumatt:
If a user doesn’t have published (approved) add-ons, only show their profile page to themselves and admins. Otherwise it should be a 404.

muffinresearch:
@jvillalobos is this something we will need for the first release?

muffinresearch:
@jvillalobos Scratch that - I think it is since we're going to hang the edit profile functionality off it.

muffinresearch:
For reference this is the current page on the legacy frontend:

<img width="1086" alt="user_info_for_muffinresearch____add-ons_for_firefox" src="https://user-images.githubusercontent.com/1514/30438640-9f8f79f2-9969-11e7-8822-fac0fdf0dd47.png">

ExE-Boss:
I think that it would be better if users with no published add-ons would still have their profile visible, but marked as a user, not a developer. Possibly with their name and ofher fields hidden.

This is because the user page shows all the reviews and, optionally, collections that that user made or followed.

muffinresearch:
Thinking that since we're not doing the settings page it would make sense to leave out the user profile view for now too.

pwalm:
Profile pages!! There will be some issues to iron out, but this is the first pass.
Public view:
![user profile - public view - 1366px](https://user-images.githubusercontent.com/6797635/34632151-b6e8d8ec-f228-11e7-901a-9c73e28f3d43.png)

User view:
![user profile - my view - 1366px](https://user-images.githubusercontent.com/6797635/34632157-bf1a1c2e-f228-11e7-95bc-330b8a5670ed.png)

Account settings:
![user profile - account settings - 1366px](https://user-images.githubusercontent.com/6797635/34632163-c8f0b172-f228-11e7-8f4b-79b9beade058.png)

ioanarusiczki:
@willdurand 
I think the Add-ons developer and Theme Artist don't show up yet on the card, is this going to be implemented or it's been dropped?
While checking I filed https://github.com/mozilla/addons-frontend/issues/5019
[...]

PR #5032, opened 2018-05-17
Add a edit link to user profile page

### LINKED ISSUE #4585: Send abuse report should not work for self reporting
STR:
1. Load AMO dev and log in with an account that has no admin rights.
2. From View Profile click on report this user for abuse.
3. Type some text and send it.

Expected results:
A message that user cannot report himself for abuse is displayed.

Actual results:
The abuse report is sent. 

Notes:

self reporting

### LINKED ISSUE #4997: Add a edit link to user profile page when it makes sense
The user profile page should have a link to the edit page when:

1. the current viewed profile is the current logged-in user's profile
2. the current logged-in user has permission to edit a user profile (admin or user with `Users:Edit` permission)

Does that make sense? If so, where should we put this link?

We could maybe add a text link below the "report" button:

screen shot 2018-05-15 at 11 58 36

willdurand:
Fix #4997
Fix #4585 

---

Admin:

screen shot 2018-05-17 at 14 20 48

Current user (owner):

screen shot 2018-05-17 at 14 20 55

Logged-out or no permission:

screen shot 2018-05-17 at 14 21 15

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import invariant from 'invariant';
 import * as React from 'react';
 import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
@@ -7,15 +8,22 @@ import { compose } from 'redux';
 import AddonsByAuthorsCard from 'amo/components/AddonsByAuthorsCard';
 import NotFound from 'amo/components/ErrorPage/NotFound';
 import ReportUserAbuse from 'amo/components/ReportUserAbuse';
-import { fetchUserAccount, getUserByUsername } from 'amo/reducers/users';
+import {
+  fetchUserAccount,
+  getCurrentUser,
+  getUserByUsername,
+  hasPermission,
+} from 'amo/reducers/users';
 import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
+  USERS_EDIT,
 } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
 import { removeProtocolFromURL, sanitizeUserHTML } from 'core/utils';
+import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
 import DefinitionList, { Definition } from 'ui/components/DefinitionList';
 import LoadingText from 'ui/components/LoadingText';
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Looking at the mockups from https://github.com/mozilla/addons-frontend/issues/2749#issuecomment-355688372, this seems different in a couple of ways:

1. The button on the mockup just says "Edit profile" not "Edit this profile".
2. When viewing your own profile, the "Edit profile" button replaces the "Report this user for abuse" button. I wonder if it's possible to report yourself for abuse? That doesn't seem to make much sense. If you cannot (or should not be able to) report yourself for abuse, then maybe we can replace the button rather than have 2 buttons when viewing your own profile.

willdurand:
> The button on the mockup just says "Edit profile" not "Edit this profile".

Mockups often differ in terms of content.. and it is tricky to know what we should be using. Happy to remove the "this" though.



> When viewing your own profile, the "Edit profile" button replaces the "Report this user for abuse" button. I wonder if it's possible to report yourself for abuse? That doesn't seem to make much sense. If you cannot (or should not be able to) report yourself for abuse, then maybe we can replace the button rather than have 2 buttons when viewing your own profile.

I agree but that's a different issue: https://github.com/mozilla/addons-frontend/issues/4585.

src/amo/components/UserProfile/index.js
```diff
@@ -187,8 +222,18 @@ export function mapStateToProps(
   state: {| users: UsersStateType |},
   ownProps: Props,
 ) {
+  const currentUser = getCurrentUser(state.users);
+  const user = getUserByUsername(state.users, ownProps.params.username);
+
+  let hasEditPermission = currentUser && user && currentUser.id === user.id;
+  if (currentUser && hasPermission(state, USERS_EDIT)) {
```
bobsilverberg: You could have done this with an `||` above, but I guess you wanted to avoid a complicated conditional statement?

src/amo/components/UserProfile/index.js
```diff
@@ -60,9 +70,23 @@ export class UserProfileBase extends React.Component<Props> {
...
   }
 
+  getEditURL() {
+    const { currentUser, user } = this.props;
+
+    invariant(user, 'user is required');
+    invariant(currentUser, 'currentUser is required');
+
+    if (currentUser.id === user.id) {
+      return `/users/edit`;
```
bobsilverberg: Because there is no string substitution here, and because this is also referenced in a test, would it be worth making this a constant that can be imported by the test?

bobsilverberg:
> I agree but that's a different issue: #4585.

I see, but I wonder if it's worth addressing at least the button part of that now, while you're already modifying the code. Seem like it could be a quick win, rather than waiting to revisit it as part of that P4 issue. What do you think?

willdurand:
> I see, but I wonder if it's worth addressing at least the button part of that now, while you're already modifying the code. Seem like it could be a quick win, rather than waiting to revisit it as part of that P4 issue. What do you think?

Yes, you are right! 👍

PR #5038, opened 2018-05-18
Fix margin in user profile edit page

### LINKED ISSUE #5035: Adjust the height of the View/Edit profile card to match the Account card's height
STR:

1. Log in to AMO-dev
2. Click on Edit My Profile on the user menu (https://addons-dev.allizom.org/en-US/firefox/users/edit)
3. Compare the height of the Edit/View Profile card (left)  with the Account card (right) 

Actual result:
There is a small height difference between the Edit/View Profile card and the Account card

Expected result:
The Edit/View profile card should be lowered to match the other card's height 

Notes:
- reproduced on AMO-dev, with FF60, Win10x64

image

willdurand:
Fix #5035

---

Before:

After:

screen shot 2018-05-18 at 10 42 04

### DIFF
--- a/src/amo/components/UserProfileEdit/styles.scss
+++ b/src/amo/components/UserProfileEdit/styles.scss
@@ -37,7 +37,7 @@ $font-size-aside: 10px;
   }
 }
 
-.UserProfileEdit-form-messages,
+.UserProfileEdit-form-messages .Notice,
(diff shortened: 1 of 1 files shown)


rebmullin:
@willdurand  - this looks good!

PR #5039, opened 2018-05-18
Login button user edit

### LINKED ISSUE #4965: User profile edit page should probably not return a 404 when user is not logged in
See: https://github.com/mozilla/addons-frontend/pull/4922#discussion_r186486440

I don't really know what should be done here, but the legacy frontend redirects to FxA. Maybe we should be doing this too. Displaying a login button would be the easiest option, but this would also show a very empty page with nothing else than a button.

### LINKED ISSUE #5034: A Server error page is displayed if user logs out on the Edit Profile page
STR:

1. Log in to AMO-dev
2. Select Edit my Profile from the user menu
3. On the Edit Profile page, click on Log out

Actual result:
A Server error is displayed when the user logs out while being on the Edit Profile page

Expected result:
User is redirected back on AMO homepage when he logs out on the Edit Profile page

Notes:
- reproduced on AMO-dev, with FF60, Win10x64

server error

willdurand:
Fix #5034
Fix #4965

---

I have added a login button to the user profile edit page:

screen shot 2018-05-18 at 14 44 09

Important: the button is named "Log in to continue" because we don't know if the user is going to edit their profile, another profile (admin) or get a 404 because the user cannot access this page.

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -228,7 +228,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
             <AuthenticateButton
               noIcon
               location={router.location}
-              logInText={i18n.gettext('Log in to continue')}
+              logInText={i18n.gettext('Log in to edit the profile')}
             />
           </Card>
         </div>
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -19,6 +19,7 @@ import {
 import AuthenticateButton from 'core/components/AuthenticateButton';
 import { USERS_EDIT } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
+import log from 'core/logger';
 import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
 import Button from 'ui/components/Button';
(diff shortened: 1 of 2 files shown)


src/amo/components/UserProfileEdit/index.js
```diff
@@ -210,11 +216,37 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
 
-    if (!currentUser || (currentUser && user && !hasEditPermission)) {
+    if (!currentUser) {
+      return (
+        <div className="UserProfileEdit">
+          <Card className="UserProfileEdit-user-links">
+            <AuthenticateButton
+              noIcon
+              location={router.location}
+              logInText={i18n.gettext('Log in to continue')}
```
bobsilverberg: I saw your comment about why this is just `Log in to continue`, but I wonder if we could come up with something a bit more meaningful. I'm not sure we need to worry about the accuracy of the button's message in the case where a user is going to get a 404, because they shouldn't be trying to access that page anyway. So maybe `Log in to edit the profile`?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -690,10 +692,97 @@ describe(__filename, () => {
...
+  });
+
+  it('does not dispatch fetchUserAccount() when user logs out on a user edit page', () => {
+    const username = 'some-user';
+
+    const { store } = signInUserWithUsername(username);
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+
+    // The user edits their profile.
+    const root = renderUserProfileEdit({ params: { username }, store });
```
bobsilverberg: I can see that this test is testing what you mean it to, but it seems like it would be even more explicit if the user were editing a profile that belonged to another user, rather than their own profile.

Issue #5023, opened 2018-05-16
User should be redirected at the top of the page after submitting a profile update


AlexandraMoga:
STR:

1. Log in to AMO-dev
2. Click on Edit My Profile from the user menu (https://addons-dev.allizom.org/en-US/firefox/users/edit)
3. Edit one of the available fileds
4. Click on Update my profile

Actual result
The user remains at the bottom of the page, unable to see the success message that confirms his account update

Expected result:
The user is redirected at the top of the page after he clicks on Update my Profile (this is how it functioned on the legacy pages)

Notes:
- reproduced on AMO-dev with FF60, Win10x64

![profile update](https://user-images.githubusercontent.com/31961530/40121884-4297a224-592b-11e8-935d-f32c972f182b.gif)

willdurand:
I believe the legacy frontend reloads the page when updating the information, hence the "move to top". I guess we could use the line below in the `UserProfileEdit` component to scroll to top:

```js
ReactDOM.findDOMNode(this).scrollIntoView();
```

Do we want to do that?

bobsilverberg:
I would prefer we just implement #4998 instead.

AlexandraMoga:
@bobsilverberg #4998 won't cover the scenario in which the user receives an error after submitting a profile update:
![error update prof](https://user-images.githubusercontent.com/31961530/40219752-63a9adfe-5a7f-11e8-88dd-35db4b411b2a.gif)

aastha980:
I would like to work on this issue

willdurand:
I noticed the presence of https://github.com/taion/react-router-scroll in our dependencies, maybe we can use it.

willdurand:
@tsl143 this issue is available if you wish.

PR #5051, opened 2018-05-19
Fixes : 5023

tsl143:
Fixes: #5023


willdurand:
Unfortunately, we won't go that way. We will instead redirect the user to their profile page.

Issue #5021, opened 2018-05-16
Edit my profile - The biography text box can be resized and after typing in text the layout of the page is broken


ioanarusiczki:
STR:
1. Load AMO dev and log in then go to edit my profile page
2. Expand the Biography text box from the bottom right corner - increase it's size (as in the .gifs below)
3. Type in some text and press enter
4. Observe the layout of the page

Expected result:
Page is displayed correctly.

Actual result:
The layout of the page is broken - Text box and Update my profile button are truncated in some cases depending on how much the text box is expanded.

Notes:
This issue is reproducible on AMO dev with FF60(Win10). 
Workaround: the page is correctly displayed after refreshing it. 

![the page layout1](https://user-images.githubusercontent.com/33448286/40117247-09380b74-591f-11e8-9d46-bc20d08282cc.gif)
 
Another example
![the page layout2](https://user-images.githubusercontent.com/33448286/40117398-85c37a8e-591f-11e8-828a-ac41fc66eebe.gif)

willdurand:
So it is a browser feature I guess, and to me, it makes sense to allow a user to resize the `textarea`. Because there is no page reload on the frontend, the browser keeps the width that has been set by the user.

I don't know if that something we can change, because the UI is rather responsive and nothing really breaks. User can still reduce the size of the textarea.

tsl143:
We can keep 100% width of textarea by default and restrict the `resize` only to vertical, in this way resize will never distort the design.
For text area we are already giving help text in next line so giving 100% width to text area by default will make sense.

<img width="745" alt="image" src="https://user-images.githubusercontent.com/5318732/40271614-10b0b144-5bbd-11e8-9c1a-d20f5ddf9c5c.png">

ioanarusiczki:
Verified fixed on AMO dev with FF60(Win10)
Because resizing is restricted to only vertical this is no longer reproducible.

![resizing the box](https://user-images.githubusercontent.com/33448286/40354127-d92d4a1a-5dbb-11e8-81b7-5024b38567db.gif)

PR #5052, opened 2018-05-19
fixes : 5021

tsl143:
Suggesting default full width of biography and resize only vertically
Fixes : #5021

### DIFF
--- a/src/amo/components/UserProfileEdit/styles.scss
+++ b/src/amo/components/UserProfileEdit/styles.scss
@@ -86,6 +86,13 @@ $font-size-aside: 10px;
   }
 }
 
+.UserProfileEdit textarea {
+  @include respond-to(large) {
+    resize: vertical;
+    width: 100%;
+  }
+}
+
 .UserProfileEdit-username,
 .UserProfileEdit-displayName,
 .UserProfileEdit-location,
--- a/src/amo/components/UserProfileEdit/styles.scss
+++ b/src/amo/components/UserProfileEdit/styles.scss
@@ -80,7 +80,9 @@ $font-size-aside: 10px;
     background: $grey-20;
     cursor: not-allowed;
   }
+}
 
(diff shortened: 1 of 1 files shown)


src/amo/components/UserProfileEdit/styles.scss
```diff
@@ -86,6 +86,13 @@ $font-size-aside: 10px;
   }
 }
 
+.UserProfileEdit textarea {
```
willdurand: Could you use `.UserProfileEdit-biography` instead?

tsl143: Updated!

willdurand:
So that does not work anymore because the rule is overridden by another one. I think we could set the `textarea` width to `100%` all the time and `resize: vertical` too. What do you think?

```diff
diff --git a/src/amo/components/UserProfileEdit/styles.scss b/src/amo/components/UserProfileEdit/styles.scss
index 1cd36e4..b91d395 100644
--- a/src/amo/components/UserProfileEdit/styles.scss
+++ b/src/amo/components/UserProfileEdit/styles.scss
@@ -80,7 +80,9 @@ $font-size-aside: 10px;
     background: $grey-20;
     cursor: not-allowed;
   }
+}
 
+.UserProfileEdit input {
   @include respond-to(large) {
     width: 60%;
   }
@@ -95,11 +97,7 @@ $font-size-aside: 10px;
 
 .UserProfileEdit-biography {
   min-height: 3rem;
-
-  @include respond-to(large) {
-    resize: vertical;
-    width: 100%;
-  }
+  resize: vertical;
 }
 
 .UserProfileEdit-buttons-wrapper {
```

tsl143:
Sounds good, I am updating PR, I cant test locally, (I dont have the setup on my office machine :P ) hoping this fixes.

willdurand:
Thanks for your PR!

Issue #5036, opened 2018-05-18
Server error on user profile page when user does not exist


willdurand:
### Describe the problem and steps to reproduce it:

1. Make sure you are logged-in
2. Go to a user profile page with a username that does not exist

### What happened?

Server Error:

<img width="1392" alt="screen shot 2018-05-18 at 10 01 18" src="https://user-images.githubusercontent.com/217628/40223174-7124ca9c-5a82-11e8-8dd9-d569fd8ee46b.png">

### What did you expect to happen?

No server error, but a 404.

AlexandraMoga:
Verified on AMO-dev with FF60, Win 10x64. A non-existent username will now generate a 404 error.

![profile load](https://user-images.githubusercontent.com/31961530/40293897-afc4d4a8-5cdb-11e8-8b91-9a72c4999c14.gif)

@willdurand - you can see in the attached video that before the "Page not found" is displayed, the user profile card will load for a brief moment. Is this an expected behavior or can it be eliminated?

willdurand:
It can be eliminated with a different error handler.

PR #5059, opened 2018-05-21
Use fixed error handler in user profile page

### LINKED ISSUE #5060: User profile card is loaded briefly when the page is accessed with a non-existent user
STR:

1. Log in to AMO-dev
2. Access the user profile page with a non-existent user (see https://addons-dev.allizom.org/en-US/firefox/user/no-username-here-bro/)

Actual result:
The user profile card will load for a brief moment before the "Page not found" is displayed

Expected result:
"Page not found" is displayed at the instant when the profile page is accessed with a non existent user

Notes:

profile load

willdurand:
Fix #5060

---

This PR fixes the behavior observed in https://github.com/mozilla/addons-frontend/issues/5036#issuecomment-390570679 by using a "fixed" error handler, which is the handler we should be using on most of our "page level components".

It allows to synchronize the errors between the server and the client, thus avoiding unnecessary data re-fetching on the client if the server gets an error.

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -23,7 +23,7 @@ import {
   ADDON_TYPE_THEME,
   USERS_EDIT,
 } from 'core/constants';
-import { withErrorHandler } from 'core/errorHandler';
+import { withFixedErrorHandler } from 'core/errorHandler';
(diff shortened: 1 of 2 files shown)


rebmullin:
@willdurand - nice!

PR #5064, opened 2018-05-21
Enable recommendations for details pages on all servers

### LINKED ISSUE #5063: Turn on TAAR lite recommendations for stage and production.
This is to put the recommendations on details pages live in stage and prod as well as -dev and local development where it is already enabled.

muffinresearch:
Fixes #5063


willdurand:
It is interesting how many different names we have for this feature: https://github.com/mozilla/addons-frontend/blob/77f3787cd4d7b08f39c8a87ce1cca96548ccf6a3/config/default.js#L295-L296

PR #5071, opened 2018-05-22
Add api/saga/reducer to fetch/load user notifications

### LINKED ISSUE #5070: Add api/saga/reducer to fetch/load user notifications
This is the first step to add user notifications to the profile edit page, see: #5000.

willdurand:
Fix #5070

---

This PR adds everything we need to load user notifications when browsing
a user profile edit page. It does not show anything in the UI yet, though.

### DIFF
--- a/src/amo/reducers/users.js
+++ b/src/amo/reducers/users.js
@@ -20,8 +20,18 @@ export const LOAD_CURRENT_USER_ACCOUNT: 'LOAD_CURRENT_USER_ACCOUNT' = 'LOAD_CURR
 export const FETCH_USER_ACCOUNT: 'FETCH_USER_ACCOUNT' = 'FETCH_USER_ACCOUNT';
 export const LOAD_USER_ACCOUNT: 'LOAD_USER_ACCOUNT' = 'LOAD_USER_ACCOUNT';
 export const DELETE_USER_PICTURE: 'DELETE_USER_PICTURE' = 'DELETE_USER_PICTURE';
+export const FETCH_USER_NOTIFICATIONS: 'FETCH_USER_NOTIFICATIONS' = 'FETCH_USER_NOTIFICATIONS';
+export const LOAD_USER_NOTIFICATIONS: 'LOAD_USER_NOTIFICATIONS' = 'LOAD_USER_NOTIFICATIONS';
 
 export type UserId = number;
+
+export type NotificationType = {|
+  enabled: boolean,
+  mandatory: boolean,
+  name: string,
+|};
+
+export type NotificationsType = Array<NotificationType>;
 
 // Basic user account object fields, returned by the API.
 export type ExternalUserType = {|
@@ -55,6 +65,7 @@ export type ExternalUserType = {|
 
 export type UserType = {|
   ...ExternalUserType,
+  notifications: NotificationsType | null,
 |};
 
 export type UsersStateType = {
@@ -227,6 +238,52 @@ export const deleteUserPicture = (
   };
 };
 
+type FetchUserNotificationsParams = {|
+  errorHandlerId: string,
+  username: string,
+|};
+
+type FetchUserNotificationsAction = {|
+  type: typeof FETCH_USER_NOTIFICATIONS,
+  payload: FetchUserNotificationsParams,
+|};
+
+export const fetchUserNotifications = ({
+  errorHandlerId,
+  username,
+}: FetchUserNotificationsParams): FetchUserNotificationsAction => {
+  invariant(errorHandlerId, 'errorHandlerId is required');
+  invariant(username, 'username is required');
+
+  return {
+    type: FETCH_USER_NOTIFICATIONS,
+    payload: { errorHandlerId, username },
+  };
+};
+
+type LoadUserNotificationsParams = {|
+  notifications: NotificationsType,
+  username: string,
+|};
+
+type LoadUserNotificationsAction = {|
+  type: typeof LOAD_USER_NOTIFICATIONS,
+  payload: LoadUserNotificationsParams,
+|};
+
+export const loadUserNotifications = ({
+  notifications,
+  username,
+}: LoadUserNotificationsParams): LoadUserNotificationsAction => {
+  invariant(notifications, 'notifications is required');
+  invariant(username, 'username is required');
+
+  return {
+    type: LOAD_USER_NOTIFICATIONS,
+    payload: { notifications, username },
+  };
+};
+
 export const getUserById = (users: UsersStateType, userId: number) => {
   invariant(userId, 'userId is required');
   return users.byID[userId];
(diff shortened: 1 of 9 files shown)


tests/unit/amo/api/test_users.js
```diff
@@ -134,4 +136,32 @@ describe(__filename, () => {
       mockApi.verify();
     });
   });
+
+  describe('userNotifications', () => {
+    const mockResponse = () => createApiResponse({
+      jsonData: createUserNotificationsResponse(),
+    });
+
+    const getParams = (params = {}) => {
```
bobsilverberg: We've got the same `getParams` issue that I mentioned in another review, and I don't recall now what your response was, but my concern was the fact that we add all this code to have a flexible way of generating params for tests, but then never use any of it and just call `getParams` once.

src/amo/reducers/users.js
```diff
@@ -317,16 +390,29 @@ const reducer = (
...
+      return {
+        ...state,
+        ...addUserToState({ notifications, state, user }),
+      };
+    }
+    case LOAD_USER_NOTIFICATIONS: {
+      const { notifications, username } = action.payload;
+      const user = getUserByUsername(state, username);
+
+      if (!user) {
```
bobsilverberg: When would this happen, and if it does happen should that be an error?

bobsilverberg:
> Argh, I don't know what I was thinking... Sure I think it would simplify a bunch of things to do the same thing no matter the user.

What *I* thought you were thinking is that we need to get the current user often, but we don't always need the notifications - we only need them when editing. So that's why you didn't want to always fetch them for the current user. But I think simplifying the code would outweigh that issue (if that is what you were concerned about).

willdurand:
@bobsilverberg I think we should actually split the two actions (loading a user and loading their notifications) because we use the "load user" action for displaying the profile and we won't be able to fetch the user notifications all the time (current user not logged-in, current user with no permissions, etc.).

So I think we should simplify by only loading the notifications for a user who is being edited. Does that make sense?

bobsilverberg:
> So I think we should simplify by only loading the notifications for a user who is being edited. Does that make sense?

So just make it a separate action that always needs to be called when editing a profile? Yeah, that sounds simpler and also addresses the potential performance issue.

src/amo/reducers/users.js
willdurand: Redux `reducer()` function should be pure and always return a state, which is why I'd like not to throw errors here.

I believe Flow encouraged me to ensure we have a user. In real life, there is no way to load notifications without the user being loaded first.

willdurand:
@bobsilverberg so I think this is good for a review, I'll create a separate PR for the dispatch logic in the component because there are a bunch of new test cases and the diff got too big.

willdurand:
> @bobsilverberg so I think this is good for a review, I'll create a separate PR for the dispatch logic in the component because there are a bunch of new test cases and the diff got too big.

Well, actually I reduced the diff so it should be ok. I made two very distinct commits to ease your life 😉

willdurand:
@bobsilverberg just rebased this PR :)

src/amo/reducers/users.js
```diff
@@ -355,6 +425,24 @@ const reducer = (
         ...addUserToState({ state, user }),
       };
     }
+    case LOAD_USER_NOTIFICATIONS: {
+      const { notifications, username } = action.payload;
+
+      const user = getUserByUsername(state, username);
+
+      invariant(user, 'user is required');
```
bobsilverberg: I realize that `notifications` can be `null`, so Flow won't complain about this, but would we ever be dispatching `LOAD_USER_NOTIFICATIONS` without having any `notifications`? I.e., would it make sense to also have `invariant(notifications, 'notifications is required');` here as well?

tests/unit/amo/api/test_users.js
```diff
@@ -136,7 +160,7 @@ describe(__filename, () => {
           method: 'DELETE',
           state: params.api,
         })
-        .returns(mockResponse());
+        .returns(createApiResponse());
```
bobsilverberg: Why this change?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -125,58 +128,111 @@ describe(__filename, () => {
...
     }));
+    sinon.assert.calledWith(dispatchSpy, fetchUserNotifications({
+      errorHandlerId: root.instance().props.errorHandler.id,
+      username,
+    }));
   });
 
-  it('does not dispatch fetchUserAccount action if there is no username', () => {
-    const { store } = signInUserWithUsername('tofumatt');
+  it('only dispatches fetchUserNotifications action when the current logged-in user is being edited', () => {
```
bobsilverberg: Nit: I found this description a bit confusing, until I figured it out. I interpreted the `only` as "it will only do this when", like as a condition, whereas I now realize you mean "it will only dispatch fetchUserNotifications and not fetchUserAccount". Do you think you can find a way to word it that is less ambiguous?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -125,58 +128,111 @@ describe(__filename, () => {
...
+      username,
+      notifications: createUserNotificationsResponse(),
+    }));
+
+    dispatchSpy.reset();
 
     // This happens when loading the user edit profile page of the current
     // logged-in user (e.g., page refresh).
-    renderUserProfileEdit({ params: {}, store });
+    renderUserProfileEdit({ errorHandler, params: {}, store });
```
bobsilverberg: Why did you need to create an `errorHandler` in this test now, when one wasn't needed before? We're not asserting anything about it.

tests/unit/amo/reducers/test_users.js
```diff
@@ -112,6 +106,63 @@ describe(__filename, () => {
...
+    it('loads notifications for a user', () => {
+      const userId = 12345;
+      const user = createUserAccountResponse({ id: userId, username: 'john' });
+      const notifications = createUserNotificationsResponse();
+
+      const prevState = reducer(initialState, loadCurrentUserAccount({ user }));
+
+      expect(prevState.byID[userId]).toEqual({
+        ...user,
+        notifications: null,
```
bobsilverberg: You've kind of already tested this above, in `it('sets notifications to `null` when loading the current user'`, but I guess there's no harm in asserting it again here.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: Please, do let me know every time my English is not correct 🤗 

I will rephrase this description.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: copy/pasta 😞

PR #5074, opened 2018-05-22
I5022 fixing collection edit load issue

### LINKED ISSUE #5022: Edit collection page is not loading anymore after page refresh
STR:

1. Log in to AMO dev
2. Go to My Collections from the user menu and select one of your collections to view its details page
3. Click on Edit Collection
4. Refresh the page, then click on Save Collection

Actual result:
The list of add-ons and the collection details are not loading 

Expected result:
The list of add-ons and the collection details are loaded 

Notes:
- another valid way to reproduce this is by manually changing the page in the URL after clicking on Edit Collection
- also reproduced when Edit Collection link is opened in a new tab
- reproduced on AMO-dev with FF60, Win 10x64

coll page load

rebmullin:
fixes #5022

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -82,13 +82,11 @@ export class CollectionBase extends React.Component<Props> {
       return;
     }
 
-
-
     const newCollection = collection && collection.numberOfAddons === 0;
 
     // If its loading and we don't have all our addon items
     // OR if its an empty collection, we can stop here.
-    if (loading && !newCollection || newCollection) {
+    if ((loading && !newCollection) || newCollection) {
       return;
     }
 
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -82,6 +82,16 @@ export class CollectionBase extends React.Component<Props> {
       return;
     }
 
+
+
+    const newCollection = collection && collection.numberOfAddons === 0;
+
+    // If its loading and we don't have all our addon items
+    // OR if its an empty collection, we can stop here.
+    if (loading && !newCollection || newCollection) {
+      return;
+    }
+
     let collectionChanged = false;
     let addonsPageChanged = false;
     let { location } = this.props;
@@ -107,16 +117,11 @@ export class CollectionBase extends React.Component<Props> {
       collectionChanged = true;
     }
 
-    // When switching into edit mode, refresh the collection add-ons.
-    if (this.props.editing && !nextProps) {
-      addonsPageChanged = true;
-    } else if (loading && !this.props.editing) {
-      // Since we are refreshing the collection addon-ons in edit-mode
-      // we only want to stop if we are outside of this.
-      return;
-    }
+    const collectionHasAddons = (collection
+      && collection.addons
+      && collection.addons.length > 0);
 
-    if (!collection || collectionChanged) {
+    if (!collection || !collectionHasAddons || collectionChanged) {
       this.props.dispatch(fetchCurrentCollection({
         errorHandlerId: errorHandler.id,
         page: location.query.page,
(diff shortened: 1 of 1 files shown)


src/amo/components/Collection/index.js
```diff
@@ -114,6 +110,10 @@ export class CollectionBase extends React.Component<Props> {
     // When switching into edit mode, refresh the collection add-ons.
     if (this.props.editing && !nextProps) {
       addonsPageChanged = true;
+    } else if (loading && !this.props.editing) {
```
bobsilverberg: Thanks @rebmullin. This fixes the problem, but I'm a bit confused about it. If `loading` is true, then why do we need to not `return` when we're in edit mode? It seems like if `loading` is true then we're waiting for a fetch action to return, and that should be true whether we're in edit mode or not. For some reason, in edit mode, when we refresh the page, `loading` is true, but yet a fetch is never completed which updates `loading` to false.

That seems like a bug, and I'm concerned that this fix is just a band-aid that will hide that other problem. Can you look into it a bit more to see if you can figure out the source of the issue I just described?

rebmullin: thanks @bobsilverberg. I'll take another look

rebmullin:
hey @bobsilverberg  - After looking into this a bit it seems that as you noted the fetch doesn’t complete b/c it is stopped at this loading pt. I *think* it’s still in the loading phase from mapStateToProps's logic but i am not entirely sure if this is where the issue comes from. I don't see where `FETCH_CURRENT_COLLECTION_PAGE` is being dispatched from when it does stop here - at the loading point and where `current.loading` is being set to `true` since it doesn't *appear* to be called..?
I am not sure if this is the best solution but I’ve added a few additional checks to see if the collection exists in addition to loading and make sure it’s not a brand new collection so it continues on if it needs to..

note: This one was a bit tricky/time consuming to debug b/c I literally had to restart the server after almost every change. I tried updating SSR setting in config but I don’t see the issue when I am in that mode.

src/amo/components/Collection/index.js
```diff
@@ -111,12 +115,11 @@ export class CollectionBase extends React.Component<Props> {
       collectionChanged = true;
     }
 
-    // When switching into edit mode, refresh the collection add-ons.
-    if (this.props.editing && !nextProps) {
-      addonsPageChanged = true;
-    }
+    const collectionHasAddons = (collection
```
bobsilverberg: Nit: There's not really a need for the parentheses here.

src/amo/components/Collection/index.js
```diff
@@ -111,12 +115,11 @@ export class CollectionBase extends React.Component<Props> {
...
+      && collection.addons
+      && collection.addons.length > 0);
 
-    if (!collection || collectionChanged) {
+    if (!collection || !collectionHasAddons || collectionChanged) {
```
bobsilverberg: I'm not sure I follow why we needed to add the check for `!collectionHasAddons` here.

src/amo/components/Collection/index.js
```diff
@@ -308,10 +311,16 @@ export const mapStateToProps = (
       hasPermission(state, COLLECTIONS_EDIT);
   }
 
+  let collectionToLoad = loading;
```
bobsilverberg: This is an odd variable name as it makes me think it will contain a collection (i.e., the collection to load).

src/amo/components/Collection/index.js
rebmullin: what i believe i was seeing was that  sometimes there was a collection but the addons hadn't loaded yet..

src/amo/components/Collection/index.js
rebmullin: lol..yes i think i need to improve on my var naming skills :).

src/amo/components/Collection/index.js
rebmullin: oops.. yep...i forgot to remove these..

bobsilverberg:
@rebmullin It look me a long time, but I finally figured out what was causing this issue, and have a patch to fix it. Thanks for taking the time to look into it, but I think this patch is no longer needed so I'm going to close this PR.

PR #5089, opened 2018-05-23
Extract / Merge / Debug locales

bobsilverberg:


bobsilverberg:
It's the discopane ui tests that are failing, and it has nothing to do with this patch, so I'm merging.

PR #5097, opened 2018-05-24
Updating geckodriver for discopane tests

jrbenny35:
This is an attempt to try and get the discopane tests more stable, first by updating the geckodriver version.


jrbenny35:
@bobsilverberg @willdurand I  ran this like 5 times, and it seems to have helped.

PR #5098, opened 2018-05-24
Added vertical scroll bar for screenshots

### LINKED ISSUE #4887: Image Orientation not proper
In the **Screenshots** section of any add-on the image is not properly oriented 

### What happened?

Full Image View :

capture2

In the ScrollBar the image is cut from bottom and which doesn't look good.

capture3

 Either there should be a vertical scroll bar too or the dimensions of the Screenshots section should be changed to get proper view.  

Mentored by: @curiouslearner @realslimshanky

Vimal-Raghubir:
Changed the display of screenshots by adding a vertical scroll bar since the lower portions of screenshots were being cut off. See #4887 for the before screenshot.

Change showing vertical scroll bar and portion changed in overflow-y section (Had a value of hidden before):

vertical

Vertical bar in actual display:
after

`Fixes #4887`

### DIFF
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -17,7 +17,7 @@ $screenshot-width: 320px;
   list-style-type: none;
   margin: 0;
   overflow-x: scroll;
-  overflow-y: hidden;
+  overflow-y: visible;
   padding: 0;
   width: auto;
 }


rebmullin:
Thanks @Vimal-Raghubir for helping out : ) Unfortunately on desktop (FF on mac), I don't see a vertical scrollbar and so the scrollbar still overlaps the image a bit at the bottom:
Alt text

I am wondering if we just add a bit of height to the psw-thumbnails section if that will work? Maybe about ~ 220px. note: this might need to be adjusted for mobile/smaller breakpoints tho. 

Also - as an alternative, we may want to extend the height of the `ScreenShots-viewport` and/ or change/shrink the image height a bit to avoid the vertical scrollbar altogether (if your fix of vertical scrollbar does appear on Windows).

Vimal-Raghubir:
After testing I realized the second suggestion doesn't work, however the first one does. By extending the psw-thumbnails section height to 220px the problem is fixed! :)

Vimal-Raghubir:
New PR can be seen at #5133.

PR #5101, opened 2018-05-24
I4744 adding key fixes ssr bug (seen on addon detail in AddonsByAuthorsCard)

### LINKED ISSUE #4744: Wrong alignment for "More themes by this user" if view listing is pressed on your submitted theme
Steps to reproduce:
1.Login with an account where you have more themes submitted and approved
2.In developer hub, open one of your submitted themes
3.Press View Listing link for that theme

Expected results:
The themes listed under "More themes by this user" are correctly aligned (6 themes listed , 3 on every row)

Actual results:
Each theme is displayed on one line

NOTES:
Reproducible on Stage and -dev 

lexa

rebmullin:
fixes #4744 

if the theme has a description/summary, the ShowMoreCard would be displayed which uses the Card component. I think b/c there were no keys here it was not acting in an unexpected way.  By adding a varying key value, this looks to be fixed..

Both screenshots are after refreshing the page (SSR bug)

Before (everything looks like it's using AddsonByAuthorsCard except the section classes are wrong):
Alt text

After:
Alt text

*Note*: i wasn't loving this key name but i didn't see a lot of options here. We might need to find a more unique name..

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -388,12 +388,14 @@ export class AddonBase extends React.Component {
       return null;
     }
 
-    {/* Adding wrapping divs here seems to address what we think is a
+    {
+      /* Adding wrapping divs here seems to address what we think is a
       reconcillation issue —— which causes the classname to not always get added
       correctly (e.g.: when the page is refreshed and the addon has
       a description).
       See https://github.com/mozilla/addons-frontend/issues/4744
-    */}
+    */
+    }
     return (
       <div>
         <AddonsByAuthorsCard
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -388,15 +388,23 @@ export class AddonBase extends React.Component {
       return null;
     }
 
+    {/* Adding wrapping divs here seems to address what we think is a
+      reconcillation issue —— which causes the classname to not always get added
+      correctly (e.g.: when the page is refreshed and the addon has
+      a description).
+      See https://github.com/mozilla/addons-frontend/issues/4744
+    */}
     return (
-      <AddonsByAuthorsCard
-        addonType={addon.type}
-        authorDisplayName={addon.authors[0].name}
-        authorUsernames={addon.authors.map((author) => author.username)}
-        className="Addon-MoreAddonsCard"
-        forAddonSlug={addon.slug}
-        numberOfAddons={6}
-      />
+      <div>
+        <AddonsByAuthorsCard
+          addonType={addon.type}
+          authorDisplayName={addon.authors[0].name}
+          authorUsernames={addon.authors.map((author) => author.username)}
+          className="Addon-MoreAddonsCard"
+          forAddonSlug={addon.slug}
+          numberOfAddons={6}
+        />
+      </div>
     );
   }
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -388,14 +388,13 @@ export class AddonBase extends React.Component {
       return null;
     }
 
-    {
-      /* Adding wrapping divs here seems to address what we think is a
+    /* Adding wrapping divs here seems to address what we think is a
       reconcillation issue —— which causes the classname to not always get added
       correctly (e.g.: when the page is refreshed and the addon has
       a description).
       See https://github.com/mozilla/addons-frontend/issues/4744
     */
-    }
+
     return (
       <div>
         <AddonsByAuthorsCard
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -47,7 +47,6 @@ export default class AddonsCard extends React.Component<Props> {
 <<<<<<< HEAD
 =======
   cardContainer: React.ElementRef<any> | null;
-  keyCount: number;
 
 >>>>>>> fixing flow stuff
   static defaultProps = {
@@ -120,7 +119,6 @@ export default class AddonsCard extends React.Component<Props> {
     );
     return (
       <CardList
-        key={addons && addons.length}
         {...otherProps}
         className={allClassNames}
         ref={(ref) => { this.cardContainer = ref; }}
--- a/src/amo/components/AddonsCard/index.js
+++ b/src/amo/components/AddonsCard/index.js
@@ -56,16 +56,6 @@ export default class AddonsCard extends React.Component<Props> {
     placeholderCount: DEFAULT_API_PAGE_SIZE,
   };
 
-  constructor(props: Props) {
-    super(props);
-
-    this.keyCount = 0;
-  }
-
-  getKey = () => {
-    return this.keyCount++;
-  }
-
   render() {
     const {
       addonInstallSource,
@@ -130,7 +120,7 @@ export default class AddonsCard extends React.Component<Props> {
     );
     return (
       <CardList
-        key={`AddonsCard${this.getKey()}`}
+        key={addons && addons.length}
         {...otherProps}
         className={allClassNames}
         ref={(ref) => { this.cardContainer = ref; }}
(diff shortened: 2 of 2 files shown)


rebmullin:
> Note: i wasn't loving this key name but i didn't see a lot of options here. We might need to find a more unique name..

i decided to use a function to generate new and unique keys

rebmullin:
@bobsilverberg / @willdurand - can one of you take a look at this when you have time? 

note: i am note sure about the CI error tho...any idea?

src/amo/components/AddonsCard/index.js
```diff
@@ -91,6 +102,7 @@ export default class AddonsCard extends React.Component<Props> {
       type && `AddonsCard--${type}`);
     return (
       <CardList
+        key={`AddonsCard${this.getKey()}`}
```
bobsilverberg: I don't see why we'd need to add a key here, or maybe I'm not understanding. I thought we only needed a key when we're asking react to render a number of items in a loop. Here we're only rendering a single `CardList` component, so why would we need a key?

rebmullin: Right, you should definitely use in the looping array scenarios but I’ve also seen when react does unexpected things (like renaming a class to sibling's class - which is what it does here - see screenshots above), adding a key usually can help with this. Here are a few articles that discuss this to some degree. 

https://www.jstips.co/en/react/keys-in-children-components-are-important/

https://reactjs.org/docs/lists-and-keys.html

These article are talking more in the looping array sense (similar to what you mentioned above ) but they discuss react being able to differentiate between the siblings (which is what it seems to be having trouble doing here).  

This seems to be the problem code (from Addons file):
```javascript
<div className="Addon-main-content">
     {this.renderAddonsByAuthorsCard({ isForTheme: true })}
     …
     {this.renderShowMoreCard()}
     ...
</div>
```
For some reason react is not able to differentiate this for some reason when there is a description (which is what `renderShowMoreCard` renders) and adding a key helps. I think because they both have the `Card` component as (nested) child..

Maybe there is a another better way but as far as I understand adding a key doesn't hurt anything, it just helps react out with differentiating between siblings if needed.

note 1: I also noticed that If you add a div around either render function/or component show above it is another way to help with the issue... but I thought it was better to use the key approach so we are not changing the html structure.

note 2: I could also move adding the key to the `Card` component (vs `CardList`) since that is the component that is used by both `ShowMoreCard` and `AddonCard`

src/amo/components/AddonsCard/index.js
```diff
@@ -34,13 +34,24 @@ type Props = {|
...
   }
 
+  constructor(props: Props) {
+    super(props);
+
+    this.keyCount = 0;
+  }
+
+  getKey = () => {
+    return this.keyCount++;
```
willdurand: This could unfortunately lead to SSR issues if client and server don't come up with the same values for some reason. It would be better to have fixed keys.

willdurand:
> thanks @willdurand. So I tried using length of addons as a key which seems to help the issue too. Will that work better? I am not sure if that's what you mean by a "fixed" key nor am I totally clear on why the values would be different but let me know know if this is any better since this approach uses a prop value that is passed down (vs the counter)

So I spent a considerable amount of time debugging react and I think there is an issue during reconciliation. I am not able to come up with a minimal reproducible setup though, so I am not sure how to report this problem to the React team.

I suggest that we surround the `AddonsByAuthorsCard` component with a `<div>` in `renderAddonsByAuthorsCard()` (including a comment about why it is needed).

rebmullin:
thanks @willdurand for investigating this one 1 too : )

I was thinking the key way was better but if you think the div way is okay, I can def use that approach instead. I will update..

willdurand:
> I was thinking the key way was better but if you think the div way is okay, I can def use that approach instead. I will update..

Yes, that would be good to have keys but I am a little worried by the `key` generation here. Usually, dynamic values don't play well with SSR.

rebmullin:
hey @willdurand, I've updated this but just curious how do you know when SSR code is different than client code? Is it just a matter of comparing what's in the DOM via inspector vs viewing the page source? When I do this tho, it still shows they're are differences between the 2 (assuming this is a valid way to check) - even tho it does look better w/ this update - I'm just wondering if this is okay?

bobsilverberg:
@willdurand Seeing as you're the one mostly helping @rebmullin with this, I'm going to unassign it from myself and assign it to you.

willdurand:
> hey @willdurand, I've updated this but just curious how do you know when SSR code is different than client code? Is it just a matter of comparing what's in the DOM via inspector vs viewing the page source? When I do this tho, it still shows they're are differences between the 2 (assuming this is a valid way to check) - even tho it does look better w/ this update - I'm just wondering if this is okay?

React usually output warnings when HTML is different between client and server.

willdurand:
@kumar303 have you ever seen this kind of issue?

rebmullin:
@willdurand @kumar303, I wanted to see about revisiting this one - is this okay to merge or should we hold off for now?

src/amo/components/Addon/index.js
```diff
@@ -388,15 +388,25 @@ export class AddonBase extends React.Component {
...
     return (
-      <AddonsByAuthorsCard
-        addonType={addon.type}
-        authorDisplayName={addon.authors[0].name}
-        authorUsernames={addon.authors.map((author) => author.username)}
-        className="Addon-MoreAddonsCard"
-        forAddonSlug={addon.slug}
-        numberOfAddons={6}
-      />
+      <div>
```
kumar303: Does this actually work? You need to make sure you restart your dev server when testing it because that will reload the code running on the server. The client code reloads automatically which would have introduced a new dom element when compared to the server rendered dom and forced a re-render.

kumar303:
I asked a question about this patch because I'm not sure it will fix the issue.

Coincidentally, I am experiencing the same problem in https://github.com/mozilla/addons-frontend/issues/5447 where the server rendered `className` is different from the client rendered one.

It appears that this is expected behavior! That is, React does not update class names in the DOM when hydrating server HTML.

> The styles not being patched up is expected. That's an intentional breaking change.

https://github.com/facebook/react/issues/10591#issuecomment-326635627

src/amo/components/Addon/index.js
willdurand: Yes, I believe I did a full restart because I don't trust the hot reload, I have many weird issues so I often restart.

kumar303:
I was curious to dive in and see if this issue is indeed the same as https://github.com/mozilla/addons-frontend/issues/5447 and it is.

First, the `<div>` fix does seem to fix it for some odd reason 😎 . It's probably triggering an edge case in React's reconciliation process so we shouldn't rely on it as a permanent fix. However, it works so I would be fine to merge this and wait until it breaks again 🤷‍♀️ 

The reason this is the same `className` issue is that when `ShowMoreCard` renders on the server, it renders [in an expanded state](https://github.com/mozilla/addons-frontend/blob/f7529e876f881cb5f0f7b580153cf066dbb63ae3/src/ui/components/ShowMoreCard/index.js#L28). When it hydrates on the client, it looks at `contents.clientHeight` and changes state into [a non-expanded state](https://github.com/mozilla/addons-frontend/blob/f7529e876f881cb5f0f7b580153cf066dbb63ae3/src/ui/components/ShowMoreCard/index.js#L44). This state change causes a re-render but the only thing changed is `className`. React does not patch up `className` changes when checking virtual DOM differences between server and client.

The [hydrate docs](https://reactjs.org/docs/react-dom.html#hydrate) suggest dealing with this by using `setState` but that's what we're doing and it's still not working. Or maybe I am misunderstanding the docs. Does anyone have ideas for how to mitigate this?

kumar303:
r+ after fixing the tests, that is

willdurand:
@kumar303 I believe the use of a div produces a different tree, hence no reconciliation optimization triggered because it is not the same "path".

willdurand:
@rebmullin any chance to rebase + prettier this patch? Would be cool to have it in this week's push.

rebmullin:
@willdurand/ @kumar303, I just rebased...how's it look now? I wasn't clear tho, do I need to do a test for this change?

willdurand:
> @willdurand / @kumar303, I just rebased...how's it look now? I wasn't clear tho, do I need to do a test for this change?

I don't think we can add a test case for that unfortunately.

PR #5104, opened 2018-05-25
Update instructions to build FF

willdurand:


docs/moz-addon-manager.md
```diff
@@ -96,94 +96,57 @@ at a URL like `localhost:3000`.
...
-
- const WEBAPI_INSTALL_HOSTS = ["addons.mozilla.org", "testpilot.firefox.com"];
+@@ -57,7 +57,7 @@ const WEBAPI_INSTALL_HOSTS = ["addons.mo
  const WEBAPI_TEST_INSTALL_HOSTS = [
    "addons.allizom.org", "addons-dev.allizom.org",
    "testpilot.stage.mozaws.net", "testpilot.dev.mozaws.net",
-   "example.com",
-+  "localhost",
+-  "example.com",
++  "example.com", "localhost",
```
rebmullin: does this part need to be updated? or if the point is to update the order, should we keep it on its own line as it was before?

docs/moz-addon-manager.md
```diff
@@ -96,94 +96,57 @@ at a URL like `localhost:3000`.
...
--    return false;
--  }
-
-   nsAutoCString host;
-   rv = uri->GetHost(host);
-   if (NS_FAILED(rv)) {
-     return false;
-   }
-
-   return IsValidHost(host);
```
rebmullin: should we leave this part in the patch - that this should be removed? I didn't see anything after this info here about removing this piece:

```javascript
@@ -57,6 +57,7 @@ IsValidHost(const nsACString& host) {
          host.LowerCaseEqualsLiteral("discovery.addons-dev.allizom.org") ||
          host.LowerCaseEqualsLiteral("testpilot.stage.mozaws.net") ||
          host.LowerCaseEqualsLiteral("testpilot.dev.mozaws.net") ||
 +        host.LowerCaseEqualsLiteral("localhost") ||
          host.LowerCaseEqualsLiteral("example.com")) {
        return true;
      }
```
see: https://monosnap.com/file/zFkJrRKzqrQfqzi3l9UW5Ny9jY17E2#embed

docs/moz-addon-manager.md
```diff
@@ -96,94 +96,57 @@ at a URL like `localhost:3000`.
...
 
 This patch will:
 1. allow you to use an HTTP connection
 2. allow any page at `localhost` (on any port) to access `mozAddonManager` and
    perform some theme actions.
 
+You can use `patch` to apply this patch:
+
+```
+patch -p1 < /path/to/patch-file.diff
```
rebmullin: this will be handy 👍

rebmullin:
@willdurand thanks for doing this : ) just a few questions above..

docs/moz-addon-manager.md
willdurand: I took the patch "as is" from my setup, it seems they have changed the formatting of this list..

docs/moz-addon-manager.md
willdurand: It is already in the code if I am not mistaken

willdurand: ah well wait, you are right actually. I thought we would not need it but I might be wrong.

willdurand:
should be better now

willdurand:
@rebmullin should be better now

docs/moz-addon-manager.md
willdurand: ah well now I understand what you meant, sorry yeah you're right, I should not have changed that.

rebmullin: @willdurand  - were you going to revert your change here? i wasn't sure...

rebmullin: or rather move to its own line?

PR #5109, opened 2018-05-25
WIP: Sort add-ons in a collection

### LINKED ISSUE #3991: Sort add-ons in a collection

Depends on: https://github.com/mozilla/addons-server/issues/8354

The user should be able to:
* View a collection
* Sort the add-ons by popularity, name, and time added

Here's how it looked on the classic site:

screenshot 2017-11-28 12 02 38

bobsilverberg:
Fixes #3991 

Depends on https://github.com/mozilla/addons-server/issues/8354

This is still a work in progress, although I think it's close. The problem (other than the fact I haven't written tests yet) is that it's not actually working.

When I select a sort order from the Select box, it does dispatch `fetchCurrentCollectionPage`, and I can see that the API code calls the API with the expected `sort` query parameter, and I can also see that `loadCurrentCollectionPage` is run after the query returns, but it doesn't update the list of add-ons on the page. I would have thought that would happen automatically after  `loadCurrentCollectionPage` completes, but apparently not.

@willdurand could you take a look and see if there's something obvious that I've not done/done wrong?

### DIFF
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -3,6 +3,7 @@ import invariant from 'invariant';
 
 import { callApi, allPages, validateLocalizedString } from 'core/api';
 import type {
+  CollectionFilters,
   ExternalCollectionAddon,
   ExternalCollectionDetail,
 } from 'amo/reducers/collections';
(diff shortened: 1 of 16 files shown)


bobsilverberg:
This has been replaced by #5414

PR #5113, opened 2018-05-25
I5011 improve collection success msg appearance

### LINKED ISSUE #5011: Improve appearance of confirmation message displayed when adding an add-on to a collection
The current implementation looks like this:

screenshot 2018-05-15 16 18 44

I think it would be better if the message was closer to the "Search" select box, and also @pwalm requested that the message fade out after 5 seconds.

rebmullin:
fixes #5011 

video
https://monosnap.com/image/52gnX55LZiRElcneQSKGCGWiJAHHwN.png

I feel like we should add some transition effect here too...

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -63,6 +63,7 @@ type Props = {|
   isCollectionBeingModified: boolean,
   page: number | null,
   router: ReactRouterType,
+  setTimeout: Function,
   siteLang: ?string,
   siteUserId: number | null,
 |};
@@ -76,6 +77,10 @@ type State = {|
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
+  static defaultProps = {
+    setTimeout: (fn) => setTimeout(fn, MESSAGE_RESET_TIME),
+  };
+
   constructor(props: Props) {
     super(props);
     this.state = this.propsToState(props);
@@ -83,25 +88,23 @@ export class CollectionManagerBase extends React.Component<Props, State> {
 
   componentWillReceiveProps(props: Props) {
     const existingId = this.props.collection && this.props.collection.id;
+    const { hasAddonBeenAdded: hasAddonBeenAddedNew } = props;
+    const { hasAddonBeenAdded } = this.props;
     if (props.collection && props.collection.id !== existingId) {
       // Only reset the form when receiving a collection that the
       // user is not already editing. This prevents clearing the form
       // in a few scenarios such as pressing the submit button.
       this.setState(this.propsToState(props));
     }
-    if (this.props.hasAddonBeenAdded !== props.hasAddonBeenAdded) {
+    if (hasAddonBeenAdded !== hasAddonBeenAddedNew) {
       this.setState({
         addonAddedStatus: props.hasAddonBeenAdded ?
           ADDON_ADDED_STATUS_SUCCESS : null,
-      }, () => {
-        if (props.hasAddonBeenAdded) {
-          setTimeout(() => {
-            this.setState({
-              addonAddedStatus: null,
-            });
-          }, MESSAGE_RESET_TIME);
-        }
       });
+    }
+
+    if (hasAddonBeenAddedNew && hasAddonBeenAddedNew !== hasAddonBeenAdded) {
+      this.props.setTimeout(this.resetMessageStatus);
     }
   }
 
@@ -236,6 +239,12 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       userId: siteUserId,
     }));
     this.setState({ addonAddedStatus: ADDON_ADDED_STATUS_PENDING });
+  };
+
+  resetMessageStatus = () => {
+    this.setState({
+      addonAddedStatus: null,
+    });
   };
 
   propsToState(props: Props) {
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -38,6 +38,8 @@ import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
+const MESSAGE_FADEOUT_TIME = 5000;
+
 export const ADDON_ADDED_STATUS_PENDING: 'ADDON_ADDED_STATUS_PENDING'
   = 'ADDON_ADDED_STATUS_PENDING';
 export const ADDON_ADDED_STATUS_SUCCESS: 'ADDON_ADDED_STATUS_SUCCESS'
@@ -89,6 +91,14 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       this.setState({
         addonAddedStatus: props.hasAddonBeenAdded ?
           ADDON_ADDED_STATUS_SUCCESS : null,
+      }, () => {
+        if (props.hasAddonBeenAdded) {
+          setTimeout(() => {
+            this.setState({
+              addonAddedStatus: null,
+            });
+          }, MESSAGE_FADEOUT_TIME);
+        }
       });
     }
   }
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -4,6 +4,7 @@ import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
+import ReactCSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
 import { compose } from 'redux';
 import config from 'config';
 
@@ -38,7 +39,8 @@ import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
-const MESSAGE_FADEOUT_TIME = 5000;
+const MESSAGE_RESET_TIME = 5000;
+const MESSAGE_FADEOUT_TIME = 450;
 
 export const ADDON_ADDED_STATUS_PENDING: 'ADDON_ADDED_STATUS_PENDING'
   = 'ADDON_ADDED_STATUS_PENDING';
@@ -97,7 +99,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
             this.setState({
               addonAddedStatus: null,
             });
-          }, MESSAGE_FADEOUT_TIME);
+          }, MESSAGE_RESET_TIME);
         }
       });
     }
@@ -327,11 +329,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
             value={this.state.slug}
           />
         </div>
-        {this.state.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
-          <Notice type="success">
-            {i18n.gettext('Added to collection')}
-          </Notice>
-        )}
+
+        <ReactCSSTransitionGroup
+          className="NoticePlaceholder"
+          component="div"
+          transitionName="overlay"
+          transitionEnterTimeout={MESSAGE_FADEOUT_TIME}
+          transitionLeaveTimeout={MESSAGE_FADEOUT_TIME}
+        >
+          {this.state.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
+            <Notice type="success">
+              {i18n.gettext('Added to collection')}
+            </Notice>
+          )}
+        </ReactCSSTransitionGroup>
+
         {!creating &&
           <AutoSearchInput
             inputName="collection-addon-query"
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -40,7 +40,7 @@ import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
-const MESSAGE_RESET_TIME = 5000;
+export const MESSAGE_RESET_TIME = 5000;
 const MESSAGE_FADEOUT_TIME = 450;
 
 export const ADDON_ADDED_STATUS_PENDING: 'ADDON_ADDED_STATUS_PENDING'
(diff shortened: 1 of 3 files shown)


bobsilverberg:
> I feel like we should add some transition effect here too...

I agree, we should use a css transition to make it fade out, rather than just disappear after 5 seconds.

rebmullin:
hey @bobsilverberg  - I've updated to use a transition. Please let me know what you think about this and some reallocation of the spacing (I've adjusted spacing here a bit b/c IMO it still looked a little odd when the space changes):

https://monosnap.com/image/mdmViit8JIIB6UPrBLlGP0s7G9dTNf.png

also - I noticed the library has updated to v2: 
https://github.com/reactjs/react-transition-group. Should I create a new issue to do an upgrade? It looks like its used in the disco pane too..

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -61,6 +64,10 @@ describe(__filename, () => {
     });
   });
 
+  afterEach(() => {
+    clock.restore();
```
bobsilverberg: If we're only using the clock in one test, is there a reason we have to do this in `afterEach`?

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -708,6 +715,17 @@ describe(__filename, () => {
...
+
+    let state = root.state();
+    expect(state.addonAddedStatus).toEqual(ADDON_ADDED_STATUS_SUCCESS);
+
+    // After 5 seconds the Notice will go away.
+    clock.tick(5010);
+
+    state = root.state();
+    expect(state.addonAddedStatus).toEqual(null);
+    root.setProps({ hasAddonBeenAdded: true });
```
bobsilverberg: I'm not sure why you're setting the state here again, and I would expect odd results because of this. Was this just a line of code left in by accident?

rebmullin:
>I assumed we'd just use css transitions manually, but if it makes it easier that's cool.

We could but I think this library helps with transitioning when the html is removed from the DOM. 

>Regarding the version, that's odd because Greenkeeper usually keeps us up to date on out-of-date dependencies. Are you sure the new version is the latest stable version?

From what i can tell, v2 has been out for a while (almost a year), altho it does say v1 is still actively maintained

here's a screenshot of before and after (to get an idea of spacing adjustments):
Alt text

Alt text

also - here is an updated video which i * think * might show a little better:
https://monosnap.com/file/5Wi6NTTC5DPQlouBmOOZxaIzJh5QD9#

I agree the transition seems a little fast. I did look at photon which I think suggests even faster but yeh I agree.. best to see what @pwalm thinks for this scenario

tests/unit/amo/components/TestCollectionManager.js
rebmullin: that's what I thought too but it seems that I need to get the updated state (when I do it this way it works :/) 
I also had to setProps too which I thought was odd but again it worked this way. Is it possible this is happening b/c i am missing some other piece of code here?

tests/unit/amo/components/TestCollectionManager.js
rebmullin: sure

rebmullin: altho - i don't think it would hurt having this here... I am just thinking if another tests is added with a clock..

I also see clock some other files where it appears only to be used 1 time and has this similar set up...

src/amo/components/CollectionManager/index.js
```diff
@@ -89,6 +93,14 @@ export class CollectionManagerBase extends React.Component<Props, State> {
       this.setState({
         addonAddedStatus: props.hasAddonBeenAdded ?
           ADDON_ADDED_STATUS_SUCCESS : null,
+      }, () => {
```
willdurand: We decided not to use the `setState` second argument in https://github.com/mozilla/addons-frontend/issues/3077.

src/amo/components/CollectionManager/index.js
```diff
@@ -89,6 +93,14 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+        if (props.hasAddonBeenAdded) {
+          setTimeout(() => {
```
willdurand: I think we could inject `setTimeout` as a default prop and pass a function that executes directly in the test cases.

willdurand:
> Regarding the version, that's odd because Greenkeeper usually keeps us up to date on out-of-date dependencies. Are you sure the new version is the latest stable version?

I think we never upgraded by fear of breaking things, but we should...

bobsilverberg:
> I think we never upgraded by fear of breaking things, but we should...

Thanks @willdurand. @rebmullin could you open a separate issue for that?

pwalm:
Looks beautiful! Ship it!

src/amo/components/CollectionManager/index.js
rebmullin: oh? ok.. let me try another way...

src/amo/components/CollectionManager/index.js
rebmullin: hey @willdurand  - do you mean something like this?

```javascript
  static defaultProps = {
    setTimeout: () => {},
  };
```

I guess I am not totally clear of the benefit of this..? let me know if i am missing something or if something like this is ok:


```javascript:
componentWillReceiveProps(props: Props) {
    ....
 
    if (hasAddonBeenAdded !== hasAddonBeenAddedNew) {
      this.setState({
        addonAddedStatus: props.hasAddonBeenAdded ?
          ADDON_ADDED_STATUS_SUCCESS : null,
      });
    }
    
    // If the message status has been set, we'll reset the message status after 5 seconds
    if (hasAddonBeenAddedNew && hasAddonBeenAddedNew !== hasAddonBeenAdded) {
      setTimeout(this.resetMessageStatus, MESSAGE_RESET_TIME);
    }
  }
```

rebmullin:
> ..@rebmullin could you open a separate issue for that?

sure @bobsilverberg. But how would we go about updating this if Greenkeeper didn't? is it just a manually change?

src/amo/components/CollectionManager/index.js
willdurand: I mean having a prop called `setTimeout` that is configured with `setTimeout` :D

In the tests, we can pass `(func) => func()` to that prop.

rebmullin:
> In the tests, we can pass (func) => func() to that prop. That way no need to deal with ticks in the tests..

hey @willdurand, I still had to use ticks so I may be missing something. Can you please take a look at my changes when you can? thanks

willdurand:
> hey @willdurand, I still had to use ticks so I may be missing something. Can you please take a look at my changes when you can? thanks

Sorry, here is what I had in mind. Note that we may have to rename the prop from `setTimeout` to `_setTimeout` (and to write `_setTimeout: setTimeout` in the `defaultProps`). The idea is to fully control the `setTimeout` without having to deal with clock/tick. We often do this for things like `window` or `config` for example.

```diff
diff --git a/src/amo/components/CollectionManager/index.js b/src/amo/components/CollectionManager/index.js
index 554d44fe3..31cf9c8e7 100644
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -78,7 +78,7 @@ type State = {|
 
 export class CollectionManagerBase extends React.Component<Props, State> {
   static defaultProps = {
-    setTimeout: (fn) => setTimeout(fn, MESSAGE_RESET_TIME),
+    setTimeout,
   };
 
   constructor(props: Props) {
@@ -104,7 +104,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
     }
 
     if (hasAddonBeenAddedNew && hasAddonBeenAddedNew !== hasAddonBeenAdded) {
-      this.props.setTimeout(this.resetMessageStatus);
+      this.props.setTimeout(this.resetMessageStatus, MESSAGE_RESET_TIME);
     }
   }
 
diff --git a/tests/unit/amo/components/TestCollectionManager.js b/tests/unit/amo/components/TestCollectionManager.js
index ff08d05b1..b8a2d7730 100644
--- a/tests/unit/amo/components/TestCollectionManager.js
+++ b/tests/unit/amo/components/TestCollectionManager.js
@@ -44,7 +44,6 @@ const simulateAutoSearchCallback = (props = {}) => {
 };
 
 describe(__filename, () => {
-  let clock;
   let fakeRouter;
   let store;
   const apiHost = config.get('apiHost');
@@ -55,7 +54,6 @@ describe(__filename, () => {
   beforeEach(() => {
     fakeRouter = createFakeRouter();
     store = dispatchClientMetadata().store;
-    clock = sinon.useFakeTimers();
     dispatchSignInActions({
       lang,
       store,
@@ -64,10 +62,6 @@ describe(__filename, () => {
     });
   });
 
-  afterEach(() => {
-    clock.restore();
-  });
-
   const getProps = ({
     collection = createInternalCollection({
       detail: createFakeCollectionDetail(),
@@ -707,7 +701,8 @@ describe(__filename, () => {
   });
 
   it('displays a notification for 5 seconds after an add-on has been added', () => {
-    const root = render({});
+    const setTimeoutSpy = sinon.stub();
+    const root = render({ setTimeout: setTimeoutSpy });
 
     expect(root.find(Notice)).toHaveLength(0);
 
@@ -715,18 +710,15 @@ describe(__filename, () => {
 
     expect(root.find(Notice)).toHaveLength(1);
     expect(root.find(Notice).children()).toHaveText('Added to collection');
+    expect(root).toHaveState('addonAddedStatus', ADDON_ADDED_STATUS_SUCCESS);
+    sinon.assert.calledWith(setTimeoutSpy, root.instance().resetMessageStatus, 5000);
 
-    const state = root.state();
-    expect(state.addonAddedStatus).toEqual(ADDON_ADDED_STATUS_SUCCESS);
-
-    const reset = () => {
-      root.setState({ addonAddedStatus: null });
-    };
-
-    // After 5 seconds the Notice will go away.
-    root.setProps(setTimeout(reset));
+    // Simulate the setTimeout behavior.
+    root.instance().resetMessageStatus();
+    // See: https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before
+    root.update();
 
-    clock.tick(6000);
+    expect(root).toHaveState('addonAddedStatus', null);
     expect(root.find(Notice)).toHaveLength(0);
   });
 
```

Does that make sense?

src/amo/components/CollectionManager/index.js
```diff
@@ -72,25 +79,40 @@ type State = {|
 |};
 
 export class CollectionManagerBase extends React.Component<Props, State> {
+  static defaultProps = {
+    setTimeout,
+    _window: typeof window !== 'undefined' ? window : {},
```
willdurand: The alternative would be to `bind` the `setTimeout` function to `window`:

```js
    setTimeout: typeof window !== 'undefined' ? window.setTimeout.bind(window) : () => {},
```

src/amo/components/CollectionManager/index.js
```diff
@@ -317,11 +345,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
             value={this.state.slug}
           />
         </div>
-        {this.state.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
-          <Notice type="success">
-            {i18n.gettext('Added to collection')}
-          </Notice>
-        )}
+
+        <ReactCSSTransitionGroup
```
willdurand: I forgot how to use this component.. do we have to keep it all the time in the UI tree? (or can we put it in the condition where `Notice` is).

src/amo/components/CollectionManager/index.js
```diff
@@ -317,11 +345,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+          className="NoticePlaceholder"
+          component="div"
+          transitionName="overlay"
```
willdurand: Is it a predefined name or custom?

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -699,15 +700,27 @@ describe(__filename, () => {
...
 
     expect(root.find(Notice)).toHaveLength(0);
 
     root.setProps({ hasAddonBeenAdded: true });
 
     expect(root.find(Notice)).toHaveLength(1);
     expect(root.find(Notice).children()).toHaveText('Added to collection');
+
+    expect(root).toHaveState('addonAddedStatus', ADDON_ADDED_STATUS_SUCCESS);
+    sinon.assert.calledWith(setTimeoutSpy, root.instance().resetMessageStatus, 5000);
```
willdurand: I guess here we can use `MESSAGE_RESET_TIME`

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -699,15 +700,27 @@ describe(__filename, () => {
...
+
+    // Simulate the setTimeout behavior.
+    root.instance().resetMessageStatus();
+    // See: https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before
```
willdurand: Maybe we can use this link with a git tag in the url: https://github.com/airbnb/enzyme/blob/enzyme%403.3.0/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before. That way it should always be valid.

src/amo/components/CollectionManager/index.js
rebmullin: No i meant to change that... thanks for catching

src/amo/components/CollectionManager/index.js
rebmullin: yes - it has to always be in the DOM to work

rebmullin:
@willdurand, thanks for the feedback. how's it look now?

PR #5119, opened 2018-05-28
Index user ID by *lowercase* username in users reducer to avoid lookup misses

### LINKED ISSUE #5118: Users reducer should store username in lowercase (in `byUsername` attribute)

Browse a user profile with a username that has been stored with capital letters in database, for example `Madonna` but use `madonna` in the URL: http://localhost:3000/en-US/firefox/user/madonna/ (in -dev).

### What happened?

The profile is not loaded and is stuck in an inconsistent state (loading mode).

The profile is successfully displayed if the user exists, otherwise 404. In this case, the user does exist.

The problem comes from the `users` reducer not storing the usernames in lowercase in the `byUsername` attribute, which implies lookup problems as the one described above.

willdurand:
Fix #5118

### DIFF
--- a/tests/unit/amo/reducers/test_users.js
+++ b/tests/unit/amo/reducers/test_users.js
@@ -63,6 +63,22 @@ describe(__filename, () => {
 
       expect(currentUserID).toEqual(null);
     });
+
+    it('indexes a loaded user ID by username', () => {
+      const state = reducer(initialState, loadUserAccount({
+        user: createUserAccountResponse({ id: 12345, username: 'JohN' }),
+      }));
+
+      expect(state.byUsername).toHaveProperty('john', 12345);
+    });
+
+    it('indexes the current user ID by username', () => {
+      const state = reducer(initialState, loadCurrentUserAccount({
+        user: createUserAccountResponse({ id: 12345, username: 'JohN' }),
+      }));
+
+      expect(state.byUsername).toHaveProperty('john', 12345);
+    });
   });
 
   describe('finishEditUserAccount', () => {
@@ -233,6 +249,15 @@ describe(__filename, () => {
       });
 
       expect(getUserByUsername(state.users, 'Biggie')).toBeUndefined();
+    });
+
+    it('is case insensitive', () => {
+      const { state } = dispatchSignInActions({
+        userProps: { id: 500, username: 'Tupac' },
+      });
+
+      expect(getUserByUsername(state.users, 'tupac'))
+        .toEqual(state.users.byID[500]);
     });
   });
 });
--- a/tests/unit/amo/reducers/test_users.js
+++ b/tests/unit/amo/reducers/test_users.js
@@ -64,7 +64,7 @@ describe(__filename, () => {
       expect(currentUserID).toEqual(null);
     });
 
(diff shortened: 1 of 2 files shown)


tests/unit/amo/reducers/test_users.js
```diff
@@ -63,6 +63,22 @@ describe(__filename, () => {
 
       expect(currentUserID).toEqual(null);
     });
+
+    it('indexes a loaded user ID by username', () => {
```
bobsilverberg: The only change you made to the reducer is to ensure that the usernames used for the key of `byUsername` is always lower case, so why do these two tests need to be added. Is this just missing coverage that you discovered while working on the patch?

willdurand: That was not covered before. If I revert my changes in the reducer, those tests will fail.

willdurand:
> This looks good, thanks @willdurand. I added a question about some of the new tests, but it doesn't block my approval. I was quite confused though, by the commit message/PR description, and also by the description in issue #5118, because they all seem to indicate that prior to this patch we were not indexing my username, but in fact we were.
> 
> This patch isn't about adding indexing, it's about ensuring that the index keys can be used in a case-insensitive manner.
> 
> I suppose once it lands the commit message will just be part of history that may never be looked at again, but I'd prefer a message that describes exactly what changes the patch makes, and I think currently it's inaccurate and confusing.

heh, thanks. Yes, I think I was already thinking about writing the description and forgot to make sure the titles made sense... Fixed in both the PR and issue.

tests/unit/amo/reducers/test_users.js
bobsilverberg: Right, but they will only fail because you're mixing the case. If you hadn't done that then similar tests would have passed without your patch. You have a different test that verifies that it's case insensitive, and these tests don't describe themselves as testing that.

willdurand: Indeed, I don't know what I was thinking this morning but the test descriptions are not really accurate... Thanks for pointing this out.

I'll update them but I want to keep them because they test the logic to add users into the state. The other test is for the selector. This selector could have been updated to find a user by a username in a case insensitive manner (by iterating over the users for instance) but I chose not to do that because I really want to store usernames in lowercase too. Hence the two different kind of tests.

bobsilverberg: I just thought of one last thing: Is it possible for there to be two users, with different userids and the same username but with different casing? E.g., a `Bob` and a `bob`? If so, wouldn't that present a potential issue with this implementation?

willdurand: Yes, I thought about that too but I assumed that would not be possible because the user API returned a user for a username with a different case (that's how I stumbled upon this bug actually).

willdurand: It has been confirmed on IRC: `username` is unique.

willdurand: ... unique in a case insensitive way, see: https://github.com/mozilla/addons-server/blob/e4d1817c934cbd700c3700d67ffd08417f54d099/src/olympia/users/forms.py#L176-L179

Issue #5000, opened 2018-05-15
Add notification settings in user profile edit page


willdurand:
Users should be able to edit their notification settings on the user profile edit page:

<img width="513" alt="notifs" src="https://user-images.githubusercontent.com/217628/40051670-e5b1ad06-583b-11e8-9d5c-5cd1ab38a151.png">

willdurand:
This feature depends on https://github.com/mozilla/addons-server/issues/8286.

willdurand:
Also I will not be able to ship this one next week (some parts might land though).

AlexandraMoga:
Verified on AMO-dev with FF60 win10x64 and Android 7.0

Users are able to select/deselect notification options when updating their profiles. Notification options are also displayed in accordance to the type of user (i.e regular user or developer) 

![notif user](https://user-images.githubusercontent.com/31961530/41045107-5a6bd59c-69b0-11e8-88ec-104f65a0e2d6.gif)


@willdurand I've noticed that the Notifications component seems to be the only one that loads somewhat independently from the rest of the sections. Also, in the loading state it appears to be loading fewer notification options. I am just wondering if there is anything that can be done to improve this aspect. On the legacy pages for example, I can't see this behavior.

willdurand:
> @willdurand I've noticed that the Notifications component seems to be the only one that loads somewhat independently from the rest of the sections.

Yes, because we have to retrieve the notifications data separately.

> Also, in the loading state it appears to be loading fewer notification options.

Yes, we chose to display only 2 "loading notifications" because users who are not developers won't have more than 2 notifications.

PR #5122, opened 2018-05-28
UI for user notifications (readonly)

### LINKED ISSUE #5127: Create a user notifications component
In order to allow a user to view/edit their notifications, we need a notifications component that is able to display such notifications with `input[type="checkbox"]` and a loading state (when notifications are not fetched from the API yet).

willdurand:
Fix #5127 – Refs #5000 – ~~Depends on #5071~~

---

It is read-only for now.

## Screenshots

Loading state: I chose to display only two notifications because that's what users who are not developers will see anyway.

screen shot 2018-05-28 at 16 25 53

List of all the notifications returned by the API:

screen shot 2018-05-28 at 16 13 08

Small screens:

screen shot 2018-05-28 at 16 13 16

RTL:

screen shot 2018-05-28 at 16 18 09

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -9,6 +9,7 @@ import { compose } from 'redux';
 
 import Link from 'amo/components/Link';
 import NotFound from 'amo/components/ErrorPage/NotFound';
+import UserProfileEditNotifications from 'amo/components/UserProfileEditNotifications';
 import UserProfileEditPicture from 'amo/components/UserProfileEditPicture';
 import {
   editUserAccount,
@@ -440,7 +441,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
               className="UserProfileEdit--Card"
               header={i18n.gettext('Profile')}
             >
-              <p className="UserProfileEdit-aside">
+              <p className="UserProfileEdit-profile-aside">
                 {isEditingCurrentUser ? i18n.gettext(
                   `Tell users a bit more information about yourself. These
                   fields are optional, but they'll help other users get to know
@@ -563,6 +564,33 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                   ].join(' '),
                 })}
               </p>
+            </Card>
+
+            <Card
+              className="UserProfileEdit--Card"
+              header={i18n.gettext('Notifications')}
+            >
+              <p className="UserProfileEdit-notifications-aside">
+                {isEditingCurrentUser ? i18n.gettext(
+                  `From time to time, Mozilla may send you email about upcoming
+                  releases and add-on events. Please select the topics you are
+                  interested in.`
+                ) : i18n.gettext(
+                  `From time to time, Mozilla may send this user email about
+                  upcoming releases and add-on events. Please select the
+                  topics this user may be interested in.`
+                )}
+              </p>
+
+              <UserProfileEditNotifications user={user} />
+
+              {isEditingCurrentUser && (
+                <p className="UserProfileEdit-notifications--help">
+                  {i18n.gettext(`Mozilla reserves the right to contact you
+                    individually about specific concerns with your hosted
+                    add-ons.`)}
+                </p>
+              )}
             </Card>
 
             <div className="UserProfileEdit-buttons-wrapper">
--- a/src/amo/components/UserProfileEditNotifications/index.js
+++ b/src/amo/components/UserProfileEditNotifications/index.js
@@ -1,0 +1,120 @@
+/* @flow */
+import makeClassName from 'classnames';
+import * as React from 'react';
+import { compose } from 'redux';
+import invariant from 'invariant';
+
+import translate from 'core/i18n/translate';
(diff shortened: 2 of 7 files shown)


src/amo/components/UserProfileEditNotifications/index.js
```diff
@@ -0,0 +1,120 @@
...
+    case 'sdk_upgrade_fail':
+      return i18n.gettext('my sdk-based add-on cannot be upgraded');
+    case 'sdk_upgrade_success':
+      return i18n.gettext('my sdk-based add-on is upgraded successfully');
+    case 'upgrade_fail':
+      return i18n.gettext(`my add-on's compatibility cannot be upgraded`);
+    case 'upgrade_success':
+      return i18n.gettext(`my add-on's compatibility is upgraded successfully`);
+    default:
+      invariant(false, `No corresponding label for notification "${name}"`);
```
willdurand: Not sure what to do here. I saw this in the react code, and I find it quite elegant but then we have a coverage miss.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -440,7 +441,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
               className="UserProfileEdit--Card"
               header={i18n.gettext('Profile')}
             >
-              <p className="UserProfileEdit-aside">
+              <p className="UserProfileEdit-profile-aside">
```
bobsilverberg: Both of these new classes, `UserProfileEdit-profile-aside` and `UserProfileEdit-notifications-aside`, point to the same css definition (which used to be just `UserProfileEdit-aside`), so why create two different classes for them? Why not just add the existing `UserProfileEdit-aside` class to the new aside you created for notifications?

src/amo/components/UserProfileEditNotifications/index.js
```diff
@@ -0,0 +1,120 @@
...
+  mandatory: boolean,
+  name: string,
+|};
+
+const createNotification = ({
+  enabled,
+  label,
+  mandatory,
+  name,
+}: CreateNotificationParams) => (
```
bobsilverberg: Can we have a return type here?

tests/unit/amo/components/TestUserProfileEditNotifications.js
```diff
@@ -0,0 +1,100 @@
...
+        user={null}
+        {...props}
+      />,
+      UserProfileEditNotificationsBase,
+    );
+  };
+
+  it('renders loading notifications without a user', () => {
+    const name = 'some-input-name';
+    const root = render({ name, user: null });
```
bobsilverberg: I don't see a `name` prop on `UserProfileEditNotifications`.

src/amo/components/UserProfileEditNotifications/index.js
bobsilverberg: Here's an idea: change `createUserNotificationsResponse` to accept an array which can override the array which is returned by default. Then write a test where you create an array of notifications with a "bad" name (it can be the only notification in the array), and pass it into `createUserNotificationsResponse` and use the resulting array of notifications in a call to `loadUserNotifications`. When you render the component that should trigger the error and give us coverage for this line.

In fact it may be overkill to do all of that with `createUserNotificationsResponse` because all you're doing is passing in an array and expecting the same array back, but I suppose if something were to change in the future about the response then it would be nice to have it encapsulated there.

src/amo/components/UserProfileEdit/index.js
willdurand: I want to be able to use them in the test suite (I think).

src/amo/components/UserProfileEditNotifications/index.js
willdurand: @bobsilverberg thanks for the review! You're right, I could get coverage for this with unknown notifications, but what should be the behavior here? `invariant` only? Throwing an exception?

Same question for the UI: if we have no label, then what? We don't display anything for that notification?

That's why I kept things simple, because `invariant` will trigger a warning for us (developers) and I believe not returning anything will just not display a label for that notification (and that won't break the UI entirely).

bobsilverberg: Well, in the test I would expect `invariant` to throw an exception, which we would assert (although we don't usually include assertions for calls to `invariant`, I think this is an exception (pun inteneded!)).

In terms of the display, I guess if we have no label we probably shouldn't be showing the notification input at all. It's bad not to show an input when there's a field for it, but I don't see any point in showing an input without a label. So I think the code for `createNotification` should return nothing if there is no label. Do you agree?

willdurand: I do agree. Then, next question is: I can maybe return `null` in the `default` case and log a warning in `createNotification` to let developers know why one notification is missing. WDYT? That would be testable as usual and we would not have to introduce a test case to check `invariant`.

bobsilverberg: Sure, that sounds good.

PR #5128, opened 2018-05-29
Allow to update user notifications

### LINKED ISSUE #5000: Add notification settings in user profile edit page

willdurand:
Fix #5000 – ~~Depends on #5122~~

---

This PR allows to update user notifications.

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -32,7 +32,11 @@ import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
 import Notice from 'ui/components/Notice';
 import OverlayCard from 'ui/components/OverlayCard';
-import type { UsersStateType, UserType } from 'amo/reducers/users';
+import type {
+  NotificationsUpdateType,
+  UsersStateType,
+  UserType,
+} from 'amo/reducers/users';
 import type { ApiStateType } from 'core/reducers/api';
 import type { DispatchFunc } from 'core/types/redux';
 import type { ErrorHandlerType } from 'core/errorHandler';
@@ -64,6 +68,7 @@ type FormValues = {|
   displayName: string | null,
   homepage: string | null,
   location: string | null,
+  notifications: NotificationsUpdateType,
   occupation: string | null,
   picture: File | null,
   username: string,
@@ -239,6 +244,20 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     }
   }
 
+  onNotificationChange = (event: SyntheticEvent<HTMLInputElement>) => {
+    event.stopPropagation();
+
+    const { name, checked } = event.currentTarget;
+
+    this.setState((prevState) => ({
+      notifications: {
+        ...prevState.notifications,
+        [name]: checked,
+      },
+      successMessage: null,
+    }));
+  }
+
   onPictureDelete = (event: SyntheticEvent<HTMLButtonElement>) => {
     event.preventDefault();
 
@@ -272,6 +291,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
       displayName,
       homepage,
       location,
+      notifications,
       occupation,
       picture,
       username,
@@ -281,6 +301,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
 
     dispatch(editUserAccount({
       errorHandlerId: errorHandler.id,
+      notifications,
       picture,
       userFields: {
         biography,
@@ -300,6 +321,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
       displayName: '',
       homepage: '',
       location: '',
+      notifications: {},
       occupation: '',
       picture: null,
       username: this.props.username,
@@ -627,7 +649,10 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                 )}
               </p>
 
-              <UserProfileEditNotifications user={user} />
+              <UserProfileEditNotifications
+                onChange={this.onNotificationChange}
+                user={user}
+              />
 
               {isEditingCurrentUser && (
                 <p className="UserProfileEdit-notifications--help">
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -165,6 +165,13 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     }
 
     if (wasUpdating && !isUpdating && !errorHandler.hasError()) {
+      if (newUser && !newUser.notifications) {
+        dispatch(fetchUserNotifications({
+          errorHandlerId: errorHandler.id,
+          username: newUser.username,
+        }));
+      }
+
       this.setState({
         pictureData: null,
         successMessage: i18n.gettext('Profile successfully updated'),
--- a/src/amo/components/UserProfileEditNotifications/index.js
+++ b/src/amo/components/UserProfileEditNotifications/index.js
@@ -48,6 +48,7 @@ type CreateNotificationParams = {|
   label: React.Element<typeof LoadingText> | string | null,
   mandatory: boolean,
   name: string,
+  onChange?: Function,
 |};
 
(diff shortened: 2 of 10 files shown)


tests/unit/amo/sagas/test_users.js
```diff
@@ -276,6 +338,10 @@ describe(__filename, () => {
   describe('fetchUserNotifications', () => {
     it('calls the API to fetch the notifications of a user', async () => {
       const username = 'tofumatt';
+
+      const user = createUserAccountResponse({ username });
+      sagaTester.dispatch(loadCurrentUserAccount({ user }));
```
willdurand: I have added this because I saw an `invariant` error in the logs (but it did not fail the test case because the test case).

willdurand:
@bobsilverberg rebased this PR, ready for a review :)

src/amo/components/UserProfileEdit/index.js
```diff
@@ -194,6 +199,20 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     }
   }
 
+  onNotificationChange = (event: SyntheticEvent<HTMLInputElement>) => {
+    event.stopPropagation();
+
+    const { name, checked } = event.currentTarget;
+
+    this.setState((prevState) => ({
```
bobsilverberg: I didn't know you could do this with `setState`. Nice.

src/amo/components/UserProfileEditNotifications/index.js
```diff
@@ -48,13 +48,15 @@ type CreateNotificationParams = {|
   label: React.Element<typeof LoadingText> | string | null,
   mandatory: boolean,
   name: string,
+  onChange?: Function,
```
bobsilverberg: Sometimes instead of just `Function`, we specify the type using the function signature, so Flow knows exactly what the function should look like.

src/amo/reducers/users.js
```diff
@@ -141,7 +143,8 @@ export const finishEditUserAccount = (): FinishEditUserAccountAction => {
 
 type EditUserAccountParams = {|
   errorHandlerId: string,
-  picture?: File | null,
+  notifications: NotificationsUpdateType,
+  picture: File | null,
```
bobsilverberg: Why does this patch include this change? It doesn't seem to relate to notifications. Is it just an unrelated problem you found when writing this patch?

src/amo/reducers/users.js
```diff
@@ -152,15 +155,17 @@ type EditUserAccountAction = {|
 |};
 
 export const editUserAccount = ({
-  errorHandlerId, picture, userFields, userId,
+  errorHandlerId, notifications, picture, userFields, userId,
 }: EditUserAccountParams): EditUserAccountAction => {
   invariant(errorHandlerId, 'errorHandlerId is required');
+  invariant(notifications, 'notifications are required');
+  invariant(picture !== undefined, 'picture is required');
```
bobsilverberg: Why do we need to check explicitly for `picture !== undefined`?

src/amo/sagas/users.js
```diff
@@ -13,6 +13,7 @@ import {
   currentUserAccount as currentUserAccountApi,
   deleteUserPicture as deleteUserPictureApi,
   editUserAccount as editUserAccountApi,
+  updateUserNotifications as updateUserNotificationsApi,
```
bobsilverberg: Nit: This just jumped out at me now, I'm not sure why, maybe because I have also been working on an api file. In some other saga files, e.g., collections, we just do `import * as api from 'amo/api/collections';` and then call them in the code like `api.getCollectionDetail`. This seems a bit cleaner than having to create all these aliases.

src/amo/sagas/users.js
```diff
@@ -62,6 +64,19 @@ export function* editUserAccount({
     });
 
     yield put(loadUserAccount({ user }));
+
+    if (Object.keys(notifications).length) {
```
bobsilverberg: You could also just do `if (notifications !== {})`, I think, although I'm not sure that's any better.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -562,6 +565,7 @@ describe(__filename, () => {
 
     sinon.assert.calledWith(dispatchSpy, editUserAccount({
       errorHandlerId: errorHandler.id,
+      notifications: {},
```
bobsilverberg: There doesn't seem to be a test that covers sending a new notifications object to `editUserAccount`. It could maybe be added to this test case, or perhaps you want to add a new test.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -582,7 +604,10 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                 )}
               </p>
 
-              <UserProfileEditNotifications user={user} />
+              <UserProfileEditNotifications
+                onChange={this.onNotificationChange}
```
bobsilverberg: This doesn't seem to be covered in a test. Should there be a test that verifies that `UserProfileEditNotifications` received the expected `onChange` property?

src/amo/reducers/users.js
willdurand: it can be null

src/amo/reducers/users.js
willdurand: I realized picture was always present, no need to mark it as optional.

src/amo/components/UserProfileEdit/index.js
willdurand: it is covered by "stores updated notifications in state", which makes sure the right behavior is triggered when this prop gets executed.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: ah right, thank you. We test that notifications get passed and we test `editUserAccount` on its own though, but one more test won't hurt :+1:

src/amo/sagas/users.js
willdurand: That makes sense.

src/amo/sagas/users.js
willdurand: mmh for some reasons, I have to think more when reading `notifications !== {}` than `.length`, because we use `.length` for arrays and strings already.

src/amo/components/UserProfileEditNotifications/index.js
willdurand: I think that would make sense for custom functions, but here we only pass a React event handler. That being said, I could add a type like `(e: SyntheticEvent<HTMLInputElement>) => void` but I don't know if that would be super useful.

willdurand:
@bobsilverberg updated the PR, thanks for the review!

willdurand:
@bobsilverberg I had to rebase this PR again, looks reviewable now 😛

tests/unit/amo/sagas/test_users.js
```diff
@@ -203,6 +206,63 @@ describe(__filename, () => {
...
+      const allNotifications = createUserNotificationsResponse();
+      allNotifications[0].enabled = notifications.reply;
+
+      const userFields = {
+        biography: 'I fell into a burning ring of fire.',
+        location: 'Folsom Prison',
+      };
+
+      mockApi
+        .expects('editUserAccount')
```
bobsilverberg: I know this is old code, and I have no idea why I'm just noticing this now, but it seems odd to have a function/action called `editUserAccount`. Shouldn't it be `updateUserAccount`? This is not something that should be addressed in this patch, but I wanted to raise it as a possible code quality issue. What do you think?

willdurand: We never use `updateUserAccount` or `UPDATE_USER_*` for the user profile part. I think that would make sense though, because I introduced some `updateXXX` methods for user notifications for instance.

willdurand: See: https://github.com/mozilla/addons-frontend/issues/5195

Issue #4279, opened 2018-01-31
Allow user to delete their profile


tofumatt:
We need API/saga/reducer code to allow a user to delete their profile.

Note: I think this button might want to be red or something more dangerous-looking. Not sure.

<img width="513" alt="screenshot 2018-01-31 14 36 44" src="https://user-images.githubusercontent.com/90871/35628602-51f449c8-0694-11e8-848b-6142c7dc7a28.png">

willdurand:
@pwalm quick & dirty:

<img width="1552" alt="screen shot 2018-05-29 at 18 41 46" src="https://user-images.githubusercontent.com/217628/40675548-fa92f31c-636f-11e8-9301-05a2818a793f.png">

Any comments? (Note: I don't really like this modal, the overlay blue is weird...)

bobsilverberg:
I agree that that looks really weird, and as if some malware has taken over the user's computer!

willdurand:
I revisited the modal with better UI and content (which can still be reworked later):

![2018-05-30 11 53 44](https://user-images.githubusercontent.com/217628/40716259-3a5b2a6a-6400-11e8-85ee-6cb140b5520f.gif)

PR #5131, opened 2018-05-29
Add reducer/saga/action/api to delete a user account

willdurand:
Refs #4279

---

Not sure whether I will do everyting in this PR or I will separate the
plumbing and the UI (as I've started here).

### DIFF
--- a/src/amo/api/users.js
+++ b/src/amo/api/users.js
@@ -89,3 +89,18 @@ export function deleteUserPicture({ api, userId }: {|
     state: api,
   });
 }
+
+export function deleteUserAccount({ api, userId }: {|
+  api: ApiStateType,
+  userId: UserId,
+|}) {
+  invariant(api, 'api state is required.');
+  invariant(userId, 'userId is required.');
+
+  return callApi({
+    auth: true,
+    endpoint: `accounts/account/${userId}`,
+    method: 'DELETE',
+    state: api,
+  });
+}
--- a/src/amo/reducers/users.js
+++ b/src/amo/reducers/users.js
@@ -22,6 +22,8 @@ export const LOAD_USER_ACCOUNT: 'LOAD_USER_ACCOUNT' = 'LOAD_USER_ACCOUNT';
 export const DELETE_USER_PICTURE: 'DELETE_USER_PICTURE' = 'DELETE_USER_PICTURE';
 export const FETCH_USER_NOTIFICATIONS: 'FETCH_USER_NOTIFICATIONS' = 'FETCH_USER_NOTIFICATIONS';
 export const LOAD_USER_NOTIFICATIONS: 'LOAD_USER_NOTIFICATIONS' = 'LOAD_USER_NOTIFICATIONS';
+export const DELETE_USER_ACCOUNT: 'DELETE_USER_ACCOUNT' = 'DELETE_USER_ACCOUNT';
+export const UNLOAD_USER_ACCOUNT: 'UNLOAD_USER_ACCOUNT' = 'UNLOAD_USER_ACCOUNT';
 
 export type UserId = number;
 
@@ -201,6 +203,50 @@ export const loadUserAccount = ({
   return {
     type: LOAD_USER_ACCOUNT,
     payload: { user },
+  };
+};
+
+export type DeleteUserAccountParams = {|
+  errorHandlerId: string,
+  userId: UserId,
+|};
+
+type DeleteUserAccountAction = {|
+  type: typeof DELETE_USER_ACCOUNT,
+  payload: DeleteUserAccountParams,
+|};
+
+export const deleteUserAccount = (
+  { errorHandlerId, userId }: DeleteUserAccountParams
+): DeleteUserAccountAction => {
+  invariant(userId, 'userId is required');
+
+  return {
+    type: DELETE_USER_ACCOUNT,
+    payload: {
+      errorHandlerId,
+      userId,
+    },
+  };
+};
+
+type UnloadUserAccountParams = {|
+  userId: UserId,
+|};
+
+type UnloadUserAccountAction = {|
+  type: typeof UNLOAD_USER_ACCOUNT,
+  payload: UnloadUserAccountParams,
+|};
+
+export const unloadUserAccount = (
+  { userId }: UnloadUserAccountParams
+): UnloadUserAccountAction => {
+  invariant(userId, 'userId is required');
+
+  return {
+    type: UNLOAD_USER_ACCOUNT,
+    payload: { userId },
   };
 };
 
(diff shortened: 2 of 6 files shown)


src/amo/reducers/users.js
```diff
@@ -22,6 +22,8 @@ export const LOAD_USER_ACCOUNT: 'LOAD_USER_ACCOUNT' = 'LOAD_USER_ACCOUNT';
 export const DELETE_USER_PICTURE: 'DELETE_USER_PICTURE' = 'DELETE_USER_PICTURE';
 export const FETCH_USER_NOTIFICATIONS: 'FETCH_USER_NOTIFICATIONS' = 'FETCH_USER_NOTIFICATIONS';
 export const LOAD_USER_NOTIFICATIONS: 'LOAD_USER_NOTIFICATIONS' = 'LOAD_USER_NOTIFICATIONS';
+export const DELETE_USER_ACCOUNT: 'DELETE_USER_ACCOUNT' = 'DELETE_USER_ACCOUNT';
+export const REMOVE_USER_ACCOUNT: 'REMOVE_USER_ACCOUNT' = 'REMOVE_USER_ACCOUNT';
```
willdurand: @bobsilverberg would you have an idea on how to improve this? "DELETE" comes from the user action (and also from the API call), so I like to have a `DELETE_USER_ACCOUNT` dispatched when user wants to remove their profile.

Yet, I also need an action to remove the user data from the state when the deletion has completed, hence the `REMOVE` because I want to remove the user data.

Given that we use `LOAD` to add a user to the state, can we maybe use `UNLOAD`?

bobsilverberg: Hmm, `UNLOAD` does make sense, as we're unloading it from state. I was also going to say that maybe it could be xxx_USER_DATA as opposed to xxx_USER_ACCOUNT, because we're removing data from state, as opposed to removing an actual account, but yes, `UNLOAD` probably makes the most sense in this context.

willdurand:
pinging @rebmullin too because it is time for bigger reviews now 😉

src/amo/api/users.js
```diff
@@ -89,3 +89,18 @@ export function deleteUserPicture({ api, userId }: {|
     state: api,
   });
 }
+
+export function deleteUserAccount({ api, userId }: {|
+  api: ApiStateType,
+  userId: UserId,
+|}) {
```
bobsilverberg: I just noticed that we don't have any return types for any of these functions, not even the ones that are expected to return something, and I think that's something we should fix. Could be a follow-up to this as it's unrelated to this patch.

src/amo/reducers/users.js
```diff
@@ -204,6 +206,50 @@ export const loadUserAccount = ({
...
+
+type DeleteUserAccountAction = {|
+  type: typeof DELETE_USER_ACCOUNT,
+  payload: DeleteUserAccountParams,
+|};
+
+export const deleteUserAccount = (
+  { errorHandlerId, userId }: DeleteUserAccountParams
+): DeleteUserAccountAction => {
+  invariant(userId, 'userId is required');
```
bobsilverberg: Shouldn't we also have an `invariant` for `errorHandlerId`?

src/amo/reducers/users.js
```diff
@@ -449,6 +495,20 @@ const reducer = (
...
+    case UNLOAD_USER_ACCOUNT: {
+      const { userId } = action.payload;
+
+      const newState = { ...state };
+
+      if (newState.byID[userId]) {
+        const { username } = newState.byID[userId];
+
+        delete newState.byID[userId];
+        delete newState.byUsername[username];
```
bobsilverberg: Should `currentUserID` also be set to `null` in here, or is that done elsewhere, prior to calling `UNLOAD_USER_ACCOUNT`?

tests/unit/amo/api/test_users.js
```diff
@@ -166,4 +167,24 @@ describe(__filename, () => {
       mockApi.verify();
     });
   });
+
+  describe('deleteUserAccount', () => {
+    it('deletes a user profile', async () => {
+      const state = dispatchSignInActions().store.getState();
```
bobsilverberg: Do we need to sign in for this test, or could you just create a `const` for `userId` and use that?

tests/unit/amo/reducers/test_users.js
```diff
@@ -163,6 +164,28 @@ describe(__filename, () => {
...
   });
 
+  describe('unloadUserAccount', () => {
+    it('unloads a user from the state given a user ID', () => {
+      const userId = 12345;
+      const username = 'john';
+
+      const user = createUserAccountResponse({ id: userId, username });
+      const prevState = reducer(initialState, loadUserAccount({ user }));
+
```
bobsilverberg: Would it be worth adding some check assertions here for `state.byID[userId]` and `state.byUsername[username]` to verify that the call to `loadUserAccount` successfully loaded the data? Otherwise the assertions at the end could be yielding false positives.

src/amo/reducers/users.js
willdurand: interesting that you noticed this here, I completely forgot when writing this patch. it is done in the next patch though, can we keep things as is here or do you want me to update the code here?

src/amo/api/users.js
willdurand: Do we have a type for that? or is it just a generic `Promise`? I'll file an issue.

tests/unit/amo/api/test_users.js
willdurand: we need the api state too.

src/amo/api/users.js
bobsilverberg: In the collections API we use `Promise<void>` if we do not expect a return value, and `Promise<type>` if we do, e.g., `Promise<ExternalCollectionDetail>`.

src/amo/reducers/users.js
bobsilverberg: I think it's fine to land this as is and accept the change in the next patch (which I noticed). 😄

tests/unit/amo/api/test_users.js
bobsilverberg: In other tests we use `dispatchClientMetadata().store.getState()` to get the state, which I suppose is six of one, half a dozen of the other, but if we don't explicitly need a signed in user, maybe we should use the former?

willdurand: I see. This action requires authentication so I'd keep a signed user because of that.

src/amo/api/users.js
willdurand: See: https://github.com/mozilla/addons-frontend/issues/5177

src/amo/reducers/users.js
willdurand: of course... sorry!

tests/unit/amo/reducers/test_users.js
willdurand: Added.

willdurand:
Thanks for the review!

willdurand:
Thank you ❤️

PR #5133, opened 2018-05-29
Fixed height issues with Screenshots section

Vimal-Raghubir:
Changed the display of the ScreenShots by extending the height of the psw-thumbnails section to 220px and same for the viewport height as recommended by @rebmullin . This is related to issue #4887. 

More details regarding the fix:
After testing I realized the two suggestions work hand in hand. They are both based on the height variable defined earlier in that file as 200px. So changing this to 220px changes both values and this resolves the display issue. However, the screenshot image's height is also reliant on this variable so when this value is changed to 220px it does not fix the issue. So the suggested fix is to make this image's height be 20px less of the variable's height and change the variable value to 220px. This 20px less fix is calculated at runtime by performing this operation on a new variable $image-height as opposed to hardcoding it. :+1

### DIFF
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -1,5 +1,6 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
+
 $screenshot-height: 220px;
 $image-height: $screenshot-height - 20px;
 $screenshot-width: 320px;
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -1,6 +1,7 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
 
+//image-height has to be 20px smaller to ensure it doesn't overflow in the screenshots section.
 // See: https://github.com/mozilla/addons-frontend/issues/4887
 $image-height: $screenshot-height - 20px;
 $screenshot-height: 220px;
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -1,6 +1,7 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-
+// screenshot-height value was increased to handle issue #4887
+// image-height is used to ensure that the image's height is always 20px less than screenshot-height for #4887 fix.
 $screenshot-height: 220px;
 $image-height: $screenshot-height - 20px;
 $screenshot-width: 320px;
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -1,6 +1,7 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-$screenshot-height: 200px;
+$screenshot-height: 220px;
+$image-height: $screenshot-height - 20px;
 $screenshot-width: 320px;
 
 .ScreenShots .pswp-thumbnails {
@@ -34,7 +35,7 @@ $screenshot-width: 320px;
 .ScreenShots-image {
   cursor: pointer;
   display: inline-block;
-  height: $screenshot-height;
+  height: $image-height;
   width: $screenshot-width;
 }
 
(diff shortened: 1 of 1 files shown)


Vimal-Raghubir:
New changes have been tested in an SCSS validator as indicated by the screenshot below.
scss

src/amo/css/ScreenShots.scss
```diff
@@ -1,6 +1,8 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-$screenshot-height: 200px;
+
+$screenshot-height: 220px;
+$image-height: $screenshot-height - 20px;
```
rebmullin: could you please add comment here so its clear why we're doing this -  maybe describe the issue and add a reference to the issue

src/amo/css/ScreenShots.scss
```diff
@@ -1,6 +1,11 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-$screenshot-height: 200px;
+// screenshot-height value was increased for 
+// https://github.com/mozilla/addons-frontend/issues/4887
+// image-height ensures the image is always 20px less than screenshot-height
+// 20px accounts for the approx size of the horizontal scrollbar
```
willdurand: Could you only write `See: https://github.com/mozilla/addons-frontend/issues/4887` as comment? Preferably above the `image-height` var. Thanks!

Vimal-Raghubir: So remove all the other comments?

willdurand: yes, keep it simple :)

src/amo/css/ScreenShots.scss
```diff
@@ -1,6 +1,9 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-$screenshot-height: 200px;
+
+// See: https://github.com/mozilla/addons-frontend/issues/4887
```
rebmullin: I defer to @willdurand here but IMHO I think it's a little better to have *some* description so you really don't have to dig into the issue

willdurand: Sorry I missed your previous comment about the original description. I found it too verbose and not really useful, so if we can come up with something short & accurate, I'd say let's add a comment above the `See: ...`. Thanks for raising this point @rebmullin

Vimal-Raghubir: How about: image-height has to be 20px smaller to ensure it doesn't overflow in the screenshots section.

willdurand: I like it, thanks @Vimal-Raghubir.

Please, don't forget the period at the end because we tend to write (syntactically) correct English sentences in our comments.

src/amo/css/ScreenShots.scss
```diff
@@ -1,6 +1,10 @@
 @import "~core/css/inc/mixins";
 @import "~amo/css/inc/vars";
-$screenshot-height: 200px;
+
+//image-height has to be 20px smaller to ensure it doesn't overflow in the screenshots section.
+// See: https://github.com/mozilla/addons-frontend/issues/4887
+$image-height: $screenshot-height - 20px;
+$screenshot-height: 220px;
```
rebmullin: @Vimal-Raghubir  - I just noticed this but screenshot-height should come before you use it on line 6

rebmullin:
@Vimal-Raghubir,  looks good, thanks : )

caitmuenster:
Thanks so much, @Vimal-Raghubir! Your contribution has been added to our r[ecognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#May_2018). 

Would you be interested in creating an account on [mozillians.org](https://mozillians.org)? I'd be happy to vouch for you!

Vimal-Raghubir:
Hi @caitmuenster . Just signed up an account on mozillians.org with my github account. Thank you :)

caitmuenster:
Yay! You are now vouched. :) 

Welcome onboard! I look forward to seeing you around the project!

PR #5140, opened 2018-05-30
Delete account UI

willdurand:
Fix mozilla/addons-server#4279 – depends on mozilla/addons-server#5131 

---

This PR adds the UI to actually delete a user profile. Users are redirected to the home page in all cases, and users who delete their own profiles are logged-out. Copy can be reworked later. There is a known (minor) issue related to cookies, see: https://github.com/mozilla/addons/issues/5548.

The copy is:

> Attention: you are about to delete your profile. Are you sure?
>
> If you confirm this irreversible action, the following data will be removed: profile picture, profile details (including username, email, display name, location, home page, biography, occupation) and notification preferences. Other data such as ratings and reviews will be anonymized.
>
> Important: if you own add-ons, you have to transfer them to other users or to delete them before you can delete your profile.
>
> Cancel | Yes, delete my profile

## Screenshots

screen shot 2018-05-30 at 16 24 25

Warning for regular users:

screen shot 2018-05-30 at 16 24 28

Warning for admins:

screen shot 2018-05-30 at 16 24 13

RTL:

screen shot 2018-05-30 at 16 25 08

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -700,13 +700,16 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                 ), ['strong'])
               }
             />
-            {isEditingCurrentUser && (
-              <p>
-                {i18n.gettext(`Important: if you own add-ons, you have to
+            <p>
+              {isEditingCurrentUser ? (
+                i18n.gettext(`Important: if you own add-ons, you have to
                   transfer them to other users or to delete them before you
-                  can delete your profile.`)}
-              </p>
-            )}
+                  can delete your profile.`)
+              ) : (
+                i18n.gettext(`Important: a user profile can only be deleted if
+                  the user does not own any add-ons.`)
+              )}
+            </p>
             <div className="UserProfileEdit-buttons-wrapper">
               <Button
                 buttonType="alert"
--- a/tests/unit/amo/components/TestUserProfileEdit.js
+++ b/tests/unit/amo/components/TestUserProfileEdit.js
@@ -1081,8 +1081,9 @@ describe(__filename, () => {
       'Attention: You are about to delete a profile. Are you sure?'
     );
 
-    expect(root.find('.UserProfileEdit-deletion-modal').find('p'))
-      .toHaveLength(1);
+    expect(root.find('.UserProfileEdit-deletion-modal').find('p').at(1))
+      .toHaveText(oneLine`Important: a user profile can only be deleted if the
+        user does not own any add-ons.`);
 
     expect(root.find('.UserProfileEdit-confirm-button').children())
       .toHaveText('Yes, delete this profile');


pwalm:
Looks great! The only thing is that we need to [follow the overlay spec in the design system](https://design.firefox.com/photon/patterns/overlay.html) for the modal.

pwalm:
@MeridelW Can you please take a look at the copy in this "delete my account" modal and make sure it follows our content guidelines?

willdurand:
> Looks great! The only thing is that we need to follow the overlay spec in the design system for the modal.

:heart: Finally, we will get rid of this weird blue!

willdurand:
Overlay updated in https://github.com/mozilla/addons-frontend/pull/5149

src/amo/components/UserProfileEdit/index.js
```diff
@@ -173,6 +178,46 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+    dispatch(deleteUserAccount({
+      errorHandlerId: errorHandler.id,
+      userId: user.id,
+    }));
+
+    if (currentUser.id === user.id) {
+      dispatch(logOutUser());
+    }
+
+    router.push(`/${lang}/${clientApp}`);
```
bobsilverberg: Would it be nice to have a route that we could push that just says "display the home page", as opposed to having to construct one using `lang` and `clientApp`? It just seems like that shouldn't be necessary.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -610,9 +655,79 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+            ) : i18n.gettext(
+              `Attention: you are about to delete a profile. Are you sure?`
+            )}
+            onEscapeOverlay={this.onCancelProfileDeletion}
+            visibleOnLoad
+          >
+            <p
+              // eslint-disable-next-line react/no-danger
+              dangerouslySetInnerHTML={
+                sanitizeHTML(i18n.sprintf(
```
bobsilverberg: I guess we need this in case a translator tries to sneak some bad html into a translation?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -610,9 +655,79 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+              </Button>
             </div>
           </div>
         </form>
+
+        {this.state.showProfileDeletionModal && (
+          <OverlayCard
+            className="UserProfileEdit-deletion-modal"
+            header={isEditingCurrentUser ? i18n.gettext(
+              `Attention: you are about to delete your profile. Are you sure?`
```
bobsilverberg: It seems like the "y" in "you" should be capitalized as it's the beginning of a sentence.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -610,9 +655,79 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+            ) : i18n.gettext(
+              `Attention: you are about to delete a profile. Are you sure?`
```
bobsilverberg: As above, should "y" be "Y"?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -610,9 +655,79 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+                    notification preferences. Other data such as ratings
+                    and reviews will be anonymized.`
+                  ), {
+                    startStrong: '<strong>',
+                    endStrong: '</strong>',
+                  }
+                ), ['strong'])
+              }
+            />
+            {isEditingCurrentUser && (
```
bobsilverberg: Shouldn't we also warn admins that they cannot delete profiles of other users who currently have add-ons? Or do we just expect them to know that?

src/amo/reducers/users.js
```diff
@@ -505,6 +505,10 @@ const reducer = (
 
         delete newState.byID[userId];
         delete newState.byUsername[username];
+
+        if (newState.currentUserID === userId) {
+          newState.currentUserID = null;
```
bobsilverberg: Ah, there it is. That was a comment I made on PR #5131.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1002,6 +1025,184 @@ describe(__filename, () => {
...
+    sinon.assert.notCalled(dispatchSpy);
+
+    // User confirms account deletion.
+    root.find('.UserProfileEdit-confirm-button').simulate(
+      'click',
+      createFakeEvent()
+    );
+
+    sinon.assert.callCount(dispatchSpy, 1);
+    sinon.assert.calledWith(dispatchSpy, deleteUserAccount({
```
bobsilverberg: Just to be explicit, do you want to also do an assert.notCalled, or is that overkill?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -173,6 +178,46 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+    invariant(currentUser, 'currentUser is required');
+    invariant(user, 'user is required');
+
+    dispatch(deleteUserAccount({
+      errorHandlerId: errorHandler.id,
+      userId: user.id,
+    }));
+
+    if (currentUser.id === user.id) {
+      dispatch(logOutUser());
```
bobsilverberg: Because you've updated `unloadUserAccount` to also set `currentUserID` to `null`, do we even need to dispatch `logOutUser` at all? All it does it set `currentUserID` to null.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1002,6 +1025,184 @@ describe(__filename, () => {
...
+    const root = renderUserProfileEdit({ store });
+
+    expect(root.find('.UserProfileEdit-deletion-modal')).toHaveLength(0);
+
+    root.find('.UserProfileEdit-delete-button').simulate(
+      'click',
+      createFakeEvent()
+    );
+
+    expect(root.find('.UserProfileEdit-deletion-modal')).toHaveLength(1);
```
bobsilverberg: It seems like it would be worth including the assertion for `preventDefault` as you did for the delete button on the main form.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1002,6 +1025,184 @@ describe(__filename, () => {
...
+
+    dispatchSpy.reset();
+
+    // User opens the modal.
+    root.find('.UserProfileEdit-delete-button').simulate(
+      'click',
+      createFakeEvent()
+    );
+
+    sinon.assert.notCalled(dispatchSpy);
```
bobsilverberg: As above, can we assert about `preventDefault` here too?

src/amo/components/UserProfileEdit/index.js
willdurand: I don't think we could do better for now, but we should definitely investigate once we upgrade React Router or move to something else (https://github.com/mozilla/addons-frontend/issues/2161) .

src/amo/components/UserProfileEdit/index.js
willdurand: Is it? I trust you on this because I often mix British, American and Broken English but I know that we don't always capitalize the first word after a colon.

src/amo/components/UserProfileEdit/index.js
willdurand: Indeed, and that's what we do everywhere when we need to insert HTML tags in translations.

src/amo/components/UserProfileEdit/index.js
willdurand: I was thinking that admins would be aware but maybe we could have a message for them.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: With both `callCount` and `calledWith`, I thought that it would be enough.

src/amo/components/UserProfileEdit/index.js
willdurand: Yes, because `logOutUser` is also used here: https://github.com/mozilla/addons-frontend/blob/c74668240dc8a472837e405b85f0977f7b027ba9/src/core/reducers/api.js#L113-L114

src/amo/components/UserProfileEdit/index.js
willdurand: Indeed! We use this pattern for translation strings including HTML (links most of the time).

src/amo/components/UserProfileEdit/index.js
bobsilverberg: I don't actually know for sure, that was just an assumption on my part. Maybe we should ask @MeridelW for a final sign-off on a final set of screen shots?

willdurand: @bobsilverberg yes, I will open a new issue and assign @MeridelW to it :)

src/amo/components/UserProfileEdit/index.js
bobsilverberg: You say "maybe" above, but you didn't change it. I'm fine either way - just wanted to raise it as something to consider.

willdurand:
@bobsilverberg thanks! I will definitely add the alternative text for admins. I cannot merge this PR yet though as it depends on #5131 😃

PR #5143, opened 2018-05-30
Consistently use username as the user identifier in collections reducers and sagas

### LINKED ISSUE #5142: Consistently use one user identifier in collections reducers and sagas
While working on the "Delete a collection" feature, I discovered that we are inconsistently using either the string `username` or the numeric `userId` in different reducer and saga methods, and I see no reason to do so. All of the APIs are designed to accept either a numeric `userId` or a string `username`, and both are guaranteed to be unique to an individual user. We should update our code to either always use `userId` or always use `username` to remove the inconsistency and resulting confusion in the codebase.

bobsilverberg:
Fixes #5142 

I decided to use the string `username` as the identifier in all cases, rather than the numeric `userId`, mainly because of this [comment](https://github.com/mozilla/addons-frontend/blob/38c186ed9a33f950db1ebbeb09b6cb386fd9d1fd/src/amo/api/collections.js#L133), which has to do with the fact that we sometimes redirect to a URL and want that URL to be username-based.

Sorry for the large PR. It's *mostly* just direct changes from `userId` to `user` in a lot of places, but there are a few places I had to change some logic as well, mostly in the components.

### DIFF
--- a/src/amo/api/collections.js
+++ b/src/amo/api/collections.js
@@ -15,7 +15,7 @@ import type {
 export type GetCollectionParams = {|
   api: ApiStateType,
   slug: string,
-  user: string | number,
+  user: string,
 |};
 
 export const getCollectionDetail = (
@@ -92,7 +92,7 @@ export const getAllCollectionAddons = async (
 type ListCollectionsParams = {|
   api: ApiStateType,
   nextURL?: string,
-  user: string | number,
+  user: string,
 |};
 
 export const listCollections = (
@@ -260,7 +260,7 @@ type ModifyCollectionAddonBaseParams = {|
   addonId: number,
   api: ApiStateType,
   slug: string,
-  user: string | number,
+  user: string,
   _modifyCollectionAddon?: (any) => Promise<void>,
 |};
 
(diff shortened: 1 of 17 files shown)


willdurand:
@bobsilverberg before I fully review this PR (which is a nice one 💪), why did you rename everything to `user` if we only pass the `username` everywhere? I thought we used `user` for both `userId` or `username` (because `userIdOrName` would be a bad variable name).

bobsilverberg:
I used `user` because there was existing code that uses `user`, defined as a `string` when all we ever pass is `username`, so rather than change all the code, I figured I'd just keep `user` and do away with `userId`. I infer that you are suggesting I should have used `username` instead, and that definitely makes sense. I will update it tomorrow.

bobsilverberg:
Ok, I changed everything to be `username`.

willdurand:
@bobsilverberg do you also see lots of conflicts for this PR?

bobsilverberg:
> do you also see lots of conflicts for this PR?

@willdurand Probably because I merged PR #5165. I'll rebase and fix those.

bobsilverberg:
Ok @willdurand I have rebased and fixed the (large number of) conflicts, so this should be ready for a look.

bobsilverberg:
Thanks @willdurand!

Issue #4998, opened 2018-05-15
Redirect user to profile page when profile has been updated?


willdurand:
When a user edits their profile and clicks "update", the current behavior is to display a "success" message and to stay on the same (edit) page. It is the behavior on the legacy frontend as well.

Should we change that? For instance, we could redirect the user to their profile once updated.

pwalm:
Correct: Clicking update should just take you to the profile page.

ioanarusiczki:
Verified fixed on AMO dev with FF60(Win10&Android 8.0)
![update profile](https://user-images.githubusercontent.com/33448286/41278989-f09946b2-6e33-11e8-8894-8c6008c3b082.gif)

PR #5144, opened 2018-05-31
I4998 - redirects to user profile page on user profile save

tsl143:
Fixes: #4998 redirects to user profile page on user profile edit save

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -165,17 +165,8 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     }
 
     if (wasUpdating && !isUpdating && !errorHandler.hasError()) {
-      if (newUser && !newUser.notifications) {
-        dispatch(fetchUserNotifications({
-          errorHandlerId: errorHandler.id,
-          username: newUser.username,
-        }));
-      }
-
-      this.setState({
-        pictureData: null,
-        successMessage: i18n.gettext('Profile successfully updated'),
-      });
+      router.push(`/${lang}/${clientApp}/user/${newUsername}/`);
+      return;
     }
 
     if (oldUser && oldUser.picture_url && newUser && !newUser.picture_url) {
--- a/tests/unit/amo/components/TestUserProfileEdit.js
+++ b/tests/unit/amo/components/TestUserProfileEdit.js
@@ -643,8 +643,18 @@ describe(__filename, () => {
     }));
   });
 
-  it('renders a success message when user profile has been updated', () => {
-    const { store } = signInUserWithUsername('tofumatt');
+  it('redirects to user profile page when user profile has been updated', () => {
+    const username = 'tofumatt';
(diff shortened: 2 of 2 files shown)


src/amo/components/UserProfileEdit/index.js
```diff
@@ -154,11 +161,16 @@ export class UserProfileEditBase extends React.Component<Props, State> {
       });
     }
 
+    if (wasUpdating && !isUpdating) {
```
tsl143: This is done to scroll top in case of both error and success. In case of success while page is reloading user can see success notice.

tsl143:
@willdurand , I had a PR for 5023 (#5051 ), which was closed on pretext of this bug.
Hence I used this bug as an opportunity to over that case too.
Your thoughts?

willdurand:
> @willdurand , I had a PR for 5023 (#5051 ), which was closed on pretext of this bug.
> Hence I used this bug as an opportunity to over that case too.
> Your thoughts?

The `scrollTo` thing is only for errors and profile picture deletion, we'll do this separately.

So I think we can redirect the user to their profile when update has been performed (and not show a message anymore). We have to use `router.push()` though, not `window.location`.

tsl143:
Okay, so in that case, shall I already remove `notice` component from `userprofileedit` ?

willdurand:
> Okay, so in that case, shall I already remove notice component from userprofileedit ?

I think we can keep it for error messages, only the confirmation message will be removed.

tsl143:
Error is already dealt by error handler.rendererror
Notice is explicitly used for success message.

tsl143:
@willdurand , I got your point :)
Updated the PR with `router.push` approach.
Thanks.

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -576,7 +576,17 @@ describe(__filename, () => {
   });
 
   it('renders a success message when user profile has been updated', () => {
```
willdurand: You should probably edit this test case description.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -155,10 +155,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     }
 
     if (wasUpdating && !isUpdating && !errorHandler.hasError()) {
-      this.setState({
-        pictureData: null,
-        successMessage: i18n.gettext('Profile successfully updated'),
-      });
+      router.push(`/${lang}/${clientApp}/user/${newUsername}/`);
```
willdurand: I think we should have a `return` below no?

tsl143: good to have

tsl143: I can do that tonight :)

willdurand:
I went ahead and fixed/rebased your branch :)

tsl143:
Thanks, got caught up with something :)

willdurand:
Thanks @tsl143!

PR #5152, opened 2018-05-31
Add a ConfirmButton and use it to delete a profile picture

### LINKED ISSUE #5151: Add a confirmation step before deleting a profile picture

1. As a logged-in user, edit your profile
2. Make sure you have uploaded a profile picture
3. Click "delete this picture" below the profile picture

### What happened?

The profile picture is deleted.

A confirmation step would be nice to avoid deletion by accident.

We should create a `ConfirmButton` component for this use case, so that we can reuse it elsewhere.

willdurand:
Fix #5151

---

Notes:

- I chose to hide this action on small screens. I am not sure we really need it on a smartphone...
- The "cancel" style for the button is fixed in another PR (#5140)

### DIFF
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
-import Button from 'ui/components/Button';
+import ConfirmButton from 'ui/components/ConfirmButton';
 import UserAvatar from 'ui/components/UserAvatar';
 import type { UserType } from 'amo/reducers/users';
 import type { I18nType } from 'core/types/i18n';
@@ -68,14 +68,14 @@ export const UserProfileEditPictureBase = ({
       </label>
 
       {(user && user.picture_url) && (
-        <p>
-          <Button
-            className="UserProfileEditPicture-delete-button"
-            onClick={onDelete}
-          >
-            {i18n.gettext('Delete this picture')}
-          </Button>
-        </p>
+        <ConfirmButton
+          buttonType="cancel"
+          className="UserProfileEditPicture-delete-button"
+          message={i18n.gettext('Do you really want to delete this picture?')}
+          onConfirm={onDelete}
+        >
+          {i18n.gettext('Delete this picture')}
+        </ConfirmButton>
       )}
     </section>
   );
--- a/src/amo/components/UserProfileEditPicture/styles.scss
+++ b/src/amo/components/UserProfileEditPicture/styles.scss
@@ -46,7 +46,14 @@ $picture-size: 128px;
   }
 
   .UserProfileEditPicture-delete-button {
-    cursor: pointer;
-    width: $picture-size;
+    margin: 12px 0;
+
+    .ConfirmButton-default-button {
+      width: $picture-size;
+    }
+
+    @include respond-to(medium) {
+      width: $picture-size;
+    }
   }
 }
--- a/src/ui/components/ConfirmButton/index.js
+++ b/src/ui/components/ConfirmButton/index.js
@@ -11,13 +11,13 @@ import './styles.scss';
 
 
 type Props = {|
-  buttonType?: string,
-  cancelButtonText?: string,
-  cancelButtonType?: string,
+  buttonType: string,
+  cancelButtonText: string,
+  cancelButtonType: string,
   children: React.Element<any>,
   className?: string,
-  confirmButtonText?: string,
-  confirmButtonType?: string,
+  confirmButtonText: string,
+  confirmButtonType: string,
   i18n: I18nType,
   message: string,
   onConfirm: Function,
--- a/src/ui/components/ConfirmButton/index.js
+++ b/src/ui/components/ConfirmButton/index.js
@@ -1,0 +1,122 @@
+/* @flow */
+import makeClassName from 'classnames';
+import * as React from 'react';
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+import Button from 'ui/components/Button';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+
+type Props = {|
+  buttonType?: string,
+  cancelButtonText?: string,
+  cancelButtonType?: string,
+  children: React.Element<any>,
+  className?: string,
+  confirmButtonText?: string,
+  confirmButtonType?: string,
+  i18n: I18nType,
+  message: string,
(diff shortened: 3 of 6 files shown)


src/ui/components/ConfirmButton/index.js
```diff
@@ -0,0 +1,122 @@
...
+      message,
+    } = this.props;
+
+    const { showConfirmation } = this.state;
+
+    const classNames = makeClassName('ConfirmButton', className, {
+      'ConfirmButton--show-confirmation': showConfirmation,
+    });
+
+    if (this.state.showConfirmation) {
```
rebmullin: just `if (showConfirmation) }` ?

tests/unit/amo/components/TestUserProfileEditPicture.js
```diff
@@ -101,22 +102,29 @@ describe(__filename, () => {
     sinon.assert.callCount(onSelect, 1);
   });
 
-  it('renders a "delete" button when user has a picture URL', () => {
+  it('renders a "delete" confirm button when user has a picture URL', () => {
```
rebmullin: should we add a test for showing the default button ('Delete this picture') button too?

also should we get more specific and show a test that this (the delete confirm) only comes up after clicking the "Delete this picture" button? or is that just assumed?

src/amo/components/UserProfileEditPicture/styles.scss
```diff
@@ -46,7 +46,16 @@ $picture-size: 128px;
   }
 
   .UserProfileEditPicture-delete-button {
-    cursor: pointer;
-    width: $picture-size;
+    display: none;
```
rebmullin: i know you made note about this but was wondering if we should keep this. I guess i thought it was always better to have the same functionality - as much as possible - between the different breakpoints, no?

src/ui/components/ConfirmButton/index.js
willdurand: good catch!

src/amo/components/UserProfileEditPicture/styles.scss
willdurand: maybe yeah

src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -68,14 +68,14 @@ export const UserProfileEditPictureBase = ({
...
-            {i18n.gettext('Delete this picture')}
-          </Button>
-        </p>
+        <ConfirmButton
+          buttonType="cancel"
+          className="UserProfileEditPicture-delete-button"
+          message={i18n.gettext('Do you really want to delete this picture?')}
+          onConfirm={onDelete}
+        >
+          {i18n.gettext('Delete this picture')}
```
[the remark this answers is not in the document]
willdurand: nope it should look like a simple link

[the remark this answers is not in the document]
rebmullin: should we could keep `cursor: pointer` style then - you removed it below it seems - to show it's a link?

src/amo/components/UserProfileEditPicture/styles.scss
rebmullin: even on the same device it could show up on landscape orientation and not portrait so just wanted to make sure it's what we want : )

src/amo/components/UserProfileEditPicture/index.js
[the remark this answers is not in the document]
willdurand: The style for this button will be fixed in another PR, which I expected to land soon

tests/unit/amo/components/TestUserProfileEditPicture.js
rebmullin: I guess I do see the delete-button below but maybe if the description is a little different that will be a little clearer ~ something like "renders a delete link which triggers a confirm (delete) button when user has a picture URL"

rebmullin:
@willdurand, this looks good to me just a few notes/qs above. 

I'll ask @bobsilverberg too to take a look since you requested him initially : )

tests/unit/amo/components/TestUserProfileEditPicture.js
willdurand: I will change to `renders a "delete" ConfirmButton when user [...]` so that it indicates that we're using a component. Would that work for you?

src/amo/components/UserProfileEditPicture/styles.scss
willdurand: Ok, that was silly. Thanks for catching this, I will make the button always available.

src/ui/components/ConfirmButton/index.js
```diff
@@ -0,0 +1,122 @@
...
+
+import translate from 'core/i18n/translate';
+import Button from 'ui/components/Button';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+
+type Props = {|
+  buttonType?: string,
```
bobsilverberg: Why is this marked as optional when it has a default value? Won't it always have a value of at least `neutral` because of the defaults?

src/ui/components/ConfirmButton/index.js
```diff
@@ -0,0 +1,122 @@
...
+  cancelButtonText?: string,
+  cancelButtonType?: string,
```
bobsilverberg: Same question for this one, and for `confirmButtonType` below.

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -0,0 +1,140 @@
...
+    root.find(Button).simulate('click', createFakeEvent());
+
+    expect(root).toHaveClassName('ConfirmButton--show-confirmation');
+
+    root.find('.ConfirmButton-cancel-button').simulate(
+      'click',
+      createFakeEvent()
+    );
+
+    expect(root).not.toHaveClassName('ConfirmButton--show-confirmation');
```
bobsilverberg: You have a test to check that `onConfirm` is called when the user clicks the "confirm" button, so maybe it would be worth adding an assertion into this test to assert that `onConfirm` is not called when a user clicks the "cancel" button?

src/ui/components/ConfirmButton/index.js
willdurand: Interesting, so I was about to write: it is optional as in "can be omitted when using the component" and not as in "nullable", but I checked the Flow documentation and things have changed apparently:

> Note: You don’t need to make foo nullable in your Props type. Flow will make sure that foo is optional if you have a default prop for foo.

https://flow.org/en/docs/react/components/

Removing the `?` now :)

Issue #4578, opened 2018-03-16
Edit comments about an add-on in a collection


kumar303:
Tracker: https://github.com/mozilla/addons-frontend/issues/2782

A user should be able to:

- Sign in
- View an existing collection that they created
- Edit the collection (which opens the overlay)
- Add/edit comments about an add-on in the collection

bobsilverberg:
@pwalm I'm wondering about the UX for this. I have attached the current mockup from Sketch, but it is a bit out of date as it still shows the overlay. The mockup looks like this:

![collections - edit collection with notes - 1366px](https://user-images.githubusercontent.com/142755/40670697-e5e0e4de-6337-11e8-8f46-b55036090a8b.png)

And the current edit page looks like this:

<img width="1218" alt="screenshot 2018-05-29 12 09 20" src="https://user-images.githubusercontent.com/142755/40671164-23c414c8-6339-11e8-9b42-73a57f7e9e3b.png">

I can pretty much infer what to do with the new screen based on the mockup, but I'm unsure about the workflow. 

I assume that a textarea should appear when a user clicks on the "notes" icon for an addon, but how do they save notes? Should there be a save button somewhere to go with the textarea? Should we automatically show a text area for any add-ons that already have notes? In not, in what other way can we indicate to a user that the add-on has notes?

Maybe we can discuss this in this week's Front End review meeting?

pwalm:
Yeah, let's talk about it in the meeting.

pwalm:
Ok, here we go (forgive my artistry):
Create comment mode:
![add-ons - 2018-05-29 18 59 59](https://user-images.githubusercontent.com/6797635/40738645-f0e37a6a-63f8-11e8-9d43-eeb2fda0a76e.jpg)

Commented state (once it has been saved):
![add-ons - 2018-05-29 18 59 59 2](https://user-images.githubusercontent.com/6797635/40738669-06254ab6-63f9-11e8-9501-8863e2437622.jpg)

Edit mode (returning to edit an existing comment):
![add-ons - 2018-05-29 18 59 59 3](https://user-images.githubusercontent.com/6797635/40738681-12449202-63f9-11e8-9044-040c798a0b7d.jpg)

Clicking "Delete" just deletes the comment, no confirmation needed.

ioanarusiczki:
Verified fixed on AMO dev with FF60 (Win10 and Android 8.0).
I filed a separate issue https://github.com/mozilla/addons-frontend/issues/5315

PR #5170, opened 2018-06-02
Reducer/Saga changes to support updating the notes for a collection add-on

bobsilverberg:
Ref: #4578 
~~Depends on: #5165~~

### DIFF
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -3,7 +3,7 @@ import { oneLine } from 'common-tags';
 import invariant from 'invariant';
 
 import { createInternalAddon } from 'core/reducers/addons';
-import type { AddonType, ExternalAddonType } from 'core/types/addons';
+import type { CollectionAddonType, ExternalAddonType } from 'core/types/addons';
 import type { LocalizedString } from 'core/types/api';
 
 export const ADD_ADDON_TO_COLLECTION: 'ADD_ADDON_TO_COLLECTION'
@@ -47,7 +47,7 @@ export const DELETE_COLLECTION_ADDON_NOTES: 'DELETE_COLLECTION_ADDON_NOTES'
   = 'DELETE_COLLECTION_ADDON_NOTES';
 
 export type CollectionType = {
-  addons: Array<AddonType> | null,
+  addons: Array<CollectionAddonType> | null,
   authorId: number,
   authorName: string,
   authorUsername: string,
@@ -710,12 +710,12 @@ export const deleteCollectionAddonNotes = ({
 
 export const createInternalAddons = (
   items: ExternalCollectionAddons
-): Array<AddonType> => {
-  return items.map((item) => {
+): Array<CollectionAddonType> => {
+  return items.map(({ addon, notes }) => {
     // This allows to have a consistent way to manipulate addons in the app.
     return {
-      ...createInternalAddon(item.addon),
-      collectionNotes: item.notes,
+      ...createInternalAddon(addon),
+      notes,
     };
   });
 };
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -41,6 +41,10 @@ export const FINISH_COLLECTION_MODIFICATION: 'FINISH_COLLECTION_MODIFICATION'
 export const REMOVE_ADDON_FROM_COLLECTION: 'REMOVE_ADDON_FROM_COLLECTION'
   = 'REMOVE_ADDON_FROM_COLLECTION';
 export const DELETE_COLLECTION: 'DELETE_COLLECTION' = 'DELETE_COLLECTION';
+export const UPDATE_COLLECTION_ADDON: 'UPDATE_COLLECTION_ADDON'
+  = 'UPDATE_COLLECTION_ADDON';
+export const DELETE_COLLECTION_ADDON_NOTES: 'DELETE_COLLECTION_ADDON_NOTES'
+  = 'DELETE_COLLECTION_ADDON_NOTES';
 
 export type CollectionType = {
   addons: Array<AddonType> | null,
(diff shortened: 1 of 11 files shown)


bobsilverberg:
@willdurand After spending some time on the UI portion of this feature, I think I want to make some changes to this approach. The only thing we ever do when updating a collection add-on is to add/change/remove notes, so I'm thinking that the reducer actions should be named thusly. I.e., `updateCollectionAddonNote`, as opposed to just `updateCollectionAddon`, and I also think it's worth adding a reducer method for `removeCollectionAddonNote`. As it stands, the implementation for `remove` is to just perform an `update` with a `notes` value of `''`, but I think it makes sense to encapsulate the logic for that in the reducer with its own method.

What do you think?

willdurand:
> @willdurand After spending some time on the UI portion of this feature, I think I want to make some changes to this approach. The only thing we ever do when updating a collection add-on is to add/change/remove notes, so I'm thinking that the reducer actions should be named thusly. I.e., updateCollectionAddonNote, as opposed to just updateCollectionAddon

@bobsilverberg that makes a lot of sense.

> and I also think it's worth adding a reducer method for removeCollectionAddonNote. As it stands, the implementation for remove is to just perform an update with a notes value of '', but I think it makes sense to encapsulate the logic for that in the reducer with its own method.

well, this is up to you. Usually, we extract the logic of the switch/case when it is too big (like > 15-20 LOC) and/or reused for another action.

bobsilverberg:
> well, this is up to you. Usually, we extract the logic of the switch/case when it is too big (like > 15-20 LOC) and/or reused for another action.

My thought was more about encapsulating the logic for removing a note in the reducer, rather than having it in the calling component. If we just have `updateCollectionAddonNote` then the caller would need to know that, in order to actually remove a note, they need to pass a `''` as the value of `notes`, and it seems like it would be better for a caller to not have to know that. It also opens the door later to a different implementation of removing notes if that changes (which it may).

willdurand:
> My thought was more about encapsulating the logic for removing a note in the reducer, rather than having it in the calling component. If we just have updateCollectionAddonNote then the caller would need to know that, in order to actually remove a note, they need to pass a '' as the value of notes, and it seems like it would be better for a caller to not have to know that. It also opens the door later to a different implementation of removing notes if that changes (which it may).

Ah. I see, sorry. Having two different actions is a better design indeed.

bobsilverberg:
I changed my mind and am not going to rename the method, because all I'd be doing is making the name longer, but I did introduce a separate reducer method for deleting the notes for a collection add-on, which simply makes use of the existing saga for updating a collection add-on. This has also been rebased against the changes from #5165, so should be ready for a review.

willdurand:
> I did introduce a separate reducer method for deleting the notes for a collection add-on, which simply makes use of the existing saga for updating a collection add-on

I would prefer two different actions and action creators (what you call "reducer method" I guess) bound to the same saga because it is a bit strange to have an `UPDATE` that is actually a `DELETE` in the redux logs.

The API documentation indicates that `null` can be sent, and I wonder if that would make sense to send `null` rather than an empty string.

bobsilverberg:
> I would prefer two different actions and action creators (what you call "reducer method" I guess) bound to the same saga because it is a bit strange to have an UPDATE that is actually a DELETE in the redux logs.

Sure, that makes sense.

> The API documentation indicates that null can be sent, and I wonder if that would make sense to send null rather than an empty string.

We cannot because of https://github.com/mozilla/addons-server/issues/7832. Using `''` is a workaround for now. I will add a comment to the code pointing to that issue.

willdurand:
> We cannot because of mozilla/addons-server#7832. Using '' is a workaround for now. I will add a comment to the code pointing to that issue.

ah sorry, thanks for letting me know!

bobsilverberg:
@willdurand I have made the changes we discussed, and have also added the ability for the collections reducer to load `collectionNotes` for an add-on that is in a collection, as we need those both for displaying the notes and also for creating/editing/removing the notes. This is ready for another look.

src/amo/reducers/collections.js
```diff
@@ -642,12 +646,77 @@ export const deleteCollection = ({
...
+    payload: {
+      addonId, errorHandlerId, notes, page, slug, username,
+    },
+  };
+};
+
+// The DeleteCollectionAddonNotesAction is identical to the
+// UpdateCollectionAddonAction, with the only difference being that
+// we pass '' into the notes field. Therefore we can reuse the
+// UpdateCollectionAddonParams type to describe the payload.
```
willdurand: Thanks!

src/amo/reducers/collections.js
```diff
@@ -642,12 +646,77 @@ export const deleteCollection = ({
...
+  addonId, errorHandlerId, page, slug, username,
+}: UpdateCollectionAddonParams = {}): DeleteCollectionAddonNotesAction => {
+  invariant(addonId, 'The addonId parameter is required');
+  invariant(errorHandlerId, 'The errorHandlerId parameter is required');
+  invariant(page, 'The page parameter is required');
+  invariant(slug, 'The slug parameter is required');
+  invariant(username, 'The username parameter is required');
+
+  // For delete we set the notes field to an empty string and call update.
+  // TODO: Use a null instead when https://github.com/mozilla/addons-server/issues/7832 is fixed.
```
willdurand: 👍

src/core/types/addons.js
```diff
@@ -173,6 +173,8 @@ export type AddonType = {|
   isRestartRequired: boolean,
   isWebExtension: boolean,
   themeData?: ThemeData,
+  // This is present for collection add-ons
+  collectionNotes?: string | null,
```
willdurand: I don't think we should be doing this, because it weakens our `AddonType` IMO, meaning the Flow checker will be less accurate and potentially miss issues.

Since this is only used in the "collections" context, that would probably make sense to have a new type:

```js
type CollectionAddonType = {|
  ...AddonType,
  notes: string | null,
|};
```

What do you think?

willdurand: I also think it would make sense to do it here.

tests/unit/amo/helpers.js
```diff
@@ -353,15 +353,25 @@ export const createFakeCollectionDetail = ({
...
+
+export const createFakeCollectionAddons = ({
+  addons = [createFakeCollectionAddon()],
+} = {}) => {
   return addons.map((addon) => ({
-    addon,
+    addon: addon.addon,
     downloads: 0,
-    notes: null,
+    notes: addon.notes,
```
willdurand: You could also destructure the unique `map` parameter: `.map({ addon, notes }) => ({`

tests/unit/amo/reducers/test_collections.js
```diff
@@ -169,8 +171,10 @@ describe(__filename, () => {
         detail: collectionDetail,
       }));
 
+      const fakeCollectionAddon =
+        createFakeCollectionAddon({ addon: { ...fakeAddon, id: 333 }, notes });
```
willdurand: I usually find more readable to write:

```js
const fakeCollectionAddon = createFakeCollectionAddon({
  addon: { ...fakeAddon, id: 333 },
  notes,
});
```

By having one parameter per line, I find it easier to parse the different parameters at a given level (like here `addon` and `note` are the two params given to `createFakeCollectionAddon`).

tests/unit/amo/reducers/test_collections.js
```diff
@@ -444,6 +451,24 @@ describe(__filename, () => {
       expect(collectionInState.addons)
         .toEqual(createInternalAddons(addons));
     });
+
+    it('loads notes for collection add-ons', () => {
+      const notes = 'These are some notes.';
+      const fakeCollectionAddon =
+        createFakeCollectionAddon({ notes });
```
willdurand: These two lines should fit on a single line, no?

tests/unit/amo/reducers/test_collections.js
```diff
@@ -444,6 +451,24 @@ describe(__filename, () => {
...
+      });
+      const collection = createFakeCollectionDetail({ addons });
+
+      const state = loadCollectionIntoState({
+        state: initialState, collection, addons,
+      });
+
+      const collectionInState = state.byId[collection.id];
+      expect(collectionInState.addons[0].collectionNotes)
+        .toEqual(notes);
```
willdurand: In general, it is better to use the full available width. 80 chars is not a lot, but these two lines should have less chars than that 🤓

src/amo/reducers/collections.js
```diff
@@ -642,12 +646,77 @@ export const deleteCollection = ({
...
+  return {
+    type: DELETE_COLLECTION_ADDON_NOTES,
+    payload: {
+      addonId, errorHandlerId, notes: '', page, slug, username,
```
willdurand: Would it make sense to add a test case for this action creator that returns an empty string for `notes`? (so that we notice it when we'll change to `null` once the API is fixed).

src/core/types/addons.js
bobsilverberg: I originally tried to create a new type for `CollectionAddonType`, but it was causing Flow nightmares. I agree it would be better, so I'll try again. 👍

src/amo/reducers/collections.js
bobsilverberg: That is being tested here: https://github.com/bobsilverberg/addons-frontend/blob/6977e2f7448ce56d8a3640be13925b5d1ae9e2de/tests/unit/amo/sagas/test_collections.js#L880. This is generally (I believe) the way we test action creators.

Do you disagree?

willdurand: Yup 👍

bobsilverberg:
Flow is happy so this is ready for another review.

tests/unit/amo/reducers/test_collections.js
willdurand: This has not been changed.

bobsilverberg: I thought I changed that. Odd. I'll fix it.

PR #5172, opened 2018-06-03
I-5003 show pagination when adding 25+ addons in a row

### LINKED ISSUE #5003: No pagination when adding more than 25 addons to a collection in a row

- Create a new collection
- Add more than 25 add-ons (good luck)

### What happened?

- No pagination appears at the bottom
- The list is limited to 25 add-ons

- No pagination or the list with all the add-ons (no limited)

xu3u4:
Fixes #5003 

After adding a new addon, the addons in that collection will be reloaded. When loading the new addons, we can get updated addon count from API.
Previously, we didn't use it so that the `numberOfAddons` passed to pagination is still the old value.
So once we pass the new addon count, pagination will appear accordingly.


src/amo/reducers/collections.js
```diff
@@ -315,14 +316,15 @@ type LoadCurrentCollectionPageAction = {|
 
 export const loadCurrentCollectionPage = ({
   addons,
+  numberOfAddons,
 }: LoadCurrentCollectionPageParams = {}): LoadCurrentCollectionPageAction => {
   if (!addons) {
     throw new Error('addons are required');
   }
```
bobsilverberg: We should probably add a check here that `numberOfAddons` isn't `null` or `undefined`. We cannot just check against `!numberOfAddons` because `0` is a valid value, but we could check that it's a number, which would protect us against a `null` or `undefined`.

You'll notice that some of the functions in this file use `invariant` instead of an `if` statement, and I'd suggest using that here, for both the existing check for `addons` and the new one you'll add for `numberOfAddons`.

xu3u4: Cool! Thanks.
I will fix it soon.

src/amo/reducers/collections.js
```diff
@@ -315,14 +316,15 @@ type LoadCurrentCollectionPageAction = {|
...
 export const loadCurrentCollectionPage = ({
   addons,
+  numberOfAddons,
 }: LoadCurrentCollectionPageParams = {}): LoadCurrentCollectionPageAction => {
-  if (!addons) {
-    throw new Error('addons are required');
-  }
+
+  invariant(addons, 'The addons parameter is required');
+  invariant(numberOfAddons, 'The numberOfAddons parameter is required');
```
bobsilverberg: This looks good, but `0` is a valid value for `numberOfAddons`, and this code will fail with a value of `0`, which is why I suggested earlier that you could either check that `numberOfAddons` is neither `undefined` or `null`, or just check that it's a number.

xu3u4: oh sorry.
Please check again. Thank you.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -156,7 +156,7 @@ describe(__filename, () => {
     it('cannot load collection page without a current collection', () => {
```
bobsilverberg: We don't generally write tests for our calls to `invariant`, so this test can be removed entirely.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -665,7 +667,15 @@ describe(__filename, () => {
     it('throws an error when addons are missing', () => {
       expect(() => {
         loadCurrentCollectionPage();
-      }).toThrow('addons are required');
+      }).toThrow('The addons parameter is required');
+    });
+
+    it('throws an error when numberOfAddons is missing', () => {
```
bobsilverberg: The same goes for these two test cases: we can remove them if all they are testing is the `invariant` checks.

src/amo/reducers/collections.js
```diff
@@ -315,14 +316,16 @@ type LoadCurrentCollectionPageAction = {|
 
 export const loadCurrentCollectionPage = ({
   addons,
+  numberOfAddons,
 }: LoadCurrentCollectionPageParams = {}): LoadCurrentCollectionPageAction => {
-  if (!addons) {
-    throw new Error('addons are required');
+  invariant(addons, 'The addons parameter is required');
+  if (typeof numberOfAddons !== 'number') {
```
bobsilverberg: We can still use `invariant` here, it would look like:

`invariant(typeof numberOfAddons === 'number', 'The numberOfAddons parameter must be a number');`

tests/unit/amo/reducers/test_collections.js
```diff
@@ -662,10 +664,12 @@ describe(__filename, () => {
   });
 
   describe('loadCurrentCollectionPage()', () => {
-    it('throws an error when addons are missing', () => {
+    it('throws an error when numberOfAddons is missing', () => {
```
bobsilverberg: With the change above, you can remove this test.

tests/unit/amo/reducers/test_collections.js
xu3u4: I think here is to test the case `without a current collection` and it doesn't use `invariant`, so maybe it is ok to keep?

src/amo/reducers/collections.js
xu3u4: Oh! Yes. It looks better, thanks.

tests/unit/amo/reducers/test_collections.js
bobsilverberg: Right, I missed that. Thanks @xu3u4. 👍

xu3u4:
@bobsilverberg Thank you a lot for helping me with this PR!

bobsilverberg:
@xu3u4 My pleasure!

Issue #5174, opened 2018-06-04
Static themes: Cropping makes it hard to tell what you're going to get at small viewports


muffinresearch:
Maybe there's a limit to what can be done here but as viewport sizes decrease it gets harder to see enough of a static theme preview to be useful.

This is going to be particularly an issue with LWTs that are converted, because they rely on right-aligned images.

URL: https://addons-dev.allizom.org/en-US/firefox/search/?platform=mac&q=static+theme

Here's a mobile size viewport:

![screen shot 2018-06-04 at 10 22 07](https://user-images.githubusercontent.com/1514/40909445-77aa89d0-67e1-11e8-809c-33d2615d6dfa.png)

Here's a wider view on the same data:

![screen shot 2018-06-04 at 10 24 46](https://user-images.githubusercontent.com/1514/40909497-93601a00-67e1-11e8-8db2-31a68720ff83.png)

Do we have alternative images that we can show at mobile viewports?

muffinresearch:
This will need some discussion as to how to best handle this.

rebmullin:
Per our discussion at all-hands and [addons-server-issue-8412](https://github.com/mozilla/addons-server/issues/8412), I think we are a good place with this, i.e. we are going to get some new images sizes and scale down if necessary by adjusting the css - which I've started in [this PR](https://github.com/mozilla/addons-frontend/pull/5225).

rebmullin:
also additional notes from discussion at all-hands:

implement srcset once the additional sizes are ready:
mozilla/addons-server#8412 (comment)

Update themes grid in the following places:
Homepage,
Themes landing,
Category landing
Note for above pages: 
desktop: change from 4 to 3 col grid
~tablet: change from 4 to 2 col grid
~mobile: should be 1 col - should already be this way


AuthorsAddoncards,
Note for above component: 
desktop: change from 3 to 2 col grid
landscape tablet: change from 3 to 2 grid
portrait tablet && mobile: should be 1 col - should already be this way

rebmullin:
note: we are holding off on srcset to further discuss. Will update this PR # shortly..

rebmullin:
note for QA:

disregard srcset (per my notes above - a follow up PR will happen if that is needed)

make sure columns match what i've above (except note for tablet grids - on home, landing and cat pages, we are not able to switch the results so it is still 3 here)

make sure images aren't cropped

please double check nothing is broken or looks off

[...]

Issue #5018, opened 2018-05-16
Static theme preview should be positioned similar to LWT preview in theme details page


AlexandraMoga:
STR:

1. Log in to AMO-dev and go to a static theme details page - see https://addons-dev.allizom.org/en-US/firefox/addon/my-first-sentence/
2. Open the same LWT theme's detail page on AMO stage - https://addons.allizom.org/en-US/firefox/addon/my-first-sentence/?src=rating
3. Compare the two theme previews

Actual result:
Static theme preview is displayed in a Screenshots section which doesn't clearly mark the theme design as the main focus on the theme details page

Expected result:
Static theme preview should be displayed in the same card LWT are currently displayed; also, the Screenshots section shouldn't really be present on the ST detail page

Notes:
- reproduced on AMO dev with FF60, Win10x64

![image](https://user-images.githubusercontent.com/31961530/40113039-bb286dc4-5910-11e8-8da3-a2e8db5a6587.png)

bobsilverberg:
This seems valid. I know supporting static themes on the front end is something new. Do we have issues open for it? Are there mockups for exactly how they should look?

rebmullin:
I think the "mockups" are to follow the design of the current theme (LWT) - REF: https://github.com/mozilla/addons-frontend/issues/4899#issuecomment-385553081

// cc @jvillalobos  please correct me if i am wrong

jvillalobos:
Yes, the layout should be as close as possible to current lwt. Not sure if the issues to work on this have been filed already. /cc @eviljeff

eviljeff:
Correct.  The only difference with the preview vs. lwt header is UX defined the size as exactly as 680 x 92 (as of mozilla/addons-server#8248).  Frontend displays something between 770 x 90 and 630 x 90 now, depending on if there are badges, (the actual header image is 680 x 100) but the static theme preview shown on the detail page shouldn't be cropped (and ideally shouldn't be scaled either).

(The "thumbnail" of the preview is smaller and to be used on list results, shelves, etc, and can be cropped at the right hand size)

rebmullin:
how should we handle the header in mobile / smaller devices? Is this okay to scale here or should we use the same height and cut the right side off? 

regarding "shelves" - where can i see that? Is that this multi column view:
![Alt text](https://monosnap.com/image/dCuLndowhikqmNr5UZlfzNbvdzqPf5.png)

or something different?

@jvillalobos @eviljeff

[...]

PR #5178, opened 2018-06-04
I5173 static theme persona style fix

### LINKED ISSUE #5173: Right-cropping theme previews will hide important elements in the theme design
Follow up for #5067 

STR:
1. Load AMO-dev homrpage (https://addons-dev.allizom.org/en-US/firefox/)
2. Examine the Featured Themes shelf

Actual result:
The cropped theme previews are hiding important elements from the theme design

Expected result:
The user should be able to see a more complete representation of the theme design when looking at the Featured themes shelf

Notes:
- the same behavior is visible on Theme's homepage and in More themes by artist card (present on theme details page and in User profile page)
- it is also the design used for smaller screen sizes
- maybe a good point to address in https://github.com/mozilla/addons-server/issues/8412
- reproduced on AMO-dev with FF60, Win 10x64 and Android 7.0

[1] AMO homepage
image

[2] Theme homepage
image

[3] More themes by author card
image

rebmullin:
fixes #5173 

for persona theme images, i am reverting back to `top right` styles so nothing is cut off and to give better sense of the theme style

### DIFF
--- a/src/amo/components/SearchResult/styles.scss
+++ b/src/amo/components/SearchResult/styles.scss
@@ -64,10 +64,7 @@
 
   .SearchResult--persona & {
     height: 100px;
-  }
-
-  .AddonsCard--horizontal .SearchResult--persona & {
-    height: $theme-height-default;
+    object-position: top right;
   }
 }
 


rebmullin:
note: this is somewhat of a short term solution ...
once lwt migration happens we will need to figure how to identify these migrated themes since static themes are set up to be cropped on the right (see https://github.com/mozilla/addons-frontend/issues/5018#issuecomment-390280137) and lwt are set up for right alignment

also - there is an issue which also brings this up #5174

willdurand:
> note: this is somewhat of a short term solution ...
> once lwt migration happens we will need to figure how to identify these migrated themes since static themes are set up to be cropped on the right (see #5018 (comment)) and lwt are set up for right alignment
> 
> also - there is an issue - #5174 - which also brings this up

Right, I think #5174 is the right place to discuss because we obviously need to discuss this point with UX and Product.

rebmullin:
>Right, I think #5174 is the right place to discuss because we obviously need to discuss this point with UX and Product.

agreed..

PR #5161, opened 2018-06-01
API/Reducer/Saga changes for delete a collection

bobsilverberg:
Ref: #4195 

This depends on #5143 (and therefore contains the commits from that as well) and will be followed by a PR to implement the delete functionality.

PR #5186, opened 2018-06-04
Delete a collection - UI portion

### LINKED ISSUE #4195: Delete a collection
After creating a collection (see https://github.com/mozilla/addons-frontend/issues/4003 or use the legacy site), the user should be able to enter the edit mode (see https://github.com/mozilla/addons-frontend/issues/4002) and then click the delete button shown in the bottom left of the overlay. This will remove the collection from all listings.

screenshot 2018-01-22 11 21 04

bobsilverberg:
Fixes #4195 

Depends on #5161 

This adds a "Delete this collection" button to the Collection screen, both in display and edit mode, which only appears if the user has access to delete the collection. For now I have styled it similar to the "Delete this picture" button on the User Profile.

Detail screen:

screenshot 2018-06-04 13 30 18

With confirmation:

screenshot 2018-06-04 13 31 57

Edit screen:

screenshot 2018-06-04 13 30 48

With confirmation:

screenshot 2018-06-04 13 32 32

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -289,9 +289,11 @@ export class CollectionBase extends React.Component<Props> {
 
   renderDeleteButton() {
     const { hasEditPermission, i18n } = this.props;
+
     if (!hasEditPermission) {
       return null;
     }
+
     return (
       <ConfirmButton
         buttonType="cancel"
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -10,6 +10,7 @@ import AddonsCard from 'amo/components/AddonsCard';
 import Link from 'amo/components/Link';
 import {
   removeAddonFromCollection,
+  deleteCollection,
   fetchCurrentCollection,
   fetchCurrentCollectionPage,
   getCurrentCollection,
@@ -28,6 +29,7 @@ import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
+import ConfirmButton from 'ui/components/ConfirmButton';
 import LoadingText from 'ui/components/LoadingText';
 import MetadataCard from 'ui/components/MetadataCard';
 import type {
@@ -74,6 +76,28 @@ export class CollectionBase extends React.Component<Props> {
 
   componentWillReceiveProps(nextProps: Props) {
     this.loadDataIfNeeded(nextProps);
+  }
+
+  onDelete = (event: SyntheticEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+
+    const { dispatch, errorHandler, collection } = this.props;
+
+    invariant(collection, 'collection is required');
+
+    const {
+      slug,
+      authorUsername: username,
+    } = collection;
+
+    invariant(slug, 'slug is required');
+    invariant(username, 'username is required');
+
+    dispatch(deleteCollection({
+      errorHandlerId: errorHandler.id,
+      slug,
+      username,
+    }));
   }
 
   loadDataIfNeeded(nextProps?: Props) {
@@ -263,6 +287,23 @@ export class CollectionBase extends React.Component<Props> {
     /* eslint-enable react/no-danger */
   }
 
+  renderDeleteButton() {
+    const { hasEditPermission, i18n } = this.props;
+    if (!hasEditPermission) {
+      return null;
+    }
+    return (
+      <ConfirmButton
+        buttonType="cancel"
+        className="Collection-delete-button"
+        message={i18n.gettext('Do you really want to delete this collection?')}
+        onConfirm={this.onDelete}
+      >
+        {i18n.gettext('Delete this collection')}
+      </ConfirmButton>
+    );
+  }
+
   renderCollection() {
     const {
       collection,
@@ -278,6 +319,7 @@ export class CollectionBase extends React.Component<Props> {
       <div className="Collection-wrapper">
         <Card className="Collection-detail">
           {this.renderCardContents()}
+          {this.renderDeleteButton()}
         </Card>
         <div className="Collection-items">
           <AddonsCard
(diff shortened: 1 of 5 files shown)


bobsilverberg:
I see that #5161 was approved, so I'll merge that and rebase this now, to make reviewing the PR easier.

tests/unit/amo/components/TestCollection.js
```diff
@@ -886,6 +947,43 @@ describe(__filename, () => {
...
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const preventDefaultSpy = sinon.spy();
+    const errorHandler = createStubErrorHandler();
+
+    const wrapper = renderComponent({ errorHandler, store });
+
+    dispatchSpy.reset();
+
+    // This emulates a user clicking the delete button and confirming.
+    wrapper.instance().onDelete(
```
willdurand: In my opinion, we should not use `instance()` that much, and especially not here. I'd rather find the `ConfirmButton` and use `.prop` to get the handler:

```js
const onDelete = wrapper.find(ConfirmButton).prop('onConfirm');
onDelete(createFakeEvent({ ... }));
```

I consider the `onDelete` method "internal" and therefore "private". If we rename this method for some reason, we'll have to update this test case, even if the behavior has not changed. If we stick to the props (the public interface/API in React), then we won't have to update this test case.

src/amo/sagas/collections.js
```diff
@@ -338,6 +339,10 @@ export function* deleteCollection({
       errorHandlerId: errorHandler.id,
       username,
     }));
```
willdurand: So it seems the `fetchUserCollectionsAction` does not remove the old/existing collections for the current logged-in user. It means deleting a collection and clicking the "back" button will load the previously deleted collection. I am not sure that's what we want.

src/amo/components/Collection/styles.scss
```diff
@@ -86,3 +86,11 @@
   margin-top: $padding-page;
   width: 100%;
 }
+
+.Collection-delete-button {
+  margin: 12px 0;
```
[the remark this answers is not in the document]
bobsilverberg: I changed it to `margin-top: 16px;` because I think we tend to go with factors of 4. Let me know if that seems okay.

tests/unit/amo/components/TestCollection.js
bobsilverberg: Very good point, thanks for the suggestion.

src/amo/sagas/collections.js
bobsilverberg: I think the issue is actually with `loadUserCollections`, not `fetchUserCollections` in that it loads all the newly found collections, but adds them to any existing collections. I believe that is because of this line: https://github.com/mozilla/addons-frontend/blob/master/src/amo/reducers/collections.js#L968. If we were to remove that, then `loadUserCollections` would always load a fresh set of collections, which seems like what we want to do. Do you have any idea why the code might have been written that way to begin with? When might we want to just add to an existing set of user collections?

bobsilverberg:
Thanks for the review, @willdurand. I have addressed all of the comments except for the one about the collection not being removed from state, to which I've added a response and some follow-up questions. I think in any case that could be treated as a follow-up issue, if you think that's okay. This is ready for another review.

src/amo/components/Collection/styles.scss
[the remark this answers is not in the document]
willdurand: that's perfect!

src/amo/sagas/collections.js
willdurand: The line 968 of the `collections` reducer modifies a `userCollections` attribute in the state, but when I load a collection page directly, this attribute is an empty object:

screen shot 2018-06-05 at 19 17 09

It is only after I delete the collection that this `userCollections` attribute gets populated.

willdurand: Isn't it `loadCollectionIntoState()` that always keep adding data to the state?

bobsilverberg: Sorry, @willdurand, I am getting confused by this discussion. I'm now unclear about what you think the issue actually is. Is it as simple as making sure that a collection gets unloaded from state when we delete it, or is there more to it than that?

In either case, do you think we can make this a follow-up issue because it doesn't really interfere with the delete feature and only represents an annoyance for an edge-case?

willdurand: The issue is that we don't unload a deleted collection. I think it could be done in this patch (because the diff is rather small and also because I noticed this bug while reviewing this patch). Let me know if you prefer a follow-up issue though.

bobsilverberg: Ok, I've pushed a commit which unloads the collection after deleting it. Could you please take another look?

src/amo/components/Collection/index.js
```diff
@@ -263,6 +287,23 @@ export class CollectionBase extends React.Component<Props> {
     /* eslint-enable react/no-danger */
   }
 
+  renderDeleteButton() {
+    const { hasEditPermission, i18n } = this.props;
+    if (!hasEditPermission) {
+      return null;
+    }
+    return (
```
willdurand: nit: maybe a blank line before the `return` statement would be nice for readability.

src/amo/sagas/collections.js
```diff
@@ -334,10 +335,17 @@ export function* deleteCollection({
     };
     yield call(api.deleteCollection, params);
 
+    // Unload the collection from state.
+    yield put(deleteCollectionBySlug(slug));
```
willdurand: it is definitely a follow-up issue, but having an action named `deleteCollection` that calls `deleteCollectionBySlug` is super weird. I was like "oh shhhh, I missed this `deleteCollection` action when reviewing #5161" but actually, `deleteCollection` and `deleteCollectionBySlug` have different purposes.

Can we rename `deleteCollectionBySlug` to `unloadCollectionBySlug`?

bobsilverberg: Good idea. I will file a follow-up.

bobsilverberg: I opened https://github.com/mozilla/addons-frontend/issues/5238

Issue #4809, opened 2018-04-16
Paginate extensions and themes views on user profile so they are all accessible


ioanarusiczki:
Prerequisites:
Make sure that you have an account with more than 3 extensions and 6 themes submitted and approved.

STR:
1. Load AMO dev and go to your User Profile page.
2. Observe the displayed extensions and themes. 
3. Refresh page multiple times. 
4. Observe the displayed extensions and themes.
5. Check for a See more by Username link.

Expected results:
Different extensions are displayed after refreshing page or a See more link will show a complete list of extensions/themes.

Actual result:
The same extensions and themes are displayed. See more link is not available.

Notes:
This issue is reproducible on AMO dev with FF59 (Win10). 

![extensionsthemes](https://user-images.githubusercontent.com/33448286/38810613-fb14b20a-418f-11e8-978d-61b3be273ca3.png)

bobsilverberg:
I'm not sure about whether we should randomize the extensions displayed, like we do for the Hero elements on the home page, but I agree we should at least provide a link somewhere to see all of the extensions/themes for an author.

@jvillalobos what do you think this needs? @pwalm Where do you think the link should go?

jvillalobos:
I would prefer that we do two things:

1) We show more add-ons by default in the developer page. At least 10 of each kind.
2) Have pagination on those views, so it's possible to look at the next page of extensions or themes without having to go to a different page. This means the default view should not be randomized. Sorting by users is what makes most sense to me.

bobsilverberg:
@jvillalobos or @ioanarusiczki would one of you mind filing two separate issues for the suggestions at https://github.com/mozilla/addons-frontend/issues/4809#issuecomment-381748215? I suppose we could morph this issue into one of them, as it sounds like we are no longer requesting a "See more" link.

ioanarusiczki:
@jvillalobos Sounds like an elegant approach 👍

willdurand:
Hi @pwalm, would you have an input on this? We don't have paginated components yet, only pagination at a "page" level (e.g., search results). When trying to reuse the pagination below each card, it looks like this:

![screenshot-2018-5-28 user profile for add-ons for firefox](https://user-images.githubusercontent.com/217628/40609775-a31d6058-6267-11e8-9384-253f5b78d5f6.png)

I don't think it looks nice **but** I don't think a lot of developers have more than 10 extensions, but a few users have lots of themes (and likely no extensions)

[...]

PR #5191, opened 2018-06-05
Update the Paginate component

### LINKED ISSUE #3100: Pagination doesn't match the mocks

willdurand:
Fix #3100 – Refs #4809

---

This is going to be... huge :confused: The `Paginate` is old and I will have to iterate over it. The idea of this PR is:

- to put pagination inside the cards, according to the mockups
- to refresh the `Paginate` UI and be closer to the mockups ("closer", not exactly like the mockups)

Notes:

- I really need this to work on the last user profile features
- I rewrote the tests for the `Card` component to use Enzyme
- I focused on search results, collections, and reviews. I don't think the `Paginate` component is used elsewhere

## Screenshots

Search results:
screen shot 2018-06-06 at 15 40 54

Collection:
screen shot 2018-06-06 at 15 41 02

Reviews:
screen shot 2018-06-06 at 15 41 59

Collection small screen:
screen shot 2018-06-06 at 15 41 10

Search small screen:
screen shot 2018-06-06 at 15 41 22


willdurand:
Hi @bobsilverberg, would you have some time to look at it?

I have 2 main interrogation points:

- how do you feel about passing the `Paginate` component to the `Card` as `footer` prop? My idea is that it mimics what we are doing already.
- can we live with the update paginator UI for a while? (cf. screenshots)

bobsilverberg:
@willdurand Yes, I will find some time today to look at it.

src/ui/components/Card/index.js
```diff
@@ -18,32 +20,40 @@ export default class Card extends React.Component {
...
+      (footerText && footerLink) ||
+      (footerLink && footerNode) ||
+      (footerText && footerNode)
+    ) {
+      throw new Error(oneLine`You can only specify exactly one of these props:
+        footer, footerLink or footerText.`);
     } else if (footerText) {
       footer = footerText;
-      footerClass = 'Card-footer-text';
+      footerClass += ' Card-footer-text';
```
bobsilverberg: We tend to always use template string for this sort of thing (even though concatenation is briefer).

src/ui/components/Card/index.js
```diff
@@ -18,32 +20,40 @@ export default class Card extends React.Component {
...
     } else if (footerLink) {
       footer = footerLink;
-      footerClass = 'Card-footer-link';
+      footerClass += ' Card-footer-link';
```
bobsilverberg: We tend to always use template string for this sort of thing (even though concatenation is briefer).

src/ui/components/Card/index.js
```diff
@@ -53,28 +63,21 @@ export default class Card extends React.Component {
           'Card--no-header': !header,
           'Card--no-footer': !footer,
         })}
-        ref={(ref) => { this.cardContainer = ref; }}
```
bobsilverberg: We don't need these `ref`s anymore?

src/ui/components/Card/index.js
```diff
@@ -18,32 +20,40 @@ export default class Card extends React.Component {
...
     } = this.props;
 
     let footer;
-    let footerClass;
-    if (footerText && footerLink) {
-      throw new Error(
-        'You cannot specify footerLink and footerText at the same time');
+    let footerClass = 'Card-footer';
+
+    if (
```
bobsilverberg: This whole `footerText` vs `footerLink` seems like a lot of work, and from what I can tell the only difference between them is the class that is eventually used for the enclosing `div`. The contents are treated exactly the same. So I'm wondering if a different approach might yield cleaner code. E.g., always just accept a `footer` property, which could be text, a Link or a node. And then also accept a `footerClass` property, or maybe even determine the class from the type of object passed in the `footer` property. What do you think?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -530,15 +531,16 @@ describe(__filename, () => {
       dispatchAddon();
       dispatchAddonReviews();
       const root = render();
-      expect(root.find(Paginate))
+      expect(shallow(root.find('.AddonReviewList-reviews').prop('footer')))
```
bobsilverberg: I'm a bit concerned that we're not verifying that the `footer` property is in fact a `Paginator`. Is there a simple way to add that to the test? The same goes for the other tests which are verifying that the `Paginator` is correctly configured.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -567,7 +571,8 @@ describe(__filename, () => {
       const root = render({
         location: fakeRouterLocation({ query: { page: 3 } }),
```
bobsilverberg: It seems like we should make `page` a `const`, which equals `3`.

tests/unit/ui/components/TestCard.js
```diff
@@ -1,78 +1,103 @@
...
   it('uses photon class if marked', () => {
     const root = render({ children: 'hello', photonStyle: true });
-    expect(root.cardContainer.className).toContain('Card--photon');
+    expect(root).toHaveClassName('Card--photon');
   });
 
   it('shows header if supplied', () => {
     const root = render({ header: 'foo' });
-    expect(root.header).toBeTruthy();
+    expect(root.find('.Card-header')).toHaveLength(1);
```
bobsilverberg: Should we be verifying the text of the `header` as well?

src/ui/components/Card/index.js
bobsilverberg: Ah, it looks like we only needed them for the tests. Is that correct?

willdurand:
Hey @bobsilverberg, awesome! Thank you SO much for the review, I did not even expect that (I should have been more explicit to avoid you waste your time).

> I'm not sure what your second question "can we live with the update paginator UI for a while? (cf. screenshots)" means. 🤷‍♂️

Sorry, if you open Sketch, you'll see that the latest mockups show a different pagination UI:

screen shot 2018-06-06 at 21 34 54
It is not yet the same but it would be too much work I think. I tried to find a balance between what we need and what we want if that makes sense.

src/ui/components/Card/index.js
willdurand: Yes, we rarely use `ref` in source files.

src/ui/components/Card/index.js
willdurand: Yes, that makes sense but I was worried to break everything. Since we are on the same page, I think it would make sense to do it, but maybe in a follow-up patch?

tests/unit/amo/components/TestAddonReviewList.js
willdurand: Yes, I think we could do better here. First, introducing a variable named correctly and we can probably verify the component passed to this prop.

willdurand:
Ok, so I refactored the `SearchResults` component too. It feels good to remove old code 😋 

It is now ready for a real look, thanks @bobsilverberg!

src/core/components/PaginatorLink/index.js
```diff
@@ -40,10 +40,14 @@ export default class PaginatorLink extends React.Component {
...
     if (currentPage === page || page < 1 || page > pageCount) {
+      const classNames = makeClassName('Paginate-item', className, {
+        'Paginate-item--current-page': currentPage === page,
+      });
+
       return (
         <Button
-          buttonType="light"
-          className={makeClassName('Paginate-item', className)}
+          buttonType="cancel"
```
bobsilverberg: I know we're using this elsewhere, but it just occurred to me that it's odd we're using a `buttonType` of `cancel` when we're not actually creating a cancel button. Should we have a new type that's more about appearance and not function?

src/amo/components/SearchResults/index.js
```diff
@@ -1,72 +1,72 @@
...
-  static propTypes = {
-    count: PropTypes.number,
-    filters: PropTypes.object,
-    i18n: PropTypes.object.isRequired,
-    loading: PropTypes.bool,
-    results: PropTypes.arrayOf(PropTypes.object),
-  }
+type Props = {|
+  count: number,
+  filters: Object,
```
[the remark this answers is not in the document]
willdurand: I'd be happy to follow-up on that, I only converted this file to flow but did not mean to fix everything here. The PR is already big enough IMHO. I do agree that we should do better here, but can we open a new issue for Flow?

src/core/components/PaginatorLink/index.js
willdurand: Yes, we could rename that in a follow-up since this prop is not used as the html button type anyway.

willdurand:
> Thanks for the updates @willdurand. I have reviewed the code changes, but haven't had time to look at the tests yet. I may not have time to get to that today.

Thanks! I'd love to get this PR merged before the freeze today, because I need it to finish the user profile feature by the end of this month 😨 (quarter goal).

src/amo/components/SearchResults/index.js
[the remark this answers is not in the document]
bobsilverberg: I guess a follow-up is fine, although in the past we've tried to avoid that for flow issues, as sometimes the follow-up never gets done. But I'm sure you'll make sure it gets fixed. 😄

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -530,15 +531,23 @@ describe(__filename, () => {
...
     });
 
     it('configures a paginator with the right Link', () => {
       dispatchAddon();
       dispatchAddonReviews();
+
       const root = render();
-      expect(root.find(Paginate)).toHaveProp('LinkComponent', Link);
+      const footer = root.find('.AddonReviewList-reviews').prop('footer');
+      const paginator = shallow(footer);
```
bobsilverberg: Is it worth creating a `paginator` variable when you're only using it once below?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -549,25 +558,39 @@ describe(__filename, () => {
       ];
       dispatchAddon();
       dispatchAddonReviews({ reviews });
+
       const root = render();
-      expect(root.find(Paginate)).toHaveProp('count', reviews.length);
+      const footer = root.find('.AddonReviewList-reviews').prop('footer');
+      const paginator = shallow(footer);
```
bobsilverberg: Is it worth creating a paginator variable when you're only using it once below?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -549,25 +558,39 @@ describe(__filename, () => {
...
 
     it('sets the paginator to page 1 without a query', () => {
       dispatchAddon();
       dispatchAddonReviews();
+
       // Render with an empty query string.
       const root = render({ location: fakeRouterLocation() });
-      expect(root.find(Paginate)).toHaveProp('currentPage', 1);
+      const footer = root.find('.AddonReviewList-reviews').prop('footer');
+      const paginator = shallow(footer);
```
bobsilverberg: Is it worth creating a paginator variable when you're only using it once below?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -549,25 +558,39 @@ describe(__filename, () => {
...
       dispatchAddon();
       dispatchAddonReviews();
+
       const root = render({
-        location: fakeRouterLocation({ query: { page: 3 } }),
+        location: fakeRouterLocation({ query: { page } }),
       });
-      expect(root.find(Paginate)).toHaveProp('currentPage', 3);
+      const footer = root.find('.AddonReviewList-reviews').prop('footer');
+      const paginator = shallow(footer);
```
bobsilverberg: Is it worth creating a paginator variable when you're only using it once below?

tests/unit/core/components/TestPaginate.js
```diff
@@ -155,6 +159,31 @@ describe('<Paginate />', () => {
...
+
+        return root.instance().getCurrentPage();
+      };
+
+      it('returns the current page', () => {
+        expect(getCurrentPage({ currentPage: 123 })).toEqual(123);
+      });
+
+      // This happens when passing the value of a query string parameter
+      // directly to the component (hightly probable).
```
bobsilverberg: s/b `highly`

tests/unit/ui/components/TestCard.js
```diff
@@ -1,78 +1,104 @@
...
     const root = render({ footerLink: <a href="/some-link">Some link</a> });
-    expect(root.footer).toBeTruthy();
-    expect(root.footer.textContent).toEqual('Some link');
-    expect(root.footer.className).toEqual('Card-footer-link');
+    expect(root.find('.Card-footer')).toHaveLength(1);
+    expect(root.find('.Card-footer')).toHaveText('Some link');
+    expect(root.find('.Card-footer')).toHaveClassName('Card-footer-link');
+  });
+
+  it('throws an error both footerLink and footerText props are passed', () => {
```
bobsilverberg: s/b `throws an error *if* both footerLink and footerText props are passed`

tests/unit/ui/components/TestCard.js
```diff
@@ -1,78 +1,104 @@
...
+    expect(() => {
+      render({
+        footerLink: <a href="/some-link">Some link</a>,
+        footerText: 'something else',
+      });
+    }).toThrowError(/can only specify exactly one of these props/);
+  });
+
+  it('throws an error both footerLink and footer props are passed', () => {
```
bobsilverberg: s/b `throws an error *if* both footerLink and footer props are passed`

tests/unit/ui/components/TestCard.js
```diff
@@ -1,78 +1,104 @@
...
+      });
+    }).toThrowError(/can only specify exactly one of these props/);
   });
 
-  it('throws an error if you mix footer content', () => {
-    expect(() => render({
-      footerLink: <a href="/some-link">Some link</a>,
-      footerText: 'something else',
-    })).toThrowError(/cannot specify footerLink and footerText/);
+  it('throws an error both footer and footerText props are passed', () => {
```
bobsilverberg: s/b `throws an error *if* both footer and footerText props are passed`

src/core/components/Paginate/index.js
```diff
@@ -43,36 +49,45 @@ export class PaginateBase extends React.Component {
...
+
+    // We're showing the maximum number of pages on each side, start and end
+    // are correct.
     return makePageNumbers({ start, end });
   }
 
   render() {
     const {
-      LinkComponent, count, currentPage, i18n, pathname, queryParams,
+      LinkComponent,
```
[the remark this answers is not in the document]
willdurand: I think so yes.

src/core/components/PaginatorLink/index.js
willdurand: https://github.com/mozilla/addons-frontend/issues/5317

tests/unit/amo/components/TestAddonReviewList.js
willdurand: I've done that for readability purpose, because shallowing a footer does not really mean it is a `paginator` under the hood.

tests/unit/ui/components/TestCard.js
willdurand: heh, super good catch!

tests/unit/amo/components/TestAddonReviewList.js
willdurand: Ok, got it. I'll remove the variable and add a comment instead 😅

tests/unit/amo/components/TestAddonReviewList.js
willdurand: Added a comment instead of a variable.

PR #5202, opened 2018-06-05
I4810/I5077-combine static/lw theme results on author cards

### LINKED ISSUE #4810: Static themes are not displayed on the new User info card
Prerequisites:
Make sure you have an account with static themes submitted and approved

STR:
1. Load AMO dev and click on View Profile.
2. Observe the "More themes by" section.

Expected result:
Static themes are displayed in the "More themes" section.

Actual result:
Only lightweight themes are displayed.

Notes: 

static themes

### LINKED ISSUE #5077: Combined theme (lw and static) results are not displayed together on the AddonsByAuthorsCard (on addon detail page)
go to a static theme detail page:
https://addons-dev.allizom.org/en-US/firefox/addon/rm-static-theme-3/?src=search

or go to lwt detail page:
https://addons-dev.allizom.org/en-US/firefox/addon/rm-test-new-theme/?src=search

you will only see "more themes" of that same type 

if an author has different types of themes it should be combined here

rebmullin:
fixes #4810 
fixes #5077 

User info card - before:
Alt text

User info card - after:
Alt text

Static theme detail page- before (more themes only show other static themes):
Alt text

Lightweight theme detail page - before (more themes only show other lightweight themes):
Alt text

Static theme detail page - after:
Alt text

Lightweight theme detail page - after:
Alt text

### DIFF
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -19,6 +19,7 @@ module.exports = {
     'cookieMaxAge',
     'cookieSecure',
     'enableClientConsole',
+    'enableStaticThemes',
     'defaultLang',
     'isDeployed',
     'isDevelopment',
@@ -75,4 +76,7 @@ module.exports = {
 
   // https://sentry.prod.mozaws.net/operations/addons-frontend-disco-prod/
   publicSentryDsn: 'https://b9e70d0dca144344a7a5674c29b08355@sentry.prod.mozaws.net/186',
+
+  // Enable static themes.
+  enableStaticThemes: false,
 };
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -76,7 +76,4 @@ module.exports = {
 
   // https://sentry.prod.mozaws.net/operations/addons-frontend-disco-prod/
   publicSentryDsn: 'https://b9e70d0dca144344a7a5674c29b08355@sentry.prod.mozaws.net/186',
-
-  // Enable static themes.
-  enableStaticThemes: false,
 };
--- a/config/default-disco.js
+++ b/config/default-disco.js
@@ -19,7 +19,6 @@ module.exports = {
     'cookieMaxAge',
     'cookieSecure',
     'enableClientConsole',
-    'enableStaticThemes',
     'defaultLang',
     'isDeployed',
     'isDevelopment',
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -2,7 +2,7 @@
 import deepcopy from 'deepcopy';
 import invariant from 'invariant';
 
-import { ADDON_TYPE_THEMES_SEARCH } from 'core/constants';
+import { ADDON_TYPE_THEMES_FILTER } from 'core/constants';
 import { createInternalAddon } from 'core/reducers/addons';
 import type {
   ExternalAddonType,
@@ -190,7 +190,7 @@ const reducer = (
     case LOAD_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
       const { addonType } = action.payload;
-      const pageSize = addonType === ADDON_TYPE_THEMES_SEARCH ?
+      const pageSize = addonType === ADDON_TYPE_THEMES_FILTER ?
         THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
 
       if (action.payload.forAddonSlug) {
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -2,7 +2,7 @@
 import deepcopy from 'deepcopy';
 import invariant from 'invariant';
 
-import { ADDON_TYPE_THEMES, ADDON_TYPE_THEMES_FILTER } from 'core/constants';
+import { ADDON_TYPE_THEMES } from 'core/constants';
 import { createInternalAddon } from 'core/reducers/addons';
 import type {
   ExternalAddonType,
@@ -154,8 +154,8 @@ export const getAddonsForUsernames = (
       return state.byAddonId[id];
     })
     .filter((addon) => {
-      const type = getAddonTypeFilter(addonType);
-      return addonType ? type.includes(addon.type) : true;
+      const addonTypeFilter = getAddonTypeFilter(addonType);
+      return addonType ? addonTypeFilter.includes(addon.type) : true;
     })
     .filter((addon) => {
       return addon.slug !== excludeSlug;
@@ -174,8 +174,7 @@ const reducer = (
   switch (action.type) {
     case FETCH_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
-      const { addonType: type } = action.payload;
-      const addonType = getAddonTypeFilter(type);
+      const { addonType } = action.payload;
 
       if (action.payload.forAddonSlug) {
         newState.byAddonSlug = {
@@ -194,12 +193,10 @@ const reducer = (
     }
     case LOAD_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
-      const { addonType: type } = action.payload;
+      const { addonType } = action.payload;
 
-      const addonType = getAddonTypeFilter(type);
-      const pageSize = ADDON_TYPE_THEMES.includes(type) ||
-        addonType === ADDON_TYPE_THEMES_FILTER ? THEMES_BY_AUTHORS_PAGE_SIZE :
-        EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
+      const pageSize = ADDON_TYPE_THEMES.includes(addonType) ?
+        THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
 
       if (action.payload.forAddonSlug) {
         newState.byAddonSlug = {
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -154,7 +154,6 @@ export const getAddonsForUsernames = (
       return state.byAddonId[id];
     })
     .filter((addon) => {
-      // return addonType ? addon.type === addonType : true;
       const addonTypeFilter = getAddonTypeFilter(addonType);
       return addonType ? addonTypeFilter.includes(addon.type) : true;
     })
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -154,6 +154,7 @@ export const getAddonsForUsernames = (
       return state.byAddonId[id];
     })
     .filter((addon) => {
+      // return addonType ? addon.type === addonType : true;
       const addonTypeFilter = getAddonTypeFilter(addonType);
       return addonType ? addonTypeFilter.includes(addon.type) : true;
     })
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -174,33 +174,33 @@ const reducer = (
   switch (action.type) {
     case FETCH_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
-      const { addonType } = action.payload;
+      const { addonType, authorUsernames, forAddonSlug } = action.payload;
 
-      if (action.payload.forAddonSlug) {
+      if (forAddonSlug) {
         newState.byAddonSlug = {
           ...newState.byAddonSlug,
-          [action.payload.forAddonSlug]: undefined,
+          [forAddonSlug]: undefined,
         };
       }
 
       newState.loadingFor[joinAuthorNamesAndAddonType(
-        action.payload.authorUsernames, addonType)] = true;
+        authorUsernames, addonType)] = true;
 
       newState.byAuthorNamesAndAddonType[joinAuthorNamesAndAddonType(
-        action.payload.authorUsernames, addonType)] = null;
+        authorUsernames, addonType)] = null;
 
       return newState;
     }
     case LOAD_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
-      const { addonType } = action.payload;
+      const { addonType, authorUsernames, forAddonSlug } = action.payload;
 
       const pageSize = ADDON_TYPE_THEMES.includes(addonType) ?
         THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
 
-      if (action.payload.forAddonSlug) {
+      if (forAddonSlug) {
         newState.byAddonSlug = {
-          [action.payload.forAddonSlug]: action.payload.addons
+          [forAddonSlug]: action.payload.addons
             .slice(0, pageSize)
             .map((addon) => addon.id),
         };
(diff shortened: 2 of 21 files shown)


src/core/constants.js
```diff
@@ -48,6 +49,8 @@ export const ADDON_TYPE_LANG = 'language';
 export const ADDON_TYPE_OPENSEARCH = 'search';
 export const ADDON_TYPE_STATIC_THEME = 'statictheme';
 export const ADDON_TYPE_THEME = 'persona';
+export const ADDON_TYPE_THEMES_FILTER = config.get('enableStaticThemes') ?
+  `${ADDON_TYPE_THEME},${ADDON_TYPE_STATIC_THEME}` : ADDON_TYPE_THEME;
```
rebmullin: should I be testing this config setting?

willdurand: I think so yes, and that's why I don't think we should be doing this.

Could we have a `getAddonTypeFilter()` helper function instead? I think this method could take the `addon` and `_config = config` as arguments, and return the filter for the API call (which is located in the sagas, right?)

I would also try to avoid replacing `addon.type` to `addon.addonFilter` as much as I could (ideally we should not add `addonFilter` to the addon type, but I might miss something).

rebmullin: @willdurand, thanks for quick feedback. Let me revisit this..

rebmullin:
hey @willdurand, I know I am getting some travis errors (which i have to look into more), but can you please take another look when you can?

src/amo/reducers/addonsByAuthors.js
```diff
@@ -172,6 +174,8 @@ const reducer = (
   switch (action.type) {
     case FETCH_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
+      const { addonType: type } = action.payload;
+      const addonType = getAddonTypeFilter(type);
```
willdurand: Why not having `addonTypeFilter` as a variable here?

src/amo/reducers/addonsByAuthors.js
```diff
@@ -181,16 +185,21 @@ const reducer = (
...
     }
     case LOAD_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
-      const pageSize = action.payload.addonType === ADDON_TYPE_THEME ?
-        THEMES_BY_AUTHORS_PAGE_SIZE : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
+      const { addonType: type } = action.payload;
+
+      const addonType = getAddonTypeFilter(type);
+      const pageSize = ADDON_TYPE_THEMES.includes(type) ||
+        addonType === ADDON_TYPE_THEMES_FILTER ? THEMES_BY_AUTHORS_PAGE_SIZE :
```
willdurand: Why do you want to check `ADDON_TYPE_THEMES_FILTER`?

src/amo/sagas/addonsByAuthors.js
```diff
@@ -18,13 +19,14 @@ export function* fetchAddonsByAuthors({ payload }) {
   yield put(errorHandler.createClearingAction());
 
   try {
+    const type = getAddonTypeFilter(addonType);
```
willdurand: We should probably stick to "addonTypeFilter" because it is a bit confusing to mix type, addonType, and addonTypeFilter.

src/core/utils/index.js
```diff
@@ -317,3 +319,12 @@ export const decodeHtmlEntities = (string) => {
...
 };
+
+export const getAddonTypeFilter = (addonType, _config = config) => {
+  const isTheme = ADDON_TYPE_THEMES.includes(addonType);
+  if (!_config.get('enableStaticThemes') || !isTheme) {
+    return addonType;
+  }
+
+  return ADDON_TYPE_THEMES_FILTER;
+};
```
willdurand: 👍

src/core/constants.js
```diff
@@ -60,7 +61,8 @@ export type AddonTypeType =
   | typeof ADDON_TYPE_LANG
   | typeof ADDON_TYPE_OPENSEARCH
   | typeof ADDON_TYPE_STATIC_THEME
-  | typeof ADDON_TYPE_THEME;
+  | typeof ADDON_TYPE_THEME
+  | typeof ADDON_TYPE_THEMES_FILTER;
```
willdurand: Why do we need to treat `ADDON_TYPE_THEMES_FILTER` as an add-on type? To me, it is only a filter for the API, not a type.

src/amo/reducers/addonsByAuthors.js
willdurand: Or maybe just `typeFilter` or `type`...  And the `joinAuthorNamesAndAddonType()` function should be renamed to `joinAuthorNamesAndType()` or `joinAuthorNamesAndTypeFilter()`:

```js
const { addonType } = action.payload;
const type = getAddonTypeFilter(addonType);

joinAuthorNamesAndType(action.payload.authorUsernames, type)
```

willdurand: mmh another question: do we actually have to modify this reducer?

src/core/constants.js
rebmullin: good catch, thanks

src/amo/reducers/addonsByAuthors.js
rebmullin: I was using the modified type in saga for this reducer so addonType came through as the modified type but it seems we can remove this here too

src/amo/reducers/addonsByAuthors.js
rebmullin: you're right..I thought we needed this but we don't :/

rebmullin: Actually, I don't think this update is even needed...
I think I can keep most of this in saga/places where we need to filter

rebmullin:
thanks @willdurand for all the feedback. can you please take another look when you can?

src/amo/reducers/addonsByAuthors.js
```diff
@@ -153,7 +154,8 @@ export const getAddonsForUsernames = (
       return state.byAddonId[id];
     })
     .filter((addon) => {
-      return addonType ? addon.type === addonType : true;
+      const addonTypeFilter = getAddonTypeFilter(addonType);
+      return addonType ? addonTypeFilter.includes(addon.type) : true;
```
willdurand: What is the type of `addonTypeFilter`? It is a `string`, no? Does `includes()` work on strings?

rebmullin: yep : ) [strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) and [arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) : )

src/core/reducers/addons.js
```diff
@@ -94,7 +97,7 @@ export function loadAddonResults(
 }
 
 export function getGuid(addon: ExternalAddonType): string {
-  if (addon.type === ADDON_TYPE_THEME) {
+  if (ADDON_TYPE_THEMES.includes(addon.type)) {
```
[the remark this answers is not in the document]
rebmullin: yeh i'll revert this..

src/amo/reducers/addonsByAuthors.js
willdurand: ok, cool. it is a bit black magic but I guess it is fine here.

src/core/constants.js
```diff
@@ -48,6 +48,7 @@ export const ADDON_TYPE_LANG = 'language';
 export const ADDON_TYPE_OPENSEARCH = 'search';
 export const ADDON_TYPE_STATIC_THEME = 'statictheme';
 export const ADDON_TYPE_THEME = 'persona';
+export const ADDON_TYPE_THEMES_FILTER = 'persona,statictheme';
```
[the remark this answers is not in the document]
rebmullin: yeh I noticed this too and can def update

tests/unit/core/utils/test_index.js
```diff
@@ -706,4 +709,30 @@ describe(__filename, () => {
...
+    });
+
+    it('enabledStaticThemes set to false in config will return ADDON_TYPE_THEME', () => {
+      const fakeConfig = getFakeConfig({ enableStaticThemes: false });
+      const addon = createInternalAddon({ type: ADDON_TYPE_THEME, config: fakeConfig });
+      expect(getAddonTypeFilter(addon.type, fakeConfig)).toEqual(ADDON_TYPE_THEME);
+    });
+
+    it('enabledStaticThemes set to true in config will return ADDON_TYPE_EXTENSION if it is not a theme', () => {
+      const fakeConfig = getFakeConfig({ enableStaticThemes: true });
```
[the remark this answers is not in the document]
rebmullin: yep, that sounds like a good idea : )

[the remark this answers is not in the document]
rebmullin: actually...this is what this test case already covers. let me know if I a missing something tho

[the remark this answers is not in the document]
willdurand: ah oops, I wanted to write `enableStaticThemes` set to `false`

config/default-disco.js
```diff
@@ -75,4 +76,7 @@ module.exports = {
 
   // https://sentry.prod.mozaws.net/operations/addons-frontend-disco-prod/
   publicSentryDsn: 'https://b9e70d0dca144344a7a5674c29b08355@sentry.prod.mozaws.net/186',
+
+  // Enable static themes.
+  enableStaticThemes: false,
```
willdurand: I am not sure we need to edit this file actually.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -172,29 +174,33 @@ const reducer = (
   switch (action.type) {
     case FETCH_ADDONS_BY_AUTHORS: {
       const newState = deepcopy(state);
+      const { addonType, authorUsernames, forAddonSlug } = action.payload;
```
willdurand: Just to be sure: is there anything that has changed in this `case` block? It looks like nothing has been changed besides destructuring `action.payload`.

src/amo/sagas/addonsByAuthors.js
```diff
@@ -18,13 +19,14 @@ export function* fetchAddonsByAuthors({ payload }) {
   yield put(errorHandler.createClearingAction());
 
   try {
+    const addonTypeFilter = getAddonTypeFilter(addonType);
```
willdurand: We might not need this variable, we can inline the call to `getAddonTypeFilter()` in the `filters` object below.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -563,6 +594,17 @@ describe(__filename, () => {
...
+  it('shows extensions in header for a static theme with multiple authors and without More text ', () => {
+    const root = renderAddonsWithType({
+      addonType: ADDON_TYPE_STATIC_THEME,
+      showMore: false,
+      multipleAuthors: true,
+    });
+
+    expect(root.find(AddonsCard))
+      .toHaveProp('header', 'Themes by these artists');
+  });
```
willdurand: Thanks for adding coverage here, I wonder why it was not done before since you did not make any changes to `AddonsByAuthorsCard` in this patch.

src/amo/reducers/addonsByAuthors.js
willdurand: We should probably add a test case for this update, I see that we have a `getAddonsForUsernames selector` describe block in the test file.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
rebmullin: I *think* it was here before.. Could it have gotten lost in a rebase?

src/amo/reducers/addonsByAuthors.js
rebmullin: yeh in these 2 cases here it seems that is all that has changed at this point..

tests/unit/amo/components/TestAddonsByAuthorsCard.js
willdurand: mmh, interesting. Maybe GitHub shows the wrong diff. No worries, we'll see right before you merge this one :)

src/amo/reducers/addonsByAuthors.js
rebmullin: hmm i struggled with these tests. Let me know if what I have is enough :/

config/default-disco.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
     'cookieMaxAge',
     'cookieSecure',
     'enableClientConsole',
+    'enableStaticThemes',
```
willdurand: I don't think you need to add this entry here.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -153,7 +154,9 @@ export const getAddonsForUsernames = (
       return state.byAddonId[id];
     })
     .filter((addon) => {
-      return addonType ? addon.type === addonType : true;
+      // return addonType ? addon.type === addonType : true;
```
willdurand: Can we remove this line instead of commenting it?

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -18,15 +23,17 @@ describe(__filename, () => {
   const fakeAuthorTwo = { ...fakeAuthor, username: 'test2', id: 61 };
   const fakeAuthorThree = { ...fakeAuthor, username: 'test3', id: 71 };
 
-  function fakeAddons() {
+  function fakeAddons(type = ADDON_TYPE_EXTENSION) {
```
willdurand: We usually use an object to gather all the arguments of a function, because it is more convenient if we want to add new arguments (and also because we don't have to remember the order of the args).

```js
function fakeAddons({ type = ADDON_TYPE_EXTENSION } = {}) {
  // ...
}
```

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -509,6 +529,61 @@ describe(__filename, () => {
       ]);
     });
 
+    it("displays lightweight theme types when filtering for authors' themes", () => {
```
willdurand: Could we write `(it) returns lightweight themes when [...]` instead?

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -509,6 +529,61 @@ describe(__filename, () => {
...
+
+      expect(getAddonsForUsernames(
+        state, authorUsernames, ADDON_TYPE_THEME
+      )).toEqual([
+        createInternalAddon(addons.firstAddon),
+        createInternalAddon(addons.secondAddon),
+      ]);
+    });
+
+    it("displays static theme types when filtering for authors' themes ", () => {
```
willdurand: Could we write `(it) returns static themes when [...]` instead?

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -509,6 +529,61 @@ describe(__filename, () => {
...
+
+      expect(getAddonsForUsernames(
+        state, authorUsernames, ADDON_TYPE_STATIC_THEME
+      )).toEqual([
+        createInternalAddon(addons.firstAddon),
+        createInternalAddon(addons.secondAddon),
+      ]);
+    });
+
+    it("displays extension types when filtering for authors' extensions", () => {
```
willdurand: Could we write `(it) returns extensions when [...]` instead?

willdurand: Also I am a bit surprised that we did not have tests for that before... 🤷‍♂️

src/amo/reducers/addonsByAuthors.js
rebmullin: yes that was not intended :/

PR #5208, opened 2018-06-06
Hide notifications information for non-developer users

### LINKED ISSUE #5206: Information text regarding 'hosted addons' should not be present in the notifications section for regular users
STR:

1. Log in to AMO dev with a non-developer account
2. Click on Edit My Profile
3. Scroll down to the Notifications section
4. Observe the information text present in the section

Actual result:
There is an information text explaining that the user might receive notifications concerning his hosted addons

Expected result:
Since the user is not a developer, this information text should not be present in the notifications section

Notes:

image

willdurand:
Fix #5206

---

This PR introduces a `isDeveloper()` selector to quickly determine
whether a user is an artist or add-on developer.

Before:

After:

screen shot 2018-06-06 at 11 52 17

### DIFF
--- a/tests/unit/amo/components/TestUserProfileEdit.js
+++ b/tests/unit/amo/components/TestUserProfileEdit.js
@@ -360,6 +360,31 @@ describe(__filename, () => {
       forbidden.`
     );
 
+    expect(root.find('.UserProfileEdit-notifications--help')).toHaveLength(0);
+  });
+
+  it('renders a help text about add-on notifications for users who are developers', () => {
+    const root = renderUserProfileEdit({
+      userProps: {
+        ...defaultUserProps,
+        is_addon_developer: true,
+      },
+    });
+
+    expect(root.find('.UserProfileEdit-notifications--help')).toHaveText(
+      oneLine`Mozilla reserves the right to contact you individually about
+      specific concerns with your hosted add-ons.`
+    );
+  });
+
+  it('renders a help text about add-on notifications for users who are artists', () => {
+    const root = renderUserProfileEdit({
+      userProps: {
+        ...defaultUserProps,
+        is_artist: true,
+      },
+    });
+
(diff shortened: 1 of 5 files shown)


tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -360,6 +360,31 @@ describe(__filename, () => {
...
+    const root = renderUserProfileEdit({
+      userProps: {
+        ...defaultUserProps,
+        is_artist: true,
+      },
+    });
+
     expect(root.find('.UserProfileEdit-notifications--help')).toHaveText(
       oneLine`Mozilla reserves the right to contact you individually about
       specific concerns with your hosted add-ons.`
```
rebmullin: should we add a test if user is_artist or is_addon_developer but they're not editing the current user, it should not appear? for example, if look at this page as admin:  `/en-US/firefox/user/Contdedisplaytestadmin/edit/`, I don't see it

willdurand: I don't think so because it is only shown to current users anyway (see the diff, `isEditingCurrentUser` was already there).

rebmullin: you're right that was already there but i guess it brought to my attention (since it's related) that maybe we should have a test for this, no? or is there an assumption that the user is typically the editing user so it doesn't need to be tested?

willdurand: @rebmullin I think it is covered by the assertions in this test case: https://github.com/mozilla/addons-frontend/blob/60a74fd4d6af21df150faa53f031f05496d6dcb4/tests/unit/amo/components/TestUserProfileEdit.js#L897-L900. WDYT?

rebmullin: @willdurand  ..ahh I didn't see this.. cool then 👍

PR #5213, opened 2018-06-06
Make sure we never pass null to a textarea

### LINKED ISSUE #5212: A 'null' value is set in the biography field when the user saves his profile without adding a biography
STR:

1. Load AMO-dev
2. Create e new account and go through all the confirmation steps until you land on the Edit profile page
3. Update your profile information, making sure that you also include a picture
4. Leave the Biography field empty
5. Click on Update Profile then go to View My Profile
6. Check the biography section

Actual result
There is a 'null' value present in the biography section on the User profile page 

Expected result:
The biography section on the User profile page should be empty 

Notes:
- when navigating to the Edit screen again, the 'null' value is also present in the biography field
- I was able to reproduce only when updating the profile for a new user
- the issue will reproduce only when a profile picture is added
- reproduced on AMO-dev with Firefox 60, Win10x64

bio null

willdurand:
Fix #5212

---

This PR fixes an issue caused by a `textarea` component, which does not
accept `null` as a valid value. It outputs a warning in the dev tools
but also converts `null` to `"null"` (string) on submit, causing the
issue reported in #5212.

### DIFF
--- a/src/amo/api/users.js
+++ b/src/amo/api/users.js
@@ -35,7 +35,8 @@ export function editUserAccount({ api, picture, userId, ...editableFields }: {|
     const form = new FormData();
     // Add all the editable fields, one by one.
     Object.keys(editableFields).forEach((key: string) => {
-      form.set(key, editableFields[key]);
+      // We cannot send `null` values, only string values.
+      form.set(key, editableFields[key] === null ? '' : editableFields[key]);
     });
     // Add the picture file.
     form.set('picture_upload', picture);
--- a/src/amo/api/users.js
+++ b/src/amo/api/users.js
@@ -35,8 +35,9 @@ export function editUserAccount({ api, picture, userId, ...editableFields }: {|
     const form = new FormData();
     // Add all the editable fields, one by one.
     Object.keys(editableFields).forEach((key: string) => {
-      // We cannot send `null` values, only string values.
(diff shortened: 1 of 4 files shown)


willdurand:
Oh 😕

willdurand:
> I think we have to dig into why this only happens on uploading an picture. Otherwise, you do not see this happen...

Indeed, great hint, thanks!

`FormData` cannot send `null` values, that's why it gets sent as `"null"`. I fixed the API call to convert `null` values to empty strings when we send `FormData` body (which only happens when the picture is uploaded). What do you think?

src/amo/api/users.js
```diff
@@ -35,7 +35,8 @@ export function editUserAccount({ api, picture, userId, ...editableFields }: {|
     const form = new FormData();
     // Add all the editable fields, one by one.
     Object.keys(editableFields).forEach((key: string) => {
-      form.set(key, editableFields[key]);
+      // We cannot send `null` values, only string values.
+      form.set(key, editableFields[key] === null ? '' : editableFields[key]);
```
rebmullin: is it worth it to do something like this ~:
```javascript
const fieldValue = editableFields[key];
form.set(key, fieldValue === null ? '' : fieldValue);
```

willdurand: sure!

PR #5218, opened 2018-06-06
Fetch user notifications after profile picture deletion

### LINKED ISSUE #5214: Notification options are stuck in a loading state after the user delets his profile picture
STR:

1. Log in to AMO-dev
2. Go to Edit My Profile and set a profile picture if you don't have one already
3. Delete your profile picture 
4. Check the Notifications section

Actual result
Notification options are stuck in a loading state after the profile picture is deleted

Expected result:
Notification options should not be affected when a user deletes his profile picture 

Notes:
- reproduced on AMO -dev with FF60 Win10x64

delete pic

willdurand:
Fix #5214

---

This PR guarantees that the user notifications are (re)loaded after the
user has deleted their profile picture. It was not the case before this
PR because, after the profile picture deletion, we load the new user
data (from the `DELETE` API call response), thus removing the stored
user notifications for this user.

### DIFF
--- a/tests/unit/amo/components/TestUserProfileEdit.js
+++ b/tests/unit/amo/components/TestUserProfileEdit.js
@@ -1081,6 +1081,39 @@ describe(__filename, () => {
     expect(root).toHaveState('pictureData', null);
   });
 
+  it('fetches the user notifications after having deleted a profile picture', () => {
+    const username = 'tofumatt';
+    const { store } = dispatchSignInActions({
+      userProps: {
+        ...defaultUserProps,
+        picture_url: 'https://example.org/pp.png',
+        username,
+      },
+    });
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const errorHandler = createStubErrorHandler();
+
+    const root = renderUserProfileEdit({ errorHandler, store });
+    const user = getCurrentUser(store.getState().users);
+
+    dispatchSpy.reset();
+    sinon.assert.notCalled(dispatchSpy);
+
+    // The user profile picture has been successfully deleted.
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1081,6 +1081,39 @@ describe(__filename, () => {
...
+      },
+    });
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const errorHandler = createStubErrorHandler();
+
+    const root = renderUserProfileEdit({ errorHandler, store });
+    const user = getCurrentUser(store.getState().users);
+
+    dispatchSpy.reset();
+    sinon.assert.notCalled(dispatchSpy);
```
bobsilverberg: This seems a bit odd to assert that the `dispatchSpy` was not called immediately after you `reset` it. Doesn't resetting it guarantee that it won't have been called at this point?

willdurand:
> Is it worth considering loading just the user data and not overwriting the notifications with empty notifications, or would that just complicate things more than it's worth?
> 
> That's interesting. I don't have an immediate answer but I will investigate!

I think that could work, closing and reopening a PR!

willdurand:
@bobsilverberg thanks a bunch for your proposal, I feel a bit stupid because it SO much better and I did not think about it at all... Just opened #5253.

bobsilverberg:
If I felt stupid every time you suggested something I hadn't thought of, I'd feel very stupid indeed! I.e., no reason to feel bad, that's why we have reviews! I'm glad my suggestion was useful. 😄

PR #5225, opened 2018-06-07
I5174-fix for search listing/mobile context

rebmullin:
potential fix/brainstorming for #5174

I was playing around with the styles and thought in these contexts - for the search listing view and detail header (below bp 900px) -  that these looked okay:

search listing:
Alt text

header on detail:
Alt text

any thoughts? or are they too small?

altho, we still need to figure out the following context if the alignment is still left:
Alt text

there are a few options to maybe help:
-remove or change to height auto (for all breakpoints)
-use a similar listing view as search (or even maybe 2 cols, vs the current 3 for larger bps -if Ux signs off), for eg:
Alt text
-discuss alignment right for static themes?

also we need to look at home, landing pgs, etc where it's 4 cols ..

cc @muffinresearch @eviljeff

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -150,11 +150,6 @@ export class AddonBase extends React.Component {
       const previewHeader = addon.previews.length && addon.previews[0];
 
       let previewURL =
-        previewHeader && previewHeader.thumbnail_url
-          ? previewHeader.thumbnail_url
-          : null;
-
-      const previewUrlLarge =
         previewHeader && previewHeader.image_url
           ? previewHeader.image_url
           : null;
@@ -171,23 +166,11 @@ export class AddonBase extends React.Component {
         previewURL = addon.previewURL;
       }
 
-      const imageAtts = {};
-      if (previewUrlLarge) {
-        const imageWidth = previewHeader && previewHeader.image_size[0];
-        const thumbWidth = previewHeader && previewHeader.thumbnail_size[0];
-        if (imageWidth && thumbWidth) {
-          // If viewing on retina, it should only show the larger size with
-          // the current widths available
-          imageAtts.srcSet = `${previewURL} ${thumbWidth}w, ${previewUrlLarge} ${imageWidth}w`;
-        }
-      }
-
       const headerImage = (
         <img
           alt={label}
           className="Addon-theme-header-image"
           src={previewURL}
-          {...imageAtts}
         />
       );
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -148,8 +148,16 @@ export class AddonBase extends React.Component {
 
     if (this.addonIsTheme()) {
       const previewHeader = addon.previews.length && addon.previews[0];
-      let previewURL = previewHeader.thumbnail_url || null;
-      const previewURLLarge = previewHeader.image_url || null;
+
+      let previewURL =
+        previewHeader && previewHeader.thumbnail_url
+          ? previewHeader.thumbnail_url
+          : null;
+
+      const previewUrlLarge =
+        previewHeader && previewHeader.image_url
+          ? previewHeader.image_url
+          : null;
 
       const label = addon
         ? i18n.sprintf(i18n.gettext('Preview of %(title)s'), {
@@ -164,13 +172,13 @@ export class AddonBase extends React.Component {
       }
 
       const imageAtts = {};
-      if (previewURLLarge) {
-        const imageSize = previewHeader.image_size[0];
-        const thumbSize = previewHeader.thumbnail_size[0];
-        if (imageSize && thumbSize) {
+      if (previewUrlLarge) {
+        const imageWidth = previewHeader && previewHeader.image_size[0];
+        const thumbWidth = previewHeader && previewHeader.thumbnail_size[0];
+        if (imageWidth && thumbWidth) {
           // If viewing on retina, it should only show the larger size with
           // the current widths available
-          imageAtts.srcSet = `${previewURL} ${thumbSize}w, ${previewURLLarge} ${imageSize}w`;
+          imageAtts.srcSet = `${previewURL} ${thumbWidth}w, ${previewUrlLarge} ${imageWidth}w`;
         }
       }
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -147,10 +147,9 @@ export class AddonBase extends React.Component {
     const { addon, i18n } = this.props;
 
     if (this.addonIsTheme()) {
-      let previewURL =
-        addon.previews.length > 0 && addon.previews[0].image_url
-          ? addon.previews[0].image_url
-          : null;
+      const previewHeader = addon.previews.length && addon.previews[0];
+      let previewURL = previewHeader.thumbnail_url || null;
+      const previewURLLarge = previewHeader.image_url || null;
 
       const label = addon
         ? i18n.sprintf(i18n.gettext('Preview of %(title)s'), {
@@ -164,11 +163,24 @@ export class AddonBase extends React.Component {
         previewURL = addon.previewURL;
       }
 
+      const imageAtts = {};
+      if (previewURLLarge) {
+        const imageSize = previewHeader.image_size[0];
+        const thumbSize = previewHeader.thumbnail_size[0];
+        if (imageSize && thumbSize) {
+          // If viewing on retina, it should only show the larger size with
+          // the current widths available
+          imageAtts.srcSet = `${previewURL} ${thumbSize}w, ${previewURLLarge} ${imageSize}w`;
+        }
+      }
+
+      const imageClassName = 'Addon-theme-header-image';
       const headerImage = (
         <img
           alt={label}
-          className="Addon-theme-header-image"
+          className={imageClassName}
           src={previewURL}
+          {...imageAtts}
         />
       );
 
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -182,11 +182,10 @@ export class AddonBase extends React.Component {
         }
       }
 
-      const imageClassName = 'Addon-theme-header-image';
       const headerImage = (
         <img
           alt={label}
-          className={imageClassName}
+          className="Addon-theme-header-image"
           src={previewURL}
           {...imageAtts}
         />
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -13,6 +13,7 @@ import { setViewContext } from 'amo/actions/viewContext';
 import CategoryHeader from 'amo/components/CategoryHeader';
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import NotFound from 'amo/components/ErrorPage/NotFound';
+import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { categoriesFetch } from 'core/actions/categories';
 import {
   ADDON_TYPE_EXTENSION,
@@ -252,10 +253,18 @@ export class CategoryBase extends React.Component {
 
     const { html } = this.contentForType(addonType);
 
+    const isAddonTheme = isTheme(addonType);
+
+    const themeProps = {};
+
+    if (isAddonTheme) {
+      themeProps.placeholderCount = LANDING_PAGE_THEME_COUNT;
+    }
+
     return (
       <div
         className={makeClassName('Category', {
-          'Category--theme': isTheme(addonType),
+          'Category--theme': isAddonTheme,
         })}
       >
         {category && (
@@ -278,7 +287,7 @@ export class CategoryBase extends React.Component {
             footerLink={html.featuredFooterLink}
             header={html.featuredHeader}
             loading={loading}
-            isTheme={isTheme(addonType)}
+            {...themeProps}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -291,7 +300,7 @@ export class CategoryBase extends React.Component {
             footerText={html.highlyRatedFooterText}
             header={html.highlyRatedHeader}
             loading={loading}
-            isTheme={isTheme(addonType)}
+            {...themeProps}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -304,7 +313,7 @@ export class CategoryBase extends React.Component {
             footerText={html.trendingFooterText}
             header={html.trendingHeader}
             loading={loading}
-            isTheme={isTheme(addonType)}
+            {...themeProps}
           />,
         )}
       </div>
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -13,7 +13,6 @@ import { setViewContext } from 'amo/actions/viewContext';
 import CategoryHeader from 'amo/components/CategoryHeader';
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import NotFound from 'amo/components/ErrorPage/NotFound';
-import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { categoriesFetch } from 'core/actions/categories';
 import {
   ADDON_TYPE_EXTENSION,
@@ -255,12 +254,6 @@ export class CategoryBase extends React.Component {
 
     const isAddonTheme = isTheme(addonType);
 
-    const themeProps = {};
-
-    if (isAddonTheme) {
-      themeProps.placeholderCount = LANDING_PAGE_THEME_COUNT;
-    }
-
     return (
       <div
         className={makeClassName('Category', {
@@ -286,8 +279,8 @@ export class CategoryBase extends React.Component {
             footerText={html.featuredFooterText}
             footerLink={html.featuredFooterLink}
             header={html.featuredHeader}
+            isTheme={isAddonTheme}
             loading={loading}
-            {...themeProps}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -299,8 +292,8 @@ export class CategoryBase extends React.Component {
             footerLink={html.highlyRatedFooterLink}
             footerText={html.highlyRatedFooterText}
             header={html.highlyRatedHeader}
+            isTheme={isAddonTheme}
             loading={loading}
-            {...themeProps}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -312,8 +305,8 @@ export class CategoryBase extends React.Component {
             footerLink={html.trendingFooterLink}
             footerText={html.trendingFooterText}
             header={html.trendingHeader}
+            isTheme={isAddonTheme}
             loading={loading}
-            {...themeProps}
           />,
         )}
       </div>
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -278,6 +278,7 @@ export class CategoryBase extends React.Component {
             footerLink={html.featuredFooterLink}
             header={html.featuredHeader}
             loading={loading}
+            isTheme={isTheme(addonType)}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -290,6 +291,7 @@ export class CategoryBase extends React.Component {
             footerText={html.highlyRatedFooterText}
             header={html.highlyRatedHeader}
             loading={loading}
+            isTheme={isTheme(addonType)}
           />,
         )}
         {this.renderIfNotEmpty(
@@ -302,6 +304,7 @@ export class CategoryBase extends React.Component {
             footerText={html.trendingFooterText}
             header={html.trendingHeader}
             loading={loading}
+            isTheme={isTheme(addonType)}
           />,
         )}
       </div>
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -1,4 +1,5 @@
 /* eslint-disable react/no-unused-prop-types */
+import makeClassName from 'classnames';
 import PropTypes from 'prop-types';
 import config from 'config';
 import * as React from 'react';
@@ -29,6 +30,7 @@ import {
   apiAddonType,
   apiAddonTypeIsValid,
   getAddonTypeFilter,
+  isTheme,
 } from 'core/utils';
 
 import './styles.scss';
@@ -251,7 +253,11 @@ export class CategoryBase extends React.Component {
     const { html } = this.contentForType(addonType);
 
     return (
-      <div className="Category">
+      <div
+        className={makeClassName('Category', {
+          'Category--theme': isTheme(addonType),
+        })}
+      >
         {category && (
           <Helmet>
             <title>{`${category.name} – ${html.title}`}</title>
--- a/src/amo/components/FeaturedCollectionCard/index.js
+++ b/src/amo/components/FeaturedCollectionCard/index.js
@@ -3,6 +3,7 @@ import * as React from 'react';
 import makeClassNames from 'classnames';
 
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
+import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { INSTALL_SOURCE_FEATURED_COLLECTION } from 'core/constants';
 import type { AddonType } from 'core/types/addons';
 
@@ -13,17 +14,13 @@ type Props = {|
   className: string,
   footerText: string,
   header: string,
-  isTheme?: boolean,
+  isTheme: boolean,
   loading: boolean,
   slug: string,
   username: string,
 |};
 
 export default class FeaturedCollectionCard extends React.Component<Props> {
-  static defaultProps = {
-    isTheme: false,
-  };
-
   render() {
     const {
       addons,
@@ -36,6 +33,12 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
       username,
     } = this.props;
 
+    const themeProps = {};
+
+    if (isTheme) {
+      themeProps.placeholderCount = LANDING_PAGE_THEME_COUNT;
+    }
+
     return (
       <LandingAddonsCard
         addonInstallSource={INSTALL_SOURCE_FEATURED_COLLECTION}
@@ -46,8 +49,8 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
         header={header}
         footerText={footerText}
         footerLink={`/collections/${username}/${slug}/`}
-        isTheme={isTheme}
         loading={loading}
+        {...themeProps}
       />
     );
   }
--- a/src/amo/components/FeaturedCollectionCard/index.js
+++ b/src/amo/components/FeaturedCollectionCard/index.js
@@ -3,7 +3,6 @@ import * as React from 'react';
 import makeClassNames from 'classnames';
 
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
-import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { INSTALL_SOURCE_FEATURED_COLLECTION } from 'core/constants';
 import type { AddonType } from 'core/types/addons';
 
@@ -33,12 +32,6 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
       username,
     } = this.props;
 
-    const themeProps = {};
-
-    if (isTheme) {
-      themeProps.placeholderCount = LANDING_PAGE_THEME_COUNT;
-    }
-
     return (
       <LandingAddonsCard
         addonInstallSource={INSTALL_SOURCE_FEATURED_COLLECTION}
@@ -50,7 +43,7 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
         footerText={footerText}
         footerLink={`/collections/${username}/${slug}/`}
         loading={loading}
-        {...themeProps}
+        isTheme={isTheme}
       />
     );
   }
--- a/src/amo/components/FeaturedCollectionCard/index.js
+++ b/src/amo/components/FeaturedCollectionCard/index.js
@@ -1,27 +1,36 @@
 /* @flow */
 import * as React from 'react';
+import makeClassNames from 'classnames';
 
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import { INSTALL_SOURCE_FEATURED_COLLECTION } from 'core/constants';
 import type { AddonType } from 'core/types/addons';
+
+import './styles.scss';
 
 type Props = {|
   addons: Array<AddonType>,
   className: string,
   footerText: string,
   header: string,
+  isTheme?: boolean,
   loading: boolean,
   slug: string,
   username: string,
 |};
 
 export default class FeaturedCollectionCard extends React.Component<Props> {
+  static defaultProps = {
+    isTheme: false,
+  };
+
   render() {
     const {
       addons,
       className,
       footerText,
       header,
+      isTheme,
       loading,
       slug,
       username,
@@ -31,10 +40,13 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
       <LandingAddonsCard
         addonInstallSource={INSTALL_SOURCE_FEATURED_COLLECTION}
         addons={addons}
-        className={className}
+        className={makeClassNames(className, {
+          'FeaturedCollection--theme': isTheme,
+        })}
         header={header}
         footerText={footerText}
         footerLink={`/collections/${username}/${slug}/`}
+        isTheme={isTheme}
         loading={loading}
       />
     );
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -10,6 +10,7 @@ import FeaturedCollectionCard from 'amo/components/FeaturedCollectionCard';
 import HomeHeroBanner from 'amo/components/HomeHeroBanner';
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import Link from 'amo/components/Link';
+import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { fetchHomeAddons } from 'amo/reducers/home';
 import {
   ADDON_TYPE_EXTENSION,
@@ -51,25 +52,26 @@ export const getFeaturedCollectionsMetadata = (i18n) => {
         'See more social media customization extensions',
       ),
       header: i18n.gettext('Social media customization'),
+      isTheme: false,
       ...FEATURED_COLLECTIONS[0],
-      isTheme: true,
     },
     {
       footerText: i18n.gettext('See more dynamic downloaders'),
       header: i18n.gettext('Dynamic downloaders'),
+      isTheme: false,
       ...FEATURED_COLLECTIONS[1],
     },
     {
       footerText: i18n.gettext('See more summer themes'),
       header: i18n.gettext('Summer themes'),
-      ...FEATURED_COLLECTIONS[2],
       isTheme: true,
+      ...FEATURED_COLLECTIONS[2],
     },
     {
       footerText: i18n.gettext('See more must-have media extensions'),
       header: i18n.gettext('Must-have media'),
+      isTheme: false,
       ...FEATURED_COLLECTIONS[3],
-      isTheme: true,
     },
   ];
 };
@@ -258,7 +260,6 @@ export class HomeBase extends React.Component {
             addonInstallSource={INSTALL_SOURCE_FEATURED}
             addons={featuredThemes}
             className="Home-FeaturedThemes"
-            header={i18n.gettext('Featured themes')}
             footerText={i18n.gettext('See more featured themes')}
             footerLink={{
               pathname: '/search/',
@@ -269,8 +270,9 @@ export class HomeBase extends React.Component {
                 featured: true,
               },
             }}
+            header={i18n.gettext('Featured themes')}
             loading={loading}
-            isTheme
+            placeholderCount={LANDING_PAGE_THEME_COUNT}
           />
         )}
 
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -52,6 +52,7 @@ export const getFeaturedCollectionsMetadata = (i18n) => {
       ),
       header: i18n.gettext('Social media customization'),
       ...FEATURED_COLLECTIONS[0],
+      isTheme: true,
     },
     {
       footerText: i18n.gettext('See more dynamic downloaders'),
@@ -62,11 +63,13 @@ export const getFeaturedCollectionsMetadata = (i18n) => {
       footerText: i18n.gettext('See more summer themes'),
       header: i18n.gettext('Summer themes'),
       ...FEATURED_COLLECTIONS[2],
+      isTheme: true,
     },
     {
       footerText: i18n.gettext('See more must-have media extensions'),
       header: i18n.gettext('Must-have media'),
       ...FEATURED_COLLECTIONS[3],
+      isTheme: true,
     },
   ];
 };
@@ -267,6 +270,7 @@ export class HomeBase extends React.Component {
               },
             }}
             loading={loading}
+            isTheme
           />
         )}
 
@@ -311,7 +315,6 @@ export class HomeBase extends React.Component {
             },
           }}
           loading={loading}
-          isTheme
         />
 
         {(loading || collections[3]) && (
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -311,6 +311,7 @@ export class HomeBase extends React.Component {
             },
           }}
           loading={loading}
+          isTheme
         />
 
         {(loading || collections[3]) && (
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -10,7 +10,6 @@ import FeaturedCollectionCard from 'amo/components/FeaturedCollectionCard';
 import HomeHeroBanner from 'amo/components/HomeHeroBanner';
 import LandingAddonsCard from 'amo/components/LandingAddonsCard';
 import Link from 'amo/components/Link';
-import { LANDING_PAGE_THEME_COUNT } from 'amo/constants';
 import { fetchHomeAddons } from 'amo/reducers/home';
 import {
   ADDON_TYPE_EXTENSION,
@@ -271,8 +270,8 @@ export class HomeBase extends React.Component {
               },
             }}
             header={i18n.gettext('Featured themes')}
+            isTheme
             loading={loading}
-            placeholderCount={LANDING_PAGE_THEME_COUNT}
           />
         )}
 
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -42,7 +42,8 @@ export default class LandingAddonsCard extends React.Component<Props> {
     } = this.props;
 
     let footerLinkHtml = null;
-    if (addons && addons.length >= placeholderCount) {
+    const count = isTheme ? LANDING_PAGE_THEME_COUNT : placeholderCount;
+    if (addons && addons.length >= count) {
       let linkTo = footerLink;
       if (linkTo && typeof linkTo === 'object') {
         // As a convenience, fix the query parameter.
@@ -63,7 +64,7 @@ export default class LandingAddonsCard extends React.Component<Props> {
         header={header}
         type="horizontal"
         loading={loading}
-        placeholderCount={isTheme ? LANDING_PAGE_THEME_COUNT : placeholderCount}
+        placeholderCount={count}
       />
     );
   }
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -4,10 +4,7 @@ import * as React from 'react';
 
 import AddonsCard from 'amo/components/AddonsCard';
 import Link from 'amo/components/Link';
-import {
-  LANDING_PAGE_EXTENSION_COUNT,
-  LANDING_PAGE_THEME_COUNT,
-} from 'amo/constants';
+import { LANDING_PAGE_EXTENSION_COUNT } from 'amo/constants';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 import type { AddonType } from 'core/types/addons';
 
@@ -18,13 +15,13 @@ type Props = {|
   footerLink?: Object | string | null,
   footerText?: string,
   header?: React.Node,
-  isTheme?: boolean,
   loading: boolean,
+  placeholderCount: number,
 |};
 
 export default class LandingAddonsCard extends React.Component<Props> {
   static defaultProps = {
-    isTheme: false,
+    placeholderCount: LANDING_PAGE_EXTENSION_COUNT,
   };
 
   render() {
@@ -35,13 +32,9 @@ export default class LandingAddonsCard extends React.Component<Props> {
       footerLink,
       footerText,
       header,
-      isTheme,
       loading,
+      placeholderCount,
     } = this.props;
-
-    const placeholderCount = isTheme
-      ? LANDING_PAGE_THEME_COUNT
-      : LANDING_PAGE_EXTENSION_COUNT;
 
     let footerLinkHtml = null;
     if (addons && addons.length >= placeholderCount) {
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -18,11 +18,15 @@ type Props = {|
   footerLink?: Object | string | null,
   footerText?: string,
   header?: React.Node,
-  isTheme: boolean,
+  isTheme?: boolean,
   loading: boolean,
 |};
 
 export default class LandingAddonsCard extends React.Component<Props> {
+  static defaultProps = {
+    isTheme: false,
+  };
+
   render() {
     const {
       addonInstallSource,
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -34,8 +34,12 @@ export default class LandingAddonsCard extends React.Component<Props> {
       loading,
     } = this.props;
 
-    const filterType = footerLink.query ? footerLink.query.addonType : null;
-    const addonTypeArray = filterType ? filterType.split(',') : [];
+    const filterType =
+      footerLink && footerLink.query && footerLink.query.addonType
+        ? footerLink.query.addonType
+        : null;
+    const addonTypeArray =
+      filterType && typeof filterType === 'string' ? filterType.split(',') : [];
     const placeholderCount = addonTypeArray.find((type) => isTheme(type))
       ? LANDING_PAGE_THEME_ADDON_COUNT
       : LANDING_PAGE_ADDON_COUNT;
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -5,12 +5,11 @@ import * as React from 'react';
 import AddonsCard from 'amo/components/AddonsCard';
 import Link from 'amo/components/Link';
 import {
-  LANDING_PAGE_ADDON_COUNT,
-  LANDING_PAGE_THEME_ADDON_COUNT,
+  LANDING_PAGE_EXTENSION_COUNT,
+  LANDING_PAGE_THEME_COUNT,
 } from 'amo/constants';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 import type { AddonType } from 'core/types/addons';
-import { isTheme } from 'core/utils';
 
 type Props = {|
   addonInstallSource?: string,
@@ -19,6 +18,7 @@ type Props = {|
   footerLink?: Object | string | null,
   footerText?: string,
   header?: React.Node,
+  isTheme: boolean,
   loading: boolean,
 |};
 
@@ -31,18 +31,13 @@ export default class LandingAddonsCard extends React.Component<Props> {
       footerLink,
       footerText,
       header,
+      isTheme,
       loading,
     } = this.props;
 
-    const filterType =
-      footerLink && footerLink.query && footerLink.query.addonType
-        ? footerLink.query.addonType
-        : null;
-    const addonTypeArray =
-      filterType && typeof filterType === 'string' ? filterType.split(',') : [];
-    const placeholderCount = addonTypeArray.find((type) => isTheme(type))
-      ? LANDING_PAGE_THEME_ADDON_COUNT
-      : LANDING_PAGE_ADDON_COUNT;
+    const placeholderCount = isTheme
+      ? LANDING_PAGE_THEME_COUNT
+      : LANDING_PAGE_EXTENSION_COUNT;
 
     let footerLinkHtml = null;
     if (addons && addons.length >= placeholderCount) {
--- a/src/amo/components/LandingAddonsCard/index.js
+++ b/src/amo/components/LandingAddonsCard/index.js
@@ -4,7 +4,10 @@ import * as React from 'react';
 
 import AddonsCard from 'amo/components/AddonsCard';
 import Link from 'amo/components/Link';
-import { LANDING_PAGE_EXTENSION_COUNT } from 'amo/constants';
+import {
+  LANDING_PAGE_EXTENSION_COUNT,
+  LANDING_PAGE_THEME_COUNT,
+} from 'amo/constants';
 import { convertFiltersToQueryParams } from 'core/searchUtils';
 import type { AddonType } from 'core/types/addons';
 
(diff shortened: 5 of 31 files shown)


rebmullin:
update: will finish this up once new images are added so we can implement srcset: https://github.com/mozilla/addons-server/issues/8412#issuecomment-398846907

rebmullin:
@willdurand @bobsilverberg @kumar303  - can anyone please take a look?

note: I am aware that I have some flow issues which I am looking into :)

src/amo/components/SearchResult/index.js
```diff
@@ -55,16 +55,41 @@ export class SearchResultBase extends React.Component<InternalProps> {
...
     let imageURL = iconURL;
 
+    const imageAtts = {};
+
     if (isTheme) {
+      const hasPreviews = addon.previews.length;
+      // Since only newly created static themes will have more than one preview
+      // we will set up a fallback for now.
+      const previewFallback = hasPreviews && addon.previews[0];
+      const previewSearch = hasPreviews && addon.previews[1];
```
rebmullin: It looks like flow is not happy unless I do something like this:

```javascript
const previewFallback = addon && addon.previews && addon.previews[0];
const previewSearch = addon && addon.previews && addon.previews[1];
```

basically, removing `hasPreviews`

I thought the above way above was cleaner and eliminating some redundancy. Am i missing something with flow :/?

src/amo/components/Addon/index.js
```diff
@@ -178,10 +178,9 @@ export class AddonBase extends React.Component {
     const type = addon ? addon.type : ADDON_TYPE_EXTENSION;
 
     if (this.addonIsTheme()) {
-      let previewURL =
-        addon.previews.length > 0 && addon.previews[0].image_url
-          ? addon.previews[0].image_url
-          : null;
+      const previewHeader = addon.previews.length && addon.previews[0];
```
bobsilverberg: It looks like this would set `previewHeader` to `false` if `addon.previews.length` is `0`, which would then cause the two statements below (e.g., `previewHeader.thumbnail_url`) to throw exceptions as `previewHeader` wouldn't be an object.

src/amo/components/Category/index.js
```diff
@@ -251,7 +253,11 @@ export class CategoryBase extends React.Component {
     const { html } = this.contentForType(addonType);
 
     return (
-      <div className="Category">
+      <div
```
bobsilverberg: I've been trying to figure out where this is used in the site to test it, but I cannot find it anywhere. Where is this actually used?

src/amo/components/LandingAddonsCard/index.js
```diff
@@ -33,9 +32,18 @@ export default class LandingAddonsCard extends React.Component<Props> {
       footerText,
       header,
       loading,
-      placeholderCount,
     } = this.props;
 
+    const filterType =
```
bobsilverberg: This logic seems overly complicated. Would it be simpler to just pass a value for `hasThemes` into `LandingAddonsCard` as a prop? Would the caller always know if it's a `LandingAddonsCard` for themes?

src/amo/components/SearchResult/index.js
```diff
@@ -55,15 +55,39 @@ export class SearchResultBase extends React.Component<InternalProps> {
...
     let imageURL = iconURL;
 
+    const imageAtts = {};
+
     if (isTheme) {
+      // Since only newly created static themes will have more than one preview
+      // we will set up a fallback for now.
+      const previewFallback = addon && addon.previews && addon.previews[0];
+      const previewSearch = addon && addon.previews && addon.previews[1];
+      const previewImage = previewSearch || previewFallback;
```
bobsilverberg: Just out of interest, why do we prefer `previews[1]` over `previews[0]`?

src/amo/components/SearchResult/index.js
```diff
@@ -109,7 +133,12 @@ export class SearchResultBase extends React.Component<InternalProps> {
       <div className="SearchResult-result">
         <div className={iconWrapperClassnames}>
           {imageURL ? (
-            <img className="SearchResult-icon" src={imageURL} alt="" />
+            <img
+              className="SearchResult-icon"
+              src={imageURL}
+              alt={addon ? `${addon.name}` : ''}
```
bobsilverberg: This is a change from before. Is this a new requirement from #5174?

src/amo/constants.js
```diff
@@ -42,3 +42,4 @@ export const FEATURED_ADDONS_TO_LOAD = 25;
 // Number of add-ons in the featured, trending, and highest rated landing page
 // sections.
 export const LANDING_PAGE_ADDON_COUNT = 4;
+export const LANDING_PAGE_THEME_ADDON_COUNT = 3;
```
bobsilverberg: Should this just be `LANDING_PAGE_THEME_COUNT`? The `ADDON` part seems redundant. Perhaps the one above should also be changed to `LANDING_PAGE_EXTENSION_COUNT`?

tests/unit/amo/components/TestAddon.js
```diff
@@ -879,7 +879,7 @@ describe(__filename, () => {
     expect(image.type()).toEqual('img');
     expect(image).toHaveClassName('Addon-theme-header-image');
     expect(image.prop('src')).toEqual(
-      'https://addons.cdn.mozilla.net/123/image.png',
+      'https://addons.cdn.mozilla.net/7123/image.png',
```
bobsilverberg: This change was made because we now expect `thumbnail_url` to be used rather than `image_url`, correct? If so, we shouldn't be relying on default values and should be setting a specific value for the `addon` in the test during setup. I've commented on this a bit more at length in a comment about `TestSearchResult.js`.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,50 @@ describe(__filename, () => {
...
+  it('displays the thumbnail image as the default src for static theme', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      type: ADDON_TYPE_STATIC_THEME,
+    });
+    const root = render({ addon });
+    const image = root.find('.SearchResult-icon');
+
+    expect(image.prop('src')).toEqual(
+      'https://addons.cdn.mozilla.net/4444/image.png',
```
bobsilverberg: We generally prefer to set data in our test that we're going to assert about, so rather than checking that this has the value from `fakeAddon`, you should create an add-on in the test, using `fakeAddon` as a basis, but providing your own value for `thumbnail_url`.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,50 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      type: ADDON_TYPE_STATIC_THEME,
+    });
+
+    const root = render({ addon });
+    const image = root.find('.SearchResult-icon');
+
+    expect(image.prop('srcSet')).toEqual(
+      'https://addons.cdn.mozilla.net/4444/image.png 200w, https://addons.cdn.mozilla.net/55555/image.png 400w',
```
bobsilverberg: As above, set specific data to assert on.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,50 @@ describe(__filename, () => {
...
+      type: ADDON_TYPE_STATIC_THEME,
+    });
+
+    delete addon.previews[1];
+
+    const root = render({ addon });
+    const image = root.find('.SearchResult-icon');
+
+    expect(image.prop('src')).toEqual(
+      'https://addons.cdn.mozilla.net/7123/image.png',
```
bobsilverberg: As above, set specific data to assert on.

src/amo/components/SearchResult/index.js
eviljeff: Assuming `previews` is directly from the api, the first item is the header preview, and the second item is the list (search) preview that's longer, and that's ideally the one you want to use.  (Also, it's a list, so if there wasn't a `previews[0]` there wouldn't be a `previews[1]`)

eviljeff: or... I don't really understand what the code does I'm commenting on 🤣 (did I get it right @rebmullin ?)

src/amo/components/SearchResult/index.js
kumar303: It's a limitation of Flow (a bug, too) -- it probably doesn't know that `hasPreviews` points to `addon.previews`. Sometimes you have to write verbose code to satisfy Flow. It's a trade-off.

src/amo/components/SearchResult/index.js
rebmullin: @bobsilverberg, yeh as @eviljeff mentions the first array item is for the header and the 2nd is for search/shelves/grid..

@eviljeff , I guess I was going off this comment [here](https://github.com/mozilla/addons-server/issues/8412#issuecomment-400050380). So I temporarily set up a fallback since I noticed on older static themes, the previews didn't come up.  I think I added a comment for this too. Anyway, I thought it was better to set it up this way..?

src/amo/components/SearchResult/index.js
rebmullin: well not a requirement but on another PR this came up...that maybe it was better to have some alt rather than have it be empty. I can try to dig that comment up if that helps..

src/amo/components/LandingAddonsCard/index.js
rebmullin: thanks for the suggestion @bobsilverberg: ) I will take another look...

src/amo/components/Category/index.js
rebmullin: I am doing this in a few places since I need to override the grid style for the themes. For example, the Category--theme styles look like this:

```css
.Category--theme {
  .Card-contents {
    .AddonsCard-list {
      grid-template-columns: repeat(3, 33%);
    }
  }
}
```

rebmullin: should this be tested?

tests/unit/amo/components/TestAddon.js
rebmullin: ahh...ok i'll take another look at this. thanks @bobsilverberg

tests/unit/amo/components/TestSearchResult.js
rebmullin: I think I got it...I'll update...thanks : )

src/amo/components/SearchResult/index.js
rebmullin: okay good to know... I wasn't sure if there was something I could have done differently with flow. thanks for looking and clarifying : )

src/amo/components/SearchResult/index.js
eviljeff: @rebmullin yeah.  It _shouldn't_ happen on stage and prod, but it's a sensible fallback.

src/amo/components/Addon/index.js
rebmullin: hmm...actually if it comes across as `undefined`, I believe it should be assigned `null`. But I am going to revisit tho anyway to see about cleaning up :/

src/amo/components/SearchResult/index.js
rebmullin: ok cool : ) yeh it was a temporary set up really for dev :|

src/amo/reducers/home.js
```diff
@@ -127,11 +131,12 @@ const reducer = (
         ...state,
         collections: collections.map((collection) => {
           if (collection) {
-            return collection.results
-              .slice(0, LANDING_PAGE_ADDON_COUNT)
-              .map((item) => {
-                return createInternalAddon(item.addon);
-              });
+            const sliceEnd = isTheme(collection.results[0].addon.type)
```
rebmullin: not sure if this is ideal way to figure out but it seems to work so maybe this is okay..?

rebmullin:
hey @bobsilverberg , do you mind taking another look at this when you can? There were some code changes (outside of here) and so I made a few updates to coordinate with those. I also made some updates per your requests : )

rebmullin:
> Perhaps after you make your changes we can meet to discuss that so I can better understand what's required.

definitely! Basically tho, we don't want cropping to happen otherwise it's hard to get an idea of what the theme really is, especially in smaller areas (like mobile, etc). We can follow up with a chat too if you'd like : )

bobsilverberg:
A question just occurred to me while reviewing this: Could we have a featured collection on the home page that contains both extensions and themes? If so, how will we deal with that? Would it use the 3-column format?

willdurand:
> A question just occurred to me while reviewing this: Could we have a featured collection on the home page that contains both extensions and themes? If so, how will we deal with that? Would it use the 3-column format?

that's what we have in -dev currently, no? https://addons-dev.allizom.org/en-US/firefox/

bobsilverberg:
> that's what we have in -dev currently, no? addons-dev.allizom.org/en-US/firefox

Yes, but that's just because I'm lazy and just added some random add-ons to the collections. It's probably not realistic for the collections in question.

rebmullin:
> Could we have a featured collection on the home page that contains both extensions and themes? If so, how will we deal with that? Would it use the 3-column format?

I guess I didn't realize that these could be mixed. I would think 3-column but I guess we can check with @pwalm on this too to be sure.

rebmullin:
> Could we have a featured collection on the home page that contains both extensions and themes? If so, how will we deal with that? Would it use the 3-column format?

per @pwalm this is not a realistic scenario : )

src/amo/components/Category/index.js
rebmullin: I added a test for this..

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,97 @@ describe(__filename, () => {
...
+          image_size: [400, 200],
+          thumbnail_size: [200, 100],
+        },
+        {
+          id: 2,
+          caption: 'Image 2',
+          image_url: 'https://addons.cdn.mozilla.net/thumb/1.png',
+          thumbnail_url: headerImageThumb,
+          image_size: [400, 200],
+          thumbnail_size: [200, 100],
```
rebmullin: I wasn't sure if I need to put in all this data or just the part that I need..?

rebmullin: rereading your comment [here](https://github.com/mozilla/addons-frontend/pull/5225#discussion_r201109747) that's what it seems like but wanted to double check..

src/amo/reducers/home.js
bobsilverberg: Hmm, yeah, this could be an issue if we had mixed content in a collection, but we were just told today that that wouldn't happen, so I guess it's okay to do. There is a problem though, that you could have an empty collection, in which case `collection.results[0]` would be `undefined` and this code would throw an exception, so you need to check for that. I think you could just check for `collection.results.length` here and not do any other processing if it's empty, and let it fall through to the `return null` at line 141.

src/amo/sagas/landing.js
```diff
@@ -24,9 +28,14 @@ export function* fetchLandingAddons({
   try {
     const state = yield select(getState);
     const { api } = state;
+
+    const pageSize = isTheme(addonType)
```
bobsilverberg: You could just inline this below, like:

```js
    const filters = {
      addonType,
      page_size: isTheme(addonType)
        ? LANDING_PAGE_THEME_COUNT
        : LANDING_PAGE_EXTENSION_COUNT,
    };
```

tests/unit/amo/components/TestAddon.js
```diff
@@ -866,34 +866,52 @@ describe(__filename, () => {
   });
 
   it('renders a static theme preview as an image', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/1.png';
     const root = shallowRender({
       addon: createInternalAddon({
         ...fakeTheme,
         type: ADDON_TYPE_STATIC_THEME,
+        previews: [
```
bobsilverberg: It's a bit annoying that you have to specify all of this, but I think it's correct to pass valid values into the component. You could save yourself from having to do it here by creating a new object in `tests/unit/amo/helpers.js` for `fakePreview`, and then using that to populate `fakeAddon` in the same file, and also using `fakePreview` here and only overriding the one property the test cares about.

tests/unit/amo/components/TestAddon.js
```diff
@@ -866,34 +866,52 @@ describe(__filename, () => {
...
     expect(image.prop('alt')).toEqual('Preview of Dancing Daisies by MaDonna');
   });
 
   it('renders the preview image from the previews array if it exists for the lightweight theme', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/1.png';
     const root = shallowRender({
       addon: createInternalAddon({
         ...fakeTheme,
         type: ADDON_TYPE_THEME,
+        previews: [
```
bobsilverberg: See my suggestion about about `fakePreview`.

tests/unit/amo/components/TestSearchResult.js
bobsilverberg: You only need to override the part the test cares about, but the way `fakeAddon` is set up you do need this. I made a suggestion above about creating a `fakePreview` which you could also use here and then you'd only have to override the one property.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,97 @@ describe(__filename, () => {
...
+
+  it('displays srcSet values if preview has multiple options', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/1.png';
+    const headerImageFull = 'https://addons.cdn.mozilla.net/full/1.png';
+    const thumbWidth = 200;
+    const fullWidth = 400;
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      type: ADDON_TYPE_STATIC_THEME,
+      previews: [
```
bobsilverberg: See the suggestion above about `fakePreview`. Also, in this test you care that you get the value from `previews[1]` and not `previews[0]`, so you should override both of those with different values, so you can be sure you're getting the expected one back.

tests/unit/amo/sagas/test_home.js
```diff
@@ -66,7 +69,7 @@ describe(__filename, () => {
 
       const baseArgs = { api: state.api };
       const baseFilters = {
-        page_size: LANDING_PAGE_ADDON_COUNT,
+        page_size: LANDING_PAGE_EXTENSION_COUNT,
```
bobsilverberg: Because we are expecting different results for different API calls, I don't think we should keep `baseFilters`. We should just specify exactly what is expected in each case (like you've done for the themes case).

tests/unit/amo/sagas/test_home.js
```diff
@@ -174,7 +178,7 @@ describe(__filename, () => {
 
         const baseArgs = { api: state.api };
         const baseFilters = {
```
bobsilverberg: Same here. Do away with `baseFilters`.

src/amo/components/Home/index.js
```diff
@@ -62,11 +63,13 @@ export const getFeaturedCollectionsMetadata = (i18n) => {
       footerText: i18n.gettext('See more summer themes'),
       header: i18n.gettext('Summer themes'),
       ...FEATURED_COLLECTIONS[2],
+      isTheme: true,
     },
     {
       footerText: i18n.gettext('See more must-have media extensions'),
       header: i18n.gettext('Must-have media'),
       ...FEATURED_COLLECTIONS[3],
+      isTheme: true,
```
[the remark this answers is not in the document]
rebmullin: hey @bobsilverberg, it looks like right now these are all themes on dev, so just want to verify this?

src/amo/components/LandingAddonsCard/index.js
```diff
@@ -15,13 +18,13 @@ type Props = {|
   footerLink?: Object | string | null,
   footerText?: string,
   header?: React.Node,
+  isTheme?: boolean,
   loading: boolean,
-  placeholderCount: number,
```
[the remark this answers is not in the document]
rebmullin: yeh let's keep it overridable : ) thanks

src/amo/components/Home/index.js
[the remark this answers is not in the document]
bobsilverberg: That was my mistake when populating the collection. They should all be extensions.

tests/unit/amo/components/TestAddon.js
rebmullin: Yeh I like this and will this out, thanks for suggestion : )

rebmullin:
Thanks @bobsilverberg  for tackling this review : ) I know it’s a big one and its a bit time consuming..

So I think I addressed all of your requests.  

One thing to note is swapped out isTheme with placeholderCount in the LandingAddonsCard since it seemed to do similar things. Let me know what you think about that..

Also, I’ve updated added the fakePreview. Hopefully this is better too : )

Also, can I ask a favor :) when you have a min..can you please update the collections so the social media ones on the homepage are correct? I think it’s the first and last that need updating and are only supposed to be showing extensions (right?). I tried to do it but it said I don’t have permissions :/

bobsilverberg:
> when you have a min..can you please update the collections so the social media ones on the homepage are correct?

I've done this.

src/amo/components/Category/index.js
```diff
@@ -250,8 +253,20 @@ export class CategoryBase extends React.Component {
 
     const { html } = this.contentForType(addonType);
 
+    const isAddonTheme = isTheme(addonType);
```
bobsilverberg: This only seems to be used once, below, so it's unnecessary. You could just do: `if (isTheme(addonType))` below instead of `if (isAddonTheme)`.

src/amo/components/Category/index.js
```diff
@@ -250,8 +253,20 @@ export class CategoryBase extends React.Component {
...
+
+    const themeProps = {};
```
bobsilverberg: This use of `themeProps` feels a bit heavyweight to me. Why not just always pass a `placeholderCount` property to `LandingAddonsCard` and set it to the appropriate value, or maybe just pass an `isTheme` property to `LandingAddonsCard` and allow it to determine the correct value for `placeholderCount`?

src/amo/components/FeaturedCollectionCard/index.js
```diff
@@ -22,20 +27,30 @@ export default class FeaturedCollectionCard extends React.Component<Props> {
       className,
       footerText,
       header,
+      isTheme,
       loading,
       slug,
       username,
     } = this.props;
 
+    const themeProps = {};
```
bobsilverberg: Same comment as above. This is making it look like passing an `isTheme` to `LandingAddonsCard` would be the best option, as then you could remove logic from this and from `Category`.

src/amo/components/Home/index.js
```diff
@@ -266,7 +270,9 @@ export class HomeBase extends React.Component {
                 featured: true,
               },
             }}
+            header={i18n.gettext('Featured themes')}
             loading={loading}
+            placeholderCount={LANDING_PAGE_THEME_COUNT}
```
bobsilverberg: As discussed above, if we just pass `isTheme` to `LandingAddonsCard` we could remove this logic too.

src/amo/components/LandingPage/index.js
```diff
@@ -228,9 +230,19 @@ export class LandingPageBase extends React.Component {
         to customize Firefox and make the browser all your own.`),
     };
 
+    const isAddonTheme = isTheme(addonType);
+
+    const themeProps = {};
```
bobsilverberg: You could replace this too!

tests/unit/amo/components/TestAddon.js
```diff
@@ -866,34 +867,47 @@ describe(__filename, () => {
   });
 
   it('renders a static theme preview as an image', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/54321.png';
+    const newPreview = [
```
bobsilverberg: You don't really need this `newPreview` variable - you could just inline this into the call to `shallowRender`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -866,34 +867,47 @@ describe(__filename, () => {
...
-    expect(image.prop('src')).toEqual(
-      'https://addons.cdn.mozilla.net/123/image.png',
-    );
+    expect(image.prop('src')).toEqual(headerImageThumb);
     expect(image.prop('alt')).toEqual('Preview of Dancing Daisies by MaDonna');
   });
 
   it('renders the preview image from the previews array if it exists for the lightweight theme', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/12345.png';
+    const newPreview = [
```
bobsilverberg: You don't really need this `newPreview` variable - you could just inline this into the call to `shallowRender`.

tests/unit/amo/components/TestCategory.js
```diff
@@ -500,6 +500,34 @@ describe(__filename, () => {
     );
   });
 
+  it('renders a theme class name if type is a theme', () => {
+    const root = render(
+      { params: { visibleAddonType: visibleAddonType(ADDON_TYPE_THEME) } },
```
bobsilverberg: Do you need to specify both `params` and `paramOverrides`, that seems odd.

tests/unit/amo/components/TestCategory.js
```diff
@@ -500,6 +500,34 @@ describe(__filename, () => {
...
+        },
+      },
+    );
+
+    expect(root).toHaveClassName('Category--theme');
+  });
+
+  it('renders without a theme class name if type is an extension', () => {
+    const root = render(
+      { params: { visibleAddonType: visibleAddonType(ADDON_TYPE_EXTENSION) } },
```
bobsilverberg: Do you need to specify both `params` and `paramOverrides`, that seems odd.

tests/unit/amo/components/TestFeaturedCollectionCard.js
```diff
@@ -55,6 +55,20 @@ describe(__filename, () => {
     expect(root.find(LandingAddonsCard)).toHaveProp('className', className);
   });
 
+  it('renders a theme class name to LandingAddonsCard if the collection contains theme addons', () => {
```
bobsilverberg: Because the logic isn't actually tied to "if the collection contains theme addons", but rather to the value of the `isTheme` property, the test should be worded that way as well.

tests/unit/amo/components/TestFeaturedCollectionCard.js
```diff
@@ -55,6 +55,20 @@ describe(__filename, () => {
...
+    const root = render({ isTheme: true });
+    expect(root.find(LandingAddonsCard)).toHaveClassName(
+      'FeaturedCollection--theme',
+    );
+  });
+
+  it('does not render a theme class name to LandingAddonsCard if the collection contains extension addons', () => {
```
bobsilverberg: Because the logic isn't actually tied to "if the collection contains theme addons", but rather to the value of the isTheme property, the test should be worded that way as well.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,85 @@ describe(__filename, () => {
     expect(root).toHaveClassName('SearchResult--theme');
   });
 
+  it('displays the thumbnail image as the default src for static theme', () => {
+    const headerImageThumb =
+      'https://addons.cdn.mozilla.net/mytestthumb/12345.png';
+    const newPreview = [
```
bobsilverberg: You don't really need this `newPreview` variable, nor the `addon` variable below - you could just inline these into the call to `render`.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,85 @@ describe(__filename, () => {
...
+    expect(image.prop('src')).toEqual(headerImageThumb);
+  });
+
+  it('displays srcSet values if preview has multiple options', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/12345.png';
+    const headerImageFull = 'https://addons.cdn.mozilla.net/full/54321.png';
+    const thumbWidth = 450;
+    const fullWidth = 900;
+
+    const newPreview = [
```
bobsilverberg: You don't really need this `newPreview` variable, nor the `addon` variable below - you could just inline these into the call to `render`.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -144,6 +144,85 @@ describe(__filename, () => {
...
+    expect(image.prop('srcSet')).toEqual(
+      `${headerImageThumb} ${thumbWidth}w, ${headerImageFull} ${fullWidth}w`,
+    );
+  });
+
+  // TODO: This can be removed once migration happens.
+  it('displays a fallback image for themes that only have 1 preview option', () => {
+    const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/1.png';
+
+    const newPreview = [
```
bobsilverberg: You don't really need this `newPreview` variable, nor the `addon` variable below - you could just inline these into the call to `render`.

src/amo/components/Category/index.js
rebmullin: I am using this 2x - on line 260 and line 267. is that ok then?

src/amo/components/Category/index.js
rebmullin: ok sure, I like the placeholder way : )

src/amo/components/FeaturedCollectionCard/index.js
rebmullin: I think I like the placeholder way that you suggested above.

src/amo/components/Home/index.js
rebmullin: i guess imo the placeholder and isTheme seemed to be doing the same thing so why not just pass in the placeholder prop..?

tests/unit/amo/components/TestCategory.js
rebmullin: oh! let me double check

tests/unit/amo/components/TestCategory.js
rebmullin: oh! will double check : /

tests/unit/amo/components/TestFeaturedCollectionCard.js
rebmullin: okay..

tests/unit/amo/components/TestFeaturedCollectionCard.js
rebmullin: I am still not sure about using isTheme prop :/ but I will try it out

tests/unit/amo/components/TestSearchResult.js
rebmullin: yep i'll do that :)

src/amo/components/Home/index.js
rebmullin: otherwise there would be no way to override the placeholder value for a theme, right?

src/amo/components/Category/index.js
rebmullin: otherwise, you wouldn't be able to override the placeholder count for a theme, right?

rebmullin: i guess you could if you don't pass in isTheme and just use the placeholder count...hmm. I'll try it out : )

src/amo/components/Category/index.js
rebmullin: now I using it many times :)

rebmullin:
> I've done this.

perfect, thanks!

src/amo/components/Category/index.js
bobsilverberg: So you are! 👍

src/amo/components/Home/index.js
bobsilverberg: It looks like you went the `isTheme` route, rather than the `placeholderCount` route, so I assume your comments above are no longer relevant? You could always pass in `placeholderCount` instead if you really needed to override the number of placeholders, right?

tests/unit/amo/components/TestFeaturedCollectionCard.js
```diff
@@ -55,6 +55,20 @@ describe(__filename, () => {
     expect(root.find(LandingAddonsCard)).toHaveProp('className', className);
   });
 
+  it("renders a theme class name to LandingAddonsCard if the collection's isTheme property is true", () => {
```
bobsilverberg: This is still a little inaccurate. You are passing a value for `isTheme` directly into the component, so it's not really accurate to say "if the collection's isTheme property is true" because the component doesn't know where this `isTheme` value comes from. I think a better description might be "renders a theme class name to LandingAddonsCard when isTheme is true".

tests/unit/amo/components/TestFeaturedCollectionCard.js
```diff
@@ -55,6 +55,20 @@ describe(__filename, () => {
...
+    const root = render({ isTheme: true });
+    expect(root.find(LandingAddonsCard)).toHaveClassName(
+      'FeaturedCollection--theme',
+    );
+  });
+
+  it("does not render a theme class name to LandingAddonsCard if the collection's isTheme property is false", () => {
```
bobsilverberg: See above for my comment about the descriptions of these tests. I think you should remove the part about it being a collection's isTheme property.

tests/unit/amo/components/TestFeaturedCollectionCard.js
bobsilverberg: Are you still unsure about this? It seems to work pretty well to me. What issues do you have with it?

tests/unit/amo/components/TestFeaturedCollectionCard.js
rebmullin: yikes..I will get this : ) Thanks for suggestion : ) I will update..

src/amo/components/Home/index.js
rebmullin: yep, I did stick with the isTheme route b/c I saw as you say here, I could override this. So I think it's good now : ) thanks

rebmullin:
> You seemed unsure about the isTheme property approach. Are you still concerned about that?

I think after reviewing your suggestion and going through with the implementation I see that it can be overridden (which was my concern). So as long as you're okay with the way I've implemented this, I think it's good now : )

src/amo/reducers/home.js
```diff
@@ -126,12 +130,13 @@ const reducer = (
       return {
         ...state,
         collections: collections.map((collection) => {
-          if (collection) {
-            return collection.results
-              .slice(0, LANDING_PAGE_ADDON_COUNT)
-              .map((item) => {
-                return createInternalAddon(item.addon);
-              });
+          if (collection && collection.results && collection.results.length) {
```
bobsilverberg: We need a test for the case where a collection exists, but it is empty, in which case we expect a `null` to be returned.

src/amo/sagas/landing.js
```diff
@@ -26,7 +30,9 @@ export function* fetchLandingAddons({
     const { api } = state;
     const filters = {
       addonType,
-      page_size: LANDING_PAGE_ADDON_COUNT,
+      page_size: isTheme(addonType)
+        ? LANDING_PAGE_THEME_COUNT
```
bobsilverberg: There doesn't seem to be a test for the case where `isTheme()` is `true`.

src/amo/components/Addon/index.js
```diff
@@ -164,11 +171,24 @@ export class AddonBase extends React.Component {
...
+        const imageWidth = previewHeader && previewHeader.image_size[0];
+        const thumbWidth = previewHeader && previewHeader.thumbnail_size[0];
+        if (imageWidth && thumbWidth) {
+          // If viewing on retina, it should only show the larger size with
+          // the current widths available
+          imageAtts.srcSet = `${previewURL} ${thumbWidth}w, ${previewUrlLarge} ${imageWidth}w`;
+        }
+      }
+
+      const imageClassName = 'Addon-theme-header-image';
```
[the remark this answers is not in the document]
rebmullin: hmm.. I am not sure. I am guessing I thought it was going to be used more than once..? I'll revert this

tests/unit/amo/components/TestAddon.js
```diff
@@ -866,34 +867,44 @@ describe(__filename, () => {
...
+          {
+            ...fakePreview,
+            thumbnail_url: headerImageThumb,
+          },
+        ],
       }),
     });
     const image = root.find('.Addon-theme-header-image');
     expect(image.type()).toEqual('img');
     expect(image).toHaveClassName('Addon-theme-header-image');
```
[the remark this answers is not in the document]
rebmullin: omg, yes, you're right...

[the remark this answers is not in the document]
rebmullin: sorry! I'll update

tests/unit/amo/components/TestCategory.js
```diff
@@ -500,6 +500,30 @@ describe(__filename, () => {
     );
   });
 
+  it('renders a theme class name if type is a theme', () => {
+    const root = render(
+      { params: { visibleAddonType: visibleAddonType(ADDON_TYPE_THEME) } },
+      {
+        autoDispatchCategories: false,
```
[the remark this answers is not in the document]
rebmullin: I'll double check..

src/amo/components/Addon/index.js
```diff
@@ -164,11 +171,24 @@ export class AddonBase extends React.Component {
         previewURL = addon.previewURL;
       }
 
+      const imageAtts = {};
+      if (previewUrlLarge) {
```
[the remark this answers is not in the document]
rebmullin: oh! checking...

src/amo/components/Category/index.js
```diff
@@ -271,6 +279,7 @@ export class CategoryBase extends React.Component {
             footerText={html.featuredFooterText}
             footerLink={html.featuredFooterLink}
             header={html.featuredHeader}
+            isTheme={isAddonTheme}
```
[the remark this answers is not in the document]
rebmullin: 👀

src/amo/reducers/home.js
rebmullin: just checking...this won't suffice? https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/reducers/test_home.js#L60

rebmullin:
@bobsilverberg, thanks again for all the review. Can you please take another look? I think it's close 🤞  I just have a question above...

https://github.com/mozilla/addons-frontend/pull/5225#discussion_r203204344

rebmullin:
@bobsilverberg, hold up a few mins? I noticed an issue that I want to fix first

rebmullin:
@bobsilverberg, okay bug should be fixed now :) Please review when you can, thanks

tests/unit/amo/components/TestCategory.js
```diff
@@ -500,6 +500,77 @@ describe(__filename, () => {
...
+      },
+    );
+
+    expect(root).toHaveClassName('Category--theme');
+  });
+
+  it('renders without a theme class name if type is an extension', () => {
+    const root = render(
+      {
+        params: { visibleAddonType: visibleAddonType(ADDON_TYPE_EXTENSION) },
```
bobsilverberg: This is not consistent with the tests above, where you use `paramOverrides` instead of `params`.

src/amo/components/Category/index.js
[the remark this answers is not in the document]
bobsilverberg: What's been added here is code that passes the value of `isAddonTheme` into the `isTheme` prop of the `LandingAddonsCard` for each of the three instances, which have the classNames `FeaturedAddons`, `HighlyRatedAddons` and `TrendingAddons`, so this could be tested in a single test which could either check that each of those specific elements was passed the correct prop value, or could just check that all instances of `LandingAddonsCard` was passed the correct prop value. You don't need two different tests for extensions and themes, just one test that makes sure that `isTheme` on the instances of `LandingAddonsCard` is what we expect it to be.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,197 @@ describe(__filename, () => {
...
 
+  it('does not render an image if the isAllowedOrigin is false', () => {
+    const root = render({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_STATIC_THEME,
+        previews: [
+          {
+            ...fakePreview,
+            thumbnail_url: 'http://example.com',
```
bobsilverberg: How do we know this is not an allowed origin? We shouldn't write a test where we assume the setup will work. We need to be able to inject a value for `isAllowedOrigin` into the component in order to test this properly, and then use a `sinon.stub` in the test.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,197 @@ describe(__filename, () => {
...
+  });
+
+  it("renders an image's alt attribute as its addon name", () => {
+    const alt = 'pretty image';
+    const root = render({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        name: alt,
+        type: ADDON_TYPE_THEME,
+        previews: [
```
bobsilverberg: Because we're only testing the `img`'s `alt` property, we shouldn't need to override the `previews` property at all.

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,197 @@ describe(__filename, () => {
...
+          },
+        ],
+      }),
+    });
+    const image = root.find('.SearchResult-icon');
+    expect(image.prop('src')).toEqual(headerImageThumb);
+    expect(image).not.toHaveProp('srcSet');
+  });
+
+  it('renders image without srcSet if there is no large preview image url', () => {
```
bobsilverberg: There are 3 different things that are checked which could result in no `srcSet` if any are false: no image URL, no `imageSize` or no `thumbSize`, so I think you could combine the two tests you have here into 1, which tests each of those cases. You could use `it.each()` for this. See https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestHome.js#L67 for an example.

src/amo/reducers/home.js
bobsilverberg: No, in that case `collection` is `null`, but we need a test for where a collection exists, but is empty.

tests/unit/amo/sagas/testLanding.js
```diff
@@ -156,7 +242,7 @@ describe(__filename, () => {
       const baseFilters = {
         addonType,
         category,
-        page_size: LANDING_PAGE_ADDON_COUNT,
+        page_size: LANDING_PAGE_EXTENSION_COUNT,
```
bobsilverberg: Should this be `LANDING_PAGE_THEME_COUNT`?

tests/unit/amo/sagas/testLanding.js
```diff
@@ -226,7 +312,7 @@ describe(__filename, () => {
       const baseArgs = { api: apiState };
       const baseFilters = {
         addonType,
-        page_size: LANDING_PAGE_ADDON_COUNT,
+        page_size: LANDING_PAGE_EXTENSION_COUNT,
```
bobsilverberg: Should this be `LANDING_PAGE_THEME_COUNT`?

tests/unit/amo/sagas/testLanding.js
```diff
@@ -54,17 +58,97 @@ describe(__filename, () => {
       );
     }
 
-    it('fetches landing page addons from the API', async () => {
+    it('fetches extensions landing page addons from the API', async () => {
```
bobsilverberg: I'm not entirely sure, but I wonder if you could use `it.each()` to avoid writing two tests that are essentially the same, but just have a different input and a different expectation?

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -106,6 +107,77 @@ describe(__filename, () => {
     expect(root.find(AddonsCard)).toHaveProp('footerLink', null);
   });
 
+  it('hides the footer link when there are less add-ons than LANDING_PAGE_THEME_COUNT', () => {
+    const addons = [
+      createInternalAddon({
```
bobsilverberg: This test is assuming that `1` (the number of add-ons you're creating) is less than `LANDING_PAGE_THEME_COUNT`, but the test doesn't know that to be the case. I think you need to use `LANDING_PAGE_THEME_COUNT`-1 in order to figure out how many add-ons to create.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -106,6 +107,77 @@ describe(__filename, () => {
...
+        type: ADDON_TYPE_THEME,
+        slug: 'custom-addon',
+      }),
+    ];
+    const root = render({ addons, isTheme: true });
+    expect(root.find(AddonsCard)).toHaveProp('footerLink', null);
+  });
+
+  it('shows the footer link when there are more or as many add-ons than LANDING_PAGE_THEME_COUNT', () => {
+    const addons = [
```
bobsilverberg: As above, we shouldn't assume the test knows how many add-ons to create, but should base that number on the actual value of `LANDING_PAGE_THEME_COUNT`.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -106,6 +107,77 @@ describe(__filename, () => {
...
+      createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+        slug: 'custom-addon-2',
+      }),
+    ];
+    const root = render({ addons, isTheme: true });
+    expect(root.find(AddonsCard)).not.toHaveProp('footerLink', null);
+
+    expect(root.find(AddonsCard).prop('footerLink')).toMatchObject({
```
bobsilverberg: I don't think you need this complicated check. You just need to check that the `footerLink` prop is a `Link` component.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -106,6 +107,77 @@ describe(__filename, () => {
...
+        slug: 'custom-addon-1',
+      }),
+      createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+        slug: 'custom-addon-2',
+      }),
+    ];
+    const root = render({ addons, isTheme: true });
+    expect(root.find(AddonsCard)).not.toHaveProp('footerLink', null);
```
bobsilverberg: Because you're checking the contents of the `footerLink` prop, you don't need this assert.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -106,6 +107,77 @@ describe(__filename, () => {
...
+      props: {
+        to: { pathname: '/some-path/', query: {} },
+        children: 'some text',
+      },
+      _owner: null,
+      _store: {},
+    });
+  });
+
+  it('shows the footer link when there are as many or more add-ons than placeholderCount', () => {
```
bobsilverberg: The comments I made about `it.(shows the footer link when there are more or as many add-ons than LANDING_PAGE_THEME_COUNT)` apply to this one as well.

bobsilverberg:
@rebmullin For some reason I ended up submitting 3 reviews, plus I replied to a number of your comments, so please make sure you look for all of my comments/replies when updating this PR.

tests/unit/amo/sagas/testLanding.js
rebmullin: yes..:/

tests/unit/amo/components/TestSearchResult.js
rebmullin: hmm.. ok let me look into this more

src/amo/components/SearchResult/index.js
```diff
@@ -35,6 +36,7 @@ type InternalProps = {|
 
 export class SearchResultBase extends React.Component<InternalProps> {
   static defaultProps = {
+    isAllowedOrigin: defaultIsAllowedOrigin,
```
rebmullin: so I set this up this way for testing (this was how I interpreted part of your comment [here](https://github.com/mozilla/addons-frontend/pull/5225#discussion_r203726631)

tests/unit/amo/components/TestSearchResult.js
rebmullin: okay, I've updated this...let me know what you think : )

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,170 @@ describe(__filename, () => {
...
+      }),
     });
-    const root = render({ addon });
 
     expect(root).toHaveClassName('SearchResult--theme');
   });
 
+  it('does not render an image if the isAllowedOrigin is false', () => {
+    const differentOriginUrl = 'http://example.com';
+    const isAllowedOriginSpy = sinon.spy();
```
bobsilverberg: This is the right idea, but we don't actually need to verify that the `isAllowedOriginSpy` was called, rather, we need to make sure that the `isAllowedOriginSpy` returns `false`, so you could use a `sinon.stub` and make sure that it returns `false`, like so:

`const isAllowedOriginSpy = sinon.stub().returns(false);`

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,170 @@ describe(__filename, () => {
...
+      }),
+    });
+    const image = root.find('.SearchResult-icon');
+
+    expect(image.prop('srcSet')).toEqual(
+      `${headerImageThumb} ${thumbWidth}w, ${headerImageFull} ${fullWidth}w`,
+    );
+  });
+
+  it.each([0, 1])(
```
bobsilverberg: We actually need to test 3 different things, not just 2. It's a bit more complex, but something like this should work:


```diff
diff --git a/tests/unit/amo/components/TestSearchResult.js b/tests/unit/amo/components/TestSearchResult.js
index 9e00954..6dee78c 100644
--- a/tests/unit/amo/components/TestSearchResult.js
+++ b/tests/unit/amo/components/TestSearchResult.js
@@ -240,9 +240,9 @@ describe(__filename, () => {
     );
   });

-  it.each([0, 1])(
-    `renders image without srcSet if there are no preview image sizes or large image %s`,
-    (index) => {
+  it.each([{ image_url: '' }, { image_size: [] }, { thumbnail_size: [] }])(
+    `renders image without srcSet if there are no preview image sizes or large image %o`,
+    (propOverride) => {
       const headerImageThumb = 'https://addons.cdn.mozilla.net/thumb/12345.png';
       const headerImageFull = 'https://addons.cdn.mozilla.net/full/54321.png';

@@ -258,10 +258,11 @@ describe(__filename, () => {
             },
             {
               ...fakePreview,
-              image_url: index === 0 ? headerImageFull : '',
+              image_url: headerImageFull,
               thumbnail_url: headerImageThumb,
-              image_size: index === 1 ? [400, 200] : [],
-              thumbnail_size: index === 1 ? [200, 100] : [],
+              image_size: [400, 200],
+              thumbnail_size: [200, 100],
+              ...propOverride,
             },
           ],
         }),
```

tests/unit/amo/reducers/test_home.js
```diff
@@ -74,6 +118,21 @@ describe(__filename, () => {
       expect(homeState.collections[0]).toEqual(null);
     });
 
+    it('returns null for an empty collection', () => {
+      const { store } = dispatchClientMetadata();
+
+      store.dispatch(
+        loadHomeAddons({
+          collections: [{}],
```
bobsilverberg: This should be a valid CollectionAddonsListResponse with an empty set of add-ons, e.g.,

```js
          collections: [
            createFakeCollectionAddonsListResponse({
              addons: [],
            }),
          ],
```

tests/unit/amo/sagas/testLanding.js
```diff
@@ -54,84 +58,90 @@ describe(__filename, () => {
...
     }
 
-    it('fetches landing page addons from the API', async () => {
-      const addonType = ADDON_TYPE_EXTENSION;
-      const baseArgs = { api: apiState };
-      const baseFilters = {
-        addonType,
-        page_size: LANDING_PAGE_ADDON_COUNT,
-      };
+    it.each([0, 1])(
```
bobsilverberg: Using a numbered index and translation logic in the test is more difficult to follow and maintain than using actual values. Here's an example that should work:

```diff
diff --git a/tests/unit/amo/sagas/testLanding.js b/tests/unit/amo/sagas/testLanding.js
index d477187..9d3bfb8 100644
--- a/tests/unit/amo/sagas/testLanding.js
+++ b/tests/unit/amo/sagas/testLanding.js
@@ -58,17 +58,23 @@ describe(__filename, () => {
       );
     }

-    it.each([0, 1])(
-      `fetches landing page addons from the API position %s`,
-      async (index) => {
-        const addonType = index === 0 ? ADDON_TYPE_EXTENSION : ADDON_TYPE_THEME;
+    it.each([
+      {
+        addonType: ADDON_TYPE_EXTENSION,
+        pageSize: LANDING_PAGE_EXTENSION_COUNT,
+      },
+      {
+        addonType: ADDON_TYPE_THEME,
+        pageSize: LANDING_PAGE_THEME_COUNT,
+      },
+    ])(
+      `fetches landing page addons from the API for %o`,
+      async (testConfig) => {
+        const { addonType, pageSize } = testConfig;
         const baseArgs = { api: apiState };
         const baseFilters = {
           addonType,
-          page_size:
-            index === 0
-              ? LANDING_PAGE_EXTENSION_COUNT
-              : LANDING_PAGE_THEME_COUNT,
+          page_size: pageSize,
         };

         const featured = createAddonsApiResult([
```

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -75,6 +108,77 @@ describe(__filename, () => {
     expect(root.find(AddonsCard)).toHaveProp('footerLink', null);
   });
 
+  it('hides the footer link when there are less add-ons than LANDING_PAGE_THEME_COUNT', () => {
+    const addons = [];
+
+    for (let i = 0; i < LANDING_PAGE_THEME_COUNT - 1; i++) {
```
bobsilverberg: You could probably use:

`const addons = Array(LANDING_PAGE_THEME_COUNT - 1).fill(createInternalAddon(fakeAddon));`

like is done in the `render` function at the top of the file, rather than manually pushing items onto an array.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -75,6 +108,77 @@ describe(__filename, () => {
...
+    const root = render({ addons, isTheme: true });
+    expect(root.find(AddonsCard)).toHaveProp('footerLink', null);
+  });
+
+  it('shows the footer link when there are more or as many add-ons as LANDING_PAGE_THEME_COUNT', () => {
+    const addons = [];
+    const footerLink = 'footer-link-path';
+    const footerText = 'footer link text';
+
+    for (let i = 0; i < LANDING_PAGE_THEME_COUNT; i++) {
```
bobsilverberg: As above, `Array().fill` is shorter.

tests/unit/amo/components/TestLandingAddonsCard.js
```diff
@@ -75,6 +108,77 @@ describe(__filename, () => {
...
+    );
+  });
+
+  it('shows the footer link when there are more or as many add-ons as placeholderCount', () => {
+    const addons = [];
+    const footerLink = 'footer-link-path';
+    const footerText = 'footer link text';
+    const placeholderCount = 2;
+
+    for (let i = 0; i < placeholderCount; i++) {
```
bobsilverberg: As above, `Array().fill` is shorter.

tests/unit/amo/components/TestSearchResult.js
rebmullin: yeh that's so much better 👍

tests/unit/amo/sagas/testLanding.js
rebmullin: 👍

tests/unit/amo/components/TestSearchResult.js
rebmullin: I *think* I got it. let me know if the update is okay : )

rebmullin:
@bobsilverberg, fingers crossed it's ready but let me know :)

src/amo/components/SearchResult/index.js
willdurand: It is a good pattern! I think we tend to name the prop with a leading underscore (`_isAllowedOrigin`) so that there is no need to alias the import + we know that this prop is injected with a default value/implementation in the code (because of the underscore).

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -131,19 +131,93 @@ describe(__filename, () => {
...
+          },
+        ],
+      }),
+    });
+    const image = root.find('.SearchResult-icon');
+
+    expect(image.prop('src')).toEqual(headerImageFull);
+  });
+
+  // TODO: This can be removed once migration happens.
```
willdurand: Can we link to a GH issue?

PR #5236, opened 2018-06-07
I3601 error handling for unknown type in rating manager

### LINKED ISSUE #3601: Static theme details page return a 500 error page if logged out
Steps to reproduce:
1. Submit a static theme i.e. https://addons-dev.allizom.org/en-US/firefox/addon/static-theme-ex2/
2. Being logged out load its details page using the new frontend pages 

Expected results:
Theme details page is loaded.

Actual results:
A 500 error page is loaded.

Notes/Issues:
- being logged in does not reproduce the issue, even if you are not the author of the theme
- using old desktop pages doesn't reproduce the issue, also

Verified on AMO-dev Ff56(Win 7).
Video for this issue:
static theme

### LINKED ISSUE #5102: Rating Manager component text should be updated for static theme type

**(once static theme is added as a valid addon type - see PR #5067)
make sure you're logged out
find a static theme addon: eg: https://addons-dev.allizom.org/en-US/firefox/addon/rm-static-theme-3/?src=search
Alt text

expect if on env where static themes are enabled text to be correct (match lwt/persona type text) 

see somewhat related issue #3601 - also error handling error should be looked at too as noted in the issue

rebmullin:
fixes #3601 

PR #5237 should help with this but here i am addressing the error handling:

basically, i just took this out so it's a little bit more consistent (you can access this page when you're already logged in) - let me know what you think

### DIFF
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -17,7 +17,6 @@ import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_LANG,
   ADDON_TYPE_OPENSEARCH,
-  ADDON_TYPE_STATIC_THEME,
   ADDON_TYPE_THEME,
   validAddonTypes as defaultValidAddonTypes,
 } from 'core/constants';
@@ -142,7 +141,6 @@ export class RatingManagerBase extends React.Component<Props, State> {
         return i18n.gettext('Log in to rate this language pack');
       case ADDON_TYPE_OPENSEARCH:
         return i18n.gettext('Log in to rate this search plugin');
-      case ADDON_TYPE_STATIC_THEME:
       case ADDON_TYPE_THEME:
         return i18n.gettext('Log in to rate this theme');
       case ADDON_TYPE_EXTENSION:
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -150,7 +150,7 @@ export class RatingManagerBase extends React.Component<Props, State> {
       default: {
         const validAddonType = validAddonTypes.includes(addonType);
         log.warn(`Using generic prompt for ${validAddonType ?
-          'add-on' : 'unknown extension'} type: ${addonType}`
+          'add-on' : 'unknown add-on'} type: ${addonType}`
         );
         return i18n.gettext('Log in to rate this add-on');
       }
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -150,8 +150,7 @@ export class RatingManagerBase extends React.Component<Props, State> {
       default: {
         const validAddonType = validAddonTypes.includes(addonType);
         log.warn(`Using generic prompt for ${validAddonType ?
-          'add-on' : 'unknown add-on'} type: ${addonType}`
-        );
+          'add-on' : 'unknown add-on'} type: ${addonType}`);
         return i18n.gettext('Log in to rate this add-on');
       }
     }
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -17,6 +17,7 @@ import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_LANG,
   ADDON_TYPE_OPENSEARCH,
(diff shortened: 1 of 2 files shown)


willdurand:
> rebmullin (Rebecca Mullin) added this to the 2018.06.21 milestone 31 minutes ago

@rebmullin I think we don't set milestones on PRs.

rebmullin:
lol @willdurand...oops, I've done that a couple of time :/  Meant to do on the issues..

rebmullin:
note: I am also going to revert changes for #5102 since there is another PR for this..

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -333,8 +333,8 @@ describe('RatingManager', () => {
     });
 
     it('cannot render a login prompt for unknown extension types', () => {
```
willdurand: can or cannot?

rebmullin: 🤦‍♂️

rebmullin: yes "can" ... i'll update

willdurand: Cool, I think we should link to the original issue too.

src/amo/components/RatingManager/index.js
```diff
@@ -147,12 +147,13 @@ export class RatingManagerBase extends React.Component<Props, State> {
...
         return i18n.gettext('Log in to rate this extension');
-      default:
-        if (!validAddonTypes.includes(addonType)) {
-          throw new Error(`Unknown extension type: ${addonType}`);
-        }
-        log.warn(`Using generic prompt for add-on type: ${addonType}`);
+      default: {
+        const validAddonType = validAddonTypes.includes(addonType);
+        log.warn(`Using generic prompt for ${validAddonType ?
+          'add-on' : 'unknown extension'} type: ${addonType}`
```
willdurand: so if the addon type is invalid, it will print `using generic prompt for unknown extension`, but is it really an extension? Why do you want to update the log message here?

rebmullin: well that's what the message said before basically - in that it referred to it as an extension if it was not valid : `Unknown extension type: ${addonType}`. Since we are using a prompt now (instead of throwing the error), I really just added this to the beginning of the original text

willdurand: mmh but this does not look correct to me, because add-on is the most generic term if I am not mistaken (i.e. a web extension is an add-on).

rebmullin: I see what you're saying...so what about what this:
```javascript
   log.warn(`Using generic prompt for ${validAddonType ?
     'add-on' : 'unknown addon-on'} type: ${addonType}`
```

willdurand: :+1: perfect!

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -338,9 +338,12 @@ describe('RatingManager', () => {
       expect(prompt).toContain('theme');
     });
 
-    it('cannot render a login prompt for unknown extension types', () => {
-      expect(() => getAuthPromptForType('xul'))
-        .toThrowError(/Unknown extension type: xul/);
+    // Since you can view this page if you're logged in -
+    // to be consistent - we'll do the same if you're logged out.
+    // See https://github.com/mozilla/addons-frontend/issues/3601.
```
willdurand: I like to write "See:" (with the colon) 😉

src/amo/components/RatingManager/index.js
```diff
@@ -147,12 +147,13 @@ export class RatingManagerBase extends React.Component<Props, State> {
...
-      default:
-        if (!validAddonTypes.includes(addonType)) {
-          throw new Error(`Unknown extension type: ${addonType}`);
-        }
-        log.warn(`Using generic prompt for add-on type: ${addonType}`);
+      default: {
+        const validAddonType = validAddonTypes.includes(addonType);
+        log.warn(`Using generic prompt for ${validAddonType ?
+          'add-on' : 'unknown add-on'} type: ${addonType}`
+        );
```
willdurand: Can we put this parenthesis on the line above it?

PR #5237, opened 2018-06-07
updating Rating Manager component text for static theme

### LINKED ISSUE #5102: Rating Manager component text should be updated for static theme type

championshuttler:
Fixes #5102 
cc - @rebmullin 

screenshot from 2018-06-07 21-41-43

### DIFF
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -142,9 +142,8 @@ export class RatingManagerBase extends React.Component<Props, State> {
         return i18n.gettext('Log in to rate this language pack');
       case ADDON_TYPE_OPENSEARCH:
         return i18n.gettext('Log in to rate this search plugin');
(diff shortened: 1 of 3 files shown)


championshuttler:
@rebmullin Haha I think I am late by few seconds only

rebmullin:
ha @championshuttler yep sure thing : ) I'll revert my changes..

championshuttler:
@rebmullin, thanks for being so kind :dancing_men:

Issue #5230, opened 2018-06-07
Report this user for abuse -  Mozilla's add-on policies link is not available


ioanarusiczki:
STR:
1. Load AMO dev and go to an add-on or a theme's detail page
2. Click on the developer's name
3. Click Report this user for abuse
4. Observe the Mozilla's Add-on Policies

Expected results:
"Mozilla's Add-on Policies" is a link to https://developer.mozilla.org/de/Add-ons/AMO/Policy/Reviews

Actual results:
Link is not available yet.

Notes:
This issue is reproducible on AMO dev - new View Profile page - with FF60(Win10).

![report user and the policies](https://user-images.githubusercontent.com/33448286/41088940-65bd0b0c-6a49-11e8-8eab-57c22bef3441.gif)

willdurand:
Interesting, I think the link never appeared.

ioanarusiczki:
It appears now. Also appears in Report a user for abuse too. Verified fixed on AMO dev with FF60(Win10)
![report for abuse](https://user-images.githubusercontent.com/33448286/41159013-b7285b00-6b33-11e8-92dc-e694cf42fc5f.gif)

PR #5240, opened 2018-06-07
Edit comments about an addon in a collection

### LINKED ISSUE #4578: Edit comments about an add-on in a collection

bobsilverberg:
Fixes #4578 

~~Depends on #5170~~


bobsilverberg:
Here's the edit screen with a couple of add-ons that already have a comment:

screenshot 2018-06-07 17 02 45

And editing the first comment:

screenshot 2018-06-07 17 04 08

And here's a screen where the first add-on didn't have a comment, so I clicked on the comment icon to open the form to add a new comment:

screenshot 2018-06-07 17 05 23

bobsilverberg:
This has been rebased on top of PR #5170

src/amo/components/EditableCollectionAddon/index.js
```diff
@@ -56,11 +111,59 @@ export class EditableCollectionAddonBase extends React.Component<Props> {
...
+          </div>
+        )}
       </li>
     );
   }
 }
 
+export const extractId = (ownProps: Props) => {
+  const { addon } = ownProps;
+  return `editable-collection-addon-${addon ? addon.id : ''}`;
```
willdurand: From what I am seeing, the `addon` is always present.

src/amo/reducers/collections.js
```diff
@@ -680,18 +680,23 @@ export const updateCollectionAddon = ({
...
 };
 
-// The DeleteCollectionAddonNotesAction is identical to the
-// UpdateCollectionAddonAction, with the only difference being that
-// we pass '' into the notes field. Therefore we can reuse the
-// UpdateCollectionAddonParams type to describe the payload.
+type DeleteCollectionAddonNotesParams = {|
+  addonId: number,
+  errorHandlerId: string,
+  notes?: string,
```
willdurand: mmh I guess you added this because you need it for the action's payload. The problem is that now, the action's payload could not have `notes`. Flow would not say anything and I have the feeling that having `notes` would be useful to delete a note.

Maybe we should keep `UpdateCollectionAddonParams` for the parameters and have `DeleteCollectionAddonNotesParams` with `notes` required.

tests/unit/amo/components/TestCollection.js
```diff
@@ -981,6 +989,168 @@ describe(__filename, () => {
...
+      errorHandler,
+      location: fakeRouterLocation({ query: { page } }),
+      store,
+    });
+
+    fakeDispatch.reset();
+
+    // This simulates the user clicking the "Delete" button on the
+    // EditableCollectionAddon component's comment form.
+    root.instance().deleteNote(addonId, errorHandler);
```
willdurand: Would it make sense to access this function _via_ the prop?

```js
const deleteNote = root.find(AddonCard).prop('deleteNote');
deleteNote(addonId, errorHandler);
```

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+    });
+
+    it('shows the read-only version of the notes area if there are notes', () => {
+      const notes = 'Some notes.';
+      const addon = {
+        ...fakeAddon,
+        notes,
+      };
+      const root = render({ addon });
+      const notesArea = root.find('.EditableCollectionAddon-notes');
```
willdurand: Should we add an assertion for this `notesArea`? (like `toHaveLength(1)`?)

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      expect(notesArea.find(Icon)).toHaveProp('name', 'comments-blue');
+      expect(root.find('.EditableCollectionAddon-notes-read-only'))
+        .toHaveLength(1);
+      expect(notesArea.find('.EditableCollectionAddon-notes-content'))
+        .toHaveProp('dangerouslySetInnerHTML', sanitizeHTML(notes));
```
willdurand: We usually use `toHaveHTML` so that we don't call `sanitizeHTML` here and we can assert the content in order to avoid issues such as #5230.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+
+    it('shows an empty notes form when the comment icon is clicked', () => {
+      const root = render();
+
+      expect(root.find('.EditableCollectionAddon-notes')).toHaveLength(0);
+      const commentIcon = root.find('.EditableCollectionAddon-create-note');
+      const clickEvent = createFakeEvent();
+      commentIcon.simulate('click', clickEvent);
+      expect(root.find('.EditableCollectionAddon-notes')).toHaveLength(1);
+      const notesForm = root.find('.EditableCollectionAddon-notes-form');
```
willdurand: Can we have some more blank lines? 😅

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      // The form should not be shown.
+      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
+    });
+
+    it('shows an empty notes form when the comment icon is clicked', () => {
+      const root = render();
+
+      expect(root.find('.EditableCollectionAddon-notes')).toHaveLength(0);
+      const commentIcon = root.find('.EditableCollectionAddon-create-note');
+      const clickEvent = createFakeEvent();
```
willdurand: I don't think we are using this variable more than once, so I'd suggest to not create it.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      // The read-only portion should not be shown.
+      expect(root.find('.EditableCollectionAddon-notes-read-only'))
+        .toHaveLength(0);
+      expect(notesForm).toHaveProp('microButtons', true);
+      expect(notesForm).toHaveProp('onDelete', root.instance().onDeleteNote);
+      expect(notesForm).toHaveProp('onDismiss', root.instance().onDismissNoteForm);
+      expect(notesForm).toHaveProp('onSubmit', root.instance().onSaveNote);
+      expect(notesForm).toHaveProp('placeholder', 'Add a comment about this add-on.');
+      expect(notesForm).toHaveProp('submitButtonText', 'Save');
+      expect(notesForm).toHaveProp('text', undefined);
```
willdurand: mmh why is this `undefined`? `addon.notes` should be either a string or `null`, but not `undefined`.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      const notes = 'Some notes.';
+      const addon = {
+        ...fakeAddon,
+        notes,
+      };
+
+      const root = render({ addon });
+      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
+      const editButton = root.find('.EditableCollectionAddon-notes-edit-button');
+      const clickEvent = createFakeEvent();
```
willdurand: I don't think we are using this variable more than once, so I'd suggest to not create it.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      const notes = 'Some notes.';
+      const addon = {
+        ...fakeAddon,
+        notes,
+      };
+
+      let root = render({ addon });
+      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
+      const editButton = root.find('.EditableCollectionAddon-notes-edit-button');
+      const clickEvent = createFakeEvent();
```
willdurand: I don't think we are using this variable more than once, so I'd suggest to not create it.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+      editButton.simulate('click', clickEvent);
+      let notesForm = root.find('.EditableCollectionAddon-notes-form');
+      expect(notesForm).toHaveLength(1);
+
+      // This simulates the user clicking the "Cancel" button on the
+      // DismissibleTextForm.
+      const onDismissNoteForm = notesForm.props().onDismiss;
```
willdurand: Could we write `.prop('onDismiss')` instead?

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+        notes,
+      };
+      const deleteNote = sinon.spy();
+      const errorHandler = createStubErrorHandler();
+
+      const root = render({ addon, deleteNote, errorHandler });
+
+      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
+      const editButton = root.find('.EditableCollectionAddon-notes-edit-button');
+      const clickEvent = createFakeEvent();
```
willdurand: I don't think we are using this variable more than once, so I'd suggest to not create it.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -76,4 +83,164 @@ describe(__filename, () => {
...
+        notes,
+      };
+      const saveNote = sinon.spy();
+      const errorHandler = createStubErrorHandler();
+
+      const root = render({ addon, errorHandler, saveNote });
+
+      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
+      const editButton = root.find('.EditableCollectionAddon-notes-edit-button');
+      const clickEvent = createFakeEvent();
```
willdurand: I don't think we are using this variable more than once, so I'd suggest to not create it.

src/amo/components/EditableCollectionAddon/index.js
```diff
@@ -56,11 +111,59 @@ export class EditableCollectionAddonBase extends React.Component<Props> {
...
+                  submitButtonText={i18n.gettext('Save')}
+                  text={addon.notes}
+                />
+              </React.Fragment>
+            ) : (
+              <div className="EditableCollectionAddon-notes-read-only">
+                <span
+                  className="EditableCollectionAddon-notes-content"
+                  // eslint-disable-next-line react/no-danger
+                  dangerouslySetInnerHTML={sanitizeHTML(addon.notes)}
```
willdurand: should we respect new lines? If so, we should probably use `nl2br` and white-list `<br>`.

src/amo/components/Collection/index.js
```diff
@@ -228,6 +237,64 @@ export class CollectionBase extends React.Component<Props> {
...
+  deleteNote: DeleteAddonNoteFunc = (
+    addonId: number, errorHandler: ErrorHandlerType
+  ) => {
+    const {
+      collection,
+      dispatch,
+      location,
+    } = this.props;
+
+    invariant(addonId, 'addonId is required');
```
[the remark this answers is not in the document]
bobsilverberg: Good point. I think I was being overzealous with my use of `invariant`. I guess we only need to use it when Flow tells us we need it (because of optional properties).

src/amo/components/EditableCollectionAddon/index.js
```diff
@@ -4,24 +4,64 @@ import invariant from 'invariant';
...
+
+export class EditableCollectionAddonBase extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      editingNote: false,
+    };
+  }
+
+  onEditNote = (event: SyntheticEvent<any>) => {
```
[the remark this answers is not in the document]
bobsilverberg: I could use `HTMLElement` here instead. I'll do that.

src/amo/components/EditableCollectionAddon/index.js
bobsilverberg: True, thanks.

src/amo/reducers/collections.js
bobsilverberg: I'm not sure I fully understand what you've suggested, but what I did was remove `notes` from `DeleteCollectionAddonNotesParams` and used `UpdateCollectionAddonParams` to describe the payload, and that seems to work and I think it supports your case.

tests/unit/amo/components/TestEditableCollectionAddon.js
bobsilverberg: Well, we use it a number of times in the test after that to find more things, so that seems like maybe overkill. The test will fail if it's not found, but I guess we'll get a better error if we include that assertion. I'll add it.

bobsilverberg:
Thanks for the review @willdurand. I have addressed all of your code-level comments and pushed a new commit. Regarding the other issues you raised:

> - does it make sense to display the "delete" button when the user adds notes for the first time?

I think you're right, that doesn't really make sense. I'll change that.

> - should we disable the "submit" button when editing notes without changing anything?

Yes, that makes sense too. That seems like more of an issue with the `DismissibleTextForm` that would manifest in other places it's used, but I can include that change in this PR.

I'll make those additional changes and push another commit for them.

bobsilverberg:
Well, those last two changes were actually far easier than I thought they'd be. Everything has now been addressed and I've pushed another commit.

bobsilverberg:
Thanks @willdurand. I know it was a huge PR. Hopefully there won't be another quite so big for some time.

PR #5254, opened 2018-06-08
Display comments for collection addons

### LINKED ISSUE #3890: Collection comments missing from new layout
[Re-filed from mozilla/addons-server#6892]

This depends on https://github.com/mozilla/addons-frontend/issues/4193

When adding comments to an add-on when creating a collection, these comments are no longer shown to users of the collection with the new layout.

### What happened?
Comments can still be entered as part of the collection, but not shown to users of the collection.

I should be able to enter the comments and users should be able to see them.

Comments can still be entered:

However, the view of the collection no longer displays any of the comments.

For me, the usefulness of the comments was to help users understand WHY either I use them or why THEY might want to use them.  

I don't have a screenshot of my comments in the old format and was only able to find an old screenshot  via Google Images.

bobsilverberg:
Fixes #3890 

Depends on #5240 

Screenshot of collection details page with notes:

screenshot 2018-06-08 08 48 29

### DIFF
--- a/src/amo/components/SearchResult/index.js
+++ b/src/amo/components/SearchResult/index.js
@@ -134,7 +134,7 @@ export class SearchResultBase extends React.Component<Props> {
                 <Icon name="comments-blue" />
                 {i18n.gettext('Add-on note')}
               </h4>
-              <div
+              <p
                 className="SearchResult-note-content"
                 // eslint-disable-next-line react/no-danger
                 dangerouslySetInnerHTML={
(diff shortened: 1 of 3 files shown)


src/amo/components/SearchResult/index.js
```diff
@@ -127,6 +127,22 @@ export class SearchResultBase extends React.Component<Props> {
...
             </div>
           ) : null}
+
+          {addon && addon.notes && (
+            <div className="SearchResult-note">
+              <h4 className="SearchResult-note-header">
+                <Icon name="comments-blue" />
+                {i18n.gettext('Add-on note')}
+              </h4>
+              <div
```
willdurand: This could be a paragraph, no?

bobsilverberg: Sure, why not? 😄

PR #5256, opened 2018-06-08
Only scroll the first time we render an error or message

### LINKED ISSUE #5250: Edit profile page scrolls up when trying to edit a field after an error is generated
STR:
1. Log in to AMO-dev
2. Go to Edit my Profile
3. Edit some of the fields, making sure that at least one of the values entered will generate an error (i.e. an invalid image type)
4. Click on update profile
5. After the corresponding errors are generated, scroll down and edit one of the available fields

Actual result:
The Edit profile page will automatically scroll up when trying to edit a field after an error is generated

Expected result:
The Edit profile page is not moving when trying to edit a field after an error is generated

Notes:
- reproduced on AMO -dev with FF60, Win10x64
- it think this was  introduced by #5224 

scroll up

willdurand:
Fix #5250

---

This PR reduces the number calls to `window.scroll`: it is not called
when we re-render because of a new error or a success message, but only
the first time!

### DIFF
--- a/tests/unit/amo/components/TestUserProfileEdit.js
+++ b/tests/unit/amo/components/TestUserProfileEdit.js
@@ -815,10 +815,7 @@ describe(__filename, () => {
       userId: user.id,
     });
 
-    const _window = {
-      scroll: sinon.spy(),
-    };
-    const root = renderUserProfileEdit({ errorHandler, store, _window });
+    const root = renderUserProfileEdit({ errorHandler, store });
 
     expect(root.find(Notice)).toHaveLength(0);
     expect(root.find('.UserProfileEdit-submit-button'))
@@ -835,8 +832,6 @@ describe(__filename, () => {
 
     expect(root.find('.UserProfileEdit-submit-button'))
       .toHaveProp('disabled', false);
-
-    sinon.assert.calledWith(_window.scroll, 0, 0);
   });
 
   it('renders a Not Found page when logged-in user cannot edit another user', () => {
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1400,4 +1401,81 @@ describe(__filename, () => {
...
+    sinon.assert.notCalled(_window.scroll);
+
+    root.setProps({
+      errorHandler: createStubErrorHandler(new Error('some error')),
+    });
+
+    sinon.assert.calledWith(_window.scroll, 0, 0);
+  });
+
+  it('does not scroll if we already scrolled because of an error', () => {
```
bobsilverberg: This test is described as `we already scrolled because of an error` but it doesn't actually cause a scroll because of an error first, it just starts off with an activated errorHandler, so I don't think the test case (exactly) matches the description. Would it be better to do what you've done in `it('scrolls to the top of the page when an error is rendered')`, above, as the set-up for this test, so we actually do scroll first, and then check that we don't scroll again on re-render?

Either that, or change the description if what we're really testing is a pre-existing error, and not whether we've already scrolled or not.

willdurand: ah right, I wanted to avoid the scroll set-up and directly inject the error handler with an error because this was not super-important (since in both cases, it leads to the same result), but I understand that it is not crystal clear, so I have added the set-up to first scroll and then make sure we do not scroll again.

PR #5257, opened 2018-06-08
Fixes #5177

### LINKED ISSUE #5177: Add Flow type for users API
The user API does not have Flow types for the return values. We started to add
return types for collections, and we should be doing this for all APIs.

Gabbyjose:
Updated the User.js file in api to include return types based on the user saga and user reducer. Left a couple of comments on the user api file with some questions to make sure I have the right value in which the promise will resolve to


willdurand:
Hi @Gabbyjose, so this looks good but you have huge merge conflicts. I think you should start a new branch from the latest `master` branch, and add the changes you did in the files `src/amo/api/users.js` and `src/amo/reducers/users.js`. Those changes look good to me.

Gabbyjose:
Thanks @willdurand, will do that!

PR #5261, opened 2018-06-08
updating react-transition-group version and components

### LINKED ISSUE #5155: Update react-transition-group to the latest version and the components that use it
the react-transition-group package has a newer version and we should look into upgrading to that (either through Greenkeeper if possible (?) or explicitly updating in package.json)

The package is used currently in the Disco and Collection Manager so these components need to be adjusted to work with the newer version

championshuttler:
Fixes #5155

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -5,7 +5,7 @@ import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
-import ReactCSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
+import { TransitionGroup, CSSTransition } from 'react-transition-group';
 import { compose } from 'redux';
 import config from 'config';
 
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -346,19 +346,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
           />
         </div>
 
-        <ReactCSSTransitionGroup
+        <TransitionGroup
           className="NoticePlaceholder"
-          component="div"
-          transitionName="NoticePlaceholder-transition"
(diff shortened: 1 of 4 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -346,19 +346,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
           className="NoticePlaceholder"
-          component="div"
-          transitionName="NoticePlaceholder-transition"
-          transitionEnterTimeout={MESSAGE_FADEOUT_TIME}
-          transitionLeaveTimeout={MESSAGE_FADEOUT_TIME}
         >
-          {this.state.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
+          <CSSTransition
+            component={null}
+            timeout={MESSAGE_FADEOUT_TIME}
```
championshuttler: I think we can remove the constant `MESSAGE_FADEOUT_TIME` and direct pass the value `450` to timeout

src/amo/components/CollectionManager/index.js
```diff
@@ -346,19 +346,21 @@ export class CollectionManagerBase extends React.Component<Props, State> {
...
+            classNames="NoticePlaceholder-transition"
+          >
+            {this.state.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
             <Notice type="success">
               {i18n.gettext('Added to collection')}
             </Notice>
-          )}
-        </ReactCSSTransitionGroup>
+            )}
+          </CSSTransition>
```
championshuttler: @rebmullin , I am little confused about where to add `CSSTransition`.

src/amo/components/CollectionManager/index.js
willdurand: no, we should use constants

rebmullin:
Hey @championshuttler, let me know if you need any assistance with this one : )

championshuttler:
Hi @rebmullin when I am running `yarn disco` with your code its showing some errors and not hosting on localhost properly

PR #5275, opened 2018-06-12
Allow a single section to fails to load on the home page without breaking the home page

### LINKED ISSUE #4387: If a single section fails to load on the home page, all of them fail to load
Currently, if any of the sections defined on the home page via `LandingAddonsCard` components fails to load its data (for example, if a collection does not exist in the environment), then all of the sections fail to load and an error message is displayed at the top of the page:

feb-14-2018 07-46-18

We should probably fix this so that any sections that _can_ load _do_ load. There was a suggestion to just hide any sections that fail to load, but I am concerned if we do that, and there's a problem with a section, we might never know about it. Maybe we could display an error message in the section that fails to load? Or is there a way we can make sure we are informed if a situation arises where a section fails to load and is subsequently hidden?

bobsilverberg:
Fixes #4387

### DIFF
--- a/tests/unit/amo/reducers/test_home.js
+++ b/tests/unit/amo/reducers/test_home.js
@@ -56,6 +56,30 @@ describe(__filename, () => {
       ]);
     });
 
+    it('loads a null for a missing collection', () => {
+      const { store } = dispatchClientMetadata();
+
+      store.dispatch(loadHomeAddons({
+        collections: [null],
+        featuredExtensions: createAddonsApiResult([fakeAddon]),
+        featuredThemes: createAddonsApiResult([fakeTheme]),
+      }));
+
+      const homeState = store.getState().home;
+
+      expect(homeState.resultsLoaded).toEqual(true);
+      expect(homeState.collections)
+        .toHaveLength(1);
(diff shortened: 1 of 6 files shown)


bobsilverberg:
@willdurand This should be ready for a review now, thanks!

tests/unit/amo/reducers/test_home.js
```diff
@@ -56,6 +56,30 @@ describe(__filename, () => {
...
+        featuredThemes: createAddonsApiResult([fakeTheme]),
+      }));
+
+      const homeState = store.getState().home;
+
+      expect(homeState.resultsLoaded).toEqual(true);
+      expect(homeState.collections)
+        .toHaveLength(1);
+      expect(homeState.collections[0])
+        .toEqual(null);
```
willdurand: Nit: those two lines above could be put on a single line.

jrbenny35:
Thanks for this Bob!!!

bobsilverberg:
No problem, @jrbenny35. This should fix the tests that fail every time we push changes to the home page. If it doesn't, please let me know.

PR #5282, opened 2018-06-14
Send errors to Sentry

### LINKED ISSUE #5278: Make sure we report errors to Sentry
In #4051, we introduced a component to catch all uncaught errors (using `componentDidCatch()`). When an error occurs, we serve a server error page to the end user and we log the error.

~~The problem is that this error is lost because logging happens on the client side, and we never get any notification and we do not send anything to Sentry.~~

~~Before this patch #4051, Sentry was notified because the sentry lib catches all uncaught errors, but it is not the case anymore.~~ => We should make sure that logging with `console.error` sends data to Sentry.

We should send errors caught by the `ErrorPage` component as described here: https://docs.sentry.io/clients/javascript/integrations/react/#expanded-usage

willdurand:
Fix #5278

---

This PR sends errors to Sentry when possible.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -16,6 +16,7 @@
 # TODO: this can go away after
 # https://github.com/facebook/flow/issues/2801
 ./flow/libs/intl.js.flow
+./flow/libs/raven.js.flow
 
 [options]
 # This maps all Sass/SCSS imports to a dummy Flow file to suppress import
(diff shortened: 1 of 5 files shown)


willdurand:
Well, that does not work as intended actually. Closing for now, and will reopen later.

PR #5284, opened 2018-06-14
Update SimulateClientError to trigger render errors

### LINKED ISSUE #5278: Make sure we report errors to Sentry

willdurand:
Fix #5278 

---

This PR adds a new button to the `SimulateClientError` that throws an error in the `render()` method, in order to test our integration with Sentry (and `componentDidCatch()` in `ErrorPage`). It also adds more Flow coverage.

### DIFF
--- a/src/core/components/ErrorPage/index.js
+++ b/src/core/components/ErrorPage/index.js
@@ -1,28 +1,35 @@
+/* @flow */
 import * as React from 'react';
-import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import log from 'core/logger';
 import { getErrorComponent as getErrorComponentDefault } from 'core/utils';
 import { loadErrorPage } from 'core/reducers/errorPage';
+import type { ErrorPageState } from 'core/reducers/errorPage';
+import type { DispatchFunc } from 'core/types/redux';
 
 
-export class ErrorPageBase extends React.Component {
-  static propTypes = {
-    children: PropTypes.node.isRequired,
-    dispatch: PropTypes.func.isRequired,
-    errorPage: PropTypes.object,
-    getErrorComponent: PropTypes.func,
-  }
+type Props = {|
+  children: React.Node,
+  dispatch: DispatchFunc,
+  errorPage: {
+    error: Error,
+    hasError: boolean,
+    statusCode: number,
+  },
+  getErrorComponent: Function,
+|};
 
+export class ErrorPageBase extends React.Component<Props> {
   static defaultProps = {
     errorPage: {},
     getErrorComponent: getErrorComponentDefault,
   }
 
-  componentDidCatch(error, info) {
+  componentDidCatch(error: Error, info: Object) {
     const { dispatch } = this.props;
+
     dispatch(loadErrorPage({ error }));
     log.error('Caught application error:', error, info);
   }
@@ -41,8 +48,7 @@ export class ErrorPageBase extends React.Component {
   }
 }
 
-
-export const mapStateToProps = (state) => ({
+export const mapStateToProps = (state: {| errorPage: ErrorPageState |}) => ({
   errorPage: state.errorPage,
 });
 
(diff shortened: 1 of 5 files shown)


tests/unit/core/containers/error-simulation/TestSimulateClientError.js
```diff
@@ -20,27 +19,49 @@ describe('SimulateClientError', () => {
...
+    );
   }
 
   it('lets you trigger an error', () => {
     const root = render();
-    const button = root.querySelector('button');
-    expect(() => Simulate.click(button)).toThrowError(/simulated client error/);
+
+    expect(() => {
+      root.find(Button).at(0).simulate('click');
```
kumar303: Instead of `find(Button).at(0)` I'd rather see it query by class name now that we have two buttons.

src/core/containers/error-simulation/SimulateClientError.js
```diff
@@ -9,21 +9,41 @@ import './SimulateClientError.scss';
...
       'Nice! Check Sentry' : '💣 Go ahead, trigger an error';
+
+    if (this.state.throwRenderError) {
+      throw new Error('This is a simulated client render error');
+    }
+
     return (
       <div className="SimulateClientError">
-        <Button buttonType="neutral" onClick={this.onClick}>{prompt}</Button>
+        <Button buttonType="neutral" onClick={this.onClick}>
```
kumar303: While you're in here, it would be nice if the component showed a paragraph explaining why this simulation page exists and what you can do with it. Maybe try to think of something that would have helped you when you first saw it.

src/core/reducers/errorPage.js
```diff
@@ -1,15 +1,38 @@
+/* @flow */
 import { LOCATION_CHANGE } from 'react-router-redux';
 
-const LOAD_ERROR_PAGE = 'LOAD_ERROR_PAGE';
+const LOAD_ERROR_PAGE: 'LOAD_ERROR_PAGE' = 'LOAD_ERROR_PAGE';
 
-export const initialState = {
+export type ErrorPageState = {|
+  clearOnNext: boolean,
+  error: Object | null,
```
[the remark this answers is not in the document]
willdurand: Yes, I am a bit confused though. I pushed an update with my understanding of the current code base.

willdurand:
Updated the page:

screen shot 2018-06-19 at 13 33 25

src/core/components/ErrorPage/index.js
```diff
@@ -1,28 +1,35 @@
...
-  }
-
+import type { ErrorPageState } from 'core/reducers/errorPage';
+import type { DispatchFunc } from 'core/types/redux';
+
+
+type Props = {|
+  children: React.Node,
+  dispatch: DispatchFunc,
+  errorPage: {
```
kumar303: What I meant in the last review is that this can be:

```js
type Props = {|
  errorPage: ErrorPageState,
  ...
|};
```

src/core/components/ErrorPage/index.js
```diff
@@ -1,28 +1,35 @@
...
+    error: Error,
+    hasError: boolean,
+    statusCode: number,
+  },
+  getErrorComponent: Function,
```
kumar303: this can be `typeof getErrorComponentDefault`

kumar303:
I finished the review and I don't have any additional change requests. Thanks for adding the description -- this is really helpful.

src/core/components/ErrorPage/index.js
willdurand: ah... right 🤦‍♂️

willdurand:
> I finished the review and I don't have any additional change requests. Thanks for adding the description -- this is really helpful.

Thanks for the help on this one. I wish we could make those pages more visible somehow...

PR #5289, opened 2018-06-15
Fix the problem with private collections on the home page

### LINKED ISSUE #5287: Collection shelf titles are moving to the next shelf if one of the homepage collections is made private
STR:
1. Log in to AMO dev with an Admin account
2. Click on See more for one of the homepage collections  - i.e. Translation Tools is the best example
3. Click on Edit collection and switch to the legacy site
4. Change the visibility status of this collection to Pivate
5. Open AMO -dev homepage in a separate tab - make sure you are logged out
6. Check the collection shelves' titles

Actual result:
The shelf titles are moving to the next available collection. In this case, Translation Tools title moves to the Privacy matters shelf and the Privacy matters title moves to the Tame your Tabs shelf

Expected result:
Shelf titles stay relevant to the collection shelf they belong to

Notes:
- I know that the Private option is not going to be used anymore when the new collection pages will land, but this seemed to be an interesting issue which should be looked into in any case
- reproduced on AMO -dev with FF60 WIn10x64 and looks most likely as a regression from #4387 

image

bobsilverberg:
Fixes #5287

### DIFF
--- a/tests/unit/amo/sagas/test_home.js
+++ b/tests/unit/amo/sagas/test_home.js
@@ -141,141 +141,76 @@ describe(__filename, () => {
       expect(loadAction).toEqual(expectedLoadAction);
     });
 
-    it('loads a null for a collection that returns a 401', async () => {
-      const state = sagaTester.getState();
+    it.each([401, 404])(
+      'loads a null for a collection that returns a %i',
+      async (status) => {
+        const state = sagaTester.getState();
 
-      const error = createApiError({ response: { status: 401 } });
+        const error = createApiError({ response: { status } });
 
-      const firstCollectionSlug = 'collection-slug';
-      const firstCollectionUser = 'user-id-or-name';
+        const firstCollectionSlug = 'collection-slug';
+        const firstCollectionUser = 'user-id-or-name';
 
-      const baseArgs = { api: state.api };
-      const baseFilters = {
-        page_size: LANDING_PAGE_ADDON_COUNT,
-      };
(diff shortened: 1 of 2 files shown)


tests/unit/amo/sagas/test_home.js
```diff
@@ -141,7 +141,75 @@ describe(__filename, () => {
       expect(loadAction).toEqual(expectedLoadAction);
     });
 
-    it('loads a null for a collection that cannot be fetched', async () => {
+    it('loads a null for a collection that returns a 401', async () => {
+      const state = sagaTester.getState();
```
willdurand: I'd like to try this new Jest 23 feature called "table driven tests" 🔥: https://facebook.github.io/jest/blog/#jest-each. (Documentation: https://facebook.github.io/jest/docs/en/api.html#testeachtable-name-fn)

We could write this test case that way:

```js
it.each([401, 404])('loads a null for a collection that returns a %d', async (status) => {
  // ...
});
```

WDYT? This patch is small enough to give it a try I'd say. I think we'll have to upgrade to Jest 23 though :)

bobsilverberg: Ok, I've updated the test to use `it.each` and it works well. Thanks for the tip. This now depends on PR #5300 landing first.

bobsilverberg:
Oops, @willdurand, I thought this had an r+wc from you, but I see it didn't. I just merged it, so if you want to take a look I can post a follow-up patch if there are any issues.

PR #5291, opened 2018-06-15
Add missing biography title in user profile

### LINKED ISSUE #5145: Biography section is missing a title on the User profile card
STR

1. Log in to AMO-dev
2. Open a user profile page that includes a biography (https://addons-dev.allizom.org/en-US/firefox/user/MaDonna-3/)
3. Check the biography section

Actual result:
There is no title marking the beginning of the user biography

Expected result:
The biography section should include a title, in order to be consistent with the rest of the user data listed on the profile card

Notes:
- reproduced on AMO dev with FF60, Win10x64

image

willdurand:
Fix #5145

---

This PR adds a title for the bio of a user (if supplied).

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -194,15 +194,18 @@ export class UserProfileBase extends React.Component<Props> {
                   />
                 ) : <LoadingText />}
               </Definition>
+              {user && user.biography && user.biography.length ? (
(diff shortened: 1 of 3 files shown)


willdurand:
> for consistency with the other dl/dds, did you want to remove the bio's p margin-top style?

good catch, thanks!

PR #5295, opened 2018-06-16
Fixes 4599: adds user type in user profile

### LINKED ISSUE #4599: Add-ons developer and Theme artist options are not available on the user's info card
As a followup for #4283 
STR:
1. Load AMO dev and click on View Profile.
2. Check if Add-ons developer and Theme Artist are present on the right upper part of the card as in [here](https://user-images.githubusercontent.com/90871/35629616-fcb243f4-0696-11e8-8b0b-024e96ec04cd.png)

Expected result:
The check boxes and the two options are present.

Actual result:
The two options are not displayed.

Notes:

infocard

tsl143:
Fixes #4599 

**Before**
image

**After**
screen shot 2018-06-16 at 10 12 15 pm


src/amo/components/UserProfile/index.js
```diff
@@ -115,13 +115,29 @@ export class UserProfileBase extends React.Component<Props> {
...
-      <React.Fragment>
+      <div className="UserProfile-head">
         <UserAvatar className="UserProfile-avatar" user={user} />
 
+        <div className="UserProfile-tags">
+          {
+            user && user.is_addon_developer &&
+            <div className="UserProfile-developer">
+              {i18n.gettext('Add-ons developer')}
+              <i className="developer" />
```
willdurand: I think we should use the `Icon` component here.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -41,6 +41,8 @@ describe(__filename, () => {
       display_name: 'Matt MacTofu',
       userId: 500,
       username: 'tofumatt',
+      is_addon_developer: true,
+      is_artist: true,
       ...props,
```
willdurand: we should probably do that in each test case (one for each icon separately).

tsl143:
@willdurand got it, actually `@include icon($name: "developer");` and `isDeveloper` are nice functions :)

willdurand:
Then again, thanks for contributing @tsl143!

Issue #5234, opened 2018-06-07
A component is changing an uncontrolled input of type text to be controlled


willdurand:
When creating a new collection, I got this warning in the devtools:

```
Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by CollectionManagerBase)
    in form (created by CollectionManagerBase)
    in CollectionManagerBase (created by Connect(CollectionManagerBase))
    in Connect(CollectionManagerBase) (created by Connect(Connect(CollectionManagerBase)))
    in Connect(Connect(CollectionManagerBase)) (created by Translate[Connect(Connect(CollectionManagerBase))])
    in Translate[Connect(Connect(CollectionManagerBase))] (created by withRouter(Translate[Connect(Connect(CollectionManagerBase))]))
    in withRouter(Translate[Connect(Connect(CollectionManagerBase))]) (created by CollectionListBase)
    in div (created by Card)
    in section (created by Card)
    in Card (created by CollectionListBase)
    in div (created by CollectionListBase)
    in div (created by CollectionListBase)
    in CollectionListBase (created by Connect(CollectionListBase))
    in Connect(CollectionListBase) (created by Translate[Connect(CollectionListBase)])
    in Translate[Connect(CollectionListBase)] (created by component)
    in component (created by RouterContext)
    in ErrorPageBase (created by Connect(ErrorPageBase))
    in Connect(ErrorPageBase) (created by AppBase)
    in div (created by AppBase)
    in div (created by AppBase)
    in div (created by AppBase)
    in NestedStatus (created by SideEffect(NestedStatus))
    in SideEffect(NestedStatus) (created by AppBase)
    in AppBase (created by Translate[AppBase])
    in Translate[AppBase] (created by Connect(Translate[AppBase]))
    in Connect(Translate[AppBase]) (created by RouterContext)
    in RouterContext (created by Router)
    in ScrollBehaviorContext (created by Router)
    in Router
    in Provider
    in I18nProvider
```

I see that the internal state of the `CollectionManager` component as `undefined` values for the corresponding `input` values. It should never be `undefined` (or `null`), only strings that can be empty.

[...]

PR #5303, opened 2018-06-18
Move the create collection feature from the CollectionList to the Collection component

### LINKED ISSUE #5302: Move the create collection feature from the CollectionList to the Collection component
This is some refactoring that makes sense to do before tackling issue #3142. The create collection feature was located at the `CollectionList` component back when we were still using an overlay for the form, but it doesn't make sense to have it there anymore. This will remove the `CollectionList` component and use the `Collection` component instead for the "create a collection" feature.

For QA:

Please check for any regressions of the "create a collection", "edit a collection" and "view a collection" features.

bobsilverberg:
Fixes #5302


tests/unit/amo/components/TestCollection.js
```diff
@@ -145,6 +161,21 @@ describe(__filename, () => {
     expect(wrapper.find('.Collection-placeholder')).toHaveLength(0);
   });
 
+  it('hides placeholder text if the user is not logged in', () => {
```
willdurand: Isn't it a duplicate test? If user is not logged in, there is no placeholder text. The other test `it hides placeholder text when creating a collection if not logged in'` already covers this case, no?

tests/unit/amo/components/TestCollection.js
```diff
@@ -855,6 +915,23 @@ describe(__filename, () => {
...
 
+  it('renders AuthenticateButton when creating and not signed in', () => {
+    const { store } = dispatchClientMetadata();
+    const location = fakeRouterLocation({
+      pathname: '/create/url/',
+    });
+    const root = renderComponent({ store, creating: true, location });
+
+    const authButton = root.find(AuthenticateButton);
+    expect(authButton).toHaveProp('location', location);
```
willdurand: we could maybe have an expectation for the button text, since it can be different.

bobsilverberg:
> I noticed these warnings in the devtools:
>
>```
>Warning: `value` prop on `input` should not be null. Consider using an empty string >to clear the component or `undefined` for uncontrolled components.
>```

I'll look into this one.

>```
>Warning: A component is changing an uncontrolled input of type text to be >controlled. Input elements should not switch from uncontrolled to controlled (or >vice versa). Decide between using a controlled or uncontrolled input element for >the lifetime of the component. More info: https://fb.me/react-controlled-components
>```

Is this just a duplicate of #5234, or is this another instance of the same problem? Do you know if this is new with this PR?

willdurand:
> Is this just a duplicate of #5234, or is this another instance of the same problem? Do you know if this is new with this PR?

mmh I think it is the same, sorry for the confusion. I would even say that both warnings reported above are related.

bobsilverberg:
> mmh I think it is the same, sorry for the confusion. I would even say that both warnings reported above are related.

In that case I'd prefer to address #5234 as a follow-up, as it seems to be unrelated to any code landing now.

willdurand:
> In that case I'd prefer to address #5234 as a follow-up, as it seems to be unrelated to any code landing now.

Works for me!

tests/unit/amo/components/TestCollection.js
```diff
@@ -129,6 +129,22 @@ describe(__filename, () => {
...
+
+    const wrapper = renderComponent({ creating: true, store });
+
+    expect(wrapper.find('.Collection-placeholder')).toHaveLength(1);
+    expect(wrapper.find('.Collection-placeholder').text()).toEqual(
+      'First, create your collection. Then you can add extensions and themes.');
+  });
+
+  it('hides placeholder text when creating a collection if not logged in', () => {
+    const wrapper = renderComponent({ creating: true });
```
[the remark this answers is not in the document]
bobsilverberg: It's the default, but I guess we should, yeah, in the interest of the setup for the test being explicit. 👍

tests/unit/amo/components/TestCollection.js
bobsilverberg: There's only one code path: you are correct that if a user is not logged in there will be no placeholder, regardless of any other circumstances, but it's a different use case. The earlier test is for the "create" use case, while this is for the "display" use case. We know the code, so we can say it's testing the same thing, but is it still worth having tests for the 2 different use cases? If someone were to change the implementation and we didn't have this case it's possible that a bug could be introduced. I'll update the description to make it more obvious that it's different than the "create" test.

PR #5307, opened 2018-06-18
I4917 screenshot size fix

### LINKED ISSUE #4197: Add-on screenshot images up-scaling degrades quality
I noticed that on AMO screens/images are up-scaled which causes them to lose clarity, especially for small images.

Example:
https://addons.mozilla.org/en-US/firefox/addon/foxytab/

Transferred from: https://github.com/mozilla/addons/issues/627

rebmullin:
fixes #4197 - 

instead of scaling up we are now we are using the actual image size. My understanding is that eventually we want to use srcset when the images are all set up (let me know if this assumption is not correct)

With the smaller image it looks a little odd so I have also centered the caption a bit more (it was left aligned in it's container which was centered). 

before:
Alt text

after:
Alt text

### DIFF
--- a/src/amo/components/ScreenShots.js
+++ b/src/amo/components/ScreenShots.js
@@ -8,8 +8,6 @@ import 'react-photoswipe/lib/photoswipe.css';
 
 import 'amo/css/ScreenShots.scss';
 
-export const HEIGHT = 1024;
-export const WIDTH = 1366; // This is the same as our $max-content-width;
 const PHOTO_SWIPE_OPTIONS = {
   closeEl: true,
   captionEl: true,
@@ -37,14 +35,14 @@ const formatPreviews = (previews) => (
   previews.map((preview) => ({
     src: preview.image_url,
     thumbnail_src: preview.thumbnail_url,
-    h: HEIGHT,
-    w: WIDTH,
+    h: preview.image_size[1],
+    w: preview.image_size[0],
     title: preview.caption,
   }))
 );
 
 export const thumbnailContent = (item) => (
-  <img src={item.src} className="ScreenShots-image" height={HEIGHT} width={WIDTH} alt="" title={item.title} />
+  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt="" title={item.title} />
 );
 
 export default class ScreenShots extends React.Component {
--- a/src/amo/components/ScreenShots.js
+++ b/src/amo/components/ScreenShots.js
@@ -42,7 +42,7 @@ const formatPreviews = (previews) => (
 );
 
 export const thumbnailContent = (item) => (
-  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt="" title={item.title} />
+  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt={item.title} title={item.title} />
 );
 
 export default class ScreenShots extends React.Component {
--- a/src/amo/components/ScreenShots.js
+++ b/src/amo/components/ScreenShots.js
@@ -42,7 +42,14 @@ const formatPreviews = (previews) => (
 );
 
 export const thumbnailContent = (item) => (
-  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt={item.title} title={item.title} />
+  <img
+    alt={item.title}
+    className="ScreenShots-image"
+    height={item.h}
+    src={item.src}
+    title={item.title}
+    width={item.w}
+  />
 );
 
 export default class ScreenShots extends React.Component {
--- a/src/amo/css/ScreenShots.scss
+++ b/src/amo/css/ScreenShots.scss
@@ -7,8 +7,14 @@ $screenshot-height: 220px;
 $image-height: $screenshot-height - 20px;
 $screenshot-width: 320px;
 
-.ScreenShots .pswp-thumbnails {
-  line-height: 0;
+.ScreenShots {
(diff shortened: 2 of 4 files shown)


tests/unit/amo/components/TestScreenShots.js
```diff
@@ -4,22 +4,27 @@ import { renderIntoDocument } from 'react-dom/test-utils';
...
 import ScreenShots, {
-  HEIGHT,
-  WIDTH,
   thumbnailContent,
 } from 'amo/components/ScreenShots';
 
+const HEIGHT = 400;
+const WIDTH = 200;
+
 describe('<ScreenShots />', () => {
```
willdurand: Could you please update this description with `__filename`?

tests/unit/amo/components/TestScreenShots.js
```diff
@@ -49,7 +54,7 @@ describe('<ScreenShots />', () => {
   });
 
   it('renders custom thumbnail', () => {
-    const item = { src: 'https://foo.com/img.png', title: 'test title' };
+    const item = { src: 'https://foo.com/img.png', title: 'test title', h: HEIGHT, w: WIDTH };
```
willdurand: It would be better to pass different height/width here to make sure we use the values from `image_size`.

tests/unit/amo/components/TestScreenShots.js
willdurand: Also, I see that we are using `renderIntoDocument`. It could be nice to rewrite the test case with Enzyme.

src/amo/css/ScreenShots.scss
```diff
@@ -7,8 +7,14 @@ $screenshot-height: 220px;
 $image-height: $screenshot-height - 20px;
 $screenshot-width: 320px;
 
-.ScreenShots .pswp-thumbnails {
-  line-height: 0;
+.ScreenShots {
+  .pswp-thumbnails {
+    line-height: 0;
```
willdurand: Why is it needed?

src/amo/components/ScreenShots.js
```diff
@@ -37,14 +35,14 @@ const formatPreviews = (previews) => (
...
-    w: WIDTH,
+    h: preview.image_size[1],
+    w: preview.image_size[0],
     title: preview.caption,
   }))
 );
 
 export const thumbnailContent = (item) => (
-  <img src={item.src} className="ScreenShots-image" height={HEIGHT} width={WIDTH} alt="" title={item.title} />
+  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt="" title={item.title} />
```
willdurand: We could probably use `item.title` as `alt` content. It is not a super-best-practice, but it is better than having an empty `alt` I guess.

src/amo/css/ScreenShots.scss
rebmullin: do you mean the `line-height` style?  This was on the line above I just moved it so I could nest the caption style as well

willdurand: heh. 🤕

tests/unit/amo/components/TestScreenShots.js
rebmullin: ahh..ok sure!

src/amo/components/ScreenShots.js
willdurand: For accessibility, we should not have the same content for both `alt` and `title`.

`title` should be the title of the image (e.g. [The Treachery of Images](https://en.wikipedia.org/wiki/The_Treachery_of_Images)) and `alt` should be a description of what the image shows (e.g., `The picture shows a pipe. Below it, Magritte painted, "Ceci n'est pas une pipe.", French for "This is not a pipe."`).

src/amo/components/ScreenShots.js
```diff
@@ -37,14 +35,14 @@ const formatPreviews = (previews) => (
...
-    w: WIDTH,
+    h: preview.image_size[1],
+    w: preview.image_size[0],
     title: preview.caption,
   }))
 );
 
 export const thumbnailContent = (item) => (
-  <img src={item.src} className="ScreenShots-image" height={HEIGHT} width={WIDTH} alt="" title={item.title} />
+  <img src={item.src} className="ScreenShots-image" height={item.h} width={item.w} alt={item.title} title={item.title} />
```
willdurand: I guess we could also write the props on different lines:

```
<img
  src={item.src}
  ...
/>
```

tests/unit/amo/components/TestScreenShots.js
```diff
@@ -49,14 +53,14 @@ describe('<ScreenShots />', () => {
   });
 
   it('renders custom thumbnail', () => {
-    const item = { src: 'https://foo.com/img.png', title: 'test title' };
+    const item = { src: 'https://foo.com/img.png', title: 'test title', h: 123, w: 1234 };
```
willdurand: You could introduce the variables `h` and `w`.

tests/unit/amo/components/TestScreenShots.js
```diff
@@ -53,7 +53,10 @@ describe(__filename, () => {
   });
 
   it('renders custom thumbnail', () => {
-    const item = { src: 'https://foo.com/img.png', title: 'test title', h: 123, w: 1234 };
+    const h = 123;
+    const w = 1234;
```
willdurand: you should reuse these variables in the `expect().toEqual()`.

Issue #5306, opened 2018-06-18
Autocomplete search results on mobile are awkwardly small


kumar303:
### What happened?

* Load AMO on a mobile device with a small screen
* Type the name of an add-on in the search box
* Wait for search results to autocomplete
* Read the results and select one that matches your query

This technically works but the results are very small which makes them awkward to deal with: 
* The text gets cut off so it's hard to read the results. 
* The click targets are not very big so it's hard to select results.

### What did you expect to happen?

Autocomplete results should be easier to deal with on mobile.

ioanarusiczki:
Verified this issue on AMO dev and AMO stage with FF60(Android 8.0)
The search suggestions are easier to read and to be selected - what I see is that the user/display name won't be entirely displayed if it's a bit longer because of the search bar.

![search bar and suggestions on mobile device](https://user-images.githubusercontent.com/33448286/41648669-3c2c00ca-7482-11e8-9f3e-f78fb15139b7.png)

kumar303:
> what I see is that the user/display name won't be entirely displayed if it's a bit longer because of the search bar.

I think that's fine. It's enough of a hint to indicate that tapping on it will give you a menu of user related actions.

PR #5308, opened 2018-06-18
Make the search box bigger on small mobile screens

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5306

Autocomplete search results on mobile are too small. This maximizes them in the current layout as a simple solution. A more complex solution would be to make the search box go full width when you focus on it (and hide the login button) but that requires a lot of new state management logic (I think).

## Before

screenshot 2018-06-18 11 33 48
screenshot 2018-06-18 11 32 07

## After

screenshot 2018-06-18 11 43 51
screenshot 2018-06-18 11 44 28

This improvement will look more dramatic on wider screen devices, such as Android Pixels.

### DIFF
--- a/src/amo/components/Header/styles.scss
+++ b/src/amo/components/Header/styles.scss
@@ -7,7 +7,7 @@
 .Header {
   background: $ink-90;
   display: grid;
-  grid-template-columns: auto min-content;
+  grid-template-columns: auto auto auto min-content;
   min-height: 112px;
   padding: 0 12px;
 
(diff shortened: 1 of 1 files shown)


src/amo/components/Header/styles.scss
```diff
@@ -7,7 +7,7 @@
 .Header {
   background: $ink-90;
   display: grid;
-  grid-template-columns: min-content auto;
+  grid-template-columns: auto auto auto min-content;
```
willdurand: Maybe we could use `repeat(3, auto) min-content`

kumar303: Ah, cool, I'll try it. At some point we should redo this grid to use named areas. It's pretty ridiculous with numbered columns but I didn't want to tackle it in this PR.

PR #5309, opened 2018-06-18
List user collections

### LINKED ISSUE #3142: List my collections and edit / delete / add
Refs: https://github.com/mozilla/addons-frontend/issues/2782

No mocks for this one yet. 

This will be a page to view/edit/add your collections, sorted by last updated.

For comparison

collections_by_muffinresearch____add-ons_for_firefox

bobsilverberg:
Fixes #3142 

~~Depends on PR #5303~~

### DIFF
--- a/src/amo/components/CollectionList/index.js
+++ b/src/amo/components/CollectionList/index.js
@@ -27,7 +27,10 @@ import type { ReactRouterLocation } from 'core/types/router';
 
 import './styles.scss';
 
-export type Props = {|
+export type Props = {||};
+
+export type InternalProps = {|
+  ...Props,
   collections: Array<CollectionType> | null,
   currentUsername: string | null,
   dispatch: DispatchFunc,
@@ -38,7 +41,7 @@ export type Props = {|
   location: ReactRouterLocation,
 |};
 
-export class CollectionListBase extends React.Component<Props> {
+export class CollectionListBase extends React.Component<InternalProps> {
   componentDidMount() {
     const {
       collections,
@@ -107,7 +110,7 @@ export class CollectionListBase extends React.Component<Props> {
         <div className="CollectionList-wrapper">
           <Card
             className="CollectionList-info"
-            header={i18n.gettext('About collections')}
+            header={i18n.gettext('Collections')}
           >
             {!isLoggedIn ? (
               <AuthenticateButton
@@ -162,13 +165,15 @@ export const mapStateToProps = (state: {|
   };
 };
 
-export const extractId = (ownProps: Props) => {
+export const extractId = (ownProps: InternalProps) => {
   const { currentUsername } = ownProps;
   return currentUsername || '';
 };
 
-export default compose(
+const CollectionList: React.ComponentType<Props> = compose(
   connect(mapStateToProps),
   translate(),
   withFixedErrorHandler({ fileName: __filename, extractId }),
 )(CollectionListBase);
+
+export default CollectionList;
--- a/src/amo/components/CollectionList/index.js
+++ b/src/amo/components/CollectionList/index.js
@@ -1,10 +1,12 @@
 /* @flow */
+import makeClassName from 'classnames';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import {
-  expandCollections, fetchUserCollections,
+  expandCollections,
+  fetchUserCollections,
 } from 'amo/reducers/collections';
 import { getCurrentUser } from 'amo/reducers/users';
 import AuthenticateButton from 'core/components/AuthenticateButton';
@@ -12,8 +14,10 @@ import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
+import UserCollection from 'ui/components/UserCollection';
 import type {
-  CollectionsState, CollectionType,
+  CollectionsState,
+  CollectionType,
 } from 'amo/reducers/collections';
 import type { UsersStateType } from 'amo/reducers/users';
 import type { ErrorHandlerType } from 'core/errorHandler';
@@ -22,7 +26,6 @@ import type { DispatchFunc } from 'core/types/redux';
 import type { ReactRouterLocation } from 'core/types/router';
 
 import './styles.scss';
-
 
 export type Props = {|
   collections: Array<CollectionType> | null,
@@ -36,10 +39,6 @@ export type Props = {|
 |};
 
 export class CollectionListBase extends React.Component<Props> {
-  static defaultProps = {
-    creating: false,
-  };
-
   componentDidMount() {
     const {
       collections,
@@ -50,97 +49,99 @@ export class CollectionListBase extends React.Component<Props> {
     } = this.props;
 
     if (currentUsername && !loadingUserCollections && !collections) {
-      dispatch(fetchUserCollections({
-        errorHandlerId: errorHandler.id, username: currentUsername,
-      }));
+      dispatch(
+        fetchUserCollections({
+          errorHandlerId: errorHandler.id,
+          username: currentUsername,
+        }),
+      );
     }
   }
 
-  render() {
-    const { i18n, isLoggedIn, location, collections } = this.props;
-    const myCollectionsHeader = i18n.gettext('My Collections');
-
-    if (!isLoggedIn) {
-      return (
-        <AuthenticateButton
-          noIcon
-          location={location}
-          logInText={i18n.gettext('Log in to view your collections')}
-        />
-      );
-    }
+  renderCollections(header: string) {
+    const { i18n, collections } = this.props;
+    const noCollectionsText = i18n.gettext('You do not have any collections.');
 
     const collectionElements = [];
 
-    if (collections && collections.length) {
+    if (collections) {
       collections.forEach((collection) => {
         const { authorUsername, id, name, numberOfAddons, slug } = collection;
-        const numberText = i18n.sprintf(
-          i18n.ngettext(
-            '%(total)s add-on', '%(total)s add-ons', numberOfAddons
-          ),
-          { total: i18n.formatNumber(numberOfAddons) },
-        );
         collectionElements.push(
-          <li className="CollectionList-collection" key={id}>
-            <a
-              className="CollectionList-collection-link"
-              href={`/collections/${authorUsername}/${slug}/`}
-            >
-              <h2 className="CollectionList-collection-name">
-                {name}
-              </h2>
-              <p className="CollectionList-collection-number">
-                {numberText}
-              </p>
-            </a>
-          </li>
+          <UserCollection
+            authorUsername={authorUsername}
+            id={id}
+            name={name}
+            numberOfAddons={numberOfAddons}
+            slug={slug}
+          />,
         );
       });
+    } else {
+      // Create 4 "loading" components.
+      for (let count = 0; count < 4; count++) {
+        collectionElements.push(<UserCollection id={count} loading />);
+      }
     }
+
+    const footer = collectionElements.length ? null : noCollectionsText;
+
+    return (
+      <Card
+        className={makeClassName('CardList', 'CollectionList-list')}
+        footer={footer}
+        header={header}
+      >
+        {collectionElements.length && (
+          <ul className="CollectionList-listing">{collectionElements}</ul>
+        )}
+      </Card>
+    );
+  }
+
+  render() {
+    const { i18n, isLoggedIn, location } = this.props;
+    const myCollectionsHeader = i18n.gettext('My Collections');
 
     return (
       <div className="CollectionList">
         <div className="CollectionList-wrapper">
-          <Card
-            className="CollectionList-info"
-            header={myCollectionsHeader}
-            key="info"
-          >
-            <p className="CollectionList-info-text">
-              {i18n.gettext(`Collections make it easy to keep track of favorite
-                add-ons and share your perfectly customized browser with others.`)}
-            </p>
-            <Button
-              buttonType="action"
-              className="CollectionList-create"
-              puffy
-              to="/collections/add/"
-              type="button"
-            >
-              {i18n.gettext('Create a collection')}
-            </Button>
+          <Card className="CollectionList-info" header={myCollectionsHeader}>
+            {!isLoggedIn ? (
+              <AuthenticateButton
+                noIcon
+                location={location}
+                logInText={i18n.gettext('Log in to view your collections')}
+              />
+            ) : (
+              <React.Fragment>
+                <p className="CollectionList-info-text">
+                  {i18n.gettext(`Collections make it easy to keep track of favorite
+                    add-ons and share your perfectly customized browser with others.`)}
+                </p>
+                <Button
+                  buttonType="action"
+                  className="CollectionList-create"
+                  puffy
+                  to="/collections/add/"
+                  type="button"
+                >
+                  {i18n.gettext('Create a collection')}
+                </Button>
+              </React.Fragment>
+            )}
           </Card>
-          <Card
-            className="CollectionList-list"
-            header={myCollectionsHeader}
-            key="list"
-          >
-            {collectionElements.length ? (
-              <ul className="CollectionList-listing">
-                {collectionElements}
-              </ul>
-            ) : null}
-          </Card>
+          {isLoggedIn ? this.renderCollections(myCollectionsHeader) : null}
         </div>
       </div>
     );
   }
 }
 
-export const mapStateToProps = (
-  state: {| collections: CollectionsState, users: UsersStateType |}
-) => {
+export const mapStateToProps = (state: {|
+  collections: CollectionsState,
+  users: UsersStateType,
+|}) => {
   const { collections, users } = state;
 
   const currentUser = getCurrentUser(users);
@@ -155,10 +156,8 @@ export const mapStateToProps = (
   return {
     currentUsername,
     isLoggedIn: !!currentUser,
-    loadingUserCollections:
-      userCollections ? userCollections.loading : false,
-    collections:
-      expandCollections(collections, userCollections),
+    loadingUserCollections: userCollections ? userCollections.loading : false,
+    collections: expandCollections(collections, userCollections),
   };
 };
 
--- a/src/amo/components/CollectionList/index.js
+++ b/src/amo/components/CollectionList/index.js
@@ -1,5 +1,4 @@
 /* @flow */
-import makeClassName from 'classnames';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -14,6 +13,7 @@ import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
+import CardList from 'ui/components/CardList';
 import UserCollection from 'ui/components/UserCollection';
 import type {
   CollectionsState,
@@ -58,7 +58,7 @@ export class CollectionListBase extends React.Component<Props> {
     }
   }
 
-  renderCollections(header: string) {
+  renderCollections() {
     const { i18n, collections } = this.props;
     const noCollectionsText = i18n.gettext('You do not have any collections.');
 
@@ -87,26 +87,28 @@ export class CollectionListBase extends React.Component<Props> {
     const footer = collectionElements.length ? null : noCollectionsText;
 
     return (
-      <Card
-        className={makeClassName('CardList', 'CollectionList-list')}
+      <CardList
+        className="CollectionList-list"
         footer={footer}
-        header={header}
+        header={i18n.gettext('My collections')}
       >
         {collectionElements.length && (
           <ul className="CollectionList-listing">{collectionElements}</ul>
         )}
-      </Card>
+      </CardList>
     );
   }
 
   render() {
     const { i18n, isLoggedIn, location } = this.props;
-    const myCollectionsHeader = i18n.gettext('My Collections');
 
     return (
       <div className="CollectionList">
         <div className="CollectionList-wrapper">
-          <Card className="CollectionList-info" header={myCollectionsHeader}>
+          <Card
+            className="CollectionList-info"
+            header={i18n.gettext('About collections')}
+          >
             {!isLoggedIn ? (
               <AuthenticateButton
                 noIcon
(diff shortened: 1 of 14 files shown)


bobsilverberg:
Screenshot:

screenshot 2018-06-18 16 01 03

Note that I have not implemented pagination for this, pending a response from @pwalm.

bobsilverberg:
I've rebased this on top of #5303

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+          <Card
+            className="CollectionList-list"
+            header={myCollectionsHeader}
+            key="list"
+          >
+            {collectionElements.length ? (
+              <ul className="CollectionList-listing">
+                {collectionElements}
+              </ul>
+            ) : null}
```
willdurand: Hum, we should probably have some `LoadingText` because it looks a bit broken when waiting for the data to be retrieved:

screen shot 2018-06-22 at 14 10 52

willdurand: I also noticed that we don't have a special "empty state" when the user does not have any collections yet. It shows the same page as above ^

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+              to="/collections/add/"
+              type="button"
+            >
+              {i18n.gettext('Create a collection')}
+            </Button>
+          </Card>
+          <Card
+            className="CollectionList-list"
+            header={myCollectionsHeader}
+            key="list"
```
willdurand: Do you need these keys here and in the previous card?

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+    const myCollectionsHeader = i18n.gettext('My Collections');
+
+    if (!isLoggedIn) {
+      return (
+        <AuthenticateButton
+          noIcon
+          location={location}
+          logInText={i18n.gettext('Log in to view your collections')}
+        />
+      );
```
willdurand: It would be nice to reuse the `Card` on the left to display this button, because it looks a bit weird currently:

screen shot 2018-06-22 at 14 17 52

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,58 @@
...
+  .Card-contents {
+    font-size: $font-size-m-smaller;
+  }
+}
+
+@include respond-to(large) {
+  .CollectionList-wrapper {
+    display: grid;
+    grid-auto-flow: column dense;
+    grid-gap: 0 $padding-page-l;
```
willdurand: On small screens, there is no padding/margin between the two cards:

screen shot 2018-06-22 at 14 19 39

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,58 @@
...
+    margin: 10px 0 0;
+  }
+}
+
+.CollectionList-create {
+  margin-top: 18px;
+  width: 100%;
+}
+
+ul.CollectionList-listing {
```
willdurand: no sure you need to specify the HTML tag here.

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,58 @@
...
+    list-style-type: none;
+  }
+
+  a:link {
+    text-decoration: none;
+
+    .CollectionList-collection-name {
+      color: $black;
+      font-size: $font-size-default;
+      font-weight: 600;
```
willdurand: We probably have a mixin for that, like `@include font-bold()` from https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/mixins.scss

src/amo/routes.js
```diff
@@ -52,6 +53,7 @@ export default (
       <Route path="collections/:username/:slug/" component={Collection} />
       {config.get('enableNewCollectionsUI') && (
         <React.Fragment>
+          <Route path="collections/" component={CollectionList} />
```
willdurand: I think we need to change the dropdown menu to use this link, because currently it uses `/collections/${siteUser.username}/`.

src/amo/reducers/collections.js
```diff
@@ -866,6 +866,15 @@ export const localizeCollectionDetail = (
   };
 };
 
+export const expandCollections = (
```
willdurand: We probably need tests for this function.

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,58 @@
...
+      margin-bottom: 0;
+    }
+  }
+
+  .CollectionList-collection-number {
+    color: $grey-50;
+    font-size: $font-size-m-smaller;
```
willdurand: The `font-weight` should probably be adjusted according to the mock-ups:

screen shot 2018-06-22 at 14 26 43

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+    if (currentUsername && !loadingUserCollections && !collections) {
+      dispatch(fetchUserCollections({
+        errorHandlerId: errorHandler.id, username: currentUsername,
+      }));
+    }
+  }
+
+  render() {
+    const { i18n, isLoggedIn, location, collections } = this.props;
+    const myCollectionsHeader = i18n.gettext('My Collections');
```
willdurand: I know that this complies with the mock-ups but it looks weird to have the same title for the two cards that are displayed next to each other.

bobsilverberg:
FYI, this isn't quite ready for another review yet.

src/amo/components/CollectionList/index.js
willdurand: The minimum would be to name the left card (in LTR mode) "About collections".

src/amo/components/Header/index.js
```diff
@@ -105,7 +105,7 @@ export class HeaderBase extends React.Component {
               <DropdownMenuItem>
                 <Link
                   className="Header-user-menu-collections-link"
-                  href={`/collections/${siteUser.username}/`}
+                  href="/collections/"
```
willdurand: I think the prop should be `to` when we'll enable the collection feature, so maybe you want to do the same as for the user profile:

https://github.com/mozilla/addons-frontend/blob/1e08b34c1b7a62c57e48ee2229bc6eeec9e39b70/src/amo/components/Header/index.js#L68-L70

src/ui/components/UserCollection/index.js
```diff
@@ -0,0 +1,73 @@
...
+    href = `/collections/${authorUsername}/${slug}/`;
+    numberText = i18n.sprintf(
+      i18n.ngettext('%(total)s add-on', '%(total)s add-ons', numberOfAddons),
+      { total: i18n.formatNumber(numberOfAddons) },
+    );
+  }
+
+  return (
+    <li className="CollectionList-collection" key={id}>
+      <a className="CollectionList-collection-link" href={href || '#'}>
```
willdurand: Could we use a `Link` component here (and have `null` instead of `#` as fallback value)?

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,173 @@
...
+                <p className="CollectionList-info-text">
+                  {i18n.gettext(`Collections make it easy to keep track of favorite
+                    add-ons and share your perfectly customized browser with others.`)}
+                </p>
+                <Button
+                  buttonType="action"
+                  className="CollectionList-create"
+                  puffy
+                  to="/collections/add/"
+                  type="button"
```
willdurand: I don't think we should be adding this `type` prop, in part because the style is broken in Chrome:

screen shot 2018-06-25 at 14 22 58

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+  margin-top: 18px;
+  width: 100%;
+}
+
+.CollectionList-listing {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  margin: 0;
+  padding: 0;
+
```
willdurand: I dont really know how to leave this comment but if you had a odd number of collections, the Card is broken:

screen shot 2018-06-25 at 14 25 38

screen shot 2018-06-25 at 14 25 35

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,173 @@
...
+      for (let count = 0; count < 4; count++) {
+        collectionElements.push(<UserCollection id={count} loading />);
+      }
+    }
+
+    const footer = collectionElements.length ? null : noCollectionsText;
+
+    return (
+      <Card
+        className={makeClassName('CardList', 'CollectionList-list')}
```
willdurand: Why not using a `CardList` component here if you pass the `CardList` class?

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+    padding: $padding-page-l;
+  }
+
+  .Card-contents {
+    font-size: $font-size-m-smaller;
+  }
+}
+
+@include respond-to(large) {
+  .CollectionList-wrapper {
```
willdurand: If I remember correctly, we tend to put the mixins inside the class definitions, not the other way around.

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  margin: 0;
+  padding: 0;
+
+  .CollectionList-collection {
+    list-style-type: none;
+  }
+
+  a:link {
```
willdurand: You might not need this if you use a `Link` component.

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+    text-decoration: none;
+
```
willdurand: It would be nice to have some sort of UI feedback when we hover a collection (I mean more than the cursor). Maybe something like the `AddonsCard`?

tests/unit/amo/components/TestCollectionList.js
```diff
@@ -0,0 +1,266 @@
...
+      ...props,
+    };
+
+    return shallowUntilTarget(
+      <CollectionList {...allProps} />,
+      CollectionListBase,
+    );
+  };
+
+  it('dispatches fetchUserCollections as expected', () => {
```
willdurand: I'd prefer something more descriptive than "as expected".

tests/unit/amo/components/TestCollectionList.js
```diff
@@ -0,0 +1,266 @@
...
+  });
+
+  it('renders the collection listing info card', () => {
+    const { store } = dispatchSignInActions();
+    const root = renderComponent({ store });
+
+    expect(root.find('.CollectionList-info')).toHaveLength(1);
+    expect(root.find('.CollectionList-info')).toHaveProp(
+      'header',
+      'My Collections',
```
willdurand: I'd rather expect "About collections"

tests/unit/amo/components/TestCollectionList.js
```diff
@@ -0,0 +1,266 @@
...
+      fetchUserCollections({
+        errorHandlerId: createStubErrorHandler().id,
+        username,
+      }),
+    );
+
+    const root = renderComponent({ store });
+
+    expect(root.find('.CollectionList-list')).toHaveLength(1);
+    expect(root.find('.CollectionList-list')).toHaveProp(
```
willdurand: This was tested before no?

tests/unit/amo/components/TestCollectionList.js
```diff
@@ -0,0 +1,266 @@
...
+      }),
+    );
+
+    const root = renderComponent({ store });
+
+    expect(root.find('.CollectionList-listing')).toHaveLength(1);
+
+    const userCollections = root.find(UserCollection);
+    expect(userCollections).toHaveLength(2);
+    for (let count = 0; count < 2; count++) {
```
willdurand: maybe you could iterate over `collections` instead?

tests/unit/amo/reducers/test_collections.js
```diff
@@ -1074,4 +1075,67 @@ describe(__filename, () => {
...
+      const collections = expandCollections(state, meta);
+      expect(collections[0]).toEqual(
+        createInternalCollection({ detail: firstCollection }),
+      );
+      expect(collections[1]).toEqual(
+        createInternalCollection({ detail: secondCollection }),
+      );
+    });
+
+    it('returns undefined for an unloaded collection', () => {
```
willdurand: Does it actually return `undefined` or does it skip the non-existing collections?

willdurand: I suppose it should skip invalid IDs as well.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -1074,4 +1075,67 @@ describe(__filename, () => {
...
+          username,
+          collections: [firstCollection],
+        }),
+      );
+
+      const collections = expandCollections(state, meta);
+      expect(collections[0]).toEqual(
+        createInternalCollection({ detail: firstCollection }),
+      );
+      expect(collections[1]).toEqual(undefined);
```
willdurand: Is `collections.length === 2`?

tests/unit/ui/components/TestUserCollection.js
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+
+import UserCollection, {
+  UserCollectionBase,
+} from 'ui/components/UserCollection';
+import { fakeI18n, shallowUntilTarget } from 'tests/unit/helpers';
+import LoadingText from 'ui/components/LoadingText';
+
+function render(customProps = {}) {
```
willdurand: We could probably move this function into the top-level `describe` block, what do you think?

tests/unit/ui/components/TestUserCollection.js
```diff
@@ -0,0 +1,78 @@
...
+    };
+
+    const root = render(props);
+
+    expect(root.find('.CollectionList-collection')).toHaveLength(1);
+    expect(root.find('.CollectionList-collection-link')).toHaveProp(
+      'href',
+      `/collections/${props.authorUsername}/${props.slug}/`,
+    );
+    expect(root.find('.CollectionList-collection-name').childAt(0)).toHaveText(
```
willdurand: I think we tend to use `children()` instead of `childAt()` for such assertions.

tests/unit/ui/components/TestUserCollection.js
```diff
@@ -0,0 +1,78 @@
...
+      props.name,
+    );
+    expect(
+      root.find('.CollectionList-collection-number').childAt(0),
+    ).toHaveText(`${props.numberOfAddons} add-ons`);
+  });
+
+  it('renders singluar text for a collection with 1 add-on', () => {
```
willdurand: s/singluar/singular (I guess)

src/ui/components/UserCollection/index.js
```diff
@@ -0,0 +1,73 @@
...
+      </a>
+    </li>
+  );
+};
+
+const UserCollection: React.ComponentType<Props> = compose(translate())(
+  UserCollectionBase,
+);
+
+export default UserCollection;
```
willdurand: I don't find any SCSS file for this component. I think there is a display issue with long collection names:

screen shot 2018-06-25 at 14 25 30

willdurand: In addition, on small screens, would be probably better to have one collection per row.

src/ui/components/UserCollection/index.js
```diff
@@ -0,0 +1,73 @@
...
+    invariant(slug, 'slug is required');
+    invariant(
+      Number.isInteger(numberOfAddons),
+      'numberOfAddons must be a number',
+    );
+
+    href = `/collections/${authorUsername}/${slug}/`;
+    numberText = i18n.sprintf(
+      i18n.ngettext('%(total)s add-on', '%(total)s add-ons', numberOfAddons),
+      { total: i18n.formatNumber(numberOfAddons) },
```
willdurand: For some reasons, RTL langs return a sort of dot when there are 0 add-ons in the collection:

screen shot 2018-06-25 at 14 54 07


I am not sure if it is expected or not.

bobsilverberg: I believe that the dot is just what is output for the language for a 0, but I'm not 100% certain of that. It seems to be consistent though.

tests/unit/amo/reducers/test_collections.js
bobsilverberg: The current behaviour, which I just extracted from `AddAddonToCollection/index.js` returns `undefined` in both cases, but I agree it probably makes more sense for them to be skipped. I'll change it.

tests/unit/amo/reducers/test_collections.js
bobsilverberg: Currently yes, but as per your comment and my response above, I'll change that.

src/ui/components/UserCollection/index.js
willdurand: Yes, I was wondering.. I *think* we can probably leave it as is.

bobsilverberg:
@willdurand I just figured out the CSS issue for odd number of collections. I'll add a commit shortly.

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+  render() {
+    const { i18n, isLoggedIn, location } = this.props;
+
+    return (
+      <div className="CollectionList">
+        <div className="CollectionList-wrapper">
+          <Card
+            className="CollectionList-info"
+            header={i18n.gettext('About collections')}
```
willdurand: Phil said "Collections"

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,174 @@
...
+    collections: expandCollections(collections, userCollections),
+  };
+};
+
+export const extractId = (ownProps: Props) => {
+  const { currentUsername } = ownProps;
+  return currentUsername || '';
+};
+
+export default compose(
```
willdurand: We should use the pattern described here now: https://github.com/mozilla/addons-frontend#flow

src/amo/components/CollectionList/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+.CollectionList-wrapper {
+  @include respond-to(large) {
+    display: grid;
+    grid-auto-flow: column dense;
+    grid-gap: 0 $padding-page-l;
+    grid-template-columns: minmax(300px, 35%) 1fr;
+    margin: 10px 0 0;
+
+    .CollectionList-list.CardList {
```
willdurand: Maybe it should be the `Card-contents` that needs `background: $white`, currently it looks like this:

screen shot 2018-06-26 at 14 40 16
screen shot 2018-06-26 at 14 40 20

src/ui/components/UserCollection/index.js
```diff
@@ -0,0 +1,72 @@
...
+  name?: string,
+  numberOfAddons?: number,
+  slug?: string,
+|};
+
+type InjectedProps = {|
+  i18n: I18nType,
+|};
+
+type InternalProps = { ...Props, ...InjectedProps };
```
willdurand: So the pattern has slightly evolved: we don't need `InjectedProps` anymore, you can spread `Props` into `InternalProps`:

```js
type InternalProps = {|
  ...Props,
  i18n: I18nType,
|};
```

src/ui/components/UserCollection/index.js
```diff
@@ -0,0 +1,72 @@
...
+    linkTo = `/collections/${authorUsername}/${slug}/`;
+    numberText = i18n.sprintf(
+      i18n.ngettext('%(total)s add-on', '%(total)s add-ons', numberOfAddons),
+      { total: i18n.formatNumber(numberOfAddons) },
+    );
+  }
+
+  return (
+    <li className="UserCollection" key={id}>
+      <Link className="UserCollection-link" to={linkTo || null}>
```
willdurand: I think you could initialize `linkTo` to `null` above so that you don't need `|| null` here.

src/ui/components/UserCollection/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  color: $black;
+  flex-grow: 1;
+  font-size: $font-size-default;
+  margin-bottom: 0;
+  width: 100%;
+  word-wrap: break-word;
+
+  .UserCollection-link:focus &,
+  .UserCollection-link:hover & {
```
willdurand: We could maybe add `:active` here.

tests/unit/amo/components/TestHeader.js
```diff
@@ -98,6 +98,7 @@ describe(__filename, () => {
     const link = wrapper.find('.Header-user-menu-collections-link');
 
     expect(link).toHaveLength(1);
+    expect(link.prop('href')).toEqual('/collections/');
```
willdurand: So this test will fail as soon as we'll turn the feature on. Is that expected? Maybe we could have tests with `getFakeConfig()`: https://github.com/mozilla/addons-frontend/blob/d9b067d4a53c25821d6defc2eb75583ac4892f1f/tests/unit/helpers.js#L459

tests/unit/ui/components/TestUserCollection.js
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    const root = render(props);
+
+    expect(root.find('.UserCollection-number').children()).toHaveText(
+      `${props.numberOfAddons} add-on`,
+    );
+  });
+
+  it('renders loading text when no collection props are passed', () => {
```
willdurand: it is rather when "loading" is set to `true`, no?

src/amo/components/CollectionList/index.js
bobsilverberg: Ok, I updated it. It's a bit odd because it doesn't actually have any public `Props`, all props are internal. I still followed the pattern explicitly, which will make it easier to update if we do add a public prop later.

tests/unit/amo/components/TestHeader.js
bobsilverberg: The `Header` component does not currently allow us to inject a `config` object, so we can't use `getFakeConfig()` without changing that. I see that this issue was avoided for the View Profile and Edit Profile links by simply not asserting anything about the href/to property. Because these flags are temporary, maybe it's fine to just do the same for the collections like, rather than going through the trouble of updating the `Header` component and creating tests based on the value of the flag?

Or, if you think we should enhance all of the tests to assert on the link href/to, then maybe that can be a follow-up?

For now (pending your response), I'll just remove that assertion so the test won't fail when the flag changes.

tests/unit/ui/components/TestUserCollection.js
bobsilverberg: Yes, that description was leftover from when I was doing it differently.

src/amo/components/CollectionList/index.js
```diff
@@ -0,0 +1,179 @@
...
+import type { I18nType } from 'core/types/i18n';
+import type { DispatchFunc } from 'core/types/redux';
+import type { ReactRouterLocation } from 'core/types/router';
+
+import './styles.scss';
+
+export type Props = {||};
+
+export type InternalProps = {|
+  ...Props,
```
willdurand: Because it is a "page-level component", I think we could just use `Props` here, no need for an empty type `Props` and `InternalProps`.

willdurand:
> For now (pending your response), I'll just remove that assertion so the test won't fail when the flag changes.

Perfect!

PR #5310, opened 2018-06-18
update sinon to version 6.0.0

bobsilverberg:
Closes #5265 

I had to replace calls to `spy.reset()` with calls to `spy.resetHistory()`, which is a bit odd, because apparently `spy.reset()` was removed in version 5, which we were previously on. Maybe it was just deprecated in version 5, and removed entirely in version 6. In any case this fixes the test failures for me locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -273,7 +273,7 @@
     "sass-loader": "^7.0.1",
     "semver": "^5.3.0",
     "shelljs": "^0.8.0",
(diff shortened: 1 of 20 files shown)


bobsilverberg:
I don't know why this is failing on CI, and the messages are odd, making it look like my changes were not picked up. I'm going to close this and open a new one.

PR #5311, opened 2018-06-18
update sinon to version 6.0.0

bobsilverberg:
Closes #5265

I had to replace calls to spy.reset() with calls to spy.resetHistory(), which is a bit odd, because apparently spy.reset() was removed in version 5, which we were previously on. Maybe it was just deprecated in version 5, and removed entirely in version 6. In any case this fixes the test failures for me locally.


bobsilverberg:
Ok, a rebase fixed it. This is passing locally and on CI, so merging.

PR #5328, opened 2018-06-20
Hide some inputs for users who are not developers

### LINKED ISSUE #4964: Don't show all inputs to users who don't have a public-facing user profile page
> We should not show these to users who don't have a public-facing user profile page (e.g., developers). It's just noise and may encourage them to enter a lot of text (especially the bio) which no one will see. It also gets in the way of settings, like notifications, below.

In other words, the following inputs should be hidden:

- homepage
- location
- occupation
- biography

willdurand:
Fix #4964

---

This PR hides the following inputs on the edit page when user is not a
developer:

- avatar/profile picture
- biography
- homepage
- location
- occupation

We hide such inputs because users who are not developers will not have a
public profile page anymore, hence there is no need for them to provide
these information.


tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -352,17 +394,24 @@ describe(__filename, () => {
       <a href="https://support.mozilla.org/kb/change-primary-email-address-firefox-accounts">Need help?</a></p>`
     );
 
+    expect(root.find('.UserProfileEdit-homepage--help')).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-biography--help')).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-notifications--help')).toHaveLength(0);
```
rebmullin: hmm to me this maybe could be a little clearer.. i mean this help text should be displayed when user is an dev or artist - should that be made cleared by test description? maybe something like ~ `renders help text section for email only if user is not a developer/artist` ?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -865,8 +916,11 @@ describe(__filename, () => {
       },
     });
 
-    // Create a user with another username.
-    const user = createUserAccountResponse({ username: 'willdurand' });
+    // Create a user who is a developer with another username.
+    const user = createUserAccountResponse({
+      is_addon_developer: true,
```
rebmullin: is it overkill to have a separate test for `is_artist` too?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -142,6 +154,31 @@ describe(__filename, () => {
     expect(root.find('.UserProfileEdit-deletion-modal')).toHaveLength(0);
   });
 
+  it('renders user profile page for current logged-in user (non-developer)', () => {
```
rebmullin: maybe to be a little clearer we could state what is different for developer case in description ~ `renders user profile page with limited fields for current logged-in user (non-developer)`  ?

src/amo/components/UserProfileEdit/index.js
```diff
@@ -549,104 +551,102 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
-                className="UserProfileEdit-location"
-                disabled={!user}
-                id="location"
-                name="location"
-                onChange={this.onFieldChange}
-                value={this.state.location}
-              />
+              {isDeveloper(user) && (
+                <React.Fragment>
+                  <label className="UserProfileEdit--label" htmlFor="homepage">
```
rebmullin: is it possible that these fields could already set up by some users? for example even a picture which maybe a regular user might add..
I am just wondering if we should do anything to hide these on the user page too (outside of edit) since now they won't be able to change this with these changes, right? This may be an edge case or not possible tho...do you know?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -352,17 +394,24 @@ describe(__filename, () => {
       <a href="https://support.mozilla.org/kb/change-primary-email-address-firefox-accounts">Need help?</a></p>`
     );
 
+    expect(root.find('.UserProfileEdit-homepage--help')).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-biography--help')).toHaveLength(0);
+    expect(root.find('.UserProfileEdit-notifications--help')).toHaveLength(0);
+  });
+
+  it('renders more help sections when user is a developer', () => {
```
rebmullin: maybe ~ `renders all help text sections when user is a developer`  ?

PR #5329, opened 2018-06-20
Always show the authenticate button

### LINKED ISSUE #5314: "Log in to edit the profile" is not available after the user logs out from Edit profile page
STR:
1. Load AMO dev and log in using an account then go to Edit profile page
2. Log out
3. Observe the page

Expected result:
Log in to edit the profile blue button is available down below the Menu bar.

Actual result:
A grey empty rectangle area is displayed.

Notes:
This issue is reproducible on AMO dev with FF60(Android 8.0).
It is not reproducible on Win10.

no edit button

willdurand:
Fix #5314

---

This PR fixes a CSS mistake that hides the Authenticate button on small screens.

Before:

screen shot 2018-06-20 at 12 01 35

After:

screen shot 2018-06-20 at 12 01 21

### DIFF
--- a/src/amo/components/UserProfileEdit/styles.scss
+++ b/src/amo/components/UserProfileEdit/styles.scss
@@ -13,6 +13,7 @@ $font-size-aside: 10px;
   }
 }
 
+.UserProfileEdit-authenticate,
 .UserProfileEdit-user-links {
   display: none;
 
@@ -35,6 +36,10 @@ $font-size-aside: 10px;
       list-style-type: none;
     }
   }
+}
+
+.UserProfileEdit-authenticate {
+  display: block;
 }
 
 .UserProfileEdit-form-messages .Notice,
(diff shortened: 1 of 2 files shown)


src/amo/components/UserProfileEdit/styles.scss
```diff
@@ -13,6 +13,7 @@ $font-size-aside: 10px;
   }
 }
 
+.UserProfileEdit-authenticate,
 .UserProfileEdit-user-links {
```
rebmullin: do we need this here? below you have the opposite..

rebmullin: not actually you even need to declare `display:block` if you remove this

willdurand: I want to leverage the rest of the CSS for larger screens actually and I force `display: block` for all screen sizes. Does that make sense?

rebmullin: yep got it 👍

PR #5354, opened 2018-06-21
Paginate user extensions and themes

### LINKED ISSUE #4809: Paginate extensions and themes views on user profile so they are all accessible

willdurand:
Fix #4809 

~~⚠️ Depends on #5352~~

---

This PR allows to paginate extensions and/or themes of a given user, on their profile page. Pagination is per card, so that we can paginate extensions and themes independently (two different query parameters).

A good user to try this PR: https://addons-dev.allizom.org/en-US/firefox/user/l4/

~~TODO: more tests~~

### DIFF
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -1,16 +1,22 @@
 /* @flow */
+/* eslint-disable react/no-unused-prop-types */
 import makeClassName from 'classnames';
 import deepEqual from 'deep-eql';
+import invariant from 'invariant';
 import * as React from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
 
 import AddonsCard from 'amo/components/AddonsCard';
+import Link from 'amo/components/Link';
 import {
   fetchAddonsByAuthors,
   getAddonsForUsernames,
+  getCountForAuthorNames,
   getLoadingForAuthorNames,
 } from 'amo/reducers/addonsByAuthors';
+import Paginate from 'core/components/Paginate';
 import {
   ADDON_TYPE_DICT,
   ADDON_TYPE_EXTENSION,
@@ -18,24 +24,33 @@ import {
   ADDON_TYPE_STATIC_THEME,
   ADDON_TYPE_THEME,
   ADDON_TYPE_THEMES,
+  SEARCH_SORT_POPULAR,
 } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
-import type { AddonsByAuthorsState } from 'amo/reducers/addonsByAuthors';
+import type {
+  AddonsByAuthorsState,
+  FetchAddonsByAuthorsParams,
+} from 'amo/reducers/addonsByAuthors';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import type { AddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
 import type { DispatchFunc } from 'core/types/redux';
+import type { ReactRouterLocation } from 'core/types/router';
 
 import './styles.scss';
 
 type Props = {|
   addonType?: string,
+  addons?: Array<AddonType>,
   authorDisplayName: string,
   authorUsernames: Array<string>,
   className?: string,
   forAddonSlug?: string,
   numberOfAddons: number,
+  pageParam: string,
+  paginate: boolean,
+  pathname?: string,
   showMore?: boolean,
 
   // AddonsCard accepts these props which are drilled in.
@@ -45,28 +60,47 @@ type Props = {|
 
 type InjectedProps = {|
   addons?: Array<AddonType>,
+  count: number | null,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
   loading?: boolean,
+  location: ReactRouterLocation,
 |};
 
 type InternalProps = { ...Props, ...InjectedProps };
 
+type DispatchFetchAddonsByAuthorsParams = {|
+  addonType: $PropertyType<FetchAddonsByAuthorsParams, 'addonType'>,
+  authorUsernames: $PropertyType<FetchAddonsByAuthorsParams, 'authorUsernames'>,
+  forAddonSlug: $PropertyType<FetchAddonsByAuthorsParams, 'forAddonSlug'>,
+  page: $PropertyType<FetchAddonsByAuthorsParams, 'page'>,
+|};
+
 export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
   static defaultProps = {
+    pageParam: 'page',
+    paginate: false,
+    showMore: true,
     showSummary: false,
     type: 'horizontal',
-    showMore: true,
   };
 
   componentWillMount() {
-    const { addonType, authorUsernames, forAddonSlug } = this.props;
+    const {
+      addonType,
+      authorUsernames,
+      forAddonSlug,
+      location,
+      pageParam,
+      paginate,
+    } = this.props;
 
     this.dispatchFetchAddonsByAuthors({
       addonType,
       authorUsernames,
       forAddonSlug,
+      page: this.getCurrentPage({ location, paginate, pageParam }),
     });
   }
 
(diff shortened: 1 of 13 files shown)


src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -51,64 +63,123 @@ type Props = {|
...
   }: Props) {
     const {
       addonType: oldAddonType,
       authorUsernames: oldAuthorNames,
       forAddonSlug: oldForAddonSlug,
+      location: oldLocation,
     } = this.props;
 
+    const newPage = paginate ?
+      oldLocation.query[pageParam] !== newLocation.query[pageParam] : false;
```
kumar303: `newLocation` could be undefined, right? I'm not sure how to easily express that in a Flow type since it only applies to `componentWillReceiveProps`.

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -51,64 +63,123 @@ type Props = {|
...
   }
 
-  dispatchFetchAddonsByAuthors({ addonType, authorUsernames, forAddonSlug }: Object) {
+  getCurrentPage({
+    location, paginate, pageParam,
+  }: {|
+    location: ReactRouterLocation,
+    pageParam: string,
+    paginate: boolean,
+  |}): ?number {
```
kumar303: `number | void` might be better

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -51,64 +63,123 @@ type Props = {|
...
+
+    return Number.isNaN(currentPage) || currentPage < 1 ? 1 : currentPage;
+  }
+
+  dispatchFetchAddonsByAuthors({
+    addonType,
+    authorUsernames,
+    forAddonSlug,
+    page,
+  }: Object) {
```
kumar303: Can you add types for these? It may be helpful to use `$PropertyType` on the existing signature type for `fetchAddonsByAuthors`.

src/amo/components/AddonsByAuthorsCard/index.js
willdurand: I don't think it can be undefined since the `withRouter` HOC is used.

src/amo/components/AddonsByAuthorsCard/index.js
willdurand: thanks

src/amo/components/AddonsByAuthorsCard/index.js
kumar303: Really? But when React calls `componentWillReceiveProps()` it does not go through any HOC as far as I know. I'm thinking of a situation where only partial props are updated, which I imagine could look like:

```js
instance.componentWillReceiveProps({ forAddonSlug: 'some-other-addon' })
```

kumar303: I learned that `componentWillReceiveProps()` receives the next `this.props` so it will be a complete object.

src/amo/components/AddonsCard/styles.scss
```diff
@@ -8,12 +8,16 @@
 
     .Card-contents {
       background: $white;
-      border-bottom-left-radius: $border-radius-default;
```
kumar303: Are you sure it's safe to do this? This component is used a lot:

```
$ crg '<AddonsCard' src/
src/amo/components/AddonRecommendations/index.js
166:      <AddonsCard

src/amo/components/AddonsByAuthorsCard/index.js
321:      <AddonsCard

src/amo/components/Collection/index.js
423:            <AddonsCard

src/amo/components/LandingAddonsCard/index.js
53:      <AddonsCard

src/amo/components/SearchResults/index.js
61:        <AddonsCard
```

willdurand:
@kumar303 everything looks OK, even though Codecov reports a coverage drop in `HostPermissions`, which I did not change.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -228,6 +228,35 @@ const reducer = (
...
+          addonType,
+        );
+
+        if (addonsToRemove) {
+          for (const addonToRemove of addonsToRemove) {
+            newState.byAddonId[addonToRemove.id] = undefined;
+
+            if (addonToRemove.authors) {
+              for (const author of addonToRemove.authors) {
+                newState.byUsername[author.username] = newState.byUsername[
```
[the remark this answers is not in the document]
willdurand: The problem is we store two different set of add-ons for a given user: extensions and themes. If I reset the state by applying your patch, I clear the list of extensions the second after I navigate to page 2 of the themes.

Your patch is what I initially wrote, but then I tried to play with pagination and boom... I was always resetting too much data.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -228,6 +228,35 @@ const reducer = (
...
+
+        const addonsToRemove = getAddonsForUsernames(
+          newState,
+          authorUsernames,
+          addonType,
+        );
+
+        if (addonsToRemove) {
+          for (const addonToRemove of addonsToRemove) {
+            newState.byAddonId[addonToRemove.id] = undefined;
```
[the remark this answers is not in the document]
willdurand: Right, I think that should work. It was mostly to clean things correctly, but I guess we can play safe and never remove anything from `byAddonId`.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -81,15 +85,20 @@ describe(__filename, () => {
...
+    multipleAuthors = false,
+    count = null,
+  }) {
     const pageSize =
       addonType === ADDON_TYPE_THEME
         ? THEMES_BY_AUTHORS_PAGE_SIZE
         : EXTENSIONS_BY_AUTHORS_PAGE_SIZE;
 
     const addons = [];
+    const nbAddons = count || pageSize;
```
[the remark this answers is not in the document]
willdurand: sure, that would make sense. I wish I could use `numberOfAddons` but that was already in-use...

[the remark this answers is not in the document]
willdurand: ahm I guess that would make more sense indeed.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -464,21 +464,41 @@ describe(__filename, () => {
...
 
     expect(root.find(AddonsByAuthorsCard).at(1)).toHaveProp(
       'addonType',
       ADDON_TYPE_THEME,
     );
+    expect(root.find(AddonsByAuthorsCard).at(1)).toHaveProp(
+      'pageParam',
+      'page_t',
+    );
+    expect(root.find(AddonsByAuthorsCard).at(1)).toHaveProp('paginate', true);
```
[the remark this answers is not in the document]
willdurand: How would you ensure that the seecond `AddonsByAuthorsCard` will have pagination enabled?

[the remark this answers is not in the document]
kumar303: Oh, whoops, I missed the `.at(1)` part of `root.find(AddonsByAuthorsCard).at(1)`. I read `0` and `1` as the same. Carry on 😁

src/amo/reducers/addonsByAuthors.js
```diff
@@ -228,6 +228,33 @@ const reducer = (
...
+          newState,
+          authorUsernames,
+          addonType,
+        );
+
+        if (addonsToRemove) {
+          for (const addonToRemove of addonsToRemove) {
+            if (addonToRemove.authors) {
+              for (const author of addonToRemove.authors) {
+                newState.byUsername[author.username] = newState.byUsername[
```
kumar303: Thanks for the explanation. I understand that `addonType` is important now. The comment up above didn't help me understand for some reason. When I re-read it, I think the comment does explain it.

src/amo/reducers/addonsByAuthors.js
```diff
@@ -228,6 +228,33 @@ const reducer = (
           ...newState.byAddonSlug,
           [forAddonSlug]: undefined,
         };
+      } else if (authorUsernames.length) {
```
kumar303: It will be surprising that this reducer behaves differently in the presence of `forAddonSlug`. Is it not possible to do it like this?

```diff
diff --git a/src/amo/reducers/addonsByAuthors.js b/src/amo/reducers/addonsByAuthors.js
index bc184e7db..2cd024516 100644
--- a/src/amo/reducers/addonsByAuthors.js
+++ b/src/amo/reducers/addonsByAuthors.js
@@ -228,7 +228,9 @@ const reducer = (
           ...newState.byAddonSlug,
           [forAddonSlug]: undefined,
         };
-      } else if (authorUsernames.length) {
+      }
+
+      if (authorUsernames.length) {
         // Potentially remove add-ons loaded for these authors with this add-on
         // type, so that we can load new add-ons in the UI (pagination). We do
         // not do it when `forAddonSlug` is defined to avoid re-fetching the
```

I know a test fails but I'm looking at `Addon` and I'm not sure if it matters or not. If it *does* matter then that tells me we need a new action instead of re-using the `FetchAddonsByAuthor` action. 

However, I don't think it's worth holding up your patch for such a big change. It would be fine to land it as is. 

I just wanted to point out that this could potentially create confusion in the future if someone were not intimately familiar with the implementation of this action/reducer logic.

src/amo/components/AddonsCard/styles.scss
willdurand: Well, I thought that would be ok but after a closer look, that was a bit silly. I have a better change for what I wanted to achieve here.

src/amo/components/AddonsByAuthorsCard/styles.scss
```diff
@@ -34,6 +34,13 @@
       }
     }
 
+    &:not(.Card--no-footer) {
```
kumar303: 👍

src/amo/reducers/addonsByAuthors.js
willdurand: No, I don't think it matters. The only reason I've done that is because I did not want to reset the state for the "add-ons by authors" on the add-on detail page to avoid re-fetching those add-ons. Apart from that, there is no reason not to do it.

kumar303: > The only reason I've done that is because I did not want to reset the state for the "add-ons by authors" on the add-on detail page to avoid re-fetching those add-ons.

That sounds like a premature optimization and one that might be better fixed a different way (if it ever becomes a problem).

willdurand: Right, I don't really see this as a premature optimization because it was the old behavior but I agree that we can fix it in a different way if needed.

Issue #5342, opened 2018-06-20
Add Flow types to prevent misuse of component props, part 2 (hard)


kumar303:
This is part 2 (the hard part) to preventing Flow typed code from calling a component with invalid props. For details, see: https://github.com/mozilla/addons-frontend/issues/3563

This second step fixes components that generated 3 or more errors. If the diffs become too big I'll spit it into further patches for easier reviewing.

- ui/components/HostPermissions (7 errors)
- ui/components/Hero (20 errors)
- ui/components/DismissibleTextForm (9 errors)
- ui/components/ConfirmButton (10 errors)
- amo/components/AddonReview (16 errors)
- amo/components/AddonReviewListItem (19 errors)
- amo/components/AddonsByAuthorsCard (8 errors)
- amo/components/AutoSearchInput (16 errors)
- amo/components/Categories (5 errors)
- amo/components/CollectionManager (10 errors)
- amo/components/EditableCollectionAddon (6 errors)
- amo/components/ErrorPage/NotFound (5 errors)
- amo/components/FlagReview (9 errors)
- amo/components/RatingManager (11 errors)
- amo/components/ReportAbuseButton (5 errors)
- amo/components/ReportUserAbuse (6 errors)
- amo/components/SearchResult (3 errors)

PR #5355, opened 2018-06-21
Flow types to prevent misuse of component props, part

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5342

This revealed one bug in how an array was passed instead of a string: https://github.com/mozilla/addons-frontend/blob/473e0936083d3a991c54be2ee45dc9ae228b540f/src/amo/components/UserProfile/index.js#L252


src/amo/components/AddonReview/index.js
```diff
@@ -34,25 +37,34 @@ type RefreshAddonFunction = (
...
+|};
+
+type InternalProps = { ...Props, ...InjectedProps };
+
 type State = {|
   reviewBody: ?string,
 |};
 
-export class AddonReviewBase extends React.Component<Props, State> {
+export class AddonReviewBase extends React.Component<InternalProps, State> {
```
willdurand: Why is this type `Internal`? I am not sure to understand. I see that we often use `CompBase` as class name, so maybe it could be named `BaseProps`. It does not really matter though.

src/amo/components/AddonReview/index.js
```diff
@@ -61,16 +73,17 @@ export class AddonReviewBase extends React.Component<Props, State> {
...
     this.state = {
       reviewBody: props.review.body,
     };
+    invariant(props.createLocalState, 'props.createLocalState() was undefined');
     this.localState = props.createLocalState(`AddonReview:${props.review.id}`);
     this.checkForStoredState();
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  componentWillReceiveProps(nextProps: InternalProps) {
```
willdurand: Or maybe `Props` should be `BaseProps` and we could use `Props` for the set of all props.

src/amo/components/AddonsCard/index.js
```diff
@@ -78,6 +79,9 @@ export default class AddonsCard extends React.Component<Props> {
         // any summary or description.
         const isTheme = ADDON_TYPE_THEMES.includes(addon.type);
         if (editing) {
+          invariant(deleteNote, 'props.deleteNote was undefined');
```
willdurand: We use the present tense in most (all?) of our `invariant` calls.

src/amo/components/EditableCollectionAddon/index.js
```diff
@@ -122,12 +127,12 @@ export class EditableCollectionAddonBase extends React.Component<Props, State> {
                 <DismissibleTextForm
                   className="EditableCollectionAddon-notes-form"
                   microButtons
-                  onDelete={addon.notes && this.onDeleteNote}
+                  onDelete={addon.notes ? this.onDeleteNote : null}
```
willdurand: Thank you.

src/amo/components/RatingManager/index.js
```diff
@@ -158,6 +168,7 @@ export class RatingManagerBase extends React.Component<Props, State> {
 
   renderLogInToRate() {
     const { AuthenticateButton, addon, location } = this.props;
+    invariant(AuthenticateButton, 'props.AuthenticateButton was undefined');
```
willdurand: Isn't it useless since `AuthenticateButton` is declared in the `defaultProps`?

src/amo/components/EditableCollectionAddon/index.js
kumar303: Flow demanded it 😄

kumar303:
In response to your two comments about `defaultProps`, I think there is a bug in Flow where it doesn't know that a prop gets a default value. It thinks the prop can be undefined when in practice it can't be since the `defaultProps` value will be used. All I could think to do was to make each prop optional and then add defensive code with `invariant()`.

kumar303:
Aha, the bug is because Flow can't follow the object spread between `Props` and `InternalProps`. [Example](https://flow.org/try/#0JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQdMAnmFnAEJpsC8cA3nThwAFsgB2AEwA2WAMKixAcywAuOAAoAbsikBXVXDG6QAIyxQAlHG4A+OJojAJdAL71GLNgAViYdLwEhAHoguAB1YSwxOCYIXThdVDZkOFQwSmQJABo4ADEpCAB3OFxxMQhCYEVdSjg4+BhIkogxMSx8YBa4GAg4CSwCZF0pGB8IPwA6QTgJ2c4k13dcKTR0AGUpJ3M4LAAPGCiJdGw8GAnZEkg2sRgAHjG-HIQ7QNSYZBhgXD6BoZGH-z8aZCBTSOQKZRqdRWWz8FzTNwMISoEhYRrAJTQoFCITo1ATdLjfGgmTycTKdQAJgs9CE8PhdFuGy2UEMyBAWG4ZBRHPRSjIQRsQA). I'll see if it's reported yet.

kumar303:
It looks like an obscure bug that only applies to function types: https://github.com/facebook/flow/issues/6502

src/amo/components/AddonReview/index.js
kumar303: They are internal because only the component will see things like the mapped props. No calling code will need to pass in these props.

src/amo/components/AddonReview/index.js
kumar303: I think `Props` is nice because those are the actual component props that any calling code will need to care about. We could make it explicit like `ExternalProps` but, I dunno, I think it's implied.

willdurand: Alright, I am fine with that.

src/amo/components/CollectionManager/index.js
```diff
@@ -115,6 +121,7 @@ export class CollectionManagerBase extends React.Component<Props, State> {
 
   componentWillUnmount() {
     if (this.timeout) {
+      invariant(this.props.clearTimeout, 'props.clearTimeout is undefined');
```
willdurand: I see that sometimes you add parentheses, and sometimes not. I think we should keep parentheses to indicate functions, but I'd get rid of the `prop.` prefix because when we get the invariant error, the file and line are indicated.

PR #5358, opened 2018-06-21
I5078 include statictheme in search filter if enabled

### LINKED ISSUE #5078: Combined lightweight and static theme results when filtering by theme in search
when you search for an addon and then filter by Theme, static themes are not included in the results. 

static themes will be included in the Theme filter results

also make sure come up on the home page, any landing pages and in its respective category

rebmullin:
fixes #5078

### DIFF
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -24,7 +24,11 @@ import {
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
-import { apiAddonType, apiAddonTypeIsValid, getAddonTypeFilter } from 'core/utils';
+import {
+  apiAddonType,
+  apiAddonTypeIsValid,
+  getAddonTypeFilter,
+} from 'core/utils';
 
 import './styles.scss';
 
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -24,7 +24,7 @@ import {
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
-import { apiAddonType, apiAddonTypeIsValid } from 'core/utils';
+import { apiAddonType, apiAddonTypeIsValid, getAddonTypeFilter } from 'core/utils';
 
 import './styles.scss';
 
@@ -123,6 +123,7 @@ export class CategoryBase extends React.Component {
 
   contentForType = (addonType) => {
     const { i18n, params } = this.props;
+    const themeFilter = getAddonTypeFilter(ADDON_TYPE_THEME);
 
     const contentForTypes = {
       [ADDON_TYPE_EXTENSION]: {
@@ -164,7 +165,7 @@ export class CategoryBase extends React.Component {
         featuredFooterLink: {
           pathname: '/search/',
           query: {
-            addonType: ADDON_TYPE_THEME,
+            addonType: themeFilter,
             category: params.slug,
             featured: true,
           },
@@ -174,7 +175,7 @@ export class CategoryBase extends React.Component {
         trendingFooterLink: {
           pathname: '/search/',
           query: {
-            addonType: ADDON_TYPE_THEME,
+            addonType: themeFilter,
             category: params.slug,
             sort: SEARCH_SORT_TRENDING,
           },
@@ -184,7 +185,7 @@ export class CategoryBase extends React.Component {
         highlyRatedFooterLink: {
           pathname: '/search/',
           query: {
-            addonType: ADDON_TYPE_THEME,
+            addonType: themeFilter,
             category: params.slug,
             sort: SEARCH_SORT_TOP_RATED,
           },
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -133,10 +133,9 @@ export class CategoryBase extends React.Component {
 
   contentForType = (addonType) => {
     const { i18n, params } = this.props;
-    const themeFilter = getAddonTypeFilter(
-      ADDON_TYPE_THEME,
-      this.props._config,
-    );
+    const themeFilter = getAddonTypeFilter(ADDON_TYPE_THEME, {
+      _config: this.props._config,
+    });
 
     const contentForTypes = {
       [ADDON_TYPE_EXTENSION]: {
--- a/src/amo/components/Category/index.js
+++ b/src/amo/components/Category/index.js
@@ -1,5 +1,6 @@
 /* eslint-disable react/no-unused-prop-types */
 import PropTypes from 'prop-types';
+import config from 'config';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -34,6 +35,7 @@ import './styles.scss';
 
 export class CategoryBase extends React.Component {
   static propTypes = {
+    _config: PropTypes.object,
     addonTypeOfResults: PropTypes.string,
     categoryOfResults: PropTypes.string,
     categories: PropTypes.object,
@@ -50,6 +52,10 @@ export class CategoryBase extends React.Component {
     }).isRequired,
     trendingAddons: PropTypes.array,
     resultsLoaded: PropTypes.bool.isRequired,
+  };
+
+  static defaultProps = {
+    _config: config,
   };
 
   componentWillMount() {
@@ -127,7 +133,10 @@ export class CategoryBase extends React.Component {
 
   contentForType = (addonType) => {
     const { i18n, params } = this.props;
-    const themeFilter = getAddonTypeFilter(ADDON_TYPE_THEME);
+    const themeFilter = getAddonTypeFilter(
+      ADDON_TYPE_THEME,
+      this.props._config,
+    );
 
     const contentForTypes = {
       [ADDON_TYPE_EXTENSION]: {
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -328,7 +328,7 @@ export const decodeHtmlEntities = (string) => {
   return entities.decode(string);
 };
 
-export const getAddonTypeFilter = (addonType, _config = config) => {
+export const getAddonTypeFilter = (addonType, { _config = config }) => {
   const isTheme = ADDON_TYPE_THEMES.includes(addonType);
   if (!_config.get('enableStaticThemes') || !isTheme) {
     return addonType;
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -328,7 +328,7 @@ export const decodeHtmlEntities = (string) => {
   return entities.decode(string);
 };
 
-export const getAddonTypeFilter = (addonType, { _config = config }) => {
+export const getAddonTypeFilter = (addonType, { _config = config } = {}) => {
   const isTheme = ADDON_TYPE_THEMES.includes(addonType);
   if (!_config.get('enableStaticThemes') || !isTheme) {
     return addonType;
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -453,32 +453,20 @@ describe(__filename, () => {
 
     const landingShelves = root.find(LandingAddonsCard);
 
-    expect(landingShelves.at(0)).toHaveProp('footerLink', {
-      pathname: `/search/`,
-      query: {
-        addonType: ADDON_TYPE_THEMES_FILTER,
-        category: fakeCategory.slug,
-        featured: true,
-      },
-    });
+    expect(landingShelves.at(0)).toHaveProp('footerLink');
+    expect(landingShelves.at(0).props().footerLink.query.addonType).toEqual(
+      ADDON_TYPE_THEMES_FILTER,
+    );
 
-    expect(landingShelves.at(1)).toHaveProp('footerLink', {
-      pathname: '/search/',
-      query: {
-        addonType: ADDON_TYPE_THEMES_FILTER,
-        category: fakeCategory.slug,
-        sort: SEARCH_SORT_TOP_RATED,
-      },
-    });
+    expect(landingShelves.at(1)).toHaveProp('footerLink');
+    expect(landingShelves.at(1).props().footerLink.query.addonType).toEqual(
+      ADDON_TYPE_THEMES_FILTER,
+    );
 
-    expect(landingShelves.at(2)).toHaveProp('footerLink', {
-      pathname: '/search/',
-      query: {
-        addonType: ADDON_TYPE_THEMES_FILTER,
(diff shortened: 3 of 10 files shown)


tests/unit/amo/components/TestCategory.js
```diff
@@ -436,6 +438,86 @@ describe(__filename, () => {
...
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+
+    const root = render({ _config: fakeConfig }, { autoDispatchCategories: false });
+
+    const landingShelves = root.find(LandingAddonsCard);
+
+    expect(landingShelves.at(0)).toHaveProp('footerLink', {
+      pathname: `/search/`,
+      query: {
+        addonType: ADDON_TYPE_THEMES_FILTER,
```
rebmullin: Ideally we would have something like the above but for some reason these tests are failing when I pass in this fakeConfig. It seems enableStaticTheme still comes up as false. Any idea what i am missing?

rebmullin:
@willdurand @kumar303  - can either of you please take a look? FYI, I had a little trouble with testing the fakeConfig :/ it wasn't working as expected so i may be missing something..? If you have any ideas, please let me know.. thanks : )

tests/unit/amo/components/TestCategory.js
willdurand: I think the problem is that the `Category` component does not receive a `config` object, hence it cannot be passed to the `getAddonTypeFilter()` function.

tests/unit/amo/components/TestCategory.js
```diff
@@ -436,6 +438,92 @@ describe(__filename, () => {
...
+    });
+  });
+
+  it('renders the themes shelves with the ADDON_TYPE_THEME filter if static theme is disabled', () => {
+    _categoriesFetch();
+    _categoriesLoad();
+    _getLanding();
+    _loadLanding();
+
+    const fakeConfig = getFakeConfig({ enableStaticThemes: false });
```
willdurand: This works only because it is `false` by default, but not because you have a specific config.

rebmullin: right, i just thought it was better to be explicit since this is the what the test describes, no?

willdurand: What I mean is that you could set `enableStaticThemes` to `true`, the test would not fail because as far as I know, `config` is not a prop of `Category`.

tests/unit/amo/components/TestCategory.js
rebmullin: yep, I already tried it that way too and it does the same. I mean if you do it outside of here, it returns what you expect (`ADDON_TYPE_THEMES_FILTER`) value but in the query it still returns the value `ADDON_TYPE_THEME`:

Alt text

willdurand: I am not sure to follow. If `Category` accepts a `_config` prop, defaulting to `config`, and `geAddonTypFilter()` is called with `this.props._config` as a second argument, I don't understand why it would not work (if `_config` is passed to the `render` function in the test case).

rebmullin: you're right! I was passing in a 2nd arg here (at 1 point) but not as a prop so it didn't work this way. thank you 👍

tests/unit/amo/components/TestCategory.js
rebmullin: yeh i get it now 🤦‍♂️  thanks:)

rebmullin:
thanks @kumar303, sure whomever can help me get this in ideally for tomorrow's tag that would be great // cc @willdurand  : )

tests/unit/amo/components/TestCategory.js
```diff
@@ -436,6 +438,92 @@ describe(__filename, () => {
...
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+
+    const root = render(
+      { _config: fakeConfig },
+      { autoDispatchCategories: false },
+    );
+
+    const landingShelves = root.find(LandingAddonsCard);
+
+    expect(landingShelves.at(0)).toHaveProp('footerLink', {
```
kumar303: It looks like this test is only necessary to cover the `addonType` logic, right? If so, you could simplify it like this:

```diff
diff --git a/tests/unit/amo/components/TestCategory.js b/tests/unit/amo/components/TestCategory.js
index 1c92fc4de..b674d778f 100644
--- a/tests/unit/amo/components/TestCategory.js
+++ b/tests/unit/amo/components/TestCategory.js
@@ -453,14 +453,10 @@ describe(__filename, () => {
 
     const landingShelves = root.find(LandingAddonsCard);
 
-    expect(landingShelves.at(0)).toHaveProp('footerLink', {
-      pathname: `/search/`,
-      query: {
-        addonType: ADDON_TYPE_THEMES_FILTER,
-        category: fakeCategory.slug,
-        featured: true,
-      },
-    });
+    expect(landingShelves.at(0)).toHaveProp('footerLink');
+    expect(landingShelves.at(0).props().footerLink.query.addonType).toEqual(
+      ADDON_TYPE_THEMES_FILTER,
+    );
 
     expect(landingShelves.at(1)).toHaveProp('footerLink', {
       pathname: '/search/',
```
It's not necessary to test the other parts of `query` because [this test](https://github.com/rebmullin/addons-frontend/blob/9080651ffe3f3bac00975457c6fb23552daf65b1/tests/unit/amo/components/TestCategory.js#L374) already checks them. It's good to remove all unnecessary assertions from a test because it will make failures easier to track down (there will be fewer failing tests).

The other tests can also be simplified in this same way.

tests/unit/amo/components/TestLandingPage.js
```diff
@@ -286,6 +288,88 @@ describe(__filename, () => {
     });
   });
 
+  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+
+    store.dispatch(
```
kumar303: I see a dispatch helper already in this test file. Can you use it like this?

```js
_getAndLoadLandingAddons({ addonType: ADDON_TYPE_THEME });
```

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +250,20 @@ describe(__filename, () => {
 
     expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
   });
+
+  it('sets themes filters shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+    const root = render({ _config: fakeConfig });
+    expect(root.instance().addonTypeOptions()[3].value).toEqual(
```
kumar303: This test is brittle because it relies on `addonTypeOptions()` which is just an implementation detail of the component. Instead, it should look at the `<Select \>` and `<option />` components in the rendered output using standard enzyme selectors. You may need a map function to find the value you want but it should be possible.

src/amo/components/Category/index.js
```diff
@@ -123,6 +133,10 @@ export class CategoryBase extends React.Component {
 
   contentForType = (addonType) => {
     const { i18n, params } = this.props;
+    const themeFilter = getAddonTypeFilter(
```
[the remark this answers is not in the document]
rebmullin: yeh, I really like this : ) I will update and try to use this moving forward where I can

tests/unit/amo/components/TestCategory.js
rebmullin: ok i see... i will update : )

tests/unit/amo/components/TestSearchFilters.js
rebmullin: I see... ok let me know how i have it now  works.

rebmullin:
thanks @kumar303 @willdurand for all the feedback : ). I've made some changes. When you can please take another look.

src/amo/components/Category/index.js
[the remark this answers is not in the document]
rebmullin: qq tho - how come we don't use flow here? I've been meaning to ask as I see it in some places but not all.. just curious if that's intentional or there is reason?

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +251,34 @@ describe(__filename, () => {
 
     expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
   });
+
+  it('sets themes filters shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+    const root = render({ _config: fakeConfig });
+    const selectOption = root.find(Select);
+
+    selectOption
```
kumar303: This approach looks better but there's a problem in that the test won't ever fail. This is because the `expect()` call happen in an if branch that may not be reached.

You can do it a couple different ways. Here's one idea:

```js
const optionValues = selectOption.at(1).props().children.map(
  (option) => option.props.value
);
expect(optionValues).toContain(ADDON_TYPE_THEMES_FILTER);
```

There is probably a better way. Also, try to use the [children](http://airbnb.io/enzyme/docs/api/ShallowWrapper/children.html) functions instead because it will make the code more portable among React versions.

tests/unit/amo/containers/TestHome.js
```diff
@@ -164,6 +166,36 @@ describe(__filename, () => {
     expect(shelf).toHaveProp('loading', true);
   });
 
+  it('renders a featured themes shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+    const root = render({ _config: fakeConfig });
+
+    const shelves = root.find(LandingAddonsCard);
+    const shelf = shelves.find('.Home-FeaturedThemes');
+    expect(shelf).toHaveProp('footerLink', {
```
kumar303: This assertion should also be updated to only look at `query`, right? I'm not sure if the other parts are relevant.

tests/unit/amo/containers/TestHome.js
```diff
@@ -164,6 +166,36 @@ describe(__filename, () => {
...
+    });
+  });
+
+  it('renders a featured themes shelf with the ADDON_TYPE_THEME filter if static theme is disabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: false });
+    const root = render({ _config: fakeConfig });
+
+    const shelves = root.find(LandingAddonsCard);
+    const shelf = shelves.find('.Home-FeaturedThemes');
+    expect(shelf).toHaveProp('footerLink', {
```
kumar303: Same comment here

src/core/utils/index.js
```diff
@@ -328,7 +328,7 @@ export const decodeHtmlEntities = (string) => {
   return entities.decode(string);
 };
 
-export const getAddonTypeFilter = (addonType, _config = config) => {
+export const getAddonTypeFilter = (addonType, { _config = config }) => {
```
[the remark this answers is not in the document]
rebmullin: oh yeh i see... thanks

tests/unit/amo/containers/TestHome.js
rebmullin: yep, you're right. i'll update

rebmullin:
hi @kumar303, thanks so much for the quick review. How's it look now?

rebmullin:
actually hold up...i have to fix 1 more test :/

rebmullin:
Hey @kumar303, I think it's ready now for re-review : )

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +251,28 @@ describe(__filename, () => {
...
+    expect(optionValues).toContain(ADDON_TYPE_THEMES_FILTER);
+  });
+
+  it('sets themes filters shelf with the ADDON_TYPE_THEME filter if static theme is disabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: false });
+    const root = render({ _config: fakeConfig });
+    const selectFilters = root.find(Select);
+
+    const optionValues = selectFilters
+      .find('.SearchFilters-AddonType')
```
kumar303: This works but you can just get it by ID:

```js
const optionValues = root.find('#SearchFilters-AddonType')
  .children()
  .map((option) => option.props().value);
```

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +251,28 @@ describe(__filename, () => {
...
     expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
   });
+
+  it('sets themes filters shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+    const root = render({ _config: fakeConfig });
+    const selectFilters = root.find(Select);
+
+    const optionValues = selectFilters
+      .find('.SearchFilters-AddonType')
```
kumar303: Same thing here

tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -244,4 +251,28 @@ describe(__filename, () => {
...
     expect(root.find('.SearchFilters-AddonType')).toHaveLength(0);
   });
+
+  it('sets themes filters shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+    const fakeConfig = getFakeConfig({ enableStaticThemes: true });
+    const root = render({ _config: fakeConfig });
+    const selectFilters = root.find(Select);
+
+    const optionValues = selectFilters
+      .find('#SearchFilters-AddonType')
```
kumar303: Oh, I only suggested this because it would allow you to do the find without having to initialize it with `Select`:

```diff
diff --git a/tests/unit/amo/components/TestSearchFilters.js b/tests/unit/amo/components/TestSearchFilters.js
index 635946348..c04db3627 100644
--- a/tests/unit/amo/components/TestSearchFilters.js
+++ b/tests/unit/amo/components/TestSearchFilters.js
@@ -255,10 +255,9 @@ describe(__filename, () => {
   it('sets themes filters shelf with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
     const fakeConfig = getFakeConfig({ enableStaticThemes: true });
     const root = render({ _config: fakeConfig });
-    const selectFilters = root.find(Select);
 
-    const optionValues = selectFilters
-      .find('.SearchFilters-AddonType')
+    const optionValues = root
+      .find('#SearchFilters-AddonType')
       .children()
       .map((option) => option.props().value);
     expect(optionValues).toContain(ADDON_TYPE_THEMES_FILTER);
```

Again, it's not a big deal and I know we are tagging soon 😀

PR #5373, opened 2018-06-22
Show "My reviews" on user profile page

### LINKED ISSUE #4282: Show "My reviews" on user profile page
The UX mocks have a "my reviews" section on the profile page. This is largely useful for your own profile as we won't be showing non-developer pages (which is, I'd think, the main place one would look for reviews).

screenshot 2018-01-31 14 54 16

willdurand:
Fix #4282

~~:warning: depends on #5354~~

---

- [x] basic feature
- [x] write tests
- [x] show only for current user


src/amo/components/UserProfile/index.js
```diff
@@ -27,35 +33,55 @@ import {
...
+  canEditProfile: boolean,
   currentUser: UserType | null,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
-  canEditProfile: boolean,
   i18n: I18nType,
   isOwner: boolean,
+  location: ReactRouterLocation,
   params: {| username: string |},
+  reviewCount: number | null,
```
kumar303: This might be a good time to move mapped props to `InternalProps`.

src/amo/components/UserProfile/index.js
```diff
@@ -103,7 +160,71 @@ export class UserProfileBase extends React.Component<Props> {
...
+      >
+        <ul>
+          {reviews.map((review) => {
+            const reviewBodySanitized = sanitizeHTML(nl2br(review.body), [
+              'br',
+            ]);
+
+            return (
+              <li key={String(review.id)}>
+                <div className="AddonReviewListItem">
```
kumar303: This should move to a `UserReview` component and should include any styles that were defined in `AddonReviewListItem`.

Ideally, [`AddonReviewListItem`](https://github.com/mozilla/addons-frontend/blob/ca492751c7dd4482ecefdccd5c1af779953e358f/src/amo/components/AddonReviewListItem/index.js#L234) should be refactored to use `UserReview` but that could come in a separate issue since it won't be a small task.

src/amo/reducers/reviews.js
```diff
@@ -176,6 +186,21 @@ export const changeViewState = ({
...
+  state: ReviewState,
+  userId: number,
+): {|
+  reviewCount: number,
+  reviews: Array<UserReviewType>,
+|} | null => {
+  return state.byUserId[userId]
+    ? {
+        reviewCount: state.byUserId[userId].reviewCount,
+        reviews: state.byUserId[userId].reviews.map((id) => state.byId[id]),
```
kumar303: This should be `expandReviewObjects({ state, reviews: state.byUserId[userId].reviews })`

src/amo/sagas/reviews.js
```diff
@@ -51,6 +54,35 @@ function* fetchReviews({
...
 
+function* fetchUserReviews({
+  payload: { errorHandlerId, page, userId },
+}: FetchUserReviewsAction): Generator<any, any, any> {
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  try {
+    const state = yield select(getState);
+
+    const response = yield call(getReviews, {
```
kumar303: Huh. It looks like this saga never got Flow types. That's a bummer because when I added them, I see some subtle bugs in all of these API calls 👻 :

```diff
diff --git a/src/amo/api/reviews.js b/src/amo/api/reviews.js
index e46fb8e97..003d6a5c2 100644
--- a/src/amo/api/reviews.js
+++ b/src/amo/api/reviews.js
@@ -135,7 +135,7 @@ export const replyToReview = ({
   });
 };
 
-type GetReviewsParams = {|
+export type GetReviewsParams = {|
   // This is the addon ID, slug, or guid.
   addon?: number | string,
   apiState?: ApiStateType,
diff --git a/src/amo/sagas/reviews.js b/src/amo/sagas/reviews.js
index 1094e2281..7a7ebe55c 100644
--- a/src/amo/sagas/reviews.js
+++ b/src/amo/sagas/reviews.js
@@ -21,6 +21,7 @@ import {
 } from 'amo/constants';
 import log from 'core/logger';
 import { createErrorHandler, getState } from 'core/sagas/utils';
+import type { GetReviewsParams } from 'amo/api/reviews';
 import type {
   FetchReviewsAction,
   FetchUserReviewsAction,
@@ -62,13 +63,14 @@ function* fetchUserReviews({
   try {
     const state = yield select(getState);
 
-    const response = yield call(getReviews, {
+    const params: GetReviewsParams = {
       api: state.api,
       // Hide star-only ratings (reviews that do not have a body).
       filter: 'without_empty_body',
       page,
       user: userId,
-    });
+    };
+    const response = yield call(getReviews, params);
 
     yield put(
       setUserReviews({
```

src/amo/components/UserProfile/index.js
```diff
@@ -307,10 +431,14 @@ export function mapStateToProps(
...
 
+  const reviews = user ? getReviewsByUserId(state.reviews, user.id) : null;
+
   return {
-    currentUser,
     canEditProfile,
+    currentUser,
     isOwner,
+    reviewCount: reviews ? reviews.reviewCount : null,
+    reviews: reviews ? reviews.reviews.slice(0, DEFAULT_API_PAGE_SIZE) : null,
```
kumar303: Why does this need to be sliced? Doesn't the API return the correct number of results since we sent it a page param?

tests/unit/amo/sagas/testReviews.js
```diff
@@ -288,4 +290,59 @@ describe(__filename, () => {
...
+
+    it('fetches reviews from the API', async () => {
+      const userId = 123;
+      const reviews = [fakeReview];
+
+      mockApi
+        .expects('getReviews')
+        .once()
+        .withArgs({
+          api: apiState,
```
kumar303: This key should be `apiState`, not `api`.

src/amo/actions/reviews.js
```diff
@@ -127,6 +131,76 @@ export function fetchReviews({
...
+export type SetUserReviewsAction = {|
+  type: typeof SET_USER_REVIEWS,
+  payload: {|
+    reviewCount: number,
+    reviews: Array<UserReviewType>,
+    userId: number,
+  |},
+|};
+
+export const setUserReviews = ({
```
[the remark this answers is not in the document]
willdurand: > What do you think? I don't mind if you do it in a follow-up issue. Part of that process would be to refactor the existing code to no longer use setAddonReviews but that too could come in a separate issue.

I was about to do it when I saw myself copy/pasting code, but then the patch would have been too large and would have touched many unrelated files.. I can definitely do it in a follow-up patch.

src/amo/components/UserProfile/index.js
willdurand: Yes, I was very sad not to be able to reuse the existing component :/

src/amo/actions/reviews.js
```diff
@@ -127,6 +131,76 @@ export function fetchReviews({
...
+    Array.isArray(reviews),
+    'reviews are required and must be an array',
+  );
+  invariant(userId, 'userId is required');
+
+  return {
+    type: SET_USER_REVIEWS,
+    payload: {
+      reviewCount,
+      reviews: reviews.map((review) => denormalizeReview(review)),
```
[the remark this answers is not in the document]
willdurand: ah right, that makes sense.

src/amo/actions/reviews.js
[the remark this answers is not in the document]
willdurand: Filed: https://github.com/mozilla/addons-frontend/issues/5415

src/amo/components/UserProfile/index.js
willdurand: Filed: https://github.com/mozilla/addons-frontend/issues/5416

src/amo/sagas/reviews.js
willdurand: wow, thank you!

src/amo/components/UserProfile/index.js
willdurand: mmh good catch, I think I don't need it anymore.

willdurand:
I believe I fixed most of the small issues and opened 2 issues for the bigger improvements.

src/amo/components/UserProfile/index.js
kumar303: Why not just move it to `UserReview` since you're already adding code here anyway? The `AddonReviewListItem` changes can definitely come later.

willdurand: ah, you mean creating this new component here? The diff is `+740 −46` already.

src/amo/components/UserProfile/index.js
```diff
@@ -295,8 +422,8 @@ export class UserProfileBase extends React.Component<Props> {
 }
 
 export function mapStateToProps(
-  state: {| users: UsersStateType |},
-  ownProps: Props,
+  state: {| reviews: ReviewState, users: UsersStateType |},
+  ownProps: InternalProps,
```
kumar303: This should be `Props` because `mapStateToProps` only receives the initial component. Did you get an error? I only see a reference to `ownProps.params` which is in `Props`.

src/amo/actions/reviews.js
```diff
@@ -170,7 +236,7 @@ export const setAddonReviews = ({
     payload: {
       addonSlug,
       reviewCount,
-      reviews: reviews.map((review) => denormalizeReview(review)),
```
kumar303: Thanks for fixing this one too.

src/amo/sagas/reviews.js
```diff
@@ -51,6 +59,37 @@ function* fetchReviews({
...
+  payload: { errorHandlerId, page, userId },
+}: FetchUserReviewsAction): Generator<any, any, any> {
+  const errorHandler = createErrorHandler(errorHandlerId);
+
+  try {
+    const state = yield select(getState);
+
+    const params: GetReviewsParams = {
+      apiState: state.api,
+      // Hide star-only ratings (reviews that do not have a body).
```
kumar303: Hmm, maybe we should show star-only ratings for user reviews? I'm not sure. It could help someone if they were trying to edit a rating. In any case, it's a future considering for UX I suppose.

src/amo/components/UserProfile/index.js
willdurand: ah no, no error. good catch

src/amo/sagas/reviews.js
willdurand: Good question.. I'll ask @jvillalobos

PR #5389, opened 2018-06-25
Hide last updated date when value is `null`

### LINKED ISSUE #5387: Last updated date can be `null` on an add-on detail page
I guess this only occurs when browsing the add-on detail page of an add-on with self-hosted versions (for instance, when creating and signing a brand new extension with `web-ext`):

screen shot 2018-06-25 at 12 16 38

It shows "invalid date" in the "More information" card.

willdurand:
Fix #5387

### DIFF
--- a/tests/unit/amo/components/TestAddonMoreInfo.js
+++ b/tests/unit/amo/components/TestAddonMoreInfo.js
@@ -416,4 +416,27 @@ describe(__filename, () => {
 
     expect(root.find('.AddonMoreInfo-version-history')).toHaveLength(0);
   });
+
+  it('renders the last updated date', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      last_updated: Date.now(),
+    });
+    const root = render({ addon });
+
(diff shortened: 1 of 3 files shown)


tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -416,4 +416,27 @@ describe(__filename, () => {
...
+  it('renders the last updated date', () => {
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      last_updated: Date.now(),
+    });
+    const root = render({ addon });
+
+    expect(root.find('.AddonMoreInfo-last-updated')).toHaveLength(1);
+    expect(root.find('.AddonMoreInfo-last-updated').children()).toIncludeText(
+      'a few seconds ago',
```
bobsilverberg: Would it be safer to check that it includes today's date, rather than the text `a few seconds ago`? I'm just wondering if there could possibly be timing issues that could produce an intermittent failure.

willdurand: I thought about it too but this component displays relative time here, so I am not sure I can use a fixed date.

PR #5393, opened 2018-06-25
Fix tooltip for email field

### LINKED ISSUE #5391: Add a more intuitive tooltip for account management links in Edit profile page
STR:

1. Log in to AMO-dev
2. Go to Edit my profile
3. Hover over the 'Need help?' and 'Manage Firefox Accounts...' links

Actual result:
A tooltip is displayed having the following text: "Email address cannot be changed here"

Expected result:
A more informative tooltip should be added to help users better understand the function of those links

Notes:
- reproduce don AMO-dev with FF60, Win10x64

fxa tooltip

willdurand:
Fix #5391 

---

This PR moves the title to the field (instead of the parent div).

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -506,15 +506,16 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                 value={this.state.username}
               />
 
-              <div title={i18n.gettext('Email address cannot be changed here')}>
+              <div>
                 <label className="UserProfileEdit--label" htmlFor="email">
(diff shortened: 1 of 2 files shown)


willdurand:
> also, were you going to add tooltips for "Need help?" and for "Manage Firefox Accounts.."? I could be wrong but I thought the emphasis in the issue were more with these links

I don't think we need tooltips because we rarely (never?) add tooltips on links.


> is there a way to adjust the style a bit so when the tooltip comes up its not blocked by the cursor:

I don't think so. That's the default cursor for disabled inputs and it is a bit unfortunate to have it like this.

Issue #5398, opened 2018-06-25
Express Flow external/internal prop types more concisely


kumar303:
As a follow-up to https://github.com/mozilla/addons-frontend/issues/5342, intoduce a more concise pattern for expressing external and internal component props.

willdurand:
> As a follow-up to #5342, intoduce a more concise pattern for expressing external and internal component props.

Could you maybe document a bit? I guess we don't want to use `InjectedProps` anymore?

PR #5399, opened 2018-06-25
Express external/internal props more concisely

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5398

* When coming back after the weekend and looking at this code I realized that `InjectedProps` is unecessary. We can just use `Props` and `InternalProps`. I updated all the code so that no one copies/pastes the wrong thing for future code.
* After understanding https://github.com/facebook/flow/issues/6502 a bit more, I moved some default prop types to `InternalProps`. This is only possible for props that are solely used for testing since we don't have Flow coverage for tests. I think it's nicer this way, though, because we won't have to add a bunch of invariant checks for the default props.

### DIFF
--- a/src/core/components/ErrorPage/index.js
+++ b/src/core/components/ErrorPage/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -11,15 +12,14 @@ import type { DispatchFunc } from 'core/types/redux';
 
 type Props = {|
   children: React.Node,
-  getErrorComponent: typeof getErrorComponentDefault,
+  getErrorComponent?: typeof getErrorComponentDefault,
 |};
 
-type InjectedProps = {|
+type InternalProps = {|
+  ...Props,
   dispatch: DispatchFunc,
   errorPage: ErrorPageState,
 |};
-
-type InternalProps = { ...Props, ...InjectedProps };
 
 export class ErrorPageBase extends React.Component<InternalProps> {
   static defaultProps = {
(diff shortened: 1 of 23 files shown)


src/core/components/ErrorPage/index.js
```diff
@@ -36,6 +36,7 @@ export class ErrorPageBase extends React.Component<InternalProps> {
 
   render() {
     const { children, errorPage, getErrorComponent } = this.props;
+    invariant(getErrorComponent, 'getErrorComponent() is undefined');
```
willdurand: So is the pattern "if a prop has a default value and the prop is not for testing purpose only, we mark it as optional and we add an `invariant` call"?

kumar303: Yes.

kumar303:
Good idea on adding docs 😅  I should have added that in the first patch. Added.

src/core/components/ErrorPage/index.js
bobsilverberg: I'm confused. If it has a default, why do we need the `invariant` at all? In case someone explicitly passes `undefined` into `getErrorComponent`? Or is this to deal with a Flow issue?

kumar303: Flow normally handles `defaultProps` just fine. It sees a required prop but then knows it can be left undefined because of the default value.

In https://github.com/mozilla/addons-frontend/issues/3563 I wanted to leverage Flow to help us validate component properties for those components wrapped in HOCs. Before any of the patches to address it, we were completely flying blind. You could instantiate any HOC wrapped component with invalid props or missing props and Flow would not complain.

The solution was not straight forward. It consists of manually declaring a type hint for the final exported component. However, it only requires one extra line of code and it will validate usage of props.

A side effect of manually declaring a type hint is that Flow loses the ability to infer `defaultProps`. This means we will now need to declare these props as optional even though internally they can be handled like required props. Using invariant should make it pretty easy to deal with this change.

PR #5402, opened 2018-06-26
I5073 tracking install updates

### LINKED ISSUE #5073: Update install tracking to special case static themes
The lightweight theme (LWT) installation uses its own API but static themes will be installed in the same way as extensions are. 

We'll need to special case the install events to not just use the "addon" action for static themes. 

One question is, should we special use something other than `theme` for the action so that we can tell static-themes apart from LWTs or just use `theme` for continuity?

rebmullin:
fixes #5073 

I’ve updated event category and action info so it's a bit more descriptive for tracking events:

Category will now display “AMO Theme [install type]”  if it its static or light weight theme, otherwise it will display “AMO Addon [install type]”.

For action, since static theme is installed the same way as an extension, I just added the type for clarification and to distinguish it from any other type of extension.

let me know if you think anything can be improved or made more clear.

Also, I am testing this via the console info. Is there a better way to test this to be sure this is coming through as expected or is this enough?


rebmullin:
hey @willdurand @bobsilverberg @kumar303 - is anyone free to take a look at this 1?

bobsilverberg:
Sure @rebmullin, I'll take a look after I've finished reviewing https://github.com/mozilla/addons-frontend/pull/5484

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
 
   return ADDON_TYPE_THEMES_FILTER;
 };
+
+export const getExtensionTypeAction = (type = ADDON_TYPE_EXTENSION) => {
```
bobsilverberg: This seems to duplicate the logic also found in `getAction` in `src/core/tracking.js`. Is there a need for both functions?

Also, I think the whole function could be reduced to:

```
export const getExtensionTypeAction = () => {
  return type === ADDON_TYPE_STATIC_THEME
    ? TRACKING_TYPE_STATIC_THEME
    : TRACKING_TYPE_EXTENSION;
}
```

Also, `getExtensionTypeAction` doesn't seem like an accurate description. I think this is more like `getTrackingAddonType`.

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
...
+  let action = TRACKING_TYPE_EXTENSION;
+
+  if (type === ADDON_TYPE_STATIC_THEME) {
+    action = TRACKING_TYPE_STATIC_THEME;
+  }
+
+  return action;
+};
+
+export const getAddonCategory = ({
```
bobsilverberg: `getAddonCategory` seems a bit of a vague name for this function, as it's not really the category of an add-on. Perhaps `getAddonEventCategory` would be better?

src/core/constants.js
```diff
@@ -195,8 +196,17 @@ export const INSTALL_ERROR = 'INSTALL_ERROR';
 
 // Tracking categories.
 export const INSTALL_CATEGORY = 'AMO Addon / Theme Installs';
```
bobsilverberg: Do we still need this? Won't it always be either `INSTALL_THEME_CATEGORY` or `INSTALL_EXTENSION_CATEGORY`?

src/core/constants.js
```diff
@@ -195,8 +196,17 @@ export const INSTALL_ERROR = 'INSTALL_ERROR';
...
+export const INSTALL_THEME_CATEGORY = 'AMO Theme Installs';
+export const INSTALL_EXTENSION_CATEGORY = 'AMO Addon Installs';
+
 export const INSTALL_STARTED_CATEGORY = 'AMO Addon / Theme Installs Started';
+export const INSTALL_THEME_STARTED_CATEGORY = 'AMO Theme Installs Started';
+export const INSTALL_EXT_STARTED_CATEGORY = 'AMO Addon Installs Started';
+
 export const UNINSTALL_CATEGORY = 'AMO Addon / Theme Uninstalls';
```
bobsilverberg: Do we still need this? Won't it always be either `UNINSTALL_THEME_CATEGORY` or `UNINSTALL_EXTENSION_CATEGORY`?

src/core/constants.js
```diff
@@ -195,8 +196,17 @@ export const INSTALL_ERROR = 'INSTALL_ERROR';
 
 // Tracking categories.
 export const INSTALL_CATEGORY = 'AMO Addon / Theme Installs';
+export const INSTALL_THEME_CATEGORY = 'AMO Theme Installs';
+export const INSTALL_EXTENSION_CATEGORY = 'AMO Addon Installs';
+
 export const INSTALL_STARTED_CATEGORY = 'AMO Addon / Theme Installs Started';
```
bobsilverberg: Do we still need this? Won't it always be either `INSTALL_THEME_STARTED_CATEGORY` or `INSTALL_EXT_STARTED_CATEGORY`?

Also, for consistency it might be better to make the latter the full `INSTALL_EXTENSION_STARTED_CATEGORY`.

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
...
+  type = ADDON_TYPE_EXTENSION,
+  installCategory = INSTALL_CATEGORY,
```
bobsilverberg: I think this would be clearer without any default values. You don't seem to ever call it without a `type` currently, and you do call it twice without an `installCategory`, but I think the calling code would be clearer if you always passed both.

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
...
+} = {}) => {
+  let category;
+  const isTheme = ADDON_TYPE_THEMES.includes(type);
```
bobsilverberg: Does this meet the requirements discussed at https://github.com/mozilla/addons-frontend/issues/5073#issuecomment-391068712? It seemed like we were being asked to differentiate between events for LWTs and static themes and I'm not sure we're doing that here. I may be misinterpreting the code and/or the requirements, but I think we need to be sure we're doing what's being asked.

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
...
+
+  switch (installCategory) {
+    case INSTALL_STARTED_CATEGORY:
+      category = isTheme
+        ? INSTALL_THEME_STARTED_CATEGORY
+        : INSTALL_EXTENSION_CATEGORY;
```
bobsilverberg: Should this be `INSTALL_EXT_STARTED_CATEGORY`?

src/core/utils/index.js
```diff
@@ -336,3 +348,39 @@ export const getAddonTypeFilter = (addonType, _config = config) => {
...
+
+export const getAddonCategory = ({
+  type = ADDON_TYPE_EXTENSION,
+  installCategory = INSTALL_CATEGORY,
+} = {}) => {
+  let category;
+  const isTheme = ADDON_TYPE_THEMES.includes(type);
+
+  switch (installCategory) {
+    case INSTALL_STARTED_CATEGORY:
```
bobsilverberg: I find the naming and use of these constants confusing. This constant `INSTALL_STARTED_CATEGORY` is actually only used here as a switch, so it could contain a value like `install:started`, but it in fact contains the value `AMO Addon / Theme Installs Started` even though that's never used as an actual category (because the actual category will be either `INSTALL_THEME_STARTED_CATEGORY` or `INSTALL_EXTENSION_CATEGORY`).

I think it would make sense to create a new set of constants with names like `INSTALL_STARTED_ACTION`, which contain values like `install:started`, and use those. The constant names using `*_CATEGORY` are being overloaded and it's making it quite confusing to follow.

src/core/utils/index.js
rebmullin: I *think* this is okay b/c I am updating the action property for this as this is currently the property that extension shows as 'addon' (the 1st comment mentions addressing [this](https://github.com/mozilla/addons-frontend/issues/5073#issue-325345556)). For static theme, this is where I am using 'addon:statictheme' to distinguish it from a regular extension install. 

The category piece is where I was addressing the 2nd part of the 2nd comment:

> Both stats can be added together if there's a need for an "all themes" stat, right?

Here I am using the category to show "AMO Themes [install type]" for both static theme and lightweight theme. I was thinking this could be a way to show the aggregated theme installs.

Also, it might make it clearer if you see exactly what data comes through on each type of install:
static theme:
Alt text

lightweight theme:
Alt text

extension:
Alt text

I guess there are different ways to group it and I can def move things around too if you think it might be clearer another way

src/core/utils/index.js
rebmullin: thanks @bobsilverberg ...I'll take another look at this.

src/core/constants.js
rebmullin: this was set up this way so the "general" category could be passed in and then depending on the type, the more descriptive category was used. I'll take another look tho to see about cleaning this up.

yes, I meant to change ext => extension. Thanks for catching that one :)

src/core/utils/index.js
bobsilverberg: Looking at the data it does look okay to me. Thanks! I think I misunderstood how the category was being used.

src/core/utils/index.js
rebmullin: oh?! i didn't not see the `getAction`. I'll take a look...

Regarding the name tho, it's not really an add on type  - right now it's either "addon" or "theme". But i'll revisit this name if it is in fact needed

src/core/constants.js
rebmullin: this was set up this way so the "general" category could be passed in and then depending on the type, the more descriptive category was used. I'll take another look tho to see about cleaning this up.

src/core/utils/index.js
rebmullin: ok cool, I thought that might help "explain" more what I was thinking : )

src/core/constants.js
rebmullin: yep, I liked your suggestion about using *_ACTION instead and will remove these "general" categories which were really being misused :/

rebmullin:
@bobsilverberg, thanks so much for the initial review : ) I've made some updates per your comments. Can you see how it looks now?

Also, from I can tell travis mentions a failure in a py file. Is that something that needs to be addressed before this gets merged?

src/core/tracking.js
```diff
@@ -159,9 +171,34 @@ export function getAction(type) {
...
       [ADDON_TYPE_EXTENSION]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_OPENSEARCH]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_STATIC_THEME]: TRACKING_TYPE_STATIC_THEME,
       [ADDON_TYPE_THEME]: TRACKING_TYPE_THEME,
     }[type] || TRACKING_TYPE_INVALID
   );
 }
 
+export const getAddonEventCategory = (type, installAction) => {
+  let category;
```
bobsilverberg: Instead of defining a `category` here, populating it in the switch, and returning it at the end, could you just return the category from inside the switch?  Something like this:

```js
export const getAddonEventCategory = (type, installAction) => {
  const isTheme = ADDON_TYPE_THEMES.includes(type);

  switch (installAction) {
    case INSTALL_STARTED_ACTION:
      return isTheme
        ? INSTALL_THEME_STARTED_CATEGORY
        : INSTALL_EXTENSION_STARTED_CATEGORY;
    case UNINSTALL_ACTION:
      return isTheme
        ? UNINSTALL_THEME_CATEGORY
        : UNINSTALL_EXTENSION_CATEGORY;
    default:
      return isTheme ? INSTALL_THEME_CATEGORY : INSTALL_EXTENSION_CATEGORY;
  }
};
```

src/core/tracking.js
```diff
@@ -159,9 +171,34 @@ export function getAction(type) {
   return (
```
bobsilverberg: I find the name of this function confusing, because it doesn't return an "action" at all, it returns a code for a type of add-on (e.g., `addon`, `addon:statictheme`, etc.). So maybe it should be `getAddonTypeForAction` or something like that?

Also, we should have unit tests for this function.

tests/unit/core/TestInstallAddon.js
```diff
@@ -1128,8 +1134,34 @@ describe(`${__filename}: withInstallHelpers`, () => {
...
-          action: TRACKING_TYPE_EXTENSION,
-          category: INSTALL_CATEGORY,
+          action: getAction(ADDON_TYPE_EXTENSION),
+          category: getAddonEventCategory(ADDON_TYPE_EXTENSION, INSTALL_ACTION),
+          label: addon.name,
+        });
+      });
+    });
+
+    it('tracks a static theme addon install', () => {
```
bobsilverberg: I don't see a test for tracking *the start* of a static theme add-on install.

tests/unit/core/TestInstallAddon.js
```diff
@@ -1322,7 +1385,7 @@ describe(`${__filename}: withInstallHelpers`, () => {
       return uninstall(addon).then(() => {
         sinon.assert.calledWith(fakeTracking.sendEvent, {
           action: TRACKING_TYPE_THEME,
```
bobsilverberg: `uninstall` calls `getAction` for this value, so you should do that in the test as well.

tests/unit/core/TestInstallAddon.js
```diff
@@ -1322,7 +1385,7 @@ describe(`${__filename}: withInstallHelpers`, () => {
...
-          category: UNINSTALL_CATEGORY,
+          category: getAddonEventCategory(addon.type, UNINSTALL_ACTION),
```
bobsilverberg: Elsewhere you have used an explicit type constant (e.g., `ADDON_TYPE_THEME`) as opposed to `addon.type` and I think it would be good to do that here as well, in case something is wrong with the test setup and you don't end up with the type of `addon` you expect.

tests/unit/core/TestInstallAddon.js
```diff
@@ -1347,7 +1410,7 @@ describe(`${__filename}: withInstallHelpers`, () => {
       return uninstall(addon).then(() => {
         sinon.assert.calledWith(fakeTracking.sendEvent, {
           action: 'invalid',
```
bobsilverberg: I think it would be better to use the constant `TRACKING_TYPE_INVALID` here.

tests/unit/core/TestInstallAddon.js
```diff
@@ -1347,7 +1410,7 @@ describe(`${__filename}: withInstallHelpers`, () => {
...
-          category: UNINSTALL_CATEGORY,
+          category: getAddonEventCategory(addon.type, UNINSTALL_ACTION),
```
bobsilverberg: You could declare a constant for `invalidType` at the top of the test and use it both to create the add-on and also here, instead of `addon.type`.

src/core/tracking.js
```diff
@@ -159,9 +171,34 @@ export function getAction(type) {
...
     {
       [ADDON_TYPE_EXTENSION]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_OPENSEARCH]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_STATIC_THEME]: TRACKING_TYPE_STATIC_THEME,
       [ADDON_TYPE_THEME]: TRACKING_TYPE_THEME,
     }[type] || TRACKING_TYPE_INVALID
   );
 }
 
+export const getAddonEventCategory = (type, installAction) => {
```
bobsilverberg: We need tests for this function, please.

src/core/tracking.js
rebmullin: I think it was getting the action for the event but I see what you're saying. I can def update as you suggest..

tests/unit/core/TestInstallAddon.js
rebmullin: oh? ok i'll take a look..

tests/unit/core/TestInstallAddon.js
rebmullin: yep, good catch.. thanks : )

src/core/tracking.js
bobsilverberg: Hmm, I know I suggested `getAddonTypeForAction`, but looking at the code again the word "action" is really just being used because we're using this value to populate the `action` argument of the `tracking.sendEvent` function. But it doesn't really represent at action at all. What is represents is the type of add-on being used to record a tracking event. So maybe `getAddonTypeForTracking` makes more sense. What do you think?

tests/unit/core/test_tracking.js
```diff
@@ -150,17 +168,89 @@ describe('Tracking', () => {
...
   });
 
   it('returns invalid for unknown type', () => {
-    expect(getAction('whatever')).toEqual('invalid');
+    expect(getAddonTypeForAction('whatever')).toEqual(TRACKING_TYPE_INVALID);
+  });
+});
+
+describe('getAddonEventCategory', () => {
+  it('returns Addon Installs Started category if type is a extension and install started is the action type', () => {
```
bobsilverberg: Thanks for adding these tests @rebmullin. The descriptions are quite long, and I don't think you need to specify the actual value you expect in the description. It could just be something like `'returns the expected category when type is extension and installAction is install started'`. The same goes for the other tests below.

rebmullin:
> You'll probably be upset with me...

bob~of course not! I appreciate the feedback and thoughtful consideration : )

> So maybe getAddonTypeForTracking makes more sense. What do you think?

yep, I like it and agree it is a better name for its purpose : )

PR #5404, opened 2018-06-26
Introduce a useFiltersFromLocation property to AutoSearchInput to support changes to CollectionManager

bobsilverberg:
Supports #3991 

I need to be able to prevent the AutoSearchInput component from automatically creating filters based on the location query parameters for my changes to the CollectionManager to support sorting add-ons in a collection, and this seemed like a simple way of doing that.


bobsilverberg:
Actually, I just realized there's a better way of doing this. Sorry for the noise.

PR #5405, opened 2018-06-26
Allow AutoSearchInput to be called without a location property

bobsilverberg:
Supports #3991

I need to be able to prevent the AutoSearchInput component from automatically creating filters based on the location query parameters for my changes to the CollectionManager to support sorting add-ons in a collection, and this seems like a simple way of doing that.

### DIFF
--- a/tests/unit/amo/components/TestAutoSearchInput.js
+++ b/tests/unit/amo/components/TestAutoSearchInput.js
@@ -297,6 +297,28 @@ describe(__filename, () => {
       );
     });
 
+    it('can be used without a location', () => {
+      const { store } = dispatchClientMetadata();
+      const dispatch = sinon.spy(store, 'dispatch');
+      const root = render({
+        store,
+      });
+
+      const query = 'ad blocker';
+      fetchSuggestions({ root, query });
+
+      sinon.assert.calledWith(
+        dispatch,
+        autocompleteStart({
+          errorHandlerId: root.instance().props.errorHandler.id,
+          filters: {
+            operatingSystem: OS_WINDOWS,
+            query,
+          },
+        }),
+      );
+    });
+
     it('lets you override the default operating system', () => {
       const { store } = dispatchClientMetadata();
       const dispatch = sinon.spy(store, 'dispatch');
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAutoSearchInput.js
```diff
@@ -297,6 +297,28 @@ describe(__filename, () => {
...
+        store,
+      });
+
+      const query = 'ad blocker';
+      fetchSuggestions({ root, query });
+
+      sinon.assert.calledWith(
+        dispatch,
+        autocompleteStart({
+          errorHandlerId: root.instance().props.errorHandler.id,
```
willdurand: It would be better to use `createStubErrorHandler()` and pass it to `render()`.

bobsilverberg:
> How do you want to use this component without a location because it uses the withRouter() HOC, which (always?) injects the location.

I guess that's what I guess for separating this change from the code that will use it. I didn't test this new version with my code that needs it. 

I didn't realize that it automatically gets a `location` injected because we do pass a `location` in when it is called: 

- https://github.com/mozilla/addons-frontend/blob/d9b067d4a53c25821d6defc2eb75583ac4892f1f/src/amo/components/SearchForm/index.js#L58
- https://github.com/mozilla/addons-frontend/blob/b8a86fce7ba51ed7681ad05354f0dfa9f6af1932/src/amo/components/CollectionManager/index.js#L412

which seems odd.

I'll test this patch with my other code to see if it works as expected.

tests/unit/amo/components/TestAutoSearchInput.js
bobsilverberg: Yes, I really just copied an existing test and only made the changes necessary to it.

willdurand:
> I didn't realize that it automatically gets a location injected because we do pass a location in when it is called:

ah, then the `withRouter` HOC should probably be removed.

bobsilverberg:
Removing `withRouter` fixes it for me, and if we're always passing in a `location` (except when we don't want it to use the `location`) then we don't need `withRouter` anymore. I've updated the PR to remove it.

bobsilverberg:
Sorry, @willdurand, I'm getting a bit flustered trying to juggle all these PRs I'm working on and teasing apart, so I did a `git commit --amend` where I shouldn't have so the change to remove `withRouter` got folded into the original commit. I did fix the test in a separate commit though. This should be ready for another look.

PR #5407, opened 2018-06-26
Changes to api/reducer/saga code to support sorting collection add-ons

bobsilverberg:
Supports #3991 
Builds on #5405 

Although this doesn't directly depend on #5405, the next PR in the series does, which also depends on this, which is why they're all linked. For the purposes of review you can ignore the first two commits, which belong to #5405.

This introduces changes to api, reducer and saga code to support sorting collection add-ons, in particular the use of the `filters` property. It also includes changes to components so they make use of that new `filters` property. The main changes to the UI for sorting collection add-ons will come in the next PR.


src/amo/components/CollectionManager/index.js
```diff
@@ -409,7 +409,6 @@ export class CollectionManagerBase extends React.Component<
             inputPlaceholder={i18n.gettext(
               'Find an add-on to include in this collection',
             )}
-            location={router.location}
```
willdurand: Do we need the `router` somewhere? If not, we can probably remove the `withRouter` HOC.

src/core/constants.js
```diff
@@ -133,6 +133,18 @@ export const SEARCH_SORT_RANDOM = 'random';
 export const SEARCH_SORT_RELEVANCE = 'relevance';
 export const SEARCH_SORT_UPDATED = 'updated';
 
+// Collection add-ons sort values
+export const COLLECTION_SORT_DATE_ADDED: 'added' = 'added';
```
willdurand: If the other one is `DESCENDING`, this one should be `ASCENDING` no?

tests/unit/amo/components/TestCollection.js
```diff
@@ -432,69 +449,72 @@ describe(__filename, () => {
...
         errorHandlerId: errorHandler.id,
-        page,
+        filters: newFilters,
         username: defaultUser,
         slug: defaultSlug,
       }),
     );
   });
 
-  it('defaults to first page when there is no page and collection is loaded', () => {
```
bobsilverberg: This test case was removed because there is no longer a condition where the component is instantiated without a `page`. `mapStateToProps` now generates the `filters` prop and it defaults the `page` filter to `1`, which is tested in the test case for `it('dispatches fetchCurrentCollection on mount')`. I can create a separate test case to assert the default values for `filters`, if you prefer, but it seemed unnecessary as they're tested in that case.

willdurand:
> I did remove a number of test cases, but they are all justified, I believe. I will go through the diff and add comments to explain the reasons.

mmh, I don't think we should ever remove a test case that covers a bug report (except if the source file/code has been removed, maybe).

tests/unit/amo/components/TestCollection.js
```diff
@@ -879,31 +949,9 @@ describe(__filename, () => {
...
-
-    const editLink = wrapper.find('.Collection-edit-link').find(Button);
-    expect(editLink).toHaveLength(1);
-    expect(editLink).toHaveProp(
-      'to',
-      `/collections/${defaultUser}/${defaultCollectionDetail.slug}/edit/`,
-    );
-  });
-
-  it('includes the page number in the edit link', () => {
```
bobsilverberg: This test case was removed because what it is testing is also tested in the case for `it('links internally to a Collection edit page')`. We no longer include the text `?page=${page}` in the `pathname`, but rather pass along `filters` in the `query` property of the `to` prop. All of that is being tested in the test case I mention above. There really doesn't seem to be a need for two different tests now. We had 2 before because sometimes the text for `?page=${page}` was included in `pathname` and sometimes it was not, but now we always pass `filters` into the `query` property of the `to`.

tests/unit/amo/components/TestCollection.js
```diff
@@ -1094,51 +1144,7 @@ describe(__filename, () => {
...
-        addonId,
-        errorHandlerId: errorHandler.id,
-        page,
-        slug: collectionDetail.slug,
-        username: collectionDetail.author.username,
-      }),
-    );
-  });
-
-  it('dispatches removeAddonFromCollection when removeAddon is called without a page defined', () => {
```
bobsilverberg: This test, as well as all the others that are for `dispatches xxx when yyy is called without a page defined` was removed because there is no longer a condition where the component doesn't have a `page` prop. It will always have a `filters` prop and that it always passed, so there's no need for 2 tests for each of these cases (one with a `page` prop and one without).

tests/unit/amo/components/TestCollection.js
```diff
@@ -1240,51 +1248,7 @@ describe(__filename, () => {
...
         addonId,
         errorHandlerId: errorHandler.id,
-        page,
-        slug: collectionDetail.slug,
-        username: collectionDetail.author.username,
-      }),
-    );
-  });
-
-  it('dispatches deleteCollectionAddonNotes when deleteNote is called without a page defined', () => {
```
bobsilverberg: As mentioned above, this test case was removed because there's no need to test the functionality `without a page defined`, because that no longer applies to this component.

tests/unit/amo/components/TestCollection.js
```diff
@@ -1315,53 +1281,7 @@ describe(__filename, () => {
...
-        errorHandlerId: errorHandler.id,
-        notes,
-        page,
-        slug: collectionDetail.slug,
-        username: collectionDetail.author.username,
-      }),
-    );
-  });
-
-  it('dispatches updateCollectionAddon when saveNote is called without a page defined', () => {
```
bobsilverberg: As mentioned above, this test case was removed because there's no need to test the functionality `without a page defined`, because that no longer applies to this component.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -714,39 +715,7 @@ describe(__filename, () => {
         slug: collection.slug,
         editing: true,
         errorHandlerId: errorHandler.id,
-        page,
-        username: signedInUsername,
-      }),
-    );
-  });
-
-  it('dispatches addAddonToCollection with a default page of 1 when selecting an add-on', () => {
```
bobsilverberg: This test case was removed because this component used to contain logic that defaulted the `page` to `1` if no page was specified, but now this component always receives a `filters` prop instead and therefore has no logic for specifying a default value for `page`.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -801,15 +795,6 @@ describe(__filename, () => {
         fetchCurrentCollectionPage(partialParams);
       }).toThrow('username is required');
     });
-
-    it('throws an error when page is missing', () => {
```
bobsilverberg: This test case was removed because the reducer method no longer expects a `page` parameter.

src/amo/components/CollectionManager/index.js
bobsilverberg: We use `router` in `onCancel`.

PR #5414, opened 2018-06-26
Sort add-ons in a collection

### LINKED ISSUE #3991: Sort add-ons in a collection

bobsilverberg:
Fixes #3991 
Depends on ~~#5407~~ 

This finishes the "Sort add-ons in a collection" feature.

screenshot 2018-06-26 13 56 26

The user should be able to sort by any of the 4 fields, and the sort sequence should be "remembered" when they move between pages, and even when switching from "View" to "Edit" mode. Both the sort sequence and page should also be maintained when actions are performed, such as adding an add-on to a collection, removing an add-on from a collection, and updating/removing an add-on note.

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -36,11 +36,6 @@ import {
 } from 'core/constants';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
-import log from 'core/logger';
-import {
-  convertFiltersToQueryParams,
-  convertQueryParamsToFilters,
-} from 'core/searchUtils';
 import { sanitizeHTML } from 'core/utils';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
@@ -116,6 +111,13 @@ export class CollectionBase extends React.Component<InternalProps> {
     this.loadDataIfNeeded(nextProps);
   }
 
+  convertFiltersToQueryParams(filters: CollectionFilters) {
+    return {
+      page: filters.page,
+      collection_sort: filters.collectionSort,
+    };
+  }
+
   onDelete = (event: SyntheticEvent<HTMLButtonElement>) => {
     event.preventDefault();
 
@@ -151,7 +153,7 @@ export class CollectionBase extends React.Component<InternalProps> {
     }`;
     router.push({
       pathname,
-      query: convertFiltersToQueryParams(newFilters),
+      query: this.convertFiltersToQueryParams(newFilters),
     });
   };
 
@@ -268,7 +270,7 @@ export class CollectionBase extends React.Component<InternalProps> {
       // https://github.com/mozilla/addons-frontend/issues/4293
       props.to = {
         pathname: this.editUrl(),
-        query: convertFiltersToQueryParams(filters),
+        query: this.convertFiltersToQueryParams(filters),
       };
     } else {
       props.href = this.editUrl();
@@ -471,7 +473,7 @@ export class CollectionBase extends React.Component<InternalProps> {
           count={collection.numberOfAddons}
           currentPage={filters.page}
           pathname={editing ? this.editUrl() : this.url()}
-          queryParams={convertFiltersToQueryParams(filters)}
+          queryParams={this.convertFiltersToQueryParams(filters)}
         />
       ) : null;
 
@@ -575,12 +577,10 @@ export const mapStateToProps = (
   const { loading } = state.collections.current;
   const { location } = ownProps;
 
-  const filtersFromLocation = convertQueryParamsToFilters(location.query);
   const filters = {
-    page: filtersFromLocation.page || 1,
+    page: location.query.page || 1,
     collectionSort:
-      filtersFromLocation.collectionSort ||
-      COLLECTION_SORT_DATE_ADDED_DESCENDING,
+      location.query.collection_sort || COLLECTION_SORT_DATE_ADDED_DESCENDING,
   };
 
   const currentUser = getCurrentUser(state.users);
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -1,5 +1,4 @@
 /* @flow */
-import { oneLineTrim } from 'common-tags';
 import config from 'config';
 import deepEqual from 'deep-eql';
 import invariant from 'invariant';
@@ -25,7 +24,7 @@ import { getCurrentUser, hasPermission } from 'amo/reducers/users';
 import AuthenticateButton from 'core/components/AuthenticateButton';
 import Paginate from 'core/components/Paginate';
 import {
-  COLLECTION_SORT_DATE_ADDED,
+  COLLECTION_SORT_DATE_ADDED_ASCENDING,
   COLLECTION_SORT_DATE_ADDED_DESCENDING,
   COLLECTION_SORT_NAME,
   COLLECTION_SORT_POPULARITY,
@@ -68,7 +67,6 @@ export type Props = {|
   collection: CollectionType | null,
   creating: boolean,
   editing: boolean,
-  filters: CollectionFilters,
   loading: boolean,
 |};
 
@@ -78,6 +76,7 @@ type InternalProps = {|
   clientApp: string,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
+  filters: CollectionFilters,
   hasEditPermission: boolean,
   i18n: I18nType,
   isLoggedIn: boolean,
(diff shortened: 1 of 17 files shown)


src/amo/components/Collection/index.js
```diff
@@ -124,6 +133,50 @@ export class CollectionBase extends React.Component<InternalProps> {
     );
   };
 
+  onSortSelect = (event: SyntheticEvent<any>) => {
```
willdurand: Can we do better than `any`?

src/amo/components/Collection/index.js
```diff
@@ -124,6 +133,50 @@ export class CollectionBase extends React.Component<InternalProps> {
...
+
+    event.preventDefault();
+
+    const sortValue = event.currentTarget.value;
+    const newFilters = {
+      ...filters,
+      sort: sortValue,
+    };
+
+    const pathname = oneLineTrim`/${lang}/${clientApp}
```
willdurand: It is a bit hard to parse to be honest.

src/ui/components/Button/index.js
```diff
@@ -27,7 +27,7 @@ export default class Button extends React.Component {
     href: PropTypes.string,
     micro: PropTypes.bool,
     puffy: PropTypes.bool,
-    to: PropTypes.string,
+    to: PropTypes.string || PropTypes.object,
```
willdurand: I think we should use `PropTypes.oneOfType([])`

tests/unit/amo/components/TestCollection.js
```diff
@@ -1304,6 +1333,51 @@ describe(__filename, () => {
...
+        const wrapper = renderComponent({
+          editing,
+          location: fakeRouterLocation({ query: { page, sort } }),
+          params: { username, slug },
+          router,
+          store,
+        });
+
+        // This emulates a user choosing an item from the sort select.
+        wrapper.instance().onSortSelect(
```
willdurand: Could we actually use `wrapper.simulate()` here? We should use `instance()` "parsimoniously".

src/amo/components/Collection/index.js
bobsilverberg: I converted it to a single line and prettier broke it up for me, so I guess that's how it will be.

bobsilverberg:
@willdurand Do you think you'll get a chance to look at this again before you disappear?

willdurand:
> @willdurand Do you think you'll get a chance to look at this again before you disappear?

Yes, that's on my list for tomorrow since you're at the workshop today 😄

bobsilverberg:
Ok, I just thought if there were more changes required I could get to them tonight, which would give you time for another review tomorrow.

willdurand:
> Ok, I just thought if there were more changes required I could get to them tonight, which would give you time for another review tomorrow.

I'll see what I can do 😄

src/amo/components/Collection/index.js
```diff
@@ -500,7 +578,9 @@ export const mapStateToProps = (
   const filtersFromLocation = convertQueryParamsToFilters(location.query);
   const filters = {
     page: filtersFromLocation.page || 1,
-    sort: filtersFromLocation.sort || COLLECTION_SORT_DATE_ADDED_DESCENDING,
+    collectionSort:
```
willdurand: So I see why you've changed the `sort` parameter but the problem is that it remains in the URL when navigating to the search results page. I think we should handle this differently.

bobsilverberg:
> About the query parameters issue, if you do not use convertQueryParamsToFilters(), then there is no need to modify the AutoSearchInput component.
>
> This component also uses convertQueryParamsToFilters() for creating the filters from the query string, but it ignores/removes parameters that it does not know. I found this while trying to reproduce the issue on the user profile pages, but the user profile pages are not affected.
>
> I think that would probably reduce the overall diff of your patch. Since you have modified the query parameter name already, it should work. You probably only need to make it a default prop and use it to retrieve the value in location.query.

@willdurand I don't understand these comments. The problem (or what I understand the problem to be) is that the `collection_sort` query parameter is being picked up by the `AutoSearchInput` component at the top of the page. I don't see how removing the use of `convertQueryParamsToFilters` in `Collections/index.js` will have any affect on that.

I'm also confused because I implemented the exact solution you suggested, and it works, and prevents the `collection_sort` query parameter from being picked up by the `AutoSearchInput` component, but now you're saying to do something different yet again. I'm really not understanding what the problem with the current patch is, nor what you are suggesting, or why it would be better than the changes I just implemented.

willdurand:
> I don't see how removing the use of convertQueryParamsToFilters in Collections/index.js will have any affect on that.

@bobsilverberg if you don't use it, then you can remove `collection_sort` from the list of `paramToFilter`. (Why would you keep it, right?) Therefore, the autosuggest component will automatically skip this parameter because convertQueryParamsToFilters() only preserves the filters in this map of param and filter names, and removes the extra query parameters/filters.

bobsilverberg:
> @bobsilverberg if you don't use it, then you can remove collection_sort from the list of paramToFilter. (Why would you keep it, right?) Therefore, the autosuggest component will automatically skip this parameter because convertQueryParamsToFilters() only preserves the filters in this map of param and filter names, and removes the extra query parameters/filters.

Ah, I thought you were specifically concerned with not having `collection_sort` appear on the URL when using the `AutoSearchInput` at the top of the page, which is what I did in my last change. I can make the above change.

willdurand:
> I'm also confused because I implemented the exact solution you suggested

Sorry, that was a only suggestion. Testing it in real life allowed me to find a simpler solution because the original suggestion is a hack: we would have to do it for all parameters of potentially all pages with query params.

Using convertQueryParamsToFilters() only for search would make sense because it was created to deal with legacy query params and new filter names (I think, given what is inside `paramToFilter`)

bobsilverberg:
Ok, I've pushed a new commit. 

I had to remove the `CollectionAddonsSortType` because I couldn't get Flow to recognize that the string coming from the select input was a valid instance of `CollectionAddonsSortType` without having to hardcode all the values of `CollectionAddonsSortType` into an `invariant` call. I tried using the constants I'd defined, e.g.:

```js
    invariant(
    [
        COLLECTION_SORT_DATE_ADDED_ASCENDING,
        COLLECTION_SORT_DATE_ADDED_DESCENDING,
        COLLECTION_SORT_NAME,
        COLLECTION_SORT_POPULARITY,
      ].includes(collectionSort),
      'collectionSort does not have a valid value',
    );
``` 

but that didn't work. Flow would only be happy with an `invariant` like:

```js
    invariant(
      ['added', '-added', 'name', 'popularity']
        .includes(collectionSort),
      'collectionSort does not have a valid value',
    );
``` 

and I really don't think we should be repeating those values. I don't think we _need_ the security of the `CollectionAddonsSortType`, and I think just using `string` is fine. We don't tend to define a type for valid types in `core/constants.js`.

bobsilverberg:
@kumar303 Now that @willdurand is away for a week, do you think you could take a look at this one, please and thank you?

src/amo/components/Collection/styles.scss
```diff
@@ -44,6 +44,19 @@
...
+  margin-top: 12px;
+}
+
+.Sort-label {
+  color: $grey-50;
+  display: block;
+}
+
+.Sort-select {
+  margin: 0 0 5px;
```
kumar303: This should be `6px` to match our specs. We typically use 2, 4, 6, 12...

tests/unit/amo/components/TestCollection.js
```diff
@@ -663,10 +657,35 @@ describe(__filename, () => {
...
+    expect(wrapper.find('.Sort-label')).toHaveText('Sort add-ons by');
+    expect(wrapper.find('.Sort-select')).toHaveProp('defaultValue', sort);
+    expect(wrapper.find('.Sort-select')).toHaveProp(
+      'onChange',
+      wrapper.instance().onSortSelect,
+    );
+    wrapper
+      .find('.Sort-select')
+      .children()
+      .forEach((option, index) => {
```
kumar303: This is a risky pattern because if the array has zero items, the test will pass. To make it fail, you'd write it more like:

```js
const options = wrapper.find('.Sort-select').children();

sortOptions.forEach((item, index) => {
  expect(options.at(index)).toHaveProp('value', item.value);
  expect(options.at(index)).toHaveText(item.children);
});
```

That is, the `sortOptions` array is under your control so it should lead the assertion.

bobsilverberg:
> I also noticed that filters are getting lost after an edit.

Yes, that's odd. I was sure that was working in an earlier iteration. It must be something I did in the last commit or two to address review concerns. I'll see if I can figure it out and fix it.

bobsilverberg:
Thanks for the review @kumar303. I have addressed all of the issues and also fixed that problem with the sort filter disappearing when switching from edit to view.

kumar303:
r+ when tests pass, that is

PR #5428, opened 2018-06-27
Add .prettierignore file

### LINKED ISSUE #5423: Add a `.prettierignore` file to tell Prettier not to process some files
As @bobsilverberg pointed out on IRC, Prettier should not process the locale files (because it does not do anything anyway and it is a waste of time).

We should create a `.prettierignore` file with the following content:

```
src/locale/
```

willdurand:
Fix #5423

---

This should speed up the command a bit, but also allow us to enable Prettier on other files faster.

JS files at the root have been prettier-ized because... we use Prettier on all JS files.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,19 @@
+# exclude everything by default
+*.*
+# exclude these files
+Dockerfile
+# exclude these directories
+/assets/
+/bin/
+/config/
+/dist/
+/docker/
+/flow/
+/locale/
+/node_modules/
+/src/core/fonts/
+/src/disco/video/
+/src/locale/
+/tests/ui/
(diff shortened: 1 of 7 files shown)


.prettierignore
```diff
@@ -0,0 +1,19 @@
+# exclude everything by default
+*.*
+# exclude these files
+Dockerfile
+# exclude these directories
+/assets/
+/bin/
+/config/
```
kumar303: Are you sure we shouldn't format `bin` or `config`?

willdurand: Not now, but soon 😈

willdurand:
I suppose I could have request a review from both of you, first come, first served.

PR #5429, opened 2018-06-27
Prettier markdown files

willdurand:
:warning: Depends on mozilla/addons-frontend#5428

Fix mozilla/addons#11971

---

Because... why not?

### DIFF
--- a/docs/moz-addon-manager.md
+++ b/docs/moz-addon-manager.md
@@ -58,7 +58,6 @@ After you have everything, the bootstrap script prompts you to check out a clone
 
 Change into the source directory and apply this patch to allow a local server at a URL like `localhost:3000`.
 
-<!-- prettier-ignore -->
 ```diff
 diff --git a/browser/base/content/browser-addons.js b/browser/base/content/browser-addons.js
 --- a/browser/base/content/browser-addons.js
@@ -112,7 +111,6 @@ diff --git a/toolkit/mozapps/extensions/AddonManagerWebAPI.cpp b/toolkit/mozapps
 
    nsAutoCString host;
    rv = uri->GetHost(host);
-
 ```
 
 This patch will:


README.md
```diff
@@ -16,54 +16,54 @@ Please submit all security-related bugs through Bugzilla using the [web security
...
 Here are some commands you can run:
 
-| Command                     | Description |
-|-----------------------------|-------------|
-| yarn amo                    | Start the dev server/proxy (for amo) using data from Docker |
-| yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/) |
-| yarn amo:no-proxy           | Start the dev server without a proxy (for amo) using data from Docker |
-| yarn amo:stage              | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/) |
+| Command                     | Description                                                                                                |
+| --------------------------- | ---------------------------------------------------------------------------------------------------------- |
```
kumar303: This entire table looks weird but I guess it renders ok

docs/moz-addon-manager.md
```diff
@@ -98,18 +97,18 @@ diff --git a/browser/base/content/browser-addons.js b/browser/base/content/brows
 +++ b/browser/base/content/browser-addons.js
 @@ -664,12 +664,11 @@ var LightWeightThemeWebInstaller = {
    },
- 
+
    _isAllowed(principal) {
 -    if (!principal || !principal.URI || !principal.URI.schemeIs("https")) {
 +    if (!principal || !principal.URI) {
```
kumar303: Can you make it ignore this patch diff?

willdurand:
mmh I suppose I could use https://prettier.io/docs/en/options.html#prose-wrap, because we are not consistent currently. I like to wrap content at 80chars, thoughts?

kumar303:
> I like to wrap content at 80chars, what do you think?

Actually, we did make a decision once to never wrap prose. I forget why but I configured vim to auto-wrap for me so I don't care. I think it would be a good idea to enforce unwrapped prose.

kumar303:
Maybe @muffinresearch remembers why we decided to always unwrap prose.

README.md
willdurand: Yeah, it renders correctly

docs/moz-addon-manager.md
willdurand: doh, I overlooked that part.

willdurand:
> Actually, we did make a decision once to never wrap prose. I forget why but I configured vim to auto-wrap for me so I don't care. I think it would be a good idea to enforce unwrapped prose.

ah well, I can live with that I guess. I can definitely set `wrap-prose` to `never`.

muffinresearch:
> Maybe @muffinresearch remembers why we decided to always unwrap prose.

I don't recall. In general IMHO 80 char limits do seem a bit weird in 2018.

willdurand:
should be better now (and rebased)

kumar303:
> 80 char limits do seem a bit weird in 2018

But it's 2018 and I have a 24" monitor that lets me do three horizontal splits in vim! But, yeah, this only applies to code.

docs/moz-addon-manager.md
```diff
@@ -21,95 +14,69 @@ Refresh the page, open the JavaScript console, and you should see a global `mozA
...
 
-After you have everything, the bootstrap script prompts you to
-check out a clone of [mozilla-central](https://hg.mozilla.org/mozilla-central/).
+After you have everything, the bootstrap script prompts you to check out a clone of [mozilla-central](https://hg.mozilla.org/mozilla-central/).
 
-Change into the source directory and apply this patch to allow a local server
-at a URL like `localhost:3000`.
+Change into the source directory and apply this patch to allow a local server at a URL like `localhost:3000`.
 
+<!-- prettier-ignore -->
```
kumar303: This doesn't seem to have any effect 😛  Or maybe you need to revert the first patch that formatted the diff code?

kumar303: That is, the diff code has still been altered in comparison to master.

muffinresearch:
> But it's 2018 and I have a 24" monitor that lets me do three horizontal splits in vim! But, yeah, this only applies to code.

Out of genuine curiosity what char count does that give you per split?

PR #5432, opened 2018-06-27
add prettier to stylelint

### LINKED ISSUE #5431: Apply Prettier to SCSS files
Prettier supports SCSS/CSS, so we could use to prettier-ize our style files.

Our `stylelint` config will conflict with Prettier because we use double quotes in our SCSS files, but not in our JS files. This seems inconsistent, so I guess, we'll need to update the rule in `.stylelintrc` to use `"string-quotes": "single"`.

championshuttler:
Fixes #5431

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -41,5 +41,6 @@
         "after-comment"
       ]
     }],
+    "string-quotes": "single"
   }
 }


willdurand:
that won't be enough, please see: https://github.com/mozilla/addons-frontend/tree/prettier-scss (last commit(s))

championshuttler:
Hi @willdurand i saw your commits on `prettier-scss` branch, you applied prettier again so I tried to do the same here on my forked repo https://github.com/championshuttler/addons-frontend/commit/c054b25235ea9737e90363a06274c56c2ae458a0 . Is it looking fine? I tested after that comment with yarn on this branch and it is not showing any stylelint errors after that

willdurand:
@championshuttler your best chance is to checkout my branch I think

PR #5439, opened 2018-06-27
Use user data when the API does not return the updated data right away after a user profile update

### LINKED ISSUE #5219: 'About:addons' notification is unchecked when the user accesses the edit profile screen right after choosing this subscription
STR:

1. Log in to AMO -dev with a developer account
2. Click on Edit My Profile
3. Scroll down to the Notifications sections and opt-in for the 'about:addons' newsletter
4. Click on Update my profile
5. Once you are redirected to the View profile page, select again Edit my profile
6. Check the options selected in the notifications section 

Actual result:
The 'about:addons' newsletter appears deselected on the Edit profile page, while the subscription on the email preference center has been registered.  

Expected result:
The 'about:addons' newsletter should appear selected on the Edit profile page

Notes:
- the checkbox will be marked if you refresh the Edit profile page
- the same scenario applies for the reverse action, i.e. unchecking the 'about:addons' newsletter from the Edit profile page

basket notif

### LINKED ISSUE #5252: The updated profile photo and add-on custom icon is displayed only after refreshing page - the previous image is displayed instead
STR:
1. Load AMO dev and go to Edit profile page to replace the existing profile picture  but make sure it's >3MB (I attached the samples I used)
2. Update profile
3. Observe the image displayed in View profile
4. Click the Edit profile button and observe the image

Expected result:
The new profile picture is displayed.

Actual result:
The previous picture is displayed. Only after page is refreshed the new image is visible.

Notes:
This issue is reproducible on AMO dev with FF60(Win10). Also because it's an intermittent issue and I've noticed that in most cases is reproducible with larger than 3MB images I attached my samples.
It is not reproducible on android or with old pages.

previous image displayed

Sometimes the image is not available - page must be refreshed
no image

[above 3MB.zip](https://github.com/mozilla/addons-frontend/files/2084278/above.3MB.zip)

willdurand:
Fix #5252
Fix #5219

---

The idea is: because we have the uploaded file and we are able to display this picture in the edit page, why not using it after having successfully submitted the form?

Same for the Basket integration issue: we now what the user has chosen, so let's use it. Because it is only overridden in the `edit` saga, I believe it is quite safe.

---

TODO:

- [x] add tests for the notification

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -328,6 +328,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
       notifications,
       occupation,
       picture,
+      pictureData,
       username,
     } = this.state;
 
@@ -338,6 +339,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
         errorHandlerId: errorHandler.id,
         notifications,
         picture,
+        pictureData,
         userFields: {
           biography,
           display_name: displayName,
--- a/src/amo/reducers/users.js
+++ b/src/amo/reducers/users.js
@@ -150,6 +150,7 @@ type UpdateUserAccountParams = {|
   errorHandlerId: string,
   notifications: NotificationsUpdateType,
   picture: File | null,
+  pictureData: string | null,
   userFields: UserEditableFieldsType,
   userId: UserId,
 |};
@@ -163,18 +164,31 @@ export const updateUserAccount = ({
   errorHandlerId,
   notifications,
   picture,
+  pictureData,
   userFields,
   userId,
 }: UpdateUserAccountParams): UpdateUserAccountAction => {
   invariant(errorHandlerId, 'errorHandlerId is required');
   invariant(notifications, 'notifications are required');
-  invariant(picture !== undefined, 'picture is required');
   invariant(userFields, 'userFields are required');
   invariant(userId, 'userId is required');
 
+  invariant(picture !== undefined, 'picture is required');
+
+  if (picture) {
+    invariant(pictureData, 'pictureData is required when picture is present');
+  }
+
   return {
     type: UPDATE_USER_ACCOUNT,
-    payload: { errorHandlerId, notifications, picture, userFields, userId },
+    payload: {
+      errorHandlerId,
+      notifications,
+      picture,
+      pictureData,
+      userFields,
+      userId,
+    },
   };
 };
 
--- a/src/amo/sagas/users.js
+++ b/src/amo/sagas/users.js
@@ -85,6 +85,9 @@ export function* updateUserAccount({
         );
         if (index !== -1) {
           allNotifications[index].enabled = notifications.announcements;
+          log.debug(
+            'Optimistically set user value for "announcements" notification',
+          );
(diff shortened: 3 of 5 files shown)


willdurand:
@kumar303 I'd like to have your input on this idea. What do you think?

It is just a rough idea, but it definitely works. As I said in the PR description, it is restricted to the "edit" action, so a pretty small error surface I believe. And also, I did not think about all the edge cases but tried some scenarios manually and did not trigger any error or weird behavior.

src/amo/reducers/users.js
```diff
@@ -163,18 +164,31 @@ export const updateUserAccount = ({
...
   invariant(errorHandlerId, 'errorHandlerId is required');
   invariant(notifications, 'notifications are required');
-  invariant(picture !== undefined, 'picture is required');
   invariant(userFields, 'userFields are required');
   invariant(userId, 'userId is required');
 
+  invariant(picture !== undefined, 'picture is required');
+
+  if (picture) {
+    invariant(pictureData, 'pictureData is required when picture is present');
```
kumar303: Did you mean 'pictureData is required when picture is NOT present' ?

src/amo/sagas/users.js
```diff
@@ -51,6 +58,14 @@ export function* updateUserAccount({
       ...userFields,
     });
 
+    if (picture) {
```
kumar303: Did you mean `if (pictureData)` ?

willdurand:
Added tests, so should be good now.

I re-add a note about `pictureData` here: it is only required when `picture` is supplied, so either the two parameters are provided or none. `picture` is the uploaded file (binary) that will be sent to the API while `pictureData` is the string representation (base64) used to display the preview of the image (initially, in the edit form only).

src/amo/components/UserProfileEdit/index.js
```diff
@@ -328,6 +328,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
       notifications,
       occupation,
       picture,
+      pictureData,
```
kumar303: This code would make more sense to me if these were named `pictureURL` and `pictureDataURL` but no need to change them now -- just an observation.

src/amo/sagas/users.js
```diff
@@ -60,6 +75,19 @@ export function* updateUserAccount({
...
+      if (typeof notifications.announcements !== 'undefined') {
+        // The Salesforce integration is asynchronous and takes a lot of time
+        // so we set the notification to whatever the user has chosen,
+        // otherwise we would display the wrong notification value.
+        // See: https://github.com/mozilla/addons-frontend/issues/5219
+        const index = allNotifications.findIndex(
+          (notification) => notification.name === 'announcements',
+        );
+        if (index !== -1) {
+          allNotifications[index].enabled = notifications.announcements;
```
kumar303: Since this is an edge case which only applies to `announcements`, it would be helpful to log a message saying that the client optimistically set this value (or something).

src/amo/components/UserProfileEdit/index.js
willdurand: I don't think these information are URLs _per se_ though.

Issue #2771, opened 2017-07-12
Add a permalink to a specific review / single review page


tofumatt:
Use the time of review in the `ReviewList` as a permalink to the specific review, which we'll need a simple page for.

grahamperrin:
Hi! Looking good with the go-live. Absence of permalinks was one of just three significant issue that I noticed within the first few minutes. 

Highlighted, pink (with Firefox using a SeaMonkey user-agent string, for the traditional view of things): 

![2017-11-02 17 48 35 conex reviews add-ons for firefox](https://user-images.githubusercontent.com/192271/32342612-cb159826-bff8-11e7-8543-e64e4b654f04.png)

grahamperrin:
> … specific review, which we'll need a simple page for.

Should I/we raise a separate issue for that? For tracking purposes; to be labelled [component: legacy pages](https://github.com/mozilla/addons-frontend/labels/component%3A%20legacy%20pages).

eviljeff:
The legacy site uses the Rating|Review id in the permalink, so that will need to work as well / instead of a timestamp.

bobsilverberg:
I feel like we discussed this when discussing what to include in the Q3 project, but I don't recall if we came up with any solutions. Is our goal to have a page for a single review, like on the legacy site, or is it to bring the user to the list of reviews for the add-on in question, with the linked review at the top (or anchored in some way)? If the latter there's an issue that a specific review could end up on any page of the reviews listing for an add-on, so it might be tricky to tell which page to show.

For reference, here's the page that shows a single review from the legacy site:

![screenshot 2018-08-29 11 23 58](https://user-images.githubusercontent.com/142755/44797788-0ac96580-ab7e-11e8-8047-b736a4a21f5d.png)

Pinging @pwalm for input.

kumar303:
> bring the user to the list of reviews for the add-on in question, with the linked review at the top

I think this would be the most useful view. When you click on a permalink for a review, you want to see the review but it would also be helpful to see a card underneath that of all other reviews. Rather than dealing with anchors, I say just put the permalinked review at the top in its own card.

pwalm:
Proposed change: The permalink to the review is a page with the review broken out up top into its own block, with all others below for context.
<img width="1482" alt="screen shot 2018-08-31 at 2 41 45 pm" src="https://user-images.githubusercontent.com/6797635/44937009-08a50980-ad2c-11e8-9821-bb27a249b107.png">

[...]

PR #5444, opened 2018-06-28
Create UserReview component

### LINKED ISSUE #5416: Create a `UserReview` component for `AddonReviewListItem` and `UserProfile`
From https://github.com/mozilla/addons-frontend/pull/5373#discussion_r198266615:

> This should move to a UserReview component and should include any styles that were defined in AddonReviewListItem.
>
>Ideally, AddonReviewListItem should be refactored to use UserReview but that could come in a separate issue since it won't be a small task.

willdurand:
Fix #5416

---

This PR introduces a new `UserReview` component to avoid duplicating code.

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -31,12 +31,7 @@ import {
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
-import {
-  nl2br,
-  removeProtocolFromURL,
-  sanitizeHTML,
-  sanitizeUserHTML,
-} from 'core/utils';
+import { removeProtocolFromURL, sanitizeUserHTML } from 'core/utils';
 import Button from 'ui/components/Button';
 import Card from 'ui/components/Card';
 import CardList from 'ui/components/CardList';
@@ -45,7 +40,7 @@ import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
 import Rating from 'ui/components/Rating';
 import UserAvatar from 'ui/components/UserAvatar';
-import UserRating from 'ui/components/UserRating';
+import UserReview from 'ui/components/UserReview';
 import type { AppState } from 'amo/store';
 import type { UserReviewType } from 'amo/actions/reviews';
 import type { UserType } from 'amo/reducers/users';
@@ -206,28 +201,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
       >
         <ul>
           {reviews.map((review) => {
-            const reviewBodySanitized = sanitizeHTML(nl2br(review.body), [
-              'br',
-            ]);
+            const byLine = (
+              <Link
+                title={i18n.gettext('Browse the reviews for this add-on')}
+                to={`/addon/${review.addonSlug}/reviews/`}
+              >
+                {i18n.moment(review.created).fromNow()}
+              </Link>
+            );
 
             return (
               <li key={String(review.id)}>
-                <div className="AddonReviewListItem">
-                  <p
-                    className="AddonReviewListItem-body"
-                    // eslint-disable-next-line react/no-danger
-                    dangerouslySetInnerHTML={reviewBodySanitized}
-                  />
-                  <div className="AddonReviewListItem-byline">
-                    <UserRating styleSize="small" review={review} readOnly />
-                    <Link
-                      title={i18n.gettext('Browse the reviews for this add-on')}
-                      to={`/addon/${review.addonSlug}/reviews/`}
-                    >
-                      {i18n.moment(review.created).fromNow()}
-                    </Link>
-                  </div>
-                </div>
+                <UserReview review={review} byLine={byLine} />
               </li>
             );
           })}
--- a/src/ui/components/UserReview/index.js
+++ b/src/ui/components/UserReview/index.js
@@ -1,0 +1,54 @@
+/* @flow */
+import * as React from 'react';
+
+import { nl2br, sanitizeHTML } from 'core/utils';
+import LoadingText from 'ui/components/LoadingText';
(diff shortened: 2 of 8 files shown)


src/amo/components/UserProfile/index.js
```diff
@@ -206,28 +201,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
...
-                    <UserRating styleSize="small" review={review} readOnly />
-                    <Link
-                      title={i18n.gettext('Browse the reviews for this add-on')}
-                      to={`/addon/${review.addonSlug}/reviews/`}
-                    >
-                      {i18n.moment(review.created).fromNow()}
-                    </Link>
-                  </div>
-                </div>
+                <UserReview review={review} byLine={byLine} />
```
kumar303: I tried this and it actually works quite well:

```diff
diff --git a/src/amo/components/UserProfile/index.js b/src/amo/components/UserProfile/index.js
index 6d0f2e8a8..11e48d01c 100644
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -7,6 +7,7 @@ import { compose } from 'redux';
 
 import { fetchUserReviews } from 'amo/actions/reviews';
 import AddonsByAuthorsCard from 'amo/components/AddonsByAuthorsCard';
+import AddonReviewListItem from 'amo/components/AddonReviewListItem';
 import Link from 'amo/components/Link';
 import NotFound from 'amo/components/ErrorPage/NotFound';
 import ReportUserAbuse from 'amo/components/ReportUserAbuse';
@@ -212,7 +213,7 @@ export class UserProfileBase extends React.Component<InternalProps> {
 
             return (
               <li key={String(review.id)}>
-                <UserReview review={review} byLine={byLine} />
+                <AddonReviewListItem review={review} location={location} />
               </li>
             );
           })}
```

- Editing mostly works. I had to refresh the page to see it. Selecting stars isn't working.
- Developer responses are shown which is nice because it allows the developer to clarify things.
- The *flag review* menu is there but we could add a prop to hide it. If we were to show user reviews to *other* site users then the *flag review* menu would make sense here.
- This makes the byline consistent but I'm not sure if that's what we want.

Just a thought. It could be done in future issues.

kumar303: Oh, actually the *flag review* menu makes perfect sense here for developer replies.

src/amo/components/UserProfile/index.js
```diff
@@ -206,28 +201,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
...
           {reviews.map((review) => {
-            const reviewBodySanitized = sanitizeHTML(nl2br(review.body), [
-              'br',
-            ]);
+            const byLine = (
+              <Link
+                title={i18n.gettext('Browse the reviews for this add-on')}
+                to={`/addon/${review.addonSlug}/reviews/`}
+              >
+                {i18n.moment(review.created).fromNow()}
```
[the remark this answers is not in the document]
willdurand: > Also, it is surprising to me that each review links to the browse reviews page (will users be able to figure this out?).

That's a pretty common pattern (it would be better to directly link to the review though).

> Anyway, I'm just thinking out loud. I don't have a concrete change request here but it feels like some part of the abstraction is missing.

I did not find a better way to fix that, but I can think about it again.

> As an aside, is the UX spec for user profile reviews out of date? It doesn't match what we have.

mmh yes, I think this has changed a bit but we also did not implement the latest UI for the add-on card/list (this is a global change).

src/amo/components/UserProfile/index.js
willdurand: mmh, so you'd like to have write access to the reviews on the user profile page?

willdurand:
Should be ready for another look if UI suggestions were ideas for later.

willdurand:
rebased.

src/amo/components/UserProfile/index.js
kumar303: Sure, I don't see why not. The feature has been requested before. And it would keep things consistent across the site.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -807,35 +802,21 @@ describe(__filename, () => {
...
 
-    expect(root.find('.AddonReviewListItem')).toHaveLength(reviews.length);
-
-    const firstReview = root.find('.AddonReviewListItem');
+    expect(root.find(UserReview)).toHaveLength(reviews.length);
 
-    expect(firstReview.find('.AddonReviewListItem-body').html()).toContain(
-      fakeReview.body,
-    );
+    const byLine = shallow(root.find(UserReview).prop('byLine'), {
```
kumar303: If you change this to `shallow(<div>{root.find(UserReview).prop('byLine')}</div>)` I think it will work without needing the `store` context.

src/ui/components/UserReview/index.js
```diff
@@ -0,0 +1,54 @@
...
+        dangerouslySetInnerHTML={sanitizeHTML(nl2br(review.body), ['br'])}
+      />
+    );
+  }
+
+  return (
+    <div className="UserReview">
+      {body}
+      <div className="UserReview-byLine">
+        {review && !isReply ? (
```
kumar303: Since `isReply` no longer makes sense in this context, I think it would be better to name it more explicitly like `showRating` or something.

tests/unit/amo/components/TestUserProfile.js
willdurand: mmh, nope because the `byLine` is the `Link` component, which requires a `store`.

willdurand:
> I think we should turn AddonReviewListItem into UserReview so that the user profile lists reviews consistently, including developer replies and edit links. This could be filed as a follow-up issue. Otherwise, I know for sure we will need to redesign the byline logic when we add permalinks: #2771 At the very least, this byline abstraction isn't quite right because there is still some duplicated logic.

Filed #5598.

PR #5445, opened 2018-06-28
Pass errorHandler from UserProfile to AddonsByAuthorsCards

### LINKED ISSUE #5442: Navigating to a non existent page in More add-on by author cards in the User profile page will lock the cards in a loading state
STR:

1. Load AMO -dev
2. Open a user profile page that lists both add-ons and themes - see https://addons-dev.allizom.org/en-US/firefox/user/AlexandraMoga1/?page_e=7&page_t=6
3. Set the page query for the two cards to a non-existent page (i.e. 7 for add-ons and 6 for themes for this case)

Actual result:
The two cards are locked in a loading state when a non-existent page is requested

Expected result:
Display (maybe) a Page not found on the card that is causing the error to appear

Notes:
- reproduced on AMO -dev with FF60, WIn10x64

[1] Page when both cards are requesting an invalid page
screen shot 2018-06-28 at 09 11 21-fullpage

[2] Page when only on card is requesting an invalid page

image

willdurand:
Fix #5442

---

This PR allows to return a 404 when extensions and/or themes on a user
profile cannot be loaded (and the API returns a 404 when trying to fetch
these add-ons), likely due to an invalid query parameter value.

Passing the `errorHandler` from the `UserProfile` component to the
`AddonsByAuthorsCard` components is required to make sure we return a
`NotFound` component on the server, thus returning a real `404` HTTP
response.

I also decided to hide the `AddonsByAuthorsCard` when an error occurs,
because it is better than being stuck in a loading state. On an add-on
detail page, it is "secondary data" so does not matter much I guess,
and on the user profile, the `UserProfile` will display the error.

I don't know if the added tests are enough or if we could find a better
patch, but I don't really see how we could do better here.

### DIFF
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -28,6 +28,7 @@ import {
 } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
+import log from 'core/logger';
 import type {
   AddonsByAuthorsState,
   FetchAddonsByAuthorsParams,
@@ -45,6 +46,7 @@ type Props = {|
   authorDisplayName: string,
   authorUsernames: Array<string>,
   className?: string,
+  errorHandler?: ErrorHandlerType,
   forAddonSlug?: string,
   numberOfAddons: number,
   pageParam: string,
@@ -62,7 +64,6 @@ type InternalProps = {|
   addons?: Array<AddonType>,
   count: number | null,
   dispatch: DispatchFunc,
-  errorHandler: ErrorHandlerType,
   i18n: I18nType,
   loading?: boolean,
   location: ReactRouterLocation,
@@ -176,6 +177,8 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       filtersForPagination.sort = SEARCH_SORT_POPULAR;
     }
 
+    invariant(errorHandler, 'errorHandler is required');
+
     this.props.dispatch(
       fetchAddonsByAuthors({
         addonType,
@@ -196,6 +199,7 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       authorDisplayName,
       authorUsernames,
       className,
+      errorHandler,
       i18n,
       loading,
       numberOfAddons,
@@ -204,6 +208,13 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       showSummary,
       type,
     } = this.props;
+
+    invariant(errorHandler, 'errorHandler is required');
+
+    if (errorHandler.hasError()) {
+      log.warn('Captured API Error:', errorHandler.capturedError);
+      return null;
+    }
 
     if (!loading && (!addons || !addons.length)) {
       return null;
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -28,7 +28,6 @@ import {
 } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
-import log from 'core/logger';
 import type {
   AddonsByAuthorsState,
   FetchAddonsByAuthorsParams,
@@ -212,8 +211,7 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
     invariant(errorHandler, 'errorHandler is required');
 
     if (errorHandler.hasError()) {
-      log.warn('Captured API Error:', errorHandler.capturedError);
-      return null;
+      return errorHandler.renderError();
     }
 
     if (!loading && (!addons || !addons.length)) {
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -146,11 +146,9 @@ export class UserProfileBase extends React.Component<InternalProps> {
   }
 
   getURL() {
-    const { user } = this.props;
+    const { params } = this.props;
 
-    invariant(user, 'user is required');
-
-    return `/user/${user.username}/`;
+    return `/user/${params.username}/`;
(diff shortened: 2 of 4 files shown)


src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -205,6 +209,13 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       type,
     } = this.props;
 
+    invariant(errorHandler, 'errorHandler is required');
+
+    if (errorHandler.hasError()) {
```
kumar303: This seems like the wrong thing to do because it is swallowing any error that may have been produced by the sagas in this component. If I remove this if statement, the original bug you set out to fix is still fixed: I see the 404 page.

However, swallowing errors looks like an unrelated bug that you may as well fix. I think you can fix it like this:

```diff
diff --git a/src/amo/components/AddonsByAuthorsCard/index.js b/src/amo/components/AddonsByAuthorsCard/index.js
index e9c395e43..74873c400 100644
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -211,11 +211,6 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
 
     invariant(errorHandler, 'errorHandler is required');
 
-    if (errorHandler.hasError()) {
-      log.warn('Captured API Error:', errorHandler.capturedError);
-      return null;
-    }
-
     if (!loading && (!addons || !addons.length)) {
       return null;
     }
@@ -341,17 +336,20 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
     }
 
     return (
-      <AddonsCard
-        addons={addons}
-        className={classnames}
-        footer={paginator}
-        header={header}
-        loading={loading}
-        placeholderCount={numberOfAddons}
-        showMetadata
-        showSummary={showSummary}
-        type={type}
-      />
+      <React.Fragment>
+        {errorHandler.renderErrorIfPresent()}
+        <AddonsCard
+          addons={addons}
+          className={classnames}
+          footer={paginator}
+          header={header}
+          loading={loading}
+          placeholderCount={numberOfAddons}
+          showMetadata
+          showSummary={showSummary}
+          type={type}
+        />
+      </React.Fragment>
     );
   }
 }

```

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -968,4 +970,31 @@ describe(__filename, () => {
       );
     });
   });
+
+  it('does not render anything when an API error is thrown', () => {
```
kumar303: Why is this needed? It doesn't seem necessary to solve the bug of rendering a 404 page.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -896,6 +919,62 @@ describe(__filename, () => {
...
+      id: 'some-id',
+      dispatch: store.dispatch,
+    });
+
+    const root = renderUserProfile({ errorHandler, store });
+
+    // Simulate an error thrown inside the second AddonsByAuthorsCard.
+    root
+      .find(AddonsByAuthorsCard)
+      .at(1)
```
kumar303: When you have two tests that look the same, it's an indication that you aren't testing the unit of logic in the correct place.

I think you can replace these two tests with one test that makes sure `UserProfile` renders a 404 `NotFound` component given an `errorHandler` with a 404 error. It doesn't matter *where* the error handler gets its error from. The set up could look like:

```js
const { store } = dispatchSignInActions();
  id: 'some-id',
  dispatch: store.dispatch,
});
errorHandler.handle(
  createApiError({
    response: { status: 500 },
    jsonResponse: { message: 'internal server error' },
  })
);

const root = renderUserProfile({ errorHandler, store });
// Check for the error...
```

The reason you don't need to test the `errorComponent` of each `AddonsByAuthorsCard` is because you already have a test to make sure that their `errorComponent` props are coming from `UserProfile`.

src/amo/components/AddonsByAuthorsCard/index.js
kumar303: > I also decided to hide the AddonsByAuthorsCard when an error occurs,
because it is better than being stuck in a loading state.

Ah, I just read this part in your description.

In this case, you should switch it in an if statement, like:

```js
render() {
    return errorHandler.renderHoweverItsCalled();
  } else {
    return <AddonsCard ...>
  }
}
```
I don't think you should hide the error, though.

src/amo/components/UserProfile/index.js
```diff
@@ -385,13 +383,14 @@ export class UserProfileBase extends React.Component<InternalProps> {
...
+          {params &&
+            params.username && (
               <div className="UserProfile-addons-and-reviews">
                 <AddonsByAuthorsCard
                   addonType={ADDON_TYPE_EXTENSION}
-                  authorDisplayName={user.name}
-                  authorUsernames={[user.username]}
+                  authorDisplayName={user ? user.name : params.username}
+                  authorUsernames={[params.username]}
+                  errorHandler={errorHandler}
```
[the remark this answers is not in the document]
willdurand: Yeah, definitely. It is fine here but nothing we can generalize :/

src/amo/components/AddonsByAuthorsCard/index.js
willdurand: >If I remove this if statement, the original bug you set out to fix is still fixed: I see the 404 page.

Yes, I added this after..

tests/unit/amo/components/TestUserProfile.js
willdurand: right, I don't know what I was thinking.

willdurand:
So I believe I addressed all the comments.

tests/unit/amo/components/TestUserProfile.js
kumar303: Heh, no worries. It's always easier to spot when doing a code review.

Issue #5447, opened 2018-06-28
Search tools install buttons looks inactive if opened directly from their URLs


vcarciu:
Steps to reproduce:
1.Go to any search tool opening it directly from a link (https://addons.mozilla.org/en-US/firefox/addon/alltheweb-news/?src=search)
2. Observe the install button

Expected results:
Button is active

Actual results:
Button is inactive

NOTES:
Not reproducible if opening any search tool from More-Search Tools in AMO

AlexandraMoga:
@kumar303 note that this is not a new issue. We have been seing this for a while and had an issue filed some time ago, which is still open. See issue https://github.com/mozilla/addons-frontend/issues/4047

I think this one is a duplicate considering that the old issue has more background.

PR #5450, opened 2018-06-28
Never render opensearch install buttons on the server

### LINKED ISSUE #4047: 'Add to Firefox' button looks inactive on Search Tool details page

kumar303:
~~Fixes https://github.com/mozilla/addons-frontend/issues/5447~~

Fixes #4047

**Problem**: On the server, opensearch add-ons are always incompatible because they require `window.external.AddSearchProvider`. This renders the install button with a disabled class. When the client code renders, it looks at `window.external` and changes the class of the button. Apparently, (for performance reasons, I guess) class values [are not updated in the DOM](https://github.com/facebook/react/issues/10591#issuecomment-326635627) when hydrating server code.

**Solution**: Well, I dunno, all I could think of was to never render these types of buttons on the server. I'm not sure how this effects SEO but it's probably fine. This approach introduces a new DOM node on the client so it will always render.

This was a breaking change introduced in React 16.


tofumatt:
😱 Wow, nice find.

PR #5461, opened 2018-06-29
Only run Prettier on modified files

### LINKED ISSUE #5448: Only run prettier on changed files
For contributors who are relying on the `yarn prettier` command, it will be a slow and annoying. We should give them something like `yarn prettier-dev` which will only run prettier on the changed files.

It looks like [Pretty-Quick](https://github.com/azz/pretty-quick#pretty-quick) will do this for us.

SeanPrashad:
Fixes #5448 - only run Prettier on changed files

This PR introduces [`pretty-quick`](https://www.npmjs.com/package/pretty-quick), an `npm` package that will only apply [Prettier](https://github.com/prettier/prettier#intro) to modified files that differ from the `master` branches version (with the CLI argument `--branch master`).

What this means for us is that every contributor now must run `yarn prettier-dev` in order to format the files that they have modified at the **end** of their PR submission.

See [here](https://giphy.com/gifs/42xwX28NURrNAkKch4/fullscreen) for an (outdated) example!


package.json
```diff
@@ -268,6 +269,7 @@
     "postcss-loader": "2.1.4",
     "potools": "^0.2.0",
     "prettier": "1.13.6",
+    "pretty-quick": "1.6.0",
```
kumar303: We pinned `prettier` to an exact version so that all devs are using the same one. I guess we should do the same here which means you'll need to add this to the exception list in the test suite. The failing test should have info about it.

package.json
```diff
@@ -26,6 +26,7 @@
     "lint": "npm run eslint && npm run stylelint",
     "prettier": "prettier --write '**'",
     "prettier-ci": "npm run prettier -- --list-different || (echo '\n\nThis failure means you did not run `yarn prettier` before committing\n\n' && exit 1)",
+    "prettier-dev": "pretty-quick --staged",
```
kumar303: Why `--staged`? The `--branch` mode sounds like more of what we would want since I think it will check all files in the pull request. This might be better than `--staged` since you won't have to remember to run the command before committing your files. In other words, you could safely run it at the _end_ of your work in the pull request.

SeanPrashad: Good idea - I've switched over the cli arg to `--branch master` which should compare the current branches changes vs the master branch and apply prettier to any files that have changed.

SeanPrashad:
I've updated the PR with the two suggestions you had @kumar303. Feel free to pull it down and try some testing when you get a moment!

SeanPrashad:
Travis is reporting something wonky with Selenium:

```
selenium.common.exceptions.TimeoutException: Message: AddOnInstallComplete was not shown.
.tox/discopane-ui-tests/lib/python3.6/site-packages/selenium/webdriver/support/wait.py:80: TimeoutException
------------------------------- pytest-selenium --------------------------------
Driver log: /tmp/pytest-of-travis/pytest-0/test_theme_installs0/driver.log
URL: https://example.com:4000/en-US/firefox/discovery/pane/63.0/linux/normal
WARNING: Failed to gather log types: Message: GET /session/e8e9b99a-16dd-4cf5-bb55-c1c3345b4ef0/log/types did not match a known command
===================== 1 failed, 5 passed in 42.12 seconds ======================
ERROR: InvocationError for command '/home/travis/build/mozilla/addons-frontend/.tox/discopane-ui-tests/bin/pytest --driver Firefox tests/ui/tests/test_discopane.py' (exited with code 1)
```

Should I try to force another build and see if it fails again? I don't think this happened because of my changes..

willdurand:
You probably want to rebase your branch, we merged a fix for these tests recently.

SeanPrashad:
Thanks for that tip @willdurand - the rebase fixed things!

package.json
```diff
@@ -26,6 +26,7 @@
     "lint": "npm run eslint && npm run stylelint",
     "prettier": "prettier --write '**'",
     "prettier-ci": "npm run prettier -- --list-different || (echo '\n\nThis failure means you did not run `yarn prettier` before committing\n\n' && exit 1)",
```
kumar303: Whoops, this line should give `yarn prettier-dev` as a hint for those who do not RTFM.

tests/unit/TestPackageJson.js
```diff
@@ -12,7 +12,8 @@ const packageJson = JSON.parse(
 
 // postcss-loader: https://github.com/mozilla/addons-frontend/issues/4592
 // prettier: https://github.com/mozilla/addons-frontend/issues/5312
-const skipDevDeps = ['postcss-loader', 'prettier'];
+// pretty-quick: https://github.com/mozilla/addons-frontend/pull/5461/files#r199189927
```
kumar303: This comment doesn't need to link to a PR. Just explain it like this:

```js
// prettier and pretty-quick: we pin these so that all developers format code with the exact same version
```

tests/unit/TestPackageJson.js
```diff
@@ -12,7 +12,8 @@ const packageJson = JSON.parse(
 
 // postcss-loader: https://github.com/mozilla/addons-frontend/issues/4592
 // prettier: https://github.com/mozilla/addons-frontend/issues/5312
```
kumar303: Since the new comment now covers both `prettier` and `prettier-quick`, you can delete this one.

package.json
SeanPrashad: Never knew what [RTFM](https://www.urbandictionary.com/define.php?term=RTFM) mean't until today 😅

SeanPrashad:
Thanks for the feedback - I've addressed all of the nits which means this should be good to go!

Issue #4262, opened 2018-01-30
Investigate A/B testing Component or framework


muffinresearch:
This issue is for investigating prior art for an A/B testing solution for React components. 

For our initial use-case we'd like to try and do this for the Homepage switch to cards in place of the current homepage hero content.

## Example libs to look at:

* react-ab https://github.com/olahol/react-ab (looks good because it's simple and we can easily tie it in with our current GA instrumentation.
* Afaik [Optimizely](https://www.optimizely.com/) has been used elsewhere at mozilla, but I don't know much about how it might work.
* [Amplitude](https://auth.mozilla.auth0.com/samlp/ql6V6DHXglhTCq9qe6DPdV3eAW2da44F) is used by Firefox Accounts and Pocket. I *think* it can be used for A/B testing but I'm not sure. There is a [presentation](https://docs.google.com/presentation/d/1dhC5JsPDfi7LSQVJ53UmHGzSmUWAHg8Y9uf_8LZrg18/edit#slide=id.g2af8fbdae6_0_69)!
* [INSERT OTHERS HERE]

## Arbitrary requirements:

* Should ideally supprt GA - anything else may require legal + privacy unless already in use on other mozilla properties.
* Should be CSP compliant - optimizely historically had issues with this, though they may be fixed already.
* Needs to work within our Universal App setup.

willdurand:
[react-ab](https://github.com/olahol/react-ab) looks interesting but we might want to contribute code to it in order to add [react-cookie](https://github.com/reactivestack/cookies/tree/master/packages/react-cookie).

kumar303:
A custom A/B solution was added in [the TAAR patch](https://github.com/mozilla/addons-frontend/pull/4942/files#diff-cd4abfb75699370a11298ce429e85e10R63) and then removed in https://github.com/mozilla/addons-frontend/pull/5467

willdurand:
We are adding a a/b HOC leveraging the TAAR patch linked above.

willdurand:
Looks like this issue has been fixed.

PR #5467, opened 2018-07-01
Remove the A/B test from the TADAr so that recommendations are 100% of traffic

### LINKED ISSUE #5408: Remove the A/B test from the TADAr so that recommendations are 100% of traffic
Ahead of making incremental changes to the underlying TADAr (Telemetry Aware Details Add-ons Recommendations) engine we'd like to turn of the fallback part of the A/B test so that recommendations are shown to all users (the fall-back will still apply in cases where TAAR errors).

bobsilverberg:
Fixes #5408 

This ensures that 100% of users will request actual recommendations from TADAr. As per the comment at https://github.com/mozilla/addons-frontend/issues/5408#issuecomment-400971189 we will not be making any changes to the data sent to GA at this time.

These changes are currently isolated to the `AddonRecommendations` component. Although we could potentially make these changes at a lower level, removing the `recommended` property from `FetchRecommendationsParams`, I think it's best to just leave that as is for now, as there is a suggestion we may be doing further experiments with this feature in the future.

### DIFF
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -110,7 +110,6 @@ export class AddonRecommendationsBase extends React.Component<Props> {
       fetchRecommendations({
         errorHandlerId: this.props.errorHandler.id,
         guid,
-        recommended: true,
       }),
     );
   }
--- a/src/amo/components/AddonRecommendations/index.js
+++ b/src/amo/components/AddonRecommendations/index.js
@@ -2,7 +2,6 @@
 import makeClassName from 'classnames';
 import invariant from 'invariant';
 import * as React from 'react';
-import defaultCookie from 'react-cookie';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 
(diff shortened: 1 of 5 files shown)


bobsilverberg:
@kumar303 I know I've asked you to review 3 PRs now, but this one is a P2 which we're really hoping to land this week, so if you don't mind could you prioritize this one?

src/amo/components/AddonRecommendations/index.js
```diff
@@ -133,14 +105,12 @@ export class AddonRecommendationsBase extends React.Component<Props> {
...
-  cohort: CohortName;
-
-  dispatchFetchRecommendations({ guid, recommended }: Object) {
+  dispatchFetchRecommendations(guid: string) {
     this.props.dispatch(
       fetchRecommendations({
         errorHandlerId: this.props.errorHandler.id,
         guid,
-        recommended,
+        recommended: true,
```
kumar303: I think it would be nicer if this parameter defaulted to `true` in `fetchRecommendations()` so that the caller can leave it undefined.

src/amo/components/AddonRecommendations/index.js
```diff
@@ -29,53 +28,33 @@ import type { DispatchFunc } from 'core/types/redux';
...
-    randomizer: Math.random,
     recommendations: null,
     tracking: defaultTracking,
   };
 
   componentDidMount() {
-    const { addon, cookie, randomizer, recommendations, tracking } = this.props;
-
-    // Set a cohort for the experiment.
-    this.cohort = cookie.load(TAAR_COHORT_COOKIE_NAME);
```
kumar303: It looks like this turned out to be simpler than `react-ab` so that's nice. Simple is good. However, the next time we want to A/B test something we will have to start from scratch. I would like to see a more general purpose solution and it's fine to keep an issue open for that. I guess the tracker is https://github.com/mozilla/addons-frontend/issues/4262 ?

Any reason why the cohort/randomizer/GA code wasn't wrapped up in a higher order component or similar type of helper? At the very least, a function like `getCohort()` to wrap up the cookie logic would be useful.

bobsilverberg: There was no particular reason that it wasn't encapsulated in an HOC, other than that I didn't think to do it. We actually started with an implementation that used `react-ab`, and then I realized that we actually didn't need that at all. Then it just kind of evolved from there. Agreed that in the future, if we need to do something similar, we should try to abstract it, and we can always look through the history of this component to see what we did before.

willdurand: I second Bob's comment. We decided to keep things simple and we did not know what to expect from this experiment. Now that it worked, it would be interesting to create a HOC for further use. We should probably keep this idea in an issue.

PR #5477, opened 2018-07-02
Support allowing collaborators to edit a Mozilla collection

### LINKED ISSUE #5163: Support allowing collaborators to edit a Mozilla collection
Depends on ~~https://github.com/mozilla/addons-server/issues/8509~~

As documented in the above issue, there will be one permission that indicates that a user can edit any Mozilla-owned collection, as well as a permission that indicates that a user can edit one particular Mozilla-owned collection.

bobsilverberg:
Fixes #5163

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -274,8 +274,14 @@ export const ADDONS_REVIEW = 'Addons:Review';
 export const THEMES_REVIEW = 'Personas:Review';
 // Can view statistics for all addons, regardless of privacy settings.
 export const STATS_VIEW = 'Stats:View';
-// Can edit all collections.
-export const COLLECTIONS_EDIT = 'Collections:Edit';
+// Can edit all Mozilla-owned collections.
+export const MOZILLA_COLLECTIONS_EDIT = 'Admin:Curation';
+// The username which corresponds to Mozilla-owned collections.
+export const MOZILLA_COLLECTIONS_USERNAME = 'mozilla';
+// Can edit the special Featured Themes collection.
+export const FEATURED_THEMES_COLLECTION_EDIT = 'Collections:Contribute';
+// The slug for the special Featured Themes collection.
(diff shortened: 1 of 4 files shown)


src/core/constants.js
```diff
@@ -274,8 +274,14 @@ export const ADDONS_REVIEW = 'Addons:Review';
 export const THEMES_REVIEW = 'Personas:Review';
 // Can view statistics for all addons, regardless of privacy settings.
 export const STATS_VIEW = 'Stats:View';
-// Can edit all collections.
-export const COLLECTIONS_EDIT = 'Collections:Edit';
```
kumar303: Are you sure we don't need this anymore? It seems like it was intended for superuser collection editors that aren't admins.

bobsilverberg: We got rid of that permission, which is the reason for this issue. Only the two permissions being implemented in this issue should allow someone who is not a global admin to edit a collection which they do not own. This change has already been made on the back end (disabling the `Collections:Edit` permission.

bobsilverberg:
Also note that this time I did change user permissions and test interactively with the site locally to ensure that the links were either displayed or hidden as expected with different permissions.

src/core/constants.js
```diff
@@ -274,8 +274,14 @@ export const ADDONS_REVIEW = 'Addons:Review';
...
+// Can edit all Mozilla-owned collections.
+export const MOZILLA_COLLECTIONS_EDIT = 'Admin:Curation';
+// The username which corresponds to Mozilla-owned collections.
+export const MOZILLA_COLLECTIONS_USERNAME = 'mozilla';
```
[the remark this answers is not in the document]
kumar303: We only put `_config` in `defaultProps` for testing but there are other ways to test it. I say don't worry about it for this patch. If development becomes a burden we could consider moving these to a config.

bobsilverberg:
Thanks @kumar303

Issue #4655, opened 2018-03-28
Add new routes for new collections UI


bobsilverberg:
We'll need new routes in nginx to route collection requests to the new front-end. Here is what we need:

- [ ] Edit collection: collections/:user/:slug/edit/ (landed in #4623)
- [ ] Create collection: collections/add/ (depends on #4003)
- [ ] My collections: collections/

These should both be enabled on dev only for now, and I will update this issue when we want them turned on for stage and prod.

bqbn:
Both endpoints, `collections/:user/:slug/edit/` and `collections/add/` are routed to the mobile site on -dev.

Please test.

Please ping me when this is ready for -stage and/or -prod.

bobsilverberg:
@bqbn Could you please enable the third route listed above `collections/` on -dev?

bqbn:
`collections/` should now be routed to the mobile site on -dev.

bobsilverberg:
Thanks @bqbn!

bobsilverberg:
@bqbn These are also now ready to be configured on stage. Could you please enable the rules on stage for: 

- collections/:user/:slug/edit/
- collections/add/
- collections/

Thanks!

bqbn:
The said 3 endpoints are now routed to the mobile site on -stage.

bobsilverberg:
Thanks @bqbn

bobsilverberg:
@bqbn We also need these routes enabled on prod to coincide with this week's push. Can you please take care of that? The rules are: 

- collections/:user/:slug/edit/
- collections/add/
- collections/

Thanks!

bqbn:
PR https://github.com/mozilla-services/cloudops-deployment/pull/2257 merged.

Please add a note to the etherpad as a reminder for the push day.

bobsilverberg:
> Please add a note to the etherpad as a reminder for the push day.

Done, thanks @bqbn!

AlexandraMoga:
Verified fixed in prod

PR #5480, opened 2018-07-03
Enable the new collections UI on stage

bobsilverberg:
I have also requested the nginx rules be updated via #4655

### DIFF
--- a/config/stage-amo.js
+++ b/config/stage-amo.js
@@ -25,5 +25,6 @@ module.exports = {
     },
   },
 
+  enableNewCollectionsUI: true,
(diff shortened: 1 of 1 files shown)


willdurand:
> I have also requested the nginx rules be updated via #4655

can we merge this after the tag to avoid enabling two features on -stage at once or is it ok to do it so?

bobsilverberg:
> can we merge this after the tag to avoid enabling two features on -stage at once or is it ok to do it so?

I _think_ it's okay to do so, but I guess we can hold off until the next release if you'd prefer

Issue #5343, opened 2018-06-20
Control local component state with Redux


kumar303:
It's convenient to use `this.setState()` to change local component state. However, all uses of `setState()` will break Redux. By that I mean we will lose the ability for Redux to track our state changes. Here are things that will be broken:

* The Redux devtool will not let you time travel between state changes if you use `setState()`
* A future hypothetical error reporter (which would [send all state changes to Sentry](https://github.com/mozilla/addons-frontend/issues/2176)) would not know about `setState()` so it would not be very useful

**All components should use Redux instead of setState()**. 

There is one exception: capturing keystrokes in a text field. For this case, we should still use `setState()` but be sure to store the final typed value in Redux. This exception is because triggering all reducers on every keystroke could hurt performance.

**But, wait! I *love* `setState()` and I don't want to create reducer/action boilerplate just for one component.** 

I feel you. Thus, this issue will focus on an abstraction that lets you use something like `setState()` but with Redux under the hood. I looked at existing libs for this but they don't support SSR.

See https://github.com/mozilla/addons-frontend/issues/1250 for converting all of `setState()` to Redux.

ioanarusiczki:
Verified the notes on AMO dev and AMO stage with FF61(Win10 and Android 8.0):  add, edit, delete - it seems to work without problems.

PR #5484, opened 2018-07-03
Control local component state with Redux

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5343

It would be great to get some feedback on this (the more eyes the better). I've only replaced `setState()` with `setUIState()` in one component, `EditableCollectionManager` to get a feel for it.

This is only a small example so it's not dramatic but here are some screencasts in the Redux tool to illustrate how switching to `setUIState()` will open up some possibilities for us.

When hiding/showing a note was controlled with `setState()`, it was completely invisible from the Redux tool (and anything else that works with Redux state). You can see that replaying all actions would not show the edit form:

<hr/>
<details>
<summary>Using setState(), the old way</summary>

addons-frontend-coll-edit-before mov

</details>
<hr/>

When notes are controlled with `setUIState()`, it is now possible to fully debug all UI state changes (including the edit form) with the Redux tool:

<hr/>
<details>
<summary>Using setUIState(), the new way</summary>

addons-frontend-coll-edit-after mov

</details>
<hr/>

The actual text input (changing the note text) is still handled by `setState()` in these examples which is why you don't see any typing.


src/core/reducers/uiState.js
```diff
@@ -0,0 +1,129 @@
...
+      ...dispatchProps,
+      setUIState: (change: Object) => {
+        dispatch(setUIState({ id: stateProps.uiStateID, change }));
+      },
+      // This is just for testing to simulate the props you'll
+      // get after a Redux action is dispatched.
+      // Using shallowUntilTarget() and root.update() won't
+      // actually update all of the HOC wrapped components
+      // ... I guess?
+      simulateUIStateProps: ({ store }: {| store: Object |}) => {
```
willdurand: Any change to move this function outside of the HOC? Can't this code live in `applyUIStateChanges()` for instance?

src/core/reducers/uiState.js
```diff
@@ -0,0 +1,129 @@
...
+  invariant(extractId, 'extractId is required');
+  invariant(defaultState, 'defaultState is required');
+
+  const mapStateToProps = (
+    state: {| uiState: UIStateType |},
+    props: Object,
+  ) => {
+    const uiStateID =
+      props.uiStateID || generateId({ fileName, id: extractId(props) });
+    const uiState = state.uiState[uiStateID] || defaultState;
```
willdurand: Do we really need a `defaultState`? I wish we had a `initialState` instead, similar to:

```
constructor(props)

File: src/core/reducers/uiState.js [hunk 3]
```diff
@@ -0,0 +1,129 @@
...
+  return (WrappedComponent) => {
+    return compose(
+      connect(
+        mapStateToProps,
+        undefined,
+        mergeProps,
+      ),
+    )(WrappedComponent);
+  };
+};
```
willdurand: It would be less confusing to move the HOC code in a `withUIState.js` file.

src/core/reducers/uiState.js
kumar303: Let me try to explain the problem from the perspective of `EditableCollectionAddon`.

It was doing this to initialize state:

```js
constructor(props) {
  super(props);
  this.state = {
    editingNote: false,
  };
}
```

I introduced `withUIState({ defaultState: ... })` to mimic this but because Redux state is persistent across all component instances (because `extractId()` only looks at props), I needed this to reset the state for *every* new instance:

```js
componentWillMount() {
  this.props.setUIState({ editingNote: false });
}
```

This allowed it to function like before.

I can think of some different solutions here. Switching to `componentDidMount()` is an obvious solution but setting an initial state really isn't an async operation so let me think of some other approaches.

willdurand: Yes, I figured out that my code snippet above fixed only one part of the problem, but not exactly what I was highlighting. We need something to initialize the state in `uiState` before the first `render()`.

src/core/reducers/uiState.js
kumar303: I tried this but ran into issues with the variables encapsulated by `withUIState()`. I made a late change where `uiStateID` is exposed in props so let me try again now that this computed variable is accessible.

src/core/reducers/uiState.js
kumar303: Ah, makes sense. I'll move it.

src/core/reducers/uiState.js
willdurand: mmh, that might work actually since the props are initialized with `defaultState`. We always have valid data injected by the HOC. By spreading the `stateProps`, the first call to `setUIState()` should add the `defaultState` overrode by `change` into redux. That should reproduce the behavior of `this.state` into the constructor as well as what `setState()` does.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -172,15 +186,24 @@ describe(__filename, () => {
...
     it('shows a populated notes form when the edit button icon is clicked', () => {
       const notes = 'Some notes.';
 
-      const root = render({ notes });
-      expect(root.find('.EditableCollectionAddon-notes-form')).toHaveLength(0);
-      const editButton = root.find(
-        '.EditableCollectionAddon-notes-edit-button',
-      );
-      editButton.simulate('click', createFakeEvent());
+      const root = renderAndEditNote({ notes });
```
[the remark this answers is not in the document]
kumar303: Oh, whoops, this test should still click the icon.

My idea was for any test that works solely in the edit-note state, they should just enter that state directly.

tests/unit/amo/components/TestEditableCollectionAddon.js
```diff
@@ -172,15 +186,24 @@ describe(__filename, () => {
       ).toHaveLength(0);
     });
 
+    it('changes UI state when the comment icon is clicked', () => {
+      const { store } = dispatchClientMetadata();
+      const root = render({ store });
+      applyUIStateChanges({ root, store });
```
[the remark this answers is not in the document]
kumar303: I agree. I'm going to try and think of some other ways to solve it. I'm not totally sure why Enzyme doesn't automatically update props with Redux state but we use a similar pattern of manually applying Redux state in other tests already.

kumar303:
@willdurand this probably isn't complete yet but it's ready for another look. Here are the main things I addressed:

- All test code resides in the helper now (no more `simulateUIStateProps`)
- All components use `componentDidMount()` to reset state which makes it behave similar to `this.setState()`

Outstanding issues:
- I'd like to not require `applyUIStateChanges()` after every state change in tests. I haven't thought of a way to solve that yet.

src/core/withUIState.js
```diff
@@ -0,0 +1,124 @@
...
+import * as React from 'react';
+import invariant from 'invariant';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+
+import { normalizeFileNameId } from 'core/utils';
+import { setUIState } from 'core/reducers/uiState';
+import type { UIStateType } from 'core/reducers/uiState';
+
+type ExtractIdType = (props: Object) => string;
```
willdurand: I think we used `Func` as suffix in the past: `ExtractIdFunc`.

src/core/withUIState.js
```diff
@@ -0,0 +1,124 @@
...
+        // props.uiState doesn't match initialState?
+        this.props.setUIState(initialState);
+      }
+
+      render() {
+        return <WrappedComponent {...this.props} />;
+      }
+    }
+
+    return compose(
```
willdurand: We might not need `compose` here.

src/core/withUIState.js
```diff
@@ -0,0 +1,124 @@
...
+        //
+        // TODO: Optimize this by only dispatching if
+        // props.uiState doesn't match initialState?
+        this.props.setUIState(initialState);
+      }
+
+      render() {
+        return <WrappedComponent {...this.props} />;
+      }
+    }
```
willdurand: It would be cool to set a display name: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

tests/unit/core/testWithUIState.js
```diff
@@ -0,0 +1,160 @@
...
+          <div className="overlay">
+            <button className="close-button" onClick={this.closeOverlay}>
+              Close overlay
+            </button>
+          </div>
+        );
+      }
+    }
+
+    const Overlay = compose(
```
willdurand: You probably don't need `compose` here.

tests/unit/core/testWithUIState.js
```diff
@@ -0,0 +1,160 @@
...
+      }),
+    )(OverlayBase);
+
+    const render = ({ id = 'some-id', ...props } = {}) => {
+      const root = shallowUntilTarget(
+        <Overlay id={id} store={store} {...props} />,
+        OverlayBase,
+      );
+      // Apply initial state change from componentWillMount.
+      applyUIStateChanges({ root, store });
```
willdurand: Could it be because of `shallowUntilTarget`? Would a plain old `shallow` + `dive` work?

tests/unit/core/testWithUIState.js
```diff
@@ -0,0 +1,160 @@
...
+        }),
+      )(ThingBase);
+
+      const root = shallowUntilTarget(<Thing store={store} />, ThingBase);
+
+      expect(root.instance().props.uiState).toEqual(initialState);
+    });
+
+    it('always resets initial state per component instance', () => {
+      class ThingBase extends React.Component {
```
willdurand: I think you could probably use a functional component to reduce the amount of code in this test:

```
const ThingBase = () => <div />;
```

tests/unit/helpers.js
```diff
@@ -611,3 +613,40 @@ export const createUserNotificationsResponse = () => {
...
+  const ownProps = root.instance().props;
+  invariant(
+    ownProps.uiStateID,
+    'uiStateID cannot be undefined; was the component wrapped in withUIState()?',
+  );
+
+  const mapStateToProps = createUIStateMapper({
+    uiStateID: ownProps.uiStateID,
+  });
+  const _stateProps = mapStateToProps(store.getState(), ownProps);
```
willdurand: Why is this variable prefixed with underscore?

src/core/withUIState.js
kumar303: Oh yeah, duh

tests/unit/core/testWithUIState.js
kumar303: You can't use `root.instance().props` on a stateless component because there is no instance. You also can't use `root.props()` because that only returns *rendered* props, not props passed to the stateless component's function. I think `instance()` is helpful here because we are testing the affect of *injecting* props into a component. It's not testing the rendering of a component, unlike our typical usage of Enzyme.

willdurand: ah, I see. But the HOC should work with stateless components though, no?

tests/unit/core/testWithUIState.js
kumar303: Nope. I got the tests passing with this:

```js
const root = shallow(<EditableCollectionAddon ...>).dive().dive().dive().dive();
```

...but `applyUIStateChanges()` was still necessary.

tests/unit/core/testWithUIState.js
kumar303: Yes, the HOC will work fine with stateless components. The `root.instance().props` call is only necessary to make an assertion about the received props while testing.

src/core/withUIState.js
kumar303: Good idea. Done.

kumar303:
I'm not totally happy with `applyUIStateChanges()` in tests but I think it's worth pushing forward as is for a couple reasons:

- I think in most cases it won't be necessary to call `applyUIStateChanges()`. I prefer tests that simulate a *single* user event and assert that a *single* Redux action was dispatched. This helps us think about the application in a pure way: *receive state* -> *render*. A lot of our tests rely on multiple state changes and multiple render calls.
- The patch as is will allow us to convert many `setState` calls to Redux. In a future patch we could improve the testing experience.

Issue #5475, opened 2018-07-02
Featured Extensions `src` not being tracked in all cases


diox:
Re-filed from https://github.com/mozilla/addons-server/issues/8263. 

**Original description by @devaneymoz:**

Go to the AMO homepage. Content programming changes frequently, but typically we have a "Featured extensions" shelf somewhere on the page. 

If you find it, click one of the featured extensions in the module display. For example, I see Enhancer for YouTube, I click it, I go to its detail page, and the URL looks like so:

https://addons.mozilla.org/firefox/addon/enhancer-for-youtube/?src=featured

Now, go back to the homepage. This time, instead of clicking on an individual featured extension in the shelf, click the "See more featured extensions" link on the far right of the module. Find Enhancer for YouTube and click thru to its detail page. The URL will look like so:

https://addons.mozilla.org/firefox/addon/enhancer-for-youtube/?src=search

It appears we're tracking the discovery flow differently (featured vs. search) when in truth they both should be the same -- src=featured

It will be difficult to measure the effectiveness of AMO's primary "Featured Extensions" collection if portions of its effect are measured as "search" results.

diox:
**Original comment by @jvillalobos:**

When you click on "See more featured extensions", you land on a search page, which is why the `src` changes to `search`.

Would it make sense to use the `featured` value whenever the search page has the "featured only" checkbox selected?

While testing this, I also noticed that if you go directly to the featured add-ons collection and click on an entry, the `src` is `collection`, so we might want to special-case the `src` from that collection as well.

diox:
Original reply by @devaneymoz:

I think it does make sense to use the "featured" value whenever the search page has the "featured only" checkbox selected, since that's effectively taking the same action as navigating to the featured collection.

ioanarusiczki:
Verified fixed on AMO dev with FF61(Win10)
 The same URL is now available in both places (when clicked the extension is clicked from AMO homepage and then from See more... link) 
https://addons-dev.allizom.org/en-US/firefox/addon/hyfy-screen-recorder/?src=featured
Same behaviour for the themes from Featured Themes shelf.

![same url](https://user-images.githubusercontent.com/33448286/42385034-596016c6-8144-11e8-951d-2a6f119ede29.gif)

PR #5489, opened 2018-07-03
Fix install source on featured search results

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5475

### DIFF
--- a/src/amo/components/SearchResults/index.js
+++ b/src/amo/components/SearchResults/index.js
@@ -4,7 +4,7 @@ import { compose } from 'redux';
 
 import AddonsCard from 'amo/components/AddonsCard';
 import Paginate from 'core/components/Paginate';
-import { INSTALL_SOURCE_SEARCH } from 'core/constants';
+import { INSTALL_SOURCE_FEATURED, INSTALL_SOURCE_SEARCH } from 'core/constants';
 import translate from 'core/i18n/translate';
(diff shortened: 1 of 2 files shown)


kumar303:
> The only question I have is about the fact that issue
#5475 also seems to suggest that if a user clicks on an add-on from the featured add-ons collection the src should also be `featured` as opposed to `collection`.

That's fixed separately in https://github.com/mozilla/addons-frontend/pull/5488

Well, if I misunderstood your comment just let me know and I'll open another issue.

bobsilverberg:
> That's fixed separately in #5488

> Well, if I misunderstood your comment just let me know and I'll open another issue.

It may be me who's misunderstanding, but I thought that what @jvillalobos was talking about at https://github.com/mozilla/addons-frontend/issues/5475#issuecomment-401806467 when he said "While testing this, I also noticed that if you go directly to the featured add-ons collection and click on an entry, the src is collection, so we might want to special-case the src from that collection as well." was that we should change the `src` on the add-on links on the View Collection page for the featured add-ons collection. That has not been addressed, but maybe we don't want to do that.

jvillalobos:
Yeah, so there are two cases:

1) A collection featured in the homepage, which is covered by #5488.

2) The Featured Add-ons collection, which is a special case. So, regardless of the way users land on the Featured Add-ons collection, the src should be the same (`featured`).

PR #5500, opened 2018-07-04
Fix display issues with long displayName on UserProfile page

### LINKED ISSUE #5494: Display name will appear cut when is longer than 25 chars on the user profile page
STR:
1. Load AMO stage and go to Edit profile page
2. Edit your display name , make sure it contains more than 25 chars
3. Update profile and observe it on the view profile page

Expected result:
Name is displayed without layout issues.

Actual result:
Name appears cut.

Notes:
This issue is reproducible when editing the username too in case the user won't edit the displayname, also when edited with a long unbroken string it will appear cut in the Extension by section.
This issue is reproducible on AMO dev and AMO stage with FF61(Win10). 

unbrokenstringdisplay

bobsilverberg:
Fixes #5494 

Before:

screenshot 2018-07-04 13 38 23

After:

screenshot 2018-07-04 13 38 55

### DIFF
--- a/src/amo/components/SearchResult/styles.scss
+++ b/src/amo/components/SearchResult/styles.scss
@@ -152,6 +152,7 @@
 
 .SearchResult-author {
   display: none;
+  overflow: auto;
   word-wrap: break-word;
 
   @include respond-to(medium) {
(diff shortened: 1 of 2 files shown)


willdurand:
> I don't have a strong opinion but another option is to just cut it off with an ellipsis.

I'd prefer to leave it as is and not use an ellipsis.

bobsilverberg:
It's also consistent with many other things on the site where we wrap and don't use an ellipsis, so I'm going to land it as is. Thanks for the review and feedback @kumar303 and @willdurand.

PR #5501, opened 2018-07-04
12 July Homepage content update

### LINKED ISSUE #5486: 12 July Homepage content update
For the **12 July** push, please update homepage content to reflect the programming in this [doc](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#heading=h.t4gj29t9yklr).

bobsilverberg:
Fixes #5486 

Note that this one is a bit different. For the past few months I've been adding and deleting code from the reducer and saga whenever the data sources for the home page changed. I was about to do that again, but then I noticed that some custom logic was added to the home page to deal with featured themes, and I didn't want to just delete all of that and the associated tests.

What I decided to do was keep pretty much all of the logic for featured themes intact, both in the `Home` component and also in the reducer and saga, and just introduce a prop into `Home` called `includeFeaturedThemes` which can be used to indicate whether the home page needs featured themes or not. I know this isn't particularly scalable, but it seems like a reasonable idea for this shelf because of the custom logic. Eventually we'll be able to do away with all of this extra work when we move to a homepage curation tool, but I think that still quite away off.


src/amo/reducers/home.js
```diff
@@ -95,10 +97,14 @@ export const loadHomeAddons = ({
 type Action = FetchHomeAddonsAction | LoadHomeAddonsAction;
 
 const createInternalAddons = (
-  response: ApiAddonsResponse,
+  response: ApiAddonsResponse | null,
 ): Array<AddonType> => {
-  return response.result.results.map((slug) =>
-    createInternalAddon(response.entities.addons[slug]),
+  if (response === null) {
```
kumar303: Even though we have pretty good Flow coverage, I think it's dangerous to allow a `null` `response`. It could lead to subtle bugs. It's also a surprising interface because why would you ask to create internal add-ons out of a null response? I would rather see the reducer fixed at the source:

```diff
diff --git a/src/amo/reducers/home.js b/src/amo/reducers/home.js
index 1f9034ffe..859ee0bd8 100644
--- a/src/amo/reducers/home.js
+++ b/src/amo/reducers/home.js
@@ -139,7 +139,9 @@ const reducer = (
           return null;
         }),
         featuredExtensions: createInternalAddons(featuredExtensions),
-        featuredThemes: createInternalAddons(featuredThemes),
+        featuredThemes: featuredThemes
+          ? createInternalAddons(featuredThemes)
+          : [],
         resultsLoaded: true,
       };
     }

```

This also would make it more clear how that parameter is optional.

tests/unit/amo/components/TestHome.js
```diff
@@ -80,8 +82,8 @@ describe(__filename, () => {
     const shelves = root.find(LandingAddonsCard);
     const shelf = shelves.find('.Home-FeaturedCollection').at(1);
 
-    expect(shelf).toHaveProp('header', 'Privacy matters');
-    expect(shelf).toHaveProp('footerText', 'See more privacy extensions');
+    expect(shelf).toHaveProp('header', 'Dynamic downloaders');
```
kumar303: Ideally, changing the order of shelves like this shouldn't require us to update any tests. Maybe in the future we can rewrite these so that they use `COLLECTIONS_TO_FETCH` (or some kind of config) to make all assertions.

tests/unit/amo/sagas/test_home.js
```diff
@@ -142,6 +144,50 @@ describe(__filename, () => {
...
+      const expectedLoadAction = loadHomeAddons({
+        collections,
+        featuredExtensions,
+        featuredThemes: null,
+      });
+
+      await sagaTester.waitFor(expectedLoadAction.type);
+      mockSearchApi.verify();
+
+      const calledActions = sagaTester.getCalledActions();
```
kumar303: This is brittle. You should use the return value of `waitFor` instead:

```js
const loadAction = await sagaTester.waitFor(expectedLoadAction.type);
mockSearchApi.verify();
```

tests/unit/amo/sagas/test_home.js
```diff
@@ -142,6 +144,50 @@ describe(__filename, () => {
...
+      const baseFilters = {
+        page_size: LANDING_PAGE_ADDON_COUNT,
+      };
+
+      const collections = [];
+
+      const featuredExtensions = createAddonsApiResult([fakeAddon]);
+      mockSearchApi
+        .expects('search')
+        .withArgs({
```
kumar303: Verifying the call parameters is not relevant to this test (unless I'm missing something?). Also, the call parameters are already verified by other tests. Repeating the validation would introduce a maintenance burden so I'd rather see it left out:

```diff
diff --git a/tests/unit/amo/sagas/test_home.js b/tests/unit/amo/sagas/test_home.js
index 1c6600441..45bd7adfd 100644
--- a/tests/unit/amo/sagas/test_home.js
+++ b/tests/unit/amo/sagas/test_home.js
@@ -145,28 +145,11 @@ describe(__filename, () => {
     });
 
     it('does not fetch featured themes if includeFeaturedThemes is false', async () => {
-      const state = sagaTester.getState();
-
-      const baseArgs = { api: state.api };
-      const baseFilters = {
-        page_size: LANDING_PAGE_ADDON_COUNT,
-      };
-
       const collections = [];
 
       const featuredExtensions = createAddonsApiResult([fakeAddon]);
       mockSearchApi
         .expects('search')
-        .withArgs({
-          ...baseArgs,
-          filters: {
-            ...baseFilters,
-            addonType: ADDON_TYPE_EXTENSION,
-            featured: true,
-            sort: SEARCH_SORT_RANDOM,
-          },
-          page: 1,
-        })
         .returns(Promise.resolve(featuredExtensions));
 
       _fetchHomeAddons({
```

kumar303:
> there is still one place left that you need to protect against undefined featured themes

If we convert this saga file to Flow, it would have shown the error ✨

tests/unit/amo/components/TestHome.js
bobsilverberg: Yes, I've been wanting to clean this up and add some more metadata that can be used by both the file and tests, so I've gone ahead and done that. It's made for some significant changes, but it's all much better, IMO.

bobsilverberg:
> If we convert this saga file to Flow, it would have shown the error ✨ 

Done in the next commit. 😄

bobsilverberg:
Thanks for the detailed review and for catching those issues @kumar303! I've addressed all of your concerns and even cleaned up the saga test a bit more.

src/amo/components/Home/index.js
```diff
@@ -43,6 +45,33 @@ export const isFeaturedCollection = (
...
 };
 
+export const getFeaturedCollectionsMetadata = (i18n: I18nType) => {
+  return [
+    {
+      footerText: i18n.gettext(
+        'See more social media customization extensions',
+      ),
+      header: i18n.gettext('Social media customization'),
+      ...FEATURED_COLLECTIONS[0],
```
kumar303: Ugh, it's awkward that the metadata will be defined in two places. I don't have a better idea though.

src/amo/components/Home/index.js
```diff
@@ -211,78 +251,74 @@ export class HomeBase extends React.Component {
...
-            addonInstallSource={INSTALL_SOURCE_FEATURED_COLLECTION}
+          <FeaturedCollectionCard
             addons={collections[0]}
             className="Home-FeaturedCollection"
-            header={i18n.gettext('Translation tools')}
-            footerText={i18n.gettext('See more translation tools')}
-            footerLink={`/collections/${FEATURED_COLLECTIONS[0].username}/${
-              FEATURED_COLLECTIONS[0].slug
-            }/`}
+            collectionMetadata={featuredCollectionsMetadata[0]}
```
kumar303: The pattern of passing in a complex object is not very React-y and I think it will make some things difficult. Flow validation of the property values is a bit cumbersome, for example.

Why not just pass them in as props? You could spread the object into props, which would be fine.

```js
<FeaturedCollectionCard
  addons={collections[0]}
  className="Home-FeaturedCollection"
  {...featuredCollectionsMetadata[0]}
/>
```

Using spreads without Flow coverage is a little risky but it's still OK.

tests/unit/amo/reducers/test_home.js
```diff
@@ -70,15 +70,23 @@ describe(__filename, () => {
...
-        createInternalAddon(fakeTheme),
-      ]);
+    });
+
+    it('loads an empty array if featured themes is null', () => {
+      const { store } = dispatchClientMetadata();
+
+      store.dispatch(
+        loadHomeAddons({
+          collections: [null],
```
kumar303: Is that realistic? Could it be `collections: []` instead?

src/amo/components/Home/index.js
```diff
@@ -43,6 +45,33 @@ export const isFeaturedCollection = (
   });
 };
 
+export const getFeaturedCollectionsMetadata = (i18n: I18nType) => {
```
kumar303: This isn't a Flow-typed file so all annotations will be ignored. Thus, it's better to remove them so they don't bitrot.

src/amo/components/FeaturedCollectionCard/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+type Props = {|
+  addons: Array<AddonType>,
+  className: string,
+  collectionMetadata: LandingCardCollectionMetadata,
+  loading: boolean,
+|};
+
+export default class FeaturedCollectionCard extends React.Component<Props> {
+  render() {
```
kumar303: Because we use shallow rendering, this function won't ever get executed and you'll see a reduced code coverage score. I'd like to see a simple test for it to make sure it executes without throwing. Since none of the props are dynamic, the test won't have to do much.

kumar303: Well, huh. The codecov report says it's covered. Let me know if I'm missing something.

src/amo/components/Home/index.js
bobsilverberg: Yeah, the need for `i18n` for the header and footer text leaves me in this position. :-(

src/amo/components/Home/index.js
bobsilverberg: Yes, this looks better. I could add Flow coverage to this file, but I guess I've already done way more than was required for issue #5501 ;-)

tests/unit/amo/reducers/test_home.js
bobsilverberg: The saga will load a `null` into the `collections` array if there is a problem fetching a collection, as is evidenced in the test above, but I guess we don't need to reuse that here. So yes, I can just make it `collections: []`.

src/amo/components/Home/index.js
bobsilverberg: Ah, yes, I didn't notice that. Although I was surprised that no other types were imported. I guess I should have put two and two together.

src/amo/components/FeaturedCollectionCard/index.js
bobsilverberg: Is this because the test uses `shallow` rather than `shallowUntilTarget`? Would it make sense to change it to the latter? If not, how exactly do I write a test targeting the `render` method?

bobsilverberg: Oh, sorry, I guess I saw your first comment and not your second. Does this mean nothing needs to be done for this?

kumar303: Sorry, ignore me. I didn't even notice `TestFeaturedCollectionCard.js` for some reason.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -72,13 +73,10 @@ export class HomeHeroBannerBase extends React.Component<Props> {
               life—work, personal, etc.—separated by color-coded tabs`)}
         </p>
       </HeroSection>,
-      <HeroSection key="hero-9" linkTo="/addon/worldwide-radio/">
-        <h3>{i18n.gettext('Worldwide Radio')}</h3>
+      <HeroSection key="hero-9" linkTo="/addon//transparent-standalone-image/">
```
EnTeQuAk: Uhh, noticed a double-slash.

PR #5507, opened 2018-07-05
Fix link to My Collections when new collections UI is disabled

### LINKED ISSUE #5506: “View My Collections” does not redirect the user correctly on AMO-prod

1.Create some collections
2.Hover over the username.
3.Click on "View My Collections".

### What happened?
The user is redirected to "Collections" page and not "Collections by [User]".
The user to be redirected to "Collections by [User]".

Reproduced on FF61.0, AMO-prod under Win 7 64-bit and Android 8.0.0
amo-prod

bobsilverberg:
Fixes #5506

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -24,6 +24,7 @@ import './styles.scss';
 
 export class HeaderBase extends React.Component {
   static propTypes = {
+    _config: PropTypes.object,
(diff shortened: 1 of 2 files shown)


bobsilverberg:
@kumar303 This fixes a P1 bug that we need to land before today's push, so if you could take a look ASAP that would be great.

Issue #5514, opened 2018-07-06
Make src consistent for feature homepage collection links


kumar303:
Follow-up to https://github.com/mozilla/addons-frontend/issues/5476

STR:
- Go to the home page
- Scroll to a featured collection, such as *Translation tools*
- Take note that links within that area include `?src=homepage-collection-featured` in the URL
- Click on *See more featured...*
- Take note of the links to each add-on

Expected result: all links after clicking *See more featured...* should contain the same identifier in the URL, `?src=homepage-collection-featured`

Actual result: the links contain `?src=collection`

The source should be the same in both cases (`src=homepage-collection-featured`), regardless of whether it's clicked directly from the homepage shelf, or from within the full collection.

ioanarusiczki:
Verified fixed on AMO dev with FF61(Win10)

![srchomepagecollectionfeatured](https://user-images.githubusercontent.com/33448286/42437831-ed2605fe-8366-11e8-8445-984840670687.gif)

PR #5515, opened 2018-07-06
Make src consistent for feature homepage collection links 

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5514


kumar303:
I think this will create conflicts with your July content update patch, sorry 😬

PR #5524, opened 2018-07-09
Reset thse byUserId data when adding a new review

### LINKED ISSUE #5470: Reviews are displayed on the user's profile page only after refreshing page
STR:
1. Load AMO dev and submit a review for an add-on
2. Go to View my profile and check if the review is displayed

Expected result:
The review is displayed after clicking View My profile.

Actual result:
No review is available.

Notes:
The Review is available only after refreshing the page or navigation to dev hub and then back to View profile is made.
This issue is reproducible on AMO dev with FF61(Win10). 

review showp up

willdurand:
Fix #5470

### DIFF
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -264,6 +264,11 @@ export default function reviewsReducer(
       return {
         ...state,
         byId: storeReviewObjects({ state, reviews: [payload] }),
+        byUserId: {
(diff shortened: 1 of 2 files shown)


willdurand:
> I think we should strive for this pattern whenever caching does not pose a problem: the user updates some data -> invalidate local state to force a re-fetch.

amen to that.

PR #5526, opened 2018-07-09
Do not add br between html tags

### LINKED ISSUE #5473: Profile page biography field creates invalid markup
STR:

Edit user profile and add the following:

```html
<ul>
<li>ul li</li>
</ul>
```

The output on the page after saving is:

```html
<ul>
<br>
<li>ul li</li>
<br>
</ul>
```

There shouldn't be brs created inside the `ul`.

willdurand:
Fix #5473

---

This PR removes `<br />` occurences between HTML tags.

### DIFF
--- a/tests/unit/core/utils/test_index.js
+++ b/tests/unit/core/utils/test_index.js
@@ -406,12 +406,18 @@ describe(__filename, () => {
       expect(nl2br(null)).toEqual('');
     });
 
+    it('returns mixed content with <br/>', () => {
+      expect(nl2br('foo\nbar\n\n<b>bold</b>')).toEqual(
+        'foo<br />bar<br />\n<b>bold</b>',
+      );
+    });
+
     it('returns valid HTML', () => {
       const htmlValue =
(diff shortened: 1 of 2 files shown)


tests/unit/core/utils/test_index.js
```diff
@@ -404,6 +404,21 @@ describe(__filename, () => {
...
+
+    it('returns mixed content with <br/>', () => {
+      expect(nl2br('foo\nbar\n\n<b>bold</b>')).toEqual(
+        'foo<br />bar<br />\n<b>bold</b>',
+      );
+    });
+
+    it('returns valid HTML', () => {
+      const htmlValue =
+        '<ul>\n<li>ul\nli<strong>foo\nbar</strong>\n</li>\n</ul>';
```
kumar303: There is a lot going on in this test:
- line breaks between tags are preserved (multiple instances)
- line breaks in tag content are converted (multiple instances)
- tag content that looks like tags do not trick the converter

It would be easier to read the failures if you broke it down into a few tests, each focusing on a single case.

Issue #5532, opened 2018-07-09
Support Notice button links


kumar303:
The existing `<Notice>` component lets you create an action button with a callback but not one which is an anchor tag link. In https://github.com/mozilla/addons-frontend/issues/5438 we need it to be a link.

PR #5533, opened 2018-07-09
Support Notice button links

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5532


willdurand:
I'm going to merge this PR because the `master` branch is broken and this PR contains a fix for it (which is not isolated in a commit I can cherry-pick unfortunately).

kumar303:
> ...which is not isolated in a commit...

ah, whoops. That did cross my mind at some point. Thanks.

PR #5539, opened 2018-07-10
Enable user profile pages on all envs

willdurand:
Refs #5541

---

I decided to keep the feature flag in case something goes really wrong. If everything works as intended, I'll remove the feature flag code in the next tag.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -288,10 +288,8 @@ module.exports = {
   // Expose the new screen for editing a collection.
   enableNewCollectionsUI: false,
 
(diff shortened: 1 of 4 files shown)


bobsilverberg:
Oh, I guess we should have an issue for removing the special logic that uses that flag once everything is considered stable.

willdurand:
> Oh, I guess we should have an issue for removing the special logic that uses that flag once everything is considered stable.

Ah good call, filed https://github.com/mozilla/addons-frontend/issues/5544

PR #5542, opened 2018-07-10
Apply Prettier to SCSS files

### LINKED ISSUE #5431: Apply Prettier to SCSS files

willdurand:
Fix #5431

### DIFF
--- a/src/amo/css/inc/vars.scss
+++ b/src/amo/css/inc/vars.scss
@@ -1,4 +1,4 @@
-@import "~core/css/inc/vars";
+@import '~core/css/inc/vars';
 
 $body-font-color: #000;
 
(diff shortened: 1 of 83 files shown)


src/amo/css/inc/vars.scss
```diff
@@ -14,12 +14,34 @@ $accent-color: #ebebeb;
...
+      min-width: 720px,
+    ),
+  extraLarge: only screen and
+    (
+      min-width: 900px,
+    ),
+  extraExtraLarge: only screen and
+    (
+      min-width: 1150px,
+    )
```
willdurand: This is weird.. 🤷‍♂️

src/ui/components/LoadingText/styles.scss
```diff
@@ -35,6 +35,10 @@
 
 @for $delay from 1 through 3 {
   .LoadingText--delay-#{$delay} {
-    animation: placeHolderShimmer #{$delay * 1.5}s infinite cubic-bezier(0.65, 0.05, 0.36, 1);
+    animation: placeHolderShimmer
+      #{$delay *
+      1.5}s
```
willdurand: This is weird too.

willdurand:
I don't know how you feel about that, there are annoying things but it does a decent job otherwise..

src/amo/css/inc/vars.scss
```diff
@@ -14,12 +14,34 @@ $accent-color: #ebebeb;
 // it with `@include respond-to(medium) { [...] }`.
 $breakpoints: (
   // This breakpoint exists for tweaking of very small screens.
-  extraSmall: only screen and (max-width: 349px),
-  mediumOnly: only screen and (min-width: 500px) and (max-width: 719px),
-  medium: only screen and (min-width: 500px),
-  large: only screen and (min-width: 720px),
-  extraLarge: only screen and (min-width: 900px),
-  extraExtraLarge: only screen and (min-width: 1150px),
+    extraSmall: only screen and
```
kumar303: Huh. This seems like it should not be indented.

PR #5559, opened 2018-07-11
adding manage my submissions under tool

### LINKED ISSUE #5487: Navigate to Manage My Submissions from AMO header
@MeridelW and I did a walkthrough of how developers manage their extensions, and it looks like developers currently can't navigate to 'manage my add-ons' from the AMO header: 

screen shot 2018-07-03 at 1 49 16 pm

Can we add it to the drop down under 'Tools'? Otherwise, developers have to scroll down to the footer to access the Developer Hub, or they have to start a "submit new extension/theme" and then navigate to "manage my add-ons".

championshuttler:
Fixes #5487

### DIFF
--- a/tests/unit/amo/components/TestHeader.js
+++ b/tests/unit/amo/components/TestHeader.js
@@ -185,7 +185,9 @@ describe(__filename, () => {
       userProps: { username },
     });
     const wrapper = renderHeader({ store });
-    const link = wrapper.find('[className = "Header-user-menu-developers-submissions-on-link"]');
+    const link = wrapper.find(
+      '.Header-user-menu-developers-submissions-on-link',
+    );
 
     expect(link).toHaveLength(1);
     expect(link.prop('children')).toEqual('Manage My Submissions');
--- a/tests/unit/amo/components/TestHeader.js
+++ b/tests/unit/amo/components/TestHeader.js
@@ -178,4 +178,17 @@ describe(__filename, () => {
 
     expect(link).toHaveLength(0);
   });
+
+  it('displays manage my submissions when user is logged in', () => {
+    const username = 'championshuttler';
+    const { store } = dispatchSignInActions({
+      userProps: { username },
+    });
+    const wrapper = renderHeader({ store });
+    const link = wrapper.find('.Header-user-menu-developers-submissions-on-link');
+
+    expect(link).toHaveLength(1);
+    expect(link.prop('children')).toEqual('Manage My Submissions');
+    expect(link).toHaveProp('href', '/developers/addons/');
+  });
 });
--- a/tests/unit/amo/components/TestHeader.js
+++ b/tests/unit/amo/components/TestHeader.js
@@ -179,18 +179,16 @@ describe(__filename, () => {
     expect(link).toHaveLength(0);
   });
 
-  it('displays manage my submissions when user is logged in', () => {
+  it('displays a "manage my submissions" link when user is logged in', () => {
     const username = 'championshuttler';
     const { store } = dispatchSignInActions({
       userProps: { username },
     });
     const wrapper = renderHeader({ store });
-    const link = wrapper.find(
-      '.Header-user-menu-developers-submissions-on-link',
-    );
+    const link = wrapper.find('.Header-user-menu-developers-submissions-link');
 
     expect(link).toHaveLength(1);
-    expect(link.prop('children')).toEqual('Manage My Submissions');
+    expect(link.children()).toHaveText('Manage My Submissions');
     expect(link).toHaveProp('href', '/developers/addons/');
   });
 });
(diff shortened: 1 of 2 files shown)


championshuttler:
@willdurand I tried to write test case , I know it has bugs :cry: . Test cases become nightmare for me :/ . I can understand the test cases but when it comes to writing test cases I got stuck

tests/unit/amo/components/TestHeader.js
```diff
@@ -178,4 +178,17 @@ describe(__filename, () => {
...
     expect(link).toHaveLength(0);
   });
+
+  it('displays manage my submissions when user is logged in', () => {
+    const username = 'championshuttler';
+    const { store } = dispatchSignInActions({
+      userProps: { username },
+    });
+    const wrapper = renderHeader({ store });
+    const link = wrapper.find('.Header-user-menu-developers-submissions-on-link');
```
willdurand: You need to find "something" that exists. Here you use a CSS class name, which is nice but it does not exist in the `Header` component. Add a `className` prop to the `Link` and it should work.

willdurand: @championshuttler do you need more information? I think you almost got it right.

championshuttler: @willdurand I was searching for it that how to add it and also tried in test cases writtern already but did not get any lead . Can you guide me how to do that?

willdurand: What if you add a `className` prop to the link you've added? Its value will be `Header-user-menu-developers-submissions-on-link`

tests/unit/amo/components/TestHeader.js
```diff
@@ -185,7 +185,7 @@ describe(__filename, () => {
       userProps: { username },
     });
     const wrapper = renderHeader({ store });
-    const link = wrapper.find('.Header-user-menu-developers-submissions-on-link');
+    const link = wrapper.find('[className = "Header-user-menu-developers-submissions-on-link"]');
```
willdurand: The test case was correct. Look at your source file and see if you can find `.Header-user-menu-developers-submissions-on-link` 😉

tests/unit/amo/components/TestHeader.js
championshuttler: @willdurand I updated it according to this airbnb.io/enzyme/docs/api/ReactWrapper/find.html# . But its failing again with error `Failed to parse selector: [className = "Header-user-menu-developers-submissions-on-link"]`

tests/unit/amo/components/TestHeader.js
championshuttler: Oops, i think no :/

championshuttler: ```.Header-user-menu-developers-submissions-on-link``` it does not exist :stuck_out_tongue:

willdurand: You should add this class name to the `Link` you have added, but in the source file.

`wrapper.find()` tries to find components into the source component (the component/file under test). Your test case is correct, because it tries to find the link you have added. The problem is that you look for a component with a given class name (`.Header-user-menu-developers-submissions-on-link`), which is good. But, because you don't specify it in the source file, the test case fails because it does not find anything, like you :)

So, add this class name to the link you have added in the source file.

tests/unit/amo/components/TestHeader.js
willdurand: Yes, the previous version of your code was ok: `wrapper.find('.Header-user-menu-developers-submissions-on-link')`

championshuttler:
Done :smile: 

screenshot from 2018-07-17 00-23-35

tests/unit/amo/components/TestHeader.js
championshuttler: @willdurand got it , thanks for the help :heart:  .

tests/unit/amo/components/TestHeader.js
```diff
@@ -178,4 +178,19 @@ describe(__filename, () => {
 
     expect(link).toHaveLength(0);
   });
+
+  it('displays manage my submissions when user is logged in', () => {
```
willdurand: can you write: "displays a "manage my submissions" link when user is logged in"

tests/unit/amo/components/TestHeader.js
```diff
@@ -178,4 +178,19 @@ describe(__filename, () => {
...
+    const { store } = dispatchSignInActions({
+      userProps: { username },
+    });
+    const wrapper = renderHeader({ store });
+    const link = wrapper.find(
+      '.Header-user-menu-developers-submissions-on-link',
+    );
+
+    expect(link).toHaveLength(1);
+    expect(link.prop('children')).toEqual('Manage My Submissions');
```
willdurand: Could you use `.children()` instead of `.prop('children')`?

tests/unit/amo/components/TestHeader.js
championshuttler: Done :+1:

tests/unit/amo/components/TestHeader.js
championshuttler: :100:

willdurand:
thanks @championshuttler!

PR #5561, opened 2018-07-11
Introduce a `AppState` type for redux

### LINKED ISSUE #4058: Define and use a single Redux application state type
We currently define Flow types for Redux application state in an ad-hoc manner. Example:

```js
import type { CollectionsState } from 'amo/reducers/collections';

const mapStateToProps = (state: {| collections: CollectionsState |}) => {
  // Do something with state.collections, but nothing else.
};
```

However, Redux always provides the exact same `state` object to every `mapStateToProps` function. 

We should define a shared type and use it everywhere, like this:

```js
import type { AppState } from 'amo/types/appState';

const mapStateToProps = (state: AppState) => {
  // Do something with state.collections or whatever...
};
```

willdurand:
Fix #4058

---

Other changes:

- All sub-state (or "small" state) have a type that is the name of the reducer + the `State` suffix:

```
reducers/abuse  => AbuseState
reducers/fooBar => FooBarState
```

### DIFF
--- a/src/amo/store.js
+++ b/src/amo/store.js
@@ -76,6 +76,18 @@ export type AppState = {|
   viewContext: ViewContextState,
 |};
 
+// This is a type function that takes a state type and returns a reducer
+// type, i.e. a function that accepts and returns the same state type.
+/* eslint-disable no-undef */
+type CreateReducerType = <AnyState>(
+  AnyState,
+) => (AnyState, action: Object) => AnyState;
+/* eslint-enable no-undef */
+
+// Given AppState, create a type for all possible application reducers.
+// See https://flow.org/en/docs/types/utilities/#toc-objmap
+type AppReducersType = $ObjMap<AppState, CreateReducerType>;
+
 type CreateStoreParams = {|
   history: Object,
   initialState: Object,
@@ -86,34 +98,35 @@ export default function createStore({
   initialState = {},
 }: CreateStoreParams = {}) {
   const sagaMiddleware = createSagaMiddleware();
+  const reducers: AppReducersType = {
+    abuse,
+    addons,
+    addonsByAuthors,
+    api,
+    autocomplete,
+    categories,
+    collections,
+    errors,
+    errorPage,
+    formOverlay,
+    heroBanners,
+    home,
+    infoDialog,
+    installations,
+    landing,
+    languageTools,
+    recommendations,
+    redirectTo,
+    reviews,
+    routing,
+    search,
+    uiState,
+    userAbuseReports,
+    users,
+    viewContext,
+  };
   const store = _createStore(
-    combineReducers({
-      abuse,
-      addons,
-      addonsByAuthors,
-      api,
-      autocomplete,
-      categories,
-      collections,
-      errors,
-      errorPage,
-      formOverlay,
-      heroBanners,
-      home,
-      infoDialog,
-      installations,
-      landing,
-      languageTools,
-      recommendations,
-      redirectTo,
-      reviews,
-      routing,
-      search,
-      uiState,
-      userAbuseReports,
-      users,
-      viewContext,
-    }),
+    combineReducers(reducers),
     initialState,
     middleware({
       routerMiddleware: routerMiddleware(history),
--- a/src/amo/store.js
+++ b/src/amo/store.js
@@ -1,3 +1,4 @@
+/* @flow */
 import { createStore as _createStore, combineReducers } from 'redux';
 import createSagaMiddleware from 'redux-saga';
 import { browserHistory } from 'react-router';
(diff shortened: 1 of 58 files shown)


src/amo/store.js
```diff
@@ -28,11 +29,62 @@ import redirectTo from 'core/reducers/redirectTo';
...
+  routing: Object,
+  search: SearchState,
+  uiState: UIStateState,
+  userAbuseReports: UserAbuseReportsState,
+  users: UsersState,
+  viewContext: ViewContextState,
+|};
+
+type CreateStoreParams = {|
+  history: Object,
```
bobsilverberg: Can we do any better than `Object` for this?

src/core/components/AuthenticateButton/index.js
```diff
@@ -98,15 +99,12 @@ export class AuthenticateButtonBase extends React.Component<InternalProps> {
...
+  api: ApiState,
   handleLogIn: HandleLogInFunc,
   siteUser: UserType | null,
 |};
 
-export const mapStateToProps = (state: {|
-  api: ApiStateType,
-  users: UsersStateType,
-|}): StateMappedProps => ({
+export const mapStateToProps = (state: AppState): StateMappedProps => ({
```
bobsilverberg: We don't seem to specify a return type for `mapStateToProps` anywhere else, so why are we doing it here? Perhaps this should be removed, or perhaps an issue should be filed to do this for all instances of `mapStateToProps`?

src/core/reducers/uiState.js
```diff
@@ -3,7 +3,7 @@ import invariant from 'invariant';
 
 export const SET_UI_STATE = 'SET_UI_STATE';
 
-export type UIStateType = {
+export type UIStateState = {
```
bobsilverberg: This is a bit of a mouthful, but I guess it follows the convention.

src/ui/components/FormOverlay/index.js
```diff
@@ -152,10 +152,7 @@ export class FormOverlayBase extends React.Component<Props> {
   }
 }
 
-const mapStateToProps = (
-  state: {| formOverlay: FormOverlayState |},
-  ownProps: Props,
-): $Shape<Props> => {
+const mapStateToProps = (state: AppState, ownProps: Props): $Shape<Props> => {
```
bobsilverberg: Ah, I see here's another example of specifying a return type, but using `$Shape<Props>` rather than an explicit type.

src/core/components/AuthenticateButton/index.js
willdurand: That was there before... I am not sure we need to type hint the return value since it is never used directly.

src/amo/store.js
willdurand: mmh this comes from a third-party lib, and we don't use flow-typed.

src/ui/components/FormOverlay/index.js
willdurand: That was there before too 🤷‍♂️

bobsilverberg: Right, well if we don't need these, and this is a cleanup patch, we might as well clean them up.

willdurand: I do agree.

src/core/components/AuthenticateButton/index.js
bobsilverberg: It seems odd that you'd get an error though. 🤷‍♂️

willdurand:
I wrote a new eslint rule to help me with this refactoring, and the rule proved to be useful:

- I found code that could use the new `AppState` type
- selectors are not consistent (we should probably pass `state: AppState` in our selectors, instead of the specialized states)

src/amo/store.js
```diff
@@ -28,11 +29,62 @@ import redirectTo from 'core/reducers/redirectTo';
...
+import type { ApiState } from 'core/reducers/api';
+import type { ErrorPageState } from 'core/reducers/errorPage';
+import type { FormOverlayState } from 'core/reducers/formOverlay';
+import type { LanguageToolsState } from 'core/reducers/languageTools';
+import type { InstallationsState } from 'core/reducers/installations';
+import type { RedirectToState } from 'core/reducers/redirectTo';
+import type { SearchState } from 'core/reducers/search';
+import type { UIStateState } from 'core/reducers/uiState';
+
+export type AppState = {|
```
kumar303: This is super great but if we add a new reducer and forget to update `AppState`, Flow will not show an error.

After some quick digging, I found a way to solve that. It's a little cryptic (!) but I added comments to explain what's happening. I tested removing / adding a reducer and Flow does show an error. Flow also wanted me to make a small change to the `reviews` reducer. 

Can you add this patch?

```diff
diff --git a/.eslintrc b/.eslintrc
index 2687f1a3b..4d09ac1f1 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,8 @@
     "document": true,
     "webpackIsomorphicTools": true,
     "ga": true,
+    // https://flow.org/en/docs/types/utilities/#toc-objmap
+    "$ObjMap": true,
     // https://flow.org/en/docs/types/utilities/#toc-propertytype
     "$PropertyType": true,
     // These are undocumented utilities that ship with Flow
diff --git a/src/amo/reducers/reviews.js b/src/amo/reducers/reviews.js
index eaca60906..f5161eba7 100644
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -167,7 +167,7 @@ export const changeViewState = ({
   state,
   reviewId,
   stateChange,
-}: ChangeViewStateParams = {}): $Shape<ReviewsState> => {
+}: ChangeViewStateParams = {}): ReviewsState => {
   const change = { ...stateChange };
 
   const existingFlag = state.view[reviewId] ? state.view[reviewId].flag : {};
diff --git a/src/amo/store.js b/src/amo/store.js
index 411554cb0..f7fd2106e 100644
--- a/src/amo/store.js
+++ b/src/amo/store.js
@@ -76,6 +76,18 @@ export type AppState = {|
   viewContext: ViewContextState,
 |};
 
+// This is a type function that takes a state type and returns a reducer
+// type, i.e. a function that accepts and returns the same state type.
+/* eslint-disable no-undef */
+type CreateReducerType = <AnyState>(
+  AnyState,
+) => (AnyState, action: Object) => AnyState;
+/* eslint-enable no-undef */
+
+// Given AppState, create a type for all possible application reducers.
+// See https://flow.org/en/docs/types/utilities/#toc-objmap
+type AppReducersType = $ObjMap<AppState, CreateReducerType>;
+
 type CreateStoreParams = {|
   history: Object,
   initialState: Object,
@@ -86,34 +98,35 @@ export default function createStore({
   initialState = {},
 }: CreateStoreParams = {}) {
   const sagaMiddleware = createSagaMiddleware();
+  const reducers: AppReducersType = {
+    abuse,
+    addons,
+    addonsByAuthors,
+    api,
+    autocomplete,
+    categories,
+    collections,
+    errors,
+    errorPage,
+    formOverlay,
+    heroBanners,
+    home,
+    infoDialog,
+    installations,
+    landing,
+    languageTools,
+    recommendations,
+    redirectTo,
+    reviews,
+    routing,
+    search,
+    uiState,
+    userAbuseReports,
+    users,
+    viewContext,
+  };
   const store = _createStore(
-    combineReducers({
-      abuse,
-      addons,
-      addonsByAuthors,
-      api,
-      autocomplete,
-      categories,
-      collections,
-      errors,
-      errorPage,
-      formOverlay,
-      heroBanners,
-      home,
-      infoDialog,
-      installations,
-      landing,
-      languageTools,
-      recommendations,
-      redirectTo,
-      reviews,
-      routing,
-      search,
-      uiState,
-      userAbuseReports,
-      users,
-      viewContext,
-    }),
+    combineReducers(reducers),
     initialState,
     middleware({
       routerMiddleware: routerMiddleware(history),

```

willdurand: ah good call, I used the new type on this object too but then flow complained too much so I removed it...

PR #5564, opened 2018-07-11
Better I18nType

### LINKED ISSUE #5563: Improve the `I18nType` Flow type
The Flow type `I18nType` is an alias for `Object`. It would make sense to create a better type for it, i.e. declare the methods/objects that are exposed.

willdurand:
Fix #5563

### DIFF
--- a/src/amo/components/SearchResult/index.js
+++ b/src/amo/components/SearchResult/index.js
@@ -48,7 +48,7 @@ export class SearchResultBase extends React.Component<InternalProps> {
     const { addon, i18n, showMetadata, showSummary } = this.props;
 
     const isTheme = this.addonIsTheme();
-    const averageDailyUsers = addon && addon.average_daily_users;
+    const averageDailyUsers = addon ? addon.average_daily_users : null;
 
     // Fall-back to default icon if invalid icon url.
     const iconURL = getAddonIconUrl(addon);
@@ -157,7 +157,7 @@ export class SearchResultBase extends React.Component<InternalProps> {
         <h3 className="SearchResult-users SearchResult--meta-section">
           <Icon className="SearchResult-users-icon" name="user-fill" />
           <span className="SearchResult-users-text">
-            {addon ? (
+            {averageDailyUsers !== null && averageDailyUsers !== undefined ? (
               i18n.sprintf(
                 i18n.ngettext(
                   '%(total)s user',
--- a/src/ui/components/ConfirmButton/index.js
+++ b/src/ui/components/ConfirmButton/index.js
@@ -13,7 +13,7 @@ type Props = {|
   buttonType?: string,
   cancelButtonText?: string | null,
   cancelButtonType?: string,
-  children: React.Element<any>,
+  children: React.Element<any> | string,
   className?: string,
   confirmButtonText?: string | null,
   confirmButtonType?: string,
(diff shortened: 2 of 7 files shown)


src/amo/components/SearchResult/index.js
```diff
@@ -157,7 +157,7 @@ export class SearchResultBase extends React.Component<InternalProps> {
         <h3 className="SearchResult-users SearchResult--meta-section">
           <Icon className="SearchResult-users-icon" name="user-fill" />
           <span className="SearchResult-users-text">
-            {addon ? (
+            {averageDailyUsers ? (
```
bobsilverberg: What if `averageDailyUsers` is `0`?

src/ui/components/ConfirmButton/index.js
```diff
@@ -13,7 +13,7 @@ type Props = {|
   buttonType?: string,
   cancelButtonText?: string | null,
   cancelButtonType?: string,
-  children: React.Element<any>,
+  children: React.Element<any> | string,
```
bobsilverberg: I don't understand why this was needed. Did Flow have a problem with this, or is this just something you noticed? `children` doesn't seem to be passed into any `i18n` functions.

src/ui/components/UserCollection/index.js
```diff
@@ -43,7 +43,7 @@ export const UserCollectionBase = (props: InternalProps) => {
     invariant(name, 'name is required');
     invariant(slug, 'slug is required');
     invariant(
-      Number.isInteger(numberOfAddons),
+      numberOfAddons && Number.isInteger(numberOfAddons),
```
bobsilverberg: What if `numberOfAddons` is `0`?

src/amo/components/SearchResult/index.js
willdurand: ah good catch

src/ui/components/ConfirmButton/index.js
willdurand: Because we use this component that way:

```
<ConfirmButton>{i18n.gettext('submit')}</ConfirmButton>
```

which is perfectly valid, but Flow started to complain that `React.Element<any>` was not compatible with `string`.

src/amo/components/SearchResult/index.js
```diff
@@ -157,7 +157,7 @@ export class SearchResultBase extends React.Component<InternalProps> {
         <h3 className="SearchResult-users SearchResult--meta-section">
           <Icon className="SearchResult-users-icon" name="user-fill" />
           <span className="SearchResult-users-text">
-            {addon ? (
+            {averageDailyUsers !== null && averageDailyUsers !== undefined ? (
```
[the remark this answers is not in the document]
willdurand: yes, but Flow is complaining :(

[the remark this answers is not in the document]
bobsilverberg: I guess if Flow is complaining then we've defined `addon` so that `average_daily_users` can be `undefined`. Is that accurate?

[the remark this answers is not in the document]
willdurand: indeed.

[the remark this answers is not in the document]
willdurand: that might be a mistake though, because we often mix `foo?: type` and `foo: type | null`.

PR #5565, opened 2018-07-11
Rename state param to apiState in callApi()

### LINKED ISSUE #3346: callApi `state` param should be called `apiState`
The `state` param in `callApi` actually just accepts `state.api`. We should rename it accordingly because that's confusing.

willdurand:
Fix #3346


willdurand:
rebased

PR #5572, opened 2018-07-11
Add focus style on upload button and notification checkboxes

### LINKED ISSUE #5567: Choose photo button and checkbox labels have no focus styles
If navigating with a keyboard in the edit user profile page, there's no focus styles for the user "Choose photo" upload button:

user_profile_for_muffinresearch__script_alert_0___script__ _add-ons_for_firefox

The checkboxes similarly have no focus styles. If using a keyboard to navigate there's no way of knowing that you're on a checkbox to be able to interact with it.

user_profile_for_muffinresearch__script_alert_0___script__ _add-ons_for_firefox

### LINKED ISSUE #5573: User profile inputs and textarea have no focus styles
Using a keyboard to focus the inputs and textarea there are no focus styles.

user_profile_for_muffinresearch__script_alert_0___script__ _add-ons_for_firefox

user_profile_for_muffinresearch__script_alert_0___script__ _add-ons_for_firefox

willdurand:
Fix #5567 
Fix #5573

---

I have slightly updated the focus style to match the Photon spec: https://design.firefox.com/photon/components/buttons.html#behaviours.

Before:

After:

### DIFF
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
+import withUIState from 'core/withUIState';
 import ConfirmButton from 'ui/components/ConfirmButton';
 import UserAvatar from 'ui/components/UserAvatar';
 import type { UserType } from 'amo/reducers/users';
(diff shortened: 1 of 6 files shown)


src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -19,71 +20,105 @@ type Props = {|
...
+      user,
+    } = this.props;
+
+    const altText = user
+      ? i18n.sprintf(i18n.gettext('Profile picture for %(name)s'), {
+          name: user.name,
+        })
+      : null;
+
+    const buttonClass = makeClassName(
```
kumar303: I guess this was in the old patch too but this should be wrapped in a `Button` component. Otherwise, it makes `Button` very fragile to change since its style logic is scattered about the application. I assume there was a reason why you couldn't use the `Button` component so could you file a bug to solve that?

willdurand: yes, the reason is that it is not really a button, it is a span that looks like a button. It is a trick to style a `input[type="file"]`... I'll open an issue.

willdurand: Issue: https://github.com/mozilla/addons-frontend/issues/5623https://github.com/mozilla/addons-frontend/issues/5623

Issue #1260, opened 2016-10-19
Standardize on component layout


kumar303:
Our components typically have associated Sass and image files and we have a couple different styles of organizational layout. We should pick one and convert everything to that style.

Here are the current approaches:

Component is self contained:

```
src/amo/components/SystematicDysfunctioner/index.js
src/amo/components/SystematicDysfunctioner/SystematicDysfunctioner.scss
src/amo/components/SystematicDysfunctioner/img/thing.svg
```

Component is mingled with everything else:

```
src/amo/components/SystematicDysfunctioner.js
src/amo/components/SystematicDysfunctioner.scss
src/amo/components/SomeOtherThing.js
```

Component parts are scattered:

```
src/amo/components/SystematicDysfunctioner.js
src/amo/css/SystematicDysfunctioner.scss
src/amo/img/thing.svg
```

I like the self-contained approach. However, I think it would be fair to allow `src/amo/components/SystematicDysfunctioner.js` _only_ if there are no other files associated with the component. Thoughts?

tofumatt:
I really dislike the contained style SOLELY because it means I have a bunch
of index.js files open in my editor and I find that really hard to scan. I
also find it much easier to find files in the flat folder structure.

I think putting scss and js together with the same name and different
extensions is nice, and then perhaps a folder of the same name for assets.

kumar303:
I guess another self-contained approach could be this:
````
src/amo/components/SystematicDysfunctioner.js
src/amo/components/SystematicDysfunctioner/SystematicDysfunctioner.scss
src/amo/components/SystematicDysfunctioner/img/thing.svg
````
although I dislike the clutter of more than one component item in the directory of components.

The thing I like about the self-contained approach (`SystematicDysfunctioner/index.js`) is that it's easy to move stuff around, such as when you extract something into core.

In any case, it doesn't matter too much to me what we pick. We just need to pick something otherwise it will be super annoying to try and find files any time you're looking for one.

mstriemer:
I prefer the self-contained approach, even if there's only a .js file.

> src/amo/components/SystematicDysfunctioner/index.js
> src/amo/components/SystematicDysfunctioner/SystematicDysfunctioner.scss
> src/amo/components/SystematicDysfunctioner/img/thing.svg

or

> src/amo/components/SystematicDysfunctioner/index.js

That being said `src/amo/components/SystematicDysfunctioner.js` without styles/images isn't the worst thing but I'd rather it be consistent.

[...]

Issue #5323, opened 2018-06-19
Remove video from top of about:addons (disco pane)


jenniferdavidson:
As a first step in making the content at the top of about:addons better, remove the video from the top of about:addons. 

The screenshot below shows content and a link to the video. Remove the video. 
![disco pane screenshot](https://user-images.githubusercontent.com/911251/41627175-04a396bc-73d4-11e8-9051-a5fbd2e43351.png)

vcarciu:
Verified as fixed : 
![removevideo](https://user-images.githubusercontent.com/1583842/42769749-cdeb4f1c-892b-11e8-95ab-996422f70e55.png)

brassy-:
Hey, vcarciu!

Here how I'm expecting to see the content below.
<img width="704" alt="screen shot 2018-07-17 at 10 40 20" src="https://user-images.githubusercontent.com/6309827/42806073-de85656a-89ad-11e8-8301-e6cbcbee3dc7.png">

To achive this I simply made some change to `.disco-content` 

```
.disco-content 
{
    font-size: 16px;
    margin-right: 50px;
    max-width: 620px;
    line-height: 1.5;
}
```
and set `.video-wrapper` on `display: none`. 
Feel free to implement it in the way you prefer — I assume have the video on display none can be bad for performance.

vcarciu:
@willdurand : can we made the changes requested above? I will remove the verified fixed label .

willdurand:
Do we really need this margin-right of `50px`? Here is the content with a larger font size + line-height, but no margin:

![screen shot 2018-07-17 at 13 43 49](https://user-images.githubusercontent.com/217628/42815384-84618798-89c7-11e8-89d2-61a0fa2ce114.png)

brassy-:
We don’t IMHO, but we need the max-width!

Il giorno mar 17 lug 2018 alle 13:44 William Durand <
notifications@github.com> ha scritto:

willdurand:
> We don’t IMHO, but we need the max-width!

roger that, thanks!

vcarciu:
@brassy- : Please have a look and let us know if everything is ok now : 
![layout](https://user-images.githubusercontent.com/1583842/42817704-91bf64b8-89d7-11e8-9e16-c875644a51b6.png)

brassy-:
Thank you! It looks good :)

PR #5576, opened 2018-07-12
Cleanup disco

willdurand:
Refs #1260

---

- remove the `containers/` directory
- move scss files next to their components


willdurand:
> Also just checking..was it officially decided to remove the container dirs and just move everything to components?

Yes, we don't use the `containers/` dir anymore.

> I still see an image in the disco/img dir. can we move that its respective dir? It looks like the DiscoPane component uses this, right?

let me check

willdurand:
> I still see an image in the disco/img dir. can we move that its respective dir? It looks like the DiscoPane component uses this, right?

good idea!

rebmullin:
@willdurand, I see some videos here too. Should we do the same as we do for images (put in the components dir) or what do you think?

willdurand:
> @willdurand, I see some videos here too. Should we do the same as we do for images (put in the components dir) or what do you think?

Good catch, although I'll have to remove the video in #5323

PR #5579, opened 2018-07-12
Cleanup ui

willdurand:
Refs #1260

---

- move images to `img/` directory
- rename scss files to `styles.scss`


willdurand:
> Just checking...I still see some vars.scss files in individual dirs (e.g: Icons and Ratings). is that okay? or should we move these to the css/vars dir?

I think that's fine

kumar303:
jason-i-love-it

PR #5582, opened 2018-07-12
use sinon.assert everywhere

### LINKED ISSUE #2437: Replace sinon calls wrapped with expect to use sinon.assert directly

willdurand:
Fix #2437 

---

- use `sinon.assert` everywhere
- remove more unused code

### DIFF
--- a/tests/unit/amo/components/TestRatingManager.js
+++ b/tests/unit/amo/components/TestRatingManager.js
@@ -101,7 +101,6 @@ describe(__filename, () => {
       loadSavedReview,
     });
 
-    sinon.assert.called(loadSavedReview);
     sinon.assert.calledWith(loadSavedReview, {
       apiState: signedInApiState,
       userId,
@@ -391,7 +390,6 @@ describe(__filename, () => {
           .returns(Promise.resolve({ ...fakeReview, ...params }));
 
         return actions.submitReview(params).then(() => {
-          sinon.assert.called(dispatch);
           sinon.assert.calledWith(dispatch, setReview(fakeReview));
           mockApi.verify();
         });
@@ -422,7 +420,6 @@ describe(__filename, () => {
           })
           .then(() => {
             mockApi.verify();
-            sinon.assert.called(dispatch);
             sinon.assert.calledWith(dispatch, setReview(fakeReview));
           });
       });
--- a/tests/unit/amo/components/TestRatingManager.js
+++ b/tests/unit/amo/components/TestRatingManager.js
@@ -101,9 +101,8 @@ describe(__filename, () => {
       loadSavedReview,
     });
 
-    expect(loadSavedReview.called).toEqual(true);
-    const args = loadSavedReview.firstCall.args[0];
-    expect(args).toEqual({
+    sinon.assert.called(loadSavedReview);
+    sinon.assert.calledWith(loadSavedReview, {
       apiState: signedInApiState,
       userId,
       addonId: addon.id,
@@ -126,7 +125,7 @@ describe(__filename, () => {
       }),
     });
     return root.onSelectRating(5).then(() => {
-      expect(submitReview.called).toEqual(true);
+      sinon.assert.called(submitReview);
 
       const call = submitReview.firstCall.args[0];
       expect(call.versionId).toEqual(321);
@@ -147,7 +146,7 @@ describe(__filename, () => {
       userReview: setReview(fakeReview).payload,
     });
     return root.onSelectRating(5).then(() => {
-      expect(submitReview.called).toBeTruthy();
+      sinon.assert.called(submitReview);
 
       const call = submitReview.firstCall.args[0];
       expect(call.reviewId).toBeTruthy();
@@ -183,7 +182,7 @@ describe(__filename, () => {
       addon,
     });
     return root.onSelectRating(newReview.rating).then(() => {
-      expect(submitReview.called).toBeTruthy();
+      sinon.assert.called(submitReview);
 
       // Make sure the review is submitted in a way where it will be
       // newly created against the current version.
@@ -200,7 +199,7 @@ describe(__filename, () => {
     const FakeAddonReview = sinon.spy(() => <div />);
     const root = render({ AddonReview: FakeAddonReview, userReview });
 
-    expect(FakeAddonReview.called).toEqual(false);
+    sinon.assert.notCalled(FakeAddonReview);
 
     return root.onSelectRating(5).then(() => {
       sinon.assert.called(FakeAddonReview);
@@ -246,10 +245,10 @@ describe(__filename, () => {
     const userId = null; // logged out
     const root = render({ AddonReview: FakeAddonReview, userReview, userId });
 
-    expect(FakeAddonReview.called).toEqual(false);
+    sinon.assert.notCalled(FakeAddonReview);
 
     return root.onSelectRating(5).then(() => {
-      expect(FakeAddonReview.called).toBeFalsy();
+      sinon.assert.notCalled(FakeAddonReview);
     });
   });
 
@@ -259,7 +258,7 @@ describe(__filename, () => {
 
     const root = render({ UserRating: UserRatingStub, userReview });
 
-    expect(UserRatingStub.called).toEqual(true);
+    sinon.assert.called(UserRatingStub);
     const props = UserRatingStub.firstCall.args[0];
     expect(props.onSelectRating).toEqual(root.onSelectRating);
     expect(props.review).toEqual(userReview);
@@ -270,7 +269,7 @@ describe(__filename, () => {
 
     render({ UserRating: UserRatingStub, userReview: null });
 
-    expect(UserRatingStub.called).toEqual(true);
+    sinon.assert.called(UserRatingStub);
     const props = UserRatingStub.firstCall.args[0];
     expect(props.rating).toBe(undefined);
   });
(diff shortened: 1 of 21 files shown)


willdurand:
rebased.

willdurand:
re-rebased 😄

willdurand:
re-re-rebased 😅

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -101,9 +101,8 @@ describe(__filename, () => {
       loadSavedReview,
     });
 
-    expect(loadSavedReview.called).toEqual(true);
-    const args = loadSavedReview.firstCall.args[0];
-    expect(args).toEqual({
+    sinon.assert.called(loadSavedReview);
```
kumar303: Whenever you use `assert.calledWith(theFunction, ...)`, it's not necessary to precede it with `assert.called(theFunction)`. In fact, it's worse because the first test failure you would see is not as informative.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -392,9 +391,8 @@ describe(__filename, () => {
           .returns(Promise.resolve({ ...fakeReview, ...params }));
 
         return actions.submitReview(params).then(() => {
-          expect(dispatch.called).toEqual(true);
-          const action = dispatch.firstCall.args[0];
-          expect(action).toEqual(setReview(fakeReview));
+          sinon.assert.called(dispatch);
```
kumar303: This should be removed since there is a `calledWith` after it

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -424,8 +422,8 @@ describe(__filename, () => {
           })
           .then(() => {
             mockApi.verify();
-            expect(dispatch.called).toEqual(true);
-            expect(dispatch.firstCall.args[0]).toEqual(setReview(fakeReview));
+            sinon.assert.called(dispatch);
```
kumar303: This should be removed

tests/unit/core/TestAddonManager.js
```diff
@@ -125,70 +124,68 @@ describe(__filename, () => {
...
+        await addonManager.install(fakeInstallUrl, fakeCallback, {
           _mozAddonManager: fakeMozAddonManager,
           src: 'home',
-        })
-        .then(unexpectedSuccess, () =>
-          expect(fakeInstallObj.install.called).toBeTruthy(),
-        );
+        });
+
+        done.fail(new Error('this should not be called'));
```
kumar303: This pattern is a bit clunky to me. I think it's more succinct in promise form (as the test was before) and it's less prone to errors such as forgetting to call `done()`.

tests/unit/core/testErrorHandler.js
```diff
@@ -174,8 +174,9 @@ describe(__filename, () => {
       const error = new Error();
       errorHandler.handle(error);
 
-      expect(store.dispatch.called).toBeTruthy();
-      expect(store.dispatch.firstCall.args[0]).toEqual(
+      sinon.assert.called(store.dispatch);
```
kumar303: This can be removed

tests/unit/core/testErrorHandler.js
```diff
@@ -357,8 +358,9 @@ describe(__filename, () => {
     it('dispatches an error', () => {
       const error = new Error();
       errorHandler.handle(error);
-      expect(errorHandler.dispatch.called).toBeTruthy();
-      expect(errorHandler.dispatch.firstCall.args[0]).toEqual(
+      sinon.assert.called(errorHandler.dispatch);
```
kumar303: This can be removed

tests/unit/core/testErrorHandler.js
```diff
@@ -396,8 +398,9 @@ describe(__filename, () => {
 
     it('clears an error', () => {
       errorHandler.clear();
-      expect(errorHandler.dispatch.called).toBeTruthy();
-      expect(errorHandler.dispatch.firstCall.args[0]).toEqual(
+      sinon.assert.called(errorHandler.dispatch);
```
kumar303: This can be removed

tests/unit/core/utils/test_index.js
```diff
@@ -610,16 +608,15 @@ describe(__filename, () => {
...
+      sinon.assert.calledWith(_config.get, configKey);
     });
 
     it('passes through component and props when enabled', () => {
       const _config = { get: () => true };
       const SomeComponent = sinon.spy(() => <div />);
       render({ color: 'orange', size: 'large' }, { SomeComponent, _config });
 
-      expect(SomeComponent.called).toBeTruthy();
+      sinon.assert.called(SomeComponent);
```
kumar303: I assume you held off on converting this call because it was only partially matching the arguments? For cases like that, use `sinon.match()` like this:

```diff
diff --git a/tests/unit/core/utils/test_index.js b/tests/unit/core/utils/test_index.js
index 74c780123..44396a0b7 100644
--- a/tests/unit/core/utils/test_index.js
+++ b/tests/unit/core/utils/test_index.js
@@ -616,10 +616,10 @@ describe(__filename, () => {
       const SomeComponent = sinon.spy(() => <div />);
       render({ color: 'orange', size: 'large' }, { SomeComponent, _config });
 
-      sinon.assert.called(SomeComponent);
-      const props = SomeComponent.firstCall.args[0];
-      expect(props.color).toEqual('orange');
-      expect(props.size).toEqual('large');
+      sinon.assert.calledWith(
+        SomeComponent,
+        sinon.match({ color: 'orange', size: 'large' }),
+      );
     });
   });
 
```

However, in this case I think matching all call arguments would work fine:

```diff
diff --git a/tests/unit/core/utils/test_index.js b/tests/unit/core/utils/test_index.js
index 74c780123..dfed5e860 100644
--- a/tests/unit/core/utils/test_index.js
+++ b/tests/unit/core/utils/test_index.js
@@ -616,10 +616,10 @@ describe(__filename, () => {
       const SomeComponent = sinon.spy(() => <div />);
       render({ color: 'orange', size: 'large' }, { SomeComponent, _config });
 
-      sinon.assert.called(SomeComponent);
-      const props = SomeComponent.firstCall.args[0];
-      expect(props.color).toEqual('orange');
-      expect(props.size).toEqual('large');
+      sinon.assert.calledWith(
+        SomeComponent,
+        { color: 'orange', size: 'large' },
+      );
     });
   });
 
```

It's usually a good idea to partially match call arguments so that the test is only checking exactly what it needs to. Otherwise, making an unrelated change to a function signature could force you to update more tests than you need to.

tests/unit/ui/components/TestRating.js
```diff
@@ -65,40 +65,40 @@ describe(__filename, () => {
     const onSelectRating = sinon.stub();
     const root = render({ onSelectRating });
     selectRating(root, 1);
-    expect(onSelectRating.called).toEqual(true);
-    expect(onSelectRating.firstCall.args[0]).toEqual(1);
+    sinon.assert.called(onSelectRating);
```
kumar303: This can be removed

tests/unit/ui/components/TestRating.js
```diff
@@ -65,40 +65,40 @@ describe(__filename, () => {
...
+    sinon.assert.calledWith(onSelectRating, 1);
   });
 
   it('lets you select a two star rating', () => {
     const onSelectRating = sinon.stub();
     const root = render({ onSelectRating });
     selectRating(root, 2);
-    expect(onSelectRating.called).toEqual(true);
-    expect(onSelectRating.firstCall.args[0]).toEqual(2);
+    sinon.assert.called(onSelectRating);
```
kumar303: This can be removed

tests/unit/ui/components/TestRating.js
```diff
@@ -65,40 +65,40 @@ describe(__filename, () => {
...
+    sinon.assert.calledWith(onSelectRating, 2);
   });
 
   it('lets you select a three star rating', () => {
     const onSelectRating = sinon.stub();
     const root = render({ onSelectRating });
     selectRating(root, 3);
-    expect(onSelectRating.called).toEqual(true);
-    expect(onSelectRating.firstCall.args[0]).toEqual(3);
+    sinon.assert.called(onSelectRating);
```
kumar303: This can be removed

tests/unit/ui/components/TestRating.js
```diff
@@ -65,40 +65,40 @@ describe(__filename, () => {
...
+    sinon.assert.calledWith(onSelectRating, 3);
   });
 
   it('lets you select a four star rating', () => {
     const onSelectRating = sinon.stub();
     const root = render({ onSelectRating });
     selectRating(root, 4);
-    expect(onSelectRating.called).toEqual(true);
-    expect(onSelectRating.firstCall.args[0]).toEqual(4);
+    sinon.assert.called(onSelectRating);
```
kumar303: This can be removed

tests/unit/ui/components/TestRating.js
```diff
@@ -65,40 +65,40 @@ describe(__filename, () => {
...
+    sinon.assert.calledWith(onSelectRating, 4);
   });
 
   it('lets you select a five star rating', () => {
     const onSelectRating = sinon.stub();
     const root = render({ onSelectRating });
     selectRating(root, 5);
-    expect(onSelectRating.called).toEqual(true);
-    expect(onSelectRating.firstCall.args[0]).toEqual(5);
+    sinon.assert.called(onSelectRating);
```
kumar303: This can be removed

tests/unit/core/TestAddonManager.js
willdurand: reverted

PR #5584, opened 2018-07-12
Avoid scrollbar in user profile name

willdurand:
This is a slightly better fix for #5494 that avoids a scroll bar to be displayed.


willdurand:
> I am wondering if we should fix this too... I think if you use a similar style here too it will look better. What do you think?

yup, fixed it too, thanks!

Issue #5557, opened 2018-07-10
Implement dismissible Photon notices


kumar303:
Photon [message bars](https://design.firefox.com/photon/components/message-bars.html) can be dismissed. We need to implement the **X** button for this.

This is required for https://github.com/mozilla/addons-frontend/issues/5438

PR #5591, opened 2018-07-12
[WIP] Implement dismissible Photon notices #5557

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5557

WIP


src/core/withUIState.js
```diff
@@ -72,15 +72,25 @@ export const mergeUIStateProps = (
...
 
+type withUIStateParams = {|
+  // This should always be set to __filename for ID purposes.
+  fileName: string,
+  // A function that takes component props and returns a string to identify this state.
+  extractId: ExtractIdFunc,
+  // An Object that defines the initial state.
+  initialState: Object,
+  // When true, new component instances will be reset to initialState.
+  newStatePerInstance?: boolean,
```
willdurand: I don't really understand why this is need. It totally breaks the fact that `setUIState()` is supposed to mimic `setState()`.

I don't know why one would want to share state between instances that are different, but I don't think it should be stored in this reducer anyway, because that would obviously be not feasible with `setState()`.

kumar303: If you dismiss a notice with `id=survey`, go to another page, and come back to the same page, React creates a new instance of the survey notice. It needs to remember that the notice was dismissed.

My thinking was that `{ newStatePerInstance: false }` is where `withUIState` becomes more of an all-purpose Redux abstraction around state when it pertains only to a single component. By default, `withUIState` would still mimic `this.setState()` -- this will prevent surprises when converting older code.

Anyway, I still need to think this one through but as it stands, `withUIState()` cannot be used to track notice dismissals which seems a shame.

I actually think the pattern of relying on `constructor() { this.state = {} }` is sloppy because it means you don't have to care about changing state for any action that unmounts the component. This was the problem in `EditableCollectionAddon`; it should have been resetting its state [after notes get saved](https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/EditableCollectionAddon/index.js#L94) but it was not. This feels like a bug to me. It only worked because of `constructor() { this.state = {} }`.

kumar303: I can take this part out for now just to keep this patch simpler. It may be that the `this.setState()` behavior is fine for dismissing most other notices. Maybe the survey notice code will just need a good ol' custom reducer.

tests/unit/core/testErrorHandler.js
```diff
@@ -54,7 +54,9 @@ function createWrappedComponent({
 
   const tree = renderIntoDocument(
     <I18nProvider i18n={fakeI18n()}>
-      <ComponentWithErrorHandling store={store} {...customProps} />
+      <Provider store={store}>
+        <ComponentWithErrorHandling {...customProps} />
+      </Provider>
```
willdurand: 👻

PR #5592, opened 2018-07-12
I5364 updating about this theme logic

### LINKED ISSUE #5364: "About this theme" should be displayed for static themes
Moving https://github.com/mozilla/addons-server/issues/8554 to front-end
STR:
1. Load AMO dev and go to a static theme's listed page
2. Observe that the same text is displayed at the summary (text at the left of the Install Button and into the "Above this theme" section)

Expected result:
"About this theme" section is not displayed.

Actual result:
About this theme section is available and repeats the same description as the summary on the left side of the Install button.

Notes:
about this theme

rebmullin:
fixes #5364

## Before (lightweight themes)
Alt text

## After (lightweight themes)
Alt text

## Before (static themes - without a description)
Alt text

## After (static themes - without a description)
Alt text

## Before / After (static themes - with a description) 
Alt text

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -262,7 +262,7 @@ export class AddonBase extends React.Component {
 
     let title;
     const descriptionProps = {};
-    let hideAbout = false;
+    let showAbout = true;
 
     if (addon) {
       switch (addon.type) {
@@ -289,10 +289,11 @@ export class AddonBase extends React.Component {
       const description = addon.description ? addon.description : addon.summary;
       // For any theme type, we want to hide the summary text here since that is
       // already displayed in the header.
-      hideAbout =
+      showAbout = !(
         (description === addon.summary &&
           addon.type === ADDON_TYPE_STATIC_THEME) ||
-        addon.type === ADDON_TYPE_THEME;
+        addon.type === ADDON_TYPE_THEME
+      );
 
       if (!description || !description.length) {
         return null;
@@ -303,7 +304,7 @@ export class AddonBase extends React.Component {
       descriptionProps.children = <LoadingText width={100} />;
     }
 
-    return !hideAbout ? (
+    return showAbout ? (
       <ShowMoreCard header={title} className="AddonDescription">
         <div
           className="AddonDescription-contents"
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -289,11 +289,10 @@ export class AddonBase extends React.Component {
       const description = addon.description ? addon.description : addon.summary;
       // For any theme type, we want to hide the summary text here since that is
       // already displayed in the header.
-      showAbout = !(
-        (description === addon.summary &&
+      showAbout =
+        ((description === addon.summary &&
           addon.type === ADDON_TYPE_STATIC_THEME) ||
-        addon.type === ADDON_TYPE_THEME
-      );
+          addon.type === ADDON_TYPE_THEME) === false;
 
       if (!description || !description.length) {
         return null;
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -262,6 +262,7 @@ export class AddonBase extends React.Component {
 
     let title;
     const descriptionProps = {};
+    let hideAbout = false;
 
     if (addon) {
       switch (addon.type) {
@@ -286,6 +287,13 @@ export class AddonBase extends React.Component {
       }
 
       const description = addon.description ? addon.description : addon.summary;
+      // For any theme type, we want to hide the summary text here since that is
+      // already displayed in the header.
+      hideAbout =
+        (description === addon.summary &&
+          addon.type === ADDON_TYPE_STATIC_THEME) ||
+        addon.type === ADDON_TYPE_THEME;
+
       if (!description || !description.length) {
         return null;
       }
@@ -295,7 +303,7 @@ export class AddonBase extends React.Component {
       descriptionProps.children = <LoadingText width={100} />;
     }
 
-    return (
+    return !hideAbout ? (
       <ShowMoreCard header={title} className="AddonDescription">
         <div
           className="AddonDescription-contents"
@@ -305,7 +313,7 @@ export class AddonBase extends React.Component {
           {...descriptionProps}
         />
       </ShowMoreCard>
-    );
+    ) : null;
   }
 
   renderVersionReleaseNotes() {
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -735,8 +735,8 @@ describe(__filename, () => {
 
     expect(root.find('.AddonDescription')).toHaveLength(1);
 
-    expect(root.find('.AddonDescription-contents').html()).toContain(
-      description,
+    expect(root.find('.AddonDescription-contents')).toHaveHTML(
+      `<div class="AddonDescription-contents">${description}</div>`,
     );
   });
 
@@ -766,7 +766,9 @@ describe(__filename, () => {
 
     expect(root.find('.AddonDescription')).toHaveLength(1);
 
-    expect(root.find('.AddonDescription-contents').html()).toContain(summary);
+    expect(root.find('.AddonDescription-contents')).toHaveHTML(
+      `<div class="AddonDescription-contents">${summary}</div>`,
+    );
   });
 
   it("displays an extension's description on the show more card if there are both description and summary", () => {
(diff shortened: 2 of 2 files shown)


src/amo/components/Addon/index.js
```diff
@@ -295,7 +303,7 @@ export class AddonBase extends React.Component {
       descriptionProps.children = <LoadingText width={100} />;
     }
 
-    return (
+    return !hideAbout ? (
```
willdurand: Maybe we should invert the condition and use a variable like `showAboutCard`? It seems weird to negate this variable to follow the happy path.

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +711,83 @@ describe(__filename, () => {
     expect(rootNode.querySelector('.AddonDescription')).toEqual(null);
   });
 
+  it("does not display a lightweight theme's show more card", () => {
```
willdurand: Isn't is specific to the add-on description? If so, I'd update the test description and find `.AddonDescription`.

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +711,83 @@ describe(__filename, () => {
...
+        ...fakeAddon,
+        type: ADDON_TYPE_EXTENSION,
+        summary: 'my theme is very cool',
+        description,
+      }),
+    });
+
+    expect(root.find(ShowMoreCard)).toHaveLength(1);
+
+    expect(root.find('.AddonDescription-contents').html()).toContain(
```
willdurand: Would `toHaveHTML()` work here?

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +711,83 @@ describe(__filename, () => {
...
+  it("does not display a lightweight theme's show more card", () => {
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+        summary: 'my theme is very cool',
+      }),
+    });
+
+    expect(root.find(ShowMoreCard)).toHaveLength(0);
```
willdurand: I think we should find `.AddonDescription`, it is more explicit IMO.

tests/unit/amo/components/TestAddon.js
```diff
@@ -630,18 +631,6 @@ describe(__filename, () => {
...
-  it('sets a title for the description of a lightweight theme', () => {
-    const root = shallowRender({
-      addon: createInternalAddon({
-        ...fakeAddon,
-        type: ADDON_TYPE_THEME,
-      }),
-    });
-    expect(root.find('.AddonDescription').prop('header')).toContain(
-      'About this theme',
-    );
```
willdurand: Do we have such a test case somewhere else?

src/amo/components/Addon/index.js
rebmullin: I actually had the but it seemed that I had to check for more things that way...

but I'll double ck this again..

willdurand: ah I see what you mean, then maybe invert the condition when you assign the value, but not here.

tests/unit/amo/components/TestAddon.js
rebmullin: well it's the whole card:

```javascript
  return showAbout ? (
      <ShowMoreCard header={title} className="AddonDescription">
        <div
          className="AddonDescription-contents"
          ref={(ref) => {
            this.addonDescription = ref;
          }}
          {...descriptionProps}
        />
      </ShowMoreCard>
    ) : null;
```javascript

should I still be checking specifically for the description?

File: tests/unit/amo/components/TestAddon.js (hunk 4 above)
Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: oh ok, I see...

File: tests/unit/amo/components/TestAddon.js (hunk 2 above)
Comment (willdurand) [reply tests/unit/amo/components/TestAddon.js]: Ah sorry if I was not clear. If we introduce a second `ShowMoreCard`, then we might have issues. You should target this specific `ShowMoreCard` _via_ its classname.

File: tests/unit/amo/components/TestAddon.js (hunk 5 above)
Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: well I broke it up into types... lightweight theme type is under this test name now:

`does not display a lightweight theme's show more card`

File: tests/unit/amo/components/TestAddon.js (hunk 3 above)
Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: oh?! let me check : )

File: tests/unit/amo/components/TestAddon.js (hunk 2 above)
Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: ahh...okay, I see... thanks

File: tests/unit/amo/components/TestAddon.js (hunk 5 above)
Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: will that work?

Comment (willdurand) [reply tests/unit/amo/components/TestAddon.js]: oops, I am sorry. I meant "assertion" and not "test case". We need to check the `header` prop once at least.

Comment (rebmullin) [reply tests/unit/amo/components/TestAddon.js]: ok if the header is part of the showMore card which is not used in this case, then I am not sure how to check this in a different way than I already have:

```javascript
it("does not display a lightweight theme's show more card", () => {
    const root = shallowRender({
      addon: createInternalAddon({
        ...fakeAddon,
        type: ADDON_TYPE_THEME,
        summary: 'my theme is very cool',
      }),
    });

    expect(root.find('.AddonDescription')).toHaveLength(0);
  });
```

src/amo/components/Addon/index.js
```diff
@@ -286,6 +287,14 @@ export class AddonBase extends React.Component {
       }
 
       const description = addon.description ? addon.description : addon.summary;
+      // For any theme type, we want to hide the summary text here since that is
+      // already displayed in the header.
+      showAbout = !(
```
willdurand: Could we explicit this `!`? Maybe we could write `(condition ...) === false`?

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +710,85 @@ describe(__filename, () => {
     expect(rootNode.querySelector('.AddonDescription')).toEqual(null);
   });
 
+  it("does not display a lightweight theme's show more card", () => {
```
willdurand: > does not display a lightweight theme's description

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +710,85 @@ describe(__filename, () => {
...
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+        summary: 'my theme is very cool',
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it("displays a static theme's description on the show more card", () => {
```
willdurand: > displays a static theme's description

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +710,85 @@ describe(__filename, () => {
...
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(1);
+
+    expect(root.find('.AddonDescription-contents')).toHaveHTML(
+      `<div class="AddonDescription-contents">${description}</div>`,
+    );
+  });
+
+  it("does not display a static theme's show more card if there is no description", () => {
```
willdurand: > does not display anything if a static theme has no description

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +710,85 @@ describe(__filename, () => {
...
+        type: ADDON_TYPE_STATIC_THEME,
+        summary: 'my theme is very cool',
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it("displays an extension's summary on the show more card", () => {
```
willdurand: > displays the extension's summary when there is no description

tests/unit/amo/components/TestAddon.js
```diff
@@ -746,6 +710,85 @@ describe(__filename, () => {
...
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(1);
+
+    expect(root.find('.AddonDescription-contents')).toHaveHTML(
+      `<div class="AddonDescription-contents">${summary}</div>`,
+    );
+  });
+
+  it("displays an extension's description on the show more card if there are both description and summary", () => {
```
willdurand: > displays the extension's description when both description and summary are supplied

tests/unit/amo/components/TestAddon.js
willdurand: mmh what I wanted is an assertion for `header === 'About this theme'` but now I realize there are test cases for that above in the file (not shown in the GitHub diff panel), and we don't want that anyway for lightweight themes.

Nevermind 😄

rebmullin: ok cool : )

src/amo/components/Addon/index.js
rebmullin: ahh sure..okay!

PR #5595, opened 2018-07-13
SurveyNotice placeholder

kumar303:
See https://github.com/mozilla/addons-frontend/issues/5438

Because we're up against the clock, I'd like to commit this placeholder to master and do an L10n extraction, just to give our localizers a jump start. We will need the survey to go live, fully localized, next Thursday.


src/amo/components/SurveyNotice/index.js
```diff
@@ -0,0 +1,41 @@
...
+  const { i18n } = props;
+  return (
+    <Notice
+      actionHref="https://qsurvey.mozilla.com/s3/addons-mozilla-org-survey"
+      actionText={i18n.gettext('Take short survey')}
+      dismissible
+      type="generic"
+    >
+      {i18n.gettext(
+        'Thanks for visiting this site! Please take a minute or two to tell Firefox about your experience.',
```
willdurand: Firefox or Mozilla?

kumar303: That's the copy they requested

willdurand: Ok, I wouldn't expect to talk to a product (and did not know that Firefox was/had a persona) but that's fine, I guess.

kumar303:
> Ok, I wouldn't expect to talk to a product

Heh, yeah, the prose is a bit awkward as well (...a minute or two to...) but since this is a rush job and an ephemeral feature, I'm not too bothered.

PR #5597, opened 2018-07-13
I5550 converting confirm btn to use setUIState

### LINKED ISSUE #5550: Convert setState() in ConfirmButton to setUIState()
The [ConfirmButton](https://github.com/mozilla/addons-frontend/blob/54c4a4ddbfaea24c27620bead2254b183ba8bc61/src/ui/components/ConfirmButton/index.js) component can be converted to using [`setUIState()`](https://github.com/mozilla/addons-frontend/issues/5343).

It's only using `setState()` to show a prompt about what the user is about to confirm.

**For QA**: please make sure that confirm buttons on the collection and user profile screens still work as expected.

rebmullin:
fixes #5550 

I wasn't totally clear if I need to do anything to the HOC. I saw this component in 2 places - the collections and the user profile - and it looks like it works with the changes in both places still. 

let me know what you think : )

### DIFF
--- a/src/ui/components/ConfirmButton/index.js
+++ b/src/ui/components/ConfirmButton/index.js
@@ -3,7 +3,6 @@ import makeClassName from 'classnames';
 import * as React from 'react';
 import { compose } from 'redux';
 
-import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import withUIState from 'core/withUIState';
 import Button from 'ui/components/Button';
@@ -117,16 +116,11 @@ export class ConfirmButtonBase extends React.Component<InternalProps> {
   }
 }
 
-export const extractId = (ownProps: Props) => {
-  return `confirmButton-${ownProps.className || ''}`;
-};
-
 const ConfirmButton: React.ComponentType<Props> = compose(
   translate(),
-  withFixedErrorHandler({ fileName: __filename, extractId }),
   withUIState({
     fileName: __filename,
-    extractId,
+    extractId: () => '',
     initialState: initialUIState,
   }),
 )(ConfirmButtonBase);
--- a/src/ui/components/ConfirmButton/index.js
+++ b/src/ui/components/ConfirmButton/index.js
@@ -3,7 +3,9 @@ import makeClassName from 'classnames';
 import * as React from 'react';
 import { compose } from 'redux';
 
+import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
+import withUIState from 'core/withUIState';
 import Button from 'ui/components/Button';
 import type { I18nType } from 'core/types/i18n';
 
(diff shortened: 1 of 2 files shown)


src/ui/components/ConfirmButton/index.js
```diff
@@ -118,8 +117,18 @@ export class ConfirmButtonBase extends React.Component<InternalProps, State> {
   }
 }
 
-const ConfirmButton: React.ComponentType<Props> = compose(translate())(
-  ConfirmButtonBase,
-);
+export const extractId = (ownProps: Props) => {
+  return `confirmButton-${ownProps.className}`;
```
rebmullin: I wasn't totally clear about this - will this work?

rebmullin: I guess i need to change this anyway since classname is not required :/

src/ui/components/ConfirmButton/index.js
```diff
@@ -118,8 +117,18 @@ export class ConfirmButtonBase extends React.Component<InternalProps, State> {
   }
 }
 
-const ConfirmButton: React.ComponentType<Props> = compose(translate())(
-  ConfirmButtonBase,
-);
+export const extractId = (ownProps: Props) => {
+  return `confirmButton-${ownProps.className || ''}`;
```
rebmullin: This doesn't feel ideal to me. Should I be passing in an id to the component? Or should I make this function optional? I didn't feel like that was a good idea either tho :/

src/ui/components/ConfirmButton/index.js
willdurand: mmh, do you really need an ID? I think you can write:

```
extractId: () => '',
```

in the HOC declaration.

rebmullin: ahh! ok that will work i think. I wasn't clear if I should be passing in *something*..

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -89,59 +97,60 @@ describe(__filename, () => {
...
-    root.find(Button).simulate('click', createFakeEvent());
-
-    expect(root.find('.ConfirmButton-cancel-button')).toHaveProp(
-      'buttonType',
-      cancelButtonType,
-    );
-    expect(root.find('.ConfirmButton-cancel-button').children()).toHaveText(
-      cancelButtonText,
-    );
-  });
```
willdurand: Why did you remove this test case?

rebmullin: hmm..good question...let me double check : )

rebmullin: I think I thought this was dupe but I see it's slightly different so I will add this back. thanks for catching this!

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -107,41 +117,59 @@ describe(__filename, () => {
...
+  it('changes UI state when the confirm button is clicked', () => {
+    const { store } = dispatchClientMetadata();
+    const root = render({ store });
+
+    expect(root.instance().props.uiState.showConfirmation).toEqual(false);
+
+    root.find(Button).simulate('click', createFakeEvent());
+    applyUIStateChanges({ root, store });
+
+    expect(root.instance().props.uiState.showConfirmation).toEqual(true);
```
willdurand: Not sure you really this test case, because it should be already tested by the previous test cases.

rebmullin: ok I was wondering but I saw an explicit test for this in the other place it was already set up. But I can def removed if it's overkill here..

willdurand: > vut I saw an explicit test for this in the other place it was already set up

ah, really? I looked too, and found nothing 😅

rebmullin: sorry I should have posted where I saw this (I was being lazy :/), but [here](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestEditableCollectionAddon.js#L224)

willdurand: mmh interesting. I'd argue that we should not be testing this because we should focus on the behaviors. If the value of this prop changes, then we should observe "something", in this case a text with two buttons should be either displayed or hidden.

But maybe @kumar303 has a different opinion.

PR #5601, opened 2018-07-14
Cleanup core

willdurand:
Refs #1260

---

- use `styles.scss` file


willdurand:
I obviously forgot to push/open this one.

willdurand:
> I couldn't find the PR but I thought we removed HoverIntent recently..? Should this be removed or is it possible we will need this?

mmh I don't know. I don't think it is useful to keep unused things. We can still get them back later if needed.

willdurand:
> I couldn't find the PR but I thought we removed HoverIntent recently..? Should this be removed or is it possible we will need this?

mmh actually you are right. I don't know what happened to `HoverIntent`..

willdurand:
@rebmullin sorry, I interpreted your comment as "I thought we already deleted `HoverIntent`" and I thought it was deleted here, but looks like not and did not find any commit with the removal of this component... so: https://github.com/mozilla/addons-frontend/pull/5618

willdurand:
ah well no. It was removed in #5582, which is not merged yet... 🤦‍♂️

PR #5602, opened 2018-07-14
Fix top-level describe block descriptions

### LINKED ISSUE #2928: Show test suite file names in failure output

willdurand:
Fix #2928 

---

This PR makes sure we always have a top-level `describe` block with `__filename` as a description.

Note: the diff is large because I added a few top-level `describe` blocks in test files where there were multiple `describe` top-level blocks (hence re-indenting almost all the code in the files).

### DIFF
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -1421,104 +1421,104 @@ describe(__filename, () => {
       expect(extractId(props)).toEqual('some-slug');
     });
   });
-});
 
-describe('mapStateToProps', () => {
-  let store;
+  describe('mapStateToProps', () => {
+    let store;
 
-  beforeEach(() => {
-    store = createStore().store;
-  });
+    beforeEach(() => {
+      store = createStore().store;
+    });
 
-  function signIn(params) {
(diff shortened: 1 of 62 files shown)


willdurand:
I used this eslint rule 🙈 https://github.com/willdurand/eslint-plugin-amo/blob/master/README.md#describe-with-filename

willdurand:
also rebased this one

tests/unit/amo/components/TestAddon.js
```diff
@@ -1423,7 +1423,7 @@ describe(__filename, () => {
   });
 });
 
-describe('mapStateToProps', () => {
+describe(__filename, () => {
```
rebmullin: here is the double occurrence...

Issue #5608, opened 2018-07-16
Only reviews are visible after editing the username


ioanarusiczki:
STR:
1. Load AMO dev and go to View My Profile , make sure your account displays content for "Extensions by",  "Themes by" and "My reviews" sections.
2. Click Edit profile and update your username.
3. Observe the profile page.

Expected result:
All sections are displayed.

Actual result:
Only "my reviews" is visible. 

Notes:
After refreshing the page all the sections are displayed. 
It is not reproducible when editing the display name.
This issue is reproducible on AMO dev and AMO stage with FF61(Win10). 
 
![only reviews](https://user-images.githubusercontent.com/33448286/42753156-a76eb280-88f8-11e8-8795-a51d68146bb9.gif)

willdurand:
Interesting, I cannot reproduce locally but I can reproduce it in -dev.

willdurand:
Ok, I think I found the issue. When updating a username, the search API does not return the add-ons to display in the two cards ("My extensions" and "My themes"), because it has not been re-indexed I suppose.

willdurand:
Filed an issue on addons-server: https://github.com/mozilla/addons-server/issues/8901

ioanarusiczki:
This is no longer reproducible on AMO dev with FF63 (Win10)

![editing the username](https://user-images.githubusercontent.com/33448286/47564285-24441f00-d92d-11e8-999b-176a85ac734f.gif)

PR #5609, opened 2018-07-16
Enable redux-devtools in -dev

### LINKED ISSUE #5622: Enable redux devtools in -dev
We'd like to enable the redux devtools in -dev (and later other environments).

For QA, in -dev only:

1. install a browser extension, see: https://github.com/zalmoxisus/redux-devtools-extension#installation.
  For Firefox on AMO: https://addons.mozilla.org/en-US/firefox/addon/remotedev/
2. browse: https://addons-dev.allizom.org/
3. open the browser devtools OR click the extension icon
4. you should see something like this:

screen shot 2018-07-16 at 23 32 06

willdurand:
Fix #5622

---

This PR introduces a new config paramter `enableDevTools` that can be
use to enable devtools in any environment.

I'd like to debug #5608 in -dev and I think it would be useful to have
the Redux devtools available in this environment. FWIW, Airbnb has the
Redux devtools in all envs, including -prod, which is super great.

As a first step, we can enable the devtools in -dev, and see if we need
to turn them on for -stage and/or -prod later.

I also removed a package that we are not using at all. We use the
following Redux devtools:
https://github.com/zalmoxisus/redux-devtools-extension, which does not
need any extra dependency. It is only enabled if the browser extension
is installed and opened.

### DIFF
--- a/tests/unit/core/test_store.js
+++ b/tests/unit/core/test_store.js
@@ -1,15 +1,14 @@
 import { middleware } from 'core/store';
+import { getFakeConfig } from 'tests/unit/helpers';
 
 describe(__filename, () => {
-  function configForDev(isDevelopment, config = {}) {
-    const finalConfig = { isDevelopment, ...config };
-
-    return {
-      get(key) {
-        return finalConfig[key];
-      },
-    };
-  }
+  const configForDev = (isDevelopment, config = {}) => {
+    return getFakeConfig({
+      isDevelopment,
+      server: false,
+      ...config,
+    });
+  };
 
   it('includes the middleware in development', () => {
     const _createLogger = sinon.stub();
@@ -19,7 +18,7 @@ describe(__filename, () => {
         _createLogger,
       }),
     ).toBe('function');
-    expect(_createLogger.called).toEqual(true);
+    sinon.assert.called(_createLogger);
   });
 
   it('does not apply middleware if not in development', () => {
@@ -30,7 +29,7 @@ describe(__filename, () => {
         _createLogger,
       }),
     ).toBe('function');
-    expect(_createLogger.called).toEqual(false);
+    sinon.assert.notCalled(_createLogger);
   });
 
   it('handles a falsey window while on the server', () => {
@@ -43,7 +42,7 @@ describe(__filename, () => {
         _window,
       }),
     ).toBe('function');
-    expect(_createLogger.called).toEqual(true);
+    sinon.assert.called(_createLogger);
   });
 
   it('does not create a logger for the server', () => {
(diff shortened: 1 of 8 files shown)


tests/unit/core/test_store.js
```diff
@@ -56,34 +57,33 @@ describe('core store middleware', () => {
...
+
     const createStore = () => {};
     expect(typeof enhancer(createStore)).toBe('function');
   });
 
-  it('adds the devtools store enhancer in development', () => {
+  it('adds the devtools store enhancer when config enables it', () => {
     const _window = {
-      devToolsExtension: sinon.spy((createStore) => createStore),
+      __REDUX_DEVTOOLS_EXTENSION__: sinon.spy((createStore) => createStore),
```
bobsilverberg: I'm not sure why we need to pass `(createStore) => createStore` into `sinon.spy`. It doesn't seem as if it's actually doing anything.

willdurand: I think it is only for reproducing the behavior of the function. Do you want me to remove it?

bobsilverberg: Well, I found it confusing, so if it doesn't actually do anything maybe it's better to remove it to avoid similar confusion in the future.

willdurand: 👍

tests/unit/core/test_store.js
```diff
@@ -3,6 +3,7 @@ import { middleware } from 'core/store';
 describe('core store middleware', () => {
   function configForDev(isDevelopment, config = {}) {
     const finalConfig = { isDevelopment, ...config };
+
```
kumar303: I know this is just old code but `configForDev` can use [`getFakeConfig()`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/helpers.js#L461) now.

tests/unit/core/test_store.js
```diff
@@ -56,34 +57,33 @@ describe('core store middleware', () => {
...
+  it('adds the devtools store enhancer when config enables it', () => {
     const _window = {
-      devToolsExtension: sinon.spy((createStore) => createStore),
+      __REDUX_DEVTOOLS_EXTENSION__: sinon.spy(),
     };
-    expect(typeof middleware({ _config: configForDev(true), _window })).toBe(
-      'function',
-    );
-    expect(_window.devToolsExtension.called).toEqual(true);
+    const _config = configForDev(true, { enableDevTools: true });
```
kumar303: since this is no longer dependent on `isDevelopment`, the test doesn't need to set that in the config.

PR #5610, opened 2018-07-16
Add UTM parameters on Disco Pane

### LINKED ISSUE #5346: Add UTM parameters to Disco Pane links
Per [this spec](https://docs.google.com/document/d/1zQddLWnKj4TyGOrEbECLlEENN0iljUjoBrGvDFz8yrI/edit), we should add UTM parameters to all links on disco pane, so they are easy to track in analytics dashboards.

There are 3 links I see that need this treatment.

Link to add-on listing on each Disco Pane entry:
* `utm_source=discovery.addons.mozilla.org`
* `utm_medium=firefox-browser`
* `utm_content=discopane-entry-link`
* `src=api` (not an UTM parameter, but it's used internally by AMO to track downloads in the stats dashboards for developers; looks like we dropped it at some point)

See More Add-ons button:
* `utm_source=discovery.addons.mozilla.org`
* `utm_medium=firefox-browser`
* `utm_content=see-more-link`
* `src=api` (not an UTM parameter, but it's used internally by AMO to track downloads in the stats dashboards for developers; looks like we dropped it at some point)

Link to Privacy Policy:
* `utm_source=discovery.addons.mozilla.org`
* `utm_medium=firefox-browser`
* `utm_content=privacy-policy-link`

The last one doesn't point to AMO, but the referral info could still be of use for people managing mozilla.org

willdurand:
Fix #5346

---

This PR adds UTM parameters on disco pane links (when possible).

### DIFF
--- a/tests/unit/disco/components/TestDiscoPane.js
+++ b/tests/unit/disco/components/TestDiscoPane.js
@@ -17,9 +17,10 @@ import { ErrorHandler } from 'core/errorHandler';
 import I18nProvider from 'core/i18n/Provider';
 import { createInternalAddon } from 'core/reducers/addons';
 import { getDiscoResults } from 'disco/actions';
-import createStore from 'disco/store';
 import { NAVIGATION_CATEGORY } from 'disco/constants';
 import * as helpers from 'disco/components/DiscoPane';
+import createStore from 'disco/store';
+import { makeQueryStringWithUTM } from 'disco/utils';
 import {
   createStubErrorHandler,
   fakeI18n,
@@ -335,6 +336,20 @@ describe(__filename, () => {
   });
 
   describe('See more add-ons link', () => {
+    it('shows a "more add-ons" link with UTM params', () => {
+      const root = render();
+      const button = root.find('.amo-link').find(Button);
+
+      expect(button).toHaveLength(1);
+      expect(button).toHaveProp(
+        'href',
+        `https://addons.mozilla.org/${makeQueryStringWithUTM({
+          utm_content: 'see-more-link',
+          src: 'api',
+        })}`,
+      );
+    });
+
     it('tracks see more addons link being clicked', () => {
       const root = render();
 
--- a/tests/unit/disco/components/TestFooter.js
+++ b/tests/unit/disco/components/TestFooter.js
@@ -1,16 +1,17 @@
 import * as React from 'react';
 
 import Footer, { FooterBase } from 'disco/components/Footer';
+import { makeQueryStringWithUTM } from 'disco/utils';
 import { fakeI18n, shallowUntilTarget } from 'tests/unit/helpers';
 
-function render(props = {}) {
-  return shallowUntilTarget(
-    <Footer i18n={fakeI18n()} {...props} />,
-    FooterBase,
-  );
-}
-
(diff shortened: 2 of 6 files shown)


rebmullin:
@willdurand, this needs to be rebased

willdurand:
@rebmullin rebased, thanks!

tests/unit/disco/components/TestDiscoPane.js
```diff
@@ -335,6 +335,17 @@ describe(__filename, () => {
...
 
   describe('See more add-ons link', () => {
+    it('shows a "more add-ons" link with UTM params', () => {
+      const root = render();
+      const button = root.find('.amo-link').find(Button);
+
+      expect(button).toHaveLength(1);
+      expect(button).toHaveProp(
+        'href',
+        'https://addons.mozilla.org/?utm_source=discovery.addons.mozilla.org&utm_medium=firefox-browser&utm_content=see-more-link&src=api',
```
bobsilverberg: Should we be hardcoding this expectation, or should we be using the `makeQueryStringWithUTM` function to generate it?

tests/unit/disco/components/TestFooter.js
```diff
@@ -22,7 +22,7 @@ describe('Footer', () => {
 
     expect(root.find('.Footer-privacy-link')).toHaveProp(
       'href',
-      'https://www.mozilla.org/privacy/websites/',
+      'https://www.mozilla.org/privacy/websites/?utm_source=discovery.addons.mozilla.org&utm_medium=firefox-browser&utm_content=privacy-policy-link',
```
bobsilverberg: Should we be hardcoding this expectation, or should we be using the `makeQueryStringWithUTM` function to generate it?

tests/unit/disco/test_utils.js
```diff
@@ -1,14 +1,39 @@
...
+    });
+  });
+
+  describe('makeQueryStringWithUTM', () => {
+    it('returns a valid query string', () => {
+      const utm_content = 'utm-content';
+      const queryString = makeQueryStringWithUTM({ utm_content });
+
+      expect(queryString).toEqual(
+        `?utm_source=discovery.addons.mozilla.org&utm_medium=firefox-browser&utm_content=${utm_content}`,
```
bobsilverberg: Because `makeQueryStringWithUTM` internally calls `makeQueryString` from `core/api`, we should probably also call that as part of the expectation. We know, for now, it's not going to return anything other than what we expect here, but `makeQueryStringWithUTM` doesn't know that, and this test is now relying on `makeQueryString` never changing.

tests/unit/disco/components/TestDiscoPane.js
willdurand: ahm, that's the thing I don't really know to be honest. I think, maybe I should use `makeQueryStringWithUTM()` yes, otherwise I am testing this function as well.

willdurand:
> For future reference, it makes things a bit easier to review if you add new commits to a PR rather than squashing things locally and replacing the original commit.

Sure thing. I try to avoid that and only push new commits, but started to rebase + fix at the same time :/

willdurand:
Thanks!

PR #5612, opened 2018-07-16
Upgrade react-transition-group

### LINKED ISSUE #5155: Update react-transition-group to the latest version and the components that use it

willdurand:
Fix #5155

---

This PR upgrades `react-transition-group` to its major version 2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,7 +204,7 @@
     "react-router-scroll": "0.4.4",
     "react-super-responsive-table": "4.2.4",
     "react-textarea-autosize": "7.0.3",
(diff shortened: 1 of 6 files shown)


rebmullin:
@willdurand, this needs a rebase..

willdurand:
@rebmullin ah right, good catch, thanks! rebased.

PR #5613, opened 2018-07-16
I5277 removes default static theme icon from reviews

### LINKED ISSUE #5277: The default icon is displayed for the static themes search suggestions
STR:
1. Load AMO dev and make a search for a static theme.
2. Observe the icon displayed below into the search suggestions.

Expected result:
The static theme's square image is visible.

Actual result:
The general icon (green icon) is displayed.

Notes:
This issue is also reproducible with collections - when making a Search for a static theme.

The General Search
static themes - icons while search

The search suggestions for the collection feature
icons in search suggestions collections

rebmullin:
fixes #5277

### DIFF
--- a/src/amo/components/AddonReviewList/index.js
+++ b/src/amo/components/AddonReviewList/index.js
@@ -1,5 +1,6 @@
 /* @flow */
 /* eslint-disable react/no-unused-prop-types */
+import makeClassName from 'classnames';
 import { oneLine } from 'common-tags';
 import * as React from 'react';
 import Helmet from 'react-helmet';
@@ -14,7 +15,6 @@ import { setViewContext } from 'amo/actions/viewContext';
 import { expandReviewObjects } from 'amo/reducers/reviews';
 import { fetchAddon, getAddonBySlug } from 'core/reducers/addons';
 import Paginate from 'core/components/Paginate';
-import { ADDON_TYPE_STATIC_THEME } from 'core/constants';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
@@ -248,7 +248,12 @@ export class AddonReviewListBase extends React.Component<Props> {
       ) : null;
 
     return (
-      <div className="AddonReviewList">
+      <div
+        className={makeClassName(
+          'AddonReviewList',
+          `AddonReviewList--${addon && addon.type}`,
+        )}
+      >
         {addon && (
           <Helmet>
             <title>{header}</title>
@@ -259,16 +264,13 @@ export class AddonReviewListBase extends React.Component<Props> {
 
         <Card className="AddonReviewList-addon">
           <div className="AddonReviewList-header">
-            {addon &&
-              addon.type !== ADDON_TYPE_STATIC_THEME && (
-                <div className="AddonReviewList-header-icon">
-                  {addon ? (
-                    <Link to={this.addonURL()}>{iconImage}</Link>
-                  ) : (
-                    iconImage
-                  )}
-                </div>
+            <div className="AddonReviewList-header-icon">
+              {addon ? (
+                <Link to={this.addonURL()}>{iconImage}</Link>
+              ) : (
+                iconImage
               )}
+            </div>
             <div className="AddonReviewList-header-text">
               <h1 className="visually-hidden">{header}</h1>
               <h2 className="AddonReviewList-header-addonName">{addonName}</h2>
--- a/src/amo/components/AddonReviewList/index.js
+++ b/src/amo/components/AddonReviewList/index.js
@@ -251,7 +251,7 @@ export class AddonReviewListBase extends React.Component<Props> {
       <div
         className={makeClassName(
           'AddonReviewList',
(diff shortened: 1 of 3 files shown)


tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -124,9 +125,6 @@ describe(__filename, () => {
     it('waits for an addon and reviews to load', () => {
       const location = fakeRouterLocation();
       const root = render({ addon: null, location });
-      expect(root.find('.AddonReviewList-header-icon img').prop('src')).toEqual(
```
rebmullin: since addon type is not known yet, this is hard to test unless i remove this like i've done here. Not sure if that's ideal tho... I am wondering if should revert this approach and just use styles to hide the icon and then we can leave this test? @willdurand , thoughts?

willdurand: Why do you want to remove it? It seems like you still want the `fallbackIcon`, no?

rebmullin: Static themes don't have their own icons and the default icon isn't supposed to be displayed. The [issue here](https://github.com/mozilla/addons-frontend/issues/5277#issuecomment-400533331) mentions white space but since it stands alone (vs in a list with other types), I was going to remove it and left align the text

willdurand: But when there is no addon yet, it should display the fallback icon, no?

Once the addon is loaded, you can either remove it or replace it with the actual addon's icon, but before I would keep the fallback icon.

rebmullin: You're right, it probably should keep the fallback when there is no addon. This failing test made me rethink my approach so I wanted a 2nd opinion. Cool tho, that helps me : ) So I'll add that back and will use css to hide the static theme icon. thank you!

src/amo/components/AddonReviewList/index.js
```diff
@@ -247,7 +248,12 @@ export class AddonReviewListBase extends React.Component<Props> {
       ) : null;
 
     return (
-      <div className="AddonReviewList">
+      <div
+        className={makeClassName(
+          'AddonReviewList',
+          `AddonReviewList--${(addon && addon.type) || ''}`,
```
willdurand: With `makeClassName`, you can have conditions with objects (the keys being the class names and the values tghe conditions):

```js
makeClassName('AddonReviewList, {
  'AddonReviewList--${addon.type}: addon && addon.type, 
})
```

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -468,6 +469,19 @@ describe(__filename, () => {
       expect(img).toHaveProp('src', addon.icon_url);
     });
 
+    it('renders a class name with its type', () => {
+      const addon = {
+        ...fakeAddon,
+        type: ADDON_TYPE_STATIC_THEME,
+      };
+      dispatchAddon(addon);
```
willdurand: You could probably move the `addon` above into the `dispatchAddon()`, like the test case after this one.

src/amo/components/AddonReviewList/index.js
rebmullin: I actually had something similar but slightly diff b/c it's a dynamic class name:
```javascript
   <div
        className={makeClassName('AddonReviewList', {
          [`AddonReviewList--${addon.type}`]: addon && addon.type,
        })}
      >
```
but flow complains about this :/

is it better to just hard code the class name..just for static theme type?

willdurand: What's the complaint?

willdurand: ah I think I know. Then:

```
  ...
  // Flow requires an extra check in the class name.
  [`AddonReviewList--${addon && addon.type}`]: addon && addon.type,
```

rebmullin:
```bash
Cannot get addon.type because property type is missing in null [1].

 [1]  40│   addon: AddonType | null,
        :
     250│     return (
     251│       <div
     252│         className={makeClassName('AddonReviewList', {
     253│           [`AddonReviewList--${addon.type}`]: addon && addon.type,
     254│         })}
     255│       >
     256│         {addon && (

```

rebmullin: ok let me check that. I thought I tried that too :/ one sec..

rebmullin: sorry this is the top part of the flow info:
`Cannot coerce addon && addon.type to string because null [1] should not be coerced.`

rebmullin: full response:
```bash
Cannot coerce addon && addon.type to string because null [1] should not be coerced.

 [1]  40│   addon: AddonType | null,
        :
     250│     return (
     251│       <div
     252│         className={makeClassName('AddonReviewList', {
     253│           [`AddonReviewList--${addon && addon.type}`]: addon && addon.type,
     254│         })}
     255│       >
     256│         {addon && (
```

willdurand: Well, then let's write maybe:

```
makeClassName('foo', addon && addon.type ? addon.type : null)
```

Issue #5508, opened 2018-07-05
Address UX issue for Edit Collection form for Collections:Contribute use case


bobsilverberg:
This originated from a bug report at https://github.com/mozilla/addons-frontend/issues/5499.

There is a permission called `Collections:Contribute` which allows a user to edit the list of add-ons that exists inside the `featured-personas` collection. It does not, however, allow a user to edit that collection's details (e.g., change the name, slug, etc.).

A user with that permission is able to access the Edit page for that collection, because that's where the functions are that allow a user to edit the add-ons in a collection. Those functions also currently work, i.e., a user can add a new add-on, remove an existing add-on, or change the notes for an existing add-on. 

The issue is that the user is not allowed to edit the collection's fields, so if/when they click the "Save collection" button they will see an error message. It seems like we should remove or disable the "Save" button in this case, and probably also the "Cancel" button, but that leaves a user with no way to navigate back to the Collection Details view other than hitting the back button. I think we should add something to the screen (maybe a "Back to details" button) which would allow the user to click to return to the details view.

What do you think @pwalm?

willdurand:
How about hiding the card completely? Or, better, displaying the "read-only" version of it like on a regular collection page (and not with disabled fields)?

bobsilverberg:
@willdurand That's an even better idea that what we discussed in the meeting. I think I'll do that!

pwalm:
Yep, display the "read-only" version

bobsilverberg:
The UI for this is ready and has been submitted via PR #5636, but there is still a problem which has been reported at https://github.com/mozilla/addons-server/issues/8747. It is still not possible for a user with one of the above permissions to add an add-on to a collection when it should be allowed. That's an API issue that will be addressed by https://github.com/mozilla/addons-server/issues/8747, hopefully soon.

ioanarusiczki:
This is verified fixed on AMO -dev with FF61(Win10)
The user with Collections:Contribute permission can return to the collection by clicking "Done Editing"

![ux issue collections contribute](https://user-images.githubusercontent.com/33448286/44210433-1de83880-a16f-11e8-87a8-bef1c5ed8973.gif)

The https://github.com/mozilla/addons-frontend/issues/5935 however was reopened because it seems there are some problems after adding or removing add-ons for the featured-personas collection.

PR #5619, opened 2018-07-16
Create a CollectionDetails component that can be shared between Collection and CollectionManager

### LINKED ISSUE #5570: Create a CollectionDetails component that can be shared between Collection and CollectionManager
To support issue #5508 we're going to want to display the collection details both from the `Collection` component and from the `CollectionManager` component, so the logic, which is currently in `Collection` should be extracted into its own component.

bobsilverberg:
Fixes #5570 

Note that although `hasEditPermission` no longer seems to be necessary, I will need it when I implement the changes to `CollectionManager` that are required to support #5508, so I don't really want to remove it now just to add it back in again in the next PR. It will allow the `CollectionManager` to know whether to display an edit form or the `CollectionDetails` component.


bobsilverberg:
You can look at PR #5636 to see how the `hasEditPermission` prop will be used.

src/amo/components/CollectionDetails/index.js
```diff
@@ -0,0 +1,90 @@
...
+  CollectionFilters,
+  CollectionType,
+} from 'amo/reducers/collections';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+export type Props = {|
+  collection: CollectionType | null,
+  editUrl: string,
```
kumar303: Requiring `editUrl` puts an unnecessary burden on the caller. `CollectionDetail` should be able to create an edit URL with just the `collection` object.

We've struggled with how to do this in the past. I think the best way is to introduce two new functions in `reducers/collections`: `collectionUrl()` and `editCollectionUrl()` (or named whatever).

There are a couple scenarios in `Collection` where it needs to create a URL without having a `collection` object but the function could allow for that by accepting `authorUsername` and `slug` values.

bobsilverberg: I agree, and will add the functions to this PR.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -1261,4 +1263,79 @@ describe(__filename, () => {
...
+          authorUsername: 'a different username',
+          collection,
+          collectionSlug: 'a-different-slug',
+        }),
+      ).toEqual(`/collections/${authorUsername}/${slug}/`);
+    });
+  });
+
+  describe('collectionEditUrl', () => {
+    it('returns an edit URL for a collection', () => {
```
kumar303: Rather than repeat all of the logic you've already tested, I think it's better just to test with dependency injection. This will help keep the tests maintainable as well as enforce a contract between the dependencies.

Example:

```js
const collectionEditUrl = ({
  authorUsername,
  slug,
  collection,
  _collectionUrl = collectionUrl
}) {
  return `${_collectionUrl(...)}edit/`;
}

// After that you really only need one test:

const _collectionUrl = sinon.spy(() => '/base/url/');
expect(collectionEditUrl({ ...params, _collectionUrl })).toEqual(
  '/base/url/edit/',
);
sinon.assert.calledWith(_collectionUrl, params);
```

PR #5620, opened 2018-07-16
Link to AMO experience survey

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5438

I had to change the headers on all pages and this revealed a couple inconsistencies (error pages, user profile, collections) which are now fixed.

screenshot 2018-07-16 16 15 19
screenshot 2018-07-16 16 15 33


src/core/reducers/survey.js
```diff
@@ -0,0 +1,45 @@
...
+
+export type SurveyState = {|
+  wasDismissed: boolean,
+|};
+
+export const initialState = {
+  wasDismissed: false,
+};
+
+type DismissSurveyAction = {|
```
bobsilverberg: It's consistent throughout the file, and it's a nice short name, but at first glance it seems like these actions affect a "Survey", whereas they really affect a `SurveyNotice`. We're not showing and dismissing a survey, we're showing and dismissing a notice. Maybe the names should be changed to be clearer, or maybe that just adds unnecessary letters.

src/disco/store.js
```diff
@@ -30,6 +31,7 @@ export default function createStore({
       infoDialog,
       redirectTo,
       routing,
+      survey,
```
bobsilverberg: Why do we need this for the disco pane? Are we showing the survey link there as well?

tests/unit/core/components/TestSurveyNotice.js
```diff
@@ -0,0 +1,102 @@
...
+    };
+    return shallowUntilTarget(<SurveyNotice {...props} />, SurveyNoticeBase);
+  };
+
+  it('configures a Notice', () => {
+    const root = render();
+
+    const notice = root.find(Notice);
+    expect(notice).toHaveProp('type', 'generic');
+    expect(notice).toHaveProp('dismissible', true);
```
bobsilverberg: Is it worth also asserting about other properties of `Notice`, such as `actionText` and `id`?

src/core/components/SurveyNotice/index.js
```diff
@@ -0,0 +1,106 @@
...
+    return null;
+  }
+
+  const onDismiss = () => {
+    // Even though a dismissal action is dispatched here, also save a
+    // cookie to manually synchronize state. The server code will load
+    // the cookie and synchronize state as part of the request.
+    // TODO: make this synchronization more automatic.
+    // See https://github.com/mozilla/addons-frontend/issues/5617
+    _cookie.save(_config.get('dismissedExperienceSurveyCookieName'), '', {
```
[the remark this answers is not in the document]
kumar303: Ah, yes. I started looking into that but Firefox nightly no longer lets you inspect cookie times so I forgot about it.

src/amo/components/CategoriesPage/styles.scss
```diff
@@ -2,7 +2,7 @@
 @import '~core/css/inc/mixins';
 
 .CategoriesPage {
-  margin: $padding-page;
+  @include page-padding();
```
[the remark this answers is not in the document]
kumar303: Yes, I was worried about that but I checked each one and they are fine -- none of them rely on margin collapse quirks.

src/core/reducers/api.js
```diff
@@ -64,6 +64,7 @@ export type UserAgentInfoType = {|
 
 export type ApiState = {
   clientApp: ?string,
+  // See config.get('langs') for all possible values.
```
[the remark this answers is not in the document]
kumar303: I didn't know what it was when I was looking at the type so I had to hunt it down through a lot of different calls. I wanted to go back in time and help past me understand it without the hunting part.

src/disco/store.js
kumar303: It's because the server reads the cookie and dispatches an action for all requests. I could wrap those calls in an if-amo statement but it would add more complexity and possibly introduce bugs in Discopane since the code path would be different. With this approach, the two apps are more in sync which means possibly less bugs.

I actually dislike the way Discopane runs with a limited set of reducers. It feels like a premature optimization. All it does for us right now is trigger potential bugs.

tests/unit/core/components/TestSurveyNotice.js
kumar303: I don't think so. Actually, this test was a placeholder and I should probably delete it. I don't think it's worth adding tests for statically configured properties. Once a property dynamically depends on conditional logic, it's worth adding a test.

src/core/reducers/survey.js
kumar303: The end result is still a survey so I think it makes sense.

src/core/components/SurveyNotice/index.js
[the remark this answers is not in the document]
kumar303: This is fixed now. I set it to expire in 180 days.

kumar303:
@bobsilverberg thanks for the review -- yes to both of your questions. The button now dismisses the survey and the link opens in a new tab. It's ready for another look.

src/core/components/SurveyNotice/index.js
```diff
@@ -0,0 +1,118 @@
...
+      .split('/')
+      .slice(2)
+      .join('/'),
+  });
+  const surveyUrl = `https://qsurvey.mozilla.com/s3/addons-mozilla-org-survey?${source}`;
+
+  return (
+    <Notice
+      actionHref={surveyUrl}
+      actionOnClick={actionOnClick}
```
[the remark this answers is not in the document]
kumar303: Ha. No, you are totally right. Thanks. It survived a bit of refactoring where before it was overly complex.

PR #5621, opened 2018-07-16
I5220-remove-addon/collection 404 fix

### LINKED ISSUE #5220: Page shows 404 when all addons in the second page are deleted
1. Editing a collection (an existing collection)
2. Having more than 25 addons in a collection, so that the pagination shows up.
3. Go to the second page, and delete them all.

### What happened?
It shows `Page not found`. 

Ideally, it should be redirected to previous page, page 1 in this case.

vlpebxfoo1

rebmullin:
fixes #5220 

I wanted to check into this approach before I changed any tests. I hope that's okay...

Also, just a note, originally I had a nested try catch inside the catch but I wasn't sure if this was ideal (based on various feedback when I looked into this a bit) plus I saw a few lint issues which are avoided when I do it this way. 

let me know what you think : )

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -53,6 +53,7 @@ import type { CollectionAddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
 import type { DispatchFunc } from 'core/types/redux';
 import type {
+  ReactRouterHistoryType,
   ReactRouterLocationType,
   ReactRouterMatchType,
 } from 'core/types/router';
@@ -76,6 +77,7 @@ type InternalProps = {|
   errorHandler: ErrorHandlerType,
   filters: CollectionFilters,
   hasEditPermission: boolean,
+  history: ReactRouterHistoryType,
   i18n: I18nType,
   isLoggedIn: boolean,
   isOwner: boolean,
@@ -247,9 +249,11 @@ export class CollectionBase extends React.Component<InternalProps> {
     invariant(slug, 'slug is required');
     invariant(username, 'username is required');
 
-    const pages = (collection.numberOfAddons - 1) / collection.pageSize;
-    const isNewPage = pages % 1 === 0;
-    const newFilter = Math.ceil(pages) || 1;
+    const { pageSize } = collection;
+    const pages = pageSize ? (collection.numberOfAddons - 1) / pageSize : null;
+    const isNewPage = pages && pages % 1 === 0;
+    const newFilter = (pages && Math.ceil(pages)) || 1;
+    const filter = isNewPage ? newFilter : filters.page;
 
     dispatch(
       removeAddonFromCollection({
@@ -257,7 +261,7 @@ export class CollectionBase extends React.Component<InternalProps> {
         errorHandlerId: errorHandler.id,
         filters: {
           ...filters,
-          page: isNewPage ? newFilter : filters.page,
+          page: filter,
         },
         slug,
         username,
@@ -268,10 +272,10 @@ export class CollectionBase extends React.Component<InternalProps> {
       const { location } = history;
 
       history.push({
-        ...location,
+        pathname: location.pathname,
         query: {
           ...location.query,
-          page: newFilter,
+          page: filter,
         },
       });
     }
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -238,7 +238,7 @@ export class CollectionBase extends React.Component<InternalProps> {
   }
 
   removeAddon: RemoveCollectionAddonFunc = (addonId: number) => {
-    const { collection, dispatch, errorHandler, filters } = this.props;
+    const { collection, dispatch, errorHandler, filters, history } = this.props;
 
     invariant(collection, 'collection is required');
 
@@ -249,7 +249,7 @@ export class CollectionBase extends React.Component<InternalProps> {
 
     const pages = (collection.numberOfAddons - 1) / collection.pageSize;
     const isNewPage = pages % 1 === 0;
-    const filter = Math.ceil(pages) || 1;
+    const newFilter = Math.ceil(pages) || 1;
 
     dispatch(
       removeAddonFromCollection({
@@ -257,7 +257,7 @@ export class CollectionBase extends React.Component<InternalProps> {
         errorHandlerId: errorHandler.id,
         filters: {
           ...filters,
-          page: isNewPage ? filter : filters.page,
+          page: isNewPage ? newFilter : filters.page,
         },
         slug,
         username,
@@ -265,13 +265,13 @@ export class CollectionBase extends React.Component<InternalProps> {
     );
 
     if (isNewPage) {
-      const { location } = this.props.history;
+      const { location } = history;
 
       history.push({
         ...location,
         query: {
           ...location.query,
-          page: filter,
+          page: newFilter,
         },
       });
     }
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -104,6 +104,24 @@ export type SaveAddonNoteFunc = (
   notes: string,
 ) => void;
 
+// Let's find out what page we'll be on after we remove
+// the item from the collection.
+const getCollectionPage = (collection) => {
+  const { numberOfAddons, pageSize } = collection;
+
+  if (pageSize) {
+    return (numberOfAddons - 1) / pageSize;
+  }
+
+  return 1;
+};
+
+const switchCollectionPage = (collection) => {
+  // If the number is an integer, then we can
+  // switch the page.
+  return Number.isInteger(getCollectionPage(collection));
+};
+
 export class CollectionBase extends React.Component<InternalProps> {
   addonPlaceholderCount: number;
 
@@ -249,11 +267,13 @@ export class CollectionBase extends React.Component<InternalProps> {
     invariant(slug, 'slug is required');
     invariant(username, 'username is required');
 
-    const { pageSize } = collection;
-    const pages = pageSize ? (collection.numberOfAddons - 1) / pageSize : null;
-    const isNewPage = pages && pages % 1 === 0;
-    const newFilter = (pages && Math.ceil(pages)) || 1;
-    const filter = isNewPage ? newFilter : filters.page;
+    const isNewPage = switchCollectionPage(collection);
+
+    let page = filters.page;
+
+    if (isNewPage) {
+      page = Math.ceil(getCollectionPage(collection)) || 1;
+    }
 
     dispatch(
       removeAddonFromCollection({
@@ -261,7 +281,7 @@ export class CollectionBase extends React.Component<InternalProps> {
         errorHandlerId: errorHandler.id,
         filters: {
           ...filters,
-          page: filter,
+          page,
         },
         slug,
         username,
@@ -275,7 +295,7 @@ export class CollectionBase extends React.Component<InternalProps> {
         pathname: location.pathname,
         query: {
           ...location.query,
-          page: filter,
+          page,
         },
       });
     }
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -104,29 +104,17 @@ export type SaveAddonNoteFunc = (
   notes: string,
 ) => void;
 
-// This gets the new collection page.
-const getNewCollectionPage = (collection) => {
-  const { numberOfAddons, pageSize } = collection;
-
-  if (pageSize) {
-    return Math.ceil((numberOfAddons - 1) / pageSize) || 1;
-  }
-
-  return 1;
-};
-
-const shouldChangePage = (collection) => {
+export const computeNewCollectionPage = (
+  collection: CollectionType,
+): number => {
   const { numberOfAddons, pageSize } = collection;
 
   let page = 1;
-
   if (pageSize) {
-    page = (numberOfAddons - 1) / pageSize;
+    page = Math.ceil((numberOfAddons - 1) / pageSize);
   }
 
-  // If the number is an integer, then we should
-  // change the page.
-  return Number.isInteger(page);
+  return page || 1;
 };
 
 export class CollectionBase extends React.Component<InternalProps> {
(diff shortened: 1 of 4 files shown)


rebmullin:
Thanks for the feedback @bobsilverberg. I'll take another look into...

rebmullin:
@willdurand @kumar303, I've made some changes (per the first review). When you have some time, please let me know if this is any better : )

src/amo/components/Collection/index.js
```diff
@@ -247,15 +249,36 @@ export class CollectionBase extends React.Component<InternalProps> {
     invariant(slug, 'slug is required');
     invariant(username, 'username is required');
 
+    const { pageSize } = collection;
+    const pages = pageSize ? (collection.numberOfAddons - 1) / pageSize : null;
+    const isNewPage = pages && pages % 1 === 0;
+    const newFilter = (pages && Math.ceil(pages)) || 1;
```
willdurand: 3 lines of basic maths, yet I don't understand anything..

I would like to see a helper function above the class with a name explaining what it is supposed to do, and we should not mix `&&` and `||` and ternary operators. I would like to avoid the modulo 1 as well, because this is JS black magic.

src/amo/components/Collection/index.js
```diff
@@ -247,15 +249,36 @@ export class CollectionBase extends React.Component<InternalProps> {
...
+    const filter = isNewPage ? newFilter : filters.page;
```
willdurand: why not naming this variable `page`?

tests/unit/amo/components/TestCollection.js
```diff
@@ -1268,6 +1268,66 @@ describe(__filename, () => {
         username: detail.author.username,
       }),
     );
+
+    sinon.assert.notCalled(root.instance().props.history.push);
+  });
+
+  it('set up new filters and a page when removeAddon removes the last addon on the page', () => {
```
willdurand: s/set up new filters/updates the page

src/amo/components/Collection/index.js
rebmullin: yikes ok : )let me see if I can clarify and clean up a bit

rebmullin:
@willdurand, thanks for feedback. Please let me know if this is any better : )
thanks

src/amo/components/Collection/index.js
```diff
@@ -102,6 +104,24 @@ export type SaveAddonNoteFunc = (
   notes: string,
 ) => void;
 
+// Let's find out what page we'll be on after we remove
+// the item from the collection.
+const getCollectionPage = (collection) => {
```
willdurand: Should it be `getPreviousCollectionPage`?

src/amo/components/Collection/index.js
```diff
@@ -102,6 +104,24 @@ export type SaveAddonNoteFunc = (
...
+  const { numberOfAddons, pageSize } = collection;
+
+  if (pageSize) {
+    return (numberOfAddons - 1) / pageSize;
+  }
+
+  return 1;
+};
+
+const switchCollectionPage = (collection) => {
```
willdurand: `shouldChangePage` maybe?

src/amo/components/Collection/index.js
```diff
@@ -247,15 +267,38 @@ export class CollectionBase extends React.Component<InternalProps> {
     invariant(slug, 'slug is required');
     invariant(username, 'username is required');
 
+    const isNewPage = switchCollectionPage(collection);
+
+    let page = filters.page;
+
+    if (isNewPage) {
+      page = Math.ceil(getCollectionPage(collection)) || 1;
```
willdurand: Why do you need `Math.ceil` here? That should be done in `getCollectionPage`, no?

willdurand:
> Please let me know if this is any better : )

Yeah that looks definitely better!

src/amo/components/Collection/index.js
rebmullin: I was going to move it up there but then swithcCollectionPage would always return true here so I decided to leave this way. I guess I can make this into another function if that would make it more clear..?

src/amo/components/Collection/index.js
rebmullin: that works : )

src/amo/components/Collection/index.js
rebmullin: actually here we are trying to figure out the collection page...
let me think about this..

rebmullin: ok I think I am going to combine the ceil func inside this function (as you've suggested/asked about below) and so i've changed things a bit...

regarding this name, I decided to go with getNewCollectionPage as it's only called when its determined that we need to switch the page.

rebmullin:
hey @willdurand, can you please take another look at this one if you have time :)

src/amo/components/Collection/index.js
```diff
@@ -102,6 +104,31 @@ export type SaveAddonNoteFunc = (
...
+  return 1;
+};
+
+const shouldChangePage = (collection) => {
+  const { numberOfAddons, pageSize } = collection;
+
+  let page = 1;
+
+  if (pageSize) {
+    page = (numberOfAddons - 1) / pageSize;
```
willdurand: I see that we do that in the function above. Can we do merge the two functions?

```js
export const computeNewCollectionPage = (
  collection: CollectionType,
): number | null => {
  const { numberOfAddons, pageSize } = collection;

  let page = 1;
  if (pageSize) {
    page = (numberOfAddons - 1) / pageSize;
  }

  if (!Number.isInteger(page)) {
    return null;
  }

  return Math.ceil(page) || 1;
};
```

willdurand: In which case do we have `page` that is not an integer?

willdurand: So, I think we can do this:

```js
export const computeNewCollectionPage = (
  collection: CollectionType,
): number => {
  const { numberOfAddons, pageSize } = collection;

  let page = 1;
  if (pageSize) {
    page = Math.ceil((numberOfAddons - 1) / pageSize);
  }

  return page || 1;
};


```

willdurand: and use it like this:

```js
  removeAddon: RemoveCollectionAddonFunc = (addonId: number) => {
    const { collection, dispatch, errorHandler, filters, history } = this.props;

    invariant(collection, 'collection is required');

    const { slug, authorUsername: username } = collection;

    invariant(username, 'username is required');

    let { page } = filters;
    let shouldPushNewRoute = false;
    const newCollectionPage = computeNewCollectionPage(collection);

    if (page !== newCollectionPage) {
      page = newCollectionPage;
      shouldPushNewRoute = true;
    }

    dispatch(
      removeAddonFromCollection({
        addonId,
        filters: {
          ...filters,
          page,
        },
        slug,
        username,
      }),
    );

    if (shouldPushNewRoute) {
      const { location } = history;

      history.push({
        pathname: location.pathname,
        query: {
          ...location.query,
          page,
        },
      });
    }
  };

```

What do you think? That would prevent a push to happen for no reason, and also make the code more understandable, no?

tests/unit/amo/components/TestCollection.js
```diff
@@ -1268,6 +1268,66 @@ describe(__filename, () => {
         username: detail.author.username,
       }),
     );
+
+    sinon.assert.notCalled(root.instance().props.history.push);
+  });
+
+  it('updates the page when removeAddon removes the last addon on the page', () => {
+    const history = createFakeHistory();
```
willdurand: You probably want to use it, you should pass it a `location` that is created a few lines after:

``` js
const location = createFakeLocation({ query: { page, collection_sort: sort } });
const history = createFakeHistory({ location });
```

Then you can pass both `location` and `history` and use the latter in `assert.test.calledWith(histor.push, ...)`.

tests/unit/amo/components/TestCollection.js
```diff
@@ -1268,6 +1268,66 @@ describe(__filename, () => {
...
+    const authorUserId = 11;
+    const { store } = dispatchSignInActions({ userId: authorUserId });
+
+    const addons = createFakeCollectionAddons();
+    const addonId = addons[0].addon.id;
+    const detail = createFakeCollectionDetail({
+      authorId: authorUserId,
+      // This will simulate only 1 item on the 2nd page.
+      count: DEFAULT_API_PAGE_SIZE + 1,
```
willdurand: We probably need a test with `DEFAULT_API_PAGE_SIZE + 2`.

tests/unit/amo/components/TestCollection.js
```diff
@@ -1268,6 +1268,66 @@ describe(__filename, () => {
...
+      location: createFakeLocation({ query: { page, collection_sort: sort } }),
+      store,
+    });
+
+    fakeDispatch.resetHistory();
+
+    // This simulates the user clicking the "Remove" button on the
+    // EditableCollectionAddon component.
+    root.instance().removeAddon(addonId);
+    sinon.assert.callCount(fakeDispatch, 1);
```
kumar303: I prefer to see `assert.calledWith` come before `assert.callCount` because `calledWith` will give a better error message on a failure.

tests/unit/amo/components/TestCollection.js
kumar303: You should pass this to `renderComponent()` to make the test more robust. In other words, you want to keep it in sync with your assertion.

tests/unit/amo/components/TestCollection.js
```diff
@@ -1268,6 +1268,66 @@ describe(__filename, () => {
...
+      removeAddonFromCollection({
+        addonId,
+        errorHandlerId: errorHandler.id,
+        filters: { page: 1, collectionSort: sort },
+        slug: detail.slug,
+        username: detail.author.username,
+      }),
+    );
+
+    sinon.assert.calledWith(root.instance().props.history.push, {
```
kumar303: After passing `history` to `renderComponent`, I think you could make this `sinon.assert.calledWith(history.push, ...)`

src/amo/components/Collection/index.js
rebmullin: unless I a missing something, the only difference I really see is this piece:

```javascript
  let { page } = filters;

    if (isNewPage) {
      page = getNewCollectionPage(collection);
    }
...
```

As far as pushing the page, I was already doing that - only pushing if isNewPage was true but I agree your names make it a little clearer

willdurand: Exactly! You did the hard work by finding the issue and the patch to fix it. I just made it a bit more readable because I really had troubles understanding it.

tests/unit/amo/components/TestCollection.js
rebmullin: I thought about this but the original case I guess assumes and does this really. Maybe it would be okay to just update that case to include this info so it's clear?

tests/unit/amo/components/TestCollection.js
rebmullin: yeh works, thanks!

src/amo/components/Collection/index.js
rebmullin: and my comments too : )

tests/unit/amo/components/TestCollection.js
willdurand: That should probably be covered to distinguish the happy path (page = 1).

rebmullin:
hey @willdurand, how's it look now?

tests/unit/amo/components/TestCollection.js
```diff
@@ -1234,8 +1234,12 @@ describe(__filename, () => {
     });
     const errorHandler = createStubErrorHandler();
     const fakeDispatch = sinon.spy(store, 'dispatch');
-    const page = 123;
+    const page = 5;
```
willdurand: 😄 I noticed this one too. Do we know why it is evaluated to 5?

tests/unit/amo/components/TestCollection.js
```diff
@@ -1257,7 +1262,61 @@ describe(__filename, () => {
     // This simulates the user clicking the "Remove" button on the
     // EditableCollectionAddon component.
     root.instance().removeAddon(addonId);
+    sinon.assert.calledWith(
+      fakeDispatch,
+      removeAddonFromCollection({
+        addonId,
+        errorHandlerId: errorHandler.id,
+        filters: { page: 5, collectionSort: sort },
```
willdurand: It is probably fine to have `page`, it has been assigned to `5` above.

tests/unit/amo/components/TestCollection.js
rebmullin: yes : )

should i show the math here? basically it should be ~ numberofaddons / pageSize

willdurand: Looking at the helper functions, we don't really create 5*25 add-ons in the collection, hence my surprise. I think it is because `count = 123` in a helper function. Is that the reason?

rebmullin: yeh, the count comes from 123 - from the helper function. Now that we have updated the logic to check if the page != newCollectionPage, it needs to be the correct page number here otherwise we won't get the expected results.

I could either change the count or the page to do this and went with the page...

I am not totally clear of your question maybe, so let me know if that helps

willdurand: Can we fix the `count` value to be more realistic?

rebmullin: Sure

tests/unit/amo/helpers.js
```diff
@@ -347,7 +347,7 @@ export function dispatchAutocompleteResults({
 
 export const createFakeCollectionDetail = ({
   name = 'My Addons',
-  count = 123,
+  count = 30,
```
willdurand: That will likely break other tests. I think we should override `count` in `createFakeCollectionDetail()`, just for this test case.

rebmullin: oh ok sure..

tests/unit/amo/components/TestCollection.js
```diff
@@ -1231,11 +1231,16 @@ describe(__filename, () => {
     const addonId = addons[0].addon.id;
     const detail = createFakeCollectionDetail({
       authorId: authorUserId,
+      count: 30,
```
willdurand: I'd add a comment saying that we pretend there are > 25 addons so that we are on page 2.

rebmullin: what about doing something like this: 

```javascript
 // This will simulate a few items on the 2nd page.
 count: DEFAULT_API_PAGE_SIZE + 2,
```

willdurand: Yes!

rebmullin: lol..

but I just realized this is actually identical to the test I have below now. I I guess I can remove this

rebmullin:
thanks for the review everyone : )

PR #5625, opened 2018-07-17
I5568 static theme display on disco

### LINKED ISSUE #5568: Update logic in disco pane so static themes is displayed similiarly to lightweight themes

once static themes are [added to the disco pane](https://github.com/mozilla/addons-server/issues/8572), we should make sure it looks similar to lightweight themes

rebmullin:
fixes #5568 

note: the functionality updates are coming in another PR

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -315,13 +315,3 @@ export const addQueryParamsToHistory = ({
   return qhistory(history, _stringify, _parse);
 };
 
-export const getPreviewImage = (addon, { version = 0, full = true } = {}) => {
-  const preview = addon.previews.length && addon.previews[version];
-  if (preview) {
-    const previewSize = full ? 'image_url' : 'thumbnail_url';
-    return preview[previewSize] && isAllowedOrigin(preview[previewSize])
-      ? preview[previewSize]
-      : null;
-  }
-  return null;
-};
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -314,12 +314,12 @@ export const addQueryParamsToHistory = ({
 }) => {
   return qhistory(history, _stringify, _parse);
 
-export const getPreviewImage = (
-  previews,
-  { version = 0, full = true } = {},
-) => {
-  const previewsType = full ? 'image_url' : 'thumbnail_url';
-  return previews && previews.length > 0 && previews[version][previewsType]
-    ? previews[version][previewsType]
-    : null;
+
+export const getPreviewImage = (addon, { version = 0, full = true } = {}) => {
+  const preview = addon.previews.length && addon.previews[version];
+  if (preview) {
+    const previewSize = full ? 'image_url' : 'thumbnail_url';
+    return preview[previewSize] || null;
+  }
+  return null;
 };
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -313,7 +313,7 @@ export const addQueryParamsToHistory = ({
   _stringify = stringify,
 }) => {
   return qhistory(history, _stringify, _parse);
-
+};
 
 export const getPreviewImage = (addon, { version = 0, full = true } = {}) => {
   const preview = addon.previews.length && addon.previews[version];
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -313,4 +313,13 @@ export const addQueryParamsToHistory = ({
   _stringify = stringify,
 }) => {
   return qhistory(history, _stringify, _parse);
+
+export const getPreviewImage = (
+  previews,
+  { version = 0, full = true } = {},
+) => {
+  const previewsType = full ? 'image_url' : 'thumbnail_url';
+  return previews && previews.length > 0 && previews[version][previewsType]
+    ? previews[version][previewsType]
+    : null;
 };
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -319,7 +319,9 @@ export const getPreviewImage = (addon, { version = 0, full = true } = {}) => {
   const preview = addon.previews.length && addon.previews[version];
   if (preview) {
     const previewSize = full ? 'image_url' : 'thumbnail_url';
-    return preview[previewSize] || null;
+    return preview[previewSize] && isAllowedOrigin(preview[previewSize])
+      ? preview[previewSize]
+      : null;
   }
   return null;
 };
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -314,4 +314,3 @@ export const addQueryParamsToHistory = ({
 }) => {
   return qhistory(history, _stringify, _parse);
 };
-
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -122,33 +122,36 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const {
-      i18n,
-      name,
-      previewURL,
-      previews,
-      type,
-      getBrowserThemeData,
-    } = this.props;
+    const { type } = this.props;
 
     if (ADDON_TYPE_THEMES.includes(type)) {
-      const imageUrl =
+      const {
+        getBrowserThemeData,
+        i18n,
+        name,
+        previewURL,
+        previews,
+      } = this.props;
+
+      let imageUrl =
         previews && previews.length > 0 && previews[0].image_url
           ? previews[0].image_url
           : null;
 
-      const label = sprintf(i18n.gettext('Preview of %(name)s'), { name });
+      if (!imageUrl && type === ADDON_TYPE_THEME) {
+        imageUrl = previewURL;
+      }
 
       const headerImage = (
         <img
-          alt={label}
+          alt={sprintf(i18n.gettext('Preview of %(name)s'), { name })}
           className="Addon-theme-header-image"
-          src={imageUrl || previewURL}
+          src={imageUrl}
         />
       );
 
       /* eslint-disable jsx-a11y/href-no-hash, jsx-a11y/anchor-is-valid */
-      return !imageUrl && type === ADDON_TYPE_THEME ? (
+      return type === ADDON_TYPE_THEME ? (
         <a
           className="theme-image"
           data-browsertheme={getBrowserThemeData()}
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -31,7 +31,7 @@ import translate from 'core/i18n/translate';
 import { withInstallHelpers } from 'core/installAddon';
 import { getAddonByGUID } from 'core/reducers/addons';
 import tracking, { getAddonTypeForTracking } from 'core/tracking';
-import { getPreviewImage } from 'core/utils';
+import { getPreviewImage, isTheme } from 'core/utils';
 import { sanitizeHTMLWithExternalLinks } from 'disco/utils';
 import { getClientCompatibility as _getClientCompatibility } from 'core/utils/compatibility';
 import LoadingText from 'ui/components/LoadingText';
@@ -123,13 +123,14 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { type } = this.props.addon;
+    const { addon } = this.props;
+    const { type } = addon;
 
-    if (ADDON_TYPE_THEMES.includes(type)) {
-      const { addon, getBrowserThemeData, i18n } = this.props;
-      const { name, previewURL, previews } = addon;
+    if (isTheme(type)) {
+      const { getBrowserThemeData, i18n } = this.props;
+      const { name, previewURL } = addon;
 
-      let imageUrl = getPreviewImage(previews);
+      let imageUrl = getPreviewImage(addon);
 
       if (!imageUrl && type === ADDON_TYPE_THEME) {
         imageUrl = previewURL;
@@ -164,7 +165,7 @@ export class AddonBase extends React.Component {
   getDescription() {
     const { description, type } = this.props;
 
-    if (ADDON_TYPE_THEMES.includes(type)) {
+    if (isTheme(type)) {
       return null;
     }
 
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -31,6 +31,7 @@ import translate from 'core/i18n/translate';
 import { withInstallHelpers } from 'core/installAddon';
 import { getAddonByGUID } from 'core/reducers/addons';
 import tracking, { getAddonTypeForTracking } from 'core/tracking';
+import { getPreviewImage } from 'core/utils';
 import { sanitizeHTMLWithExternalLinks } from 'disco/utils';
 import { getClientCompatibility as _getClientCompatibility } from 'core/utils/compatibility';
 import LoadingText from 'ui/components/LoadingText';
@@ -122,21 +123,13 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { type } = this.props;
+    const { type } = this.props.addon;
 
     if (ADDON_TYPE_THEMES.includes(type)) {
-      const {
-        getBrowserThemeData,
-        i18n,
-        name,
-        previewURL,
-        previews,
-      } = this.props;
+      const { addon, getBrowserThemeData, i18n } = this.props;
+      const { name, previewURL, previews } = addon;
 
-      let imageUrl =
-        previews && previews.length > 0 && previews[0].image_url
-          ? previews[0].image_url
-          : null;
+      let imageUrl = getPreviewImage(previews);
 
       if (!imageUrl && type === ADDON_TYPE_THEME) {
         imageUrl = previewURL;
@@ -161,9 +154,7 @@ export class AddonBase extends React.Component {
           {headerImage}
         </a>
       ) : (
-        <a className="theme-image" href="#">
-          {headerImage}
-        </a>
+        <React.Fragment>{headerImage}</React.Fragment>
       );
     }
 
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -14,7 +14,6 @@ import InstallButton from 'core/components/InstallButton';
 import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
-  ADDON_TYPE_THEMES,
   CLICK_CATEGORY,
   DOWNLOAD_FAILED,
   ERROR,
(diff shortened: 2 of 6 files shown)


src/disco/components/Addon/index.js
```diff
@@ -121,31 +123,58 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { getBrowserThemeData, i18n, name, previewURL } = this.props;
-    if (this.props.type === ADDON_TYPE_THEME) {
+    const { type } = this.props;
```
willdurand: I *think* `type` comes from the `addon` prop. I know it is available directly but that's because we spread the whole `addon` if I remember correctly. I think it would be slightly better to retrieve `type` from the `addon`.

This comment also applies to `name`, `previews` and probably `previewURL` too.

src/disco/components/Addon/index.js
```diff
@@ -121,31 +123,58 @@ export class AddonBase extends React.Component {
...
           onClick={this.installTheme}
         >
-          <img
-            src={previewURL}
-            alt={sprintf(i18n.gettext('Preview of %(name)s'), { name })}
-          />
+          {headerImage}
+        </a>
+      ) : (
+        <a className="theme-image" href="#">
```
willdurand: Do we really need a `a` here? It has no link and no `onClick` handler.

tests/unit/disco/components/TestAddon.js
```diff
@@ -386,22 +388,59 @@ describe(__filename, () => {
...
+    {
+      type: ADDON_TYPE_STATIC_THEME,
+      previews: [
+        {
+          ...fakePreview,
+          image_url: 'https://addons.cdn.mozilla.net/full/54321.png',
+        },
+      ],
+    },
+  ])(`Addon for %o`, async (testConfig) => {
```
willdurand: I like your use of `.each` but I would like to see a better description. Currently it is:

```
  ● /Users/williamdurand/projects/mozilla/addons-frontend/tests/unit/disco/components/TestAddon.js › Addon for { type: 'statictheme',
  previews:
   [ { id: 1,
       caption: 'Image 1',
       image_url: 'https://addons.cdn.mozilla.net/full/54321.png',
       thumbnail_url: 'https://addons.cdn.mozilla.net/thumb/1.png',
       image_size: [ 400, 200, [length]: 2 ],
       thumbnail_size: [ 200, 100, [length]: 2 ] },
     [length]: 1 ] } › renders the correct image for a theme
```

it is not really readable. I think here that would make more sense to have to set of tests because you are not exactly testing the same things it seems (otherwise `previews` would have the same content, no?)

Alternatively, you can use and configure `previews` in the `beforeEach` block:

```
describe.each([ADDON_TYPE_THEME, ADDON_TYPE_STATIC_THEME])(`Addon with type = %s`, (type) => {
});
```

src/disco/components/Addon/index.js
rebmullin: Yeh I guess it makes sense to move this to the functionality PR where it does have the its own onClick

tests/unit/disco/components/TestAddon.js
rebmullin: I saw some of the same repeating tests but I see what you're saying. Let me try this way you suggest with just type.

rebmullin:
> One question: do you also see the static theme being enabled time to time? Not always. I don't really know why (and it does not matter).

hey @willdurand, are you seeing this on dev? static themes in disco is supposed to be enabled here. @eviljeff set it up that way a  a few weeks ago...
I am actually not sure how this is switched off tho and if QA has been toggling this for testing..? If they are it should be more "on" then "off" tho... I've not seen it disabled since I knew it was enabled...

tests/unit/disco/components/TestAddon.js
rebmullin: let me know if how it have it now is ok. I extracted some of it..

rebmullin:
hey @willdurand, thanks for the feedback. Do you mind taking another look at this one?

src/core/utils/index.js
```diff
@@ -304,3 +304,13 @@ export const normalizeFileNameId = (filename) => {
...
+
+export const getPreviewImage = (
+  previews,
+  { version = 0, full = true } = {},
+) => {
+  const previewsType = full ? 'image_url' : 'thumbnail_url';
+  return previews && previews.length > 0 && previews[version][previewsType]
+    ? previews[version][previewsType]
+    : null;
+};
```
rebmullin: I've only used the function here in this related section.. but I'll set up a follow issue/PR to use this in the other places too

src/disco/components/Addon/index.js
```diff
@@ -121,31 +124,48 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { getBrowserThemeData, i18n, name, previewURL } = this.props;
-    if (this.props.type === ADDON_TYPE_THEME) {
+    const { type } = this.props.addon;
+
+    if (ADDON_TYPE_THEMES.includes(type)) {
```
willdurand: Should we use `isTheme()` here?

src/disco/components/Addon/index.js
```diff
@@ -121,31 +124,48 @@ export class AddonBase extends React.Component {
...
+      const { addon, getBrowserThemeData, i18n } = this.props;
```
willdurand: Would it make sense to get the addon above? (You use `this.props.addon`) to retrieve the `type`, but maybe we can get the `addon` first, and then its type.

src/disco/components/Addon/index.js
```diff
@@ -121,31 +124,48 @@ export class AddonBase extends React.Component {
...
           onClick={this.installTheme}
         >
-          <img
-            src={previewURL}
-            alt={sprintf(i18n.gettext('Preview of %(name)s'), { name })}
-          />
+          {headerImage}
         </a>
+      ) : (
+        <React.Fragment>{headerImage}</React.Fragment>
```
willdurand: Do you really need a fragment here?

src/disco/components/Addon/index.js
```diff
@@ -121,31 +124,48 @@ export class AddonBase extends React.Component {
...
     }
+
     return null;
   }
 
   getDescription() {
     const { description, type } = this.props;
 
-    if (type === ADDON_TYPE_THEME) {
+    if (ADDON_TYPE_THEMES.includes(type)) {
```
willdurand: `isTheme()`?

tests/unit/disco/components/TestAddon.js
```diff
@@ -386,25 +388,57 @@ describe(__filename, () => {
     });
   });
 
-  describe('<Addon type="theme"/>', () => {
-    let root;
+  describe.each([ADDON_TYPE_THEME, ADDON_TYPE_STATIC_THEME])(
+    `Addon for %s`,
```
willdurand: > Addon with type = %s

What do you think?

tests/unit/disco/components/TestAddon.js
```diff
@@ -386,25 +388,57 @@ describe(__filename, () => {
...
+  });
 
+  describe('<Addon type="theme"/>', () => {
     it('calls installTheme on click', () => {
       const installTheme = sinon.stub();
-      const addon = result;
+      const addon = {
+        ...result,
+        type: ADDON_TYPE_THEME,
+      };
```
willdurand: What is `result` here? Also, I think we should be using `createInternalAddon()` from `core/reducers/addons`.

tests/unit/disco/components/TestAddon.js
```diff
@@ -386,25 +388,57 @@ describe(__filename, () => {
...
-    it('does render the theme image for a theme', () => {
-      expect(root.find('.theme-image')).toHaveLength(1);
-    });
+      it("doesn't render the logo for a theme", () => {
+        expect(root.find('.logo')).toHaveLength(0);
+      });
 
-    it("doesn't render the logo for a theme", () => {
-      expect(root.find('.logo')).toHaveLength(0);
+      it("doesn't render the description for a theme", () => {
```
willdurand: You might not need "for a theme" every time since it is grouped into a `describe` block that should mention the type of the addon.

src/core/utils/index.js
willdurand: Would it make sense to pass the `addon` object instead?

Can you find a way to reuse this helper function somewhere else?

eviljeff:
>> One question: do you also see the static theme being enabled time to time? Not always. I don't really know why (and it does not matter).

> hey @willdurand, are you seeing this on dev? static themes in disco is supposed to be enabled here. @eviljeff set it up that way a a few weeks ago...
I am actually not sure how this is switched off tho and if QA has been toggling this for testing..? If they are it should be more "on" then "off" tho... I've not seen it disabled since I knew it was enabled...

yeah static themes should be included in the API response on if `disco-staticthemes-dev` waffle switch is enabled.

src/disco/components/Addon/index.js
rebmullin: If I don't have this, it breaks. This is kind of temporary tho anyway, b/c in the functionality PR, we should add back the `a` wrapper tag which has an onClick (a special case for static themes). I realize your work might override that with your work but I am just going off what I know my changes would be for now :)

willdurand: mmh, what exactly is breaking? A fragment renders nothing, and I believe `headerImage` is a valid element.

willdurand:
> hey @willdurand, are you seeing this on dev? static themes in disco is supposed to be enabled here. @eviljeff set it up that way a a few weeks ago...
> I am actually not sure how this is switched off tho and if QA has been toggling this for testing..? If they are it should be more "on" then "off" tho... I've not seen it disabled since I knew it was enabled...

Sorry, I was not clear. I meant, when I click to install a theme, it gets enabled sometimes, but not always. I am not sure why because I thought that was not possible.

tests/unit/disco/components/TestAddon.js
rebmullin: results comes from `loadedAddons` function which is actually used in the DiscoPane component:

`const result = loadedAddons(_state)[0];`

Is that okay?

willdurand: ah ok, yes. sorry I was a bit lost but not your fault :)

src/core/utils/index.js
rebmullin: hmm..I am not sure I am understanding why we should pass in `addon` instead... But let me take another look...

willdurand: I think I gave you a bad hint about this function. Do you think we can reuse it somewhere else? The code looks similar but not exactly the same (there are calls to `isAllowedOrigin()` for instance).

rebmullin:
> Sorry, I was not clear. I meant, when I click to install a theme, it gets enabled sometimes, but not always. I am not sure why because I thought that was not possible.

I think this is part of the bug with static theme. I had to explicitly call `enable()` for static themes. I know you said you removed/refactored the Switch but I had something like this:

https://github.com/mozilla/addons-frontend/pull/5683/files#diff-fd3a9b2c8bff28448f81457a3e53f828R25

That PR is a little bit outdated (needs a rebase) but I believe you should always see it working there

src/disco/components/Addon/index.js
rebmullin: I see this in console:

Caught application error: Invariant Violation: "Objects are not valid as a React child (found: object with keys {headerImage}). If you meant to render a collection of children, use an array instead.

Alt text

src/core/utils/index.js
rebmullin: oh sorry I missed your comment above about isAllowedOrigin and I guess the hint too:) ...

Yeh I guess I missed that piece of it. I am honestly not sure why it's in some places but not others (isAllowedOrigin). I guess I can add it to the getPreviewImage function. Would that work?

rebmullin: So I added this isAllowedOrigin to the function. It just adds an extra step which I thought would be okay.  I *think* now it should cover the other cases too. Let me know what you think.

willdurand:
@rebmullin I have integrated most of your changes in #5712 🎉

rebmullin:
> I have integrated most of your changes in #5712 

that's awesome @willdurand. I'll close this and #5683 out once that goes through 🎉

willdurand:
@rebmullin what's the status of this PR?

rebmullin:
@willdurand, oh? I thought you were kind of taking this over. I can def update this in the am tho if you need it. let me know...

src/core/utils/index.js
```diff
@@ -314,3 +314,14 @@ export const addQueryParamsToHistory = ({
 }) => {
   return qhistory(history, _stringify, _parse);
 };
+
+export const getPreviewImage = (addon, { version = 0, full = true } = {}) => {
```
willdurand: Can we move this function to `src/core/imageUtils.js`?

willdurand: Is it really a `version` or just an index?

src/disco/components/Addon/index.js
```diff
@@ -121,31 +123,49 @@ export class AddonBase extends React.Component {
   }
 
   getThemeImage() {
-    const { getBrowserThemeData, i18n, name, previewURL } = this.props;
-    if (this.props.type === ADDON_TYPE_THEME) {
+    const { addon } = this.props;
+    const { type } = addon;
```
willdurand: You could probably destructure `type` from `this.props.addon`;

src/disco/components/Addon/index.js
```diff
@@ -121,31 +123,49 @@ export class AddonBase extends React.Component {
...
+
+    if (isTheme(type)) {
+      const { getBrowserThemeData, i18n } = this.props;
+      const { name, previewURL } = addon;
+
+      let imageUrl = getPreviewImage(addon);
+
+      if (!imageUrl && type === ADDON_TYPE_THEME) {
+        imageUrl = previewURL;
```
willdurand: Does that mean `imageUrl` can be `null` for a static theme? What should we be displaying then?

src/disco/components/Addon/index.js
willdurand: mmh, this works for me:

```js
      return type === ADDON_TYPE_THEME ? (
        <a
          className="theme-image"
          data-browsertheme={getBrowserThemeData()}
          href="#"
          onClick={this.installTheme}
        >
          {headerImage}
        </a>
      ) : (
        headerImage
      );

```

src/disco/components/Addon/index.js
```diff
@@ -233,7 +253,7 @@ export class AddonBase extends React.Component {
     }
 
     const addonClasses = makeClassName('addon', {
-      theme: type === ADDON_TYPE_THEME,
+      theme: ADDON_TYPE_THEMES.includes(type),
```
willdurand: Why not using `isTheme()` here?

tests/unit/core/utils/test_index.js
```diff
@@ -792,4 +797,81 @@ describe(__filename, () => {
...
+            ...fakePreview,
+            image_url: 'https://addons.cdn.mozilla.net/image.not.used.here.png',
+          },
+        ],
+      });
+      const image = getPreviewImage(addon);
+      expect(image).toEqual(fullImage);
+    });
+
+    it('returns the full image from the 2nd item in the previews array', () => {
```
willdurand: I get it but this is weird to see `2nd` in the description and `version: 1` in the assertion.

tests/unit/core/utils/test_index.js
```diff
@@ -792,4 +797,81 @@ describe(__filename, () => {
       });
     });
   });
+
+  describe('getPreviewImage', () => {
+    it('returns the first full image from the previews array', () => {
+      const fullImage = 'https://addons.cdn.mozilla.net/full/12345.png';
```
willdurand: Can we use `${config.get('amoCDN')}` instead of `https://addons.cdn.mozilla.net` everywhere?

src/disco/components/Addon/index.js
rebmullin: Static themes should always have this. This was only set up for persona. Initially I was thinking after migration (of lwt themes) persona would have the previews too but it actually should change the type too and never hit here after the migration happens. If that makes sense..

tests/unit/core/utils/test_index.js
rebmullin: hmm ok..let me try something else

src/core/utils/index.js
rebmullin: yeh index is better, thanks

rebmullin: yep, I'll move that too..

tests/unit/disco/components/TestAddon.js
```diff
@@ -385,25 +387,58 @@ describe(__filename, () => {
...
+          type: ADDON_TYPE_STATIC_THEME,
+        },
+      });
+      const image = shallowRoot.find('.Addon-theme-header-image');
+      expect(image.prop('src')).toEqual(fullImage);
+      expect(image).toHaveProp('alt', `Preview of ${newAddonName}`);
     });
+  });
 
+  describe('<Addon type="theme"/>', () => {
```
willdurand: Can we rename this description to: `addon with type "theme"`

tests/unit/disco/components/TestAddon.js
```diff
@@ -385,25 +387,58 @@ describe(__filename, () => {
...
 
-    it("doesn't render the logo for a theme", () => {
-      expect(root.find('.logo')).toHaveLength(0);
+      it("doesn't render the description", () => {
+        expect(root.find('.editorial-description')).toHaveLength(0);
+      });
+    },
+  );
+
+  describe('<Addon type="statictheme"/>', () => {
```
willdurand: Can we rename this description to: `addon with type "static theme"`

tests/unit/disco/components/TestAddon.js
```diff
@@ -385,25 +387,58 @@ describe(__filename, () => {
...
 
+  describe('<Addon type="theme"/>', () => {
     it('calls installTheme on click', () => {
       const installTheme = sinon.stub();
-      const addon = result;
+      const addon = {
+        ...result,
+        type: ADDON_TYPE_THEME,
+        previews: [],
+      };
```
willdurand: Why do we need this change?

tests/unit/disco/components/TestAddon.js
```diff
@@ -385,25 +387,58 @@ describe(__filename, () => {
...
+            {
+              ...fakePreview,
+              image_url: fullImage,
+            },
+          ],
+          type: ADDON_TYPE_STATIC_THEME,
+        },
+      });
+      const image = shallowRoot.find('.Addon-theme-header-image');
+      expect(image.prop('src')).toEqual(fullImage);
```
willdurand: You can use `toHaveProp` too here.

tests/unit/disco/components/TestAddon.js
rebmullin: previews has to be empty and type persona. The getPreviewImage checks inside the add for this info and I believe the default is type extension with previews. SO i am just updating it to match the results for lightweight theme

PR #5630, opened 2018-07-17
Update eslint-plugin-amo

willdurand:
I have updated the `eslint-plugin-amo` with more rules, see: https://github.com/willdurand/eslint-plugin-amo#rules.


willdurand:
ah well, node 6 😄 – going to release a fix for that.

willdurand:
> I don't understand your comment at #5630 (comment), but regardless this works locally for me. r+

@bobsilverberg the plugin had a small issue on node 6 and I found it thanks to our test suite here. So I released a new version of the plugin to make it compatible with node 6 again.

Issue #5635, opened 2018-07-17
Disable the "Save collection" button unless one of the fields in the edit form has been changed


bobsilverberg:
As discussed in last week's front-end UX meeting, the following changes should be made to the collection edit screen:

1. Move the select which is used to add an add-on to a collection from the edit form into the card below (the one that currently holds the sort select).
2. Disable the "Save collection" button unless one of the fields in the edit form has been changed.

bobsilverberg:
I've decided to split this into two issues. The first requirement above was moved into issue #5675, so I am going to repurpose this issue to be solely about the second requirement, which is to disable the "Save collection" button unless one of the fields in the edit form has been changed.

AlexandraMoga:
Verified on AMO -dev with FF61, WIn10x64 and Android 8.0

The Save button for the edit collection form is disabled until one of the fields - name, description, url - are edited.

![save button](https://user-images.githubusercontent.com/31961530/43131744-6e28452c-8f42-11e8-94da-4b9753ea4a0d.gif)


I have filed a follow up issue #5702  for my concern regarding the Save button state after an add-on is added/removed from the collection

PR #5636, opened 2018-07-17
Allow the CollectionManager to display CollectionDetails instead of an edit form

bobsilverberg:
Ref #5508 
Depends on #5619 

Note that this is built on top of the patch from #5619, and therefore only the second commit is relevant to this patch.

This displays the collection details instead of an edit form when a user does not have full edit permission for a collection. Note that this is not a full solution for #5508, as we also need to move the select box which is used to add an add-on to a collection into the card with the sort select in order to give a user access to add an add-on to a collection in this scenario. That latter part will be addressed by issue #5635. I didn't include it in this patch to try to keep the size small.


bobsilverberg:
I think it actually makes sense to implement #5635 before this, so I'm going to mark this as blocked for now.

bobsilverberg:
This is now ready to be reviewed. The UI works, but there is one outstanding issue which has been reported at mozilla/addons-server#8747. It is still not possible for a user with one of the above permissions to add an add-on to a collection when it should be allowed. A user can remove add-ons from a collection and add/change/remove notes for add-ons, but cannot add an add-on.

src/amo/components/CollectionManager/index.js
```diff
@@ -211,12 +213,24 @@ export class CollectionManagerBase extends React.Component<
...
       i18n,
       isCollectionBeingModified,
+      showEditForm,
       siteLang,
     } = this.props;
     const { description, name, slug } = this.state;
 
+    if (!showEditForm) {
+      return (
+        <CollectionDetails
```
kumar303: Handling this logic here puts an undue burden on `CollectionManager`. The logic burden is already on `Collection` so why not continue handling it there? Unless I'm missing something, keeping it in `Collection` is as simple as:

```diff
diff --git a/src/amo/components/Collection/index.js b/src/amo/components/Collection/index.js
index ab3bb3b14..067d5e0b3 100644
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -318,7 +318,8 @@ export class CollectionBase extends React.Component<InternalProps> {
       showEditButton,
     } = this.props;
 
-    if (creating || editing) {
+    const managingCollection = creating || editing;
+    if (managingCollection && hasEditPermission) {
       return (
         <CollectionManager
           collection={collection}
@@ -333,7 +334,7 @@ export class CollectionBase extends React.Component<InternalProps> {
       <CollectionDetails
         collection={collection}
         filters={filters}
-        showEditButton={showEditButton}
+        showEditButton={showEditButton && !managingCollection}
       />
     );
   }
```

src/amo/components/CollectionManager/index.js
```diff
@@ -211,12 +213,24 @@ export class CollectionManagerBase extends React.Component<
...
+          collection={collection}
+          filters={filters}
+          showEditButton={false}
```
kumar303: Once you enter edit mode as a maintainer, there is no way to get out of edit mode (besides pressing the browser *back* button). I suggest showing a button like *Finish editing* which will behave the same as *Cancel* on the edit form.

PR #5639, opened 2018-07-17
Create a CollectionSort component

### LINKED ISSUE #5638: Create a CollectionSort component to reduce the code in the Collection component
As we're going to relocate the sort select in #5635, it seems like a good time to extract the code for the sort select into its own component, and that's what this issue is about.

bobsilverberg:
Fixes #5638 
Depends on #5619 

This only depends on #5619 because I didn't want to have to go back to fix conflicts which I know will exist, so it doesn't really depend on that, but I built it on top of it.

### DIFF
--- a/src/amo/components/CollectionSort/index.js
+++ b/src/amo/components/CollectionSort/index.js
@@ -1,0 +1,87 @@
+/* @flow */
+import * as React from 'react';
+import { compose } from 'redux';
+
+import {
+  COLLECTION_SORT_DATE_ADDED_ASCENDING,
+  COLLECTION_SORT_DATE_ADDED_DESCENDING,
+  COLLECTION_SORT_NAME,
+  COLLECTION_SORT_POPULARITY,
+} from 'core/constants';
+import translate from 'core/i18n/translate';
+import Card from 'ui/components/Card';
+import Select from 'ui/components/Select';
+import type { CollectionFilters } from 'amo/reducers/collections';
+import type { I18nType } from 'core/types/i18n';
+
+import './styles.scss';
+
+export type Props = {|
+  filters: CollectionFilters,
+  onSortSelect: (event: SyntheticEvent<HTMLSelectElement>) => void,
+|};
+
+type InternalProps = {|
+  ...Props,
+  i18n: I18nType,
+|};
+
+export class CollectionSortBase extends React.Component<InternalProps> {
+  sortOptions() {
+    const { i18n } = this.props;
+
+    return [
+      {
+        label: i18n.gettext('Newest first'),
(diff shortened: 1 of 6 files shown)


src/amo/components/CollectionSort/index.js
```diff
@@ -0,0 +1,87 @@
...
+        <form>
+          <label className="CollectionSort-label" htmlFor="Sort-Select">
+            {i18n.gettext('Sort add-ons by')}
+          </label>
+          <Select
+            className="CollectionSort-select"
+            defaultValue={filters.collectionSort}
+            id="CollectionSort-select"
+            name="sort"
+            onChange={onSortSelect}
```
kumar303: I'd suggest keeping `onSortSelect` defined within this component. The logic belongs here. The only case where it wouldn't belong here is if two components wanted to do different things with the sort select but I can't imagine how they would. 

You can use [`withRouter`](https://github.com/mozilla/addons-frontend/blob/9131e88b0f6a07e98ba2fd1918bef73b013de6f9/src/amo/components/Addon/index.js#L648) to get access to the router and use a standard `mapStateToProps` to get access to `api` state. Also, a shared `editCollectionUrl()` will come in handy.

tests/unit/amo/components/TestCollectionSort.js
```diff
@@ -0,0 +1,65 @@
...
+    );
+  };
+
+  it('renders a sort select', () => {
+    const sort = COLLECTION_SORT_NAME;
+    const onSortSelect = sinon.stub();
+
+    const root = render({ filters: { collectionSort: sort }, onSortSelect });
+
+    expect(root.find('.CollectionSort-label')).toHaveText('Sort add-ons by');
```
kumar303: I don't think it's useful to test statically configured data like this. It's also a maintenance burden because when you change the text you have to update the test.

bobsilverberg:
FYI, I rebased this to make the review simpler.

src/amo/components/CollectionSort/index.js
```diff
@@ -0,0 +1,145 @@
...
+              return (
+                <option key={option.value} value={option.value}>
+                  {option.label}
+                </option>
+              );
+            })}
+          </Select>
+        </form>
+      </Card>
+    ) : null;
```
kumar303: I think it would be better to render this form even without a collection (i.e. when loading). It should render fine -- it would just throw an error if you clicked on a sort option before the collection loaded (which would be hard to do in real life). The old component worked like this.

It's a good habit to try and render as much as possible while data is loading. This makes the page less jumpy when the final content loads.

PR #5642, opened 2018-07-17
I5640 adding extract id to confirm btn

rebmullin:


src/ui/components/ConfirmButton/index.js
```diff
@@ -116,11 +117,18 @@ export class ConfirmButtonBase extends React.Component<InternalProps> {
   }
 }
 
+export const extractId = (ownProps: Props) => {
+  const buttonId =
+    ownProps.id ||
+    `${ownProps.className || ''}-${ownProps.message.replace(/\s/g, '')}`;
```
rebmullin: I thought it was better to remove the spaces from the message. Is this ok?

kumar303: It might be helpful to keep the ID readable so I'd suggest replacing spaces with underscores: `ownProps.message.replace(/\s/g, '_')`

src/ui/components/ConfirmButton/index.js
```diff
@@ -116,11 +117,18 @@ export class ConfirmButtonBase extends React.Component<InternalProps> {
...
+  return `confirmButton-${buttonId}`;
```
kumar303: prefixing with `confirmButton` is redundant since it will already be prefixed with the filename

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -162,4 +165,19 @@ describe(__filename, () => {
     expect(root).not.toHaveClassName('ConfirmButton--show-confirmation');
     sinon.assert.calledWith(onConfirmSpy, event);
   });
+
+  describe('extractId', () => {
+    it("contains its ID if it's passed in", () => {
+      const id = 'special-button';
```
kumar303: ✨

src/ui/components/ConfirmButton/index.js
willdurand: Honestly, I'd rather make `id` required than playing with props like this. It will be unreadable in the logs/devtools 🙁

kumar303: Okay, I'm fine with making `id` required. It will force the caller to think about uniqueness which is probably a good thing.

rebmullin: okay, sure thing. I can make id required..

tests/unit/ui/components/TestConfirmButton.js
rebmullin: :)

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -162,4 +166,11 @@ describe(__filename, () => {
     expect(root).not.toHaveClassName('ConfirmButton--show-confirmation');
     sinon.assert.calledWith(onConfirmSpy, event);
   });
+
+  describe('extractId', () => {
+    it('returns a unique ID provided by the ID prop', () => {
+      const id = 'special-button';
+      expect(extractId({ id })).toEqual(id);
```
kumar303: This needs to pass all component props like `extractId(getProps({ id: 'special-button' }))`.

It may not seem important now, but some refactoring later could require new props which might inadvertently break the test. Besides, you already have an abstraction to generate default props. It just needs a tweak like:

```diff
diff --git a/tests/unit/ui/components/TestConfirmButton.js b/tests/unit/ui/components/TestConfirmButton.js
index 81cdd4b62..48c84655a 100644
--- a/tests/unit/ui/components/TestConfirmButton.js
+++ b/tests/unit/ui/components/TestConfirmButton.js
@@ -14,16 +14,20 @@ import {
 import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
 
 describe(__filename, () => {
-  const render = ({ children, i18n = fakeI18n(), ...props } = {}) => {
+  const getProps = ({ i18n = fakeI18n(), ...props } = {}) => {
+    return {
+      i18n,
+      id: 'Collection-confirm-delete',
+      message: 'some warning message',
+      onConfirm: sinon.stub(),
+      store: dispatchClientMetadata().store,
+      ...props,
+    };
+  };
+
+  const render = ({ children , ...otherProps } = {}) => {
     return shallowUntilTarget(
-      <ConfirmButton
-        i18n={i18n}
-        id="Collection-confirm-delete"
-        message="some warning message"
-        onConfirm={sinon.stub()}
-        store={dispatchClientMetadata().store}
-        {...props}
-      >
+      <ConfirmButton {...getProps(otherProps)}>
         {children || 'the default text of this button'}
       </ConfirmButton>,
       ConfirmButtonBase,
```

src/amo/components/Collection/index.js
```diff
@@ -426,6 +426,7 @@ export class CollectionBase extends React.Component<InternalProps> {
       <ConfirmButton
         buttonType="cancel"
         className="Collection-delete-button"
+        id="Collection-delete-button"
```
kumar303: I suppose it would make sense to use a shared variable here like `id={className}`

src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -99,6 +99,7 @@ export class UserProfileEditPictureBase extends React.Component<InternalProps> {
             <ConfirmButton
               buttonType="cancel"
               className="UserProfileEditPicture-delete-button"
+              id="UserProfileEditPicture-delete-button"
```
kumar303: Same here.

src/ui/components/ConfirmButton/index.js
```diff
@@ -18,6 +18,7 @@ type Props = {|
   className?: string,
   confirmButtonText?: string | null,
   confirmButtonType?: string,
+  id: string,
```
kumar303: Because we don't have 100% Flow coverage, this also needs to be protected in the `render()` function like:

```js
render() {
  invariant(this.props.id, 'The id property is required');
}
```

rebmullin: Actually, a few follow up ?s here: So should I do the same with the message prop (and other required props)? The message text is actually used here in the render vs the id which is only used in the extract function).

rebmullin:
hi @kumar303, I've updated but I had a question above. When you can, please take another look?

src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -67,6 +67,8 @@ export class UserProfileEditPictureBase extends React.Component<InternalProps> {
       },
     );
 
+    const ConfirmButtonClassName = 'UserProfileEditPicture-delete-button';
```
kumar303: nitpick: variable names should always be camel case starting with a lower case letter unless they refer to a class.

src/ui/components/ConfirmButton/index.js
```diff
@@ -67,10 +69,13 @@ export class ConfirmButtonBase extends React.Component<InternalProps> {
       confirmButtonText,
       confirmButtonType,
       i18n,
+      id,
       message,
       uiState,
     } = this.props;
 
+    invariant(id, 'The id property is required');
```
kumar303: In response to your question about adding `invariant()` for `message` and `onConfirm` (the other required props), yes, I think it would be nice to fix that while you're in here.

rebmullin: yep, just wanted to be sure, thanks : )

PR #5652, opened 2018-07-18
Upgrade React Router and HMR 🚀

### LINKED ISSUE #99: Look to remove babel-plugin-react-transform.
This will allow HMR (hot module reloading) to work on functional components by using the HMR API directly. 

See https://github.com/gaearon/react-transform-boilerplate#you-may-not-need-it for more info.

### LINKED ISSUE #2161: Update to react-router-4

willdurand:
Fix #2161 
Fix #99 

---

This PR upgrades React Router to v4 (lib fully rewritten) and configures Hot Module Replacement with [React Hot Loader](https://github.com/gaearon/react-hot-loader). This work will ease the upgrade to Webpack 4 and should help a lot to hot reload redux code (reducers, sagas).

TODO:

- [x] make sure everything work as expected
  - [x] `disco`
  - [x] `amo`
- [x] fix the test suite
- [x] review HMR setup for React
- [x] add tests for new `ScrollToTop` component
- [x] add tests for new `addQueryParamsToHistory()` function
- [x] add tests for new added function in `server/base.js`
- [x] add tests for new `Root` component

Redux/Saga HMR will be handled separately.

### source files

- Route parameters are now available in `match.params` (instead of `params` before):

```js
<Route path="/foo/:bar" component={Foo} />

// Before:
const Foo = ({ params }) => <p>{params.bar}</p>;

// After:
const Foo = ({ match }) => <p>{match.params.bar}</p>;
```

- Do not access `location` via `history.location` (even if it is possible, it is not recommended). Instead, accept `location` as prop:

```js
const Foo = ({ location }) => <div />;

export withRouter(Foo);
```

- `withRouter` HOC (or `Route`) provides the following props: `match`, `location` and `history`.

- New Flow types in `core/types/router`: `ReactRouterLocationType`, `ReactRouterHistoryType` and `ReactRouterMatchType`. The idea is to use the latter as follows:

```js
type InternalProps = {|
  match: {|
    ...ReactRouterMatchType,
    params: {|
      username: string,
    |}
  |},
|};
```

- Query parameters are still available into the `location` object thanks to a new `addQueryParamsToHistory()` helper. React Router v4 does not automatically parse the query string and the `location.query` object was removed. I reintroduced it by decorating the `history` object. TL;DR: `location.query` is available in our app because we made it available.

### test files

- Use `createContextWithFakeRouter()` (`tests/unit/helpers.js`) when you test a component that uses the `withRouter` HOC:

```js
shallowUntilTarget(<Component {...props} />, ComponentBase, {
  shallowOptions: createContextWithFakeRouter(),
});
```

- `fakeRouterLocation()` helper is deprecated, use `createFakeLocation()` instead.

- `createFakeRouter()` has been renamed to `createFakeHistory()` because there is no `router` object anymore, it is called `history` now.


willdurand:
rebased!

bobsilverberg:
> Route parameters are now available in match.params (instead of params before)

Do we think there's any value in doing something ourselves so that we can still access params via `params` instead of `match.params`? It would require a lot fewer changes to our code, but I guess we're probably better off moving to the newer API.

src/core/components/Root/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+import I18nProvider from 'core/i18n/Provider';
+import type { I18nType } from 'core/types/i18n';
+import type { ReactRouterHistoryType } from 'core/types/router';
+
+type Props = {|
+  children: React.Node,
+  history: ReactRouterHistoryType,
+  i18n: I18nType,
+  store: Object,
```
kumar303: We have a rudimentary [store type](https://github.com/mozilla/addons-frontend/blob/d7e6609988c0f2a079e7e88d36b086d7c56c79d2/src/core/types/redux.js#L9) you can use.

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -135,6 +135,9 @@ describe(__filename, () => {
     return shallowUntilTarget(
       <AddonsByAuthorsCard {...props} />,
       AddonsByAuthorsCardBase,
+      {
+        shallowOptions: createContextWithFakeRouter({ location }),
```
kumar303: This seems pretty common. Could we add a new option to `shallowUntilTarget` like `{ withFakeRouter: true }` or something?

tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -83,12 +83,13 @@ describe(__filename, () => {
       store = dispatchSignInActions({ userProps }).store;
     }
 
-    fakeRouter = createFakeRouter({ params });
+    fakeHistory = createFakeHistory();
```
bobsilverberg: Nit: This is inconsistent with the way we usually set this sort of thing up. We usually either create the `fakeHistory` instance in the test and pass it into `render` or we use `beforeEach` to set it up. This way actually seems fine, but it's different from all our other tests, which makes it slightly confusing to a reviewer.

tests/unit/core/TestScrollToTop.js
```diff
@@ -0,0 +1,58 @@
...
+
+    root.setProps({ location: createFakeLocation() });
+    sinon.assert.calledWith(_window.scrollTo, 0, 0);
+  });
+
+  it('does not call window.scrollTo() when window is not defined', () => {
+    const _window = null;
+
+    const root = render({ _window });
+    root.setProps({ location: createFakeLocation() });
```
bobsilverberg: There isn't an assertion here that proves that `window.scrollTo()` wasn't called. I realize it's hard/impossible to do that, considering that `_window` is `null`, but the test description suggests that it is verifying that `window.scrollTo()` is not called. Do we need this test at all? Or maybe the test should be about making sure an error is not thrown when window is not defined and the location changes?

tests/unit/core/utils/test_index.js
```diff
@@ -761,4 +764,32 @@ describe(__filename, () => {
...
+    it('parses the query string to build the query object', () => {
+      const history = createFakeHistory({
+        location: createFakeLocation({ query: null, search: 'foo=123' }),
+      });
+
+      expect(history).toHaveProperty('location.query', null);
+
+      const historyWithQueryParams = addQueryParamsToHistory({ history });
+
+      expect(historyWithQueryParams.location.query).toEqual({
```
bobsilverberg: In the test above you use `expect(historyWithQueryParams).toHaveProperty('location.query', ...)` can you not use that here as well?

src/amo/client.js
```diff
@@ -3,8 +3,12 @@ import makeClient from 'core/client/base';
...
 
-import routes from './routes';
+import App from './components/App';
 import sagas from './sagas';
 import createStore from './store';
 
-makeClient(routes, createStore, { sagas });
+makeClient(App, createStore, { sagas });
+
+if (module.hot) {
```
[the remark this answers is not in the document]
kumar303: It's a [magic NodeJS global](https://nodejs.org/api/globals.html#globals_global_objects).

src/amo/components/AddonReviewList/index.js
```diff
@@ -95,9 +103,14 @@ export class AddonReviewListBase extends React.Component<Props> {
...
         fetchReviews({
-          addonSlug: params.addonSlug,
+          addonSlug: match.params.addonSlug,
           errorHandlerId: errorHandler.id,
-          page: location.query.page,
+          // TODO: so, there is a test case (`it dispatches fetchReviews with
+          // an invalid page variable`) that conflicts with `fetchReviews()`
+          // requiring a page of type `number`. We should decide whether `page`
+          // can be anything OR restrict to integer values.
+          // $FLOW_FIXME
```
kumar303: I think you could just change this action to accept a string `page`. The [API page param](https://github.com/mozilla/addons-frontend/blob/master/src/amo/api/reviews.js#L143) may need to change to a string as well but that's fine.

tests/unit/core/TestScrollToTop.js
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
```
kumar303: I'd expect `TestScrollToTop` to live at ` tests/unit/core/components/TestScrollToTop.js`

tests/unit/core/TestScrollToTop.js
kumar303: For cases like this I'd typically put a comment at the top of the call, like:

```js
// Make sure this doesn't throw.
root.setProps({ location: createFakeLocation() });
```

src/amo/components/CollectionManager/index.js
```diff
@@ -143,7 +148,8 @@ export class CollectionManagerBase extends React.Component<
         createCollection({
           ...payload,
           defaultLocale: siteLang,
-          includeAddonId: router.location.query.include_addon_id,
+          // $FLOW_FIXME: query parameter values are string, not number.
```
kumar303: Same thing here: this code should have always treated the query string param as a string. Sigh, javascript. Could you file a bug to fix this?

tests/unit/amo/components/TestAddonsByAuthorsCard.js
willdurand: Not sure because we often want to pass something. Would it add some value?

src/amo/components/Addon/index.js
```diff
@@ -643,11 +641,10 @@ export function mapStateToProps(state, ownProps) {
 }
 
 export const extractId = (ownProps) => {
-  return ownProps.params.slug;
+  return ownProps.match.params.slug;
 };
 
 export default compose(
-  withRouter,
```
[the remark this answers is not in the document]
willdurand: I want to introduce a `pages` directory for these components.

src/amo/components/SearchTools/index.js
```diff
@@ -10,20 +10,17 @@ import type { SearchFilters } from 'amo/components/AutoSearchInput';
...
 |};
 
 export class SearchToolsBase extends React.Component<Props> {
   render() {
-    const { filters, pathname, ...otherProps } = this.props;
+    const { filters } = this.props;
 
     return (
       <Search
-        {...otherProps}
```
kumar303: What was it relying on this spread for?

src/amo/routes.js
```diff
@@ -1,85 +0,0 @@
...
-import ReviewGuide from 'amo/components/StaticPages/ReviewGuide';
-import SearchPage from 'amo/components/SearchPage';
-import ServerError from 'amo/components/ErrorPage/ServerError';
-import UserProfile from 'amo/components/UserProfile';
-import UserProfileEdit from 'amo/components/UserProfileEdit';
-import SimulateAsyncError from 'core/components/error-simulation/SimulateAsyncError';
-import SimulateClientError from 'core/components/error-simulation/SimulateClientError';
-import SimulateSyncError from 'core/components/error-simulation/SimulateSyncError';
-
-// If you add a new route here, check that the nginx rules maintained by ops
```
kumar303: Could you put this comment back? It's still important.

src/core/server/base.js
```diff
@@ -262,127 +265,105 @@ function baseServer(
...
+        log.warn('Server side rendering is disabled.');
 
-            return hydrateOnClient({ res, pageProps });
-          }
-        } catch (preLoadError) {
-          log.info(oneLine`Caught an error in match() before rendering:
+        return hydrateOnClient({ res, pageProps });
+      }
+    } catch (preLoadError) {
+      log.info(oneLine`Caught an error in match() before rendering:
```
kumar303: It might help avoid confusion if you update this message. It no longer runs in `match()`.

src/disco/components/App/index.js
```diff
@@ -36,21 +47,57 @@ export class AppBase extends React.Component {
...
-export function mapStateToProps(state, ownProps) {
+function mapStateToProps(state, ownProps) {
+  const match = matchPath(ownProps.location.pathname, {
+    path: DISCO_PANE_PATH,
+    exact: true,
+  });
+
   return {
-    browserVersion: ownProps.params.version,
+    browserVersion: match.params.version,
```
kumar303: Can't this be `ownProps.match.params.version`?

tests/unit/helpers.js
```diff
@@ -509,20 +510,43 @@ export const fakeRouterLocation = (props = {}) => {
   };
 };
 
+// TODO: remove this alias and use `createFakeLocation()` only.
```
kumar303: It only got renamed, right? Another option is to keep the name `fakeRouterLocation` and file an issue to rename everything to `createFakeLocation` later.

src/amo/components/UserProfileEdit/index.js
```diff
@@ -53,10 +55,11 @@ type Props = {|
...
   lang: string,
-  // The routing `params` prop is used in `mapStateToProps()`.
-  // eslint-disable-next-line react/no-unused-prop-types
-  params: {| username: string |},
-  router: ReactRouterType,
+  match: {|
+    ...ReactRouterMatchType,
+    params: {| username: string |},
+  |},
+  history: ReactRouterHistoryType,
```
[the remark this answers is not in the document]
willdurand: Thanks, I'll try to apply these changes everywhere.

I filed https://github.com/mozilla/eslint-plugin-amo/issues/8 because I am a bit tired of these comments in general. Prettier does not fix this unfortunately, but hopefully we'll have a ESLint rule soon 🙏

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
willdurand: I filed https://github.com/mozilla/addons-frontend/issues/5735

src/amo/components/AddonReviewList/index.js
```diff
@@ -313,10 +326,17 @@ export class AddonReviewListBase extends React.Component<Props> {
 }
 
 export function mapStateToProps(state: AppState, ownProps: Props) {
-  if (!ownProps || !ownProps.params || !ownProps.params.addonSlug) {
-    throw new Error('The component had a falsey params.addonSlug parameter');
+  if (
+    !ownProps ||
+    !ownProps.match ||
```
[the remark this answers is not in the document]
willdurand: I don't know. I think it should not be possible not to have `ownProps` or `ownProps.match` or `ownProps.match.params`, but that was the legacy code..

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -383,10 +383,10 @@ export const mapStateToProps = (state: AppState, ownProps: Props) => {
 };
 
 const AddonsByAuthorsCard: React.ComponentType<Props> = compose(
+  withRouter,
```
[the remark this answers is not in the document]
willdurand: Yep, I filed https://github.com/mozilla/eslint-plugin-amo/issues/9

src/amo/components/App/index.js
```diff
@@ -214,33 +236,173 @@ export class AppBase extends React.Component<Props> {
...
+            />
+
+            <div className="App-content">
+              <div className="App-content-wrapper">
+                <ErrorPage getErrorComponent={getErrorComponent}>
+                  <div className="App-banner">
+                    <SurveyNotice location={location} />
+                  </div>
+
+                  <Switch>
```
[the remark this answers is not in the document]
willdurand: I moved this code to a new `Routes` component.

src/amo/components/SearchForm/index.js
```diff
@@ -55,7 +61,7 @@ export class SearchFormBase extends React.Component<Props> {
       >
         <AutoSearchInput
           inputName="q"
-          location={router.location}
+          location={history.location}
```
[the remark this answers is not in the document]
willdurand: 😅

src/ui/components/UserCollection/index.js
```diff
@@ -56,7 +58,7 @@ export const UserCollectionBase = (props: InternalProps) => {
 
   return (
     <li className="UserCollection" key={id}>
-      <Link className="UserCollection-link" to={linkTo}>
```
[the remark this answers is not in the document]
willdurand: Nope, that was a cause of a lot of warnings.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1203,17 +1213,6 @@ describe(__filename, () => {
       const link = allLinks[0];
       const path = link.props.to;
       expect(path).toEqual('/addon/chill-out/reviews/');
-
-      return new Promise((resolve, reject) => {
-        match({ location: path, routes }, (error, redirectLocation, props) => {
-          if (error) {
-            return reject(error);
-          }
-          // Check to make sure it is a valid routed path.
```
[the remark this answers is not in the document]
willdurand: I don't think so. I don't really understand why we would need to, suddenly, test routing-related things (and also, I don't really know how to port this old code)

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -105,21 +102,21 @@ describe(__filename, () => {
 
   describe('<AddonReviewList/>', () => {
     it('requires location params', () => {
-      expect(() => render({ router: { params: null } })).toThrowError(
-        /component had a falsey params\.addonSlug/,
+      expect(() => render({ match: null })).toThrowError(
+        /component had a falsey match\.params\.addonSlug/,
```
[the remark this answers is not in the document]
willdurand: Yep, removed this code actually. You're 💯% right.

src/amo/components/AddonReviewList/index.js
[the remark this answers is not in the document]
willdurand: Removed this code after having seen your comment on the test cases.

tests/unit/amo/components/TestApp.js
```diff
@@ -165,20 +158,36 @@ describe(__filename, () => {
...
+    });
     const _navigator = { userAgent: 'Firefox 10000000.0' };
-    render({ _navigator, setUserAgent, userAgent: '' });
 
-    sinon.assert.calledWith(setUserAgent, _navigator.userAgent);
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+
+    render({ _navigator, store });
+
+    sinon.assert.calledWith(
```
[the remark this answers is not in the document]
willdurand: If I remember correctly, all test cases were failing and I had to rewrite most of them, so I took the time to do things better.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -55,26 +56,28 @@ describe(__filename, () => {
...
   }) => {
     return {
       collection,
       creating: false,
       filters: {},
+      history,
       i18n: fakeI18n(),
-      location: fakeRouterLocation(),
-      router,
+      location: createFakeLocation(),
```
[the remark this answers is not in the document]
willdurand: Because that would be too much changes. I realized that the diff would be huge. I filed https://github.com/mozilla/addons-frontend/issues/5736.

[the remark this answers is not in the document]
willdurand: I reverted this change btw.

tests/unit/amo/components/TestUserProfileEdit.js
willdurand: heh, you're right. That's not nice! I have changed the code to initialize a default `history` object in the `renderUserProfileEdit()` function and pass a `history` in the test cases where it is needed.

tests/unit/core/utils/test_index.js
willdurand: good catch, I can do it!

src/amo/components/App/index.js
```diff
@@ -214,33 +236,173 @@ export class AppBase extends React.Component<Props> {
...
-          <div className="App-content">
-            <div className="App-content-wrapper">
-              <ErrorPage getErrorComponent={getErrorComponent}>
-                <div className="App-banner">
-                  <SurveyNotice location={location} />
-                </div>
-                {children}
-              </ErrorPage>
+        <ScrollToTop>
+          <div className="amo">
```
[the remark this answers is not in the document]
willdurand: I believe I have not changed that, it was there for years 😸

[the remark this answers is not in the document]
willdurand: Updated!

src/amo/components/SearchTools/index.js
willdurand: I don't know but I don't see any useful undocumented props that we could pass to the `Search` component...

src/amo/routes.js
willdurand: ah right, I forgot.

src/core/server/base.js
willdurand: ah oops.

src/disco/components/App/index.js
willdurand: No because it is outside a `Route`, that is why I have to use the `matchPath()` function.

tests/unit/helpers.js
willdurand: Yes, I filed https://github.com/mozilla/addons-frontend/issues/5736

willdurand:
> Router navigation no longer works after a hot module reload. I can reproduce it in multiple places. For example, load the home page. Open App/index.js and give it an <h1> in render() or something. Wait for the bundle to update. Try to click any internal link like going to an add-on detail page. The URL changes but the UI does not update.

That sounds pretty bad.

willdurand:
@kumar303 so, there is something I wish I would not have to do right now (in this PR) but that was needed to fix the routing issue.

It makes HMR faster/better, but the `makeClient()` function is gone, and we have a `createClient()` instead that initializes almost everything only once, and returns a `renderApp()` function used to render the app once + every time HMR is performed. In the near future, we'll also return the `store` to configure HMR for redux (I think). I have a working redux HMR setup locally BTW 😎

willdurand:
@kumar303 here is the commit for the HMR with redux: https://github.com/mozilla/addons-frontend/commit/b5f49aae9eeb05622ab5923e12e84eb42791f462

tests/unit/amo/components/TestAddonsByAuthorsCard.js
kumar303: oh, I thought it was not important to reference `location` in most cases. I guess that wouldn't be helpful then.

src/amo/client.js
```diff
@@ -1,10 +1,20 @@
...
 
-makeClient(routes, createStore, { sagas });
+createClient(createStore, { sagas }).then(({ renderApp }) => {
+  renderApp(App);
+
+  if (module.hot) {
+    module.hot.accept('./components/App', () => {
+      // eslint-disable-next-line global-require
+      const NextApp = require('./components/App').default;
+      renderApp(NextApp);
```
kumar303: I'm not sure if this is the culprit but when a module reloads, it unmounts and remounts *everything*. It didn't do that before. Is there a way to fix that? If not, I guess it's not a big deal.

Here is an example. Go to the collection list. Click on a collection. Click *Delete this collection* so that it asks you the prompt -- are you sure? Edit some other component, like `Collection`. After the reload, you can see that `ConfirmButton` got remounted which caused its state to get reset. This is not helpful for hot module debugging. This doesn't happen on the master branch.

willdurand: That's the recommended way now, but I've seen people configureing redux HMR for that reason. I'll try and see if my redux patch fixes that issue.

willdurand: The `ConfirmButton` might not be the best example because the `withUIState()` HOC explicitly resets the state on mount. I guess that's why it gets closed on code change.

kumar303: Resetting state on mount always felt wrong to me in `withUIState()` but I didn't have concrete reasons for why. This is a concrete reason. I filed https://github.com/mozilla/addons-frontend/issues/5764 to fix it.

PR #5666, opened 2018-07-19
Use the router to navigate to the "create collection" page in AddAddonToCollection

### LINKED ISSUE #5664: User is directed to the old form when selecting Create a collection from the add-on details page
STR:

1. Log in to AMO stage
2. Select an add-on to view its details page
3. Scroll down to the Add to collection feature and select Create a new collection

Actual result:
User lands on the old create a collection form

Expected result:
User lands on the new create a collection form

Notes:
 - reproduced on AMO dev and stage with FF61, Win10x64 and Android 8.0

coll old form

willdurand:
Fix #5664

### DIFF
--- a/src/amo/components/AddAddonToCollection/index.js
+++ b/src/amo/components/AddAddonToCollection/index.js
@@ -1,8 +1,8 @@
 /* @flow */
-/* global window */
 import makeClassName from 'classnames';
 import * as React from 'react';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
 import { compose } from 'redux';
 
 import {
@@ -24,19 +24,26 @@ import type { DispatchFunc } from 'core/types/redux';
 import type { CollectionType } from 'amo/reducers/collections';
 import type { AppState } from 'amo/store';
 import type { ElementEvent } from 'core/types/dom';
+import type { ReactRouterType } from 'core/types/router';
 
 import './styles.scss';
 
 type Props = {|
-  _window: typeof window | Object,
   addon: AddonType | null,
+|};
+
+type InternalProps = {|
+  ...Props,
   // These are all user collections that the current add-on is a part of.
   addonInCollections: Array<CollectionType> | null,
   className?: string,
+  clientApp: string,
   currentUsername: string | null,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  router: ReactRouterType,
+  lang: string,
   loadingAddonsInCollections: boolean,
   loadingUserCollections: boolean,
   // These are all collections belonging to the user.
@@ -51,15 +58,12 @@ type SelectData = {|
   disabled: boolean,
 |};
 
-export class AddAddonToCollectionBase extends React.Component<Props> {
+export class AddAddonToCollectionBase extends React.Component<InternalProps> {
   optionSelectHandlers: { [key: string]: OnSelectOptionType };
 
-  static defaultProps = {
-    _window: typeof window !== 'undefined' ? window : {},
-  };
-
-  constructor(props: Props) {
+  constructor(props: InternalProps) {
     super(props);
+
     this.optionSelectHandlers = {};
   }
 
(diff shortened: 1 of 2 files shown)


willdurand:
I've see this `TODO` in the code, is it still valid?

```
          // TODO: show create collection overlay when it's implemented.
          // See
          // https://github.com/mozilla/addons-frontend/issues/4003
          // https://github.com/mozilla/addons-frontend/issues/3993
```

bobsilverberg:
The reference to https://github.com/mozilla/addons-frontend/issues/4003 was just a dependency on having the "Create collection" feature in the new UI, but the reference to https://github.com/mozilla/addons-frontend/issues/3993 is slightly different. It's about being able to create a new collection from this point, and have the add-on added to that collection after creation. Because a user is currently not able to select the "Create new collection" option and have the add-on "remembered", this patch does not regress functionality, but we may want to consider bumping the priority of #3993 if we ever intend to support it.

src/amo/components/AddAddonToCollection/index.js
```diff
@@ -24,19 +24,26 @@ import type { DispatchFunc } from 'core/types/redux';
...
 type Props = {|
-  _window: typeof window | Object,
   addon: AddonType | null,
+|};
+
+type InternalProps = {|
+  ...Props,
   // These are all user collections that the current add-on is a part of.
   addonInCollections: Array<CollectionType> | null,
   className?: string,
```
bobsilverberg: It doesn't look like `className` is internal.

tests/unit/amo/components/TestAddAddonToCollection.js
```diff
@@ -57,8 +59,12 @@ describe(__filename, () => {
     userId = 1,
     username = 'some-user',
     collections = [createFakeCollectionDetail({ authorUsername: username })],
+    lang,
```
bobsilverberg: Nit: Should these be in alphabetical order?

src/amo/components/AddAddonToCollection/index.js
willdurand: it does not look like it is used actually :p

willdurand:
> The reference to #4003 was just a dependency on having the "Create collection" feature in the new UI, but the reference to #3993 is slightly different. It's about being able to create a new collection from this point, and have the add-on added to that collection after creation. 

So could we remove the link to #4003?

> Because a user is currently not able to select the "Create new collection" option and have the add-on "remembered", this patch does not regress functionality, but we may want to consider bumping the priority of #3993 if we ever intend to support it.

Now that I fixed the navigation, the add-on could be stored into the redux state, and reused during the "create a collection" scenario. What do you think?

bobsilverberg:
> So could we remove the link to #4003?

Yes.

> Now that I fixed the navigation, the add-on could be stored into the redux state, and reused during the "create a collection" scenario. What do you think?

That's only a small part of the fix, because right now it's not possible to create a collection with an add-on in it, so we'd also have to adjust at least the saga, if not more. I think we should first figure out if we still want to support this, as it's been unsupported via the new front end since the new front end was implemented. I'll add a comment to the issue.

willdurand:
> That's only a small part of the fix, because right now it's not possible to create a collection with an add-on in it, so we'd also have to adjust at least the saga, if not more. I think we should first figure out if we still want to support this, as it's been unsupported via the new front end since the new front end was implemented. I'll add a comment to the issue.

I am a bit surprised. We do know how to add an add-on to a collection when it exists, so why can't we reuse this logic?

- User goes to the add-on detail page and click "create a collection"
- User is redirected to the create page
- User creates the collection
- Once created, the add-on is added automatically

bobsilverberg:
Yes, we could certainly do that, we just don't. As I mentioned in issue #3993, it would require some sort of UX change to inform the user, on the "Create a collection" page, that the specific add-on will be added after the collection is created, but that's certainly not a deal breaker.

willdurand:
> Yes, we could certainly do that, we just don't. As I mentioned in issue #3993, it would require some sort of UX change to inform the user, on the "Create a collection" page, that the specific add-on will be added after the collection is created, but that's certainly not a deal breaker.

That makes a lot of sense!

PR #5669, opened 2018-07-19
Create a CollectionAddAddon component to extract the display and logic for the AutoSearchInput for adding an add-on to a collection

### LINKED ISSUE #5667: Create a CollectionAddAddon component to extract the display and logic for the AutoSearchInput for adding an add-on to a collection
Supports #5635 

We're going to move the AutoSearchInput for adding an add-on to a collection from the current edit form into the card where the Sort select currently resides. To support this we should extract all the logic and display code for it into its own component.

bobsilverberg:
Fixes #5667 

Note that while the diff is quite large, it's really mostly just moving code from one component to another. I did also take the opportunity to use the new `withUIState` HOC for the new component because it seemed silly to create a new component and still use local state.

### DIFF
--- a/src/amo/components/CollectionAddAddon/index.js
+++ b/src/amo/components/CollectionAddAddon/index.js
@@ -28,24 +28,12 @@ import './styles.scss';
 export const MESSAGE_RESET_TIME = 5000;
 const MESSAGE_FADEOUT_TIME = 450;
 
-export const ADDON_ADDED_STATUS_NONE: 'ADDON_ADDED_STATUS_NONE' =
-  'ADDON_ADDED_STATUS_NONE';
-export const ADDON_ADDED_STATUS_PENDING: 'ADDON_ADDED_STATUS_PENDING' =
-  'ADDON_ADDED_STATUS_PENDING';
-export const ADDON_ADDED_STATUS_SUCCESS: 'ADDON_ADDED_STATUS_SUCCESS' =
-  'ADDON_ADDED_STATUS_SUCCESS';
-
-export type AddonAddedStatusType =
-  | typeof ADDON_ADDED_STATUS_NONE
-  | typeof ADDON_ADDED_STATUS_PENDING
-  | typeof ADDON_ADDED_STATUS_SUCCESS;
-
 type UIStateType = {|
-  addonAddedStatus: AddonAddedStatusType,
+  addonWasAdded: boolean,
 |};
 
 const initialUIState: UIStateType = {
-  addonAddedStatus: ADDON_ADDED_STATUS_NONE,
+  addonWasAdded: false,
 };
 
 export type Props = {|
@@ -81,9 +69,7 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
     const { hasAddonBeenAdded } = this.props;
     if (hasAddonBeenAdded !== hasAddonBeenAddedNew) {
       this.props.setUIState({
-        addonAddedStatus: props.hasAddonBeenAdded
-          ? ADDON_ADDED_STATUS_SUCCESS
-          : ADDON_ADDED_STATUS_NONE,
+        addonWasAdded: props.hasAddonBeenAdded,
       });
     }
 
@@ -111,15 +97,9 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
     } = this.props;
     const { addonId } = suggestion;
 
-    invariant(addonId, 'addonId cannot be empty');
-    invariant(
-      collection,
-      'A collection must be loaded before you can add an add-on to it',
-    );
-    invariant(
-      currentUsername,
-      'Cannot add to collection because you are not signed in',
-    );
+    invariant(addonId, 'addonId is required');
+    invariant(collection, 'collection is required');
+    invariant(currentUsername, 'currentUsername is required');
 
     dispatch(
       addAddonToCollection({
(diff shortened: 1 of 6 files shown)


src/amo/components/CollectionAddAddon/index.js
```diff
@@ -0,0 +1,207 @@
...
+          selectSuggestionText={i18n.gettext('Add to collection')}
+        />
+      </div>
+    );
+  }
+}
+
+export const extractId = (props: Props) => {
+  const { collection } = props;
+  return collection ? collection.slug : '';
```
kumar303: For practical purposes this might be fine but switching to `collection.id` would be more stable.

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -0,0 +1,207 @@
...
+  render() {
+    const { errorHandler, i18n, uiState } = this.props;
+
+    return (
+      <div className="CollectionAddAddon">
+        {errorHandler.hasError() ? (
+          errorHandler.renderError()
+        ) : (
+          <TransitionGroup className="NoticePlaceholder">
+            {uiState.addonAddedStatus === ADDON_ADDED_STATUS_SUCCESS && (
```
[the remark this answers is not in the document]
bobsilverberg: Yes, I must have done something differently during the initial implementation which required these different states, but then moved away from it, because they clearly aren't needed now. Thanks for catching this.

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -0,0 +1,187 @@
...
+
+  render() {
+    const { errorHandler, i18n, uiState } = this.props;
+
+    return (
+      <div className="CollectionAddAddon">
+        {errorHandler.hasError() ? (
+          errorHandler.renderError()
+        ) : (
+          <TransitionGroup className="NoticePlaceholder">
```
kumar303: Whoops, I didn't notice this last time. This class `NoticePlaceholder` needs to be prefixed with the component name (`CollectionAddAddon`). This is a convention we follow as an attempt to prevent CSS class name collisions. 

The [CSS style guide](https://github.com/mozilla/addons-frontend/blob/master/docs/css.md) explains this convention but we also need to make docs like this easier to find (see https://github.com/mozilla/addons-frontend/issues/2223).

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -0,0 +1,187 @@
...
+            {uiState.addonWasAdded && (
+              <CSSTransition
+                classNames="NoticePlaceholder-transition"
```
kumar303: Same comment here

src/amo/components/CollectionAddAddon/styles.scss
```diff
@@ -0,0 +1,35 @@
...
+    border: 1px solid $grey-50;
+    border-radius: $border-radius-xs;
+  }
+
+  .AutoSearchInput--autocompleteIsOpen .AutoSearchInput-query {
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+
+  .Notice {
```
kumar303: This is a bit confusing because I don't see any elements getting rendered underneath `<Notice />`. Could you add a comment about why it's needed? Maybe it's because the css transition component alters the dom?

As an aside, this transition functionality seems like it should live in `Notice` but we could worry about that later, when we want to add transitions to another notice.

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -0,0 +1,262 @@
...
+import Notice from 'ui/components/Notice';
+
+const simulateAutoSearchCallback = (props = {}) => {
+  return simulateComponentCallback({
+    Component: AutoSearchInput,
+    ...props,
+  });
+};
+
+const setAddonWasAddedUiStateToTrue = ({ username, root, store }) => {
```
kumar303: I think a better name for this would be `addAddonToCollection` or something that conveys the user's *action* rather than the component's internal representation of the user's action.

We typically name these types of functions the same as the dispatched action but with an underscore prefix. In this case: `_addonAddedToCollection`. It's not a hard and fast rule, though.

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -0,0 +1,262 @@
...
+      createFakeAutocompleteResult({ name: 'uBlock Origin' }),
+    );
+    const selectSuggestion = simulateAutoSearchCallback({
+      root,
+      propName: 'onSuggestionSelected',
+    });
+    selectSuggestion(suggestion);
+
+    applyUIStateChanges({ root, store });
+    expect(root.instance().props.uiState.addonWasAdded).toEqual(false);
```
kumar303: This is testing the implementation of the component rather than the functionality of the component. I think it's better to test functionality because it lets the implementation evolve freely. 

I think you can actually just delete this test. The functionality is already tested in *it removes the notification after a new add-on has been selected*.

bobsilverberg:

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -0,0 +1,187 @@
...
+        errorHandlerId: errorHandler.id,
+        filters,
+        slug: collection.slug,
+        username: currentUsername,
+      }),
+    );
+    this.props.setUIState({ addonWasAdded: false });
+  };
+
+  timeout: TimeoutID;
```
kumar303: I guess we don't have a rule for this but these typically go at the top of the class. If you get a lint error from react-sort-order (or whatever) just disable it at the top. I hate that rule.

src/amo/components/CollectionAddAddon/styles.scss
```diff
@@ -0,0 +1,37 @@
...
+    border-radius: $border-radius-xs;
+  }
+
+  .AutoSearchInput--autocompleteIsOpen .AutoSearchInput-query {
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+
+  // The Notice component creates children which are affected by these classes,
+  // which are added by the CSSTransition component.
```
kumar303: Thanks

Issue #5672, opened 2018-07-19
The loading screen is jarring while editing a collection


kumar303:
### Describe the problem and steps to reproduce it:

- Log in
- Create a new collection
- Edit the collection
- Add/remove add-ons

### What happened?

Each time you add or remove an add-on, the loading screen fills the whole page which is jarring.

### What did you expect to happen?

The loading screen could be a little less jarring.

AlexandraMoga:
Verified fixed on AMO -dev with FF61, Win10x64

![less loading screen](https://user-images.githubusercontent.com/31961530/43071805-8c1372bc-8e7c-11e8-8fa0-28125ed9e1b4.gif)

PR #5673, opened 2018-07-19
A less jarring loading screen while editing a collection

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5672

I noticed this while testing out the spiffy new collections and it bugged me. I thought it was easy to fix and it totally wasn't! But I started it towards the end of the day anyway. Sigh.

<details>
<summary>Before</summary>

addons-frontend-collection-before mov

</details>

<details>
<summary>After</summary>

addons-frontend-collection-after mov

</details>

### DIFF
--- a/src/amo/components/Collection/index.js
+++ b/src/amo/components/Collection/index.js
@@ -94,6 +94,8 @@ export type SaveAddonNoteFunc = (
 ) => void;
 
 export class CollectionBase extends React.Component<InternalProps> {
+  addonPlaceholderCount: number;
+
   static defaultProps = {
     _config: config,
     _isFeaturedCollection: isFeaturedCollection,
@@ -101,8 +103,25 @@ export class CollectionBase extends React.Component<InternalProps> {
     editing: false,
   };
 
+  constructor(props: InternalProps) {
+    super(props);
+    this.addonPlaceholderCount = 5;
+    this.maybeResetAddonPlaceholderCount();
+  }
+
+  maybeResetAddonPlaceholderCount() {
+    const { collection } = this.props;
+    if (collection && collection.addons && collection.addons.length) {
+      this.addonPlaceholderCount = collection.addons.length;
+    }
+  }
+
   componentWillMount() {
     this.loadDataIfNeeded();
+  }
+
+  componentDidUpdate() {
+    this.maybeResetAddonPlaceholderCount();
   }
 
   componentWillReceiveProps(nextProps: InternalProps) {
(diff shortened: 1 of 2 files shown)


src/amo/components/Collection/index.js
```diff
@@ -94,17 +94,36 @@ export type SaveAddonNoteFunc = (
...
 
+  constructor(props: InternalProps) {
+    super(props);
+    this.addonPlaceholderCount = 5;
+    this.maybeResetAddonPlaceholderCount();
+  }
+
+  maybeResetAddonPlaceholderCount() {
+    const { collection } = this.props;
+    if (collection && collection.addons && collection.addons.length) {
```
bobsilverberg: It seems like this could just be: 

`this.addonPlaceholderCount = collection && collection.addons && collection.addons.length || 5;`

and you could then do away with setting the default in `constructor`. Maybe using a `const` in place of the `5` would be a good idea in this case.

tests/unit/amo/components/TestCollection.js
```diff
@@ -572,6 +594,58 @@ describe(__filename, () => {
     expect(wrapper.find(AddonsCard)).toHaveProp('editing', false);
   });
 
+  it('sets a default placeholder count', () => {
+    const wrapper = renderComponent();
+    expect(wrapper.find(AddonsCard)).toHaveProp('placeholderCount', 5);
```
bobsilverberg: If we made this a `const` and exported it then we could use the value from the component here, rather than hardcoding `5`.

src/amo/components/Collection/index.js
kumar303: This won't work because when `loading=true`, `collection` is null which means it will reset the value back to 5. We want it to use the previous value instead.

This whole approach is a little bit un-react-y in that it relies on the state of the component. The `render()` cycle is not working with the current value of `addonPlaceholderCount`, it's working with the *previous* value. This is just one of those patterns that's hard to achieve cleanly in React.

kumar303:
> What do you think about using 3 for the default, rather than 5?

Sounds good. I changed it to 3. I'm not really sure what the default number of add-ons in a collection would be. I should really ask for access to STMO.

kumar303:
This seemed like a general purpose feature so I looked into moving the logic to `AddonsCard`. None of the other components using `AddonsCard` would benefit from it, though. We should keep this in mind in case we need it somewhere else. For now, YAGNI, so, I won't move it.

src/amo/components/Collection/index.js
kumar303: I added a comment to this code to make it more clear.

PR #5676, opened 2018-07-20
 Move the select which is used to add an add-on to a collection out of the edit form

### LINKED ISSUE #4617: Search field in edit collection form has an action arrow with no functionality
Follow up for #4579 

STR:

1. Log in to AMO-dev and select a collection to view its details page (i.e. https://addons-dev.allizom.org/en-US/firefox/collections/mozilla/bookmark-managers/)
2. Click on Edit collection link
3. In the edit collection dialog, look at the search field in the far right

Actual result:
There is an action arrow similar to the one from the AMO header search, but for collections this has no practical function

Expected result:
The arrow should not be present in the search field included in the edit collection component

Notes:
- reproduced on AMO-dev, FF59, Win10x64 and Android 7.0

arrow

### LINKED ISSUE #5286: "This add-on already belongs to the collection" should be displayed above Search
STR:
1. Load AMO dev and go to your collections, edit one of the existing collections or create a new one
2. Add an extension or theme to the collection
3. Try to add the previous extension again to the collection and observe the message displayed

Expected result:
"This add-on already belongs to the collection" message is displayed above the Search.

Actual result:
"This add-on already belongs to the collection" message is displayed above Collection name field.

Notes:
Reproducible on AMO dev with FF60(Win10&Android 8.0).

On small devices such as android phones if the message is above "Collection name" name field to observe the red message the user must scroll to top.
the red message - small devices

### LINKED ISSUE #5675: Move the select which is used to add an add-on to a collection out of the edit form
Move the select which is used to add an add-on to a collection from the edit form into the card that appears below the edit form (the one that currently holds the sort select).

This is an attempt to make it more obvious that saving collection metadata is an operation independent of adding add-ons to the collection.

**For QA** please make sure that all collection editing functionality (saving metadata and adding/removing add-ons) still works.

bobsilverberg:
Fixes #5675 
Fixes #5286
Fixes #4617
Depends on #5669 

There is one UX quirk that may be an issue. When you add an add-on to a collection, the card increases in size to show the "Added to collection" notice, and then, when the notice fades away, the card reduces in size. The fact that it changes its size to show the notice is consistent with what we do when there's an error, but it's a bit odd because of the notice disappearing and the card shrinking. See the screenshot as an example:

jul-20-2018 09-47-13

The only option I can think of to mitigate this is to reserve enough blank space at the top of the card for the message, but then that looks bad whenever the message is not visible, which is why I wrote it the way it currently is. Do you have an opinion @pwalm?

Before:

screenshot 2018-07-20 09 51 12
After:

screenshot 2018-07-20 09 49 27

### DIFF
--- a/src/amo/components/CollectionAddAddon/styles.scss
+++ b/src/amo/components/CollectionAddAddon/styles.scss
@@ -4,8 +4,7 @@
 .CollectionAddAddon {
   .ErrorList,
   .CollectionAddAddon-noticePlaceholder {
-    margin-top: 16px;
-    min-height: 32px;
+    margin: 0;
   }
 
   .AutoSearchInput-label {
--- a/src/amo/components/CollectionAddAddon/styles.scss
+++ b/src/amo/components/CollectionAddAddon/styles.scss
@@ -2,18 +2,33 @@
 @import '~amo/css/inc/vars';
 
 .CollectionAddAddon {
+  margin-bottom: 24px;
+
   .ErrorList,
   .CollectionAddAddon-noticePlaceholder {
-    margin: 0;
+    margin-bottom: 6px;
   }
 
   .AutoSearchInput-label {
-    padding-top: 4px;
+    padding-bottom: 6px;
+  }
+
+  .AutoSearchInput-search-box {
+    padding-top: 6px;
   }
 
   .AutoSearchInput-query {
     border: 1px solid $grey-50;
     border-radius: $border-radius-xs;
+  }
+
+  .AutoSearchInput-icon-magnifying-glass {
+    top: 12px;
+  }
+
+  .AutoSearchInput-submit-button,
+  .SearchSuggestion-icon-arrow {
+    display: none;
   }
 
   .AutoSearchInput--autocompleteIsOpen .AutoSearchInput-query {
--- a/src/amo/components/CollectionSort/styles.scss
+++ b/src/amo/components/CollectionSort/styles.scss
@@ -5,7 +5,3 @@
   display: block;
   padding-bottom: 6px;
 }
-
-.CollectionSort-select {
-  margin: 0 0 6px;
-}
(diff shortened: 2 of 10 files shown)


bobsilverberg:
Thanks for the review @kumar303. I believe I have addressed all the layout issues, although the AutoSearchInput was a bit difficult to deal with. Please take another look and let me know what you think.

src/amo/components/CollectionSort/styles.scss
```diff
@@ -1,12 +1,9 @@
...
-  margin-top: 12px;
-}
-
 .CollectionSort-label {
   color: $grey-50;
   display: block;
+  padding-bottom: 6px;
 }
 
 .CollectionSort-select {
```
kumar303: The old code had `margin: 0 0 6px` but you can delete that entirely. You can just rely on the card padding. Here you will see that the extra margin causes the layout to look lopsided:

screenshot 2018-07-24 14 24 09
screenshot 2018-07-24 14 28 14

src/amo/components/CollectionAddAddon/styles.scss
```diff
@@ -2,21 +2,35 @@
...
     border-radius: $border-radius-xs;
   }
 
+  .AutoSearchInput-icon-magnifying-glass {
+    top: 12px;
+  }
+
+  .AutoSearchInput-submit-button,
+  .SearchSuggestion-icon-arrow {
+    display: none;
```
kumar303: This seems fine but I feel like their might be an accessibility issue with hiding the "button" (if anyone even knows that it's a button). We can worry about it later, though. The button was just a no-op before.

src/amo/components/CollectionSort/styles.scss
kumar303: btw, I was able to use the Redux devtool to travel back in time to this state. However, the auto search input breaks when you time travel since the third party component we use is implemented with `setState()`.

PR #5680, opened 2018-07-20
Redirect to the slug that is returned by the API call when creating or editing a collection

### LINKED ISSUE #5424: Creating a new collection named "favorites" will generate a 404 on the new frontend
STR:

1. Log in to AMO -dev with a user that doesn't have a 'My favorite add-on' collection
2. Go to My Collections and click on Create Collection
3. Enter _favorites_ in the collection name field
4. Click on Create Collection

Actual result:
A Page not Found is displayed

Expected result:
The user remains on the same screen and continues to edit his collection

Notes:
 - reproduced on AMO -dev with FF60, Win10x64

favorites 404

bobsilverberg:
Fixes #5424

### DIFF
--- a/tests/unit/amo/sagas/test_collections.js
+++ b/tests/unit/amo/sagas/test_collections.js
@@ -567,7 +567,9 @@ describe(__filename, () => {
           slug: returnedSlug,
         });
 
-        mockApi.expects('updateCollection').returns(collectionDetailResponse);
+        mockApi
+          .expects('updateCollection')
+          .returns(Promise.resolve(collectionDetailResponse));
 
         const collectionSlug = 'some-collection';
         const updateFilters = { page: 2 };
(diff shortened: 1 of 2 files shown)


tests/unit/amo/sagas/test_collections.js
```diff
@@ -559,22 +559,28 @@ describe(__filename, () => {
...
-        mockApi.expects('updateCollection').returns(Promise.resolve());
+      it('redirects to the resulting slug after update', async () => {
+        const submittedSlug = 'submitted-slug';
+        const returnedSlug = 'returned-slug';
+
+        const collectionDetailResponse = createFakeCollectionDetail({
+          slug: returnedSlug,
+        });
+
+        mockApi.expects('updateCollection').returns(collectionDetailResponse);
```
kumar303: This should return a promise:
```js
        mockApi
          .expects('updateCollection')
          .returns(Promise.resolve(collectionDetailResponse));
```
I think it only works as is because sinon stubs are a bit magical and they have a fallback promise implementation (this is just a guess). Relying on magic makes me nervous.

PR #5683, opened 2018-07-20
WIP-Disco pane static theme functionality updates

### LINKED ISSUE #5681: Update logic in disco pane so static themes are installed similiarly to lightweight themes

Right now, you can't see the theme when it's toggled on. It seems to install but be enabled

You should also be able to click the image to "toggle on"

If should appear to work the same way lightweight themes

notes for dev: 
if you run `setCurrentStatus` on the change of status in `componentsWillReceiveProps`, you see the following with toggling on/off:

first toggle on: 
you see theme

first toggle off:
you see theme turn off and uninstalls

2nd toggle on:
it's seems to download/install but it gets disabled. You can click the switch here and enables  ** this is the issue - need to look into this more

2nd toggle off:
you see theme turns off and uninstalls

rebmullin:
fixes #5681 

branched off of #5625 (which addresses more visual parts)

TODO: update tests

### DIFF
--- a/src/amo/components/Header/index.js
+++ b/src/amo/components/Header/index.js
@@ -146,6 +146,15 @@ export class HeaderBase extends React.Component {
                   {i18n.gettext('Submit a New Theme')}
                 </Link>
               </DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link
+                  className="Header-user-menu-developers-submissions-link"
+                  href="/developers/addons/"
+                  prependClientApp={false}
+                >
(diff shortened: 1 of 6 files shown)


willdurand:
@rebmullin the `InstallSwitch` code is about to be removed because of #3954. If you could explain clearly what is needed in an issue, I could probably add the feature in the `InstallButton`.

willdurand:
Thanks @rebmullin! We can keep this PR opened (and as-is). I'll try to merge it into #5712.

willdurand:
This patch has been merged into #5712.

PR #5689, opened 2018-07-23
Remove unused webpack-dev-server dep

willdurand:
Removing this dev dependency as we are not using it, we use middleware instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -292,7 +292,6 @@
     "tosource": "^1.0.0",
     "webpack": "^3.1.0",
     "webpack-dev-middleware": "^1.10.2",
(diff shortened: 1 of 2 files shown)


willdurand:
> We have a lint rule to catch extraneous package dependencies but I guess it's turned off for dev deps: mozilla/eslint-config-amo:index.js@master#L60

I guess lots of false-positive. I tried a few tools to find more unused deps and... that's often not accurate.

kumar303:
> We have a lint rule to catch extraneous package dependencies

Ah, wait, that's serving a different purpose: it makes sure everything in `src` is defined as prod deps.

PR #5692, opened 2018-07-23
Accept numeric IDs to browse a user profile page

### LINKED ISSUE #5594: Support numeric IDs in user profile URLs
The legacy frontend allows to browse a user profile page using the numeric ID of a user and not their username: `/user/123/`. Having a numeric username is not allowed so there is no conflict.

The new/current user profile pages require the username, and passing the user ID will not work (404 returned). FWIW, the change to make it work is quite easy.

@kewisch told me he used such URLs with numeric IDs in STMO (but could change to use usernames instead).

Do we want to support numeric IDs in user profile URLs? Only for user profile pages, not for editing a user though (so only URLs like `/user/{ID or username}/`).

willdurand:
Fix #5594 

---

I had this fix in one of my local branches... This PR allows a user profile page to be shown when using a user ID in the URL (for backward compatibility purpose).

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -18,6 +18,7 @@ import { getReviewsByUserId } from 'amo/reducers/reviews';
 import {
   fetchUserAccount,
   getCurrentUser,
+  getUserById,
   getUserByUsername,
   hasPermission,
   isDeveloper,
@@ -141,9 +142,9 @@ export class UserProfileBase extends React.Component<InternalProps> {
   }
 
   getURL() {
-    const { params } = this.props;
+    const { params, user } = this.props;
 
-    return `/user/${params.username}/`;
+    return `/user/${user ? user.username : params.username}/`;
   }
 
   getEditURL() {
@@ -382,7 +383,7 @@ export class UserProfileBase extends React.Component<InternalProps> {
                 <AddonsByAuthorsCard
                   addonType={ADDON_TYPE_EXTENSION}
                   authorDisplayName={user ? user.name : params.username}
-                  authorUsernames={[params.username]}
+                  authorUsernames={[user ? user.username : params.username]}
                   errorHandler={errorHandler}
                   numberOfAddons={EXTENSIONS_BY_AUTHORS_PAGE_SIZE}
                   pageParam="page_e"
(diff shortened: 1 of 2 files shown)


src/amo/components/UserProfile/index.js
```diff
@@ -415,8 +416,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
 }
 
 export function mapStateToProps(state: AppState, ownProps: Props) {
+  const { username } = ownProps.params;
+
   const currentUser = getCurrentUser(state.users);
-  const user = getUserByUsername(state.users, ownProps.params.username);
+
+  // `getUserByUsername()` requires a string as second argument.
+  let user = getUserByUsername(state.users, `${username}`);
```
willdurand: I am pretty sure this only occurs in the test suite, not in real life..

src/amo/components/UserProfile/index.js
```diff
@@ -141,9 +142,9 @@ export class UserProfileBase extends React.Component<InternalProps> {
   }
 
   getURL() {
-    const { params } = this.props;
+    const { params, user } = this.props;
 
-    return `/user/${params.username}/`;
+    return `/user/${user ? user.username : params.username}/`;
```
bobsilverberg: Because you have this logic three times in the file, might it be worth creating a `getUsername` helper that returns `user ? user.username : params.username`?

Also, I'm not sure if it matters (I imagine not), but this will also cause the links in the `Paginate` component and the `AddonsByAuthorsCard`s to be converted from numeric ids into usernames. I.e., if I visit `/user/12345/`, the links will be to `/user/some-username/`.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -926,6 +926,15 @@ describe(__filename, () => {
     expect(root.find(NotFound)).toHaveLength(1);
   });
 
+  it('renders a user profile when URL contains a user ID', () => {
+    const { store } = signInUserWithUsername('black-panther');
+    const user = getCurrentUser(store.getState().users);
+
+    const root = renderUserProfile({ params: { username: user.id }, store });
+
+    expect(root.find('.UserProfile')).toHaveLength(1);
```
bobsilverberg: It seems like this test should also be checking that the profile page that is rendered in fact belongs to the user who corresponds with the `id` used? It's probably also a good idea to check the properties passed into `Paginate` and the `AddonsByAuthorsCard`s because that logic has changed with this patch.

src/amo/components/UserProfile/index.js
willdurand: Yes, I think that's fine if we link to username in the `Paginate` component. The main idea is to allow the legacy URLs to be served.

willdurand:
@bobsilverberg sorry for the previous iteration, you're right: more tests were needed!

PR #5694, opened 2018-07-23
Allow a Button to be read only

### LINKED ISSUE #5623: Try to create a read-only Button component (e.g. for styling `input[type="file"]`) instead of reusing its CSS
In the `UserProfileEditPicture` component, the style of the `Button` component is used to create a fake button (`<span />`) that is used to style a `input[type="file"]`.

Instead of borrowing the style, we should create a read-only `Button` component OR update the existing `Button` component to render a `<span />` given certain prop (`readonly` or `fake` maybe).

willdurand:
Fix #5623

### DIFF
--- a/src/amo/components/UserProfileEditPicture/index.js
+++ b/src/amo/components/UserProfileEditPicture/index.js
@@ -5,6 +5,7 @@ import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
 import withUIState from 'core/withUIState';
+import Button from 'ui/components/Button';
 import ConfirmButton from 'ui/components/ConfirmButton';
 import UserAvatar from 'ui/components/UserAvatar';
 import type { UserType } from 'amo/reducers/users';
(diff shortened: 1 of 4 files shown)


src/amo/components/UserProfileEditPicture/index.js
```diff
@@ -93,7 +86,15 @@ export class UserProfileEditPictureBase extends React.Component<InternalProps> {
             onFocus={this.onFocus}
             type="file"
           />
-          <span className={buttonClass}>{i18n.gettext('Choose Photo…')}</span>
+          <Button
+            buttonType="action"
+            disabled={!user}
+            className="UserProfileEditPicture-select-button"
```
rebmullin: nit: c before d  : )

willdurand:
> I saw in the issue you put "fake"... I am not sure if that is the right name either but it may be a little bit clearer...so something similar to that? What do you think?

There was already a `disabled` prop, so I guess `readOnly` would make sense but I am open to suggestions. I don't like `fake` because it is a real button, it does not link to anything though (hence the "read only").

bobsilverberg:
When I took a quick look at this I also thought that `readOnly` wasn't a good word for it. How about `createLink: false`?

rebmullin:
Yeh maybe that or maybe ~ `mockButton`? or `triggerInput`?

willdurand:
> When I took a quick look at this I also thought that readOnly wasn't a good word for it. How about createLink: false?

> Yeh maybe that or maybe ~ mockButton? or triggerInput?

How about `noLink` then?

willdurand:
@rebmullin updated with `noLink` as prop name, thanks a lot!

Issue #5670, opened 2018-07-19
Add analytics event to "Take the survey" button


jenniferdavidson:
Add an analytics event on the "Take the survey" button that's part of the message bar implemented here in Issue #5438 so we can tell how many people have clicked on the button.

kumar303:
From IRC:

> Surveygizmo only records if they’ve completed the first page.

so we don't really have any *view* events at all.

jenniferdavidson:
@kumar303 Correct. [Screenshot of Fall Off Report from SurveyGizmo](https://screenshots.firefox.com/X4mIK3s0YZhzFfet/app.surveygizmo.com)

kumar303:
OK, we'll have analytics going out next push (Thursday). The new google analytics category will be `AMO Addon / Experience Survey Notice` and we'll have these events:

- `Shown survey notice` (notice appeared for this user)
- `Visited survey` (they clicked on the button)
- `Dismissed survey notice` (they clicked the X)

PR #5696, opened 2018-07-23
Track interaction with the survey notice for analytics

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5670

### DIFF
--- a/src/core/components/SurveyNotice/index.js
+++ b/src/core/components/SurveyNotice/index.js
@@ -9,6 +9,7 @@ import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
 import { dismissSurvey } from 'core/reducers/survey';
+import tracking from 'core/tracking';
 import Notice from 'ui/components/Notice';
 import type { ReactRouterLocation } from 'core/types/router';
 import type { I18nType } from 'core/types/i18n';
@@ -16,6 +17,11 @@ import type { DispatchFunc } from 'core/types/redux';
 import type { AppState } from 'amo/store';
 
 import './styles.scss';
+
+export const TRACKING_CATEGORY = 'AMO Addon / Experience Survey Notice';
+export const TRACKING_ACTION_DISMISSED = 'Dismissed survey notice';
+export const TRACKING_ACTION_SHOWN = 'Shown survey notice';
+export const TRACKING_ACTION_VISITED = 'Visited survey';
 
 type Props = {|
   location: ReactRouterLocation,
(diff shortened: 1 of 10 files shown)


src/core/components/SurveyNotice/index.js
```diff
@@ -73,34 +92,58 @@ export const SurveyNoticeBase = ({
...
+        .slice(2)
+        .join('/'),
+    });
+    const surveyUrl = `https://qsurvey.mozilla.com/s3/addons-mozilla-org-survey?${source}`;
+
+    return (
+      <Notice
+        actionHref={surveyUrl}
+        actionOnClick={this.onClickSurveyLink}
+        actionTarget="_blank"
```
muffinresearch: Is this setting `rel="noopener noreferrer"` somewhere else? If not it should be - see https://mathiasbynens.github.io/rel-noopener/ for more info.

src/core/components/SurveyNotice/index.js
```diff
@@ -26,40 +33,52 @@ type InternalProps = {|
...
+      'de',
+      'en-US',
+      'es',
+      'fr',
+      'ja',
+      'pl',
+      'pt-BR',
+      'ru',
+      'zh-CN',
+      'zh-TW',
```
willdurand: Would it make sense to have this array in `config/`?

src/core/components/SurveyNotice/index.js
```diff
@@ -73,34 +92,58 @@ export const SurveyNoticeBase = ({
...
+      category: SURVEY_CATEGORY,
+    });
+  };
+
+  onClickSurveyLink = () => {
+    this.dismissNotice();
+    this.props._tracking.sendEvent({
+      action: SURVEY_ACTION_VISITED,
+      category: SURVEY_CATEGORY,
+    });
```
willdurand: I suppose we could have a method in this component to `sendTrackingEvent({ action })` since only the `action` changes.

src/core/components/SurveyNotice/index.js
```diff
@@ -73,34 +92,58 @@ export const SurveyNoticeBase = ({
...
+
+    // Pass along a source derived from the current URL path but with
+    // the preceding language path removed.
+    const source = querystring.stringify({
+      source: location.pathname
+        .split('/')
+        .slice(2)
+        .join('/'),
+    });
+    const surveyUrl = `https://qsurvey.mozilla.com/s3/addons-mozilla-org-survey?${source}`;
```
willdurand: We have a `addQueryParams()` helper too in `core/utils`. I am wondering if that would do the trick here.

tests/unit/core/components/TestSurveyNotice.js
```diff
@@ -32,6 +40,12 @@ describe(__filename, () => {
     return shallowUntilTarget(<SurveyNotice {...props} />, SurveyNoticeBase);
   };
 
+  const getNoticeProp = (root, prop) => {
+    const notice = root.find(Notice);
+    expect(notice).toHaveProp(prop);
+    return notice.props()[prop];
```
willdurand: Can we write `notice.prop(prop)` instead?

tests/unit/helpers.js
```diff
@@ -654,3 +654,11 @@ export function setUIState({ root, change, store }) {
 export function fakeCookie(overrides = {}) {
   return { load: sinon.stub(), save: sinon.stub(), ...overrides };
 }
+
+export const createFakeTracking = (overrides = {}) => {
+  return {
+    sendEvent: sinon.stub(),
+    setDimension: sinon.stub(),
+    ...overrides,
+  };
```
willdurand: Thank you for this helper.

tests/unit/core/components/TestSurveyNotice.js
kumar303: Hah. Whoops.

src/core/components/SurveyNotice/index.js
kumar303: It would add some indirection and I don't think we'll need it. The survey will be disabled soon.

src/core/components/SurveyNotice/index.js
kumar303: Ah, thanks. I think we need to fix this at a lower level: https://github.com/mozilla/addons-frontend/issues/5700

For now, I just fixed the `Notice` button.

kumar303:
@willdurand thanks. It's ready for another look.

PR #5697, opened 2018-07-23
Disable the "Save collection" button unless one of the fields in the edit form has been changed

### LINKED ISSUE #5635: Disable the "Save collection" button unless one of the fields in the edit form has been changed

bobsilverberg:
Fixes #5635

### DIFF
--- a/tests/unit/amo/components/TestCollectionManager.js
+++ b/tests/unit/amo/components/TestCollectionManager.js
@@ -432,9 +432,15 @@ describe(__filename, () => {
   });
 
   it('disables and enables form buttons when modification status changes', () => {
-    const renderAndCheckButtons = (shouldBeDisabled) => {
-      const root = render();
+    const name = 'Collection name';
 
+    const collection = createInternalCollection({
+      detail: createFakeCollectionDetail({ name }),
+    });
+
+    let root = render({ collection });
+
+    const buttonsAreDisabled = (shouldBeDisabled) => {
       expect(root.find('.CollectionManager-cancel')).toHaveProp(
         'disabled',
         shouldBeDisabled,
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -432,9 +432,15 @@ describe(__filename, () => {
...
-      const root = render();
+    const name = 'Collection name';
 
+    const collection = createInternalCollection({
+      detail: createFakeCollectionDetail({ name }),
+    });
+
+    let root = render({ collection });
+
+    const buttonsAreDisabled = (shouldBeDisabled) => {
```
willdurand: I think it makes sense to have such a helper but I would pass `root` to be more explicit and maybe find a more suitable name to reflect the expectations. Could be `checkButtonsState({ root, disabled: true/false });`

Would that make sense?

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -445,12 +451,62 @@ describe(__filename, () => {
...
+    const name = 'Collection name';
+    const slug = 'collection-slug';
+
+    const collection = createInternalCollection({
+      detail: createFakeCollectionDetail({ description, name, slug }),
+    });
+
+    const root = render({ collection });
+
+    const saveIsDisabled = (isDisabled) => {
```
willdurand: Same comment as above for this function.

PR #5704, opened 2018-07-24
Sort by popularity should use descending order

### LINKED ISSUE #5581: Sort by popularity should use descending order
### STR:
1.  Open any collection i.e: https://addons.allizom.org/en-US/firefox/collections/mozilla/ad-blockers/
2. Go to "Sort add-ons by" drop down menu and select popularity.

### Expected result:
Most popular add-ons are displayed first

### Actual result
The add-ons are not sorted by popularity

### Notes:
Verified in FF61 Win7x64. Issue can be reproduced in amo-dev amo-stage and amo-prod
Please view this screencast:
collectionspopularity

bobsilverberg:
Fixes #5581 

Note that I looked for an opportunity to test this change, but as all we're doing is changing the value of a constant, there isn't really anything to test. We don't generally test values of constants.

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -140,7 +140,8 @@ export const SEARCH_SORT_UPDATED = 'updated';
 export const COLLECTION_SORT_DATE_ADDED_ASCENDING = 'added';
 export const COLLECTION_SORT_DATE_ADDED_DESCENDING = '-added';
 export const COLLECTION_SORT_NAME = 'name';
-export const COLLECTION_SORT_POPULARITY = 'popularity';
+// Note that popularity sort should be in descending order.
+export const COLLECTION_SORT_POPULARITY = '-popularity';
 
 // Operating system for add-ons and files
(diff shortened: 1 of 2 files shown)


src/core/constants.js
```diff
@@ -140,7 +140,8 @@ export const SEARCH_SORT_UPDATED = 'updated';
 export const COLLECTION_SORT_DATE_ADDED_ASCENDING = 'added';
 export const COLLECTION_SORT_DATE_ADDED_DESCENDING = '-added';
 export const COLLECTION_SORT_NAME = 'name';
-export const COLLECTION_SORT_POPULARITY = 'popularity';
+// Note that popularity sort should be in descending order.
+export const COLLECTION_SORT_POPULARITY = '-popularity';
```
willdurand: Can we suffix this constant with `_DESCENDING` to make things explicit (and avoid the comment here)?

bobsilverberg: Good idea, thanks!

PR #5706, opened 2018-07-24
Add a AnimatedIcon component

willdurand:
Refs #3954 

---

I want to extract this component from the rest of the PR for #3954 because it will be quite big otherwise. The icon looks like this:


tests/unit/ui/components/TestAnimatedIcon.js
```diff
@@ -0,0 +1,35 @@
...
+
+    expect(root.find(Icon)).toHaveLength(1);
+    expect(root).toHaveClassName('AnimatedIcon');
+    expect(root).toHaveClassName('AnimatedIcon-foo');
+  });
+
+  it('allows a custom className', () => {
+    const root = shallow(<AnimatedIcon name="bar" className="sup" />);
+
+    expect(root).toHaveClassName('AnimatedIcon');
```
bobsilverberg: Do you need these first two assertions? They are already done in the first test. I suppose you may want to make sure that when a custom `className` is passed it doesn't mess with the normal expected `classNames`?

tests/unit/ui/components/TestAnimatedIcon.js
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import AnimatedIcon from 'ui/components/AnimatedIcon';
+import Icon from 'ui/components/Icon';
+
+describe(__filename, () => {
+  it('renders an Icon and sets CSS classes', () => {
+    const root = shallow(<AnimatedIcon name="foo" />);
```
bobsilverberg: It may be overkill, but I think we tend to use consts in any case where we have a value in setup and in assertions, i.e., 

```js
const name = 'foo';
const root = shallow(<AnimatedIcon name={name} />);
expect(root).toHaveClassName(`AnimatedIcon-${name}`);
```

This comment applies to all the tests below.

tests/unit/ui/components/TestAnimatedIcon.js
willdurand: Yes, I don't want to replace the default `className`, I want to add it to the list of default ones.

muffinresearch:
Is there any alternative ways to achieve the same/similar end result without such a large SVG? I'm sure it will compress and can be minfied but this seems quite heavy at 77.2 KB.

willdurand:
> Is there any alternative ways to achieve the same/similar end result without such a large SVG? I'm sure it will compress and can be minfied but this seems quite heavy at 77.2 KB.

I have reduced its size a bit but forgot to optimize it with a tool.

willdurand:
I reduced the size of the SVG by 2.7

muffinresearch:
> I reduced the size of the SVG by 2.7

That's a good improvement. I wonder if this could be done without a sprite at all. But we can probably consider that separately without blocking on this.

willdurand:
> That's a good improvement. I wonder if this could be done without a sprite at all? But we can probably consider that separately without blocking on this.

👍

PR #5710, opened 2018-07-24
Update user profile copy

### LINKED ISSUE #5560: The delete user account modal should contain more info
### STR:
1.  Go to edit user profile page: https://addons.allizom.org/en-US/firefox/users/edit
2. Click the "Delete My Account" button
3. Observe the wording

### Actual result
There is no mention that the user will be able to log in again with the deleted account but that account will not be linked with the deleted one.

### Expected result:
The wording should look like on the old pages: 
oldpagesdeleteaccount

### Notes:
- reproduced on AMO -dev  and AMO -stage, FF60 Win7
Please view this screencast: 
newpagesdeleteaccount

willdurand:
Fix #5560

---

Copy: https://docs.google.com/document/d/1nnv1gwiQe5K3tvYXdsA2QmtWKO6XD75aD31V0hoyI_k/edit?ts=5b5600fb

### DIFF
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -541,7 +541,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                     href="https://accounts.firefox.com/settings"
                     className="UserProfileEdit-manage-account-link"
                   >
-                    {i18n.gettext('Manage Firefox Accounts...')}
+                    {i18n.gettext('Manage Firefox Accounts…')}
                   </a>
                 )}
               </div>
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestUserProfileEdit.js
```diff
@@ -1254,7 +1254,7 @@ describe(__filename, () => {
     expect(modal).toHaveLength(1);
     expect(modal).toHaveProp(
       'header',
-      'IMPORTANT: Deleting your account is irreversible.',
+      'IMPORTANT: Deleting your Firefox Add-ons profile is irreversible.',
```
bobsilverberg: Similar to my comment above, this is essentially static text, except for the fact that it conditionally says either "your Firefox Add-ons profile" or "this Firefox Add-ons profile", so it might ease the maintenance burden if we just check that the `header` contains the part of the string we expect to change.

Although maybe that will make the test more confusing, so I suppose either way is suboptimal. 🤷‍♂️

src/amo/components/UserProfileEdit/index.js
```diff
@@ -782,21 +782,31 @@ export class UserProfileEditBase extends React.Component<Props, State> {
...
+              </p>
+            )}
             <p
               // eslint-disable-next-line react/no-danger
               dangerouslySetInnerHTML={sanitizeHTML(
                 i18n.sprintf(
                   isEditingCurrentUser
                     ? i18n.gettext(
                         `%(strongStart)sNOTE:%(strongEnd)s You cannot delete
-                        your account if you are the %(linkStart)sauthor of any
```
bobsilverberg: I know this is older code, but should these links open in a new tab/window? It seems odd to interrupt the flow of deleting a user profile by navigating away from this modal. It would take multiple steps to get back here after looking at the information at the linked page(s).

willdurand: I thought we could do it but `DOMPurify` apparently removes the `_target` attribute 😞

PR #5712, opened 2018-07-24
Install button reloaded (AMO and Disco Pane)

### LINKED ISSUE #3954: Re-implement mozAddonManager features in the install button

### LINKED ISSUE #5325: Replace about:addons (disco pane) toggle switches with Add to Firefox buttons
We want the experience in about:addons to be consistent with the experience on addons.mozilla.org. Right now, the about:addons disco pane experience uses a toggle switch, and addons.mozilla.org uses a button.

Wait to implement the button in disco pane, until the button is updated in addons.mozilla.org.

willdurand:
Fix #3954 
Fix #5325 

---

This PR:

- rewrites the `InstallButton` component into a `NewInstallButton` component to use the `mozAddonManager`:
  - install add-ons
  - uninstall add-ons
  - re-enable add-ons
- use the new `NewInstallButton` on AMO in -dev and locally (feature flag)
- use the new `NewInstallButton` on Disco Pane in -dev and locally (feature flag)

It works without the `mozAddonManager`.

### Disco Pane

Disabled state because status = UNKNOWN or not compatible:

Enabled state (in patched FF):

One theme installed:

Disabled state:

screen shot 2018-07-25 at 18 36 41

### AMO

Disabled state because status = UNKNOWN or not compatible:

Enabled state:

Installed state:

Disabled state:

screen shot 2018-07-25 at 18 07 01


willdurand:
Should be ready for a first look. I think the only feature that was not present before and that is still needed is to install+enable static themes at the same time on Disco Pane, but there are changes in another PR related to that so I'll do it as a follow-up I guess.

willdurand:
I also think that we should preload the animation because the svg might not appear on time. That's what the Pocket people did, but they did not use React or a component... so I am not sure how to do that in a nice manner (yet).

muffinresearch:
> Should be ready for a first look. I think the only feature that was not present before and that is still needed is to install+enable static themes at the same time on Disco Pane, but there are changes in another PR related to that so I'll do it as a follow-up I guess.

@willdurand Bear in mind that static themes are going to be enabled for upload this week (2nd). If this code lands it needs to handle the static theme enabling case before it hits production. It will probably make sense to land this after the tag tomorrow to provide more time and maximise testing.

willdurand:
> @willdurand Bear in mind that static themes are going to be enabled for upload this week (2nd). If this code lands it needs to handle the static theme enabling case before it hits production. It will probably make sense to land this after the tag tomorrow to provide more time and maximise testing.

@muffinresearch yep, it mostly works with static themes. Enabling works except when you click the preview image in disco pane (which is not working currently anyway). I also double-checked the potential double-enabling issue, and this is handled (the button was flickering in any case).

I think this week's release is packed though (with static themes + RR v4) so I am fine with merging this patch after tomorrow's tag.

kumar303:
Would it be possible to break out the patch for https://github.com/mozilla/addons-frontend/issues/3954 into a separate PR? It should be as easy as copying and pasting `Addon/index.js`, `InstallButton/index.js`, `TestAddon.js`, and `TestInstallButton.js` into a new branch based on master (I think). This would make it easier to review and also safer to land since it would only affect AMO. We could follow up with the Discopane patches.

willdurand:
> Would it be possible to break out the patch for #3954 into a separate PR?

Sure, I can break it down. Sorry, I should have done that earlier (the goal is to enable this new button on both AMO and Disco Pane at the same time, though).

willdurand:
mmh now that I have changed this PR, I remember why I did not split it earlier. The `InstallButton` is used on both AMO and Disco apps. Fixing that would probably add more changes than supporting the `InstallButton` on the Disco Pane.

willdurand:
I re-added the changes to support the Disco Pane (last commit).

willdurand:
rebased

src/core/components/InstallButton/index.js
```diff
@@ -194,195 +185,148 @@ export class InstallButtonBase extends React.Component {
...
 
-    // OpenSearch plugins display their own prompt so using the "Add to
-    // Firefox" button regardless on mozAddonManager support is a better UX.
-    const useButton =
-      (hasAddonManager !== undefined && !hasAddonManager) ||
-      addon.type === ADDON_TYPE_OPENSEARCH ||
-      this.props.useButton;
-    let button;
-
-    const { compatible } = getClientCompatibility({
```
kumar303: Is this not needed anymore? I don't see it. Previously, I think this was used to gray out the buttons in certain scenarios.

src/core/components/InstallButton/index.js
```diff
@@ -194,195 +185,148 @@ export class InstallButtonBase extends React.Component {
...
 
-    if (addon.type === ADDON_TYPE_THEME) {
-      button = (
-        <Button
-          buttonType="action"
-          className={buttonClass}
-          disabled={buttonIsDisabled}
-          data-browsertheme={JSON.stringify(getThemeData(addon))}
-          onClick={this.installTheme}
-          puffy
```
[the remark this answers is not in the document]
willdurand: Which spec? because the spec in the issue does not have the same dimensions obviously.

src/core/components/InstallButton/index.js
```diff
@@ -194,195 +185,148 @@ export class InstallButtonBase extends React.Component {
...
-          href={installURL}
-          prependClientApp={false}
-          prependLang={false}
-          puffy
-        >
-          <Icon name="plus" />
-          {buttonText}
-        </Button>
-      );
+    const buttonIsDisabled = disabled === false ? status === UNKNOWN : disabled;
```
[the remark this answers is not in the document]
willdurand: How and why would one want to use the mozAddonManager? If one wants to add add-ons, requiring mozAddonManager makes sense, no? Isn't it the new/recommended API? 

Which fallback should we use otherwise? The very old InstallTrigger? I checked and mozAddonManager is available in all supported FF versions.

src/core/components/InstallButton/index.js
```diff
@@ -194,195 +185,148 @@ export class InstallButtonBase extends React.Component {
...
-          type={this.props.type}
-          uninstall={this.props.uninstall}
-          updateURL={this.props.updateURL}
-          version={this.props.version}
-        />
-        {button}
+      <div className={makeClassName('InstallButton', className)}>
+        {this.showLoadingAnimation() ? (
+          <div className="InstallButton-loading">
+            <AnimatedIcon alt={this.getButtonText()} name="loading" />
```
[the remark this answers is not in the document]
willdurand: We might have other animated icons in the future, given that we are about to introduce this one and UX likes the FF animated icons. Of course we could merge this component into InstallButton right now, and extract the AnimatedIcon later.

src/core/components/InstallButton/index.js
willdurand: The parent component does that in both cases already. There is an issue about too many compatibility checks.

src/core/components/InstallButton/index.js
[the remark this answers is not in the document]
willdurand: i think the button still contains the link to the theme, so a user could probably download the theme and install it manually, no? Could it be a valid fallback?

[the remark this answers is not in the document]
willdurand: In addition, currently AMO does not use mozAddonManager at all IIRC, so that's why -dev shows an enabled button. What would be the result with the disco app though?

src/disco/components/Addon/index.js
```diff
@@ -226,13 +227,13 @@ export class AddonBase extends React.Component {
   };
 
   clickHeadingLink = (e) => {
-    const { type, name, _tracking } = this.props;
+    const { addon, _tracking } = this.props;
```
rebmullin: is this still needed?

tests/unit/core/components/TestInstallButton.js
```diff
@@ -82,146 +105,113 @@ describe(__filename, () => {
...
-    expect(switchComponent.props()).toMatchObject(addon.themeData);
+    const icon = button.find(Icon);
+    expect(icon).toHaveLength(1);
+    expect(icon).toHaveProp('name', 'plus');
   });
 
-  it('renders a theme button when mozAddonManager is not available', () => {
-    const addon = createInternalAddon(fakeTheme);
-    const root = render({ hasAddonManager: false, addon });
+  it('renders a button for themes', () => {
```
rebmullin: nit: above you have renders a 'Button' for themes..

also, I am wondering if we should have one for static theme too as this is only specific to lightweight theme I believe

willdurand:
@kumar303 I am going to feature-flag this feature, using a temporary `InstallButton` component until we are happy with everything (fallback, edge cases, etc.). I'll start with the provided features here + make sure we can have a link if `mozAddonManager` is not available.

tests/unit/core/components/TestInstallButton.js
```diff
@@ -82,146 +105,113 @@ describe(__filename, () => {
...
-    const root = render({
-      addon: createInternalAddon(
-        createFakeAddon({
-          type: ADDON_TYPE_EXTENSION,
-          files: [{ platform: OS_ALL, url: installURL }],
-        }),
-      ),
-      hasAddonManager: false,
-    });
+  it('calls the `installTheme` helper to install a theme', () => {
```
rebmullin: should we set anything up for static theme too?

rebmullin: never mind...I see this below : )

tests/unit/core/components/TestInstallButton.js
```diff
@@ -541,121 +456,149 @@ describe(__filename, () => {
...
+    expect(root.find(AnimatedIcon)).toHaveLength(1);
+  });
 
-    it('does not find a file hash without a current version', () => {
-      const addon = createInternalAddon(
-        createFakeAddon({
-          current_version: undefined,
-        }),
-      );
+  it('sets a `alt` prop to the AnimatedIcon when status is DOWNLOADING', () => {
```
rebmullin: nit: an `alt` ?

src/core/components/InstallButton/index.js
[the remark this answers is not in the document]
rebmullin: I have the same question here..is it intended to remove puffy for the install versions?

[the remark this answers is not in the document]
rebmullin: at least in AMO..

willdurand:
Feature flag ✅

I'll address the comments left so far and will ping you for new reviews. It is a huge diff now because I copied the files, which avoids to impact production code and will ease the full switch later (by `git mv`'ing the `NewInstallButton` into the `InstallButton`).

src/disco/components/Addon/index.js
willdurand: why not? it is needed for the tracking event.

willdurand:
> I noticed for static themes now it shows a GA "Started install" & "Install" 2x. I know we have the special case for static themes. This might be fine but just want to make sure this is okay and not actually sending dupe info..

Good catch! Thanks!

src/disco/components/Addon/index.js
rebmullin: ok was just verifying..I wasn't sure if I ever saw this called

src/disco/components/Addon/index.js
```diff
@@ -299,13 +309,27 @@ export class AddonBase extends React.Component {
...
+              addon={addon}
+              className="Addon-install-button"
+              defaultInstallSource={defaultInstallSource}
+              disabled={!compatible}
+              enable={enable}
+              install={install}
+              installTheme={installTheme}
+              puffy={false}
+              status={status || UNKNOWN}
+              uninstall={uninstall}
```
rebmullin: I think we need to pass in `hasAddonManager` (true) as prop here, right? Right now I don't think it works without this..

willdurand: oh heh, good catch!

rebmullin: I am wondering if that should also be part of the check of using the New Button..? Or what happens if it doesn't have that? does it fallback to the trigger method that was used before? or does it just do nothing?

src/core/components/NewInstallButton/index.js
```diff
@@ -0,0 +1,336 @@
...
+  installExtension = async (event: SyntheticEvent<HTMLAnchorElement>) => {
+    const { addon, enable, install } = this.props;
+
+    event.preventDefault();
+    event.stopPropagation();
+
+    await install();
+
+    if (addon.type === ADDON_TYPE_STATIC_THEME) {
+      await enable();
```
rebmullin: I think we want to do something similar with the old set up too (in case this isn't turned on right away). Right now it only installs without enabling..

willdurand: I suppose this should be done in a separate PR then.

src/disco/components/Addon/index.js
rebmullin: ohhh! I see...it's just tracking the click : )

src/disco/components/Addon/index.js
willdurand: fixed!

rebmullin:
hey @willdurand, just verifying the change of button size for mobile. Before it looked something like this:

Alt text

src/core/components/NewInstallButton/index.js
rebmullin: sure, we have the header click too anyway. It could be a part of that work..

src/disco/components/Addon/index.js
rebmullin: yep, I see 👍

src/amo/components/Addon/index.js
```diff
@@ -514,8 +526,22 @@ export class AddonBase extends React.Component {
...
-                {addon &&
-                  isFireFox && (
+                {showInstallButton &&
+                  config.get('enableNewInstallButton') && (
+                    <NewInstallButton
+                      addon={addon}
+                      defaultInstallSource={defaultInstallSource}
+                      disabled={!isCompatible}
+                      enable={enable}
+                      hasAddonManager={hasAddonManager}
```
rebmullin: this should always be true pretty much right? is there a possibility this could be false? If there is that case, I am wondering if we should wrap this button with this conditional check? b/c if it's it false, it won't work anyway, right?

tests/unit/core/components/TestNewInstallButton.js
```diff
@@ -0,0 +1,547 @@
...
+  it('configures a different install method for search providers', () => {
+    const root = renderOpenSearch();
+
+    expect(root.find(Button)).toHaveProp(
+      'onClick',
+      root.instance().installOpenSearch,
+    );
+  });
+
+  it('calls the `installTheme` helper to install a theme', () => {
```
rebmullin: maybe you should clarify here '..a lightweight theme' ?

src/amo/components/Addon/index.js
willdurand: It can be `false`, see: https://github.com/mozilla/addons-frontend/issues/3954#issuecomment-410210610. It works without the manager, but you don't have a status and therefore it is almost like the current install button.

rebmullin: sorry question this was meant for disco where I saw I nothing happening (it didn't install).

rebmullin: basically on disco - what I see is that the status shows 'uninstalled' (vs unknown) which means the button doesn't get disabled and this isn't covered in the button clicks options. let me know if you see something different tho..

willdurand: I am not sure I understand.

kumar303:
In an earlier patch, I commented about how install buttons are now disabled without `mozAddonManager`. You responded with this question:

> How and why would one want to not use the mozAddonManager? If one wants to add add-ons, requiring mozAddonManager makes sense, no? Isn't it the new/recommended API?

The main case for using the site without `mozAddonManager` is local development since `localhost:3000` does not have access to `mozAddonManager`. I'd like to preserve the way this currently works, if possible. Even if you can't fully do an install, it is helpful to see an enabled button when the add-on is enabled. If these states no longer make sense then we can revisit.

src/core/components/InstallButton/index.js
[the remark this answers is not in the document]
kumar303: by specs I meant the Sketch file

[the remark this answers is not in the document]
kumar303: ...but maybe the Sketch file is out of sync with the issue?

kumar303:
I commented before trying out the new patch. Installing an add-on/theme on `localhost:3000` without `mozAddonManager` now works again 😎 Thanks.

src/disco/components/Addon/index.js
```diff
@@ -226,25 +231,32 @@ export class AddonBase extends React.Component {
   };
 
   clickHeadingLink = (e) => {
-    const { type, name, _tracking } = this.props;
+    const { addon, _tracking } = this.props;
 
     if (e.target.nodeName.toLowerCase() === 'a') {
       _tracking.sendEvent({
-        action: getAddonTypeForTracking(type),
+        action: getAddonTypeForTracking(addon.type),
```
kumar303: 🔥 I hope this is correct 😀 I'm sure it is. The old reliance on spreads was absolutely bonkers.

src/core/components/NewInstallButton/index.js
```diff
@@ -0,0 +1,336 @@
...
+
+type Props = {|
+  addon: AddonType,
+  className?: string,
+  defaultInstallSource: string,
+  disabled: boolean,
+  location: ReactRouterLocationType,
+  puffy: boolean,
+  status: string,
+  // From `withInstallHelpers()`, see: `src/core/installAddon.js`.
```
kumar303: Is this comment correct? I see all props below this line getting set explicitly in `amo/Addon` and `disco/Addon`. I still see them set in `core/installAddon` as well so I'm a bit confused TBH.

tests/unit/core/components/TestNewInstallButton.js
```diff
@@ -0,0 +1,547 @@
+/* global Node */
```
kumar303: we have this in `.eslintrc` already

src/amo/components/Addon/index.js
rebmullin: sorry for not being clear.  Basically tho if you hasMoz... is false, you can't install anything on disco. This seems like something we should fix, right? I thought initially the buttons might be disabled in this case but they are actually still active and then it doesn't hit any case here so clicking the button doesn't do anything:

```javascript
if (!buttonIsDisabled) {
      if ([ENABLED, INSTALLED].includes(status)) {
        buttonProps.buttonType = 'neutral';
        buttonProps.onClick = this.uninstallAddon;
        buttonProps.className = makeClassName(
          buttonProps.className,
          'NewInstallButton-button--uninstall',
        );
      } else if (status === DISABLED) {
        buttonProps.buttonType = 'neutral';
        buttonProps.onClick = this.enableAddon;
        buttonProps.className = makeClassName(
          buttonProps.className,
          'NewInstallButton-button--enable',
        );
      } else if (addon.type === ADDON_TYPE_THEME) {
        buttonProps['data-browsertheme'] = JSON.stringify(getThemeData(addon));
        buttonProps.onClick = this.installTheme;
      } else if (hasAddonManager) {
        buttonProps.onClick =
          addon.type === ADDON_TYPE_OPENSEARCH
            ? this.installOpenSearch
            : this.installExtension;
      } else if (status === UNINSTALLED) {
        buttonProps.onClick = this.installExtension;
      }
    }
```

does this help clarify?

src/core/components/InstallButton/index.js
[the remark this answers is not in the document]
willdurand: Who knows? I configured the button to stay the same on AMO and not "puffy" on Disco (which seems to be the spec).

src/amo/components/Addon/index.js
willdurand: for me it does work, I can install themes and extensions, without the `mozAddonManager`. I have fixed this in the meantime.

rebmullin: oh? ok maybe it's just an issue on my end then? I have the ff build but it actually might be a bit outdated and related to that:/

willdurand: You have to make sure all the config prefs are correctly set: https://addons-frontend.readthedocs.io/en/latest/moz-addon-manager/

willdurand:
> @muffinresearch knows some proxy tricks to simulate how anti-virus software hijacks root certs so we can test it out. It's an edge case, for sure.

Yes, we discussed it. I am going to file an issue to double-check that!

willdurand:
🎉

PR #5717, opened 2018-07-24
I5707 adding all addon types to action list for GA

### LINKED ISSUE #5707: Look into GA install type to see why there is an invalid type appearing

In GA, there used to be just an extension or theme type but now an invalid category is coming up (on prod). 

This has something to do with the action type. We should look into these changes pertaining to this

rebmullin:
fixes #5707 

So it looks like we using extension tracking for install as a catchall if it wasn’t a lightweight theme or open search type.  

Earlier we were not using `getAddonTypeForTracking` (previously called `getAction`) for install, only uninstalls it seems. So on a side but related note I think you should probably see invalid coming up on uninstalls..

Anyway, I’ve added all the missing types so in theory we shouldn’t see invalid anymore if it’s one of these types. Just a note, I am still tracking these with the extension tracking type - is that what we want? Or should each addon type have their very own tracking type?

These are currently the tracking types that are set up:
```javascript
// Tracking add-on types
export const TRACKING_TYPE_EXTENSION = 'addon';
export const TRACKING_TYPE_STATIC_THEME = ADDON_TYPE_STATIC_THEME;
export const TRACKING_TYPE_THEME = 'theme';
export const TRACKING_TYPE_INVALID = 'invalid';
```

Here are some example screenshots for a dictionary type (notice the GA info in console at the bottom of each screenshot)
BEFORE: 
Alt text

AFTER:
Alt text

// cc @muffinresearch @willdurand

### DIFF
--- a/src/core/tracking.js
+++ b/src/core/tracking.js
@@ -170,18 +170,12 @@ export class Tracking {
 }
 
 export function getAddonTypeForTracking(type) {
-  const extensionTrackingTypes = [
-    ADDON_TYPE_DICT,
-    ADDON_TYPE_EXTENSION,
-    ADDON_TYPE_LANG,
-    ADDON_TYPE_OPENSEARCH,
-  ];
-
-  const index = extensionTrackingTypes.indexOf(type);
-
   return (
     {
-      [extensionTrackingTypes[index]]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_DICT]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_EXTENSION]: TRACKING_TYPE_EXTENSION,
+      [ADDON_TYPE_LANG]: TRACKING_TYPE_EXTENSION,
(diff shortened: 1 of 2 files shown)


src/core/tracking.js
```diff
@@ -168,10 +170,18 @@ export class Tracking {
...
+    ADDON_TYPE_OPENSEARCH,
+  ];
+
+  const index = extensionTrackingTypes.indexOf(type);
+
   return (
     {
-      [ADDON_TYPE_EXTENSION]: TRACKING_TYPE_EXTENSION,
-      [ADDON_TYPE_OPENSEARCH]: TRACKING_TYPE_EXTENSION,
+      [extensionTrackingTypes[index]]: TRACKING_TYPE_EXTENSION,
```
willdurand: Could we simply add all the constants in this map? I find confusing the use of `index`. I think it is fine to list all the constants:

```
{
  ...
  [ADDON_TYPE_DICT]: TRACKING_TYPE_EXTENSION,
  [ADDON_TYPE_EXTENSION]: TRACKING_TYPE_EXTENSION,
  ...
}
```

PR #5719, opened 2018-07-24
I5703 adjusting styles to minimize big icon while waiting for addon

### LINKED ISSUE #5703: A glitch appears when opening a theme listing page
STR:

1. Load AMO -dev
2. Select a theme to view its listing page

Actual result:
A glitch appears (a zoomed in default icon) right before the theme details page is loaded

Expected result:
There are no display issues when visiting a theme details page

Notes:
- reproduced on AMO dev with FF61, Win10x64 and Android 8.0
- the issue is also visible on stage but a little bit less pronounced 

theme icon glitch

rebmullin:
fixes #5703 

I did not see this issue consistently and I am not sure if this will be the end all solution but it does  seem to minimize the icon size at least. I believe this icon should be appearing as a placeholder while the we are waiting for the addon to load (right?). So this just gives a max height to make it look a little nicer. Note that I am going by lwt's height since it is a little bigger and for now since it's it should I think it should work for both.

### DIFF
--- a/src/core/css/inc/vars.scss
+++ b/src/core/css/inc/vars.scss
@@ -48,3 +48,6 @@ $max-modal-width: 768px;
 // Theme image dimensions
 $theme-height-default: 92px;
 $theme-width-default: 680px;
+
+// Lightweight images are slightly different
+$theme-height-legacy: 100px;
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -116,6 +116,7 @@
 .Addon-theme-header {
   border-radius: $border-radius-default;
   height: auto;
+  max-height: $theme-height-legacy;
   max-width: $theme-width-default;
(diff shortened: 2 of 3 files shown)


src/core/css/inc/vars.scss
```diff
@@ -48,3 +48,6 @@ $max-modal-width: 768px;
 // Theme image dimensions
 $theme-height-default: 92px;
 $theme-width-default: 680px;
+
+// Lightweight images are slightly different
+$theme-height-legacy: 100px;
```
willdurand: We can't use `$theme-height-default` because legacy themes have a height of `100px`? I don't understand the comment above, so I am guessing :p

rebmullin: oops :) Sorry my comment wasn't clear but yes, that's right. I am using the `100px` bc lightweight themes still have this.

PR #5730, opened 2018-07-25
I5555 overlay convert setstate to set ui state

### LINKED ISSUE #5555: Convert setState() in Overlay to setUIState()
The [Overlay](https://github.com/mozilla/addons-frontend/blob/54c4a4ddbfaea24c27620bead2254b183ba8bc61/src/ui/components/Overlay/index.js) component can be converted to using [`setUIState()`](https://github.com/mozilla/addons-frontend/issues/5343).

It's using `setState()` to show/hide the overlay content.

**For QA**: please make sure the overlay on the user profile screen still works as expected.

rebmullin:
fixes #5555 (wip)

I had few follow up questions maybe before the official review:

Should I use update `Overlay` to use flow or is this okay how I have it?

I’ve updated the tests in the `TestOverlay` to use enzyme - that’s ideally what we want, right?  

I see that `TestOverlayCard` is using `renderIntoDocument` too and since updating `TestOverlay` actually took a bit of time for me to change I wanted to make sure before I did any more work that this is something we want.

Clicking the background in the overlay, as far as I can tell, doesn’t work :/ 
Should it? I checked dev and it didn't work for me either. On stage, it seems to work if I click in the header..? :/

### DIFF
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -220,7 +220,7 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
       linkEnd: '</a>',
     });
 
-    const overlayClassName = 'AddonReview';
+    const overlayClassName = 'AddonReview-overlay';
 
     return (
       <OverlayCard
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -220,11 +220,14 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
       linkEnd: '</a>',
     });
 
+    const overlayClassName = 'AddonReview';
+
     return (
       <OverlayCard
         visibleOnLoad
         onEscapeOverlay={this.props.onEscapeOverlay}
-        className="AddonReview"
+        className={overlayClassName}
+        id={overlayClassName}
       >
         <h2 className="AddonReview-header">{i18n.gettext('Write a review')}</h2>
         {/* eslint-disable react/no-danger */}
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -452,6 +452,8 @@ export class UserProfileEditBase extends React.Component<Props, State> {
 
     const userProfileURL = `/user/${username}/`;
 
+    const overlayClassName = 'UserProfileEdit-deletion-modal';
+
     return (
       <div className="UserProfileEdit">
         {user && (
@@ -758,7 +760,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
 
         {this.state.showProfileDeletionModal && (
           <OverlayCard
-            className="UserProfileEdit-deletion-modal"
+            className={overlayClassName}
             header={
               isEditingCurrentUser
                 ? i18n.gettext(
@@ -768,6 +770,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                     `IMPORTANT: Deleting this Firefox Add-ons profile is irreversible.`,
                   )
             }
+            id={overlayClassName}
             onEscapeOverlay={this.onCancelProfileDeletion}
             visibleOnLoad
           >
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -221,8 +221,10 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     this.setState({ showProfileDeletionModal: true });
   };
 
-  onCancelProfileDeletion = (e: SyntheticEvent<HTMLButtonElement>) => {
-    e.preventDefault();
+  onCancelProfileDeletion = (event?: SyntheticEvent<any>) => {
+    if (event) {
+      event.preventDefault();
+    }
 
     this.setState({ showProfileDeletionModal: false });
   };
--- a/src/amo/components/UserProfileEdit/index.js
+++ b/src/amo/components/UserProfileEdit/index.js
@@ -221,10 +221,8 @@ export class UserProfileEditBase extends React.Component<Props, State> {
     this.setState({ showProfileDeletionModal: true });
   };
 
-  onCancelProfileDeletion = (event?: SyntheticEvent<any>) => {
-    if (event) {
-      event.preventDefault();
-    }
+  onCancelProfileDeletion = (e: SyntheticEvent<HTMLButtonElement>) => {
+    e.preventDefault();
 
     this.setState({ showProfileDeletionModal: false });
   };
@@ -773,7 +771,6 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                   )
             }
             id={overlayClassName}
-            onEscapeOverlay={this.onCancelProfileDeletion}
             visibleOnLoad
           >
             <p>
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -8,26 +8,12 @@ import withUIState from 'core/withUIState';
 
 import './styles.scss';
 
-// TODO: look into this as I don't see this working on the frontend
-// even in other envs.
-export const onClickBackground = (
-  event: SyntheticEvent<HTMLButtonElement>,
-  // eslint-disable-next-line no-use-before-define
-  _this: OverlayBase,
-) => {
-  if (_this.props.onEscapeOverlay) {
-    _this.props.onEscapeOverlay();
-  }
-  _this.hide();
-};
-
 type Props = {|
   className?: string,
   children: React.Element<any>,
   id: string,
-  onEscapeOverlay?: Function,
+  onEscapeOverlay?: () => void,
   visibleOnLoad?: boolean,
-  _onClickBackground: Function,
 |};
 
 type UIStateType = {|
@@ -45,7 +31,6 @@ type InternalProps = {|
 export class OverlayBase extends React.Component<InternalProps> {
   static defaultProps = {
     visibleOnLoad: false,
-    _onClickBackground: onClickBackground,
   };
 
   componentWillReceiveProps(nextProps: InternalProps) {
@@ -64,13 +49,14 @@ export class OverlayBase extends React.Component<InternalProps> {
     this.props.setUIState({ visible: false });
   }
 
-  show() {
-    this.props.setUIState({ visible: true });
-  }
-
-  toggle() {
-    this.props.setUIState({ visible: !this.props.uiState.visible });
-  }
+  // TODO: look into this as I don't see this working on the frontend
+  // even in other envs.
+  onClickBackground = () => {
+    if (this.props.onEscapeOverlay) {
+      this.props.onEscapeOverlay();
+    }
+    this.hide();
+  };
 
   render() {
     const { children, className, id, uiState } = this.props;
@@ -86,7 +72,7 @@ export class OverlayBase extends React.Component<InternalProps> {
       >
         <div
           className="Overlay-background"
-          onClick={(e) => this.props._onClickBackground(e, this)}
+          onClick={this.onClickBackground}
           role="presentation"
         />
         <div className="Overlay-contents">{children}</div>
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -81,11 +81,11 @@ export class OverlayBase extends React.Component<InternalProps> {
   }
 }
 
-export const extractId = (ownProps: InternalProps) => {
+export const extractId = (ownProps: Props) => {
   return ownProps.id;
 };
 
-const Overlay: React.ComponentType<InternalProps> = compose(
+const Overlay: React.ComponentType<Props> = compose(
   withUIState({
     fileName: __filename,
     extractId,
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -34,10 +34,8 @@ export class OverlayBase extends React.Component<InternalProps> {
   };
 
   componentDidMount() {
-    const { uiState, visibleOnLoad } = this.props;
-    if (visibleOnLoad !== uiState.visible) {
-      this.props.setUIState({ visible: visibleOnLoad });
-    }
+    const { visibleOnLoad } = this.props;
+    this.props.setUIState({ visible: visibleOnLoad });
   }
 
   componentWillReceiveProps(nextProps: InternalProps) {
@@ -50,10 +48,6 @@ export class OverlayBase extends React.Component<InternalProps> {
     ) {
       this.props.setUIState({ visible: visibleOnLoadNew });
     }
-  }
-
-  componentWillUnmount() {
-    this.hide();
   }
 
   onClickBackground = (event: SyntheticEvent<any>) => {
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -1,35 +1,47 @@
+/* @flow */
 import invariant from 'invariant';
 import makeClassName from 'classnames';
 import * as React from 'react';
-import PropTypes from 'prop-types';
 import { compose } from 'redux';
 
 import withUIState from 'core/withUIState';
 
 import './styles.scss';
 
-const initialUIState = { visible: false };
-
-// TODO: I don't see this working on the frontend
+// TODO: look into this as I don't see this working on the frontend
 // even in other envs.
-export const onClickBackground = (event, _this) => {
+export const onClickBackground = (
+  event: SyntheticEvent<HTMLButtonElement>,
+  _this: OverlayBase,
+) => {
   if (_this.props.onEscapeOverlay) {
     _this.props.onEscapeOverlay();
   }
   _this.hide();
 };
 
-export class OverlayBase extends React.Component {
-  static propTypes = {
-    children: PropTypes.node,
-    className: PropTypes.string,
-    id: PropTypes.string.isRequired,
-    visibleOnLoad: PropTypes.bool,
-    uiState: PropTypes.object,
-    setUIState: PropTypes.func,
-    _onClickBackground: PropTypes.func,
-  };
+type Props = {|
+  className?: string,
+  children: React.Element<any> | string,
+  id: string,
+  onEscapeOverlay: Function,
+  visibleOnLoad?: boolean,
+  _onClickBackground: Function,
+|};
 
+type UIStateType = {|
+  visible: boolean,
+|};
+
+const initialUIState: UIStateType = { visible: false };
+
+type InternalProps = {|
+  ...Props,
+  setUIState: ($Shape<UIStateType>) => void,
+  uiState: UIStateType,
+|};
+
+export class OverlayBase extends React.Component<InternalProps> {
   static defaultProps = {
     visibleOnLoad: false,
     uiState: initialUIState,
@@ -37,7 +49,7 @@ export class OverlayBase extends React.Component {
     _onClickBackground: onClickBackground,
   };
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps(nextProps: InternalProps) {
     const { uiState: uiStateOld } = this.props;
 
     if (
@@ -76,20 +88,13 @@ export class OverlayBase extends React.Component {
           onClick={(e) => this.props._onClickBackground(e, this)}
           role="presentation"
         />
-        <div
-          className="Overlay-contents"
-          ref={(ref) => {
-            this.overlayContents = ref;
-          }}
-        >
-          {children}
-        </div>
+        <div className="Overlay-contents">{children}</div>
       </div>
     );
   }
 }
 
-export const extractId = (ownProps) => {
+export const extractId = (ownProps: InternalProps) => {
   return ownProps.id;
 };
 
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -33,29 +33,40 @@ export class OverlayBase extends React.Component<InternalProps> {
     visibleOnLoad: false,
   };
 
+  componentDidMount() {
+    const { uiState, visibleOnLoad } = this.props;
+    if (visibleOnLoad !== uiState.visible) {
+      this.props.setUIState({ visible: visibleOnLoad });
+    }
+  }
+
   componentWillReceiveProps(nextProps: InternalProps) {
-    const { uiState: uiStateOld } = this.props;
+    const { uiState } = this.props;
     const { visibleOnLoad: visibleOnLoadNew } = nextProps;
 
     if (
-      (visibleOnLoadNew && !uiStateOld.visible) ||
-      uiStateOld.visible !== nextProps.uiState.visible
+      visibleOnLoadNew !== undefined &&
+      visibleOnLoadNew !== uiState.visible
     ) {
       this.props.setUIState({ visible: visibleOnLoadNew });
     }
   }
 
-  hide() {
-    this.props.setUIState({ visible: false });
+  componentWillUnmount() {
+    this.hide();
   }
 
-  // TODO: look into this as I don't see this working on the frontend
-  // even in other envs.
-  onClickBackground = () => {
+  onClickBackground = (event: SyntheticEvent<any>) => {
+    event.preventDefault();
     if (this.props.onEscapeOverlay) {
       this.props.onEscapeOverlay();
     }
+
     this.hide();
+  };
+
+  hide = () => {
+    this.props.setUIState({ visible: false });
   };
 
   render() {
--- a/src/ui/components/Overlay/index.js
+++ b/src/ui/components/Overlay/index.js
@@ -12,6 +12,7 @@ import './styles.scss';
 // even in other envs.
 export const onClickBackground = (
   event: SyntheticEvent<HTMLButtonElement>,
+  // eslint-disable-next-line no-use-before-define
   _this: OverlayBase,
 ) => {
   if (_this.props.onEscapeOverlay) {
@@ -22,9 +23,9 @@ export const onClickBackground = (
 
 type Props = {|
   className?: string,
-  children: React.Element<any> | string,
+  children: React.Element<any>,
   id: string,
-  onEscapeOverlay: Function,
+  onEscapeOverlay?: Function,
   visibleOnLoad?: boolean,
   _onClickBackground: Function,
 |};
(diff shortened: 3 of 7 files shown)


src/ui/components/Overlay/index.js
```diff
@@ -1,32 +1,48 @@
...
+    uiState: PropTypes.object,
+    setUIState: PropTypes.func,
+    onClickBackground: PropTypes.func,
   };
 
   static defaultProps = {
     visibleOnLoad: false,
+    uiState: initialUIState,
+    setUIState: () => {},
+    onClickBackground: () => {},
```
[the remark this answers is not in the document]
rebmullin: Yep, thanks for catching that. I'll update this.

Regarding it not working, it's not working in dev or stage either. I'll look into this a bit more tho..

src/ui/components/Overlay/index.js
```diff
@@ -1,32 +1,48 @@
...
 import './styles.scss';
 
-export default class Overlay extends React.Component {
+const initialUIState = { visible: false };
+
+export class OverlayBase extends React.Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
+    id: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
rebmullin: We do want to make sure they are unique so I think `id` would be better here. It sounds like you reached the same conclusion :) right?

src/amo/components/AddonReview/index.js
```diff
@@ -220,11 +220,14 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
       linkEnd: '</a>',
     });
 
+    const overlayClassName = 'AddonReview';
```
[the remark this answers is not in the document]
rebmullin: I agree this is better. thanks :)

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,125 @@
...
+  };
 
   it('renders an Overlay', () => {
     const root = render();
-    expect(root.overlayContainer).toBeTruthy();
-    expect(root.overlayBackground).toBeTruthy();
-    expect(root.overlayContainer.tagName).toEqual('DIV');
-    expect(root.overlayContainer.className).toContain('Overlay');
+    expect(root).toHaveClassName('Overlay');
+    expect(root.find('.Overlay')).toHaveLength(1);
```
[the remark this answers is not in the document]
rebmullin: 🤦‍♂️  yep...

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,125 @@
...
+    applyUIStateChanges({ root, store });
+    expect(root).not.toHaveClassName('Overlay--visible');
+  });
 
-    root.toggle();
-    expect(root.overlayContainer.className).not.toContain('Overlay--visible');
+  describe('extractId', () => {
+    it('returns a unique ID provided by the ID prop', () => {
+      const id = 'custom-overlay';
+      expect(extractId(getProps({ id }))).toEqual(id);
```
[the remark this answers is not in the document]
rebmullin: hey @bobsilverberg, this came up here in [this PR](https://github.com/mozilla/addons-frontend/pull/5642#discussion_r203423927). So i followed the same here. Given that explanation, is this okay here?

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,125 @@
...
+    applyUIStateChanges({ root, store });
+
+    // Applying this 2 times here to mimic componentWillReceiveProps
+    // in that it see the uiState changes till the 2nd run through
+    applyUIStateChanges({ root, store });
+
+    expect(root.instance().props.uiState.visible).toEqual(true);
   });
 
   it('hides when you click the background', () => {
```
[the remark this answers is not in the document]
rebmullin: hmm yeh, I will add a check for the visible classname, thanks!

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,125 @@
...
+      uiState: { visible: false },
+    });
+
+    expect(root.instance().props.uiState.visible).toEqual(false);
+
+    applyUIStateChanges({ root, store });
+
+    // Applying this 2 times here to mimic componentWillReceiveProps
+    // in that it see the uiState changes till the 2nd run through
+    applyUIStateChanges({ root, store });
```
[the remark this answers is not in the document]
rebmullin: I'm actually not 100% but I believe we have to do this 2x here b/c the first time when it goes through componentWillReceiveProps, it gets changed but you don't see it till the change till the 2nd run through. There def might be a better to show this in the test. I will update the comment too a little bit..

src/ui/components/Overlay/index.js
```diff
@@ -35,31 +51,33 @@ export default class Overlay extends React.Component {
...
+          'Overlay--visible': uiState.visible,
         })}
         ref={(ref) => {
           this.overlayContainer = ref;
         }}
       >
         <div
-          onClick={this.onClickBackground}
+          onClick={this.props.onClickBackground}
           ref={(ref) => {
```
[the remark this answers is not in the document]
rebmullin: I am just curious why we are removing this? Is it ok to remove the one below too then - the overlayContents?

rebmullin:
> As with the upgrading to Flow, I don't think it's necessary to do this every time we touch a test file. Because you have not had to make any changes to TestOverlayCard I don't think we should bloat the patch with updating that one right now.

Regarding the TestOverlayCard, I actually do pass in a required `id` now so I would need to add this check, right? A bigger concern tho is that now I see the following when I run `yarn test-ci`:

 ● /Users/rmullin/Sites/www/addons-frontend/tests/unit/ui/components/TestOverlayCard.js › is toggled

```bash
    Invariant Violation: Could not find "store" in either the context or props of "Connect(WithUIState(OverlayBase))". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(WithUIState(OverlayBase))".

      11 | describe(__filename, () => {
      12 |   function render(props = {}) {
    > 13 |     return renderIntoDocument(<OverlayCard {...props} />);
         |            ^
      14 |   }
      15 | 
      16 |   it('renders an OverlayCard', () => {
```

I feel b/c this is connected to Overlay, I will have to update this anyway..? I'll look into to see if it's a quick update (till we can do a follow up PR to update this)

tests/unit/ui/components/TestOverlayCard.js
```diff
@@ -4,13 +4,33 @@ import {
...
+      id: 'OverlayCard',
+      store,
+      ...props,
+    };
+  };
+
   function render(props = {}) {
-    return renderIntoDocument(<OverlayCard {...props} />);
+    return findRenderedComponentWithType(
+      renderIntoDocument(
```
rebmullin: I left this in place for now as initially I was focused on just the Overlay. 

also a note here, on why I removed some of this tests here:

When I provide this Provider wrapper, the `overlayContainer` isn't available anymore. You will see it removed in this PR but even with this added back, this is coming up undefined.

I was running into this similar issue too with hide, show, toggle - but I didn't even see these used here (or in Overlay - altho I left in Overlay for now).

the Overlay tests seem to do some of this already (such as check for default set up).

let me know if this is an issue

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,141 @@
...
-    const root = render({ visibleOnLoad: true, onEscapeOverlay });
-    Simulate.click(root.overlayBackground);
+    const root = render({
+      visibleOnLoad: true,
+      onEscapeOverlay,
+    });
+    const btn = root.find('.Overlay-background');
+    btn.simulate('click', clickEvent);
+    const rootInstance = root.instance();
+    rootInstance.props._onClickBackground(clickEvent, rootInstance);
```
kumar303: Why do you need to invoke the method? Clicking on `'.Overlay-background'` should be enough -- was it not working in the test?

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,110 @@
...
-      this.setState({ visible: nextProps.visibleOnLoad });
+    if (
+      (visibleOnLoadNew && !uiStateOld.visible) ||
+      uiStateOld.visible !== nextProps.uiState.visible
+    ) {
+      this.props.setUIState({ visible: visibleOnLoadNew });
     }
   }
 
-  onClickBackground = () => {
```
kumar303: It looks like maybe you moved this to a prop just for testing? I am not sure you need it for testing (I explained why in the comments).

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,110 @@
...
-          this.overlayContainer = ref;
-        }}
       >
         <div
-          onClick={this.onClickBackground}
-          ref={(ref) => {
-            this.overlayBackground = ref;
-          }}
           className="Overlay-background"
+          onClick={(e) => this.props._onClickBackground(e, this)}
```
kumar303: You could use [`bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) instead but it looks like maybe you don't need it as a prop.

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,110 @@
...
+    _this.props.onEscapeOverlay();
+  }
+  _this.hide();
+};
+
+type Props = {|
+  className?: string,
+  children: React.Element<any>,
+  id: string,
+  onEscapeOverlay?: Function,
```
kumar303: If you make this `onEscapeOverlay?: () => void` then Flow will validate the function signature. It's more fine-grained than `Function`; it says it's a function that takes no arguments.

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,110 @@
...
       </div>
     );
   }
 }
+
+export const extractId = (ownProps: InternalProps) => {
+  return ownProps.id;
+};
+
+const Overlay: React.ComponentType<InternalProps> = compose(
```
kumar303: The type hint should be `React.ComponentType<Props>` to guarantee that callers pass in all required properties of `Props`, not `InternalProps`.

src/ui/components/OverlayCard/index.js
```diff
@@ -67,3 +60,17 @@ export default class OverlayCard extends React.Component {
     );
   }
 }
+
+export const extractId = (ownProps) => {
+  return ownProps.id;
+};
+
+const OverlayCard = compose(
+  withUIState({
```
kumar303: Unless I'm missing something, this component does not need to be wrapped in `withUIState`.

It may have been confusing because the old component was implemented in a way that doesn't make sense for a React component. It was saving a ref to its child `<Overlay>` so it could show/hide the overlay but as far as I recall, nothing ever used this feature within `OverlayCard`. The React way to do it is to pass along state changes via props which it does already through `visibleOnLoad`. I think that's all it needs to do.

src/ui/components/OverlayCard/index.js
```diff
@@ -47,6 +39,7 @@ export default class OverlayCard extends React.Component {
     return (
       <Overlay
         onEscapeOverlay={this.props.onEscapeOverlay}
+        id={id}
         visibleOnLoad={visibleOnLoad}
         ref={(ref) => {
```
kumar303: You can just remove all the refs, I think

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,141 @@
...
-    const root = render({ children: 'hello' });
-    expect(root.overlayContents).toBeTruthy();
-    expect(root.overlayContents.textContent).toContain('hello');
+    const html = '<div>overriding child div</div>';
+    const root = render({ children: html });
+    expect(
+      root
+        .children()
+        .at(1)
+        .props().children,
```
kumar303: Is it possible to use the [children](http://airbnb.io/enzyme/docs/api/ShallowWrapper/children.html) wrapper functions instead?

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,141 @@
...
 
   it('is visible when the `visibleOnLoad` prop is passed', () => {
-    const root = render({ visibleOnLoad: true });
-    expect(root.overlayContainer.className).toContain('Overlay--visible');
+    const { store } = dispatchClientMetadata();
+
+    const root = render({
+      store,
+      visibleOnLoad: true,
+      uiState: { visible: false },
```
kumar303: The `uiState` prop is meant for internal use (does it even work to pass in a value like this?). It's not helpful to pass this in a test. You should use [`setUIState()`](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/helpers.js#L670) but in this case the default is `visibleOnLoad: false` so I don't think you need to initialize the state.

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,141 @@
...
+    });
+
+    expect(root).not.toHaveClassName('Overlay--visible');
+
+    applyUIStateChanges({ root, store });
+
+    // Applying this 2 times here to mimic how uiState works in componentWillReceiveProps:
```
kumar303: This should be done in a separate test using `root.setProps()`. I think it will be easier, too.

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,141 @@
...
   it('is shown and hidden when `hide()` and `show()` are called', () => {
-    const root = render();
+    const { store } = dispatchClientMetadata();
+    const root = render({ store });
+
+    expect(root).not.toHaveClassName('Overlay--visible');
 
-    root.show();
-    expect(root.overlayContainer.className).toContain('Overlay--visible');
+    root.instance().show();
```
kumar303: I think you can get rid of the public `.show()`, `.hide()`, and `.toggle()` methods because, as far as I see, nothing is even using these methods. As I pointed out up above, this is a very un-React way to do things. It requires storing refs to components which we try not to do anymore, especially in production code. I think both `AddonReview` and `UserProfileEdit` are showing/hiding with the `visibleOnLoad` prop -- this should be enough.

kumar303:
> A bigger concern tho is that now I see the following when I run yarn test-ci:
> /Users/rmullin/Sites/www/addons-frontend/tests/unit/ui/components/TestOverlayCard.js › is toggled
>
>    Invariant Violation: Could not find "store" in either the context or props of "Connect(WithUIState(OverlayBase))"

This was because `withUIState()` introduces a Redux constraint. This error should go away when you remove `withUIState()` from `OverlayCard` (I don't think you need it there).

src/ui/components/Overlay/index.js
rebmullin: okay, I was looking at other components (as a guideline) and saw this in many places. I saw that Internal Props uses Props so I thought this made sense to me. I can def update I just wanted to let you know where I was coming from : )

src/ui/components/OverlayCard/index.js
rebmullin: I didn't initially do this or want to do this but I was getting complaints from flow. See my comment above: https://github.com/mozilla/addons-frontend/pull/5730#issuecomment-408176529.  

I'll take another look tho. Maybe there is a better way.

src/ui/components/OverlayCard/index.js
rebmullin: will do : )

tests/unit/ui/components/TestOverlay.js
rebmullin: yep, let me take another look

tests/unit/ui/components/TestOverlay.js
rebmullin: Right,  I wanted to do b/c I didn't even see it used. So cool, I'll remove this : ) thanks

src/ui/components/Overlay/index.js
rebmullin: hmm i don't think bind was working for me but maybe I  don't need it as you suggest. I'll take another look..

tests/unit/ui/components/TestOverlay.js
rebmullin: hmm it is actually working. I'll remove..

tests/unit/ui/components/TestOverlay.js
rebmullin: hmm ok let me take another look, thanks kumar

src/ui/components/OverlayCard/index.js
rebmullin: Actually if I do this, I will have to update the OverlayCard (since it's using these refs in its tests). I guess I was trying to avoid that in this PR at least..
I am still having that store issue with this file so I may have to anyway tho :/
I'll follow up : )

tests/unit/ui/components/TestOverlay.js
rebmullin: actually, we are calling the hide method in the clickbackground function so I'll leave that for now..

rebmullin:
> This was because withUIState() introduces a Redux store requirement. This error should go away when you remove withUIState() from OverlayCard (I don't think you need it there).

I was still seeing this when I tried to update the tests in OverlayCard to use shallow :/ 

For now, I've removed all the withUIState from this file tho and kept the renderIntoDocument wrapped with a Provider (to prevent this same message noted above) and it seems to be "okay"

src/ui/components/Overlay/index.js
rebmullin: oh! I see... this is just at the end of the file...

tests/unit/ui/components/TestOverlay.js
kumar303: Yes, it's OK for the component to call its own `.hide()` method internally.

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,96 @@
...
+|};
+
+export class OverlayBase extends React.Component<InternalProps> {
   static defaultProps = {
     visibleOnLoad: false,
   };
 
-  constructor(props) {
-    super(props);
-    this.state = { visible: props.visibleOnLoad };
```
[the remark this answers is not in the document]
kumar303: Could you also add a test to cover this?

src/amo/components/AddonReview/index.js
```diff
@@ -220,11 +220,14 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
       linkEnd: '</a>',
     });
 
+    const overlayClassName = 'AddonReview-overlay';
```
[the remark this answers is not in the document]
rebmullin: 🤦‍♂️  yep...thanks for catching this..

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,96 @@
...
+  componentWillReceiveProps(nextProps: InternalProps) {
+    const { uiState: uiStateOld } = this.props;
+    const { visibleOnLoad: visibleOnLoadNew } = nextProps;
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.visibleOnLoad !== undefined) {
-      this.setState({ visible: nextProps.visibleOnLoad });
+    if (
+      (visibleOnLoadNew && !uiStateOld.visible) ||
+      uiStateOld.visible !== nextProps.uiState.visible
```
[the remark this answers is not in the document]
rebmullin: It seems in order to get this to work, I need to reset the state in componentWillUnMount:

```javascript
 componentWillUnmount() {
    this.props.setUIState({ visible: false });
  }
```

otherwise, I can't always reopen it. Egs:

on the detail page, clicking on star (similar thing happens with profile page):
click 1 - it opens
click 2 - nothing happens
click 3 - it opens

on review page, clicking on "edit my review" link:
click 1 - it opens
click 2, etc - nothing happens

I do see the clicking on background working if i click near the top of the page(seems a little sporadic too)

**** Actually after testing this a bit more this (unmount noted above) only seems to work if you close via submitting. Using the background click does not unmount it so i'm still investigating this :/

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,96 @@
...
+    ) {
+      this.props.setUIState({ visible: visibleOnLoadNew });
     }
   }
 
+  hide() {
+    this.props.setUIState({ visible: false });
+  }
+
+  // TODO: look into this as I don't see this working on the frontend
```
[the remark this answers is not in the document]
rebmullin: thanks for helping with this @kumar303 : ) it's just that I am still seeing some issues (noted above about not being able to reopen after this first click). I thought I had a workaround with the unmount but then the clicking on background still is causing issues. I'll keep you posted..

src/ui/components/Overlay/index.js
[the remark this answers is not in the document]
rebmullin: Regarding your question about about `visibleOnLoad`..these 2 components that are set up to use it are always passing this in as true.  These components arent mounted until their respective state in their components show true. so I don't think it's ever changes value at least in these 2 cases. if that makes sense..

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,96 @@
...
+      id: 'Overlay',
+      className: 'Overlay',
+      store: dispatchClientMetadata().store,
+      ...props,
+    };
+  };
+
+  const render = ({ children, ...otherProps } = {}) => {
+    const html = '<div>a child div</div>';
+    return shallowUntilTarget(
```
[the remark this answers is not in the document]
rebmullin: Ok I'll update here. (I guess I need to update this in confirmButton too..? )

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,96 @@
...
+    const className = 'I-am-so-over-it';
+    const root = render({ className });
+    expect(root).toHaveClassName(className);
   });
 
   it('renders children', () => {
-    const root = render({ children: 'hello' });
-    expect(root.overlayContents).toBeTruthy();
-    expect(root.overlayContents.textContent).toContain('hello');
+    const html = '<div>overriding child div</div>';
```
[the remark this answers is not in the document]
rebmullin: when I take this out of a string, I see this:
(with this test case):
```javascript
 expect(
      root
        .find('.Overlay-contents')
        .children()
        .html(),
    ).toEqual(html);
```

shows me this:

```bash
Expected value to equal:
      <div>overriding child div</div>
    Received:
      "<div>overriding child div</div>"
```
is something like this better:
```javascript
 const text = 'overriding div text..';
    const html = <div>{text}</div>;
    const root = render({ visibleOnLoad: true, children: html });

    expect(
      root
        .find('.Overlay-contents')
        .children()
        .html(),
    ).toContain(text);
```

src/ui/components/Overlay/index.js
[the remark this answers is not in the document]
rebmullin: I think this is b/c the components that use this are relying on their own state detection to display/mount the component so this is always passed in as true at this point.

I think we have a test case which about cover this but I think I can tweak it a bit to fit a little better

src/amo/components/UserProfileEdit/index.js
```diff
@@ -768,6 +770,7 @@ export class UserProfileEditBase extends React.Component<Props, State> {
                     `IMPORTANT: Deleting this Firefox Add-ons profile is irreversible.`,
                   )
             }
+            id={overlayClassName}
             onEscapeOverlay={this.onCancelProfileDeletion}
```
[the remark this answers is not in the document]
rebmullin: I actually think we do need this (for now) anyway b/c this updates the state back in the parent ...at least it seems that way in the UserProfileEdit. I have to dig in a bit more to REview..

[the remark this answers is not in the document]
rebmullin: I *think* once we get this UIstate into these parents components, I should be able to reset their state in Overlay and then I can remove this onEscapeOverlay. Would that be okay to do that in a follow up PR?

kumar303:
https://github.com/mozilla/addons-frontend/issues/5764 just landed so you can try `resetOnUnmount=true` to see if it's helpful. I feel like maybe the `visibleOnLoad` prop should be enough, though. I'd have to look at a patch so maybe just get it working first :)

src/amo/components/UserProfileEdit/index.js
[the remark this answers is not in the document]
rebmullin: Yep, I ended up removing this in the UserProfileEdit...

src/ui/components/Overlay/index.js
```diff
@@ -1,80 +1,101 @@
...
+export const extractId = (ownProps: Props) => {
+  return ownProps.id;
+};
+
+const Overlay: React.ComponentType<Props> = compose(
+  withUIState({
+    fileName: __filename,
+    extractId,
+    initialState: initialUIState,
+  }),
```
rebmullin: I know it said before it was working without resetting this before but it seems to be okay with the default now 🤷‍♂️

tests/unit/ui/components/TestOverlay.js
```diff
@@ -1,73 +1,107 @@
...
+    const text = 'overriding div text..';
+    const html = <div>{text}</div>;
+    const root = render({ visibleOnLoad: true, children: html });
+
+    expect(
+      root
+        .find('.Overlay-contents')
+        .children()
+        .html(),
+    ).toContain(text);
```
rebmullin: I am still not sure about this. I believe I tried a few a things here. let me know if this works...

rebmullin:
@kumar303, thanks for all the feedback and helping out with this one.

I *think* it's all working now. There are 3 sections that I am testing/viewing this:

Star rating manager (seen on addon detail pg and on the reviews pg)
Edit my review (seen on the review pg)
Delete my profile (user profile pg)

You can do back-to-back clicks after closing via the buttons/submit or even clicking the background.

note: I still only saw the background working when you clicked above-ish the modal.

kumar303:
> I still only saw the background working when you clicked above-ish the modal.

That's a known issue, not something to worry about

kumar303:
You can disregard the CI failure; it will be fixed in https://github.com/mozilla/addons-frontend/issues/5819

tests/unit/ui/components/TestOverlay.js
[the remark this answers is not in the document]
rebmullin: ahh...I see : ) thanks

tests/unit/ui/components/TestOverlayCard.js
```diff
@@ -4,18 +4,38 @@ import {
...
+        </Provider>,
+      ),
+      OverlayCard,
+    );
   }
 
   it('renders an OverlayCard', () => {
     const root = render();
-    expect(root.overlayCard).toBeTruthy();
+    expect(root.props.className).toEqual('OverlayCard');
```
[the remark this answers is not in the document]
rebmullin: lol..ok sure : )

willdurand:
> You can disregard the CI failure; it will be fixed in #5819

Fixed :)

PR #5739, opened 2018-07-26
Change rating count to review count on add-on detail page

### LINKED ISSUE #5434: Change rating count to review count on add-on detail page
screen shot 2018-06-27 at 11 01 58 am

- Change the "Ratings" count on the detail page to a "Reviews" count.
- Make the "Reviews" text and accompanying number clickable, linking to the all reviews page.

bobsilverberg:
Fixes #5434 

Before:

screenshot 2018-07-26 09 14 34

After:

screenshot 2018-07-26 09 13 49

### DIFF
--- a/src/amo/components/AddonMeta/index.js
+++ b/src/amo/components/AddonMeta/index.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import { compose } from 'redux';
 
+import Link from 'amo/components/Link';
 import translate from 'core/i18n/translate';
 import type { AddonType } from 'core/types/addons';
 import MetadataCard from 'ui/components/MetadataCard';
@@ -50,13 +51,37 @@ export class AddonMetaBase extends React.Component<InternalProps> {
     let reviewTitle;
     if (!addon) {
       reviewCount = null;
-      reviewTitle = i18n.gettext('Ratings');
+      reviewTitle = i18n.gettext('Reviews');
     } else if (addonReviewCount) {
       reviewCount = i18n.formatNumber(addonReviewCount);
       reviewTitle = i18n.ngettext('Review', 'Reviews', addonReviewCount);
     } else {
       reviewTitle = i18n.gettext('No Reviews');
     }
+
+    const reviewsContent =
+      addon && reviewCount ? (
+        <Link
+          className="AddonMeta-reviews-content-link"
+          to={`/addon/${addon.slug}/reviews/`}
+        >
+          {reviewCount}
+        </Link>
+      ) : (
+        reviewCount
+      );
+
+    const reviewsTitle =
+      addon && reviewCount ? (
+        <Link
+          className="AddonMeta-reviews-title-link"
+          to={`/addon/${addon.slug}/reviews/`}
+        >
+          {reviewTitle}
+        </Link>
+      ) : (
+        reviewTitle
+      );
 
     return (
       <div className="AddonMeta">
@@ -69,8 +94,8 @@ export class AddonMetaBase extends React.Component<InternalProps> {
               title: userTitle,
             },
             {
-              content: reviewCount,
-              title: reviewTitle,
+              content: reviewsContent,
+              title: reviewsTitle,
             },
             {
               content: (
--- a/src/amo/components/AddonMeta/index.js
+++ b/src/amo/components/AddonMeta/index.js
@@ -59,29 +59,24 @@ export class AddonMetaBase extends React.Component<InternalProps> {
       reviewTitle = i18n.gettext('No Reviews');
     }
 
-    const reviewsContent =
-      addon && reviewCount ? (
-        <Link
-          className="AddonMeta-reviews-content-link"
-          to={`/addon/${addon.slug}/reviews/`}
-        >
-          {reviewCount}
-        </Link>
-      ) : (
-        reviewCount
-      );
+    const reviewsLink =
+      addon && reviewCount ? `/addon/${addon.slug}/reviews/` : null;
 
-    const reviewsTitle =
-      addon && reviewCount ? (
-        <Link
-          className="AddonMeta-reviews-title-link"
-          to={`/addon/${addon.slug}/reviews/`}
-        >
-          {reviewTitle}
-        </Link>
-      ) : (
-        reviewTitle
-      );
+    const reviewsContent = reviewsLink ? (
+      <Link className="AddonMeta-reviews-content-link" to={reviewsLink}>
+        {reviewCount}
+      </Link>
+    ) : (
+      reviewCount
+    );
+
+    const reviewsTitle = reviewsLink ? (
+      <Link className="AddonMeta-reviews-title-link" to={reviewsLink}>
+        {reviewTitle}
+      </Link>
+    ) : (
+      reviewTitle
+    );
 
     return (
       <div className="AddonMeta">
--- a/src/amo/components/AddonMeta/index.js
+++ b/src/amo/components/AddonMeta/index.js
@@ -30,6 +30,8 @@ export class AddonMetaBase extends React.Component<InternalProps> {
     const averageRating = addon && addon.ratings ? addon.ratings.average : null;
     const addonRatingCount =
       addon && addon.ratings ? addon.ratings.count : null;
+    const addonReviewCount =
+      addon && addon.ratings ? addon.ratings.text_count : null;
     const averageDailyUsers = addon ? addon.average_daily_users : null;
 
     let userCount = '';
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Thanks for the review @rebmullin, and thanks for catching that missing requirement! I've added the links, only creating links when there are actually reviews to link to. The tests got a bit ugly because of the way the `Link`s are embedded inside the `MetadataCard`, but I think my approach is the most appropriate for enzyme.

src/amo/components/AddonMeta/index.js
```diff
@@ -39,14 +40,38 @@ export class AddonMetaBase extends React.Component<Props> {
...
+    } else if (addonReviewCount) {
+      reviewCount = i18n.formatNumber(addonReviewCount);
+      reviewTitle = i18n.ngettext('Review', 'Reviews', addonReviewCount);
     } else {
-      reviewTitle = i18n.gettext('No Ratings');
+      reviewTitle = i18n.gettext('No Reviews');
     }
 
+    const reviewsContent =
+      addon && reviewCount ? (
```
rebmullin: could this be a variable since you use down below too? maybe something like `const hasReviews`...

src/amo/components/AddonMeta/index.js
```diff
@@ -39,14 +40,38 @@ export class AddonMetaBase extends React.Component<Props> {
...
+        </Link>
+      ) : (
+        reviewCount
+      );
+
+    const reviewsTitle =
+      addon && reviewCount ? (
+        <Link
+          className="AddonMeta-reviews-title-link"
+          to={`/addon/${addon.slug}/reviews/`}
```
rebmullin: could this url be a var too, since it's used here and above?

src/amo/components/AddonMeta/styles.scss
```diff
@@ -36,3 +36,24 @@
   @include margin-end(-2.5px);
   @include margin-start(-2.5px);
 }
+
+.AddonMeta .MetadataCard-content,
+.AddonMeta .MetadataCard-title {
+  a,
+  a:link,
+  a:visited {
+    color: $grey-50;
```
rebmullin: do we want to change the color here for the number too? 

not that this should be a blocker but I am also wondering if there is a more visual way to let the user know it's a link? Maybe this is okay..? But right now unless you hover over it, it's hard to tell..

src/amo/components/AddonMeta/styles.scss
```diff
@@ -36,3 +36,24 @@
...
+  }
+
+  a:active,
+  a:hover {
+    color: $link-color;
+  }
+
+  a:focus {
+    outline: 1px dotted $link-color;
+    outline: auto 5px -webkit-focus-ring-color;
```
rebmullin: do we still need this 2nd outline?

rebmullin:
> The tests got a bit ugly because of the way the Links are embedded inside the MetadataCard, but I think my approach is the most appropriate for enzyme.

The tests looked okay to me but maybe @willdurand or @kumar303 can take a quick look too

src/amo/components/AddonMeta/styles.scss
bobsilverberg: I did what @kumar303 suggested and used the default [focus mixin](https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/mixins.scss#L92).

src/amo/components/AddonMeta/index.js
bobsilverberg: I tried, and Flow won't let me do that. If I don't explicitly specify `addon && reviewCount`, but rather put it into a `const`, Flow complains that `addon` can be null. I'm surprised about this, and Kumar agrees it seems like a Flow bug, but I don't know if there's much we can do about it.

Here's a [simple example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCAJsXAHZgC8YA3gD6phjECGGrAXHeawLZ5uAZwwAnAJbkA5gF8w9MOQCuMGABpU9GQG50AYwoiwGPEZoAKVqQrcSZSguWqAlNQB8dJmAPkjAC1YhABF2VmowK3swADJoiOtyADo2Dl1mcSgwcwDg0NdaLwBIUTwMJVFKSIpk0MTeATSwGVQdVCA) of the bug.

src/amo/components/AddonMeta/styles.scss
bobsilverberg: I did this to match the mocks. I think it would be a bit ugly to have the title or the number in blue in among all the grey text, although you make a good point that the fact that it's a link is not obvious.

What do you think @pwalm?

tests/unit/amo/components/TestAddonMeta.js
```diff
@@ -89,48 +90,105 @@ describe(__filename, () => {
...
+        getReviewTitle(root)
+          .find(Link)
+          .children(),
+      ).toHaveText('Reviews');
+      expect(getReviewCount(root).find(Link)).toHaveProp(
+        'to',
+        `/addon/${slug}/reviews/`,
+      );
+      expect(
+        getReviewCount(root)
```
kumar303: I think it would be worth moving up `getReviewCount(root)` and `getReviewTitle(root)` to a shared variable.

src/amo/components/AddonMeta/styles.scss
```diff
@@ -42,3 +42,23 @@
   display: flex;
   justify-content: center;
 }
+
+.AddonMeta .MetadataCard-content,
+.AddonMeta .MetadataCard-title {
+  a,
+  a:link,
+  a:visited {
+    color: $grey-50;
```
kumar303: I think the number should be black, like before, instead of gray. This will match the mocks.

Your patch:

screenshot 2018-08-14 10 38 40


The mocks:

screenshot 2018-08-14 10 39 04


I also have questions about how to make these look more like links but let's save that for a follow-up. I like how the links turn blue when you hover over them.

src/amo/components/AddonMeta/index.js
```diff
@@ -48,14 +51,33 @@ export class AddonMetaBase extends React.Component<InternalProps> {
...
+    const reviewsContent = reviewsLink ? (
+      <Link className="AddonMeta-reviews-content-link" to={reviewsLink}>
+        {reviewCount}
+      </Link>
+    ) : (
+      reviewCount
+    );
+
+    const reviewsTitle = reviewsLink ? (
+      <Link className="AddonMeta-reviews-title-link" to={reviewsLink}>
```
kumar303: Hmm. I thought I fixed the centering for this but I guess not.

Your patch:

screenshot 2018-08-14 10 46 39


Could you add some centering? TBH I'm not sure why `justify-content: center` doesn't already do that. I guess it only applies to the block, not the text.

```diff
diff --git a/src/amo/components/AddonMeta/styles.scss b/src/amo/components/AddonMeta/styles.scss
index 0a8fb8ecb..58dd7df8a 100644
--- a/src/amo/components/AddonMeta/styles.scss
+++ b/src/amo/components/AddonMeta/styles.scss
@@ -41,6 +41,7 @@
 .AddonMeta-rating-title {
   display: flex;
   justify-content: center;
+  text-align: center;
 }
 
 .AddonMeta .MetadataCard-content,
```

screenshot 2018-08-14 10 47 14

src/amo/components/AddonMeta/index.js
rebmullin: ahh yep, I've had similar issues..

src/amo/components/AddonMeta/styles.scss
```diff
@@ -41,4 +41,31 @@
...
+    font-weight: normal;
+    text-decoration: none;
+  }
+
+  a.AddonMeta-reviews-content-link,
+  a:link.AddonMeta-reviews-content-link,
+  a:visited.AddonMeta-reviews-content-link {
```
kumar303: You can DRY this up with the Sass ampersand operator:

```diff
diff --git a/src/amo/components/AddonMeta/styles.scss b/src/amo/components/AddonMeta/styles.scss
index 0783997da..492f12092 100644
--- a/src/amo/components/AddonMeta/styles.scss
+++ b/src/amo/components/AddonMeta/styles.scss
@@ -54,10 +54,12 @@
     text-decoration: none;
   }
 
-  a.AddonMeta-reviews-content-link,
-  a:link.AddonMeta-reviews-content-link,
-  a:visited.AddonMeta-reviews-content-link {
-    color: $black;
+  a.AddonMeta-reviews-content-link {
+    &,
+    &:link,
+    &:visited {
+      color: $black;
+    }
   }
 
   a:active,
```

PR #5741, opened 2018-07-26
Add Hybrid Content Telemetry

### LINKED ISSUE #4528: Implement Hybrid Content Telemetry on Disco Pane
This issue is to implement the [Hybrid Content Telemetry](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/hybrid-content.html#) on the discovery pane.

The library requires a permission - this is being added in: https://bugzilla.mozilla.org/show_bug.cgi?id=1444114 - the work to implement is not necessarily blocked by this though.

Once added this should expose the relevant APIs to the browser. 

* We should look to implement it in a way that mirrors the current GA tracking  - the lib has been designed for this case so that should be straight-forward.
* It should be run in parallel with the google tracking initially and allow for turning each one on or off independently of each other.
* The usage of HCT should be setup with a config var so that observing DNT is configurable based on the implementation. This is because for this implementation which is in-product use (e.g. the disco pane which is shown in Firefox) [HCT will observe telemetry prefs in Firefox](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/hybrid-content.html#js-api). However, should we in the future want to use this on a web property we may want DNT to be the control as to whether data is collected.

muffinresearch:
Fixes #4528 

This branch adds support for Hybrid Content Telemetry tracking on the discovery pane. It's currently configured off, but once this passes review I'll enable it on -dev and it will need a data review before we turn it on in production.

If enabled it loads a separate library asset (it could be merged into the disco bundle if preferred.

In addition to the core HCT parts and tests along the way it implements:

* Adds handling of dynamic imports (from a node perspective) to prevent build and test errors. This uses the es6 dynamic import syntax which is now preferred over things like require.ensure.
* Adds some tests to protect tracking constants from modification.

The main tracking call is based on an initialization promise. This way if called too soon it will wait until the promise is resolved.


willdurand:
How would you feel about moving most of the code in `disco/tracking` instead? (and remove the babel/webpack setup to dynamically import code)

I think `disco/tracking` could expose the same API as `core/tracking` by decorating it: forwarding all calls to the `core/tracking` instance and adding the new call to HCT in `sendEvent()`.

muffinresearch:
@willdurand I considered something along similar lines initially, but the problem with it is that the install tracking is shared between AMO and the disco pane. So how would we structure things to make it clear where the tracking lib is coming from?

It would be possible to make this work with webpack magic (like we do for loggin on the client and server), but I think we should try and avoid doing that if possible. Are there other options for clearer separation?

willdurand:
> @willdurand I considered something along similar lines initially, but the problem with it is that the install tracking is shared between AMO and the disco pane. So how would we structure things to make it clear where the tracking lib is coming from?

What exactly needs the tracking? Components like `InstallButton` could be configured with a `tracking` prop to pass the right instance.

muffinresearch:
Incidently the non es5 code in the lib is breaking uglification during builds, I'll need to file a bug on hct about that to get that resolved quickly unless we can upgrade to uglify-es sooner.

muffinresearch:
> What exactly needs the tracking that is also used by the disco pane? Components like InstallButton could be configured with a tracking prop to pass the right instance.

That could work, what would we need to set that up, can we have tracking passed from the app downwards so it's available anywhere anyone wants to use it?

willdurand:
> That could work, what would we need to set that up, can we have tracking passed from the app downwards so it's available anywhere anyone wants to use it?

We should find the components used by both AMO and Disco Pane and that needs tracking, then we'll have to add a new `tracking` prop, but I believe most of them already have a `_tracking` (hidden) prop.

muffinresearch:
Will and I discussed this branch last week and whilst we both like the idea of splitting the code for tracking between Disco and AMO, keeping the hct part of a single tracking file is going to be less error prone because it won't be possible to introduce code that might use the wrong tracking instance. This also avoids having anything set up to pass tracking around the application.

The patch has also been updated with the following changes:

* Uses `await`/`async` rather than `then()`
* Updates to the patched version of the hct lib that should prevent uglify errors during builds.
* Introduces Invariant to catch unregistered data being passed to telemetry.

bin/server.js
```diff
@@ -4,7 +4,9 @@ const path = require('path');
 
 const chalk = require('chalk');
 const chokidar = require('chokidar');
-require('babel-register');
+require('babel-register')({
+  plugins: ['dynamic-import-node'],
```
willdurand: Can't we enable it globally in the `.babelrc` file?

src/core/tracking.js
```diff
@@ -59,33 +92,36 @@ export class Tracking {
...
+    this.logPrefix = `[GA: ${this.trackingEnabled ? 'ON' : 'OFF'}]`;
 
-    if (this.enabled) {
+    if (this.trackingEnabled) {
       /* eslint-disable */
       // Snippet from Google UA docs: http://bit.ly/1O6Dsdh
       window.ga =
         window.ga ||
         function() {
+          /* istanbul ignore next */
```
willdurand: I have mixed feelings regarding the use of such comments. It is okay to not reach 100% code coverage sometimes. We just lie to ourselves here. If we refactor the code and leave this comment for some reasons, we may miss useful code coverage.

src/core/tracking.js
```diff
@@ -98,6 +134,44 @@ export class Tracking {
...
+    if (typeof window !== 'undefined' && hctEnabled === true) {
+      log.info('Setting up the Hybrid Content Telemetry lib');
+      // Note: special webpack comments must be after the module name or
+      // babel-plugin-dynamic-import-node will blow-up.
+      const hybridContentTelemetry = await import('mozilla-hybrid-content-telemetry/HybridContentTelemetry-lib');
+      /* webpackChunkName: "disco-hct" */
+      await hybridContentTelemetry.initPromise();
+      let logHctReason;
+      if (!hybridContentTelemetry) {
+        /* istanbul ignore next */
```
willdurand: same here.

src/core/tracking.js
```diff
@@ -98,6 +134,44 @@ export class Tracking {
...
+    this.hctInitPromise = this.initHCT({ _config });
+  }
+
+  async initHCT({ _config = config } = {}) {
+    const hctEnabled = _config.get('hctEnabled');
+    if (typeof window !== 'undefined' && hctEnabled === true) {
+      log.info('Setting up the Hybrid Content Telemetry lib');
+      // Note: special webpack comments must be after the module name or
+      // babel-plugin-dynamic-import-node will blow-up.
+      const hybridContentTelemetry = await import('mozilla-hybrid-content-telemetry/HybridContentTelemetry-lib');
```
willdurand: The `await` should be used inside a `try/catch` block to avoid a potential issue with unhandled promises.

src/core/tracking.js
```diff
@@ -107,11 +181,44 @@ export class Tracking {
...
+          )}`,
+        );
+        invariant(
+          telemetryObjects.includes(object),
+          `Object "${object}" must be one of the registered values: ${telemetryObjects.join(
+            ',',
+          )}`,
+        );
+        hybridContentTelemetry.recordEvent(
+          'disco.interaction',
```
willdurand: This value could probably become a constant.

src/core/tracking.js
```diff
@@ -107,11 +181,44 @@ export class Tracking {
...
+          method,
+          object,
+          data.value,
+        );
+      } else {
+        this.log(`Not logging to telemetry canUpload(): ${canUpload}`);
```
willdurand: > [...] because canUpload() returned: ${canUpload}

tests/unit/core/test_tracking.js
```diff
@@ -27,25 +31,25 @@ import {
   UNINSTALL_THEME_CATEGORY,
 } from 'core/constants';
 
+function stubConfig(overrides = {}) {
```
willdurand: I think you should use `getFakeConfig()` from `tests/unit/helpers`.

tests/unit/core/test_tracking.js
```diff
@@ -27,25 +31,25 @@ import {
...
+  const config = {
+    trackingEnabled: true,
+    trackingId: 'sample-tracking-id',
+    ...overrides,
+  };
+  return { get: sinon.spy((key) => config[key]) };
+}
+
+function createTracking(overrides = {}) {
```
willdurand: We also have a `createFakeTracking()` helper in `tests/unit/helpers` if you don't need the full object.

tests/unit/core/test_tracking.js
```diff
@@ -353,4 +357,194 @@ describe(__filename, () => {
       sinon.assert.calledWith(fakeLog.log, 'Do Not Track is enabled');
     });
   });
+
+  describe('Hybrid Content Telemetry filterIdentifier()', () => {
+    it('should return content up to the default max length', () => {
+      expect(filterIdentifier('abcdefghijklmnopqrstuvwxyz').length).toEqual(20);
+    });
+
+    it('should return content up to the supplie max length', () => {
```
willdurand: supplied?

tests/unit/core/test_tracking.js
```diff
@@ -353,4 +357,194 @@ describe(__filename, () => {
       sinon.assert.calledWith(fakeLog.log, 'Do Not Track is enabled');
     });
   });
+
+  describe('Hybrid Content Telemetry filterIdentifier()', () => {
+    it('should return content up to the default max length', () => {
+      expect(filterIdentifier('abcdefghijklmnopqrstuvwxyz').length).toEqual(20);
```
willdurand: You may want to use [`toHaveLength()`](https://jest-bot.github.io/jest/docs/expect.html#tohavelengthnumber) instead of `toEqual`.

tests/unit/core/test_tracking.js
```diff
@@ -353,4 +357,194 @@ describe(__filename, () => {
...
+    });
+
+    it('should return content up to the supplie max length', () => {
+      expect(
+        filterIdentifier('abcdefghijklmnopqrstuvwxyz', { maxLen: 10 }).length,
```
willdurand: You could also repeat a letter `N` times to be more explicit (I had to count the number of letters). I think `'a'.repeat(20)` should work. This comment also applies to the previous test case.

tests/unit/core/test_tracking.js
```diff
@@ -353,4 +357,194 @@ describe(__filename, () => {
...
+    });
+
+    it('should not change the tracking category constants for extension uninstalls', () => {
+      expect(UNINSTALL_EXTENSION_CATEGORY).toEqual('AMO Addon Uninstalls');
+    });
+
+    it('should not change the tracking category constants for clicks', () => {
+      expect(CLICK_CATEGORY).toEqual('AMO Addon / Theme Clicks');
+    });
+  });
```
willdurand: 👍

tests/unit/core/test_tracking.js
willdurand: Given the test cases below, I think you need the full `Tracking` object so forget this comment.

tests/unit/core/test_tracking.js
```diff
@@ -353,4 +357,194 @@ describe(__filename, () => {
...
+
+    it('should throw if unregistered method is used', () => {
+      canUploadStub.callsFake(() => true);
+      const tracking = createTracking({
+        _config: stubConfig({ hctEnabled: true }),
+      });
+      const badTrackingData = Object.assign({}, trackingData, {
+        method: 'failMethod',
+      });
+      return expect(tracking._hct(badTrackingData)).rejects.toThrow(
```
willdurand: We usually don't test the invariants. That's why we use this lib (to stop writing such test cases). Invariant is not used in production.

muffinresearch: Ah ok, makes sense 👍

tests/unit/core/test_tracking.js
muffinresearch: I used the convention already established in this file. Would it be best to update all of these in one go independently?

bin/server.js
muffinresearch: If we do that it would be transpiled for the client when it doesn't need to be, because webpack understands `import()`.

willdurand:
When trying this patch locally, with `hctEnabled: true` and `yarn disco`, I get this error in the devtools:

```
ReferenceError: webpackJsonp is not defined[Learn More] disco-hct-1810bede5c7921845153.js:1:1
<anonymous>
http://127.0.0.1:3001/disco-hct-1810bede5c7921845153.js:1:1

```

willdurand:
@muffinresearch the issue above occurs because the `disco-hct` chunk gets injected into the HTML page. We already exclude `i18n` chunks, so I think a patch like this would work:

```diff
diff --git a/src/core/components/ServerHtml/index.js b/src/core/components/ServerHtml/index.js
index fceb76b..9df02dd 100644
--- a/src/core/components/ServerHtml/index.js
+++ b/src/core/components/ServerHtml/index.js
@@ -33,7 +33,7 @@ export default class ServerHtml extends Component {
     const leafName = filePath.split('/').pop();
     let sriProps = {};
     // Only output files for the current app.
-    if (leafName.startsWith(appName) && !leafName.includes('i18n')) {
+    if (leafName.startsWith(appName) && !/(i18n|hct)/.test(leafName)) {
       if (includeSri) {
         sriProps = {
           integrity: sriData[leafName],
```

Actually, it works because I tested it, but I don't think we should be doing this. A config entry would probably make more sense.

muffinresearch:
@willdurand I think this suggests the magic bundle-splitting comment is broken. Are you up to date? Because I was having issues with prettier mangling that comment.

willdurand:
> @willdurand I think this suggests the magic bundle-splitting comment is broken. Are you up to date? Because I was having issues with prettier mangling that comment.

I am, and nothing is broken. The devtools network panel indicates that `disco-hct.js` is fetched twice. If you look at the generated HTML, you'll probably find a `script` tag requiring the `disco-htc.js` file.

muffinresearch:
Ah I see what your saying now. That's a good catch 👍

muffinresearch:
I think I might stop this being bundle split. That would solve the problem too. I wanted to double check the bundle-sizes before landing this just to make sure this isn't bloating the amo file.

willdurand:
> I think I might stop this being bundle split. That would solve the problem too. I wanted to double check the bundle-sizes before landing this just to make sure this isn't bloating the amo file.

That's a good use case for showing how code splitting works though, I am not against keeping it.

willdurand:
@muffinresearch I disabled uglify and there are lots of `unused ...` comments, which means dead code elimination works. I believe React dev stuff are removed too, I cannot find anything.

```
/* unused harmony export defaultMergeProps */
/* unused harmony export wrapMergePropsFunc */
/* unused harmony export whenMergePropsIsFunction */
/* unused harmony export whenMergePropsIsOmitted */
```

muffinresearch:
@willdurand I've run out of time for landing this patch this week since other more pressing concerns took precedence.

Here's a patch for wrapping the hct code in blocks that should allow for elimination: 

https://gist.github.com/muffinresearch/b82d27de251758ed79e467e08b4d41fa

I was testing something like this previously and got no useful savings.

willdurand:
I'll try to continue this work and investigate the issue about dead code elimination.

willdurand:
I just rebased this branhc.

willdurand:
So, [your patch](https://github.com/mozilla/addons-frontend/pull/5741#issuecomment-412123447) allows webpack to replace config values and enable dead code elimination (there is no need to change the uglify config though, it is already the default behavior):

```
    // Attempt to enable Hybrid content telemetry.
    if (false) {
      this.hctInitPromise = this.initHCT({ _config: _config });
    }
```

Yet, from what I am seeing, webpack compiles the rest of the JS in a different manner, i.e. not keeping class and methods (it creates objects instead). That's why there is no dead code elimination, because webpack is unable to wrap this code into a `if (false) {}` block.

willdurand:
@muffinresearch I don't see a huge value in adding the patch given in https://github.com/mozilla/addons-frontend/pull/5741#issuecomment-412123447 because we won't remove a lot of code.

Yet, the code splitting still makes sense because we still extract the lib and it will never get loaded if we do not use it. In order to make it work, we still need to fix the issue mentioned in https://github.com/mozilla/addons-frontend/pull/5741#issuecomment-410981861.

muffinresearch:
@willdurand ok, thanks for digging into it. I'll fix the issue and aim to land this for next week.

muffinresearch:
@willdurand ok I think this is ready for another look. I fixed up the chunk exclusion using a config var and after fixing up some missing categories/methods it's looking to work with both the new and old disco buttons.

muffinresearch:
> Yet, from what I am seeing, webpack compiles the rest of the JS in a different manner, i.e. not keeping class and methods (it creates objects instead). That's why there is no dead code elimination, because webpack is unable to wrap this code into a if (false) {} block.

I wonder if this is something that would be fixed with tree-shaking?

src/core/components/ServerHtml/index.js
```diff
@@ -8,6 +8,10 @@ import config from 'config';
 
 import { LTR } from 'core/constants';
 
+const JS_CHUNK_EXCLUDES = new RegExp(
+  `(?:${config.get('jsChunkExclusions').join('|')})`,
+);
```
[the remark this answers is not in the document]
willdurand: how about creating the regex in the constructor then?

src/core/tracking.js
```diff
@@ -107,6 +146,49 @@ export class Tracking {
...
+      log.info('Setting up the Hybrid Content Telemetry lib');
+      // Note: special webpack comments must be after the module name or
+      // babel-plugin-dynamic-import-node will blow-up.
+      try {
+        // prettier-ignore
+        const hybridContentTelemetry = await import(
+          'mozilla-hybrid-content-telemetry/HybridContentTelemetry-lib'
+          /* webpackChunkName: "disco-hct" */
+        );
+        await hybridContentTelemetry.initPromise();
```
willdurand: I noticed that when we do not have access to hct telemetry, this method stops here and there is no more logs, no errors, nothing..

muffinresearch: Hmm that would seem to be a bug in the lib. It should probably reject in this case.

muffinresearch: I've filed a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1490284

src/core/components/ServerHtml/index.js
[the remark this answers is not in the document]
muffinresearch: What do you think is missing from a testing perspective?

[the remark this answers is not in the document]
willdurand: @muffinresearch ah sorry, I missed your last comment.

Ideally, we should be able to inject `config` with a custom config, for example, a specific value for `jsChunkExclusions`. The problem with global constants is that they are... global and less testable. We cannot easily write a unit test to cover a change in this regex for instance. That's why you had to add code in this file: https://github.com/mozilla/addons-frontend/pull/5741/commits/66e00fefa5754c2e560e730e83e2226a84c3b296#diff-6a0e1012d7256e185eee9b88a87725a7.

A more "pure" unit test (I think) would have been to inject a custom `_config` and call `getStatic()`. That would be slightly more isolated.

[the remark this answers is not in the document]
willdurand: It could probably be done later, though, since this works.

src/core/tracking.js
willdurand: added: "blocked by upstream" label

muffinresearch:
So the latest version of the HCT lib and latest nightly fixes the promise issue. But older browsers won't see the promise reject, however on the basis that this doesn't seem to impact GA in the meantime and that we'll do further testing on -dev before it's released I think this is fine.

PR #5749, opened 2018-07-27
HMR for Redux

### LINKED ISSUE #1096: Make redux actions hot reloadable again

willdurand:
Fix #1096

---

Add HMR for redux code. It is possible to add a new reducer without reloading and/or to change the code of a reducer. I have added some tests for the `core/client/base.js` file because the HMR code depends on it.

I enabled the `NamedModulesPlugin` webpack plugin to ease HMR debugging (it prints the filenames instead of numbers when something cannot be updated).


src/amo/client.js
```diff
@@ -7,11 +10,18 @@ import App from './components/App';
 import sagas from './sagas';
 import createStore from './store';
 
-createClient(createStore, { sagas }).then(({ renderApp }) => {
+createClient(createStore, { sagas }).then(({ history, renderApp, store }) => {
   renderApp(App);
 
   if (module.hot) {
-    module.hot.accept('./components/App', () => {
+    module.hot.accept(['./sagas', './store'], () => {
```
kumar303: It would be nice to see a comment here explaining that sagas aren't hot reloadable but this is needed to facilitate reducer/store reloading.

src/amo/client.js
```diff
@@ -7,11 +10,18 @@ import App from './components/App';
...
+      // eslint-disable-next-line global-require
+      const { reducers } = require('./store');
+      const nextRootReducer = connectRouter(history)(combineReducers(reducers));
```
kumar303: To minimize bugs, I'd like to see `connectRouter(history)(combineReducers(reducers))` moved to a function like `createRootReducer()` (or something) so it can also be called [in the real store](https://github.com/mozilla/addons-frontend/blob/bc32a8c978894f7dc2fbd1135cdc710a07dd6aec/src/amo/store.js#L133).

src/disco/client.js
```diff
@@ -19,10 +21,17 @@ if (config.get('trackingSendInitPageView') === false) {
...
 
-createClient(createStore, { sagas }).then(({ renderApp }) => {
+createClient(createStore, { sagas }).then(({ history, renderApp, store }) => {
   renderApp(App);
 
   if (module.hot) {
+    module.hot.accept(['./sagas', './store'], () => {
+      // eslint-disable-next-line global-require
+      const { reducers } = require('./store');
+      const nextRootReducer = connectRouter(history)(combineReducers(reducers));
```
kumar303: This may also benefit from a `createRootReducer()` function.

tests/unit/core/client/test_base.js
```diff
@@ -0,0 +1,51 @@
...
+describe(__filename, () => {
+  describe('createClient()', () => {
+    const fakeCreateStore = () => {
+      return {
+        sagaMiddleware: null,
+        store: null,
+      };
+    };
+
+    const fakeFastClick = {
```
kumar303: I'd rather see this initialized before every test otherwise the call history of the stub *might* leak across tests. It might not because of [sandboxing](https://github.com/mozilla/addons-frontend/blob/master/tests/setup.js#L71) but I'd feel better if it were a new instance mainly as not to obscure the intention of the code 😀 

For easy access in assertions, we usually do it like:

```js
let fakeFastClick;

beforeEach(() => {
  fakeFastClick = {
    attach: sinon.stub(),
  };
});
```

webpack.dev.config.babel.js
```diff
@@ -71,6 +71,7 @@ export default Object.assign({}, webpackConfig, {
       'react-dom/umd/react-dom.development.js',
     ),
     new webpack.NormalModuleReplacementPlugin(/^redux$/, 'redux/dist/redux.js'),
+    new webpack.NamedModulesPlugin(),
```
kumar303: D'oh. Good find. It looks like this will be the default for development in [webpack 4](https://github.com/webpack/webpack/issues/3476#issuecomment-368300958).

src/amo/client.js
willdurand: That totally makes sense :+1:

willdurand:
rebased.

PR #5760, opened 2018-07-27
Sort all the props \o/

### LINKED ISSUE #5775: Sort destructured props
This issue is a follow-up to some comments in #5652. When upgrading to React Router v4, most of the props have been sorted but some might have been missed, and we did not have an ESLint rule for that. The ESLint rule has since been created: https://github.com/mozilla/eslint-plugin-amo#sort-destructured-props.

---

We (try to) enforce sorted destructured properties in function arguments and destructured assignements:

```js
const foo = ({ a, b }) => {};

const { a, b } = this.props;
```

This rule allows us to always know where to insert a new property, thus avoiding unnecessary thinking. We follow the "natural sort order":

```js
const { _underscoreProps, CapitalizedProps, aLowerCase, bLowerCase, ...otherProps } = props;
```

⚠️ Thanks to the ESLint rule, we do not have to do anything anymore (it is fixed automatically), **except** when a prop is used as a default value of another prop in the same destructuration:

```js
const foo = ({
  userId = 123,
  aPropThatShouldBeListedFirst = userId,
}) = > {
};
```

The ESLint rule will probably try to re-order the list of props as follows:

```js
const foo = ({
  aPropThatShouldBeListedFirst = userId,
  userId = 123,
}) = > {
};
```

But another ESLint rule will throw an error because `userId` is used before it is declared. In this case, it is best to disable the `amo/sort-destructured-props` props:

```js
// eslint-disable-next-line amo/sort-destructured-props
const foo = ({
  userId = 123,
  aPropThatShouldBeListedFirst = userId,
}) = > {
};
```

willdurand:
Fix #5775

---

- [x] upgrade `eslint-plugin-amo` to version `1.5.0`
- [x] apply `eslint --fix` to automatically sort all the destructured props


willdurand:
> can we at least do it only for multi-line destructors?

This would be tricky to automate because ESLint does not really know whether a statement has newlines or not (it is all about the syntax).

kumar303:
> ESLint does not really know whether a statement has newlines or not (it is all about the syntax)

Maybe it could count characters or words to guess whether or not it will span multiple lines?

willdurand:
I'm going to close this PR for now. I leave the ticket open though.

PR #5765, opened 2018-07-27
Fix normalization problems in rating/review state

### LINKED ISSUE #3221: Fix review by user state so that it stores object IDs not objects
**For QA:**
Please check for regressions in the flows related to leaving a star rating, writing a review, editing a review, and browsing reviews.

As of https://github.com/mozilla/addons-frontend/pull/3220 the review state has a map of reviews by ID. All other parts of the state link to IDs instead of objects. However, I didn't have time to fix the user state because it wasn't relevant for the patch and it introduced some complexities. It will rot away so we should fix it.

kumar303:
Fixes #3221

The old code had oh so many problems. I focused on fixing these:
* It had a custom state mapping of user review objects which was out of sync with other parts of the state. Mainly, this custom mapping was not pointing its review objects to `state.byId`.
* It tried to simulate how the server sets and resets `review.isLatest`

My solutions:
* There is a new state mapping dedicated to latest user reviews
* No more insane key nesting (`reviewsState[userId][addonId][versionId]`)
* No server logic is simulated. The code simply clears the cache of server objects where necessary.

To keep this patch small-ish and focused, I **did not** fix these problems:
* `TestRatingManager` should use `shallowUntilTarget`
* `TestRatingManager` should dispatch actions to generate mapped props
* `RatingManager` should use sagas for async actions
* The reviews reducer should be a duck
* Action types should not be littered in `core/constants.js`
* `RatingManager` should use `withUIState()` (this actually causes a real bug that I'm about to file)


tests/unit/amo/reducers/testReviews.js
```diff
@@ -812,4 +710,143 @@ describe(__filename, () => {
...
+      );
+
+      expect(
+        state.latestByAddonVersion[
+          latestByAddonVersionKey({ addonId, userId, versionId })
+        ],
+      ).toEqual(review.id);
+    });
+
+    it('can set the latest review to null', () => {
```
rebmullin: is this possible?

tests/unit/amo/reducers/testReviews.js
```diff
@@ -155,49 +155,8 @@ describe(__filename, () => {
...
-      }),
-    );
-
-    // Make sure all reviews co-exist by userId, addonId, review ID.
-    expect(state[1][1][1].rating).toEqual(1);
-    expect(state[1][2][2].rating).toEqual(5);
-    expect(state[2][2][3].rating).toEqual(4);
-  });
-
   it('preserves existing add-on reviews', () => {
```
rebmullin: it seems like we have 2 test cases with this same name.  or should we keep both but clarify a bit more?

src/amo/actions/reviews.js
```diff
@@ -431,3 +432,30 @@ export const clearAddonReviews = ({
...
+  userId: number,
+|};
+
+export type SetLatestReviewAction = {|
+  type: typeof SET_LATEST_REVIEW,
+  payload: SetLatestReviewParams,
+|};
+
+export const setLatestReview = (
+  params: SetLatestReviewParams,
```
willdurand: Is this a new pattern you would like to introduce? Usually, we list all the properties of that object.

src/amo/components/RatingManager/index.js
```diff
@@ -90,14 +93,25 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+    const {
+      apiState,
+      loadSavedReview,
+      userId,
+      addon,
+      userReview,
+      version,
+    } = this.props;
+    if (userId && addon && userReview === undefined) {
+      log.debug(`loading a saved rating (if it exists) for user ${userId}`);
```
willdurand: nit: I believe we capitalize the first word of a log message.

src/amo/components/RatingManager/index.js
```diff
@@ -232,7 +246,7 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
             <UserRating
               readOnly={!isLoggedIn}
               onSelectRating={this.onSelectRating}
-              review={userReview}
+              review={userReview || undefined}
```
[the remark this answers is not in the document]
willdurand: +1

src/amo/components/RatingManager/index.js
```diff
@@ -245,25 +259,18 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
-      addonId ${ownProps.addon.id}, versionId ${ownProps.version.id}`);
-
-    const allUserReviews = state.reviews[userId] || {};
-    const addonReviews = allUserReviews[ownProps.addon.id] || {};
-    const latestId = Object.keys(addonReviews).find(
-      (reviewId) => addonReviews[reviewId].isLatest,
+  if (userId && ownProps.addon) {
+    log.debug(
+      `Looking for latest review of addon:${ownProps.addon.id}/version:${
+        ownProps.version.id
```
willdurand: You may want to create a variable for `versionId`, it is used twice.

src/amo/components/RatingManager/index.js
```diff
@@ -276,25 +283,53 @@ export const mapStateToProps = (state: AppState, ownProps: Props) => {
...
       }
     });
   },
 
   submitReview(params) {
-    return submitReview(params).then((review) => dispatch(setReview(review)));
+    return submitReview(params).then((review) => {
+      // The API could possibly return a null review.version if that
+      // version was deleted. In that case, we fall back to the submitted
+      // versionId which came from the page data. It is highly unlikely
```
willdurand: Confidence? Hopefully 100% because we'll never get the information in -prod since `invariant` calls are stripped.

src/amo/reducers/reviews.js
```diff
@@ -209,12 +189,50 @@ export const getReviewsByUserId = (
     : null;
 };
 
+export const latestByAddonVersionKey = ({
```
willdurand: How about prefixing this function with `make`?

src/amo/reducers/reviews.js
```diff
@@ -209,12 +189,50 @@ export const getReviewsByUserId = (
...
+  userId,
+  addonId,
+  versionId,
+}: {|
+  userId: number,
+  addonId: number,
+  versionId: number,
+|}) => {
+  return `user${userId}-addon${addonId}-version${versionId}`;
```
willdurand: Do we need to read those values? If so, we could use a slash as separator and use the dash between the prefix and its value.

src/amo/reducers/reviews.js
```diff
@@ -72,70 +74,48 @@ export type ReviewsState = {|
...
 };
 
-type ExpandReviewObjectsParams = {|
-  state: ReviewsState,
-  reviews: Array<number>,
-|};
+export const selectReview = (
+  reviewsState: ReviewsState,
+  id: number,
+): UserReviewType | void => {
```
willdurand: Grr. I thought that `void` was a way to express "this function does not return anything" but Flow actually uses `void` to express `undefined`... 🤦‍♂️

src/amo/reducers/reviews.js
```diff
@@ -209,12 +189,50 @@ export const getReviewsByUserId = (
...
+  reviewsState: ReviewsState,
+  userId: number,
+  addonId: number,
+  versionId: number,
+|}): UserReviewType | null | void => {
+  const key = latestByAddonVersionKey({ userId, addonId, versionId });
+  const userReviewId = reviewsState.latestByAddonVersion[key];
+
+  if (userReviewId === null) {
+    // This means an action had previously attempted to fetch the latest
```
willdurand: Does that mean you make the distinction between `null` and `undefined` in the caller code?

willdurand: Otherwise, I would restrict the return value type to either `UserReviewType` or `null`.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -106,10 +110,32 @@ describe(__filename, () => {
...
+    const userId = 12889;
+    const addon = createInternalAddon({ ...fakeAddon, id: 3344 });
+    const version = {
+      ...fakeAddon.current_version,
+      id: 9966,
+    };
+    const loadSavedReview = sinon.spy();
+
+    render({
+      apiState: signedInApiState,
```
willdurand: Too bad this one does not come from `dispatchSignInActions().store`.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -264,6 +290,16 @@ describe(__filename, () => {
...
 
+  it('sets an undefined UserRating review when none exists', () => {
+    const UserRatingStub = sinon.spy(() => <div />);
+
+    render({ UserRating: UserRatingStub, userReview: null });
+
+    sinon.assert.calledWithMatch(UserRatingStub, {
+      review: undefined,
+    });
+  });
```
willdurand: ah, here is the test case related to the `|| undefined` above I guess. I am not sure why `undefined` is better than `null` though: is it only for respecting the `Props` definition of the `UserRating` component? We could also accept `null` as a valid value.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -377,30 +413,68 @@ describe(__filename, () => {
     });
 
     describe('submitReview', () => {
-      it('posts the reviw and dispatches the created review', () => {
+      it('posts the review and dispatches the created review', () => {
```
willdurand: 🥇

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -377,30 +413,68 @@ describe(__filename, () => {
...
+        const versionId = 54321;
+        const params = {
+          rating: fakeReview.rating,
+          apiState: { ...signedInApiState, token: 'new-token' },
+          addonId: fakeAddon.id,
+          versionId,
+        };
+
+        // Simulate a review for a deleted add-on version.
+        const reviewResult = { ...fakeReview, version: undefined };
```
willdurand: Should we use `denormalizeReview()`?

willdurand:
Tested locally, works as intended.

src/amo/components/RatingManager/index.js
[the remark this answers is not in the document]
kumar303: Yes, falsey values need to fall back to `undefined`, e.g. when `userReview` could be `null`. The Flow type for `UserRating` doesn't allow `null`. The difference could be subtle but important. In this case a `null` value for `userReview` is handled differently than an `undefined` value.

src/amo/reducers/reviews.js
```diff
@@ -72,70 +74,48 @@ export type ReviewsState = {|
   byAddon: ReviewsByAddon,
   byId: ReviewsById,
   byUserId: ReviewsByUserId,
+  latestByAddonVersion: {
```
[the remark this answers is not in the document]
kumar303: Sure, I can change it

src/amo/reducers/reviews.js
```diff
@@ -209,12 +189,50 @@ export const getReviewsByUserId = (
...
+  const userReviewId = reviewsState.latestByAddonVersion[key];
+
+  if (userReviewId === null) {
+    // This means an action had previously attempted to fetch the latest
+    // user review but one does not exist.
+    return null;
+  }
+
+  // Return the review object or undefined if no attempt has been made
+  // to fetch the latest user review.
```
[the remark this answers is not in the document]
kumar303: I can reword it. It's saying that if the component has not yet attempted to fetch the latest user review, `userReview` will be undefined.

src/amo/components/RatingManager/index.js
```diff
@@ -116,9 +130,9 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
     };
 
     if (userReview) {
-      log.info(`Editing reviewId ${userReview.id}`);
+      log.debug(`Editing reviewId ${userReview.id}`);
```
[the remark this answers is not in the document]
kumar303: I realized the logging in this component is a bit noisy. We don't really filter out debug/info but the messages are really just for debugging so I switched them in case we want to filter it out later.

src/amo/components/RatingManager/index.js
kumar303: I'm 99% confident we'll never get it (on this page anyway) since there is no way to load a deleted version. The code has to be defensive here only because the Flow type says it's *possible* which is technically true.

src/amo/reducers/reviews.js
kumar303: Isn't that the same, though? Not returning anything is the same as returning a literal `undefined`.

src/amo/reducers/reviews.js
kumar303: > Does that mean you make the distinction between null and undefined in the caller code?

Yes. The component only attempts to fetch a `userReview` [when the value is `undefined`](https://github.com/mozilla/addons-frontend/pull/5765/files#diff-883b82043ac0668a2dfdc54fe6914a8cR108). If the value is `null`, it knows that a fetch was already attempted and that no `userReview` exists.

This is a newer pattern that has been working really well, especially for SSR. This component was very old. It was looking for a user review every time an instance was created, regardless if it already had one or not. It was like smashing everything with hammer until it dies.

tests/unit/amo/components/TestRatingManager.js
kumar303: This whole test file needs to be rewritten. Yes, it should be using `dispatchSignInActions()`.

tests/unit/amo/components/TestRatingManager.js
kumar303: The `null` vs `undefined` pattern isn't really anything new. It's the same thing we do in other components with one key difference: loading a `null` (or empty) `userReview` is OK. In most other components, loading empty data renders a 404. Thus, they don't handle the case of rendering with empty data.

tests/unit/amo/components/TestRatingManager.js
kumar303: This is a normalized review object, i.e. one returned from the API with all nested data structures in the JSON.

kumar303:
@willdurand @rebmullin thanks for the reviews!

PR #5773, opened 2018-07-29
I5772-licenses that arent links are breaking pages

### LINKED ISSUE #5772: Addons that have the All Rights Reserved license are now showing up broken

Addons that have the "All Rights Reserved" license are showing up broken 

eg:
https://addons-dev.allizom.org/en-US/firefox/addon/wide-zmeurica/
https://addons-dev.allizom.org/en-US/firefox/addon/source/
https://addons-dev.allizom.org/en-US/firefox/addon/rm-theme-new-sizes/

### What happened?

I think this is related to the recent Router changes and Link being a little bit more strict.

Pages would appear

rebmullin:
fixes #5772

## BEFORE - extension
Alt text

## BEFORE - static theme
Alt text

## AFTER - extension
Alt text

## AFTER - static theme
Alt text

originally I updated just the AddonMoreInfo component but I this might be better to update in the Link...

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -93,23 +93,6 @@ export class AddonMoreInfoBase extends React.Component<Props> {
 
     const currentVersion = addon.current_version;
     const lastUpdated = addon.last_updated;
-    const { license } = currentVersion;
-    let licenseHasUrl = null;
-
-    if (license) {
-      licenseHasUrl = license.url ? (
-        <Link
-          className="AddonMoreInfo-license-link"
-          href={license.url}
-          prependClientApp={false}
-          prependLang={false}
-        >
-          {license.name}
-        </Link>
-      ) : (
-        <span className="AddonMoreInfo-license-type">{license.name}</span>
-      );
-    }
 
     return this.renderDefinitions({
       homepage,
@@ -131,7 +114,17 @@ export class AddonMoreInfoBase extends React.Component<Props> {
             },
           )
         : null,
-      versionLicenseLink: currentVersion ? licenseHasUrl : null,
+      versionLicenseLink:
+        currentVersion && currentVersion.license ? (
+          <Link
+            className="AddonMoreInfo-license-link"
+            href={currentVersion.license.url}
+            prependClientApp={false}
+            prependLang={false}
+          >
+            {currentVersion.license.name}
+          </Link>
+        ) : null,
       privacyPolicyLink: addon.has_privacy_policy ? (
         <Link
           className="AddonMoreInfo-privacy-policy-link"
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -93,6 +93,22 @@ export class AddonMoreInfoBase extends React.Component<Props> {
 
     const currentVersion = addon.current_version;
     const lastUpdated = addon.last_updated;
+    const { license } = currentVersion;
+    let licenseHasUrl = null;
+    if (license) {
+      licenseHasUrl = license.url ? (
+        <Link
+          className="AddonMoreInfo-license-link"
+          href={license.url}
+          prependClientApp={false}
+          prependLang={false}
+        >
+          {license.name}
+        </Link>
+      ) : (
+        <span className="AddonMoreInfo-license-type">{license.name}</span>
+      );
+    }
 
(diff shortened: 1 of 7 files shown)


willdurand:
> originally I updated just the AddonMoreInfo component but I this might be better to update in the Link...

Why not? I don't think we want to allow `href` with a `null` value.

rebmullin:
> Why not? I don't think we want to allow href with a null value.

I guess I was thinking we might want to keep it this way to protect other cases where this might happen but I really wasn't sure. I'll revert this and recommit the update in the AddonMoreInfo component..

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -91,8 +91,25 @@ export class AddonMoreInfoBase extends React.Component<Props> {
       );
     }
 
-    const currentVersion = addon.current_version;
     const lastUpdated = addon.last_updated;
+    const currentVersion = addon.current_version;
```
willdurand: I would have left this line where it was before.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -91,8 +91,25 @@ export class AddonMoreInfoBase extends React.Component<Props> {
...
+    const { license } = currentVersion;
+    let licenseHtml = null;
```
willdurand: You could probably name this variable `versionLicenseLink`, no?

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -185,6 +185,31 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      current_version: {
+        ...fakeAddon.current_version,
+        license: { name: 'justText', url: null },
+      },
+    });
+    const root = render({ addon });
+    const link = root.find('.AddonMoreInfo-license-type');
+    expect(link.children()).toIncludeText('justText');
```
willdurand: Where do you check that it is not a link?

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -185,6 +185,31 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      current_version: {
+        ...fakeAddon.current_version,
+        license: null,
+      },
+    });
+    const root = render({ addon });
+    expect(root.find('.AddonMoreInfo-license')).toHaveLength(0);
+  });
```
willdurand: I am surprised that this test case did not exist before.

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -91,8 +91,25 @@ export class AddonMoreInfoBase extends React.Component<Props> {
...
+        <Link
+          className="AddonMoreInfo-license-link"
+          href={license.url}
+          prependClientApp={false}
+          prependLang={false}
+        >
+          {license.name}
+        </Link>
+      ) : (
+        <span className="AddonMoreInfo-license-type">{license.name}</span>
```
willdurand: Should it be `type` or `name`?

src/amo/components/AddonMoreInfo/index.js
rebmullin: well it's not always a link, right? so I thought this was more fitting...

src/amo/components/AddonMoreInfo/index.js
rebmullin: yeh I guess name : )

tests/unit/amo/components/TestAddonMoreInfo.js
rebmullin: Yep, i'll add another case : ) thanks

src/amo/components/AddonMoreInfo/index.js
willdurand: `versionLicenseLink` is the key used a few line below.

rebmullin: Altho, I see that's what we are calling it in the function..

rebmullin: okay I see..I'll just rename this

tests/unit/amo/components/TestAddonMoreInfo.js
rebmullin: yeh me too..

Issue #2402, opened 2017-05-18
Automatically restart the dev server when config files change


kumar303:
When any config file changes, the dev server needs to be restarted. We already have a file watcher so let's try to do this automatically.

bobsilverberg:
This is something I think is worth still trying to do. I did a bit of digging through related issues/PRs and it looks like https://github.com/mozilla/addons-frontend/pull/5777 supposedly fixed this, but as far as I can tell it isn't fixed. Any changes to config files still require a manual restart of the server.

Are others seeing this as fixed, and if not is this still something we want to try to do?

PR #5777, opened 2018-07-30
Reload server when source files change

### LINKED ISSUE #5530: Automatically reload dev server code on every change
We use a hot reloader but it only applies to client side code. The server code does not hot reload. It should!

Here's why:
- First, you really need to develop with SSR enabled since it catches so many bugs
- React [reconciliation](https://reactjs.org/docs/reconciliation.html) has a lot of optimizations now which means our dev environment goes quickly out of sync until you manually restart the server
- Restarting the server is very slow
- Switching code branches or doing a yarn update is especially annoying

This issue will probably be harder than https://github.com/mozilla/addons-frontend/issues/2402 but maybe not.

willdurand:
Fix #5530

---

After having looked at the Internet to know what other people do, I
started to play with the webpack `require.cache` object. That was the
second option after "nodemon" (a tool to restart a node daemon, which is
a no-go for our setup given that it takes ages to boot the whole amo
app).

This `require.cache` object persists an `import` in memory to avoid
reading a source file all the time. The intuition was to clear this
cache every time a source file changed so that webpack would require
this file again, on the server. The problem is that the webpack config
(and webpack server) is located in `bin/webpack-dev-server.js`, which is
not the same file (and therefore process) as the "server code".

I tried to change the `runServer()` function (in `core/server/base.js`)
to stop/start a server after a change, but this did not work either. In
fact, the Isomorphic webpack plugin does not like to be restarted. That
did not work as expected.

At some point, I saw that every time I was editing `bin/server.js`, this
file would get reloaded, and it was super-fast! I tried to make some
changes in the source code, saved this `bin/server.js`, and boom! The
server was serving the new source code! :tada:

That's why I ended up writing this patch. A file watcher watches all the
files in `src/` and triggers a (soft) server reload. Last but not least,
it does not seem to break Hot Module Replacement (the client does not
loose the websocket connection).

## About #2402

This technique would probably work too. If I add `config/` in the list of
directories watched by the watcher, the server code would be able to use
the latest edited version of the config (I tried for instance to change
the value of `allowErrorSimulation`). The problem is that the client
code does not get reloaded when the config changes, leading to a
difference between what the server and client render.

I think I could use a watcher in `bin/webpack-dev-server.js` too, and
clear `require.cache` when the config is updated. Not sure, but that
might do the trick, but that'll be for another PR!


bin/server.js
```diff
@@ -22,4 +25,18 @@ if (config.util.getEnv('NODE_ENV') === 'development') {
...
+require('core/server/base').runServer().then(() => {
+  if (config.get('isDevelopment')) {
+    const basePath = config.get('basePath');
+    const watcher = chokidar.watch(path.join(basePath, 'src'));
+
+    watcher.on('ready', () => {
+      console.log('👀 The file watcher for server hot reload is ready.');
+
+      watcher.on('all', () => {
+        // Touching this file triggers a server reload.
```
kumar303: This is pretty non-intuitive (I have no idea how it works). Could you elaborate on this comment? Or is it magical to you, too?

willdurand: We require `piping` a few lines above, I don't really know why this is there but it is super useful to reload a node server obviously: https://github.com/mdlawson/piping.

PR #5778, opened 2018-07-30
Prevent the test suite from ever calling a real API

### LINKED ISSUE #1138: Prevent the test suite from ever calling a real API
- write a test for an action that calls the API
- forget to mock out the `callApi` method
### What happened?

This will attempt to make a real API request (depending on what server is configured). The result will probably be a harmless 401 Unauthorized since no real tokens are used in the test suite but this can cause slowness.

It should not make any real API request at all even though it was a mistake not to mock the component.

willdurand:
Fix #1138

---

I tested by removing all the "mock" code of a saga test suite and in
some API test cases in the disco test suite.

### DIFF
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -73,6 +73,11 @@ global.sinon = sinon.createSandbox();
 // Stub the magic constant webpack normally supplies.
 global.CLIENT_CONFIG = require('core/utils').getClientConfig(config);
 
+// See: https://github.com/mozilla/addons-frontend/issues/1138
+global.fetch = () => {
+  throw new Error('API calls MUST be mocked.');
+};
+
 afterEach(() => {
   global.sinon.restore();
 });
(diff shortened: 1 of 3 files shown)


tests/setup.js
```diff
@@ -73,6 +73,11 @@ global.sinon = sinon.createSandbox();
 // Stub the magic constant webpack normally supplies.
 global.CLIENT_CONFIG = require('core/utils').getClientConfig(config);
 
+// See: https://github.com/mozilla/addons-frontend/issues/1138
+global.fetch = () => {
+  throw new Error('API calls MUST be mocked.');
```
kumar303: It would be nice if this message included the URL that was getting fetched since we don't always get a stack trace, especially in sagas. Seeing the URL should help track it down. I guess you'll have to look for `input.url || input` in [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Syntax).

tests/unit/amo/sagas/test_collections.js
```diff
@@ -127,6 +127,11 @@ describe(__filename, () => {
         .once()
         .returns(Promise.reject(error));
 
+      mockApi
+        .expects('getCollectionAddons')
```
kumar303: 👀

PR #5779, opened 2018-07-30
Switch to firefox dev edition

willdurand:
Refs #5781 

---

This PR changes the Firefox version from nightly to dev edition in order
to get a green feedback again. The issue is related to a "menu" button
not being found anymore. All FF versions < 63 are not affected.

I added a job for nightly so that we can keep an eye on it, but this job
won't fail the build at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,27 @@ jobs:
     - stage:
       <<: *tox
       addons:
-        firefox: latest-nightly
+        firefox: latest-dev
         hosts: example.com
       env: TOXENV=discopane-ui-tests GECKODRIVER=0.20.1 MOZ_HEADLESS=1 PATH=./node_modules/.bin:$PATH
+      install:
+        - yarn
+        - yarn start-func-test-server &
+      before_script:
+        - wget -O /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/releases/download/v$GECKODRIVER/geckodriver-v$GECKODRIVER-linux64.tar.gz
+        - mkdir $HOME/geckodriver && tar xvf /tmp/geckodriver.tar.gz -C $HOME/geckodriver
+        - export PATH=$HOME/geckodriver:$PATH
+        - firefox --version
+        - geckodriver --version
+        - pip install tox
+        # Wait for server to be available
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -70,9 +70,32 @@ jobs:
         - pip install tox
         # Wait for server to be available
         - until $(curl --output /dev/null --silent --head --fail -k https://example.com:4000); do printf '.'; sleep 5; done
+    - stage:
+      <<: *tox
+      addons:
+        firefox: latest-nightly
+        hosts: example.com
+      env: ALLOWED_TO_FAIL=1 TOXENV=discopane-ui-tests GECKODRIVER=0.20.1 MOZ_HEADLESS=1 PATH=./node_modules/.bin:$PATH
```
kumar303: If this is allowed to fail, no one will care about it. I don't think it will be helpful to add an additional stage here.

.travis.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     - stage:
       <<: *tox
       addons:
-        firefox: latest-nightly
+        firefox: latest-dev
```
kumar303: I think this is fine as a temporary fix. I would like to see `latest-nightly` commented out with a link to a bug for re-enabling it. We should unblock master then investigate why it's failing on Nightly and try to get back on Nightly.

.travis.yml
willdurand: Well, I will care about it. Otherwise, how do we know if this gets fixed on Nightly?

kumar303: > Otherwise, how do we know if this gets fixed on Nightly?

The person looking into fixing the tests for Nightly would be running it locally.

But, sure, that's fine if you want to leave in a new stage temporarily while someone works on a fix.

.travis.yml
kumar303: Is there a bug on file for putting the tests back on Nightly? I'd like to see a comment here with that bug link.

willdurand: I opened https://github.com/mozilla/addons-frontend/issues/5781

willdurand:
There is a new issue on `master` but unrelated to this patch, merging...

PR #5787, opened 2018-07-30
I5703 fixing scaled icon flash

### LINKED ISSUE #5703: A glitch appears when opening a theme listing page

rebmullin:
fixes #5703 

From what I could tell the url was changed in the src *after* the classname was updated and for a split second you see the icon large b/c it takes on the new classname styles:

Alt text

In the header, the only way that I found that fixes this is to change the html structure or to use a key on the image/wrapping div. This might be another reconciliation issue (?). I decided to go the html approach - changing out a div for a span. let me know what you think..

note: i did a little bit of clean up too : )

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -164,13 +164,10 @@ export class AddonBase extends React.Component {
         previewURL = addon.previewURL;
       }
 
-      const themeClassName = 'Addon-theme-header';
-
       return (
         <div
+          className="Addon-theme-header"
           key="Addon-theme-image"
-          className={themeClassName}
-          id={themeClassName}
           role="presentation"
         >
           <img
@@ -183,7 +180,7 @@ export class AddonBase extends React.Component {
     }
 
     return (
-      <div className="Addon-icon">
+      <div className="Addon-icon" key="Addon-icon-header">
         <img
           alt={label}
           className="Addon-icon-image"
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -150,7 +150,7 @@ export class AddonBase extends React.Component {
       ? i18n.sprintf(i18n.gettext('Preview of %(title)s'), {
           title: addon.name,
         })
-      : '';
+      : null;
 
     if (addon && isTheme(addon.type)) {
       const previewHeader = addon.previews.length && addon.previews[0];
@@ -167,7 +167,12 @@ export class AddonBase extends React.Component {
       const themeClassName = 'Addon-theme-header';
 
       return (
-        <div className={themeClassName} id={themeClassName} role="presentation">
+        <div
+          key="Addon-theme-image"
+          className={themeClassName}
+          id={themeClassName}
+          role="presentation"
+        >
           <img
             alt={label}
             className="Addon-theme-header-image"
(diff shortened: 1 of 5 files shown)


src/amo/components/Addon/index.js
```diff
@@ -154,43 +160,31 @@ export class AddonBase extends React.Component {
...
       );
     }
 
-    const iconUrl = getAddonIconUrl(addon);
-
     return (
-      <div className="Addon-icon">
-        <img className="Addon-icon-image" alt="" src={iconUrl} />
-      </div>
+      <span className="Addon-icon">
```
willdurand: Ah well, that's a bit annoying to have to do that but I understand. Would adding `key` work though? I'd like to use `key` over `span` because it is more the react way.

tests/unit/amo/components/TestAddon.js
```diff
@@ -298,6 +298,8 @@ describe(__filename, () => {
 
     // Since withInstallHelpers relies on this, make sure it's initialized.
     expect(root.instance().props.platformFiles).toEqual({});
+
+    expect(root.find('.Addon-icon img').prop('alt')).toEqual('');
```
willdurand: Could we set the value to `null` so that it would not render an empty `alt` attribute?

src/amo/components/Addon/index.js
rebmullin: yep seems to works too so I can switch : )

tests/unit/amo/components/TestAddon.js
rebmullin: oh I thought empty attribute was better than leaving it off but sure I can update

src/amo/components/Addon/index.js
```diff
@@ -154,42 +160,35 @@ export class AddonBase extends React.Component {
...
-      );
+      const themeClassName = 'Addon-theme-header';
 
       return (
         <div
-          className="Addon-theme-header"
-          id="Addon-theme-header"
+          key="Addon-theme-image"
+          className={themeClassName}
+          id={themeClassName}
```
willdurand: Do we really need an `id` here? I find it strange to use the `themeClassName`. If the `id` is not useful, let's remove it and also remove the `themeClassName`.

src/amo/components/Addon/index.js
```diff
@@ -154,42 +160,35 @@ export class AddonBase extends React.Component {
...
+            src={previewURL}
+          />
         </div>
       );
     }
 
-    const iconUrl = getAddonIconUrl(addon);
-
     return (
       <div className="Addon-icon">
```
willdurand: Don't we need a key here too?

tests/unit/amo/components/TestSearchResult.js
```diff
@@ -177,6 +177,16 @@ describe(__filename, () => {
     expect(root.find('.SearchResult-icon')).not.toHaveProp('alt', '');
   });
 
+  it('renders a loading class name while there is no addon', () => {
+    const root = render({
```
willdurand: Could be put on a single line.

src/amo/components/Addon/index.js
rebmullin: Yeh I guess I just left it b/c it was there already. Sure I can remove this

src/amo/components/Addon/index.js
rebmullin: I didn't seem like it was needed but to be sure I can add this here too

PR #5792, opened 2018-07-31
Use cheap-module-eval-source-map in webpack dev

### LINKED ISSUE #4508: Source maps are not working on dev

Open the web console and trigger some kind of error. See https://github.com/mozilla/addons-frontend/issues/4507

### What happened?

The error points to the minified source

Source maps should point to the actual source

I tried in both Firefox and Chrome to be sure.

willdurand:
Fix #4508

---

This `devtool` value is the recommended one, and it is faster than the
previous one: https://webpack.js.org/configuration/devtool/.

From what I know, the main difference is that it does not link to the
column, only to the line (where an error has occurred for instance). I
don't think it really matters because we only have 80 columns anyway.

### DIFF
--- a/webpack.dev.config.babel.js
+++ b/webpack.dev.config.babel.js
@@ -44,7 +44,7 @@ for (const app of appsBuildList) {
 }
 
 export default Object.assign({}, webpackConfig, {
-  devtool: 'heap-module-eval-source-map',
+  devtool: 'cheap-module-eval-source-map',
   context: path.resolve(__dirname),
   entry: entryPoints,
   output: Object.assign({}, webpackConfig.output, {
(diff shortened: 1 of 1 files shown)


webpack.dev.config.babel.js
```diff
@@ -44,7 +44,7 @@ for (const app of appsBuildList) {
 }
 
 export default Object.assign({}, webpackConfig, {
-  devtool: 'inline-source-map',
+  devtool: 'heap-module-eval-source-map',
```
muffinresearch: Is this a typo? Title of the PR says "cheap" this says "heap".

willdurand: heh 🤕

willdurand: thanks!

PR #5793, opened 2018-07-31
Do not run tests twice after editing a file

### LINKED ISSUE #2639: `jest --watch` runs tests twice after editing a file

Run `yarn test`, edit a file.

### What happened?

Sometimes the tests are run twice

They should only be run once for each file change

This is happening because our [WebpackIsomorphicToolsPlugin](https://github.com/mozilla/addons-frontend/blob/65d4e802edad8b78f0df955e363ef38214179c5d/webpack.prod.config.babel.js#L52) auto-generates a `webpack-assets.json` file. The change to this file triggers a second test run.

It is not yet possible to [ignore changes to this file](https://github.com/facebook/jest/issues/3923) and I haven't found a good workaround yet.

willdurand:
Fix #2639

---

I find this behavior annoying and found the GH issue last night, so here is a patch.
I hooked into the Jest process to see which files were triggering a re-run.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -38,5 +38,9 @@ module.exports = {
     'jest-watch-typeahead/filename',
     'jest-watch-typeahead/testname',
   ],
+  watchPathIgnorePatterns: [
+    '<rootDir>/bin/server.js',
+    '<rootDir>/webpack-assets.json',
+  ],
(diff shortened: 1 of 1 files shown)


kumar303:
Thanks!

PR #5795, opened 2018-07-31
Fix CSS order

### LINKED ISSUE #3565: Normalize is included in the middle of the CSS bundle
Since our style is fairly heavily specified I'm not sure how much this is a problem but normalize is not currently included at the top of our stylesheet.

willdurand:
Fix #3565

---

Please read: https://github.com/mozilla/addons-frontend/issues/3565#issuecomment-409204307


muffinresearch:
Ah good detective work there!

doff

PR #5803, opened 2018-07-31
Rename denormalizeReview() to createInternalReview()

### LINKED ISSUE #3219: Rename denormalizeReview() to createInternalReview()

willdurand:
Fix #3219

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -40,7 +40,7 @@ export type UserReviewType = {|
   versionId: number | null,
 |};
 
-export function denormalizeReview(
+export function createInternalReview(
   review: ExternalReviewType | ExternalReviewReplyType,
 ): UserReviewType {
   return {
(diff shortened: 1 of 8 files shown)


kumar303:
🌈 🍉 🍭 🍓

PR #5804, opened 2018-07-31
Move createInternalReview() into the reduce

### LINKED ISSUE #3342: Move denormalizeReview() / createInternalReview() into the reducer
The `denormalizeReview()` function (which will get renamed to `createInternalReview()` in https://github.com/mozilla/addons-frontend/issues/3219) is called in an action creator. It is bad practice to add this kind of logic to action creators because action functions are not replayable. In other words, hypothetically, we would not be able to reproduce a user reported bug in this function. We should always move logic like this to the reducer.

willdurand:
Fix #3342

---

Do not use `createInternalReview()` into an action creator.


src/amo/reducers/reviews.js
```diff
@@ -313,6 +316,21 @@ export default function reviewsReducer(
     }
     case SET_REVIEW: {
       const { payload } = action;
+      const review = createInternalReview(payload);
+
+      return {
+        ...state,
+        byId: storeReviewObjects({ state, reviews: [review] }),
+        byUserId: {
```
kumar303: It's a shame that this is a duplication of the `SET_INTERNAL_REVIEW` case. However, I don't see a clean way to DRY it up. People complain about how Redux paradigms can be hard to encapsulate in abstractions and this is one such example.

Unless you have an idea for how to abstract this cleanly, you'll need to add test coverage for the `SET_INTERNAL_REVIEW` case by copying [this test](https://github.com/mozilla/addons-frontend/blob/5cdc08bf0cccb083f487aed8de70e449b37cd0b6/tests/unit/amo/reducers/testReviews.js#L109). If I comment it out in your current patch, the tests pass.

tests/unit/amo/components/TestAddonReview.js
```diff
@@ -444,20 +443,5 @@ describe(__filename, () => {
...
       });
     });
-
-    describe('setDenormalizedReview', () => {
-      it('dispatches a setReview action', () => {
-        const review = {
-          ...defaultReview,
-          body: 'some body',
-        };
-        actions.setDenormalizedReview(review);
```
kumar303: We still need this test because it covers how the component [exposes a function](https://github.com/mozilla/addons-frontend/blob/5cdc08bf0cccb083f487aed8de70e449b37cd0b6/src/amo/components/AddonReview/index.js#L283) for dispatching the action.

It's odd way to write a component. I believe it was originally done this way for ease of testing in a world that predated `shallowUntilTarget` (where `sinon.stub(store, 'dispatch')` was not an option).

src/amo/reducers/reviews.js
willdurand: I could extract the common logic into a function at least.

I tried to use the same action, looking for `userId` in the `payload` to determine whether `createInternalReview()` was needed, but Flow complained.

src/amo/reducers/reviews.js
```diff
@@ -227,12 +229,32 @@ export const selectLatestUserReview = ({
   return selectReview(reviewsState, userReviewId);
 };
 
+// This is used in the SET_REVIEW and SET_INTERNAL_REVIEW cases.
```
kumar303: Do you think this comment is necessary? Usually when I want to see where a function is used I just search for it. Comments like tend to go stale quickly, too.

src/amo/reducers/reviews.js
```diff
@@ -227,12 +229,32 @@ export const selectLatestUserReview = ({
...
+const addReviewInState = ({
```
kumar303: nit: this would read a bit better as `addReviewToState`

src/amo/reducers/reviews.js
```diff
@@ -313,15 +335,14 @@ export default function reviewsReducer(
...
-        byId: storeReviewObjects({ state, reviews: [payload] }),
-        byUserId: {
-          ...state.byUserId,
-          // This will trigger a refresh from the server.
-          [payload.userId]: undefined,
-        },
-      };
+      const review = createInternalReview(payload);
+
+      return addReviewInState({ state, review });
```
kumar303: Actually, yeah, this abstraction totally works 😃 I'm not sure why I thought it wouldn't.

tests/unit/amo/reducers/testReviews.js
```diff
@@ -124,6 +132,32 @@ describe(__filename, () => {
     expect(state.byUserId[userId]).not.toBeDefined();
   });
 
+  it('resets the byUserId data when adding an internal review', () => {
```
kumar303: When you factor out a shared piece of logic (e.g. `addReviewInState()`) you have two options:
1. Duplicate the tests to cover the shared logic in every place you depend on it
2. Add unit tests for the new shared function and use a sinon mock to *guarantee* the function is getting called correctly

You have taken the duplication approach here but I'd rather see you move to the second option.

First, you'll need to set it up for dependency injection:

```diff
diff --git a/src/amo/reducers/reviews.js b/src/amo/reducers/reviews.js
index 0b629a179..b11445a46 100644
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -230,7 +230,7 @@ export const selectLatestUserReview = ({
 };
 
 // This is used in the SET_REVIEW and SET_INTERNAL_REVIEW cases.
-const addReviewInState = ({
+export const addReviewInState = ({
   state,
   review,
 }: {|
@@ -266,6 +266,9 @@ type ReviewActionType =
 export default function reviewsReducer(
   state: ReviewsState = initialState,
   action: ReviewActionType,
+  {
+    _addReviewInState = addReviewInState,
+  }: {| _addReviewInState: typeof addReviewInState |} = {},
 ) {
   switch (action.type) {
     case SEND_REPLY_TO_REVIEW:
@@ -337,12 +340,12 @@ export default function reviewsReducer(
       const { payload } = action;
       const review = createInternalReview(payload);
 
-      return addReviewInState({ state, review });
+      return _addReviewInState({ state, review });
     }
     case SET_INTERNAL_REVIEW: {
       const { payload } = action;
 
-      return addReviewInState({ state, review: payload });
+      return _addReviewInState({ state, review: payload });
     }
     case SET_REVIEW_REPLY: {
       const reviewId = action.payload.originalReviewId;
```

Next, add unit tests for `addReviewInState()` and give yourself two test cases for `SET_REVIEW` and `SET_INTERNAL_REVIEW` with mock functions to make sure it's used.

willdurand: 😞

tests/unit/amo/components/TestAddonReview.js
```diff
@@ -445,17 +445,17 @@ describe(__filename, () => {
...
           ...defaultReview,
           body: 'some body',
         };
-        actions.setDenormalizedReview(review);
+        actions.setInternalReview(review);
 
         sinon.assert.called(dispatch);
         const action = dispatch.firstCall.args[0];
-        expect(action.type).toEqual(SET_REVIEW);
+        expect(action.type).toEqual(SET_INTERNAL_REVIEW);
```
kumar303: You could clean this up while you're in here but no big deal.

```js
sinon.assert.calledWith(dispatch, setInternalReview(review));
```

willdurand: ah yes, makes sense

PR #5811, opened 2018-08-01
Use uglifyjs-webpack-plugin

### LINKED ISSUE #5796: Upgrade uglify to cope with es6 code
The uglify we are currently using cannot handle non-es5 code. 

To fix this we will need to use uglify-es in place of uglify. This is only a problem if a npm package contains non-transpiled code.

We previously made this change in https://github.com/mozilla/addons-frontend/pull/4868 but it had to be reverted because webpack produced hashes that were the same even though the content of the files changed.

We'll need to understand why this occurred and we might need to make a change to ensure that hashes are different - a provision for that is here if we need it as a last resort:

https://github.com/mozilla/addons-frontend/blob/bcae92c4ce3bef0cdbfdfec708d7c667d998f139/bin/build-locales#L83

willdurand:
Fix #5796

---

I tried to build the `amo` app and could not get the same hashes when
introducing a change in the source code. I was not able to reproduce
the original issue, but I am optimistic :smile:

This PR will allow us to compile not-only-ES5 code!

### DIFF
--- a/webpack.prod.config.babel.js
+++ b/webpack.prod.config.babel.js
@@ -4,7 +4,6 @@ import path from 'path';
 import config from 'config';
 import ExtractTextPlugin from 'extract-text-webpack-plugin';
 import SriPlugin from 'webpack-subresource-integrity';
-import webpack from 'webpack';
 import WebpackIsomorphicToolsPlugin from 'webpack-isomorphic-tools/plugin';
 import UglifyJsPlugin from 'uglifyjs-webpack-plugin';
 
--- a/webpack.prod.config.babel.js
+++ b/webpack.prod.config.babel.js
@@ -6,6 +6,7 @@ import ExtractTextPlugin from 'extract-text-webpack-plugin';
 import SriPlugin from 'webpack-subresource-integrity';
 import webpack from 'webpack';
 import WebpackIsomorphicToolsPlugin from 'webpack-isomorphic-tools/plugin';
+import UglifyJsPlugin from 'uglifyjs-webpack-plugin';
 
 import SriDataPlugin from './src/core/server/sriDataPlugin';
 import { getPlugins, getRules } from './webpack-common';
@@ -45,11 +46,14 @@ export default {
       allChunks: true,
     }),
     // optimizations
-    new webpack.optimize.UglifyJsPlugin({
-      sourceMap: true,
-      comments: false,
-      compress: {
-        drop_console: true,
+    new UglifyJsPlugin({
+      uglifyOptions: {
+        output: {
+          comments: false,
+        },
+        compress: {
+          drop_console: true,
+        },
       },
     }),
     new WebpackIsomorphicToolsPlugin(webpackIsomorphicToolsConfig),
--- a/webpack.prod.config.babel.js
+++ b/webpack.prod.config.babel.js
@@ -46,6 +46,8 @@ export default {
     }),
     // optimizations
     new UglifyJsPlugin({
+      // Do not change these options without busting the cache.
+      // See: https://github.com/mozilla/addons-frontend/issues/5796
       uglifyOptions: {
         output: {
           comments: false,
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -80,7 +80,7 @@ for (const appName of appsBuildList) {
 
     // This is used to force cache-busting of the JS via changing all locale JS.
     // DO NOT REMOVE! See https://github.com/mozilla/addons-frontend/issues/4927
-    localeObject._cacheBust = '2018.05.04';
+    localeObject._cacheBust = '2018.08.02';
 
     fs.writeFileSync(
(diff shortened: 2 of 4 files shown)


muffinresearch:
For "belt and braces" would it be worth bumping the date for cache-busting purposes? Last time the problems didn't show up until production and then it was too late. 

Did you also check the hashes vs contents of the locale files?

willdurand:
> For "belt and braces" would it be worth bumping the date for cache-busting purposes? Last time the problems didn't show up until production and then it was too late.

Does that mean -dev and -stage were not affected? If so, then I'd like to refrain my positivity. What does this date do exactly?

> Did you also check the hashes vs contents of the locale files?

I did and hashes were different.

muffinresearch:
> Does that mean -dev and -stage were not affected? If so, then I'd like to refrain my positivity. What does this date do exactly?

I added a link in my comment above - see https://github.com/mozilla/addons-frontend/blob/3701a9a3afb85a3571435dc97c4f28d9f7a07efa/bin/build-locales#L83 - by modifying this it causes the file-name to be different on all JS because the core files include the locale references. Changing the object in the locale also changes the content of every locale file too, so it's a hack that quickly ensures the hashes will be different for every JS file.

Updating that might be worth doing since it should exclude any possibility of the hashes being wrong and a client receiving cached statics. That said if you can see a place where they fixed something between the time we had this issue and if we're very confident this might be unnecessary - however, the way I see it there's no real problem in busting the cache this way especially if this update looks to be changing all the hashes anyway.

willdurand:
> Did you also check the hashes vs contents of the locale files?

STR:

1. checkout the `uglify` branch
2. `yarn install`
3. `NODE_APP_INSTANCE=amo NODE_ENV=production yarn build`
4. it will build the app in `dist/`, save the list of files somewhere
5. Now make a change in a locale file, and then repeat 3. and 4.

```
// version without a change
-rw-r--r-- 1 williamdurand  53K Aug  1 18:07 dist/amo-i18n-fr-b27f34a7d45d720bb6a7.js

// version with a change
-rw-r--r-- 1 williamdurand  53K Aug  1 18:10 dist/amo-i18n-fr-02ba1110a58b16afc47c.js
```

willdurand:
> - by modifying this it causes the file-name to be different on all JS because the core files include the locale references. Changing the object in the locale also changes the content of every locale file too, so it's a hack that quickly ensures the hashes will be different for every JS file.
> 
> Updating that might be worth doing since it should exclude any possibility of the hashes being wrong and a client receiving cached statics. That said if you can see a place where they fixed something between the time we had this issue and if we're very confident this might be unnecessary - however, the way I see it there's no real problem in busting the cache this way especially if this update looks to be changing all the hashes anyway.

What if we remove this magic line? That should also change the hashes, no?

muffinresearch:
> What if we remove this magic line? That should also change the hashes, no?

It would but I'd be inclined to leave it in case we need it again in the future.

willdurand:
ohhhh I see.

One question though: if the dist content changes but the source is exactly the same, why would it be a problem? It could be due to some optimizations for instance. For example, if the hash is calculated based on the source file (that would be weird and it is unlikely but not so much an issue).

muffinresearch:
> One question though: if the dist content changes but the source is exactly the same, why would it be a problem? It could be due to some optimizations for instance. For example, if the hash is calculated based on the source file (that would be weird and it is unlikely but not so much an issue).

The problem before was that it broke SRI (subresource integrity) because the cached files were loaded but didn't match the new SRI hashes, and this caused some of the webpack code to throw because it couldn't load files. So it's very important we don't ship something with same webpack generated filenames which contains modified content.

willdurand:
> The problem before was that it broke SRI (subresource integrity) because the cached files were loaded but didn't match the new SRI hashes, and this caused some of the webpack code to throw because it couldn't load files. So it's very important we don't ship something with same webpack generated filenames which contains modified content.

Ah ok, I understand now.

willdurand:
@muffinresearch so shasum are different before and after this patch, but filenames remain the same.

I believe we would get the same outcome if we wanted to change the uglify config though. Not sure if there is a real issue there, besides the SRI you mentioned. What I mean is that if we want to upgrade to uglify-es and/or change the uglify config, we'll have to invalidate the cache manually (using a new `_cacheBust` date).

muffinresearch:
@willdurand yep I think it's fine as long as we bump the cache-bust date. As you say presumably the webpack filenames are derived from the source which is why this happens.

webpack.prod.config.babel.js
```diff
@@ -45,11 +45,14 @@ export default {
       allChunks: true,
     }),
     // optimizations
-    new webpack.optimize.UglifyJsPlugin({
-      sourceMap: true,
-      comments: false,
-      compress: {
-        drop_console: true,
+    new UglifyJsPlugin({
```
muffinresearch: Would be good to add a comment here saying not to make changes without busting the cache.

PR #5823, opened 2018-08-02
Fix pathname in SearchFilters

### LINKED ISSUE #5822: Filtering is not working for Search Tools
Steps to reproduce:
1.Go to AMO Stage and from "More" dropdown menu, select Search Tools
2.Try to apply any filter from the "Filter Results" section

Expected results:
All filters are working

Actual results:
No filter is applied for Search Tools

NOTES:
Not reproducible in Prod
Not reproducible when filtering themes or extensions
 
Please see a video with the bug: 
searchfilter

willdurand:
Fix #5822

---

I reverted some changes made in #5652 (RRv4 upgrade) because that's the
root cause of #5822. The `SearchFilters` must navigate to the search
page from the "search tools" page.

### DIFF
--- a/tests/unit/amo/components/TestSearchFilters.js
+++ b/tests/unit/amo/components/TestSearchFilters.js
@@ -25,18 +25,19 @@ import {
 describe(__filename, () => {
   let fakeHistory;
 
-  function render({
-    filters = {},
-    pathname = '/en-US/android/search/',
-    ...props
-  } = {}) {
+  function render({ filters = {}, pathname = '/search/', ...props } = {}) {
     const errorHandler = createStubErrorHandler();
     const { store } = dispatchClientMetadata();
 
     store.dispatch(searchStart({ errorHandlerId: errorHandler.id, filters }));
 
     return shallowUntilTarget(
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestSearchFilters.js
```diff
@@ -25,18 +25,19 @@ import {
 describe(__filename, () => {
   let fakeHistory;
 
-  function render({
-    filters = {},
-    pathname = '/en-US/android/search/',
-    ...props
-  } = {}) {
+  function render({ filters = {}, pathname = '/search/', ...props } = {}) {
```
muffinresearch: Is there a reason for removing the clientApp and lang from pathname here?

willdurand: Yes, this prop name is a bit misleading: it is a developer-defined prop called pathname, not the pathname from the routing layer. It should be called something like "the last part of the pathname".

It is supplied via the `Search` component, which defines the default value:  https://github.com/mozilla/addons-frontend/blob/82c44dad7d818dadc60d17e9b93e88088edd4779/src/amo/components/Search/index.js#L69

PR #5826, opened 2018-08-02
Use Pino for logging

### LINKED ISSUE #2423: Unify logging
I'm not sure how much value there is in special server vs client-side logging. 

Let's decide whether this is still needed and consolidate on a single logging solution that works the same in node and browsers.

This is best done after the change to use Jest.

willdurand:
Fix #2423

---

Use `pino` instead of bunyan.

I think we should do something to the `logRequests()` middleware because
it makes everything unreadable in the server logs... I don't think we
need to log both the full `req` and `res` objects. => changed to `log.debug()`

~~I added the ability to disable logging _via_ the config, which is useful
for the test suite :)~~


willdurand:
`yarn.lock` conflicts but that's not an issue.

src/core/client/logger.js
```diff
@@ -1,58 +0,0 @@
...
-  const app = `[${appName}]`;
-
-  if (_function.prototype.bind) {
-    consoleFunc = _function.prototype.bind.call(
-      consoleMethod,
-      _consoleObj,
-      app,
-    );
-  } else {
-    // Fallback for IE < 10;
```
kumar303: Do you know what browsers pino supports? I can't find anything definitive on their readme.

config/test.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   allowErrorSimulation: true,
   enableClientConsole: true,
   isDeployed: false,
+  enableLogging: false,
```
kumar303: I don't think this is helpful because you can already run the tests with `yarn test --silent`.

I would rather [patch up Jest](https://github.com/facebook/jest/issues/4156) to expose its DefaultReporter so we can write a reporter that only shows logging for failing tests.

src/core/client/logger.js
willdurand: heh, looks like IE 10 might not work: https://github.com/pinojs/pino/issues/311.

config/test.js
willdurand: oh really? ok, let's remove it then.

src/core/client/logger.js
willdurand: I think it works for IE10 and more: https://github.com/pinojs/pino/pull/323, which looks ok to me (according to Analytics, we had 0.36% users on Internet Explorer from January 1st, 2018).

willdurand:
> I no longer see debug messages in the client (example). Could you put back the debug messages? It would be great if the log level was configurable. After that, we could turn off debug messages by default. I still want to see them when I need them, though.

ahm, default level is `info` indeed. Let's use `debug`. How would you like to configure the level? `config/*`? Also, I guess this has changed because the current client logger was homemade, right?

> As for the logRequests() middleware, could we switch this to log.debug() so that it's hidden by default? I think that would be nice.

Yep, I can do that (although I wrote https://github.com/willdurand/pino-devtools and this is nicely displayed now).

kumar303:
> How would you like to configure the level? `config/*`? 

Yes, I just meant in the config files.

> ... I wrote https://github.com/willdurand/pino-devtools and this is nicely displayed now

That's cool! I would like to use that. In the screenshot I see `<no message> Object { req: {...}, res: {...}...` This is nicer because it's shorter but this still looks like noise to me. It's not so helpful. We might as well make those `debug` messages.

bin/proxy.js
```diff
@@ -4,15 +4,13 @@ const fs = require('fs');
...
 const config = require('config');
 const cookie = require('cookie');
 const httpProxy = require('http-proxy');
+const pino = require('pino');
 
-const log = bunyan.createLogger({
-  name: 'proxy',
-  app: config.get('appName'),
-  serializers: bunyan.stdSerializers,
+const log = pino({
```
kumar303: Could you pass `level` here from the config?

I think it might be better to change the [request logging](https://github.com/mozilla/addons-frontend/blob/0e9d837c326c446145959607704ed9f68899690a/bin/proxy.js#L63) to a debug message. What do you think?

src/core/middleware/logRequests.js
```diff
@@ -7,5 +7,5 @@ export function logRequests(req, res, next) {
   next();
   const finish = new Date();
   const elapsed = finish - start;
-  log.info({ req, res, start, finish, elapsed });
+  log.debug({ req, res, start, finish, elapsed });
```
kumar303: 🌈 

When you actually turn on debug, though, the server log is pretty much useless. Ha. I am really not sure about this middleware.

bin/proxy.js
willdurand: > I think it might be better to change the request logging to a debug message. What do you think?

@kumar303 I like to have those logs, I read them often. We could switch them to `debug`, but then we should probably have `loggingLevel: 'debug'` for the `development` env?

willdurand: I decided to turn the debug logs on in `development` because it keeps compatibility with the current behavior and it really makes sense to have such logs in development. Otherwise, we'll never seen/read them, never. And I'd be mad if I had to restart the server to read a debug log.

kumar303: > I like to have those logs, I read them often

OK, I was on the fence about it. Let's keep them as `info`.

> I decided to turn the debug logs on in development because it keeps compatibility with the current behavior

Sure. I find it a bit noisy so I'll probably set it to `info` locally and see how it goes.

> I'd be mad if I had to restart the server to read a debug log.

That's the age old logging problem 😁 I find that *most* bugs are reproducible so I don't mind switching into debug mode only when needed. For that one time the bug is not reproducible it will be annoying, yes.

PR #5827, opened 2018-08-02
[l10n][az] Machine translation for AMO Frontend (manually verified)

mastizada:
Machine Translation of locale files for Azerbaijani locale.

App matches already translated strings in other projects in Pontoon and when match is 100% adds it to the po file.

All translations are manually verified.


mastizada:
Only nodejs related test failed, must be something is broken in app. Django related tests (po is related to django) passed.

kumar303:
Sorry about the error in TravisCI. This is unrelated and has been fixed on the master branch.

Thanks for your patch.

Issue #5832, opened 2018-08-02
Add saga and reducer for add-on star rating summaries


kumar303:
We need a saga to fetch star rating summaries and a reducer for storing them. This will support https://github.com/mozilla/addons-frontend/issues/5437

Issue #5437, opened 2018-06-27
Show star rating breakdown on detail page


pwalm:
Show the star rating breakdown at the top of the detail page. Example:
<img width="1005" alt="screen shot 2018-06-27 at 11 28 25 am" src="https://user-images.githubusercontent.com/6797635/41992487-715053b6-79fd-11e8-9892-ca2a1ed29c88.png">

Spec:
<img width="1507" alt="screen shot 2018-06-27 at 11 28 32 am" src="https://user-images.githubusercontent.com/6797635/41992495-76964d26-79fd-11e8-84b6-11dff1a40ef6.png">

ioanarusiczki:
I filed https://github.com/mozilla/addons-frontend/issues/5846 and https://github.com/mozilla/addons-frontend/issues/5847 for the scenarios when the user cancels the given rating.

@kumar303 What I've noticed about the layout:
- the spec above mentioned a Fira Sans-Regular font, I have with Inspector a font-family: Fira Sans,sans-serif font (Could be the same but I'm not sure)
- the spec mentioned 15 px font size for the Ratingsbystar-count, I see it as 14 px with Inspector
- the length of the bar is 315.45 px and I see above that it should have 306px

font size
![font size](https://user-images.githubusercontent.com/33448286/43781955-d318e8ae-9a66-11e8-9b89-48e5f3ea882d.png)
bar lenght
![lenght of the bar](https://user-images.githubusercontent.com/33448286/43781825-808c781c-9a66-11e8-8c66-f6770ba0af70.png)


Verified on AMO dev with FF61(Win10 & Android 8.0).

kumar303:
Thanks for the detailed observations. I don't think any new issues are needed. Here are some responses:

> the spec above mentioned a Fira Sans-Regular font, I have with Inspector a font-family: Fira Sans,sans-serif font

Yep, they're the same. This is the default font we already use on the site.

> the spec mentioned 15 px font size for the Ratingsbystar-count, I see it as 14 px with Inspector

I chose 14px because it is the closest standard font size [we already have](https://github.com/mozilla/addons-frontend/blob/f3efa1158a7e0b9f36d36e25b378bd4abb19a827/src/core/css/inc/vars.scss#L21). I filed https://github.com/mozilla/addons-frontend/issues/5866 to update font sizes.

> the length of the bar is 315.45 px and I see above that it should have 306px

[...]

Issue #4529, opened 2018-03-09
New add-on version will not display a user rating/review assigned to a previous version


AlexandraMoga:
STR:

1. Log in to AMO and pick an extension to view its details page
2. Write a review for this add-on and make sure your review is posted
3. Submit a new version for this extension and approve it
4. Go to add-on details page again and check the rating section

Actual result:
There is no record of a previous rating left by your user (stars are not colored)

Expected result:
A user rating assigned to a previous version should not be cleared when a new add-on version is submitted (stars should be colored)

Notes: 
- if the user goes to add-on reviews page, he can see that a review he previously posted is listed
- reproduced on all AMO servers with FF58 WIn10x64 and Android 7.0

![image](https://user-images.githubusercontent.com/31961530/37215496-c2c75b6a-23c0-11e8-8b68-173a326886d4.png)

![image](https://user-images.githubusercontent.com/31961530/37215331-5464241e-23c0-11e8-8e68-779c5193e42a.png)

muffinresearch:
@jvillalobos is this expected/by design?

jvillalobos:
No, we shouldn't be filtering ratings by version.

bobsilverberg:
I'm not sure how we should address this. We currently record ratings/reviews for a particular add-on version. Is this issue suggesting that we want to stop doing that, and only have one review per add-on per user (i.e., not one review per version)? 

@jvillalobos @kumar303

jvillalobos:
> only have one review per add-on per user (i.e., not one review per version)?

That's what I think we should do, yes.

bobsilverberg:
I think this will mostly need to be dealt with on the back end then, although our code will have to change to not submit/request a version any longer.

@diox What do you think?

kumar303:
The backend API is designed for one review per version. We may be able to workaround this structure on the frontend, I'm not sure.

kumar303:
Related: we have some dead code that used to support the case of editing a review for a previous version. It no longer works and is not covered by any tests. https://codecov.io/gh/mozilla/addons-frontend/src/76640e28035b1581096b49ac6064cef37635ad4f/src/amo/components/RatingManager/index.js#L129

Fixing the code I pasted up above is not trivial. I'm just posting it here for reference.

bobsilverberg:
I think it would be best to deal with it on the backend, which is why I pinged @diox. If we have to deal with it on the frontend we might be able to do that, but we should investigate a backend fix first, imo.

[...]

PR #5833, opened 2018-08-02
Add saga and reducer for add-on star rating summaries

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5832

I envision that this will be used in a new `RatingSummary` component (see https://github.com/mozilla/addons-frontend/issues/5437). It will first check for rating summaries in state and then dispatch an action to fetch them if they don't exist.

Docs on the API for star rating summaries: http://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#list-ratings


src/amo/reducers/reviews.js
```diff
@@ -81,6 +85,9 @@ export type ReviewsState = {|
     // or null if one does not exist yet.
     [userIdAddonIdVersionId: string]: number | null,
   },
+  ratingSummary: {
+    [addonId: string]: RatingSummaryType,
```
willdurand: Why would `addonId` become a string here? It has been typed as `number` in `FetchRatingSummaryParams`.

src/amo/sagas/reviews.js
```diff
@@ -60,6 +69,41 @@ function* fetchReviews({
   }
 }
 
+function* fetchRatingSummary({
+  payload: { errorHandlerId, addonId },
+}: FetchRatingSummaryAction): Generator<any, any, any> {
```
willdurand: Why is Flow not complaining here? `type` is required in the `FetchRatingSummaryAction` definition.

src/amo/reducers/reviews.js
```diff
@@ -255,8 +269,10 @@ type ReviewActionType =
   | SetAddonReviewsAction
   | SetInternalReviewAction
   | SetLatestReviewAction
+  | SetRatingSummaryAction
   | SetReviewAction
   | SetReviewReplyAction
+  | SetUserReviewsAction
```
willdurand: 🔥

tests/unit/amo/sagas/testReviews.js
```diff
@@ -354,11 +434,10 @@ describe(__filename, () => {
 
       _fetchUserReviews();
 
-      const errorAction = errorHandler.createErrorAction(error);
-      await sagaTester.waitFor(errorAction.type);
+      const expectedAction = errorHandler.createErrorAction(error);
+      const action = await sagaTester.waitFor(expectedAction.type);
```
willdurand: that's cool!

src/amo/reducers/reviews.js
```diff
@@ -244,6 +252,12 @@ export const addReviewToState = ({
       // This will trigger a refresh from the server.
       [review.userId]: undefined,
     },
+    ratingSummary: {
+      ...state.ratingSummary,
+      // When adding a new rating, reset the cache of rating summaries.
+      // This will trigger a refresh from the server.
+      [review.addonId]: undefined,
```
rebmullin: I am just trying to understand this ...so here you set to undefined and then you update this below in 'SET_RATING_SUMMARY'? this forces any update to update this immediately?

src/amo/reducers/reviews.js
```diff
@@ -91,6 +98,7 @@ export const initialState: ReviewsState = {
   byId: {},
   byUserId: {},
   latestUserReview: {},
+  ratingSummary: {},
```
rebmullin: I am just thinking if this name would be clearer maybe with the name groupedRatings (just to have consistency with the api.?)... or maybe ratingTotals..?

src/amo/actions/reviews.js
```diff
@@ -132,6 +135,50 @@ export function fetchReviews({
...
+  errorHandlerId: string,
+|};
+
+export type FetchRatingSummaryAction = {|
+  type: typeof FETCH_RATING_SUMMARY,
+  payload: FetchRatingSummaryParams,
+|};
+
+export function fetchRatingSummary({
+  addonId,
```
rebmullin: I see in the api docs, that this could this be addon slug. I am wondering if this could be an alternate arg to pass in or does that make it too messy (I do see you use this addonId as a key in some places)? Just wondering : )

src/amo/sagas/reviews.js
```diff
@@ -60,6 +69,41 @@ function* fetchReviews({
...
+  const errorHandler = createErrorHandler(errorHandlerId);
+  try {
+    const state = yield select(getState);
+
+    const params: GetReviewsParams = {
+      addon: addonId,
+      apiState: state.api,
+      show_grouped_ratings: true,
```
rebmullin: I see in the api, it has a version number. Will we ever need that here? 

also, as far as I could tell, this is only in the api v4. So there is that dependency on upgrading in order to use this?

tests/unit/amo/reducers/testReviews.js
```diff
@@ -891,5 +892,81 @@ describe(__filename, () => {
       const state = addReviewToState({ state: prevState, review });
       expect(state.byUserId[userId]).not.toBeDefined();
     });
+
+    it('resets ratingSummary when adding a new rating', () => {
```
rebmullin: It seems like it should but just double checking that this will cover updating a rating too?

tests/unit/amo/reducers/testReviews.js
```diff
@@ -891,5 +892,81 @@ describe(__filename, () => {
...
+        setRatingSummary({
+          addonId,
+          summary,
+        }),
+      );
+
+      expect(state.ratingSummary[addonId]).toEqual(summary);
+    });
+
+    it('preserves existing rating summaries', () => {
```
rebmullin: does each version have it's own summary too? No, right?

tests/unit/amo/sagas/testReviews.js
```diff
@@ -102,10 +103,95 @@ describe(__filename, () => {
...
+    const ratingSummaryResponse = (
+      summary = {
+        1: 0,
+        2: 0,
+        3: 0,
+        4: 0,
+        5: 0,
+      },
+    ) => {
+      // When requesting with show_grouped_ratings=true, a specical
```
rebmullin: nit: specical => special

tests/unit/amo/reducers/testReviews.js
rebmullin: actually it might but not sure if there is a use case for this..?

src/amo/reducers/reviews.js
kumar303: This was weird. I did get a Flow error (but not before?). I think the error only popped up when I added a new switch case. Maybe this needed the [`default: (action: empty;)` syntax](https://flow.org/en/docs/react/redux/#toc-typing-redux-reducers) which I never fully understood. We don't use it.

tests/unit/amo/sagas/testReviews.js
kumar303: It's [undocumented](https://github.com/wix/redux-saga-tester/pull/49). Sigh. We should probably do a sweep and convert all the old saga test suites. A lot of them are already using it.

willdurand: Yep, and probably fork and maintain our own `SagaTester` lib...

src/amo/sagas/reviews.js
kumar303: This API endpoint is a bit overloaded. We don't need to filter by version to get the grouped ratings.

I tested the v3 endpoint and it works. We should probably figure out if we can upgrade to v4 soon. We probably can.

tests/unit/amo/reducers/testReviews.js
kumar303: Ah, yes, I forgot that. Thanks.

tests/unit/amo/reducers/testReviews.js
kumar303: Hmm, I need to think about that for a minute.

kumar303: It is correct as is.

See https://github.com/mozilla/addons-frontend/issues/4529#issuecomment-400836661:

"we shouldn't be filtering ratings by version."

Good question. I always get confused when I think of this because the old site was way more complicated.

kumar303: > does each version have it's own summary too?

btw, the answer to this is technically yes because you can make a request to the API endpoint with `version=N` and you'll only get ratings submitted for that version.

src/amo/reducers/reviews.js
kumar303: Whoops, I suppose it should be `number` then. Object keys can only be [strings or symbols](https://stackoverflow.com/questions/6066846/keys-in-javascript-objects-can-only-be-strings) (which we don't use) so Flow doesn't really care if we make it `number` or `string` here. It will always end up as a string. You're right, though; the input should always be a number.

src/amo/sagas/reviews.js
kumar303: This is just destructor syntax. It is safe and acceptable to only partially destruct an object.

src/amo/sagas/reviews.js
```diff
@@ -60,6 +69,41 @@ function* fetchReviews({
   }
 }
 
+function* fetchRatingSummary({
+  payload: { errorHandlerId, addonId },
+}: FetchRatingSummaryAction): Generator<any, any, any> {
+  const errorHandler = createErrorHandler(errorHandlerId);
+  try {
+    const state = yield select(getState);
```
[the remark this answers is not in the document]
kumar303: I don't think the grouped ratings have a hard dependency on review objects. You could render grouped ratings without a review object.

Or did I misunderstand?

Enforcing data dependencies in Redux can be tricky. If we need access to data from another part of the state (say, `userId`) we typically would require it in the initial action.

src/amo/reducers/reviews.js
kumar303: The pattern we strive for looks like this:

1. When the forthcoming `RatingSummary` component is constructing, it checks `reviewState.ratingSummary[addonId]`
2. If the value is `undefined`, it dispatches `fetchRatingSummary()` and renders a loading screen
3. If the value is **not** `undefined`, it renders the data

The line you commented on is the reducer for `setRating()` or `setInternalRating()`. Those are both actions for creating or modifying a rating. Whenever that happens, we reset the `ratingSummary` so that `RatingSummary` knows to re-fetch it from the API. It keeps us in sync with the API.

If you ever see something that isn't following this pattern, it probably needs fixing.

We had to get creative in the past to work around stale API caching but those days should hopefully be over. The API has evolved.

src/amo/sagas/reviews.js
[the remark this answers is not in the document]
rebmullin: oh? maybe I misunderstood..? okay, I thought it was a part of reviews..

Ok that makes sense..to do the ck the initial action.

src/amo/reducers/reviews.js
rebmullin: cool, thanks for breaking it down : )

kumar303:
Thanks for the reviews

Issue #1299, opened 2016-10-25
Switch to using ducks for actions/actionTypes/reducers


mstriemer:
Having multiple files for actions, reducers and actionTypes (constants) is a little cumbersome and these things normally change together. There's another format that some people use called [ducks](https://github.com/erikras/ducks-modular-redux) that combines these together.

> ### Rules
> 
> A module...
> 1. MUST `export default` a function called `reducer()`
> 2. MUST `export` its action creators as functions
> 3. MUST have action types in the form `npm-module-or-app/reducer/ACTION_TYPE`
> 4. MAY export its action types as `UPPER_SNAKE_CASE`, if an external reducer needs to listen for them, or if it is a published reusable library

This would give us something like:

``` js
// src/disco/actions/discoResults.js
const LOAD_RESULTS = 'disco/discoResults/LOAD_RESULTS';

export function loadResults(results) {
  return {
    type: LOAD_RESULTS,
    payload: { results },
  };
}

export default reducer(state = [], action) {
  switch (action.type) {
    case LOAD_RESULTS:
      return [...action.payload.results];
    default:
      return state;
  }
}
```

I'm not sure where we'd want to store these, could be in a folder called `ducks`, `actions`, `reducers` or something else.

I found another article while looking for this format that suggests [organising your app even more like how django does](http://jaysoo.ca/2016/02/28/organizing-redux-application/) so we'd have `addons/components`, `reviews/components`, etc. That might play nicely into this but could be another incremental step if we'd like to do it.

Inspired by #1281.

tofumatt:
Oh, that is really cool! 👍🏻

muffinresearch:
Seems like a good approach, I've always found having these things disconnected makes it hard to see overview of the relationship between them.

kumar303:
putting them both in one place makes sense to me

willdurand:
We do this for a while now, so let's close this issue.

PR #5848, opened 2018-08-06
Refactor infoDialog reducer

### LINKED ISSUE #5861: Refactor infoDialog reducer
Refs #5860 

---

I am not sure we are still using this code but the `infoDialog` reducer is not written correctly: it does not return a consistent state shape and it does not have action creators. Let's do all of this + add Flow types.

willdurand:
Fix #5861

---

There are issues in the server logs, one of these issues is related to
`InfoDialog`, so I looked at the code and rewrote the reducer to fix the
error correctly.

### DIFF
--- a/src/core/reducers/infoDialog.js
+++ b/src/core/reducers/infoDialog.js
@@ -1,16 +1,52 @@
-import { CLOSE_INFO, SHOW_INFO } from 'core/constants';
+/* @flow */
+export const CLOSE_INFO: 'CLOSE_INFO' = 'CLOSE_INFO';
+export const SHOW_INFO: 'SHOW_INFO' = 'SHOW_INFO';
 
-export default function infoDialog(state = {}, { type, payload }) {
-  switch (type) {
+export type InfoDialogState = {|
+  data: Object,
+  show: boolean,
+|};
+
+export const initialState: InfoDialogState = {
+  data: {},
+  show: false,
+};
+
+type ShowInfoDialogParams = Object;
+
+type ShowInfoDialogAction = {|
+  type: typeof SHOW_INFO,
+  payload: ShowInfoDialogParams,
+|};
+
+export const showInfoDialog = (
+  payload: ShowInfoDialogParams,
+): ShowInfoDialogAction => {
+  return { type: SHOW_INFO, payload };
+};
+
+type CloseInfoDialogAction = {|
+  type: typeof CLOSE_INFO,
+|};
+
+export const closeInfoDialog = (): CloseInfoDialogAction => {
+  return { type: CLOSE_INFO };
+};
+
+type Action = CloseInfoDialogAction | ShowInfoDialogAction;
+
+export default function infoDialog(
+  state: InfoDialogState = initialState,
+  action: Action,
+) {
+  switch (action.type) {
     case SHOW_INFO:
       return {
         show: true,
-        data: payload,
+        data: action.payload,
       };
     case CLOSE_INFO:
-      return {
-        show: false,
-      };
+      return initialState;
     default:
       return state;
   }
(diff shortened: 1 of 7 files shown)


src/core/reducers/infoDialog.js
```diff
@@ -1,16 +1,52 @@
-import { CLOSE_INFO, SHOW_INFO } from 'core/constants';
+/* @flow */
+export const CLOSE_INFO: 'CLOSE_INFO' = 'CLOSE_INFO';
+export const SHOW_INFO: 'SHOW_INFO' = 'SHOW_INFO';
```
rebmullin: why move this out of constants?

also, I've seen this in other places too but isn't the above the same as: `export const SHOW_INFO: 'SHOW_INFO';` ?

src/core/reducers/infoDialog.js
```diff
@@ -1,16 +1,52 @@
...
 
-export default function infoDialog(state = {}, { type, payload }) {
-  switch (type) {
+export type InfoDialogState = {|
+  data: Object,
```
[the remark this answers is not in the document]
willdurand: Yep, there are "stuff" inside but I don't really know what. Before, it was inconsistent (sometimes `data` was part of the state, sometimes not, which is not something we should be doing).

src/core/reducers/infoDialog.js
willdurand: so I moved them because we want to have ducks everywhere (see #1299).

willdurand:
> @willdurand, I know you described its purpose in the intro but did you want to add a issue associated with this? or any other labels? Also, I am not sure if we need QA here...

mmh, yes let's have an issue for that.

willdurand:
> it looks good from what I can tell but wasn't sure where I could see this on the frontend. I usually like to look at that too to be sure..

Yeah, that's weird right? I have to double-check somehow but I don't really know why we have this code..

src/core/reducers/infoDialog.js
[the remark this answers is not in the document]
willdurand: Actually, I found a better type.

src/core/reducers/infoDialog.js
willdurand: We do this because Flow does not read the value of a constant. For Flow, the type of a constant is always `string`, unless we specify our own type `: 'OUR_TYPE'`. That way, we have a specific type that is not `string` but the constant value.

willdurand:
> Yeah, that's weird right? I have to double-check somehow but I don't really know why we have this code..

Alrighty, it is not being used yet but it will be soon. This is related to the mozAddonManager...

src/core/reducers/infoDialog.js
```diff
@@ -1,16 +1,58 @@
-import { CLOSE_INFO, SHOW_INFO } from 'core/constants';
+/* @flow */
+export const CLOSE_INFO: 'CLOSE_INFO' = 'CLOSE_INFO';
+export const SHOW_INFO: 'SHOW_INFO' = 'SHOW_INFO';
 
-export default function infoDialog(state = {}, { type, payload }) {
-  switch (type) {
+type InfoDialogData = {|
+  addonName: string,
+  closeAction: Function,
```
kumar303: Can you file an issue to stop dispatching a function in the action like this? It is very bad practice because it breaks any feature that relies on serializing the action stack.

If you have an idea for a lint rule to prevent anyone from dispatching a function, that would be nice.

kumar303: https://redux.js.org/basics/actions

> Actions are plain JavaScript objects

willdurand: Yup, I have it on my TODO post-it. Thanks for the review!

willdurand: Done!

Issue #5435, opened 2018-06-27
Making rating stars yellow


pwalm:
<img width="1071" alt="screen shot 2018-06-27 at 11 01 58 am" src="https://user-images.githubusercontent.com/6797635/41991443-466f5e2e-79fa-11e8-8cab-c725f6d566ac.png">

Make the rating stars on the detail page yellow:
filled: RGBA: 255,233,0,100
empty: RGBA: 255,233,0,25

**NOTE**: the wording, *4 star average*, will be handled in https://github.com/mozilla/addons-frontend/issues/5436

ioanarusiczki:
Verified fixed on AMO dev with FF61(Win 10&Android 8.0)

@kumar303 I couldn't find using the Inspector the RGBA values as mentioned above for the rating stars. 
> filled: RGBA: 255,233,0,100
empty: RGBA: 255,233,0,25

For the rating bar, I found rgba(255, 233, 0, 0.75) - yellow. 
The rating stars have : rgb(29, 29, 29) and rgba(0, 0, 0, 0); 
Let me know if it's ok. Thanks!

Issue #1292, opened 2016-10-25
Optimise SVG


muffinresearch:
Most of the source SVGs are un-optimised we should optimise them with svgo or similar.

muffinresearch:
Looks like it's possible to do this with a webpack plugin which will also optimize images - doing in the build step seems like a really good idea.

willdurand:
I believe this has been done, we definitely use svgo.

PR #5855, opened 2018-08-06
Show a star rating breakdown for add-ons

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5437

Note that the *Overall rating* has not been updated to match the UI mocks because that will happen in https://github.com/mozilla/addons-frontend/issues/5435

A typical view:

screenshot 2018-08-06 11 11 33

screenshot 2018-08-06 11 22 28

screenshot 2018-08-06 11 13 34

Add-on without any ratings:

screenshot 2018-08-06 11 12 18

Loading state:

screenshot 2018-08-06 11 20 11

Error state:

screenshot 2018-08-06 11 06 44


tests/unit/amo/components/TestRatingsByStar.js
```diff
@@ -0,0 +1,289 @@
...
+    store.dispatch(setGroupedRatings({ addonId: addon.id, grouping }));
+    const root = render({ addon });
+
+    const barValues = root.find('.RatingsByStar-barValue');
+    for (const index of [0, 1, 2, 3, 4]) {
+      expect(barValues.at(index)).toHaveProp('style', { width: '0%' });
+    }
+  });
+
+  it('renders 0% bar values when ratings are all 0', () => {
```
rebmullin: doesn't this mean ratings don't exist? (same as above?)

src/amo/components/RatingsByStar/index.js
```diff
@@ -0,0 +1,149 @@
...
+      );
+    }
+  }
+
+  renderBarValue(starCount: number) {
+    const { addon } = this.props;
+    invariant(addon, 'addon is required');
+
+    let width = 0;
+    if (addon.ratings && addon.ratings.count > 0) {
```
rebmullin: maybe this can be its own function ~ isFullBar?

src/amo/components/RatingsByStar/index.js
```diff
@@ -0,0 +1,149 @@
...
+      width = Math.round((starCount / addon.ratings.count) * 100);
+    }
+
+    return (
+      <div
+        className={makeClassName(
+          'RatingsByStar-bar',
+          'RatingsByStar-barValue',
+          {
+            'RatingsByStar-barValueLessThan100': width < 100,
```
rebmullin: **OR**...maybe this class could be called RatingsByStar-fullBar - and flip styles for this new class

I don't know if this is any better actually but at first I thought this meant something else

src/amo/components/RatingsByStar/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+            if (!errorHandler.hasError()) {
+              if (groupedRatings) {
+                starCount = groupedRatings[star];
+              }
+
+              starCountNode = loading ? (
+                <LoadingText minWidth={95} />
+              ) : (
+                i18n.formatNumber(starCount || 0)
```
rebmullin: I dont know if this matters but in the comp (I am going by [this](https://github.com/mozilla/addons-frontend/issues/5437)), the numbers don't show comma. We might actually want this but wanted to verify this.

Also, the font was slightly different - I'm showing 15px here (in the "comp"). I know it's not a drastic difference from what we have (14px) but I don't know how particularly we should be ...? What do you think?

src/amo/components/RatingsByStar/index.js
```diff
@@ -0,0 +1,149 @@
...
+              );
+            }
+
+            return (
+              <React.Fragment key={star}>
+                <div className="RatingsByStar-star">
+                  <span>{i18n.formatNumber(star)}</span>
+                  <Icon name="star-active" />
```
rebmullin: I guess this will always be active ..if even there are no ratings?

src/ui/components/MetadataCard/index.js
```diff
@@ -1,22 +1,23 @@
...
+const MetadataCard = ({ className, metadata }: Props) => {
+  invariant(metadata, 'The metadata property is required');
 
   return (
-    <div className="MetadataCard">
-      {props.metadata.map(({ content, title } = {}) => {
+    <div className={makeClassName('MetadataCard', className)}>
+      {metadata.map(({ content, title } = {}) => {
         if (content === undefined) {
           throw new Error('content is required');
```
rebmullin: just curious - I know this isn't part of your patch but can/should we use invariant here?

src/amo/components/RatingsByStar/styles.scss
```diff
@@ -0,0 +1,60 @@
...
+  display: flex;
+
+  .Icon {
+    @include margin-start(6px);
+  }
+}
+
+.RatingsByStar-barContainer {
+  align-items: center;
+  display: flex;
```
rebmullin: I dont know that either style is needed here

src/amo/components/RatingsByStar/index.js
```diff
@@ -0,0 +1,149 @@
...
+  errorHandler: ErrorHandlerType,
+  groupedRatings?: GroupedRatingsType,
+  i18n: I18nType,
+|};
+
+export class RatingsByStarBase extends React.Component<InternalProps> {
+  componentWillMount() {
+    // TODO: this is intended to load on the server (before mount) but it
+    // does not.
+    // See: https://github.com/mozilla/addons-frontend/issues/5854
```
rebmullin: I guess this might get potentially fixed with the updates @willdurand referenced earlier (in an email)?

src/ui/components/Icon/img/star-active.svg
```diff
@@ -0,0 +1,10 @@
...
+        <g id="Extension-Detail---1366px" transform="translate(-140.000000, -607.000000)" fill="#FFE900" fill-rule="nonzero">
+            <path d="M154.994575,670.99995 C153.704598,671.000763 152.477615,670.442079 151.630967,669.468394 C150.784319,668.49471 150.401158,667.201652 150.580582,665.923653 L153.046749,648.259919 L141.193762,635.514481 C140.080773,634.318044 139.711733,632.608076 140.232152,631.058811 C140.752571,629.509546 142.078939,628.369589 143.688275,628.088421 L160.214424,625.130961 L168.013827,609.468577 C168.767364,607.955994 170.3113,607 172.000594,607 C173.689888,607 175.233824,607.955994 175.98736,609.468577 L183.790813,625.130961 L200.329111,628.08437 C201.934946,628.371492 203.25546,629.513805 203.771316,631.062053 C204.287172,632.610301 203.915846,634.316807 202.803377,635.51043 L190.954439,648.26397 L193.420606,665.923653 C193.652457,667.578241 192.93975,669.223573 191.574418,670.185702 C190.209085,671.147831 188.420524,671.265104 186.941351,670.489485 L172.002619,662.698806 L157.047688,670.50569 C156.413201,670.833752 155.708782,671.003331 154.994575,670.99995 Z" id="Star-outline"></path>
+        </g>
+    </g>
+</svg>
```
rebmullin: would there be an benefit into making this an icon font? Not something to do here necessarily but this reminded me to ask about this...

tests/unit/amo/components/TestRatingsByStar.js
kumar303: One case is `addon.ratings.count = 0` and the other is `addon.ratings = undefined`. It's a little weird but I think we could legitimately encounter either case. The `addon.ratings` property is computed so there could be a split second when they are out of sync.

src/amo/components/RatingsByStar/index.js
kumar303: > the numbers don't show comma

Ah, I didn't notice that. We need localized numbers for languages like Arabic and I think in other languages it will help users parse the result. German users will be able to read this more naturally, for example: `1.234.567`

> Also, the font was slightly different - I'm showing 15px here

I went with the [closest font size variable](https://github.com/mozilla/addons-frontend/blob/master/src/core/css/inc/vars.scss#L21) just for consistency with the rest of the site. I think we need to re-sync those variables with the spec because they are off in other places too.

src/amo/components/RatingsByStar/index.js
kumar303: mmh, yeah, `active` is not a good name. I'll change it.

src/ui/components/MetadataCard/index.js
kumar303: Yes, that line pre-dates our usage of invariant. I didn't change it because I think some tests will need updating and I didn't want to make it a bigger patch.

src/amo/components/RatingsByStar/styles.scss
kumar303: without this, the bar was not vertically centered. Since the bar has a height of 16px there was some small whitespace but I wonder now if those were due to `line-height` problems that I've since fixed.

tests/unit/amo/components/TestRatingsByStar.js
rebmullin: I see..

src/ui/components/MetadataCard/index.js
rebmullin: I figured I just want to make sure : )

src/amo/components/RatingsByStar/index.js
kumar303: Yes, it will. We're targeting that fix for Q4 so I wanted to make sure and leave this as `componentWillMount()` to make it clear what the intention was.

src/ui/components/Icon/img/star-active.svg
kumar303: Possibly. I went with SVG just because it's easy to sync it with the Sketch spec. We also need to optimize our SVGs (https://github.com/mozilla/addons-frontend/issues/1292).

src/amo/components/RatingsByStar/styles.scss
rebmullin: hmm ok, I didn't see this...

I guess leave if you see something else/it's breaking..

src/amo/components/RatingsByStar/index.js
rebmullin: okay I figured, just double checking things : )

src/amo/components/RatingsByStar/styles.scss
kumar303: Actually, it made me think to force a line-height of `16px` which fixes everything 🌤 I could then remove the vertical centering.

kumar303:
Thanks for the review

tests/unit/amo/components/TestRatingsByStar.js
```diff
@@ -0,0 +1,285 @@
...
+    });
+  };
+
+  const createErrorHandlerWithError = () => {
+    const errorHandler = new ErrorHandler({
+      id: 'some-id',
+      dispatch: store.dispatch,
+    });
+    errorHandler.handle(new Error('some unexpected error'));
+    return errorHandler;
```
willdurand: If that is slightly better than using `createStubErrorHandler()`, I think we should change this helper function (assuming the main concern is injecting the `capturedError` directly, without using `handle()`).

PR #5693, opened 2018-07-23
DEMO of srcset / I5691 (for Preview in Search and Header)

rebmullin:
fixes #5691

PR #5856, opened 2018-08-06
DEMO of picture el / I5691 (for Preview in Search and Header)

rebmullin:
Here is a demo using picture el for the Addon detail header image and in the Search results. It seemed safe to use given [caniuse/#feat-picture](https://caniuse.com/#feat=picture). 

I *think* generally speaking, in these cases it's still a little better to use srcset (since the images are the same - just different sizes) but right now we need more sizes (and take into consideration resolution) to see it kick in. Right now with srcset, I only really see this happening it on smaller non-retina devices. Until we get more sizes, I think the pic el is good alternate solution as it loads the smaller (thumbs) version of the images for mobile. In search esp this can be good since it's a list view of results. We might have to eventually address resolution issues here too but at least we can control this and make this change happen if we want to. For now everything I looked at looked okay... 
I arbitrarily picked 500 as the breakpoint but we can obviously adjust that..

I am personally a fan of the picture el since we have a little bit more control with the breakpoints. Let me know what you guys think : )

note: I skipped updating tests for now 

this might be a good static theme to view since it has updated preview sizes:
http://localhost:3000/en-US/firefox/addon/eco-theme/

@willdurand @kumar303 @bobsilverberg @muffinresearch 

you can compare this to PR #5693 

Also, let me know if it is more helpful to discuss/see this in another way..

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -159,17 +159,32 @@ export class AddonBase extends React.Component {
         previewURL = addon.previewURL;
       }
 
+      const previewURLThumb = getPreviewImage(addon, { full: false });
+
+      const previewImage = previewURLThumb ? (
+        <picture>
+          <source srcSet={previewURL} media="(min-width: 500px)" />
+          <img
+            alt={label}
+            className="Addon-theme-header-image"
+            src={previewURLThumb}
+          />
+        </picture>
+      ) : (
+        <img
+          alt={label}
+          className="Addon-theme-header-image"
+          src={previewURL}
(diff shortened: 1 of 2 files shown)


rebmullin:
before I move forward with this (updating any tests:), I want to get your opinions on this. I think for now this works better than srcset so I've decided to close that PR for now. I'll keep that branch active for a bit if you want to take a look..

willdurand:
closing as not updated for 1 month and it has conflicts. I think we can move the discussion into an issue if that's needed.

PR #5864, opened 2018-08-07
Use pino-devtools to view the server logs

### LINKED ISSUE #4041: View server logs in browser devtools
The `addons-frontend` server logs are pretty hard to deal with in the shell console. We should try hooking up [Chrome logger](https://addons.mozilla.org/en-US/firefox/addon/chromelogger/) to view them in browser devtools. We can probably use this [express middleware](https://github.com/olahol/express-chrome-logger).

willdurand:
Fix #4041

---

This PR uses [pino-devtools](https://github.com/willdurand/pino-devtools) to view the server logs:

screen shot 2018-08-07 at 17 48 57

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
       }
     },
     "start-dev": {
-      "command": "npm run clean && concurrently --kill-others 'npm run webpack-dev-server' 'node bin/server.js | pino'",
+      "command": "npm run clean && concurrently --kill-others 'npm run webpack-dev-server' 'node bin/server.js | pino-devtools'",
       "env": {
         "ENABLE_PIPING": "true",
         "NODE_ENV": "development",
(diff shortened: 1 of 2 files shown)


willdurand:
@kumar303 finally, this is now node6 compatible!

willdurand:
> It would be nice if pino-devtools accepted env variables to set options.

I think that's doable.

> It would be fantastic if the AMO dev server included a script tag to client.js so that we could see the server log messages mixed in with the client log messages (we could prefix them with [server] for filtering).

Interesting approach, I like it. We can probably do that. 

> The auto-opening isn't too helpful for me because I don't develop in my default Firefox profile. I use a custom one. It's not a big deal but I usually have a lot of tabs open so it's a little awkward. I'd rather see client.js integrated into AMO itself.

This can be disabled. I like the idea of having a client in AMO. 

> It's a shame that restarting the dev server opens a new tab to pino-devtools even if you already have one open. I don't see any way to fix that, though.

That's only doable with Chrome and MacOS via applescript as far as I know.

PR #5870, opened 2018-08-07
I5552 updating expandable card to use setuistate

### LINKED ISSUE #5552: Convert setState() in ExpandableCard to setUIState()
The [ExpandableCard](https://github.com/mozilla/addons-frontend/blob/54c4a4ddbfaea24c27620bead2254b183ba8bc61/src/ui/components/ExpandableCard/index.js) component can be converted to using [`setUIState()`](https://github.com/mozilla/addons-frontend/issues/5343).

It's only using `setState()` to expand/collapse the card.

**For QA**: please make sure that tapping *filter results* in the mobile view of the search page still expands/collapses.

screenshot 2018-07-10 14 04 56
screenshot 2018-07-10 14 05 10

rebmullin:
fixes #5552

### DIFF
--- a/tests/unit/ui/components/TestExpandableCard.js
+++ b/tests/unit/ui/components/TestExpandableCard.js
@@ -1,23 +1,35 @@
-import { mount } from 'enzyme';
 import * as React from 'react';
 
 import ExpandableCard, {
   ExpandableCardBase,
+  extractId,
 } from 'ui/components/ExpandableCard';
 import {
+  applyUIStateChanges,
   createFakeEvent,
   fakeI18n,
   shallowUntilTarget,
 } from 'tests/unit/helpers';
-
-function render(props) {
-  return shallowUntilTarget(
-    <ExpandableCard i18n={fakeI18n()} {...props} />,
-    ExpandableCardBase,
-  );
-}
+import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
 
 describe(__filename, () => {
+  const getProps = ({ i18n = fakeI18n(), ...props } = {}) => {
+    return {
+      i18n,
+      id: 'expandableCard',
+      store: dispatchClientMetadata().store,
+      ...props,
+    };
+  };
+
+  function render({ children, ...otherProps } = {}) {
+    const props = getProps(otherProps);
+    return shallowUntilTarget(
+      <ExpandableCard {...props}>{children || 'some info'}</ExpandableCard>,
+      ExpandableCardBase,
+    );
+  }
+
   it('renders a ExpandableCard', () => {
     const root = render();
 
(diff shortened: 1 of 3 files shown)


tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -32,24 +44,17 @@ describe(__filename, () => {
...
 
-    // Clicking on the toggle should expand the card.
-    root.find('.ExpandableCard-ToggleLink').simulate('click', fakeEvent);
+    const root = render({ store });
 
-    sinon.assert.called(fakeEvent.preventDefault);
-    expect(card()).toHaveClassName('ExpandableCard--expanded');
+    // We want to simulate card's link onclick.
+    // TODO: look into why this isn't available with find
+    root.instance().onClick(createFakeEvent());
```
rebmullin: I know maybe this isn't the way I should be doing this but when I do something like this:

`root.find(Card).simulate('click', createFakeEvent());` or looked for the `a` tag or the a's class (root.find('.ExpandableCard-ToggleLink')..., it didn't work :/

Should I be looking into mount here?

tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -1,23 +1,35 @@
-import { mount } from 'enzyme';
```
kumar303: 🙏

tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -32,24 +44,17 @@ describe(__filename, () => {
...
+    const { store } = dispatchClientMetadata();
 
-    // Clicking on the toggle should expand the card.
-    root.find('.ExpandableCard-ToggleLink').simulate('click', fakeEvent);
+    const root = render({ store });
 
-    sinon.assert.called(fakeEvent.preventDefault);
-    expect(card()).toHaveClassName('ExpandableCard--expanded');
+    // We want to simulate card's link onclick.
+    // TODO: look into why this isn't available with find
```
kumar303: The element with `onClick` is not part of the root which is why you can't get it with `root.find()`. It's part of the JSX which was passed to the `header` prop of `Card`. You'll need to get that prop and pass it to `shallow()`. That will give you a new shallow wrapper root that you can simulate `click` on.

I mentioned it in the *Testing JSX passed through properties* of [this article](https://hacks.mozilla.org/2018/04/testing-strategies-for-react-and-redux/) which goes into detail about how to do it with some example code.

tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -32,24 +44,17 @@ describe(__filename, () => {
   });
 
   it('toggles when clicked', () => {
```
kumar303: I think this component could benefit from one other test that simulates two clicks and makes sure the state changes accordingly. Currently it only simulates one click which doesn't exercise the toggling part.

tests/unit/ui/components/TestExpandableCard.js
rebmullin: ahh ok yes! thank you

rebmullin:
thanks @kumar303 for the review. I've made some updates. can you please take another look?

tests/unit/ui/components/TestExpandableCard.js
```diff
@@ -32,24 +45,26 @@ describe(__filename, () => {
...
-    // We have to mount for this test because of the usage of `setState`
-    // in this component.
-    const root = mount(<ExpandableCard i18n={fakeI18n()} />);
-    const card = () => root.find('.ExpandableCard');
-    const fakeEvent = createFakeEvent();
+    const { store } = dispatchClientMetadata();
+
+    const root = render({ store });
+    const cardHeader = root.prop('header');
```
kumar303: It's not mandatory but if you check for `header` right above this line it will make a better test failure:

```js
expect(root).toHaveProp('header');
const cardHeader = root.prop('header');
// ...
```

PR #5872, opened 2018-08-07
Add custom reporter to reduce the test output

### LINKED ISSUE #2980: Jest test output is too big because of console logging
This has been happening realistically since Jest but I think the bundle size has made it more intense. Basically our test output is so huge that Travis stops rendering it and it makes looking up failed tests quite difficult.

I don't think there's an easy fix but I'm filing this here because having to download the raw logs for failing tests is pretty annoying and locally I'll often run `yarn test -- --silent` because otherwise it's way too loud.

We should fix things up so test output, especially on Travis, is nice and concise so we can see–at a glance–test failures.

willdurand:
Fix #2980
~~:warning: depends on #5871~~

---

That works on my machine™️ Maybe we should test this reporter, but I don't really know how. I guess we'll notice when things will break :D


willdurand:
rebased.

jest.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     // Replaces the following formats with an empty module.
     '^.+\\.(scss|css|svg|woff|woff2|mp4|webm)$': '<rootDir>/tests/emptyModule',
   },
+  reporters: ['<rootDir>/tests/fingers-crossed-reporter.js'],
```
kumar303: I think it's ready for a proper name! I'd suggest putting it in a directory called `tests/jest-reporters/`

willdurand: Isn't it ready for a proper name? We hope we won't need to output console logs, that's the idea behind this name :)

willdurand:
> I wanted it to only show the output for a failing test but it shows the output for the whole suite. This is better than before but less than ideal.

Yep I saw that too, but it is better than absolutely all logs. We don't have too many test cases per test suite.

Scimonster:
Maybe consider spinning this off into its own module so other people can easily take advantage of it?

willdurand:
> Maybe consider spinning this off into its own module so other people can easily take advantage of it?

Not sure, it works well for us in this project but I am not sure it is generic enough to be reused elsewhere.

PR #5873, opened 2018-08-07
Make rating stars yellow

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5435

Notes:
- Read-only stars are no longer hollow/open; they are closed. The spec never had them like this so I think it was a mistake.
- I cleaned up the SVG file names
- Everything "gold" is now "yellow" because the photon color spec calls it yellow.
- I converted the test suite to Enzyme, refs https://github.com/mozilla/addons-frontend/issues/1104

Before:

screenshot 2018-08-07 18 10 31

After:

screenshot 2018-08-07 17 16 38

RTL:

screenshot 2018-08-07 18 16 48

Half stars:

screenshot 2018-08-07 17 15 10
screenshot 2018-08-07 17 15 31

No ratings:

screenshot 2018-08-07 17 17 50

Updated review listing:

screenshot 2018-08-07 17 15 57

No regressions in the main star selector:

screenshot 2018-08-07 17 16 14


src/amo/components/UserProfile/index.js
```diff
@@ -351,7 +351,7 @@ export class UserProfileBase extends React.Component<InternalProps> {
                   <Rating
                     rating={user.average_addon_rating}
                     readOnly
-                    styleName="small"
+                    styleSize="small"
```
kumar303: Found this after the Flow conversion 👻

src/ui/components/UserRating/index.js
```diff
@@ -32,11 +32,11 @@ export const UserRatingBase = (props: Props) => {
   return (
     <Rating
       className={className}
-      isOwner={isOwner}
       onSelectRating={onSelectRating}
       rating={(review && review.rating) || 0}
-      readOnly={readOnly}
+      readOnly={readOnly || false}
```
willdurand: Would it make sense to have a default value for the `readOnly` prop instead?

tests/unit/ui/components/TestRating.js
```diff
@@ -1,58 +1,53 @@
...
 function render(customProps = {}) {
   const props = {
     i18n: fakeI18n(),
     ...customProps,
   };
-  return findRenderedComponentWithType(
-    renderIntoDocument(<Rating {...props} />),
-    RatingBase,
-  );
+  return shallowUntilTarget(<Rating {...props} />, RatingBase);
```
willdurand: 😍

src/ui/components/UserRating/index.js
kumar303: We could but it won't make this Flow error go away:

```
Cannot create Rating element because undefined [1] is incompatible with boolean [2] in property readOnly.

     src/ui/components/UserRating/index.js
 [1] 15│   readOnly?: boolean,
       :
     39│     } = this.props;
     40│
     41│     return (
     42│       <Rating
     43│         className={className}
     44│         onSelectRating={onSelectRating}
     45│         rating={(review && review.rating) || 0}
     46│         readOnly={readOnly}
     47│         styleSize={styleSize}
     48│         yellowStars={isOwner}
     49│       />
     50│     );
     51│   }
     52│ }

     src/ui/components/Rating/index.js
 [2] 21│   readOnly: boolean,
```

The reason is that while Flow does understand `defaultProps` in normal cases, it doesn't know about `defaultProps` in our [type hint](https://github.com/mozilla/addons-frontend/blob/def32ce267fdd4c6aadaefd70237f381573a6456/src/ui/components/UserRating/index.js#L51) (an abnormal case) so we are forced to make `readOnly` an optional property.

I looked into making the type hint also know about `defaultProps` but I couldn't get it to work. I don't think we have a granular enough [type helper](https://flow.org/en/docs/react/types/) for it. [`React.ElementConfig`](https://flow.org/en/docs/react/types/#toc-react-elementconfig) is what we want but it doesn't let us swap `InternalProps` for `Props`.

PR #5874, opened 2018-08-07
I5553 showmorecard to use setuistate

### LINKED ISSUE #5553: Convert setState() in ShowMoreCard to setUIState()
The [ShowMoreCard](https://github.com/mozilla/addons-frontend/blob/54c4a4ddbfaea24c27620bead2254b183ba8bc61/src/ui/components/ShowMoreCard/index.js) component can be converted to using [`setUIState()`](https://github.com/mozilla/addons-frontend/issues/5343).

It's using `setState()` to expand/collapse content.

**For QA**: please make sure that long add-on descriptions can still be expanded/collapsed.

rebmullin:
fixes #5553 

this component seems so similar to the expandable Card I am wondering if they should be combined...

I also I had a few notes/ questions below : )

### DIFF
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -34,6 +34,7 @@ type InternalProps = {|
   i18n: I18nType,
   setUIState: ($Shape<UIStateType>) => void,
   uiState: UIStateType,
+  children: React.Element<any>,
 |};
 
 const initialUIState: UIStateType = {
@@ -45,19 +46,33 @@ export class ShowMoreCardBase extends React.Component<InternalProps> {
   contents: HTMLElement | null;
 
   componentWillReceiveProps(nextProps: InternalProps) {
-    const html =
-      this.props.children.props.dangerouslySetInnerHTML &&
-      this.props.children.props.dangerouslySetInnerHTML.__html;
+    const { children } = this.props;
+    const { children: newChildren } = nextProps;
 
-    const htmlNew =
-      nextProps.children.props.dangerouslySetInnerHTML &&
-      nextProps.children.props.dangerouslySetInnerHTML.__html;
+    let html =
+      children.props &&
+      children.props.dangerouslySetInnerHTML &&
+      children.props.dangerouslySetInnerHTML.__html;
+
+    let newHtml =
+      newChildren.props &&
+      newChildren.props.dangerouslySetInnerHTML &&
+      newChildren.props.dangerouslySetInnerHTML.__html;
+
+    // If it's not html, let check that text too.
+    if (!html && children && !children.props) {
+      html = children;
+    }
+
+    if (newHtml !== 'undefined' && newChildren && !newChildren.props) {
+      newHtml = newChildren;
+    }
 
     // Reset UIState if component html has changed.
     // This is needed b/c if you return to an addon that you've
     // already visited the component doesn't hit unmount again and the store
     // keeps the last component's UIState which isn't what we want.
-    if (htmlNew && html !== htmlNew) {
+    if (newHtml && html !== newHtml) {
       this.props.setUIState({
         ...initialUIState,
       });
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -19,7 +19,7 @@ export const MAX_HEIGHT = 150;
 
 type UIStateType = {|
   showAllContent: boolean,
-  disableTruncateCall: boolean,
+  readMoreExpanded: boolean,
 |};
 
 type Props = {|
@@ -38,43 +38,62 @@ type InternalProps = {|
 
 const initialUIState: UIStateType = {
   showAllContent: true,
-  disableTruncateCall: false,
+  readMoreExpanded: false,
 };
 
 export class ShowMoreCardBase extends React.Component<InternalProps> {
   contents: HTMLElement | null;
 
-  onClick = (event: SyntheticEvent<HTMLAnchorElement>) => {
-    const { setUIState } = this.props;
-
-    event.preventDefault();
-
-    setUIState({ showAllContent: true });
-
-    // This is used to prevent any additional
-    // unnecessary calls to the truncate function.
-    setUIState({ disableTruncateCall: true });
-  };
-
   componentWillReceiveProps(nextProps: InternalProps) {
-    // Once read more has been clicked, the card has been opened so we no longer
-    // have to run this truncate for the component.
-    if (!this.props.uiState.disableTruncateCall) {
+    const html =
+      this.props.children.props.dangerouslySetInnerHTML &&
+      this.props.children.props.dangerouslySetInnerHTML.__html;
+
+    const htmlNew =
+      nextProps.children.props.dangerouslySetInnerHTML &&
+      nextProps.children.props.dangerouslySetInnerHTML.__html;
+
+    // Reset UIState if component html has changed.
+    // This is needed b/c if you return to an addon that you've
+    // already visited the component doesn't hit unmount again and the store
+    // keeps the last component's UIState which isn't what we want.
+    if (htmlNew && html !== htmlNew) {
+      this.props.setUIState({
+        ...initialUIState,
+      });
+    }
+
+    // If the read more has already been expanded, we can skip
+    // the call to truncate.
+    // Ideally this would only be called one time and it wouldn't be
+    // needed after the initial set up but we need this here (vs componentDidMount)
+    // to get an accurate clientHeight.
+    if (!this.props.uiState.readMoreExpanded) {
       this.truncateToMaxHeight(this.contents);
     }
   }
 
   truncateToMaxHeight = (contents: HTMLElement | null) => {
+    const { uiState } = this.props;
     if (contents) {
       // If the contents are short enough they don't need a "show more" link; the
       // contents are expanded by default.
-      if (
-        this.props.uiState.showAllContent &&
-        contents.clientHeight >= MAX_HEIGHT
-      ) {
-        this.props.setUIState({ showAllContent: false });
+      if (uiState.showAllContent && contents.clientHeight >= MAX_HEIGHT) {
+        this.props.setUIState({
+          ...uiState,
+          showAllContent: false,
+        });
       }
     }
+  };
+
+  onClick = (event: SyntheticEvent<HTMLAnchorElement>) => {
+    event.preventDefault();
+
+    this.props.setUIState({
+      showAllContent: true,
+      readMoreExpanded: true,
+    });
   };
 
   render() {
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -64,7 +64,7 @@ export class ShowMoreCardBase extends React.Component<InternalProps> {
       html = children;
     }
 
-    if (newHtml !== 'undefined' && newChildren && !newChildren.props) {
+    if (!newHtml && newChildren && !newChildren.props) {
       newHtml = newChildren;
     }
 
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -23,7 +23,7 @@ type UIStateType = {|
 |};
 
 type Props = {|
-  children: any,
+  children: React.Element<any>,
   className?: string,
   header?: React.Element<any> | string,
   id: string,
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -3,7 +3,6 @@
 import invariant from 'invariant';
 import makeClassName from 'classnames';
 import * as React from 'react';
-// import PropTypes from 'prop-types';
 import { compose } from 'redux';
 
 import translate from 'core/i18n/translate';
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -40,23 +40,36 @@ const initialUIState: UIStateType = { expanded: true };
 export class ShowMoreCardBase extends React.Component<InternalProps> {
   contents: HTMLElement | null;
 
-  componentDidMount() {
-    this.truncateToMaxHeight(this.contents);
-  }
+  hasExpanded = false;
 
   onClick = (event: SyntheticEvent<HTMLAnchorElement>) => {
     event.preventDefault();
     this.props.setUIState({ expanded: true });
   };
 
+  componentWillReceiveProps() {
+    if (!this.hasExpanded) {
+      this.truncateToMaxHeight(this.contents);
+    }
+  }
+
+  shouldComponentUpdate(nextProps: InternalProps) {
+    if (
+      !this.props.uiState.expanded &&
+      this.props.uiState.expanded !== nextProps.uiState.expanded
+    ) {
+      this.hasExpanded = true;
+    }
+
+    return true;
+  }
+
   truncateToMaxHeight = (contents: HTMLElement | null) => {
     // If the contents are short enough they don't need a "show more" link; the
     // contents are expanded by default.
     if (contents) {
-      if (contents.clientHeight >= MAX_HEIGHT) {
+      if (this.props.uiState.expanded && contents.clientHeight >= MAX_HEIGHT) {
         this.props.setUIState({ expanded: false });
-      } else {
-        this.props.setUIState({ expanded: true });
       }
     }
   };
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -47,21 +47,18 @@ export class ShowMoreCardBase extends React.Component<InternalProps> {
     this.props.setUIState({ expanded: true });
   };
 
-  componentWillReceiveProps() {
-    if (!this.hasExpanded) {
-      this.truncateToMaxHeight(this.contents);
-    }
-  }
+  componentWillReceiveProps(nextProps: InternalProps) {
+    const {
+      uiState: { expanded },
+    } = this.props;
 
-  shouldComponentUpdate(nextProps: InternalProps) {
-    if (
-      !this.props.uiState.expanded &&
-      this.props.uiState.expanded !== nextProps.uiState.expanded
-    ) {
+    if (!expanded && expanded !== nextProps.uiState.expanded) {
       this.hasExpanded = true;
     }
 
-    return true;
+    if (!this.hasExpanded) {
+      this.truncateToMaxHeight(this.contents);
+    }
   }
 
   truncateToMaxHeight = (contents: HTMLElement | null) => {
@@ -129,5 +126,6 @@ export default compose(
     fileName: __filename,
     extractId,
     initialState: initialUIState,
+    resetOnUnmount: true,
   }),
 )(ShowMoreCardBase);
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -23,7 +23,7 @@ type UIStateType = {|
 |};
 
 type Props = {|
-  children: React.Element<any> | string,
+  children: any,
   className?: string,
   header?: React.Element<any> | string,
   id: string,
@@ -34,7 +34,6 @@ type InternalProps = {|
   i18n: I18nType,
   setUIState: ($Shape<UIStateType>) => void,
   uiState: UIStateType,
-  children: React.Element<any>,
 |};
 
 const initialUIState: UIStateType = {
--- a/src/ui/components/ShowMoreCard/index.js
+++ b/src/ui/components/ShowMoreCard/index.js
@@ -18,7 +18,8 @@ import './styles.scss';
 export const MAX_HEIGHT = 150;
 
 type UIStateType = {|
-  expanded: boolean,
+  showAllContent: boolean,
+  disableTruncateCall: boolean,
 |};
 
 type Props = {|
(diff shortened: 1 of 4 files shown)


src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,66 @@
...
+|};
+
+type InternalProps = {|
+  ...Props,
+  i18n: I18nType,
+  setUIState: ($Shape<UIStateType>) => void,
+  uiState: UIStateType,
+|};
 
-  componentWillReceiveProps() {
```
rebmullin: I decided to remove this b/c with these updates, it now triggers every time setUiState is called ....so it causes an endless loop. It seemed like in one of the tests below that there is a possibility that the child could change and we want to see the truncatetoMaxHeight get called for this but I still was questioning whether this was something we needed..? Next time the page loads, it should get updated (I think).

let me know if you see any issues with this..

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,81 @@
...
   it('renders children', () => {
     const root = render({ children: 'Hello I am description' });
-
-    expect(root.childAt(0)).toHaveText('Hello I am description');
+    const contents = root.find('.ShowMoreCard-contents');
+    expect(contents).toHaveText('Hello I am description');
   });
 
-  it('executes truncateToMaxHeight when it recieves props', () => {
-    const root = mount(<ShowMoreCardBase i18n={fakeI18n()} />);
```
rebmullin: I explained why I removed this above. let me know if you disagree / feel this IS needed..

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,81 @@
...
-
-    const contentNode = component.contents;
-    const truncateToMaxHeight = sinon.spy(component, 'truncateToMaxHeight');
-    root.setProps(); // simulate any kind of update to properties
-
-    sinon.assert.calledWith(truncateToMaxHeight, contentNode);
-  });
-
-  it('expands if new child content is smaller', () => {
-    const root = mount(
```
rebmullin: same: I explained why I removed this above. let me know if you disagree / feel this IS needed..

src/ui/components/ShowMoreCard/index.js
rebmullin: hmm..I guess I could check if there are prop changes to anything besides the uiState...

rebmullin: for now tho, I am thinking I will as I have it. I am still not convinced the props will be changing too often...

rebmullin: I've looked into this so more and I was not able to get this to work :/ 
When I leave this in componentDidMount the height on pages that are not the original page is only about 19px :/  Even tho when I inspect the node it I see the correct height there..so I am not entirely sure why it's showing the 19px. I see it get updated to the accurate amount on one of the other/later component life cycles (such as componentWillReceiveProps or shouldComponentUpdate etc, but if you call truncate from here, it updates the uiState prop and runs through this again (loop).

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,112 @@
...
 
-    const contentNode = component.contents;
-    const truncateToMaxHeight = sinon.spy(component, 'truncateToMaxHeight');
-    root.setProps(); // simulate any kind of update to properties
-
-    sinon.assert.calledWith(truncateToMaxHeight, contentNode);
-  });
+    const contentNode = root.instance().contents;
 
-  it('expands if new child content is smaller', () => {
```
rebmullin: I will work on adding this back if we end up going this route...

rebmullin:
I don’t know if this is okay or seems hacky :/ but I could not get an accurate height without putting this truncate function in one of the later life cycles and using a flag. At first I didn't think I needed this flag at all but I ran into issues when there were more than 1 ShowMore on the page in that the 2nd link wouldn’t work without clicking a 2nd time...

There might be a cleaner/better way to do all this or maybe this is good case where we just keep setState...?

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,82 @@
...
-  }
+type Props = {|
+  children: React.Element<any> | string,
+  className?: string,
+  header?: React.Element<any> | string,
+  id: string,
+|};
 
-  componentWillReceiveProps() {
-    this.truncateToMaxHeight(this.contents);
```
bobsilverberg: To follow up on our conversation, I'm wondering whether you can address the issue by just adding a check into the existing `componentWillReceiveProps` that checks whether `children` has changed, and only calls `this.truncateToMaxHeight` is `children` has in fact changed.

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,82 @@
...
+  uiState: UIStateType,
+|};
 
-  onClick = (event) => {
+const initialUIState: UIStateType = { expanded: true };
+
+export class ShowMoreCardBase extends React.Component<InternalProps> {
+  contents: HTMLElement | null;
+
+  hasExpanded = false;
```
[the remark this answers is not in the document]
rebmullin: I did notice this but flow didn't complain about this so I thought that maybe this was ok..? I will revisit this tho..

src/ui/components/ShowMoreCard/index.js
rebmullin: As mentioned I did something similar but it wasn't working. I revisited this just to be sure but I am running into the same issue where the height is not accurate when it first see there is a difference (if that makes sense). This was the same issue that I was running into with adding this into just componentDidMount. Here's the full code that I am testing now. Maybe you will see something I am not tho:

https://pastebin.com/GeUt3BHp

rebmullin:
@bobsilverberg, let me know what you think of the updates. I made some changes and I think it's a little cleaner. I added some comments too - to help clarify. Please let me know what you think.

note: I still need to update tests

rebmullin:
@bobsilverberg, thanks for checking this out:)  

So I just tried this again to see if maybe I missed something when I was testing this late last night : / and this IS still working for me. So I am a little baffled..? I even tried that link you noted above...

Also just a quick thought - are you looking at mobile? the read more for the extension part only shows up in mobile (this matches what I see on dev too). 

I am also just curious, do you see any errors in the console. This is an old branch, so I am just wondering if that might mess things up (?) unless you reinstall. Let me try rebasing and maybe you can double check this again if you don't mind :)

bobsilverberg:
> Also just a quick thought - are you looking at mobile? the read more for the extension part only shows up in mobile (this matches what I see on dev too).

Oh, duh!  I think I was even the one who implemented that change! No, I was totally looking at it on desktop and expecting the "Read more" link to be there. My apologies. I'll take another look now.

bobsilverberg:
@rebmullin Yeah, so I was wrong about it not working at all, it does work, on mobile, the first time you access the page, but I was right about there being a problem with `disableTruncateCall`. If you click a "Read more" link for one add-on, and then navigate to a different add-on, the "About" box starts off as expanded for that new add-on too. I think it should revert back to "non-expanded" when a new add-on is loaded. Do you agree?

rebmullin:
> If you click a "Read more" link for one add-on, and then navigate to a different add-on, the "About" box starts off as expanded for that new add-on too. I think it should revert back to "non-expanded" when a new add-on is loaded. Do you agree?

Yeh I definitely agree that it should work this way. I thought it *was* doing this:/

This is what I thought I show in the [video](https://monosnap.com/file/9JuVp0L0n82wBb4UJ3jN7uhE2RVfld/#) here - going from one add on to another...


Anyway, I am going to take another look to see what's going and see if I can replicate the issue that you're seeing.

I'll keep you posted. Thanks so much for taking another look @bobsilverberg  : )

bobsilverberg:
I think it only happens when you don't navigate away from the Details page. See this video for an example:

aug-28-2018 15-54-47

rebmullin:
@bobsilverberg, ahh..thanks for the video! that's a big help : )

rebmullin:
hey @bobsilverberg, do you mind taking another review pass at this one? I tried to add some comments to clarify why I was doing some things. let me know if any questions : )


(I will do tests when the approach is deemed ok).

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,107 @@
...
+
+export class ShowMoreCardBase extends React.Component<InternalProps> {
+  contents: HTMLElement | null;
+
+  componentWillReceiveProps(nextProps: InternalProps) {
+    const html =
+      this.props.children.props.dangerouslySetInnerHTML &&
+      this.props.children.props.dangerouslySetInnerHTML.__html;
+
+    const htmlNew =
```
bobsilverberg: Nit, `newHtml` sounds better (to me) than `htmlNew`.

rebmullin:
> This is fantastic...

OMG, I am so happy to hear this : ) Been battling with this one for a while..

I will update tests soon. thanks!

src/ui/components/ShowMoreCard/index.js
rebmullin: When I was looking at updating tests for this and I realized that maybe children won't always be dangerouslySetInnerHTML...so I may need to tweak this a bit : /

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,122 @@
...
+      children.props.dangerouslySetInnerHTML.__html;
+
+    let newHtml =
+      newChildren.props &&
+      newChildren.props.dangerouslySetInnerHTML &&
+      newChildren.props.dangerouslySetInnerHTML.__html;
+
+    // If it's not html, let check that text too.
+    if (!html && children && !children.props) {
+      html = children;
```
rebmullin: I believe this will cover all basis..

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
-    root.setProps({ children: 'short content' });
-    expect(root.state('expanded')).toEqual(true);
+    // Simulating read more has been expanded already.
+    const root = render({
+      store: {
+        ...store,
+        store: {
+          uiState: {
+            readMoreExpanded: true,
+          },
```
rebmullin: I am hoping this is ok or do I need to show it all out:

it hits truncate
it shows read more
user clicks on read more
state is change to readMoreExpanded: true
this calls componentWillReceive Props again and now it should not update. 

I wasn't sure...I kind of skipped to the part where is was just true. Let me know if that's a problem..

rebmullin:
hey @bobsilverberg, I've updated the tests. Please let me know what you think now: )

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,122 @@
...
+  showAllContent: true,
+  readMoreExpanded: false,
+};
+
+export class ShowMoreCardBase extends React.Component<InternalProps> {
+  contents: HTMLElement | null;
 
-  componentWillReceiveProps() {
-    this.truncateToMaxHeight(this.contents);
+  componentWillReceiveProps(nextProps: InternalProps) {
```
rebmullin: @willdurand, I guess I should replace this? and replace with shouldComponentUpdate (maybe) or another life cycle?

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
 
-    // We cannot directly target the footer link, so we simulate the click here.
-    root.instance().onClick(createFakeEvent());
-    root.update();
+    moreLink.simulate('click', createFakeEvent());
 
-    expect(root).toHaveClassName('ShowMoreCard--expanded');
+    applyUIStateChanges({ root, store });
 
-    expect(root).toHaveProp('footerLink', null);
```
bobsilverberg: You removed this check that `footertLink` is now null. Does this check no longer apply?

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
+          dangerouslySetInnerHTML={{
+            __html: '<span>This is different Text.</span>',
+          }}
+        />
+      ),
+    });
+    /* eslint-enable react/no-danger */
 
-    expect(root.childAt(0)).toHaveText('Hello I am description');
+    sinon.assert.called(dispatchSpy);
```
bobsilverberg: The component does not call `dispatch` itself directly. I assume this is called when `setUIState` is called internally, but it seems like it would be better to just check for `setUIState` to be called, rather than checking for `dispatch` to be called.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
+    const truncateToMaxHeight = sinon.spy(
+      root.instance(),
+      'truncateToMaxHeight',
     );
 
-    // We have to manually set the expanded flag to false because we don't have
-    // a `clientHeight` in the tests.
-    root.setState({ expanded: false });
+    // We are simulating the truncate method call.
+    root.instance().truncateToMaxHeight({ clientHeight: MAX_HEIGHT + 1 });
```
bobsilverberg: Why do we have to call this first?

tests/unit/ui/components/TestShowMoreCard.js
bobsilverberg: I think because you're specifically testing that it doesn't call `truncateToMaxHeight` if it's already expanded, then just starting the test in the expanded state is fine.

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,122 @@
...
+      children.props &&
+      children.props.dangerouslySetInnerHTML &&
+      children.props.dangerouslySetInnerHTML.__html;
+
+    let newHtml =
+      newChildren.props &&
+      newChildren.props.dangerouslySetInnerHTML &&
+      newChildren.props.dangerouslySetInnerHTML.__html;
+
+    // If it's not html, let check that text too.
```
bobsilverberg: Let's try to use proper grammar in comments. This should be something like: `If it's not html, check for plain text.`

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,122 @@
...
+    if (!html && children && !children.props) {
+      html = children;
+    }
+
+    if (newHtml !== 'undefined' && newChildren && !newChildren.props) {
+      newHtml = newChildren;
+    }
+
+    // Reset UIState if component html has changed.
+    // This is needed b/c if you return to an addon that you've
```
bobsilverberg: I would prefer we use full words rather than abbreviations in comments. Please change `b/c` to `because`.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
     expect(root).toHaveClassName(className);
   });
 
   it('renders children', () => {
     const root = render({ children: 'Hello I am description' });
+    const contents = root.find('.ShowMoreCard-contents');
+    expect(contents).toHaveText('Hello I am description');
+  });
+
+  it('dispatches if the html children has changed', () => {
```
bobsilverberg: I think this would make more sense as `dispatches if the children's html has changed`.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,223 @@
...
+    /* eslint-enable react/no-danger */
 
-    expect(root.childAt(0)).toHaveText('Hello I am description');
+    sinon.assert.called(dispatchSpy);
   });
 
-  it('executes truncateToMaxHeight when it recieves props', () => {
-    const root = mount(<ShowMoreCardBase i18n={fakeI18n()} />);
-    const component = root.instance();
+  it('dispatches if the children has changed', () => {
```
bobsilverberg: I think this would be better as: `dispatches if the children's text has changed`.

src/ui/components/ShowMoreCard/index.js
rebmullin: sorry this is a bad habit of mine..I'll update thanks!

tests/unit/ui/components/TestShowMoreCard.js
rebmullin: I guess we don't need to. I was simulating the whole process (does this when it firsts renders too).  I'll remove this..

tests/unit/ui/components/TestShowMoreCard.js
rebmullin: Oh?! I think I inadvertently removed that. I'll add that back, thanks!

tests/unit/ui/components/TestShowMoreCard.js
rebmullin: I see what you mean but for some reason when I spy on setUIState, the tests don't work. I'll revisit this tho to be sure..

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,126 @@
...
+  header?: React.Element<any> | string,
+  id: string,
+|};
+
+type InternalProps = {|
+  ...Props,
+  i18n: I18nType,
+  setUIState: ($Shape<UIStateType>) => void,
+  uiState: UIStateType,
+  children: React.Element<any>,
```
rebmullin: I remove this here flow complains because of the following:
```javascript
let html =
      children.props &&
      children.props.dangerouslySetInnerHTML &&
      children.props.dangerouslySetInnerHTML.__html;
...
```
it seems weird to have this in both spots (in Props too), but I guess this is okay?

rebmullin:
hey @bobsilverberg, I made some changes. Please take another look and let me know how it looks nice. Hopefully we are close :)

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,126 @@
...
+      newChildren.props &&
+      newChildren.props.dangerouslySetInnerHTML &&
+      newChildren.props.dangerouslySetInnerHTML.__html;
+
+    // If it's not html, check for plain text.
+    if (!html && children && !children.props) {
+      html = children;
+    }
+
+    if (newHtml !== 'undefined' && newChildren && !newChildren.props) {
```
bobsilverberg: Why are you checking `!html` at line 63, but here you're checking `newHtml !== 'undefined'`? Also, shouldn't it be just `undefined` and not `'undefined'`? Maybe I'm not understanding the logic of this statement.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,214 @@
...
+      store: dispatchClientMetadata().store,
+      ...props,
+    };
+  };
+
+  function render({ children, ...otherProps } = {}) {
+    const props = getProps(otherProps);
     return shallowUntilTarget(
-      <ShowMoreCard i18n={fakeI18n()} {...props} />,
+      <ShowMoreCard {...props}>{children || 'some info'}</ShowMoreCard>,
```
bobsilverberg: Could you make `'some info'` be the default value of the `children` argument in the function signature, rather than having this `||` here?

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,214 @@
...
-    expect(root).not.toHaveClassName('ShowMoreCard--expanded');
+    applyUIStateChanges({ root, store });
 
-    // We have to shallow this prop because we pass a component to it.
-    expect(shallow(root.prop('footerLink')).html()).toContain(
-      '<span class="visually-hidden">Expand to</span> Read more',
-    );
+    expect(root).toHaveProp('footerLink');
+    const footerLink = root.prop('footerLink');
+    const cardLink = shallow(footerLink);
```
bobsilverberg: You could probably do without this `cardLink` const and just have
 `const moreLink = shallow(footerLink).find('.ShowMoreCard-expand-link');`

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,214 @@
...
-  it('expands if new child content is smaller', () => {
-    const root = mount(
-      <ShowMoreCardBase i18n={fakeI18n()}>
-        This would be very long content, but we cannot get `clientHeight` in the
-        tests, so this will be forced below (via setState()).
-      </ShowMoreCardBase>,
+  it("does not call resetUIState if the children's html is the same", () => {
+    /* eslint-disable react/no-danger */
+    const root = render({
+      children: (
```
bobsilverberg: Because these values are the same, it might be better to set a `const` for `children` and just use that in both the `render` and the `setProps` calls.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,214 @@
...
+      ),
+    });
+    /* eslint-enable react/no-danger */
+
+    sinon.assert.notCalled(resetUIStateSpy);
+  });
+
+  it("does not call resetUIState if the children's text is the same", () => {
+    const root = render({
+      children: 'Some text',
```
bobsilverberg: As above, using a `const` for `children` might be better.

tests/unit/ui/components/TestShowMoreCard.js
```diff
@@ -1,105 +1,214 @@
...
+    const { store } = dispatchClientMetadata();
 
-    expect(root.state('expanded')).toEqual(false);
-    // This will call `componentWillReceiveProps()`, the content of `children`
-    // is for example purpose.
-    root.setProps({ children: 'short content' });
-    expect(root.state('expanded')).toEqual(true);
+    // Simulating read more has been expanded already.
+    const root = render({
+      store: {
```
bobsilverberg: I realize this contradicts what I said earlier, but I think the problem with what you're doing here - trying to render it with the correct state - will be fragile. The test shouldn't have to understand the internals of `uiState`. So maybe you do need to do some setup with the component in order to get it into the state you want for the start of the test.

src/ui/components/ShowMoreCard/index.js
rebmullin: I see what you mean but I am sort of wondering now if we should even check this at this point. It's adding an extra level of complexity for something that is not even set up on site at this point. I feel like with this type of component - where it's pulling in variable text (which has variable length) - it's text from the user so it would always be set up with `dangerousSetInnerHtml`...

@bobsilverberg, what do you think?

rebmullin: I guess we have this in the test so we should keep this and work it out..

tests/unit/ui/components/TestShowMoreCard.js
rebmullin: okay, I've set it up to do all the steps to get it to this point then

rebmullin:
@bobsilverberg, can you please take another look? :) 🤞

src/ui/components/ShowMoreCard/index.js
```diff
@@ -1,55 +1,125 @@
...
-    header: PropTypes.node,
-    i18n: PropTypes.object.isRequired,
-  };
+type UIStateType = {|
+  showAllContent: boolean,
+  readMoreExpanded: boolean,
+|};
+
+type Props = {|
+  children: any,
```
bobsilverberg: Why does this need to be `any`? Could it be 'React.Node`?

rebmullin: @bobsilverberg I used this to fix the issue mentioned above:
https://github.com/mozilla/addons-frontend/pull/5874#discussion_r217543489

I seems I can also do the following if it's better:
`children: React.Element<any> | Object,`

bobsilverberg: I just tried your code with just `React.Element<any>` (i.e., not `React.Element<any> | Object`) and I didn't get any Flow errors. Does that not work for you? `Object` isn't really much better than `any`.

rebmullin:
thanks @bobsilverberg ++ for the review on this one :)

Issue #5598, opened 2018-07-13
Move `AddonReviewListItem` logic into the `UserReview` component


willdurand:
The `AddonReviewListItem` and `UserReview` are similar. In fact, `AddonReviewListItem` relies on the `UserReview`, which is a "read-only" component to display user reviews. The `UserReview` component has been introduced in #5416 to avoid duplicate style, extracting style from `AddonReviewListItem`.

We could go a step further by merging both component, so that we can use the same component when displaying user reviews, no matter if we want read-only or read-write mode.

kumar303:
This is technically a dupe of https://github.com/mozilla/addons-frontend/issues/6104 now. Even though https://github.com/mozilla/addons-frontend/issues/6104 was filed later on, I'm going to keep that one open since it will help QA understand which pages to test for this change.

PR #5882, opened 2018-08-08
Re-style developer responses on the user profile page

### LINKED ISSUE #5684: Review/rating replies are shown on profile

- Reply to review/rating on an extension
- Go to profile
- Scroll to "My reviews" section

### What happened?
Replies are visible as if they are reviews the user wrote.

Replies are not reviews, they do not have a star rating associated with them and leaving a review on your own extension would be weird, right?

https://addons.mozilla.org/en-US/firefox/user/freaktechnik/

image

which are replies (to reviews from anonymous users, so I doubt the users will ever get to see the replies :frowning_face: ):

image

willdurand:
Fix #5684

---

Possible thanks to https://github.com/mozilla/addons-server/pull/9099.

This PR updates the style of a review on the user profile page, when this review is a developer response.

Before:

screen shot 2018-08-08 at 15 42 50

After:

screen shot 2018-08-08 at 15 42 29

No change on the reviews page:

screen shot 2018-08-08 at 15 42 41


src/amo/components/AddonReviewListItem/index.js
```diff
@@ -270,7 +270,7 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
     );
 
     return (
-      <UserReview review={review} byLine={byLine} showRating={isReply}>
+      <UserReview review={review} byLine={byLine} showRating={!isReply}>
```
willdurand: I completely missed that but that was illogical... 👻

kumar303:
For consistency, why not just make developer replies on the profile look like a normal developer reply but without the indent?

43840797-de30fcb6-9b21-11e8-947c-501d9f93b3de

You could re-use the same *Developer response* header with the left pointing arrow. This should allow you to merge the two so they are sharing the same component code, no?

willdurand:
> You could re-use the same Developer response header with the left pointing arrow. This should allow you to merge the two so they are sharing the same component code, no?

@kumar303 I don't think I have the original rating so I cannot display it, which means the left pointing arrow would point to nothing. And we don't want the action links so far. When fetching reviews for a user, we don't get a lot of information.

willdurand:
> The arrow (to me) feels more like an icon to indicate that it's a reply -- that's all. I guess that's fine, though. I don't like how the byline is used as a header but I already dislike how the profile ratings look different than add-on ratings. I feel like they should look the same. Treating them differently is going to create complications when we try to add features to the review list.
> 
> (I'm just thinking out loud here.)

There is #5598 that aims at unifying both components (for a single review) but not for the two lists. I don't know what would be the best approach.

I guess I could tryo to fix #5598 and see how I can leverage this work to fix #5684.

src/amo/components/UserProfile/styles.scss
```diff
@@ -120,3 +120,13 @@ $avatar-size: 64px;
 .UserProfile-reviews {
   margin-top: 0;
 }
+
+.UserProfile-review--is-reply {
+  background-color: transparentize($blue-50, 0.95);
```
kumar303: The more features we add, the more we duplicate styles. I would like to move towards unifying these components soon.

tests/unit/amo/components/TestUserProfile.js
```diff
@@ -808,21 +809,60 @@ describe(__filename, () => {
...
+    const root = renderUserProfile({ location, params, store });
+
+    expect(root.find('.UserProfile-review')).toHaveLength(reviews.length);
+    expect(root.find('.UserProfile-review')).toHaveClassName(
+      'UserProfile-review--is-reply',
+    );
+
+    const byLine = shallow(root.find(UserReview).prop('byLine'), {
+      // The `Link` component needs the store.
+      context: { store },
```
kumar303: You may be able to avoid it with this (but maybe not).

```js
const byLine = shallow(<div>{root.find(UserReview).prop('byLine')}</div>);
```

PR #5885, opened 2018-08-08
I5875 static theme not appearing on shelves fix

rebmullin:
fixes mozilla/addons#12177 

I decided to adjust the filters only here (and hopefully not mess with this https://github.com/mozilla/addons/issues/12177 until after the migration as it gets messy if we change/modify this)

note: On dev you can see static themes appearing on homepage and themes landing shelves and cat causes with these changes

### DIFF
--- a/tests/unit/amo/sagas/testLanding.js
+++ b/tests/unit/amo/sagas/testLanding.js
@@ -65,7 +65,7 @@ describe(__filename, () => {
         pageSize: LANDING_PAGE_EXTENSION_COUNT,
       },
       {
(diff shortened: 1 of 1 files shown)


rebmullin:
note: I am getting this the following in the tests and I am not sure how to fix this:

`Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.`

I do vaguely remember seeing this before but it went away. I also tried upping time just to see and i still this message... any idea on how to fix this?

rebmullin:
@kumar303, thanks for looking this one.

I will open a follow up issue as you suggest..

PR #5896, opened 2018-08-09
[WIP] Setup yarn offline

### LINKED ISSUE #3034: Verify integrity of dependencies in production with yarn
After https://github.com/mozilla/addons-frontend/issues/1391 we will be using `yarn` to install dependencies in production. 

Let's make sure all packages get installed without any surprises.

* A production install should never upgrade a package or its dependencies, even if a sub-dependency is declared with a loose version range. Please see [this detailed example](https://github.com/mozilla/addons-frontend/pull/5897#issuecomment-419574560) of what we should prevent.
* Verify the integrity of all packages. If one gets tampered with, the install step should fail. (Does yarn already do this by default? Maybe we need [`yarn check`](https://yarnpkg.com/en/docs/cli/check))

This issue is about **preventing malware on NPM from creeping into a production build** (see the [eslint malware scenario](https://www.theregister.co.uk/2018/07/12/npm_eslint/) as an example). It would still be possible for a developer to unknowingly introduce malware when upgrading a dependency -- that problem is out of scope.

willdurand:
Fix #3034

---

- [x] setup offline mode
- [x] use offline mode everywhere


willdurand:
So I can't make it work, it says:

```
error Can't make a request in offline mode ("https://registry.yarnpkg.com/babel-polyfill/-/babel-polyfill-6.26.0.tgz")
```

The package is checked in though, so I don't understand...

PR #5897, opened 2018-08-09
use yarn.lock when building the apps

### LINKED ISSUE #3034: Verify integrity of dependencies in production with yarn

### LINKED ISSUE #6194: Make deploy builds more predictable with yarn.lock
Related to #3034 

Currently, we do not use the `yarn.lock` file to download/install the dependencies in the `Dockerfile`, which means the build process does not rely on the `yarn.lock` file at all.

We should fix this by using the `yarn.lock` file when running `yarn install`.

For QA: both AMO and Disco Pane should be deployed without any problem (it will be either fine or completely broken).

willdurand:
Fix #6194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ cache:
 # Allow to display job names on Travis-CI, see:
 # https://github.com/travis-ci/travis-ci/issues/5898#issuecomment-362490313
 script: FORCE_COLOR=0 yarn $COMMAND
+
+# See: https://github.com/mozilla/addons-frontend/issues/3034
+install: yarn install --pure-lockfile
+
 jobs:
   include:
     # Test the build process.
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM node:6-slim
 
 # Install node_modules into a different directory to avoid npm/npm#9863.
 RUN mkdir -p /srv/node
-ADD package.json /srv/node/
+COPY package.json /srv/node/
+COPY yarn.lock /srv/node/
 WORKDIR /srv/node
 
 # This file has been downloaded from: https://dl.yarnpkg.com/debian/pubkey.gpg
@@ -28,7 +29,7 @@ RUN buildDeps=' \
     # apt-get purge -y $buildDeps && \
     rm -rf /var/lib/apt/lists/*
 
-ADD . /srv/code/
+COPY . /srv/code/
 WORKDIR /srv/code
 
 # The V2 Pipeline expects version.json to be located in /app/
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN buildDeps=' \
     # install deps
     apt-get update -y && \
     apt-get install -y --no-install-recommends $buildDeps && \
-    yarn install && \
+    yarn install --pure-lockfile && \
     # cleanup
     # apt-get purge -y $buildDeps && \
     rm -rf /var/lib/apt/lists/*


willdurand:
I cannot make the `--offline` approach work, so I guess this is a good start.

willdurand:
> Let me know if I'm missing something but I don't think this solves #3034 (I just updated its description btw). Aside from not solving the issue, I don't see any benefit to adding --pure-lockfile.
> 
> As far as I can tell, --pure-lockfile only stops yarn install from generating a lock file, that's it.

It is not clear to me what this option does actually. I re-read the GH issues and it now seems pointless: I don't understand why people keep mentioning it when it comes to looking for "how to install dependencies according to the lock file?". Anyway...

> I cloned the repo in a temp directory, edited the version of a dependency in package.json, and ran yarn install --pure-lockfile. It still tried to install the new dependency version from package.json.

That sounds like the correct behavior to me given that you change the `package.json` file without changing the `yarn.lock` file. More info here: https://github.com/yarnpkg/yarn/issues/5270#issuecomment-362016699. Those STR are unfortunately not matching what we want to achieve in #3034 IIRC.

Both files should be synchronized, and that's why we keep pushing `Sync yarn.lock with master` commits time to time. `yarn install` does whatever needs to be done to make sure `package.json` is satisfied, which should be the case because we synchronize both files all the time.
 
> Maybe we need --frozen-lockfile? I think that will fail if package.json is introducing new modules/versions. Getting a failure like this in production/dev/stage is too late, though.

Adding such an option on Travis means we can stop using Greenkeeper because all PRs will be red :/

We could potentially add a Travis job for builds on `master`, but that means we cannot merge Greenkeeper PRs directly from the web UI (which is fine to me, though).

> Anyway, I'm not convinced that either option will solve this part of #3034 :
> 
>     A production install should never upgrade a package (or its dependencies) beyond what's in our lock file.

I tend to disagree: that will be the case with this PR given that it adds the `yarn.lock` file when installing the dependencies in the Docker image build process.

> I think we need to consider a new approach. We either need an offline mirror or something like shrinkwrap.

Offline mode is probably too much honestly, we would have to commit 50-60MB of `tar` files and maintain them. This is probably the best paranoid option, though.

We currently don't use the `yarn.lock` file at all, we are very lucky that everything works. Having that should partially fix the issue mentioned in the OP.

The second part could be fixed by running `yarn check` during the Docker image build, which is used by the UI tests IIRC so we would see the errors in travis-ci.

WDYT?

willdurand:
@kumar303 ^

bobsilverberg:
@willdurand Do we need to keep this PR open, or are we looking at other alternatives to solve the issue?

willdurand:
> @willdurand Do we need to keep this PR open, or are we looking at other alternatives to solve the issue?

I think it should stay open until I get the chance to talk to @kumar303 about it again, hopefully I'll have an update by the end of this week.

kumar303:
Sorry for missing the last comment.

I'll come back to this requirement again:

> A production install should never upgrade a package or its dependencies beyond what's in our lock file (or something else that declares all package versions).

You mentioned in the last comment that your patch will fix this. How?

Let's take `isomorphic-fetch` as an example. We require a specific version of it (not a loose version):

https://github.com/mozilla/addons-frontend/blob/9f14a306602179c0ac4f38b1335a77b234b4ef0d/package.json#L184

But in the lock file, its dependency on `node-fetch` is loose (if I understand how lock files work--maybe I don't!):

https://github.com/mozilla/addons-frontend/blob/9f14a306602179c0ac4f38b1335a77b234b4ef0d/yarn.lock#L4992-L4997

What if an attacker hacks an account and publishes `1.0.2` of `node-fetch` with a coin miner in it? Won't the next production install of `addons-frontend` pull in `1.0.2` of `node-fetch`? If not, how does your patch prevent that?

kumar303:
> A production install should never upgrade a package or its dependencies beyond what's in our lock file (or something else that declares all package versions).

I also just realized that I worded this incorrectly. What I mean is to never upgrade any *sub-dependency* beyond its version that existed at the time we pinned the top-level *dependency*.

We accomplish this in Python by pinning *all* package versions (even sub-dependencies) and installing the requirements with [`--no-deps`](https://github.com/mozilla/addons-server/blob/f922c5e1f212f71eceac8d3fd5a620d1b8a17032/Dockerfile.deploy#L73).

willdurand:
> You mentioned in the last comment that your patch will fix this. How?

@kumar303 mainly by using the `yarn.lock` file, this file is not used yet so it is a miracle that things work and get installed properly. It can only be better. Yarn folks say that the `yarn.lock` file ensures reproducible builds (and yarn in general, but does that mean it is SemVer-compliant? Or better?)

According to [this issue](https://github.com/yarnpkg/yarn/issues/4814), `frozen-lockfile` will only fail if `package.json` and `yarn.lock` files are not in sync, which is not what we want either.

> I also just realized that I worded this incorrectly. What I mean is to never upgrade any sub-dependency beyond its version that existed at the time we pinned the top-level dependency.
> 
> We accomplish this in Python by pinning all package versions (even sub-dependencies) and installing the requirements with --no-deps.

Offline mirror is probably the feature that allows that: https://yarnpkg.com/blog/2016/11/24/offline-mirror/.

To sum up, I really think we need that `yarn.lock` file as part of the build process, that will be better than without. We can still open a separate issue for the paranoid mode (a.k.a. pinning all package versions of everything). WDYT?

kumar303:
> I really think we need that yarn.lock file as part of the build process, that will be better than without. 

I agree. If you want to merge your patch as is then that's fine with me. I don't think it should close https://github.com/mozilla/addons-frontend/issues/3034 but you could make another issue to track it like "Make deploy builds more predictable with `yarn.lock`" or something.

> We can still open a separate issue for the paranoid mode (a.k.a. pinning all package versions of everything).

I intended https://github.com/mozilla/addons-frontend/issues/3034 to be that issue and I think the bullet point about making sure no packages can be upgraded captures that intent. I will update the issue to clarify what I meant in that sentence.

willdurand:
@kumar303 alrighty, updated PR title/desc. and created a different issue. r?

PR #5903, opened 2018-08-09
I5682-static themes install/enable update

### LINKED ISSUE #5681: Update logic in disco pane so static themes are installed similiarly to lightweight themes

rebmullin:
fixes #5681


src/core/components/AMInstallButton/index.js
```diff
@@ -119,14 +120,16 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
...
 
     event.preventDefault();
     event.stopPropagation();
 
     await install();
 
-    if (addon.type === ADDON_TYPE_STATIC_THEME) {
+    const isEnabled = await getCurrentStatus();
+
+    if (addon.type === ADDON_TYPE_STATIC_THEME && !isEnabled) {
```
willdurand: I think we should call `getCurrentStatus()` inside the `if`, because it is only needed for static themes.

src/core/installAddon.js
```diff
@@ -333,6 +334,29 @@ export class WithInstallHelpers extends React.Component {
     }
   }
 
+  getCurrentStatus() {
```
willdurand: I believe this method should be named `isAddonEnabled` because we only return the `isEnabled` value.

src/core/installAddon.js
```diff
@@ -333,6 +334,29 @@ export class WithInstallHelpers extends React.Component {
...
+    const {
+      _addonManager,
+      addon: { guid },
+      hasAddonManager,
+    } = this.props;
+
+    if (!hasAddonManager) {
+      log.info('No addon manager, cannot set add-on status');
```
willdurand: I am not sure we need this because other methods don't have it, since we'll use them only when we have mozAddonManager.

src/core/installAddon.js
```diff
@@ -333,6 +334,29 @@ export class WithInstallHelpers extends React.Component {
...
+      return Promise.resolve();
+    }
+
+    log.info('Getting add-on status');
+    return _addonManager
+      .getAddon(guid)
+      .then((addon) => {
+        return addon.isEnabled;
```
willdurand: I see that this code is the same as `setCurrentStatus`, which is called on every prop change.

I did not know we were using `getAddon` before so I wonder if this is needed. Maybe relying on `status` is enough.

src/core/components/AMInstallButton/index.js
willdurand: Given that the status is likely up-to-date, can we maybe just do:

```
if (addon.type === ADDON_TYPE_STATIC_THEME && status !== ENABLED) {
  ...
}
```

src/core/installAddon.js
rebmullin: yep i was going to rename this : )

src/core/installAddon.js
willdurand: Ok, I missed the condition on the GUID. That makes sense performance-wise.

> I don't know that if we'll we see ENABLED right after install. I compared by testing install of a regular extension and its status right after install shows UNINSTALLED...which doesn't seem right to me..? So there may be a deeper issue here to fix.

oO, that does not sound right.

>  For now maybe I can check for the DISABLED state (in case it gets fixed and is working more like an extension and show UNINSTALLED). if that all makes sense : )

We can probably live with a new method in `installAddon`. I'd like to investigate more the potential bug above.

rebmullin: More on this

> I don't know that if we'll we see ENABLED right after install. I compared by testing install of a regular extension and its status right after install shows UNINSTALLED...which doesn't seem right to me..? So there may be a deeper issue here to fix.

In the render, it does show a more accurate status. For a regular extension it says INSTALLED. If you refresh the page it shows ENABLED.

I think after the install, its just not updated at that point yet so it actually still shows the older status..

src/disco/components/Addon/index.js
```diff
@@ -45,9 +45,9 @@ const CSS_TRANSITION_TIMEOUT = { enter: 700, exit: 300 };
 
 export const installStaticTheme = async (event, props) => {
   event.preventDefault();
-  const { status } = props;
+  const { hasAddonManager, status } = props;
 
-  if (status === UNINSTALLED) {
+  if (hasAddonManager && status === UNINSTALLED) {
```
willdurand: ah yes, I think we should not have to call this method if we have no mozAddonManager, but it should not be done here IMO. Can we prevent the call where we bind the onClick event?

tests/unit/core/TestInstallAddon.js
```diff
@@ -601,6 +601,30 @@ describe(__filename, () => {
...
+          ...fakeAddon,
+          _addonManager: fakeAddonManager,
+        });
+        const { root } = renderWithInstallHelpers({
+          ...addon,
+          addon: { guid },
+          _addonManager: fakeAddonManager,
+        });
+
+        guid = root.instance().props.guid;
```
rebmullin: to me this seemed a little weird that I needed to do this but it doesn't work without this..?

src/disco/components/Addon/index.js
```diff
@@ -242,6 +244,30 @@ export class AddonBase extends React.Component {
     }
   };
 
+  installStaticTheme = async (event) => {
```
rebmullin: Originally I had this outside of the class so I could spy on it being called but I decided to move back inside the class b/c I couldn't seem to spy on the install and enable methods which I thought was more import to spy on (rather than the function itself). hopefully this is okay..

rebmullin:
> I think after the install, its just not updated at that point yet so it actually still shows the older status..

@willdurand, given the above, I am going to keep this isAddonEnabled method. let me know if you see any issues w/ this now.

src/disco/components/Addon/index.js
rebmullin: i thought about doing that but then there will still be a link where we want to do event.preventDefault...

unless I make it not a link with this condition...is that better?

src/core/installAddon.js
```diff
@@ -333,6 +334,23 @@ export class WithInstallHelpers extends React.Component {
...
+    } = this.props;
+
+    log.info('Getting add-on enabled status');
+    return _addonManager
+      .getAddon(guid)
+      .then((addon) => {
+        return addon.isEnabled;
+      })
+      .catch((error) => {
+        log.error(`Caught error from addonManager: ${error}`);
```
willdurand: Should we return `false`?

willdurand: Can we move to async/await syntax?

src/core/installAddon.js
```diff
@@ -298,6 +298,7 @@ export class WithInstallHelpers extends React.Component {
     WrappedComponent: PropTypes.func.isRequired,
     _addonManager: PropTypes.object,
     _tracking: PropTypes.object,
+    addon: PropTypes.object,
```
willdurand: We have the guid in the props already, no?

src/disco/components/Addon/index.js
```diff
@@ -61,6 +62,7 @@ export class AddonBase extends React.Component {
     iconUrl: PropTypes.string,
     install: PropTypes.func.isRequired,
     installTheme: PropTypes.func.isRequired,
+    isAddonEnabled: PropTypes.bool,
```
willdurand: that's a function, no? maybe the name of the function is not good enough as it implies a boolean.

src/disco/components/Addon/index.js
```diff
@@ -242,6 +244,30 @@ export class AddonBase extends React.Component {
...
+    event.preventDefault();
+    const {
+      enable,
+      hasAddonManager,
+      isAddonEnabled,
+      install,
+      status,
+    } = this.props;
+
+    if (hasAddonManager && status === UNINSTALLED) {
```
willdurand: do we really need to check the addonManager?

src/disco/components/Addon/index.js
```diff
@@ -137,30 +139,30 @@ export class AddonBase extends React.Component {
 
       let imageUrl = getPreviewImage(addon);
 
-      if (!imageUrl && type === ADDON_TYPE_THEME) {
+      let headerLinkProps = {
+        className: 'theme-image',
+        href: '#',
```
willdurand: we should probably link to the static theme, we have it no?

src/core/installAddon.js
rebmullin: > Can we move to async/await syntax?

that would be cool : ) i actually thought about that but was trying to keep consistent with other funcs..

src/disco/components/Addon/index.js
rebmullin: hmm..good point..

src/disco/components/Addon/index.js
rebmullin: the link here is supposed to install/enable (like lightweight theme currently does)

src/disco/components/Addon/index.js
rebmullin: i thought it would good to ck to be sure. We do pass this val into the button so thought there could be scenario would it's false for some reason here too, no?

src/core/installAddon.js
rebmullin: linting was complaining about this so I added this here

rebmullin: actually after changing to async/await, it seems that we can remove..

src/disco/components/Addon/index.js
willdurand: even if we disable the mozAddonManager?

rebmullin: well right now..the way it's set up it wouldn't do anything. I guess we could link the header in this case to the addon detail page

On a side but related note if mozAddonManager is false, then this value would be passed to the buttons and I don't see the buttons working here either in this case. Is this expected?

willdurand: The install button works with or without the `mozAddonManager` by offering the link to the extension in a `href` prop.

src/core/installAddon.js
```diff
@@ -333,6 +333,22 @@ export class WithInstallHelpers extends React.Component {
...
+      addon: { guid },
+    } = this.props;
+
+    log.info('Getting add-on enabled status');
+    try {
+      const addon = await _addonManager.getAddon(guid);
+      return addon.isEnabled;
+    } catch (error) {
+      log.error(`Caught error from addonManager: ${error}`);
+      return false;
```
willdurand: Can we move the return statement outside the catch block? so that it shows that this function returns `false` by default.

src/disco/components/Addon/index.js
willdurand: I think what we could do is that if `hasAddonManager` is `false`, then we do not surround the `` with a link. No link, no action, for both lightweight and static themes.

That would also allow you to get rid of the test in `installStaticTheme()`.

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -393,19 +393,48 @@ describe(__filename, () => {
...
+      getAddonEnabledStatus: sinon.stub().returns(false),
+    });
 
     const event = createFakeEvent();
 
     const installButton = root.find('.AMInstallButton-button');
     await installButton.simulate('click', event);
 
-    sinon.assert.calledOnce(install);
+    await sinon.assert.calledOnce(install);
```
willdurand: oh, we really have to `await` here?

tests/unit/core/TestInstallAddon.js
willdurand: That does not sound good 😛

src/amo/components/Addon/index.js
```diff
@@ -69,6 +69,7 @@ export class AddonBase extends React.Component {
     enable: PropTypes.func.isRequired,
     errorHandler: PropTypes.object.isRequired,
     getClientCompatibility: PropTypes.func,
+    getAddonEnabledStatus: PropTypes.func.isRequired,
```
willdurand: Thanks for renaming this function. I still don't like the name but I don't find a better name..

tests/unit/core/components/TestAMInstallButton.js
rebmullin: yeh it wasn't hitting enable without this..should I be doing something else?

tests/unit/core/TestInstallAddon.js
rebmullin: lol..

actually I just noticed tho I think w/ the changes (I am not sure exactly what), its ok now without...?
I'll remove now...

src/amo/components/Addon/index.js
rebmullin: ok i'll keep for now. We can update if think of something better : )

tests/unit/disco/components/TestAddon.js
```diff
@@ -443,42 +444,103 @@ describe(__filename, () => {
...
+    };
+
+    it('renders the correct image', () => {
+      const root = renderWithStaticTheme();
+      const image = root.find('.Addon-theme-header-image');
       expect(image).toHaveProp('src', fullImage);
       expect(image).toHaveProp('alt', `Preview of ${newAddonName}`);
     });
+
+    it("calls install and enable helper functions when clicking on the static theme's header image if hasAddonManager is true", async () => {
```
rebmullin: I am wondering if it's better to group some of these tests here in a `hasAddonManager is false` (or true) describe?

src/amo/components/Addon/index.js
willdurand: Maybe `getEnabledState`? or we can revert back to `isAddonEnabled`...

src/core/components/AMInstallButton/index.js
```diff
@@ -47,6 +47,7 @@ type Props = {|
   defaultInstallSource: string,
   disabled: boolean,
   enable: () => Promise<any>,
+  getAddonEnabledStatus: () => Promise<any>,
```
willdurand: I think we can use `Promise<boolean>`, no?

src/core/installAddon.js
```diff
@@ -333,6 +333,23 @@ export class WithInstallHelpers extends React.Component {
     }
   }
 
+  async getAddonEnabledStatus() {
+    const {
+      _addonManager,
+      addon: { guid },
+    } = this.props;
+
+    log.info('Getting add-on enabled status');
```
willdurand: Not sure we really need this `log.info`

src/disco/components/Addon/index.js
```diff
@@ -242,6 +248,23 @@ export class AddonBase extends React.Component {
...
+    event.preventDefault();
+    const { enable, getAddonEnabledStatus, install, status } = this.props;
+
+    if (status === UNINSTALLED) {
+      await install();
+
+      // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1477328
+      // on why we might explicitly be calling the enable function
+      // after install
+      const isEnabled = await getAddonEnabledStatus();
```
willdurand: Maybe this could go outside the `status === UNINSTALLED` block

src/core/installAddon.js
```diff
@@ -333,6 +333,23 @@ export class WithInstallHelpers extends React.Component {
...
+    try {
+      const addon = await _addonManager.getAddon(guid);
+      return addon.isEnabled;
+    } catch (error) {
+      log.error(`Caught error from addonManager: ${error}`);
```
willdurand: I'd rather write:


```
log.error('could not determine whether the add-on was enabled', error);
```

src/disco/components/Addon/index.js
```diff
@@ -242,6 +248,24 @@ export class AddonBase extends React.Component {
...
+
+    if (status === UNINSTALLED) {
+      await install();
+    }
+
+    const isEnabled = await getAddonEnabledStatus();
+
+    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1477328
+    // on why we might explicitly be calling the enable function
+    // after install
```
willdurand: the `See: <link` is enough I think

tests/unit/core/components/TestAMInstallButton.js
willdurand: I see. I think we could avoid the `await` here and above with `setImmediate()`.

willdurand: The problem I see here is that the test case above does not `await` the `install` and therefore no matter what happens the assertion `sinon.assert.notCalled(enable);` will always be correct. That means a bug will never be caught.

willdurand: @kumar303 would you have an idea to make it better?

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -393,19 +393,48 @@ describe(__filename, () => {
...
     sinon.assert.calledOnce(enable);
     sinon.assert.calledOnce(event.preventDefault);
     sinon.assert.calledOnce(event.stopPropagation);
   });
 
+  it("does not call the `enable` after the `install `helper for a static theme if it's already enabled", async () => {
```
willdurand: enable helper?

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,29 @@ describe(__filename, () => {
...
+        });
+        const { root } = renderWithInstallHelpers({
+          ...addon,
+          addon: { guid },
+          _addonManager: fakeAddonManager,
+        });
+
+        const { getAddonEnabledStatus } = root.instance().props;
+
+        return getAddonEnabledStatus().then((addonObject) => {
```
willdurand: we can probably switch to async/await here too.

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,29 @@ describe(__filename, () => {
         .returns(false);
     });
 
+    describe('getAddonEnabledStatus', () => {
+      it('calls addonManager.getAddon()', () => {
```
willdurand: > returns true when the add-on is enabled

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,29 @@ describe(__filename, () => {
...
+          _addonManager: fakeAddonManager,
+        });
+
+        const { getAddonEnabledStatus } = root.instance().props;
+
+        return getAddonEnabledStatus().then((addonObject) => {
+          sinon.assert.calledWith(fakeAddonManager.getAddon, addonObject.guid);
+        });
+      });
+    });
```
willdurand: We need a test case for when there is an error (`it('returns false when there is an error'`)

tests/unit/disco/components/TestAddon.js
```diff
@@ -443,42 +444,102 @@ describe(__filename, () => {
...
               image_url: fullImage,
             },
           ],
-          type: ADDON_TYPE_STATIC_THEME,
         },
+        status: UNINSTALLED,
+        install,
+        enable,
+        hasAddonManager: true,
+        getAddonEnabledStatus: sinon.stub().returns(false),
```
willdurand: That should probably be `resolves` instead of `returns` because it is a promise

tests/unit/disco/components/TestAddon.js
```diff
@@ -443,42 +444,102 @@ describe(__filename, () => {
...
+
+    it("does not call enable helper functions when clicking on the static theme's header image if addon is already enabled", async () => {
+      install = sinon.spy();
+      enable = sinon.spy();
+      const root = renderWithStaticTheme({
+        getAddonEnabledStatus: sinon.stub().returns(true),
+      });
+      const themeImage = root.find('.theme-image-link');
+      await themeImage.simulate('click', createFakeEvent());
+      await sinon.assert.called(install);
```
willdurand: that's so weird... Maybe we should not use `simulate('click')` and call `await root.instance().installStaticTheme()` directly.. (cc @kumar303)

tests/unit/core/components/TestAMInstallButton.js
kumar303: > @kumar303 would you have an idea to make it better?

This is the first time I've seen `sinon.assert.calledOnce()` used as a promise and I can't find any documentation on this behavior so I don't think we should rely on it. My guess is that it only works by chance in that it waits for one more tick on the event loop. Otherwise, what would cause the `calledOnce()` promise to resolve? It is not clear.

I would like to see these tests rewritten with proper async handling.

The problem is that the click handlers rely on promises. Here is a suggestion that doesn't fully cover hooking up the click handlers but I'd be fine with it since it's a simple change from what you have already and it's probably good enough.

```diff
diff --git a/tests/unit/core/components/TestAMInstallButton.js b/tests/unit/core/components/TestAMInstallButton.js
index 3d1b60693..b8bbf2233 100644
--- a/tests/unit/core/components/TestAMInstallButton.js
+++ b/tests/unit/core/components/TestAMInstallButton.js
@@ -403,9 +403,10 @@ describe(__filename, () => {
     const event = createFakeEvent();
 
     const installButton = root.find('.AMInstallButton-button');
-    await installButton.simulate('click', event);
+    const onClick = installButton.prop('onClick');
+    await onClick(event);
 
-    await sinon.assert.calledOnce(install);
+    sinon.assert.calledOnce(install);
     sinon.assert.calledOnce(enable);
     sinon.assert.calledOnce(event.preventDefault);
     sinon.assert.calledOnce(event.stopPropagation);
```

kumar303: > I see. I think we could avoid the `await` here and above with `setImmediate()`.

IMO `setImmediate()` is a hack and we shouldn't invest code in it.

tests/unit/disco/components/TestAddon.js
rebmullin: this works, thank you.... the way @kumar303 suggested below works here too. is one way better than the other?

rebmullin:
> Are there any other TestAMInstallButton tests (not related to this patch) that need updating for the async behavior?

I'll take a look..

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,48 @@ describe(__filename, () => {
...
+        });
+        const { root } = renderWithInstallHelpers({
+          ...addon,
+          addon: { guid },
+          _addonManager: fakeAddonManager,
+        });
+        const { isAddonEnabled } = root.instance().props;
+        const isEnabled = await isAddonEnabled();
+
+        expect(isEnabled).toEqual(true);
```
rebmullin: is this acceptable/sufficient? or do i need to show it returning false and true individually?

rebmullin:
Thanks @kumar303 @willdurand for the reviews. Please let me know how's it looking now?

tests/unit/disco/components/TestAddon.js
willdurand: it is fine to use Kumar's proposal (that is: getting the prop and invoking it)

src/disco/components/Addon/index.js
```diff
@@ -242,6 +248,22 @@ export class AddonBase extends React.Component {
     }
   };
 
+  installStaticTheme = async (event) => {
+    event.preventDefault();
```
willdurand: can you put this after the `const { ... } = this.props;`

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -393,19 +393,52 @@ describe(__filename, () => {
     const enable = sinon.spy();
     const install = sinon.spy();
 
-    const root = render({ addon, enable, install });
+    const root = render({
```
willdurand: We should probably update the previous test case too (line 380 of that file)

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -393,19 +393,52 @@ describe(__filename, () => {
...
+    const onClick = installButton.prop('onClick');
+    await onClick(event);
 
     sinon.assert.calledOnce(install);
     sinon.assert.calledOnce(enable);
     sinon.assert.calledOnce(event.preventDefault);
     sinon.assert.calledOnce(event.stopPropagation);
   });
 
+  it("does not call the `enable` helper after the `install `helper for a static theme if it's already enabled", async () => {
```
willdurand: s/`install `helper/`install` helper

tests/unit/core/test_installAddon.js
willdurand: that's fine.

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,49 @@ describe(__filename, () => {
         .returns(false);
     });
 
+    describe('isAddonEnabled', () => {
+      it('returns add-on status', async () => {
+        const fakeAddonManager = getFakeAddonManagerWrapper();
+        let guid;
+        const addon = createInternalAddon({
+          ...fakeAddon,
+          _addonManager: fakeAddonManager,
```
willdurand: Why would the add-on need a `_addonManager`? That does not look good to me.

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,49 @@ describe(__filename, () => {
...
+        });
+        const { root } = renderWithInstallHelpers({
+          ...addon,
```
willdurand: Can you remove this please? We should not do that anymore.

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,49 @@ describe(__filename, () => {
         .returns(false);
     });
 
+    describe('isAddonEnabled', () => {
+      it('returns add-on status', async () => {
```
willdurand: This test would be better as is:


```js
      it('returns true when the add-on is enabled', async () => {
        const fakeAddonManager = getFakeAddonManagerWrapper();
        const addon = createInternalAddon(fakeAddon);

        const { root } = renderWithInstallHelpers({
          addon,
          _addonManager: fakeAddonManager,
        });

        const { isAddonEnabled } = root.instance().props;
        const isEnabled = await isAddonEnabled();

        sinon.assert.calledWith(fakeAddonManager.getAddon, addon.guid);
        expect(isEnabled).toEqual(true);
      });
```

tests/unit/core/test_installAddon.js
rebmullin: I've left this in otherwise the assert doesn't work. You also have kept it below so I assume this is okay..?

tests/unit/core/components/TestAMInstallButton.js
rebmullin: yep i missed that one, thanks!

tests/unit/core/test_installAddon.js
willdurand: the test case I provided above passes without that so I'd like to see it removed. It is not a good idea to add invalid props.

rebmullin: hmm i see this:

```bash
● /Users/rmullin/Sites/www/addons-frontend/tests/unit/core/test_installAddon.js › withInstallHelpers › isAddonEnabled › returns true when the add-on is enabled

    AssertError: expected stub to be called with arguments 

      614 |         const isEnabled = await isAddonEnabled();
      615 | 
    > 616 |         sinon.assert.calledWith(fakeAddonManager.getAddon, addon.guid);
          |                      ^
      617 |         expect(isEnabled).toEqual(true);
      618 |       });
```

you don't see this?

tests/unit/disco/components/TestAddon.js
rebmullin: ok sure I'll update this.. thanks!

tests/unit/core/test_installAddon.js
```diff
@@ -601,6 +601,42 @@ describe(__filename, () => {
...
+
+      it('returns false when there is an error', async () => {
+        const fakeAddonManager = getFakeAddonManagerWrapper({
+          // Resolve a null addon which will trigger an exception.
+          getAddon: Promise.resolve(null),
+        });
+
+        const { root } = renderWithInstallHelpers({
+          addon: createInternalAddon({
+            addon: null,
```
willdurand: this does not work, it should be null directly and not createInternalAddon.

rebmullin: I did initially try this but I see this without doing it this way:

```bash
 TypeError: Cannot read property 'guid' of null

      337 |     const {
      338 |       _addonManager,
    > 339 |       addon: { guid },
          |                ^
      340 |     } = this.props;
```

rebmullin: should I be handling this another way in installAddon?

rebmullin: I see...I'll update..thanks

PR #5905, opened 2018-08-10
Test files as per standard- Fixes #5616

### LINKED ISSUE #5616: Standardize test files
See also: #1260

---

I'd like to have a common naming rule for the test files, currently we have:

- `TestComponent` for a React `Component`
- `test_someThing` for a `someThing` JS file

But it is not really consistent and not always used that way (for instance, we have `testSomeThing` files sometimes).

I suggest we use the two rules above and we stick to them.

nabendu82:
`Fixes #5616`

Changed the test files as per

-    TestComponent for a React Component
-    test_someThing for a someThing JS file


src/amo/actions/reviews.js
```diff
@@ -1,5 +1,5 @@
 /* @flow */
-import invariant from 'invariant';
+import invariant from '../../../../../../../Library/Caches/typescript/2.9/node_modules/@types/invariant';
```
willdurand: I don't think this is needed.

tests/unit/core/i18n/TestProvider.js
```diff
@@ -1,6 +1,6 @@
-import { mount } from 'enzyme';
-import * as React from 'react';
-import PropTypes from 'prop-types';
+import { mount } from '../../../../../../../../Library/Caches/typescript/2.9/node_modules/@types/enzyme';
+import * as React from '../../../../../../../../Library/Caches/typescript/2.9/node_modules/@types/react';
+import PropTypes from '../../../../../../../../Library/Caches/typescript/2.9/node_modules/@types/prop-types';
```
willdurand: This should not be changed.

willdurand:
@nabendu82 I rebased your branch. That's a nice work. I saw a few more files to rename though:

- `tests/unit/Test*.js`
- `tests/unit/testHelpers.js`
- `tests/unit/testSetup.js`
- `tests/unit/core/Test*.js`
- `tests/unit/core/testErrorHandler.js`
- `tests/unit/core/testWithUIState.js`

PR #5911, opened 2018-08-10
Rewrite RatingManager tests with Enzyme

### LINKED ISSUE #1104: Test all React components with Enzyme

willdurand:
Fix #1104

### DIFF
--- a/tests/unit/amo/components/TestRatingManager.js
+++ b/tests/unit/amo/components/TestRatingManager.js
@@ -1,13 +1,6 @@
 import * as React from 'react';
-import { Provider } from 'react-redux';
-import {
-  findRenderedComponentWithType,
-  renderIntoDocument,
-} from 'react-dom/test-utils';
 
-import { setAuthToken } from 'core/actions';
 import { createInternalAddon } from 'core/reducers/addons';
-import { loadCurrentUserAccount } from 'amo/reducers/users';
 import {
   ADDON_TYPE_DICT,
   ADDON_TYPE_EXTENSION,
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestRatingManager.js
```diff
@@ -16,73 +9,76 @@ import {
...
-            <RatingManagerWithI18n {...props} />
-          </Provider>
-        </I18nProvider>,
-      ),
-      RatingManagerBase,
+    return shallowUntilTarget(<RatingManager {...props} />, RatingManagerBase);
+  }
+
+  const createStoreWithLatestReview = ({
+    addon = fakeAddon,
```
kumar303: To make the intention of this more clear, it should be:

```js
addon = createInternalAddon(fakeAddon),
```

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -120,64 +112,74 @@ describe(__filename, () => {
...
+    });
+    const version = { id: 321 };
+
+    const { store } = dispatchSignInActions();
+
+    const root = render({ addon, store, submitReview, version });
+
+    return root
+      .instance()
+      .onSelectRating(5)
```
kumar303: And, again, if we were using sagas this test could simulate an actual click. 🤔

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -537,6 +582,8 @@ describe(__filename, () => {
   });
 
   describe('mapStateToProps', () => {
```
kumar303: Since you ported to `shallowUntilTarget`, these tests can all either be deleted or switched over to working against `render()`. Rejoice! 🌴  

One way to approach it is to start breaking things and see what tests are already covering their logic. You may need to add a few tests where the logic is not fully covered.

```diff
diff --git a/src/amo/components/RatingManager/index.js b/src/amo/components/RatingManager/index.js
index 10087dda6..662a995c0 100644
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -270,7 +270,7 @@ export const mapStateToProps = (state: AppState, ownProps: Props) => {
   }
 
   return {
-    apiState: state.api,
+    //apiState: state.api,
     userReview,
     userId,
   };
diff --git a/tests/unit/amo/components/TestRatingManager.js b/tests/unit/amo/components/TestRatingManager.js
index e771e4bf2..bfb8fda2b 100644
--- a/tests/unit/amo/components/TestRatingManager.js
+++ b/tests/unit/amo/components/TestRatingManager.js
@@ -581,7 +581,7 @@ describe(__filename, () => {
     });
   });
 
-  describe('mapStateToProps', () => {
+  describe.skip('mapStateToProps', () => {
     let store;
 
     function getMappedProps({

```

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -190,144 +192,177 @@ describe(__filename, () => {
...
+          }),
+        );
+      });
   });
 
   it('renders and configures AddonReview after submitting a rating', () => {
-    const userReview = createInternalReview(fakeReview);
-    const FakeAddonReview = sinon.spy(() => <div />);
-    const root = render({ AddonReview: FakeAddonReview, userReview });
+    const FakeAddonReview = () => <div />;
```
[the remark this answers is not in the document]
willdurand: Yeah, I was not sure about what to do here so I left most of the code as is to keep a small-ish diff. I am fine with fixing those stuff though.

tests/unit/amo/components/TestRatingManager.js
willdurand: oops, right

tests/unit/amo/components/TestRatingManager.js
willdurand: > You may need to add a few tests where the logic is not fully covered.

Looks like I did not have to write new test cases.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -190,144 +192,163 @@ describe(__filename, () => {
...
+        expect(root.find(AddonReview)).toHaveLength(1);
 
-      // Now make sure the callback is configured.
-      expect(root.state.showTextEntry).toEqual(true);
-      // Trigger the callback just like AddonReview would
-      // after completion.
-      props.onReviewSubmitted();
-      expect(root.state.showTextEntry).toEqual(false);
-    });
+        // Now make sure the callback is configured.
```
kumar303: I think this comment bitrotted prior to your patch. Maybe just remove it? I have no idea what it means.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -190,144 +192,163 @@ describe(__filename, () => {
...
+        expect(root).toHaveState('showTextEntry', true);
+
+        // Trigger the callback just like AddonReview would after completion.
+        root.instance().onReviewSubmitted();
```
kumar303: This should exercise the configured callback prop instead:

```js
const callback = root.find(AddonReview).prop('onReviewSubmitted');
callback();
```

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -190,144 +192,163 @@ describe(__filename, () => {
...
-    return root.onSelectRating(5).then(() => {
-      sinon.assert.notCalled(FakeAddonReview);
-    });
+    expect(root.find(AddonReview)).toHaveLength(0);
+
+    return root
+      .instance()
+      .onSelectRating(5)
+      .then(() => {
+        root.update();
```
kumar303: When I remove this the test still passes. Can it be removed? `setState()` changes are applied to the shallow wrapper automatically.

Issue #3919, opened 2017-11-15
Show star rating / review summary in review listing


kumar303:
The Add-on review listing should show a summary of star ratings and reviews to help users better assess the quality of the add-on before installing it.

Please check the mocks before implementing but currently it looks like:

<img width="508" alt="screenshot 2017-11-15 11 56 37" src="https://user-images.githubusercontent.com/55398/32852024-5a497ef6-c9fc-11e7-83cc-84980ed97b43.png">

This will look similar to the summary shown in detail pages: https://github.com/mozilla/addons-frontend/issues/3918

pwalm:
Design updated and referenced in this issue: https://github.com/mozilla/addons-frontend/issues/5462

kumar303:
@pwalm : The current listing page has clickable stars for the user to select their rating ([example](https://addons.mozilla.org/en-US/firefox/addon/emoji-cheatsheet/reviews/)). I'm not sure that we can still use a star summary for that and if we can, I think it needs some changes to make it obvious that the stars are clickable.

kumar303:
We discussed this on IRC. The old listing page let you create/edit a rating or review. @pwalm says the new listing page *should not*. It should be the read-only view from the specs.

pwalm:
Correct! Read only everywhere except the rating block on the detail page

ioanarusiczki:
- the rating block is only editable from the detail page once clicked https://addons-dev.allizom.org/en-US/firefox/addon/adblock-edge/reviews/ the page has just read only properties.
- stars are increased compared to the old layout.
- the user must click the Read All link to see the review, the same when it is edited.

Verified fixed on AMO dev with FF61(Win10&Android 8.0)

PR #5913, opened 2018-08-10
Show star rating / review summary in review listing 

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3919

### DIFF
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -10,6 +10,12 @@
 
   .Rating {
     margin: 6px 0 12px;
+    padding: 0 10px;
+
+    @include respond-to(large) {
+      padding-left: 200px;
+      padding-right: 200px;
+    }
   }
 }
 
--- a/src/amo/components/RatingManager/styles.scss
+++ b/src/amo/components/RatingManager/styles.scss
@@ -1,22 +1,14 @@
 @import '~core/css/inc/mixins';
-@import '~core/css/inc/vars';
-@import '~ui/components/Rating/vars';
 
-$button-margin: 13px;
-$button-height: $rating-l-star-size + $rating-l-star-gutter -
-  ($button-margin * 2);
-
-.RatingManager-log-in-to-rate {
+.RatingManager-ratingControl {
+  align-items: center;
+  display: flex;
+  justify-content: center;
   position: relative;
 }
 
 .RatingManager-log-in-to-rate-button {
-  align-items: center;
-  display: flex;
   font-size: $font-size-s;
-  height: $button-height;
-  justify-content: center;
-  margin-top: $rating-top-margin + $button-margin;
   padding: 0;
   position: absolute;
   width: 100%;
--- a/src/amo/components/RatingManager/styles.scss
+++ b/src/amo/components/RatingManager/styles.scss
@@ -1,3 +1,4 @@
+@import '~core/css/inc/mixins';
 @import '~core/css/inc/vars';
 @import '~ui/components/Rating/vars';
 
@@ -29,5 +30,11 @@ $button-height: $rating-l-star-size + $rating-l-star-gutter -
 
   legend {
     word-wrap: break-word;
+  }
+
+  .Rating {
+    @include respond-to(large) {
+      margin: 12px;
+    }
   }
 }
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -140,10 +140,8 @@ export class RatingBase extends React.Component<InternalProps> {
 
     return (
       <div className={allClassNames} title={description}>
-        <span className="Rating-star-group">
-          {this.renderRatings()}
-          <span className="visually-hidden">{description}</span>
-        </span>
+        {this.renderRatings()}
+        <span className="visually-hidden">{description}</span>
       </div>
     );
   }
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -101,7 +101,7 @@ export class RatingBase extends React.Component<InternalProps, StateType> {
     this.setState({ hoveringOverStar: star });
   };
 
-  stopHovering = (event: SyntheticEvent<any>) => {
+  stopHovering = () => {
     this.setState({ hoveringOverStar: null });
   };
 
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -28,13 +28,22 @@ type InternalProps = {|
   i18n: I18nType,
 |};
 
-export class RatingBase extends React.Component<InternalProps> {
+type StateType = {|
+  hoveringOverStar: number | null,
+|};
+
+export class RatingBase extends React.Component<InternalProps, StateType> {
   static defaultProps = {
     className: '',
     readOnly: false,
     styleSize: 'large',
     yellowStars: false,
   };
+
+  constructor(props: InternalProps) {
+    super(props);
+    this.state = { hoveringOverStar: null };
+  }
 
   onSelectRating = (event: SyntheticEvent<HTMLButtonElement>) => {
     event.preventDefault();
@@ -85,19 +94,35 @@ export class RatingBase extends React.Component<InternalProps> {
     );
   };
 
+  onHoverStar = (star: number) => {
+    if (this.props.readOnly) {
+      return;
+    }
+    this.setState({ hoveringOverStar: star });
+  };
+
+  stopHovering = (event: SyntheticEvent<any>) => {
+    this.setState({ hoveringOverStar: null });
+  };
+
   renderRatings() {
     const { readOnly } = this.props;
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
     return [1, 2, 3, 4, 5].map((thisRating) => {
+      let isSelected = thisRating - rating <= 0.25;
+      if (this.state.hoveringOverStar !== null) {
+        isSelected = thisRating <= this.state.hoveringOverStar;
+      }
       const props = {
         className: makeClassName('Rating-star', `Rating-rating-${thisRating}`, {
-          'Rating-selected-star': thisRating - rating <= 0.25,
+          'Rating-selected-star': isSelected,
           'Rating-half-star':
             thisRating - rating > 0.25 && thisRating - rating <= 0.75,
         }),
         key: `rating-${thisRating}`,
+        onMouseEnter: () => this.onHoverStar(thisRating),
         title: this.renderTitle(rating, readOnly, thisRating),
       };
 
@@ -138,7 +163,11 @@ export class RatingBase extends React.Component<InternalProps> {
     );
 
     return (
-      <div className={allClassNames} title={description}>
+      <div
+        className={allClassNames}
+        title={description}
+        onMouseLeave={this.stopHovering}
+      >
         {this.renderRatings()}
         <span className="visually-hidden">{description}</span>
       </div>
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -92,12 +92,11 @@ export class RatingBase extends React.Component<InternalProps> {
 
     return [1, 2, 3, 4, 5].map((thisRating) => {
       const props = {
-        className: makeClassName('Rating-star', {
+        className: makeClassName('Rating-star', `Rating-rating-${thisRating}`, {
           'Rating-selected-star': thisRating - rating <= 0.25,
           'Rating-half-star':
             thisRating - rating > 0.25 && thisRating - rating <= 0.75,
         }),
-        id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
         title: this.renderTitle(rating, readOnly, thisRating),
       };
--- a/src/ui/components/Rating/index.js
+++ b/src/ui/components/Rating/index.js
@@ -92,7 +92,7 @@ export class RatingBase extends React.Component<InternalProps> {
 
     return [1, 2, 3, 4, 5].map((thisRating) => {
       const props = {
-        className: makeClassName('Rating-choice', {
+        className: makeClassName('Rating-star', {
           'Rating-selected-star': thisRating - rating <= 0.25,
           'Rating-half-star':
             thisRating - rating > 0.25 && thisRating - rating <= 0.75,
--- a/src/ui/components/Rating/styles.scss
+++ b/src/ui/components/Rating/styles.scss
@@ -1,14 +1,25 @@
 @import '~core/css/inc/mixins';
 @import './vars';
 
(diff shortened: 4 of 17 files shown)


src/ui/components/Rating/index.js
```diff
@@ -85,20 +94,35 @@ export class RatingBase extends React.Component<InternalProps> {
...
-        className: makeClassName('Rating-choice', {
-          'Rating-selected-star': thisRating - rating <= 0.25,
+        className: makeClassName('Rating-star', `Rating-rating-${thisRating}`, {
+          'Rating-selected-star': isSelected,
           'Rating-half-star':
             thisRating - rating > 0.25 && thisRating - rating <= 0.75,
         }),
-        id: `Rating-rating-${thisRating}`,
         key: `rating-${thisRating}`,
+        onMouseEnter: () => this.onHoverStar(thisRating),
```
kumar303: I couldn't get a pure CSS hover to work due to the `grid-gap` which causes a flicker. I couldn't think of an alternative that didn't involve implementing a custom gap which is a pitfall of doom. Switching to grid fixes so many other layout issues that I think it's an acceptable trade-off.

src/amo/components/AddonReview/styles.scss
```diff
@@ -10,6 +10,12 @@
 
   .Rating {
     margin: 6px 0 12px;
+    padding: 0 10px;
```
kumar303: The values in this file are inconsistent but I was just preventing regressions. This component is going away soon.

src/ui/components/Rating/index.js
```diff
@@ -28,14 +28,23 @@ type InternalProps = {|
...
   static defaultProps = {
     className: '',
     readOnly: false,
     styleSize: 'large',
     yellowStars: false,
   };
 
+  constructor(props: InternalProps) {
+    super(props);
+    this.state = { hoveringOverStar: null };
```
kumar303: I chose `setState` here because I wanted the UI to be snappy when hovering. Switching to `setUIState()` also burdens the caller with having to generate an ID. I dunno, I'm still thinking on it.

src/ui/components/Rating/index.js
```diff
@@ -85,20 +94,35 @@ export class RatingBase extends React.Component<InternalProps> {
...
+
   renderRatings() {
     const { readOnly } = this.props;
     // Accept falsey values as if they are zeroes.
     const rating = this.props.rating || 0;
 
     return [1, 2, 3, 4, 5].map((thisRating) => {
+      let isSelected = thisRating - rating <= 0.25;
+      if (this.state.hoveringOverStar !== null) {
+        isSelected = thisRating <= this.state.hoveringOverStar;
```
rebmullin: can we extract out hoveringOverStar since it's used a few times?

src/amo/components/RatingManager/styles.scss
```diff
@@ -30,4 +23,10 @@ $button-height: $rating-l-star-size + $rating-l-star-gutter -
   legend {
     word-wrap: break-word;
   }
+
+  .Rating {
+    @include respond-to(large) {
+      margin: 12px;
```
rebmullin: maybe we could keep this margin at smaller viewports too? At least on top:

Alt text

what do you think?

src/ui/components/Rating/index.js
```diff
@@ -85,20 +94,35 @@ export class RatingBase extends React.Component<InternalProps> {
...
+      }
       const props = {
-        className: makeClassName('Rating-choice', {
-          'Rating-selected-star': thisRating - rating <= 0.25,
+        className: makeClassName('Rating-star', `Rating-rating-${thisRating}`, {
+          'Rating-selected-star': isSelected,
           'Rating-half-star':
             thisRating - rating > 0.25 && thisRating - rating <= 0.75,
         }),
-        id: `Rating-rating-${thisRating}`,
```
bobsilverberg: Why don't we need an `id`? Is it because we are rendering an html element and not a react component?

kumar303:
> I did have 1 question about the location of the stars on the reviews page - the mock shows it up top. Is that just an old mock?

@rebmullin thanks for the review. The mock you referenced is up to date. Are you saying that the star rating graph does not appear at the top of the page? Can you send me a URL and screenshot? It appears at the top for me: https://screenshots.firefox.com/m65oS801QVj6sXne/localhost

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -517,13 +525,11 @@ describe(__filename, () => {
...
-      dispatchAddon(addon);
-      const root = render();
-
-      const h3 = root.find('.AddonReviewList-header-authors');
+      const header = renderAddonHeader({ addon });
+      const h3 = header.find('.AddonReviewList-header-authors').render();
 
-      expect(h3.render().text()).toEqual('by Hayley');
-      expect(h3.render().find('a')).toHaveLength(0);
+      expect(h3.text()).toEqual('by Hayley');
```
bobsilverberg: Nit: Ideally `Haley` would be in a `const` rather than hardcoded twice.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -540,13 +546,11 @@ describe(__filename, () => {
...
-      dispatchAddon(addon);
-      const root = render();
+      const header = renderAddonHeader({ addon });
+      const h3 = header.find('.AddonReviewList-header-authors').render();
 
-      const h3 = root.find('.AddonReviewList-header-authors');
-
-      expect(h3.render().text()).toEqual('by Chantal, Leroy');
-      expect(h3.render().find('a')).toHaveLength(2);
+      expect(h3.text()).toEqual('by Chantal, Leroy');
```
bobsilverberg: Nit: Ideally `Chantal` and `Leroy` would be in `const`s rather than hardcoded twice.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -540,13 +546,11 @@ describe(__filename, () => {
...
+      expect(h3.find('a')).toHaveLength(2);
```
bobsilverberg: Should the test also be checking that the `href` of the `a` tags are the expected values?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -640,76 +644,100 @@ describe(__filename, () => {
...
-      onReviewSubmitted();
+    it('renders RatingsByStar with an add-on', () => {
+      const addon = { ...fakeAddon };
+      dispatchAddon(addon);
+      const root = render();
+      const ratingsByStar = root.find(RatingsByStar);
 
-      sinon.assert.calledWith(dispatchSpy, clearAddonReviews({ addonSlug }));
+      expect(ratingsByStar).toHaveProp('addon');
+      expect(ratingsByStar.prop('addon')).toMatchObject({ id: addon.id });
```
bobsilverberg: Why not just `expect(ratingsByStar).toHaveProp('addon', addon);`?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -640,76 +644,100 @@ describe(__filename, () => {
...
 
-      const manager = root.find(RatingManager);
-      expect(manager).toHaveProp('onReviewSubmitted');
-      const onReviewSubmitted = manager.prop('onReviewSubmitted');
+    it('renders a fixed star average', () => {
+      dispatchAddon({
+        ...fakeAddon,
+        ratings: {
+          ...fakeAddon.ratings,
+          average: 4.6667,
```
bobsilverberg: Nit: Would it be better to use a `const` for this and in the assertion?

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -640,76 +644,100 @@ describe(__filename, () => {
...
 
-      sinon.assert.calledWith(history.push, {
-        pathname: `/${lang}/${clientApp}${root.instance().url()}`,
-        query: { page: 1 },
+    it('renders whole number star averages', () => {
+      dispatchAddon({
+        ...fakeAddon,
+        ratings: {
+          ...fakeAddon.ratings,
+          average: 4.0,
```
bobsilverberg: Nit: Would it be better to use a `const` for this and in the assertion?

tests/unit/ui/components/TestRating.js
```diff
@@ -227,6 +227,51 @@ describe(__filename, () => {
...
+    );
+  });
+
+  it('finishes hovering on mouseLeave', () => {
+    const root = render();
+
+    const rating = 3;
+    const hoverStar = getStar({ root, rating });
+    hoverStar.simulate('mouseEnter', createFakeEvent());
+
```
bobsilverberg: Would it be useful to have a check assertion in here that the star has the `Rating-selected-star` class, to avoid a possible false positive for the next assertion?

tests/unit/ui/components/TestRating.js
```diff
@@ -238,14 +283,26 @@ describe(__filename, () => {
...
     });
 
+    it('does nothing when you hover over stars', () => {
+      const root = render({ readOnly: true });
+
+      const rating = 3;
+      const hoverStar = getStar({ root, rating });
+      hoverStar.simulate('mouseEnter', createFakeEvent());
+
+      expect(getStar({ root, rating })).not.toHaveClassName(
```
bobsilverberg: This test is checking for a specific behaviour being absent, but the test says "does nothing when you hover over stars", so would it be any better to check that the classNames of the star are the same before and after simulating the hover, rather than checking for the absence of this specific class?

src/ui/components/Rating/index.js
kumar303: The id was just used for testing. It doesn't really make sense to create an ID like this because on a page like the review listing, there could be many versions of `Rating` with `thisRating==4`.

rebmullin:
> The mock you referenced is up to date. Are you saying that the star rating graph does not appear at the top of the page? Can you send me a URL and screenshot? It appears at the top for me: screenshots.firefox.com/m65oS801QVj6sXne/localhost

hey @kumar303, sorry for not being clearer. And actually what I am bringing up might be technically out of scope (of your PR anyway) ..but just to clarify, I was referring to the stars pointed to here:
Alt text

otherwise is looking good to me 👍

kumar303:
@bobsilverberg 

> I've looked over this and I'm a bit confused about why so many changes were necessary to implement this for the review listing.

- The old implementation never worked quite right so lots of components that rendered ratings had to employ hacks to get things to work. These all needed updating and this makes up the bulk of the changes.
- Part of this patch was removing `AddonReviewList`'s ability to open an editor for the rating. When I removed this I simplified the logic and tests but that created a lot of churn in the diff.

> I'm also a bit surprised that we need a grid for the Rating component.

I am open to other implementation ideas but I think grid is a good fit.

Adding big rating stars to the top of the pre-existing `RatingsByStar` component seemed like a simple patch but the first problem I ran into was that `Rating` was implemented with fixed star sizes. It was not possible to match the mocks without adding a new custom size and this got me thinking that the current implementation didn't handle multiple device sizes well at all. 

I couldn't think of a way to make it more responsive with the flex-box implementation -- trying to account for a gutter and dealing with the quirks of percentage widths just didn't seem like a good use of time, especially considering that grid is designed specifically for this type of layout.

Once I got through the grid implementation, I remembered that hovering over stars was achieved with a sibling selector and that won't work over the grid gaps. D'oh. It's unfortunate that I had to resort to hover events but I still think the pros of the grid shine through. 

The stars are now much easier to place inside of content (notice that all negative margin hacks are gone!) and the layout looks a lot better when you resize the screen.

src/ui/components/Rating/styles.scss
```diff
@@ -1,41 +1,52 @@
...
-    background-image: url('./img/closed-star-yellow.svg');
+  @include respond-to(extraExtraLarge) {
+    &.Rating--large {
+      grid-column-gap: 12px;
+    }
   }
+}
+
+.Rating-star {
+  background: url('./img/closed-star-dim-gray.svg') center/contain no-repeat;
```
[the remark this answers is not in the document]
kumar303: They could be! I thought about this too while working on it but didn't want to explode the patch any further. I filed this to use an inline SVG which I think would solve it nicely. ﻿﻿https://github.com/mozilla/addons-frontend/issues/5925

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -100,15 +96,27 @@ describe(__filename, () => {
...
       const location = createFakeLocation();
       const root = render({ addon: null, location });
-      expect(root.find('.AddonReviewList-header-icon img').prop('src')).toEqual(
-        fallbackIcon,
-      );
+      const header = getAddonHeader(root);
+
+      expect(
+        header.find('.AddonReviewList-header-icon img').prop('src'),
+      ).toEqual(fallbackIcon);
```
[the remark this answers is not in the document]
kumar303: Ah, thanks. I was moving fast trying to the get the old code working.

src/ui/components/Rating/index.js
kumar303: Oh, wait, you mean for the JSX? Ah, yes, it does need an `id`. Good catch! The linter didn't catch this because I was setting it in a props object and using a spread for the JSX constructor.

kumar303: Ignore that last comment. Heh. I was confusing `id` with `key`. Each array item already has a `key` so this will keep DOM mutations to a minimum.

src/amo/components/RatingManager/styles.scss
kumar303: Good idea, thanks.

src/ui/components/Rating/index.js
bobsilverberg: Ha, I was confusing `id` with `key` too!

tests/unit/amo/components/TestAddonReviewList.js
kumar303: Because the test doesn't access to the internal add-on object. I added a comment to make it less confusing.

tests/unit/amo/components/TestAddonReviewList.js
kumar303: Since our test files don't use Flow, I think it's a bit safer to hard code conversions like this. The alternative would be to re-implement the `.toFixed(1)` call but that kind of thing can lead to bitrot without static typing.

tests/unit/amo/components/TestAddonReviewList.js
kumar303: Same reason as above.

tests/unit/ui/components/TestRating.js
kumar303: I don't think it would be useful because that's a duplication of the previous test. The false positive wouldn't be an issue since the logic is already covered -- at least one test would fail.

PR #5917, opened 2018-08-11
updating rtl and ltr as constant

### LINKED ISSUE #3362: i18n `getDirection()` helper should rely on constants

championshuttler:
[WIP]
Fixes #3362

how to approach with scss files, should I import constants from `core/constants` or any better approach

### DIFF
--- a/src/core/constants.js
+++ b/src/core/constants.js
@@ -299,3 +299,6 @@ export const RATINGS_MODERATE = 'Ratings:Moderate';
 export const USERS_EDIT = 'Users:Edit';
 // Super powers. It means absolutely all permissions.
 export const ALL_SUPER_POWERS = '*:*';
+
+export const RTL = 'rtl';
+export const LTR = 'ltr';
--- a/src/core/components/ServerHtml/index.js
+++ b/src/core/components/ServerHtml/index.js
@@ -6,6 +6,8 @@ import serialize from 'serialize-javascript';
 import Helmet from 'react-helmet';
 import config from 'config';
 
+import { LTR } from 'core/constants';
+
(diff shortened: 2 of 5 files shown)


willdurand:
> how to approach with scss files, should I import constants from core/constants or any better approach

don't touch the SCSS code I'd say.

championshuttler:
@willdurand haha ok 😜

championshuttler:
Then I think it's ready for first review 😉

src/core/constants.js
```diff
@@ -299,3 +299,6 @@ export const RATINGS_MODERATE = 'Ratings:Moderate';
 export const USERS_EDIT = 'Users:Edit';
 // Super powers. It means absolutely all permissions.
 export const ALL_SUPER_POWERS = '*:*';
+
```
championshuttler: Should I add a comment here like `Left-to-right and right-to-left variables` .

willdurand: nope

willdurand:
There is one occurrence that needs to be changed in `src/core/components/ServerHtml/index.js`

willdurand:
@championshuttler There is one occurrence that needs to be changed in `src/core/components/ServerHtml/index.js`

championshuttler:
Done , next level stupid i am :man_facepalming:

Issue #5926, opened 2018-08-13
The flag a review menu text gets squished on mobile


kumar303:
### Describe the problem and steps to reproduce it:

Go to any add-on review listing on mobile

### What happened?

There are many scenarios where the *Flag* text appears squished, like this:

<img width="322" alt="screenshot 2018-08-13 16 50 17" src="https://user-images.githubusercontent.com/55398/44060068-fb08f00a-9f18-11e8-8488-4474ecdfbf2f.png">

### What did you expect to happen?

It should not get squished

AlexandraMoga:
@kumar303  I think the issue described above is the same as https://github.com/mozilla/addons-frontend/issues/3424 and in both cases the layout issue is caused by an unbroken string in user's display name.

The issue did not seem to reproduce when only 'normal' display names are present on the reviews page:

![image](https://user-images.githubusercontent.com/31961530/44076328-7a9c3f36-9fa8-11e8-88c7-8618fb0b2876.png)

kumar303:
Actually, https://github.com/mozilla/addons-frontend/issues/3424 is similar but different. It's almost fixed! I have a patch.

AlexandraMoga:
Verified on AMO stage with FF61, Win10x64 and Android 8.0
The Flag button looks fine to me. 

![image](https://user-images.githubusercontent.com/31961530/44201480-fd5db580-a151-11e8-8726-470ee4c2b433.png)

PR #5927, opened 2018-08-13
Don't break words in review control links

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5926

**Before**

screenshot 2018-08-13 17 28 37

**After**

screenshot 2018-08-13 21 32 23
screenshot 2018-08-13 21 33 09
screenshot 2018-08-13 21 35 07


kumar303:
ah, yes, we should probably move the buttons below

PR #5936, opened 2018-08-14
Use the collection owner when adding an add-on to a collection from the collection edit screen

### LINKED ISSUE #5935: Use the collection owner when adding an add-on to a collection from the collection edit screen
The function that should allow users with either `Admin:Curation` or `Collections:Contribute` to add add-ons to collections that they do not personally own is currently not working. That's because the code that calls the API assumes that the current user is the owner of the collection. We need to change the code in `CollectionAddAddon` to use the username of the owner of the collection, rather than the username of the current user, when trying to add an add-on to a collection.

bobsilverberg:
Fixes #5935


bobsilverberg:
Thanks @kumar303. I have tested it interactively and it does fix the issue. I've also removed that extra code, thanks for catching that.

Issue #3424, opened 2017-10-10
A user with long name that leaves a review causes layout issues for "Reply to this review" option on device


ValentinaPC:
Prerequisites:
The person who reviewed your add-on has long account name;
User uses an Android device;

Steps to reproduce:
1. Sign in as a developer of an add-on
2. Go to the detail page of your add-on 
3. Click to read all reviews
4. Observe the "Reply to this review" option

Expected results:
There are no display or layout issues.

Actual results:
The link is not displayed, only the reply arrow.

Notes/Issues:
Verified on AMO-dev FF56(Android 7.2.1)
Screenshot for this issue:
![2017-10-10_1315](https://user-images.githubusercontent.com/15685960/31381424-26163b3e-adbd-11e7-9b86-4760bcacaa18.png)

ioanarusiczki:
Verified fixed on AMO dev with FF61(Android 8.0) 

![small screen - button is now visible](https://user-images.githubusercontent.com/33448286/44347362-44281400-a4a0-11e8-9353-347079f8fa3f.png)

PR #5951, opened 2018-08-15
Push review controls to a new line on small screens (fixes Reply to Review layout issues)

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/3424 by moving review controls to a new row on small screens. This has come up before so it will solve a few problems.

<hr>

Here is how it looked **before**. Notice that it's not as bad as the original screenshot in the issue, thanks to https://github.com/mozilla/addons-frontend/issues/5926

screenshot 2018-08-15 10 13 06

<hr>

Here's how it looks **after** the patch:

screenshot 2018-08-15 12 55 47
screenshot 2018-08-15 12 55 59
screenshot 2018-08-15 12 56 23
screenshot 2018-08-15 12 57 03
screenshot 2018-08-15 12 57 41
screenshot 2018-08-15 12 58 05

<hr>

Large screens:

screenshot 2018-08-15 12 57 30
screenshot 2018-08-15 14 30 37

### DIFF
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -31,10 +31,10 @@
 
 .AddonReviewListItem-allControls {
   display: flex;
-  justify-content: end;
   // Push all controls down to a new row on small screens.
   grid-column-end: span 2;
   grid-column-start: 1;
+  justify-content: end;
   margin-top: 6px;
 
   @include respond-to(medium) {
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -5,7 +5,6 @@
 
 .AddonReviewListItem {
   .Rating {
-    align-self: start;
     height: $default-line-height;
   }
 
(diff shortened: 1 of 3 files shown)


src/amo/components/AddonReviewListItem/styles.scss
```diff
@@ -1,15 +1,30 @@
 @import '~photon-colors/photon-colors';
+@import '~core/css/inc/vars';
 @import '~core/css/inc/mixins';
 @import '~amo/css/inc/vars';
 
+.AddonReviewListItem {
+  .Rating {
+    align-self: start;
```
bobsilverberg: This doesn't seem to be needed. Changing it doesn't have any impact on the position of the rating stars. Is it there "just in case"?

kumar303: Huh. I guess it's not needed anymore, I'll remove it. At one point it solved the vertical centering shown in the *before* screenshot up above but that's not a problem anymore.

kumar303:
Thanks @rebmullin -- I was struggling with medium sized screens and I also just noticed that the developer reply doesn't have stars so I'd need even more grid configuration for that one. I don't think the grid approach is working so I took another stab at it with flex and it works a lot better!

Moved to https://github.com/mozilla/addons-frontend/pull/5961

PR #5953, opened 2018-08-15
Display a notification when an add-on is removed from a collection

### LINKED ISSUE #5937: Display a notification when an add-on is removed from a collection
This is a spin-off from #5748, because that issue discusses more than just the notification.

For this issue:
- when a user removes an add-on from a collection, display a notification similar to the one that is displayed when a user adds an add-on to a collection
- the message bar should be grey rather than  green in this case

bobsilverberg:
Fixes #5937

<details>
  <summary>New screen in edit mode</summary>

screenshot 2018-08-15 14 47 33

</details>

<details>
  <summary>New screen with notice about adding an add-on</summary>

screenshot 2018-08-15 14 48 28

</details>

<details>
  <summary>New screen with notice about removing an add-on</summary>

screenshot 2018-08-15 14 48 13

</details>

### DIFF
--- a/src/amo/components/CollectionAddAddon/index.js
+++ b/src/amo/components/CollectionAddAddon/index.js
@@ -11,6 +11,7 @@ import { addAddonToCollection } from 'amo/reducers/collections';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import withUIState from 'core/withUIState';
+import Card from 'ui/components/Card';
 import Notice from 'ui/components/Notice';
 import type { SuggestionType } from 'amo/components/AutoSearchInput';
 import type {
@@ -27,12 +28,18 @@ import './styles.scss';
 export const MESSAGE_RESET_TIME = 5000;
 const MESSAGE_FADEOUT_TIME = 450;
 
+export const addonAddedAction: 'added' = 'added';
+export const addonRemovedAction: 'removed' = 'removed';
+
 type UIStateType = {|
-  addonWasAdded: boolean,
+  completedAddonAction:
+    | typeof addonAddedAction
+    | typeof addonRemovedAction
+    | null,
 |};
 
 const initialUIState: UIStateType = {
-  addonWasAdded: false,
+  completedAddonAction: null,
 };
 
 export type Props = {|
@@ -46,6 +53,7 @@ type InternalProps = {|
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   hasAddonBeenAdded: boolean,
+  hasAddonBeenRemoved: boolean,
   i18n: I18nType,
   setTimeout: Function,
   setUIState: (state: $Shape<UIStateType>) => void,
@@ -65,15 +73,34 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
   };
 
   componentWillReceiveProps(props: InternalProps) {
-    const { hasAddonBeenAdded: hasAddonBeenAddedNew } = props;
-    const { hasAddonBeenAdded } = this.props;
-    if (hasAddonBeenAdded !== hasAddonBeenAddedNew) {
+    const {
+      hasAddonBeenAdded: hasAddonBeenAddedNew,
+      hasAddonBeenRemoved: hasAddonBeenRemovedNew,
+    } = props;
+    const { errorHandler, hasAddonBeenAdded, hasAddonBeenRemoved } = this.props;
+
+    const addStatusChanged = hasAddonBeenAdded !== hasAddonBeenAddedNew;
+    const removeStatusChanged = hasAddonBeenRemoved !== hasAddonBeenRemovedNew;
+
+    if (addStatusChanged) {
       this.props.setUIState({
-        addonWasAdded: props.hasAddonBeenAdded,
+        completedAddonAction: hasAddonBeenAddedNew ? addonAddedAction : null,
       });
     }
 
-    if (hasAddonBeenAddedNew && hasAddonBeenAddedNew !== hasAddonBeenAdded) {
+    if (removeStatusChanged) {
+      this.props.setUIState({
+        completedAddonAction: hasAddonBeenRemovedNew
+          ? addonRemovedAction
+          : null,
+      });
+    }
+
+    if (
+      (addStatusChanged || removeStatusChanged) &&
+      (hasAddonBeenAddedNew || hasAddonBeenRemovedNew)
+    ) {
+      errorHandler.clear();
       this.timeout = this.props.setTimeout(
         this.resetMessageStatus,
         MESSAGE_RESET_TIME,
@@ -82,10 +109,17 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
   }
 
   componentWillUnmount() {
+    this.resetMessageStatus();
+  }
+
+  resetMessageStatus = () => {
+    this.props.setUIState({
+      completedAddonAction: null,
+    });
     if (this.timeout) {
       this.props.clearTimeout(this.timeout);
     }
-  }
+  };
 
   onAddonSelected = (suggestion: SuggestionType) => {
     const { collection, dispatch, errorHandler, filters } = this.props;
@@ -105,31 +139,33 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
         username: collection.authorUsername,
       }),
     );
-    this.props.setUIState({ addonWasAdded: false });
-  };
-
-  resetMessageStatus = () => {
-    this.props.setUIState({
-      addonWasAdded: false,
-    });
+    this.resetMessageStatus();
   };
 
   render() {
     const { errorHandler, i18n, uiState } = this.props;
 
     return (
-      <div className="CollectionAddAddon">
+      <Card className="CollectionAddAddon">
         {errorHandler.hasError() ? (
           errorHandler.renderError()
         ) : (
           <TransitionGroup className="CollectionAddAddon-noticePlaceholder">
-            {uiState.addonWasAdded && (
+            {uiState.completedAddonAction && (
               <CSSTransition
                 classNames="CollectionAddAddon-noticePlaceholder-transition"
                 timeout={MESSAGE_FADEOUT_TIME}
               >
-                <Notice type="success">
-                  {i18n.gettext('Added to collection')}
+                <Notice
+                  type={
+                    uiState.completedAddonAction === addonAddedAction
+                      ? 'success'
+                      : 'generic'
+                  }
+                >
+                  {uiState.completedAddonAction === addonAddedAction
+                    ? i18n.gettext('Added to collection')
+                    : i18n.gettext('Removed from collection')}
                 </Notice>
               </CSSTransition>
             )}
@@ -145,7 +181,7 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
           onSuggestionSelected={this.onAddonSelected}
           selectSuggestionText={i18n.gettext('Add to collection')}
         />
-      </div>
+      </Card>
     );
   }
 }
@@ -158,6 +194,7 @@ export const extractId = (props: Props) => {
 export const mapStateToProps = (state: AppState) => {
   return {
     hasAddonBeenAdded: state.collections.hasAddonBeenAdded,
+    hasAddonBeenRemoved: state.collections.hasAddonBeenRemoved,
   };
 };
 
--- a/src/amo/components/CollectionAddAddon/index.js
+++ b/src/amo/components/CollectionAddAddon/index.js
@@ -32,14 +32,11 @@ export const addonAddedAction: 'added' = 'added';
 export const addonRemovedAction: 'removed' = 'removed';
 
 type UIStateType = {|
-  completedAddonAction:
-    | typeof addonAddedAction
-    | typeof addonRemovedAction
-    | null,
+  addonAction: typeof addonAddedAction | typeof addonRemovedAction | null,
 |};
 
 const initialUIState: UIStateType = {
-  completedAddonAction: null,
+  addonAction: null,
 };
 
 export type Props = {|
@@ -84,15 +81,13 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
 
     if (addStatusChanged) {
       this.props.setUIState({
-        completedAddonAction: hasAddonBeenAddedNew ? addonAddedAction : null,
+        addonAction: hasAddonBeenAddedNew ? addonAddedAction : null,
       });
     }
 
     if (removeStatusChanged) {
       this.props.setUIState({
-        completedAddonAction: hasAddonBeenRemovedNew
-          ? addonRemovedAction
-          : null,
+        addonAction: hasAddonBeenRemovedNew ? addonRemovedAction : null,
       });
     }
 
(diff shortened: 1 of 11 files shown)


src/amo/components/CollectionAddAddon/index.js
```diff
@@ -27,12 +28,18 @@ import './styles.scss';
 export const MESSAGE_RESET_TIME = 5000;
 const MESSAGE_FADEOUT_TIME = 450;
 
+export const addonAddedAction: 'added' = 'added';
+export const addonRemovedAction: 'removed' = 'removed';
+
 type UIStateType = {|
-  addonWasAdded: boolean,
+  completedAddonAction:
```
kumar303: This would read better to me as `addonAction`.

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -65,15 +73,34 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
...
+          : null,
       });
     }
 
-    if (hasAddonBeenAddedNew && hasAddonBeenAddedNew !== hasAddonBeenAdded) {
+    if (
+      (addStatusChanged || removeStatusChanged) &&
+      (hasAddonBeenAddedNew || hasAddonBeenRemovedNew)
+    ) {
+      errorHandler.clear();
```
kumar303: Usually this happens automatically because the saga which is using the error handler will clear it the next time you run the saga. I see why it's necessary here. Hey, it works 🤷‍♀️

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -82,10 +109,17 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
   }
 
   componentWillUnmount() {
+    this.resetMessageStatus();
+  }
+
+  resetMessageStatus = () => {
```
kumar303: `resetMessages`, maybe? I suggest that because clearing the fader timeout doesn't really have to do with status.

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -160,16 +180,73 @@ describe(__filename, () => {
...
+      username: signedInUsername,
+    });
+    const root = render({ setTimeout: window.setTimeout, store });
+
+    expect(root.find(Notice)).toHaveLength(0);
+
+    _addonRemovedFromCollection({ root, store });
+
+    expect(root.find(Notice)).toHaveLength(1);
+    expect(root.find(Notice).children()).toHaveText('Removed from collection');
```
kumar303: This should also test that `type` gets set to `generic`.

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -160,16 +180,73 @@ describe(__filename, () => {
...
-    expect(root.instance().props.uiState.addonWasAdded).toEqual(false);
+    expect(root.instance().props.uiState.completedAddonAction).toEqual(null);
     expect(root.find(Notice)).toHaveLength(0);
   });
 
+  it('clears the errorHandler when an add-on is added', () => {
+    const errorHandler = createStubErrorHandler(
+      new Error('Unexpected API error'),
+    );
+    errorHandler.clear = sinon.spy();
```
kumar303: It would be better to use the sinon feature for this so that we'll see an error in case `clear()` gets renamed in a refactor.

```js
const clearStub = sinon.stub(errorHandler, 'clear');
...
sinon.assert.called(clearStub);
```

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -160,16 +180,73 @@ describe(__filename, () => {
...
+    _addonAddedToCollection({ username: signedInUsername, root, store });
+
+    sinon.assert.called(errorHandler.clear);
+  });
+
+  it('clears the errorHandler when an add-on is removed', () => {
+    const errorHandler = createStubErrorHandler(
+      new Error('Unexpected API error'),
+    );
+    errorHandler.clear = sinon.spy();
```
kumar303: Here, too.

tests/unit/amo/reducers/test_collections.js
```diff
@@ -495,6 +497,22 @@ describe(__filename, () => {
       expect(state.hasAddonBeenAdded).toEqual(false);
     });
 
+    it('sets a hasAddonBeenRemoved flag when beginning to remove an add-on from a collection', () => {
```
kumar303: I think you mean 'it unsets a hasAddonBeenRemoved flag...', right?

tests/unit/amo/reducers/test_collections.js
```diff
@@ -495,6 +497,22 @@ describe(__filename, () => {
...
+          addonId: 1,
+          collectionId: 3,
+          errorHandlerId: 'error-handler',
+          filters: {},
+          slug: 'some-collection',
+          username: 'some-user',
+        }),
+      );
+
+      expect(state.hasAddonBeenRemoved).toEqual(false);
```
kumar303: Since this is the same as what's in `initialState`, it would be better to apply `addonRemovedFromCollection()` to state before applying `removeAddonFromCollection()`.

tests/unit/amo/sagas/test_collections.js
```diff
@@ -763,6 +764,14 @@ describe(__filename, () => {
 
       const fetchAction = await sagaTester.waitFor(expectedFetchAction.type);
       expect(fetchAction).toEqual(expectedFetchAction);
+
+      const expectedRemovedAction = addonRemovedFromCollection();
+
+      const removedAction = await sagaTester.waitFor(
```
kumar303: I would rather see this moved to its own test. In the case of breakage, it will be easier to read the test output and know how many things broke.

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -105,31 +139,33 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
...
-                <Notice type="success">
-                  {i18n.gettext('Added to collection')}
+                <Notice
+                  type={
+                    uiState.completedAddonAction === addonAddedAction
+                      ? 'success'
+                      : 'generic'
+                  }
+                >
+                  {uiState.completedAddonAction === addonAddedAction
```
rebmullin: can we extract out completedAddonAction since it's used a few times? 

Also/or maybe this itself could be a var `const isAdded = uiState.completedAddonAction === addonAddedAction`

tests/unit/amo/components/TestCollectionAddAddon.js
```diff
@@ -160,16 +180,73 @@ describe(__filename, () => {
...
-    expect(root.instance().props.uiState.addonWasAdded).toEqual(true);
+    expect(root.instance().props.uiState.completedAddonAction).toEqual(
+      addonAddedAction,
+    );
+
+    // Trigger the setTimeout behavior.
+    clock.tick(MESSAGE_RESET_TIME);
+
+    applyUIStateChanges({ root, store });
+    expect(root.instance().props.uiState.completedAddonAction).toEqual(null);
```
rebmullin: do we need to test state here too? I guess I was under the impression that checking if it exists (the next line) is enough..?

src/amo/reducers/collections.js
```diff
@@ -1308,6 +1322,20 @@ const reducer = (
...
+      return {
+        ...state,
+        hasAddonBeenRemoved: false,
+      };
+    }
+
+    case ADDON_REMOVED_FROM_COLLECTION: {
+      return {
+        ...state,
+        hasAddonBeenRemoved: true,
```
rebmullin: I don't know if this matters but should this get reset back to false? I was looking at redux and saw after it was removed it just stayed in this true state? just trying to understand : )

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -105,31 +139,33 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
...
+                    ? i18n.gettext('Added to collection')
+                    : i18n.gettext('Removed from collection')}
```
rebmullin: when I remove an item in smaller screens (where the search sits below), for some reason it scrolls back to the top of the screen after the message fades...the 'Added to collection' does not do this. Do you see this too?

src/amo/components/CollectionAddAddon/styles.scss
```diff
@@ -1,8 +1,23 @@
...
+  margin: $padding-page 0;
+
+  @include respond-to(large) {
+    margin-bottom: $padding-page-l;
+    margin-top: 0;
+  }
+
+  .Card-contents {
+    overflow: visible;
+    padding-top: calc(20px - 6px);
```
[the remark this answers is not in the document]
bobsilverberg: It's the margin from the notice. I added a comment.

tests/unit/amo/components/TestCollectionAddAddon.js
bobsilverberg: Yes, I suppose what I'm testing here is just the internal implementation, although, because `uiState.completedAddonAction` has three potential values, I kind of like checking that it's been set to what I expect here, in addition to checking that the notice has disappeared.

src/amo/reducers/collections.js
bobsilverberg: I'm not sure about this. I can see what you mean - it does seem like it should get reset at some point, but I'm not sure at what point that should be. Would it be after the notice fades out? Having the code as it is doesn't seem to cause any problems, so I think it's okay, but let's see if @kumar303 has any concerns about this as well.

tests/unit/amo/components/TestCollectionAddAddon.js
rebmullin: ahh..ok that makes sense

src/amo/components/CollectionAddAddon/index.js
rebmullin: sure thing!

src/amo/reducers/collections.js
kumar303: Actually, this crossed my mind when reviewing too but I decided that it's not necessary for the current behavior and will only add more complexity to the reducer and the tests. I don't think it's worth it.

src/amo/components/CollectionAddAddon/styles.scss
```diff
@@ -1,8 +1,24 @@
...
+
+  @include respond-to(large) {
+    margin-bottom: $padding-page-l;
+    margin-top: 0;
+  }
+
+  .Card-contents {
+    overflow: visible;
+    // 6px is the margin from the CollectionAddAddon-noticePlaceholder.
+    padding-top: calc(20px - 6px);
```
rebmullin: do we need calc here? can this just be `padding-top: 14px`

src/amo/components/CollectionAddAddon/index.js
```diff
@@ -65,27 +70,51 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
...
-    const { hasAddonBeenAdded: hasAddonBeenAddedNew } = props;
-    const { hasAddonBeenAdded } = this.props;
-    if (hasAddonBeenAdded !== hasAddonBeenAddedNew) {
+    const {
+      hasAddonBeenAdded: hasAddonBeenAddedNew,
+      hasAddonBeenRemoved: hasAddonBeenRemovedNew,
+    } = props;
+    const { errorHandler, hasAddonBeenAdded, hasAddonBeenRemoved } = this.props;
+
+    const addStatusChanged = hasAddonBeenAdded !== hasAddonBeenAddedNew;
```
kumar303: The complexity of having to compare the previous value with the new value struck me as odd when I saw the first patch but it only just dawned on me as to what the problem is.

You're mixing two state managers. You're mixing `setUIState()` with the `collections` reducer. You should be able to accomplish all of this with just the `collections` reducer.

Rendering the notification should be as simple as checking the mapped state prop:
```diff
diff --git a/src/amo/components/CollectionAddAddon/index.js b/src/amo/components/CollectionAddAddon/index.js
index 317831e7e..b71bb903d 100644
--- a/src/amo/components/CollectionAddAddon/index.js
+++ b/src/amo/components/CollectionAddAddon/index.js
@@ -138,7 +138,13 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
   };
 
   render() {
-    const { errorHandler, i18n, uiState } = this.props;
+    const {
+      errorHandler,
+      hasAddonBeenAdded,
+      hasAddonBeenRemoved,
+      i18n,
+      uiState,
+    } = this.props;
     const { addonAction } = uiState;
     const addonAdded = addonAction === addonAddedAction;
 
@@ -148,13 +154,13 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
           errorHandler.renderError()
         ) : (
           <TransitionGroup className="CollectionAddAddon-noticePlaceholder">
-            {addonAction && (
+            {(hasAddonBeenAdded || hasAddonBeenRemoved) && (
               <CSSTransition
                 classNames="CollectionAddAddon-noticePlaceholder-transition"
                 timeout={MESSAGE_FADEOUT_TIME}
               >
-                <Notice type={addonAdded ? 'success' : 'generic'}>
-                  {addonAdded
+                <Notice type={hasAddonBeenAdded ? 'success' : 'generic'}>
+                  {hasAddonBeenAdded
                     ? i18n.gettext('Added to collection')
                     : i18n.gettext('Removed from collection')}
                 </Notice>
```

However, this means you also need to hide the notification by dispatching an action in a timeout. Something like this:

```diff
diff --git a/src/amo/components/CollectionAddAddon/index.js b/src/amo/components/CollectionAddAddon/index.js
index 317831e7e..0b0e2e226 100644
--- a/src/amo/components/CollectionAddAddon/index.js
+++ b/src/amo/components/CollectionAddAddon/index.js
@@ -71,11 +71,19 @@ export class CollectionAddAddonBase extends React.Component<InternalProps> {
 
   componentWillReceiveProps(props: InternalProps) {
     const {
+      dispatch,
       hasAddonBeenAdded: hasAddonBeenAddedNew,
       hasAddonBeenRemoved: hasAddonBeenRemovedNew,
+      setTimeout,
     } = props;
     const { errorHandler, hasAddonBeenAdded, hasAddonBeenRemoved } = this.props;
 
+    if (hasAddonBeenAdded) {
+      setTimeout(() => {
+        dispatch(hideAddonAddedNotification());
+      }, MESSAGE_FADEOUT_TIME);
+    }
+
     const addStatusChanged = hasAddonBeenAdded !== hasAddonBeenAddedNew;
     const removeStatusChanged = hasAddonBeenRemoved !== hasAddonBeenRemovedNew;
 
```

Both of these seem like simple changes but it would take quite a bit of redesign so I don't suggest you try and do it in this patch. You could do it in a future patch if you like but it would of course take time and effort. The net benefit is probably just that `componentWillReceiveProps()` would become simpler.

Anyway, just food for thought!

tests/unit/amo/sagas/test_collections.js
```diff
@@ -732,6 +733,31 @@ describe(__filename, () => {
...
+        slug: 'some-other-slug',
+        username: 'some-other-user',
+      };
+
+      mockApi
+        .expects('removeAddonFromCollection')
+        .once()
+        .returns(Promise.resolve());
+
+      _removeAddonFromCollection(params);
```
kumar303: This can just be `_removeAddonFromCollection()` since none of the parameters are important for testing the behavior of dispatching the removal action.

src/amo/reducers/collections.js
kumar303: > I think it's a little awkward to track ephemeral data such as notifications in Redux.

Actually, duh, it isn't awkward at all. You just dispatch one action to show the notification and another action from `setTimeout` to hide it. 

I left a detailed review comment!

src/amo/components/CollectionAddAddon/styles.scss
kumar303: I think it's kinda nice to see see that 20px is the starting point. It might be helpful when we finally fix all 20px margins 😛 They don't follow the pixel spec.

bobsilverberg: Yeah, I did that to be explicit about how the 14px is calculated.

src/amo/components/CollectionAddAddon/index.js
bobsilverberg: Thanks @kumar303, you're right, that's much better and much simpler. Because of that, I'm actually inclined to fix it now even though it's a bit of work. Otherwise the "worse" approach I'm using will just become more technical debt and we have a big enough backlog of that already. I don't think it will be a *huge* amount of work, so I'll try to get it done in time for the next tag.

PR #5954, opened 2018-08-15
Add an underline to the links to "all reviews" from the AddonMeta component

### LINKED ISSUE #5952: Add an underline to the links to "all reviews" from the AddonMeta component
The number of reviews and the title that reads "Reviews" was turned into a link in the `AddonMeta` component, with the number being displayed in black text and the title in grey text. To make these more obviously links, we should also display an underline for them. Both of them should become blue in colour when hovered over.

bobsilverberg:
Fixes #5952 

Before:

screenshot 2018-08-15 15 10 05

Note that hovering over the number "153" in the *before* version does not change the color of the number to blue.

After:

screenshot 2018-08-15 15 12 29

After, while hovering over the number of reviews:

screenshot 2018-08-15 15 08 19

Note also that hovering over the text "Reviews" changes the color to blue in both the before and after versions.

### DIFF
--- a/src/amo/components/AddonMeta/styles.scss
+++ b/src/amo/components/AddonMeta/styles.scss
@@ -51,7 +51,6 @@
   a:visited {
     color: $grey-50;
     font-weight: normal;
-    text-decoration: none;
   }
 
   a.AddonMeta-reviews-content-link {
@@ -59,6 +58,11 @@
     &:link,
     &:visited {
       color: $black;
+    }
+
+    &:active,
+    &:hover {
+      color: $link-color;
     }
   }
 


src/amo/components/AddonMeta/styles.scss
```diff
@@ -60,6 +59,11 @@
     &:visited {
       color: $black;
     }
+
+    &:active,
```
bobsilverberg: I'm not sure why I have to repeat these here, and they're not simply inherited from the definition for `a:active, a:hover` below, but without this explicit override the link remains black when being hovered over.

rebmullin: I don't thin the below styles do anything..?

it looks like the style is basically this:

```css
a.AddonMeta-reviews-content-link a.active {
 // styles
}
```

it's styling for nested links so I that shouldn't work b/c that shouldn't be valid html.. if that make sense. I think you can probably remove the below styles

kumar303: > I'm not sure why I have to repeat these here...

I think it's because [the whole anchor](https://github.com/mozilla/addons-frontend/blob/c3e8e3a73bca79e1d2f8dfbb14c80b4a41853d25/src/amo/components/AddonMeta/styles.scss#L57) is set to black and due to the specificity, the rule below it does not override it. I think you could remove the line I linked to and then you won't have to repeat [this rule](https://github.com/mozilla/addons-frontend/blob/c3e8e3a73bca79e1d2f8dfbb14c80b4a41853d25/src/amo/components/AddonMeta/styles.scss#L65).

An ampersand in Sass expands to the compiled selector after applying all nested brackets.

bobsilverberg: Removing the rule for [`&`](https://github.com/mozilla/addons-frontend/blob/c3e8e3a73bca79e1d2f8dfbb14c80b4a41853d25/src/amo/components/AddonMeta/styles.scss#L57) doesn't work if I also remove the [repeated rule](https://github.com/mozilla/addons-frontend/blob/c3e8e3a73bca79e1d2f8dfbb14c80b4a41853d25/src/amo/components/AddonMeta/styles.scss#L65). The link remains black when hovered.

I also tried replacing the rule with the ampersands with

```
  a:link.AddonMeta-reviews-content-link,
  a:visited.AddonMeta-reviews-content-link {
    color: $black;
  }
```

to avoid overriding the whole anchor, but that also doesn't help and the link remains black when hovered over.

I cannot seem to get it to turn blue unless I do the repeated declaration as exists in the current patch. Does anyone have any other suggestions to try?

rebmullin: oops/sorry..disregard my above comment. I misread this and thought this was nested...

rebmullin: I don't really see a way around it b/c you need override the specificity

kumar303:
> I'm not sure why, though.

Maybe it's because even though you're hovering over it, the element is still a `:link` (i.e. a non-visited anchor with an href).

Issue #5945, opened 2018-08-14
Saga for creating / updating an add-on review or rating


kumar303:
In order to create a new review manager (see https://github.com/mozilla/addons-frontend/issues/5942) we need a saga for creating / updating an add-on review or rating.

Issue #5942, opened 2018-08-14
Write a review inside of a text-box, not in an overlay


kumar303:
This is part of a new UX for reviews/ratings. See #5519

- Go to an add-on detail page
- Sign in
- Select a star to rate an add-on

Instead of seeing an overlay, you should be able to write your review in a text box.

**QA**, please see the list of known issues for this patch at https://github.com/mozilla/addons-frontend/pull/5997

<img width="495" alt="screenshot 2018-08-14 16 33 26" src="https://user-images.githubusercontent.com/55398/44120134-f3a0d634-9fe0-11e8-8b63-eccd17375328.png">
<img width="487" alt="screenshot 2018-08-14 16 33 37" src="https://user-images.githubusercontent.com/55398/44120137-f659019e-9fe0-11e8-9a1d-fd2a5b202016.png">

AlexandraMoga:
Verified on AMO -dev with FF61 and Android 8.0

The current functionality is:
- rating is saved when a star is selected
- the review form is opened when a star is selected
- a different star rating can be selected and saved inside the review form
- a review can be written inside the the new textbox
- the review form can be dismissed by clicking on the "Nevermind, ...." button

@kumar303 I have two cases that haven't been specifically mentioned in the "known issues" list from #5997
I have logged #6120 and #6121 for more details

PR #5963, opened 2018-08-16
Saga for creating / updating an add-on review

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5945

This will support the UI in https://github.com/mozilla/addons-frontend/issues/5942 . Either this or the reducer will probably need tweaks for the UI but this patch should be a good foundation to build on.


src/amo/constants.js
```diff
@@ -1,31 +1,4 @@
 /* @flow */
-// Action types.
```
kumar303: Sorry for the noise. It was bothering me that these action types were not in the action module. This cleans up `amo/constants` quite a bit.

src/amo/actions/reviews.js
```diff
@@ -1,33 +1,41 @@
...
-  SHOW_REPLY_TO_REVIEW_FORM,
-} from 'amo/constants';
 import type { FlagReviewReasonType } from 'amo/constants';
 import type {
   ExternalReviewReplyType,
   ExternalReviewType,
   GroupedRatingsType,
 } from 'amo/api/reviews';
 
+export const CLEAR_ADDON_REVIEWS: 'CLEAR_ADDON_REVIEWS' = 'CLEAR_ADDON_REVIEWS';
```
bobsilverberg: This pattern seems nice, having the actions in `actions/reviews.js`, and only having "true" reducer code in the `reducers/reviews.js` file, but it's also inconsistent across the site. Is this an old pattern or a new pattern? I ask because it's not what we did with collections, and a lot of that is new code.

I'm not suggesting we go back and change a bunch of stuff just to be consistent, but it would be good to know, moving forward, what pattern we'd like to use, and it might be worth converting some things as we touch them, if it's not a huge amount of work.

kumar303: The new pattern is to put actions in the reducer module. The reviews code is pretty old but I didn't want to make the patch too crazy by moving it to the reducer. This is a good baby step, though.

src/amo/api/reviews.js
```diff
@@ -53,6 +53,8 @@ export type SubmitReviewParams = {|
   versionId?: number,
 |};
 
+export type SubmitReviewResponse = ExternalReviewType;
```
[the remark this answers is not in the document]
kumar303: I had to import it into the saga for type hinting (since Flow can't follow saga's cryptic API) so I thought it would be better to give it more specific alias.

kumar303:
> I was wondering why we didn't already have a saga for doing this, and then I saw that we were just dispatching API calls directly from components. 

Yes, don't do that for future components! Many review components were implemented pre-saga and they "work" but the test suites are overly complicated. The saga pattern is really good for testing because the component only has to dispatch an action, which is easy to test.

kumar303:
Thanks for the review.

PR #5969, opened 2018-08-17
showing a warning for non public add-0n

### LINKED ISSUE #4244: Show a warning when viewing a non-public add-on

As an admin, log in to AMO and view the detail page for a disabled add-on

*or*

As a developer, upload an add-on, make it non-public, log in to AMO and view its detail page

### What happened?

You see a complete detail page as if nothing was different

You should see a warning saying that the add-on is not public but you're seeing it anyway because you have elevated privileges.

This has created confusion a couple of times from admins or developers who are not expecting to see a public listing.

It should be pretty easy to fix:

```diff
diff --git a/src/amo/components/Addon/index.js b/src/amo/components/Addon/index.js
index a2d4eec21..52e2b4d4e 100644
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -53,6 +53,7 @@ import Card from 'ui/components/Card';
 import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
 import ShowMoreCard from 'ui/components/ShowMoreCard';
+import Notice from 'ui/components/Notice';
 
 import './styles.scss';
 
@@ -537,6 +538,11 @@ export class AddonBase extends React.Component {
                 reason={compatibility.reason}
               />
             ) : null}
+            {addon && addon.status !== 'public' ? (
+              <Notice type="error">
+                {i18n.gettext('This is not a public listing. You are only seeing it because of elevated permissions.')}
+              </Notice>
+            ) : null}
             <header className="Addon-header">
               {this.headerImage({ compatible: isCompatible })}
 
```

championshuttler:
Fixes #4244


src/core/types/addons.js
```diff
@@ -1,9 +1,11 @@
 /* @flow */
 import type { AddonTypeType } from 'core/constants';
 
+export const STATUS_PUBLIC: 'public' = 'public';
```
rebmullin: I personally think the constant here should be moved to core/constants

willdurand: I second @rebmullin's comment

willdurand: @rebmullin is right, but I am not even sure we really need a constant outside the component.

championshuttler: So do I need to move this to core/constants or can we use 'public' directly?

willdurand: You can use a constant but only in the component file.

rebmullin: also, I think this could just be `export const STATUS_PUBLIC = 'public';`

championshuttler: mine dumbness++ :man_facepalming:

rebmullin: hey @championshuttler, sorry for the confusion but I think what @willdurand suggests is better... to put this constant in the component file itself. If this ends up being a constant we need over and over we can move it back the core/constants file

rebmullin:
hi @championshuttler, just making sure you saw the feedback on the questions above:

https://github.com/mozilla/addons-frontend/issues/4244#issuecomment-414024659

championshuttler:
Hey @willdurand , can you please review it, I and @rebmullin tried another way also to write test case but it's not working.

We wrote test case firstly for verification of admin/developer and  `error` will appear to developer/admins only. 
```
  it('displays a notice on a disabled addon to [admin/developer]', () => {
      userProps: {
        is_addon_developer: true,
        display_name: 'imadev',
        userId: 100,
        username: 'iamdev',
      },
    });

    const addon = createInternalAddon({
      ...fakeAddon,
      status: 'disabled',
    });

    const root = shallowRender({ addon, store });
  });
  });
```

But this test case is working in both case like when we set true for `is_addon_developer` but its also passing when its `false` :astonished:

championshuttler:
Added 10 px margin .
Before
https://camo.githubusercontent.com/e2ab12e980ba57ffbd9533b6ba3d228c342cbd1e/68747470733a2f2f6d6f6e6f736e61702e636f6d2f696d6167652f4955485071356f4e32314a5430696634324c4a7258724f303377787a34572e706e67
image

After: 
screenshot_2018-08-22 idkhtml add-ons for firefox 2

On Mobile 
Before 
movile

After 
screenshot_2018-08-22 idkhtml add-ons for firefox 1

src/ui/components/Notice/styles.scss
```diff
@@ -11,6 +11,7 @@
   grid-gap: 4px;
   grid-template-columns: min-content auto;
   line-height: 1.1;
+  margin-bottom: 10px;
```
willdurand: This will probably add a `margin` to all `Notice` components, which might not be a good idea.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1566,4 +1568,42 @@ describe(__filename, () => {
       expect(installButton).toHaveProp('uninstall', uninstall);
     });
   });
+
+  it('displays a notice to the admin/developer on a disabled addon', () => {
```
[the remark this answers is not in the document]
willdurand: There should be one test case for an admin and one test case for the developer of the add-on.

`is_addon_developer` does not control anything because the API performs the permissions check. We should probably leave a comment mentioning that: "non-public add-ons require an account listed as a developer of the add-on or admin rights".

src/ui/components/Notice/styles.scss
championshuttler: Yes, you are correct, it will add margin to every Notice component.
 Should I define inline css for it?

championshuttler:
Hi @willdurand i added test case for developer can you please review :see_no_evil:

src/ui/components/Notice/styles.scss
championshuttler: I think here https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/Addon/styles.scss#L67 ;)

willdurand: Then, I'll add a class name to the `Notice` you've added and change the SCSS like this:

```scss
.AddonCompatibilityError,
.YourNewClassName {
  ...
}
```

championshuttler: Got it :dancer:

championshuttler: Fixed  :man_dancing: 

After changes
screenshot_2018-08-24 idkhtml add-ons for firefox 1
screenshot_2018-08-24 idkhtml add-ons for firefox

src/amo/components/Addon/index.js
```diff
@@ -500,7 +500,7 @@ export class AddonBase extends React.Component {
               />
             ) : null}
             {addon && addon.status !== STATUS_PUBLIC ? (
-              <Notice type="error">
+              <Notice type="error" className="Non-public-addon-notice">
```
willdurand: I think it should start with `Addon`, like `Addon-non-public-notice`.

championshuttler: :man_facepalming: Done :)

rebmullin:
hey @championshuttler, travis is reporting an error. I think if you rebase, it should get fixed..

tests/unit/amo/components/TestAddon.js
```diff
@@ -1566,4 +1568,71 @@ describe(__filename, () => {
       expect(installButton).toHaveProp('uninstall', uninstall);
     });
   });
+  // non-public add-ons require an account listed as a developer of the add-on or admin rights
```
rebmullin: hey @championshuttler, can you capitalize the first letter Non-public.. and end with a period?

championshuttler: :+1:

tests/unit/amo/components/TestAddon.js
```diff
@@ -1566,4 +1568,71 @@ describe(__filename, () => {
...
+    const addon = createInternalAddon({
+      ...fakeAddon,
+      status: 'disabled',
+    });
+
+    const root = shallowRender({
+      addon,
+      store: dispatchSignInActions({
+        userProps: {
+          permissions: [ALL_SUPER_POWERS],
```
rebmullin: so @willdurand this is a sufficient test? if we do `permission: []` here, the test still passes (same with the below case - if we say `is_addon_developer: false` for the dev case). I understand the API will return a 403/404 if they are not authorized/logged in but I guess I am still a little confused myself on the usefulness of these tests shown in this way b/c if you change the details in the way I noted above it still passes. Maybe we should add more comment about this not even being reached (or the test case that checks this)? If that makes sense..

But if you're ok with this, then I'll defer to you : )

willdurand: right, maybe a single test case would be enough.

championshuttler: @willdurand thanks for the quick reply , should i delete the test case for developer?

willdurand: yeah, I think @rebmullin is right. we should have test cases that fail if we change something, so I suppose we only need one test case that is the opposite of when api returns 403.

rebmullin: @willdurand, so there is already this test case:

```javascript
it('renders NotFound page for forbidden add-on - 403 error', () => {

    const error = createApiError({
      response: { status: 403 },
      jsonResponse: { message: 'You do not have permission.' },
    });
      id: 'some-id',
      dispatch: store.dispatch,
    });
    errorHandler.handle(error);

    const root = renderComponent({ errorHandler, store });
  });
```

is this sufficient? I mean it shows that a 403 gives Notfound page..

and maybe we can just add the following to the above test:
`expect(root.find(Notice)).toHaveLength(0);` 

would that work?

(on top of merging the above dev and admin tests as you suggest above)

rebmullin: still using the comment above the test case:
`// Non-public add-ons require an account listed as a developer of the add-on or admin rights.`

willdurand: We keep the comment and we add a test case indicating that we display a notice message when an admin or the developer views the page.

rebmullin:
looks 👍  to me. Thanks @championshuttler for tackling this one : )

PR #5974, opened 2018-08-17
Made homepage more responsive(720-760px width)

### LINKED ISSUE #4087: AMO homepage has layout issues for German on smaller screen sizes
STR:

1. Load AMO homepage in responsive design (~720 - 750 width)
2. Change your locale to Deutsch
3. Observe the layout of the Collections shelf (Customize the way Firefox works...)

Expected result:
There are no layout issues

Actual result;
There is little to no spacing between the Collection titles

Notes:
- reproduced on all AMO servers with FF57 Win10 and Android 7.0 (tablet).

image

hritvi:
Fixes #4087 

<b>After</b>
screenshot from 2018-08-17 18-35-16
screenshot from 2018-08-17 18-41-16

<b>Before</b>
https://user-images.githubusercontent.com/31961530/34110069-36ece7dc-e40e-11e7-9404-4518457dff13.png

### DIFF
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -144,8 +144,10 @@ export class HomeBase extends React.Component {
               to={`/collections/mozilla/${collectionSlug}/`}
               className="Home-SubjectShelf-link"
             >
-              <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
-              <span>{title}</span>
+              <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
+                <span class="listTitle-vw">{title}</span>
+              </Icon>
+              <span class="listTitle">{title}</span>
             </Link>
           </li>
         ))}
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -145,9 +145,9 @@ export class HomeBase extends React.Component {
               className="Home-SubjectShelf-link"
             >
               <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
-                <span class="Home-SubjectShelf-link-text-vw">{title}</span>
+                <span className="Home-SubjectShelf-link-text-vw">{title}</span>
               </Icon>
-              <span class="Home-SubjectShelf-link-text">{title}</span>
+              <span className="Home-SubjectShelf-link-text">{title}</span>
             </Link>
           </li>
         ))}
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -145,9 +145,9 @@ export class HomeBase extends React.Component {
               className="Home-SubjectShelf-link"
             >
               <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
-                <span class="listTitle-vw">{title}</span>
+                <span class="Home-SubjectShelf-link-text-vw">{title}</span>
               </Icon>
-              <span class="listTitle">{title}</span>
+              <span class="Home-SubjectShelf-link-text">{title}</span>
             </Link>
           </li>
         ))}
--- a/src/amo/components/Home/index.js
+++ b/src/amo/components/Home/index.js
@@ -144,10 +144,8 @@ export class HomeBase extends React.Component {
               to={`/collections/mozilla/${collectionSlug}/`}
               className="Home-SubjectShelf-link"
             >
-              <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
-                <span className="Home-SubjectShelf-link-text-vw">{title}</span>
-              </Icon>
-              <span className="Home-SubjectShelf-link-text">{title}</span>
+              <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
+              <span>{title}</span>
             </Link>
           </li>
         ))}
--- a/src/amo/components/Home/styles.scss
+++ b/src/amo/components/Home/styles.scss
@@ -20,7 +20,7 @@
   .Card-header {
     display: block;
 
-    @include respond-to(large) {
+    @media only screen and (min-width: 760px) {
       display: none;
     }
   }
@@ -33,7 +33,7 @@
     overflow: auto;
     width: 100%;
 
-    @include respond-to(large) {
+    @media only screen and (min-width: 760px){
       border-top-left-radius: $border-radius-default;
       border-top-right-radius: $border-radius-default;
       flex-wrap: nowrap;
@@ -47,7 +47,7 @@
   display: none;
   height: auto;
 
-  @include respond-to(large) {
+  @media only screen and (min-width: 760px) {
     align-self: center;
     display: block;
     max-width: 130px;
--- a/src/amo/components/Home/styles.scss
+++ b/src/amo/components/Home/styles.scss
@@ -80,15 +80,15 @@
 .Home-SubjectShelf-list {
   align-self: center;
   display: flex;
-  flex-flow: row wrap;
+  flex-flow: column wrap;
   justify-content: space-between;
   margin: 0 auto;
   overflow: auto;
   padding: 0;
   width: 100%;
 
-  @include respond-to(extraSmall) {
-    flex-flow: column wrap;
+  @include respond-to(medium) {
+    flex-flow: row wrap;
   }
 
   @include respond-to(large) {
@@ -137,11 +137,11 @@
   padding-left: 75px;
   margin-bottom: 5px;
   transform: scale(0.8);
-  width: 100%;
+  width: 250%;
   text-align: left;
 
-  @include respond-to(extraSmall) {
-    width: 250%;
+  @include respond-to(medium) {
+    width: 100%;
   }
 
   @include respond-to(large) {
--- a/src/amo/components/Home/styles.scss
+++ b/src/amo/components/Home/styles.scss
@@ -58,14 +58,14 @@
   }
 }
 
-.listTitle {
+.Home-SubjectShelf-link-text {
   display: none;
   @include respond-to(large) {
     display: block;
   }
 }
 
-.listTitle-vw {
+.Home-SubjectShelf-link-text-vw {
   display: inline;
   @include respond-to(large) {
     display: none;
--- a/src/amo/components/Home/styles.scss
+++ b/src/amo/components/Home/styles.scss
@@ -20,7 +20,7 @@
   .Card-header {
     display: block;
 
-    @media only screen and (min-width: 760px) {
+    @include respond-to(mediumcard) {
       display: none;
     }
   }
@@ -33,7 +33,7 @@
     overflow: auto;
     width: 100%;
 
-    @media only screen and (min-width: 760px){
+    @include respond-to(mediumcard) {
       border-top-left-radius: $border-radius-default;
       border-top-right-radius: $border-radius-default;
       flex-wrap: nowrap;
@@ -47,7 +47,7 @@
   display: none;
   height: auto;
 
-  @media only screen and (min-width: 760px) {
+  @include respond-to(mediumcard) {
     align-self: center;
     display: block;
     max-width: 130px;
@@ -55,6 +55,20 @@
 
   @include respond-to(extraLarge) {
     max-width: 250px;
+  }
+}
+
+.listTitle {
+  display: none;
+  @include respond-to(large) {
+    display: block;
+  }
+}
+
+.listTitle-vw {
+  display: inline;
+  @include respond-to(large) {
+    display: none;
   }
 }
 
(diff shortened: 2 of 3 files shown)


willdurand:
@hritvi how about implementing this proposal instead? https://github.com/mozilla/addons-frontend/issues/4087#issuecomment-370063825

hritvi:
Yes, we can also go with that proposal but I suppose the icon size has to be changed for that which won't be a good option.

hritvi:
And I can't understand why travis-ci tests are failing. Can you give me an insight into this @willdurand. As far as I can understand, is this because of syntax?

willdurand:
@hritvi Travis CI fails because of a syntax issue. You can solve it by running `yarn prettier-dev`, then committing the generated changes.

hritvi:
<code>yarn prettier-dev</code> returned with Everything is Awesome! And I would be glad to hear your viewpoints on the above mentioned proposal.

willdurand:
I don't think we should introduce custom media queries. We should deal with the existing breakpoints, so I'd like to consider a different fix like changing the breakpoint from `large` to `medium` maybe.

That being said, trying to comply with the proposal in https://github.com/mozilla/addons-frontend/issues/4087#issuecomment-370063825 would make even more sense.

hritvi:
Following that.

src/amo/components/Home/index.js
```diff
@@ -144,8 +144,10 @@ export class HomeBase extends React.Component {
               to={`/collections/mozilla/${collectionSlug}/`}
               className="Home-SubjectShelf-link"
             >
-              <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
-              <span>{title}</span>
+              <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
+                <span class="listTitle-vw">{title}</span>
+              </Icon>
+              <span class="listTitle">{title}</span>
```
willdurand: Do we really need to add new HTML here?

src/amo/css/inc/vars.scss
```diff
@@ -30,6 +30,10 @@ $breakpoints: (
     (
       min-width: 500px,
     ),
+  mediumcard: only screen and
```
willdurand: I don't think we should be adding new breakpoints. Can we use the existing ones?

src/amo/components/Home/styles.scss
```diff
@@ -58,6 +58,20 @@
   }
 }
 
+.listTitle {
```
willdurand: We tend to prefix the CSS class names by the component name. In your case, the CSS class is inside the `Home-SubjectShelf-link` block, so I'd reuse it. For example: `.Home-SubjectShelf-link-text`.

src/amo/css/inc/vars.scss
willdurand: Yep, looks good with `medium`.

hritvi: The issue dealt with increasing the readability of this view(German):
https://user-images.githubusercontent.com/31961530/34110069-36ece7dc-e40e-11e7-9404-4518457dff13.png
If I remove the breakpoint, I am afraid that there would be no other view for viewports between 720px and 760px.
I can add media queries for the same but that doesn't go well with the rest of the code flow.

willdurand: ah, then use `extraLarge`.

hritvi: done :+1:

src/amo/components/Home/index.js
```diff
@@ -144,8 +144,10 @@ export class HomeBase extends React.Component {
               to={`/collections/mozilla/${collectionSlug}/`}
               className="Home-SubjectShelf-link"
             >
-              <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
-              <span>{title}</span>
+              <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
+                <span class="Home-SubjectShelf-link-text-vw">{title}</span>
```
kumar303: In React JSX you'll need to use DOM attribute names, not normal HTML attributes. In this case, `class=` should be `className=`.

src/amo/components/Home/index.js
```diff
@@ -144,8 +144,10 @@ export class HomeBase extends React.Component {
...
+              </Icon>
+              <span class="Home-SubjectShelf-link-text">{title}</span>
```
kumar303: Same thing here: `className=`

src/amo/components/Home/styles.scss
```diff
@@ -73,6 +87,10 @@
   padding: 0;
   width: 100%;
 
+  @include respond-to(extraSmall) {
```
kumar303: We don't have a breakpoint named `extraSmall` defined in https://github.com/mozilla/addons-frontend/blob/851f05b08ece38bc475f9be42d348f06eb15fa7b/src/core/css/inc/vars.scss#L59

Do you still need to add this rule? It looks like things are working without it.

src/amo/components/Home/styles.scss
```diff
@@ -109,19 +127,26 @@
...
   height: 64px;
+  padding-top: 25px;
+  padding-left: 75px;
   margin-bottom: 5px;
   transform: scale(0.8);
   width: 100%;
+  text-align: left;
+
+  @include respond-to(extraSmall) {
+    width: 250%;
```
kumar303: Do you still need this rule? I don't think it was ever getting applied because `extraSmall` is not a valid breakpoint.

src/amo/components/Home/styles.scss
hritvi: Ya, this breakpoint is used to generate this screen for viewports of size less than 349px as suggested by @pwalm. 
https://user-images.githubusercontent.com/6797635/36923791-3948abc2-1e21-11e8-83c7-48fd86d5a92b.png

rebmullin: hi @hritvi : ) since we do mobile first approach you should be able to set the mobile styles in the initial rules and then you can override those styles for example at the medium (500px)/higher breakpoint (see already defined breakpoints [here](https://github.com/mozilla/addons-frontend/blob/851f05b08ece38bc475f9be42d348f06eb15fa7b/src/core/css/inc/vars.scss#L59))

```css 
.Home-SubjectShelf-list {
   flex-flow: column wrap; // mobile
   ...other styles

   @include respond-to(medium) {
      flex-flow: row wrap // overrides mobile
      ...override other mobile styles or set new ones for this bp here
    }
}
```

src/amo/components/Home/index.js
```diff
@@ -144,8 +144,10 @@ export class HomeBase extends React.Component {
               to={`/collections/mozilla/${collectionSlug}/`}
               className="Home-SubjectShelf-link"
             >
-              <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
-              <span>{title}</span>
+              <Icon name={`Home-SubjectShelf-${collectionSlug}`}>
```
rebmullin: hi @hritvi : ) I *think* we can maybe keep the same markup and still style this using flex / grid. Have you tried that? It's ok to change markup when needed but I am just wondering if this is really needed here..? 

Also, just curious - what does "vw" stand for in `Home-SubjectShelf-link-text-vw`?

hritvi: Ya, I tried that but could not get a satisfactory result @rebmullin. Any advice by you would be of great help.

'vw' stands for viewportwidth, the content in <code>Home-SubjectShelf-link-text-vw</code>will be displayed according to the viewport width, that's why the name. Here too, any other name suggested by you would help. Thanks

rebmullin: does something like this work?

html:
```html
...
 <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
 <span className="Home-SubjectShelf-text">{title}</span>
...
```

```css
.Home-SubjectShelf-list-item {
  flex-grow: 1;
  list-style-type: none;
  margin: 15px 0;
  text-align: center;
  width: 100%;

  @include respond-to(medium) {
    width: 40%;
  }

  @include respond-to(large) {
    margin: 0;
  }
}

.Home-SubjectShelf-link {
  font-size: $font-size-m-smaller;
  text-decoration: none;
  display: flex;
  align-items: center;

  @include respond-to(large) {
    display: inline-block;
  }

  &:link,
  &:visited {
    color: $black;
  }

  &:active,
  &:focus,
  &:hover {
    color: $link-color;
    text-decoration: underline;
  }
}

.Home-SubjectShelf-list-item .Icon {
  background-position: left;
  background-repeat: no-repeat;
  background-size: 64px;
  content: '';
  display: block;
  flex-grow: 1;
  height: 64px;
  margin-bottom: 5px;
  transform: scale(0.8);

  @include respond-to(medium) {
    background-position: center;
    transform: initial;
    width: auto;
  }

  @include respond-to(large) {
    width: 100%;
  }
}

.Home-SubjectShelf-text {
  // 70px accounts for approx width of Icon plus some cushion
  width: calc(100% - 70px);
  text-align: left;

  @include respond-to(medium) {
    // 100px accounts for approx width of Icon plus some cushion
    width: calc(100% - 100px);
  }

  @include respond-to(large) {
    width: auto;
  }
}
```
sorry I am not showing diff here but something like seems to work. Basically I am toggling widths and display at different breakpoints.

You might have to adjust it more to get it to match the comps better.

hritvi: Thanks for your help :smiley: 
I tried that but it doesn't seem to work as it should be.
screenshot from 2018-09-06 11-01-26

I agree with you on toggling widths but we also have to toggle height where the text would lie and to make it lie in the center, I suppose, we do have to add some extra HTML.

rebmullin: hmmm.. this is what I see on my end:
Alt text

can you see that you have flex and align-items styles added to the link class (with the html I have noted above):

```css
.Home-SubjectShelf-link {
...
  display: flex;
  align-items: center;
...
```

hritvi: It worked @rebmullin with some minor changes. Thanks :smile:

hritvi:
@willdurand @rebmullin I have made the changes. But I can't understand why the tests are failing. Can you give me an insight into this. Thanks :smile:

src/amo/components/Home/styles.scss
```diff
@@ -94,6 +117,12 @@
 .Home-SubjectShelf-link {
   font-size: $font-size-m-smaller;
   text-decoration: none;
+  display: flex;
+  align-items: center;
```
championshuttler: @hritvi , for fixing errors you need to arrange css properties in alphabetical order . You can check the error here https://travis-ci.org/mozilla/addons-frontend/jobs/425625754#L517-L536

hritvi:
Finaaly the tests have passed :smile: Thanks @championshuttler.

championshuttler:
@hritvi great :+1:

src/amo/components/Home/styles.scss
```diff
@@ -118,11 +149,17 @@
   height: 64px;
   margin-bottom: 5px;
   transform: scale(0.8);
-  width: 100%;
+  width: 100px;
 
   @include respond-to(medium) {
+    background-position: left;
```
rebmullin: since this is same as mobile, we can remove this one

src/amo/components/Home/styles.scss
```diff
@@ -58,6 +58,23 @@
...
+  @include respond-to(medium) {
+    width: calc(
+      100% - 100px
+    ); /* 100px accounts for approx width of Icon plus some cushion */
+  }
+
+  @include respond-to(large) {
+    width: auto;
+  }
+}
```
rebmullin: hi @hritvi : ) I am not sure you need any of this anymore if you're not using [flex-grow](https://github.com/mozilla/addons-frontend/pull/5974#discussion_r215327406) here. So I think you can remove all this, as it doesn't seem to do anything now. This means you can remove the class (Home-SubjectShelf-link-text) on the span above. 

Please let me know if you see something different..

rebmullin:
hey @hritvi, I think it's very close : )  We just need to clean up a few more styles and then I think it's good. Please let me know if you have any questions.

src/amo/components/Home/index.js
```diff
@@ -145,7 +145,7 @@ export class HomeBase extends React.Component {
               className="Home-SubjectShelf-link"
             >
               <Icon name={`Home-SubjectShelf-${collectionSlug}`} />
-              <span>{title}</span>
+              <span className="Home-SubjectShelf-link-text">{title}</span>
```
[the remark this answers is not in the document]
hritvi: Ya, we don't need it anymore. I am sorry I just forgot to remove that :sweat_smile:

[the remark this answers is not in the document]
rebmullin: no worries, thanks for updating :)

rebmullin:
@hritvi, this looks great :) Thanks for all your work here 💯

PR #5982, opened 2018-08-18
Fixes #5964 resolves scrolling on top on clicking remove button

### LINKED ISSUE #5964: Clicking on "Remove" (while editing a collection), brings the user to the top of the page on mobile

Go to ~ Mobile size
Find a collection to edit that already has addons.
Click on "Remove" 
You should see a "Removed from collection" message (might be easier to see on desktop b/c of the issue noted below)
 

### What happened?

User gets sent back to the top of the page

It would keep the same scroll position (similar to what happens when you add an addon) and you would see the "Removed from collection" message

ShivangiKakkar:
Fix #5964 
#### Description
 Instead of comparing the page number’s using `!=` we were comparing the types (`!==`) which was always resulting in a true condition thus reloading the page everytime. :D


ShivangiKakkar:
I am not sure how to remove the lint error due to using `!=`

willdurand:
We should fix this issue in a different manner, which is always considering query param values as string values, not numbers.

ShivangiKakkar:
@willdurand 
oh okay! Got your point.
Today I read about the evil twins `!=` and `==`:P

rebmullin:
hi @ShivangiKakkar, are you still looking into this one? Do you need any help?

ShivangiKakkar:
@rebmullin Hey, I was under the impression that @willdurand is working on it. This issue is free if anyone wants to work on it. :)

ShivangiKakkar:
@rebmullin Hey! I was under the impression that @willdurand is looking into it. But I can surely try again, can you please guide me a bit in the direction I need to think so as to resolve this? Thank you :)

willdurand:
> However, we still see this scroll to top happen when we change the route

I think this is OK, we have a routing middleware to actually do that if I remember correctly. I don't think we should spend too much time fixing this right now because it is expected to lose the scroll position following a page refresh.

ShivangiKakkar:
@willdurand @rebmullin Please review again. Thanks for the help :)

src/amo/components/Collection/index.js
```diff
@@ -267,7 +267,7 @@ export class CollectionBase extends React.Component<InternalProps> {
     const newCollectionPage = computeNewCollectionPage(collection);
 
     if (page !== newCollectionPage) {
-      page = newCollectionPage;
+      page = Number(newCollectionPage);
```
rebmullin: I think this is ok to leave as string type. If you look at the original type of the page var, you can see it's a string there.

rebmullin:
Hi @ShivangiKakkar, regarding the test failing here:

`tests/unit/amo/components/TestCollection.js`

It's a similar issue - type issue - that is coming up in the tests. The page type is defined as a number type (in the test) but it should now be a string (to match your changes in the code). If you update the few tests where this happens, it should then pass.

ShivangiKakkar:
@rebmullin @willdurand ,Would you please review it again? Sorry for taking this much time. Thanks :)

ShivangiKakkar:
@willdurand @rebmullin, On removing the last addon it was directing the `page` to `0` that doesn't exist, resulting in an error.
So I put the check on the `return` statement. It seems to work fine now. Please review the changes. Thanks!

ShivangiKakkar:
@willdurand @rebmullin, On removing the last addon it was directing the `page` to `0` that doesn't exist, resulting in an error.
So I put the check on the `return` statement. It seems to work fine now. Please review the changes. Thanks!

willdurand:
@ShivangiKakkar I think you need to rebase this branch (it should have no conflicts, but rebase is needed as some files have been renamed)

ShivangiKakkar:
@willdurand @rebmullin, I did something really terrible while rebasing it I guess, So to avoid conflicts all together I am opening an entirely new PR for the issue. Sorry :(

PR #5992, opened 2018-08-20
Move logic for Collection Details card into its own component

### LINKED ISSUE #5990: Move logic for Collection Details card into its own component
To support #5748, and further simplify the `Collection` component, move the logic that renders the details card (which can end up being either a `CollectionDetails` component or a `CollectionManager` component) into its own component.

bobsilverberg:
Fixes #5990

### DIFF
--- a/src/amo/components/CollectionDetailsCard/index.js
+++ b/src/amo/components/CollectionDetailsCard/index.js
@@ -1,0 +1,102 @@
+/* @flow */
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import CollectionDetails from 'amo/components/CollectionDetails';
+import CollectionManager from 'amo/components/CollectionManager';
+import { getCurrentUser, hasPermission } from 'amo/reducers/users';
+import {
+  FEATURED_THEMES_COLLECTION_EDIT,
+  FEATURED_THEMES_COLLECTION_SLUG,
+  MOZILLA_COLLECTIONS_EDIT,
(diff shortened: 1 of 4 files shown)


src/amo/components/CollectionDetailsCard/index.js
```diff
@@ -0,0 +1,102 @@
...
+  filters: CollectionFilters,
+|};
+
+type InternalProps = {|
+  ...Props,
+  hasEditPermission: boolean,
+  showEditButton: boolean,
+|};
+
+export class CollectionDetailsCardBase extends React.Component<InternalProps> {
```
kumar303: This looks like it could be a stateless component.

src/amo/components/CollectionDetailsCard/index.js
```diff
@@ -0,0 +1,102 @@
...
+  let hasEditPermission = false;
+  let showEditButton = false;
+
+  if (collection && currentUser) {
+    hasEditPermission =
+      isOwner ||
+      // User can edit mozilla collections, and it is a mozilla collection.
+      (collection.authorUsername === MOZILLA_COLLECTIONS_USERNAME &&
+        hasPermission(state, MOZILLA_COLLECTIONS_EDIT));
+    showEditButton =
```
kumar303: The "edit button" is out of context here so I think it would be nicer to call this something more like `hasMaintainerPermission`. You would still use it like `<CollectionDetails showEditButton={hasMaintainerPermission} />` because showing the edit button makes sense in that context.

PR #5996, opened 2018-08-21
Node 8 :sunglasses:

### LINKED ISSUE #5215: Upgrade to node 8 in production
Node 6 is now in maintenance LTS and since node 8 has been active LTS we should be OK to switch over to it in production.

* Update to node 8 on -dev and test everything is ok
* Update on stage and then production.

willdurand:
Fix #5215

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Never submit security-related bugs through a Github Issue or by email.
 
 ## Requirements
 
-- You need [Node](https://nodejs.org/) 6.x which is the current [LTS](https://github.com/nodejs/LTS) (long term support) release.
+- You need [Node](https://nodejs.org/) 8.x which is the current [LTS](https://github.com/nodejs/LTS) (long term support) release.
 - Install [yarn](https://yarnpkg.com/en/) to manage dependencies and run scripts.
 
 The easiest way to manage multiple node versions in development is to use [nvm](https://github.com/creationix/nvm).
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "private": true,
   "engines": {
-    "node": ">=6 <=8"
+    "node": "<=8.10"
   },
   "scripts": {
     "build": "npm run clean && better-npm-run build",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "private": true,
   "engines": {
-    "node": ">=6 <=8"
+    "node": "<=8.10"
```
kumar303: Did you mean `>=8.10` ?

README.md
```diff
@@ -19,7 +19,7 @@ Never submit security-related bugs through a Github Issue or by email.
 
 ## Requirements
 
-- You need [Node](https://nodejs.org/) 6.x which is the current [LTS](https://github.com/nodejs/LTS) (long term support) release.
+- You need [Node](https://nodejs.org/) 8.x which is the current [LTS](https://github.com/nodejs/LTS) (long term support) release.
```
kumar303: 💃

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "private": true,
   "engines": {
-    "node": ">=6 <=8"
+    "node": ">=8.10, <=10"
```
kumar303: Do we even need to restrict it to less than 10? I'm not sure why we were doing that before.

package.json
willdurand: heh, yeah >= 8.10 is actually better... but we don't use this config anyway (see: https://github.com/yarnpkg/yarn/issues/2800)

package.json
willdurand: I think the idea was to describe which node versions we want to support. This field is never used anyway as I commented above, so maybe we can remove it entirely?

package.json
```diff
@@ -4,9 +4,6 @@
   "description": "Universal front-end projects to complement addons-server.",
   "main": "index.js",
   "private": true,
-  "engines": {
-    "node": ">=6 <=8"
```
kumar303: Why did you remove it entirely? I think it would be nice (for contributors) to give them an error when they aren't on node 8. This should produce an error for them:

```
    "node": ">=8.10"
```

Issue #5943, opened 2018-08-14
View and edit a saved add-on review


kumar303:
This is part of a new UX for reviews/ratings. See #5519

- Go to an add-on detail page
- rate an add-on
- write a review (see https://github.com/mozilla/addons-frontend/issues/5942)

Your saved review should be visible. You should be able to edit it.

<img width="482" alt="screenshot 2018-08-14 16 34 02" src="https://user-images.githubusercontent.com/55398/44120426-f011f1e6-9fe1-11e8-9bee-0ecb4b8ffdd2.png">

kumar303:
> The screen does not match the mock; this is intentional

Let me expand on my rationale for this. The **current mock** (the image at the top of this issue) has these problems:

- When submitting a star rating (no text), it is not obvious how to edit the rating. One can simply change the selected star but there is no prompt for this so it may not be clear. My approach builds upon the current style for a saved review so that there is an *Edit my rating* button. It's important to be able to edit a rating because someone may leave a low star rating out of frustration and then have a change of heart later.
- There was no way way to delete a star rating. I added a *Delete my rating* button.
- Similarly, there was no way to delete a saved review. This is an important feature. I added a *Delete my review* button.
- Because the saved review builds upon the existing style of rendering a review with controls at the bottom, I thought it no longer made sense to show a standalone *Update or edit your review* button. There is an *Edit my review* button which is clear enough.
- The mock shows *posted by you, just now* at the bottom but putting this at the bottom introduces more inconsistency to the style of reviews that we already have. I think it is not so different to put this by line at the top which is what I did.
- The mock explicitly calls a user *you* in *posted by you, just now*. I don't think this is necessary because the entire saved review is rendered in a card with the title *Rate your experience*. I think it's obvious that you wrote it. Also, dropping *you* keeps things consistent with the current listing style.

As a reminder, the UX mocks for this feature were known to be incomplete. Rather than wait for completion, we needed to move forward. As I worked through it and was able to interact with the UI, I discovered many problems that were fixable.

AlexandraMoga:
Verified on AMO -dev with FF62, Win10x64 and Android 8.0

- this is the layout of the rate your experience card after a review and a developer reply are submitted:

![image](https://user-images.githubusercontent.com/31961530/45422805-05017300-b69a-11e8-8edb-5f0ec405e90a.png)

@kumar303  I have a few observations:
1. the need to refresh the page in order to see your review is permanent or will it disappear once #6162 is fixed?
[...]

Issue #5941, opened 2018-08-14
When you select a star, indicate that the add-on rating has been saved


kumar303:
This is part of a new UX for reviews/ratings. See https://github.com/mozilla/addons-frontend/issues/5519

- Go to an add-on detail page
- Sign in
- Select a star to rate an add-on

A notice should tell you that the star rating was saved.

<img width="491" alt="screenshot 2018-08-14 16 32 51" src="https://user-images.githubusercontent.com/55398/44120010-9422ab88-9fe0-11e8-91f4-7b8ef9a141f2.png">

suhailsinghbains:
Hi, would love to work on this bug.
Are you talking about the showing a Notice from Notice component.

kumar303:
Thanks for offering to help! This is not a good issue to work on because it's already assigned to a developer. The best way to find issues to work on is to check this list https://github.com/mozilla/addons-frontend/labels/contrib%3A%20welcome These are the issues we've labeled as `contrib: welcome` because they are ready to work on.

AlexandraMoga:
Verified fixed on AMO -dev with FF62, Win10x64 and Android 8.0

![saved rating](https://user-images.githubusercontent.com/31961530/45625846-8b50f700-ba96-11e8-9011-3e91250d13b7.gif)

Issue #5519, opened 2018-07-06
Improve the experience of rating and writing a review


kumar303:
This is a placeholder issue for UX changes to improve the experience of rating an add-on, writing a review, and seeking support.

UX specs are in progress. Check the [PRD](https://docs.google.com/document/d/1oUz9Ng9v159IPdW4UqHhbNspWa9m5SaicAAX5jU7OjA/edit#) for details. Ideas:

- When you select a star, make it obvious that the rating has been saved (https://github.com/mozilla/addons-frontend/issues/5941)
- Show a generic button for reporting any kind of issue with an add-on, not just abuse. This will open a new flow for things like seeking support from the developer or reporting a bug. (https://github.com/mozilla/addons-frontend/issues/6199)
- Write a review inside of a text-box, not in an overlay (https://github.com/mozilla/addons-frontend/issues/5942, https://github.com/mozilla/addons-frontend/issues/5944)
- Make it obvious how to edit your written review (https://github.com/mozilla/addons-frontend/issues/5943)

muffinresearch:
@kumar303 do you need a tracker issue for this? Now that we have projects we're typically using them to replace tracker issues.

kumar303:
@muffinresearch I guess I meant a "tracker" (in quotes). We don't have anything filed currently so I wanted to put something in the project. I know this will be big but once we have UX specs we can split this apart and file separate issues.

pwalm:
OK, revised the flow:
<img width="1729" alt="screen shot 2018-07-23 at 3 50 20 pm" src="https://user-images.githubusercontent.com/6797635/43107085-47443c02-8e90-11e8-8648-494e8d4c379b.png">

kumar303:
@pwalm this looks great, thanks! Some feedback...

> Roll over stars to select rating

Could we say *select a star* or something to indicate that they need to click/tap on a star? Or would that be obvious?

In the read-only view of the written review, there is a pipe character separating the posted date and *Edit review*. We don't use pipes elsewhere and this strays from the current read-only view of written reviews. Could we make them consistent?

kumar303:
@jvillalobos :

What should we do for the *Contact the developer for support* button when the developer has not specified a support URL? Just hide the button? It would be helpful if we could email the developer the written message but I know that's a big feature.

[...]

PR #5997, opened 2018-08-21
Write a review inside of a text-box, not in an overlay 

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5942

There are some **known issues** with this patch:
- There is no indication that the initial star rating is saved. This will be addressed in https://github.com/mozilla/addons-frontend/issues/5941
- After submitting a review, the page does not render a new state even though the review was saved. The next state to render will be handled in https://github.com/mozilla/addons-frontend/issues/5943
- This patch is an incremental approach to solving https://github.com/mozilla/addons-frontend/issues/5519 . It is hidden behind a config flag.
- The text box does not show a list of tips as seen in the UX mock. This is because it's not possible to put new lines in a `textarea` placeholder.
- Unlike the UX mock, the *report abuse* button is always shown. This was intentional in an effort dissuade users from submitting one-star reviews when they really mean to report abuse.

<details>
<summary>Watch a screencast</summary>

addons-frontend-inline-review mov

</details>

### DIFF
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -54,7 +54,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
   };
 
   render() {
-    const { i18n, review } = this.props;
+    const { errorHandler, i18n, review } = this.props;
 
     const reviewGuide = i18n.sprintf(
       i18n.gettext(
@@ -76,10 +76,9 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
       'Write about your experience with this add-on.',
     );
 
-    // TODO: render errors
-
     return (
       <div className="AddonReviewManager">
+        {errorHandler.renderErrorIfPresent()}
         <div className="AddonReviewManager-starRating">
           {i18n.gettext('Your star rating:')}
           <Rating
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -64,7 +64,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
   render() {
     const { errorHandler, i18n, review, flashMessage } = this.props;
 
-    const reviewGuide = i18n.sprintf(
+    const reviewGuideLink = i18n.sprintf(
       i18n.gettext(
         'Please follow our %(linkStart)sreview guidelines%(linkEnd)s.',
       ),
@@ -76,7 +76,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
 
     /* eslint-disable react/no-danger */
     const formFooter = (
-      <span dangerouslySetInnerHTML={sanitizeHTML(reviewGuide, ['a'])} />
+      <span dangerouslySetInnerHTML={sanitizeHTML(reviewGuideLink, ['a'])} />
     );
     /* eslint-enable react/no-danger */
 
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -80,7 +80,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
       <div className="AddonReviewManager">
         {errorHandler.renderErrorIfPresent()}
         <div className="AddonReviewManager-starRating">
-          {i18n.gettext('Your star rating:')}
+          <span>{i18n.gettext('Your star rating:')}</span>
           <Rating
             className="AddonReviewManager-Rating"
             onSelectRating={this.onSubmitRating}
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -3,7 +3,12 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
-import { updateAddonReview } from 'amo/actions/reviews';
+import {
+  SAVED_RATING,
+  STARTED_SAVE_RATING,
+  STARTED_SAVE_REVIEW,
+  updateAddonReview,
+} from 'amo/actions/reviews';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
@@ -11,7 +16,7 @@ import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import Rating from 'ui/components/Rating';
 import type { DispatchFunc } from 'core/types/redux';
 import type { ErrorHandlerType } from 'core/errorHandler';
-import type { UserReviewType } from 'amo/actions/reviews';
+import type { FlashMessageType, UserReviewType } from 'amo/actions/reviews';
 import type { I18nType } from 'core/types/i18n';
 import type { OnSubmitParams } from 'ui/components/DismissibleTextForm';
 
@@ -26,6 +31,7 @@ type InternalProps = {|
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  flashMessage?: FlashMessageType | void,
 |};
 
 export class AddonReviewManagerBase extends React.Component<InternalProps> {
@@ -54,7 +60,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
   };
 
   render() {
-    const { errorHandler, i18n, review } = this.props;
+    const { errorHandler, i18n, review, flashMessage } = this.props;
 
     const reviewGuide = i18n.sprintf(
       i18n.gettext(
@@ -88,9 +94,21 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
             styleSize="small"
             yellowStars
           />
+          <span
+            className={makeClassName('AddonReviewManager-savedRating', {
+              'AddonReviewManager-savedRating-hidden':
+                flashMessage !== STARTED_SAVE_RATING &&
+                flashMessage !== SAVED_RATING,
+            })}
+          >
+            {flashMessage === STARTED_SAVE_RATING
+              ? i18n.gettext('Saving')
+              : i18n.gettext('Saved')}
+          </span>
         </div>
         <DismissibleTextForm
           formFooter={formFooter}
+          isSubmitting={flashMessage === STARTED_SAVE_REVIEW}
           onSubmit={this.onSubmitReview}
           placeholder={placeholder}
           puffyButtons
@@ -102,6 +120,12 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     );
   }
 }
+
+const mapStateToProps = (state: AppState) => {
+  return {
+    flashMessage: state.reviews.flashMessage,
+  };
+};
 
 export const extractId = (props: Props): string => {
   return props.review.id.toString();
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -92,7 +92,9 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
           <Rating
             className="AddonReviewManager-Rating"
             onSelectRating={this.onSubmitRating}
-            rating={review.rating}
+            rating={
+              flashMessage === STARTED_SAVE_RATING ? undefined : review.rating
+            }
             styleSize="small"
             yellowStars
           />
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import makeClassName from 'classnames';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -14,6 +15,7 @@ import translate from 'core/i18n/translate';
 import { sanitizeHTML } from 'core/utils';
 import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import Rating from 'ui/components/Rating';
+import type { AppState } from 'amo/store';
 import type { DispatchFunc } from 'core/types/redux';
 import type { ErrorHandlerType } from 'core/errorHandler';
 import type { FlashMessageType, UserReviewType } from 'amo/actions/reviews';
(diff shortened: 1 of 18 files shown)


src/amo/components/AddonReviewManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+  render() {
+    const { errorHandler, i18n, review, flashMessage } = this.props;
+
+    const reviewGuide = i18n.sprintf(
+      i18n.gettext(
+        'Please follow our %(linkStart)sreview guidelines%(linkEnd)s.',
+      ),
+      {
+        linkStart: '<a href="/review_guide">',
```
bobsilverberg: Should this open in a new tab?

src/ui/components/Rating/index.js
```diff
@@ -140,7 +140,7 @@ export class RatingBase extends React.Component<InternalProps, StateType> {
         key: `rating-${thisRating}`,
         onClick: undefined,
         onMouseEnter: () => this.onHoverStar(thisRating),
-        title: this.renderTitle(rating, readOnly, thisRating),
+        title: this.renderTitle(rating, readOnly || false, thisRating),
```
bobsilverberg: Why not just define `readOnly` in `renderTitle` as `readOnly?`? This `|| false` looks odd to me.

kumar303:
> The mocks suggest you can assign a rating by clicking a star, and not be forced/prompted to also enter a review.

That's coming in a future patch: https://github.com/mozilla/addons-frontend/issues/5941

src/amo/components/AddonReviewManager/index.js
kumar303: I guess it would be nice but I can't figure out how to do it. When I add `target=_blank` it gets stripped out by `sanitizeHTML()`.

src/ui/components/Rating/index.js
kumar303: I didn't want it to be `readOnly?` in `renderTitle` because that function relies on three positional arguments and I wanted Flow to catch accidental omissions. I think I can solve this a different way, though.

src/amo/components/AddonReviewManager/index.js
willdurand: If you add both target and rel, does that get stripped out too?

kumar303: I just tried `rel="noopener noreferrer"` and target is still stripped.

Upon digging, it looks like we could do it with a [hook](https://github.com/cure53/DOMPurify/blob/master/demos/hooks-target-blank-demo.html#L31) but this might be a bit much to fit into the patch. I see that Discopane does something similar but that changes all links.

willdurand: > I just tried rel="noopener noreferrer" and target is still stripped.

😞 

> Upon digging, it looks like we could do it with a hook but this might be a bit much to fit into the patch. I see that Discopane does something similar but that changes all links.

Yeah I remember having done that in https://github.com/mozilla/addons-frontend/issues/3554. Before, we had target/rel attributes on links everywhere.

We could maybe allow `target` and `rel` (`ALLOWED_ATTR`) or write a different hook.

kumar303: Allowing any kind of target value seems risky so I'd rather not do that.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -450,6 +454,80 @@ describe(__filename, () => {
...
+      expect(manager).toHaveLength(1);
+      expect(manager).toHaveProp('review', root.instance().props.userReview);
+
+      expect(root.find(UserRating)).toHaveLength(0);
+    });
+
+    it('hides AddonReviewManager when pressing the cancel button', async () => {
+      const root = renderInline({ store: createStoreWithLatestReview() });
+      root.setState({ showTextEntry: true });
+
```
bobsilverberg: Just wondering, why don't we have to call `root.update()` here, whereas we had to in the [previous test](https://github.com/mozilla/addons-frontend/blob/d36773d58b782bd001b53d9cd75c44695d50e5df/tests/unit/amo/components/TestRatingManager.js#L481)?

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -450,6 +454,80 @@ describe(__filename, () => {
...
+      root
+        .find('.RatingManager-cancelTextEntryButton')
+        .simulate('click', createFakeEvent());
+
```
bobsilverberg: `cancelTextEntry` also calls `event.preventDefault()` and we usually verify that in a test.

src/amo/components/RatingManager/index.js
```diff
@@ -182,40 +190,91 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+    event.preventDefault();
+    this.setState({ showTextEntry: false });
+  };
 
-    invariant(addon, 'addon is required');
-    invariant(version, 'version is required');
+  renderTextEntry() {
+    const { _config, userReview } = this.props;
+
+    if (!userReview || !this.isSignedIn()) {
```
bobsilverberg: There's a test for not rendering the `AddonReviewManager` if the user isn't signed in, but not for the case where `userReview` is falsey.

src/amo/components/RatingManager/index.js
```diff
@@ -182,40 +190,91 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+        </fieldset>
+      </form>
+    );
+  }
+
+  renderInlineReviewControls() {
+    const { i18n } = this.props;
+    const { showTextEntry } = this.state;
+
+    return showTextEntry && this.isSignedIn() ? (
```
bobsilverberg: Should there be a test for the case when `showTextEntry` is true but the user is not signed in, or do we consider that to be not possible?

tests/unit/amo/components/TestRatingManager.js
kumar303: Because `root.setState()` does a re-render internally. In the previous test, the state was changed asynchronously in an event handler which is why `root.update()` was needed. This is actually bad practice because the component could be unmounted which would throw an error (there is a bug about this, too). When we move that component to sagas it can go away.

tests/unit/amo/components/TestAddonReviewManager.js
```diff
@@ -0,0 +1,218 @@
...
+    const form = root.find(DismissibleTextForm);
+    expect(form).toHaveProp('text', body);
+  });
+
+  it('renders a DismissibleTextForm placeholder', () => {
+    const root = render();
+
+    const form = root.find(DismissibleTextForm);
+    expect(form).toHaveProp('placeholder');
+    expect(form.prop('placeholder')).toContain('Write about');
```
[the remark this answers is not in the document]
kumar303: Ah, true, good catch.

tests/unit/amo/components/TestAddonReviewManager.js
```diff
@@ -0,0 +1,218 @@
...
+    expect(message).toHaveText('Saved');
+    expect(message).not.toHaveClassName(
+      '.AddonReviewManager-savedRating-hidden',
+    );
+  });
+
+  it('hides a flashed rating message', () => {
+    const { store } = dispatchClientMetadata();
+    // Set a message then hide it.
+    store.dispatch(flashReviewMessage(SAVED_RATING));
```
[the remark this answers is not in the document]
kumar303: This test covers the negative part of the if statement. Both of the equality checks are already covered by [this](https://github.com/mozilla/addons-frontend/pull/5997/files#diff-8b7f80610063d6c88bad7762912e34e2R150) and [this](https://github.com/mozilla/addons-frontend/pull/5997/files#diff-8b7f80610063d6c88bad7762912e34e2R163).

tests/unit/amo/components/TestAddonReviewManager.js
```diff
@@ -0,0 +1,218 @@
...
+    expect(message).toHaveClassName('.AddonReviewManager-savedRating-hidden');
+  });
+
+  it('enters a submitting review state', () => {
+    const { store } = dispatchClientMetadata();
+    store.dispatch(flashReviewMessage(STARTED_SAVE_REVIEW));
+
+    const root = render({ store });
+
+    expect(root.find(DismissibleTextForm)).toHaveProp('isSubmitting', true);
```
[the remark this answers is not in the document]
kumar303: Good call

src/amo/components/RatingManager/index.js
kumar303: Ah, that's true. That state actually wasn't rendering correctly either. I hope to revisit these render states when I can get rid of the old behavior in RatingManager. It should be controlled more from Redux.

src/amo/components/RatingManager/index.js
kumar303: Actually, it is possible. You can begin editing your review and then--without saving it--click the *log out* button at the top of the page. This state is covered here: https://github.com/mozilla/addons-frontend/pull/5997/files#diff-5d954e9f650d2beb87d2ca0344afe003R502

kumar303:
@bobsilverberg thanks for the great feedback. It's ready for another look now.

src/amo/components/RatingManager/index.js
```diff
@@ -175,47 +206,78 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+              review={!this.isSignedIn() ? null : userReview}
+            />
+          </div>
+        </fieldset>
+      </form>
+    );
+  }
+
+  renderInlineReviewControls() {
+    const { i18n, userReview } = this.props;
```
bobsilverberg: This no longer needs `userReview`.

src/amo/components/RatingManager/index.js
```diff
@@ -175,47 +206,78 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+    const { showTextEntry } = this.state;
```
bobsilverberg: This no longer needs `showTextEntry`.

src/amo/components/RatingManager/index.js
```diff
@@ -175,47 +206,78 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+    ) : (
+      this.renderUserRatingForm()
+    );
+  }
+
+  render() {
+    const { _config, addon, version } = this.props;
+
+    invariant(addon, 'addon is required');
+    invariant(version, 'version is required');
```
bobsilverberg: This doesn't actually seem to need `version` at all.

src/ui/components/Rating/index.js
```diff
@@ -71,7 +71,7 @@ export class RatingBase extends React.Component<InternalProps, StateType> {
   // that surrounds the read-only set of stars.
   renderTitle = (
     rating: ?number,
-    readOnly: boolean,
+    readOnly: boolean | void,
```
rebmullin: is this the same as `readOnly?: boolean`? above we have it that way ..so I'm just maybe not clear on the difference so just double checking : )

src/amo/components/AddonReviewManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+  errorHandler: ErrorHandlerType,
+  i18n: I18nType,
+  flashMessage?: FlashMessageType | void,
+|};
+
+export class AddonReviewManagerBase extends React.Component<InternalProps> {
+  onSubmitRating = (rating: number) => {
+    const { errorHandler, dispatch, review } = this.props;
+
+    dispatch(
```
rebmullin: should we have invariant checks here?

src/amo/components/AddonReviewManager/index.js
rebmullin: could this be a Link tag (if we are not using target=_blank)?

src/amo/components/RatingManager/index.js
kumar303: `addon` and `version` are the two external props that are required. Since `RatingManager` is used by `Addon` which doesn't have Flow coverage, I decided we should put some `invariant()` calls in `render()` just to catch any mistakes.

src/amo/components/RatingManager/index.js
```diff
@@ -175,47 +206,77 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
...
+
+    return this.shouldShowTextEntry() ? (
+      <Button
+        href="#cancelTextEntry"
+        onClick={this.cancelTextEntry}
+        className="RatingManager-cancelTextEntryButton"
+        buttonType="neutral"
+        puffy
+      >
+        {i18n.gettext("Nevermind, I don't want to write a review")}
```
rebmullin: this might be more of a UX question, but if the use already has a review, maybe this text should change to indicate that? Otherwise maybe it's unclear that it won't delete the review (if it exists already)..?

src/ui/components/Rating/index.js
kumar303: `renderTitle` is called with positional arguments, like `renderTitle(3, true, 4)`. This kind of function is hard to maintain in JavaScript -- it's easy to mess up a call when you are refactoring code. We use Flow to catch mistakes.

If I were to define the `readOnly` parameter (positional argument 2) as `readOnly?` then Flow would allow that argument to be *omitted* from the call. I don't want that. However, I do want to allow the *value* of `readOnly` to be a literal `undefined`.

src/amo/components/AddonReviewManager/index.js
kumar303: It's not necessary because all of these properties cannot be empty (see [InternalProps](https://github.com/mozilla/addons-frontend/pull/5997/files/0887b8484e3fc3f4de326e36109d599fbb9dcebf#diff-dddd6fbd1ffe3a0900aa9b0da5db5fe8R31)). If a caller leaves them empty, Flow will show an error about it. Thus, we don't need any extra `invariant()` calls.

rebmullin: ok, sorry for dupe questions :/ I guess I still have to sort this out in my mind a bit..I think b/c I remember adding them in some other scenarios. Maybe render..

src/amo/components/AddonReviewManager/index.js
rebmullin: I mean if we are linking internally I guess I thought the Link tag was usually better to keep as SPA?

src/ui/components/Rating/index.js
rebmullin: ahh i see. cool, thanks for breaking it down : )

src/amo/components/RatingManager/index.js
kumar303: I totally missed that -- thanks! I will fix it.

kumar303:
Thanks again for all the reviews

Issue #6012, opened 2018-08-22
Saga / Reducer for flashing messages about the inline review form


kumar303:
There are two parts of the form for https://github.com/mozilla/addons-frontend/issues/5942 that are saved independently: the star rating and the review text. We need to temporarily show ("flash") messages about each action.

PR #6013, opened 2018-08-22
Saga/Reducer for flashing review related messages

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6012

Here's a visual of what I'd like to do with the data (below). The *saving* and *saved* message will flash on the screen temporarily. Also, the review form needs to be handled independently so I need a separate message to disable the submit button while it's in progress.

<details>
<summary>See a rough demo of the UI</summary>

addons-frontend-inline-review mov

</details>

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -18,6 +18,8 @@ export const FETCH_GROUPED_RATINGS: 'FETCH_GROUPED_RATINGS' =
   'FETCH_GROUPED_RATINGS';
 export const FETCH_REVIEWS: 'FETCH_REVIEWS' = 'FETCH_REVIEWS';
 export const FETCH_USER_REVIEWS: 'FETCH_USER_REVIEWS' = 'FETCH_USER_REVIEWS';
+export const FLASH_REVIEW_MESSAGE: 'FLASH_REVIEW_MESSAGE' =
+  'FLASH_REVIEW_MESSAGE';
 export const HIDE_EDIT_REVIEW_FORM: 'HIDE_EDIT_REVIEW_FORM' =
   'HIDE_EDIT_REVIEW_FORM';
 export const HIDE_REPLY_TO_REVIEW_FORM: 'HIDE_REPLY_TO_REVIEW_FORM' =
@@ -582,5 +584,32 @@ export const updateAddonReview = ({
   return {
     type: UPDATE_ADDON_REVIEW,
     payload: { body, errorHandlerId, rating, reviewId },
+  };
+};
+
+export const ABORTED: 'aborted' = 'aborted';
+export const SAVED_RATING: 'saved-rating' = 'saved-rating';
+export const SAVED_REVIEW: 'saved-review' = 'saved-review';
+export const STARTED_SAVE_RATING: 'started-save-rating' = 'started-save-rating';
+export const STARTED_SAVE_REVIEW: 'started-save-review' = 'started-save-review';
+
+export type FlashMessageType =
+  | typeof ABORTED
+  | typeof SAVED_RATING
+  | typeof SAVED_REVIEW
+  | typeof STARTED_SAVE_RATING
+  | typeof STARTED_SAVE_REVIEW;
+
+export type FlashReviewMessageAction = {|
+  type: typeof FLASH_REVIEW_MESSAGE,
+  payload: { message: FlashMessageType | void },
+|};
+
+export const flashReviewMessage = (
+  message: FlashMessageType | void,
+): FlashReviewMessageAction => {
+  return {
+    type: FLASH_REVIEW_MESSAGE,
+    payload: { message },
   };
 };
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -594,6 +594,7 @@ export const STARTED_SAVE_RATING: 'started-save-rating' = 'started-save-rating';
 export const STARTED_SAVE_REVIEW: 'started-save-review' = 'started-save-review';
 
 export type FlashMessageType =
+  | void
   | typeof ABORTED
   | typeof SAVED_RATING
   | typeof SAVED_REVIEW
@@ -602,11 +603,11 @@ export type FlashMessageType =
 
 export type FlashReviewMessageAction = {|
   type: typeof FLASH_REVIEW_MESSAGE,
-  payload: { message: FlashMessageType | void },
+  payload: { message: FlashMessageType },
 |};
 
 export const flashReviewMessage = (
-  message: FlashMessageType | void,
+  message: FlashMessageType,
 ): FlashReviewMessageAction => {
   return {
     type: FLASH_REVIEW_MESSAGE,
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -20,6 +20,8 @@ export const FETCH_REVIEWS: 'FETCH_REVIEWS' = 'FETCH_REVIEWS';
 export const FETCH_USER_REVIEWS: 'FETCH_USER_REVIEWS' = 'FETCH_USER_REVIEWS';
 export const FLASH_REVIEW_MESSAGE: 'FLASH_REVIEW_MESSAGE' =
   'FLASH_REVIEW_MESSAGE';
+export const HIDE_FLASHED_REVIEW_MESSAGE: 'HIDE_FLASHED_REVIEW_MESSAGE' =
+  'HIDE_FLASHED_REVIEW_MESSAGE';
 export const HIDE_EDIT_REVIEW_FORM: 'HIDE_EDIT_REVIEW_FORM' =
   'HIDE_EDIT_REVIEW_FORM';
 export const HIDE_REPLY_TO_REVIEW_FORM: 'HIDE_REPLY_TO_REVIEW_FORM' =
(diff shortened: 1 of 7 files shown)


src/amo/reducers/reviews.js
```diff
@@ -91,6 +94,8 @@ export type ReviewsState = {|
   view: {
     [reviewId: number]: ViewStateByReviewId,
   },
+  // Short-lived messages about reviews.
+  flashMessage?: FlashMessageType,
```
rebmullin: I see this could be undefined but even so should be added to the initial Review State? or is that intentionally left out b/c of this?

src/amo/actions/reviews.js
```diff
@@ -584,3 +586,30 @@ export const updateAddonReview = ({
...
+export type FlashMessageType =
+  | typeof ABORTED
+  | typeof SAVED_RATING
+  | typeof SAVED_REVIEW
+  | typeof STARTED_SAVE_RATING
+  | typeof STARTED_SAVE_REVIEW;
+
+export type FlashReviewMessageAction = {|
+  type: typeof FLASH_REVIEW_MESSAGE,
+  payload: { message: FlashMessageType | void },
```
rebmullin: can `void` just be added to `flashMessageType` above?

src/amo/sagas/reviews.js
```diff
@@ -50,6 +56,9 @@ import type {
   UpdateAddonReviewAction,
 } from 'amo/actions/reviews';
 
+// Number of millesconds that a message should be flashed on screen.
+export const FLASH_MESSAGE_DURATION = 2000;
```
rebmullin: It seems that this only applies to the SAVED messages duration.?

src/amo/sagas/reviews.js
```diff
@@ -200,7 +209,16 @@ function* manageAddonReview(
   const { body, errorHandlerId, rating } = action.payload;
   const errorHandler = createErrorHandler(errorHandlerId);
 
+  const savingRating = !!rating;
+  const savingReview = !!body;
```
rebmullin: With the new set up it always be one or the other (even on the initial create), is that right?

src/amo/sagas/reviews.js
```diff
@@ -235,9 +253,23 @@ function* manageAddonReview(
...
+    // Make the message disappear after some time.
+    yield new Promise((resolve) => setTimeout(resolve, FLASH_MESSAGE_DURATION));
+    yield put(flashReviewMessage(undefined));
   } catch (error) {
-    log.warn(`Failed to submit review with action ${action.type}: ${error}`);
+    log.warn(
+      `Failed to create/update review with action ${action.type}: ${error}`,
+    );
     yield put(errorHandler.createErrorAction(error));
+    yield put(flashReviewMessage(ABORTED));
```
rebmullin: on error, after aborting, should we reset and set back to undefined? or is that overkill?

src/amo/sagas/reviews.js
kumar303: It could be both. They are independent of each other.

src/amo/actions/reviews.js
kumar303: Well, huh. I initially wanted it like that but Flow gave me some errors that I couldn't figure out. I just tried it again and Flow is happy 🤷‍♂️  The error must have been related to calling code that is now fixed.

src/amo/sagas/reviews.js
kumar303: Correct. I can rename it.

src/amo/sagas/reviews.js
kumar303: I considered it but with that approach the caller would never know if the saga had aborted or if the message had just faded away. Nothing is relying on the abort message yet so it doesn't matter. The message just needs to change into *something* on error to make the other one fade away.

src/amo/reducers/reviews.js
kumar303: Yeah, probably a good idea. Done.

kumar303:
@rebmullin I think I addressed all your questions. Can you take another look?

src/amo/actions/reviews.js
```diff
@@ -584,3 +586,31 @@ export const updateAddonReview = ({
     payload: { body, errorHandlerId, rating, reviewId },
   };
 };
+
+export const ABORTED: 'aborted' = 'aborted';
+export const SAVED_RATING: 'saved-rating' = 'saved-rating';
+export const SAVED_REVIEW: 'saved-review' = 'saved-review';
+export const STARTED_SAVE_RATING: 'started-save-rating' = 'started-save-rating';
```
willdurand: Why did you switch to lowercase for flow types/values?

src/amo/sagas/reviews.js
```diff
@@ -235,9 +253,25 @@ function* manageAddonReview(
     );
 
     yield put(setReview(reviewFromResponse));
+
+    if (savingRating) {
+      yield put(flashReviewMessage(SAVED_RATING));
+    }
+    if (savingReview) {
+      yield put(flashReviewMessage(SAVED_REVIEW));
```
willdurand: Is there any flickering if both `savingRating` and `savingReview` are `true`?

src/amo/sagas/reviews.js
```diff
@@ -235,9 +253,25 @@ function* manageAddonReview(
...
+    }
+
+    // Make the message disappear after some time.
+    yield new Promise((resolve) =>
+      setTimeout(resolve, FLASH_SAVED_MESSAGE_DURATION),
+    );
+    yield put(flashReviewMessage(undefined));
```
willdurand: That's a bit weird to me, I'd rather write another action creator (`hideFlashReviewMessage`), even if it does the same thing under the hood.

src/amo/sagas/reviews.js
```diff
@@ -235,9 +253,25 @@ function* manageAddonReview(
...
+      yield put(flashReviewMessage(SAVED_RATING));
+    }
+    if (savingReview) {
+      yield put(flashReviewMessage(SAVED_REVIEW));
+    }
+
+    // Make the message disappear after some time.
+    yield new Promise((resolve) =>
+      setTimeout(resolve, FLASH_SAVED_MESSAGE_DURATION),
+    );
```
willdurand: Does that block the saga for a certain `FLASH_SAVED_MESSAGE_DURATION`? 🤔

tests/unit/helpers.js
```diff
@@ -701,3 +701,44 @@ export const createFakeTracking = (overrides = {}) => {
...
+
+// Save a reference to the real setTimeout in case a test uses a mock
+// sinon clock.
+const globalSetTimeout = setTimeout;
+
+/*
+ * Wait until the saga dispatches an action causing isMatch(action)
+ * to return true. Throw an error if it doesn't.
+ */
+export async function matchingSagaAction(
```
willdurand: I don't see any test case for this helper function.

tests/unit/amo/sagas/test_reviews.js
```diff
@@ -577,5 +602,85 @@ describe(__filename, () => {
 
       expect(action).toEqual(expectedAction);
     });
+
+    it('flashes the start of saving a rating', async () => {
```
willdurand: is "to flash" a verb? I always used "flash" as an adjective for "flash messages" (as opposed to persistent messages).

src/amo/sagas/reviews.js
willdurand: I was about to say that given the non-trivial code required to write test cases, that might not be a good idea, but then I checked the redux-saga doc and this seems a rather used pattern since there is a dedicated effect for that: https://redux-saga.js.org/docs/api/#delayms-val

src/amo/actions/reviews.js
kumar303: The lower case values are just what gets saved to Redux state. I don't think they need to be upper case. This is also consistent with how we do it elsewhere:

https://github.com/mozilla/addons-frontend/blob/c698d47a25d7f0b93626f0ecff2691817d80061e/src/amo/constants.js#L2-L9

https://github.com/mozilla/addons-frontend/blob/c698d47a25d7f0b93626f0ecff2691817d80061e/src/amo/reducers/recommendations.js#L13-L16

src/amo/sagas/reviews.js
kumar303: Oh, good call on `delay()`. It looks like that [implements cancellation](https://github.com/redux-saga/redux-saga/blob/master/packages/delay-p/src/index.js#L10) which seems important. I'll try to use `delay()` instead.

src/amo/actions/reviews.js
willdurand: well, it would be more consistent if we had only one case for all redux constants.

tests/unit/helpers.js
kumar303: Whoops, good call. We should cover throwing on timeout.

tests/unit/amo/sagas/test_reviews.js
kumar303: I always thought of it as a verb in this context to say *flash a message on the screen temporarily*.

src/amo/sagas/reviews.js
kumar303: In this UI it won't be possible to trigger both at the same time. If it comes up, we could handle it pretty easily by accepting both messages.

src/amo/actions/reviews.js
kumar303: I didn't really want to use constants at all here because I thought Flow would be enough to prevent mistakes. However, Flow won't show an error if you make an equality comparison with a bogus string like:

```js
if (flashMessage === 'started-typo') {
}
```

...which may be a Flow bug but I can think of why they may not want it to be an error. The constants will help prevent mistakes like that.

kumar303:
Thanks for the reviews, these were really helpful.

Issue #6007, opened 2018-08-22
User profile pages return a 5xx if the username starts with a 0


diox:
Copied over from https://github.com/mozilla/addons-server/issues/9137 which was filed by @bqbn :

### Describe the problem and steps to reproduce it:
Looking at our daily access logs, I'm surprised to see that 50% of the 5xx are caused by people access users' profile pages. For instance, on 8/10/2018 we had 173 5xx on one proxy instance, and 86 of them happened when people accessing `/<locale>/users/<user_id>` endpoint.

And particularly, [1] accounted for 40% of all 5xx, or 69 of them. The `0xy9en` user seems to have a few popular addons (for example [2] and [3]) and thus its user profile page attracted many hits, but unfortunately they all ended up with 5xx.

[1] https://addons.mozilla.org/en-US/firefox/user/0xy9en/
[2] https://addons.mozilla.org/en-US/firefox/addon/hotspot-shield-free-vpn-proxy-/
[3] https://addons.mozilla.org/en-US/firefox/addon/touch-vpn/?src=search

### What happened?

Not sure. A full list of user profile endpoints that returned 5xx is following:

```
/tr/firefox/user/000mayte000/
/fr/firefox/user/00Thathy00/
/en-GB/android/user/0SuperNova0/
/en-US/firefox/user/0owesleyo0/
/bg/firefox/user/0Adrian0Avalos0/
/tr/firefox/user/0_0_HellishHeaven_In_Flesh123456/
/ru/firefox/user/0veR/
/ru/firefox/user/000MIIX/
/ru/firefox/user/0x6b/
/nl/firefox/user/00adef/
/en-GB/firefox/user/0o0rsi/
/en-US/firefox/user/0leg-cool/
/en-US/firefox/user/0xy9en/
/fa/firefox/user/0xy9en/
/hsb/firefox/user/0xy9en/
/id/firefox/user/0xy9en/
/vi/firefox/user/0xy9en/
/ar/firefox/user/0xy9en/
/es/firefox/user/0xy9en/
/ru/firefox/user/0xy9en/
/en-US/firefox/user/00crimson/
/pt-BR/firefox/user/000000000o/
/zh-CN/android/user/0l1v3r/
/ja/android/user/0langy/
/zh-TW/android/user/0xSobky/
/ru/firefox/user/0xy9en/?src=api
/zh-CN/android/user/0bruno/?src=addon-detail
/uk/firefox/user/0l1v3r/?src=addon-detail
```

[...]

PR #6020, opened 2018-08-22
Handle zero-prefixed usernames in UserProfile

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6007

I checked all other usage of `parseInt()` and I don't think we need to fix anything else. It's surprising that `parseInt('0something', 10)` returns `0` 🤡 👯

### DIFF
--- a/src/amo/components/UserProfile/index.js
+++ b/src/amo/components/UserProfile/index.js
@@ -448,7 +448,7 @@ export function mapStateToProps(state: AppState, ownProps: Props) {
   // `getUserByUsername()` requires a string as second argument.
   let user = getUserByUsername(state.users, `${username}`);
 
-  if (!user) {
+  if (!user && /^[0-9]+$/.test(username)) {
     const userId = parseInt(username, 10);
     user = !Number.isNaN(userId) ? getUserById(state.users, userId) : undefined;
   }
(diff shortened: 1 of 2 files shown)


src/amo/components/UserProfile/index.js
```diff
@@ -448,7 +448,7 @@ export function mapStateToProps(state: AppState, ownProps: Props) {
   // `getUserByUsername()` requires a string as second argument.
   let user = getUserByUsername(state.users, `${username}`);
 
-  if (!user) {
+  if (!user && /^[0-9]+$/.test(username)) {
```
willdurand: I have been thinking about this patch and one edge case would be `username = 0`.

kumar303: Any username consisting of all numbers would defeat it, yes. I tried it and it's not possible:

screenshot 2018-08-23 12 33 44

But I tried many numbers at first (including `0`) and they were all taken usernames 😛 That might just be dirty dev data, I'm not sure.

willdurand: Yep but 0 is a special case, no? It will throw an error because of the invariant.

kumar303: Ah, right, probably. We'll see if `0` sends us a bug report. Until then I'm not worried.

PR #6028, opened 2018-08-23
Remove disco react warnings

### LINKED ISSUE #6059: Cleanup errors/warnings when running `yarn disco`
There are a lot of errors related to invalid props when starting disco locally. We should fix those errors. See #5860 for amo.

For QA:

- the install button should continue to work in -stage (which is not the same button as in -dev)
- the disco pane should work as intended in -dev (i.e. installing/removing add-ons with the mozAddonManager)

willdurand:
Fix #6059

---

This PR removes react warnings in the disco app, mostly by fixing invalid prop types.

### DIFF
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -8,14 +8,11 @@ import {
 } from 'core/constants';
 import { addQueryParams } from 'core/utils';
 
-const testHasWindow = () => typeof window !== 'undefined';
-
-export function hasAddonManager({ hasWindow = testHasWindow, navigator } = {}) {
-  // Returns undefined if it cannot be determined if mozAddonManager is supported (likely server
-  // rendering with no window). Otherwise returns true/false based on mozAddonManager in navigator.
-  if (!navigator && !hasWindow()) {
-    return undefined;
+export function hasAddonManager({ navigator } = {}) {
+  if (typeof window === 'undefined') {
+    return false;
   }
+
   return 'mozAddonManager' in (navigator || window.navigator);
 }
 
--- a/src/disco/components/DiscoPane/index.js
+++ b/src/disco/components/DiscoPane/index.js
@@ -38,6 +38,7 @@ export class DiscoPaneBase extends React.Component {
     }).isRequired,
     results: PropTypes.arrayOf(PropTypes.object).isRequired,
     _addChangeListeners: PropTypes.func,
+    _config: PropTypes.object,
     _tracking: PropTypes.object,
   };
 
@@ -45,6 +46,7 @@ export class DiscoPaneBase extends React.Component {
     AddonComponent: Addon,
     mozAddonManager: config.get('server') ? {} : navigator.mozAddonManager,
     _addChangeListeners: addChangeListeners,
+    _config: config,
     _tracking: tracking,
   };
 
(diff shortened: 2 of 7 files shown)


src/core/addonManager.js
```diff
@@ -8,14 +8,11 @@ import {
...
 
-const testHasWindow = () => typeof window !== 'undefined';
-
-export function hasAddonManager({ hasWindow = testHasWindow, navigator } = {}) {
-  // Returns undefined if it cannot be determined if mozAddonManager is supported (likely server
-  // rendering with no window). Otherwise returns true/false based on mozAddonManager in navigator.
-  if (!navigator && !hasWindow()) {
-    return undefined;
+export function hasAddonManager({ navigator } = {}) {
+  if (typeof window === 'undefined') {
```
rebmullin: should we still do a check for navigator? is it possible it won't be defined either?

tests/unit/disco/components/TestDiscoPane.js
```diff
@@ -327,12 +319,25 @@ describe(__filename, () => {
   });
 
   describe('componentDidMount', () => {
-    it('sets events', () => {
+    it('does not set events on the server', () => {
```
rebmullin: I guess this ties in to your above change but still not clear why this change is needed..

src/disco/components/DiscoPane/index.js
```diff
@@ -81,9 +83,15 @@ export class DiscoPaneBase extends React.Component {
   componentDidMount() {
     const {
       _addChangeListeners,
+      _config,
       handleGlobalEvent,
       mozAddonManager,
     } = this.props;
+
+    if (_config.get('server')) {
```
rebmullin: I'm just curious what issue is this fixing...

src/core/addonManager.js
willdurand: It does not matter at this point, because the next condition will be false.

tests/unit/disco/components/TestDiscoPane.js
willdurand: because it was the same result before, but very obscure. it

src/disco/components/DiscoPane/index.js
willdurand: it does not fix an issue but make the code more readable.

willdurand: it is related to your previous question/my previous answer.

src/core/addonManager.js
rebmullin: 🤦‍♂️ yep I see : )

src/disco/components/DiscoPane/index.js
rebmullin: is this even needed tho? I thought componentDidMount lifecycle was only called on client?

src/disco/components/DiscoPane/index.js
```diff
@@ -38,13 +38,15 @@ export class DiscoPaneBase extends React.Component {
...
     _addChangeListeners: PropTypes.func,
+    _config: PropTypes.object,
     _tracking: PropTypes.object,
   };
 
   static defaultProps = {
     AddonComponent: Addon,
     mozAddonManager: config.get('server') ? {} : navigator.mozAddonManager,
     _addChangeListeners: addChangeListeners,
+    _config: config,
```
rebmullin: just checking..do we still need this?

willdurand: ah nope

willdurand: removed!

PR #6029, opened 2018-08-23
Fix CSP for disco and amo locally

### LINKED ISSUE #2984: Turn on CSP for local development
We rely on CSP (Content Security Policy) rules in production and on our hosted dev servers but we don't have CSP enabled for local development. This is mostly because the dynamic assets served by the webpack server are a bit tricky to deal with. If possible, let's turn on CSP so that our local environment will see failures just like production.

willdurand:
Fix #2984

---

Explanations in comments.

I got a lot of CSP issues in the disco app locally, so I investigated a bit. It turned out that amo had CSP disabled locally but not disco. I tried to fix the CSP errors for disco and tried to enable CSP on amo, which actually worked.

### DIFF
--- a/src/core/middleware/security.js
+++ b/src/core/middleware/security.js
@@ -37,7 +37,7 @@ export function csp({ _config = config, noScriptStyles, _log = log } = {}) {
     _config.get('CSP') !== 'false' ? deepcopy(_config.get('CSP')) : false;
 
   if (cspConfig) {
-    if (noScriptStyles) {
+    if (noScriptStyles && !_config.get('isDevelopment')) {
       const hash = crypto
         .createHash('sha256')
         .update(noScriptStyles)
--- a/src/core/middleware/security.js
+++ b/src/core/middleware/security.js
@@ -2,6 +2,7 @@ import crypto from 'crypto';
 import fs from 'fs';
 import path from 'path';
 
+import { oneLine } from 'common-tags';
 import helmet from 'helmet';
 import config from 'config';
 import deepcopy from 'deepcopy';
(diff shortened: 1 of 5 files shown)


webpack.dev.config.babel.js
```diff
@@ -44,7 +44,7 @@ for (const app of appsBuildList) {
 }
 
 export default Object.assign({}, webpackConfig, {
-  devtool: 'cheap-module-eval-source-map',
+  devtool: 'cheap-module-source-map',
```
willdurand: By switching to `cheap-module-source-map`, we don't use `eval` anymore, avoiding LOTS of CSP errors. It is slightly slower apparently, but I did not notice any big impact.

src/core/middleware/security.js
```diff
@@ -37,7 +37,7 @@ export function csp({ _config = config, noScriptStyles, _log = log } = {}) {
     _config.get('CSP') !== 'false' ? deepcopy(_config.get('CSP')) : false;
 
   if (cspConfig) {
-    if (noScriptStyles) {
+    if (noScriptStyles && !_config.get('isDevelopment')) {
```
kumar303: I'm ambivalent about using `!isDevelopment` here because it obscures the fact that the local server is relying on `unsafe-inline`. Maybe one day the local server will not.

I thought of suggesting `!config.get('allowUnsafeInline')` to make it more explicit but I also like how `isDevelopment` is something that would never leak into production.

What do you think?

kumar303: I think it would be good to add [a test for this](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/core/middleware/test_cspMiddleware.js).

willdurand: ah, right

willdurand: I also thought about introducing a new config variable but as you wrote already, too many config vars may lead to vars leaking into the wrong environment.

It seems slightly safer to me to use `isDevelopment`.

willdurand: I have added a `log.debug` call when we omit the hash-source (and therefore allow `unsafe-inline`).

Issue #6014, opened 2018-08-22
Show loading indicator for rating stars


kumar303:
We never implemented a loading indicator for rating stars when the add-on detail page loads. It could benefit from one. It's definitely needed for https://github.com/mozilla/addons-frontend/issues/5942 (a new inline review form) because the stars look jumpy when you save them.

**For QA**:

- Sign in and rate an add-on on the detail page by selecting its stars
- Refresh the page
- You should now see an indication that the saved star rating is loading before it shows your previously saved rating.

hritvi:
Hey @kumar303. Can I work on this issue? Thanks in advance :)

ioanarusiczki:
Verified fixed on AMO dev with FF61(Win10&Android 8.0)
The loading indicator for the rated stars

![loading state for rated stars](https://user-images.githubusercontent.com/33448286/44712213-f5d6ce80-aab8-11e8-9e81-d80a0297cc1a.gif)

kumar303:
Hello @hritvi. Sorry but I missed your question until just now. The best way to find issues is to look through the [`contrib: welcome`](https://github.com/mozilla/addons-frontend/labels/contrib%3A%20welcome) label.

PR #6031, opened 2018-08-23
Show loading indicator for rating stars

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6014

<details>
<summary>Large stars example</summary>

addons-frontend-stars-big mov

</details>

<details>
<summary>Small stars example</summary>

addons-frontend-stars-small mov

</details>

### DIFF
--- a/src/amo/components/RatingManager/index.js
+++ b/src/amo/components/RatingManager/index.js
@@ -83,15 +83,9 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
   }
 
   componentDidMount() {
-    const {
-      apiState,
-      loadSavedReview,
-      userId,
-      addon,
-      userReview,
-      version,
-    } = this.props;
-    if (userId && addon && userReview === undefined) {
+    const { apiState, loadSavedReview, userId, addon, version } = this.props;
+
+    if (this.shouldLoadUserReview()) {
       log.debug(`Loading a saved rating (if it exists) for user ${userId}`);
       loadSavedReview({
         apiState,
@@ -101,6 +95,12 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
         versionId: version.id,
       });
     }
+  }
+
+  shouldLoadUserReview() {
+    const { userId, userReview } = this.props;
+    // Only try to load a userReview when the user is signed in.
+    return Boolean(userId && userReview === undefined);
   }
 
   onSelectRating = (rating: number) => {
@@ -196,9 +196,12 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
   };
 
   render() {
-    const { i18n, addon, userId, userReview } = this.props;
+    const { i18n, addon, userId, userReview, version } = this.props;
     const { showTextEntry } = this.state;
     const isLoggedIn = Boolean(userId);
+
+    invariant(addon, 'addon is required');
+    invariant(version, 'version is required');
 
     const prompt = i18n.sprintf(
       i18n.gettext('How are you enjoying your experience with %(addonName)s?'),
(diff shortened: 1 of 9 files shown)


src/ui/components/UserRating/index.js
```diff
@@ -12,6 +12,7 @@ import type { UserReviewType } from 'amo/actions/reviews';
 type Props = {|
   className?: string,
   isOwner?: boolean,
+  loading?: boolean,
```
bobsilverberg: Wouldn't `loading` just be the same as `!readOnly && review === undefined`? Or do you need this for testing purposes?

tests/unit/ui/components/TestRating.js
```diff
@@ -228,7 +239,7 @@ describe(__filename, () => {
   });
 
   it('selects stars on hover', () => {
-    const root = render();
+    const root = render({ loading: false });
```
bobsilverberg: You mostly use `renderWithRating()` to end up with a component that isn't in a loading state in these tests (including in "does nothing when you hover over stars", below), but for these two tests you use `render({ loading: false })`. Is there a reason for that difference in these two tests?

src/amo/components/RatingManager/index.js
```diff
@@ -196,10 +196,13 @@ export class RatingManagerBase extends React.Component<InternalProps, State> {
   };
 
   render() {
-    const { i18n, addon, userId, userReview } = this.props;
+    const { i18n, addon, userId, userReview, version } = this.props;
     const { showTextEntry } = this.state;
     const isLoggedIn = Boolean(userId);
 
+    invariant(addon, 'addon is required');
+    invariant(version, 'version is required');
```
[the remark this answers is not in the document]
kumar303: I removed an if statement that was checking the existence of `addon`. It wasn't necessary because `addon` is required. However, `<Addon>` is the caller of `<RatingManager>` and that component isn't covered by Flow so I thought it would be a good idea to add some invariant protection. While I was at it, I added protection for `version` which is the only other required prop.

src/ui/components/UserRating/index.js
kumar303: The `RatingManager` needs to present stars that the user can select to rate the component. In this case, `rating==undefined` because no rating has been set. My intention was to let `RatingManager` control the loading state but let me double check to see I still need to do that.

kumar303:
Thanks. I think I addressed all comments. Can you take another look?

In the previous approach, I added the `loading` prop to allow a calling component to control the loading state in addition to an implicit loading state. This adds a dimension of complexity but I thought I needed it. 

After your comments, I took a fresh look and found out how to do everything with an implicit loading state. This simplified the logic in `Rating`. One drawback is it makes the calling code a bit less explicit but that's probably OK.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -124,6 +124,38 @@ describe(__filename, () => {
     sinon.assert.notCalled(loadSavedReview);
   });
 
+  it('renders loading stars before the saved review has loaded', () => {
+    const addon = createInternalAddon({ ...fakeAddon });
+    const { store } = dispatchSignInActions();
+
+    const root = render({ addon, store });
+
+    expect(root.find(UserRating)).toHaveProp('review', undefined);
```
bobsilverberg: I see what you mean about the impact of your recent change. The code does seem cleaner, but it's a bit less clear exactly how it's working. For example, in this test, it's described as "renders loading stars..." but doesn't have an assertion that seems to match (i.e., the assertion doesn't say anything about loading, rather it's about the contents of the `review` prop).

Maybe the descriptions of these tests should be updated to describe what we actually expect, which is a specific value for the `review` prop?

kumar303: Yeah, I can see how changing these descriptions could be helpful. The loading state is now the responsibility of `Rating`.

PR #6041, opened 2018-08-24
Add new config for ui-tests.

b4handjr:
Fix https://github.com/mozilla/addons/issues/12264

---

Just a small addition to assist in the ui tests.

### DIFF
--- a/config/uitests-amo.js
+++ b/config/uitests-amo.js
@@ -1,11 +1,12 @@
-// For use on addons-server, or with the command `amo:ui-tests` and the
-// HOSTNAME  variable as `ui-tests`.
+// For use on addons-server, or with the command `amo:ui-tests`.
 module.exports = {
   apiHost: 'http://olympia.test',
-  CSP: false,
   proxyApiHost: 'http://olympia.test',
   proxyPort: 3000,
   proxyEnabled: true,
+
+  CSP: false,
+
   // Setting this to false returns add-ons that are not compatible but means
   // developers can pull from a much larger dataset on the local/-dev/-stage
   // servers. Set this to true to only get compatible add-ons (this is what
@@ -14,6 +15,7 @@ module.exports = {
   fxaConfig: 'default',
   trackingEnabled: false,
   loggingLevel: 'debug',
-  enableAMInstallButton: true,
   enableNodeStatics: true,
+
+  enableAMInstallButton: true,
 };


config/uitests-amo.js
```diff
@@ -0,0 +1,15 @@
...
+  apiHost: 'http://olympia.test',
+  proxyApiHost: 'http://olympia.test',
+  proxyPort: 3000,
+  proxyEnabled: true,
+  // Setting this to false returns add-ons that are not compatible but means
+  // developers can pull from a much larger dataset on the local/-dev/-stage
+  // servers. Set this to true to only get compatible add-ons (this is what
+  // happens in production) but get a lot fewer add-ons in search results.
+  restrictSearchResultsToAppVersion: false,
+  fxaConfig: 'default',
```
willdurand: We should probably use the correct value here, i.e. `local`, and get rid of th `FXA_CONFIG` env var in the `package.json`.

config/uitests-amo.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  apiHost: 'http://olympia.test',
+  proxyApiHost: 'http://olympia.test',
+  proxyPort: 3000,
+  proxyEnabled: true,
```
willdurand: I see that the `uitests-disco` config has a different setup, why?

Also, why do we have a `uitests` config and a `functional-development-amo`  config?

config/uitests-amo.js
jrbenny35: Well I have to use default for the tests, that is why I am creating this config. I can just set it here though.

config/uitests-amo.js
jrbenny35: ```uitests-disco``` runs in a different environment and doesnt need fxa login. I should get rid of the other script as well.

willdurand: right but the api/proxy are not related to fxa, no?

jrbenny35: Not that I know of.

jrbenny35:
> Other question: how do you load this config given that NODE_CONFIG_ENV has not been set in package.json?

I use the node config option ```{short_hostname}-{instance}.EXT```. I set the host name in the addons-server docker-compose.yml.

willdurand:
> I use the node config option {short_hostname}-{instance}.EXT. I set the host name in the addons-server docker-compose.yml.

ah right, I forgot. Maybe adding a comment to those files would make sense because most of the config files use `NODE_CONFIG_ENV`.

config/uitests-amo.js
willdurand: does that mean `uitests-disco` relies on an API that is not reachable at `http://olympia.test`?

config/uitests-amo.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  // For use on addons-server, or with the command `amo:ui-tests` and the
+  // HOSTNAME  variable as `ui-tests`.
```
willdurand: Instead of writing this, add a `NODE_CONFIG_ENV` entry in `package.json` with `ui-tests` as value.

config/uitests-amo.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
willdurand: The comment you wrote below should be put above:

> For use on addons-server, or with the command `amo:ui-tests`.

config/uitests-amo.js
```diff
@@ -0,0 +1,17 @@
...
+  // Setting this to false returns add-ons that are not compatible but means
+  // developers can pull from a much larger dataset on the local/-dev/-stage
+  // servers. Set this to true to only get compatible add-ons (this is what
+  // happens in production) but get a lot fewer add-ons in search results.
+  restrictSearchResultsToAppVersion: false,
+  fxaConfig: 'default',
+  trackingEnabled: false,
+  loggingLevel: 'debug',
+  enableAMInstallButton: true,
+};
```
willdurand: Don't you need `enableNodeStatics: true,` here too? I see it used in different "test-like" envs.

package.json
```diff
@@ -62,6 +63,14 @@
       "env": {
         "AMO_CDN": "https://addons-dev-cdn.allizom.org",
         "PROXY_API_HOST": "https://addons-dev.allizom.org",
+        "FXA_CONFIG": "default",
```
willdurand: I think you just changed the FXA config of our local setup, which is not nice 😛 

It should be `local` here, not `default`.

package.json
```diff
@@ -62,6 +63,14 @@
...
+        "CSP": false,
+        "NODE_APP_INSTANCE": "amo"
+      }
+    },
+    "amo:ui-tests": {
+      "command": "better-npm-run start-dev-proxy",
+      "env": {
         "FXA_CONFIG": "local",
```
willdurand: Can you remove this entry? and also `CSP` below. It is better to use the config file.

willdurand: You can add `NODE_CONFIG_ENV: ui-tests` though.

config/uitests-amo.js
willdurand: You can specify `CSP: false` here too.

config/uitests-amo.js
willdurand: so, it should be `local` I guess.

package.json
willdurand: What? Then why having `FAX_CONFIG` set here?

config/uitests-amo.js
```diff
@@ -0,0 +1,19 @@
+// For use on addons-server, or with the command `amo:ui-tests` and the
+// HOSTNAME  variable as `ui-tests`.
```
willdurand: I don't think we should mention the `HOSTNAME` variable (in part because we load the config _via_ `NODE_CONFIG_ENV`)

config/uitests-amo.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = {
+  apiHost: 'http://olympia.test',
+  CSP: false,
```
willdurand: Can you put this line after `proxyEnabled` and add one empty line before and after? I'd like to make it super easy to see that `CSP` is disabled.

config/uitests-amo.js
```diff
@@ -0,0 +1,19 @@
...
+  proxyEnabled: true,
+  // Setting this to false returns add-ons that are not compatible but means
+  // developers can pull from a much larger dataset on the local/-dev/-stage
+  // servers. Set this to true to only get compatible add-ons (this is what
+  // happens in production) but get a lot fewer add-ons in search results.
+  restrictSearchResultsToAppVersion: false,
+  fxaConfig: 'default',
+  trackingEnabled: false,
+  loggingLevel: 'debug',
+  enableAMInstallButton: true,
```
willdurand: This is a feature-flag, can you add it at the end, with a blank line before?

config/uitests-amo.js
willdurand: Do you want `default` or `local` here?

PR #6045, opened 2018-08-24
Add more tracking events

### LINKED ISSUE #5940: Hook up all analytics/tracking to the mozAddonManager install button

Perform all the following actions with the new `mozAddonManager` install button from https://github.com/mozilla/addons-frontend/issues/3954 (this is only on in dev):
- install add-on / theme / etc...
- uninstall add-on / theme / etc...
- install and click the *Manage* button
- install but then cancel in the doorhanger prompt
- install an add-on in a way that fails (you can do this on dev by setting incorrect prefs)

Watch the developer console.

### What happened?

Some Google Analytics events are missing.

You should see Google Analytics tracking events for each of these actions.

~~At the very least, we need to hook up the equivalent of successful installs~~ (I think they are handled already in https://github.com/mozilla/addons-frontend/blob/4578186b4c2c3dcf1fcf4f0670ac2f8adcb5aa18/src/core/installAddon.js#L475).

There are a couple unused tracking methods in `AMInstallButton` that may need to be removed.

willdurand:
Fix #5940

---

- [x] install add-on / theme / etc...
- [x] uninstall add-on / theme / etc...
- [x] install and click the Manage button => assuming it is `enable`
- [x] install but then cancel in the doorhanger prompt
- [x] ~~install an add-on in a way that fails (you can do this on dev by setting incorrect prefs)~~

Update: I decided to not add error/failure events yet as it is unclear to me what we want to track.


src/core/constants.js
```diff
@@ -190,17 +190,34 @@ export const UNINSTALL_COMPLETE = 'UNINSTALL_COMPLETE';
...
 // Changing these strings will break existing statistics without
 // updating the category matching at the same time.
+export const ENABLE_THEME_CATEGORY = 'AMO Theme Activations';
+export const ENABLE_EXTENSION_CATEGORY = 'AMO Addon Activations';
+
 export const INSTALL_THEME_CATEGORY = 'AMO Theme Installs';
 export const INSTALL_EXTENSION_CATEGORY = 'AMO Addon Installs';
 
+export const INSTALL_THEME_CANCELLED_CATEGORY = 'AMO Theme Installs Cancelled';
+export const INSTALL_EXTENSION_CANCELLED_CATEGORY =
```
rebmullin: I am wondering it's better to have the more specific description first followed by the type (theme vs extension):
INSTALL_CANCELLED_THEME_CATEGORY
INSTALL_CANCELLED_EXTENSION_CATEGORY

INSTALL_ERRORED_THEME_CATEGORY
INSTALL_ERRORED_EXTENSION_CATEGORY (etc)

rebmullin: also - not sure since how it is important to notify anyone about these updates since we are not technically changing the existing ones, we are just adding new ones..

src/core/installAddon.js
```diff
@@ -122,11 +126,23 @@ export function makeProgressHandler(dispatch, guid) {
         type: INSTALL_CANCELLED,
         payload: { guid },
       });
+
+      _tracking.sendEvent({
+        action: getAddonTypeForTracking(type),
+        category: getAddonEventCategory(type, INSTALL_CANCELLED_ACTION),
```
rebmullin: is it possible to track this for open search and lang?

rebmullin: also, i can't seem to comment above (around like 119) - should this sendEvent INSTALL_ERROR_ACTION

src/core/constants.js
```diff
@@ -190,17 +190,34 @@ export const UNINSTALL_COMPLETE = 'UNINSTALL_COMPLETE';
 export const INSTALL_ERROR = 'INSTALL_ERROR';
 
 // Tracking install actions.
+export const ENABLE_ACTION = 'enable';
 export const INSTALL_ACTION = 'install';
+export const INSTALL_CANCELLED_ACTION = 'install:cancelled';
+export const INSTALL_ERRORED_ACTION = 'install:errored';
+export const INSTALL_FAILED_ACTION = 'install:failed';
```
rebmullin: what's the difference between errored and failed here? is install failed specific to that and error any other kind of error?

src/core/components/AMInstallButton/index.js
```diff
@@ -115,6 +115,10 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
       addonName: addon.name,
       type: addon.type,
     });
+    this.trackInstallSucceeded({
+      addonName: addon.name,
+      type: addon.type,
```
rebmullin: since type and name are used a few times, maybe we can extract them out here?

src/core/constants.js
```diff
@@ -190,17 +190,34 @@ export const UNINSTALL_COMPLETE = 'UNINSTALL_COMPLETE';
...
 export const INSTALL_STARTED_ACTION = 'install:started';
 export const UNINSTALL_ACTION = 'uninstall';
 
 // Tracking Event Categories.
 // WARNING: Do not change these without notifying data + metrics teams.
 // Changing these strings will break existing statistics without
 // updating the category matching at the same time.
+export const ENABLE_THEME_CATEGORY = 'AMO Theme Activations';
+export const ENABLE_EXTENSION_CATEGORY = 'AMO Addon Activations';
```
rebmullin: hmm I'm wondering if Activation could be clearer maybe..?

Bc install could mean this too, right? Could it just be `AMO [type] Enabled.` We are using past tense verbs below so maybe this is ok and better since it's consistent with those..? what do you think?

src/core/installAddon.js
rebmullin: oh I guess this is below?

src/core/installAddon.js
```diff
@@ -481,13 +516,19 @@ export class WithInstallHelpers extends React.Component {
...
           setInstallState({
             guid,
             status: ERROR,
             error: FATAL_INSTALL_ERROR,
           }),
         );
+        _tracking.sendEvent({
+          action: getAddonTypeForTracking(type),
+          category: getAddonEventCategory(type, INSTALL_ERRORED_ACTION),
+          label: name,
```
rebmullin: should we also sendEvent that install_failed_action? I am still not super clear on which one to use if you are choosing one over the other

tests/unit/core/test_tracking.js
```diff
@@ -279,6 +291,81 @@ describe(__filename, () => {
...
+      ).toEqual(INSTALL_EXTENSION_CANCELLED_CATEGORY);
+    });
+
+    it('returns the expected category when type is lightweight theme and installAction is cancelled', () => {
+      expect(
+        getAddonEventCategory(ADDON_TYPE_THEME, INSTALL_CANCELLED_ACTION),
+      ).toEqual(INSTALL_THEME_CANCELLED_CATEGORY);
+    });
+
+    it('returns the expected category when type is static theme and installAction is cancelled', () => {
```
rebmullin: I don't see this option to cancel with static theme, do you?

src/core/installAddon.js
```diff
@@ -405,11 +419,25 @@ export class WithInstallHelpers extends React.Component {
...
+      name,
+      type,
+    } = this.props;
 
     return _addonManager
       .enable(guid)
       .then(() => {
+        _tracking.sendEvent({
+          action: getAddonTypeForTracking(type),
+          category: getAddonEventCategory(type, ENABLE_ACTION),
```
rebmullin: this also comes up when we install a theme since we are still special casing this? is that ok?

tests/unit/core/test_tracking.js
```diff
@@ -279,6 +291,81 @@ describe(__filename, () => {
...
     });
+
+    it('returns the expected category when type is extension and installAction is cancelled', () => {
+      expect(
+        getAddonEventCategory(ADDON_TYPE_EXTENSION, INSTALL_CANCELLED_ACTION),
+      ).toEqual(INSTALL_EXTENSION_CANCELLED_CATEGORY);
+    });
+
+    it('returns the expected category when type is lightweight theme and installAction is cancelled', () => {
+      expect(
```
rebmullin: I don't see this option to cancel with lightweight theme, do you?

src/core/components/AMInstallButton/index.js
willdurand: yup

src/core/constants.js
willdurand: I was looking for a category name that says "Enable" like we have the "installs" and the "uninstalls", but I don't think the "enables" work, right?

src/core/constants.js
willdurand: You mean re-sorting the constants?

src/core/installAddon.js
willdurand: no I don't think it is possible because we do not have such an event.

willdurand: You mean sending an event for `DOWNLOAD_FAILED`?

src/core/installAddon.js
willdurand: ah, right. maybe we could add an argument to not send an event in that case.

src/core/installAddon.js
rebmullin: right or maybe that is covered below.

I haven't been able to test the error stuff myself yet..

src/core/installAddon.js
rebmullin: OR maybe we can add enable tracking to the others (extensions and lightweight theme) since it technically does do this on install..? but that might be more involved..

tests/unit/core/test_tracking.js
rebmullin: sorry just to clarify I see the following for extension (Add AND Cancel options):

Alt text

but I wasn't see this for any of the themes. If you see these then never mind..

src/core/constants.js
rebmullin: I just meant renaming them a bit. I might be overthinking this : ) but just thought it might be clearer when you first look at the code to see it the way I have above. 

Immediately you can see INSTALL_CANCELLED_[TYPE] or INSTALL_ERROR_[TYPE] vs the theme or extension breaking up the description (IMO). If that doesn't makes any sense or you disagree that it's more obvious this way, then just disregard. It was just a "thinking out loud" suggestion :)

src/core/constants.js
rebmullin: Right that didn't exist before. I just thought description "Enabled" might be more clear since this (currently) doesn't come up when you install (altho maybe it should as I noted below/on another comment)

src/core/installAddon.js
rebmullin: > is it possible to track this for open search and lang?

I saw when i tried install open search, there was a cancel option (altho it doesn't look a little different from a regular extension).

I guess you can disregard my comment above INSTALL_ERROR_ACTION for now as you said you mentioned that would be revisiting this :) Basically tho I wasn't sure if we should be tracking this altho I realized you might have already set that up somewhere else (in install catch?)

tests/unit/core/test_tracking.js
willdurand: ooh right!

src/core/constants.js
willdurand: I'm going to remove those FAILED/ERRORED events for now.

src/core/constants.js
willdurand: ah right, makes sense

tests/unit/core/test_tracking.js
willdurand: indeed, there is no button to cancel a theme installation (lightweight or static theme), but we have to cover this by test, no?

src/core/installAddon.js
willdurand: unfortunately, we cannot do anything for open search add-ons. I tried and failed. The `AddSearchProvider` opens a sort of confirm popup but there is no return value available and it is not a promise-based API.

src/core/installAddon.js
willdurand: I have added a prop to the `enable` helper, which prevents the tracking event to be sent when we are installing an add-on.

tests/unit/core/test_tracking.js
willdurand: I see what you mean but for completeness, we should test that, no?

willdurand:
@rebmullin just updated the code, thanks a lot for your review and super useful comments! I removed the error/failure tracking events for now.

tests/unit/core/test_installAddon.js
```diff
@@ -1068,6 +1105,40 @@ describe(__filename, () => {
...
+            category: getAddonEventCategory(
+              ADDON_TYPE_EXTENSION,
+              ENABLE_ACTION,
+            ),
+            label: addon.name,
+          });
+        });
+      });
+
+      it('does not end a tracking event when "sendTrackingEvent" is false', () => {
```
rebmullin: should "end" be "send" here?

tests/unit/core/test_installAddon.js
```diff
@@ -1011,20 +1027,38 @@ describe(__filename, () => {
...
+        });
       });
 
       it('sets status to error on onInstallFailed', () => {
         const dispatch = sinon.spy();
         const guid = '{my-addon}';
-        const handler = makeProgressHandler(dispatch, guid);
+        const name = 'my-addon';
+        const type = ADDON_TYPE_EXTENSION;
+        const handler = createProgressHandler({ dispatch, guid, name, type });
```
rebmullin: do we still need this update now (with name and type)?

src/core/installAddon.js
```diff
@@ -416,12 +422,28 @@ export class WithInstallHelpers extends React.Component {
...
+      guid,
+      iconUrl,
+      name,
+      type,
+    } = this.props;
 
     return _addonManager
       .enable(guid)
       .then(() => {
+        if (sendTrackingEvent) {
```
rebmullin: I don't see this being called/sent on disco. shouldn't it be?

src/core/constants.js
rebmullin: so you'll have a follow up issue for this?

tests/unit/core/test_tracking.js
rebmullin: yeh I see .. yep i think that makes sense : )

tests/unit/core/test_installAddon.js
willdurand: yep

tests/unit/core/test_tracking.js
rebmullin: yep, I see...that makes sense. thanks : )

src/core/installAddon.js
willdurand: mmh, how? I tried with a theme and an extension.

rebmullin: oh? does it work for you?

willdurand: Yes

rebmullin: hmm that's weird. I see it now too :/ ok cool : )

PR #6049, opened 2018-08-24
Add INACTIVE status for extensions

### LINKED ISSUE #5929: Button's state is "Enable" instead of "Add to Firefox" after clicking "Remove" from Add-on manager
STR:
1. Load AMO dev and go to an add-ons detail page. 
2. Click the "Add to Firefox" button. 
3. Open a new tab and from add-on manager click the "Remove" button for the installed add-on.
4. Go back to the add-ons page and observe the state of the button, refresh page.

Actual result:
The button has the "+Enable" state. 

Expected result:
The button has the "+Add to Firefox" and the add-on can be re-installed.

Notes:
This issue is reproducible with FF63 (Win10).  

enable state instead of add to firefox

willdurand:
Partially fix #5929

---

I said "partially" because there is some work to do on the browser side. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1477155#c1.

This allows to get a "Add to Firefox" button when an **extension** has been removed from "about:addons". It does not work for themes, though. 

I also fixed the test cases because `getAddonById` clearly does not return `type` as we are used to use: it is either `extension` or `theme`, not `persona`, `statictheme`, etc.

### DIFF
--- a/src/core/installAddon.js
+++ b/src/core/installAddon.js
@@ -24,6 +24,7 @@ import {
   FATAL_ERROR,
   FATAL_INSTALL_ERROR,
   FATAL_UNINSTALL_ERROR,
+  INACTIVE,
   INSTALL_ACTION,
   INSTALL_ERROR,
   INSTALL_CANCELLED,
@@ -73,6 +74,7 @@ import {
   USER_AGENT_OS_WINDOWS,
 } from 'core/reducers/api';
 import { showInfoDialog } from 'core/reducers/infoDialog';
+import { isTheme } from 'core/utils';
 
 export function installTheme(
   node,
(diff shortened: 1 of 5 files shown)


src/core/installAddon.js
```diff
@@ -377,8 +382,15 @@ export class WithInstallHelpers extends React.Component {
     return _addonManager
       .getAddon(guid)
       .then(
-        (addon) => {
-          const status = addon.isActive && addon.isEnabled ? ENABLED : DISABLED;
+        ({ isActive, isEnabled }) => {
```
kumar303: It would be really nice to define types and get Flow coverage for `addonManager()` because it's undocumented and hard to understand.

willdurand: Filed https://github.com/mozilla/addons-frontend/issues/6088

willdurand:
> The custom handling for themes is unfortunate but maybe you'll get some answers to that in the bug soon. If you don't hear back, send a need-info request to aswan. The bug you commented on is very low priority because about:addons is ancient code and no one wants to touch it unless they have a really good reason.

Thanks, I set a reminder in +8 days for myself.

PR #6060, opened 2018-08-24
Storybook setup start - phase 1

rebmullin:
notes/questions:

I really liked the knobs interface which can easily change out the props but for some reason I wasn't sure if this would work well with Percy(?):
https://www.npmjs.com/package/@storybook/addon-knobs

I haven't looked into Percy yet tho so for now I will leave this : )

Storybook version:
Using v3 / stable version
4 is in alpha (I haven’t really looked into this)

Where to eventually publish to:
GitHub pages, Heroku?

Percy (for snapshots):
https://docs.percy.io/docs/storybook-for-react
If we end up keeping similar setup, will work with this?
Specifically - will knobs addon work with Percy? 

Show additional/helpful information?
Do we want to show source?
Do we want to show propType info*?

* Note: with flow props, this is still TBD how to be done

Regarding the structure, I thought about moving these inside the src directory but leave as is for now.


willdurand:
Thanks for starting this! I think it is good to start with only two components.

Early feedback:

- can we NOT have webpack integration?
- can we try not to add a lot of plugins?

rebmullin:
hey @willdurand, thanks for the quick feedback : )

I could not get this up and running without overriding their [webpack](https://storybook.js.org/configurations/default-config/#webpack). It was complaining about sass and even the core/logger coming from the components. I don't know that it needs to include exactly *all* that we have set up for dev/prod but this was the quickest way I got it working :/

Oh? So we don't really want too many plugins? Can you clarify what is "a lot"? How do you feel about the knobs plugin (assuming it works ok w/ percy)?

also, any thoughts about showing propType info and source here? 


@kumar303 @bobsilverberg, if you have any additional input too that would be great:)

willdurand:
> I could not get this up and running without overriding their webpack. It was complaining about sass and even the core/logger coming from the components. I don't know that it needs to include exactly all that we have set up for dev/prod but this was the quickest way I got it working :/

Yeah, that's because Storybook also uses webpack, but they say we can use our webpack config if it is webpack2-compliant, which is maybe not the case (we use webpack 3): https://storybook.js.org/configurations/custom-webpack-config/. What you have seem to work though, it only requires a bit of cleanup and a simpler filename (`webpack.config.js`).

> Oh? So we don't really want too many plugins? Can you clarify what is "a lot"? How do you feel about the knobs plugin (assuming it works ok w/ percy)?
> 
> also, any thoughts about showing propType info and source here?

rebmullin:
thanks @willdurand..

> storybook.js.org/configurations/custom-webpack-config. What you have seem to work though, it only requires a bit of cleanup and a simpler filename (webpack.config.js).

yep initially I had started with this...something much simpler than what I have but the core/logger also caused issues. I think I needed one of the plugins you had set up for this (I don't remember which one specifically). But I spent a lot of time this before I decided to just plugin to what we already had set up and worked . I can look into what I can take out but it seem that we need a lot of this set up.

> I've seen some Router stuff too, have you see gvaldambrini/storybook-router:packages/react@master?

hmm I googled about this b/c I was definitely getting complaints about the store and found something similar to what I have (altho as I mentioned in my comments this needed revisiting). So yeh I can def revisit this. I don't think I saw this package tho...so I'll see if we can use that instead. That would be super if that resolves that issue.

I personally liked knobs but I see what you mean. I thought this might be an issue so I can definitely strip this out. I can also strip out the device and the other plugins (for now?) to make it a cleaner/smaller patch. I think that makes good sense.

thanks  :)

.storybook/config.js
```diff
@@ -0,0 +1,34 @@
...
+// Override some global-y setup options.
+// See https://www.npmjs.com/package/@storybook/addon-options.
+setOptions({
+  name: 'Mozilla Addons frontend',
+  url: 'https://github.com/mozilla/addons-frontend',
+});
+
+
+// Global Decorators.
+addDecorator(withKnobs)
```
kumar303: The *Knobs* section is super helpful

package.json
```diff
@@ -41,7 +41,9 @@
     "test-coverage-once": "bin/config-check.js && better-npm-run jest --coverage",
     "test-once": "bin/config-check.js && better-npm-run jest && npm run lint",
     "version-check": "bin/version-check.js",
-    "webpack-dev-server": "npm run build-locales && better-npm-run webpack-dev-server"
+    "webpack-dev-server": "npm run build-locales && better-npm-run webpack-dev-server",
+    "storybook": "NODE_PATH=src start-storybook -p 9001",
```
kumar303: I think if you adjusted the webpack config with `resolve` then you won't need `NODE_PATH` (but I'm not sure).

package.json
```diff
@@ -228,6 +230,14 @@
     "webpack-isomorphic-tools": "3.0.6"
   },
   "devDependencies": {
+    "@storybook/addon-actions": "^3.4.10",
+    "@storybook/addon-console": "1.0.4",
```
kumar303: We use a caret to keep dev dependencies loose (this cuts down on greenkeeper spam). Any reason why everything you added can't have caret versions, e.g. `"^1.0.4"`?

stories/Button.js
```diff
@@ -0,0 +1,92 @@
...
+				cancel: 'cancel',
+				confirm: 'confirm',
+				alert: 'alert',
+				none: 'none',
+			};
+
+			const buttonType = select(
+				'buttonType',
+				btnTypeOptions,
+				'none',
```
kumar303: Is this setting the default to `none`? I think `neutral` would be a better default because it's more common. I'm not sure why `none` is the default type for `Button`.

.storybook/App.override.styles.scss
```diff
@@ -0,0 +1,7 @@
+html,
+body {
+	background: #fff;
+	// Giving extra space at top so does get covered
```
kumar303: Did you mean *does not* ?

rebmullin:
hi @kumar303, yeh out of habit I was installing pkgs using npm. Maybe that did it...?
I will remove this now : )

.storybook/webpack.config.babel.js
```diff
@@ -0,0 +1,19 @@
...
+import config from 'config';
+import fs from 'fs';
+import { getPlugins, getRules } from '../webpack-common';
+import { babelOptions } from '../webpack.dev.config.babel';
+
+
+// TODO: See if any of the module options are overkill here
+// or if there is anything we need to change.
+module.exports = {
+  module: {
```
kumar303: To future proof this, you should probably spread the rest of `module`:

```js
import webpackDevConfig, { babelOptions } from '../webpack.dev.config.babel';

module.exports = {
  module: {
    ...webpackDevConfig.module,
    rules: getRules({ babelOptions, bundleStylesWithJs: true }),
  },
};
```

stories/Button.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { MemoryRouter } from 'react-router';
+
+import { storiesOf, addDecorator } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { linkTo } from '@storybook/addon-links';
+import { withInfo } from '@storybook/addon-info';
+import { text, boolean, select } from '@storybook/addon-knobs';
+
+import Provider from '../.storybook/Provider';
```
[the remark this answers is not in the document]
kumar303: ... or move `.storybook` to `stories/setup`?

bobsilverberg:
@rebmullin I haven't looked closely at the code, but I tried it out locally and just wanted to say it's very cool. Thanks for putting the effort into making this happen!

stories/Button.js
rebmullin: yeh, I was just mimicking what the component shows as the default. is this ok given that? or still update that here?

.storybook/App.override.styles.scss
rebmullin: ha..yes : )

package.json
rebmullin: oops...this was unintentional. They npm module usually gives a snippet of code and I just copied that and didn't realize. I should be able to update that easily.

package.json
rebmullin: Yes it is supposedly supposed to fix this issue but I haven't had success yet:/ There is 1 module (I think) that I can't get to resolve: core/polyfill. I think if I don't piggyback off the current setup (which imports this) and @willdurand  suggested anyway, I may have more success. I'm going to circle back to this one...

.storybook/config.js
rebmullin: yeh i think so..i really liked this one too : )

.storybook/addons.js
```diff
@@ -0,0 +1,15 @@
...
+// import '@storybook/addon-links/register';
+
+const devices = devicesJSON.extensions.map(({ device }) => ({
+  uid: device.title,
+  title: device.title,
+  width: device.screen.vertical.width,
+  height: device.screen.vertical.height,
+}));
+
+registerScissors(devices);
```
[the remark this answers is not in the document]
rebmullin: yeh i thought so : )

.storybook/Provider.js
```diff
@@ -0,0 +1,34 @@
...
+// TODO: look into why stote/state isn't seen
+// by Button for example when you use the "to" prop:
+// it complains about the following (missing) clientApp and lang props :/
+// Do we need to add it to html and get it that way..? maybe preview-head.html?
+// The below only handles the Link url issue noted above.
+let initialState;
+const { store } = createStore({
+  history,
+  initialState: {
+    api: {
```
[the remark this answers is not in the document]
rebmullin: oh cool! this is exactly what I need. thank you : )

.storybook/webpack.config.babel.js
```diff
@@ -0,0 +1,19 @@
...
+  	rules: getRules({ babelOptions, bundleStylesWithJs: true }),
+  },
+
+  plugins : [
+  	...getPlugins(),
+  ],
+};
```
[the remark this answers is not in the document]
rebmullin: I have been kind of banging my head with this one... 
It seems that I can actually do this `import Button from 'ui/components/Button'` without resolve but I don't know why. I have to look into this more..

stories/Button.js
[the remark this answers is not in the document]
rebmullin: ok, sure thing ...I will look into this

rebmullin:
Hey @kumar303, thanks so much for the feedback : ) So I've been having issues with webpack 😕 with my first attempt at some of the changes you noted above but I will look into this more.

Also, I know you said you liked scissors and knobs but you're on the same page as @willdurand right? That these should be stripped out for this initial patch? 


thanks @bobsilverberg :)

stories/Button.js
rebmullin: I guess this might be moot.. since we are removing knobs (for now?)..

rebmullin:
> What you have seem to work though, it only requires a bit of cleanup and a simpler filename (webpack.config.js).

@willdurand the reason i set it up the way I did with the babel imported was b/c it wasn't recognizing import or any es6 unless I did it this way. Sass loader is easy to set up but the main reason why I was trying to import your settings was b/c of the import of core/logger that was causing an issue. I think the plugins that you had set up helped with that. I'm going to revisit this tho so I'll follow back up

rebmullin:
> I am not a fan of the knobs plugin because I think we should use Storybook like a living style guide and therefore write multiple stories for a component. I don't think we really need to update props, we should show examples.

the reason I like knobs is specifically for a component like Button where there are so many permutations. I guess right now I will show each example as you suggest and we maybe can revisit this after percy's set up.

stories/Button.js
```diff
@@ -0,0 +1,184 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import Button from 'ui/components/Button';
+
+const label = 'Hello Button';
+
+storiesOf('Button', module)
+	.add('default (no buttonType)', () => <Button>{label}</Button>)
```
rebmullin: I left out href and top and noLink props as they didn't seem to have an effect on the look. Let me know if you think these still should be added.

package.json
```diff
@@ -41,7 +41,9 @@
     "test-coverage-once": "bin/config-check.js && better-npm-run jest --coverage",
     "test-once": "bin/config-check.js && better-npm-run jest && npm run lint",
     "version-check": "bin/version-check.js",
-    "webpack-dev-server": "npm run build-locales && better-npm-run webpack-dev-server"
+    "webpack-dev-server": "npm run build-locales && better-npm-run webpack-dev-server",
+    "storybook": "cross-env NODE_PATH=src start-storybook -p 9001 -c stories/setup",
```
rebmullin: I could not get this working without NODE_PATH :/ I also saw that it was good idea include this package if we are going to use this...

stories/setup/webpack.config.js
```diff
@@ -0,0 +1,3 @@
+require('babel-register');
```
rebmullin: The main reason I am trying to use babel here is so I can import getClientConfig. I can't seem to use any of those utils files/functions if it is not done this way (without having an es5 version).

rebmullin:
hey @kumar303, do you want to take another look at this one? I've made most of the updates you and @willdurand suggested I think or have comments on why..

kumar303:
Can you fix up the conflicts when you get a chance?

stories/Button.js
```diff
@@ -0,0 +1,184 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import Button from 'ui/components/Button';
+
+const label = 'Hello Button';
+
+storiesOf('Button', module)
+  .add('default (no buttonType)', () => <Button>{label}</Button>)
```
kumar303: We can definitely iterate on this but at first glance, I don't think it will be useful to have one component style per page. I'd like to see many component styles on one page so that I can make a change to the code and see it all auto-update in front of me.

How about making one page per button type and showing each variation like this?

```diff
diff --git a/stories/Button.js b/stories/Button.js
index 0fa64a1c4..c61bf0796 100644
--- a/stories/Button.js
+++ b/stories/Button.js
@@ -6,6 +6,28 @@ import Button from 'ui/components/Button';
 const label = 'Hello Button';
 
 storiesOf('Button', module)
+  .add('action', () => {
+    return (
+      <dl className="stories-list">
+        <dt>Default</dt>
+        <dd>
+          <Button buttonType="action">{label}</Button>
+        </dd>
+        <dt>disabled=true</dt>
+        <dd>
+          <Button buttonType="action" disabled>{label}</Button>
+        </dd>
+        <dt>puffy=true</dt>
+        <dd>
+          <Button buttonType="action" puffy>{label}</Button>
+        </dd>
+        <dt>micro=true</dt>
+        <dd>
+          <Button buttonType="action" micro>{label}</Button>
+        </dd>
+      </dl>
+    );
+  })
   .add('default (no buttonType)', () => <Button>{label}</Button>)
   .add('default (no buttonType), disabled', () => (
     <Button disabled>{label}</Button>
diff --git a/stories/setup/App.override.styles.scss b/stories/setup/App.override.styles.scss
index 8b948128a..cdccdc5f1 100644
--- a/stories/setup/App.override.styles.scss
+++ b/stories/setup/App.override.styles.scss
@@ -1,6 +1,22 @@
+@import '~photon-colors/photon-colors';
+
 html,
 body {
   background: #fff;
   // Giving extra cushion so nothing gets covered.
   padding: 20px 10px 10px;
 }
+
+.stories-list {
+  display: grid;
+  grid-spacing: 0;
+  grid-template-columns: 1fr 1fr;
+
+  dd, dt {
+    align-items: center;
+    border-bottom: 1px solid $grey-30;
+    display: flex;
+    margin: 0;
+    padding: 6px;
+  }
+}
```

screenshot 2018-09-05 10 37 01

stories/Button.js
```diff
@@ -0,0 +1,184 @@
+import React from 'react';
```
kumar303: Let's use Flow for all stories. This will be crucial in making sure the story code doesn't bit rot as the app evolves.

```diff
diff --git a/package.json b/package.json
index 49cbb374c..b79219d42 100644
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "eslint": "eslint .",
     "flow": "flow",
     "flow:check": "flow check",
-    "flow:dev": "flow stop; flow start; chokidar .flowconfig flow/ src/ tests/ -i flow/logs/flow.log -c 'flow status' --initial",
+    "flow:dev": "flow stop; flow start; chokidar .flowconfig flow/ stories/ src/ tests/ -i flow/logs/flow.log -c 'flow status' --initial",
     "stylelint": "stylelint --syntax scss **/*.scss",
     "lint": "npm run eslint && npm run stylelint",
     "prettier": "prettier --write '**'",
diff --git a/stories/Badge.js b/stories/Badge.js
index 51863dfe0..493b02b42 100644
--- a/stories/Badge.js
+++ b/stories/Badge.js
@@ -1,3 +1,4 @@
+/* @flow */
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 
diff --git a/stories/Button.js b/stories/Button.js
index 0fa64a1c4..a1a49d80c 100644
--- a/stories/Button.js
+++ b/stories/Button.js
@@ -1,3 +1,4 @@
+/* @flow */
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 
```

stories/index.js
```diff
@@ -0,0 +1,10 @@
+// Add global styles
+import 'amo/components/App/styles.scss';
+import 'core/css/inc/lib.scss';
+
+// These styles override some App styles that make it hard to see
+// in storybook.
+import './setup/App.override.styles.scss';
+
+import './Badge';
```
kumar303: could you move this to `./ui/Badge`? I think that will help us scale up all the components we will need to add.

stories/index.js
```diff
@@ -0,0 +1,10 @@
...
+import './Button';
```
kumar303: Similarly, move this to `./core/Button` to follow its own path.

stories/setup/App.override.styles.scss
```diff
@@ -0,0 +1,6 @@
+html,
+body {
+  background: #fff;
+  // Giving extra cushion so nothing gets covered.
+  padding: 20px 10px 10px;
```
kumar303: I think it would be nice to follow our app specs to make everything look consistent. Following the specs would mean using pixel values divisible by 2 or 4, etc:

```
padding: 24px 12px 12px;
```

stories/Button.js
rebmullin: yep, I like this. Before I was playing with knobs, I had something similar (well..all one way page) but I think this way is better. I will work on this..

stories/Button.js
rebmullin: good idea 👍  I will add!

stories/setup/App.override.styles.scss
rebmullin: oh?! sure thing

package.json
```diff
@@ -42,7 +42,9 @@
...
```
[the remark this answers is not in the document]
rebmullin: ahh! missed that, thanks..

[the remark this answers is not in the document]
rebmullin: ahh... ok sure thing.

rebmullin:
conflicts again :/  
I get this fixed..

stories/index.js
rebmullin: hi @kumar303, can you clarify what you mean here? Do you want the stories to live actually in the `ui/Badge` component:

```javascript
storiesOf('Badge', module)
  .add('default (no type)', () => <Badge label={label} />)
  .add('experimental type', () => <Badge type="experimental" label={label} />)
  .add('featured type', () => <Badge type="featured" label={label} />)
  .add('restart-required type', () => (
    <Badge type="restart-required" label={label} />
  ))
...
```

stories/index.js
rebmullin: same thing ...do you mind clarifying please? :)

stories/Button.js
willdurand: Instead of manually doing this, can we explore storybook addons? There are at least two addons for that:

- https://github.com/Checkfront/react-storybook-addon-chapters
- https://github.com/evgenykochetkov/react-storybook-addon-props-combinations

That's also why I initially requested a smaller patch with the bare minimum setup.. So that we can iterate over it, and actually ship things. FWIW, I have worked with Storybook in the past.

rebmullin: ok so..I'll hold off on this piece for now? We can explore if chapters will do something similar in a follow up PR? will that work?

kumar303: > That's also why I initially requested a smaller patch with the bare minimum setup

I have been requesting changes because I didn't think the first pass was set up to meet our needs.

> Instead of manually doing this, can we explore storybook addons?

This sounds promising. To keep it moving, I suggest doing it in a follow-up issue.

stories/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+	"rules": {
```
kumar303: This indent is off

rebmullin: hmm I thought I fixed that but I guess not 🤷‍♀️  will try again : )

PR #6062, opened 2018-08-24
Reducer/saga/api code to delete a review

### LINKED ISSUE #6054: Reducer/Saga/API code to delete a review
To support #5495 and #3075, we need reducer/saga/api support for deleting a review.

bobsilverberg:
Fixes #6054

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -39,8 +39,7 @@ export const SEND_REPLY_TO_REVIEW: 'SEND_REPLY_TO_REVIEW' =
 export const SEND_REVIEW_FLAG: 'SEND_REVIEW_FLAG' = 'SEND_REVIEW_FLAG';
 export const UPDATE_ADDON_REVIEW: 'UPDATE_ADDON_REVIEW' = 'UPDATE_ADDON_REVIEW';
 export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
-export const CLEAR_ADDON_REVIEWS_FOR_REVIEWID: 'CLEAR_ADDON_REVIEWS_FOR_REVIEWID' =
-  'CLEAR_ADDON_REVIEWS_FOR_REVIEWID';
+export const CLEAR_ADDON_REVIEWS: 'CLEAR_ADDON_REVIEWS' = 'CLEAR_ADDON_REVIEWS';
 
 export type UserReviewType = {|
   addonId: number,
@@ -632,21 +631,21 @@ export const deleteAddonReview = ({
   };
 };
 
-type ClearAddonReviewsForReviewIdParams = {|
+type ClearAddonReviewsParams = {|
   reviewId: number,
 |};
 
-export type ClearAddonReviewsForReviewIdAction = {|
-  type: typeof CLEAR_ADDON_REVIEWS_FOR_REVIEWID,
-  payload: ClearAddonReviewsForReviewIdParams,
+export type ClearAddonReviewsAction = {|
+  type: typeof CLEAR_ADDON_REVIEWS,
+  payload: ClearAddonReviewsParams,
 |};
 
-export const clearAddonReviewsForReviewId = ({
+export const clearAddonReviews = ({
   reviewId,
-}: ClearAddonReviewsForReviewIdParams): ClearAddonReviewsForReviewIdAction => {
+}: ClearAddonReviewsParams): ClearAddonReviewsAction => {
   invariant(reviewId, 'reviewId is required');
   return {
-    type: CLEAR_ADDON_REVIEWS_FOR_REVIEWID,
+    type: CLEAR_ADDON_REVIEWS,
     payload: { reviewId },
   };
 };
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -39,7 +39,8 @@ export const SEND_REPLY_TO_REVIEW: 'SEND_REPLY_TO_REVIEW' =
 export const SEND_REVIEW_FLAG: 'SEND_REVIEW_FLAG' = 'SEND_REVIEW_FLAG';
 export const UPDATE_ADDON_REVIEW: 'UPDATE_ADDON_REVIEW' = 'UPDATE_ADDON_REVIEW';
 export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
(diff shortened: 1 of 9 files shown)


src/amo/api/reviews.js
```diff
@@ -262,3 +263,28 @@ export const flagReview = ({
...
+  reviewId: number,
+|};
+
+export const deleteReview = ({
+  apiState,
+  errorHandler,
+  reviewId,
+}: DeleteReviewParams = {}): Promise<void> => {
+  return new Promise((resolve) => {
+    invariant(reviewId, 'The reviewId parameter is required');
```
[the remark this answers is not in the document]
bobsilverberg: I agree, consistency would be good. I guess, because these are going to eventually be stripped out for production, we could go with short and sweet. How about simply "xxx is required", i.e., in this case "reviewId is required"?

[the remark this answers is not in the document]
willdurand: I like "xxx is required"

src/amo/actions/reviews.js
```diff
@@ -39,6 +38,9 @@ export const SEND_REPLY_TO_REVIEW: 'SEND_REPLY_TO_REVIEW' =
   'SEND_REPLY_TO_REVIEW';
 export const SEND_REVIEW_FLAG: 'SEND_REVIEW_FLAG' = 'SEND_REVIEW_FLAG';
 export const UPDATE_ADDON_REVIEW: 'UPDATE_ADDON_REVIEW' = 'UPDATE_ADDON_REVIEW';
+export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
+export const CLEAR_ADDON_REVIEWS_FOR_REVIEWID: 'CLEAR_ADDON_REVIEWS_FOR_REVIEWID' =
```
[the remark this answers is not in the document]
bobsilverberg: I agree with changing `CLEAR` to `UNLOAD`. I guess I went with `CLEAR` because I was replacing another action that was already called `CLEAR`. That's also part of the reason I used `CLEAR_ADDON_REVIEWS_FOR_REVIEWID`, because originally I wasn't sure that we could just remove `CLEAR_ADDON_REVIEWS` so I needed the name to be different.

I don't think `UNLOAD_ADDON_REVIEW` makes sense though, because what we're doing is not unloading an addon review; what we're doing is unloading any caches of reviews which contained this review, which I think is a bit different. Maybe just `UNLOAD_ADDON_REVIEWS` would be okay for that, but I kind of like the `FOR_REVIEWID` part because it makes it more explicit exactly what we're doing. Either works for me though. What do you think?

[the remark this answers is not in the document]
willdurand: I'd go for `UNLOAD_ADDON_REVIEWS` then :)

bobsilverberg:
Thanks for the review @willdurand. This is ready for another look.

src/amo/actions/reviews.js
```diff
@@ -39,6 +38,8 @@ export const SEND_REPLY_TO_REVIEW: 'SEND_REPLY_TO_REVIEW' =
   'SEND_REPLY_TO_REVIEW';
 export const SEND_REVIEW_FLAG: 'SEND_REVIEW_FLAG' = 'SEND_REVIEW_FLAG';
 export const UPDATE_ADDON_REVIEW: 'UPDATE_ADDON_REVIEW' = 'UPDATE_ADDON_REVIEW';
+export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
+export const CLEAR_ADDON_REVIEWS: 'CLEAR_ADDON_REVIEWS' = 'CLEAR_ADDON_REVIEWS';
```
willdurand: so, no `UNLOAD`?

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -296,43 +296,124 @@ describe(__filename, () => {
...
+          slug: addonSlug,
+        },
+        user: { id: userId },
+      };
+
+      let state;
+
+      // Initialize values into the byId, byAddon, byUserId and groupedRatings buckets.
+      state = reviewsReducer(startState, setReview(review));
+      expect(state.byId[reviewId].addonId).toEqual(addonId);
```
willdurand: I know it is convenient to add expectations in a function helper, but then it gets executed all the time and it is not easy to see when it breaks (since we usually find expectations in the test cases, not in helper functions). I think we should either not have expectations or they should be in the test cases.

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -296,43 +296,124 @@ describe(__filename, () => {
...
+      state = loadReviewDataIntoState({
+        startState: state,
+        addonId: addonId2,
+        addonSlug: addonSlug2,
+        grouping,
+        reviewId: reviewId2,
+        userId: userId2,
+      });
+
+      state = reviewsReducer(state, clearAddonReviews({ reviewId }));
```
willdurand: Adding expectations before this call to define preconditions would make this test case more readable because one would directly see the impact of `clearAddonReviews`

src/amo/actions/reviews.js
bobsilverberg: Oops, I forgot that part. :-P

bobsilverberg:
Thanks again @willdurand. I'm not sure if the changes I've made are exactly what you're looking for. Please take another look and let me know.

src/amo/components/EditableCollectionAddon/index.js
```diff
@@ -202,7 +202,7 @@ const EditableCollectionAddon: React.ComponentType<Props> = compose(
     fileName: __filename,
     extractId,
     initialState: initialUIState,
-    resetOnUnmount: true,
+    resetOnUnmount: false,
```
willdurand: I don't think we need this, right?

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -295,44 +295,135 @@ describe(__filename, () => {
...
+
+      const verifyUnrelatedData = () => {
+        expect(state.byId[reviewId2].addonId).toEqual(addonId2);
+        expect(state.byAddon[addonSlug2].reviews).toEqual([reviewId2]);
+        expect(state.byUserId[userId2].reviews).toEqual([reviewId2]);
+        expect(state.groupedRatings[addonId2]).toEqual(grouping);
+      };
+
+      // Verify that the unrelated data has been loaded for the reviewId.
+      verifyUnrelatedData();
```
willdurand: maybe it is not worth adding a function here, it is used only twice.

src/amo/components/EditableCollectionAddon/index.js
bobsilverberg: Oops, again!

Issue #5359, opened 2018-06-21
clean up / remove any lightweight theme logic once migration is ready


rebmullin:
### Describe the problem and steps to reproduce it:

There is code specific to lightweight themes which I believe can be removed once the migration is ready:

there is some logic that defines valid types, which includes persona. We should be able to remove and replace with statictheme. 

remove all the themedata set up that lightweight themes is using

update the search filter

other stuff?

---

**For QA:** please make sure all pages with static themes behave as before. There should be no obvious regression. In addition, here is a list of "known changes":

- [search result pages with "recommended" themes](https://addons-dev.allizom.org/en-US/firefox/search/?platform=mac&recommended=true&type=statictheme) should have a new HTML title ("Recommended themes" instead of "Recommended add-ons")
- the contribution card on [a static theme detail page](https://addons.mozilla.org/en-US/firefox/addon/matte-black-v2/) now shows "Support this artist" rather than "Support this author" (bug fix, yay)

kumar303:
> there is some logic that defines valid types, which includes persona.

Just FYI, this logic has been intertwined in surprising ways over the history of the code base so removing it may not be so trivial.

kumar303:
See https://github.com/mozilla/addons-frontend/pull/7627 for some initial removal

diox:
The migration happened some time ago, can we make some progress on this now ? We assumed we could remove most of the personas-related code in addons-server, and it caused https://github.com/mozilla/addons-frontend/issues/8480, which forced us to add back a hack pretending some lightweight theme categories still exist.

muffinresearch:
This will need further investigation, but in looking at this today, It might be worth considering if it would be easier to change the `ADDON_TYPE_THEME` from `persona` to `statictheme` and remove the `ADDON_TYPE_STATIC_THEME` instead of the other way around.

diox:
I suspect a few of these occurrences are hack to work around the subtleties of personas though. I think it's worth double-checking each.

willdurand:
> - the contribution card now shows "Support this artist" rather than "Support this developer" (bug fix, yay)

@jvillalobos I am not sure whether we have static themes with "contribution cards", but I think we should had have that when we introduced the static themes. Thoughts?

willdurand:
> We do support contributions for themes. [Here are the ones](https://sql.telemetry.mozilla.org/queries/65715/source#166931) that have it enabled, and I see the button in their page.

[...]

PR #6069, opened 2018-08-26
Display nothing when there is no addon description

### LINKED ISSUE #6058: Don't repeat Summary on product page if there's no Description
Add-ons can have a short Summary and a longer Description, both of which appear in the product page. If the Description is missing, we currently duplicate the Summary and show that in place of a Description. Instead of doing that, the Description section should not show up at all. 

It looks like this was fixed for themes already, in https://github.com/mozilla/addons-frontend/issues/5364#issuecomment-404312967

SeanPrashad:
Fixes #6058 - When there is no description present for an add-on, show nothing instead of showing an exact duplicate of the add-ons summary.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -278,9 +278,8 @@ export class AddonBase extends React.Component {
       // For any theme type, we want to hide the summary text here since that is
       // already displayed in the header.
       showAbout =
-        ((description === addon.summary &&
-          addon.type === ADDON_TYPE_STATIC_THEME) ||
-          addon.type === ADDON_TYPE_THEME) === false;
+        (description === addon.summary || addon.type === ADDON_TYPE_THEME) ===
+        false;
 
       if (!description || !description.length) {
         return null;
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -187,6 +187,10 @@
       // Fixes screenshot overflow in FF 52 ESR.
       max-width: 100%;
       overflow-x: hidden;
+
+      .Card:first-child {
+        margin-top: 0;
+      }
     }
 
     .AddonDescription-contents {
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -784,6 +784,76 @@ describe(__filename, () => {
     expect(root.find('.AddonDescription')).toHaveLength(0);
   });
 
+  it('does not display anything when the lightweight theme has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_THEME,
+        summary,
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it('does not display anything when the generic add-on has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_COMPLETE_THEME,
+        summary,
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it('does not display anything when the search plugin has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_OPENSEARCH,
+        summary,
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it('does not display anything when the language pack has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_LANG,
+        summary,
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it('does not display anything when the dictionary has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_DICT,
+        summary,
+        description: null,
+      }),
+    });
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
   it("displays the extension's description when both description and summary are supplied", () => {
     const description = 'some cool description';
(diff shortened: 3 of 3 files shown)


src/amo/components/Addon/index.js
```diff
@@ -271,7 +271,7 @@ export class AddonBase extends React.Component {
           title = i18n.gettext('About this add-on');
       }
 
-      const description = addon.description ? addon.description : addon.summary;
+      const description = addon.description ? addon.description : null;
```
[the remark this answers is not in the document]
SeanPrashad: Just curious @willdurand - what was the git command you ran to do the diff? Or did you edit the files locally and run the command to show me what to do? 😆

SeanPrashad:
Thanks for the quick feedback! I've slapped in the changes as requested but I'm wondering if the CSS rule I've added is in the correct location. I've tested it on 2-3 add-ons (with only the recommended add-ons present) and everything is looking good so far:

margin-top

src/amo/components/Addon/styles.scss
```diff
@@ -185,6 +185,10 @@
       // Fixes screenshot overflow in FF 52 ESR.
       max-width: 100%;
       overflow-x: hidden;
+
+      .Card {
```
willdurand: not sure this is safe, I'd go for a more restrictive rule, like `.Card:first-child` or something similar.

src/amo/components/Addon/index.js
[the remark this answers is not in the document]
willdurand: I checked out your PR with `git checkout` (_via_ [hub](https://github.com/github/hub)), then I edited the files locally and ran `git diff`. Actually, I ran `git diff | pbcopy` to copy the content of `git diff` in the clipboard to be able to paste it in a comment here.

src/amo/components/Addon/index.js
```diff
@@ -271,28 +270,23 @@ export class AddonBase extends React.Component {
...
-      const description = addon.description ? addon.description : addon.summary;
-      // For any theme type, we want to hide the summary text here since that is
-      // already displayed in the header.
-      showAbout =
-        ((description === addon.summary &&
-          addon.type === ADDON_TYPE_STATIC_THEME) ||
-          addon.type === ADDON_TYPE_THEME) === false;
+      const { description } = addon;
 
       if (!description || !description.length) {
```
willdurand: Maybe we should also check that `description !== summary`, no? I am thinking about that because of the test case "does not display a lightweight theme's summary" you changed.

SeanPrashad: Yeah I was thinking about why that test had changed.. I'll try it out later today when I get home from work!

SeanPrashad: I tried adding `description !== summary` but 12 other tests broke and my gut instinct told me that wasn't probably the right route. Instead, I update that single test since `description` wasn't passed in at all - what are your thoughts?

src/amo/components/Addon/index.js
```diff
@@ -274,28 +273,23 @@ export class AddonBase extends React.Component {
           title = i18n.gettext('About this add-on');
       }
 
-      const description = addon.description ? addon.description : addon.summary;
-      // For any theme type, we want to hide the summary text here since that is
-      // already displayed in the header.
-      showAbout =
-        ((description === addon.summary &&
-          addon.type === ADDON_TYPE_STATIC_THEME) ||
-          addon.type === ADDON_TYPE_THEME) === false;
```
rebmullin: Hey @SeanPrashad, thanks for helping with this one : ) 

So I think we still need something like we had above. I know it gets a little confusing b/c the different types have different fields. This should be the breakdown:

Extensions,
Static themes:

Description - yes (optional)
Summary - yes (required on both)
-----

Lwt themes:
Description - yes (optional)
Summary - always null

You can inspect this in the redux tool too if that helps.


so I *think* you can just revert a lot of what was here and just remove static theme piece ~:

```javascript

showAbout =
        (description === addon.summary || addon.type === ADDON_TYPE_THEME) ===
        false;
...
```

Basically for a lightweight theme we never want to show that bottom card (it will never have new/different text). For other types we just need to see if it's actually using the `addon.description` text. If the "description" here is set to the `addon.description` we know the addon has both fields (since summary is required) which means we should display both texts. If the "description" is set to the `addon.summary` then we can hide in the bottom card since since we are already displaying this text in the header. 

note: I'm just thinking it might be a *little* less confusing if we renamed `const description = ...` to `const text=...` since it could really either be addon.description or addon.summary?


Does that help? or let me know if any questions :)

SeanPrashad: Thanks for the super detailed explanation @rebmullin! It makes more sense now :)

tests/unit/amo/components/TestAddon.js
```diff
@@ -781,11 +781,8 @@ describe(__filename, () => {
       }),
     });
 
-    expect(root.find('.AddonDescription')).toHaveLength(1);
-
-    expect(root.find('.AddonDescription-contents')).toHaveHTML(
-      `<div class="AddonDescription-contents">${summary}</div>`,
-    );
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+    expect(root.find('.AddonDescription-contents')).toHaveLength(0);
```
rebmullin: Thanks for the updates @SeanPrashad : ) It's looking good. 

Just a few more notes here:

So I don't think we need both of these "expects"; If `AddonDescription` is 0 then `AddonDescription-contents` will be 0 too. I think initially we had the extra check here b/c we were checking what the text was.

We also need to update the test case's description here. I believe it's outdated now. You can probably put something similar to what the static theme test case shows (I believe it's right above this one).

Lastly, we should have similar test cases now for the other add-on types too. Like language, dictionary, search..etc.

tests/unit/amo/components/TestAddon.js
```diff
@@ -781,11 +781,77 @@ describe(__filename, () => {
...
+
+    expect(root.find('.AddonDescription')).toHaveLength(0);
+  });
+
+  it('does not display anything when the generic add-on has no description', () => {
+    const summary = 'my theme is very cool';
+    const root = shallowRender({
+      addon: createInternalAddon({
+        ...fakeAddon,
+        type: ADDON_TYPE_COMPLETE_THEME,
```
rebmullin: hmm..I have never really seen this type before : /

We can just update the description so it matches: "..when the complete theme has no description"?

SeanPrashad:
@rebmullin I've updated that test description as requested. For the list of all other add-on types, I was looking at the different types found [here](https://github.com/mozilla/addons-frontend/blob/master/tests/unit/amo/components/TestAddon.js#L38) - should we get a second opinion on the new additions?

rebmullin:
@SeanPrashad, yep, thanks :) I did see that before but I never really saw that being used I guess so it wasn't on my radar :/  It may be obsolete (?) but I guess we can worry about that [here](#5359) potentially..

I think this is looking 👍 but sure..we can see what @willdurand says too :)

PR #6073, opened 2018-08-27
Tidy/simplify SCSS imports

### LINKED ISSUE #6097: Simplify SCSS imports
It is super difficult to find the right imports all the time, so we end copy/pasting imports from another file and tweak them if that does not work. This leads to weird import statements in some files.

By having only one file to import, it makes things easier. Each SCSS file will now use **only one import**, one of:

```
@import '~amo/css/styles';
// or
@import '~core/css/styles';
// or
@import '~disco/css/styles';
// or
@import '~ui/css/styles';
```

In addition, no app-specific style should leak in `core` or another app.

For QA: please make sure there is no visual regression on both AMO and Disco Pane, all pages. By visual regression, I mean no blank page or missing style. I don't think we'll get very small regressions, it should be obvious.

willdurand:
Fix #6097

---

This PR cleans the different SCSS imports to avoid "amo" SCSS to leak in
`src/ui` or `src/core` so that "disco" can also use the components in
those directories. Also remove a few unused variables.

In addition, I created one file to import for each app, instead of having
different imports, not always sorted correctly. (I did not create a unique
file for the "core" components yet, though).

Why? It is super difficult to find the right imports all the time, so we
end copy/pasting imports from another file and tweak them if that does not
work. This leads to weird import statements in some files. By having only
one file to import, it makes things easier.

Why (2)? I reviewed a PR and saw some code in disco that was coming from
amo, so I decided to clean up our SCSS a bit.

I built the amo app with `master` and this PR and did not notice any size
difference. I did not see any visual regression either.

### DIFF
--- a/src/amo/components/App/styles.scss
+++ b/src/amo/components/App/styles.scss
@@ -1,4 +1,5 @@
 @import '~normalize.css';
+@import '~core/css/inc/lib';
 @import '~amo/css/styles';
 
 html,
--- a/src/amo/components/App/styles.scss
+++ b/src/amo/components/App/styles.scss
@@ -1,4 +1,4 @@
-@import '~normalize.css';
+@import '~normalize';
 @import '~core/css/inc/lib';
 @import '~amo/css/styles';
 
--- a/src/amo/components/App/styles.scss
+++ b/src/amo/components/App/styles.scss
@@ -1,10 +1,5 @@
 @import '~normalize.css';
-@import '~photon-colors/photon-colors';
-@import '~core/css/inc/vars';
-@import '~core/css/inc/lib';
-@import '~core/css/inc/mixins';
-@import '~amo/css/inc/vars';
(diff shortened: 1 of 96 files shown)


src/amo/components/App/styles.scss
```diff
@@ -1,10 +1,6 @@
 @import '~normalize.css';
```
kumar303: At some point we need to fix this.

```
[0] DEPRECATION WARNING on line 1, column 8 of stdin:
[0] Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
```

willdurand: ah, interesting.

willdurand: more info at: https://github.com/sass/node-sass/issues/2362#issuecomment-388634848

willdurand:
(builds were green on travis)

PR #6074, opened 2018-08-27
Fix comments for react-onclickoutside

### LINKED ISSUE #5879: react-onclickoutside breaks shallowUntilTarget
For some reasons, the `react-onclickoutside` HOC does not work with `shallowUntilTarget()` unless we disable the lifecycle methods:

```js
  const renderComponent = (componentInstance, target) => {
    return shallowUntilTarget(componentInstance, target, {
      // TODO: ideally, we would like to enable the lifecycle methods, but it
      // produces unexpected errors, related to Enzyme 3.
      // See: http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#lifecycle-methods.
      shallowOptions: { disableLifecycleMethods: true },
    });
  };
```

We have this code in the `DropdownMenu` component and I have added it to the `InfoDialog` component too. We should investigate to see what the root cause is and potentially fix the issue to avoid disabling the lifecycle methods.

If that does not work at all, well... let's use this issue as documentation.

willdurand:
Close #5879

### DIFF
--- a/tests/unit/core/components/TestInfoDialog.js
+++ b/tests/unit/core/components/TestInfoDialog.js
@@ -14,12 +14,9 @@ describe(__filename, () => {
       ...props,
     };
 
-    // This is needed because of `react-onclickoutside`, see:
-    // https://github.com/mozilla/addons-frontend/issues/5879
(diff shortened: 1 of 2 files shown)


willdurand:
Merging as it is only a comment change.

PR #6075, opened 2018-08-27
Greenkeeper/monorepo.enzyme 1.3.0

willdurand:
Looks like it fails, see: https://github.com/airbnb/enzyme/issues/1795


willdurand:
Fixed by #6174

PR #6083, opened 2018-08-27
I6082 ratings overlay style adjustment

### LINKED ISSUE #6082: When overlay is open to rate an addon, the stars styles are now above overlay

got to an addon detail page and click on a star to rate. 
You see stars above overlay 

### What happened?

You see stars above overlay and even modal if screen is small enough:

Alt text

It could be a regression from PR #6056 (maybe?) and something just needs to slightly adjusted to account for this.

rebmullin:
fixes #6082

### DIFF
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -13,7 +13,7 @@
     padding: 0 10px;
 
     @include respond-to(large) {
-      min-width: 768px;
+      min-width: 700px;
       padding-left: 200px;
       padding-right: 200px;
     }
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -13,8 +13,7 @@
     padding: 0 10px;
 
     @include respond-to(large) {
-      padding-left: 200px;
-      padding-right: 200px;
+      min-width: 520px;
     }
   }
 }
--- a/src/amo/components/AddonReview/styles.scss
+++ b/src/amo/components/AddonReview/styles.scss
@@ -13,7 +13,9 @@
     padding: 0 10px;
 
     @include respond-to(large) {
-      min-width: 520px;
+      min-width: 768px;
+      padding-left: 200px;
+      padding-right: 200px;
     }
   }
 }
--- a/src/ui/components/Overlay/styles.scss
+++ b/src/ui/components/Overlay/styles.scss
@@ -1,5 +1,11 @@
 @import '~photon-colors/photon-colors';
 
+$z-index: (
+  overlay: 1000,
+  background: 1001,
+  contents: 1002,
+);
+
 .Overlay {
   display: none;
 }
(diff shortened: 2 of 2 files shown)


bobsilverberg:
@rebmullin Do you mind adding before and after screenshots for this one?

rebmullin:
## BEFORE (currently on Dev but not Stage):
Alt text

## AFTER (should be similar/same as Stage now):
Alt text

src/amo/components/AddonReview/styles.scss
```diff
@@ -13,6 +13,7 @@
     padding: 0 10px;
 
     @include respond-to(large) {
+      min-width: 768px;
```
willdurand: What happens between 720px and 768px?

src/ui/components/Overlay/styles.scss
```diff
@@ -13,6 +13,12 @@
   position: fixed;
   top: 0;
   width: 100%;
+  z-index: 2;
```
willdurand: That looks a bit weird because the z-index for `.Overlay-contents` is at `1001`

src/ui/components/Overlay/styles.scss
```diff
@@ -13,6 +13,12 @@
...
+
+  + form {
```
willdurand: That looks like it won't fix everything. It works only because there is a form in the other component, no? Also the use of `+` is usually the last option in CSS.

willdurand:
screen shot 2018-08-27 at 21 14 44

kumar303:
You could also use a sass variable to track the `z-index` property

src/ui/components/Overlay/styles.scss
rebmullin: yeh but that's a different level in the DOM hierarchy so it's different..

rebmullin:
thanks for everyone's feedback : )

@willdurand, how's it look now tho?

src/ui/components/Overlay/styles.scss
```diff
@@ -1,5 +1,11 @@
 @import '~photon-colors/photon-colors';
 
+$z-index: (
```
willdurand: I think the idea was to have a variable that we can increment or that is the z-index of the overlay and then we do `$z-index + 1` and `$z-index + 2`.

rebmullin: ahh! I see..thanks

kumar303:
Thanks for taking care of this regression so fast, @rebmullin

src/ui/components/Overlay/styles.scss
```diff
@@ -1,5 +1,7 @@
 @import '~photon-colors/photon-colors';
 
+$z-index: 1000;
```
kumar303: We should probably create a mixin someday to [manage z-index across the whole site](https://www.smashingmagazine.com/2014/06/sassy-z-index-management-for-complex-layouts/). It's not a pressing problem yet.

PR #6084, opened 2018-08-27
Enable deleting a review from the review listing page

### LINKED ISSUE #3075: There is no way to delete own review
1. To go AMO mobile website and visit any add-ons details page. ( Example- https://addons.mozilla.org/en-US/firefox/addon/real-madrid-blue/ )
2. Write a review
### What happened?
There is option for the review owner to delete his/her review
There should be option for the review owner to delete his/her review like we have in the classic desktop website.
The following is my own written review. There is no option to delete it.
screenshot_2017-09-07-21-47-03-962_org mozilla fennec_aurora

bobsilverberg:
Fixes #3075 

This adds a "Delete my review" link to reviews that are deletable by the current user, and a "Delete my reply" link to replies that are deletable by the current user. Clicking the link deletes the review/reply immediately and refreshes the list of reviews.

I considered using the `ConfirmButton` as we've used for deleting a collection and a user profile, but I think that's probably unnecessary for deleting a review/reply.

A user's review:

screenshot 2018-08-27 14 23 49

A developer's response:

screenshot 2018-08-27 14 27 45

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -611,6 +611,7 @@ export const hideFlashedReviewMessage = (): HideFlashedReviewMessageAction => {
 
 type DeleteAddonReviewParams = {|
   errorHandlerId: string,
+  isReplyToReviewId?: number,
   reviewId: number,
 |};
 
@@ -621,6 +622,7 @@ export type DeleteAddonReviewAction = {|
 
 export const deleteAddonReview = ({
   errorHandlerId,
+  isReplyToReviewId,
   reviewId,
 }: DeleteAddonReviewParams) => {
   invariant(errorHandlerId, 'errorHandlerId is required');
@@ -628,7 +630,7 @@ export const deleteAddonReview = ({
 
   return {
     type: DELETE_ADDON_REVIEW,
-    payload: { errorHandlerId, reviewId },
+    payload: { errorHandlerId, isReplyToReviewId, reviewId },
   };
 };
 
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -260,7 +260,7 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 'AddonReviewListItem-control',
                 confirmButtonClassName,
               )}
-              id={confirmButtonClassName}
+              id={`${confirmButtonClassName}-${review.id}`}
               message={
                 isReply
                   ? i18n.gettext('Do you really want to delete this reply?')
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import makeClassName from 'classnames';
 import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
@@ -20,10 +21,11 @@ import {
   showEditReviewForm,
   showReplyToReviewForm,
 } from 'amo/actions/reviews';
+import ConfirmButton from 'ui/components/ConfirmButton';
+import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
 import UserReview from 'ui/components/UserReview';
-import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import type { UserReviewType } from 'amo/actions/reviews';
 import type { UserType } from 'amo/reducers/users';
 import type { AppState } from 'amo/store';
@@ -229,6 +231,8 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
       byLine = <LoadingText />;
     }
 
+    const confirmButtonClassName = 'AddonReviewListItem-delete';
+
     const controls = (
       <div className="AddonReviewListItem-allControls">
         {siteUser && review && review.userId === siteUser.id ? (
@@ -250,15 +254,24 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
-            <a
-              href="#delete"
-              onClick={this.onClickToDeleteReview}
-              className="AddonReviewListItem-delete AddonReviewListItem-control"
+            <ConfirmButton
+              buttonType="cancel"
+              className={makeClassName(
+                'AddonReviewListItem-control',
+                confirmButtonClassName,
+              )}
+              id={confirmButtonClassName}
+              message={
+                isReply
+                  ? i18n.gettext('Do you really want to delete this reply?')
+                  : i18n.gettext('Do you really want to delete this review?')
+              }
+              onConfirm={this.onClickToDeleteReview}
             >
               {isReply
                 ? i18n.gettext('Delete my reply')
                 : i18n.gettext('Delete my review')}
-            </a>
+            </ConfirmButton>
           </React.Fragment>
         ) : null}
 
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -256,10 +256,12 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
             </a>
             <ConfirmButton
               buttonType="cancel"
+              cancelButtonType="neutral"
               className={makeClassName(
                 'AddonReviewListItem-control',
                 confirmButtonClassName,
               )}
+              confirmButtonText={i18n.gettext('Delete')}
               id={`${confirmButtonClassName}-${review.id}`}
               message={
                 isReply
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -47,6 +47,7 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
+  deletingReview: boolean,
   dispatch: DispatchFunc,
   editingReview: boolean,
   errorHandler: ErrorHandlerType,
@@ -199,6 +200,7 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
   render() {
     const {
       addon,
+      deletingReview,
       editingReview,
       errorHandler,
       i18n,
@@ -254,26 +256,32 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
-            <ConfirmButton
-              buttonType="cancel"
-              cancelButtonType="neutral"
-              className={makeClassName(
-                'AddonReviewListItem-control',
-                confirmButtonClassName,
-              )}
-              confirmButtonText={i18n.gettext('Delete')}
-              id={`${confirmButtonClassName}-${review.id}`}
-              message={
-                isReply
-                  ? i18n.gettext('Do you really want to delete this reply?')
-                  : i18n.gettext('Do you really want to delete this review?')
-              }
-              onConfirm={this.onClickToDeleteReview}
-            >
-              {isReply
-                ? i18n.gettext('Delete my reply')
-                : i18n.gettext('Delete my review')}
-            </ConfirmButton>
+            {deletingReview ? (
+              <span className="AddonReviewListItem-control AddonReviewListItem-deleting">
+                {i18n.gettext('Deleting…')}
+              </span>
+            ) : (
+              <ConfirmButton
+                buttonType="cancel"
+                cancelButtonType="neutral"
+                className={makeClassName(
+                  'AddonReviewListItem-control',
+                  confirmButtonClassName,
+                )}
+                confirmButtonText={i18n.gettext('Delete')}
+                id={`${confirmButtonClassName}-${review.id}`}
+                message={
+                  isReply
+                    ? i18n.gettext('Do you really want to delete this reply?')
+                    : i18n.gettext('Do you really want to delete this review?')
+                }
+                onConfirm={this.onClickToDeleteReview}
+              >
+                {isReply
+                  ? i18n.gettext('Delete my reply')
+                  : i18n.gettext('Delete my review')}
+              </ConfirmButton>
+            )}
           </React.Fragment>
         ) : null}
 
@@ -323,18 +331,21 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
 }
 
 export function mapStateToProps(state: AppState, ownProps: Props) {
+  let deletingReview = false;
   let editingReview = false;
   let replyingToReview = false;
   let submittingReply = false;
   if (ownProps.review) {
     const view = state.reviews.view[ownProps.review.id];
     if (view) {
+      deletingReview = view.deletingReview;
       editingReview = view.editingReview;
       replyingToReview = view.replyingToReview;
       submittingReply = view.submittingReply;
     }
   }
   return {
+    deletingReview,
     editingReview,
     siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
     replyingToReview,
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import invariant from 'invariant';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
@@ -12,6 +13,7 @@ import log from 'core/logger';
 import { getCurrentUser, hasPermission } from 'amo/reducers/users';
 import { isAddonAuthor } from 'core/utils';
 import {
+  deleteAddonReview,
   hideEditReviewForm,
   hideReplyToReviewForm,
   sendReplyToReview,
@@ -54,6 +56,20 @@ type InternalProps = {|
 |};
 
 export class AddonReviewListItemBase extends React.Component<InternalProps> {
+  onClickToDeleteReview = (event: SyntheticEvent<HTMLElement>) => {
+    const { dispatch, errorHandler, isReplyToReviewId, review } = this.props;
+    event.preventDefault();
+
+    invariant(review, 'review is required');
+    dispatch(
+      deleteAddonReview({
+        errorHandlerId: errorHandler.id,
+        reviewId: review.id,
+        isReplyToReviewId,
+      }),
+    );
+  };
+
   onClickToEditReview = (event: SyntheticEvent<any>) => {
     const { dispatch, isReplyToReviewId, review } = this.props;
     event.preventDefault();
@@ -233,6 +249,15 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
               {isReply
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
+            </a>
+            <a
+              href="#delete"
+              onClick={this.onClickToDeleteReview}
+              className="AddonReviewListItem-delete AddonReviewListItem-control"
+            >
+              {isReply
+                ? i18n.gettext('Delete my reply')
+                : i18n.gettext('Delete my review')}
             </a>
           </React.Fragment>
         ) : null}
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -256,7 +256,7 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
-            {deletingReview ? (
+            {deletingReview && !errorHandler.hasError() ? (
               <span className="AddonReviewListItem-control AddonReviewListItem-deleting">
                 {i18n.gettext('Deleting…')}
               </span>
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -30,6 +30,19 @@
   }
 }
 
+.AddonReviewListItem-delete {
+  .ConfirmButton-cancel-button,
+  .ConfirmButton-confirm-button,
+  .ConfirmButton-default-button {
+    font-size: 12px;
+  }
+
+  .ConfirmButton-default-button {
+    align-items: flex-start;
+    color: $blue-50;
+  }
+}
+
 .AddonReviewListItem-control {
   @include margin-start(6px);
 
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -34,12 +34,16 @@
   .ConfirmButton-cancel-button,
   .ConfirmButton-confirm-button,
   .ConfirmButton-default-button {
-    font-size: 12px;
+    font-size: $font-size-s;
   }
 
   .ConfirmButton-default-button {
     align-items: flex-start;
     color: $blue-50;
+  }
+
+  .ConfirmButton-buttons {
+    flex-flow: row-reverse;
   }
 }
 
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -500,8 +500,6 @@ export default function reviewsReducer(
 
       if (reviewData) {
         const { addonId, addonSlug, userId } = reviewData;
-        delete newState.byId[reviewId];
-        delete newState.view[reviewId];
         return {
           ...newState,
           byAddon: {
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -490,6 +490,7 @@ export default function reviewsReducer(
       if (reviewData) {
         const { addonId, addonSlug, userId } = reviewData;
         delete newState.byId[reviewId];
+        delete newState.view[reviewId];
         return {
           ...newState,
           byAddon: {
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -2,6 +2,7 @@
 import { oneLine } from 'common-tags';
 
 import {
+  DELETE_ADDON_REVIEW,
   FLASH_REVIEW_MESSAGE,
   HIDE_FLASHED_REVIEW_MESSAGE,
   UNLOAD_ADDON_REVIEWS,
@@ -22,6 +23,7 @@ import {
   createInternalReview,
 } from 'amo/actions/reviews';
 import type {
+  DeleteAddonReviewAction,
   UnloadAddonReviewsAction,
   FlagReviewAction,
   FlashMessageType,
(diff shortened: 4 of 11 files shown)


src/amo/sagas/reviews.js
```diff
@@ -297,7 +297,7 @@ function* deleteAddonReview({
       reviewId,
     });
 
-    yield put(unloadAddonReviews({ reviewId }));
+    yield put(unloadAddonReviews({ reviewId: isReplyToReviewId || reviewId }));
```
kumar303: Why does this switch the ID in the case of a reply? Can't you always just use `reviewId`?

There is a bug that you can reproduce like this:
- As a developer, submit a reply
- View your user profile page where you can see your reply
- Go back to the add-on listing and delete your reply
- View your user profile page again.

The reply you just deleted is still visible. It disappears if you reload the page, though.

screenshot 2018-08-28 10 02 17

kumar303: It would be nice if the reducer for `unloadAddonReviews` also unloaded `state.view[reviewId]`. I don't see any bugs because of it but it would leave the UI a bit fragile since `view` is used to control things like showing and hiding the reply / edit forms.

https://github.com/mozilla/addons-frontend/blob/82e55db015d072c1baa8fe59265fcfd337c66966/src/amo/reducers/reviews.js#L97

Clearing `state.view[reviewId]` is also important since it functions independently of `state.byId`.

bobsilverberg: I guess we need to clear both. I hadn't checked that scenario. We do need to clear based on `isReplyToReviewId` for a reply, otherwise we don't clear data for reviews which had this as a reply, but yes, we also need to clear data for the actual `reviewId` of the reply.

bobsilverberg: I was actually wondering about this when I wrote `unloadAddonReviews` but wasn't sure if it was necessary. Thanks for letting me know. 👍

src/amo/actions/reviews.js
```diff
@@ -611,6 +611,7 @@ export const hideFlashedReviewMessage = (): HideFlashedReviewMessageAction => {
 
 type DeleteAddonReviewParams = {|
   errorHandlerId: string,
+  isReplyToReviewId?: number,
```
rebmullin: nit: I guess when I see a var start with "is" like this I think it's a boolean. can this just be replyToReviewId ? or is that more involved/ a bigger patch for a change like this?

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -188,6 +189,49 @@ describe(__filename, () => {
...
+    deleteButton.simulate('click', clickEvent);
+
+    sinon.assert.called(clickEvent.preventDefault);
+    sinon.assert.calledWith(
+      dispatchSpy,
+      deleteAddonReview({
+        errorHandlerId: errorHandler.id,
+        reviewId: review.id,
+      }),
+    );
```
rebmullin: should you show that this link has length 0 now (after using the delete link) ~

.find('.AddonReviewListItem-delete').toHaveLength(0)

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -707,6 +751,50 @@ describe(__filename, () => {
...
+      deleteButton.simulate('click', clickEvent);
+
+      sinon.assert.called(clickEvent.preventDefault);
+      sinon.assert.calledWith(
+        dispatchSpy,
+        deleteAddonReview({
+          errorHandlerId: errorHandler.id,
+          reviewId: review.id,
+          isReplyToReviewId: originalReviewId,
+        }),
```
rebmullin: same here? should we show the that link is gone after the dispatch?

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -54,6 +56,20 @@ type InternalProps = {|
 |};
 
 export class AddonReviewListItemBase extends React.Component<InternalProps> {
+  onClickToDeleteReview = (event: SyntheticEvent<HTMLElement>) => {
+    const { dispatch, errorHandler, isReplyToReviewId, review } = this.props;
+    event.preventDefault();
+
+    invariant(review, 'review is required');
```
rebmullin: just curious, does invariant need to be here and in actions/reducer? If its ok, should we add errorHandler too? it seems that's required too, right?

bobsilverberg: Because we've told Flow that `review` can be undefined or null (it's declared as `review?: UserReviewType | null`), we need to use `invariant` here to tell Flow that we never expect it to be undefined or null at this point, and if it is an exception should be thrown.

bobsilverberg:
Thanks for the review @rebmullin.

> Also, not sure if you know about this already but after deleting my review, I still see the my star rating on the reviews page (it's cleared out on the detail page). It seems it should be cleared out on both, right?

Do you mean you still see it on the user profile page? If so that should be fixed with the latest commit, that I just pushed. If not, then I'm not sure where it is that you still see if after deleting.

src/amo/actions/reviews.js
bobsilverberg: Yeah, I thought the same thing at first, but this is what the same variable is called elsewhere. I suppose I can just change it here.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -234,6 +254,24 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
+            <ConfirmButton
+              buttonType="cancel"
+              className={makeClassName(
+                'AddonReviewListItem-control',
+                confirmButtonClassName,
+              )}
+              id={confirmButtonClassName}
```
rebmullin: I think there should only be 1 id on a page so maybe id can be concatenated with review id so they are unique?

rebmullin:
also, i *think* the styles were ok before but now it looks off:

Alt text

do you see that too? or is this intended?

rebmullin:
@bobsilverberg, sorry I meant to add a screenshot for my comment above. This on the reviews detail page:

Alt text

see the stars on the left side (With my 4 star rating)? If you go back to the detail page, you can see it's removed there. 

Does that help clarify?

src/amo/components/AddonReviewListItem/index.js
rebmullin: to clarify: should only have 1 id with the same name that is...

src/amo/components/AddonReviewListItem/index.js
rebmullin: ok I think I see..thanks : )

tests/unit/amo/components/TestAddonReviewListItem.js
bobsilverberg: This is an interesting idea, and might have been simple before changing this to use `ConfirmButton`, but now the test does not actually click on the button, it just calls the `onConfirm` method of the button. In order to do this now we'd have to first click on the `ConfirmButton` link and then click on the actual "Confirm" button that is displayed by the `ConfirmButton` component. That seems like way too much for this test. 

What do you think, @kumar303? Can you think of a simpler way of confirming that the "Delete" link is no longer present immediately after the deletion occurs?

src/amo/components/AddonReviewListItem/index.js
bobsilverberg: Ah, good point!

src/amo/actions/reviews.js
rebmullin: yeh I kind of remember that (it's in other places)..I think this is a nit so i would say only if it's not too involved otherwise that can be a follow up patch/issue if you think it's worth it..

bobsilverberg:
@rebmullin I fixed that style issue so I'm not sure why you're seeing it now. Here's how it looks to me:

screenshot 2018-08-28 15 03 58

bobsilverberg:
@rebmullin 

> Does that help clarify?

No, not really ;-)

Do you mean the big yellow stars on the left? Those always represent the average rating on the reviews page, not your current rating. Are you expecting those to change?

bobsilverberg:
@kumar303 and @rebmullin Note that I noticed another bug which I might just want to wait to address in a follow-up. It's the fact that the total number of reviews for the add-on is not updated when you delete your review. I think there's even more cache cleanup that may need to be done to address that.

rebmullin:
> Do you mean the big yellow stars on the left? Those always represent the average rating on the reviews page, not your current rating. Are you expecting those to change?

oh my bad! I thought that was my rating...

Please disregard then : )

bobsilverberg:
@kumar303 I've addressed this issues you raised, including implementing the `ConfirmButton`. I tried to style it as closely to the current screen as possible, but I'm not convinced that it looks as good as it might.

rebmullin:
hmm @bobsilverberg, styles do look off to me but maybe it's just on my end :/

bobsilverberg:
Here's what it currently looks like when confirming:

screenshot 2018-08-28 15 10 06

rebmullin:
@bobsilverberg @kumar303, would confirm of delete review look better in a modally style here?  ~

Alt text

bobsilverberg:
> would confirm of delete review look better in a modally style here? 

That does look a bit nicer to me. Is that something we have built-in to `ConfirmButton` right now, or is that something you just mocked up?

rebmullin:
@bobsilverberg, initially I was thinking we could use the actual overlay/ modal but then thought that might be too much (since it overlays the whole page). We do have something for that...
You might be able to be override for this case here easily but the version I show above, I just mocked up by changing a few styles...
maybe that could be follow up issue if you guys want to try to go this way.?

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -234,6 +254,24 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
+            <ConfirmButton
```
kumar303: Thanks for adding this. It will help. Here are some small suggestions that I think would make it look better.

- Make it say *Delete* to be clear
- Use a proper cancel button to differentiate the word *Cancel* from the text prompt
- Always render the buttons in a row, even on mobile. I think the default of switching to column mode on mobile was more for the case of cancel text (not a button).

screenshot 2018-08-28 15 00 13
screenshot 2018-08-28 15 00 32
screenshot 2018-08-28 15 07 12

```diff
diff --git a/src/amo/components/AddonReviewListItem/index.js b/src/amo/components/AddonReviewListItem/index.js
index 46607cc3a..18a10c8dd 100644
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -255,6 +255,8 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 : i18n.gettext('Edit my review')}
             </a>
             <ConfirmButton
+              cancelButtonType="neutral"
+              confirmButtonText={i18n.gettext('Delete')}
               buttonType="cancel"
               className={makeClassName(
                 'AddonReviewListItem-control',
diff --git a/src/amo/components/AddonReviewListItem/styles.scss b/src/amo/components/AddonReviewListItem/styles.scss
index e6b0da742..4eb99974a 100644
--- a/src/amo/components/AddonReviewListItem/styles.scss
+++ b/src/amo/components/AddonReviewListItem/styles.scss
@@ -41,6 +41,10 @@
     align-items: flex-start;
     color: $blue-50;
   }
+
+  .ConfirmButton-buttons {
+    flex-flow: row-reverse;
+  }
 }
 
 .AddonReviewListItem-control {
```

src/amo/components/AddonReviewListItem/styles.scss
```diff
@@ -30,6 +30,19 @@
   }
 }
 
+.AddonReviewListItem-delete {
+  .ConfirmButton-cancel-button,
+  .ConfirmButton-confirm-button,
+  .ConfirmButton-default-button {
+    font-size: 12px;
```
kumar303: This needs to use `$font-size-s`

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -188,6 +189,58 @@ describe(__filename, () => {
...
+      reviewUserId: 987,
+    });
+    const root = render({ review });
+
+    expect(
+      renderControls(root).find('.AddonReviewListItem-delete'),
+    ).toHaveLength(0);
+  });
+
+  it('dispatches deleteReview when a user clicks the delete link', () => {
```
kumar303: It might be good to update this spec

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -188,6 +189,58 @@ describe(__filename, () => {
...
+      '.AddonReviewListItem-delete',
+    );
+    const clickEvent = createFakeEvent();
+    deleteButton.simulate('click', clickEvent);
+
+    // This emulates a user clicking the delete button and confirming.
+    const onDelete = deleteButton.prop('onConfirm');
+    onDelete(createFakeEvent({ preventDefault: preventDefaultSpy }));
+
+    sinon.assert.calledOnce(preventDefaultSpy);
```
kumar303: This should work, no? The properties are already spies.

```js
const deleteEvent = createFakeEvent();
onDelete(deleteEvent);
sinon.assert.calledOnce(deleteEvent.preventDefault);
```

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -188,6 +189,58 @@ describe(__filename, () => {
...
+    const dispatchSpy = sinon.spy(store, 'dispatch');
+    const preventDefaultSpy = sinon.spy();
+    const root = render({ review });
+    const { errorHandler } = root.instance().props;
+
+    const deleteButton = renderControls(root).find(
+      '.AddonReviewListItem-delete',
+    );
+    const clickEvent = createFakeEvent();
+    deleteButton.simulate('click', clickEvent);
```
kumar303: This is clicking `<ConfirmButton>`, right? I don't think that's necessary and it's also not a good idea because it reaches into the implementation of `ConfirmButton`. You can trust that `ConfirmButton` works as advertised. All you need to do is make sure its props are set correctly.

tests/unit/amo/components/TestAddonReviewListItem.js
kumar303: https://github.com/mozilla/addons-frontend/pull/6084/files#r213419834

> should you show that this link has length 0 now (after using the delete link)

No, because that's the responsibility of `ConfirmButton`.

However, I don't see a test case for this! Good catch. 

Can you add one to https://github.com/mozilla/addons-frontend/blob/master/tests/unit/ui/components/TestConfirmButton.js ?

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -707,6 +760,57 @@ describe(__filename, () => {
...
+      );
+      expect(deleteLink).toHaveLength(1);
+      expect(deleteLink.children()).toHaveText('Delete my reply');
+      expect(deleteLink).toHaveProp(
+        'message',
+        'Do you really want to delete this reply?',
+      );
+    });
+
+    it('dispatches deleteReview when a user clicks the delete link for a developer reply', () => {
```
kumar303: It can just be "dispatches deleteReview when a user deletes a developer reply"

tests/unit/amo/components/TestAddonReviewListItem.js
```diff
@@ -707,6 +760,57 @@ describe(__filename, () => {
...
+      const { errorHandler } = root.instance().props;
+
+      const deleteButton = renderControls(root).find(
+        '.AddonReviewListItem-delete',
+      );
+      // This emulates a user clicking the delete button and confirming.
+      const onDelete = deleteButton.prop('onConfirm');
+      onDelete(createFakeEvent({ preventDefault: preventDefaultSpy }));
+
+      sinon.assert.calledOnce(preventDefaultSpy);
```
kumar303: I think this can also be `sinon.assert.calledOnce(deleteEvent.preventDefault);`

tests/unit/amo/sagas/test_reviews.js
```diff
@@ -719,6 +728,56 @@ describe(__filename, () => {
...
+      _deleteAddonReview({ isReplyToReviewId, reviewId });
+
+      const expectedAction = unloadAddonReviews({
+        reviewId,
+      });
+      const action = await matchingSagaAction(
+        sagaTester,
+        matchReviewId(expectedAction),
+      );
+      expect(action).toEqual(expectedAction);
```
kumar303: This is probably redundant. I guess it doesn't hurt, though.

tests/unit/helpers.js
```diff
@@ -710,11 +711,13 @@ export async function matchingSagaAction(
   { maxTries = 50 } = {},
 ) {
   let calledActions = [];
+  let foundAction;
   let matched = false;
```
kumar303: I think you can delete `matched` now and just look for a truthy `foundAction`.

src/amo/components/AddonReviewListItem/index.js
willdurand: If we move this button/confirm button to one side (right side for instance), that might look better and might be easier to style as we could add some `margin-start` and/or a thin vertical border.

tests/unit/amo/components/TestAddonReviewListItem.js
bobsilverberg: I think we're talking about two different things here. I believe what @rebmullin was suggesting is a test that shows that after deleting the review/reply the edit link no longer appears, but maybe not because the entire review/reply would disappear so the edit link must also disappear.

I think what you're referring to @kumar303 is a test that shows that when you click the link for a `ConfirmButton`, that, in addition to the confirmation dialog appearing, the link also disappears. Is that correct?

src/amo/components/AddonReviewListItem/index.js
bobsilverberg: What do you think of @willdurand's suggestion @kumar303? I agree that the confirm dialog might look better if it were all the way on the left, but I'm not sure we want to move the "Edit" link to the right of the "Flag" link.

tests/unit/amo/components/TestAddonReviewListItem.js
bobsilverberg: Yeah, that was just me forgetting to remove that line when I changed it to just call the `onConfirm` prop.

bobsilverberg:
I have addressed all of the outstanding issues. Here are a couple of screenshots of the new confirm dialog:

screenshot 2018-08-29 10 12 40
screenshot 2018-08-29 10 13 07

src/amo/components/AddonReviewListItem/index.js
kumar303: I'm not sure what this means. The confirm button is already on the right side. I think it would be good to land this patch so that people can try it on dev then follow up with further UX changes -- e.g. Phil may have some ideas for improvement, too.

src/amo/reducers/reviews.js
```diff
@@ -490,6 +490,7 @@ export default function reviewsReducer(
       if (reviewData) {
         const { addonId, addonSlug, userId } = reviewData;
         delete newState.byId[reviewId];
+        delete newState.view[reviewId];
```
kumar303: Well, huh. I learned today that you can't use `delete` in a reducer.

Here's how to break the app:
- Submit a review
- Go to the review listing page
- Delete the review
- Open the Redux Devtool
- Jump back to the state triggered by the action `DELETE_ADDON_REVIEW`

The app is now broken (empty page, error in the console) because `delete` erased objects that were stored in previous states. Here's how to fix it (we need to always do this, so I filed https://github.com/mozilla/addons-frontend/issues/6102):

```diff
diff --git a/src/amo/reducers/reviews.js b/src/amo/reducers/reviews.js
index 240b26304..aadc7cb82 100644
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -489,10 +489,16 @@ export default function reviewsReducer(
 
       if (reviewData) {
         const { addonId, addonSlug, userId } = reviewData;
-        delete newState.byId[reviewId];
-        delete newState.view[reviewId];
         return {
           ...newState,
+          byId: {
+            ...newState.byId,
+            [reviewId]: undefined,
+          },
+          view: {
+            ...newState.view,
+            [reviewId]: undefined,
+          },
           byAddon: {
             ...newState.byAddon,
             [addonSlug]: undefined,
```

I know that the only broken piece here is the Redux Devtool but I'd like to preserve that functionality. In this case, it's really important because I wanted to see what the loading state looks like. Without Redux time travel, it's hard to "pause" the app in a loading state.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -54,6 +58,20 @@ type InternalProps = {|
 |};
 
 export class AddonReviewListItemBase extends React.Component<InternalProps> {
+  onClickToDeleteReview = (event: SyntheticEvent<HTMLElement>) => {
+    const { dispatch, errorHandler, isReplyToReviewId, review } = this.props;
+    event.preventDefault();
+
+    invariant(review, 'review is required');
+    dispatch(
+      deleteAddonReview({
```
kumar303: I just noticed that when this action dispatches, the UI does not indicate that anything is happening. This could be confusing to users on a slow connection. In fact, they may think nothing is happening and they may try to delete the review again.

Here's a simple idea for a fix:

screenshot 2018-08-29 10 31 05


<details>
<summary>patch</summary>

```diff
diff --git a/src/amo/components/AddonReviewListItem/index.js b/src/amo/components/AddonReviewListItem/index.js
index eb78c9925..2dd647224 100644
--- a/src/amo/components/AddonReviewListItem/index.js
+++ b/src/amo/components/AddonReviewListItem/index.js
@@ -47,6 +47,7 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
+  deletingReview: boolean,
   dispatch: DispatchFunc,
   editingReview: boolean,
   errorHandler: ErrorHandlerType,
@@ -199,6 +200,7 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
   render() {
     const {
       addon,
+      deletingReview,
       editingReview,
       errorHandler,
       i18n,
@@ -254,26 +256,32 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
                 ? i18n.gettext('Edit my reply')
                 : i18n.gettext('Edit my review')}
             </a>
-            <ConfirmButton
-              buttonType="cancel"
-              cancelButtonType="neutral"
-              className={makeClassName(
-                'AddonReviewListItem-control',
-                confirmButtonClassName,
-              )}
-              confirmButtonText={i18n.gettext('Delete')}
-              id={`${confirmButtonClassName}-${review.id}`}
-              message={
-                isReply
-                  ? i18n.gettext('Do you really want to delete this reply?')
-                  : i18n.gettext('Do you really want to delete this review?')
-              }
-              onConfirm={this.onClickToDeleteReview}
-            >
-              {isReply
-                ? i18n.gettext('Delete my reply')
-                : i18n.gettext('Delete my review')}
-            </ConfirmButton>
+            {deletingReview ? (
+              <span className="AddonReviewListItem-control">
+                {i18n.gettext('Deleting…')}
+              </span>
+            ) : (
+              <ConfirmButton
+                buttonType="cancel"
+                cancelButtonType="neutral"
+                className={makeClassName(
+                  'AddonReviewListItem-control',
+                  confirmButtonClassName,
+                )}
+                confirmButtonText={i18n.gettext('Delete')}
+                id={`${confirmButtonClassName}-${review.id}`}
+                message={
+                  isReply
+                    ? i18n.gettext('Do you really want to delete this reply?')
+                    : i18n.gettext('Do you really want to delete this review?')
+                }
+                onConfirm={this.onClickToDeleteReview}
+              >
+                {isReply
+                  ? i18n.gettext('Delete my reply')
+                  : i18n.gettext('Delete my review')}
+              </ConfirmButton>
+            )}
           </React.Fragment>
         ) : null}
 
@@ -323,18 +331,21 @@ export class AddonReviewListItemBase extends React.Component<InternalProps> {
 }
 
 export function mapStateToProps(state: AppState, ownProps: Props) {
+  let deletingReview = false;
   let editingReview = false;
   let replyingToReview = false;
   let submittingReply = false;
   if (ownProps.review) {
     const view = state.reviews.view[ownProps.review.id];
     if (view) {
+      deletingReview = view.deletingReview;
       editingReview = view.editingReview;
       replyingToReview = view.replyingToReview;
       submittingReply = view.submittingReply;
     }
   }
   return {
+    deletingReview,
     editingReview,
     siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
     replyingToReview,
diff --git a/src/amo/reducers/reviews.js b/src/amo/reducers/reviews.js
index 240b26304..87070a35f 100644
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -2,6 +2,7 @@
 import { oneLine } from 'common-tags';
 
 import {
+  DELETE_ADDON_REVIEW,
   FLASH_REVIEW_MESSAGE,
   HIDE_FLASHED_REVIEW_MESSAGE,
   UNLOAD_ADDON_REVIEWS,
@@ -22,6 +23,7 @@ import {
   createInternalReview,
 } from 'amo/actions/reviews';
 import type {
+  DeleteAddonReviewAction,
   UnloadAddonReviewsAction,
   FlagReviewAction,
   FlashMessageType,
@@ -75,6 +77,7 @@ export type FlagState = {
 };
 
 type ViewStateByReviewId = {|
+  deletingReview: boolean,
   editingReview: boolean,
   flag: FlagState,
   replyingToReview: boolean,
@@ -175,6 +178,7 @@ export const changeViewState = ({
     view: {
       ...state.view,
       [reviewId]: {
+        deletingReview: false,
         editingReview: false,
         replyingToReview: false,
         submittingReply: false,
@@ -270,6 +274,7 @@ export const addReviewToState = ({
 };
 
 type ReviewActionType =
+  | DeleteAddonReviewAction
   | FlagReviewAction
   | FlashReviewMessageAction
   | UnloadAddonReviewsAction
@@ -296,6 +301,12 @@ export default function reviewsReducer(
   }: {| _addReviewToState: typeof addReviewToState |} = {},
 ) {
   switch (action.type) {
+    case DELETE_ADDON_REVIEW:
+      return changeViewState({
+        state,
+        reviewId: action.payload.reviewId,
+        stateChange: { deletingReview: true },
+      });
     case SEND_REPLY_TO_REVIEW:
       return changeViewState({
         state,
```

</details>

What do you think?

bobsilverberg: I like it! It's pretty convenient that we don't need a `finishedDeleting` action because the delete process itself will clear the state of the `deletingReview` flag. 👍

src/amo/reducers/reviews.js
bobsilverberg: @kumar303 Did you try it with this patch? Because even with this patch the Redux devtools are still broken after I delete a review. I think it's because of the way the reducer works. I can look into it further, but I wanted to check whether implementing this solution actually fixed the redux devtools for you, or if you just assumed it would.

bobsilverberg:
Thanks for the review @kumar303. I've fixed the reducer to not delete the keys, although that doesn't seem to have fixed the redux devtools. I've also implemented the "Deleting…" message.

PR #6095, opened 2018-08-28
Fix notice for non-public add-ons

willdurand:
Refs #4244

### DIFF
--- a/tests/unit/amo/components/TestAddon.js
+++ b/tests/unit/amo/components/TestAddon.js
@@ -1568,8 +1568,9 @@ describe(__filename, () => {
     });
   });
 
-  // Non-public add-ons require an account listed as a developer of the add-on or admin rights.
-  it('displays a notice on a disabled addon to admin/developer', () => {
+  // Non-public add-ons require an account listed as a developer of the add-on
+  // or admin rights.
+  it('displays a notice to admin/developer when add-on is not fully reviewed', () => {
     const addon = createInternalAddon({
       ...fakeAddon,
       status: 'disabled',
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestAddon.js
```diff
@@ -1568,8 +1568,9 @@ describe(__filename, () => {
     });
   });
 
-  // Non-public add-ons require an account listed as a developer of the add-on or admin rights.
-  it('displays a notice on a disabled addon to admin/developer', () => {
+  // Non-public add-ons require an account listed as a developer of the add-on
+  // or admin rights.
+  it('displays a notice to admin/developer when add-on is not fully reviewed', () => {
```
rebmullin: can you just put "..is not reviewed"? Unless there are steps for review that maybe I am not aware of?

Also, it *seems* like it could also be reviewed and have this be "invisible" too so I am wondering if this description is completely accurate?

tests/unit/amo/components/TestAddon.js
```diff
@@ -1579,7 +1580,7 @@ describe(__filename, () => {
     expect(root.find(Notice)).toHaveLength(1);
   });
 
-  it('does not display a disabled notice for a public addon', () => {
+  it('does not display a notice when add-on is fully reviewed', () => {
```
rebmullin: when add on is ~ "...visible (shared) and has been reviewed"?

tests/unit/amo/components/TestAddon.js
willdurand: > can you just put "..is not reviewed"? Unless there are steps for review that maybe I am not aware of?
> 
> Also, it seems like it could also be reviewed and have this be "invisible" too so I am wondering if this description is completely accurate?

It is "fully reviewed" according to this: https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#detail. There are indeed different review states.

rebmullin: ahh I see..public is "fully reviewed."

I guess me marking this as "invisible" on a reviewed addon is also what confused me:
Alt text

willdurand: yes, it is super weird. invisible === disabled apparently... but it affects the `is_disabled` prop and not the status -_-

rebmullin: I still think we should mention something about this (that it would shown if reviewed but invisible)

willdurand: we have another test case for that below. if we ever need yet another condition, I'd like not to update other descriptions with "... and ...". This description still holds.

rebmullin: oh? I didn't see the other test case. if that's there then cool.

PR #6096, opened 2018-08-28
Fix AMInstallButton alignment on addon detail page

### LINKED ISSUE #6092: Install button is aligned to the left when there is no summary available
STR:

1. Open AMO -dev
2. Go to an add-on details page that does not have a summary or description - https://addons-dev.allizom.org/en-US/firefox/addon/green-winter/?src=hotness
3. Observe the position of the Install button

Actual result:
Install button is aligned to the left

Expected result:
Install button is aligned to the right

Notes:
- observed both for themes (lwt and static) and extensions, but it will be most common for lwt
- reproduced on AMO -dev with FF61, Win10x64

image

willdurand:
Fix #6092

---

Before:

screen shot 2018-08-28 at 17 49 07

After:

screen shot 2018-08-28 at 17 49 05

### DIFF
--- a/src/amo/components/Addon/styles.scss
+++ b/src/amo/components/Addon/styles.scss
@@ -6,6 +6,7 @@
   @include page-padding();
 
   .Addon-summary-and-install-button-wrapper {
+    .AMInstallButton,
     .InstallButton,
     .Button--get-firefox {
       @include respond-to(medium) {
@@ -151,6 +152,7 @@
   word-wrap: break-word;
 }
 
+.Addon .AMInstallButton,
 .Addon .InstallButton {
   @include respond-to(medium) {
     @include margin-start(auto);
@@ -160,6 +162,7 @@
   }
 }
 
+.Addon .AMInstallButton-button,
 .Addon .InstallButton-button,
 .Button--get-firefox {
   width: 100%;


src/amo/components/Addon/styles.scss
```diff
@@ -151,6 +152,7 @@
   word-wrap: break-word;
 }
 
+.Addon .AMInstallButton,
```
willdurand: I am not sure about that, but it was set for `InstallButton` so I added it..

src/amo/components/Addon/styles.scss
```diff
@@ -160,6 +162,7 @@
   }
 }
 
+.Addon .AMInstallButton-button,
```
willdurand: I am not sure about that, but it was set for `InstallButton` so I added it..

willdurand: ah this one fixes that:

Before

screen shot 2018-08-28 at 17 53 51


After

screen shot 2018-08-28 at 17 53 46

rebmullin: just checking...do we want this? on stage it is full width on ~ mobile (with the installButton styles). This (full width on mobile) is also consistent with the "..get firefox now" button on non-firefox (chrome) for example. Or should it be smaller now in this case?

willdurand: this is not mobile here

rebmullin: ok cool, I see. 👍  then : )

Issue #6102, opened 2018-08-29
Reducers should not delete any object in state (this breaks the Redux devtool)


kumar303:
**For QA**: please check for regressions in the following features:
- Delete a collection
- Delete your user account

<hr>

When a reducer handles a deletion action, sometimes it does so with `delete`:

```js
case DELETE_A_THING: {
  const newState = { ...state };
  delete newState.byId[thingId];
  return newState;
}
```

This works for normal site usage but it breaks Redux time travel. You cannot go back in time to the previous state because all references to the old object have been removed. Going back in time will probably result in thrown error somewhere.

Here is a safe way to delete objects:

```js
case DELETE_A_THING: {
  return {
    ...state,
    byId: {
      ...state.byId,
      [thingId]: undefined,
    },
  };
}
```

Let's update all reducers.

ioanarusiczki:
I tested for regressions on AMO dev with FF61 (Win10&Android8.0)
> Delete a collection
   Delete your user account
Delete works for both.

PR #6106, opened 2018-08-29
Remove delete statements from all reducers

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6102

### DIFF
--- a/src/amo/reducers/users.js
+++ b/src/amo/reducers/users.js
@@ -548,12 +548,19 @@ const reducer = (
       if (newState.byID[userId]) {
         const { username } = newState.byID[userId];
 
-        delete newState.byID[userId];
-        delete newState.byUsername[username];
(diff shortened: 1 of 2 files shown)


src/amo/reducers/users.js
```diff
@@ -548,12 +548,19 @@ const reducer = (
...
         const { username } = newState.byID[userId];
 
-        delete newState.byID[userId];
-        delete newState.byUsername[username];
-
-        if (newState.currentUserID === userId) {
-          newState.currentUserID = null;
-        }
+        return {
+          ...newState,
```
bobsilverberg: Do we even need `newState` anymore with this change? Could it just be `state`?

kumar303: Duh. Thanks.

PR #6108, opened 2018-08-30
Deleting a review does not update the number of reviews for the add-on

### LINKED ISSUE #6105: Deleting a review does not update the number of reviews for the add-on
~~Depends on #6084~~

STR:
1. Go to an add-on's detail page.
2. Create a review for the add-on.
3. Note the number of reviews the add-on has (in the upper right-hand card).
4. Click on the link to view all reviews for the add-on.
5. Observe the number of reviews shown in the header for the review listing. It should be (and is) the same number as observed in step 3.
6. Delete your review.
7. Observe the number of reviews shown in the header for the review listing. It should be decremented by 1.
8. Click on the add-on's name in the left-hand card to navigate back to the details for the add-on.
9. Observe the number of reviews the add-on has (in the upper right-hand card). It should also have been decremented by 1.

Expected:
The number of reviews in steps 7 and 9 should be 1 less than the number from step 3.

Actual:
The number of reviews in steps 7 and 9 are the same as the number from step 3.

bobsilverberg:
Fixes #6105

### DIFF
--- a/src/amo/components/AddonReviewList/index.js
+++ b/src/amo/components/AddonReviewList/index.js
@@ -43,7 +43,7 @@ import type { I18nType } from 'core/types/i18n';
 
 import './styles.scss';
 
-type Props = {|
+type InternalProps = {|
   addon: AddonType | null,
   addonIsLoading: boolean,
   clientApp: string,
@@ -64,16 +64,16 @@ type Props = {|
   history: ReactRouterHistoryType,
 |};
 
-export class AddonReviewListBase extends React.Component<Props> {
+export class AddonReviewListBase extends React.Component<InternalProps> {
   componentWillMount() {
     this.loadDataIfNeeded();
   }
 
-  componentWillReceiveProps(nextProps: Props) {
+  componentWillReceiveProps(nextProps: InternalProps) {
     this.loadDataIfNeeded(nextProps);
   }
 
-  loadDataIfNeeded(nextProps?: Props) {
+  loadDataIfNeeded(nextProps?: InternalProps) {
     const lastAddon = this.props.addon;
     const nextAddon = nextProps && nextProps.addon;
     const {
@@ -338,7 +338,7 @@ export class AddonReviewListBase extends React.Component<Props> {
   }
 }
 
-export function mapStateToProps(state: AppState, ownProps: Props) {
+export function mapStateToProps(state: AppState, ownProps: InternalProps) {
   const { addonSlug } = ownProps.match.params;
   const reviewData = state.reviews.byAddon[addonSlug];
 
@@ -358,7 +358,7 @@ export function mapStateToProps(state: AppState, ownProps: Props) {
   };
 }
 
-export const extractId = (ownProps: Props) => {
+export const extractId = (ownProps: InternalProps) => {
   const {
     location,
     match: { params },
@@ -366,6 +366,8 @@ export const extractId = (ownProps: Props) => {
 
   return `${params.addonSlug}-${location.query.page || ''}`;
 };
+
+type Props = {||};
 
 const AddonReviewList: React.ComponentType<Props> = compose(
   connect(mapStateToProps),
--- a/src/amo/components/AddonReviewList/index.js
+++ b/src/amo/components/AddonReviewList/index.js
@@ -43,7 +43,12 @@ import type { I18nType } from 'core/types/i18n';
 
 import './styles.scss';
 
+type Props = {|
+  location: ReactRouterLocationType,
+|};
+
 type InternalProps = {|
+  ...Props,
   addon: AddonType | null,
   addonIsLoading: boolean,
   clientApp: string,
@@ -51,7 +56,6 @@ type InternalProps = {|
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
   lang: string,
-  location: ReactRouterLocationType,
   match: {|
     ...ReactRouterMatchType,
     params: {
@@ -366,8 +370,6 @@ export const extractId = (ownProps: InternalProps) => {
 
   return `${params.addonSlug}-${location.query.page || ''}`;
 };
-
-type Props = {||};
 
 const AddonReviewList: React.ComponentType<Props> = compose(
   connect(mapStateToProps),
--- a/src/amo/components/AddonReviewList/index.js
+++ b/src/amo/components/AddonReviewList/index.js
@@ -12,7 +12,11 @@ import RatingsByStar from 'amo/components/RatingsByStar';
 import { fetchReviews } from 'amo/actions/reviews';
 import { setViewContext } from 'amo/actions/viewContext';
 import { expandReviewObjects } from 'amo/reducers/reviews';
-import { fetchAddon, getAddonBySlug } from 'core/reducers/addons';
+import {
+  fetchAddon,
+  getAddonBySlug,
+  isAddonLoading,
+} from 'core/reducers/addons';
 import Paginate from 'core/components/Paginate';
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
@@ -40,11 +44,12 @@ import type { I18nType } from 'core/types/i18n';
 import './styles.scss';
 
 type Props = {|
-  i18n: I18nType,
   addon: AddonType | null,
+  addonIsLoading: boolean,
   clientApp: string,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
+  i18n: I18nType,
   lang: string,
   location: ReactRouterLocationType,
   match: {|
@@ -75,6 +80,7 @@ export class AddonReviewListBase extends React.Component<Props> {
       addon,
       dispatch,
       errorHandler,
+      addonIsLoading,
       match: { params },
       reviews,
     } = {
@@ -88,7 +94,9 @@ export class AddonReviewListBase extends React.Component<Props> {
     }
 
     if (!addon) {
-      dispatch(fetchAddon({ slug: params.addonSlug, errorHandler }));
+      if (!addonIsLoading) {
+        dispatch(fetchAddon({ slug: params.addonSlug, errorHandler }));
+      }
     } else if (
       // This is the first time rendering the component.
       !nextProps ||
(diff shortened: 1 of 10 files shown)


src/core/reducers/addons.js
```diff
@@ -11,17 +12,20 @@ import type {
...
+type AddonID = number;
 
 type ExternalAddonMap = {
   [addonSlug: string]: ExternalAddonType,
 };
 
 export type LoadAddonsAction = {|
   payload: {| addons: ExternalAddonMap |},
-  type: string,
+  type: typeof LOAD_ADDONS,
```
kumar303: 👍

src/core/reducers/addons.js
```diff
@@ -346,6 +371,24 @@ export default function addonsReducer(
         bySlug,
       };
     }
+    case UNLOAD_ADDON: {
```
kumar303: Why not adjust `UnloadAddonReviewsAction` to have `addon` in the payload and have this switch case look for `UNLOAD_ADDON_REVIEWS`? In other words, I don't think you need a separate `unloadAddon()` action. I can't imagine any situation other than deleting reviews where we would need a dedicated `unloadAddon()` action.

src/amo/components/AddonReviewListItem/index.js
```diff
@@ -60,12 +60,20 @@ type InternalProps = {|
...
+    } = this.props;
     event.preventDefault();
 
+    invariant(addon, 'addon is required');
     invariant(review, 'review is required');
     dispatch(
       deleteAddonReview({
+        addon,
         errorHandlerId: errorHandler.id,
         reviewId: review.id,
```
kumar303: I thought it would be clever to change this from `review.id` to `review` so you could use `review.addon` but that doesn't get us a `guid`. Oh, well.

https://github.com/mozilla/addons-frontend/blob/eae17b8d2dad542e97ecc00cffe5fc870554467f/src/amo/api/reviews.js#L13-L16

kumar303:
This seems like the right fix but after your patch are you still running into server caching issues? https://github.com/mozilla/addons-server/issues/9293

bobsilverberg:
> Do you think we need a dedicated unloadAddon() action? It would simplify the patch without introducing a new action.

I was actually surprised we didn't have one already! But thinking about it some more, I guess if we don't foresee any other cases where we'd need to do it we can forego creating a new action at this time. I'll give it a shot.

bobsilverberg:
Well, I tried the different approach, but there's a problem that actually creates an infinite loop. We need to make sure that the code to remove the cached reviews happen before the code to remove the cached add-on, otherwise we end up calling `FETCH_ADDON` over and over again. I actually encountered this on this patch when I put the call to `unloadAddon` before the calls to `unloadAddonReviews`. It worked when I put `unloadAddon` at the end, but now I'm thinking maybe I just got lucky with that and there still might be a timing issue. 

@kumar303 what do you think? Can we leave it as is, or do I need to dig into that further?

bobsilverberg:
@kumar303 I pushed my changes that cause the infinite loop, so you can see what I'm talking about. All of the tests pass, but if you try to run it interactively you'll see the issue.

bobsilverberg:
I think that requiring an `addon` as a required argument to `unloadAddonReviews` is going to cause other problems down the road, so I think I'm going to change it to just use `addonId` instead.

kumar303:
> I pushed my changes that cause the infinite loop

I'm pretty sure what's happening is this:
* The review gets deleted
* The add-on gets set to `undefined` in the reducer
* This causes `Addon` to dispatch `fetchAddon()` repeatedly since the add-on is still `undefined` while it is being loaded from the API

Here is a patch I added that stopped the infinite loop which is why I'm posing the theory up above.

```diff
diff --git a/src/core/sagas/addons.js b/src/core/sagas/addons.js
index 4f833e482..b86dd0c49 100644
--- a/src/core/sagas/addons.js
+++ b/src/core/sagas/addons.js
@@ -3,7 +3,7 @@
 // Disabled because of
 // https://github.com/benmosher/eslint-plugin-import/issues/793
 /* eslint-disable import/order */
-import { call, put, select, takeEvery } from 'redux-saga/effects';
+import { call, put, select, takeLatest, throttle } from 'redux-saga/effects';
 /* eslint-enable import/order */
 
 import { fetchAddon as fetchAddonFromApi } from 'core/api';
@@ -29,5 +29,5 @@ export function* fetchAddon({
 }
 
 export default function* addonsSaga(): Generator<any, any, any> {
-  yield takeEvery(FETCH_ADDON, fetchAddon);
+  yield throttle(3000, FETCH_ADDON, fetchAddon);
 }
```

If my theory is correct then **how is this only happening now?** And how does splitting it into two unload actions solve it? Maybe my theory is not correct.

Usually we prevent loops like this by introducing a loading indicator in the state. Something like `FETCH_ADDON` would immediately set `loading=true` so that the component knows not to start another fetch routine. Example: 

https://github.com/mozilla/addons-frontend/blob/22bdd384e4a39b3e4f80ae5acfdeb67f5cd6f1b5/src/amo/components/LandingPage/index.js#L110-L121

Notice `if ... !loading`.

When I dug around in some other components, I noticed that we don't actually check for loading flags all that much. Hmm.

Hopefully this helps. I didn't get time to dig much deeper.

bobsilverberg:
Thanks for the feedback @kumar303, that does help and I'll try out your patch. I  suspect (as it seems you do too) that something different is going on, but I'll dig a bit deeper. I think it specifically has to do with the interaction between reviews and add-ons.

bobsilverberg:
It seems that there's some interaction between `fetchAddon`, `mapStateToProps` and `componentWillReceiveProps` that is causing `componentWillReceiveProps` to be called over and over again, and when it finds that `addon` is `null` it calls `fetchAddon` again, and so on. I cannot quite figure out what's triggering it over and over again, but I do think @kumar303's comments above about having a loading state for add-ons makes a lot of sense, and if we do that we can eliminate this problem by not fetching an add-on if it's already being loaded. The latest version of this patch does just that.

src/amo/components/AddonReviewList/index.js
```diff
@@ -40,11 +44,12 @@ import type { I18nType } from 'core/types/i18n';
 import './styles.scss';
 
 type Props = {|
-  i18n: I18nType,
   addon: AddonType | null,
+  addonIsLoading: boolean,
```
kumar303: Can you create `InternalProps` and move all the mapped and injected props there? I'm not sure why this component never got it.

src/core/reducers/addons.js
```diff
@@ -330,6 +356,7 @@ export default function addonsReducer(
 
         if (addon.slug) {
           bySlug[addon.slug] = addon.id;
+          loadingBySlug[addon.slug] = false;
```
kumar303: I realized that doing assignments like this is bad for the same reason as using the [delete statement](https://github.com/mozilla/addons-frontend/issues/6102) -- it breaks the Redux devtool. It would probably be annoying to rewrite this whole switch case so maybe file a follow-up?

src/amo/sagas/reviews.js
```diff
@@ -298,7 +298,7 @@ function* deleteAddonReview({
       reviewId,
     });
 
-    yield put(unloadAddonReviews({ reviewId }));
+    yield put(unloadAddonReviews({ addonId, reviewId }));
     if (isReplyToReviewId) {
       yield put(unloadAddonReviews({ reviewId: isReplyToReviewId }));
```
kumar303: Why aren't you passing `addonId` here, too? This would also let you make `addonId` required which I think would help make the reducer a little more robust.

src/amo/components/AddonReviewList/index.js
bobsilverberg: Sure, I'll do that.

src/core/reducers/addons.js
bobsilverberg: I might just do it now. Let me see how much time I think it will take.

src/amo/sagas/reviews.js
bobsilverberg: I did this on purpose because we've already cleared the add-on's data in the call to `yield put(unloadAddonReviews({ addonId, reviewId }));` so there's no reason to clear it again. I was trying to be efficient.

I can see your reasoning, though. With it being optional a user might call `unloadAddonReviews` without it not realizing they they should be passing it in in order to unload the addon `too`. I'll change this.

src/amo/components/AddonReviewList/index.js
bobsilverberg: Hmm, This component doesn't actually have any `Props` that aren't internal. I added a type for `Props` that I defined as `{||}` to use for the component declaration.

src/core/reducers/addons.js
bobsilverberg: Actually, I'll file a follow-up. It's going to take some work to address this.

bobsilverberg: I opened #6147.

willdurand: How is it breaking the Redux devtools? `loadingBySlug` is a new object created L349.

kumar303: I took a closer look at this specific case and it's safe (I think) but this is a very dangerous pattern so we should avoid it. Here is one small change which would mutate the previous state -- it would be easy to introduce this by accident.

```js
const prevState = {
  byID: {
    '123': { name: 'first' },
  },
};

const nextState = {
  byID: {
    ...prevState.byID,
  },
};
  
nextState.byID['123'].name = 'second';
  
console.log('previous:', prevState.byID[123].name);
console.log('next:', nextState.byID[123].name);
```

Get the [babel translation here](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBADgJwKYDcDKUCGUkwLwwDeAUDDAEYCeAkgCIBcRpZMA5AIwBMAzK44TDCYAtkkasAZgEsE0VjAC-AGmbLiCgNzFioSLDBIAHlAzZcBEmSp1-zMgDpHiVKZz3rtFWTWbmxA8auSO40tADaHDysALr2QqL4bBBIugAmrFpkOuAQIAA2wXkgAOYAFKzOKFIgAK4Q9KxK8MjoWG4eYVzcsfFIAJRaurkF9kVlrAFQDU2TQSF0nTw9Iv0aQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=)

When you run this, you can see that the previous state was mutated:

screenshot 2018-09-06 09 55 59

This is because objects are passed by "copy of a reference," similar to Python.

PR #6114, opened 2018-08-31
Fix problem with "Deleting" message after deleting a reply

### LINKED ISSUE #6113: "Deleting my reply" button is not correctly displayed when the developer deletes and writes a new reply
STR:
1. Load AMO dev and as a developer reply to a review on the listed review page
2. Delete the reply and and click reply to this review
3. Write a reply again and click publish reply
4. Observe the "Delete my reply" button

Actual result:
Button is inactive and is visible as "Delete..."

Expected result:
Delete my reply button is blue and active

Notes:
This issue is reproducible on AMO dev with FF61(Win10 and Android 8.0)
Workaround: refresh page

delete

bobsilverberg:
Fixes #6113

### DIFF
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -495,9 +495,15 @@ export default function reviewsReducer(
         payload: { reviewId },
       } = action;
 
-      const newState = { ...state };
-      const reviewData = state.byId[reviewId];
+      const newState = {
+        ...state,
+        view: {
+          ...state.view,
+          [reviewId]: undefined,
+        },
+      };
 
+      const reviewData = state.byId[reviewId];
       if (reviewData) {
         const { addonId, addonSlug, userId } = reviewData;
         return {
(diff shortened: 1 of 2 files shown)


src/amo/reducers/reviews.js
```diff
@@ -495,9 +495,15 @@ export default function reviewsReducer(
         payload: { reviewId },
       } = action;
 
-      const newState = { ...state };
-      const reviewData = state.byId[reviewId];
+      const newState = {
+        ...state,
+        view: {
```
kumar303: Whoops, good catch here by QA

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -391,6 +391,24 @@ describe(__filename, () => {
       expect(state.view[reviewId]).toEqual(undefined);
     });
 
+    it('unloads cached view data for a reply', () => {
```
kumar303: When I read this test I was really confused because `reviewId` seemed no different than `replyId`. I think it would have helped me if this spec were more explicit, something like: 'it unloads cached view data even for deleted reviews'. After that, you could add a comment explaining how deleting a reply then adding another reply would enter this state.

Sadly, I think this bug was a side effect of deciding not to clean up after the deletion (https://github.com/mozilla/addons-frontend/pull/6084#discussion_r213735440). I still think that's fine, though.

PR #6119, opened 2018-09-02
Put existing visited collection in collections

### LINKED ISSUE #5930: Pagination is not displayed on the collection details page for a new collection with >25 add-ons
STR

1. Log in to AMO
2. Go to your Collections page and Create a new collection
3. Add more that 25 add-ons to your new collection
4. Click on View my Collections from the user menu and select the new collection to view its details
5. Scroll down to the bottom of the page and try to navigate to the next page

Actual result:
The pagination component is not present on the collection details

Expected result:
Pagination is present for a new collection with >25 add-ons

Notes:
- reproduced on all AMO servers with FF61, Win10x64
- issue is solved when the collection details page is reloaded

**this video depicts what happens in Steps 4 and 5**

no pagination

xu3u4:
Fixes #5930 

#### Description
When users go from collection A page(after refresh) to collections and back to collection A page, if addons number over 25, the pagination is disappeared.
The initial pageSize send from `LOAD_USER_COLLECTIONS` to `loadCollectionIntoState` is always null. So if the `existingCollection` is stored to collections, it can have all the existing data, including `pageSize`.

fixed

### DIFF
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -907,6 +907,7 @@ export const loadCollectionIntoState = ({
   // make sure we don't overwrite any existing addons.
   if (!internalCollection.addons && existingCollection) {
     internalCollection.addons = existingCollection.addons;
+    internalCollection.pageSize = existingCollection.pageSize;
   }
 
   return {
--- a/tests/unit/amo/reducers/test_collections.js
+++ b/tests/unit/amo/reducers/test_collections.js
@@ -702,6 +702,7 @@ describe(__filename, () => {
 
   describe('loadCollectionIntoState', () => {
     it('preserves existing collection addons', () => {
+      const fakePageSize = 5;
       const fakeCollectionAddon = createFakeCollectionAddon({
         addon: { ...fakeAddon, id: 1 },
       });
@@ -717,6 +718,7 @@ describe(__filename, () => {
         state: initialState,
         collection,
         addons,
+        pageSize: fakePageSize,
       });
 
       // Simulate loading it a second time but without addons.
@@ -724,6 +726,7 @@ describe(__filename, () => {
 
       const collectionInState = state.byId[collection.id];
       expect(collectionInState.addons).toEqual(createInternalAddons(addons));
+      expect(collectionInState.pageSize).toEqual(fakePageSize);
     });
 
     it('loads notes for collection add-ons', () => {


src/amo/reducers/collections.js
```diff
@@ -898,16 +898,14 @@ export const loadCollectionIntoState = ({
...
-    detail: collection,
-    items: addons,
-    pageSize,
-  });
-  // In case the new collection isn't loaded with add-ons,
-  // make sure we don't overwrite any existing addons.
-  if (!internalCollection.addons && existingCollection) {
-    internalCollection.addons = existingCollection.addons;
-  }
+  const internalCollection =
```
bobsilverberg: This works, but it always loads everything from the `existingCollection` if one exists and there are no `addons`, so it will ignore anything in the `collection` parameter, which may not be what we want.

It might be better to add a line into the `if` block at line 908 to also take `pageSize` from `existingCollection` and use it in `internalCollection` just as we've done with `addons`:

```diff
diff --git a/src/amo/reducers/collections.js b/src/amo/reducers/collections.js
index 5ee5cee..745a4ee 100644
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -898,14 +898,17 @@ export const loadCollectionIntoState = ({
   pageSize,
 }: LoadCollectionIntoStateParams): CollectionsState => {
   const existingCollection = state.byId[collection.id];
-  const internalCollection =
-    !addons && existingCollection
-      ? existingCollection
-      : createInternalCollection({
-          detail: collection,
-          items: addons,
-          pageSize,
-        });
+  const internalCollection = createInternalCollection({
+    detail: collection,
+    items: addons,
+    pageSize,
+  });
+  // In case the new collection isn't loaded with add-ons,
+  // make sure we don't overwrite any existing addons.
+  if (!internalCollection.addons && existingCollection) {
+    internalCollection.addons = existingCollection.addons;
+    internalCollection.pageSize = existingCollection.pageSize;
+  }

   return {
     ...state,
```

Also, the test for the reducer should be updated to include a test for this new use case.

xu3u4: Ok, I see.
Sorry, my computer was broken 😭
I will fix it soon.

xu3u4: @bobsilverberg 
Thanks for your advise.
I have updated it. 
Please check.

PR #6135, opened 2018-09-04
Extract/Merge/Debug Amo locales

bobsilverberg:


smayer97:
Does this mean that this will fix a "bug" where comments in collections were only pulling up if you are using the same locale version as when the comment was created, e.g. create comment with US locale, then access collection with GB locale and comments would not pull up, but would will pull up using the US locale?

(sorry I do not have the reference but there was an entry in this database for this issue).

bobsilverberg:
@smayer97 No, this has nothing to do with that. This just sets up the files for localizers.

smayer97:
Ok thx.

Issue #4935, opened 2018-05-04
Language packs listed on AMO are not compatible with latest Firefox release version


AlexandraMoga:
STR:

1. Open AMO homepage
2. Navigate to the language tools listing page (https://addons.mozilla.org/en-US/firefox/language-tools/)
3. Select one of the available Language Packs to open its details page. **EDIT**: The Language Pack must be configured to *only* support a version of Firefox higher than the one you are using. 

Actual result:
Language pack compatibility is set to Firefox 60.0 as min compatible version

Expected result:
Language packs listed on AMO are compatible with the latest release version of Firefox

Notes:
- reproduced on AMO prod with FF59, WIn10x64

![lg packs](https://user-images.githubusercontent.com/31961530/39629010-77209842-4fb3-11e8-9bef-c8cad43a8778.gif)

willdurand:
xref: #8180

diox:
Note that specifically for langpacks, the [language tools API](http://addons-server.readthedocs.io/en/latest/topics/api/addons.html#language-tools) does provide a way to return only compatible versions, through the `appversion` parameter. It's not currently available for the detail API however, so using this would require refactoring the language tools page to directly download the language packs from that page.

Firefox is going to use this API to display available language packs directly from the browser, FYI.

kumar303:
This is going to keep happening for every Firefox release due to changes in the release engineering team. See https://github.com/mozilla/addons-server/issues/8180#issuecomment-417406841

I think we need to come up with a workaround.

Maybe we can just show a second banner (under the error) linking to the previous, compatible version of the XPI? In other words, a quick and dirty hack that fetches [previous versions](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#versions-list) and filters by `appversion` to get the XPI link. 

The goal would be to let people install their language pack in a somewhat obvious way.

diox:
Once https://github.com/mozilla/addons-server/issues/8180 lands **the only thing needed on the frontend will be to send `app` and `appversion` parameters when doing a detail API request**. Then, *only for langpacks*, the API will overwrite `current_version` in the result, returning the latest compatible public version according to the parameters passed.

kumar303:
I tested this locally with https://addons-dev.allizom.org/en-US/firefox/addon/finn-language-pack/ and it is doing the right thing 💃

AlexandraMoga:
[...]

PR #6138, opened 2018-09-04
Show compatible langpacks based on app version when possible

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/4935

~~In the interest of time and simplicity, I did not actually test this by setting up multiple versions of language packs.~~ I was able to test this with http://localhost:3000/en-US/firefox/addon/finn-language-pack/ (see https://github.com/mozilla/addons-server/issues/8180#issuecomment-418687835)

Also, I opened up a can of worms here. The moving of `refreshAddon` was to fix a circular dependency and the new polyfill code was probably always a bug but I think moving `refreshAddon` made it worse.

### DIFF
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -11,7 +11,8 @@ import {
   setInternalReview as _setInternalReview,
   setReview,
 } from 'amo/actions/reviews';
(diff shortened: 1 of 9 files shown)


kumar303:
Thanks for the review

Issue #5527, opened 2018-07-09
upgrade api from v3 to v4


rebmullin:
upgrade api from v3 to v4

---

- [x] 2018-05-18: renamed /reviews/ endpoint to /ratings/ https://github.com/mozilla/addons-server/issues/6849
- [x] 2018-05-25: renamed rating.rating property to rating.score https://github.com/mozilla/addons-server/pull/8332
- [x] 2018-06-05: dropped rating.title property https://github.com/mozilla/addons-server/issues/8144
- [x] 2018-07-12: added type property to autocomplete API. This change was also backported to the v3 API. https://github.com/mozilla/addons-server/issues/8803
- ~~2018-07-19: localised field values are always returned as objects, even if only a single language is requested. Setting a localised value with a string is removed too - it must always be an object of one or more translations. https://github.com/mozilla/addons-server/issues/8794~~
- [x] 2018-07-18: added previews property to discovery API addons object. This change was also backported to the v3 API. https://github.com/mozilla/addons-server/issues/8863
- [x] 2018-07-20: dropped downloads property from the collection add-ons results. https://github.com/mozilla/addons-server/issues/8944
- [x] 2018-08-16: added is_developer_reply property to ratings. This changed was also backported to the v3 API. https://github.com/mozilla/addons-server/issues/8993
- [x] 2018-09-13: added name and icon_url properties to the addon object in ratings. This changed was also backported to the v3 API. https://github.com/mozilla/addons-server/issues/9357
- [x] 2018-09-27: backed out “localised field values are always returned as objects” change from 2018-07-19 from v4 API. This is intended to be temporary change while addons-frontend upgrades. On addons-dev and addons stage enviroments the previous behavior is available as api/v4dev. The v4dev api is not available on AMO production server. https://github.com/mozilla/addons-server/issues/9467

kumar303:
We need to do this. Today, when new features land in the API they have to be *backported* to v3 if we want to use them. This is unsustainable.

The v4 changelog is a fairly small list: https://addons-server.readthedocs.io/en/latest/topics/api/overview.html#v4-api-changelog

We should implement the ability to downgrade a single endpoint to v3 in case any hard to fix bug comes up.

Also, related: https://github.com/mozilla/addons-server/issues/8899

willdurand:
Most new features have been backported and BC breaks are relatively easy, except this one:

[...]

PR #6142, opened 2018-09-05
Add version param to callApi

willdurand:
Refs #5527 

---

This PR adds a `version` parameter to the `callApi` function, used by most higher-level API calls, along with a `apiVersion` config parameter. This will ease the API version upgrade described in #5527.

### DIFF
--- a/tests/unit/core/api/test_index.js
+++ b/tests/unit/core/api/test_index.js
@@ -414,12 +414,13 @@ describe(__filename, () => {
     });
 
     it('uses the config version as default value', async () => {
-      const _config = getFakeConfig({ apiVersion: '456' });
+      const apiVersion = '456';
+      const _config = getFakeConfig({ apiVersion });
       const endpoint = '/some-endpoint/';
 
       mockWindow
         .expects('fetch')
-        .withArgs(sinon.match('/api/456/some-endpoint/'))
+        .withArgs(sinon.match(`/api/${apiVersion}/some-endpoint/`))
         .returns(createApiResponse());
 
       await api.callApi({ _config, endpoint });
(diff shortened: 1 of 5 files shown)


tests/unit/core/api/test_index.js
```diff
@@ -399,6 +399,32 @@ describe(__filename, () => {
...
+        .expects('fetch')
+        .withArgs(sinon.match(`/api/${version}/some-endpoint/`))
+        .returns(createApiResponse());
+
+      await api.callApi({ endpoint, version });
+      mockWindow.verify();
+    });
+
+    it('uses the config version as default value', async () => {
+      const _config = getFakeConfig({ apiVersion: '456' });
```
bobsilverberg: Nit: `const apiVersion = 456;` so you don't have to hardcode it below.

tests/unit/core/api/test_index.js
```diff
@@ -546,28 +572,38 @@ describe(__filename, () => {
...
-      expect(getStartLoginQs(location)).toEqual({ to: '/foo?bar=BAR' });
+      expect(getStartLoginQs({ location })).toEqual({ to: '/foo?bar=BAR' });
     });
 
     it('includes the next path the config if set', () => {
-      sinon
-        .stub(config, 'get')
-        .withArgs('fxaConfig')
-        .returns('my-config');
+      const _config = getFakeConfig({ fxaConfig: 'my-config' });
```
bobsilverberg: Nit: use `const`s here to avoid hardcoding the same values in the assertion.

tests/unit/core/api/test_index.js
```diff
@@ -546,28 +572,38 @@ describe(__filename, () => {
...
-      expect(getStartLoginQs(location)).toEqual({
+
+      expect(getStartLoginQs({ _config, location })).toEqual({
         to: '/foo',
         config: 'my-config',
       });
     });
+
+    it('uses the API version from config', () => {
+      const _config = getFakeConfig({ apiVersion: 'v789' });
```
bobsilverberg: Nit: use a `const` for `apiVersion`.

PR #6150, opened 2018-09-06
Remove spreads in Addon components

### LINKED ISSUE #1416: Don't spread add-on and installation in mapStateToProps

willdurand:
Fix #1416

---

- [x] remove spreads in `Addon` (amo)
- [x] remove spreads in `Addon` (disco)

### DIFF
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -313,6 +313,7 @@ export class AddonBase extends React.Component {
             />
           ) : (
             <InstallButton
+              {...this.props.addon}
               {...this.props}
               className="Addon-install-button"
               defaultInstallSource={defaultInstallSource}
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -29,7 +29,6 @@ import {
 } from 'core/constants';
 import translate from 'core/i18n/translate';
 import { withInstallHelpers } from 'core/installAddon';
-import { getAddonByGUID } from 'core/reducers/addons';
 import tracking, { getAddonTypeForTracking } from 'core/tracking';
 import { isTheme } from 'core/utils';
 import { sanitizeHTMLWithExternalLinks } from 'disco/utils';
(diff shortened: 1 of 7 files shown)


src/disco/components/Addon/index.js
```diff
@@ -327,16 +335,13 @@ export function mapStateToProps(state, ownProps) {
   // `ownProps.guid` is already "normalized" with `getGuid()` in the
   // `DiscoPane` container component.
   const installation = state.installations[ownProps.guid] || {};
-  const addon = getAddonByGUID(state, ownProps.guid);
```
willdurand: This is not needed because `DiscoPane` passes the `addon` object to each `Addon` component.

kumar303:
Could you take a look at the conflicts?

tests/unit/amo/components/TestAddon.js
```diff
@@ -1623,17 +1623,6 @@ describe(__filename, () => {
...
-      const description = 'whatever';
-      fetchAddon({ addon: { ...fakeAddon, description } });
-      const props = _mapStateToProps();
-
-      // Make sure a random addon prop gets passed as a component prop
-      // so that the withInstallHelpers HOC works.
-      expect(props.description).toEqual(description);
-    });
-
     it('must convert all installed addon props to component props', () => {
```
bobsilverberg: This doesn't seem like a valid description for this test case anymore.

tests/unit/amo/components/TestAddon.js
```diff
@@ -1644,11 +1633,11 @@ describe(__filename, () => {
           status: INSTALLED,
         }),
       );
-      const { needsRestart } = _mapStateToProps();
+      const { installStatus } = _mapStateToProps();
 
       // Make sure a random installedAddon prop gets passed as a component prop
       // so that the withInstallHelpers HOC works.
-      expect(needsRestart).toEqual(false);
+      expect(installStatus).toEqual(INSTALLED);
```
bobsilverberg: Nit: set a `const` for `INSTALLED` at the top of the test.

tests/unit/amo/components/TestAddon.js
willdurand: oh, good catch!

PR #6151, opened 2018-09-06
Add tracking for INSTALL_DOWNLOAD_FAILED event

### LINKED ISSUE #6099: There is no analytics event sent when a failure is received during installation
Follow up for #5940 

STR:
1. Open your Firefox browser
2. Go to about:config and set `xpinstall.signatures.dev-root` to `false`
3. Go to AMO -dev and open the browser console
4. Install an extension
5. Observe the analytics events sent

Actual result:
There is no event sent mentioning the installation failure

Expected result:
An event that reports the installation failure should be sent

Notes:
- reproduced on AMO -dev with FF61, Win10x64

willdurand:
Fix #6099

---

In case of a (download) error, a GA event is now sent. See STR in #6099.


willdurand:
thanks!

PR #6153, opened 2018-09-06
Fix setCurrentStatus() call

### LINKED ISSUE #6152: AMInstallButton status is not updated

1. Go to https://addons-dev.allizom.org/en-US/firefox/addon/enhancer-for-youtube/ with a browser that has `mozAddonManager`
2. Click on the "Search by Image" add-on located in the "Other popular extensions"

### What happened?

The `AMInstallButton` (new button) is disabled:

screen shot 2018-09-06 at 15 35 21

The button is NOT disabled.

Could be a regression related to #6094.

willdurand:
Fix #6152

### DIFF
--- a/tests/unit/core/test_installAddon.js
+++ b/tests/unit/core/test_installAddon.js
@@ -200,6 +200,27 @@ describe(__filename, () => {
     sinon.assert.calledWith(_addonManager.getAddon, '@new-guid');
   });
 
+  it('sets status when add-on is loaded on update', () => {
+    const Component = componentWithInstallHelpers();
+    const _addonManager = getFakeAddonManagerWrapper({
+      getAddon: Promise.resolve({
+        isActive: true,
+        isEnabled: true,
+      }),
+    });
+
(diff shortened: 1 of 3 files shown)


tests/unit/core/test_installAddon.js
```diff
@@ -200,6 +200,27 @@ describe(__filename, () => {
     sinon.assert.calledWith(_addonManager.getAddon, '@new-guid');
   });
 
+  it('sets status when add-on is loaded on update', () => {
+    const Component = componentWithInstallHelpers();
+    const _addonManager = getFakeAddonManagerWrapper({
+      getAddon: Promise.resolve({
```
kumar303: It's not necessary for this patch but at some point this should really resolve the final object using a helper. I see that sometimes [`type` is passed](https://github.com/mozilla/addons-frontend/blob/6a6a293a48620b477376e407e72f264562ad557c/tests/unit/core/test_installAddon.js#L163) so a helper would be better to manager the return value.

willdurand:
thanks for the quick review!

PR #6157, opened 2018-09-06
I6018 hero setup w/ab test

### LINKED ISSUE #6018: Hero area in homepage is too tall for common screen resolutions
Here's how the hero area looks in a 1280x720 viewport:
screen shot 2018-08-22 at 11 15 16
So, even if the window is maximized, most of the AMO homepage real estate is being taken over by the hero area, which is also not very informative and has been shown to confuse some users.

@pwalm will be looking into new designs for this for Q4, but @kumar303 suggested we could do something more short-term by maybe detecting the resolution and scaling the hero area down for those cases. Let's see what can be done and decide if it's a good stopgap measure.

/cc @devaneymoz

rebmullin:
fixes #6018 

This is just a start/proposal for ab hoc. It seems like ab testing will always have /need a cookie, so I set it up a round that. Let me know if that is the wrong assumption.. 
Also, I ran into some issues (like with passing in vals to this) so I welcome early feedback : )

note: this still needs tests


src/core/components/Experiment/index.js
```diff
@@ -0,0 +1,158 @@
...
+  BName: string,
+|}) {
+  invariant(nameId, 'nameId is required');
+  invariant(AName, 'AName is required');
+  invariant(BName, 'BName is required');
+
+  return (WrappedComponent: Function) =>
+    compose(
+      connect(
+        _mergeExperimentProps({ WrappedComponent, nameId, AName, BName }),
```
willdurand: I am wondering why you connect this component here?

rebmullin: I was having some issues getting it to pass in values without some similar function. I wouldn't be surprised if there was better to do this but I saw something similar to this in another HOC and this did what I needed it to do :). The key thing was passing in the params...

I can try cleaning this up a bit more but I still connect it seems so the values will be available.

willdurand: I can give it a try but the simpler the better, and I don't see why we need to connect here. We don't use redux.

willdurand: I think you can move the whole HOC class (`Experiment`) into this function, that would give you access to all the props you want from the HOC function.

Maybe we can rename it `withExperiment`. Also, I don't think we put HOCs in the `components/` folder. If you rename the entire file with `withExperiment.js`, then you can move it to `src/core/`.

That's a really good start, once cosmetics changes have been made, we can start writing tests :)

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,58 +1,91 @@
...
+export const AB_HOME_HERO_VARIANT_A = 'ScaledDown';
+export const AB_HOME_HERO_VARIANT_B = 'Large';
+
 export class HomeHeroBannerBase extends React.Component<Props> {
   sections() {
-    const { i18n } = this.props;
-
+    const { i18n, trackClick } = this.props;
+    // TODO: make this into an array that you can map through.
+    // This might help eliminate manual prop set up / redundancy.
```
willdurand: Yep, doing this before your PR would reduce the diff of this one.

src/core/components/Experiment/index.js
```diff
@@ -0,0 +1,158 @@
...
+import type { AppState } from 'amo/store';
+
+type Props = {|
+  AName: string,
+  BName: string,
+  nameId: string,
+|};
+
+type UIStateType = {|
+  abTest: string | void,
```
willdurand: Why do you need a ui state btw? And why a "ui state" and not a regular redux state?

src/core/components/Experiment/index.js
```diff
@@ -0,0 +1,158 @@
...
+    const { WrappedComponent, _cookie, nameId, uiState, ...props } = this.props;
+
+    invariant(nameId, 'nameId is required');
+
+    const exposedPropHelpers = {
+      trackClick: (...args) => this.trackClick(...args),
+    };
+
+    // When the cookie is initially set it will not be seen on client
+    // until there is a refresh so we are also checking uiState.
```
willdurand: `componentDidMount` is only executed on the client, so we will never set a cookie on the server if I understand correctly. Do you really need this? That was not present before AFAIK and that would simplify lots of things here.

rebmullin: I tried doing this in componentWillMount but I thought I saw a complaint about the classes not matching (on server and client). I can move it back/try that again tho...

To be clear tho - do I really need what? :) the uiState?

src/core/components/Experiment/index.js
rebmullin: well I *think* this would have been a case where I used regular state (b/c of situation where class wasn't assigned on client on initial visit) here but thought we were trying to use UIState instead..? I'll revisit this..

src/amo/components/HomeHeroBanner/index.js
rebmullin: oh? I thought it should be a separate PR..? Sure tho, I can update here :)

src/core/components/Experiment/index.js
rebmullin: hmm..I tried that before but couldn't get it (note: this is a case where I have to restart the server every time it's updated/wrong :/)...but I'll recheck this one too:)

Yep - that makes sense. I'll move things around as you suggest, thanks.

src/core/components/Experiment/index.js
rebmullin: hmm I guess I don't see this now (about the class mismatch so I will it back and then this will resolve - i believe - the client issue noted above)

src/amo/components/HomeHeroBanner/index.js
willdurand: Yes, that would be nice as a separate PR but still merged before this PR here :)

rebmullin: new PR for this #6180 : )

src/ui/components/Hero/styles.scss
```diff
@@ -40,6 +40,19 @@
   }
 }
 
+.HomeHeroBanner--experiment-is-on {
```
willdurand: I think this should have a more meaningful name, explaining what it does rather than why it is displayed.

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+      if (this.abTestCookie) {
+        _tracking.sendEvent({
+          action: `${abNameId} Page View`,
+          category: `AMO ${this.abTestCookie}`,
+          label: '',
+        });
+      }
+    }
+
+    trackClick = (e: SyntheticEvent<any>, url: string = '') => {
```
willdurand: I am not sure the HOC should do that, because it is up to the wrapped component to decide. For instance, what if we want to have a different action or label?

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+
+      this.abTestCookie = _cookie.load(`AB_${abNameId}_COOKIE`);
+
+      if (this.abTestCookie === undefined) {
+        this.abTestCookie =
+          randomizer() >= 0.5
+            ? `AB_TEST_${abNameId}_${ANameVariant}`
+            : `AB_TEST_${abNameId}_${BNameVariant}`;
+        _cookie.save(`AB_${abNameId}_COOKIE`, this.abTestCookie, {
+          path: '/', // TODO: make this flexible too possibly.
```
willdurand: That could be another prop passed to `withExperiment`, with default values.

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+  _tracking: typeof tracking,
+  randomizer: () => number,
+  WrappedComponent: Function,
+|};
+
+export const withExperiment = ({ nameId, AName, BName }: Props) => (
+  WrappedComponent: Function,
+) => {
+  class WithExperiment extends React.Component<InternalProps> {
+    abTestCookie: string | void;
```
willdurand: could be simply `experimentCookie`

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+import * as React from 'react';
+import invariant from 'invariant';
+import cookie from 'react-cookie';
+
+import tracking from 'core/tracking';
+
+type Props = {|
+  AName: string,
+  BName: string,
+  nameId: string,
```
willdurand: could it be simply `id`?

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
+/* @flow */
+import * as React from 'react';
+import invariant from 'invariant';
+import cookie from 'react-cookie';
+
+import tracking from 'core/tracking';
+
+type Props = {|
+  AName: string,
```
willdurand: could it be `variantA` and `variantB`?

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+        ...props
+      } = this.props;
+
+      // We'll call the "AName" variant the "on" / "true" variant.
+      const isOn =
+        _cookie.load(`AB_${abNameId}_COOKIE`) ===
+        `AB_TEST_${abNameId}_${ANameVariant}`;
+
+      const exposedPropHelpers = {
+        trackClick: (...args) => this.trackClick(...args),
```
willdurand: I think the HOC should not do that, but rather pass a prop that tells the wrapper component which variant it is in.

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+
+      invariant(ANameVariant, 'AName is required');
+      invariant(BNameVariant, 'BName is required');
+      invariant(abNameId, 'nameId is required');
+
+      this.abTestCookie = _cookie.load(`AB_${abNameId}_COOKIE`);
+
+      if (this.abTestCookie === undefined) {
+        this.abTestCookie =
+          randomizer() >= 0.5
```
willdurand: this should probably be a prop of `withExperiment`, so that it is not a HOC that does 50-50, but it can probably be done in a follow-up issue.

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+        _cookie,
+        AName: ANameVariant,
+        nameId: abNameId,
+        ...props
+      } = this.props;
+
+      // We'll call the "AName" variant the "on" / "true" variant.
+      const isOn =
+        _cookie.load(`AB_${abNameId}_COOKIE`) ===
+        `AB_TEST_${abNameId}_${ANameVariant}`;
```
willdurand: If you pass the variant name/id, then you don't need another prop.

src/core/withExperiment.js
willdurand: we can also probably extract the cookie name in a method.

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+    static defaultProps = {
+      _cookie: cookie,
+      _tracking: tracking,
+      AName,
+      BName,
+      nameId,
+      randomizer: Math.random,
+    };
+
+    componentWillMount() {
```
willdurand: ideally, we should not use that anymore. maybe `componentDidMount()`?

willdurand: or do you need the server side?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,18 +1,26 @@
...
 
 type Props = {|
   i18n: I18nType,
+  trackClick: () => void,
+  experimentIsOn: boolean,
 |};
 
+export const AB_HOME_HERO_TEST_NAME = 'HOME_HERO';
+export const AB_HOME_HERO_VARIANT_A = 'ScaledDown';
+export const AB_HOME_HERO_VARIANT_B = 'Large';
```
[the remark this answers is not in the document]
rebmullin: I was just doing by this: https://github.com/mozilla/addons-frontend/issues/6018#issuecomment-417731488

src/ui/components/Hero/styles.scss
rebmullin: hmm I thought it would be better to generalize it but sure I can update..

src/core/withExperiment.js
rebmullin: ok, I wasn't sure about this but thought it *could* be useful. I will update :)

src/core/withExperiment.js
rebmullin: ok, I was wondering about this one too but figured 50-50 was common with these types of test

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
willdurand: yeah, I meant for the terms to use.

src/core/withExperiment.js
willdurand: no need to change that right now

src/ui/components/Hero/styles.scss
willdurand: My idea is: if we keep this style, then we'll have nothing to update here once we get rid of the experiment.

rebmullin: I see ..that makes sense

src/core/withExperiment.js
```diff
@@ -0,0 +1,114 @@
...
+        this.abTestCookie =
+          randomizer() >= 0.5
+            ? `AB_TEST_${abNameId}_${ANameVariant}`
+            : `AB_TEST_${abNameId}_${BNameVariant}`;
+        _cookie.save(`AB_${abNameId}_COOKIE`, this.abTestCookie, {
+          path: '/', // TODO: make this flexible too possibly.
+        });
+      }
+
+      if (this.abTestCookie) {
```
rebmullin: hey @willdurand , just checking..you think this too should be removed from the experiment Hoc?

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
rebmullin: oh @willdurand, my understanding is it should be all viewports if the experiment is "on" (I will verify this to be sure). But does that answer your question?

src/core/withExperiment.js
rebmullin: okie..

rebmullin:
@willdurand ++, thank you for all the help with this one 👍 

I think I've made all the updates. can you please take another look?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -206,26 +236,47 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
-        <HeroSection key={hero.url} linkTo={hero.url}>
+        <HeroSection key={url} linkTo={url} onClick={this.trackExperimentClick}>
           <h3>{hero.title}</h3>
           <p>{hero.description}</p>
         </HeroSection>
       );
     });
   }
 
+  trackExperimentClick = (e: SyntheticEvent<any>, url: string) => {
```
rebmullin: I am using this name so we remember to remove this once the experiment is over...

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
...
+|};
+
+type Props = {|
+  cookieConfig?: Object,
+  id: string,
+  variantA: string,
+  variantB: string,
+|};
+
+type InternalProps = {|
```
rebmullin: I noticed you renamed this but I wasn't sure why. Anyway, so I left for now but wanted to double check with you on this since we usually keep this name or let me know if I am missing something

rebmullin:
also, just wanted to make sure this was ok - that the cookie is set to expire on session. It seems you actually have to "Quit" the browser for it to be removed tho. Is this ok or should we be explicitly expiring this at some later date?

src/core/withExperiment.js
willdurand: I renamed this type to be more explicit. `InternalProps` is nice when it is in a "component" file, with only a component definition (class or function) in it. In this file, there is mainly a HOC (that uses a component class).

willdurand:
> also, just wanted to make sure this was ok - that the cookie is set to expire on session. It seems you actually have to "Quit" the browser for it to be removed tho. Is this ok or should we be explicitly expiring this at some later date?

mmh, I think it should never expire, no?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,49 @@
...
 import HeroSection from 'ui/components/HeroSection';
 import type { I18nType } from 'core/types/i18n';
+import tracking from 'core/tracking';
+import { withExperiment } from 'core/withExperiment';
 
 import './styles.scss';
 
 type Props = {|
   i18n: I18nType,
+  trackExperimentClick: () => void,
```
willdurand: I think this is not needed anymore because it is part of this component.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,49 @@
...
+  componentDidMount() {
+    const { _tracking, variant } = this.props;
+
+    if (variant) {
+      _tracking.sendEvent({
+        action: `${AB_HOME_HERO_TEST_NAME} Page View`,
+        category: `AMO ${AB_HOME_HERO_TEST_NAME}_EXPERIMENT: ${variant}`,
+        label: '',
+      });
+    }
```
willdurand: Do we actually want to send page view events?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -206,26 +236,47 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
   }
 
+  trackExperimentClick = (e: SyntheticEvent<any>, url: string) => {
+    const { _tracking, variant } = this.props;
+
+    _tracking.sendEvent({
+      action: `${AB_HOME_HERO_TEST_NAME} Click`,
+      category: `AMO ${AB_HOME_HERO_TEST_NAME}_EXPERIMENT: ${variant}`,
+      label: url,
+    });
```
willdurand: same comments above apply to this `sendEvent` call

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
+/* @flow */
+import * as React from 'react';
+import invariant from 'invariant';
+import cookie from 'react-cookie';
+
+export type ExposedWithExperimentProps = {|
```
willdurand: Sorry for having given you the wrong name. I double checked and we already use another name pattern that is: `<HOC>InjectedProps`, so `WithExperimentInjectedProps`.

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
...
+type Props = {|
+  cookieConfig?: Object,
+  id: string,
+  variantA: string,
+  variantB: string,
+|};
+
+type InternalProps = {|
+  ...Props,
+  WrappedComponent: Function,
```
willdurand: Can we use `React.ComponentType<any>`?

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
...
+|};
+
+const defaultCookieConfig = { path: '/' };
+
+export const withExperiment = ({
+  cookieConfig = defaultCookieConfig,
+  id: defaultId,
+  variantA: defaultVariantA,
+  variantB: defaultVariantB,
+}: Props) => (WrappedComponent: Function) => {
```
willdurand: Can we use `React.ComponentType<any>`?

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    constructor(props: InternalProps) {
+      super(props);
+
+      const { _cookie, id, randomizer, variantA, variantB } = this.props;
+
+      invariant(id, 'id is required');
+      invariant(variantA, 'variantA is required');
+      invariant(variantB, 'variantB is required');
```
willdurand: I think, if you move those `invariant` calls above the class, then we'll be able to validate the HOC call and we would not need those calls here (if that makes sense).

I'd put the `invariant` above the class, on the `default*` variables.

src/core/withExperiment.js
```diff
@@ -0,0 +1,76 @@
...
+        _cookie.save(this.getCookieName(), this.experimentCookie, cookieConfig);
+      }
+    }
+
+    getCookieName() {
+      return `experiment_${this.props.id}`;
+    }
+
+    render() {
+      const { _cookie, ...props } = this.props;
```
willdurand: we could filter the passed props a bit by removing the props used by this HOC:

```
const {
  _cookie,
  variantA,
  variantB,
  id,
  randomizer,
 ...props,
} = this.props;
```

That way, wrapped component won't receive the following props: id, variantA, variantB, randomizer.

src/ui/components/Hero/styles.scss
```diff
@@ -40,6 +40,19 @@
   }
 }
 
+.HomeHeroBanner--scaled-down {
```
willdurand: `--small`?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,49 @@
...
 
-export class HomeHeroBannerBase extends React.Component<Props> {
+type InternalProps = {|
+  ...Props,
+  _tracking: typeof tracking,
+|};
+
+export const AB_HOME_HERO_TEST_NAME = 'HOME_HERO';
+export const AB_HOME_HERO_VARIANT_A = 'ScaledDown';
+export const AB_HOME_HERO_VARIANT_B = 'Large';
```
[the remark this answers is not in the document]
willdurand: also, do we have a common language (with product) for those values? home hero is fine, but for the variants? We could maybe go for "small" and "large".

src/ui/components/HeroSection/index.js
```diff
@@ -32,6 +33,7 @@ export default class HeroSection extends React.Component<Props> {
         {linkTo ? (
           <Link
             className="HeroSection-link-wrapper"
+            onClick={(e) => this.props.onClick(e, linkTo)}
```
willdurand: We could also pass `onClick` directly here, and `bind` the `onExperimentClick` method in the parent component. That would be better because otherwise, how do we know that `linkTo` will be passed to the parent component here?

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -33,4 +40,55 @@ describe(__filename, () => {
     expect(carousel).toHaveLength(1);
     expect(carousel).toHaveProp('random', true);
   });
+
+  it('calls tracking on home page view', () => {
```
willdurand: not sure we want this because we want to track user engagement. This will only show a 50-50 result (given that we should show the small hero UI to 50% of the users)

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -33,4 +40,55 @@ describe(__filename, () => {
...
+      label: '',
+    });
+
+    sinon.assert.calledOnce(fakeTracking.sendEvent);
+
+    const heroBanner = root.find(Hero);
+
+    // We'll use the first item in the home heroes array as an example.
+    const heroItem = heroBanner.prop('sections')[0];
+    const heroLink = shallow(heroItem).find('.HeroSection-link-wrapper');
```
willdurand: Why not using `find()` and `.at(0)`?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable react/no-multi-comp, react/prop-types */
```
willdurand: Why is `react/prop-types` needed?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,74 @@
...
+
+    function componentWithExperiment({ experimentProps, props } = {}) {
+      const TestComponent = withExperiment({
+        id: 'ABtest',
+        variantA: 'AName',
+        variantB: 'BName',
+        ...experimentProps,
+      })(SomeComponentBase);
+
+      return shallow(<TestComponent {...props} />);
```
willdurand: Why not using `shallowUntilTarget()` and look for `SomeComponentBase`?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,74 @@
...
+describe(__filename, () => {
+  describe('withExperiment', () => {
+    class SomeComponentBase extends React.Component {
+      render() {
+        return <div className="component" />;
+      }
+    }
+
+    function componentWithExperiment({ experimentProps, props } = {}) {
+      const TestComponent = withExperiment({
```
willdurand: It should be `const SomeComponent =`

src/amo/components/HomeHeroBanner/index.js
rebmullin: Well what if they never click on a hero? how we will know which version they are looking at?

src/core/withExperiment.js
rebmullin: oh?! ok sure I'll take a look..

src/core/withExperiment.js
rebmullin: ok that makes sense

tests/unit/core/test_withExperiment.js
rebmullin: oops..I'll remove : )

src/amo/components/HomeHeroBanner/index.js
willdurand: we compare clicks between the two variants?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,49 @@
...
+  static defaultProps = {
+    _tracking: tracking,
+  };
+
+  componentDidMount() {
+    const { _tracking, variant } = this.props;
+
+    if (variant) {
+      _tracking.sendEvent({
+        action: `${AB_HOME_HERO_TEST_NAME} Page View`,
```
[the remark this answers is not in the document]
rebmullin: ok sure..I'll revisit this

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
rebmullin: 👍  I will update

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,49 @@
...
+        category: `AMO ${AB_HOME_HERO_TEST_NAME}_EXPERIMENT: ${variant}`,
```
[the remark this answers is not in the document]
rebmullin: will this work for category: `AMO HOME_HERO Experiment`

src/amo/components/HomeHeroBanner/index.js
rebmullin: hey @willdurand, this is what I am referring to: https://github.com/mozilla/addons-frontend/issues/6018#issuecomment-418814678

src/ui/components/HeroSection/index.js
rebmullin: hmm.I thought I tried something like that but it wasn't working. Let me recheck this tho..

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
rebmullin: I like it 👍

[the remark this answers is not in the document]
rebmullin: I was stuck on using the var here but I guess we don't need to

src/amo/components/HomeHeroBanner/index.js
willdurand: oh shoot, right.. we also want to measure bounce rate.

tests/unit/core/test_withExperiment.js
rebmullin: I don't see the variant prop when I use shallowUntilTarget:

```javascript
const root = componentWithExperiment();
const componentBase = root.find('.component');

// this is ok
expect(componentBase).toHaveLength(1);

// this one fails
expect(componentBase).toHaveProp('variant');
```
am I missing something?

src/amo/components/HomeHeroBanner/index.js
willdurand: sorry, yes so this is needed. we want to test both engagement (clicking on links) and bounce rate (users stay on the site instead of abandoning it)

src/ui/components/HeroSection/index.js
willdurand: I'd say we have to make `onExperimentClick` a non-bound method:

```js
onExperimentClick() {
}
```

and call `bind` in the `onClick` prop:

```
onClick={this.onExperimentClick.bind(this, url)}
```

tests/unit/core/test_withExperiment.js
willdurand: What if you find `SomeComponentBase` instead?

tests/unit/amo/components/TestHomeHeroBanner.js
rebmullin: hey @willdurand, can you clarify where you mean?

willdurand: ahm, nevermind. I did not read the code correctly and missed the line 76...

I am not sure we want to do that though. Can't we just get the `onClick` prop of the first item and assert that?

rebmullin: I am trying to simulate the click on the hero's respective link ('.HeroSection-link-wrapper'). It doesn't seem to work without. Is this what you mean? I had tried I think what you're mentioning here earlier too but it didn't work:

```bash
 TypeError: heroItem.simulate is not a function

      77 |     const heroItem = heroBanner.prop('sections')[0];
      78 | 
    > 79 |     heroItem.simulate('click', createEvent());
```


I *think* b/c the onclick is passed down to the inner component..?

the following seems to work? is this any better:

```javascript
const heroItem = heroBanner.prop('sections')[0];
heroItem.props.onClick(); 
```

tests/unit/core/test_withExperiment.js
rebmullin: I tried that too and still doesn't work:

```javascript
const root = componentWithExperiment();
const componentBase = root.find(SomeComponentBase);

// this fails (says it should be 0)
expect(componentBase).toHaveLength(1);
...
```

does that look right? Maybe I need to adjust something in the setup function..?

src/amo/components/HomeHeroBanner/index.js
rebmullin: so how I have this then is ok?

src/ui/components/HeroSection/index.js
rebmullin: for some reason bind wasn't working for me. I have it working in a slightly different way. let me know if that way is ok..

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,72 @@
...
+        id: 'ABtest',
+        variantA: 'AName',
+        variantB: 'BName',
+        ...experimentProps,
+      })(SomeComponentBase);
+
+      return shallow(<SomeComponent {...props} />);
+    }
+
+    it('passes the variant prop', () => {
```
willdurand: s/passes the/injects a

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,72 @@
...
+      const root = componentWithExperiment();
+      expect(root).toHaveProp('variant');
+    });
+
+    it('loads and saves cookie', () => {
```
willdurand: we should split this test case into multiple test cases:

- it loads a cookie upon construction
- it creates a cookie upon construction if none has been loaded
- it does not create a cookie upon construction if one has been loaded

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,72 @@
...
+      sinon.assert.calledWith(_cookie.load, `experiment_${id}`);
+
+      sinon.assert.calledWith(
+        _cookie.save,
+        `experiment_${id}`,
+        root.instance().experimentCookie,
+      );
+    });
+
+    it('saves cookie with custom cookie config', () => {
```
willdurand: it allows a custom cookie configuration

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,72 @@
...
+
+      sinon.assert.calledWith(
+        _cookie.save,
+        `experiment_${id}`,
+        root.instance().experimentCookie,
+        overrideCookieConfig,
+      );
+    });
+  });
+});
```
willdurand: You can also add a test case for the display name, similar to:

```js
    it('sets a display name', () => {
      expect(Overlay.displayName).toMatch(/WithUIState\(OverlayBase\)/);
    });
```

src/core/withExperiment.js
```diff
@@ -0,0 +1,83 @@
+/* @flow */
+import * as React from 'react';
+import invariant from 'invariant';
+import cookie from 'react-cookie';
+
+export type WithExperimentInjectedProps = {|
+  variant: string,
+|};
+
+type Props = {|
```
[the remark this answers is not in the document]
rebmullin: right..

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -33,4 +40,53 @@ describe(__filename, () => {
     expect(carousel).toHaveLength(1);
     expect(carousel).toHaveProp('random', true);
   });
+
+  it('calls tracking on home page view', () => {
+    const fakeTracking = createFakeTracking();
+    const root = shallowRender({ _tracking: fakeTracking });
+
+    const { variant } = root.instance().props;
```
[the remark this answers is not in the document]
rebmullin: I can look into doing this another way..? but no unfortunately it was not available with `root.prop('variant')`

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,72 @@
+import { shallow } from 'enzyme';
+import * as React from 'react';
+
+import { withExperiment } from 'core/withExperiment';
+import { fakeCookie } from 'tests/unit/helpers';
+
+describe(__filename, () => {
+  describe('withExperiment', () => {
```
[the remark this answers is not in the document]
rebmullin: 🤦‍♂️ yep..

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -206,26 +238,50 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
     });
   }
 
+  trackExperimentClick = (e: SyntheticEvent<any>, url: string) => {
+    const { _tracking, variant } = this.props;
+
+    _tracking.sendEvent({
+      action: `${variant} | Click`,
+      category: AB_HOME_HERO_EXPERIMENT_CATEGORY,
+      label: url,
```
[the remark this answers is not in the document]
rebmullin: I thought about this but decided to use what we already had but sure I can update :)

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,51 @@
...
+    _tracking: tracking,
+  };
+
+  componentDidMount() {
+    const { _tracking, variant } = this.props;
+
+    if (variant) {
+      _tracking.sendEvent({
+        action: `${variant} | Page View`,
+        category: AB_HOME_HERO_EXPERIMENT_CATEGORY,
```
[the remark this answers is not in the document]
rebmullin: I guess I misunderstood but yeh I can move event type to category

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,51 @@
...
   i18n: I18nType,
 |};
 
-export class HomeHeroBannerBase extends React.Component<Props> {
+type InternalProps = {|
+  ...Props,
+  _tracking: typeof tracking,
+|};
+
+export const AB_HOME_HERO_TEST_NAME = 'HOME_HERO';
```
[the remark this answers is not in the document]
rebmullin: > Is TEST_NAME the "experiment" name?

AB_HOME_HERO_EXPERIMENT better?

> why do you want to keep the value all capitalized?

It doesn't have to be all uppercase I guess..I can make it lowercase.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,51 @@
...
 
 import './styles.scss';
 
 type Props = {|
+  ...WithExperimentInjectedProps,
   i18n: I18nType,
 |};
 
-export class HomeHeroBannerBase extends React.Component<Props> {
+type InternalProps = {|
```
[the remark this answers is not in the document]
rebmullin: oh I see..ok thanks

src/core/withExperiment.js
```diff
@@ -0,0 +1,83 @@
...
+    experimentCookie: string | void;
+
+    static defaultProps = {
+      _cookie: cookie,
+      id: defaultId,
+      randomizer: Math.random,
+      variantA: defaultVariantA,
+      variantB: defaultVariantB,
+    };
+
```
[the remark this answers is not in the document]
rebmullin: just curious - why move tracking to the constructor here?

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
rebmullin: oh!? I guess this was already there 🤦‍♂️

src/amo/components/HomeHeroBanner/index.js
willdurand: nope, bounce rate is computed by google analytics. we will know which home gets the best rate thanks to the "page view" event you send

src/ui/components/HeroSection/index.js
willdurand: looks good

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
willdurand: yep, otherwise the cookie name is a bit weird :D

src/amo/components/HomeHeroBanner/index.js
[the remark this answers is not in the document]
willdurand: I had to double check the google analytics doc. From my understanding, one category gathers several actions.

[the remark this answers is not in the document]
willdurand: You want to gather everything in one category? Instead of having two categories with each two actions (our 2 variants)?

src/core/withExperiment.js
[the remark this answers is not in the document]
willdurand: d'oh sorry, this should not be part of the diff, I was testing something...

tests/unit/amo/components/TestHomeHeroBanner.js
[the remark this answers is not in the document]
willdurand: leave it as is then 🤷‍♂️

src/core/withExperiment.js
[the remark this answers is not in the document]
rebmullin: ahh ok, I'll ignore that : )

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -206,26 +234,51 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
 
     return heroes.map((hero) => {
+      const { title, url } = hero;
       return (
-        <HeroSection key={hero.url} linkTo={hero.url}>
-          <h3>{hero.title}</h3>
+        <HeroSection
+          key={url}
+          linkTo={url}
+          experimentTitle={title}
```
rebmullin: note: I added this here so it would be available for the tests. 
Also I named it so we know to remove this when the test is done.

rebmullin:
@willdurand, thanks for all the feedback : ) Can you please see how it is now?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,47 @@
...
+import makeClassName from 'classnames';
 
 import translate from 'core/i18n/translate';
 import Hero from 'ui/components/Hero';
 import HeroSection from 'ui/components/HeroSection';
 import type { I18nType } from 'core/types/i18n';
+import tracking from 'core/tracking';
+import {
+  withExperiment,
+  type WithExperimentInjectedProps,
```
willdurand: Can we split this import into two imports? One regular `import` and one `import type`? We put all `import type` after the regular `import`.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -1,19 +1,47 @@
...
+
+export class HomeHeroBannerBase extends React.Component<InternalProps> {
+  static defaultProps = {
+    _tracking: tracking,
+  };
+
+  componentDidMount() {
+    const { _tracking, variant } = this.props;
+
+    if (variant) {
```
willdurand: can we have no variant?

src/amo/components/HomeHeroBanner/index.js
willdurand: mmh that does not look good to me because tests should not change our production code.

src/core/withExperiment.js
```diff
@@ -0,0 +1,87 @@
...
+import invariant from 'invariant';
+import cookie from 'react-cookie';
+
+import { getDisplayName } from 'core/utils';
+
+export type WithExperimentInjectedProps = {|
+  variant: string,
+|};
+
+type WithExperimentProps = {|
```
willdurand: I would put a `w` instead of `W` (in `with`) to match the HOC name.

src/core/withExperiment.js
```diff
@@ -0,0 +1,87 @@
...
+  cookieConfig?: Object,
```
willdurand: I think we can do slightly better:

```
type CookieConfig = {|
  path?: string,
|};

// ...

type WithExperimentProps = {| 
  cookieConfig?: CookieConfig,
  // ...
|};
```

we can reuse `CookieConfig`:

```
const defaultCookieConfig: CookieConfig = { path: '/' }; 
```

src/ui/components/HeroSection/index.js
```diff
@@ -10,7 +10,10 @@ import './styles.scss';
 
 type Props = {|
   children?: any,
+  // eslint-disable-next-line react/no-unused-prop-types
```
willdurand: that's why I think we should not have that

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  function setComponentWithExperiment({ experimentProps } = {}) {
+    return withExperiment({
+      id: 'ABtest',
+      variantA: 'AName',
+      variantB: 'BName',
+      ...experimentProps,
+    })(SomeComponentBase);
+  }
```
willdurand: Can you merge this function within `renderComponentWithExperiment`?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,111 @@
...
+
+  function renderComponentWithExperiment({ experimentProps, props } = {}) {
```
willdurand: I think you can use `render`, that's shorter

src/amo/components/HomeHeroBanner/index.js
rebmullin: you're right..I guess this isn't needed

tests/unit/core/test_withExperiment.js
rebmullin: initially i had it separate but I need it separate for the displayName case

src/ui/components/HeroSection/index.js
rebmullin: hmm ok I'll see if I can the title another way..

tests/unit/core/test_withExperiment.js
willdurand: oops sorry I wanted to remove this comment, it does not imply for the reason you gave

rebmullin: ok cool, we're on the same page then : )

src/core/withExperiment.js
```diff
@@ -5,37 +5,41 @@ import cookie from 'react-cookie';
...
+type withExperimentProps = {|
+  cookieConfig?: CookieConfig,
   id: string,
   variantA: string,
   variantB: string,
 |};
 
-type WithExperimentInternalProps = {|
-  ...WithExperimentProps,
+type withExperimentInternalProps = {|
```
willdurand: I would not change this one.

src/core/withExperiment.js
```diff
@@ -5,37 +5,41 @@ import cookie from 'react-cookie';
 
 import { getDisplayName } from 'core/utils';
 
-export type WithExperimentInjectedProps = {|
+export type withExperimentInjectedProps = {|
```
willdurand: not sure about changing this one but I get why you did it.

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -26,11 +35,61 @@ describe(__filename, () => {
...
+    const heroLink = shallow(heroItem).find('.HeroSection-link-wrapper');
+
+    expect(heroLink).toHaveLength(1);
+
+    heroLink.simulate('click', createFakeEvent());
+
+    sinon.assert.calledWith(fakeTracking.sendEvent, {
+      action: variant,
+      category: `${AB_HOME_HERO_EXPERIMENT_CATEGORY} / Click`,
+      label: heroItem.props.children[0].props.children,
```
rebmullin: I know this looks kin of ugly but I couldn't seem to get at this in a cleaner way :/

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -89,7 +89,7 @@ describe(__filename, () => {
     sinon.assert.calledWith(fakeTracking.sendEvent, {
       action: variant,
       category: `${AB_HOME_HERO_EXPERIMENT_CATEGORY} / Click`,
-      label: heroItem.props.experimentTitle,
+      label: heroItem.props.children[0].props.children,
```
willdurand: That's a lot of `.` here... Question: can we get access to `sections()` (the method) and from there, get the first data and use those data to "find" components and as expected values?

src/core/withExperiment.js
rebmullin: lol..should I change this back?

rebmullin: nm..I saw your comment above

tests/unit/amo/components/TestHomeHeroBanner.js
rebmullin: The sections method returns the react component but I think if move the heroes outside of this function and call that in the test here something like that could work. If that's better I'll update..

rebmullin: I don't know if it's much better in that it's almost as long but let me know what you think..

rebmullin: what do you think?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -204,28 +228,54 @@ export class HomeHeroBannerBase extends React.Component<Props> {
         url: '/addon/britannica-insights/',
       },
     ];
+  }
 
-    return heroes.map((hero) => {
+  sections() {
+    return this.getHeroes().map((hero) => {
+      const { title, url } = hero;
```
willdurand: You could even extract `description` here

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,109 @@
...
+      props: {
+        _cookie,
+      },
+    });
+
+    sinon.assert.called(_cookie.load);
+  });
+
+  it('creates a cookie upon construction if none has been loaded', () => {
+    const id = 'Hero';
```
willdurand: can you lowercase the value?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,109 @@
...
+
+    sinon.assert.calledWith(
+      _cookie.save,
+      `experiment_${id}`,
+      root.instance().experimentCookie,
+    );
+  });
+
+  it('does not create a cookie upon construction if one has been loaded', () => {
+    const id = 'Hero';
```
willdurand: can you lowercase the value?

tests/unit/core/test_withExperiment.js
```diff
@@ -0,0 +1,109 @@
...
+      },
+    });
+
+    sinon.assert.notCalled(_cookie.save);
+  });
+
+  it('it allows a custom cookie configuration', () => {
+    const overrideCookieConfig = { path: '/test' };
+    const _cookie = fakeCookie();
+    const id = 'layoutTest';
```
willdurand: can you add a different value to this `id`, like `custom_cookie_config`?

src/amo/components/HomeHeroBanner/index.js
willdurand: And add a blank line below too.

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -26,11 +35,63 @@ describe(__filename, () => {
...
+  it('sends a tracking event on hero click', () => {
+    const fakeTracking = createFakeTracking();
+
+    const root = shallowRender({
+      _tracking: fakeTracking,
+    });
+
+    const { variant } = root.instance().props;
+
+    const heroBanner = root.find(Hero);
```
willdurand: can you rename the variable to `hero`?

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -26,11 +35,63 @@ describe(__filename, () => {
...
+
+    // We'll use the first item in the home heroes array as an example.
+    const heroItem = heroBanner.prop('sections')[0];
```
willdurand: can you rename this variable to `firstHeroItem`?

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -26,11 +35,63 @@ describe(__filename, () => {
...
+    const firstHeroTitle = root.instance().getHeroes()[0].title;
+
+    const heroLink = shallow(heroItem).find('.HeroSection-link-wrapper');
```
willdurand: can you rename this variable to `firstHeroLink`?

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -26,11 +35,63 @@ describe(__filename, () => {
...
+    expect(heroLink).toHaveLength(1);
+
+    heroLink.simulate('click', createFakeEvent());
+
+    sinon.assert.calledWith(fakeTracking.sendEvent, {
+      action: variant,
+      category: `${AB_HOME_HERO_EXPERIMENT_CATEGORY} / Click`,
+      label: firstHeroTitle,
+    });
+  });
```
willdurand: Can we add a `calledOnce`?

rebmullin: I actually removed that b/c it's called once for the "page view" (on render) and once for the "click". So it would technically be calledTwice? I'll add that with a comment..?

rebmullin:
> Time to see how it behaves in real life!

:)

thanks for all the help here @willdurand ++

tests/unit/amo/components/TestHomeHeroBanner.js
willdurand: oh right, then nevermind

rebmullin: Ok cool..I'll remove the calledTwice then

Issue #6143, opened 2018-09-05
Develop mozAddonManager with example.com


kumar303:
Currently, developing `mozAddonManager` features requires [building a custom Firefox](https://github.com/mozilla/addons-frontend/blob/master/docs/moz-addon-manager.md) which is cumbersome and time consuming. Let's see if we can develop without a custom build of Firefox.

Instructions:

1. Build a dev server

```
NODE_APP_INSTANCE=amo NODE_ENV=dev yarn build
```

2. Run the dev server

```
NODE_APP_INSTANCE=amo NODE_ENV=dev USE_HTTPS_FOR_DEV=true SERVER_HOST=example.com yarn start
```

3. Make sure you are resolving `example.com` to your local server in `/etc/hosts`:

```
127.0.0.1   example.com
```

4. Create a Firefox profile and set the following prefs in `about:config` [per instructions](https://github.com/mozilla/addons-frontend/blob/master/docs/moz-addon-manager.md#turning-on-mozaddonmanager-for-development):

```
extensions.webapi.testing = true
xpinstall.signatures.dev-root = true
```

5. Open https://example.com:4000/
6. Since this is a secure connection and we haven't installed certs into Firefox, you will have to ignore the cert warning.

Typing `navigator.mozAddonManager` into the JavaScript console on the page shows that it's available.

**TODO**

* [ ] get console logs to appear
* [ ] get the new install buttons to show up

willdurand:
I took a different approach in https://github.com/mozilla/addons-frontend/pull/6210

PR #6168, opened 2018-09-07
Listen to onOperationCancelled to fix the install status

### LINKED ISSUE #562: Unable to reflect pending uninstall undo from about:addons#Extensions
STR:
- Install an add-on via the new disco pane
- Open about:addons#Extensions
- Click remove - see the state change to uninstalling (a restart UI should be shown here - that will be covered elsewhere)
- Click undo

What happens:
- Nothing

What should happen:
- We should get an event telling us the user has cancelled the pending uninstallation.

I've filed this platform bug to expose `onOperationCancelled` https://bugzilla.mozilla.org/show_bug.cgi?id=1279046

### LINKED ISSUE #6164: Listen to `onOperationCancelled`

1. Install an add-on from AMO
2. Go to about:addons
3. Remove this add-on
4. Undo the removed add-on in about:addons
5. Close about:addons

### What happened?

Closing about:addons when the add-on is restored does not send any event, causing AMO -dev to not know the status of the add-on. In other words, the install button on AMO is in a "pending" state.

The install button on AMO should not have a "pending" state.

We should listen to `onOperationCancelled` to update the state of the add-on, which will fix the button status.

Related issues:

- https://github.com/mozilla/addons-frontend/issues/562
- https://github.com/mozilla/addons-frontend/issues/3562
- https://bugzilla.mozilla.org/show_bug.cgi?id=1477155

### LINKED ISSUE #6227: Without refreshing the page, the add-ons button state is "Add to Firefox" instead of "Remove"
Similar with https://github.com/mozilla/addons-frontend/issues/5929 

STR:
1. Install an extension from the add-ons detail page - using the new buttons AMInstallButton
2. Open the add-on manager in a new tab click "remove"
3. Go to the add-ons detail page and refresh it to see the "Add to Firefox" state 
4. Go back to add-on manager tab and click "Undo"
5. Observe the button's state on the detail page

Expected result:
"Remove" is displayed.

Actual result:
"Add to Firefox" is displayed.

Notes:
This issue is reproducible on AMO stage with FF64 (Win10)
Workaround: page must be refreshed again

only after refresh

willdurand:
Fix #6164
Fix #562
Fix #6227

~~⚠️ depends on #6237~~

---

The `onOperationCancelled` event is fired when a user clicks the "undo"
button in `about:addons` (after having removed an add-on from this
panel). This fixes a bug where the install button state was still
"pending" on the AMO side.

Update: this `onOperationCancelled` event seems to be only sent for extensions.

### DIFF
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -163,11 +163,12 @@ export function addChangeListeners(
     needsRestart: boolean,
   |}) => void,
   mozAddonManager: MozAddonManagerType,
+  { _log = log }: {| _log: typeof log |} = {},
 ) {
   function handleChangeEvent(e: AddonChangeEvent) {
     const { id: guid, type, needsRestart } = e;
 
-    log.info('Event received', { type, id: guid, needsRestart });
+    _log.info('Event received', { type, id: guid, needsRestart });
 
     if (type === ON_OPERATION_CANCELLED_EVENT) {
       // We need to retrieve the correct status for this add-on.
@@ -182,7 +183,7 @@ export function addChangeListeners(
           });
         })
         .catch((error) => {
-          log.error(
+          _log.error(
             'Unexpected error after having received onOperationCancelled event',
             error,
           );
@@ -203,7 +204,7 @@ export function addChangeListeners(
 
   if (mozAddonManager && mozAddonManager.addEventListener) {
     for (const event of GLOBAL_EVENTS) {
-      log.info(`adding event listener for "${event}"`);
+      _log.info(`adding event listener for "${event}"`);
       mozAddonManager.addEventListener(event, handleChangeEvent);
     }
 
@@ -212,9 +213,9 @@ export function addChangeListeners(
       handleChangeEvent,
     );
 
-    log.info('Global change event listeners have been initialized');
+    _log.info('Global change event listeners have been initialized');
   } else {
-    log.info('mozAddonManager.addEventListener not available');
+    _log.info('mozAddonManager.addEventListener not available');
   }
 
   return handleChangeEvent;
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -8,6 +8,7 @@ import {
   GLOBAL_EVENT_STATUS_MAP,
   INACTIVE,
   INSTALL_EVENT_LIST,
+  ON_OPERATION_CANCELLED_EVENT,
   SET_ENABLE_NOT_AVAILABLE,
 } from 'core/constants';
 import { addQueryParams, isTheme } from 'core/utils';
@@ -43,7 +44,7 @@ type OptionalParams = {|
 
 type GetAddonStatusParams = {|
   addon: FirefoxAddon,
-  type: string,
+  type?: string,
 |};
 
 export function getAddonStatus({ addon, type }: GetAddonStatusParams) {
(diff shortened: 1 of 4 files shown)


src/core/constants.js
```diff
@@ -245,6 +245,7 @@ export const GLOBAL_EVENT_STATUS_MAP = {
   onUninstalled: UNINSTALLED,
   onEnabling: ENABLING,
   onDisabling: DISABLING,
+  onOperationCancelled: INSTALLED,
```
kumar303: Should this always be mapped to install?

What about this:
- Go to `about:addons`
- Go to Extensions
- Click *Disable* for one of the extensions
- Click *Remove* on that same extension
- Click *undo*

After clicking undo, the extension goes back to disabled in `about:addons`. What happens to the install button for that case?

kumar303: I either need https://github.com/mozilla/addons-frontend/issues/6143 or a new hard drive to test with `mozAddonManager` 🙁

willdurand: mmh right, that will set the "installed" state to the AMO button whereas the `about:addons` shows "disabled".

I am not sure what to do though:

1. the initial issue is a super rare edge case
2. the code to hook into those listeners is not super extendable
3. your STR are a even more super rare edge case 😛 

We cannot easily set the previous state because, if we follow your STR, the previous state is "UNINSTALLING", not "DISABLED". We could filter the events to get "stable" events only and retain the last one, but this seems to introduce a lot of complexity.

willdurand: > I either need #6143 or a new hard drive to test with `mozAddonManager` 🙁

no more excuse now 😸

willdurand: so... I came up with a new patch here, which depends on https://github.com/mozilla/addons-frontend/pull/6237

willdurand:
Rebased and ready for a new look!

src/core/addonManager.js
```diff
@@ -164,18 +165,39 @@ export function addChangeListeners(
...
+          const status = getAddonStatus({ addon });
+
+          return callback({
+            guid,
+            status,
+            needsRestart,
+          });
+        })
+        .catch((error) => {
+          log.error(
```
kumar303: This catch block is missing test coverage. I think it's important to cover this case because it won't get used a lot in real life. Yes, I know that statement sounds weird 😁

tests/unit/core/test_addonManager.js
```diff
@@ -238,41 +256,79 @@ describe(__filename, () => {
...
+      const guid = 'foo@whatever';
+      const needsRestart = false;
+
+      await handleChangeEvent({
+        id: guid,
+        needsRestart,
+        type: ON_OPERATION_CANCELLED_EVENT,
+      });
+
+      sinon.assert.calledWith(fakeEventCallback, {
```
kumar303: Are you making sure your tests fail at least once? If you don't do test first development then you need to comment out the code or something like that to check failures. 

This test never fails. I don't know if it's because of this patch or a previous one but there are some leaky stubs (this patch is just meant to illustrate the problem):

```diff
diff --git a/tests/unit/core/test_addonManager.js b/tests/unit/core/test_addonManager.js
index a0b16fc90..aabaf52cb 100644
--- a/tests/unit/core/test_addonManager.js
+++ b/tests/unit/core/test_addonManager.js
@@ -256,6 +256,11 @@ describe(__filename, () => {
       addEventListener: sinon.stub(),
     };
 
+    beforeEach(() => {
+      fakeEventCallback.resetHistory();
+      fakeMozAddonManager.addEventListener.resetHistory();
+    });
+
     describe('global events', () => {
       const handleChangeEvent = addonManager.addChangeListeners(
         fakeEventCallback,
```

willdurand: Interesting. It actually fails if the event does not have the right `type`.

tests/unit/core/test_addonManager.js
```diff
@@ -238,41 +256,116 @@ describe(__filename, () => {
       addEventListener: sinon.stub(),
     };
 
-    const handleChangeEvent = addonManager.addChangeListeners(
-      fakeEventCallback,
-      fakeMozAddonManager,
-    );
+    beforeEach(() => {
+      fakeEventCallback.resetHistory();
+      fakeMozAddonManager.addEventListener.resetHistory();
```
kumar303: I took a closer look and you don't need to redefine `fakeMozAddonManager` here at all. You can just delete it. The current definition looks right:

https://github.com/mozilla/addons-frontend/blob/eb1134bf3a4f00e7f49be92aee7532aca8e41189/tests/unit/core/test_addonManager.js#L42-L46

PR #6170, opened 2018-09-07
Preload install animation (SVG)

### LINKED ISSUE #6169: Preload install animation (svg) to avoid a flicker on the very first time

1. Go to a fresh add-on detail page
2. Install an add-on
3. Observe the animation

### What happened?

Sometimes (depending on the network), the SVG does not load the very first time or it is late because the SVG is downloaded when the animation is shown.

No rendering issue.

Preloading the SVG will fix the issue.

willdurand:
Fix #6169

---

This PR merges the `AnimatedIcon` component into the `AMInstallButton`
so that we can preload the SVG properly.


tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -525,31 +524,31 @@ describe(__filename, () => {
     expect(root.find('.AMInstallButton-button')).toHaveLength(0);
 
     expect(root.find('.AMInstallButton-loading')).toHaveLength(1);
-    expect(root.find(AnimatedIcon)).toHaveLength(1);
+    expect(root.find(Icon)).toHaveLength(1);
```
rebmullin: should we add this here:

`expect(root.find(Icon)).toHaveProp('name', 'loading');`

to show the Icon is in fact loading?

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -499,20 +498,20 @@ describe(__filename, () => {
   });
 
   it.each([DOWNLOADING, ENABLING, INSTALLING, UNINSTALLING])(
-    'renders an AnimatedIcon when add-on is %s',
+    'renders an Icon when add-on is %s',
```
rebmullin: maybe 'renders a loading Icon when..'?

willdurand:
> I also like the idea of using css animation here if we can:
> 
> #6169 (comment)
> 
> maybe in a follow up issue..?

Yup, that's an interesting idea.

tests/unit/core/components/TestAMInstallButton.js
willdurand: ah right

PR #6180, opened 2018-09-07
I6179 update structure home hero array

### LINKED ISSUE #6179: Change the structure of home banner array so it can be mapped
We want to do this so we can map through the array and return the React component from this. Also will help clean things up a bit:
https://github.com/mozilla/addons-frontend/pull/6157/files#r215910683

QA: make sure heroes are not broken and look the same on home page

rebmullin:
fixes #6179


src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +18,18 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
-        <h3>{i18n.gettext('Britannica Insights')}</h3>
-
-        <p>{i18n.gettext('Add Britannica facts to your search results')}</p>
-      </HeroSection>,
-    ];
+    return heroes.map((hero, index) => {
+      const { description, title } = hero;
+      return (
+        /* eslint-disable react/no-array-index-key */
+        <HeroSection key={`hero-${index}`} linkTo={hero.url}>
```
kumar303: Using the array index as a key might cause some confusion during development when adding/removing items from the array (and hot reloading). What about using `hero.title` and `hero.url` as the key?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +18,18 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
+          <h3>{i18n.sprintf(i18n.gettext('%(title)s'), { title })}</h3>
```
kumar303: Due to the way we extract strings for localization, this approach won't work. Instead, you'll need to pass in `i18n` to a factory function to get the `heroes` array, like:

```js
function getHeroItems(i18n: I18nType) {
  return [
    {
      title: i18n.gettext('Facebook Container'),
      description: i18n.gettext('Prevent Facebook from tracking you around the web'),
      url: '...',
    }
  ];
}
```

willdurand: @rebmullin your code works but we cannot extract localized strings that way. The extraction works by parsing the code, looking for `gettext()` calls and taking the first parameter of each of these calls (very simplified version). If we use a variable, then there is no string to extract, only a variable, which has no value during extraction time. That is why we must always use plain old strings in `gettext()` calls.

rebmullin: thanks @willdurand 

I thought I saw this used with sprintf in other places so I assumed this was okay. Is this a problem here b/c in this case, it's only a variable?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +18,17 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
-      <HeroSection key="hero-29" linkTo="/addon/britannica-insights/">
-        <h3>{i18n.gettext('Britannica Insights')}</h3>
-
-        <p>{i18n.gettext('Add Britannica facts to your search results')}</p>
-      </HeroSection>,
-    ];
+    const heroes = getHeroItems(i18n);
+
+    return heroes.map((hero) => {
+      const { description, title } = hero;
```
willdurand: I think you do not need this line. Use `hero.title` and `hero.description`, just like you used `hero.url`.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +18,17 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
-      </HeroSection>,
-      <HeroSection key="hero-29" linkTo="/addon/britannica-insights/">
-        <h3>{i18n.gettext('Britannica Insights')}</h3>
-
-        <p>{i18n.gettext('Add Britannica facts to your search results')}</p>
-      </HeroSection>,
-    ];
+    const heroes = getHeroItems(i18n);
+
+    return heroes.map((hero) => {
```
willdurand: I don't think you need to create a `heroes` variable.

src/amo/homeHeroItems.js
```diff
@@ -0,0 +1,188 @@
+import type { I18nType } from 'core/types/i18n';
+
+export function getHeroItems(i18n: I18nType) {
```
willdurand: Can you put this in the `HomeHeroBanner` file directly?

willdurand: I'd add all this code in `sections()` in `HomeHeroBanner`. Maybe you can use your `heroes` variable for that!

src/amo/components/HomeHeroBanner/index.js
rebmullin: I see.. thanks @willdurand : )

src/amo/components/HomeHeroBanner/index.js
rebmullin: yep 👍

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +17,203 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
+        url: '/addon/tabliss/',
+      },
+      {
+        title: i18n.gettext('Tree Style Tab'),
+        description: i18n.gettext(
+          'Have a ton of open tabs? Organize them in a tidy sidebar',
+        ),
+        url: '/addon/tree-style-tab/',
+      },
+
```
willdurand: useless blank line

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -17,202 +17,203 @@ export class HomeHeroBannerBase extends React.Component<Props> {
...
+        description: i18n.gettext(
+          'Add Britannica facts to your search results',
+        ),
+        url: '/addon/britannica-insights/',
+      },
     ];
+
+    return heroes.map((hero) => {
+      return (
+        <HeroSection key={`hero-${hero.url}`} linkTo={hero.url}>
```
willdurand: you can probably use `key={hero.url}` only given that slugs are unique.

src/amo/components/HomeHeroBanner/index.js
rebmullin: thanks

PR #6183, opened 2018-09-07
Allow for a review to be featured at the top of the All Reviews page

### LINKED ISSUE #2771: Add a permalink to a specific review / single review page

bobsilverberg:
Fixes #2771 

There are (at least) a couple of issues:

1. I'm having the same problem I had with recommendations and SSR where the sagas do not finish in time and the page is broken for SSR, unless I move the processing to `componentDidMount`. This does cause a noticeable delay in all the content appearing on the page, so we may want to find another way of dealing with this. I've added a comment to the code about this as well.

2. I had trouble coming up with a good way to test that a particular action was not dispatched. I have also added a comment to the code about this and welcome any suggestions.

The new version of the All Reviews page is located at `/review/[reviewId]/`, e.g., http://localhost:3000/en-US/firefox/review/563237/, and currently looks like this:

screenshot 2018-09-07 15 36 15

Here's what it looks like if it's not your own review:

screenshot 2018-09-10 08 36 13

### DIFF
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -1,6 +1,5 @@
 /* @flow */
 import makeClassName from 'classnames';
-import { oneLineTrim } from 'common-tags';
 import config from 'config';
 import invariant from 'invariant';
 import * as React from 'react';
@@ -301,17 +300,18 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
     const noAuthor = shortByLine || this.isReply();
 
     if (review) {
-      const timestamp = i18n.moment(review.created).fromNow();
-      const linkStart = oneLineTrim`<a href="/${lang}/${clientApp}/addon/${
-        review.addonSlug
-      }/reviews/${review.id}/">`;
+      const timestamp = `
+        <a href="/${lang}/${clientApp}/addon/${review.addonSlug}/reviews/${
+        review.id
+      }/">
+          ${i18n.moment(review.created).fromNow()}
+        </a>
+      `;
       const byLineString = noAuthor
         ? // translators: Example in English: "posted last week"
-          i18n.gettext('posted %(linkStart)s%(timestamp)s%(linkEnd)s')
+          i18n.gettext('posted %(timestamp)s')
         : // translators: Example in English: "by UserName123, last week"
-          i18n.gettext(
-            'by %(authorName)s, %(linkStart)s%(timestamp)s%(linkEnd)s',
-          );
+          i18n.gettext('by %(authorName)s, %(timestamp)s');
 
       byLine = (
         <span
@@ -322,8 +322,6 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           dangerouslySetInnerHTML={sanitizeHTML(
             i18n.sprintf(byLineString, {
               authorName: review.userName,
-              linkStart,
-              linkEnd: '</a>',
               timestamp,
             }),
             ['a'],
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -318,6 +318,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           className={makeClassName('', {
             'AddonReviewCard-authorByLine': !noAuthor,
           })}
+          // eslint-disable-next-line react/no-danger
           dangerouslySetInnerHTML={sanitizeHTML(
             i18n.sprintf(byLineString, {
               authorName: review.userName,
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -9,6 +9,7 @@ import { compose } from 'redux';
 import AddonReview from 'amo/components/AddonReview';
 import AddonReviewManager from 'amo/components/AddonReviewManager';
 import FlagReviewMenu from 'amo/components/FlagReviewMenu';
+import Link from 'amo/components/Link';
 import { ADDONS_EDIT } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
@@ -312,6 +313,14 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </span>
         );
       }
+      byLine = (
+        <Link
+          title={i18n.gettext('Go to this review')}
+          to={`/addon/${review.addonSlug}/reviews/${review.id}/`}
+        >
+          {byLine}
+        </Link>
+      );
     } else {
       byLine = <LoadingText />;
     }
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -1,5 +1,6 @@
 /* @flow */
 import makeClassName from 'classnames';
+import { oneLineTrim } from 'common-tags';
 import config from 'config';
 import invariant from 'invariant';
 import * as React from 'react';
@@ -9,13 +10,12 @@ import { compose } from 'redux';
 import AddonReview from 'amo/components/AddonReview';
 import AddonReviewManager from 'amo/components/AddonReviewManager';
 import FlagReviewMenu from 'amo/components/FlagReviewMenu';
-import Link from 'amo/components/Link';
 import { ADDONS_EDIT } from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
 import { getCurrentUser, hasPermission } from 'amo/reducers/users';
-import { isAddonAuthor } from 'core/utils';
+import { isAddonAuthor, sanitizeHTML } from 'core/utils';
 import {
   deleteAddonReview,
   hideEditReviewForm,
@@ -55,11 +55,13 @@ type Props = {|
 type InternalProps = {|
   ...Props,
   _config: typeof config,
+  clientApp: string,
   deletingReview: boolean,
   dispatch: DispatchFunc,
   editingReview: boolean,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  lang: string,
   replyingToReview: boolean,
   siteUser: UserType | null,
   siteUserHasReplyPerm: boolean,
@@ -279,47 +281,53 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       _config,
       addon,
       className,
+      clientApp,
       deletingReview,
       editingReview,
       errorHandler,
       flaggable,
       i18n,
-      siteUserHasReplyPerm,
+      lang,
       replyingToReview,
       review,
       shortByLine,
       showRating,
       siteUser,
+      siteUserHasReplyPerm,
       verticalButtons,
     } = this.props;
 
     let byLine;
+    const noAuthor = shortByLine || this.isReply();
 
     if (review) {
       const timestamp = i18n.moment(review.created).fromNow();
-      if (shortByLine || this.isReply()) {
-        // translators: Example in English: "posted last week"
-        byLine = i18n.sprintf(i18n.gettext('posted %(timestamp)s'), {
-          timestamp,
-        });
-      } else {
-        byLine = (
-          <span className="AddonReviewCard-authorByLine">
-            {/* translators: Example in English: "by UserName123, last week" */}
-            {i18n.sprintf(i18n.gettext('by %(authorName)s, %(timestamp)s'), {
-              authorName: review.userName,
-              timestamp,
-            })}
-          </span>
-        );
-      }
+      const linkStart = oneLineTrim`<a href="/${lang}/${clientApp}/addon/${
+        review.addonSlug
+      }/reviews/${review.id}/">`;
+      const byLineString = noAuthor
+        ? // translators: Example in English: "posted last week"
+          i18n.gettext('posted %(linkStart)s%(timestamp)s%(linkEnd)s')
+        : // translators: Example in English: "by UserName123, last week"
+          i18n.gettext(
+            'by %(authorName)s, %(linkStart)s%(timestamp)s%(linkEnd)s',
+          );
+
       byLine = (
-        <Link
-          title={i18n.gettext('Go to this review')}
-          to={`/addon/${review.addonSlug}/reviews/${review.id}/`}
-        >
-          {byLine}
-        </Link>
+        <span
+          className={makeClassName('', {
+            'AddonReviewCard-authorByLine': !noAuthor,
+          })}
+          dangerouslySetInnerHTML={sanitizeHTML(
+            i18n.sprintf(byLineString, {
+              authorName: review.userName,
+              linkStart,
+              linkEnd: '</a>',
+              timestamp,
+            }),
+            ['a'],
+          )}
+        />
       );
     } else {
       byLine = <LoadingText />;
@@ -466,11 +474,13 @@ export function mapStateToProps(state: AppState, ownProps: Props) {
     }
   }
   return {
+    clientApp: state.api.clientApp,
     deletingReview,
     editingReview,
-    siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
+    lang: state.api.lang,
     replyingToReview,
     siteUser: getCurrentUser(state.users),
+    siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
     submittingReply,
   };
 }
--- a/src/amo/components/FeaturedAddonReview/index.js
+++ b/src/amo/components/FeaturedAddonReview/index.js
@@ -1,6 +1,8 @@
 /* @flow */
 import * as React from 'react';
+import NestedStatus from 'react-nested-status';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
 import { compose } from 'redux';
 
 import { fetchReview } from 'amo/actions/reviews';
@@ -22,7 +24,6 @@ import './styles.scss';
 
 type Props = {|
   addon: AddonType | null,
-  location: ReactRouterLocationType,
   reviewId: number,
 |};
 
@@ -33,6 +34,7 @@ type InternalProps = {|
   featuredReview?: UserReviewType,
   i18n: I18nType,
   loadingReview: boolean,
+  location: ReactRouterLocationType,
 |};
 
 export class FeaturedAddonReviewBase extends React.Component<InternalProps> {
@@ -79,9 +81,11 @@ export class FeaturedAddonReviewBase extends React.Component<InternalProps> {
     const featuredReviewCard =
       errorHandler.hasError() &&
       errorHandler.capturedError.responseStatusCode === 404 ? (
-        <div className="FeaturedAddonReview-notfound">
-          {i18n.gettext('The review was not found.')}
-        </div>
+        <NestedStatus code={404}>
+          <div className="FeaturedAddonReview-notfound">
+            {i18n.gettext('The review was not found.')}
+          </div>
+        </NestedStatus>
       ) : (
         <AddonReviewCard
           addon={addon}
@@ -123,6 +127,7 @@ export const extractId = (ownProps: InternalProps) => {
 };
 
 const FeaturedAddonReview: React.ComponentType<Props> = compose(
+  withRouter,
   connect(mapStateToProps),
   translate(),
   withFixedErrorHandler({ fileName: __filename, extractId }),
--- a/src/amo/components/FeaturedAddonReview/index.js
+++ b/src/amo/components/FeaturedAddonReview/index.js
@@ -1,0 +1,131 @@
+/* @flow */
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import { fetchReview } from 'amo/actions/reviews';
+import AddonReviewCard from 'amo/components/AddonReviewCard';
+import { selectReview } from 'amo/reducers/reviews';
+import { withFixedErrorHandler } from 'core/errorHandler';
+import translate from 'core/i18n/translate';
+import log from 'core/logger';
+import Card from 'ui/components/Card';
+import type { UserReviewType } from 'amo/actions/reviews';
+import type { AppState } from 'amo/store';
+import type { ErrorHandlerType } from 'core/errorHandler';
+import type { AddonType } from 'core/types/addons';
+import type { I18nType } from 'core/types/i18n';
+import type { DispatchFunc } from 'core/types/redux';
+import type { ReactRouterLocationType } from 'core/types/router';
+
+import './styles.scss';
+
+type Props = {|
+  addon: AddonType | null,
+  location: ReactRouterLocationType,
+  reviewId: number,
+|};
+
+type InternalProps = {|
+  ...Props,
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  featuredReview?: UserReviewType,
+  i18n: I18nType,
+  loadingReview: boolean,
+|};
+
+export class FeaturedAddonReviewBase extends React.Component<InternalProps> {
+  constructor(props: InternalProps) {
+    super(props);
+    this.loadDataIfNeeded();
+  }
+
+  componentWillReceiveProps(nextProps: InternalProps) {
+    this.loadDataIfNeeded(nextProps);
+  }
+
+  loadDataIfNeeded(nextProps?: InternalProps) {
+    const {
+      dispatch,
+      errorHandler,
+      featuredReview,
+      loadingReview,
+      reviewId,
+    } = {
+      ...this.props,
+      ...nextProps,
+    };
+
+    if (errorHandler.hasError()) {
+      log.warn('Not loading data because of an error');
+      return;
+    }
+
+    if (!featuredReview && !loadingReview) {
+      dispatch(fetchReview({ reviewId, errorHandlerId: errorHandler.id }));
+    }
+  }
+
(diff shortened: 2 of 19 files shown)


bobsilverberg:
I added a commit which updates the permalink on the UserProfile page, so this is now feature complete for #2771.

src/amo/components/AddonReviewList/index.js
```diff
@@ -289,6 +317,16 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
       );
     }
 
+    let featuredReviewHeader;
+    if (featuredReview) {
+      featuredReviewHeader =
+        siteUser && featuredReview.userId === siteUser.id
+          ? i18n.gettext('My Review')
```
kumar303: I don't think it is necessary to customize the title for owners. It can always say `Review by %(userName)s`.

src/amo/components/AddonReviewList/index.js
```diff
@@ -54,22 +57,30 @@ type InternalProps = {|
...
-  history: ReactRouterHistoryType,
+  reviewsAreLoading: boolean,
+  siteUser: UserType | null,
 |};
 
 export class AddonReviewListBase extends React.Component<InternalProps> {
-  componentWillMount() {
+  // TODO: Note: I had to change this from componentWillMount to componenDidMount
+  // otherwise this patch breaks with SSR.
+  // I know this is not ideal, but I'm not sure how else to deal with it at this time.
```
kumar303: Could you explain how it breaks? I'm pretty sure we need to keep rendering reviews on the server for SEO.

src/amo/components/Routes/index.js
```diff
@@ -46,6 +46,11 @@ const Routes = ({ _config = config }: Props = {}) => (
       path="/:lang/:application/addon/:addonSlug/reviews/"
       component={AddonReviewList}
     />
+    <Route
+      exact
+      path="/:lang/:application/review/:reviewId"
+      component={AddonReviewList}
```
kumar303: Did you consider making a new `FeaturedAddonReview` component? The new component could render `<AddonReviewList>` underneath the featured review. This separation of concerns might make the logic easier to reason about.

src/amo/components/AddonReviewList/index.js
bobsilverberg: Sure, that sounds reasonable.

src/amo/components/AddonReviewList/index.js
bobsilverberg: The page never finishes loading. It's left with loading indicators for each of the areas. I encountered this before, and the explanation, as I understood it from @willdurand was that the sagas that are dispatched on the server do not complete in time, so the fetches are dispatched, but the loads never occur.

I agree that we should endeavour to keep rendering the reviews on the server, but this also needs to be addressed. Maybe moving it into its own component, as you suggest below, might help. It might be okay to not render the reviews on the server just for the case of a featuredReview permalink destination.

src/amo/components/Routes/index.js
bobsilverberg: Yes, I wasn't too happy about how much I had to change `AddonReviewList` to support this, so I'll give that a try.

src/amo/components/AddonReviewList/index.js
kumar303: The way our SSR render cycle works is that the component has to dispatch *all* data loading actions in the *first* render. After that, the server will wait for all running sagas to complete and then re-render the page. This process is fragile but it should be possible to achieve SSR, theoretically.

bobsilverberg: I think the problem comes because we now have a case where we need to fetch a review first (via id), before we can fetch an add-on or a list of reviews, because only after fetching the review do we know the add-on's slug. I suppose we could just add the slug as part of the link and always require it - but we could encounter the case where a review that was written at one point is for an add-on whose slug has changed, and that would cause a problem.

bobsilverberg: I resolved this issue by always requiring an `addonSlug` parameter.

src/amo/components/Routes/index.js
bobsilverberg: It got complicated trying to do that because the current list also contains a summary card. I did extract the Featured review into its own component and that seems to have simplified things a bit.

bobsilverberg:
@kumar303 This is ready for another look.

src/amo/components/Routes/index.js
```diff
@@ -46,6 +46,11 @@ const Routes = ({ _config = config }: Props = {}) => (
       path="/:lang/:application/addon/:addonSlug/reviews/"
       component={AddonReviewList}
     />
+    <Route
+      exact
+      path="/:lang/:application/addon/:addonSlug/reviews/:reviewId"
```
kumar303: Doesn't this need to come before the route above it since that one matches `/:lang/:application/addon/:addonSlug/reviews/`?

tests/unit/amo/api/test_reviews.js
```diff
@@ -431,4 +432,30 @@ describe(__filename, () => {
...
+      mockApi
+        .expects('callApi')
+        .withArgs({
+          auth: true,
+          endpoint: `reviews/review/${params.reviewId}/`,
+          errorHandler: params.errorHandler,
+          method: 'GET',
+          apiState: params.apiState,
+        })
+        .returns(Promise.resolve(fakeResponse));
```
kumar303: I learned recently that you can just do `.resolves(fakeResponse)`

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -341,6 +368,32 @@ describe(__filename, () => {
       sinon.assert.calledWith(dispatch, setViewContext(addon.type));
     });
 
+    it('does not dispatch a view context if there is no add-on', () => {
+      const dispatch = sinon.stub(store, 'dispatch');
+      const errorHandler = createStubErrorHandler();
+      render({ errorHandler });
+
+      // Note: Because we expect setViewContext NOT to be called, we don't have a specific
+      // addon.type argument to use for an assert.neverCalledWith, so instead we verify that
```
kumar303: Can you use `sinon.assert.notCalledWithMatch(dispatch, sinon.match({ type: '...' }))`? I haven't tried it.

tests/unit/amo/components/TestRoutes.js
```diff
@@ -15,7 +15,7 @@ describe(__filename, () => {
 
   it('renders the routes for the amo app', () => {
     const root = render();
-    expect(root.find(Route)).toHaveLength(26);
+    expect(root.find(Route)).toHaveLength(27);
```
kumar303: Hmm. I didn't know about this test. This doesn't seem useful.

kumar303: This wasn't a change request, just me thinking out loud.

src/amo/components/FeaturedAddonReview/index.js
```diff
@@ -0,0 +1,131 @@
...
+    const featuredReviewHeader = featuredReview
+      ? i18n.sprintf(i18n.gettext('Review by %(userName)s'), {
+          userName: featuredReview.userName,
+        })
+      : null;
+
+    const featuredReviewCard =
+      errorHandler.hasError() &&
+      errorHandler.capturedError.responseStatusCode === 404 ? (
+        <div className="FeaturedAddonReview-notfound">
```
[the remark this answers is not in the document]
bobsilverberg: I tried that but it doesn't seem to be working, but maybe I don't understand what it's supposed to do. Is it supposed to cause the response from the server for the page to have a 404 status? If so then I'm not seeing that. FTR, I'm also not seeing that working for the `NotAuthorized` component, so maybe it doesn't work when run locally? I've implemented the change anyway, but I don't see it doing anything.

src/amo/components/Routes/index.js
bobsilverberg: I assume not, because it's working fine for me, but I'll move it anyway.

bobsilverberg: Maybe the `exact` in the props is causing it to work?

tests/unit/amo/components/TestRoutes.js
bobsilverberg: Agreed, but I wasn't sure if I should just delete it as part of this patch, so I didn't.

bobsilverberg:
I've added all the missing test coverage.

src/amo/sagas/reviews.js
```diff
@@ -325,14 +326,17 @@ function* fetchReview({
...
     };
 
+    console.log('---- in saga, params: ', params);
+
     const response: ExternalReviewType = yield call(getReview, params);
+    console.log('---- in saga, response: ', response);
 
     yield put(setReview(response));
   } catch (error) {
+    console.log('---- in saga, error: ', error);
```
willdurand: I don't think we want to keep those `console.log` calls :)

bobsilverberg: Why not? ;-)

src/amo/components/FeaturedAddonReview/index.js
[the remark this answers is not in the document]
kumar303: It's kinda weird (not very React-y) but it should work, even in development. 

https://github.com/mozilla/addons-frontend/blob/001398bdcac1427675e1fb56472b1a4791402d98/src/core/server/base.js#L103

src/amo/components/FeaturedAddonReview/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const featuredReviewHeader = featuredReview
+      ? i18n.sprintf(i18n.gettext('Review by %(userName)s'), {
+          userName: featuredReview.userName,
+        })
+      : null;
+
+    const featuredReviewCard =
+      errorHandler.hasError() &&
+      errorHandler.capturedError.responseStatusCode === 404 ? (
+        <NestedStatus code={404}>
```
kumar303: BTW, I don't know if your previous comment about this was outdated but this is working as intended. I see a 404 response status.

src/amo/reducers/reviews.js
```diff
@@ -429,6 +461,18 @@ export default function reviewsReducer(
         },
       });
     }
+    case FETCH_REVIEWS: {
+      const {
```
kumar303: nit: this looks like it could fit on one line

It's a bummer that prettier won't collapse short lines automatically.

src/amo/components/AddonReviewList/index.js
```diff
@@ -320,25 +329,33 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
...
+            <FeaturedAddonReview addon={addon} reviewId={reviewId} />
+          )}
+          <CardList
+            className="AddonReviewList-reviews-listing"
+            footer={paginator}
+            header={reviewCountHTML}
+          >
+            <ul>
+              {allReviews.map((review, index) => {
+                if (!reviewId || (review && review.id !== reviewId)) {
```
kumar303: Oh, bummer. This is going to mess up the paginator. Rather than trying to add a bunch of complex code to make the paginator work, I suggest you leave it like this (or maybe don't exclude reviews?) and file an API bug to add a new parameter like `exclude_review_id=N`.

tests/unit/amo/components/TestAddonReviewList.js
```diff
@@ -620,7 +682,9 @@ describe(__filename, () => {
       it('configures a paginator with the right Link', () => {
         const root = renderWithPagination();
 
-        const footer = root.find('.AddonReviewList-reviews').prop('footer');
+        const footer = root
```
kumar303: This and a bunch of other tests like they could benefit from a helper such as:

```js
expect(renderFooter(root)).toHaveProp('LinkComponent', Link);
```

tests/unit/amo/components/TestFeaturedAddonReview.js
```diff
@@ -0,0 +1,182 @@
...
+      fetchReview({
+        reviewId,
+        errorHandlerId: errorHandler.id,
+      }),
+    );
+  });
+
+  it('fetches a review when the reviewId changes', () => {
+    const firstReviewId = 1;
+    const secondReviewId = 1;
```
kumar303: It would be helpful to change the ID here 😎 

Actually, I think this test is also missing some setup, right?

```js
    store.dispatch(setReview({ ...fakeReview, id: firstReviewId }));
```

tests/unit/amo/components/TestFeaturedAddonReview.js
```diff
@@ -0,0 +1,182 @@
...
+    errorHandler.handle(
+      createApiError({
+        response: { status: 404 },
+        apiURL: 'https://some/api/endpoint',
+        jsonResponse: { message: 'not found' },
+      }),
+    );
+
+    const root = render({ errorHandler });
+    expect(root.find('.FeaturedAddonReview-notfound')).toHaveLength(1);
```
kumar303: This should also check for `NestedStatus` to make sure it gets a 404 status.

tests/unit/amo/components/TestFeaturedAddonReview.js
kumar303: It's great to get into the habit of watching your test fail at least once. You can do it by writing the tests first or commenting out the code that passes the test.

src/amo/components/FeaturedAddonReview/index.js
willdurand: Why do we use `NestedStatus` here and not the `NotFound` component?

src/amo/components/AddonReviewList/index.js
bobsilverberg: I'm not sure what you mean by mess up the paginator. Do you mean that if you click to go to another page you'll no longer see your featured review at the top? If so, then yes, that's how it works now. I wasn't sure how big an issue that was, because if you're going to start paging through reviews then do you really need to still see the featured review at the top?

bobsilverberg: The problem with not excluding the review is that if you look at your own review, and go to edit or delete it, you see the UI in both the FeaturedReviewCard and the first AddonReviewCard in the listing, which looks pretty weird.

tests/unit/amo/components/TestFeaturedAddonReview.js
bobsilverberg: Oops! Yes, I usually do try to make sure I can make a test fail, but I guess I failed (pun intended) at that this time.

As for the setup, we don't strictly need that, as it's okay for the component to fetch the review for the first id and then also fetch the review for the second id. That's why I'm using `dispatch.resetHistory()`. I think the test is still valid, although we are asking the component to fetch two reviews. Do you feel it is important to change it and make sure the first review is already loaded during test setup?

bobsilverberg:
This is ready for another look.

src/amo/reducers/reviews.js
kumar303: If you put it all on one line, prettier will leave it. If you break any part of it on to a new line, prettier will break the whole thing. It's a weird feature. I wish it were consistent.

src/amo/components/FeaturedAddonReview/index.js
kumar303: The first attempt at this patch was to use a new top level component for the featured review but that approach didn't work. That approach would have made it easier to show a generic 404 page. I don't know if it's worth the effort of prop drilling to make it work in this new approach.

src/amo/reducers/reviews.js
bobsilverberg: I did put it all on one line, and prettier still changed it for me. It's doing the change automatically in my IDE, so maybe that behaves differently from just using the command line. With my IDE, it will not let me put it all on one line.

src/amo/components/AddonReviewList/index.js
bobsilverberg: I opened https://github.com/mozilla/addons-server/issues/9424 to allow us to exclude a review from the API results.

tests/unit/amo/components/TestFeaturedAddonReview.js
kumar303: > Do you feel it is important to change it and make sure the first review is already loaded during test setup?

No, that's not important if you don't think it's necessary. I haven't looked at the newer patch but I couldn't see how the old patch was successfully handling changes from one review to another. That logic *is* important. I'll look at the new patch.

kumar303:
I hadn't thought it was possible to view a rating-only extension on the listing page but it is since they are linked from the user profile page.

screenshot 2018-09-14 11 59 56


Amazingly, the screen totally works as expected. React is pretty great! It might not work if you are *not* signed in but I can't think of any scenario where that could happen in normal site usage.

kumar303:
> I hadn't thought it was possible to view a rating-only extension

I suppose we could change the title here to *Rating by kumar_dev2* (instead of *Review*) but that's not a big thing.

tests/unit/amo/components/TestFeaturedAddonReview.js
```diff
@@ -0,0 +1,184 @@
...
+        reviewId,
+        errorHandlerId: errorHandler.id,
+      }),
+    );
+  });
+
+  it('fetches a review when the reviewId changes', () => {
+    const firstReviewId = 1;
+    const secondReviewId = 2;
+    const dispatch = sinon.stub(store, 'dispatch');
```
kumar303: This test fails if you load the first review (this is the setup I was suggesting in previous comments):

```diff
diff --git a/tests/unit/amo/components/TestFeaturedAddonReview.js b/tests/unit/amo/components/TestFeaturedAddonReview.js
index 303cd6f22..0fb68a360 100644
--- a/tests/unit/amo/components/TestFeaturedAddonReview.js
+++ b/tests/unit/amo/components/TestFeaturedAddonReview.js
@@ -75,8 +75,11 @@ describe(__filename, () => {
   it('fetches a review when the reviewId changes', () => {
     const firstReviewId = 1;
     const secondReviewId = 2;
-    const dispatch = sinon.stub(store, 'dispatch');
     const errorHandler = createStubErrorHandler();
+    store.dispatch(
+      fetchReview({ errorHandlerId: errorHandler.id, reviewId: firstReviewId }),
+    );
+    const dispatch = sinon.stub(store, 'dispatch');
 
     const root = render({
       errorHandler,
```

However, I can't actually figure out a way to break the site because of this omission in `loadDataIfNeeded()`. I thought I could load one review then click another ID to load another but that works. I think it works because the router unmounts the component each time. 

Because of that, I wonder if you should just delete this test and not worry about it until it's a problem. What do you think? I'm surprised it works as is, though, because I've seen this case pose a problem in other components.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -317,6 +318,14 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </span>
         );
       }
+      byLine = (
+        <Link
```
kumar303: The issue calls for only linking [the timestamp](https://github.com/mozilla/addons-frontend/blob/442ee43fafba36f4c27fb693f23bd417c87b4011/src/amo/components/AddonReviewCard/index.js#L304) which I like better. What do you think? Using a timestamp for permalinks is common practice.

kumar303: I think it would look better if the links were underlined in `$grey-50` -- they will fit with the header style better. It will be more of an issue when moving the link to the timestamp.

screenshot 2018-09-14 12 38 21

tests/unit/amo/components/TestFeaturedAddonReview.js
bobsilverberg: This is odd because there isn't even any logic in `FeaturedAddonReview` to deal with this scenario, so no wonder the test doesn't pass! I'm not sure why I even wrote it. Having said that, we should deal with it in the component so I've updated it to do that. I also changed the test. I think there's a flaw in your logic above, because you call `fetchReview` but not `setReview` so that will leave `loadingReview` as `true` which would cause the test to fail because it won't dispatch `fetchReview` if `loadingReview` is true.

I've fixed all of this in both the component and the test.

src/amo/components/AddonReviewCard/index.js
bobsilverberg: Yeah, I was being lazy. I agree that's better, but it means a more complicated l10n string, because we now need to embed the link in the middle of the string. Either that, or make it 2 different strings, but that would go against i18n philosophy, I think.

I'll try going the first route, embedding the link in the string.

bobsilverberg:
@kumar303 This is ready for another look. Maybe it can make today's tag?

tests/unit/amo/components/TestFeaturedAddonReview.js
kumar303: Oh, yeah, duh. I thought I was dispatching `setReview()`.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -278,40 +281,55 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
-        byLine = (
-          <span className="AddonReviewCard-authorByLine">
-            {/* translators: Example in English: "by UserName123, last week" */}
-            {i18n.sprintf(i18n.gettext('by %(authorName)s, %(timestamp)s'), {
+      const linkStart = oneLineTrim`<a href="/${lang}/${clientApp}/addon/${
+        review.addonSlug
+      }/reviews/${review.id}/">`;
+      const byLineString = noAuthor
+        ? // translators: Example in English: "posted last week"
+          i18n.gettext('posted %(linkStart)s%(timestamp)s%(linkEnd)s')
```
kumar303: Is there any benefit to injecting the HTML tags with sprintf? It makes the string more complicated for localizers. How about this?

```diff
diff --git a/src/amo/components/AddonReviewCard/index.js b/src/amo/components/AddonReviewCard/index.js
index abd386830..aba4842ae 100644
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -301,17 +301,18 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
     const noAuthor = shortByLine || this.isReply();
 
     if (review) {
-      const timestamp = i18n.moment(review.created).fromNow();
-      const linkStart = oneLineTrim`<a href="/${lang}/${clientApp}/addon/${
-        review.addonSlug
-      }/reviews/${review.id}/">`;
+      const timestamp = `
+        <a href="/${lang}/${clientApp}/addon/${review.addonSlug}/reviews/${
+        review.id
+      }/">
+          ${i18n.moment(review.created).fromNow()}
+        </a>
+      `;
       const byLineString = noAuthor
         ? // translators: Example in English: "posted last week"
-          i18n.gettext('posted %(linkStart)s%(timestamp)s%(linkEnd)s')
+          i18n.gettext('posted %(timestamp)s')
         : // translators: Example in English: "by UserName123, last week"
-          i18n.gettext(
-            'by %(authorName)s, %(linkStart)s%(timestamp)s%(linkEnd)s',
-          );
+          i18n.gettext('by %(authorName)s, %(timestamp)s');
 
       byLine = (
         <span
@@ -322,8 +323,6 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           dangerouslySetInnerHTML={sanitizeHTML(
             i18n.sprintf(byLineString, {
               authorName: review.userName,
-              linkStart,
-              linkEnd: '</a>',
               timestamp,
             }),
             ['a'],

```

src/amo/components/AddonReviewCard/index.js
```diff
@@ -278,40 +281,55 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
+          );
+
+      byLine = (
+        <span
+          className={makeClassName('', {
+            'AddonReviewCard-authorByLine': !noAuthor,
+          })}
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={sanitizeHTML(
+            i18n.sprintf(byLineString, {
```
kumar303: Can we make these links underlined but `$grey-50` like the text? I think that would make the header look more cohesive. What do you think?

screenshot 2018-09-18 09 39 12

src/amo/components/AddonReviewCard/index.js
bobsilverberg: Duh. That's so much better. Thanks!

src/amo/components/AddonReviewCard/index.js
bobsilverberg: Yes, sorry, I recall you mentioned that earlier, but it got lost in the comments. Tbh I'm experiencing fatigue with this PR, but it seems like it's getting very close.

kumar303: It is super close, yes! It was a huge feature. For future reference, it would have been easier to review if you'd split the actions and reducer code into a separate patch.

As a general rule, I think 400 lines of changes is optimal for a review. Pushing it to 600 is usually OK but anything beyond that just becomes difficult.

bobsilverberg: My apologies. I think when I started it I didn't realize how big it was going to be, so I just started plowing through everything. I will try to take some time to think about chunking the work in the future before I just jump in and start coding.

bobsilverberg:
@kumar303 I've addressed the latest comments. Sorry about missing the one about the link colour earlier.

src/ui/components/UserReview/styles.scss
```diff
@@ -36,4 +36,19 @@
   color: $sub-text-color;
   display: flex;
   font-size: $font-size-s;
+
+  a,
+  a:link,
+  a:visited {
+    color: $grey-50;
```
kumar303: Ah, shoot. This needs to be `$sub-text-color` to match the text. They are actually different. Oops.

PR #6187, opened 2018-09-10
Remove componentWillMount() calls

### LINKED ISSUE #6186: Remove componentWillMount() calls (react upgrade)
See: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#migrating-from-legacy-lifecycles

For QA: pages should load without any problem (i.e. no obvious failures)

willdurand:
Fix #6186

---

This PR updates our codebase to get rid of the `componentWillMount()` calls and use the `constructor()` instead, as specified [here](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data):

> When supporting server rendering, it’s currently necessary to provide the data synchronously – componentWillMount was often used for this purpose but the constructor can be used as a replacement. The upcoming suspense APIs will make async data fetching cleanly possible for both client and server rendering.

### DIFF
--- a/src/amo/components/Addon/index.js
+++ b/src/amo/components/Addon/index.js
@@ -104,7 +104,9 @@ export class AddonBase extends React.Component {
     getClientCompatibility: _getClientCompatibility,
   };
 
-  componentWillMount() {
+  constructor(props) {
+    super(props);
+
     const {
       addon,
       addonIsLoading,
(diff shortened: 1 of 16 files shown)


willdurand:
CI fails because of bundlesize and +0.11KB added (for some reasons).

PR #6210, opened 2018-09-11
Add new amo:dev-https command

### LINKED ISSUE #6143: Develop mozAddonManager with example.com

willdurand:
Fix #6143

---

~~This setup works at 99%. I am not able to install extensions though~~ (edit: it works with Nightly), but the rest looks good. CSP complains about sha1 signatures too, but this is not related.

All you need to get started is to import [the CA certificate](https://github.com/mozilla/addons-frontend/blob/14624803fc0d8db424cb2da471f68911163c6253/bin/local-dev-server-certs/example.com.ca.crt.pem) in Firefox:

screen shot 2018-09-11 at 16 07 46

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Here are some commands you can run:
 | --------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
 | yarn amo                    | Start the dev server/proxy (for amo) using data from Docker                                                                   |
 | yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/)                         |
+| yarn amo:dev-https          | Same as `amo:dev` but with HTTPS                                                                                              |
 | yarn amo:no-proxy           | Start the dev server without a proxy (for amo) using data from Docker                                                         |
 | yarn amo:stage              | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/)                         |
 | yarn build                  | Build an app specified with the `NODE_APP_INSTANCE` environment variable.                                                     |
--- a/README.md
+++ b/README.md
@@ -33,38 +33,38 @@ The easiest way to manage multiple node versions in development is to use [nvm](
 
 Here are some commands you can run:
 
-| Command                     | Description                                                                                                                   |
-| --------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
-| yarn amo                    | Start the dev server/proxy (for amo) using data from Docker                                                                   |
-| yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/)                         |
-| yarn amo:dev-https          | Same as `amo:dev` but with HTTPS                                                                                              |
-| yarn amo:no-proxy           | Start the dev server without a proxy (for amo) using data from Docker                                                         |
-| yarn amo:stage              | Start the dev server/proxy (for amo) using data from the staging server (https://addons.allizom.org/)                         |
-| yarn build                  | Build an app specified with the `NODE_APP_INSTANCE` environment variable.                                                     |
-| yarn build-all              | Build all the applications.                                                                                                   |
-| yarn build-ci               | Run the `build-all` and `bundlesize` npm scripts.                                                                             |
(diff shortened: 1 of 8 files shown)


kumar303:
That's cool to hear you got this working! I'll try it out later today. Could you also add a section to https://github.com/mozilla/addons-frontend/blob/master/docs/moz-addon-manager.md about how to use it?

willdurand:
> That's cool to hear you got this working! I'll try it out later today. Could you also add a section to /docs/moz-addon-manager.md@master about how to use it?

I was thinking about not advertising this feature right now and submit a doc PR later.

kumar303:
> I was thinking about not advertising this feature right now

Why not? I am not a fan of hidden features. Also, it's sometimes hard to find time and motivation to document things later which is why I always try to push for docs in the original patch -- it could start with minimal docs, though.

README.md
```diff
@@ -37,6 +37,7 @@ Here are some commands you can run:
 | --------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
 | yarn amo                    | Start the dev server/proxy (for amo) using data from Docker                                                                   |
 | yarn amo:dev                | Start the dev server/proxy (for amo) using data from the dev server (https://addons-dev.allizom.org/)                         |
+| yarn amo:dev-https          | Same as `amo:dev` but with HTTPS                                                                                              |
```
kumar303: It really only works with `example.com` so I think this should mention that.

bin/proxy.js
```diff
@@ -52,31 +55,46 @@ function getHost(req) {
...
+const proxy = httpProxy.createProxyServer({ secure: false });
+
+const handler = (req, res) => {
   const host = getHost(req);
   return proxy.web(req, res, {
     target: host,
     changeOrigin: true,
     autoRewrite: true,
-    protocolRewrite: 'http',
+    protocolRewrite: protocol,
```
kumar303: Hmm, does that work when proxying non-https URLs? I guess it would only come into play if we tried to proxy a local Docker API via `olympia.test`. No one really uses that so it's not a big deal.

willdurand: I don't know, before it was `http` only so I don't know.. Maybe we could remove this option if we get an issue?

And if no one is using the `olympia.test` env (docker env, no?), then we could maybe remove it entirely?

willdurand:
> I would like to see documentation added as part of this patch but I'd be OK with doing it in a separate issue as long we work on it ASAP (in fact, I'd be happy to write the docs if you want to kick me the issue).

added some doc!

bin/proxy.js
kumar303: Yes, `olympia.test` is for Docker and we want to support it even though no one uses it. It's the only way you can develop without Internet access. However, I'm not worried about your change on this line. Let's see what happens.

willdurand: I tried to develop without Internet access last time... Ended up writing a Go proxy to cache everything before going offline -_-

kumar303:
Thanks for the docs! This is really helpful because all team members will need to do this at some point.

Issue #5944, opened 2018-08-14
Edit review in a text box (not an overlay) on the listing page


kumar303:
This is part of a new UX for reviews/ratings. See #5519

- Go to an add-on detail page
- Rate an add-on
- Write a review (see https://github.com/mozilla/addons-frontend/issues/5942)
- Go to the review listing page
- Click *Edit my review*

This should let you edit the review in a text box, not an overlay.

ioanarusiczki:
Verified on AMO dev with FF62(Win10 and Android 8.0)
Filed 2 issues here https://github.com/mozilla/addons-frontend/issues/6272 and here https://github.com/mozilla/addons-frontend/issues/6271

@kumar303 I need a confirmation for this because I think it is an expected behavior but not completely sure : if the text box is opened for edit on the detail page it will be displayed as open on the listing review page too if the user clicks the "Read all reviews" link (see the .gif below). Text typed in without clicking the "update review" button is not visible if user navigates from one page to the other clicking the "read all reviews" from the detail page and the name of the add-on from the listing review page. 

![the opened text box](https://user-images.githubusercontent.com/33448286/45688136-3e822480-bb59-11e8-8a23-a8e568042c60.gif)

Issue #2043, opened 2017-03-13
Comment date is not updated after comment edit


ValentinaPC:
Steps to reproduce:
1. Add a comment to a theme/add-on on Android i.e. https://addons-dev.allizom.org/en-US/android/addon/full-screen-252573/
2. Wait a while and edit your comment

Expected results:
Date of the comment is updated to the moment that the comment was edited.

Actual results:
Date of the comment is not the same.

Notes/Issues:
Verified on FF52(Android 6.0.1). Issue is reproducing on both dev and stage.
Comment edited today: 
![edit_001](https://cloud.githubusercontent.com/assets/15685960/23851321/c2dc3a9a-07eb-11e7-963c-0cbf3303e005.png)

kumar303:
might be related to https://github.com/mozilla/addons-frontend/issues/1664

muffinresearch:
This is still an issue.

kumar303:
Oh, I see. We explicitly show the creation time:

https://github.com/mozilla/addons-frontend/blob/2eecafc5e7de5c5b67b45364e33fb696b270d2f3/src/amo/components/AddonReviewListItem/index.js#L198

pwalm:
Hmmm... How about: "posted 3 days ago (edited 5 mins ago)"

Repeating "ago" twice seems weird, but it works for me.

tofumatt:
While I’m rarely one to point to reddit as a way to do things 😅, I think their edit system works well here. They just show an asterisk next to the time posted and add a title attribute for when it was edited iirc.

I think adding a bracketed “edited” with a title attribute for more info (“last edited 5 mins ago”) would work eg: `posted 3 days ago <span title=“Last edited 5 minutes ago”>(edited)</span>`. It prevents the “ago” appearing twice, shows a post was edited, and preserves relative dates so it’s easy to compare creation date with last edited date.

tofumatt:
Side note: it might be worth ignoring displaying “edited” if it was edited within a few minutes of creation date as that is likely a typo fix or similar, though obviously there’s an abuse potential there. It’s small though–the developer would need to reply to a review that was entirely changed within a few minutes to look foolish. Maybe if there’s a developer reply we show a post was edited no matter what.

pwalm:
@tofumatt Let's try the Reddit way

kumar303:
I'm opening this up to contributors. I can mentor.

SeanPrashad:
[...]

PR #6220, opened 2018-09-11
Edit a review inline on the listing page

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/5944

<details>
<summary>Screencast</summary>

addons-frontend-inline-edit mov

</details>

### DIFF
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -118,8 +118,16 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
           onSubmit={this.onSubmitReview}
           placeholder={placeholder}
           puffyButtons
-          submitButtonText={i18n.gettext('Submit review')}
-          submitButtonInProgressText={i18n.gettext('Submitting review')}
+          submitButtonText={
+            review.body
+              ? i18n.gettext('Update review')
+              : i18n.gettext('Submit review')
+          }
+          submitButtonInProgressText={
+            review.body
+              ? i18n.gettext('Updating review')
+              : i18n.gettext('Submitting review')
+          }
           text={review.body}
         />
       </div>
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -25,6 +25,7 @@ import type { OnSubmitParams } from 'ui/components/DismissibleTextForm';
 import './styles.scss';
 
 type Props = {|
+  onCancel?: () => void,
   review: UserReviewType,
 |};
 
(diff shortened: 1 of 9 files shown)


willdurand:
Isn't it a bit weird to allow review editing after a reply to the review has been left?

kumar303:
> Isn't it a bit weird to allow review editing after a reply to the review has been left?

Editing a review in general can be confusing, especially since developers get email notifications with the original text. 

We have this on file to address it: https://github.com/mozilla/addons-frontend/issues/2043 The decision was to just say *edited* next to the date. A fancier solution would be to do it like github where you can view all previous versions.

src/amo/components/AddonReviewManager/index.js
```diff
@@ -113,6 +114,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
         <DismissibleTextForm
           formFooter={formFooter}
           isSubmitting={flashMessage === STARTED_SAVE_REVIEW}
+          onDismiss={onCancel || undefined}
```
bobsilverberg: Won't `onCancel` be `undefined` if it's falsey?

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -806,9 +809,60 @@ describe(__filename, () => {
...
+      store.dispatch(showEditReviewForm({ reviewId: review.id }));
+      const dispatchSpy = sinon.spy(store, 'dispatch');
+
+      const root = renderInline({ review });
+
+      const manager = root.find(AddonReviewManager);
+      expect(manager).toHaveProp('onCancel');
+
+      dispatchSpy.resetHistory();
+      const onCancel = manager.prop('onCancel');
```
bobsilverberg: Nit: this `const` seems a bit redundant.

kumar303:
> When editing an existing review, the button says "Submit review" but I think "Save review" might be better

Good idea. I added it.

> It feels a bit odd to be able to edit the rating by clicking on the stars with the box open ... I can see how it might be confusing to a user

I think the *Saved* notification will help address this so I'm not too concerned. Let's get some feedback on the UI as is and see what Phil thinks.

> do you think we should consider also allowing a person to edit their rating by clicking on the small stars even with the inline form not visible?

I guess we could? I'm not too worried about it because clicking *Edit review* is a pretty intuitive way to do it if you want to change your rating from the listing page.

tests/unit/amo/components/TestAddonReviewCard.js
kumar303: I don't have a strong opinion but, to me, `manager.prop('onCancel')()` is harder to read than `onCancel()`.

PR #6221, opened 2018-09-11
Made homepage more responsive

### LINKED ISSUE #4087: AMO homepage has layout issues for German on smaller screen sizes

hritvi:
fixes #4087


src/amo/components/Home/styles.scss
```diff
@@ -105,7 +119,7 @@
 }
 
 .Home-SubjectShelf-list-item .Icon {
-  background-position: center;
+  background-position: left;
```
rebmullin: hi @hritvi, something I just noticed is if it's rtl, we should have:

`background-position: right`

Alt text

and then we just need to override this for the larger breakpoint..

hritvi:
@rebmullin In [this]( https://github.com/mozilla/addons-frontend/issues/4087#issuecomment-370063825) design it is ltr....so what should I consider?

willdurand:
> @rebmullin In [this](https://github.com/mozilla/addons-frontend/issues/4087#issuecomment-370063825) design it is ltr....so what should I consider?

it must work in both RTL and LTR mode.

hritvi:
I am sorry but I didn't get your point. Should I add a check whether it is rtl or ltr because just doing <code>background position: right</code> won't do the job as it would be rtl for all the screens with viewport size less than 500px(If that's what is required than it's fine).

willdurand:
> I am sorry but I didn't get your point. Should I add a check whether it is rtl or ltr because just doing `background position: right` won't do the job as it would be rtl for all the screens with viewport size less than 500px(If that's what is required than it's fine).

I guess I am a bit lost in this discussion 😄

We should add something like:

```scss
[dir='rtl'] & {
  background-position: right
}
```

in `.Home-SubjectShelf-list-item .Icon`

hritvi:
I have updated the PR but I don't know how to check it. Could you help me with this. Thanks :smile:

hritvi:
It is working :smile: 

screen shot 2018-09-12 at 21 25 32

screen shot 2018-09-12 at 21 25 23

src/amo/components/Home/styles.scss
```diff
@@ -1,5 +1,10 @@
 @import '~amo/css/styles';
 
+*[dir='rtl'] {
+  background-position: right;
+  unicode-bidi: embed;
+}
```
rebmullin: hi @hritvi,

I think you're on the right track here :) but we *only* want to do this in scope of the home icon here. So something like ~:

```css
.Home-SubjectShelf-list-item .Icon
     background-position: left; 
     …other styles


     [dir='rtl'] & {
        background-position: right; // this will override the above style only when it's rtl

         // you’ll probably need to add here the large breakpoint too to change it
         // back to center again 
      }
   ….
```

see before:
Alt text

after (with icons more to the right):
Alt text

does that make sense?

hritvi: screenshot from 2018-09-13 12-40-57
Does this works?

src/amo/components/Home/styles.scss
```diff
@@ -114,11 +128,26 @@
...
+    width: 100%;
+  }
+}
+
+*[dir='rtl'] {
+  .Home-SubjectShelf-list-item .Icon {
+    background-position: right;
+    unicode-bidi: embed;
+    @include respond-to(large) {
+      background-position: center;
```
rebmullin: hi @hritvi, can we insert this above (and use sass' nesting here)?

```css
.Home-SubjectShelf-list-item .Icon {
   ...
   @include respond-to(large) {
    background-position: center;
    width: 100%;
  }
  
  // insert this here 
  [dir='rtl'] & {
    background-position: right;
    // unicode-bidi: embed;
    @include respond-to(large) {
      background-position: center;
    }
  }
}

technically what you have is right but I think we should take advantage of sass' nesting since it can be done pretty easily here.

File: src/amo/components/Home/styles.scss [hunk 4]
```diff
@@ -114,11 +128,26 @@
...
+  @include respond-to(large) {
+    background-position: center;
+    width: 100%;
+  }
+}
+
+*[dir='rtl'] {
+  .Home-SubjectShelf-list-item .Icon {
+    background-position: right;
+    unicode-bidi: embed;
```
rebmullin: I think switching the position here is enough.. you can remove this `unicode-bit: embed`

rebmullin:
@hritvi, I think it's very close..just a few more small changes : )

src/amo/components/Home/styles.scss
hritvi: I suppose we can't do that because the <code>dir</code> tag is an attribute of HTML and not the span with the class Home-SubjectShelf-list-item. So we can't include it in class

I tried 
```
@include respond-to(large) {
    background-position: center;
    width: 100%;
  }
  html[dir=rtl] & {
    .Home-SubjectShelf-list-item .Icon {
      background-position: right;
      @include respond-to(large) {
        background-position: center;
      }
    }
```
But browser is not able to render the CSS written in the [dir='rtl'] block.

rebmullin: hi @hritvi, it seems that you re-nested the `.Home-SubjectShelf-list-item .Icon` again which will make it not work...

so you want to have something like this:

```diff
--- a/src/amo/components/Home/styles.scss
+++ b/src/amo/components/Home/styles.scss

.Home-SubjectShelf-list-item .Icon {
  background-position: left;
  background-repeat: no-repeat;
  background-size: 64px;
  border-radius: $border-radius-default;
  content: '';
  display: block;
  height: 64px;
  margin-bottom: 5px;
  transform: scale(0.8);
  width: 35%;

  @include respond-to(medium) {
    transform: initial;
  }

  @include respond-to(large) {
    background-position: center;
    width: 100%;
  }

+  [dir='rtl'] & {
+    background-position: right;
+
+    @include respond-to(large) {
+      background-position: center;
+    }
+  }
}

- *[dir='rtl'] {
-  .Home-SubjectShelf-list-item .Icon {
-    background-position: right;
-    unicode-bidi: embed;
-    @include respond-to(large) {
-      background-position: center;
-    }
-  }

...
```
Can you try that? I think it's just a matter of fixing the nesting issue...
Let me know how that works for you :)

rebmullin: or let me know if any questions :)

hritvi: Yes, it worked. I have updated the PR. Thanks :smile:

rebmullin:
@hritvi, thanks for updating : )

Can you resolve the conflicts here? I think some things were moved around so it's affecting your PR it seems..

src/amo/pages/Home/styles.scss
```diff
@@ -114,11 +128,24 @@
...
+  @include respond-to(large) {
+    background-position: center;
+    width: 100%;
+  }
+
+  [dir='rtl'] & {
+    background-position: right;
+
+    @include respond-to(large) {
+      background-position: center;
```
willdurand: why do you need to set this again? It is set by the previous block line 138.

hritvi: because the `rtl` overwrites the center background position and therefore I have to set it to center in the `rtl` mode also for large viewports.

rebmullin:
@hritvi, this looks 👍

caitmuenster:
Yay @hritvi! Thanks for the patch! We've added your contribution to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition). 🎉

Welcome onboard and congrats for getting your patch merged!

hritvi:
Thanks @caitmuenster :smile:

PR #6226, opened 2018-09-12
More docs on using example.com to develop mozAddonManager

kumar303:
Follow up to https://github.com/mozilla/addons-frontend/issues/6143

### DIFF
--- a/docs/moz-addon-manager.md
+++ b/docs/moz-addon-manager.md
@@ -55,8 +55,8 @@ It is possible to serve the local development version of this project with HTTPS
 
 3. In your custom Firefox profile, go to `about:config`, void the warranty, and set these prefs. Afterwards, restart Firefox for them to take effect.
 
-    - set `extensions.webapi.testing` to `true` [to turn on `mozAddonManager`](#turning-on-mozaddonmanager-in--dev-and--stage-environments)
-    - set `xpinstall.signatures.dev-root` to `true` [to install add-ons](#install-add-ons-in--dev-and--stage-environments)
+   - set `extensions.webapi.testing` to `true` [to turn on `mozAddonManager`](#turning-on-mozaddonmanager-in--dev-and--stage-environments)
+   - set `xpinstall.signatures.dev-root` to `true` [to install add-ons](#install-add-ons-in--dev-and--stage-environments)
 
 4. Start `addons-frontend` with the command below:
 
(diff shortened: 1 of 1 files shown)


docs/moz-addon-manager.md
```diff
@@ -53,24 +53,29 @@ It is possible to serve the local development version of this project with HTTPS
...
-This allows you to browse the project at https://example.com:3000/ (and not `localhost`). In order to get access to the `mozAddonManager` and be able to install add-ons, you still need to set two prefs:
+This allows you to browse the project at https://example.com:3000/ (and not `localhost`).
 
-- `extensions.webapi.testing` set to `true` [to turn on `mozAddonManager`](#turning-on-mozaddonmanager-in--dev-and--stage-environments)
-- `xpinstall.signatures.dev-root` set to `true` [to install add-ons](#install-add-ons-in--dev-and--stage-environments)
+You are all set to develop with `mozAddonManager`!
 
-You are all set!
+**Known issues**:
+- You cannot yet sign in when the site is running from `example.com`.
```
willdurand: ah! I did not try that. What if you sign in at https://addons-dev.allizom.org?

docs/moz-addon-manager.md
```diff
@@ -53,24 +53,29 @@ It is possible to serve the local development version of this project with HTTPS
    - Choose the file you have downloaded before (`example.com.ca.crt.pem`)
    - Accept this CA certificate to trust websites
 
-3. Start this project with the command below:
+3. In your custom Firefox profile, go to `about:config`, void the warranty, and set these prefs. Afterwards, restart Firefox for them to take effect.
+
+    - set `extensions.webapi.testing` to `true` [to turn on `mozAddonManager`](#turning-on-mozaddonmanager-in--dev-and--stage-environments)
+    - set `xpinstall.signatures.dev-root` to `true` [to install add-ons](#install-add-ons-in--dev-and--stage-environments)
+
+4. Start `addons-frontend` with the command below:
```
willdurand: Well, actually it only works for AMO, not Disco Pane (yet).

docs/moz-addon-manager.md
kumar303: The problem lies in our [`fxaConfig`](https://github.com/mozilla/addons-frontend/blob/f24407c8975cbe997dcdcfd9e66f8977558cfe3d/config/development-amo.js#L11) which points to this on the server for `local`:

https://github.com/mozilla/addons-server/blob/91f3183c05fcf60eb03502704118a0e2ab0801f3/settings.py#L105-L106

The `redirect_url` parameter there actually *has not effect* but if you load the dashboard to see the configuration for the client ID, its redirect indeed points back to `localhost:3000`. 

We need a new configuration with a redirect back to `https://example.com:3000` and that's pretty much it, I think.

docs/moz-addon-manager.md
kumar303: OK, I'll add it to the known issues.

PR #6228, opened 2018-09-12
Fix AMInstallButton for opensearch add-ons

### LINKED ISSUE #6225: Failing to install Search Tools after creating a new FF profile
STR:
1. Load AMO stage - and go to More click on Search Tools.
2. Try to install a search tools from the list (the install button will probably look as being inactive) but refreshing page will work.  You will notice that the search tool can be installed.
3. Create a new profile and go again in More click on Search Tools.
4. Try to install again a search tool from the list, click +Add to Firefox
5. Observe the page.

Expected results:
The search tools is installed without problems.

Actual results:
Page's layout is broken as displayed below and the search tool cannot be installed.

Notes:
This issue is reproducible on AMO dev and AMO stage with FF62, FF64 (Win10, Mac OS X 10133)
 
The inactive button from step 2
disabledinstallsearch

The new profile issue 
cannot install search tools

willdurand:
Fix #6225

### DIFF
--- a/src/core/components/AMInstallButton/index.js
+++ b/src/core/components/AMInstallButton/index.js
@@ -299,11 +299,10 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
       } else if (addon.type === ADDON_TYPE_THEME) {
         buttonProps['data-browsertheme'] = JSON.stringify(getThemeData(addon));
         buttonProps.onClick = this.installTheme;
+      } else if (addon.type === ADDON_TYPE_OPENSEARCH) {
+        buttonProps.onClick = this.installOpenSearch;
       } else if (hasAddonManager) {
-        buttonProps.onClick =
-          addon.type === ADDON_TYPE_OPENSEARCH
-            ? this.installOpenSearch
-            : this.installExtension;
+        buttonProps.onClick = this.installExtension;
       }
     }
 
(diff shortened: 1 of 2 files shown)


kumar303:
Well, huh. I guess my Nightly setup is not allowing me to install open search extensions with `enableAMInstallButton` either true or false. Any ideas on what I might be missing?

The first try seems to go OK:

screenshot 2018-09-13 09 26 42

but then the button does not change and the extension is not installed (there's nothing in `about:addons`). I see this in the console:

screenshot 2018-09-13 09 28 25

When I try to install it again, I see this message:

screenshot 2018-09-13 09 28 37

src/core/components/AMInstallButton/index.js
```diff
@@ -263,7 +263,11 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
     });
 
     const buttonIsDisabled =
-      disabled === false ? hasAddonManager && status === UNKNOWN : disabled;
+      disabled === false
+        ? hasAddonManager &&
+          status === UNKNOWN &&
+          addon.type !== ADDON_TYPE_OPENSEARCH
```
kumar303: Is this because we don't get a status from `mozAddonManager` for opensearch extensions?

willdurand: Yes.

I don't think there is a bugzilla issue for that given that opensearch plugins are installed with `window.external.AddSearchProvider` (which is why the opensearch add-ons have no status)

kumar303: Ah, that's fine. I forgot that it's a completely separate API.

PR #6235, opened 2018-09-13
Refactor doc a bit, and add more content

willdurand:
Fix mozilla/addons#10318

---

This PR moves some doc sections from the `README` to the `docs/` folder, adding new pages.

It also adds a new page about the custom HOCs we use.


docs/high-order-components.md
```diff
@@ -0,0 +1,89 @@
+# High Order Components
```
kumar303: These docs are cool. Thanks. I'm a little worried that they will fall out of sync with the code, though. They're fine for now but maybe we can consider something later like [Sphinx for JavaScript](https://github.com/erikrose/sphinx-js). I also really like how easy it is to interlink docs with Sphinx. Hmm. I can't tell how well Sphinx supports Markdown these days, though.

kumar303:
Sure, that sounds like it could work as long as the `docs/` landing page provides enough guidance to discover all pages.

Issue #2595, opened 2017-06-15
We should not change icon_url to iconUrl


tofumatt:
### Describe the problem and steps to reproduce it:
This code https://github.com/mozilla/addons-frontend/blob/master/src/core/reducers/addons.js#L12 doesn't really do much but is really annoying and weird to debug. We should remove it.

### What happened?
I wanted to refactor a test.

### What did you expect to happen?
It was annoying to refactor because we rename this one property.

### Anything else we should know?
@kumar303 is going to run out of GIFs, be sure to bring him extra at the All-Hands.

willdurand:
Code is here AFAIK: https://github.com/mozilla/addons-frontend/blob/afc5cc00788734e3dc74190559097a022825e559/src/core/reducers/addons.js#L194-L197

SIMRAN1:
@willdurand I want to work on this issue. May I proceed?

willdurand:
@SIMRAN1 yep

willdurand:
We actually agreed on not doing this (see: https://github.com/mozilla/addons-frontend/pull/6241#discussion_r217512958).

kumar303:
The add-ons reducer was aliasing `iconUrl` to `icon_url` to accommodate this bonkers code in Discopane that everyone was afraid to touch:

https://github.com/mozilla/addons-frontend/blob/d6c8b4a34204f8848bfceae6f5afbe599432fc86/src/disco/containers/DiscoPane.js#L129

This issue was simply about cleaning up the reducer by removing the unused alias:

https://github.com/mozilla/addons-frontend/blob/3949e3e2bc4d79c0a5510213b9b2f48f942dccd8/src/core/reducers/addons.js#L198-L201

As far as I can tell, Discopane is no longer camel casing keys so I think it's safe to remove?

@willdurand can we re-open this issue? I think you were talking about something completely different in https://github.com/mozilla/addons-frontend/pull/6241#discussion_r217512958

willdurand:
mmh I thought it was a more generic issue stating that aliasing `iconUrl` to `icon_url` was counter-productive, not just this prop. I don't understand why we would like to keep a 1:1 binding for this property and not for other props to be honest.

> As far as I can tell, Discopane is no longer camel casing keys so I think it's safe to remove?

I think so yes.

ioanarusiczki:
@willdurand Is this for "qa not needed" label or is there something we could test ?

willdurand:
[...]

PR #6241, opened 2018-09-13
Add addonIconUrl and addonName to internal and external review objects

### LINKED ISSUE #6130: Add more add-on info to UserReviewType
We're going to need more add-on fields in `UserReviewType` to support #5858, and we're also going to need access to the `addon` object to support #6104 after #6108 lands, so I think the easiest thing to do would be to include `addon` in `UserReviewType`. 

I'm not sure if that might introduce caching issues, though, but I suppose we could have the same issues even if we just store the `name` and `iconUrl` of the add-on in the `review` object, so this seems like a reasonable approach to me.

@kumar303 I'm going to start working on this, unless you have any objections or better ideas.

bobsilverberg:
Fixes #6130 

Note: There doesn't seem to be a way to actually test this change, not from the api/action side anyway. I suppose Flow will help us when we make use of these new properties in other components.

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -43,7 +43,7 @@ export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
 export const UNLOAD_ADDON_REVIEWS: 'UNLOAD_ADDON_REVIEWS' =
   'UNLOAD_ADDON_REVIEWS';
 
-export type SimplifiedAddonType = {|
+export type ReviewAddonType = {|
   iconUrl?: string,
   id: number,
   name?: string,
@@ -51,7 +51,7 @@ export type SimplifiedAddonType = {|
 |};
 
 export type UserReviewType = {|
-  addon: SimplifiedAddonType,
+  reviewAddon: ReviewAddonType,
   body?: string,
   created: Date,
   id: number,
@@ -70,7 +70,7 @@ export function createInternalReview(
   review: ExternalReviewType | ExternalReviewReplyType,
 ): UserReviewType {
   return {
-    addon: {
+    reviewAddon: {
       iconUrl: review.addon.icon_url,
       id: review.addon.id,
       name: review.addon.name,
@@ -518,7 +518,8 @@ export const setReviewWasFlagged = ({
 };
 
 type SetLatestReviewParams = {|
-  addon: SimplifiedAddonType,
+  addonId: number,
+  addonSlug: string,
   review: ExternalReviewType | null,
   userId: number,
   versionId: number,
@@ -530,19 +531,21 @@ export type SetLatestReviewAction = {|
 |};
 
 export const setLatestReview = ({
-  addon,
+  addonId,
+  addonSlug,
   versionId,
   review,
   userId,
 }: SetLatestReviewParams): SetLatestReviewAction => {
-  invariant(addon, 'addon is required');
+  invariant(addonId, 'addonId is required');
+  invariant(addonSlug, 'addonSlug is required');
   invariant(review !== undefined, 'review is required');
   invariant(userId, 'userId is required');
   invariant(versionId, 'versionId is required');
 
   return {
     type: SET_LATEST_REVIEW,
-    payload: { addon, review, userId, versionId },
+    payload: { addonId, addonSlug, review, userId, versionId },
   };
 };
 
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -44,7 +44,9 @@ export const UNLOAD_ADDON_REVIEWS: 'UNLOAD_ADDON_REVIEWS' =
   'UNLOAD_ADDON_REVIEWS';
 
 export type UserReviewType = {|
+  addonIconUrl: string,
   addonId: number,
+  addonName: string,
   addonSlug: string,
   body?: string,
   created: Date,
@@ -64,7 +66,9 @@ export function createInternalReview(
   review: ExternalReviewType | ExternalReviewReplyType,
 ): UserReviewType {
   return {
+    addonIconUrl: review.addon.icon_url,
     addonId: review.addon.id,
+    addonName: review.addon.name,
     addonSlug: review.addon.slug,
     body: review.body,
     created: review.created,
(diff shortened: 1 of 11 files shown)


src/amo/actions/reviews.js
```diff
@@ -43,7 +43,9 @@ export const UNLOAD_ADDON_REVIEWS: 'UNLOAD_ADDON_REVIEWS' =
   'UNLOAD_ADDON_REVIEWS';
 
 export type UserReviewType = {|
+  addonIconUrl: string,
```
willdurand: There is this issue about `icon_url` that should not be renamed to `iconUrl`: https://github.com/mozilla/addons-frontend/issues/2595. I guess it applies here too.

src/amo/actions/reviews.js
```diff
@@ -63,7 +65,9 @@ export function createInternalReview(
   review: ExternalReviewType | ExternalReviewReplyType,
 ): UserReviewType {
   return {
+    addonIconUrl: review.addon.icon_url,
     addonId: review.addon.id,
+    addonName: review.addon.name,
     addonSlug: review.addon.slug,
```
willdurand: Why not exposing directly `addon: review.addon`? We had 2 fields before so that was okay, but it has just doubled here...

bobsilverberg: True, but it still only contains a small amount of add-on information. Do you think it won't be confusing to developers who might expect a full `addon` object?

src/amo/actions/reviews.js
bobsilverberg: I'm not sure I agree with #2595. Eslint will complain if we try to use variables with underscores in them, which I think is one of the reasons it was done this way in the first place. 

I'm not sure there will be a benefit in changing it back to `icon_url` and I fear it will introduce other hassles.

@kumar303 What do you think?

src/amo/actions/reviews.js
willdurand: Let's call this property `simplifiedAddon` (or `partialAddon`): https://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#rating-detail-object.

src/amo/actions/reviews.js
willdurand: > Eslint will complain if we try to use variables with underscores in them, which I think is one of the reasons it was done this way in the first place.

Yep, I am not sure about this issue either. It's good that you don't agree, I second that (although I understand why the issue has been filed).

bobsilverberg: Well, maybe we should just close #2595 then. Too bad a contributor is already working on it. Any objections to that, @kumar303?

src/amo/actions/reviews.js
bobsilverberg: I'm not convinced this will buy us much. 

If we do this, we now have to use `simplifiedAddon.xxx` to access each property, or `const { id, name, etc } = simplifiedAddon`, each of which represents more code. It will also require changes to every component that uses add-on information from a review, whereas this patch requires no changes to any other components.

Not every consumer will need access to these two extra props, so many can remain unchanged indefinitely.

If you feel strongly about this I'll make the change, but I am not convinced of its value.

src/amo/actions/reviews.js
willdurand: alright, let's closing it.

src/amo/actions/reviews.js
willdurand: That's what we do for `reply` already. Before, there were only 2 props to flatten, so that was okay, but now it is 4 and maybe tomorrow 5 or more. I don't really see the point of naming props `addonXXX` instead of having a single data structure for them. I see flow type definitions duplicated in this patch.

I don't know, it seems to me that it won't be future-proof.

bobsilverberg: Fair enough. I'll make the change.

bobsilverberg:
@willdurand I made the changes. There are now a lot more changes to review!

src/amo/actions/reviews.js
kumar303: Calling this `addonIconUrl` seems fine to me. The issue you linked to has nothing to do with this 👻  It has to do with some scary code in Discopane that I think is long gone. I summarized it here: https://github.com/mozilla/addons-frontend/issues/2595#issuecomment-421492095

src/amo/actions/reviews.js
```diff
@@ -42,9 +42,15 @@ export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
...
+  iconUrl?: string,
+  id: number,
+  name?: string,
+  slug: string,
+|};
+
 export type UserReviewType = {|
-  addonId: number,
-  addonSlug: string,
+  addon: SimplifiedAddonType,
```
kumar303: Even though Flow will catch mistakes, I agree that we should not call this `addon`. We should call it `reviewAddon` something.

src/amo/actions/reviews.js
```diff
@@ -42,9 +42,15 @@ export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
 export const UNLOAD_ADDON_REVIEWS: 'UNLOAD_ADDON_REVIEWS' =
   'UNLOAD_ADDON_REVIEWS';
 
+export type SimplifiedAddonType = {|
```
kumar303: Could we call it `ReviewAddonType`? I think we have a similar type somewhere for the `addon` field in the response of a [search request](https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#get--api-v4-addons-search-) because it too is missing some fields (but I can't find the type now).

src/amo/actions/reviews.js
```diff
@@ -42,9 +42,15 @@ export const DELETE_ADDON_REVIEW: 'DELETE_ADDON_REVIEW' = 'DELETE_ADDON_REVIEW';
 export const UNLOAD_ADDON_REVIEWS: 'UNLOAD_ADDON_REVIEWS' =
   'UNLOAD_ADDON_REVIEWS';
 
+export type ReviewAddonType = {|
+  iconUrl?: string,
+  id: number,
+  name?: string,
```
willdurand: Why are some props like `name` and `iconUrl` optional?

bobsilverberg: Good question. I'm sure there must have been a reason I did it like that, but I cannot find one now. I'll change it.

PR #6247, opened 2018-09-14
Add pages directories

### LINKED ISSUE #5735: Proposal to identify "page-level" components
We do have some components that are "page-level" components because we do not instantiate them ourselves, we pass them to React Router, which does the instantiation and so on.

Those components should be identified as "page-level" components. Those components get props from the React Router layer (without a HOC) and we should make it clear that they are a bit special.

I see two options:

1. we introduce a new directory `pages` either in `src/<project>/` or `src/<project>/components/`
2. we suffix these components with `Page`, but I don't really like the idea

I'd go for a new directory, probably: `src/<project>/components/pages/`.

What do you think?

willdurand:
Fix #5735 

---


tests/unit/amo/components/pages/TestCollectionList.js
```diff
@@ -1,10 +1,10 @@
-import { oneLineTrim } from 'common-tags';
+import { oneLine } from 'common-tags';
```
rebmullin: did you mean to update this here?

willdurand:
> I think I'm personally more of a fan of `src/<project>/pages/` - just from a “it’s easier to spot” point of view. I feel like it could a little lost with all the components in each project. I guess as long as it’s separated for the reasons you mentioned in the issue, then this will work 👍

the thing is those "things" are still components.

willdurand:
> Quick question - should App stay outside?

I think so yeah, it has no "route".

rebmullin:
>  think so yeah, it has no "route".

I know ..it just thought since it was a wrapper I wasn't sure if it should look "special" too : )

src/amo/components/AddonsCard/index.js
```diff
@@ -12,7 +12,7 @@ import type {
   DeleteAddonNoteFunc,
   RemoveCollectionAddonFunc,
   SaveAddonNoteFunc,
-} from 'amo/components/Collection';
+} from 'amo/components/pages/Collection';
```
kumar303: I'd prefer `amo/pages/Collection`. I get that they are technically components but I just don't want to go digging for them -- there are enough component sub-directories as it is. I'd rather see it moved up a level.

src/amo/components/App/index.js
```diff
@@ -31,7 +31,7 @@ import {
   VIEW_CONTEXT_HOME,
   maximumSetTimeoutDelay,
 } from 'core/constants';
-import DefaultErrorPage from 'core/components/ErrorPage';
+import DefaultErrorPage from 'core/components/pages/ErrorPage';
```
kumar303: I don't think `ErrorPage` makes sense in `pages`. I know that there are routes for them but they are just debugging routes used for development. I see `NotAuthorized`, `NotFound`, and so on as low-level components that routed pages will import and render.

src/disco/components/Routes/index.js
```diff
@@ -3,12 +3,12 @@ import config from 'config';
 import * as React from 'react';
 import { Route, Switch } from 'react-router-dom';
 
-import GenericError from 'core/components/ErrorPage/GenericError';
-import NotFound from 'core/components/ErrorPage/NotFound';
-import DiscoPane from 'disco/components/DiscoPane';
-import SimulateAsyncError from 'core/components/error-simulation/SimulateAsyncError';
-import SimulateClientError from 'core/components/error-simulation/SimulateClientError';
-import SimulateSyncError from 'core/components/error-simulation/SimulateSyncError';
+import GenericError from 'core/components/pages/ErrorPage/GenericError';
```
kumar303: Again, I don't think these components make sense in `pages`.

src/amo/components/AddonsCard/index.js
willdurand: Let's do this!

PR #6249, opened 2018-09-14
Fixing layout issue for rating button for some languages

### LINKED ISSUE #6248: Layout issues for review banner for some languages
Ref  issue #2056  and PR #6233 

After 6233 it looks like 
screenshot_2018-09-14 lastpass add-ons fur android

Better to add some padding to blue button so it will be easy to read .

championshuttler:
Fixes #6248

### DIFF
--- a/src/ui/components/Button/styles.scss
+++ b/src/ui/components/Button/styles.scss
@@ -66,6 +66,7 @@ $puffy-font-size: 15px;
         font-size: $micro-font-size;
         height: auto;
         min-height: 24px;
+        padding: 1px 16px;
       }
 
       .Icon {


championshuttler:
Before 
screenshot_2018-09-14 lastpass add-ons fur android

After : 
screenshot_2018-09-14 lastpass add-ons fur android 1

championshuttler:
For `0 16px` it would be like 
screenshot_2018-09-14 lastpass add-ons fur android 2

src/ui/components/Button/styles.scss
```diff
@@ -66,6 +66,7 @@ $puffy-font-size: 15px;
         font-size: $micro-font-size;
         height: auto;
         min-height: 24px;
+        padding: 1px 16px;
```
willdurand: How about having a `2px` or `4px` padding?

kumar303: This will affect all buttons so it's not the right fix.

The problematic button is already special:

https://github.com/mozilla/addons-frontend/blob/9fd03080ce7b0d2d4db3ab7e748e7fa59f08d422/src/amo/components/RatingManager/styles.scss#L12

kumar303:
Ah, I just noticed that you fixed this a different way in https://github.com/mozilla/addons-frontend/pull/6233 . I'll close this PR but just let me know if I'm mistaken.

willdurand:
@kumar303 the idea was to give some more padding to these buttons after the patch in #6233, which makes buttons nicer yet hard to read.

championshuttler:
Hi @kumar303 , yeah as William mentioned this PR is about to make the button easy to read. Can we re-open it :)

Thanks

kumar303:
> @kumar303 the idea was to give some more padding to these buttons

OK

championshuttler:
Made the requested changes in correct file

After: 
screen shot 2018-10-01 at 20 19 07

willdurand:
aaaah oops, then my bad. Indeed, the PR could be re-submitted.

PR #6250, opened 2018-09-14
Add a check for non-restartless add-ons

### LINKED ISSUE #6240: Installing a non-restartless extension says it is corrupt

- Go to https://addons.mozilla.org/en-US/firefox/language-tools/
- Follow a link to one of the first dictionaries: https://addons.mozilla.org/en-US/firefox/addon/akan-ns%C9%9Bmfuaasekyer%C9%9B/
- Install it in Firefox 62

### What happened?

It cannot be installed because it's non-restartless, which is not supported. In the browser console:

```
1536858701111	addons.xpi	WARN	Download of https://addons.mozilla.org/firefox/downloads/file/376172/akan_spelling_dictionary-0.9.1-typefix-fx.xpi?src=dp-btn-primary failed: Error: Non-restartless extensions no longer supported (resource://gre/modules/addons/XPIInstall.jsm:646:13) JS Stack trace: loadManifestFromRDF@XPIInstall.jsm:646:13
loadFromRDF@XPIInstall.jsm:786:23
```

In the UI:

screenshot 2018-09-13 12 11 50

We should show an incompatible banner just like for other incompatible extensions.

It's weird that this extension is listed at all but that's a separate bug: https://github.com/mozilla/addons-server/issues/9418

willdurand:
Fix #6240 

---

~~I think we still need to know when non-restartless add-on support has been dropped because the current patch makes all non-restartless add-ons incompatible.~~ FF 61 is the answer.

### DIFF
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -173,8 +173,19 @@ export function getClientCompatibility({
     downloadUrl = FACEBOOK_CONTAINER_DOWNLOAD_URL;
   }
 
+  let compatible = agent.compatible && supportsClientApp;
+
+  const { browser } = userAgentInfo;
+  const isFF61 =
+    browser.name === 'Firefox' && mozCompare(browser.version, '61.0') >= 0;
+
+  if (compatible && isFF61 && addon && addon.isRestartRequired === true) {
+    compatible = false;
+    reason = INCOMPATIBLE_UNSUPPORTED_PLATFORM;
+  }
+
   return {
-    compatible: agent.compatible && supportsClientApp,
+    compatible,
     downloadUrl,
     maxVersion,
     minVersion,
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -145,6 +145,7 @@ export function getClientCompatibility({
   clientApp,
   userAgentInfo,
   _window = typeof window !== 'undefined' ? window : {},
+  _log = log,
 } = {}) {
   // Check compatibility with client app.
   const { supportsClientApp, maxVersion, minVersion } = getCompatibleVersions({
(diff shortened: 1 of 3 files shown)


src/core/utils/compatibility.js
```diff
@@ -173,8 +173,19 @@ export function getClientCompatibility({
     downloadUrl = FACEBOOK_CONTAINER_DOWNLOAD_URL;
   }
 
+  let compatible = agent.compatible && supportsClientApp;
+
+  const { browser } = userAgentInfo;
```
kumar303: I think it would be nice to wrap all of this code in `if (addon.isRestartRequired === true) { ...` since otherwise it's irrelevant.

tests/unit/core/utils/test_compatibility.js
```diff
@@ -702,6 +702,67 @@ describe(__filename, () => {
...
+        },
+      });
+
+      expect(
+        getClientCompatibility({
+          addon,
+          clientApp,
+          userAgentInfo,
+        }),
+      ).toEqual({
```
kumar303: This would be more concise and maintainable as `.toMatchObject({ compatible: false, reason: '...' });`

tests/unit/core/utils/test_compatibility.js
```diff
@@ -702,6 +702,67 @@ describe(__filename, () => {
...
```
kumar303: Same here

tests/unit/helpers.js
```diff
@@ -167,6 +167,8 @@ export const userAgentsByPlatform = {
       rv:33.0) Gecko/20100101 Firefox/33.0`,
     firefox57: oneLine`Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:57.0)
       Gecko/20100101 Firefox/57.1`,
+    firefox61: oneLine`Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:57.0)
```
kumar303: It would be nice to change this to `rv:61.0` just to avoid confusion. I see the line above it needs changing too 🤒

src/core/utils/compatibility.js
```diff
@@ -173,8 +173,19 @@ export function getClientCompatibility({
...
+  const isFF61 =
+    browser.name === 'Firefox' && mozCompare(browser.version, '61.0') >= 0;
+
+  if (compatible && isFF61 && addon && addon.isRestartRequired === true) {
+    compatible = false;
+    reason = INCOMPATIBLE_UNSUPPORTED_PLATFORM;
```
kumar303: re-using this reason seems fine but what about a `log.debug()` message saying that `isRestartRequired` triggered the incompatibility?

tests/unit/helpers.js
willdurand: heh, oops

PR #6256, opened 2018-09-14
Storybook format

### LINKED ISSUE #6159: Format storybook existing components (button, badge)
Currently we just have these 2 components set up - Button and Badge.

We should make it look something like described here:
https://github.com/mozilla/addons-frontend/pull/6060#discussion_r215322058

Check out addons to mentioned here to see if this makes it easier, etc

rebmullin:
fixes #6159 

I personally like chapters over the combinations in that it was a bit more flexible, so I am using that here.

I followed the design that came up [here](https://github.com/mozilla/addons-frontend/pull/6060#discussion_r215322058) as I really like that : )
Of course can tweak it as needed...

Also not sure how you think Button looks - in that it's all one 1 page..? or do you think each type deserves its own page? We can move that too but I thought in this case anyway it made more sense to leave on the 1 page.

note: I haven't added proptypes or source here but I think we can. The styling didn't look super nice so I left it off for now.


stories/ui/Button.js
```diff
@@ -1,185 +1,82 @@
...
+  ];
+}
+
+function createChapters() {
+  return buttonTypes.map((type) => {
+    return {
+      title: type,
+      sections: createSections(type).map((section) => {
+        return {
+          subtitle: `${section.description}`,
```
kumar303: I like the style of showing a string of props and values but manually defining this as `description` is going to be hard to maintain. I say remove `description` from `section` and just render the same thing by inspecting `section.props`. If you make it a common function then `Badge` can use it too.

stories/ui/Badge.js
```diff
@@ -6,16 +6,34 @@ import Badge from 'ui/components/Badge';
...
+  {
+    type: 'not-compatible',
+  },
+  {
+    type: 'requires-payment',
+  },
+];
+
+storiesOf('Badge', module).addWithChapters('Badge variations', {
+  useTheme: true,
```
kumar303: Rather than set this to `false` for every component, it looks like you can just set it as a default (according to the docs):

```diff
diff --git a/stories/setup/config.js b/stories/setup/config.js
index eadb40aea..a4a8ad884 100644
--- a/stories/setup/config.js
+++ b/stories/setup/config.js
@@ -9,6 +9,7 @@ setDefaults({
     allowSourceToggling: false,
     showPropTables: false,
     allowPropTablesToggling: false,
+    useTheme: false,
   },
 });
 
```

However, I don't see a visual difference by changing this parameter in either case.

stories/ui/Button.js
```diff
@@ -1,185 +1,82 @@
-/* @flow */
```
kumar303: Can you put this back? I don't see any errors. If you run into an error, you can use [`$FLOW_FIXME` comments](https://github.com/mozilla/addons-frontend#flow).

kumar303: I know that `Button` doesn't have Flow coverage today but hopefully it will soon!

stories/ui/Badge.js
```diff
@@ -6,16 +6,34 @@ import Badge from 'ui/components/Badge';
...
+  chapters: [
+    {
+      sections: sections.map((section) => {
+        return {
+          subtitle: `type=${section.type}`,
+          sectionFn: () => <Badge {...section} label={label} />,
```
kumar303: It's really important that we get Flow coverage for props because otherwise the stories will drift out of sync quickly. The best idea I have is to add a type hint. Any other ideas? This isn't so bad, though:

```diff
diff --git a/src/ui/components/Badge/index.js b/src/ui/components/Badge/index.js
index f13266b4e..be5608e79 100644
--- a/src/ui/components/Badge/index.js
+++ b/src/ui/components/Badge/index.js
@@ -5,7 +5,7 @@ import Icon from 'ui/components/Icon';
 
 import './styles.scss';
 
-type Props = {|
+export type Props = {|
   label: string,
   type?:
     | 'experimental'
diff --git a/stories/ui/Badge.js b/stories/ui/Badge.js
index 9146ddd40..13f50d727 100644
--- a/stories/ui/Badge.js
+++ b/stories/ui/Badge.js
@@ -3,6 +3,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 
 import Badge from 'ui/components/Badge';
+import type { Props as BadgeProps } from 'ui/components/Badge';
 
 const label = 'Hello Badge';
 
@@ -29,9 +30,10 @@ storiesOf('Badge', module).addWithChapters('Badge variations', {
   chapters: [
     {
       sections: sections.map((section) => {
+        const props: BadgeProps = { label, ...section };
         return {
           subtitle: `type=${section.type}`,
-          sectionFn: () => <Badge {...section} label={label} />,
+          sectionFn: () => <Badge {...props} />,
         };
       }),
     },

```

stories/setup/storybook.override.styles.scss
```diff
@@ -0,0 +1,37 @@
...
+  // Giving extra cushion so nothing gets covered.
+  padding: 24px 12px 12px;
+}
+
+// This overrides some chapter styles.
+.section-container {
+  border-bottom: 1px solid $background-grey;
+  display: block;
+  // We need to use !important here to override chapter's inline style.
+  margin-bottom: 0 !important;
```
kumar303: The need for `!important` seems kind of fragile. I noticed that the chapters add-on has `useTheme`, so, would it be easier to make a custom theme? I can't find info on theming in the docs so maybe you already looked into this?

rebmullin: yes good catch! actually I thought I had this set to off but 1 file didn't have this set correctly and so I missed this. Anyway, when I move it to config and set useTheme correctly, it does seem to remove the inline styles so I can remove these !importants :)

stories/ui/Badge.js
rebmullin: It seems to be a minimal difference as far as I can see. But there are some inline styles that are added so yes, I will make sure to set this to false. As noted above it should eliminate the !importants added to some chapter styles that I set up.

stories/ui/Button.js
rebmullin: > Can you put this back? 

oops! yes, that was inadvertently removed..will this add back.

stories/ui/Button.js
rebmullin: yep I agree and like your suggestion. I will update..

stories/ui/Button.js
rebmullin: maybe I should set up an issue for Button to get flow coverage?

stories/ui/Button.js
rebmullin: this part is WIP..

stories/setup/config.js
```diff
@@ -1,11 +1,27 @@
-import { configure } from '@storybook/react';
+import { configure, setAddon } from '@storybook/react';
 import { setOptions } from '@storybook/addon-options';
+import chaptersAddon, { setDefaults } from 'react-storybook-addon-chapters';
+
+// TBD: overrides chapters defaults. Do we want this?
```
willdurand: I think the more default values, the better :)

stories/index.js
```diff
@@ -2,9 +2,7 @@
 import 'amo/components/App/styles.scss';
 import 'core/css/inc/lib.scss';
 
-// These styles override some App styles that make it hard to see
-// in storybook.
-import './setup/App.override.styles.scss';
+import './setup/storybook.override.styles.scss';
```
willdurand: This filename could be `styles.scss`, no need to make it extra long.

stories/setup/config.js
willdurand: We can at least enable `allowPropTablesToggling`, WDYT?

rebmullin: I like the idea of showing props (and source ) but there are few issues with using it here:

1. it only works with Proptypes. I looked into this a bit (when I was looking into to storybook) and there is supposedly a way to get it to work with flow but I've yet to have success with this :/ I'll circle back tho...

2. with the current set up, it looks like this: 
Alt text

it displays per chapter/section it seems. I think we only need to show it 1 per component, right? So I thought we can use chapters to display not there here but in another way, altho this will be a bit more work... or do you think this is okay?


Regarding the other reasons of overriding the default options:

`allowSourceToggling: false` - same reasons as propsToggling
`showSource: false` - same reasons...
`useTheme: false` - this prevents using !important on styles

stories/ui/Button.js
kumar303: > maybe I should set up an issue for Button to get flow coverage?

Sure!

stories/setup/config.js
rebmullin: I added this in a different way (in it's own section so it would not be redundant)..let me know what you think 

note: flow is still an issue which I need to look into. 

Also - thoughts about showing source? (via toggle for each variation of just generic source at the top like I currently have set up for props)

stories/setup/config.js
```diff
@@ -1,11 +1,42 @@
...
+  setDefaults as setAddonInfoDefaults,
+} from '@storybook/addon-info';
+
+// TBD: overrides chapters defaults. Do we want this?
+setDefaults({
+  sectionOptions: {
+    allowPropTablesToggling: false,
+    allowSourceToggling: true,
+    showSource: false,
+    useTheme: false,
```
rebmullin: I sort of went into why I left these this way here: https://github.com/mozilla/addons-frontend/pull/6256#discussion_r218960154

actually I turned source toggle on but I personally think we don't need this here. 

Regarding the display of props, it doesn't seem that this works with flow type as far as I can tell so I set up addon-info for this (on it's own page).

rebmullin:
@willdurand @kumar303, I've made some changes. I am not sure if these 2 components represent well all the components so I may have to refactor a bit more but let me know if you like the general look or have any feedback about the source and props info I posted above : )

rebmullin:
I've decided to move move the ui items under the prop info so we don't have to worry too much about styles (I can add this back to the right side but if there is more text it just means more style adjustments have to be made). So to keep it easy I moved it. 

Also, for now I am removing the source toggle option. IMO this isn't needed..

let me know what you guys think : )

stories/setup/styles.scss
```diff
@@ -0,0 +1,53 @@
...
+  .section-subtitle {
+    color: #bbb;
+
+    @include respond-to(large) {
+      margin: 0;
+    }
+  }
+
+  .section-button-container {
+    float: left;
```
kumar303: I can't find `.section-button-container` in the rendered page but could this use flex instead? Floats are hard to get right.

stories/setup/webpack.config.babel.js
```diff
@@ -53,6 +53,6 @@ module.exports = {
     alias: {
       normalize: 'normalize.css/normalize.css',
     },
-    modules: [path.resolve('./src'), 'node_modules'],
+    modules: [path.resolve('./src'), path.resolve('./stories'), 'node_modules'],
```
kumar303: It's not a good idea to resolve `./stories` like this because [`src/ui`](https://github.com/mozilla/addons-frontend/tree/master/src/ui) will overshadow `stories/ui`.

Maybe we could solve it by adjusting the directory layout?

```
stories/setup/
stories/index.js
stories/stories/ <-- everything would now go here.
```

However, I don't see why you need to resolve the path to `./stories` in the first place. Is that just something you want to use in the future?

If you can, maybe try getting things working with relative imports first so we can see how annoying it is.

stories/ui/Badge.js
```diff
@@ -1,21 +1,55 @@
 /* @flow */
 import React from 'react';
 import { storiesOf } from '@storybook/react';
+import { withInfo } from '@storybook/addon-info';
 
+// $FLOW_FIXME flow not liking the path here ( will look into this more)
```
kumar303: You will need to add a [resolver](https://github.com/mozilla/addons-frontend/blob/master/.flowconfig#L22-L23) for Flow too but I think that's a bit risky (I explained why in the comment up above).

stories/ui/Badge.js
```diff
@@ -1,21 +1,55 @@
...
+  )
+  .addWithChapters('Badge variations', {
+    chapters: [
+      ...createChapters({
+        Component: Badge,
+        chapters: types,
+        children: label,
+        createPropsMatrix,
+        otherChapterProps: {
+          title: undefined,
```
kumar303: This is a bit confusing. Could you add a comment explaining why you need it?

stories/utils.js
```diff
@@ -0,0 +1,48 @@
...
+import React from 'react';
+
+type PropsMatrixType = {|
+  props: Object,
+|};
+
+type ChapterType = string | number | void;
+
+type CreateChapterParams = {|
+  // TODO: look into why React.Node or ~ React.ComponentType<any>
```
kumar303: The React import just needed fixing.

```diff
diff --git a/stories/utils.js b/stories/utils.js
index f8657a8f8..97939e957 100644
--- a/stories/utils.js
+++ b/stories/utils.js
@@ -1,5 +1,5 @@
 /* @flow */
-import React from 'react';
+import * as React from 'react';
 
 type PropsMatrixType = {|
   props: Object,
@@ -10,7 +10,7 @@ type ChapterType = string | number | void;
 type CreateChapterParams = {|
   // TODO: look into why React.Node or ~ React.ComponentType<any>
   // doesn't work here :/.
-  Component: Function,
+  Component: React.ComponentType<any>,
   chapters: Array<string | void>,
   children?: any,
   createPropsMatrix: (any) => Array<PropsMatrixType>,
```

stories/setup/webpack.config.babel.js
rebmullin: actually it was working as expected but linting and flow were complaining about this when I ran those checks : / 

I'll remove what I've added above as you suggest tho w/ relative paths for now

stories/utils.js
rebmullin: ahh! yes..thank you : )

stories/setup/styles.scss
rebmullin: actually I think this is when I was showing the toggle source (which I've since removed). I had to override their style (despite the useTheme false setting we had). Anyway, it's moot at this point so I'll remove this : )

stories/ui/Badge.js
rebmullin: yep I'll do relative paths for now.

PR #6263, opened 2018-09-17
Remove iconUrl in addon

### LINKED ISSUE #2595: We should not change icon_url to iconUrl

willdurand:
Fix #2595

### DIFF
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -195,11 +195,6 @@ export function createInternalAddon(apiAddon: ExternalAddonType): AddonType {
 
     // These are custom properties not in the API response.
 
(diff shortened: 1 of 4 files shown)


rebmullin:
Yeh I thought I saw themes still had this too but now I don't 🤷‍♂️

kumar303:
don-draper-bliss

PR #6266, opened 2018-09-17
Fix :focus button style

### LINKED ISSUE #6251: Install new buttons look like being clicked - highlighted with blue
STR:
1. Load AMO dev and go to an add-ons detail page
2. Click the install button
3. After the add-on is installed click on "Remove" button
4. Observe the button's "Add to Firefox" borders 

Expected result:
"Add to Firefox" button is not highlighted with a blue border.

Actual result:
"Add to firefox" is highlighted. 

Notes:
This is reproducible on AMO dev with FF64(Win10). 
Reproducible when the extension is enabled and then the "Remove" buttons is displayed with blue borders.
Also, when a lightweight theme is installed, "Remove" is displayed with the blue border, then the "Install to Firefox" too if installation is repeated.
The issue is also reproducible with Disco Pane.

An extension
highlighted buttons1

Lightweight theme 
highlighted buttons2

Disco pane
highlighted buttons3

willdurand:
Fix #6251

---

I found the polyfill in this article (which contains information about why it is needed): https://fvsch.com/styling-buttons/#sticky-focus

We never noticed that before because most of our buttons disappear on click or we redirect users somewhere else.

### DIFF
--- a/package.json
+++ b/package.json
@@ -214,6 +214,7 @@
     "express": "4.16.2",
     "extract-text-webpack-plugin": "3.0.2",
     "fastclick": "1.0.6",
+    "focus-visible": "4.1.5",
     "full-icu": "1.2.1",
     "helmet": "3.13.0",
     "history": "4.7.2",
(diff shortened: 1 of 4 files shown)


kumar303:
This is cool. Do you think there are any accessibility issues? I guess since it preserves focus for key interaction then it's Okay. 

@muffinresearch knows a lot about accessibility ^

PR #6268, opened 2018-09-17
Show install errors on AMO

### LINKED ISSUE #6261: There is no error message displayed when an add-on fails to install with the new install buttons
Follow up for #6240 

Prerequisites:
- open an unbranded version of Firefox
- set the following prefs:
`extensions.webapi.testing` =  `true`
`xpinstall.signatures.required` = `false`
`xpinstall.signatures.dev-root` = `true`
- optional pref:
`extensions.webextPermissionPrompts` = `false`

STR:

1. Load AMO dev
2. Search for an add-on that should fail to install - see https://addons-dev.allizom.org/en-US/firefox/addon/akan-spelling-dict-issue-6240/
3. Install the add-on

Actual result:
The add-on fails to install but there is no error message displayed

Expected result:
An error message should be displayed for every failed install attempt 

Notes:
- the reason why this drew my attention is that before the new install buttons, even on the configuration mentioned in the prerequisites, we used to receive an error when an add-on failed to install - i.e. a "corrupt" or "incompatible" message
- this scenario needs to be tested with an unbranded Firefox version, as the official version will always throw a "corrupt" message on -dev and stage because of the signatures 
- reproduced on AMO -dev and stage with FF62, Win10x64

error on install

willdurand:
Fix #6261

---

This is something I was not aware of until today... It is already
available in Disco Pane, so it is mostly mimicking what is there.

screen shot 2018-09-17 at 18 57 08

### DIFF
--- a/src/core/installAddon.js
+++ b/src/core/installAddon.js
@@ -7,7 +7,7 @@ import { oneLine } from 'common-tags';
 import config from 'config';
 
 import { getAddonIconUrl } from 'core/imageUtils';
-import { setInstallState } from 'core/actions/installations';
+import { setInstallError, setInstallState } from 'core/actions/installations';
 import log from 'core/logger';
 import themeInstall, { getThemeData } from 'core/themeInstall';
 import tracking, {
@@ -29,7 +29,6 @@ import {
   INSTALL_CANCELLED,
   INSTALL_CANCELLED_ACTION,
   INSTALL_DOWNLOAD_FAILED_ACTION,
-  INSTALL_ERROR,
   INSTALL_FAILED,
   INSTALL_STARTED_ACTION,
   INSTALL_STARTED_THEME_CATEGORY,
@@ -155,10 +154,7 @@ export function makeProgressHandler({
     } else if (event.type === 'onDownloadEnded') {
       dispatch(setInstallState({ guid, status: INSTALLING }));
     } else if (event.type === 'onDownloadFailed') {
-      dispatch({
-        type: INSTALL_ERROR,
-        payload: { guid, error: DOWNLOAD_FAILED },
-      });
+      dispatch(setInstallError({ guid, error: DOWNLOAD_FAILED }));
 
       _tracking.sendEvent({
         action: getAddonTypeForTracking(type),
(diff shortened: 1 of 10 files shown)


src/core/installAddon.js
```diff
@@ -155,10 +154,7 @@ export function makeProgressHandler({
     } else if (event.type === 'onDownloadEnded') {
       dispatch(setInstallState({ guid, status: INSTALLING }));
     } else if (event.type === 'onDownloadFailed') {
-      dispatch({
-        type: INSTALL_ERROR,
-        payload: { guid, error: DOWNLOAD_FAILED },
-      });
+      dispatch(setInstallError({ guid, error: DOWNLOAD_FAILED }));
```
kumar303: 🏖 🍍 

Nice. I forgot about this.

tests/unit/disco/components/TestAddon.js
```diff
@@ -146,9 +151,7 @@ describe(__filename, () => {
         error: FATAL_INSTALL_ERROR,
       });
       const error = root.find('.notification.error');
-      expect(error.find('p').html()).toContain(
-        'An unexpected error occurred during installation.',
-      );
+      expect(error.find('p').html()).toContain(getMessage(FATAL_INSTALL_ERROR));
```
kumar303: These tests are now a duplication of what you added to `utils/test_addons.js`. I would like to see all of them deleted except for one or two which you can use as integration tests to make sure `disco/Addon.js` is calling `getErrorMessage()` and rendering the result (similar to what you added in `pages/TestAddon.js`).

tests/unit/core/utils/test_addons.js
```diff
@@ -0,0 +1,26 @@
...
+describe(__filename, () => {
+  describe('getErrorMessage', () => {
+    it.each([
+      [INSTALL_FAILED, 'Installation failed'],
+      [DOWNLOAD_FAILED, 'Download failed'],
+      [FATAL_INSTALL_ERROR, 'error occurred during installation'],
+      [FATAL_UNINSTALL_ERROR, 'error occurred during uninstallation'],
+      [FATAL_ERROR, 'unexpected error occurred.'],
+      ['SOME OTHER ERROR', 'unexpected error occurred.'],
+    ])('returns a translated message for %s', (error, message) => {
```
kumar303: nit: to avoid confusion with localization, you could say "returns a readable message for..."

willdurand: ah right!

PR #6269, opened 2018-09-17
Storybook gh-pages / adding travis script for this

### LINKED ISSUE #6254: Publish storybook to gh-pages - part 2
this includes setting up storybook config:

https://github.com/mozilla/addons-frontend/pull/6202#issuecomment-420749075

and integrating with travis

rebmullin:
fixes #6254 

I am using the following as references:

linter repo: https://github.com/mozilla/addons-linter/blob/master/.travis.yml#L38

storybook deployer ci: https://github.com/storybooks/storybook-deployer#deploying-storybook-as-part-of-a-ci-service

I am not sure if there is more to do

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,5 +81,6 @@ jobs:
     - stage:
       <<: *tox
       env: TOXENV=dennis-lint
-  after_success:
-  - test $TRAVIS_BRANCH = "master" && yarn storybook-deploy --ci
+    - stage: Storybook deployment
+      script: echo "Deploying to GitHub Pages..."
+      after_success: yarn storybook-deploy --ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,4 +81,5 @@ jobs:
     - stage:
       <<: *tox
       env: TOXENV=dennis-lint
-  after_success: yarn storybook-deploy -- --ci
+  after_success:
+  - test $TRAVIS_BRANCH = "master" && yarn storybook-deploy --ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,5 +82,6 @@ jobs:
       <<: *tox
       env: TOXENV=dennis-lint
     - stage: Storybook deployment
+      language: node_js
       script: echo "Deploying to GitHub Pages..."
       after_success: yarn storybook-deploy --ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,5 +83,6 @@ jobs:
       env: TOXENV=dennis-lint
     - stage: Storybook deployment
       language: node_js
+      node_js: '8'
       script: echo "Deploying to GitHub Pages..."
       after_success: yarn storybook-deploy --ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,4 +85,4 @@ jobs:
       language: node_js
       node_js: '8'
       script: echo "Deploying to GitHub Pages..."
-      after_success: yarn storybook-deploy --ci
+      after_success: yarn storybook-deploy --ci --remote=upstream
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,4 +85,4 @@ jobs:
       language: node_js
       node_js: '8'
       script: echo "Deploying to GitHub Pages..."
-      after_success: yarn storybook-deploy --ci --remote=upstream
+      after_success: yarn storybook-deploy --ci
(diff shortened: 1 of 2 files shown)


rebmullin:
> .. script name looks a bit weird

ha yes : ) thanks..

> You should probably not hide the output right away..

so you mean remove this piece?: /dev/null 2>&1

willdurand:
> so you mean remove this piece?: /dev/null 2>&1

yep

package.json
```diff
@@ -351,6 +351,11 @@
     "webpack-hot-middleware": "^2.18.0",
     "webpack-subresource-integrity": "^1.0.0-rc.1"
   },
+  "storybook-deployer": {
+    "gitUsername": "echo $GH_USER",
+    "gitEmail": "echo $GH_EMAIL",
```
rebmullin: I am not too sure about to refer to these vars (assuming these are set in travis?)

willdurand: can we use the defaults?

.travis.yml
```diff
@@ -81,3 +81,4 @@ jobs:
     - stage:
       <<: *tox
       env: TOXENV=dennis-lint
+  after_success: yarn storybook-deploy -- --ci
```
willdurand: Because we use yarn, I don't think we need `--`, we can directly pass arguments to the script:

```
yarn storybook-deploy --ci
```

In addition, we'll need a condition to only run `after_success` from the `master` branch.

package.json
rebmullin: yeh I can remove this if we are leaving the defaults.

.travis.yml
rebmullin: I added a condition. let me know if this works, otherwise I might need your help in adding a "stage"

willdurand: We should not have such a condition right now, because we are not able to test that way.

I'd suggest you to remove your last commit, close this PR, push your branch directly here and open a PR. It should deploy the storybook to `gh-pages`. If that works, you'll have to add the condition to only deploy on `master`.

willdurand:
```js
    - stage: Storybook deployment
      script: echo "Deploying to GitHub Pages..."
      after_success: yarn storybook-deploy --ci
```

rebmullin:
@willdurand, there is a gh-pages branch now:) 

*but* I am looking into why the url doesn't seem to work (even after some time): https://mozilla.github.io/addons-frontend/

willdurand:
@rebmullin that looks good to me as well, 🎉 

I don't know why it is not available at https://mozilla.github.io/addons-frontend/, it is probably related to some project's settings.

rebmullin:
@willdurand, also regarding the following: `/dev/null 2>&1` @muffinresearch mentioned potentially keeping this to be extra safe nothing's leaked. Are you ok with adding this back?

rebmullin:
@willdurand, I am also noticing travis is running more often :/ I am wondering if this is related to these changes? even outside of this PR I am noticing this... is that possible?

willdurand:
Looks like everything is working as expected now, so let's add a condition to only deploy on the `master` branch: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditional-builds.

I believe all we need is a `if` parameter for your newly added stage.

rebmullin:
> I don't think it is risky but why not. We won't know what will happen when things will break though.

If there is an issue I guess we can always turn it off temporarily

> I am also noticing travis is running more often :/ I am wondering if this is related to these changes? even outside of this PR I am noticing this... is that possible?

Maybe it's actually not travis but there are codecov reports that show up many times in a PR it seems.

willdurand:
> Maybe it's actually not travis but there are codecov reports that show up many times in a PR it seems.

it is codecov and it is annoying.. nothing we can do thugh.

package.json
```diff
@@ -351,6 +351,11 @@
     "webpack-hot-middleware": "^2.18.0",
     "webpack-subresource-integrity": "^1.0.0-rc.1"
   },
+  "storybook-deployer": {
+    "gitUsername": "Custom Username",
+    "gitEmail": "custom@email.com",
+    "commitMessage": "Deploy Storybook to Github Pages [skip ci]"
```
willdurand: Can you only keep the `commitMessage` if that works? Also, would be nice to put `[skip ci]` as a prefix and not as a suffix.

PR #6273, opened 2018-09-18
Add config for withExperiment HOC

### LINKED ISSUE #6274: A/B tests should be controlled by config flags
The A/B test for the hero page doesn't currently have a config flag. One should be added since this will make it easier to enable in various envs as well as ensuring if we need to turn it off it's easy to do so.

willdurand:
Fixes #6274

---

This PR adds two features:

1. make the `witExperiment` HOC config-aware in a generic way
2. disable the home hero experiment by default, and enable it in -dev only

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -108,6 +108,7 @@ module.exports = {
     'enableExperienceSurvey',
     'enableInlineAddonReview',
     'enableStaticThemes',
+    'experiments',
     'fxaConfig',
     'isDeployed',
     'isDevelopment',
@@ -312,4 +313,10 @@ module.exports = {
   enableAMInstallButton: false,
 
   enableInlineAddonReview: false,
+
+  // The withExperiment HOC relies on this config to enable/disable A/B
+  // experiments.
+  experiments: {
+    home_hero: false,
+  },
 };
--- a/src/core/withExperiment.js
+++ b/src/core/withExperiment.js
@@ -1,12 +1,15 @@
 /* @flow */
+import config from 'config';
 import * as React from 'react';
 import invariant from 'invariant';
 import cookie from 'react-cookie';
 
+import log from 'core/logger';
 import { getDisplayName } from 'core/utils';
 
 export type WithExperimentInjectedProps = {|
-  variant: string,
+  experimentEnabled: boolean,
+  variant: string | null,
 |};
 
 type CookieConfig = {|
@@ -23,6 +26,7 @@ type withExperimentProps = {|
 type withExperimentInternalProps = {|
   ...withExperimentProps,
   WrappedComponent: React.ComponentType<any>,
+  _config: typeof config,
   _cookie: typeof cookie,
   randomizer: () => number,
 |};
@@ -43,6 +47,7 @@ export const withExperiment = ({
     experimentCookie: string | void;
 
     static defaultProps = {
+      _config: config,
       _cookie: cookie,
       id: defaultId,
       randomizer: Math.random,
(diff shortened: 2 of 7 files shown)


config/default.js
```diff
@@ -312,4 +313,10 @@ module.exports = {
   enableAMInstallButton: false,
 
   enableInlineAddonReview: false,
+
+  // The withExperiment HOC relies on this config to enable/disable A/B
+  // experiments.
+  experiments: {
```
rebmullin: should we have this match the hoc name: `withExperiments: { .. }`

src/core/withExperiment.js
```diff
@@ -1,12 +1,15 @@
...
 import invariant from 'invariant';
 import cookie from 'react-cookie';
 
+import log from 'core/logger';
 import { getDisplayName } from 'core/utils';
 
 export type WithExperimentInjectedProps = {|
-  variant: string,
+  experimentEnabled: boolean,
+  variant: string | null,
```
rebmullin: I feel like this should be undefined (string | void) if not being used, no?

willdurand: if it is undefined, we cannot determine whether the prop has been passed or forgotten.

config/default.js
```diff
@@ -312,4 +313,10 @@ module.exports = {
...
+    home_hero: false,
```
rebmullin: do we want this to be false? or do we want to make sure to set this on for stage and prod? I guess I was under the impression this was supposed to go live this week?

willdurand: There is no rush and no deadline. This will go live next week I suppose.

config/default.js
willdurand: mmh, I think it is shorter to use `experiments` and to have a list of experiments here. The name `withExperiment` is an implementation detail to me.

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -1,61 +1,67 @@
...
-    const root = shallowRender();
+    const root = render();
     const carousel = root.find(Hero);
 
     expect(carousel).toHaveLength(1);
     expect(carousel).toHaveProp('random', true);
   });
 
-  it('sends a tracking event when first rendered on the client', () => {
+  it('sends a tracking event when first render on the client', () => {
```
rebmullin: I feel like this sounds a little off maybe..Can we put back rendered or maybe ~ "sends a tracking event *on* first client render" ?

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -96,4 +102,37 @@ describe(__filename, () => {
       label: firstHeroTitle,
     });
   });
+
+  it('does not send a tracking event on first render when config experiment is disabled', () => {
```
rebmullin: should we have a test case for the variant - to show it's null (or undefined as suggested above) somewhere here?

rebmullin: if disabled..

willdurand: mmh, probably yeah

willdurand: added, thanks!

src/core/withExperiment.js
rebmullin: mmm..ok..

config/default.js
rebmullin: ok I'll defer to you : )

PR #6276, opened 2018-09-18
Fix some flow definitions for AMInstallButton

willdurand:
I am working on a bigger patch that brings flow to disco pane. Doing this made me discover these issues.

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -386,6 +386,7 @@ export class AddonBase extends React.Component {
       install,
       installStatus,
       installTheme,
+      setCurrentStatus,
       uninstall,
       userAgentInfo,
     } = this.props;
@@ -535,6 +536,7 @@ export class AddonBase extends React.Component {
                       hasAddonManager={hasAddonManager}
                       install={install}
                       installTheme={installTheme}
+                      setCurrentStatus={setCurrentStatus}
                       status={installStatus}
                       uninstall={uninstall}
                       isAddonEnabled={isAddonEnabled}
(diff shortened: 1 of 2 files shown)


src/core/components/AMInstallButton/index.js
```diff
@@ -45,23 +45,23 @@ import type { ReactRouterLocationType } from 'core/types/router';
 import './styles.scss';
 
 type Props = {|
+  ...WithInstallHelpersInjectedProps,
   addon: AddonType,
   className?: string,
   defaultInstallSource: string,
   disabled: boolean,
-  location: ReactRouterLocationType,
```
rebmullin: should you setCurrentStatus that you have added above?

src/amo/pages/Addon/index.js
```diff
@@ -535,6 +536,7 @@ export class AddonBase extends React.Component {
                       hasAddonManager={hasAddonManager}
                       install={install}
                       installTheme={installTheme}
+                      setCurrentStatus={setCurrentStatus}
```
rebmullin: I don't see this being used in the AMInstallButton - is that something you're adding in a later PR?

willdurand: nope, but it needs to be passed.

src/core/components/AMInstallButton/index.js
willdurand: it is part of `WithInstallHelperInjectedProps`.

rebmullin: ahh..I see..cool 👍

PR #6278, opened 2018-09-18
Add flow type to disco store

### LINKED ISSUE #6279: Add flow types to Disco store
Supports #2917 

---

In order to make the disco pane code easier to update (and more readable), we should have more flow coverage. This issue is about adding flow types to the disco store.

willdurand:
Fix #6279

---

It is a small step for a better disco app.

### DIFF
--- a/src/core/types/store.js
+++ b/src/core/types/store.js
@@ -1,0 +1,14 @@
+import type { ReactRouterHistoryType } from 'core/types/router';
+
+// This is a type function that takes a state type and returns a reducer
+// type, i.e. a function that accepts and returns the same state type.
+/* eslint-disable no-undef */
+export type CreateReducerType = <AnyState>(
+  AnyState,
+) => (AnyState, action: Object) => AnyState;
+/* eslint-enable no-undef */
+
+export type CreateStoreParams = {|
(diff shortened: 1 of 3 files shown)


kumar303:
Awesome! Is there an issue for this? Since we don't use PRs themselves for tracking I think it always helps to add an issue (even just a one-liner).

willdurand:
> Awesome! Is there an issue for this? Since we don't use PRs themselves for tracking I think it always helps to add an issue (even just a one-liner).

Ah right sorry, yes I filed one for this and will open new ones for the other bits.

src/core/types/store.js
```diff
@@ -0,0 +1,14 @@
+import type { ReactRouterHistoryType } from 'core/types/router';
+
+// This is a type function that takes a state type and returns a reducer
+// type, i.e. a function that accepts and returns the same state type.
+/* eslint-disable no-undef */
```
kumar303: You could try removing this to see what happens. It's some kind of eslint/flow bug.

willdurand: it does not work :( I tried and it complains about `AnyState` not being defined..

Issue #6286, opened 2018-09-18
Tighten up home page shelf on mobile


kumar303:
As a follow-up to https://github.com/mozilla/addons-frontend/issues/4087 , the shelf needs tightening up to match the mock provided in https://github.com/mozilla/addons-frontend/issues/4087#issuecomment-370063825

ioanarusiczki:
Verified fixed on AMO -dev with FF62 (Android 8.0)

![first section on android](https://user-images.githubusercontent.com/33448286/45950870-19cef680-c00a-11e8-93a7-86154ccf286e.png)

PR #6287, opened 2018-09-18
Tighten up home page shelf on mobile

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6286

**Before**

screenshot 2018-09-18 17 16 16

**After**

screenshot 2018-09-18 17 16 38

Here are some other screens with the new patch:

screenshot 2018-09-18 17 17 55
screenshot 2018-09-18 17 18 19

### DIFF
--- a/src/amo/pages/Home/styles.scss
+++ b/src/amo/pages/Home/styles.scss
@@ -82,11 +82,12 @@
 .Home-SubjectShelf-list-item {
   flex-grow: 1;
   list-style-type: none;
-  margin: 15px 0;
+  margin: 0;
   text-align: center;
   width: 100%;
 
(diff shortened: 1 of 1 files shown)


src/amo/pages/Home/styles.scss
```diff
@@ -82,11 +82,12 @@
 .Home-SubjectShelf-list-item {
   flex-grow: 1;
   list-style-type: none;
-  margin: 15px 0;
+  margin: 0;
```
rebmullin: can you just leave out margin if it's 0 here?

kumar303: Ah, yes I can.

PR #6293, opened 2018-09-19
Add flow types to disco reducers and sagas

### LINKED ISSUE #6292: Add flow types to Disco reducers and sagas
Supports #2917 

---

In order to make the disco pane code easier to update (and more readable), we should have more flow coverage. This issue is about adding flow types to the disco reducers and sagas.

willdurand:
Fix #6292


src/disco/reducers/discoResults.js
```diff
@@ -1,11 +1,95 @@
-import { LOAD_DISCO_RESULTS } from 'disco/constants';
+/* @flow */
+import invariant from 'invariant';
 
-export default function discoResults(state = [], { type, payload }) {
-  switch (type) {
+import type { AddonType, ExternalAddonType } from 'core/types/addons';
+
+export const GET_DISCO_RESULTS: 'GET_DISCO_RESULTS' = 'GET_DISCO_RESULTS';
```
rebmullin: could this be called FETCH_DISCO_RESULTS... it's more consistent with other get/fetch names that I am seeing

src/disco/reducers/discoResults.js
```diff
@@ -1,11 +1,95 @@
...
+
+  return {
+    type: LOAD_DISCO_RESULTS,
+    payload: { entities, result },
+  };
+}
+
+type Action = LoadDiscoResultsAction | GetDiscoResultsAction;
+
+export default function discoResults(
```
rebmullin: can we call this reducer? it seems that we are doing this in other places...

src/disco/reducers/discoResults.js
```diff
@@ -1,11 +1,95 @@
...
+  state: DiscoResultsState = initialState,
+  action: Action,
+): DiscoResultsState {
+  switch (action.type) {
     case LOAD_DISCO_RESULTS: {
```
rebmullin: should we explicitly add GET_DISCO_RESULTS here?

src/disco/reducers/discoResults.js
willdurand: In a follow-up, yep. I tried to not change any code here, just moved things and added flow types.

src/disco/reducers/discoResults.js
willdurand: We don't have anything to change so not sure..

src/disco/reducers/discoResults.js
willdurand: I tend to agree with you but it is not consistent.. there are reducers that are named.

rebmullin: ok I just thought our eventual goal was [this](https://github.com/mozilla/addons-frontend/issues/1299) so I figured why you were changing this, we should update:)

I guess this is still moving in this direction tho so 👏

willdurand: Yeah, naming it `reducer` or not does not really matter. It is now a Duck(s) :)

src/core/reducers/addons.js
```diff
@@ -24,16 +25,16 @@ type ExternalAddonMap = {
...
 |};
 
 // TODO: We should remove this method and move all calls to `loadAddonResults`.
 // This function relies on normalizr messing with our response data.
 // See: https://github.com/mozilla/addons-frontend/issues/2917
-export function loadAddons(entities: {|
-  addons?: ExternalAddonMap,
-|}): LoadAddonsAction {
+export function loadAddons(entities: {
+  addons: ExternalAddonMap,
```
rebmullin: should this be ` ExternalAddonMap | ExternalDiscoAddonMap` too?

src/disco/reducers/discoResults.js
rebmullin: ok cool, I wasn't sure if that mattered..cool tho now I know :)

src/core/reducers/addons.js
willdurand: ah good catch, added, thanks!

PR #6298, opened 2018-09-19
Add flow types to Addon disco component

### LINKED ISSUE #6294: Add flow types to Addon disco component
Supports #2917 

---

In order to make the disco pane code easier to update (and more readable), we should have more flow coverage. This issue is about adding flow types to the `Addon` disco component.

willdurand:
Fixes #6294 

---

The patch for normalizr should hopefully fix more test code. It is hard to recover from the "spread all the properties"-gate.

### DIFF
--- a/src/core/reducers/installations.js
+++ b/src/core/reducers/installations.js
@@ -20,13 +20,10 @@ export type InstalledAddon = {
   error?: string,
   guid: $PropertyType<AddonType, 'guid'>,
   needsRestart?: boolean,
-  // TODO: merge with core/constants.validInstallStates
-  // once that file supports Flow.
-  status?:
+  status:
     | 'DISABLED'
     | 'DISABLING'
     | 'DOWNLOADING'
-    | 'ENABLED'
     | 'ENABLED'
     | 'ENABLING'
     | 'ERROR'
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -54,7 +54,7 @@ type InternalProps = {|
   clientApp: string,
   defaultInstallSource: string,
   error: string | void,
-  getBrowserThemeData: Function,
+  getBrowserThemeData: () => string,
   i18n: I18nType,
   status: $PropertyType<InstalledAddon, 'status'>,
   userAgentInfo: UserAgentInfoType,
--- a/src/disco/components/Addon/index.js
+++ b/src/disco/components/Addon/index.js
@@ -1,8 +1,7 @@
-/* eslint-disable react/no-danger */
+/* @flow */
 import config from 'config';
 import makeClassName from 'classnames';
 import * as React from 'react';
-import PropTypes from 'prop-types';
 import { CSSTransition, TransitionGroup } from 'react-transition-group';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
@@ -20,7 +19,6 @@ import {
   UNINSTALLED,
   UNKNOWN,
   validAddonTypes,
-  validInstallStates,
 } from 'core/constants';
 import translate from 'core/i18n/translate';
 import { withInstallHelpers } from 'core/installAddon';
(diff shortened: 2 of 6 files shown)


tests/unit/disco/components/TestAddon.js
```diff
@@ -289,20 +284,24 @@ describe(__filename, () => {
     });
 
     it('does render a logo for an extension', () => {
-      const root = renderAddon({ addon: result });
+      const { heading, ...addon } = result;
+      const root = renderAddon({ addon, heading });
```
rebmullin: it's not clear to me why you have to pass in heading here since it's not part of the test.

tests/unit/disco/components/TestAddon.js
```diff
@@ -289,20 +284,24 @@ describe(__filename, () => {
...
 
       expect(root.find('.logo')).toHaveLength(1);
     });
 
     it("doesn't render a theme image for an extension", () => {
-      const root = renderAddon({ addon: result });
+      const { heading, ...addon } = result;
+      const root = renderAddon({ addon, heading });
```
rebmullin: it's not clear to me why you have to pass in heading here since it's not part of the test.

tests/unit/disco/components/TestAddon.js
```diff
@@ -289,20 +284,24 @@ describe(__filename, () => {
...
 
       expect(root.find('.Addon-ThemeImage-link')).toHaveLength(0);
       expect(root.find(ThemeImage)).toHaveLength(0);
     });
 
     it('throws on invalid add-on type', () => {
-      const root = renderAddon({ addon: result });
+      const { heading, ...addon } = result;
+      const root = renderAddon({ addon, heading });
```
rebmullin: same here

tests/unit/disco/components/TestAddon.js
```diff
@@ -195,7 +195,8 @@ describe(__filename, () => {
     });
 
     it('renders the heading', () => {
-      const root = renderAddon({ addon: result });
+      const { heading, ...addon } = result;
+      const root = renderAddon({ addon, heading });
```
rebmullin: I think if you're showing the heading as passed in props it should override the default so it's more clear otherwise, can you just leave that out here?

src/disco/components/Addon/index.js
```diff
@@ -20,60 +19,56 @@ import {
...
+type InternalProps = {|
+  ...Props,
+  ...WithInstallHelpersInjectedProps,
+  _config: typeof config,
+  _getClientCompatibility: typeof getClientCompatibility,
+  _tracking: typeof tracking,
+  clientApp: string,
+  defaultInstallSource: string,
+  error: string | void,
+  getBrowserThemeData: Function,
```
rebmullin: nothing more specific than Function?

src/disco/components/Addon/index.js
```diff
@@ -158,7 +154,7 @@ export class AddonBase extends React.Component {
     );
   }
 
-  installTheme = (event) => {
+  installTheme = (event: SyntheticEvent<any>) => {
```
rebmullin: I guess we can't be any more specific than this?

tests/unit/disco/components/TestAddon.js
willdurand: here, I need it because of the first expectation L305.

tests/unit/disco/components/TestAddon.js
willdurand: I actually fixed the expected value :)

src/core/reducers/installations.js
```diff
@@ -20,14 +20,11 @@ export type InstalledAddon = {
   error?: string,
   guid: $PropertyType<AddonType, 'guid'>,
   needsRestart?: boolean,
-  // TODO: merge with core/constants.validInstallStates
```
kumar303: Did you consider doing this? You could use `$ObjMap` but it would require a change to `validInstallStates` in `core/constants`.

src/disco/components/Addon/index.js
willdurand: Not sure which element would make sense here. Since it is not used, I don't know..

src/disco/pages/DiscoPane/index.js
```diff
@@ -167,7 +167,8 @@ export class DiscoPaneBase extends React.Component<InternalProps> {
         {this.renderFindMoreButton({ position: 'top' })}
 
         {results.map((item) => (
-          <Addon addon={item} key={item.guid} />
+          // $FLOW_FIXME: addon should be a AddonType
```
kumar303: could you provide a type hint like `results.map((item: AddonType) => ...)`. Otherwise, you lose a lot of coverage on this line.

tests/unit/disco/components/TestAddon.js
rebmullin: yeh the test passes without the heading being passed in (I think)..I believe b/c it's in the default props..

yep, I did see L305 but it wasn't clear to me why that was there either 😕

src/disco/pages/DiscoPane/index.js
willdurand: are you sure given that `results` is typed?

src/core/reducers/installations.js
willdurand: I double checked that both were in sync, and decided to use `$PropertyType<InstalledAddon, 'status'>`.

src/disco/pages/DiscoPane/index.js
willdurand: Ah I see what you mean I guess, well not now but soon-ish everything will have a correct type. Right now, it is being weird. I'll try your idea though.

willdurand: That creates a huge list of Flow errors :/

src/core/reducers/installations.js
kumar303: How do you mean? I still see duplication:

This:

https://github.com/mozilla/addons-frontend/blob/21a551e4e993e8b5b9b012588537cdea58eb138b/src/core/reducers/installations.js#L23-L35

Versus:

https://github.com/mozilla/addons-frontend/blob/21a551e4e993e8b5b9b012588537cdea58eb138b/src/core/constants.js#L16-L30


The comment you removed was about making them share a common type.

willdurand: The constant is still being used in the `InstallSwitch`. I cannot remove the constant entirely yet.

src/disco/pages/DiscoPane/index.js
willdurand: In any case, this was (and still is) fixed in #6325.

src/core/reducers/installations.js
willdurand: Yes!

PR #6315, opened 2018-09-20
Change discoResults state from array to object

### LINKED ISSUE #6314: Fix discoResults state
Supports #2917 

---

In order to make the disco pane code easier to update (and more readable), we should have a better state for the `discoResults` reducer. Currently, this reducer has a non-object state (array), which makes it difficult to add new properties to the state.

This issue is about changing the state shape from array to object.

For QA: please make sure Disco Pane still lists the add-ons as usual.

willdurand:
Fix #6314


willdurand:
thanks!

PR #6319, opened 2018-09-20
Add flow types to DiscoPane (disco) component

### LINKED ISSUE #6318: Add flow types to DiscoPane (disco) component
Supports #2917 

---

In order to make the disco pane code easier to update (and more readable), we should have more flow coverage. This issue is about adding flow types to the `DiscoPane` disco component.

willdurand:
Fixes #6318

---

I also removed some unused code.

### DIFF
--- a/src/disco/pages/DiscoPane/index.js
+++ b/src/disco/pages/DiscoPane/index.js
@@ -1,7 +1,7 @@
+/* @flow */
 /* global navigator */
 import invariant from 'invariant';
 import * as React from 'react';
-import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 import config from 'config';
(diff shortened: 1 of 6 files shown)


src/disco/pages/DiscoPane/index.js
```diff
@@ -18,36 +18,52 @@ import { NAVIGATION_CATEGORY } from 'disco/constants';
...
-      }).isRequired,
-    }).isRequired,
-    results: PropTypes.arrayOf(PropTypes.object).isRequired,
-    _addChangeListeners: PropTypes.func,
-    _tracking: PropTypes.object,
-  };
-
+type Props = {|
+  location: ReactRouterLocationType,
+  match: {|
```
rebmullin: seems like these should be in InteralProps, no?

src/disco/pages/DiscoPane/index.js
```diff
@@ -18,36 +18,52 @@ import { NAVIGATION_CATEGORY } from 'disco/constants';
...
+  |},
+|};
+
+type InternalProps = {|
+  ...Props,
+  _addChangeListeners: typeof addChangeListeners,
+  _tracking: typeof tracking,
+  dispatch: DispatchFunc,
+  errorHandler: ErrorHandlerType,
+  handleGlobalEvent: Function,
```
rebmullin: is it better to be a little more specific ~ (Object) => void ?

src/disco/pages/DiscoPane/index.js
willdurand: Not here because this is a page-level component, which gets the `location` and `match` props from the `Route` component in the `Routes` component (if that makes sense :D)

PR #6325, opened 2018-09-20
Remove normalizr in disco app

### LINKED ISSUE #6328: Create explicit API data mappings (i.e. replace normalizr) for Disco Pane
We use magical schemas (via [normalizr](https://github.com/paularmstrong/normalizr)) to store API data in a mappable format. Sometimes this helps look up objects better but other times we have code that *unmaps* it entirely.

Instead of using `normalizr`, let's just map what we need explicitly in our reducers. It might require more code but the mapping logic will be a lot easier to follow. This is what our reducers are for!

---

Let's do this for Disco Pane!

For QA: please make sure the Disco Pane works as expected (listing of add-ons, installs)

willdurand:
Fixes #6328

~~Depends on #6298~~
~~Depends on #6315~~
~~Depends on #6319~~

---

With the other PRs merged, this will be easier to review! I am sorry for the big diff though, it was waaay smaller until I got stuck with the `tests/unit/disco/components/TestAddon.js` file, which I entirely rewrote with `shallowUntilTarget`.

### DIFF
--- a/src/core/actions/installations.js
+++ b/src/core/actions/installations.js
@@ -31,7 +31,6 @@ export const setInstallError = ({
   error,
 }: SetInstallErrorParams): SetInstallErrorAction => {
   invariant(guid, 'guid is required');
-  invariant(error, 'error is required');
 
   return {
     type: INSTALL_ERROR,
--- a/src/core/installAddon.js
+++ b/src/core/installAddon.js
@@ -559,7 +559,6 @@ export class WithInstallHelpers extends React.Component<
       WrappedComponent,
       _addonManager,
       _installTheme,
-      _tracking,
       ...passThroughProps
     } = this.props;
 
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -21,7 +21,6 @@ import type {
   ExternalAddonType,
   ThemeData,
 } from 'core/types/addons';
-import type { ExternalDiscoAddonMap } from 'disco/reducers/discoResults';
 
 export const LOAD_ADDONS: 'LOAD_ADDONS' = 'LOAD_ADDONS';
 export const FETCH_ADDON: 'FETCH_ADDON' = 'FETCH_ADDON';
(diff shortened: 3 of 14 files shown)


src/core/actions/installations.js
```diff
@@ -31,7 +31,6 @@ export const setInstallError = ({
   error,
 }: SetInstallErrorParams): SetInstallErrorAction => {
   invariant(guid, 'guid is required');
-  invariant(error, 'error is required');
```
willdurand: This is needed because an ancient test case was explicitly testing an install error with no error message.

src/core/installAddon.js
```diff
@@ -559,7 +559,6 @@ export class WithInstallHelpers extends React.Component<
       WrappedComponent,
       _addonManager,
       _installTheme,
-      _tracking,
```
willdurand: This is needed to let the `_tracking` prop pass through this HOC.

src/core/reducers/addons.js
```diff
@@ -229,7 +228,11 @@ export function createInternalAddon(apiAddon: ExternalAddonType): AddonType {
 
   const currentVersion = apiAddon.current_version;
 
-  if (currentVersion && currentVersion.files.length > 0) {
+  if (
+    currentVersion &&
+    currentVersion.files &&
```
willdurand: One test case was failing because of this, not sure it is realistic but it does not hurt I guess.

willdurand:
(rebased)

willdurand:
(rebased)

src/disco/reducers/discoResults.js
```diff
@@ -63,26 +66,52 @@ export function getDiscoResults({
...
+  };
+};
+
+type CreateExternalAddonMapParams = {|
+  results: Array<ExternalDiscoResultType>,
+|};
+
+export const createExternalAddonMap = ({
+  results,
+}: CreateExternalAddonMapParams): ExternalAddonMap => {
```
kumar303: It's unfortunate to say that this returns `ExternalAddonMap` because the `addon` objects are missing a lot of properties. I know that if you type it correctly then Flow will explode so I'm not sure what to do. I worry that something in Discopane will break one day because it's relying on a non-existent property (such as `addon.description` or whatever).

tests/unit/disco/components/TestAddon.js
```diff
@@ -491,112 +610,74 @@ describe(__filename, () => {
...
+        id: 'some-theme-id',
+        name: 'some-theme-name',
         type: ADDON_TYPE_THEME,
-        previews: [],
       };
+
+      store.dispatch(
+        setInstallState({
+          // We use this function because the GUID is changed for lightweight
+          // themes (to mimic Firefox's internal behavior).
```
kumar303: Does the magic guid swapping still work? I am not set up very well to test it locally.

willdurand: Yep, it works because `createInternalAddon()` calls the `getGuid()` function.

PR #6326, opened 2018-09-20
Hide review listing if the only review is also the featured review

### LINKED ISSUE #6316: The header for the reviews list should be hidden when there is only one review available for an add-on
STR:

1. Open AMO stage
2. Go to an add-on with only one listed review
3. Click on the review permalink - see https://addons.allizom.org/en-US/firefox/addon/vladify-1-0-1/reviews/715929/

Actual result:
The reviews list header is displayed even if there are no other reviews listed for this add-on. Also, the header doesn't have a border-bottom-radius

Expected result:
The header should be hidden when there are no other reviews listed for an add-on

Notes:
- reproduced on -dev and stage with FF62, WIn10x64

image

bobsilverberg:
Fixes #6316

### DIFF
--- a/src/amo/pages/AddonReviewList/index.js
+++ b/src/amo/pages/AddonReviewList/index.js
@@ -213,20 +213,20 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
       header = <LoadingText />;
     }
 
-    const addonRatingCount =
+    const addonReviewCount =
       addon && addon.ratings ? addon.ratings.text_count : null;
     let addonName;
     let reviewCountHTML;
-    if (addon && addonRatingCount !== null) {
+    if (addon && addonReviewCount !== null) {
       addonName = <Link to={this.addonURL()}>{addon.name}</Link>;
       reviewCountHTML = i18n.sprintf(
         i18n.ngettext(
           '%(total)s review for this add-on',
           '%(total)s reviews for this add-on',
-          addonRatingCount,
+          addonReviewCount,
         ),
         {
-          total: i18n.formatNumber(addonRatingCount),
+          total: i18n.formatNumber(addonReviewCount),
         },
       );
     } else {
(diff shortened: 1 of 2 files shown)


src/amo/pages/AddonReviewList/index.js
```diff
@@ -333,29 +333,36 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
...
+          (!reviewId ||
+            allReviews.length > 1 ||
+            allReviews[0].id.toString() !== reviewId) && (
+            <CardList
+              className="AddonReviewList-reviews-listing"
+              footer={paginator}
+              header={reviewCountHTML}
+            >
+              <ul>
+                {allReviews.map((review, index) => {
```
rebmullin: what about using filter and map (or even reduce):

may be better/cleaner to just get params instead of this includes way but this is just an example:

```javascript
const reviewListing = allReviews
      .filter((item) => item && !this.props.location.pathname.includes(item.id))
      .map((review, index) => {
        // TODO: Remove this and use the API to filter out the featured review once
        // https://github.com/mozilla/addons-server/issues/9424 is fixed.

        return (
          <li key={String(index)}>
            <AddonReviewCard addon={addon} review={review} />
          </li>
        );
      });

....

<ul>{reviewsListing}</ul>
```

src/amo/pages/AddonReviewList/index.js
```diff
@@ -333,29 +333,36 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
...
-                    <li key={String(index)}>
-                      <AddonReviewCard addon={addon} review={review} />
-                    </li>
-                  );
-                }
-                return null;
-              })}
-            </ul>
-          </CardList>
+          {// TODO: remove this condition after
```
kumar303: This is totally fine as is but you can also put comments in JSX like `{/* some long comment */}`, FYI.

src/amo/pages/AddonReviewList/index.js
```diff
@@ -333,29 +333,36 @@ export class AddonReviewListBase extends React.Component<InternalProps> {
...
+          //  https://github.com/mozilla/addons-server/issues/9424 is fixed
+          //  and integrated.
+          (!reviewId ||
```
kumar303: This if statement is pretty complicated. How about just replacing `allReviews.map((review, index) => ...)` (below) with a variable? Then you can simply check the length.

src/amo/pages/AddonReviewList/index.js
bobsilverberg: This is nice, but the issue is that we want to hide the entire card.

src/amo/pages/AddonReviewList/index.js
bobsilverberg: I'm not sure I follow what you mean, but I think I can take your idea *and* @rebmullin's idea and combine them. If I filter the array first then I can just check its length. Maybe that's what you meant.

src/amo/pages/AddonReviewList/index.js
rebmullin: right this is just for the inner mapping code....if that makes sense..sorry if that wasn't clear..

bobsilverberg: Ah, sorry. I think you inspired me to a better solution anyway. :-)

src/amo/pages/AddonReviewList/index.js
kumar303: Yes, I think that's the same idea.

Issue #6272, opened 2018-09-18
Bigger stars for editing ratings for the review listing page


ioanarusiczki:
STR:
1. Load AMO dev and from an add-ons detail page make a review.
2. Open the review listing page and click the Edit my review button.
3. Try to modify the rating -tap on a star.
4. Observe the rating stars.

Expected:
Stars are easy to tap without zoom in.

Actual:
Stars are small andnot too easy to tap without zoom in.

Notes:
Reproducible on AMO dev with FF62(Android 8.0)

![android stars](https://user-images.githubusercontent.com/33448286/45678810-e3900380-bb3f-11e8-8b22-8073238d3355.png)

PR #6327, opened 2018-09-20
Bigger rating stars on small screens when editing a review

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6272

<hr/>

**Before**

screenshot 2018-09-20 09 59 23

<hr/>

**After**

screenshot 2018-09-20 09 57 58

<hr/>

In other locales:

screenshot 2018-09-20 09 57 01
screenshot 2018-09-20 09 58 26

On larger screens it's the same as before:

screenshot 2018-09-20 09 58 54
screenshot 2018-09-20 09 57 29

### DIFF
--- a/src/amo/components/AddonReviewManager/styles.scss
+++ b/src/amo/components/AddonReviewManager/styles.scss
@@ -5,13 +5,6 @@
 .AddonReviewManager {
   font-size: $font-size-default;
 
-  .AddonReviewManager-Rating.Rating {
-    height: $default-line-height;
-    margin: 0;
-
-    @include margin-start(6px);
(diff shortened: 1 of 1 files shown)


kumar303:
> just checking we don't want this for larger tap devices (like ipad, etc)...?

Hmm, I'm not sure. It looks a little odd to me (screenshot below). In this edit state, the focus is on the text entry form which is why I think the small stars work well. Maybe we can just address it in a future issue if it's a problem.

screenshot 2018-09-20 14 48 19

kumar303:
Thanks for the review

PR #6332, opened 2018-09-20
Simplify AddonCompatibilityError disco component

### LINKED ISSUE #6331: Disco Pane's AddonCompatibilityError component does not need Redux
I was about to fix #6323 and took a look at the component used in Disco Pane (since one suggestion [here](https://github.com/mozilla/addons-frontend/issues/6240#issuecomment-423125561) was about reusing the message from Disco Pane).

I found a component that requires redux but does not use it. We can simplify this component a lot! (and add Flow types).

For QA: disco pane incompatibility errors should be displayed.

willdurand:
Fix #6331

### DIFF
--- a/tests/unit/disco/components/TestAddonCompatibilityError.js
+++ b/tests/unit/disco/components/TestAddonCompatibilityError.js
@@ -8,30 +8,18 @@ import {
   INCOMPATIBLE_NOT_FIREFOX,
   INCOMPATIBLE_UNDER_MIN_VERSION,
 } from 'core/constants';
-import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
 import { fakeI18n, shallowUntilTarget } from 'tests/unit/helpers';
 
 describe(__filename, () => {
   function render({ ...props }) {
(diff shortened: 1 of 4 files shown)


tests/unit/disco/components/TestAddonCompatibilityError.js
```diff
@@ -40,35 +28,29 @@ describe(__filename, () => {
...
 
     expect(content.text()).toContain(
       'Firefox for iOS does not currently support add-ons.',
     );
   });
 
-  it('renders a generic message when reason code not known', () => {
-    const root = render({ reason: 'fake reason' });
+  it('renders a generic message when not FF', () => {
+    const root = render({ reason: INCOMPATIBLE_NOT_FIREFOX });
```
bobsilverberg: Nit: We don't technically need this test because there is no code path for this reason. This will take the same path as `render({ reason: 'fake reason' })`, but I suppose it's nice to have a test for this as this is a reason that we could expect to actually be set. I'll leave it up to you whether to include this test or remove it.

PR #6340, opened 2018-09-21
Expose experiments and feature flags on /__frontend_version__

### LINKED ISSUE #6363: Expose feature flags and experiments on `/__frontend_version__`
In order to quickly know/verify which features and/or experiments (A/B) are enabled/disabled for a given app and env, let's expose them on `/__frontend_version__`, which is an existing endpoint that returns a JSON response with some version numbers.

willdurand:
Fixes #6363 

~~⚠️ depends on #6364~~

---

This PR exposes the feature flags and experiments on `/__frontend_version__` (as discussed with @kumar303 and @bqbn on IRC).

~~It has a white-list for config variables we do not want to expose given that our current prefix for feature flags is `enable`. We could probably use a `feature_flags` object parameter (like `experiments`) to avoid accidental exposure.~~ => we adopted the `enableFeature` prefix in #6362.

screen shot 2018-09-21 at 12 31 51


willdurand:
@kumar303 I requested a review from you to get early feedback on this. I wonder if we should move feature flags to a namespace (object config parameter) first to avoid the white-list and the need for a prefix. That works as is though and I find it quite elegant so we can still iterate on it.

tests/unit/core/utils/test_server.js
```diff
@@ -0,0 +1,90 @@
...
+
+import {
+  getFeatureFlagName,
+  viewFrontendVersionHandler,
+} from 'core/utils/server';
+import { getFakeConfig } from 'tests/unit/helpers';
+
+describe(__filename, () => {
+  describe('viewFrontendVersionHandler', () => {
+    const basePath = path.join(__dirname, '__fixtures__');
```
kumar303: nit: I don't think `__fixtures__` needs so many underscores here because there is no chance of naming collisions.

src/core/server/base.js
```diff
@@ -201,7 +202,7 @@ function baseServer(
   // Following the ops monitoring convention, return version info at this URL.
```
kumar303: While you're in here could you mention that this is specifically a Dockerflow convention? And link to https://github.com/mozilla-services/Dockerflow

tests/unit/helpers.js
```diff
@@ -442,9 +442,9 @@ export function createFakeUserAbuseReport({
 // if (fakeConfig.get('isDevelopment')) {
 //   ...
 // }
-export const getFakeConfig = (params = {}) => {
+export const getFakeConfig = (params = {}, forceKey = false) => {
```
kumar303: Because you never only add one option, I'd rather see it as an object. This also makes calling code easier to read.

```js
export const getFakeConfig = (params = {}, { forceKey = false } = {}) => {
  // ...
}
```

Additionally, since this function will be called in test code which doesn't have Flow coverage, optional positional arguments are dangerous (they are hard for refactoring).

kumar303: `allowUnknownKeys` would make more sense to me

src/core/utils/server.js
```diff
@@ -0,0 +1,72 @@
+/* @flow */
+import path from 'path';
+
+import fs from 'fs-extra';
+import config from 'config';
+import type { $Request, $Response } from 'express';
+
+import log from 'core/logger';
+
+const EXCLUDES_LIST = [
```
kumar303: > I wonder if we should move feature flags to a namespace (object config parameter) first to avoid the white-list

Hmm, yes, that would make it a lot easier to expose the flag parameters but would it make things awkward for overriding? How would you do it? Current you can do this:

https://github.com/mozilla/addons-frontend/blob/40d1ba4e8fe83bfdcf4bc44ba3854a5ba8a4ea61/config/dev-amo.js#L29

It's nice that you don't have to worry about other flags in that one line.

However, maybe it would also be nice to *explicitly* see the entire object of feature flags getting set for each env? I'm not sure.

tests/unit/core/utils/test_server.js
willdurand: works for me (old habit, sorry)

tests/unit/helpers.js
willdurand: works for me

src/core/utils/server.js
willdurand: I think I can make a change in `getFakeConfig()` to allow:

```js
getFakeConfig({ 'feature_flags.enableInlineAddonReview': true });
```

but that might introduce a bit of complexity..

An alternative would be to find a different prefix for feature flags (other than `enable`).

kumar303: `enableFeature` as a prefix looks like it would still keep the variables readable.

src/core/server/base.js
```diff
@@ -198,10 +199,11 @@ function baseServer(
     });
   }
 
-  // Following the ops monitoring convention, return version info at this URL.
+  // Following the ops monitoring Dockerflow convention, return version info at
+  // this URL. See: https://github.com/mozilla-services/Dockerflow
   app.get('/__version__', viewVersion);
```
kumar303: Why doesn't this use `viewFrontendVersionHandler()`? I don't think we should maintain separate implementations.

src/core/utils/server.js
```diff
@@ -0,0 +1,63 @@
...
+  const version = path.join(_config.get('basePath'), versionFilename);
+
+  const featureFlags = Object.keys(_config)
+    .filter((key) => {
+      return key.startsWith(PREFIX);
+    })
+    .reduce((map, key) => {
+      return {
+        ...map,
+        [getFeatureFlagName(key)]: _config.get(key),
```
kumar303: Can this just be `[key]: _config.get(key)`? It will still be readable. I think this would be better because someone may think to copy/paste what they see in the JSON and grep for it in the code.

src/core/utils/server.js
```diff
@@ -0,0 +1,63 @@
...
+          `Could not stat version file ${version}: ${error.toString()}`,
+        );
+        res.sendStatus(415);
+      } else {
+        const versionJson = await fs.readJson(version);
+
+        res.json({
+          ...versionJson,
+          experiments,
+          feature_flags: featureFlags,
```
kumar303: `featureFlags` (instead of `feature_flags`) would be more consistent with our code base.

src/core/server/base.js
willdurand: Alrighty, I can add it here too.

src/core/utils/server.js
willdurand: Yep, although `grep` would still work as it is only about removing a prefix, but why not.

src/core/utils/server.js
willdurand: This endpoint will be used by another app in the near future, so making it pretty could have been nice. The API we consume uses snake case for the JSON properties, so I just used that here too.

willdurand: I think the backend (which will eventually expose feature flags) will use snake case, so that would be nice to be consistent.

src/core/utils/server.js
willdurand: Thinking about it again, I am not sure it is wise to keep the (internal) prefix.

kumar303: I still don't see how it's useful to translate the config keys into shorter keys but I don't feel very strong about it. I am worried about confusing someone who would be looking at this output and thinking that it maps to config keys but I guess they would grep for the suffix and find what they're looking for anyway, so 🤷‍♀️

PR #6342, opened 2018-09-21
Reset the reviews view state when LOCATION_CHANGE is dispatched

### LINKED ISSUE #6300: The review form should not be left open when switching pages
Followup for https://github.com/mozilla/addons-frontend/issues/5944

STR:
1. Load AMO dev and make a review on the detail page for an add-on/theme
2. Click edit my review, observe that the form is opened
3. Click Read all reviews link
4. Observe the review text box 

Expected results:
The text box is closed - once edit my review is clicked it will open the form.

Actual results:
The text box is in edit mode - opened.

Notes: 
This issue is reproducible on AMO dev with FF62(Win10). 

the opened text box1

bobsilverberg:
Fixes #6300

### DIFF
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -1,5 +1,6 @@
 /* @flow */
 import { oneLine } from 'common-tags';
+import { LOCATION_CHANGE } from 'connected-react-router';
 
 import {
   DELETE_ADDON_REVIEW,
@@ -24,7 +25,6 @@ import {
   HIDE_REPLY_TO_REVIEW_FORM,
   createInternalReview,
 } from 'amo/actions/reviews';
-import { SET_VIEW_CONTEXT } from 'core/constants';
 import type {
   DeleteAddonReviewAction,
   FetchReviewAction,
@@ -51,7 +51,6 @@ import type {
 } from 'amo/actions/reviews';
 import type { GroupedRatingsType } from 'amo/api/reviews';
 import type { FlagReviewReasonType } from 'amo/constants';
-import type { ViewContextActionType } from 'amo/reducers/viewContext';
 import type { AppState } from 'amo/store';
 
 type ReviewsById = {
@@ -313,8 +312,7 @@ type ReviewActionType =
   | SetReviewReplyAction
   | SetUserReviewsAction
   | ShowEditReviewFormAction
-  | ShowReplyToReviewFormAction
-  | ViewContextActionType;
+  | ShowReplyToReviewFormAction;
 
 export default function reviewsReducer(
   state: ReviewsState = initialState,
@@ -579,7 +577,7 @@ export default function reviewsReducer(
       }
       return newState;
     }
-    case SET_VIEW_CONTEXT: {
+    case LOCATION_CHANGE: {
       return {
         ...state,
         view: {},
--- a/src/amo/reducers/reviews.js
+++ b/src/amo/reducers/reviews.js
@@ -24,6 +24,7 @@ import {
   HIDE_REPLY_TO_REVIEW_FORM,
   createInternalReview,
 } from 'amo/actions/reviews';
+import { SET_VIEW_CONTEXT } from 'core/constants';
 import type {
   DeleteAddonReviewAction,
   FetchReviewAction,
@@ -50,6 +51,7 @@ import type {
 } from 'amo/actions/reviews';
 import type { GroupedRatingsType } from 'amo/api/reviews';
 import type { FlagReviewReasonType } from 'amo/constants';
+import type { ViewContextActionType } from 'amo/reducers/viewContext';
 import type { AppState } from 'amo/store';
 
 type ReviewsById = {
@@ -311,7 +313,8 @@ type ReviewActionType =
   | SetReviewReplyAction
   | SetUserReviewsAction
   | ShowEditReviewFormAction
-  | ShowReplyToReviewFormAction;
+  | ShowReplyToReviewFormAction
+  | ViewContextActionType;
 
 export default function reviewsReducer(
   state: ReviewsState = initialState,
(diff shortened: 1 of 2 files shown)


tests/unit/amo/reducers/test_reviews.js
```diff
@@ -1219,4 +1221,20 @@ describe(__filename, () => {
...
+    it('resets the view state', () => {
+      let state;
+
+      state = reviewsReducer(
+        undefined,
+        showEditReviewForm({
+          reviewId: 1,
+        }),
+      );
+      state = reviewsReducer(undefined, setViewContext(VIEW_CONTEXT_HOME));
```
bobsilverberg: I thought about adding a check assertion here, to verify that `state.view` is not simply `{}` here, but we are already testing that in the test for `showEditReviewForm` above, so it's redundant.

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -1219,4 +1221,20 @@ describe(__filename, () => {
...
+  describe('setViewContext', () => {
+    it('resets the view state', () => {
+      let state;
+
+      state = reviewsReducer(
+        undefined,
+        showEditReviewForm({
+          reviewId: 1,
+        }),
+      );
```
bobsilverberg: I thought about adding a check assertion here, to verify that `state.view` is not simply `{}` here, but we are already testing that in the test for `showEditReviewForm` above, so it's redundant.

src/amo/reducers/reviews.js
```diff
@@ -576,6 +579,12 @@ export default function reviewsReducer(
       }
       return newState;
     }
+    case SET_VIEW_CONTEXT: {
```
kumar303: I see what you were trying to do here but relying on this action is fragile. It's not guaranteed that `setViewContext()` will be dispatched and, actually, I think it's a bug that it is (the context is the same between both views). 

`LOCATION_CHANGE` would be better because we know it gets dispatched whenever the URL changes. Example:

https://github.com/mozilla/addons-frontend/blob/40d1ba4e8fe83bfdcf4bc44ba3854a5ba8a4ea61/src/core/reducers/errorPage.js#L59

src/amo/reducers/reviews.js
```diff
@@ -576,6 +579,12 @@ export default function reviewsReducer(
...
+      return {
+        ...state,
+        view: {},
```
kumar303: This is brutal because it resets all view state for reviews / ratings, some of which may be used in unrelated components.

All we need to do is hide the form so I suggest `editingReview: false`. If we need to reset more things in the future we could but for now I think that's all we need.

By the way, I considered just hiding the form explicitly on unmount. This works but it breaks hot reloading so I don't think it's a good solution.

In `AddonReviewManager`, I added:

```js
componentWillUnmount() {
  const { review, dispatch } = this.props;
  dispatch(hideEditReviewForm({ reviewId: review.id }));
}
```

kumar303: Oh, I see how there's a problem with not having `reviewId`. 

I haven't looked into the implications of resetting the entire view state but if you think it's harmless then I suppose it will work.

The only problem with `componentWillUnmount()` is how it breaks hot reloading. That's not a deal breaker, just an annoyance. If that's a better way to solve it then it's fine.

bobsilverberg: I've looked through all of the code that uses the `reviews.view` state, and it seems to me that all of it is used for maintaining state on an individual page. I think if the location changes, that would render all of that data useless/unnecessary, so I do think it would be fine to clear the whole thing out on a `LOCATION_CHANGE` action. This will also avoid breaking hot reloading. I have made this change.

kumar303:
Please update this PR title before you merge it.

src/amo/reducers/reviews.js
```diff
@@ -576,6 +577,12 @@ export default function reviewsReducer(
       }
       return newState;
     }
+    case LOCATION_CHANGE: {
+      return {
+        ...state,
+        view: {},
```
kumar303: I'm looking at the side effects of this. I guess one is that the app will forget which reviews you had flagged (for spam, etc) as you navigate between listing pages. That's probably fine, though.

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -1219,4 +1221,21 @@ describe(__filename, () => {
...
+
+      state = reviewsReducer(undefined, { type: LOCATION_CHANGE });
```
kumar303: typo:

```diff
diff --git a/tests/unit/amo/reducers/test_reviews.js b/tests/unit/amo/reducers/test_reviews.js
index a6636968d..8ee9ae095 100644
--- a/tests/unit/amo/reducers/test_reviews.js
+++ b/tests/unit/amo/reducers/test_reviews.js
@@ -1233,7 +1233,7 @@ describe(__filename, () => {
         }),
       );
 
-      state = reviewsReducer(undefined, { type: LOCATION_CHANGE });
+      state = reviewsReducer(state, { type: LOCATION_CHANGE });
 
       expect(state.view).toEqual({});
     });
```

How did this test ever fail for you? When I asked if you watch your tests fail you said yes but now I'm giving you the side eye 🍌

src/amo/reducers/reviews.js
bobsilverberg: Thanks @kumar303. That is a bit of a pain, although it matches the behaviour if you reload the site as well (i.e., you don't see that reviews you've previously flagged are flagged unless you're in the same redux session), so I guess it's not too bad.

I could change the logic to keep the flagged view state intact and only reset the other data, but it would mean looping through all of the existing reviewIds. It sounds like you don't favour that over the current approach, but I just thought I'd check. Which do you prefer?

kumar303: I prefer your approach, no need to change it. Yeah, you could write some complex code to iterate through the entire view map and only reset `editingReview: false` but I don't think it's worth the complexity.

tests/unit/amo/reducers/test_reviews.js
kumar303: Moving quickly is good! We use code review to catch things like this (which doesn't always work, heh) -- I think it's great to keep moving fast. We also have a great QA team 😄

Issue #6352, opened 2018-09-21
Source maps to JS are broken for hosted servers


kumar303:
### Describe the problem and steps to reproduce it:

- Open the developer console
- load the home page to AMO on a hosted server (such as dev)

### What happened?

You see log messages that point to the compressed JS bundle:

```
Intl.NumberFormat exists           amo-83ad424c9df72dac3e49.js:1:886144
```

### What did you expect to happen?

You see log messages that point to the actual JS source. Here's what it looks like in our local development server:

```
Intl.NumberFormat exists           browser.js:179 
```

### Anything else we should know?

Here is an example of the JS bundle on dev (note that this changes on every re-deployment) https://addons-amo-dev-cdn.allizom.org/amo-83ad424c9df72dac3e49.js

As far as I can tell, our webpack configuration is correct. Thus, the source map should exist at https://addons-amo-dev-cdn.allizom.org/amo-83ad424c9df72dac3e49.js.map but it does not. It's a 404.

kumar303:
I can verify that this a problem in the code, not with anything ops-related.

ioanarusiczki:
@kumar303 I verified it on AMO dev with FF62 -win10
Everytime I reload AMO -dev I receive 
Intl.NumberFormat exists                                  amo-6c0035b542186f97c434.js:1:884745 

![console logs](https://user-images.githubusercontent.com/33448286/46021142-9c2de800-c0e8-11e8-9d83-586166d03e13.png)

kumar303:
Source maps are working for me in Firefox, Chrome, and Sentry.

Here is a better way to verify it:
* Go to https://addons-dev.allizom.org/en-US/firefox/simulate-client-error/
* Click *Go ahead, trigger an error*
* Open Tools > Web Developer > Web Console

If you should see an error pointing to the `index.js` file then source maps are working:

<img width="753" alt="screenshot 2018-09-26 10 11 38" src="https://user-images.githubusercontent.com/55398/46089653-9bdb2e80-c174-11e8-961e-b5154e16dd83.png">

I also verified that the source maps are working for Sentry, which is the main bug I was trying to solve.

ioanarusiczki:
@kumar303  I tried it with https://addons-dev.allizom.org/en-US/firefox/simulate-client-error/ and  I also received  index.js:16:2 - on AMO dev with FF62(Win10)


[...]

PR #6355, opened 2018-09-21
Fix source maps on hosted servers

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6352

I confirmed the fix before and after with this command:

```
NODE_APP_INSTANCE=amo NODE_ENV=production yarn build
```

### DIFF
--- a/webpack.prod.config.babel.js
+++ b/webpack.prod.config.babel.js
@@ -46,6 +46,7 @@ export default {
     }),
     // optimizations
     new UglifyJsPlugin({
+      sourceMap: true,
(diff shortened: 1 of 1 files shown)


webpack.prod.config.babel.js
```diff
@@ -46,6 +46,7 @@ export default {
     }),
     // optimizations
     new UglifyJsPlugin({
+      sourceMap: true,
```
willdurand: Since I removed this line accidentally, do you think we can add a comment?

PR #6364, opened 2018-09-24
Rename feature flags

### LINKED ISSUE #6362: Use `enableFeature` prefix for all feature flag config parameters
In order to better identify the feature flags among all the config parameters, let's adopt the `enableFeature` prefix for these config params. The idea of this issue is to rename all parameters.

For QA:

- static themes support should be enabled on all envs
- the new install button should be available in -dev and -stage on AMO, in -dev only on Disco Pane
- inline add-on review should be in -dev only (AMO)
- the experience survey should be disabled entirely
- add-on recommendations (AMO detail page) should be enabled

willdurand:
Fix #6362


config/default-amo.js
```diff
@@ -118,5 +118,5 @@ module.exports = {
   authTokenValidFor: 2592000, // 30 days
 
   // Turn on recommendations in details pages.
-  enableAddonRecommendations: true,
+  enableFeatureAddonRecommendations: true,
```
bobsilverberg: Nit: I know it's not really what this patch is about, but I wonder if it would make sense to remove any flags that are no longer needed. I don't think we really need this `enableAddonRecommendations` flag anymore, as it was just in place to control the feature during the initial development/rollout. Rather than make a bunch of changes to a flag that actually should be removed, it might be more efficient to just remove it.

willdurand: That's a good idea. Usually, we have an issue for that. If the issue is missing for this flag, then I'll file one and work on it. I don't want to introduce too much changes in this PR.

willdurand: See: #6374

PR #6371, opened 2018-09-24
Rename config file for -prod env

### LINKED ISSUE #6365: Fix production config filename to match ops naming

I read the production logs.

### What happened?

> WARNING: NODE_ENV value of 'prod' did not match any deployment config file names.

No config warning.

Related ops issue: https://github.com/mozilla-services/cloudops-deployment/pull/2441

It is not a critical issue given that the default config is our production config.

### Proposal

We should rename our production config files from `production.js` to `prod.js`.

willdurand:
Fixes #6365


willdurand:
> I'm a bit confused. If our current production config is actually `default.js`, why do we need this `prod.js` file at all? Is it actually being used by anything, and if so, will this name change break that?

@bobsilverberg it is loaded by `node-config`, which merges many different files: https://github.com/lorenwest/node-config/wiki/Configuration-Files#file-load-order to build the final config for a given environment.

`default.js` is the default config for ALL apps, then come a default config file for each app, then other files per environment and/or app.

The config files related to a hosted environment are used on our hosted environments. We have `dev.js`, `stage.js` and currently `production.js`. Those files are loaded at runtime by the server instances.

Ops did some changes related to how configuration is being loaded at runtime and it looks like they use "prod" for the -prod environment. We have to rename our `production.js` file to make sure it is loaded at runtime and get rid of the warning message.

I wrote "`default` is our production environment" because that's the approach we take, but we need those environment files so that the config is loaded successfully (without any warning). This `production.js` file has been there forever, and it is written that we should not make any change to it (as much as we can).

I don't think we rename `default.js` to `prod.js` though, because that would not be loaded all the time. So we have to keep a `default.js` for the shared config, and keep one env file for each hosted env (dev, stage, prod) to make sure the config is loaded without a warning.

This was never a problem before (like 3 weeks ago) because there was an issue with the environment variable loading the config (the value of this variable was not accurate). In this PR: https://github.com/mozilla-services/cloudops-deployment/pull/2441, this issue has been fixed but it introduced a small issue (which is fixed by this patch) in -prod, a fairly harmless bug though.

The "ops" patch makes sure we use `NODE_CONFIG_ENV` to load the config instead of `NODE_ENV`, which must be set to `production` all the time. `NODE_CONFIG_ENV` should be either: `dev`, `stage` or `prod`.

bobsilverberg:
@willdurand I notice that there are still a number of places where we have a `NODE_ENV` of `production` rather than `prod`, including some webpack config files and some documentation.

Should all of those be changed to `prod` as well?

willdurand:
I noticed that too, the best thing we can do is add the NODE_CONFIG_ENV var set to "prod".

bobsilverberg:
I'm not sure what you mean by "the best thing we can do is". Can we not change all of these instances to `prod` from `production`?

willdurand:
Sorry, so we should not change `NODE_ENV`, it must be set to `production`, but we can probably add another env variable for the config (`NODE_CONFIG_ENV`), which should be set to `prod`.

PR #6373, opened 2018-09-24
27 September homepage content update

### LINKED ISSUE #6302: 27 September homepage content update
Please update homepage content for the **27 Sept** push as detailed in the programming [doc](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#). 

@muffinresearch @bobsilverberg

bobsilverberg:
Fixes #6302 

screenshot 2018-09-24 08 30 40

### DIFF
--- a/src/amo/components/Routes/index.js
+++ b/src/amo/components/Routes/index.js
@@ -38,11 +38,7 @@ const Routes = ({ _config = config }: Props = {}) => (
     {/* TODO: Post launch update this URL and redirect see #3374/ */}
     <Route exact path="/:lang/review_guide" component={ReviewGuide} />
 
-    <Route
-      exact
-      path="/:lang/:application/"
-      component={(props) => <Home {...props} includeFeaturedThemes />}
-    />
+    <Route exact path="/:lang/:application/" component={Home} />
 
     <Route exact path="/:lang/:application/addon/:slug/" component={Addon} />
     <Route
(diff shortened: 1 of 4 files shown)


src/amo/components/Routes/index.js
```diff
@@ -38,7 +38,11 @@ const Routes = ({ _config = config }: Props = {}) => (
     {/* TODO: Post launch update this URL and redirect see #3374/ */}
     <Route exact path="/:lang/review_guide" component={ReviewGuide} />
 
-    <Route exact path="/:lang/:application/" component={Home} />
+    <Route
+      exact
+      path="/:lang/:application/"
+      component={(props) => <Home {...props} includeFeaturedThemes />}
```
willdurand: I'd rather flip the default prop, otherwise this component is going to change often, and it should not (ideally only when adding new routes).

bobsilverberg: We've done that in the past, but it just seemed wrong to me. The default is a default, and can be overridden by specified props, so that seems like the correct way to do it. We also have a test which relies on the current value of the default prop, so that test would have to change each time we change the default, which also seems like unnecessary work.

What is the concern about this component having to change when we want to flip this? Might be be better to make it a config option instead?

willdurand: It is a page-level component, so the default value configures it entirely. There is no way another prop can be injected in theory, except if you do what you've done here, which does not look good to me either because this is likely not covered by a test case.

Changing the test case makes sense to me, because the test case reflects the state of the page (given that it is a page-level component). If we want future-proof test cases, we can pass the prop with a fixed value and not rely on the default value.

bobsilverberg: Ok, I've changed it back.

PR #6378, opened 2018-09-24
Avoid unhandled promise when creating history

### LINKED ISSUE #6367: UnhandledPromiseRejectionWarning in production

I read the production logs.

### What happened?

I saw:

```
(node:201) UnhandledPromiseRejectionWarning: URIError: Pathname "/en-US/android/collections/mozilla%df%27/privacy-matters/" could not be decoded. This is likely caused by an invalid percent-encoding.
 at createLocation (/srv/node/node_modules/history/LocationUtils.js:51:13)
 at /srv/node/node_modules/history/createMemoryHistory.js:57:74
 at Array.map (<anonymous>)
 at createMemoryHistory (/srv/node/node_modules/history/createMemoryHistory.js:56:32)
 at createHistory (/srv/code/src/core/server/base.js:44:14)
 at _callee$ (/srv/code/src/core/server/base.js:281:21)
 at tryCatch (/srv/node/node_modules/regenerator-runtime/runtime.js:65:40)
 at Generator.invoke [as _invoke] (/srv/node/node_modules/regenerator-runtime/runtime.js:303:22)
 at Generator.prototype.(anonymous function) [as next] (/srv/node/node_modules/regenerator-runtime/runtime.js:117:21)
 at step (/srv/code/src/core/server/base.js:113:191)
 at /srv/code/src/core/server/base.js:113:437
 at new Promise (<anonymous>)
 at /srv/code/src/core/server/base.js:113:99
 at /srv/code/src/core/server/base.js:243:3
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at trailingSlashesMiddleware (/srv/code/src/core/middleware/trailingSlash.js:48:10)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at prefixMiddleware (/srv/code/src/core/middleware/prefixMiddleware.js:144:10)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at csp (/srv/node/node_modules/helmet-csp/index.js:59:7)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at xXssProtection (/srv/node/node_modules/x-xss-protection/index.js:26:7)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at nosniff (/srv/node/node_modules/dont-sniff-mimetype/index.js:4:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at frameguard (/srv/node/node_modules/frameguard/index.js:38:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at hsts (/srv/node/node_modules/hsts/index.js:42:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at logRequests (/srv/code/src/core/middleware/logRequests.js:7:3)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at responseTime (/srv/node/node_modules/response-time/index.js:61:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at /srv/node/node_modules/raven/lib/client.js:560:9
 at bound (domain.js:303:14)
 at runBound (domain.js:314:12)
 at Raven.context (/srv/node/node_modules/raven/lib/client.js:431:33)
 at ravenRequestMiddleware (/srv/node/node_modules/raven/lib/client.js:557:12)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at expressInit (/srv/node/node_modules/express/lib/middleware/init.js:40:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at query (/srv/node/node_modules/express/lib/middleware/query.js:45:5)
 at Layer.handle [as handle_request] (/srv/node/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/srv/node/node_modules/express/lib/router/index.js:317:13)
 at /srv/node/node_modules/express/lib/router/index.js:284:7
 at Function.process_params (/srv/node/node_modules/express/lib/router/index.js:335:12)
 at next (/srv/node/node_modules/express/lib/router/index.js:275:10)
 at Function.handle (/srv/node/node_modules/express/lib/router/index.js:174:3)
 at Function.handle (/srv/node/node_modules/express/lib/application.js:174:10)
 at Server.app (/srv/node/node_modules/express/lib/express.js:39:9)
 at emitTwo (events.js:126:13)
 at Server.emit (events.js:214:7)
 at parserOnIncoming (_http_server.js:634:12)
 at HTTPParser.parserOnHeadersComplete (_http_common.js:117:17)
 (node:201) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 5)
```

No unhandled promise in production code, ever.

I don't know whether this is hyper-critical, but it looks quite bad.

willdurand:
Fixes #6367

---

This patch makes sure there is no unhandled promise when creating react
router's history with an invalid URI. It seems to be a known issue, but
I don't know if that's the best fix. At least, it does not throw anymore.


src/core/server/base.js
```diff
@@ -39,10 +39,17 @@ import {
...
-export const createHistory = ({ req }) => {
-  return addQueryParamsToHistory({
-    history: createMemoryHistory({ initialEntries: [req.url] }),
-  });
+export const createHistory = ({ req, _log = log }) => {
+  let history;
+
+  try {
+    history = createMemoryHistory({ initialEntries: [req.url] });
+  } catch (err) {
```
kumar303: Catching and recovering from this error seems like it would leave the app in a very broken state -- the history will not have a starting URL. Really, the upstream library needs a fix. Until then I think we should let it throw and trigger a 500.

However, throwing this error does not trigger a 500 (it's an uncaught promise and thus a gateway timeout). This is the real bug; the patch here needs to fix that. All your current patch does is prevent one kind of error from getting thrown. If another one gets thrown, we will have an uncaught promise again.

We need a try/catch in [`app.use()`](https://github.com/mozilla/addons-frontend/blob/adb2ad0132798af9f5fee9e5c98c542265ba5b40/src/core/server/base.js#L243) so that it looks more like:

```js
  app.use(async (req, res, next) => {
    try {
      if (isDevelopment) {
        // ...
      }
      // Everthing else in the handler...
    } catch (handlerError) {
      log.error(`Caught an unexpected error while handling the request: ${handlerError}`);
      return next(handlerError);
    }
  });
```

After that change, I get an immediate 500 response, no uncaught promise, and I see a stack trace pointing to the error.

But, wait, why don't I see our generic 500 page? 

Here lies the second bug we need to fix. The error handler encountered an error while processing the error. Let's guard against that in [the error handler](https://github.com/mozilla/addons-frontend/blob/adb2ad0132798af9f5fee9e5c98c542265ba5b40/src/core/server/base.js#L387) like this:

```js
  app.use((error, req, res, next) => {
    try {
      // All error handler code...
      return showErrorPage({ createStore, error, status: 500, req, res, config });
    } catch (recoveryError) {
      log.error(`Additionally, the error handler caught an error: ${recoveryError}`);
      log.error({ err: recoveryError }); // log the stack trace.

      // Pass the original error to the next error handler.
      return next(error);
    }
  });
```

With this change, I see the first stack trace, the one we needed to know about all along.

willdurand: Thanks @kumar303, you're right it is better to fix the root cause than patching the visible problem. My mind was not fresh enough, I'll revisit this patch by tomorrow.

willdurand: > Catching and recovering from this error seems like it would leave the app in a very broken state -- the history will not have a starting URL. Really, the upstream library needs a fix. Until then I think we should let it throw and trigger a 500.

The history had `/`, which is enough to show a nice error page. Any navigation will trigger a `LOCATION_CHANGE`, so the state is not super broken, only the location is not correct but that's because the URL is not valid. That's the main issue and not sure the upstream lib can do anything about it.

willdurand: > But, wait, why don't I see our generic 500 page?
> 
> Here lies the second bug we need to fix. The error handler encountered an error while processing the error. Let's guard against that in [the error handler](https://github.com/mozilla/addons-frontend/blob/adb2ad0132798af9f5fee9e5c98c542265ba5b40/src/core/server/base.js#L387) like this:

I still don't see any change, the nice error page is not displayed because we need the store, which needs a history. I wish we had static error pages, at least for 5xx.

kumar303: >> Here lies the second bug we need to fix. The error handler encountered an error while processing the error.

> I still don't see any change

The stack traces are different (but they are hard to read because of the async generator implementation).

Without protecting the error handler, the stack trace points to the error handler as the culprit:

https://github.com/mozilla/addons-frontend/blob/eaf095d9631c21ac98533fc56f00d981373d0cc2/src/core/server/base.js#L382

...and then `createHistory()`:

https://github.com/mozilla/addons-frontend/blob/eaf095d9631c21ac98533fc56f00d981373d0cc2/src/core/server/base.js#L101

But the real culprit was in the request handler:

https://github.com/mozilla/addons-frontend/blob/eaf095d9631c21ac98533fc56f00d981373d0cc2/src/core/server/base.js#L267

The code example I provided above puts a focus on the real stack trace but logs the side effect stack trace, too.

kumar303:
The new changes look good at a glance. Send me a re-review request when the conflicts are resolved and I'll try it out.

willdurand:
> The new changes look good at a glance. Send me a re-review request when the conflicts are resolved and I'll try it out.

@kumar303 done, thanks!

tests/unit/helpers.js
```diff
@@ -732,3 +732,13 @@ export async function matchingSagaAction(
   }
   return foundAction;
 }
+
+export const getFakeLogger = (params = {}) => {
```
kumar303: Good idea. I'm surprised we didn't have one. Could you file an issue to port all existing test code over to this? Maybe it would be a *good first bug*. I see about 8 calls that look like `fakeLog = { ... }`.

tests/unit/core/server/test_server.js
```diff
@@ -394,6 +398,35 @@ describe(__filename, () => {
       expect(response.status).toEqual(301);
       expect(response.headers.location).toEqual(newURL);
     });
+
+    it('catches all errors and returns a 500', async () => {
```
kumar303: Nice test 👍

tests/unit/helpers.js
willdurand: Yep, this bothered me several times.

PR #6383, opened 2018-09-24
Make review link a Link rather than an <a>

### LINKED ISSUE #6289: Clicking a link to a review triggers a page refresh

1. Go to the page listing the reviews of an add-on
2. Click the relative time of one review (it has a link)

### What happened?

A page refresh.

Client-side navigation, no page refresh.

The problem comes from these lines: https://github.com/mozilla/addons-frontend/blob/3547c977acd733874dd2b908ae171c2091e4f496/src/amo/components/AddonReviewCard/index.js#L303-L309

We use the `a` HTML element and not the `Link` React component, by-passing the client-side navigation. We should not be doing this but I am not sure to know how to fix this.

NB: there is a similar problem with the link to a user profile from an add-on detail page.

bobsilverberg:
Fixes #6289 

Thanks for the suggestion in #6289, @kumar303. It seems to have worked well.

### DIFF
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -302,6 +302,10 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
         : // translators: Example in English: "by UserName123, last week"
           i18n.gettext('by %(authorName)s, %(timestamp)s');
 
+      // This somewhat odd code is needed because we want to wrap the timestamp
+      // inside a Link component, and the whole thing is inside a string that
+      // needs to be localized, so we localize the string first and then
+      // inject the timestamp, wrapped in a Link, inside the string.
       const localized = i18n.sprintf(byLineString, {
         authorName: review.userName,
         timestamp: '__timestamp__',
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -6,6 +6,7 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
+import Link from 'amo/components/Link';
 import AddonReview from 'amo/components/AddonReview';
 import AddonReviewManager from 'amo/components/AddonReviewManager';
 import FlagReviewMenu from 'amo/components/FlagReviewMenu';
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonReviewCard/index.js
```diff
@@ -294,33 +291,33 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
           i18n.gettext('posted %(timestamp)s')
         : // translators: Example in English: "by UserName123, last week"
           i18n.gettext('by %(authorName)s, %(timestamp)s');
 
+      const localized = i18n.sprintf(byLineString, {
+        authorName: review.userName,
+        timestamp: '__timestamp__',
+      });
+
+      const parts = localized.split('__timestamp__');
```
kumar303: Since this code is a bit convoluted, it would be nice to explain why it's needed in a comment.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -294,33 +291,33 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
+      const allParts = [parts.shift(), timestamp, ...parts];
```
kumar303: Rendering the remaining `...parts` isn't covered by any tests. It would be helpful to add coverage for it. You could do `sinon.stub(i18n, 'gettext')` to inject any kind of L10n string.

kumar303: Similar to my last comment, it seems that `parts.shift()` doesn't have test coverage either.

willdurand: How does that work with RTL?

bobsilverberg: That's a good question. I'm guessing that any localizations done in RTL languages will already be in the correct order, so I think this will "just work".

bobsilverberg:
Thanks for the review @kumar303. I added a comment, and also a test. I'm not sure if the test is perfect. I originally was asserting that the text was exactly 
```
`${firstPart} <Connect(LinkBase) /> ${lastPart}` 
```

in order to prove that the `Link` was inserted in between the parts, but I didn't really like that assertion.

src/amo/components/AddonReviewCard/index.js
kumar303: Hmm, we should test this. I think it will probably work, though.

kumar303: I just meant to manually test it. I did so by comparing the Hebrew and Arabic views to the same review listing on -dev. It matches -dev so I think it's working fine.

PR #6386, opened 2018-09-24
Enable new install button on AMO (all envs, including -prod)

### LINKED ISSUE #6379: Enable new install button on AMO (all envs, including -prod)
It's time.

willdurand:
Fix #6379


rebmullin:
@willdurand, can you remind me why we are not enabling for disco here? Or will you be opening another issue for that?

willdurand:
@rebmullin the disco pane is running a shield study, so we cannot enable the buttons here yet

Issue #6172, opened 2018-09-07
Do not duplicate the webpack config in storybook


willdurand:
The `stories/setup/webpack.config.babel.js` duplicates a lot of webpack config that is already present in `webpack.dev.config.babel.js`.

We should avoid that and require the `webpack.dev.config.babel.js` file in  `stories/setup/webpack.config.babel.js`. From there, we can reuse parts of the config we are interested in. Here is a working patch:

```diff
diff --git a/stories/setup/webpack.config.babel.js b/stories/setup/webpack.config.babel.js
index 8633795ab..d52f2b18e 100644
--- a/stories/setup/webpack.config.babel.js
+++ b/stories/setup/webpack.config.babel.js
@@ -1,58 +1,9 @@
-import 'core/polyfill';
-
-import path from 'path';
-
-import config from 'config';
-import autoprefixer from 'autoprefixer';
-import webpack from 'webpack';
-
-import { getClientConfig } from 'core/utils';
+import webpackDevConfig from '../../webpack.dev.config.babel';
 
 module.exports = {
   module: {
-    rules: [
-      {
-        test: /\.svg$/,
-        use: [{ loader: 'svg-url-loader', options: { limit: 10000 } }],
-      },
-      {
-        test: /\.scss$/,
-        use: [
-          { loader: 'style-loader' },
-          { loader: 'css-loader', options: { importLoaders: 2 } },
-          {
-            loader: 'postcss-loader',
-            options: {
-              plugins: [
-                autoprefixer({
-                  browsers: ['last 2 versions'],
-                  grid: false,
-                }),
-              ],
-            },
-          },
-          { loader: 'sass-loader', options: { outputStyle: 'expanded' } },
-        ],
-      },
-    ],
-  },
-  // The following plugins are needed to help handle
-  // server-side imported components such as core/logger.
-  // Is this the best way??
-  plugins: [
-    new webpack.DefinePlugin({
-      CLIENT_CONFIG: JSON.stringify(getClientConfig(config)),
-    }),
-
-    new webpack.NormalModuleReplacementPlugin(
-      /config$/,
-      'core/client/config.js',
-    ),
-  ],
-  resolve: {
-    alias: {
-      normalize: 'normalize.css/normalize.css',
-    },
-    modules: [path.resolve('./src'), 'node_modules'],
+    rules: webpackDevConfig.module.rules,
   },
+  plugins: webpackDevConfig.plugins,
+  resolve: webpackDevConfig.resolve,
[...]

PR #6389, opened 2018-09-24
I6387 storybook setup followup

### LINKED ISSUE #6387: Storybook set up - phase 2
I tested pulling in some other components and we still need:

Provider wrapper:
Alt text

wepback: png, etc loader, window on server issue

rebmullin:
fixes #6387

### DIFF
--- a/stories/.eslintrc
+++ b/stories/.eslintrc
@@ -3,6 +3,15 @@
     "import/no-extraneous-dependencies": ["error", {
       // Allow dev-dependencies in this directory.
       "devDependencies": true
-    }]
+    }],
+    "import/no-unresolved": [
+      "error", {
+        // TODO: Look for a better way.
+        // This was the only I could disable the
+        // tests resolve linting issue for now (without
+        // adding disable eslint [this rule] to the file itself).
+        "ignore": [ "tests" ]
+      }
+    ]
   }
 }
--- a/stories/.eslintrc
+++ b/stories/.eslintrc
@@ -7,7 +7,7 @@
     "import/no-unresolved": [
       "error", {
         // TODO: Look for a better way.
-        // This was the only I could disable the
+        // This was the only way I could disable the
         // tests resolve linting issue for now (without
         // adding disable eslint [this rule] to the file itself).
         "ignore": [ "tests" ]
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -1,8 +1,7 @@
 /* @flow */
 import { createBrowserHistory } from 'history';
 import React from 'react';
-import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
-import { fakeI18n } from 'tests/unit/helpers';
+import { dispatchClientMetadata, fakeI18n } from 'tests/unit/helpers';
 
 import Root from 'core/components/Root';
 import { addQueryParamsToHistory } from 'core/utils';
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -3,18 +3,15 @@ import { createBrowserHistory } from 'history';
 import React from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
 
-// Without core/polyfill we see the following error:
-// "regeneratorRuntime is not defined".
-import 'core/polyfill';
 import createStore from 'amo/store';
 import { addQueryParamsToHistory } from 'core/utils';
 
 const history = addQueryParamsToHistory({
-  history: createBrowserHistory(),
+	history: createBrowserHistory(),
 });
 
 const { store } = createStore({ history });
 
 export default function Provider({ story }: Object) {
-  return <ReduxProvider store={store}>{story}</ReduxProvider>;
+	return <ReduxProvider store={store}>{story}</ReduxProvider>;
 }
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -1,0 +1,20 @@
+/* @flow */
+import { createBrowserHistory } from 'history';
+import React from 'react';
+import { Provider as ReduxProvider } from 'react-redux';
+
+// Without core/polyfill we see the following error:
+// "regeneratorRuntime is not defined".
+import 'core/polyfill';
+import createStore from 'amo/store';
+import { addQueryParamsToHistory } from 'core/utils';
+
+const history = addQueryParamsToHistory({
+  history: createBrowserHistory(),
+});
+
+const { store } = createStore({ history });
+
+export default function Provider({ story }: Object) {
+  return <ReduxProvider store={store}>{story}</ReduxProvider>;
+}
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -1,16 +1,23 @@
 /* @flow */
 import { createBrowserHistory } from 'history';
 import React from 'react';
-import { Provider as ReduxProvider } from 'react-redux';
 
-// Without core/polyfill we see the following error:
-// "regeneratorRuntime is not defined".
-import 'core/polyfill';
+import Root from 'core/components/Root';
+import { addQueryParamsToHistory } from 'core/utils';
 
 import { dispatchClientMetadata } from '../../tests/unit/amo/helpers';
+import { fakeI18n } from '../../tests/unit/helpers';
 
 const { store } = dispatchClientMetadata();
 
+const history = addQueryParamsToHistory({
+  history: createBrowserHistory(),
+});
+
 export default function Provider({ story }: Object) {
-	return <ReduxProvider store={store}>{story}</ReduxProvider>;
+  return (
+    <Root store={store} history={history} i18n={fakeI18n}>
+      {story}
+    </Root>
+  );
 }
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -3,15 +3,14 @@ import { createBrowserHistory } from 'history';
 import React from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
 
-import createStore from 'amo/store';
-import { addQueryParamsToHistory } from 'core/utils';
+// Without core/polyfill we see the following error:
+// "regeneratorRuntime is not defined".
+import 'core/polyfill';
 
-const history = addQueryParamsToHistory({
-  history: createBrowserHistory(),
-});
+import { dispatchClientMetadata } from '../../tests/unit/amo/helpers';
 
-const { store } = createStore({ history });
+const { store } = dispatchClientMetadata();
 
 export default function Provider({ story }: Object) {
-  return <ReduxProvider store={store}>{story}</ReduxProvider>;
+	return <ReduxProvider store={store}>{story}</ReduxProvider>;
 }
--- a/stories/setup/Provider.js
+++ b/stories/setup/Provider.js
@@ -1,12 +1,11 @@
 /* @flow */
 import { createBrowserHistory } from 'history';
 import React from 'react';
+import { dispatchClientMetadata } from 'tests/unit/amo/helpers';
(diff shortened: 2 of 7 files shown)


stories/setup/Provider.js
```diff
@@ -0,0 +1,20 @@
+/* @flow */
+import { createBrowserHistory } from 'history';
+import React from 'react';
+import { Provider as ReduxProvider } from 'react-redux';
+
+// Without core/polyfill we see the following error:
+// "regeneratorRuntime is not defined".
+import 'core/polyfill';
```
kumar303: When I remove this, I don't see any errors. What are the steps to reproduce?

In any case, it's already imported in the webpack config so I'm not sure why it's needed twice. https://github.com/mozilla/addons-frontend/blob/5ad0920f1855cb28f51c6cf426791cd984b43f39/stories/setup/webpack.config.babel.js#L1

stories/setup/Provider.js
```diff
@@ -0,0 +1,20 @@
...
+import createStore from 'amo/store';
+import { addQueryParamsToHistory } from 'core/utils';
+
+const history = addQueryParamsToHistory({
+  history: createBrowserHistory(),
+});
+
+const { store } = createStore({ history });
+
+export default function Provider({ story }: Object) {
```
kumar303: I don't see this used anywhere. Maybe you need it for a forthcoming patch? If that's the case, I'd rather see this all applied to the forthcoming patch.

stories/setup/webpack.config.babel.js
```diff
@@ -48,6 +52,12 @@ module.exports = {
       /config$/,
       'core/client/config.js',
     ),
+
+    // This swaps the server side window object with a standard browser window.
+    new webpack.NormalModuleReplacementPlugin(
+      /core\/window/,
+      'core/browserWindow.js',
+    ),
```
willdurand: Why exactly don't we reuse the existing production config?

stories/setup/Provider.js
```diff
@@ -0,0 +1,17 @@
...
+  return <ReduxProvider store={store}>{story}</ReduxProvider>;
+}
```
willdurand: Why do we need this? Can't we pass a `store` like we do in the test suite?

stories/setup/webpack.config.babel.js
rebmullin: I think we did this at one pt but I thought we were only going to pass what we needed. I can revisit this tho..

stories/setup/Provider.js
rebmullin: actually I had to add this back with the changes I made. I removed this from the webpack tho which I didn't see had any impact.

stories/setup/Provider.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+import { createBrowserHistory } from 'history';
+import React from 'react';
+import { Provider as ReduxProvider } from 'react-redux';
+
+// Without core/polyfill we see the following error:
+// "regeneratorRuntime is not defined".
+import 'core/polyfill';
```
kumar303: Would it be possible to put this in a more central entry point? I worry that `setup/Provider` might be too late in the bundling process. What if a module that comes before it needs the polyfill? Putting it in `setup/webpack.config.babel.js` seemed like a central entry point. Does that no longer work?

stories/setup/webpack.config.babel.js
```diff
@@ -48,10 +50,17 @@ module.exports = {
       /config$/,
       'core/client/config.js',
     ),
+
+    // This swaps the server side window object with a standard browser window.
+    new webpack.NormalModuleReplacementPlugin(
```
kumar303: It looks like this whole plugin array can be built with [`getPlugins()`](https://github.com/mozilla/addons-frontend/blob/master/webpack-common.js#L140) -- did you try that? If it doesn't work, maybe you can add some parameters to that function for how you need it to change?

stories/setup/Provider.js
```diff
@@ -0,0 +1,16 @@
...
+
+import { dispatchClientMetadata } from '../../tests/unit/amo/helpers';
```
kumar303: I see an alias in the webpack config for `tests`. I think you can use that for this import path, right?

stories/setup/Provider.js
```diff
@@ -0,0 +1,16 @@
...
+
+const { store } = dispatchClientMetadata();
+
+export default function Provider({ story }: Object) {
+	return <ReduxProvider store={store}>{story}</ReduxProvider>;
```
willdurand: We have a `Root` component that could be useful, maybe: https://github.com/mozilla/addons-frontend/blob/master/src/core/components/Root/index.js

stories/setup/Provider.js
rebmullin: It didn't work :/ but let me trying moving it maybe to config or somewhere higher in the chain if I can..

stories/setup/webpack.config.babel.js
rebmullin: I think I had that originally so let me revisit this one : )

stories/setup/Provider.js
rebmullin: sigh webpack... if I adjust the resolve info a bit yes I think it will work.

stories/setup/Provider.js
rebmullin: oh! I'll check it out, thanks @willdurand  :)

stories/setup/Provider.js
rebmullin: I seem to be having issues with this too :/  It works but then lint and flow complain about not finding the module. If I don't put the resolve at all, then this then component complains about it 😕

stories/setup/webpack.config.babel.js
rebmullin: the circular plugin causes issue it seems so if I add a check for that I think this will work

stories/setup/webpack.config.babel.js
```diff
@@ -1,57 +1,16 @@
...
-    new webpack.NormalModuleReplacementPlugin(
-      /config$/,
-      'core/client/config.js',
-    ),
-  ],
+  plugins: [...getPlugins({ includeCircularDependencyPlugin: false })],
   resolve: {
     alias: {
       normalize: 'normalize.css/normalize.css',
+      tests: path.resolve('./tests'),
```
rebmullin: This is not working as expected. If I leave this out, the component can't find the module. If I add it I still need to do ../../test etc..

stories/setup/webpack.config.babel.js
```diff
@@ -36,22 +38,11 @@ module.exports = {
...
```
kumar303: I'd like to see some duplication removed from this webpack config if possible, otherwise it will be a maintenance burden. We need to share as much common config as we can.

I made this change and everything seems to work. Could you apply this patch to share the common test alias?

 I tried using `getRules()` but webpack complains about schema errors which I don't understand. Duplicating the rules are OK for now but I'd like to see that fixed later on if possible.

```diff
diff --git a/stories/setup/webpack.config.babel.js b/stories/setup/webpack.config.babel.js
index d4f3f648c..4495c9847 100644
--- a/stories/setup/webpack.config.babel.js
+++ b/stories/setup/webpack.config.babel.js
@@ -4,9 +4,11 @@ import path from 'path';
 
 import autoprefixer from 'autoprefixer';
 
+import webpackConfig from '../../webpack.prod.config.babel';
 import { getPlugins } from '../../webpack-common';
 
 module.exports = {
+  ...webpackConfig,
   module: {
     rules: [
       {
@@ -39,11 +41,4 @@ module.exports = {
     ],
   },
   plugins: [...getPlugins({ includeCircularDependencyPlugin: false })],
-  resolve: {
-    alias: {
-      normalize: 'normalize.css/normalize.css',
-      tests: path.resolve('./tests'),
-    },
-    modules: [path.resolve('./src'), 'node_modules'],
-  },
 };

```

stories/setup/webpack.config.babel.js
```diff
@@ -36,22 +38,11 @@ module.exports = {
...
-    new webpack.DefinePlugin({
-      CLIENT_CONFIG: JSON.stringify(getClientConfig(config)),
-    }),
-
-    new webpack.NormalModuleReplacementPlugin(
-      /config$/,
-      'core/client/config.js',
-    ),
-  ],
+  plugins: [...getPlugins({ includeCircularDependencyPlugin: false })],
```
kumar303: Well, huh. That exposed a real bug. Please add a todo comment to re-enable this plugin (it's helpful!) after https://github.com/mozilla/addons-frontend/issues/6561 is fixed.

stories/setup/webpack.config.babel.js
rebmullin: I've cleaned this up as suggested. I had to use the prod config as you have here. Dev doesn't work b/c of certain components (like log, etc). 

Regarding the getRules...if we don't want to show prop types, then we can use this. I kind of find seeing the props types helpful so I looked into this a bit more and noticed if we did not include the babel-loader this works. So I basically have this set up as a config now for the function.

stories/setup/Provider.js
```diff
@@ -0,0 +1,23 @@
+/* @flow */
+import { createBrowserHistory } from 'history';
+import React from 'react';
+
+import Root from 'core/components/Root';
+import { addQueryParamsToHistory } from 'core/utils';
+
+import { dispatchClientMetadata } from '../../tests/unit/amo/helpers';
```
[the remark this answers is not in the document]
rebmullin: thanks for this : )

stories/setup/webpack.config.babel.js
rebmullin: Actually maybe I'll do a separate PR for this as part of this [issue](https://github.com/mozilla/addons-frontend/issues/6172)

rebmullin: I did extract this out in a new PR: #6571

stories/setup/webpack.config.babel.js
```diff
@@ -1,58 +1,18 @@
...
-
-import path from 'path';
-
-import config from 'config';
-import autoprefixer from 'autoprefixer';
-import webpack from 'webpack';
-
-import { getClientConfig } from 'core/utils';
+import { getPlugins, getRules } from '../../webpack-common';
+import webpackConfig from '../../webpack.prod.config.babel';
```
rebmullin: note: I did extract out these changes into a new PR #6571

rebmullin:
hey @kumar303, can you please take another look? 

notes: 
I've updated the webpack config in PR #6571. 

Unfortunately, I am still having an issue with lint (for the test path):/  I've tried playing with eslint import resolver but that did not seem to work. For now, I turned this rule off in eslintrc since the module and now flow are ok with this. Please let me know if I missing something or if what I have will be ok for now.

stories/.eslintrc
```diff
@@ -3,6 +3,15 @@
...
       "devDependencies": true
-    }]
+    }],
+    "import/no-unresolved": [
+      "error", {
+        // TODO: Look for a better way.
+        // This was the only way I could disable the
+        // tests resolve linting issue for now (without
+        // adding disable eslint [this rule] to the file itself).
+        "ignore": [ "tests" ]
```
kumar303: I suppose that's fine for now. We should probably use the [`webpack`](https://www.npmjs.com/package/eslint-import-resolver-webpack) resolver but I couldn't get it to work.

stories/setup/config.js
```diff
@@ -5,6 +5,10 @@ import infoAddon, {
   setDefaults as setAddonInfoDefaults,
 } from '@storybook/addon-info';
 
+// Without core/polyfill we see the following error:
```
kumar303: You don't really need this comment. The polyfill is required for lots of things to work.

PR #6390, opened 2018-09-24
I6310 adding flow to btn component

### LINKED ISSUE #6310: Add flow types to Button component

rebmullin:
fixes #6310

### DIFF
--- a/src/ui/components/Button/index.js
+++ b/src/ui/components/Button/index.js
@@ -1,12 +1,30 @@
+/* @flow */
 import makeClassName from 'classnames';
 import { oneLine } from 'common-tags';
 import * as React from 'react';
-import PropTypes from 'prop-types';
 
 import Link from 'amo/components/Link';
 import log from 'core/logger';
 
 import './styles.scss';
+
+type Props = {|
+  buttonType: string | 'none',
+  children?: React.Node,
+  className?: string,
+  disabled?: boolean | null,
+  externalDark?: boolean,
+  href?: string,
+  micro?: boolean,
+  name?: number,
+  noLink?: boolean,
+  onClick?: Function | null,
+  puffy?: boolean,
+  to?: string | Object,
+  target?: string,
+  type?: string,
+  'data-browsertheme'?: string,
+|};
 
 const BUTTON_TYPES = [
   'neutral',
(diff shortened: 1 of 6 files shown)


src/ui/components/Button/index.js
```diff
@@ -51,7 +58,9 @@ export default class Button extends React.Component {
       to,
       ...rest
     } = this.props;
-    const props = { ...rest };
+    const props: Object = { ...rest };
```
rebmullin: I am not sure if Object is ok here or if I should create up something like "internalProps". Initially I went that route but it wasn't working so with buttonType. if that's preferred tho, I will look into that more..

src/ui/components/Button/index.js
```diff
@@ -1,13 +1,31 @@
...
 import * as React from 'react';
-import PropTypes from 'prop-types';
 
 import Link from 'amo/components/Link';
 import log from 'core/logger';
 
 import './styles.scss';
 
+type Props = {|
+  buttonType: string | 'none',
```
willdurand: We should use a better type that is the intersection of all possible values of `BUTTON_TYPES`. This must be done manually though...

```
type ButtonType =
  | 'neutual'
  | 'action'
  | ...;
```

src/ui/components/Button/index.js
willdurand: We can use `$Rest` I think.

src/ui/components/Button/index.js
willdurand: Yes, it means there are other components to fix. See the diff I pasted above, that should help you (hopefully)!

rebmullin: oh I see in your recommended patch how you we are importing this to each file and fixing this 👍

src/ui/components/Button/index.js
```diff
@@ -1,13 +1,40 @@
...
+  | 'cancel'
+  | 'confirm'
+  | 'alert'
+  | 'none';
+
+type Props = {|
+  buttonType: ButtonType,
+  children?: React.Node,
+  className?: string,
+  disabled: boolean | null,
```
willdurand: I think we should not have that, it should be `boolean` only

src/ui/components/Button/index.js
```diff
@@ -1,13 +1,40 @@
...
+  externalDark?: boolean,
+  href?: string,
+  micro: boolean,
+  name?: number,
+  noLink: boolean,
+  onClick?: Function | null,
+  puffy: boolean,
+  to?: string | Object,
```
willdurand: Can you add a `// TODO: make a better Object type` here?

src/ui/components/Button/index.js
willdurand: It requires a change in the `UserProfile` (or `UserProfileEdit`) component.

rebmullin: ah! I missed this above. I'll update..

PR #6392, opened 2018-09-25
Fixes #6275: Redirecting /user/<userId> to /user/<username>

### LINKED ISSUE #6275: Redirect user profile URLs using user id to the canonical variant using username
- Visit https://addons.mozilla.org/en-US/firefox/user/4757633/

### What happened?
- The page loads directly

- The client should be redirect to the canonical URL using the username instead of the user id, which is https://addons.mozilla.org/en-US/firefox/user/mozilla/

- We do that for add-ons already: https://addons.mozilla.org/en-US/firefox/addon/954390/ redirects to https://addons.mozilla.org/en-US/firefox/addon/facebook-container/

suhailsinghbains:
Fixes #6275
On visiting https://addons.mozilla.org/en-US/firefox/user/4757633/ , the page is redirected to https://addons.mozilla.org/en-US/firefox/user/mozilla/

No pictures could be attached, as a Video demonstrating would be appropriate.
Thanks @willdurand for being my mentor.

### DIFF
--- a/src/amo/pages/UserProfile/index.js
+++ b/src/amo/pages/UserProfile/index.js
@@ -67,11 +67,13 @@ type Props = {|
 type InternalProps = {|
   ...Props,
   canEditProfile: boolean,
+  clientApp: string,
   currentUser: UserType | null,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
   isOwner: boolean,
+  lang: string,
   pageSize: number | null,
   reviewCount: number | null,
   reviews: Array<UserReviewType> | null,
@@ -83,9 +85,11 @@ export class UserProfileBase extends React.Component<InternalProps> {
     super(props);
 
     const {
+      clientApp,
       dispatch,
       errorHandler,
       isOwner,
+      lang,
       location,
       match: { params },
       reviews,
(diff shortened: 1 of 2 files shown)


src/amo/pages/UserProfile/index.js
```diff
@@ -112,6 +113,16 @@ export class UserProfileBase extends React.Component<InternalProps> {
         }),
       );
     }
+    let idOrNameCheck = { ...location}.pathname.split('/')[4];
```
willdurand: hu oh, what's that? Why not using `location` directly? In any case I don't think you need the `location`, there should be a `match` prop that does that (i.e. extracting URL parameters).

src/amo/pages/UserProfile/index.js
```diff
@@ -112,6 +113,16 @@ export class UserProfileBase extends React.Component<InternalProps> {
...
+    if(/^[0-9]+$/.test(idOrNameCheck)){
```
willdurand: I think you can use `\d+`

src/amo/pages/UserProfile/index.js
```diff
@@ -112,6 +113,16 @@ export class UserProfileBase extends React.Component<InternalProps> {
...
+      const newUrl= { ...user };
```
willdurand: You can remove that and use `user.username` in the `url:` value.

suhailsinghbains: I'm getting this error and hence, does not redirect. I just cannot use any property of 'user'.
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'username' of undefined

willdurand: ah, I see. Yes, because we need to fetch the user first. So we need the user to send the server redirect.

willdurand:
@suhailsinghbains hi, do you need any help?

suhailsinghbains:
> 
> 
> @suhailsinghbains hi, do you need any help?

Yes please, I tried yarn test, every test passed, but, here travis is giving error.

suhailsinghbains:
The yarn amo:dev project works perfectly(redirection)
But, I couldn't figure out how to satisfy the test. I had been trying for past 3 hours. Please help me @willdurand

willdurand:
>  But, I couldn't figure out how to satisfy the test. I had been trying for past 3 hours. Please help me @willdurand

That's because we have to update some other tests, because the current logic uses the `username` as default parameter. This requires some changes and I filed #6526 for that (I'm going to work on it first, then you'll be able to rebase this PR on top of my patch for #6526)

suhailsinghbains:
Ohh, Thank God, I was thinking that I was disturbing you for such small things :)

muffinresearch:
@suhailsinghbains hi - have you had a chance to take another look at this?

suhailsinghbains:
@muffinresearch Hi, I got busy with the exams, so, didn't get time to go through again.
>That's because we have to update some other tests, because the current logic uses the username as default parameter. This requires some changes and I filed #6526 for that (I'm going to work on it first, then you'll be able to rebase this PR on top of my patch for #6526)

This issue(#6526 ) is open though, 
Still, I'll have a look :)

willdurand:
@suhailsinghbains thanks a bunch for your work here. Unfortunately, things have changed a bit and also a lot of other things happened in the meantime, so I am not able to accept this PR anymore (or to help you iterate on it).

You can find other issues by looking at the [opened `contrib: welcome` issues](https://github.com/mozilla/addons-frontend/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3A%22contrib%3A+welcome%22). Those issues should be safer (we won't reject your PRs) 😛

PR #6400, opened 2018-09-25
Simplify the 'rating saved' notification

### LINKED ISSUE #6345: Simplify the 'rating saved' notification

- Sign in
- Go to an add-on you haven't rated
- Select a star to rate the add-on

### What happened?

You see a green notification banner

You should see a text notification with a green check mark (no banner)

<details>
<summary>Example</summary>

ratings and reviews flow v2

</details>

bobsilverberg:
Fixes #6345 

New "saving" message:

screenshot 2018-09-25 10 13 19

New "saved" message:

screenshot 2018-09-25 10 12 08

### DIFF
--- a/src/amo/components/RatingManagerNotice/index.js
+++ b/src/amo/components/RatingManagerNotice/index.js
@@ -28,7 +28,7 @@ const RatingManagerNotice = ({
 
   if (type) {
     return (
-      <Notice type={type} {...props}>
+      <Notice type={type} light {...props}>
         {message}
       </Notice>
     );
--- a/src/amo/components/RatingManagerNotice/styles.scss
+++ b/src/amo/components/RatingManagerNotice/styles.scss
@@ -1,10 +1,5 @@
 @import '~amo/css/styles';
 
-.RatingManager .AddonReviewCard-allControls {
-  padding-bottom: 3px;
-  padding-top: 2px;
-}
-
 .RatingManager .RatingManagerNotice-savedRating-hidden {
   display: none;
 }
--- a/src/amo/components/RatingManagerNotice/styles.scss
+++ b/src/amo/components/RatingManagerNotice/styles.scss
@@ -5,6 +5,6 @@
   text-align: center;
 }
 
-.RatingManagerNotice-savedRating-hidden {
+.RatingManager .RatingManagerNotice-savedRating-hidden {
   display: none;
 }
--- a/src/amo/components/RatingManagerNotice/styles.scss
+++ b/src/amo/components/RatingManagerNotice/styles.scss
@@ -1,8 +1,8 @@
 @import '~amo/css/styles';
 
-.RatingManagerNotice-savedRating {
-  margin-bottom: 12px;
-  text-align: center;
+.RatingManager .AddonReviewCard-allControls {
+  padding-bottom: 3px;
+  padding-top: 2px;
 }
 
 .RatingManager .RatingManagerNotice-savedRating-hidden {
--- a/src/ui/components/Notice/styles.scss
+++ b/src/ui/components/Notice/styles.scss
@@ -14,6 +14,7 @@
     background-color: transparent;
     display: flex;
     justify-content: center;
+    padding: 0;
   }
 
   &.Notice-dismissible {
--- a/src/ui/components/Notice/styles.scss
+++ b/src/ui/components/Notice/styles.scss
@@ -14,6 +14,7 @@
     background-color: transparent;
     display: flex;
     justify-content: center;
+    line-height: $default-line-height;
     padding: 0;
   }
 
@@ -45,6 +46,14 @@
   width: 16px;
 }
 
+.Notice-light .Notice-icon {
+  // Reduce the icon size for light notices so that the total
+  // height of the notice adds up to $default-line-height.
+  background-size: 11px 11px;
+  height: 11px;
+  width: 11px;
+}
+
 .Notice-dismisser-button {
   cursor: pointer;
   padding: 0;
@@ -54,6 +63,10 @@
   display: inline-block;
   margin: 0;
   padding: 4px 0;
+
+  .Notice-light & {
+    padding: 0;
+  }
 
   // The spec calls for additional spacing
   // in front of the action button. This is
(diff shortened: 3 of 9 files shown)


tests/unit/ui/components/TestRatingManagerMessage.js
```diff
@@ -0,0 +1,45 @@
+import { shallow } from 'enzyme';
```
kumar303: This test file needs to move to ` tests/unit/amo/components/TestRatingManagerMessage.js`. I missed that in the original patch.

tests/unit/amo/components/TestRatingManager.js
```diff
@@ -398,10 +397,11 @@ describe(__filename, () => {
...
     const root = render({ store });
 
-    const message = root.find(RatingManagerNotice);
-    expect(message).toHaveProp('message', 'Star rating saved');
-    expect(message).toHaveProp('type', successType);
-    expect(message).toHaveProp('hideMessage', false);
+    const notice = root.find(RatingManagerMessage);
+    expect(notice).toHaveProp('hideMessage', false);
+    expect(shallow(notice.prop('message'))).toHaveText(
+      '<Icon /> Star rating saved',
```
kumar303: This seems fragile to me since it relies on Enzyme's debug output (showing the nested components in the shallow wrapper). I guess it's probably fine but really the test seems like it's checking the text. You could do that directly:

```diff
diff --git a/tests/unit/amo/components/TestRatingManager.js b/tests/unit/amo/components/TestRatingManager.js
index 6fcd1f4ba..4f56c5f9b 100644
--- a/tests/unit/amo/components/TestRatingManager.js
+++ b/tests/unit/amo/components/TestRatingManager.js
@@ -399,8 +399,8 @@ describe(__filename, () => {
 
     const notice = root.find(RatingManagerMessage);
     expect(notice).toHaveProp('hideMessage', false);
-    expect(shallow(notice.prop('message'))).toHaveText(
-      '<Icon /> Star rating saved',
+    expect(shallow(notice.prop('message')).text()).toContain(
+      'Star rating saved',
     );
   });
 
```

You could then add an `expect(wrapper.find(Icon)).toHaveLength(1)` but I'm not sure that it's necessary.

kumar303:
> Why not use the Notice component and add either a new type or a prop to change its UI

I like that idea! I hadn't thought of it. This should be an easy change in the Notice stylesheet.

bobsilverberg:
Here's the new version which uses a new "light" notice.

screenshot 2018-09-25 14 13 04

screenshot 2018-09-25 14 12 40

Note that I left the icon visible for the generic "saving" message as well as displaying the icon for the success "saved" message because it seems like a nice thing to have, but I can remove it if we'd rather only have a check mark for "saved" and no icon for "saving".

src/ui/components/Notice/styles.scss
```diff
@@ -10,6 +10,12 @@
   line-height: 1.1;
   padding: 4px;
 
+  &.Notice-light {
```
kumar303: I think you should also remove the padding for `light` notices. Since it doesn't have a colorful container, I think the padding is not helpful. In this case, it expands the margin which looks inconsistent with all other margins:

screenshot 2018-09-25 14 18 40

src/amo/components/RatingManagerNotice/index.js
```diff
@@ -28,7 +28,7 @@ const RatingManagerNotice = ({
 
   if (type) {
     return (
-      <Notice type={type} {...props}>
+      <Notice type={type} light {...props}>
```
kumar303: I think it would be a nice feature if *Delete rating* gets hidden when the *Saved* message is visible. This would make the screen less jumpy.

So, instead of:

screenshot 2018-09-25 14 18 40


...make it look like:

screenshot 2018-09-25 14 27 32

I guess it would require a new property on `AddonReviewCard` like `hideControls` or something.

What do you think?

bobsilverberg: Sounds like a good idea. I've implemented this. I had to monkey around with some padding to remove any jumpiness, and it looks fine when I run it locally now.

src/amo/components/RatingManagerNotice/styles.scss
```diff
@@ -1,10 +1,10 @@
 @import '~amo/css/styles';
 
-.RatingManagerNotice-savedRating {
-  margin-bottom: 12px;
-  text-align: center;
+.RatingManager .AddonReviewCard-allControls {
+  padding-bottom: 3px;
+  padding-top: 2px;
```
kumar303: This feels like a fragile way to keep the two blocks in sync.

I think the key to fixing this is to make the `light` notices appear like any other line of text. Here are a few ideas. What do you think?

```diff
diff --git a/src/amo/components/RatingManagerNotice/styles.scss b/src/amo/components/RatingManagerNotice/styles.scss
index 69990f701..5b723d982 100644
--- a/src/amo/components/RatingManagerNotice/styles.scss
+++ b/src/amo/components/RatingManagerNotice/styles.scss
@@ -1,10 +1,5 @@
 @import '~amo/css/styles';
 
-.RatingManager .AddonReviewCard-allControls {
-  padding-bottom: 3px;
-  padding-top: 2px;
-}
-
 .RatingManager .RatingManagerNotice-savedRating-hidden {
   display: none;
 }
diff --git a/src/ui/components/Notice/styles.scss b/src/ui/components/Notice/styles.scss
index 30fb9a1e3..5bdb9624a 100644
--- a/src/ui/components/Notice/styles.scss
+++ b/src/ui/components/Notice/styles.scss
@@ -14,6 +14,7 @@
     background-color: transparent;
     display: flex;
     justify-content: center;
+    line-height: $default-line-height;
     padding: 0;
   }
 
@@ -45,6 +46,14 @@
   width: 16px;
 }
 
+.Notice-light .Notice-icon {
+  // Reduce the icon size for light notices so that the total
+  // height of the notice adds up to $default-line-height.
+  background-size: 11px 11px;
+  height: 11px;
+  width: 11px;
+}
+
 .Notice-dismisser-button {
   cursor: pointer;
   padding: 0;
@@ -55,6 +64,10 @@
   margin: 0;
   padding: 4px 0;
 
+  .Notice-light & {
+    padding: 0;
+  }
+
   // The spec calls for additional spacing
   // in front of the action button. This is
   // applied even without a button to strike
```

bobsilverberg: Nice! That works well. I'll update the css. Thanks @kumar303.

PR #6402, opened 2018-09-25
Do not dispatch setLatestReview after updating a review

### LINKED ISSUE #6271: The previous text is displayed after changing the star rating
STR:
1. Load AMO dev, login and go to an add-ons detail page to make a rating
2. Click the Read all reviews link and open the reviews page 
3. Click on "Edit my review" and edit the text then click on a star to change it's rating (or the same star)
4. Observe the text

Actual result:
The previous without edits is visible.

Expected result:
The edited text is available.

Notes:
Reproducible on AMO dev with FF62 (Win10). 
Not reproducible with "Edit my review" from the add-ons detail page.

previous text

bobsilverberg:
Fixes #6271 

This is a bit odd. The thing that's causing the list of reviews to reload is the call to `setLatestReview` in the reviews saga. I added that in https://github.com/mozilla/addons-frontend/commit/e4e68bb734b272cabf5723ddd8b66e459bc058a7#diff-170eeec007168b88041e21f7420f424a in order to fix https://github.com/mozilla/addons-frontend/issues/5941, but now it seems that we do not need that anymore. If I remove the call to `setLatestReview` from the saga, the notifications still work.

I'm not sure if this is because of other things that have landed in the meantime, but it is a bit confusing. From what I can tell this change does not break anything and fixes the issue.

### DIFF
--- a/src/amo/sagas/reviews.js
+++ b/src/amo/sagas/reviews.js
@@ -39,7 +39,6 @@ import {
   hideReplyToReviewForm,
   setAddonReviews,
   setGroupedRatings,
-  setLatestReview,
   setReview,
   setReviewReply,
   setReviewWasFlagged,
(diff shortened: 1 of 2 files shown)


bobsilverberg:
> Do you want to re-assign this issue to me?

Sure.

Issue #6265, opened 2018-09-17
Summary's rating bar won't turn yellow after the given rating


ioanarusiczki:
STR:
1. Load AMO dev and go to an add-ons/theme detail page which has no ratings given yet
2. Click a rating star (such as 5/5) and observe the review summary displayed on the right top corner

Expected result:
Number of ratings increased with one and the yellow bar changes color.

Actual result:
The yellow bar that indicated the 5 star rating is still greyed out.

Notes:
This issue is reproducible on AMO dev with FF62(Win10&Android 8.0).
Refreshing page will show the correct display.

![the yellow bar](https://user-images.githubusercontent.com/33448286/45624030-c43a9d00-ba91-11e8-9862-15f656922379.gif)

kumar303:
Thanks for the report. This will be fixed in https://github.com/mozilla/addons-frontend/issues/5986

kumar303:
This may be unrelated to #5986. I'm looking into it now.

kumar303:
Good catch. I see what the problem is. There was a similar bug where the review count above the star graph was not getting updated. The same patch will fix that, too.

ioanarusiczki:
@kumar303 I cannot reproduce this scenario anymore. It's verified fixed on AMO dev with FF62(Win10 and Android 8.0)

![fixed issue](https://user-images.githubusercontent.com/33448286/45950015-a11a6b00-c006-11e8-89e9-e07c122cc92b.gif)

kumar303:
I had to revert this because of regressions. See https://github.com/mozilla/addons-frontend/pull/6399

I'm going to fix the original bug in a different way.

ioanarusiczki:
I re-checked this issue today and it's not reproducible - AMO dev FF62 - Win10 &Android 8.0

![yelow bar is yellow now](https://user-images.githubusercontent.com/33448286/46289039-72653d00-c590-11e8-84ff-cbd03d512535.gif)

Issue #6356, opened 2018-09-21
Editing a developer reply from a permalink is visually awkward


kumar303:
### Describe the problem and steps to reproduce it:

- Sign in as a developer
- Go to one of your add-ons that has some reviews
- Go to the review listing
- Reply to a review
- Click the timestamp of the review to view its detail page
- From the detail page, click *Edit my review*

### What happened?

It looks like:

<img width="746" alt="screenshot 2018-09-21 16 22 28" src="https://user-images.githubusercontent.com/55398/45906838-3c1f0500-bdbb-11e8-8ec5-9876e4db30a8.png">

### What did you expect to happen?

* There shouldn't be any stars
* ~~You should only see one review form. Multiple forms are open because the reply is also rendered further down in the listing.~~
  * https://github.com/mozilla/addons-frontend/issues/6443 has been opened to address this.
* ~~The link to edit should have been *Edit my reply*~~
* The buttons should say *Update reply*, etc

AlexandraMoga:
If you venture on selecting a star, an error message will also be displayed:

![image](https://user-images.githubusercontent.com/31961530/46019096-fed0b500-c0e3-11e8-8fe0-fd82a40799d7.png)

bobsilverberg:
@kumar303 maybe a change landed since you reported this, but there is one aspect of your report I cannot reproduce.  You say:

> The link to edit should have been Edit my reply

I do see "Edit reply" as the link in the featured review card:

![screenshot 2018-09-26 15 37 50](https://user-images.githubusercontent.com/142755/46104821-57668780-c1a2-11e8-9c10-e439a5adb599.png)

Is that bug no longer valid, or am I misunderstanding?

kumar303:
> I do see "Edit reply" as the link in the featured review card

Yep, this is fixed now. I crossed it out in the description. It looks like the other things I mentioned in the description still need fixing.

bobsilverberg:
> The page must be refreshed to see the updated text down below in the list of reviews. Same if the response is edited from the list of reviews, not a big issue but let me know if we should track this with another issue.

I opened https://github.com/mozilla/addons-frontend/issues/6441, which I think addresses the issue you describe. Do you agree @ioanarusiczki?

ioanarusiczki:
@bobsilverberg Yes, I agree. It's in here https://github.com/mozilla/addons-frontend/issues/6441

PR #6409, opened 2018-09-25
Update add-on data after submitting a review / rating

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6265 (again) by simply reloading the add-on to get new rating and review counts. 

My first proof was to add an entirely new action + reducer but luckily I hadn't yet spent time writing tests 😁


bobsilverberg:
Yeah, that makes sense, although even in the update case dealing with caching can be tricky. I think I may have come across a case when working on #6356 where our current caching setup is going to cause an issue.

Issue #6192, opened 2018-09-10
Investigate ways to reduce the main JS bundle size (amo app)


willdurand:
The `amo` JS bundle gets bigger and bigger as we add more features. We use [bundlesize](https://github.com/siddharthkp/bundlesize) to fail a Travis CI build if the bundle gets too big given a threshold. The threshold has been increased in #6191 to avoid failing PRs but we should see if we can do anything to reduce the bundle size.

Various ideas: tree shaking, code-splitting, dead code elimination, babel-preset-env (and browserlist rules), ...

diox:
Worth keeping this issue open to try to find a solution (might be a good candidate for our [Performance project](https://github.com/orgs/mozilla/projects/150)). We went from 400 KB to 475 KB limit in 2 years, this is only getting worse.

diox:
From duplicate:
> Since we do Server Side Rendering and our JavaScript bundle is at the bottom of the pages, its size doesn't seem to have a huge negative impact on Core Web Vitals, but still, reducing that size would help performance in general - make the site more reactive for things that do depend on client-side JavaScript. So let's look at what we can do.

willdurand:
See: https://github.com/mozilla/addons-frontend/issues/10171

willdurand:
As of March 9, 2021, the JS bundle size is 385KB.

willdurand:
I identified a kinda big dependency (especially given that it's only used once) in #10179. I am not sure what else we can do.

willdurand:
I think we can close this issue now.

PR #6411, opened 2018-09-25
I6192 setting up code splitting by route

rebmullin:
This sets up some code-splitting at the routing level. I was following this:

https://reactjs.org/docs/code-splitting.html#route-based-code-splitting. 

Regarding tree-shaking, I didn't have much luck with this (i.e. I didn't see any difference) but I can investigate that more..
I can also look into babel-preset-env but is there a generally guideline for this - - do we have env list, rules, etc in mind?

We can also customize the Loading a bit more (for errors too) but thought that I would get some feedback on this before I went into that more :)


willdurand:
Thanks for starting this, do you have an associated issue? I am wondering where this idea comes from.

My 2 cents: we should really upgrade to Webpack 4 and probably define goals before taking this direction (and measure). While code-splitting looks nice, it is likely not bug-free and could make some things more complicated. It was already tricky to make the HCT patch work well, so moving all pages to code-splitting seems too bold for me.

rebmullin:
I was looking at #6192. I'll close this tho and we can revisit after the webpack 4 upgrade..

PR #6415, opened 2018-09-26
Fix language tools listing

### LINKED ISSUE #6145: fa-IR dictionary doesn't appear on Language Tools page
This dictionary doesn't appear in the [Language Tools](https://addons.mozilla.org/firefox/language-tools/) page: https://addons.mozilla.org/addon/lilak-persian-dictionary/.

The target locale was already set to `fa-IR`, and it can be seen in the Language Tools page in the old front end. There are two entries for Persian on the old page. The new one only has one for `fa`.

willdurand:
Fixes #6145

---

This PR fixes #6145 in two ways:

- it adds `fa-IR` (Persian for Iran) in the list of supported languages,
  which creates a new row for the related language tools
- it lists more add-ons by adding the add-ons with a long locale (e.g.,
  `az-IR`) in the sets of add-ons (dict or lang pack) of the
  corresponding short locale (e.g., `az`), but only if this long locale
  is not supported (otherwise there would be add-ons listed multiple
  times (e.g., the Austrian German would be listed in "German" `de` and
  "German (Austria)" `de-AT`).

### DIFF
--- a/tests/unit/amo/pages/TestLanguageTools.js
+++ b/tests/unit/amo/pages/TestLanguageTools.js
@@ -230,7 +230,7 @@ describe(__filename, () => {
     expect(root.find('title')).toHaveText('Dictionaries and Language Packs');
   });
 
-  it('renders add-ons for all variants of a short locale', () => {
+  it('renders add-ons for all variants of a short locale on a single row (only one supported language)', () => {
     // The short locale is `az` here.
     const addons = [
       createFakeLanguageTool({
@@ -264,7 +264,7 @@ describe(__filename, () => {
     ]);
   });
 
-  it('does not render add-ons for all variants of a short locale when the variant is a supported language', () => {
+  it('renders add-ons on two different rows corresponding to two supported languages', () => {
     // The short locale is `fa` here, which is in the list of supported
     // languages (`src/core/languages.js`) together with `fa-IR`.
     const addons = [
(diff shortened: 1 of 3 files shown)


tests/unit/amo/pages/TestLanguageTools.js
```diff
@@ -229,4 +229,78 @@ describe(__filename, () => {
...
+    const root = renderShallow({ store });
+
+    expect(root.find('.LanguageTools-table-row')).toHaveLength(2);
+
+    expect(root.find(LanguageToolList).at(0)).toHaveProp('languageTools', [
+      addons[0],
+    ]);
+    expect(root.find(LanguageToolList).at(1)).toHaveProp('languageTools', [
+      addons[1],
+      addons[2],
```
rebmullin: Aren't you showing here that it *does* render the addons listed above? I don't see where /how you are showing it's not rendering. I am not sure I fully understand this so maybe I am missing something..

willdurand: The test case description could be rephrased I think, all add-ons are eventually displayed. In this test case, they should not be displayed on the same "row" though.

rebmullin: Oh I see...yeh IMO that would help to rephrase that then

tests/unit/amo/pages/TestLanguageTools.js
```diff
@@ -229,4 +229,78 @@ describe(__filename, () => {
 
     expect(root.find('title')).toHaveText('Dictionaries and Language Packs');
   });
+
+  it('renders add-ons for all variants of a short locale', () => {
```
rebmullin: should this say in 1 row (to differentiate from the following test case)?

PR #6420, opened 2018-09-26
Use API v4

### LINKED ISSUE #5527: upgrade api from v3 to v4

willdurand:
Fixes #5527 

---

I did not fix all the props, especially around ratings since we use this term for everything (I kept `rating` for props, but I used `score` for data).


tests/unit/core/api/test_index.js
```diff
@@ -82,7 +82,7 @@ describe(__filename, () => {
       const endpoint = 'project-ă-ă-â-â-日本語';
       mockWindow
         .expects('fetch')
-        .withArgs(sinon.match(`/api/v3/${endpoint}/`))
```
bobsilverberg: I wonder if it would be worth making the version a `const` in this file so that if we ever need to change this again it can be changed in one place instead of in multiple tests?

PR #6425, opened 2018-09-26
Recreate AMInstallButton animation in pure CSS

### LINKED ISSUE #6207: Create "loading" animation in CSS
From: https://github.com/mozilla/addons-frontend/issues/6169#issuecomment-419372479

---

Let's try to recreate the SVG animation in plain old CSS to avoid a (rather) huge SVG to be loaded and animated in CSS. WIP at: https://jsfiddle.net/muffinresearch/9x7bshnz/77/

On IRC:

> Probably need to consider if it's possible to use other props to get hardware acceleration for that animation. height/width won't.

willdurand:
Fixes #6207 

---

This should be better performance-wise.


src/core/components/AMInstallButton/index.js
```diff
@@ -321,13 +321,17 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
...
+              title={this.getButtonText()}
             >
-              <span className="AMInstallButton-loading-icon">
-                <Icon alt={this.getButtonText()} name="loading" />
-              </span>
+              <div className="AMInstallButton-loader">
+                <div className="AMInstallButton-loader-container">
+                  <div className="AMInstallButton-loader-ball" />
+                </div>
+                <span className="visually-hidden">{this.getButtonText()}</span>
```
rebmullin: can the `this.getButtonText()` be set to a var since it's used a few times?

src/core/components/AMInstallButton/styles.scss
```diff
@@ -59,57 +24,106 @@
...
       [dir='rtl'] & {
-        // For some reasons, there is a small horizontal offset in RTL mode.
-        margin-left: -1px;
+        animation-name: move-rtl;
       }
     }
+
+    .AMInstallButton-loader-ball {
+      animation: squash $speed ease-in-out infinite alternate;
+      background: $blue-60;
```
rebmullin: where is this defined?

willdurand: This is part of [photon](https://design.firefox.com/photon/welcome.html), which has a lib required in `src/ui/css/vars.scss`.

src/core/components/AMInstallButton/index.js
willdurand: yep

PR #6430, opened 2018-09-26
I5928 collection remove pg switch fix

### LINKED ISSUE #5928: Pagination jumps to the last page when an add-on is removed in the edit collection form
STR:

1. Log in to AMO -dev
2. Access you Collections page
3. Select a collection that has >25 add-on to edit it
4. Remove an add-on from the current page
5. After removal, check again the page number in the URL

Actual result:
Collection jumps to the last page in the collection

Expected result:
User remains on the same page after an add-on is deleted (provided that he doesn't clear a whole page of add-ons)

Notes:
- in the video below, it is clearly visible that after deleting an add-on from page 4, pagination jumps to page 7, which is the last page in the collection
- reproduced on ANO dev with FF61, Win10x64 
- probably a regression from #5621 

wrong page

rebmullin:
fixes #5928

### DIFF
--- a/src/amo/pages/Collection/index.js
+++ b/src/amo/pages/Collection/index.js
@@ -106,10 +106,14 @@ export const computeNewCollectionPage = (
 
   let page = '1';
   if (pageSize) {
-    page = Math.ceil((numberOfAddons - 1) / pageSize);
-    if (parseInt(currentPage, 10) < page) {
-      page = currentPage;
+    const lastPage = Math.ceil((numberOfAddons - 1) / pageSize);
+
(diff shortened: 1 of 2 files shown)


src/amo/pages/Collection/index.js
```diff
@@ -100,12 +100,20 @@ export type SaveAddonNoteFunc = (
 
 export const computeNewCollectionPage = (
   collection: CollectionType,
+  currentPage: string,
```
willdurand: We should switch to a single-object argument here:

```js
{ collection, currentPage }
```

PR #6435, opened 2018-09-27
I6338 adjusting back functionality on collection cancel

### LINKED ISSUE #6338: The create a collection form is displayed instead of the collection page after "cancel" is clicked
STR:
1. Load AMO dev and go to View my Collections and click on Create a collection
2. Change the language from the footer. I used DE
3. Click the "Cancel" button
4. Observe the page 

Actual result:
"Create a collection" form is displayed in English. 

Expected result:
The collection page is displayed in German.

Notes:
This issue is reproducible on AMO dev and AMO stage with all languages FF62(Win10)

language1

rebmullin:
fixes #6338

### DIFF
--- a/src/amo/components/CollectionManager/index.js
+++ b/src/amo/components/CollectionManager/index.js
@@ -42,7 +42,7 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
-  clientApp: ?string,
+  clientApp: string,
   currentUsername: string,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
@@ -50,7 +50,7 @@ type InternalProps = {|
   i18n: I18nType,
   isCollectionBeingModified: boolean,
   location: ReactRouterLocationType,
-  siteLang: ?string,
+  siteLang: string,
 |};
 
 type State = {|
@@ -80,10 +80,11 @@ export class CollectionManagerBase extends React.Component<
   }
 
   onCancel = (event: SyntheticEvent<HTMLButtonElement>) => {
-    const { creating, dispatch, history } = this.props;
+    const { clientApp, creating, dispatch, history, siteLang } = this.props;
 
     if (creating) {
-      history.goBack();
+      // This is used over history.goBack in case the language changes.
+      history.push(`/${siteLang}/${clientApp}/collections/`);
     }
 
     event.preventDefault();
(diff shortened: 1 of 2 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -42,15 +42,15 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
-  clientApp: ?string,
+  clientApp: string,
```
bobsilverberg: Do you know why these were marked as optional before? That suggests there is a case when they might be missing but now we're saying they will never be missing. Is that valid?

src/amo/components/CollectionManager/index.js
```diff
@@ -80,10 +80,11 @@ export class CollectionManagerBase extends React.Component<
   }
 
   onCancel = (event: SyntheticEvent<HTMLButtonElement>) => {
-    const { creating, dispatch, history } = this.props;
+    const { clientApp, creating, dispatch, history, siteLang } = this.props;
 
     if (creating) {
-      history.goBack();
+      // This is used over history.goBack in case the language changes.
```
bobsilverberg: I don't think we need this comment, as we can no longer see that this used to call `history.goBack`, so I'm not sure what the comment adds.

rebmullin: ok yeh I guess it loses context so yep I'll remove.. thanks : )

src/amo/components/CollectionManager/index.js
```diff
@@ -80,10 +80,14 @@ export class CollectionManagerBase extends React.Component<
...
   onCancel = (event: SyntheticEvent<HTMLButtonElement>) => {
-    const { creating, dispatch, history } = this.props;
+    const { clientApp, creating, dispatch, history, siteLang } = this.props;
 
     if (creating) {
-      history.goBack();
+      if (siteLang) {
+        history.push(`/${siteLang}/${clientApp}/collections/`);
+      } else {
+        history.goBack();
```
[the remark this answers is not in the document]
bobsilverberg: It makes sense to me, if we're saying that `siteLang` can be missing.

tests/unit/amo/components/TestCollectionManager.js
```diff
@@ -668,8 +668,27 @@ describe(__filename, () => {
...
+    simulateCancel(root);
+
+    sinon.assert.calledWith(
+      fakeHistory.push,
+      `/${siteLang}/${clientApp}/collections/`,
+    );
+  });
+
+  it('calls history.goBack() when creating and language is falsey', () => {
+    const siteLang = null;
```
bobsilverberg: To match the Flow description, this should actually be `undefined`, not `null`, because `null` would be considered invalid by Flow.

Issue #6441, opened 2018-09-27
When saving changes to a reply, it is not updated in other places


bobsilverberg:
There are a couple of different STRs:

From a permalink:

1. Log in as an admin and go to the list of reviews for an add-on.
2. Add a reply to one of the reviews.
3. Click the permalink of the reply you just created to be brought to the permalinked page for the reply.
4. In the "Reply by..." card, click the "Edit reply" link.
5. Change the text of the reply and click "Update reply".

Expected result:
The text of the reply is changed in both the "Reply by..." card and in the embedded reply which is visible in the list of reviews.

Actual result:
The text of the reply is changed in the "Reply by..." card but the old text is visible in the list of reviews.

From the list of reviews:

1. Log in as an admin and go to the list of reviews for an add-on.
2. Add a reply to one of the reviews.
3. Click the permalink of the reply you just created to be brought to the permalinked page for the reply.
4. In the embedded reply in the list of reviews (i.e., the one with the blue background), click the "Edit reply" link.
5. Change the text of the reply and click "Update reply".

Expected result:
The text of the reply is changed in both the "Reply by..." card and in the embedded reply which is visible in the list of reviews.

Actual result:
The text of the reply is changed in the embedded reply which is visible in the list of reviews but the old text is visible in the the "Reply by..." card.

ioanarusiczki:
@bobsilverberg This issue is reproducible with developer accounts too. If page is refreshed the last edit will be correctly displayed in both places.  - FF62(Win10)

![developer account](https://user-images.githubusercontent.com/33448286/46338050-cd05a400-c637-11e8-96aa-4bf9dc80a9d9.gif)

kumar303:
We could fix this but showing the two forms is just a bug that we will fix in https://github.com/mozilla/addons-frontend/issues/6443 Thus, I'm closing this.

bobsilverberg:
I'm not sure that this is related to the two forms, but maybe I'm misunderstanding the issue. To me this is about the fact that editing a reply in one place does not clear the cache and refresh the reply in another place. E.g., editing a reply in the Featured review card does not update the text of the reply as it appears in an embedded reply further down the page. I'm not sure how not displaying two forms would directly address that.

Issue #6438, opened 2018-09-27
Deleting a review from its permalink shows 'review was not found'


kumar303:
### Describe the problem and steps to reproduce it:

* Sign in as a non-administrator
* Review an add-on
* Go to the review listing
* Click the timestamp of the review you just wrote to view its detail page
* On the detail page, click *Delete review* and confirm deletion

### What happened?

The review is deleted but you now see *The review was not found* at the top of the page.

### What did you expect to happen?

The page should redirect you to the review listing.

bobsilverberg:
I implemented this "review was not found" message to address the case where someone shares a permalink to a review, but the review has been deleted. If we no longer show a message but rather just redirect the user to the review listing page, with no acknowledgment that they intended to visit a particular review, I think that could be confusing to them. I think it better to support that use case, and have a minor annoyance of an extra message at the top of the page immediately after you delete a review from a permalink.

We could change the message to "The review has been deleted by the author", which would be more specific and would address both cases.

What do you think @kumar303?

kumar303:
Ah, ok, that makes sense and I see how it could be helpful. I'm not sure if we can easily determine whether a review was deleted or it simply doesn't exist. It's probably not worth the effort. I'm just going to close this for now.

PR #6442, opened 2018-09-27
Editing a developer reply from a permalink is visually awkward

### LINKED ISSUE #6356: Editing a developer reply from a permalink is visually awkward

bobsilverberg:
Fixes #6356 

Before:

screenshot 2018-09-27 13 42 25

After:

screenshot 2018-09-27 13 41 35

### DIFF
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -77,6 +77,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
       puffyButtons,
     } = this.props;
 
+    const isReply = review.isDeveloperReply;
     const reviewGuideLink = i18n.sprintf(
       i18n.gettext(
         'Please follow our %(linkStart)sreview guidelines%(linkEnd)s.',
@@ -88,7 +89,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     );
 
     /* eslint-disable react/no-danger */
-    const formFooter = !review.isDeveloperReply ? (
+    const formFooter = !isReply ? (
       <span dangerouslySetInnerHTML={sanitizeHTML(reviewGuideLink, ['a'])} />
     ) : (
       undefined
@@ -100,15 +101,12 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     );
 
     let submitButtonText = i18n.gettext('Submit review');
-    if (review.body) {
-      submitButtonText = review.isDeveloperReply
-        ? i18n.gettext('Update reply')
-        : i18n.gettext('Update review');
-    }
-
     let submitButtonInProgressText = i18n.gettext('Submitting review');
     if (review.body) {
-      submitButtonInProgressText = review.isDeveloperReply
+      submitButtonText = isReply
+        ? i18n.gettext('Update reply')
+        : i18n.gettext('Update review');
+      submitButtonInProgressText = isReply
         ? i18n.gettext('Updating reply')
         : i18n.gettext('Updating review');
     }
@@ -116,7 +114,7 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     return (
       <div className="AddonReviewManager">
         {errorHandler.renderErrorIfPresent()}
-        {!review.isDeveloperReply && (
+        {!isReply && (
           <div className="AddonReviewManager-starRating">
             <span>{i18n.gettext('Your star rating:')}</span>
             <Rating
--- a/src/amo/components/AddonReviewManager/index.js
+++ b/src/amo/components/AddonReviewManager/index.js
@@ -88,8 +88,10 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     );
 
     /* eslint-disable react/no-danger */
-    const formFooter = (
+    const formFooter = !review.isDeveloperReply ? (
       <span dangerouslySetInnerHTML={sanitizeHTML(reviewGuideLink, ['a'])} />
+    ) : (
+      undefined
     );
     /* eslint-enable react/no-danger */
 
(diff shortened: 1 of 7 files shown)


bobsilverberg:
Note that I uncovered a bug while testing this which I opened at https://github.com/mozilla/addons-frontend/issues/6441. Note also that I did not address the issue of two reply forms opening in the patch, but opened https://github.com/mozilla/addons-frontend/issues/6443 as a follow-up.

tests/unit/amo/sagas/test_reviews.js
```diff
@@ -674,6 +676,26 @@ describe(__filename, () => {
...
+        );
+
+      _updateAddonReview();
+
+      const expectedAction = hideFlashedReviewMessage();
+      await sagaTester.waitFor(expectedAction.type);
+
+      const unexpectedAction = fetchAddon({ errorHandler, slug });
+
+      expect(sagaTester.getCalledActions().map((a) => a.type)).not.toContain(
```
rebmullin: can you just do this here:
` expect(sagaTester.numCalled(unexpectedAction.type)).toEqual(0);` ?

tests/unit/amo/sagas/test_reviews.js
```diff
@@ -792,6 +814,44 @@ describe(__filename, () => {
...
+
+      const unexpectedAction = setLatestReview({
+        addonId,
+        addonSlug,
+        review: externalReview,
+        userId,
+        versionId,
+      });
+
+      expect(sagaTester.getCalledActions().map((a) => a.type)).not.toContain(
```
rebmullin: same here? or the mapping necessary?

src/amo/components/AddonReviewManager/index.js
```diff
@@ -88,41 +88,59 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
...
+    let submitButtonText = i18n.gettext('Submit review');
+    if (review.body) {
+      submitButtonText = review.isDeveloperReply
+        ? i18n.gettext('Update reply')
+        : i18n.gettext('Update review');
+    }
+
+    let submitButtonInProgressText = i18n.gettext('Submitting review');
+    if (review.body) {
+      submitButtonInProgressText = review.isDeveloperReply
```
rebmullin: just a thought..can this be moved up above (under the above review.body)?

src/amo/components/AddonReviewManager/index.js
```diff
@@ -88,41 +88,59 @@ export class AddonReviewManagerBase extends React.Component<InternalProps> {
     );
 
     /* eslint-disable react/no-danger */
-    const formFooter = (
+    const formFooter = !review.isDeveloperReply ? (
```
rebmullin: is worth setting review.isDeveloperReply to a var since it's used a few times here?

bobsilverberg:
@rebmullin Regarding the issue you're seeing with the delete functionality, this seems kind of similar to issue #6441, so maybe whatever we do to address that will also fix this issue. If not, then your idea of re-directing to the reviews listing after deleting a reply might be a good solution.

tests/unit/amo/sagas/test_reviews.js
bobsilverberg: Agreed, this seems simpler. I was just following the model we use everywhere else. Thanks @rebmullin!

src/amo/sagas/reviews.js
```diff
@@ -281,21 +281,23 @@ function* manageAddonReview(
...
-    yield put(
-      setLatestReview({
-        addonId: reviewFromResponse.addon.id,
-        addonSlug: reviewFromResponse.addon.slug,
-        review: reviewFromResponse,
-        userId: reviewFromResponse.user.id,
-        versionId: reviewFromResponse.version.id,
-      }),
-    );
+    if (!reviewFromResponse.is_developer_reply) {
```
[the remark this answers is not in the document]
bobsilverberg: I did this because `reviewFromResponse.version` is missing and therefore the `invariant` is being triggered. I realized that if we are updating a reply we really don't need to do either of these things (`setLatestReview` or `fetchAddon`), so it made sense to just not do them.

Looking closer now, I'm not sure that the fact that `version` is missing from the response is expected, and maybe that's a bug on the server side. The [docs](https://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#rating-detail-object) don't mention that `version` is optional. I guess I'll file a bug on addons-server for that, but it still makes sense to me that we wouldn't dispatch either of those actions for a reply update as neither are needed.

@kumar303 What do you think?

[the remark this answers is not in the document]
bobsilverberg: I opened https://github.com/mozilla/addons-server/issues/9550 to report the missing `version`, although I now suspect that that is expected.

bobsilverberg:
Actually, @rebmullin, can you confirm whether you are logged in with an admin account when you saw the delete issue? If you were then it might be invalid. Admin users can see deleted reviews, and there is no data currently available to allow us to know that a review/reply has been deleted. So if you are an admin and you are at the details screen for a review/reply and you delete it, you will still see if in the FeaturedReviewCard because the app will still be able to retrieve it from the API.

bobsilverberg:
Thanks for the reviews @rebmullin and @kumar303. A new version is now available for review.

kumar303:
@rebmullin 

> Click delete my review (or reply) ... I would think it would return you back to the reviews page maybe.? 

Yep, I noticed this the other day, too. We will fix it here: https://github.com/mozilla/addons-frontend/issues/6438

src/amo/sagas/reviews.js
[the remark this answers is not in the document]
kumar303: Interesting. It looks like someone had already figured out that the version could be missing because our Flow type is correct:

https://github.com/mozilla/addons-frontend/blob/a238ae43335c366c34569a29d6c021130ef982c0/src/amo/api/reviews.js#L31-L43

kumar303: > it still makes sense to me that we wouldn't dispatch either of those actions for a reply update as neither are needed.

Yes, I agree. I see it now. I guess `fetchAddon` would be possible since it doesn't require `version` but I can't think of why we need it.

PR #6456, opened 2018-09-28
Don't refresh listings when updating a review

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6271

For inserting and deleting reviews, we need to refresh add-on listings so that the data gets re-fetched from the API and the pages stay in sync. For updates, we don't need to do that. Also the update form is on a listing page (d'oh) so refreshing causes a couple bugs like the one here.

This patch does not refresh listings on update.


kumar303:
> I don't like the way we've had to change setReview to accept an isUpdate argument ... I guess it kind of feels like that shouldn't be its responsibility.

Yep, I agree.

I went through a bit of trial and error to fix this bug. The "right" way to fix it IMO is to automatically detect updates in the reducer. This doesn't work, however, because both `setReview` and `setLatestReview` need to know if it's an update. Depending on which one happens first, the second one can no longer automatically detect if it's an update.

With that said and after a weekend to stew on it, I can try to make `setLatestReview` *not* need to know if an update is happening. I'm not sure if that's possible but I can give it a try. After that, maybe I can move things back to the reducer.

> I wonder if it mightn't be better to wait until the new UI is enabled on prod ... at which point we can remove all of the code that supports the old UI

I would be fine with reworking this after deleting the old code. I would also like to fix this bug before enabling the new UI, if possible. It's pretty janky and there could be actual data loss if someone were in the middle of writing a review (like the STR for this bug).

kumar303:
> I can try to make setLatestReview not need to know if an update is happening.

I got this to to work 🎉 

Now I can give the reducer the sole responsibility of auto-detecting updates.

kumar303:
@bobsilverberg can you take another look? This moves the update detection to the reducer which I think makes the most sense. The problem with the last patch is `setLatestReview` was trying to do too much. I have moved some of its logic into `setReview`.

src/amo/reducers/reviews.js
```diff
@@ -261,28 +261,74 @@ export const selectLatestUserReview = ({
...
+  newReviewBody: ?string,
+|}) {
+  const existingReview = selectReview(reviewsState, newReviewId);
+  if (!existingReview) {
+    return false;
+  }
+
+  // If this update is actually turning a rating into a review then
+  // it's not an update.
+  if (!existingReview.body && newReviewBody) {
```
bobsilverberg: This causes a problem for the User Profile page. If I click the "Write a review" button for a rating-only review, when I submit the review it is not treated as an update and therefore the page is refreshed. If I also had a text box open editing an existing review at the time, I would lose my state. 

I don't think this is that big a deal, as it's a more extreme edge case, and I'm not sure how we can make it work the way we want both from the User Profile page and from the RatingManager (because that *does* wanted it treated like an insert and not an update), but it's some food for thought.

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -981,120 +987,160 @@ describe(__filename, () => {
...
       state = reviewsReducer(
         state,
-        setGroupedRatings({ addonId, grouping: grouping1 }),
+        setGroupedRatings({
+          addonId,
+          grouping: { 1: 0, 2: 0, 3: 0, 4: 0, 5: 0 },
+        }),
       );
-      state = reviewsReducer(
+
```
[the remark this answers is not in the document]
kumar303: The storage of grouped ratings is already covered by another test (below) so it would be redundant to re-test it.

https://github.com/mozilla/addons-frontend/blob/166c06d965446530bfcf168128ee011c4419ee04/tests/unit/amo/reducers/test_reviews.js#L1158-L1166

src/amo/reducers/reviews.js
kumar303: > [While submitting a new review, ] if I also had a text box open editing an existing review at the time, I would lose my state.

I don't think that's worth fixing. The user would not do this intentionally.

kumar303: > If I click the "Write a review" button for a rating-only review, when I submit the review it is not treated as an update and therefore the page is refreshed

I have a fix for this.

kumar303:
The user profile refresh was a good catch, thanks. This was a bug in the logic for when and when not to refresh listing data. I fixed this and also made the implementation of `addReviewToState` more direct. I re-tested all behavior so I think this is ready to go but feel free to comment after I merge if you see anything worth following up on.

PR #6461, opened 2018-10-01
Add locale to HTML titles

### LINKED ISSUE #5761: Add locale to page titles
We need to de-duplicate titles, adding the locale the title will help to differentiate what is essentially the same title in both place where just the locale is different. 

The locale should be appended to the end of the title in brackets e.g `This is the title (en-US)`, this garantees non duplicate titles per locale irrespective of the state of translations.

willdurand:
Fix #5761

---

This PR adds the locale to each HTML title.

### DIFF
--- a/src/amo/components/App/index.js
+++ b/src/amo/components/App/index.js
@@ -60,6 +60,7 @@ type Props = {|
   handleGlobalEvent: () => void,
   i18n: I18nType,
   isHomePage: boolean,
+  lang: string,
   location: ReactRouterLocationType,
   logOutUser: () => void,
   mozAddonManager: $PropertyType<MozNavigator, 'mozAddonManager'>,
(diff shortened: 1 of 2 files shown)


src/amo/components/App/index.js
```diff
@@ -205,17 +206,33 @@ export class AppBase extends React.Component<Props> {
...
 
     if (clientApp === CLIENT_APP_ANDROID) {
-      defaultTitle = i18n.gettext('Add-ons for Android');
-      titleTemplate = i18n.gettext('%s – Add-ons for Android');
+      defaultTitle = i18n.sprintf(
+        i18n.gettext('Add-ons for Android (%(locale)s)'),
+        i18nValues,
+      );
+      titleTemplate = i18n.sprintf(
+        i18n.gettext('%%s – Add-ons for Android (%(locale)s)'),
```
muffinresearch: Presumably dennis linter will catch this escaping being broken if a localizer accidentally removes one of the `%`?

willdurand: I will introduce a variable because dennis does not do anything.. `%%` is supported as regular content (https://github.com/willkg/dennis/commit/3c65a1ddca3628852c67d3c182adc63fcbaed330), thus it is subject to change.

PR #6463, opened 2018-10-01
Use 301 status code in local redirects

### LINKED ISSUE #5677: Replace 302 locale redirects with 301
To improve the SEO signals we should use a 301 for locale redirects instead of a 302.

willdurand:
Fix #5677

### DIFF
--- a/src/amo/pages/SearchPage/index.js
+++ b/src/amo/pages/SearchPage/index.js
@@ -91,7 +91,7 @@ export class SearchPageBase extends React.Component<InternalProps> {
 
       props.dispatch(
         sendServerRedirect({
-          status: 302,
+          status: 301,
           url: `/${lang}/${clientApp}/search/${queryString}`,
         }),
       );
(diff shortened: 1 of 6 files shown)


willdurand:
> Bear in mind the implicit caching around a 301 might make it undesirable in some cases.
> 
> Erring on the side of caution if you assume that the 301'd location will always be returned in the future for the same original URL request that's probably the best way to see that it's the right fit for a given case.

Indeed. This PR updates the status code in three cases:

- supporting the value `all` for the "operating system" search parameter, which is a legacy value (we use nothing/empty now). Using 301 makes sense to get rid of this parameter.
- supporting the old `atype` search parameter. Using 301 makes sense to get rid of this parameter too, since we map its value to the `addonType` equivalent.
- "fixing" the URL (prefix middleware), which makes sure the client is valid, the locale is valid too (and with the right case), etc. I also think it is worth using 301 here.

PR #6467, opened 2018-10-01
Use getFileHash in installAddon HOC

### LINKED ISSUE #1208: Pass add-on hash into mozAddonsManager
Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1308261 for more.

willdurand:
Fixes #1208 

---

I reused an existing function (moved to `core/utils/addons`), so the diff looks big but it is not. The main part is calling `getFileHash` in the `install()` method of the HOC.

### DIFF
--- a/src/core/addonManager.js
+++ b/src/core/addonManager.js
@@ -29,7 +29,7 @@ type FirefoxAddon = {|
 
 export type MozAddonManagerType = {|
   addEventListener: (eventName: string, handler: Function) => void,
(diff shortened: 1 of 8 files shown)


willdurand:
> I didn't run the code but it looks good to me. Do installs break if you hard code some garbage for hash? That would be a good way to test that it's hooked up, especially since mozAddonManager isn't documented too well.

Yep, that's what I did and it fails with a message "download failed".

PR #6471, opened 2018-10-01
Add admin and review links for add-on details page

### LINKED ISSUE #4632: Add admin and review links for add-on details page
Broken off from #2756.
~~Depends on #5548~~

These links should appear in the "More Information" box, at the bottom of the "Add-on Links" section. The following links are needed:

* "Edit add-on". Goes to `/developers/addon/SLUG/edit`. Requires `Addons:Edit` permission to appear.
* "Admin add-on". Goes to `/admin/addon/manage/SLUG/`. Requires `AdminTools:View` permission to appear. 
* "Content review add-on". Goes to `/reviewers/review-content/SLUG`. Requires  `Addons:ContentReview` permission to appear.
* "Code review add-on". Goes to `/reviewers/review/SLUG`. Requires `Addons:PostReview` permission to appear. Should not appear for themes.
* "Review theme". Goes to `/reviewers/review/SLUG`. Requires `Personas:Review` permission to appear. Should only appear for themes.

bobsilverberg:
Fixes #4632 

Here's an example for an extension for a user with all of the permissions:

screenshot 2018-10-01 11 11 12

### DIFF
--- a/src/amo/components/AddonAdminLinks/index.js
+++ b/src/amo/components/AddonAdminLinks/index.js
@@ -3,7 +3,6 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
-import Link from 'amo/components/Link';
 import {
   ADDONS_CONTENTREVIEW,
   ADDONS_EDIT,
@@ -45,7 +44,7 @@ export class AddonAdminLinksBase extends React.Component<InternalProps> {
       i18n,
     } = this.props;
 
-    if (!addon) {
+    if (addon === null) {
       return null;
     }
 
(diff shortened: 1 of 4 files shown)


bobsilverberg:
Thanks for the review @willdurand. It does feel much better as a separate component. As it is a significant change, do you mind taking another look? 

Also, I'm having trouble with Flow. I understand the issue, but I'm not sure how to make Flow happy. The problem is that `DefinitionList` expects its children to be of type `Definition`, but I am trying to use an instance of `AddonAdminLinks` as a child.  `AddonAdminLinks` does return a `Definition`, but that doesn't seem to make Flow happy. I thought maybe I could have `AddonAdminLinks` return a `<ul>` instead, and then wrap that in a `<Definition>` in `AddonMoreInfo`, but the problem with that is that we don't know whether to display anything from within `AddonMoreInfo` - it's only `AddonAdminLinks` that knows whether to display anything or not.

Any advice you might have for dealing with this would be appreciated.

bobsilverberg:
I had a thought about this in the middle of the night last night. Maybe I should not try to put the component inside the exiting `DefinitionList`, but inside its own at the end. I'll try that and see if it resolves the issue.

bobsilverberg:
Ok, I fixed the Flow issue and this is ready for another look @willdurand.

tests/unit/amo/components/TestAddonAdminLinks.js
```diff
@@ -0,0 +1,134 @@
...
+import {
+  dispatchClientMetadata,
+  dispatchSignInActions,
+  fakeAddon,
+  fakeTheme,
+} from 'tests/unit/amo/helpers';
+import { fakeI18n, shallowUntilTarget } from 'tests/unit/helpers';
+
+describe(__filename, () => {
+  const { store } = dispatchClientMetadata();
```
kumar303: This is dangerous because it means the state of each test will leak into the others.

You need to reset the state before each test within `beforeEach(() => ...)` or move the initialization into `render()` or create it with a common helper or something else.

bobsilverberg: Oops, good catch. Thanks!

bobsilverberg:
I have updated the PR with the following changes:

1. Fixed the test issue as reported by @kumar303 at https://github.com/mozilla/addons-frontend/pull/6471#discussion_r222333455.
2. Added the new requirements as requested by @wagnerand at https://github.com/mozilla/addons-frontend/issues/4632#issuecomment-426299194.
3. Added some more tests to verify that links are not shown when their permission is missing.
4. Changed the `Link`s to `<a>`s because they are external.

src/amo/components/AddonAdminLinks/index.js
```diff
@@ -0,0 +1,157 @@
...
+          href={`/reviewers/review-content/${addon.slug}`}
+        >
+          {i18n.gettext('Content review add-on')}
+        </a>
+      </li>
+    ) : null;
+
+    const codeReviewLinkText = isTheme(addon.type)
+      ? i18n.gettext('Review theme')
+      : i18n.gettext('Code review add-on');
```
kumar303: *Review add-on code*, maybe?

tests/unit/amo/components/TestAddonAdminLinks.js
```diff
@@ -0,0 +1,219 @@
...
+    const perms = Array.isArray(permissions) ? permissions : [permissions];
+    return render({
+      addon:
+        addon === undefined
+          ? createInternalAddon({
+              ...fakeAddon,
+              slug,
+            })
+          : addon,
+      store: dispatchSignInActions({
```
kumar303: If you wanted to, you could just re-use the local store:

```js
const renderWithPermissions = ({ addon, permissions }) => {
  perms = ...;
  dispatchSignInActions({ store, userProps: { permissions: perms } });
  return render({ ... });
};
```

Without that, it might be surprising if someone tried to dispatch something from the local store.

tests/unit/amo/components/TestAddonAdminLinks.js
```diff
@@ -0,0 +1,219 @@
...
+  });
+
+  it('shows a theme review link for a theme if the user has permission', () => {
+    const root = render({
+      addon: createInternalAddon({
+        ...fakeTheme,
+        slug,
+      }),
+      store: dispatchSignInActions({
+        userProps: { permissions: [THEMES_REVIEW] },
```
kumar303: can this not use `renderWithPermissions`?

tests/unit/amo/components/TestAddonAdminLinks.js
```diff
@@ -0,0 +1,219 @@
...
+  });
+
+  it('does not show a theme review link if the user does not have permission', () => {
+    const root = render({
+      addon: createInternalAddon({
+        ...fakeTheme,
+        slug,
+      }),
+      store: dispatchSignInActions({
+        userProps: { permissions: [ADDONS_EDIT] },
```
kumar303: This looks like it can use `renderWithPermissions`, too.

tests/unit/amo/components/TestAddonAdminLinks.js
```diff
@@ -0,0 +1,219 @@
...
+      />,
+      AddonAdminLinksBase,
+    );
+  }
+
+  const renderWithPermissions = ({ addon, permissions }) => {
+    const perms = Array.isArray(permissions) ? permissions : [permissions];
+    return render({
+      addon:
+        addon === undefined
```
kumar303: This code would be identical:

```js
const renderWithPermissions = ({ 
  addon = createInternalAddon({
    ...fakeAddon,
    slug,
  }), 
  permissions, 
}) => {
  ...
};
```

src/amo/components/AddonAdminLinks/index.js
bobsilverberg: That does sound better to me, but the text comes directly from https://github.com/mozilla/addons-frontend/issues/4632#issue-308115330. I know that "code review" is a thing, so maybe the current value of "Code review add-on" is still preferred.

@jvillalobos Which do you think is a better label for the link:
* Code review add-on
* Review add-on code

jvillalobos: I like "Review add-on code"

bobsilverberg:
The current failure is due to bundle size, for which I've submitted https://github.com/mozilla/addons-frontend/pull/6519. I think this is fine to merge.

flodolo:
This PR introduces a few strings that are absolutely obscure without context. Localizers need context to localize them, and that's where comments can help.

> Admin add-on

Does it mean "Manage add-on", an "Add-on for admins", something else completely?

> Admin Links

Manage links, links for admin, …?

> Content review add-on
> Admin add-on status

I really don't know about these.

willdurand:
@flodolo yep, sorry about that. There is an issue for this problem: https://github.com/mozilla/addons-frontend/issues/6607

PR #6472, opened 2018-10-01
Add description meta tag on add-on detail pages

### LINKED ISSUE #5766: Add <meta name="description”> for details pages
Details pages should have a meta description tag that uses the summary text for the add-on.

From https://docs.google.com/document/d/1W9LOw7wgbpY6Mw5Zd4V71hJ0YahY4DTNl23MHVpG0_E/edit?ts=5b5b40e5#heading=h.aqjhycqt8ulj

The heading for details pages should be:

“Download xyz for Firefox. Description of addon.”

willdurand:
Fixes #5766 

---

I did not provide a fallback in case the summary is not available since the content should be truncated if we use the full description (and that might lead to a weird content).

Because the name of the add-on is part of the content, there is no risk of similar content if we don't have a summary (content will still be unique).

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -398,6 +398,24 @@ export class AddonBase extends React.Component {
     );
   }
 
+  getMetaDescription() {
+    const { addon, i18n } = this.props;
+
+    if (!addon) {
+      return null;
+    }
+
+    const content = i18n.sprintf(
+      i18n.gettext('Download %(addonName)s for Firefox. %(summary)s'),
+      {
+        addonName: addon.name,
+        summary: addon.summary,
+      },
+    );
+
+    return <meta name="description" content={content} />;
+  }
+
   render() {
     const {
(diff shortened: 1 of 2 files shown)


src/amo/pages/Addon/index.js
```diff
@@ -398,6 +398,24 @@ export class AddonBase extends React.Component {
...
+
+    if (!addon) {
+      return null;
+    }
+
+    const content = i18n.sprintf(
+      i18n.gettext('Download %(addonName)s for Firefox. %(summary)s'),
+      {
+        addonName: addon.name,
+        summary: addon.summary,
```
rebmullin: do we still want to factor in persona which has descriptions for now? see https://github.com/mozilla/addons-frontend/pull/6069#discussion_r214071501

willdurand: I think it is fine as is, because we are about to stop supporting lightweight themes.

PR #6489, opened 2018-10-02
Add more logging statements and remove CSS for HomeHeroBanner

### LINKED ISSUE #6477: Sometimes the hero banner jumps in size on first page load

* Clear all your cookies
* Load the prod home page
* Refresh the page a lot (10 or 20 times, nothing specific)

### What happened?

Sometimes, the hero banner jumps in size. It can go from small to big or big to small.

The hero banner should not jump in size.

This happened both when I was pinned to a small banner and a large banner with the A/B cookie

### LINKED ISSUE #6490: Remove `HomeHeroBanner` "small" variant and tracking code
We do not want to show the "small" variant of the `HomeHeroBanner`, which was part of an A/B experiment. This issue is about removing this variant, but not entirely yet since we want to debug a potential issue in #6477.

We can also remove tracking code.

willdurand:
Fixes #6477 
Fixes #6490

### DIFF
--- a/src/amo/components/HomeHeroBanner/index.js
+++ b/src/amo/components/HomeHeroBanner/index.js
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { compose } from 'redux';
 import makeClassName from 'classnames';
 
+import log from 'core/logger';
 import translate from 'core/i18n/translate';
 import Hero from 'ui/components/Hero';
 import HeroSection from 'ui/components/HeroSection';
@@ -33,8 +34,12 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
     const { _tracking, experimentEnabled, variant } = this.props;
 
     if (!experimentEnabled) {
+      log.info('[HomeHeroBanner.componentDidMount] experiment not enabled');
+
       return;
     }
+
+    log.info('[HomeHeroBanner.componentDidMount] variant is:', variant);
 
     _tracking.sendEvent({
       action: variant,
@@ -312,6 +317,8 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
   };
 
   render() {
+    log.info('[HomeHeroBanner.render] variant is:', this.props.variant);
+
     const homeBannerClass = makeClassName('HomeHeroBanner', {
       'HomeHeroBanner--small': this.props.variant === AB_HOME_HERO_VARIANT_A,
     });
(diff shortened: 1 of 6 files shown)


src/core/withExperiment.js
```diff
@@ -69,9 +69,24 @@ export const withExperiment = ({
 
       this.experimentCookie = _cookie.load(this.getCookieName());
 
+      log.info(
```
rebmullin: just a quick question: I've read that `this.props` and `props` are supposed to be the same in the constructor but I've ran into an issue where it was actually different so things didn't work as expected. Maybe that was a deeper issue but if I am remembering correctly it was a similar situation where it was ok locally but not on another env. I am wondering if we should log out the different props just to see those..?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -15,7 +16,6 @@ import './styles.scss';
 
 type InternalProps = {|
   ...WithExperimentInjectedProps,
-  _tracking: typeof tracking,
```
rebmullin: can remove the tracking import above to make linting happy :) ?

willdurand: d'oh!

src/core/withExperiment.js
willdurand: which props would you like to see here?

rebmullin: something like ~

`log.info(this.props)`;

AND

`log.info(props)`

just to make sure they are the same

willdurand: the problem is that `config` and `i18n` are part of each variable, and it is big!

rebmullin: oh right. So maybe just the props that we are using then ~

log.info(this.props._cookie);
log.info(this.props.variantA)
log.info(this.props.variantB)

log.info(props._cookie);
log.info(props.variantA);
log.info(props.variantB)

?

willdurand: added!

src/core/withExperiment.js
```diff
@@ -60,6 +60,18 @@ export const withExperiment = ({
     constructor(props: withExperimentInternalProps) {
       super(props);
 
+      log.info('[WithExperiment.constructor] props:', {
+        _cookie: props._cookie,
+        variantA: props.variantA,
+        variantB: props.variantB,
+      });
+
+      log.info('[WithExperiment.constructor] this.props:', {
```
kumar303: Since this exact component class doesn't use inheritance for anything, I don't see how this log statement would be helpful. It probably won't hurt, though.

Issue #6349, opened 2018-09-21
Enable the inline review flow in production


kumar303:
We have a new review writing / editing flow where the form appears inline as opposed to an overlay. See this for a screencast: https://github.com/mozilla/addons-frontend/pull/6231

Let's turn this on in production by enabling its config flag after these issues are closed:

* ~~https://github.com/mozilla/addons-frontend/issues/6300 : The review form should not be left open when switching pages~~
* ~~https://github.com/mozilla/addons-frontend/issues/6345 : Simplify the 'Rating saved' notification~~
* ~~https://github.com/mozilla/addons-frontend/issues/6346 : Minor tweaks to buttons and links in review flow~~
* ~~https://github.com/mozilla/addons-frontend/issues/6271 : The previous text is displayed after changing the star rating~~

These will be the next issues we follow up with after enabling the flow:
* https://github.com/mozilla/addons-frontend/issues/6347 : Simplify the confirmation screen for deleting a review / rating
* https://github.com/mozilla/addons-frontend/issues/5986 : The star breakdown doesn't always update after submitting a rating

pwalm:
Most recent mock up for reference:
![ratings and reviews flow v3](https://user-images.githubusercontent.com/6797635/45907297-44237880-bdac-11e8-852a-c1780b894b0e.png)

bobsilverberg:
Note that landing this should fix #6414 and #6417.

kumar303:
We decided to push this out early, before all issues are fixed. This is because there are some bugs in production related to all of the recent code changes and enabling the new flow will fix them.

We will enable the inline review flow on Thursday 2018-10-04.

PR #6498, opened 2018-10-03
Enable the inline review flow in production

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6349


kumar303:
Oh, duh. I forgot about all the tests that need the config flag turned off.

kumar303:
Oh yeah, I keep forgetting to file an issue about deleting the old code: https://github.com/mozilla/addons-frontend/issues/6500 🎉

kumar303:
I think this is good to go. The test failure is a locale problem that I'll fix separately.

Issue #6501, opened 2018-10-03
Invalid variable in zh_TW and sl locales


kumar303:
Introduced by https://github.com/mozilla/addons-frontend/commit/5d9d9925030779d70911d61938fb2ca28c4ea703 :

```
>>> Working on: /home/travis/build/mozilla/addons-frontend/locale/zh_TW/LC_MESSAGES/amo.po
E201: invalid variables: %s
446:#: src/amo/components/App/index.js:235
447:msgid "%(title)s – Add-ons for Android (%(locale)s)"
448:msgstr "%s - Android 版附加元件站（%(locale)s）"
```

PR #6502, opened 2018-10-03
Fix invalid variable in zh_TW and sl locales

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6501


kumar303:
The tox env is broken for me, too 🤢  I'll worry about that later.

PR #6506, opened 2018-10-03
11 October homepage content update

### LINKED ISSUE #6493: 11 October homepage content update
For the **11 October** push, please update homepage content as detailed [here](https://docs.google.com/document/d/1g4RuGKW__eTHNd277rKPMM46ygonXzW68jSs2OXEkPM/edit#). 

@muffinresearch

bobsilverberg:
Fixes #6493

### DIFF
--- a/src/amo/components/HomeHeroBanner/index.js
+++ b/src/amo/components/HomeHeroBanner/index.js
@@ -298,13 +298,6 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
         url: '/addon/groupspeeddial/',
       },
       {
-        title: i18n.gettext('Search Site'),
-        description: i18n.gettext(
-          'Search within just the domain you’re visiting',
-        ),
-        url: '/addon/search-site-we/',
-      },
-      {
         title: i18n.gettext('Undo Close Tab Button'),
         description: i18n.gettext('Never lose a page again'),
         url: '/addon/undo-close-tab-button/',
--- a/src/amo/components/HomeHeroBanner/index.js
+++ b/src/amo/components/HomeHeroBanner/index.js
@@ -271,6 +271,44 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
         description: i18n.gettext('Customized mouse gestures'),
         url: '/addon/foxy-gestures/',
       },
+      {
+        title: i18n.gettext('Dark Reader'),
+        description: i18n.gettext(
+          'Go dark for a better web viewing experience',
+        ),
+        url: '/addon/darkreader/',
+      },
+      {
+        title: i18n.gettext('CookieBro'),
+        description: i18n.gettext('Automatically delete unwanted cookies'),
+        url: '/addon/cookiebro/',
+      },
+      {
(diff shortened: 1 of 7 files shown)


src/amo/sagas/home.js
```diff
@@ -76,12 +76,12 @@ export function* fetchHomeAddons({
       page_size: LANDING_PAGE_THEME_COUNT,
     },
   };
-  const trendingExtensionsParams: SearchParams = {
+  const popularExtensionsParams: SearchParams = {
```
rebmullin: I wonder if there is any easier way to toggle this.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -271,6 +271,44 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
...
+      },
+      {
+        title: i18n.gettext('Group Speed Dial'),
+        description: i18n.gettext(
+          'Visual bookmarks for your favorite places on the web',
+        ),
+        url: '/addon/groupspeeddial/',
+      },
+      {
+        title: i18n.gettext('Search Site'),
```
rebmullin: it looks like we already have this one

src/amo/sagas/home.js
bobsilverberg: I agree that this is a pain, and there is probably a better way, but we keep saying that this is "temporary" and we'll have better tools to manage this stuff in the future, so we just persist with the manual "search and replace" approach for now. I have done a couple of things over the past year or so that have made incremental improvements, but we still have to make a lot of changes when we want to change one of the shelves.

src/amo/components/HomeHeroBanner/index.js
bobsilverberg: Good catch, thanks!

bobsilverberg:
> Shouldn’t "See more parental controls" be "See more parental controls extensions"?

I used "See more parental controls" because the extensions are parental controls. @devaneymoz what do you prefer for the link text?

bobsilverberg:
Thanks @devaneymoz. In that case no changes are needed.

TonnesM:
@bobsilverberg and @devaneymoz  Very well then.

I tend to agree about the thoughts above, but I’m not quite sure whether or not they’re entirely valid, and or at least for any language. These where my thoughts:

1) All similar strings describing theme and extension types look like "See more [type] extensions/themes".
2) "Parental controls" is a widely used term of itself and typically referred to as features/settings that I don’t think should be considered or findable as (types of) extensions or other types/pieces of software of their own, i.e. it could or maybe should be considered as a category.
3) In some cases (like for my locale) "Parental controls" is translated as e.g. "Parental guidance", partly because "controls" can be hard to translate in this context - German for instance uses "Schutz"(Protection). That also means plural forms may not be possible for certain locales so they need to add "extensions" anyway, if they understand the string at all. It could also lead to translation inconsistency when unnoticed - see Pontoon and Transvision for comparison for all locales.

Therefor my choice would be to add "extensions" here. Nevertheless, you can keep it as is, localizers may be able to find this PR and decide for themselves.

PR #6511, opened 2018-10-04
Expose JSON linked data on amo detail pages

### LINKED ISSUE #4129: Expose Hidden Metadata in Search Engine Results
When you search for an add-on on a search engine like Google, you can see the extension's average ratings from Google Play and the Chrome Store (if the add-on is available on multiple browsers), but not the average rating from AMO. An example is below: 

screen shot 2018-01-04 at 3 06 12 pm

Could we expose an extension's AMO rating on search engine results?

willdurand:
Fixes #4129

---

You can use this tool to verify the structured data: https://search.google.com/structured-data/testing-tool

~~I do not yet expose the number of downloads (as mentioned in #4129) because we don't know if we want to expose it. It can be added later in any case.~~

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -47,7 +47,7 @@ import {
 } from 'core/constants';
 import { withInstallHelpers } from 'core/installAddon';
 import { isTheme, nl2br, sanitizeHTML, sanitizeUserHTML } from 'core/utils';
-import { getErrorMessage } from 'core/utils/addons';
+import { getAddonJsonLinkedData, getErrorMessage } from 'core/utils/addons';
 import { getClientCompatibility as _getClientCompatibility } from 'core/utils/compatibility';
 import { getAddonIconUrl, getPreviewImage } from 'core/imageUtils';
 import translate from 'core/i18n/translate';
@@ -493,6 +493,16 @@ export class AddonBase extends React.Component {
     return tags;
   }
 
+  renderJsonLinkedData() {
+    const { addon } = this.props;
+
+    return (
+      <script type="application/ld+json">
+        {JSON.stringify(getAddonJsonLinkedData({ addon }))}
+      </script>
+    );
+  }
+
   render() {
     const {
       addon,
@@ -591,6 +601,7 @@ export class AddonBase extends React.Component {
             <link rel="canonical" href={addon.url} />
             <meta name="description" content={this.getPageDescription()} />
             {this.renderMetaOpenGraph()}
+            {this.renderJsonLinkedData()}
           </Helmet>
         )}
 
--- a/src/core/utils/addons.js
+++ b/src/core/utils/addons.js
@@ -10,6 +10,7 @@ import {
   INSTALL_FAILED,
 } from 'core/constants';
 import log from 'core/logger';
+import { getPreviewImage } from 'core/imageUtils';
 import type { AddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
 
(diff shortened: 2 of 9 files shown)


src/core/utils/addons.js
```diff
@@ -62,3 +63,51 @@ export const getFileHash = ({
...
+  if (ratings) {
+    aggregateRating = {
+      '@type': 'AggregateRating',
+      ratingCount: ratings.count,
+      ratingValue: ratings.average,
+    };
+  }
+
+  return removeUndefinedProps({
+    '@context': 'http://schema.org',
```
rebmullin: should this be: `https://schema.org/WebApplication` - in the same in the issue it has this? I see type has WebApplication tho so maybe this is okay..?

src/core/utils/addons.js
```diff
@@ -62,3 +63,51 @@ export const getFileHash = ({
...
+
+export const getAddonJsonLinkedData = ({
+  addon,
+}: {|
+  addon: AddonType,
+|}): Object | null => {
+  const { current_version: currentVersion, ratings } = addon;
+
+  let aggregateRating;
+  if (ratings) {
```
rebmullin: it wasn't clear to me in the issue if we should always be adding this or only if had a min amount of reviews (or rating too it mentioned).

tests/unit/core/components/TestServerHtml.js
```diff
@@ -85,7 +93,7 @@ describe(__filename, () => {
     const root = render();
     const js = root.find('script');
 
-    expect(js.at(1)).toHaveProp('src', '/foo/disco-blah.js');
+    expect(js.at(2)).toHaveProp('src', '/foo/disco-blah.js');
```
rebmullin: It's not clear to me why this has changed..?

tests/unit/core/components/TestServerHtml.js
```diff
@@ -112,8 +120,8 @@ describe(__filename, () => {
     const root = render();
     const js = root.find('script');
 
-    expect(js.at(1)).toHaveProp('integrity', 'sha512-disco-js');
-    expect(js.at(1)).toHaveProp('crossOrigin', 'anonymous');
+    expect(js.at(2)).toHaveProp('integrity', 'sha512-disco-js');
+    expect(js.at(2)).toHaveProp('crossOrigin', 'anonymous');
```
rebmullin: nit: same here..

src/core/utils/addons.js
```diff
@@ -62,3 +63,51 @@ export const getFileHash = ({
...
+    }
+  });
+  return newObject;
+}
+
+export const getAddonJsonLinkedData = ({
+  addon,
+}: {|
+  addon: AddonType,
+|}): Object | null => {
```
rebmullin: can we be any more specific than Object or could it be too variable to have something more specific than this?

src/amo/pages/Addon/index.js
```diff
@@ -467,6 +467,16 @@ export class AddonBase extends React.Component {
     }
   }
 
+  renderJsonLinkedData() {
+    const { addon } = this.props;
+
+    return (
+      <script type="application/ld+json">
+        {JSON.stringify(getAddonJsonLinkedData({ addon }))}
```
rebmullin: since you have this this could return null should we wrap this in a condition?

willdurand: Actually, it should not return `null`, good catch

src/core/utils/addons.js
willdurand: I don't think we can since it is not a fixed data structure.

src/core/utils/addons.js
willdurand: Yeah, it is not clear what to do. My understanding is that we should not round numbers and not hide/show numbers from time to time, i.e. always showing the ratings. That's what Google does.

src/core/utils/addons.js
willdurand: I think that was a mistake in the issue because this does not work. I validated the schema with the Google validator.

tests/unit/core/components/TestServerHtml.js
willdurand: There is now a script injected before those scripts. It is the script tag for the JSON-LD data.

tests/unit/core/components/TestServerHtml.js
willdurand: There is now a script injected before those scripts. It is the script tag for the JSON-LD data.

rebmullin:
via the tool, I see the following when I look at an addon without a rating:

Alt text

do you see the same? I am looking at redux and it seems that ratings won't always be null/undefined:

Alt text

willdurand:
ah, interesting. then I shoukd check count too

willdurand:
rebased and updated to take `count` into account

willdurand:
> even if there is no rating, we want to add this script? It seems like we do since maybe the other info is useful too but wanted to verify..

Yes, we need the other info in any case.

willdurand:
Thanks! Marking as blocked because I'd like @rraue's input about the low ratings.

PR #6520, opened 2018-10-04
Update the versions reducer to use the correct file

### LINKED ISSUE #6510: Update the versions reducer to use the correct file
This is a follow-up to #6473. The current code in `reducers/versions.js` just uses the first file for a given version to provide the `created` and `filesize` properties, but we should really find the correct file to use, using logic similar to that found in [`findInstallURL`](https://github.com/mozilla/addons-frontend/blob/2018.10.11/src/core/installAddon.js#L222). This will likely involve refactoring [`installAddon.js`](https://github.com/mozilla/addons-frontend/blob/2018.10.11/src/core/installAddon.js) and possibly other modules as well.

bobsilverberg:
Fixes #6510

### DIFF
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -1,10 +1,12 @@
 /* @flow */
 import invariant from 'invariant';
 
+import { createPlatformFiles } from 'core/reducers/addons';
+import { findFileForPlatform } from 'core/utils';
 import type {
   AddonCompatibilityType,
-  AddonFileType,
   ExternalAddonVersionType,
+  PlatformFilesType,
 } from 'core/types/addons';
 
 export const FETCH_VERSIONS: 'FETCH_VERSIONS' = 'FETCH_VERSIONS';
@@ -13,7 +15,7 @@ export const LOAD_VERSIONS: 'LOAD_VERSIONS' = 'LOAD_VERSIONS';
 export type AddonVersionType = {
   compatibility?: AddonCompatibilityType,
   created: string,
-  files: Array<AddonFileType>,
+  platformFiles: PlatformFilesType,
   filesize: number,
   id: number,
   license: { name: string, url: string },
@@ -24,18 +26,24 @@ export type AddonVersionType = {
 export const createInternalVersion = (
   version: ExternalAddonVersionType,
 ): AddonVersionType => {
-  // TODO: We should find the correct file, for now we'll just take the first one.
-  // Note that multiple files per version is going away, but not quite yet.
-  const file = version.files[0];
-  return {
+  const addonVersion = {
     compatibility: version.compatibility,
-    created: file.created,
-    files: version.files,
-    filesize: file.size,
+    platformFiles: createPlatformFiles(version),
     id: version.id,
     license: { name: version.license.name, url: version.license.url },
     releaseNotes: version.release_notes,
     version: version.version,
+  };
+
+  const file = findFileForPlatform({
+    platformFiles: addonVersion.platformFiles,
+    userAgentInfo: state.api.userAgentInfo, // this won't work currently
+  });
+
+  return {
+    ...addonVersion,
+    created: file.created,
+    filesize: file.size,
   };
 };
 
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -2,8 +2,8 @@
 import invariant from 'invariant';
 
 import { createPlatformFiles } from 'core/reducers/addons';
-import type { UserAgentInfoType } from 'core/reducers/api';
 import { findFileForPlatform } from 'core/utils';
+import type { UserAgentInfoType } from 'core/reducers/api';
 import type {
   AddonCompatibilityType,
   ExternalAddonVersionType,
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -132,16 +132,19 @@ export const getVersionInfo = ({
   _findFileForPlatform = findFileForPlatform,
   userAgentInfo,
   version,
-}: GetVersionInfoParams): VersionInfoType => {
+}: GetVersionInfoParams): VersionInfoType | null => {
   const file = _findFileForPlatform({
     platformFiles: version.platformFiles,
     userAgentInfo,
   });
 
-  return {
-    created: file ? file.created : undefined,
-    filesize: file ? file.size : undefined,
-  };
+  if (file) {
+    return {
+      created: file.created,
+      filesize: file.size,
+    };
+  }
+  return null;
 };
 
 type Action = FetchVersionsAction | LoadVersionsAction;
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -2,6 +2,7 @@
 import invariant from 'invariant';
 
 import { createPlatformFiles } from 'core/reducers/addons';
+import type { UserAgentInfoType } from 'core/reducers/api';
 import { findFileForPlatform } from 'core/utils';
 import type {
   AddonCompatibilityType,
@@ -14,9 +15,7 @@ export const LOAD_VERSIONS: 'LOAD_VERSIONS' = 'LOAD_VERSIONS';
 
 export type AddonVersionType = {
   compatibility?: AddonCompatibilityType,
-  created: string,
   platformFiles: PlatformFilesType,
-  filesize: number,
   id: number,
   license: { name: string, url: string },
   releaseNotes?: string,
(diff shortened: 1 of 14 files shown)


src/amo/reducers/versions.js
```diff
@@ -24,19 +26,25 @@ export type AddonVersionType = {
...
+    platformFiles: createPlatformFiles(version),
     id: version.id,
     license: { name: version.license.name, url: version.license.url },
     releaseNotes: version.release_notes,
     version: version.version,
   };
+
+  const file = findFileForPlatform({
+    platformFiles: addonVersion.platformFiles,
+    userAgentInfo: state.api.userAgentInfo, // this won't work currently
```
bobsilverberg: My idea was that when we create the internal version, it would make sense to do this lookup and store the info we need from the correct `file` object. The issue (or one issue) is that I don't have access to `userAgentInfo` at this point.

Instead of doing this I could just leave `created` and `filesize` out of the `AddonVersionType` and calculate them on the fly, each time I need to render a version, but that seems like processing that's worth avoiding.

At this point I see a couple of options:

1. Do it like I've done it above, and pass `userAgentInfo` in the action payload which would result in the call to `createInternalVersion` (which will likely come from a saga).

2. Leave these properties out and create a function/functions that will return the `create` and `filesize`. I guess that function would still live in this reducer? I'd then have to call it inside a loop when I am rendering the versions.

@kumar303 what do you think?

kumar303: > 1. pass `userAgentInfo` in the action payload
> 2. call a selector function to get `created` and `filesize` when rendering the versions.

Yes, I think these are the two best options. I am torn but I guess either would be fine. 

I think there would be a negligible performance penalty if you call a selector function when rendering. Premature optimization is usually not a great idea. 

As for passing `userAgentInfo` in the action payload, this has the downside in that the app will no longer react to state changes affecting `userAgentInfo`. In practice that's probably fine because `userAgentInfo` never changes and I can't think of any reason why it would in the future.

I don't know if this helps. I think you could go in either direction 🤷‍♂️

tests/unit/disco/components/TestAddon.js
```diff
@@ -466,9 +466,6 @@ describe(__filename, () => {
       const reason = 'WHATEVER';
 
       const root = render({
-        addonProps: {
```
bobsilverberg: Note: This was required to avoid an error. I could have ensured that the `current_version` used had a `files` property, but I don't think this has any impact on the test. From what I can tell it's just the `compatible: false` that controls the outcome of this test, so I believe it is fine to just remove these lines.

bobsilverberg:
This is no longer a WIP and is ready for a review.

bobsilverberg:
The build is failing for this PR with the following message:

```
/home/travis/build/mozilla/addons-frontend/node_modules/babel-register/node_modules/babel-core/lib/transformation/file/index.js:590
      throw err;
      ^
SyntaxError: /home/travis/build/mozilla/addons-frontend/src/amo/img/icons/default-64.png: Unexpected character '�' (1:0)
> 1 | �PNG
    | ^
  2 | 
  3 | 
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mozilla-addons-frontend@0.0.1 build-all: `npm run clean && NODE_APP_INSTANCE=amo better-npm-run build && NODE_APP_INSTANCE=disco better-npm-run build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the mozilla-addons-frontend@0.0.1 build-all script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2018-10-09T16_11_20_146Z-debug.log
```

and I have no idea why. This patch hasn't touched that file (/src/amo/img/icons/default-64.png).

I am also seeing this locally. Perhaps the error message is a red herring, but I'm not sure what else to check. Can anyone help?

bobsilverberg:
One final note: There are few more tests that could be improved by using a mock of `findInstallURL` as I have done for some of the tests in this patch. 

Specifically, the tests for `setCurrentStatus` and `install` in test_installAddon.js, as well as tests in `TestAMInstallButton.js` and `TestInstallButton.js`. I did not do this because this patch is big enough as it is, and the changes would be fairly large. The tests are fine as they are, but could be "improved". I could open a code quality issue to update them at some point if we think that would be valuable.

bobsilverberg:
In case anyone is actively following along, I may have found the cause of the error, which is totally unrelated to that file.

src/amo/reducers/versions.js
```diff
@@ -121,6 +117,33 @@ export const getLoadingBySlug = ({ slug, state }: GetBySlugParams): boolean => {
...
+  userAgentInfo,
+  version,
+}: GetVersionInfoParams): VersionInfoType => {
+  const file = _findFileForPlatform({
+    platformFiles: version.platformFiles,
+    userAgentInfo,
+  });
+
+  return {
+    created: file ? file.created : undefined,
```
kumar303: What is this function going to be used for? I guess it's for an upcoming patch?

I'm concerned about when `file` is falsy. I think it should not create an object with empty properties for that case because it could be confusing. Can it return `null` in that case instead of an object with empty properties?

src/core/installAddon.js
```diff
@@ -221,48 +163,33 @@ type FindInstallUrlParams = {|
...
-    installURL = platformFiles[OS_ALL].url;
-  }
+  const installURL = platformFile && platformFile.url;
 
   if (!installURL) {
     // This could happen for themes which do not have version files.
     log.debug(
-      oneLine`No install URL exists for platform "${agentOsName}"
-      (mapped to "${platform}"); platform files:`,
+      oneLine` No file exists for os "${userAgentInfo.os.toString()}"; platform files:`,
```
kumar303: Any reason why there's a space at the beginning? It doesn't really matter, just wondering.

src/core/types/addons.js
```diff
@@ -156,6 +156,17 @@ export type ExternalAddonType = {|
   weekly_downloads: number,
 |};
 
+export type PlatformFilesType = {|
+  // This seems necessary to help Flow know that computed
+  // keys always return an AddonFileType.
+  [anyPlatform: string]: ?AddonFileType,
```
kumar303: It looks like you can delete this now. I think it was only happening before because of how the platform files object was created by assignment in a loop.

tests/unit/amo/reducers/test_versions.js
```diff
@@ -61,6 +65,22 @@ describe(__filename, () => {
     ]);
   });
 
+  describe(createInternalVersion, () => {
```
kumar303: I think you meant this as a string 👀

tests/unit/amo/reducers/test_versions.js
```diff
@@ -61,6 +65,22 @@ describe(__filename, () => {
...
+          name: fakeVersion.license.name,
+          url: fakeVersion.license.url,
+        },
+        releaseNotes: fakeVersion.release_notes,
+        version: fakeVersion.version,
+      });
+    });
+  });
+
   describe(getLoadingBySlug, () => {
```
kumar303: Hmm. I guess this file wasn't using string specs? I think it should be a string.

tests/unit/amo/reducers/test_versions.js
```diff
@@ -74,4 +94,32 @@ describe(__filename, () => {
       expect(getVersionsBySlug({ slug: 'some-slug', state })).toBe(null);
     });
   });
+
+  describe(getVersionInfo, () => {
```
kumar303: This spec should be a string, too.

tests/unit/core/reducers/test_addons.js
```diff
@@ -789,4 +781,38 @@ describe(__filename, () => {
...
+      };
+      expect(
+        createPlatformFiles({
+          ...fakeVersion,
+          files: [windowsFile, unknownFile],
+        }),
+      ).toEqual({
+        ...defaultPlatformFiles,
+        [OS_WINDOWS]: windowsFile,
+        [unknownPlatform]: unknownFile,
```
kumar303: This is a special case and I think it warrants its own test. Also, you can delete [this one](https://github.com/mozilla/addons-frontend/blob/159016d1a7bbbad8a997ea978d3accce1f6b9095/tests/unit/core/reducers/test_addons.js#L263) because it makes more sense as a unit test of `createPlatformFiles` now.

src/amo/reducers/versions.js
bobsilverberg: Yes, it will be used in an upcoming patch. It's the selector that will be used when rendering a version to retrieve the info needed from the correct file. You're right, it would be better to just return null when `file` is falsey, which I actually realized when working on the next patch. I'll change this.

src/core/installAddon.js
bobsilverberg: No reason. I'm not sure how that got there, but it's a mistake. I'll fix it.

src/core/reducers/addons.js
```diff
@@ -166,14 +204,7 @@ export function createInternalAddon(apiAddon: ExternalAddonType): AddonType {
...
     // These are custom properties not in the API response.
-
-    platformFiles: {
-      all: undefined,
-      android: undefined,
-      linux: undefined,
-      mac: undefined,
-      windows: undefined,
-    },
+    platformFiles: createPlatformFiles(apiAddon.current_version),
```
kumar303: How about `currentPlatformFiles` ?

However, I don't totally understand why this is needed. Can't the `versions` reducer respond to the same load action and store a version object like any other version object? After that, we could add `addon.currentVersionId` as a link to the version object.

bobsilverberg: I hadn't considered it when writing this patch, but now I'm working on https://github.com/mozilla/addons-frontend/issues/6634 and it's become obvious that something needs to change. Using an id is an interesting idea I hadn't considered. So we'd store the "real" version info in the versions state and just store a link to it in the addon? It feels a bit weird to reference an item in a different reducer's state, but I guess that's ok. Is that what you're suggesting?

Issue #6529, opened 2018-10-05
Extract the confirmation dialog from ConfirmButton


kumar303:
For https://github.com/mozilla/addons-frontend/issues/6347 we need direct control over the confirmation dialog.

This is to extract the dialog panel from `ConfirmButton` into a new component. The `ConfirmButton` component should look and behave exactly as before.

**For QA**: please test for regressions in the following scenarios:
* Deleting a rating
* Deleting a review
* Deleting a developer response
* Deleting a user profile picture
* Deleting a collection

PR #6530, opened 2018-10-05
Extract confirmation dialog from ConfirmButton

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6529

### DIFF
--- a/tests/unit/ui/components/TestConfirmButton.js
+++ b/tests/unit/ui/components/TestConfirmButton.js
@@ -5,7 +5,7 @@ import ConfirmButton, {
   ConfirmButtonBase,
   extractId,
 } from 'ui/components/ConfirmButton';
-import ConfirmButtonDialog from 'ui/components/ConfirmButtonDialog';
+import ConfirmationDialog from 'ui/components/ConfirmationDialog';
 import {
   applyUIStateChanges,
   createFakeEvent,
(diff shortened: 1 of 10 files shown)


tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -99,121 +129,33 @@ describe(__filename, () => {
...
+    onCancel(createFakeEvent());
 
-    root
-      .find('.ConfirmButton-cancel-button')
-      .simulate('click', createFakeEvent());
     applyUIStateChanges({ root, store });
-
     expect(root).not.toHaveClassName('ConfirmButton--show-confirmation');
-    sinon.assert.notCalled(onConfirmSpy);
+    expect(root.find(ConfirmationDialog)).toHaveLength(0);
```
rebmullin: should we show above that the dialog existed prior to clicking on Cancel:
 `expect(root.find(ConfirmationDialog)).toHaveLength(1);`

tests/unit/ui/components/TestConfirmButton.js
```diff
@@ -99,121 +129,33 @@ describe(__filename, () => {
...
-    applyUIStateChanges({ root, store });
-
-    expect(root).toHaveClassName('ConfirmButton--show-confirmation');
-    sinon.assert.notCalled(onConfirmSpy);
+    const onConfirm = getDialogProp(root, 'onConfirm');
 
+    // Simulate a confirmation.
     const event = createFakeEvent();
-    root.find('.ConfirmButton-confirm-button').simulate('click', event);
+    onConfirm(event);
```
rebmullin: same here

tests/unit/ui/components/TestConfirmButton.js
kumar303: Sure, I can add that. It's a kind of a duplication of other test code but since we don't have Flow coverage in tests, mistakes can happen 🤷‍♀️

kumar303:
Thanks for the quick review

PR #6542, opened 2018-10-08
Remove the static theme flag

### LINKED ISSUE #5802: Remove the static theme flag after has been verified in PROD

This should come after #5799 and it has been verified as good in prod. We should remove this enable flag altogether

willdurand:
Fixes #5802

### DIFF
--- a/tests/unit/amo/pages/TestCategory.js
+++ b/tests/unit/amo/pages/TestCategory.js
@@ -23,6 +23,7 @@ import ErrorList from 'ui/components/ErrorList';
 import {
   createStubErrorHandler,
   fakeI18n,
+  getFakeConfig,
   shallowUntilTarget,
 } from 'tests/unit/helpers';
 import {
@@ -434,7 +435,7 @@ describe(__filename, () => {
     });
   });
 
-  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter', () => {
+  it('sets the correct footer links for themes', () => {
     _categoriesFetch();
     _categoriesLoad();
     _getLanding();
--- a/tests/unit/amo/pages/TestCategory.js
+++ b/tests/unit/amo/pages/TestCategory.js
@@ -23,7 +23,6 @@ import ErrorList from 'ui/components/ErrorList';
 import {
   createStubErrorHandler,
   fakeI18n,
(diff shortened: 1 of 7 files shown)


tests/unit/amo/pages/TestLandingPage.js
```diff
@@ -317,17 +316,15 @@ describe(__filename, () => {
     });
   });
 
-  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
-    const fakeConfig = getFakeConfig({ enableFeatureStaticThemes: true });
-
+  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter', () => {
```
rebmullin: Before I believe we added this test just b/c of the config now I am wondering if we need this..? See [PR/commit](https://github.com/mozilla/addons-frontend/commit/cb33d11172242d4a1146c79edad9a6e2a5e2bfe1) that added this. There's nothing for extension so I am just thinking..

if we do keep it I would adjust the name slightly and should we add to the end ~ "...if it's a theme type" and consider adding an extensions test too b/c if we have this for theme should we add something for extensions? I feel like that's what we usually do to show the different types, right?

tests/unit/amo/pages/TestCategory.js
```diff
@@ -434,18 +433,13 @@ describe(__filename, () => {
     });
   });
 
-  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter if static theme is enabled', () => {
+  it('renders the themes shelves with the ADDON_TYPE_THEMES_FILTER filter', () => {
```
rebmullin: Before I believe we added this test just b/c of the config now I am wondering if we need this..? See [PR/commit](https://github.com/mozilla/addons-frontend/commit/cb33d11172242d4a1146c79edad9a6e2a5e2bfe1#diff-6077ad35fee63ec7e99470d3e202d584R441) that added this. 

if we do keep it I would adjust the name slightly and should we add to the end ~ "...if it's a theme type" and consider adding an extensions test too b/c if we have this for theme should we add something for extensions? I feel like that's what we usually do to show the different types, right?

note: I have similar comment below

PR #6545, opened 2018-10-08
Remove normalizr

### LINKED ISSUE #2917: Create explicit API data mappings (i.e. replace normalizr) for AMO

willdurand:
Fixes #2917

---

Remove all the things!


kumar303:
D'oh. Helpers-magedden is upon us.

willdurand:
Thanks @kumar303, it is fixed now!

src/core/api/index.js
```diff
@@ -227,8 +220,7 @@ export function callApi({
         }
         throw fetchError;
       },
-    )
-    .then((response) => (schema ? normalize(response, schema) : response));
```
kumar303: 🌤

PR #6547, opened 2018-10-08
I6538 remove ab exp/logs from homepg

### LINKED ISSUE #6538: Remove withExperiment HOC from the HomeHeroBanner component + extra log calls
We should remove the `withExperiment` HOC from the `HomeHeroBanner` component as we do not run this experiment anymore. Extra log calls added in https://github.com/mozilla/addons-frontend/pull/6489 should be removed too, and we should also remove the config parameter.

rebmullin:
fixes #6538

### DIFF
--- a/src/amo/components/HomeHeroBanner/index.js
+++ b/src/amo/components/HomeHeroBanner/index.js
@@ -1,7 +1,6 @@
 /* @flow */
 import * as React from 'react';
 import { compose } from 'redux';
-import makeClassName from 'classnames';
 
 import { INSTALL_SOURCE_HERO_PROMO } from 'core/constants';
 import log from 'core/logger';
@@ -9,36 +8,16 @@ import translate from 'core/i18n/translate';
 import { addQueryParams } from 'core/utils';
 import Hero from 'ui/components/Hero';
 import HeroSection from 'ui/components/HeroSection';
-import { withExperiment } from 'core/withExperiment';
 import type { I18nType } from 'core/types/i18n';
-import type { WithExperimentInjectedProps } from 'core/withExperiment';
 import type { HeroSectionsType } from 'ui/components/Hero';
 
 import './styles.scss';
 
 type InternalProps = {|
-  ...WithExperimentInjectedProps,
   i18n: I18nType,
 |};
 
-export const AB_HOME_HERO_EXPERIMENT = 'homeHero';
-export const AB_HOME_HERO_VARIANT_A = 'small';
-export const AB_HOME_HERO_VARIANT_B = 'large';
-export const AB_HOME_HERO_EXPERIMENT_CATEGORY = 'AMO Home Hero Experiment';
-
 export class HomeHeroBannerBase extends React.Component<InternalProps> {
-  componentDidMount() {
-    const { experimentEnabled, variant } = this.props;
-
-    if (!experimentEnabled) {
-      log.info('[HomeHeroBanner.componentDidMount] experiment not enabled');
-
-      return;
-    }
-
-    log.info('[HomeHeroBanner.componentDidMount] variant is:', variant);
-  }
-
   getHeroes() {
     const { i18n } = this.props;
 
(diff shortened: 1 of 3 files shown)


willdurand:
Note: we are not ready to merge this PR since we did not find the root cause of the HomeHeroBanner issue.

rebmullin:
hey @willdurand, what's the status on this one? Should I still keep this one open?

willdurand:
I don't find anything in Kibana, so I guess the problem is somewhere else (React reconciliation?). I don't know what to do now since it was impossible to reproduce the bug in any env but -prod, and logs don't contain any weird behaviors...

@kumar303 what do you think? Do you want to take a closer look? Or should we remove the code (this PR) and see if the issue happens again during the next a/b experiment?

willdurand:
@rebmullin we found the root cause of the issue. We need to fix it first and make sure it has been fixed entirely. So I think either you close this PR right now and re-open it later (after next week I believe), or you keep it as-is (and rebase regularly).

willdurand:
> It would be hard to use our logs to determine that the `react-cookie` upgrade (#6803) fixes this bug.

Can't we make sure we won't see any logs with `cookie saved` twice?

kumar303:
Yes, we could but I don't know how to search for that in kibana. I only found the dupes by scanning the first page of results with my eyes and I wouldn't want to do that across many pages.

willdurand:
> Yes, we could but I don't know how to search for that in kibana. I only found the dupes by scanning the first page of results with my eyes and I wouldn't want to do that across many pages.

can we try that during the next tag?

rebmullin:
hey @willdurand, this has been updated. Can you please see how it looks now?

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -314,27 +292,16 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
...
 
-const HomeHeroBanner: React.ComponentType<InternalProps> = compose(
-  translate(),
-  withExperiment({
-    id: AB_HOME_HERO_EXPERIMENT,
-    variantA: AB_HOME_HERO_VARIANT_A,
-    variantB: AB_HOME_HERO_VARIANT_B,
-  }),
-)(HomeHeroBannerBase);
+const HomeHeroBanner: React.ComponentType<InternalProps> = compose(translate())(
```
willdurand: You can get rid of `compose()` too and write: `... = translate()(HomeHeroBannerBase);`

src/core/withExperiment.js
```diff
@@ -59,43 +59,18 @@ export const withExperiment = ({
...
-      });
-
-      log.info('[WithExperiment.constructor] this.props:', {
-        variantA: this.props.variantA,
-        variantB: this.props.variantB,
-      });
-
       if (!this.isExperimentEnabled()) {
-        log.debug(`Experiment "${props.id}" is not enabled by config.`);
+        log.debug(`Experiment "${defaultId}" is not enabled by config.`);
```
willdurand: I believe this should be `props.id`

tests/unit/amo/components/TestHomeHeroBanner.js
```diff
@@ -2,35 +2,18 @@ import * as React from 'react';
...
   fakeI18n,
-  getFakeConfig,
   shallowUntilTarget,
 } from 'tests/unit/helpers';
 
-// Skip `withCookies` HOC since Enzyme does not support the React Context API.
-// See: https://github.com/mozilla/addons-frontend/issues/6839
-jest.mock('react-cookie', () => ({
-  withCookies: (component) => component,
-}));
```
willdurand: 🏆

src/amo/components/HomeHeroBanner/index.js
rebmullin: oh? So this is only used when using connect/redux?

src/core/withExperiment.js
rebmullin: oops! I didn't mean to change this.. thanks!

src/amo/components/HomeHeroBanner/index.js
willdurand: this is only useful when you have more than one HOC

PR #6549, opened 2018-10-08
I5737 updating page params type

### LINKED ISSUE #5737: The query parameter values should be string values, not numbers, everywhere
The `page` parameter used for API calls is not always considered a string value, leading to inconsistencies and issues with Flow. It applies to ALL query parameters.

See also:

- https://github.com/mozilla/addons-frontend/blob/master/src/amo/api/reviews.js#L143
- https://github.com/mozilla/addons-frontend/pull/5652/files/d7e6609988c0f2a079e7e88d36b086d7c56c79d2#r205250173
- https://github.com/mozilla/addons-frontend/pull/5652#discussion_r205260797

rebmullin:
fixes #5737

this was a little bigger than I expected :/

would it be better to somehow split this?


rebmullin:
> Are there any other parameters that are number and should be string values?

I missed the links so I will look into those too. But I do have a question..should all these params that are numbers here be strings, for eg:

the issue links to this:
https://github.com/mozilla/addons-frontend/blob/master/src/amo/api/reviews.js#L146

what about these ~:
https://github.com/mozilla/addons-frontend/blob/master/src/core/api/search.js#L15

I guess initially I thought it was just the q values but I wanted to clarify.

Also, I started looking into these a bit and it was a lot of updating which is fine but I am just wondering if it would be ok to split those into new issues/PRs..?

rebmullin:
hey @willdurand, so I split up the issue as we discussed above. can you please let me know how this looks now (for just updating the page value now)?

PR #6556, opened 2018-10-09
adding developer policies to footer

### LINKED ISSUE #6548: Add "Developer Policies" to footer
With input from legal and Mike Conca, we should add the Developer Policies (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/AMO/Policy) to the footer of the AMO front-end site. 

When we launch a new Developer Hub landing page in November (formerly known as the "developer launchpad"), we want to clearly point devs to the policies in the footer, which employs the front-end footer (rather than current Dev Hub footer).

See the attached screenshot for illustration of the requested change.

developer policies

championshuttler:
Fixes #6548

### DIFF
--- a/bin/loading-page.html
+++ b/bin/loading-page.html
@@ -67,6 +67,10 @@
                   <!-- /react-text -->
                 </a>
               </li>
+              <li>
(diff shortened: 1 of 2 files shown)


bobsilverberg:
All this did was add some static text, that is always displayed, so I didn't think it warranted a test, tbh.

PR #6557, opened 2018-10-09
Redirect slug

### LINKED ISSUE #4271: Mixed case version of slug leads to blank page
Re-filed copy of https://github.com/mozilla/addons/issues/586

(Please include as many details as possible.)

Go to an addon location (https://addons.mozilla.org/en-US/firefox/addon/snaplinksplus/), but change the case of the name (https://addons.mozilla.org/en-US/firefox/addon/snapLinksplus/) and you don't get an error page, just a blank screen.

### What happened?

You don't get an error page, just a blank screen.

Get an error page, like with (https://addons.mozilla.org/en-US/firefox/addon/sadmknmsadlksa/)

Working: https://addons.mozilla.org/en-US/firefox/addon/snaplinksplus/
Working: https://addons.mozilla.org/en-US/firefox/addon/sadmknmsadlksa/
Not working: https://addons.mozilla.org/en-US/firefox/addon/snapLinksplus/

Looks like this:

add-ons_for_firefox

As noted by @diox in https://github.com/mozilla/addons/issues/586#issuecomment-357675409 this worked on the old site. The only issue is the legacy site shows the same content on a different URL (I'm not sure if it uses rel=canonical) - the ideal solution is probably to provide a 301 cacheable redirect to the lower case version.

willdurand:
Fixes #4271

---

Send a redirect when an add-on detail slug is not lower case OR a collection username/slug is not lower case.

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -125,9 +125,10 @@ export class AddonBase extends React.Component {
       const { slug } = params;
 
       if (addon) {
-        // We want to make sure the slug converts to a positive
-        // number/ID before we try redirecting.
-        if (slugIsPositiveID(slug)) {
+        // We want to make sure the slug converts to a positive number/ID
+        // before we try redirecting. We also want to make sure the case of the
+        // slug parameter matches the add-on's slug case.
+        if (slugIsPositiveID(slug) || addon.slug !== slug) {
           // We only load add-ons by slug, but ID must be supported too because
           // it is a legacy behavior.
           dispatch(
--- a/src/amo/pages/Collection/index.js
+++ b/src/amo/pages/Collection/index.js
@@ -37,6 +37,7 @@ import {
 import { withFixedErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import log from 'core/logger';
+import { sendServerRedirect } from 'core/reducers/redirectTo';
 import Card from 'ui/components/Card';
 import ConfirmButton from 'ui/components/ConfirmButton';
 import type {
@@ -69,6 +70,7 @@ type InternalProps = {|
   ...Props,
   _config: typeof config,
   _isFeaturedCollection: typeof isFeaturedCollection,
+  clientApp: string,
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   filters: CollectionFilters,
@@ -225,11 +227,31 @@ export class CollectionBase extends React.Component<InternalProps> {
 
     if (
       collection &&
-      (collection.slug !== params.slug ||
+      (collection.slug.toLowerCase() !== params.slug.toLowerCase() ||
         collection.authorUsername.toLowerCase() !==
           params.username.toLowerCase())
     ) {
       collectionChanged = true;
+    }
+
+    // See: https://github.com/mozilla/addons-frontend/issues/4271
+    if (
+      !collectionChanged &&
+      collection &&
+      (params.username !== collection.authorUsername ||
+        params.slug !== collection.slug)
+    ) {
+      const { lang, clientApp } = this.props;
+
+      this.props.dispatch(
+        sendServerRedirect({
+          status: 301,
+          url: `/${lang}/${clientApp}/collections/${
+            collection.authorUsername
+          }/${collection.slug}/`,
+        }),
+      );
+      return;
     }
 
     if (!collection || collectionChanged) {
@@ -510,7 +532,9 @@ export class CollectionBase extends React.Component<InternalProps> {
 }
 
 export const mapStateToProps = (state: AppState, ownProps: InternalProps) => {
-  const { loading } = state.collections.current;
+  const { api, collections } = state;
+
+  const { loading } = collections.current;
   const { creating, location } = ownProps;
 
   const filters = {
@@ -526,10 +550,12 @@ export const mapStateToProps = (state: AppState, ownProps: InternalProps) => {
     collection && currentUser && collection.authorId === currentUser.id;
 
   return {
+    clientApp: api.clientApp,
     collection,
     filters,
     isLoggedIn: !!currentUser,
     isOwner,
+    lang: api.lang,
     loading,
   };
 };
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -343,7 +343,7 @@ export default function addonsReducer(
         ...state,
         loadingBySlug: {
           ...state.loadingBySlug,
-          [slug]: true,
+          [slug.toLowerCase()]: true,
         },
       };
     }
--- a/src/core/reducers/addons.js
+++ b/src/core/reducers/addons.js
@@ -293,7 +293,11 @@ export const getAddonBySlug = (
   state: AppState,
   slug: string,
 ): AddonType | null => {
-  const addonId = state.addons.bySlug[slug];
+  if (typeof slug !== 'string') {
+    return null;
+  }
+
+  const addonId = state.addons.bySlug[slug.toLowerCase()];
 
   return getAddonByID(state, addonId);
 };
@@ -308,7 +312,11 @@ export const getAddonByGUID = (
 };
 
 export const isAddonLoading = (state: AppState, slug: string): boolean => {
-  return Boolean(state.addons.loadingBySlug[slug]);
+  if (typeof slug !== 'string') {
+    return false;
+  }
+
+  return Boolean(state.addons.loadingBySlug[slug.toLowerCase()]);
 };
 
 export const getAllAddons = (state: AppState): Array<AddonType> => {
@@ -355,8 +363,8 @@ export default function addonsReducer(
         byID[`${addon.id}`] = addon;
 
         if (addon.slug) {
-          bySlug[addon.slug] = addon.id;
-          loadingBySlug[addon.slug] = false;
+          bySlug[addon.slug.toLowerCase()] = addon.id;
+          loadingBySlug[addon.slug.toLowerCase()] = false;
         }
 
         if (addon.guid) {
@@ -390,11 +398,11 @@ export default function addonsReducer(
           },
           bySlug: {
             ...state.bySlug,
-            [addon.slug]: undefined,
+            [addon.slug.toLowerCase()]: undefined,
           },
           loadingBySlug: {
             ...state.loadingBySlug,
-            [addon.slug]: false,
+            [addon.slug.toLowerCase()]: undefined,
           },
         };
       }
--- a/tests/unit/amo/pages/TestAddon.js
+++ b/tests/unit/amo/pages/TestAddon.js
@@ -620,6 +620,30 @@ describe(__filename, () => {
     sinon.assert.callCount(fakeDispatch, 1);
   });
 
+  // See: https://github.com/mozilla/addons-frontend/issues/4271.
+  it("dispatches a server redirect when slug param case does not match the add-on's slug case", () => {
+    const clientApp = CLIENT_APP_FIREFOX;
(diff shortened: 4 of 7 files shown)


src/amo/pages/Addon/index.js
```diff
@@ -125,11 +131,12 @@ export class AddonBase extends React.Component {
       const { slug } = params;
 
       if (addon) {
-        // We want to make sure the slug converts to a positive
-        // number/ID before we try redirecting.
-        if (slugIsPositiveID(slug)) {
+        // We want to make sure the slug converts to a positive number/ID
+        // before we try redirecting.
+        if (slugIsPositiveID(slug) || isNotLowerCase(slug)) {
```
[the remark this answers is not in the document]
willdurand: it must always be lowercase and it must be enforced now, if the API does not lowercase the slug in the response, it's an API bug.

src/amo/pages/Addon/index.js
```diff
@@ -743,7 +750,10 @@ export class AddonBase extends React.Component {
 }
 
 export function mapStateToProps(state, ownProps) {
-  const { slug } = ownProps.match.params;
+  const slug =
+    typeof ownProps.match.params.slug === 'string'
```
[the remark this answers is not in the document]
willdurand: given that we retrieve the slug based on the routing parameter, we must use the lowercase version, because the only slug valid that should be stored is lowercase.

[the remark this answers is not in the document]
willdurand: I think it is because the test case fails if you remove this line.

src/amo/pages/Collection/index.js
```diff
@@ -223,6 +226,24 @@ export class CollectionBase extends React.Component<InternalProps> {
       }
     }
 
+    // See: https://github.com/mozilla/addons-frontend/issues/4271
+    if (
+      collection &&
+      (isNotLowerCase(params.username) || isNotLowerCase(params.slug))
```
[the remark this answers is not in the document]
willdurand: right, let's ignore this for the username. that's clearly a mistake.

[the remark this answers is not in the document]
willdurand: we'll get rid of usernames soon anyway, user ids won't have this problem in the near future.

src/amo/pages/Addon/index.js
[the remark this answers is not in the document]
bobsilverberg: I'm not sure I follow. Try testing with http://localhost:3000/en-CA/firefox/addon/Beach_View/.

src/amo/pages/Addon/index.js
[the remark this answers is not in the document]
bobsilverberg: Ah, I see. I think I better understand how the two changes in `Addon` interact now.

willdurand:
I am marking this PR as blocked because I thought the API was already ready, but it is not the case: the API must return lowercase slugs all the time.

willdurand:
Ok, so the original idea supported by Product was to lowercase slugs all the time, but @eviljeff made a case to support mixed cases since we have done that in the past.

I thought about it last night, and I think we can start by making mixed case work first, and see to lowercase slugs in the near future.

src/core/reducers/addons.js
```diff
@@ -390,11 +398,11 @@ export default function addonsReducer(
           },
           bySlug: {
             ...state.bySlug,
-            [addon.slug]: undefined,
+            [addon.slug.toLowerCase()]: undefined,
           },
           loadingBySlug: {
             ...state.loadingBySlug,
-            [addon.slug]: false,
+            [addon.slug.toLowerCase()]: undefined,
```
willdurand: I think it makes sense to remove it from the state, rather than keeping it.

src/core/reducers/addons.js
```diff
@@ -355,8 +363,8 @@ export default function addonsReducer(
         byID[`${addon.id}`] = addon;
 
         if (addon.slug) {
-          bySlug[addon.slug] = addon.id;
-          loadingBySlug[addon.slug] = false;
+          bySlug[addon.slug.toLowerCase()] = addon.id;
```
bobsilverberg: Should you also be using `slug.toLowerCase()` in `FETCH_ADDON` (line 346)?

tests/unit/amo/pages/TestAddon.js
```diff
@@ -620,6 +620,30 @@ describe(__filename, () => {
...
+  it("dispatches a server redirect when slug param case does not match the add-on's slug case", () => {
+    const clientApp = CLIENT_APP_FIREFOX;
+    const { store } = dispatchClientMetadata({ clientApp });
+    const addon = createInternalAddon(fakeAddon);
+    store.dispatch(_loadAddons({ addon }));
+
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+    renderComponent({
+      // Set the slug param to all uppercase
+      params: { slug: addon.slug.toUpperCase() },
```
bobsilverberg: Nit: There's no guarantee that `slug` in `fakeAddon` isn't already all upper case, so we should probably set a manual value for `slug` in the test. Of course, the test would fail if that were the case, but it still feels like the right thing to do, to me.

tests/unit/amo/pages/TestCollection.js
```diff
@@ -1392,6 +1371,67 @@ describe(__filename, () => {
       }),
     );
   });
+  it('sends a server redirect when username is not lower case', () => {
```
bobsilverberg: Should this description be about the case not matching, as opposed to "not lower case"?

tests/unit/amo/pages/TestCollection.js
```diff
@@ -1392,6 +1371,67 @@ describe(__filename, () => {
...
+
+    _loadCurrentCollection({ store });
+
+    fakeDispatch.resetHistory();
+
+    const collection = getCurrentCollection(store.getState().collections);
+
+    const params = {
+      slug: collection.slug,
+      username: collection.authorUsername.toUpperCase(),
```
bobsilverberg: Nit: As with the nit I filed above, I think we should set a known value for the collection `authorUsername` rather than assuming it's not already upper case.

tests/unit/amo/pages/TestCollection.js
```diff
@@ -1392,6 +1371,67 @@ describe(__filename, () => {
...
+    const fakeDispatch = sinon.spy(store, 'dispatch');
+
+    _loadCurrentCollection({ store });
+
+    fakeDispatch.resetHistory();
+
+    const collection = getCurrentCollection(store.getState().collections);
+
+    const params = {
+      slug: collection.slug.toUpperCase(),
```
bobsilverberg: Nit: As with the nit I filed above, I think we should set a known value for the collection `slug` rather than assuming it's not already upper case.

tests/unit/amo/pages/TestCollection.js
```diff
@@ -1392,6 +1371,67 @@ describe(__filename, () => {
...
+        status: 301,
+        url: `/${lang}/${clientApp}/collections/${collection.authorUsername}/${
+          collection.slug
+        }/`,
+      }),
+    );
+    sinon.assert.calledOnce(fakeDispatch);
+  });
+
+  it('sends a server redirect when slug is not lower case', () => {
```
bobsilverberg: Should this description be about the case not matching, as opposed to "not lower case"?

tests/unit/core/reducers/test_addons.js
```diff
@@ -101,6 +101,21 @@ describe(__filename, () => {
     });
   });
 
+  it('stores all add-on slugs in lowercase', () => {
+    const addonResults = [
```
bobsilverberg: Nit: You could avoid hardcoding these by making `FIRST` and `SeCond` consts and then using `toLowerCase()` in the assertion.

tests/unit/core/reducers/test_addons.js
```diff
@@ -566,6 +593,15 @@ describe(__filename, () => {
         createInternalAddon(fakeAddon),
       );
     });
+
+    it('is case insensitive', () => {
+      const { store } = dispatchClientMetadata();
+      store.dispatch(loadAddons(createFetchAddonResult(fakeAddon).entities));
+
+      expect(
+        getAddonBySlug(store.getState(), fakeAddon.slug.toUpperCase()),
```
bobsilverberg: Nit: As above, we're assuming that `fakeAddon.slug` is not already in upper case.

src/core/reducers/addons.js
```diff
@@ -293,7 +293,11 @@ export const getAddonBySlug = (
   state: AppState,
   slug: string,
 ): AddonType | null => {
-  const addonId = state.addons.bySlug[slug];
+  if (typeof slug !== 'string') {
```
bobsilverberg: This doesn't seem to be covered by a test, at least not a test where `slug` is neither a string nor `null` nor `undefined`.

src/core/reducers/addons.js
```diff
@@ -308,7 +312,11 @@ export const getAddonByGUID = (
 };
 
 export const isAddonLoading = (state: AppState, slug: string): boolean => {
-  return Boolean(state.addons.loadingBySlug[slug]);
+  if (typeof slug !== 'string') {
```
bobsilverberg: This doesn't seem to be covered by a test, at least not a test where `slug` is neither a string nor `null` nor `undefined`.

src/core/reducers/addons.js
bobsilverberg: Looking again, wouldn't Flow protect us from this ever happening?

src/core/reducers/addons.js
bobsilverberg: Looking again, wouldn't Flow protect us from this ever happening?

src/core/reducers/addons.js
willdurand: Flow won't be helpful at runtime.

There are 3 test cases for string, null or undefined. I think that's enough, but if yu like, I can still add another one with a number.

src/core/reducers/addons.js
willdurand: same as above

tests/unit/amo/pages/TestCollection.js
willdurand: ah right

tests/unit/amo/pages/TestCollection.js
willdurand: ah right

tests/unit/amo/pages/TestAddon.js
willdurand: well, calling the method implies that it is not already the case, but I can of course be more verbose and create an addon with a lowercase slug.

src/core/reducers/addons.js
willdurand: oh right, good catch!

tests/unit/core/reducers/test_addons.js
willdurand: I'd argue that the test case above explicitly proves that.

bobsilverberg: I'd argue that that statement proves my case. ;-)

src/core/reducers/addons.js
bobsilverberg: I'm not sure it's about what I like. If the code is written to return a `null` when a number is passed, it seems like we should have a test for that. Otherwise it could have just been `if(!slug)`.

tests/unit/core/reducers/test_addons.js
willdurand: haha, I changed it anyway

src/core/reducers/addons.js
willdurand: fair enough, updated!

PR #6558, opened 2018-10-09
Do not index featured reviews

willdurand:
Refs #5758

---

We do not want to index those pages, so let's tell robots about it.

### DIFF
--- a/tests/unit/amo/pages/TestAddonReviewList.js
+++ b/tests/unit/amo/pages/TestAddonReviewList.js
@@ -735,6 +735,38 @@ describe(__filename, () => {
       const root = render();
       expect(root.find('title')).toHaveLength(0);
     });
+
+    it('does not render a robots meta tag', () => {
+      const reviews = [
+        { ...fakeReview, id: 1, score: 1 },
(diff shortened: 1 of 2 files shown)


tests/unit/amo/pages/TestAddonReviewList.js
```diff
@@ -735,6 +735,38 @@ describe(__filename, () => {
...
+      ];
+      dispatchAddon(createInternalAddon(fakeAddon));
+      dispatchAddonReviews({ reviews });
+
+      const root = render({
+        params: { reviewId: reviews[0].id.toString() },
+      });
+
+      expect(root.find('meta[name="robots"]')).toHaveLength(1);
+      expect(root.find('meta[name="robots"]')).toHaveProp(
```
bobsilverberg: Nit: We don't really need this second assertion as it's asserting what is essentially static text. I think the test would be fine simply checking that the `robots` meta tag is there, but it's your call if you want to keep it.

willdurand: I prefer to keep it because if we change the value by mistake, it can become problematic.

PR #6566, opened 2018-10-09
Refactor theme data

### LINKED ISSUE #3892: Theme description is not listed if installed from Discovery Pane
Steps to reproduce:
1. Pick a theme with description - i.e. https://addons-dev.allizom.org/en-US/firefox/addon/all-fall/
2. Install it from AMO
3. Go to about:addons Themes tab and observe theme description
4. Go to Discovery Pane and install the same theme
5. Repeat step #3

Expected results:
Theme description is listed.

Actual results:
Theme description is not listed if installed from Discovery pane.

Notes/Issues:
Verified on AMO-dev FF58(Win 7).
Screenshots for this issue:
descr-amo
descr-disco_pane

Relates with #3798

willdurand:
Fixes #3892

---

This PR removes a bunch of code related to `theme_data`. I am confident now that

- the `AMInstallButton` is used on AMO and soon on Disco Pane
- flow coverage on Disco Pane is good
- old `InstallSwitch`/`InstallButton` components are scheduled for deletion

This actually fixes a small issue.

In order to test the old code (not for QA but for reviewers):

- make sure to disable the `enableFeatureAMInstallButton` flag in `config/development-disco.js`
- make sure you have the prefs `xpinstall.signatures.dev-root` and `extensions.webapi.testing` set to `true` too
- you should see the switch buttons with `yarn disco:https`

Installing a LW theme should work. Once done, go to `about:addons` and make sure there is your theme listed as installed (with a description) in the `themes` tab. That's it!

You can also test with the new install button because it is also affected (see STR of #3892).

### DIFF
--- a/tests/unit/core/reducers/test_addons.js
+++ b/tests/unit/core/reducers/test_addons.js
@@ -124,6 +124,7 @@ describe(__filename, () => {
       isRestartRequired: false,
       isWebExtension: true,
       isMozillaSignedExtension: false,
+      themeData: null,
     });
   });
 
@@ -135,12 +136,7 @@ describe(__filename, () => {
     // what we expect it to below.
     const expectedTheme = {
       ...theme,
-      // Expect theme_data to be merged into the addon.
-      ...theme.theme_data,
-      themeData: {
-        ...theme.theme_data,
-        description: theme.description,
-      },
+      themeData: theme.theme_data,
       description: theme.description,
       guid: getGuid(theme),
       platformFiles: {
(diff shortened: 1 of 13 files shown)


tests/unit/core/reducers/test_addons.js
```diff
@@ -251,26 +247,6 @@ describe(__filename, () => {
     });
   });
 
-  it('does not use description from theme_data', () => {
```
rebmullin: does this mean we ARE using the description from theme_data now?

willdurand: yes, for LWT themes

rebmullin: Should we keep this as test then, to show that it's using the right description?

willdurand: hum, no because this is not relevant anymore. Yet, the test case above should have a new description... From: `stores a modified theme object` to `stores the original theme object`

Issue #6458, opened 2018-09-28
Simplify delete confirmation UI in review listing


kumar303:
### Describe the problem and steps to reproduce it:

- Sign in
- Rate an add-on and write a review
- Go to the review listing
- Find the review and click *Delete review*

### What happened?

It shows a prompt with two buttons

### What did you expect to happen?

It should be simplified like https://github.com/mozilla/addons-frontend/issues/6347

pwalm:
Proposed:
<img width="1070" alt="screen shot 2018-09-28 at 1 52 56 pm" src="https://user-images.githubusercontent.com/6797635/46233013-07183280-c326-11e8-8590-19fc80fb6d32.png">

When "Delete review" is clicked, the prompt appears in place.

ioanarusiczki:
Verified this issue on AMO dev with FF62 - Win10&Android 8.0 
I filed two separate issues 
https://github.com/mozilla/addons-frontend/issues/6612 
https://github.com/mozilla/addons-frontend/issues/6613

@kumar303 I don't know if the "Flag" button should be or can be removed for personal reviews and replies - user cannot flag it's own review or reply anyway.

Issue #6347, opened 2018-09-21
Simplify the confirmation screen for deleting a review / rating


kumar303:
~~Depends on #6454~~

Here is how to see the current deletion prompts:

- Sign in
- Go to an add-on detail page
- Rate the add-on
- Click *Delete rating*
- Or ... Write a review
- Click *Delete review*

The screens for each of these should be simplified. Here is an example (below) but check Sketch for an up to date version.

## Known issues

* The stars will be made smaller according to the mocks in https://github.com/mozilla/addons-frontend/issues/6577
* When deleting a review, the prompt at the top of *Rate your experience* will be added in https://github.com/mozilla/addons-frontend/issues/6577
* @pwalm called for underlining all control links (*Edit*, *Delete*, *Flag*, etc) for accessibility but the mocks have not been updated to show this yet
* We intend to always show the *Report this add-on for abuse* button even though it's not in the mocks. See https://github.com/mozilla/addons-frontend/issues/6307#issuecomment-426713063
* The *Edit review* and *Delete review* links are not aligned to the right (like the mock) because the card size is very small. This needs more discussion which will happen in https://github.com/mozilla/addons-frontend/issues/6403

![ratings and reviews flow v2](https://user-images.githubusercontent.com/55398/45894396-ceabae00-bd93-11e8-9b44-cdb7bc82eb2e.png)

bobsilverberg:
> Here is an example but check Sketch for an up to date version.

@kumar303 / @pwalm I looked in the Sketch file but cannot find anything like this anywhere? Are they there and I'm just not looking in the correct place?

Issue #6305, opened 2018-09-19
Hide edit link when confirming review deletion


kumar303:
STR:
- Select a star to rate an add-on
- Write a review and submit the review
- Click *Delete my review* to open the confirmation screen.

You see this:

<img width="358" alt="screenshot 2018-09-19 14 42 51" src="https://user-images.githubusercontent.com/55398/45777040-4d7cdb80-bc1a-11e8-96e3-f10ff9ec59e9.png">

...but the *Edit my review* link should be hidden.

This came up as feedback in a recent UX meeting.

ioanarusiczki:
This is not reproducible anymore - verified on AMO dev with FF62 (Win10&Android 8.0)

![no edit available while deleting](https://user-images.githubusercontent.com/33448286/46877132-cd2c4d80-ce48-11e8-922b-47a9dc713fce.gif)

PR #6567, opened 2018-10-09
Simplify the confirm delete UI for reviews / ratings

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6458; fixes https://github.com/mozilla/addons-frontend/issues/6347 (because it only required changing the prompt); fixes https://github.com/mozilla/addons-frontend/issues/6305 (which is irrelevant now)

Read https://github.com/mozilla/addons-frontend/issues/6347 for some known issues -- e.g. where this patch strays from the mocks. I have discussed some discrepancies with Phil but let me know if you have questions.

New screens:

screenshot 2018-10-09 17 10 58
screenshot 2018-10-09 17 17 37
screenshot 2018-10-09 17 18 01
screenshot 2018-10-09 17 18 19
screenshot 2018-10-09 17 18 39
screenshot 2018-10-09 17 19 11
screenshot 2018-10-09 17 19 24
screenshot 2018-10-09 17 19 43
screenshot 2018-10-09 17 20 54
screenshot 2018-10-09 17 21 16

### DIFF
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -305,6 +305,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       i18n,
       replyingToReview,
       review,
+      smallCard,
       submittingReply,
     } = this.props;
 
@@ -338,6 +339,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
             addon={addon}
             isReplyToReviewId={review.id}
             review={review.reply}
+            smallCard={smallCard}
           />
         )}
       </div>
@@ -487,6 +489,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
     return (
       <div
         className={makeClassName('AddonReviewCard', className, {
+          'AddonReviewCard-isReply': this.isReply(),
           'AddonReviewCard-ratingOnly': this.isRatingOnly(),
           'AddonReviewCard-viewOnly': !editingReview,
           'AddonReviewCard-smallCard': smallCard,
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -406,7 +406,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
     }
 
     let controlsAreVisible = showControls;
-    if (beginningToDeleteReview && !smallCard) {
+    if (beginningToDeleteReview && (this.isRatingOnly() || !smallCard)) {
       controlsAreVisible = false;
     }
 
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -17,6 +17,8 @@ import log from 'core/logger';
 import { getCurrentUser, hasPermission } from 'amo/reducers/users';
 import { isAddonAuthor } from 'core/utils';
 import {
+  beginDeleteAddonReview,
+  cancelDeleteAddonReview,
   deleteAddonReview,
   hideEditReviewForm,
   hideReplyToReviewForm,
@@ -25,7 +27,7 @@ import {
   showReplyToReviewForm,
 } from 'amo/actions/reviews';
 import Button from 'ui/components/Button';
-import ConfirmButton from 'ui/components/ConfirmButton';
+import ConfirmationDialog from 'ui/components/ConfirmationDialog';
 import DismissibleTextForm from 'ui/components/DismissibleTextForm';
 import Icon from 'ui/components/Icon';
 import LoadingText from 'ui/components/LoadingText';
@@ -51,6 +53,7 @@ type Props = {|
   shortByLine?: boolean,
   showControls?: boolean,
   showRating?: boolean,
+  // TODO: rename to smallCard ?
   verticalButtons?: boolean,
 |};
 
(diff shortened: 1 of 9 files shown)


tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -296,17 +348,131 @@ describe(__filename, () => {
...
+    deleteLink.simulate('click', fakeEvent);
+
+    sinon.assert.calledWith(
+      dispatchSpy,
+      beginDeleteAddonReview({ reviewId: review.id }),
+    );
+    sinon.assert.called(fakeEvent.preventDefault);
+  });
+
+  it('renders generic delete confirmation buttons', () => {
```
bobsilverberg: Nit: This is a general comment about a the tests that follow. I find it a bit hard to grasp what these are testing without looking closely at the code, which I suppose should maybe be expected. The issue is that there are tests for `smallCard: true` and tests for `smallCard: false` but they're not labelled that way. It seems like the tests for `smallCard: false` are stated like they're the default behaviour (e.g., `renders generic delete confirmation buttons` and `renders a confirmation prompt for deleting a review`) whereas the tests where `smallCard: true` are sometimes specifically called out and sometimes not (e.g., `renders delete confirmation buttons for a rating` and `does not render a confirmation message for smallCard=true`). I wonder if the tests might be easier to understand if you divided these into two `describe` blocks, one for `smallCard: true` and one for `smallCard: false`?

I know you said you considered adding some `describe` blocks and decided it would make things more confusing, but maybe this would be an improvement?

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -330,7 +496,7 @@ describe(__filename, () => {
       }),
     );
 
-    const root = render({ review });
+    const root = render({ review, smallCard: true });
```
bobsilverberg: Why `smallCard: true`? Doesn't the message also appear when `smallCard === false`?

src/amo/reducers/reviews.js
```diff
@@ -364,13 +384,13 @@ export default function reviewsReducer(
...
         reviewId: action.payload.reviewId,
-        stateChange: { editingReview: true },
+        stateChange: { beginningToDeleteReview: false, editingReview: true },
       });
     case SHOW_REPLY_TO_REVIEW_FORM:
       return changeViewState({
         state,
         reviewId: action.payload.reviewId,
-        stateChange: { replyingToReview: true },
+        stateChange: { beginningToDeleteReview: false, replyingToReview: true },
```
bobsilverberg: Just wondering, what's the use case for this? I couldn't seem to replicate a condition where I am deleting a review, but can also reply to it.

tests/unit/amo/reducers/test_reviews.js
```diff
@@ -570,13 +572,27 @@ describe(__filename, () => {
...
-          reviewId: review.id,
-        }),
+        showEditReviewForm({ reviewId: review.id }),
       );
 
       expect(state.view[review.id].editingReview).toEqual(true);
     });
+
+    it('exits beginDeleteAddonReview state', () => {
+      const review = { ...fakeReview, id: 837 };
```
bobsilverberg: Nit: This applies to most tests added to this file: You don't really need to set a specific `id` for the `fakeReview` because you are never referring to that `id` directly. You just use `review.id` throughout the test, so it's unnecessary for the test to "know" what the `id` is.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -450,66 +489,83 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
+          />
         )}
-        {errorHandler.renderErrorIfPresent()}
         {this.renderReply()}
+        {siteUser &&
+          review &&
+          review.userId === siteUser.id &&
+          this.isRatingOnly() &&
+          !editingReview && (
+            <Button
```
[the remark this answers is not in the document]
kumar303: This was unintended. I'll fix it.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -366,15 +407,18 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       byLine = <LoadingText />;
     }
 
-    const confirmButtonClassName = 'AddonReviewCard-delete';
+    let controlsAreVisible = showControls;
+    if (beginningToDeleteReview && (this.isRatingOnly() || !smallCard)) {
```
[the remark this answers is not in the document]
kumar303: Good idea. I like it. Fixed.

src/amo/reducers/reviews.js
kumar303: In an earlier version the controls were still visible while you were deleting so it was possible back then.

tests/unit/amo/reducers/test_reviews.js
kumar303: All of those tests do rely on `review.id` for things like when they check `state.view[review.id].editingReview`. Sometimes `fakeReview` is passed in as a default value to helpers so I thought it would be safer to give each one a unique ID. I guess it's not so important.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -46,18 +48,18 @@ type Props = {|
   className?: string,
   flaggable?: boolean,
   isReplyToReviewId?: number,
-  smallerWriteReviewButton?: boolean,
   review?: UserReviewType | null,
   shortByLine?: boolean,
   showControls?: boolean,
   showRating?: boolean,
-  verticalButtons?: boolean,
+  smallCard?: boolean,
```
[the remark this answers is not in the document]
kumar303: Names are hard 🍤 I just changed it to `slim` which better illustrates the problem: the container doesn't have much width. I also added a comment explaining how it's used.

kumar303:
Thanks. It's ready for another look.

Issue #6561, opened 2018-10-09
Fix circular dependency in `tests/unit/helpers.js`


kumar303:
If you run `yarn storybook` from https://github.com/mozilla/addons-frontend/pull/6389 with the circular dependencies plugin enabled, you see:

```
ERROR in Circular dependency detected:
tests/unit/amo/helpers.js -> tests/unit/helpers.js -> tests/unit/amo/helpers.js

ERROR in Circular dependency detected:
tests/unit/helpers.js -> tests/unit/amo/helpers.js -> tests/unit/helpers.js
```

This is a totally legit error but we don't see it anywhere else. I guess it's because the code in `./stories` is importing test files in a webpack bundle, something we don't normally do.

Let's fix the circular dependency.

After this fix, the circular dependency plugin can be enabled for storybook.

kumar303:
I'm open to ideas on how to fix this. One possibility is to merge everything into one helper. Another is to split helpers into smaller modules according to what they do (we do have a lot of helpers now).

kumar303:
Merging `tests/unit/amo/helpers.js` into `tests/unit/helpers.js` sounds good to me.

PR #6571, opened 2018-10-09
I6172 cleaning up stories webpack config

### LINKED ISSUE #6172: Do not duplicate the webpack config in storybook

rebmullin:
fixes #6172 

I had to use the Prod config as Dev doesn't work b/c of certain components (like log, etc).

Regarding the getRules...if we don't want to show prop types, then we can use this as is (without the modification that I've added here). I personally find seeing the props types helpful so I looked into this a bit more and noticed if we did not include the babel-loader this works. I am not quite sure why this would cause a problem but hopefully this workaround will be okay :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -283,6 +283,7 @@
     "babel-jest": "^23.0.0",
     "babel-loader": "^7.0.0",
     "babel-plugin-dynamic-import-node": "^1.2.0",
+    "babel-plugin-react-docgen": "^1.9.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
--- a/package.json
+++ b/package.json
@@ -283,7 +283,6 @@
     "babel-jest": "^23.0.0",
     "babel-loader": "^7.0.0",
     "babel-plugin-dynamic-import-node": "^1.2.0",
-    "babel-plugin-react-docgen": "^1.9.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
--- a/stories/setup/webpack.config.babel.js
+++ b/stories/setup/webpack.config.babel.js
@@ -1,12 +1,18 @@
 import { getPlugins, getRules } from '../../webpack-common';
-import webpackConfig from '../../webpack.prod.config.babel';
+import webpackConfig, {
+  babelOptions as defaultBabelOptions,
+} from '../../webpack.dev.config.babel';
+
+// We need to add the docgen plugin to get Flow to work.
+const babelOptions = {
+  ...defaultBabelOptions,
+  plugins: [...defaultBabelOptions.plugins, 'react-docgen'],
+};
 
 module.exports = {
   ...webpackConfig,
   module: {
-    rules: [
-      ...getRules({ bundleStylesWithJs: true, includeBabelLoader: false }),
-    ],
+    rules: [...getRules({ babelOptions, bundleStylesWithJs: true })],
   },
   // TODO: add the circularDependency plugin back once the following issue is fixed:
   // https://github.com/mozilla/addons-frontend/issues/6561.
--- a/stories/setup/webpack.config.babel.js
+++ b/stories/setup/webpack.config.babel.js
@@ -14,11 +14,5 @@ module.exports = {
   module: {
     rules: [...getRules({ babelOptions, bundleStylesWithJs: true })],
   },
-  // TODO: add the circularDependency plugin back once the following issue is fixed:
-  // https://github.com/mozilla/addons-frontend/issues/6561.
-  plugins: [
-    ...getPlugins({
-      includeCircularDependencyPlugin: false,
-    }),
-  ],
+  plugins: [...getPlugins()],
 };
--- a/stories/setup/webpack.config.babel.js
+++ b/stories/setup/webpack.config.babel.js
@@ -1,4 +1,4 @@
-import { getPlugins, getRules } from '../../webpack-common';
+import { getRules } from '../../webpack-common';
 import webpackConfig, {
   babelOptions as defaultBabelOptions,
 } from '../../webpack.dev.config.babel';
@@ -14,5 +14,4 @@ module.exports = {
   module: {
     rules: [...getRules({ babelOptions, bundleStylesWithJs: true })],
   },
-  plugins: [...getPlugins()],
 };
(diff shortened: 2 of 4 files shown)


willdurand:
> I had to use the Prod config as Dev doesn't work b/c of certain components (like log, etc).

Why exactly? The patch I pasted in #6172 was working fine. I don't see any reference to `log` in the webpack prod config.

stories/setup/webpack.config.babel.js
```diff
@@ -1,58 +1,18 @@
...
-  // The following plugins are needed to help handle
-  // server-side imported components such as core/logger.
-  // Is this the best way??
+  // TODO: add the circularDependency plugin back once the following issue is fixed:
+  // https://github.com/mozilla/addons-frontend/issues/6561.
   plugins: [
-    new webpack.DefinePlugin({
-      CLIENT_CONFIG: JSON.stringify(getClientConfig(config)),
+    ...getPlugins({
+      includeCircularDependencyPlugin: false,
```
willdurand: I suggest fixing https://github.com/mozilla/addons-frontend/issues/6561 first so that we don't forget to actually do it.

stories/setup/webpack.config.babel.js
```diff
@@ -1,58 +1,18 @@
...
-                  browsers: ['last 2 versions'],
-                  grid: false,
-                }),
-              ],
-            },
-          },
-          { loader: 'sass-loader', options: { outputStyle: 'expanded' } },
-        ],
-      },
+      ...getRules({ bundleStylesWithJs: true, includeBabelLoader: false }),
```
willdurand: `bundleStylesWithJs` set to `true` is the default in the dev webpack config.

webpack-common.js
```diff
@@ -118,13 +122,20 @@ export function getRules({ babelOptions, bundleStylesWithJs = false } = {}) {
     ];
   }
 
-  return [
-    {
+  let babelLoader = {};
+  // The main purpose for this is when we call getRules from the storybook setup
+  // (stories), it causes flow not to work.
+  if (includeBabelLoader) {
```
willdurand: Do you know why this isn't working with Flow?

rebmullin: I dug in a little bit more and saw if I added the following as plugin to the babelrc file it would keep flow types here

"babel-plugin-react-docgen"

So I will make sure to include this : )

rebmullin:
> Why exactly? The patch I pasted in #6172 was working fine. I don't see any reference to log in the webpack prod config.

very weird..I am not sure but it's working for me now too 🤷‍♂️

stories/setup/webpack.config.babel.js
rebmullin: yeh but I need to still call getRules to override babeloptions (unless I add the plugin I noted below to the root .babelrc file. Maybe that's an option too..)

stories/setup/webpack.config.babel.js
```diff
@@ -1,58 +1,24 @@
...
-
-import config from 'config';
-import autoprefixer from 'autoprefixer';
-import webpack from 'webpack';
-
-import { getClientConfig } from 'core/utils';
+// We need to add this extra plugin to get Flow to work.
+const babelOptions = {
+  ...defaultBabelOptions,
+  plugins: 'babel-plugin-react-docgen',
```
rebmullin: maybe I should create it's own babelrc file here to make it more flexible or what do you think?

stories/setup/webpack.config.babel.js
```diff
@@ -1,58 +1,18 @@
...
-    ),
-  ],
-  resolve: {
-    alias: {
-      normalize: 'normalize.css/normalize.css',
-    },
-    modules: [path.resolve('./src'), 'node_modules'],
+    rules: [...getRules({ babelOptions, bundleStylesWithJs: true })],
   },
+  plugins: [...getPlugins()],
```
willdurand: Is this needed?

package.json
```diff
@@ -283,6 +283,7 @@
     "babel-jest": "^23.0.0",
     "babel-loader": "^7.0.0",
     "babel-plugin-dynamic-import-node": "^1.2.0",
+    "babel-plugin-react-docgen": "^1.9.0",
```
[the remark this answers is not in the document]
willdurand: I am not sure, if it is already a dependency of Storybook, why adding it here? Does eslint complain?

stories/setup/webpack.config.babel.js
rebmullin: hmm..I thought I saw an issue so I thought I had to override dev but actually it's fine so I've removed this.

package.json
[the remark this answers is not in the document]
rebmullin: No, I guess it's fine so I'll remove this..

PR #6591, opened 2018-10-11
Create a formatFilesize utilitiy which deals with i18n

### LINKED ISSUE #6586: Create a filesize utilitiy which deals with i18n
This is needed to support https://github.com/mozilla/addons-frontend/issues/2770. We need to display a user-friendly file size, given a number of bytes returned by the API. This needs to also take into account l10n, so `12345678` would display as "11.77 MB" in English, but might display as "11.77 ميغابايت" in Arabic [1] and as "11,77 Mo" in French [2].

[1] https://translate.google.com/#en/ar/11.77%20MB
[2] https://translate.google.com/#en/fr/11.77%20MB

bobsilverberg:
Fixes #6586

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,7 @@
     "express": "4.16.2",
     "extract-text-webpack-plugin": "3.0.2",
     "fastclick": "1.0.6",
+    "filesize": "3.6.1",
     "focus-visible": "4.1.5",
     "fs-extra": "7.0.0",
     "full-icu": "1.2.1",
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -254,12 +254,12 @@ export const formatFilesize = ({
     TB: i18n.gettext('%(formattedSize)s TB'),
   };
 
-  const [sizeNum, sizeName] = _filesize(size).split(' ');
-  if (!sizeNum || !sizeName) {
+  const [sizeNumber, sizeName] = _filesize(size).split(' ');
+  if (!sizeNumber || !sizeName) {
     return null;
   }
 
-  const formattedSize = i18n.formatNumber(sizeNum);
+  const formattedSize = i18n.formatNumber(sizeNumber);
   const sizeString = sizeStrings[sizeName];
 
   return sizeString ? i18n.sprintf(sizeString, { formattedSize }) : null;
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -249,26 +249,25 @@ export const formatFilesize = ({
   const sizeStrings = {
     // These are the expected values for the unit of measure returned by
     // filesize. Realistically we shouldn't get anything back larger than TB.
-    // translators: B is an abbreviation of Bytes in English.
-    // Localize it if necessary but use a short abbreviation.
+    /* eslint-disable max-len */
+    // translators: B is an abbreviation of Bytes in English. Localize it if necessary but use a short abbreviation.
     B: i18n.gettext('%(localizedSize)s B'),
-    // translators: KB is an abbreviation of Kilobytes in English.
-    // Localize it if necessary but use a short abbreviation.
+    // translators: KB is an abbreviation of Kilobytes in English. Localize it if necessary but use a short abbreviation.
     KB: i18n.gettext('%(localizedSize)s KB'),
-    // translators: MB is an abbreviation of Megabytes in English.
-    // Localize it if necessary but use a short abbreviation.
+    // translators: MB is an abbreviation of Megabytes in English. Localize it if necessary but use a short abbreviation.
     MB: i18n.gettext('%(localizedSize)s MB'),
-    // translators: GB is an abbreviation of Gigabytes in English.
-    // Localize it if necessary but use a short abbreviation.
+    // translators: GB is an abbreviation of Gigabytes in English. Localize it if necessary but use a short abbreviation.
     GB: i18n.gettext('%(localizedSize)s GB'),
-    // translators: TB is an abbreviation of Terabytes in English.
-    // Localize it if necessary but use a short abbreviation.
+    // translators: TB is an abbreviation of Terabytes in English. Localize it if necessary but use a short abbreviation.
     TB: i18n.gettext('%(localizedSize)s TB'),
+    /* eslint-enable max-len */
   };
 
   const [sizeNumber, sizeName] = _filesize(size).split(' ');
   if (!sizeNumber || !sizeName) {
-    _log.error(`Filesize returned unparseable data for size ${size}`);
+    _log.error(
+      `Filesize returned sizeNumber: "${sizeNumber}", sizeName: "${sizeName}" size "${size}"`,
+    );
     return i18n.formatNumber(size);
   }
 
--- a/src/core/i18n/utils.js
+++ b/src/core/i18n/utils.js
@@ -234,6 +234,7 @@ function oneLineTranslationString(translationKey) {
 
 type FormatFilesizeParams = {|
   _filesize: typeof filesize,
+  _log: typeof log,
   i18n: I18nType,
   size: number,
 |};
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -214,6 +214,7 @@
     "express": "4.16.2",
     "extract-text-webpack-plugin": "3.0.2",
     "fastclick": "1.0.6",
+    "filesize": "3.6.1",
```
rebmullin: when you update the package.json file, this should update the yarn.lock file. Can you make sure that's added too?

tests/unit/core/i18n/test_utils.js
```diff
@@ -509,4 +510,48 @@ describe(__filename, () => {
...
+    it('returns null for an invalid unit of measure from filesize', () => {
+      const _filesize = sinon.stub().returns('123 BOB');
+      expect(_formatFilesize({ _filesize })).toEqual(null);
+    });
+
+    it.each(['B', 'KB', 'MB', 'GB', 'TB'])(
+      'calls i18n.sprintf with the expected substitution for size %s',
+      (sizeName) => {
+        const i18n = fakeI18n();
+        const size = '123';
```
rebmullin: shouldn't '123' be a number (not a string)?

src/core/i18n/utils.js
```diff
@@ -230,6 +232,39 @@ function oneLineTranslationString(translationKey) {
...
+    // These are the expected values for the unit of measure returned by
+    // filesize. Realistically we shouldn't get anything back larger than TB.
+    B: i18n.gettext('%(formattedSize)s B'),
+    KB: i18n.gettext('%(formattedSize)s KB'),
+    MB: i18n.gettext('%(formattedSize)s MB'),
+    GB: i18n.gettext('%(formattedSize)s GB'),
+    TB: i18n.gettext('%(formattedSize)s TB'),
+  };
+
+  const [sizeNum, sizeName] = _filesize(size).split(' ');
```
rebmullin: nit: would it be better to have sizeNum be sizeNumber?

tests/unit/core/i18n/test_utils.js
bobsilverberg: I don't think so. It's used as a string everywhere in the test, and it's also going to be a string inside `formatFilesize` when passed into `sprintf`.

bobsilverberg:
Thanks for the review @rebmullin. I've addressed two of your issues, but I don't think `'123'` should be a number in the test.

tests/unit/core/i18n/test_utils.js
rebmullin: yep that makes sense

src/core/i18n/utils.js
```diff
@@ -230,6 +232,39 @@ function oneLineTranslationString(translationKey) {
...
+// Translates a file size in bytes into a localized user-friendly format.
+export const formatFilesize = ({
+  _filesize = filesize,
+  i18n,
+  size,
+}: FormatFilesizeParams): string | null => {
+  const sizeStrings = {
+    // These are the expected values for the unit of measure returned by
+    // filesize. Realistically we shouldn't get anything back larger than TB.
+    B: i18n.gettext('%(formattedSize)s B'),
```
kumar303: I don't think this will be obvious to localizers so you should give each one of these an explantory comment. Example:
```js
const sizeStrings = {
  // translators: B is an abbreviation of Bytes in English. Localize it if necessary but use a short abbreviation.
  B: i18n.gettext('%(formattedSize)s B'),
  // translators: KB is an abbreviation of Kilobytes in English. Localize it if necessary but use a short abbreviation.
  KB: i18n.gettext('%(formattedSize)s KB'),
};
```

kumar303: Calling this `%(localizedSize)s` may give translators a better hint about what's happening.

src/core/i18n/utils.js
```diff
@@ -230,6 +232,39 @@ function oneLineTranslationString(translationKey) {
...
+    KB: i18n.gettext('%(formattedSize)s KB'),
+    MB: i18n.gettext('%(formattedSize)s MB'),
+    GB: i18n.gettext('%(formattedSize)s GB'),
+    TB: i18n.gettext('%(formattedSize)s TB'),
+  };
+
+  const [sizeNumber, sizeName] = _filesize(size).split(' ');
+  if (!sizeNumber || !sizeName) {
+    return null;
```
kumar303: Returning `null` seems unhelpful here. I would suggest doing this:

* Log an error so that Sentry will tell us we passed input that confused `filesize()`. If we see one of these for something out of our control, we could downgrade the log message to a warning.
* Return `i18n.formatNumber(sizeNumber)` without any units because I think that is a pretty good fallback. It may not make sense but it would make more sense than an empty string.

src/core/i18n/utils.js
```diff
@@ -230,6 +232,39 @@ function oneLineTranslationString(translationKey) {
...
+  }
+
+  const formattedSize = i18n.formatNumber(sizeNumber);
+  const sizeString = sizeStrings[sizeName];
+
+  return sizeString ? i18n.sprintf(sizeString, { formattedSize }) : null;
```
kumar303: Similar to above, the case of an unmapped unit would be more helpful like this:

* Log an error so that Sentry will tell us that we have failed to map a known unit
* Return `i18n.formatNumber(sizeNumber)` without any units.

bobsilverberg:

src/core/i18n/utils.js
```diff
@@ -230,6 +232,56 @@ function oneLineTranslationString(translationKey) {
...
+  _filesize = filesize,
+  _log = log,
+  i18n,
+  size,
+}: FormatFilesizeParams): string | null => {
+  const sizeStrings = {
+    // These are the expected values for the unit of measure returned by
+    // filesize. Realistically we shouldn't get anything back larger than TB.
+    // translators: B is an abbreviation of Bytes in English.
+    // Localize it if necessary but use a short abbreviation.
```
kumar303: It's a bit lame but you have to put the `translators:` prefixed comment all on one line. You can test it by running `bin/extract-locales` then checking `amo.pot`.

src/core/i18n/utils.js
```diff
@@ -230,6 +232,56 @@ function oneLineTranslationString(translationKey) {
...
+    GB: i18n.gettext('%(localizedSize)s GB'),
+    // translators: TB is an abbreviation of Terabytes in English.
+    // Localize it if necessary but use a short abbreviation.
+    TB: i18n.gettext('%(localizedSize)s TB'),
+  };
+
+  const [sizeNumber, sizeName] = _filesize(size).split(' ');
+  if (!sizeNumber || !sizeName) {
+    _log.error(`Filesize returned unparseable data for size ${size}`);
```
kumar303: Seeing some actual values may be more helpful for debugging:

```
Filesize returned sizeNumber="${sizeNumber}", sizeName="${sizeName}" for size "${size}"
```

tests/unit/core/i18n/test_utils.js
```diff
@@ -509,4 +510,73 @@ describe(__filename, () => {
...
+      i18n = fakeI18n(),
+      size = 123,
+    }) => {
+      return utils.formatFilesize({ _filesize, _log, i18n, size });
+    };
+
+    it('receives an expected string from filesize', () => {
+      // This is a sanity check, verifying that filesize does in fact return
+      // what we expect.
+      // Without it filesize could break and we'd never know.
```
kumar303: I like this test but do you really need to use a mock version of `filesize()` for the other tests? What about using the real function?

tests/unit/core/i18n/test_utils.js
```diff
@@ -509,4 +510,73 @@ describe(__filename, () => {
...
+    });
+
+    it('returns the size for an invalid string from filesize', () => {
+      const size = 987;
+      const _filesize = sinon.stub().returns('123');
+      expect(_formatFilesize({ _filesize, size })).toEqual(size.toString());
+    });
+
+    it('logs an error for an invalid string from filesize', () => {
+      const fakeLog = {
```
kumar303: You can use the shiny new `fakeLogger` now:

https://github.com/mozilla/addons-frontend/blob/8649ddcd51b59aaff94c4a1b5ff53803fae10841/tests/unit/helpers.js#L1120-L1128

src/core/i18n/utils.js
bobsilverberg: Hmm, I was wondering about that so I checked some other code and saw translation notes on multiple lines, e.g., [here](https://github.com/mozilla/addons-frontend/blob/2018.10.11/src/amo/components/AddonMoreInfo/index.js#L125-L126) and [here](https://github.com/mozilla/addons-frontend/blob/2018.10.11/src/amo/pages/AddonReviewList/index.js#L302-L303). I guess those are bad. Maybe we need to open a code quality issue to check for multi-line translation hints?

tests/unit/core/i18n/test_utils.js
bobsilverberg: I just figured that mocking was better, so that we're isolating the test to the code we're concerned with. I guess I can use the real function everywhere, in which case I probably don't even need to configure the function to accept a mock. That seems contrary to the way we usually write tests though.

tests/unit/core/i18n/test_utils.js
```diff
@@ -509,4 +510,70 @@ describe(__filename, () => {
...
+    const _formatFilesize = ({
+      _filesize,
+      _log,
+      i18n = fakeI18n(),
+      size = 123,
+    }) => {
+      return utils.formatFilesize({ _filesize, _log, i18n, size });
+    };
+
+    it('receives an expected string from filesize', () => {
```
kumar303: I don't think you need this test because `formats the number returned by filesize` covers the same thing.

PR #6595, opened 2018-10-11
Add a test case for `onLocationChanged`

### LINKED ISSUE #6560: Add a test case for `onLocationChanged` helper function/action
See: https://github.com/mozilla/addons-frontend/pull/6466#discussion_r223745798

The idea is to make sure our version of `onLocationChanged` does not get out of sync with the upstream lib.

willdurand:
Fixes #6560

---

This test case makes sure our helper function is always in sync with the
upstream lib.

### DIFF
--- a/tests/unit/test_helpers.js
+++ b/tests/unit/test_helpers.js
@@ -1,7 +1,11 @@
 /* eslint-disable react/no-multi-comp */
-import { shallow } from 'enzyme';
 import React, { Component } from 'react';
+import { ConnectedRouter } from 'connected-react-router';
+import { shallow, mount } from 'enzyme';
+import { createMemoryHistory } from 'history';
 import { compose } from 'redux';
+import { Provider } from 'react-redux';
+import { Route } from 'react-router-dom';
 import SagaTester from 'redux-saga-tester';
 
 // Disabled because of
@@ -10,9 +14,11 @@ import SagaTester from 'redux-saga-tester';
 import { put, takeLatest } from 'redux-saga/effects';
 /* eslint-enable import/order */
 
+import createStore from 'amo/store';
 import {
   getFakeConfig,
   matchingSagaAction,
+  onLocationChanged,
   shallowUntilTarget,
   unexpectedSuccess,
 } from 'tests/unit/helpers';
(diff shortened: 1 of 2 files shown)


tests/unit/amo/test_helpers.js
```diff
@@ -0,0 +1,38 @@
...
+      );
+
+      dispatchSpy.resetHistory();
+
+      const pathname = '/foo';
+      history.push(pathname);
+
+      sinon.assert.calledWith(
+        dispatchSpy,
+        sinon.match(onLocationChanged({ pathname })),
```
kumar303: I think this should not use `sinon.match()` because we *do* want to know if there are more or less props in the payload. Was there a dynamic value giving you trouble? If so, you could create a custom sinon matcher.

kumar303: Okay. Here's an example of a custom matcher:

https://github.com/mozilla/addons-frontend/blob/6dcad19b0dc9fb21ed6b42cfe351fc0ec803ede6/tests/unit/helpers.js#L845-L867

willdurand: I see. It seems a lot of code to me, for a matcher that won't be reused.

The use of `sinon.match` here catches the following bugs:

- `action` set to another constant in the upstream lib
- `location` not set or with a different shape (i.e. `pathname` not present anymore)

That's why I thought that would be enough. Let me come back with something stricter yet still simple.

tests/unit/test_helpers.js
```diff
@@ -293,4 +299,34 @@ describe(__filename, () => {
...
+      );
+
+      dispatchSpy.resetHistory();
+
+      const pathname = '/foo';
+      history.push(pathname);
+
+      sinon.assert.calledWith(
+        dispatchSpy,
+        onLocationChanged({ pathname, key: sinon.match.string }),
```
kumar303: Ah, yes, I was thinking of `sinon.match.string` but I thought it only worked for mock expectations. I think this will be helpful.

willdurand: Yes, that's super interesting!

Issue #1515, opened 2016-12-21
Missing moment locales for several of our own locales


tofumatt:
Once #1514 is merged we're using [moment] for our dates but it's missing locales for the following locales we support:

* [ ] ast - Asturian (missing)
* [ ] cak - Kaqchikel (missing)
* [x] bn_BD - Bengali - moment has `bn` so we could map this.
* [x] dbl - debug ltr - Could map to `x-pseudo`
* [x] dbr - debug rtl - Could map to `x-pseudo`
* [ ] dsb - Lower Sorbian  (missing)
* [x] fy_NL - Frisian - moment has `fy` so this could be mapped
* [ ] ga_IE - Gaelic irish (missing)
* [ ] hsb - Upper Sorbian (missing)
* [ ] kab - Kabyle (missing)
* [x] nb_NO - Norwegian Bokmål  can be mapped to `nb`.
* [x] nn_NO - Nynorsk  can be mapped to `nn`.
* [x] mn - Mongolian present. No mapping required.
* [x] pt_PT - Portuguese - available as `pt` needs to be mapped.
* [x] sv_SE - Swedish - available as `sv` needs to be mapped.

We should add them to moment or help them get added.

List updated by @muffinresearch on at 1st August 2018

willdurand:
@muffinresearch it is written we can add a mapping for some locales, but I don't know how to do that.

I am not sure our code allows that. There is a mapping for short -> long locales, but `pt -> pt_PT` is there and moment still complains about `pt-PT` being missing for instance.

muffinresearch:
I think we'd need a moment specific mapping and do that replacement here: https://github.com/mozilla/addons-frontend/blob/f8d1f2c7f4aaa6053330c9742a688469920384d4/bin/build-locales

tofumatt:
Oops, I think I wasn't clear in the issue but if I recall correctly this was meant to track mapping our locales to moment's. Our mappings are in https://github.com/mozilla/addons-frontend/blob/master/config/default.js#L236 but yeah they don't do it for moment.

I'm guessing the issue is that moment doesn't like the dashes?

tofumatt:
Indeed, @muffinresearch nailed it 😄

willdurand:
Mapping added in https://github.com/mozilla/addons-frontend/issues/6623

KevinMind:
Old Jira Ticket: https://mozilla-hub.atlassian.net/browse/ADDFRNT-114

PR #6614, opened 2018-10-12
Add moment locale mapping

### LINKED ISSUE #6623: Allow to map moment locales
See also: #1515

---

We should add some config and logic in (`bin/build-locales`) to map unsupported moment locales to supported moment locales (for example, using the short locale that groups similar languages).

willdurand:
Refs #1515 
Fixes #6623

---

## Before

(see the updated date in the left column)

screen shot 2018-10-12 at 15 59 20

## After

(see the updated date in the left column)

screen shot 2018-10-12 at 15 52 04

### DIFF
--- a/bin/build-locales
+++ b/bin/build-locales
@@ -14,10 +14,9 @@ const glob = require('glob');
 const shelljs = require('shelljs');
 const chalk = require('chalk');
 const toSource = require('tosource');
-const oneLine = require('common-tags').oneLine;
+const { oneLine } = require('common-tags');
 
 const appName = config.get('appName');
-const momentLangMap = config.get('momentLangMap');
 
 if (!appName) {
   console.log(chalk.red('Please specify the appName with NODE_APP_INSTANCE'));
@@ -27,6 +26,24 @@ if (!appName) {
 const localeDir = path.join(__dirname, '../locale');
 const poFiles = glob.sync(`${localeDir}/**/${appName}.po`);
 const dest = path.join(__dirname, '../src/locale/');
+
+const getLocaleModulePath = (locale, { _config = config } = {}) => {
+  const momentLangMap = _config.get('momentLangMap');
+
+  const momentLocale = locale.replace('_', '-').toLowerCase();
+  const mappedLocale = momentLangMap[momentLocale];
+
+  let localeModulePath = `moment/locale/${momentLocale}`;
+
+  if (mappedLocale) {
+    localeModulePath = `moment/locale/${mappedLocale}`;
+
+    console.info(oneLine`${locale} is mapped to ${mappedLocale} in
+      configuration, therefore ${mappedLocale} will be used.`);
+  }
+
+  return localeModulePath;
+};
 
 poFiles.forEach((pofile) => {
   const dir = path.dirname(pofile);
@@ -50,17 +67,7 @@ poFiles.forEach((pofile) => {
   if (appName === 'amo') {
     var defineLocale = null;
     try {
-      const momentLocale = locale.replace('_', '-').toLowerCase();
-      const mappedLocale = momentLangMap[momentLocale];
-
-      let localeModulePath = `moment/locale/${momentLocale}`;
-
-      if (mappedLocale) {
-        localeModulePath = `moment/locale/${mappedLocale}`;
-
-        console.info(oneLine`${locale} is mapped to ${mappedLocale} in
-          configuration, therefore ${mappedLocale} will be used.`);
-      }
+      const localeModulePath = getLocaleModulePath(locale);
 
       // Check for the locale first; if it doesn't exist we don't have
       // a moment locale that matches.
@@ -92,3 +99,7 @@ poFiles.forEach((pofile) => {
   fs.writeFileSync(
     localeOutputFilePath, `module.exports = ${toSource(localeObject)}`);
 });
+
+module.exports = {
+  getLocaleModulePath,
+};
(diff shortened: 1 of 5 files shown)


config/default.js
```diff
@@ -242,6 +242,19 @@ module.exports = {
...
+  // See: https://github.com/mozilla/addons-frontend/issues/1515
+  momentLangMap: {
+    'bn-bd': 'bn',
+    'fy-nl': 'fy',
+    'nb-no': 'nb',
+    'nn-no': 'nn',
+    'pt-pt': 'pt',
+    'sv-se': 'sv',
+    dbl: 'x-pseudo',
+    dbr: 'x-pseudo',
```
rebmullin: given that these are being removed in #6624, do we need to include dbl and dbr here?

willdurand: It depends on which PR lands first 😋

config/default.js
```diff
@@ -240,6 +240,17 @@ module.exports = {
   },
   rtlLangs: ['ar', 'fa', 'he', 'ur'],
   defaultLang: 'en-US',
+  // Mapping for missing moment locales that we can still map to existing ones.
```
rebmullin: would this be a little clearer ~ Some missing moment locales can be mapped to existing ones.

tests/unit/test_config.js
```diff
@@ -27,4 +28,18 @@ describe(__filename, () => {
...
+  describe('momentLangMap', () => {
+    it('should not map locales provided by upstream', () => {
+      const momentLangMap = config.get('momentLangMap');
+
+      Object.keys(momentLangMap).forEach((locale) => {
+        // If it does not throw, it means the file likely exists.
+        expect(() => {
+          // eslint-disable-next-line import/no-dynamic-require
+          require(`moment/locale/${locale}.js`);
+        }).toThrow();
```
rebmullin: I am just not sure I understand this, should this throw on an invalid file? When I changed this to be a random file (that doesn't exist), it didn't throw an error so just double checking this..

bin/build-locales
```diff
@@ -50,7 +51,17 @@ poFiles.forEach((pofile) => {
     var defineLocale = null;
     try {
       const momentLocale = locale.replace('_', '-').toLowerCase();
-      const localeModulePath = `moment/locale/${momentLocale}`;
```
kumar303: If you replace this with `getLocaleModulePath()` then you could add unit tests for it. I think it would be helpful to have tests for such a function.

tests/unit/test_config.js
willdurand: It is expected to throw, otherwise it means the locale file exists.

For example, if I change the test case like this:

```diff
diff --git a/tests/unit/test_config.js b/tests/unit/test_config.js
index 4a7030bc2..366b960ac 100644
--- a/tests/unit/test_config.js
+++ b/tests/unit/test_config.js
@@ -33,7 +33,8 @@ describe(__filename, () => {
     it('should not map locales provided by upstream', () => {
       const momentLangMap = config.get('momentLangMap');
 
-      Object.keys(momentLangMap).forEach((locale) => {
+      //Object.keys(momentLangMap).forEach((locale) => {
+      ['fr'].forEach((locale) => {
         // If it does not throw, it means the file likely exists.
         expect(() => {
           // eslint-disable-next-line import/no-dynamic-require

```

I get this error:

```
 FAIL  tests/unit/test_config.js
  ● /Users/williamdurand/projects/mozilla/addons-frontend/tests/unit/test_config.js › momentLangMap › should not map locales provided by upstream

    expect(function).toThrow(undefined)

    Expected the function to throw an error.
    But it didn't throw anything.

      40 |           // eslint-disable-next-line import/no-dynamic-require
      41 |           require(`moment/locale/${locale}.js`);
    > 42 |         }).toThrow();
         |            ^
      43 |       });
      44 |     });
      45 |   });

      at tests/unit/test_config.js:42:12
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (tests/unit/test_config.js:37:14)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 2 passed, 3 total
```

bin/build-locales
willdurand: yep, good call. let's start testing those scripts.

bin/build-locales
```diff
@@ -27,6 +27,24 @@ const localeDir = path.join(__dirname, '../locale');
 const poFiles = glob.sync(`${localeDir}/**/${appName}.po`);
 const dest = path.join(__dirname, '../src/locale/');
 
+const getLocaleModulePath = (locale, { _config = config } = {}) => {
```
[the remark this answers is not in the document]
willdurand: We don't import code from `src` in the bin scripts, but maybe we should?

tests/unit/bin/test_build-locales.js
```diff
@@ -0,0 +1,23 @@
+import { getLocaleModulePath } from 'bin/build-locales';
```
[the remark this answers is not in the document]
willdurand: we could configure webpack/node to create modules and use `@amo` instead of `amo`.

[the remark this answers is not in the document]
willdurand: (or just `amo`)

bin/build-locales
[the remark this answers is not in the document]
kumar303: I didn't think of that. I guess we could.

tests/unit/bin/test_build-locales.js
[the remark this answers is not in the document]
kumar303: If this would allow us to stop configuring module paths in four different places then I like this idea! I filed https://github.com/mozilla/addons-frontend/issues/6677

Issue #6612, opened 2018-10-12
Delete and Cancel buttons are not separated


ioanarusiczki:
Followup for https://github.com/mozilla/addons-frontend/issues/6458

STR:
1. Load AMO dev login, go to an add-ons detail page and make a review
2. Open the detailed reviews page
3. Try to delete the review - observe the Delete and Cancel

Actual result:
Delete and Cancel are one next the other - look attached

Expected result:
There's some space between the two buttons -the way the[ mocks](https://github.com/mozilla/addons-frontend/pull/6567) show it

Notes:
This issue is reproducible on AMO dev with FF62- Win10 & Android 8.0
Also reproducible for developer replies

Desktop display
![buttons on desktop](https://user-images.githubusercontent.com/33448286/46874625-43798180-ce42-11e8-9792-f4c6b6b01455.png)

Android display
![android buttons](https://user-images.githubusercontent.com/33448286/46874661-5ab86f00-ce42-11e8-86de-b6725e8bb9d3.png)

SeanPrashad:
Hey @ioanarusiczki 😁! I'd love to pick this up if that's ok

kumar303:
Hi @SeanPrashad thanks for your interest. Not all issues are available to work on because they may be in our critical path or may have a different priority level. If you want to work on ratings features, I have recently added more to the [`contrib: welcome`](https://github.com/mozilla/addons-frontend/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3A%22component%3A+add-on+ratings%22++label%3A%22contrib%3A+welcome%22) label.

SeanPrashad:
Thanks @kumar303 - I'll take a peek there

kumar303:
Thanks for the report, @ioanarusiczki . I used the newer [multi-column feature](https://caniuse.com/#feat=multicolumn) which is supposedly supported by Firefox 62 but I guess not. I'll think of a different solution.

ioanarusiczki:
@kumar303 It's fixed now. I verified it on AMO dev with FF62 (Win10 & Android 8.0)

Desktop

![buttons desktop](https://user-images.githubusercontent.com/33448286/47007481-f991e380-d140-11e8-9802-ec666a911259.png)

Android 

![buttons android](https://user-images.githubusercontent.com/33448286/47007269-85efd680-d140-11e8-9d77-ac5c7d418995.png)

PR #6618, opened 2018-10-12
Use plain ol' margin-ing instead of row-gap

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6612 with a more direct margin approach because [can I use `row-gap`](https://caniuse.com/#feat=multicolumn) is full of lies! Browsers are hard.

**Before**

screenshot 2018-10-12 17 32 11
screenshot 2018-10-12 17 39 57

**After**

screenshot 2018-10-12 17 32 29
screenshot 2018-10-12 17 33 04
screenshot 2018-10-15 09 29 42

screenshot 2018-10-15 09 27 41

### DIFF
--- a/src/amo/components/AddonReviewCard/styles.scss
+++ b/src/amo/components/AddonReviewCard/styles.scss
@@ -189,9 +189,12 @@
   flex-direction: column;
   flex-wrap: wrap;
 
-  .ConfirmationDialog-confirm-button,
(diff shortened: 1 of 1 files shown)


kumar303:
whoops, I remember why I used `row-gap` now. The responsive mobile view makes this tricky.

kumar303:
Thanks for the review

PR #6620, opened 2018-10-13
Do not render report flag for reviews by the current user

### LINKED ISSUE #6416: "Flag" should be removed from the profile page
STR:
1. Load AMO dev and go to the user profile page, make sure you have some reviews
2. Observe the buttons displayed

Expected result:
Edit reviews and Delete review are available.

Actual result:
Flag is present too - it has no functionality since the user cannot flag his own reviews (or developer responses)

Notes:
Issue is reproducible on AMO dev with FF62 -Win10

own reviews cannot be flagged

SeanPrashad:
Fixes #6416: "Flag" should be removed from the profile page

This PR prevents the "flag" button from being rendered for reviews made by the current user. Essentially, if Sean leaves a review, he should not be allowed to flag his own comment.

**Before:**

before_removing_flag

**After:**

after_removing_flag

### DIFF
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -481,7 +481,8 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </a>
         ) : null}
 
-        {flaggable && review ? (
+        {/* Do not render the "flag" button for replies made by the current user */}
+        {flaggable && review && siteUser && siteUser.id !== review.userId ? (
           <FlagReviewMenu
             isDeveloperReply={this.isReply()}
             openerClass="AddonReviewCard-control"
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -423,6 +423,11 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       (review.userId === siteUser.id ||
         (this.isReply() && this.siteUserCanManageReplies()));
 
+    /* Do not render the "flag" button for reviews made by the current user */
+    const showFlagButton = !siteUser
+      ? true
+      : !(review && siteUser.id === review.userId);
+
     const controls = controlsAreVisible ? (
       <div className="AddonReviewCard-allControls">
         {review && showEditControls ? (
@@ -481,8 +486,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </a>
         ) : null}
 
-        {/* Do not render the "flag" button for replies made by the current user */}
-        {flaggable && review && siteUser && siteUser.id !== review.userId ? (
+        {flaggable && review && showFlagButton ? (
           <FlagReviewMenu
             isDeveloperReply={this.isReply()}
             openerClass="AddonReviewCard-control"
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -423,11 +423,6 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       (review.userId === siteUser.id ||
         (this.isReply() && this.siteUserCanManageReplies()));
 
-    /* Do not render the "flag" button for reviews made by the current user */
-    const showFlagButton = !siteUser
-      ? true
-      : !(review && siteUser.id === review.userId);
-
     const controls = controlsAreVisible ? (
       <div className="AddonReviewCard-allControls">
         {review && showEditControls ? (
@@ -486,7 +481,8 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </a>
         ) : null}
 
-        {flaggable && review && showFlagButton ? (
+        {/* Do not render the "flag" button for reviews made by the current user */}
+        {flaggable && review && (!siteUser || siteUser.id !== review.userId) ? (
           <FlagReviewMenu
             isDeveloperReply={this.isReply()}
             openerClass="AddonReviewCard-control"
--- a/tests/unit/amo/components/TestAddonReviewCard.js
+++ b/tests/unit/amo/components/TestAddonReviewCard.js
@@ -554,7 +554,7 @@ describe(__filename, () => {
   });
 
   it('lets you flag a review', () => {
-    const review = createReviewAndSignInAsUnrelatedUser();
+    const review = _setReview(fakeReview);
     const root = render({ review });
 
     const flag = renderControls(root).find(FlagReviewMenu);
@@ -569,12 +569,11 @@ describe(__filename, () => {
   });
 
   it('lets you flag a developer reply', () => {
-    const review = createReviewAndSignInAsUnrelatedUser();
-    review.isDeveloperReply = true;
-    const root = render({ review });
+    const { reply } = _setReviewReply();
+    const root = renderReply({ reply });
 
     const flag = renderControls(root).find(FlagReviewMenu);
-    expect(flag).toHaveProp('review', review);
+    expect(flag).toHaveProp('review', reply);
     expect(flag).toHaveProp('isDeveloperReply', true);
   });
 
--- a/tests/unit/amo/components/TestAddonReviewCard.js
+++ b/tests/unit/amo/components/TestAddonReviewCard.js
@@ -554,7 +554,7 @@ describe(__filename, () => {
   });
 
   it('lets you flag a review', () => {
-    const review = _setReview(fakeReview);
+    const review = createReviewAndSignInAsUnrelatedUser();
     const root = render({ review });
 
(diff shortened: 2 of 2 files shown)


SeanPrashad:
For the failing tests, it looks like `siteUser` is not set for some of them since they are stubbed and I'm trying to figure out how to properly set it without touching too much!

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -562,11 +569,12 @@ describe(__filename, () => {
   });
 
   it('lets you flag a developer reply', () => {
-    const { reply } = _setReviewReply();
-    const root = renderReply({ reply });
+    const review = createReviewAndSignInAsUnrelatedUser();
```
SeanPrashad: This test was giving me some trouble - not sure if this is the route for us to take. Feedback greatly appreciated!

kumar303:
I think the direction of this patch is good -- hiding the flag menu for all screens. I'd rather implement this feature consistently rather than just for the user profile page. I personally don't like hidden UI controls because it can be confusing to know what state you're in but everyone else argued against that so I concede 😁

src/amo/components/AddonReviewCard/index.js
```diff
@@ -481,7 +481,8 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </a>
         ) : null}
 
-        {flaggable && review ? (
+        {/* Do not render the "flag" button for replies made by the current user */}
+        {flaggable && review && siteUser && siteUser.id !== review.userId ? (
```
kumar303: If the user is signed out, the *Flag* link should always be visible. Otherwise, if someone wants to flag a review, they might not know that they have to sign in first.

This is what it looks like (there is actually a style regression that I will fix right now! https://github.com/mozilla/addons-frontend/issues/6656)

screenshot 2018-10-16 11 45 43

SeanPrashad:
I've updated the logic to **always** render the flag button if the user isn't logged in:

post_feedback_from_kumar

src/amo/components/AddonReviewCard/index.js
```diff
@@ -481,7 +486,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
           </a>
         ) : null}
 
-        {flaggable && review ? (
+        {flaggable && review && showFlagButton ? (
```
kumar303: I liked how simple this was in your first patch. It was a lot easier to read. What about this?

```suggestion
        {flaggable && review && (!siteUser || siteUser.id !== review.userId) ? (
```

kumar303: You also need to add a test that makes sure the flag button is not hidden when the user is signed out. After writing the test, you can comment out the part that allowed non-signed in users to see the flag link; the test should fail. This helps you know that the test is working.

tests/unit/amo/components/TestAddonReviewCard.js
kumar303: > This test was checking a developer's reply to an existing review. I think I changed the logic behind it - would appreciate feedback.

Why did you need to change the test? I put it back to how it was and it passed for me. Maybe it was related to not being signed in?

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -547,7 +554,7 @@ describe(__filename, () => {
   });
 
   it('lets you flag a review', () => {
-    const review = _setReview(fakeReview);
```
kumar303: I think you can put this back now, too. The test passes for me.

src/amo/components/AddonReviewCard/index.js
```diff
@@ -423,6 +423,11 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       (review.userId === siteUser.id ||
         (this.isReply() && this.siteUserCanManageReplies()));
 
+    /* Do not render the "flag" button for reviews made by the current user */
```
kumar303: Prefix comments with `//` (double slashes) when they don't need to span multiple lines.

SeanPrashad:
Thanks for the great feedback @kumar303! ✌🏽I've reverted the other tests back and added a new test to [check for the flag button when the user wasn't authenticated](https://github.com/mozilla/addons-frontend/pull/6620/files#diff-472cf21dd9e14bab7babf939d647df8dR586). I think this should be good to 🚢

kumar303:
> Not sure if it's just me, but I can't reply to the third comment you made for your review

It's not just you. Github is buggy. I copied/pasted the quote of what I was replying to because I had actually pressed the reply button on your last comment and I knew Github is buggy around this behavior (for some reason). I guess they broke it further.

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -576,6 +583,21 @@ describe(__filename, () => {
     expect(root.find(FlagReviewMenu)).toHaveLength(0);
   });
 
+  it('renders the flag button when not signed in', () => {
+    const root = render({ review: _setReview(fakeReview) });
+
+    expect(renderControls(root).find(FlagReviewMenu)).toHaveLength(1);
+  });
+
+  it('hides the flag button if you wrote the review', () => {
```
kumar303: When I undo your patch, this test should fail but it does not. Try to make it fail.

Here's how I undid your patch:

```diff
diff --git a/src/amo/components/AddonReviewCard/index.js b/src/amo/components/AddonReviewCard/index.js
index 13013647a..ff535434a 100644
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -482,7 +482,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
         ) : null}
 
         {/* Do not render the "flag" button for reviews made by the current user */}
-        {flaggable && review && (!siteUser || siteUser.id !== review.userId) ? (
+        {flaggable && review ? (
           <FlagReviewMenu
             isDeveloperReply={this.isReply()}
             openerClass="AddonReviewCard-control"
```

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -576,6 +583,21 @@ describe(__filename, () => {
     expect(root.find(FlagReviewMenu)).toHaveLength(0);
   });
 
+  it('renders the flag button when not signed in', () => {
```
kumar303: Actually, I just realized that you don't need this test because we already have one:

https://github.com/mozilla/addons-frontend/blob/61424ad5a9a200c5abfe3dfda013c1fd916217fa/tests/unit/amo/components/TestAddonReviewCard.js#L556

However, I would suggest changing the spec from 'lets you flag a review' to 'shows FlagReviewMenu when signed out'

tests/unit/amo/components/TestAddonReviewCard.js
SeanPrashad: Ok - here's how I think we should proceed:

The current test:

* Signs the user in
* Creates a review
* Checks that the flag button isn't rendered

The ideal test:

* Signs the user in
* Creates a review
* Checks that the flag button isn't rendered
* **Signs the user out**
* **Checks that the flag button is rendered**

SeanPrashad:
Your tip made me realize that I forgot to check if the flag button would be rendered _after_ the user logged out! I've updated this in the test and removed the old helper function I had written.

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -576,6 +576,21 @@ describe(__filename, () => {
     expect(root.find(FlagReviewMenu)).toHaveLength(0);
   });
 
+  it('hides the flag button if you wrote the review', () => {
+    const originalReviewId = 123;
+    const review = signInAndDispatchSavedReview({
+      siteUserId: originalReviewId,
```
kumar303: This variable name is confusing but I see what you are trying to do. You are letting the helper take over to set `siteUserId` and `reviewUserId` to the same value. Instead of letting the helper do this implicitly, your test should explicitly set these both to the same value so that the intention is clear.

SeanPrashad: Yeah, I didn't want to introduce another helper that wrapped the `signInAndDispatchSavedReview()` helper so I just used it directly. I've updated the variable naming to `siteUserId` so it hopefully helps us recognize that this is an authenticated user.

tests/unit/amo/components/TestAddonReviewCard.js
```diff
@@ -576,6 +576,21 @@ describe(__filename, () => {
...
+    });
+    let root = render({ review });
+
+    expect(renderControls(root).find(FlagReviewMenu)).toHaveLength(0);
+
+    store.dispatch(logOutUser());
```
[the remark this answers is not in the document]
SeanPrashad: ✅  All done

PR #6625, opened 2018-10-15
Remove debug locales (`dbr` and `dbl`)

### LINKED ISSUE #6624: Remove debug locales (`dbr` and `dbl`)
We build "debug" locales for testing/QA purpose but it seems no one is using them and we get fast feedback from the localizers when something is not translatable anyway.

We should remove the code related to debug locales and simplify the i18n extraction.

willdurand:
Fixes #6624


muffinresearch:
more-red-than-green-award-sm

PR #6616, opened 2018-10-12
Create an addon-summary card to use on the Review Listing and All Versions pages

bobsilverberg:
Fixes #6605

PR #6630, opened 2018-10-15
Use Link component in AddonTitle

### LINKED ISSUE #6628: Use react-router for author links in an add-on detail page
See also: #6526 #6626 

---

Once #6626 lands, we can update its internals to use the `Link` component for each author link, instead of using hardcoded HTML. This is similar to what has been done in the AddonReviewCard (for the permalink with the timestamp).

willdurand:
Fixes #6628

---

This PR brings react router into the URLs of each author of an add-on, on the add-on detail pages. The list of authors is now fully translatable!

### RTL

screen shot 2018-10-15 at 16 23 26

### LTR

screen shot 2018-10-15 at 16 23 36

### DIFF
--- a/src/amo/components/AddonTitle/index.js
+++ b/src/amo/components/AddonTitle/index.js
@@ -1,8 +1,8 @@
 /* @flow */
 import * as React from 'react';
 
+import Link from 'amo/components/Link';
 import translate from 'core/i18n/translate';
-import { sanitizeHTML } from 'core/utils';
 import LoadingText from 'ui/components/LoadingText';
 import type { AddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
@@ -19,46 +19,44 @@ type InternalProps = {|
 |};
 
 export const AddonTitleBase = ({ addon, i18n }: InternalProps) => {
-  let children;
-  let htmlTitle;
+  const authors = [];
 
-  if (addon) {
-    let title;
-    if (!addon.authors) {
-      title = addon.name;
-    } else {
-      const authorList = addon.authors.map((author) => {
-        if (author.url) {
-          return `<a href="${author.url}">${author.name}</a>`;
-        }
+  if (addon && addon.authors) {
+    const nbAuthors = addon.authors.length;
 
-        return author.name;
-      });
-
-      title = i18n.sprintf(
-        // translators: Example: The Add-On <span>by The Author</span>
-        i18n.gettext('%(addonName)s %(startSpan)sby %(authorList)s%(endSpan)s'),
-        {
-          addonName: addon.name,
-          authorList: authorList.join(', '),
-          startSpan: '<span class="AddonTitle-author">',
-          endSpan: '</span>',
-        },
+    addon.authors.forEach((author, index) => {
+      authors.push(
+        author.url ? (
+          <Link key={author.id} to={`/user/${author.username}/`}>
+            {author.name}
+          </Link>
+        ) : (
+          author.name
+        ),
       );
-    }
 
-    htmlTitle = sanitizeHTML(title, ['a', 'span']);
-  } else {
-    children = <LoadingText width={70} />;
+      if (index + 1 < nbAuthors) {
+        authors.push(', ');
+      }
+    });
   }
 
   return (
-    <h1
-      className="AddonTitle"
-      // eslint-disable-next-line react/no-danger
-      dangerouslySetInnerHTML={htmlTitle}
-    >
-      {children}
+    <h1 className="AddonTitle">
+      {addon ? (
+        <React.Fragment>
+          {addon.name}
+          {authors.length > 0 && (
+            <span className="AddonTitle-author">
+              {// translators: Example: "add-on" by "some authors"
+              i18n.gettext('by')}{' '}
+              {authors}
+            </span>
+          )}
+        </React.Fragment>
+      ) : (
+        <LoadingText width={70} />
+      )}
     </h1>
   );
 };
--- a/src/amo/components/AddonTitle/index.js
+++ b/src/amo/components/AddonTitle/index.js
@@ -1,9 +1,13 @@
 /* @flow */
 import * as React from 'react';
+import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


willdurand:
As per our discussion, we should wait for #6616 first.

src/amo/components/AddonTitle/index.js
```diff
@@ -19,46 +19,44 @@ type InternalProps = {|
...
-    >
-      {children}
+    <h1 className="AddonTitle">
+      {addon ? (
+        <React.Fragment>
+          {addon.name}
+          {authors.length > 0 && (
+            <span className="AddonTitle-author">
+              {// translators: Example: "add-on" by "some authors"
+              i18n.gettext('by')}{' '}
```
bobsilverberg: This doesn't seem like it will work for l10n. You are dictating that a space will always follow the translated version of `by` but we cannot know that will always be correct.

tests/unit/amo/components/TestAddonTitle.js
```diff
@@ -26,54 +26,64 @@ describe(__filename, () => {
...
 
-    expect(root.html()).toContain(name);
+    expect(root).toIncludeText(name);
   });
 
   it('renders a single author', () => {
-    const authorUrl = 'http://olympia.test/en-US/firefox/user/krupa/';
+    const author = {
+      ...fakeAddon.authors[0],
+      username: 'some-username',
```
bobsilverberg: You don't technically need to specify a specific `username` in the test setup because you are never referring to this later. You just use `author.username` in your assertion, so it doesn't really matter what the `username` is. You don't specify an author name here, so why specify a `username`?

tests/unit/amo/components/TestAddonTitle.js
```diff
@@ -26,54 +26,64 @@ describe(__filename, () => {
...
+      `/user/${author1.username}/`,
+    );
+    expect(root.find(Link).at(1)).toHaveProp('children', author2.name);
+    expect(root.find(Link).at(1)).toHaveProp(
+      'to',
+      `/user/${author2.username}/`,
+    );
+    // This assertion is used to make sure we add commas between each author
+    // link/name.
+    expect(root).toIncludeText(
```
bobsilverberg: This seems a bit fragile. I think we try to avoid using enzyme output in assertions. Could you just check that there is a `, ` in the string somewhere? Maybe check that there is only one?

src/amo/components/AddonTitle/index.js
```diff
@@ -19,46 +19,44 @@ type InternalProps = {|
...
-    let title;
-    if (!addon.authors) {
-      title = addon.name;
-    } else {
-      const authorList = addon.authors.map((author) => {
-        if (author.url) {
-          return `<a href="${author.url}">${author.name}</a>`;
-        }
+  if (addon && addon.authors) {
+    const nbAuthors = addon.authors.length;
```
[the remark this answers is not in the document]
willdurand: Flow complains if I don't do this.. (because it wants me to re-check `addon.authors`)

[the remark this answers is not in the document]
willdurand: I changed the var name to `authorCount`

src/amo/components/AddonTitle/index.js
willdurand: That's a good call, because first time I tried in RTL mode, I thought it worked, but that does not quite work.

tests/unit/amo/components/TestAddonTitle.js
willdurand: nevermind, I found a better way!

willdurand:
The patch got a bit bigger because I did not handle RTL mode correctly as you pointed out 🥇 But now the PR got screenshots :D and the content is fully translatable.

src/amo/components/AddonTitle/index.js
```diff
@@ -16,53 +20,73 @@ type Props = {|
...
+            <span className="AddonTitle-author">
+              {isRTL ? (
+                <React.Fragment>
+                  {authors}{' '}
+                  {// translators: Example: add-on "by" some authors
+                  i18n.gettext('by')}
+                </React.Fragment>
+              ) : (
+                <React.Fragment>
+                  {i18n.gettext('by')} {authors}
```
bobsilverberg: This should probably have a translators hint as well.

Having said that, are you sure this will work for all languages? You've accounted for RTL, but is it possible that some language wouldn't use a space as a separator? You're still dictating what will appear between the "by" and the list of authors and I'm a bit concerned that that still might be an issue. I feel like we need to have them translate the full "by " (i.e., "by" plus a space), but I could be wrong about that. I realize it probably will make the code trickier to allow for that.

willdurand: Those spaces are "UI" spaces and not "writing" spaces if that makes sense. We need an empty space between text and components/HTML. I could have used `&nbsp;` but decided to only use a space.

willdurand: > I feel like we need to have them translate the full "by " (i.e., "by" plus a space), but I could be wrong about that

That was my initial idea but `gettext()` does not let you do that anyway 😞 It is still an improvement over what we have right now though

willdurand: > This should probably have a translators hint as well.

I don't know because the other one has a hint already. So I thought it would create only one entry, and not two, but sure I can add a comment.

bobsilverberg:
@willdurand Just wondering, is this the last change you're planning to make to `AddonTitle` or should I wait for another one before continuing my work on #6616? Also, are you waiting on anything from me on this one?

willdurand:
> Just wondering, is this the last change you're planning to make to `AddonTitle` or should I wait for another one before continuing my work on #6616?

I think I am done with this component. The next change will be to use `userId` instead of `username` in author URLs.

> Also, are you waiting on anything from me on this one?

A final review/approval? 🙏

bobsilverberg:
> A final review/approval?

You didn't flag it for a re-review so I wasn't sure if you were done with it or not. I'll look at it now.

PR #6632, opened 2018-10-15
Add date and last-modified meta tags

### LINKED ISSUE #5916: Add creation and update dates to meta tags in detail pages
The `<head>` of detail pages should include the following tags:

`<meta name="date" content="2018-07-24T16:01:00+0200" />`
`<meta name="last-modified" content="2018-07-24T16:01:00+0200" />`

The `date` value should correspond to the add-on's creation time, and the `last-modified` value should correspond to the last updated time.

/cc @rraue

willdurand:
Fixes #5916


src/amo/pages/Addon/index.js
```diff
@@ -600,6 +600,8 @@ export class AddonBase extends React.Component {
             <title>{this.getPageTitle()}</title>
             <link rel="canonical" href={addon.url} />
             <meta name="description" content={this.getPageDescription()} />
+            <meta name="date" content={addon.created} />
+            <meta name="last-modified" content={addon.last_updated} />
```
rebmullin: since last_updated can be null would it be better to wrap this in a condition? REF: https://github.com/mozilla/addons-frontend/pull/6632/files#diff-9e6cdc9033004286c61eaefcf496afe2R133

willdurand: I suppose yes, but I don't think `last_updated` can be `null` actually. It seems to be always returned by the API.

@EnTeQuAk can `last_updated` be `null` or empty?

willdurand: As per our IRC discussion, it *might* happen, so let's be defensive here. Thanks @rebmullin, that's an extremely relevant review!

PR #6638, opened 2018-10-15
Store versions by id in the versions reducer state

### LINKED ISSUE #6637: Store versions by id in the versions reducer state
This is to support #6634. To try to reduce the size of that patch, which will be big, update the versions reducer to store versions by id.

bobsilverberg:
Fixes #6637

### DIFF
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -13,10 +13,12 @@ import type {
 export const FETCH_VERSIONS: 'FETCH_VERSIONS' = 'FETCH_VERSIONS';
 export const LOAD_VERSIONS: 'LOAD_VERSIONS' = 'LOAD_VERSIONS';
 
+type VersionIdType = number;
+
 export type AddonVersionType = {
   compatibility?: AddonCompatibilityType,
   platformFiles: PlatformFilesType,
-  id: number,
+  id: VersionIdType,
   license: { name: string, url: string },
   releaseNotes?: string,
   version: string,
@@ -36,12 +38,16 @@ export const createInternalVersion = (
 };
 
 export type VersionsState = {
+  byId: {
+    [id: number]: AddonVersionType,
+  },
   bySlug: {
-    [slug: string]: { versions: Array<AddonVersionType>, loading: boolean },
+    [slug: string]: { versionIds: Array<VersionIdType>, loading: boolean },
   },
 };
 
 export const initialState: VersionsState = {
+  byId: {},
   bySlug: {},
 };
 
@@ -106,7 +112,13 @@ export const getVersionsBySlug = ({
   invariant(state, 'state is required');
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
+  const versionIds = infoForSlug && infoForSlug.versionIds;
+  if (versionIds) {
+    return versionIds.map((versionId) => {
+      return state.byId[versionId];
+    });
+  }
+  return null;
 };
 
 export const getLoadingBySlug = ({ slug, state }: GetBySlugParams): boolean => {
@@ -124,27 +136,50 @@ type VersionInfoType = {|
 
 type GetVersionInfoParams = {|
   _findFileForPlatform?: typeof findFileForPlatform,
+  state: VersionsState,
   userAgentInfo: UserAgentInfoType,
-  version: AddonVersionType,
+  versionId: VersionIdType,
 |};
 
 export const getVersionInfo = ({
   _findFileForPlatform = findFileForPlatform,
+  state,
   userAgentInfo,
-  version,
+  versionId,
 }: GetVersionInfoParams): VersionInfoType | null => {
-  const file = _findFileForPlatform({
-    platformFiles: version.platformFiles,
-    userAgentInfo,
-  });
+  const version = state.byId[versionId];
 
-  if (file) {
-    return {
-      created: file.created,
-      filesize: file.size,
-    };
+  if (version) {
+    const file = _findFileForPlatform({
+      platformFiles: version.platformFiles,
+      userAgentInfo,
+    });
+
+    if (file) {
+      return {
+        created: file.created,
+        filesize: file.size,
+      };
+    }
   }
+
   return null;
+};
+
+type GetVersionByIdParams = {|
+  id: VersionIdType,
+  state: VersionsState,
+|};
+
+export const getVersionById = ({
+  id,
+  state,
+}: GetVersionByIdParams): AddonVersionType | null => {
+  invariant(id, 'id is required');
+  invariant(state, 'state is required');
+
+  const version = state.byId[id];
+  return version || null;
 };
 
 type Action = FetchVersionsAction | LoadVersionsAction;
@@ -161,7 +196,7 @@ const reducer = (
         bySlug: {
           ...state.bySlug,
           [slug]: {
-            versions: null,
+            versionIds: null,
             loading: true,
           },
         },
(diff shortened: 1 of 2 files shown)


bobsilverberg:
@kumar303 This is a small step towards what we were discussing at https://github.com/mozilla/addons-frontend/pull/6520#discussion_r225199455.

src/amo/reducers/versions.js
```diff
@@ -106,7 +112,13 @@ export const getVersionsBySlug = ({
   invariant(state, 'state is required');
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
+  const versionIds = infoForSlug && infoForSlug.versionIds;
+  if (versionIds) {
+    return versionIds.map((versionId) => {
+      return state.byId[versionId];
```
kumar303: This should be `return getVersionById({ id: versionId, state })`

src/amo/reducers/versions.js
```diff
@@ -124,29 +136,52 @@ type VersionInfoType = {|
...
-    platformFiles: version.platformFiles,
-    userAgentInfo,
-  });
-
-  if (file) {
-    return {
-      created: file.created,
-      filesize: file.size,
-    };
+  const version = state.byId[versionId];
```
kumar303: This should use `getVersionById`

src/amo/reducers/versions.js
kumar303: This should throw an error if the version object doesn't exist in state. It would mean the app is severely broken. If you let it return `undefined` (or `null`) then it might create an infinite loading loop depending on the loading code.

bobsilverberg:

src/amo/reducers/versions.js
```diff
@@ -106,7 +128,19 @@ export const getVersionsBySlug = ({
...
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
+  const versionIds = infoForSlug && infoForSlug.versionIds;
+  if (versionIds) {
+    return versionIds.map((versionId) => {
+      const version = getVersionById({ id: versionId, state });
+      if (!version) {
+        throw new Error(
+          `No version was found for slug ${slug} using versionId ${versionId}`,
```
kumar303: The coverage report says this doesn't have a test

src/amo/reducers/versions.js
```diff
@@ -106,7 +128,19 @@ export const getVersionsBySlug = ({
   invariant(state, 'state is required');
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
+  const versionIds = infoForSlug && infoForSlug.versionIds;
```
kumar303: Why is this checking for the presence of `infoForSlug.versionIds` ? The [type](https://github.com/mozilla/addons-frontend/blob/7200d59388557d36d35dbede5b19d492af794e5c/src/amo/reducers/versions.js#L45) says it will always be an array. Does the type need updating?

bobsilverberg: The code was not in sync with the Flow definition, but I fixed that. It now should always be an array, so I can remove the extra check.

src/amo/reducers/versions.js
bobsilverberg: There doesn't actually seem to be a way to create this condition, other than manually monkeying around with the state in the test. That's generally frowned upon, but I don't think we can create a state where this error would occur just by calling reducer methods. I'll add a test to get the coverage, but I'm just trying to preempt the comment about the test.

bobsilverberg:
Thanks @kumar303. It's ready for another look.

src/amo/reducers/versions.js
```diff
@@ -161,7 +201,7 @@ const reducer = (
         bySlug: {
           ...state.bySlug,
           [slug]: {
-            versions: null,
+            versionIds: [],
```
kumar303: Why do you need to initialize this to an empty array? We typically initialize to `null` so the frontend can know for sure that no attempt has been made to fetch the results. In other words, an empty array could be the legitimate result of a fetch.

tests/unit/amo/reducers/test_versions.js
```diff
@@ -89,9 +90,26 @@ describe(__filename, () => {
...
+    it('throws when an expected version is not found for a slug', () => {
+      const slug = 'some-slug';
+      const versions = [fakeVersion];
+      const state = versionsReducer(
+        undefined,
+        loadVersions({ slug, versions }),
+      );
+
+      // We have to manually remove the version in order to trigger the error.
+      delete state.byId[fakeVersion.id];
```
kumar303: I understand this better now. You don't need a test for this. I'll suggest something better than throwing an error.

src/amo/reducers/versions.js
```diff
@@ -101,12 +123,23 @@ type GetBySlugParams = {|
...
+}: GetBySlugParams): Array<AddonVersionType> => {
   invariant(slug, 'slug is required');
   invariant(state, 'state is required');
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
+  if (infoForSlug) {
+    return infoForSlug.versionIds.map((versionId) => {
+      const version = getVersionById({ id: versionId, state });
+      if (!version) {
```
kumar303: I didn't realize that `loadAddons` is guaranteed to [load a version object](https://github.com/mozilla/addons-frontend/blob/e6b1a6d90bec1ddae26ae360fff074c59953a499/src/amo/reducers/versions.js#L221-L224). Because of that, you don't need to throw an error. 

However, I think it would be nice to add `invariant(version, ...)` in case someone refactors the `loadAddons` reducer and accidentally removes the `byId` part. Calls to `invariant()` are always executed which means you don't need to add test coverage for them -- there is no chance of introducing unexpected errors.

src/amo/reducers/versions.js
bobsilverberg: Ah, that's probably why I had it as `null` in the first place. I changed it in the last commit. I'll change it back to `null`.

src/amo/reducers/versions.js
```diff
@@ -101,12 +126,22 @@ type GetBySlugParams = {|
 export const getVersionsBySlug = ({
   slug,
   state,
-}: GetBySlugParams): Array<AddonVersionType> | null => {
+}: GetBySlugParams): Array<AddonVersionType> => {
   invariant(slug, 'slug is required');
   invariant(state, 'state is required');
 
   const infoForSlug = state.bySlug[slug];
-  return (infoForSlug && infoForSlug.versions) || null;
```
kumar303: Why does this no longer return `null` for non-existent versions? It seems like that leaves no way to detect if versions by slug have been loaded or not. I haven't seen the component patches for this so maybe it's fine?

tests/unit/amo/reducers/test_versions.js
```diff
@@ -36,7 +37,7 @@ describe(__filename, () => {
       fetchVersions({ errorHandlerId: 1, slug }),
     );
 
-    expect(getVersionsBySlug({ slug, state })).toBe(null);
+    expect(getVersionsBySlug({ slug, state })).toEqual([]);
```
kumar303: Similar to my comment on the last review, we usually set data to null when fetching. Otherwise, how would you be able to know if the API returned a legitimately empty array of versions?

src/amo/reducers/versions.js
bobsilverberg: You're right - a consumer may want to be able to differentiate between null and no records found. I've changed it back.

tests/unit/amo/reducers/test_versions.js
bobsilverberg: I've changed that back.

PR #6641, opened 2018-10-16
I6531 search result title update

### LINKED ISSUE #6531: Display Category name on Category page
When you go to a category page, for example https://addons.mozilla.org/en-US/firefox/search/?category=download-management&sort=rating&type=extension, you see "531 add-ons found" but no indication that these are add-ons in the Download Management category.

We should change it to say "X add-ons found in $Category" for all category pages.

rebmullin:
fixes #6531

### DIFF
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -5,7 +5,11 @@ import { compose } from 'redux';
 
 import { fetchCategories } from 'core/reducers/categories';
 import translate from 'core/i18n/translate';
-import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEMES_FILTER } from 'core/constants';
+import {
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_STATIC_THEME,
+  ADDON_TYPE_THEMES_FILTER,
+} from 'core/constants';
 import { withErrorHandler } from 'core/errorHandler';
 import Card from 'ui/components/Card';
 import type { AppState } from 'amo/store';
@@ -27,7 +31,7 @@ type InternalProps = {|
   filters: SearchFilters,
   hasCategory: boolean,
   i18n: I18nType,
-  loading: boolean,
+  loadingSearch: boolean,
 |};
 
 export class SearchContextCardBase extends React.Component<InternalProps> {
@@ -42,11 +46,11 @@ export class SearchContextCardBase extends React.Component<InternalProps> {
   }
 
   render() {
-    const { categoryName, count, filters, i18n, loading } = this.props;
+    const { categoryName, count, filters, i18n, loadingSearch } = this.props;
     const { addonType, query } = filters;
     let searchText;
 
-    if (!loading) {
+    if (!loadingSearch) {
       switch (addonType) {
         case ADDON_TYPE_EXTENSION:
           if (categoryName) {
@@ -152,11 +156,11 @@ export class SearchContextCardBase extends React.Component<InternalProps> {
           }
           break;
       }
-    } else if (loading && query) {
+    } else if (loadingSearch && query) {
       searchText = i18n.sprintf(i18n.gettext('Searching for "%(query)s"'), {
         query,
       });
-    } else if (loading) {
+    } else if (loadingSearch) {
       searchText = i18n.gettext('Searching for add-ons');
     }
 
@@ -187,30 +191,30 @@ export function mapStateToProps(state: AppState) {
   if (currentCategory) {
     const categoriesState = state.categories.categories;
     const { clientApp } = state.api;
-    let translatedCategory = {};
 
     if (categoriesState && clientApp) {
-      Object.keys(categoriesState[clientApp]).forEach((type) => {
-        const searchType = categoriesState[clientApp][type];
-        Object.keys(searchType).forEach((category) => {
-          const searchCategory = searchType[category];
-          const { name, slug } = searchCategory;
+      const appTypes = categoriesState[clientApp];
 
-          if (slug === currentCategory) {
-            translatedCategory = {
-              [slug]: {
-                name,
-              },
-            };
+      if (appTypes) {
+        if (
+          filters &&
+          filters.addonType &&
+          typeof filters.addonType === 'string'
+        ) {
+          // eslint-disable-next-line prefer-destructuring
+          let addonType = filters.addonType;
+          if (addonType === ADDON_TYPE_THEMES_FILTER) {
+            addonType = ADDON_TYPE_STATIC_THEME;
           }
-        });
-      });
-    }
 
-    categoryName =
-      translatedCategory && translatedCategory[currentCategory]
-        ? translatedCategory[currentCategory].name
-        : null;
+          const categories = appTypes[addonType];
+
+          if (categories) {
+            categoryName = categories[currentCategory].name;
+          }
+        }
+      }
+    }
   }
 
   return {
@@ -218,7 +222,7 @@ export function mapStateToProps(state: AppState) {
     categoryName,
     count: search.count || 0,
     filters,
-    loading: search.loading,
+    loadingSearch: search.loading,
   };
 }
 
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -1,5 +1,4 @@
 /* @flow */
-// import PropTypes from 'prop-types';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
--- a/src/amo/components/SearchContextCard/index.js
+++ b/src/amo/components/SearchContextCard/index.js
@@ -79,92 +79,6 @@ export class SearchContextCardBase extends React.Component {
             );
           }
           break;
-        case ADDON_TYPE_DICT:
-          if (category) {
-            const categoryName = category.replace('-', ' ');
-
-            if (query) {
-              searchText = i18n.sprintf(
-                i18n.ngettext(
-                  '%(count)s result found for "%(query)s" in %(categoryName)s',
-                  '%(count)s results found for "%(query)s" in %(categoryName)s',
-                  count,
-                ),
-                { count: i18n.formatNumber(count), query, categoryName },
-              );
-            } else {
-              searchText = i18n.sprintf(
-                i18n.ngettext(
-                  '%(count)s result found in %(categoryName)s',
-                  '%(count)s results found in %(categoryName)s',
-                  count,
-                ),
-                { count: i18n.formatNumber(count), categoryName },
-              );
-            }
-          } else if (query) {
-            searchText = i18n.sprintf(
-              i18n.ngettext(
-                '%(count)s result found for "%(query)s"',
-                '%(count)s results found for "%(query)s"',
-                count,
-              ),
-              { count: i18n.formatNumber(count), query },
-            );
-          } else {
-            searchText = i18n.sprintf(
-              i18n.ngettext(
-                '%(count)s result found',
-                '%(count)s results found',
-                count,
-              ),
-              { count: i18n.formatNumber(count) },
-            );
-          }
-          break;
-        case ADDON_TYPE_LANG:
-          if (category) {
-            const categoryName = category.replace('-', ' ');
-
-            if (query) {
-              searchText = i18n.sprintf(
-                i18n.ngettext(
-                  '%(count)s result found for "%(query)s" in %(categoryName)s',
-                  '%(count)s results found for "%(query)s" in %(categoryName)s',
-                  count,
-                ),
-                { count: i18n.formatNumber(count), query, categoryName },
-              );
-            } else {
-              searchText = i18n.sprintf(
-                i18n.ngettext(
-                  '%(count)s result found in %(categoryName)s',
-                  '%(count)s results found in %(categoryName)s',
-                  count,
-                ),
-                { count: i18n.formatNumber(count), categoryName },
-              );
-            }
-          } else if (query) {
-            searchText = i18n.sprintf(
-              i18n.ngettext(
-                '%(count)s result found for "%(query)s"',
-                '%(count)s results found for "%(query)s"',
-                count,
-              ),
-              { count: i18n.formatNumber(count), query },
-            );
-          } else {
-            searchText = i18n.sprintf(
-              i18n.ngettext(
-                '%(count)s result found',
-                '%(count)s results found',
-                count,
-              ),
-              { count: i18n.formatNumber(count) },
-            );
-          }
-          break;
         case ADDON_TYPE_THEMES_FILTER:
           if (category) {
             const categoryName = category.replace('-', ' ');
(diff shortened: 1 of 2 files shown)


src/amo/components/SearchContextCard/index.js
```diff
@@ -29,67 +29,121 @@ export class SearchContextCardBase extends React.Component {
...
-            '%(count)s results found for "%(query)s"',
-            count,
-          ),
-          { count: i18n.formatNumber(count), query },
-        );
+    if (!loading) {
+      switch (addonType) {
+        case ADDON_TYPE_EXTENSION:
+          if (category) {
+            const categoryName = category.replace('-', ' ');
```
willdurand: We cannot do this unfortunately. We need translated strings, which means a lot of new strings and/or new API calls to load the categories.

rebmullin: thanks for the quick feedback. I'll revisit this one...

rebmullin: actually, quick question on this...

doesn't sprintf in combination with i18n.ngettext translate this? I am not clear why this won't work :/

willdurand: That's because the category name needs to be translated too

kumar303:
Hi @rebmullin. I started looking at this but when I merged it into master, I got this error:

```
[1544824787994271700] ERROR (amo/18292 on indigo.local):
[1]     err: {
[1]       "type": "Error",
[1]       "message": "Cannot find module 'core/actions/categories'",
[1]       "stack":
[1]               ...
[1]               at Object.<anonymous> (/Users/kumar/dev/addons-frontend/src/amo/components/SearchContextCard/index.js:6:1)
```

It looks like that got removed, maybe to combine it with `reducers/categories`?

rebmullin:
ahh yes @kumar303.. this is very outdated I guess. I'll update this! Thanks for the tip:)

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +179,72 @@ export class SearchContextCardBase extends React.Component {
...
+
   return {
-    count: state.search.count,
-    filters: state.search.filters,
-    loading: state.search.loading,
+    hasCategory: !!currentCategory,
+    categoryName,
+    count: search.count,
+    filters,
+    loading: search.loading,
```
bobsilverberg: Shouldn't this be the `loading` value from the `categories` state, not the `search` state, or do you perhaps need both? After looking at the tests, it seems like at least some of the time you are concerned with whether the `categories` are loading, but it looks like other times you might be concerned with whether the search results are loading.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -111,162 +134,393 @@ describe(__filename, () => {
...
+    const { store } = dispatchSearchResults({
+      filters: {
+        addonType: ADDON_TYPE_THEMES_FILTER,
+        category: 'causes',
         query: 'test',
       },
     });
 
+    _fetchCategories({ store });
+    _loadCategories({ store });
```
bobsilverberg: I mentioned this in the code for the component as well, but I find this confusing, this test says "when loading is false", and the categories are not loading because you have fetched and loaded them, but the `loading` that is being used in the component is from the `search` state, not from the `categories` state.

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -111,162 +134,393 @@ describe(__filename, () => {
...
+    });
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header')).toIncludeText(
+      '2 themes found',
     );
   });
 
   it('should render singular form when only one result is found with addonType ADDON_TYPE_DICT', () => {
```
bobsilverberg: Do we still need tests specifically for `ADDON_TYPE_DICT` when that is no longer being specifically dealt with in the code? Should we instead just have a set of tests for when the type is neither extension nor theme?

tests/unit/amo/components/TestSearchContextCard.js
```diff
@@ -111,162 +134,393 @@ describe(__filename, () => {
...
+    });
+
+    const root = render({ store });
+
+    expect(root.find('.SearchContextCard-header')).toIncludeText(
+      '2 extensions found',
     );
   });
 
   it('should render singular form when only one result is found with addonType ADDON_TYPE_LANG', () => {
```
bobsilverberg: As mentioned above, do we still need tests specifically for `ADDON_TYPE_LANG` when that is no longer being specifically dealt with in the code? Should we instead just have a set of tests for when the type is neither extension nor theme?

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +179,72 @@ export class SearchContextCardBase extends React.Component {
...
+    const { clientApp } = state.api;
+    const allCategories = [];
+
+    if (categoriesState && clientApp) {
+      Object.keys(categoriesState[clientApp]).forEach((type) => {
+        const searchType = categoriesState[clientApp][type];
+        Object.keys(searchType).forEach((category) => {
+          const searchCategory = searchType[category];
+          const { slug } = searchCategory;
+          const { name } = searchCategory;
```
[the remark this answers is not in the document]
rebmullin: yeh that's weird that I did it this way lol

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +179,72 @@ export class SearchContextCardBase extends React.Component {
...
+        });
+      });
+    }
+
+    const translatedCategory =
+      allCategories.length &&
+      allCategories.find(
+        (category) =>
+          category[currentCategory] &&
+          category[currentCategory].slug === currentCategory,
```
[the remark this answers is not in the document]
rebmullin: It seems that I was unnecessarily looping through categories 2 times:/ I will see about improving this!

src/amo/components/SearchContextCard/index.js
```diff
@@ -98,15 +169,6 @@ export class SearchContextCardBase extends React.Component {
       searchText = i18n.gettext('Loading add-ons');
```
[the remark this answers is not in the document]
rebmullin: I believe there is a difference but I got to recheck this

src/amo/components/SearchContextCard/index.js
```diff
@@ -98,15 +169,6 @@ export class SearchContextCardBase extends React.Component {
...
     } else if (!loading && count === 0) {
```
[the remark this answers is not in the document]
rebmullin: hmm...let me check..

src/amo/components/SearchContextCard/index.js
```diff
@@ -1,95 +1,166 @@
...
-    count: PropTypes.number,
-    filters: PropTypes.object,
-    i18n: PropTypes.object.isRequired,
-    loading: PropTypes.bool.isRequired,
-  };
+type Props = {||};
+
+type InternalProps = {|
+  ...Props,
+  categoryName?: string,
```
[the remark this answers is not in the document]
rebmullin: it could also be undefined too so I guess i can either do 

`categoryName?: string | null`

or

`category: string | null | void`

I am not sure if 1  is better than the other..?

I wonder why flow didn't complain about this tho 🤔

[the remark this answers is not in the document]
rebmullin: I just saw your comment below about defining this as null (which I will do:) thanks!

src/amo/components/SearchContextCard/index.js
rebmullin: yeh maybe you're right (I need both). I will look into this further with fresh 👀

tests/unit/amo/components/TestSearchContextCard.js
rebmullin: good point. I think I was just showing as a non extension/theme examples but I think that might be covered. I'll double check and remove or make this more generic.

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +179,72 @@ export class SearchContextCardBase extends React.Component {
...
+
+    if (categoriesState && clientApp) {
+      Object.keys(categoriesState[clientApp]).forEach((type) => {
+        const searchType = categoriesState[clientApp][type];
+        Object.keys(searchType).forEach((category) => {
+          const searchCategory = searchType[category];
+          const { slug } = searchCategory;
+          const { name } = searchCategory;
+          allCategories.push({
+            [slug]: {
```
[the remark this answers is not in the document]
rebmullin: I redid this:)

src/amo/components/SearchContextCard/index.js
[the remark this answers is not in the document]
rebmullin: yeh that was a mistake it seems. I don't think we actually need this. It should be caught by the default

src/amo/components/SearchContextCard/index.js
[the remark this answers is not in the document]
rebmullin: I guess we can change this to make it more consistent. I don't know what it was different but this is what was there originally it seems.

src/amo/components/SearchContextCard/index.js
rebmullin: When there is a category, then we do need to worry about category but I think we can assume when we have categoryName, categories have loaded so I *think* this is okay.

I think because search depends on the category info (if it's present), then I think this is okay to use this for loading here. It is confusing but does that make sense?

tests/unit/amo/components/TestSearchContextCard.js
rebmullin: maybe I should put "...when search is loaded" ?. Is that any clearer?

I need to call the categorie functions in order to set the `categoryName`. But I am also using `dispatchSearchResults` which starts and loads the search. 

Does that make sense?

src/amo/components/SearchContextCard/index.js
rebmullin: I also removed the test I added `should not fetch categories if there is a category filter but it's already loading` since I am not going to using the loading check any longer. If `categoryName` is present, I am using this to determine if we should do the dispatch. I think this is the right way to handle this but let me know if you don't agree :/

tests/unit/amo/components/TestSearchContextCard.js
rebmullin: I updated the tests a bit...let me know if that's better now

rebmullin:
hey @bobsilverberg, thanks for taking this one on:)  So I addressed I think most of your comments. The only thing I am not 100% about is the loading stuff still. It's a little confusing and I tried to explain it above but let me know if that makes sense. If not we can discuss further:)

src/amo/components/SearchContextCard/index.js
bobsilverberg: I've looked over the lastest patch and I think you're right - you don't need loading status for categories, just for search. It might make sense to change the prop name from `loading` to `loadingSearch` to that it's clear which loading status we're concerned with.

rebmullin:
> Do you think this will work?

this @bobsilverberg! yep this is definitely a lot cleaner. It mostly works, I just need to adjust it slightly for themes (as filter is statictheme,persona). I'll get that updated soon

rebmullin:
hey @bobsilverberg, thanks for all your help here. Can you please take another look? Hopefully it's close now:) 🤞

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +172,64 @@ export class SearchContextCardBase extends React.Component {
...
+          // eslint-disable-next-line prefer-destructuring
+          let addonType = filters.addonType;
+          if (addonType === ADDON_TYPE_THEMES_FILTER) {
+            addonType = ADDON_TYPE_STATIC_THEME;
+          }
+
+          const categories = appTypes[addonType];
+
+          if (categories) {
+            categoryName = categories[currentCategory].name;
```
bobsilverberg: `categories[currentCategory]` could be `undefined` here, so I think you need `categoryName = categories[currentCategory] ? categories[currentCategory].name || null;`. Maybe there should also be a test for that, which would have caught this?

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +172,64 @@ export class SearchContextCardBase extends React.Component {
...
+    if (categoriesState && clientApp) {
+      const appTypes = categoriesState[clientApp];
+
+      if (appTypes) {
+        if (
+          filters &&
+          filters.addonType &&
+          typeof filters.addonType === 'string'
+        ) {
+          // eslint-disable-next-line prefer-destructuring
```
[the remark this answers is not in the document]
rebmullin: yeh this was flow thing

src/amo/components/SearchContextCard/index.js
rebmullin: hmm let me recheck

src/amo/components/SearchContextCard/index.js
```diff
@@ -117,15 +172,64 @@ export class SearchContextCardBase extends React.Component {
...
+  const { filters = {} } = search;
+
+  let currentCategory;
+  let categoryName = null;
+
+  if (
+    search &&
+    filters &&
+    filters.category &&
+    typeof filters.category === 'string'
```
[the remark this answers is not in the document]
rebmullin: I think if there is no category, we wouldn't have to do anything else either which is what this checks. 

I guess I can include it here but then I have to set up a default empty string value to addonType (to make flow happy). This is one of the reasons why I moved it here. 

Just to be clear this is what I have:

```diff
+  let addonType = '';
+
   if (
     search &&
     filters &&
+    filters.addonType &&
+    typeof filters.addonType === 'string' &&
     filters.category &&
     typeof filters.category === 'string'
   ) {
     currentCategory = filters.category;
+
+    addonType = filters.addonType;
+    if (addonType === ADDON_TYPE_THEMES_FILTER) {
+      addonType = ADDON_TYPE_STATIC_THEME;
+    }
   }
```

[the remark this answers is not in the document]
rebmullin: I still think how I have this is okay since the **category** filter is ultimately the determining factor, right? I understand that `addonType` would be present too but you could also have a search with just `addonType`

with that being said, I could move this above as you suggest I just prefer how it was really only defined where it was being used (also I didn't have to assign the default value as a string). Maybe that doesn't matter tho..?

e.g.: https://addons-dev.allizom.org/en-US/firefox/search/?platform=mac&q=test&type=extension

rebmullin:
@bobsilverberg, I've added a test but let see comment above regarding the addonType check...let me know if i missing something.

src/amo/components/SearchContextCard/index.js
bobsilverberg: This still doesn't seem right to me. I don't think the test you added is addressing this use case because your test loads no categories, so `categories` will be falsey above, but when you want to emulate in the test is `categories` being true, but `categories[currentCategory]` returning `undefined`, because the specific category requested doesn't exist. You can test this interactively by applying your patch and going to http://localhost:3000/en-US/firefox/search/?category=bookmarksx&sort=rating&type=extension

rebmullin:
> You can test this interactively by applying your patch and going to http://localhost:3000/en-US/firefox/search/?category=bookmarksx&sort=rating&type=extension

@bobsilverberg, I am having a little bit of trouble with this last test.

Is this sufficient?:

```javascript
 it('does not render a categoryName when the category is invalid', () => {
    dispatchSearchResults({
      // The API does not return addon results if the category is invalid.
      addons: {},
      filters: {
        category: 'bad-category',
      },
      store: _store,
    });

    const root = render();

    expect(root.find('.SearchContextCard-header').text()).toEqual(
      '0 results found',
    );
  });
```

As noted above, the API returns nothing in this scenario. 

I did try to replicate the API error but it didn't look like this mattered either way when I pass in the errorHandler to render:

```javascript
      id: 'some-id',
      dispatch: _store.dispatch,
    });
    errorHandler.handle(
      createApiError({
        response: { status: 400 },
        jsonResponse: { message: 'Bad Request' },
      }),
    );
```

if this first test shown above is sufficient, then disregard this part regarding the errorHandler.

rebmullin:
I've decided this should be okay:

```javascript
t('does not render a categoryName when the category is invalid', () => {
    dispatchSearchResults({
      // The API does not return addon results if the category is invalid.
      addons: {},
      filters: {
        category: 'bad-category',
      },
      store: _store,
    });

    const root = render();

    expect(root.find('.SearchContextCard-header').text()).toEqual(
      '0 results found',
    );
  });
```

PR #6643, opened 2018-10-16
Make sure internal collections have a string value

### LINKED ISSUE #6635: Bad collection data can cause every detail page to error

* Create a collection with a leading emoji in the title so that the title is essentially blank as per https://github.com/mozilla/addons-server/issues/8923
* Ensure you are signed in
* Visit any details page

### What happened?

* Every details page is a "Server error"
* In the console there is: 'Caught application error: TypeError: "e.name is null; can't access its "localeCompare" property"'

* Even if the collection data is broken it should not result in the details pages being error pages.

willdurand:
This PR should fix #6635.

### DIFF
--- a/src/amo/reducers/collections.js
+++ b/src/amo/reducers/collections.js
@@ -877,7 +877,7 @@ export const createInternalCollection = ({
   description: detail.description,
   id: detail.id,
   lastUpdatedDate: detail.modified,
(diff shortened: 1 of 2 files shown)


bobsilverberg:
Thanks @willdurand!

Issue #6648, opened 2018-10-16
Consider a different title for Android pages


willdurand:
We currently use `Firefox` for the Firefox pages and `Android` for the Android pages, e.g.:

- homepage with `clientApp` set to `firefox`: `Add-ons for Firefox`
- homepage with `clientApp` set to `android`: `Add-ons for Android`
- add-on detail page (`firefox`): `Stratiform – Get this Extension for 🦊 Firefox (en-US)`
- add-on detail page (`android`): `Stratiform – Get this Extension for 🦊 Android (en-US)`

Are we happy with this? @bobsilverberg and I find that not having "Firefox for Android" sounds a bit odd.

willdurand:
@MeridelW I think that's a question for you :)

MeridelW:
I agree not having "Firefox for Android" sounds odd. My once concern with including it is that Google character count limits might cut off the "Android" portion. This will only happen in cases where the extension name is longer. We could cut out the "for" to save some characters. How about this?

Get this Extension for 🦊 Firefox Android (en-US)

bobsilverberg:
Sounds good to me! 👍

PR #6649, opened 2018-10-16
Extract AddonHead component

### LINKED ISSUE #6647: Extract `AddonHead` component from `Addon`
In order to support #6644, we should extract a new component called `AddonHead` from the existing `Addon` component, which would contain the `Helmet` block in the `render()` method.

It will be easier to work on the metadata and it will reduce the size of this (too) big `Addon` component.

For QA: please make sure page title/description and other metadata (JSON+LD, other `meta` tags are present).

willdurand:
Fixes #6647

---

Except 2 or 3 `invariant` statements, code has been moved without modification.

### DIFF
--- a/src/amo/components/AddonHead/index.js
+++ b/src/amo/components/AddonHead/index.js
@@ -1,0 +1,225 @@
+/* @flow */
+import invariant from 'invariant';
+import * as React from 'react';
+import Helmet from 'react-helmet';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import {
+  ADDON_TYPE_DICT,
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_LANG,
+  ADDON_TYPE_OPENSEARCH,
+  ADDON_TYPE_STATIC_THEME,
+  ADDON_TYPE_THEME,
+  CLIENT_APP_ANDROID,
+} from 'core/constants';
+import translate from 'core/i18n/translate';
+import { getPreviewImage } from 'core/imageUtils';
+import { getAddonJsonLinkedData } from 'core/utils/addons';
+import type { AppState } from 'amo/store';
+import type { I18nType } from 'core/types/i18n';
+import type { AddonType } from 'core/types/addons';
+
+type Props = {|
+  addon: AddonType | null,
+|};
+
+type InternalProps = {|
+  ...Props,
+  clientApp: string,
+  i18n: I18nType,
+  lang: string,
+|};
+
+export class AddonHeadBase extends React.Component<InternalProps> {
+  getPageTitle() {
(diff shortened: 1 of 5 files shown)


tests/unit/amo/components/TestAddonHead.js
```diff
@@ -0,0 +1,200 @@
...
+      lang,
+    });
+    const root = render({ addon, store });
+
+    expect(root.find('title')).toHaveText(
+      `${addon.name} – Get this ${name} for 🦊 Firefox (${lang})`,
+    );
+  });
+
+  it.each([
```
bobsilverberg: Nit: It might make for a more messy test, but you could have 2 nested `it.each` calls, one which loops through the 2 values of `clientApp` and another that loops through the values of `type`, instead of two complete tests for `CLIENT_APP_FIREFOX` and `CLIENT_APP_ANDROID`. It would remove a lot of duplication, but it might be a bit ugly.

src/amo/components/AddonHead/index.js
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }
+
+  getPageDescription() {
+    const { addon, i18n } = this.props;
+
+    invariant(addon, 'addon is required');
+
+    return i18n.sprintf(
+      i18n.gettext('Download %(addonName)s for Firefox. %(summary)s'),
```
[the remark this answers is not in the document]
willdurand: I don't think so here, because in both cases it is an add-on for Firefox, right? I mean, downloading something for Android would mean an Android app to me, not a FF app.

This is super confusing, but if we get a better title like "Firefox for Android" in https://github.com/mozilla/addons-frontend/issues/6648, I'll make the change here too

tests/unit/amo/components/TestAddonHead.js
willdurand: ah, that makes me think about a matrix instead:

```
[
  [CLIENT_APP_FIREFOX, ADDON_TYPE_DICT, 'Dictionary'],
  ...,
  [CLIENT_APP_ANDROID, ADDON_TYPE_DICT, 'Dictionary'],
  ...,
]
```

willdurand: The thing is we would need the title somewhere, mmh that's tricky actually.

src/amo/components/AddonHead/index.js
```diff
@@ -0,0 +1,225 @@
...
+
+    const i18nValues = {
+      addonName: addon.name,
+      locale: lang,
+    };
+
+    if (clientApp === CLIENT_APP_ANDROID) {
+      switch (addon.type) {
+        case ADDON_TYPE_DICT:
+          return i18n.sprintf(
```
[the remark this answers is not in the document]
willdurand: oh, right I like it!

PR #6663, opened 2018-10-17
Add alternate links to add-on detail pages

### LINKED ISSUE #6644: Add hreflang attribute/alternate links to cope with duplicate content and reduce crawl budget
In #5755, we added `canonical` links to most pages (those without query parameters). We did that to avoid duplicate content (because of the random query parameters added for analytics, etc.). Yet, we still provide one `canonical` link per `locale` and `clientApp`.

Each AMO URL starts with: `/{locale}/{clientApp}/`.

We support a lot of locales and two apps: `firefox` and `android`. Content may change a lot between two locales but it does not change that much between `firefox` and `android`, except that some add-ons are not available/shown/compatible with Android, including static themes that are about to be numerous (because LWT will be migrated to static themes soon).

Therefore, there are two points to discuss/address: locales and client apps.

## Locales
Good read: https://audisto.com/insights/guides/30/

I believe we could do better by adding `alternate` links with a `hreflang` property. Those links tell crawlers that a page has variants, given a locale in our case.

For each indexed page, and in addition to the `canonical` link that must be a self referencing link (that's already the case), we should add an `alternate` link for each supported locale:

```
<link rel="alternate" hreflang="en-US" href="https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/">
<link rel="alternate" hreflang="en-GB" href="https://addons.mozilla.org/en-GB/firefox/addon/ublock-origin/">
<!-- self referencing canonical link -->
<link rel="canonical" href="https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/">
```

Each locale must be [ISO_639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) compliant.

We can use the `x-default` special locale to denote a "default" page. Since AMO is capable of content negotiation (automatically sets the locale), we can probably use this special locale with a special URL without a locale:

```
<link rel="alternate" hreflang="x-default" href="https://addons.mozilla.org/firefox/addon/ublock-origin/">
```

**Does that sound good?**

## Client apps

**Do we really want to crawl Android pages?** Those pages are very similar to the Firefox pages (because we basically do `s/Firefox/Android/g` to replace all occurrences of "Firefox" with "Android").

What would be the best strategy to do that? I'd be incline to use a meta `robots` with `noindex, follow` for all pages as soon as we have `clientApp === 'android'`.

What do you think? /cc @rraue 

---

## Update for QA (and reviewers)

- We do NOT deindex android pages
- We add a `x-default` link pointing to the `en-US` page
- We add a `en` link pointing to the `en-US` page
- We add a `pt` link pointing to the `pt-PT` page
- We ignore ISO 639-2 locales for now (locales with three letters, e.g., `ast`, `cak`, etc.)
- We focus on add-on detail pages here

willdurand:
Fixes #6644

~~⚠️ depends on #6649~~

---

This PR adds new `link` tags in the add-on detail pages.

You can try it by using https://app.hreflang.org/, which requires three steps:

1. Install `localtunnel`: https://localtunnel.github.io/www/
2. Run it: `lt --port 3000`
3. Update this line to add `xxx.localtunnel.me` instead of `${_config.get('baseURL')}`: https://github.com/mozilla/addons-frontend/blob/7a6dd48ee5050cc863d6fdefff2dd24eafbe24b4/src/amo/utils.js#L64
4. Browse an add-on detail page like: `xxx.localtunnel.me/en-US/firefox/addon/awesome-screenshot-plus-/?`
5. Paste this URL into https://app.hreflang.org/ and click the "test" button

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -111,6 +111,7 @@ module.exports = {
     'experiments',
     'fxaConfig',
     'hctEnabled',
+    'hrefLangsMap',
     'isDeployed',
     'isDevelopment',
     'langMap',
@@ -122,6 +123,7 @@ module.exports = {
     'trackingEnabled',
     'trackingId',
     'trackingSendInitPageView',
+    'unsupportedHrefLangs',
     'validClientAppUrlExceptions',
     'validClientApplications',
     'validLocaleUrlExceptions',
@@ -230,6 +232,22 @@ module.exports = {
     'zh-CN',
     'zh-TW',
   ],
+  // Exclusion list of unsupported locales for alternate links, see:
+  // https://github.com/mozilla/addons-frontend/issues/6644
+  unsupportedHrefLangs: [
+    'ast',
+    'cak',
+    'dsb',
+    'hsb',
+    'kab',
+  ],
+  // Map of locale aliases for "alternate" links, see:
+  // https://github.com/mozilla/addons-frontend/issues/6644
+  hrefLangsMap: {
+    'x-default': 'en-US',
+    en: 'en-US',
+    pt: 'pt-PT',
+  },
   // Map of short langs to longer ones.
   langMap: {
     en: 'en-US',
--- a/tests/unit/amo/components/TestAddonHead.js
+++ b/tests/unit/amo/components/TestAddonHead.js
@@ -280,7 +280,7 @@ describe(__filename, () => {
 
   // This test case ensures the production configuration is taken into account.
   it.each([['x-default', 'en-US'], ['pt', 'pt-PT'], ['en', 'en-US']])(
-    'renders a "x-default" alternate link',
+    'renders a "%s" alternate link',
     (hrefLang, locale) => {
       const addon = createInternalAddon(fakeAddon);
       const baseURL = 'https://example.org';
--- a/tests/unit/amo/components/TestAddonHead.js
+++ b/tests/unit/amo/components/TestAddonHead.js
@@ -242,9 +242,12 @@ describe(__filename, () => {
     const clientApp = CLIENT_APP_FIREFOX;
 
     const _hrefLangs = ['x-default'];
+    const aliasKey = 'x-default';
+    const aliasValue = 'en-US';
     const hrefLangsMap = {
-      'x-default': 'en-US',
+      [aliasKey]: aliasValue,
     };
+
     const _config = getFakeConfig({ baseURL, hrefLangsMap });
     const { store } = dispatchClientMetadata({ clientApp });
 
@@ -253,13 +256,11 @@ describe(__filename, () => {
     expect(root.find('link[rel="alternate"]')).toHaveLength(_hrefLangs.length);
     expect(root.find('link[rel="alternate"]').at(0)).toHaveProp(
       'hrefLang',
-      _hrefLangs[0],
+      aliasKey,
     );
     expect(root.find('link[rel="alternate"]').at(0)).toHaveProp(
       'href',
-      `${baseURL}/${hrefLangsMap[_hrefLangs[0]]}/${clientApp}/addon/${
-        addon.slug
-      }/`,
+      `${baseURL}/${aliasValue}/${clientApp}/addon/${addon.slug}/`,
     );
   });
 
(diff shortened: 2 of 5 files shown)


tests/unit/amo/components/TestAddonHead.js
```diff
@@ -197,4 +207,73 @@ describe(__filename, () => {
...
+    };
+    const _config = getFakeConfig({ baseURL, hrefLangsMap });
+    const { store } = dispatchClientMetadata({ clientApp });
+
+    const root = render({ _config, _hrefLangs, addon, store });
+
+    expect(root.find('link[rel="alternate"]')).toHaveLength(_hrefLangs.length);
+    expect(root.find('link[rel="alternate"]').at(0)).toHaveProp(
+      'hrefLang',
+      _hrefLangs[0],
```
bobsilverberg: Nit: I wonder if the test might be a bit more readable if you were to set `const`s equal to `x-default` and `en-US` and then use those to set values for `_hrefLangs` and `hrefLangsMap`, and then in the asserts you could just use the `const`s again, rather than things like `_hrefLangs[0]` and `hrefLangsMap[_hrefLangs[0]]`?

tests/unit/amo/components/TestAddonHead.js
```diff
@@ -197,4 +207,73 @@ describe(__filename, () => {
...
+        addon.slug
+      }/`,
+    );
+  });
+
+  it('does not render any links for unsupported alternate link locales', () => {
+    const addon = createInternalAddon(fakeAddon);
+    const lang = 'fr';
+
+    const _hrefLangs = ['fr', 'en-US'];
```
bobsilverberg: Nit: You could use `lang` instead of `'fr'` here.

willdurand:
> While the tests added to `TestAddonHead` are good unit tests for the behaviour you've added, I think we might also want an integration test for the aliased locales.
> 
> I suppose we don't usually test config values, so that would be an argument against it, but the config values contain important logic. I.e., we have a requirement that `x-default` always creates a link to `en-US` and that `pt` always creates a link to `pt-PT`, so I wonder if it would be worth having a test that _does not_ mock `hrefLangsMap` in `config` and verify that we get the expected link tags?

Yep, that works for me.

config/default.js
```diff
@@ -230,6 +232,22 @@ module.exports = {
     'zh-CN',
     'zh-TW',
   ],
+  // Exclusion list of unsupported locales for alternate links, see:
+  // https://github.com/mozilla/addons-frontend/issues/6644
+  unsupportedHrefLangs: [
```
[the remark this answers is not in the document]
willdurand: I don't think we should extend this existing data structure. Its purpose is to list most of the existing languages. From there, we use config to select the languages we support.

config/default.js
```diff
@@ -230,6 +232,22 @@ module.exports = {
...
+    'ast',
+    'cak',
+    'dsb',
+    'hsb',
+    'kab',
+  ],
+  // Map of locale aliases for "alternate" links, see:
+  // https://github.com/mozilla/addons-frontend/issues/6644
+  hrefLangsMap: {
```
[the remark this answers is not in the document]
willdurand: Technically, this map could be used to override some locales.

tests/unit/amo/components/TestAddonHead.js
willdurand: yep

tests/unit/amo/components/TestAddonHead.js
willdurand: yep

tests/unit/amo/components/TestAddonHead.js
```diff
@@ -276,4 +277,23 @@ describe(__filename, () => {
 
     expect(root.find('link[rel="alternate"]')).toHaveLength(0);
   });
+
+  // This test case ensures the production configuration is taken into account.
+  it.each([['x-default', 'en-US'], ['pt', 'pt-PT'], ['en', 'en-US']])(
+    'renders a "x-default" alternate link',
```
bobsilverberg: Nit: Should this description use one of the variables from the `it.each` instead of always saying `"x-default"`?

willdurand: oops

PR #6665, opened 2018-10-17
I6433-new home hero

### LINKED ISSUE #6433: Redesign hero area and point to user guides
For the [Homepage Hero Area](https://docs.google.com/document/d/1PZwtYeWC2_eWBOX4a8QUBoOHeAZpEEOFRzEhpsgFrKo/edit#), the new design will point to new user guide pages that are also part of this project.

The homepage will show a few (about 3), but there will probably be a couple more we have in reserve so the content changes regularly. The ones we show will be decided in the monthly homepage refresh that @devaneymoz requests, but further down the line they will be controlled by an admin tool (we might get into this in Q1 2019, so I wanted to mention it in case it changes anything about the implementation).

Final design to be provided by @pwalm.

rebmullin:
fixes #6433


src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+  }
+
+  render() {
+    return (
+      <div className="HomeHero">
+        {this.getHeroHeader()}
+
+        <div className="HomeHero-cards">
+          <Hero name="Home" sections={this.getHeroSections()} />
```
rebmullin: I am using the Hero component here with the expectation that this will work in a similar way to the current home page hero - where items are randomly rotated. We just need to add the `random` prop, and then it should be ready to go...

rebmullin:
hey @pwalm, can you please look at the screenshots shown above and advise if this works for the various sizes? I don't see font info in invision so I am eyeballing this as best I can :) 

Also, I have a few follow up questions:

- Do we want to use specific urls for the guide sections?

- I've made each topic's entire card clickable, is this okay?

- I thought I saw (somewhere) that there might be more cards in the future and these should be rotated...is that right?

config/default.js
```diff
@@ -320,6 +321,8 @@ module.exports = {
 
   enableFeatureInlineAddonReview: true,
 
+  enableNewHomeHero: false,
```
rebmullin: I was thinking I would turn this on in the dev configs once the guide pages are set up

willdurand: Shouldn't we add the pages first then?

Also, a feature flag must start with `enableFeature`.

rebmullin: >Shouldn't we add the pages first then?

I am working on the other pages but wanted to tackle this page first. I can create another issue to turn this "on" in dev once those are ready. Would that be okay?


I've updated the feature flag name, thanks : )

willdurand: ok, that sounds good.

(also, that looks super shiny)

pwalm:
> Do we want to use specific urls for the guide sections?
Yes, @devaneymoz can give guidance on that

> I've made each topic's entire card clickable, is this okay?
Yes!

> I thought I saw (somewhere) that there might be more cards in the future and these should be rotated...is that right?
Correct.

config/default.js
```diff
@@ -349,6 +350,8 @@ module.exports = {
 
   enableFeatureInlineAddonReview: true,
 
+  enableFeatureNewHomeHero: false,
```
willdurand: Maybe you can call this feature `HomeHeroGuides`. Otherwise we might end up with a `NewNewHomeHero`, etc. next time.

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+import Card from 'ui/components/Card';
+import Icon from 'ui/components/Icon';
+
+import './styles.scss';
+
+type InternalProps = {|
+  i18n: I18nType,
+|};
+
+export class HomeHeroBase extends React.Component<InternalProps> {
```
willdurand: How about we call this component `HomeHeroGuides`?

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+import Link from 'amo/components/Link';
+import translate from 'core/i18n/translate';
+import Hero from 'ui/components/Hero';
+import type { I18nType } from 'core/types/i18n';
+import Card from 'ui/components/Card';
+import Icon from 'ui/components/Icon';
+
+import './styles.scss';
+
+type InternalProps = {|
```
willdurand: You could spread an empty `Props` type maybe, because `Props` is what you should pass to ` React.ComponentType` at the bottom of the page.

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+  i18n: I18nType,
+|};
+
+export class HomeHeroBase extends React.Component<InternalProps> {
+  getHeroes() {
+    const { i18n } = this.props;
+
+    // TODO: get urls
```
willdurand: File an issue and link to it here please.

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+    return [
+      {
+        title: i18n.gettext('Stay Safe Online'),
+        icon: 'stop-hand',
+        url: '/guides/stay-safe',
```
willdurand: I think we put a trailing slash to each URL.

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+      },
+    ];
+  }
+
+  getHeroHeader() {
+    const { i18n } = this.props;
+
+    return (
+      <div className="HomeHero-header">
+        <h2 className="HomeHero-title">
```
willdurand: That should probably be `HomeHero-header-title`, since it is inside `HomeHero-header`.

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+          {i18n.gettext('Extensions are like apps for your browsers.')}
+        </h2>
+        <h3 className="HomeHero-subtitle">
```
willdurand: Same here, that should be `HomeHero-header-subtitle`

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+  getHeroSections() {
+    return this.getHeroes().map((hero) => {
+      const { title, icon, url } = hero;
+
+      return (
+        <Link key={url} to={url} className="HomeHero-link">
+          <Card className="HomeHero-card">
+            <Icon className="HomeHero-icon" name={icon} />
+            <h4>
+              <span>{title}</span>
```
willdurand: Why do you need a span inside a `h4`?

src/amo/components/HomeHero/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+  getHeroSections() {
+    return this.getHeroes().map((hero) => {
+      const { title, icon, url } = hero;
+
+      return (
+        <Link key={url} to={url} className="HomeHero-link">
+          <Card className="HomeHero-card">
+            <Icon className="HomeHero-icon" name={icon} />
+            <h4>
```
willdurand: I guess this would need a css class too.

src/amo/components/HomeHero/index.js
willdurand: You can probably use `React.PureComponent` here, see: https://reactjs.org/docs/react-api.html#reactpurecomponent

src/amo/components/HomeHero/styles.scss
```diff
@@ -0,0 +1,98 @@
...
+  .Card-contents {
+    align-items: center;
+    background-color: $white;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    padding: 24px 0 12px;
+
+    @include respond-to(large) {
+      min-height: 138px;
```
willdurand: Why 138 exactly?

src/amo/components/HomeHero/styles.scss
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @include respond-to(extraLarge) {
+      min-height: 0;
+    }
+  }
+}
+
+.HomeHero-link {
+  &:link {
```
willdurand: I don't remember exactly, but this might not be supported everywhere (I mean, the way you think it should be supported)

src/amo/components/HomeHero/styles.scss
```diff
@@ -0,0 +1,98 @@
...
+    color: $black;
+    text-decoration: none;
+  }
+}
+
+.HomeHero-icon {
+  &.Icon {
+    height: 40px;
+    width: 40px;
```
willdurand: I would introduce a sass variable for this value, like `$icon-size`.

src/amo/pages/Home/index.js
```diff
@@ -245,7 +246,11 @@ export class HomeBase extends React.Component {
 
         {errorHandler.renderErrorIfPresent()}
 
-        <HomeHeroBanner />
+        {_config.get('enableFeatureNewHomeHero') ? (
+          <HomeHero />
```
willdurand: Yep, `HomeHeroGuides` would be cute here :)

tests/unit/amo/components/TestHomeHero.js
```diff
@@ -0,0 +1,38 @@
...
+} from 'tests/unit/helpers';
+import Hero from 'ui/components/Hero';
+
+describe(__filename, () => {
+  const defaultProps = {
+    i18n: fakeI18n(),
+    store: dispatchClientMetadata().store,
+  };
+
+  function render(props) {
```
willdurand: Could you please initialize the default props in the function's signature?

Also, could you please use the arrow syntax instead of the old `function`?

tests/unit/amo/components/TestHomeHero.js
```diff
@@ -0,0 +1,38 @@
...
+    return shallowUntilTarget(
+      <HomeHero {...defaultProps} {...props} />,
+      HomeHeroBase,
+    );
+  }
+
+  it('renders a HomeHero', () => {
+    const root = render();
+    expect(root).toHaveClassName('HomeHero');
```
willdurand: You can add some more assertions for the title, subtitle, etc. No need to do exact matches, it should just be there.

willdurand:
I think we should change the `margin` on small screens, it is odd to see a difference:

screen shot 2018-10-25 at 17 45 46

(I don't like how it is on desktop too, but I can live with that)

Still on mobile, I think we can reduce the font sizes, it is a bit too big, no?

screen shot 2018-10-25 at 17 47 11

src/amo/components/HomeHero/styles.scss
rebmullin: hey, can you tell me what exactly is not supported? I see `&:link` used in many places so let me know if you mean something else..

src/amo/components/HomeHero/styles.scss
rebmullin: for ~tablet, the "Enhance your media Experience" makes the card expand because the text wraps to a 2nd line. I noticed it was approx this height. So I went with that so the other cards would match and it would be balanced.

src/amo/components/HomeHero/index.js
rebmullin: Actually I can remove this..

src/amo/components/HomeHero/index.js
rebmullin: oh? okay, I'll try that..

tests/unit/amo/components/TestHomeHeroGuides.js
```diff
@@ -0,0 +1,43 @@
...
+    expect(root.find('.HomeHeroGuides-header-subtitle')).toHaveLength(1);
+  });
+
+  it('renders the hero sections', () => {
+    const root = render();
+    const heroSections = root.find(Hero);
+
+    expect(heroSections).toHaveLength(1);
+    expect(heroSections).toHaveProp('sections');
+    expect(heroSections.props().sections).toHaveLength(3);
```
rebmullin: is this enough here? I wasn't sure if need to get more specific than this here..

src/amo/components/HomeHero/styles.scss
willdurand: That's true for `en-US` but likely not for `de` (German) for instance.

src/amo/components/HomeHero/styles.scss
willdurand: well, nevermind. I don't remember exactly tbh.. One question though: why do you need to target the links like this? It only select those links when they are not yet visited, no?

rebmullin: Well initially I had something like this:

```css
.HomeHero-link {
   color: $black;
   text-decoration: none;
}
```

but it (the color specifically) was overridden by the `a:link` styles. I can try nesting tho if you think that's better..

src/amo/components/HomeHero/styles.scss
rebmullin: oh? will it be longer for de? I guess I didn't think of it being longer...

let me look into another approach then..

rebmullin:
hey @willdurand, thanks for the review. I've updated..can you take another look please ?

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+import Hero from 'ui/components/Hero';
+import type { I18nType } from 'core/types/i18n';
+import Card from 'ui/components/Card';
+import Icon from 'ui/components/Icon';
+
+import './styles.scss';
+
+type Props = {};
+
+type InternalProps = {
```
willdurand: You should always use the strict notation: `{| ... |}`

willdurand: (Event for `Props`)

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+    );
+  }
+
+  getHeroSections() {
+    return this.getHeroes().map((hero) => {
+      const { title, icon, url } = hero;
+
+      return (
+        <Link key={url} to={url} className="HomeHeroGuides-link">
```
willdurand: Since you name this "hero sections", I'd expect the className to be `HomeHeroGuides-section`.

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+      );
+    });
+  }
+
+  render() {
+    return (
+      <div className="HomeHeroGuides">
+        {this.getHeroHeader()}
+
+        <div className="HomeHeroGuides-cards">
```
willdurand: Since you name the different cards "hero sections", I'd expect the className to be `HomeHeroGuides-sections`.

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+          <Hero name="Home" sections={this.getHeroSections()} />
```
willdurand: What is this name exactly? And why is it set to `Home` and `not `HomeHeroGuides` or `HomeHeroGuides-Hero`?

tests/unit/amo/components/TestHomeHeroGuides.js
```diff
@@ -0,0 +1,33 @@
...
+      HomeHeroGuidesBase,
+    );
+  };
+
+  it('renders a HomeHeroGuides', () => {
+    const root = render();
+
+    expect(root).toHaveClassName('HomeHeroGuides');
+    expect(root.find('.HomeHeroGuides-header-title')).toHaveLength(1);
+    expect(root.find('.HomeHeroGuides-header-subtitle')).toHaveLength(1);
```
willdurand: You could maybe assert partial content here:

```
     expect(root.find('.HomeHeroGuides-header-title')).toMatch(/Extensions are/); 
     expect(root.find('.HomeHeroGuides-header-subtitle')).toMatch(/They add features/); 
```

tests/unit/amo/components/TestHomeHeroGuides.js
willdurand: yep, although you should use `prop('sections')` instead of `.props().sections`.

tests/unit/amo/components/TestHomeHeroGuides.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('renders the hero sections', () => {
+    const root = render();
+    const heroSections = root.find(Hero);
```
willdurand: This should be named `hero`, not `heroSections`, no?

tests/unit/amo/components/TestHomeHeroGuides.js
```diff
@@ -0,0 +1,33 @@
...
+
+    expect(heroSections).toHaveLength(1);
+    expect(heroSections).toHaveProp('sections');
+    expect(heroSections.props().sections).toHaveLength(3);
+  });
```
willdurand: You could probably also "shallow" the `heroSections` (which should be renamed `hero`) and see if you can find a `Link`.

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+          {i18n.gettext(
+            'They add features to Firefox to make browsing faster, smarter, or just plain fun.',
+          )}
+        </h3>
+      </div>
+    );
+  }
+
+  getHeroSections() {
+    return this.getHeroes().map((hero) => {
```
willdurand: I did not pay attention to that before but the `Hero` component expects a list of `HeroSection` components. Why not reusing those? Flow does not complain (but it should...)

rebmullin: I actually missed that too :/

let me try to get it to work with this.

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
+/* @flow */
+import * as React from 'react';
+import { compose } from 'redux';
+
+import translate from 'core/i18n/translate';
+import type { I18nType } from 'core/types/i18n';
```
willdurand: `import type` should be at the bottom of the `import` section.

src/amo/components/HomeHeroGuides/index.js
```diff
@@ -0,0 +1,95 @@
...
+    );
+  }
+
+  getHeroSections() {
+    return this.getHeroes().map((hero) => {
+      const { title, icon, url } = hero;
+
+      return (
+        <HeroSection key={url} linkTo={url} styleName="HomeHeroGuides">
+          <Card className="HomeHeroGuides-card">
```
willdurand: This could be `HomeHeroGuides-section` or `-item` or something that is not the  underlying React component. Then you can reuse it in the children:

- `HomeHeroGuides-section-icon`
- `HomeHeroGuides-section-title`

src/ui/components/Hero/styles.scss
```diff
@@ -40,7 +40,7 @@
   }
 }
 
-.Hero .HeroSection {
+.Hero .HeroSection.HeroSection-styleName--default {
```
willdurand: How about:

```suggestion
.Hero .HeroSection-styleName--default {
```

src/ui/components/HeroSection/index.js
```diff
@@ -34,7 +44,7 @@ export default class HeroSection extends React.Component<Props> {
           <Link
             className="HeroSection-link-wrapper"
             onClick={this.props.onClick}
-            to={addQueryParams(linkTo, { src: INSTALL_SOURCE_HERO_PROMO })}
+            {...createToLink(linkTo, linkSource)}
```
willdurand: It would make sense to configure this link in the parent component then. Instead of introducing a new `linkSource` prop, you can call `addQueryParams()` in `src/amo/components/HomeHeroBanner/index.js`. This component will become more generic that way.

src/amo/components/HomeHeroBanner/styles.scss
```diff
@@ -42,33 +42,35 @@ $colors-third-item: ('cyan', #15aabf, #1098ad) ('teal', #12b886, #0ca678);
 }
 
 .HeroSection {
-  &:nth-child(1) {
-    $color: nth($colors-first-item, 1);
-    $bg-color: nth($color, 2);
-    $fg-color: nth($color, 3);
+  &.HeroSection-styleName--default {
```
willdurand: I would not make this change, instead you can update the class name L44. There is no need to add yet another nested level. See my suggestion above.

src/amo/components/HomeHeroBanner/styles.scss
```diff
@@ -42,33 +42,35 @@ $colors-third-item: ('cyan', #15aabf, #1098ad) ('teal', #12b886, #0ca678);
 }
 
 .HeroSection {
```
willdurand:
```suggestion
.HeroSection-styleName--default {
```

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -293,7 +294,11 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
       const { description, title, url } = hero;
 
       return (
-        <HeroSection key={url} linkTo={url}>
+        <HeroSection
+          key={url}
+          linkTo={url}
```
willdurand: I'd rather add the query params here (see comment below).

src/amo/components/HomeHeroBanner/styles.scss
willdurand: Maybe `default` is not a great name anymore. We could change it in the `Hero` component to: `random-color` maybe?

src/amo/components/HomeHeroGuides/index.js
rebmullin: ahh..I see.. yep will do...

src/ui/components/HeroSection/index.js
rebmullin: yep that's a lot better!

src/amo/components/HomeHeroBanner/styles.scss
rebmullin: yeh I agree. `random-color` sounds fine to me...

src/ui/components/Hero/styles.scss
```diff
@@ -15,54 +15,3 @@
     padding: $padding-page-l;
   }
 }
-
-.Hero-contents {
```
rebmullin: this is all very specific to the current HomeHeroBanner, so I moved a lot of these styles over to `HomeHeroBanner.scss`

rebmullin:
@willdurand, thanks for another review. Can you please see how it looks now? :)

src/ui/components/Hero/styles.scss
willdurand: okay, I guess that's fine

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -293,7 +295,11 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
       const { description, title, url } = hero;
 
       return (
-        <HeroSection key={url} linkTo={url}>
+        <HeroSection
+          key={url}
+          linkTo={addQueryParams(url, { src: INSTALL_SOURCE_HERO_PROMO })}
+          styleName="random-color"
```
willdurand: You should cover this line with a test case.

src/amo/components/HomeHeroBanner/index.js
```diff
@@ -293,7 +295,11 @@ export class HomeHeroBannerBase extends React.Component<InternalProps> {
       const { description, title, url } = hero;
 
       return (
-        <HeroSection key={url} linkTo={url}>
+        <HeroSection
+          key={url}
+          linkTo={addQueryParams(url, { src: INSTALL_SOURCE_HERO_PROMO })}
```
willdurand: You should cover this line with a test case.

tests/unit/ui/components/TestHeroSection.js
```diff
@@ -29,14 +28,12 @@ describe(__filename, () => {
...
-    const root = shallowRender({ linkTo: '/whatever/' });
+    const linkTo = '/somewhere/';
+    const root = shallowRender({ linkTo });
     const link = root.find(Link);
 
     expect(link).toHaveProp('className', 'HeroSection-link-wrapper');
-    expect(link).toHaveProp(
-      'to',
-      `/whatever/?src=${INSTALL_SOURCE_HERO_PROMO}`,
-    );
```
willdurand: Since this has been removed, it is important to add a test case in the test file of `HomeHeroBanner`.

willdurand:
@rebmullin I forgot, the padding on small screens is not correct:

screen shot 2018-10-29 at 16 33 05

PR #6670, opened 2018-10-18
Provide translation hint for admin link

### LINKED ISSUE #6669: "Admin Links" string needs a localization comment
It's unclear what Admin means in this context, and I'm pretty sure everyone is going to translate it as "manage", given similar strings in the file.
https://github.com/mozilla/addons-frontend/blob/master/src/amo/components/AddonAdminLinks/index.js#L144

championshuttler:
Fixes #6669

### DIFF
--- a/src/amo/components/AddonAdminLinks/index.js
+++ b/src/amo/components/AddonAdminLinks/index.js
@@ -141,7 +141,12 @@ export class AddonAdminLinksBase extends React.Component<InternalProps> {
 
     return (
       <DefinitionList className="AddonAdminLinks">
-        <Definition term={i18n.gettext('Admin Links')}>
+        <Definition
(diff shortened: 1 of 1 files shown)


src/amo/components/AddonAdminLinks/index.js
```diff
@@ -141,7 +141,12 @@ export class AddonAdminLinksBase extends React.Component<InternalProps> {
 
     return (
       <DefinitionList className="AddonAdminLinks">
-        <Definition term={i18n.gettext('Admin Links')}>
+        <Definition
+          term={
+            // translators: This action allows a admin to access the private links.
```
bobsilverberg: I think this would be better as `This is a list of links to administrative functions`.

PR #6675, opened 2018-10-18
Setup a/b experiment: install button color

### LINKED ISSUE #6144: Run A/B test for install button (green vs. blue)

addons.mozilla.org product pages get ~4.3M views.  UX changes that improve add-on installation are high impact.   There was a [study done](https://mana.mozilla.org/wiki/display/FIREFOX/AMO+Optimization+for+Non-Firefox+Users+Landing+on+Add-on+Pages), that may have some correlations for ideas for the upcoming Product Page tweaks to drive engagement.  

The study looked for UX changes that improved engagement - specific to AMO pages visited from non-Fx browsers.   In that case the Action was to install Firefox (from non-Fx Browser), whereas users with Firefox we want them to install the add-on.  

### What happened?

Version B had ~10 times more engagement.  So highly worth testing out some of the relatable ideas in the new designs for both our AMO pages from Fx browsers and non-Fx Browsers.  @pwalm @atsay @jvillalobos 

Worth talking to Cmore - on who the UX and study folks were that tried the various UX components.  I think only one of the ideas in the study made it into the AMO pages from non-Fx browsers (the big green button).

Data driven decisions on the page UX and content have helped fix the download button (big increase in engagement) and add TAAR to DiscoPane (0.8% increase in a user's probability to install overall, and a 1.2% increase for non-en-US users.  Users with TAAR retained 1% better than those without TAAR).

willdurand:
Fixes #6144

---

Only enabled in -dev for now.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -356,6 +356,6 @@ module.exports = {
   // experiments.
   experiments: {
     home_hero: true,
-    install_button_color: true,
+    install_button_color: false,
   },
 };
--- a/config/default.js
+++ b/config/default.js
@@ -356,5 +356,6 @@ module.exports = {
   // experiments.
   experiments: {
     home_hero: true,
+    install_button_color: true,
   },
 };
--- a/src/core/components/AMInstallButton/index.js
+++ b/src/core/components/AMInstallButton/index.js
@@ -32,6 +32,7 @@ import tracking, {
   getAddonEventCategory,
 } from 'core/tracking';
 import { isTheme } from 'core/utils';
+import { withExperiment } from 'core/withExperiment';
 import Button from 'ui/components/Button';
 import Icon from 'ui/components/Icon';
 import type { AppState } from 'amo/store';
@@ -41,6 +42,7 @@ import type { AddonType } from 'core/types/addons';
 import type { I18nType } from 'core/types/i18n';
 import type { ReactRouterLocationType } from 'core/types/router';
 import type { ButtonType } from 'ui/components/Button';
+import type { WithExperimentInjectedProps } from 'core/withExperiment';
 
 import './styles.scss';
 
@@ -57,6 +59,7 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
+  ...WithExperimentInjectedProps,
   _config: typeof config,
   _log: typeof log,
   _tracking: typeof tracking,
@@ -83,6 +86,11 @@ type ButtonProps = {|
 
 const TRANSITION_TIMEOUT = 150;
 
+const EXPERIMENT_ID = 'install_button_color';
+export const EXPERIMENT_CATEGORY = 'AMO Install Button Color Experiment';
+export const VARIANT_BLUE = 'blue';
+export const VARIANT_GREEN = 'green';
+
 export class AMInstallButtonBase extends React.Component<InternalProps> {
   static defaultProps = {
     _config: config,
@@ -91,6 +99,20 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
     _window: typeof window !== 'undefined' ? window : {},
     puffy: true,
   };
+
+  componentDidMount() {
+    const { _log, _tracking, experimentEnabled, variant } = this.props;
+
+    if (!experimentEnabled) {
+      _log.debug(`Experiment "${EXPERIMENT_ID}" is disabled.`);
+      return;
+    }
+
+    _tracking.sendEvent({
+      action: variant,
+      category: EXPERIMENT_CATEGORY,
+    });
+  }
 
   installTheme = (event: SyntheticEvent<HTMLAnchorElement>) => {
     const { addon, status, installTheme } = this.props;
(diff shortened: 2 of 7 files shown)


config/default.js
```diff
@@ -335,5 +335,6 @@ module.exports = {
   // experiments.
   experiments: {
     home_hero: true,
+    install_button_color: true,
```
kumar303: Is there a reason why these aren't camel case?

src/core/components/AMInstallButton/index.js
```diff
@@ -91,12 +99,28 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
...
+    });
+  }
+
   installTheme = (event: SyntheticEvent<HTMLAnchorElement>) => {
     const { addon, status, installTheme } = this.props;
 
     event.preventDefault();
     event.stopPropagation();
 
+    this.trackInstallClick();
```
kumar303: We already track the install and activation events so I don't think you need to duplicate the click event here. I'm not an analytics pro but I think all we need to do is add one event (on `componentDidMount`, maybe?) to link the user to the variant of the button. After that we should be able to visualize conversions but also see which conversion is linked to which button variant.

config/default.js
willdurand: The only reason I can provide is that this ID is used in the cookie name, which is prettier all in snake case.

src/core/components/AMInstallButton/index.js
willdurand: Sorry, it looks like I wasted your time.

I thought that this was okay because we don't necessarily need to measure the bounce rate. All we want is to determine which button color gets clicked the most, which should be the case with this current implementation, unless I am missing something.

If we add some "page view" event (in `componentDidMount()` as you said), I still think we need to add a "click" event because it seems tricky to correlate events from different categories in GA.

kumar303: > If we add some "page view" event (in componentDidMount() as you said), I still think we need to add a "click" event because it seems tricky to correlate events from different categories in GA.

I'm pretty sure GA already has the concept of a user session so adding a "page view" with `variant=blue | green` still seems like it would work to me. I suggest asking Jorge how he plans to visualize it. I'd imagine he could make a report of all clicks (using our existing events and maybe summing together all categories) then filtering by which button variant the user was shown.

If you still think that we need a generic click event (i.e. without a specific category) then I'd rather see that added alongside the other existing event: https://github.com/mozilla/addons-frontend/blob/511d9666d402e4ba30f5d1790e04784967df8b70/src/core/installAddon.js#L436-L440

willdurand: So, last time, we setup the GA report ourselves. I can try to do the same with your approach though, that should be doable, but I'll be sure when I'll have enough data in GA -dev.

> If you still think that we need a generic click event (i.e. without a specific category) then I'd rather see that added alongside the other existing event: [/src/core/installAddon.js@`511d966`#L436-L440](https://github.com/mozilla/addons-frontend/blob/511d9666d402e4ba30f5d1790e04784967df8b70/src/core/installAddon.js#L436-L440)

We can't because we should also capture clicks when the mozAddonManager is not used.

tests/unit/core/components/TestAMInstallButton.js
```diff
@@ -640,4 +646,45 @@ describe(__filename, () => {
 
     expect(root.find(Button)).toHaveProp('disabled', false);
   });
+
+  describe('install_button_color experiment', () => {
+    const renderWithExperiment = (props = {}) => {
+      return render({
+        experimentEnabled: true,
```
kumar303: I guess I hadn't noticed this pattern before but I think it would be nicer to test it with a config object. That would cover the integration between our config parameter and the experiment (which is important).

```js
return render({
  _config: getFakeConfig({ experiments: { install_button_color: true } }),
  ...props,
});
```

config/default.js
```diff
@@ -356,5 +356,6 @@ module.exports = {
   // experiments.
   experiments: {
     home_hero: true,
+    install_button_color: false,
```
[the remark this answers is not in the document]
willdurand: I actually like cookie names in snake case, that's just me. I will file an issue to change to camel case because that will be more consistent.

src/core/components/AMInstallButton/index.js
```diff
@@ -363,6 +389,11 @@ const AMInstallButton: React.ComponentType<Props> = compose(
   withRouter,
   connect(mapStateToProps),
   translate(),
+  withExperiment({
```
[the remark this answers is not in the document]
willdurand: haha, yeah, one day.

willdurand:
> I think this will get you the GA data you need but I guess you'll know for sure once testing it in dev. The green button looks good to me locally.

I'll try to generate some data in -dev and see if I can enable the experiment this week.

config/default.js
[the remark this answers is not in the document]
willdurand: See: https://github.com/mozilla/addons-frontend/issues/6821

PR #6684, opened 2018-10-19
Pass author IDs to AddonsByAuthorsCard

### LINKED ISSUE #5608: Only reviews are visible after editing the username

### LINKED ISSUE #6610: AddonsByAuthorsCard should take a list of user IDs, not usernames
To support https://github.com/mozilla/addons-server/issues/8302, we need to change the `AddonsByAuthorsCard` component to take a list of user IDs instead of usernames.

This issue is blocked by an API change requested here: https://github.com/mozilla/addons-server/issues/8901

### LINKED ISSUE #9735: Enable the Sponsored add-ons shelf on the homepage
Flip the flag so that the Sponsored add-ons shelf will appear on the homepage on stage and prod.

Requested by @diox via Matrix on behalf of @jvillalobos

willdurand:
Fixes #5608
Fixes #6610

### DIFF
--- a/tests/unit/amo/components/TestAddonsByAuthorsCard.js
+++ b/tests/unit/amo/components/TestAddonsByAuthorsCard.js
@@ -37,6 +37,9 @@ import {
 } from 'tests/unit/helpers';
 
 describe(__filename, () => {
+  const randomAuthorId1 = 123;
+  const randomAuthorId2 = 456;
+
   const fakeAuthorOne = {
     ...fakeAuthor,
     name: 'Krupa',
@@ -254,7 +257,7 @@ describe(__filename, () => {
   it('should render nothing if add-ons are null', () => {
     const root = render({
       addons: null,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
     });
 
     expect(root).not.toHaveClassName('AddonsByAuthorsCard');
@@ -267,7 +270,7 @@ describe(__filename, () => {
     store.dispatch(
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_EXTENSION,
-        authorIds: ['test2'],
+        authorIds: [randomAuthorId2],
         errorHandlerId: errorHandler.id,
         pageSize: EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
       }),
@@ -275,7 +278,7 @@ describe(__filename, () => {
 
     const root = render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       store,
     });
@@ -295,7 +298,7 @@ describe(__filename, () => {
 
     render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       numberOfAddons,
       store,
@@ -305,7 +308,7 @@ describe(__filename, () => {
       dispatchSpy,
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_EXTENSION,
-        authorIds: ['test2'],
+        authorIds: [randomAuthorId2],
         errorHandlerId: errorHandler.id,
         pageSize: numberOfAddons,
       }),
@@ -320,7 +323,7 @@ describe(__filename, () => {
 
     const root = render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       numberOfAddons,
       store,
@@ -330,14 +333,14 @@ describe(__filename, () => {
 
     root.setProps({
       addonType: ADDON_TYPE_THEME,
-      authorIds: ['test1'],
+      authorIds: [randomAuthorId1],
     });
 
     sinon.assert.calledWith(
       dispatchSpy,
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_THEME,
-        authorIds: ['test1'],
+        authorIds: [randomAuthorId1],
         errorHandlerId: errorHandler.id,
         pageSize: numberOfAddons,
       }),
@@ -349,14 +352,14 @@ describe(__filename, () => {
 
     root.setProps({
       addonType: ADDON_TYPE_THEME,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
     });
 
     sinon.assert.calledWith(
       dispatchSpy,
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_THEME,
-        authorIds: ['test2'],
+        authorIds: [randomAuthorId2],
         errorHandlerId: errorHandler.id,
         pageSize: numberOfAddons,
       }),
@@ -371,7 +374,7 @@ describe(__filename, () => {
 
     const root = render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       numberOfAddons,
       store,
@@ -381,14 +384,14 @@ describe(__filename, () => {
 
     root.setProps({
       addonType: ADDON_TYPE_OPENSEARCH,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
     });
 
     sinon.assert.calledWith(
       dispatchSpy,
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_OPENSEARCH,
-        authorIds: ['test2'],
+        authorIds: [randomAuthorId2],
         errorHandlerId: errorHandler.id,
         pageSize: numberOfAddons,
       }),
@@ -403,7 +406,7 @@ describe(__filename, () => {
 
     const root = render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       numberOfAddons,
       store,
@@ -419,7 +422,7 @@ describe(__filename, () => {
       dispatchSpy,
       fetchAddonsByAuthors({
         addonType: ADDON_TYPE_EXTENSION,
-        authorIds: ['test2'],
+        authorIds: [randomAuthorId2],
         errorHandlerId: errorHandler.id,
         forAddonSlug: 'testing',
         pageSize: numberOfAddons,
@@ -434,7 +437,7 @@ describe(__filename, () => {
 
     const root = render({
       addonType: ADDON_TYPE_EXTENSION,
-      authorIds: ['test2'],
+      authorIds: [randomAuthorId2],
       errorHandler,
       store,
     });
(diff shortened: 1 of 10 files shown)


tests/unit/amo/pages/TestUserProfile.js
```diff
@@ -530,19 +530,11 @@ describe(__filename, () => {
     );
   });
 
-  it('renders AddonsByAuthorsCard without a user', () => {
+  it('does not render AddonsByAuthorsCards without a user', () => {
```
bobsilverberg: Is the reason this used to be able to render an `AddonsByAuthorsCard` without a user but can no longer because it used to be able to fetch add-ons via the `username` but now it needs an actual user in order to have access to the user's id? If so this makes sense, but it is a change in behaviour, correct? Is it something we're okay with?

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -166,7 +169,7 @@ describe(__filename, () => {
       const state = reducer(
         previousState,
         fetchAddonsByAuthors({
-          authorUsernames: ['author2'],
+          authorIds: ['author2'],
```
bobsilverberg: Ids should be numbers.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -183,7 +186,7 @@ describe(__filename, () => {
         undefined,
         loadAddonsByAuthors({
           addons: [fakeAddon],
-          authorUsernames: [fakeAddon.authors[0].username],
+          authorIds: [fakeAddon.authors[0].id],
```
bobsilverberg: Just looking at this test in isolation, there's no guarantee that `fakeAddon.authors[0].id` isn't the same as `randomAuthorId1`. Maybe it would be better to have 2 new consts in this test for the ids used here and in the second call to `fetchAddonsByAuthors` so that it's obvious to the test that they are different values. I don't think it actually matters that this is `fakeAddon.authors[0].id` here - it looks like it could be any number.

bobsilverberg: Having said all that, it actually looks like the `authorIds` are irrelevant to this test. It would behave the same way if the authors are the same or different, so maybe it should be rewritten to be more simpler, as it suggests that they are expected to be different in the test.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -652,18 +614,18 @@ describe(__filename, () => {
...
       );
 
-      expect(getAddonsForUsernames(state, ['test2', 'no-addons-user'])).toEqual(
-        [
-          createInternalAddon(addons.firstAddon),
-          createInternalAddon(addons.secondAddon),
-        ],
-      );
+      expect(
+        getAddonsForAuthorIds(state, [fakeAuthorTwo.id, randomAuthorId2]),
```
bobsilverberg: I find this very hard to follow. How am I supposed to know to expect that I'll get back `addons.firstAddon` and `addons.secondAddon` when I call `getAddonsForAuthorIds` passing in `fakeAuthorTwo.id` and `randomAuthorId2`? There's nothing in the test that makes that clear, so it's very difficult to know that this test is actually set up correctly for the expected outcome.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -690,13 +654,15 @@ describe(__filename, () => {
...
-          authorUsernames: ['test', 'test2'],
+          authorIds: [fakeAuthorOne.id, fakeAuthorTwo.id],
           count: Object.values(addons).length,
           pageSize: EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
         }),
       );
 
-      expect(getAddonsForUsernames(state, ['test', 'test2'])).toEqual([
+      expect(
+        getAddonsForAuthorIds(state, [fakeAuthorOne.id, fakeAuthorTwo.id]),
```
bobsilverberg: I find this similarly confusing, although not as bad as the one I commented on above. I guess we're supposed to assume that `firstAddon` has `fakeAuthorOne` as an author and `secondAddon` has `fakeAuthorTwo` as an author?

I realize you probably didn't write these tests and are just refactoring some code, but I am disturbed by how opaque the set up is for many of these tests.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -812,30 +794,30 @@ describe(__filename, () => {
...
           addons: Object.values(addons),
-          authorUsernames: ['fakeUsername'],
+          authorIds: [randomAuthorId1],
           count: Object.values(addons).length,
           pageSize: EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
         }),
       );
 
-      expect(getAddonsForUsernames(state, ['nobody'])).toBeNull();
+      expect(getAddonsForAuthorIds(state, [999])).toBeNull();
```
bobsilverberg: How about `randomAuthorId1 + 1` instead of `999`. Then we guarantee they're different.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -854,31 +836,31 @@ describe(__filename, () => {
...
+      authorIds: [randomAuthorId1],
       count: 0,
       errorHandlerId: 'error-handler-id',
       pageSize: EXTENSIONS_BY_AUTHORS_PAGE_SIZE,
     };
 
-    it('returns count for just authorUsernames', () => {
-      const count = 123;
+    it('returns count for just authorIds', () => {
+      const count = randomAuthorId1;
```
bobsilverberg: This feels weird. I know it works, but it makes it seem like there's a relationship between the authorId passed in and the count, which there is not. I think it would be better to just set `count` to a number.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -888,11 +870,11 @@ describe(__filename, () => {
...
       );
 
-      expect(getCountForAuthorNames(state, ['author1'])).toEqual(count);
+      expect(getCountForAuthorIds(state, [randomAuthorId1])).toEqual(count);
     });
 
-    it('returns count for authorUsernames + addonType', () => {
-      const count = 123;
+    it('returns count for authorIds + addonType', () => {
+      const count = randomAuthorId1;
```
bobsilverberg: Again, I think this makes the test harder to understand.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -904,24 +886,24 @@ describe(__filename, () => {
...
-        getCountForAuthorNames(state, ['author1'], ADDON_TYPE_THEME),
+        getCountForAuthorIds(state, [randomAuthorId1], ADDON_TYPE_THEME),
       ).toEqual(count);
     });
 
     it('returns null when there is no match', () => {
       const state = reducer(undefined, loadAddonsByAuthors(params));
 
-      expect(getCountForAuthorNames(state, ['author2'])).toEqual(null);
+      expect(getCountForAuthorIds(state, ['author2'])).toEqual(null);
```
bobsilverberg: `authorId` should be a number and you've used a string here.

tests/unit/amo/reducers/test_addonsByAuthors.js
```diff
@@ -904,24 +886,24 @@ describe(__filename, () => {
...
+    it('returns null when no authorIds provided', () => {
       const state = reducer(undefined, loadAddonsByAuthors(params));
 
-      expect(getCountForAuthorNames(state, [])).toEqual(null);
+      expect(getCountForAuthorIds(state, [])).toEqual(null);
     });
 
     it('resets count when fetching add-ons by authors', () => {
-      const count = 123;
+      const count = randomAuthorId1;
```
bobsilverberg: ibid

tests/unit/amo/components/TestAddonsByAuthorsCard.js
```diff
@@ -258,7 +254,7 @@ describe(__filename, () => {
   it('should render nothing if add-ons are null', () => {
     const root = render({
       addons: null,
-      authorUsernames: ['test2'],
+      authorIds: ['test2'],
```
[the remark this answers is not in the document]
willdurand: ah right, I guess my regex did not replace everything correctly...

tests/unit/amo/pages/TestUserProfile.js
willdurand: > Is the reason this used to be able to render an AddonsByAuthorsCard without a user but can no longer because it used to be able to fetch add-ons via the username but now it needs an actual user in order to have access to the user's id? If so this makes sense

Yep.

>  but it is a change in behaviour, correct?

Yep, mostly because it fixes a bug when a user changes their username.

> Is it something we're okay with?

We do not have any other choice currently, but https://github.com/mozilla/addons-frontend/pull/6686 will fix that by adding a "loading state" to the `AddonsByAuthorsCard` component, which enables us to always render this component. But that's the future 😃

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: > I realize you probably didn't write these tests and are just refactoring some code, but I am disturbed by how opaque the set up is for many of these tests.

I did not write any of these test cases indeed. It was already painful to adapt them... 😢

willdurand: I am going to correct everything you mentioned in this review, but I will leave other improvements for the next PR or next code quality PR, because this PR is quite big already.

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: I set the same `authorIds` for both actions, but I don't really know what this test case is doing tbh. I just `s/username/id` here. That's probably not smart, but I prefer that over introducing a bug/weakness.

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: oops, sorry about that. Search & replace false positives... I should have caught that, sorry 😞

tests/unit/amo/reducers/test_addonsByAuthors.js
willdurand: I found a way to be more verbose in those test cases, I haven't replaced everything, but I could do that in a follow-up or in the next PR (https://github.com/mozilla/addons-frontend/pull/6686) maybe.

willdurand:
Thanks @bobsilverberg, this should be ready for a second review!

PR #6686, opened 2018-10-19
Always render AddonsByAuthorsCard in user profile pages

### LINKED ISSUE #6309: No loading indicators when paging through user profile reviews

- Sign in
- Leave lots of reviews
- View your user profile
- Click *Next* to see the next page of your reviews

### What happened?

The page loads without loading indicators

<details>
<summary>Screencast</summary>

addons-frontend-profile-reviews mov

</details>

The page should show loading indicators otherwise it will be confusing on slower connections.

### LINKED ISSUE #6775: Always render AddonsByAuthorsCard

1. go to https://addons-dev.allizom.org/en-US/firefox/user/abine/
2. refresh
3. observe

### What happened?

The left card is rendered but the right side is empty. Then the right side starts to render some stuff, then it is fully loaded with the add-ons by the user.

Current state in -dev:

The full layout should be rendered in a "loading" state.

Expected state locally (hence the no-css render at the beginning):

willdurand:
Fixes #6775
~~:warning: depends on #6684~~

---

This PR changes the way we use the `AddonsByAuthorsCard` and adds a
better "loading" state to it. Now, we can render the component all the
time.

The `AddonsByAuthorsCard` is rendered with loading animations on first
render (from the server) and it does not shake.

### DIFF
--- a/src/amo/components/AddonsByAuthorsCard/index.js
+++ b/src/amo/components/AddonsByAuthorsCard/index.js
@@ -28,6 +28,7 @@ import {
 import { withErrorHandler } from 'core/errorHandler';
 import translate from 'core/i18n/translate';
 import { isTheme } from 'core/utils';
+import LoadingText from 'ui/components/LoadingText';
 import type { FetchAddonsByAuthorsParams } from 'amo/reducers/addonsByAuthors';
 import type { AppState } from 'amo/store';
 import type { ErrorHandlerType } from 'core/errorHandler';
@@ -40,8 +41,8 @@ import './styles.scss';
 
 type Props = {|
   addonType?: string,
-  authorDisplayName: string,
-  authorIds: Array<number>,
+  authorDisplayName: string | null,
+  authorIds: Array<number> | null,
   className?: string,
   errorHandler?: ErrorHandlerType,
   forAddonSlug?: string,
@@ -94,6 +95,10 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       paginate,
     } = props;
 
+    if (!authorIds) {
+      return;
+    }
+
     this.dispatchFetchAddonsByAuthors({
       addonType,
       authorIds,
@@ -116,6 +121,10 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
       forAddonSlug: oldForAddonSlug,
       location: oldLocation,
     } = this.props;
+
+    if (!newAuthorIds) {
+      return;
+    }
 
     const newPage = paginate
       ? oldLocation.query[pageParam] !== newLocation.query[pageParam]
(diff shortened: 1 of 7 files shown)


src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -359,25 +373,29 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
...
 
-  let addons = getAddonsForAuthorIds(
-    state.addonsByAuthors,
-    authorIds,
-    addonType,
-    forAddonSlug,
-  );
-  addons = addons ? addons.slice(0, numberOfAddons) : addons;
+  let addons = [];
+  let loading = true;
```
kumar303: When you set `loading` explicitly to `true` here, it is out of sync with the Redux state which seems wrong. Maybe the Redux state needs fixing instead? Or were you trying to show loading bars before it begins loading? If so, why?

willdurand: This component is used for secondary data only. I want it to have a "loading state" upon construction, so that it can always be rendered.

Currently, we need the `user` to render it, which means a condition to render/not render it. I want to always render it.

willdurand:
> I'm still not entirely sure what's going on with this patch. Maybe I need to see the STR of the bug you are fixing. Without your patch, I still see loading indicators when the page first loads -- they just don't appear until `FETCH_ADDONS_BY_AUTHORS` is dispatched but that seems correct.

Sorry for the confusion, I linked this PR to a new issue. I don't know why I mixed the previous issue and this fix.. context switching maybe or just me having my head in the clouds :)

I want the component to always be rendered on the server. If you look at the patch and `master` with the redux devtools, you'll see that the right side of the user profile is not rendered by the server.

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -359,25 +373,29 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
...
   const { addonType, authorIds, forAddonSlug, numberOfAddons } = ownProps;
 
-  let addons = getAddonsForAuthorIds(
-    state.addonsByAuthors,
-    authorIds,
-    addonType,
-    forAddonSlug,
-  );
-  addons = addons ? addons.slice(0, numberOfAddons) : addons;
+  let addons = [];
```
kumar303: I don't have a change request but I want to point out that initializing this to an empty array is risky. An empty array should be reserved for the case when the API returned an empty set. Before an API fetch has been made, the value should be `undefined` or `null`.

These issues will come to light when you try to make it [load](https://github.com/mozilla/addons-frontend/blob/0b6d7c48c5a13733833e5795925101ecfb57397e/src/amo/components/AddonsByAuthorsCard/index.js#L97) only once on the server in a future patch so I guess you could wait until then. Right now the selector and probably other things would need changes to support conditional loading.

src/amo/components/AddonsByAuthorsCard/index.js
```diff
@@ -359,25 +373,29 @@ export class AddonsByAuthorsCardBase extends React.Component<InternalProps> {
...
-    state.addonsByAuthors,
-    authorIds,
-    addonType,
-  );
+    loading = getLoadingForAuthorIds(
+      state.addonsByAuthors,
+      authorIds,
+      addonType,
+    );
+    loading = loading === null ? true : loading;
```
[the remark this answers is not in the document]
willdurand: Thank you for this patch ❤️

src/amo/components/AddonsByAuthorsCard/index.js
willdurand: Yep, good call. I'll change that

PR #6688, opened 2018-10-19
Store the versionId of the current version in addon

### LINKED ISSUE #6634: Store the versionId of the current version in addon
Depends on #6637 

**Edited**

While working on https://github.com/mozilla/addons-frontend/issues/6629 I realized that although we have a new `AddonVersionType` which represents an internal representation of a version, we're still storing a version with an `ExternalAddonVersionType` in `addon.current_version`, which complicates things when you want to be able to pass either a version or an add-on into a function.

Instead of this, we can just store the `versionId` of the current version in the add-on and use a selector to retrieve the current version when needed.

In order to avoid having to change a huge amount of code, I will introduce code to store the version id, but not remove the existing code for `current_version` yet. I will also update the versions reducer to make sure that we always store version details in response to an action that also loads add-ons. After that is complete we can start migrating individual components to access the version via a selector rather than via `addon.current_version`. Once everthing has been migrated we can remove the code for `current_version`.

bobsilverberg:
Fixes #6634 

This stores the `versionId` of the `current_version` in the `addon` object. It also updates the versions reducer so that every time an add-on is loaded we also load data for the add-on's `current_version` in the versions redux state.

This does not remove `current_version` from `addon` nor does it update any components to access the version via `currentVersionId`. That will be done in small pieces in separate PRs after this change lands.

### DIFF
--- a/src/amo/reducers/versions.js
+++ b/src/amo/reducers/versions.js
@@ -1,7 +1,17 @@
 /* @flow */
 import invariant from 'invariant';
 
-import { createPlatformFiles } from 'core/reducers/addons';
+import { LOAD_ADDONS_BY_AUTHORS } from 'amo/reducers/addonsByAuthors';
+import {
+  LOAD_COLLECTION_ADDONS,
+  LOAD_CURRENT_COLLECTION,
+  LOAD_CURRENT_COLLECTION_PAGE,
+} from 'amo/reducers/collections';
+import { LOAD_HOME_ADDONS } from 'amo/reducers/home';
+import { LOAD_RECOMMENDATIONS } from 'amo/reducers/recommendations';
+import { LANDING_LOADED } from 'core/constants';
+import { LOAD_ADDON_RESULTS, createPlatformFiles } from 'core/reducers/addons';
+import { SEARCH_LOADED } from 'core/reducers/search';
 import { findFileForPlatform } from 'core/utils';
 import type { UserAgentInfoType } from 'core/reducers/api';
 import type {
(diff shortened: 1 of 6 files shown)


src/amo/reducers/versions.js
```diff
@@ -234,6 +246,90 @@ const reducer = (
...
+        },
+      };
+    }
+
+    case LANDING_LOADED: {
+      const { featured, highlyRated, trending } = action.payload;
+
+      const newVersions = {};
+      for (const apiResponse of [featured, highlyRated, trending]) {
+        if (apiResponse) {
```
kumar303: It would be nice to get some test coverage for when `apiResponse` is falsy. I commented out this statement and I don't see any failures.

tests/unit/amo/reducers/test_versions.js
```diff
@@ -173,4 +195,335 @@ describe(__filename, () => {
...
+          loadAddonsByAuthors({
+            addons: [
+              {
+                ...fakeAddon,
+                current_version: version,
+              },
+            ],
+            authorUsernames: [fakeAddon.authors[0].username],
+            count: 1,
+            pageSize: DEFAULT_API_PAGE_SIZE,
```
kumar303: This is repeated everywhere and it's not relevant to any of the tests. That's a good reason to make a `_loadAddonsByAuthors` helper with some defaults. It could also calculate `count`.

src/amo/reducers/versions.js
bobsilverberg: [None of these can be falsey](https://github.com/mozilla/addons-frontend/blob/2018.10.18-1/src/amo/actions/landing.js#L21-L29), so I can just remove this check.

PR #6698, opened 2018-10-21
I5893 hide sort filter on collection creation

### LINKED ISSUE #5893: Sort add-on section is displayed on the Create a collection from
STR:

1. Log in to AMO -dev
2. Select View My Collections from the user menu
3. Click on Create a collection
4. Observe the fileds available in the create a collection form - https://addons.allizom.org/en-US/firefox/collections/add/

Actual result:
The Sort add-ons section is present, even if the collection doesn't have any add-ons included yet

Expected result:
The "Sort add-ons' section should be hidden in the create a collection form

Notes:
- reproduced on all AMO servers with FF61, Win10x64
- this was fixed once in https://github.com/mozilla/addons-frontend/issues/5628 but it might have been reintroduced at some point (maybe by https://github.com/mozilla/addons-frontend/pull/5676)
- this time, the section is present only in the create collection form, when the user is logged in

image

xu3u4:
Fixes #5893 
The sort filter should be hidden when creating a collection.

Before:
image

After:
image

### DIFF
--- a/tests/unit/amo/pages/TestCollection.js
+++ b/tests/unit/amo/pages/TestCollection.js
@@ -837,6 +837,7 @@ describe(__filename, () => {
     expect(wrapper.find('.Collection-wrapper')).toHaveLength(1);
     expect(wrapper.find(AddonsCard)).toHaveLength(0);
     expect(wrapper.find(CollectionDetailsCard)).toHaveProp('creating', true);
+    expect(wrapper.find(CollectionControls)).toHaveLength(0);
   });
 
   it('does not render the pagination when no add-ons in the collection', () => {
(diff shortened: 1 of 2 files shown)


tests/unit/amo/pages/TestCollection.js
```diff
@@ -771,6 +771,13 @@ describe(__filename, () => {
     });
   });
 
+  it('does not render a CollectionControls component when creating', () => {
+    const { store } = dispatchSignInActions();
+    const wrapper = renderComponent({ creating: true, store });
+
+    expect(wrapper.find(CollectionControls)).toHaveLength(0);
```
willdurand: I think you could move this assertion in the test case below (around L840)

`it('renders a create collection page', ...);`

xu3u4: Ok, thanks.

PR #6716, opened 2018-10-23
I6715-AddonTitle update

### LINKED ISSUE #6715: Make the AddonTitle component a bit more flexible
Right now this component only returns an h1 tag. If we could make this a bit more flexible for the guide pages (and other components) we could make use of this more. 

See the updated addon cards:

Alt text

REF:
https://mozilla.invisionapp.com/d/main/#/console/15640858/324863887/preview

rebmullin:
fixes #6715

### DIFF
--- a/src/amo/components/AddonTitle/index.js
+++ b/src/amo/components/AddonTitle/index.js
@@ -14,7 +14,7 @@ import type { I18nType } from 'core/types/i18n';
 import './styles.scss';
 
 type Props = {|
-  ComponentTag?: string,
+  as?: string,
   addon: AddonType | null,
   linkToAddon?: boolean,
 |};
@@ -26,7 +26,7 @@ type InternalProps = {|
 |};
 
 export const AddonTitleBase = ({
-  ComponentTag = 'h1',
+  as: Component = 'h1',
   addon,
   i18n,
   isRTL,
@@ -59,7 +59,7 @@ export const AddonTitleBase = ({
   }
 
   return (
-    <ComponentTag className="AddonTitle">
+    <Component className="AddonTitle">
       {addon ? (
         <React.Fragment>
           {linkToAddon ? (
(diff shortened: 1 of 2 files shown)


src/amo/components/AddonTitle/index.js
```diff
@@ -56,35 +58,37 @@ export const AddonTitleBase = ({
     });
   }
 
-  return (
-    <h1 className="AddonTitle">
```
willdurand: Quick question: why not declaring a prop like `component` that defaults to `h1` and use it here?

rebmullin: yeh let me try that... thanks!

src/amo/components/AddonTitle/index.js
```diff
@@ -14,6 +14,7 @@ import type { I18nType } from 'core/types/i18n';
 import './styles.scss';
 
 type Props = {|
+  ComponentTag?: string,
```
willdurand: react-bootstrap uses `as` as prop, and i find it nice, but maybe `component` would be enough.

src/amo/components/AddonTitle/index.js
```diff
@@ -25,6 +26,7 @@ type InternalProps = {|
 |};
 
 export const AddonTitleBase = ({
+  ComponentTag = 'h1',
```
willdurand: If you go for `as` or `component`, you can then alias the prop to `Component` so that it can be instantiated by React:

```suggestion
  as: Component = 'h1',
```

src/amo/components/AddonTitle/index.js
rebmullin: hmm..I thought ComponentTag was more cleared since it's not an actual component and just the tag name. But `as` looks interesting...
I'll take a look at that:)

src/amo/components/AddonTitle/index.js
rebmullin: @willdurand, I've updated using "as" but I am still wondering if that's super clear. Do you think a "tag" or "wrapperTag" prop might be more clear? 

set up in this same way tho:

`tag: Component = 'h1',`

PR #6720, opened 2018-10-23
I6718 AMInstall button label

### LINKED ISSUE #6718: Make the AMInstallButton accept a custom label
Currently you can't adjust the install label but we need this for the guide page. 

See mock:

Alt text

rebmullin:
fixes #6718

### DIFF
--- a/src/core/components/AMInstallButton/index.js
+++ b/src/core/components/AMInstallButton/index.js
@@ -47,6 +47,7 @@ import './styles.scss';
 type Props = {|
   ...WithInstallHelpersInjectedProps,
   addon: AddonType,
+  buttonLabel?: string,
   className?: string,
   defaultInstallSource: string,
   disabled: boolean,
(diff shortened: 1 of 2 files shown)


src/core/components/AMInstallButton/index.js
```diff
@@ -47,6 +47,7 @@ import './styles.scss';
 type Props = {|
   ...WithInstallHelpersInjectedProps,
   addon: AddonType,
+  buttonLabel?: string,
```
willdurand: how about we call it `defaultButtonText` because that's what it is, no?

rebmullin: I see what you're saying (since it's under default in the switch) but I guess I don't think of this as default really because it's not used if it's not passed in.

willdurand: mmh hat's the default state, though

src/core/components/AMInstallButton/index.js
```diff
@@ -216,6 +217,9 @@ export class AMInstallButtonBase extends React.Component<InternalProps> {
         return i18n.gettext('Uninstalling');
       case INACTIVE:
       default:
+        if (defaultButtonText) {
+          return defaultButtonText;
+        }
```
willdurand: I'd add a blank line below.

PR #6721, opened 2018-10-23
Code to support AddonMoreInfo with internal version

### LINKED ISSUE #6651: Update AddonMoreInfo to use a version stored in the versions redux state
Depends on #6634 

As part of removing `current_version` from `addon`, update `AddonMoreInfo` to use a version retrieved via `mapStateToProps` instead of using `addon.current_version`.

bobsilverberg:
Fixes #6651

### DIFF
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -5,6 +5,7 @@ import { compose } from 'redux';
 
 import AddonAdminLinks from 'amo/components/AddonAdminLinks';
 import Link from 'amo/components/Link';
+import { getVersionById } from 'amo/reducers/versions';
 import { STATS_VIEW } from 'core/constants';
 import translate from 'core/i18n/translate';
 import { hasPermission } from 'amo/reducers/users';
@@ -17,8 +18,9 @@ import {
 import Card from 'ui/components/Card';
 import DefinitionList, { Definition } from 'ui/components/DefinitionList';
 import LoadingText from 'ui/components/LoadingText';
-import type { I18nType } from 'core/types/i18n';
+import type { AddonVersionType } from 'amo/reducers/versions';
 import type { AppState } from 'amo/store';
+import type { I18nType } from 'core/types/i18n';
 
 type Props = {|
   addon: AddonType | null,
@@ -26,14 +28,15 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
+  hasStatsPermission: boolean,
   i18n: I18nType,
   userId: number | null,
-  hasStatsPermission: boolean,
+  version: AddonVersionType | null,
 |};
 
 export class AddonMoreInfoBase extends React.Component<InternalProps> {
   listContent() {
-    const { addon, i18n, userId, hasStatsPermission } = this.props;
+    const { addon, hasStatsPermission, i18n, userId, version } = this.props;
 
     if (!addon) {
       return this.renderDefinitions({
@@ -96,9 +99,8 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
       );
     }
 
-    const currentVersion = addon.current_version;
     const lastUpdated = addon.last_updated;
-    const license = currentVersion && currentVersion.license;
+    const license = version && version.license;
     let versionLicenseLink = null;
 
     if (license) {
@@ -122,9 +124,7 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
       supportEmail,
       statsLink,
       version:
-        currentVersion && addonHasVersionHistory(addon)
-          ? currentVersion.version
-          : null,
+        version && addonHasVersionHistory(addon) ? version.version : null,
       versionLastUpdated: lastUpdated
         ? i18n.sprintf(
             // translators: This will output, in English:
@@ -269,10 +269,15 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
   }
 }
 
-export const mapStateToProps = (state: AppState) => {
+export const mapStateToProps = (state: AppState, ownProps: Props) => {
+  const { addon } = ownProps;
   return {
-    userId: state.users.currentUserID,
     hasStatsPermission: hasPermission(state, STATS_VIEW),
+    userId: state.users.currentUserID,
+    version:
+      addon && addon.currentVersionId
+        ? getVersionById({ id: addon.currentVersionId, state: state.versions })
+        : null,
   };
 };
 
--- a/src/amo/components/AddonMoreInfo/index.js
+++ b/src/amo/components/AddonMoreInfo/index.js
@@ -31,12 +31,18 @@ type InternalProps = {|
   hasStatsPermission: boolean,
   i18n: I18nType,
   userId: number | null,
-  version: AddonVersionType | null,
+  currentVersion: AddonVersionType | null,
 |};
 
 export class AddonMoreInfoBase extends React.Component<InternalProps> {
   listContent() {
-    const { addon, hasStatsPermission, i18n, userId, version } = this.props;
+    const {
+      addon,
+      hasStatsPermission,
+      i18n,
+      userId,
+      currentVersion,
+    } = this.props;
 
     if (!addon) {
       return this.renderDefinitions({
@@ -100,7 +106,7 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
     }
 
     const lastUpdated = addon.last_updated;
-    const license = version && version.license;
+    const license = currentVersion && currentVersion.license;
     let versionLicenseLink = null;
 
     if (license) {
@@ -124,7 +130,9 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
       supportEmail,
       statsLink,
       version:
-        version && addonHasVersionHistory(addon) ? version.version : null,
+        currentVersion && addonHasVersionHistory(addon)
+          ? currentVersion.version
+          : null,
       versionLastUpdated: lastUpdated
         ? i18n.sprintf(
             // translators: This will output, in English:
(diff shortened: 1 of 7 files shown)


tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -188,42 +197,22 @@ describe(__filename, () => {
...
+    store.dispatch(_loadVersions({ license: { name: 'justText', url: null } }));
+    const root = render({});
     expect(root.find('.AddonMoreInfo-license-link')).toHaveLength(0);
 
     const link = root.find('.AddonMoreInfo-license-name');
     expect(link.children()).toIncludeText('justText');
   });
 
-  it('does not render any license info if the license is null', () => {
-    const addon = createInternalAddon({
```
rebmullin: what happens if the license is null? Is that case covered still?

src/amo/components/AddonMoreInfo/index.js
```diff
@@ -269,10 +269,15 @@ export class AddonMoreInfoBase extends React.Component<InternalProps> {
...
 }
 
-export const mapStateToProps = (state: AppState) => {
+export const mapStateToProps = (state: AppState, ownProps: Props) => {
+  const { addon } = ownProps;
   return {
-    userId: state.users.currentUserID,
     hasStatsPermission: hasPermission(state, STATS_VIEW),
+    userId: state.users.currentUserID,
+    version:
```
rebmullin: nit: could we still call this `currentVersion`? that name would still be relevant right? I also think `version.version` (above) would be a bit clearer if the object and property name were different..

What do you think?

tests/unit/amo/components/TestAddonMoreInfo.js
bobsilverberg: I think when I wrote this patch I had updated the type definition so that license could never be null, which is why I removed this test, but then, during the review cycle I made license allowed to be null, so you're right, we need this test back. I have brought it back.

src/amo/components/AddonMoreInfo/index.js
bobsilverberg: Good suggestion, I like it! I've made that change too.

bobsilverberg:
> @bobsilverberg, I am not sure if this is expected but the license section doesn't show up anymore on a few extension pages that I happen to check... but I see it showing up in dev.

Thanks @rebmullin. That's definitely a bug. I'm looking into it now and I can see using the Redux tools that the data in state isn't what I would have expected it to be. I'm not sure why yet. I'll try to figure that out, and also figure out how this wasn't caught by a test.

bobsilverberg:
I found and fixed the problem, @rebmullin, including adding a test for the condition that caused it. Thanks again for catching that! This is ready for another look.

src/core/types/addons.js
```diff
@@ -53,7 +53,7 @@ type PartialExternalAddonVersionType = {|
 export type ExternalAddonVersionType = {|
   ...PartialExternalAddonVersionType,
   // The `text` property is omitted from addon.current_version.license.
-  license: { name: string, url: string },
+  license?: { name: string, url: string },
```
kumar303: The type you changed here is supposed to have a defined `license`. The one without a license is [`PartialExternalAddonVersionType`](https://github.com/mozilla/addons-frontend/blob/4181a0940045b271fae07fec217f70a9da61709e/src/core/types/addons.js#L39). Sometimes it's hard to make this kind of thing work with Flow but could you check the place where you expected to get a warning about an undefined `license`? You might be able to fix it by declaring the input as `PartialExternalAddonVersionType | ExternalAddonVersionType`

bobsilverberg: Ah, that makes more sense. I wasn't sure what the heck `PartialExternalAddonVersionType` was even for! I wasn't actually getting any warnings without the `?`, but I knew from the code I had to change that when recommendations are returned we don't get a `license` or `release_notes` for the `current_version`.

If I remove the `?` Flow is happy, but I think maybe something is incorrect somewhere. I tried to use `PartialExternalAddonVersionType` but then Flow had me running in circles because suddenly there were all kinds of errors.

@kumar303 Do you think this is something that needs to be fixed, or is it okay to just leave it as is, without the `?`.

kumar303: > I wasn't sure what the heck PartialExternalAddonVersionType was even for!

Maybe add a comment above it that would help your former self understand it?

> is it okay to just leave it as is, without the `?`.

Yeah, I suggest doing that. It's a bit of a headache trying to get Flow to understand union types consisting of complex objects so no need to spend a lot of time on it.

bobsilverberg: Nevermind, I got it working, although I'm not sure if it's correct. I'll flag you for a review, @kumar303.

src/amo/reducers/recommendations.js
```diff
@@ -83,7 +87,7 @@ export const fetchRecommendations = ({
 };
 
 export type LoadRecommendationsParams = {|
-  addons: Array<ExternalAddonType>,
+  addons: Array<PartialExternalAddonType>,
```
kumar303: Yep, this matches the API docs 👍

src/core/types/addons.js
```diff
@@ -158,6 +158,11 @@ export type ExternalAddonType = {|
   weekly_downloads: number,
 |};
 
+export type PartialExternalAddonType = {|
+  ...ExternalAddonType,
+  current_version?: PartialExternalAddonVersionType,
```
kumar303: Maybe just remove this type?

It doesn't seem very useful since `ExternalAddonType` already has a `current_version` definition to accommodate both types:

https://github.com/mozilla/addons-frontend/blob/a6765efa490a36617f4c3d908c0c5e7f4cac8eb6/src/core/types/addons.js#L120

Alternatively, you could try fixing `ExternalAddonType`.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -17,12 +18,16 @@ import {
   fakeAddon,
   fakeI18n,
   fakeTheme,
+  fakeVersion,
   shallowUntilTarget,
 } from 'tests/unit/helpers';
 import LoadingText from 'ui/components/LoadingText';
 
 describe(__filename, () => {
-  const { store } = dispatchClientMetadata();
```
kumar303: D'oh. How did that creep in? Sigh.

I did a quick search. This seemed to be the only one.

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -89,15 +94,15 @@ describe(__filename, () => {
     delete partialAddon.homepage;
     delete partialAddon.support_email;
     delete partialAddon.support_url;
-    const root = render({ addon: createInternalAddon(partialAddon) });
```
kumar303: I guess this was serving the purpose of initializing `support_url` to an empty string (or whatever). Initializing to empty props seems like a good idea so maybe put this back?

tests/unit/amo/components/TestAddonMoreInfo.js
```diff
@@ -158,28 +163,32 @@ describe(__filename, () => {
...
 
-  it('renders the version number of an add-on', () => {
-    const addon = createInternalAddon({
-      ...fakeAddon,
-      current_version: {
-        ...fakeAddon.current_version,
-        version: '2.0.1',
-      },
+  const _loadVersions = (versionProps = {}) => {
+    return loadVersions({
```
kumar303: Usually we just include `store.dispatch()` in the helper itself since that's what you always need to do.

tests/unit/amo/reducers/test_versions.js
```diff
@@ -444,14 +444,49 @@ describe(__filename, () => {
...
+            id: versionId,
+          }),
+        ).toEqual(createInternalVersion(versionForRecommendations));
+      });
+
+      it('does not overwrite license and releaseNotes', () => {
+        const slug = 'some-slug';
+        let state = versionsReducer(
+          undefined,
+          loadVersions({ slug, versions: [version] }),
```
kumar303: I think this would be a better test if it explicitly defined `license` and `release_notes` rather than rely on implicit values.

src/core/types/addons.js
bobsilverberg: I'm confused. Are you suggesting doing away with `PartialExternalAddonType`? That's what I'm using for the type in the recommendations reducer, which you commented on seems correct. If I get rid of that then these changes don't really accomplish much, do they? To me the whole point was to document the the `loadRecommendations` action gets `PartialExternalAddonType` objects as opposed to `ExternalAddonType` objects.

tests/unit/amo/components/TestAddonMoreInfo.js
bobsilverberg: Are you just commenting that this was a problem before so it's good that I changed it, or are you pointing out something I need to change here?

tests/unit/amo/components/TestAddonMoreInfo.js
bobsilverberg: The problem here is that `partialAddon` is already an internal add-on object, so calling `createInternalAddon` and passing it in doesn't really make sense. `createInternalAddon` is only supposed to accept an `ExternalAddonType`.

tests/unit/amo/components/TestAddonMoreInfo.js
kumar303: Nothing to change. I was just surprised that the old code was so broken. Thanks for fixing it.

tests/unit/amo/components/TestAddonMoreInfo.js
kumar303: It makes sense to call `createInternalAddon()` a second time because several properties were deleted. The `createInternalAddon()` function will normalize those properties (I think).

bobsilverberg: As discussed on IRC, adding this back in does nothing.

Issue #6722, opened 2018-10-23
Log action types dispatched in a server response


kumar303:
I think it might be useful to log all action types that are dispatched in a server response to help visualize server state. It won't be as nice as full state logging (https://github.com/mozilla/addons-frontend/issues/6719) but it would be easier to implement, at least.

ioanarusiczki:
@kumar303 What should be tested on the frontend ? Please provide some STR or mark it "qa not needed".
Thanks!

willdurand:
@ioanarusiczki there is nothing to see for QA I think, I marked the issue as "qa: not needed". Thanks!

PR #6724, opened 2018-10-23
Use a minimal Redux logger during development on the server

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6722

I think this will be useful. Example:

screenshot 2018-10-23 12 48 53

### DIFF
--- a/src/core/store.js
+++ b/src/core/store.js
@@ -5,7 +5,7 @@ import { createLogger } from 'redux-logger';
 
 import log from 'core/logger';
 
-export const minimalReduxLogger = (store) => (next) => (action) => {
+export const minimalReduxLogger = () => (next) => (action) => {
   log.info('Dispatching', action.type);
   return next(action);
 };
--- a/src/core/store.js
+++ b/src/core/store.js
@@ -3,6 +3,13 @@ import config from 'config';
 import { applyMiddleware, compose } from 'redux';
 import { createLogger } from 'redux-logger';
 
+import log from 'core/logger';
+
+export const minimalReduxLogger = (store) => (next) => (action) => {
+  log.info('Dispatching', action.type);
+  return next(action);
+};
+
 /*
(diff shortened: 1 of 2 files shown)


src/core/store.js
```diff
@@ -3,26 +3,40 @@ import config from 'config';
 import { applyMiddleware, compose } from 'redux';
 import { createLogger } from 'redux-logger';
 
+import log from 'core/logger';
+
+export const minimalReduxLogger = () => (next) => (action) => {
```
bobsilverberg: The changes seem fine to me, but I don't really understand everything that is going on. What's actually happening here with this `next` and `action`?

tests/unit/core/test_store.js
```diff
@@ -87,4 +107,20 @@ describe(__filename, () => {
     expect(typeof middleware({ _config, _window })).toBe('function');
     sinon.assert.notCalled(_window.__REDUX_DEVTOOLS_EXTENSION__);
   });
+
+  describe('minimalReduxLogger', () => {
+    it('process an action', () => {
```
bobsilverberg: I don't really understand what's going on in this test either. I've read through it a few times, but there's always a point at which I get lost. :-(

src/core/store.js
kumar303: The signature of Redux middleware is to call `next(action)` which returns a new state after all reducers have processed the action. Docs: https://redux.js.org/advanced/middleware#the-final-approach

tests/unit/core/test_store.js
kumar303: It simulates how Redux middleware works.

kumar303: Did the documentation help? It's a well defined interface so I don't think it will change.

kumar303:
@bobsilverberg thanks for the review. I added some explanations but let me know if you think it could use a follow-up to make things more clear.

PR #6726, opened 2018-10-23
Update Addon to use a version stored in the versions redux state

### LINKED ISSUE #6653: Update Addon to use a version stored in the versions redux state
Depends on #6634 

As part of removing `current_version` from `addon`, update `Addon` to use a version retrieved via `mapStateToProps` instead of using `addon.current_version`.

bobsilverberg:
Fixes #6653

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -199,14 +199,18 @@ export class AddonBase extends React.Component {
   }
 
   renderRatingsCard() {
-    const { RatingManager, addon, i18n, location, version } = this.props;
+    const { RatingManager, addon, i18n, location, currentVersion } = this.props;
     let content;
     let footerPropName = 'footerText';
 
     let ratingManager;
-    if (addon && version) {
+    if (addon && currentVersion) {
       ratingManager = (
-        <RatingManager addon={addon} location={location} version={version} />
+        <RatingManager
+          addon={addon}
+          location={location}
+          version={currentVersion}
+        />
       );
     } else {
       ratingManager = (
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Thanks for the review @willdurand. I rebased this against master and also made a small change based on a suggestion from @rebmullin at https://github.com/mozilla/addons-frontend/pull/6721#discussion_r227821993.

src/amo/pages/Addon/index.js
```diff
@@ -622,10 +622,15 @@ export function mapStateToProps(state, ownProps) {
 
   let addonsByAuthors;
   let installedAddon = {};
+  let currentVersion = null;
 
   if (addon) {
     addonsByAuthors = getAddonsForSlug(state.addonsByAuthors, addon.slug);
     installedAddon = state.installations[addon.guid] || {};
+    currentVersion = getVersionById({
```
willdurand: I don't really like that this selector does not follow the existing conventions (that is: `getVersionId(versionsState,  id)`) but 🤷‍♂️

bobsilverberg: Do you mean because it accepts an object as opposed to two positional arguments?

bobsilverberg: We don't seem to be 100% consistent about that, e.g., https://github.com/mozilla/addons-frontend/blob/2018.10.25/src/amo/reducers/recommendations.js#L118-L126 (although admittedly I wrote that one too). Do feel that's an important convention to maintain? I feel like using an object is more explicit.

kumar303: I always try to avoid positional arguments whenever possible but that's mostly due to my experience with JavaScript before Flow and my experience with Flow before strict arity. Even after all that, positional arguments made up of primitives are still not very type safe -- it's hard for the compiler to detect if they are out of order.

willdurand: > Do you mean because it accepts an object as opposed to two positional arguments?

Yes.

I know we use objects and functions with arity = 1 a lot. I tend to like it because I am a fan of named arguments. That was just a comment I left because I could not simply skip it. I am not sure there is any actionable task behind it, except maybe discussing conventions at the All Hands.

PR #6728, opened 2018-10-23
I6740-guide page's saga/reducer set up

### LINKED ISSUE #6740: Set up guide page's saga/reducer
this is to support work for the guide pages #6661

rebmullin:
fixes #6740 

I might need some help with the tests here :)


src/core/reducers/search.js
```diff
@@ -21,6 +21,7 @@ export type FiltersType = {|
   page?: string,
   query?: Object,
   sort?: string,
+  guid?: string,
```
willdurand: Do you really need to change this file?

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+
+import { createPlatformFiles } from 'core/reducers/addons';
+import type { AddonType } from 'core/types/addons';
+
+export const FETCH_GUIDE_ADDONS: 'FETCH_GUIDE_ADDONS' = 'FETCH_GUIDE_ADDONS';
+export const LOAD_GUIDE_ADDONS: 'LOAD_GUIDE_ADDONS' = 'LOAD_GUIDE_ADDONS';
+
+export type GuideType = {|
+  addons: Array<AddonType> | void,
+|};
```
willdurand: Is it me or everything is compacted? Can you add some blank lines if that's the case?

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+import invariant from 'invariant';
+
+import { createPlatformFiles } from 'core/reducers/addons';
+import type { AddonType } from 'core/types/addons';
+
+export const FETCH_GUIDE_ADDONS: 'FETCH_GUIDE_ADDONS' = 'FETCH_GUIDE_ADDONS';
+export const LOAD_GUIDE_ADDONS: 'LOAD_GUIDE_ADDONS' = 'LOAD_GUIDE_ADDONS';
+
+export type GuideType = {|
+  addons: Array<AddonType> | void,
```
willdurand: It would be better to say `| null`.

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
+/* @flow */
+import invariant from 'invariant';
+
+import { createPlatformFiles } from 'core/reducers/addons';
+import type { AddonType } from 'core/types/addons';
+
+export const FETCH_GUIDE_ADDONS: 'FETCH_GUIDE_ADDONS' = 'FETCH_GUIDE_ADDONS';
+export const LOAD_GUIDE_ADDONS: 'LOAD_GUIDE_ADDONS' = 'LOAD_GUIDE_ADDONS';
+
+export type GuideType = {|
```
willdurand: I guess it is the `GuideState`, and this will be useful in the `amo/store` file.

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
+/* @flow */
```
willdurand: I'd name this file `guides.js`.

src/amo/reducers/guide.js
willdurand: It should probably be `GuidesSate` (plural).

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+import { createPlatformFiles } from 'core/reducers/addons';
+import type { AddonType } from 'core/types/addons';
+
+export const FETCH_GUIDE_ADDONS: 'FETCH_GUIDE_ADDONS' = 'FETCH_GUIDE_ADDONS';
+export const LOAD_GUIDE_ADDONS: 'LOAD_GUIDE_ADDONS' = 'LOAD_GUIDE_ADDONS';
+
+export type GuideType = {|
+  addons: Array<AddonType> | void,
+|};
+export type FetchGuideParams = {|
```
willdurand: It should probably be `FetchGuidesParams`

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+  guid: string,
+|};
+export type FetchGuideAction = {|
+  type: typeof FETCH_GUIDE_ADDONS,
+  payload: FetchGuideParams,
+|};
+export type LoadGuideAction = {|
+  type: typeof LOAD_GUIDE_ADDONS,
+  payload: GuideType,
+|};
```
willdurand: Can you add the types above each function that uses the types?

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+export const initialState: GuideType = {
+  addons: undefined,
```
willdurand: Yep, would be better to use `null` here.

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+};
+export const loadGuideAddons = ({ addons }: GuideType): LoadGuideAction => {
+  invariant(addons, 'addons is required');
```
willdurand: s/is/are?

src/amo/reducers/guide.js
```diff
@@ -0,0 +1,72 @@
...
+
+  const mappedResults =
+    addons &&
+    addons.map((addon) => {
+      return {
+        ...addon,
+        platformFiles: createPlatformFiles(addon.current_version),
```
willdurand: Why exactly do you need this?

Actually, I am not sure add-ons should be stored in this reducer. We have a reducer (`addons`) for the add-ons. I'd store everything there, using `loadAddonResults()`.

rebmullin: Well we need the the search api for this so we can pass multiple guids..

I didn't see the addons within the search results with this info which we need for the AMIntallButton..

src/core/reducers/search.js
willdurand: ah, that's because you use `SearchParams` here. I missed that, so I guess it is alright. (I did not expect this file to be changed)

src/amo/reducers/guides.js
```diff
@@ -0,0 +1,76 @@
...
+
+export type LoadGuidesAction = {|
+  type: typeof LOAD_GUIDES_ADDONS,
+  payload: GuidesState,
+|};
+
+export const initialState: GuidesState = {
+  addons: null,
+};
+export const loadGuidesAddons = ({ addons }: GuidesState): LoadGuidesAction => {
```
willdurand: I believe this is not needed. Let's add the addons in the `addons` reducer instead.

rebmullin: hey @willdurand, can you please clarify what you mean?

are you saying that the redux structure should be something like ~:
```javascript
addons 
  ByID: {}
  ByGuid: {}
  ...etc
  guides: [] // add here?
```

Also, I wasn't clear what isn't needed..

willdurand: don't load the addons in this reducer, use the existing `addons` reducer.

rebmullin: okay I'll revisit this.. thanks!

willdurand: Sorry but I don't understand. You fetch addons from the search API. Instead of loading them in a `guides` reducer, you can use the existing `loadAddonResults()` action in your `guides` saga, no? That would load all the add-ons in the `addons` reducer.

You should not have to touch the addons reducer code.

rebmullin:
hey @willdurand, I've made some changes. Can you please take another look?

src/amo/reducers/guides.js
```diff
@@ -0,0 +1,79 @@
...
+  state: GuidesState = initialState,
+  action: ActionType,
+): GuidesState => {
+  switch (action.type) {
+    case FETCH_GUIDES_ADDONS:
+      return {
+        ...state,
+        loading: true,
+      };
+    case UPDATE_GUIDES_ADDONS_LOADED_STATUS:
```
willdurand: You probably don't need this. The good thing about Redux is that ALL reducers get ALL actions, so you could just "listen to" `LOAD_ADDON_RESULTS`. I'd say that when you get `LOAD_ADDON_RESULTS`, then you set `loading` to `false`, and when you fetch, you set it to `true`.

src/amo/sagas/guides.js
```diff
@@ -0,0 +1,45 @@
...
+} from 'amo/reducers/guides';
+import log from 'core/logger';
+import { createErrorHandler, getState } from 'core/sagas/utils';
+import { loadAddonResults } from 'core/reducers/addons';
+import type { SearchParams } from 'core/api/search';
+import type { FetchGuidesAction } from 'amo/reducers/guides';
+
+export function* fetchGuidesAddons({
+  payload: { errorHandlerId, guid },
+}: FetchGuidesAction): Generator<any, any, any> {
```
willdurand: You can probably use `Saga` from `core/types/sagas`.

src/amo/sagas/guides.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const params: SearchParams = {
+      api: state.api,
+      filters: {
+        guid,
+      },
+    };
+
+    const guideAddons = yield call(searchApi, params);
+    const { results } = guideAddons;
```
willdurand: You can probably do that in the line above:

```js
const { results } yield call(searchApi, params);
```

src/amo/sagas/guides.js
```diff
@@ -0,0 +1,45 @@
...
+
+    yield put(loadAddonResults({ addons: results }));
+  } catch (error) {
+    log.warn('Search for guide addons failed:', error);
+    yield put(errorHandler.createErrorAction(error));
+  }
+
+  yield put(updateGuideAddonsLoadedStatus({ loading: false }));
```
willdurand: I believe you can omit this.

src/amo/sagas/guides.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export default function* guideAddonsSaga(): Generator<any, any, any> {
```
willdurand: You can probably use `Saga` from `core/types/sagas`.

src/amo/store.js
```diff
@@ -66,6 +67,7 @@ type InternalAppState = {|
   errorPage: ErrorPageState,
   errors: Object,
   formOverlay: FormOverlayState,
+  guides: Object,
```
willdurand: You should use `GuidesState` instead of `Object`.

src/amo/sagas/guides.js
rebmullin: of course! I looked at one point but needed to rebase and forgot to circle back. I'll do that now..

src/amo/reducers/guides.js
rebmullin: I didn't realize this but this makes sense. thanks for the tip!

rebmullin:
hey @willdurand. I've made some updates per your comments. Can you please take another look? :)

src/amo/reducers/guides.js
```diff
@@ -0,0 +1,61 @@
+/* @flow */
+import invariant from 'invariant';
+
+import { LOAD_ADDON_RESULTS } from 'core/reducers/addons';
+
+export const FETCH_GUIDES_ADDONS: 'FETCH_GUIDES_ADDONS' = 'FETCH_GUIDES_ADDONS';
+export const UPDATE_GUIDES_ADDONS_LOADED_STATUS: 'UPDATE_GUIDES_ADDONS_LOADED_STATUS' =
+  'UPDATE_GUIDES_ADDONS_LOADED_STATUS';
```
willdurand: You can remove this constant now.

tests/unit/amo/reducers/test_guides.js
```diff
@@ -0,0 +1,30 @@
...
+      expect(state).toEqual(initialState);
+    });
+
+    it('updates the loading flag status', () => {
+      const state = guidesReducer(
+        undefined,
+        fetchGuidesAddons({ guid: 'test', errorHandlerId: 'test' }),
+      );
+
+      const newState = guidesReducer(state, { type: FETCH_GUIDES_ADDONS });
```
willdurand: I am not sure I understand why you did that? You should call `fetchGuidesAddons` here, but it is done 2 lines above.

tests/unit/amo/reducers/test_guides.js
```diff
@@ -0,0 +1,30 @@
...
+      expect(newState.loading).toEqual(true);
+
+      const afterAddonsLoadedState = guidesReducer(state, {
+        type: LOAD_ADDON_RESULTS,
+      });
```
willdurand: Can you call `loadAddonResults()` instead (the action creator rather than writing the action by hand)

tests/unit/amo/sagas/test_guides.js
```diff
@@ -0,0 +1,93 @@
...
+  const guid = 'support@lastpass.com,{b76ed4e7-12a6-4f25-a27b-fc3f93289008}';
+
+  let errorHandler;
+  let mockApi;
+  let sagaTester;
+
+  beforeEach(() => {
+    errorHandler = createStubErrorHandler();
+    mockApi = sinon.mock(searchApi);
+    const initialState = dispatchClientMetadata().state;
```
willdurand: Can you have a blank line before and after that line?

src/amo/reducers/guides.js
```diff
@@ -0,0 +1,61 @@
...
+  loading: boolean,
+|};
+
+export const initialState: GuidesState = {
+  loading: false,
+};
+
+export type FetchGuidesParams = {|
+  errorHandlerId: string,
+  guid: string,
```
willdurand: shouldn't this be `guids` since you can have multiple `guid`?

tests/unit/amo/sagas/test_guides.js
```diff
@@ -0,0 +1,93 @@
...
+        })
+        .once()
+        .resolves(guideAddons);
+
+      _fetchGuidesAddons({ guid });
+
+      const { results } = guideAddons;
+
+      const expectedAction = loadAddonResults({ addons: results });
+      const loadAction = await sagaTester.waitFor(expectedAction.type);
```
willdurand: s/loadAction/action

tests/unit/amo/sagas/test_guides.js
```diff
@@ -0,0 +1,93 @@
...
+        fetchGuidesAddons({
+          errorHandlerId: errorHandler.id,
+          ...params,
+        }),
+      );
+    }
+
+    it("calls the API to fetch a guide's addons", async () => {
+      const state = sagaTester.getState();
+      const guideAddons = { results: [fakeAddon, fakeAddon] };
```
willdurand: I'd introduce a variable named `addons` for the array of `fakeAddon`. And I'd write the following instead:

```
.resolves({ results: addons });
```

src/amo/reducers/guides.js
rebmullin: ugh sorry about that..I missed that :/

tests/unit/amo/reducers/test_guides.js
rebmullin: yep, you're right 🤦‍♀️

tests/unit/amo/sagas/test_guides.js
rebmullin: maybe to help avoid this in the future (since I see this a lot it seems), when should I have blank lines?

src/amo/reducers/guides.js
rebmullin: well, I did think about that but I kept as guid since that matches the search filter name. I can change this tho..

tests/unit/amo/reducers/test_guides.js
```diff
@@ -0,0 +1,31 @@
...
+
+    it('updates the loading flag status', () => {
+      const state = guidesReducer(
+        undefined,
+        fetchGuidesAddons({ guids: 'test,test2', errorHandlerId: 'test' }),
+      );
+
+      expect(state.loading).toEqual(true);
+
+      const afterAddonsLoadedState = guidesReducer(
```
willdurand: s/afterAddonsLoadedState/newState maybe

src/amo/reducers/guides.js
```diff
@@ -0,0 +1,59 @@
...
+  loading: boolean,
+|};
+
+export const initialState: GuidesState = {
+  loading: false,
+};
+
+export type FetchGuidesParams = {|
+  errorHandlerId: string,
+  guids: string,
```
willdurand: Maybe it would be less error prone to use `Array<string>` here and to do `.join(',')` in the saga. That would also communicate the intent better.

src/amo/reducers/guides.js
rebmullin: I ended up changing this to guids. I do think that is a better/clearer since there will most likely be multiple guids in this case.

src/amo/reducers/guides.js
rebmullin: I started work on the Guide Page itself and I am doing the `join(`) there. I am not sure if it really matters but I guess I can move this over to the saga..

willdurand: it would be better to move it because expecting a string with comma separated values is a weak contract. If it is done in the saga, then you can enforce it only once and callers don't need to know that (the comma separated values become an implementation detail)

Issue #6729, opened 2018-10-23
Initialize the client with a predictable Redux state object


kumar303:
For sanity, we should initialize the client (i.e. rendered HTML) with a predictable Redux state object.

**For QA**: This change could affect the entire app but only for the first page load. Please test as much as possible with this approach:
* Load the site, navigate to another page
* Click the browser refresh button
* Make sure the page looks correct after refresh
* Navigate somewhere else, click refresh, verify, and repeat...

<hr />

How is the current Redux state object unpredictable, you ask? Well, let me begin by *mourning the hours of my life lost to this bug*. Thanks, let's continue. You can see it in action like this:

* Go to https://addons-dev.allizom.org/en-US/firefox/addon/icloud-test-dev/ in Chrome so you can use Redux Devtools. Be sure to load the URL directly instead of navigating there from another page.
* With Redux Devtools, select the `@@INIT` action and open the *State* tab to see the initial state that the page was loaded with.
* Open up `addonsByAuthors` and you should see something like: `loadingFor: { ValentinaDDD-extension: true }`. 

This means the client will think it's in a loading state **yet nothing will ever load**! 

If you use the logger from https://github.com/mozilla/addons-frontend/pull/6724 then you'll see how the final server render dispatches `FETCH_ADDONS_BY_AUTHORS` just quick enough to modify state before `ServerHtml` renders its HTML.

Funny enough, the `AddonsByAuthors` component does not check if data has already been loaded or is in the process of loading before [beginning to load data](https://github.com/mozilla/addons-frontend/blob/5e45570bdd5f27db987a62385297e0b332ae97d8/src/amo/components/AddonsByAuthorsCard/index.js#L97). Hmm, I wonder if there is a trail of tears behind that decision! All other components would have trouble here because it would be impossible to know if it is safe to begin loading data or not.

ioanarusiczki:
I tested this on a new profile, loaded AMO dev with FF62(Win10), went on another site then returned and refreshed AMO (mostly I used links that redirect from AMO to other pages). The homepage reloads without any layout issues.
![issue with refreshing page](https://user-images.githubusercontent.com/33448286/47421809-08ecde80-d78a-11e8-8b4e-fe9525352521.gif)

willdurand:
@ioanarusiczki I am marking this issue as "qa: extra" because we need to repeat the STR for all pages of AMO. This change is rather critical.

ioanarusiczki:
[...]

PR #6730, opened 2018-10-23
Initialize the client with a predictable app state

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6729 by serializing the application state immediately after all sagas have finished and *before* the final application component begins rendering. 

This ensures that the client always has a predictable app state. In other words, we already know that we only get *one* chance to load data: the first render and any saga that was triggered by it. This patch makes that implementation detail predictable. 

Before this patch, the constructors of components in the *second* render might have been able to modify the initial state, depending on timing. I don't know if any page was relying on that buggy behavior but I tested cold loads of many pages with this patch and they seem to work.

The steps in https://github.com/mozilla/addons-frontend/issues/6729 for reproducing the bug are no longer possible after this patch. I no longer see the page render in an infinite loading state.

### DIFF
--- a/src/core/components/ServerHtml/index.js
+++ b/src/core/components/ServerHtml/index.js
@@ -15,6 +15,7 @@ const JS_CHUNK_EXCLUDES = new RegExp(
 export default class ServerHtml extends Component {
   static propTypes = {
     appName: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


willdurand:
let's merge this asap, so that QA can make sure nothing goes wrong.

PR #6739, opened 2018-10-24
Extract a CategoryHead component

### LINKED ISSUE #6738: Extract a CategoryHead component
In order to add new `meta` tags to the category pages (#5767), let's create a `CategoryHead` component that is similar to the `AddonHead` component. This new component will contain all the `Helmet` logic for a category page.

For QA: the category pages should have correct HTML titles and `canonical` links (in the HTML).

willdurand:
Fixes #6738

### DIFF
--- a/src/amo/components/CategoryHead/index.js
+++ b/src/amo/components/CategoryHead/index.js
@@ -15,7 +15,6 @@ import type { I18nType } from 'core/types/i18n';
 
 type Props = {|
   category: CategoryType | null,
-  type: 'extension' | 'theme',
 |};
 
 type InternalProps = {|
@@ -27,12 +26,12 @@ type InternalProps = {|
 
 export class CategoryHeadBase extends React.PureComponent<InternalProps> {
   getPageTitle() {
-    const { category, i18n, type } = this.props;
+    const { category, i18n } = this.props;
 
     invariant(category, 'category is required');
 
     let title;
-    switch (type) {
+    switch (category.type) {
       case ADDON_TYPE_EXTENSION:
         title = i18n.gettext('%(categoryName)s – Extensions');
         break;
--- a/src/amo/components/CategoryHead/index.js
+++ b/src/amo/components/CategoryHead/index.js
@@ -1,0 +1,99 @@
+/* @flow */
+import config from 'config';
+import invariant from 'invariant';
+import * as React from 'react';
+import Helmet from 'react-helmet';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+import { getCanonicalURL } from 'amo/utils';
+import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import translate from 'core/i18n/translate';
+import type { AppState } from 'amo/store';
+import type { I18nType } from 'core/types/i18n';
+
+type CategoryType = {|
+  application: string,
+  description: string,
+  id: number,
+  misc: boolean,
+  name: string,
+  slug: string,
+  type:
+    | 'extension'
+    | 'theme'
+    | 'dictionary'
+    | 'search'
+    | 'language'
+    | 'persona',
+  weight: number,
+|};
+
+type Props = {|
+  category: CategoryType | null,
+  type: 'extension' | 'theme' | null,
+|};
+
+type InternalProps = {|
+  ...Props,
+  _config: typeof config,
+  i18n: I18nType,
+  locationPathname: string,
+|};
+
+export class CategoryHeadBase extends React.PureComponent<InternalProps> {
+  getPageTitle() {
+    const { category, i18n, type } = this.props;
+
+    invariant(category, 'category is required');
+
+    let title;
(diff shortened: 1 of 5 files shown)


src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,99 @@
...
+  misc: boolean,
+  name: string,
+  slug: string,
+  type:
+    | 'extension'
+    | 'theme'
+    | 'dictionary'
+    | 'search'
+    | 'language'
+    | 'persona',
```
rebmullin: do we need all of these types? on the current category pages, I only show extension and persona? let me know if I am missing something..

src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,99 @@
...
+import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import translate from 'core/i18n/translate';
+import type { AppState } from 'amo/store';
+import type { I18nType } from 'core/types/i18n';
+
+type CategoryType = {|
+  application: string,
+  description: string,
+  id: number,
+  misc: boolean,
```
rebmullin: hmm..I wonder what "misc" is for?

src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,99 @@
...
+    | 'dictionary'
+    | 'search'
+    | 'language'
+    | 'persona',
+  weight: number,
+|};
+
+type Props = {|
+  category: CategoryType | null,
+  type: 'extension' | 'theme' | null,
```
rebmullin: will type ever be null?

src/amo/components/CategoryHead/index.js
rebmullin: I see...maybe this is for the "other" category..

willdurand: that's part of the API response: https://addons-server.readthedocs.io/en/latest/topics/api/categories.html#get--api-v4-addons-categories-

src/amo/components/CategoryHead/index.js
willdurand: The API returns more categories than that, I, listed all thd types mentioned in https://addons-server.readthedocs.io/en/latest/topics/api/categories.html#get--api-v4-addons-categories-

src/amo/components/CategoryHead/index.js
willdurand: I am not sure, maybe not. Let me change that

src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+    let title;
+    switch (type) {
+      case ADDON_TYPE_EXTENSION:
+        title = i18n.gettext('%(categoryName)s – Extensions');
+        break;
+      case ADDON_TYPE_THEME:
+        title = i18n.gettext('%(categoryName)s – Themes');
+        break;
+      default:
```
rebmullin: will the default ever hit if type is always defined? Maybe this could be an if / else?

src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,83 @@
...
+import { getCanonicalURL } from 'amo/utils';
+import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import translate from 'core/i18n/translate';
+import type { AppState } from 'amo/store';
+import type { CategoryType } from 'amo/types/categories';
+import type { I18nType } from 'core/types/i18n';
+
+type Props = {|
+  category: CategoryType | null,
+  type: 'extension' | 'theme',
```
rebmullin: shouldn't type be extension or persona?

src/amo/components/CategoryHead/index.js
willdurand: This is more future-proof.

src/amo/components/CategoryHead/index.js
willdurand: Indeed, but I guess we could even remove this prop. Each category has a `type`.

rebmullin: yep that's better 👍

src/amo/components/CategoryHead/index.js
```diff
@@ -0,0 +1,82 @@
...
+  _config: typeof config,
+  i18n: I18nType,
+  locationPathname: string,
+|};
+
+export class CategoryHeadBase extends React.PureComponent<InternalProps> {
+  getPageTitle() {
+    const { category, i18n } = this.props;
+
+    invariant(category, 'category is required');
```
rebmullin: isn't this covered by this check below:

```javascript
if (!category) {
  return null;
}
```

willdurand: It is but Flow does not get it, that's why I use an `invariant` call here.

willdurand:
Thanks!

Issue #6717, opened 2018-10-23
Make the user object available to SSR rendered components


kumar303:
### Describe the problem and steps to reproduce it:

* Create a component that waits for `siteUser` (e.g. [`getCurrentUser(state.users)`](https://github.com/mozilla/addons-frontend/blob/a02777625bb33f8eebff8eea38420e6b11c845f6/src/amo/reducers/users.js#L374)) before it dispatches a loading action
* Create a page requiring that component to load all its data on the server, before a response is written

### What happened?

This is not possible because `siteUser` does not predictably load before the [first server render](https://github.com/mozilla/addons-frontend/blob/44a07385698e25abad6247cb4ac8bbb53edd0642/src/core/server/base.js#L363)

### What did you expect to happen?

It should be possible to rely on `siteUser` when loading data in a server response.

### Anything else we should know?

* SSR data loading is already very fragile. We only get *one* render cycle to dispatch all loading sagas. Really, we need a holistic fix to this that lets us load all necessary data on the server in a predictable manner. Fixing this would be a big architectural change.
* Instead of loading `currentUser` [in a saga](https://github.com/mozilla/addons-frontend/blob/a02777625bb33f8eebff8eea38420e6b11c845f6/src/amo/sagas/users.js#L22) (which is triggered [right after parsing the cookie](https://github.com/mozilla/addons-frontend/blob/44a07385698e25abad6247cb4ac8bbb53edd0642/src/core/server/base.js#L309)), we could use a promise and make sure this promise completes before the first render. This would be a quick fix but it may have performance implications.

kumar303:
For context, I ran into this when trying to render user review permissions on the server for https://github.com/mozilla/addons/issues/526 . However, I am able to just load the necessary data on the client.

bobsilverberg:
Kumar and I discussed this a bit at ReactConf, although we didn't discuss any possible solutions. I think we agreed that we should discuss this as a group at the next all-hands, discussing the problems and challenges we face with our current approach and possible soltuions.

kumar303:
It would be nice if we could think of a non-drastic solution, one that worked with our current saga strategy. My only idea for that is to keep rendering the app in a loop until it stops triggering sagas. Ha! 

Another nice thing would be to load data on the server exactly how we load it on the client (however we decide to do it).

@willdurand I think you had some ideas around this?

PR #6745, opened 2018-10-25
Let unlisted add-on owners/developers reply to reviews

kumar303:
Fixes https://github.com/mozilla/addons/issues/526

Originally I wanted to modify the current [`fetchReviews`](https://github.com/mozilla/addons-frontend/blob/12ad39894c60fa462ce06edb8b60029d8e029d5a/src/amo/pages/AddonReviewList/index.js#L128-L137) call but this is complicated because I can't depend on `siteUser` (https://github.com/mozilla/addons-frontend/issues/6717). My current patch just introduces a separate dispatch which is fine, I guess. We could fix it up later.

### DIFF
--- a/src/amo/components/AddonReviewCard/index.js
+++ b/src/amo/components/AddonReviewCard/index.js
@@ -71,6 +71,7 @@ type InternalProps = {|
   i18n: I18nType,
   replyingToReview: boolean,
   siteUser: UserType | null,
+  siteUserCanManageReplies: boolean,
   siteUserHasReplyPerm: boolean,
   submittingReply: boolean,
 |};
@@ -283,11 +284,6 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
     return i18n.gettext('Keep review');
   }
 
-  siteUserCanManageReplies() {
-    const { siteUserCanReply, siteUserHasReplyPerm } = this.props;
-    return siteUserCanReply || siteUserHasReplyPerm;
-  }
-
   renderReply() {
     const {
       addon,
@@ -354,6 +350,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       showControls,
       showRating,
       siteUser,
+      siteUserCanManageReplies,
       slim,
     } = this.props;
 
@@ -408,7 +405,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
       review &&
       siteUser &&
       (review.userId === siteUser.id ||
-        (this.isReply() && this.siteUserCanManageReplies()));
+        (this.isReply() && siteUserCanManageReplies));
 
     const controls = controlsAreVisible ? (
       <div className="AddonReviewCard-allControls">
@@ -455,7 +452,7 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
         !replyingToReview &&
         !review.reply &&
         !this.isReply() &&
-        this.siteUserCanManageReplies() &&
+        siteUserCanManageReplies &&
         siteUser &&
         review.userId !== siteUser.id ? (
           <a
@@ -559,13 +556,16 @@ export function mapStateToProps(state: AppState, ownProps: Props) {
     }
   }
 
+  const siteUserHasReplyPerm = hasPermission(state, ADDONS_EDIT);
+
   return {
     beginningToDeleteReview,
     deletingReview,
     editingReview,
     replyingToReview,
     siteUser: getCurrentUser(state.users),
-    siteUserHasReplyPerm: hasPermission(state, ADDONS_EDIT),
+    siteUserCanManageReplies: ownProps.siteUserCanReply || siteUserHasReplyPerm,
+    siteUserHasReplyPerm,
     submittingReply,
   };
 }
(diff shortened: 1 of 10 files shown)


src/amo/components/AddonReviewCard/index.js
```diff
@@ -285,22 +284,8 @@ export class AddonReviewCardBase extends React.Component<InternalProps> {
...
-      return _siteUserCanManageReplies();
-    }
-    if (!siteUser) {
-      return false;
-    }
-    return (
-      isAddonAuthor({ addon, userId: siteUser.id }) || siteUserHasReplyPerm
-    );
+    const { siteUserCanReply, siteUserHasReplyPerm } = this.props;
+    return siteUserCanReply || siteUserHasReplyPerm;
```
willdurand: I believe this could live in an injected prop instead of a method now.

src/amo/pages/AddonReviewList/index.js
```diff
@@ -63,7 +71,8 @@ type InternalProps = {|
   pageSize: number | null,
   reviewCount?: number,
   reviews?: Array<UserReviewType>,
-  areReviewsLoading: boolean,
+  siteUser: UserType | null,
+  siteUserCanReplyToReviews: ?boolean,
```
willdurand: Can you use `| void` instead? I like it better because it clearly shows that `undefined` as a value is important (which seems to be the case in `componentDidMount()`). `null` is [allowed here](https://flow.org/en/docs/types/maybe/), and it is not clear to me whether it is wanted.

That being said, do you use `undefined` here because the selector returns `undefined`? I don't really like this mix of values vs undefined vs null. I especially don't like `undefined` because if you forget a `return` or something, Flow won't complain.

We should really think about using a solid [Option/Maybe lib](https://github.com/alexanderjarvis/maybe)...

</ rant >

tests/unit/amo/pages/TestAddonReviewList.js
```diff
@@ -692,6 +714,129 @@ describe(__filename, () => {
...
+      expect(items).toHaveLength(2);
+
+      expect(items.at(0)).toHaveProp('siteUserCanReply', true);
+      expect(items.at(1)).toHaveProp('siteUserCanReply', true);
+    });
+
+    it('passes siteUserCanReply to FeaturedAddonReview', () => {
+      const reviewId = 8765;
+      const addon = { ...fakeAddon };
+      dispatchAddon(addon);
```
willdurand: This should really be called `_loadAddonResults()`

tests/unit/amo/pages/TestAddonReviewList.js
```diff
@@ -692,6 +714,129 @@ describe(__filename, () => {
...
+      dispatchAddonReviews({ reviews: [{ ...fakeReview }] });
```
willdurand: and this `_setAddonReviews()`

tests/unit/amo/pages/TestAddonReviewList.js
```diff
@@ -91,6 +98,21 @@ describe(__filename, () => {
     store.dispatch(action);
   };
 
+  const dispatchUserReviewPermissions = ({
```
willdurand: We can probably call this function: `_signInUserAndSetReviewPermissions`

src/amo/pages/AddonReviewList/index.js
kumar303: Whoops, the question mark type was not intended. We could benefit from a lint rule to disallow question mark types in favor of explicit `null` and/or `void` types but there is a lot of existing code to fix.

> do you use undefined here because the selector returns undefined? I don't really like this mix of values vs undefined vs null.

I was allowing `undefined` because the reducer sets `canReplyToReviews` to `undefined` before it has been fetched it from the API. That is a pattern I'm used to: `undefined` usually means it has *not* been fetched from the API whereas `null` means it has been fetched from the API and does not exist. However, that isn't relevant here since the API result will always be a `boolean`. Thus, I changed this initialization to `null`.

>  I especially don't like undefined because if you forget a return or something, Flow won't complain.

If you unintentionally omit a return value from a function, Flow or lint usually complains. Lint will complain about inconsistent return values and Flow will complain if the returned value doesn't match another input type. But, yeah, JavaScript. I changed the initialization to `null` as suggested.

> We should really think about using a solid Option/Maybe lib...

I would rather stick to primitives, if we can. I think with Flow coverage it's pretty safe to make comparisons to primitives. We could clean up our consistency for how we do it, though.

willdurand: > That is a pattern I'm used to: `undefined` usually means it has _not_ been fetched from the API whereas `null` means it has been fetched from the API and does not exist.

The problem with this "pattern" is that it is your pattern, and no one is supposed to know it because I don't think it is documented or enforced. I know it because you left comments in the past, but contributors or other developers might not know it.

A lib for true Option/Maybe types would allow us to introduce this pattern (pretty much) in an robust and explicit manner.

kumar303: Yes, we definitely need to improve our docs for which patterns we use on the site. However, anyone who has tried to build an SSR compliant loading state on AMO should know the pattern by now.

kumar303:
Thanks for the review

Issue #5767, opened 2018-07-27
Add <meta name="description”> for all other pages


muffinresearch:
Following on from https://github.com/mozilla/addons-frontend/issues/5766 we should also add descriptions to all other relevant pages represented by the new frontend.

See also https://docs.google.com/document/d/1W9LOw7wgbpY6Mw5Zd4V71hJ0YahY4DTNl23MHVpG0_E/edit?ts=5b5b40e5#heading=h.aqjhycqt8ulj

## Pages

⚠️ all of the page below are available for Firefox and Android (Firefox for Android):

- [x] Homepage: https://addons.mozilla.org/en-US/firefox/
- [x] Landing Extensions: https://addons.mozilla.org/en-US/firefox/extensions/
- [x] Landing Themes: https://addons.mozilla.org/en-US/firefox/themes/
- [x] Search tools: https://addons.mozilla.org/en-US/firefox/search-tools/
- [x] Reviews for an add-on: https://addons.mozilla.org/en-US/firefox/addon/user-agent-string-switcher/reviews/ --> https://github.com/mozilla/addons-frontend/pull/6712
- [x] user profile of a developer or artist or both: https://addons.mozilla.org/en-US/firefox/user/Testfahrer/
- Themes / category pages (https://addons.mozilla.org/en-US/firefox/themes/):
  - [x] Abstract
  - [x] Causes
  - [x] Fashion
  - [x] Film and TV
  - [x] Firefox
  - [x] Foxkeh
  - [x] Holiday
  - [x] Music
  - [x] Nature
  - [x] Other
  - [x] Scenery
  - [x] Seasonal
  - [x] Solid
  - [x] Sports
  - [x] Websites
- [x] About page: https://addons.mozilla.org/en-US/about
- [x] Review guide: https://addons.mozilla.org/en-US/review_guide
- [x] Collection pages (we can probably rely on the description of the collection, but there might be something to add before): https://addons.mozilla.org/en-US/firefox/collections/mozilla/ad-blockers/

Firefox-only pages:

- [x] Dictionaries and Lang Packs: https://addons.mozilla.org/en-US/firefox/language-tools/
- Extensions / category pages (https://addons.mozilla.org/en-US/firefox/extensions/):
  - [x] Alerts & Updates
  - [x] Appearance
  - [x] Bookmarks
  - [x] Download Management
  - [x] Feeds, News & Blogging
  - [x] Games & Entertainment
  - [x] Language Support
  - [x] Other
[...]

PR #6747, opened 2018-10-25
Add description meta tag on category pages

willdurand:
Refs #5767
~~:warning: depends on #6739~~

---

The categories API now exposes the descriptions, so we can just use them in the frontend.

### DIFF
--- a/tests/unit/amo/components/TestCategoryHead.js
+++ b/tests/unit/amo/components/TestCategoryHead.js
@@ -65,4 +65,30 @@ describe(__filename, () => {
 
     expect(root.find('title')).toHaveText(category.name);
   });
+
+  it('renders a "description" meta tag when category description is available', () => {
+    const description = 'some description for a category';
+    const category = { ...fakeCategory, description };
+    const root = render({ category });
+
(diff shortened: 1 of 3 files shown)


willdurand:
rebased!

tests/unit/amo/components/TestCategoryHead.js
```diff
@@ -65,4 +65,30 @@ describe(__filename, () => {
...
+    const root = render({ category });
+
+    expect(root.find('meta[name="description"]')).toHaveLength(1);
+    expect(root.find('meta[name="description"]')).toHaveProp(
+      'content',
+      description,
+    );
+  });
+
+  it('does not render a "description" meta tag when category description is null', () => {
```
rebmullin: I'm looking at api docs and it doesn't say this could be null. Usually I've seen string|null so just wanted to double check this :)

willdurand: it should say that given that before my patch on the server, all descriptions were null.. I'll update the API doc ☺️

willdurand: I filed: https://github.com/mozilla/addons-server/issues/9818

PR #6756, opened 2018-10-26
Add Flow types to the user saga

### LINKED ISSUE #5080: Add Flow types to the user saga
The `user` saga does not have Flow types because the initial thoughts about Flow were to use it one some specific files, but since #4523, we should add Flow types in sagas too 🎉 

Let's do this.

willdurand:
Fixes #5080

### DIFF
--- a/src/core/types/sagas.js
+++ b/src/core/types/sagas.js
@@ -1,0 +1,3 @@
+/* @flow */
+
+export type Saga = Generator<any, any, any>;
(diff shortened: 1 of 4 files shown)


src/core/types/sagas.js
```diff
@@ -0,0 +1,3 @@
+/* @flow */
+
+export type Saga = Generator<any, any, any>;
```
rebmullin: 👍

PR #6758, opened 2018-10-26
Use official babel-gettext-extractor

### LINKED ISSUE #6457: Look at switching back to Sentry's babel-gettext-extractor lib
Looks like they've pulled in the changes from the Mozilla fork so we can switch back to it and deprecate the mozilla version ([once the linter switches too](https://github.com/mozilla/addons-linter/issues/2211)).

See https://github.com/getsentry/babel-gettext-extractor

willdurand:
Fixes #6457

---

I manually verified this PR by:

- looking at the fork and upstream libs to make sure all changes in the fork have been added to the upstream/original lib
- extracting/merging the locales for `amo` on `master` and on this PR and compared both results (changes are similar, so that looks good)

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,7 +282,7 @@
     "autoprefixer": "^9.0.0",
     "babel-core": "^6.24.1",
     "babel-eslint": "^10.0.1",
-    "babel-gettext-extractor": "git+https://github.com/muffinresearch/babel-gettext-extractor.git#0d39d3882bc846e7dcb6c9ff6463896c96920ce6",
+    "babel-gettext-extractor": "^3.0.0",
     "babel-jest": "^23.0.0",
     "babel-loader": "^7.0.0",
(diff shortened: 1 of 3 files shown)


rebmullin:
I do..I installed this a few weeks back when I was looking into locales. 

I didn't have an issue a few weeks back when I did the locale PRs so this is new :/

willdurand:
> I do..I installed this a few weeks back when I was looking into locales.
> 
> I didn't have an issue a few weeks back when I did the locale PRs so this is new :/

What's the output of `which msgmerge` on your system?

Do you use `brew` to install your tools? If so, make sure you have `brew install gettext` and you can try `brew link gettext --force` to make sure you get access to the binaries via your `$PATH`.

rebmullin:
I do use `brew`.  However, I was fiddling with my bash_profile file for something else the other day so maybe that was the issue. Anyway tho..your suggestion above fixed my issue 🎉

PR #6761, opened 2018-10-27
Add custom match parameters for category pages

### LINKED ISSUE #2346: Use explicit regex for accepted add-on types in AMO routes
Once we get the new router (which will probably happen in https://github.com/mozilla/addons-frontend/issues/2161) then we can change all of these routes:

````
<Route path=":visibleAddonType/categories/" component={CategoryList} />
````

to a specific regex list of accepted types, like:

````
<Route path=":visibleAddonType(addon|theme|...)/categories/" component={CategoryList} />
````

This will prevent confusing 500 errors when you type in a URL like `/en-US/firefox/users/register/`.

### PR DESCRIPTION
**Pull request by SeanPrashad**
Fixes #2346: Use explicit regex for accepted add-on types in AMO routes

React's `Route` component contains a [`path`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md#path-string--string) prop which lets us specify [custom match parameters](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0#custom-match-parameters). 

By specifying the acceptable addon types, we can remove the `404` code that checks if the specified addon type (`themes` or `extensions`) is allowed.

### DIFF
--- a/src/amo/components/Categories/index.js
+++ b/src/amo/components/Categories/index.js
@@ -107,16 +107,6 @@ export class CategoriesBase extends React.Component<InternalProps> {
     }
     const classNameProp = classnames('Categories', className);
 
-    if (!errorHandler.hasError() && !loading && !categories.length) {
-      return (
-        <Card className={classNameProp}>
-          <p className="Categories-none-loaded-message">
-            {i18n.gettext('No categories found.')}
-          </p>
-        </Card>
-      );
-    }
-
     return (
       <Card className={classNameProp} header={i18n.gettext('Categories')}>
         {errorHandler.renderErrorIfPresent()}
--- a/src/amo/components/Categories/index.js
+++ b/src/amo/components/Categories/index.js
@@ -107,6 +107,16 @@ export class CategoriesBase extends React.Component<InternalProps> {
     }
     const classNameProp = classnames('Categories', className);
 
+    if (!errorHandler.hasError() && !loading && !categories.length) {
+      return (
+        <Card className={classNameProp}>
+          <p className="Categories-none-loaded-message">
+            {i18n.gettext('No categories found.')}
+          </p>
+        </Card>
+      );
+    }
+
     return (
       <Card className={classNameProp} header={i18n.gettext('Categories')}>
         {errorHandler.renderErrorIfPresent()}
--- a/tests/unit/amo/components/TestCategories.js
+++ b/tests/unit/amo/components/TestCategories.js
@@ -255,16 +255,6 @@ describe(__filename, () => {
     ).toHaveProp('children', 'Travel');
   });
 
-  it('renders a no categories found message', () => {
-    const categoriesResponse = { results: [] };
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by SeanPrashad** [comment]
Example of how to use regex for the path: https://reacttraining.com/react-router/web/example/url-params

**Comment by SeanPrashad** [comment]
Here's a gif after the update.. which won't really show anything new on the UI but it has the regex for `extensions` and `themes` in place:

reproducing_regex_bug

File: src/amo/components/Categories/index.js [hunk 1]
```diff
@@ -106,16 +106,6 @@ export class CategoriesBase extends React.Component<InternalProps> {
     }
     const classNameProp = classnames('Categories', className);
 
-    if (!errorHandler.hasError() && !loading && !categories.length) {
```
willdurand: I don't think we should remove this `if` block because it seems unrelated (it handles the case of an empty API response, which might happen).

tests/unit/amo/components/TestCategories.js
```diff
@@ -255,16 +255,6 @@ describe(__filename, () => {
     ).toHaveProp('children', 'Travel');
   });
 
-  it('renders a no categories found message', () => {
```
willdurand: We should probably keep this test case too.

src/amo/components/Categories/index.js
SeanPrashad: Ah ok - I thought this was the block to validate the addon type but I wasn't sure.

SeanPrashad:
Removing those three tests for validating the `visibleAddonType ` breaks three tests:

1. https://github.com/mozilla/addons-frontend/blob/0b8a709bb5a599b2803725b69f2946d492c27952/tests/unit/amo/pages/TestLandingPage.js#L417-L420

2. https://github.com/mozilla/addons-frontend/blob/0b8a709bb5a599b2803725b69f2946d492c27952/tests/unit/amo/pages/TestLandingPage.js#L422-L430

3. https://github.com/mozilla/addons-frontend/blob/0b8a709bb5a599b2803725b69f2946d492c27952/tests/unit/amo/pages/TestCategory.js#L231-L244

All 3 of these tests relate to the old functionality where it was possible to have an add-on type that wasn't `extensions` or `themes`.

My gut instinct is telling me that we should not remove them, but update their functionality to check for a `NotFound` component that's rendered. However, when we override the `visibleAddonType` to be something outside of `extensions` or `themes`, an error is thrown [here](https://github.com/mozilla/addons-frontend/blob/0b8a709bb5a599b2803725b69f2946d492c27952/src/core/utils/index.js#L184-L191) but this test doesn't actually reflect the new functionality because that is not possible anymore.

@willdurand What are your thoughts on completely removing these tests?

willdurand:
@SeanPrashad I would remove the three test cases, they don't make sense anymore.

willdurand:
Thanks @SeanPrashad!

PR #6767, opened 2018-10-29
Remove componentWillReceiveProps() calls

### LINKED ISSUE #6188: Remove componentWillReceiveProps calls (react upgrade)
See: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change

For QA: there should be no issue when navigating between pages (for example, when browsing multiple add-ons, refreshing a page, then navigating to some other add-ons linked on a page, etc.)

---

I filed issues for the following components:

- `Addon`: https://github.com/mozilla/addons-frontend/issues/6770
- `AutoSearchInput`: https://github.com/mozilla/addons-frontend/issues/6771
- `UserProfileEdit`: https://github.com/mozilla/addons-frontend/issues/6769

willdurand:
Fixes #6188

---

~~I don't know if I can remove all of them yet, so let's do this step by step. I start with the "easy" ones.~~ After this patch, only 3 `componentWillReceiveProps()` calls remain but there is one open issue for each occurrence.

### DIFF
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -78,10 +78,10 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
     debounce: defaultDebounce,
   };
 
-  static getDerivedStateFromProps(props: InternalProps) {
+  static getDerivedStateFromProps(props: InternalProps, state: State) {
     const { review } = props;
 
-    if (review) {
+    if (review && state.reviewBody !== review.body) {
       return {
         reviewBody: review.body,
       };
--- a/src/amo/components/AddonReview/index.js
+++ b/src/amo/components/AddonReview/index.js
@@ -78,20 +78,28 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
     debounce: defaultDebounce,
   };
 
+  static getDerivedStateFromProps(props: InternalProps, state: State) {
+    const { review } = props;
+
+    if (review) {
+      return {
+        reviewBody: review.body,
+      };
+    }
+
+    // No state update necessary
+    return null;
+  }
+
   constructor(props: InternalProps) {
     super(props);
+
     this.state = {
       reviewBody: props.review.body,
     };
+
     this.localState = props.createLocalState(`AddonReview:${props.review.id}`);
     this.checkForStoredState();
-  }
-
-  componentWillReceiveProps(nextProps: InternalProps) {
-    const { review } = nextProps;
-    if (review) {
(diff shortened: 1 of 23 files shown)


src/amo/components/CollectionManager/index.js
```diff
@@ -54,29 +54,53 @@ type InternalProps = {|
...
 
-  componentWillReceiveProps(props: InternalProps) {
-    const existingId = this.props.collection && this.props.collection.id;
-    if (props.collection && props.collection.id !== existingId) {
-      // Only reset the form when receiving a collection that the
-      // user is not already editing. This prevents clearing the form
-      // in a few scenarios such as pressing the submit button.
-      this.setState(this.propsToState(props));
-    }
+    this.state = {
```
kumar303: I see a lot of duplication between this and `getDerivedStateFromProps`. Maybe they can share an intermediary function?

src/ui/components/ShowMoreCard/index.js
```diff
@@ -44,43 +44,43 @@ const initialUIState: UIStateType = {
...
+  componentDidUpdate(prevProps: InternalProps) {
+    const { children: prevChildren } = prevProps;
+    const { children, uiState } = this.props;
 
     let html =
+      prevChildren.props &&
+      prevChildren.props.dangerouslySetInnerHTML &&
+      prevChildren.props.dangerouslySetInnerHTML.__html;
+
+    let newHtml =
```
bobsilverberg: Nit: Because we've changed from `componentWillReceiveProps` to `componentDidUpdate`, in the context of this method we're really dealing with `html` and `oldHtml`, rather than `html` and `newHtml`. It might be clearer if you changed the names and switched them around.

src/amo/components/CollectionManager/index.js
bobsilverberg: Snap!

src/amo/components/AddonReview/index.js
```diff
@@ -78,22 +78,30 @@ export class AddonReviewBase extends React.Component<InternalProps, State> {
     debounce: defaultDebounce,
   };
 
+  static getDerivedStateFromProps(props: InternalProps) {
+    const { review } = props;
+
+    if (review) {
```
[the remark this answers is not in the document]
willdurand: That could be interesting indeed.

src/ui/components/ShowMoreCard/index.js
willdurand: ah right, thanks!

PR #6779, opened 2018-10-30
snyk wizard

willdurand:


willdurand:
looks like this is not needed.

PR #6784, opened 2018-10-30
Update permissions code to use information from a version rather than an addon

### LINKED ISSUE #6782: Update permissions code to use information from a version rather than an addon
`current_version` and `platformFiles` are going to be removed from `addon` so permissions code should be updated to work with a specific `version` object and not use `current_version` or `platformFiles` from an `addon` object.

bobsilverberg:
Fixes #6782


bobsilverberg:
Thanks for the review @kumar303. I'm trying to keep them small. The one for `installAddon` is probably going to be quite large.

Issue #6500, opened 2018-10-03
Delete all code related to the old overlay review flow


kumar303:
The inline review flow is landing! https://github.com/mozilla/addons-frontend/issues/6349

Let's delete all code for the old overlay flow.

**For QA**: please make sure that creating / editing ratings and reviews still works.

willdurand:
🏆

ioanarusiczki:
I tested the ratings, reviews and developer replies on the detail page, from the user profile page and the reviews page on AMO dev with FF63(Win10).
Besides the [known issues](https://github.com/mozilla/addons-frontend/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3A%22component%3A+add-on+ratings%22label%3A%22type%3A+bug%22+label%3A%22priority%3A+p3%22+) I filed https://github.com/mozilla/addons-frontend/issues/6827

PR #6789, opened 2018-10-30
Delete code related to the old overlay review flow

kumar303:
Fixes https://github.com/mozilla/addons-frontend/issues/6500

This will be annoying to review, sorry. I guess just try to submit ratings and write a review? I have manually tested everything I can think of.

Notes:

- The RatingManager changes are extra confusing because I removed old-not-old code that was used for the *first* iteration of enabling the inline review flow. In the latest code, much of this logic was moved to AddonReviewCard.
- I deleted some legitimate tests from TestRatingManager because I found duplicate tests already covering the same logic
- I rewrote some tests in TestRatingManager that were legitimate but that were using the old overlay for execution


rebmullin:
also...

kumar303:
Thanks @rebmullin

PR #6795, opened 2018-10-31
Extract HrefLang component

willdurand:
Fixes mozilla/addons#12608

---

This PR creates a new component called `HrefLang` that renders a set of `alternate` HTML links. We have to use `Helmet` in this component and we cannot put `HrefLang` in an existing `Helmet` block because that's how Helmet works...

> Helmet takes plain HTML tags and outputs plain HTML tags. 
> – https://github.com/nfl/react-helmet

### DIFF
--- a/src/amo/components/HeadLinks/index.js
+++ b/src/amo/components/HeadLinks/index.js
@@ -31,12 +31,10 @@ export class HeadLinksBase extends React.PureComponent<InternalProps> {
   render() {
     const { _config, _hrefLangs, clientApp, lang, to } = this.props;
 
-    if (_config.get('unsupportedHrefLangs').includes(lang)) {
-      return null;
-    }
-
     invariant(to.charAt(0) === '/', 'The `to` prop must start with a slash.');
 
+    const includeAlternateLinks =
+      _config.get('unsupportedHrefLangs').includes(lang) === false;
     const hrefLangsMap = _config.get('hrefLangsMap');
 
     return (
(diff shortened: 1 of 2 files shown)


willdurand:
Here is how it is going to be used: https://github.com/mozilla/addons-frontend/pull/6796/commits/548eb4316dc34bf9b23068c82aeb61ecf9056cad

src/amo/components/AddonHead/index.js
```diff
@@ -205,26 +175,29 @@ export class AddonHeadBase extends React.Component<InternalProps> {
...
-        <script type="application/ld+json">
-          {JSON.stringify(getAddonJsonLinkedData({ addon }))}
-        </script>
-      </Helmet>
+      <React.Fragment>
+        <Helmet titleTemplate={null}>
+          <title>{this.getPageTitle()}</title>
+
+          <link
+            rel="canonical"
```
kumar303: Would it make sense to move this into the new component you created? It seems like when rendering alternate links that you'd always want to include the canonical link, too. It also seems like you could safely build this within `HrefLang` with the `to` property.

src/amo/components/AddonHead/index.js
```diff
@@ -6,6 +6,7 @@ import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
+import HrefLang from 'amo/components/HrefLang';
```
kumar303: `HrefLang` doesn't immediately convey what this component does. How about `AltLangLinks`? Or something.

willdurand: `AlternateLinks`?

src/amo/components/AddonHead/index.js
kumar303: `HeadLinks` wfm

src/amo/components/AddonHead/index.js
kumar303: I like `AlternateLinks`

willdurand:
Updated!

src/amo/components/HeadLinks/index.js
```diff
@@ -0,0 +1,85 @@
...
+  static defaultProps = {
+    _config: config,
+    _hrefLangs: hrefLangs,
+  };
+
+  render() {
+    const { _config, _hrefLangs, clientApp, lang, to } = this.props;
+
+    if (_config.get('unsupportedHrefLangs').includes(lang)) {
+      return null;
```
kumar303: Do we still need to render the `rel="canonical"` link for this case? The old code was rendering it.

willdurand: oh 🤦‍♂️

PR #6806, opened 2018-11-01
Update compatibility.js to work with a specific version

### LINKED ISSUE #6629: Update compatibility.js to work with a specific version
Depends on #6634

The functions in [`compatibility.js`](https://github.com/mozilla/addons-frontend/blob/2018.10.11/src/core/utils/compatibility.js) are currently designed to work with an `addon` object and extract the version info from `addon.current_version`. In order to support #2770 we need to be able to pass a specific version into some of these functions.

This issue is about changing `compatibility.js` to allow us to do that.

For QA:

Please test installing an add-on and a theme, both from AMO and disco pane. Also check for regressions in the messaging around incompatible extensions, both in terms of incompatible versions and for when Firefox is not the client.

bobsilverberg:
Fixes #6629

### DIFF
--- a/src/core/utils/compatibility.js
+++ b/src/core/utils/compatibility.js
@@ -1,5 +1,7 @@
+/* @flow */
 /* global window */
 import { oneLine } from 'common-tags';
+import invariant from 'invariant';
 import mozCompare from 'mozilla-version-comparator';
 
 import {
@@ -15,6 +17,9 @@ import {
 } from 'core/constants';
 import { findInstallURL } from 'core/installAddon';
 import log from 'core/logger';
+import type { AddonVersionType } from 'core/reducers/versions';
+import type { UserAgentInfoType } from 'core/reducers/api';
+import type { AddonType } from 'core/types/addons';
 
 // HACK: This is the GUID for the Facebook Container
 // add-on, which has a special-cased download URL supplied.
@@ -23,21 +28,37 @@ export const FACEBOOK_CONTAINER_ADDON_GUID = '@contain-facebook';
 export const FACEBOOK_CONTAINER_DOWNLOAD_URL =
   'https://www.mozilla.org/firefox/facebookcontainer/';
 
-export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
+export type GetCompatibleVersionsParams = {|
+  _log?: typeof log,
+  addon: AddonType,
+  clientApp: string,
+  currentVersion: AddonVersionType,
+|};
+
+export type CompatibleVersionsType = {|
+  supportsClientApp: boolean,
+  maxVersion: string | null,
+  minVersion: string | null,
+|};
+
+export function getCompatibleVersions({
+  _log = log,
+  addon,
+  clientApp,
+  currentVersion,
+}: GetCompatibleVersionsParams = {}): CompatibleVersionsType {
   let maxVersion = null;
   let minVersion = null;
   // Assume the add-on is incompatible until we see explicit support
   // in current_version.compatibility
   let supportsClientApp = false;
 
-  if (addon) {
-    if (
-      addon.current_version &&
-      addon.current_version.compatibility[clientApp]
-    ) {
+  if (currentVersion.compatibility) {
+    const compatInfo = currentVersion.compatibility[clientApp];
+    if (compatInfo) {
       supportsClientApp = true;
-      maxVersion = addon.current_version.compatibility[clientApp].max;
-      minVersion = addon.current_version.compatibility[clientApp].min;
+      maxVersion = compatInfo.max;
+      minVersion = compatInfo.min;
     }
 
     if (!supportsClientApp) {
@@ -48,19 +69,34 @@ export function getCompatibleVersions({ _log = log, addon, clientApp } = {}) {
   return { supportsClientApp, maxVersion, minVersion };
 }
 
+export type IsCompatibleWithUserAgentParams = {|
+  _findInstallURL?: typeof findInstallURL,
+  _log?: typeof log,
+  _window?: typeof window | Object,
+  addon: AddonType,
+  currentVersion: AddonVersionType,
+  maxVersion: string | null,
+  minVersion: string | null,
+  userAgentInfo: UserAgentInfoType,
+|};
+
+export type CompatibleWithUserAgentType = {|
+  compatible: boolean,
+  reason: string | null,
+|};
+
 export function isCompatibleWithUserAgent({
   _findInstallURL = findInstallURL,
+  _log = log,
+  _window = typeof window !== 'undefined' ? window : {},
   addon,
+  currentVersion,
   maxVersion,
   minVersion,
   userAgentInfo,
-  _log = log,
-  _window = typeof window !== 'undefined' ? window : {},
-} = {}) {
+}: IsCompatibleWithUserAgentParams = {}): CompatibleWithUserAgentType {
   // If the userAgent is false there was likely a programming error.
-  if (!userAgentInfo) {
-    throw new Error('userAgentInfo is required');
-  }
+  invariant(userAgentInfo, 'userAgentInfo is required');
 
   const { browser, os } = userAgentInfo;
 
@@ -83,7 +119,7 @@ export function isCompatibleWithUserAgent({
     // confused by this as tofumatt was.
     // See: https://github.com/mozilla/addons-frontend/issues/2074#issuecomment-286983423
     if (maxVersion && mozCompare(browser.version, maxVersion) === 1) {
-      if (addon.current_version.is_strict_compatibility_enabled) {
+      if (currentVersion.isStrictCompatibilityEnabled) {
         return { compatible: false, reason: INCOMPATIBLE_OVER_MAX_VERSION };
       }
 
@@ -105,7 +141,7 @@ export function isCompatibleWithUserAgent({
       }
 
       // `minVersion` is always respected, regardless of
-      // `is_strict_compatibility_enabled`'s value.
+      // `isStrictCompatibilityEnabled`'s value.
       return { compatible: false, reason: INCOMPATIBLE_UNDER_MIN_VERSION };
     }
 
@@ -120,11 +156,12 @@ export function isCompatibleWithUserAgent({
     // we need to make sure it has a matching platform file
     // to work around some bugs.
     // See https://github.com/mozilla/addons-server/issues/6576
+    const { platformFiles } = currentVersion;
     if (
       addon.type === ADDON_TYPE_EXTENSION &&
       !_findInstallURL({
         appendSource: false,
-        platformFiles: addon.platformFiles,
+        platformFiles,
         userAgentInfo,
       })
     ) {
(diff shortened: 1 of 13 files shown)


bobsilverberg:
Some discopane ui tests are failing, and I think it's because of my patch, but not necessarily a problem with my patch. I tested discopane locally with this patch and it seems fine.

@jrbenny35 Perhaps we could discuss this when you have some time?

bobsilverberg:
@diox I think the problem with the ui tests is that this patch relies on https://github.com/mozilla/addons-server/issues/9855. Is there anything that needs to be done to either addons-frontend or addons-server to bring the ui tests up to date with the changes in https://github.com/mozilla/addons-server/issues/9855?

willdurand:
> @diox I think the problem with the ui tests is that this patch relies on [mozilla/addons-server#9855](https://github.com/mozilla/addons-server/issues/9855). Is there anything that needs to be done to either addons-frontend or addons-server to bring the ui tests up to date with the changes in [mozilla/addons-server#9855](https://github.com/mozilla/addons-server/issues/9855)?

@bobsilverberg you should run the ui tests locally to be sure but the config seems to point to the -prod API. The `uitests` env is used for those tests and I don't see any config for the API host, so it should be the `default` one, i.e. -prod.

bobsilverberg:
> you should run the ui tests locally to be sure but the config seems to point to the -prod API. The uitests env is used for those tests and I don't see any config for the API host, so it should be the default one, i.e. -prod.

I tried to run them locally, but couldn't get it working. If the tests use the prod API that would explain the issue, as I don't think https://github.com/mozilla/addons-server/issues/9855 is on prod yet.

bobsilverberg:
@kumar303 you asked about this in IRC. The code is ready to be reviewed. I'm still trying to figure out exactly what is causing the ui test failures, but I suspect it's a mismatch between the API that I expect (which is in dev) and the old API (which is still in prod). I'm not sure what, if anything we can do about that. @diox do you have any suggestions/ideas?

bobsilverberg:
Hmm, there seem to be conflicts now too. I'll address those.

willdurand:
@bobsilverberg here is the stackstrace of the ui tests:

```
redux-saga error: uncaught at fetchDiscoveryAddons
Invariant Violation: id is required
    at invariant (/srv/node/node_modules/invariant/invariant.js:40:15)
    at getVersionById (/srv/code/src/core/reducers/versions.js:130:3)
    at reducer (/srv/code/src/core/reducers/versions.js:273:37)
    at combination (/srv/node/node_modules/redux/lib/combineReducers.js:133:29)
    at /srv/node/node_modules/connected-react-router/lib/reducer.js:50:30
    at dispatch (/srv/node/node_modules/redux/lib/createStore.js:178:22)
    at /srv/node/node_modules/connected-react-router/lib/middleware.js:25:18
    at /srv/node/node_modules/redux-saga/lib/internal/middleware.js:72:22
    at dispatch (/srv/node/node_modules/redux/lib/applyMiddleware.js:45:18)
    at /srv/node/node_modules/redux-saga/lib/internal/utils.js:265:12
    at /srv/node/node_modules/redux-saga/lib/internal/proc.js:500:52
    at exec (/srv/node/node_modules/redux-saga/lib/internal/scheduler.js:25:5)
    at flush (/srv/node/node_modules/redux-saga/lib/internal/scheduler.js:66:5)
    at asap (/srv/node/node_modules/redux-saga/lib/internal/scheduler.js:39:5)
    [...]
```

bobsilverberg:
Thanks @willdurand. That makes sense and is what I'd expect with the API issue.

bobsilverberg:
@willdurand How did you get that stack trace?

willdurand:
> @willdurand How did you get that stack trace?

I ran the ui tests locally, get and cleaned up the Docker logs.

kumar303:
As @willdurand mentioned, the UI tests are not using the local, Docker hosted API. See https://github.com/mozilla/addons-frontend/issues/6830 . The intention was always that they use the local API for integration and predictability.

I guess until that issue is resolved, you should disable the UI tests. It would be worse to ship a broken build because we might miss other failures.

I suggest commenting out the TravisCI job config for now.

kumar303:
D'oh. I was getting confused by the CircleCI tests. The TravisCI tests do not build a local API through Docker. I cancelled https://github.com/mozilla/addons-frontend/issues/6830 because it's not easily actionable.

The best thing I can think of is to disable the UI tests until the change you need lands in prod. Maybe someone has a better idea?

bobsilverberg:
Thanks for the advice @willdurand and @kumar303. I'm not really comfortable with disabling the ui tests for this one, as although I believe my code is not the issue, I wouldn't want to land something that is broken. It can wait until the API changes land on prod. 

Having said that, I think the code could still be reviewed so that it's ready to land as soon as the tests do pass.

bobsilverberg:
@willdurand It looks like https://github.com/mozilla/addons-frontend/issues/6836 fixed the ui tests for this PR. Many thanks!

bobsilverberg:
@willdurand or @kumar303 I guess my hint on IRC wasn't explicit enough. Could someone please review this PR? It's been outstanding for nearly a week and it's blocking two of my other PRs (plus others I haven't submitted yet). Thanks in advance for your help with this.

kumar303:
I spent 100% of my non-meeting time yesterday reviewing code but I had set this one aside until the tests were fixed. I'm reviewing it now. Thanks for your patience.

kumar303:
For future reference, I think it's fine to skip a functional test in order to land a patch more quickly. We shouldn't let a known false positive hold us back.

tests/unit/core/utils/test_compatibility.js
```diff
@@ -319,105 +311,100 @@ describe(__filename, () => {
...
+      const currentVersion = createInternalVersion({
+        ...fakeVersion,
+        compatibility: {},
       });
 
-      expect(maxVersion).toEqual(null);
-      expect(minVersion).toEqual(null);
-    });
-
-    it('returns null if current_version does not exist', () => {
```
kumar303: What happened to this feature? Is it still supported in some way? We need to support add-ons without `current_version` for the case of incomplete add-ons. You can view this type of add-on via AMO (not DiscoPane) as an admin and sometimes as a developer.

tests/unit/disco/components/TestAddon.js
```diff
@@ -411,6 +420,29 @@ describe(__filename, () => {
...
+        _getClientCompatibility,
+        addon,
+      });
+
+      sinon.assert.calledWith(_getClientCompatibility, {
+        addon: createInternalAddon(addon),
+        clientApp,
+        currentVersion: createInternalVersion(fakeVersion),
+        // userAgentInfo stored in state.api only includes browser and os.
+        userAgentInfo: { browser: userAgentInfo.browser, os: userAgentInfo.os },
```
kumar303: Recreating this object from scratch is fragile. I think it would be fine to just compare it to `root.instance().props.userAgentInfo` because we have Flow type safety to help cover that prop.

src/core/utils/compatibility.js
```diff
@@ -142,22 +179,42 @@ export function isCompatibleWithUserAgent({
...
+  _window?: typeof window | {},
+  addon: AddonType,
+  clientApp: string,
+  currentVersion: AddonVersionType,
+  userAgentInfo: UserAgentInfoType,
+|};
+
+export type ClientCompatibilityType = {|
+  compatible: boolean,
+  downloadUrl?: string,
```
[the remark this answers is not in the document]
bobsilverberg: Changing it to `string | null` would require also changing the consumer, which I'd rather not have to do in this patch, so I've changed it to `string | void`. Just wondering, why is `downloadUrl: string | void` considered better than `downloadUrl?: string`?

src/disco/components/Addon/index.js
```diff
@@ -305,17 +310,23 @@ export class AddonBase extends React.Component<InternalProps> {
 
 function mapStateToProps(state: AppState, ownProps: Props) {
   const addon = getAddonByID(state, ownProps.addonId);
+  let currentVersion = null;
 
   let installation = {};
-  if (addon) {
+  if (addon && addon.currentVersionId) {
```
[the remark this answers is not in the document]
bobsilverberg: You're right, unless there's a bug in the reducer we should never get an `addon` without a `currentVersionId`. I'll add the `invariant` as you suggest.

tests/unit/core/utils/test_compatibility.js
```diff
@@ -319,105 +311,100 @@ describe(__filename, () => {
...
 
-      expect(isCompatibleWithUserAgent({ addon, userAgentInfo })).toMatchObject(
-        { compatible: true },
-      );
+      expect(
+        _isCompatibleWithUserAgent({ addon, currentVersion, userAgentInfo }),
+      ).toMatchObject({ compatible: true });
+    });
+
+    it('works with a version', () => {
```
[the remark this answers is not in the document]
bobsilverberg: I have no idea what this is either! It's the exact same test as `should mark Firefox 10 as incompatible with a minVersion of 10.1`. I spent a lot of time on this refactor across multiple branches so I guess at some point I lost my mind and wrote this test. I'll delete it.

tests/unit/core/utils/test_compatibility.js
bobsilverberg: `getCompatibleVersions` no longer accesses `current_version` but rather accepts `currentVersion` as an argument, and it is required to have a value. So this test is no longer meaningful. I've added an `invariant` into `getCompatibleVersions` to enforce it in the case that Flow doesn't catch it.

bobsilverberg:
Thanks for the review @kumar303. I've addressed the issues and pushed a new commit.

src/core/utils/compatibility.js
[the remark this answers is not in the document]
kumar303: ok, that's fine. For the case of a return value object it's not such a big deal but it's a big deal for parameter objects because `downloadUrl?` would allow the caller to forget to pass in the parameter. This could be a mistake. We want Flow to catch those mistakes but still allow the caller to say `someCall({ downloadUrl: undefined })`.

tests/unit/core/utils/test_compatibility.js
kumar303: > `currentVersion` ... is required to have a value. I've added an invariant into `getCompatibleVersions` to enforce it in the case that Flow doesn't catch it.

What I'm saying is that a `null` `currentVersion` object is something we need to support in real life. You can get an add-on with a `null` `current_version` property by looking at the detail page for an incomplete submission either as an admin or a developer.

kumar303:
I'm still concerned about how `null` versions are no longer supported. I added some details to the thread above.

tests/unit/core/utils/test_compatibility.js
bobsilverberg: Ah, I see the issue. I am dealing with that currently by *not calling* `getClientCompatibility` from `Addon` if `currentVersion` is falsey. I do the check in `Addon` as that's the only place where this is called. Flow ensures that we don't try to call `getClientCompatibility` without a `currentVersion`, but I guess my current approach is less encapsulated.  I really should add back in the logic to accept a `null` for `currentVersion` and just return a null in that case. I'll do that.

bobsilverberg:
Ok, I believe I'd addressed the `currentVersion === null` issue. Thanks @kumar303!

src/core/utils/compatibility.js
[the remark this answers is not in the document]
bobsilverberg: Just following up on this because I still don't fully understand (and I commented on some `?`s in https://github.com/mozilla/addons-frontend/pull/6865). Are you saying that we should *never* have optional arguments? That every caller should have to pass either `null` or `undefined` explicitly into every function, even if the argument is irrelevant to the caller?

bobsilverberg:
Thanks for the reviews, @kumar303!

> I tried to test it locally but I can't find any incompatible add-ons, even on hosted dev. Do you know of any? I'm not too worried because QA will help with this.

I wasn't able to confirm the compatibility messages with an incompatible add-on. Like you, I'm going to rely on QA to help with that.

src/core/utils/compatibility.js
```diff
@@ -23,21 +28,36 @@ export const FACEBOOK_CONTAINER_ADDON_GUID = '@contain-facebook';
...
+
+export type CompatibleVersionsType = {|
+  supportsClientApp: boolean,
+  maxVersion: string | null,
+  minVersion: string | null,
+|};
+
+export function getCompatibleVersions({
+  _log = log,
+  addon,
```
willdurand: Why did you keep `addon` here? It is not used anymore, except in a log call.

bobsilverberg: That's the reason I kept it. I assumed we wanted to log the `addon` in that `log` call. I suppose we could use `currentVersion` instead, but that contains less immediately obvious information. 

Do you think we should change this to remove `addon` as an argument?

willdurand: no, I wanted to double-check. It makes sense to keep it in order to get a better log message 👍

PR #6809, opened 2018-11-01
I6661-privacy guide initial setup

rebmullin:
fixes mozilla/addons#12550 (initial setup for the privacy page)

REF: 
design:
https://mozilla.invisionapp.com/d/main/#/console/15640858/324863887/preview

content:
https://docs.google.com/document/d/1u1tbRo_ivfEAh_er2XKPciuYLARiNa2N5qXPGC0jfcY/edit?ts=5bbcb799

### DIFF
--- a/tests/unit/amo/pages/TestGuides.js
+++ b/tests/unit/amo/pages/TestGuides.js
@@ -99,14 +99,14 @@ describe(__filename, () => {
     );
   });
 
-  it(`renders an HTML title`, () => {
+  it('renders an HTML title', () => {
     const content = getContent('privacy', fakeI18n());
     const root = render({ content });
 
     expect(root.find('title')).toHaveText(content.title);
   });
 
-  it(`renders a HeadLinks component`, () => {
+  it('renders a HeadLinks component', () => {
     const root = render();
 
     expect(root.find(HeadLinks)).toHaveLength(1);


src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,251 @@
...
+          });
+
+          return (
+            <div className="Guides-section" key={addon && addon.url}>
+              <h2 className="Guides-section-title">{sections[key].header}</h2>
+              <p className="Guides-section-description">
+                {sections[key].description}
+              </p>
+
+              <GuidesAddonCard
```
willdurand: This PR is ultra big. How about splitting it? I think you could easily omit the `GuidesAddonCard` component and ship the rest of the PR, WDYT? Maybe we could also have only one `privacy` section too?

rebmullin: yep, I can definitely do that. thanks for the suggestion : )

regarding the `privacy` section suggestion - do you mean combining the `getGuids` and `getContent` function?

willdurand: no sorry, I meant just having one entry in the `sections` so that it reduces the diff even more.

src/amo/components/Routes/index.js
```diff
@@ -41,6 +42,15 @@ const Routes = ({ _config = config }: Props = {}) => (
     <Route exact path="/:lang/:application/" component={Home} />
 
     <Route exact path="/:lang/:application/addon/:slug/" component={Addon} />
+
+    {_config.get('enableFeatureHomeHeroGuides') && (
+      <Route
+        exact
+        path="/:lang/:application/guides/:guidesSlug"
```
willdurand: I think you can use just `slug` here.

willdurand: And we should add a trailing slash too.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+|};
+
+export const getGuids = (guidesSlug: string): Array<string> | null => {
+  switch (guidesSlug) {
+    case 'privacy':
+      return ['{446900e4-71c2-419f-a6a7-df9c091e268b}'];
+    // The following will be addressed in another issue.
+    // See https://github.com/mozilla/addons-frontend/issues/6744.
+    case 'organize-tabs-and-bookmarks':
+    case 'enhance-your-media-experience':
```
willdurand: I suppose you can delete this code for now (the two `case` above). There is a issue for that anyway.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+              'Explore more %(linkStart)spassword manager%(linkEnd)s staff picks.',
+            ),
+            exploreUrl: '/collections/mozilla/password-managers/',
+          },
+        ],
+      };
+    // The following will be addressed in another issue.
+    // See https://github.com/mozilla/addons-frontend/issues/6744.
+    case 'organize-tabs-and-bookmarks':
+    case 'enhance-your-media-experience':
```
willdurand: Same here.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+    default:
+      return null;
+  }
+};
+
+export const getContent = (guidesSlug: string, i18n: I18nType): Object => {
```
willdurand: You should create a flow type for each content entry, instead of `Object`.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+            exploreUrl: '/collections/mozilla/password-managers/',
+          },
+        ],
+      };
+    // The following will be addressed in another issue.
+    // See https://github.com/mozilla/addons-frontend/issues/6744.
+    case 'organize-tabs-and-bookmarks':
+    case 'enhance-your-media-experience':
+    default:
+      return {};
```
willdurand: I don't think we should have that as a return value. Maybe `null`?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+export class GuidesBase extends React.Component<InternalProps> {
+  constructor(props: InternalProps) {
+    super(props);
+
+    const { guidesSlug, guids } = this.props;
+
+    this.props.dispatch(
+      fetchGuidesAddons({
+        guids,
+        errorHandlerId: guidesSlug,
```
willdurand: I think you should use `withFixedErrorHandler()` (HOC), which will inject a `errorHandler` in this component (of type `ErrorHandlerType`).

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+      }),
+    );
+  }
+
+  getGuidesSectionsHtml = (
```
willdurand: Technically, it is not HTML.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+    sections: Array<Object>,
```
willdurand: You need a type here.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+  ): React.ChildrenArray<React.Node> => {
+    const { addons, clientApp, i18n, lang } = this.props;
+
+    return addons.length
+      ? addons.map((addon, key) => {
+          // TODO: look into having these links use the Router (vs 'a' tag).
```
willdurand: Yep, we should definitely have that.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+    );
+  }
+
+  getGuidesSectionsHtml = (
+    sections: Array<Object>,
+  ): React.ChildrenArray<React.Node> => {
+    const { addons, clientApp, i18n, lang } = this.props;
+
+    return addons.length
+      ? addons.map((addon, key) => {
```
willdurand: What is this `key`? I find it hard to understand how `addons` and `sections` are related.

willdurand: Is it the array index?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+          const exploreMore = i18n.sprintf(sections[key].exploreMore, {
+            linkStart: `<a href="/${lang}/${clientApp}${
+              sections[key].exploreUrl
+            }">`,
+            linkEnd: '</a>',
+          });
+
+          return (
+            <div className="Guides-section" key={addon && addon.url}>
+              <h2 className="Guides-section-title">{sections[key].header}</h2>
```
willdurand: Would it make sense to use a map for `sections`? We could use add-on's `guid`s as keys.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+          <p className="Guides-header-intro">{introText}</p>
+        </div>
+        {this.getGuidesSectionsHtml(sections)}
+      </div>
+    );
+  };
+
+  render() {
+    const { i18n } = this.props;
+    const guidesPage = this.getGuidesPage();
```
willdurand: Why not moving the logic from `getGuidesPage()` directly here?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+    if (!guidesPage) {
+      return <NotFound />;
+    }
+
+    const { match } = this.props;
+    const { guidesSlug } = match.params;
+    const { title } = getContent(guidesSlug, i18n);
+    const pageTitle = i18n.sprintf('%(pageTitle)s', {
+      pageTitle: title,
+    });
```
willdurand: There is nothing to translate here, so you can use `title` directly in `<title>{title}</title>`.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,214 @@
...
+}
+
+export const mapStateToProps = (state: AppState, ownProps: Props) => {
+  const { match } = ownProps;
+  const { guidesSlug } = match.params;
+
+  const addons = [];
+  const guids = getGuids(guidesSlug) || [];
+
+  guids.forEach((guid) => {
```
willdurand: I think this should be done in the `render()` method instead, so that you can map each guid to a "section". That would avoid the implicit contract that is: the order of the array of add-ons.

src/amo/pages/Guides/index.js
rebmullin: Each addon that is displayed, is showing its own addon data along with some additional data from the sections array. Since the indexes match up, I am using this to match up the addon with it's corresponding sections data. I am not sure if it will help but I can rename the key to index..?

src/amo/pages/Guides/index.js
rebmullin: yeh I like that..I can map through sections instead.

src/amo/pages/Guides/index.js
rebmullin: but I have to have state to use this reducer: `getAddonByGUID`. Is it okay to pass all of state as a prop?

or let me know if I miss something..

src/amo/pages/Guides/index.js
rebmullin: Altho this might come up again when I need to reference the affiliated addon...

rebmullin: maybe we can chat about this on weds?

src/amo/pages/Guides/index.js
willdurand: oh, I see now. Maybe you can build the map here then, instead of having `guids` as an array, let's have a `{[guid: string]: AddonType }`

src/amo/pages/Guides/index.js
willdurand: What are affiliated add-ons?

rebmullin: each section item corresponds with its own addon

willdurand: ah yep, so each section can also be mapped by add-ons `guid`, isnt it?

src/amo/pages/Guides/index.js
rebmullin: Another thought I had, is I can just add the addon guid to its respective section and then pass the guid to the `GuidesAddonCard`:

```javascript
 <GuidesAddonCard
     addonGuid={section.addonGuid}
  />
```

Then I can use the `getAddonByGUID` reducer from `GuidesAddonCard`. I am going to try that. Hopefully you will agree : )

src/amo/pages/Guides/index.js
rebmullin: yeh as I mentioned below, I will add a guid to each section. I think this way is definitely clearer :)

rebmullin:
@willdurand, thanks so much for the review. I've made some changes..can you please take another look? :)

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+type SectionsType = {|
+  addonGuid: string,
+  header: string,
+  description: string,
+  addonCustomText: string,
+  exploreMore: string,
+  exploreUrl: string,
+|};
+
+type GetContentType = {|
```
willdurand: Or maybe `GuideType`?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+export const getContent = (
+  slug: string,
+  i18n: I18nType,
+): GetContentType | null => {
+  switch (slug) {
+    case 'privacy':
+      return {
+        title: i18n.gettext('Stay Safe Online'),
+        introText: i18n.gettext(
+          'The web is a wonderful but wild place. Your personal data can be used without your consent, your activities spied upon, and your passwords stolen. Fortunately, extensions can help fortify your online privacy and security.',
```
willdurand: You can use backticks to wrap the text at ~80 chars.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+export class GuidesBase extends React.Component<InternalProps> {
+  constructor(props: InternalProps) {
+    super(props);
+
+    const { errorHandler, i18n, match } = this.props;
+    const { slug } = match.params;
+    const content = getContent(slug, i18n);
+
+    if (!content) {
+      return null;
```
willdurand: `return;` should be enough, no?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+    );
+  }
+
+  getGuidesSections = (
+    sections: Array<SectionsType>,
+  ): React.ChildrenArray<React.Node> => {
+    const { clientApp, i18n, lang } = this.props;
+
+    return sections.map((section) => {
+      const exploreMore = i18n.sprintf(section.exploreMore, {
```
willdurand: `exploreMoreLink`?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+
+      return (
+        <div className="Guides-section" key={section.exploreUrl}>
+          <h2 className="Guides-section-title">{section.header}</h2>
+          <p className="Guides-section-description">{section.description}</p>
+
+          {/* eslint-disable react/no-danger */}
+          <div
+            className="Guides-section-explore-more"
+            dangerouslySetInnerHTML={sanitizeHTML(exploreMore, ['a'])}
```
willdurand: You can probably use `// eslint-disable react/no-danger` directly above `dangerouslySetInnerHTML` to avoid the two comments.

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,187 @@
...
+      return <NotFound />;
+    }
+
+    const { introText, icon, title, sections } = content;
+
+    return (
+      <div className="Guides-page">
+        <Helmet>
+          <title>{title}</title>
+        </Helmet>
```
willdurand: We will probably want to add the `HeadLinks` component (may need a rebase). All you have to do is adding `<HeadLinks />` below the `</Helmet>` block (not inside, outside, below).

willdurand: Write a test case for that.

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,96 @@
...
+  shallowUntilTarget,
+} from 'tests/unit/helpers';
+import Guides, { extractId, GuidesBase, getContent } from 'amo/pages/Guides';
+
+describe(__filename, () => {
+  let content;
+  let store;
+  let slug;
+  let i18n;
+  let guids;
```
willdurand: I am not a fan of global variables in test files. This is super tricky to refactor.

willdurand: Can you write a `getProps` function instead? That function can have all the defaults set in `beforeEach`.

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  it('renders a Guides Page', () => {
+    const root = render();
+
+    expect(root.find('title')).toHaveText(content.title);
+    expect(root.find('.Guides')).toHaveLength(1);
+    expect(root.find('.Guides-header-icon')).toHaveLength(1);
+    expect(root.find('.Guides-header-page-title')).toHaveLength(1);
+    expect(root.find('.Guides-header-intro')).toHaveLength(1);
```
willdurand: You should probably get some content (`GuideType`) first, then pass that content to the `render()` function along with the slug, and then assert for content strings in each section here. We know that it will be rendered given that the parent component (`.Guides`) is rendered.

tests/unit/amo/pages/TestGuides.js
willdurand: There is no need to keep the `beforeEach()`

rebmullin:
@willdurand, thanks again for the review : ) I've updated again...can you please take another look when you can? Hopefully it's close now 🤞

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,193 @@
...
+        <HeadLinks />
```
willdurand: You can add blank lines before and after to identify the different logic blocks.

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,108 @@
...
+} from 'tests/unit/helpers';
+import Guides, { extractId, GuidesBase, getContent } from 'amo/pages/Guides';
+
+describe(__filename, () => {
+  const getProps = ({
+    store = dispatchClientMetadata().store,
+    i18n = fakeI18n(),
+    dispatch = store.dispatch,
+    slug = 'privacy',
+    content = getContent('privacy', i18n),
```
willdurand: you might even be able to use `content = getContent(slug, i18n)` here

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,108 @@
...
+      dispatch,
+      i18n,
+      match,
+      slug,
+      store,
+      ...customProps,
+    };
+  };
+
+  const render = ({ ...customProps } = {}) => {
```
willdurand:
```suggestion
  const render = (customProps = {}) => {
```

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,108 @@
...
+    expect(root.find(HeadLinks)).toHaveLength(1);
+
+    expect(root.find('.Guides')).toHaveLength(1);
+    expect(root.find('.Guides-header-icon')).toHaveLength(1);
+    expect(root.find('.Guides-header-page-title')).toHaveLength(1);
+    expect(root.find('.Guides-header-intro')).toHaveLength(1);
+
+    expect(root.find('.Guides-section')).toHaveLength(1);
+    expect(root.find('.Guides-section-title').text()).toContain(
+      'Create and manage strong passwords',
```
willdurand: You may want to use `content.title`, no?

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,108 @@
...
+    );
+    expect(root.find('.Guides-section-description').text()).toContain(
+      'Password managers can help you',
```
willdurand: Could we use `content.description`?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,193 @@
...
+  constructor(props: InternalProps) {
+    super(props);
+
+    const { errorHandler, i18n, match } = this.props;
+    const { slug } = match.params;
+    const content = getContent(slug, i18n);
+
+    if (!content) {
+      return;
+    }
```
willdurand: Is this covered by a test case?

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,193 @@
...
+    if (!content) {
+      return <NotFound />;
+    }
+
+    const { introText, icon, title, sections } = content;
+
+    return (
+      <div className="Guides-page">
+        <Helmet>
+          <title>{title}</title>
```
willdurand: We need a test case for that. Search for "renders an HTML title" in the test files, you'll find examples.

src/amo/pages/Guides/index.js
willdurand: This also needs to be covered by a test case. Simply make sure it is present with `toHaveLength(1)`.

```js
it(`renders a HeadLinks component`, () => {});
```

src/amo/pages/Guides/index.js
```diff
@@ -0,0 +1,193 @@
...
+  ): React.ChildrenArray<React.Node> => {
+    const { clientApp, i18n, lang } = this.props;
+
+    return sections.map((section) => {
+      const exploreMoreLink = i18n.sprintf(section.exploreMore, {
+        linkStart: `<a class="Guides-section-explore-more-link" href="/${lang}/${clientApp}${
+          section.exploreUrl
+        }">`,
+        linkEnd: '</a>',
+      });
```
willdurand: Do you have an issue opened for that? (I think I saw something a while ago) Make sure there is an issue, no need to reference it here though.

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,108 @@
...
+    );
+    expect(root.find('.Guides-section-explore-more')).toHaveHTML(
+      oneLine`<div class="Guides-section-explore-more">Explore more
+      <a href="/en-US/android/collections/mozilla/password-managers/"
+      class="Guides-section-explore-more-link">password manager</a> staff picks.</div>`,
+    );
```
willdurand: Note: you will likely have to update those assertions when you'll add more sections. Maybe you want to use `.find()` and then `.at(0)`.

src/amo/pages/Guides/index.js
rebmullin: yep, I added this expect but not as it's own separate test. I can extract that out tho..

tests/unit/amo/pages/TestGuides.js
rebmullin: I have to dig into the sections array for this which I can do..

rebmullin: or did you mean the page title or the section title? or both? :)

for example the getContent will return this whole object here:
```javasscript
case 'privacy':
      return {
        title: i18n.gettext('Stay Safe Online'),
        introText: i18n.gettext(
          `The web is a wonderful but wild place. Your personal data can be used
           without your consent, your activities spied upon, and your passwords
           stolen. Fortunately, extensions can help fortify your online privacy
           and security.`,
        ),
        icon: 'stop-hand',
        sections: [
          // Bitwarden free password manager
          {
            addonGuid: '{446900e4-71c2-419f-a6a7-df9c091e268b}',
            header: i18n.gettext('Create and manage strong passwords'),
            description: i18n.gettext(
              `Password managers can help you create secure passwords, store your
               passwords (safely) in one place, and give you easy access to your
               login credentials wherever you are.`,
            ),
            addonCustomText: i18n.gettext(
              `Fully encrypted password protection. Store your data securely and
               access logins across devices.`,
            ),
            exploreMore: i18n.gettext(
              'Explore more %(linkStart)spassword manager%(linkEnd)s staff picks.',
            ),
            exploreUrl: '/collections/mozilla/password-managers/',
          },
        ],
      };
```

src/amo/pages/Guides/index.js
rebmullin: yep i did this but I've added as part of the "renders a Guide Page" test case..I can extract this out too tho..

src/amo/pages/Guides/index.js
rebmullin: yep it WAS there but it's missing now:/ I will add this back, thanks!

tests/unit/amo/pages/TestGuides.js
rebmullin: I just need to test 1 item tho, right? (`.at(0)` for example), not each section?

src/amo/pages/Guides/index.js
willdurand: yep, that'd be nice

tests/unit/amo/pages/TestGuides.js
willdurand: yep, one section is probably fine

src/amo/pages/Guides/index.js
rebmullin: yeh there IS an issue so I think I'll reference that here:)

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,140 @@
...
+    const sectionExploreLink = root.find('.Guides-section-explore-more').at(0);
+
+    expect(sectionExploreLink).toHaveHTML(
+      oneLine`<div class="Guides-section-explore-more">Explore more
+      <a href="/en-US/android/collections/mozilla/password-managers/"
+      class="Guides-section-explore-more-link">password manager</a> staff picks.</div>`,
+    );
+  });
+
+  it(`renders an HTML title`, () => {
```
willdurand: You probably don't need the backticks here

tests/unit/amo/pages/TestGuides.js
```diff
@@ -0,0 +1,140 @@
...
+    const content = getContent('privacy', fakeI18n());
+    const root = render({ content });
+
+    expect(root.find('title')).toHaveText(content.title);
+  });
+
+  it(`renders a HeadLinks component`, () => {
```
willdurand: same here

PR #6812, opened 2018-11-02
Always load the latest saved review, regardless of version

kumar303:
This fixes https://github.com/mozilla/addons-frontend/issues/4529 by removing the `versionId` parameter when looking for saved reviews. I confirmed that when omitting this parameter, the API returns the latest saved review for the add-on, regardless of version. I also verified that no errors are thrown if the user had reviewed an add-on multiple times at different versions.

As part of this patch, I moved the old review fetching logic to a saga.

### DIFF
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -296,7 +296,6 @@ export function setGroupedRatings({
 
 type FetchLatestUserReviewParams = {|
   addonId: number,
-  addonSlug: string,
   errorHandlerId: string,
   userId: number,
 |};
@@ -308,12 +307,10 @@ export type FetchLatestUserReviewAction = {|
 
 export function fetchLatestUserReview({
   addonId,
-  addonSlug,
   errorHandlerId,
   userId,
 }: FetchLatestUserReviewParams): FetchLatestUserReviewAction {
   invariant(addonId, 'addonId is required');
-  invariant(addonSlug, 'addonSlug is required');
   invariant(errorHandlerId, 'errorHandlerId is required');
   invariant(userId, 'userId is required');
 
@@ -321,7 +318,6 @@ export function fetchLatestUserReview({
     type: FETCH_LATEST_USER_REVIEW,
     payload: {
       addonId,
-      addonSlug,
       errorHandlerId,
       userId,
     },
@@ -613,7 +609,6 @@ export const setReviewWasFlagged = ({
 
 type SetLatestReviewParams = {|
   addonId: number,
-  addonSlug: string,
   review: ExternalReviewType | null,
   userId: number,
 |};
@@ -625,18 +620,16 @@ export type SetLatestReviewAction = {|
 
 export const setLatestReview = ({
   addonId,
-  addonSlug,
   review,
   userId,
 }: SetLatestReviewParams): SetLatestReviewAction => {
   invariant(addonId, 'addonId is required');
-  invariant(addonSlug, 'addonSlug is required');
   invariant(review !== undefined, 'review is required');
   invariant(userId, 'userId is required');
 
   return {
     type: SET_LATEST_REVIEW,
-    payload: { addonId, addonSlug, review, userId },
+    payload: { addonId, review, userId },
   };
 };
 
--- a/src/amo/actions/reviews.js
+++ b/src/amo/actions/reviews.js
@@ -20,6 +20,8 @@ export const FETCH_REVIEW: 'FETCH_REVIEW' = 'FETCH_REVIEW';
 export const FETCH_REVIEW_PERMISSIONS: 'FETCH_REVIEW_PERMISSIONS' =
   'FETCH_REVIEW_PERMISSIONS';
 export const FETCH_REVIEWS: 'FETCH_REVIEWS' = 'FETCH_REVIEWS';
+export const FETCH_LATEST_USER_REVIEW: 'FETCH_LATEST_USER_REVIEW' =
+  'FETCH_LATEST_USER_REVIEW';
 export const FETCH_USER_REVIEWS: 'FETCH_USER_REVIEWS' = 'FETCH_USER_REVIEWS';
 export const FLASH_REVIEW_MESSAGE: 'FLASH_REVIEW_MESSAGE' =
   'FLASH_REVIEW_MESSAGE';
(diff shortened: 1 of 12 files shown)


kumar303:
> Why are we even bothering with accepting and passing a versionId when getting the latest review for an add-on?

I thought that maybe we'd need it one day but I can't think of any realistic scenarios so, yeah, I'll remove it.

kumar303:
> It seems to me like we could just remove a lot of the code that looks at versionId

Actually, yeah, I forgot to do this!

kumar303:
I simplified all the old code which had previously supported user reviews across versions.

src/amo/components/RatingManager/index.js
```diff
@@ -82,53 +64,23 @@ type InternalProps = {|
...
-    };
-
-    if (userReview) {
-      log.debug(`Editing reviewId ${userReview.id}`);
-      params.reviewId = userReview.id;
-    } else {
-      log.debug(oneLine`Submitting a new review for
-        versionId ${params.versionId || '[empty]'}`);
-    }
+    const { addon, dispatch, errorHandler, userReview, version } = this.props;
```
bobsilverberg: It doesn't show up in this diff, but I think you can also remove `version` from the `render` method of this component.

src/amo/sagas/reviews.js
```diff
@@ -376,6 +379,53 @@ function* deleteAddonReview({
   }
 }
 
+function* fetchLatestUserReview({
+  payload: { addonId, addonSlug, errorHandlerId, userId },
+}: FetchLatestUserReviewAction): Generator<any, any, any> {
```
bobsilverberg: I saw @willdurand use the [`Saga` type](https://github.com/mozilla/addons-frontend/blob/master/src/core/types/sagas.js#L3) for this in a recent PR. Should we be doing that?

src/amo/api/reviews.js
```diff
@@ -186,37 +186,36 @@ export function getReviews({
   });
 }
 
-export type GetLatestReviewParams = {|
+export type GetLatestUserReviewParams = {|
   addon: number,
```
bobsilverberg: I know this is old code, but it is a bit counter-intuitive. Should we be using `addonId` and `userId` here instead of `addon` and `user`?

src/amo/sagas/reviews.js
kumar303: How is that any more helpful? Also, I just learned that `Generator` is the equivalent of `Generator<any, any, any>` so maybe we should use that? In any case, our approach is that we do not use type safety for covering what a generator yields and receives.

src/amo/api/reviews.js
kumar303: Since it's an API function that maps directly to [this `addon` parameter](https://addons-server.readthedocs.io/en/latest/topics/api/ratings.html#list-ratings), the value can be either an `addonId` or an `addonSlug`.

src/amo/sagas/reviews.js
willdurand: It is a meaningful alias (and it is shorter).

src/amo/components/RatingManager/index.js
kumar303: The `render` function calls `invariant` to say that `version` is required. I think that's a good place to do so because you'll see it early on in development. FYI, `Addon` doesn't have Flow coverage and it still uses `RatingManager` so this invariant call is important.

src/amo/actions/reviews.js
```diff
@@ -292,12 +294,40 @@ export function setGroupedRatings({
   };
 }
 
-type FetchUserReviewsParams = {|
+type FetchLatestUserReviewParams = {|
+  addonId: number,
+  addonSlug: string,
```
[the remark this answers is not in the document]
kumar303: Good catch, thanks. I think that was a relic from when the reducer had to reset some cache based on slug or something.

src/amo/actions/reviews.js
```diff
@@ -311,7 +341,11 @@ export function fetchUserReviews({
   errorHandlerId,
   userId,
   page = '1',
-}: FetchUserReviewsParams): FetchUserReviewsAction {
+}: {|
```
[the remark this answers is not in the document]
kumar303: > Why did you decide to use an explicit type for FetchLatestUserReviewParams 

because it's shared by the action payload

> ...but do away with the explicit type here and just create an inline type definition?

because this one is not shared with anything. I find inline object types preferable; you can read the code top to bottom without having to look up a reference to the type.

src/amo/sagas/reviews.js
kumar303: It's not shorter when you count the import statement 😁 but that's fine, I can use it. I'll add a comment to the type definition to explain that we don't rely on type safety for covering what a saga yields and receives. Maybe that will help someone be less confused when they first see it.

src/amo/api/reviews.js
kumar303: Whoops, this should have been `addon: number | string`

kumar303:
Thanks for the review. I think I addressed all issues but let me know if I missed anything.

src/amo/components/RatingManager/index.js
bobsilverberg: Why not just put the `invariant` in `onSelectRating`, where `version` is actually needed? I understand that having it in `render` is like an early warning system, but it still feels odd to me. At some point we won't even need `version` at all (when the back end is changed to just not care about versions when it comes to reviews). I guess it's fine to keep it in `render`, but I don't love it.

src/amo/api/reviews.js
bobsilverberg: That makes more sense.

src/amo/components/RatingManager/index.js
kumar303: I considered that but putting `invariant` in `onSelectRating` means that the error will only happen when a user tries to create a rating. That would be a crappy error to experience in real life. I know we do this in other places but that's only for `nullable` properties where it wouldn't make sense to throw in `render`. But, yeah, once we can change the backend for creating new ratings then the `version` requirement will go away.

kumar303: > once we can change the backend for creating new ratings...

I'm actually not sure we will ever be able to do that. I think it's important to know which version a user was referring to when they first rated an add-on.

PR #6819, opened 2018-11-04
Fixes 6735 | Blank space in URLs redirect

### LINKED ISSUE #6735: Blank space in URLs can cause the requested page to get stuck in a loading state
STR:

1. Load the following page - https://addons-dev.allizom.org/en-US/firefox/addon/stratiform%20/
2. Observe the layout

Actual result:
The page content is not displayed and it looks stuck in a continuous loading state. There are a lot of CSP reports in the console.

Expected result:
The page requested should return a 404, or should ignore the blank space, as the API does

Notes:
- if you call the same page from the API, its details are retrieved. The API seems to ignore the blank space: see https://addons-dev.allizom.org/api/v4/addons/addon/stratiform%20/
- I have also tried with a different page - https://addons-dev.allizom.org/en-US/firefox/addon/stratiform/reviews/563890%20/ and in this case, a 404 will be triggered, but from a legacy page. Again, the API will ignore the blank space and will return the review details: see https://addons-dev.allizom.org/api/v4/ratings/rating//563890%20/
- reproduced on all AMO servers with F62, WIn10x64

image

tsl143:
Fixes #6735 
Trims the trailing spaces of `slug` in `match.params`. This triggers 301 redirect to and hence removes loading state.

- [x] This PR relates to an existing open issue and there are no existing PRs open for the same issue.
- [x] Add `Fixes #ISSUENUM` at the top of your PR.
- [x] Add a description of the the changes introduced in this PR.
- [x] The change has been successfully run locally.
- [x] Add tests to cover the changes added in this PR.

### DIFF
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -611,7 +611,8 @@ export class AddonBase extends React.Component {
 }
 
 export function mapStateToProps(state, ownProps) {
-  const { slug } = ownProps.match.params;
+  let { slug = '' } = ownProps.match.params;
+  slug = typeof slug === 'string' ? slug.trim() : slug;
   let addon = getAddonBySlug(state, slug);
 
   // It is possible to load an add-on by its ID but in the routing parameters,
--- a/src/amo/pages/Addon/index.js
+++ b/src/amo/pages/Addon/index.js
@@ -611,7 +611,7 @@ export class AddonBase extends React.Component {
 }
 
 export function mapStateToProps(state, ownProps) {
-  let { slug = '' } = ownProps.match.params;
+  let { slug } = ownProps.match.params;
   slug = typeof slug === 'string' ? slug.trim() : slug;
   let addon = getAddonBySlug(state, slug);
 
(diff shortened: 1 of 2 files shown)


src/amo/pages/Addon/index.js
```diff
@@ -611,7 +611,8 @@ export class AddonBase extends React.Component {
 }
 
 export function mapStateToProps(state, ownProps) {
-  const { slug } = ownProps.match.params;
+  let { slug = '' } = ownProps.match.params;
```
bobsilverberg: I'm not sure that you need to default `slug` to `''` here. You're checking below that `slug` is a string before trying to call `trim()` on it, so I don't think you need to default it to anything. 

It's interesting that the [Flow definition of `getAddonBySlug`](https://github.com/mozilla/addons-frontend/blob/2018.11.08/src/core/reducers/addons.js#L319-L322) states that `slug` must be a string, but it is also checking internally for `typeof string`, so that suggests that `slug` could be something other than a string.

tsl143:
1. I put slug=' ' just out of habit, I do this before any variable on which
string operations are done, in this case trim. I see no harm in that, but
can totally be removed if we are very sure.

2. There are numeric slugs! I don't remember exactly(not on computer rn)
but I think for some legacy slugs. You can find it in test cases.

PR #6825, opened 2018-11-05
Upgrade react-cookie

### LINKED ISSUE #6803: Upgrade react-cookie for reliable cookie state
**For QA**: please check for regressions in any feature that relies on a cookie:
* log in / log out
* A/B test for green/blue buttons https://github.com/mozilla/addons-frontend/issues/6144
* The *view classic desktop site* link in the footer

<hr/>

We need to upgrade `react-cookie` for better [SSR support](https://github.com/reactivestack/cookies/issues/79#issuecomment-298998732). See https://github.com/mozilla/addons-frontend/issues/6477#issuecomment-434843845 for details on how the older library is likely creating bugs.

willdurand:
Fixes #6803


tests/unit/amo/components/TestApp.js
```diff
@@ -22,12 +22,18 @@ import {
...
-  fakeCookie,
+  fakeCookies,
   fakeI18n,
   shallowUntilTarget,
   userAuthToken,
 } from 'tests/unit/helpers';
 
+// Skip `withCookies` HOC since Enzyme does not support the React Context API.
+// See: https://github.com/mozilla/addons-frontend/issues/6839
+jest.mock('react-cookie', () => ({
```
kumar303: This seems fine as a temporary hack but why not define it once in [`setup.js`](https://github.com/mozilla/addons-frontend/blob/master/tests/setup.js)? From a quick test on my end, it should work.

willdurand: Actually, that breaks lots of tests. It is better to mock when needed, not everywhere..

kumar303: Can you just pull this into a helper like `setUpMockReactCookie()` ? This will make it easier to find and delete all the code when we remove it. I don't know what's up with the `jest` magic global so if that doesn't work then it's fine.

src/core/server/base.js
```diff
@@ -208,6 +208,9 @@ function baseServer(
     app.use(middleware.serveAssetsLocally());
   }
 
+  // Cookies.
```
kumar303: I don't think this comment adds anything. The code already explains it.

However, it might help to add a comment saying how this will patch `universalCookies` on to `request` since that part is a bit magical.

tests/unit/core/server/test_server.js
```diff
@@ -71,11 +72,14 @@ export class ServerTestHelper {
     global.webpackIsomorphicTools = {
       assets: () => fakeAssets,
     };
+    // We don't want the `document` object on the server.
+    delete global.document;
```
kumar303: Good idea. Maybe delete `window`, too?

willdurand: I deleted `document` mainly because `universal-cookie` uses it to determine if it is in server or client mode.

tests/unit/amo/components/TestApp.js
willdurand: Looks like it does not work well...

PR #6832, opened 2018-11-05
Use -dev API for ui tests

### LINKED ISSUE #6836: Use -dev API for ui tests
We should use the `-dev` API instead of `-prod` in the `uitests` run on Travis CI to avoid issues like the one described in https://github.com/mozilla/addons-frontend/pull/6806#issuecomment-435511037.

willdurand:
Fixes #6836

---

This PR changes the `uitests` config to use the -dev API instead of -prod.

It fixes the test case that installs an extension, ~~but not (yet) the theme install one.~~ Fixed by installing a LWT and not a static theme.


kumar303:
Thanks!

PR #6840, opened 2018-11-06
Remove `componentWillReceiveProps()` in `AutoSearchInput`

### LINKED ISSUE #6771: Remove `componentWillReceiveProps()` in `AutoSearchInput`
See also: #6188 

---

The following patch works but one test case is still failing (see after the patch):

```diff
diff --git a/src/amo/components/AutoSearchInput/index.js b/src/amo/components/AutoSearchInput/index.js
index 2fd203486..91ad64b6d 100644
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -109,6 +109,17 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
     showInputLabel: true,
   };
 
+  static getDerivedStateFromProps(props: InternalProps, state: State) {
+    if (props.query !== state.searchValue) {
+      return {
+        ...state,
+        searchValue: props.query || '',
+      };
+    }
+
+    return null;
+  }
+
   constructor(props: InternalProps) {
     super(props);
     invariant(props.inputName, 'The inputName property is required');
@@ -143,14 +154,6 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
     };
   }
 
-  componentWillReceiveProps(nextProps: InternalProps) {
-    const { query } = nextProps;
-
-    if (this.props.query !== query) {
-      this.setState({ searchValue: query || '' });
-    }
-  }
-
   createFiltersFromQuery(query: string) {
     const { location, userAgentInfo } = this.props;
     // Preserve any existing search filters.
```

Here is the failing test case:

  ● /Users/williamdurand/projects/mozilla/addons-frontend/tests/unit/amo/components/TestAutoSearchInput.js › search input › only sets an updated query if it is unique

    expect(received).toMatchObject(expected)

    Expected value to match object:
      {"value": "panda themes"}
    Received:
      {"className": "AutoSearchInput-query", "id": "AutoSearchInput-query", "maxLength": 100, "minLength": 3, "name": "query", "onChange": [Function anonymous], "placeholder": "Find add-ons", "type": "search", "value": "ad blocker"}
    Difference:
    - Expected
    + Received

      Object {
    -   "value": "panda themes",
    +   "value": "ad blocker",
      }

      134 |       root.setProps({ query });
      135 |
    > 136 |       expect(getInputProps(root)).toMatchObject({ value: typedQuery });
          |                                   ^
      137 |     });
      138 |
      139 |     it('handles typing text into the input', () => {

      at Object.<anonymous> (tests/unit/amo/components/TestAutoSearchInput.js:136:35)

From what I understand, the test case tries to cover the following scenario:

1. set a value for a `query` prop
2. this value is copied to the component's state in the constructor
3. the state is updated via a child component
4. new props are given to the component
5. we expect that the `query` prop is NOT copied to the component's state that time

I find this scenario a bit weird because we assume a state property can be updated by two different "sources", yet not always.

Is that a real life scenario?

willdurand:
Fixes #6771

---

This PR also updates some related files.

### DIFF
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -2,12 +2,12 @@
 /* eslint-disable react/prop-types, react/no-unused-prop-types */
 import { oneLine } from 'common-tags';
 import makeClassName from 'classnames';
-import invariant from 'invariant';
 import defaultDebounce from 'lodash.debounce';
 import * as React from 'react';
 import Autosuggest from 'react-autosuggest';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
+import { withRouter } from 'react-router-dom';
 
 import SearchSuggestion from 'amo/components/SearchSuggestion';
 import { withFixedErrorHandler } from 'core/errorHandler';
@@ -53,10 +53,8 @@ type Props = {|
   // This is the name property of the <input> tag.
   inputName: string,
   inputPlaceholder?: string,
-  location?: ReactRouterLocationType,
-  onSearch: (SearchFilters) => void,
+  onSearch?: (SearchFilters) => void,
   onSuggestionSelected: (SuggestionType) => void,
-  query?: string,
   // This is accessibility text for what selecting a suggestion will do.
   selectSuggestionText: string,
   showInputLabel?: boolean,
@@ -75,6 +73,7 @@ type InternalProps = {|
   dispatch: DispatchFunc,
   errorHandler: ErrorHandlerType,
   i18n: I18nType,
+  location: ReactRouterLocationType,
 |};
 
 type State = {|
@@ -111,16 +110,6 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
 
   constructor(props: InternalProps) {
     super(props);
-    invariant(props.inputName, 'The inputName property is required');
-    invariant(props.onSearch, 'The onSearch property is required');
-    invariant(
-      props.onSuggestionSelected,
-      'The onSuggestionSelected property is required',
-    );
-    invariant(
-      props.selectSuggestionText,
-      'The selectSuggestionText property is required',
-    );
 
     this.dispatchAutocompleteStart = this.props.debounce(
       ({ filters }) => {
@@ -139,16 +128,16 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
 
     this.state = {
       autocompleteIsOpen: false,
-      searchValue: props.query || '',
+      searchValue: this.getSearchValueFromProps(props),
     };
   }
 
-  componentWillReceiveProps(nextProps: InternalProps) {
-    const { query } = nextProps;
-
-    if (this.props.query !== query) {
-      this.setState({ searchValue: query || '' });
+  getSearchValueFromProps({ location, inputName }: InternalProps) {
+    if (!location.query) {
+      return '';
     }
+
+    return location.query[inputName] || '';
   }
 
   createFiltersFromQuery(query: string) {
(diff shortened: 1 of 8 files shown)


src/amo/components/AutoSearchInput/index.js
```diff
@@ -111,16 +110,6 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
 
   constructor(props: InternalProps) {
     super(props);
-    invariant(props.inputName, 'The inputName property is required');
```
kumar303: 👍 Cool. We have Flow coverage for all calling components now 🎉

src/amo/components/AutoSearchInput/index.js
```diff
@@ -236,8 +225,12 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
       this.searchInput.blur();
     }
 
-    const filters = this.createFiltersFromQuery(this.state.searchValue);
-    this.props.onSearch(filters);
+    const { onSearch } = this.props;
+
+    if (onSearch) {
+      const filters = this.createFiltersFromQuery(this.state.searchValue);
+      onSearch(filters);
```
kumar303: I don't see test coverage for not calling this when `onSearch` is undefined. Even though we have Flow it would be nice to get a test for it.

tests/unit/amo/components/TestAutoSearchInput.js
```diff
@@ -108,30 +114,33 @@ describe(__filename, () => {
...
 
-      const query = 'ad blocker';
-      root.setProps({ query });
+      const root = render({ location });
 
-      expect(getInputProps(root)).toMatchObject({ value: query });
+      expect(root).toHaveState('searchValue', '');
     });
 
     it('only sets an updated query if it is unique', () => {
```
kumar303: This spec could use an update. I'm not sure if it even made sense with the old code 😛

src/amo/components/AutoSearchInput/index.js
willdurand: If it is `undefined`, how do we test it? 🤔

willdurand:
> It looks good so far but I'm having trouble testing it due to [mozilla/addons-server#9919](https://github.com/mozilla/addons-server/issues/9919)

Interesting. I did not hit this when I tested myself.

src/amo/components/AutoSearchInput/index.js
kumar303: An adequate test would be to trigger a search with an undefined `onSearch` prop and make sure no errors are thrown. For example, this would have thrown an error:

```diff
diff --git a/src/amo/components/AutoSearchInput/index.js b/src/amo/components/AutoSearchInput/index.js
index dc99b8af4..1ca6c95b7 100644
--- a/src/amo/components/AutoSearchInput/index.js
+++ b/src/amo/components/AutoSearchInput/index.js
@@ -227,10 +227,8 @@ export class AutoSearchInputBase extends React.Component<InternalProps, State> {
 
     const { onSearch } = this.props;
 
-    if (onSearch) {
-      const filters = this.createFiltersFromQuery(this.state.searchValue);
-      onSearch(filters);
-    }
+    const filters = this.createFiltersFromQuery(this.state.searchValue);
+    onSearch(filters);
   };
 
   handleSearchChange = (
```

However, I thought about it more. I think it's OK to just rely on type safety here. No need to add a test.

tests/unit/amo/components/TestAutoSearchInput.js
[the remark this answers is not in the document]
willdurand: wfm

PR #6844, opened 2018-11-06
Only self-canonicalized pages must have hreflang

### LINKED ISSUE #6842: Hreflang fix (only self-canonicalized pages must have hreflang)

Our new hreflang setup has one flaw. Its producing a lot of "no return tag" errors which can confuse the whole setup.
screen shot 2018-11-06 at 14 32 11

They appear on pages with a parameter because these pages are not self-canonicalized and tehrefor they get no returning hreflang from the linked pages.

Example: 

/af/firefox/addon/image-search-pro/?src=hotness
canonical: /af/firefox/addon/image-search-pro/
hreflang to: /en-US/firefox/addon/image-search-pro/

/en-US/firefox/addon/image-search-pro/ returns of course only to /af/firefox/addon/image-search-pro/ and therefor the setup is broken.

Pages not self canonicalized must not have any hreflang in the head-section. And they do not need it, as a crawler will follow the canonical link in the head and then process the hreflang on the page pointed on by the canonical. Thats why the rule is as well, that no 301 or 404 must be included in any hreflang: it cant return the hreflang link and only a complete redundant hreflang setup garantees to be interpreted correctly.

willdurand:
Fixes #6842

### DIFF
--- a/tests/unit/amo/components/TestHeadLinks.js
+++ b/tests/unit/amo/components/TestHeadLinks.js
@@ -23,7 +23,7 @@ describe(__filename, () => {
     'renders alternate links with hreflang for %s',
     (clientApp) => {
       const baseURL = 'https://example.org';
-      const lang = 'en-US';
+      const lang = 'fr';
       const to = '/some-url';
 
       const _hrefLangs = ['fr', 'en-US'];
@@ -55,7 +55,7 @@ describe(__filename, () => {
   it('renders alternate links for aliased locales', () => {
     const baseURL = 'https://example.org';
     const clientApp = CLIENT_APP_FIREFOX;
-    const lang = 'en-US';
+    const lang = 'de';
     const to = '/some-url';
 
     const _hrefLangs = ['x-default'];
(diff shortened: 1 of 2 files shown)


tests/unit/amo/components/TestHeadLinks.js
```diff
@@ -23,11 +23,16 @@ describe(__filename, () => {
...
       const baseURL = 'https://example.org';
+      const lang = 'en-US';
       const to = '/some-url';
 
       const _hrefLangs = ['fr', 'en-US'];
       const _config = getFakeConfig({ baseURL });
-      const { store } = dispatchClientMetadata({ clientApp });
+      const { store } = dispatchClientMetadata({
+        clientApp,
+        lang,
```
bobsilverberg: Nit: I was trying to follow why these changes were needed to the test, so I commented out this line and found that the test still passed. I realized that this is because `en-US` is the default value for `lang` in `dispatchClientMetadata`, so maybe it would be better to set `lang` to a different value than `en-US`.

PR #6848, opened 2018-11-06
Update core/utils/addons.js to use a direct version

### LINKED ISSUE #6846: Update core/utils/addons.js to use a direct version
This supports #2770. As part of removing `current_version` and `platformFiles` from `addon` and using an actual version from the versions redux state for that data, we need to update the code in `core/utils/addons.js` to not expect to find that data in an addon.

For QA: Please check installing add-ons on both AMO and disco pane.

bobsilverberg:
Fixes #6846

### DIFF
--- a/src/amo/components/AddonHead/index.js
+++ b/src/amo/components/AddonHead/index.js
@@ -18,6 +18,7 @@ import {
 } from 'core/constants';
 import translate from 'core/i18n/translate';
 import { getPreviewImage } from 'core/imageUtils';
+import { getVersionById } from 'core/reducers/versions';
 import { getAddonJsonLinkedData } from 'core/utils/addons';
 import type { AppState } from 'amo/store';
 import type { AddonVersionType } from 'core/reducers/versions';
@@ -27,12 +28,12 @@ import type { AddonType } from 'core/types/addons';
 type Props = {|
   _getAddonJsonLinkedData?: typeof getAddonJsonLinkedData,
   addon: AddonType | null,
-  currentVersion: AddonVersionType | null,
 |};
 
 type InternalProps = {|
   ...Props,
   clientApp: string,
+  currentVersion: AddonVersionType | null,
   i18n: I18nType,
   lang: string,
 |};
@@ -175,11 +176,20 @@ export class AddonHeadBase extends React.Component<InternalProps> {
   }
 }
 
-const mapStateToProps = (state: AppState) => {
+const mapStateToProps = (state: AppState, ownProps: Props) => {
+  const { addon } = ownProps;
   const { clientApp, lang } = state.api;
+  let currentVersion = null;
+  if (addon && addon.currentVersionId) {
+    currentVersion = getVersionById({
+      id: addon.currentVersionId,
+      state: state.versions,
+    });
+  }
 
   return {
     clientApp,
+    currentVersion,
     lang,
   };
 };
--- a/src/amo/components/AddonHead/index.js
+++ b/src/amo/components/AddonHead/index.js
@@ -20,11 +20,14 @@ import translate from 'core/i18n/translate';
 import { getPreviewImage } from 'core/imageUtils';
 import { getAddonJsonLinkedData } from 'core/utils/addons';
 import type { AppState } from 'amo/store';
+import type { AddonVersionType } from 'core/reducers/versions';
 import type { I18nType } from 'core/types/i18n';
 import type { AddonType } from 'core/types/addons';
 
 type Props = {|
+  _getAddonJsonLinkedData?: typeof getAddonJsonLinkedData,
   addon: AddonType | null,
+  currentVersion: AddonVersionType | null,
 |};
 
 type InternalProps = {|
@@ -35,6 +38,10 @@ type InternalProps = {|
 |};
 
 export class AddonHeadBase extends React.Component<InternalProps> {
+  static defaultProps = {
+    _getAddonJsonLinkedData: getAddonJsonLinkedData,
+  };
+
   getPageTitle() {
     const { addon, clientApp, i18n, lang } = this.props;
 
@@ -133,7 +140,8 @@ export class AddonHeadBase extends React.Component<InternalProps> {
   }
 
   render() {
-    const { addon } = this.props;
+    const { _getAddonJsonLinkedData, addon, currentVersion } = this.props;
+    invariant(_getAddonJsonLinkedData, '_getAddonJsonLinkedData is required.');
 
     if (!addon) {
       return null;
@@ -149,7 +157,7 @@ export class AddonHeadBase extends React.Component<InternalProps> {
           <title>{this.getPageTitle()}</title>
 
           <script type="application/ld+json">
-            {JSON.stringify(getAddonJsonLinkedData({ addon }))}
+            {JSON.stringify(_getAddonJsonLinkedData({ addon, currentVersion }))}
           </script>
         </Helmet>
 
--- a/src/amo/components/AddonHead/index.js
+++ b/src/amo/components/AddonHead/index.js
@@ -26,12 +26,12 @@ import type { I18nType } from 'core/types/i18n';
 import type { AddonType } from 'core/types/addons';
 
 type Props = {|
-  _getAddonJsonLinkedData?: typeof getAddonJsonLinkedData,
   addon: AddonType | null,
 |};
 
 type InternalProps = {|
   ...Props,
+  _getAddonJsonLinkedData: typeof getAddonJsonLinkedData,
   clientApp: string,
   currentVersion: AddonVersionType | null,
   i18n: I18nType,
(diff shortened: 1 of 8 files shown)


src/amo/components/AddonHead/index.js
```diff
@@ -19,11 +19,14 @@ import translate from 'core/i18n/translate';
...
 import { getAddonJsonLinkedData } from 'core/utils/addons';
 import type { AppState } from 'amo/store';
+import type { AddonVersionType } from 'core/reducers/versions';
 import type { I18nType } from 'core/types/i18n';
 import type { AddonType } from 'core/types/addons';
 
 type Props = {|
+  _getAddonJsonLinkedData?: typeof getAddonJsonLinkedData,
   addon: AddonType | null,
+  currentVersion: AddonVersionType | null,
```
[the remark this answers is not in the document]
bobsilverberg: I explained this a bit at https://github.com/mozilla/addons-frontend/pull/6813#discussion_r232242077, although this case is different, because I am  allowing a `null` for `currentVersion` in the props. In this case I guess I was just thinking that, again, this would always be called from a component that has a `currentVersion` prop, so why not just pass it, rather than ask redux to retrieve it yet again. It seems like less processing to do that, but maybe that's not something to be concerned about. Is our philosophy that we should always retrieve data from the redux store rather than accept it via a prop, even if the called will have the same data in a prop of its own?

src/core/installAddon.js
```diff
@@ -216,6 +217,7 @@ type WithInstallHelpersProps = {|
   _installTheme: typeof installTheme,
   _tracking: typeof tracking,
   addon: AddonType,
+  currentVersion: AddonVersionType,
```
willdurand: So, I know why you did that but we should not do that. Injecting such props like this is super obscure.

I know we do that for `addon` (which would be tricky to change) and `userAgentInfo` (which could probably be read from Redux as described below) already, but that does not mean we should do that. I was able to remove some props in a previous refactoring, and that was a bit painful, so let's not do that again :p 

I believe we could create a `mapStateToProps` function in the `withInstallHelpers()` HOC: https://github.com/mozilla/addons-frontend/blob/9f51f3ae053720970e7c8d355f00b718b8902045/src/core/installAddon.js#L566-L569

Here is a sketch of patch:

```diff
diff --git a/src/core/installAddon.js b/src/core/installAddon.js
index 0700509df..a255832e7 100644
--- a/src/core/installAddon.js
+++ b/src/core/installAddon.js
@@ -563,8 +563,24 @@ export function withInstallHelpers({
       WrappedComponent,
     )})`;
 
+    const mapStateToProps = (
+      // We cannot use `AppState` because this HOC is shared between `amo` and
+      // `disco`.
+      state: {| versions: VersionsState |},
+      ownProps: WithInstallHelpersProps
+    ) => {
+      const currentVersion = getVersionById({
+        id: ownProps.addon.currentVersionId,
+        state: state.versions,
+      });
+
+      return {
+        currentVersion,
+      };
+    };
+
     return connect(
-      undefined,
+      mapStateToProps,
       _makeMapDispatchToProps({ WrappedComponent, defaultInstallSource }),
     )(WithInstallHelpers);
   };

```

If that works, we could also move `userAgentInfo` there (in a follow-up patch). WDYT?

bobsilverberg:
Thanks for the review @willdurand. This is ready for another look.

src/core/installAddon.js
bobsilverberg: It seems to work well, thanks @willdurand. I've opened https://github.com/mozilla/addons-frontend/issues/6878 as a follow-up.

willdurand: That's great news! 🎉

src/amo/components/AddonHead/index.js
```diff
@@ -133,7 +141,8 @@ export class AddonHeadBase extends React.Component<InternalProps> {
   }
 
   render() {
-    const { addon } = this.props;
+    const { _getAddonJsonLinkedData, addon, currentVersion } = this.props;
+    invariant(_getAddonJsonLinkedData, '_getAddonJsonLinkedData is required.');
```
willdurand: Do you need this invariant statement because of flow?

src/core/installAddon.js
```diff
@@ -563,8 +572,26 @@ export function withInstallHelpers({
       WrappedComponent,
     )})`;
 
+    const mapStateToProps = (
+      state: AmoAppState | DiscoAppState,
```
willdurand: Yep, I was thinking about that too. Looks like it works, so 🏅

src/core/installAddon.js
```diff
@@ -563,8 +572,26 @@ export function withInstallHelpers({
...
+      ownProps: WithInstallHelpersProps,
+    ) => {
+      const { addon } = ownProps;
+      const currentVersion =
+        addon && addon.currentVersionId
```
willdurand: Why do you need the `addon &&` part? An addon must be provided. It is enforced twice, once with the Flow type (being `AddonType` only) and with an `invariant` call.

src/core/utils/addons.js
```diff
@@ -43,16 +44,19 @@ export const getErrorMessage = ({
 export const getFileHash = ({
   addon,
   installURL,
+  version,
 }: {|
   addon: AddonType,
   installURL: string,
+  version?: AddonVersionType,
```
willdurand: I thought you wanted to use `| void` instead of `?:`?

tests/unit/core/test_installAddon.js
```diff
@@ -85,7 +85,7 @@ function componentWithInstallHelpers({
 }
 
 const defaultProps = (overrides = {}) => {
-  const { store } = createStore();
+  const store = overrides.store || createStore().store;
```
willdurand: I suppose the alternative would be:

```js
const defaultProps = ({ store = createStore().store, ...overrides } = {}) => {
  sinon.stub(store, 'dispatch');
  ...
};
```

tests/unit/core/test_installAddon.js
```diff
@@ -992,29 +1001,37 @@ describe(__filename, () => {
...
         const { root } = renderWithInstallHelpers({
           _addonManager: fakeAddonManager,
-          addon,
           defaultInstallSource,
+          store,
         });
         const { install } = root.instance().props;
 
-        return install(addon).then(() => {
+        return install().then(() => {
```
willdurand: mmh, interesting.

src/core/installAddon.js
willdurand: If `addon &&` is required for some reasons, there is a bug somewhere and we should treat `addon` as a nullable. If it is only because of the two tests case that are failing in `test_installAddon.js`, then I guess we should fix them (and we can remove the spread of `addon` there).

src/amo/components/AddonHead/index.js
bobsilverberg: Yes, because it's optional Flow thinks it could be `undefined` here. I guess Flow doesn't understand `defaultProps`. I can remove the `invariant` if I create a new const here and set it to `_getAddonJsonLinkedData || getAddonJsonLinkedData` so I've done that in the next commit, but I feel like that's a bit ugly.

src/core/installAddon.js
bobsilverberg: There's a problem somewhere because it's not just the tests that fail without this. It looks like the Flow definition at https://github.com/mozilla/addons-frontend/blob/2018.11.08/src/core/installAddon.js#L218 is incorrect, as I'm getting a `null` for `addon` when running the code. That may be a bug introduced by my patch - I'll dig deeper. It doesn't help that `src/amo/pages/Addon/index.js` isn't covered by Flow.

src/core/utils/addons.js
bobsilverberg: FTR it was @kumar303, not me that had an issue with it, but yes, this looks like a case where `AddonVersionType | void` would be better. Although looking closer, it seems like `version` could be required as I don't see any cases where it wouldn't exist when calling this code. I'll look into that too.

src/amo/components/AddonHead/index.js
willdurand: hum, interesting. What if you move `_getAddonJsonLinkedData` into the `InternalProps` and make it not optional?

src/core/installAddon.js
willdurand: >  It doesn't help that `src/amo/pages/Addon/index.js` isn't covered by Flow.

yep :(

I can see why the Flow type is incorrect: we don't have an add-on when we load an add-on detail page.

src/amo/components/AddonHead/index.js
bobsilverberg: That works, thanks!

src/core/installAddon.js
bobsilverberg: As discussed on IRC leaving the `addon &&` for now is probably the best solution given our current code.

tests/unit/core/test_installAddon.js
bobsilverberg: Yes, that feels better. Thanks!

tests/unit/core/test_installAddon.js
bobsilverberg: Yeah, I tried and tried to understand why we were passing `addon` into `install` at all, and could not. Then discovered we don't even need it.

PR #6851, opened 2018-11-06
Add admin user link to user profile page

### LINKED ISSUE #6487: Add admin user link to user profile page
(Similar to #4632 but for users)

This link would appear in the left box, somewhere below the ratings (along with "Report user for abuse" and "Edit user" buttons):

"Admin user". Goes to `/admin/models/users/userprofile/ID/` (note that the slug won't work). Requires `Users:Edit` and `Admin:Tools` to appear.

bobsilverberg:
Fixes #6487 

Before:

screenshot 2018-11-06 12 55 25

After:

screenshot 2018-11-06 12 51 41

### DIFF
--- a/src/amo/pages/UserProfile/index.js
+++ b/src/amo/pages/UserProfile/index.js
@@ -28,6 +28,7 @@ import Paginate from 'core/components/Paginate';
 import {
   ADDON_TYPE_EXTENSION,
   ADDON_TYPE_THEME,
+  ADMIN_TOOLS,
   USERS_EDIT,
 } from 'core/constants';
 import { withFixedErrorHandler } from 'core/errorHandler';
@@ -65,6 +66,7 @@ type Props = {|
 
 type InternalProps = {|
   ...Props,
+  canAdminUser: boolean,
   canEditProfile: boolean,
   currentUser: UserType | null,
   dispatch: DispatchFunc,
@@ -264,6 +266,7 @@ export class UserProfileBase extends React.Component<InternalProps> {
 
   render() {
     const {
+      canAdminUser,
       canEditProfile,
       errorHandler,
       i18n,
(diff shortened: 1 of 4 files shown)


src/amo/pages/UserProfile/index.js
```diff
@@ -416,6 +419,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
                 {i18n.gettext('Edit profile')}
               </Button>
             ) : null}
+
+            {canAdminUser && user ? (
+              <Button
+                className="UserProfile-admin-link"
+                buttonType="neutral"
+                href={`/admin/models/users/userprofile/${user.id}/`}
```
rebmullin: I know in the issue it says this url won't work, right? It wasn't clear to me why tho...
Anyway, I'm just wondering if this is something we should hide till it does work?

src/amo/pages/UserProfile/index.js
```diff
@@ -416,6 +419,18 @@ export class UserProfileBase extends React.Component<InternalProps> {
                 {i18n.gettext('Edit profile')}
               </Button>
             ) : null}
+
+            {canAdminUser && user ? (
```
rebmullin: nit: since you check for the user object when setting `canAdminUser` (https://github.com/mozilla/addons-frontend/pull/6851/files#diff-8ac5e13d9f69883f4468abd7d2787b98R493), is this needed again?

src/amo/pages/UserProfile/index.js
bobsilverberg: I think @wagnerand was just pointing out that the `username` (he said "slug") won't work and that's why we needed to use the `id` instead. The link does work - I tested it.

src/amo/pages/UserProfile/index.js
bobsilverberg: Unfortunately Flow doesn't seem to be able to keep track of that, so while we know that we don't really need this check, Flow seems to require it.

bobsilverberg:
Thanks for the review @rebmullin! I've added some responses to your comments.

src/amo/pages/UserProfile/index.js
rebmullin: hmm doesn't work for me. Is it behind the vpn? maybe that's why I can't see it...

src/amo/pages/UserProfile/index.js
rebmullin: ahh..makes sense then. thanks!

src/amo/pages/UserProfile/index.js
bobsilverberg: Yes, it's behind the VPN

bobsilverberg:
Thanks for the review @rebmullin!

PR #6857, opened 2018-11-07
Static themes and Android

### LINKED ISSUE #6689: Withhold reference to themes from Android
When Lightweight themes (LWTS) migrate to static themes there will be no themes on AMO that are installable by Android. 

Because of that we need to hide the references to themes for the Android site. 

- [x] Hide the navigations for themes on the Android site
- [x] Hide the theme search filters.
- [x] Hide the featured themes cards on the homepage
- [x] Hide the "Change the way Firefox looks with themes." card on the homepage.
- [x] Double-check that the search does not return static themes for android theme searches - e.g. if someone gets this by making a manual search.
- [x] theme category pages should return a 404: http://addons-dev.allizom.org/en-US/android/themes/fashion/
- [x] theme categories page should return a 404: http://addons-dev.allizom.org/en-US/android/themes/categories/
- [x] theme landing page should return a 404: http://addons-dev.allizom.org/en-US/android/themes/
- [x] a not found page should not have a link to the themes page

All changes should be config based so that we can reverse this in future.

willdurand:
Fixes #6689

---

I believe everything is covered. I tried to browse this version locally and I cannot get any static themes stuff.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -350,6 +350,8 @@ module.exports = {
 
   enableFeatureHomeHeroGuides: false,
 
+  // This flag is enabled by default because the feature already exists and we
+  // want to disable it.
   // See: https://github.com/mozilla/addons-frontend/issues/6689
   enableFeatureStaticThemesForAndroid: true,
 
--- a/config/default.js
+++ b/config/default.js
@@ -110,6 +110,7 @@ module.exports = {
     'enableDevTools',
     'enableFeatureExperienceSurvey',
     'enableFeatureHomeHeroGuides',
+    'enableFeatureStaticThemesForAndroid',
     'enableRequestID',
     'experiments',
     'fxaConfig',
@@ -348,6 +349,9 @@ module.exports = {
   dismissedExperienceSurveyCookieName: 'dismissedExperienceSurvey',
 
   enableFeatureHomeHeroGuides: false,
+
+  // See: https://github.com/mozilla/addons-frontend/issues/6689
+  enableFeatureStaticThemesForAndroid: true,
 
   // The withExperiment HOC relies on this config to enable/disable A/B
   // experiments.
--- a/src/amo/components/SectionLinks/index.js
+++ b/src/amo/components/SectionLinks/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import config from 'config';
 import makeClassName from 'classnames';
 import * as React from 'react';
 import { compose } from 'redux';
@@ -29,6 +30,7 @@ import type { ReactRouterHistoryType } from 'core/types/router';
 import './styles.scss';
 
 type Props = {|
+  _config: typeof config,
   className?: string,
   clientApp: string,
   dispatch: DispatchFunc,
@@ -38,6 +40,10 @@ type Props = {|
 |};
 
 export class SectionLinksBase extends React.Component<Props> {
+  static defaultProps = {
+    _config: config,
+  };
+
   setClientApp = (event: Object) => {
     event.preventDefault();
 
@@ -50,16 +56,19 @@ export class SectionLinksBase extends React.Component<Props> {
   };
 
   render() {
-    const { className, clientApp, i18n, viewContext } = this.props;
+    const { _config, className, clientApp, i18n, viewContext } = this.props;
     const isExploring = [VIEW_CONTEXT_EXPLORE, VIEW_CONTEXT_HOME].includes(
       viewContext,
     );
 
     let forBrowserNameText;
+    let showThemes = true;
+
     if (clientApp === CLIENT_APP_FIREFOX) {
       forBrowserNameText = i18n.gettext('for Firefox');
     } else if (clientApp === CLIENT_APP_ANDROID) {
       forBrowserNameText = i18n.gettext('for Android');
+      showThemes = _config.get('enableFeatureStaticThemesForAndroid');
     }
 
     return (
@@ -88,16 +97,22 @@ export class SectionLinksBase extends React.Component<Props> {
             {i18n.gettext('Extensions')}
           </Link>
         </li>
-        <li>
-          <Link
-            className={makeClassName('SectionLinks-link', {
-              'SectionLinks-link--active': viewContext === ADDON_TYPE_THEME,
-            })}
-            to={`/${visibleAddonType(ADDON_TYPE_THEME)}/`}
-          >
-            {i18n.gettext('Themes')}
-          </Link>
-        </li>
+        {showThemes && (
+          <li>
+            <Link
+              className={makeClassName(
+                'SectionLinks-link',
+                'SectionLinks-link-theme',
+                {
+                  'SectionLinks-link--active': viewContext === ADDON_TYPE_THEME,
+                },
+              )}
+              to={`/${visibleAddonType(ADDON_TYPE_THEME)}/`}
+            >
+              {i18n.gettext('Themes')}
+            </Link>
+          </li>
+        )}
         <li>
           <DropdownMenu
             className="SectionLinks-link SectionLinks-dropdown"
--- a/src/amo/components/SectionLinks/index.js
+++ b/src/amo/components/SectionLinks/index.js
@@ -7,6 +7,7 @@ import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 
 import Link from 'amo/components/Link';
+import { shouldShowThemes } from 'amo/utils';
 import { setClientApp } from 'core/actions';
 import {
   ADDON_TYPE_EXTENSION,
@@ -62,13 +63,10 @@ export class SectionLinksBase extends React.Component<Props> {
     );
 
     let forBrowserNameText;
-    let showThemes = true;
-
     if (clientApp === CLIENT_APP_FIREFOX) {
       forBrowserNameText = i18n.gettext('for Firefox');
     } else if (clientApp === CLIENT_APP_ANDROID) {
       forBrowserNameText = i18n.gettext('for Android');
-      showThemes = _config.get('enableFeatureStaticThemesForAndroid');
     }
 
     return (
@@ -97,7 +95,7 @@ export class SectionLinksBase extends React.Component<Props> {
             {i18n.gettext('Extensions')}
           </Link>
         </li>
-        {showThemes && (
+        {shouldShowThemes({ _config, clientApp }) && (
           <li>
             <Link
               className={makeClassName(
--- a/src/amo/components/SuggestedPages/index.js
+++ b/src/amo/components/SuggestedPages/index.js
@@ -1,19 +1,41 @@
+/* @flow */
+import config from 'config';
 import * as React from 'react';
-import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { compose } from 'redux';
 
 import Link from 'amo/components/Link';
-import { ADDON_TYPE_EXTENSION, ADDON_TYPE_THEME } from 'core/constants';
+import {
+  ADDON_TYPE_EXTENSION,
+  ADDON_TYPE_THEME,
+  CLIENT_APP_ANDROID,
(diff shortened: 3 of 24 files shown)


rebmullin:
just a quick visual review on the frontend and I have a few questions/notes:

- the extensions page is 404'ing on my end - does this happen for you?

- if you go to a static theme directly, what do you expect to happen? Will the install button just be disabled and a Notice displayed?

willdurand:
> * the extensions page is 404'ing on my end - does this happen for you?

Fixed!

> * if you go to a static theme directly, what do you expect to happen? Will the install button just be disabled and a Notice displayed?

I think so yeah, but I'll double-check with @muffinresearch.

willdurand:
> > * the extensions page is 404'ing on my end - does this happen for you?
> 
> Fixed!

ahm. not fixed actually, I have one more thing to fix...

config/default.js
```diff
@@ -110,6 +110,7 @@ module.exports = {
     'enableDevTools',
     'enableFeatureExperienceSurvey',
     'enableFeatureHomeHeroGuides',
+    'enableFeatureStaticThemesForAndroid',
```
rebmullin: maybe it's just me but I find this name a little confusing since we currently are not displaying static themes on Android, yet you have "true" here…

maybe it can be renamed to just "enableFeatureThemesForAndroid" ?

willdurand: That sounds correct to me, the feature is about enabling static themes for Android, which is the case by default but not in -dev anymore. When we'll push that to production, the default config will be `false`.

rebmullin: oh ok..I didn't realize static themes were on android on stage etc. I was looking at dev and saw they weren't there.

src/amo/components/SuggestedPages/index.js
```diff
@@ -1,19 +1,41 @@
...
   };
 
   render() {
-    const { i18n } = this.props;
+    const { _config, clientApp, i18n } = this.props;
+
+    const showThemes =
+      clientApp === CLIENT_APP_ANDROID
+        ? _config.get('enableFeatureStaticThemesForAndroid')
+        : true;
```
rebmullin: is it worth it to make this a little helper function since you use this many times?

tests/unit/amo/pages/TestCategory.js
```diff
@@ -749,4 +750,39 @@ describe(__filename, () => {
...
+
+  it('renders a 404 when clientApp is Android and enableFeatureStaticThemesForAndroid is false', () => {
+    store = dispatchClientMetadata({ clientApp: CLIENT_APP_ANDROID }).store;
+    const _config = getFakeConfig({
+      enableFeatureStaticThemesForAndroid: false,
+    });
+
+    const root = render({ _config, store });
+
+    expect(root.find(NotFound)).toHaveLength(1);
```
rebmullin: should we add this too: `expect(root).toHaveClassName('Category');`

willdurand:
Ok, I overlooked a few things obviously but that should be better now!

tests/unit/amo/pages/TestCategory.js
rebmullin: sorry I meant `expect(root).not.toHaveClassName('Category');` or `expect(root.find('Category')).toHaveLength(0);`

tests/unit/amo/pages/TestLandingPage.js
```diff
@@ -614,4 +627,54 @@ describe(__filename, () => {
...
+
+  it('renders a 404 when clientApp is Android and enableFeatureStaticThemesForAndroid is false', () => {
+    store = dispatchClientMetadata({ clientApp: CLIENT_APP_ANDROID }).store;
+    const _config = getFakeConfig({
+      enableFeatureStaticThemesForAndroid: false,
+    });
+
+    const root = renderTheme({ _config, store });
+
+    expect(root.find(NotFound)).toHaveLength(1);
```
rebmullin: should we add `expect(root.find('LandingPage')).toHaveLength(0);` here ?

rebmullin:
I was just double checking things and came across a static theme in a collection, for example:

https://addons-dev.allizom.org/en-US/android/collections/rebdev-test/testing1234/?page=1&collection_sort=-added - is this okay?

config/default.js
willdurand: Ok, I will add a comment then.

src/amo/components/SuggestedPages/index.js
willdurand: Yep, thanks for the idea!

willdurand:
> > * if you go to a static theme directly, what do you expect to happen? Will the install button just be disabled and a Notice displayed?
> 
> I think so yeah, but I'll double-check with @muffinresearch.

@rebmullin so, I'll file a new issue for that.

willdurand:
Updated!

config/default.js
rebmullin: thank you :)

tests/unit/amo/pages/TestCategoriesPage.js
```diff
@@ -53,4 +74,52 @@ describe(__filename, () => {
...
+    expect(root.find(Categories)).toHaveLength(1);
+  });
+
+  it('does not return a 404 when addonType is not "theme"', () => {
+    const { store } = dispatchClientMetadata({ clientApp: CLIENT_APP_ANDROID });
+    const _config = getFakeConfig({
+      enableFeatureStaticThemesForAndroid: false,
+    });
+
+    const root = render({ _config, store });
```
rebmullin: I know extension is the default type (right?) but I think it would be a little clearer if you were to pass in the type here since that's the description for this test case.

src/amo/components/SectionLinks/index.js
```diff
@@ -50,16 +56,19 @@ export class SectionLinksBase extends React.Component<Props> {
...
 
   render() {
-    const { className, clientApp, i18n, viewContext } = this.props;
+    const { _config, className, clientApp, i18n, viewContext } = this.props;
     const isExploring = [VIEW_CONTEXT_EXPLORE, VIEW_CONTEXT_HOME].includes(
       viewContext,
     );
 
     let forBrowserNameText;
+    let showThemes = true;
```
rebmullin: maybe `const showThemes = shouldShowThemes({ _config, clientApp }))` ? or maybe you can just wrap shouldShowThemes itself around the li without assigning it since it's only used 1 time here?

tests/unit/amo/pages/TestCategoriesPage.js
rebmullin: I did just notice that you did have a separate render for theme so maybe this is okay then..

tests/unit/amo/pages/TestCategory.js
```diff
@@ -110,6 +111,51 @@ describe(__filename, () => {
...
+
+    return renderWithCategory({ category, props });
+  };
+
+  const renderAndroidTheme = (props = {}) => {
+    const category = {
+      ...fakeCategory,
+      // This should really be et to CLIENT_APP_ANDROID *but* the reducer
+      // replaces android "theme" categories with firefox categories here:
+      // https://github.com/mozilla/addons-frontend/blob/0ebf7b7fc559dee2cf704bad04905c8c8e430905/src/core/reducers/categories.js#L72-L82
```
rebmullin: sorry for my ignorance but what does "et" mean here?

tests/unit/amo/pages/TestCategoriesPage.js
```diff
@@ -53,4 +74,52 @@ describe(__filename, () => {
 
     expect(root.find(HeadLinks)).toHaveLength(1);
   });
+
+  it('returns a 404 when clientApp is Android and enableFeatureStaticThemesForAndroid is false', () => {
```
rebmullin: should you mention that this is only for themes in the test description?

tests/unit/amo/pages/TestCategory.js
```diff
@@ -749,4 +795,52 @@ describe(__filename, () => {
       expect(root.find(NotFound)).toHaveLength(0);
     });
   });
+
+  it('renders a 404 when clientApp is Android and enableFeatureStaticThemesForAndroid is false', () => {
```
rebmullin: same here - should you mention that this is only for a theme in the test description?

src/amo/pages/Home/index.js
```diff
@@ -282,7 +286,7 @@ export class HomeBase extends React.Component {
           />
         )}
 
-        {includeFeaturedThemes && (
+        {includeFeaturedThemes && showThemes && (
```
rebmullin: what about the FeaturedCollection section (see ~ https://github.com/mozilla/addons-frontend/pull/6857/files#diff-9d92c2d79f19315442c24a9765c04dd7L50)? I've seen themes there too (not a lot I don't think but sometimes). In this PR you can see it was removed https://github.com/mozilla/addons-frontend/commit/c169c8a4e7d87a08f53b3e30e3b38c6afa7a406b#diff-9d92c2d79f19315442c24a9765c04dd7 (I couldn't easily get access to the one where it was added but hopefully this shows what I mean).

Maybe this is okay for now since it's not currently set up this way..? What do you think?

tests/unit/amo/pages/TestLandingPage.js
```diff
@@ -614,4 +627,55 @@ describe(__filename, () => {
 
     expect(root.find(HeadLinks)).toHaveLength(1);
   });
+
+  it('renders a 404 when clientApp is Android and enableFeatureStaticThemesForAndroid is false', () => {
```
rebmullin: I've made a similar comment above but I'm wondering if we should specify in the description that's is for a theme only?

tests/unit/amo/pages/TestCategoriesPage.js
willdurand: That's why I added `renderThemeCategories()` indeed.

tests/unit/amo/pages/TestCategory.js
willdurand: nothing 😭 that should be "set"

tests/unit/amo/pages/TestCategoriesPage.js
willdurand: I'd assume that `renderThemeCategories()` would be enough. The opposite test case explicitly says `does not return a 404 when addonType is not "theme"`

src/amo/pages/Home/index.js
willdurand: I think we cannot hide what we don't know..

tests/unit/amo/pages/TestLandingPage.js
willdurand: Every time is calls a `renderTheme` function and we mention `enableFeatureStaticThemesForAndroid` so I don't think it is needed here.

willdurand:
Fixed a few things and rebased, thanks @rebmullin!

src/amo/pages/Home/index.js
rebmullin: okay I just didn't want anything to sneak in inadvertently later on so just double checking :)

tests/unit/amo/pages/TestCategory.js
rebmullin: lol.

src/amo/pages/Home/index.js
willdurand: Collections are not something "theme" or "extension" specific unfortunately, so we cannot really do anything anyway...

willdurand: oh, interesting. Then, let me file an issue for that because this PR gets too big.

willdurand: Can we defer this change to https://github.com/mozilla/addons-frontend/issues/6870? I think it is fine given that we don't have any featured "theme" collections currently.

willdurand:
Thanks for the quick and great reviews!

